PR #9124, opened 2021-01-15
Make chain cover index bg update go faster

erikjohnston:
We do this by allowing a single iteration to process multiple rooms at a time, as there are often a lot of really tiny rooms, which can massively slow things down.

Basically, instead of processing each room individually and pulling out the events in order of `(depth, stream_ordering)`, we instead pull out events in order of `(room_id, depth, stream_ordering)`, which allows us to process multiple small rooms in one go.

### DIFF
--- a/synapse/storage/databases/main/events_bg_updates.py
+++ b/synapse/storage/databases/main/events_bg_updates.py
@@ -16,6 +16,8 @@
 import logging
 from typing import Dict, List, Optional, Tuple
 
+import attr
+
 from synapse.api.constants import EventContentFields
 from synapse.api.room_versions import KNOWN_ROOM_VERSIONS
 from synapse.events import make_event_from_dict
@@ -26,6 +28,25 @@ from synapse.storage.types import Cursor
 from synapse.types import JsonDict
 
 logger = logging.getLogger(__name__)
+
+
+@attr.s
+class _CalculateChainCover:
+    """Return value for _calculate_chain_cover_txn.
+    """
+
+    # The last room_id/depth/stream processed.
+    room_id = attr.ib(type=str)
+    depth = attr.ib(type=int)
+    stream = attr.ib(type=int)
+
+    # Number of rows processed
+    processed_count = attr.ib(type=int)
+
+    # Map from room_id to last depth/stream processed for each room that we have
+    # processed all events for (i.e. the rooms we can flip the
+    # `has_auth_chain_index` for)
+    finished_room_map = attr.ib(type=Dict[str, Tuple[int, int]])
 
 
 class EventsBackgroundUpdatesStore(SQLBaseStore):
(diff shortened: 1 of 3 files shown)


synapse/storage/databases/main/events_bg_updates.py
```diff
@@ -715,142 +736,37 @@ def get_rejected_events(
     async def _chain_cover_index(self, progress: dict, batch_size: int) -> int:
         """A background updates that iterates over all rooms and generates the
         chain cover index for them.
+
+        Returns:
+            tuple of rows processed, if the background update has finished and
+            the new progress dict.
```
clokep, line 742: I'm having some trouble following what this is trying to say. (Also I think the type hint needs updating?)

synapse/storage/databases/main/events_bg_updates.py
```diff
@@ -860,42 +776,160 @@ def _calculate_auth_chain(
...
+        single_room: bool,
+    ) -> _CalculateChainCover:
+        """Calculate the chain cover for `batch_size` events, ordered by
+        `(room_id, depth, stream)`.
+
+        Args:
+            txn,
+            last_room_id,
+            last_depth,
+            last_stream,
```
clokep, line 829: Can we fill these in?

synapse/storage/databases/main/events_bg_updates.py
```diff
@@ -860,42 +776,160 @@ def _calculate_auth_chain(
...
+        # need to include the room passed in if it's not included in the result
+        # set (as we then know we've processed all events in said room).
+        #
+        # This is the set of rooms that we can now safely flip the
+        # `has_auth_chain_index` bit for.
+        finished_rooms = {
+            row[5]: (row[3], row[4]) for row in rows if row[5] != new_last_room_id
+        }
+        if last_room_id not in finished_rooms and last_room_id != new_last_room_id:
+            finished_rooms[last_room_id] = (last_depth, last_stream)
```
clokep, line 903: This covers the case where the `last_room_id` was already done and no work was necessary this iteration?

synapse/storage/databases/main/events_bg_updates.py
```diff
@@ -860,42 +776,160 @@ def _calculate_auth_chain(
...
+
+        # Put the results in the necessary format for
+        # `_add_chain_cover_index`
+        event_to_room_id = {row[0]: row[5] for row in rows}
+        event_to_types = {row[0]: (row[1], row[2]) for row in rows}
+
+        # Calculate the new last position we've processed up to.
+        new_last_depth = rows[-1][3] if rows else last_depth  # type: int
+        new_last_stream = rows[-1][4] if rows else last_stream  # type: int
+        new_last_room_id = rows[-1][5] if rows else ""  # type: str
```
clokep, line 890: Since we're ordering by room ID first, is this really the last stream ID / depth? Is it possible for an earlier room ID to have a later stream ID?

tests/storage/test_event_chain.py
```diff
@@ -508,79 +506,181 @@ def test_background_update(self):
...
         )
 
         # Ugh, have to reset this flag
-        store.db_pool.updates._all_done = False
+        self.store.db_pool.updates._all_done = False
 
         while not self.get_success(
-            store.db_pool.updates.has_completed_background_updates()
+            self.store.db_pool.updates.has_completed_background_updates()
         ):
```
clokep, line 669: We might want to assert that more than one iteration was necessary.

tests/storage/test_event_chain.py
```diff
@@ -508,79 +506,181 @@ def test_background_update(self):
...
+        self.get_success(
+            self.store.db_pool.runInteraction(
+                "test",
+                self.store._get_auth_chain_difference_using_cover_index_txn,
+                room_id1,
+                states1,
+            )
+        )
+
+    def test_background_update_single_large_room(self):
```
clokep, line 641: Should we have a similar test with multiple rooms? (Or does the test above already have multiple iterations in it?)

synapse/storage/databases/main/events_bg_updates.py
```diff
@@ -28,6 +30,25 @@
 logger = logging.getLogger(__name__)
 
 
+@attr.s
```
clokep, line 33:
```suggestion
@attr.s(slots=True, frozen=True)
```

synapse/storage/databases/main/events_bg_updates.py
erikjohnston: Yup, but we didn't know we were done last time

synapse/storage/databases/main/events_bg_updates.py
erikjohnston: Oh argh, that is a holdover from a previous attempt to make this better :facepalm:

synapse/storage/databases/main/events_bg_updates.py
erikjohnston: They're last in the sense that they're from the last row, and where we want to stream from next time round. In some ways they should be treated as a tuple, but I think if we made them a tuple in the code it'd just confuse things.

clokep: Ah, OK! The three are just a pointer into the ordered list. It isn't meant to be the *maximum*, just the last one processed. 👍

clokep (approved):
Let's give it a try. 👍

erikjohnston:
Woo thanks!

PR #9125, opened 2021-01-15
Remove dependency on distutils

richvdh:
`distutils` is pretty much deprecated these days, and replaced with `setuptools`. It's also annoying because you can't `pip install` it, and it's hard to figure out which debian package we should depend on to make sure it's there.

Since we only use it for a tiny function anyway, let's just vendor said function into our codebase.

I *think* this fixes an issue on debian where upgrading `python3-distutils` could lead to `No module named 'distutils.util'`:
 1. You're following Debian testing, which has recently changed the default python version from 3.7 to 3.9. Currently, you have `python3.7` installed, as well as `python3-distutils==3.7`. You also have `matrix-synapse-py3==1.24` installed (or, equivalently, the `buster` version of 1.25).
 2. You upgrade `python3-distutils` on your system, which removes `/usr/lib/python3.7/distutils/util.py`, but leaves `python3.7` in place.
 3. Synapse now fails with the above import error.

In any case, this seems like a good thing to do.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -31,7 +31,6 @@ Pre-Depends: dpkg (>= 1.16.1)
 Depends:
  adduser,
  debconf,
- python3-distutils|libpython3-stdlib (<< 3.6),
  ${misc:Depends},
  ${shlibs:Depends},
  ${synapse:pydepends},
--- a/changelog.d/9125.misc
+++ b/changelog.d/9125.misc
@@ -1,0 +1,1 @@
+Remove dependency on `distutils`.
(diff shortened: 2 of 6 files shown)


debian/control
```diff
@@ -31,7 +31,6 @@ Pre-Depends: dpkg (>= 1.16.1)
 Depends:
  adduser,
  debconf,
- python3-distutils|libpython3-stdlib (<< 3.6),
```
richvdh, line 34: nb that `python3.X` depends on `libpython3.X-stdlib` so we don't need to pull it in explicitly.

This was here to say that we didn't need a separate `python3-distutils` provided `libpython3-stdlib` was old enough. Now we don't need a separate python3-distutils at all, so the whole line is redundant.

erikjohnston (approved):

PR #9127, opened 2021-01-15
Test OIDC login with multiple OIDC providers

richvdh:
The idea here is to make sure that an OIDC login works correctly when more than one OIDC provider is configured. It extends the existing `MultiSSOTestCase.test_multi_sso_redirect_to_oidc` testcase into a full OIDC login test.

There are three commits here: probably easiest to review individually.

### DIFF
--- a/changelog.d/9127.feature
+++ b/changelog.d/9127.feature
@@ -1,0 +1,1 @@
+Add support for multiple SSO Identity Providers.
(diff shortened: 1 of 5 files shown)


clokep (changes requested):
This looks fine, but tests are failing on Py3.5.

richvdh:
it's failing due to a bug fixed in #9128 (which is why that's passing despite being based on this branch). I could shuffle the PRs around a bit to get them to both pass, but it's likely to be a bit fiddly. wdyt?

clokep:
> turns out it's failing due to a bug fixed in #9128 (which is why that's passing despite being based on this branch). I could shuffle the PRs around a bit to get them to both pass, but it's likely to be a bit fiddly. wdyt?

I'll take a look at that shortly then! 👍

richvdh:
I've squash-merged #9128, which is a superset of the changes in this PR, thus ensuring we don't have a commit on develop which breaks the tests.

The two PRs are somewhat related anyway.

PR #9017, opened 2021-01-04
Add initial support for a "pick your IdP" page

richvdh:
During login, if there are multiple IdPs enabled, offer the user a choice of IdPs.

This is deliberately a bit basic for now. You can see what it looks like at https://matrix.org/_matrix/media/r0/download/sw1v.org/pFnpZJJawgBgsseJuynxximd.

PR #9128, opened 2021-01-15
Fix bugs in handling clientRedirectUrl

richvdh:
- don't drop duplicate query-params, or params with no value (a long-standing issue which has affected all three SSO impls forever)
 - in the IDP-picker, allow utf-8 in the URL (introduced in #9017)

Based on #9127


clokep (approved):
LGTM!

PR #9137, opened 2021-01-16
Fedora Prerequisites are (possibly) inaccurate

compu42:
`python3-devel` is required to build the wheel for Twisted, otherwise it'll just fail. `psycopg2` is also required for Postgresql. I don't believe these come with the "Development Tools" group, but I might be mistaken.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -156,9 +156,11 @@ Installing prerequisites on CentOS 8 or Fedora>26:
 ```sh
 sudo dnf install libtiff-devel libjpeg-devel libzip-devel freetype-devel \
                  libwebp-devel tk-devel redhat-rpm-config \
-                 python3-virtualenv libffi-devel openssl-devel python3-devel phycopg2 
+                 python3-virtualenv libffi-devel openssl-devel python3-devel python3-phycopg2 
 sudo dnf groupinstall "Development Tools"
 ```
+
+Signed-off-by: Joseph Arnault <computerdude90042@outlook.com>
 
 Installing prerequisites on CentOS 7 or Fedora<=25:
 
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -156,7 +156,7 @@ Installing prerequisites on CentOS 8 or Fedora>26:
 ```sh
 sudo dnf install libtiff-devel libjpeg-devel libzip-devel freetype-devel \
                  libwebp-devel tk-devel redhat-rpm-config \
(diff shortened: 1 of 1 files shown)


clokep:
@compu42 Can you [sign-off on your commit](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md#sign-off)? Thanks!

INSTALL.md
```diff
@@ -156,7 +156,7 @@ Installing prerequisites on CentOS 8 or Fedora>26:
 ```sh
 sudo dnf install libtiff-devel libjpeg-devel libzip-devel freetype-devel \
                  libwebp-devel tk-devel redhat-rpm-config \
-                 python3-virtualenv libffi-devel openssl-devel
+                 python3-virtualenv libffi-devel openssl-devel python3-devel phycopg2
```
callahad, line 159: Thank you for the contribution! Could you please double-check these package names? I thought the Fedora package for psycopg2 was named `python3-psycopg2`

callahad (changes requested):

INSTALL.md
compu42: right, my bad, yes it is is `python3-psycopg2`, and I'm uncertain if `libpq-devel` would be better

callahad:
Using Fedora 33 on ARM, I can confirm that `libpq-devel` and `python3-devel` are required for `pip install -e '.[all,test,lint]' to succeed.

Attempting to `pip install --no-binary ':all:' -e '.[all,test,lint]'` reveals further dependencies on `libxml2-devel` and `libxslt-devel`.

It also trips over matrix-org/python-unpaddedbase64#4: our `unpaddedbase64` library doesn't provide a source distribution.

And a few dependencies need `wheel` installed so that `bdist_wheel` is available when building.

I did not need to explicitly install `tk-devel` or `redhat-rpm-config`, which were introduced years ago in d7739c4e375195c20b21d7f43248543ecf4ed3f3 and 53254551f0ce9911251e2a7383abdf29a3ced6eb respectively.

I'll tidy those up and get this merged.

callahad:
Upon looking more closely at install.md, we have *lots* of very, very outdated and ill-maintained information in there.

We should probably purge it and move most distro-specific information to the GitHub wiki or similar.

callahad:
I've continued this, cleaned it up, and pushed an additional commit over at #9322. Thank you, @compu42 !

PR #9144, opened 2021-01-18
Enforce all replication HTTP clients calls use kwargs

erikjohnston:
Should prevent https://github.com/matrix-org/synapse/pull/9130 from happening again

### DIFF
--- a/changelog.d/9144.misc
+++ b/changelog.d/9144.misc
@@ -1,0 +1,1 @@
+Enforce that replication HTTP clients are called with keyword arguments only.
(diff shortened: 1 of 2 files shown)


clokep:
Looks like old-deps is failing with invalid syntax errors. 😢

erikjohnston:
> Running setup.py install for frozendict: finished with status 'error'

Uh oh

clokep:
> > Running setup.py install for frozendict: finished with status 'error'
> 
> Uh oh

This actually seems to be #9146.

clokep (approved):
This should be fine if the build passes.

PR #8950, opened 2020-12-15
Add a maximum size for well-known lookups.

clokep:
I noticed we already had a way to enforce a maximum size when reading a response from Twisted. There's no reason for a .well-known lookup to be large so we should bail after a short amount of data. I chose 50kb in here, but that's somewhat arbitrary, the maximum reasonable size for a well-known response is:

```python
len("""{
    "m.server": "<maximum domain length>:65535"
}
""")
```

Note that the maximum domain length is 253 characters. So this leads to roughly 28 + 253 < 300 bytes. We can bikeshed the maximum size if you'd like.

Note that I don't *think* there's anything else useful to put in there.

PR #9145, opened 2021-01-18
Properly raise the exception when the body exceeds the max size.

### LINKED ISSUE #9132: UnboundLocalError raised when response body exceeds max size
A bug with 1.25.0 was reported in the Synapse Package Maintainer's channel:

> I got a nice Traceback in 1.25.0 after start:
>
>     2021-01-15 20:32:45,345 - synapse.http.matrixfederationclient - 987 - WARNING - GET-25- {GET-O-1} [matrix.org] Requested file is too large > 10485760 bytes
>     2021-01-15 20:32:45,345 - synapse.rest.media.v1.media_repository - 417 - ERROR - GET-25- Failed to fetch remote media matrix.org/cPeSAplLYzzcKlpJjLtwlzrT
>     Traceback (most recent call last):                                                                                     
>       File "/usr/local/lib/python3.7/site-packages/synapse/rest/media/v1/media_repository.py", line 384, in _download_remote_file
>         "allow_remote": "false"                                                                                            
>       File "/usr/local/lib/python3.7/site-packages/synapse/http/matrixfederationclient.py", line 1004, in get_file
>         length,                                                                                                            
>     UnboundLocalError: local variable 'length' referenced before assignment
>
> From a (very) quick glance at the source I would guess that synapse hits `except BodyExceededMaxSize` before the `length = ...` statement in the try-block, which causes the `logger.info` statement to reference a non-existing `length` variable.
> Otherwise things appear to work fine. :D

I suspect this was introduced as part of this commit:

https://github.com/matrix-org/synapse/commit/ff5c4da1289cb5e097902b3e55b771be342c29d6#diff-a53ea02bf6a0cbb98e925b47e71f4ffb149075536543c1eca15366fc65c90aefL977-R983

The diagnosis looks right to me: when `read_body_with_max_size` throws, it skips the `length` assignment statement two lines later. However, we don't raise or return from the exception handler, so we continue through to a logging statement which tries to use the unassigned `length` variable.

clokep:
Fixes #9132

### DIFF
--- a/changelog.d/9145.bugfix
+++ b/changelog.d/9145.bugfix
@@ -1,0 +1,1 @@
+Fix "UnboundLocalError: local variable 'length' referenced before assignment" errors when the response body exceeds the expected size. This bug was introduced in v1.25.0.
--- a/synapse/http/client.py
+++ b/synapse/http/client.py
@@ -724,7 +724,7 @@ class SimpleHttpClient:
                 read_body_with_max_size(response, output_stream, max_size)
             )
         except BodyExceededMaxSize:
-            SynapseError(
+            raise SynapseError(
                 502,
                 "Requested file is too large > %r bytes" % (max_size,),
                 Codes.TOO_LARGE,
--- a/synapse/http/matrixfederationclient.py
+++ b/synapse/http/matrixfederationclient.py
@@ -996,7 +996,7 @@ class MatrixFederationHttpClient:
             logger.warning(
                 "{%s} [%s] %s", request.txn_id, request.destination, msg,
             )
-            SynapseError(502, msg, Codes.TOO_LARGE)
+            raise SynapseError(502, msg, Codes.TOO_LARGE)
         except Exception as e:
             logger.warning(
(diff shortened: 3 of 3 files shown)


richvdh (commented):
:facepalm: 

is there a similar problem in `http/client.py` (https://github.com/matrix-org/synapse/pull/8950/files#diff-66444b92f547d9edb90914be0717bbe92d7b02ce10b65b4546b6c629af563ae9R726) ?

clokep:
Yes...yes there is. 🤦

changelog.d/9145.bugfix
```diff
@@ -0,0 +1 @@
+Fix "UnboundLocalError: local variable 'length' referenced before assignment" errors when the response body exceeds the expected size. This bug was introduced in v1.25.0.
```
clokep, line 1: I think the error here is the same in both cases, although the stack would be different.

richvdh (approved):
lgtm

PR #9110, opened 2021-01-13
Land support for multiple OIDC providers

richvdh:
This is the final step for supporting multiple OIDC providers concurrently.

First of all, we reorganise the config so that you can specify a list of OIDC providers, instead of a single one. Before:

```yaml
oidc_config:
   enabled: true
   issuer: "https://oidc_provider"
   # etc
```

After:

```
oidc_providers:
 - idp_id: prov1
   issuer: "https://oidc_provider"

 - idp_id: prov2
   issuer: "https://another_oidc_provider"
```

The old format is still grandfathered in.

With that done, it's then simply a matter of having `OidcHandler` instantiate a new `OidcProvider` for each configured provider.

There are two commits here: one for the code, and one for the docs/sample config. Probably easier to review them separately

Fixes #8927.

~~Based on #9109.~~

PR #9153, opened 2021-01-18
Fix error messages from OIDC config parsing

richvdh:
#9110 reshuffled the logic for parsing OIDC config, but didn't update the error messages to match.

### DIFF
--- a/synapse/config/oidc_config.py
+++ b/synapse/config/oidc_config.py
@@ -280,7 +280,7 @@ def _parse_oidc_provider_configs(config: JsonDict) -> Iterable["OidcProviderConf
     """
     validate_config(MAIN_CONFIG_SCHEMA, config, ())
 
-    for i, p in enumerate(config.get("oidc_providers")) or []:
+    for i, p in enumerate(config.get("oidc_providers") or []):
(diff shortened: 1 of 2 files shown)


synapse/config/oidc_config.py
```diff
@@ -280,8 +280,8 @@ def _parse_oidc_provider_configs(config: JsonDict) -> Iterable["OidcProviderConf
     """
     validate_config(MAIN_CONFIG_SCHEMA, config, ())
 
-    for p in config.get("oidc_providers") or []:
-        yield _parse_oidc_config_dict(p)
+    for i, p in enumerate(config.get("oidc_providers")) or []:
```
clokep, line 283: I suspect this should be:
```suggestion
    for i, p in enumerate(config.get("oidc_providers") or []):
```

This should handle bogus values better (e.g. `oidc_providers: null` or something.

clokep (commented):
Looks like mypy is unhappy too.

synapse/config/oidc_config.py
richvdh: :facepalm: thanks

clokep (approved):
LGTM.

PR #9154, opened 2021-01-18
Support icons for Identity Providers

richvdh:
~~Based on ~~#9153~~ #9159~~. A few preliminary commits to get some of the validation tools in the right place.


clokep:
Seems fine overall! 👍 Needs the base-PR updated to fix the build failure though.

clokep:
#9161 has an additional import of `parse_and_validate_server_name` that will need to be updated.

synapse/util/stringutils.py
```diff
@@ -26,6 +27,15 @@
...
 client_secret_regex = re.compile(r"^[0-9a-zA-Z\.\=\_\-]+$")
 
+# https://matrix.org/docs/spec/client_server/r0.6.1#matrix-content-mxc-uris,
+# together with https://github.com/matrix-org/matrix-doc/issues/2177 which basically
+# says "there is no grammar for media ids"
+#
+# The server_name part of this is purposely lax: use parse_and_validate_mxc for
+# additional validation.
+#
+mxc_re = re.compile("^mxc://([^/]+)/([^/#?]+)$")
```
clokep, line 37: Should this be `MXC_RE` (as `VALID_HOST_REGEX` is).

clokep (approved):
Seems good to me. 👍

synapse/util/stringutils.py
richvdh: well if we're doing that, let's call it `MXC_REGEX` ?

anyway, sure.

PR #9159, opened 2021-01-19
Give `public_baseurl` a default value

richvdh:
Context: #9154 adds a new template, and in the default template I want to use an `mxc_to_http` filter, but that isn't defined unless `public_baseurl` is set, so currently trying to start synapse without it set will explode with a mysterious exception about templates.


clokep (approved):
I think this looks reasonable!

PR #9161, opened 2021-01-19
Validate the server name for public rooms endpoint.

clokep:
The server provided to the `/publicRooms` endpoint was bogus and throwing exceptions from the HTTP client layer. We can catch this much earlier at the REST API layer though.

Fixes https://sentry.matrix.org/sentry/synapse-matrixorg/issues/119354/


richvdh (approved):
lgtm!

PR #9162, opened 2021-01-19
Add a dockerfile for running a set of Synapse worker processes

anoadragon453:
This PR adds a Dockerfile and some supporting files to the `docker/` directory. The Dockerfile's intention is to spin up a container with:

* A Synapse main process.
* Any desired worker processes, defined by a `SYNAPSE_WORKERS` environment variable supplied at runtime.
* A redis for worker communication.
* A nginx for routing traffic.
* A supervisord to start all worker processes and monitor them if any go down.

Note that **this is not currently intended to be used in production**. If you'd like to use Synapse workers with Docker, instead make use of the official image, with one worker per container. The purpose of this dockerfile is currently to allow testing Synapse in worker mode with the [Complement](https://github.com/matrix-org/complement/) test suite.

`configure_workers_and_start.py` is where most of the magic happens in this PR. It reads from environment variables (documented in the file) and creates all necessary config files for the processes. It is the entrypoint of the Dockerfile, and thus is run any time the docker container is spun up, recreating all config files in case you want to use a different set of workers. One can specify which workers they'd like to use by setting the `SYNAPSE_WORKERS` environment variable (as a comma-separated list of arbitrary worker names) or by setting it to `*` for all worker processes. We will be using the latter in CI.

Huge thanks to @MatMaul for helping get this all working :tada: This PR is paired with its equivalent on the Complement side: https://github.com/matrix-org/complement/pull/62.

Note, for the purpose of testing this PR before it's merged: You'll need to (re)build the base Synapse docker image for everything to work (`matrixdotorg/synapse:latest`). Then build the worker-based docker image on top (`matrixdotorg/synapse:workers`). 

TODO:

- ~~Take off the /dev/null's from postgres and caddy. It can hide startup errors, and they seem to mostly be quiet after startup.~~

This is probably best reviewed with all commits at once.


anoadragon453:
I've found that the `federation_sender` worker currently causes some tests `TestOutboundFederationSend` to fail, which I'm currently investigating.

anoadragon453:
> I've found that the `federation_sender` worker currently causes some tests (`TestOutboundFederationSend`) to fail. I'm currently investigating this.

Still not sure what's causing this. I propose we continue on with it disabled for now. @MatMaul will investigate further in the coming day. I've written up an issue for this as well: https://github.com/matrix-org/synapse/issues/9192

anoadragon453:
Hm, and now media traffic isn't reaching the media worker... some more debugging needed.

anoadragon453:
> Hm, and now media traffic isn't reaching the media worker... some more debugging needed.

Turns out this was only a problem when using the Complement Dockerfile, and was a result of traffic being sent to the wrong port. Things have been fixed on the [Complement side](https://github.com/matrix-org/complement/pull/62/commits/ed25ea61686169a67f6724b2b58525b6a43b38a5).

docker/Dockerfile-workers
```diff
@@ -0,0 +1,25 @@
...
+
+# Volume for user-editable config files, logs etc.
+VOLUME ["/data"]
+
+# A script to read environment variables and create the necessary
+# files to run the desired worker configuration. Will start supervisord.
+COPY ./docker/configure_workers_and_start.py /configure_workers_and_start.py
+ENTRYPOINT ["/configure_workers_and_start.py"]
+
+# TODO: Healthcheck? Which worker to ask? Can we ask supervisord?
```
anoadragon453, line 25: I don't consider this necessary for getting this PR merged, as we're mostly just concerned about running Complement tests.

docker/conf/homeserver.yaml
```diff
@@ -40,7 +40,7 @@ listeners:
         compress: false
   {% endif %}
 
-  - port: 8008
+  - port: {{ SYNAPSE_PORT or 8008 }}
```
erikjohnston, line 43: Why are we making this configurable?

docker/Dockerfile-workers
erikjohnston: Then let's kill the comment.

docker/configure_workers_and_start.py
```diff
@@ -0,0 +1,422 @@
...
+stdout_logfile=/dev/stdout
+stdout_logfile_maxbytes=0
+stderr_logfile=/dev/stderr
+stderr_logfile_maxbytes=0
+autorestart=unexpected
+exitcodes=0
+
+""" % (
+        config_path,
+    )
```
erikjohnston, line 261: Please can we split these out into separate template files, it's really hard to read when everything is inline.

docker/configure_workers_and_start.py
```diff
@@ -0,0 +1,422 @@
...
+    client_max_body_size 100M;
+    """
+    nginx_config_body = ""  # to modify below
+    nginx_config_template_end = """
+    # Send all other traffic to the main process
+    location ~* ^(\\/_matrix|\\/_synapse) {
+        proxy_pass http://localhost:8080;
+        proxy_set_header X-Forwarded-For $remote_addr;
+    }
+}
```
erikjohnston, line 284: Why nginx? We use haproxy everywhere else.

erikjohnston (commented):
I think we should restrict this to complement, I don't think its worth having a docker image that sort-of-but-not-really can be used to run workers. 

This will also need to support running multiple workers as a given type, e.g. `event_persisters`, so that we can test sharding. If we're doing this as a complement specific docker image then I'd vote we just support "monolith" or "all the workers" modes.

(I think its probably to ask whether it would have been easier to make complement support docker compose?)

docker/configure_workers_and_start.py
anoadragon453: Nginx was the solution that was reached for first, back in the [design doc](https://docs.google.com/document/d/1XcwrO_JX2erurW82czk7ivsfQUQSFMsrGPmPfKqnGnY/edit#), mostly due to its simple config format. I can replace it with HAProxy instead, though it'd be a bit of a pain.

docker/conf/homeserver.yaml
anoadragon453: So that the reverse proxy can listen on 8008 and route traffic instead. Complement will only contact the homeserver on 8008/8448 (and making it configurable on Complement's end has been rejected).

anoadragon453:
> This will also need to support running multiple workers as a given type, e.g. `event_persisters`, so that we can test sharding.

I plan to add that afterwards in a separate PR, even though I don't think it'll be too complex of a change. Just modifying `worker_extra_conf` and `shared_extra_conf` based on the worker type and count.

> If we're doing this as a complement specific docker image then I'd vote we just support "monolith" or "all the workers" modes.

Although it's added complexity, I think it's rather useful to be able to test with different worker configurations, especially while developing locally. It allows you to test sharded versus single workers. We might only run one v. the other in CI, but I think having the flexibility is nice.

It was at least useful to be able to run and fix Complement tests without the `federation_sender` enabled while it was borked.

> (I think its probably to ask whether it would have been easier to make complement support docker compose?)

I think that would be rather difficult at this point, as it's all been built around the assumption of a single docker container per homeserver. Pretty sure I asked the same at the start of all this :)

anoadragon453:
Adding sharded workers means we'll need to possibly redefine what it means to run "all the workers". Currently one can pass `SYNAPSE_WORKERS=*` to mean one instance of all possible worker types.

To define multiple instances of one worker type, I'd imagine they'd just specify `SYNAPSE_WORKERS=...federation_sender,federation_sender,...`. That then makes the `*` option in a bit of an odd spot. I don't really want to see `SYNAPSE_WORKERS=*,federation_sender`. So perhaps I should just remove `*`.

We currently set `SYNAPSE_WORKERS` here: https://github.com/matrix-org/complement/blob/9a4ff413df89d12494f725ec05076e0b4c97f990/dockerfiles/SynapseWorkers.Dockerfile#L49-L50

docker/configure_workers_and_start.py
anoadragon453: Although I will note to myself that there is quite a lot we can crib from https://github.com/matrix-org/sytest/blob/c75ff0449b5b62bdebfdb9024170988b519628d9/lib/SyTest/Homeserver/Synapse.pm#L1145-L1305.

anoadragon453:
Actually we may need to think about this a bit differently given how Complement works.

Complement will spin up a homeserver container *per-test*, which would leave us with either a lot of separate log files for each test, or some very long log files that are appended to during each test. Currently Complement makes this much nicer by by only outputting the homeserver logs to the console when a test fails.

We could also have Complement drop docker logs to disk on a test failure. These would not be separated by worker, however changes to logging mentioned earlier would prefix each log line with the worker name, which makes them separable.

> the structured log format may be useful for combined logs

This might allow standard software to separate out the logs better, but would make the stdout logs a lot harder to read plainly. At the moment I don't see much benefit to putting the logs into other software as our structured logging is still quite basic?

At the moment I'm just envisioning the logs being barfed out into buildkite's logs, hence just adding the worker name as a prefix. And potentially Complement just dropping those ones to disk for tests that failed may be enough.

erikjohnston:
(Side bar: err, spinning up a container per-test sounds like a recipe for excruciatingly slow CI?)

richvdh:
> (Side bar: err, spinning up a container per-test sounds like a recipe for excruciatingly slow CI?)

you should probably have voiced that concern at some earlier point in Complement's development... ;)

erikjohnston:
> At the moment I'm just envisioning the logs being barfed out into buildkite's logs, hence just adding the worker name as a prefix. And potentially Complement just dropping those ones to disk for tests that failed may be enough.

Given the constraints I guess that is reasonable

MatMaul:
> The alternative is to create a log file artifact for each worker, like we do with Sytest, though I find it a bit overwhelming to look through each worker's file individually when debugging. If one wanted to separate them still, they could just grep the interspersed logs to get the logs from a single worker.

I was thinking about having both, dump in stdout (thx for your work on prepending worker name @anoadragon453 ), and use a specific folder (/logs ?) that we can binmount where we dump one folder per worker, with logrotated logs there. I guess we should have a toggle for that, something like `DUMP_LOGS_TO_DISK` ? Or just do it with some sensible defaults so we don't eat too much space even if not needed (1w retention with 1d compressed logrotate ?) ?

I can have a look at some point this week if we want to go this road.

anoadragon453:
@MatMaul remember that Complement starts the docker container, so even if we do specify a volume, that may require Complement to set up a logging point. Probably best we start on the Complement side and work out a standard for dropping logs to disk.

MatMaul:
I was more thinking about end users here to be honest. Stdout with the worker name prefixed should be enough to handle Complement use case right ?

docker/configure_workers_and_start.py
anoadragon453: It looks like if we want to support more than just the [basic load-balancing techniques](https://nginx.org/en/docs/http/load_balancing.html), then we'll likely need to switch over to HAProxy.

I'll check how big of a task this will be, but at the moment I'm leaning towards just supporting round-robin LB for now with nginx, and then switch to HAProxy after the fact. Doing so shouldn't require any changes in Synapse/Complement.

anoadragon453: So nginx does actually support load-balancing by request IP, or by hashing something in the request: https://nginx.org/en/docs/http/ngx_http_upstream_module.html#hash

But actually thinking about it again I don't think we really want any load-balancing techniques other than round-robin for the Complement usecase. We certainly don't want to load-balance by IP, as all requests from our test infra would be coming from localhost.

In fact I think what we really want is consistency in load-balancing between test runs. And round-robin gives us exactly that.

anoadragon453:
> I was more thinking about end users here to be honest. Stdout with the worker name prefixed should be enough to handle Complement use case right ?

Ah right! Yes, and we already have a volume configured for the Dockerfile-workers image (`/data`) which we could just dump logs into.

I think a toggle, default on or off, would be sensible. Dumping the logs wouldn't be useful for the Complement usecase as it won't mount volumes.

anoadragon453:
@MatMaul I've added an environment variable which toggles whether worker processes write their logs to disk. It does *not* tell the main process to write its logs to disk.

We could have it do so, however the main process' log config is currently generated only once, and is then stored in the `/data` volume. We do not modify/regenerate it again. So if you then toggled the option off, all workers - who have their log config generated on each start - would not log to disk, but your main process would.

I'm not sure of a better way to do this than just making the env var for worker processes only, and leaving it up to the sysadmin to change the main process' log config manually (just as they would change the main process' homeserver.yaml config when needed).

docker/configure_workers_and_start.py
erikjohnston: The main reason to use haproxy rather than nginx is that haproxy is kinda what we "officially" support. It doesn't matter too much but its nice to be consitent

docker/README.md
```diff
@@ -208,4 +211,57 @@ healthcheck:
 ## Using jemalloc
 
 Jemalloc is embedded in the image and will be used instead of the default allocator.
-You can read about jemalloc by reading the Synapse [README](../README.md)
\ No newline at end of file
+You can read about jemalloc by reading the Synapse [README](../README.md)
+
+### Running all worker processes in a single container
```
erikjohnston, line 216:
```suggestion
### Running all worker processes in a single container for testing
```

Just to make it *super* clear

erikjohnston (commented):
I wonder if we should move some of the stuff specific to the worker docker to `conf-workers/` instead? It's a bit confusing what is being used for what etc. 

I also think this would have been a lot simpler if instead of making it all configurable we just had the majority of it as static config, but I'm also conscious that this has been open for ages now.

erikjohnston:
I'd quite like to get thoughts from other people here about whether this is good enough for now.

anoadragon453:
> I wonder if we should move some of the stuff specific to the worker docker to `conf-workers/` instead? It's a bit confusing what is being used for what etc. 

Good idea.

> I also think this would have been a lot simpler if instead of making it all configurable we just had the majority of it as static config, but I'm also conscious that this has been open for ages now.

I think generating config files is still the way to go, rather than including a separate config for each worker. The added complexity on top of that which allows for customising which workers are in use isn't too substantial I don't think.

Plus as mentioned earlier it can be handy to easily be able to specify your own setup of workers, such as when Kegan did for his project recently.

docker/README.md
```diff
@@ -53,6 +53,7 @@ The following environment variables are supported in `generate` mode:
 * `SYNAPSE_SERVER_NAME` (mandatory): the server public hostname.
 * `SYNAPSE_REPORT_STATS` (mandatory, `yes` or `no`): whether to enable
   anonymous statistics reporting.
+* `SYNAPSE_HTTP_PORT`: the port Synapse should start listen for http traffic on.
```
richvdh, line 56: should document the default here?

richvdh:
```suggestion
* `SYNAPSE_HTTP_PORT`: the port Synapse should listen on for http traffic.
```

docker/README.md
```diff
@@ -73,6 +74,8 @@ docker run -d --name synapse \
     matrixdotorg/synapse:latest
 ```
 
+assuming 8008 is the port Synapse is listening for traffic on.
```
richvdh, line 77:
```suggestion
assuming 8008 is the port Synapse is configured to listen on for http traffic.
```

docker/README.md
```diff
@@ -208,4 +211,57 @@ healthcheck:
 ## Using jemalloc
 
 Jemalloc is embedded in the image and will be used instead of the default allocator.
-You can read about jemalloc by reading the Synapse [README](../README.md)
\ No newline at end of file
+You can read about jemalloc by reading the Synapse [README](../README.md)
```
richvdh, line 214:
```suggestion
You can read about jemalloc by reading the Synapse [README](../README.md).
```

docker/README.md
```diff
@@ -208,4 +211,57 @@ healthcheck:
...
+
+### Running all worker processes in a single container for testing
```
richvdh, line 216: I feel like this should be a separate file. This file is copied to https://hub.docker.com/r/matrixdotorg/synapse, and it's not really relevant to the main `matrixdotorg/synapse` image. (indeed, it would be rather confusing for people reading it at https://hub.docker.com/r/matrixdotorg/synapse)

docker/README.md
```diff
@@ -208,4 +211,57 @@ healthcheck:
...
+handled by the container. A external reverse proxy should be placed in front the exposed
+HTTP port for the purposes of TLS termination.
+
+Note that the worker Dockerfile is based off the main `[Dockerfile](Dockerfile)`. If you
+would like to build from the current checkout, first build the main Synapse docker image
+using the instructions in [Building the image](#building-the-image), then afterwards build
+the worker image:
+
+```
+docker build -t matrixdotorg/synapse:workers -f docker/Dockerfile-workers .
```
richvdh, line 236: given it's built from a different Dockerfile, I think it should have a different image name, rather than just being a tag on the regular image:

```suggestion
docker build -t matrixdotorg/synapse-workers -f docker/Dockerfile-workers .
```

docker/README.md
```diff
@@ -208,4 +211,57 @@ healthcheck:
...
+`WORKERS_CONFIG` variable in [this script](configure_workers_and_start.py), which the
+Dockerfile makes use of to generate appropriate worker, nginx and supervisord config files.
+
+Sharding is supported for a subset of workers, in line with the [worker documentation]
+(../docs/workers.md). To run multiple instances of a given worker type, simply specify 
+the type multiple times in `SYNAPSE_WORKERS`
+(e.g `SYNAPSE_WORKERS=event_creator,event_creator...`).
+
+Otherwise, `SYNAPSE_WORKERS` can either be left empty or unset to spawn no workers (
+leaving only the main process). The container is configured to use redis-based worker
```
richvdh, line 262:
```suggestion
Otherwise, `SYNAPSE_WORKERS` can either be left empty or unset to spawn no workers 
(leaving only the main process). The container is configured to use redis-based worker
```

docker/conf/log.config
```diff
@@ -2,9 +2,34 @@ version: 1
...
+  file:
+    class: logging.handlers.TimedRotatingFileHandler
+    formatter: precise
+    filename: {{ LOG_FILE_PATH or "homeserver.log" }}
+    when: "midnight"
+    backupCount: 6  # Does not include the current log file.
+    encoding: utf8
+
+  # Default to buffering writes to log file for efficiency. This means that
+  # will be a delay for INFO/DEBUG logs to get written, but WARNING/ERROR
```
richvdh, line 21:
```suggestion
  # there will be a delay for INFO/DEBUG logs to get written, but WARNING/ERROR
```

docker/conf/log.config
```diff
@@ -17,6 +42,12 @@ loggers:
 
 root:
     level: {{ SYNAPSE_LOG_LEVEL or "INFO" }}
+
+{% if LOG_FILE_PATH %}
+    handlers: [console, buffer]
+{% else %}
+    # Write logs to the console
```
richvdh, line 49:
```suggestion
```

docker/configure_workers_and_start.py
```diff
@@ -0,0 +1,555 @@
...
+    --config-path=/conf/workers/shared.yaml \
+    --config-path=/conf/workers/{name}.yaml
+autorestart=unexpected
+priority=500
+exitcodes=0
+stdout_logfile=/dev/stdout
+stdout_logfile_maxbytes=0
+stderr_logfile=/dev/stderr
+stderr_logfile_maxbytes=0
+""".format_map(
```
richvdh, line 406: recommend sticking this sort of thing in a global constant, so that it doesn't interrupt the flow of the function.

richvdh (commented):
> I'd quite like to get thoughts from other people here about whether this is good enough for now.

Generally, I think it's fine. I've mentioned a few nits below, but I think the main thing I'd like to see changed is to move the docs out of the main docker/README, both because it's a separate docker image, but also because it helps point out the fact that this probably isn't something you want to be using unless you know what you're doing.

docker/README.md
anoadragon453: Oh indeed, I completely forgot that it ends up on Docker Hub. Good call.

anoadragon453: Although I'm struggling to come up with a good place to put it... Perhaps README-testing.md? Or maybe it should go in `docs/dev/` or CONTRIBUTING instead?

richvdh: I'd go with a `README-testing.md` in the docker dir, since it's mostly describing how the docker image works. We can always add references to it from elsewhere (eg it would be good to stick a comment in the the dockerfile itself).

anoadragon453: I've now done this.

docker/README-testing.md
```diff
@@ -0,0 +1,135 @@
...
+    matrixdotorg/synapse-workers
+```
+
+Substituting `POSTGRES*` variables for those that match a postgres host you have 
+available (usually a running postgres docker container).
+
+The `SYNAPSE_WORKERS` environment variable is a comma-separated list of workers to use
+when running the container. All possible worker names are defined by the keys of the
+`WORKERS_CONFIG` variable in [this script](configure_workers_and_start.py), which the
+Dockerfile makes use of to generate appropriate worker, nginx and supervisord config files.
```
anoadragon453, line 117: I just realised we have a `SYNAPSE_WORKER` environment variable on the base Synapse image already, and adding a different variable that's just the plural form is bound to cause confusion.

`SYNAPSE_WORKER_TYPES` sounds sensible?

anoadragon453: I went with `SYNAPSE_WORKER_TYPES`.

docker/README-testing.md
```diff
@@ -0,0 +1,136 @@
...
+
+```sh
+docker build -t matrixdotorg/synapse-workers -f docker/Dockerfile-workers .
+```
+
+This will build an image with the tag` matrixdotorg/synapse-workers`.
+
+It's worth noting at this point that this image is fully functional, and 
+can be used for testing against locally. See instructions for using the container 
+under [Running the Dockerfile-worker image](#Running the Dockerfile-worker image) below.
```
richvdh, line 69: this link is broken.

docker/README-testing.md
```diff
@@ -0,0 +1,136 @@
...
+    -p 8008:8008 \
+    -e SYNAPSE_SERVER_NAME=my.matrix.host \
+    -e SYNAPSE_REPORT_STATS=no \
+    -e POSTGRES_HOST=postgres -e POSTGRES_USER=postgres POSTGRES_PASSWORD=somesecret \
+    -e SYNAPSE_WORKER_TYPES=synchrotron,media_repository,user_dir \
+    -e SYNAPSE_WORKERS_WRITE_LOGS_TO_DISK=1 \
+    matrixdotorg/synapse-workers
+```
+
+Substituting `POSTGRES*` variables for those that match a postgres host you have
```
richvdh, line 111:
```suggestion
... substituting `POSTGRES*` variables for those that match a postgres host you have 
```

docker/README-testing.md
```diff
@@ -0,0 +1,136 @@
...
+you've set up a volume according to the [usual Synapse docker instructions](README.md).
+Then run something along the lines of:
+
+```
+docker run -d --name synapse \
+    --mount type=volume,src=synapse-data,dst=/data \
+    -p 8008:8008 \
+    -e SYNAPSE_SERVER_NAME=my.matrix.host \
+    -e SYNAPSE_REPORT_STATS=no \
+    -e POSTGRES_HOST=postgres -e POSTGRES_USER=postgres POSTGRES_PASSWORD=somesecret \
```
richvdh, line 105:
```suggestion
    -e POSTGRES_HOST=postgres \
    -e POSTGRES_USER=postgres \
    -e POSTGRES_PASSWORD=somesecret \
```

docker/README-testing.md
```diff
@@ -0,0 +1,136 @@
...
+available (usually a running postgres docker container).
+
+The `SYNAPSE_WORKER_TYPES` environment variable is a comma-separated list of workers to
+use when running the container. All possible worker names are defined by the keys of the
+`WORKERS_CONFIG` variable in [this script](configure_workers_and_start.py), which the
+Dockerfile makes use of to generate appropriate worker, nginx and supervisord config
+files.
+
+Sharding is supported for a subset of workers, in line with the [worker documentation]
+(../docs/workers.md). To run multiple instances of a given worker type, simply specify
```
richvdh, line 121: this link is also broken.

docker/configure_workers_and_start.py
```diff
@@ -20,9 +20,9 @@
 # The environment variables it reads are:
 #   * SYNAPSE_SERVER_NAME: The desired server_name of the homeserver.
 #   * SYNAPSE_REPORT_STATS: Whether to report stats.
-#   * SYNAPSE_WORKERS: A comma separated list of worker names as specified in WORKER_CONFIG
-#                      below. Leave empty for no workers, or set to '*' for all possible
-#                      workers.
+#   * SYNAPSE_WORKER_TYPES: A comma separated list of worker names as specified in WORKER_CONFIG
+#                           below. Leave empty for no workers, or set to '*' for all possible
+#                           workers.
```
richvdh, line 25:
```suggestion
#   * SYNAPSE_WORKER_TYPES: A comma separated list of worker names as specified in WORKER_CONFIG
#          below. Leave empty for no workers, or set to '*' for all possible workers.
```

richvdh (approved):
lgtm once you fix the nits below

PR #9163, opened 2021-01-19
Return a 404 if no valid thumbnail can be found

clokep:
Currently the logic for choosing a thumbnail (among the already existing thumbnails -- so not when auto-generating thumbnails is turned on) errors if no matching thumbnail is found. This leads to a 500 error returned to the client when it is reasonable to return a 404.

This should be reviewable commit-by-commit
* The first two commits are just some setup/re-organization.
* The third commit is the "real" fix.
* The fourth and fifth commits are some abstracting (if these are controversial I'm OK backing them out).

Fixes at least https://sentry.matrix.org/sentry/synapse-matrixorg/issues/127518/

### DIFF
--- a/changelog.d/9163.bugfix
+++ b/changelog.d/9163.bugfix
@@ -1,1 +1,1 @@
-Fix a longstanding bug where Synapse would return a 500 error when a thumbnail did not exist (and auto-generation of thumbnails was not enabled).
+Fix a long-standing bug where Synapse would return a 500 error when a thumbnail did not exist (and auto-generation of thumbnails was not enabled).
--- a/changelog.d/9163.bugfix
+++ b/changelog.d/9163.bugfix
@@ -1,0 +1,1 @@
+Fix a longstanding bug where Synapse would return a 500 error when a thumbnail did not exist (and auto-generation of thumbnails was not enabled).
--- a/synapse/rest/media/v1/_base.py
+++ b/synapse/rest/media/v1/_base.py
@@ -300,6 +300,7 @@ class FileInfo:
         thumbnail_height (int)
         thumbnail_method (str)
         thumbnail_type (str): Content type of thumbnail, e.g. image/png
+        thumbnail_length (int): The size of the media file, in bytes.
     """
 
     def __init__(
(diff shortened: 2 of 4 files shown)


turt2live:
How much would you hate me if I mentioned that the [spec](https://matrix.org/docs/spec/client_server/r0.6.1#get-matrix-media-r0-thumbnail-servername-mediaid) doesn't actually have a 404 code but seems to have every other error code?

clokep:
> How much would you hate me if I mentioned that the [spec](https://matrix.org/docs/spec/client_server/r0.6.1#get-matrix-media-r0-thumbnail-servername-mediaid) doesn't actually have a 404 code but seems to have every other error code?

I wouldn't hate you. 😄 What is a server meant to respond if a thumbnail does not exist then? It looks like a generic 400 could be returned. (As an aside -- I find it really confusing in the spec about whether the detailed error codes are the only allowed error codes or not.)

turt2live:
In synapse's case a 400 is probably best aligned with the spec, but a 404 is probably more appropriate tbh.

The spec is meant to strictly define the error states, but in practice 500 is allowed and so are a bunch of other "default" http states. We have also forgotten to define errors on quite a few endpoints...

clokep:
> In synapse's case a 400 is probably best aligned with the spec, but a 404 is probably more appropriate tbh.

Change them to a 400 is fairly trivial, I'd likely want to ensure the rest of the codes are sane in the media repo at the same time though... 🤷

erikjohnston:
> How much would you hate me if I mentioned that the [spec](https://matrix.org/docs/spec/client_server/r0.6.1#get-matrix-media-r0-thumbnail-servername-mediaid) doesn't actually have a 404 code but seems to have every other error code?

We return 404 if the media doesn't exist already. and it'd feel weird to change that to a 404 so :shrug:

clokep:
> Although: should we not be returning some sort of thumbnail if we have the media?

The bug was from a situation where we do have some thumbnails, but not of the *type* that is requested (e.g. cropped vs. scaled).

> Looks good. Is there an easy way of adding a test for this?

I'll take a look. 👍

clokep:
I was able to add a couple of basic tests here that failed on develop and now pass! 🎉

PR #9164, opened 2021-01-19
Try to recover from unknown media encodings.

clokep:
Sometimes we get media with unknown character encodings, e.g. Shift JIS (`x-sjis`). Instead of erroring, try to decode these as UTF-8.

Note that since we're only looking at the start of an HTML document it will usually contain a lot of metadata syntax and hopefully not too many invalid characters. In reality most of the situations I saw this happening the body we were trying to decode was purely HTML markup and whitespace.

Fixes https://sentry.matrix.org/sentry/synapse-matrixorg/issues/127535/


synapse/rest/media/v1/preview_url_resource.py
```diff
@@ -701,15 +701,17 @@ def decode_and_calc_og(
...
     except UnicodeDecodeError:
         # blindly try decoding the body as utf-8, which seems to fix
         # the charset mismatches on https://google.com
         parser = etree.HTMLParser(recover=True, encoding=request_encoding)
         tree = etree.fromstring(body.decode("utf-8", "ignore"), parser)
-        og = _calc_og(tree, media_uri)
+    except LookupError:
+        # blindly try decoding the body as utf-8.
+        parser = etree.HTMLParser(recover=True, encoding="utf-8")
+        tree = etree.fromstring(body, parser)
```
erikjohnston, line 712: Do we want to wrap this in a `try/except` so we can handle the error if this fails? Or does that happen higher up?

clokep: There isn't additional error handling. I'm not really sure what this will throw unfortunately, and the lxml docs are a bit lacking. I'll take a look though and see if there's something reasonable we can do here.

erikjohnston: I think when we're calling into a library its probably fine to just have a generic `except Exception:`?

synapse/rest/media/v1/preview_url_resource.py
```diff
@@ -717,32 +717,28 @@ def decode_and_calc_og(
...
             tree = etree.fromstring(body.decode("utf-8", "ignore"), parser)
+            return _calc_og(tree, media_uri)
         except Exception as e:
             logger.warning("Failed to parse HTML body as UTF-8: %s" % (e,))
             return {}
     except Exception as e:
         logger.warning("Failed to parse HTML body: %s" % (e,))
         return {}
 
-    return _calc_og(tree, media_uri)
```
erikjohnston: I'm not sure it makes sense to do this? As a) if it fails with `UnicodeDecodeError` we'll pointlessly retry and b) the log messages won't really be accurate.

erikjohnston (approved):

synapse/rest/media/v1/preview_url_resource.py
clokep: I'm not sure what this is referring to. This code changed in the final commit (when I added tests).

erikjohnston: I mean, why are we moving them inside the try/except blocks?

clokep: They've always been inside the try-except block. An early commit on this branch moved them out, then I realized an error can be thrown from them (and that's why they were inside of them) so I moved them back.

erikjohnston: Ooooh, right. I'm still tempted to move `return _calc_og(tree, media_uri)`  out and wrap it in a separate try/catch, but :shrug: up to you

clokep: If the `_calc_og` raises, then we need to create the new `tree` with the `body.decode("utf-8", "ignore")` bit, I'll try to make this a bit more readable, but I'm not sure...

erikjohnston: Oh, I really wasn't expecting it to fail with further decode errors? So `etree.fromstring` *and* `_calc_og` may run into decoding issues? Is that because `etree.fromstring` doesn't fully parse it and then `_calc_og` runs into it? Or?

I feel like I'm being distinctly unhelpful here, let's discuss this when you're around so we can get this over the line without another set  of back and forths

clokep: The `UnicodeDecodeError` comes out of `_calc_og` when you start iterating the tree, *not* when you call `etree.fromstring`. Maybe that's where the confusion lies?

erikjohnston: Aaaah, yeah OK. I think you can just ignore me then

PR #9165, opened 2021-01-19
Avoid trusting the content when calculating presentable names.

clokep:
Ensure that event content has the keys we expect to see before using them to calculate presentable names.

This doesn't ensure that they content is strings.

Fixes https://sentry.matrix.org/sentry/synapse-matrixorg/issues/133489/

### DIFF
--- a/changelog.d/9165.bugfix
+++ b/changelog.d/9165.bugfix
@@ -1,0 +1,1 @@
+Fix a longstanding bug where invalid data could cause errors when calculating the presentable room name for push.
--- a/changelog.d/9165.bugfix
+++ b/changelog.d/9165.bugfix
@@ -1,1 +1,1 @@
-Fix a longstanding bug where invalid data could cause errors when calculating the presentable room name for push.
+Fix a long-standing bug where invalid data could cause errors when calculating the presentable room name for push.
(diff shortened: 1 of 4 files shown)


erikjohnston (approved):
I wonder if we should be adding some unit tests for this sort of stuff?

clokep:
@erikjohnston I added some tests, not sure if they're worth a quick gander, but they are pretty much reverse-engineered from the code + have some garbage content to try to ensure there aren't exceptions. I ran these tests on develop and roughly half of them fail.

changelog.d/9165.bugfix
```diff
@@ -0,0 +1 @@
+Fix a longstanding bug where invalid data could cause errors when calculating the presentable room name for push.
```
clokep, line 1:
```suggestion
Fix a long-standing bug where invalid data could cause errors when calculating the presentable room name for push.
```

clokep:
I did also wonder while doing this if we should be asserting that the properties are strings before using them, as right now we assume they are if they exist.

erikjohnston (approved):

erikjohnston:
> I did also wonder while doing this if we should be asserting that the properties are strings before using them, as right now we assume they are if they exist.

Probably tbh, but that can be done separately.

PR #9166, opened 2021-01-19
Use the account data handler in additional places.

clokep:
Regression from #9104.

I audited the rest of the methods on the new `AccountDataHandler` and couldn't find any other callers that directly used the datastore.

See https://sentry.matrix.org/sentry/synapse-matrixorg/issues/197118/

### DIFF
--- a/changelog.d/9166.feature
+++ b/changelog.d/9166.feature
@@ -1,0 +1,1 @@
+Add experimental support for moving off receipts and account data persistence off master.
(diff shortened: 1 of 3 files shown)


clokep:
There's also https://sentry.matrix.org/sentry/synapse-matrixorg/issues/197111/, but I think that was transient during the deployment. wdyt @erikjohnston?

erikjohnston:
> There's also https://sentry.matrix.org/sentry/synapse-matrixorg/issues/197111/, but I think that was transient during the deployment. wdyt @erikjohnston?

Yeah, I messed up the deployment

erikjohnston (approved):

PR #9176, opened 2021-01-20
Speed up chain cover calculation

erikjohnston:
If we are persisting many state events, e.g. due to a join or from the background update, we ended up doing a lot of small queries for each state event. This PR batches up those queries.

Commits should be reviewable one by one.

### DIFF
--- a/synapse/storage/databases/main/events.py
+++ b/synapse/storage/databases/main/events.py
@@ -625,6 +625,7 @@ class PersistEventsStore:
             events_to_calc_chain_id_for,
             chain_map,
         )
+        chain_map.update(new_chain_tuples)
 
         db_pool.simple_insert_many_txn(
             txn,
@@ -792,8 +793,6 @@ class PersistEventsStore:
         #      new chain if the sequence number has already been allocated.
         #
 
-        new_chain_tuples = {}  # type: Dict[str, Tuple[int, int]]
-
         existing_chains = set()  # type: Set[int]
         tree = []  # type: List[Tuple[str, Optional[str]]]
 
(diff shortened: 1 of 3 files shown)


synapse/storage/databases/main/events.py
```diff
@@ -826,14 +825,28 @@ def _allocate_chain_ids(
...
+        # number of new chain IDs in one call, replacing all temporary
+        # objects with real allocated chain IDs.
+
+        unallocated_chain_ids = set()  # type: Set[object]
+        new_chain_tuples = {}  # type: Dict[str, Tuple[Any, int]]
+        for event_id, auth_event_id in tree:
+            # If we reference an auth_event_id we fetch the allocated chain ID,
+            # either from the existing `chain_map` or the newly generated
+            # `new_chain_tuples` map.
+            if auth_event_id:
```
clokep, line 844: If `auth_event_id` is not set, I *think* this will use a previous `existing_chain_id`, which I think leaks between loops?

erikjohnston: oh, woops

synapse/storage/databases/main/events.py
```diff
@@ -794,6 +752,138 @@ def _add_chain_cover_index(
...
+        existing_chains = set()  # type: Set[int]
+        tree = []  # type: List[Tuple[str, Optional[str]]]
+
+        # We need to do this in a topologically sorted order as we want to
+        # generate chain IDs/sequence numbers of an event's auth events before
+        # the event itself.
+        for event_id in sorted_topologically(
+            events_to_calc_chain_id_for, event_to_auth_chain
+        ):
+            existing_chain_id = None
```
clokep, line 805: I found where your clearing of `existing_chain_id` went! This line can be removed -- `existing_chain_id` is *always* set below before being used.

clokep (approved):
Looks like it will do the same thing, I *think* I agree that it will be faster?

erikjohnston:
> Looks like it will do the same thing, I _think_ I agree that it will be faster?

It helped a bunch on jki.re for the background update, probably won't change much when just inserting single state events at a time

synapse/storage/databases/main/events.py
erikjohnston: Yeah, though it's very confusing that it leaks between loops :)

clokep: I don't think it does actually...since it is also set, before being checked. It is scoped to the inner loop only?

erikjohnston: Oh, argh, yes, I see what you're saying. This line can be removed

PR #9177, opened 2021-01-20
Tighten the restrictions on `idp_id`

richvdh:
MSC2858 has moved the goalposts here: see https://github.com/matrix-org/matrix-doc/pull/2858/commits/c7063a4eeca073ecc9cac0368381e8b7338d98fd#diff-23a449bd815bfd357851e38da761fd5d5d2aef4fc26f50cbefe581e64ecafaa7L49, noting that that link should be to https://github.com/matrix-org/matrix-doc/pull/2758


clokep:
Looks like the max length also changed from 128 to 255?

clokep:
For reference, the `idp_id` now uses https://github.com/matrix-org/matrix-doc/blob/rav/proposals/textual_identifier_grammar/proposals/2758-textual-id-grammar.md#proposal

clokep (approved):

clokep:
> Looks like the max length also changed from 128 to 255?

We decided to ignore this, we can open it up later if we want.

PR #9180, opened 2021-01-20
Avoid a warning when using `twine check`.

clokep:
This avoids a warning when using `twine check dist/*` (which we should be running in CI to avoid things like 4c37d2acd58f9cbcdc262d25a75f8b38c50ed7f9).


clokep:
I don't see a downside of doing this, but let me know if you think there is?

callahad (approved):
lgtm

PR #9181, opened 2021-01-20
Use execute_batch instead of executemany in places 

erikjohnston:
`execute_batch` does fewer round trips in postgres than `executemany`, but does not give a correct `txn.rowcount` result after.

### DIFF
--- a/synapse/storage/database.py
+++ b/synapse/storage/database.py
@@ -267,8 +267,7 @@ class LoggingTransaction:
 
             self._do_execute(lambda *x: execute_batch(self.txn, *x), sql, args)
         else:
-            for val in args:
-                self.execute(sql, val)
+            self.executemany(sql, args)
 
     def execute_values(self, sql: str, *args: Any) -> List[Tuple]:
         """Corresponds to psycopg2.extras.execute_values. Only available when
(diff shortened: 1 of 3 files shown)


clokep:
I assume this replaces each call where we don't care about the result?

synapse/storage/database.py
```diff
@@ -267,8 +267,7 @@ def execute_batch(self, sql: str, args: Iterable[Iterable[Any]]) -> None:
 
             self._do_execute(lambda *x: execute_batch(self.txn, *x), sql, args)
         else:
-            for val in args:
-                self.execute(sql, val)
+            self.executemany(sql, args)
```
clokep, line 270: Is it expected that this has any performance difference?

clokep (approved):

erikjohnston:
> I assume this replaces each call where we don't care about the result?

There are more I think, I just did the obvious ones that looked to be on hot paths.

synapse/storage/database.py
erikjohnston: The docs are unclear, so possibly? Either way probably clearer to shove it down a layer :shrug:

clokep: Hopefully the underlying code is optimized, yeah. 👍

PR #9184, opened 2021-01-20
Add a check for duplicate IdP ids

richvdh:


synapse/config/oidc_config.py
```diff
@@ -43,6 +44,17 @@ def read_config(self, config, **kwargs):
...
+        # check we don't have any duplicate idp_ids
+        # XXX: this won't detect clashes with other IdP providers using other SSO
+        # mechanisms (such as SAML or CAS); that will be detected when we set up the
+        # listeners but by then synapse will have forked, so it's not ideal.
+        c = Counter([i.idp_id for i in self.oidc_providers])
+        for idp_id, count in c.items():
+            if count > 1:
+                raise ConfigError(
+                    "Multiple OIDC providers have the idp_id %r." % idp_id
+                )
```
clokep, line 56: Should this give a path of `oidc_providers` at least? Not sure if that's even reasonable.

richvdh: maybe, but it could be a clash between an entry in `oidc_providers` and `oidc_config`

clokep: Ah, right, the backwards compatibility bit. I think this is OK then.

clokep (approved):

synapse/config/oidc_config.py
```diff
@@ -43,6 +44,17 @@ def read_config(self, config, **kwargs):
         except DependencyException as e:
             raise ConfigError(e.message) from e
 
+        # check we don't have any duplicate idp_ids
+        # XXX: this won't detect clashes with other IdP providers using other SSO
+        # mechanisms (such as SAML or CAS); that will be detected when we set up the
+        # listeners but by then synapse will have forked, so it's not ideal.
```
clokep, line 50: We could prefix them with an implementation specific bit if we wanted to avoid people having to care about this. 🤷 E.g. if you use `google` we would make that `oidc-google` or something of that nature.

richvdh: true. If we're going to change it, we better do it before 1.26 ships and people start using the feature and getting entries in their `user_external_ids` table.

counterpoint: *not* including the SSO mechanism here means that we leave the path open for SSO providers to switch from, say, SAML to OIDC in the future without needing a db update (though that also requires the unique remote user id to remain constant, which seems unlikely on reflection)

richvdh: on reflection: I think this change makes sense. I'm going to do it in a branch targetting release-v1.26.

richvdh: ok that's done in #9189. I don't think it changes anything in this PR apart from making this comment redundant so I'm going to remove it.

clokep: Yeah, that seems fine. 👍

PR #9188, opened 2021-01-21
Use execute_batch in more places

erikjohnston:
c.f. #9181

### DIFF
--- a/synapse/storage/databases/main/events_bg_updates.py
+++ b/synapse/storage/databases/main/events_bg_updates.py
@@ -139,8 +139,6 @@ class EventsBackgroundUpdatesStore(SQLBaseStore):
         max_stream_id = progress["max_stream_id_exclusive"]
         rows_inserted = progress.get("rows_inserted", 0)
 
-        INSERT_CLUMP_SIZE = 1000
-
         def reindex_txn(txn):
             sql = (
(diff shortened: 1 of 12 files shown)


synapse/storage/databases/main/events_bg_updates.py
```diff
@@ -178,9 +176,7 @@ def reindex_txn(txn):
 
             sql = "UPDATE events SET sender = ?, contains_url = ? WHERE event_id = ?"
 
-            for index in range(0, len(update_rows), INSERT_CLUMP_SIZE):
-                clump = update_rows[index : index + INSERT_CLUMP_SIZE]
-                txn.executemany(sql, clump)
```
clokep, line 183: After this change, this will essentially end up as `txn.executemany(sql, update_rows)` for SQLite. Is that OK or should we do this clumping inside of our `execute_batch` method for SQLite?

erikjohnston: I think the clumping is a bit spurious, I believe `executemany` is fine handling large lists.

clokep: The docs don't say anything about trying to limit the length of the iterator used, so... 🤷

clokep (approved):

PR #9189, opened 2021-01-21
Prefix idp_id with "oidc-"

richvdh:
... to avoid clashes with other SSO mechanisms

cf https://github.com/matrix-org/synapse/pull/9184#discussion_r561228817

### DIFF
--- a/changelog.d/9189.misc
+++ b/changelog.d/9189.misc
@@ -1,0 +1,1 @@
+Add an `oidc-` prefix to any `idp_id`s which are given in the `oidc_providers` configuration.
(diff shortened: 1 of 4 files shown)


clokep (approved):
I think this looks fine, an alternative that came to mind was to store the SSO type (OIDC/SAML/CAS) as a separate column from the ID. I don't see a strong reason to do it that way though.

richvdh:
> an alternative that came to mind was to store the SSO type (OIDC/SAML/CAS) as a separate column from the ID.

note that the `idp_id` is also used as an identifier for msc2858 and the idp-picker, where it also needs to be unique, so I don't think that approach would help much.

PR #9190, opened 2021-01-21
Increase perf of handling concurrent use of StreamIDGenerators.

erikjohnston:
We have seen a failure mode here where if there are many in flight unfinished IDs then marking an ID as finished takes a lot of CPU (as calling `deque.remove` iterates over the list)

### DIFF
--- a/changelog.d/9190.misc
+++ b/changelog.d/9190.misc
@@ -1,0 +1,1 @@
+Increase perf of handling concurrent use of `StreamIDGenerators`.
--- a/changelog.d/9190.misc
+++ b/changelog.d/9190.misc
@@ -1,1 +1,1 @@
-Increase perf of handling concurrent use of `StreamIDGenerators`.
+Improve performance of concurrent use of `StreamIDGenerators`.
--- a/synapse/storage/util/id_generators.py
+++ b/synapse/storage/util/id_generators.py
@@ -102,7 +102,10 @@ class StreamIdGenerator:
             )
 
         # We use this as an ordered set, as we want to efficiently append items,
-        # remove items and get the first item.
+        # remove items and get the first item. Since we insert IDs in order, the
+        # insertion ordering will ensure its in the correct ordering.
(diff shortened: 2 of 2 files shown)


synapse/storage/util/id_generators.py
```diff
@@ -101,7 +100,10 @@ def __init__(self, db_conn, table, column, extra_tables=[], step=1):
             self._current = (max if step > 0 else min)(
                 self._current, _load_current_id(db_conn, table, column, step)
             )
-        self._unfinished_ids = deque()  # type: Deque[int]
+
+        # We use this as an ordered set, as we want to efficiently append items,
+        # remove items and get the first item.
```
richvdh, line 105: can you explain what the values of the dict are? same as the key? `0`? the former I think?

synapse/storage/util/id_generators.py
```diff
@@ -101,7 +100,10 @@ def __init__(self, db_conn, table, column, extra_tables=[], step=1):
...
+        self._unfinished_ids = OrderedDict()  # type: OrderedDict[int, int]
```
richvdh, line 106: `OrderedDict` is ordered by insertion order rather than key, iirc. Is that what you intend here? (possibly it amounts to the same thing?)

changelog.d/9190.misc
```diff
@@ -0,0 +1 @@
+Increase perf of handling concurrent use of `StreamIDGenerators`.
```
richvdh, line 1:
```suggestion
Improve performance of concurrent use of `StreamIDGenerators`.
```

richvdh (commented):
generally lgtm

synapse/storage/util/id_generators.py
erikjohnston: Yeah, they amount to the same thing (and is the same situation as with deque). Will expand the comment

richvdh (approved):

PR #9038, opened 2021-01-07
Run the linters on a consistent list of files

richvdh:
We were running some linters on some files and some on others. Extract a common setting and use it everywhere.

As a side-effect, we limit the files we run `black` on, so if you have any random directories in your working copy, it won't go and mess them up.

PR #9191, opened 2021-01-21
Align the directories linted in CI with the defaults in scripts-dev/lint.sh

anoadragon453:
The lists of source directories to lint between `tox.ini` and `lint.sh` became out of sync. This PR tightens them up and adds some comments reminding any future readers to keep the list in sync.

There's certainly something to be said about maintaining two lists. Perhaps in the future `tox.ini` should just call `lint.sh` instead. For now, these directories change so rarely that I don't think it's a big deal.

### DIFF
--- a/changelog.d/9191.misc
+++ b/changelog.d/9191.misc
@@ -1,0 +1,1 @@
+Add some missing source directories to the automatic linting script.
--- a/scripts-dev/lint.sh
+++ b/scripts-dev/lint.sh
@@ -80,7 +80,7 @@ else
   # then lint everything!
   if [[ -z ${files+x} ]]; then
     # Lint all source code files and directories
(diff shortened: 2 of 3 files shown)


clokep (approved):
This seems OK to me. Also see #9038 where this list was normalized a bit.

clokep:
After this change the lint script seems to keep wanting to modify `synapse/stubs/txredisapi.pyi`? It just wants to remove a blank line though. 😕

anoadragon453:
@clokep  I've been noticing this for a while with another `.pyi` file as well. Turns out it's a bug in `isort`, and is fixed in the latest version! As a result I've created a PR to update `isort`: https://github.com/matrix-org/synapse/pull/9222

PR #9193, opened 2021-01-21
Fix receipts or account data not being sent down sync

erikjohnston:
Introduced in #9104 

This wasn't picked up by the tests as this is all fine the first time you run Synapse (after upgrading), but then when you restart the wrong value is pulled from `stream_positions`.

### DIFF
--- a/synapse/storage/util/sequence.py
+++ b/synapse/storage/util/sequence.py
@@ -45,6 +45,21 @@ and run the following SQL:
 See docs/postgres.md for more information.
 """
 
+_INCONSISTENT_STREAM_ERROR = """
+Postgres sequence '%(seq)s' is inconsistent with associated stream position
+of '%(stream_name)s' in the 'stream_positions' table.
+
+This is likely a programming error and should be reported at
+https://github.com/matrix-org/synapse.
+
+A temporary workaround to fix this error is to shut down Synapse (including
+any and all workers) and run the following SQL:
+
+    DELETE FROM stream_positions WHERE stream_name = '%(stream_name)s';
+
+This will need to be done every time the server is restarted.
+"""
+
 
 class SequenceGenerator(metaclass=abc.ABCMeta):
     """A class which generates a unique sequence of integers"""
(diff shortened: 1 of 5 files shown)


synapse/storage/util/sequence.py
```diff
@@ -60,14 +75,20 @@ def check_consistency(
...
-        This is to handle various cases where the sequence value can get out
-        of sync with the table, e.g. if Synapse gets rolled back to a previous
+        This is to handle various cases where the sequence value can get out of
+        sync with the table, e.g. if Synapse gets rolled back to a previous
         version and the rolled forwards again.
+
+        If a stream name is given then will check that any value in the
+        `stream_positions` table is less than or equal to the current sequence
+        value. If it isn't then it's likely that streams have been crossed
+        somewhere (e.g. two ID generators have the same stream name).
```
clokep, line 91: Something seems off with the first sentence -- "then will check" is missing a word maybe?

erikjohnston: Woops, I think that is "If a stream name is given then **this** will check...."

synapse/storage/util/sequence.py
```diff
@@ -118,6 +143,18 @@ def check_consistency(
...
+        # If we have an associated stream check the stream_positions table.
+        max_in_stream_positions = None
+        if stream_name:
+            txn.execute(
+                "SELECT MAX(stream_id) FROM stream_positions WHERE stream_name",
+                (stream_name,),
+            )
+            rows = txn.fetchall()
+            if rows and row[0][0] is not None:
+                max_in_stream_positions = row[0][0]
```
clokep, line 156: Can this be `fetchone` instead?

erikjohnston: Done

synapse/storage/util/sequence.py
```diff
@@ -118,6 +143,18 @@ def check_consistency(
...
+        # If we have an associated stream check the stream_positions table.
+        max_in_stream_positions = None
+        if stream_name:
+            txn.execute(
+                "SELECT MAX(stream_id) FROM stream_positions WHERE stream_name = ?",
+                (stream_name,),
+            )
+            row = txn.fetchone()
+            if row and row[0] is not None:
+                max_in_stream_positions = row[0][0]
```
clokep, line 156: Should this now be `row[0]`? (Also the check for `row[0] is not None` isn't really necessary since we'll just set the max back to `None`, but 🤷

erikjohnston: sigh

clokep (approved):
LGTM.

PR #9198, opened 2021-01-21
Precompute joined hosts and store in Redis

erikjohnston:
This is a fairly noddy implementation, but hopefully makes sense as a first cut

### DIFF
--- a/synapse/federation/sender/__init__.py
+++ b/synapse/federation/sender/__init__.py
@@ -199,16 +199,22 @@ class FederationSender:
                     if not event.internal_metadata.should_proactively_send():
                         return
 
-                    # We check the external cache for the destinations, which is
-                    # stored per state group.
-                    destinations = None
-                    sg = await self._external_cache.get(
-                        "event_to_prev_state_group", event.event_id
-                    )
-                    if sg:
-                        destinations = await self._external_cache.get(
-                            "get_joined_hosts", str(sg)
+                    destinations = None  # type: Optional[Set[str]]
+                    if not event.prev_event_ids():
+                        # If there are no prev event IDs then the state is empty
+                        # and so no remote servers in the room
+                        destinations = set()
+                    else:
+                        # We check the external cache for the destinations, which is
+                        # stored per state group.
+
+                        sg = await self._external_cache.get(
+                            "event_to_prev_state_group", event.event_id
                         )
+                        if sg:
+                            destinations = await self._external_cache.get(
+                                "get_joined_hosts", str(sg)
+                            )
 
                     if destinations is None:
                         try:
--- a/synapse/federation/sender/__init__.py
+++ b/synapse/federation/sender/__init__.py
@@ -71,6 +71,7 @@ class FederationSender:
         self.server_name = hs.hostname
 
         self.store = hs.get_datastore()
+        self.state_store = hs.get_datastores().state
         self.state = hs.get_state_handler()
 
         self.clock = hs.get_clock()
@@ -142,6 +143,8 @@ class FederationSender:
             self._wake_destinations_needing_catchup,
         )
 
+        self._external_cache = hs.get_external_cache()
+
     def _get_per_destination_queue(self, destination: str) -> PerDestinationQueue:
         """Get or create a PerDestinationQueue for the given destination
 
@@ -197,22 +200,34 @@ class FederationSender:
                     if not event.internal_metadata.should_proactively_send():
                         return
 
-                    try:
-                        # Get the state from before the event.
-                        # We need to make sure that this is the state from before
-                        # the event and not from after it.
-                        # Otherwise if the last member on a server in a room is
-                        # banned then it won't receive the event because it won't
-                        # be in the room after the ban.
-                        destinations = await self.state.get_hosts_in_room_at_events(
-                            event.room_id, event_ids=event.prev_event_ids()
+                    # We check the external cache for the destinations, which is
+                    # stored per state group.
+                    destinations = None
+                    sg = await self._external_cache.get_cache(
+                        "event_to_prev_state_group", event.event_id
+                    )
+                    if sg:
+                        destinations = await self._external_cache.get_cache(
+                            "get_joined_hosts", str(sg)
                         )
-                    except Exception:
-                        logger.exception(
-                            "Failed to calculate hosts in room for event: %s",
-                            event.event_id,
-                        )
-                        return
+
+                    if destinations is None:
+                        try:
+                            # Get the state from before the event.
+                            # We need to make sure that this is the state from before
+                            # the event and not from after it.
+                            # Otherwise if the last member on a server in a room is
+                            # banned then it won't receive the event because it won't
+                            # be in the room after the ban.
+                            destinations = await self.state.get_hosts_in_room_at_events(
+                                event.room_id, event_ids=event.prev_event_ids()
+                            )
+                        except Exception:
+                            logger.exception(
+                                "Failed to calculate hosts in room for event: %s",
+                                event.event_id,
+                            )
+                            return
 
                     destinations = {
                         d
--- a/synapse/federation/sender/__init__.py
+++ b/synapse/federation/sender/__init__.py
@@ -71,7 +71,6 @@ class FederationSender:
         self.server_name = hs.hostname
 
         self.store = hs.get_datastore()
-        self.state_store = hs.get_datastores().state
         self.state = hs.get_state_handler()
 
         self.clock = hs.get_clock()
--- a/synapse/federation/sender/__init__.py
+++ b/synapse/federation/sender/__init__.py
@@ -202,11 +202,11 @@ class FederationSender:
                     # We check the external cache for the destinations, which is
                     # stored per state group.
                     destinations = None
-                    sg = await self._external_cache.get_cache(
+                    sg = await self._external_cache.get(
                         "event_to_prev_state_group", event.event_id
                     )
                     if sg:
-                        destinations = await self._external_cache.get_cache(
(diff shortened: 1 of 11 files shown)


synapse/federation/sender/__init__.py
```diff
@@ -197,22 +200,34 @@ async def handle_event(event: EventBase) -> None:
...
-                        # be in the room after the ban.
-                        destinations = await self.state.get_hosts_in_room_at_events(
-                            event.room_id, event_ids=event.prev_event_ids()
-                        )
-                    except Exception:
-                        logger.exception(
-                            "Failed to calculate hosts in room for event: %s",
-                            event.event_id,
+                    # We check the external cache for the destinations, which is
+                    # stored per state group.
```
clokep, line 204: Why not have a `self._use_external_cache` check here like when we set the entries?

synapse/handlers/message.py
```diff
@@ -432,6 +432,9 @@ def __init__(self, hs: "HomeServer"):
 
         self._ephemeral_events_enabled = hs.config.enable_ephemeral_messages
 
+        self._external_cache = hs.get_external_cache()
+        self._use_external_cache = hs.config.redis.redis_enabled
```
clokep, line 436: I wonder if this should be `redis enabled and more than one federation sender`?

synapse/replication/tcp/external_cache.py
```diff
@@ -0,0 +1,112 @@
...
+                host=hs.config.redis_host,
+                port=hs.config.redis_port,
+                password=hs.config.redis.redis_password,
+                reconnect=True,
+            )
+
+    def _get_redis_key(self, cache_name: str, key: str) -> str:
+        return "cache_v1:%s:%s" % (cache_name, key)
+
+    async def set_cache(
```
clokep, line 71: Our other caches seem to name these methods as simply `set` and `get`

I also wonder if instead of `expire_seconds` we should do `expiry_ms` as the `ExpiringCache` does.

synapse/replication/tcp/external_cache.py
```diff
@@ -0,0 +1,112 @@
...
+        """
+
+        if self._redis_connection is None:
+            return None
+
+        result = await self._redis_connection.get(self._get_redis_key(cache_name, key))
+
+        logger.debug("Got cache result %s %s: %r", cache_name, key, result)
+
+        get_counter.labels(cache_name, result is not None).inc()
```
clokep, line 103: There's the whole `register_cache` thing that does some metrics stuff, not sure that's appropriate to use here.

In general though, I think we would want to be able to calculate a cache hit rate, which I think we need more info than this?

synapse/federation/sender/__init__.py
clokep: Oh I guess this is because the external cache just quickly returns `None` if not being used?

synapse/handlers/message.py
clokep: Or at the very least if the federation sending is done by the main process it probably doesn't make sense to use this?

synapse/replication/tcp/external_cache.py
erikjohnston: Yeah, I'm sort of in two minds about that as these work differently. You should be able to work out the cache hit ratio via `get_counter{result="true"}/sum(get_counter)` I think?

synapse/handlers/message.py
erikjohnston: Maybe, though it is worth noting that it's a) probably cheaper to calculate it on event creator, as you'll already have the state in the cache, and b) means that you can run multiple event creators to split up the work without having to run multiple federation senders.

synapse/federation/sender/__init__.py
erikjohnston: Yeah, exactly, it's kinda like with logging where you sometimes do `if logger.isEnabledFor(...)` to protect from extra work. It's probably clearer to change the `set` path to always have an `ExternalCache` instance and call a `self._external_cache.is_enabled()`, or something

synapse/federation/sender/__init__.py
```diff
@@ -71,6 +71,7 @@ def __init__(self, hs: "synapse.server.HomeServer"):
         self.server_name = hs.hostname
 
         self.store = hs.get_datastore()
+        self.state_store = hs.get_datastores().state
```
clokep, line 74: I think this is unused?

synapse/federation/sender/__init__.py
clokep: 🤷 Either way seems OK, I think having a single specific spot for the enabled logic to live makes sense.

synapse/handlers/message.py
clokep: Ah, right that makes sense. I guess it is a reasonable assumption that anyone running workers would be running at least one of an event creator or federation sender?

synapse/replication/tcp/external_cache.py
clokep: Oh, you should be able to! I think I was more of just surprised this was done differently.

synapse/handlers/message.py
erikjohnston: Yeah, and its really not the end of the world if they aren't.

synapse/replication/tcp/external_cache.py
clokep: Annoyingly the `ExpiringCache` uses `expiry_ms` (with a y). I don't know if matters much.

Also do you have thoughts on `set`/`get` vs. `set_cache`/`get_cache`?

clokep:
Do we need any tests for this or would they just be mocking out the world?

synapse/replication/tcp/external_cache.py
erikjohnston: I KNEW THERE WAS SOMETHING I FORGOT TO DO

clokep:
> Do we need any tests for this or would they just be mocking out the world?

We talked about this in #synapse-dev:matrix.org and decided that we wouldn't really be testing anything with unit tests. This is testing via worker mode of sytest though! 👍

clokep (changes requested):
Per the changes to `get` / `set` for the API.

synapse/replication/tcp/external_cache.py
```diff
@@ -0,0 +1,118 @@
...
+        if hs.config.redis.redis_enabled:
+            from synapse.replication.tcp.redis import lazyConnection
+
+            logger.info(
+                "Connecting to redis (host=%r port=%r) for external cache",
+                hs.config.redis_host,
+                hs.config.redis_port,
+            )
+
+            # First create the connection for sending commands.
```
richvdh, line 59: first?

synapse/replication/tcp/external_cache.py
```diff
@@ -0,0 +1,118 @@
...
+        result = await self._redis_connection.get(self._get_redis_key(cache_name, key))
+
+        logger.debug("Got cache result %s %s: %r", cache_name, key, result)
+
+        get_counter.labels(cache_name, result is not None).inc()
+
+        if not result:
+            return None
+
+        # For some reason the integers get magically converted back to integers
```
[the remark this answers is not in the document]
erikjohnston: Oh wow, that's quite special. Yeah, I agree we should turn it off.

synapse/replication/tcp/external_cache.py
```diff
@@ -0,0 +1,118 @@
...
+            from synapse.replication.tcp.redis import lazyConnection
+
+            logger.info(
+                "Connecting to redis (host=%r port=%r) for external cache",
+                hs.config.redis_host,
+                hs.config.redis_port,
+            )
+
+            # First create the connection for sending commands.
+            self._redis_connection = lazyConnection(
```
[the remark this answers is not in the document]
erikjohnston: Yeah, that's a good point. I'm not really sure how a) cheap the connections are and b) whether we want to segregate the cache stuff from the replication pub/sub. Since the commands get pipelined the connections shouldn't get congested, I guess?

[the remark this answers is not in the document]
erikjohnston: I've changed it so that we share the outbound connection.

erikjohnston:
I've added a few more commits with the aim of getting a better cache hit ratio, in a SyTest run it looks like we're now at 100% cache hit ratio.

synapse/replication/tcp/external_cache.py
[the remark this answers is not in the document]
erikjohnston: I'll do it in a separate PR though I think

synapse/handlers/message.py
```diff
@@ -978,6 +982,44 @@ async def handle_new_client_event(
             await self.store.remove_push_actions_from_staging(event.event_id)
             raise
 
+    async def _cache_joined_hosts_for_event(self, event: EventBase):
+        """Precalculate the joined hosts at the event, when using Redis, so that
+        external federation senders don't have to recalculate it themselves.
+        """
```
clokep, line 988:
```suggestion
    async def _cache_joined_hosts_for_event(self, event: EventBase) -> None:
        """Precalculate the joined hosts at the event, when using Redis, so that
        external federation senders don't have to recalculate it themselves.
        """
```

I find this docstring a bit confusing too, but couldn't figure out a good way to reword it.

Also should this be public (no `_` prefix) since it is called from another handler?

clokep (approved):
LGTM!

synapse/handlers/message.py
erikjohnston: Yes it should be public

erikjohnston:
Thanks @clokep !

PR #9199, opened 2021-01-21
Do not require the CAS service URL setting (use public_baseurl instead).

### LINKED ISSUE #7198: CAS should use the public_baseurl instead of a separate cas_service_url
When using CAS you currently need to define a `cas_service_url` in the homeserver configuration. This ends up being a duplicate of the `public_baseurl` configuration and should be fairly easy to get rid of.

Reasons for doing this:
* Not having duplicated config is good. 👍 
* This will match what SAML requires.

We'll need to deprecate the old option before just outright removing it.

clokep:
Since `public_baseurl` is always defined as of #9159, it becomes easy to automatically use this for CAS if `service_url` is not defined.

By default if `service_url` is defined that will be used instead.

Fixes #7198

### DIFF
--- a/synapse/config/cas.py
+++ b/synapse/config/cas.py
@@ -30,7 +30,7 @@ class CasConfig(Config):
 
         if self.cas_enabled:
             self.cas_server_url = cas_config["server_url"]
-            self.cas_service_url = cas_config["service_url"]
+            self.cas_service_url = cas_config.get("service_url") or self.public_baseurl
             self.cas_displayname_attribute = cas_config.get("displayname_attribute")
             self.cas_required_attributes = cas_config.get("required_attributes") or {}
         else:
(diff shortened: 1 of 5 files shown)


synapse/config/cas.py
```diff
@@ -30,7 +30,13 @@ def read_config(self, config, **kwargs):
 
         if self.cas_enabled:
             self.cas_server_url = cas_config["server_url"]
-            self.cas_service_url = cas_config["service_url"]
+            public_base_url = cas_config.get("service_url") or self.public_baseurl
+            if public_base_url[-1] != "/":
+                public_base_url += "/"
```
erikjohnston, line 35: This will change `self.public_baseurl`, which may end up confusing people?

erikjohnston (approved):

synapse/config/cas.py
clokep: It should not since strings are immutable: `public_base_url` doesn't end up as a reference to `self.public_baseurl` unless there's additional magic here that I don't know?

clokep: Also in reality this code won't run if `self.public_baseurl` is used:

https://github.com/matrix-org/synapse/blob/0cd2938bc854d947ae8102ded688a626c9fac5b5/synapse/config/server.py#L167-L168

PR #8394, opened 2020-09-24
Hotfix: disable autoescape by default when rendering Jinja2 templates

anoadragon453:
#8037 changed the default `autoescape` option when rendering Jinja2 templates from `False` to `True`. This caused some bugs, noticeably around redirect URLs being escaped in SAML2 auth confirmation templates, causing those URLs to break for users.

This change returns the previous behaviour as it stood. We may want to look at each template individually and see whether autoescaping is a good idea at some point, but for now lets just fix the breakage.

PR #8037, opened 2020-08-06
Use the default templates when a custom template file cannot be found

anoadragon453:
Fixes https://github.com/matrix-org/synapse/issues/6583

This PR:

* Adds a new function, `read_templates` to `EmailConfig` that can take a custom template directory and some template filenames and will load them as `jinja2.Template` objects. If a filename isn't found in the given custom template directory, then the default synapse template directory is used instead.
  - This is currently located in `EmailConfig` due to the rest of the templating code being there. I'd like to at some point I'd like to move this method to a more "utils" area, and to factor out the different template config options into config files that are relevant to each template's purpose (similar to how sso's templates have been done), but I think doing so in this PR would make it too large.
* Switches up the rest of the codebase to just use this function.

I also took out some safe imports of `jinja2` and `bleach`, as I believe they were once optional imports, but are now just included in the required dependencies..

Best reviewed commit-by-commit.

PR #9200, opened 2021-01-21
Clean-up the template loading code

clokep:
This attempts to clean-up some of the template loading code a bit:

* Enables auto-escape by default for HTML files (this was originally done in #8037, but partially disabled in #8394).
* Adds a new `read_template` method for reading a single template.
* Some logic clean-up.

A few things to look out for here:
* Does the SSO redirect issue still exist from #8394? I don't know how to reproduce this, maybe @anoadragon453 remembers.
* Do we need to call this out in the upgrade notes?

### DIFF
--- a/.buildkite/scripts/test_old_deps.sh
+++ b/.buildkite/scripts/test_old_deps.sh
@@ -10,4 +10,7 @@ apt-get install -y python3.5 python3.5-dev python3-pip libxml2-dev libxslt-dev x
 
 export LANG="C.UTF-8"
 
+# Prevent virtualenv from auto-updating pip to an incompatible version
+export VIRTUALENV_NO_DOWNLOAD=1
+
 exec tox -e py35-old,combine
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,20 @@
+Synapse 1.26.0rc2 (2021-01-25)
+==============================
+
+Bugfixes
+--------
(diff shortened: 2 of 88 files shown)


clokep:
Following-up on a conversation from #synapse-dev:matrix.org from earlier today -- both of the redirect URLs for SSO were already being manually escaped, so turning on autoescape shouldn't change anything here. It is also possible that #9128 would help with whatever problem we had been seeing.

I'll double check with some of the SSO setups though to ensure that nothing seems amiss.

erikjohnston (commented):
I think the code looks fine, I don't know the answers to the questions posed in the PR description though

clokep:
> I'll double check with some of the SSO setups though to ensure that nothing seems amiss.

I did some testing here and the redirect flows for both login and UIAuth were still working fine with this branch.

clokep:
> I think the code looks fine, I don't know the answers to the questions posed in the PR description though

I think the first question is answered, for the second one, I suspect we should probably mention something about it.

dkasak:
We should definitely mention something so that people know what's up if it breaks their stuff, IMO.

clokep:
I added a note about this to the upgrade docs.

erikjohnston (approved):

erikjohnston:
@clokep looks like you need to merge in develop?

PR #9203, opened 2021-01-22
Add some configuration settings to make profile data more private

AndrewFerr:
This adds two off-by-default settings that each restrict access to profile data (displaynames and avatars):
- **include_profile_data_on_invite**: If false, prevents invite events from including the invitee's profile data. Otherwise, all it takes for someone to access any user's profile data is to know/guess their mxid and invite them to a room.
- **allow_profile_lookup_over_federation**: If false, causes any requests for profile data over federation to respond with `M_FORBIDDEN`.

I've had these tweaks on my own server for quite some time now, and felt it would be helpful to share them.

### Pull Request Checklist

### DIFF
--- a/synapse/handlers/message.py
+++ b/synapse/handlers/message.py
@@ -381,6 +381,12 @@ class EventCreationHandler:
 
         self.room_invite_state_types = self.hs.config.room_invite_state_types
 
+        self.membership_types_to_include_profile_data_in = (
+            {Membership.JOIN, Membership.INVITE}
+            if self.hs.config.include_profile_data_on_invite
+            else {Membership.JOIN}
+        )
+
         self.send_event = ReplicationSendEventRestServlet.make_client(hs)
 
         # This is only used to get at ratelimit function, and maybe_kick_guest_users
@@ -494,7 +500,7 @@ class EventCreationHandler:
             membership = builder.content.get("membership", None)
             target = UserID.from_string(builder.state_key)
 
-            if membership in {Membership.JOIN, Membership.INVITE}:
+            if membership in self.membership_types_to_include_profile_data_in:
                 # If event doesn't include a display name, add one.
                 profile = self.profile_handler
                 content = builder.content
--- a/synapse/handlers/profile.py
+++ b/synapse/handlers/profile.py
@@ -309,6 +309,16 @@ class ProfileHandler(BaseHandler):
         await self._update_join_states(requester, target_user)
 
     async def on_profile_query(self, args: JsonDict) -> JsonDict:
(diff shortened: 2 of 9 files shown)


clokep:
@AndrewFerr Can you regenerate the sample config for the updates, you should be able to run `./scripts-dev/generate_sample_config` to do it. Thanks!

synapse/handlers/message.py
```diff
@@ -381,6 +381,12 @@ def __init__(self, hs: "HomeServer"):
 
         self.room_invite_state_types = self.hs.config.room_invite_state_types
 
+        self.event_types_to_include_profile_data_in = (
```
erikjohnston, line 384: This should probably be called `membership_types_...`?

synapse/handlers/profile.py
```diff
@@ -309,6 +309,16 @@ async def set_avatar_url(
         await self._update_join_states(requester, target_user)
 
     async def on_profile_query(self, args: JsonDict) -> JsonDict:
+        if (
+            not self.hs.config.allow_profile_lookup_over_federation
+            and args.get("origin") != self.hs.hostname
+        ):
```
erikjohnston, line 315: Since this is only called from the federation API I think we can just not bother with the check `args.get("origin") != self.hs.hostname`? I think a) it makes it simpler and b) we are sort of trying to limit the places where we do these sort of checks to make any future vhosting efforts easier.

Since you're there could you also add a docstring to the function, something like:

```python
async def on_profile_query(self, args: JsonDict) -> JsonDict:
    """Handles federation profile query requests.
    """
```

erikjohnston (changes requested):
Broadly looks good, thanks! Just a few minor things that could be tidied up

clokep:
@AndrewFerr Looks like this needs some style fixed up! You can run `./scripts-dev/lint.sh` to do this automatically (although the fix in this case is pretty simple).

AndrewFerr:
> @AndrewFerr Looks like this needs some style fixed up! You can run `./scripts-dev/lint.sh` to do this automatically (although the fix in this case is pretty simple).

Sorry about that! The change was so small I assumed it wouldn't need to be linted again. Lesson learned :)

clokep:
Arg and now this seems to be hitting the py3.5 failure which was fixed on develop. We'll need to merge develop into this branch to fix that. 😢

AndrewFerr:
How about I squash my changes and rebase (in a new PR if preferable)?

erikjohnston (approved):
Sorry for taking so long to get back to this!

PR #9206, opened 2021-01-22
Allow testing old deps with postgres

erikjohnston:
This is so that we can test old dependencies with postgres

### DIFF
--- a/.buildkite/scripts/setup_old_deps.sh
+++ b/.buildkite/scripts/setup_old_deps.sh
@@ -9,5 +9,3 @@ apt-get update
 apt-get install -y python3.5 python3.5-dev python3-pip libxml2-dev libxslt-dev xmlsec1 zlib1g-dev tox
 
 export LANG="C.UTF-8"
-
-exec tox -e py35-old,combine
--- a/tox.ini
+++ b/tox.ini
@@ -101,7 +101,7 @@ usedevelop=true
 
 # A test suite for the oldest supported versions of Python libraries, to catch
 # any uses of APIs not available in them.
-[testenv:py35-old]
+[testenv:py35-{old,old-postgres}]
 skip_install=True
 deps =
     # Ensure a version of setuptools that supports Python 3.5 is installed.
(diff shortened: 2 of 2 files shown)


tox.ini
```diff
@@ -114,11 +114,17 @@ deps =
...
+setenv =
+    postgres: SYNAPSE_POSTGRES = 1
+
 commands =
     # Make all greater-thans equals so we test the oldest version of our direct
     # dependencies, but make the pyopenssl 17.0, which can work against an
     # OpenSSL 1.1 compiled cryptography (as older ones don't compile on Travis).
-    /bin/sh -c 'python -m synapse.python_dependencies | sed -e "s/>=/==/g" -e "s/psycopg2==2.6//" -e "s/pyopenssl==16.0.0/pyopenssl==17.0.0/" | xargs -d"\n" pip install'
+    #
+    # Also strip out psycopg2 unless we need it.
```
clokep, line 125: This is to ensure that we don't import it when it isn't available?

.buildkite/scripts/setup_old_deps.sh
```diff
@@ -9,5 +9,3 @@ apt-get update
 apt-get install -y python3.5 python3.5-dev python3-pip libxml2-dev libxslt-dev xmlsec1 zlib1g-dev tox
 
 export LANG="C.UTF-8"
```
richvdh, line 11: this export goes nowhere. I suspect it was here for a reason; you'll need to figure out if that reason was important.

.buildkite/scripts/setup_old_deps.sh
```diff
@@ -9,5 +9,3 @@ apt-get update
...
-
-exec tox -e py35-old,combine
```
richvdh, line 13: there are comments in the header which will need updating.

tox.ini
```diff
@@ -114,11 +114,17 @@ deps =
     coverage
     coverage-enable-subprocess==1.0
 
+setenv =
+    postgres: SYNAPSE_POSTGRES = 1
+
```
richvdh, line 119: I believe this should be magically inherited from the defaults in `testenv`. could you verify one way or the other?

richvdh (changes requested):

PR #9207, opened 2021-01-22
Implementation of MSC2938

Yoric:
### Context

This patch implements (a subset of) MSC 2938, i.e. the ability for users to report content to room moderators instead of system administrators.

### Pull Request Checklist


clokep:
@Yoric asked for some early feedback on this.

synapse/rest/client/v2_alpha/report_event.py
```diff
@@ -58,17 +81,108 @@ async def on_POST(self, request, room_id, event_id):
                 Codes.BAD_JSON,
             )
 
+        target = body.get("org.matrix.msc2938.target", "homeserver-admins")
+
+        if target == "room-moderators":
+            # Report event to room moderators.
+            # This branch has further safety checks (e.g. can the user actually see the event?)
+            return await self._target_room_moderators(room_id=room_id, event_id=event_id, user_id=user_id, reason=body["reason"], score=body["score"])
```
clokep, line 89: I find the `_target` prefix unclear -- maybe `_report_to_room_moderators` and `_report_to_homeserver_admin` would be clearer?

synapse/rest/client/v2_alpha/report_event.py
```diff
@@ -58,17 +81,108 @@ async def on_POST(self, request, room_id, event_id):
...
-            reason=body["reason"],
-            content=body,
+            reason=reason,
+            content=content,
             received_ts=self.clock.time_msec(),
         )
 
         return 200, {}
 
+    async def _target_room_moderators(self, room_id, event_id, user_id, reason, score):
```
clokep, line 112: Please add docstrings and type hints.

This seems *very* different than how we handle the ones for homeserver admins. Does this behavior make sense?

In particular note that the server notices system is disabled by default in Synapse.

In particular is there need for any more state to be kept of these reports (e.g. the admin one gets stored into a table).

synapse/rest/client/v2_alpha/report_event.py
```diff
@@ -58,17 +81,108 @@ async def on_POST(self, request, room_id, event_id):
...
+        # Now make sure that the user was able to witness the event.
+        events = await filter_events_for_client(
+            self.storage,
+            user_id,
+            [event],
+            is_peeking=True
+        )
+        if len(events) == 0:
+            raise SynapseError(
+                HTTPStatus.FORBIDDEN,
```
clokep, line 134: I suspect this should still be a `BAD_REQUEST`.

clokep:
This relates to matrix-org/matrix-doc#2938.

synapse/rest/client/v2_alpha/report_event.py
```diff
@@ -58,17 +81,108 @@ async def on_POST(self, request, room_id, event_id):
...
+                Codes.FORBIDDEN
+            )
+
+        # Find the moderators
+        power_level_event = await self.state.get_current_state(
+            room_id, EventTypes.PowerLevels, ""
+        )
+        ban_level = power_level_event.content.get("ban", 50)
+        kick_level = power_level_event.content.get("kick", 50)
+        moderator_level = max(ban_level, kick_level)
```
clokep, line 145: Does this definition of moderator match other places in the matrix spec?

synapse/rest/client/v2_alpha/report_event.py
```diff
@@ -58,17 +81,108 @@ async def on_POST(self, request, room_id, event_id):
...
+            raise SynapseError(
+                HTTPStatus.FORBIDDEN,
+                "User cannot witness this event",
+                Codes.FORBIDDEN
+            )
+
+        # Find the moderators
+        power_level_event = await self.state.get_current_state(
+            room_id, EventTypes.PowerLevels, ""
+        )
```
clokep, line 142: I suspect this can return `None`, but not 100% sure.

synapse/rest/client/v2_alpha/report_event.py
```diff
@@ -58,17 +81,108 @@ async def on_POST(self, request, room_id, event_id):
...
+                level = power_level_event.content.get("users_default", 0)
+            if level >= moderator_level:
+                hs = UserID.from_string(user_id).domain
+                moderators_on_hs = moderators_by_hs.get(hs, [])
+                moderators_on_hs.append(member_id)
+
+
+        # Dispatch report
+        for (hs, moderators) in moderators_by_hs.items():
+            if hs == self.hs.domain:
```
clokep, line 164: `self.hs.is_mine` is a better way to do this.

synapse/rest/client/v2_alpha/report_event.py
```diff
@@ -31,12 +40,26 @@
...
+
     def __init__(self, hs):
         super().__init__()
         self.hs = hs
         self.auth = hs.get_auth()
+
+
         self.clock = hs.get_clock()
+        self.federation_client = hs.get_federation_client()
+        self.snm = hs.get_server_notices_manager()
```
clokep, line 59: Please use a longer variable name.

synapse/rest/client/v2_alpha/report_event.py
```diff
@@ -58,17 +81,108 @@ async def on_POST(self, request, room_id, event_id):
...
+                            "eventId": event_id,
+                            "userId": user_id,
+                            "score": score,
+                            "reason": reason
+                        }
+                    )
+            else:
+                # Dispatch report to remote homeserver.
+                # FIXME: Dispatch report to remote moderators.
+                pass
```
clokep, line 183: You'll want to do something with the federation handler here most likely. Reading the MSC I don't know if there's quite enough information in there to implement this.

clokep (commented):
I put some thoughts down on the code here. I think my main concern is how different reporting to a room moderator vs. a server admin is (one stores in the db, the other sends a notices message). I'm also a bit concerned that this might not make sense on other homeserver implementations or for all Synapse configs.

synapse/rest/client/v2_alpha/report_event.py
Yoric: > Please add docstrings and type hints.
> 
> This seems _very_ different than how we handle the ones for homeserver admins. Does this behavior make sense?

Well, this is being summarized/discussed [here](https://github.com/matrix-org/matrix-doc/pull/2938). It seems to be the best way for moderators to actually receive the info.

> In particular note that the server notices system is disabled by default in Synapse.

It is enabled on matrix.org and EMS, though, right?

> In particular is there need for any more state to be kept of these reports (e.g. the admin one gets stored into a table).

Not as far as I can tell. It's a simple message. We can't do that for the homeserver admin because the homeserver admin doesn't have an account, but in the future, we could imagine changing how it works for admins.

synapse/rest/client/v2_alpha/report_event.py
Yoric: As far as I can tell, there is no definition of moderator in other places of the matrix spec:

> Clients may wish to assign names to particular power levels. A suggested mapping is as follows: - 0 User - 50 Moderator - 100 Admin

Is pretty much it. Part of the MSC actually defines what a moderator is: someone who can both kick and ban.

synapse/rest/client/v2_alpha/report_event.py
Yoric: Ok, I've been using what I found around the Synapse source code, but I'll adapt :)

ShadowJonathan:
Continuing thoughts from https://github.com/matrix-org/matrix-doc/pull/2938, maybe the best course of actions would be to send EDUs to target servers (which're more likely to be dropped safely if the remote server does not support this), and add a hook into the federation receiving mechanism to check for these EDUs, and then add them to an internal database?

Afterwards, when a potential MSC for homeserver-to-homeserver rooms is made ready, the mechanism can switch to that approach, while keeping the database and federation scoop.

This makes the federation-injection approach a glass-cannon hotfix for now (since it's fairly urgent), with the downside of it not being resilient to server downtime of the other side, while in the future a better approach can be steamrolled.

Yoric:
Ok, ready for a new round of review.

Yoric:
@clokep Review ping?

synapse/rest/client/v2_alpha/report_event.py
```diff
@@ -31,16 +44,39 @@
...
 
+    # As we cannot import HomeServer without creating a circular dependency,
+    # use a forward reference.
+    hs: 'HomeServer'
+    auth: Auth
+    federation_sender: Optional[FederationSender]
+    server_notices: ServerNoticesManager
+    state: StateHandler
+    storage: Storage
+    store: DataStore
```
richvdh, line 55: this should be unnecessary - annotate the `hs` param to `__init__` and mypy/pycharm should figure out the rest.

(it's also not valid syntax in python 3.5, though since this is unlikely to land before we drop support for py3.5, that's probably unimportant.

synapse/event_auth.py
```diff
@@ -409,7 +409,7 @@ def _can_send_event(event: EventBase, auth_events: StateMap[EventBase]) -> bool:
     if user_level < send_level:
         raise AuthError(
             403,
-            "You don't have permission to post that to the room. "
+            ("You don't have permission to post that %s to the room. " % event.type)
```
richvdh, line 412:
```suggestion
            ("You don't have permission to post %s to the room. " % (event.type,))
```

prefer to pass an explicit tuple to %.

synapse/rest/client/v2_alpha/report_event.py
```diff
@@ -58,17 +94,250 @@ async def on_POST(self, request, room_id, event_id):
                 Codes.BAD_JSON,
             )
 
+        target = body.get("org.matrix.msc2938.target", "homeserver-admins")
```
richvdh, line 97: experimental features like this should normally be gated by an `experimental_features` setting: see https://github.com/matrix-org/synapse/blob/develop/synapse/config/experimental.py.

synapse/rest/client/v2_alpha/report_event.py
```diff
@@ -31,16 +44,39 @@
...
         super().__init__()
         self.hs = hs
         self.auth = hs.get_auth()
+
+        self.federation_sender = None
+        if hs.should_send_federation():
+            self.federation_sender = hs.get_federation_sender()
+            hs.get_federation_registry().register_edu_handler(
+                "org.matrix.m.content_report", self.on_receive_report_through_federation
+            )
```
richvdh, line 67: note that whether you will *receive* EDUs on this worker is unrelated to `hs.should_send_federation`:

```suggestion
        hs.get_federation_registry().register_edu_handler(
            "org.matrix.m.content_report", self.on_receive_report_through_federation
        )
```

synapse/rest/client/v2_alpha/report_event.py
```diff
@@ -31,16 +44,39 @@
...
         self.clock = hs.get_clock()
+        self.federation_client = hs.get_federation_client()
+        self.server_notices = hs.get_server_notices_manager()
+        self.state = hs.get_state_handler()
         self.store = hs.get_datastore()
+        self.storage = hs.get_storage()
```
richvdh, line 73: the line between "what goes in `synapse.rest.client`" and "what goes in `synapse.handlers`" is fuzzy, but if you're dealing with federation stuff, that's a pretty clear sign to me that stuff needs factoring out into `synapse.handlers`.

I suggest making an `AbuseReportHandler`, which can be called from the on_POST method here, and can contain the EDU handler.

synapse/rest/client/v2_alpha/report_event.py
```diff
@@ -58,17 +94,250 @@ async def on_POST(self, request, room_id, event_id):
...
+    ):
+        """
+        Report an event to the homeserver admin.
+        This is typically meant to be used to report an entire room, e.g. for harboring illegal activities.
+
+        - room_id The room in which the event took place.
+        - event_id The event to report. We do not check whether the event took place in that room.
+        - user_id The user who reported the event. We do not check whether the user can actually witness the event.
+        - reason The human-readable reason provided by the user.
+        - content A JSON dictionary `{reason: String?, score: Number?}`.
```
richvdh, line 140: Have a look at the rest of Synapse for conventions for documenting parameters. (As per https://github.com/matrix-org/synapse/blob/develop/docs/code_style.md#general-rules, we use the conventions at https://google.github.io/styleguide/pyguide.html#doc-function-args)

richvdh (commented):
@Yoric what are you looking for here? Feedback on the MSC? Feedback on whether the impl matches the MSC? Are there particular parts of the impl you have questions on?

The impl looks generally plausible, though as I point out I'd like to see part of it split out to a Handler class.

I'm a little worried about DoS vectors: it seems like it would be easy for a malicious client or federated server to spew hundreds of abuse reports which stack up into server notices. (Of course, that vector exists today, but it's hard to turn into a proper DoS.)  Maybe we should apply some pre-emptive rate-limiting?

Yoric:
> @Yoric what are you looking for here? Feedback on the MSC? Feedback on whether the impl matches the MSC? Are there particular parts of the impl you have questions on?

Essentially, I'd like to land this at some point during the next few weeks and then start experimenting with using this in the front-end. So, I'm looking for anything that can bring me closer to that end!

(I will apply your feedback asap)

Yoric:
> Maybe we should apply some pre-emptive rate-limiting?

Done.

Yoric:
Note: I still need to write sytests.

richvdh:
[looks like your tests are failing]

synapse/handlers/abuse_reports.py
```diff
@@ -0,0 +1,372 @@
...
+            raise LimitExceededError(
+                retry_after_ms=int(1000 * (time_allowed - time_now_s))
+            )
+
+        room_id = RoomID.from_string(room_id_)
+        event_id = None
+        try:
+            event_id = EventID.from_string(event_id_)
+        except Exception:
+            event_id = EventID.from_string("%s:%s" % (event_id_, self.hs.hostname))
```
richvdh, line 108: this isn't a thing we do anywhere else - event ids should be treated as complete, opaque, identifiers.

Yoric: Yeah, I was misled by the existence of `EventID`.

richvdh (changes requested):
the diff (https://github.com/matrix-org/synapse/pull/9207/files) seems to contain a bunch of unrelated changes: can you fix that so that this can be reviewed, please?

As a matter of general principle, in future please avoid rebasing PR branches once they have received a review, so that it's possible to see how it has changed since the previous review.

Yoric:
Apologies, I'm not entirely clear how that happened.

richvdh:
> > As a matter of general principle, in future please avoid rebasing PR branches once they have received a review, so that it's possible to see how it has changed since the previous review.
> 
> I'm not sure how I can do that when the review process has started 4 months ago and the work on typing has changed much of the underlying code since then. What's the policy in such cases?

I agree if it's a complete rewrite then there's not much value in a merge rather than a rebase, but I'm assuming it hasn't completely changed since I first reviewed this (11 days ago).

Yoric:
> I agree if it's a complete rewrite then there's not much value in a merge rather than a rebase, but I'm assuming it hasn't completely changed since I first reviewed this (11 days ago).

I believe that we're using different meanings of the word "rebase". I'll try and avoid overwriting history in the future!

richvdh:
I'm referring to `git rebase` ;-p. Thanks!

richvdh (changes requested):

synapse/handlers/abuse_reports.py
```diff
@@ -0,0 +1,360 @@
...
+        if target == "room-moderators":
+            # Report event to room moderators as a server notice.
+            # This branch has further safety checks (e.g. can the user actually see the event?)
+            return await self._report_to_room_moderators(
+                room_id=room_id,
+                event_id=event_id,
+                user_id=user_id,
+                reason=body["reason"],
+                score=body["score"],
+                nature=body.get("nature", None),
```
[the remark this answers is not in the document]
Yoric: You're right, I'll add it to the MSC.

synapse/handlers/abuse_reports.py
```diff
@@ -22,114 +22,80 @@
...
-        (
-            allowed,
-            time_allowed,
-        ) = await self._room_key_request_rate_limiter.can_do_action(requester=requester)
-        if not allowed:
-            raise LimitExceededError(
-                retry_after_ms=int(1000 * (time_allowed - time_now_s))
-            )
+        """
+        Report an event as abuse.
```
richvdh, line 61:
```suggestion
        Handle a request from a client to report an event as abuse.
```

synapse/handlers/abuse_reports.py
```diff
@@ -22,114 +22,80 @@
...
-                "Param 'score' must be an integer",
-                Codes.BAD_JSON,
-            )
+        Args:
+            user_id: The user who reported the event. We do not check whether the user can actually witness the event.
+            event_id: The event to report. We do not check whether the event took place in that room.
+            room_id: The room in which the event took place.
+            reason: The human-readable reason provided by the user.
+            content: A JSON dictionary `{reason: String?, score: Number?}`.
+            score Optionally, a "badness" score where -100 is "really bad" and 0 is "acceptable".
```
richvdh, line 71:
```suggestion
            score: Optionally, a "badness" score where -100 is "really bad" and 0 is "acceptable".
```

richvdh, line 71: in what way is it optional? the type declaration is `int`, not `Optional`.

synapse/handlers/abuse_reports.py
```diff
@@ -22,114 +22,80 @@
...
+            nature An optional flag to aid classification and prioritization of abuse reports, e.g. "m.abuse.spam".
```
richvdh, line 72:
```suggestion
            nature: An optional flag to aid classification and prioritization of abuse reports, e.g. "m.abuse.spam".
```

synapse/config/experimental.py
```diff
@@ -36,3 +36,6 @@ def read_config(self, config: JsonDict, **kwargs):
 
         # MSC3026 (busy presence state)
         self.msc3026_enabled = experimental.get("msc3026_enabled", False)  # type: bool
+
+        # MSC2938 (report to moderator)
+        self.msc2938_enabled = experimental.get("msc2938_enabled", False)  # type: bool
```
richvdh, line 41: I don't think this is actually used?

synapse/handlers/abuse_reports.py
```diff
@@ -186,12 +152,13 @@ async def _report_to_room_moderators(
...
-        - reason The human-readable reason provided by the user.
-        - nature An optional flag to aid classification and prioritization of abuse reports, e.g. "m.abuse.spam".
-        - score Optionally, a "badness" score where -100 is "really bad" and 0 is "acceptable".
+        Args:
+            room_id The room in which the event took place.
+            event_id The event to report. We do not check whether the event took place in that room.
+            user_id The user who reported the event. We do not check whether the user can actually witness the event.
+            reason The human-readable reason provided by the user.
+            nature An optional flag to aid classification and prioritization of abuse reports, e.g. "abuse.spam".
+            score Optionally, a "badness" score where -100 is "really bad" and 0 is "acceptable".
```
richvdh, line 161: all need `:`

richvdh: (also, could they be in the right order, please?)

synapse/handlers/abuse_reports.py
```diff
@@ -205,7 +172,9 @@ async def _report_to_room_moderators(
         event = await self.store.get_event(event_id=event_id, check_room_id=None)
         if event.room_id != room_id.to_string():
```
richvdh, line 173: again I ask: why not let `get_event` do this check for you?

Generally it's a pretty poor use of your time and mine for me to make the same review comment multiple times. If something I've written is unclear, just comment on the thread and I'll do my best to clarify.

synapse/rest/client/v2_alpha/report_event.py
```diff
@@ -35,16 +39,48 @@ def __init__(self, hs: "HomeServer"):
...
 
     async def on_POST(self, request: SynapseRequest, room_id, event_id):
-        logger.debug("on_POST %s" % event_id)
         requester = await self.auth.get_user_by_req(request)
         body = parse_json_object_from_request(request)
-        return await self.abuse_report_handler.report(
-            request, requester.user, body, room_id, event_id
+        await self._rate_limiter.ratelimit(requester=requester)
+
+        if not isinstance(body["reason"], str):
```
richvdh, line 55: this will 500 if there is no `reason` property.

```suggestion
        reason = body.get("reason")
        if not isinstance(reason, str):
```

likewise below for `score`.

richvdh (changes requested):
getting there! could you implement the `msc2938_enabled` setting properly please?

richvdh:
@Yoric I believe you're planning to abandon this in favour of MSC3215 - are you happy for us to close this PR?

Yoric:
Yes, let's close it.

PR #9209, opened 2021-01-22
Add an admin API for shadow-banning users.

clokep:
This expands the current shadow-banning feature to be usable via the admin API and adds documentation for it.

A shadow-banned users receives successful responses to their client-server API requests, but the events are not propagated into rooms.

Shadow-banning a user should be used as a tool of last resort and may lead to confusing or broken behaviour for the client.

Most of this PR is just the mechanics of piping through from the REST servlet layer to the datastore, although invalidating the cache is a bit wobbly. I don't think we have a helper for invalidating multiple cache entries at once, but I didn't want to blow away the whole cache.

This builds on the piece of work originally done in #8028 by adding an admin API. (That PR was split up so there's a bunch of other PRs that were actually merged. It is a bit confusing, sorry about that.)

### DIFF
--- a/synapse/storage/databases/main/registration.py
+++ b/synapse/storage/databases/main/registration.py
@@ -360,6 +360,35 @@ class RegistrationWorkerStore(CacheInvalidationWorkerStore):
 
         await self.db_pool.runInteraction("set_server_admin", set_server_admin_txn)
 
+    async def set_shadow_banned(self, user: UserID, shadow_banned: bool) -> None:
+        """Sets whether a user shadow-banned.
+
+        Args:
+            user: user ID of the user to test
+            shadow_banned: true iff the user is to be shadow-banned, false otherwise.
+        """
+
+        def set_shadow_banned_txn(txn):
+            self.db_pool.simple_update_one_txn(
+                txn,
+                table="users",
+                keyvalues={"name": user.to_string()},
+                updatevalues={"shadow_banned": shadow_banned},
+            )
+            # In order for this to apply immediately, reset all access tokens of this user.
+            tokens = self.db_pool.simple_select_onecol_txn(
+                txn,
+                table="access_tokens",
+                keyvalues={"user_id": user.to_string()},
+                retcol="token",
+            )
+            for token in tokens:
+                self._invalidate_cache_and_stream(
+                    txn, self.get_user_by_access_token, (token,)
+                )
+
+        await self.db_pool.runInteraction("set_shadow_banned", set_shadow_banned_txn)
+
     def _query_for_auth(self, txn, token: str) -> Optional[TokenLookupResult]:
         sql = """
             SELECT users.name as user_id,
--- a/synapse/storage/databases/main/registration.py
+++ b/synapse/storage/databases/main/registration.py
@@ -375,7 +375,7 @@ class RegistrationWorkerStore(CacheInvalidationWorkerStore):
                 keyvalues={"name": user.to_string()},
                 updatevalues={"shadow_banned": shadow_banned},
             )
-            # In order for this to apply immediately, reset all access tokens of this user.
+            # In order for this to apply immediately, clear the cache for this user.
             tokens = self.db_pool.simple_select_onecol_txn(
                 txn,
                 table="access_tokens",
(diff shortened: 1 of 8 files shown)


synapse/storage/databases/main/registration.py
```diff
@@ -360,6 +360,35 @@ def set_server_admin_txn(txn):
...
+        """
+
+        def set_shadow_banned_txn(txn):
+            self.db_pool.simple_update_one_txn(
+                txn,
+                table="users",
+                keyvalues={"name": user.to_string()},
+                updatevalues={"shadow_banned": shadow_banned},
+            )
+            # In order for this to apply immediately, reset all access tokens of this user.
```
erikjohnston, line 378: I think you mean clear the cache? When I first read this I thought you meant deleting the access tokens

erikjohnston (approved):

synapse/storage/databases/main/registration.py
clokep: Yeah I mean clear the cache. 👍 I'll update the wording.

dkoao:
IMHO, shadow banning the user is never a good thing. You should be clear and upfront with the user if you don't want them to participate in your servers.

It's honestly disappointing seeing this behaviour from Matrix.

gardenBanter:
Very sad that it has come to this even on Matrix. As the owner of one of the larger homeservers, I think that this will deeply undermine the trust of our users in the service.

pizzadude:
Shadow banning is evil.

callahad:
Hi. I made the call for us to merge this.

You're absolutely correct that shadowbanning is aggressively disrespectful, and should not be used for routine moderation. However, I do believe that legitimate scenarios exist where it is one of the only effective mitigations against persistent, intentional, abusive behavior. And thus, it should be optionally available *as a tool of last resort* to server admins.

The strength federated systems is that each community is empowered to set its own policies around moderation, including whether or not shadowbanning is ever considered acceptable on their homeserver.

> I think that this will deeply undermine the trust of our users in the service.

@gardenBanter Are there ways we could mitigate that diminution in trust?

Perhaps some sort of requirement around intent and disclosure, e.g., making the API disabled by default unless overridden in the config, and somehow publicly exposing whether or not the API is turned on? Or automated transparency reports around its usage on a homeserver?

MiguelX413:
I also think it should be unmerged.

9Morello:
Please, reconsider this.. I also believe this is better off unmerged.
> Perhaps some sort of requirement around intent and disclosure, e.g., making the API disabled by default unless overridden in the config, and somehow publicly exposing whether or not the API is turned on? Or automated transparency reports around its usage on a homeserver?

@callahad Thanks for taking your time to write a serious reply to concerned users. I'm not the person you asked, but imo both are steps in the right direction.

gwofford11:
Shadowbanning is inherently deceitful. This should be unmerged.

arthurrasmusson:
Please remove this functionality.

krruzic:
Please do both of these. I think the other admins concern about trust is a really good point, the idea that an admin can invisibly and permanently silence someone will both lower signups and increase abandonment. The bare minimum is SOME transparency. 

> Perhaps some sort of requirement around intent and disclosure, e.g., making the API disabled by default unless overridden in the config, and somehow publicly exposing whether or not the API is turned on?

ottersarecool:
This is a big ooof, please unmerge this

ara4n:
Hi all: we added shadowbanning back in August in https://github.com/matrix-org/synapse/pull/8034, and while I completely agree that it should **never** be used for moderation purposes, it is nevertheless useful as a measure of last resort against spammers.  We're not going to back it out, nor are we going to back out this PR which simply exposes an API for the server admin (n.b. not room admins) to use.

Meanwhile, it's worth noting there are much worse things that server admins can do to abuse users than shadowbanning (e.g. steal their access_token and snoop on their account; tarpitting them; randomly dropping traffic), so I really don't think this should be controversial.

Something we could consider doing is to turn it off by default, to discourage casual use, and keep it as an emergency measure.

Thanks all for the feedback on this; I think we have enough datapoints to know how strongly some folks feel about this - I'm going to lock the issue now.

PR #9217, opened 2021-01-25
Use an older version of pip for py35 old deps.

clokep:
pip 21.0 dropped support for Python 3.5.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -2,6 +2,7 @@
 envlist = packaging, py35, py36, py37, py38, py39, check_codestyle, check_isort
 
 [base]
+download = False
 deps =
     python-subunit
     junitxml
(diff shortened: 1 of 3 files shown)


callahad:
We're green and ready for review. (Okay, the sytest / workers / pg / redis job is still running, but it's been going 21 minutes without fail and it should be green by the time you read this)

erikjohnston:
It looks like we're not updating pip in the unit tests (even for newer pythons), is that expected? Or am I reading the test output wrong?

tox.ini
```diff
@@ -2,6 +2,7 @@
 envlist = packaging, py35, py36, py37, py38, py39, check_codestyle, check_isort
 
 [base]
+download = False
```
richvdh, line 5: what does this do? I don't think adding entries to `base` does anything unless it gets referenced from another section?

clokep:
I'm going to merge this so our builds work again. It sounds like the team is relatively happy with this from #synapse-dev:matrix.org conversation.

PR #9218, opened 2021-01-25
Periodically send pings to detect dead Redis connections

### LINKED ISSUE #9010: replication connection died
It looks like the connection between synapse and redis died, and synapse didn't notice.

In the logs, we saw:

```
2021-01-03 19:19:16,659 - synapse.metrics.background_process_metrics - 216 - ERROR - send-cmd-55720- Background process 'send-cmd' threw an exception
  File "/usr/local/lib/python3.8/site-packages/synapse/metrics/background_process_metrics.py", line 212, in run
    result = await result
  File "/usr/local/lib/python3.8/site-packages/synapse/replication/tcp/redis.py", line 185, in _async_send_command
    await make_deferred_yieldable(
  File "/usr/local/lib/python3.8/site-packages/twisted/internet/defer.py", line 654, in _runCallbacks
    current.result = callback(current.result, *args, **kw)
  File "/usr/local/lib/python3.8/site-packages/txredisapi.py", line 533, in handle_reply
    raise r
txredisapi.ConnectionError: Lost connection
2021-01-03 19:19:16,696 - synapse.metrics - 576 - INFO - - Collecting gc 1
2021-01-03 19:19:16,857 - twisted - 254 - INFO - send-cmd-55721- Discarding dead connection.
```

Note that there are *two* TCP connections between each synapse process and redis (one for sending, one for receiving - and the one causing the error above is the *sending* side). Nevertheless, it's plausible to imagine that both sockets drop at a similar time, and this is reflected in the "number of clients" graph from redis:

image

(the synchrotron was restarted at 19:47: note the extra connection).

No replication data was received after this point. My hypothesis is that the TCP socket died, and the synapse side didn't notice due to the lack of TCP keepalives or other outgoing data.

erikjohnston:
Hopefully fixes #9010

### DIFF
--- a/synapse/replication/tcp/redis.py
+++ b/synapse/replication/tcp/redis.py
@@ -191,7 +191,7 @@ class RedisSubscriber(txredisapi.SubscriberProtocol, AbstractConnection):
 
 
 class SynapseRedisFactory(txredisapi.RedisFactory):
-    """A subclass of RedisFactory that ensures that periodically sends pings
+    """A subclass of RedisFactory that periodically sends pings
     to ensure that we detect dead connections.
     """
 
--- a/synapse/replication/tcp/redis.py
+++ b/synapse/replication/tcp/redis.py
@@ -60,16 +60,16 @@ class RedisSubscriber(txredisapi.SubscriberProtocol, AbstractConnection):
     immediately after initialisation.
 
     Attributes:
-        handler: The command handler to handle incoming commands.
-        stream_name: The *redis* stream name to subscribe to and publish from
-            (not anything to do with Synapse replication streams).
-        outbound_redis_connection: The connection to redis to use to send
+        synapse_handler: The command handler to handle incoming commands.
+        synapse_stream_name: The *redis* stream name to subscribe to and publish
+            from (not anything to do with Synapse replication streams).
+        synapse_outbound_redis_connection: The connection to redis to use to send
             commands.
     """
 
-    handler = None  # type: ReplicationCommandHandler
-    stream_name = None  # type: str
-    outbound_redis_connection = None  # type: txredisapi.RedisProtocol
+    synapse_handler = None  # type: ReplicationCommandHandler
+    synapse_stream_name = None  # type: str
+    synapse_outbound_redis_connection = None  # type: txredisapi.RedisProtocol
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
@@ -89,19 +89,19 @@ class RedisSubscriber(txredisapi.SubscriberProtocol, AbstractConnection):
         # it's important to make sure that we only send the REPLICATE command once we
         # have successfully subscribed to the stream - otherwise we might miss the
         # POSITION response sent back by the other end.
-        logger.info("Sending redis SUBSCRIBE for %s", self.stream_name)
-        await make_deferred_yieldable(self.subscribe(self.stream_name))
+        logger.info("Sending redis SUBSCRIBE for %s", self.synapse_stream_name)
+        await make_deferred_yieldable(self.subscribe(self.synapse_stream_name))
         logger.info(
             "Successfully subscribed to redis stream, sending REPLICATE command"
         )
-        self.handler.new_connection(self)
+        self.synapse_handler.new_connection(self)
         await self._async_send_command(ReplicateCommand())
         logger.info("REPLICATE successfully sent")
 
         # We send out our positions when there is a new connection in case the
         # other side missed updates. We do this for Redis connections as the
         # otherside won't know we've connected and so won't issue a REPLICATE.
-        self.handler.send_positions_to_connection(self)
+        self.synapse_handler.send_positions_to_connection(self)
 
     def messageReceived(self, pattern: str, channel: str, message: str):
         """Received a message from redis.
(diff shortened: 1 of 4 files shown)


synapse/replication/tcp/redis.py
```diff
@@ -206,34 +255,37 @@ def __init__(
...
-        super().__init__()
-
-        # This sets the password on the RedisFactory base class (as
-        # SubscriberFactory constructor doesn't pass it through).
-        self.password = hs.config.redis.redis_password
+        super().__init__(
+            hs,
+            "subscriber",
+            None,
+            1,
```
clokep, line 262: It would probably be clearer to use kwargs for these. (`None` and `1` seem to be dbid and poolsize, but I had to fliip back and forth between classes to see that.)

synapse/replication/tcp/redis.py
```diff
@@ -206,34 +255,37 @@ def __init__(
...
+        p.synapse_handler = self.synapse_handler
+        p.synapse_outbound_redis_connection = self.synapse_outbound_redis_connection
+        p.synapse_stream_name = self.synapse_stream_name
 
         return p
 
 
 def lazyConnection(
-    reactor,
+    hs: "HomeServer",
```
clokep, line 288: The docstring for this could use updating.

synapse/replication/tcp/redis.py
```diff
@@ -183,11 +184,59 @@ async def _async_send_command(self, cmd: Command):
...
+        handler: Type = txredisapi.ConnectionHandler,
+        charset: str = "utf-8",
+        password: Optional[str] = None,
+        replyTimeout: Optional[int] = None,
+        convertNumbers: Optional[int] = True,
+    ):
+        # We want to ensure that we timeout when sending pings on dead
+        # connections, rather than just hanging.
+        if replyTimeout is None:
+            replyTimeout = 30
```
clokep, line 214: Should we just set this as the default above?

synapse/replication/tcp/redis.py
```diff
@@ -183,11 +184,54 @@ async def _async_send_command(self, cmd: Command):
...
+
+    @wrap_as_background_process("redis_ping")
+    async def _send_ping(self):
+        for connection in self.pool:
+            try:
+                await make_deferred_yieldable(connection.ping())
+            except Exception:
+                logger.warning("Failed to send ping to a redis connection")
 
-class RedisDirectTcpReplicationClientFactory(txredisapi.SubscriberFactory):
```
clokep, line 190: Looks like `SubscriberFactory` is actually [just `RedisFactory` with some defaults](https://github.com/IlyaSkriblovsky/txredisapi/blob/c89a49829d1c48baa10d03c41d39c261c02ad6e1/txredisapi.py#L2366-L2371). So this change is really using our own `SynapseRedisFactory` instead, which uses different defaults (+ handles pinging).

I'm guessing you did the ping logic in a separate class so it can be re-used for #9198? Or just for separation of concerns?

erikjohnston: I should have made that clear. The reason that its separate is that we can then use `SynapseRedisFactory` for the normal non-subscriber connection, and so share the ping logic between the two.

synapse/replication/tcp/redis.py
```diff
@@ -183,11 +184,54 @@ async def _async_send_command(self, cmd: Command):
...
-            self.outbound_redis_connection.publish(self.stream_name, encoded_string)
+            self.synapse_outbound_redis_connection.publish(
+                self.synapse_stream_name, encoded_string
+            )
+        )
+
+
+class SynapseRedisFactory(txredisapi.RedisFactory):
+    """A subclass of RedisFactory that ensures that periodically sends pings
+    to ensure that we detect dead connections.
```
clokep, line 195:
```suggestion
    """A subclass of RedisFactory that periodically sends pings
    to ensure that we detect dead connections.
```

synapse/replication/tcp/redis.py
```diff
@@ -206,65 +250,62 @@ def __init__(
...
 
 def lazyConnection(
-    reactor,
+    hs: "HomeServer",
     host: str = "localhost",
     port: int = 6379,
     dbid: Optional[int] = None,
     reconnect: bool = True,
-    charset: str = "utf-8",
     password: Optional[str] = None,
```
clokep, line 288: Not sure if password is still useful or if we should just use `hs.config.redis.redis_password` 🤷

synapse/replication/tcp/redis.py
```diff
@@ -206,65 +250,62 @@ def __init__(
...
     ):
 
-        super().__init__()
-
-        # This sets the password on the RedisFactory base class (as
-        # SubscriberFactory constructor doesn't pass it through).
-        self.password = hs.config.redis.redis_password
+        super().__init__(
+            hs,
+            uuid="subscriber",
```
clokep, line 255: I think previously we defaulted this to `None`? Not sure if this change matters much?

erikjohnston: Yeah, I don't think it matters but it was annoying me that they want a string and we were passing in a None.

synapse/replication/tcp/redis.py
erikjohnston: Yeah, I'm slightly in two minds about this. Either we read the host/port/dbid/password from the config or we pass them all in, I think it only really matters if we think we might have multiple redis instances configured :shrug:

synapse/replication/tcp/redis.py
clokep: Fair enough! 👍  I wasn't sure if this truly need to be unique is the only reason I raised an eyebrow.

erikjohnston: I think it only needs to be unique for concurrent connections, and we'll only ever have one subscriber connection :+1:

synapse/replication/tcp/redis.py
clokep: Ah, fair enough about the host/port. Being consistent sounds good. Ignore this!

clokep (approved):
LGTM!

PR #9221, opened 2021-01-25
Clarify security note regarding the domain Synapse is hosted on.

dkasak:
The intent is to make it clearer that running Synapse on the same domain as another sensitive web application is not a good idea due to possibility of XSS.

### Pull Request Checklist

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -149,21 +149,24 @@ For details on having Synapse manage your federation TLS certificates
 automatically, please see `<docs/ACME.md>`_.
 
 
-Security Note
+Security note
 =============
 
-Matrix serves raw user generated data in some APIs - specifically the `content
-repository endpoints <https://matrix.org/docs/spec/client_server/latest.html#get-matrix-media-r0-download-servername-mediaid>`_.
+Matrix serves raw, user-supplied data in some APIs -- specifically the `content
+repository endpoints`_.
 
-Whilst we have tried to mitigate against possible XSS attacks (e.g.
-https://github.com/matrix-org/synapse/pull/1021) we recommend running
-matrix homeservers on a dedicated domain name, to limit any malicious user generated
-content served to web browsers a matrix API from being able to attack webapps hosted
-on the same domain.  This is particularly true of sharing a matrix webclient and
-server on the same domain.
+.. _content repository endpoints: https://matrix.org/docs/spec/client_server/latest.html#get-matrix-media-r0-download-servername-mediaid
 
-See https://github.com/vector-im/riot-web/issues/1977 and
-https://developer.github.com/changes/2014-04-25-user-content-security for more details.
+Whilst we make a reasonable effort to mitigate against XSS attacks (such as
+using `CSP`_), ideally a Matrix homeserver should be hosted on a dedicated
+domain name not shared by other web applications. This particularly applies to
+sharing the domain with Matrix web clients and other sensitive applications
+like webmail.
+
+.. _CSP: https://github.com/matrix-org/synapse/pull/1021
+
+Using a separate domain ensures that even if an XSS is found in Synapse, the
+impact to other applications will be minimal.
 
 
 Upgrading an existing Synapse
--- a/README.rst
+++ b/README.rst
@@ -157,15 +157,36 @@ repository endpoints`_.
 
 .. _content repository endpoints: https://matrix.org/docs/spec/client_server/latest.html#get-matrix-media-r0-download-servername-mediaid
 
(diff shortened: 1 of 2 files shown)


clokep:
@dkasak Can you add a newsfile to this? Something like `changelog.d/9221.doc` -> "Clarify security note regarding hosting Synapse on the same domain as other web applications."

dkasak:
Done.

clokep:
Thank you! 👍 I think the CI error is fixed in #9146 / unrelated.

README.rst
```diff
@@ -149,21 +149,24 @@ For details on having Synapse manage your federation TLS certificates
...
-content served to web browsers a matrix API from being able to attack webapps hosted
-on the same domain.  This is particularly true of sharing a matrix webclient and
-server on the same domain.
+.. _content repository endpoints: https://matrix.org/docs/spec/client_server/latest.html#get-matrix-media-r0-download-servername-mediaid
 
-See https://github.com/vector-im/riot-web/issues/1977 and
-https://developer.github.com/changes/2014-04-25-user-content-security for more details.
+Whilst we make a reasonable effort to mitigate against XSS attacks (such as
+using `CSP`_), ideally a Matrix homeserver should be hosted on a dedicated
+domain name not shared by other web applications. This particularly applies to
```
clokep, line 162: Should we define better what a "dedicated domain name" means? Is a subdomain OK (e.g. matrix.example.com where other web apps might be on example.com or webmail.example.com)?

dkasak: The intention is to recommend a completely separate domain name, not a subdomain. This is because hosting on a subdomain has some subtleties and gotchas around it, such as a subdomain being able to set cookies for its ancestor domains, which then also get sent to sibling subdomains. So I think we should say that the ideal scenario is to host on a domain that is completely unrelated to any sensitive web apps (i.e. `example1.com` and `example2.com`).

How about

> Whilst we make a reasonable effort to mitigate against XSS attacks (such as using `CSP`_), the ideal scenario is for the Matrix homeserver to be hosted on a domain that is completely separate from any domain hosting any other sensitive web applications.
>
> That is, if you have a sensitive web application running on `example1.com`, the `public_baseurl` setting in Synapse's config file should not be set to `example1.com` nor any of its subdomains, but to another domain `example2.com`.
>
> Note that most of the risk involved is mitigated by ensuring that the homeserver is at least hosted on a different subdomain. Therefore, at the very least, you should ensure that if a sensitive web application is running on `example1.com`, the `public_baseurl` is something like `subdomain.example1.com`. However, there are some subtleties around this, such as subdomains being able to set cookies on their ancestor domains. Whether this would be problematic will depend on the concrete web applications involved.
>
> Conclusion: Ideally, host on a domain that is completely separate from domains used by other applications. At the very least, host on a subdomain that is not shared by a sensitive application.

I also see now that I accidentally dropped the developer.github.com reference and the previous discussion in the riot-web repo. The former seems useful to have, but I'm not sure the latter clarifies anything or whether it just makes it more confusing (judging by the comments there).

dkasak: Of course, since this is now mentioning the public base *URL*, we can't talk about it being set *to* a *domain name*. *sigh* I'll revise.

anoadragon453:
Heads up that merging in develop will be required to allow CI to pass due to us recently dropping Py3.5 support.

README.rst
richvdh: @dkasak are you still planning to revise this? I think that some examples would certainly be helpful here.

dkasak: Yes, *examples*, that's what I need! Thanks. And yes, I'll try taking another stab at it soon (probably tomorrow).

README.rst
```diff
@@ -149,21 +149,45 @@ For details on having Synapse manage your federation TLS certificates
...
+
+.. _some attacks: https://en.wikipedia.org/wiki/Session_fixation#Attacks_using_cross-subdomain_cookie
+
+To illustrate this with an example, if your Element Web or other sensitive web
+application is hosted on ``A.example1.com``, you should ideally host Synapse on
+``example2.com``. Some amount of protection is offered by hosting on
+``B.example1.com`` instead, so this is also acceptable in some scenarios.
+However, you should *not* host your Synapse on ``A.example1.com``.
+
+Note that all of the above refers exclusively to the domain used in Synapse's
```
dkasak, line 185: I'm not sure whether this paragraph is confusing or clarifying. I thought it was good to mention the setting it is talking about explicitly.

richvdh: I think it's good.

richvdh (approved):
ltgm!

PR #9223, opened 2021-01-25
Add type hints to various handlers.

clokep:
This adds type hints to almost all the remaining handlers that don't have them (except the `e2e_*` ones, which I have in a separate branch). This should be reviewable commit-by-commit. Most of it is straightforward, but I'll mark a couple of odd spots.

### DIFF
--- a/synapse/handlers/acme.py
+++ b/synapse/handlers/acme.py
@@ -14,6 +14,7 @@
 # limitations under the License.
 
 import logging
+from typing import TYPE_CHECKING
 
 import twisted
 import twisted.internet.error
@@ -21,6 +22,9 @@ from twisted.web import server, static
 from twisted.web.resource import Resource
 
 from synapse.app import check_bind_error
+
+if TYPE_CHECKING:
+    from synapse.app.homeserver import HomeServer
 
 logger = logging.getLogger(__name__)
 
@@ -35,12 +39,12 @@ solutions, please read https://github.com/matrix-org/synapse/blob/master/docs/AC
 
 
 class AcmeHandler:
-    def __init__(self, hs):
+    def __init__(self, hs: "HomeServer"):
         self.hs = hs
         self.reactor = hs.get_reactor()
         self._acme_domain = hs.config.acme_domain
 
-    async def start_listening(self):
+    async def start_listening(self) -> None:
         from synapse.handlers import acme_issuing_service
 
         # Configure logging for txacme, if you need to debug
@@ -85,7 +89,7 @@ class AcmeHandler:
             logger.error(ACME_REGISTER_FAIL_ERROR)
             raise
 
-    async def provision_certificate(self):
+    async def provision_certificate(self) -> None:
 
         logger.warning("Reprovisioning %s", self._acme_domain)
 
@@ -110,5 +114,3 @@ class AcmeHandler:
         except Exception:
             logger.exception("Failed saving!")
             raise
-
-        return True
--- a/synapse/handlers/acme_issuing_service.py
+++ b/synapse/handlers/acme_issuing_service.py
@@ -22,8 +22,10 @@ only need (and may only have available) if we are doing ACME, so is designed to 
 imported conditionally.
 """
 import logging
+from typing import Dict, Iterable, List
 
 import attr
+import pem
 from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives import serialization
 from josepy import JWKRSA
@@ -36,20 +38,27 @@ from txacme.util import generate_private_key
 from zope.interface import implementer
 
 from twisted.internet import defer
+from twisted.internet.interfaces import IReactorTCP
 from twisted.python.filepath import FilePath
 from twisted.python.url import URL
+from twisted.web.resource import IResource
 
 logger = logging.getLogger(__name__)
 
 
-def create_issuing_service(reactor, acme_url, account_key_file, well_known_resource):
+def create_issuing_service(
+    reactor: IReactorTCP,
+    acme_url: str,
+    account_key_file: str,
+    well_known_resource: IResource,
+) -> AcmeIssuingService:
     """Create an ACME issuing service, and attach it to a web Resource
 
     Args:
         reactor: twisted reactor
-        acme_url (str): URL to use to request certificates
-        account_key_file (str): where to store the account key
-        well_known_resource (twisted.web.IResource): web resource for .well-known.
+        acme_url: URL to use to request certificates
+        account_key_file: where to store the account key
+        well_known_resource: web resource for .well-known.
             we will attach a child resource for "acme-challenge".
 
     Returns:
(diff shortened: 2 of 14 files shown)


synapse/handlers/acme.py
```diff
@@ -110,5 +114,3 @@ async def provision_certificate(self):
         except Exception:
             logger.exception("Failed saving!")
             raise
-
-        return True
```
clokep, line 114: This method is just awiated and the return type is never used.

synapse/handlers/groups_local.py
```diff
@@ -284,27 +295,7 @@ async def create_group(self, group_id, user_id, content):
...
         else:
-            local_attestation = self.attestations.create_attestation(group_id, user_id)
-            content["attestation"] = local_attestation
-
-            content["user_profile"] = await self.profile_handler.get_profile(user_id)
-
-            try:
-                res = await self.transport_client.create_group(
-                    get_domain_from_id(group_id), group_id, user_id, content
-                )
```
clokep, line 295: [`TransportLayerClient`](https://github.com/matrix-org/synapse/blob/30fba6210834a4ecd91badf0c8f3eb278b72e746/synapse/federation/transport/client.py#L33) does not have a `create_group`, since the groups code is to be replaced it seemed prudent to just fail quickly here.

synapse/handlers/search.py
```diff
@@ -418,13 +427,10 @@ async def search(self, user, content, batch=None):
 
         state_results = {}
         if include_state:
-            rooms = {e.room_id for e in allowed_events}
-            for room_id in rooms:
+            for room_id in {e.room_id for e in allowed_events}:
                 state = await self.state_handler.get_current_state(room_id)
                 state_results[room_id] = list(state.values())
 
-            state_results.values()
```
clokep, line 426: I don't think this was doing anything, but maybe there was *magic* it was doing?

synapse/handlers/user_directory.py
```diff
@@ -145,10 +145,6 @@ async def _unsafe_process(self) -> None:
         if self.pos is None:
             self.pos = await self.store.get_user_directory_stream_pos()
 
-        # If still None then the initial background update hasn't happened yet
-        if self.pos is None:
-            return None
```
clokep, line 150: `get_user_directory_stream_pos` always returns an `int` by my reading so this code was unreachable.

synapse/handlers/user_directory.py
```diff
@@ -232,7 +228,7 @@ async def _handle_deltas(self, deltas: List[Dict[str, Any]]) -> None:
                         continue
 
                     if change:  # The user joined
-                        event = await self.store.get_event(event_id, allow_none=True)
```
clokep, line 235: Right below this we assume that the event exists, so we either need to not `allow_none` *or* handle `event` being `None`.

synapse/storage/databases/main/stats.py
```diff
@@ -339,7 +340,7 @@ async def get_earliest_token_for_stats(self, stats_type: str, id: str) -> int:
         )
 
     async def bulk_update_stats_delta(
-        self, ts: int, updates: Dict[str, Dict[str, Dict[str, Counter]]], stream_id: int
+        self, ts: int, updates: Dict[str, Dict[str, Counter[str]]], stream_id: int
```
clokep, line 343: This was just wrong, `Counter` is already a (special) mapping.

synapse/handlers/acme_issuing_service.py
```diff
@@ -83,18 +92,20 @@ class ErsatzStore:
...
-    def store(self, server_name, pem_objects):
+    def store(
+        self, server_name: bytes, pem_objects: Iterable[pem.AbstractPEMObject]
+    ) -> defer.Deferred:
         self.certs[server_name] = [o.as_bytes() for o in pem_objects]
         return defer.succeed(None)
 
 
-def load_or_create_client_key(key_file):
+def load_or_create_client_key(key_file: str) -> JWKRSA:
```
erikjohnston, line 104: I wonder if we should be writing stubs for libraries. Since we ignore the lib in mypy.ini it'll be treated as `Any`, which is distinctly unhelpful, and perhaps even worse than not having a type hint at all?

We do this in other places though, so happy to leave this as is and discuss the broader issue separately.

synapse/handlers/search.py
erikjohnston: That looks spurious to me. `Dict.values` shouldn't have any side effect.

synapse/handlers/user_directory.py
erikjohnston: Hmm. So I guess this could break if we get a rejected event in our state delta? Since this is a background job that won't make any progress if it breaks(?) I'd be tempted to keep the `allow_none` and ignore if it is `None` (as if change was false).

synapse/handlers/stats.py
```diff
@@ -110,10 +122,10 @@ async def _unsafe_process(self):
             )
 
             for room_id, fields in room_count.items():
-                room_deltas.setdefault(room_id, {}).update(fields)
+                room_deltas.setdefault(room_id, Counter()).update(fields)
 
             for user_id, fields in user_count.items():
-                user_deltas.setdefault(user_id, {}).update(fields)
+                user_deltas.setdefault(user_id, Counter()).update(fields)
```
erikjohnston, line 122: I think this is fine because `fields` is a `Dict[str, int]`, which in this case is effectively the same as `Counter[str]`?

clokep: Pretty much, the alternative would be to make `user_deltas` a `Dict[str, Dict[str, int]]` instead of a `Dict[str, Counter[str]]` I think? Would that be better?

synapse/handlers/acme_issuing_service.py
clokep: I think that would make sense to do as a separate step? 🤷 The ACME stuff in particular is deprecated I think, so maybe not worthwhile to write stubs there.

synapse/handlers/stats.py
erikjohnston: Nah, it's fine, just wanted to check my understanding of the change was correct :)

synapse/handlers/acme_issuing_service.py
erikjohnston: True, I just know that we sometimes get confused at why mypy isn't picking up obvious bugs because of this sort of thing, but that can be dealt with separately

synapse/handlers/user_directory.py
clokep: Ah, that probably makes more sense. 👍

clokep: Looking closer at this code, I don't think just ignoring it is the right change, I think we still need to handle that it occurred, but not try to do anything with profile data?

erikjohnston: Oh, possibly. TBH I'm not too concerned as this is a) for the user directory and b) for a case which really shouldn't happen (though has in the past)

clokep: I made this change!

synapse/handlers/acme_issuing_service.py
clokep: I filed #9228.

erikjohnston (approved):

Issue #8334, opened 2020-09-16
Investigate if there are unspecced endpoints that can be removed


clokep:
We should investigate whether there are any unspecced (and unused) endpoints in Synapse.

See #8177 about some of the admin APIs.

I took a look at some of the federation APIs and client APIs also, will put some results in here tomorrow.

clokep:
I collated a bunch of this on [a spreadsheet](https://docs.google.com/spreadsheets/d/1PLAf7BTOp0mAAj5BSZAHp1CFkHDcjAYxuvKYK-ObIQE/), but a summary is below.

## Federation

### Unspecced

* The various groups endpoints (under `/_matrix/federation/v1/groups/`), but are to be replaced (I believe).
* A room complexity endpoint (`/_matrix/federation/unstable/rooms/{roomId}/complexity`) was added in #5216.

### Errors

* `/_matrix/federation/v1/3pid/onbind` (implemented by [`On3pidBindServlet`](https://github.com/matrix-org/synapse/blob/c619253db80c8d1c606dc40756dd3c9e3a55a9fb/synapse/federation/transport/server.py#L666-L671)) accepts `POST`, while [the specification](https://matrix.org/docs/spec/server_server/latest#put-matrix-federation-v1-3pid-onbind) says it should accept `PUT`.

## Client

### Prefix

* All endpoints appear under `/_matrix/client/unstable/`, which is probably OK.
* Most endpoints appear under `/_matrix/client/r0/`, unless they're unstable endpoints.
* Some endpoints appear under `/_matrix/client/api/v1/`.

### Unspecced

* The various groups endpoints (under `/_matrix/client/(unstable|r0)/groups/`, `/_matrix/client/(unstable|r0)/create_group`, under `/_matrix/client/(unstable|r0)/publicised_groups`, and `/_matrix/client/(unstable|r0)/joined_groups`), but are to be replaced (I believe).
* `/_matrix/client/(unstable|api/v1|r0)/pushers/remove` (implemented by [`PushersRemoveRestServlet`](https://github.com/matrix-org/synapse/blob/ea26e9a98b0541fc886a1cb826a38352b7599dbe/synapse/rest/client/v1/pusher.py#L147-L186)).
* Listing appservices (`/_matrix/client/(unstable|api/v1|r0)/directory/list/appservice/{networkId}/{roomId}`), added in #1676.
[...]

PR #6585, opened 2019-12-21
Support CS API v0.6.0

anoadragon453:
Figured I'd get the ball rolling on this one so we don't forget about it.

Fixes https://github.com/matrix-org/synapse/issues/6045

PR #9224, opened 2021-01-26
r0.6.1 support: /rooms/:roomId/aliases endpoint

turt2live:
*Author's note: Please read https://github.com/matrix-org/synapse/pull/9224#issuecomment-767869965*

[MSC2432](https://github.com/matrix-org/matrix-doc/pull/2432) added this endpoint originally but it has since been included in the spec for nearly a year. 

~~We can't reasonably remove the unstable endpoint as client implementations still rely on it, and those clients cannot rely on `r0.6.x` appearing in the `/versions` to detect if this stable endpoint exists. It's proposed to introduce a transition period for implementations which may be stuck on the old endpoint of 2 months (minimum). After that point, Synapse can (and should) remove the unstable endpoint without warning. The exact timelines are to be determined by the synapse team, however.~~

This is progress towards https://github.com/matrix-org/synapse/issues/8334

### Pull Request Checklist


richvdh:
we should ask ourselves how we missed this. It should have been checked when we started declaring support for r0.6 (#6585, ftr).

changelog.d/9224.misc
```diff
@@ -0,0 +1 @@
+Add stable `/rooms/:roomId/aliases` endpoint from r0.6.0 (MSC2432 historically).
```
clokep, line 1: I think this is a bugfix?

turt2live: it could also be a feature :P 

I'm not sure this case was considered when building the changelog system, so happy to do whatever people are most happy with.

clokep:
> we should ask ourselves how we missed this. It should have been checked when we started declaring support for r0.6 (#6585, ftr).

Do we have any process for finding these? I wonder if having a helper to generate the experimental URLs would help auditing all experimental features easier. 🤷

turt2live:
IPAc1CV

Synapse doesn't actually support r0.6.1 which is when this endpoint was introduced. In the spec's `rX` numbering series there is a patch-but-not-really version where this endpoint got introduced (I think practically we should have brought it in for r0.7, but whatever). Synapse doesn't advertise support for r0.6.1 and doesn't support much of it, so this is actually not an omission and in fact work towards a feature.

This should be handled as a very late transition from unstable -> stable, which clients should be made to understand. Leaving to @callahad to coordinate.

clokep:
Note that https://matrix.org/docs/spec/client_server/r0.6.1#changelog is the rest of the changes that were in r0.6.1, with a quick look I think Synapse has all those changes already.

richvdh:
I think we should also ask ourselves how this trivial PR ended up bit-rotting for 4 months :/

changelog.d/9224.misc
richvdh: yeah sounds like it should be a feature. also:

```suggestion
Add new endpoint `/_matrix/client/r0/rooms/{roomId}/aliases` from Client-Server API r0.6.1 (previously [MSC2432](https://github.com/matrix-org/matrix-doc/pull/2432)).
```

richvdh (approved):

synapse/rest/client/v1/room.py
```diff
@@ -892,7 +892,7 @@ class RoomAliasListServlet(RestServlet):
             r"^/_matrix/client/unstable/org\.matrix\.msc2432"
             r"/rooms/(?P<room_id>[^/]*)/aliases"
         ),
-    ]
+    ] + list(client_patterns("/rooms/(?P<room_id>[^/]*)/aliases$"))
```
richvdh, line 895:
```suggestion
    ] + client_patterns("/rooms/(?P<room_id>[^/]*)/aliases$", unstable=False)
```

richvdh (changes requested):

synapse/rest/client/v1/room.py
richvdh: it's already a list. more importantly, we shouldn't add it to `unstable`.

richvdh: oh does dropping the `list` upset mypy? if so, ignore that bit.

turt2live:
> I think we should also ask ourselves how this trivial PR ended up bit-rotting for 4 months :/

Like with most things I contribute, time and energy got reallocated and this ended up being forgotten about with no one to really coordinate it.

richvdh:
> > I think we should also ask ourselves how this trivial PR ended up bit-rotting for 4 months :/
> 
> Like with most things I contribute, time and energy got reallocated and this ended up being forgotten about with no one to really coordinate it.

To be clear, I see the main failure here being within the Synapse team process. It certainly wasn't a dig at you as contributor.

richvdh:
the remaining failures are https://github.com/matrix-org/sytest/issues/1055, so I'll close this.

PR #9227, opened 2021-01-26
Add a note to changelog about redis usage

erikjohnston:
c.f. #9198

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,9 @@
+Unreleased
+==========
+
+Note that this release includes a change in Synapse to use Redis as a cache, as well as a pub/sub mechanism, if Redis support is enabled. No action is needed by server administrators.
+
+
 Synapse 1.26.0rc2 (2021-01-25)
 ==============================
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,7 +1,7 @@
 Unreleased
 ==========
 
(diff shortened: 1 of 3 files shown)


CHANGES.md
```diff
@@ -1,3 +1,9 @@
+Unreleased
+==========
+
+Note that this release includes a change in Synapse to use Redis as a cache, as well as a pub/sub mechanism, if Redis support is enabled. No action is needed by server administrators.
```
erikjohnston, line 4: I don't know if its worth saying that we don't expect much increase in memory usage? Matrix.org is currently running at ~20MB.

clokep (approved):
This seems OK. I do wonder if people will have questions about what is being cached? 🤷

CHANGES.md
clokep: It might be worth mentioning that we don't expect the Redis usage to dramatically increase.

clokep:
Should we update anything in the workers doc: https://github.com/matrix-org/synapse/blob/master/docs/workers.md#main-processworker-communication (or the arch diagram in it).

CHANGES.md
erikjohnston: Done

erikjohnston:
> This seems OK. I do wonder if people will have questions about what is being cached?

I think at that point they can look at the PR, I don't think its useful to spell it out anywhere, especially since I imagine we'll be caching more stuff in Redis as time goes by.

Issue #9213, opened 2021-01-23
The default OpenId Connect user mapping provider fail to initialize with default config


tmortagne:
Enable OpenID Connect, configure the required properties (`issuer`, `client_id`, etc.) and don't change anything to `user_mapping_provider` (i.e. use the default user mapping behavior).

The OIDC section of the configuration look like this:

```
# Enable OpenID Connect (OIDC) / OAuth 2.0 for registration and login.
#
# See https://github.com/matrix-org/synapse/blob/master/docs/openid.md
# for some example configurations.
#
oidc_config:
  # Uncomment the following to enable authorization against an OpenID Connect
  # server. Defaults to false.
  #
  enabled: true

  # Uncomment the following to disable use of the OIDC discovery mechanism to
  # discover endpoints. Defaults to true.
  #
  #discover: false

  # the OIDC issuer. Used to validate tokens and (if discovery is enabled) to
  # discover the provider's endpoints.
  #
  # Required if 'enabled' is true.
  #
  issuer: "https://myhost/oidc/"

  # oauth2 client id to use.
  #
  # Required if 'enabled' is true.
  #
  client_id: "matrix"

  # oauth2 client secret to use.
  #
  # Required if 'enabled' is true.
  #
  client_secret: "dontcare"

  # auth method to use when exchanging the token.
  # Valid values are 'client_secret_basic' (default), 'client_secret_post' and
  # 'none'.
  #
  #client_auth_method: client_secret_post

  # list of scopes to request. This should normally include the "openid" scope.
  # Defaults to ["openid"].
  #
  #scopes: ["openid", "profile"]

  # the oauth2 authorization endpoint. Required if provider discovery is disabled.
  #
  #authorization_endpoint: "https://accounts.example.com/oauth2/auth"
  # the oauth2 token endpoint. Required if provider discovery is disabled.
  #
  #token_endpoint: "https://accounts.example.com/oauth2/token"

  # the OIDC userinfo endpoint. Required if discovery is disabled and the
  # "openid" scope is not requested.
  #
  #userinfo_endpoint: "https://accounts.example.com/userinfo"

  # URI where to fetch the JWKS. Required if discovery is disabled and the
  # "openid" scope is used.
  #
  #jwks_uri: "https://accounts.example.com/.well-known/jwks.json"

  # Uncomment to skip metadata verification. Defaults to false.
  #
  # Use this if you are connecting to a provider that is not OpenID Connect
  # compliant.
  # Avoid this in production.
[...]

PR #9229, opened 2021-01-26
Pass a dict, instead of None, to modules if a None config is specified in the homeserver config

anoadragon453:
Fixes https://github.com/matrix-org/synapse/issues/9213

If a Synapse module's config block were empty in YAML, thus being translated to a `Nonetype` in Python, then some modules could fail as that None ends up getting passed to their `parse_config` method. Modules are expected to accept a `dict` instead.

This PR ensures that if the user does end up specifying an empty config block (such as what [the default oidc config in the sample config](https://github.com/matrix-org/synapse/blob/5310808d3bebd17275355ecd474bc013e8c7462d/docs/sample_config.yaml#L1816-L1845) states) then `None` is not passed to the module. An empty dict is passed instead.

This code assumes that no existing modules are relying on receiving a `None` config block, but I'd really hope that they aren't.


changelog.d/9229.misc
```diff
@@ -0,0 +1 @@
+Fix a bug where `None` was passed to Synapse modules instead of an empty dictionary if an empty module `config` block was provided in the homeserver config.
```
erikjohnston, line 1: If this is a bug fix it should probably be a `.bugfix`? I could go either way on it though

erikjohnston (approved):

changelog.d/9229.misc
anoadragon453: Yeah, it should probably be a bugfix, especially if the changelog starts that way :blush:

PR #9231, opened 2021-01-26
Skip newsfile check in CI if not a pull request

callahad:

### Pull Request Checklist


clokep:
Would it make more sense to do this in the pipeline so the step doesn't even run (instead of having the step run and just return early). I think something like [an `if: build.pull_request.id` conditional](https://buildkite.com/docs/pipelines/conditionals)?

callahad:
Probably; will PR against pipeline instead 🙂

PR #9232, opened 2021-01-26
Add type hints to E2E handler.

clokep:
This finishes up type hints for the handlers (building on #9223).

The e2e_keys package is quite confusing (especially that it has a lot of nested dicts), but I think I've gotten them all.

This should be reviewable commit-by-commit, but each commit is essentially converting a file so 🤷 . I ended up adding some `JsonDict`-ness to the device handler too since mypy was complaining about something...


synapse/handlers/e2e_keys.py
```diff
@@ -667,16 +686,18 @@ async def upload_signing_keys_for_user(self, user_id, keys):
...
+    ) -> JsonDict:
         """Upload device signatures for cross-signing
 
         Args:
-            user_id (string): the user uploading the signatures
-            signatures (dict[string, dict[string, dict]]): map of users to
+            user_id: the user uploading the signatures
+            signatures: map of users to
                 devices to signed keys. This is the submission from the user; an
                 exception will be raised if it is malformed.
```
anoadragon453, line 698: Do you mind reflowing this paragraph?

synapse/handlers/e2e_keys.py
```diff
@@ -1135,15 +1162,17 @@ async def _retrieve_cross_signing_keys_for_remote_user(
...
 
     Args:
-        key (dict): the key data to verify
-        user_id (str): the user whose key is being checked
-        key_type (str): the type of key that the key should be
-        signing_key (VerifyKey): (optional) the signing key that the key should
+        key: the key data to verify
+        user_id: the user whose key is being checked
+        key_type: the type of key that the key should be
+        signing_key: (optional) the signing key that the key should
```
anoadragon453, line 1175: We can probably drop the `(optional)` here.

anoadragon453 (approved):
Heh, I ended up commenting on the `from_user_id` None-ness before seeing your last commit.

Regardless, happy to see the big reduction of lines into one in `mypy.ini` :tada: 

I've noted some very minor QoL changes below.

synapse/handlers/e2e_keys.py
clokep: Of course not. 👍

PR #8870, opened 2020-12-03
Default to the blacklisting reserved IP ranges.

clokep:
This was a change @richvdh and I discussed after #8821 that we should default to blacklisting all private IP ranges instead of asking people to type them in. This allows us to add / remove some in the future if necessary.

Currently if this option is unspecified / empty than no IP ranges are blacklisted.

PR #9240, opened 2021-01-27
Convert blacklisted IPv4 addresses to compatible IPv6 addresses.

clokep:
This automatically converts IPv4 addresses to some obscure IPv6 address ranges:

* IPv6 with embedded IPv4
* IPv4-compatible IPv6
* 6to4 addresses

This expands on some of the conversation in https://github.com/matrix-org/synapse/pull/8870#discussion_r535485437.

~~An alternative approach to this would be to automatically do these conversions for any IPv4 addresses that get blocked at runtime, not just for the default.~~ This approach is now taken.


clokep:
> An alternative approach to this would be to automatically do these conversions for any IPv4 addresses that get blocked at runtime, not just for the default.

From some chatter I think we're going to switch to this instead.

synapse/config/repository.py
```diff
@@ -187,15 +185,13 @@ def read_config(self, config, **kwargs):
...
 
-            self.url_preview_ip_range_blacklist = IPSet(
-                config["url_preview_ip_range_blacklist"]
-            )
-
             # we always blacklist '0.0.0.0' and '::', which are supposed to be
             # unroutable addresses.
-            self.url_preview_ip_range_blacklist.update(["0.0.0.0", "::"])
+            self.url_preview_ip_range_blacklist = generate_ip_set(
+                config["url_preview_ip_range_blacklist"], ["0.0.0.0", "::"]
```
clokep: I wonder if these default IPs should be CIDRs?

synapse/config/server.py
```diff
@@ -294,32 +355,28 @@ def read_config(self, config, **kwargs):
...
         # Attempt to create an IPSet from the given ranges
-        try:
-            self.ip_range_blacklist = IPSet(ip_range_blacklist)
-        except Exception as e:
-            raise ConfigError("Invalid range(s) provided in ip_range_blacklist.") from e
+
         # Always blacklist 0.0.0.0, ::
-        self.ip_range_blacklist.update(["0.0.0.0", "::"])
+        self.ip_range_blacklist = generate_ip_set(
+            ip_range_blacklist, ["0.0.0.0", "::"], config_path=("ip_range_blacklist",)
```
clokep, line 361: I wonder if these default IPs should be CIDRs?

synapse/config/server.py
```diff
@@ -40,6 +41,66 @@
...
+        try:
+            network = IPNetwork(ip)
+        except AddrFormatError as e:
+            raise ConfigError(
+                "Invalid IP range provided: %s." % (ip,), config_path
+            ) from e
+        result.add(network)
+
+        # It is possible that these already exist in the set, but that's OK.
+        if ":" not in str(network):
```
erikjohnston, line 95: *How* is there not a `network.is_ipv4()`???

synapse/config/server.py
erikjohnston: I think so? Probably?

erikjohnston (approved):

synapse/config/server.py
clokep: I think because every IPv4 network is technically an IPv6 network too since all of IPv4 is included in IPv6? 🤷

clokep: So really all of this works around a bug in netaddr that it doesn't understand the different forms of IPv6 addresses really encompass the same networks as the IPv4 ones. 🤷

erikjohnston: :shrug:

synapse/config/server.py
clokep: Updated 0.0.0.0 -> 0.0.0.0/8 which matches the default blacklist.

clokep:
Based on https://en.wikipedia.org/wiki/Reserved_IP_addresses I also added a few more IP ranges for testing networks and IPv6 multicast (which doesn't make sense).

synapse/config/server.py
```diff
@@ -66,6 +129,10 @@
     "fe80::/10",
     # Unique local addresses.
     "fc00::/7",
+    # Testing networks.
+    "2001:db8::/32",
+    # Multicast.
+    "ff00::/8",
```
dkasak, line 135: Let's also add `fec0::/10:` Site Local Addresses (deprecated - RFC 3879), previously mentioned [here](https://github.com/matrix-org/synapse/pull/8870#discussion_r536253356).

```suggestion
    "ff00::/8",
    # Site-local addresses
    "fec0::/10",
```

docs/sample_config.yaml
```diff
@@ -177,6 +178,8 @@ pid_file: DATADIR/homeserver.pid
 #  - '::1/128'
 #  - 'fe80::/10'
 #  - 'fc00::/7'
+#  - '2001:db8::/32'
+#  - 'ff00::/8'
```
dkasak, line 182:
```suggestion
#  - 'ff00::/8'
#  - 'fec0::/10'
```

docs/sample_config.yaml
```diff
@@ -1002,6 +1006,8 @@ media_store_path: "DATADIR/media_store"
...
```
dkasak, line 1010:
```suggestion
#  - 'ff00::/8'
#  - 'fec0::/10'
```

dkasak (changes requested):

clokep:
@dkasak / @erikjohnston Are we happy with the current bit of this PR? I think it is in good shape now.

erikjohnston (approved):
Code looks fine, I think the ip ranges makes sense, but let's wait for @dkasak to confirm

synapse/config/repository.py
```diff
@@ -187,16 +185,17 @@ def read_config(self, config, **kwargs):
...
+                config["url_preview_ip_range_blacklist"],
+                ["0.0.0.0", "::"],
+                config_path=("url_preview_ip_range_blacklist",),
+            )
 
-            self.url_preview_ip_range_whitelist = IPSet(
-                config.get("url_preview_ip_range_whitelist", ())
+            self.url_preview_ip_range_whitelist = generate_ip_set(
+                config.get("url_preview_ip_range_whitelist", ()),
+                config_path=("url_preview_ip_range_blacklist",),
```
dkasak, line 198: Shouldn't this be `"url_preview_ip_range_whitelist"`?

```suggestion
                config_path=("url_preview_ip_range_whitelist",),
```

synapse/config/server.py
```diff
@@ -40,6 +41,66 @@
...
+    """
+    Generate an IPSet from a list of IP addresses or CIDRs.
+
+    Additionally, for each IPv4 network in the list of IP addresses, also
+    includes the corresponding IPv6 networks.
+
+    This includes:
+
+    * IPv4-Compatible IPv6 Address (see RFC 4291, section 2.5.5.1)
+    * IPv4-Mapped IPv6 Address (see RFC 4291, section 2.5.5.1)
```
dkasak, line 73:
```suggestion
    * IPv4-Mapped IPv6 Address (see RFC 4291, section 2.5.5.2)
```

dkasak (changes requested):
Looks good to me too, apart from two minor things.

(do we Approve or Request changes in such scenarios?)

clokep:
> Looks good to me too, apart from two minor things.

Good catches!

> (do we Approve or Request changes in such scenarios?)

My rule of thumb is request changes if you want to look at them again before they get merged; approve if you're OK with the author making the changes and merging.

dkasak (approved):

PR #9242, opened 2021-01-27
Add 'brand' field to MSC2858 response

richvdh:
We've decided to add a 'brand' field to help clients decide how to style the
buttons.

Also, fix up the allowed characters for idp_id, while I'm in the area.


clokep:
Out of curiosity -- how does this differ from having an icon and name? What additional styling would a client add?

richvdh:
> Out of curiosity -- how does this differ from having an icon and name? What additional styling would a client add?

Great question, but it's one I'm going to take to the MSC, because it'll get lost in here.

richvdh:
I added some notes to https://github.com/matrix-org/matrix-doc/blob/t3chguy/msc/multi-sso/proposals/2858-Multiple-SSO-Identity-Providers.md#styling-information-as-an-alternative-to-brand.

erikjohnston (approved):
Code looks good. As per MSC I wonder whether we actually need the brands to be namespaced.

synapse/config/oidc_config.py
```diff
@@ -226,11 +230,22 @@ def generate_config_section(self, config_dir_path, server_name, **kwargs):
...
     "required": ["issuer", "client_id", "client_secret"],
     "properties": {
-        # TODO: fix the maxLength here depending on what MSC2528 decides
-        #   remember that we prefix the ID given here with `oidc-`
-        "idp_id": {"type": "string", "minLength": 1, "maxLength": 128},
+        "idp_id": {
+            "type": "string",
+            "minLength": 1,
+            # MSC2858 allows a maxlen of 255, but we prefix with "oidc-"
+            "maxLength": 251,
```
clokep, line 237: I think this should be 250: `len("oidc-") == 5`

PR #9245, opened 2021-01-27
Support for scraping email addresses from OIDC providers

richvdh:


synapse/config/oidc_config.py
```diff
@@ -143,9 +143,9 @@ def generate_config_section(self, config_dir_path, server_name, **kwargs):
         #
         #           For the default provider, the following settings are available:
         #
-        #             sub: name of the claim containing a unique identifier for the
-        #                 user. Defaults to 'sub', which OpenID Connect compliant
-        #                 providers should provide.
+        #             subject_claim: name of the claim containing a unique identifier
```
clokep, line 146: This was just documented wrong it looks like?

synapse/handlers/oidc_handler.py
```diff
@@ -1213,16 +1210,23 @@ async def map_user_attributes(
...
-                display_name = None
+        emails = []  # type: List[str]
+        email = render_template_field(self._config.email_template)
+        if email:
+            emails.append(email)
 
-        return UserAttributeDict(localpart=localpart, display_name=display_name)
+        return UserAttributeDict(
+            localpart=localpart, display_name=display_name, emails=emails
+        )
```
clokep, line 1229: FWIW I had wonder if we should just return `synapse.handlers.sso.UserAttributes` here instead of having a class that is almost identical. 🤷 I think I had mypy issues with it though and couldn't easily fix it.

No change necessary, just idly thinking. :)

clokep (approved):
Seems good to me. 👍

synapse/config/oidc_config.py
richvdh: yup.

synapse/handlers/oidc_handler.py
richvdh: I had the same thought. Then I realised this was an extension module, so changing the expected return type could break existing implementations, and suddenly I lost enthusiasm...

clokep: I'm pretty sure that was what I did as well. 😟

Issue #3971, opened 2018-09-27
Presence is increasingly heavy


spantaleev:
I'm not sure if #3962 is completely different, but this is something that I've been noticing for a while..

It seems like sending out presence updates whenever my presence changes, would cause 100% CPU usage for a while on my small VPS.

An easy way to trigger it is to use the riot-ios app. Just opening the app and then sending it to the background would cause it to hit `/_matrix/client/r0/presence/{userId}/status` with `PUT` requests (either updating presence to `online` or to `unavailable`.. and after some more inactivity, to `offline` by a Synapse background task, it seems).

Doing that would cause 100% CPU for a while. I imagine Synapse tries to notify many other servers about the presence change. While this (non-important thing) is going on, Synapse would be slow to respond to other requests.

If I just keep alternating between backgrounding and foregrounding the riot-ios app, I can effectively keep my homeserver at 100% CPU.

Normally though, a few seconds after backgrounding the app (which sets my presence as `unavailable`), due to a subsequent foregrounding of the app or due to a `/sync` by another client of mine (on desktop or something), my presence status would change back to `online` and cause the same thing once again.

Maybe a few things could be investigated:

- whether riot-ios should try to set presence as `unavailable` at all, especially given that other clients may be syncing at the same time and telling Synapse I'm `online`..

- even if a given client says `unavailable`, whether the server should accept that, given that other clients (devices) may be syncing and setting another status at the same time

- whether the server should be so quick to accept and propagate a presence status, when said status might change once again some couple of seconds later. `/sync` is usually called by clients with a long-polling timeout of 30 seconds, so there usually may be something that re-sets the presence status after as little as 30 seconds. Do federated clients care about sub-30-seconds granularity? Perhaps presence changes can be debounced for 30-60 seconds before actually kicking in

- whether propagating presence to other servers should be so heavy. Perhaps the code can be optimized or deprioritized, such that it won't disturb other server operations

------

Using `use_presence = false` eliminates the problem, at the expense of presence not working.

I do like having presence, but I don't care about it being so accurate and so fast to propagate (at the expense of other operations).

neilisfragile:
Hi @spantaleev thanks for detailed issue. I think you're probably right that we are too keen to send out presence updates to other servers in the room.

I don't have an immediate fix, but one thing you might do is check your extremities per https://github.com/matrix-org/synapse/issues/1760 because the problem is that we can end up doing state res for all your rooms to see who's in there.

spantaleev:
[...]

PR #9251, opened 2021-01-28
use_presence: Default to False

rht:
This is a pre-emptive measure so that a user's first install experience is smoother.

Tracking issue with use_presence's performance problem:
https://github.com/matrix-org/synapse/issues/3971.

### Pull Request Checklist

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -581,6 +581,12 @@ This also requires the optional `lxml` python dependency to be  installed. This
 in turn requires the `libxml2` library to be available - on  Debian/Ubuntu this
 means `apt-get install libxml2-dev`, or equivalent for your OS.
 
+### Presence
+
+By default, presence is disabled because because it consumes a lot CPU. If you
+want to turn it on anyway, you can set `use_presence: true`. The tracking issue
+is https://github.com/matrix-org/synapse/issues/3971.
+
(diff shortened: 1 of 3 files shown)


rht:
I will fix the test cases if there is a consensus that presence should be disabled by default.

rht:
Actually, since this default value being set to False is temporary, there is less hassle with just saying in INSTALL.md that `use_presence` should be off.

clokep:
I don't think we can just change the default here since it would cause people using this feature to have it automatically disabled when upgrading?

rht:
Yeah, better if just recommend in INSTALL.md to turn it off for now?

clokep:
> Yeah, better if just recommend in INSTALL.md to turn it off for now?

This is mentioned in the README already: https://github.com/matrix-org/synapse#help-synapse-is-slow-and-eats-all-my-ramcpu

All of this could be a bit better organized intro a troubleshooting section, but I don't think that belongs in the INSTALL file.

I'm going to close this for now since the proposed solution is not backwards compatible.

rht:
I have seen that text in README. Unfortunately it is a remedy after people have noticed their RAM and CPU shot up. It's better to warn people in advance in INSTALL.md.

PR #9255, opened 2021-01-28
Cache config options in SSL verification

erikjohnston:
Reading from the config object is *slow*.

### DIFF
--- a/changelog.d/9238.misc
+++ b/changelog.d/9238.misc
@@ -1,1 +1,0 @@
-Add ratelimited to 3PID `/requestToken` API.
--- a/changelog.d/9238.misc
+++ b/changelog.d/9238.misc
@@ -1,0 +1,1 @@
+Add ratelimited to 3PID `/requestToken` API.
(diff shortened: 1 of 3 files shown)


synapse/crypto/context_factory.py
```diff
@@ -125,19 +125,24 @@ def __init__(self, config):
...
 
+    def get_options(self, host: bytes):
         # IPolicyForHTTPS.get_options takes bytes, but we want to compare
         # against the str whitelist. The hostnames in the whitelist are already
         # IDNA-encoded like the hosts will be here.
         ascii_host = host.decode("ascii")
 
         # Check if certificate verification has been enabled
-        should_verify = self._config.federation_verify_certificates
+        should_verify = self._should_verify
```
clokep, line 141: Looks like this variable is unneeded?

erikjohnston: It gets set and used below?

clokep: Arg, I didn't see that it was set again inside the if-block!

changelog.d/9238.misc
```diff
@@ -0,0 +1 @@
+Add ratelimited to 3PID `/requestToken` API.
```
clokep, line 1: I think this got pulled over from the wrong branch.

clokep (approved):

changelog.d/9238.misc
erikjohnston: ETOOMANYBRANCHES

erikjohnston: How did that event happen :/

Issue #9256, opened 2021-01-28
KeyError in synapse.push.mailer in get_message_vars


clokep:
See https://sentry.matrix.org/sentry/synapse-matrixorg/issues/190798/

It looks like this is happening when a user is no longer in the room when prepping to send an email notification (this is likely due to the user leaving during the delay between an event being received and sending the email).

For some situations we should likely just ignore the notification and move on, but in some situations we probably want to notify the user that they've been kicked / banned / whatever.

PR #9257, opened 2021-01-28
Handle empty rooms when generating email notifications.

clokep:
~~This got a bit out of hand for how I started, I can split this into separate PRs if we would like.~~ The overall goal is to fix the sentry error from #9256 where a member is no longer in the room state (for whatever reason) when generating the push notification.

This can happen in two spots which are handled separately:
1. `get_message_vars`, where the state is used to get the member's name and avatar URL, we can provide reasonable defaults by using the MXID and no avatar.
2. `make_summary_text`, where we can provide slightly more generic summaries.

~~While working on this I ended up refactoring the `make_summary_text` method, then fixing a bug in it. I can split this out to a separate PR if that is clearer.~~ This was split out into #9260.

### DIFF
--- a/changelog.d/9257.bugfix
+++ b/changelog.d/9257.bugfix
@@ -1,0 +1,1 @@
+Fix long-standing KeyError in synapse.push.mailer in get_message_vars.
--- a/changelog.d/9257.bugfix
+++ b/changelog.d/9257.bugfix
@@ -1,1 +1,1 @@
-Fix long-standing KeyError in synapse.push.mailer in get_message_vars.
+Fix long-standing bug where sending email push would fail for rooms that the server had since left.
--- a/synapse/push/mailer.py
+++ b/synapse/push/mailer.py
@@ -257,7 +257,7 @@ class Mailer:
         rooms = []
 
         for r in rooms_in_order:
-            roomvars = await self.get_room_vars(
+            roomvars = await self._get_room_vars(
                 r, user_id, notifs_by_room[r], notif_events, state_by_room[r]
             )
             rooms.append(roomvars)
@@ -273,7 +273,7 @@ class Mailer:
             # Only one room has new stuff
             room_id = list(notifs_by_room.keys())[0]
 
-            summary_text = await self.make_summary_text_single_room(
+            summary_text = await self._make_summary_text_single_room(
                 room_id,
                 notifs_by_room[room_id],
                 state_by_room[room_id],
@@ -281,13 +281,13 @@ class Mailer:
                 user_id,
             )
         else:
-            summary_text = await self.make_summary_text(
+            summary_text = await self._make_summary_text(
                 notifs_by_room, state_by_room, notif_events, reason
             )
 
         template_vars = {
             "user_display_name": user_display_name,
-            "unsubscribe_link": self.make_unsubscribe_link(
+            "unsubscribe_link": self._make_unsubscribe_link(
                 user_id, app_id, email_address
             ),
             "summary_text": summary_text,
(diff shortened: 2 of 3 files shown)


changelog.d/9257.bugfix
```diff
@@ -0,0 +1 @@
+Fix long-standing KeyError in synapse.push.mailer in get_message_vars.
```
erikjohnston, line 1:
```suggestion
Fix long-standing bug where sending email push would fail for rooms that the server had since left.
```

Maybe?

synapse/push/mailer.py
```diff
@@ -443,10 +443,15 @@ async def get_message_vars(
...
-        sender_avatar_url = sender_state_event.content.get("avatar_url")
+        sender_state_event_id = room_state_ids.get(("m.room.member", event.sender))
+        if sender_state_event_id:
+            sender_state_event = await self.store.get_event(sender_state_event_id)
+            sender_name = name_from_member_event(sender_state_event)
+            sender_avatar_url = sender_state_event.content.get("avatar_url")
+        else:
+            # The sender is no longer in the room for some reason.
+            sender_name = event.sender
+            sender_avatar_url = None
```
erikjohnston, line 454: I wonder if we should use the state at the time of the event? Using the MXID is quite ugly so we should probably avoid it where possible.

I think we can just do:

```python
state = await self.state_store.get_state_for_event(StateFilter.from_types((("m.room.member", event.sender),))
````

Falling back to the sender MXID if that is also empty?

And then ditto below

**Review by erikjohnston — APPROVED** [review]

**Review by erikjohnston — CHANGES_REQUESTED** [review]
That wasn't meant to be a green tick :(

File: synapse/push/mailer.py (hunk 2 above)
Comment (clokep) [reply synapse/push/mailer.py:454]: I don't think this works though since (in this case) you don't have an event ID to pass to `get_state_for_event`. Looking through the state store I don't see anything that lets you not pass in an event ID.

Comment (erikjohnston) [reply synapse/push/mailer.py:454]: Oh, I unhelpfully forgot to put the event ID in the example. I mean using `event.event_id` to get the state at the time of the event that we're pushing? Or is `event` something else?

Comment (clokep) [reply synapse/push/mailer.py:468]: That worked! 🎉 

I'm having a bit of trouble figure out how to do this below though since we seem to lose the "event" at some point. (This code is quite confusing...)

**Review by erikjohnston — APPROVED** [review]
`StateFilter.from_types` takes a collection of tuples of type/state key

File: synapse/push/mailer.py [hunk 3]
```diff
@@ -431,22 +470,51 @@ async def get_notif_vars(
...
+        state_key = ("m.room.member", event.sender)
+        sender_state_event_id = room_state_ids.get(state_key)
+        if sender_state_event_id:
+            sender_state_event = await self.store.get_event(
+                sender_state_event_id
+            )  # type: Optional[EventBase]
+        else:
+            # Attempt to check the historical state for the room.
+            historical_state = await self.state_store.get_state_for_event(
+                event.event_id, StateFilter.from_types((state_key,))
```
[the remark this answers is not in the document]
clokep: `state_key` is defined above as `("m.room.member", event.sender)` -- is this incorrect? Maybe the name of the variable is confusing?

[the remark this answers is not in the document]
erikjohnston: OH, yes, the name is very confusing as `state_key` is the name of the field in the event and the second item in the tuple. Maybe something like `state_tuple` or `type_state_key` or something?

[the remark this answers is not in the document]
clokep: I renamed it to `type_state_key` in both places!

erikjohnston:
:shipit:

PR #9258, opened 2021-01-28
Ratelimit invites by room and target user

erikjohnston:


synapse/config/ratelimiting.py
```diff
@@ -107,6 +107,15 @@ def read_config(self, config, **kwargs):
...
         )
 
+        self.rc_invites_per_room = RateLimitConfig(
+            config.get("rc_invites", {}).get("per_room", {}),
+            defaults={"per_second": 0.003, "burst_count": 10},
+        )
+        self.rc_invites_per_user = RateLimitConfig(
+            config.get("rc_invites", {}).get("per_user", {}),
+            defaults={"per_second": 0.003, "burst_count": 5},
+        )
```
anoadragon453, line 117: Should these be `0.03` instead of `0.003`?

synapse/handlers/room_member.py
```diff
@@ -144,6 +155,12 @@ async def _user_left_room(self, target: UserID, room_id: str) -> None:
         """
         raise NotImplementedError()
 
+    def ratelimit_invite(self, room_id: str, invitee_user_id: str):
+        """Ratelimit invites py room and by target user.
```
anoadragon453, line 159:
```suggestion
        """Ratelimit invites by room and by target user.
```

tests/handlers/test_federation.py
```diff
@@ -191,6 +191,97 @@ def test_rejected_state_event_state(self):
...
+                    room_version,
+                )
+            )
+
+        self.get_failure(
+            self.handler.on_invite_request(
+                other_server, create_invite_for("@user-4:test"), room_version,
+            ),
+            exc=LimitExceededError,
+        )
```
anoadragon453, line 239: So my local server would send a 429 back over federation to the remote server? I'm guessing this would just result in a failure on the remote side (possibly without a useful message)?

changelog.d/9258.feature
```diff
@@ -0,0 +1 @@
+Add ratelimites to invites in rooms and to specific users.
```
anoadragon453, line 1:
```suggestion
Add ratelimits to invites in rooms and to specific users.
```

synapse/handlers/room_member.py
```diff
@@ -387,8 +404,11 @@ async def update_membership_locked(
                 raise SynapseError(403, "This room has been blocked on this server")
 
         if effective_membership_state == Membership.INVITE:
+            target_id = target.to_string()
+            self.ratelimit_invite(room_id, target_id)
```
anoadragon453, line 408: I wonder whether we should condition this on the `ratelimit` argument to this function. Unfortunately doing so would allow users to mass invite by supplying a long list of users to `/createRoom`. However not doing so might cause the server to get rate-limited accidentally?

synapse/handlers/room_member.py
dkasak, line 159:
```suggestion
        """Ratelimit invites by room and by target user.
```

anoadragon453 (changes requested):
A few edge cases to think about, but on the whole looks good.

tests/handlers/test_federation.py
erikjohnston: I believe the 429 will be proxied through to the client.

synapse/handlers/room_member.py
erikjohnston: Oh good point. I can also limit the number of invites in `/createRoom` by the burst count.

synapse/config/ratelimiting.py
```diff
@@ -107,6 +107,15 @@ def read_config(self, config, **kwargs):
             defaults={"per_second": 0.003, "burst_count": 5},
         )
 
+        self.rc_invites_per_room = RateLimitConfig(
+            config.get("rc_invites", {}).get("per_room", {}),
+            defaults={"per_second": 0.3, "burst_count": 10},
+        )
+        self.rc_invites_per_user = RateLimitConfig(
+            config.get("rc_invites", {}).get("per_user", {}),
+            defaults={"per_second": 0.003, "burst_count": 5},
```
anoadragon453, line 116: This doesn't match the sample config.

synapse/config/ratelimiting.py
```diff
@@ -174,6 +185,14 @@ def generate_config_section(self, **kwargs):
         #rc_3pid_validation:
         #  per_second: 0.003
         #  burst_count: 5
+        #
+        #rc_invites:
+        #  per_room:
+        #    per_second: 0.3
+        #    burst_count: 10
+        #  per_user:
+        #    per_second: 0.03
```
anoadragon453, line 194: Did you mean for these two `per_second`s to be different?

tests/handlers/test_federation.py
anoadragon453: Looks like a `HttpResponseException` with code 429 will get raised:

https://github.com/matrix-org/synapse/blob/2649d545a551dd126d73d34a6e3172916ea483e0/synapse/federation/federation_client.py#L783-L805

So it won't be a `LimitExceededError` and won't have `retry_after_ms` nor `M_LIMIT_EXCEEDED` but it miiiight be handled correctly by clients?

We may just want to catch `429`, craft and raise a `LimitExceededError` to be extra safe.

anoadragon453 (changes requested):
Getting there!

synapse/config/ratelimiting.py
erikjohnston: SIGH I THOUGHT I DOUBLE CHECKED

synapse/config/ratelimiting.py
erikjohnston: Yes, so we want a relatively high `per_room` ratelimit to allow room upgrades, but we can limit the number of invites a user receives a lot (one every 5 minutes for `0.003`)

tests/handlers/test_federation.py
erikjohnston: Oh thank you. Have fixed this now

synapse/federation/federation_client.py
```diff
@@ -810,7 +810,7 @@ async def _do_send_invite(
                         "User's homeserver does not support this room version",
                         Codes.UNSUPPORTED_ROOM_VERSION,
                     )
-            elif e.code == 403:
+            elif e.code in (403, 429):
                 raise e.to_synapse_error()
```
anoadragon453, line 814: Cool, thanks for added. This won't pass through `retry_after_ms` but I presume the client will have some default amount of time they'll wait for.

anoadragon453 (approved):
lgtm!

synapse/federation/federation_client.py
erikjohnston: Oh, hmm, why on earth don't we proxy through all the fields??

I don't think clients actually care about `retry_after_ms`...

docs/sample_config.yaml
```diff
@@ -862,6 +864,14 @@ log_config: "CONFDIR/SERVERNAME.log.config"
...
 #  per_second: 0.003
 #  burst_count: 5
+#
+#rc_invites:
+#  per_room:
+#    per_second: 0.3
+#    burst_count: 10
+#  per_user:
+#    per_second: 0.003
+#    burst_count: 5
```
richvdh, line 874: if this is a ratelimit per recipient, does that mean that a spammer can DoS a given recipient so that they can't receive any invites from genuine users any more? what happens to genuine invites - are they just lost somehow?

richvdh: (could this sort of info be added to the comments in the config file, please?)

erikjohnston: Hmm, true. This does mean In some ways this is somewhat by design, the idea is that you don't want to have users inundated by invites from a lot of different fake accounts. The invites will see the 429 though, so at least its a visible thing. I guess we could ratelimit by `(sending_server, target_user)` for inbound remote invites so that a spammer would have to have accounts on different servers?

richvdh: maybe. Honestly, I'm kinda questioning how useful this limit is as it stands.

PR #9259, opened 2021-01-28
Define a new Synapse module that can manipulate user directory search results

anoadragon453:
This PR creates a new Synapse module type that is intended to interact with the user directory. In its current state, the only method that can be overridden is  `get_search_query_ordering` which, given a database type, will return SQL that defines how user directory search results should be ordered.

The motivation for this is that DINUM would like to be able to return users with a specific string in their displayname higher than other results. It could potentially be used by others to manipulate search results in whichever way they wish, though admittedly in a limited way currently. There is potential to expand this module to manipulate other areas of the user directory search in the future though.

DINUM's instance of the module is located [here](https://github.com/matrix-org/matrix-synapse-user-directory-search-dinum), and should be considered during review.

Note: I'm aware that having a module pass direct SQL to Synapse is a bit dangerous, however so are the abilities of many other modules. Modules aren't meant to be installed willy-nilly.

I'm also aware that it's a bit gross to request *part* of an SQL query from a module. I didn't want to allow the module to manipulate the return values from the query however, as that would require a much more generic method of processing the results of the query.

Reviewable commit-by-commit.

### DIFF
--- a/changelog.d/9259.feature
+++ b/changelog.d/9259.feature
@@ -1,0 +1,1 @@
+Add a new Synapse module type for manipulating user directory search results.
--- a/docs/sample_config.yaml
+++ b/docs/sample_config.yaml
@@ -2545,6 +2545,21 @@ spam_checker:
 #user_directory:
 #  enabled: true
 #  search_all_users: false
+#
+#  # Server admins can define a Python module that implements extra rules for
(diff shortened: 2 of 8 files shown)


anoadragon453:
This PR has been superseded by https://github.com/matrix-org/synapse/pull/9383.

PR #9260, opened 2021-01-28
Refactor email summary generation.

clokep:
Split out of #9257.

This is reviewable commit-by-commit. For the first commit you likely want to disable whitespace.

It has two changes:

* The second commit fixes a case where no summary text was returned.
* The use of `messages_from_person` vs. `messages_from_person_and_others` was tweaked slightly to be based on whether there was 1 sender or multiple senders, not based on if there was 1 room or multiple rooms.

### DIFF
--- a/changelog.d/9260.misc
+++ b/changelog.d/9260.misc
@@ -1,0 +1,1 @@
+Refactor the generation of summary text for email notifications.
--- a/synapse/push/mailer.py
+++ b/synapse/push/mailer.py
@@ -272,7 +272,11 @@ class Mailer:
             room_id = list(notifs_by_room.keys())[0]
 
             summary_text = await self.make_summary_text_single_room(
-                notifs_by_room[room_id], state_by_room[room_id], notif_events, user_id
+                room_id,
+                notifs_by_room[room_id],
+                state_by_room[room_id],
+                notif_events,
+                user_id,
             )
         else:
             summary_text = await self.make_summary_text(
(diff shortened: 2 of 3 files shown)


erikjohnston:
> This changes the behaviour of the summary returned to be based on the number of senders, instead of based on the number of rooms.

Is that what we want? I'm failing to grok really what that means, as I'd have naively thought it'd be weird to have message from one sender, but in different rooms, look like they were from the same room?

clokep:
> > This changes the behaviour of the summary returned to be based on the number of senders, instead of based on the number of rooms.
> 
> Is that what we want? I'm failing to grok really what that means, as I'd have naively thought it'd be weird to have message from one sender, but in different rooms, look like they were from the same room?

I don't know, but I think it previously didn't make sense.

In the case of a single room with multiple notifications and no room name we used (call this Case A):

https://github.com/matrix-org/synapse/blob/fa50e4bf4ddcb8e98d44700513a28c490f80f02b/synapse/config/emailconfig.py#L37

In the case of multiple rooms and the room causing the notification to be unnamed we used (call this Case B):

https://github.com/matrix-org/synapse/blob/fa50e4bf4ddcb8e98d44700513a28c490f80f02b/synapse/config/emailconfig.py#L40

In both cases the `person` is from `descriptor_from_member_events`:

https://github.com/matrix-org/synapse/blob/6f7417c3db54c9545e93b0428303f29973468d39/synapse/push/presentable_names.py#L168-L193

So for Case A this was generally OK, but for Case B this could end up with:

> You have messages on FooApp from Bob and 3 others and others...

So the idea is that for both of them we'll use the number of senders, which will end up with reasonable messages, I hope! This change is all in https://github.com/matrix-org/synapse/pull/9260/commits/6f33c1eb255039445ccd2de0fe3eaa07cbf71134 I think.

erikjohnston:
Oh, OK. For some reason I was blanking that this was just for the summary, rather than the full email...

erikjohnston (approved):

clokep:
Yeah, this is only for the summaries! And the changed cases are only in the case of the room not having a name, which I don't think is common? 🤷

PR #9262, opened 2021-01-28
Split out a separate endpoint to complete SSO registration

richvdh:
There are going to be a couple of paths to get to the final step of SSO reg, and I want the URL in the browser to consistent. So, let's move the final step onto a separate path, which we redirect to.

There are a couple of preparatory commits here, which should be reviewed separately.

### DIFF
--- a/changelog.d/9262.feature
+++ b/changelog.d/9262.feature
@@ -1,0 +1,1 @@
+Improve the user experience of setting up an account via single-sign on.
--- a/synapse/rest/synapse/client/sso_register.py
+++ b/synapse/rest/synapse/client/sso_register.py
@@ -1,0 +1,50 @@
+# -*- coding: utf-8 -*-
+# Copyright 2021 The Matrix.org Foundation C.I.C.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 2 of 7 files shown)


changelog.d/9262.feature
```diff
@@ -0,0 +1 @@
+Improve the user experience of setting up an account via single-sign on.
```
richvdh, line 1: obviously, there is more to follow.

synapse/rest/synapse/client/sso_register.py
```diff
@@ -0,0 +1,50 @@
...
+    def __init__(self, hs: "HomeServer"):
+        super().__init__()
+        self._sso_handler = hs.get_sso_handler()
+
+    async def _async_render_GET(self, request: Request) -> None:
+        try:
+            session_id = get_username_mapping_session_cookie_from_request(request)
+        except SynapseError as e:
+            logger.warning("Error fetching session cookie: %s", e)
+            self._sso_handler.render_error(request, "bad_session", e.msg, code=e.code)
```
anoadragon453, line 48: I don't think this function takes an arg `code`.

anoadragon453 (approved):
lgtm besides one quirk. Also yay for separate resources.

synapse/rest/synapse/client/sso_register.py
anoadragon453: Nvm, was looking at `synapse-dinsic` :innocent:

Issue #9263, opened 2021-01-29
Trying to set a password during UI Auth while also failing to authenticate during a stage will result in the password not being saved


anoadragon453:
Related: https://github.com/matrix-org/matrix-doc/issues/2907, https://github.com/matrix-org/synapse/pull/8009

CC: @clokep 

DINUM were finding after upgrading to Synapse v1.23.0 that email registration was no longer working due to "Missing params: password". Their flow consisted of the following requests.

Check with the homeserver what the required flows to register are:
```
POST /register

{}

# get back a sessionID
```

Request a validation token from Synapse (no Sydent involved here):
```
POST /register/email/requestToken

{some client_secret and email address info}

# get back a sid (session ID for validating the email)
```

Request /register again, this time **with a password** which we want to save. We're doing to do a `m.login.email.identity` here even though we know that it's going to fail. Note that `m.login.email.identity` is the only flow the server gives us. We just want to save the password.

```
POST /register

{
  "auth": {
    "threepid_creds": { threepid_validation stuff that won't succeed because the link in the email hasn't been pressed yet },
    "session": "what_we_got_from_the_first_request",
    "type": "m.login.email.identity"
  },
  "password": "please_save_this_password123"
}

# Get back a 401. Fair, I expected this. But my password_hash has been recorded, right?
```

The client got a 401 one, but not from a `InteractiveAuthIncompleteError` exception, but a `LoginError` coming from here:

https://github.com/matrix-org/synapse/blob/c9c0ad5e204f309f2686dbe250382e481e0f82c2/synapse/handlers/ui_auth/checkers.py#L191-L192

Now unfortunately, the code that stores `password_hash` in `ui_auth_sessions` for subsequent UI Auth requests will only do so if a `InteractiveAuthIncompleteError` is raised, *not* `LoginError`:

https://github.com/matrix-org/synapse/blob/789d9ebad3043b54a7c70cfadb41af7a20ce3877/synapse/rest/client/v2_alpha/register.py#L513-L536

So uh oh. A `password_hash` never got saved. In the case of DINUM, a user now clicks the link in their email and is brought to a *completely separate client* which doesn't know what the password is supposed to be.

That client (having validated the threepid) then makes a final request to `/register`:

```
POST /register

{
  "auth": {
    "threepid_creds": { threepid_validation stuff that won't succeed because the link in the email hasn't been pressed yet }
  }
}

# Gets back a 400! Missing params: password
```

And thus registration fails.

---

[...]

PR #9265, opened 2021-01-29
Prevent email UIA failures from raising a LoginError

anoadragon453:
Context, Fixes: https://github.com/matrix-org/synapse/issues/9263

In the past to fix an issue with old Riots re-requesting threepid validation tokens, we raised a `LoginError` during UIA instead of `InteractiveAuthIncompleteError`. This is now breaking the way Tchap logs in - which isn't standard, but also isn't disallowed by the spec.

An easy fix is just to remove the 4 year old workaround.

### DIFF
--- a/changelog.d/9265.bugfix
+++ b/changelog.d/9265.bugfix
@@ -1,0 +1,1 @@
+Prevent password hashes from getting dropped if a client failed threepid validation during a User Interactive Auth stage.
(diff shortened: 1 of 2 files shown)


changelog.d/9265.bugfix
```diff
@@ -0,0 +1 @@
+Prevent password hashes from getting dropped if a client failed threepid validation during a User Interactive Auth stage.
```
clokep, line 1: Does it make sense to mention that this removes a workaround for Element < *some number*?

clokep: Or at least "for old Riots"?

clokep (approved):
LGTM minus the changelog entry.

changelog.d/9265.bugfix
anoadragon453: Oh, indeed yes. The fix was added in Riot v0.7.4. I'll add that to the changelog.

PR #8227, opened 2020-09-01
Add /_synapse/client to the reverse proxy docs

anoadragon453:
Fixes: https://github.com/matrix-org/synapse/issues/8154

This PR adds a information about forwarding `/_synapse/client` endpoints through your reverse proxy. The first of these endpoints are introduced in https://github.com/matrix-org/synapse/pull/8004.

Question: Do we need to update `workers.md` with this endpoint as well?

Issue #7530, opened 2020-05-19
Potential bug in SAML + worker mode


clokep:
I'm unsure if this will be a problem in reality or is just a potential for issues, but figured I should document it. This is somewhat similar to #6705, but is:
* Specific to SAML.
* Not specific to UI authentication (e.g. it will apply to login/registration as well).

The SAML handler stores state about ongoing SAML requests in memory (see uses of [`_outstanding_requests_dict`](https://github.com/matrix-org/synapse/blob/3c8a57f080a66a3d4d146adf7020c18b397bcf6c/synapse/handlers/saml_handler.py#L76-L77) in the `synapse.handlers.saml_handler.SamlHandler` class).

In worker mode, it is possible for a request to get created and the callback to occur on different workers causing an error about an unrequested SAML response.

I believe the workaround is to ensure that the following endpoints all go to the same worker:
* `/_matrix/client/r0/login/sso/redirect`
* `/_matrix/saml2/authn_response`
* `/_matrix/client/r0/auth/(org.matrix.login.sso|m.login.sso)/fallback/web`

clokep:
I believe this is a duplicate of #7056.

clokep:
Based on the conversation in #7056, it seems like this might be a different cause that ends up in the same error.

babolivier:
@clokep If you're reopening it, could you please rename this issue with something more explicit? "Potential bug in SAML + worker mode" sounds quite hazy.

localguru:
Hi,

I think same problem here. I run 3 generic.worker via nginx `ip_hash` balancing, login SAML only. When saving keys at e.g. at logout I get this error 
```
{"errcode":"M_UNRECOGNIZED","error":"Unrecognized request"}
```
and the client ends up on URL `https://.../_matrix/client/r0/auth/m.login.sso/fallback/web?session=xxxx` .

error from generic.worker log:
```
2020-10-29 16:41:58,485 - synapse.http.server - 76 - INFO - GET-1699 - <XForwardedForRequest at 0x7fb2efc15da0 method='GET' uri='/_matrix/client/r0/auth/m.login.sso/fallback/web?session=xxxx' clientproto='HTTP/1.0' site=18103> SynapseError: 400 - Unrecognized request
[...]

PR #9271, opened 2021-01-30
Fix SSO on workers

### LINKED ISSUE #8966: workers.md gives incorrect advice about various endpoints
https://github.com/matrix-org/synapse/blob/develop/docs/workers.md#synapseappgeneric_worker says:

> Additionally, the following endpoints should be included if Synapse is configured to use SSO...
> ```
> # OpenID Connect requests.
> ^/_matrix/client/(api/v1|r0|unstable)/login/sso/redirect$
> ^/_synapse/oidc/callback$
> 
> # SAML requests.
> ^/_matrix/client/(api/v1|r0|unstable)/login/sso/redirect$
> ^/_matrix/saml2/authn_response$
> 
> # CAS requests.
> ^/_matrix/client/(api/v1|r0|unstable)/login/(cas|sso)/redirect$
> ^/_matrix/client/(api/v1|r0|unstable)/login/cas/ticket$
> ```

However, none of these endpoints are actually enabled on worker processes, so any attempts to send them to a worker will result in a 404.

richvdh:
Fixes #8966.

~~Based on #9262.~~ Commits should be reviewable independently.

### DIFF
--- a/docs/workers.md
+++ b/docs/workers.md
@@ -256,24 +256,27 @@ Additionally, the following endpoints should be included if Synapse is configure
 to use SSO (you only need to include the ones for whichever SSO provider you're
 using):
 
+    # for all SSO providers
+    ^/_matrix/client/(api/v1|r0|unstable)/login/sso/redirect
+    ^/_synapse/client/pick_idp$
+    ^/_synapse/client/pick_username
+    ^/_synapse/client/sso_register$
+
     # OpenID Connect requests.
-    ^/_matrix/client/(api/v1|r0|unstable)/login/sso/redirect$
     ^/_synapse/oidc/callback$
 
     # SAML requests.
-    ^/_matrix/client/(api/v1|r0|unstable)/login/sso/redirect$
     ^/_matrix/saml2/authn_response$
 
     # CAS requests.
-    ^/_matrix/client/(api/v1|r0|unstable)/login/(cas|sso)/redirect$
     ^/_matrix/client/(api/v1|r0|unstable)/login/cas/ticket$
+
+Ensure that all SSO logins go to a single process.
+For multiple workers not handling the SSO endpoints properly, see
+[#7530](https://github.com/matrix-org/synapse/issues/7530).
 
 Note that a HTTP listener with `client` and `federation` resources must be
 configured in the `worker_listeners` option in the worker config.
-
-Ensure that all SSO logins go to a single process (usually the main process). 
-For multiple workers not handling the SSO endpoints properly, see
-[#7530](https://github.com/matrix-org/synapse/issues/7530).
 
 #### Load balancing
 
--- a/docs/workers.md
+++ b/docs/workers.md
@@ -225,7 +225,6 @@ expressions:
     ^/_matrix/client/(api/v1|r0|unstable)/joined_groups$
     ^/_matrix/client/(api/v1|r0|unstable)/publicised_groups$
     ^/_matrix/client/(api/v1|r0|unstable)/publicised_groups/
-    ^/_synapse/client/password_reset/email/submit_token$
 
     # Registration/login requests
     ^/_matrix/client/(api/v1|r0|unstable)/login$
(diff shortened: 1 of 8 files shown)


synapse/storage/databases/main/registration.py
```diff
@@ -443,6 +443,26 @@ def f(txn):
 
         return await self.db_pool.runInteraction("get_users_by_id_case_insensitive", f)
 
+    async def record_user_external_id(
```
richvdh, line 446: this is moving from `RegistrationStore` to `RegistrationWorkerStore`. There's no caching on this table so I think it's safe to move.

docs/workers.md
```diff
@@ -225,7 +225,6 @@ expressions:
     ^/_matrix/client/(api/v1|r0|unstable)/joined_groups$
     ^/_matrix/client/(api/v1|r0|unstable)/publicised_groups$
     ^/_matrix/client/(api/v1|r0|unstable)/publicised_groups/
-    ^/_synapse/client/password_reset/email/submit_token$
```
richvdh, line 228: tangential, but we may as well fix it while we're here. 

This line was added in #8227. AFAICT, it has never worked on a worker.

synapse/rest/synapse/client/__init__.py
```diff
@@ -42,6 +42,18 @@ def build_synapse_client_resource_tree(hs: "HomeServer") -> Mapping[str, Resourc
...
+    # rely on optional dependencies.
+    if hs.config.oidc_enabled:
+        from synapse.rest.oidc import OIDCResource
+
+        resources["/_synapse/oidc"] = OIDCResource(hs)
+
+    if hs.config.saml2_enabled:
+        from synapse.rest.saml2 import SAML2Resource
+
+        resources["/_matrix/saml2"] = SAML2Resource(hs)
```
clokep, line 55: Does it make sense to note that this is under `_matrix` for backwards compat?

docs/workers.md
```diff
@@ -256,25 +256,28 @@ Additionally, the following endpoints should be included if Synapse is configure
...
     # OpenID Connect requests.
-    ^/_matrix/client/(api/v1|r0|unstable)/login/sso/redirect$
     ^/_synapse/oidc/callback$
 
     # SAML requests.
-    ^/_matrix/client/(api/v1|r0|unstable)/login/sso/redirect$
     ^/_matrix/saml2/authn_response$
 
     # CAS requests.
-    ^/_matrix/client/(api/v1|r0|unstable)/login/(cas|sso)/redirect$
```
clokep, line 268: Note that this one was slightly different in that it has `(cas|sso)` in it. Not sure if any clients still use the `cas` path though.

docs/workers.md
```diff
@@ -256,25 +256,28 @@ Additionally, the following endpoints should be included if Synapse is configure
...
     ^/_matrix/client/(api/v1|r0|unstable)/login/cas/ticket$
 
-Note that a HTTP listener with `client` and `federation` resources must be
-configured in the `worker_listeners` option in the worker config.
-
-Ensure that all SSO logins go to a single process (usually the main process). 
+Ensure that all SSO logins go to a single process.
```
clokep, line 273: This is probably true for the UI auth endpoints too?

clokep (approved):
Looks good generally! I left a couple of comments, but nothing major.

synapse/rest/synapse/client/__init__.py
richvdh: sorry, note where? as a comment here?

docs/workers.md
richvdh: yeahh, I'm hoping that's old enough that basically nobody will encounter it.

docs/workers.md
richvdh: I'm not sure it is? afaik nothing is tracked in-memory for UIA (at least for CAS/OIDC; SAML has https://github.com/matrix-org/synapse/issues/7530)

clokep: Right, the username mapping sessions and extra attributes only matter on login, not during UI auth. OK.

synapse/rest/synapse/client/__init__.py
clokep: As a comment, yes. (My thinking being that this is in the `synapse.rest.synapse` package someone might be confused in the future at why we're registering a `_matrix` URL here. 🤷 ) Not a big deal either way though.

richvdh: ok, I added a comment.

PR #9272, opened 2021-01-30
Improve styling and wording of SSO redirect confirm template

richvdh:
~~Based on #9271.~~

### DIFF
--- a/synapse/module_api/__init__.py
+++ b/synapse/module_api/__init__.py
@@ -279,7 +279,11 @@ class ModuleApi:
         )
 
     async def complete_sso_login_async(
-        self, registered_user_id: str, request: SynapseRequest, client_redirect_url: str
+        self,
+        registered_user_id: str,
+        request: SynapseRequest,
+        client_redirect_url: str,
+        new_user: bool = False,
     ):
         """Complete a SSO login by redirecting the user to a page to confirm whether they
         want their access token sent to `client_redirect_url`, or redirect them to that
@@ -291,9 +295,11 @@ class ModuleApi:
             request: The request to respond to.
             client_redirect_url: The URL to which to offer to redirect the user (or to
                 redirect them directly if whitelisted).
+            new_user: set to true to use wording for the consent appropriate to a user
+                who has just registered.
         """
         await self._auth_handler.complete_sso_login(
-            registered_user_id, request, client_redirect_url,
+            registered_user_id, request, client_redirect_url, new_user=new_user
         )
 
     @defer.inlineCallbacks
--- a/synapse/res/templates/sso_redirect_confirm.html
+++ b/synapse/res/templates/sso_redirect_confirm.html
@@ -3,12 +3,34 @@
 <head>
     <meta charset="UTF-8">
     <title>SSO redirect confirmation</title>
+    <meta name="viewport" content="width=device-width, user-scalable=no">
+    <style type="text/css">
+      {% include "sso.css" without context %}
+    </style>
 </head>
     <body>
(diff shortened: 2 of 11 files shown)


synapse/module_api/__init__.py
```diff
@@ -279,7 +279,11 @@ def complete_sso_login(
         )
 
     async def complete_sso_login_async(
-        self, registered_user_id: str, request: SynapseRequest, client_redirect_url: str
+        self,
+        registered_user_id: str,
+        request: SynapseRequest,
+        client_redirect_url: str,
+        new_user: bool = False,
```
clokep, line 286: Does it make sense to add this to the `complete_sso_login` version as well? I'm not sure what even calls these anymore.

synapse/res/templates/sso_redirect_confirm.html
```diff
@@ -3,12 +3,34 @@
...
+    </style>
 </head>
     <body>
-        <p>The application at <span style="font-weight:bold">{{ display_url | e }}</span> is requesting full access to your <span style="font-weight:bold">{{ server_name }}</span> Matrix account.</p>
-        <p>If you don't recognise this address, you should ignore this and close this tab.</p>
-        <p>
-            <a href="{{ redirect_url | e }}">I trust this address</a>
-        </p>
+        <header>
+            <!-- {% if new_user %} -->
```
clokep, line 13: Why putting this in HTML comment tags?

synapse/res/templates/sso_redirect_confirm.html
```diff
@@ -3,12 +3,34 @@
...
+            <h1>Your account is now ready</h1>
+            <p>You've made your account on {{ server_name | e }}.</p>
```
clokep, line 15: With autoescape on I don't think we should be manually escaping these as it is confusing (although it shouldn't cause an error). (Or is #9200 not on the social login branch?)

synapse/res/templates/sso_redirect_confirm.html
```diff
@@ -3,12 +3,34 @@
...
+            <!-- {% else %} -->
+            <h1>Log in</h1>
+            <!-- {% endif %} -->
+            <p>Continue to confirm you trust <strong>{{ display_url | e }}</strong>.</p>
+        </header>
+        <main>
+            <!-- {% if user_profile.avatar_url and user_profile.display_name %} -->
```
clokep, line 22: Should this code account for having only a display name and not an avatar?

synapse/module_api/__init__.py
richvdh: given it's deprecated, I'm not sure there's any value in updating it. Any modules which want to modify the value of `new_user` should call the async version.

synapse/res/templates/sso_redirect_confirm.html
richvdh: to make it it easier to develop outside the template engine, I believe.

synapse/res/templates/sso_redirect_confirm.html
richvdh: > (Or is #9200 not on the social login branch?)

yup, that, I'm afraid.

synapse/res/templates/sso_redirect_confirm.html
richvdh: the design is not to show this section at all if we don't have an avatar (which we never will, currently).

That said, it ought to account for having an avatar and not a display name...

synapse/res/templates/sso_redirect_confirm.html
clokep: I've never seen this pattern before. What makes it easier to develop without these? We don't seem to do this in any other templates. Personally I find it makes my eyes ignore the Jinja logic.

PR #9275, opened 2021-01-30
Replace username picker with a template

richvdh:
There's some prelimiary work here to pull out the construction of a jinja environment to a separate function.

I wanted to load the template at display time rather than load time, so that it's easy to update on the fly. Honestly, I think we should do this with all our templates: the risk of ending up with malformed templates is far outweighed by the improved turnaround time for an admin trying to update them.

~~Based on #9272.~~


clokep:
I think this PR is going to conflict pretty badly with #9200, which doesn't seem to be on the social login branch. 😢

synapse/rest/synapse/client/pick_username.py
```diff
@@ -73,11 +69,40 @@ async def _async_render_GET(self, request: Request):
...
+        idp_id = session.auth_provider_id
+        template_params = {
+            "idp": self._sso_handler.get_identity_providers()[idp_id],
+            "user_attributes": {
+                "display_name": session.display_name,
+                "emails": session.emails,
+            },
+        }
+
+        template = self._jinja_env.get_template("sso_auth_account_details.html")
```
clokep, line 102: Most of our code seems to get the templates "early" so that errors pop up during start-up instead of when a user hits the page. (And they do it in the config code...) This is different...but I'm not sure it matters.

I don't love loading all the templates early and storing them, sure you only render them once, but it means you're storing them in memory always. 🤷

richvdh:
> I think this PR is going to conflict pretty badly with #9200, which doesn't seem to be on the social login branch. cry

yes, I fear so. I'll have to unpick that in future.

synapse/res/templates/sso_auth_account_details.html
```diff
@@ -0,0 +1,115 @@
...
+        <div class="username_input">
+          <label for="field-username">Username</label>
+          <div class="prefix">@</div>
+          <input type="text" name="username" id="field-username" autofocus required pattern="[a-z0-9\-=_\/\.]+">
+          <div class="postfix">:{{ server_name }}</div>
+        </div>
+        <input type="submit" value="Continue" class="primary-button">
+        <!-- {% if user_attributes %} -->
+        <section class="idp-pick-details">
+          <h2><img src="{{idp.idp_icon | mxc_to_http(24, 24)}}"/>Information from {{ idp.idp_name }}</h2>
```
clokep, line 91: In #9200 I tried to standardize the formatting of our Jinja a bit by adding spaces after `{{` and before `}}` and before/after `|`. 🤷

docs/sample_config.yaml
```diff
@@ -1968,6 +1969,35 @@ sso:
...
+    #
+    #         * idp_id: unique identifier for the IdP
+    #         * idp_name: user-facing name for the IdP
+    #         * idp_icon: if specified in the IdP config, an MXC URI for an icon
+    #              for the IdP
+    #         * idp_brand: if specified in the IdP config, a textual identifier
+    #              for the brand of the IdP
+    #
+    #     * user_attributes: an object containing details about the user that
+    #       we received from the IdP. May have the following attributes:
```
clokep, line 1992: The template seems to also handle `avatar_url`, although none of our SSO code gets that right now.

synapse/res/templates/sso_auth_account_details.html
clokep: (Also something something about linting / autoformatting here... Our JS/HTML files are a bit of a mess in this.)

synapse/rest/synapse/client/pick_username.py
richvdh: sorry yes. I mentioned this in the PR description; wdyt?

docs/sample_config.yaml
richvdh: indeed it does. It's "future-proof".

synapse/rest/synapse/client/pick_username.py
clokep: I think that I should read PR descriptions more closely. It sounds like we're in agreement. Sorry about that!

synapse/res/templates/sso_auth_account_details.html
richvdh: oops yes. Updated, I hope.

docs/sample_config.yaml
clokep: Alright! Does it make sense to document it now or "in the future"? 😄

synapse/res/templates/sso_auth_account_details.js
```diff
@@ -0,0 +1,76 @@
...
+        } else if(json.available) {
+            return {available: true};
+        } else {
+            return {message: username + " is not available, please choose another."};
+        }
+    });
+}
+
+function validateUsername(username) {
+    usernameField.setCustomValidity("");
```
clokep, line 38: TIL this is a thing. 👍

clokep (approved):
Seems like an improvement! 👍

docs/sample_config.yaml
richvdh: I don't think we should document it here until it's supported.

Issue #8896, opened 2020-12-08
Integrate T&C consent into SSO registration flow


richvdh:
When a new user registers on the server via SSO, it would be nice to present them with the consent pages, rather than making them do the consent flow afterwards.

richvdh:
fixed by #9276

PR #4682, opened 2019-02-19
Fix registration on workers

erikjohnston:
There turns out to be a bunch of code paths not tested by sytest, including adding threepids and accepting ToS.

We fix this by adding a replication API to run post registration actions. We run this as a separate step, rather than as part of the existing replication API, mainly to reduce the amount of refactoring that would otherwise need to happen.

c.f. #4666

PR #9276, opened 2021-01-30
Collect terms consent from the user during SSO registration

richvdh:
kinda does what it says on the tin I think?

some awfulness to make the consent registration work on workers via the existing replication APIs. We should clean this up.

~~Based on #9271 and #9275.~~

Fixes: #8896

### DIFF
--- a/synapse/handlers/sso.py
+++ b/synapse/handlers/sso.py
@@ -144,6 +144,7 @@ class UsernameMappingSession:
 
     # choices made by the user
     chosen_localpart = attr.ib(type=Optional[str], default=None)
+    terms_accepted_version = attr.ib(type=Optional[str], default=None)
 
 
 # the HTTP cookie used to track the mapping session id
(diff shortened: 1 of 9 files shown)


synapse/handlers/sso.py
```diff
@@ -816,6 +851,15 @@ async def register_sso_user(self, request: Request, session_id: str) -> None:
...
         )
 
+        auth_result = {}
+        if session.terms_accepted_version:
+            # TODO: make this less awful.
+            auth_result[LoginType.TERMS] = True
+
+        await self._registration_handler.post_registration_actions(
+            user_id, auth_result, access_token=None
+        )
```
erikjohnston, line 861: Was this just not getting called before?

erikjohnston (approved):

synapse/handlers/sso.py
richvdh: indeed it wasn't. It didn't matter, because the only thing it does other than register consent is registering 3pids, which we do in `_register_mapped_user` anyway.

doing it this way feels horrible, because I'm basically having to pretend that we're doing a UIA registration here, but I don't want to get into refactoring all that right now. I don't really understand why `post_registration_actions` is separate from `register_user` anyway. (It seems to have arrived in #4682: any thoughts?)

erikjohnston: I think I probably did it like that because there was some logic/processing that could be done on the workers between the two calls, or something. I can't really remember if I'm honest.

I ha

PR #9277, opened 2021-01-30
Make importing display name and email optional

richvdh:
~~based on #9275~~

### DIFF
--- a/synapse/handlers/sso.py
+++ b/synapse/handlers/sso.py
@@ -14,7 +14,16 @@
 # limitations under the License.
 import abc
 import logging
-from typing import TYPE_CHECKING, Awaitable, Callable, Dict, List, Mapping, Optional
+from typing import (
+    TYPE_CHECKING,
+    Awaitable,
+    Callable,
+    Dict,
+    Iterable,
+    List,
+    Mapping,
+    Optional,
+)
 from urllib.parse import urlencode
 
 import attr
@@ -144,6 +153,8 @@ class UsernameMappingSession:
 
     # choices made by the user
     chosen_localpart = attr.ib(type=Optional[str], default=None)
+    use_display_name = attr.ib(type=bool, default=True)
+    emails_to_use = attr.ib(type=List[str], factory=list)
 
 
 # the HTTP cookie used to track the mapping session id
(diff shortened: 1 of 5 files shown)


synapse/rest/synapse/client/pick_username.py
```diff
@@ -113,11 +113,19 @@ async def _async_render_POST(self, request: SynapseRequest):
 
         try:
             localpart = parse_string(request, "username", required=True)
+            use_display_name = parse_boolean(request, "use_display_name", default=False)
+
+            try:
+                emails_to_use = [
+                    val.decode("utf-8") for val in request.args.get(b"use_email", [])
+                ]  # type: List[str]
```
clokep, line 121: Does this list need to be checked against the list that was given to the user? (Right now it seems like it could be used to inject *any* email?)

richvdh: that is (hopefully) checked in the handler.

richvdh: specifically, here: https://github.com/matrix-org/synapse/pull/9277/commits/f96c0dec56670885428823b7c6ee225ca10cd609#diff-05be763854c3b4a72fd8c85ffb5a6b3a10da786f46c53ca4b64d1f150d8fc992R744

clokep: Ah, OK! Just done in a different layer than expected!

clokep: Arg, #6739 adds a `parse_list_from_args` which would be useful here.

synapse/handlers/sso.py
```diff
@@ -720,11 +736,26 @@ async def handle_submit_username_request(
...
         """
         session = self.get_mapping_session(session_id)
 
         # update the session with the user's choices
         session.chosen_localpart = localpart
+        session.use_display_name = use_display_name
+
+        session.emails_to_use = []
+        for email in emails_to_use:
+            # this is O(N^2), but N is small...
```
clokep, line 750: This makes me wonder if they should both be sets (hopefully none of our code explodes if there's duplicates given!).

clokep (approved):

synapse/rest/synapse/client/pick_username.py
richvdh: that feels a bit overengineered, if I'm honest.

synapse/handlers/sso.py
richvdh: hrm, duplicates are a thing I hadn't considered. I think I will make them sets.

clokep:
Changes look reasonable FTR. 👍 ✅

PR #9280, opened 2021-01-31
Remove tls_fingerprints option

### LINKED ISSUE #8424: rip out tls_fingerprints option from homeserver.yaml
this option has no use any more and we should get rid of it and any code that reads it.

jerinjtitus:
Signed-off-by: Jerin J Titus <72017981+jerinjtitus@users.noreply.github.com>

### Pull Request Checklist

Fixes #8424.


anoadragon453:
The diff in this PR looked to be a little mixed to a bad merge or similar. I've removed everything besides the intended changes.

clokep:
This has some unused imports:

```
synapse/config/tls.py:20:1: F401 'hashlib.sha256' imported but unused
synapse/config/tls.py:23:1: F401 'unpaddedbase64.encode_base64' imported but unused
scripts-dev/convert_server_keys.py:1:1: F401 'hashlib' imported but unused
```

clokep:
I wonder if we should raise an error if this value is added to the config? I don't know our process for removing values from the config. 😢

richvdh:
>  I don't know our process for removing values from the config. 

IIRC there are a number of removed config options that make synapse print/log errors if you try to use them, so that's an option.

In this case I don't think setting `tls_fingerprints` is actively harmful, so I don't think there's any need to error out.

changelog.d/9280.misc
```diff
@@ -0,0 +1 @@
+Ripped out an option from `homeserver.yaml`.
```
clokep, line 1:
```suggestion
Removed support for the deprecated `tls_fingerprints` configuration setting.
```

You can add `Contributed by <your name>.` to the end if you would like!

This should also be named `changelog.d/9280.removal`, I think.

clokep (commented):
I think this looks good overall. There seems to be a reference in `synapse.app._base` to the `tls_fingerprints` which shows that the `read_certificate_from_disk` method can be further simplified...I think. 😄

changelog.d/9280.misc
jerinjtitus: Thanks for the review and sure adding name won't hurt.

jerinjtitus:
> I think this looks good overall. There seems to be a reference in `synapse.app._base` to the `tls_fingerprints` which shows that the `read_certificate_from_disk` method can be further simplified...I think.

Yes, I had that doubt. It would be great if you can give some insight into it's working, so that I could simplify it. The comment 
`# we only need the certificate for the tls_fingerprints.`
made me think that the whole function's purpose was around `tls_fingerprints` option, but not sure how exactly.

clokep:
> > I think this looks good overall. There seems to be a reference in `synapse.app._base` to the `tls_fingerprints` which shows that the `read_certificate_from_disk` method can be further simplified...I think.
> 
> Yes, I had that doubt. It would be great if you can give some insight into it's working, so that I could simplify it. The comment
> `# we only need the certificate for the tls_fingerprints.`
> made me think that the whole function's purpose was around `tls_fingerprints` option, but not sure how exactly.

In the case where there's no TLS listeners the only reason that access to the certificate is needed is to add it to the `tls_fingerprints` option, so if we're not loading that anymore.

If we take a look at `read_certificate_from_disk` with your changes and consider `require_cert_and_key` to be `False`, then the only thing it does is (maybe) set `self.tls_certificate`. So I think the call to it from `_base` with `require_cert_and_key=False` can go away and then we can simplify the code to always load the certificates.

Looking at it a bit more I'm actually unsure that `tls_certificate` is even necessary at all after this change?

jerinjtitus:
>Looking at it a bit more I'm actually unsure that tls_certificate is even necessary at all after this change?

I am doubtful about that too. Should we file a new issue on it?

clokep:
> > Looking at it a bit more I'm actually unsure that tls_certificate is even necessary at all after this change?
> 
> I am doubtful about that too. Should we file a new issue on it?

I see no reason not to remove it as part of this.

jerinjtitus:
> I see no reason not to remove it as part of this.
I suggested it because there was an ambiguity on if it's necessary or not. If we are sure it's not, then I can start working on it.

clokep:
> > I see no reason not to remove it as part of this.
> 
> I suggested it because there was an ambiguity on if it's necessary or not. If we are sure it's not, then I can start working on it.

I didn't see it used, but it is worth double checking! We can also remove it and see if anything breaks. 🤷

clokep:
I merged develop into this since it is quite out of date.

richvdh:
we should decide if this is ready to merge, or if it should be closed.

erikjohnston:
I think it makes sense to merge? No server since v1.0 should be looking at TLS fingerprints right?

(I added need discussion to make sure we actually make a decision sometime this cycle)

richvdh:
> I think it makes sense to merge? No server since v1.0 should be looking at TLS fingerprints right?

yes, we can certainly remove the `tls_fingerprints` value from the /keys response (if that's not already done).  The question is whether this PR does that - it just needs a review.

clokep:
I think after this PR there's some additional dead code which could be removed. We could do that after the fact though.

scripts-dev/convert_server_keys.py
```diff
@@ -54,15 +53,9 @@ def convert_v1_to_v2(server_name, valid_until, keys, certificate):
         "server_name": server_name,
         "verify_keys": {key_id: {"key": key} for key_id, key in keys.items()},
         "valid_until_ts": valid_until,
-        "tls_fingerprints": [fingerprint(certificate)],
```
richvdh, line 57: I don't think this is necessary, but on the other hand, I don't think this script is useful any more.

synapse/rest/key/v2/local_key_resource.py
```diff
@@ -89,14 +84,11 @@ def response_json_object(self):
...
             }
 
-        tls_fingerprints = self.config.tls_fingerprints
-
         json_object = {
             "valid_until_ts": self.valid_until_ts,
             "server_name": self.config.server_name,
             "verify_keys": verify_keys,
             "old_verify_keys": old_verify_keys,
-            "tls_fingerprints": tls_fingerprints,
```
richvdh, line 99: for reference: this has been unspecced since at least the first version of the server-server API: https://matrix.org/docs/spec/server_server/r0.1.0#get-matrix-key-v2-server-keyid.

richvdh (approved):

PR #9285, opened 2021-02-01
Ensure pushers are deleted for deactivated accounts

erikjohnston:


synapse/handlers/deactivate_account.py
```diff
@@ -120,6 +120,11 @@ async def deactivate_account(
 
         await self.store.user_set_password_hash(user_id, None)
 
+        # Most of the pushers will have been deleted when we logged out the
```
richvdh, line 123: is this true?

changelog.d/9285.bugfix
```diff
@@ -0,0 +1 @@
+Fix bug where we deleted correctly delete all pushers for a user when they deactivated their account.
```
clokep, line 1: I don't think this makes sense.

synapse/handlers/deactivate_account.py
erikjohnston: Ugh, the logic looks to be broken when using workers.

erikjohnston: This has now been fixed

erikjohnston:
I rebased without thinking I'm afraid, but the code hasn't changed.

synapse/storage/databases/main/pusher.py
```diff
@@ -373,3 +373,44 @@ def delete_pusher_txn(txn, stream_id):
...
+            )
+
+            self.db_pool.simple_delete_txn(
+                txn,
+                table="pushers",
+                keyvalues={"user_name": user_id},
+            )
+
+            for stream_id, pusher in zip(stream_ids, pushers):
+                self.db_pool.simple_insert_txn(
```
clokep, line 402: Looks like this could be `simple_insert_many_txn`, but not sure if that's an optimization or not. (Is there any reason *not* to do that?)

clokep (approved):
Looks reasonable.

synapse/storage/databases/main/pusher.py
erikjohnston: Good point, `simple_insert_many` is quicker

PR #9286, opened 2021-02-01
Improve styling and wording of SSO UIA templates

### LINKED ISSUE #9171: SSO-over-UIA flow is ugly and not mobile-friendly
1. It starts with something nice in the client app, like: 

   > image

   That bit is fine.

2. After you click "Single Sign On", you get taken to a new tab, which looks like:

   > image

   That is (a) ugly, (b) not mobile-friendly, (c) we don't seem to know if "sign-on" is hyphenated or not.

   For context, [MSC2454](https://github.com/matrix-org/matrix-doc/blob/master/proposals/2454-ui-interactive-auth-for-sso.md#why-we-need-user-to-confirm-before-the-sso-flow) explains why this page is required.

3. Clicking "reauthenticate with single sign-on" takes you to the SSO provider's pages, which are outside our control and probably fine.

4. Then when you finish SSO you end up back at a page which looks like:

   > image

   ... which is also ugly and not mobile-friendly. It's also missing a full stop. It's worth noting that the intention is that the user should never see this page, but various bugs (https://github.com/vector-im/element-web/issues/16219, https://github.com/matrix-org/matrix-doc/issues/2977) mean that is not the case today.

5. There are also some error pages a user might see along the way. 

    For example, a generic error page:

    > image

   Or, "you seem to have logged into your SSO provider with a different user to what we expected":

   > image

   In both cases we do a poor job of explaining to the user how they should proceed. Oh and now single-sign-on has *two* hyphens!

There's a general theme here that these templates are very basic and would benefit from some wordsmithing and design. *However*, note that they are part of the *homeserver*, not the client, so Element branding is inappropriate. For `matrix.org`, we could consider customising the default templates with Matrix branding, but it would be nice to consider whether we can improve the default templates that are shipped as part of the Synapse package so that all deployments can benefit from the improvements.

richvdh:
fixes #9171

### DIFF
--- a/synapse/res/templates/sso_auth_success.html
+++ b/synapse/res/templates/sso_auth_success.html
@@ -1,18 +1,27 @@
-<html>
-<head>
-    <title>Authentication Successful</title>
-    <script>
-    if (window.onAuthDone) {
-        window.onAuthDone();
-    } else if (window.opener && window.opener.postMessage) {
-        window.opener.postMessage("authDone", "*");
-    }
-    </script>
-</head>
+<!DOCTYPE html>
+<html lang="en">
+    <head>
+        <meta charset="UTF-8">
+        <title>Authentication successful</title>
+        <meta name="viewport" content="width=device-width, user-scalable=no">
+        <style type="text/css">
+            {% include "sso.css" without context %}
+        </style>
+        <script>
+            if (window.onAuthDone) {
+                window.onAuthDone();
+            } else if (window.opener && window.opener.postMessage) {
+                window.opener.postMessage("authDone", "*");
+            }
+        </script>
+    </head>
     <body>
-        <div>
-            <p>Thank you</p>
-            <p>You may now close this window and return to the application</p>
-        </div>
+        <header>
+            <h1>Thank you</h1>
+            <p>
+                Now we know it’s you, you can close this window and return to the
+                application.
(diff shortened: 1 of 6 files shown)


erikjohnston (approved):
Sure! Though looks like some changes are duplicated with #9287?

richvdh:
hrm; I seem to have messed up.

richvdh:
Let's try that again. I've now removed the duplicated commit and added  f9ab7d3.

richvdh:
(sorry about that)

synapse/res/templates/sso_auth_success.html
```diff
@@ -1,18 +1,27 @@
...
+    </head>
     <body>
-        <div>
-            <p>Thank you</p>
-            <p>You may now close this window and return to the application</p>
-        </div>
+        <header>
+            <h1>Thank you</h1>
+            <p>
+                Now we know it’s you, you can close this window and return to the
```
erikjohnston, line 22: Hah, this sounds slightly ominous.

erikjohnston (approved):

synapse/res/templates/sso_auth_success.html
erikjohnston: (I think its fine though)

PR #9288, opened 2021-02-01
Put OIDC callback URI under /_synapse/client.

### LINKED ISSUE #9050: OIDC SSO does not work with documented reverse-proxy settings
per the reverse-proxy docs, https://github.com/matrix-org/synapse/blob/master/docs/reverse_proxy.md, you only need to forward `/_matrix` and `/_synapse/client` to synapse.

However, the OIDC login flow calls back to `/_synapse/oidc/callback`, so people following the documentation get a 404.

richvdh:
Fixes #9050.

Some notes on this for posterity: it's hard to change this without breaking backwards-compatibility, because:
 * each time we redirect to the IdP, we tell it which URI we expect it to redirect back to
 * the IdP has a whitelist, and we've told everyone to configure their IdPs to expect `/_synapse/oidc/callback`.

Still, this seems like a thing that needs fixing, and it's better to rip the plaster off now than wait for even more people to deploy OIDC.


clokep (approved):
Seems reasonable.

The error people will get without making this change will be unable to login, correct?

richvdh:
> The error people will get without making this change will be unable to login, correct?

it'll be an error from their IdP:

image

Aricg:
this was fun. but agreed about ripping the plaster, we are not live yet. lol.

PR #9289, opened 2021-02-01
Put SAML callback URI under /_synapse/client.

richvdh:
Largely for consistency with OIDC, but also just to tidy it up while we're in the area.

Based on #9288.

(Unlike #9288, we can preserve backwards-compat simply by making both the old and the new endpoints work, because in the case of SAML, the redirectUrl is stored at the SAML IdP).


clokep (approved):
I think this is OK, but the diff is a bit hard to read.

clokep:
Should we also do `/login/cas/ticket` while we're at it? I don't think this is part of the C-S spec?

richvdh:
> Should we also do /login/cas/ticket while we're at it? I don't think this is part of the C-S spec?

yeeees, but that's defined as a `Servlet` rather than a `Resource`, which makes it more fiddly to move :)

richvdh:
rebased; hopefully the diff will be clearer.

PR #9281, opened 2021-02-01
Reorganize CONTRIBUTING.md

Yoric:
### Pull Request Checklist

* [X] Pull request is based on the develop branch
* [x] Pull request includes a [changelog file](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md#changelog). The entry should:
  - Be a short description of your change which makes sense to users. "Fixed a bug that prevented receiving messages from other servers." instead of "Moved X method from `EventStore` to `EventWorkerStore`.".
  - Use markdown where necessary, mostly for `code blocks`.
  - End with either a period (.) or an exclamation mark (!).
  - Start with a capital letter.
* [x] Pull request includes a [sign off](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md#sign-off)
* [X] Code style is correct (run the [linters](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md#code-style))

PR #9293, opened 2021-02-02
Add -prof tox environment

ShadowJonathan:
### Pull Request Checklist

### DIFF
--- a/changelog.d/9293.misc
+++ b/changelog.d/9293.misc
@@ -1,0 +1,1 @@
+Add profiling test environment.
--- a/tox.ini
+++ b/tox.ini
@@ -62,6 +62,9 @@ setenv =
 passenv = *
 
 commands =
+    # can be invoked by adding -prof to any environment (e.g. 'py36-postgres-prof'), does not work with '-j X'
+    prof: python -m cProfile -o "{envname}.cprof" "{envbindir}/trial" {env:TRIAL_FLAGS:} {posargs:tests} {env:TOXSUFFIX:}
+
     # the "env" invocation enables coverage checking for sub-processes. This is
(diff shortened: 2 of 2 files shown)


ShadowJonathan:
(I *think* it would also be handy to add this to `CONTRIBUTING.md`, but i have no idea how to go about that when #9281 is also going on)

richvdh (commented):
I'm not really convinced this is worthwhile: it's just extra stuff to maintain, which isn't going to be used very often. I think if you're going to run a profiler, you can run `trial` yourself.

richvdh (changes requested):
.

ShadowJonathan:
How exactly would this add maintenance load? The only thing I can imagine is a gotcha for "oh, so I can just ignore `!prof`", this change will only make it so that *any* test run is able to be profiled (which can be helpful if it's included in the tox config by default), which can then be added to `CONTRIBUTING.md`. Synapse swapped out plainly running `trial` with deterministic `tox`, adding a profiler option to it would only make sense, as every other option has already been folded into `tox`.

I understand if y'all dont wanna accept this PR, but if you really just wanna put a full dot on that decision, just close it.

clokep:
> How exactly would this add maintenance load? The only thing I can imagine is a gotcha for "oh, so I can just ignore `!prof`" when reading `tox.ini`

tox configs are already confusing enough to read and this adds additional cognitive load each time we need to modify it.

> Synapse swapped out plainly running `trial` with deterministic `tox`, adding a profiler option to it would only make sense, as every other option has already been folded into `tox`.

I'm not sure what this means, I think most of the team runs tests directly via trial (and not using tox).

I agree with @richvdh's assessment.

PR #9296, opened 2021-02-02
Fix formatting for "bad session" error during sso registration flow

richvdh:
We seem to have a bit of a mix of error handling approaches: some happens in the `rest` layer, and some in the handler, so this feels a bit messy. I'm not really sure how to improve it though.


clokep (approved):

clokep:
> some happens in the `rest` layer, and some in the handler, so this feels a bit messy. I'm not really sure how to improve it though.

I don't have ideas of how to improve this at the moment either. I've usually tried to do input validation in the REST layer, but you'll still have some potential of throwing errors via the handlers layer. 🤷

PR #9297, opened 2021-02-02
social login Fix username validation javascript

richvdh:
this one from @bwindels could probably do with going into 1.27.0, since it's quite broken in the RC.


clokep (approved):
Looks plausible. 😄 Was it only broken on old browsers or...?

It would be nice to be able to test some of the JS code we have, but that would be a rather large undertaking.

richvdh:
it was pretty broken everywhere :/

Issue #9294, opened 2021-02-02
[umbrella] Adding pg8000 as a database driver


ShadowJonathan:
Using [pg8000](https://pypi.org/project/pg8000/), a pure-python postgres database driver, will make it possible to use pypy in a performant manner, as pypy has poor interaction with it's C-API.

This issue is thus a sub-issue of #8888, and tracks progress in integrating pg8000 into the codebase.

### TODO

- [ ] Pre: Generalize the current database drivers further to minimise engine-specific behaviour (#9298)
  - Adding in `BaseDatabaseEngine.sql_type` property to if-else based on `"sqlite"` or `"postgres"`
- [ ] Pre: Rename `PostgresEngine` to `PsycoPg2Engine` (#XXXX)
- [ ] Add `Pg8000Engine` (#XXXX)
- [ ] Post: Add `"postgres"` "virtual" database name to switch to `psycopg2` on cpython, and `pg8000` on pypy (#XXXX)
- [ ] Post: Add `with engine.new_conn() as conn:`-esc context manager usage everywhere, to ensure proper closure of cursors, transactions, and connections in pypy (#XXXX)

erikjohnston:
> Using [pg8000](https://pypi.org/project/pg8000/), a pure-python postgres database driver, will make it possible to use pypy in a performant manner, as pypy has poor interaction with it's C-API.

I'm slightly surprised that it'd be faster than the other connectors, given its written in python. Do you have benchmarks or blog posts or something to show its happening?

ShadowJonathan:
On pypy, `pg8000` is faster than `psycopg2cffi` when placed next to eachother, due to that poor C-API interaction I mentioned earlier.

This comes from some rough `tox -e py36` vs `tox -e pypy3` test runs i did in #9123:

```
using psycopg2(cffi): (-j 3)
py36-postgres: Ran 1501 tests in 706.022s
pypy3-postgres: Ran 1501 tests in 2548.308s

using pg8000: (-j 3)
pypy3-postgres: Ran 1501 tests in 763.756s
```

I don't have blogposts specifically for pg8000, but it's a matter of principle; pypy in general has a JIT in it's central premise, which speeds up bare python to machine code, and [this blog post](https://morepypy.blogspot.com/2018/09/inside-cpyext-why-emulating-cpython-c.html) talks a bit about why pypy has poor performance with C APIs in general (it mentions CFFI being "better", but it's not on-par with performance it should be able to achieve, it's still not fully optimised).

[...]

PR #9298, opened 2021-02-02
Generalize the current database drivers further to minimise engine-specific behaviour

ShadowJonathan:
### Pull Request Checklist

### DIFF
--- a/changelog.d/9298.misc
+++ b/changelog.d/9298.misc
@@ -1,0 +1,1 @@
+Generalise various internal database components.
--- a/synapse/replication/slave/storage/_base.py
+++ b/synapse/replication/slave/storage/_base.py
@@ -27,7 +27,9 @@ logger = logging.getLogger(__name__)
 class BaseSlavedStore(CacheInvalidationWorkerStore):
     def __init__(self, database: DatabasePool, db_conn, hs):
         super().__init__(database, db_conn, hs)
-        if isinstance(self.database_engine, PostgresEngine):
+        if isinstance(
+            self.database_engine, PostgresEngine
+        ):  # todo see if generalizable to is_postgres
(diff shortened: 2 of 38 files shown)


ShadowJonathan:
(I invite people to look at the added `todo` items and give suggestions wherever possible)

clokep:
My quick thought on this is that I'm not sure that this really generalizes things, the PR could use a more in-depth description of what this is accomplishing and why.

It seems to mostly be swapping out an isinstance check for a property saying whether something is postgres or sqlite. I understand this is related to #9294 -- would it be possible to use the same `PostgresEngine` wrapper for that? If not, then why?

ShadowJonathan:
It wont be possible to use the same wrapper, the `PostgresEngine` isinstance check comes with the assumption that it'll either be the `psycopg2` driver, or the `sqlite3` driver, we need to generalise this to the point of "oh, what SQL dialect do i need to use?" (postgresql or sqlite), some of these `isinstance` checks were left in because the branch started using `psycopg2`-specific logic.

This PR is aimed to make the codebase more intentionally oblivious which exact driver is being used at all times, and removing assumptions it'll then either be a binary of `psycopg2` or `sqlite3` passing through, so it needs to first be generalized to the two SQL dialects coming up (`postgresql` and `sqlite`), with driver-specific logic having fallbacks that don't assume the binary of drivers.

ShadowJonathan:
Question; does `MultiWriterIdGenerator` use `psycopg2`-specific assumptions? e.g. connections being thread-safe and/or `psycopg2`-specific methods/behaviour.

ShadowJonathan:
Closing this PR as the changes need to be less vague and targeted, i'll come back to this later (hopefully).

Issue #9295, opened 2021-02-02
Rework the storage framework to work nicely with PEP 249


ShadowJonathan:
This is a "maybe" issue, but some elements of [PEP 249](https://www.python.org/dev/peps/pep-0249/) have already been implicitly coded into the codebase, such as `synapse.storage.types`, however, it could be easier to directly work with PEP 249's DB-API-v2 schema and approach to simplify the storage framework even further, and make (potential) future database engines easier to support.

(Things such as module [`.threadsafety`](https://www.python.org/dev/peps/pep-0249/#threadsafety) properties can maybe help futher granularize database pooling mechanisms dynamically.)

richvdh:
what would this actually involve, in your mind? The codebase assumes database APIs which comply PEP249, as both psycopg2 and sqlite do.

ShadowJonathan:
More directly work with PEP249 interfaces which's quirks are caught in the database engine modules, instead of embedding the quirks all over the place.

Basically: Only consume `PEP249` APIs, specialise on database type, and the database driver needs to be invisible.

Ofc only if this is possible with the way the framework is set up.

The pros of this will be that it's incredibly easy to work with multiple database engine types, and possibly multiple SQL database types, but that's far-off.

clokep:
I'm still not sure what concrete changes are proposed here -- can you clarify what "storage framework" means to you? (Is it a particular class? A particular module?)

ShadowJonathan:
Ah sorry, it's quite ambiguous, but i mean the target area of every part of synapse that either;
- Holds a reference to a `Connection`, `Cursor` or `Transaction` object (to keep those parts in mind)
- Directly works on aforementioned objects with SQL statements
- Needs to or utilises certain quirks from database drivers and directly or indirectly invokes them by passing values that end up being used by those drivers.

With "storage framework" i meant the entirety of `synapse.storage`, but i have a feeling even that module-based scope might not be good enough, but we'll see.

richvdh:
can you give us some examples of what would change here? As far as I'm aware almost all of the code you mention already uses PEP-249-compliant interfaces.

ShadowJonathan:
[...]

PR #9299, opened 2021-02-02
Detail typing of Connection and Cursor to further match PEP 249

ShadowJonathan:
### Pull Request Checklist

Related to #9295 

### DIFF
--- a/changelog.d/9299.misc
+++ b/changelog.d/9299.misc
@@ -1,1 +1,1 @@
-Update internal database-related types to better match PEP 249.
+Update the `Cursor` type hints to better match PEP 249.
--- a/changelog.d/9299.misc
+++ b/changelog.d/9299.misc
@@ -1,0 +1,1 @@
+Update internal database-related types to better match PEP 249.
--- a/synapse/storage/prepare_database.py
+++ b/synapse/storage/prepare_database.py
@@ -619,9 +619,10 @@ def _get_or_create_schema_state(
 
     txn.execute("SELECT version, upgraded FROM schema_version")
     row = txn.fetchone()
-    current_version = int(row[0]) if row else None
+    current_version = int(row[0]) if row is not None and len(row) > 0 else None
 
     if current_version:
+        assert row is not None
         txn.execute(
             "SELECT file FROM applied_schema_deltas WHERE version >= ?",
             (current_version,),
--- a/synapse/storage/prepare_database.py
+++ b/synapse/storage/prepare_database.py
@@ -619,10 +619,9 @@ def _get_or_create_schema_state(
 
     txn.execute("SELECT version, upgraded FROM schema_version")
     row = txn.fetchone()
-    current_version = int(row[0]) if row else None
 
-    if current_version:
-        assert row is not None
+    if row is not None:
+        current_version = int(row[0])
         txn.execute(
             "SELECT file FROM applied_schema_deltas WHERE version >= ?",
             (current_version,),
--- a/synapse/storage/prepare_database.py
+++ b/synapse/storage/prepare_database.py
@@ -619,7 +619,7 @@ def _get_or_create_schema_state(
 
     txn.execute("SELECT version, upgraded FROM schema_version")
     row = txn.fetchone()
-    current_version = int(row[0]) if row is not None and len(row) > 0 else None
+    current_version = int(row[0]) if row else None
 
     if current_version:
         assert row is not None
--- a/synapse/storage/types.py
+++ b/synapse/storage/types.py
@@ -44,14 +44,16 @@ class Cursor(Protocol):
         self,
     ) -> Optional[
         Sequence[
+            # fixme: this is an approximate typing of dbapi2's cursor.description, going off of sqlite3 and other
+            #  drivers, it may not be entirely accurate.
             Tuple[
-                Any,
-                Any,
+                str,
                 Optional[Any],
-                Optional[Any],
-                Optional[Any],
-                Optional[Any],
-                Optional[Any],
+                Optional[int],
+                Optional[int],
+                Optional[int],
+                Optional[int],
+                Optional[int],
             ]
         ]
     ]:
--- a/synapse/storage/types.py
+++ b/synapse/storage/types.py
@@ -12,7 +12,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-from typing import Any, Iterable, Iterator, List, Optional, Tuple
+from typing import Any, Iterator, List, Mapping, Optional, Sequence, Tuple, Union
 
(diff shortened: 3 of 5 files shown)


synapse/storage/prepare_database.py
```diff
@@ -619,9 +619,10 @@ def _get_or_create_schema_state(
 
     txn.execute("SELECT version, upgraded FROM schema_version")
     row = txn.fetchone()
-    current_version = int(row[0]) if row else None
+    current_version = int(row[0]) if row is not None and len(row) > 0 else None
```
ShadowJonathan, line 622: (better to be explicit)

synapse/storage/prepare_database.py
```diff
@@ -619,9 +619,10 @@ def _get_or_create_schema_state(
...
 
     if current_version:
+        assert row is not None
```
ShadowJonathan, line 625: (needed for mypy to stop talking about `row` being `Optional[]`, is already assured by `current_version` being truthy)

synapse/storage/prepare_database.py
clokep: Please revert this change, it is clearer to treat all false-y values the same.

synapse/storage/types.py
```diff
@@ -12,30 +12,49 @@
...
+    ) -> Optional[
+        Sequence[
+            Tuple[
+                Any,
+                Any,
+                Optional[Any],
+                Optional[Any],
+                Optional[Any],
+                Optional[Any],
+                Optional[Any],
```
clokep, line 54: It would be nice if these had more specific types than `Any` -- it seems the first is a `str` and some of the others are `Optional[int]`?

synapse/storage/types.py
```diff
@@ -12,30 +12,49 @@
...
+            ]
+        ]
+    ]:
         return None
```
clokep, line 58: While we're here?

```suggestion
        ...
```

synapse/storage/util/sequence.py
```diff
@@ -106,7 +106,15 @@ def __init__(self, sequence_name: str):
...
         txn.execute("SELECT nextval(?)", (self._sequence_name,))
-        return txn.fetchone()[0]
+        fetch_res = txn.fetchone()
+        if fetch_res is None:
+            # this should technically not happen, unless the connection/cursor/transaction has been corrupted
+            raise RuntimeError(
+                "nextval for self._sequence_name {} returned empty".format(
+                    repr(self._sequence_name)
+                )
+            )
```
clokep, line 116: Since these cannot happen it probably makes more sense to do `assert fetch_res is not None`. (Same below.)

synapse/storage/types.py
```diff
@@ -12,30 +12,49 @@
...
-from typing import Any, Iterable, Iterator, List, Optional, Tuple
+from typing import Any, Iterator, List, Mapping, Optional, Sequence, Tuple, Union
 
 from typing_extensions import Protocol
 
 """
 Some very basic protocol definitions for the DB-API2 classes specified in PEP-249
 """
 
+_Parameters = Union[Sequence[Any], Mapping[str, Any]]
```
clokep, line 23: It is unclear to me if PEP249 actually means `Sequence` when it refers to `Sequence` or not. The type shed hints for this use `Iterable`, not `Sequence`: https://github.com/python/typeshed/blob/master/stdlib/sqlite3/dbapi2.pyi#L142

ShadowJonathan: I think it's safer to do `Sequence`, the difference between `Iterable` and `Sequence` is that `Iterable` could be a generator or a `set`, `Sequence` cannot, and I'd be willing to bet that most db implementations are betting on an indexable iterable, ergo a sequence, to be passed along.

Though if you still think `Iterable` is better, just say so, but that's my stand.

synapse/storage/prepare_database.py
ShadowJonathan: It's not due to the change above, but `txn.fetchone()` is `Optional[]`, and while mypy does not do proper advanced branch checking to see that `current_version` is truthy if `row` is truthy (or `not None`), to see that correlation.

The above change (`row` being truthy vs `row` being `not None`) didn't change this error, i put this assert here to assure mypy that this is okay.

synapse/storage/types.py
ShadowJonathan: Can you find a source on it? Because I couldn't find a better one than the PEP and some other `dbapi`-esc database files.

synapse/storage/prepare_database.py
ShadowJonathan:
```suggestion
    current_version = int(row[0]) if row else None
```

synapse/storage/types.py
clokep: I'm fairly certain the column names must be strings. The others I was simply basing it off the names, but as we don't use them in synapse it most likely doesn't matter.

Interestingly [sqlite3's `description`](https://docs.python.org/3/library/sqlite3.html#sqlite3.Cursor.description) also has `None` as the second field.

synapse/storage/types.py
clokep: > I'd be willing to bet that most db implementations are betting on an indexable iterable, ergo a sequence, to be passed along.

I would be surprised if they need it to be indexable, but you're correct that generators / sets cannot be used here. I don't think it matters too much, if this works as `Sequence` than that's fine with me.

synapse/storage/prepare_database.py
clokep: So I think this code can be simplified a bit, by changing the `if current_version` to `if row is not None` and then setting `current_version = int(row[0])` as the first line in the if-statement.

synapse/storage/types.py
ShadowJonathan: ...i'll make the second one `Optional` as well then, i'll make the first `str`, and the last 5 `int`, does that sound good?
I'll also add a note that the exact types of those 7 tuple-indexes are relatively undefined, in case we encounter a problem like this one again in the future.

synapse/storage/types.py
```diff
@@ -12,31 +12,52 @@
...
 
     @property
-    def description(self) -> Any:
-        return None
+    def description(
+        self,
+    ) -> Optional[
+        Sequence[
+            # fixme: this is an approximate typing of dbapi2's cursor.description, going off of sqlite3 and other
+            #  drivers, it may not be entirely accurate.
```
clokep, line 48:
```suggestion
            # Note that this is an approximate typing based on sqlite3 and other
            # drivers, and may not be entirely accurate.
```

changelog.d/9299.misc
```diff
@@ -0,0 +1 @@
+Update internal database-related types to better match PEP 249.
```
clokep, line 1: Update the `Cursor` type hints to better match PEP 249.

clokep (approved):
Thanks!

PR #9300, opened 2021-02-02
social login: add noopener to terms link

richvdh:
this is basically a security thing, aiui


clokep (approved):
Seems fine. Are there other links we have in our templates that should use this while we're at it?

richvdh:
> Seems fine. Are there other links we have in our templates that should use this while we're at it?

I believe not, because I think this is the only place we open the link in a new tab (and `noopener` only matters if the opener remains open)

PR #9305, opened 2021-02-02
SRV request logging

Bubu:
### Pull Request Checklist


Bubu:
I still had this lying around. This was motivated by a daylong quest to figure out why my federation was working only for some remote servers. (turned out to be a dns problem...)

synapse/http/federation/matrix_federation_agent.py
```diff
@@ -323,12 +323,17 @@ async def _resolve_server(self) -> List[Server]:
         if port or _is_ip_literal(host):
             return [Server(host, port or 8448)]
 
+        logger.debug("Looking up SRV record for %s", host)
```
clokep, line 326: I believe that `host` is `bytes` here so this will end up logging:

> Looking up SRV record for b'foo'

I think you'll want to to `host.decode()` here to get it to a string.

This might also be true in other spots. 😄

clokep (changes requested):

synapse/http/federation/matrix_federation_agent.py
Bubu: Right, I got tricked by that multiple times already.

Should be fixed.

synapse/http/federation/matrix_federation_agent.py
```diff
@@ -323,12 +323,19 @@ async def _resolve_server(self) -> List[Server]:
         if port or _is_ip_literal(host):
             return [Server(host, port or 8448)]
 
+        logger.debug("Looking up SRV record for %s", host.decode())
```
richvdh, line 326: this will throw an exception if `host` contains invalid utf-8. 

I'm not sure if that is a real concern here, but as a rule best to user `decode(errors='replace')` or somesuch for logging.

clokep: I don't think it is a real concern here since `host` has to be ascii? (Probably?) But yeah using something like that sounds better. :)

Bubu: I also don't think we can get to this point with an undecodable host... but anyway, still makes a lot of sense to do that. Changed.

clokep (approved):
LGTM.

PR #9119, opened 2021-01-14
allow specifying https:// proxy

Bubu:
### Pull Request Checklist

* [x] Pull request is based on the develop branch
* [x] Pull request includes a [changelog file](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md#changelog). The entry should:
  - Be a short description of your change which makes sense to users. "Fixed a bug that prevented receiving messages from other servers." instead of "Moved X method from `EventStore` to `EventWorkerStore`.".
  - Use markdown where necessary, mostly for `code blocks`.
  - End with either a period (.) or an exclamation mark (!).
  - Start with a capital letter.
* [x] Pull request includes a [sign off](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md#sign-off)
* [x] Code style is correct (run the [linters](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md#code-style))

Issue #9090, opened 2021-01-12
Support for https traffic to the proxy


Bubu:
**Description:**

Note: this is different from talking https via the proxy (via HTTPS_PROXY). This is about supporting https for the synapse -> proxy-server connection.

From what I understand this feature is not that widely supported yet (though the article below is from 2016). But for complete proxy support for synapse this might be worth adding.

Context: https://daniel.haxx.se/blog/2016/11/26/https-proxy-with-curl/

Bubu:
There's a comment about this here: https://github.com/matrix-org/synapse/blob/develop/synapse/http/proxyagent.py#L187-L189

Bubu:
I have a working commit for this now: https://git.bubu1.eu/Bubu/synapse/-/commit/d7c77ce51ed002af94316c1c277c0d020cd26368

Need to add tests, etc.

anoadragon453:
This looks promising, thanks @Bubu! In-progress PR is here: https://github.com/matrix-org/synapse/pull/9119

Issue #9088, opened 2021-01-12
Support for proxy exceptions (e.g. NO_PROXY env var)


Bubu:
**Description:**

Or I guess another way of specifying domains/hosts which should be exempt from proxy-ing.

From the curl man page:
```
      NO_PROXY <comma-separated list of hosts/domains>
              list of host names that shouldn't go through any proxy. If set to an asterisk '*' only, it matches all hosts. Each name in this list is matched as either a domain name which contains the hostname, or  the  hostname  it‐
              self.

              This environment variable disables use of the proxy even when specified with the -x, --proxy option. That is NO_PROXY=direct.example.com curl -x http://proxy.example.com http://direct.example.com accesses the target URL
              directly, and NO_PROXY=direct.example.com curl -x http://proxy.example.com http://somewhere.example.com accesses the target URL through the proxy.

              The list of host names can also be include numerical IP addresses, and IPv6 versions should then be given without enclosing brackets.

              IPv6 numerical addresses are compared as strings, so they will only match if the representations are the same: "::1" is the same as "::0:1" but they don't match.
```

I don't currently know how this would interact with the ip_range_blacklist :thinking:

Bubu:
> I don't currently know how this would interact with the ip_range_blacklist thinking

The reason why this interacts at all is: https://github.com/matrix-org/synapse/pull/9084 as the blacklist is not applied for connections going through the proxy.

I think that the simplest and sanest solution would be that everything not going through the proxy (because it's in NO_PROXY) would then be subject to the blacklist/whitelist again and if you need to connect to an internal IP for something that should be explicitly mentioned in the whitelist.

Bubu:
Ah, there's an issue here already: https://github.com/matrix-org/synapse-dinsic/issues/70

Bubu:
I believe this is fixed since 1.29 via https://github.com/matrix-org/synapse/pull/9372

Issue #8859, opened 2020-12-02
Support proxy server for outbound federation


sim0nx:
**Description:**
In a setup where there is the synapse server (machine B) behind a reverse proxy (machine A), the two being on two different machines, and the machine running synapse not having direct internet access (but only via proxy) it should be possible to proxy the federation connections.
Just as is done e.g. for notifications.

<internet> -- reverse proxy (machine A)  -- synapse (machine B, access to internet only via proxy)

There was already an issue on this (#4198) although the implementation which resulted out of that (#6239) explicitly does not support proxy support for federation.

clokep:
Thanks for filing this! It is definitely something we should do at some point. I'm going to close this as a duplicate of #8660.

PR #9306, opened 2021-02-02
support federation queries through http connect proxy

### LINKED ISSUE #8660: Add support for HTTPS proxy for federation requests
This will be helpful for certain use cases, but mainly I'd like to see if using one reduces CPU usage for small servers that end up making lots of connections at once.

Bubu:
Can be specified by HTTPS_PROXY env var.

pass unfiltered reactor to federation agent for proxy support

Signed-off-by: Marcus Hoffmann <bubu@bubu1.eu>

### Pull Request Checklist

Fixes #8660

### DIFF
--- a/synapse/http/federation/matrix_federation_agent.py
+++ b/synapse/http/federation/matrix_federation_agent.py
@@ -36,7 +36,7 @@ from synapse.http.client import BlacklistingAgentWrapper
 from synapse.http.connectproxyclient import HTTPConnectProxyEndpoint
 from synapse.http.federation.srv_resolver import Server, SrvResolver
 from synapse.http.federation.well_known_resolver import WellKnownResolver
-from synapse.http.proxyagent import ProxyAgent
+from synapse.http.proxyagent import ProxyAgent, parse_username_password
 from synapse.logging.context import make_deferred_yieldable, run_in_background
 from synapse.types import ISynapseReactor
 from synapse.util import Clock
@@ -76,7 +76,7 @@ class MatrixFederationAgent:
         tls_client_options_factory: Optional[FederationPolicyForHTTPS],
         user_agent: bytes,
         ip_blacklist: IPSet,
-        proxy_reactor: IReactorCore = None,
+        proxy_reactor: Optional[ISynapseReactor] = None,
         _srv_resolver: Optional[SrvResolver] = None,
         _well_known_resolver: Optional[WellKnownResolver] = None,
     ):
@@ -231,6 +231,8 @@ class MatrixHostnameEndpointFactory:
             srv_resolver = SrvResolver()
 
         self._srv_resolver = srv_resolver
+        self.https_proxy_creds, https_proxy = parse_username_password(https_proxy)
+
         self.https_proxy_endpoint = proxyagent.http_proxy_endpoint(
             https_proxy, proxy_reactor
         )
@@ -242,6 +244,7 @@ class MatrixHostnameEndpointFactory:
             self._srv_resolver,
             parsed_uri,
             self.https_proxy_endpoint,
+            self.https_proxy_creds,
         )
 
 
@@ -265,12 +268,15 @@ class MatrixHostnameEndpoint:
         srv_resolver: SrvResolver,
         parsed_uri: URI,
         https_proxy_endpoint: Optional[IStreamClientEndpoint],
+        https_proxy_creds,
     ):
         self._reactor = reactor
 
         self._parsed_uri = parsed_uri
 
         self.https_proxy_endpoint = https_proxy_endpoint
+
+        self.https_proxy_creds = https_proxy_creds
 
         # set up the TLS connection params
         #
@@ -309,8 +315,17 @@ class MatrixHostnameEndpoint:
                         port,
                         self.https_proxy_endpoint,
                     )
+                    connect_headers = Headers()
+                    # Determine whether we need to set Proxy-Authorization headers
+                    if self.https_proxy_creds:
+                        # Set a Proxy-Authorization header
+                        connect_headers.addRawHeader(
+                            b"Proxy-Authorization",
+                            self.https_proxy_creds.as_proxy_authorization_value(),
+                        )
+
                     endpoint = HTTPConnectProxyEndpoint(
-                        self._reactor, self.https_proxy_endpoint, host, port
+                        self._reactor, self.https_proxy_endpoint, host, port, headers=connect_headers
                     )
                 else:
                     logger.debug("Connecting to %s:%i", host.decode("ascii"), port)
--- a/synapse/http/federation/matrix_federation_agent.py
+++ b/synapse/http/federation/matrix_federation_agent.py
@@ -326,7 +326,11 @@ class MatrixHostnameEndpoint:
                         )
 
                     endpoint = HTTPConnectProxyEndpoint(
-                        self._reactor, self.https_proxy_endpoint, host, port, headers=connect_headers
+                        self._reactor,
+                        self.https_proxy_endpoint,
+                        host,
+                        port,
+                        headers=connect_headers,
                     )
                 else:
                     logger.debug("Connecting to %s:%i", host.decode("ascii"), port)
--- a/synapse/http/federation/matrix_federation_agent.py
+++ b/synapse/http/federation/matrix_federation_agent.py
@@ -14,6 +14,7 @@
 import logging
 import urllib.parse
 from typing import Any, Generator, List, Optional
+from urllib.request import getproxies, proxy_bypass
 
 from netaddr import AddrFormatError, IPAddress, IPSet
 from zope.interface import implementer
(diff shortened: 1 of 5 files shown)


Bubu:
@clokep I'd like your feedback here, some notes:

* This has been running in production on an instance of synapse that only supports outgoing connections via a proxy. So I'm somewhat certain that in general it does the correct thing.
* It obviously still needs tests. I didn't actually look into that yet.
* this doesn't attempt to tackle the fact that the blacklisting should probably be only happening at the proxy instead of in synapse. I think this can be solved independently though. 

* From the *will not be used for* list from https://github.com/matrix-org/synapse/pull/6239 this only adds fed traffic as there was no immediate need for the other things on that list (in fact adding some of those would need #9088 to be solved)
* This might be solving #8859, though the reasoning in that issue is currently not clear to me (ssl offloading vs. us needed that just for connectivity to other servers)
* The thing I'm unsure about is how to best introduce this: It changes existing functionality if you use a proxy server for anything... but on the other hand introducing more (nonstandard) env vars or config options for this doesn't seem satisfactory either. Maybe add this together with #9088 and pointing it out in the upgrading notes would be enough?

clokep:
I didn't look in detail yet, but a quick look makes it look reasonable. Is this only HTTP*S* proxy support? (No HTTP proxy support?)

I've set the review flag so that someone can provide feedback!

Bubu:
> I didn't look in detail yet, but a quick look makes it look reasonable. Is this only HTTPS proxy support? (No HTTP proxy support?)

This is https only (target server is on https, *not* the proxy server like #9090) because I was under the impression that federation traffic requires a tls connection anyway. (I should be easy enough to pass through the http_proxy as well, if that is necessary for some reason.)

synapse/server.py
```diff
@@ -388,7 +388,9 @@ def get_federation_http_client(self) -> MatrixFederationHttpClient:
         tls_client_options_factory = context_factory.FederationPolicyForHTTPS(
             self.config
         )
-        return MatrixFederationHttpClient(self, tls_client_options_factory)
+        return MatrixFederationHttpClient(
+            self, tls_client_options_factory, https_proxy=os.getenvb(b"HTTPS_PROXY"),
```
clokep, line 392: I suspect we would also want to support `http_proxy` (which I suspect is not too much more work).

synapse/http/federation/matrix_federation_agent.py
```diff
@@ -211,13 +231,17 @@ def __init__(
             srv_resolver = SrvResolver()
 
         self._srv_resolver = srv_resolver
+        self.https_proxy_endpoint = proxyagent.http_proxy_endpoint(
+            https_proxy, proxy_reactor
+        )
```
clokep, line 236: Shouldn't we not create this endpoint if `https_proxy` is not given?

clokep (commented):
> * This has been running in production on an instance of synapse that only supports outgoing connections via a proxy. So I'm somewhat certain that in general it does the correct thing.

Awesome news! I know that there's interest by others in running with this too!

> * It obviously still needs tests. I didn't actually look into that yet.

Yeah...it is rather unfortunate that the federation client is separate from the other HTTP client, but I don't think fixing that is trivial.

> * this doesn't attempt to tackle the fact that the blacklisting should probably be only happening at the proxy instead of in synapse. I think this can be solved independently though.

I think non-federation requests somewhat have the same issue right now, so I don't see this as a problem to solve here.

> * From the _will not be used for_ list from #6239 this only adds fed traffic as there was no immediate need for the other things on that list (in fact adding some of those would need #9088 to be solved)

Ideally we would eventually add proxy support for all of these locations, I think. But not 100% sure. This fixes #8660, might be worth filing an issue with the remaining places we don't use a proxy?

> * This might be solving #8859, though the reasoning in that issue is currently not clear to me (ssl offloading vs. us needed that just for connectivity to other servers)

This solves #8660, which I've added to the description (#8859 was marked as duplicate of #8660).

> * The thing I'm unsure about is how to best introduce this: It changes existing functionality if you use a proxy server for anything... but on the other hand introducing more (nonstandard) env vars or config options for this doesn't seem satisfactory either. Maybe add this together with #9088 and pointing it out in the upgrading notes would be enough?

For #6239 it seems we just turned it on, I don't think we would do something different here. I suspect documenting it in the upgrade notes should be enough.

synapse/server.py
Bubu: Is it possible that federation requests could end up with a http target? I was under the impression that https is a requirement for federation?

synapse/http/federation/matrix_federation_agent.py
Bubu: `http_proxy_endpoint(...)` returns `None` when you give it `None` as an input, so I think this is fine?

synapse/server.py
dkasak: The `HTTP_PROXY` and `HTTPS_PROXY` variables refer to the type of connection between the client and the proxy, not the kind of requests that the proxy will be proxying upstream. The proxy should support both HTTP and HTTPS requests regardless of this setting.

dkasak: Though actually, I always feel a bit insane when talking about this, because it seems there are programs that flat out disagree on this. I just noticed your #9090 🤔

tzyl:
Thanks for starting this @Bubu! This would also be really useful for me and I'm running a similar patch to support a setup where outbound traffic flows through an egress proxy host. A follow up that I would also want would be to support the `NO_PROXY` convention environment variable (as urllib and other non-python software does) which I'm happy to contribute following this.

synapse/server.py
clokep: Arg I seem to have made the same mistake as @dkasak in getting those two confused. I think the federation APIs only talk via HTTPS.

clokep:
@Bubu based on the conversation above...do you think this is ready for more feedback or what's the next step?

richvdh:
@Bubu is this ready for review, or are you still working on it? It looks like a couple of conflicts have cropped up.

f1-outsourcing:
I also have a problem with the federation traffic not going through the proxy. Can anyone point me to a patch that I can apply to 1.31.0 ?

Bubu:
@richvdh I resolved the conflicts. In general review is appreciated though it still lacks functioning tests.

(I currently don't have time to look into writing those, though potentially @chrisguida wanted to look at that.)

chrisguida:
> @richvdh I resolved the conflicts. In general review is appreciated though it still lacks functioning tests.
> 
> (I currently don't have time to look into writing those, though potentially @chrisguida wanted to look at that.)

If anyone else wants to take on writing the test so that this can be merged, please go ahead. I've run this code and it works, but figuring out how to write the test properly will most likely take me a while since I'm not too familiar with the testing framework. Otherwise I'll get to it when time permits.

synapse/http/federation/matrix_federation_agent.py
```diff
@@ -72,6 +76,7 @@ def __init__(
         tls_client_options_factory: Optional[FederationPolicyForHTTPS],
         user_agent: bytes,
         ip_blacklist: IPSet,
+        proxy_reactor: Optional[ISynapseReactor] = None,
```
richvdh, line 79: please add new params to the docstring (on the class, in this case). As well as making the code maintainable in future, it's important to help with review.

synapse/http/matrixfederationclient.py
```diff
@@ -279,6 +279,7 @@ def __init__(self, hs, tls_client_options_factory):
             tls_client_options_factory,
             user_agent,
             hs.config.federation_ip_range_blacklist,
+            proxy_reactor=hs.get_reactor(),
```
richvdh, line 282: given that we already pass the ip blacklist into `MatrixFederationAgent`, why not move the construction of `BlacklistingReactorWrapper` down to it, rather than having to pass in two reactors?

richvdh (changes requested):
ok, I don't really understand what's going on in this code. Please update the docstrings to help me follow it.

tests/http/federation/test_matrix_federation_agent.py
```diff
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
```
dklimpel, line 1:
```suggestion

```

This is no longer needed. See #9786

dklimpel:
> > (I currently don't have time to look into writing those, though potentially @chrisguida wanted to look at that.)
> 
> If anyone else wants to take on writing the test so that this can be merged, please go ahead. I've run this code and it works, but figuring out how to write the test properly will most likely take me a while since I'm not too familiar with the testing framework. Otherwise I'll get to it when time permits.

I will work on it after #10360 and #9119 is merged.
Some of this code and tests is needed to keep it simpler and avoid conflicts.

ShadowJonathan:
I'm seeing uses of `urllib` here, isn't it possible to use `urllib3` for this?

dklimpel:
> I'm seeing uses of `urllib` here, is it possible to use `urllib3` for this?
> 
> My main gripe with it is that `urllib3` is newer, but also that `urllib` produces `bytes` instead of `str`, which speaks of python 2 behaviour. Those bytes may not be UTF8, or properly parsed from the environment, so for compatibility i'd suggest trying to see if you can get `urllib3` to work with this.

I am not sure if it is useful to use the library only here.
If would have to check them certainly in whole Synapse.

There is also a discussion about it there: https://github.com/matrix-org/synapse/pull/9119#discussion_r592441358

clokep:
I'm going to close this in lieu of #10475 where @dklimpel picked up this work. Hopefully that's OK! 🎉

Bubu:
Sure, sorry for not having the badwidth to continue here :-(.

And thanks for picking this up @dklimpel.

PR #9307, opened 2021-02-03
Add debug for OIDC flow

richvdh:
just some improvements to the logging here, to help understand the failure modes.

### DIFF
--- a/synapse/handlers/oidc_handler.py
+++ b/synapse/handlers/oidc_handler.py
@@ -123,7 +123,6 @@ class OidcHandler:
         Args:
             request: the incoming request from the browser.
         """
-
         # The provider might redirect with an error.
         # In that case, just display it as-is.
         if b"error" in request.args:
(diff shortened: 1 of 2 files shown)


synapse/handlers/oidc_handler.py
```diff
@@ -183,14 +186,16 @@ async def handle_oidc_callback(self, request: SynapseRequest) -> None:
...
+            logger.exception("Invalid session for OIDC callback")
             self._sso_handler.render_error(request, "invalid_session", str(e))
             return
         except MacaroonInvalidSignatureException as e:
-            logger.exception("Could not verify session")
+            logger.exception("Could not verify session for OIDC callback")
             self._sso_handler.render_error(request, "mismatching_session", str(e))
             return
 
+        logger.info("Received OIDC callback for IdP %s", session_data.idp_id)
```
clokep, line 197: I wonder if more of these log lines should have info about the IdP in use.

clokep (approved):
Looks OK to me. I feel like it might not be great to log some of the tokens (security wise), but at debug it should be OK.

PR #9310, opened 2021-02-03
Clarify documentation about escaping URLs in templates.

clokep:
#9200 failed to update some documentation in the sample config.


clokep:
I see no reason not to put this into 1.27.0 as it is a docs only change for something added in 1.27.0rc1.

richvdh (approved):
lgtm.

PR #9311, opened 2021-02-03
Add `check_media_file_for_spam` spam checker hook

erikjohnston:
This is to allow us to reject known bad data, or e.g. do some anti virus scanning.

Some notes:
* This responds with generic 404 for remote content, but a 400 "Bad Content" for uploads. Is that what we want?
* If we reject remote media we'll keep downloading and then deleting the data for each request.

An example usage:

https://github.com/matrix-org/synapse/blob/d71524f686315645b7c0b4de0e816cdf2410d836/tests/rest/media/v1/test_media_storage.py#L433-L437

### DIFF
--- a/synapse/events/spamcheck.py
+++ b/synapse/events/spamcheck.py
@@ -17,6 +17,8 @@
 import inspect
 from typing import TYPE_CHECKING, Any, Dict, List, Optional, Tuple, Union
 
+from synapse.rest.media.v1._base import FileInfo
+from synapse.rest.media.v1.media_storage import ReadableFileWrapper
 from synapse.spam_checker_api import RegistrationBehaviour
 from synapse.types import Collection
 from synapse.util.async_helpers import maybe_awaitable
@@ -214,3 +216,48 @@ class SpamChecker:
                     return behaviour
 
         return RegistrationBehaviour.ALLOW
+
+    async def check_media_file_for_spam(
+        self, file_wrapper: ReadableFileWrapper, file_info: FileInfo
+    ) -> bool:
+        """Checks if a piece of newly uploaded media should be blocked.
+
+        This will be called for local uploads, downloads of remote media, each
+        thumbnail generated for those, and web pages/images used for URL
+        previews.
+
+        Note that care should be taken to not do blocking IO operations in the
+        main thread. For example, to get the contents of a file a module
+        should do::
+
+            async def check_media_file_for_spam(
+                self, file: ReadableFileWrapper, file_info: FileInfo
+            ) -> bool:
+                buffer = BytesIO()
+                await file.write_chunks_to(buffer.write)
+
+                if buffer.getvalue() == b"Hello World":
+                    return True
+
+                return False
+
+
+        Args:
+            file: An object that allows reading the contents of the media.
+            file_info: Metadata about the file.
+
+        Returns:
+            True if the media should be blocked or False if it should be
+            allowed.
+        """
+
+        for spam_checker in self.spam_checkers:
+            # For backwards compatibility, only run if the method exists on the
+            # spam checker
+            checker = getattr(spam_checker, "check_media_file_for_spam", None)
+            if checker:
+                spam = await maybe_awaitable(checker(file_wrapper, file_info))
+                if spam:
+                    return True
+
+        return False
--- a/synapse/rest/media/v1/media_storage.py
+++ b/synapse/rest/media/v1/media_storage.py
@@ -16,13 +16,17 @@ import contextlib
 import logging
 import os
 import shutil
-from typing import IO, TYPE_CHECKING, Any, Optional, Sequence
+from typing import IO, TYPE_CHECKING, Any, Callable, Optional, Sequence
+
+import attr
 
 from twisted.internet.defer import Deferred
 from twisted.internet.interfaces import IConsumer
 from twisted.protocols.basic import FileSender
 
+from synapse.api.errors import NotFoundError
 from synapse.logging.context import defer_to_thread, make_deferred_yieldable
+from synapse.util import Clock
 from synapse.util.file_consumer import BackgroundFileConsumer
 
 from ._base import FileInfo, Responder
@@ -58,6 +62,8 @@ class MediaStorage:
         self.local_media_directory = local_media_directory
         self.filepaths = filepaths
         self.storage_providers = storage_providers
+        self.spam_checker = hs.get_spam_checker()
+        self.clock = hs.get_clock()
 
     async def store_file(self, source: IO, file_info: FileInfo) -> str:
         """Write `source` to the on disk media store, and also any other
@@ -127,18 +133,29 @@ class MediaStorage:
                     f.flush()
                     f.close()
 
+                    spam = await self.spam_checker.check_media_file_for_spam(
+                        ReadableFileWrapper(self.clock, fname), file_info
+                    )
+                    if spam:
+                        logger.info("Blocking media due to spam checker")
+                        # Note that we'll delete the stored media, due to the
+                        # try/except below. The media also won't be stored in
+                        # the DB.
+                        raise SpamMediaException()
+
                     for provider in self.storage_providers:
                         await provider.store_file(path, file_info)
 
                     finished_called[0] = True
 
                 yield f, fname, finish
-        except Exception:
+        except Exception as e:
             try:
                 os.remove(fname)
             except Exception:
                 pass
-            raise
+
+            raise e from None
 
         if not finished_called:
             raise Exception("Finished callback not called")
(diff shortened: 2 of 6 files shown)


synapse/rest/media/v1/media_storage.py
```diff
@@ -302,3 +316,44 @@ def write_to_consumer(self, consumer: IConsumer) -> Deferred:
...
+
+
+@attr.s(slots=True)
+class ReadableFileWrapper:
+    """Wrapper that allows reading a file in chunks, yielding to the reactor,
+    and writing to an IO stream.
+
+    This is simplified `FileSender` that takes an IO object rather than an
+    `IConsumer`.
+    """
```
erikjohnston, line 336: The rationale here is that we don't want to give the raw path to the spam checker so that people don't just do `file.read()`, potentially blocking the reactor while we do large reads. Instead we do what `FileSender` does and read small chunks, yielding to the reactor between reads. I'm not 100% if that's the right way of doing things, but at least it's what we currently do for file downloads.

daenney:
I don't think a 400 is appropriate. "Bad request" generally means there's something malformed about the request the client did, not "the content itself is illegal".

https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400
> The HyperText Transfer Protocol (HTTP) 400 Bad Request response status code indicates that the server cannot or will not process the request due to something that is perceived to be a client error (e.g., malformed request syntax, invalid request message framing, or deceptive request routing).

The 422, unprocessable entity, is probably the closest thing
https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/422
> The HyperText Transfer Protocol (HTTP) 422 Unprocessable Entity response status code indicates that the server understands the content type of the request entity, and the syntax of the request entity is correct, but it was unable to process the contained instructions.

synapse/rest/media/v1/media_storage.py
richvdh: it's really not the right way of doing things. Irrespective of whether non-blocking IO is perfect, it's better than doing regular IO on the reactor thread.

It's probably good enough for a quick hack, but it would be nice to come back to it and either make a non-blocking-io thing which integrates with the twisted reactor, or give up and uses threads.

synapse/rest/media/v1/media_storage.py
```diff
@@ -302,3 +316,44 @@ def write_to_consumer(self, consumer: IConsumer) -> Deferred:
...
+
+    CHUNK_SIZE = 2 ** 14
+
+    clock = attr.ib(type=Clock)
+    path = attr.ib(type=str)
+
+    async def write_to_io(self, io: BinaryIO):
```
clokep, line 343: Maybe this should just take a callback and if someone wants to write to a buffer they can?

synapse/events/spamcheck.py
```diff
@@ -214,3 +216,47 @@ async def check_registration_for_spam(
...
+        Returns:
+            True if the media should be blocked or False if it should be
+            allowed.
+        """
+
+        for spam_checker in self.spam_checkers:
+            # For backwards compatibility, only run if the method exists on the
+            # spam checker
+            checker = getattr(spam_checker, "check_media_file_for_spam", None)
+            spam = await maybe_awaitable(checker(file_wrapper, file_info))
```
clokep, line 258: I think this code might not really work properly with multiple spam checkers?

synapse/rest/media/v1/media_storage.py
clokep: Ah I see this gives control of how often to request data to the spam checker while I'm suggesting keep that control, which essentially would be a `IProducer` and ask the spam checker to be an `IConsumer`?

synapse/events/spamcheck.py
erikjohnston: Why? Though I have forgotten to add an if check there

synapse/rest/media/v1/media_storage.py
erikjohnston: True

synapse/events/spamcheck.py
clokep: Oh, this should be OK. I completely misread how to use the `ReadableFileWrapper` API. It is a one-shot to read the file in chunks. I thought it returned only a single chunk and you had to keep calling it.

My only comment would be that this could be inefficient with multiple spam checkers, but that's not a huge deal.

synapse/events/spamcheck.py
```diff
@@ -214,3 +216,48 @@ async def check_registration_for_spam(
...
+        main thread. For example, to get the contents of a file a module
+        should do::
+
+            async def check_media_file_for_spam(
+                self, file: ReadableFileWrapper, file_info: FileInfo
+            ) -> bool:
+                buffer = BytesIO()
+                await file.write_chunks_to(buffer.write)
+
+                if bytes.getvalue() == b"Hello World":
```
clokep, line 239:
```suggestion
                if buffer.getvalue() == b"Hello World":
```

erikjohnston:
> I don't think a 400 is appropriate. "Bad request" generally means there's something malformed about the request the client did, not "the content itself is bad/undesirable".
> 
> https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400
> 
> > The HyperText Transfer Protocol (HTTP) 400 Bad Request response status code indicates that the server cannot or will not process the request due to something that is perceived to be a client error (e.g., malformed request syntax, invalid request message framing, or deceptive request routing).
> 
> The 422, unprocessable entity, is probably the closest thing
> https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/422
> 
> > The HyperText Transfer Protocol (HTTP) 422 Unprocessable Entity response status code indicates that the server understands the content type of the request entity, and the syntax of the request entity is correct, but it was unable to process the contained instructions.

I think if we are going to make it obvious to the clients they have been blocked by the spam checking module then 400 is probably the more widely expected code, I think 422 is just going to freak people out. Eh, I don't know.

synapse/rest/media/v1/media_storage.py
erikjohnston: Have filed https://github.com/matrix-org/synapse/issues/9312

daenney:
I’m not sure the purpose of status codes is to scare clients/users. Shouldn’t that be up to the user agent, like Ele web? A 422 could more easily be rendered as a “hey this file is really bad” in the UI compared to a 400 that could mean “we broke something at the http request level”

synapse/rest/media/v1/media_storage.py
```diff
@@ -302,3 +316,41 @@ def write_to_consumer(self, consumer: IConsumer) -> Deferred:
...
     def __exit__(self, exc_type, exc_val, exc_tb):
         self.open_file.close()
+
+
+class SpamMediaException(NotFoundError):
+    """The media was blocked by a spam checker, so we simply 404 the request (in
+    the same way as if it was quarantined).
+    """
+
+    pass
```
clokep, line 326:
```suggestion
```

synapse/events/spamcheck.py
```diff
@@ -214,3 +216,48 @@ async def check_registration_for_spam(
...
         return RegistrationBehaviour.ALLOW
+
+    async def check_media_file_for_spam(
+        self, file_wrapper: ReadableFileWrapper, file_info: FileInfo
+    ) -> bool:
+        """Checks if a piece of newly uploaded media should be blocked.
+
+        This will be called for local uploads, downloads of remote media, each
+        thumbnail generated for those, and web pages/images used for URL
+        previews.
```
clokep, line 227: It seems a bit odd to me to do this for thumbnail generation, but I think it just due to where we're hooking into the code here?

synapse/rest/media/v1/media_storage.py
```diff
@@ -127,18 +133,26 @@ async def finish():
                     f.flush()
                     f.close()
 
+                    spam = await self.spam_checker.check_media_file_for_spam(
+                        ReadableFileWrapper(self.clock, fname), file_info
+                    )
+                    if spam:
+                        logger.info("Blocking media due to spam checker")
```
clokep, line 140: Should we include some info about what was blocked while logging?

synapse/rest/media/v1/media_storage.py
```diff
@@ -127,18 +133,26 @@ async def finish():
...
+                        raise SpamMediaException()
```
clokep, line 141: It might be good to document somewhere what happens in this case. It seems the file gets deleted, I assume it doesn't end up in the database?

clokep (commented):
I think this looks good overall, but I'm unsure if @richvdh is happy with the IO conversation.

richvdh:
> I think this looks good overall, but I'm unsure if @richvdh is happy with the IO conversation.

I've not gone through the code, but from what I saw I'm happy with this under a "no worse than what we have" banner.

synapse/events/spamcheck.py
erikjohnston: Yeah. I can make it not do thumbnails, but its not clear that is the right thing to do. Spam checkers can check that themselves.

synapse/rest/media/v1/media_storage.py
erikjohnston: We don't actually know very much about the media annoyingly, e.g. we don't know the media ID. I hope we can extract enough info from the request logs.

synapse/rest/media/v1/media_storage.py
erikjohnston: I've added a comment here, hopefully that is good enough?

synapse/rest/media/v1/media_storage.py
clokep: Doesn't `file_info` have some stuff in it?

erikjohnston: Yeah, it has some stuff, but not the media ID. We could log if its a remote download or not, but I'm not sure that is particularly useful given you can get that from the request logs?

clokep: Oh, I see all that info is pretty much in the requests logs. 👍

clokep (approved):

PR #9313, opened 2021-02-03
Backout changes for automatically calculating the public baseurl

### LINKED ISSUE #9264: Login fails after update from 1.25.0 to 1.26.0 - syncing forever ... many rooms
### Description

We did an update to our system today from V 1.25.0 to the recent 1.26.0. Since that time no Users can log in anymore.

Still logged in users do not see any problem. Even after a restart of synapse.

### Steps to reproduce

Point the webbrowser (Firefox or safari Mac, or FF Linux) to the element webpage or open the element app on Mac, linux, iOS and point it to the URL.

Enter the login username and password for an AD user.

In the webbrowser I get the message: "_Syncing, If you've joined lots of rooms, this might take a while._"
On the iOS iPad I get a spinning logo.

I'd expect to be logged in an beeing granted access to all my room etc.

<!--

``` (three backticks, on a line on their own), so that they are formatted legibly.
-->

### Version information

- **Homeserver**:

matrix.dangermouse.filmakademie.de
<!--

listen on that port.)
-->
- **Version**:

{"server_version":"1.26.0","python_version":"3.6.9"}

- **Install method**:

install: apt install matrix-synapse-py3
update: apt upgrade

- **Platform**:
<!--
-->
Ubuntu 18.04.5 LTS

Virtual server.


**Note**:

Nevertheless: Amazing project, great Work !!!!  - Kowtow

### PR DESCRIPTION
**Pull request by clokep**
This is an attempt to fix #9264:

* Reverts most of the changes from #9159.
* Cleans-up some checks for `public_baseurl`.
* Avoids loading templates that are troublesome if `public_baseurl` is set, in which case they won't be used anyway.
* Old configs should work fine now *except* if you're using CAS, which now requires `public_baseurl` (and refuses to start if it is not configured). This matches the behavior of the other SSO implementations (and is the least bad way I could figure out how to do this).

This approach feels pretty brittle to me, but I'm not sure if there's a more clever thing to do in the short term.

### DIFF
--- a/synapse/config/cas.py
+++ b/synapse/config/cas.py
@@ -37,13 +37,6 @@ class CasConfig(Config):
             if not public_baseurl:
                 raise ConfigError("cas_config requires a public_baseurl to be set")
 
-            # If the service_url is given, use it for backwards compatibility.
-            # Otherwise, use the public_baseurl.
-            if cas_config.get("service_url"):
-                public_baseurl = cas_config["service_url"]
-                if public_baseurl[-1] != "/":
-                    public_baseurl += "/"
-
             # TODO Update this to a _synapse URL.
             self.cas_service_url = public_baseurl + "_matrix/client/r0/login/cas/ticket"
             self.cas_displayname_attribute = cas_config.get("displayname_attribute")
--- a/synapse/config/cas.py
+++ b/synapse/config/cas.py
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from ._base import Config
+from ._base import Config, ConfigError
 
 
 class CasConfig(Config):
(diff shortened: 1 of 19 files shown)

### CODE REVIEW

File: synapse/config/cas.py [hunk 1]
```diff
@@ -30,13 +30,22 @@ def read_config(self, config, **kwargs):
...
-                public_base_url += "/"
+
+            # The public baseurl is required because it is used by the redirect
+            # template.
+            public_baseurl = self.public_baseurl
+            if not public_baseurl:
+                raise ConfigError("cas_config requires a public_baseurl to be set")
+
+            # If the service_url is given, use it for backwards compatibility.
+            # Otherwise, use the public_baseurl.
```
clokep, line 41: If we're requiring `public_baseurl` anyway, maybe we should just rip out support for this fallback? They should be the same anyway?

richvdh: yeah, probably.

synapse/config/sso.py
```diff
@@ -51,6 +49,22 @@ def read_config(self, config, **kwargs):
...
+        self.sso_redirect_confirm_template = None  # type: Optional[jinja2.Template]
+        if self.public_baseurl:
+            (
+                self.sso_login_idp_picker_template,
+                self.sso_redirect_confirm_template,
+            ) = self.read_templates(
+                ["sso_login_idp_picker.html", "sso_redirect_confirm.html"],
+                self.sso_template_dir,
+            )
+
```
richvdh, line 67: this feels a bit fragile, and obscure. Those templates may be the only ones that currently use `mxc_to_http` by default, but I can see us updating other templates to use it in the future, or people customising other templates and wondering why it mysteriously fails.

it might be better to arrange for `mxc_to_http` to always be defined (so that the template can be loaded), but then raise an exception at render time, with an explanation that `public_baseurl` is required in the error?

synapse/handlers/auth.py
```diff
@@ -1474,6 +1474,8 @@ def _complete_sso_login(
         # URL we redirect users to.
         redirect_url_no_params = client_redirect_url.split("?")[0]
 
+        assert self._sso_redirect_confirm_template is not None
+
```
richvdh, line 1478: if we're going down this route, it would be good to add some more comments on these assertions explaining why it is `Optional`, and why we don't expect it to be `None` (ie, why is the assertion safe?)

synapse/config/sso.py
clokep: I agree, I don't love it. A runtime error might make more sense, especially since we do always expect it to be configured when needed.

clokep: I made this change -- `mxx_to_http` will now raise a runtime error if `public_baseurl` is not set. It isn't expected that this actually happens since we do check in the config the situations where we expect it to be required.

richvdh (approved):
lgtm

clokep:
This seems to fail due to sytest having different config for public baseurl vs. the CAS service url:

* `public_baseurl`: http://localhost:8801
* `cas_config.service_url`: https://localhost:44349

Will see if I can get those to be the same...

PR #9317, opened 2021-02-04
Fix braces in OIDC sample config.

clokep:
This escapes the braces in the OIDC sample config so they're proper Jinja2 templates.


richvdh (approved):
doh

Issue #9318, opened 2021-02-04
Add 'raise X from e' in more places


ShadowJonathan:
While working on #9279 i noticed that a lot of exceptions raised often don't re-use the exception when raising;
```py
# this
try:
  # yada
except Exception:
  raise CustomException()

# instead of this
try:
  # yada daba doo
except Exception as e:
  raise CustomException() from e
```
The latter adds more context and a longer and more detailed traceback to look at.

The request;
- Add more of these in a lot of places (either `raise X from e`, or putting them into twisted `Failure` objects that're thrown somewhere)
  - If it won't be added to the re-raised exception, put down a message why in the comment (sensitive information, etc.)

~~- (Maybe) Make it a coding policy/find a linter to enforce raised exception usage, unless explicitly suppressed.~~

ShadowJonathan:
(im putting this here as an earmark for *maybe* myself to work on this after the PR for #9279 is finished and merged (or closed))

erikjohnston:
This happens automatically most of the time, from https://docs.python.org/3/tutorial/errors.html#exception-chaining:

> Exception chaining happens automatically when an exception is raised inside an except or finally section.

Though there are a bunch of places where we have a `try/except` within an `except`, where we probably *do* want to be explicit about which exception we're chaining from (i.e. we usually want to ignore any nested exception)

ShadowJonathan:
:thinking: Hmmm, I didn't know that.

Still, I think a thorough check should be done to see the messing about with `Failure`s I see sometimes, but i'll make a separate issue for that.

Note: I still think an explicit `from e` syntax is cleaner than letting it happen automatically, especially as sometimes the actual raising of exceptions isn't that idiomatic in this codebase, but i agree it shouldnt be a "hard and fast" rule then.

ShadowJonathan:
Just checked the twisted API docs, it looks `Failure` also works with capturing current exceptions, so I won't open an issue for that, but I still personally think explicit > implicit.

PR #9320, opened 2021-02-04
Add flake8-bugbear to the test suite, resolve resulting lint errors

### LINKED ISSUE #9279: Add flake8-bugbear to the test suite
Adding the `flake8-bugbear` plugin can help us catch bugs early before they occur, adding it on the codebase produces a enormous amount of the following codes;

- `B001 Do not use bare 'except:', it also catches unexpected events like memory errors, interrupts, system exit, and so on.  Prefer 'except Exception:'.  If you're sure what you're doing, be explicit and write 'except BaseException:'.`
- `B006 Do not use mutable data structures for argument defaults.  They are created during function definition time. All calls to the function reuse this one instance of that data structure, persisting changes between them.`
- `B007 Loop control variable 'encoded_signature' not used within the loop body. If this is intended, start the name with an underscore.`
- `B008 Do not perform function calls in argument defaults.  The call is performed only once at function definition time. All calls to your function will reuse the result of that definition-time function call.  If this is intended, assign the function call to a module-level variable and use that variable as a default value.`
- `B010 Do not call setattr with a constant attribute value, it is not any safer than normal property access.`
- `B306 'BaseException.message' has been deprecated as of Python 2.6 and is removed in Python 3. Use 'str(e)' to access the user-readable message. Use 'e.args' to access arguments passed to the exception.`

I think that's reasonable to take a closer look at where bugbear notes these linting errors, as it indeed can look likely for these bugs to exist there.

ShadowJonathan:
The scope of this PR is the following;
- adding `flake8-bugbear` to the `lint` extras
- fixing all resulting `flake8` errors
- adding various small fixes in the name of explicitness; (discussed in #9318)
  - `raise X from e`
  - `Failure(e)`

PR contains some comments on areas where wider intention isn't (as) clear.

### Pull Request Checklist

Closes #9279

### DIFF
--- a/synapse/app/__init__.py
+++ b/synapse/app/__init__.py
@@ -24,7 +24,7 @@ logger = logging.getLogger(__name__)
 try:
     python_dependencies.check_requirements()
 except python_dependencies.DependencyException as e:
-    sys.stderr.writelines(e.message)
+    sys.stderr.writelines(e.message)  # noqa: B306
     sys.exit(1)
 
 
--- a/synapse/crypto/context_factory.py
+++ b/synapse/crypto/context_factory.py
@@ -191,7 +191,7 @@ def _context_info_cb(ssl_connection, where, ret):
         # ... we further assume that SSLClientConnectionCreator has set the
         # '_synapse_tls_verifier' attribute to a ConnectionVerifier object.
         tls_protocol._synapse_tls_verifier.verify_context_info_cb(ssl_connection, where)
-    except BaseException as e:  # taken from the twisted implementation
+    except Exception as e:  # taken from the twisted implementation
         logger.exception("Error during info_callback", exc_info=e)
         f = Failure(e)
         tls_protocol.failVerification(f)
--- a/synapse/crypto/context_factory.py
+++ b/synapse/crypto/context_factory.py
@@ -191,9 +191,9 @@ def _context_info_cb(ssl_connection, where, ret):
         # ... we further assume that SSLClientConnectionCreator has set the
         # '_synapse_tls_verifier' attribute to a ConnectionVerifier object.
         tls_protocol._synapse_tls_verifier.verify_context_info_cb(ssl_connection, where)
-    except:  # noqa: E722, taken from the twisted implementation
-        logger.exception("Error during info_callback")
-        f = Failure()
+    except BaseException as e:  # taken from the twisted implementation
+        logger.exception("Error during info_callback", exc_info=e)
+        f = Failure(e)
         tls_protocol.failVerification(f)
 
 
@@ -219,7 +219,7 @@ class SSLClientConnectionCreator:
         # ... and we also gut-wrench a '_synapse_tls_verifier' attribute into the
         # tls_protocol so that the SSL context's info callback has something to
         # call to do the cert verification.
-        setattr(tls_protocol, "_synapse_tls_verifier", self._verifier)
+        tls_protocol._synapse_tls_verifier = self._verifier
         return connection
 
 
--- a/synapse/federation/units.py
+++ b/synapse/federation/units.py
@@ -18,6 +18,7 @@ server protocol.
 """
 
 import logging
+from typing import Optional
 
 import attr
 
@@ -98,7 +99,7 @@ class Transaction(JsonEncodedObject):
         "pdus",
     ]
 
-    def __init__(self, transaction_id=None, pdus: list = None, **kwargs):
+    def __init__(self, transaction_id=None, pdus: Optional[list] = None, **kwargs):
         """ If we include a list of pdus then we decode then as PDU's
         automatically.
         """
--- a/synapse/federation/units.py
+++ b/synapse/federation/units.py
@@ -98,7 +98,7 @@ class Transaction(JsonEncodedObject):
         "pdus",
     ]
 
-    def __init__(self, transaction_id=None, pdus=[], **kwargs):
+    def __init__(self, transaction_id=None, pdus: list = None, **kwargs):
         """ If we include a list of pdus then we decode then as PDU's
         automatically.
         """
@@ -107,7 +107,7 @@ class Transaction(JsonEncodedObject):
         if "edus" in kwargs and not kwargs["edus"]:
             del kwargs["edus"]
 
-        super().__init__(transaction_id=transaction_id, pdus=pdus, **kwargs)
+        super().__init__(transaction_id=transaction_id, pdus=pdus or [], **kwargs)
 
     @staticmethod
     def create_new(pdus, **kwargs):
--- a/synapse/handlers/device.py
+++ b/synapse/handlers/device.py
@@ -157,7 +157,7 @@ class DeviceWorkerHandler(BaseHandler):
             # The user may have left the room
             # TODO: Check if they actually did or if we were just invited.
             if room_id not in room_ids:
-                for key, event_id in current_state_ids.items():
+                for key, _ in current_state_ids.items():
                     etype, state_key = key
                     if etype != EventTypes.Member:
                         continue
@@ -180,7 +180,7 @@ class DeviceWorkerHandler(BaseHandler):
                 log_kv(
                     {"event": "encountered empty previous state", "room_id": room_id}
                 )
-                for key, event_id in current_state_ids.items():
+                for key, _ in current_state_ids.items():
                     etype, state_key = key
                     if etype != EventTypes.Member:
                         continue
@@ -199,7 +199,7 @@ class DeviceWorkerHandler(BaseHandler):
             for state_dict in prev_state_ids.values():
                 member_event = state_dict.get((EventTypes.Member, user_id), None)
                 if not member_event or member_event != current_member_id:
-                    for key, event_id in current_state_ids.items():
+                    for key, _ in current_state_ids.items():
                         etype, state_key = key
                         if etype != EventTypes.Member:
                             continue
@@ -713,7 +713,7 @@ class DeviceListUpdater:
                 # This can happen since we batch updates
                 return
 
-            for device_id, stream_id, prev_ids, content in pending_updates:
+            for device_id, stream_id, prev_ids, _ in pending_updates:
                 logger.debug(
                     "Handling update %r/%r, ID: %r, prev: %r ",
                     user_id,
@@ -739,7 +739,7 @@ class DeviceListUpdater:
             else:
                 # Simply update the single device, since we know that is the only
                 # change (because of the single prev_id matching the current cache)
-                for device_id, stream_id, prev_ids, content in pending_updates:
+                for device_id, stream_id, _, content in pending_updates:
                     await self.store.update_remote_device_list_cache_entry(
                         user_id, device_id, content, stream_id
                     )
--- a/synapse/handlers/device.py
+++ b/synapse/handlers/device.py
@@ -157,7 +157,7 @@ class DeviceWorkerHandler(BaseHandler):
             # The user may have left the room
             # TODO: Check if they actually did or if we were just invited.
             if room_id not in room_ids:
-                for key, _ in current_state_ids.items():
+                for key in current_state_ids.keys():
                     etype, state_key = key
                     if etype != EventTypes.Member:
                         continue
@@ -180,7 +180,7 @@ class DeviceWorkerHandler(BaseHandler):
                 log_kv(
                     {"event": "encountered empty previous state", "room_id": room_id}
                 )
-                for key, _ in current_state_ids.items():
+                for key in current_state_ids.keys():
                     etype, state_key = key
                     if etype != EventTypes.Member:
                         continue
@@ -199,7 +199,7 @@ class DeviceWorkerHandler(BaseHandler):
             for state_dict in prev_state_ids.values():
                 member_event = state_dict.get((EventTypes.Member, user_id), None)
                 if not member_event or member_event != current_member_id:
-                    for key, _ in current_state_ids.items():
+                    for key in current_state_ids.keys():
                         etype, state_key = key
                         if etype != EventTypes.Member:
                             continue
--- a/synapse/http/proxyagent.py
+++ b/synapse/http/proxyagent.py
@@ -14,6 +14,7 @@
 # limitations under the License.
 import logging
 import re
+from typing import Optional
 
 from zope.interface import implementer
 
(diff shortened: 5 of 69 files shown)


tests/rest/client/v1/utils.py
```diff
@@ -184,7 +184,13 @@ def send(self, room_id, body=None, txn_id=None, tok=None, expect_code=200):
         )
 
     def send_event(
-        self, room_id, type, content={}, txn_id=None, tok=None, expect_code=200
+        self,
+        room_id,
+        type,
+        content: Optional[dict] = None,
```
ShadowJonathan, line 190: Temporary, It isn't clear if this is `dict` or `Optional[dict]`, because a `None` could've been passed through, which would've then serialised below via `json.dumps` into `null`.

Is this type correct or not? (I'll edit this snippet to adjust)

tests/rest/client/v2_alpha/test_relations.py
```diff
@@ -628,7 +629,7 @@ def _send_relation(
         relation_type,
         event_type,
         key=None,
-        content={},
+        content: Optional[dict] = None,
```
ShadowJonathan, line 632: In the docstring below, `content` is possibly `None`, however the default value was always `{}` instead of `None`, should something be changed about this (the type, the docstring, etc.), or is this fine just as it is annotated right now?

synapse/util/async_helpers.py
```diff
@@ -485,8 +485,8 @@ def time_it_out():
 
         try:
             deferred.cancel()
-        except:  # noqa: E722, if we throw any exception it'll break time outs
-            logger.exception("Canceller failed during timeout")
+        except BaseException as e:  # if we throw any exception it'll break time outs
```
ShadowJonathan, line 488: Was `BaseException` always intended to be caught, or should it be `Exception`?

synapse/crypto/context_factory.py
```diff
@@ -191,9 +191,9 @@ def _context_info_cb(ssl_connection, where, ret):
         # ... we further assume that SSLClientConnectionCreator has set the
         # '_synapse_tls_verifier' attribute to a ConnectionVerifier object.
         tls_protocol._synapse_tls_verifier.verify_context_info_cb(ssl_connection, where)
-    except:  # noqa: E722, taken from the twisted implementation
-        logger.exception("Error during info_callback")
-        f = Failure()
+    except BaseException as e:  # taken from the twisted implementation
```
ShadowJonathan, line 194: Was `BaseException` always intended to be caught, or should it be `Exception`?

synapse/logging/context.py
```diff
@@ -683,10 +684,10 @@ def run_in_background(f, *args, **kwargs):
     current = current_context()
     try:
         res = f(*args, **kwargs)
-    except:  # noqa: E722
+    except BaseException as e:
```
ShadowJonathan, line 687: Was `BaseException` always intended to be caught, or should it be `Exception`?

tests/util/test_logcontext.py
```diff
@@ -75,8 +75,8 @@ def check_logcontext():
             try:
                 self.assertIs(current_context(), sentinel_context)
                 d2.callback(None)
-            except BaseException:
-                d2.errback(twisted.python.failure.Failure())
+            except BaseException as e:
```
ShadowJonathan, line 78: Was `BaseException` always intended to be caught, or should it be `Exception`?

ShadowJonathan (commented):
Left some comments where I need some input.

ShadowJonathan:
(Post-PR) Per #9279 I also mentioned `B90`, which is a subclass of "opinionated" linting violations from bugbear, as of commit 3ef7c6c, the following violations are reported;
```
synapse/rest/media/v1/_base.py:289:1: B903 Data class should either be immutable or use __slots__ to save memory. Use collections.namedtuple to generate an immutable class, or enumerate the attributes in a __slot__ declaration in the class to leave attributes mutable.
synapse/config/appservice.py:38:33: B902 Invalid first argument 'cls' used for instance method. Use the canonical first argument name in methods, i.e. self.
synapse/config/api.py:35:33: B902 Invalid first argument 'cls' used for instance method. Use the canonical first argument name in methods, i.e. self.
synapse/config/tracer.py:50:33: B902 Invalid first argument 'cls' used for instance method. Use the canonical first argument name in methods, i.e. self.
synapse/config/_base.py:34:1: B903 Data class should either be immutable or use __slots__ to save memory. Use collections.namedtuple to generate an immutable class, or enumerate the attributes in a __slot__ declaration in the class to leave attributes mutable.
synapse/replication/http/_base.py:129:36: B902 Invalid first argument **kwargs used for instance method. Use the canonical first argument name in methods, i.e. self.
tests/test_visibility.py:147:9: B901 Using `yield` together with `return x`. Use native `async def` coroutines or put a `# noqa` comment on this line if this was intentional.
tests/test_visibility.py:173:9: B901 Using `yield` together with `return x`. Use native `async def` coroutines or put a `# noqa` comment on this line if this was intentional.
tests/test_visibility.py:196:9: B901 Using `yield` together with `return x`. Use native `async def` coroutines or put a `# noqa` comment on this line if this was intentional.
tests/storage/test_state.py:77:9: B901 Using `yield` together with `return x`. Use native `async def` coroutines or put a `# noqa` comment on this line if this was intentional.
tests/test_utils/html_parsers.py:52:15: B902 Invalid first argument '_' used for instance method. Use the canonical first argument name in methods, i.e. self.
tests/rest/media/v1/test_url_preview.py:97:17: B902 Invalid first argument '_self' used for instance method. Use the canonical first argument name in methods, i.e. self.
tests/rest/client/test_transactions.py:49:13: B901 Using `yield` together with `return x`. Use native `async def` coroutines or put a `# noqa` comment on this line if this was intentional.
tests/util/caches/test_descriptors.py:231:21: B901 Using `yield` together with `return x`. Use native `async def` coroutines or put a `# noqa` comment on this line if this was intentional.
```

These aren't taken into account with how the linter is currently set up, for that i need to manually `select=` them via `setup.cfg`, which I might want to do in another PR.

ShadowJonathan:
(also before anyone comments on it, no, i cant make this PR smaller (than 200 lines), because i have to both add flake8-bugbear *and* fix all the small problems that pop up while going over the new linting suite errors)

tests/rest/client/v2_alpha/test_relations.py
ShadowJonathan: Test failures show that this should be `= {}`, docstring below is just what you can put in, which makes sense, i guess.

tests/rest/client/v1/utils.py
clokep: If it can be `None` (which this obviously can due to the default value) then it should be `Optional[dict]`.

clokep:
@ShadowJonathan When doing large repetitive changes like this it can help to do one of the following:

* Do not enable all of the checks at once and do them in separate PRs.
* Make the fixes in separate commits which clearly state what each commit is fixing ("Do not use mutable default arguments.", etc.)

Also note that a PR above a couple of hundred lines of diff generally takes much longer to review. Small PRs are better!

tests/rest/client/v1/utils.py
ShadowJonathan: The default value before was `{}`, bugbear bugged (heh) about this because `{}` is mutable and instantiated at compile-time, which can bring bugs, i changed it to `None` here to then do `content or {}` below, which effectively re-defaults `{}` for this parameter.

The question is; Should it? (`content or {}`), because before someone could still explicitly pass a `None` value, i'm just sanity-checking here if someone *should* pass a `None` value here at all, and so the annotation should be `dict` instead of `Optional[dict]` (because it's either the defaulted `{}` or another dict value, not and never `None`)

ShadowJonathan:
I *could* do this if there was a `select=` key for `flake8` in `setup.cfg`, which makes me able to enable sub-categories for each of the checks and then create PRs for each, but right now i'd rather want to present this one (which will take a second, which is fine for me) instead of making the `select=` PR first and then waiting for that one before making and chopping up this one.

(The problem is is that i cant "enable just some of the checks", `flake8` works either based on defaults with `ignore`d checks, or `select`ed checks (together with `ignore`d ones). 

I *could* make PRs that in parallel that have every other category explicitly `ignore`d other than the one that the PR is for, but that'd create a mess (you'd have to merge `ignore=B001,B002` with `ignore=B002,B003`, and then do this again with `ignore=B001,B003`. I could rebase every PR when previous ones are merged, but that'd be tedious and confusing, y'all have already indicated you don't want rebased PRs without reason.

Finally I could also then just bring those "small fix PRs" in serial, but that'd take a while to be fixed, i'd have to keep track of every separated change locally (which may slowly become out of date), and a bunch of other stuff that's just plain overhead.

If the `select=` was present, then i could just do `select=...,B001`, `select=...,B002`, `select=...,B003` PRs, which then are able to be merged together more nicely (`select=...,B001,B002`), and when done, a final PR could make it just the broad `B` category, but alas.)

TL;DR: Next time I'll make sure the small changes are better able to be applied into seperate PRs. :+1:

tests/rest/client/v1/utils.py
clokep: I don't believe that `None` is a valid value there.

synapse/util/async_helpers.py
clokep: These can probably all be `Exception`.

synapse/handlers/device.py
```diff
@@ -157,7 +157,7 @@ async def get_user_ids_changed(
             # The user may have left the room
             # TODO: Check if they actually did or if we were just invited.
             if room_id not in room_ids:
-                for key, event_id in current_state_ids.items():
+                for key, _ in current_state_ids.items():
```
clokep, line 160: Looks like this could also use `current_state_Ids.keys()`.

synapse/federation/units.py
```diff
@@ -98,7 +98,7 @@ class Transaction(JsonEncodedObject):
         "pdus",
     ]
 
-    def __init__(self, transaction_id=None, pdus=[], **kwargs):
+    def __init__(self, transaction_id=None, pdus: list = None, **kwargs):
```
clokep, line 101: In general when we have an option which does not need to be provided It should be set to a type of `Optional[foo]`. There's quite a few place that this should be updated in this PR.

synapse/app/__init__.py
```diff
@@ -24,7 +24,7 @@
 try:
     python_dependencies.check_requirements()
 except python_dependencies.DependencyException as e:
-    sys.stderr.writelines(e.message)
+    sys.stderr.writelines(e.message)  # noqa: B306
```
clokep, line 27: Should we update this to be `str(e)`?

synapse/handlers/device.py
ShadowJonathan: oh whoops, i did that a few more times, i didn't see this was an `.items` call, good catch :+1:

synapse/http/proxyagent.py
```diff
@@ -64,7 +64,7 @@ def __init__(
         self,
         reactor,
         proxy_reactor=None,
-        contextFactory=BrowserLikePolicyForHTTPS(),
+        contextFactory: BrowserLikePolicyForHTTPS = None,
```
clokep, line 67: This is the in wrong type, it should be `IPolicyForHTTPS`, I think.

synapse/app/__init__.py
ShadowJonathan: `DependencyException.message` is actually a property that formats it properly, thus the noqa, this is in a few other places too

synapse/federation/units.py
ShadowJonathan: If you could point that out, that'd be helpful, because I think otherwise i'd break some expectations, and i dont have a full knowledge where i can and where i cant

clokep: Everywhere it defaults to `None` should be `Optional` IMO.

ShadowJonathan: they defaulted to mutable constructors (`{}`, `[]`) before, so i assume they werent meant to be optional, but should i make it so?

clokep: That pretty much meant you did not need to provide it, we're now making that equivalent to `None`, so they're optional, yes.

synapse/http/proxyagent.py
ShadowJonathan: I'll also make it `Optional`

clokep:
Overall I like the direction this is heading; would you mind restructuring it so it's easier to review?

I think that many of the changes in here are good (I'm especially happy to see the change to stop using mutable input arguments), but this PR tries to do too many unrelated things and it hinders discussion of individual changes. I'm going to close this so that we can start afresh.

In particular I'm afraid of introducing regressions and it being difficult to bisect exactly what caused it. I would suggest splitting this work up into smaller pieces. A few ways this could be done:

## Strategy 1

* Split up the PR into separate types of changes (e.g. one for the mutable input arguments, one for using `keys()` instead of `items()`, etc.), this could be done by running `flake8-bugbear` locally and fixing the issues.
* After those have been merged, introduce `flake8-bugbear` and enable it. This might involve fixing some minor discrepancies that occurred in the meantime.

Depending on the size of each of the changes, some of them could be grouped together if each change is in separate commit with a clear commit message.

## Strategy 2

* Introduce `flake8-bugbear` and disable most of the checks.
* Fix the issues found and create a new PR (essentially a scaled down version of this).
* Create subsequent PRs (in series or in parallel) that fix remaining issues.

This will cause conflicts, but they will be easy to fix since it is just the single line of which checks are enabled. As with above, some of the changes could be grouped if they're done in separate commits.

ShadowJonathan:
I'm going to take a variant of strategy 2; i'll create a PR (like [mentioned before](https://github.com/matrix-org/synapse/pull/9320#issuecomment-773410951)) that'll explicitly enable checks in `setup.cfg` (on separate lines), once that's merged, i'll create a few PRs that all individually enable checks on `bugbear` and then also subsequently resolve their issues, i'll lead this with a PR that'll add `flake8-bugbear` to the dependencies, and enable the checks that do not cause issues. I'll trail the induvidual PRs with a final PR that'll condensate the `B001,B002,B003` into just `B` and the like.

Does that sound good?

clokep:
I think that sounds reasonable! Thanks for being understanding!

ShadowJonathan:
(note to self; make tracker issue from it)

PR #9326, opened 2021-02-04
Share the code for handling required attributes between CAS & SAML.

clokep:
CAS & SAML had very similar (almost identical) code for handling required attributes from the response. Also adds tests! Should be easy to add to OIDC in the future if necessary.

I think this was the last piece of code to abstract between the various SSO implementations.


richvdh (approved):
lgtm

PR #9342, opened 2021-02-06
Update delegate.md

OttoHollmann:
Maybe worth to explicitly mention to edit `public_baseurl` when using delegation.

### Pull Request Checklist
* [x]  Pull request is based on the develop branch
* [ ]  Pull request includes a [changelog file](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md#changelog). The entry should:
  
  * Be a short description of your change which makes sense to users. "Fixed a bug that prevented receiving messages from other servers." instead of "Moved X method from `EventStore` to `EventWorkerStore`.".
  * Use markdown where necessary, mostly for `code blocks`.
  * End with either a period (.) or an exclamation mark (!).
  * Start with a capital letter.
* [x]  Pull request includes a [sign off](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md#sign-off)
* [x]  Code style is correct (run the [linters](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md#code-style))

### DIFF
--- a/docs/delegate.md
+++ b/docs/delegate.md
@@ -51,6 +51,12 @@ However, if you really need it, you can find some documentation on how such a
 record should look like and how Synapse will use it in [the Matrix
 specification](https://matrix.org/docs/spec/server_server/latest#resolving-server-names).
 
+## Miscellaneous
+
+If you decided to use delegation, don't forget edit also your `homeserver.yaml`
+configuration and specify `public_baseurl` (e.g. `https://synapse.example.com/`)
+otherwise you will not be able to sign in.
+
 ## Delegation FAQ
 
 ### When do I need delegation?


richvdh (changes requested):
I might be missing something, but as far as I'm aware, the need for `public_baseurl` is unrelated to delegation?

OttoHollmann:
Let me describe steps which I done.
I've started with generating config files with `docker run --rm synapse generate` with env `SYNAPSE_SERVER_NAME=example.com` and verifyed that everything works well.
Then I've decided to setup delegation as described in `delegation.md`. Everything seemd to be OK except login, none of my client were able to login and after providing my credentials it freezed (I think with no error on server site). And then I spent more than hour debugging, why I can't login. Finally after setting up `public_baseurl` to `synapse.example.com` I was able to login.

So maybe it's unrelated, but I will appreciate it there will be some mention like in this PR. Or did I'm anything wrong and the recommended steps to setup delegation are different?

richvdh:
I just think it's unrelated to delegation, or indeed federation. You need a correct `public_baseurl` whether or not you are configuring federation, so it's a very odd thing to put in `delegation.md`.

It might be worth calling out the need for it in `INSTALL.md`, but on the other hand, #9313 should mean that it's no longer necessary to specify `public_baseurl`.

OttoHollmann:
Yes, it seems to be solved with PR you mentioned. But anyway, I think it is related, see:
https://github.com/matrix-org/synapse/blob/9dbfec0b15ef52508b0ddcd767b48a6edbc847fd/docs/sample_config.yaml#L1286

Or we can move it into FAQ section. It's up to you, you can close this issue it you don't want to merge this PR.

richvdh:
> But anyway, I think it is related, see:

That's about delegating 3pid validation, not federation.

Thank you very much for contributing, but I'm afraid I feel like the change will confuse people, so I'm going to go ahead and close this.

PR #9353, opened 2021-02-08
Implement a CachedCall to handle boilerplate of caching results

richvdh:
This feels like a relatively common pattern which is hard to get right.

I've used `oidc.jwks` as an example for how it could be used.

### DIFF
--- a/synapse/util/caches/cached_call.py
+++ b/synapse/util/caches/cached_call.py
@@ -1,0 +1,115 @@
+# -*- coding: utf-8 -*-
+# Copyright 2021 The Matrix.org Foundation C.I.C.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import Awaitable, Callable, Generic, Optional, TypeVar
+
+from twisted.internet.defer import Deferred
+
+from synapse.logging.context import make_deferred_yieldable, run_in_background
+
+TV = TypeVar("TV")
+
+
+class CachedCall(Generic[TV]):
+    """A wrapper for asynchronous calls whose results should be shared
+
+    This is useful for wrapping asynchronous functions, where there might be multiple
+    callers, but we only want to call the underlying function once (and have the result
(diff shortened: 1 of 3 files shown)


richvdh:
I'm interested in thoughts on this. I don't think the tests are passing; in any case, if it's a good idea, I want to add unit tests. But if the answer is "why don't you just ....", then I won't pursue it further.

synapse/util/caches/cached_call.py
```diff
@@ -0,0 +1,97 @@
...
+        """
+        Args:
+            f: The underlying function. Only one call to this function will be alive
+                at once (per instance of CachedCall)
+        """
+        self._callable = f  # type: Optional[Callable[[], Awaitable[TV]]]
+        self._deferred = None  # type: Optional[Deferred]
+        self._result = None  # type: Union[None, Failure, TV]
+
+    async def get(self) -> TV:
```
clokep, line 66: Maybe slighter nicer ergonomics if we used `__call__` instead?

This would let you do something like:

```python
async def handle_request() -> X:
    # We can call this multiple times, but it will result in a single call to
    # _load_prop().
    return await _cached_val()
```

This feels a bit nicer since you've wrapped a function and then it returns a function-like thing and might let you use it as a decorator?

synapse/util/caches/cached_call.py
```diff
@@ -0,0 +1,97 @@
...
+        #    and any eventual exception may not be caught.
+
+        # we can now await the deferred, and once it completes, return the result.
+        await make_deferred_yieldable(self._deferred)
+
+        # I *think* this is the easiest way to correctly raise a Failure without having
+        # to gut-wrench into the implementation of Deferred.
+        d = Deferred()
+        d.callback(self._result)
+        return await d
```
clokep, line 97: I'm not sure this will work properly for errors, do we need to call `d.errback` if it is an instance of `Failure`? (I don't think `Deferred` does anything special if you callback a `Failure`.)

synapse/util/caches/cached_call.py
richvdh: hrm, maybe. I always feel like relying on `__call__` is a bit magical, and tend to prefer that the interactions are made explicit, which is why I did it this way. I could be persuaded though.

clokep: My thought is that it makes it more like [`functools.cached_property`](https://docs.python.org/dev/library/functools.html#functools.cached_property), but perhaps that isn't the goal. 😄

synapse/util/caches/cached_call.py
richvdh: It does work, because `errback(x)` and `callback(x)` are both just thin wrappers around `_startRunCallbacks(x)`. You could argue that this isn't using Deferred's API as it's intended, which might be fair.

richvdh:
This is now superceded by #9362.

PR #9358, opened 2021-02-09
Fix #8518 (sync requests being cached wrongly on timeout)

### LINKED ISSUE #8518: Sync requests immediately return with empty payload
### Description

### Steps to reproduce

- **Homeserver**: matrix.org

ShadowJonathan:
This fixes #8518 by adding a conditional check on `SyncResult` in a function when `prev_stream_token == current_stream_token`, as a sanity check. In `CachedResponse.set.<remove>()`, the result is immediately popped from the cache if the conditional function returns "false".

This prevents the caching of a timed-out `SyncResult` (that has `next_key` as the stream key that produced that `SyncResult`). The cache is prevented from returning a `SyncResult` that makes the client request the same stream key over and over again, effectively making it stuck in a loop of requesting and getting a response immediately for as long as the cache keeps those values.

### Pull Request Checklist

Fixes #8518

### DIFF
--- a/changelog.d/9358.misc
+++ b/changelog.d/9358.misc
@@ -1,0 +1,1 @@
+Added a fix that invalidates cache for empty timed-out sync responses.
--- a/synapse/handlers/sync.py
+++ b/synapse/handlers/sync.py
@@ -40,7 +40,7 @@ from synapse.types import (
 from synapse.util.async_helpers import concurrently_execute
 from synapse.util.caches.expiringcache import ExpiringCache
 from synapse.util.caches.lrucache import LruCache
-from synapse.util.caches.response_cache import NoTimedCache, ResponseCache
+from synapse.util.caches.response_cache import ResponseCache
 from synapse.util.metrics import Measure, measure_func
 from synapse.visibility import filter_events_for_client
 
@@ -277,8 +277,9 @@ class SyncHandler:
         user_id = sync_config.user.to_string()
         await self.auth.check_auth_blocking(requester=requester)
 
-        res = await self.response_cache.wrap(
+        res = await self.response_cache.wrap_conditional(
             sync_config.request_key,
+            lambda result: since_token != result.next_batch,
             self._wait_for_sync_for_user,
             sync_config,
             since_token,
@@ -330,11 +331,6 @@ class SyncHandler:
             else:
                 lazy_loaded = "false"
             non_empty_sync_counter.labels(sync_type, lazy_loaded).inc()
-
-        # fixme hack: sanity check to invalidate cache, prevent a self-referral loop
-        #  replace with contextvars approach once it gets working on twisted (twisted/twisted#1262)
-        if since_token == result.next_batch:
-            result = NoTimedCache(result)  # type: ignore # because it is unwrapped in ResponseCache.set.<remove>()
 
         return result
 
--- a/synapse/handlers/sync.py
+++ b/synapse/handlers/sync.py
@@ -40,7 +40,7 @@ from synapse.types import (
 from synapse.util.async_helpers import concurrently_execute
 from synapse.util.caches.expiringcache import ExpiringCache
 from synapse.util.caches.lrucache import LruCache
-from synapse.util.caches.response_cache import ResponseCache
+from synapse.util.caches.response_cache import NoTimedCache, ResponseCache
 from synapse.util.metrics import Measure, measure_func
 from synapse.visibility import filter_events_for_client
 
@@ -330,6 +330,10 @@ class SyncHandler:
             else:
                 lazy_loaded = "false"
             non_empty_sync_counter.labels(sync_type, lazy_loaded).inc()
+
+        # hack: sanity check to invalidate cache, prevent a loop
+        if since_token == result.next_batch:
+            result = NoTimedCache(result)  # type: ignore # because it is unwrapped in ResponseCache.set.<remove>()
 
         return result
 
--- a/synapse/handlers/sync.py
+++ b/synapse/handlers/sync.py
@@ -331,7 +331,8 @@ class SyncHandler:
                 lazy_loaded = "false"
             non_empty_sync_counter.labels(sync_type, lazy_loaded).inc()
 
-        # hack: sanity check to invalidate cache, prevent a loop
+        # fixme hack: sanity check to invalidate cache, prevent a self-referral loop
+        #  replace with contextvars approach once it gets working on twisted (twisted/twisted#1262)
         if since_token == result.next_batch:
             result = NoTimedCache(result)  # type: ignore # because it is unwrapped in ResponseCache.set.<remove>()
 
--- a/synapse/util/caches/response_cache.py
+++ b/synapse/util/caches/response_cache.py
@@ -114,7 +114,7 @@ class ResponseCache(Generic[T]):
                 )
             else:
                 self.pending_result_cache.pop(key, None)
-            return r
+            return r.obj if isinstance(r, NoTimedCache) else r
 
         result.addBoth(remove)
         return result.observe()
--- a/synapse/util/caches/response_cache.py
+++ b/synapse/util/caches/response_cache.py
@@ -13,9 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import logging
-from typing import TYPE_CHECKING, Any, Callable, Dict, Generic, Optional, TypeVar
-
-import attr
+from typing import TYPE_CHECKING, Any, Callable, Dict, Generic, Optional, Set, TypeVar
 
 from twisted.internet import defer
 
@@ -31,11 +29,6 @@ logger = logging.getLogger(__name__)
 T = TypeVar("T")
 
 
-@attr.s(slots=True, frozen=True)
-class NoTimedCache:
-    obj = attr.ib()
-
-
 class ResponseCache(Generic[T]):
     """
     This caches a deferred response. Until the deferred completes it will be
@@ -47,6 +40,7 @@ class ResponseCache(Generic[T]):
     def __init__(self, hs: "HomeServer", name: str, timeout_ms: float = 0):
         # Requests that haven't finished yet.
         self.pending_result_cache = {}  # type: Dict[T, ObservableDeferred]
+        self.pending_conditionals = {}  # type: Dict[T, Set[Callable[[Any], bool]]]
 
         self.clock = hs.get_clock()
         self.timeout_sec = timeout_ms / 1000.0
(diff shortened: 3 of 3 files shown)


ShadowJonathan:
I could make this better by using `contextvars`, but i've heard that so far it does not correctly work with twisted, so i'm not doing that now, and instead this could be replaced with `contextvars` when it *does* work.

ShadowJonathan:
(Note: i am fully aware this is kind-of a hack, and i'm prodding with this PR to see if this hack is acceptable for the problem)

erikjohnston:
Thanks for digging into the bug. I think your approach is probably fine, it doesn't really fill me with joy but oh well. Perhaps one refinement/alternative would be to have `ResponseCache` take a `should_cache: Callable[[T], bool]`  arg, which is called on every returned result to see if it should be cached? The default would effectively be `lambda _: True` (i.e. cache everything), and for the sync cache could either be `lambda r: since_token != r.next_batch` or even just `bool` (as the sync results are falsey if they're empty, and we may simply just not want to cache empty results).

Jonathon/et al.: Thoughts?

(This will also need a test)

ShadowJonathan:
> (as the sync results are falsey if they're empty, and we may simply just not want to cache empty results)

I don't wanna bet on this, but i'll look at `should_cache`, thanks for the in-depth reply!

I'll also look at making a test somewhere

ShadowJonathan:
I made it so that `ResponseCache` has a second method; `wrap_conditional`, which expects a position argument of a callable, which all have to return `True` for it to actually be cached in a timed fashion.

ShadowJonathan:
@erikjohnston there seems to be no `test_responsecache` in `tests.util.caches`, do you;

a. want me to make one, and *only* add a test for the conditional.
b. want me to make one, add a test for the conditional, and then some other tests (e.g. timeout and actually checking if it's cached or not)
c. do nothing for now, have the tests arrive in another PR later

?

erikjohnston:
Ugh, I really thought there were tests already. I think adding tests should be easy enough, so if you could do that then that would be great. Some simple tests plus some tests of the new logic would be fab, assuming that by the time you've done the tests for the new logic it'll be trivial to do some simple tests as well.

ShadowJonathan:
highlighting comment to commit here to support changes;
> ```make first argument to ResponseCache synapse.util.Clock instead of HomeServer, to make more accessible to tests
> + as a result, hs.get_clock() needs to be added in all places
> + and in some places, HomeServer type annotation should be added
> ```

ShadowJonathan:
(should i add `, add tests for ResponseCache.` to the newspaper?)

synapse/federation/federation_server.py
```diff
@@ -450,6 +451,7 @@ async def _on_state_ids_request_compute(self, room_id, event_id):
     async def _on_context_state_request_compute(
         self, room_id: str, event_id: str
     ) -> Dict[str, list]:
+        pdus = []  # type: Iterable[EventBase]
```
ShadowJonathan, line 454: this is needed, because else mypy will complain about the `pdu =` in the second branch not being `List[EventBase]`

ShadowJonathan:
Some conversation in #synapse-dev uncovered that adding tests in this PR probably isn't the best option, so i'll make another PR that depends on this one, and adds tests for ResponseCache

ShadowJonathan:
(the failing sytest test on the latest commit is flaky, btw, can someone restart that?)

anoadragon453 (approved):
This looks fine for me, thanks for clearly describing what caused the bug in the description!

ShadowJonathan:
Oh, the description is actually outdated right now, let me update that.

anoadragon453:
Thanks! Merging as the tests in https://github.com/matrix-org/synapse/pull/9458 are passing.

changelog.d/9358.misc
```diff
@@ -0,0 +1 @@
+Added a fix that invalidates cache for empty timed-out sync responses.
```
richvdh, line 1: shouldn't this have been a `.bugfix`?

synapse/handlers/sync.py
```diff
@@ -277,8 +277,9 @@ async def wait_for_sync_for_user(
         user_id = sync_config.user.to_string()
         await self.auth.check_auth_blocking(requester=requester)
 
-        res = await self.response_cache.wrap(
+        res = await self.response_cache.wrap_conditional(
             sync_config.request_key,
+            lambda result: since_token != result.next_batch,
```
richvdh, line 282: this should have had a comment. Why does doing this give the right behaviour? (I shouldn't have to go and find the PR that added it, and even having done so it's hard to correlate the description to the code.)

changelog.d/9358.misc
ShadowJonathan: shoot, yes, could you change the file extension in the `develop` branch?

synapse/util/caches/response_cache.py
```diff
@@ -112,6 +117,31 @@ def remove(r):
         result.addBoth(remove)
         return result.observe()
 
+    def add_conditional(self, key: T, conditional: Callable[[Any], bool]):
+        self.pending_conditionals.setdefault(key, set()).add(conditional)
```
richvdh, line 121: this seems like a dangerous thing to add to the public interface. It seems like it would be very easy to use in racy way.

ShadowJonathan: should i signal that `add_conditional` is private by prefixing it with `_` (in a new PR)?

synapse/handlers/sync.py
ShadowJonathan: It was described in the PR, but i'll take it in another PR to add a comment here, i.e. "small fixes for `ResponseCache`"

synapse/util/caches/response_cache.py
richvdh: since it's only called in one place, I'd just inline it.

synapse/handlers/sync.py
richvdh: > It was described in the PR description,

yeah as I said: having to go and find the PR and grok it is a pain.

synapse/util/caches/response_cache.py
```diff
@@ -112,6 +117,31 @@ def remove(r):
...
+        """The same as wrap(), but adds a conditional to the final execution.
+
+        When the final execution completes, *all* conditionals need to return True for it to properly cache,
+        else it'll not be cached in a timed fashion.
+        """
+
+        # See if there's already a result on this key that hasn't yet completed. Due to the single-threaded nature of
+        # python, adding a key immediately in the same execution thread will not cause a race condition.
+        result = self.get(key)
+        if not result or isinstance(result, defer.Deferred) and not result.called:
```
richvdh, line 140: I think this is correct, but I had to go and look up the precedence of `and` vs `or` to check it. Please use parens to disambiguate in future.

synapse/util/caches/response_cache.py
```diff
@@ -112,6 +117,31 @@ def remove(r):
...
+            self.add_conditional(key, should_cache)
```
richvdh, line 141: is it correct that the conditional is added when `not result.called` ?  I'd argue that cachability of the result should be determined by the `callback` being used, rather than any subsequent `callbacks` (which are discarded).

richvdh, line 141: I feel like this is duplicating *half* the logic of `wrap`, which is kinda the worst of all worlds. Could you not make `should_cache` nullable, and have `wrap` call `wrap_conditional` rather than the other way around?

richvdh (commented):
sorry to write a bunch of comments after the event; I wanted to review it in the light of #9507 and these were a few things I spotted.

TL;DR is that this could be easier to grok but I can't see any obvious bugs here.

synapse/util/caches/response_cache.py
ShadowJonathan: operator precedence is `or` -> `and` -> `not`, so this is actually
`((not result) or isinstance(result, defer.Deferred)) and (not result.called)`

I had to check after your previous comment, it could actually be `if not result and not result.called`, but because `result` can be a plain return value, i need to insert the check to `isinstance(..., defer.Deferred)`

Yes, the conditional is added to be executed after the result is called, i do agree that now i realise that any future calls can be like "hey, thjis shouldn't be cached anyways because the returned value doesn't agree with what i already got from somewhere else", in which case I need to rewrite this to execute the conditional locally whenever it's called, and evict the cache when it's not valid according to the new conditional. (the `callLater` already taking place would simply execute anyways, no way to properly garbage collect that at that point to my knowledge, if it was an asyncio `Task`, i could also map it somewhere to then `.cancel()` it, if this is possible with `callLater` handles or something, please tell me)

ShadowJonathan: I initially did that, but that became a bit of a mess, and you'd had to shift around the positional arguments on *any* `wrap` calls currently in place.

I also didn't wanna make it a keyword argument, because that'd shadow any potential keyword arguments to the inner call. While i know it is trivial to "just add it", i didn't want to because it is non-explicit to anyone not aware of this change to `ResponseCache`, and so it is possible for it to become a bug.

richvdh: > I initially did that, but that became a bit of a mess, and you'd had to shift around the positional arguments on any wrap calls currently in place.

I'm not suggesting changing the signature of `wrap`.

Wrap would be:

```python
    def wrap(
        self, key: T, callback: "Callable[..., Any]", *args: Any, **kwargs: Any
    ) -> defer.Deferred:
        return self.wrap_conditional(key, None, callback, *args, **kwargs)
```

why is that a mess?

richvdh: as discussed at considerable length in the room: I don't agree with your logic here. I think that supporting multiple `should_cache` callbacks per key significantly complicates the implentation, and semantically is dubious at best.

ShadowJonathan: hmmm, i'll change it to that in that PR, then :+1:

ShadowJonathan: So i can do three things here (sorry to repeat again);
1. Only the *very first* call will add the conditional (replacing the `dict[T, set[Callable]]` with just a plain `dict[T, Callable]` mapping).
2. The first callers (to call `wrap` before the call completes) will be able to contribute a `conditional`, this is how it's currently implemented.
3. Any subsequent call to `wrap_conditional`, even after the result has been set, can veto and evict the cache.

I didn't get a clear signal if it should be `1` or `2`, that's why i'm asking again.

synapse/handlers/sync.py
ShadowJonathan: Alright, i'll add a comment describing what that conditional does :+1:

synapse/util/caches/response_cache.py
ShadowJonathan: Alright, i'll just make it a simple `Dict[T, Callable]` mapping then

changelog.d/9358.misc
ShadowJonathan: Ah, it's already suggested in #9522 to include it there, i'll resolve this then :+1:

synapse/util/caches/response_cache.py
ShadowJonathan: Actually, more discussion in #9522 on this.

clokep:
I've backed this out in aee10768d8a9bf6c3df80859d2c135607aa1bf80. I think we'll want to:

* Re-land this with the additional edge-case handled.
* Ensure the tests cover the additional edge-case.

I think the current known issue of #8518 is better than the unknown impact of the regression issue #9507. Unfortunately with this change we did not have confidence in creating a new release candidate or deploying to matrix.org, which are necessities for the team.

PR #9361, opened 2021-02-09
Remove unneeded type constraints on 3rd party protocol lookup responses

babolivier:
Synapse checks that the `fields` object in 3rd party protocol lookup responses from app services only includes string values, but [the spec](https://matrix.org/docs/spec/application_service/r0.1.2#get-matrix-app-v1-thirdparty-user-protocol) says there's no such constraint.

### DIFF
--- a/changelog.d/9361.bugfix
+++ b/changelog.d/9361.bugfix
@@ -1,0 +1,1 @@
+Fix a bug causing Synapse to impose the wrong type constraints on fields when processing responses from appservices to `/_matrix/app/v1/thirdparty/user/{protocol}`.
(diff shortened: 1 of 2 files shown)


clokep (approved):
Seems reasonable.

clokep:
I debated briefly if we should assert it a str / int / bool / `None` or something, but looks the spec just says it is anything...

Hopefully other bits of our code don't assume that the fields are a string?

babolivier:
> Hopefully other bits of our code don't assume that the fields are a string?

When testing my appservice (which relies on Synapse passing through lookup results including boolean values in the fields), I had no result with this check in Synapse, and was getting results passed through when I removed it. On top of this `query_3pe` is only in two code path and in both cases we pass through the appservice's response directly (with only the checks in `query_3pe`). So I believe it's the only place we do this assumption.

PR #9362, opened 2021-02-09
Clean up caching/locking of OIDC metadata load

richvdh:
There are a few commits here, with the broad goal of cleaning up the way we construct the metadata cache. Sorry it won't be particularly easy to see what's going on - I couldn't find a way to break it down to make it much more obvious.

### DIFF
--- a/synapse/handlers/oidc_handler.py
+++ b/synapse/handlers/oidc_handler.py
@@ -246,6 +246,7 @@ class OidcProvider:
 
         self._token_generator = token_generator
 
+        self._config = provider
         self._callback_url = hs.config.oidc_callback_url  # type: str
 
         self._scopes = provider.scopes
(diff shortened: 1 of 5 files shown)


synapse/handlers/oidc_handler.py
```diff
@@ -363,30 +360,48 @@ def _uses_userinfo(self) -> bool:
...
+
+    async def _load_metadata(self) -> OpenIDProviderMetadata:
+        # init the metadata from our config
+        metadata = OpenIDProviderMetadata(
+            issuer=self._config.issuer,
+            authorization_endpoint=self._config.authorization_endpoint,
+            token_endpoint=self._config.token_endpoint,
+            userinfo_endpoint=self._config.userinfo_endpoint,
+            jwks_uri=self._config.jwks_uri,
+        )  # type: OpenIDProviderMetadata
```
clokep, line 393: I'm surprised you need to specify the type here?

clokep (approved):
I think this looks OK.

synapse/handlers/oidc_handler.py
richvdh: so am I. removed.

Issue #9366, opened 2021-02-10
[tracker] flake8-bugbear progression


ShadowJonathan:
This is a tracker issue for adding [flake8-bugbear](https://pypi.org/project/flake8-bugbear/) in short steps, and to keep track of it's progress.

The original PR was #9320, it has been split in the following PRs;

- [x] ~~Add `flake8-bugbear` dependency, put `B00` in ignore list, and fix remaining violations (#9499)~~
- [x] Finish check-fixing PRs (in order of altering `ignore` keys): 
  - [x] ~~`B00` -> `B006,B007,B008`: #9659~~
  - [x] ~~`B006,B007,B008` -> `B007`: #9682~~
  - [x] ~~`B007` -> `""`: #9838~~
- [ ] Discuss and merge making flake8 explicitly `select` checks (issue #XXXX, PR #XXXX)
  - [ ] Add one `B` code on `select=` (#XXXX)
  - [ ] Enable opinionated checks (`B90`) (#XXXX)
- [ ] (Side-quest) Update bugbear to [21.4](https://github.com/PyCQA/flake8-bugbear/releases/tag/21.4.1) (which adds a new check) (#XXXX)

(Some more information in #9279 about some specifics of these checks)

(#9370 was originally part of this tracker issue, but was denied, [some discussion followed](https://matrix.to/#/!XaqDhxuTIlvldquJaV:matrix.org/$Rx9vt5oTauoF21aNQhfSP67FN5RstYhqO6aiUXp_C2c?via=matrix.org&via=vector.modular.im&via=envs.net))

ShadowJonathan:
Optionally i'd also like to enable `B90`, (as said in #9279), or a select few of these checks, these can happen after the second PR, but just to note that.

ShadowJonathan:
(made #9368 and 9369 to be deliberated for after the first PR)

clokep:
@ShadowJonathan Looks like a lot of this is done! 🎉 Any update on here that we should add?

ShadowJonathan:
Already added all those updates, next step is "Discuss and merge making flake8 explicitly select checks", i'll make a properly-worded Process issue for that whenever i have time, currently bugbear is "there", but i'd like to also enable some of those "opinionated" checks.

Additionally, I'd wanna update bugbear to the latest version, to fix the newest checks that were added, lemme add that one now...

clokep:
[...]

PR #9370, opened 2021-02-10
Explicitly add flake8 selections

ShadowJonathan:
### Pull Request Checklist

Part of #9366

### DIFF
--- a/changelog.d/9370.misc
+++ b/changelog.d/9370.misc
@@ -1,0 +1,1 @@
+Explicitly select flake8 checks (as opposed to the defaults).
--- a/setup.cfg
+++ b/setup.cfg
@@ -25,7 +25,7 @@ select=
     # https://github.com/adamchainz/flake8-comprehensions#rules
     C4
 
-# Ignored checks take precedence above selected checks.
+# Ignored checks take precedence over selected checks.
 # Please also include a description as to why to include these, if possible.
 ignore=
     # "line break before binary operator"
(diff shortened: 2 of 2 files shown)


ShadowJonathan:
I used `flake8 setup.py -vvvv 2> flake8.txt` and https://flake8.pycqa.org/en/latest/user/violations.html#selecting-violations-with-flake8 to determine which selections are default, apart from flake8's default, `C4` (`flake8-comprehensions`) was the only other check that was turned on by default.

clokep:
I'm a bit confused by this -- I thought the goal was to enable additional checks only for flake8-bugbear? If that's not possible, then I think we'll need to ignore all the ones that don't pass yet.

ShadowJonathan:
I outlined it [here](https://github.com/matrix-org/synapse/pull/9320#issuecomment-773410951), but TL;DR: It's easier to individually add the checks to `select=` rather than remove them from `ignore=` one-by-one, the latter is more susceptible to merge conflicts, and it's easier to make git figure out the diff being "ah, just adding some lines here" rather than "oh, i need to remove this line, and this line, and this line...".

It's also simply easier to work with, and easier to observe explicitly enabled > implicitly enabled, i already created some other issues which touched on those explicitly-disabled checks, to create discussion to be able to remove them.

clokep:
I'm not too keen to merge this as I'm not a huge fan of disabling these checks (i.e. if we never finish a project to enable bugbear and new checks get added to flake8 then we'll be missing out). I think the approach should be to disable just the bugbear checks and start fixing those issues individually.

PR #9374, opened 2021-02-10
Finish adding type hints to the HomeServer

clokep:
There were a few spots that we didn't have return types on `HomeServer` this adds them! Unfortunately there's a few spots where we return different types, which makes the code rather messy. I'm not 100% sure this is worth it, although it did actually find a couple of discrepancies.


synapse/handlers/e2e_keys.py
```diff
@@ -226,16 +229,23 @@ async def do_remote_query(destination):
...
                             user_id=user_id
                         )
 
-                    user_devices = user_devices["devices"]
+                    if user_device_resync is None:
+                        failures[destination] = {
+                            "status": 503,
+                            "message": "Unable to resync devices",
+                        }
+                        continue
```
clokep, line 246: It seems that this method assumes that `user_device_resync` will raise if the host cannot be reached, but it actually eats errors and returns `None` instead. 😢

synapse/handlers/set_password.py
```diff
@@ -33,9 +33,7 @@ class SetPasswordHandler(BaseHandler):
     def __init__(self, hs: "HomeServer"):
         super().__init__(hs)
         self._auth_handler = hs.get_auth_handler()
-        device_handler = hs.get_device_handler()
-        assert isinstance(device_handler, DeviceHandler)
-        self._device_handler = device_handler
+        self._device_handler = hs.get_device_handler()
```
erikjohnston: So something was trying to construct this on a worker? Can we make it so that that doesn't happen?

erikjohnston (commented):
LGTM apart from the one Q

synapse/handlers/set_password.py
clokep: I'll look back into it!

clokep:
Deleting this since it is easier to do it in bits and pieces.

PR #9376, opened 2021-02-10
Support for form_post in OIDC responses

### LINKED ISSUE #9319: support for `form_post` response mode for OAuth2
https://openid.net/specs/oauth-v2-form-post-response-mode-1_0.html defines a `form_post` response mode for returning Authorization Response parameters from the Authorization Endpoint. This mode is required by Apple when requesting non-default scopes (https://developer.apple.com/documentation/sign_in_with_apple/sign_in_with_apple_js/incorporating_sign_in_with_apple_into_other_platforms), which we believe is required to enable FaceID/TouchID.

I think that supporting `form_post` is pretty trivial; we should do so.

richvdh:
Apple want to POST the OIDC auth response back to us rather than using query-params; add the necessary support to make that work.

Fixes #9319.

### DIFF
--- a/synapse/handlers/oidc_handler.py
+++ b/synapse/handlers/oidc_handler.py
@@ -47,8 +47,26 @@ if TYPE_CHECKING:
 
 logger = logging.getLogger(__name__)
 
-SESSION_COOKIE_NAME = b"oidc_session"
-SESSION_COOKIE_NAME_IOS_HACK = b"oidc_session_no_samesite"
+# we want the cookie to be returned to us even when the request is the POSTed
+# result of a form on another domain, as is used with `response_mode=form_post`.
+#
+# Modern browsers will not do so unless we set SameSite=None; however *older*
+# browsers (including all versions of Safari on iOS 12?) don't support
+# SameSite=None, and interpret it as SameSite=Strict:
+# https://bugs.webkit.org/show_bug.cgi?id=198181
+#
+# As a rather painful workaround, we set *two* cookies, one with SameSite=None
+# and one with no SameSite, in the hope that at least one of them will get
+# back to us.
+#
+# Secure is necessary for SameSite=None (and, empirically, also breaks things
+# on iOS 12.)
+#
+# Here we have the names of the cookies, and the options we use to set them.
+_SESSION_COOKIES = [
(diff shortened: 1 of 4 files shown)


synapse/handlers/oidc_handler.py
```diff
@@ -149,26 +150,29 @@ async def handle_oidc_callback(self, request: SynapseRequest) -> None:
...
-        )
+        # Removing the cookies early avoids spamming the provider with token requests.
+        for cookie_name, options in [
+            (SESSION_COOKIE_NAME, b"; SameSite=None"),
+            (SESSION_COOKIE_NAME_IOS_HACK, b""),
+        ]:
+            request.cookies.append(
+                b"%s=; Path=/_synapse/client/oidc; Expires=Thu, Jan 01 1970 00:00:00 UTC; "
+                b"HttpOnly; Secure%s" % (cookie_name, options)
+            )
```
clokep, line 175: This is to get around Twisted not supporting setting `SameSite=None`?

synapse/handlers/oidc_handler.py
```diff
@@ -693,14 +697,33 @@ async def handle_redirect_request(
...
+        #
+        # Secure is necessary for SameSite=None (and, empirically, also breaks things
+        # on iOS 12.)
+        #
+        # we have to build the cookie by hand rather than calling request.addCookie
+        # to work around https://twistedmatrix.com/trac/ticket/10088
+        #
+        for cookie_name, options in [
+            (SESSION_COOKIE_NAME, b"; Secure; SameSite=None"),
+            (SESSION_COOKIE_NAME_IOS_HACK, b""),
```
clokep, line 721: These don't seem to match above where we clear out the cookies. In that case *both* are set to `Secure`.

We should probably pull these out to constants.

clokep (changes requested):
Looks OK based on the previous conversations we had about the security aspects of this, I think there's some mismatches though. See the inline comments.

synapse/handlers/oidc_handler.py
richvdh: yes.

synapse/handlers/oidc_handler.py
richvdh: Ooops. fair.

clokep (approved):
LGTM!

PR #9377, opened 2021-02-10
Convert test cases to use HomeserverTestCase

clokep:
This converts some tests to use `HomeserverTestCase` (and stop using `inlineCallbacks`). Tests only change so should be pretty OK if tests pass. 👍 

### DIFF
--- a/tests/handlers/test_auth.py
+++ b/tests/handlers/test_auth.py
@@ -16,28 +16,21 @@ from mock import Mock
 
 import pymacaroons
 
-from twisted.internet import defer
-
-import synapse
-import synapse.api.errors
-from synapse.api.errors import ResourceLimitError
+from synapse.api.errors import AuthError, ResourceLimitError
 
 from tests import unittest
 from tests.test_utils import make_awaitable
-from tests.utils import setup_test_homeserver
 
 
-class AuthTestCase(unittest.TestCase):
-    @defer.inlineCallbacks
-    def setUp(self):
-        self.hs = yield setup_test_homeserver(self.addCleanup)
-        self.auth_handler = self.hs.get_auth_handler()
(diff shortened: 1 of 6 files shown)


tests/handlers/test_auth.py
```diff
@@ -76,29 +67,25 @@ def verify_nonce(caveat):
...
         )
         self.assertEqual("a_user", user_id)
 
         # when we advance the clock, the token should be rejected
-        self.hs.get_clock().now = 6000
-        with self.assertRaises(synapse.api.errors.AuthError):
-            yield defer.ensureDeferred(
-                self.auth_handler.validate_short_term_login_token_and_get_user_id(token)
-            )
+        self.reactor.advance(6)
```
clokep, line 78: `advance` takes seconds, so 6000 -> 6.

tests/handlers/test_auth.py
```diff
@@ -109,110 +96,98 @@ def test_short_term_login_token_cannot_replace_user_id(self):
...
+            self.auth_handler.validate_short_term_login_token_and_get_user_id(
+                self._get_macaroon().serialize()
+            ),
+            ResourceLimitError,
+        )
 
-    @defer.inlineCallbacks
     def test_mau_limits_parity(self):
+        # Ensure we're not at the unix epoch.
+        self.reactor.advance(1)
```
clokep, line 146: I think this might actually be due to a bug in `AuthBlocking` where the timestamp isn't checked against `None`:

https://github.com/matrix-org/synapse/blob/7932d4e9f7e140ef05a3099e18120101019ec3e1/synapse/api/auth_blocking.py#L108-L110

This would only ever show up in tests or if we find a DeLorean to go back to 1970.

richvdh (approved):
lgtm!

PR #9379, opened 2021-02-10
Move getEvent cache to externalCache

Half-Shot:
This PR moves the getEvent cache to the external cache class (backed by redis). ~We could potentially have a two level cache where the process still has a smaller LRU, but this PR at least ensures there is one cache for events for all workers :)~

This PR implements Redis as a level 2 event cache, where the existing event cache serves as a level 1 cache.


synapse/storage/databases/main/events_worker.py
```diff
@@ -491,9 +492,12 @@ async def _get_events_from_cache_or_db(self, event_ids, allow_rejected=False):
         return event_entry_map
 
     def _invalidate_get_event_cache(self, event_id):
+        if self._external_cache.is_enabled():
+            self._external_cache.delete("getEvent", event_id)
```
richvdh, line 496: you have to `await` a coroutine for it to actually do anything.

richvdh (commented):
This looks exciting!

I suspect a two-layer cache will be essential though: we end up fetching some events a lot of times, and you don't necessarily want to be deserialising them every time (not least because that could mean you end up with many copies of the same event in memory at the same time, which could backfire)

also, you might want to consider storing redactions separately to events, so that the entries are immutable and you don't have to worry about cache invalidation.

synapse/storage/databases/main/events_worker.py
Half-Shot: I've since made this a background process, as we depend on this function to be sync all over the place. I am a bit worried that doing this might break areas of Synapse that rely on this cache to be accurate though?

Half-Shot:
>I suspect a two-layer cache will be essential though: we end up fetching some events a lot of times, and you don't necessarily want to be deserialising them every time (not least because that could mean you end up with many copies of the same event in memory at the same time, which could backfire)

I've ended up changing the logic so we use the existing memory cache first, and fall back to the external cache. The idea being that if you want to save on worker memory, you tune down your event_cache_size which controls your L1 cache and Redis can be an L2 cache. If you don't jhave Redis support, you only get the L1 cache.

>also, you might want to consider storing redactions separately to events, so that the entries are immutable and you don't have to worry about cache invalidation.

Ah, so the reason we invalidate the cache is to then populate it with another entry that includes the redacted_event? If so, then two caches definitely make more sense to me. Would also solve the racyness of an async delete.

 >Configurable cache size / limit
 
I did a bit of homework on this. It sounds like the way most people use Redis is that they set it's memory limit, and applications basically fill the cache while Redis handles evicting entries to keep it under the memory limit. With this in mind, rather than adding complicated logic to Synapse to keep the Redis cache a safe size, we should probably ask administrators to configure their Redis cache to evict based on lfu (least-frequently-used), and to set a sensible memory limit.

For reference, I left half-shot.uk running for a bit overnight and Redis has grown to 1.2GB bearing in mind that I've not setup any cache eviction settings on it yet.

synapse/storage/databases/main/events_worker.py
Half-Shot: Following on from https://github.com/matrix-org/synapse/pull/9379#issuecomment-777386106, we should probably use two caches.

synapse/storage/databases/main/events_worker.py
```diff
@@ -492,8 +494,46 @@ async def _get_events_from_cache_or_db(self, event_ids, allow_rejected=False):
 
     def _invalidate_get_event_cache(self, event_id):
         self._get_event_cache.invalidate((event_id,))
+        if self._external_cache.is_enabled():
+            # XXX: Is there danger in doing this?
+            # We could hold a set of recently evicted keys in memory if
+            # we need this to by syncronous?
```
jaller94, line 500:
```suggestion
            # we need this to be synchronous?
```

Half-Shot:
Having spoken with @richvdh in #synapse-dev, we'd like to avoid cache invalidation so I've hunted down the areas where we invalidate:

Usages of `EventsWorkerStore._invalidate_get_event_cache:
- `PersistEventSore._update_room_depths_txn` (can we just mutate the event in the cache?)
- `Store._store_redaction` (store redacted event in a seperate cache?)
- `CacheInvalidationWorkerStore._invalidate_caches_for_event`
  - Called for backfilling ("We fetched some old events and either we had never seen that event before or it went from being an outlier to not.")
    Do we need to invalidate?
  - Called for new events ("We received a new event, or an event went from being an outlier to not"
    Do we need to invalidate?


Probably need to pick the brain of @erikjohnston here a bit to see what he thinks.

synapse/storage/databases/main/events_worker.py
```diff
@@ -492,8 +494,46 @@ async def _get_events_from_cache_or_db(self, event_ids, allow_rejected=False):
...
+        if self._external_cache.is_enabled():
+            # XXX: Is there danger in doing this?
+            # We could hold a set of recently evicted keys in memory if
+            # we need this to be synchronous?
+            run_as_background_process(
+                "getEvent_external_cache_delete",
+                self._external_cache.delete,
+                GET_EVENT_CACHE_NAME,
+                event_id,
+            )
```
Half-Shot, line 506: Danger danger

synapse/storage/databases/main/events_worker.py
```diff
@@ -62,7 +62,8 @@
 EVENT_QUEUE_THREADS = 3  # Max number of threads that will fetch events
```
Half-Shot, line 62: Seeing a lot of tests fail due to timeouts over replication. I am not having much luck understanding why, and I'm not able to reproduce the problem in a live environment.

synapse/storage/databases/main/events_worker.py
```diff
@@ -62,7 +62,8 @@
...
 EVENT_QUEUE_ITERATIONS = 3  # No. times we block waiting for requests for events
 EVENT_QUEUE_TIMEOUT_S = 0.1  # Timeout when waiting for requests for events
-
+EVENT_EXPIRY_TIME = 24 * 60 * 60 * 1000
```
Half-Shot, line 65: Not sure if we really want to have a timeout at all. 24h seemed reasonable at the time, but typically this means the cache gets thrashed every 24 hours. Maybe we should have no limit and expect admins to regulate redis memory usage through configuration?

synapse/storage/databases/main/events_worker.py
richvdh: You probably know this already, but this is absolutely not safe: if we modify the cached event (for example due to redaction), and then try to fetch the event again, there's a good chance of getting the unmodified version back from the cache.

Half-Shot: Yup, this is horrible. Even if we maintained a separate redaction cache, the set operation is async. I'm trying to think of ways to make modifying the cache here synchronous.

synapse/storage/databases/main/events_worker.py
Half-Shot: I removed this, there was a significant performance penalty when the events eventually expired out of the cache. Better to let redis handle cache eviction.

erikjohnston: An alternative would be to update the expiry on each access

synapse/storage/databases/main/events_worker.py
richvdh: I don't really follow what you mean by "the set operation". You mean, storing the redaction in the cache? I don't think that matters - persisting the redaction to the database is also async.

erikjohnston: To invalidate the Redis cache I think we need track "in flight" invalidations, so that any query for an entry in the process of being invalidated just returns `None`. We'll also need to do this for any `set` operation (for mutable caches) to ensure that we don't get stale results.

When combining with database caches extra care needs to be taken to ensure that streaming the cache invalidation happens *after* we've deleted stuff from Redis. This is probably going to be a bit fiddly to get right, especially for event caches that get invalidated due to the events stream. I *think* it can be done by doing the `await cache.delete` in the context manager that assigns new stream IDs for events (as we won't stream invalidation until after).

We also then need to ensure that we handle the case where a worker re-inserts stale data between the Redis cache being cleared and the DB invalidation being streamed to the workers. Re-invalidating the Redis cache on the workers when it receives the invalidation doesn't solve this, as it still allows workers to (briefly) clobber the Redis cache with stale data, which will be visible to workers that already invalidated the cache (and may then get put back into the in-memory cache?). l don't know quite how to fix that, other than introducing sequence numbers so that "newer" versions of events don't get clobbered by older versions.

TBH I think that we should try hard to avoid this and just have immutable caches (or mutable caches only in Redis). That probably means just storing the unredacted event dict (and possibly internal metadata?), and have everything else be queried from the DB and cached solely in memory. Hopefully the mutable bits are small and so we can cache a lot more in memory, which should hopefully still give us some benefit.

erikjohnston:
Looks like this PR has a bunch of spurious commits ? Looks like a rebase that included a merge commit.

synapse/storage/databases/main/events_worker.py
richvdh: I still agree with the general sentiment of "let's do everything we can to avoid cache invalidation in Redis", but something I don't quite follow here:

> To invalidate the Redis cache I think we need track "in flight" invalidations

In principle, why can't we just treat the Redis cache invalidation as part of the async "persist event" operation, provided we do the invalidation of the in-memory-cache (and streaming that invalidation to workers) after the redis invalidation completes?   Any operation which tries to read the event between the start of the "persist event" and the in-memory invalidation might get an old copy of the event, but isn't that true anyway?

Also:

> (or mutable caches only in Redis). 

is that really what you meant to type?

erikjohnston: > > To invalidate the Redis cache I think we need track "in flight" invalidations
> 
> In principle, why can't we just treat the Redis cache invalidation as part of the async "persist event" operation, provided we do the invalidation of the in-memory-cache (and streaming that invalidation to workers) after the redis invalidation completes? Any operation which tries to read the event between the start of the "persist event" and the in-memory invalidation might get an old copy of the event, but isn't that true anyway?

I think that's fine, so long as we do wait for the Redis invalidation (rather than doing that as a background task). For the current database caches we try really hard to immediately invalidate caches after we commit to the DB, to avoid having inconsistencies due to having a mix of live and stale data. That may be less of a concern with things we persist in Redis, and for the event cache its probably not a problem. I'm not sure I'm comfortable with saying its true generally though.



> > (or mutable caches only in Redis).
> 
> is that really what you meant to type?

I mean I think its fine to cache things in Redis that are mutable if we don't also cache things in memory (and so avoid the need to coordinate the invalidation between the Redis and in-memory caches)

richvdh: > For the current database caches we try really hard to immediately invalidate caches after we commit to the DB

I don't really follow this. There's inherently a window between the commit and the cache invalidation: there is no "immediately". You might say we'll be widening that window, which is true, but if we have code that depends on that race not happening it might be best to find it than wait for the 1-in-a-million time where it bites...

None of this changes the fact that I think an architecture that avoids Redis cache invalidation will be much easier to reason about (and probably more performant)!

synapse/storage/databases/main/events_worker.py
```diff
@@ -62,7 +62,8 @@
 EVENT_QUEUE_THREADS = 3  # Max number of threads that will fetch events
 EVENT_QUEUE_ITERATIONS = 3  # No. times we block waiting for requests for events
 EVENT_QUEUE_TIMEOUT_S = 0.1  # Timeout when waiting for requests for events
-
+GET_EVENT_CACHE_NAME = "getEvent"
+EXTERNAL_CACHE_EXPIRY_SECONDS = 6 * 60 * 60 # 6 hours
```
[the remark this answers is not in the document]
Half-Shot: Sounds good on both counts.

synapse/replication/tcp/external_cache.py
```diff
@@ -81,14 +101,16 @@ async def set(self, cache_name: str, key: str, value: Any, expiry_ms: int) -> No
             )
         )
 
-    async def get(self, cache_name: str, key: str) -> Optional[Any]:
+    async def get(self, cache_name: str, key: str, expiry_s: Optional[int] = None) -> Optional[Any]:
```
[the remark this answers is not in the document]
erikjohnston: We can change this method signature to `expiry_ms` and then convert it to seconds before calling txredisapi?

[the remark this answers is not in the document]
Half-Shot: Oh I see, yeah that works.

Half-Shot:
>Looks like this PR has a bunch of spurious commits ? Looks like a rebase that included a merge commit.

Yeah, I hadn't noticed that until it was too late. I've corrected the branch now.

synapse/storage/databases/main/events_worker.py
Half-Shot: Done!

erikjohnston:
@Half-Shot what's the state of this ooi?

richvdh:
this seems to be bitrotting and unloved, so I'm going to close it.

PR #9381, opened 2021-02-11
Update black, and run auto formatting over the codebase

MadLittleMods:
This PR accomplishes:

 - Update black version to the latest
 - Run black auto formatting over the codebase
    - Run autoformatting according to [`docs/code_style.md
`](https://github.com/matrix-org/synapse/blob/80d6dc9783aa80886a133756028984dbf8920168/docs/code_style.md)
 - Update `code_style.md` docs around installing black to use the correct version

### Pull Request Checklist


MadLittleMods:
The [`./scripts-dev/lint.sh`](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md#code-style) and [`black` formatter](https://github.com/matrix-org/synapse/blob/80d6dc9783aa80886a133756028984dbf8920168/docs/code_style.md#formatting-tools) seem to be at odds. How do I make them agree so my editor can auto-format correctly?

erikjohnston:
Does `lint.sh` do something different than `black` locally for you? It should just be calling `black`.

The usual reason for descrepancies is because we pin our version of black so that CI doesn't break whenever they release a new version:

https://github.com/matrix-org/synapse/blob/3f58fc848d0002de4605bed91603a1f9f245d128/setup.py#L98-L103

Ideally we'd update our pinned version fairly quickly, but we usually forget. Two options: 1) you can just install the older version of black via `pip install black==19.10b0`, or b) update the version in `setup.py`

MadLittleMods:
> b) update the version in setup.py

Perfect! Thanks for pointing out the version @erikjohnston! Everything seems to be happy now 😃 

Perhaps the [docs for installing black](https://github.com/matrix-org/synapse/blob/80d6dc9783aa80886a133756028984dbf8920168/docs/code_style.md) could be updated as well to make sure we match the version when installing that way. Can we define it in [`python_dependencies.py`](https://github.com/matrix-org/synapse/blob/2c9b4a5f16b7950e7ebb2e5f445e7ae925e61113/synapse/python_dependencies.py#L49) and reference from there?

ShadowJonathan:
Maybe the title of this PR should be changed to reflect what is going on? e.g. ""Update black, and run auto formatting over the codebase"?

clokep:
> Perhaps the [docs for installing black](https://github.com/matrix-org/synapse/blob/80d6dc9783aa80886a133756028984dbf8920168/docs/code_style.md) could be updated as well to make sure we match the version when installing that way.

We should match the instructions in contributing to use `pip install -e ".[lint,mypy]"`

MadLittleMods:
Thanks @clokep 🐊, update the docs!

contrib/experiments/cursesio.py
```diff
@@ -63,8 +63,8 @@ def print_log(self, text):
         self.redraw()
 
     def redraw(self):
-        """ method for redisplaying lines
-            based on internal list of lines """
+        """method for redisplaying lines
+        based on internal list of lines"""
```
clokep, line 67:
```suggestion
        """method for redisplaying lines based on internal list of lines"""
```

synapse/appservice/api.py
```diff
@@ -227,7 +227,9 @@ async def push_bulk(
 
         try:
             await self.put_json(
-                uri=uri, json_body=body, args={"access_token": service.hs_token},
+                uri=uri,
+                json_body=body,
+                args={"access_token": service.hs_token},
```
clokep, line 232: I was surprised by this change, but I think it is due to:

> added support for pre-splitting collections by putting an explicit trailing comma inside

From the [black changelog](https://github.com/psf/black/blob/master/CHANGES.md).

If you remove the trailing comma than black won't reformat this since it is on a short enough line. I don't have a strong opinion about whether we would want to remove the commas instead of reformatting. 🤷 

This seems to make up most of the diff fwiw.

MadLittleMods: We're making a change either way. I'd prefer not to fix them all 😅

Is there a way to retry the build? It looks like it failed on one of the runs, https://buildkite.com/matrix-dot-org/synapse/builds/15308

anoadragon453: I've retried it for you. You should see a Retry button if you're signed in to Buildkite and have the appropriate permissions. If you don't have the latter, ask in the CI Internal room.

MadLittleMods: Thanks @anoadragon453! Looks all ✅ now. I've asked in https://matrix.to/#/!QaKLeuyXEEnDvFiPex:matrix.org/$1613508293261011LJpKF:matrix.org?via=matrix.org&via=sw1v.org&via=amorgan.xyz

docs/code_style.md
```diff
@@ -8,16 +8,16 @@ errors in code.
 
 The necessary tools are detailed below.
 
+First install them with:
+
+    pip install -e ".[lint,mypy]"
```
MadLittleMods, line 13: Updated install instructions to use the version already specified instead of the latest.

Installing the `latest` version leads to mismatches between editor auto-formating and `lint.sh`

richvdh (approved):
:ship:

MadLittleMods:
Thanks @erikjohnston @ShadowJonathan @clokep @anoadragon453 @richvdh @callahad for the help getting this through 🤘🦏

PR #9383, opened 2021-02-11
Add a config option to prioritise local users in user directory search results

anoadragon453:
This PR adds a homeserver config option, `user_directory.prefer_local_users`, that when enabled will show local users higher in user directory search results than remote users. This option is off by default.

Note that turning this on doesn't necessarily mean that remote users will always be put below local users, but they should be assuming all other ranking factors (search query match, profile information present etc) are identical.

This is useful for, say, University networks that are openly federating, but want to prioritise local students and staff in the user directory over other random users.

Reviewable commit-by-commit.

### DIFF
--- a/synapse/storage/databases/main/user_directory.py
+++ b/synapse/storage/databases/main/user_directory.py
@@ -558,6 +558,11 @@ class UserDirectoryStore(UserDirectoryBackgroundUpdateStore):
     def __init__(self, database: DatabasePool, db_conn, hs):
         super().__init__(database, db_conn, hs)
 
+        self._prioritise_local_users_in_search = (
+            hs.config.user_directory_search_prioritise_local_users
+        )
+        self._server_name = hs.config.server_name
+
     async def remove_from_user_dir(self, user_id: str) -> None:
         def _remove_from_user_dir_txn(txn):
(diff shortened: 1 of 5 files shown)


synapse/storage/databases/main/user_directory.py
```diff
@@ -750,6 +755,29 @@ async def search_user_dir(self, user_id, search_term, limit):
...
+            elif isinstance(self.database_engine, Sqlite3Engine):
+                # Note that we need to include a comma at the end for valid SQL
+                statement = "user_id LIKE ? DESC,"
+            else:
+                # This should be unreachable.
+                raise Exception("Unrecognized database engine")
+            additional_ordering_statements.append(statement)
+
+            # Append the local server name as an argument to the final query
+            ordering_arguments += ("%:" + self._server_name,)
```
erikjohnston, line 779: I think it would probably make more sense to split this in two and put into the db specific branches below, otherwise matching up the query structures is quite confusing

anoadragon453: Done in e87885d81. I agree that it does look a bit cleaner, and will be nicer if we ever (heaven forbid) add another DB engine implementation.

erikjohnston (approved):

PR #9384, opened 2021-02-11
Allow OIDC config to override discovered values

### LINKED ISSUE #9347: OIDC: allow individual endpoints to be overridden without disabling discovery
It's sometimes useful to be able to tweak individual endpoints for the OpenID Connect auth provider, for example to add additional parameters to the "authentication" API. One example is for Google, which [specifies](https://developers.google.com/identity/protocols/oauth2/openid-connect#authenticationuriparameters) a `prompt` parameter telling Google to let the user pick between one of their accounts.

Currently the only way for an admin to add these parameters is by disabling `discovery` and manually configuring *all* of the OAuth2 endpoints. I think we should reverse this and make configured endpoints override any discovered ones.

richvdh:
Fixes #9347

### DIFF
--- a/changelog.d/9384.misc
+++ b/changelog.d/9384.misc
@@ -1,0 +1,1 @@
+Allow OIDC config to override discovered values.
--- a/synapse/handlers/oidc_handler.py
+++ b/synapse/handlers/oidc_handler.py
@@ -383,21 +383,30 @@ class OidcProvider:
         return await self._provider_metadata.get()
 
     async def _load_metadata(self) -> OpenIDProviderMetadata:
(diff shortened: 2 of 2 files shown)


clokep (approved):
This seems like it has some chance of breaking working configurations (if someone configured `discover` to `True`, but filled in broken other info). I don't see a better way to do this besides "your config is broken" though. 🤷

richvdh:
> This seems like it has some chance of breaking working configurations (if someone configured discover to True, but filled in broken other info).

yeah, indeed. That would be silly though.

PR #9390, opened 2021-02-11
Add http2 to the nginx example config

auscompgeek:
### Pull Request Checklist

### DIFF
--- a/changelog.d/9390.doc
+++ b/changelog.d/9390.doc
@@ -1,0 +1,1 @@
+Add HTTP/2 support to the nginx example configuration. Contributed by David Vo.
--- a/docs/reverse_proxy.md
+++ b/docs/reverse_proxy.md
@@ -40,12 +40,12 @@ the reverse proxy and the homeserver.
 
 ```
 server {
-    listen 443 ssl;
-    listen [::]:443 ssl;
+    listen 443 ssl http2;
+    listen [::]:443 ssl http2;
 
     # For the federation port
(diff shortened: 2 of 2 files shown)


clokep:
A couple of questions:

* This causes nginx to terminate HTTP/2, but it still talks HTTP/1.1 to the backend?
* Is this available in older nginx versions? (Is someone blindly copying and pasting this likely going to break something?)

auscompgeek:
> This causes nginx to terminate HTTP/2, but it still talks HTTP/1.1 to the backend?

Yes.

> Is this available in older nginx versions?

`http2` was added in nginx 1.9.5, released in 2015: https://www.nginx.com/blog/nginx-1-9-5/

clokep:
Hmmm, their docs do say that HTTP/2 support isn't built by default though: https://nginx.org/en/docs/http/ngx_http_v2_module.html

auscompgeek:
I can't imagine that'll be a problem in practice unless people are building their own nginx. [Debian's nginx-full](https://packages.debian.org/buster/nginx-full) (the default provider of the `nginx` package) ships with HTTP/2 support, and from what I can tell [Alpine also ship nginx with HTTP/2](https://git.alpinelinux.org/aports/tree/main/nginx/APKBUILD?h=3.9-stable#n188).

clokep (approved):

clokep:
Fair enough. I hope it would give a reasonable error message in that case anyway.

Issue #9351, opened 2021-02-08
Redis connections stop reconnecting if they time out while authenticating


erikjohnston:
If txredisapi gets an exception while waiting for the reply to `AUTH`, it will close the connection and disable reconnections (likely the idea is that if you get a wrong password it shouldn't keep trying). However, since any exception triggers this it can cause Synapse to permanently stop talking to redis even if the failure is transient (e.g. if the synapse process wedges at 100% CPU the `AUTH` request can time out).

This leads to failure modes where workers can silently stop receiving updates, or message sending stops working, etc. The only way to fix it is to manually restart the process. Not all cases will trigger alerts, so this is a fairly terrible thing to occur.

Probably the easiest way, but hacky, way of fixing this is making the `continueTrying` attribute on our factories ignore attempts to set it to `False`.

c.f. https://sentry.matrix.org/sentry/synapse-matrixorg/issues/198649/

callahad:
We believe this has been resolved.

PR #9391, opened 2021-02-11
Ensure that we never stop reconnecting to redis

erikjohnston:
Hopefully a fix for https://github.com/matrix-org/synapse/issues/9351, albeit in a truly awful way.

### DIFF
--- a/synapse/replication/tcp/redis.py
+++ b/synapse/replication/tcp/redis.py
@@ -47,7 +47,7 @@ T = TypeVar("T")
 V = TypeVar("V")
 
 
-@attr.s(slots=True, frozen=True)
+@attr.s
 class ConstantProperty(Generic[T, V]):
     """A descriptor that returns the given constant, ignoring attempts to set
     it.
--- a/synapse/replication/tcp/redis.py
+++ b/synapse/replication/tcp/redis.py
@@ -15,8 +15,9 @@
 
 import logging
 from inspect import isawaitable
-from typing import TYPE_CHECKING, Optional, Type, cast
+from typing import TYPE_CHECKING, Generic, Optional, Type, TypeVar, cast
 
+import attr
 import txredisapi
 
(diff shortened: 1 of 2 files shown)


synapse/replication/tcp/redis.py
```diff
@@ -42,6 +43,24 @@
 
 logger = logging.getLogger(__name__)
 
+T = TypeVar("T")
+V = TypeVar("V")
+
+
+@attr.s(slots=True, frozen=True)
+class ConstantProperty(Generic[T, V]):
```
clokep, line 51: Can you just use the `@property` decorator (or does that raise when attempting to set it?)

synapse/replication/tcp/redis.py
```diff
@@ -42,6 +43,24 @@
...
+    """A descriptor that returns the given constant, ignoring attempts to set
+    it.
+    """
+
+    constant = attr.ib()  # type: V
```
clokep, line 56: FWIW we can now do the following I think?

```suggestion
    constant: V = attr.ib()
```

erikjohnston: I thought that was py3.6?

clokep: Oh, maybe. I'm guessing this is why you need the `cast()` call?

synapse/replication/tcp/redis.py
erikjohnston: Yeah, that raises. We could do `continueTrying = property(lambda _: True, lambda _a, _b: True)` but eww

synapse/replication/tcp/redis.py
erikjohnston: I believe it *should* understand descriptors enough to understand, but empirically it didn't. Possibly because of the `#type: ignore` (as mypy doesn't like overriding a bool attribute with a descriptor that returns a bool...)

clokep (approved):
Seems ugly, but should work.

Issue #8956, opened 2020-12-16
Presence logic for handling newly joined rooms is very inefficient and does duplicate work


erikjohnston:
When a user joins a room Synapse needs to decide what presence updates to send out. For local users joining a room we need to send a presence update to all users in the room, for remote users joining we need to send out all local users presence to the remote user.

This is implemented by streaming the current state delta stream, however when the server joins a federated room for the first time the state delta will include join events for *all* members. This leads to a situation where we correctly process the fact that a local user has joined a room and send out presence updates, and *then* incorrectly process each remote join event as if they're a new user join, leading to a lot of unnecessary work and superfluous presence events being generated and sent over federation. 

https://github.com/matrix-org/synapse/blob/6d02eb22dfde9551c515acaf73503e2500e00eaf/synapse/handlers/presence.py#L860-L953

anoadragon453:
> This leads to a situation where we correctly process the fact that a local user has joined a room and send out presence updates, and then incorrectly process each remote join event as if they're a new user join, leading to a lot of unnecessary work and superfluous presence events being generated and sent over federation.

For the latter, what happens is that we get a large bundle of state deltas to process when we join a room. However we indeed end up processing all the existing remote joins as new user joins.

Save checking whether the joins are older than our local user join (which there may not be a reliable way to do so? `stream_ordering`, `depth`?) we can just drop the remote joins in this instance, as a local user join means we'll be sending presence to all current hosts in the room regardless.

anoadragon453:
The first half of this (not processing each remote join as a new server joining the room) was solved in https://github.com/matrix-org/synapse/pull/9402.

The second half (only sending presence updates when this is the first time a remote server shares a room with the local user), will be tackled in a separate PR.

PR #9394, opened 2021-02-11
Remove dead handled_events set in invite_join

anoadragon453:
This PR removes a set that was created and [initially used](https://github.com/matrix-org/synapse/commit/1d2a0040cff8d04cdc7d7d09d8f04a5d628fa9dd#diff-0bc92da3d703202f5b9be2d3f845e375f5b1a6bc6ba61705a8af9be1121f5e42R435-R436), but is no longer today.

May help cut down a bit on the time it takes to accept invites.

### DIFF
--- a/changelog.d/9394.misc
+++ b/changelog.d/9394.misc
@@ -1,0 +1,1 @@
+Remove some dead code from the acceptance of room invites path.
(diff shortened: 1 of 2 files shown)


clokep (approved):
LGTM! Did you just find this by inspection?

anoadragon453:
> LGTM! Did you just find this by inspection?

I just came across it while tracing where some methods were used while working on #8956. Just a coincidence :)

Issue #7988, opened 2020-07-30
Convert inlineCallbacks to async/await


clokep:
There's a desire to convert code that uses `inlineCallbacks` to `async` / `await` since it makes stack traces nicer and helps with profiling (e.g. #7670).

Some progress on this can be tracked at https://patrick.cloke.us/areweasyncyet/

I've broken down the Synapse code into modules for tracking of this work:

* [ ] `contrib.cmdclient`
* [ ] `contrib.experiments`
* [x] `docs.log_contexts` #10353
* [ ] `scripts.synapse_port_db` #6718
* [x] `synapse.api` #7363 #8031
* [ ] `synapse.app` #7620 #7868 #8213
* [x] `synapse.appservice` #7973 #8207
* [x] `synapse.crypto` #8003
* [x] `synapse.events` #7949
* [ ] `synapse.federation` #5689 #6279 #6280 #6840 #6996 #7363 #7868 #7975 #8072
* [x] `synapse.groups` #7363 #7600 #7951
* [ ] `synapse.handlers` #6275 #6280 #6484 #6496 #6504 #6517 #6559 #6837 #6991 #7261 #7338 #7363 #7396 #7459 #7460 #7507 #7557 #7561 #7640 #7649 #7678 #7679 #7727 #7775 #7851 #7860 #7871 #7884 #7912 #7935 #7939 #7945 #7950
* [ ] `synapse.http` #5475 #5585 #7874 #7975 #8016 #8214
* [ ] `synapse.logging` #7872
* [ ] `synapse.metrics` #8032
* [x] `synapse.module_api`
* [ ] `synapse.notifier` #7395
* [x] `synapse.push` #7557 #7948 #7981
* [x] `synapse.replication` #6274 #6666 #7987
* [x] `synapse.rest` #6196 #6482 #6483 #7020 #7110 #7873 #7947 #8027
* [x] `synapse.server`
* [x] `synapse.server_notices` #7363 #7394 #7996
* [x] `synapse.spam_checker_api`
* [x] `synapse.state` #7942 #8140
* [x] `synapse.storage` #6647 #7241 #7363 #7620 #7963 #7976 #8042 #8045 #8061 #8062 #8063 #8066 #8069 #8070 #8071 #8074 #8075 #8076 #8087 #8100 #8121 #8133 #8156 #8162 #8166 #8168 #8173 #8191 #8192 #8194 #8195 #8197 #8199 #8200 #8201
* [x] `synapse.streams` #8014
[...]

PR #9396, opened 2021-02-11
Convert additional test-cases to homeserver test case

clokep:
Building on #9377.

The first commit is a bit different as it actually combines two classes to use a less-mocked-out version. Should be reviewable commit by commit.

Part of #7988

### DIFF
--- a/tests/handlers/test_appservice.py
+++ b/tests/handlers/test_appservice.py
@@ -15,8 +15,6 @@
 
 from mock import Mock
 
-from twisted.internet import defer
-
 from synapse.handlers.appservice import ApplicationServicesHandler
 from synapse.types import RoomStreamToken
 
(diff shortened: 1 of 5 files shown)


tests/handlers/test_appservice.py
```diff
@@ -124,7 +126,9 @@ def test_query_room_alias_exists(self):
             Mock(room_id=room_id, servers=servers)
         )
 
-        result = self.successResultOf(self.handler.query_room_alias_exists(room_alias))
+        result = self.successResultOf(
+            defer.ensureDeferred(self.handler.query_room_alias_exists(room_alias))
+        )
```
anoadragon453: Is there a difference between `self.successResultOf` and `self.get_success`?

anoadragon453 (approved):
Looks good, thanks for doing this!

tests/handlers/test_appservice.py
clokep: `successResultOf` doesn't handle async functions on old Twisted. Note that `get_success` is only available on `HomeserverTestCase`:

https://github.com/matrix-org/synapse/blob/4b73488e811714089ba447884dccb9b6ae3ac16c/tests/unittest.py#L480-L486

PR #9399, opened 2021-02-12
Some speedups for presence

anoadragon453:
This PR speeds up presence a bit in two ways:

* First commit switches out some code run during every presence update that did the following in Python:
 
   - For every local user on the server
      - For every room they're in
        - For every user in those rooms
            - deduplicate and return that user
  
  for SQL that gets the union of 1. all distinct users who share a private room 2. all distinct users that share a public room. Which should hopefully be a bit less insane.

* Second commit is a small one that just prevents us from iterating over the list of hosts in a room twice sequentially when joining that room.

### DIFF
--- a/changelog.d/9399.misc
+++ b/changelog.d/9399.misc
@@ -1,0 +1,1 @@
+Optimise some code behind the user presence feature.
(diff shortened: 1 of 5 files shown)


anoadragon453:
So while this PR added a much more efficient query for checking what users a user shares rooms with, that query couldn't be cached, whereas the old one could. Performance monitoring with a warm db is necessary to see whether this improvement will actually make a difference.

I don't really have the capability to do that at the moment, so I'm closing this for now.

PR #9402, opened 2021-02-13
Be smarter about which hosts to send presence to when processing room joins

### LINKED ISSUE #8956: Presence logic for handling newly joined rooms is very inefficient and does duplicate work

anoadragon453:
Addresses #8956. Does not address "we should also only send presence updates when this is the first time a remote server shares a room with the local user". I'd like to tackle that in a separate PR as it will likely involve tracking which presence updates we've sent out in some form.

This PR attempts to eliminate unnecessary presence sending work when your local server joins a room, or when a remote server joins a room your server is participating in by processing state deltas in chunks rather than individually.

---

When your server joins a room for the first time, it requests the historical state as well. This chunk of new state is passed to the presence handler which, after filtering that state down to only membership joins, will send presence updates to homeservers for each join processed.

It turns out that we were being a bit naive and processing each event individually, and sending out presence updates for every one of those joins. Even if many different joins were users on the same server (hello IRC bridges), we'd send presence to that same homeserver for every remote user join we saw.

This PR attempts to deduplicate all of that by processing the entire batch of state deltas at once, instead of only doing each join individually. We process the joins and note down which servers need which presence:

* If it was a local user join, send that user's latest presence to all servers in the room
* If it was a remote user join, send the presence for all local users in the room to that homeserver

We deduplicate by inserting all of those pending updates into a dictionary of the form:

```
{
  server_name1: {presence_update1, ...},
  server_name2: {presence_update1, presence_update2, ...}
}
```

Only after building this dict do we then start sending out presence updates.

### DIFF
--- a/changelog.d/9402.bugfix
+++ b/changelog.d/9402.bugfix
@@ -1,1 +1,1 @@
-Avoid sending unnecessary presence updates when joining a room.
+Fix a bug where a lot of unnecessary presence updates were sent when joining a room.
--- a/changelog.d/9402.bugfix
+++ b/changelog.d/9402.bugfix
@@ -1,0 +1,1 @@
+Avoid sending unnecessary presence updates when joining a room.
--- a/synapse/handlers/presence.py
+++ b/synapse/handlers/presence.py
@@ -898,7 +898,7 @@ class PresenceHandler(BasePresenceHandler):
         # Send out user presence updates for each destination
         for destination, user_state_set in presence_destinations.items():
             self.federation.send_presence_to_destinations(
-                states=user_state_set, destinations=[destination]
+                destinations=[destination], states=user_state_set
             )
 
     async def _on_user_joined_room(
--- a/synapse/handlers/presence.py
+++ b/synapse/handlers/presence.py
@@ -914,7 +914,7 @@ class PresenceHandler(BasePresenceHandler):
             user_id: The ID of the user that has joined the room.
 
         Returns:
-            A tuple of (list of destinations, list of presence updates).
+            A tuple of destinations and presence updates to send to them.
         """
         if self.is_mine_id(user_id):
             # If this is a local user then we need to send their presence
--- a/synapse/handlers/presence.py
+++ b/synapse/handlers/presence.py
@@ -892,8 +892,9 @@ class PresenceHandler(BasePresenceHandler):
 
             # Insert the destinations and respective updates into our destinations dict
             for destination in destinations:
-                presence_destinations.setdefault(destination, set())
-                presence_destinations[destination].update(user_presence_states)
+                presence_destinations.setdefault(destination, set()).update(
+                    user_presence_states
+                )
 
         # Send out user presence updates for each destination
         for destination, user_state_set in presence_destinations.items():
@@ -905,7 +906,15 @@ class PresenceHandler(BasePresenceHandler):
         self, room_id: str, user_id: str
     ) -> Tuple[List[str], List[UserPresenceState]]:
         """Called when we detect a user joining the room via the current state
-        delta stream.
+        delta stream. Returns the destinations that need to be updated and the
+        presence updates to send to them.
+
+        Args:
+            room_id: The ID of the room that the user has joined.
+            user_id: The ID of the user that has joined the room.
+
+        Returns:
+            A tuple of (list of destinations, list of presence updates).
         """
         if self.is_mine_id(user_id):
             # If this is a local user then we need to send their presence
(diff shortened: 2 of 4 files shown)


synapse/handlers/presence.py
```diff
@@ -906,9 +929,45 @@ async def _on_user_joined_room(self, room_id: str, user_id: str) -> None:
...
+
+                # Check what other rooms this user shares with other users on a homeserver
+                room_ids = await self.store.get_rooms_for_destination(
+                    destination=host, max_stream_ordering=stream_id,
+                )
+
+                if len(room_ids) > 1:
+                    # This homeserver has a user in another room that we're aware of
+                    # It's not the first time we've seen this homeserver
+                    return
```
ShadowJonathan, line 966: Why a return here?

synapse/handlers/presence.py
```diff
@@ -906,9 +929,45 @@ async def _on_user_joined_room(self, room_id: str, user_id: str) -> None:
             # TODO: We should be able to filter the hosts down to those that
             # haven't previously seen the user
```
ShadowJonathan, line 930: Maybe remove this, as this PR is fixing it :P

anoadragon453:
@ShadowJonathan This PR is still a draft :P

synapse/handlers/presence.py
```diff
@@ -880,29 +884,45 @@ async def _handle_state_delta(self, deltas):
...
+        for destination, user_state_set in presence_destinations.items():
+            self.federation.send_presence_to_destinations(
+                destinations=[destination], states=user_state_set
+            )
+
+    async def _on_user_joined_room(
+        self, room_id: str, user_id: str
+    ) -> Tuple[List[str], List[UserPresenceState]]:
         """Called when we detect a user joining the room via the current state
         delta stream.
```
erikjohnston, line 908: Can you update the docstring with the return values please

synapse/handlers/presence.py
```diff
@@ -880,29 +884,45 @@ async def _handle_state_delta(self, deltas):
...
+            # Retrieve any user presence state updates that need to be sent as a result,
+            # and the destinations that need to receive it
+            destinations, user_presence_states = await self._on_user_joined_room(
+                room_id, state_key
+            )
+
+            # Insert the destinations and respective updates into our destinations dict
+            for destination in destinations:
+                presence_destinations.setdefault(destination, set())
+                presence_destinations[destination].update(user_presence_states)
```
erikjohnston, line 896: Why is this on two lines, rather than `.setdefault(..).update(..)`?

anoadragon453: Awkwardly it doesn't fit within 95 characters on one line, but I can still combine it into one contiguous statement.

synapse/handlers/presence.py
```diff
@@ -880,29 +884,54 @@ async def _handle_state_delta(self, deltas):
...
-        """
+        delta stream. Returns the destinations that need to be updated and the
+        presence updates to send to them.
+
+        Args:
+            room_id: The ID of the room that the user has joined.
+            user_id: The ID of the user that has joined the room.
 
+        Returns:
+            A tuple of (list of destinations, list of presence updates).
```
erikjohnston, line 917:
```suggestion
            A tuple of destinations and presence updates to send to them.
```

maybe?

synapse/handlers/presence.py
erikjohnston: The second lookup in the dict, `presence_destinations[destination]`, was making me sad :)

synapse/handlers/presence.py
anoadragon453: Yeah, that reads better :slightly_smiling_face:

changelog.d/9402.bugfix
```diff
@@ -0,0 +1 @@
+Avoid sending unnecessary presence updates when joining a room.
```
erikjohnston, line 1:
```suggestion
Fix a bug where a lot of unnecessary presence updates were sent when joining a room.
```

erikjohnston (approved):

PR #9408, opened 2021-02-15
Remove dead notify_for_states presence method

anoadragon453:
This PR removes a dead method that was found while working on presence performance. It seems like the commit that added this method (https://github.com/matrix-org/synapse/commit/4e1cebd56f9688d49a51929264c095356005f9a3) didn't call it from anywhere at the time either?

### DIFF
--- a/changelog.d/9408.misc
+++ b/changelog.d/9408.misc
@@ -1,0 +1,1 @@
+Clean up an unused method in the presence handler code.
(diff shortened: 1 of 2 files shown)


clokep:
@anoadragon453 Is this ready for review? Seems good to me.

anoadragon453:
Oops, yes. Not sure why it wasn't marked as such.

clokep (approved):
I double checked there were no callers, seems fine to remove! 👍 

`self.notifier` and `get_interested_parties` seem to have other callers so...

anoadragon453:
Yeah, can't ditch those guys unfortunately :yum:. Thanks!

PR #9411, opened 2021-02-15
Prep work for removing `outlier` from `internal_metadata`

richvdh:
I realised that, for every event, we were storing a second copy of the `outlier` flag in `internal_metadata`. That's 15 bytes per event. FIFTEEN!

Well anyway it feels a bit redundant, and I'd quite like to remove it. I think we can do this in two phases:

 * First (this one): stop relying on the data in the `internal_metadata` column. This should be backwards-compatible.
 * Later: stop storing `outlier` in the `internal_metadata` column. That will break rollback to older synapse versions.

### DIFF
--- a/changelog.d/9411.misc
+++ b/changelog.d/9411.misc
@@ -1,0 +1,1 @@
+Preparatory steps for removing redundant `outlier` data from `event_json.internal_metadata` column.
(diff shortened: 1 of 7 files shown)


richvdh:
I think this is ready for review

erikjohnston (approved):

anoadragon453 (approved):

PR #7785, opened 2020-07-05
Add /user/{user_id}/shared_rooms/ api

Half-Shot:
This change adds a yet-to-be-MSCed unstable feature for fetching rooms you have in common with a user. Since synapse already implements this at the storage level, this API just exposes this as an endpoint in order to provide support for https://github.com/vector-im/riot-web/issues/14339.

[MSC 2666](https://github.com/matrix-org/matrix-doc/pull/2666)

#### Leftover things to implement from the MSC

- Reject `user_id` == `other_user_id`
- Replace `rooms` with `joined`.

### Pull Request Checklist

* [x] Pull request is based on the develop branch
* [x] Pull request includes a [changelog file](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md#changelog). The entry should:
  - Be a short description of your change which makes sense to users. "Fixed a bug that prevented receiving messages from other servers." instead of "Moved X method from `EventStore` to `EventWorkerStore`.".
  - Use markdown where necessary, mostly for `code blocks`.
  - End with either a period (.) or an exclamation mark (!).
  - Start with a capital letter.
* [x] Code style is correct (run the [linters](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md#code-style))

PR #9416, opened 2021-02-16
Remove cache for get_shared_rooms_for_users

anoadragon453:
This PR remove the cache for the `get_shared_rooms_for_users` storage method (the db method driving the experimental "what rooms do I share with this user?" feature: [MSC2666](https://github.com/matrix-org/matrix-doc/pull/2666)). Currently subsequent requests to the endpoint will return the same result, even if your shared rooms with that user have changed.

The cache was added in https://github.com/matrix-org/synapse/pull/7785, but we forgot to ensure it was invalidated appropriately.

Upon attempting to invalidate it, I found that the cache had to be entirely invalidated whenever a user (remote or local) joined or left a room. This didn't make for a very useful cache, especially for a function that may or may not be called very often. Thus, I've opted to remove it instead of invalidating it.

If there's a way to partially invalidate this cache when a user leaves/joins a room, then that may be a case for bringing it back. But at the moment I don't think that's possible (especially if it requires a costly DB query just to do so).

### DIFF
--- a/synapse/storage/databases/main/user_directory.py
+++ b/synapse/storage/databases/main/user_directory.py
@@ -670,7 +670,6 @@ class UserDirectoryStore(UserDirectoryBackgroundUpdateStore):
         users.update(rows)
         return list(users)
 
-    @cached()
     async def get_shared_rooms_for_users(
         self, user_id: str, other_user_id: str
     ) -> Set[str]:
--- a/synapse/storage/databases/main/user_directory.py
+++ b/synapse/storage/databases/main/user_directory.py
@@ -497,8 +497,7 @@ class UserDirectoryBackgroundUpdateStore(StateDeltasStore):
     async def add_users_in_public_rooms(
         self, room_id: str, user_ids: Iterable[str]
     ) -> None:
-        """Insert entries into the users_who_share_private_rooms table. The first
-        user should be a local user.
+        """Insert entries into the users_in_public_rooms table.
 
         Args:
             room_id
--- a/changelog.d/9416.bugfix
+++ b/changelog.d/9416.bugfix
@@ -1,0 +1,1 @@
+Fix a bug that caused multiple calls to the experimental `shared_rooms` endpoint to return stale results.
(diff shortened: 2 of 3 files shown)


synapse/storage/databases/main/user_directory.py
```diff
@@ -506,13 +514,21 @@ async def add_users_in_public_rooms(
...
         """
 
-        await self.db_pool.simple_upsert_many(
-            table="users_in_public_rooms",
-            key_names=["user_id", "room_id"],
-            key_values=[(user_id, room_id) for user_id in user_ids],
-            value_names=(),
-            value_values=None,
-            desc="add_users_in_public_rooms",
+        def _add_users_in_public_rooms_txn(txn):
```
clokep, line 517: (Not related to this PR, but can you fix the docstring above this for `add_users_in_public_rooms`? It refers to the wrong table.)

clokep (commented):
This PR looks fine, but it does seem like we're doing a lot of work to invalidate caches for something that may or may not be called frequently.

anoadragon453:
Yeah, after thinking about this a bit I think it's probably just best to lose the cache. We'd indeed spend so much work invalidating it that I don't think it's particularly enough to offset the benefit of keeping it.

That said, ideally we would rate-limit this endpoint, though the spec doesn't suggest to: https://github.com/matrix-org/matrix-doc/pull/2666.

anoadragon453:
I've updated this PR to just remove the cache instead, as it seemed more costly to invalidate it.

synapse/storage/databases/main/user_directory.py
```diff
@@ -497,8 +497,7 @@ async def add_users_who_share_private_room(
     async def add_users_in_public_rooms(
         self, room_id: str, user_ids: Iterable[str]
     ) -> None:
-        """Insert entries into the users_who_share_private_rooms table. The first
-        user should be a local user.
+        """Insert entries into the users_in_public_rooms table.
```
clokep, line 500: Was this comment just stale?

clokep (approved):
Seems reasonable to me! 👍 The test changes just look like some code de-duplication.

anoadragon453:
> The test changes just look like some code de-duplication.

Code de-duping and making sure a second call happens to the `shared_rooms` endpoint after some change occurred.

synapse/storage/databases/main/user_directory.py
anoadragon453: I think it was a copy/paste error from the method above (`add_users_who_share_private_room`).

Issue #9403, opened 2021-02-13
docker builds broken on armv7


richvdh:
Our Docker build is currently broken on armv7, since python-cryptography now requires a rust compiler to build:

From https://app.circleci.com/pipelines/github/matrix-org/synapse/19720/workflows/47040f65-8516-4995-b250-9cae722e4efc/jobs/25662:

```
#25 5590.   Building wheel for cryptography (PEP 517): started
#25 5605.   Building wheel for cryptography (PEP 517): finished with status 'error'
#25 5605.   ERROR: Command errored out with exit status 1:
#25 5605.    command: /usr/local/bin/python /usr/local/lib/python3.8/site-packages/pip/_vendor/pep517/_in_process.py build_wheel /tmp/tmpyl06y7df
#25 5605.        cwd: /tmp/pip-install-mmoy6v11/cryptography_468fe58643ff4d02b4439954d410409c
...
#25 5605.   generating cffi module 'build/temp.linux-armv7l-3.8/_openssl.c'
#25 5605.   running build_rust
#25 5605.   
#25 5605.       =============================DEBUG ASSISTANCE=============================
#25 5605.       If you are seeing a compilation error please try the following steps to
#25 5605.       successfully install cryptography:
#25 5605.       1) Upgrade to the latest pip and try again. This will fix errors for most
#25 5605.          users. See: https://pip.pypa.io/en/stable/installing/#upgrading-pip
#25 5605.       2) Read https://cryptography.io/en/latest/installation.html for specific
#25 5605.          instructions for your platform.
#25 5605.       3) Check our frequently asked questions for more information:
#25 5605.          https://cryptography.io/en/latest/faq.html
#25 5605.       4) Ensure you have a recent Rust toolchain installed:
#25 5605.          https://cryptography.io/en/latest/installation.html#rust
#25 5605.       5) If you are experiencing issues with Rust for *this release only* you may
#25 5605.          set the environment variable `CRYPTOGRAPHY_DONT_BUILD_RUST=1`.
#25 5605.       =============================DEBUG ASSISTANCE=============================
#25 5605.   
#25 5605.   error: Can not find Rust compiler
#25 5605.   ----------------------------------------
#25 5605.   ERROR: Failed building wheel for cryptography
```

amd64 and arm64 are fine, because we use precompiled wheels from PyPI.

(see also https://github.com/pyca/cryptography/issues/5771)

richvdh:
[...]

PR #9423, opened 2021-02-17
Fix arm docker images

erikjohnston:
c.f. https://github.com/matrix-org/synapse/issues/9403

Basically building the `cryptography` package now requires rust to build, which fetches packages from crates.io. Fetching packages breaks when building for 32bit arch on 64bit platforms, so we create a base image that has the packages already downloaded. This allowis us to run rustc/cargo in offline mode, which doesn't break.

We don't want to do this by default however, as cryptography ships wheels for most platforms and so the extra steps are entirely unnecessary.

Does anyone know of a good way of testing this works on circle CI? I have tested locally

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -79,4 +79,6 @@ commands:
         type: string
     steps:
       - run: docker buildx build -f docker/Dockerfile-cargo-cache --push -t 127.0.0.1:5000/cargo_cache --platform << parameters.platforms >> --progress=plain .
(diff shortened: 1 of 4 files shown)


erikjohnston:
Note that as per https://github.com/docker/buildx#--driver-driver multi platform images can't be loaded into the local `docker images`, so we have to push our intermediate cargo cache images to a local repository that we spin up.

erikjohnston:
Also note that the last commit enables building of images for this branch and disables push to docker hub so that we can test it actually works.

erikjohnston:
Ugh, now the docker builds are timing out :/

clokep:
@erikjohnston Are we closing this then?

erikjohnston:
We've decided to drop support for ARMv7, so yeah

PR #9425, opened 2021-02-17
Fix only handling the last presence state for each user

anoadragon453:
This is a small bug that I noticed while working on #8956.

We have a for-loop which attempts to strip all presence changes for each user except for the final one, as we don't really care about older presence:

https://github.com/matrix-org/synapse/blob/9e19c6aab4b5a99039f2ddc7d3120dd3b26c274b/synapse/handlers/presence.py#L368-L371

`new_states_dict` stores this stripped copy of latest presence state for each user, before it is... put into a new variable `new_state`, which is just overridden by the subsequent for loop.

I believe this was instead meant to override `new_states`. Without doing so, it effectively meant:

1. The for loop had no effect.
2. We were still processing old presence state for users.

### DIFF
--- a/synapse/handlers/presence.py
+++ b/synapse/handlers/presence.py
@@ -349,10 +349,13 @@ class PresenceHandler(BasePresenceHandler):
                 [self.user_to_current_state[user_id] for user_id in unpersisted]
             )
 
-    async def _update_states(self, new_states):
+    async def _update_states(self, new_states: Iterable[UserPresenceState]):
         """Updates presence of users. Sets the appropriate timeouts. Pokes
         the notifier and federation if and only if the changed presence state
         should be sent to clients/servers.
+
+        Args:
+            new_states: The new user presence state updates to process.
         """
         now = self.clock.time_msec()
 
--- a/synapse/handlers/presence.py
+++ b/synapse/handlers/presence.py
@@ -349,7 +349,7 @@ class PresenceHandler(BasePresenceHandler):
                 [self.user_to_current_state[user_id] for user_id in unpersisted]
             )
 
-    async def _update_states(self, new_states: Iterable[UserPresenceState]):
+    async def _update_states(self, new_states: Iterable[UserPresenceState]) -> None:
         """Updates presence of users. Sets the appropriate timeouts. Pokes
         the notifier and federation if and only if the changed presence state
         should be sent to clients/servers.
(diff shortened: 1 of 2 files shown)


clokep:
I think add type hints to the method declaration would cause mypy to find this FTR.

anoadragon453:
> I think adding type hints to the method declaration would cause mypy to find this FTR.

@clokep Do you think so? if we added a type hint to `new_states` I'm not sure that would have caught us accidentally not assigning to it.

Personally I'm confused why `flake8` didn't catch the unused `new_state`.

clokep:
> Personally I'm confused why `flake8` didn't catch the unused `new_state`.

Well it is used, it is set into a dicationary. 😄

clokep:
> @clokep Do you think so? if we added a type hint to `new_states` I'm not sure that would have caught us accidentally not assigning to it.

I think so because `new_state` when you iterate the dictionary and when you set it to `new_states_dict.values()` are of different types? Or maybe not...well regardless we should add a type hint. 😛

anoadragon453:
> > Personally I'm confused why `flake8` didn't catch the unused `new_state`.
> 
> Well it is used, it is set into a dicationary. smile

Oh, bah fair. But it's still getting assigned to in L371 and then never getting used. That should count for something :grin:

anoadragon453:
> > @clokep Do you think so? if we added a type hint to `new_states` I'm not sure that would have caught us accidentally not assigning to it.
> 
> I think so because `new_state` when you iterate the dictionary and when you set it to `new_states_dict.values()` are of different types? Or maybe not...well regardless we should add a type hint. stuck_out_tongue

Ohh, I see what you mean. Yes, I'm all for adding new type hints, and this gives me a great excuse to do so in this PR!

anoadragon453:
> synapse/handlers/presence.py:371: error: Incompatible types in assignment (expression has type "ValuesView[UserPresenceState]", variable has type "UserPresenceState")  [assignment]
synapse/handlers/presence.py:401: error: Argument 1 to "len" has incompatible type "Iterable[UserPresenceState]"; expected "Sized"  [arg-type]
Found 2 errors in 1 file (checked 379 source files)

(You were right, by the way!)

synapse/handlers/presence.py
```diff
@@ -349,10 +349,13 @@ async def _persist_unpersisted_changes(self):
                 [self.user_to_current_state[user_id] for user_id in unpersisted]
             )
 
-    async def _update_states(self, new_states):
+    async def _update_states(self, new_states: Iterable[UserPresenceState]):
```
clokep, line 352:
```suggestion
    async def _update_states(self, new_states: Iterable[UserPresenceState]) -> None:
```

clokep (approved):

PR #9426, opened 2021-02-17
Parse ui_auth.session_timeout as a duration (instead of treating it as ms)

arya2331:
### Pull Request Checklist

  - Be a short description of your change that makes sense to users. "Fixed a bug that prevented receiving messages from other servers." instead of "Moved X method from `EventStore` to `EventWorkerStore`.".


arya2331:
Signed-off-by: Rishabh Arya <rishabharya32@gmail.com>

synapse/config/auth.py
```diff
@@ -37,7 +39,9 @@ def read_config(self, config, **kwargs):
 
         # User-interactive authentication
         ui_auth = config.get("ui_auth") or {}
-        self.ui_auth_session_timeout = ui_auth.get("session_timeout", 0)
+        self.ui_auth_session_timeout = con_fig.parse_duration(
```
clokep, line 42: Since `AuthConfig` is a subclass of `Config` this can be called as `self.parse_duration`.

It looks to me like `parse_duration` will do the correct thing if the old version is given (e.g. a raw `6000` should still work out to 6000).

We also need to update the documentation below about this configuration variable.

arya2331: @clokep I have updated PR using `self.parse_duration`.  Yes, `parse_duartion` will do the correct thing if the old version is given. I have also updated the documentation below

changelog.d/9426.misc
```diff
@@ -0,0 +1 @@
+Calling the 'parse_duration' function so that configs can be specified in terms of number of seconds/hours/etc/.
```
clokep, line 1: We want to word this in terms of how a Synapse administrator might care, so maybe:

```suggestion
The `ui_auth.session_timeout` configuration setting be specified in terms of number of seconds/minutes/etc/.
```

You can also add `Contributed by <your name>.` to the end if you'd like!

synapse/config/auth.py
```diff
@@ -106,4 +108,7 @@ def generate_config_section(self, config_dir_path, server_name, **kwargs):
             # seconds.
             #
             #session_timeout: 15000
+            # Now ui_auth also allow standard duration formats such as "15s" to allow
+            # for credential validation to last for 15 seconds.
+            # session_timeout: "15s"
```
clokep, line 113: Instead of adding a new section here let's update the above. I think if we change the first sentence to:

> The amount of time to allow a user-interactive authentication session to be active.

And then we can update the `#session_timeout: 15000` to `#session_timeout: 15s`, like you did (although note that there should be a space there).

clokep (changes requested):
Looks good! I think the documentation just needs a minor update!

clokep (approved):
Looks good if CI passes!

clokep:
Thank you!

PR #9428, opened 2021-02-17
Add back the guard against the user directory stream position being unset

### LINKED ISSUE #9420: Traceback on background process 'user_directory.notify_new_event' threw an exception
### Description

I just upgraded a homeserver from 1.26 to 1.27 and now I'm seeing a lot of these tracebacks in my homeserver.log:

```
2021-02-16 17:36:09,420 - synapse.metrics.background_process_metrics - 211 - ERROR - user_directory.notify_new_event-9 - Background process 'user_directory.notify_new_event' threw an exception
  File "/opt/venvs/matrix-synapse/lib/python3.7/site-packages/synapse/metrics/background_process_metrics.py", line 208, in run
    return await maybe_awaitable(func(*args, **kwargs))
  File "/opt/venvs/matrix-synapse/lib/python3.7/site-packages/synapse/handlers/user_directory.py", line 110, in process
    await self._unsafe_process()
  File "/opt/venvs/matrix-synapse/lib/python3.7/site-packages/synapse/handlers/user_directory.py", line 159, in _unsafe_process
    self.pos, room_max_stream_ordering
  File "/opt/venvs/matrix-synapse/lib/python3.7/site-packages/synapse/storage/databases/main/state_deltas.py", line 51, in get_current_state_deltas
    prev_stream_id = int(prev_stream_id)
TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType
```

### Version information

- **Homeserver**: Private
- **Version**: 1.27
- **Install method**: Debian package.
- **Platform**: Debian

clokep:
This adds back the code removed in https://github.com/matrix-org/synapse/pull/9223/files#r564010797.

That comment is incorrect -- the call to `get_user_directory_stream_pos` will raise an error if there's no row in `get_user_directory_stream_pos`, but row itself can have a `null` value for `stream_id`.

Fixes #9420

### DIFF
--- a/changelog.d/9428.bugfix
+++ b/changelog.d/9428.bugfix
@@ -1,0 +1,1 @@
+Fix a bug introduced in v1.27.0: "TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType." related to the user directory.
--- a/synapse/handlers/user_directory.py
+++ b/synapse/handlers/user_directory.py
@@ -143,6 +143,10 @@ class UserDirectoryHandler(StateDeltasHandler):
         if self.pos is None:
             self.pos = await self.store.get_user_directory_stream_pos()
 
+        # If still None then the initial background update hasn't happened yet.
+        if self.pos is None:
+            return None
+
         # Loop round handling deltas until we're up to date
         while True:
             with Measure(self.clock, "user_dir_delta"):
--- a/synapse/storage/databases/main/user_directory.py
+++ b/synapse/storage/databases/main/user_directory.py
@@ -707,7 +707,13 @@ class UserDirectoryStore(UserDirectoryBackgroundUpdateStore):
 
         return {row["room_id"] for row in rows}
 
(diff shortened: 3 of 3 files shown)


z3ntu:
Tested on Arch Linux, the exception is gone now.

changelog.d/9428.bugfix
```diff
@@ -0,0 +1 @@
+Fix a bug introduced in v1.27.0: "TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType." related to the user directory.
```
erikjohnston, line 1: Does this have a user visible effect?

erikjohnston (approved):

changelog.d/9428.bugfix
clokep: I don't think so since the user directory wouldn't be ready to be updated anyway?

erikjohnston: Fair :)

PR #9433, opened 2021-02-18
Drop ARMv7 from docker

erikjohnston:
It's proving incredibly hard to build in CircleCI infra.

c.f. #9403 (and #9423)


changelog.d/9433.misc
```diff
@@ -0,0 +1 @@
+Drop support for ARMv7 in Docker images.
```
clokep, line 1: Should this be a `.docker` or `.removal`?

erikjohnston: I keep forgetting those things exist

erikjohnston: I'm putting it in docker, fairly arbitrarily, but removal feels like its more for removing synapse features :shrug:

clokep (approved):
I think this is OK. I'm guessing we still want to do the multi-layered build here so that amd64 gets uploaded first? (I don't remember if both arm64 and arm/v7 were super slow.)

erikjohnston:
> I think this is OK. I'm guessing we still want to do the multi-layered build here so that amd64 gets uploaded first? (I don't remember if both arm64 and arm/v7 were super slow.)

I think its probably safest to leave as is for now either way, tbh.

PR #9434, opened 2021-02-18
Remove deprecated SAML2 callback URL since it does not work.

### LINKED ISSUE #9430: SAML SSO broken on Synapse 1.27
The [1.27 upgrade notes](https://github.com/matrix-org/synapse/blob/v1.27.0/UPGRADE.rst#changes-to-callback-uri-for-oauth2--openid-connect) claim:

> Note: a similar change is being made for SAML2; in this case the old URI `[synapse public baseurl]/_matrix/saml2` is being deprecated, but **will continue to work**, so no immediate changes are required for existing installations.

In fact, as part of the SAML2 request, the expected callback URI is sent to the SAML2 IdP, which will check that it matches what is expected and fail if not.

Hence, upgrading to Synapse 1.27 will break login for anyone using SAML :/.

clokep:
Fixes #9430

* Update the upgrade documentation to match OIDC bits.
* Stop mounting the old resource (which won't work).

### DIFF
--- a/UPGRADE.rst
+++ b/UPGRADE.rst
@@ -88,21 +88,21 @@ for example:
 Upgrading to v1.27.0
 ====================
 
-Changes to callback URI for OAuth2 / OpenID Connect
----------------------------------------------------
+Changes to callback URI for OAuth2 / OpenID Connect and SAML2
+-------------------------------------------------------------
 
-This version changes the URI used for callbacks from OAuth2 and SAML2 identity providers.
+This version changes the URI used for callbacks from OAuth2 and SAML2 identity providers:
 
-If your server is configured for single sign-on via an OpenID Connect or OAuth2 identity
-provider, you will need to add ``[synapse public baseurl]/_synapse/client/oidc/callback``
-to the list of permitted "redirect URIs" at the identity provider.
+* If your server is configured for single sign-on via an OpenID Connect or OAuth2 identity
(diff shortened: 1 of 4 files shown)


callahad (approved):
lgtm

callahad:
Newsfragment and we're good to go

UPGRADE.rst
```diff
@@ -88,20 +88,21 @@ for example:
...
 
-(Note: a similar change is being made for SAML2; in this case the old URI
-``[synapse public baseurl]/_matrix/saml2`` is being deprecated, but will continue to
-work, so no immediate changes are required for existing installations.)
+  See `docs/openid.md <docs/openid.md>`_ for more information on setting up OpenID
+  Connect.
+
+* If your server is configured for single sign-on via a SAML2 identity provider, you will
+  need to add ``[synapse public baseurl]/_synapse/client/saml2/authn_response`` as a permitted
+  "ACS location" (also known as "allowed callback URLs") at the identity provider.
```
richvdh, line 105: are they known as "callback URLs" on any real SAML2 provider? I'm not sure that's a SAML2 term.

richvdh: XML: why call it something obvious when you can call it an "Assertion Consumer Service"

clokep: Auth0 calls them that:

image

richvdh: fair enough!

PR #9436, opened 2021-02-18
SSO: redirect to public URL before setting cookies

richvdh:
... otherwise, we don't get the cookie back.

### DIFF
--- a/synapse/http/__init__.py
+++ b/synapse/http/__init__.py
@@ -14,8 +14,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import re
+from typing import Union
 
-from twisted.internet import task
+from twisted.internet import address, task
 from twisted.web.client import FileBodyProducer
 from twisted.web.iweb import IRequest
 
@@ -51,6 +52,40 @@ class QuieterFileBodyProducer(FileBodyProducer):
             FileBodyProducer.stopProducing(self)
         except task.TaskStopped:
             pass
+
+
+def get_request_uri(request: IRequest) -> bytes:
+    """Return the full URI that was requested by the client"""
+    return b"%s://%s%s" % (
+        b"https" if request.isSecure() else b"http",
+        _get_requested_host(request),
+        # despite its name, "request.uri" is only the path and query-string.
+        request.uri,
+    )
+
+
+def _get_requested_host(request: IRequest) -> bytes:
+    hostname = request.getHeader(b"host")
+    if hostname:
+        return hostname
+
+    # no Host header, use the address/port that the request arrived on
+    host = request.getHost()  # type: Union[address.IPv4Address, address.IPv6Address]
+
+    hostname = host.host.encode("ascii")
+
+    if request.isSecure() and host.port == 443:
+        # default port for https
+        return hostname
+
+    if not request.isSecure() and host.port == 80:
+        # default port for http
+        return hostname
+
+    return b"%s:%i" % (
+        hostname,
+        host.port,
+    )
 
 
 def get_request_user_agent(request: IRequest, default: str = "") -> str:
--- a/synapse/rest/client/v1/login.py
+++ b/synapse/rest/client/v1/login.py
@@ -20,6 +20,7 @@ from synapse.api.errors import Codes, LoginError, SynapseError
 from synapse.api.ratelimiting import Ratelimiter
 from synapse.appservice import ApplicationService
 from synapse.handlers.sso import SsoIdentityProvider
+from synapse.http import get_request_uri
 from synapse.http.server import HttpServer, finish_request
 from synapse.http.servlet import (
     RestServlet,
@@ -354,6 +355,7 @@ class SsoRedirectServlet(RestServlet):
             hs.get_oidc_handler()
         self._sso_handler = hs.get_sso_handler()
         self._msc2858_enabled = hs.config.experimental.msc2858_enabled
+        self._public_baseurl = hs.config.public_baseurl
 
     def register(self, http_server: HttpServer) -> None:
         super().register(http_server)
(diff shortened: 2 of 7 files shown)


synapse/http/site.py
```diff
@@ -333,14 +333,28 @@ def _should_log_request(self) -> bool:
...
     Add a layer on top of another request that only uses the value of an
     X-Forwarded-For header as the result of C{getClientIP}.
+
+    XXX: I think the right way to do this is with request.setHost().
     """
 
+    def __init__(self, *args, **kw):
+        SynapseRequest.__init__(self, *args, **kw)
+
+        forwarded_header = self.getHeader(b"x-forwarded-proto")
```
richvdh, line 346: what do people think about this? we don't currently document that this is a thing that ought to be set (and indeed, matrix.org's haproxy doesn't). Should we? or are there better ways to solve this problem?

clokep:
When would requests arrive at the wrong URL? (Is this due to delegation of some kind?)

richvdh:
> When would requests arrive at the wrong URL? 

The canonical CS API (and hence `public_baseurl` for matrix.org) is `https://matrix-client.matrix.org`, which goes directly to the synapse cluster, rather than via the general `matrix.org` cluster.

Mostly for historical reasons, it's also possible to log in via `https://matrix.org`. It seems like some clients default to that, which might be a thing we should look into getting them to change.

synapse/http/__init__.py
```diff
@@ -53,6 +54,37 @@ def stopProducing(self):
...
+
+    if request.isSecure() and host.port == 443:
+        # default port for https
+        return hostname
+
+    if not request.isSecure() and host.port == 80:
+        # default port for http
+        return hostname
+
+    return b":%i" % (host.port,)
```
clokep, line 85: Should this return hostname:port?

synapse/http/site.py
```diff
@@ -333,14 +333,28 @@ def _should_log_request(self) -> bool:
...
+        if forwarded_header is not None:
+            self._is_secure = forwarded_header.lower() == b"https"
```
clokep, line 348: Any reason not to just set `isSecure` in these two places?

synapse/rest/client/v1/login.py
```diff
@@ -373,6 +375,24 @@ def register(self, http_server: HttpServer) -> None:
...
+        if not self._public_baseurl:
+            raise SynapseError(400, "SSO requires a valid public_baseurl")
+
+        # if this isn't the expected hostname, redirect to the right one, so that we
+        # get our cookies back.
+        requested_uri = get_request_uri(request)
+        baseurl_bytes = self._public_baseurl.encode("utf-8")
+        if not requested_uri.startswith(baseurl_bytes):
+            i = requested_uri.index(b"/_matrix")
+            new_uri = baseurl_bytes[:-1] + requested_uri[i:]
```
clokep, line 387: It seems like it would be simpler to do `baseurl_bytes + request.uri`. Actually -- it seems odd that we rebuild the entire uri when we really only want to compare the scheme + host (+ port)?

synapse/http/site.py
clokep: I think this is pretty standard -- I don't know of another way to do this (and I don't see a security issue if this was sent directly via a client, it might break things but not in a "bad" way?)

synapse/http/__init__.py
richvdh: yes

synapse/http/site.py
richvdh: sorry, which two places?

synapse/rest/client/v1/login.py
richvdh: the idea is that we'll behave correctly here if `/_matrix` isn't at the root of the path. (Of course, there are plenty of other places that won't work...)

synapse/http/site.py
clokep: Sorry for the opaque comment! I guess I'm asking why there's a separate `_is_secure` and then a method for returning that value, instead of setting the `isSecure` property directly. (As part of this I also wonder why we set this in `__init__` vs. `getClientIP` which is dynamic.)

synapse/rest/client/v1/login.py
clokep: So this should redirect from https://foo.bar/whatever/_matrix/... -> https://public.baseurl/_matrix/...?

synapse/http/site.py
richvdh: well, isSecure() is defined by the base class to be a callable method, not a property.

synapse/rest/client/v1/login.py
richvdh: correct. I guess I should spell that out in a comment.

synapse/http/site.py
clokep: Ah-ha! I didn't realize that this was part of the base-class! 👍

tests/rest/client/v1/test_login.py
```diff
@@ -47,8 +47,9 @@
     HAS_JWT = False
 
 
-# public_base_url used in some tests
-BASE_URL = "https://synapse/"
+# synapse server name: used to populate public_base_url in some tests
+SYNAPSE_SERVER_PUBLIC_HOSTNAME = "synapse"
+BASE_URL = "http://%s/" % (SYNAPSE_SERVER_PUBLIC_HOSTNAME,)
```
richvdh, line 52: swapping to `http` here, because `FakeChannel.isSecure()` returns `False`, so synapse will see the requested uri as `http://synapse/...`.  Configuring that as the `public_baseurl` avoids the redirect.

tests/rest/client/v1/utils.py
```diff
@@ -550,6 +550,20 @@ def initiate_sso_login(
...
+        channel = make_request(
+            self.hs.get_reactor(),
+            self.site,
+            "GET",
+            urllib.parse.urlunsplit(("", "") + parts[2:]),
+            custom_headers=[
+                ("Host", parts[1]),
+            ],
+        )
+
```
richvdh, line 566: this was the easiest way of figuring out what the `Host` header should be set to.

tests/rest/client/v2_alpha/test_auth.py
```diff
@@ -161,7 +161,7 @@ class UIAuthTests(unittest.HomeserverTestCase):
 
     def default_config(self):
         config = super().default_config()
-        config["public_baseurl"] = "https://synapse.test"
+        config["public_baseurl"] = "http://synapse.test"
```
richvdh, line 164: as above

richvdh:
I think this is *finally* ready-for-review after a week of banging my head against various infrastructure-related problems. The commits should be reviewable independently.

anoadragon453 (approved):
This looks pretty reasonable to me.

tests/rest/client/v1/test_login.py
richvdh: fair enough. I've updated some comments.

PR #9438, opened 2021-02-18
Regenerate exact thumbnails if missing

erikjohnston:
This doesn't re-fetch remote media if we have also deleted that, but I think that's something we can add in the future if we want

### DIFF
--- a/synapse/rest/media/v1/thumbnail_resource.py
+++ b/synapse/rest/media/v1/thumbnail_resource.py
@@ -114,6 +114,7 @@ class ThumbnailResource(DirectServeJsonResource):
             m_type,
             thumbnail_infos,
             media_id,
+            media_id,
             url_cache=media_info["url_cache"],
             server_name=None,
         )
@@ -269,6 +270,7 @@ class ThumbnailResource(DirectServeJsonResource):
             method,
             m_type,
             thumbnail_infos,
+            media_id,
             media_info["filesystem_id"],
             url_cache=None,
             server_name=server_name,
@@ -282,6 +284,7 @@ class ThumbnailResource(DirectServeJsonResource):
         desired_method: str,
         desired_type: str,
         thumbnail_infos: List[Dict[str, Any]],
+        media_id: str,
         file_id: str,
         url_cache: Optional[str] = None,
         server_name: Optional[str] = None,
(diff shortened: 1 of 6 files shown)


synapse/rest/media/v1/thumbnail_resource.py
```diff
@@ -316,9 +319,53 @@ async def _select_and_respond_with_thumbnail(
...
+                    request,
+                    responder,
+                    file_info.thumbnail_type,
+                    file_info.thumbnail_length,
+                )
+                return
+
+            # If we can't find the thumbnail we regenerate it. This can happen
+            # if e.g. we've deleted the thumbnails but still have the original
+            # image somewhere.
```
clokep, line 334: I think I'm missing how we know we have the original file still? Or do we expect the generate methods to explode in that case?

erikjohnston: That is not an unreasonable question. Yes, it will explode, but we should handle it more explicitly.

synapse/rest/media/v1/thumbnail_resource.py
```diff
@@ -316,9 +319,60 @@ async def _select_and_respond_with_thumbnail(
...
+            # thumbnails), and b) have already calculated that appropriate
+            # width/height/method so we can just call the "generate exact"
+            # methods.
+
+            # First let's check that we do actually have the original image
+            # still. This will throw a 404 if we don't.
+            # TODO: We should refetch the thumbnails for remote media.
+            await self.media_storage.ensure_media_is_in_local_cache(
+                FileInfo(server_name, file_id, url_cache=url_cache)
+            )
```
clokep, line 348: Ah, this is actually the first line of `generate_remote_exact_thumbnail` and `generate_local_exact_thumbnail` 🤦 so maybe we don't need to check this again?

clokep (approved):
Seems reasonable! Not sure if it is worth including any tests for this case?

synapse/rest/media/v1/thumbnail_resource.py
erikjohnston: The second time it gets called is fast (just a check if the path exists), so I think its better to be explicit by keeping this here (and I think its good to check again in the functions to avoid foot guns)

erikjohnston:
Have a test :)

clokep (approved):
Looks reasonable if tests pass! 👍

PR #9442, opened 2021-02-18
Add type hints to DictionaryCache and TTLCache.

clokep:
As the title says. This turned out harder than I expected and I was expecting to add them to all the caches, but there's some oddities with the way we use a sentinel in this code that I figured is worth a discussion so we can hopefully solve it once and use the same pattern in other places.


synapse/util/caches/ttlcache.py
```diff
@@ -40,26 +45,27 @@ def __init__(self, cache_name, timer=time.time):
...
+        if e is not SENTINEL:
+            assert isinstance(e, _CacheEntry)
             self._expiry_list.remove(e)
 
         entry = _CacheEntry(expiry_time=expiry, ttl=ttl, key=key, value=value)
         self._data[key] = entry
         self._expiry_list.add(entry)
 
-    def get(self, key, default=SENTINEL):
+    def get(self, key: KT, default: T = SENTINEL) -> Union[VT, T]:  # type: ignore
```
clokep, line 68: I could not get mypy to not complain about this:

> synapse/util/caches/ttlcache.py:68: error: Incompatible default for argument "default" (default has type "object", argument has type "T")  [assignment]

We `dictionary_cache` seems to have a magic `_Sentinel` enum to help with this, but I couldn't get it to work.

clokep: I'm fairly keen to figure this out since we'd want to use the same pattern for all the caches FTR.

erikjohnston: After some testing I think doing `SENTINEL = object()  # type: Any` works. It's a bit cheeky since `SENTINEL` isn't a `T`, but I think its probably the easiest.

erikjohnston (commented):
Mostly looks good to me, bar the `SENTINEL` issue

erikjohnston (approved):

PR #9445, opened 2021-02-19
Add endpoint to get an event at a given timestamp - MSC3030

MadLittleMods:
Add endpoint to get an event at a given timestamp ([MSC3030](https://github.com/matrix-org/matrix-doc/pull/3030))

Complement tests: https://github.com/matrix-org/complement/pull/178

Client API endpoint. This will also go and fetch from the federation API endpoint if unable to find an event locally or we found an extremity with possibly a closer event we don't know about.
```
GET /_matrix/client/unstable/org.matrix.msc3030/rooms/<roomID>/timestamp_to_event?ts=<timestamp>&dir=<direction>
{
    "event_id": ...
    "origin_server_ts": ...
}
```

Federation API endpoint:
```
GET /_matrix/federation/unstable/org.matrix.msc3030/timestamp_to_event/<roomID>?ts=<timestamp>&dir=<direction>
{
    "event_id": ...
    "origin_server_ts": ...
}
```

---

The initial implementation spawned from resurrecting old commits from @erikjohnston, https://gitlab.matrix.org/new-vector/synapse/-/compare/develop...kraken

### Dev notes

```
COMPLEMENT_DIR=../complement ./scripts-dev/complement.sh TestJumpToDateEndpoint
```

 - [`get_events_around(room_id, event_id, before_limit, after_limit, event_filter)`](https://github.com/matrix-org/synapse/blob/428174f90249ec50f977b5ef5c5cf9f92599ee0a/synapse/storage/databases/main/stream.py#L800)
 - [`find_first_stream_ordering_after_ts_txn(txn, ts)`](https://github.com/matrix-org/synapse/blob/57501d919458f71f6505e7474e9825c00bc8ec87/synapse/storage/databases/main/event_push_actions.py#L573)
 - `get_room_event_before_stream_ordering`

---

```python
all_backward_extremity_query = """
    SELECT event_id FROM event_backward_extremities
    WHERE room_id = ?
    LIMIT 100
"""

all_forward_extremity_query = """
    SELECT event_id FROM event_forward_extremities
    WHERE room_id = ?
    LIMIT 100
"""

txn.execute(all_backward_extremity_query, (room_id,))
all_backward_extremties = txn.fetchall()
txn.execute(all_forward_extremity_query, (room_id,))
all_forward_extremties = txn.fetchall()
logger.info(
    "all_forward_extremties=%s all_backward_extremties=%s",
    all_forward_extremties,
    all_backward_extremties,
)
```

---

Function overloads:

 - https://treyhunner.com/2018/10/asterisks-in-python-what-they-are-and-how-to-use-them/#Keyword-only_arguments_without_positional_arguments

### Todo

 - [ ] Make sure this also works with historical messages from MSC2716
 - [x] If your local server has only backfilled a little bit of the room and you ask for messages in the distant past (past the backwards extremities), the server should probably try to backfill some messages from back then. Or just ask another homeserver
 - [x] Do we want to add a `?dir=[f|b]` query parameter so you can specify which direction from the date you want to get the message from. For example when I want to use this in combination with MSC2716 to find a place to import messages, I want to get before/after a certain date vs the closest date.

### Pull Request Checklist


erikjohnston:
I think we probably need to think a little bit about that query performance, I don't it's terribly efficient as it stands. I think its entirely solvable though, so this is fine to prove that the MSC is possible :+1:

erikjohnston:
I don't think there is anything to do here until the MSC progresses?

MadLittleMods:
@erikjohnston I'm going to be exploring this MSC next after MSC2716 in order to make static archives more efficient. Use server-side rendered Hydrogen with this endpoint to show what happened on that day in the archive.

synapse/rest/client/v1/room.py
```diff
@@ -1389,6 +1389,34 @@ def register_txn_path(servlet, regex_string, http_server, with_get=False):
...
+        self._store = hs.get_datastore()
+
+    async def on_GET(
+        self, request: SynapseRequest, room_id: str
+    ) -> Tuple[int, JsonDict]:
+        requester = await self._auth.get_user_by_req(request)
+        await self._auth.check_joined_room(room_id, requester.to_string())
+
+        timestamp = parse_integer(request, "ts")
+        thread_id = parse_integer(request, "thread_id", 0)
```
MadLittleMods, line 1407: What is a `thread_id`?

MadLittleMods: It looks like it was part of a threading demo and we can remove it here, https://gitlab.matrix.org/new-vector/synapse/-/commit/e6896040c71f56aa59d8f8a2d1e788d0ad79b4ac

synapse/rest/client/room.py
```diff
@@ -1070,6 +1070,32 @@ def register_txn_path(
...
+        timestamp = parse_integer(request, "ts")
+        direction = parse_string(request, "dir", default="f", allowed_values=["f", "b"])
+
+        event_id = await self._store.get_event_for_timestamp(
+            room_id, timestamp, direction
+        )
+
+        return 200, {
+            "event_id": event_id,
+        }
```
MadLittleMods, line 1096: One problem I'm seeing with going with a `/timestamp_to_event` endpoint is that there is no way to actually paginate with `/messages` after getting the `event_id`.

We probably want to consider returning a pagination stream token to use with the `from`/`to` on `/messages`

MadLittleMods: Or just not worry about it, we can make a separate request to get one:

```go
// Get a pagination token for a single event
contextRes := c.MustDoFunc(t, "GET", []string{"_matrix", "client", "r0", "rooms", roomID, "context", someEventId}, client.WithContentType("application/json"), client.WithQueries(url.Values{
	"limit": []string{"0"},
}))
```

synapse/rest/client/room.py
```diff
@@ -1239,6 +1265,7 @@ def register_servlets(
     RoomAliasListServlet(hs).register(http_server)
     SearchRestServlet(hs).register(http_server)
     RoomCreateRestServlet(hs).register(http_server)
+    TimestampLookupRestServlet(hs).register(http_server)
```
MadLittleMods, line 1268: TODO: Add under experimental feature

synapse/storage/databases/main/events_worker.py
```diff
@@ -1626,3 +1626,104 @@ def _cleanup_old_transaction_ids_txn(txn):
...
+
+            txn.execute(all_backward_extremity_query, (room_id,))
+            all_backward_extremties = txn.fetchall()
+            txn.execute(all_forward_extremity_query, (room_id,))
+            all_forward_extremties = txn.fetchall()
+            logger.info(
+                "all_forward_extremties=%s all_backward_extremties=%s",
+                all_forward_extremties,
+                all_backward_extremties,
+            )
```
MadLittleMods, line 1664: When federated `hs2` joins, the room looks like the following for them. Should `$gHGXLXGcXfFlXdlelxOOYRw3KGRTQGvJxJKyVMotJ4w` also be a forward extremity? It's not referenced by any other `prev_events` on `hs2` yet. (@matrix-org/synapse-core any help here)

I understand that forward extremities are used for the next new event comes in as the `prev_events` and kinda wouldn't make sense to tie in `$gHGXLXGcXfFlXdlelxOOYRw3KGRTQGvJxJKyVMotJ4w` that way so there is a gap in my understanding and probably the description from https://github.com/matrix-org/synapse/blob/develop/docs/development/room-dag-concepts.md#forward-extremity

```
all_forward_extremeties=[('$AEhX_px4K4MGZ7fEtngZraN7KlK8h6AGEH5JRInI0to',)]
all_backward_extremeties=[('$qWzrvnib2Oh_XaSOVsERgWSICeR-zAqx4XkyJCBNJcA',)]
```
```
(oldest)
$ETDJdE8w0fhLCgbWOf0tMwaxTi7TCTT_CXC_XhpjqNI (1637131874827) - m.room.create
$DlVohXBO6Ts7pEOPxJ_CNlw95nYOhVfWPxuypxgsDRk (1637131874885) - m.room.member
$QliEnJH5dka-EBuHQ_9UDMmtuXtGlPcTDzHhHwRpKz0 (1637131874919) - m.room.power_levels
$kwejVaW0F90hdqOtohi5faGBj4qCy4LaJ7kjaRZQIeY (1637131874960) - m.room.join_rules
$gHGXLXGcXfFlXdlelxOOYRw3KGRTQGvJxJKyVMotJ4w (1637131874992) - m.room.history_visibility
<gap>
$AEhX_px4K4MGZ7fEtngZraN7KlK8h6AGEH5JRInI0to (1637131875789) - m.room.member
(newest)
```

---

The actual full picture on `hs1`, looks like:
```
(oldest)
$ETDJdE8w0fhLCgbWOf0tMwaxTi7TCTT_CXC_XhpjqNI (1637131874827) - m.room.create
$DlVohXBO6Ts7pEOPxJ_CNlw95nYOhVfWPxuypxgsDRk (1637131874885) - m.room.member
$QliEnJH5dka-EBuHQ_9UDMmtuXtGlPcTDzHhHwRpKz0 (1637131874919) - m.room.power_levels
$kwejVaW0F90hdqOtohi5faGBj4qCy4LaJ7kjaRZQIeY (1637131874960) - m.room.join_rules
$gHGXLXGcXfFlXdlelxOOYRw3KGRTQGvJxJKyVMotJ4w (1637131874992) - m.room.history_visibility
$5Npt2_DwUfzfXWbDYy2HMlgDPCHWa2esUn5F9dQecnk (1637131875032) - m.room.message
$qWzrvnib2Oh_XaSOVsERgWSICeR-zAqx4XkyJCBNJcA (1637131875091) - m.room.message
$AEhX_px4K4MGZ7fEtngZraN7KlK8h6AGEH5JRInI0to (1637131875789) - m.room.member
(newest)
```

synapse/federation/transport/server/__init__.py
```diff
@@ -324,6 +327,13 @@ def register_servlets(
             )
 
         for servletclass in DEFAULT_SERVLET_GROUPS[servlet_group]:
+            # Only allow the `/timestamp_to_event` servlet if msc3030 is enabled
+            if (
+                servletclass == FederationTimestampLookupServlet
+                and not hs.config.experimental.msc3030_enabled
+            ):
+                continue
```
MadLittleMods, line 335: Because `FederationTimestampLookupServlet` is bundled up in the `FEDERATION_SERVLET_CLASSES` which is an immutable tuple without any access to `hs`, I had to feature flag the endpoint this way. Open to any better solutions 😀

synapse/handlers/federation.py
```diff
@@ -268,36 +299,6 @@ async def _maybe_backfill_inner(
...
-                    old_d = joined_domains.get(dom)
-                    if old_d:
-                        joined_domains[dom] = min(d, old_d)
-                    else:
-                        joined_domains[dom] = d
-                except Exception:
-                    pass
-
-            return sorted(joined_domains.items(), key=lambda d: d[1])
-
```
MadLittleMods, line 300: Just de-nesting this function so we can re-use it

synapse/federation/federation_server.py
```diff
@@ -200,6 +200,37 @@ async def on_backfill_request(
...
+            body including `event_id`.
+        """
+        with (await self._server_linearizer.queue((origin, room_id))):
+            origin_host, _ = parse_server_name(origin)
+            await self.check_server_matches_acl(origin_host, room_id)
+
+            # We only try to fetch data from the local database
+            event_id = await self.store.get_event_for_timestamp(
+                room_id, timestamp, direction
+            )
```
squahtx, line 228: Depending on the history visibility of the room, can we end up revealing the IDs (and timestamps) of events to homeservers that would otherwise be unable to view them?

synapse/federation/transport/server/__init__.py
squahtx: Perhaps we could register the servlet manually?
```py
if "federation" in servlet_groups and hs.config.experimental.msc3030_enabled:
    FederationTimestampLookupServlet(...).register(resource)
```

synapse/federation/transport/client.py
```diff
@@ -148,6 +148,38 @@ async def backfill(
...
+            FEDERATION_UNSTABLE_PREFIX,
+            "/org.matrix.msc3030/timestamp_to_event/%s",
+            room_id,
+        )
+
+        args = {"ts": [str(timestamp)], "dir": [direction]}
+
+        return await self.client.get_json(
+            destination, path=path, args=args, try_trailing_slash_on_400=True
+        )
```
squahtx, line 181: We ought to validate that the response is a `dict` and that `response["event_id"]` is a string, as expected.

synapse/handlers/room.py
```diff
@@ -1217,6 +1218,114 @@ async def filter_evts(events: List[EventBase]) -> List[EventBase]:
...
+                        ex.args,
+                    )
+                    continue
+
+        if not local_event_id:
+            raise SynapseError(
+                404,
+                "Unable to find event from %s in direction %s" % (timestamp, direction),
+                errcode=Codes.NOT_FOUND,
+            )
```
squahtx, line 1324: The return type of `Optional[str]` may be misleading because we raise instead of returning `None`

synapse/federation/transport/client.py
squahtx: Also we'll never return `None` since the remote server will 404 if it can't find a suitable event.

synapse/handlers/room.py
```diff
@@ -1217,6 +1218,114 @@ async def filter_evts(events: List[EventBase]) -> List[EventBase]:
...
+
+                    continue
+                except Exception as ex:
+                    logger.exception(
+                        "Failed to fetch /timestamp_to_event from %s because of exception(%s) %s args=%s",
+                        domain,
+                        type(ex).__name__,
+                        ex,
+                        ex.args,
+                    )
```
squahtx, line 1316: We're going to be spamming the logs if remote homeservers can't find appropriate events to return, since 404s will raise exceptions.

synapse/handlers/room.py
```diff
@@ -1217,6 +1218,114 @@ async def filter_evts(events: List[EventBase]) -> List[EventBase]:
...
+        )
+
+        # If we found an extremity, we should probably ask another homeserver
+        # first about more history in between
+        is_extremity = False
+        if local_event_id:
+            is_extremity = await self.store.check_if_event_is_extremity(
+                room_id, local_event_id
+            )
+        if not local_event_id or is_extremity:
```
squahtx, line 1271: Do we still want to ask a remote homeserver if the extremity's direction matches `direction`?

eg. if a homeserver is up to date in a room, and we're looking for an event after `timestamp`, and that event happens to be the latest one, we'll end up querying other homeservers fruitlessly.

synapse/handlers/room.py
```diff
@@ -1217,6 +1218,114 @@ async def filter_evts(events: List[EventBase]) -> List[EventBase]:
...
+            likely_domains = [
+                domain for domain, depth in curr_domains if domain != self.server_name
+            ]
+
+            # Loop through each homeserver candidate until we get a succesful response
+            for domain in likely_domains:
+                try:
+                    remote_response = await self.transport_layer.timestamp_to_event(
+                        domain, room_id, timestamp, direction
+                    )
```
squahtx, line 1290: If no homeserver has an answer for `timestamp_to_event`, they'll spam each other with queries recursively.

eg. if a client were to ask for the closest event before January 1st 1970 or after 2030

synapse/http/servlet.py
```diff
@@ -79,6 +79,35 @@ def parse_integer(
...
+
+
+@overload
+def parse_integer_from_args(
+    args: Mapping[bytes, Sequence[bytes]],
+    name: str,
+    *,
+    required: Literal[True],
+) -> int:
+    ...
```
squahtx, line 98: +1 for more ergonomic type hints!

synapse/storage/databases/main/events_worker.py
```diff
@@ -1626,3 +1626,102 @@ def _cleanup_old_transaction_ids_txn(txn):
...
+    async def get_event_for_timestamp(
+        self, room_id: str, timestamp: int, direction: str
+    ) -> Optional[str]:
+        sql_template = """
+            SELECT event_id FROM events
+            WHERE
+                origin_server_ts %s ?
+                AND room_id = ?
+            ORDER BY origin_server_ts %s
+            LIMIT 1;
```
squahtx, line 1692: Do we want to exclude rejected events? (`rejections` table)

synapse/storage/databases/main/events_worker.py
squahtx: > `event_forward_extremities` can't be from `backfilled` events or `outliers`

After asking in #synapse-devs, it turns out this is by design. I think it's best to rename the method, otherwise we'll end up with multiple conflicting concepts of a forward extremity in the code.

synapse/handlers/room.py
MadLittleMods: It shouldn't do it recursively. Only the client endpoint goes and tries to ask other federated homeservers.

But the federation endpoint acts differently. When a homeserver asks via the federation endpoint, it only tries to get an event from it's own local database.

This means that someone making a `/client/timestamp_to_event` request will only result in a single `/federation/timestamp_to_event` request going out to each participating homeserver.

---

As a thought, for a hard-limit we could check that the timestamp comes after the `m.room.create` event. But with [MSC2716](https://github.com/matrix-org/matrix-doc/pull/2716) coming around the corner, events will have timestamps before the `m.room.create` event so that isn't a viable option.

synapse/handlers/room.py
MadLittleMods: We could skip asking other federated homeservers if the event matches `event_forward_extremities`.

Can change the concept to look for gaps instead of extremities.

synapse/federation/federation_server.py
```diff
@@ -200,6 +202,69 @@ async def on_backfill_request(
...
+            filtered_events = await filter_events_for_server(
+                self.storage,
+                origin,
+                [event],
+                # TODO: Is returning an event_id for an event that's hidden by
+                # history_visibility considered leaking? It looks like we
+                # already return redacted copies of hidden events in the
+                # `/backfill` endpoint.
+                redact=True,
```
MadLittleMods, line 242: Is returning an `event_id` for an event that's hidden by `history_visibility` considered leaking?
                
`/backfill` will already return redacted versions of events that aren't visible to them in the history.

synapse/federation/federation_server.py
MadLittleMods: Is returning an `event_id` for an event that's hidden by `history_visibility` considered leaking?
                
`/backfill` will already return redacted versions of events that aren't visible to them in the history. I've re-used the same `filter_events_for_server` function for the same behavior.

synapse/federation/transport/client.py
MadLittleMods: > We ought to validate that the response is a `dict` and that `response["event_id"]` is a string, as expected.

That seems reasonable except we don't do it for any of the other functions in this file AFAICT. Do we want to make this one different?

> Also we'll never return `None` since the remote server will 404 if it can't find a suitable event.

Another homeserver implementation could potentially return anything.

synapse/storage/databases/main/events_worker.py
```diff
@@ -1626,3 +1626,102 @@ def _cleanup_old_transaction_ids_txn(txn):
             "_cleanup_old_transaction_ids",
             _cleanup_old_transaction_ids_txn,
         )
+
+    async def check_if_event_is_extremity(self, room_id: str, event_id: str) -> bool:
```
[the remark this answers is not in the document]
MadLittleMods: Updated to `event_next_to_gap`

synapse/storage/databases/main/events_worker.py
MadLittleMods: Thanks @squahtx, updated to `event_next_to_gap` and the new focus on just gaps

synapse/federation/transport/server/__init__.py
MadLittleMods: That seems like a lot more boilerplate in the `...` to keep in sync with each other.

What do you prefer?

scripts-dev/complement.sh
```diff
@@ -65,4 +65,4 @@ if [[ -n "$1" ]]; then
 fi
 
 # Run the tests!
-go test -v -tags synapse_blacklist,msc2946,msc3083,msc2403,msc2716 -count=1 "${EXTRA_COMPLEMENT_ARGS[@]}" ./tests/...
+go test -v -tags synapse_blacklist,msc2946,msc3083,msc2403,msc2716,msc3030 -count=1 "${EXTRA_COMPLEMENT_ARGS[@]}" ./tests/...
```
MadLittleMods, line 68: TODO: Potentially revert before merge

scripts-dev/complement.sh
```diff
@@ -65,4 +65,4 @@ if [[ -n "$1" ]]; then
 fi
 
 # Run the tests!
-go test -v -tags synapse_blacklist,msc2946,msc3083,msc2403 -count=1 $EXTRA_COMPLEMENT_ARGS ./tests/...
+go test -v -tags synapse_blacklist,msc2946,msc3083,msc2403,msc3030 -count=1 $EXTRA_COMPLEMENT_ARGS ./tests/...
```
MadLittleMods, line 68: TODO: Potentially revert before merge

MadLittleMods:
Thanks for the review pass @squahtx, lots of good points and I think I have it in shape again! 😊

synapse/handlers/room.py
squahtx: My mistake, I misread the code. Please ignore my comment.

synapse/federation/transport/server/__init__.py
squahtx: I've got no strong preference here. Either way works for me.

synapse/federation/federation_server.py
squahtx: I wasn't aware of the precedent from `/backfill`, thank you!

If that endpoint allows a homeserver to retrieve all past `event_id`s and `origin_server_ts`s, then we don't need to use `filter_events_for_server` here.

synapse/federation/transport/client.py
squahtx: >> We ought to validate that the response is a dict and that response["event_id"] is a string, as expected.
>
> That seems reasonable except we don't do it for any of the other functions in this file AFAICT. Do we want to make this one different?

There's an ongoing push to have more thorough mypy coverage and this file is a big pain point. The intent is to eventually have all the methods in this file perform validation, but we're not there yet. With that said, it'd be okay to omit the validation for now since someone will get around to adding it eventually.

>> Also we'll never return None since the remote server will 404 if it can't find a suitable event.
>
> Another homeserver implementation could potentially return anything.

True, but do note that `get_json` claims to return either a `JsonDict` or `list` and never `None`:
https://github.com/matrix-org/synapse/blob/develop/synapse/http/matrixfederationclient.py#L961-L970

(I suspect that `Union[JsonDict, list]` annotation is misleading and we're going to have a very bad time getting mypy to accept this file)

synapse/handlers/room.py
squahtx: I'm not sure I explained it well.

Let's say that the local homeserver has 4 events, A, B, C and D, linked linearly. The events before A and after D are not known locally.

Requests 1, 2 and 3 come in, with the timestamp represented by a `|` and direction represented by an arrow:
```
          A    B    C    D                    other events
       <--x----x----x----x-->                  <--x--...

                                      <-| request 1
       request 2 |->
            request 3 |->

           <-| request 4
```

Requests 1 and 2 do the right thing currently.
For request 1, the homeserver finds event D, whose successors are not known locally, so it would ask over federation. There's a good chance other servers know of events closer to the timestamp so this is good.
Request 2 finds event C and no federation requests are done, which is also sensible. Event C may not actually be the closest, but it's hoped this is unlikely.

Request 3 finds event D. The current code would make a federation request, but I'm suggesting that it shouldn't, for the same reason request 2 doesn't make one: the section of history between request 3's timestamp and D is known
Request 4 is basically the same as 3, but mirrored to be backwards and the same reasoning should apply.

In summary, if we are looking backwards and we find a forward extremity*, then we should continue to ask over federation.
If we are looking forwards and we find a forward extremity*, then I suggest we return it.
and the same for backward extremities, with the direction flipped.

*an event whose successors aren't known, not the definition in the docs.

squahtx: Oh, I see the code's been changed so that request 1 wouldn't federate if D is in the `event_forward_extremities` table. That looks very sensible!

synapse/storage/databases/main/events_worker.py
```diff
@@ -1626,3 +1626,136 @@ def _cleanup_old_transaction_ids_txn(txn):
...
+            # forward gap.
+            #
+            # We can't just check `event_forward_extremities` directly because that
+            # doesn't include potential backfilled and outlier events.
+            #
+            # We consider any event that is an `event_forward_extremities` as
+            # the latest in the room and not next to a gap.
+            forward_edge_query = """
+                SELECT 1 FROM event_edges
+                LEFT JOIN rejections ON event_edges.prev_event_id == rejections.event_id
```
squahtx, line 1668: Checking for rejection does make sense here, but shouldn't it be the next event that we don't want to be rejected?
```suggestion
                LEFT JOIN rejections ON event_edges.event_id == rejections.event_id
```

synapse/storage/databases/main/events_worker.py
```diff
@@ -1626,3 +1626,136 @@ def _cleanup_old_transaction_ids_txn(txn):
...
+                LEFT JOIN event_forward_extremities ON event_edges.prev_event_id == event_forward_extremities.event_id
+                WHERE
+                    event_edges.room_id = ?
+                    AND event_edges.prev_event_id = ?
+                    /* If the event in question is a forward extremity, we will
+                     * just consider the forward gap as not a gap since it's one
+                     *  of the latest events in the room.
+                     */
+                    AND event_forward_extremities.event_id IS NOT NULL
```
squahtx, line 1677: If `event_id` has no forward edges and is in `event_forward_extremities`, the query will return no rows and we'll still class the event as next to a gap. It might be easier to do a separate query for `event_forward_extremities`.

synapse/federation/federation_server.py
```diff
@@ -200,6 +202,70 @@ async def on_backfill_request(
...
+        self, origin: str, room_id: str, timestamp: int, direction: str
+    ) -> Tuple[int, Dict[str, Any]]:
+        """When we receive a federated `/timestamp_to_event` request,
+        handle all of the logic for validating and fetching the event.
+
+        Args:
+            origin: The server we received the event from
+            room_id: Room to fetch the event from
+            timestamp: The point in time we should navigate from in the given
+                direction to find the closest event.
```
squahtx, line 215:
```suggestion
            timestamp: The point in time (inclusive) we should navigate from in the
                given direction to find the closest event.
```

synapse/federation/transport/client.py
```diff
@@ -148,6 +148,38 @@ async def backfill(
...
+    ) -> Optional[JsonDict]:
+        """
+        Calls a remote federating server at `destination` asking for their
+        closest event to the given timestamp in the given direction.
+
+        Args:
+            destination: Domain name of the remote homeserver
+            room_id: Room to fetch the event from
+            timestamp: The point in time we should navigate from in the given
+                direction to find the closest event.
```
squahtx, line 163:
```suggestion
            timestamp: The point in time (inclusive) we should navigate from in the
                given direction to find the closest event.
```

synapse/handlers/room.py
```diff
@@ -1220,6 +1221,114 @@ async def filter_evts(events: List[EventBase]) -> List[EventBase]:
...
+    ) -> str:
+        """Find the closest event to the given timestamp in the given direction.
+        If we can't find an event locally or the event we have locally is next to a gap,
+        it will ask other federated homeservers for an event.
+
+        Args:
+            requester: The user making the request according to the access token
+            room_id: Room to fetch the event from
+            timestamp: The point in time we should navigate from in the given
+                direction to find the closest event.
```
squahtx, line 1246:
```suggestion
            timestamp: The point in time (inclusive) we should navigate from in the
                given direction to find the closest event.
```

synapse/handlers/room.py
```diff
@@ -1220,6 +1221,114 @@ async def filter_evts(events: List[EventBase]) -> List[EventBase]:
...
+                    remote_event_id = remote_response.get("event_id", None)
+                    if remote_event_id:
+                        # TODO: Do we want to persist this as an extremity?
+                        # TODO: I think ideally, we would try to backfill from
+                        # this event and run this whole
+                        # `get_event_for_timestamp` function again to make sure
+                        # they didn't give us an event from their gappy history.
+                        return remote_event_id
+
+                    continue
```
squahtx, line 1311:
```suggestion
```
`continue` is redundant here

synapse/handlers/room.py
```diff
@@ -1220,6 +1221,114 @@ async def filter_evts(events: List[EventBase]) -> List[EventBase]:
...
+                except Exception as ex:
+                    logger.debug(
+                        "Failed to fetch /timestamp_to_event from %s because of exception(%s) %s args=%s",
+                        domain,
+                        type(ex).__name__,
+                        ex,
+                        ex.args,
+                    )
+                    continue
```
squahtx, line 1320:
```suggestion
```

synapse/handlers/room.py
```diff
@@ -1220,6 +1221,114 @@ async def filter_evts(events: List[EventBase]) -> List[EventBase]:
...
+                        continue
+
+                    remote_event_id = remote_response.get("event_id", None)
+                    if remote_event_id:
+                        # TODO: Do we want to persist this as an extremity?
+                        # TODO: I think ideally, we would try to backfill from
+                        # this event and run this whole
+                        # `get_event_for_timestamp` function again to make sure
+                        # they didn't give us an event from their gappy history.
+                        return remote_event_id
```
squahtx, line 1309: I wonder if it'd be a good idea to amend the MSC to include `origin_server_ts` in the response, so we could discard events that are worse (further away) than the one we found locally.

squahtx (commented):
Thank you for making the changes! I gave it another review pass

synapse/storage/databases/main/events_worker.py
```diff
@@ -1626,3 +1626,136 @@ def _cleanup_old_transaction_ids_txn(txn):
             "_cleanup_old_transaction_ids",
             _cleanup_old_transaction_ids_txn,
         )
+
+    async def event_next_to_gap(self, room_id: str, event_id: str) -> bool:
```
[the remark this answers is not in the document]
MadLittleMods: Ahh, I was trying to match your previous review nit around this but I see more what you're after with `is` somewhere in it, https://github.com/matrix-org/synapse/pull/9445#discussion_r756031441

synapse/handlers/room.py
MadLittleMods: Seems decent 👍 Everything is up in the air and we can change anything. I am even considering whether we can add a pagination token to start paginating on `/messages` right away, https://github.com/matrix-org/synapse/pull/9445#discussion_r747943150

synapse/federation/federation_server.py
MadLittleMods: 👍 Removed `filter_events_for_server`. We already have `check_server_matches_acl` to see if a server is allowed in the room.

synapse/storage/databases/main/events_worker.py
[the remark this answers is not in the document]
MadLittleMods: Renamed to `is_event_next_to_gap`

synapse/storage/databases/main/events_worker.py
```diff
@@ -1626,3 +1626,136 @@ def _cleanup_old_transaction_ids_txn(txn):
...
+            # If the event in question is listed as a backward extremity, it's
+            # next to a gap.
+            backward_extremity_query = """
+                SELECT 1 FROM event_backward_extremities
+                LEFT JOIN rejections USING (event_id)
+                WHERE
+                    room_id = ?
+                    AND event_id = ?
+                    /* Make sure event is not rejected */
+                    AND rejections.event_id IS NULL
```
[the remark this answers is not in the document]
MadLittleMods: Great spot 👀, I've tried to update the comments to reflect this subtlety,

```sql
/* It's not a valid edge if the event referencing our event in
 * question is rejected.
 */
AND rejections.event_id IS NULL
```

synapse/storage/databases/main/events_worker.py
MadLittleMods: Great spot 👀, I've tried to update the comments to reflect this subtlety,

```sql
/* It's not a valid edge if the event referencing our event in
 * question is rejected.
 */
AND rejections.event_id IS NULL
```

synapse/storage/databases/main/events_worker.py
MadLittleMods: Updated and explained this edge case in the comments 👍

synapse/storage/databases/main/events_worker.py
```diff
@@ -1626,3 +1626,173 @@ def _cleanup_old_transaction_ids_txn(txn):
...
+                LIMIT 1
+            """
+
+            # If the event in question has its prev_events listed as a backward
+            # extremity, it's next to a backwards gap.
+            #
+            # We need to check this before the forward edges below as the event
+            # in question could be a forward extremity while still having a
+            # backwards gap.
+            for prev_event_id in event.prev_event_ids():
```
MadLittleMods, line 1703: Calling out a bigger change here since the last review cycle 👆

Previously, there was a logic flaw in the way I was detecting backwards gaps where I was just trying to check if the event in question was a backward extremity. But this would never be the case because when we persist an event, we clear it as a backward extremity and record it's `prev_events` as the new backward extremities.

So we really need to just whether the event in question has it's `prev_events` listed as backward extremity which is what the code is doing now.

synapse/federation/transport/client.py
MadLittleMods: > True, but do note that `get_json` claims to return either a `JsonDict` or `list` and never `None`:
https://github.com/matrix-org/synapse/blob/develop/synapse/http/matrixfederationclient.py#L961-L970

Ahh, I was looking at [`MatrixFederationRequest` which had `Optional[JsonDict]`](https://github.com/matrix-org/synapse/blob/e8ae94a22367a81049582dfdb16c743a45ca4e8b/synapse/http/matrixfederationclient.py#L179) 🤦‍♂️ (correct code is [under `MatrixFederationHttpClient` which @squahtx linked](https://github.com/matrix-org/synapse/blob/develop/synapse/http/matrixfederationclient.py#L961-L970))

I created `TimestampToEventResponse(TypedDict)` to better type what's being returned here

synapse/handlers/room.py
MadLittleMods: > In summary, if we are looking backwards and we find an event with no successors, then we should continue to ask over federation.
> If we are looking forwards and we find an event with no successors, then I suggest we return it.
and the same for events with no predecessors, with the direction flipped.

Updated 👍 and split the logic into `is_event_next_to_backward_gap` and `is_event_next_to_forward_gap`

synapse/handlers/room.py
MadLittleMods: Added `origin_server_ts` to the federation endpoint and the optimization to only return the remote event if closer than the local event.

Is it worth adding to the client endpoint as well?

MadLittleMods: I've added it to the client endpoint as well because it could be useful to make sure that a requested event is not 24 hours older than the timestamp requested for day-by-day archive for example.

synapse/federation/transport/client.py
```diff
@@ -148,6 +157,46 @@ async def backfill(
...
+        )
+
+        args = {"ts": [str(timestamp)], "dir": [direction]}
+
+        remote_response = await self.client.get_json(
+            destination, path=path, args=args, try_trailing_slash_on_400=True
+        )
+        assert isinstance(remote_response, dict)
+        assert remote_response.get("event_id", None) is not None
+        assert remote_response.get("origin_server_ts", None) is not None
```
squahtx, line 196: `assert` may not be the best tool for the job here. `assert`s are more for catching programmer mistakes rather than validation and are skipped entirely in `python -O` mode.

I took a look at how we do validation of federation responses. In `FederationClient` we raise `InvalidResponseError`s for malformed responses (see `get_room_hierarchy` in particular).

`InvalidResponseError` is defined in `federation_client.py`, which maybe suggests that's where validation should go? I'm not too clear on the distinction between `FederationClient` and `TransportLayerClient`.

We also need to check that `event_id` is a str and `origin_server_ts` is an int.
```py
        if not isinstance(remote_response, dict): ...
        if not isinstance(remote_response.get(...), str): ...
        if not isinstance(remote_response.get(...), int): ...
```

synapse/handlers/room.py
```diff
@@ -1220,6 +1221,137 @@ async def filter_evts(events: List[EventBase]) -> List[EventBase]:
...
+            room_id, timestamp, direction
+        )
+        logger.debug(
+            "get_event_for_timestamp: locally, we found event_id=%s closest to timestamp=%s",
+            local_event_id,
+            timestamp,
+        )
+
+        # Check for gaps in the history where events could be hiding in between
+        # the timestamp given and the event we were able to find localy
```
squahtx, line 1268:
```suggestion
        # the timestamp given and the event we were able to find locally
```

synapse/handlers/room.py
```diff
@@ -1220,6 +1221,137 @@ async def filter_evts(events: List[EventBase]) -> List[EventBase]:
...
+                    # they didn't give us an event from their gappy history.
+                    remote_event_id = remote_response["event_id"]
+                    origin_server_ts = remote_response["origin_server_ts"]
+
+                    # Only return the remote event if it's closer than the local event
+                    if not local_event or (
+                        local_event
+                        and abs(origin_server_ts - timestamp)
+                        < abs(local_event.origin_server_ts - timestamp)
+                    ):
```
squahtx, line 1334:
```suggestion
                    if not local_event or (
                        abs(origin_server_ts - timestamp)
                        < abs(local_event.origin_server_ts - timestamp)
                    ):
```
The 2nd `local_event` check looks redundant

synapse/storage/databases/main/events_worker.py
squahtx: I'm confused about what a backward extremity is now. Previously I thought it was an event we had, but whose `prev_events` we did not all have. After looking at the [code](https://github.com/matrix-org/synapse/blob/a4521ce0a8d252e77ca8bd261ecf40ba67511a31/synapse/storage/databases/main/events.py#L2136), it seems to agree with what you're saying.

When I prod at my synapse's database, I see a lot of events in `event_backward_extremities` that are also in `events` and some that aren't. I'm guessing the ones that are in `events` are outliers. Does an outlier always have one `prev_event` that is also a backward extremity?

synapse/storage/databases/main/events_worker.py
```diff
@@ -1626,3 +1626,189 @@ def _cleanup_old_transaction_ids_txn(txn):
...
+            Boolean indicating whether it's an extremity
+        """
+
+        def is_event_next_to_gap_txn(txn) -> bool:
+            # If the event in question is a forward extremity, we will just
+            # consider any potential forward gap as not a gap since it's one of
+            # the latest events in the room.
+            #
+            # We can't combine this query with the `forward_edge_query` below
+            # because if the event in question has no forward edges(isn't
```
squahtx, line 1696:
```suggestion
            # because if the event in question has no forward edges (isn't
```

synapse/storage/databases/main/events_worker.py
```diff
@@ -1626,3 +1626,189 @@ def _cleanup_old_transaction_ids_txn(txn):
...
+                    AND event_id = ?
+                LIMIT 1
+            """
+
+            # Check to see whether the event in question is already referenced
+            # by another event. If we don't see any edges, we're next to a
+            # forward gap.
+            #
+            # We can't just only check `event_forward_extremities` directly
+            # because that doesn't include backfilled and outlier events.
```
squahtx, line 1713: This comment is confusing. It seems to imply that `event_forward_extremities` almost does what we want, when it's the opposite - events in that list are considered to not be next to gaps.

```suggestion
```

squahtx (commented):
Thank you for adding the comments to the is-next-to-gap code! They're going to prove useful for future maintainers.

It's all minor comments and suggestions now. Overall it's looking good

synapse/storage/databases/main/events_worker.py
MadLittleMods: >  Does an `outlier` always have one `prev_event` that is also a backward extremity?

Not necessarily:

 1. I think an `outlier` could reference a a non-outlier event that we already have persisted. Imagine if we backfill an event from `bob` and we're missing a `prev_event`. Synapse will go and fetch the missing `prev_event` and store it as an `outlier` but that event could be referencing events we already have persisted normally.
    [Mermaid live editor playground](https://mermaid-js.github.io/mermaid-live-editor/edit/#eyJjb2RlIjoiZmxvd2NoYXJ0IEJUXG4gICAgc3ViZ3JhcGggcGVyc2lzdGVkXG4gICAgICAgIEMtLS0tPkItLS0-QVxuICAgIGVuZFxuXG4gICAgc3ViZ3JhcGggYmFja2ZpbGxlZFxuICAgICAgICBib2JFdmVudFxuICAgIGVuZFxuXG4gICAgYm9iRXZlbnQtLT5tZW1iZXJCb2IwKFtcIm0ucm9vbS5tZW1iZXIgKGJvYilcIl0pXG5cbiAgICBtZW1iZXJCb2IwLS0-QiIsIm1lcm1haWQiOiJ7XG4gIFwidGhlbWVcIjogXCJkZWZhdWx0XCJcbn0iLCJ1cGRhdGVFZGl0b3IiOmZhbHNlLCJhdXRvU3luYyI6dHJ1ZSwidXBkYXRlRGlhZ3JhbSI6ZmFsc2V9)
    
 1. Events can be created with no `prev_events`. Related discussion where I am playing around no `prev_event` events for MSC2716, https://github.com/matrix-org/synapse/pull/11243#discussion_r750816377

---

> I'm confused about what a backward extremity is now.

Looking at the code, [`outliers` are never cleared as a `event_backward_extremities`](https://github.com/matrix-org/synapse/blob/70cbb1a5e311f609b624e3fae1a1712db639c51e/synapse/storage/databases/main/events.py#L2214). So the `event_backward_extremities` you also see persisted in `events` are `outliers`. 

The other backward extremities are just the oldest events we know about, not necessarily persisted. Backward extremities just need to be `event_id` strings and will most likely be the `prev_events` of the oldest events we have persisted.

I wrote a bit about them in https://github.com/matrix-org/synapse/blob/develop/docs/development/room-dag-concepts.md#backwards-extremity (also see the `outliers` section) but I think it needs another iteration to get it more correct.

synapse/storage/databases/main/events_worker.py
MadLittleMods: Good point, it's a bit off. I've added the clarification to the `forward_extremity_query` comments instead and re-framed it to explain why we can't just use it by itself.

```
# `event_forward_extremities` does not include backfilled or outlier
# events so we can't rely on it to find forward gaps. We can only
# use it to determine whether a message is the latest in the room.
```

synapse/federation/transport/client.py
MadLittleMods: It looks like some of the `TransportLayerClient` functions are cheating the types and just assuming it returns `JsonDict` when they could also return `List`

I've created a `timestamp_to_event` function in `federation_client.py` to do the rest of the validation like the other functions there ⏩

synapse/federation/federation_client.py
```diff
@@ -1517,6 +1517,83 @@ async def send_request(
...
+            InvalidResponseError when the response does not have the correct
+            keys or wrong types
+        """
+        remote_response = await self.transport_layer.timestamp_to_event(
+            destination, room_id, timestamp, direction
+        )
+
+        if not isinstance(remote_response, dict):
+            raise InvalidResponseError(
+                "Response must be a JSON dictionary but received %s" % remote_response
```
squahtx, line 1552: minor suggestion: Could we use `%r` for presenting untrusted data? In case it contains weird ascii junk that we don't want in our logs.

The same for `from_json_dict` below

squahtx (approved):
LGTM!

Sorry for the lengthy review process. I really appreciate the updates to the comments and documentation you're doing!

synapse/storage/databases/main/events_worker.py
squahtx: Thanks for clearing things up for me!

MadLittleMods:
Thanks for the review @squahtx! 🐓 The daily regular drum of review was fantastic to blast this forward quickly and got this to a much better state 🤘

PR #9449, opened 2021-02-19
Include newly added sequences in the port DB script

### LINKED ISSUE #9344: Postgres port script fails to set correct initial values for some sequences
## TL;DR: Uninitialized sequences
* `receipts_sequence` (`receipts_linearized`)
* `account_data_sequence` (`room_account_data`)
* `device_inbox_sequence` (`device_inbox`)

## Steps to reproduce
I have synapse installed through the ubuntu package from the https://packages.matrix.org/debian/ repository, at version `matrix-synapse-py3/unknown,now 1.26.0+focal1 amd64 [installed]`

I ran `synapse_port_db` to migrate my synapse install from sqlite to postgres.

After the script was complete I updated the database section of my homeserver.yaml to point to the new postgres install, and started up synapse.

The problem was that synapse then failed to come up (and, as an aside, `journalctl --since=today -u matrix-synapse` was uninformative as to why it was failing). Looking at the homeserver log I saw an error message beginning with 
```
2021-02-07 16:54:16,547 - twisted - 254 - ERROR - None -  Postgres sequence 'receipts_sequence' is inconsistent with associated
2021-02-07 16:54:16,547 - twisted - 254 - ERROR - None -  table 'receipts_linearized'.
```
It then went on to provide an SQL statement to run to fix that problem.

I checked (`select * from receipts_sequence`) before running that setval statement and the current value of the sequence was in fact 1, indicating it had never been called.

After addressing that issue I got error messages about two more sequences (account_data_sequence and device_inbox_sequence) before I was able to start up my synapse.

The printed error message said "_This can happen if Synapse has been downgraded and then upgraded again, or due to a bad migration._", but I'm quite sure I've never downgraded synapse or manually run a migration against the sqlite db.

### LINKED ISSUE #9382: Postgres port script breaks the database after synapse 1.26, doesn't allow creating new DMs.
### Description

I upgraded from sqlite to postgres using `synapse_port_db`. The script ran successfully, except for the issues I mentioned in https://github.com/matrix-org/synapse/issues/9344. A few days after that, however, a new user on a different HS tried to initiate a DM with me, and I found myself unable to accept the invite with a "Failed to join room: Internal Server Error". My synapse logs indicate that the problem is `duplicate key value violates unique constraint "event_auth_chains_c_seq_index"`.

Searching the history of #synapse:matrix.org for `event_auth_chains_c_seq_index` turned up a couple more cases where people saw that error and were unable to create new rooms or DMs after migrating to postgres, so I know it's not just me. In the past the advice they had been given was to wipe their server and start fresh, but I think that kind of a data loss bug (or a db state bug for which the only remedy is data loss) is pretty unacceptable for a messaging service.

This is particularly pernicious because the postgres migration script completes successfully, and it's only possibly a few days later, when someone tries to create a new room, that you find the database is borked. This would be less of a problem if there were a clear, documented workaround (like there is for the bugs in #9344), but for this bug the crowd in #synapse:matrix.org does not know of a nondestructive fix for the problem.

### Steps to reproduce

- upgrade from sqlite to postgres on synapse 1.26
- have another user try to start a new DM room with you.
- Observe that this fails, with an error 

### Version information

- **Homeserver**:

If not matrix.org:

- **Version**: {"server_version":"1.26.0","python_version":"3.8.5"}

- **Install method**: apt: `matrix-synapse-py3/unknown,now 1.26.0+focal1 amd64 [installed]`

- **Platform**: Ubuntu 20.04.2 LTS

clokep:
Fixes #9344 and fixes #9382.

This adds additional sequences to the port db script:

* `account_data_sequence`
* `receipts_sequence`
* `event_auth_chain_id`

It also calls `check_consistency` for `event_auth_chain_id`.

#9344 also talks about `account_data_sequence` being unported, but that is in the script. There was a slight bug in that function though where we didn't await the call to restart the sequence. Maybe that was the issue.

I'm not 100% confident this is doing the correct thing, but I *think* it is.

Should we be checking the consistency of the other sequences, by the way?

### DIFF
--- a/scripts/synapse_port_db
+++ b/scripts/synapse_port_db
@@ -22,7 +22,7 @@ import logging
 import sys
 import time
 import traceback
-from typing import Dict, Optional, Set
+from typing import Dict, Iterable, Optional, Set
 
 import yaml
 
@@ -629,7 +629,9 @@ class Porter(object):
             await self._setup_state_group_id_seq()
             await self._setup_user_id_seq()
             await self._setup_events_stream_seqs()
-            await self._setup_device_inbox_seq()
+            await self._setup_sequence(
+                "device_inbox_sequence", ("device_inbox", "device_federation_outbox")
+            )
 
             # Step 3. Get tables.
             self.progress.set_state("Fetching tables")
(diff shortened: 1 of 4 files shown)


scripts/synapse_port_db
```diff
@@ -629,7 +629,13 @@ class Porter(object):
             await self._setup_state_group_id_seq()
             await self._setup_user_id_seq()
             await self._setup_events_stream_seqs()
-            await self._setup_device_inbox_seq()
+            await self._setup_sequence(
+                "device_inbox_sequence", ("device_inbox", "device_federation_outbox")
+            )
```
clokep, line 634: It seems the other sequences don't *quite* follow the same pattern so have custom functions. It could probably be abstract to handle them, but...not sure it is worth it.

synapse/storage/database.py
```diff
@@ -429,6 +429,12 @@ def get_chain_id_txn(txn):
         self.event_chain_id_gen = build_sequence_generator(
             engine, get_chain_id_txn, "event_auth_chain_id"
         )
+        with make_conn(
+            database_config, engine, "event_auth_chain_consistency"
+        ) as db_conn:
```
clokep, line 434: This feels a little fish-y, but might be the right thing to do?

clokep: Oh hmm this needs to ensure the migration has run before being executed. 😢

erikjohnston:
> #9344 also talks about `account_data_sequence` being unported, but that is in the script. There was a slight bug in that function though where we didn't await the call to restart the sequence. Maybe that was the issue.

if we don't await on a function then it likely doesn't get run, so that would explain it I think?

erikjohnston:
> Should we be checking the consistency of the other sequences, by the way?

Gah, I have a forgotten branch that ensures we always call `check_consistency` :facepalm:

erikjohnston (approved):

PR #9450, opened 2021-02-19
MSC2918 Refresh tokens implementation

sandhose:
This implements refresh tokens, as defined by [MSC2918](https://github.com/matrix-org/matrix-doc/pull/2918)

This MSC has been implemented client side in Hydrogen Web: https://github.com/vector-im/hydrogen-web/pull/235

The basics of the MSC works: requesting refresh tokens on login, having the access tokens expire, and using the refresh token to get a new one. 
Here are the remaining things to do:

 - ~~write tests (and ensure the current tests aren't broken)~~ done
 - ensure it works for other login types (incl. app services logins)
 - make it works for puppeted users
 - ~~write the schema migration for postgres (sqlite only ATM)~~ done
 - ~~also implement it when registering users~~ done
 - ~~make the access token duration configurable (and maybe also allow to disable the rotation of refresh tokens?)~~ done
 - add a background update that deletes old refresh tokens?

### Pull Request Checklist


clokep:
From conversation with @sandhose today this could use an design review (not necessarily a code review) with an eye toward gotchas and things that might completely break with this approach.

synapse/handlers/auth.py
```diff
@@ -754,13 +756,48 @@ def _auth_dict_for_flows(
...
+        ) = await self.get_refresh_token_for_user_id(
+            user_id=existing_token.user_id, device_id=existing_token.device_id
+        )
+        access_token = await self.get_access_token_for_user_id(
+            user_id=existing_token.user_id,
+            device_id=existing_token.device_id,
+            valid_until_ms=valid_until_ms,
+            refresh_token_id=new_refresh_token_id,
+        )
+        await self.store.invalidate_refresh_token(existing_token.token_id)
```
erikjohnston, line 781: We'll need to make this API idempotent somehow, i.e. if the client never gets the response to their `/refresh` request they need to be able to retry it without failing. One possible option would be to not invalidate the refresh token until the server sees the new access token in use, but there are probably other ways of doing so.

synapse/storage/databases/main/schema/delta/60/01refresh_tokens.sql.sqlite
```diff
@@ -0,0 +1,31 @@
+CREATE TABLE refresh_tokens (
```
erikjohnston, line 1: Don't we also need a `postgres` schema change?

erikjohnston (commented):
I think this mostly makes sense, beyond needing to make the requests idempotent

synapse/storage/databases/main/schema/delta/60/01refresh_tokens.sql.sqlite
sandhose: That's done now.

synapse/handlers/auth.py
sandhose: I'm now doing this by checking if any "child" refresh token was already used. I consider that a refresh token has been used either if the associated access token was used (using the `last_validated` field) or if the refresh token was consumed (used on `/refresh`).
I'll write try to write some tests to check if the behavior is correct.

sandhose:
I fixed the existing tests. The `portdb` script seems to hang though, not sure why?

erikjohnston:
> I fixed the existing tests. The `portdb` script seems to hang though, not sure why?

FTR I tried running locally at it doesn't hang. The logs on CI make it look like it hangs after its completed all the work.

synapse/storage/databases/main/schema/delta/60/01refresh_tokens.sql.postgres
```diff
@@ -0,0 +1,13 @@
+CREATE TABLE refresh_tokens (
+  id BIGINT PRIMARY KEY,
+  user_id TEXT NOT NULL,
+  device_id TEXT,
+  token TEXT NOT NULL,
+  next_token_id BIGINT REFERENCES refresh_tokens (id) ON DELETE CASCADE,
```
richvdh, line 6: this is the reason the port script is getting stuck: it tries to populate the tables in a way that constraints will be satisfied, by populating the least-constrained tables first. The logic is being confused by `refresh_tokens`'s reference to itself.

I've not looked at the rest of the code yet, but one solution might be simply to add `AND tc.table_name != ccu.table_name` to the `get_table_constraints` query in the port script.

synapse/storage/databases/main/registration.py
```diff
@@ -1042,6 +1052,54 @@ async def update_access_token_last_validated(self, token_id: int) -> None:
...
+        def _lookup_refresh_token_txn(txn) -> Optional[RefreshTokenLookupResult]:
+            txn.execute(
+                """
+                SELECT
+                    rt.id token_id,
+                    rt.user_id,
+                    rt.device_id,
+                    rt.next_token_id,
+                    (nrt.next_token_id IS NOT NULL) has_next_refresh_token_been_refreshed,
+                    (at.last_validated IS NOT NULL) has_next_access_token_been_used
```
richvdh, line 1069: I don't think that `at.last_validated` does what you need here (it records the last time that the token was revalidated with user-interactive auth).

The nearest to what you want is probably to join against `user_ips` (by user_id and access_token). `user_ips` is updated in batches so it lags by up to 5 seconds, but hopefully that's acceptable?

richvdh (commented):
looks generally good to me, modulo a few tweaks!

at some point in the near future (not necessarily part of this PR) we could do with a job to remove the expired access tokens from the database.

synapse/storage/databases/main/schema/delta/60/01refresh_tokens.sql.sqlite
```diff
@@ -0,0 +1,31 @@
...
+  id BIGINT PRIMARY KEY,
+  user_id TEXT NOT NULL,
+  device_id TEXT,
+  token TEXT NOT NULL,
+  next_token_id BIGINT REFERENCES refresh_tokens (id) ON DELETE CASCADE,
+  UNIQUE(token)
+);
+
+-- Adding the refresh_token column to access tokens
```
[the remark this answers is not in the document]
sandhose: I knew I had issues with table altering with SQLite at some point and got confused by [this old scheme delta](https://github.com/matrix-org/synapse/blob/e3bc4617fcd5c858ff02cf2d443b898db87ae8a5/synapse/storage/databases/main/schema/delta/58/28drop_last_used_column.sql.sqlite#L33). Fixed in f499d6349

synapse/storage/prepare_database.py
```diff
@@ -35,7 +35,7 @@
 
 # Remember to update this number every time a change is made to database
 # schema files, so the users will be informed on server restarts.
-SCHEMA_VERSION = 59
+SCHEMA_VERSION = 60
```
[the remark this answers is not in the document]
sandhose: I was not aware of that, fixed in f499d6349

synapse/storage/databases/main/schema/delta/60/01refresh_tokens.sql.postgres
sandhose: Will try that, thanks!

sandhose (commented):
Thanks @richvdh for the review. Applied a bunch of your suggestions, but I'm still working on it

synapse/handlers/register.py
```diff
@@ -55,6 +56,16 @@
...
 
+LoginDict = TypedDict(
+    "LoginDict",
+    {
+        "device_id": str,
+        "access_token": str,
+        "valid_until_ms": Optional[int],
+        "refresh_token": Optional[str],
+    },
+)
```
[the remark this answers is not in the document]
sandhose: I did not manage to make that work in a worker setup because it can go through the [`RegisterDeviceReplicationServlet`](https://github.com/matrix-org/synapse/pull/9450/files#diff-a3bbf601c798f027761fde58f9e58bb5a50dab2bd5017597747760f3975b1cc6L77) if it is in a worker setup, and that has to return dict I think?

synapse/handlers/auth.py
```diff
@@ -770,13 +778,128 @@ def _auth_dict_for_flows(
...
+            message = base64.urlsafe_b64decode(parts[1])
+        except binascii.Error:
+            return False
+
+        for key in self.signing_keys:
+            try:
+                key.verify_key.verify(message, signature)
+                return True
+            except nacl.signing.BadSignatureError:
+                pass
```
sandhose, line 868: Please check here if you're OK with my approach here. Tokens are basically shaped like this: `{base64(sig)}.{base64(rand)}` where `rand` is 128 random bytes and `sig` is a Ed25519 (iirc) signature of `rand` signed with the first signing key in the config.
When checking the validity of refresh token, we first check its signature (with old, rotated keys as well) before reaching the DB

synapse/handlers/register.py
```diff
@@ -666,7 +678,8 @@ async def register_device(
         is_guest: bool = False,
         is_appservice_ghost: bool = False,
         auth_provider_id: Optional[str] = None,
-    ) -> Tuple[str, str]:
+        should_issue_refresh_token: bool = False,
+    ) -> Tuple[str, str, Optional[int], Optional[str]]:
```
sandhose, line 682: Is it OK to return a bunch of things in a tuple here, or should I better return some struct/dict?

synapse/rest/client/v1/login.py
```diff
@@ -224,19 +272,23 @@ async def _do_other_login(self, login_submission: JsonDict) -> Dict[str, str]:
...
         self,
         user_id: str,
         login_submission: JsonDict,
-        callback: Optional[Callable[[Dict[str, str]], Awaitable[None]]] = None,
+        callback: Optional[Callable[[LoginResponse], Awaitable[None]]] = None,
         create_non_existent_users: bool = False,
         ratelimit: bool = True,
         auth_provider_id: Optional[str] = None,
-    ) -> Dict[str, str]:
+        should_issue_refresh_token: bool = False,
```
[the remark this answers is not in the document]
sandhose: done

synapse/rest/client/v2_alpha/register.py
```diff
@@ -702,6 +707,10 @@ async def _create_registration_details(
             )
 
             result.update({"access_token": access_token, "device_id": device_id})
+
+            if refresh_token is not None:
+                result["refresh_token"] = refresh_token
```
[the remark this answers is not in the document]
sandhose: done

synapse/storage/databases/main/registration.py
```diff
@@ -1042,6 +1054,92 @@ async def update_access_token_last_validated(self, token_id: int) -> None:
...
+            if old_next_token_id is not None:
+                self.db_pool.simple_delete_one_txn(
+                    txn,
+                    "refresh_tokens",
+                    {"id": old_next_token_id},
+                )
+
+        await self.db_pool.runInteraction(
+            "replace_refresh_token", _replace_refresh_token_txn
+        )
```
sandhose, line 1141: Please double-check this. Is this the right way to do all this in one transaction?

synapse/storage/databases/main/schema/delta/59/12refresh_tokens.sql
```diff
@@ -0,0 +1,34 @@
...
+  UNIQUE(token)
+);
+
+-- Add a reference to the refresh token generated alongside each access token
+ALTER TABLE "access_tokens"
+  ADD COLUMN refresh_token_id BIGINT REFERENCES refresh_tokens (id) ON DELETE CASCADE;
+
+-- Add a flag whether the token was already used or not
+ALTER TABLE "access_tokens"
+  ADD COLUMN used BOOLEAN NOT NULL DEFAULT FALSE;
```
sandhose, line 34: I'm a bit scared of adding a new column on this table with a default. Seems like Postgres 11 handles this properly: https://brandur.org/postgres-default but I have not tested.
The alternative is to make a null-able field instead.

tests/rest/client/v2_alpha/test_auth.py
```diff
@@ -499,3 +499,221 @@ def test_ui_auth_fails_for_incorrect_sso_user(self):
         self.delete_device(
             self.user_tok, self.device_id, 403, body={"auth": {"session": session_id}}
         )
+
+
+class RefreshAuthTests(unittest.HomeserverTestCase):
```
sandhose, line 504: I only did REST API tests there. This does not cover some cases like SSO and app services logins yet

tests/rest/client/v2_alpha/test_auth.py
```diff
@@ -499,3 +499,221 @@ def test_ui_auth_fails_for_incorrect_sso_user(self):
...
+            - and either
+                - the next access token was used
+                - the next refresh token was used
+
+        The chain of tokens goes like this:
+
+            login -|-> first_refresh -> third_refresh (fails)
+                   |-> second_refresh -> fifth_refresh
+                   |-> fourth_refresh (fails)
+        """
```
sandhose, line 643: This test is a bit convoluted. Please check if the comments are enough to understand what it does. Feel free to suggest things here if not

sandhose (commented):
@richvdh this is now ready for another review!

synapse/handlers/auth.py
sandhose: This is not the case anymore, I had not seen #5588 when I did this. Tokens now have the same shape as access tokens. I'm checking their shape + CRC before reaching the DB though.

synapse/rest/client/v1/login.py
```diff
@@ -371,6 +447,41 @@ def _get_auth_flow_dict_for_idp(
...
+                200,
+                {
+                    "access_token": access_token,
+                    "refresh_token": refresh_token,
+                    "expires_in": int(expires_in_ms / 1000),
+                },
+            )
+
+        except KeyError:
+            raise SynapseError(400, "Missing JSON keys.")
```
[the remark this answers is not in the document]
sandhose: Done in https://github.com/matrix-org/synapse/pull/9450/commits/c20f94aaa95b94d6eaae197eabbf4469c97fbfb5

synapse/storage/databases/main/registration.py
sandhose: Done in https://github.com/matrix-org/synapse/pull/9450/commits/262d1ab17405eecba65e045ae75b3f958bb4b599

synapse/storage/databases/main/schema/delta/60/01refresh_tokens.sql.postgres
sandhose: Done in https://github.com/matrix-org/synapse/pull/9450/commits/29806b4f1a18ccb56848dd66183ca825a7f25964

synapse/rest/client/v1/login.py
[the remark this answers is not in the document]
sandhose: Done in https://github.com/matrix-org/synapse/pull/9450/commits/c20f94aaa95b94d6eaae197eabbf4469c97fbfb5

synapse/rest/client/v1/login.py
```diff
@@ -138,6 +158,13 @@ def on_GET(self, request: SynapseRequest):
     async def on_POST(self, request: SynapseRequest):
         login_submission = parse_json_object_from_request(request)
 
+        param = bytes(LoginRestServlet.REFRESH_TOKEN_PARAM, "utf-8")
```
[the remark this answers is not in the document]
sandhose: fixed in https://github.com/matrix-org/synapse/pull/9450/commits/adc6eabfd88f3260a694b4038e8d2255cb151a02

synapse/rest/client/v1/login.py
```diff
@@ -138,6 +158,13 @@ def on_GET(self, request: SynapseRequest):
...
+        should_issue_refresh_token = (
+            request.args is not None
+            and param in request.args
+            and request.args[param][0] == b"true"
+        )
```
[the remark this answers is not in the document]
sandhose: fixed in https://github.com/matrix-org/synapse/pull/9450/commits/adc6eabfd88f3260a694b4038e8d2255cb151a02

sandhose:
I did something wrong while merging develop into my branch (that is not showing in the diff on GitHub), which is most probably why there is a failing test. I'll let you review in this state and if you're OK I'll rebase my branch properly end of this week.

richvdh:
> I'll rebase my branch properly end of this week.

please don't rebase now that it's been through a couple of rounds of review. A merge should be fine.

synapse/handlers/register.py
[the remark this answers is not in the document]
richvdh: fair enough.

synapse/rest/client/v1/login.py
```diff
@@ -253,6 +253,8 @@ async def _do_other_login(
 
         Args:
             login_submission:
+            should_issue_refresh_token: Wheter this login should issue a
```
richvdh, line 256: 'Wheter' has a typo, but generally I prefer to explicitly what it means if it is True:

```suggestion
            should_issue_refresh_token: True if this login should issue a
```

synapse/rest/client/v2_alpha/register.py
```diff
@@ -677,18 +689,25 @@ async def on_POST(self, request):
...
+            should_issue_refresh_token=should_issue_refresh_token,
         )
 
     async def _create_registration_details(
-        self, user_id, params, is_appservice_ghost=False
+        self,
+        user_id,
+        params,
+        is_appservice_ghost=False,
+        should_issue_refresh_token: bool = False,
```
richvdh, line 710: docstring for the new param please. (while you're here, perhaps you could move the types from the docstring into annotations on the params.)

(NB `params` is a `dict`, not an `object` - looks like someone had been writing too much javascript...)

synapse/storage/schema/main/delta/59/14refresh_tokens.sql
```diff
@@ -0,0 +1,34 @@
...
+  token TEXT NOT NULL,
+  -- When consumed, a new refresh token is generated, which is tracked by
+  -- this foreign key
+  next_token_id BIGINT REFERENCES refresh_tokens (id) ON DELETE CASCADE,
+  UNIQUE(token)
+);
+
+-- Add a reference to the refresh token generated alongside each access token
+ALTER TABLE "access_tokens"
+  ADD COLUMN refresh_token_id BIGINT REFERENCES refresh_tokens (id) ON DELETE CASCADE;
```
richvdh, line 30: I've been bitten in the past by `ON DELETE CASCADE` deleting stuff unexpectedly. Should we `ON DELETE RESTRICT` or `ON DELETE SET NULL` instead?

synapse/storage/databases/main/registration.py
```diff
@@ -1082,6 +1084,21 @@ async def update_access_token_last_validated(self, token_id: int) -> None:
             desc="update_access_token_last_validated",
         )
 
+    async def mark_access_token_as_used(self, token_id: int) -> None:
+        """Mark the access token as used, which invalidates the old refresh token.
```
richvdh, line 1088:
```suggestion
        """Mark the access token as used, which invalidates the refresh token used to obtain it.
```

synapse/handlers/auth.py
richvdh: yup the new approach looks good.

synapse/storage/schema/main/delta/59/14refresh_tokens.sql
sandhose: That helps to delete the whole chain of refresh and access tokens when the user logs out. If we only delete the latest refresh token on logout, the N-1 token becomes valid again

synapse/storage/databases/main/registration.py
```diff
@@ -1082,6 +1084,21 @@ async def update_access_token_last_validated(self, token_id: int) -> None:
...
+            token_id: The ID of the access token to update.
+        Raises:
+            StoreError if there was a problem updating this.
+        """
+        await self.db_pool.simple_update_one(
+            "access_tokens",
+            {"id": token_id},
+            {"used": True},
+            desc="mark_access_token_as_used",
+        )
```
richvdh, line 1100: note that `get_user_by_access_token` has a cache on it, so you'll have to invalidate the cache after updating the table.

I previously suggested joining to `user_ips` instead of modifying `access_tokens`: would that not be sufficient?

synapse/storage/databases/main/registration.py
```diff
@@ -1123,11 +1140,34 @@ def _lookup_refresh_token_txn(txn) -> Optional[RefreshTokenLookupResult]:
         )
 
     async def replace_refresh_token(self, token_id: int, next_token_id: int) -> None:
```
richvdh, line 1142: docstring for this please

synapse/storage/databases/main/schema/delta/59/12refresh_tokens.sql
richvdh: yeah, this is problematic in postgres 9.6 and 10. https://www.postgresql.org/docs/10/sql-altertable.html#SQL-ALTERTABLE-NOTES has some more info on it.

Best to make it nullable.

synapse/storage/databases/main/registration.py
richvdh: yup, lgtm.

synapse/handlers/register.py
richvdh: an attr.s would probably be nicer; it's ok though.

synapse/config/registration.py
```diff
@@ -152,6 +156,10 @@ def generate_config_section(self, generate_secrets=False, **kwargs):
         #
         #session_lifetime: 24h
 
+        # MSC2918
+        # TODO: docs
+        #access_token_lifetime: 5m
+
```
richvdh, line 162: we should maybe leave this out of the documentation altogether until MSC2918 is formalised, tbh.

```suggestion
```

synapse/handlers/auth.py
```diff
@@ -772,13 +774,114 @@ def _auth_dict_for_flows(
...
+
+        existing_token = await self.store.lookup_refresh_token(refresh_token)
+        if existing_token is None:
+            raise SynapseError(400, "refresh token does not exist")
+
+        if (
+            existing_token.has_next_access_token_been_used
+            or existing_token.has_next_refresh_token_been_refreshed
+        ):
+            raise SynapseError(400, "refresh token isn't valid anymore")
```
richvdh, line 807: should these be 403s, maybe?

It'd also be nice to return something other than `M_UNKNOWN` as the error code. (Error behaviour should be covered in the MSC.)

synapse/handlers/auth.py
```diff
@@ -772,13 +774,114 @@ def _auth_dict_for_flows(
...
+            device_id=existing_token.device_id,
+            valid_until_ms=valid_until_ms,
+            refresh_token_id=new_refresh_token_id,
+        )
+        await self.store.replace_refresh_token(
+            existing_token.token_id, new_refresh_token_id
+        )
+        return access_token, new_refresh_token
+
+    def verify_refresh_token(self, token: str) -> bool:
```
richvdh, line 826: Probably best to make this a private method, for clarity:

```suggestion
    def _verify_refresh_token(self, token: str) -> bool:
```

synapse/storage/databases/main/registration.py
```diff
@@ -1271,6 +1370,7 @@ async def add_access_token_to_user(
         device_id: Optional[str],
         valid_until_ms: Optional[int],
         puppets_user_id: Optional[str] = None,
+        refresh_token_id: Optional[int] = None,
```
richvdh, line 1373: docstring please

synapse/storage/databases/main/registration.py
```diff
@@ -1297,6 +1397,29 @@ async def add_access_token_to_user(
...
+        )
+
+        return next_id
+
+    async def add_refresh_token_to_user(
+        self,
+        user_id: str,
+        token: str,
+        device_id: Optional[str],
+    ) -> int:
```
richvdh, line 1412: docstring please

synapse/handlers/register.py
```diff
@@ -717,21 +738,40 @@ class and RegisterDeviceReplicationServlet.
...
         else:
+            if should_issue_refresh_token:
+                (
+                    refresh_token,
+                    refresh_token_id,
+                ) = await self._auth_handler.get_refresh_token_for_user_id(
+                    user_id,
+                    device_id=registered_device_id,
+                )
+                valid_until_ms = self.clock.time_msec() + self.access_token_lifetime
```
richvdh, line 759: it seems like we need to define how `session_lifetime` and `access_token_lifetime`  interact. Currently it looks like `session_lifetime` is just ignored if the client requests a refresh token, which isn't really any good.

`session_lifetime`'s main purpose is to force a periodic re-authentication with an external authentication provider. Obviously ultimately the OAuth2 work will provide a better solution here, but we'll need to do something in the meantime.

Can we just refuse to issue a refresh token (falling back to the existing behaviour) if a `session_lifetime` is specified?

synapse/rest/client/v1/login.py
```diff
@@ -273,28 +329,48 @@ async def _complete_login(
...
+    async def _do_token_login(
+        self, login_submission: JsonDict, should_issue_refresh_token: bool = False
+    ) -> LoginResponse:
         """
         Handle the final stage of SSO login.
 
         Args:
-             login_submission: The JSON request body.
+            login_submission: The JSON request body.
+            should_issue_refresh_token: Wheter this login should issue a
```
richvdh, line 372: wheter

tests/rest/client/v2_alpha/test_auth.py
richvdh: looks very clear to me, thanks!

synapse/rest/client/v2_alpha/register.py
sandhose: Done in https://github.com/matrix-org/synapse/pull/9450/commits/6024ed89983c0cf756a354279cf9eb464249acc1

sandhose (commented):
@richvdh Thanks a lot for the review. I think I replied to *all* open conversations. I left them open so you can check all of them & close them yourself.

As said in [#synapse-dev:matrix.org](https://matrix.to/#/!XaqDhxuTIlvldquJaV:matrix.org/$wACotMy9B6G9Ei4H4jIJQ8i578VZC6BMzyX358yrbyU?via=matrix.org&via=vector.modular.im&via=envs.net), I'm very confused by the failing test. It fails consistently, even locally, and only on this branch, even after merging back `develop`.

sandhose:
There was an issue with sqlite returning `'FALSE'` on `BOOLEAN` columns in the `py3-old` environment (and ofc, `bool('FALSE') == True`). I could not reproduce it locally, until I tried that in a docker container with the `ubuntu:bionic` image (same as in CI).

I first tried fixing it by casting to `INTEGER` (https://github.com/matrix-org/synapse/pull/9450/commits/cdfd871bc3136661fe37102e9403544f5f04b773), but the real issue was the default value of the `used` column: `FALSE` was interpreted as `'FALSE'` by sqlite, so the default inserted something that was then interpreted as truthy.

sqlite 3.22.0:
```
sqlite> CREATE TABLE a (b INT, c BOOL NOT NULL DEFAULT FALSE);
sqlite> INSERT INTO a (b) VALUES (1);
sqlite> SELECT * FROM a;
1|FALSE
```

sqlite 3.35.2:
```
sqlite>  CREATE TABLE a (b INT, c BOOL NOT NULL DEFAULT FALSE);
sqlite> INSERT INTO a (b) VALUES (1);
sqlite> SELECT * FROM a;
1|0
```

This was properly fixed in https://github.com/matrix-org/synapse/pull/9450/commits/e07ef9b8d97e4d7500d6045519af4dd83a7f5440. I had to split the Postgres migration and the SQLite migration because Postgres does not like `0` as default for a `BOOLEAN` column.

synapse/storage/databases/main/registration.py
```diff
@@ -53,6 +53,7 @@ class TokenLookupResult:
         valid_until_ms: The timestamp the token expires, if any.
         token_owner: The "owner" of the token. This is either the same as the
             user, or a server admin who is logged in as the user.
+        token_used: True if this token was used at least once in a request.
```
richvdh, line 56: it might be worth adding a comment here to note that `get_user_by_access_token` is cached, so this field might be out of date.

richvdh (approved):
modulo a few tiny niggles, looks great!

richvdh:
> There was an issue with sqlite returning `'FALSE'` on `BOOLEAN` columns in the `py3-old` environment (and ofc, `bool('FALSE') == True`). I could not reproduce it locally, until I tried that in a docker container with the `ubuntu:bionic` image (same as in CI).
> 
> I first tried fixing it by casting to `INTEGER` ([cdfd871](https://github.com/matrix-org/synapse/commit/cdfd871bc3136661fe37102e9403544f5f04b773)), but the real issue was the default value of the `used` column: `FALSE` was interpreted as `'FALSE'` by sqlite, so the default inserted something that was then interpreted as truthy.

Yeah, I think we've had this nonsense before, though I couldn't quite remember the details. Well done for tracking it down.

In the past it looks like we've adopted either of two solutions:
 * the same as you've done - see for example `local_media_repository.safe_from_quarantine` in `delta/58/08_media_safe_from_quarantine.sql.{sqlite,postgres}`.
 * Avoid `SELECT`ing from the column (ie, just use it in WHERE clauses), so that the odd behaviour in old sqlite doesn't happen.

In short, +1 to your solution here. 

I should really write this down somewhere.

richvdh:
> I should really write this down somewhere.

I made #10164 to cover this.

erikjohnston:
I think this is blocked on the MSC progressing further before we want to merge?

synapse/handlers/auth.py
```diff
@@ -793,18 +793,20 @@ async def refresh_token(
         """
 
         # Verify the token signature first before looking up the token
-        if not self.verify_refresh_token(refresh_token):
-            raise SynapseError(400, "invalid refresh token")
+        if not self._verify_refresh_token(refresh_token):
+            raise SynapseError(401, "invalid refresh token", Codes.UNKNOWN_TOKEN)
```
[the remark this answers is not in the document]
sandhose: My understanding is that 401 is more a "you're not authenticated", which could also mean "the credentials you gave me are not valid". 403 would more be a "the credentials you gave me are good, but you can't do that with them"

synapse/storage/databases/main/schema/delta/59/12refresh_tokens.sql
sandhose: I'd rather not have to worry about nullable boolean columns when querying if possible though.

Seems like we already did a lot of `ADD COLUMN ... NOT NULL DEFAULT ...` in the past:
```
synapse/storage/schema/main/delta/53/add_user_type_to_users.sql
19:ALTER TABLE users ADD COLUMN user_type TEXT DEFAULT NULL;

synapse/storage/schema/main/delta/27/forgotten_memberships.sql
26: ALTER TABLE room_memberships ADD COLUMN forgotten INTEGER DEFAULT 0;

synapse/storage/schema/main/delta/48/groups_joinable.sql
22:ALTER TABLE groups ADD COLUMN join_policy TEXT NOT NULL DEFAULT 'invite';

synapse/storage/schema/main/delta/59/14refresh_tokens.sql.postgres
17:ALTER TABLE "access_tokens"
18:  ADD COLUMN used BOOLEAN NOT NULL DEFAULT FALSE;

synapse/storage/schema/main/delta/59/11add_knock_members_to_stats.sql
16:ALTER TABLE room_stats_current ADD COLUMN knocked_members INT NOT NULL DEFAULT '0';
17:ALTER TABLE room_stats_historical ADD COLUMN knocked_members BIGINT NOT NULL DEFAULT '0';

synapse/storage/schema/main/delta/59/14refresh_tokens.sql.sqlite
17:ALTER TABLE "access_tokens"
18:  ADD COLUMN used BOOLEAN NOT NULL DEFAULT 0;

synapse/storage/schema/main/delta/56/hidden_devices.sql
18:ALTER TABLE devices ADD COLUMN hidden BOOLEAN DEFAULT FALSE;

synapse/storage/schema/main/delta/56/redaction_censor.sql
16:ALTER TABLE redactions ADD COLUMN have_censored BOOL NOT NULL DEFAULT false;

synapse/storage/schema/main/delta/58/08_media_safe_from_quarantine.sql.postgres
18:ALTER TABLE local_media_repository ADD COLUMN safe_from_quarantine BOOLEAN NOT NULL DEFAULT FALSE;

synapse/storage/schema/common/delta/58/00background_update_ordering.sql
19:ALTER TABLE background_updates ADD COLUMN ordering INT NOT NULL DEFAULT 0;

synapse/storage/schema/main/delta/58/08_media_safe_from_quarantine.sql.sqlite
18:ALTER TABLE local_media_repository ADD COLUMN safe_from_quarantine BOOLEAN NOT NULL DEFAULT 0;

synapse/storage/schema/main/delta/58/10federation_pos_instance_name.sql
20:ALTER TABLE federation_stream_position ADD COLUMN instance_name TEXT NOT NULL DEFAULT 'master';
```


Looking at the existing columns with defaults, seems like we don't have that many nullable columns with defaults:
```sql
select table_name, column_name, data_type, column_default, is_nullable
from information_schema.columns
where table_schema = 'public' 
  and column_default is not null
order by 1, 2;
```

|table_name                        |column_name         |data_type|column_default                                   |is_nullable|
|----------------------------------|--------------------|---------|-------------------------------------------------|-----------|
|access_tokens                     |used                |boolean  |false                                            |NO         |
|appservice_stream_position        |lock                |character|'X'::bpchar                                      |NO         |
|background_updates                |ordering            |integer  |0                                                |NO         |
|devices                           |hidden              |boolean  |false                                            |YES        |
|e2e_fallback_keys_json            |used                |boolean  |false                                            |NO         |
|e2e_room_keys_versions            |deleted             |smallint |0                                                |NO         |
|event_push_summary_stream_ordering|lock                |character|'X'::bpchar                                      |NO         |
|events                            |depth               |bigint   |0                                                |NO         |
|federation_stream_position        |instance_name       |text     |'master'::text                                   |NO         |
|groups                            |join_policy         |text     |'invite'::text                                   |NO         |
|instance_map                      |instance_id         |integer  |nextval('instance_map_instance_id_seq'::regclass)|NO         |
|local_media_repository            |safe_from_quarantine|boolean  |false                                            |NO         |
|push_rules                        |priority            |integer  |0                                                |NO         |
|received_transactions             |has_been_referenced |smallint |0                                                |YES        |
|redactions                        |have_censored       |boolean  |false                                            |NO         |
|room_memberships                  |forgotten           |integer  |0                                                |YES        |
|schema_version                    |lock                |character|'X'::bpchar                                      |NO         |
|stats_incremental_position        |lock                |character|'X'::bpchar                                      |NO         |
|user_directory_stream_pos         |lock                |character|'X'::bpchar                                      |NO         |
|users                             |admin               |smallint |0                                                |NO         |
|users                             |deactivated         |smallint |0                                                |NO         |
|users                             |is_guest            |smallint |0                                                |NO         |

synapse/storage/databases/main/registration.py
sandhose: Done

synapse/config/registration.py
```diff
@@ -119,10 +119,21 @@ def read_config(self, config, **kwargs):
...
         self.session_lifetime = session_lifetime
 
-        access_token_lifetime = config.get("access_token_lifetime", "5m")
-        access_token_lifetime = self.parse_duration(access_token_lifetime)
+        access_token_lifetime = config.get(
+            "access_token_lifetime", "5m" if session_lifetime is None else None
+        )
+        if access_token_lifetime is not None:
+            access_token_lifetime = self.parse_duration(access_token_lifetime)
         self.access_token_lifetime = access_token_lifetime
```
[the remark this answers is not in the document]
sandhose: Done

synapse/handlers/auth.py
```diff
@@ -772,13 +774,116 @@ def _auth_dict_for_flows(
...
+        existing_token = await self.store.lookup_refresh_token(refresh_token)
+        if existing_token is None:
+            raise SynapseError(401, "refresh token does not exist", Codes.UNKNOWN_TOKEN)
+
+        if (
+            existing_token.has_next_access_token_been_used
+            or existing_token.has_next_refresh_token_been_refreshed
+        ):
+            raise SynapseError(
+                401, "refresh token isn't valid anymore", Codes.UNKNOWN_TOKEN
```
sandhose, line 808: @richvdh then maybe *this* could be a 403 with a `FORBIDDEN` code?

sandhose:
> I think this is blocked on the MSC progressing further before we want to merge?

@erikjohnston I just took the MSC out of "draft" state to have it properly reviewed. Note that this implementation is behind the unstable prefix defined in the MSC and can be completely disabled through config (although enabled by default)

synapse/storage/databases/main/schema/delta/59/12refresh_tokens.sql
richvdh: > Seems like we already did a lot of `ADD COLUMN ... NOT NULL DEFAULT ...` in the past:

Many of those are small tables, so can be rewritten during Synapse restart without major problems. I'm surprised to see a couple of quite large tables in that list, and I wonder if they caused us trouble at the time, but I don't think that changes my position on it here: I'd like us to avoid rewriting the access_tokens table if possible.

> Looking at the existing columns with defaults, seems like we don't have that many nullable columns with defaults:

To be clear, to avoid having to rewrite the table, you need to make it nullable with no default.

synapse/handlers/auth.py
richvdh: yes, probably :)

synapse/config/registration.py
```diff
@@ -119,6 +119,11 @@ def read_config(self, config, **kwargs):
             session_lifetime = self.parse_duration(session_lifetime)
         self.session_lifetime = session_lifetime
 
+        # The `access_token_lifetime` applies for tokens that can be renewed
+        # using a refresh token, as per MSC2918. This behaviour can be disabled
+        # by setting it to `None` (`null` in the YAML config). Since it is
+        # incompatible with the `session_lifetime` mechanism, it is set to
+        # `None` by default if a `session_lifetime` is set.
```
richvdh, line 126:
```suggestion
        # The `access_token_lifetime` applies for tokens that can be renewed
        # using a refresh token, as per MSC2918. If it is `None`, the refresh
        # token mechanism is disabled.
        #
        # Since it is incompatible with the `session_lifetime` mechanism, it is set to
        # `None` by default if a `session_lifetime` is set.
```

synapse/storage/schema/main/delta/59/14refresh_tokens.sql.postgres
```diff
@@ -0,0 +1,18 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+-- Add a flag whether the token was already used or not
+ALTER TABLE "access_tokens"
+  ADD COLUMN used BOOLEAN NOT NULL DEFAULT FALSE;
```
richvdh, line 18: since github is obstinately refusing to make my reply visible, I'll write it again:



>  Seems like we already did a lot of ADD COLUMN ... NOT NULL DEFAULT ... in the past:

Many of those are small tables, so can be rewritten during Synapse restart without major problems. I'm surprised to see a couple of quite large tables in that list, and I wonder if they caused us trouble at the time, but I don't think that changes my position on it here: I'd like us to avoid rewriting the access_tokens table if possible.

>  Looking at the existing columns with defaults, seems like we don't have that many nullable columns with defaults:

To be clear, to avoid having to rewrite the table, you need to make it nullable with no default.

synapse/handlers/auth.py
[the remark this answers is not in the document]
richvdh: seems you're right here!

synapse/storage/schema/main/delta/59/14refresh_tokens.sql.postgres
sandhose: done in https://github.com/matrix-org/synapse/pull/9450/commits/18628fc3adeb28fb820c4e5e1da3b8bece4ce40f

synapse/handlers/auth.py
sandhose: done in https://github.com/matrix-org/synapse/pull/9450/commits/bcc33e25fe7bb8a724b9495ce479a1d24baa8570

richvdh (approved):
lgtm

PR #9458, opened 2021-02-22
Add ResponseCache tests

ShadowJonathan:
Summary of the changes:
- Change `ResponseCache`'s init signature to receive a `synapse.util.Clock` instead of a `HomeServer` (to make it more leaner for tests, and the new signature makes more sense, as it only ever grabbed the clock)
- Change every instantiation location of `ResponseCache` to do `hs.get_clock()` (this is where most of the "noise" of files come from.
  - Annotate some `hs` parameters properly with `HomeServer` (this is technically unrelated, caught along the way, pls notify if i should revert and add in a seperate PR)
    - Make a type annotation in `synapse/federation/federation_server.py` more explicit as mypy starts complaining about it.
 - Add the `tests.util.caches.test_responsecache` tests

### Pull Request Checklist

~~Depends on #9358, will rebase after merge~~

This PR originally included extra behaviour introduced in #9358, however, this behaviour is now under discussion in #9525, this PR doesn't contain that extra behaviour anymore.


synapse/federation/federation_server.py
```diff
@@ -454,6 +455,7 @@ async def _on_state_ids_request_compute(self, room_id, event_id):
     async def _on_context_state_request_compute(
         self, room_id: str, event_id: str
     ) -> Dict[str, list]:
+        pdus = []  # type: Iterable[EventBase]
```
clokep, line 458: I think you can add the type two lines below this (the first time `pdus` is declared) instead of allocating a list that will never be used.

ShadowJonathan: I didn't do this because it'd look wonky, but alright :+1:

ShadowJonathan: Ehhh, i still find it wonky, but i do see a way to make this more pythonic

ShadowJonathan:
In light of #9522, this needs a few more tests for if a function throws an exception while executing, if this is merged before #9522, i'll add these to a new PR (or just to #9522), if #9522 is merged before this, i'll add the extra tests to this.

ShadowJonathan:
cc @richvdh (because of conversation in `#synapse-dev`)

clokep:
@ShadowJonathan Based on #9358 being backed out, what's the status of this? I think it *could* be nice to rebase it and add the tests for the current code. We can than put the fixes necessary on top of that, which might better highlight changes in behavior? Sorry for the churn!

For now I'm going to remove this from the review queue, but please let us know if you think this is waiting for feedback still.

ShadowJonathan:
It's stalled cuz of #9525, but you're right that i can just add some of these tests to the codebase as-is, i'll try to remember to do that.

clokep:
> Some tests in here expect a behaviour that's currently under discussion, and so i wont merge or close this entirely as-is until it's clear what approach is better.

Right, I guess I was thinking it would be awesome to get the tests in that test the current behavior! :)

ShadowJonathan:
Oh, looks like i can just remove `ConditionalDeferredCacheTestCase` and be alright then, i'll leave the reference to the commit that still has this for the future, so i can grab it from there if the conversation resolves that way: a633872085c5b0f1253a97810b6b4666fbdf8226

tests/util/caches/test_responsecache.py
```diff
@@ -0,0 +1,125 @@
...
+from twisted.internet.task import Clock
+
+from synapse.util import Clock as SynapseClock
+from synapse.util.caches.response_cache import ResponseCache
+
+from tests.unittest import TestCase
+
+# few notes about test naming here:
+# 'wait': denotes tests that have an element of "waiting" before its wrapped result becomes available
+# 'expire': denotes tests that test expiry after assured existence
```
clokep, line 24: Does this make sense to be in a docstring instead?

tests/util/caches/test_responsecache.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+class DeferredCacheTestCase(TestCase):
+    def setUp(self):
+        self.reactor = Clock()
+        self.synapse_clock = SynapseClock(self.reactor)
```
clokep, line 30: I think you want to use `tests.server.get_clock` here (and it would be nice to name them the same as `reactor` and `clock` (to match `HomeserverTestCase`).

It might even make sense to create the cache object here and then use it for each test (this simplifies each individual test-case, which makes it easier to see what each test-case is testing). I know the timeouts are currently different on them, but I think the tests could be slightly modified to take that into account.

clokep (commented):
Looks good overall! I left a few (hopefully minor) comments.

tests/util/caches/test_responsecache.py
```diff
@@ -0,0 +1,125 @@
+# Copyright 2021 Vector Creations Ltd
```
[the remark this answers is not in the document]
ShadowJonathan: Because...? Do i need to change it to my name, or to `Element` (with some suffix)?

tests/util/caches/test_responsecache.py
ShadowJonathan: Originally it was meant to apply to the whole file, but since the "whole file" now is one class, i'll just move this and add it to that class, probably make it make more sense too. :+1:

tests/util/caches/test_responsecache.py
ShadowJonathan: I'll use `get_clock`, but i wont move the cache here, as they need to be variable for tests (see the naming and `timeout_ms` on each. :+1: 

One thing I *will* do is make it a separate function (with accepts a name and optionally `ms=`), so that future tests can just use that instead of instantiating it directly every time.

tests/util/caches/test_responsecache.py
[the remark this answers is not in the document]
clokep: It should be your name, I believe.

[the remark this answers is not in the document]
ShadowJonathan: Ah, i'm seeing "The Matrix.org Foundation C.I.C.", i'll change it to that, unless there are objections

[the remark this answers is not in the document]
clokep: Yeah, that's fine as well! 👍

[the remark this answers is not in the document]
clokep: (Specifically New Vector or any other variation of that shouldn't be used.)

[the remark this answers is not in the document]
ShadowJonathan: Pushed with the new copyright, please notify if this is the wrong one.

tests/util/caches/test_responsecache.py
ShadowJonathan: I changed the instantiation to `with_cache` to make it easier on the eyes, please notify if this is wrong.

tests/util/caches/test_responsecache.py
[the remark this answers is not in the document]
ShadowJonathan: Oh whoops, i didnt see these previous comments, thanks! :+1:

clokep (approved):
LGTM! Thanks!

PR #9463, opened 2021-02-22
example systemd config: propagate reloads to units

richvdh:
It should be possible to reload `synapse.target` to have the reload propagate to all the synapse units.

### DIFF
--- a/changelog.d/9463.doc
+++ b/changelog.d/9463.doc
@@ -1,0 +1,1 @@
+Update the example systemd config to propagate reloads to individual units.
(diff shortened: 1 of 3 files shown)


clokep (commented):
I think this looks OK. From reading the docs it seems `PartOf` handles only stopping and restarting of units, but not reloading. (There's also a `Requires`, but that seems to do a bunch of other stuff, but not handle reloading.)

My only question is if this should be `matrix-synapse.target` instead of just `synapse.target`?

richvdh:
> My only question is if this should be matrix-synapse.target instead of just synapse.target?

ugh, yes it should. thanks.

PR #9465, opened 2021-02-22
Fix deleting pushers when using sharded pushers.

erikjohnston:
`remove_pushers_by_access_token` has the following check:

https://github.com/matrix-org/synapse/blob/0a00b7ff14890987f09112a2ae696c61001e6cf1/synapse/push/pusherpool.py#L178-L179

The intent here was that we should ignore attempts to delete pushers on workers not responsible for that user's push. This is incorrect, as deleting pushers must happen on master. However, unless the `pusher_instances` config is set and non-empty `ShardedWorkerHandlingConfig.should_handle(..)`  will always return true (for silly backwards compat reasons, which is fixed in a separate PR), and so the method ends up doing the right thing in that case.

Since our integration tests for workers doesn't use `pusher_instances` all the tests passed, even though they broke if that option was set. 

We fix this by a) removing the condition and b) making `remove_pusher` hit master over replication when called on workers (I don't think anything on workers call it, but it seemed easier to just add support for it). We also replace the `REMOVE_PUSHER` replication command with a HTTP API (to make it reliable).

I'm planning on clearing out pushers that should have been deleted in a separate PR.

### DIFF
--- a/changelog.d/9465.bugfix
+++ b/changelog.d/9465.bugfix
@@ -1,0 +1,1 @@
+Fix deleting pushers when using sharded pushers.
--- a/docs/tcp_replication.md
+++ b/docs/tcp_replication.md
@@ -220,10 +220,6 @@ Asks the server for the current position of all streams.
 
    Acknowledge receipt of some federation data
 
-#### REMOVE_PUSHER (C)
-
-   Inform the server a pusher should be removed
-
 ### REMOTE_SERVER_UP (S, C)
 
    Inform other processes that a remote server may have come back online.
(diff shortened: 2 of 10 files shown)


clokep:
> I'm planning on clearing out pushers that should have been deleted in a separate PR.

Will this be a background update / schema delta or something manual?

erikjohnston:
> > I'm planning on clearing out pushers that should have been deleted in a separate PR.
> 
> Will this be a background update / schema delta or something manual?

I'm assuming something automated, but haven't really gotten that far yet

clokep (approved):
I think this looks fine! Do we need to update any tests for this (or will we be updating some sytests for this)?

erikjohnston:
> I think this looks fine! Do we need to update any tests for this (or will we be updating some sytests for this)?

https://github.com/matrix-org/synapse/pull/9466 picks it up, by removing the different code paths between sharded and non-sharded push set ups. Though maybe we should start running our worker mode sytests with sharded pushers?

Issue #8015, opened 2020-08-01
federation sender doesn't work when invoked as synapse.app.generic_worker


maranda:
### Description

As in the title, I followed the documentation closely to setup the worker, but when a message is propagated through Federation servers will answer with 403s

If I leave federation sending to the master instance there're no issues.

### Steps to reproduce

- Send message in a room
- Message appears locally but is not replicated over the network
- Remote servers answer 403s to federation transactions

Messages appear locally but not remotely and in logs the following warnings appear:
`
2020-08-01 11:29:46,696 - synapse.http.matrixfederationclient - 536 - WARNING - federation_transaction_transmission_loop-4108 - {PUT-O-3160} [***] Request failed: PUT matrix://***/_matrix/federation/v1/send/1596280923582: HttpResponseException("403: b'Forbidden'")
`
### Version information

- **Homeserver**: aria-net.org

If not matrix.org:

- **Version**: 1.18

- **Install method**: Ubuntu packages

- **Platform**:
Hyper-V machine with 16GB of RAM and eight virtual processors
Synapse running with 1+3 workers (2 general, 1 federation sender) with Postgresql and Redis managed via systemd
OS: Ubuntu server 20.04

maranda:
I found the issue.

Basically https://github.com/matrix-org/synapse/blob/develop/docs/systemd-with-workers/system/matrix-synapse-worker%40.service will only startup the generic worker I didn't really look inside the script so I didn't notice it.

I think that should be properly referenced in the documentation, that the script should be modified to use other worker typed e.g. synapse.app.federation_sender.

erikjohnston:
Thanks for the bug report! Part of me feels like we just make everything use the same app to avoid this problem, but until then yeah let's just fix the documentation :+1:

richvdh:
I'd so much rather fix the app than the docs...

richvdh:
I'm confused by this. Why does the `service` file changing? Why doesn't it work when it starts `-m synapse.app.generic_worker` ?

maranda:
@richvdh it's not that the worker didn't start it's that the generic worker didn't work at all for the federation sender role.

richvdh:
yes; I'm asking why that is the case. AFAICT the code that gets run is identical.

richvdh:
(that's mostly a question for the dev team!)

clokep:
There's still some special-cased code that needs to get cleaned-up:

[...]

PR #9466, opened 2021-02-22
Clean up `ShardedWorkerHandlingConfig`

erikjohnston:
See commit messages, but basically the idea is to make `ShardedWorkerHandlingConfig` behave in a saner way, rather than having special cases for situations where there are only one or no instances defined. This is done by consolidating all the logic to handle the different ways of configuring pusher and federation senders into one place.

The special casing masked problems such as those described in https://github.com/matrix-org/synapse/pull/9465


synapse/config/_base.py
```diff
@@ -876,4 +876,20 @@ def get_instance(self, key: str) -> str:
...
+@attr.s
+class RoutableShardedWorkerHandlingConfig(ShardedWorkerHandlingConfig):
+    """A version of `ShardedWorkerHandlingConfig` that is used for config
+    options where all instances know which instances are responsible for the
+    sharded work.
+    """
+
+    def __attrs_post_init__(self):
+        # We require that `self.instances` is non-empty.
+        assert self.instances
```
clokep, line 888: Should these raise `ConfigError` instead? (Or at least provide a message for how the user can fix this?)

synapse/config/_base.py
```diff
@@ -876,4 +876,20 @@ def get_instance(self, key: str) -> str:
...
+
+    def get_instance(self, key: str) -> str:
+        """Get the instance responsible for handling the given key."""
+        return self._get_instance(key)
```
clokep, line 892: Where is this defined? (Is this a magic attrs thing?)

synapse/config/_base.pyi
```diff
@@ -149,4 +149,6 @@ class ShardedWorkerHandlingConfig:
     instances: List[str]
     def __init__(self, instances: List[str]) -> None: ...
     def should_handle(self, instance_name: str, key: str) -> bool: ...
+
+class RoutableShardedWorkerHandlingConfig(ShardedWorkerHandlingConfig):
     def get_instance(self, key: str) -> str: ...
```
clokep, line 154: From the above both `ShardedWorkerHandlingConfig` and `RoutableShardedWorkerHandlingConfig` seem to define `get_instance`.

synapse/config/_base.py
erikjohnston: `_get_instance` is defined in `ShardedWorkerHandlingConfig`

synapse/config/_base.pyi
erikjohnston: `ShardedWorkerHandlingConfig` has a private `_get_instance`, rather than `get_instance`

synapse/config/_base.py
erikjohnston: We should be checking before this, as we can't give a particularly helpful message. I'll make it raise a slightly more useful exception though.

Relatedly, looks like we don't actually check that there is at least one event stream writer, I'll fix that too.

synapse/config/_base.py
```diff
@@ -845,21 +845,21 @@ class ShardedWorkerHandlingConfig:
     def should_handle(self, instance_name: str, key: str) -> bool:
         """Whether this instance is responsible for handling the given key."""
         # If multiple instances are not defined we always return true
-        if not self.instances or len(self.instances) == 1:
-            return True
+        if not self.instances:
+            return False
```
clokep, line 849: This comment seems out of date now.

clokep:
Will this fix #8015?

erikjohnston:
> Will this fix #8015?

Oooh, yes , I think so!

synapse/config/workers.py
```diff
@@ -17,9 +17,28 @@
...
+from ._base import (
+    Config,
+    ConfigError,
+    RoutableShardedWorkerHandlingConfig,
+    ShardedWorkerHandlingConfig,
+)
 from .server import ListenerConfig, parse_listener_def
 
+_FEDERATION_SENDER_WITH_SEND_FEDERATION_ENABLED_ERROR = """
+The send_federation must be disabled in the main synapse process
```
clokep, line 29: I feel like these are both missing a word -- maybe "The send_federation *setting*..."?

synapse/config/workers.py
```diff
@@ -118,12 +138,41 @@ def read_config(self, config, **kwargs):
...
+                federation_sender_instances = ["master"]
+
+            if self.worker_app == "synapse.app.federation_sender":
+                if send_federation:
+                    # If we're running federation senders, and not using
+                    # `federation_sender_instances`, then we should have
+                    # explicitly set `send_federation` to false.
+                    raise ConfigError(
+                        _FEDERATION_SENDER_WITH_SEND_FEDERATION_ENABLED_ERROR
+                    )
```
clokep, line 171: I think we've been trying to provide the path as we write new errors.

synapse/config/workers.py
```diff
@@ -118,12 +138,41 @@ def read_config(self, config, **kwargs):
...
+        #      `federation_sender_instances`.
+        #
+
+        send_federation = config.get("send_federation", True)
+
+        federation_sender_instances = config.get("federation_sender_instances")
+        if federation_sender_instances is None:
+            # Default to an empty list, which means "another, unknown, worker is
+            # responsible for it".
+            federation_sender_instances = []
```
clokep, line 157: This happens if you're using `send_federation`, but this config is being parsed on a non-federation sender worker. Is that correct?

synapse/config/workers.py
```diff
@@ -164,7 +213,37 @@ def read_config(self, config, **kwargs):
                 "Must only specify one instance to handle `receipts` messages."
             )
 
-        self.events_shard_config = ShardedWorkerHandlingConfig(self.writers.events)
+        if len(self.writers.events) == 0:
+            raise ConfigError("Must specify at least one instance to handle `events`.")
```
clokep, line 217: I wonder if it would help us remember to add these if we used `validators` on the `WriterLocations` attributes?

synapse/config/workers.py
```diff
@@ -118,12 +138,41 @@ def read_config(self, config, **kwargs):
                 )
             )
 
-        # Whether to send federation traffic out in this process. This only
-        # applies to some federation traffic, and so shouldn't be used to
-        # "disable" federation
-        self.send_federation = config.get("send_federation", True)
+        # Handle federation sender configuration.
```
clokep, line 141: It looks like the logic for sharding federation sender and push is identical, can we abstract this somehow?

I think something like `self.send_federation = self._handle_shared_config("send_federation", "federation_sender_instances", "synapse.app.federation_sender")` might do it?

tests/replication/test_pusher_shard.py
```diff
@@ -95,7 +95,7 @@ def test_send_push_single_worker(self):
 
         self.make_worker_hs(
             "synapse.app.pusher",
-            {"start_pushers": True},
+            {"start_pushers": False},
```
clokep, line 98: I'm a bit confused why we need to provide this? Does the test config default to starting push?

erikjohnston: `start_pushers` should be false in the config file, then previously it got set to true during `generic_worker.start` (which doesn't get run from here) and now gets set to true in the worker config class (which does get run)

synapse/config/workers.py
erikjohnston: Yeah, though I'm not sure how easy it is to make `ConfigError` out of validation failures?

synapse/config/workers.py
erikjohnston: Yes, exactly.

synapse/config/workers.py
erikjohnston: What do you mean? What path?

clokep: `ConfigError` takes an optional `path`:

> Where appropriate, an indication of where in the configuration the problem lies.

I don't know if it should always be provided or not.

synapse/config/workers.py
clokep: There's a few places that we raise a `ConfigError` from another exception, but not sure if it is worth refactoring that here. 🤷

synapse/config/workers.py
erikjohnston: I had a quick look at this and it seems like its just used to prefix the comment with `Error in configuration at 'path.value':`, which I don't think is useful in this case.

synapse/config/workers.py
erikjohnston: I'm slightly in two minds about this. On the one had, yes, on the other we should never have more than these two, and so I think it might actually be clearer just writing it out rather than factoring it out and passing a bunch of stuff in etc :shrug:

clokep: If there isn't additional workers that we'll need to apply this to than that sounds OK.

erikjohnston: Yeah, this only for working around the old style config options, new stuff shouldn't have that problem

clokep (approved):

Issue #7938, opened 2020-07-23
check postgres sequences for validity on startup 


richvdh:
in the last two weeks, I've seen two people who have migrated their databases to a new postgres instance and forgotten to copy the sequence generators. At best, this causes confusing errors; at worst, it could lead to corruption. 

To save people from shooting themselves in the foot, we could check at startup that there are no rows in the table with IDs above the next number in the sequence, and refuse to start if so.

richvdh:
in the meantime, to manually check that your database dump has been correctly imported to your new database:

```sql
SELECT last_value FROM state_group_id_seq;
```

... that should return a value significantly larger than 1.

erikjohnston:
#8402 fixes this *accept* for the guest user ID generation, as per: https://github.com/matrix-org/synapse/pull/8402/files#diff-50fc898a46f43529763a3c63716ba0b4R44-R49 .

PR #9470, opened 2021-02-23
Refactor to ensure we call check_consistency

erikjohnston:
The idea here is to stop people forgetting to call `check_consistency`. Folks can still just pass in `None` to the new args in `build_sequence_generator`, but hopefully they won't.

c.f.  #9382

### DIFF
--- a/synapse/storage/databases/main/registration.py
+++ b/synapse/storage/databases/main/registration.py
@@ -23,7 +23,7 @@ import attr
 from synapse.api.constants import UserTypes
 from synapse.api.errors import Codes, StoreError, SynapseError, ThreepidValidationError
 from synapse.metrics.background_process_metrics import wrap_as_background_process
-from synapse.storage.database import DatabasePool
+from synapse.storage.database import DatabasePool, LoggingDatabaseConnection
 from synapse.storage.databases.main.cache import CacheInvalidationWorkerStore
 from synapse.storage.databases.main.stats import StatsStore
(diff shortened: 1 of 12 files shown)


clokep:
Will probably want to merge the changes from #9449 into this?

clokep:
I think this is related to #7938?

synapse/storage/databases/main/registration.py
```diff
@@ -70,7 +70,12 @@ def _default_token_owner(self):
 
 
 class RegistrationWorkerStore(CacheInvalidationWorkerStore):
-    def __init__(self, database: DatabasePool, db_conn: Connection, hs: "HomeServer"):
+    def __init__(
+        self,
+        database: DatabasePool,
+        db_conn: LoggingDatabaseConnection,
```
clokep, line 76: I remember you and @richvdh having a conversation about switching these from `Connection` to `LoggingDatabaseConnection`, but do not remember the OK. Are we OK making this change?

clokep (approved):
Looks like an improvement! Just one question (and I did wonder if `check_consistency` could be made private, but don't have a strong opinion on it).

synapse/storage/databases/main/registration.py
erikjohnston: Yeah, I think we are

richvdh: he convinced me I think

Issue #9471, opened 2021-02-23
synapse blindly trusts X-Forwarded-For if x_forwarded option is enabled


richvdh:
Synapse does not check that the chain in `X-Forwarded-For` is trusted, and so an attacker can spoof their IP address if the reverse proxy does not sanitize `X-Forwarded-For`.  Ideally, we should be able to pass a set of trusted IP addresses, and synapse should only trust `X-Forwarded-For` if: 1) the request comes from a trusted IP address, and 2) every IP address in X-Forwarded-For, other than the first one, is trusted.

This can be mitigated by ensuring that the `X-Forwarded-For` header is sanitized before it hits synapse.  For example, the public-facing reverse-proxy should remove any `X-Forwarded-For` header that it receives.

The IP address seems to be used for:

- checking that AS requests come from trusted IP addresses
- rate limiting registration requests
- UI auth (maybe?)
- request logging
- last-seen IP address for devices

richvdh:
vaguely related: matrix-org/sydent#299

richvdh:
I've just realised that our [reverse proxy docs](https://github.com/matrix-org/synapse/blob/master/docs/reverse_proxy.md) say nothing about sanitizing the X-Forwarded-For header, so this might be a bit more of an urgent security problem than I thought.

richvdh:
my suggested approach for fixing this would be to allow a `trusted_proxies` setting in the `listener` configuration, which is a list of IP addresses to be compared against the client IP and `X-Forwarded-For` addresses. `x_forwarded` can be deprecated.

dklimpel:
It is also related to https://github.com/matrix-org/matrix-content-scanner/pull/36

PR #9472, opened 2021-02-23
Add support for X-Forwarded-Proto

richvdh:
rewrite XForwardedForRequest to set `isSecure()` based on `X-Forwarded-Proto`. Also implement `getClientAddress()` while we're here.

### DIFF
--- a/docs/reverse_proxy.md
+++ b/docs/reverse_proxy.md
@@ -9,23 +9,23 @@ of doing so is that it means that you can expose the default https port
 (443) to Matrix clients without needing to run Synapse with root
 privileges.
 
-**NOTE**: Your reverse proxy must not `canonicalise` or `normalise`
-the requested URI in any way (for example, by decoding `%xx` escapes).
-Beware that Apache *will* canonicalise URIs unless you specify
-`nocanon`.
+You should configure your reverse proxy to forward requests to `/_matrix` or
+`/_synapse/client` to Synapse, and have it set the `X-Forwarded-For` and
+`X-Forwarded-Proto` request headers.
 
-When setting up a reverse proxy, remember that Matrix clients and other
-Matrix servers do not necessarily need to connect to your server via the
-same server name or port. Indeed, clients will use port 443 by default,
-whereas servers default to port 8448. Where these are different, we
-refer to the 'client port' and the 'federation port'. See [the Matrix
+You should remember that Matrix clients and other Matrix servers do not
+necessarily need to connect to your server via the same server name or
+port. Indeed, clients will use port 443 by default, whereas servers default to
+port 8448. Where these are different, we refer to the 'client port' and the
+'federation port'. See [the Matrix
 specification](https://matrix.org/docs/spec/server_server/latest#resolving-server-names)
 for more details of the algorithm used for federation connections, and
 [delegate.md](<delegate.md>) for instructions on setting up delegation.
 
-Endpoints that are part of the standardised Matrix specification are
-located under `/_matrix`, whereas endpoints specific to Synapse are
-located under `/_synapse/client`.
+**NOTE**: Your reverse proxy must not `canonicalise` or `normalise`
+the requested URI in any way (for example, by decoding `%xx` escapes).
+Beware that Apache *will* canonicalise URIs unless you specify
+`nocanon`.
 
 Let's assume that we expect clients to connect to our server at
 `https://matrix.example.com`, and other servers to connect at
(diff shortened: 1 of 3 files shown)


docs/reverse_proxy.md
```diff
@@ -102,6 +103,7 @@ example.com:8448 {
     SSLEngine on
     ServerName matrix.example.com;
 
+    RequestHeader set "X-Forwarded-Proto" expr=%{REQUEST_SCHEME}
```
richvdh, line 106: Apache and caddy both set X-Forwarded-For by default, though in a way that https://github.com/matrix-org/synapse/issues/9471 is problematic.

docs/reverse_proxy.md
```diff
@@ -134,6 +137,8 @@ example.com:8448 {
 ```
 frontend https
   bind :::443 v4v6 ssl crt /etc/ssl/haproxy/ strict-sni alpn h2,http/1.1
+  http-request add-header X-Forwarded-Proto https if { ssl_fc }
```
clokep, line 140: Should this also be `set-header`?

From the [haproxy docs](https://www.haproxy.com/documentation/aloha/latest/traffic-management/lb-layer7/http-rewrite/):

> `set-header` removes first any header in existing field that matches `<name>`.

richvdh: yes.

I've also added an equivalent for `http`, which is strictly redundant if people only configure an `ssl` bind, but I figure that it might be important if it gets cut-and-pasted.

synapse/http/site.py
```diff
@@ -333,26 +337,77 @@ def _should_log_request(self) -> bool:
...
+    def getClientAddress(self) -> IAddress:
+        """
+        Return the address of the client who submitted this request.
+        """
+        if self._forwarded_for is not None:
+            return self._forwarded_for
+        return super().getClientAddress()
+
+
+@implementer(IAddress)
```
clokep, line 407: I've always found this interface bizarre that it has no fields...

synapse/http/site.py
```diff
@@ -333,26 +337,77 @@ def _should_log_request(self) -> bool:
...
 
-    def getClientIP(self):
+    # the client IP and ssl flag, as extracted from the headers.
+    _forwarded_for = None  # type: Optional[_XForwardedForAddress]
+    _forwarded_https = False  # type: bool
+
+    def requestReceived(self, command, path, version):
+        # this method is called by the Channel once the full request has been
+        # received, to dispatch the request to a resource.
+        # We can use it to set the IP address and port according to the
```
clokep, line 353: I think this should be "IP address and protocol"? (We don't seem to do anything with the port.)

synapse/http/site.py
```diff
@@ -333,26 +337,77 @@ def _should_log_request(self) -> bool:
...
+                "forwarded request lacks an x-forwarded-proto header: assuming https"
+            )
+            self._forwarded_https = True
+
+    def isSecure(self):
+        if self._forwarded_https:
+            return True
+        return super().isSecure()
+
+    def getClientIP(self) -> str:
```
clokep, line 388: `getClientIP` will now potentially return private IPs (of the reverse proxy) instead of `-` if the header isn't set.

It seems that this gets used to [fill the `user_ips` table](https://github.com/matrix-org/synapse/blob/0a00b7ff14890987f09112a2ae696c61001e6cf1/synapse/api/auth.py#L236-L242), which can be [retrieved via the whois call](https://github.com/matrix-org/synapse/blob/0a00b7ff14890987f09112a2ae696c61001e6cf1/synapse/handlers/admin.py#L40-L58), which is [a public API](https://matrix.org/docs/spec/client_server/r0.6.1#get-matrix-client-r0-admin-whois-userid).

This *is* a misconfiguration, but is a change in behavior so thought it was worth calling out.

synapse/http/site.py
richvdh: indeed.

synapse/http/site.py
richvdh: yeah, that behaviour always seemed a bit magical to me. Why `-`?

Note that `/whois` is not *strictly* "public": it can only be called by admin users and the user themselves.

clokep: My point was more of that a private IP could potentially be exposed to users who are not system administrators. (This is what I meant by public.)

richvdh: right, yes.

Well, I feel like it's a fairly basic configuration error. Don't do that.

clokep (approved):
LGTM. I wonder if we should call something out in upgrade.rst?

richvdh:
>  I wonder if we should call something out in upgrade.rst?

I don't think there's anything hugely backwards-incompatible here. The warning is harmless and clear enough imho.

erikjohnston:
> > I wonder if we should call something out in upgrade.rst?
> 
> I don't think there's anything hugely backwards-incompatible here. The warning is harmless and clear enough imho.

I feel like if its important enough for us to log at WARNING (for every request) we should probably call it out? If nothing else almost doubling our log size isn't ideal

richvdh:
> I feel like if its important enough for us to log at WARNING (for every request) we should probably call it out? 

well it's called out in the newsfile...

but fair enough I'll stick something in UPGRADE.rst too.

PR #9473, opened 2021-02-23
Handle image transparency when thumbnailing.

clokep:
Fixes https://sentry.matrix.org/sentry/synapse-matrixorg/issues/104264/

This fixes an error: `TypeError: cannot unpack non-iterable int object` when attempting to thumbnail an image that has transparency set.

This is based ons ome code near https://github.com/python-pillow/Pillow/blob/058b8d3d12de7f18fe199b3cd706e451986eaceb/src/PIL/PngImagePlugin.py#L1279-L1301


clokep:
This is failing due to a change in output starting at pillow 6.0.0. 😢

clokep:
> This is failing due to a change in output starting at pillow 6.0.0. 😢

I figured that for this image it is enough to just check that the code ran properly and not assert the output. Not ideal, but should be OK.

erikjohnston (approved):

PR #9474, opened 2021-02-23
Add back the deprecated SAML endpoint.

### LINKED ISSUE #9452: synapse 1.28rc1 breaks compat with SAML portal
one of our customers has a saml portal which expects synapse to respond to saml callbacks on `/_matrix/saml2/authn_response`. Since #9434, it doesn't.

clokep:
Partially reverts #9434, fixes #9452.

This can be removed in a future version.


richvdh (approved):
lgtm

PR #9479, opened 2021-02-24
Add SQL delta for deleting stale pushers

erikjohnston:
The select for this took ~8s on matrix.org. I think we could turn this into a background job, but I'm not sure it's worth it.

This a clean up after the fix in https://github.com/matrix-org/synapse/pull/9465

### DIFF
--- a/changelog.d/9479.bugfix
+++ b/changelog.d/9479.bugfix
@@ -1,0 +1,1 @@
+Fix deleting pushers when using sharded pushers.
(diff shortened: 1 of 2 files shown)


clokep (approved):
One-time 8s doesn't sound so bad on matrix.org. I think this is OK. Probably want to ensure that #9285 is landed in the same cycle?

erikjohnston:
> One-time 8s doesn't sound so bad on matrix.org. I think this is OK. Probably want to ensure that #9285 is landed in the same cycle?

I don't think it matters, #9285 is probably going to need its own cleanup task.

PR #9484, opened 2021-02-24
Add option for timed /sync cache

### LINKED ISSUE #3880: cache sync results for a while after the request completes
Both @lampholder and @Ralith have reported similar sounding problems where initial /sync takes ~5x longer to complete from Riot than it should.

@lampholder mentioned that he tried calling initial /sync directly from curl against his HS (bypassing nginx) without a clientside timeout and it took 3 mins, but when called by Riot (which times out reqs every 90s or so to avoid network glitches wedging a request) it takes ~15 mins.  Meanwhile, @Ralith (on an underpowered server with bad IO) takes 15 mins(!) to generate the sync when called directly, but >1h when called from Riot :/

I haven't seen logs from either, but my theory is that if the client times out the /sync request **after** the server has generated the response, but **before** it has sent it to the client, then the next time the client retries, the /sync will be calculated from scratch, and the failure mode may continue.

It looks like the current behaviour on develop caches sync requests via response_cache which blocks additional requests for the same sync parameters until any ongoing one completes, and then they all return with the same result: https://github.com/matrix-org/synapse/blob/b1580f50fea05ce02dc4663adb23704b33e30bf3/synapse/handlers/sync.py#L219-L225.

However, on the matrix-org-hotfixes branch @richvdh changed it to cache sync responses for an additional 2 minutes, so that any retrying clients will get the cached copy rather than starting from scratch: https://github.com/matrix-org/synapse/commit/0ca2857baab2d241a740a6ba5e8ed36d17756338.  This is similar to the old /initialSync snapshot_cache from https://github.com/matrix-org/synapse/pull/457 which never seems to have made it over to /sync.

I'm not seeing any additional /sync response cache?

If my theory is correct, it looks like this hotfix needs to be ported back into develop, and sytest fixed so that it doesn't fail if initial /sync returns stale data (perhaps we could add a null filter on or something as a cachebuster).

Another explanation for this could be if Twisted is correctly cancelling the ongoing /sync when the client goes away, thus throwing away the cached response.  The same fix would mitigate that, however.

ShadowJonathan:
This adds the `experimental_features.sync_cache_timeout_ms` config value. Thus this fixes #3880.

Stabilisation for general recommended use is left for a future PR or decision, after this has been tested enough (in my opinion).

### Pull Request Checklist

### DIFF
--- a/changelog.d/9484.bugfix
+++ b/changelog.d/9484.bugfix
@@ -1,0 +1,1 @@
+Add experimental config feature for timed `/sync` cache, fixes #3880.
--- a/synapse/config/experimental.py
+++ b/synapse/config/experimental.py
@@ -29,4 +29,6 @@ class ExperimentalConfig(Config):
         self.msc2858_enabled = experimental.get("msc2858_enabled", False)  # type: bool
 
         # Timeout value (in milliseconds) on sync Response Cache
-        self.sync_cache_timeout_ms = experimental.get("sync_cache_timeout_ms", 0)  # type: int
+        self.sync_cache_timeout_ms = experimental.get(
+            "sync_cache_timeout_ms", 0
+        )  # type: int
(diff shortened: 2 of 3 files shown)


ShadowJonathan:
Folded into #9739

PR #9491, opened 2021-02-24
Add a Synapse Module for configuring presence update routing

anoadragon453:
Based on https://github.com/matrix-org/synapse/pull/9650

At the moment, if you'd like to share presence between local or remote users, those users must be sharing a room together. This isn't always the most convenient or useful situation though.

This PR adds a module to Synapse that will allow deployments to set up extra logic on where presence updates should be routed. The module must implement two methods, `get_users_for_states` and `get_interested_users`. These methods are given presence updates or user IDs and must return information that Synapse will use to grant passing presence updates around.

A method is additionally added to `ModuleApi` which allows triggering a set of users to receive the current, online presence information for all users they are considered interested in. This is the equivalent of that user receiving presence information during an initial sync. 

The goal of this module is to be fairly generic and useful for a variety of applications, with hard requirements being:

* Sending state for a specific set or all known users to a defined set of local and remote users.
* The ability to trigger an initial sync for specific users, so they receive all current state.

This is quite a chunky PR, though hopefully the commits should have enough explanation in each of them to get you by.

### DIFF
--- a/docs/presence_router_module.md
+++ b/docs/presence_router_module.md
@@ -4,6 +4,9 @@ Synapse supports configuring a module that can specify additional users
 (local or remote) to should receive certain presence updates from local
 users.
 
+Note that routing presence via Application Service transactions is not
+currently supported.
+
 The presence routing module is implemented as a Python class, which will be imported by
 the running Synapse.
 
--- a/docs/presence_router_module.md
+++ b/docs/presence_router_module.md
@@ -53,10 +53,10 @@ async def get_users_for_states(
 ) -> Dict[str, Set[UserPresenceState]]:
 ```
 
-**Required.** An asynchronous class method that is passed an iterable of user presence
+**Required.** An asynchronous method that is passed an iterable of user presence
 state. This method can determine whether a given presence update should be sent to certain
 users. It does this by returning a dictionary with keys representing local or remote
-Matrix User IDs, and values being a python Set
+Matrix User IDs, and values being a python set
 of `synapse.handlers.presence.UserPresenceState` instances.
 
 Synapse will then attempt to send the specified presence updates to each user when
@@ -68,10 +68,10 @@ possible.
 async def get_interested_users(self, user_id: str) -> Union[Set[str], str]
 ```
 
-**Required.** An asynchronous class method that is passed a single Matrix User ID. This
+**Required.** An asynchronous method that is passed a single Matrix User ID. This
 method is expected to return the users that the passed in user may be interested in the
-presence of. This may be local or remote users. It does so by returning a python Set of
-Matrix User IDs, or the object
+presence of, in addition to users they share a room with. This may be local or remote
+users. It does so by returning a python set of Matrix User IDs, or the object
 `synapse.events.presence_router.PresenceRouter.ALL_USERS` to indicate that the passed
 user should receive presence information for *all* known users.
 
--- a/docs/presence_router_module.md
+++ b/docs/presence_router_module.md
@@ -4,11 +4,11 @@ Synapse supports configuring a module that can specify additional users
 (local or remote) to should receive certain presence updates from local
 users.
 
-Note that routing presence via an Application Service transactions is not
+Note that routing presence via Application Service transactions is not
 currently supported.
 
-The presence routing module is implemented as a Python class, which will be imported by
-the running Synapse.
+The presence routing module is implemented as a Python class, which will
+be imported by the running Synapse.
 
 ## Python Presence Router Class
 
@@ -40,6 +40,7 @@ required.
 ```python
 def parse_config(config_dict: dict) -> Any
 ```
+
 **Required.** A static method that is passed a dictionary of config options, and
   should return a validated config object. This method is described further in
   [Configuration](#configuration).
@@ -70,8 +71,11 @@ async def get_interested_users(self, user_id: str) -> Union[Set[str], str]
 
 **Required.** An asynchronous method that is passed a single Matrix User ID. This
 method is expected to return the users that the passed in user may be interested in the
-presence of, in addition to users they share a room with. This may be local or remote
-users. It does so by returning a python set of Matrix User IDs, or the object
+presence of. Returned users may be local or remote. The presence routed as a result of
+what this method returns is sent in addition to the updates already sent between users
+that share a room together. Presence updates are deduplicated.
+
+This method should return a python set of Matrix User IDs, or the object
 `synapse.events.presence_router.PresenceRouter.ALL_USERS` to indicate that the passed
 user should receive presence information for *all* known users.
 
--- a/docs/presence_router_module.md
+++ b/docs/presence_router_module.md
@@ -27,7 +27,8 @@ async def ModuleApi.send_local_online_presence_to(users: Iterable[str]) -> None
 
 which can be given a list of local or remote MXIDs to broadcast known, online user
 presence to (for those users that the receiving user is considered interested in). 
-It does not include state for users who are currently offline.
+It does not include state for users who are currently offline, and it can only be
+called on workers that support sending federation.
 
 ### Module structure
 
--- a/docs/presence_router_module.md
+++ b/docs/presence_router_module.md
@@ -57,7 +57,7 @@ async def get_users_for_states(
 state. This method can determine whether a given presence update should be sent to certain
 users. It does this by returning a dictionary with keys representing local or remote
 Matrix User IDs, and values being a python Set
-of `synapse.handlers.presence. UserPresenceState` instances.
+of `synapse.handlers.presence.UserPresenceState` instances.
 
 Synapse will then attempt to send the specified presence updates to each user when
 possible.
@@ -71,8 +71,9 @@ async def get_interested_users(self, user_id: str) -> Union[Set[str], str]
 **Required.** An asynchronous class method that is passed a single Matrix User ID. This
 method is expected to return the users that the passed in user may be interested in the
 presence of. This may be local or remote users. It does so by returning a python Set of
-Matrix User IDs, or the string `"ALL"` to mean that the passed user should receive
-presence information for *all* known users.
+Matrix User IDs, or the object
+`synapse.events.presence_router.PresenceRouter.ALL_USERS` to indicate that the passed
+user should receive presence information for *all* known users.
 
 For clarity, if the user `@alice:example.org` is passed to this method, and the Set
 `{"@bob:example.com", "@charlie:somewhere.org"}` is returned, this signifies that Alice
@@ -85,6 +86,7 @@ Below is an example implementation of a presence router class.
 
 ```python
 from typing import Dict, Iterable, Set, Union
+from synapse.events.presence_router import PresenceRouter
 from synapse.handlers.presence import UserPresenceState
 from synapse.module_api import ModuleApi
 
@@ -163,23 +165,27 @@ class ExamplePresenceRouter:
 
         return destination_users
 
-    async def get_interested_users(self, user_id: str) -> Union[Set[str], str]:
+    async def get_interested_users(
+        self,
+        user_id: str,
+    ) -> Union[Set[str], PresenceRouter.ALL_USERS]:
         """
         Retrieve a list of users that `user_id` is interested in receiving the
         presence of. This will be in addition to those they share a room with.
-        Optionally, the literal str "ALL" can be returned to indicate that this user 
-        should receive all incoming local and remote presence updates.
+        Optionally, the object PresenceRouter.ALL_USERS can be returned to indicate
+        that this user should receive all incoming local and remote presence updates.
+
         Note that this method will only be called for local users.
 
         Args:
           user_id: A user requesting presence updates.
 
         Returns:
-          A set of user IDs to return additional presence updates for, or "ALL" to return
-          presence updates for all other users.
+          A set of user IDs to return additional presence updates for, or
+          PresenceRouter.ALL_USERS to return presence updates for all other users.
         """
         if user_id in self._config.always_send_to_users:
-            return "ALL"
+            return PresenceRouter.ALL_USERS
 
         return set()
 ```
--- a/docs/presence_router_module.md
+++ b/docs/presence_router_module.md
@@ -1,4 +1,4 @@
-# Presence Routing Module
+# Presence Router Module
 
 Synapse supports configuring a module that can specify additional users
 (local or remote) to should receive certain presence updates from local
@@ -34,6 +34,7 @@ Below is an example implementation of a presence router class.
 from typing import Dict, Iterable, Set, Union
 from synapse.handlers.presence import UserPresenceState
 from synapse.module_api import ModuleApi
+
 class PresenceRouterConfig:
     def __init__(self):
         # Config options with their defaults
@@ -43,10 +44,12 @@ class PresenceRouterConfig:
         # A list of users to ignore presence updates for. Does not affect
         # shared-room presence relationships
         self.blacklisted_users = []  # type: List[str]
+
 class ExamplePresenceRouter:
     """An example implementation of synapse.presence_router.PresenceRouter.
     Supports routing all presence to a configured set of users, or a subset
     of presence from certain users to members of certain rooms.
+
     Args:
         config: A configuration object.
         module_api: An instance of Synapse's ModuleApi.
@@ -54,12 +57,15 @@ class ExamplePresenceRouter:
     def __init__(self, config: PresenceRouterConfig, module_api: ModuleApi):
         self._config = config
         self._module_api = module_api
+
     @staticmethod
     def parse_config(config_dict: dict) -> PresenceRouterConfig:
         """Parse a configuration dictionary from the homeserver config, do
         some validation and return a typed PresenceRouterConfig.
+
         Args:
             config_dict: The configuration dictionary.
+
         Returns:
             A validated config object.
         """
@@ -67,11 +73,14 @@ class ExamplePresenceRouter:
         config = PresenceRouterConfig()
         always_send_to_users = config_dict.get("always_send_to_users")
         blacklisted_users = config_dict.get("blacklisted_users")
+
         # Do some validation of config options... otherwise raise a
         # synapse.config.ConfigError.
         config.always_send_to_users = always_send_to_users
         config.blacklisted_users = blacklisted_users
+
         return config
+
     async def get_users_for_states(
         self,
         state_updates: Iterable[UserPresenceState],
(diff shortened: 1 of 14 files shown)


erikjohnston:
> * The ability to trigger an initial sync for specific users, so they receive all current state. (TODO)

What do you mean by this?



> Something to consider: we may want to do some validation on Synapse's side to ensure modules aren't generating their own state? Or is that a feature, rather than a bug.

I don't think its worth trying to enforce TBH,

docs/presence_routing_module.md
```diff
@@ -0,0 +1,134 @@
+# Presence Routing Module
+
+Synapse supports configuring a module that can specify additional destinations (users or
+rooms) to receive certain presence updates from local users.
+
+Both local or remote users can be specified as a destination. If a room is specified, that
+will be translated to all users in that room. The homeserver must be participating in any
+specified room.
```
erikjohnston, line 8: I'm not 100% convinced we want to be able to specify rooms TBH, I think it may end up leading to a bunch of complexity for not much gain (unless we have specific use cases for it)

erikjohnston:
I've also realised that that there is another place where we have to hook in, which is where `/sync` tries to get all other users' presence the syncing user is interested in:

https://github.com/matrix-org/synapse/blob/8bcfc2eaad66aa6e92a745d226fee4a81a921541/synapse/handlers/presence.py#L1127-L1147

docs/presence_routing_module.md
anoadragon453: It was included as it would be rather easy to support ([`get_interested_parties`](https://github.com/matrix-org/synapse/blob/8bcfc2eaad66aa6e92a745d226fee4a81a921541/synapse/handlers/presence.py#L1305-L1329) already returns this), but we don't currently have any use cases. Off the top of my head it might be useful to have something like a "presence receivers" room where you could add/remove users remotely.

anoadragon453: Now that I've read your second comment though, supporting rooms would not be as simple.

anoadragon453: Well, actually the sync code already supports having an explicit room ID attached to it. We could just add more room IDs to retrieve users from :)

anoadragon453:
>> * The ability to trigger an initial sync for specific users, so they receive all current state. (TODO)

> What do you mean by this?

When another user is determined to receive presence for all users, it's crucial to be able to send them all existing presence of all local users - not just the latest updates as the proposal currently provides. Thus, either:

1. Synapse, upon detecting a change from what the module returned last, could send all current user presence to the new user. This would ensure they are always kept up to date. However, would be wildly inefficient if the module were changing the list of returned users constantly, plus I think it reduces the flexibility of the module.
2. `get_rooms_and_users_for_states` could return another list containing users that should receive all local user presence. That still sounds a bit finicky though, especially if we're updating the list of users to send presence to asynchronously.
3. We could add a new method to `ModuleApi` that allows modules to, at any point, send all local user presence to a given set of users. This would allow the module's logic to determine when this should happen, separately from when `get_rooms_and_users_for_states` is called.

I'm leaning towards 3. My only reservation is whether this would be useful in the generic sense but... probably?

>> Something to consider: we may want to do some validation on Synapse's side to ensure modules aren't generating their own state? Or is that a feature, rather than a bug.

> I don't think its worth trying to enforce TBH,

Yeah, probably not. The sysadmin did decide to install that module on their own :slightly_smiling_face:

anoadragon453:
@erikjohnston I've gone with 3. for now and defined `ModuleApi.send_local_online_presence_to(users: List[str]) -> None`, which can be used to update a user on the current state of local user presence, which is useful if they've been offline for a while.

anoadragon453:
For the implementation of `ModuleApi.send_local_online_presence_to(users: List[str]) -> None`, I'm envisioning an implementation along the lines of the following: 

```py
self._presence_stream = self.hs.get_event_sources().sources["presence"]
#...
def ModuleApi.send_local_online_presence_to(users: List[str]) -> None:
    for user in users:
        if self.hs.is_mine(user):
            # Modify SyncHandler._generate_sync_entry_for_presence to call
            # presence_source.get_new_events with an empty `from_key` if
            # that user's ID were in a list modified by ModuleApi somewhere.
            # That user would then get all presence state on next incremental sync.
        else:
            # Retrieve presence state for all currently online users
            presence_events = await self._presence_stream.get_new_events(
                 user, from_key=None, include_offline=False
            )
            FederationSender.send_presence([ev.state for ev in presence_events])
```

Technically this method is only needed for remote users (local users could just initial sync), but it's much more convenient if the module can specify both.

erikjohnston:
Basically yeah I think that works. Instead of broadcasting all local presence states it should only send states that the given user is interested in though.

anoadragon453:
> Basically yeah I think that works. Instead of broadcasting all local presence states it should only send states that the given user is interested in though.

Oh true. And that could be every local user if used in conjunction with `get_rooms_and_users_for_states`.

docs/presence_routing_module.md
```diff
@@ -0,0 +1,137 @@
...
+        desired_updates = set()
+        for update in state_updates:
+            if update.state_key not in self._config.blacklisted_users:
+                desired_updates.add(update)
+
+        # Send all presence updates to specific users
+        for user_id in self._config.always_send_to_users:
+            destination_users[user_id] = desired_updates
+
+        return destination_users
```
anoadragon453, line 113: So this method works great for when you have a set of state updates you want to send out. But I'm increasingly realising it works terribly for when you have a user that you want to query the current state of other users for.

Where we were looking to hook in, we'd need to somehow return user IDs of users on the server that we should query presence for. `get_users_for_states` expects a set of `UserPresenceState` which at this point we don't have yet. We actually *purposefully* don't have that yet. At the linked call, we try to just figure out the IDs of users we want to pull presence of, instead of pulling all the presence right away. If we do that (so that we can pass `UserPresenceState` instances to `get_users_for_states`), then we lose all of the optimisation this function is aiming for (only pulling state at the very end after we've filtered out users).

We also can't just add another method that says "for a given user syncing, which users should we try to get updates for?" as the module won't know the user IDs of anyone else on the homeserver except those it's hoping to route presence to.

So I'm a bit stuck on how to support the collection of presence through `/sync` while honoring what `get_users_for_states` returns at the moment.

anoadragon453: Oops, I meant to link the following for where we planned to hook into the `/sync` code:

https://github.com/matrix-org/synapse/blob/e55bd0e11012ee14a8b18ece16f4837b86212f5c/synapse/handlers/presence.py#L1089.

erikjohnston: Hmm. Looking at the sync code I wonder if we can make `_get_interested_in` return a `"all"` value, which we can then filter after pulling out the states that have changed:

https://github.com/matrix-org/synapse/blob/0279e0e08623f8d9e8dc680c65b2e94ae2386d9e/synapse/handlers/presence.py#L1089-L1117

anoadragon453: Possibly, however `get_users_for_states` currently supports routing a subset of incoming presence to some users, rather than all of the presence.

So while that would support the "all" usecase, it wouldn't work if the module wanted to only route *some* presence to a user.

We could just throw our hands up and make `get_users_for_states` only support routing *all* local user presence, but the generality is further diminished.

erikjohnston: I was thinking that we could have a function `get_interested_in(user) -> Union[Set[str], All]`, to figure out which presence updates to fetch, and then pass those to another function to filter them down correctly (or reuse `get_users_for_states`).

anoadragon453: Right, so that proposed function would be in the module, correct? That sounds doable.

My only concern is that if `get_interested_in` is not implemented by the module, then `get_users_for_states` may not work as expected for local users. When local presence updates come in, `get_users_for_states` is simply used to figure out which user `/sync` streams to wake up. We would then query `get_interested_in` to determine which state events to pull.

So we'd need to trust the module developer to return similar things for both functions. We is fine and we could document that, though ideally we'd just have a single function that could be called for both cases.

erikjohnston: True. Though we can add strong "here be dragons" warnings.

I wonder if we can write default implementations, `get_interested_in` always returning `all` and then use `get_users_for_states` as the way of filtering?

anoadragon453: That would work... and then it would just be a performance hit if the module didn't implement `get_interested_in`. And worst case is the desyncronisation of results, but hopefully the documentation would do enough here to make not doing that clear.

I'll move forward with that for now, thanks!

docs/presence_router_module.md
```diff
@@ -0,0 +1,155 @@
...
+        Note that this method will only be called for local users.
+        Args:
+          user_id: A user requesting presence updates.
+        Returns:
+          A set of user IDs to return additional presence updates for, or "ALL" to return
+          presence updates for all other users.
+        """
+        if user_id in self._config.always_send_to_users:
+            return "ALL"
+        return set()
```
erikjohnston, line 113: I think this needs to be formatted a bit, at least add some newlines between the function definitions :)

docs/presence_router_module.md
```diff
@@ -0,0 +1,155 @@
...
+
+```python
+async def ModuleApi.send_local_online_presence_to(users: Iterable[str]) -> None
+```
+
+which can be given a list of local or remote MXIDs to broadcast known, online user
+presence to (for those users that the receiving user is considered interested in). 
+It does not include state for users who are currently offline.
+
+### Example
```
erikjohnston, line 29: Can we define the required methods and describe them first, and then have an example after please?

synapse/events/presence_router.py
```diff
@@ -0,0 +1,92 @@
...
+
+        Args:
+            state_updates: An iterable of user presence state updates.
+
+        Returns:
+          A dictionary of user_id -> set of UserPresenceState, indicating which
+          presence updates each user should receive.
+        """
+        if self.custom_presence_router is not None and hasattr(
+            self.custom_presence_router, "get_users_for_states"
```
erikjohnston, line 57: I wonder if we should make these both required methods? We can then error if we try and load a module that doesn't have the expected functions?

synapse/module_api/__init__.py
```diff
@@ -50,11 +51,17 @@ def __init__(self, hs, auth_handler):
...
+        self._presence_stream = hs.get_event_sources().sources["presence"]
 
         # We expose these as properties below in order to attach a helpful docstring.
         self._http_client = hs.get_simple_http_client()  # type: SimpleHttpClient
         self._public_room_list_manager = PublicRoomListManager(hs)
 
+        # The next time these users sync, they will receive the current presence
+        # state of all local users. Users are added by send_local_online_presence_to,
+        # and removed after a successful sync.
+        self.send_full_presence_to_local_users = set()
```
erikjohnston, line 63:
```suggestion
        self._send_full_presence_to_local_users = set()
```

I think we should only have private attrs here, to avoid plugins ever thinking its a good idea to mess with the attributes directly?

synapse/handlers/presence.py
```diff
@@ -1083,59 +1110,192 @@ async def get_new_events(
...
+        return presence_updates, max_token
+
+    async def _filter_all_presence_updates_for_user(
+        self,
+        user_id: str,
+        max_token: int,
+        from_key: int,
+        include_offline: bool,
+    ) -> Tuple[List[UserPresenceState], int]:
+        # Only return updates since the last sync
```
erikjohnston, line 1200: docstring please

docs/presence_router_module.md
anoadragon453: huh, there definitely *used* to be spaces there. I wonder where they went...

synapse/module_api/__init__.py
anoadragon453: Hmm true, though we do need to access this from other bits of Synapse. That's probably fine if we add a comment.

An alternative is to make this a global variable at the file level?

synapse/events/presence_router.py
anoadragon453: Yeah, I think that's a good idea. Would save people from falling into the trap of only implementing one.

docs/presence_router_module.md
```diff
@@ -0,0 +1,222 @@
+# Presence Router Module
```
anoadragon453, line 1: I was asked today about how one would implement this module for an application service, instead of a user calling `/sync`, and had a bit of a look through the code. I noticed a couple of things:

1. There's some code that determines whether an application service is interested in a user or a room. This is used to check whether we need to consider sending an event linked to a room/user to the application service. We don't hook into this yet. I suppose we could add a call to `get_interested_users` here and call it with the `user_id` argument set to the Application Service's `sender_localpart` user? `get_users_for_states` would still be called to filter presence updates once an application service had been declared interested.

    This could also be worked around by having the AS expand their namespace regexes, thus making them interested in a wide range of users (potentially all), but I think that's a bit hacky.

2. The `ModuleApi.send_local_online_presence_to` method is currently only set up to modify `/sync` responses. I know we're trying to discourage AS users from syncing. So we probably need to modify the method to check if the user is a known application service `sender_localpart`, and if so send a burst of all current user presence states in a transaction.

I'll also add some tests for application services to verify what does and doesn't already work.

anoadragon453: After testing locally appservices do not currently receive presence updates.

However in the interest of not complicating this PR further, I'm going to leave a note about this and aim to fix it in a future PR.

anoadragon453:
@erikjohnston I've force-pushed to incorporate the changes from #9650. The commits since your last review are 41f9cd1 onwards.

docs/presence_router_module.md
```diff
@@ -0,0 +1,224 @@
...
+
+Synapse supports configuring a module that can specify additional users
+(local or remote) to should receive certain presence updates from local
+users.
+
+Note that routing presence via Application Service transactions is not
```
erikjohnston, line 7:
```suggestion
Note that routing presence via an Application Service transactions is not
```

synapse/module_api/__init__.py
```diff
@@ -385,6 +392,34 @@ async def create_and_send_event_into_room(self, event_dict: JsonDict) -> EventBa
...
+
+                # Force a presence initial_sync for this user next time
+                self.send_full_presence_to_local_users.add(user)
+            else:
+                # Retrieve presence state for currently online users that this user
+                # is considered interested in
+                presence_events = await self._presence_stream.get_new_events(
+                    user, from_key=None, include_offline=False
+                )
+                await FederationSender.send_presence(
```
erikjohnston, line 419: Err, doesn't this need to be `self._federation_sender` or something?

anoadragon453: Yep! That'll teach me to not add tests for federation sending...

anoadragon453:
So tests broke briefly because we were attempting to load a `FederationSender` in `ModuleApi`'s `__init__`, which some tests call without having set `send_federation: True` in their config (as expected).

So I now only load it in `send_local_online_presence_to` and added notes everywhere cautioning that the method must only be called on workers that support federation sending.

That lead me to another thought though, in that if you configure a `PresenceRouter` per worker, and your module is doing things like running `ModuleApi.send_local_online_presence_to`  when something changes externally, that's going to happen multiple times over. So modules probably need an option to prevent more than one instance trying to broadcast presence information.

Any idea where I could add a note for that, or whether there's any precedent for configuring modules per-worker thus far? (I imagine you'd just stick a `presence.presence_router` block in each worker config...)

synapse/handlers/presence.py
```diff
@@ -1091,59 +1118,207 @@ async def get_new_events(
...
+        # explicit_room_id either. Assert cache_context is not None so we can use it
+        # without mypy complaining.
+        assert cache_context
+
+        # Check with the presence router whether we should poll additional users for
+        # their presence information
+        additional_users = await self.get_presence_router().get_interested_users(
+            user.to_string()
+        )
+        if additional_users == "ALL":
```
erikjohnston, line 1312: We should move `"ALL"` to be a constant

docs/presence_router_module.md
```diff
@@ -0,0 +1,225 @@
...
+#### `get_users_for_states`
+
+```python
+async def get_users_for_states(
+    self,
+    state_updates: Iterable[UserPresenceState],
+) -> Dict[str, Set[UserPresenceState]]:
+```
+
+**Required.** An asynchronous class method that is passed an iterable of user presence
```
erikjohnston, line 56:
```suggestion
**Required.** An asynchronous method that is passed an iterable of user presence
```

"class methods" are a different thing

docs/presence_router_module.md
```diff
@@ -0,0 +1,225 @@
...
+state. This method can determine whether a given presence update should be sent to certain
+users. It does this by returning a dictionary with keys representing local or remote
+Matrix User IDs, and values being a python Set
+of `synapse.handlers.presence. UserPresenceState` instances.
```
erikjohnston, line 60:
```suggestion
Matrix User IDs, and values being a python set
of `synapse.handlers.presence.UserPresenceState` instances.
```

docs/presence_router_module.md
```diff
@@ -0,0 +1,225 @@
...
+Synapse will then attempt to send the specified presence updates to each user when
+possible.
+
+#### `get_interested_users`
+
+```python
+async def get_interested_users(self, user_id: str) -> Union[Set[str], str]
+```
+
+**Required.** An asynchronous class method that is passed a single Matrix User ID. This
```
erikjohnston, line 71:
```suggestion
**Required.** An asynchronous method that is passed a single Matrix User ID. This
```

docs/presence_router_module.md
```diff
@@ -0,0 +1,225 @@
...
+method is expected to return the users that the passed in user may be interested in the
+presence of. This may be local or remote users. It does so by returning a python Set of
```
erikjohnston, line 73:
```suggestion
presence of, in addition to users they share a room with. This may be local or remote users. It does so by returning a python Set of
```

synapse/handlers/presence.py
```diff
@@ -1091,59 +1118,207 @@ async def get_new_events(
...
                 else:
-                    user_ids_changed = users_interested_in
+                    # This user should receive all user presence, and hasn't provided a from_key.
+                    # Send all currently known user presence states.
+                    users_to_state = await self.store.get_presence_for_all_users(
+                        include_offline=include_offline
+                    )
 
-            updates = await presence.current_state_for_users(user_ids_changed)
+                    return list(users_to_state.values()), max_token
```
erikjohnston, line 1144: Did we not decide that we wanted to filter users through `get_users_for_states` if we returned `ALL`?

anoadragon453:
> Hmm, I think its fine so long as `send_local_online_presence_to` is only called in response to the presence handler poking it, as that will then only get run on one worker?

Right, however `send_local_online_presence_to` was specifically designed to be called when the list of presence-receiving users was appended to. That can happen in response to something changing externally (say if an external, block-like data structure updated to include more users :wink:).

I think this is fine, and the module implementer would just need to be aware and only call the module on the right workers.

synapse/handlers/presence.py
anoadragon453: This was the idea originally, however I struggled to make the typing work effectively. I just tried it again now and had zero problems. Likely because I'm just comparing strings now instead of trying to use `isinstance` and similar.

synapse/handlers/presence.py
anoadragon453: We did, and this is indeed a path that does not do so! I've made this path use `_filter_all_presence_updates_for_user` in 2f16ed047, as well as made that function a bit cleaner.

synapse/module_api/__init__.py
anoadragon453: Have made it private.

erikjohnston:
@anoadragon453 the isort is failing :)

anoadragon453:
:facepalm: fixed.

erikjohnston (approved):

PR #9493, opened 2021-02-25
Update JWT login type to support JWKS, custom sub claim, and encode special chars in user ID

boti7:
This PR adds the following optional fields to the JWT login type configuration:
- `jwks_uri` allows specifying a JSON Web Key Set endpoint where RSA signing keys can be retrieved from, instead of using a secret or public key from the configuration.
- `subject_claim` allows specifying a claim other than the default "sub" to use as localpart of the Matrix ID.
- `normalize_user_id` enables mapping the localpart from other character sets using the `map_username_to_mxid_localpart` function.

### Pull Request Checklist

Signed-off-by: Botond Horvath <hello@boti7.com>


clokep:
Looks like reasonable improvements! 👍 This seems broken on Python 3.5 (which pyjwt technically doesn't support: jpadilla/pyjwt#585, Synapse will be dropping it shortly). Also could you see if any of the documentation under `docs/jwt.md` needs to be updated? (I don't know if we need to detail all the options there + the sample config, but at least mentioning some of the additional features likely makes sense.)

clokep:
@boti7 It seems that some unrelated changes were pushed onto this branch for CI. If those are improvements that would be useful we should do them as a separate PR.

Did you see my comment above? Note that we'll be dropping support for Python 3.5 soon, so might make sense to just wait on that if possible.

worldofgeese:
@clokep this was my fault. We'll roll it all back.

clokep:
(Also merging in develop should fix the mypy issues.)

worldofgeese:
@clokep I've hard reset to @boti7's last change. Let me know if there's anything else I can do.

synapse/config/jwt_config.py
```diff
@@ -29,11 +29,14 @@ def read_config(self, config, **kwargs):
...
         if jwt_config:
             self.jwt_enabled = jwt_config.get("enabled", False)
-            self.jwt_secret = jwt_config["secret"]
             self.jwt_algorithm = jwt_config["algorithm"]
 
             # The issuer and audiences are optional, if provided, it is asserted
             # that the claims exist on the JWT.
+            self.jwt_secret = jwt_config.get("secret")
+            self.jwt_jwks_uri = jwt_config.get("jwks_uri")
+            self.jwt_subject_claim = jwt_config.get("subject_claim")
```
clokep, line 38: It would be clearer to have this default to `"sub"` here instead of down in the login code!

synapse/config/jwt_config.py
```diff
@@ -29,11 +29,14 @@ def read_config(self, config, **kwargs):
         jwt_config = config.get("jwt_config", None)
         if jwt_config:
             self.jwt_enabled = jwt_config.get("enabled", False)
-            self.jwt_secret = jwt_config["secret"]
             self.jwt_algorithm = jwt_config["algorithm"]
 
             # The issuer and audiences are optional, if provided, it is asserted
             # that the claims exist on the JWT.
+            self.jwt_secret = jwt_config.get("secret")
+            self.jwt_jwks_uri = jwt_config.get("jwks_uri")
```
clokep, line 37: We should add some sort of assertion that one of these is set.

synapse/config/jwt_config.py
```diff
@@ -29,11 +29,14 @@ def read_config(self, config, **kwargs):
...
+            self.jwt_subject_claim = jwt_config.get("subject_claim")
+            self.jwt_normalize_user_id = jwt_config.get("normalize_user_id", False)
```
clokep, line 39: I'm assuming we want to default to `False` for backwards compatibility reasons?

synapse/rest/client/v1/login.py
```diff
@@ -298,11 +301,19 @@ async def _do_jwt_login(self, login_submission: JsonDict) -> Dict[str, str]:
...
 
         import jwt
+        from jwt import PyJWKClient
+
+        key = self.jwt_secret
+
+        if not key and self.jwt_jwks_uri:
+            jwks_client = PyJWKClient(self.jwt_jwks_uri)
+            signing_key = jwks_client.get_signing_key_from_jwt(token)
+            key = signing_key.key
```
clokep, line 311: It would be ideal if we only had to fetch this once (similar to the `_provider_metadata` property of `OidcHandler`).

synapse/rest/client/v1/login.py
```diff
@@ -298,11 +301,19 @@ async def _do_jwt_login(self, login_submission: JsonDict) -> Dict[str, str]:
             )
 
         import jwt
+        from jwt import PyJWKClient
+
+        key = self.jwt_secret
+
+        if not key and self.jwt_jwks_uri:
+            jwks_client = PyJWKClient(self.jwt_jwks_uri)
+            signing_key = jwks_client.get_signing_key_from_jwt(token)
```
clokep, line 310: Does this make a synchronous HTTP call? If so we ideally would do this via a `SimpleHttpClient` or push this into the background.

clokep (commented):
Looks pretty reasonable overall. I suspect it would be a nice follow-up to move the JWT code into a separate handler instead of sprinkling it into `LoginRestServlet`.

boti7:
I just made a small change to `docs/jwt.md` and I think the other options are clear from the sample config, but let me know if you want me to add anything else.

From my side it's totally fine to wait until you drop support for Python 3.5. Also merged in develop, but there are still some mypy issues.

synapse/config/jwt_config.py
boti7: At first I tried to specify the default value here, maybe I made a mistake, but it didn't work for some reason.

synapse/config/jwt_config.py
boti7: I'm not sure how you validate the config file usually. Is that OK if we raise a `ConfigError` if none of these is set?

synapse/config/jwt_config.py
boti7: Exactly.

synapse/config/jwt_config.py
clokep: That's surprising! We do that in many places, could you try again? I hope it is just changing this to `jwt_config.get("subject_claim", "sub")`.

anoadragon453:
We've now dropped Python 3.5 support :tada:

erikjohnston:
I think we're still waiting for some changes here? If nothing else there is a merge conflict. @boti7 please nudge if/when its ready for review :slightly_smiling_face:

e-001:
@boti7 I'll clean up this conflict and PR. Desperately need escaped sub claim. Auth0 lovingly uses a pipe to separate auth method and ID.

e-001:
a77ebc96a4525c76303977b9fd4041e6dcb35cff Caught up to develop and refactored for 683d6f75af0e941e9ab3bc0a985aa6ed5cc7a238. Waiting on mqsdk/synapse#1.

erikjohnston:
Looks like the JWT tests are failing when using old-deps, possibly need to bump the minimum version of the library?

boti7:
@erikjohnston It's ready for review from my side.

synapse/config/jwt.py
```diff
@@ -75,10 +85,30 @@ def generate_config_section(self, **kwargs):
...
+            #
+            #jwks_uri: "provided-by-your-issuer"
+
+            # Name of the claim containing a unique identifier for the user.
+            #
+            # Optional, defaults to `sub`.
+            #
+            #subject_claim: "sub"
+
+            # Perform normalization of the user ID and encode unsupported characters.
```
clokep, line 106: We use British spelling.

```suggestion
            # Perform normalisation of the user ID and encode unsupported characters.
```

synapse/python_dependencies.py
```diff
@@ -107,7 +107,7 @@
     "url_preview": ["lxml>=3.5.0"],
     "sentry": ["sentry-sdk>=0.7.2"],
     "opentracing": ["jaeger-client>=4.0.0", "opentracing>=2.2.0"],
-    "jwt": ["pyjwt>=1.6.4"],
+    "jwt": ["pyjwt>=2.1.0"],
```
clokep, line 110: I asked the packages whether this change would be OK. Can you remind me why we're bumping this?

synapse/rest/client/v1/login.py
clokep: Fetching this for every login seems quite inefficient. I wonder if we should do something on start-up (like OIDC). It looks like `PyJWKClient` is quite simple!

synapse/rest/client/v1/login.py
```diff
@@ -336,10 +347,14 @@ async def _do_jwt_login(self, login_submission: JsonDict) -> Dict[str, str]:
                 errcode=Codes.FORBIDDEN,
             )
 
-        user = payload.get("sub", None)
+        subject_claim = self.jwt_subject_claim or "sub"
+        user = payload.get(subject_claim, None)
```
clokep, line 351: We already did the fallback to `sub` in the config code, no need to do it again.

```suggestion
        user = payload.get(self.jwt_subject_claim, None)
```

clokep (changes requested):
Looks pretty good overall! I wonder if there's enough JWT logic here now that we should move this code to a separate `JwtHandler`. What do you think?

synapse/python_dependencies.py
clokep: Looks like supported versions of Debian and Fedora are on 1.7.1 still. Would it be possible to use that?

boti7: `PyJWKClient` was added in v2.0.0 and caching of signing keys is supported since v2.1.0: https://github.com/jpadilla/pyjwt/blob/master/CHANGELOG.rst#v210

synapse/rest/client/v1/login.py
boti7: The issue is that tests don't use the config code but set `hs.config` directly, so without specifying a fallback here, many tests in `JWTTestCase` would fail. What do you suggest to solve this?

synapse/config/jwt.py
boti7: Do you want to change the spelling in the code too?

synapse/rest/client/v1/login.py
clokep: The config code should get run during tests, see:

https://github.com/matrix-org/synapse/blob/fe604a022a7142157da7e90a40330beb2a11af7a/tests/unittest.py#L469-L472

clokep: Were you able to try this again? It should run fine during tests.

clokep:
Removing my review -- this has a couple of open questions still that are waiting for responses!

synapse/python_dependencies.py
clokep: I'm not really sure what to do about this, although I do wonder if we should be using `PyJWKClient` since it bypasses the reactor / current methods for communicating with external resources.

anoadragon453:
Hey @boti7, are you able to continue working on this? :slightly_smiling_face:

Also note that there is now a conflict in `synapse/rest/client/v1/login.py`. You'll need to simply merge the `develop` branch into yours to resolve it.

anoadragon453:
Given the lack of activity, I'm going to close this. Anyone is free to pick it back up in a new PR though!

PR #9496, opened 2021-02-25
Test that we require validated email for email pushers

erikjohnston:


clokep (approved):
Seems reasonable to me!

PR #9497, opened 2021-02-25
(Hopefully) stop leaking file descriptors in media repo

### LINKED ISSUE #9488: media repo worker leaks open FDs
There is a sawtooth on the "Open FDs" graph for the media repository: they accumulate until a Gen 2 GC happens:

image

The accumulated FDs appear to be `CLOSE_WAIT` sockets; they mostly appear to be outbound connections to port 80 or 443 so they are probably related to url preview requests.

clokep:
I think this fixes #9488 (leaking of file descriptors in the media repo). It seems that in the case where we rejected the response due to the headers returning a content size that the response is never rejected and sits around until it is garbage collected. There doesn't seem to be a way to abort a connection directly from the response so this adds another protocol which immediately errors with a `BodyExceededMaxSize` exception.

I'm not positive this will fix the issue, but I'm hopeful!

Note that the first two commits here are just munging some test code so we can add a new test.

### DIFF
--- a/tests/http/test_client.py
+++ b/tests/http/test_client.py
@@ -26,77 +26,83 @@ from tests.unittest import TestCase
 
 
 class ReadBodyWithMaxSizeTests(TestCase):
-    def setUp(self):
+    def _build_response(self, length=UNKNOWN_LENGTH):
         """Start reading the body, returns the response, result and proto"""
-        response = Mock(length=UNKNOWN_LENGTH)
-        self.result = BytesIO()
-        self.deferred = read_body_with_max_size(response, self.result, 6)
+        response = Mock(length=length)
+        result = BytesIO()
+        deferred = read_body_with_max_size(response, result, 6)
 
         # Fish the protocol out of the response.
-        self.protocol = response.deliverBody.call_args[0][0]
-        self.protocol.transport = Mock()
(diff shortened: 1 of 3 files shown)


tests/http/test_client.py
```diff
@@ -104,3 +104,18 @@ def test_additional_data(self):
...
+        # Deferred shouldn't be called yet.
+        self.assertFalse(deferred.called)
+
+        # Start sending data.
+        protocol.dataReceived(b"12345")
+        self._assert_error(deferred, protocol)
+        self._cleanup_error(deferred)
+
+        # The data is never consumed.
+        self.assertEqual(result.getvalue(), b"")
```
erikjohnston, line 121: Can we test that that transport was closed?

erikjohnston (approved):

tests/http/test_client.py
clokep: We should assert that `abortConnection` was closed, good catch!

clokep: Actually, this is already done in `_assert_error`. Is there more you mean by ensuring it is closed?

erikjohnston: Oh, I didn't spot that either. No, I think this is fine then

clokep: Great! 👍

PR #9498, opened 2021-02-25
Purge chain cover tables when purging events.

### LINKED ISSUE #9481: Federated room cannot be joined again when left and deleted before
### Description

When a federated room (e.g. #synapse:matrix.org) is joined from self hosted federated homeserver, then left, deleted via API from self homeserver, then rejoin fails.

### Steps to reproduce

1. Join some room (e.g. #synapse:matrix.org) from self hosted homeserver (same version of Synapse 1.27.0).
2. Leave that room
3. Delete that room using API `/_synapse/admin/v1/rooms/[room_id]/delete`
4. Try to join that room again
5. It fails with following error...
```
synapse.http.server: [POST-10040] Failed handle request via 'JoinRoomAliasServlet': <XForwardedForRequest at 0x7f57646fa970 method='POST' uri='/_matrix/client/r0/join/%23synapse%3Amatrix.org' clientproto='HTTP/1.1' site='8008'>
      File "/usr/lib/python3.9/site-packages/synapse/http/server.py", line 252, in _async_render_wrapper
      File "/usr/lib/python3.9/site-packages/synapse/http/server.py", line 430, in _async_render
      File "/usr/lib/python3.9/site-packages/synapse/rest/client/v1/room.py", line 301, in on_POST
        await self.room_member_handler.update_membership(
      File "/usr/lib/python3.9/site-packages/synapse/handlers/room_member.py", line 333, in update_membership
        result = await self.update_membership_locked(
      File "/usr/lib/python3.9/site-packages/synapse/handlers/room_member.py", line 549, in update_membership_locked
        remote_join_response = await self._remote_join(
      File "/usr/lib/python3.9/site-packages/synapse/handlers/room_member.py", line 1091, in _remote_join
        event_id, stream_id = await self.federation_handler.do_invite_join(
      File "/usr/lib/python3.9/site-packages/synapse/handlers/federation.py", line 1400, in do_invite_join
        max_stream_id = await self._persist_auth_tree(
      File "/usr/lib/python3.9/site-packages/synapse/handlers/federation.py", line 2050, in _persist_auth_tree
        await self.persist_events_and_notify(
      File "/usr/lib/python3.9/site-packages/synapse/handlers/federation.py", line 2925, in persist_events_and_notify
        events, max_stream_token = await self.storage.persistence.persist_events(
      File "/usr/lib/python3.9/site-packages/synapse/storage/persist_events.py", line 262, in persist_events
        ret_vals = await make_deferred_yieldable(
    twisted.internet.defer.FirstError: FirstError[#0, [Failure instance: Traceback: <class 'psycopg2.errors.UniqueViolation'>: duplicate key value violates unique constraint "event_auth_chains_pkey"
    DETAIL:  Key (event_id)=($e9U026auDHIgaZPAqlblvPupACjl7jcZDblP970dJPs) already exists.

    /usr/lib/python3.9/site-packages/synapse/metrics/background_process_metrics.py:208:run
    --- <exception caught here> ---
    /usr/lib/python3.9/site-packages/synapse/storage/persist_events.py:172:handle_queue_loop
    /usr/lib/python3.9/site-packages/synapse/storage/persist_events.py:324:persisting_queue
    /usr/lib/python3.9/site-packages/synapse/storage/persist_events.py:532:_persist_events
    /usr/lib/python3.9/site-packages/synapse/storage/databases/main/events.py:171:_persist_events_and_state_updates
    /usr/lib/python3.9/site-packages/synapse/storage/database.py:661:runInteraction
    /usr/lib/python3.9/site-packages/synapse/storage/database.py:744:runWithConnection
    /usr/lib64/python3.9/site-packages/twisted/python/threadpool.py:250:inContext
    /usr/lib64/python3.9/site-packages/twisted/python/threadpool.py:266:<lambda>
    /usr/lib64/python3.9/site-packages/twisted/python/context.py:122:callWithContext
    /usr/lib64/python3.9/site-packages/twisted/python/context.py:85:callWithContext
    /usr/lib64/python3.9/site-packages/twisted/enterprise/adbapi.py:306:_runWithConnection
    /usr/lib64/python3.9/site-packages/twisted/python/compat.py:464:reraise
    /usr/lib64/python3.9/site-packages/twisted/enterprise/adbapi.py:297:_runWithConnection
    /usr/lib/python3.9/site-packages/synapse/storage/database.py:739:inner_func
    /usr/lib/python3.9/site-packages/synapse/storage/database.py:539:new_transaction
    /usr/lib/python3.9/site-packages/synapse/logging/utils.py:71:wrapped
    /usr/lib/python3.9/site-packages/synapse/storage/databases/main/events.py:379:_persist_events_txn
    /usr/lib/python3.9/site-packages/synapse/storage/databases/main/events.py:472:_persist_event_auth_chain_txn
    /usr/lib/python3.9/site-packages/synapse/storage/databases/main/events.py:630:_add_chain_cover_index
    /usr/lib/python3.9/site-packages/synapse/storage/database.py:896:simple_insert_many_txn
    /usr/lib/python3.9/site-packages/synapse/storage/database.py:274:execute_batch
    /usr/lib/python3.9/site-packages/synapse/storage/database.py:319:_do_execute
    /usr/lib/python3.9/site-packages/synapse/storage/database.py:274:<lambda>
    /usr/lib64/python3.9/site-packages/psycopg2/extras.py:1209:execute_batch
    ]]
```

The room should be rejoined without any problem.

### Version information

- **Homeserver**: matrix.bolesiny.net
- **Version**: 1.27.0

- **Install method**: Fedora rpm package

- **Platform**: Fedora 33

clokep:
When purging a room we should be purging the corresponding entries on the chain cover index.

Two of the tables (`event_auth_chains` and `event_auth_chains_to_calculate`) are easy since they reference `event_id`, but `event_auth_chain_links` is a bit harder since it does not.

I'm not really sure this is the correct approach, but would like some feedback.

This seems to fix #9481.

I'd like to include a fix for people who historically left a room as well, but I'm not sure there's any good way to do that. (Trying to find all events which no longer exist seems like it would be expensive.)

I do not think the same code should be applied to `purge_history` (as opposed to `purge_room`) in this file since that doesn't seem to remove state events.


richvdh:
shouldn't we have had some foreign key references which would have caught this problem earlier? :/

clokep:
> shouldn't we have had some foreign key references which would have caught this problem earlier? :/

I think most of our tables don't have foreign keys. I'd like to at least add a test that joins a room, leaves that room, then re-joins it. 😢

richvdh:
> I think most of our tables don't have foreign keys

this is sort of my point. These are recently added tables; why are we still not using foreign keys? And sorry, this isn't really the place to debate it, I'm just a bit sad that we keep repeating historical mistakes.

richvdh:
btw the approach here looks correct; but I'm not that familiar with the auth chain stuff, so I can't speak to whether the details are correct.

> I do not think the same code should be applied to purge_history (as opposed to purge_room) in this file since that doesn't seem to remove state events.

I would still think we should at least delete the relevant rows in  `event_auth_chains` ?

synapse/storage/databases/main/purge_events.py
```diff
@@ -310,6 +315,29 @@ def _purge_room_txn(self, txn, room_id):
...
+
+        logger.info("[purge] removing events from event_auth_chain_links")
+        txn.executemany(
+            "DELETE FROM event_auth_chain_links WHERE "
+            "(origin_chain_id = ? AND origin_sequence_number = ?) OR "
+            "(target_chain_id = ? AND target_sequence_number = ?)",
+            (
+                (chain_id, seq_num, chain_id, seq_num)
+                for (chain_id, seq_num) in referenced_chain_id_tuples
+            ),
```
erikjohnston, line 338: Can we use multiline strings please?

erikjohnston (approved):

clokep:
@erikjohnston Any thoughts on if it would make sense to do a background update to try to fix previously purged rooms?

erikjohnston:
> @erikjohnston Any thoughts on if it would make sense to do a background update to try to fix previously purged rooms?

Ugh, we probably should do something to fix this I guess, as currently there's no way of getting out of this mess when you discover it. I think its just a case of iterating over all the rows in `event_auth_chains` and removing them if we don't have an associated event in the `events` table

waclaw66:
Is there any particular reason why Synapse doesn't use foreign keys? They would make all the maintenance much easier.

richvdh:
> Is there any particular reason why Synapse doesn't use foreign keys? They would make all the maintenance much easier and keep database consistent.

I feel like I raised this point earlier: https://github.com/matrix-org/synapse/pull/9498#issuecomment-786184632

clokep:
From our meeting today we discussed:

* Merging this as is.
* Following-up with a background update to fix for people who have already purged rooms.

PR #9499, opened 2021-02-26
Add bugbear, add B00 to ignored checks, and fix 2 violation codes

ShadowJonathan:
Adds `flake8-bugbear` to the test suite, and immediately ignores it in the `setup.cfg`, to be fixed later (see #9366)

### Pull Request Checklist

Part of #9366

### DIFF
--- a/synapse/app/__init__.py
+++ b/synapse/app/__init__.py
@@ -24,7 +24,9 @@ logger = logging.getLogger(__name__)
 try:
     python_dependencies.check_requirements()
 except python_dependencies.DependencyException as e:
-    sys.stderr.writelines(e.message)
+    sys.stderr.writelines(
+        e.message  # noqa: B306, DependencyException.message is a property
+    )
     sys.exit(1)
 
 
--- a/tests/unittest.py
+++ b/tests/unittest.py
@@ -140,7 +140,7 @@ class TestCase(unittest.TestCase):
             try:
                 self.assertEquals(attrs[key], getattr(obj, key))
             except AssertionError as e:
-                raise (type(e))(e.message + " for '.%s'" % key)
+                raise (type(e))("Assert error for '.{}':".format(key)) from e
 
     def assert_dict(self, required, actual):
         """Does a partial assert of a dict.
--- a/changelog.d/9499.misc
+++ b/changelog.d/9499.misc
@@ -1,0 +1,1 @@
+Introduce bugbear to the test suite.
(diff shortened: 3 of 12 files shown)


synapse/app/__init__.py
```diff
@@ -24,7 +24,9 @@
 try:
     python_dependencies.check_requirements()
 except python_dependencies.DependencyException as e:
-    sys.stderr.writelines(e.message)
+    sys.stderr.writelines(
+        e.message  # noqa: B306, DependencyException.message is a property
```
clokep, line 28: Should we just be using `str(e)` in each of these spots as the [bugbear docs](https://github.com/PyCQA/flake8-bugbear#python-3-compatibility-warnings) suggest?

tests/unittest.py
```diff
@@ -140,7 +140,7 @@ def assertObjectHasAttributes(self, attrs, obj):
             try:
                 self.assertEquals(attrs[key], getattr(obj, key))
             except AssertionError as e:
-                raise (type(e))(e.message + " for '.%s'" % key)
+                raise (type(e))("Assert error for '.{}':".format(key)) from e
```
clokep, line 143: Why was this one updated to not use `e.message`? It looks like this is going to lose information, or are you hoping that by doing the `from e` it will be kept?

synapse/app/__init__.py
ShadowJonathan: no, because `DependencyException.message` is a property that returns a custom value, if we'd wanna do that, we need to override `__str__` on `DependencyException`

tests/unittest.py
ShadowJonathan: `e.message` is deprecated and shouldn't be used (which is kinda the point of `B306`), with `from e`, it retains that message still in the trace back (under "caused by"), all this intermediate exception is doing is insert the `key` variable for extra information.

clokep: From your comment above I think the answer is that this is not a `DependencyException`, so `e.message` is actually deprecated.

synapse/app/__init__.py
ShadowJonathan: Hmmmm, thanks for that mention, ive subbed to it.

But in the meantime, would this be accepted? (`noqa` on this, with the reasoning behind it as well)

clokep (approved):
I don't love having to mark all of those as noqa, but I don't have a better solution. (We could turn off that check, but that doesn't seem ideal either.)

PR #9503, opened 2021-02-26
Re-run rejected metadata background update.

erikjohnston:
It landed in schema version 58 after 59 had been created, causing some
servers to not run it. The main effect of was that not all rooms had
their chain cover calculated correctly. After the BG updates complete
the chain covers will get fixed when a new state event in the affected
rooms is received.


synapse/storage/databases/main/schema/delta/59/09rejected_events_metadata.sql
```diff
@@ -13,5 +13,8 @@
  * limitations under the License.
  */
 
+-- This originally was in 58/, but landed after 59/ was created, and so some
+-- servers running develop didn't run this delta. Running it again should be
+-- safe.
 INSERT INTO background_updates (ordering, update_name, progress_json) VALUES
   (5828, 'rejected_events_metadata', '{}');
```
clokep, line 20: Does this need a new ordering?

clokep:
Also I can't really tell from the github UI -- is this a copy or a move of that file? 😕

synapse/storage/databases/main/schema/delta/59/09rejected_events_metadata.sql
erikjohnston: I don't think so, since we still want it to be run at the same time as before? We also have a BG update that relies on this.

erikjohnston:
> Also I can't really tell from the github UI -- is this a copy or a move of that file?

It's a move, so that new servers don't run both delta files.

clokep:
I'm pretty sure this should be fine. Will this cause any issues for servers which *have* run the background update (will the insert fail in the delta if it already exists? What if a server is still chunking through this background update, will it blow away progress?)

erikjohnston:
> I'm pretty sure this should be fine. Will this cause any issues for servers which _have_ run the background update (will the insert fail in the delta if it already exists? What if a server is still chunking through this background update, will it blow away progress?)

The background update should do the right thing if re-run (as it needs to handle the case where new rejected events are inserted correctly into the metadata tables). 

This will break if the previous background update is still in progress, as there's a unique index on `background_updates`, will fix.

clokep (approved):

PR #9508, opened 2021-02-27
Update reverse proxy to add OpenBSD relayd example configuration.

tlvb:
### Pull Request Checklist

*My changes are branched out from develop into doc-reverseproxy-relayd and I'm requesting a merge into develop, I assume this is what is meant.*
*Signed-off-by: Leo Bärring <leo.barring@protonmail.com>**
*N/A, no code has been updated.*


docs/reverse_proxy.md
```diff
@@ -160,6 +161,52 @@ backend matrix
...
+
+```
+table <webserver>    { 127.0.0.1 }
+table <matrixserver> { 127.0.0.1 }
+
+http protocol "https" {
+    tls { no tlsv1.0, ciphers "HIGH" }
+    tls keypair "example.com"
+    match header set "X-Forwarded-For"   value "$REMOTE_ADDR"
+    match header set "X-Forwarded-Proto" value "https"
```
anoadragon453, line 174: We [recently discovered](https://github.com/matrix-org/synapse/pull/9512) that nginx modified pass the HTTP `Host` header by default and needed to explicitly tell it to pass it through. Do you know whether relayd would pass the `Host` header to Synapse unaffected or modify/block it at all??

anoadragon453 (commented):
I am by no means an expert in `relayd`, but looking this over it looks like it would do the right thing.

docs/reverse_proxy.md
tlvb: I'm no expert either, but I ran a small test using the suggested config, and with netcat as the "server" that is forwarded to, to capture the request, and e.g. `curl https://example.com:18448/_matrix/hello` produces
```
GET /_matrix/hello HTTP/1.1
Accept: */*
Host: example.com:18448
User-Agent: curl/7.75.0
```
(port slightly altered in test config as otherwise I would get printouts from elsewhere trying to federate with my server)

And for the regular https port (`curl https://example.com`):
```
GET / HTTP/1.1
Accept: */*
Host: tlvb.se
User-Agent: curl/7.75.0
X-Forwarded-For: 256.92.34.25
X-Forwarded-Proto: https
```

So the host does not seem to be modified here.

anoadragon453: Indeed, that looks like what I would expect.

anoadragon453 (approved):
lgtm then, thanks!

PR #9510, opened 2021-02-28
Record the SSO Auth Provider in the login token

richvdh:
This great big stack of commits is a a whole load of hoop-jumping to make it easier to store additional values in login tokens, and then to actually store the SSO Identity Provider in the login token.  (Making use of that data will follow in a subsequent PR.)

The commits should be reviewable independently.


synapse/util/macaroons.py
```diff
@@ -16,14 +16,18 @@
...
 import pymacaroons
+from pymacaroons.exceptions import MacaroonVerificationFailedException
 
 
 def get_value_from_macaroon(macaroon: pymacaroons.Macaroon, key: str) -> str:
     """Extracts a caveat value from a macaroon token.
 
+    Checks that there is exactly one caveat of the form "key = <val>" in the macaroon,
+    and returns the extracted value.
+
```
richvdh, line 57: This change is being introduced mostly in the interests of safety. We've been burnt badly by this in the past (https://matrix.org/blog/2016/07/08/critical-security-vulnerability-in-synapse-0-12-to-0-16-1-inclusive), so some double-checking feels like it is in order.

synapse/handlers/auth.py
```diff
@@ -1578,13 +1584,20 @@ def generate_access_token(
...
     ) -> str:
         macaroon = self._generate_base_macaroon(user_id)
         macaroon.add_first_party_caveat("type = login")
         now = self.hs.get_clock().time_msec()
         expiry = now + duration_in_ms
         macaroon.add_first_party_caveat("time < %d" % (expiry,))
+        if auth_provider_id is not None:
+            macaroon.add_first_party_caveat(
+                "auth_provider_id = %s" % (auth_provider_id,)
+            )
```
erikjohnston, line 1600: A macaroon that doesn't have an `auth_provider_id` caveat allows clients to add any `auth_provider_id` themselves. I think we want to add a `auth_provider_id = None` (or equivalent) when `auth_provider_id` is None?

erikjohnston (commented):
I'm a bit nervous of storing metadata in macaroons, as its going to be easy to accidentally allow clients to inject arbitrary values for them. Ideally we'd rip out the macaroons and replace it with a scheme that doesn't allow mutation, though that seems like quite a bit of faff (and the spec mentions they're meant to be macaroons?).

synapse/handlers/auth.py
dkasak: What is the advantage of storing this information in a macaroon instead of recording it internally in the database?

richvdh: > A macaroon that doesn't have an auth_provider_id caveat allows clients to add any auth_provider_id themselves. I think we want to add a auth_provider_id = None (or equivalent) when auth_provider_id is None?

we were actually setting it to something not-None on all code paths. I've updated the method signature to enforce this.

richvdh: > What is the advantage of storing this information in a macaroon instead of recording it internally in the database?

it's just easier! To store it in the database, we'd have to add a table, do a bunch of sql to insert and select rows, and do a bunch more sql to delete rows once they expire. Not having to do the database queries seems more efficient too.

richvdh:
> I'm a bit nervous of storing metadata in macaroons, as its going to be easy to accidentally allow clients to inject arbitrary values for them. Ideally we'd rip out the macaroons and replace it with a scheme that doesn't allow mutation, though that seems like quite a bit of faff (and the spec mentions they're meant to be macaroons?).

Well, quite. Technically changing away from macaroons here would require an MSC, though for the same reasons as with access_tokens I'm not sure the change can make a material difference to any clients.

I considered replacing the macaroons with some other signed identifier, but didn't much fancy the faff.  I also considered replacing the `user_id` field in the *existing* macaroons with some sort of structured data to store a tuple, but that felt *terrible*.

erikjohnston (approved):
Other than the note below, LGTM.

synapse/handlers/auth.py
erikjohnston: > it's just easier! 

/me looks at the size of this PR :p

richvdh: :-p

This PR is mostly moving existing stuff about. Storing it in the DB would be a bunch of new code.

synapse/util/macaroons.py
```diff
@@ -0,0 +1,114 @@
...
+@overload
+def get_value_from_macaroon(macaroon: pymacaroons.Macaroon, key: str) -> str:
+    ...
+
+
+@overload
+def get_value_from_macaroon(
+    macaroon: pymacaroons.Macaroon, key: str, default: _TV
+) -> Union[str, _TV]:
+    ...
```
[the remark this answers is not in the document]
richvdh: yes, good point. Of course doing so also makes me realise that the OIDC code was a bit dodgy here in that it wasn't always setting a value for `ui_auth_session_id`. I don't think this was exploitable (I don't think that claiming your OIDC auth is for someone else's UIA session helps you, since the userids won't match), but it's definitely a good thing to tighten up.

So, can has another review pls? :innocent:

erikjohnston (approved):

PR #9512, opened 2021-02-28
Update nginx reverse-proxy docs

richvdh:
Turns out nginx overwrites the Host header by default.

### DIFF
--- a/docs/reverse_proxy.md
+++ b/docs/reverse_proxy.md
@@ -53,6 +53,8 @@ server {
         proxy_pass http://localhost:8008;
         proxy_set_header X-Forwarded-For $remote_addr;
         proxy_set_header X-Forwarded-Proto $scheme;
+        proxy_set_header Host $host;
+
         # Nginx by default only allows file uploads up to 1M in size
         # Increase client_max_body_size to match max_upload_size defined in homeserver.yaml
         client_max_body_size 50M;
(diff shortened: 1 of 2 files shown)


docs/reverse_proxy.md
```diff
@@ -53,6 +53,8 @@ server {
         proxy_pass http://localhost:8008;
         proxy_set_header X-Forwarded-For $remote_addr;
         proxy_set_header X-Forwarded-Proto $scheme;
+        proxy_set_header Host $host;
```
anoadragon453, line 56: We might consider using [$http_host](https://nginx.org/en/docs/http/ngx_http_core_module.html#var_http_) instead, which does some normalisation on the value.

richvdh: > which does some normalisation on the value.

where do you see that?

I think `$host` matches what we want here.

anoadragon453: In the docs I linked:

> the last part of a variable name is the field name converted to lower case with dashes replaced by underscores 

Though perhaps this doesn't matter for what we use the Host header for/we normalise in Python already.

richvdh: I don't think those words mean what you think they mean :)

> the last part of a variable **name** is the field [ie, header] **name**

name, not value :)

anoadragon453: Oh, I thought it was just informing you that `name` should be replaced by other header names on the page. The is a subtle but important distinction.

OK, indeed not super helpful then :)

anoadragon453 (approved):
This is why I've switched to caddy,

Issue #9513, opened 2021-03-01
mypy failing on Twisted 21.2.0


clokep:
The mypy check is failing on 21.2.0, which adds significantly more type hints internal to Twisted.

There's currently 196 errors, but hopefully it isn't too bad to clean them up!

clokep:
Filed Shoobx/mypy-zope#34 about an issue I'm having...

clokep:
> Filed [Shoobx/mypy-zope#34](https://github.com/Shoobx/mypy-zope/issues/34) about an issue I'm having...

It seems that mypy-zope has an issue with using the mypy cache, not sure if there will be a solution, but for now runnig `rm -r .mypy_cache` removes some spurious errors.

clokep:
This is all "fixed" now, minus the issue with the mypy cache. I'm not sure what to do about this. We can:

* Leave it and ignore the issue.
* Add a `rm .mypy_cache` to the `lint.sh` script.

I don't have other ideas. 😄

callahad:
Decision:

1. Apply the `rm` hack to `lint.sh` as a stopgap
2. (Briefly) look into upstream code to scope how difficult it would be to supply a patch

PR #9515, opened 2021-03-01
Use the proper Request in type hints.

clokep:
Twisted comes with `twisted.web.http.Request` and `twisted.web.server.Request`, we were using the former in some situations when we should always be using the latter.

This is work towards #9513, it also pins Twisted for the mypy check in the meantime, so that we can continue working in parallel to this.

This reduces the mypy errors from 196 to 129.

### DIFF
--- a/synapse/handlers/auth.py
+++ b/synapse/handlers/auth.py
@@ -36,7 +36,7 @@ import attr
 import bcrypt
 import pymacaroons
 
-from twisted.web.http import Request
+from twisted.web.server import Request
 
 from synapse.api.constants import LoginType
 from synapse.api.errors import (
@@ -481,7 +481,7 @@ class AuthHandler(BaseHandler):
             sid = authdict["session"]
 
         # Convert the URI and method to strings.
-        uri = request.uri.decode("utf-8")
+        uri = request.uri.decode("utf-8")  # type: ignore
         method = request.method.decode("utf-8")
 
         # If there's no session ID, create a new session.
--- a/synapse/replication/http/membership.py
+++ b/synapse/replication/http/membership.py
@@ -18,6 +18,7 @@ from typing import TYPE_CHECKING, List, Optional, Tuple
 from twisted.web.server import Request
 
 from synapse.http.servlet import parse_json_object_from_request
+from synapse.http.site import SynapseRequest
 from synapse.replication.http._base import ReplicationEndpoint
 from synapse.types import JsonDict, Requester, UserID
 from synapse.util.distributor import user_left_room
@@ -78,12 +79,15 @@ class ReplicationRemoteJoinRestServlet(ReplicationEndpoint):
         }
 
     async def _handle_request(  # type: ignore
-        self, request: Request, room_id: str, user_id: str
+        self, request: SynapseRequest, room_id: str, user_id: str
     ) -> Tuple[int, JsonDict]:
         content = parse_json_object_from_request(request)
 
         remote_room_hosts = content["remote_room_hosts"]
         event_content = content["content"]
+
+        requester = Requester.deserialize(self.store, content["requester"])
+        request.requester = requester
 
         logger.info("remote_join: %s into room: %s", user_id, room_id)
 
@@ -143,7 +147,7 @@ class ReplicationRemoteRejectInviteRestServlet(ReplicationEndpoint):
         }
 
     async def _handle_request(  # type: ignore
-        self, request: Request, invite_event_id: str
+        self, request: SynapseRequest, invite_event_id: str
     ) -> Tuple[int, JsonDict]:
         content = parse_json_object_from_request(request)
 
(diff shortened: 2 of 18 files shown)


synapse/handlers/auth.py
```diff
@@ -481,7 +481,7 @@ async def check_ui_auth(
             sid = authdict["session"]
 
         # Convert the URI and method to strings.
-        uri = request.uri.decode("utf-8")
+        uri = request.uri.decode("utf-8")  # type: ignore
```
clokep, line 484: The type of `request.uri` is incorrectly specified in Twisted.

synapse/replication/http/membership.py
```diff
@@ -85,10 +85,6 @@ async def _handle_request(  # type: ignore
         remote_room_hosts = content["remote_room_hosts"]
         event_content = content["content"]
 
-        requester = Requester.deserialize(self.store, content["requester"])
-
-        request.requester = requester
```
clokep, line 90: Here (and below) the `request` is never used after this line, so I don't think this is doing anything.

synapse/rest/media/v1/_base.py
```diff
@@ -49,18 +49,20 @@
...
     try:
+        # The type on postpath seems incorrect in Twisted.
+        postpath = request.postpath  # type: List[bytes]  # type: ignore
+        assert postpath
+
         # This allows users to append e.g. /test.png to the URL. Useful for
         # clients that parse the URL to see content type.
-        server_name, media_id = request.postpath[:2]
-
-        if isinstance(server_name, bytes):
```
clokep, line 56: This is *always* bytes so I don't think we need this if-statement.

synapse/rest/media/v1/upload_resource.py
```diff
@@ -88,8 +90,9 @@ async def _async_render_POST(self, request: Request) -> None:
         # TODO(markjh): parse content-dispostion
 
         try:
+            content = request.content  # type: IO  # type: ignore
```
clokep, line 93: The type of this is undefined right now in Twisted (I think)?

synapse/rest/media/v1/upload_resource.py
```diff
@@ -79,7 +79,9 @@ async def _async_render_POST(self, request: Request) -> None:
         headers = request.requestHeaders
 
         if headers.hasHeader(b"Content-Type"):
-            media_type = headers.getRawHeaders(b"Content-Type")[0].decode("ascii")
+            content_type_headers = headers.getRawHeaders(b"Content-Type")
+            assert content_type_headers  # for mypy
```
clokep, line 83: We know that the header will exist due to the above check, but mypy doesn't...

synapse/handlers/auth.py
clokep: Well it isn't "incorrect", but the placeholder is a string and the value is bytes, which is annoying.

ShadowJonathan: Maybe make notes on the places where this happens, and submit that issue on `twisted/twisted`?

clokep: I'm already planning to.

synapse/rest/media/v1/upload_resource.py
ShadowJonathan: Why the `type: ignore` after the `IO` type definition?

clokep: It overrides the definition previously calculated.

synapse/replication/http/membership.py
callahad: How confident are we? If we're getting `request` as an input parameter and than mutating it as a side effect, might that have issues elsewhere?

synapse/rest/media/v1/_base.py
```diff
@@ -49,18 +49,20 @@
 
 def parse_media_id(request: Request) -> Tuple[str, str, Optional[str]]:
     try:
+        # The type on postpath seems incorrect in Twisted.
```
callahad, line 52: Maybe add the Twisted version to this comment, so it's easier to check if it's still relevant when we stumble across this comment in a year

ShadowJonathan: (+ if you got a issue # for all of those problems on `twisted/twisted`, maybe add then to the ones currently causing trouble in mypy, or a `fixme` at least)

clokep:
```suggestion
        # The type on postpath seems incorrect in Twisted 21.2.0.
```

synapse/replication/http/membership.py
clokep: I'm not sure what you're asking. Are you asking if we're confident in this change or if the code before this change might have been masking other bugs or something?

Note that this is the replication API, which I don't think has any concept of requester? The `Requester.deserialize` doesn't do much except generate an object from a string, so it shouldn't have any real side-effects. Maybe this is here to change logging?

@erikjohnston do you know what the proper thing to do here is?

erikjohnston: Yeah, I think is so that the requester is logged, rather than the default `{None}`. We should probably make it happen automatically, somehow, rather than having it only work for a subset of the replication APIs.

erikjohnston (approved):

synapse/replication/http/membership.py
clokep: Ah, OK. The weird thing is that that implies this is actually a `SynapseRequest` then, not a `Request`. I'll see if I can rework this bit.

PR #9516, opened 2021-03-01
Fix SQL delta file taking a long time to run

### LINKED ISSUE #9504: 08delete_pushers_for_deactivated_accounts.sql takes minutes
We should convert it to a bg update.

erikjohnston:
Fixes #9504, added in #9285

### DIFF
--- a/changelog.d/9516.bugfix
+++ b/changelog.d/9516.bugfix
@@ -1,0 +1,1 @@
+Fix a bug where users' pushers were not all deleted when they deactivated their account.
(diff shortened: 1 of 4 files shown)


clokep:
@erikjohnston Can we merge in develop into this?

clokep (approved):
I think this looks OK. If someone had previously deployed develop (successfully) then it should be OK that this gets skipped since the previous migration ran OK?

erikjohnston:
> I think this looks OK. If someone had previously deployed develop (successfully) then it should be OK that this gets skipped since the previous migration ran OK?

Yes, exactly

PR #9517, opened 2021-03-01
Complete Apache instructions

crees:
Do I still need a CHANGELOG entry for a docs fix?

### DIFF
--- a/changelog.d/9517.doc
+++ b/changelog.d/9517.doc
@@ -1,0 +1,1 @@
+Complete instructions for Apache proxying
--- a/docs/reverse_proxy.md
+++ b/docs/reverse_proxy.md
@@ -98,10 +98,16 @@ example.com:8448 {
 
 ### Apache
 
+Enable `mod_proxy`, `mod_proxy_http`, `mod_proxy_balancer`, `mod_headers`
+and `mod_lbmethod_byrequests`.
+
 ```
 <VirtualHost *:443>
     SSLEngine on
-    ServerName matrix.example.com;
+    ServerName matrix.example.com:443
+
+    SSLCertificateFile "/path/to/certificate/file"
+    SSLCertificateKeyFile "/path/to/certificate/keyfile"
 
     RequestHeader set "X-Forwarded-Proto" expr=%{REQUEST_SCHEME}
     AllowEncodedSlashes NoDecode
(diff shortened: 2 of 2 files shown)


clokep:
> Do I still need a CHANGELOG entry for a docs fix?

They're preferred, yes. It helps people realize the docs were updated when taking a look at the release notes.

crees:
Changelog entry done.

docs/reverse_proxy.md
```diff
@@ -111,9 +117,13 @@ example.com:8448 {
     ProxyPassReverse /_synapse/client http://127.0.0.1:8008/_synapse/client
 </VirtualHost>
 
+Listen 8448
```
anoadragon453, line 120: Just for my own curiosity, this is indeed required as per https://httpd.apache.org/docs/2.4/bind.html.

> To implement a VirtualHost, the server must first be told to listen to the address and port to be used. Then a <VirtualHost> section should be created for the specified address and port to set the behavior of this virtual host. Note that if the <VirtualHost> is set for an address and port that the server is not listening to, it cannot be accessed.

docs/reverse_proxy.md
```diff
@@ -98,10 +98,16 @@ example.com:8448 {
 
 ### Apache
 
+Enable `mod_proxy`, `mod_proxy_http`, `mod_proxy_balancer`, `mod_headers`
+and `mod_lbmethod_byrequests`.
+
 ```
 <VirtualHost *:443>
     SSLEngine on
-    ServerName matrix.example.com;
```
anoadragon453, line 104: Thanks for catching, `;` are indeed not a thing in Apache configs. And indeed https://httpd.apache.org/docs/2.4/mod/core.html#servername mentions to include a port for `ServerName` if possible:

> For optimal reliability and predictability, you should specify an explicit hostname and port using the ServerName directive.

docs/reverse_proxy.md
```diff
@@ -98,10 +98,16 @@ example.com:8448 {
 
 ### Apache
 
+Enable `mod_proxy`, `mod_proxy_http`, `mod_proxy_balancer`, `mod_headers`
+and `mod_lbmethod_byrequests`.
```
anoadragon453, line 102: `mod_proxy_balancer` and `mod_lbmethod_byrequests` are likely only required if you require load balancing, which this example doesn't specify?

changelog.d/9517.doc
```diff
@@ -0,0 +1 @@
+Complete instructions for Apache proxying
```
anoadragon453, line 1: You will need to add punctuation (just a period) to the end of this changelog entry.

anoadragon453 (changes requested):
A minor change and a question, but otherwise lgtm. Learned a bit about Apache in the process too :slightly_smiling_face:

anoadragon453:
Hey @crees, are you still up for helping drive this PR home?

richvdh:
This seems to only need trivial changes, so I'll try and bring it up to date.

richvdh:
actually it seems to have been superceded by #9696, so I'm going to close it.

PR #9518, opened 2021-03-01
Fix additional type hints from Twisted upgrade

clokep:
Part of #9513, this is a grab bag of different improvements, but should be reviewable commit-by-commit. I think most of them are self-explanatory, but a couple are funkier. I'll try to mark the ones I think are questionable.

### DIFF
--- a/synapse/http/federation/matrix_federation_agent.py
+++ b/synapse/http/federation/matrix_federation_agent.py
@@ -14,7 +14,7 @@
 # limitations under the License.
 import logging
 import urllib.parse
-from typing import List, Optional
+from typing import Any, Generator, List, Optional
 
 from netaddr import AddrFormatError, IPAddress, IPSet
 from zope.interface import implementer
@@ -116,7 +116,7 @@ class MatrixFederationAgent:
         uri: bytes,
         headers: Optional[Headers] = None,
         bodyProducer: Optional[IBodyProducer] = None,
-    ) -> defer.Deferred:
+    ) -> Generator[defer.Deferred, Any, defer.Deferred]:
         """
         Args:
             method: HTTP method: GET/POST/etc
--- a/synapse/http/federation/matrix_federation_agent.py
+++ b/synapse/http/federation/matrix_federation_agent.py
@@ -177,17 +177,17 @@ class MatrixFederationAgent:
         # We need to make sure the host header is set to the netloc of the
         # server and that a user-agent is provided.
         if headers is None:
-            headers = Headers()
+            request_headers = Headers()
         else:
-            headers = headers.copy()
+            request_headers = headers.copy()
 
-        if not headers.hasHeader(b"host"):
-            headers.addRawHeader(b"host", parsed_uri.netloc)
-        if not headers.hasHeader(b"user-agent"):
-            headers.addRawHeader(b"user-agent", self.user_agent)
+        if not request_headers.hasHeader(b"host"):
+            request_headers.addRawHeader(b"host", parsed_uri.netloc)
+        if not request_headers.hasHeader(b"user-agent"):
+            request_headers.addRawHeader(b"user-agent", self.user_agent)
 
         res = yield make_deferred_yieldable(
-            self._agent.request(method, uri, headers, bodyProducer)
+            self._agent.request(method, uri, request_headers, bodyProducer)
         )
 
         return res
--- a/synapse/http/server.py
+++ b/synapse/http/server.py
@@ -21,6 +21,7 @@ import logging
 import types
 import urllib
 from http import HTTPStatus
+from inspect import isawaitable
 from io import BytesIO
 from typing import (
     Any,
@@ -30,6 +31,7 @@ from typing import (
     Iterable,
     Iterator,
     List,
+    Optional,
     Pattern,
     Tuple,
     Union,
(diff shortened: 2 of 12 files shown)


synapse/http/server.py
```diff
@@ -551,7 +558,7 @@ def __init__(
         request: Request,
         iterator: Iterator[bytes],
     ):
-        self._request = request
+        self._request = request  # type: Optional[Request]
```
clokep, line 561: We set it to `None` when we stop producing so we need to consider it optional here.

synapse/server.py
```diff
@@ -403,7 +404,7 @@ def get_room_shutdown_handler(self) -> RoomShutdownHandler:
         return RoomShutdownHandler(self)
 
     @cache_in_self
-    def get_sendmail(self) -> sendmail:
+    def get_sendmail(self) -> Callable[..., defer.Deferred]:
```
clokep, line 407: I started adding a better type signature here, but `sendmail` is fairly complicated and can take bytes and str as almost every argument.

tests/rest/client/v1/test_login.py
```diff
@@ -522,7 +522,9 @@ def test_multi_sso_redirect_to_cas(self):
             shorthand=False,
         )
         self.assertEqual(channel.code, 302, channel.result)
-        cas_uri = channel.headers.getRawHeaders("Location")[0]
+        location_headers = channel.headers.getRawHeaders("Location")
+        assert location_headers
```
clokep, line 526: Weirdly it seems that mypy doesn't understand `self.assertTrue(location_headers)`. 🤷 Not sure there's something better we can do for this (and similar instances).

synapse/http/federation/matrix_federation_agent.py
```diff
@@ -116,7 +116,7 @@ def request(
         uri: bytes,
         headers: Optional[Headers] = None,
         bodyProducer: Optional[IBodyProducer] = None,
-    ) -> defer.Deferred:
+    ) -> Generator[defer.Deferred, Any, defer.Deferred]:
```
clokep, line 119: We haven't switched this to `async` since `request` is part of the `IAgent` interface.

synapse/http/server.py
```diff
@@ -79,10 +81,12 @@ def return_json_error(f: failure.Failure, request: SynapseRequest) -> None:
     """Sends a JSON error response to clients."""
 
     if f.check(SynapseError):
```
anoadragon453, line 83: Should we change this to `isinstance(f.value, SynapseError)` perhaps?

synapse/http/site.py
```diff
@@ -96,25 +96,34 @@ def __repr__(self):
     def get_request_id(self):
         return "%s-%i" % (self.get_method(), self.request_seq)
 
-    def get_redacted_uri(self):
-        uri = self.uri
+    def get_redacted_uri(self) -> str:
+        """Gets the redacted URI associated with the request (or placeholder if not
+        method has yet been received).
```
anoadragon453, line 101:
```suggestion
        """Gets the redacted URI associated with the request (or placeholder if method
        has not yet been received).
```

instead perhaps?

synapse/http/site.py
```diff
@@ -96,25 +96,34 @@ def __repr__(self):
...
+
+        Note: This is necessary as the placeholder value in twisted is str
+        rather than bytes, so we need to sanitise `self.uri`.
+
+        Returns:
+            The redacted URI as as string.
```
anoadragon453, line 107:
```suggestion
            The redacted URI as a string.
```

synapse/http/site.py
```diff
@@ -96,25 +96,34 @@ def __repr__(self):
...
+    def get_method(self) -> str:
         """Gets the method associated with the request (or placeholder if not
         method has yet been received).
 
         Note: This is necessary as the placeholder value in twisted is str
         rather than bytes, so we need to sanitise `self.method`.
 
         Returns:
-            str
+            The request method as as string.
```
anoadragon453, line 122:
```suggestion
            The request method as a string.
```

synapse/http/site.py
```diff
@@ -96,25 +96,34 @@ def __repr__(self):
...
+        uri = self.uri  # type: Union[bytes, str]
         if isinstance(uri, bytes):
-            uri = self.uri.decode("ascii", errors="replace")
+            uri = uri.decode("ascii", errors="replace")
         return redact_uri(uri)
 
-    def get_method(self):
+    def get_method(self) -> str:
         """Gets the method associated with the request (or placeholder if not
         method has yet been received).
```
anoadragon453, line 116:
```suggestion
        """Gets the method associated with the request (or placeholder if method
        has not yet been received).
```

Oh, it was a c/p. Well same here too then.

tests/rest/client/v1/test_login.py
anoadragon453: Does it understand `self.assertIsNotNone(location_headers)` instead?

anoadragon453 (commented):
Thanks for wading through this! A couple questions and suggestions.

synapse/http/server.py
clokep: I don't think so. This is the proper way to check against twisted failures?

tests/rest/client/v1/test_login.py
clokep: I tried a handful of variations of this (`assertTrue`, `assertIsInstance`, `assertIsNotNone`, etc.) I think mypy just doesn't know what these are. The only references I can find to them in mypy are in the stubs file.

I was wondering if this is because were using the twisted unit test classes (although this does eventually inherit from the Python unit testing framework). 🤷

clokep: This seems to be python/mypy#5088, which points to python/mypy#4063.

anoadragon453: I see. Oh well, `assert` it is for now then.

synapse/http/server.py
anoadragon453: True, probably easier (and more likely) for us to remove a `# type: ignore` in the future than switch away from an `isinstance`.

anoadragon453 (approved):

synapse/http/server.py
clokep: I think my thought is that if we want to make that change we should do is as a discrete step, not bury it in this PR!

PR #9520, opened 2021-03-01
Add type hints to purge room and server notice admin API

dklimpel:
### Pull Request Checklist

### DIFF
--- a/synapse/rest/admin/server_notice_servlet.py
+++ b/synapse/rest/admin/server_notice_servlet.py
@@ -69,9 +69,7 @@ class SendServerNoticeServlet(RestServlet):
             self.__class__.__name__,
         )
 
-    async def on_POST(
-        self, request: SynapseRequest, txn_id: str = None
-    ) -> Tuple[int, JsonDict]:
+    async def on_POST(self, request: SynapseRequest) -> Tuple[int, JsonDict]:
         await assert_requester_is_admin(self.auth, request)
         body = parse_json_object_from_request(request)
         assert_params_in_dict(body, ("user_id", "content"))
--- a/synapse/rest/admin/server_notice_servlet.py
+++ b/synapse/rest/admin/server_notice_servlet.py
@@ -33,14 +33,7 @@ if TYPE_CHECKING:
 
 
 class SendServerNoticeServlet(RestServlet):
-    """This servlet needs some unit tests and updates:
-    The `txn_id` should be passed into `send_notice` method and then passed through
-    into `create_and_send_nonmember_event` (see RoomSendEventRestServlet for the proper
-    way to do this).
-    Also it needs an update to avoid creating duplicate server-notice rooms.
-    https://github.com/matrix-org/synapse/pull/9520#discussion_r586435068
-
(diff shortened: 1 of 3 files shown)


synapse/rest/admin/server_notice_servlet.py
```diff
@@ -66,7 +69,9 @@ def register(self, json_resource):
             self.__class__.__name__,
         )
 
-    async def on_POST(self, request, txn_id=None):
+    async def on_POST(
+        self, request: SynapseRequest, txn_id: str = None
```
clokep, line 73: Please make `txn_id` into `Optional[str]` since it can also be `None`. (This only applies to the `POST` endpoint, it seems that the `PUT` endpoint requires it?)

clokep: Right, I was thinking that `txn_id` was optionally accepted here, but looking at how it is registered above that seems to not be the case! We should just remove the `txn_id` parameter in that case.

clokep: Looking at this again I think that something confusing is happening. The `on_PUT` handler calls [`fetch_or_execute_request`](https://github.com/matrix-org/synapse/blob/c619253db80c8d1c606dc40756dd3c9e3a55a9fb/synapse/rest/client/transactions.py#L57-L66), when then calls [`fetch_or_execute`](https://github.com/matrix-org/synapse/blob/c619253db80c8d1c606dc40756dd3c9e3a55a9fb/synapse/rest/client/transactions.py#L68-L101). Both of these pass along the args and kwargs passed to it. So I think calling this API with a `PUT` right now would fail (as it would eventually call `on_POST` with too many arguments).

I'm not sure why `txn_id` is being passed through though since it is unused?

clokep: We chatted a bit about this in [#synapse-dev:matrix.org](https://matrix.to/#/!XaqDhxuTIlvldquJaV:matrix.org/$Uc5-_9Ws8gfpQl2zSsJAAqcxugfV3C3ozOrTVKohkEs?via=matrix.org&via=vector.modular.im&via=envs.net) and realize that this is highlighting a bug in the servlet!

* The `txn_id` should be added back as an `Optional[str]`.
* The `txn_id` should be passed into `send_notice` method and then passed through into `create_and_send_nonmember_event` (see `RoomSendEventRestServlet` for the proper way to do this).

It was also pointed out that some more magic might be needed to avoid creating duplicate server-notice rooms too, but maybe we can leave that to a follow-up and just add a note about it.

clokep (changes requested):
Per the review comment.

synapse/rest/admin/server_notice_servlet.py
dklimpel: Is a comment the right solution or rather an issue here?

clokep: I think the right solution is to use the correct type `Optional[str]` and file a follow-up issue.

clokep (approved):
LGTM! Thanks for making those changes (and filing the follow-up)!

PR #9521, opened 2021-03-01
Add type hints to user admin API

dklimpel:
### Pull Request Checklist

### DIFF
--- a/synapse/rest/admin/users.py
+++ b/synapse/rest/admin/users.py
@@ -278,9 +278,9 @@ class UserRestServletV2(RestServlet):
                         target_user.to_string()
                     )
 
-            user = await self.admin_handler.get_user(
-                target_user
-            )  # type: JsonDict  # type: ignore
+            user = await self.admin_handler.get_user(target_user)
+            assert user is not None
+
             return 200, user
 
         else:  # create user
@@ -338,11 +338,10 @@ class UserRestServletV2(RestServlet):
                     target_user, requester, body["avatar_url"], True
                 )
 
-            ret = await self.admin_handler.get_user(
-                target_user
-            )  # type: JsonDict  # type: ignore
+            user = await self.admin_handler.get_user(target_user)
+            assert user is not None
 
-            return 201, ret
+            return 201, user
 
 
 class UserRegisterServlet(RestServlet):
--- a/synapse/rest/admin/users.py
+++ b/synapse/rest/admin/users.py
@@ -16,7 +16,7 @@ import hashlib
 import hmac
 import logging
 from http import HTTPStatus
-from typing import TYPE_CHECKING, Dict, Optional, Tuple
+from typing import TYPE_CHECKING, Dict, List, Optional, Tuple
 
 from synapse.api.constants import UserTypes
 from synapse.api.errors import Codes, NotFoundError, SynapseError
@@ -55,7 +55,7 @@ class UsersRestServlet(RestServlet):
 
     async def on_GET(
         self, request: SynapseRequest, user_id: str
-    ) -> Tuple[int, JsonDict]:
+    ) -> Tuple[int, List[JsonDict]]:
         target_user = UserID.from_string(user_id)
         await assert_requester_is_admin(self.auth, request)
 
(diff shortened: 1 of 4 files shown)


synapse/rest/admin/users.py
```diff
@@ -165,7 +167,7 @@ def __init__(self, hs):
         self.registration_handler = hs.get_registration_handler()
         self.pusher_pool = hs.get_pusherpool()
 
-    async def on_GET(self, request, user_id):
+    async def on_GET(self, request: SynapseRequest, user_id) -> Tuple[int, JsonDict]:
```
clokep, line 170:
```suggestion
    async def on_GET(self, request: SynapseRequest, user_id: str) -> Tuple[int, JsonDict]:
```

synapse/rest/admin/users.py
```diff
@@ -272,7 +276,9 @@ async def on_PUT(self, request, user_id):
                         target_user.to_string()
                     )
 
-            user = await self.admin_handler.get_user(target_user)
+            user = await self.admin_handler.get_user(
+                target_user
+            )  # type: JsonDict  # type: ignore
```
clokep, line 281: Why ignore this?

synapse/rest/admin/users.py
```diff
@@ -346,10 +354,10 @@ class UserRegisterServlet(RestServlet):
     PATTERNS = admin_patterns("/register")
     NONCE_TIMEOUT = 60
 
-    def __init__(self, hs):
+    def __init__(self, hs: "HomeServer"):
         self.auth_handler = hs.get_auth_handler()
         self.reactor = hs.get_reactor()
-        self.nonces = {}
+        self.nonces = {}  # type: Dict
```
clokep, line 360: I think this is a mapping of strings to ints? It would be great to specify that as `Dict[str, int]`.

synapse/rest/admin/users.py
```diff
@@ -626,12 +640,14 @@ class SearchUsersRestServlet(RestServlet):
...
-    def __init__(self, hs):
+    def __init__(self, hs: "HomeServer"):
         self.hs = hs
         self.store = hs.get_datastore()
         self.auth = hs.get_auth()
 
-    async def on_GET(self, request, target_user_id):
+    async def on_GET(
+        self, request: SynapseRequest, target_user_id: str
+    ) -> Tuple[int, Optional[JsonDict]]:
```
clokep, line 650: Why is this one optional? I think `search_users` always returns a dict?

synapse/storage/databases/main/__init__.py
```diff
@@ -264,7 +264,7 @@ def _get_active_presence(self, db_conn):
 
         return [UserPresenceState(**row) for row in rows]
 
-    async def get_users(self) -> List[Dict[str, Any]]:
+    async def get_users(self) -> JsonDict:
```
clokep, line 267: I think this still returns a list? So it should be `List[JsonDict]`? Same for the other changes in this file.

clokep (changes requested):
Mostly looks good, but I have a few questions!

synapse/rest/admin/users.py
dklimpel: https://github.com/matrix-org/synapse/blob/aee10768d8a9bf6c3df80859d2c135607aa1bf80/synapse/handlers/admin.py#L60

It is in `get_user` optional, because it can be `NULL` if user does not exist.

https://github.com/matrix-org/synapse/blob/aee10768d8a9bf6c3df80859d2c135607aa1bf80/synapse/rest/admin/users.py#L192-L195
In this case the user always exist an we get a `JsonDict`.

synapse/rest/admin/users.py
dklimpel: IMO it can be `NULL` if search returns an empty result.
https://github.com/matrix-org/synapse/blob/aee10768d8a9bf6c3df80859d2c135607aa1bf80/synapse/storage/databases/main/__init__.py#L356-L372

synapse/storage/databases/main/__init__.py
dklimpel: I have made an update.

synapse/rest/admin/users.py
clokep: I think you want to do `assert user` after this line to let mypy know that it will never be `None`.

synapse/rest/admin/users.py
clokep: Ah! I was looking at the handler code. 👍  This is why adding the type hints is useful!

clokep (approved):
LGTM! Thank you!

PR #9523, opened 2021-03-01
Add logging to ObservableDeferred callbacks

ShadowJonathan:
### Pull Request Checklist

### DIFF
--- a/synapse/util/async_helpers.py
+++ b/synapse/util/async_helpers.py
@@ -76,11 +76,12 @@ class ObservableDeferred:
         def callback(r):
             object.__setattr__(self, "_result", (True, r))
             while self._observers:
+                observer = None
                 try:
-                    # TODO: Handle errors here.
-                    self._observers.pop().callback(r)
-                except Exception:
-                    pass
+                    observer = self._observers.pop()
+                    observer.callback(r)
+                except Exception as e:
+                    logger.warning("%r threw an exception on .callback(%r), ignoring...", observer, r, exc_info=e)
             return r
 
         def errback(f):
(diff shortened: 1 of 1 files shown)


richvdh (approved):
lgtm I think but I'd like another pair of eyes on it

erikjohnston (approved):
I think this makes sense. I do wonder if we want to log them as exceptions though, so things like sentry pick them up and we can fix it?

ShadowJonathan:
> I do wonder if we want to log them as exceptions though, so things like sentry pick them up and we can fix it?

@erikjohnston I have no idea how i can do that, do i need to change `warning` to `exception` or `error`? Where can i see more details about how sentry hooks into synapse, anyway?

erikjohnston:
(It works by the sentry package adding a hook to the python logging library so that it can see all the logs that go past)

ShadowJonathan:
:+1: Then i'll change the `warning` to an `error`, I think that even without a sentry library, it'd make more sense anyways.

clokep:
Re-requested review from @erikjohnston to take a look at the additional changes.

ShadowJonathan:
Oh, yeah, i forgot to mention i changed it to `exception`, because internally it still calls to `error`, it just semantically looked more appropriate for me.


stdlib for `logging`, anyhow;
```py
    def exception(self, msg, *args, exc_info=True, **kwargs):
        """
        Convenience method for logging an ERROR with exception information.
        """
        self.error(msg, *args, exc_info=exc_info, **kwargs)
```

erikjohnston (approved):

Issue #8642, opened 2020-10-24
Consider using asyncioreactor


ShadowJonathan:
### Idea

Use `twisted.internet.asyncioreactor` for synapse's twisted reactor.

### Description

This is part of my idea that synapse should consider moving forward, beyond, or at least *next* to twisted, to also use `asyncio`'s economy of libraries and functions. With the `asyncioreactor`, it's possible to mix `twisted` and `asyncio` functionality with eachother.

It's also possible to use [`uvloop`](https://github.com/MagicStack/uvloop) in addition to this, allowing a big IO performance boost.

Some links:

`asyncioreactor`: https://twistedmatrix.com/documents/current/api/twisted.internet.asyncioreactor.html

`asyncio`/`twisted` intermixing: https://gist.github.com/ldjebran/4febf298232a6fd86871df25d4dc00dd

`Deferred.asFuture()`: https://twistedmatrix.com/documents/current/api/twisted.internet.defer.Deferred.html#asFuture

Twisted and Asyncio (blog post): https://meejah.ca/blog/python3-twisted-and-asyncio

### Considerations

I have two concerns with doing this:

1. Will the existing twisted-based code be stable on a new reactor based out of `asyncio`?

2. Will the reactor impact performance (negatively)?

For 1, I want to run a server along my existing one, only joining some channels, and watch out for any crashes and the like, and/or perform test benches with `asyncioreactor` selected.

For 2, I'd want to know what the best way is for me to collect such statistics and/or how to compare them with a normal server.

erikjohnston:
FWIW I tried using the asyncioreactor the other day and it mostly seemed to work, modulo a couple of test failures. The big difference seems to be that `asyncioreactor.seconds()` doesn't return a unix timestamp, which our `Clock` implementation assumes.

richvdh:
That got fixed recently: https://github.com/twisted/twisted/pull/1237.

@ShadowJonathan we're familiar with asyncioreactor (we use it in sygnal). Feel free to try it out on Synapse and let us know how it goes.

ShadowJonathan:
@richvdh it looks like it's being planned for `20.11`, but it's not out yet: https://github.com/twisted/twisted/commit/be66f18fde8aba632117e0bf685638984377150e (look at the tags in the top row)

ShadowJonathan:
[...]

PR #9528, opened 2021-03-02
Create ISynapseReactor interface which incorporates the necessary reactors

clokep:
This allows passing around the proper interface instead of pretending that `ReactorBase` has the proper methods.

Part of #9513, fixes ~10 issues.

### DIFF
--- a/synapse/handlers/acme.py
+++ b/synapse/handlers/acme.py
@@ -73,7 +73,9 @@ class AcmeHandler:
                 "Listening for ACME requests on %s:%i", host, self.hs.config.acme_port
             )
             try:
-                self.reactor.listenTCP(self.hs.config.acme_port, srv, interface=host)
+                self.reactor.listenTCP(
+                    self.hs.config.acme_port, srv, backlog=50, interface=host
+                )
             except twisted.internet.error.CannotListenError as e:
                 check_bind_error(e, host, bind_addresses)
 
--- a/synapse/replication/tcp/redis.py
+++ b/synapse/replication/tcp/redis.py
@@ -328,6 +328,6 @@ def lazyConnection(
     factory.continueTrying = reconnect
 
     reactor = hs.get_reactor()
-    reactor.connectTCP(host, port, factory, 30)
+    reactor.connectTCP(host, port, factory, timeout=30, bindAddress=None)
 
     return factory.handler
--- a/synapse/types.py
+++ b/synapse/types.py
@@ -35,6 +35,14 @@ from typing import (
 import attr
 from signedjson.key import decode_verify_key_bytes
 from unpaddedbase64 import decode_base64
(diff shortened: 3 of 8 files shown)


synapse/handlers/acme.py
```diff
@@ -73,7 +73,9 @@ async def start_listening(self) -> None:
                 "Listening for ACME requests on %s:%i", host, self.hs.config.acme_port
             )
             try:
-                self.reactor.listenTCP(self.hs.config.acme_port, srv, interface=host)
+                self.reactor.listenTCP(
+                    self.hs.config.acme_port, srv, backlog=50, interface=host
```
clokep, line 77: mypy was complaining about not giving the required `backlog` attribute, 50 matches the default in all the reactor implementations in Twisted.

synapse/replication/tcp/redis.py
```diff
@@ -328,6 +328,6 @@ def lazyConnection(
     factory.continueTrying = reconnect
 
     reactor = hs.get_reactor()
-    reactor.connectTCP(host, port, factory, 30)
+    reactor.connectTCP(host, port, factory, timeout=30, bindAddress=None)
```
clokep, line 331: mypy was complaining about a missing `bindAddress` parameter, this gives the default value.

synapse/types.py
```diff
@@ -67,6 +74,12 @@ class Collection(Iterable[T_co], Container[T_co], Sized):  # type: ignore
 JsonDict = Dict[str, Any]
 
 
+class SynapseReactor(
+    IReactorTCP, IReactorPluggableNameResolver, IReactorTime, IReactorCore
+):
+    """The interfaces necessary for Synapse to function."""
```
ShadowJonathan, line 80: Just for clarification, is this new type meant to be a collection of interfaces that we use as "okay, this type implements these interfaces"? I'm asking because i'm thinking about how this'd potentially interact with #8642 or other reactors.

clokep: It is meant to collect the various reactor interfaces that Synapse expects, e.g. it expects to be able to speak TCP. It shouldn't have any impact on #8642, since if asyncioreactor doesn't implement these interfaces than it won't work!

ShadowJonathan: I've become lukewarm on asyncioreactor anyways, as in my research its proven to be a very thin wrapper around asyncio, with non-adequate integration with it (in my opinion).

(P.S. I cant seem to resolve this comment convo, but i consider it resolved, thanks for answering :+1:)

synapse/replication/tcp/redis.py
clokep: (This is because the interface doesn't define the default, but the implementations do.)

erikjohnston (approved):

PR #9535, opened 2021-03-03
Remove unneeded warning log line

michaelkaye:
This log line is not needed and we seem to be blocked on removing it.


richvdh (changes requested):
matrix-org-hotfixes already contains a fix to the bug you're trying to fix, so this is redundant.

PR #9536, opened 2021-03-03
Make deleting stale pushers a background update

erikjohnston:


richvdh (approved):
seems plausible

changelog.d/9536.bugfix
```diff
@@ -0,0 +1 @@
+Fix deleting pushers when using sharded pushers.
```
richvdh, line 1: is this the same as the newsfile for the PR that introduced the delta file? if not, please could it be?

erikjohnston: It is, sorry should have linked to it in the PR description: #9479

PR #9537, opened 2021-03-03
Fix 'rejected_events_metadata' background update

erikjohnston:
Turns out matrix.org has an event that has duplicate auth events (which really isn't supposed to happen, but here we are). This caused the background update to fail due to `UniqueViolation`.

This was added in v1.28.0, but we only just noticed on matrix.org due to #9503.

Sentry link: https://sentry.matrix.org/sentry/synapse-matrixorg/issues/200948/

### DIFF
--- a/changelog.d/9537.bugfix
+++ b/changelog.d/9537.bugfix
@@ -1,0 +1,1 @@
+Fix rare edge case that caused a background update to fail if the server had rejected an event that had duplicate auth events.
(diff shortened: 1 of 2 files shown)


clokep (approved):
I'm assuming the background update will retry adding this rejected event?

erikjohnston:
> I'm assuming the background update will retry adding this rejected event?

Yup, its currently sitting there repeatedly trying that event, and failing each time :)

PR #9541, opened 2021-03-03
Add a basic test for purging rooms

clokep:
This adds a basic storage layer test for purging rooms, doesn't offer too much more than the ones using the admin API, but I find it quite a bit clearer.

Related to #9498 / #9481.

### DIFF
--- a/tests/storage/test_purge.py
+++ b/tests/storage/test_purge.py
@@ -90,3 +90,28 @@ class PurgeTests(HomeserverTestCase):
         self.get_success(self.store.get_event(second["event_id"]))
         self.get_success(self.store.get_event(third["event_id"]))
         self.get_success(self.store.get_event(last["event_id"]))
+
+    def test_purge_room(self):
+        """
+        Purging a room will delete everything about it.
+        """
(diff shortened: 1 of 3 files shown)


tests/storage/test_purge.py
```diff
@@ -33,63 +31,84 @@ def make_homeserver(self, reactor, clock):
...
+        create_event = self.get_success(
+            state_handler.get_current_state(self.room_id, "m.room.create", "")
+        )
+        self.assertIsNotNone(create_event)
+
+        # Purge everything before this topological token
+        self.get_success(self.storage.purge_events.purge_room(self.room_id))
+
+        # The events aren't found.
+        self.store._invalidate_get_event_cache(create_event.event_id)
```
clokep, line 112: There's a note in `purge_room` how we should be invalidating caches. I can take a look at that but it'll grow the scope a bit.

erikjohnston (approved):

clokep:
Thanks! I'm going to update the changelog entry here since this  missed v1.29.0rc1.

PR #9543, opened 2021-03-03
Fix additional type hints

clokep:
More of #9513.

Should be reviewable commit-by-commit, they're not really related. I'll note anything odd.

This fixes 11 issues. Getting there!


synapse/federation/federation_server.py
```diff
@@ -361,7 +361,7 @@ async def process_pdus_for_room(room_id: str):
                             logger.error(
                                 "Failed to handle PDU %s",
                                 event_id,
-                                exc_info=(f.type, f.value, f.getTracebackObject()),
+                                exc_info=(f.type, f.value, f.getTracebackObject()),  # type: ignore
```
clokep, line 364: mypy things that one of these doesn't match the type signature for `logger.error`, but empirically it works...

> synapse/federation/federation_server.py:364: error: Argument "exc_info" to "error" of "Logger" has incompatible type "Tuple[Optional[Type[BaseException]], Optional[BaseException], Any]"; expected "Union[None, bool, Union[Tuple[type, BaseException, Optional[TracebackType]], Tuple[None, None, None]], BaseException]"  [arg-type]

synapse/logging/context.py
```diff
@@ -669,7 +669,7 @@ def g(*args, **kwargs):
     return g
 
 
-def run_in_background(f, *args, **kwargs):
+def run_in_background(f, *args, **kwargs) -> defer.Deferred:
```
clokep, line 672: I think this isn't 100% true actually since if the result is *not* a `Deferred` than we return it raw, which seems confusing. I think we should likely wrap that in `defer.succeed`?

clokep:
I have now realized that my commit messages here aren't the best. I think the commits aren't too complicated, but please shout if there are questions!

synapse/logging/context.py
erikjohnston: Yeah, probably. I don't think we should lie about our return value here.

synapse/federation/federation_server.py
erikjohnston: @richvdh do you remember why we can't just use `logger.exception` here?

erikjohnston (changes requested):

tests/test_utils/logging_setup.py
```diff
@@ -28,7 +28,7 @@ class ToTwistedHandler(logging.Handler):
     def emit(self, record):
         log_entry = self.format(record)
         log_level = record.levelname.lower().replace("warning", "warn")
-        self.tx_log.emit(
+        self.tx_log.emit(  # type: ignore
```
clokep, line 31: For some reason mypy things that `tx_log` here is `object`, I'm not really sure why though... Anyway, it complains that object has no `emit` method.

synapse/logging/context.py
clokep: I wrapped the raw return value in `defer.succeed`, I suspect it doesn't really make sense to call `run_in_background` with a sync function, but 🤷 This would also break when using `await`, so maybe we should make this an error instead?

tests/test_utils/logging_setup.py
clokep: I suspect this is fixed by twisted/twisted#1541?

erikjohnston (approved):

synapse/federation/federation_server.py
richvdh: er we probably can just `s/error/exception/`, if that helps, though I don't think it will.

or is your question why we pass in an explicit `exc_info` rather than letting `logging` pull it from `sys.exc_info()`? in which case the answer is that `f.getTracebackObject` has magic to build a more useful stacktrace in the face of async execution.  (It's *possible* that it's better nowadays with proper `async/await`, but I suspect it's not.)

clokep: Thanks for the background! I think leaving the ignore here is the best way forward at the moment without getting into a rabbit hole of logging.

PR #9548, opened 2021-03-04
Require AppserviceRegistrationType

Half-Shot:
(apparently we don't have an issue for this one)

Needs a fix on the sytest side for https://github.com/matrix-org/sytest/issues/1018

This change ensures that the appservice registration behaviour follows [the spec](https://spec.matrix.org/unstable/application-service-api/#client-server-api-extensions). We decided to [do this for Dendrite](https://github.com/matrix-org/dendrite/pull/1785), so it made sense to also make a PR for synapse to correct the behaviour.

In terms of who is affected by this:
- matrix-appservice-bridge based bridges <=2.50 are affected, and will need to update
- matrix-bot-sdk bridges are not affected (e.g. mx-puppet-bridge)
- mautrix-python bridges are not affected
- mautrix-go bridges are affected [<= latest](https://github.com/tulir/mautrix-go/commit/2ab811a476a6aad539eed8e613580bc18b603206)


synapse/rest/client/v2_alpha/register.py
```diff
@@ -428,15 +427,20 @@ async def on_POST(self, request):
...
-        if self.auth.has_access_token(request):
-            appservice = self.auth.get_appservice_by_req(request)
-
         # fork off as soon as possible for ASes which have completely
         # different registration flows to normal users
 
         # == Application Service Registration ==
-        if appservice:
+        if body.get("type") == AppserviceRegistrationType:
+            if not self.auth.has_access_token(request):
```
Half-Shot, line 435: Maybe we should check for this access token regardless of type, but throw if the type is wrong to help people diagnose their broken appservices?

Half-Shot: Have done so in 8bd7eb6

clokep:
I think this can't be merged until #9559 has been in a release for a bit. Do we have any ideas on how long we want to keep it deprecated?

synapse/api/constants.py
```diff
@@ -70,6 +70,11 @@ class LoginType:
     DUMMY = "m.login.dummy"
 
 
+# This is not a login type as we cannot login with it, however
+# the spec says that this is a valid registration type for appservices.
+AppserviceRegistrationType = "m.login.application_service"
```
clokep, line 75: Please use a Python style constant: `APP_SERVICE_REGISTRATION_TYPE`.

Half-Shot:
>I think this can't be merged until #9559 has been in a release for a bit. Do we have any ideas on how long we want to keep it deprecated?

All the folks I've reached out to so far have either fixed it same-day, or have already been following the spec. The change required is tiny, so I think two weeks is plenty of time?

anoadragon453:
> > I think this can't be merged until #9559 has been in a release for a bit. Do we have any ideas on how long we want to keep it deprecated?
> 
> All the folks I've reached out to so far have either fixed it same-day, or have already been following the spec. The change required is tiny, so I think two weeks is plenty of time?

Be sure to check with the myriad of application services that customers have :slightly_frowning_face:

clokep:
> > I think this can't be merged until #9559 has been in a release for a bit. Do we have any ideas on how long we want to keep it deprecated?
> 
> All the folks I've reached out to so far have either fixed it same-day, or have already been following the spec. The change required is tiny, so I think two weeks is plenty of time?

I assume this refers to application services authors? Not necessarily people who have deployed those ASes?

anoadragon453:
Uh, that wasn't meant to happen :thinking:

Half-Shot:
@clokep Authors yes, who should probably talk to their users about upgrading too. I think we can assume that it will all happen. I'll have a chat with PS

synapse/api/constants.py
```diff
@@ -70,6 +70,11 @@ class LoginType:
     DUMMY = "m.login.dummy"
 
 
+# This is not a login type as we cannot login with it, however
+# the spec says that this is a valid registration type for appservices.
```
richvdh, line 74: a "registration type" isn't really a concept the spec has.

(and no, it's certainly not a login type).

```suggestion
# This is used in the `type` parameter for /register when called by
# an appservice to register a new user.
```

richvdh: Incidentally, the uses of a magic `type: magic_value` parameter with usage completely at odds with the *normal* use of `/register` is a *terrible* idea, but that's a matter for the spec, not here.

(it should be a separate endpoint imho)

changelog.d/9548.bugfix
```diff
@@ -0,0 +1 @@
+/register now requires a `body.type` value of `m.login.appservice` when registering appservice users.
```
richvdh, line 1: this should probably be a `removal` (see https://github.com/matrix-org/synapse/blob/develop/CONTRIBUTING.md#changelog), with slightly different wording.

richvdh (dismissed):
seems generally good to me. a few nits.

synapse/rest/client/v2_alpha/register.py
```diff
@@ -457,6 +461,11 @@ async def on_POST(self, request):
             )
 
             return 200, result
+        elif self.auth.has_access_token(request):
+            raise SynapseError(
+                400,
+                "A type of m.login.application_service must be provided when registering as an appservice",
+            )
```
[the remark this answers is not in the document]
Half-Shot: I don't think there is a case currently where you can register with an access token without being an appservice?

synapse/api/constants.py
Half-Shot: Agreed, we should have an appservice registration endpoint given they don't even use all the parameters from the response anyway (e.g. they don't care about devices) so perhaps there should be a proposal made for that...

changelog.d/9548.bugfix
Half-Shot: I've made an attempt, but I'm not sure what wording change you are looking for.

changelog.d/9548.removal
```diff
@@ -0,0 +1 @@
+/register now expects a `body.type` value of `m.login.appservice` when registering appservice users, to conform to the spec.
```
anoadragon453, line 1:
```suggestion
Make `/_matrix/client/r0/register` expect a login type of `m.login.application_service` when an Application Service registers a user, to align with [the relevant spec](https://spec.matrix.org/unstable/application-service-api/#server-admin-style-permissions).
```

May be a bit wordy, but hopefully clearer. Also `s/m.login.appservice/m.login.application_service`.

synapse/api/constants.py
```diff
@@ -73,6 +73,11 @@ class LoginType:
     DUMMY = "m.login.dummy"
 
 
+# This is used in the `type` parameter for /register when called by
+# an appservice to register a new user.
+APP_SERVICE_REGISTRATION_TYPE = "m.login.application_service"
```
anoadragon453, line 78: Why wasn't this added to the `LoginType` class above?

tests/test_mau.py
```diff
@@ -232,19 +232,20 @@ def test_tracked_but_not_limited(self):
...
+        }
+
+        if appservice:
+            request_data["type"] = APP_SERVICE_REGISTRATION_TYPE
 
         channel = self.make_request(
             "POST",
             "/register",
-            request_data,
+            json.dumps(request_data),
```
anoadragon453, line 248: Not that you need to change it (as it matches the existing code), but heads up that `make_request` can actually just accept straight dictionaries these days.

synapse/api/constants.py
Half-Shot: See https://github.com/matrix-org/synapse/pull/9548#discussion_r604778903

tests/test_mau.py
Half-Shot: That's cool. Hopefully someone nice will come along and refactor the lot one day :)

Half-Shot: Oh except here I have changed it back  haven't i?

changelog.d/9548.removal
```diff
@@ -0,0 +1 @@
+Make `/_matrix/client/r0/register` expect a login type of `m.login.application_service` when an Application Service registers a user, to align with [the relevant spec](https://spec.matrix.org/unstable/application-service-api/#server-admin-style-permissions).
```
anoadragon453, line 1:
```suggestion
Make `/_matrix/client/r0/register` expect a type of `m.login.application_service` when an Application Service registers a user, to align with [the relevant spec](https://spec.matrix.org/unstable/application-service-api/#server-admin-style-permissions).
```

synapse/api/constants.py
anoadragon453: Oh pfft, sorry to have gone around in circles on that issue then.

tests/test_mau.py
anoadragon453: Mmm, I don't think so. Before it was `request_data = json.dumps(...)`: https://github.com/matrix-org/synapse/pull/9548/files#diff-9f648c7d26b378d1bdae9e1d63e7a4c75d3645634680aa26e44dc959f99aa70dL236

But hooray for updating it anyways :tada:

Half-Shot: changed again

anoadragon453 (commented):
I'd like to see this conversation resolved before I give a :heavy_check_mark:, @richvdh? https://github.com/matrix-org/synapse/pull/9548/files#r604775392

synapse/rest/client/v2_alpha/register.py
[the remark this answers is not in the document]
Half-Shot: The reason for keeping this `elif` around is that previously we used the access token to determine if the request was on the behalf of an appservice. Rather than throwing that away I wanted to throw an explicit error for people running non-compliant appservices (hence the very specific error message).

If we throw this statement away, the token will be ignored and requests that previously successfully created an appservice user may either fail with a somewhat unhelpful error (like "you didn't specify a password"), or may succeed in creating a non-appservice user which would probably do more harm than good.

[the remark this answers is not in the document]
richvdh: the point is more that the error message is ... unclear.

synapse/rest/client/v2_alpha/register.py
```diff
@@ -459,6 +463,11 @@ async def on_POST(self, request):
             )
 
             return 200, result
+        elif self.auth.has_access_token(request):
+            raise SynapseError(
+                400,
+                "A type of m.login.application_service must be provided when registering as an appservice",
```
richvdh, line 469:
```suggestion
                "An access token should not be provided on requests to /register (except if type is m.login.application_service)",
```

synapse/rest/client/v2_alpha/register.py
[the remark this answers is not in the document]
richvdh: suggested a rephrasing.

synapse/rest/client/v2_alpha/register.py
Half-Shot: Perfect, that's much better.

richvdh (approved):
lgtm

PR #9549, opened 2021-03-04
JWT OIDC secrets for Sign in with Apple

### LINKED ISSUE #9212: OpenID Connect `client_secret` should be optional
I'm surprised to see `client_secret` documented as required since it's not really the case in the OIDC protocol even if it's definitely a good practice especially for big public providers which often do require it.

### LINKED ISSUE #9220: 💣 Improve mechanism for generating SSO client secrets for Apple login
"Sign in with Apple ID" requires an oidc client secret which is an encoded JSON Web Token (https://developer.apple.com/documentation/sign_in_with_apple/generate_and_validate_tokens#3262048). It would be good to add support for calculating such secrets.

richvdh:
Apple had to be special. They want a client secret which is generated from an EC key.

Fixes #9220. Also fixes #9212 while I'm here.


clokep:
@richvdh Looks like this is failing tests!

synapse/handlers/oidc_handler.py
```diff
@@ -977,6 +993,81 @@ def _remote_id_from_userinfo(self, userinfo: UserInfo) -> str:
...
+        oauth_client_id: str,
+        oauth_issuer: str,
+        clock: Clock,
+    ):
+        self._key = key
+        self._oauth_client_id = oauth_client_id
+        self._oauth_issuer = oauth_issuer
+        self._clock = clock
+        self._cached_secret = b""
+        self._cached_secret_replacement_time = 0
```
clokep, line 1027: I'm not quite sure I understand what would use the cached values here -- does `ClientAuth` cache the secret internally?

richvdh: no it does not. `_get_secret` returns the cached value.

richvdh:
oops, helps if you commit all the files!

synapse/handlers/oidc_handler.py
clokep: I think my question was unclear -- it looked to me like `ClientAuth` was only used once and then discarded, but I see that a single instance is re-used while exchanging the code.

I guess I'm surprised that caching is necessary -- is it very inefficient to calculate the secret each time? Or are you just not supposed to do that?

clokep (approved):
I think this looks fine! I left another question about the caching, but that's mostly for my own understanding!

synapse/handlers/oidc_handler.py
clokep: Fair enough! Thanks for the explanation.

PR #9559, opened 2021-03-08
Warn that /register will soon require a type when called with an access token

Half-Shot:
This notice is giving a heads up to the planned spec compliance fix https://github.com/matrix-org/synapse/pull/9548.


docs/9559.removal
```diff
@@ -0,0 +1 @@
+Support for calling /register without a type of `m.login.application_service` when trying to register an AS user will be dropped in an upcoming release.
```
anoadragon453, line 1:
```suggestion
Registering an Application Service user without using the `m.login.application_service` login type will be unsupported in an upcoming Synapse release.
```

anoadragon453 (changes requested):
Typically we'll also call out noteworthy deprecations at the top of the release notes ([example](https://github.com/matrix-org/synapse/blob/develop/CHANGES.md#removal-warning)). Could you add a note to the top of CHANGES.md with a similar blurb? We'll format it nicely when doing the release.

anoadragon453:
Also the changelog needs to go in `changelog.d/`.

anoadragon453 (approved):
Looks like we're happy to include [the related change](https://github.com/matrix-org/synapse/pull/9548) - the discussion is now just about how long to include the notice for.

PR #9560, opened 2021-03-08
Add a list of hashes to ignore during git blame.

clokep:
This adds a file which lists revs which should be ignored when doing a `git blame`, this is useful for mass reformatting, etc. You can configure git to use it automatically via `git config blame.ignoreRevsFile .git-blame-ignore-revs`.

Note that we only want to include changes in here which we are highly confident won't change behavior (e.g. we don't want to include things like the async/await conversation).

I suspect we should also document running `git config`. Maybe in CONTRIBUTING, but I like keeping that simple... Any thoughts?

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,8 @@
+# Black reformatting (#5482).
+32e7c9e7f20b57dd081023ac42d6931a8da9b3a3
+
+# Target Python 3.5 with black (#8664).
+aff1eb7c671b0a3813407321d2702ec46c71fa56
+
+# Update black to 20.8b1 (#9381).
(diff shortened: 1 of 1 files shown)


anoadragon453:
https://github.com/matrix-org/synapse/commit/b7ddece2a6d541367a10d00d1e7851e29fce2612 is a good contender to include as well I believe.

clokep:
> [b7ddece](https://github.com/matrix-org/synapse/commit/b7ddece2a6d541367a10d00d1e7851e29fce2612) is a good contender to include as well I believe.

Probably, it does have some changes which *could* change behavior though...

erikjohnston (approved):
I think this makes sense

Issue #9492, opened 2021-02-24
federation catchup is very inefficient


richvdh:
If a remote HS takes a long time to process a transaction (perhaps because of #9490), then we will enter "catchup mode" for that homeserver. In this mode, we only ever send the most recent event per room, which means that the remote homeserver has to backfill for any intermediate events, which is, comparatively very slow (and may require calling `/state_ids`, see https://github.com/matrix-org/synapse/issues/7893 and https://github.com/matrix-org/synapse/issues/6597).

So, we have a homeserver which is struggling to keep up, and we are making its life even harder by only sending it a subset of the traffic in the room.

I'm not at all sure why we do this. Shouldn't we send all the events that the remote server hasn't received (up to a limit)?

ShadowJonathan:
What does the spec say? Shouldn't it try to send all events it can queue up (within reason)?

anoadragon453:
It looks like we fetch the oldest events that the remote has not yet received: https://github.com/matrix-org/synapse/blob/0a00b7ff14890987f09112a2ae696c61001e6cf1/synapse/federation/sender/per_destination_queue.py#L461-L465

https://github.com/matrix-org/synapse/blob/0a00b7ff14890987f09112a2ae696c61001e6cf1/synapse/storage/databases/main/transactions.py#L424-L442

rather than the most recent event per room?

richvdh:
destination_rooms stores the stream_ordering of the most recent event in each room that we *should have* sent to each destination, rather than the most recent event that we actually did send.  So when we join to events, we get the event id of that single most recent event per room.

erikjohnston:
I wonder if another thing at play here is the interaction of all the different servers in the room. Take a busy room like Matrix HQ, now after a few hours of down time a server will receive the latest event from *all* servers that sent an event during that time, for each one the receiving server will first do `/get_missing_events` to fetch up to ten missing events, then do a `/state_ids` request if there is still a gap. This means that for a single room the server can end up processing a *lot* more than just the latest event, and end up with a many small chunks of DAG int the gap between it going offline and coming back online

ShadowJonathan:
Maybe rooms need something similar to `ResponseCache.wrap()`, in the sense that getting the backlog and state_id is handled by one worker, and every subsequent request is queued to that until it is done, this might require closer and more complex coordination (even more so across workers), but it'll solve that problem.

erikjohnston:
[...]

PR #9561, opened 2021-03-08
Don't go into federation catch up mode so easily

erikjohnston:
Federation catch up mode is very inefficient if the number of events
that the remote server has missed is small, since handling gaps can be
very expensive, c.f. #9492.

Instead of going into catch up mode whenever we see an error, we instead
do so only if we've backed off from trying the remote for more than an
hour (the assumption being that in such a case it is more than a
transient failure).

### DIFF
--- a/synapse/federation/sender/per_destination_queue.py
+++ b/synapse/federation/sender/per_destination_queue.py
@@ -93,6 +93,10 @@ class PerDestinationQueue:
         self._destination = destination
         self.transmission_loop_running = False
 
+        # Flag to signal to any running transmission loop that there is new data
+        # queued up to be sent.
+        self._new_data_to_send = False
+
         # True whilst we are sending events that the remote homeserver missed
         # because it was unreachable. We start in this state so we can perform
         # catch-up at startup.
@@ -208,6 +212,10 @@ class PerDestinationQueue:
         transaction in the background.
         """
 
+        # Mark that we (may) have new things to send, so that any running
+        # transmission loop will recheck whether there is stuff to send.
+        self._new_data_to_send = True
+
         if self.transmission_loop_running:
             # XXX: this can get stuck on by a never-ending
             # request at which point pending_pdus just keeps growing.
@@ -250,6 +258,8 @@ class PerDestinationQueue:
 
             pending_pdus = []
             while True:
+                self._new_data_to_send = False
+
                 # We have to keep 2 free slots for presence and rr_edus
                 limit = MAX_EDUS_PER_TRANSACTION - 2
 
@@ -265,15 +275,6 @@ class PerDestinationQueue:
                 ) = await self._get_to_device_message_edus(limit)
 
                 pending_edus = device_update_edus + to_device_edus
-
-                # BEGIN CRITICAL SECTION
-                #
-                # In order to avoid a race condition, we need to make sure that
-                # the following code (from popping the queues up to the point
-                # where we decide if we actually have any pending messages) is
-                # atomic - otherwise new PDUs or EDUs might arrive in the
-                # meantime, but not get sent because we hold the
-                # transmission_loop_running flag.
 
                 pending_pdus = self._pending_pdus
 
@@ -320,14 +321,21 @@ class PerDestinationQueue:
                 if not pending_pdus and not pending_edus:
                     logger.debug("TX [%s] Nothing to send", self._destination)
                     self._last_device_stream_id = device_stream_id
-                    return
+
+                    # If we've gotten told about new things to send during
+                    # checking for things to send, we try looking again.
+                    # Otherwise new PDUs or EDUs might arrive in the meantime,
+                    # but not get sent because we hold the
+                    # `transmission_loop_running` flag.
+                    if self._new_data_to_send:
+                        continue
+                    else:
+                        return
 
                 # if we've decided to send a transaction anyway, and we have room, we
                 # may as well send any pending RRs
                 if len(pending_edus) < MAX_EDUS_PER_TRANSACTION:
                     pending_edus.extend(self._get_rr_edus(force_flush=True))
-
-                # END CRITICAL SECTION
 
                 success = await self._transaction_manager.send_new_transaction(
                     self._destination, pending_pdus, pending_edus
--- a/synapse/federation/sender/per_destination_queue.py
+++ b/synapse/federation/sender/per_destination_queue.py
@@ -15,23 +15,10 @@
 # limitations under the License.
 import datetime
 import logging
-from collections import deque
-from typing import (
-    TYPE_CHECKING,
-    Dict,
-    Hashable,
-    Iterable,
-    Iterator,
-    List,
-    Optional,
-    Tuple,
-    TypeVar,
-    cast,
-)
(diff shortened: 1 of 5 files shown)


synapse/federation/sender/per_destination_queue.py
```diff
@@ -422,16 +421,12 @@ async def _transaction_transmission_loop(self) -> None:
...
                     "Failed to send event %s to %s", p.event_id, self._destination
                 )
-
-            self._start_catching_up()
         except Exception:
             logger.exception("TX [%s] Failed to send transaction", self._destination)
             for p in pending_pdus:
                 logger.info(
                     "Failed to send event %s to %s", p.event_id, self._destination
                 )
```
anoadragon453, line 429: Unrelated, but I wonder if we should try to log the event IDs on one line here instead of 50. In another PR though.

anoadragon453 (commented):
So in the case of someone who's gone offline for 55 minutes, they'll still get missed messages as transactions will still be retried from other servers after only that short amount of downtime?

erikjohnston:
> So in the case of someone who's gone offline for 55 minutes, they'll still get missed messages as transactions will still be retried from other servers after only that short amount of downtime?

That's the idea.

But I think I've messed it up....

anoadragon453:
I assume #9567 is intended to sort out the usecase I mentioned? In which case this lgtm.

erikjohnston:
For context there was a problem where we dropped the PDUs that were included in a transaction even if it failed, which meant if we *didn't* go into catchup mode those events would not be sent out. The above commits make it so that we requeue up events if we fail to send them (plus a bunch of refactors to make that easier).

This isn't quite ready for review yet though.

erikjohnston:
Ok, I think this is now ready for review again. Probably easiest to look at it commit-by-commit

synapse/federation/sender/per_destination_queue.py
```diff
@@ -580,3 +501,125 @@ def _start_catching_up(self) -> None:
...
+        limit = MAX_EDUS_PER_TRANSACTION - 2
+
+        device_update_edus, dev_list_id = await self.queue._get_device_update_edus(
+            limit
+        )
+
+        if device_update_edus:
+            self._device_list_id = dev_list_id
+        else:
+            self.queue._last_device_list_stream_id = dev_list_id
```
clokep, line 543: It looks like we used to do this inside of the `if not pending_pdus and not pending_edus` block. Is this a behavior change?

synapse/federation/sender/per_destination_queue.py
```diff
@@ -580,3 +501,125 @@ def _start_catching_up(self) -> None:
...
+        pending_edus = device_update_edus + to_device_edus
+
+        pending_pdus = self.queue._pending_pdus
+
+        # We can only include at most 50 PDUs per transactions
+        pending_pdus, self.queue._pending_pdus = pending_pdus[:50], pending_pdus[50:]
+
+        pending_edus.extend(self.queue._get_rr_edus(force_flush=False))
+        pending_presence = self.queue._pending_presence
+        self.queue._pending_presence = {}
```
clokep, line 564: This will cause us to declare a new dict in some cases when we don't need to, it seems we could (instead) check `if self.queue._pending_presence` below and only overwrite this value if necessary (i.e. the `pending_presence` variable is unnecessary).

synapse/federation/sender/per_destination_queue.py
```diff
@@ -580,3 +501,125 @@ def _start_catching_up(self) -> None:
...
+            )
+
+        pending_edus.extend(
+            self.queue._pop_pending_edus(MAX_EDUS_PER_TRANSACTION - len(pending_edus))
+        )
+        while (
+            len(pending_edus) < MAX_EDUS_PER_TRANSACTION
+            and self.queue._pending_edus_keyed
+        ):
+            _, val = self.queue._pending_edus_keyed.popitem()
```
clokep, line 589: This is copied & pasted, but `popitem` is LIFO so it is theoretically possible you'll never reach the bottom of this queue, which is annoying. 😢

clokep (commented):
I think that this looks good overall, but I left some comments. Most of them can probably be ignored / deferred to another PR, but a couple are trying to understand what this is doing. 😄

synapse/federation/sender/per_destination_queue.py
```diff
@@ -577,25 +589,29 @@ async def __aenter__(self) -> Tuple[List[EventBase], List[Edu]]:
...
-        return pending_pdus, pending_edus
+        return self._pdus, pending_edus
 
     async def __aexit__(self, exc_type, exc, tb):
         if exc_type is not None:
-            # Failed to send transaction, nothing to do.
+            # Failed to send transaction. Requeue events we failed to send this
+            # time round.
+            if self._pdus:
+                self.queue._pending_pdus.extendleft(self._pdus)
```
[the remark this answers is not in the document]
erikjohnston: Annoyingly `deque` can't be sliced like that (which makes sense, given how it's implemented). We could use an iterator to pull out the first N items, so `self._pdus = list(itertools.islice(self.queue._pending_pdus, 50))`, and then pop here, but I don't know its slightly sucky.

synapse/federation/sender/per_destination_queue.py
erikjohnston: Hmm, yeah. In a separate PR i was going to make it so that we don't drop the EDUs if the transaction fails (in the same way as for PDUs), so it'll probably get magically fixed there?

synapse/federation/sender/per_destination_queue.py
erikjohnston: No, we just need to only set it after we've sent all device EDUs that were returned, and if none are returned we can immediately set it. If we don't set it when we get none returned then that is also fine, just slightly less efficient (as the DB query needs to look at a larger range).

synapse/federation/sender/per_destination_queue.py
```diff
@@ -580,3 +501,125 @@ def _start_catching_up(self) -> None:
...
+            to_device_edus,
+            device_stream_id,
+        ) = await self.queue._get_to_device_message_edus(limit)
+
+        if to_device_edus:
+            self._device_stream_id = device_stream_id
+        else:
+            self.queue._last_device_stream_id = device_stream_id
+
+        pending_edus = device_update_edus + to_device_edus
```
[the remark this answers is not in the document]
erikjohnston: Oh, yeah, could do. I have an in built phobia of concatenating collections using `+` (given it unnecessarily allocates temporary collections), but it doesn't really matter.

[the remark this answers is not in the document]
clokep: That's fair! I guess my real complain is that we should be grouping these things together so the logic to build `pending_edus` is in one spot.

synapse/federation/sender/per_destination_queue.py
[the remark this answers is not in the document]
clokep: In my example above I'm unsure why a `deque` would be helpful and it couldn't be a simple `list`?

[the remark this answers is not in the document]
erikjohnston: Ok, yes that actually makes sense. Gah, I find reasoning about lists vs deques confusing

PR #9562, opened 2021-03-08
Prevent the config-lint script erroring out on any sample_config changes

anoadragon453:
I noticed that I'd occasionally have `scripts-dev/lint.sh` fail when messing about with config options in my PR. The script calls `scripts-dev/config-lint.sh`, which attempts some validation on the sample config.

 It does this by using `sed` to edit the sample_config, and then seeing if the file changed using `git diff`.

The problem is: if you changed the sample_config as part of your commit, this script will error regardless.

This PR attempts to change the check so that existing, unstaged changes to the sample_config will not cause the script to report an invalid file.

### DIFF
--- a/scripts-dev/config-lint.sh
+++ b/scripts-dev/config-lint.sh
@@ -2,9 +2,14 @@
 # Find linting errors in Synapse's default config file.
 # Exits with 0 if there are no problems, or another code otherwise.
 
+# cd to the root of the repository
+cd `dirname $0`/..
+
+# Restore backup of sample config upon script exit
+trap "mv docs/sample_config.yaml.bak docs/sample_config.yaml" EXIT
+
 # Fix non-lowercase true/false values
 sed -i.bak -E "s/: +True/: true/g; s/: +False/: false/g;" docs/sample_config.yaml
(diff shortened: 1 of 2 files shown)


scripts-dev/config-lint.sh
```diff
@@ -2,9 +2,14 @@
...
+# Restore backup of sample config upon script exit
+trap "mv docs/sample_config.yaml.bak docs/sample_config.yaml" EXIT
+
 # Fix non-lowercase true/false values
 sed -i.bak -E "s/: +True/: true/g; s/: +False/: false/g;" docs/sample_config.yaml
-rm docs/sample_config.yaml.bak
 
 # Check if anything changed
-git diff --exit-code docs/sample_config.yaml
+diff docs/sample_config.yaml docs/sample_config.yaml.bak
```
anoadragon453, line 15: The script will now just compare the modified sample config against the `*.bak` file, and no longer uses `git diff` for comparison. `diff` will report the same error codes.

The `trap` statement will move the original sample config (`*.bak`) on top of the modified one upon script exit.

erikjohnston (approved):

PR #9568, opened 2021-03-09
Do not have mypy ignore unpaddedbase64.

clokep:
Per matrix-org/python-unpaddedbase64#7.

### DIFF
--- a/changelog.d/9568.misc
+++ b/changelog.d/9568.misc
@@ -1,0 +1,1 @@
+Do not have mypy ignore type hints from unpaddedbase64.
--- a/mypy.ini
+++ b/mypy.ini
@@ -117,9 +117,6 @@ ignore_missing_imports = True
 [mypy-saml2.*]
 ignore_missing_imports = True
 
-[mypy-unpaddedbase64]
-ignore_missing_imports = True
-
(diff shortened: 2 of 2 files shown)


ShadowJonathan (changes requested):
Does the `"unpaddedbase64>=1.1.0"` in `synapse/python_dependencies.py` not need to be changed?

clokep:
> Does the `"unpaddedbase64>=1.1.0"` in `synapse/python_dependencies.py` not need to be changed?

I do not think so, because this is only used for type checking. This isn't a run-time requirement. We generally assume you're using the latest versions of libraries when running mypy.

ShadowJonathan:
> We generally assume you're using the latest versions of libraries when running mypy.

I'd suggest changing it anyway with this PR, for determinism, because to actually update it you'd need to do `pip install -e .[all] -U`, which isn't signalled in `CONTRIBUTING.md`. The default flow for contributors is to do that when updating your branch, so this'd fail on their branch and they'd be confused why until they slap the upgrade flat on `pip install`.

Maybe a bit too dramatic, but eh.

erikjohnston (approved):

PR #9573, opened 2021-03-09
Fix the auth provider on the logins metric

richvdh:
We either need to pass the auth provider over the replication api, or make sure
we report the auth provider on the worker that received the request. I've gone
with the latter.


clokep (approved):
LGTM!

PR #9576, opened 2021-03-09
Use the chain cover index in get_auth_chain_ids

### LINKED ISSUE #9482: Use chain cover index in get_auth_chain_ids
We have an instance of Synapse that is spending a lot of CPU time in `get_auth_chain_ids`. @erikjohnston suggested the CPU load could be reduced:

> but we should probably just make `get_auth_chain_ids` use the chain cover index (its just a simplified version of get_auth_chain_diff), which should be much faster

clokep:
This should be reviewable commit-by-commit, but this is a simplified version of `get_auth_chain_difference` when using the chain cover. As with that method we bail if the chain cover is broken or non-existent and fall back to the old algorithm. I've also included some simple tests for this, I'm not sure I hit all the edge cases in them though.

Fixes #9482

### DIFF
--- a/synapse/storage/databases/main/event_federation.py
+++ b/synapse/storage/databases/main/event_federation.py
@@ -125,8 +125,8 @@ class EventFederationWorkerStore(EventsWorkerStore, SignatureWorkerStore, SQLBas
         # All the events that we've found that are reachable from the events.
         seen_events = set()  # type: Set[str]
 
-        # A map from chain ID to max sequence number reachable from any event ID.
-        chains = {}  # type: Dict[int, int]
+        # A map from chain ID to max sequence number of the given events.
+        event_chains = {}  # type: Dict[int, int]
 
         sql = """
             SELECT event_id, chain_id, sequence_number
@@ -141,7 +141,9 @@ class EventFederationWorkerStore(EventsWorkerStore, SignatureWorkerStore, SQLBas
 
             for event_id, chain_id, sequence_number in txn:
                 seen_events.add(event_id)
-                chains[chain_id] = max(sequence_number, chains.get(chain_id, 0))
+                event_chains[chain_id] = max(
+                    sequence_number, event_chains.get(chain_id, 0)
+                )
 
         # Check that we actually have a chain ID for all the events.
         events_missing_chain_info = initial_events.difference(seen_events)
@@ -165,8 +167,12 @@ class EventFederationWorkerStore(EventsWorkerStore, SignatureWorkerStore, SQLBas
             WHERE %s
         """
 
-        # (We need to take a copy of `chains` as we want to mutate it in the loop)
-        for batch in batch_iter(set(chains), 1000):
+        # A map from chain ID to max sequence number *reachable* from any event ID.
+        chains = dict(event_chains)
+
+        # We need to take a copy of `event_chains` as we need to separate chain
+        # IDs / seq nos from the given events vs. reachable ones.
+        for batch in batch_iter(event_chains, 1000):
             clause, args = make_in_list_sql_clause(
                 txn.database_engine, "origin_chain_id", batch
             )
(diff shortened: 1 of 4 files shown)


clokep:
Some of this is copy and pasted from `get_auth_chain_difference` since the need is identical, but only over a single set of events (instead of multiple state sets). I'm thinking about whether we can abstract it or not.

clokep:
Something to note: using the chain cover for `get_auth_chain_difference` might have made the performance of `get_auth_chain_ids` worse since the non-chain cover versions of those share a cache (`_event_auth_cache`) which was no longer used by `get_auth_chain_difference` once the chain cover was calculated.

erikjohnston:
> Something to note: using the chain cover for `get_auth_chain_difference` might have made the performance of `get_auth_chain_ids` worse since the non-chain cover versions of those share a cache (`_event_auth_cache`) which was no longer used by `get_auth_chain_difference` once the chain cover was calculated.

I think we could probably add some caches that would have a similar effect. Caching `event_id -> reachable chains` would likely help a lot, though possibly a bit overkill. I think I'd be fine with leaving it as is and seeing if it causes problems.

erikjohnston:
> > Some of this is copy and pasted from `get_auth_chain_difference` since the need is identical, but only over a single set of events (instead of multiple state sets). I'm thinking about whether we can abstract it or not.
> 
> I could not figure out a good way to do this that didn't calculate unnecessary things for `get_auth_chain_ids`, additionally the code is different in a couple of subtle ways and I think abstracting it might make it more confusing. If anyone has ideas of parts that can be shared I'd love to hear!

I think its fine to leave them split out, I think its almost better to be clearer here than to be DRY.

synapse/storage/databases/main/event_federation.py
```diff
@@ -54,46 +54,195 @@ def __init__(self, database: DatabasePool, db_conn, hs):
...
+                        target_chain_id in event_chains
+                        and target_sequence_number >= event_chains[target_chain_id]
+                    ):
+                        event_chains.pop(target_chain_id)
+
+        # Don't include the given events (since we're finding the auth chain of
+        # those events).
+        for chain_id in event_chains:
+            if event_chains[chain_id] == chains[chain_id]:
+                chains[chain_id] -= 1
```
erikjohnston, line 204: I'm finding the logic around this quite confusing, and I'm not sure I totally buy that it works. Though the tests do somewhat suggest it does.

Can we instead have `chains` start off being the empty set, so that it only contains *other* chains that are reachable from the initial events, and then add in the chains from `event_chains`? So something like:

```python
chains = set()

# Add all other chains reachable from initial set
for start, end in txn:
    if start.seq <= chains.get(start.id):
        chains[start.id] = start.seq

# Add initial set of chains, excluding the the seq corresponding to initial event
for id, seq in event_chains:
    chains[id] = max(seq - 1, chains.get(id))
```

erikjohnston (commented):
I think this looks good, I'm just wondering if we can make the `include_given` handling a bit clearer

synapse/storage/databases/main/event_federation.py
erikjohnston: Oh, interesting. Don't spend too long trying it again then, it may be that its just complex logic no matter how you slice it.

clokep: I agree that what you have looks cleaner, it is possible I had other bugs that got fixed though!

(Note that this also only works because we don't use a recursive implementation.)

clokep: I was able to make this change pretty easily, I think it cleans it up a bit!

synapse/storage/databases/main/event_federation.py
```diff
@@ -54,46 +54,185 @@ def __init__(self, database: DatabasePool, db_conn, hs):
...
+            )
+            txn.execute(sql % (clause,), args)
+
+            for (
+                origin_chain_id,
+                origin_sequence_number,
+                target_chain_id,
+                target_sequence_number,
+            ) in txn:
+                if origin_sequence_number <= event_chains.get(origin_chain_id, 0):
```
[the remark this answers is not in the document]
clokep: Seems that I dropped the comment from the spot I copied this from. 😢 I'll add it back.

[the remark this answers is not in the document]
erikjohnston: Hah, I didn't even bother to check that as I just assumed I was an awful person that didn't bother writing a comment in the first place!

Issue #8944, opened 2020-12-15
Document how to write plugins


erikjohnston:
We have plugin support used in various places, some of these are documented to a greater or lesser extent (e.g [SSO mapping providers](https://github.com/matrix-org/synapse/blob/develop/docs/sso_mapping_providers.md)), but we should have a generic doc they can all point to explaining:

1. How to write a plugin, e.g the function signatures of `__init__` and `parse_config`,
2. How to deploy these plugins in practice (i.e. make them a python package and install them in the venv?)

babolivier:
The first point is already covered in https://matrix-org.github.io/synapse/latest/modules/writing_a_module.html

So I'm going to rename this issue to cover only the second one.

richvdh:
related: #9944.

jaye-ww:
I think the cookie cutter template for Synapse modules should be mentioned in the docs:
https://github.com/matrix-org/synapse-module-cookiecutter-template

PR #9580, opened 2021-03-10
Fix spam checker modules documentation example

jaywink:
A parse_config method is mandatory on spam checked modules.

This relates to #8944 as well.

### DIFF
--- a/changelog.d/9580.doc
+++ b/changelog.d/9580.doc
@@ -1,1 +1,1 @@
-Fix spam checker modules documentation example to clarify parse_config is a required method.
+Clarify the spam checker modules documentation example to mention that `parse_config` is a required method.
--- a/changelog.d/9580.doc
+++ b/changelog.d/9580.doc
@@ -1,0 +1,1 @@
+Fix spam checker modules documentation example to clarify parse_config is a required method.
--- a/docs/spam_checker.md
+++ b/docs/spam_checker.md
@@ -14,7 +14,7 @@ The Python class is instantiated with two objects:
 * An instance of `synapse.module_api.ModuleApi`.
 
 It then implements methods which return a boolean to alter behavior in Synapse.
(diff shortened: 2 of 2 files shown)


changelog.d/9580.doc
```diff
@@ -0,0 +1 @@
+Fix spam checker modules documentation example to clarify parse_config is a required method.
```
clokep, line 1:
```suggestion
Clarify the spam checker modules documentation example to mention that `parse_config` is a required method.
```

docs/spam_checker.md
```diff
@@ -14,6 +14,7 @@ The Python class is instantiated with two objects:
 * An instance of `synapse.module_api.ModuleApi`.
 
 It then implements methods which return a boolean to alter behavior in Synapse.
+All the methods are mandatory to be defined.
```
clokep, line 17:
```suggestion
All the methods are must be defined.
```

docs/spam_checker.md
```diff
@@ -31,6 +32,10 @@ are documented in the `synapse.events.spamcheck.SpamChecker` class.
 The `ModuleApi` class provides a way for the custom spam checker class to
 call back into the homeserver internals.
 
+Additionally, a `parse_config` method is mandatory and receives the plugin config
+dictionary. After possible sanitizing, It must return a dictionary which
+will be passed to `__init__` later.
```
clokep, line 37: This doesn't need to return a dictionary, but any object which can go into the `__init__`, I think.

```suggestion
dictionary. After parsing, It must return an object which will be
passed to `__init__` later.
```

clokep (changes requested):
While we're updating this, can we include the `check_media_file_for_spam` method in the list of things to implement?

clokep (approved):
Looks great! Thank you!

PR #9581, opened 2021-03-10
Add `get_userinfo_by_id` method to `ModuleApi`

jaywink:
Makes it easier to fetch user details in for example spam checker modules, without needing to use api._store or figure out database interactions.


synapse/module_api/__init__.py
```diff
@@ -72,6 +72,16 @@ def public_room_list_manager(self):
         """
         return self._public_room_list_manager
 
+    def get_user_by_id(self, user_id: str) -> Optional[Dict[str, Any]]:
+        """Get user by user_id
+
+        Args:
+            user_id: Fully qualified user id.
+        Returns:
+            the user dictionary or None if not found.
```
richvdh, line 81: I'm a bit concerned about returning an unstructured dict here, with no contract about what fields it contains. What assumptions is the module allowed to make?

synapse/module_api/__init__.py
```diff
@@ -72,6 +72,16 @@ def public_room_list_manager(self):
...
+        """
+        return self._store.get_user_by_id(user_id)
```
richvdh, line 83: isn't get_user_by_id async?

richvdh (changes requested):

synapse/module_api/__init__.py
jaywink: Do we have a user type that could be constructed?

synapse/module_api/__init__.py
jaywink: Good call. Is `defer.Deferred` the correct type or is the some other way to signify with typing that it's a deferred containing either a dictionary or None? Looking around, that type seems to be set as is for return values where used.

richvdh: generally it's best to declare the function `async` and put an `await` in here.

synapse/module_api/__init__.py
richvdh: Not at present. You could define one (maybe in the `types` module, using `attr.s`), or just list the fields here.

jaywink: Extremely sorry to let this sit for such a long time. I reviewed it's still useful with the new modules architecture. There is now a `User` type. It tries to match some of the way `Requester` names things, for example using `user` instead of `user_id` or `name` (actual database column). I was tempted to call it `ModuleApiUser` to match some of the specific types found, but since it pretty much matches what comes out of the database, left it as `User`.

jaywink:
Hmm, interesting, the thing the linting is complaining about was actually formatted by `./scripts-dev/lint.sh -d` :thinking:

richvdh:
> Hmm, interesting, the thing the linting is complaining about was actually formatted by ./scripts-dev/lint.sh -d

It looks like you might be using the wrong version of `black`? It should be 21.6b0, as per `setup.py`.

Meanwhile it looks like `mypy` is flagging some valid problems.

tests/test_types.py
```diff
@@ -104,3 +111,49 @@ def testNonAscii(self):
         # this should work with either a unicode or a bytes
         self.assertEqual(map_username_to_mxid_localpart("têst"), "t=c3=aast")
         self.assertEqual(map_username_to_mxid_localpart("têst".encode()), "t=c3=aast")
+
+
+class UserTestCase(unittest.TestCase):
```
richvdh, line 116: I think this is mostly testing that `attrs` works as expected? I don't think I'd bother.

synapse/types.py
```diff
@@ -751,3 +751,32 @@ def get_verify_key_from_cross_signing_key(key_info):
     # and return that one key
     for key_id, key_data in keys.items():
         return (key_id, decode_verify_key_bytes(key_id, decode_base64(key_data)))
+
+
+@attr.s(frozen=True, slots=True)
+class User:
```
richvdh, line 757: I think we probably want to call this `UserInfo` or something rather than `User`.

synapse/types.py
```diff
@@ -751,3 +751,32 @@ def get_verify_key_from_cross_signing_key(key_info):
...
+        consent_server_notice_sent:  Version of policy documents the user has been sent.
+        consent_version:  Version of policy documents the user has consented to.
+        creation_ts:  Creation timestamp of the user.
+        deactivated:  True if the user has been deactivated.
+        is_guest:  True if the user is a guest user.
+        shadow_banned:  True if the user has been shadow-banned.
+        user_type:  User type (None for normal user, 'support' and 'bot' other options).
+    """
+
+    user = attr.ib(type="UserID")
```
richvdh, line 773: nowadays this can be written:

```
user: UserID = attr.ib()
```

... or, even more succinctly, along with `attr.s(auto_attribs=True)`:

```
user: UserID
```

(see https://www.attrs.org/en/stable/types.html?highlight=auto_attribs#type-annotations)


If you're using auto_attribs, default values can also be given this way:

```
admin: bool = False
```

- not that I think you actually need defaults, since all the values are set explicitly.

I won't insist on you changing these, but I thought it might be useful to know for the future at least.

richvdh, line 773:
```suggestion
    user_id = attr.ib(type="UserID")
```

synapse/module_api/__init__.py
```diff
@@ -174,6 +174,34 @@ def email_app_name(self) -> str:
...
+                creation_ts=user_data.get("creation_ts"),
+                consent_version=user_data.get("consent_version"),
+                consent_server_notice_sent=user_data.get("consent_server_notice_sent"),
+                app_service=app_service,
+                shadow_banned=user_data.get("shadow_banned"),
+                deactivated=user_data.get("deactivated"),
+                is_guest=user_data.get("is_guest"),
+                admin=user_data.get("admin"),
+                user_type=user_data.get("user_type"),
+            )
```
richvdh, line 203: I think this logic belongs outside the module_api, which is meant to be a very thin wrapper.

Maybe better to add a `get_user_info_by_id` alongside `store.get_user_by_id`, and call straight through to that here.

(We can also then mark `_store.get_user_by_id` deprecated, since its interface is horrible.)

synapse/types.py
```diff
@@ -751,3 +751,32 @@ def get_verify_key_from_cross_signing_key(key_info):
...
+    admin = attr.ib(type=bool, default=False, converter=bool)
```
richvdh, line 774: why do we need `converter`s on this and the others below? If we're trying to assign something other than a bool then that suggests something wrong in the caller?

synapse/types.py
```diff
@@ -751,3 +751,32 @@ def get_verify_key_from_cross_signing_key(key_info):
...
+    app_service = attr.ib(type=Optional["ApplicationService"], default=None)
+    consent_server_notice_sent = attr.ib(type=Optional[str], default=None)
+    consent_version = attr.ib(type=Optional[str], default=None)
+    creation_ts = attr.ib(type=int, default=0)
+    deactivated = attr.ib(type=bool, default=False, converter=bool)
+    is_guest = attr.ib(type=bool, default=False, converter=bool)
```
richvdh, line 780: can we be consistent about whether the bool properties have `is_` prefixes? I think it's probably clearer if they do (so `is_admin`, `is_deactivated`, etc), but the inconsistency is particularly horrible.

richvdh (changes requested):
thanks for your work on this @jaywink! I've made a few suggestions below.

synapse/types.py
jaywink: I added those since the database record from `_store.get_user_by_id` seemed to not return bools. Agreed this is not the place to do such things, removing.

synapse/module_api/__init__.py
jaywink: I believe this could be up for a new review. The last commit is due to mypy, it feels wrong needing to have that but I'm not sure if there is another way around things?

Probably best deprecate `_store.get_user_by_id` in a follow-up pr, I'm happy to do that?

synapse/storage/databases/main/registration.py
```diff
@@ -188,17 +188,17 @@ async def get_userinfo_by_id(self, user_id: str) -> Optional[UserInfo]:
...
             consent_version=user_data.get("consent_version"),
-            creation_ts=user_data.get("creation_ts"),
+            creation_ts=user_data.get("creation_ts", 0),
             is_admin=bool(user_data.get("admin")),
             is_deactivated=bool(user_data.get("deactivated")),
             is_guest=bool(user_data.get("is_guest")),
             is_shadow_banned=bool(user_data.get("shadow_banned")),
-            user_id=UserID.from_string(user_data.get("name")),
+            user_id=UserID.from_string(str(user_data.get("name"))),
             user_type=user_data.get("user_type"),
```
richvdh: > The fixes seem a bit hacky just to fix the type checks, as the return types are `Any` from the selects.

It's because you're using `user_data.get`. `.get` returns `None` if the dict doesn't contain that property. In this case, we know that the dict *will* contain each of these properties, so you should do `user_data["appservice_id"]` instead, which I think will also keep mypy happy.

synapse/storage/databases/main/registration.py
```diff
@@ -166,6 +166,42 @@ async def get_user_by_id(self, user_id: str) -> Optional[Dict[str, Any]]:
             desc="get_user_by_id",
         )
 
+    @cached()
```
richvdh, line 169: it'd be swell if `get_userinfo_by_id` and `get_user_by_id` could share a cache. I think the easiest way to do that is by having one call the other.

synapse/storage/databases/main/registration.py
```diff
@@ -166,6 +166,42 @@ async def get_user_by_id(self, user_id: str) -> Optional[Dict[str, Any]]:
...
+    async def get_userinfo_by_id(self, user_id: str) -> Optional[UserInfo]:
```
richvdh, line 170: please can you give this a docstring?

Also, please can you stick a comment or docstring on get_user_by_id to say it's deprecated and replaced by get_userinfo_by_id?

synapse/types.py
```diff
@@ -751,3 +751,32 @@ def get_verify_key_from_cross_signing_key(key_info):
     # and return that one key
     for key_id, key_data in keys.items():
         return (key_id, decode_verify_key_bytes(key_id, decode_base64(key_data)))
+
+
+@attr.s(auto_attribs=True, frozen=True, slots=True)
+class UserInfo:
+    """Represents an object with information about a user.
```
richvdh, line 758: It *is* an object with information about a user.

```suggestion
    """Holds information about a user. Result of get_userinfo_by_id.
```

richvdh (changes requested):
thanks for your perseverance here @jaywink! A couple more tweaks suggested here.

synapse/storage/databases/main/registration.py
richvdh: > Probably best deprecate _store.get_user_by_id in a follow-up pr, I'm happy to do that?

Oh I just read this. I'd roll it into this PR, since it's simple enough.

Unless you were talking about actually replacing uses of `get_user_by_id` with `get_userinfo_by_id`? Which would be amazing if you've got time, and yes a separate PR would be best.

jaywink:
Review addressed, thanks for walking through! :pray:

changelog.d/9581.feature
```diff
@@ -1 +1 @@
-Add get_userinfo_by_id method to ModuleApi.
+Add `get_userinfo_by_id` method to ModuleApi. `UserInfo` is a new type returned from the registration store. This also deprecates the `get_user_by_id` method on the registration store.
```
richvdh, line 1: `get_user_by_id` was never exposed outside internal code, so I don't think we need to mention this here.

```suggestion
Add `get_userinfo_by_id` method to ModuleApi.
```

richvdh (approved):
lgtm otherwise

github-afsaneh (commented):
@https://github.com/matrix-org/synapse/pull/9581#issue-589412758

PR #9542, opened 2021-03-03
Add a background task to purge unused chain IDs.

clokep:
Fixes #9481

This iterates through the chain cover and ensures each event exists in the database. This seems wildly inefficient, but I couldn't think of a better way to do it.

PR #9583, opened 2021-03-10
Fix a bug in the background task for purging chain cover

clokep:
One of my refactors slightly broke #9542. We need to pass a list of tuples, not just a list to `executemany`.


waclaw66:
Success! DB purged, can join deleted rooms again 🎉

erikjohnston (approved):

PR #9585, opened 2021-03-10
Ensure we use a copy of the event content dict before modifying it in serialize_event

anoadragon453:
This bug was discovered by DINUM. We were modifying `serialized_event["content"]`, which - if you've got `USE_FROZEN_DICTS` turned on or are [using a third party rules module](https://github.com/matrix-org/synapse/blob/17cd48fe5171d50da4cb59db647b993168e7dfab/synapse/events/third_party_rules.py#L73-L76) - will raise a 500 if you try to a edit a reply to a message.

`serialized_event["content"]` could be set to the edit event's content, instead of a copy of it, which is bad as we attempt to modify it. Instead, we also end up modifying the original event's content. DINUM uses a third party rules module, which meant the event's content got frozen and thus an exception was raised.

To be clear, the problem is not that the event's content was frozen. In fact doing so helped us uncover the fact we weren't copying event content correctly.

### DIFF
--- a/tests/rest/client/test_third_party_rules.py
+++ b/tests/rest/client/test_third_party_rules.py
@@ -161,6 +161,68 @@ class ThirdPartyRulesTestCase(unittest.HomeserverTestCase):
         ev = channel.json_body
         self.assertEqual(ev["content"]["x"], "y")
 
+    def test_message_edit(self):
+        """Ensure that the module doesn't cause issues with edited messages."""
+        # first patch the event checker so that it will modify the event
+        async def check(ev: EventBase, state):
+            d = ev.get_dict()
+            d["content"] = {
+                "msgtype": "m.text",
+                "body": d["content"]["body"].upper(),
+            }
+            return d
+
+        current_rules_module().check_event_allowed = check
+
+        # Send an event, then edit it.
+        channel = self.make_request(
+            "PUT",
+            "/_matrix/client/r0/rooms/%s/send/modifyme/1" % self.room_id,
+            {
+                "msgtype": "m.text",
+                "body": "Original body",
+            },
+            access_token=self.tok,
+        )
+        self.assertEqual(channel.result["code"], b"200", channel.result)
+        orig_event_id = channel.json_body["event_id"]
+
+        channel = self.make_request(
+            "PUT",
+            "/_matrix/client/r0/rooms/%s/send/m.room.message/2" % self.room_id,
+            {
+                "m.new_content": {"msgtype": "m.text", "body": "Edited body"},
+                "m.relates_to": {
+                    "rel_type": "m.replace",
+                    "event_id": orig_event_id,
+                },
+                "msgtype": "m.text",
+                "body": "Edited body",
+            },
+            access_token=self.tok,
+        )
+        self.assertEqual(channel.result["code"], b"200", channel.result)
+        edited_event_id = channel.json_body["event_id"]
+
+        # ... and check that they both got modified
+        channel = self.make_request(
+            "GET",
+            "/_matrix/client/r0/rooms/%s/event/%s" % (self.room_id, orig_event_id),
+            access_token=self.tok,
+        )
+        self.assertEqual(channel.result["code"], b"200", channel.result)
+        ev = channel.json_body
+        self.assertEqual(ev["content"]["body"], "ORIGINAL BODY")
+
+        channel = self.make_request(
+            "GET",
+            "/_matrix/client/r0/rooms/%s/event/%s" % (self.room_id, edited_event_id),
+            access_token=self.tok,
+        )
+        self.assertEqual(channel.result["code"], b"200", channel.result)
+        ev = channel.json_body
+        self.assertEqual(ev["content"]["body"], "EDITED BODY")
+
     def test_send_event(self):
         """Tests that the module can send an event into a room via the module api"""
         content = {
--- a/tests/rest/client/v2_alpha/test_relations.py
+++ b/tests/rest/client/v2_alpha/test_relations.py
@@ -17,6 +17,7 @@ import itertools
 import json
 import urllib
 
+from synapse import events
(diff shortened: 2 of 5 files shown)


erikjohnston:
Can you add a unit test for this? It's just occurred to me that we should be able to just do `@overrid_config({"user_frozen_dicts": true})` to test it

erikjohnston:
Otherwise LGTM

erikjohnston (changes requested):
LGTM, just needs a test

clokep:
I added a test for this (and merged in develop for good measure). I think this is ready!

clokep:
@anoadragon453 You might also want to make sure you take a look at the tests. They do fail on develop and pass on this branch.

erikjohnston:
I've also added a test to make sure that this works when frozen dicts are in use, which I think is the most general case of this bug.

clokep (approved):
LGTM! Does that make my test duplicative or is there still value?

I don't know who can fully approve this now that we've all touched it. :)

erikjohnston:
> LGTM! Does that make my test duplicative or is there still value?
> 
> I don't know who can fully approve this now that we've all touched it. :)

Yay, team work! I think we may as well leave the test I guess? The more tests the better?

anoadragon453:
Aaaand it looks like another test is trying to modify the event dict :)

erikjohnston:
> Aaaand it looks like another test (`tests.storage.test_redaction.RedactionTestCase.test_circular_redaction`) is trying to modify the event dict :)

Oh, a) woops, b) it means that the `USE_FROZEN_DICTS` option is leaking between tests, which is not ideal

anoadragon453:
> it means that the `USE_FROZEN_DICTS` option is leaking between tests, which is not ideal

I think this was just because we were setting on the `events` import directly, which would then affect any tests that were run after it. Since there's a config option as well, I just opted to use that, which should be tied to the test homeserver state.

erikjohnston:
> > it means that the `USE_FROZEN_DICTS` option is leaking between tests, which is not ideal
> 
> I think this was just because we were setting on the `events` import directly, which would then affect any tests that were run after it. Since there's a config option as well, I just opted to use that, which should be tied to the test homeserver state.

SNAP!!! Except we need to add a hack to make the tests pick up the config option

erikjohnston:
It does somewhat beg the question whether we should be running CI with frozendicts on? :/

anoadragon453:
> It does somewhat beg the question whether we should be running CI with frozendicts on? :/

Probably yes? Although it can make writing tests harder?

erikjohnston:
How would it make writing tests harder?

anoadragon453:
> How would it make writing tests harder?

Well `tests.storage.test_redaction.RedactionTestCase.test_circular_redaction` failed because it tries to modify part of an event. We'd need to change that test so that it operates in a different way.

erikjohnston:
Oh right, I mean it sorta feels like that's just asking for trouble anyway :shrug:

tests/rest/client/test_third_party_rules.py
```diff
@@ -161,6 +161,68 @@ async def check(ev: EventBase, state):
...
+    def test_message_edit(self):
+        """Ensure that the module doesn't cause issues with edited messages."""
+        # first patch the event checker so that it will modify the event
+        async def check(ev: EventBase, state):
+            d = ev.get_dict()
+            d["content"] = {
+                "msgtype": "m.text",
+                "body": d["content"]["body"].upper(),
+            }
+            return d
```
anoadragon453, line 173: I'm a bit confused as we try to prevent modules from modifying events? https://github.com/matrix-org/synapse/blob/0a00b7ff14890987f09112a2ae696c61001e6cf1/synapse/events/third_party_rules.py#L73-L76

tests/rest/client/v2_alpha/test_relations.py
```diff
@@ -518,6 +523,63 @@ def test_multi_edit(self):
             {"event_id": edit_event_id, "sender": self.user_id}, m_replace_dict
         )
 
+    def test_edit_reply(self):
+        """Test that a editing a reply works."""
```
anoadragon453, line 527:
```suggestion
        """Test that editing a reply works."""
```

richvdh:
> It does somewhat beg the question whether we should be running CI with frozendicts on? :/

For the record: the idea of having them turned off for the UTs was that we should avoid using a different config for CI than we should for production, otherwise we end up with things that work with frozendict but then fail in production with frozen events turned off.

anoadragon453 (commented):
± @anoadragon453 requested feedback on this change.

By the way, after fixing the config option I tried running all unit tests with `use_frozen_dicts: true` and only 4 fail!

* `CanonicalAliasTestCase.test_remove_alias`
* `CanonicalAliasTestCase.test_remove_other_alias`
* `ModuleApiTestCase.test_sending_events_into_room`
* `RedactionTestCase.test_circular_redaction`

tests/rest/client/test_third_party_rules.py
clokep: How is this different than the test above that does the same thing? 😕 Maybe I misunderstood what that test was doing...

anoadragon453: Ahaaaaa. It was me who was misunderstanding what the point of freezing the event was.

So we do allow `ThirdPartyRules` modules to modify event dicts, but it must do so by returning a dict with the new fields, and must not modify the original event object.

So the freeze is just to ensure that last bit - but we still allow modifying the event.

In that case yes, this test looks solid!

anoadragon453 (commented):
:heavy_check_mark: @anoadragon453 has approved these changes.

erikjohnston (approved):

anoadragon453:
I'm taking the executive decision to merge this now. Good work team :hammer:

PR #9586, opened 2021-03-10
Convert Requester to attrs

richvdh:
... because namedtuples suck

Also fix up a couple of other annotations to keep mypy happy.

### DIFF
--- a/synapse/handlers/auth.py
+++ b/synapse/handlers/auth.py
@@ -337,7 +337,8 @@ class AuthHandler(BaseHandler):
                 user is too high to proceed
 
         """
-
+        if not requester.access_token_id:
+            raise ValueError("Cannot validate a user without an access token")
         if self._ui_auth_session_timeout:
             last_validated = await self.store.get_access_token_last_validated(
                 requester.access_token_id
@@ -1213,7 +1214,7 @@ class AuthHandler(BaseHandler):
     async def delete_access_tokens_for_user(
         self,
         user_id: str,
-        except_token_id: Optional[str] = None,
+        except_token_id: Optional[int] = None,
         device_id: Optional[str] = None,
     ):
         """Invalidate access tokens belonging to a user
--- a/synapse/rest/media/v1/media_repository.py
+++ b/synapse/rest/media/v1/media_repository.py
@@ -35,6 +35,7 @@ from synapse.api.errors import (
 from synapse.config._base import ConfigError
 from synapse.logging.context import defer_to_thread
 from synapse.metrics.background_process_metrics import run_as_background_process
+from synapse.types import UserID
 from synapse.util.async_helpers import Linearizer
 from synapse.util.retryutils import NotRetryingDestination
 from synapse.util.stringutils import random_string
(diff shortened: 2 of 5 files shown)


synapse/handlers/auth.py
```diff
@@ -337,7 +337,8 @@ async def validate_user_via_ui_auth(
                 user is too high to proceed
 
         """
-
+        if not requester.access_token_id:
+            raise ValueError("Cannot validate a user without an access token")
```
richvdh, line 341: I think this is unreachable, but I had to add a check to keep mypy happy.

synapse/handlers/auth.py
```diff
@@ -1213,7 +1214,7 @@ async def delete_access_token(self, access_token: str):
     async def delete_access_tokens_for_user(
         self,
         user_id: str,
-        except_token_id: Optional[str] = None,
+        except_token_id: Optional[int] = None,
```
richvdh, line 1217: this was an incorrect annotation.

synapse/rest/media/v1/media_repository.py
```diff
@@ -145,7 +146,7 @@ async def create_content(
         upload_name: Optional[str],
         content: IO,
         content_length: int,
-        auth_user: str,
+        auth_user: UserID,
```
richvdh, line 149: so was this

synapse/storage/databases/main/registration.py
```diff
@@ -1510,7 +1510,7 @@ def f(txn):
     async def user_delete_access_tokens(
         self,
         user_id: str,
-        except_token_id: Optional[str] = None,
+        except_token_id: Optional[int] = None,
```
richvdh, line 1513: and this

synapse/types.py
```diff
@@ -157,8 +156,8 @@ def deserialize(store, input):
 def create_requester(
     user_id: Union[str, "UserID"],
     access_token_id: Optional[int] = None,
-    is_guest: Optional[bool] = False,
-    shadow_banned: Optional[bool] = False,
+    is_guest: bool = False,
+    shadow_banned: bool = False,
```
richvdh, line 160: and these

synapse/types.py
```diff
@@ -83,33 +83,32 @@ class ISynapseReactor(
...
-            "access_token_id",
-            "is_guest",
-            "shadow_banned",
-            "device_id",
-            "app_service",
-            "authenticated_entity",
-        ],
-    )
-):
+@attr.s
```
clokep, line 86: Should this use slots / be frozen? (I suspect it can be frozen OK since tuples aren't able to be modified?)

synapse/types.py
clokep: Can we remove the types in the docstring while we're cleaning this up?

clokep (approved):

clokep:
And now I'm realizing you didn't request review on this. 🤦  I got very excited since I've wanted to do this forever!

synapse/types.py
richvdh: yes it should. done.

synapse/types.py
richvdh: done

Issue #8393, opened 2020-09-24
Re-Activating account fails when using SSO/SAML (with local passwords disabled)


phewitt-uw:
### Description

Using a self-hosted server, and using SAML2/SSO for authentication with local password DB is disabled.  When you deactivate an account this works, but when you try to reactivate an account via API  `_synapse/admin/v2/users` by setting "deactivated" to "false" API returns error:

{
    "errcode": "M_UNKNOWN",
    "error": "Must provide a password to re-activate an account."
}

and when you try to include a password API returns error: 

{
    "errcode": "M_FORBIDDEN",
    "error": "Password change disabled"
}

### Steps to reproduce

- enable SSO
- disable local password db in config
- activate account
- deactivate account
- then try to re-activate account using API

<!--
Describe how what happens differs from what you expected.

If you can identify any relevant log snippets from _homeserver.log_, please include
those (please be careful to remove any personal or private data). Please surround them with
``` (three backticks, on a line on their own), so that they are formatted legibly.
-->

### Version information

<!-- IMPORTANT: please answer the following questions, to help us narrow down the problem -->

<!-- Was this issue identified on matrix.org or another homeserver? -->
- **Homeserver**:

If not matrix.org:

<!--
 What version of Synapse is running?

You can find the Synapse version with this command:

$ curl http://localhost:8008/_synapse/admin/v1/server_version

(You may need to replace `localhost:8008` if Synapse is not configured to
listen on that port.)
-->
- **Version**: 1.20.0

- **Install method**: docker images from matrixdotorg
<!-- examples: package manager/git clone/pip  -->

- **Platform**: kubernetes & docker
<!--
Tell us about the environment in which your homeserver is operating
distro, hardware, if it's running in a vm/container, etc.
-->

clokep:
Looks like the code around [here](https://github.com/matrix-org/synapse/blob/6fdf5775939100121ad9e6e3a8cb21192a5444d6/synapse/rest/admin/users.py#L244-L252) needs to be updated to handle the case of an account not needing a password. I'm unsure there's a good way to check if an account is an SSO account or not, although we do have a table of external IDs to internal IDs. Maybe existing in there is a reasonable solution?

anoadragon453:
Note that this seems separate from https://github.com/matrix-org/synapse/issues/8359 in that it only affects SSO users.

localguru:
Hi, anything going on here? Or is there a workaround?

callahad:
[...]

PR #9587, opened 2021-03-10
Re-Activating account when local passwords are disabled

dklimpel:
Fixes: #8393

### Pull Request Checklist

### DIFF
--- a/synapse/rest/admin/users.py
+++ b/synapse/rest/admin/users.py
@@ -269,7 +269,10 @@ class UserRestServletV2(RestServlet):
                         target_user.to_string(), False, requester, by_admin=True
                     )
                 elif not deactivate and user["deactivated"]:
-                    if "password" not in body:
+                    if (
+                        "password" not in body
+                        and self.hs.config.password_localdb_enabled
+                    ):
                         raise SynapseError(
                             400, "Must provide a password to re-activate an account."
                         )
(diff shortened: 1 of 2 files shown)


erikjohnston (approved):
Thanks!

synapse/rest/admin/users.py
```diff
@@ -269,7 +269,10 @@ async def on_PUT(
                         target_user.to_string(), False, requester, by_admin=True
                     )
                 elif not deactivate and user["deactivated"]:
-                    if "password" not in body:
+                    if (
+                        "password" not in body
+                        and self.hs.config.password_localdb_enabled
```
clokep, line 274: I'm not sure this will catch all the instances of this, would it make more sense to check if there's a mapping from SSO auth provider to this user?

In particular, I think many admins run with passwords completely disabled, and not specifically the localdb passwords disabled.

dklimpel: IMO this not the best solution. But this is like `set_password` is implemented.
https://github.com/matrix-org/synapse/blob/1baab2035265cf2543fe3c0ef5412c1ac0740c7e/synapse/handlers/set_password.py#L37-L45

To improve it, we have also change the check in `set_password`.

clokep: After further conversation in #synapse-dev:matrix.org, I suspect that this should check `self.hs.config.password_enabled and self.hs.config.password_localdb_enabled` since that is the only time it makes sense to set a password.

Issue #9456, opened 2021-02-20
Return m.change_password enabled:false if local database not used for authentication


xyzroller:
**Description:**

I have moved my users to SSO via OIDC. The user accounts previously existed on synapse and I do not want the users to be confused by any remnants of the native synapse account. I have therefore disabled local passwords in the password_config section of synapse's homeserver.yaml - `enabled: false` in addition to configuring the oidc_provider.

However, when a user logs in to synapse from element-web or element-desktop, they are still shown UI elements for changing their (local synapse) password. Trying to do this leads to an error.

From https://github.com/vector-im/element-web/issues/16499, I learned that the clients are responding to the capabilities passed by synapse - in this case because they receive `enabled: true` for the parameter `m.change_password` in their capabilities query. 

From the Synapse Admins room on matrix, I learned that `enabled:true` is sent whenever there is a password hash present in the database (which in my case there is, as the users previously authenticated against the local DB) - see [here](https://github.com/matrix-org/synapse/blob/develop/synapse/rest/client/v2_alpha/capabilities.py#L44)

The feature request is that `enabled:false` is sent for the capability `m.change_password` whenever the config option password_config.enabled is set to `false` in homeserver.yaml.

clokep:
This seems to be confusing a few different items:

* The `m.change_password` bit is unrelated to logging in (at least clients shouldn't be using that).
* The logic for choosing which login types are supported is [near here](https://github.com/matrix-org/synapse/blob/0a00b7ff14890987f09112a2ae696c61001e6cf1/synapse/handlers/auth.py#L207-L225) -- if you have `password_config.enabled` set to `False` than password will not be accepted. (Note that there's also a `password_config.localdb_enabled` flag.)
* vector-im/element-web#16499 seems to talk about *password reset*, not logging in. It does seem that there's a bug there though that we should swap that capability based on the config in someway.

So, I guess I'm unsure if you're asking about password resets or logins or something else -- could you provide a bit more description of what is happening?

xyzroller:
Sorry for the confusion. I'll try to explain a bit better.

My overall aim is that my users, who previously used credentials stored in the Synapse database but now all sign on via SSO, no longer see any UI elements that have anything to do with the password that is (still) present in the Synapse database. The credentials (still) in the Synapse database are no longer used since I set `password_config.enabled` to `false` in `homeserver.yaml` (to "force" the users to log in via SSO).

[...]

PR #4472, opened 2019-01-25
track unstable room v3

neilisfragile:
Needs a sytest

PR #9588, opened 2021-03-10
Return m.change_password.enabled=false if local database is disabled

dklimpel:
Return `m.change_password` `enabled:false` if local database not used for authentication
Fixes: #9456

I had to remove a part of the test. Because I do not know how to create an `access_token` for an user who uses SSO.

What is the best config switch for this situation? `password_config.enabled` or `password_config.localdb_enabled`?
https://github.com/matrix-org/synapse/blob/1baab2035265cf2543fe3c0ef5412c1ac0740c7e/synapse/handlers/set_password.py#L37-L45
`set_password` uses `password_config.localdb_enabled`

https://github.com/matrix-org/synapse/blob/a7a379006651ea219c2618c0a47e8f4053a9e769/docs/sample_config.yaml#L2211-L2220
The documentation says that `localdb_enabled` is ignored if `password_config.enabled` is `false`. But I can not find the code where `localdb_enabled` becomes unset or set to `false` if `localdb_enabled` is `false`.

### Pull Request Checklist

### DIFF
--- a/changelog.d/9588.bugfix
+++ b/changelog.d/9588.bugfix
@@ -1,1 +1,1 @@
-Return `m.change_password` `enabled:false` if local database not used for authentication.
+Fix the `/capabilities` endpoint to return `m.change_password` as disabled if the local password database is not used for authentication. Contributed by @dklimpel.
--- a/changelog.d/9588.bugfix
+++ b/changelog.d/9588.bugfix
@@ -1,0 +1,1 @@
+Return `m.change_password` `enabled:false` if local database not used for authentication.
--- a/synapse/handlers/auth.py
+++ b/synapse/handlers/auth.py
@@ -215,6 +215,10 @@ class AuthHandler(BaseHandler):
         self._password_enabled = hs.config.password_enabled
         self._password_localdb_enabled = hs.config.password_localdb_enabled
 
+        self._change_password = (
+            self._password_enabled and self._password_localdb_enabled
+        )
+
         # start out by assuming PASSWORD is enabled; we will remove it later if not.
         login_types = set()
         if self._password_localdb_enabled:
@@ -884,6 +888,18 @@ class AuthHandler(BaseHandler):
                 user_infos.keys(),
             )
         return result
+
+    def get_allow_change_password(self) -> bool:
+        """Get status if users on this server are allowed to change or set a password.
+        Both `config.password_enabled` and `config.password_localdb_enabled`
+        must be true.
+        It is allowed that people add a password to their account even if they created
+        that account with SSO if it is not explicitly prohibited.
+
+        Returns:
+            Whether users on this server are allowed to change or set a password
+        """
+        return self._change_password
 
     def get_supported_login_types(self) -> Iterable[str]:
         """Get a the login types supported for the /login API
--- a/synapse/handlers/auth.py
+++ b/synapse/handlers/auth.py
@@ -215,10 +215,6 @@ class AuthHandler(BaseHandler):
         self._password_enabled = hs.config.password_enabled
         self._password_localdb_enabled = hs.config.password_localdb_enabled
 
-        self._change_password = (
-            self._password_enabled and self._password_localdb_enabled
-        )
-
         # start out by assuming PASSWORD is enabled; we will remove it later if not.
         login_types = set()
         if self._password_localdb_enabled:
@@ -889,17 +885,18 @@ class AuthHandler(BaseHandler):
             )
         return result
 
-    def get_allow_change_password(self) -> bool:
-        """Get status if users on this server are allowed to change or set a password.
-        Both `config.password_enabled` and `config.password_localdb_enabled`
-        must be true.
-        It is allowed that people add a password to their account even if they created
-        that account with SSO if it is not explicitly prohibited.
+    def can_change_password(self) -> bool:
+        """Get whether users on this server are allowed to change or set a password.
+
+        Both `config.password_enabled` and `config.password_localdb_enabled` must be true.
+
+        Note that any account (even SSO accounts) are allowed to add passwords if the above
+        is true.
 
         Returns:
             Whether users on this server are allowed to change or set a password
         """
-        return self._change_password
+        return self._password_enabled and self._password_localdb_enabled
 
     def get_supported_login_types(self) -> Iterable[str]:
         """Get a the login types supported for the /login API
--- a/synapse/rest/client/v2_alpha/capabilities.py
+++ b/synapse/rest/client/v2_alpha/capabilities.py
@@ -13,11 +13,17 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import logging
+from typing import TYPE_CHECKING, Tuple
 
 from synapse.api.room_versions import KNOWN_ROOM_VERSIONS
 from synapse.http.servlet import RestServlet
+from synapse.http.site import SynapseRequest
+from synapse.types import JsonDict
 
 from ._base import client_patterns
+
+if TYPE_CHECKING:
+    from synapse.server import HomeServer
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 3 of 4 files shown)


synapse/rest/client/v2_alpha/capabilities.py
```diff
@@ -27,21 +33,17 @@ class CapabilitiesRestServlet(RestServlet):
...
         self.auth = hs.get_auth()
         self.store = hs.get_datastore()
 
-    async def on_GET(self, request):
+    async def on_GET(self, request: SynapseRequest) -> Tuple[int, JsonDict]:
         requester = await self.auth.get_user_by_req(request, allow_guest=True)
-        user = await self.store.get_user_by_id(requester.user.to_string())
-        change_password = bool(user["password_hash"])
+        await self.store.get_user_by_id(requester.user.to_string())
+        change_password = self.config.password_enabled
```
clokep, line 46: (Similar to the comment I left on https://github.com/matrix-org/synapse/pull/9587/files#r592377951)

I'm not sure this is the right way to do this, you can have both SSO user and password users in the same server. Does it make sense to allow a user who signs in via SSO to set a password? I'm not sure.

dklimpel: I had the same question in top of this PR.
Do you know if Synapse has an function to get information of login type of an user?

clokep: There's a `get_identity_providers_for_user` method on `SsoHandler` which would return an empty dict if there's no valid IdPs to log in with. I'm just unsure if this is the proper thing to do or not. 😢

dklimpel: Ok.
I think about: Is it possible eg. login in web app with SSO and in mobile app with Synapse local db password? Can an user use SSO and password login, can it switch choose how to login?

dklimpel: I have found `_get_available_ui_auth_types` and did an update.
https://github.com/matrix-org/synapse/blob/33a02f0f52a5cd793c0d123463d8a7a1e3f4f198/synapse/handlers/auth.py#L387-L388

clokep:
I wonder if we should also disable the server endpoint to change the password if no hash is currently set? This would avoid people logging in via SSO adding a password.

Again it is unclear to me whether we *want* to support that or not though. 😢

synapse/rest/client/v2_alpha/capabilities.py
clokep: From the conversation we had in #synapse-dev:matrix.org, it seems that we actually want this to be based on is whether passwords are enabled or not (i.e. we *do* want SSO users to be able to set a password if passwords are enabled).

The set password handler only allows setting this if the local password DB is available:

https://github.com/matrix-org/synapse/blob/3101a304fc2093c2eb23e38ce2b11a853a2e0c9b/synapse/handlers/set_password.py#L47-L48

It also doesn't really make sense if passwords are completely disabled, so I think we want to keep this on:

`change_password = self.config.password_enabled and self.config.password_localdb_enabled`

This should be set in the `__init__` since it is the same for all users.

Note that this logic was originally added in #4472. That doesn't seem to mention why it is checking for the password hash though.

dklimpel:
> I wonder if we should also disable the server endpoint to change the password if no hash is currently set? This would avoid people logging in via SSO adding a password.
> 
> Again it is unclear to me whether we _want_ to support that or not though. 😢

This exactly what `_get_available_ui_auth_types` does:
https://github.com/matrix-org/synapse/blob/a7a379006651ea219c2618c0a47e8f4053a9e769/synapse/handlers/auth.py#L399-L411

synapse/handlers/auth.py
```diff
@@ -885,6 +889,18 @@ async def _find_user_id_and_pwd_hash(
...
+        """Get status if users on this server are allowed to change or set a password.
+        Both `config.password_enabled` and `config.password_localdb_enabled`
+        must be true.
+        It is allowed that people add a password to their account even if they created
+        that account with SSO if it is not explicitly prohibited.
+
+        Returns:
+            Whether users on this server are allowed to change or set a password
+        """
+        return self._change_password
```
clokep, line 902: I don't think we need the `_change_password` property here, we can return `self._password_enabled and self._password_localdb_enabled`.

synapse/handlers/auth.py
```diff
@@ -885,6 +889,18 @@ async def _find_user_id_and_pwd_hash(
             )
         return result
 
+    def get_allow_change_password(self) -> bool:
+        """Get status if users on this server are allowed to change or set a password.
+        Both `config.password_enabled` and `config.password_localdb_enabled`
+        must be true.
+        It is allowed that people add a password to their account even if they created
+        that account with SSO if it is not explicitly prohibited.
```
clokep, line 897:
```suggestion
        """Get whether users on this server are allowed to change or set a password.

        Both `config.password_enabled` and `config.password_localdb_enabled` must be true.

        Note that any account (even SSO accounts) are allowed to add passwords if the above
        is true.
```

synapse/handlers/auth.py
```diff
@@ -885,6 +889,18 @@ async def _find_user_id_and_pwd_hash(
             )
         return result
 
+    def get_allow_change_password(self) -> bool:
```
clokep, line 892: I don't love this name, but am not sure what's better, maybe only `can_change_password`? 🤷

tests/rest/client/v2_alpha/test_capabilities.py
```diff
@@ -66,12 +66,4 @@ def test_get_change_password_capabilities(self):
         capabilities = channel.json_body["capabilities"]
 
         self.assertEqual(channel.code, 200)
-
-        # Test case where password is handled outside of Synapse
```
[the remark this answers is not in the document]
dklimpel: I had a try. If you set `"localdb_enabled": False` it is not possible to login an user with password. But this API needs authentification. I do not know how to create a token for an user without password.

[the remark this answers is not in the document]
dklimpel: I do a second try with this function:
https://github.com/matrix-org/synapse/blob/f87dfb94034087f5392670b6e857600102ea3f39/synapse/handlers/auth.py#L769-L775

changelog.d/9588.bugfix
```diff
@@ -0,0 +1 @@
+Return `m.change_password` `enabled:false` if local database not used for authentication.
```
clokep, line 1: I think the following is clearer:

```suggestion
Fix the `/capabilities` endpoint to return `m.change_password` as disabled if the local password database is not used for authentication. Contributed by @dklimpel.
```

clokep (approved):
Looks good! Thanks for fixing this!

xrayn:
IMO it would have been better to introduce a new configuration option for a possible password change and not make this solely dependent from the setting `localdb_enabled`. Ideal would be an option where one can disable password changes, but can still use  the local db for authentication.

Because now, it is no longer possible to mix local users and a SSO (or LDAP, ...).

This might not be a concern for many people, but actually one can no longer add local users for testing or local users for bots an alike. For small deployments this is no concern at all, but for bigger deployments, adding users to the identity management system of a company or institution might be complicated (or nearly impossible) in terms of operational procedures.

clokep:
@xrayn This code is pretty nuanced and I believe there could be bugs when having both local and SSO passwords, if you're seeing one please file a separate issue with details of your setup.

I don't see how this change would preclude having both local users and SSO (or LDAP) users.

xrayn:
> I don't see how this change would preclude having both local users and SSO (or LDAP) users.

Perhaps this is a misunderstanding, but I assume that when setting `localdb_enabled` to `false` then a user can no longer login with credentials from the localdb right?

clokep:
> > I don't see how this change would preclude having both local users and SSO (or LDAP) users.
> 
> Perhaps this is a misunderstanding, but I assume that when setting `localdb_enabled` to `false` then a user can no longer login with credentials from the localdb right?

Yes, that's the point of that setting -- it disables the local database. If you don't have a local database then there's no way for a user to change their password, which this PR fixes.

PR #9590, opened 2021-03-11
Add logging for redis connection setup

richvdh:
This makes it write logs like:

```
2021-03-11 12:29:02,800 - synapse.replication.tcp.redis - 264 - INFO - None - Connecting to redis server localhost:6379
2021-03-11 12:29:02,802 - synapse.replication.tcp.redis - 270 - INFO - None - Connection to redis server localhost:6379 failed: Connection was refused by other side: 111: Connection refused.
```

It's an attempt to get more insight into what is happening with redis connections, as well as mitigating the issue where requests mysteriously block when the redis server is down.

### DIFF
--- a/changelog.d/9590.misc
+++ b/changelog.d/9590.misc
@@ -1,0 +1,1 @@
+Add logging for redis connection setup.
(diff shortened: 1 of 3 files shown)


clokep (approved):
Seems reasonable to me. 👍  (Minus CI failing.)

PR #4358, opened 2019-01-07
Refactor request sending to have better excpetions

erikjohnston:
MatrixFederationHttpClient blindly reraised exceptions to the caller
without differentiating "expected" failures (e.g. connection timeouts
etc) versus more severe problems (e.g. programming errors).

This commit adds a RequestSendFailed exception that is raised when
"expected" failures happen, allowing the TransactionQueue to log them as
warnings while allowing us to log other exceptions as actual exceptions.

This should help with #4239, as we can now differentiate between "expected" errors and unexpected ARGH LOG THE WORLD errors

PR #9597, opened 2021-03-11
Reject concurrent transactions

### LINKED ISSUE #9489: Incoming federation transactions stack up indefinitely
Since #8342, we apply the federation ratelimiter *after* the transaction linearizer. The net effect is that, if we cannot keep up with incoming transactions, we allow them to stack up indefinitely (there can only be one thread inside the transaction linearizer, so we never hit the concurrency limit on the ratelimiter).  This can use a lot of memory, and it's not the way matrix federation transactions are meant to work (if the remote gets behind, it should drop messages until it catches up).

It feels like the ratelimter is pretty redundant there. (Admittedly it still has some use as a true *rate* limiter rather than a concurrency limiter, but that's a bit of an edge-case.)

Should we instead be using a separate ratelimiter dedicated to incoming transactions?

richvdh:
If more transactions arrive from an origin while we're still processing the first one, reject them.

Hopefully fixes #9489

### DIFF
--- a/synapse/federation/federation_server.py
+++ b/synapse/federation/federation_server.py
@@ -206,7 +206,7 @@ class FederationServer(FederationBase):
         # CRITICAL SECTION: the first thing we must do (before awaiting) is
         # add an entry to _active_transactions.
         assert origin not in self._active_transactions
-        self._active_transactions[origin] = transaction.transaction_id
+        self._active_transactions[origin] = transaction.transaction_id  # type: ignore
 
         try:
             result = await self._handle_incoming_transaction(
--- a/synapse/federation/federation_server.py
+++ b/synapse/federation/federation_server.py
@@ -170,6 +170,13 @@ class FederationServer(FederationBase):
 
         logger.debug("[%s] Got transaction", transaction_id)
 
+        # Reject malformed transactions early: reject if too many PDUs/EDUs
+        if len(transaction.pdus) > 50 or (  # type: ignore
+            hasattr(transaction, "edus") and len(transaction.edus) > 100  # type: ignore
+        ):
+            logger.info("Transaction PDU or EDU count too large. Returning 400")
+            return 400, {}
+
         # we only process one transaction from each origin at a time. We need to do
         # this check here, rather than in _on_incoming_transaction_inner so that we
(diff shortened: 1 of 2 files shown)


synapse/federation/federation_server.py
```diff
@@ -169,6 +170,33 @@ async def on_incoming_transaction(
 
         logger.debug("[%s] Got transaction", transaction_id)
 
+        # Reject malformed transactions early: reject if too many PDUs/EDUs
+        if len(transaction.pdus) > 50 or (  # type: ignore
+            hasattr(transaction, "edus") and len(transaction.edus) > 100  # type: ignore
+        ):
+            logger.info("Transaction PDU or EDU count too large. Returning 400")
+            return 400, {}
```
richvdh, line 178: I moved this up here from `_handle_incoming_transaction` to keep a sytest (which was sending concurrent transactions) happy, but it seems like generally a good thing to do.

synapse/federation/federation_server.py
```diff
@@ -112,10 +112,11 @@ def __init__(self, hs: "HomeServer"):
         # with FederationHandlerRegistry.
         hs.get_directory_handler()
 
-        self._federation_ratelimiter = hs.get_federation_ratelimiter()
```
clokep, line 115: I wonder if we should just inline the rate-limiter that used to be shared with this into `TransportLayerServer` now...

clokep (approved):
Seems to make sense. It definitely makes the code easier to reason about!

Synapse has always retried federation transactions on a 429, correct? This won't break old servers federating with newer servers?

synapse/federation/federation_server.py
clokep: I agree that it seems best to do this as early as possible. 👍

richvdh:
> Synapse has always retried federation transactions on a 429, correct? 

looks like it might have changed in #4358, but I don't think it's going to break anything too badly anyway. Worst case the sender will back off transmission for a bit, which is fine.

synapse/federation/federation_server.py
richvdh: :man_shrugging: probably...

richvdh:
> looks like it might have changed in #4358

actually looks like we retried for all response codes before that.

PR #9609, opened 2021-03-12
Add SSO attribute requirements for OIDC providers

HubbeKing:
This way, administrators may require certain attributes be part of OIDC userinfo in order to allow login/registration using the OIDC SSO Provider.

This lets one limit SSO logins to certain groups, by adding group memberships to OIDC userinfo - in Keycloak, this can be done with a Group Membership mapper.

I tested this with my own Keycloak instance, using an openid-connect client, and it behaved as expected. Unit tests and integration tests also pass, so it appears to not have broken anything.

Additionally, I added a simple unit test for the new feature.
Signed-off-by: Hubbe King <github@hubbe.club>


clokep:
Looks like a pretty good start! It seems the sample config needs to be re-generated with `scripts-dev/generate_sample_config`, also looks like there's a couple of mypy errors (you can run this yourself via `scripts-dev/lint.sh`. The newsfragment just needs a `.` on the end. 😄

HubbeKing:
Should be good now, found an f-string I left in there that broke the py3.5 CI builds, changed it to the older %s style, so it's consistent with the rest of the file as well.

synapse/config/oidc_config.py
```diff
@@ -190,6 +191,19 @@ def generate_config_section(self, config_dir_path, server_name, **kwargs):
         #           When rendering, the Jinja2 templates are given a 'user' variable,
         #           which is set to the claims returned by the UserInfo Endpoint and/or
         #           in the ID Token.
+        #   It is possible to configure Synapse to only allow logins if certain attributes
+        #   match particular values in the OIDC userinfo. The requirements can be listed under
+        #   `attribute_requirements` as shown below. All of the listed attributes must
+        #   match for the login to be permitted. Note that fields listed may need to be added
+        #   to the userinfo return in the OIDC provider, by extending or adding a scope.
+        #   If the attributes listed are not part of the default scopes, these scopes may
+        #   also need to be added to the `scopes` section of the OIDC config.
```
clokep, line 200: I think this somewhat repeats the same thing twice, maybe this is a bit clearer?

```suggestion
        #   `attribute_requirements` as shown below. All of the listed attributes must
        #   match for the login to be permitted. Additional attributes can be added to
        #   userinfo by expanding the `scopes` section of the OIDC config to retrieve
        #   additional information from the OIDC provider.
```

synapse/handlers/oidc_handler.py
```diff
@@ -278,6 +288,9 @@ def __init__(
         self._config = provider
         self._callback_url = hs.config.oidc_callback_url  # type: str
 
+        self._oidc_attribute_requirements = (
+            provider.attribute_requirements
+        )  # type: Iterable[SsoAttributeRequirement]
```
clokep, line 293: I don't think you need to provide the type here.

synapse/handlers/oidc_handler.py
```diff
@@ -854,6 +867,15 @@ async def handle_oidc_callback(
             )
 
         # otherwise, it's a login
+        logger.debug("Userinfo for OIDC login: %s", userinfo)
+
+        # Ensure that the attributes of the logged in user meet the required
+        # attributes
+        # With OIDC, we check userinfo against attribute_requirements
```
clokep, line 874:
```suggestion
        # Ensure that the attributes of the logged in user meet the required
        # attributes by checking the userinfo against attribute_requirements.
```

synapse/config/oidc_config.py
```diff
@@ -568,3 +598,6 @@ class OidcProviderConfig:
 
     # the config of the user mapping provider
     user_mapping_provider_config = attr.ib()
+
+    # required attributes to require in userinfo to allow login/registration
+    attribute_requirements = attr.ib(type=Iterable[SsoAttributeRequirement])
```
clokep, line 603: I think this can be defined as a list here and that's fine (a list is an iterable):

```suggestion
    attribute_requirements = attr.ib(type=List[SsoAttributeRequirement])
```

synapse/handlers/oidc_handler.py
```diff
@@ -854,6 +867,15 @@ async def handle_oidc_callback(
...
+        if not self._sso_handler.check_required_attributes(
+            request, userinfo, self._oidc_attribute_requirements
+        ):
```
clokep, line 877: I'm not sure this is quite the same, `check_required_attributes` expects the attributes to be a mapping of attribute names (as strings) to a list of values, I *think* `userinfo` is a mapping of attribute name to a single value.

I think the code around https://github.com/matrix-org/synapse/blob/df425c2c63d969778f725c9053dd93e57ee854e1/synapse/handlers/sso.py#L977-L979 will break pretty badly if you do things like require a value of `foo` and the actual value is `foobar`.

I think the solution is to pass in `{k: [v] for k, v in userinfo.items()}` to `check_required_attributes`.

clokep (changes requested):
Looks good overall! I left a few comments though. Thanks for working on this!

synapse/handlers/oidc_handler.py
HubbeKing: Yeah. The behavior here gets a bit odd - userinfo here is a dict, with keys being strings and values being parsed from JSON data types to python data types - so it's a mix of strings, booleans, integers, lists, and whatever other arbitrary elements have been added to the scopes for the oidc provider. 

Doing `{k: [v] for k, v in userinfo.items()}` would ensure that things like strings are handled as exact matches, but I'm unsure how `if req.value in values` acts if `values` is a list containing a list when `req.value` is only ever going to be a string, since that's what `SsoRequirement.JSON_SCHEMA` demands.

Ideally, I'd like it to support behavior like "userinfo groups contains `foobar` when groups is a list of strings", but I'm not sure how to cleanly handle that.

As it is right now, it's always a loose match - `values` here will be an integer, a string, a boolean, or a list, and `req.value` will be a string. If `values` is a string and `req.value` is a substring of it, the check passes. If `values` is a list, and it contains the string that `req.value` is, the check passes. In all other cases, the check fails.
This is perhaps okay, but it is a bit sloppy, and the documentation as written doesn't explain it very well.

Perhaps I could expand `SsoRequirement.JSON_SCHEMA` such that `req.value` here can be things other than a string, and add an `operator` field to it to support both the exact and contains style of matching? That would require changes to the CAS and SAML handlers as well though.
But that would allow `attribute_requirements` to support both the "attribute must match X" behavior it currently does in SAML and CAS, and a new "attribute must contain X" behavior, which would be ideal for things like checking group membership.
`groups` is a fairly common OIDC userinfo attribute, after all.

synapse/config/oidc_config.py
```diff
@@ -190,6 +191,19 @@ def generate_config_section(self, config_dir_path, server_name, **kwargs):
...
+        #   match for the login to be permitted. Note that fields listed may need to be added
+        #   to the userinfo return in the OIDC provider, by extending or adding a scope.
+        #   If the attributes listed are not part of the default scopes, these scopes may
+        #   also need to be added to the `scopes` section of the OIDC config.
+        #
+        #   attribute_requirements:
+        #     - attribute: family_name
+        #       value: "Stephensson"
+        #     - attribute: groups
+        #       value: "admin"
```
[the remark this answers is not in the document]
HubbeKing: I think it may be a good idea to stick to the default OIDC `profile` scope claims in the examples, rather than having parity with the SAML and CAS examples.
https://openid.net/specs/openid-connect-core-1_0.html#ScopeClaims

I picked `family_name` because I felt that was a reasonably common use case, restricting logins to one's family when running a homeserver, for instance. `groups` is used a lot when integrating OIDC providers with things like `oauth2-proxy`, but importantly `groups` is then a list of strings, which I have Opinions on how best to handle... (see below)

synapse/handlers/oidc_handler.py
clokep: Would groups be returned as a string or a list of strings? (Or is that up to the implementation to decide?)

I'll need to think a bit more about this.

HubbeKing: It should be up to the OIDC provider - as far as I'm aware the OpenID Connect spec only requires that the userinfo response is valid JSON, and that it must contain a `sub` value, and that it must not contain any empty strings or null values.

I believe most OIDC providers handle it as a list of strings, though.

synapse/config/oidc_config.py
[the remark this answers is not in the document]
clokep: Ah! I did not realize that is where these came from. 😄 The reasoning makes sense.

HubbeKing:
As we discussed in the matrix room, I made the check wrap non-list values in userinfo in lists.

It may additionally be prudent to cast non-string values that are in lists in userinfo to strings, since the `attribute_requirements` config can only contain string values, but that may be getting a bit too into the nitty gritty.

changelog.d/9609.feature
```diff
@@ -0,0 +1 @@
+Added SsoAttributeRequirements to OpenID Connect-based SSO providers. Contributed by Hubbe King.
```
clokep, line 1: Better if we phrase this in a way that someone not familiar with the code can understand:

```suggestion
Logins using OpenID Connect can require attributes on the `userinfo` response in order to login. Contributed by Hubbe King.
```

synapse/config/oidc_config.py
```diff
@@ -191,6 +192,27 @@ def generate_config_section(self, config_dir_path, server_name, **kwargs):
...
+        #   userinfo by expanding the `scopes` section of the OIDC config to retrieve
+        #   additional information from the OIDC provider.
+        #
+        #   An important thing to note here, is that OIDC claims which exist in the userinfo
+        #   as a string will be strictly matched - the attribute matches only if the userinfo
+        #   value exactly matches the configured value in `attribute_requirements`.
+        #   However, if the claim exists in the userinfo as a list, the attribute matches
+        #   if the list contains the value configured in `attribute_requirements`.
+        #   i.e. `family_name` in the userinfo MUST be "Stephensson" in the example below
+        #   i.e. `groups` in the userinfo MUST contain "admin" in the example below
```
clokep, line 208: Does this seem a bit simpler, but still accurate?

```suggestion
        #   If the OIDC claim is a list, then the attribute must match any value in the list.
        #   Otherwise, it must exactly match the value of the claim. Using the example
        #   below, the `family_name` claim MUST be "Stephensson", but the `groups`
        #   claim MUST contain "admin".
```

clokep:
I think this looks good minus the one question I left above!

synapse/config/oidc_config.py
HubbeKing: Yeah, that's less wordy but still entirely accurate. Seems good!

clokep: Woot! Thanks for making this change!

clokep (approved):
LGTM if CI passes! Thanks for working on this! 🎉

enekonieto:
If anyone is looking for the docs about this feature I found them in the [sample config](https://github.com/matrix-org/synapse/blob/develop/docs/sample_config.yaml#L1915).

PR #9612, opened 2021-03-13
Include opencontainers labels in Docker image

### LINKED ISSUE #9611: Include opencontainers labels in docker images
**Description:**

It would be nice if the official docker image of synapse included [opencontainers-compatible labels](https://github.com/opencontainers/image-spec/blob/master/annotations.md). These can, for instance, be used by [renovate](https://github.com/renovatebot/renovate) to automatically fetch the changelog associated with new releases.

languitar:
Cf. https://github.com/opencontainers/image-spec/blob/master/annotations.md#pre-defined-annotation-keys

Fixes #9611

Ideally, `org.opencontainers.image.version` would be set in CircleCI. But I am not entirely certain how the release workflow is.

### Pull Request Checklist

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -19,7 +19,7 @@ ARG PYTHON_VERSION=3.8
 FROM docker.io/python:${PYTHON_VERSION}-slim as builder
 
 LABEL org.opencontainers.image.url='https://matrix.org/docs/projects/server/synapse'
-LABEL org.opencontainers.image.documentation='https://github.com/matrix-org/synapse/blob/master/INSTALL.md'
+LABEL org.opencontainers.image.documentation='https://github.com/matrix-org/synapse/blob/master/docker/README.md'
 LABEL org.opencontainers.image.source='https://github.com/matrix-org/synapse.git'
 LABEL org.opencontainers.image.licenses='Apache-2.0'
 
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -18,6 +18,11 @@ ARG PYTHON_VERSION=3.8
 ###
 FROM docker.io/python:${PYTHON_VERSION}-slim as builder
 
+LABEL org.opencontainers.image.url='https://matrix.org/docs/projects/server/synapse'
+LABEL org.opencontainers.image.documentation='https://github.com/matrix-org/synapse/blob/master/INSTALL.md'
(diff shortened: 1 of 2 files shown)


docker/Dockerfile
```diff
@@ -18,6 +18,11 @@ ARG PYTHON_VERSION=3.8
 ###
 FROM docker.io/python:${PYTHON_VERSION}-slim as builder
 
+LABEL org.opencontainers.image.url='https://matrix.org/docs/projects/server/synapse'
+LABEL org.opencontainers.image.documentation='https://github.com/matrix-org/synapse/blob/master/INSTALL.md'
```
erikjohnston, line 22: I'm not 100% sure if these are the URLs we want to use? Do other people have opinions?

erikjohnston (commented):
Otherwise I think this looks like a good thing!

docker/Dockerfile
languitar: I wasn't sure neither. Finding something that looks official using Google isn't exactly easy.

languitar:
> Otherwise I think this looks like a good thing!

Do you think I should also include a circle ci integration to inject the version in the appropriate label in the PR, too?

docker/Dockerfile
anoadragon453: We may want to link to the docker README instead of the general INSTALL.md guide for the docker containers?

https://matrix.org/docs/projects/server/synapse is probably fine, especially since we have a link to the github in another line. We could probably do to make that page a bit prettier though.

languitar:
```suggestion
LABEL org.opencontainers.image.url='https://matrix.org/docs/projects/server/synapse'
LABEL org.opencontainers.image.documentation='https://github.com/matrix-org/synapse/tree/master/docker'
```

erikjohnston:
> > Otherwise I think this looks like a good thing!
> 
> Do you think I should also include a circle ci integration to inject the version in the appropriate label in the PR, too?

Let's merge this as is for now, but if you're keen then adding the version injection as a new PR would be welcomed :)

erikjohnston (approved):

languitar:
Am I right that this should be included in v1.31.0? Because there I can't find any of these labels on the published image.

languitar:
Yikes, I already see the problem. The labels are attached to the builder, not the final image :/ I will open a new MR to fix this.

PR #9615, opened 2021-03-15
Remove state events from /admin/v1/rooms/<room_id>/context/<event_id>

Yoric:
Real-world tests show that this API is too slow (and timeouts) because serialization
of state events can be extremely slow when there are too many state events, e.g.
in irc rooms.

### Pull Request Checklist

### DIFF
--- a/changelog.d/9615.removal
+++ b/changelog.d/9615.removal
@@ -1,0 +1,1 @@
+Remove state events from /_synapse/admin/v1/rooms/<room_id>/context/<event_id> for performance reasons.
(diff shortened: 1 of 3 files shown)


anoadragon453:
Heads up that the performance issues may be addressed by https://github.com/matrix-org/synapse/pull/9619. Plus @erikjohnston expressed concern that a Context API is supposed to include state events.

clokep:
I think this can be closed since #9619 fixed the root cause? Please re-open if this is incorrect.

PR #9616, opened 2021-03-15
Prevent bundling aggregations for state events

anoadragon453:
We're trying this out on hotfixes to see whether it helps address
performance issues. There's no need to do aggregation bundling for
state events anyways.

Note: mypy CI failure is unrelated.


erikjohnston:
> We're trying this out on hotfixes to see whether it helps address
> performance issues. There's no need to do aggregation bundling for
> state events anyways.

If its not necessary then let's just land it on mainline?

anoadragon453:
Closed in favour of a mainline implementation: #9619

PR #9618, opened 2021-03-15
Revert requiring a specific version of Twisted for mypy checks.

clokep:
This removes the pin on Twisted in the tox mypy check (originally added in #9515.

### DIFF
--- a/changelog.d/9618.misc
+++ b/changelog.d/9618.misc
@@ -1,0 +1,1 @@
+Fix incorrect type hints.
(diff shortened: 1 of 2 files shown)


clokep:
I think that this will work fine now to run `tox -e mypy` repeatedly, although running `mypy` manually (or via `scripts-dev/lint.sh`) is still broken: https://github.com/matrix-org/synapse/issues/9513#issuecomment-797672040

Any thoughts on what to do about this?

erikjohnston (approved):
I don't know what we do about running `mypy` directly, but that shouldn't stop us from running it in CI

PR #9619, opened 2021-03-15
Prevent bundling aggregations for state events

anoadragon453:
There's no need to do aggregation bundling for state events. Doing so can cause performance issues.

cc @Yoric


clokep:
Is there any concern about removing this from the C-S API will break clients? The spec didn't seem to say whether or not they should be included.

anoadragon453:
I *suspect* it's not intended for state events to have aggregation information added to them, since they already aggregate by replacing past state with the same (state_type, state_key).

erikjohnston:
> I _suspect_ it's not intended for state events to have aggregation information added to them, since they already aggregate by replacing past state with the same (state_type, state_key).

We don't bundle up aggregations in things like `/state` requests, so I think its fine.

changelog.d/9619.bugfix
```diff
@@ -0,0 +1 @@
+Prevent attempting to bundle aggregations for state events in /context APIs.
```
erikjohnston, line 1: This isn't really a bug? And if it was we should say what the effect was here.

I believe this is more about improving performance?

erikjohnston (approved):

changelog.d/9619.bugfix
anoadragon453: Good point, I found it a bit ambiguous but I think `misc` does make a bit more sense.

PR #9620, opened 2021-03-15
Handle an empty cookie as an invalid macaroon.

clokep:
Fixes https://sentry.matrix.org/sentry/synapse-matrixorg/issues/199800/

It seems that the cookie retrieved is empty, causing the pymacaroon library to raise a different exception than we expect.

### DIFF
--- a/changelog.d/9620.bugfix
+++ b/changelog.d/9620.bugfix
@@ -1,0 +1,1 @@
+Fix a bug introduced in v1.28.0 where the OpenID Connect callback endpoint could error with a `MacaroonInitException`.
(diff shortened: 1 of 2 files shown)


erikjohnston (approved):

richvdh:
I do wonder why the cookie is empty; but still, clients presenting weird data shouldn't cause exceptions.

richvdh:
merging ready for synapse 1.30

clokep:
Thanks! 🎉 

I have no idea why it would return an empty cookie. I debated treating the empty cookie the same as not having a cookie, but that behavior didn't seem right. 🤷

PR #9621, opened 2021-03-15
Only send catch up events if they're the latest in the room

erikjohnston:
This is a bit of a hack to help mitigate #9492. One of the problems there is that in a busy with lots of servers, each server will send the last event *they* sent, which results in the server that has come back online getting a smattering of events rather than just the latest event its missed. The flip side is that it relies on whatever server that did send the last event in the room actually sending that to the newly up server.

I'm not really sure if this is what we want to do. Other thing we could do is add more metadata to the sent event to help the remote Do The Right Thing, or simply send the current forward extremities of the room to them (though currently servers drop events they get via `/send` from other servers).

Thoughts?

### DIFF
--- a/changelog.d/9621.misc
+++ b/changelog.d/9621.misc
@@ -1,0 +1,1 @@
+Only send catch up events if they're the latest in the room, reducing the number of events a server receives after recovering from downtime.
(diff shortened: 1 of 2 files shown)


erikjohnston:
> or simply send the current forward extremities of the room to them (though currently servers drop events they get via `/send` from other servers).

After some talk in #synapse-dev:matrix.org, I'll give this a go

Issue #9572, opened 2021-03-09
Let the Spam Checker know whether the registration came from an SSO provider


lampholder:
Right now we call `spam_checker.check_registration_for_spam` with both native, local, Synapse registrations _and_ with the "first login" of SSO users.

If a homeserver is running with both native and SSO-provided logins, it'd be good to expose to the spam checker the nature of the registration - perhaps by passing in `auth_provider_id` to the `check_registration_for_spam` call?

anoadragon453:
@richvdh how would you feel about passing `auth_provider_id` to SpamChecker modules? Or should we try to pass some sort of opaque constant instead (AS vs. SSO vs. Native)?

richvdh:
seems reasonable enough to me

PR #9626, opened 2021-03-16
Pass SSO IdP information to spam checker's registration function

anoadragon453:
Fixes https://github.com/matrix-org/synapse/issues/9572

When a SSO user logs in for the first time, we create a local Matrix user for them. This goes through the register_user flow, which ends up triggering the spam checker. Spam checker modules don't currently have any way to differentiate between a user trying to sign up initially, versus an SSO user (whom has presumably already been approved elsewhere) trying to log in for the first time.

This PR passes `auth_provider_id` as an argument to the `check_registration_for_spam` function. This argument will contain an ID of an SSO provider (`"saml"`, `"cas"`, etc.) if one was used, else `None`.

### DIFF
--- a/synapse/events/spamcheck.py
+++ b/synapse/events/spamcheck.py
@@ -212,14 +212,21 @@ class SpamChecker:
             # spam checker
             checker = getattr(spam_checker, "check_registration_for_spam", None)
             if checker:
-                behaviour = await maybe_awaitable(
-                    checker(
-                        email_threepid,
-                        username,
-                        request_info,
-                        auth_provider_id=auth_provider_id,
+                checker_args = inspect.getfullargspec(checker)
+                # Provide auth_provider_id if the function supports it
+                if "auth_provider_id" in checker_args.args:
+                    behaviour = await maybe_awaitable(
+                        checker(
+                            email_threepid,
+                            username,
+                            request_info,
+                            auth_provider_id,
+                        )
                     )
-                )
+                else:
+                    behaviour = await maybe_awaitable(
+                        checker(email_threepid, username, request_info)
+                    )
                 assert isinstance(behaviour, RegistrationBehaviour)
                 if behaviour != RegistrationBehaviour.ALLOW:
                     return behaviour
--- a/synapse/events/spamcheck.py
+++ b/synapse/events/spamcheck.py
@@ -15,6 +15,7 @@
 # limitations under the License.
 
 import inspect
+import logging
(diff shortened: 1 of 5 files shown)


synapse/events/spamcheck.py
```diff
@@ -208,9 +212,21 @@ async def check_registration_for_spam(
             # spam checker
             checker = getattr(spam_checker, "check_registration_for_spam", None)
             if checker:
-                behaviour = await maybe_awaitable(
-                    checker(email_threepid, username, request_info)
-                )
+                checker_args = inspect.getfullargspec(checker)
+                # Provide auth_provider_id if the function supports it
+                if "auth_provider_id" in checker_args.args:
```
richvdh, line 217: should we not look at whether it has at least four positional args rather than what those args are called?

synapse/events/spamcheck.py
```diff
@@ -208,9 +212,21 @@ async def check_registration_for_spam(
             # spam checker
             checker = getattr(spam_checker, "check_registration_for_spam", None)
             if checker:
-                behaviour = await maybe_awaitable(
-                    checker(email_threepid, username, request_info)
-                )
+                checker_args = inspect.getfullargspec(checker)
```
richvdh, line 215: I think really you want `inspect.signature` here rather than `getfullargspec`. (see https://docs.python.org/3/library/inspect.html#inspect.getfullargspec)

synapse/events/spamcheck.py
```diff
@@ -208,9 +212,21 @@ async def check_registration_for_spam(
...
+                            email_threepid,
+                            username,
+                            request_info,
+                            auth_provider_id,
+                        )
+                    )
+                else:
+                    behaviour = await maybe_awaitable(
+                        checker(email_threepid, username, request_info)
+                    )
```
richvdh, line 229: it might be neater to factor the `await maybe_awaitable(...)` out of the if condtion?

richvdh (changes requested):
looks generally good, though a few nits

synapse/events/spamcheck.py
anoadragon453: This does make more sense if we add more arguments in the future. Thanks!

synapse/events/spamcheck.py
anoadragon453: Oh, good to know that `getfullargspec` is on the way out. I suppose we should also update the other users of it in the codebase.

richvdh (approved):
lgtm

PR #9631, opened 2021-03-16
Add type hints to the room member handler.

clokep:
This is spun out of #9374 as that got messy. This adds a return type to `get_room_member_handler`, which enables additional type checking.

As part of this, it complained that `forget` didn't exist on the worker version. I implemented a stub which just raises an error, which seems "wrong", but at least makes it explicit. Instead we could make a call over replication to do this. Is that the right thing to do? (Note that the endpoints that use this don't have any "GET" method so I'm not sure there's benefit to running this on workers which just call back to main. Beyond ease of deployment, that is.)

### DIFF
--- a/changelog.d/9631.misc
+++ b/changelog.d/9631.misc
@@ -1,0 +1,1 @@
+Add additional type hints to the Homeserver object.
--- a/synapse/handlers/register.py
+++ b/synapse/handlers/register.py
@@ -437,10 +437,10 @@ class RegistrationHandler(BaseHandler):
 
                 if RoomAlias.is_valid(r):
                     (
-                        room_id,
+                        room,
                         remote_room_hosts,
                     ) = await room_member_handler.lookup_room_alias(room_alias)
-                    room_id = room_id.to_string()
+                    room_id = room.to_string()
                 else:
                     raise SynapseError(
                         400, "%s was not legal room ID or room alias" % (r,)
--- a/synapse/handlers/room_member.py
+++ b/synapse/handlers/room_member.py
@@ -155,6 +155,10 @@ class RoomMemberHandler(metaclass=abc.ABCMeta):
         """
         raise NotImplementedError()
 
+    @abc.abstractmethod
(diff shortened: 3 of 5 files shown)


erikjohnston:
> As part of this, it complained that `forget` didn't exist on the worker version. I implemented a stub which just raises an error, which seems "wrong", but at least makes it explicit. Instead we could make a call over replication to do this. Is that the right thing to do? (Note that the endpoints that use this don't have any "GET" method so I'm not sure there's benefit to running this on workers which just call back to main. Beyond ease of deployment, that is.)

It's not ideal, but I think its fine for now. The best would be that any classes that required `forget` explicitly asked for the master handler, which would then fail on startup if it was done on a worker instead. That can end up being a bit fiddly and involve breaking up classes and things though

clokep:
> The best would be that any classes that required `forget` explicitly asked for the master handler, which would then fail on startup if it was done on a worker instead.

I'm guessing this is why there's a couple of handlers that do only one thing?

The other option is to add a replication API for it, which in this case is easy. 😄 I'm just not sure it makes sense.

erikjohnston (approved):

clokep:
(I should mention that I'm looking for some more general guidance as the other changes in #9374 will likely require similar changes. Having them raise runtime exceptions might be a good first step.)

erikjohnston:
Forgot to hit send on the review button :facepalm: 



> I'm guessing this is why there's a couple of handlers that do only one thing?

Yeah, probably. Though sometimes we just like making handlers :)

> The other option is to add a replication API for it, which in this case is easy.  I'm just not sure it makes sense.

If its never used its just dead code that is just going to break, so yeah, we can do that, but we should then also make it possible to run the associated API on workers

clokep:
> > The other option is to add a replication API for it, which in this case is easy.  I'm just not sure it makes sense.
> 
> If its never used its just dead code that is just going to break, so yeah, we can do that, but we should then also make it possible to run the associated API on workers

Sorry -- my assumption was that if we made a replication API we'd make it run on workers!

PR #9634, opened 2021-03-16
only save remote cross-signing keys if they're different from the current ones

uhoreg:
There seems to be a feedback loop where, if a remote user has a master signing key but no self-signing key, synapse will try to re-fetch that user's keys, which it will blindly treat as a device update, which will notify local users about the device change, which will cause them to fetch the remote user's keys, which will cause synapse to notice that they have a master key but no self-signing key, and will try to re-fetch again.  This change will only update the local cached keys after refetching the remote user's keys if anything actually changed, which breaks the loop.

### DIFF
--- a/changelog.d/9634.misc
+++ b/changelog.d/9634.misc
@@ -1,2 +1,1 @@
-Only save remote cross-signing and device keys if they're different from the
-current ones.
+Only save remote cross-signing and device keys if they're different from the current ones.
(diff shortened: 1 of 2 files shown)


erikjohnston (approved):

changelog.d/9634.misc
```diff
@@ -0,0 +1,2 @@
+Only save remote cross-signing and device keys if they're different from the
+current ones.
```
clokep, line 2: Changelogs should be one line:

```suggestion
Only save remote cross-signing and device keys if they're different from the current ones.
```

PR #8362, opened 2020-09-20
Fixed a bug with reactivating users with the admin API

dklimpel:
Fixes: #8359
Trying to reactivate a user with the admin API (`PUT
/_synapse/admin/v2/users/<user_name>`) causes an internal server error.

Seems to be a regression in #8033

### Pull Request Checklist

* [x] Pull request is based on the develop branch
* [x] Pull request includes a [changelog file](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md#changelog). The entry should:
  - Be a short description of your change which makes sense to users. "Fixed a bug that prevented receiving messages from other servers." instead of "Moved X method from `EventStore` to `EventWorkerStore`.".
  - Use markdown where necessary, mostly for `code blocks`.
  - End with either a period (.) or an exclamation mark (!).
  - Start with a capital letter.
* [x] Pull request includes a [sign off](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md#sign-off)
* [x] Code style is correct (run the [linters](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md#code-style))

Signed-off-by: Dirk Klimpel dirk@klimpel.org

PR #9636, opened 2021-03-16
Update the check whether a password may be set

dklimpel:
Follow up of #9587 
Based on #9588

Check `password_config.enabled` and `password_config.localdb_enabled` before set a new password.
Add unit tests to admin API
Add needed cache invalidation

### Pull Request Checklist

### DIFF
--- a/changelog.d/9636.bugfix
+++ b/changelog.d/9636.bugfix
@@ -1,0 +1,1 @@
+Checks if passwords are allowed before setting it for the user.
--- a/synapse/storage/databases/main/registration.py
+++ b/synapse/storage/databases/main/registration.py
@@ -1210,9 +1210,7 @@ class RegistrationBackgroundUpdateStore(RegistrationWorkerStore):
         self._invalidate_cache_and_stream(
             txn, self.get_user_deactivated_status, (user_id,)
         )
-        self._invalidate_cache_and_stream(
-            txn, self.get_user_by_id, (user_id,)
-        )
+        self._invalidate_cache_and_stream(txn, self.get_user_by_id, (user_id,))
         txn.call_after(self.is_guest.invalidate, (user_id,))
 
     @cached()
--- a/synapse/storage/databases/main/registration.py
+++ b/synapse/storage/databases/main/registration.py
@@ -1210,6 +1210,9 @@ class RegistrationBackgroundUpdateStore(RegistrationWorkerStore):
         self._invalidate_cache_and_stream(
             txn, self.get_user_deactivated_status, (user_id,)
         )
+        self._invalidate_cache_and_stream(
+            txn, self.get_user_by_id, (user_id,)
+        )
         txn.call_after(self.is_guest.invalidate, (user_id,))
 
     @cached()
--- a/tests/rest/admin/test_user.py
+++ b/tests/rest/admin/test_user.py
@@ -1003,12 +1003,23 @@ class UserRestTestCase(unittest.HomeserverTestCase):
 
     def prepare(self, reactor, clock, hs):
         self.store = hs.get_datastore()
+        self.auth_handler = hs.get_auth_handler()
 
+        # create users and get access tokens
+        # regardless of whether password login or SSO is allowed
         self.admin_user = self.register_user("admin", "pass", admin=True)
-        self.admin_user_tok = self.login("admin", "pass")
+        self.admin_user_tok = self.get_success(
+            self.auth_handler.get_access_token_for_user_id(
+                self.admin_user, device_id=None, valid_until_ms=None
+            )
+        )
 
         self.other_user = self.register_user("user", "pass", displayname="User")
-        self.other_user_token = self.login("user", "pass")
+        self.other_user_token = self.get_success(
+            self.auth_handler.get_access_token_for_user_id(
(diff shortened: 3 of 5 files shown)


tests/rest/admin/test_user.py
```diff
@@ -1507,41 +1520,89 @@ def test_change_name_deactivate_user_user_directory(self):
...
         # Deactivate the user.
+        self._deactivate_user("@user:test")
+
+        # Attempt to reactivate the user (without a password).
         channel = self.make_request(
             "PUT",
             self.url_other_user,
             access_token=self.admin_user_tok,
-            content=json.dumps({"deactivated": True}).encode(encoding="utf_8"),
+            content=json.dumps({"deactivated": False}).encode(encoding="utf_8"),
```
clokep, line 1543: We can pass just the dictionary in here instead of worrying about dumping and encoding (and the framework will handle that). It should simplify some of the code in here a bit! (There's a bunch of places to make this change, I only marked this one.)

```suggestion
            content={"deactivated": False},
```

tests/rest/admin/test_user.py
```diff
@@ -1507,41 +1520,89 @@ def test_change_name_deactivate_user_user_directory(self):
...
+        self.assertEqual("@user:test", channel.json_body["name"])
+        self.assertFalse(channel.json_body["deactivated"])
+        self.assertIsNone(channel.json_body["password_hash"])
+        self._is_erased("@user:test", False)
 
-        # Reactivate the user.
+    @override_config({"password_config": {"enabled": False}})
+    def test_reactivate_user_password_disabled(self):
+        """
+        Test reactivating another user when using SSO/SAML
```
clokep, line 1599:
```suggestion
        Test reactivating another user when using SSO.
```

tests/rest/admin/test_user.py
```diff
@@ -1507,41 +1520,89 @@ def test_change_name_deactivate_user_user_directory(self):
...
         self._is_erased("@user:test", False)
-        d = self.store.mark_user_erased("@user:test")
-        self.assertIsNone(self.get_success(d))
-        self._is_erased("@user:test", True)
 
-        # Attempt to reactivate the user (without a password).
+    @override_config({"password_config": {"localdb_enabled": False}})
+    def test_reactivate_user_localdb_disabled(self):
+        """
+        Test reactivating another user when using SSO/SAML
```
clokep, line 1565:
```suggestion
        Test reactivating another user when using SSO.
```

tests/rest/admin/test_user.py
```diff
@@ -1653,14 +1716,32 @@ def test_accidental_deactivation_prevention(self):
...
+            "/_synapse/admin/v2/users/%s" % urllib.parse.quote(user_id),
+            access_token=self.admin_user_tok,
+            content=json.dumps({"deactivated": True}).encode(encoding="utf_8"),
+        )
+        self.assertEqual(200, int(channel.result["code"]), msg=channel.result["body"])
+        self.assertTrue(channel.json_body["deactivated"])
+        self.assertIsNone(channel.json_body["password_hash"])
+        self._is_erased(user_id, False)
+        d = self.store.mark_user_erased(user_id)
+        self.assertIsNone(self.get_success(d))
```
clokep, line 1742: I'm a bit confused at why we're erasing data here?

changelog.d/9636.bugfix
```diff
@@ -0,0 +1 @@
+Checks if passwords are allowed before setting it for the user.
```
clokep, line 1: I wonder if we should consolidate the various three (?) PRs we've made around this into a single changelog entry. What do you think?

dklimpel: Yes, we can do it. But the first one (#9587) is already in 1.30.0rc1 included.

clokep: Ah, well maybe just for the last two then!

tests/rest/admin/test_user.py
dklimpel: To test the erase was part of #8362.
There was a bug because there was no test.
Now I have moved the whole part of deactivation in a private function because it is needed more than one times for these tests.

clokep: Ah, so the erasure isn't really only needed for some of the tests, but we do it for all just to simplify? That sounds fine!

synapse/storage/databases/main/registration.py
```diff
@@ -1210,6 +1210,7 @@ def set_user_deactivated_status_txn(self, txn, user_id: str, deactivated: bool):
         self._invalidate_cache_and_stream(
             txn, self.get_user_deactivated_status, (user_id,)
         )
+        self._invalidate_cache_and_stream(txn, self.get_user_by_id, (user_id,))
         txn.call_after(self.is_guest.invalidate, (user_id,))
```
clokep, line 1214: Good catch! 👍 

I don't see why we're invalidating the guest status here, but it seems like it has always been that way.

clokep (approved):

synapse/storage/databases/main/registration.py
dklimpel: I had started with implementing unit tests. The tests had always failed. That was the reason for catch this bug.

PR #9637, opened 2021-03-17
Fix bad naming of storage function

erikjohnston:
We had two functions named `get_forward_extremities_for_room` and
`get_forward_extremeties_for_room` that took different paramters. We
rename one of them to avoid confusion.


clokep (approved):
Seems good if CI passes!

PR #9638, opened 2021-03-17
Fix up types for the typing handler.

clokep:
More split off from #9374, this handles just the typing handler. I used a different approach from #9631 that is a bit more explicit. 🤷

### DIFF
--- a/synapse/rest/client/v1/room.py
+++ b/synapse/rest/client/v1/room.py
@@ -49,7 +49,7 @@ from synapse.util import json_decoder
 from synapse.util.stringutils import parse_and_validate_server_name, random_string
 
 if TYPE_CHECKING:
-    import synapse.server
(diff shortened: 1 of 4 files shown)


synapse/rest/client/v1/room.py
```diff
@@ -874,16 +874,19 @@ async def on_PUT(self, request, room_id, user_id):
         # Limit timeout to stop people from setting silly typing timeouts.
         timeout = min(content.get("timeout", 30000), 120000)
 
+        # Defer getting the typing handler since it will raise on workers.
+        typing_handler = self.hs.get_typing_writer_handler()
```
clokep, line 878: This is annoying, but the other option I came up with was to *not* register the endpoint on other works which people said was more confusing.

erikjohnston (approved):

PR #9639, opened 2021-03-17
Fix federation stall on concurrent access errors

### LINKED ISSUE #9635: Federation catchup stalls due to concurrent access
```
sender1_1        | 2021-03-16 20:00:14,270 - synapse.storage.txn - 538 - WARNING - process_event_queue_for_federation-19 - [TXN OPERROR] {store_destination_rooms_entries-9e4f} could not serialize access due to concurrent update
sender1_1        | 2021-03-16 20:00:14,335 - synapse.storage.txn - 538 - WARNING - process_event_queue_for_federation-19 - [TXN OPERROR] {store_destination_rooms_entries-9e4f} could not serialize access due to concurrent update
sender1_1        | 2021-03-16 20:00:14,380 - synapse.storage.txn - 538 - WARNING - process_event_queue_for_federation-19 - [TXN OPERROR] {store_destination_rooms_entries-9e4f} could not serialize access due to concurrent update
sender1_1        | 2021-03-16 20:00:14,414 - synapse.storage.txn - 538 - WARNING - process_event_queue_for_federation-19 - [TXN OPERROR] {store_destination_rooms_entries-9e4f} could not serialize access due to concurrent update
sender1_1        | 2021-03-16 20:00:14,451 - synapse.storage.txn - 538 - WARNING - process_event_queue_for_federation-19 - [TXN OPERROR] {store_destination_rooms_entries-9e4f} could not serialize access due to concurrent update
sender1_1        | 2021-03-16 20:00:14,510 - synapse.storage.txn - 538 - WARNING - process_event_queue_for_federation-19 - [TXN OPERROR] {store_destination_rooms_entries-9e4f} could not serialize access due to concurrent update
sender1_1        | 2021-03-16 20:00:14,771 - synapse.metrics.background_process_metrics - 210 - ERROR - process_event_queue_for_federation-19 - Background process 'process_event_queue_for_federation' threw an exception
sender1_1        | 2021-03-16 20:00:29,184 - synapse.federation.sender - 187 - DEBUG - process_event_queue_for_federation-20 - Handling 131486 -> 131586
```

My federation sender is in a simple loop: it tries to catch up with events, which it does by doing `handle_event` over batches of 100 events, which calls `store_destination_rooms_entries`, which throws an exception due to `could not serialize access due to concurrent update`, `process_event_queue_for_federation` task dies, it's `finally:` does `self._is_processing = False`, and then another process event queue task can be started in `notify_new_events` (which starts from scratch)

Here's the error:
```
sender1_1        | 2021-03-16 20:02:25,657 - synapse.metrics.background_process_metrics - 210 - ERROR - process_event_queue_for_federation-24 - Background process 'process_event_queue_for_federation' threw an exception
sender1_1        | Traceback (most recent call last):
sender1_1        |   File "/usr/local/lib/python3.8/site-packages/synapse/metrics/background_process_metrics.py", line 208, in run
sender1_1        |     return await maybe_awaitable(func(*args, **kwargs))
sender1_1        |   File "/usr/local/lib/python3.8/site-packages/synapse/federation/sender/__init__.py", line 272, in _process_event_queue_loop
sender1_1        |     await make_deferred_yieldable(
sender1_1        | twisted.internet.defer.FirstError: FirstError[#0, [Failure instance: Traceback: <class 'psycopg2.errors.SerializationFailure'>: could not serialize access due to concurrent update
sender1_1        | 
sender1_1        | /usr/local/lib/python3.8/site-packages/twisted/internet/defer.py:517:errback
sender1_1        | /usr/local/lib/python3.8/site-packages/twisted/internet/defer.py:580:_startRunCallbacks
sender1_1        | /usr/local/lib/python3.8/site-packages/twisted/internet/defer.py:662:_runCallbacks
sender1_1        | /usr/local/lib/python3.8/site-packages/twisted/internet/defer.py:1514:gotResult
sender1_1        | --- <exception caught here> ---
sender1_1        | /usr/local/lib/python3.8/site-packages/twisted/internet/defer.py:1443:_inlineCallbacks
sender1_1        | /usr/local/lib/python3.8/site-packages/twisted/python/failure.py:500:throwExceptionIntoGenerator
sender1_1        | /usr/local/lib/python3.8/site-packages/synapse/federation/sender/__init__.py:266:handle_room_events
sender1_1        | /usr/local/lib/python3.8/site-packages/synapse/federation/sender/__init__.py:254:handle_event
sender1_1        | /usr/local/lib/python3.8/site-packages/synapse/federation/sender/__init__.py:330:_send_pdu
sender1_1        | /usr/local/lib/python3.8/site-packages/synapse/storage/databases/main/transactions.py:315:store_destination_rooms_entries
sender1_1        | /usr/local/lib/python3.8/site-packages/synapse/storage/database.py:660:runInteraction
sender1_1        | /usr/local/lib/python3.8/site-packages/synapse/storage/database.py:743:runWithConnection
sender1_1        | /usr/local/lib/python3.8/site-packages/twisted/python/threadpool.py:238:inContext
sender1_1        | /usr/local/lib/python3.8/site-packages/twisted/python/threadpool.py:254:<lambda>
sender1_1        | /usr/local/lib/python3.8/site-packages/twisted/python/context.py:118:callWithContext
sender1_1        | /usr/local/lib/python3.8/site-packages/twisted/python/context.py:83:callWithContext
sender1_1        | /usr/local/lib/python3.8/site-packages/twisted/enterprise/adbapi.py:293:_runWithConnection
sender1_1        | /usr/local/lib/python3.8/site-packages/twisted/python/deprecate.py:298:deprecatedFunction
sender1_1        | /usr/local/lib/python3.8/site-packages/twisted/python/compat.py:403:reraise
sender1_1        | /usr/local/lib/python3.8/site-packages/twisted/enterprise/adbapi.py:284:_runWithConnection
sender1_1        | /usr/local/lib/python3.8/site-packages/synapse/storage/database.py:738:inner_func
sender1_1        | /usr/local/lib/python3.8/site-packages/synapse/storage/database.py:532:new_transaction
sender1_1        | /usr/local/lib/python3.8/site-packages/synapse/storage/databases/main/transactions.py:347:_store_destination_rooms_entries_txn
sender1_1        | /usr/local/lib/python3.8/site-packages/synapse/storage/database.py:275:execute_batch
sender1_1        | /usr/local/lib/python3.8/site-packages/synapse/storage/database.py:320:_do_execute
sender1_1        | /usr/local/lib/python3.8/site-packages/synapse/storage/database.py:275:<lambda>
sender1_1        | /usr/local/lib/python3.8/site-packages/psycopg2/extras.py:1209:execute_batch
sender1_1        | ]]
```

This causes my federation sender to "stall", with transmission loops constantly sending events while (undoubtedly) updating the `destination_room` table, which causes the `process_event_queue_for_federation` task to "stall" and start to loop.

ShadowJonathan:
This PR fixes #9635 by ensuring that the `store_destination_rooms_entries` db interaction gets retried indefinitely when it fails with `psycopg2.errors.SerializationFailure`.

This is fine, as this failure is simply a [rollback of the transaction if another transaction had changed the view of the table concurrently](https://www.postgresql.org/docs/10/transaction-iso.html#XACT-SERIALIZABLE), which means the "other" transaction has succeeded, and this one has failed, thus, this is infinitely retry-able until this transaction "succeeds" while the other "loses". In the very worst case, this blocks until the `destinations` table is less congested.

~~And to relieve some of that congestion, the second fix this PR gives is making sure every `handle_room_events` background task only instructs it's (eventual) `_send_pdu` calls to *only* call `self.store.store_destination_rooms_entries` when it is handling the final event in the (per-room) event list.~~

~~Again, this is fine, as `destinations_rooms` is used to calculate which events need to be caught-up with, and in the worst case (when the federation sender shuts down or aborts in the middle of a `handle_room_events` task), it'll cause the catch-up to notify more events (likely 1 more event) some servers, but it wont drop any to send.~~

~~This'll also cause federation to deliver to large rooms such as `#matrix:matrix.org` more efficiently, as the `store_destination_rooms_entries` call is only done *once* on the last event that is being caught up with, meaning that if that call is expensive, and federation events start "backing up", it'll be able to keep pace by calling this only once per room in the batch of 100 events that're submitted to per-destination queues. This'll increase performance substantially on smaller servers.~~

Edit: Some of the principles this PR was made on (for `destination_rooms`) were likely to be false, this PR will only be about the fix, and the optimisation will be tried in another PR.

### Pull Request Checklist

### DIFF
--- a/changelog.d/9639.bugfix
+++ b/changelog.d/9639.bugfix
@@ -1,0 +1,1 @@
+Fixes #9635, and optimizes federation sending.
--- a/changelog.d/9639.bugfix
+++ b/changelog.d/9639.bugfix
@@ -1,1 +1,1 @@
-Fixes #9635, and optimizes federation sending.
+Fixes #9635.
--- a/changelog.d/9639.bugfix
+++ b/changelog.d/9639.bugfix
@@ -1,1 +1,1 @@
-Fixes #9635.
+Fix bug where federation sending can stall due to `concurrent access` database exceptions when it falls behind.
--- a/synapse/federation/sender/__init__.py
+++ b/synapse/federation/sender/__init__.py
@@ -14,7 +14,7 @@
 # limitations under the License.
 
 import logging
-from typing import Dict, Hashable, Iterable, List, Optional, Set, Tuple
+from typing import Dict, Hashable, Iterable, List, Optional, Sequence, Set, Tuple
 
 from prometheus_client import Counter
 
@@ -189,7 +189,9 @@ class FederationSender:
                 if not events and next_token >= self._last_poked_id:
                     break
 
-                async def handle_event(event: EventBase) -> None:
+                async def handle_event(
+                    event: EventBase, flush_destination_rooms: bool = True
+                ) -> None:
                     # Only send events for this server.
                     send_on_behalf_of = event.internal_metadata.get_send_on_behalf_of()
                     is_mine = self.is_mine_id(event.sender)
@@ -251,7 +253,9 @@ class FederationSender:
                     logger.debug("Sending %s to %r", event, destinations)
 
                     if destinations:
-                        await self._send_pdu(event, destinations)
+                        await self._send_pdu(
+                            event, destinations, flush_destination_rooms
+                        )
 
                         now = self.clock.time_msec()
                         ts = await self.store.get_received_ts(event.event_id)
@@ -260,10 +264,11 @@ class FederationSender:
                             "federation_sender"
                         ).observe((now - ts) / 1000)
 
-                async def handle_room_events(events: Iterable[EventBase]) -> None:
+                async def handle_room_events(events: Sequence[EventBase]) -> None:
                     with Measure(self.clock, "handle_room_events"):
-                        for event in events:
-                            await handle_event(event)
+                        evs_len = len(events)
+                        for pos, event in enumerate(events, start=1):
+                            await handle_event(event, pos == evs_len)
 
                 events_by_room = {}  # type: Dict[str, List[EventBase]]
                 for event in events:
@@ -307,7 +312,12 @@ class FederationSender:
         finally:
             self._is_processing = False
 
-    async def _send_pdu(self, pdu: EventBase, destinations: Iterable[str]) -> None:
+    async def _send_pdu(
+        self,
+        pdu: EventBase,
+        destinations: Iterable[str],
+        flush_destination_rooms: bool = True,
+    ) -> None:
         # We loop through all destinations to see whether we already have
         # a transaction in progress. If we do, stick it in the pending_pdus
         # table and we'll get back to it later.
@@ -324,14 +334,15 @@ class FederationSender:
 
         assert pdu.internal_metadata.stream_ordering
 
-        # track the fact that we have a PDU for these destinations,
-        # to allow us to perform catch-up later on if the remote is unreachable
-        # for a while.
-        await self.store.store_destination_rooms_entries(
-            destinations,
-            pdu.room_id,
-            pdu.internal_metadata.stream_ordering,
-        )
+        if flush_destination_rooms:
+            # track the fact that we have a PDU for these destinations,
+            # to allow us to perform catch-up later on if the remote is unreachable
+            # for a while.
+            await self.store.store_destination_rooms_entries(
+                destinations,
+                pdu.room_id,
+                pdu.internal_metadata.stream_ordering,
+            )
 
         for destination in destinations:
             self._get_per_destination_queue(destination).send_pdu(pdu)
--- a/synapse/federation/sender/__init__.py
+++ b/synapse/federation/sender/__init__.py
@@ -14,7 +14,7 @@
 # limitations under the License.
 
 import logging
-from typing import Dict, Hashable, Iterable, List, Optional, Sequence, Set, Tuple
+from typing import Dict, Hashable, Iterable, List, Optional, Set, Tuple
 
 from prometheus_client import Counter
 
@@ -189,9 +189,7 @@ class FederationSender:
                 if not events and next_token >= self._last_poked_id:
                     break
 
-                async def handle_event(
-                    event: EventBase, flush_destination_rooms: bool = True
-                ) -> None:
+                async def handle_event(event: EventBase) -> None:
                     # Only send events for this server.
                     send_on_behalf_of = event.internal_metadata.get_send_on_behalf_of()
                     is_mine = self.is_mine_id(event.sender)
@@ -253,9 +251,7 @@ class FederationSender:
                     logger.debug("Sending %s to %r", event, destinations)
 
                     if destinations:
-                        await self._send_pdu(
-                            event, destinations, flush_destination_rooms
-                        )
+                        await self._send_pdu(event, destinations)
 
                         now = self.clock.time_msec()
                         ts = await self.store.get_received_ts(event.event_id)
@@ -264,11 +260,10 @@ class FederationSender:
                             "federation_sender"
                         ).observe((now - ts) / 1000)
 
-                async def handle_room_events(events: Sequence[EventBase]) -> None:
+                async def handle_room_events(events: Iterable[EventBase]) -> None:
                     with Measure(self.clock, "handle_room_events"):
-                        evs_len = len(events)
-                        for pos, event in enumerate(events, start=1):
-                            await handle_event(event, pos == evs_len)
+                        for event in events:
+                            await handle_event(event)
 
                 events_by_room = {}  # type: Dict[str, List[EventBase]]
                 for event in events:
@@ -312,12 +307,7 @@ class FederationSender:
         finally:
             self._is_processing = False
 
-    async def _send_pdu(
-        self,
-        pdu: EventBase,
-        destinations: Iterable[str],
-        flush_destination_rooms: bool = True,
-    ) -> None:
+    async def _send_pdu(self, pdu: EventBase, destinations: Iterable[str]) -> None:
         # We loop through all destinations to see whether we already have
         # a transaction in progress. If we do, stick it in the pending_pdus
         # table and we'll get back to it later.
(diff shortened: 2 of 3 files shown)


ShadowJonathan:
The way it increases performance is due to `_send_pdu` previously calling `store_destination_rooms_entries` once per event, this can be costly for large rooms.

Now, it calls it only once per room per `process_event_queue_for_federation` batch (max 100), which means once pressure is applied onto large rooms by local users, `process_event_queue_for_federation` will start to process larger batches, which then call `store_destination_rooms_entries` less, which means that it is able to "catch up" more efficiently per batch, until an equilibrium is reached.

erikjohnston:
Oh, this is very interesting. I'm surprised that its so contested and is introducing delays. My only concern with your approach is that it means not keeping that table up to date, and I'm not sure what the consequences of that are.

We recently introduced a change which handles the `SerializationFailure` by doing upserts outside of transactions, so I wonder if just splitting `store_destination_rooms_entries_txn` into two would magically help here, e.g. something like:

```python
    async def store_destination_rooms_entries(
        self,
        destinations: Iterable[str],
        room_id: str,
        stream_ordering: int,
    ) -> None:
        """
        Updates or creates `destination_rooms` entries in batch for a single event.

        Args:
            destinations: list of destinations
            room_id: the room_id of the event
            stream_ordering: the stream_ordering of the event
        """

        await self.db_pool.simple_upsert_many(
            table="destinations",
            key_names=("destination",),
            key_values=[(d,) for d in destinations],
            value_names=[],
            value_values=[],
            desc="store_destination_rooms_entries_dests",
        )

        rows = [(destination, room_id) for destination in destinations]
        await self.db_pool.simple_upsert_many(
            table="destination_rooms",
            key_names=("destination", "room_id"),
            key_values=rows,
            value_names=["stream_ordering"],
            value_values=[(stream_ordering,)] * len(rows),
            desc="store_destination_rooms_entries_rooms",
        )
```

If that works, then yay we can avoid some of the complexity of this PR.

ShadowJonathan:
That fixes #9635, but that doesn't fix the performance improvement I've noticed, of which I've hypothesised the following;

> Again, this is fine, as `destinations_rooms` is used to calculate which events need to be caught-up with, and in the worst case (when the federation sender shuts down or aborts in the middle of a `handle_room_events` task), it'll cause the catch-up to notify more events (likely 1 more event) some servers, but it wont drop any to send.

Which is only in the case of which `handle_room_events` gets an exception thrown, which is then thrown into `store_destination_rooms_entries`, which is caught and effectively transaction sending is reset (or stays) at the current `stream_id` in the database.

erikjohnston:
I'm sort of wondering whether that will speed things up. The DB hits should be fast, and we'll be doing quite a lot of other DB hits during processing, so I'm slightly surprised if the fixed version would actually cause a slow down. I can completely believe that when things were getting contested, and retried, etc that would cause it to be a source of slowness.

ShadowJonathan:
> I'm sort of wondering whether that will speed things up.

The federation senders now do an insert to `destinations` (for certainty that the foreign key exists), and an update to `destination_rooms`, this needs to be done concurrently, for *every* event, which is likely to result in many contentions for a slow postgres database and many federation events.

> The DB hits should be fast [...]

In my experience, with `#matrix:matrix.org`, they were not (for a small homeserver), as `destinations` and `destination_rooms` accesses/updates (on every event, for all 2k homeservers) are slow (and resulted in that concurrent access bug), i think this optimization (at least) will offer a lot to smaller and/or slower homeservers (as those will get the best performance bonus from this).

erikjohnston:
> The federation senders now do an insert to `destinations` (for certainty that the foreign key exists), and an update to `destination_rooms`, this needs to be done concurrently, for _every_ event, which is likely to result in many contentions for a slow postgres database with many federated events to multiple rooms concurrently.

The insert into `destinations` shouldn't be bad in the contended case, as its just a `DO NOTHING`. My understanding is that `destination_rooms` shouldn't have concurrent writes to the same room, so also shouldn't be contended.

The thing is that while we do do this for every event, we also *also* do a lot of reads and writes to the DB for each transaction that takes place, which is often per event *per destination*, so even if we're doing a bunch of batching up my gut instinct would be they'd be dwarfed by all the other stuff.

Though given this is an issue occurring while your federation sender is busy catching up its entirely possible that all the performance characteristic are different than the "normal" case.

ShadowJonathan:
> My understanding is that `destination_rooms` shouldn't have concurrent writes to the same room, so also shouldn't be contended.

It is, because for a batch from `process_event_queue_for_federation` is separated into the rooms that each events belong to, and so `handle_room_events` gets called for each of these batches, concurrently.

When the homeservers in each room overlap, *then* it gets contested, and so a room like `#matrix:matrix.org` causes a holdup on the database, and any other small room that shares a room with `#matrix:matrix.org` (very likely) gets no serialised access, no concurrent update.

And even for large rooms like `#matrix:matrix.org`, a call to update `destination_rooms` is *very* expensive, it would be useful to minimise this as much as possible anyways.

ShadowJonathan:
...wait, I'm realising I might have it wrong about `destination_rooms`, I'll need to think about this for a moment.

ShadowJonathan:
Yeah, i think I'll remove the optimization and split that into another PR, as it currently has some problems that needs to be hashed out individually, I'll make this PR just about the fix, then

ShadowJonathan:
(I'm sorry for all the drama, this is still one of the first times I've looked at this code, gaining more understanding the longer i looked at it 😅)

erikjohnston:
> Yeah, i think I'll remove the optimization and split that into another PR, as it currently has some problems that needs to be hashed out individually, I'll make this PR just about the fix, then

No worries! Your initial stab was perfectly reasonable, and we do similar things elsewhere. It's just one of those things that's best to avoid if possible, and I happen to have been staring at a lot of this sort of stuff recently :sweat_smile:

changelog.d/9639.bugfix
```diff
@@ -0,0 +1 @@
+Fixes #9635.
```
erikjohnston, line 1:
```suggestion
Fix bug where federation sending can stall due to `concurrent access` database exceptions when it falls behind.
```

erikjohnston (approved):

clokep:
Can we get an improved title on this PR? I find it best to put the issue that it fixes in the description and not in the title and have the title be a short bit on what the change actually is.

ShadowJonathan:
Whoops, will do

richvdh:
It might be helpful to update the description of the PR too. AFAICT it ended up being quite different to what was written originally.

Issue #9490, opened 2021-02-24
_room_pdu_linearizer starves other incoming transactions


richvdh:
I had two transactions arrive at about the same time, each containing one PDU for the same room, each of which had a `prev_events` ref to a single unknown event.

The first transaction required a significant backfill/state resolution operation, requiring over 90 minutes of activity. 

The second transaction waited patiently for in the `_room_pdu_linearizer` for the first one to finish, and once it finally got to the front of the queue, was processed within 20 seconds. So I'm now 90 minutes behind in that room, and once you get behind it's very hard to catch up.

My questions are:
 * can we somehow round-robin between transactions for a room, rather than having a single transaction hold the lock for 90 minutes?
 * Is this lock even required? If I understand correctly, when you have multiple `federation_inbound` workers, we can end up with each worker doing its own backfill here.

erikjohnston:
I think the linearizer is there to try and protect against lots of redundant work, in particular if we are busy processing an event A and then get another event B that references A, we don't want to then go and effectively fetch A and start processing it a second time concurrently. Synapse should handle it all correctly, since when we come to persist the events the event persister queues will Do The Right thing with any duplicate data.

Some other thoughts:

* When we implemented running multiple federation inbound workers i think we deliberately decided that duplicating work that is happening on *other* workers isn't the end of the world, at least we have still limited the total amount of duplicate work that can happen at any one time. (Plus cross worker locks is something that sounds hard)
* We could refine the lock so that we can process multiple events in a room at a time, so long as there aren't any dependencies, i.e. if we are processing event A then any events that come in that reference A should block waiting for it to be processed, but events that don't reference in flight/queued up events can be processed immediately. We probably *do* want to limit the in flight amount of work somehow, as otherwise we run the risk of a broken room slowly consuming up all resources on the worker, starving the other rooms from getting events.
*  :scream: 90 minutes to process an event
* Blocking responding to a transaction isn't ideal if we are spending so long processing events. Ideally we'd write such events to a staging area and return immediately, we could either do this for all events or just for events that require outbound HTTP calls (i.e. have missing prev events, etc). The problem here is that we'd have multiple workers pushing and pulling from the queue, which will require some form of coordination between the workers.


To expand a bit on queuing up events in a staging area:

* This allows batching up work, if we get a bunch of events from various servers then we can be more intelligent in how we process them (especially when other servers are sending us random events in "catch up" mode)
[...]

PR #9640, opened 2021-03-17
Make federation catchup send last event from any server.

erikjohnston:
Currently federation catchup will send the last *local* event that we failed to send to the remote. This can cause issues for large rooms where lots of servers have sent events while the remote server was down, as when it comes back up again it'll be flooded with events from various points in the DAG.

Instead, let's make it so that all the servers send the most recent events, even if its not theirs. The remote should deduplicate the
events, so there shouldn't be much overhead in doing this. Alternatively, the servers could only send local events if they were also extremities and hope that the other server will send the event over, but that is a bit risky.

c.f. https://github.com/matrix-org/synapse/issues/9490 and https://github.com/matrix-org/synapse/pull/9621

### DIFF
--- a/synapse/federation/federation_server.py
+++ b/synapse/federation/federation_server.py
@@ -35,7 +35,7 @@ from twisted.internet import defer
 from twisted.internet.abstract import isIPAddress
 from twisted.python import failure
 
-from synapse.api.constants import EduTypes, EventTypes, Membership
+from synapse.api.constants import EduTypes, EventTypes
 from synapse.api.errors import (
     AuthError,
     Codes,
@@ -63,7 +63,7 @@ from synapse.replication.http.federation import (
     ReplicationFederationSendEduRestServlet,
     ReplicationGetQueryRestServlet,
 )
-from synapse.types import JsonDict, get_domain_from_id
+from synapse.types import JsonDict
 from synapse.util import glob_to_regex, json_decoder, unwrapFirstError
 from synapse.util.async_helpers import Linearizer, concurrently_execute
 from synapse.util.caches.response_cache import ResponseCache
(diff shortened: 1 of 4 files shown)


synapse/federation/sender/per_destination_queue.py
```diff
@@ -415,22 +416,93 @@ async def _catch_up_transmission_loop(self) -> None:
...
-                rooms = [p.room_id for p in catchup_pdus]
-                logger.info("Catching up rooms to %s: %r", self._destination, rooms)
+            # We send transactions with events from one room only, as its likely
+            # that the remote will have to do additional processing, which may
+            # take some time. It's better to give it small amounts of work
+            # rather than risk the request timing out and repeatedly being
+            # retried, and not making any progress.
+            for pdu in catchup_pdus:
+                # The PDU from the DB will be the last PDU in the room from
+                # *this server* that wasn't sent to the remote. However, other
```
clokep, line 426: This is because we only know that *we* failed to send an event, but the remote server might be communicating OK with other servers in the room? (Meaning that it could actually already have the event we're trying to send to it?)

synapse/federation/sender/per_destination_queue.py
```diff
@@ -415,22 +416,93 @@ async def _catch_up_transmission_loop(self) -> None:
...
 
-            if logger.isEnabledFor(logging.INFO):
-                rooms = [p.room_id for p in catchup_pdus]
-                logger.info("Catching up rooms to %s: %r", self._destination, rooms)
+            # We send transactions with events from one room only, as its likely
+            # that the remote will have to do additional processing, which may
+            # take some time. It's better to give it small amounts of work
+            # rather than risk the request timing out and repeatedly being
+            # retried, and not making any progress.
+            for pdu in catchup_pdus:
```
clokep, line 424: Should we be grouping these by room first? Otherwise it seems like we would do some unnecessary work. (I think at least...)

synapse/federation/federation_server.py
```diff
@@ -727,27 +727,6 @@ async def _handle_received_pdu(self, origin: str, pdu: EventBase) -> None:
             if the event was unacceptable for any other reason (eg, too large,
             too many prev_events, couldn't find the prev_events)
         """
-        # check that it's actually being sent from a valid destination to
-        # workaround bug #1753 in 0.18.5 and 0.18.6
-        if origin != get_domain_from_id(pdu.sender):
```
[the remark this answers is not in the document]
erikjohnston: This needs to get removed as we're now sending events in transactions that aren't from the originating server. It's safe to remove due to the bug having long since been fixed.

synapse/federation/sender/per_destination_queue.py
erikjohnston: Err, I don't follow really. The server will only track the last event it failed to send to the remote was, which will only be local events. When the server comes back up we want it to be told about the most recent events in the room, so that it can figure out it missed stuff. We don't want to send anything other than the latest events as that will cause unnecessary processing on the remote, and the last event sent by the server to a room won't necessarily be the part of the latest events in the room (as other servers may have sent events since).

In other words, we want the remote to only receive the current forward extremities in the room; we're not trying to send it all events its missed.

clokep: > The server will only track the last event it failed to send to the remote was, which will only be local events.

This is what I was trying to confirm. 👍

clokep (approved):

Issue #9446, opened 2021-02-19
support for spaces summary API (unfederated)


richvdh:
[MSC2946](https://github.com/matrix-org/matrix-doc/pull/2946) proposes a "summary" API which is useful for spaces support. This issue is to track creation of that API. As part of this initial implementation, the "Federation API" will not be implemented. (TBD: what happens if someone queries a space that we don't have info for locally?)

richvdh:
supporting pagination is going to be a bit fiddly, so estimate 8 days or so on this, to include testing

richvdh:
pagination is descoped for now.

PR #9643, opened 2021-03-18
Initial spaces summary API

richvdh:
This is very bare-bones for now: federation will come soon, while pagination is descoped for now but will come later.

Unfortunately there is no good reference to the API that it's implementing yet. It's vaguely based on MSC2946, but there are various differences, and reading MSC2946 may well be more confusing than helpful. Writing a doc is on my todo list but I wanted to get this up for review.

Fixes: #9446

### DIFF
--- a/synapse/config/experimental.py
+++ b/synapse/config/experimental.py
@@ -28,8 +28,5 @@ class ExperimentalConfig(Config):
         # MSC2858 (multiple SSO identity providers)
         self.msc2858_enabled = experimental.get("msc2858_enabled", False)  # type: bool
 
-        # MSC2946 (space summaries)
-        self.msc2946_enabled = experimental.get("msc2946_enabled", False)  # type: bool
-
-        # MSC1772 (spaces)
-        self.msc1772_enabled = experimental.get("msc1772_enabled", False)  # type: bool
+        # Spaces (MSC1772, MSC2946, etc)
+        self.spaces_enabled = experimental.get("spaces_enabled", False)  # type: bool
--- a/synapse/config/experimental.py
+++ b/synapse/config/experimental.py
@@ -27,3 +27,9 @@ class ExperimentalConfig(Config):
 
         # MSC2858 (multiple SSO identity providers)
         self.msc2858_enabled = experimental.get("msc2858_enabled", False)  # type: bool
+
+        # MSC2946 (space summaries)
+        self.msc2946_enabled = experimental.get("msc2946_enabled", False)  # type: bool
+
+        # MSC1772 (spaces)
+        self.msc1772_enabled = experimental.get("msc1772_enabled", False)  # type: bool
--- a/synapse/handlers/space_summary.py
+++ b/synapse/handlers/space_summary.py
@@ -31,12 +31,10 @@ logger = logging.getLogger(__name__)
 
 # number of rooms to return. We'll stop once we hit this limit.
 # TODO: allow clients to reduce this with a request param.
-# TODO: increase it, probably. It's deliberately low to start with so that
-#    we can think about whether we need pagination.
-ROOMS_LIMIT = 5
+ROOMS_LIMIT = 50
 
 # max number of events to return per room.
-ROOMS_PER_SPACE_LIMIT = 5
+ROOMS_PER_SPACE_LIMIT = 50
 
 
 class SpaceSummaryHandler:
@@ -116,7 +114,7 @@ class SpaceSummaryHandler:
                 child_events = filter(_is_suggested_child_event, child_events)
 
             # if this is not the first room, and the client has specified a limit,
-            # apply it
+            # apply it (the client limit does not apply to the root room)
             if max_rooms_per_space is not None and len(processed_rooms) > 1:
                 max_rooms = min(ROOMS_PER_SPACE_LIMIT, max_rooms_per_space)
             else:
--- a/synapse/handlers/space_summary.py
+++ b/synapse/handlers/space_summary.py
@@ -31,10 +31,10 @@ logger = logging.getLogger(__name__)
 
 # number of rooms to return. We'll stop once we hit this limit.
 # TODO: allow clients to reduce this with a request param.
-ROOMS_LIMIT = 50
+MAX_ROOMS = 50
 
 # max number of events to return per room.
-ROOMS_PER_SPACE_LIMIT = 50
+MAX_ROOMS_PER_SPACE = 50
 
 
 class SpaceSummaryHandler:
@@ -86,7 +86,7 @@ class SpaceSummaryHandler:
 
         now = self._clock.time_msec()
 
-        while room_queue and len(rooms_result) < ROOMS_LIMIT:
+        while room_queue and len(rooms_result) < MAX_ROOMS:
             room_id = room_queue.popleft()
             logger.debug("Processing room %s", room_id)
             processed_rooms.add(room_id)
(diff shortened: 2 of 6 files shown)


synapse/handlers/space_summary.py
```diff
@@ -0,0 +1,202 @@
...
+    from synapse.server import HomeServer
+
+logger = logging.getLogger(__name__)
+
+# number of rooms to return. We'll stop once we hit this limit.
+# TODO: allow clients to reduce this with a request param.
+ROOMS_LIMIT = 50
+
+# max number of events to return per room.
+ROOMS_PER_SPACE_LIMIT = 50
```
clokep, line 37: I think prefixing with `MAX_` matches some other constants better.

```suggestion
MAX_ROOMS_PER_SPACE_LIMIT = 50
```

synapse/config/experimental.py
```diff
@@ -27,3 +27,9 @@ def read_config(self, config: JsonDict, **kwargs):
 
         # MSC2858 (multiple SSO identity providers)
         self.msc2858_enabled = experimental.get("msc2858_enabled", False)  # type: bool
+
+        # MSC2946 (space summaries)
+        self.msc2946_enabled = experimental.get("msc2946_enabled", False)  # type: bool
+
+        # MSC1772 (spaces)
+        self.msc1772_enabled = experimental.get("msc1772_enabled", False)  # type: bool
```
clokep, line 35: I see in the code where these are used, but it seems like it doesn't really make sense to enable one without the other, do we really need two flags?

synapse/handlers/space_summary.py
```diff
@@ -0,0 +1,202 @@
...
+
+            room_entry = await self._build_room_entry(room_id)
+            rooms_result.append(room_entry)
+
+            # look for child rooms/spaces.
+            child_events = await self._get_child_events(room_id)
+
+            if suggested_only:
+                # we only care about suggested children
+                child_events = filter(_is_suggested_child_event, child_events)
```
clokep, line 114: Seems like this could be a flag to `_get_child_events`, but this works fine too.

synapse/handlers/space_summary.py
```diff
@@ -0,0 +1,202 @@
...
+
+            # if this is not the first room, and the client has specified a limit,
+            # apply it (the client limit does not apply to the root room)
```
clokep, line 117: What does "first room" mean here? (How is it different from "root room"?)

synapse/handlers/space_summary.py
```diff
@@ -0,0 +1,202 @@
...
+            "guest_can_join": stats["guest_access"] == "can_join",
+            "room_type": room_type,
+        }
+
+        # Filter out Nones – rather omit the field altogether
+        room_entry = {k: v for k, v in entry.items() if v is not None}
+
+        return room_entry
+
+    async def _get_child_events(self, room_id: str) -> Iterable[EventBase]:
```
clokep, line 177: I understand that this returns events, but it seems like it would be clearer if this was called `_get_child_rooms` or something.

synapse/rest/client/v1/room.py
```diff
@@ -981,7 +984,54 @@ def register_txn_path(servlet, regex_string, http_server, with_get=False):
...
+            "/rooms/(?P<room_id>[^/]*)/spaces$"
+        ),
+    )
+
+    def __init__(self, hs: "synapse.server.HomeServer"):
+        super().__init__()
+        self._auth = hs.get_auth()
+        self._space_summary_handler = hs.get_space_summary_handler()
+
+    async def on_GET(self, request: Request, room_id: str):
```
clokep, line 1000:
```suggestion
    async def on_GET(self, request: Request, room_id: str) -> Tuple[int, JsonDict]:
```

synapse/rest/client/v1/room.py
```diff
@@ -981,7 +984,54 @@ def register_txn_path(servlet, regex_string, http_server, with_get=False):
...
+        requester = await self._auth.get_user_by_req(request, allow_guest=True)
+
+        return 200, await self._space_summary_handler.get_space_summary(
+            requester.user.to_string(),
+            room_id,
+            suggested_only=parse_boolean(request, "suggested_only", default=False),
+            max_rooms_per_space=parse_integer(request, "max_rooms_per_space"),
+        )
+
+    async def on_POST(self, request: Request, room_id: str):
```
clokep, line 1010:
```suggestion
    async def on_POST(self, request: Request, room_id: str) -> Tuple[int, JsonDict]:
```

clokep: Is this the same API, just using a JSON object for parameters instead of query parameters? 😕

synapse/handlers/space_summary.py
richvdh: uh, they mean the same thing - the room_id that is passed in the request, which we start walking from. I'll try to clarify the comment.

synapse/config/experimental.py
richvdh: Good question. I'm trying to look forward to a time when we start dropping support for these unstable identifiers, and wanted to leave the path open for them to land separately

I kindof agree though. I'll combine them into one option.

synapse/handlers/space_summary.py
clokep: That's what I initially thought but then second guessed myself. Thanks for clarifying!

synapse/rest/client/v1/room.py
clokep: It seems like it should be a GET API, I'm not sure if having both is less confusing though. 😄

synapse/handlers/space_summary.py
richvdh: I still end up needing a separate function to do the test, so I don't think doing the filtering inside `_get_child_events` helps much.

synapse/handlers/space_summary.py
richvdh: I've gone for `MAX_ROOMS_PER_SPACE`, as the LIMIT seemed redundant. WDYT?

synapse/handlers/space_summary.py
richvdh: I'm honestly not sure I agree. I could maybe go with `_get_child_room_events` if that would be clearer?

synapse/handlers/space_summary.py
```diff
@@ -0,0 +1,199 @@
...
+        # there should always be an entry
+        assert stats is not None, "unable to retrieve stats for %s" % (room_id,)
+
+        current_state_ids = await self._store.get_current_state_ids(room_id)
+        create_event = await self._store.get_event(
+            current_state_ids[(EventTypes.Create, "")]
+        )
+
+        # TODO: update once MSC1772 lands
+        room_type = create_event.content.get(EventContentFields.MSC1772_ROOM_TYPE)
```
richvdh, line 156: this was the reason for the second config option: I wanted its presence to remind us that there was this "unstable" code in here, beyond the experimentalness of the SpaceSummaryHandler as a whole. Anyway, I've replaced the conditions with TODOs.

synapse/handlers/space_summary.py
clokep: Sounds good to me!

synapse/handlers/space_summary.py
clokep: Re-reading it, it seems pretty clear what it is doing. I think this is OK.

synapse/handlers/space_summary.py
```diff
@@ -0,0 +1,199 @@
...
+        }
+
+        # Filter out Nones – rather omit the field altogether
+        room_entry = {k: v for k, v in entry.items() if v is not None}
+
+        return room_entry
+
+    async def _get_child_events(self, room_id: str) -> Iterable[EventBase]:
+        # look for child rooms/spaces.
+        current_state_ids = await self._store.get_current_state_ids(room_id)
```
clokep, line 179: It seems like there should be an easier way to pull the state of a room filtering for the event type (instead of having to do it after the fact). It seems we only have methods for filtering by the type + state key though, which doesn't help here.

clokep (approved):
Looks good to me!

synapse/handlers/space_summary.py
richvdh: I agree there should, but it's not a can of worms I want to open right this minute.

PR #9644, opened 2021-03-18
Implement MSC3026: busy presence state

babolivier:
This PR implements [MSC3026](https://github.com/matrix-org/matrix-doc/pull/3026), which introduces a busy presence state. This state means that the user is online and active in the app, but might not generate new presence events (e.g. the user is in a call and the window's not in focus) and therefore shouldn't get their presence changed to unavailable if no new event have been received in some time.

### DIFF
--- a/synapse/api/constants.py
+++ b/synapse/api/constants.py
@@ -51,6 +51,7 @@ class PresenceState:
     OFFLINE = "offline"
     UNAVAILABLE = "unavailable"
     ONLINE = "online"
+    BUSY = "org.matrix.msc3026.busy"
 
 
 class JoinRules:
--- a/synapse/handlers/presence.py
+++ b/synapse/handlers/presence.py
@@ -751,7 +751,7 @@ class PresenceHandler(BasePresenceHandler):
             new_fields["status_msg"] = msg
 
         if presence == PresenceState.ONLINE or (
-            self._busy_presence_enabled and presence == PresenceState.BUSY
+            presence == PresenceState.BUSY and self._busy_presence_enabled
         ):
             new_fields["last_active_ts"] = self.clock.time_msec()
 
--- a/synapse/handlers/presence.py
+++ b/synapse/handlers/presence.py
@@ -730,6 +730,7 @@ class PresenceHandler(BasePresenceHandler):
             PresenceState.ONLINE,
             PresenceState.UNAVAILABLE,
             PresenceState.OFFLINE,
+            PresenceState.BUSY,
         )
         if presence not in valid_presence:
             raise SynapseError(400, "Invalid presence state")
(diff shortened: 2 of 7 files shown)


clokep:
We've been putting some experimental features behind flags (see `synapse.config.experimental` -- would that make sense for this?

babolivier:
I'm not sure - this is a pretty small change, which only adds support for clients setting the new state (not making the server change it directly) so I'd think not, but happy to discuss it

babolivier:
Discussion in #synapse-dev concluded on that it should be behind a flag.

synapse/rest/client/versions.py
```diff
@@ -81,6 +81,8 @@ def on_GET(self, request):
                     "io.element.e2ee_forced.public": self.e2ee_forced_public,
                     "io.element.e2ee_forced.private": self.e2ee_forced_private,
                     "io.element.e2ee_forced.trusted_private": self.e2ee_forced_trusted_private,
+                    # Supports the busy presence state described in MSC3026.
+                    "org.matrix.msc3026.busy_presence": True,
```
clokep, line 85: This should be based on the configuration flag.

synapse/handlers/presence.py
```diff
@@ -744,7 +750,9 @@ async def set_state(self, target_user, state, ignore_status_msg=False):
             msg = status_msg if presence != PresenceState.OFFLINE else None
             new_fields["status_msg"] = msg
 
-        if presence == PresenceState.ONLINE:
+        if presence == PresenceState.ONLINE or (
+            self._busy_presence_enabled and presence == PresenceState.BUSY
```
clokep, line 754: Maybe for consistency?

```suggestion
            presence == PresenceState.BUSY and self._busy_presence_enabled
```

synapse/api/constants.py
```diff
@@ -51,6 +51,7 @@ class PresenceState:
     OFFLINE = "offline"
     UNAVAILABLE = "unavailable"
     ONLINE = "online"
+    BUSY = "org.matrix.msc3026.busy"
```
ShadowJonathan, line 54: Could there be any special tricks for making sure "busy" is interpreted by both it's MSC prefix and its eventual name? `busy`, something like a frozen list with the first item being the "canonical" one.

babolivier: It'll probably be easier to have two constants, e.g. `BUSY` and `MSC3026BUSY`, once the MSC is merged (and we'll also be able to get rid of the latter after some time, once most clients are up to date with the spec).

ShadowJonathan: I suggest doing that, and changing the `==`s in your PR with `in (MSC3026BUSY, BUSY)` or something.

clokep (approved):
LGTM!

synapse/api/constants.py
babolivier: Agreed, though I believe it's better to do that as a second step when the feature gets released in a stable release of the spec because implementations are not supposed to support stable notations before that point.

PR #9645, opened 2021-03-18
federation_client: stop adding URL prefix

richvdh:
this is just a dev script so I don't think we need to announce the change particularly.


clokep (approved):
Seems fine. Is this even distributed?

richvdh:
outside of the git repo you mean? I don't think so, no.

clokep:
Yeah, I meant as part of our debs or pip packages. I don't think so either. I pretty much consider anything in `scripts-dev` to be fair game for immediate change.

PR #9648, opened 2021-03-18
Add an admin API to manage ratelimit for a specific user

dklimpel:
Fixes: #6286

A few questions:

1. Is the `RoomWorkerStore` the right class for `get_ratelimit_for_user`, `set_ratelimit_for_user`, and `delete_ratelimit_for_user`?
Should I move it to `RegistrationWorkerStore` or anywhere else?
2. Is the best solution to delete a custom ratelimit with http `DELETE`. Or do you prefer to set to `null` with http `POST`?
3. When get status with http `GET` and no custom ratelimit is set, do you prefer return an empty dict (`{}`) or values with `null` (`      "messages_per_second": null, "burst_count": null`)

### Pull Request Checklist

### DIFF
--- a/docs/admin_api/user_admin_api.rst
+++ b/docs/admin_api/user_admin_api.rst
@@ -827,7 +827,7 @@ The following parameters should be set in the URL:
 Override ratelimiting for users
 ===============================
 
-This API allows to override or disable rate limiting for a specific user.
+This API allows to override or disable ratelimiting for a specific user.
 There are specific APIs to set, get and delete a ratelimit.
 
 Get status of ratelimit
@@ -861,11 +861,16 @@ The following parameters should be set in the URL:
 The following fields are returned in the JSON response body:
 
 - ``messages_per_second`` - integer - The number of actions that can
-  be performed in a second.
+  be performed in a second. `0` or `null` mean that ratelimiting is disabled
+  for this user.
 - ``burst_count`` - integer - How many actions that can be performed
   before being limited.
 
-If **no** custom ratelimit is set, the values are ``null``.
+If **no** custom ratelimit is set, an empty JSON dict is returned.
+
+.. code:: json
+
+    {}
 
 Set ratelimit
 -------------
--- a/docs/admin_api/user_admin_api.rst
+++ b/docs/admin_api/user_admin_api.rst
@@ -823,3 +823,114 @@ The following parameters should be set in the URL:
 
 - ``user_id`` - The fully qualified MXID: for example, ``@user:server.com``. The user must
   be local.
+
+Override ratelimiting for users
+===============================
+
+This API allows to override or disable rate limiting for a specific user.
+There are specific APIs to set, get and delete a ratelimit.
+
+Get status of ratelimit
+-----------------------
+
+The API is::
+
+  GET /_synapse/admin/v1/users/<user_id>/override_ratelimit
+
+To use it, you will need to authenticate by providing an ``access_token`` for a
+server admin: see `README.rst <README.rst>`_.
+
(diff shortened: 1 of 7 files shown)


erkinalp:
Spec counterpart: https://github.com/matrix-org/matrix-doc/pull/3059

erikjohnston:
> A few questions:
> 
> 1. Is the `RoomWorkerStore` the right class for `get_ratelimit_for_user`, `set_ratelimit_for_user`, and `delete_ratelimit_for_user`?
>    Should I move it to `RegistrationWorkerStore` or anywhere else?

We should probably move it out to its own file, but its probably fine where it is.

> 2. Is the best solution to delete a custom ratelimit with http `DELETE`. Or do you prefer to set to `null` with http `POST`?

`DELETE` seems like a sensible choice

> 3. When get status with http `GET` and no custom ratelimit is set, do you prefer return an empty dict (`{}`) or values with `null` (`      "messages_per_second": null, "burst_count": null`)

So, looking at the code values or either `0` or `null` can mean "disable ratelimit for user", so we shouldn't use values with `null` to mean no ratelimitng is set.

docs/admin_api/user_admin_api.rst
```diff
@@ -823,3 +823,114 @@ The following parameters should be set in the URL:
...
+  be local.
+
+**Response**
+
+The following fields are returned in the JSON response body:
+
+- ``messages_per_second`` - integer - The number of actions that can
+  be performed in a second.
+- ``burst_count`` - integer - How many actions that can be performed
+  before being limited.
```
erikjohnston, line 866: Can we document that `0` or `null` mean ratelimiting is disabled? Or we should replace `0` with `null` (or vice versa) before returning the values, for consistency sake?

erikjohnston (dismissed):
Looks good otherwise!

docs/admin_api/user_admin_api.rst
dklimpel: If i read the code correctly, then `null` is better for deactivating, as it breaks the processing sooner!?

https://github.com/matrix-org/synapse/blob/5b268997bdde04c905eed0a7c04e9e2352e35fba/synapse/handlers/_base.py#L102-L108

erikjohnston: `if not override.messages_per_second:` and `if not override.messages_per_second or override.messages_per_second == 0:` are equivalent here (since both `None` and `0` are falsey in python), which is why we've ended up in a situation where `0` and `None` mean the same thing anyway.

erikjohnston: I think the only thing we might want to do here is to change `null` to `0` (or vice versa) when we return the values from the API, for consistency. I don't really mind though.

(The fact that we treat `0` and `null` the same is a bit awful, but untangling that may be a bit of a mess)

erikjohnston (commented):
I think this LGTM, but will wait to merge until we resolve the outstanding thread about `0` vs `null`

docs/admin_api/user_admin_api.rst
clokep: You can probably just cast to an int to always return `0`?

dklimpel: Yes, I can do it.
Should I do the conversion from `None` to `0` only in admin API or direct in store `get_ratelimit_for_user`?

erikjohnston: I'd probably do it in the API for now, just to avoid unnecessarily changing things used by lots of places.

erikjohnston (approved):
Sorry this has taken so long to get through review!

PR #9650, opened 2021-03-19
Add a storage method for returning all current presence from all users

anoadragon453:
Split off from https://github.com/matrix-org/synapse/pull/9491

Adds a storage method for getting the current presence of all local users, optionally excluding those that are offline. This will be used by the code in #9491 when a PresenceRouter module informs Synapse that a given user should have `"ALL"` user presence updates routed to them. Specifically, it is used here: https://github.com/matrix-org/synapse/blob/b588f16e391d664b11f43257eabf70663f0c6d59/synapse/handlers/presence.py#L1131-L1133

Note that there is a `get_all_presence_updates` function just above. That function is intended to walk up the table through stream IDs, and is primarily used by the presence replication stream. I could possibly make use of it in the PresenceRouter-related code, but it would be a bit of a bodge.

### DIFF
--- a/synapse/storage/databases/main/presence.py
+++ b/synapse/storage/databases/main/presence.py
@@ -174,7 +174,7 @@ class PresenceStore(SQLBaseStore):
         """
         users_to_state = {}
 
-        exclude_keyvalues = {}
+        exclude_keyvalues = None
         if not include_offline:
             # Exclude offline presence state
             exclude_keyvalues = {"state": "offline"}
@@ -191,7 +191,6 @@ class PresenceStore(SQLBaseStore):
                 orderby="stream_id",
                 start=offset,
                 limit=limit,
-                keyvalues={},
                 exclude_keyvalues=exclude_keyvalues,
                 retcols=(
                     "user_id",
@@ -208,7 +207,7 @@ class PresenceStore(SQLBaseStore):
             for row in rows:
                 users_to_state[row["user_id"]] = UserPresenceState(**row)
 
-            # We've ran out of updates to query
+            # We've run out of updates to query
             if len(rows) < limit:
                 break
 
--- a/synapse/storage/databases/main/presence.py
+++ b/synapse/storage/databases/main/presence.py
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import List, Tuple
+from typing import Dict, List, Tuple
 
 from synapse.api.presence import UserPresenceState
 from synapse.storage._base import SQLBaseStore, make_in_list_sql_clause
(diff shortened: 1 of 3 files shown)


synapse/storage/databases/main/presence.py
```diff
@@ -157,5 +157,64 @@ async def get_presence_for_users(self, user_ids):
...
+
+        Args:
+            include_offline: Whether to include offline presence states
+
+        Returns:
+            A dict of user IDs to their current UserPresenceState.
+        """
+        users_to_state = {}
+
+        exclude_keyvalues = {}
```
clokep, line 177:
```suggestion
        exclude_keyvalues = None
```

synapse/storage/databases/main/presence.py
```diff
@@ -157,5 +157,64 @@ async def get_presence_for_users(self, user_ids):
...
+        offset = 0
+        while True:
+            rows = await self.db_pool.runInteraction(
+                "get_presence_for_all_users",
+                self.db_pool.simple_select_list_paginate_txn,
+                "presence_stream",
+                orderby="stream_id",
+                start=offset,
+                limit=limit,
+                keyvalues={},
```
clokep, line 194: I think you can just leave this out?

synapse/storage/databases/main/presence.py
```diff
@@ -157,5 +157,64 @@ async def get_presence_for_users(self, user_ids):
...
+                    "status_msg",
+                    "currently_active",
+                ),
+                order_direction="ASC",
+            )
+
+            for row in rows:
+                users_to_state[row["user_id"]] = UserPresenceState(**row)
+
+            # We've ran out of updates to query
```
clokep, line 211:
```suggestion
            # We've run out of updates to query
```

synapse/storage/databases/main/presence.py
```diff
@@ -157,5 +157,64 @@ async def get_presence_for_users(self, user_ids):
...
+        users_to_state = {}
+
+        exclude_keyvalues = {}
+        if not include_offline:
+            # Exclude offline presence state
+            exclude_keyvalues = {"state": "offline"}
+
+        # This may be a very heavy database query.
+        # We paginate in order to not block a database connection.
+        limit = 100
```
clokep, line 184: 100 seems kind of low, I think we usually batch by 1000?

clokep (commented):
Looks pretty good, left a couple of comments.

synapse/storage/databases/main/presence.py
anoadragon453: The only other usage of this function I see uses a limit of 100:

https://github.com/matrix-org/synapse/blob/0a00b7ff14890987f09112a2ae696c61001e6cf1/synapse/storage/databases/main/stats.py#L297-L320

But intuitively 1000 is probably also fine, given the overhead of spinning up another query?

clokep (approved):

synapse/storage/databases/main/presence.py
clokep: Fair enough. We can tweak it if we need to.

PR #9652, opened 2021-03-19
Federation API for Space summary

richvdh:
Builds on the work done in #9643 to add a federation API for space summaries.

There's a bit of refactoring of the existing client-server code first, to avoid too much duplication.

### DIFF
--- a/synapse/federation/transport/server.py
+++ b/synapse/federation/transport/server.py
@@ -18,7 +18,7 @@
 import functools
 import logging
 import re
-from typing import Optional, Tuple, Type
+from typing import TYPE_CHECKING, Container, Mapping, Optional, Sequence, Tuple, Type
 
 import synapse
 from synapse.api.constants import MAX_GROUP_CATEGORYID_LENGTH, MAX_GROUP_ROLEID_LENGTH
@@ -29,7 +29,7 @@ from synapse.api.urls import (
     FEDERATION_V1_PREFIX,
     FEDERATION_V2_PREFIX,
 )
-from synapse.http.server import JsonResource
+from synapse.http.server import HttpServer, JsonResource
 from synapse.http.servlet import (
     parse_boolean_from_args,
     parse_integer_from_args,
@@ -44,9 +44,13 @@ from synapse.logging.opentracing import (
     whitelisted_homeserver,
 )
 from synapse.server import HomeServer
-from synapse.types import ThirdPartyInstanceID, get_domain_from_id
+from synapse.types import JsonDict, ThirdPartyInstanceID, get_domain_from_id
+from synapse.util.ratelimitutils import FederationRateLimiter
 from synapse.util.stringutils import parse_and_validate_server_name
 from synapse.util.versionstring import get_version_string
+
+if TYPE_CHECKING:
+    import synapse.server
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


synapse/federation/transport/server.py
```diff
@@ -1474,18 +1512,24 @@ async def on_GET(self, origin, content, query, room_id):
 )
 
 
-def register_servlets(hs, resource, authenticator, ratelimiter, servlet_groups=None):
+def register_servlets(
+    hs: "synapse.server.HomeServer",
```
clokep, line 1516: Out of curiosity I always use `from synapse.app.homeserver import HomeServer` for these type hints. Should we be using this instead? (Also any reason for using the full dotted module name here instead of a `from` import? We don't seem to do that often.)

synapse/federation/transport/server.py
```diff
@@ -1376,6 +1380,40 @@ async def on_PUT(self, origin, content, query, group_id):
...
+        self,
+        origin: str,
+        content: JsonDict,
+        query: Mapping[bytes, Sequence[bytes]],
+        room_id: str,
+    ) -> Tuple[int, JsonDict]:
+        suggested_only = content.get("suggested_only", False)
+        if not isinstance(suggested_only, bool):
+            raise SynapseError(
+                400, "'suggested_only' must be a boolean", Codes.BAD_JSON
```
clokep, line 1397: Should this (and below) be `INVALID_PARAM` instead of `BAD_JSON`?

synapse/handlers/space_summary.py
```diff
@@ -144,12 +200,25 @@ async def _summarize_local_room(
...
+
+        # otherwise, check if the room is peekable
+        hist_vis = ""
+        hist_vis_ev = await self._state_handler.get_current_state(
+            room_id, EventTypes.RoomHistoryVisibility, ""
+        )
+        if hist_vis_ev:
+            hist_vis = hist_vis_ev.content.get("history_visibility")
+        if hist_vis == HistoryVisibility.WORLD_READABLE:
             return True
```
clokep: I think it would be clearer to indent this since it can only be reached if the event exists.

synapse/handlers/space_summary.py
```diff
@@ -76,67 +78,154 @@ async def get_space_summary(
...
+        events_result = []  # type: List[JsonDict]
+        for edge_event in itertools.islice(child_events, max_children):
+            events_result.append(
+                await self._event_serializer.serialize_event(
+                    edge_event,
+                    time_now=now,
+                    event_format=format_event_for_client_v2,
                 )
+            )
+        return (room_entry,), events_result
```
clokep, line 201: Why wrap `room_entry` in a tuple first? Will this eventually return multiple rooms or something?

synapse/handlers/space_summary.py
```diff
@@ -76,67 +78,154 @@ async def get_space_summary(
...
+
+        return {"rooms": rooms_result, "events": events_result}
+
+    async def federation_space_summary(
+        self,
+        room_id: str,
+        suggested_only: bool,
+        max_rooms_per_space: Optional[int],
+        exclude_rooms: Iterable[str],
+    ) -> JsonDict:
```
clokep, line 119: Looks like this is pretty much identical to the C-S implementation, but it:

* Doesn't take a requester (and thus doesn't check if the user is in the room, falling back to whether the room is world-readable instead).
* Doesn't use `_RoomQueueEntry`.
* Falls back to the default for the maximum number of rooms per entry.

Am I missing anything else?

synapse/handlers/space_summary.py
```diff
@@ -191,6 +280,11 @@ async def _get_child_events(self, room_id: str) -> Iterable[EventBase]:
         return (e for e in events if e.content.get("via"))
 
 
+@attr.s(frozen=True, slots=True)
+class _RoomQueueEntry:
+    room_id = attr.ib(type=str)
```
clokep, line 285: Will the via stuff be added in a separate PR? It seems premature to separate this out.

synapse/federation/transport/server.py
richvdh: > Out of curiosity I always use from synapse.app.homeserver import HomeServer

There isn't a `HomeServer` in `synapse.app.homeserver` (except by import from `synapse.server`. Do you mean `SynapseHomeServer`? In which case, that won't really be accurate for a worker process.

> (Also any reason for using the full dotted module name here instead of a from import? We don't seem to do that often.)

Fair; I'll change this.

synapse/federation/transport/server.py
richvdh: I *think* `M_INVALID_PARAM` is for query-string params, though this is very unclear, and I've opened https://github.com/matrix-org/matrix-doc/issues/3070 about it.

synapse/handlers/space_summary.py
richvdh: true.

synapse/handlers/space_summary.py
richvdh: yeahh, it's mostly for symmetry with a `_summarize_remote_room` method in #9653, which can return multiple rooms.

It's worth noting that this method can also return zero rooms. I fiddled with making the whole result Optional, but it didn't really simplify anything.

synapse/handlers/space_summary.py
richvdh: That sounds about right so far. The differences get a bit bigger in #9653: it also doesn't ever call out over federation, which means it doesn't need to track `via`s (which is why it doesn't use `_RoomQueueEntry`.

synapse/handlers/space_summary.py
richvdh: sorry, yes.

synapse/federation/transport/server.py
clokep: I suspect I've made a bunch of imports in correct! Oops. I'll put up a PR to fix this.

synapse/federation/transport/server.py
clokep: I think we're inconsistent with this in Synapse then!

synapse/handlers/space_summary.py
clokep: Got it. Thanks! I was wondering why it was separate, but that makes sense that #9653 would have this differ more.

clokep (approved):
Seems good!

synapse/federation/transport/server.py
richvdh: s/synapse/the entire matrix ecosystem. yes.

Issue #9447, opened 2021-02-19
 support for spaces summary API (federated)


richvdh:
This issue is to extend the work started in #9446 to support queries over federation

PR #9653, opened 2021-03-19
Spaces summary: call out to other servers

richvdh:
When we hit an unknown room in the space tree, see if there are other servers that we might be able to poll to get the data.

~~Based on #9652~~.  Fixes: #9447

### DIFF
--- a/synapse/federation/federation_client.py
+++ b/synapse/federation/federation_client.py
@@ -27,11 +27,13 @@ from typing import (
     List,
     Mapping,
     Optional,
+    Sequence,
     Tuple,
     TypeVar,
     Union,
 )
 
+import attr
 from prometheus_client import Counter
 
 from twisted.internet import defer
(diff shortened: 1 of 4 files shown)


erikjohnston:
Initial thought here is that the functions and classes could do with some docstrings :)

richvdh:
> Initial thought here is that the functions and classes could do with some docstrings :)

I added some better docstrings to `federation_client`. Regarding the private methods in `SpaceSummaryHandler`, I've left them undocumented, because (a) they're private and (b) I think they're kinda obvious, but I could add some if you feel strongly.

synapse/handlers/space_summary.py
```diff
@@ -200,6 +254,43 @@ async def _summarize_local_room(
...
+        room: "_RoomQueueEntry",
+        suggested_only: bool,
+        max_children: Optional[int],
+        exclude_rooms: Iterable[str],
+    ) -> Tuple[Sequence[JsonDict], Sequence[JsonDict]]:
+        room_id = room.room_id
+        logger.info("Requesting summary for %s via %s", room_id, room.via)
+
+        # we need to make the exclusion list json-serialisable
+        exclude_rooms = list(exclude_rooms)
```
clokep, line 268: Does this mean that `get_space_summary` doesn't actually take a `Sequence[str]` for `exclude_rooms`, but actually needs a `List[str]`?

synapse/handlers/space_summary.py
```diff
@@ -200,6 +254,43 @@ async def _summarize_local_room(
...
+
+        via = itertools.islice(room.via, MAX_SERVERS_PER_SPACE)
+        try:
+            res = await self._federation_client.get_space_summary(
+                via,
+                room.room_id,
```
clokep, line 274: We already have `room_id` above, no need to grab it again from the object.

synapse/handlers/space_summary.py
```diff
@@ -200,6 +254,43 @@ async def _summarize_local_room(
...
+                suggested_only=suggested_only,
+                max_rooms_per_space=max_children,
+                exclude_rooms=exclude_rooms,
+            )
+        except Exception as e:
+            logger.warning(
+                "Unable to get summary of %s via federation: %s",
+                room_id,
+                e,
+                exc_info=logger.isEnabledFor(logging.DEBUG),
```
clokep, line 284: Why only send this if the debug logging is enabled?

synapse/handlers/space_summary.py
```diff
@@ -149,20 +200,23 @@ async def federation_space_summary(
 
         while room_queue and len(rooms_result) < MAX_ROOMS:
             room_id = room_queue.popleft()
+            if room_id in processed_rooms:
+                # already done this room
+                continue
```
clokep, line 205: This would happen because we now don't check below whether the edge is processed already while extending the queue?

synapse/handlers/space_summary.py
```diff
@@ -78,35 +83,81 @@ async def get_space_summary(
...
+            # any rooms returned don't need visiting again
+            processed_rooms.update(cast(str, room.get("room_id")) for room in rooms)
+
+            # the room we queried may or may not have been returned, but don't process
+            # it again, anyway.
+            processed_rooms.add(room_id)
+
+            # XXX: is it ok that we blindly iterate through any events returned by
+            #   a remote server, whether or not they actually link to any rooms in our
+            #   tree?
```
clokep, line 144: I'm not really sure of the implications of this, we're not persisting any of this as state, so joining the room should get the proper data?

synapse/federation/federation_client.py
```diff
@@ -495,16 +500,31 @@ async def _try_destination_list(
...
+                elif failover_on_unknown_endpoint:
+                    # there is no good way to detect an "unknown" endpoint. Dendrite
+                    # returns a 404 (with no body); synapse returns a 400
+                    # with M_UNRECOGNISED.
+                    if e.code == 404 or (
+                        e.code == 400 and synapse_error.errcode == Codes.UNRECOGNIZED
+                    ):
+                        failover = True
+
+                if not failover:
```
clokep, line 518: Should this be `if failover`? I'm having some trouble following this logic. Previously the `500` error would raise a `SynapseError`, but now it won't (since `failover` is `True` in that case).

synapse/handlers/space_summary.py
richvdh: yes. It's annotated as such: https://github.com/matrix-org/synapse/pull/9653/files#diff-3cf3f534d150a0319b7bdd81c884daff978bba8f8c603d7f023d5dd5d73e1ab1R1074

synapse/handlers/space_summary.py
richvdh: basically because in 90% of cases, it's going to be overly verbose: it'll be due to servers not supporting the endpoint, or having left the room, or being down, or whatever - none of which merit a stacktrace. On the other hand, if it's going wrong for reasons we don't understand, it'll be useful to be able to see the stacktrace.

synapse/handlers/space_summary.py
richvdh: it can happen because we might find two edges pointing to the same room. I think this was actually a bug in the original impl, I just hadn't found it yet.

Example: A has two children, B and C. C is also a child of B.  First we process A, and add the discovered children (B and C) to the queue. We pop B off the queue, and add *its* children (C) to the queue. C is now in the queue twice. Without this line, we end up processing it twice.

synapse/handlers/space_summary.py
richvdh: yes it should.

The concern here is that a malicious or buggy remote server could return a load of events (and rooms, for that matter) which are nothing to do with the requested room. That means that we'll end up spending resources looking for rooms we don't actually care about, and will return that data to the client (which might either ignore it, or (incorrectly) add it to the UI). Still, there shouldn't be any lasting effect here.

synapse/federation/federation_client.py
richvdh: > Previously the 500 error would raise a SynapseError

I don't think that's right. Previously anything *except* a 500 error would raise a `SynapseError`.

I agree the logic here isn't super easy to follow. I'm very open to suggestions for how to reorganise it to make it clearer.

I was actually contemplating this logic in the shower this morning... Really, there ought to be a defined list of error codes that a well-behaved server can return for each endpoint - and anything else constitutes a misbehaving server whose response should be discounted. The trouble is that implementing that means getting into the weeds with refactoring the other code that uses `_try_destination_list`.

synapse/handlers/space_summary.py
richvdh: indeed. done.

synapse/federation/federation_client.py
clokep: Ah, yeah I had missed the `not` in front of the first clause. 😢 

I think the newer version is a bit clearer with the `failover` variable. I was just failing to see the equivalence to the old one.

I don't have any suggestions of how to make it better now. 👍

clokep (approved):
Let's give it a try! 🎉

Issue #8683, opened 2020-10-29
Improve structured logging


clokep:
Building on #8588 there's a desire to improve the output of Synapse structured logging.

Currently a "standard" log line would be turned into JSON (this was a curl to `/_matrix/client/r0/login` to get the login flows, but other logs look similar):

```
2020-10-29 07:36:00,762 - synapse.access.http.8080 - 315 - INFO - GET-0 - - - 8080 - {None} Processed request: 0.000sec/-0.000sec (0.000sec, 0.000sec) (0.000sec/0.000sec/0) 285B 200 "GET /_matrix/client/r0/login HTTP/1.1" "curl/7.64.1" [0 dbevts]
```

This would transform to (note that whitespace was added for visibility, normally this would be a single line with no whitespaec between JSON separators):

```json
{
    "log": "- - 8080 - {None} Processed request: 0.000sec/-0.000sec (0.000sec, 0.000sec) (0.000sec/0.000sec/0) 285B 200 \"GET /_matrix/client/r0/login HTTP/1.1\" \"curl/7.64.1\" [0 dbevts]",
    "namespace": "synapse.access.http.8080",
    "level": "INFO",
    "time": 1603971360.76,
    "request": "GET-0",
    "scope": null,
    "server_name": "localhost:8480"
}
```

A few improvements we'll want to make:

* The log line includes a formatted message, which makes it difficult to search in a structured way.
* Currently logging a message with additional fields requires including the fields twice (using the standard logging flow and the `extra` keyword to the logger).

clokep:
To update this a bit:

* The code to create the structured logs has been cleaned up quite a bit so I think this issue is now in a spot to include additional info on a per-log line basis.
* We now include additional fields in the structed logs due to #9654, which should make them more usable.

So the description here is still pretty valid (including the formatted log lines is not ideal), but much of the logging context is now included.

I'm no longer actively working on this so unassigning.

PR #9654, opened 2021-03-19
Record more information into structured logs

clokep:
This was some work I had going towards #8683, it adds some basic request information to the structured logs (similar to information available in the [processed log lines](https://github.com/matrix-org/synapse/wiki#what-do-all-those-fields-in-the-processed-line-mean)).

In particular this should allow pivoting on things like IP address or user ID in structured logs.

In order to do this there was a bit of refactoring necessary to `LoggingContext`. Each commit should be separately reviewable.

### DIFF
--- a/synapse/http/site.py
+++ b/synapse/http/site.py
@@ -14,7 +14,7 @@
 import contextlib
 import logging
 import time
-from typing import Optional, Type, Union
+from typing import Optional, Tuple, Type, Union
 
 import attr
 from zope.interface import implementer
@@ -126,27 +126,38 @@ class SynapseRequest(Request):
             return self.method.decode("ascii")
         return method
 
-    def get_authenticated_entity(self) -> Optional[str]:
+    def get_authenticated_entity(self) -> Tuple[Optional[str], Optional[str]]:
+        """
+        Get the "authenticated" entity of the request, which might be the user
+        performing the action, or a user being puppeted by a server admin.
+
+        Returns:
+            A tuple:
+                The first item is a string representing the user making the request.
+
+                The second item is a string or None representing the user who
+                authenticated when making this request. See
+                Requester.authenticated_entity.
+        """
         # Convert the requester into a string that we can log
         if isinstance(self.requester, str):
-            return self.requester
+            return self.requester, None
         elif isinstance(self.requester, Requester):
+            requester = self.requester.user.to_string()
             authenticated_entity = self.requester.authenticated_entity
 
             # If this is a request where the target user doesn't match the user who
-            # authenticated (e.g. and admin is puppetting a user) then we log both.
+            # authenticated (e.g. and admin is puppetting a user) then we return both.
             if self.requester.user.to_string() != authenticated_entity:
-                return "{},{}".format(
-                    authenticated_entity,
-                    self.requester.user.to_string(),
-                )
-            return authenticated_entity
+                return requester, authenticated_entity
+
+            return requester, None
         elif self.requester is not None:
             # This shouldn't happen, but we log it so we don't lose information
             # and can see that we're doing something wrong.
-            return repr(self.requester)  # type: ignore[unreachable]
+            return repr(self.requester), None  # type: ignore[unreachable]
 
-        return None
+        return None, None
 
     def render(self, resrc):
         # this is called once a Resource has been found to serve the request; in our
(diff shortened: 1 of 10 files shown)


synapse/http/site.py
```diff
@@ -126,13 +126,35 @@ def get_method(self) -> str:
             return self.method.decode("ascii")
         return method
 
+    def get_authenticated_entity(self) -> Optional[str]:
```
clokep, line 129: This feels like it should potentially be two pieces of information in the structured logs? An authenticated entity and acted entity or something.

synapse/http/site.py
```diff
@@ -334,6 +356,9 @@ def _finished_processing(self):
         except Exception as e:
             logger.warning("Failed to stop metrics: %r", e)
 
+        # Break the cycle between SynapseRequest and LoggingContext.
+        self.logcontext = None
```
clokep, line 360: I don't remember why I felt it was necessary to manually break this, but it seems plausible.

erikjohnston: Hmm, I'm a bit concerned that we do `assert self.logcontext` in various places and its a bit non-obvious to me that they will never get called after this point.

synapse/http/site.py
erikjohnston: Yeah, probably

synapse/http/site.py
```diff
@@ -312,7 +334,7 @@ def _finished_processing(self):
             ' %sB %s "%s %s %s" "%s" [%d dbevts]',
             self.getClientIP(),
             self.site.site_tag,
-            authenticated_entity,
```
erikjohnston, line 315: Can we also delete the code calculating this above?

synapse/http/site.py
clokep: This PR only adds one of these, I think, to `finish`, which immediately calls `_finished_processing` where the other one is.

I believe that `_finished_processing` can be called once (and only once), so this should be safe.

clokep: Alternately -- I suppose I could just run in the sentinel context then or something? Or do you have suggestions of what to do?

erikjohnston: I'd rather it exploded than accidentally record stuff against the sentinel context. Ideally our log contexts would be robust enough where we could trust tat we could just do `current_context()`. Maybe instead of having a pointer to the request in the log context we should store the actual values we care about? It's more space but a) no circular dependency and b) we don't recalculate the values every time we log something?

clokep: > Maybe instead of having a pointer to the request in the log context we should store the actual values we care about? It's more space but a) no circular dependency and b) we don't recalculate the values every time we log something?

I think that would work, the annoying bit becomes that there will be a couple of places that those need to be listed (since we have to handle the child-parent context relationship). I'll think a bit more about it!

erikjohnston: We could have a `RequestMetadata` struct that we point to from various places? I don't know. I'm not a massive fan of the current state of affairs but I don't think we should block this PR on it too much

clokep: I had the same though last night. 👍

clokep: I made this change, but it got a little funky since `requester` gets set later, so I needed to make it a `setter`...

synapse/logging/context.py
```diff
@@ -181,6 +180,42 @@ def __sub__(self, other: "ContextResourceUsage") -> "ContextResourceUsage":
         return res
 
 
+class ContextRequest:
```
clokep, line 183: This could probably use `attrs`, but we don't seem to in other spots in this file...

synapse/http/site.py
erikjohnston: Thanks!

synapse/logging/context.py
erikjohnston: I think that's just because the file is old, can we make it an attrs please? That way we can have type hints

erikjohnston, line 183: Can we have a docstring too please

synapse/http/site.py
```diff
@@ -126,13 +155,60 @@ def get_method(self) -> str:
...
+                self.get_request_id(),
+                self.getClientIP(),
+                self.site.site_tag,
+                # The requester is going to be unknown at this point.
+                None,
+                None,
+                self.get_method(),
+                self.get_redacted_uri(),
+                self.clientproto.decode("ascii", errors="replace"),
+                get_request_user_agent(self),
```
erikjohnston, line 209: Can these be kwargs please, it feels easy to accidentally get some of these the wrong way round

erikjohnston (approved):

PR #9655, opened 2021-03-19
Implement MSC3069: Guest support on whoami

turt2live:
MSC: https://github.com/matrix-org/matrix-doc/pull/3069


anoadragon453 (changes requested):
I would like to see a test for this, but otherwise it lgtm.

We have an example of registering a guest user [here](https://github.com/matrix-org/synapse/blob/394516ad1bb6127ab5b32a12d81ef307deb39570/tests/rest/client/v2_alpha/test_register.py#L126-L134) (although I would use the `@override_config({"allow_guest_access": True})` test decorator instead of setting the option on the homeserver directly).

turt2live:
@anoadragon453 while adding tests I noticed the deactivate test wasn't actually testing the access token - I've fixed this here, but happy to pull it out to its own PR if needed.

tests/rest/client/v2_alpha/test_account.py
```diff
@@ -398,7 +398,7 @@ def test_deactivate_account(self):
         self.assertTrue(self.get_success(store.get_user_deactivated_status(user_id)))
 
         # Check that this access token has been invalidated.
-        channel = self.make_request("GET", "account/whoami")
+        channel = self.make_request("GET", "account/whoami", access_token=tok)
```
anoadragon453, line 401: This seems fine to leave in this PR.

anoadragon453 (changes requested):
Thanks for writing the tests! Some words for helping them pass are below.

turt2live:
err, sorry - this wasn't ready for re-review just yet. Thanks for answering all the questions I was about to ask though :D

anoadragon453:
@turt2live just checking that you intend to keep working on this?

turt2live:
One day, yes. There's a small timing problem with releases on my part that I need to work out, but otherwise it's nearly there. 

Someone else could fairly easily pick this up if they wanted.

turt2live:
It doesn't seem fair to throw this back at anoa after months of inactivity, so sending back to the team.

DMRobertson (changes requested):
Looks reasonable to me! I think you need to run the linter script though. After that we can see if the new tests pass.

DMRobertson (approved):
LGTM!

synapse/rest/client/account.py
```diff
@@ -878,9 +878,13 @@ def __init__(self, hs: "HomeServer"):
...
 
     async def on_GET(self, request: SynapseRequest) -> Tuple[int, JsonDict]:
-        requester = await self.auth.get_user_by_req(request)
+        requester = await self.auth.get_user_by_req(request, allow_guest=True)
 
-        response = {"user_id": requester.user.to_string()}
+        response = {
+            "user_id": requester.user.to_string(),
+            # MSC: https://github.com/matrix-org/matrix-doc/pull/3069
+            "org.matrix.msc3069.is_guest": bool(requester.is_guest),
```
richvdh, line 886: For future reference: I'd have preferred this to be behind an opt-in config flag.

(see https://matrix-org.github.io/synapse/latest/development/experimental_features.html)

turt2live: It's not really a feature, but more of a very tedious spec bug fix.

richvdh: nevertheless, I'd prefer Synapse not to expose the unstable-named flag without people explicitly opting into it.

PR #9657, opened 2021-03-19
Allow providing credentials to HTTPS_PROXY

anoadragon453:
Addresses https://github.com/matrix-org/synapse-dinsic/issues/70

This PR causes `ProxyAgent` to attempt to extract credentials from an `HTTPS_PROXY` env var. If credentials are found, a `Proxy-Authorization` header ([details](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Proxy-Authorization)) is sent to the proxy server to authenticate against it. The headers are *not* passed to the remote server.

Also added some type hints.

Reviewable commit-by-commit.

### DIFF
--- a/changelog.d/9657.feature
+++ b/changelog.d/9657.feature
@@ -1,1 +1,1 @@
-Support credentials for proxy authentication in the `HTTPS_PROXY` environment variable.
+Add support for credentials for proxy authentication in the `HTTPS_PROXY` environment variable.
--- a/changelog.d/9657.feature
+++ b/changelog.d/9657.feature
@@ -1,0 +1,1 @@
+Support credentials for proxy authentication in the `HTTPS_PROXY` environment variable.
--- a/synapse/http/proxyagent.py
+++ b/synapse/http/proxyagent.py
@@ -12,10 +12,13 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import base64
 import logging
 import re
+from typing import Optional, Tuple
 from urllib.request import getproxies_environment, proxy_bypass_environment
 
+import attr
 from zope.interface import implementer
 
 from twisted.internet import defer
@@ -23,6 +26,7 @@ from twisted.internet.endpoints import HostnameEndpoint, wrapClientTLS
 from twisted.python.failure import Failure
 from twisted.web.client import URI, BrowserLikePolicyForHTTPS, _AgentBase
 from twisted.web.error import SchemeNotSupported
+from twisted.web.http_headers import Headers
 from twisted.web.iweb import IAgent
 
 from synapse.http.connectproxyclient import HTTPConnectProxyEndpoint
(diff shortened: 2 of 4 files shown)


synapse/http/proxyagent.py
```diff
@@ -223,16 +257,43 @@ def _http_proxy_endpoint(proxy, reactor, **kwargs):
     if proxy is None:
         return None
 
-    # currently we only support hostname:port. Some apps also support
-    # protocol://<host>[:port], which allows a way of requiring a TLS connection to the
-    # proxy.
```
erikjohnston, line 228: This comment still seems valid?

tests/http/test_proxyagent.py
```diff
@@ -240,7 +241,10 @@ def test_http_request_via_proxy(self):
         body = self.successResultOf(treq.content(resp))
         self.assertEqual(body, b"result")
 
-    @patch.dict(os.environ, {"https_proxy": "proxy.com", "no_proxy": "unused.com"})
+    @patch.dict(
+        os.environ,
+        {"https_proxy": "bob:pinkponies@proxy.com", "no_proxy": "unused.com"},
+    )
```
erikjohnston, line 247: Can you make a new test please? So that we test both paths.

erikjohnston (changes requested):
Mostly LGTM

synapse/http/proxyagent.py
anoadragon453: Sort of. We support `HTTP_PROXY` and `HTTPS_PROXY`, which does allow us to require a TLS connection to the proxy. But that's done via two separate environment variables, instead of specifying the `protocol` to a single env var.

Because of that I'm not sure if we still needed to include the bit about TLS.

tests/http/test_proxyagent.py
anoadragon453: Sure. I'll try to de-duplicate the code a bit between them.

synapse/http/proxyagent.py
erikjohnston: I think the pertinent point here is that there are different formats for this env var and we only support a subset of formats.

anoadragon453: That's fair. I'll try to work that back in!

anoadragon453: 0c5a5b9 is hopefully a bit clearer?

erikjohnston: Thanks! It just feels a bit sad to lose that information when its already there

changelog.d/9657.feature
```diff
@@ -0,0 +1 @@
+Support credentials for proxy authentication in the `HTTPS_PROXY` environment variable.
```
erikjohnston, line 1:
```suggestion
Add support for credentials for proxy authentication in the `HTTPS_PROXY` environment variable.
```

erikjohnston (approved):

synapse/http/proxyagent.py
anoadragon453: Yeah I'd missed the underlying point of the comment, assuming that it was old and just referring to the lack of being able to use TLS with a proxy :slightly_smiling_face:

PR #9659, opened 2021-03-21
Add more bugbear codes

ShadowJonathan:
Part of #9366 

Adds in more test codes except `B006`-`B008`.

### Pull Request Checklist

### DIFF
--- a/changelog.d/9659.misc
+++ b/changelog.d/9659.misc
@@ -1,0 +1,1 @@
+Add and fix more bugbear lint violations.
--- a/changelog.d/9659.misc
+++ b/changelog.d/9659.misc
@@ -1,1 +1,1 @@
-Add and fix more bugbear lint violations.
+Introduce bugbear to the test suite and fix some of it's lint violations.
--- a/synapse/crypto/context_factory.py
+++ b/synapse/crypto/context_factory.py
@@ -191,9 +191,9 @@ def _context_info_cb(ssl_connection, where, ret):
         # ... we further assume that SSLClientConnectionCreator has set the
         # '_synapse_tls_verifier' attribute to a ConnectionVerifier object.
         tls_protocol._synapse_tls_verifier.verify_context_info_cb(ssl_connection, where)
-    except BaseException as e:  # taken from the twisted implementation
+    except BaseException:  # taken from the twisted implementation
         logger.exception("Error during info_callback")
-        f = Failure(e)
+        f = Failure()
         tls_protocol.failVerification(f)
 
 
--- a/synapse/crypto/context_factory.py
+++ b/synapse/crypto/context_factory.py
@@ -191,9 +191,9 @@ def _context_info_cb(ssl_connection, where, ret):
         # ... we further assume that SSLClientConnectionCreator has set the
         # '_synapse_tls_verifier' attribute to a ConnectionVerifier object.
         tls_protocol._synapse_tls_verifier.verify_context_info_cb(ssl_connection, where)
-    except:  # noqa: E722, taken from the twisted implementation
+    except BaseException as e:  # taken from the twisted implementation
         logger.exception("Error during info_callback")
-        f = Failure()
+        f = Failure(e)
         tls_protocol.failVerification(f)
 
 
--- a/synapse/logging/context.py
+++ b/synapse/logging/context.py
@@ -689,10 +689,10 @@ def run_in_background(f, *args, **kwargs) -> defer.Deferred:
     current = current_context()
     try:
         res = f(*args, **kwargs)
-    except:  # noqa: E722
+    except Exception as e:
         # the assumption here is that the caller doesn't want to be disturbed
         # by synchronous exceptions, so let's turn them into Failures.
-        return defer.fail()
+        return defer.fail(e)
 
     if isinstance(res, types.CoroutineType):
         res = defer.ensureDeferred(res)
--- a/synapse/logging/context.py
+++ b/synapse/logging/context.py
@@ -689,10 +689,10 @@ def run_in_background(f, *args, **kwargs) -> defer.Deferred:
     current = current_context()
     try:
         res = f(*args, **kwargs)
-    except Exception as e:
+    except Exception:
         # the assumption here is that the caller doesn't want to be disturbed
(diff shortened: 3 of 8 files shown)


synapse/logging/context.py
```diff
@@ -689,10 +689,10 @@ def run_in_background(f, *args, **kwargs) -> defer.Deferred:
     current = current_context()
     try:
         res = f(*args, **kwargs)
-    except:  # noqa: E722
+    except Exception as e:
```
ShadowJonathan, line 692: explicit > implicit

I know that `defer.fail()` will take the current exception even if this wasn't captured, but imo it's more friendly to read this way.

synapse/crypto/context_factory.py
```diff
@@ -191,9 +191,9 @@ def _context_info_cb(ssl_connection, where, ret):
         # ... we further assume that SSLClientConnectionCreator has set the
         # '_synapse_tls_verifier' attribute to a ConnectionVerifier object.
         tls_protocol._synapse_tls_verifier.verify_context_info_cb(ssl_connection, where)
-    except:  # noqa: E722, taken from the twisted implementation
+    except BaseException as e:  # taken from the twisted implementation
```
ShadowJonathan, line 194: I'm not entirely sure if this is correct, but "taken from the twisted implementation" tells me that twisted (more or less) explicitly intended to also capture `BaseException` as well.

tests/server.py
```diff
@@ -593,7 +593,7 @@ def flush(self, maxbytes=None):
         if self.disconnected:
             return
 
-        if getattr(self.other, "transport") is None:
+        if hasattr(self.other, "transport"):
```
ShadowJonathan, line 596: From [this review chain](https://github.com/matrix-org/synapse/pull/9649#discussion_r597986898) I assume `hasattr` was originally intended here, instead of `getattr`.

tests/server.py
```diff
@@ -593,7 +593,7 @@ def flush(self, maxbytes=None):
         if self.disconnected:
             return
 
-        if getattr(self.other, "transport") is None:
+        if not hasattr(self.other, "transport"):
```
ShadowJonathan, line 596: (copy from prev comment)

 From [this review chain](https://github.com/matrix-org/synapse/pull/9649#discussion_r597986898) I assume `hasattr` was originally intended here, instead of `getattr`.
 
 I accidentally changed this to just `hasattr` before, no `not`, which froze the tests.

synapse/storage/database.py
```diff
@@ -670,10 +670,10 @@ async def runInteraction(
 
             for after_callback, after_args, after_kwargs in after_callbacks:
                 after_callback(*after_args, **after_kwargs)
-        except:  # noqa: E722, as we reraise the exception this is fine.
+        except Exception as e:
             for after_callback, after_args, after_kwargs in exception_callbacks:
                 after_callback(*after_args, **after_kwargs)
-            raise
+            raise e
```
clokep, line 676: I think there's no reason to declare `e` (and then raise) in this case:

```python
except Exception:
    ...
    raise
```

Should be equivalent.

synapse/util/async_helpers.py
```diff
@@ -496,8 +496,8 @@ def time_it_out():
 
         try:
             deferred.cancel()
-        except:  # noqa: E722, if we throw any exception it'll break time outs
-            logger.exception("Canceller failed during timeout")
+        except Exception as e:  # if we throw any exception it'll break time outs
+            logger.exception("Canceller failed during timeout", exc_info=e)
```
clokep, line 500: Doesn't `logger.exception` pull this automatically? It seems better to not declare it if we don't have to.

tests/server.py
clokep: This seems like what this was trying to do, yes. 👍

synapse/crypto/context_factory.py
clokep: I think we prefer *not* passing the error to `Failure` since that doesn't magic to pull out the "real" failure (or a failure with a better stack trace or something).

ShadowJonathan: Oh huh, I'll change it then 👍

synapse/util/async_helpers.py
ShadowJonathan: I like explicitness over implicitness, i think this is easier to read.

Same to the other comment.

Though if this is really not desired, I can change this back.

clokep: Please change it back.

I get the explicit vs. implicit argument, but it also adds fragility in case APIs ever change. (No, I don't really think this API will change, but philosophically if we can avoid passing things in it avoids brittleness and is easier to read.)

ShadowJonathan: I'll change it back. 👍

I don't think python really wants to emphasize itself around implicitness, but I do have to agree that with the presence of twisted in the codebase (which does a lot of implicit magic), things are a lot more fragile.

In general, i think python is mainly built around being explicit first and foremost, so that's where my intention and argument comes from.

changelog.d/9659.misc
```diff
@@ -0,0 +1 @@
+Add and fix more bugbear lint violations.
```
clokep, line 1: To match #9499:

```suggestion
Introduce bugbear to the test suite and fix some of it's lint violations.
```

clokep (approved):
LGTM if CI passes!

changelog.d/9659.misc
ShadowJonathan: I'll duplicate this on every subsequent PR, then :+1:

clokep: They get merged automatically by towncrier so this makes it a bit easier.

changelog.d/9659.misc
```diff
@@ -0,0 +1 @@
+Introduce bugbear to the test suite and fix some of it's lint violations.
```
richvdh, line 1: *its*

:innocent:

clokep: I'll fix both renditions of this. 😄

clokep: c2d4bd62a26ea0f19445aa23367fa7ee1b4e5b08

PR #9661, opened 2021-03-22
Proof of concept for GitHub Actions

callahad:
Compare to https://github.com/matrix-org/pipelines/blob/master/synapse/pipeline.yml

Per discussion in https://github.com/matrix-org/synapse/discussions/9622

(I _think_ this is complete, aside from double-checking logs to and log uploading from the Buildkite pipeline)

### Pull Request Checklist

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -180,8 +180,9 @@ jobs:
           || true
 
   sytest:
-    if: ${{ !failure() }}
-    needs: linting-done
+    if: ${{ always() }}
+    #if: ${{ !failure() }}
+    #needs: linting-done
     runs-on: ubuntu-latest
     container:
       image: matrixdotorg/sytest-synapse:${{ matrix.sytest-tag }}
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -180,9 +180,8 @@ jobs:
           || true
 
   sytest:
-    if: ${{ always() }}
-    #if: ${{ !failure() }}
-    #needs: linting-done
+    if: ${{ !failure() }}
+    needs: linting-done
     runs-on: ubuntu-latest
     container:
       image: matrixdotorg/sytest-synapse:${{ matrix.sytest-tag }}
@@ -229,9 +228,11 @@ jobs:
         run: /bootstrap.sh synapse
         working-directory: /src
       - name: Dump results.tap
+        if: ${{ always() }}
         run: cat /logs/results.tap
-      - name: Upload logs
+      - name: Upload SyTest logs
         uses: actions/upload-artifact@v2
+        if: ${{ always() }}
         with:
           name: Sytest Logs (${{ join(matrix.*, ', ') }})
           path: |
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -114,10 +114,9 @@ jobs:
           SYNAPSE_POSTGRES_HOST: localhost
           SYNAPSE_POSTGRES_USER: postgres
           SYNAPSE_POSTGRES_PASSWORD: postgres
-      - name: Dump logs #on failure
+      - name: Dump logs
         # Note: Dumps to workflow logs instead of using actions/upload-artifact
         #       This keeps logs colocated with failing jobs
-        #if: ${{ failure() }}
         run: >-
           find _trial_temp -name '*.log'
           -exec echo "::group::{}" \;
@@ -137,6 +136,14 @@ jobs:
           entrypoint: .buildkite/scripts/test_old_deps.sh
         env:
           TRIAL_FLAGS: "--jobs=2"
+      - name: Dump logs
+        # Note: Dumps to workflow logs instead of using actions/upload-artifact
+        #       This keeps logs colocated with failing jobs
+        run: >-
+          find _trial_temp -name '*.log'
+          -exec echo "::group::{}" \;
+          -exec cat {} \;
+          -exec echo "::endgroup::" \;
 
   trial-pypy:
     # Very slow; only run if the branch name includes 'pypy'
@@ -157,6 +164,14 @@ jobs:
       - run: tox -e py,combine
         env:
           TRIAL_FLAGS: "--jobs=2"
+      - name: Dump logs
+        # Note: Dumps to workflow logs instead of using actions/upload-artifact
+        #       This keeps logs colocated with failing jobs
+        run: >-
+          find _trial_temp -name '*.log'
+          -exec echo "::group::{}" \;
+          -exec cat {} \;
+          -exec echo "::endgroup::" \;
 
   sytest:
     if: ${{ !failure() }}
@@ -206,6 +221,14 @@ jobs:
       - name: Run SyTest
         run: /bootstrap.sh synapse
         working-directory: /src
+      - name: Dump logs
+        # Note: Dumps to workflow logs instead of using actions/upload-artifact
+        #       This keeps logs colocated with failing jobs
+        run: >-
+          find logs -name '*.log*' -o -name 'results.tap'
+          -exec echo "::group::{}" \;
+          -exec cat {} \;
+          -exec echo "::endgroup::" \;
 
   portdb:
     if: ${{ !failure() }} # Allow previous steps to be skipped, but not fail
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -173,6 +173,7 @@ jobs:
         BLACKLIST: ${{ matrix.workers && 'synapse-blacklist-with-workers' }}
 
     strategy:
+      fail-fast: false
       matrix:
         include:
           - sytest-tag: py35
@@ -201,7 +202,6 @@ jobs:
     - name: Prepare test blacklist
       run: cat sytest-blacklist .buildkite/worker-blacklist > synapse-blacklist-with-workers
     - name: Run SyTest
-      continue-on-error: true  # FIXME
       run: /bootstrap.sh synapse
       working-directory: /src
 
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -117,11 +117,13 @@ jobs:
       - name: Dump logs
         # Note: Dumps to workflow logs instead of using actions/upload-artifact
         #       This keeps logs colocated with failing jobs
+        #       It also ignores find's exit code; this is a best effort affair
         run: >-
           find _trial_temp -name '*.log'
           -exec echo "::group::{}" \;
           -exec cat {} \;
           -exec echo "::endgroup::" \;
+          || true
 
   trial-olddeps:
     if: ${{ !failure() }} # Allow previous steps to be skipped, but not fail
@@ -139,11 +141,13 @@ jobs:
       - name: Dump logs
         # Note: Dumps to workflow logs instead of using actions/upload-artifact
         #       This keeps logs colocated with failing jobs
+        #       It also ignores find's exit code; this is a best effort affair
         run: >-
           find _trial_temp -name '*.log'
           -exec echo "::group::{}" \;
           -exec cat {} \;
           -exec echo "::endgroup::" \;
+          || true
 
   trial-pypy:
     # Very slow; only run if the branch name includes 'pypy'
@@ -167,11 +171,13 @@ jobs:
       - name: Dump logs
         # Note: Dumps to workflow logs instead of using actions/upload-artifact
         #       This keeps logs colocated with failing jobs
+        #       It also ignores find's exit code; this is a best effort affair
         run: >-
           find _trial_temp -name '*.log'
           -exec echo "::group::{}" \;
           -exec cat {} \;
           -exec echo "::endgroup::" \;
+          || true
 
   sytest:
     if: ${{ !failure() }}
@@ -224,11 +230,13 @@ jobs:
       - name: Dump logs
         # Note: Dumps to workflow logs instead of using actions/upload-artifact
         #       This keeps logs colocated with failing jobs
+        #       It also ignores find's exit code; this is a best effort affair
         run: >-
-          find logs -name '*.log*' -o -name 'results.tap'
+          find /logs -name '*.log*' -o -name 'results.tap'
           -exec echo "::group::{}" \;
           -exec cat {} \;
           -exec echo "::endgroup::" \;
+          || true
 
   portdb:
     if: ${{ !failure() }} # Allow previous steps to be skipped, but not fail
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -31,7 +31,7 @@ jobs:
         run: scripts-dev/check_line_terminators.sh
 
   lint-newsfile:
-    if: ${{ github.base_ref == 'develop'  || contains(github.ref, 'release-') }}
+    if: ${{ github.base_ref == 'develop'  || contains(github.base_ref, 'release-') }}
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


.github/workflows/tests.yml
```diff
@@ -0,0 +1,289 @@
...
+            postgres: postgres
+            workers: workers
+            redis: redis
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Prepare test blacklist
+      run: cat sytest-blacklist .buildkite/worker-blacklist > synapse-blacklist-with-workers
+    - name: Run SyTest
+      continue-on-error: true  # FIXME
```
ShadowJonathan, line 204: why's this a `FIXME`?

.github/workflows/tests.yml
```diff
@@ -0,0 +1,289 @@
...
+      with:
+        workdir: /github/workspace
+        entrypoint: .buildkite/scripts/test_old_deps.sh
+      env:
+        TRIAL_FLAGS: "--jobs=2"
+
+
+  trial-pypy:
+    # Very slow; only run if the branch name includes 'pypy'
+    if: ${{ contains(github.ref, 'pypy') && !failure() }}
```
ShadowJonathan, line 140: Could it be made that it also runs if the PR/commit name has "pypy"? (preferably PR name)

.github/workflows/tests.yml
```diff
@@ -0,0 +1,289 @@
...
+
+
+  lint-sdist:
+    if: ${{ contains(github.ref, 'release-') }}
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-python@v2
+      with:
+        python-version: '3.x'
```
ShadowJonathan, line 56: What python version does this resolve to? 3.6?

I recommend setting this to the minimally-supported version, and/or do a matrix between the minimal + latest supported version.

.github/workflows/tests.yml
callahad: If one SyTest fails we should fail the whole matrix... but currently SyTests are a bit flakey :(

ShadowJonathan:
For some reason, [this test](https://github.com/matrix-org/synapse/runs/2164468895) flaked, but it didn't mark the job as failed.

.github/workflows/tests.yml
callahad: It could, but I think the expression would be quite a bit longer?

I think it'd be something like `github.event == 'pull_request' && (contains(github.event.pull_request.title, 'pypy') || contains(github.ref, 'pypy') ) && !failure())`?

.github/workflows/tests.yml
callahad: This resolves to the latest 3.x version; don't see a reason to matrix just for checking that wheels and sdists build... either they do or they don't.

callahad:
Oh weird. That was failing properly before I converted that whole job to run inside the sytest container... will poke at it shortly.

callahad:
(Or if you want to try to fix it yourself, turn on Actions in your fork of this repo, push up a branch to your fork, and you should be able to run and iterate on the tests to your heart's desire!)

.github/workflows/tests.yml
ShadowJonathan: Looks good, and I don't think the expression being longer is a problem, most people wouldn't see this unless they were editing this specific part of the pipeline.

Adding some comments to deconstruct what is happening could help, tho.

ShadowJonathan:
Thanks, I'll try to see if i have time this week, i just left behind a few comments on some things i noticed while glancing :P

callahad:
Ooh, I misunderstood `step.continue-on-failure`. What I really wanted was `strategy.fail-fast`

callahad:
I believe this is ready to review.

Deviations from our [buildkite pipeline](https://github.com/matrix-org/pipelines/blob/master/synapse/pipeline.yml):

- I'm not excluding `dinsic` branches from tox and mypy tests... if we still want that we can add it back in after merging the broader PR, but I want to keep the logic as clear as possible initially.
- I only check for a newsfile for PRs into develop, which I _think_ is semantically correct. Happy to broaden it back up to checking for _all_ pull requests, though. (Or I guess we also need to check for PRs going into release branches, right? Any other targets we might need?)
- I'm not uploading `_trial_temp/*/*.log` from Trial tests -- do we need these, beyond what's already in stdout?
- I'm not uploading `[ "logs/**/*.log", "logs/**/*.log.*", "logs/results.tap" ]` from SyTest  -- do we need these, beyond what's already in stdout?

babolivier:
> I'm not excluding dinsic branches from tox and mypy tests... if we still want that we can add it back in after merging the broader PR, but I wanted to keep the logic as clear as possible initially.

This shouldn't be an issue since this repo's `dinsic` branch should be considered dead now given it's been replaced by the fork over at https://github.com/matrix-org/synapse-dinsic

clokep:
> * I only check for a newsfile for PRs into develop, which I _think_ is semantically correct. Happy to broaden it back up to checking for _all_ pull requests, though. (Or I guess we also need to check for PRs going into release branches, right? Any other targets we might need?)

I suspect checking for the release branches also makes sense.

> * I'm not uploading `_trial_temp/*/*.log` from Trial tests -- do we need these, beyond what's already in stdout?

These are the server logs of why a test fails. I usually run the test locally, but sometimes they're useful if something only fails in CI (e.g. a Python or platform difference).

> * I'm not uploading `[ "logs/**/*.log", "logs/**/*.log.*", "logs/results.tap" ]` from SyTest  -- do we need these, beyond what's already in stdout?

What's on standard out for sytest isn't very useful. You really need the results.tap file. If something is failing, the log files are useful to see what the homeserver was doing.

.github/workflows/tests.yml
```diff
@@ -118,7 +118,15 @@ jobs:
         SYNAPSE_POSTGRES_HOST: localhost
         SYNAPSE_POSTGRES_USER: postgres
         SYNAPSE_POSTGRES_PASSWORD: postgres
-
+    - name: Dump logs #on failure
+      # Note: Dumps to workflow logs instead of using actions/upload-artifact
+      #       This keeps logs colocated with failing jobs
+      #if: ${{ failure() }}
```
ShadowJonathan, line 124: commented out?

.github/workflows/tests.yml
```diff
@@ -176,23 +184,23 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - sytest-tag: py35
+          - sytest-tag: stretch
```
ShadowJonathan, line 187: This is mainly a comment on sytest itself, but with `stretch`/`buster`/etc. tags its hard to parse what kind of version is being tested exactly, could you maybe add comments which map these?

```
# stretch: python 3.6
# buster: python 3.7
# testing: python 3.8
```
this, at the very least, keeps some maintainability on these tags

.github/workflows/tests.yml
callahad: Testing to make sure it works as expected -- easier to do that than induce a failure :)

.github/workflows/tests.yml
callahad: See https://github.com/matrix-org/sytest/issues/1003 for context

callahad: (I've backed this out for the moment, as we haven't actually rebuilt and retagged images yet)

richvdh:
this looks like it could do with a merge from develop to fix buildkite CI, incidentally.

(on which note: isn't the fact it is failing in buildkite *but not GHA* a red flag?)

callahad:
> this looks like it could do with a merge from develop to fix buildkite CI, incidentally.
> 
> (on which note: isn't the fact it is failing in buildkite _but not GHA_ a red flag?)

That does strike me as a red flag, but with Buildkite not GitHub Actions: This pull request targets `develop` already. If manually merging develop into this branch would cause the Buildkite tests to pass, then it indicates that Buildkite _isn't_ testing the result of the proposed merge, right?

ShadowJonathan:
@richvdh GHA, with merges, acts as if the merge has taken place *before* running the CI pipeline, this way you get an accurate representation of how the changes can affect develop.

(This is also why `bors` was mentioned by @callahad, as that tests continually while the target branch is updated, so a CI result from a week ago isn't "outdated" against the target branch)

richvdh:
> That does strike me as a red flag, but with Buildkite not GitHub Actions: This pull request targets `develop` already. If manually merging develop into this branch would cause the Buildkite tests to pass, then it indicates that Buildkite _isn't_ testing the result of the proposed merge, right?

Ah fair enough. The BK pipeline doesn't do the `merge_base_branch` for the lint steps.

richvdh:
generally this looks sensible to me.

callahad:
I may be off my rocker re: branches / merges / etc. GHA in my fork is now failing along the same lines as BuildKite. Will look more closely tonight.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,322 @@
...
+      - uses: actions/setup-python@v2
+      - run: pip install tox
+      - name: Patch Buildkite-specific test script
+        run: |
+          sed -i -e 's/\$BUILDKITE_PULL_REQUEST/${{ github.event.number }}/' \
+            scripts-dev/check-newsfragment
+      - run: scripts-dev/check-newsfragment
+
+  lint-sdist:
+    if: ${{ contains(github.ref, 'release-') }}
```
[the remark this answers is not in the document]
callahad: Figured there's no need to build wheels and sdists on `develop`... but it only takes ~15 seconds, so we could run it unconditionally. Thoughts?

.github/workflows/tests.yml
```diff
@@ -0,0 +1,322 @@
...
+
+  lint-crlf:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Check line endings
+        run: scripts-dev/check_line_terminators.sh
+
+  lint-newsfile:
+    if: ${{ github.base_ref == 'develop'  || contains(github.ref, 'release-') }}
```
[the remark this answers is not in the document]
callahad: Ah, shoot, good catch. The latter condition should be:

```diff
- contains(github.ref, 'release-')
+ contains(github.base_ref, 'release-')

So that we also look for newsfiles any time a pull request is opened against a release branch or develop.
```

.github/workflows/tests.yml
```diff
@@ -0,0 +1,322 @@
...
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Prepare test blacklist
+        run: cat sytest-blacklist .buildkite/worker-blacklist > synapse-blacklist-with-workers
+      - name: Run SyTest
+        run: /bootstrap.sh synapse
+        working-directory: /src
+      - name: Dump logs
+        # Note: Dumps to workflow logs instead of using actions/upload-artifact
```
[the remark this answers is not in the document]
callahad: Mostly just experimenting. In GHA, artifacts are stored and namespaced at the level of a pipeline run, not per step, so to get at the uploaded logs you need to jump out of the current viewer and up to the overall run summary. Which is kind of annoying.

For small enough logs, dumping them to stdout is nice enough and keeps everything in one place:

Screen Shot 2021-04-06 at 16 30 57

...but it turns out things get really slow when you try to do that for the tens of thousands of lines dumped by a large SyTest run, so I'm going back to using upload-artifact for that.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,322 @@
...
+
+  sytest:
+    if: ${{ !failure() }}
+    needs: linting-done
+    runs-on: ubuntu-latest
+    container:
+      image: matrixdotorg/sytest-synapse:${{ matrix.sytest-tag }}
+      volumes:
+        - ${{ github.workspace }}:/src
+        - ./logs:/logs
```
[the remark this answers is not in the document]
callahad: This was because I did a thinko on the `find` invocation which was only dumping the `results.tap` and ignoring the other logs.

But before I realized that, I noticed the Buildkite pipeline was mounting `./logs:/logs` and wondered if maybe that volume mount needed to be in place for some reason (it doesn't).

.github/workflows/tests.yml
```diff
@@ -0,0 +1,322 @@
...
+        with:
+          repository: "matrix-org/complement"
+          path: complement
+
+      # Build initial Synapse image
+      - run: docker build -t matrixdotorg/synapse:latest -f docker/Dockerfile .
+        working-directory: synapse
+
+      # Build another Synapse image based on the latest release, but suitable
+      # for testing federation with the image built in the prior step.
```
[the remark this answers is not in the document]
callahad: Line 311 builds `matrixdotorg/synapse:latest` from `docker/Dockerfile`; that's the "prior step"

Line 315 builds `complement-synapse` from `Synapse.Dockerfile`, which itself starts with `FROM matrixdotorg/synapse:latest`

So the `complement-synapse` image uses the `matrixdotorg/synapse:latest` image that we just built.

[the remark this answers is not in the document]
callahad: Suggestions for rewording the comment are very welcome -- I found it a bit confusing and I think the comment got a bit muddled as a result.

.github/workflows/tests.yml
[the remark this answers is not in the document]
richvdh: I'm more interested in building it for PRs, to avoid breaking them.

.github/workflows/tests.yml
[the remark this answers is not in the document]
richvdh: I thought the one I wrote for BK was reasonably clear:

```
      # Build a second docker image on top of the above image. This one sets up Synapse with a generated config file,
      # signing and SSL keys so Synapse can run and federate.
```

(It's *not* based on the latest release (of Synapse).)

callahad:
Alright, I think review concerns have been addressed.

Open question: given how GitHub Actions treats artifacts (one big namespace per pipeline run), should we _only_ upload SyTest logs when it fails? It'd make it easier to pick out the ones you want from this list:

Screen Shot 2021-04-06 at 17 45 26

Alternatively: I could name the archives to include the result status ("SyTest Logs PASS ..." or "SyTest Logs FAIL ...").

Unfortunately, I don't think there's a good way to make these viewable directly in the browser: they're just too big to stream into GHA's normal logging, and uploaded artifacts are implicitly wrapped in a zip container.

ShadowJonathan:
One vague suggestion i might have is to take those logs files as input for another pipeline step (if they failed) and extract some easily-readable information from there in console-readable format, or a shorter file that gives an analysis of what went wrong, something similar to what we're currently doing at the "top" of buildkite's build pages.

.github/workflows/tests.yml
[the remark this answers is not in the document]
callahad: Done; changing the workflow to build sdists unconditionally.

.github/workflows/tests.yml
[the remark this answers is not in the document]
callahad: Comment updated

callahad:
> One vague suggestion i might have is to take those logs files as input for another pipeline step (if they failed) and extract some easily-readable information from there in console-readable format, or a shorter file that gives an analysis of what went wrong, something similar to what we're currently doing at the "top" of buildkite's build pages.

So that's handled by the [format_tap.pl](https://github.com/matrix-org/sytest/blob/7f4ce13e7dfb790d943b90de295b13b947bce0f9/scripts/format_tap.pl) script in SyTest which gets run on `results.tap` upon failure. The result is passed to our [annotate-buildkite-plugin](https://github.com/matrix-org/annotate-buildkite-plugin)... we could probably do similar things easily enough in GHA, but probably best to not block on it.

callahad:
Reviewers: I think this is good to go

callahad:
`trial-olddeps` is failing pending updates to our test_old_deps.sh and tox.ini in #9766

Otherwise this should be good to go

callahad:
Yep, everything else is green

richvdh (approved):
lgtm, assuming that CI starts passing once we get our other ducks (#9769?) in a row.

PR #9664, opened 2021-03-22
Fixed undefined variable error in catchup

erikjohnston:
Broke in #9640

### DIFF
--- a/synapse/federation/sender/per_destination_queue.py
+++ b/synapse/federation/sender/per_destination_queue.py
@@ -481,7 +481,7 @@ class PerDestinationQueue:
                     if new_pdus:
                         room_catchup_pdus = new_pdus
                     else:
-                        room_catchup_pdus[pdu]
+                        room_catchup_pdus = [pdu]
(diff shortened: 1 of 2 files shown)


synapse/federation/sender/per_destination_queue.py
```diff
@@ -480,6 +480,8 @@ async def _catch_up_transmission_loop(self) -> None:
                     # the other sending servers are up).
                     if new_pdus:
                         room_catchup_pdus = new_pdus
+                    else:
+                        room_catchup_pdus[pdu]
```
clokep, line 484: Maybe?

```suggestion
                    else:
                        room_catchup_pdus = [pdu]
```

erikjohnston: :facepalm:

clokep (approved):
I did wonder if we should default `room_catchup_pdus` to this instead (up above when we check if stuff is in the extremities already). But this seems fine!

PR #9667, opened 2021-03-23
platform specific prerequisites in source install

fuzzmz:
Make it clearer in the source install step that the platform specific
prerequisites must be installed first.

### Pull Request Checklist

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -6,7 +6,7 @@ There are 3 steps to follow under **Installation Instructions**.
   - [Choosing your server name](#choosing-your-server-name)
   - [Installing Synapse](#installing-synapse)
     - [Installing from source](#installing-from-source)
-      - [Platform-Specific Instructions](#platform-specific-instructions)
+      - [Platform-specific prerequisites](#platform-specific-prerequisites)
         - [Debian/Ubuntu/Raspbian](#debianubunturaspbian)
         - [ArchLinux](#archlinux)
         - [CentOS/Fedora](#centosfedora)
@@ -60,17 +60,14 @@ that your email address is probably `user@example.com` rather than
 
 (Prebuilt packages are available for some platforms - see [Prebuilt packages](#prebuilt-packages).)
 
+When installing from source please make sure that the [Platform-specific prerequisites](#platform-specific-prerequisites) are already installed.
+
 System requirements:
 
 - POSIX-compliant system (tested on Linux & OS X)
 - Python 3.5.2 or later, up to Python 3.9.
 - At least 1GB of free RAM if you want to join large public rooms like #matrix:matrix.org
 
-Synapse is written in Python but some of the libraries it uses are written in
-C. So before we can install Synapse itself we need a working C compiler and the
-header files for Python C extensions. See [Platform-Specific
-Instructions](#platform-specific-instructions) for information on installing
-these on various platforms.
 
 To install the Synapse homeserver run:
 
@@ -128,7 +125,11 @@ source env/bin/activate
 synctl start
 ```
 
-#### Platform-Specific Instructions
+#### Platform-Specific prerequisites
+
+Synapse is written in Python but some of the libraries it uses are written in
+C. So before we can install Synapse itself we need a working C compiler and the
+header files for Python C extensions.
 
 ##### Debian/Ubuntu/Raspbian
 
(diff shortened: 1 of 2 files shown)


ShadowJonathan:
~~"Installation Prerequisites" would be more correct~~

fuzzmz:
Not sure if we should just mention the prereqs, or move them entirely before the source install steps.

INSTALL.md
```diff
@@ -60,6 +60,8 @@ that your email address is probably `user@example.com` rather than
 
 (Prebuilt packages are available for some platforms - see [Prebuilt packages](#prebuilt-packages).)
 
+When installing from source please make sure that the [Platform-Specific prerequisites](#platform-specific-instructions) are already installed.
```
anoadragon453, line 63:
```suggestion
When installing from source please make sure that the [Platform-Specific prerequisites](#platform-specific-prerequisites) are already installed.
```

INSTALL.md
```diff
@@ -128,7 +130,7 @@ source env/bin/activate
 synctl start
 ```
 
-#### Platform-Specific Instructions
+#### Platform-Specific prerequisites
```
anoadragon453, line 133: Please either capitalise "prerequisites" or make the first letter of "specific" lowercase so that the title is consistent.

changelog.d/9667.doc
```diff
@@ -0,0 +1 @@
+Update documentation to mention source install prerequisites before the source install steps.
```
anoadragon453, line 1:
```suggestion
Update source install documentation to mention platform prerequisites before the source install steps.
```

INSTALL.md
```diff
@@ -6,7 +6,7 @@ There are 3 steps to follow under **Installation Instructions**.
   - [Choosing your server name](#choosing-your-server-name)
   - [Installing Synapse](#installing-synapse)
     - [Installing from source](#installing-from-source)
-      - [Platform-Specific Instructions](#platform-specific-instructions)
+      - [Platform-Specific prerequisites](#platform-specific-instructions)
```
anoadragon453, line 9:
```suggestion
      - [Platform-Specific prerequisites](#platform-specific-prerequisites)
```

anoadragon453 (changes requested):
Even I recall forgetting to install dependencies when doing a source install on a random device, so this note is definitely appreciated!

Just needs a few fixes, but it's all small stuff.

I considered suggesting just moving the "Installing from source" section below the "Platform-specific prerequisites" bit, but the former is so large that I think the small note makes more sense.

fuzzmz:
@anoadragon453, thanks for the feedback, I've updated the files based on it.

INSTALL.md
```diff
@@ -68,8 +70,8 @@ System requirements:
 
 Synapse is written in Python but some of the libraries it uses are written in
 C. So before we can install Synapse itself we need a working C compiler and the
-header files for Python C extensions. See [Platform-Specific
-Instructions](#platform-specific-instructions) for information on installing
+header files for Python C extensions. See [Platform-specific
+prerequisites](#platform-specific-prerequisites) for information on installing
```
anoadragon453, line 74: Sorry, now that I look at the [rendered version](https://github.com/matrix-org/synapse/blob/4fcfacac4a1e3a31bfd69a2abbcd1ecc8e97567b/INSTALL.md#installing-from-source), we now have two links to the prerequisites next to each other. Even so, I still missed the original noticec when installing from source recently.

I think the line you've written is in a more visible spot. I would keep that, and move the following explanation:

> Synapse is written in Python but some of the libraries it uses are written in C. So before we can install Synapse itself we need a working C compiler and the header files for Python C extensions. ~~See Platform-specific prerequisites for information on installing these on various platforms.~~ (we can take this bit out)

into the top of the "Platform-specific prerequisites" section instead.

anoadragon453 (changes requested):

INSTALL.md
fuzzmz: Fixed in the latest commit.

INSTALL.md
```diff
@@ -128,7 +125,11 @@ source env/bin/activate
 synctl start
 ```
 
-#### Platform-Specific Instructions
+#### Platform-Specific prerequisites
```
anoadragon453, line 128:
```suggestion
#### Platform-specific prerequisites
```

anoadragon453 (approved):
Looks good to me, thank you very much!

(CI fail unrelated to this change.)

Issue #9409, opened 2021-02-15
we shouldn't ratelimit joins too aggressively


ara4n:
it's not unreasonable for a user to want to join a bunch of rooms - especially with rooms as spaces etc on the horizon

anoadragon453:
Related: https://github.com/matrix-org/synapse/issues/8735

ara4n:
Coincidentally:

<img width="924" alt="Screenshot 2021-02-16 at 11 55 57" src="https://user-images.githubusercontent.com/1294269/108059833-fa5b1d00-704d-11eb-8d86-1feb271a7bf2.png">

PR #9674, opened 2021-03-23
Increase default join burst ratelimiting

erikjohnston:
It's legitimate behaviour to try and join a bunch of rooms at once.

c.f. https://github.com/matrix-org/synapse/issues/9409

### DIFF
--- a/changelog.d/9674.misc
+++ b/changelog.d/9674.misc
@@ -1,0 +1,1 @@
+Increase default join ratelimiting burst rate.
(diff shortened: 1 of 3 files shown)


clokep (approved):
Seems fine. 👍

MacLemon:
Will be interesting if that rate limit may impact joining Spaces in the future. Since that will make discovery of “ALL the rooms” easier, so joins may become more bursty.

PR #9676, opened 2021-03-23
Add type hints to misc. files.

clokep:
This is an old branch I had laying around that adds some type hints to some events files and some of the files in the root `synapse` package. I don't think there's anything too crazy in them.


synapse/visibility.py
```diff
@@ -252,48 +252,46 @@ def allowed(event):
...
+    # EventBase.
     filtered_events = map(allowed, events)
 
-    # remove the None entries
-    filtered_events = filter(operator.truth, filtered_events)
-
-    # we turn it into a list before returning it.
-    return list(filtered_events)
+    # Turn it into a list and remove None entries before returning.
+    return [ev for ev in filtered_events if ev]
```
clokep, line 260: I just found this easier to read than using filter. 🤷

erikjohnston (approved):

PR #9681, opened 2021-03-24
Add type hints to for the federation sender.

clokep:
This adds an abstract-class for the `FederationSender` and `FederationRemoteSendQueue` to ensure they implement the same interface, we then return this interface from the `HomeServer` object.

Split out of #9374

### DIFF
--- a/synapse/federation/sender/__init__.py
+++ b/synapse/federation/sender/__init__.py
@@ -74,6 +74,7 @@ class AbstractFederationSender(metaclass=abc.ABCMeta):
         """This gets called when we have some new events we might want to
         send out to other servers.
         """
+        raise NotImplementedError()
 
     @abc.abstractmethod
     async def send_read_receipt(self, receipt: ReadReceipt) -> None:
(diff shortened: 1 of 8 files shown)


synapse/federation/sender/__init__.py
```diff
@@ -65,8 +68,85 @@
...
+        # to a worker.
+        return 0
+
+    @abc.abstractmethod
+    async def get_replication_rows(
+        self, instance_name: str, from_token: int, to_token: int, target_row_count: int
+    ) -> Tuple[List[Tuple[int, Tuple]], int, bool]:
+        # Dummy implementation for case where federation sender isn't offloaded
+        # to a worker.
+        return [], 0, False
```
erikjohnston: FWIW I tend to do `raise NotImplementedError` so that I don't have to make up return values

clokep: Looks like I had changed it to `pass` in a future commit, but raising `NotImplementedError` makes more sense. 👍

PR #9685, opened 2021-03-25
Rewrite complement.sh

anoadragon453:
This PR rewrites the original complement.sh script with a number of improvements:

* We now use a local checkout of Complement (configurable with `COMPLEMENT_DIR`), instead of downloading the master branch.
* You can now specify a regex of test names to run, or just run all tests.
* We now use the Synapse test blacklist tag (so all tests will pass).

### DIFF
--- a/scripts-dev/complement.sh
+++ b/scripts-dev/complement.sh
@@ -1,4 +1,4 @@
-#! /bin/bash -e
+#!/usr/bin/env bash
 # This script is designed for developers who want to test their code
 # against Complement.
 #
@@ -15,6 +15,10 @@
 #
 # ./complement.sh "TestOutboundFederation(Profile|Send)"
 #
+
+# Exit if a line returns a non-zero exit code
+set -e
+
 COMPLEMENT_DIR="${COMPLEMENT_DIR:-$(dirname $0)/../../complement}"
 
 cd "$(dirname $0)/.."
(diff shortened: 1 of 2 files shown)


scripts-dev/complement.sh
```diff
@@ -1,22 +1,40 @@
...
+# argument to the script. Complement will then only run those tests. If
+# no regex is supplied, all tests are run. Ex.
+#
+# ./complement.sh "TestOutboundFederation(Profile|Send)"
+#
+
+# Exit if a line returns a non-zero exit code
+set -e
+
+COMPLEMENT_DIR="${COMPLEMENT_DIR:-$(dirname $0)/../../complement}"
```
erikjohnston, line 22: I'm not totally comfortable reaching into random parts of the FS in such a magical way. I'd prefer if we just didn't have a default instead and fall back to the old behaviour?

anoadragon453: I was going for mirroring the behaviour of Sytest which assumes a `../synapse` directory. But I suppose it would be:

1. Handy to have the functionality of automatically downloading Complement
2. Best to avoid accidentally running code in `../complement`, in case it doesn't contain Complement

erikjohnston: Ah, didn't realise that sytest did the same. Still, its a bit toooo magical for my liking if we have a better way of doing it.

anoadragon453: We now download by default!

erikjohnston (approved):

PR #9686, opened 2021-03-25
Improve tracing for to device messages

erikjohnston:
This is to try and help track down to device messages that are getting delayed or lost.

### DIFF
--- a/synapse/handlers/devicemessage.py
+++ b/synapse/handlers/devicemessage.py
@@ -182,6 +182,9 @@ class DeviceMessageHandler:
     ) -> None:
         sender_user_id = requester.user.to_string()
 
+        message_id = random_string(16)
+        set_tag("message_id", message_id)
+
         set_tag("number_of_messages", len(messages))
         set_tag("sender", sender_user_id)
         local_messages = {}
@@ -204,6 +207,7 @@ class DeviceMessageHandler:
                         "content": message_content,
                         "type": message_type,
                         "sender": sender_user_id,
+                        "message_id": message_id,
                     }
                     for device_id, message_content in by_device.items()
                 }
@@ -212,8 +216,6 @@ class DeviceMessageHandler:
             else:
                 destination = get_domain_from_id(user_id)
                 remote_messages.setdefault(destination, {})[user_id] = by_device
-
-        message_id = random_string(16)
 
         context = get_active_span_text_map()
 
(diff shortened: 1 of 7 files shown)


synapse/handlers/devicemessage.py
```diff
@@ -221,17 +227,15 @@ async def send_device_message(
...
+            log_kv({"destination": destination})
+            remote_edu_contents[destination] = {
+                "messages": messages,
+                "sender": sender_user_id,
+                "type": message_type,
+                "message_id": message_id,
+                "org.matrix.opentracing_context": json_encoder.encode(context),
+            }
 
-        log_kv({"local_messages": local_messages})
```
anoadragon453, line 232: Looks like the new `log_kv({user_id, device_id})` above is meant to be the replacement for this `log_kv` and the one below. Is that correct?

synapse/logging/opentracing.py
```diff
@@ -259,6 +259,14 @@ def report_span(self, span):
 logger = logging.getLogger(__name__)
 
 
+class SynapseTags:
+    # The message ID of any to_device message processed
+    TO_DEVICE_MESSAGE_ID = "to_device.message_id"
+
+    # Whether the sync response has new data to be returned to the client.
+    SYNC_RESULT = "sync.new_data"
```
anoadragon453, line 267: Is the purpose of this to check whether a `to_device` message came down `/sync` for a user? If so, then won't events other than `to_device` messages lower the signal-to-noise ratio?

synapse/handlers/typing.py
```diff
@@ -86,6 +89,7 @@ def _reset(self) -> None:
         self._member_last_federation_poke = {}
         self.wheel_timer = WheelTimer(bucket_size=5000)
 
+    @wrap_as_background_process("typing.run_as_background_process")
```
anoadragon453, line 92: Should this be `"typing._handle_timeouts"` or similar instead?

anoadragon453 (commented):
Overall looks good, and hooray for more tracing :tada: 

I just had a couple questions below.

synapse/logging/opentracing.py
erikjohnston: Ish, I was more interested in signalling whether we'd found anything or not. I might go in and add per-section flags later, but I'm a bit wary of adding toooo much at once

synapse/handlers/typing.py
erikjohnston: ... yes

synapse/handlers/devicemessage.py
erikjohnston: Yup. This log was a bit useless as it gave the entire message, which then got truncated by jaeger, so you couldn't really see what was going on

anoadragon453 (approved):
lgtm

PR #9689, opened 2021-03-25
Use interpreter from $PATH instead of absolute paths in various scripts using /usr/bin/env

sandhose:
On NixOS, `bash` isn't under `/bin/bash` but rather in some directory in `$PATH`. Locally, I've been patching those scripts to make them work.
`/usr/bin/env` seems to be the only [portable way](https://unix.stackexchange.com/questions/29608/why-is-it-better-to-use-usr-bin-env-name-instead-of-path-to-name-as-my) to use binaries from the PATH as interpreters.

### Pull Request Checklist

### DIFF
--- a/changelog.d/9689.misc
+++ b/changelog.d/9689.misc
@@ -1,0 +1,1 @@
+Use interpreter from $PATH instead of absolute paths in various scripts using /usr/bin/env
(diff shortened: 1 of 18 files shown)


changelog.d/9689.misc
```diff
@@ -0,0 +1 @@
+Use interpreter from $PATH instead of absolute paths in various scripts using /usr/bin/env.
```
anoadragon453, line 1:
```suggestion
Use interpreter from `$PATH` via `/usr/bin/env` instead of absolute paths in various scripts.
```

anoadragon453 (approved):
lgtm following the discussion in #synapse-dev. Thank you!

PR #9692, opened 2021-03-25
Make it possible to use dmypy

erikjohnston:
Running `dmypy run` will do a `mypy` check while spinning up a daemon that makes rerunning `dmypy run` a lot faster.

`dmypy` doesn't support `follow_imports = silent` and has `local_partial_types` enabled, so this PR enables those options and fixes the issues that were newly raised. Note that `local_partial_types` will be enabled by default in upcoming mypy releases.

### DIFF
--- a/changelog.d/9692.misc
+++ b/changelog.d/9692.misc
@@ -1,0 +1,1 @@
+Make it possible to use `dmypy`.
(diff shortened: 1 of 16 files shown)


anoadragon453 (approved):
Just tried it out and it's wicked fast to run subsequent checks! I'd love to place this into `lint.sh`.

PR #7332, opened 2020-04-23
Add information about client delegation

bit:
Delegation also requires .well-known/matrix/client for clients to find the homeserver

PR #9693, opened 2021-03-25
In delegation guide, document public_baseurl

adamvy:
Clients using matrix-react-sdk use the `well_known.homeserver` value from the login
response, which default to server_name, when trying to communicate to the homeserver.

https://matrix.org/docs/spec/client_server/latest#post-matrix-client-r0-login

https://github.com/matrix-org/matrix-react-sdk/blob/develop/src/Login.ts#L222

### Pull Request Checklist

### DIFF
--- a/changelog.d/9693.doc
+++ b/changelog.d/9693.doc
@@ -1,0 +1,1 @@
+Update delegation guide to mention setting public_baseurl as required by matrix-react-sdk based clients.
--- a/docs/delegate.md
+++ b/docs/delegate.md
@@ -51,6 +51,18 @@ However, if you really need it, you can find some documentation on how such a
 record should look like and how Synapse will use it in [the Matrix
 specification](https://matrix.org/docs/spec/server_server/latest#resolving-server-names).
 
+## public_baseurl configuration
+
+You will also need to set public_baseurl in homeserver.yaml to the address of your actual
+synapse server (or a reverse proxy that points to synapse).
+
+```yaml
+public_baseurl: https://synapse.example.com/
+```
+
+otherwise certain clients such as element.io will attempt to reach your homeserver via the
+top level via `https://example.com/`.
+
 ## Delegation FAQ
 
 ### When do I need delegation?


richvdh (commented):
I don't think public_baseurl is related to federation delegation, so I don't think this belongs here.

See also https://github.com/matrix-org/synapse/pull/7332, https://github.com/matrix-org/synapse/pull/9342

adamvy:
Hmm.  Well right now delegate.md documents .well-known delegation.  But matrix-react-sdk uses public_baseurl as returned in the login response as its source of truth for delegation.

matrix-react-sdk: https://github.com/matrix-org/matrix-react-sdk/blob/develop/src/Login.ts#L222

as set by synapse here: https://github.com/matrix-org/synapse/blob/9dbfec0b15ef52508b0ddcd767b48a6edbc847fd/synapse/rest/well_known.py#L41

INSTALL.md says that the .well-known and the public_baseurl should match:
https://github.com/matrix-org/synapse/blob/develop/INSTALL.md#client-well-known-uri

So I think they should both be mentioned in the delegation documentation, as its unwise to change one but not the other.  You only need to set public_baseurl if it differs from servername, but that is the common case in delegation.

richvdh:
> Hmm. Well right now delegate.md documents .well-known delegation.

It documents delegation (of federation traffic) via `.well-known/matrix/server`, yes. `.well-known/matrix/client` is not used for delegation.

> But matrix-react-sdk uses public_baseurl as returned in the login response as its source of truth for delegation.
> 
> matrix-react-sdk: https://github.com/matrix-org/matrix-react-sdk/blob/develop/src/Login.ts#L222

No, this is unrelated to delegatation.

> INSTALL.md says that the .well-known and the public_baseurl should match:
> https://github.com/matrix-org/synapse/blob/develop/INSTALL.md#client-well-known-uri

There is more than one .well-known file. It says that the url in `.well-known/matrix/**client**` should match the public_baseurl.

adamvy:
Ahh I see, client well-known lookup is distinct from delegation.  Thanks for your patience and explanation!

Issue #9656, opened 2021-03-19
Missing Apache Conf


synexTG:
**Description:**

The mentioned Apache Sites conf ist not working anymore. I would be happy to see a complete working guide on how configure synapse for Apache. 

https://github.com/matrix-org/synapse/blob/develop/docs/reverse_proxy.md

richvdh:
please could you explain what is wrong with the example config?

synexTG:
If you just paste it you will notice that after the ServerName attribute the ";" is wrong also if you use this you will still land on the homepage. If you want to look deeper into that i will reinstall it on my server and send you all information.

erikjohnston:
Thanks! If you could include everything you know needs to be changed, that'd be really useful

synexTG:
So i installed all with the install.md instructions, i am using the following apache conf:

```
<VirtualHost *:80>
    ServerName mydomain.com
        Redirect permanent / https://mydomain.com

</VirtualHost>

<VirtualHost *:443>
    SSLEngine on
    ServerName mydomain.com
    SSLCertificateFile /etc/letsencrypt/live/mydomain.com/fullchain.pem
    SSLCertificateKeyFile /etc/letsencrypt/live/mydomain.com/privkey.pem
    RequestHeader set "X-Forwarded-Proto" expr=%{REQUEST_SCHEME}
    AllowEncodedSlashes NoDecode
    ProxyPass /_matrix http://127.0.0.1:8008/_matrix nocanon
    ProxyPassReverse /_matrix http://127.0.0.1:8008/_matrix
    ProxyPass /_synapse/client http://127.0.0.1:8008/_synapse/client nocanon
    ProxyPassReverse /_synapse/client http://127.0.0.1:8008/_synapse/client
</VirtualHost>
```

I am still landing on the default apache site (default index.html in the web root folder). The apache server is reloaded successfull, all needed modules are enabled (proxy_http, headers, ssl).

synexTG:
psh

richvdh:
this isn't a great place to come for support. If you need help getting set up, please go to `#synapse:matrix.org`.

synexTG:
Following information: The configurations works but it doesn't show the matrix home site. 
The only task todo, if the home site doesnt matter, is to remove the simocolons (;) in the doc.

PR #9696, opened 2021-03-26
Preserve host in example apache config

ptman:
Fixes redirect loop

### Pull Request Checklist

This is only a short documentation change which I did on github web ui.

### DIFF
--- a/UPGRADE.rst
+++ b/UPGRADE.rst
@@ -98,9 +98,12 @@ will log a warning on each received request.
 
 To avoid the warning, administrators using a reverse proxy should ensure that
 the reverse proxy sets `X-Forwarded-Proto` header to `https` or `http` to
-indicate the protocol used by the client. See the `reverse proxy documentation
-<docs/reverse_proxy.md>`_, where the example configurations have been updated to
-show how to set this header.
+indicate the protocol used by the client.
+
+Synapse also requires the `Host` header to be preserved.
+
+See the `reverse proxy documentation <docs/reverse_proxy.md>`_, where the
+example configurations have been updated to show how to set these headers.
 
 (Users of `Caddy <https://caddyserver.com/>`_ are unaffected, since we believe it
 sets `X-Forwarded-Proto` by default.)
--- a/docs/reverse_proxy.md
+++ b/docs/reverse_proxy.md
@@ -104,10 +104,11 @@ example.com:8448 {
 ```
 <VirtualHost *:443>
     SSLEngine on
-    ServerName matrix.example.com;
+    ServerName matrix.example.com
 
     RequestHeader set "X-Forwarded-Proto" expr=%{REQUEST_SCHEME}
     AllowEncodedSlashes NoDecode
+    ProxyPreserveHost on
     ProxyPass /_matrix http://127.0.0.1:8008/_matrix nocanon
     ProxyPassReverse /_matrix http://127.0.0.1:8008/_matrix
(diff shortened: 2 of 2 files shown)


docs/reverse_proxy.md
```diff
@@ -108,6 +108,7 @@ example.com:8448 {
```
richvdh, line 108: while you're here: there's a spurious `;` on the line above. Can you fix that?

UPGRADE.rst
```diff
@@ -92,13 +92,14 @@ Requirement for X-Forwarded-Proto header
 ----------------------------------------
 
 When using Synapse with a reverse proxy (in particular, when using the
-`x_forwarded` option on an HTTP listener), Synapse now expects to receive an
-`X-Forwarded-Proto` header on incoming HTTP requests. If it is not set, Synapse
-will log a warning on each received request.
+`x_forwarded` option on an HTTP listener), Synapse now expects to receive
+`X-Forwarded-Proto` and `Host` headers on incoming HTTP requests. If it is not
```
richvdh, line 96: > If it is not

"it" is now unclear in this sentence. It is meant to specifically refer to "`X-Forwarded-Proto`". 

I suggest making a separate sentence ("Synapse also requires the `Host` header to be preserved" or so)
```suggestion
`X-Forwarded-Proto` and `Host` headers on incoming HTTP requests. If it is not
```

docs/reverse_proxy.md
ptman: sure

richvdh:
fixes #9656

UPGRADE.rst
ptman: ok

UPGRADE.rst
```diff
@@ -94,11 +94,13 @@ Requirement for X-Forwarded-Proto header
...
 `X-Forwarded-Proto` header on incoming HTTP requests. If it is not set, Synapse
-will log a warning on each received request.
+will log a warning on each received request. Synapse also requires the `Host`
+header to be preserved.
 
 To avoid the warning, administrators using a reverse proxy should ensure that
 the reverse proxy sets `X-Forwarded-Proto` header to `https` or `http` to
-indicate the protocol used by the client. See the `reverse proxy documentation
+indicate the protocol used by the client and `Host` header matching
+`public_baseurl` in the config. See the `reverse proxy documentation
```
richvdh, line 103: it's actually important that the header is *preserved*, rather than that it matches the url in the config: the entire point is that synapse needs to know whether it needs to redirect the client to the correct host.

So I think it's a mistake to mention `public_baseurl` here; and in any case, the host header won't have any effect on the warning.

UPGRADE.rst
```diff
@@ -94,11 +94,13 @@ Requirement for X-Forwarded-Proto header
 When using Synapse with a reverse proxy (in particular, when using the
 `x_forwarded` option on an HTTP listener), Synapse now expects to receive an
 `X-Forwarded-Proto` header on incoming HTTP requests. If it is not set, Synapse
-will log a warning on each received request.
+will log a warning on each received request. Synapse also requires the `Host`
+header to be preserved.
```
richvdh, line 98: I worry that it will be easy to miss this one sentence. It probably merits its own paragraph (maybe before "See the reverse proxy documentation...").

richvdh (changes requested):

richvdh (approved):
lgtm. Thanks very much!

PR #9697, opened 2021-03-26
Explicitly upgrade openssl in docker file and enforce new version of cryptography

erikjohnston:
This is to ensure that we have an `openssl` version with the latest security patches.

(Apologies for the reformatting, but vscode *really* wants to reformat the dockerfile)

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -24,7 +24,6 @@ RUN apt-get update && apt-get install -y \
         libffi-dev \
         libjpeg-dev \
         libpq-dev \
-        libssl \
         libssl-dev \
         libwebp-dev \
         libxml++2.6-dev \
@@ -72,7 +71,6 @@ RUN apt-get update && apt-get install -y \
         libwebp6 \
         xmlsec1 \
         libjemalloc2 \
-        libssl \
         libssl-dev \
         openssl \
         && rm -rf /var/lib/apt/lists/*
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -24,10 +24,12 @@ RUN apt-get update && apt-get install -y \
         libffi-dev \
         libjpeg-dev \
         libpq-dev \
+        libssl \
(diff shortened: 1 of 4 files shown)


docker/Dockerfile
```diff
@@ -20,17 +20,19 @@ FROM docker.io/python:${PYTHON_VERSION}-slim as builder
...
+        build-essential \
+        libffi-dev \
+        libjpeg-dev \
+        libpq-dev \
+        libssl \
+        libssl-dev \
+        libwebp-dev \
+        libxml++2.6-dev \
+        libxslt1-dev \
+        openssl \
```
richvdh, line 32: is `openssl` actually a build dep? I'm not sure we want to start throwing random things in here just in case they get used.

docker/Dockerfile
```diff
@@ -20,17 +20,19 @@ FROM docker.io/python:${PYTHON_VERSION}-slim as builder
...
-    libxml++2.6-dev \
-    libxslt1-dev \
-    rustc \
-    zlib1g-dev \
- && rm -rf /var/lib/apt/lists/*
+        build-essential \
+        libffi-dev \
+        libjpeg-dev \
+        libpq-dev \
+        libssl \
```
richvdh, line 27: possibly the same could be said here.

docker/Dockerfile
erikjohnston: How do we figure this out?

erikjohnston: Given that we don't know if anything depends on `openssl`, even if we don't *think* so, I'm inclined to install it anyway for paranoia reasons since its a security thing

erikjohnston: Looks like openssl is already installed in the base image

callahad: Yep! Already installed, but not necessarily updated since the base images only get updated ~monthly at best. So good for us to do it here.

PR #9698, opened 2021-03-26
Suppress CryptographyDeprecationWarning

richvdh:
This warning is somewhat confusing to users, so let's suppress it


anoadragon453 (approved):
Seems to work well from local testing!

PR #9699, opened 2021-03-26
Fix the suggested pip incantation for cryptography

richvdh:
If you have the wrong version of `cryptography` installed, synapse suggests:

```
To install run:
    pip install --upgrade --force 'cryptography>=3.4.7;python_version>='3.6''
```

However, the use of ' inside '...' doesn't work, so when you run this, you get
an error.

### DIFF
--- a/synapse/python_dependencies.py
+++ b/synapse/python_dependencies.py
@@ -15,6 +15,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import itertools
 import logging
 from typing import List, Set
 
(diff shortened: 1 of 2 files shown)


synapse/python_dependencies.py
```diff
@@ -150,7 +163,7 @@ def message(self):
     @property
     def dependencies(self):
         for i in self.args[0]:
-            yield "'" + i + "'"
+            yield '"' + i + '"'
```
richvdh, line 166: this line here is the actual fix - the rest of it is just to make sure we don't do the same thing again with ".

erikjohnston (approved):

Issue #9448, opened 2021-02-19
Add m.room.create to invite_room_state


richvdh:
for spaces work, we want `m.room.create` to be included in the state that gets included with an invite (ie, the `invite_room_state`). In theory there is a config option for this, but @t3chguy tried setting it and it didn't work.

This work is to investigate and fix the config option not working.

t3chguy:
It worked when inviting over federation but with some light testing and not much digging did not work for local users.

richvdh:
fixed by #9710

PR #9700, opened 2021-03-26
Replace `room_invite_state_types` with `room_prejoin_state`

richvdh:
`room_invite_state_types` was inconvenient as a configuration setting, because
anyone that ever set it would not receive any new types that were added to the
defaults. Here, we deprecate the old setting, and replace it with a couple of
new settings under `room_prejoin_state`.

This is part of fixing #9448.

### DIFF
--- a/synapse/config/api.py
+++ b/synapse/config/api.py
@@ -55,7 +55,7 @@ class ApiConfig(Config):
            # Additional state event types to share with users when they are invited
            # to a room.
            #
-           # By default, this list is empty (so only the default event types are shared)
+           # By default, this list is empty (so only the default event types are shared).
            #
            #additional_event_types:
            #  - org.example.custom.event.type
@@ -66,7 +66,7 @@ class ApiConfig(Config):
     def _get_prejoin_state_types(self, config: JsonDict) -> Iterable[str]:
         """Get the event types to include in the prejoin state
 
-        Parsees the config and returns an iterable of the event types to be included.
+        Parses the config and returns an iterable of the event types to be included.
         """
         room_prejoin_state_config = config.get("room_prejoin_state") or {}
 
@@ -94,7 +94,7 @@ class ApiConfig(Config):
 _ROOM_INVITE_STATE_TYPES_WARNING = """\
 WARNING: The 'room_invite_state_types' configuration setting is now deprecated,
 and replaced with 'room_prejoin_state'. New features may not work correctly
-unless room_invite_state_types is removed. See the sample configuration file for
+unless 'room_invite_state_types' is removed. See the sample configuration file for
 details of 'room_prejoin_state'.
 --------------------------------------------------------------------------------
 """
--- a/synapse/config/api.py
+++ b/synapse/config/api.py
@@ -1,4 +1,4 @@
-# Copyright 2015, 2016 OpenMarket Ltd
+# Copyright 2015-2021 The Matrix.org Foundation C.I.C.
 #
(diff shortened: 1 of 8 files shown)


clokep (approved):
Seems fine if CI passes!

This will conflict / have some ramifications for the knocking implementation in #6739.

synapse/config/api.py
```diff
@@ -12,38 +12,127 @@
...
+           #
+           # Uncomment the following to disable these defaults (so that only the event
+           # types listed in 'additional_event_types' are shared). Defaults to 'false'.
+           #
+           #disable_default_event_types: true
+
+           # Additional state event types to share with users when they are invited
+           # to a room.
+           #
+           # By default, this list is empty (so only the default event types are shared)
```
anoadragon453, line 58:
```suggestion
           # By default, this list is empty (so only the default event types are shared).
```

synapse/config/api.py
```diff
@@ -12,38 +12,127 @@
...
+        if not room_prejoin_state_config.get("disable_default_event_types"):
+            yield from _DEFAULT_PREJOIN_STATE_TYPES
+
+        yield from room_prejoin_state_config.get("additional_event_types", [])
+
+
+_ROOM_INVITE_STATE_TYPES_WARNING = """\
+WARNING: The 'room_invite_state_types' configuration setting is now deprecated,
+and replaced with 'room_prejoin_state'. New features may not work correctly
+unless room_invite_state_types is removed. See the sample configuration file for
```
anoadragon453, line 97:
```suggestion
unless 'room_invite_state_types' is removed. See the sample configuration file for
```

synapse/handlers/message.py
```diff
@@ -385,7 +385,7 @@ def __init__(self, hs: "HomeServer"):
         self._events_shard_config = self.config.worker.events_shard_config
         self._instance_name = hs.get_instance_name()
 
-        self.room_invite_state_types = self.hs.config.room_invite_state_types
+        self.room_invite_state_types = self.hs.config.api.room_prejoin_state
```
anoadragon453, line 388: Is this preferred over `self.hs.config.room_prejoin_state` for performance reasons?

synapse/config/api.py
```diff
@@ -12,38 +12,127 @@
...
+           #additional_event_types:
+           #  - org.example.custom.event.type
+        """ % {
+            "formatted_default_state_types": formatted_default_state_types
+        }
+
+    def _get_prejoin_state_types(self, config: JsonDict) -> Iterable[str]:
+        """Get the event types to include in the prejoin state
+
+        Parsees the config and returns an iterable of the event types to be included.
```
anoadragon453, line 69:
```suggestion
        Parses the config and returns an iterable of the event types to be included.
```

synapse/handlers/message.py
richvdh: in my ideal world, we'd get rid of all the direct accesses to `config.foo` fields, and remove the magic in `RootConfig` that makes it work.

In the meantime, I'm working to reduce instances of it, one by one.

richvdh:
thanks @anoadragon453 for your attention to detail!

anoadragon453 (approved):

rubo77:
which is the first matrix version, where you have to change this to room_prejoin_state?

clokep:
> which is the first matrix version, where you have to change this to room_prejoin_state?

This option is first available in Synapse v1.32.0, but the code should be backwards compatible. If you're seeing a bug please file a new issue!

PR #9701, opened 2021-03-26
Use 'dmypy run' in lint.sh instead of 'mypy'

anoadragon453:
For it's obvious performance benefits. `dmypy` support landed in #9692 (~~although it appear that it doesn't actually work on develop atm...~~ Twisted just needed an upgrade :)

### DIFF
--- a/changelog.d/9701.misc
+++ b/changelog.d/9701.misc
@@ -1,0 +1,1 @@
+Use `dmypy run` in lint script for improved performance in type-checking while developing.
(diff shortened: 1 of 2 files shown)


erikjohnston (approved):
Yay! Slightly annoyingly `dmypy` has different default to `mypy`, but I think we've mostly set them all to be the same. I believe any differences cause `dmypy` to error when `mypy` succeeds, rather than vice versa, so we shouldn't end up in a situation where `lint.sh` passes locally by CI fails. Hopefully.

Issue #9651, opened 2021-03-19
How destination_rooms is currently persisted in correlation to the per-destination catch-up mechanic would likely not properly work as-is


ShadowJonathan:
While working on #9639 and thinking about how certain aspects of it actually work, I have a feeling that entries `destination_rooms` are not properly persisted for what it is supposed to do.

Currently, [`store_destination_rooms_entries`](https://github.com/matrix-org/synapse/blob/fc86c73e2eb08879dbd29c24c170780ba8839668/synapse/storage/databases/main/transactions.py#L307) is called in [`_send_pdu`](https://github.com/matrix-org/synapse/blob/844c9b0aaedc7092ee24ff0100de70fb141ab5fe/synapse/federation/sender/__init__.py#L315), which in effect is called in the [`process_event_queue_for_federation`](https://github.com/matrix-org/synapse/blob/844c9b0aaedc7092ee24ff0100de70fb141ab5fe/synapse/federation/sender/__init__.py#L178) loop on every event that is being batched to be sent.

However, this store happens *before* the events are [distributed](https://github.com/matrix-org/synapse/blob/844c9b0aaedc7092ee24ff0100de70fb141ab5fe/synapse/federation/sender/__init__.py#L347-L348) to the individual [per_destination transmission loops](https://github.com/matrix-org/synapse/blob/026503fa3b90c03996a64be95432e345434b4a82/synapse/federation/sender/per_destination_queue.py#L261), which means it is possible for these loops to encounter that the remote is not available, and so `destination_rooms` will be updated with a stream_id from *unsent events*.

This would cause the [catch-up transmission loop](https://github.com/matrix-org/synapse/blob/026503fa3b90c03996a64be95432e345434b4a82/synapse/federation/sender/per_destination_queue.py#L381) to miss some (or all) events that need to be sent to that destination if the destination is soon retried after the call to `store_destination_rooms_entries`, with events that were "marked" during that call now not being sent to the destination.

My suggestion would be to move the `store_destination_rooms_entries` task (or a simpler version of it) to [`_TransactionQueueManager.__aexit__`](https://github.com/matrix-org/synapse/blob/026503fa3b90c03996a64be95432e345434b4a82/synapse/federation/sender/per_destination_queue.py#L602), where `set_destination_last_successful_stream_ordering` is already called to set the `destinations` table (but not `destination_rooms`), and to do this in a way that it is called *once* per successful sent batch of events (for all rooms).

[...]

PR #9702, opened 2021-03-26
Experimental Federation Speedup

ShadowJonathan:
Bit of a continuation from #9639's speedup intent and #9651's confusion.

This basically speeds up federation by "squeezing" each individual dual database call (to `destinations` and `destination_rooms`), which previously happened per every event, into *one* call for an entire batch (100 max).

### Pull Request Checklist


synapse/federation/sender/__init__.py
```diff
@@ -393,34 +448,13 @@ async def handle_room_events(events: Iterable[EventBase]) -> None:
...
-        logger.debug("Sending to: %s", str(destinations))
+    def _distribute_pdus(self, destination_to_pdus: Dict[str, List[EventBase]]) -> None:
+        for destination, pdus in destination_to_pdus.items():
+            logger.debug("Sending %d to: %s", len(pdus), destination)
 
-        if not destinations:
-            return
-
-        sent_pdus_destination_dist_total.inc(len(destinations))
-        sent_pdus_destination_dist_count.inc()
```
ShadowJonathan, line 409: These became
```py
sent_pdus_destination_dist_total.inc(len(pdus))
```
per every destination.

I'm fairly sure this is incorrect, what could I replace this with?

synapse/federation/sender/__init__.py
```diff
@@ -337,34 +341,85 @@ async def handle_event(event: EventBase) -> None:
...
+                        return room_id, {
+                            event: dests
+                            for event, dests in (
+                                ret
+                                for ret in (
+                                    await handle_event(event) for event in events
+                                )
+                                if ret is not None
+                            )
+                        }
```
ShadowJonathan, line 369: I'm not sure a dict comprehension this large is actually comprehensible (heh), so i added a comment to it, but i'm still not sure if it's not too complex.

ShadowJonathan:
(Note: Some CI tests are failing with `SyntaxError: 'await' expressions in comprehensions are not supported`, however, [they are supported in 3.6+](https://stackoverflow.com/questions/40746213/how-to-use-await-in-a-python-lambda), with `1.31.0` being the last release that supports 3.5 (for which a release candidate is already out), i'm ignoring these tests.)

synapse/federation/sender/__init__.py
```diff
@@ -337,34 +341,85 @@ async def handle_event(event: EventBase) -> None:
...
+                )  # type: List[Tuple[str, Dict[EventBase, Collection[str]]]]
+
+                # Tuples of room_id + destination to their max-seen stream_ordering
+                room_with_dest_stream_ordering = {}  # type: Dict[Tuple[str, str], int]
+
+                # List of events to send to each destination
+                dest_to_events = {}  # type: Dict[str, List[EventBase]]
+
+                # For each roomID + events with destinations pair...
+                for room_id, event_and_dests in per_room_events_and_dests:
```
ShadowJonathan, line 392: For this section, i'm relatively sure there are ways to make it nicer, but peppering this with comments was a more comprehensible alternative.

If someone can see ways to make this nicer, be my guest.

synapse/federation/sender/__init__.py
ShadowJonathan: (Outdated)

synapse/federation/sender/__init__.py
```diff
@@ -337,34 +341,84 @@ async def handle_event(event: EventBase) -> None:
...
+                async def handle_room_events(
+                    room_id: str, events: Iterable[EventBase]
+                ) -> Tuple[str, Dict[EventBase, Collection[str]]]:
                     with Measure(self.clock, "handle_room_events"):
-                        for event in events:
-                            await handle_event(event)
+                        # Handle all events, skip if handle_event returns None
+                        handled_events = (
+                            ret
+                            for ret in [await handle_event(event) for event in events]
```
ShadowJonathan, line 361: if we change the `[]` comprehension into a generator (`()`), python will start throwing errors about async iterators, because to work with an async iterator in a comprehension, you have to prefix `for` with `async for`

This'll be a small memory tradeoff for readability, as this'd create a list of `len(events)`, but the surrounding generator will filter out all `None` values, which will have itself take up no space until it is consumed by the dict comprehension

so `len(events)` array of `Optional[Tuple[Eventbase, Collection[str]]`, then comprehended into `Dict[EventBase, Collection[str]]`

synapse/federation/sender/__init__.py
```diff
@@ -337,34 +341,84 @@ async def handle_event(event: EventBase) -> None:
...
+                            if ret is not None
+                        )
+
+                        # Generate event -> destination pairs for every handled event
+                        return room_id, {
+                            event: dests for event, dests in handled_events
+                        }
```
ShadowJonathan, line 368: some previous comment had a larger dict comprehension, it can be merged into one with the current setup, but i think this is more readable.

(performance-wise it doesn't matter, because above is still creating only a generator in idle until consumed by this dict comprehension)

synapse/federation/sender/__init__.py
```diff
@@ -337,34 +341,84 @@ async def handle_event(event: EventBase) -> None:
...
+                            # ...and add the event to each destination queue.
+                            event_to_dests.setdefault(destination, []).append(event)
+
+                # Bulk-store destination_rooms stream_ids
+                await self.store.bulk_store_destination_rooms_entries(
+                    room_with_dest_stream_ordering
                 )
 
+                # Send corresponding events to each destination queue
+                self._distribute_pdus(event_to_dests)
```
ShadowJonathan, line 420: distribution happens *after* database access, because then a database-access-crash-loop will cause no events to be spammed to other servers, but a crash loop will also cause federation to stall, which is much more noticeable in the logs (as only the errors will be visible)

also, i think it's "safer" to *not* have a race condition where `destination_room` values are set after pdus are actually distributed

anywho, if either of these functions throw, it wont reach `update_federation_out_pos`, and it'll loop around with `finally:` and the next poke from a database sequence

synapse/federation/sender/__init__.py
```diff
@@ -337,34 +341,84 @@ async def handle_event(event: EventBase) -> None:
...
                         ],
                         consumeErrors=True,
                     )
+                )  # type: List[Tuple[str, Dict[EventBase, Collection[str]]]]
+
+                # Tuples of room_id + destination to their max-seen stream_ordering
+                room_with_dest_stream_ordering = {}  # type: Dict[Tuple[str, str], int]
+
+                # List of events to send to each destination
+                dest_to_events = {}  # type: Dict[str, List[EventBase]]
```
ShadowJonathan, line 388: This is actually fairly cheap, the memory for `EventBase` is already allocated, and it being an object, it's simply referenced.

Each list takes `64` bytes up-front*, and then `8` bytes with each pointer, this means that for a worst-case scenario (100 events destined for `#matrix`, with around 2k servers) is `1728000` bytes (`2000 * (64 + (8*100))`), `1.7` Mb, excluding memory use for destination strings (which are immutably referenced) and `dict` internals for mapping.

(sending 1 event to 2000 servers that way is `144000` bytes, 144kb, btw)

(* the `64` bytes is the up-front cost for a python list, python's objects are quite large that way. That number is for python 3.6, and in python 3.7 they're `72` bytes, but in python 3.8 they're `58` bytes. Lists use C pointers, which are `8` bytes each.)

richvdh, line 388: for consistency with `events_by_room`, this should be called `events_by_dest`.

richvdh (commented):
a few initial thoughts here. It seems generally plausible.

synapse/federation/sender/__init__.py
```diff
@@ -337,34 +341,84 @@ async def handle_event(event: EventBase) -> None:
...
+                        # Handle all events, skip if handle_event returns None
+                        handled_events = (
+                            ret
+                            for ret in [await handle_event(event) for event in events]
+                            if ret is not None
+                        )
+
+                        # Generate event -> destination pairs for every handled event
+                        return room_id, {
+                            event: dests for event, dests in handled_events
```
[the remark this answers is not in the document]
ShadowJonathan: good point, it could also just be a tuple/list of tuples, so i can just return `handled_events` instead of making it a dict :+1:

synapse/federation/sender/__init__.py
```diff
@@ -337,34 +341,84 @@ async def handle_event(event: EventBase) -> None:
                     logger.debug("Sending %s to %r", event, destinations)
```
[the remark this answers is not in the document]
ShadowJonathan: Still true, it'll only take a small second to actually propagate it, it also acknowledges that this event passed through all the tests.

Do you want me to remove it, or place it somewhere else?

synapse/federation/sender/__init__.py
```diff
@@ -393,34 +447,13 @@ async def handle_room_events(events: Iterable[EventBase]) -> None:
         finally:
             self._is_processing = False
 
-    async def _send_pdu(self, pdu: EventBase, destinations: Iterable[str]) -> None:
-        # We loop through all destinations to see whether we already have
-        # a transaction in progress. If we do, stick it in the pending_pdus
-        # table and we'll get back to it later.
+    def _distribute_pdus(self, destination_to_pdus: Dict[str, List[EventBase]]) -> None:
+        for destination, pdus in destination_to_pdus.items():
+            logger.debug("Sending %d to: %s", len(pdus), destination)
```
richvdh, line 452:
```suggestion
            logger.debug("Sending %d pdus to %s", len(pdus), destination)
```

synapse/federation/sender/__init__.py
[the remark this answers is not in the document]
richvdh: probably get rid of it.

synapse/federation/sender/__init__.py
```diff
@@ -337,34 +341,84 @@ async def handle_event(event: EventBase) -> None:
...
+                    for event, destinations in event_and_dests.items():
+
+                        # (we got this from the database, it's filled)
+                        assert event.internal_metadata.stream_ordering
+
+                        # ...iterate over those destinations..
+                        for destination in destinations:
+                            if (
+                                room_with_dest_stream_ordering.setdefault(
+                                    (room_id, destination), 0
```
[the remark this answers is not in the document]
ShadowJonathan: It's effectively also a storage variable, we're finding the largest `stream_id`/`stream_ordering` per room per destination, from all events that find that destination valid.

Eventually it's given to `bulk_store_destination_rooms_entries`, which needs *all* this information to condense it into one `destinations` upsert, and one `destination_rooms` upsert

synapse/federation/sender/__init__.py
ShadowJonathan: Then imma guess that the new approach is correct? Then i don't understand what the old counter was doing with an additional `.inc` after having done so for every destination

richvdh: note that there are two separate counters

synapse/federation/sender/__init__.py
```diff
@@ -328,42 +340,57 @@ async def handle_event(event: EventBase) -> None:
...
                         ).observe((now - ts) / 1000)
 
-                async def handle_room_events(events: Iterable[EventBase]) -> None:
+                        return destinations
+                    return None
+
+                async def get_federatable_events_and_destinations(
+                    room_id: str, events: Iterable[EventBase]
+                ) -> Tuple[str, List[Tuple[EventBase, Collection[str]]]]:
                     with Measure(self.clock, "handle_room_events"):
```
richvdh, line 365: this Measure block needs renaming. Or, more likely, getting rid of.

ShadowJonathan: It still *does* mark some timing on how long it takes to get information for room events' destinations, so i'll rename it to `fetch_destinations_for_events`

synapse/federation/sender/__init__.py
```diff
@@ -393,34 +420,59 @@ async def handle_room_events(events: Iterable[EventBase]) -> None:
...
+                        # ...update their stream-ordering...
+                        this_stream_ordering[
+                            destination
+                        ] = event.internal_metadata.stream_ordering
+
+                    # ...and add the event to each destination queue.
+                    events_by_dest.setdefault(destination, []).append(event)
+
+            for dest, stream_id in this_stream_ordering.items():
+                room_with_dest_stream_ordering[(room_id, dest)] = stream_id
```
richvdh, line 465: why not update `room_with_dest_stream_ordering` directly, rather than going via `this_stream_ordering`?

synapse/federation/sender/per_destination_queue.py
```diff
@@ -154,7 +154,7 @@ def pending_edu_count(self) -> int:
             + len(self._pending_edus_keyed)
         )
 
-    def send_pdu(self, pdu: EventBase) -> None:
+    def send_pdu(self, *pdus: EventBase) -> None:
```
richvdh, line 157: this should probably be called `send_pdus`. Also note that the docstring refers to *a* PDU.

synapse/federation/sender/__init__.py
```diff
@@ -393,34 +420,59 @@ async def handle_room_events(events: Iterable[EventBase]) -> None:
...
+        await self.store.bulk_store_destination_rooms_entries(
+            room_with_dest_stream_ordering
         )
 
-        for destination in destinations:
-            self._get_per_destination_queue(destination).send_pdu(pdu)
+        for destination, pdus in events_by_dest.items():
+            logger.debug("Sending %d pdus to %s", len(pdus), destination)
+
+            self._get_per_destination_queue(destination).send_pdu(*pdus)
```
richvdh, line 475: my instinct is that it would be more efficient - and possibly clearer - to define `send_pdu` to take an `Iterable[EventBase]` and pass `pdus` in directly rather than using the `*` syntax. That may or may not be accurate.

synapse/federation/sender/__init__.py
```diff
@@ -393,34 +420,59 @@ async def handle_room_events(events: Iterable[EventBase]) -> None:
...
+                # ...iterate over those destinations..
+                for destination in destinations:
+                    if (
+                        this_stream_ordering.setdefault(destination, 0)
+                        < event.internal_metadata.stream_ordering
+                    ):
+                        # ...update their stream-ordering...
+                        this_stream_ordering[
+                            destination
+                        ] = event.internal_metadata.stream_ordering
```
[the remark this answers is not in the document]
richvdh: also: can't we assume that we receive events in `stream_ordering` order, so we can make this unconditional? I'm pretty sure that other things depend on this.

synapse/federation/sender/__init__.py
```diff
@@ -275,15 +275,27 @@ async def _process_event_queue_loop(self) -> None:
                 if not events and next_token >= self._last_poked_id:
                     break
 
-                async def handle_event(event: EventBase) -> None:
+                async def get_destinations_for_event(
+                    event: EventBase,
+                ) -> Optional[Collection[str]]:
```
richvdh, line 280: you could have it return an empty tuple when there are no destinations. Feels more elegant/consistent to me. YMMV.

synapse/federation/sender/__init__.py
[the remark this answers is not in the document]
ShadowJonathan: I dont want to assume we receive events in `stream_ordering`, because i have not seen that we *do*, regardless, i think this has minimal performance impact for a good fallback "collection", which can assure more redundancy.

I'll apply the suggestion above :+1:

synapse/federation/sender/__init__.py
ShadowJonathan: You suggested changing *from* that approach to this approach earlier [here](https://github.com/matrix-org/synapse/pull/9702#discussion_r605109179), though i can see how we could make this `dict[str, dict[str, int]]` instead, i'll do that and update some storage logic :+1:

synapse/federation/sender/__init__.py
```diff
@@ -328,42 +340,57 @@ async def handle_event(event: EventBase) -> None:
...
-                            for evs in events_by_room.values()
+                            run_in_background(
+                                get_federatable_events_and_destinations, rid, evs
+                            )
+                            for rid, evs in events_by_room.items()
                         ],
                         consumeErrors=True,
                     )
-                )
+                )  # type: List[Tuple[str, List[Tuple[EventBase, Collection[str]]]]]
```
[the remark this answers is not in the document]
ShadowJonathan: I think that'll have a small performance hit, but I think I can do that. :+1:

synapse/federation/sender/__init__.py
ShadowJonathan: It (effectively) doesn't change the function signature, as it's still able to receive single positional items, I'd wanna change the function signature to `send_pdus` and look for other places where it is called, but every old call to it would need it's singular argument wrapped with `(arg,)` or `[arg]`, `send_pdus([arg])`, which i personally find less readable overall than using this positional argument packing.

synapse/federation/sender/__init__.py
ShadowJonathan: I agree, but it would make checking for "do not send this event to any destination" much more cleaner by check of `is not None`, which is more explicit (imo) than `len(ret) > 0`, same with the current `None` returns, because those'd have to reply with a `Collection`, for which i find `[]`, `set()`, and `tuple()` all not very appealing to semantically read.

It *would* make it cleaner to return a `Collection` in this return type, but i think it'd suffer the readability for it overall, for me personally.

synapse/federation/sender/__init__.py
richvdh: > You suggested changing from that approach to this approach earlier here

no, that's not what I suggested there.

synapse/federation/sender/__init__.py
richvdh: > but every old call to it would need it's singular argument wrapped with (arg,)

there aren't any other calls to it, except in tests...

synapse/federation/sender/__init__.py
richvdh: > I agree, but it would make checking for "do not send this event to any destination" much more cleaner by check of `is not None`, which is more explicit (imo) than `len(ret) > 0`, same with the current `None` returns, because those'd have to reply with a `Collection`, for which i find `[]`, `set()`, and `tuple()` all not very appealing to semantically read.

I have no idea how to parse this sentence.

synapse/federation/sender/__init__.py
richvdh: (note that I only see the original version of a comment in my notifications, so I didn't see the "is that alright" you added later)

synapse/federation/sender/__init__.py
richvdh: by the way, remember that empty lists and tuples are falsey, so rather than `if len(ret) > 0:`, you can just write `if ret:`.

ShadowJonathan: > I have no idea how to parse this sentence.

TL;DR: I think it'll look less pythonic because of all the `return []`s and/or the like

synapse/federation/sender/__init__.py
ShadowJonathan: Alright, then i'll change the function signature :+1:

synapse/federation/sender/__init__.py
ShadowJonathan: then what did you suggest?

synapse/federation/sender/__init__.py
richvdh: I am unconvinced that `return ()` is "less pythonic"  than `return None`, and in any case personally I find it at least as readable.

synapse/federation/sender/__init__.py
richvdh: > you could omit `room_id` from `per_room_events_and_dests`

it seems fairly clear to me, but to spell it out:

currently, `per_room_events_and_dests` is a list of `(room_id, [(event, destinations), ...])` tuples. I assert that you don't need to keep `room_id` separately there, and could instead just make it a it a list of lists of `(event, destinations)`.

I think doing so would simplify the implementation of `get_federatable_events_and_destinations` (since it would no longer need a separate `room_id` parameter).

synapse/federation/sender/__init__.py
ShadowJonathan: ...D'oh, i thought `()` != "empty tuple", mainly cuz Rust uses this as their "empty" type, i got confused for a second.

In that case, i think using `return ()` is perfectly fine, i messed it up, i'll change this to that :+1:

synapse/federation/sender/__init__.py
ShadowJonathan: Alright, that makes sense, i'll try to change it to this :+1:

synapse/federation/sender/__init__.py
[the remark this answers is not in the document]
ShadowJonathan: I believe that parallelising per event will cause a bigger performance hit, in this regard, not because it will have to process 100 tasks concurrently suddenly (which i believe python is able to elegantly process), but because all of those events will be requesting `prev_events` and the like, which might be cached (or cachable), and many events concurrently "fighting" for this might mean more fetches than necessary, the (rough) linearisation per room, then, is intended to keep coroutines from righting over cache-misses, and to instead let every `prev_event` request lead into the next ones for subsequent events.

Just a note, i'll resolve this thread, but i just wanted to jot down my thoughts.

ShadowJonathan:
`check-style` is currently borked due to #9734 :/

ShadowJonathan:
Its fixed on develop, merged it into here to ensure that CI runs again

synapse/federation/sender/__init__.py
```diff
@@ -277,7 +271,7 @@ async def _process_event_queue_loop(self) -> None:
 
                 async def get_destinations_for_event(
                     event: EventBase,
-                ) -> Optional[Collection[str]]:
+                ) -> Collection[str]:
                     """Computes the destinations to which this event must be sent.
 
                     This returns None when there are no destinations to send to,
```
richvdh, line 277: no it doesn't. Please make sure that comments match code.

synapse/federation/sender/__init__.py
```diff
@@ -357,40 +351,28 @@ async def get_destinations_for_event(
...
-                    return None
+                    return ()
 
                 async def get_federatable_events_and_destinations(
-                    room_id: str, events: Iterable[EventBase]
-                ) -> Tuple[str, List[Tuple[EventBase, Collection[str]]]]:
-                    with Measure(self.clock, "handle_room_events"):
+                    events: Iterable[EventBase],
+                ) -> List[Tuple[EventBase, Collection[str]]]:
+                    with Measure(self.clock, "fetch_destinations_for_events"):
```
richvdh, line 359:
```suggestion
                    with Measure(self.clock, "get_destinations_for_events"):
```

synapse/federation/sender/__init__.py
```diff
@@ -357,40 +351,28 @@ async def get_destinations_for_event(
...
                         # Get destinations for events, skip if get_destinations_for_event returns None
```
richvdh, line 360: Please make sure that comments match code.

synapse/federation/sender/__init__.py
```diff
@@ -422,47 +404,38 @@ async def get_federatable_events_and_destinations(
 
     async def _distribute_events(
         self,
-        per_room_events_and_dests: List[
-            Tuple[str, List[Tuple[EventBase, Collection[str]]]]
-        ],
+        events_and_dests: List[Tuple[EventBase, Collection[str]]],
```
richvdh, line 407: I think you can relax this constraint:

```suggestion
        events_and_dests: Iterable[Tuple[EventBase, Collection[str]]],
```

synapse/federation/sender/__init__.py
```diff
@@ -422,47 +404,38 @@ async def get_federatable_events_and_destinations(
...
     ) -> None:
+        """Distribute events from the transmission loop.
```
richvdh, line 409: could you add some more words. "Distribute" doesn't really tell us much here.

synapse/federation/sender/__init__.py
```diff
@@ -422,47 +404,38 @@ async def get_federatable_events_and_destinations(
...
+
+        Args:
+            events_and_dests: A list of tuples, which are (event, destinations).
```
richvdh, line 412: what are the destinations?

synapse/federation/sender/per_destination_queue.py
```diff
@@ -155,11 +155,11 @@ def pending_edu_count(self) -> int:
             + len(self._pending_edus_keyed)
         )
 
-    def send_pdu(self, *pdus: EventBase) -> None:
-        """Add a PDU to the queue, and start the transmission loop if necessary
+    def send_pdus(self, pdus: Collection[EventBase]) -> None:
```
richvdh, line 158:
```suggestion
    def send_pdus(self, pdus: Iterable[EventBase]) -> None:
```

synapse/federation/sender/__init__.py
```diff
@@ -328,42 +334,45 @@ async def handle_event(event: EventBase) -> None:
...
+                            for (event, dests) in [
+                                (event, await get_destinations_for_event(event))
+                                for event in events
+                            ]
+                            if dests
+                        ]
+
+                events_and_dests = await get_federatable_events_and_destinations(
+                    events
+                )  # type: List[Tuple[EventBase, Collection[str]]]
```
richvdh, line 372: I'm surprised you need an annotation here. Mypy should figure it out from the declaration of `get_federatable_events_and_destinations`.

TBH I'd be inclined to inline `get_federatable_events_and_destinations` (and remove the Measure block, unless you have a particular reason to want to keep it)

richvdh (commented):
this is looking good, save a few outstanding nits.

Have you tried it in practice? Does it make much difference?

ShadowJonathan:
> Have you tried it in practice? Does it make much difference?

I haven't, but i also need to set up a generic "dev server" somewhere, and i may do so after/during this PR

synapse/federation/sender/__init__.py
ShadowJonathan: Will add some more comments, and add types in that tuple

synapse/federation/sender/__init__.py
ShadowJonathan: I want to keep the measure block for metrics, and that type annotation is a leftover from many refactors, i'll remove that (if it doesnt cause mypy to get confused) :+1:

synapse/federation/sender/__init__.py
```diff
@@ -358,8 +358,10 @@ async def get_destinations_for_event(
                 async def get_federatable_events_and_destinations(
                     events: Iterable[EventBase],
                 ) -> List[Tuple[EventBase, Collection[str]]]:
-                    with Measure(self.clock, "fetch_destinations_for_events"):
-                        # Get destinations for events, skip if get_destinations_for_event returns None
+                    with Measure(self.clock, "get_destinations_for_events"):
+                        # Fetch federation destinations per event,
+                        # skip if get_destinations_for_event returns None,
```
richvdh, line 363: this isn't true, is it? c'mon, it shouldn't be my job to point this out.

richvdh (approved):
this looks good other than the below. I'm going to try running it on sw1v.org.

synapse/federation/sender/__init__.py
ShadowJonathan: :facepalm: 

You're now legally allowed to slap me if i apply review suggestions before i've had coffee, will fix :+1:

richvdh:
:man_shrugging:

anoadragon453:
We reverted this PR due to some performance regressions in sending PDUs over federation. The details are in https://github.com/matrix-org/synapse/issues/9863.

PR #9705, opened 2021-03-27
Experimental GIF Thumbnailing support

### LINKED ISSUE #1278: Thumbnailing animated GIFs (or showing they're an animation) (SYN-251)
(Imported from https://matrix.org/jira/browse/SYN-251)

(Reported by @ara4n)

ShadowJonathan:
Hopefully resolves #1278

Adds in some fixes and refactors to `synapse.rest.media.v1.thumbnailer.Thumbnailer` that makes it able to recognise `image/gif` as an output type and transform images accordingly.

This is experimental for the following reasons;
 - Currently no GIF resources are expected on the Element client, allowing this to be on by default will cause people to be able to "visually spam" others by setting their avatars, room icons, or just sending specific "emoji" images as GIFs.
   - The solution would be a better awareness in the Element repos for animated images, which this PR calls for.
 - Processing GIFs might be heavy on servers, as each individual frame has to be transformed.
 - In it's current state, there can be some thumbnailing codepaths and transformations that might be missed by unit testing, so this hasn't been "really" tested.
 
For these reasons, GIF thumbnailing support is experimental and disabled by default.

### Pull Request Checklist


synapse/config/repository.py
```diff
@@ -70,6 +70,7 @@ def parse_thumbnail_requirements(thumbnail_sizes):
         method = size["method"]
         jpeg_thumbnail = ThumbnailRequirement(width, height, method, "image/jpeg")
         png_thumbnail = ThumbnailRequirement(width, height, method, "image/png")
+        gif_thumbnail = ThumbnailRequirement(width, height, method, "image/gif")
```
ShadowJonathan, line 73: no idea how to go further than this, as this happens in a `.config.` file, and to "enable" this, you'd need to grab something from the "experimental" config section (as custom for experimental stuff), however, i dunno how to do this without messing up imports or something

synapse/rest/media/v1/thumbnailer.py
```diff
@@ -93,27 +93,16 @@ def aspect(self, max_width: int, max_height: int) -> Tuple[int, int]:
         else:
             return (max_height * self.width) // self.height, max_height
 
-    def _resize(self, width: int, height: int) -> Image:
```
ShadowJonathan, line 96: (reordering public methods above private ones)

synapse/rest/media/v1/thumbnailer.py
```diff
@@ -129,24 +118,62 @@ def crop(self, width: int, height: int, output_type: str) -> BytesIO:
         Returns:
             BytesIO: the bytes of the encoded image ready to be written to disk
         """
+
+        return self._process_animated_aware(
+            self.image, output_type, lambda i: self._crop(i, width, height)
+        )
+
+    def _process_animated_aware(
```
ShadowJonathan, line 126: rule of three *should* apply here, but this piece is a bit complex, and the only thing that'd really change between the two functions is changing `_resize` with `_crop`, thus the abstraction

synapse/rest/media/v1/thumbnailer.py
```diff
@@ -129,24 +118,62 @@ def crop(self, width: int, height: int, output_type: str) -> BytesIO:
...
+        self, image: Image, output_type: str, transform: "Callable[[Image], Image]"
+    ) -> BytesIO:
+        if self.FORMATS[output_type] == "GIF" and getattr(image, "n_frames", 1) > 1:
+            first, *rest = (transform(i) for i in ImageSequence.Iterator(image))
+            return self._encode_image(first, output_type, append_images=rest)
+        else:
+            return self._encode_image(transform(image), output_type)
+
+    def _crop(self, image: Image, width: int, height: int) -> Image:
```
ShadowJonathan, line 135: to-private and addition of `image` to decouple it and make it able to be used as a generic image transformer

clokep:
Are you hoping to get feedback for this or still tweaking it? (Are you looking for a review?)

ShadowJonathan:
On above self-reviews, yes, I'm inviting comments, because i need some direction on this before i actually finish this, i wish I could communicate that clearer, but i think putting it on "ready" isn't one of them, as then the PR would be considered done, which it certainly isn't.

synapse/config/repository.py
richvdh: There are other interdependencies between config sections (see for example `ApiConfig` which checks `self.spaces_enabled`, which is part of `ExperimentalConfig`).  As long as `ContentRepositoryConfig` appears after `ExperimentalConfig` in the `HomeServerConfig` definition, this will work. It's not particularly elegant, but :man_shrugging:

richvdh (commented):
I'm hesitant about the concept here, even behind an experimental option, unless it's a thing that the matrix protocol actually wants to support. In the best case it becomes a maintenance burden in Synapse; in the worst case lots of people start using gifs (animated or otherwise) and it becomes a thing the matrix protocol has to support forever.

In the interest of trying to stimulate discussion on this from the spec POV, I've commented at https://github.com/matrix-org/matrix-doc/issues/1938#issuecomment-810968551.

ShadowJonathan:
I agree, I sorta found the way thumbnails are identified/retrieved fairly weird, maybe a comprehensive thumbnail endpoint (f.e. one that supports multiple representations of the same media, animated or not) could definitively solve this.

I'll wait for a resolution/MSC, until then, this PR is probably stalled.

ShadowJonathan:
This PR almost certainly has bitrotted, and it is still blocked on https://github.com/matrix-org/matrix-spec/issues/453, so i'll close this.

PR #9709, opened 2021-03-29
Fix CI by ignore type for None module import

anoadragon453:
Looks like this is the workaround(?) we use in other places where we need to test whether a module exists.

Why this only broke now, I couldn't tell you. Perhaps we configured mypy to only start checking this file recently?

### DIFF
--- a/changelog.d/9709.misc
+++ b/changelog.d/9709.misc
@@ -1,0 +1,1 @@
+Fix type-checking CI on develop.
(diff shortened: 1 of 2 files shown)


richvdh (approved):
lgtm

Bubu:
> Why this only broke now, I couldn't tell you.

Not 100% sure this is the cause but hiredis did a release yesterday.

anoadragon453:
@Bubu Aha, that'd be it then. Thanks!

PR #9710, opened 2021-03-29
Include m.room.create in invite_room_state for Spaces

richvdh:
Based on #9700


erikjohnston (approved):
LGTM, I think commit e54ada4 is the only salient one here. It does look like something odd has gone on when merging back from the base branch, so we probably want to merge in develop after the base branch has been merged to double check everything is OK

PR #9711, opened 2021-03-29
Make RateLimiter class check for ratelimit overrides

### LINKED ISSUE #9663: Do not rate limit invites if an appservice has `rate_limited: False` set
At the moment we are having issues with an EMS script that creates rooms en masse for customers looking to migrate. However calls made to /createRoom will fail due to a recent change to rate limit the number of invites allowed at room creation time. It would be good if the rate_limited: False flag could be respected here, and the limiting be disabled for users attempting to create rooms inside the AS namespace.

erikjohnston:
This should fix a class of bug where we forget to check if e.g. the appservice shouldn't be ratelimited.

We also check the `ratelimit_override` table to check if the user has ratelimiting disabled. That table is really only meant to override the event sender ratelimiting, so we don't use any values from it (as they might not make sense for different rate limits), but we do infer that if ratelimiting is disabled for the user we should disabled all ratelimits.

Fixes #9663

### DIFF
--- a/synapse/api/ratelimiting.py
+++ b/synapse/api/ratelimiting.py
@@ -46,7 +46,7 @@ class Ratelimiter:
             OrderedDict()
         )  # type: OrderedDict[Hashable, Tuple[float, int, float]]
 
-    def can_requester_do_action(
+    async def can_requester_do_action(
         self,
         requester: Requester,
         rate_hz: Optional[float] = None,
@@ -73,17 +73,18 @@ class Ratelimiter:
                 * The reactor timestamp for when the action can be performed next.
                   -1 if rate_hz is less than or equal to zero
         """
-        # Disable rate limiting of users belonging to any AS that is configured
-        # not to be rate limited in its registration file (rate_limited: true|false).
-        if requester.app_service and not requester.app_service.is_rate_limited():
-            return True, -1.0
-
-        return self.can_do_action(
-            requester.user.to_string(), rate_hz, burst_count, update, _time_now_s
+        return await self.can_do_action(
+            requester,
+            requester.user.to_string(),
+            rate_hz,
+            burst_count,
+            update,
+            _time_now_s,
         )
 
-    def can_do_action(
+    async def can_do_action(
         self,
+        requester: Optional[Requester],
         key: Hashable,
         rate_hz: Optional[float] = None,
         burst_count: Optional[int] = None,
@@ -93,6 +94,8 @@ class Ratelimiter:
         """Can the entity (e.g. user or IP address) perform the action?
 
         Args:
+            requester: The requester that is doing the action, if any. Used to check for
+                ratelimit overrides.
             key: The key we should use when rate limiting. Can be a user ID
                 (when sending events), an IP address, etc.
             rate_hz: The long term number of actions that can be performed in a second.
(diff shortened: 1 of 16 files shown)


richvdh:
Related: #9648

synapse/api/ratelimiting.py
```diff
@@ -93,6 +98,8 @@ def can_do_action(
         """Can the entity (e.g. user or IP address) perform the action?
 
         Args:
+            requester: The requester that is doing the action, if any. Used to check for
+                ratelimit overrides.
```
richvdh, line 102: I think it's worth being explicit here that we only check for the presence of a non-zero override.

synapse/api/ratelimiting.py
```diff
@@ -186,6 +212,8 @@ def ratelimit(
         """Checks if an action can be performed. If not, raises a LimitExceededError
 
         Args:
+            requester: The requester that is doing the action, if any. Used to check for
+                ratelimit overrides.
```
richvdh, line 216: ditto.

I think we should also say something about this in the docstring for `can_requester_do_action`.

synapse/handlers/auth.py
```diff
@@ -373,7 +377,9 @@ def get_new_session_data() -> JsonDict:
             )
         except LoginError:
             # Update the ratelimiter to say we failed (`can_do_action` doesn't raise).
-            self._failed_uia_attempts_ratelimiter.can_do_action(requester_user_id)
+            await self._failed_uia_attempts_ratelimiter.can_do_action(
+                requester, requester_user_id
+            )
```
richvdh, line 382: could use `requester_can_do_action` here.

synapse/handlers/auth.py
```diff
@@ -352,7 +354,9 @@ async def validate_user_via_ui_auth(
         requester_user_id = requester.user.to_string()
 
         # Check if we should be ratelimited due to too many previous failed attempts
-        self._failed_uia_attempts_ratelimiter.ratelimit(requester_user_id, update=False)
+        await self._failed_uia_attempts_ratelimiter.ratelimit(
+            requester, requester_user_id, update=False
+        )
```
richvdh, line 359: I wonder if we should make the `key` of `ratelimit` optional, and have it default to `requester.user.to_string()`.

tests/api/test_ratelimiting.py
```diff
@@ -5,38 +5,48 @@
...
 
-class TestRatelimiter(unittest.TestCase):
+class TestRatelimiter(unittest.HomeserverTestCase):
     def test_allowed_via_can_do_action(self):
-        limiter = Ratelimiter(clock=None, rate_hz=0.1, burst_count=1)
-        allowed, time_allowed = limiter.can_do_action(key="test_id", _time_now_s=0)
+        limiter = Ratelimiter(
+            store=self.hs.get_datastore(), clock=None, rate_hz=0.1, burst_count=1
+        )
+        allowed, time_allowed = self.get_success_or_raise(
```
richvdh, line 13: what's the difference between `get_success` and `get_success_or_raise`, btw? The latter looks very much like an inlined version of the former.

richvdh (commented):
looks fine otherwise

tests/api/test_ratelimiting.py
richvdh: hrm, fair enough. maybe that should be controlled via a kwarg, rather than a completely separate impl.

richvdh (approved):
lgtm

PR #9713, opened 2021-03-30
Add release helper script

erikjohnston:
This does everything up to tagging the actual release.

Sample run:

```
$ ./scripts-dev/release.py
Updating git repo...
RC [Y/n]: 
Release type (major, hotfix) [major]: 
Create new version: 1.31.0rc1? [Y/n]: 
Which branch should release be based off of? [develop]: 
Loading template...
Finding news fragments...
Rendering news fragments...
Writing to newsfile...
Staging newsfile...
Removing news fragments...
I want to remove the following files:
... < list of changelog files > ...
Is it okay if I remove those files? [Y/n]: 
Done!
... < displays diff > ...
Edit changelog? [y/N]: y
Push branch to github? [Y/n]: 
```

### DIFF
--- a/scripts-dev/release.py
+++ b/scripts-dev/release.py
@@ -169,7 +169,7 @@ def release():
             default = "master"
 
         branch_name = click.prompt(
-            "Which branch should release be based on?", default=default
+            "Which branch should the release be based on?", default=default
         )
 
         base_branch = find_ref(repo, branch_name)
@@ -195,7 +195,7 @@ def release():
     with open("synapse/__init__.py", "w") as f:
         f.write(red.dumps())
 
-    # Generate changelgs
+    # Generate changelogs
     subprocess.run("python3 -m towncrier", shell=True)
 
     # Generate debian changelogs if its not an RC.
--- a/scripts-dev/release.py
+++ b/scripts-dev/release.py
@@ -14,9 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""An interactive script to walk through the initial stages of creating a
-release, including creating release branch, updating changelog and pushing to
-GitHub.
+"""An interactive script for doing a release. See `run()` below.
 """
 
 import subprocess
@@ -29,25 +27,17 @@ from packaging import version
 from redbaron import RedBaron
 
 
-def find_ref(repo: git.Repo, ref_name: str) -> Optional[git.HEAD]:
-    """Find the branch/ref, looking first locally then in the remote."""
-    if ref_name in repo.refs:
-        return repo.refs[ref_name]
-    elif ref_name in repo.remote().refs:
-        return repo.remote().refs[ref_name]
-    else:
-        return None
-
-
-def update_branch(repo: git.Repo):
-    """Ensure branch is up to date if it has a remote"""
-    if repo.active_branch.tracking_branch():
-        repo.git.merge(repo.active_branch.tracking_branch().name)
-
-
 @click.command()
-def release():
-    """Main release command"""
+def run():
+    """An interactive script to walk through the initial stages of creating a
+    release, including creating release branch, updating changelog and pushing to
+    GitHub.
+
+    Requires the dev dependencies be installed, which can be done via:
+
+        pip install -e .[dev]
+
+    """
 
     # Make sure we're in a git repo.
     try:
@@ -234,5 +224,21 @@ def release():
     )
 
 
+def find_ref(repo: git.Repo, ref_name: str) -> Optional[git.HEAD]:
+    """Find the branch/ref, looking first locally then in the remote."""
+    if ref_name in repo.refs:
+        return repo.refs[ref_name]
+    elif ref_name in repo.remote().refs:
+        return repo.remote().refs[ref_name]
+    else:
+        return None
+
+
+def update_branch(repo: git.Repo):
+    """Ensure branch is up to date if it has a remote"""
+    if repo.active_branch.tracking_branch():
+        repo.git.merge(repo.active_branch.tracking_branch().name)
+
+
 if __name__ == "__main__":
-    release()
+    run()
--- a/scripts-dev/release.py
+++ b/scripts-dev/release.py
@@ -169,7 +169,7 @@ def release():
             default = "master"
 
         branch_name = click.prompt(
-            "Which branch should release be based off of?", default=default
+            "Which branch should release be based on?", default=default
         )
 
         base_branch = find_ref(repo, branch_name)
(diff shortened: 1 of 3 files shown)


scripts-dev/release.py
```diff
@@ -0,0 +1,238 @@
...
+"""
+
+import subprocess
+import sys
+from typing import Optional
+
+import click
+import git
+from packaging import version
+from redbaron import RedBaron
```
richvdh, line 29: the problem with this stuff is that it has to be installed in the virtualenv. We can probably add them to the dev dependencies if they are useful.

scripts-dev/release.py
```diff
@@ -0,0 +1,238 @@
...
+        # release type.
+        if current_version.is_prerelease:
+            default = release_branch_name
+        elif release_type == "major":
+            default = "develop"
+        else:
+            default = "master"
+
+        branch_name = click.prompt(
+            "Which branch should release be based off of?", default=default
```
richvdh, line 172:
```suggestion
            "Which branch should release be based on?", default=default
```

scripts-dev/release.py
```diff
@@ -0,0 +1,238 @@
...
+        )
+
+        base_branch = find_ref(repo, branch_name)
+        if not base_branch:
+            print(f"Could not find base branch {branch_name}!")
+            click.get_current_context().abort()
+
+        # Checkout the base branch and ensure its up to date
```
richvdh, line 180:
```suggestion
        # Check out the base branch and ensure it's up to date
```

richvdh (commented):
looks generally good.

scripts-dev/release.py
erikjohnston: Hmm, you're right. I'm not a huge fan of adding them as a dev dependency, but I don't have a better idea and they all make the script easier and remove boiler plate.

erikjohnston: I've added it under a new `"dev"` section, not sure if that's the right way to do it though.

scripts-dev/release.py
```diff
@@ -0,0 +1,238 @@
...
+        # release type.
+        if current_version.is_prerelease:
+            default = release_branch_name
+        elif release_type == "major":
+            default = "develop"
+        else:
+            default = "master"
+
+        branch_name = click.prompt(
+            "Which branch should release be based on?", default=default
```
anoadragon453, line 172:
```suggestion
            "Which branch should the release be based on?", default=default
```

scripts-dev/release.py
```diff
@@ -0,0 +1,238 @@
...
+    # Switch to the release branch and ensure its up to date.
+    repo.git.checkout(release_branch_name)
+    update_branch(repo)
+
+    # Update the `__version__` variable and write it back to the file.
+    version_node.value = '"' + new_version + '"'
+    with open("synapse/__init__.py", "w") as f:
+        f.write(red.dumps())
+
+    # Generate changelgs
```
anoadragon453, line 198:
```suggestion
    # Generate changelogs
```

scripts-dev/release.py
```diff
@@ -0,0 +1,238 @@
...
+    subprocess.run("python3 -m towncrier", shell=True)
+
+    # Generate debian changelogs if its not an RC.
+    if not rc:
+        subprocess.run(
+            f'dch -M -v {new_version} "New synapse release {new_version}."', shell=True
+        )
+        subprocess.run('dch -M -r -D stable ""', shell=True)
```
anoadragon453, line 206: So `dch` isn't actually available on Arch Linux - thus I need to edit the file manually. It would be handy to add a condition to this that checks whether `dch` is installed and skip+warn the user if not.

I can also just do this myself in a separate PR :slightly_smiling_face:

callahad:
https://aur.archlinux.org/packages/devscripts/ ?

On Thu, Apr 1, 2021, at 14:15, Andrew Morgan wrote:
> 

> ***@***.**** commented on this pull request.

> 
> In scripts-dev/release.py <https://github.com/matrix-org/synapse/pull/9713#discussion_r605643252>:

> > +            # locally.
+            repo.git.checkout(release_branch_name)
+            release_branch = repo.active_branch
+    else:
+        # If a branch doesn't exist we create one. We ask which one branch it
+        # should be based off, defaulting to sensible values depending on the
+        # release type.
+        if current_version.is_prerelease:
+            default = release_branch_name
+        elif release_type == "major":
+            default = "develop"
+        else:
+            default = "master"
+
+        branch_name = click.prompt(
+            "Which branch should release be based on?", default=default
> 
> ⬇️ Suggested change 
> -            "Which branch should release be based on?", default=default
+            "Which branch should the release be based on?", default=default
> 
> 
> In scripts-dev/release.py <https://github.com/matrix-org/synapse/pull/9713#discussion_r605645246>:

> > +        if not base_branch.is_remote():
+            update_branch(repo)
+
+        # Create the new release branch
+        release_branch = repo.create_head(release_branch_name, commit=base_branch)
+
+    # Switch to the release branch and ensure its up to date.
+    repo.git.checkout(release_branch_name)
+    update_branch(repo)
+
+    # Update the `__version__` variable and write it back to the file.
+    version_node.value = '"' + new_version + '"'
+    with open("synapse/__init__.py", "w") as f:
+        f.write(red.dumps())
+
+    # Generate changelgs
> 
> ⬇️ Suggested change 
> -    # Generate changelgs
+    # Generate changelogs
> 
> 
> In scripts-dev/release.py <https://github.com/matrix-org/synapse/pull/9713#discussion_r605647645>:

> > +            f'dch -M -v {new_version} "New synapse release {new_version}."', shell=True
+        )
+        subprocess.run('dch -M -r -D stable ""', shell=True)

scripts-dev/release.py
callahad: https://aur.archlinux.org/packages/devscripts/ ?

anoadragon453: Aha! I knew there was a package somewhere - but a year or so ago it was spewing out errors.

But now it works perfectly for whatever reason! Thanks :tada:

scripts-dev/release.py
```diff
@@ -0,0 +1,238 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""An interactive script to walk through the initial stages of creating a
+release, including creating release branch, updating changelog and pushing to
+GitHub.
+"""
```
richvdh, line 20: I suggest adding something about `pip install -e .[dev]` here.

richvdh (commented):
looks good. One more thing: could you do the `argparse` magic necessary so that `release.py -h` gives a help string? Just as a matter of good practice/making a nice script to use.

erikjohnston:
> looks good. One more thing: could you do the `argparse` magic necessary so that `release.py -h` gives a help string? Just as a matter of good practice/making a nice script to use.

Done. Note that the `click` library derives the help from the docstring of the `@click.command()` function

richvdh (approved):
lgtm!

PR #9717, opened 2021-03-30
Add an experimental room version to support restricted join rules.

### LINKED ISSUE #9712: Experimental join_rule for joining rooms in a space without an invite
For the Spaces project there needs to be a way to allow users to join a room based on membership in a space.

An approach for this is detailed in [MSC3083](https://github.com/matrix-org/matrix-doc/pull/3083).

The approach will be to:
* Create an experimental room version.
* Update the auth rules based on the `join_rule` / experimental room version to allow users to join based on the `allowed` key in the `m.room.join_rules` event.

Note that we want this to be in an experimental room version as it is unclear whether clients will need to make changes to support this.

clokep:
* Creates an experimental room version for a new join rule: `restricted`.
* Implements the `restricted` join rule as the same as `public` for that experimental room version.
* Adds a bunch of unit tests for the `_is_membership_change_allowed`.

Fixes #9712

### DIFF
--- a/changelog.d/9717.feature
+++ b/changelog.d/9717.feature
@@ -1,0 +1,1 @@
+Add experimental support for [MSC2962](https://github.com/matrix-org/matrix-doc/pull/2962): restricting room access via group membership.
--- a/changelog.d/9717.feature
+++ b/changelog.d/9717.feature
@@ -1,1 +1,1 @@
-Add experimental support for [MSC2962](https://github.com/matrix-org/matrix-doc/pull/2962): restricting room access via group membership.
+Add experimental support for [MSC3083](https://github.com/matrix-org/matrix-doc/pull/3083): restricting room access via group membership.
--- a/synapse/api/room_versions.py
+++ b/synapse/api/room_versions.py
@@ -69,8 +69,8 @@ class RoomVersion:
     limit_notifications_power_levels = attr.ib(type=bool)
     # MSC2174/MSC2176: Apply updated redaction rules algorithm.
     msc2176_redaction_rules = attr.ib(type=bool)
-    # MSC2962: Support the 'restricted' join_rule.
-    msc2962_join_rules = attr.ib(type=bool)
+    # MSC3083: Support the 'restricted' join_rule.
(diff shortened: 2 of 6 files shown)


synapse/api/room_versions.py
```diff
@@ -69,6 +69,8 @@ class RoomVersion:
     limit_notifications_power_levels = attr.ib(type=bool)
     # MSC2174/MSC2176: Apply updated redaction rules algorithm.
     msc2176_redaction_rules = attr.ib(type=bool)
+    # MSC2962: Support the 'restricted' join_rule.
+    msc2962_join_rules = attr.ib(type=bool)
```
clokep, line 73: This is currently MSC2962, but I think we're going to split it out. Maybe I should put up a PR to the draft to reserve a number?

richvdh: maybe do the same as https://github.com/matrix-org/matrix-doc/blob/kegan/spaces-summary/proposals/2946-spaces-summary.md

synapse/api/room_versions.py
```diff
@@ -161,5 +182,6 @@ class RoomVersions:
         RoomVersions.V5,
         RoomVersions.V6,
         RoomVersions.MSC2176,
+        RoomVersions.MSC2962,
```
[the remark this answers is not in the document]
clokep: Is it reasonable to re-use the `spaces` flag or would you prefer a new one?

synapse/api/room_versions.py
clokep: See matrix-org/matrix-doc#3083

richvdh (approved):
lgtm

changelog.d/9717.feature
```diff
@@ -0,0 +1 @@
+Add experimental support for [MSC2962](https://github.com/matrix-org/matrix-doc/pull/2962): restricting room access via group membership.
```
clokep, line 1:
```suggestion
Add experimental support for [MSC3083](https://github.com/matrix-org/matrix-doc/pull/3083): restricting room access via group membership.
```

PR #9720, opened 2021-03-31
Revert "Use 'dmypy run' in lint.sh instead of 'mypy' (#9701)"

clokep:
This reverts commit f380bb77d19e7dd4ee6f61cd489e240ee4aa8fc2.

Reverting since it gives errors that running `mypy` directly doesn't give. So the `lint.sh` script fails for me every time.

I can merge this across branches after it merges.

### DIFF
--- a/changelog.d/9720.misc
+++ b/changelog.d/9720.misc
@@ -1,0 +1,1 @@
+Revert using `dmypy run` in lint script.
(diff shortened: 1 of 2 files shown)


richvdh (approved):
lgtm

clokep:
Thanks! I merged this into `develop` as well.

PR #9721, opened 2021-03-31
Remove functionality associated with unused historical stats tables

### LINKED ISSUE #9602: room_stats_historical and user_stats_historical appear unused
We have a whole lot of code which maintains the `room_stats_historical` and `user_stats_historical` tables (and a config option, `bucket_size`, which configures how it is populated), but afaict they are never actually used.

It's not obvious to me what the purpose of these tables are for, and I'd argue that unless we have a clear usecase for them, the tables and dead code should be stripped out.

It looks like the concept of historical stats was introduced by @ara4n in the initial commits in #4338, though it subsequently got refactored out of recognition by #5971. @ara4n: any comments on the purpose of this, and/or objections to it being ripped out.

Also of interest (and would need updating) is the doc, https://github.com/matrix-org/synapse/blob/v1.29.0/docs/room_and_user_statistics.md.

xmunoz:
- Purge all associated functionality
- Update the docs
- Get unit tests passing

### Pull Request Checklist

Fixes #9602


richvdh (commented):
Hey @xmunoz, thank you very much for doing all this work!

I haven't yet read through it in detail, though at a skim it looks like it makes all the right sort of changes.

However, we need to be a bit careful here: if we simply drop the `*_historical` tables from the database, that means that it will be difficult to roll Synapse back to an earlier version. Generally we try to avoid that sort of situation, because it means that if there is a bug in the next version of Synapse, our users are a bit stuck.

So, one solution here is to just remove the code that updates the tables, but leave the tables in place for now.

(I've been thinking a bit recently about how to manage this situation: https://hackmd.io/@richvdh/BJYXQMQHO has some more details)

xmunoz:
Ok, that makes sense. Is there a suggested PR procedure for actioning this kind of change? It sounds like you'd like me to split this PR into 3 PRs, each with a number indicating the order that they should be merged. Is there a way to tag PRs with a version number, so that you will know when it needs to get merged? Are there any other example of changed that have been merged in this way? If not, that's fine too. I'm happy to be the guinea pig for this.

richvdh:
since there isn't any code that reads from the `_historical` tables currently (as far as I know?), I think we only need two PRs here: one to remove the code that writes to the tables, and later, another that finally removes the tables.

Generally what we do is just write the first PR, and then file a maintenance issue to remind us to do the next step. (The `Z-Future-Maintenance` issue label is meant for that sort of thing, though we haven't yet used it for much.)

xmunoz:
Build is failing now, but I'm not sure why. I simply added a commit to remove the sql schema change and update the changelog.

richvdh:
oh, the build is failing because `flake8-bugbear` has had a new release. Can you merge `develop` into your branch, so that you pick up https://github.com/matrix-org/synapse/pull/9734 ?

richvdh (commented):
This is amazing, thank you so much!

The only thing is that I'd prefer to hold off merging it until we've done the groundwork for schema versions outlined in https://hackmd.io/@richvdh/BJYXQMQHO. If you fancy taking that on (in a separate PR) that would be amazing - otherwise hopefully we'll get to it within the next week or so.

xmunoz:
> The only thing is that I'd prefer to hold off merging it until we've done the groundwork for schema versions outlined in https://hackmd.io/@richvdh/BJYXQMQHO. If you fancy taking that on (in a separate PR) that would be amazing - otherwise hopefully we'll get to it within the next week or so.

I'm not sure what this groundwork is or what it entails, so I'll leave it up to you.

anoadragon453:
I believe this is blocked on https://github.com/matrix-org/synapse/pull/9933.

richvdh:
#9933 is now landed, so this is unblocked. We should bump `SCHEMA_VERSION`, per the example in https://matrix-org.github.io/synapse/develop/development/database_schema.html#synapse-schema-versions.

richvdh:
That said: I'd like us to have at least one release of Synapse which checks `schema_compat_version` rather than `schema_version` before we bump `schema_version`, otherwise we'll break downgrade.

In other words: Synapse 1.37 will introduce support for `schema_compat_version`. We can land this PR in Synapse 1.38 along with a bump of `schema_version`, and it will still be possible to downgrade from 1.38 to 1.37.

richvdh (approved):
lgtm!

Issue #8782, opened 2020-11-19
Proposal: Python / PostgreSQL / Etc. Version Deprecation Policy


callahad:
Synapse currently claims to support Python 3.5 (EOL Sept. 2020) and Postgres 9.5 (EOL Feb. 2021).

We need to plan for how we will remove support for these platform dependencies in a responsible manner.

## Motivation

I'm afraid we're boxing ourselves into a corner where our dependencies might be unsupported on those platform versions, and if a security vulnerability were to arise, we wouldn't be in a position to address it without maintaining an untenable number of forks / backports or forcing all of our customers into an emergency upgrade of their infrastructure.

Articulating a support lifecycle allows our customers to make reasonable plans for managing their infrastructure and allows us to know when we can take advantage of newer platform features.

## Proposal

1. __We follow upstream support lifecycles.__
    When a version of a language, database, or Linux distribution reaches End of Life, we also withdraw our support for that version.

2. __There is no guaranteed grace period.__
    We will not intentionally break compatibility for at least one month after an EOL date, but we explicitly reserve the right to do so should security or other concerns arise in an unmaintained version of a dependency. 

3. __This comes into effect March 2021.__
    We'll announce with the next Synapse release. Synapse will be allowed to break Py3.5, Pg9.5, and Ubuntu 16.04 starting with our first release in March 2021.

## Additional Context

- Element is happy with this policy, including folks who liaise with large EMS and Professional Services customers.
- Ubuntu Xenial (16.04 LTS) ships Py3.5 and Pg9.5; it goes EOL Apr. 2021. This seems to align closely enough with the proposed timeline.
- Platform support lifecycles are relatively long. Python, Postgres, and Ubuntu tend to have annual or biannual releases with five-year support windows. This ensures multiple years of overlap between supported releases; ample time for organizations to validate and deploy new platform versions.
- Folks who are _really_ stuck on old distros can use our Docker images instead. Or make their own packages.

## Questions

- Is a one month, best-effort grace period sufficient?
- Do we need to explicitly enumerate our supported Linux distros somewhere given that we *do* build our own Debian / Ubuntu packages, or can we some how get away without doing that?
-  Does the roll-out (announce next release, implement in March) sound right?

clokep:
Cherry-picking a couple of things:

> Ubuntu Xenial (16.04 LTS) ships Py3.5 and Pg9.5; it goes EOL Feb. 2021. This seems to align nicely with the proposed timeline.

[...]

PR #9723, opened 2021-04-01
Add deprecation policy doc

erikjohnston:
Lifts the conclusions from https://github.com/matrix-org/synapse/issues/8782 to be in repo.

This is a bit of a straw man, so please feel free to quibble with the wording.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -3,6 +3,7 @@
 There are 3 steps to follow under **Installation Instructions**.
 
 - [Installation Instructions](#installation-instructions)
+  - [Platform Dependencies](#choosing-your-server-name)
   - [Choosing your server name](#choosing-your-server-name)
   - [Installing Synapse](#installing-synapse)
     - [Installing from source](#installing-from-source)
@@ -37,6 +38,14 @@ There are 3 steps to follow under **Installation Instructions**.
     - [Setting up a TURN server](#setting-up-a-turn-server)
     - [URL previews](#url-previews)
     - [Troubleshooting Installation](#troubleshooting-installation)
+
+
+## Platform dependencies
+
+Synapse uses a number of platform dependencies such as Python and PostgreSQL,
+and aims to support versions supported upstream. See the
+[deprecation policy](docs/deprecation_policy.md) document for more details.
+
 
 ## Choosing your server name
 
@@ -542,7 +551,7 @@ Alternatively, you can do so from the command line. This can be done as follows:
     register_new_matrix_user -c homeserver.yaml http://localhost:8008
     ```
 
-This will prompt you to add details for the new user, and will then connect to 
+This will prompt you to add details for the new user, and will then connect to
 the running Synapse to create the new user. For example:
 ```
 New user localpart: erikj
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -3,7 +3,6 @@
 There are 3 steps to follow under **Installation Instructions**.
 
 - [Installation Instructions](#installation-instructions)
-  - [Platform Dependencies](#choosing-your-server-name)
   - [Choosing your server name](#choosing-your-server-name)
   - [Installing Synapse](#installing-synapse)
     - [Installing from source](#installing-from-source)
@@ -38,13 +37,6 @@ There are 3 steps to follow under **Installation Instructions**.
     - [Setting up a TURN server](#setting-up-a-turn-server)
     - [URL previews](#url-previews)
     - [Troubleshooting Installation](#troubleshooting-installation)
-
-
-## Platform dependencies
-
-Synapse uses a number of platform dependencies such as Python and PostgreSQL,
-and aims to support versions supported upstream. See the
-[deprecation policy](docs/deprecation_policy.md) document for more details.
 
 
 ## Choosing your server name
--- a/README.rst
+++ b/README.rst
@@ -319,7 +319,7 @@ Platform dependencies
 =====================
 
 Synapse uses a number of platform dependencies such as Python and PostgreSQL,
-and aims to support versions supported upstream. See the
+and aims to follow supported upstream versions. See the
 `<docs/deprecation_policy.md>`_ document for more details.
 
 
--- a/README.rst
+++ b/README.rst
@@ -314,6 +314,15 @@ Testing with SyTest is recommended for verifying that changes related to the
 Client-Server API are functioning correctly. See the `installation instructions
 <https://github.com/matrix-org/sytest#installing>`_ for details.
 
+
+Platform dependencies
+=====================
+
+Synapse uses a number of platform dependencies such as Python and PostgreSQL,
+and aims to support versions supported upstream. See the
+`<docs/deprecation_policy.md>`_ document for more details.
+
+
 Troubleshooting
 ===============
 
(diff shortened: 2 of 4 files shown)


INSTALL.md
```diff
@@ -3,6 +3,7 @@
 There are 3 steps to follow under **Installation Instructions**.
```
richvdh, line 3: this is item 4... :/

I'm hesitant to put this upfront in the installation instructions: there's a lot for new users to understand here, and I'm not entirely sure it belongs here

richvdh: maybe link it from the README instead?

erikjohnston: Fair, its a bit of a tricky balance. Maybe we should point to the doc whenever we mention min requirements?

README.rst
```diff
@@ -314,6 +314,15 @@ Testing with SyTest is recommended for verifying that changes related to the
 Client-Server API are functioning correctly. See the `installation instructions
 <https://github.com/matrix-org/sytest#installing>`_ for details.
 
+
+Platform dependencies
+=====================
+
+Synapse uses a number of platform dependencies such as Python and PostgreSQL,
+and aims to support versions supported upstream. See the
```
anoadragon453, line 322:
```suggestion
and aims to follow supported upstream versions. See the
```

docs/deprecation_policy.md
```diff
@@ -0,0 +1,37 @@
...
+------
+
+For Python and PostgreSQL Synapse will follow the upstream support life cycles,
+i.e. when a version reaches End of Life Synapse will withdraw support for that
+version in future releases.
+
+The team will not intentionally break compatibility for at least one month after
+an EOL date, but does explicitly reserve the right to do so should security or
+other concerns arise in an unmaintained version of a dependency.
+
```
anoadragon453, line 19:
```suggestion
```

docs/deprecation_policy.md
```diff
@@ -0,0 +1,37 @@
...
+Details on the upstream support life cycles for Python and PostgreSQL are
+documented at https://endoflife.date/python and
+https://endoflife.date/postgresql.
+
+
+Context
+-------
+
+It is important for system admins to have a clear understanding of the platform
+requirements of Synapse and deprecation policies so that they can effectively
```
anoadragon453, line 30:
```suggestion
requirements of Synapse and its deprecation policies so that they can effectively
```

docs/deprecation_policy.md
```diff
@@ -0,0 +1,37 @@
...
+
+The team will not intentionally break compatibility for at least one month after
+an EOL date, but does explicitly reserve the right to do so should security or
+other concerns arise in an unmaintained version of a dependency.
+
+
+Details on the upstream support life cycles for Python and PostgreSQL are
+documented at https://endoflife.date/python and
+https://endoflife.date/postgresql.
+
```
anoadragon453, line 24: Do we not want to comment on the policy for any other dependencies?

callahad: Do we have other deps we care about in this regard?

docs/deprecation_policy.md
```diff
@@ -0,0 +1,37 @@
...
+Policy
+------
+
+For Python and PostgreSQL Synapse will follow the upstream support life cycles,
+i.e. when a version reaches End of Life Synapse will withdraw support for that
+version in future releases.
+
+The team will not intentionally break compatibility for at least one month after
+an EOL date, but does explicitly reserve the right to do so should security or
+other concerns arise in an unmaintained version of a dependency.
```
callahad, line 18: ...do we _actually_ want to box ourselves in to this, or should we drop when upstream drops?

erikjohnston: I'd me more than happy to drop this, it seems a bit pointless.

docs/deprecation_policy.md
anoadragon453: Probably not, but it may be nice to explicitly say that we don't.

docs/deprecation_policy.md
callahad: Make it so

callahad:
...I wonder if we could replace this whole document by just being very, very explicit up front:

> Synapse requires Python ⩾ 3.6 and PostgreSQL ⩾ 9.6.
> We follow upstream end-of-life dates for [Python](https://endoflife.date/python) and [Postgres](https://endoflife.date/postgresql).

If we wanted to be _really_ verbose we could also say:

> We expect to drop support for PostgreSQL 9.6 in November 2021, and Python 3.6 in December 2021.

erikjohnston:
> ...I wonder if we could replace this whole document by just being very, very explicit up front:
> 
> > Synapse requires Python ⩾ 3.6 and PostgreSQL ⩾ 9.6.
> > We follow upstream end-of-life dates for [Python](https://endoflife.date/python) and [Postgres](https://endoflife.date/postgresql).
> 
> If we wanted to be _really_ verbose we could also say:
> 
> > We expect to drop support for PostgreSQL 9.6 in November 2021, and Python 3.6 in December 2021.

The main thing I want is a document I can link to each time we go through this process, rather than having to come up with words each time. I also do think its beneficial to explicitly document why we're doing it, even if it is a bit obvious.

changelog.d/9723.doc
```diff
@@ -0,0 +1 @@
+Add a document describing the deprecation policy for platform dependencies.
```
clokep, line 1: Maybe just:

> Document the deprecation policy for platform dependencies.

docs/deprecation_policy.md
```diff
@@ -0,0 +1,33 @@
+Deprecation Policy for Platform Dependencies
+============================================
+
+Synapse has a number of platform dependencies, including Python and PostgreSQL.
+This document outlines the policy towards which versions we support, and how we
+drop support for versions in the future.
```
clokep, line 6: Maybe?

```suggestion
Synapse has a number of platform dependencies, including Python and PostgreSQL.
This document outlines the policy towards which versions we support, and when we
drop support for versions in the future.
```

docs/deprecation_policy.md
```diff
@@ -0,0 +1,33 @@
...
+
+
+Policy
+------
+
+For Python and PostgreSQL Synapse will follow the upstream support life cycles,
+i.e. when a version reaches End of Life Synapse will withdraw support for that
+version in future releases.
```
clokep, line 14:
```suggestion
Synapse follows the upstream support life cycles for Python and PostgreSQL,
i.e. when a version reaches End of Life Synapse will withdraw support for that
version in future releases.
```

docs/deprecation_policy.md
clokep: We can always add more in the future. 🤷

clokep (approved):
I think this looks good, minus some nitpicking of words. I like including *why* we're doing it. It is useful context for people to have.

docs/deprecation_policy.md
anoadragon453, line 6:
```suggestion
This document outlines the policy of which versions of specific dependencies
we support, and how we drop support for old versions in the future.
```

anoadragon453 (approved):
A minor wording change, but otherwise lgtm!

Issue #9574, opened 2021-03-09
Serving synapse on a subpath breaks OIDC due to session cookie having incorrect path


imaspeer:
The session cookie synapse sets before redirecting to the OIDC provider always has its path set to `/_synapse/client/oidc` regardless of the configured `public_baseurl`.

If synapse is served on a subpath rather than at the webserver's root (e.g. `domain.example/matrix`), this causes the cookie to not be sent with the callback request (since the path `/matrix/_synapse/client/oidc/callback` does not match the cookie's), which then fails with the error `missing_session`.

anoadragon453:
We currently only support endpoints being hosted at the root path of a domain, hence the hardcoded path. However we have been planning towards lifting this restriction.

I'm curious whether you've been successfully hosting your Client Server API (and Federation API?) on a subpath without running into any other issues?

imaspeer:
Hi,

We're running a non-federating server, so can't speak about federation API ; regarding client-server API, in a few weeks of running 1.28.0 with patched OIDC we have not ran into any other issue.

I'm attaching the patch to `synapse/handlers/oidc_handlers.py` we use; it works for us, however we haven't tried it against latest versions.
[oidc_cookie_prefix.patch.txt](https://github.com/matrix-org/synapse/files/6229051/oidc_cookie_prefix.patch.txt)

anoadragon453:
@imaspeer Thanks for providing a patch! I've modified it slightly and stuck it up as a PR at #9726 to get some feedback from the team on it :slightly_smiling_face:

richvdh:
@imaspeer out of interest, what is the usecase for hosting synapse at a sub-path? This came up in another context and I'm interested to know if there are important reasons we should support it.

anoadragon453:
Closing this issue now that the patch has landed - but feel free to still comment here about your use case! This likely won't be the last thing to need fixing w.r.t hosting endpoints on subpaths :)

PR #9726, opened 2021-04-01
Allow OIDC cookies to work on non-root public baseurls

anoadragon453:
Applied a (slightly modified) patch from https://github.com/matrix-org/synapse/issues/9574.

As far as I understand this would allow the cookie set during the OIDC flow to work on deployments using public baseurls that do not sit at the URL path root.

Besides the code presented here, there's a bigger question of whether we want to support non-root base URLs at all. That question has been raised on the linked issue.


synapse/handlers/oidc_handler.py
```diff
@@ -782,8 +789,14 @@ async def handle_redirect_request(
...
         for cookie_name, options in _SESSION_COOKIES:
             request.cookies.append(
-                b"%s=%s; Max-Age=3600; %s"
-                % (cookie_name, cookie.encode("utf-8"), options)
+                b"%s=%s; Max-Age=3600; Path=%s; %s"
+                % (
+                    cookie_name,
+                    cookie.encode("utf-8"),
+                    self._callback_path_prefix.encode("utf-8")
+                    + b"/_synapse/client/oidc",
```
richvdh, line 797: why not include `/_synapse/client/oidc` in `self._callback_path` ?

synapse/handlers/oidc_handler.py
```diff
@@ -282,6 +282,13 @@ def __init__(
         self._config = provider
         self._callback_url = hs.config.oidc_callback_url  # type: str
 
+        # TODO: This is pretty much a hack to get the path specified by public_baseurl.
+        # It'd probably be nicer to have a config option that lets you specify a custom
+        # path, which we'd then use here.
```
richvdh, line 287: why would we want a custom path here?

richvdh (dismissed):

synapse/handlers/oidc_handler.py
anoadragon453: I meant that instead of specifying one's `public_baseurl` as, say, `https://example.com/specific/deployment`, you would have one option for the hostname, and another for the path. This would save us from needing to parse the path from the URL, though I suppose that isn't probably wouldn't buy us *that* much.

synapse/handlers/oidc_handler.py
anoadragon453: Hmm, I suppose it's handy to have a reference to the path extracted from `public_baseurl`, which we can then add on to in places (as we do here).

But while we're only using this in one place, we can move the concatenation out of the weeds/runtime code path.

synapse/handlers/oidc_handler.py
richvdh: yeah, I think that would be more confusing than helpful.

richvdh: ... on which basis, can you update the TODO?

synapse/handlers/oidc_handler.py
```diff
@@ -285,9 +285,15 @@ def __init__(
...
         # path, which we'd then use here.
-        self._callback_path_prefix = urlparse(hs.config.public_baseurl).path
-        if self._callback_path_prefix.endswith("/"):
-            self._callback_path_prefix = self._callback_path_prefix[:-1]
+        public_baseurl_path = urlparse(hs.config.public_baseurl).path
+        if public_baseurl_path.endswith("/"):
+            public_baseurl_path = public_baseurl_path[:-1]
+
+        # Calculate the prefix for OIDC callback paths based on the public_baseurl.
+        # We'll insert this into the Path= parameter of any session cookies we set.
```
richvdh, line 293: stick this a few lines higher?

synapse/handlers/oidc_handler.py
```diff
@@ -285,9 +285,15 @@ def __init__(
         # TODO: This is pretty much a hack to get the path specified by public_baseurl.
         # It'd probably be nicer to have a config option that lets you specify a custom
         # path, which we'd then use here.
-        self._callback_path_prefix = urlparse(hs.config.public_baseurl).path
-        if self._callback_path_prefix.endswith("/"):
-            self._callback_path_prefix = self._callback_path_prefix[:-1]
+        public_baseurl_path = urlparse(hs.config.public_baseurl).path
+        if public_baseurl_path.endswith("/"):
```
richvdh, line 289: isn't it guaranteed to end with `/`? In which case, skip this, and remove the `/` on the start of the suffix?

richvdh (changes requested):
minor things

richvdh (approved):
lgtm!

PR #9734, opened 2021-04-02
Fix version for bugbear

ShadowJonathan:
Bugbear [added new checks](https://github.com/PyCQA/flake8-bugbear/releases/tag/21.4.1), however, because we're both not actively ignoring new checks (to a degree) or fixing the bugbear version, these got auto-updated in CI, breaking it because `flake8` is now disapproving of the codebase.

### Pull Request Checklist

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -99,7 +99,7 @@ CONDITIONAL_REQUIREMENTS["lint"] = [
     "isort==5.7.0",
     "black==20.8b1",
     "flake8-comprehensions",
-    "flake8-bugbear",
+    "flake8-bugbear==21.3.2",
     "flake8",
 ]
 
--- a/changelog.d/9734.misc
+++ b/changelog.d/9734.misc
@@ -1,0 +1,1 @@
+Pin flake8-bugbear's version.


ShadowJonathan:
thanks!

setup.py
```diff
@@ -99,7 +99,7 @@ def exec_file(path_segments):
     "isort==5.7.0",
     "black==20.8b1",
     "flake8-comprehensions",
-    "flake8-bugbear",
+    "flake8-bugbear==21.3.2",
```
richvdh, line 102: @callahad @ShadowJonathan: it's a shame that no attention was given to the comment above:

> We pin black so that our tests don't start failing on new releases.

This implies that we pin `flake8-bugbear` for some other reason?

ShadowJonathan: Where does that comment come from?

> This implies that we pin flake8-bugbear for some other reason?

We don't pin bugbear for any other reason, no.

richvdh: four lines above. Please remember to check the context around code you are changing.

ShadowJonathan: Ah, i didn't see that, i was mainly focused on just pinning bugbear, should an addition to the comment be made?

richvdh: it'd be nice. It may not be worth it. I'm mostly just asking that both you and @callahad remember to look for context around stuff you are changing/reviewing.

PR #9735, opened 2021-04-02
Check for proper membership during a local join of restricted rooms.

### LINKED ISSUE #9714: Experimental support for checking space membership when a local user joins a restricted room
This is a follow-on to #9712 to implement the changes necessary when a client calls `/join` on a restricted room for [MSC3083](https://github.com/matrix-org/matrix-doc/pull/3083).

* Figure out where to hook into the local join-room flow.
* Check the space membership based on the `allow` parameter of the `join_rules`.

clokep:
Expands the support for [MSC3083](https://github.com/matrix-org/matrix-doc/pull/3083) by checking membership in a space when joining a restricted room.

Fixes #9714

### DIFF
--- a/synapse/handlers/room_member.py
+++ b/synapse/handlers/room_member.py
@@ -295,8 +295,8 @@ class RoomMemberHandler(metaclass=abc.ABCMeta):
                 newly_joined = prev_member_event.membership != Membership.JOIN
                 is_invite = prev_member_event.membership == Membership.INVITE
 
-            # If the member is not already in the room, check if they should be
-            # allowed access via membership in a space.
+            # If the member is not already in the room and is not accepting an invite,
+            # check if they should be allowed access via membership in a space.
             if (
                 newly_joined
                 and not is_invite
--- a/synapse/handlers/room_member.py
+++ b/synapse/handlers/room_member.py
@@ -289,14 +289,20 @@ class RoomMemberHandler(metaclass=abc.ABCMeta):
 
         if event.membership == Membership.JOIN:
             newly_joined = True
+            is_invite = False
             if prev_member_event_id:
                 prev_member_event = await self.store.get_event(prev_member_event_id)
                 newly_joined = prev_member_event.membership != Membership.JOIN
+                is_invite = prev_member_event.membership == Membership.INVITE
 
             # If the member is not already in the room, check if they should be
             # allowed access via membership in a space.
-            if newly_joined and not await self._can_join_restricted_room(
-                prev_state_ids, room_id, user_id
+            if (
+                newly_joined
+                and not is_invite
+                and not await self._can_join_restricted_room(
+                    prev_state_ids, room_id, user_id
+                )
             ):
                 raise AuthError(
                     403,
--- a/synapse/handlers/room_member.py
+++ b/synapse/handlers/room_member.py
@@ -210,7 +210,7 @@ class RoomMemberHandler(metaclass=abc.ABCMeta):
         if join_rules_event.content.get("join_rule") != JoinRules.MSC3083_RESTRICTED:
             return True
 
-        # If allowed is of the wrong form, then only allows invites.
+        # If allowed is of the wrong form, then only allow invited users.
         allowed_spaces = join_rules_event.content.get("allow", [])
         if not isinstance(allowed_spaces, list):
             return False
--- a/synapse/handlers/room_member.py
+++ b/synapse/handlers/room_member.py
@@ -29,6 +29,7 @@ from synapse.api.errors import (
     SynapseError,
 )
 from synapse.api.ratelimiting import Ratelimiter
+from synapse.api.room_versions import RoomVersion
 from synapse.events import EventBase
 from synapse.events.snapshot import EventContext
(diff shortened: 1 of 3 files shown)


clokep:
Complement tests at matrix-org/complement#93 although this probably needs some unit tests too.

clokep:
I think this (and the complement PR) are both probably ready for review. I'm not sure what additional unit tests I should include in Synapse itself?

synapse/handlers/room_member.py
```diff
@@ -235,9 +289,25 @@ async def _local_membership_update(
...
+
+            # If the member is not already in the room, check if they should be
+            # allowed access via membership in a space.
+            if (
+                newly_joined
+                and not is_invite
+                and not await self._can_join_restricted_room(
+                    prev_state_ids, event.room_version, user_id
+                )
+            ):
```
clokep, line 306: This partially feels like we're re-implementing the auth checks here, but not sure there's a more reasonable way to do it.

clokep:
(I rebased this to pick up some changes on develop, hopefully that's OK!)

synapse/handlers/room_member.py
```diff
@@ -235,9 +289,25 @@ async def _local_membership_update(
 
         if event.membership == Membership.JOIN:
             newly_joined = True
+            is_invite = False
```
richvdh, line 292: `is_invite` implies that the current event is an invite. Suggest `user_is_invited` or somesuch

synapse/handlers/room_member.py
```diff
@@ -178,6 +179,59 @@ async def ratelimit_invite(
 
         await self._invites_per_user_limiter.ratelimit(requester, invitee_user_id)
 
+    async def _can_join_restricted_room(
+        self, state_ids: StateMap[str], room_version: RoomVersion, user_id: str
+    ) -> bool:
+        """
+        Check whether a user can join a restricted room.
```
richvdh, line 186: this gets run on *all* rooms, not just restricted ones, so I think it needs a rename and a wording tweak... maybe `_can_join_without_invite` ?

synapse/handlers/room_member.py
```diff
@@ -178,6 +179,59 @@ async def ratelimit_invite(
...
+        join_rules_event_id = state_ids.get((EventTypes.JoinRules, ""), None)
+        if not join_rules_event_id:
+            return True
+
+        # If the join rule is not restricted, this doesn't apply.
+        join_rules_event = await self.store.get_event(join_rules_event_id)
+        if join_rules_event.content.get("join_rule") != JoinRules.MSC3083_RESTRICTED:
+            return True
+
+        # If allowed is of the wrong form, then ignore it (and treat the room as public).
```
richvdh, line 210: > (and treat the room as public).

this feels like a questionable decision - we should fail towards more security.

synapse/handlers/room_member.py
```diff
@@ -178,6 +179,59 @@ async def ratelimit_invite(
...
+
+        # Get the list of joined rooms and see if there's an overlap.
+        joined_rooms = await self.store.get_rooms_for_user(user_id)
+
+        # Pull out the other room IDs, invalid data gets filtered.
+        for space in allowed_spaces:
+            if not isinstance(space, dict):
+                continue
+
+            soace_id = space.get("space")
```
richvdh, line 223: soace

synapse/handlers/room_member.py
richvdh: agreed, but :man_shrugging:

richvdh (commented):
looks generally good. a few nits.

synapse/handlers/room_member.py
clokep: I seem to have swapped this from the original MSC2962 phrasing of:

> If the `allow` key is an empty list (or not a list at all), no users are allowed to join without an invite.

To the following in MSC3083:

> If the `allow` key is an empty list (or not a list at all), then the room reverts to standard `public` join rules.

I don't really recall *why* I did that, but I'm guessing it was due to the phrasing of how `restricted` differs from other join rules:

> the same as `public` from the perspective of the auth rules, but with the additional caveat that servers are expected to check the `allow` rules before generating a `join` event (whether for a local or a remote user).

I agree though that this seems quite questionable. I'm going to change this back to the original MSC2962 phrasing.

clokep: I made this change (and updated the MSC).

clokep:
The new tests for complement aren't actually running yet unfortunately. See matrix-org/pipelines#139.

synapse/handlers/room_member.py
```diff
@@ -207,10 +210,10 @@ async def _can_join_restricted_room(
         if join_rules_event.content.get("join_rule") != JoinRules.MSC3083_RESTRICTED:
             return True
 
-        # If allowed is of the wrong form, then ignore it (and treat the room as public).
+        # If allowed is of the wrong form, then only allows invites.
```
richvdh, line 213:
```suggestion
        # If allowed is of the wrong form, then only allow invited users.
```

richvdh (approved):
lgtm

clokep:
Thank you for the review! 🎉

Issue #9525, opened 2021-03-02
[discussion] Changing the behaviour of `ResponseCache.wrap_conditional` to only honour the very first caller


ShadowJonathan:
cc @richvdh, @clokep 

I feel like this warrants some further discussion, as i think i don't exactly understand the rationale of *why* we're pretty much doing this, and i think that some gotchas werent received during discussion in `#synapse-dev`, so i'm making this verbose issue on it to get on the same page.

### Status

#9358 introduced `ResponseCache.wrap_conditional`, #9522 builds on it while fixing #9507, and #9458 adds tests that validate its functionality (as given by #9358)

This feature (as of the making of this issue) has the following functionality;

- It acts like `wrap`, but has an extra positional argument which the caller can provide a "`conditional`" with, this `conditional` would take the result of the call as it's first argument, and return a `bool` indicating if the result should be cached in a timed fashion or not.
- Currently, if another caller would call `wrap_conditional` with the same `key`, their "`conditional`" callable would be added to the set of conditionals, to be all evaluated when the call returns.
- When the call returns, this set of conditionals need *all* to return `True` (and not throw an exception, which is what #9522 fixes) for the result to be cached in a timely fashion.
- Any subsequent call to `wrap_conditional` will have their `conditional` argument ignored, no matter if it evaluated to `False` or not.

### Discussion terminology

For the purpose of discussion, I'll define the following terms;
- `wrap*` is `wrap` and/or `wrap_conditional`, this is in cases where the desired functionality is indistinguishable between the two. (e.g. "doesn't matter if a `conditional` is supplied or not")
- the "callers", or "the caller", are any function that calls `wrap*` (with the same `key`, in the context of this discussion)
- "the very first caller" is the first (timeline-wise) to call to `wrap*` or `wrap_conditional`, and start the inner call. This excludes *any* future callers.
- "the first callers" are the callers that have called `wrap*` but are waiting for the inner call to complete, awaiting on a `Deferred`-like object.
- "subsequent callers" are the callers excluded from the first callers, these callers 

### Arguments

My argument on going forward with this is to make *any* caller (first or subsequent) be able to evict/invalidate the cache when a `conditional` they give evaluates as `False`.

[...]

PR #9739, opened 2021-04-03
Introduce SyncResponseCache and add /sync response cache timeout

### LINKED ISSUE #3880: cache sync results for a while after the request completes

### LINKED ISSUE #8518: Sync requests immediately return with empty payload

ShadowJonathan:
This is a compromise between #9358's initial handling and the subsequent discussion for semantics (continuing in #9525).

This PR introduces a subclass to `synapse.util.caches.response_cache.ResponseCache` that adds the `wrap_conditional` behaviour that is applicable for `/sync`; the Very First Caller to the `wrap*` function gets to decide the condition-check function for it's result, whether or not it'll instantly expire.

This PR also adds an experimental config key: `sync_cache_timeout_ms`, the new cache timeout (in ms) for `/sync` results to expire (based on their initial arguments).

### A few notes

I'll restate it; This PR is a *compromise*, I would want to continue the discussion in #9525 first, but I don't want to have that bog the fix for #8518, thus this PR adds an *explicit subclass* to signal that.

I *will not* merge `SyncResponseCache` with `ResponseCache` because of that, because I would want to have the discussion in #9525 first.

`SyncResponseCache.wrap_conditional`'s behaviour is *custom* to `/sync`.

### Pull Request Checklist

Fixes #8518
Fixes #3880


ShadowJonathan:
(this is a *bit* over the 200-LoC limit, but 135 lines of these come from `test_sync_responsecache.py`, which is mostly boilerplate anyway.)

changelog.d/9739.feature
```diff
@@ -0,0 +1 @@
+Added experimental support to cache `/sync` responses with config key `experimental_features.sync_cache_timeout_ms` (in milliseconds).
```
ShadowJonathan, line 1: (Do we announce experimental settings like these? Is this even experimental?)

synapse/util/caches/sync_response_cache.py
```diff
@@ -0,0 +1,101 @@
...
+        if cond is None:
+            return True
+        else:
+            try:
+                # Below type annotation is needed for mypy to shush about some statements being unreachable,
+                # we essentially have to not trust other functions to be able to correctly recover from any fallout
+                # (and log it)
+                res = cond(result)  # type: Any
+            except Exception:
+                logger.exception(
```
ShadowJonathan, line 32: `.exception` because then it throws it to sentry, i think this is pretty sentry-worthy

synapse/util/caches/sync_response_cache.py
```diff
@@ -0,0 +1,101 @@
...
+                    logger.warning(
+                        "[%s]: Conditional %r returned non-bool value %r (for key %r)",
+                        self._name,
+                        cond,
+                        res,
+                        key,
+                    )
+                    # Return concrete boolean value based on falsy or truthiness.
+                    # If this raises, then so be it, then this value wasn't ever supposed to be true" or "false"
+                    # anyways, then have it be a scream test.
```
ShadowJonathan, line 51: I could be more correct by wrapping a `try`-`catch` around this, but i think that'd be a bit too over-cautious, i dont know what situations could cause `bool()` to raise, but coupled with the limited use, i think this is fine.

Still added a comment, though.

ShadowJonathan: (Yes, i'd like it to be *this* "correct")

synapse/util/caches/sync_response_cache.py
```diff
@@ -0,0 +1,101 @@
...
+                    return bool(res)
+                else:
+                    return res
+
+    # Copy this method wholesale from ResponseCache to be able to alter the inner `remove` function
+    def set(self, key: T, deferred: defer.Deferred) -> defer.Deferred:
```
ShadowJonathan, line 57: I couldn't not copy this without editing `ResponseCache` to use a method instead of the `::{{remove}}` closure, I didn't wanna touch `ResponseCache`, so i just copied this part.

Should I make `ResponseCache.set` more generic to allow for it's "remove" closure to be class-defined, or is this okay? (Could add a little bit more maintenance burden to leave it like this)

tests/util/caches/test_sync_responsecache.py
```diff
@@ -0,0 +1,134 @@
+from synapse.util.caches.sync_response_cache import SyncResponseCache
+
+from tests.server import get_clock
+from tests.unittest import TestCase
+
+# A random callback that returns an object
+CALLBACK = lambda: OBJ
+
+# An object, can be equalized to itself
+OBJ = {0}
```
ShadowJonathan, line 10: All that matters is that i have something with an identifier that `x == x`, so i can easily boilerplate it.

changelog.d/9739.bugfix
```diff
@@ -0,0 +1 @@
+Fix #8518 and #3880.
```
richvdh, line 1: These won't get auto-linked by towncrier. Please spell out the bugs you are fixing - linking back to the GH issue number isn't normally necessary because anyone who is really interested can follow the link to the PR and thence to the GH issues.

changelog.d/9739.feature
richvdh: Well yes, we announce experimental features. I don't think this should be one though. Experimental features are more appropriate when we are making changes to the Matrix protocol; this is all internal to Synapse. 

If it needs shaking out before releasing (and I'm not convinced it does), we can do that with test deployments using branches rather than with feature flags.

richvdh (commented):
well, I hear your point that you don't want to merge this with ResponseCache, but forking the code (even via a subclass) is a poor solution here: it means we have twice as much stuff to maintain, and having two similar-but-different things is always a recipe for confusion.  (Haven't you complained about the two HttpClient implementations in the past?)

I don't really think this is the right solution, and I'm not inclined to accept it. The best I can suggest is waiting for now, and I'll try to put up a PR with my suggested approach and we can try to agree as a group which is the right way forward.

richvdh:
I've *finally* put up my counter-proposal at #10157.

ShadowJonathan:
Thanks, i'll close this PR so that all attention can be focused on that one.

Issue #9715, opened 2021-03-30
Experimental support for checking space membership when a remote user joins a restricted room


clokep:
Follow-on to #9712, analogous to #9714, but for the `/make_join` / `/send_join` endpoints for [MSC3083](https://github.com/matrix-org/matrix-doc/pull/3083).

* Figure out where to hook into the make-join / send-join flow.
* Check the space membership based on the allow parameter of the join_rules.
* Reject if the server is not a member of the space. (The MSC talks a bit about peeking in this case, but this is being left unimplemented for now.)

Hopefully this can re-use much of the code from #9714, but will hook into it in a different spot.

clokep:
A couple of notes: the code for this is in the `FederationHandler`, see the `on_make_join_request` and `on_send_join_request` methods. It isn't clear to me whether we want to do it in one of those or both of those. We probably want to do it as early as possible, so probably in `/make_join`, but we might need to do it again in `/send_join` if any of the data that comes back is untrusted (but I assume the whole point of the `/make_join` / `/send_join` workflow is that you can't forge that data between them...)

Another thought is that we might want to check in both in-case the membership changed in the space between the `/make_join` and `/send_join` calls.

PR #7026, opened 2020-03-03
Remove unused federation endpoint (`query_auth`)

clokep:
This removes the unused federation endpoint at `/_matrix/federation/v1/query_auth/{roomId}/{eventId}`.

As part of this it also:
* Moves `_check_sigs_and_hash_and_fetch` from `FederationBase` to `FederationClient` (since it calls into `FederationClient.get_pdu`).
* Converts `_check_sigs_and_hash_and_fetch` to be async/await.
* Fixes type issues that `mypy` found.

~~I need to put up an MSC for this work still, but I wanted to see it run through CI.~~ See [MSC2451](https://github.com/matrix-org/matrix-doc/pull/2451).

Fixes #6997

PR #9743, opened 2021-04-05
Add type hints to the federation handler and server.

clokep:
Adds some missing type hints to the `FederationHandler` and `FederationServer`. It also removes an unused method which should have been removed in #7026.

Some pre-work for #9715.

### DIFF
--- a/synapse/federation/federation_server.py
+++ b/synapse/federation/federation_server.py
@@ -739,22 +739,20 @@ class FederationServer(FederationBase):
 
         await self.handler.on_receive_pdu(origin, pdu, sent_to_us_directly=True)
 
-    def __str__(self):
+    def __str__(self) -> str:
         return "<ReplicationLayer(%s)>" % self.server_name
 
     async def exchange_third_party_invite(
         self, sender_user_id: str, target_user_id: str, room_id: str, signed: Dict
-    ):
-        ret = await self.handler.exchange_third_party_invite(
+    ) -> None:
+        await self.handler.exchange_third_party_invite(
             sender_user_id, target_user_id, room_id, signed
         )
-        return ret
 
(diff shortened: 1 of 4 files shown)


synapse/federation/federation_server.py
```diff
@@ -739,22 +739,20 @@ async def _handle_received_pdu(self, origin: str, pdu: EventBase) -> None:
...
+        await self.handler.exchange_third_party_invite(
             sender_user_id, target_user_id, room_id, signed
         )
-        return ret
 
-    async def on_exchange_third_party_invite_request(self, event_dict: Dict):
-        ret = await self.handler.on_exchange_third_party_invite_request(event_dict)
-        return ret
+    async def on_exchange_third_party_invite_request(self, event_dict: Dict) -> None:
+        await self.handler.on_exchange_third_party_invite_request(event_dict)
```
ShadowJonathan, line 753: this method doesnt return results anymore?

clokep: Hmm...so this is a bit tricky `FederationHandler.on_exchange_third_party_invite_request` always returns `None`, but it looks like the result of this is actually returned via `FederationThirdPartyInviteExchangeServlet`, so something is a bit funky here.

It looks like it *should* be just returning `{}` instead.

clokep: Oops, actually `FederationThirdPartyInviteExchangeServlet` calls the handler, not here so that's unrelated.

ShadowJonathan: I'm just pointing to changing the method behaviour on this, i have no other context, but no worries.

clokep: 7e030a2674c5e903f84c4e9c3335c5524b35b415 will stop using the returned value from this, which is always `None`.

erikjohnston (approved):

PR #9753, opened 2021-04-06
Fix reported bugbear: too broad exception assertion

anoadragon453:
This reported issue was breaking develop for me. And to be fair, it's a valid issue to fix. The issue is well-described in the error:

>tests/config/test_load.py:40:9: B017 assertRaises(Exception): should be considered evil. It can lead to your test passing even if the code being tested is never executed due to a typo. Either assert for a more specific exception (builtin or custom), use assertRaisesRegex, or use the context manager form of assertRaises.

This issue was blocking #6739's CI from passing.

### DIFF
--- a/changelog.d/9753.misc
+++ b/changelog.d/9753.misc
@@ -1,0 +1,1 @@
+Check that a `ConfigError` is raised, rather than simply `Exception`, when appropriate in homeserver config file generation tests.
--- a/tests/config/test_load.py
+++ b/tests/config/test_load.py
@@ -20,6 +20,7 @@ from io import StringIO
 
 import yaml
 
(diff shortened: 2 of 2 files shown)


ShadowJonathan:
This was effectively fixed with #9734, because it was breaking develop over the weekend, did you accidentally download the newer version?

I'm delaying pushing the bugbear version forward and adding these fixes until ive un-ignored all the current ones, FWIW.

ShadowJonathan:
This was effectively fixed with #9734, which commit was failing on develop?

clokep (approved):
Seems to make sense. 👍

anoadragon453:
@ShadowJonathan aha, thanks for the context!

PR #9763, opened 2021-04-07
Check for space membership during a remote join of a restricted room.

### LINKED ISSUE #9715: Experimental support for checking space membership when a remote user joins a restricted room

clokep:
This depends on #9735 and #9800, but applies the same logic for remote joins. It only applies this logic during the `/send_join` aspect of the handshake (not the `/make_join`) since this seems to be where we apply other error checking logic. I think we need to do it regardless during `/send_join` so I'm not sure of the benefit of doing it in both places.

Fixes #9715


clokep:
See matrix-org/complement#99 for integration tests on this.

synapse/handlers/federation.py
```diff
@@ -1667,16 +1667,52 @@ async def on_send_join_request(self, origin: str, pdu: EventBase) -> JsonDict:
...
+
+        # Check if the user is already in the room or invited to the room.
+        user_id = event.state_key
+        prev_member_event_id = prev_state_ids.get((EventTypes.Member, user_id), None)
+        newly_joined = True
+        is_invite = False
+        if prev_member_event_id:
+            prev_member_event = await self.store.get_event(prev_member_event_id)
+            newly_joined = prev_member_event.membership != Membership.JOIN
+            is_invite = prev_member_event.membership == Membership.INVITE
```
clokep, line 1686: Much of this logic (and the if-statement below) is now duplicated between this and local joins. I wonder if we should instead update `can_join_without_invite` to raise a `SynapseError` and try to push some of this logic in there? The local join logic has a couple of interleaved bits though I think.

synapse/handlers/federation.py
```diff
@@ -1994,13 +2030,36 @@ async def _handle_new_event(
...
+            state: The state events to calculate the event context from. This is
+                ignored if context is provided.
+            auth_events: The auth events to calculate the event context from. This is
+                ignored if context is provided.
+            backfilled: True if the event was backfilled.
+
+        Returns:
+             The event context.
+        """
+        if not context:
```
clokep, line 2055: There's not that many callers to `_handle_new_event`, I wonder if it would make sense to always require calling `_prep_event` separately first to pass in the context?

synapse/handlers/federation.py
```diff
@@ -1667,16 +1667,52 @@ async def on_send_join_request(self, origin: str, pdu: EventBase) -> JsonDict:
         # would introduce the danger of backwards-compatibility problems.
         event.internal_metadata.send_on_behalf_of = origin
 
-        context = await self._handle_new_event(origin, event)
+        # Calculate the event context.
+        context = await self._prep_event(
+            origin, event, state=None, auth_events=None, backfilled=False
+        )
```
richvdh, line 1673: hrhrm. `_prep_event` has some side-effects that I'm not *sure* we want to do just here. we might want to split it in half (and see if we can think of better names while we're at it.)

synapse/handlers/federation.py
```diff
@@ -1667,16 +1667,52 @@ async def on_send_join_request(self, origin: str, pdu: EventBase) -> JsonDict:
...
+
+        # Get the current state at the to-be created event.
```
richvdh, line 1675: The event has already been created at this point...

```suggestion
        # Get the state before the new event.
```

synapse/handlers/federation.py
richvdh: I don't think it would be the end of the world to have this logic in both `_local_membership_update` and `can_join_without_invite`, to save us from duplicating the `if newly_joined and not is_invite` condition. I don't think it matters too much though.

synapse/handlers/federation.py
```diff
@@ -1667,16 +1667,52 @@ async def on_send_join_request(self, origin: str, pdu: EventBase) -> JsonDict:
...
+        user_id = event.state_key
+        prev_member_event_id = prev_state_ids.get((EventTypes.Member, user_id), None)
+        newly_joined = True
+        is_invite = False
+        if prev_member_event_id:
+            prev_member_event = await self.store.get_event(prev_member_event_id)
+            newly_joined = prev_member_event.membership != Membership.JOIN
+            is_invite = prev_member_event.membership == Membership.INVITE
+
+        # We retrieve the room member handler here as to not cause a cyclic dependency
```
richvdh, line 1688: generally I would take this as a sign that `can_join_without_invite` needs to find somewhere else to live. I appreciate that may be fiddly, though.

synapse/handlers/federation.py
richvdh: as above - I think it might make more sense to split `_prep_event` in half.

synapse/handlers/federation.py
clokep: Fair enough. We do this in a couple of other spots in this file, I think. 😢

synapse/handlers/federation.py
clokep: From my reading, the parts we might not want to do are the call to `maybe_kick_guest_users` and maybe caching joined hosts? Does that match what you were thinking?

synapse/handlers/federation.py
clokep: I'm struggling a bit to find a good spot for this, I was thinking maybe the `RoomHandler`, but that pulls in `RoomMemberHandler` so I think it would have the same issue. Maybe the `EventCreationHandler` or a new handler. 😢 

The other spot we do this seems to be in exchanging 3PIDs.

synapse/handlers/federation.py
richvdh: I think I would split it after `do_auth` (ie, at https://github.com/matrix-org/synapse/blob/94bdb01a90f09a67b2d83d1af43be060676f354c/synapse/handlers/federation.py#L2269). `do_auth` is only called there, so I think you can move the first half of `_prep_event` into `do_auth`.

clokep:
Arg, I merged this instead of #9800 by mistake. I've manually reverted it in e8816c6aced208cdf1310393a212b5109892ffbf.

PR #9766, opened 2021-04-08
Require py36, Postgres 9.6, and sqlite 3.22

richvdh:
~based on #9769~

### DIFF
--- a/synapse/storage/database.py
+++ b/synapse/storage/database.py
@@ -2059,15 +2059,12 @@ def make_in_list_sql_clause(
 KV = TypeVar("KV")
 
 
-def make_tuple_comparison_clause(
-    _database_engine: BaseDatabaseEngine, keys: List[Tuple[str, KV]]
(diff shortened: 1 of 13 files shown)


synapse/storage/database.py
```diff
@@ -2060,68 +2060,20 @@ def make_in_list_sql_clause(
 
 
 def make_tuple_comparison_clause(
-    database_engine: BaseDatabaseEngine, keys: List[Tuple[str, KV]]
+    _database_engine: BaseDatabaseEngine, keys: List[Tuple[str, KV]]
```
erikjohnston, line 2063: Can we remove this arg from all the call locations? Given we need to check that nothing uses kwargs, I don't see much point in keeping it around

erikjohnston (approved):

synapse/storage/database.py
richvdh: if you prefer, sure.

PR #9770, opened 2021-04-08
Fix sharded federation sender sometimes using 100% CPU.

erikjohnston:
We pull all destinations requiring catchup from the DB in batches.
However, if all those destinations get filtered out (due to the
federation sender being sharded), then the `last_processed` destination
doesn't get updated, and we keep requesting the same set repeatedly.


clokep (approved):
Seems fine! 👍

PR #9771, opened 2021-04-08
Enable complement tests for MSC2946.

### LINKED ISSUE #9756: Enable spaces summary (MSC2946) tests in complement
(Making this here, not the complement repo, since it probably requires some implementation changes?)

We should run the Spaces Summary tests in complement with Synapse. They do not currently pass, however.

Things to do:

* [ ] Enable spaces in the homeserver config.
* [ ] Enable the `msc2946` build tag for `complement.sh`
* [ ] Fix broken tests:
    * [ ] query whole graph
    * [ ] max_rooms_per_space
    * [ ] limit
    * [ ] redact link

Related to #9446

Note that the federation tests do pass, it is just a few of the client tests which fail.

clokep:
Fixes #9756 by providing the additional tag. This depends on matrix-org/complement#101 and will need a corresponding pipelines PR.

### DIFF
--- a/changelog.d/9771.misc
+++ b/changelog.d/9771.misc
@@ -1,0 +1,1 @@
+Enable Complement tests for [MSC2946](https://github.com/matrix-org/matrix-doc/pull/2946): Spaces Summary API.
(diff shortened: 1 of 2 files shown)


richvdh (approved):
LGTM

clokep:
I'm going to merge this, note that if you're running from a local complement checkout you'll need to ensure it is up-to-date.

PR #7709, opened 2020-06-16
Use mock from the stdlib.

clokep:
Per @anoadragon453's suggestion from looking at #7704 this switches from using the mock package to the one in the standard library.

Done with some sed magic + isort.

* Switches from importing `mock`, to importing it as a submodule of `unittest`.
* Stop installing the `mock` package.

This affects test only code so should be fine if unit tests pass.

PR #9772, opened 2021-04-08
Use mock from the stdlib.

clokep:
This is a redo of #7709, which I still had on my computer. Switches from using the mock package to the one in the standard library since we've now dropped Python 3.5.

Done with some sed magic + running isort.

* Switches from importing `mock`, to importing it as a submodule of `unittest`.
* Stop installing the `mock` package.
* Updates the isort config to consider `mock` a stdlib package.
* Runs isort.

This affects test only code so should be fine if unit tests pass. Should be reviewable commit-by-commit. I separated the automated steps from the steps where I manually changed things.

### DIFF
--- a/changelog.d/9772.misc
+++ b/changelog.d/9772.misc
@@ -1,0 +1,1 @@
+Use mock from the standard library instead of a separate package.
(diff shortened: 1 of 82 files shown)


anoadragon453 (approved):
lgtm!

PR #9773, opened 2021-04-08
Allow homeserver admins to remove appservice-initiated rooms from the public room directory

anoadragon453:
Homeserver admins can already delist rooms from the public room directory by calling [`/directory/list/room/{roomId}`](https://matrix.org/docs/spec/client_server/r0.6.1#put-matrix-client-r0-directory-list-room-roomid) with an admin access token.

However, application services can also list rooms, which will appear when supplying a `third_party_instance_id` or even just setting `include_all_networks: true` when calling [`POST /publicRooms`](https://matrix.org/docs/spec/client_server/r0.6.1#post-matrix-client-r0-publicrooms).

Application services can list and delist rooms via the [`PUT /directory/list/appservice/{networkId}/{roomId}`](https://matrix.org/docs/spec/application_service/r0.1.2#put-matrix-client-r0-directory-list-appservice-networkid-roomid) endpoint.

This PR allows homeserver admins to also delist appservice rooms via this endpoint. It also adds tests for that, as well as the happy paths of appservices (we didn't seem to have any) and the sad path of a normal user trying to call the endpoint.

### DIFF
--- a/synapse/storage/databases/main/room.py
+++ b/synapse/storage/databases/main/room.py
@@ -1396,8 +1396,6 @@ class RoomStore(RoomBackgroundUpdateStore, RoomWorkerStore, SearchStore):
                     txn,
                     table="appservice_room_list",
                     keyvalues={
-                        "appservice_id": appservice_id,
-                        "network_id": network_id,
                         "room_id": room_id,
                     },
                 )
--- a/synapse/storage/databases/main/room.py
+++ b/synapse/storage/databases/main/room.py
@@ -1359,65 +1359,58 @@ class RoomStore(RoomBackgroundUpdateStore, RoomWorkerStore, SearchStore):
         self.hs.get_notifier().on_new_replication_data()
 
     async def set_room_is_public_appservice(
-        self, room_id, appservice_id, network_id, is_public
-    ):
-        """Edit the appservice/network specific public room list.
+        self,
+        room_id: str,
+        appservice_id: Optional[str],
+        network_id: Optional[str],
+        is_public: bool,
+    ) -> None:
+        """Edit a room's visibility across one or more appservice/network-specific public
+        room lists.
 
         Each appservice can have a number of published room lists associated
         with them, keyed off of an appservice defined `network_id`, which
         basically represents a single instance of a bridge to a third party
         network.
 
+        Both the appservice and network ID are optional only when removing a room from the directory.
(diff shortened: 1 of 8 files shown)


synapse/storage/databases/main/room.py
```diff
@@ -1396,8 +1396,6 @@ def set_room_is_public_appservice_txn(txn, next_id):
                     txn,
                     table="appservice_room_list",
                     keyvalues={
-                        "appservice_id": appservice_id,
-                        "network_id": network_id,
```
erikjohnston, line 1400: Won't this mean that if we delete a room in one appservice list, it will delete them in all? While it's probably unexpected for a room to appear in multiple lists, its not wrong per se.

Should we instead be making the API require specifying the appservice ID for when done as an admin?

erikjohnston (changes requested):

synapse/storage/databases/main/room.py
anoadragon453: Indeed, it appears multiple application services can specify the same protocol (`network_id`).

The problem is that the admin may not initially know which network the room belongs to (and I'm not sure what the best way to query that information is).

I might advocate for the other direction and just state that if an admin uses the endpoint, the room is removed across all appservices that are part of that network.

erikjohnston: I was assuming that the client would know the network ID as it would just have used it to do the `/publicRooms` query, but it doesn't when getting "all rooms".

> I might advocate for the other direction and just state that if an admin uses the endpoint, the room is removed across all appservices that are part of that network.

If we're doing that then maybe we should always delete the room across all room lists?

anoadragon453: By "all room lists" you mean the client public room list as well, right?

Would it be better to just have an Admin API to delete a room across both client and appservice room lists?

Otherwise I'm not sure how this PR will work without different requirements depending on the requester (which feels weird).

erikjohnston: > Would it be better to just have an Admin API to delete a room across both client and appservice room lists?

Yeah, that might make more sense, and is probably what we want? 

Separately, its not unreasonable for someone to want to delete entries from a particular room list, so it would be good to not make an API that designs that out.

anoadragon453:
This is still on my backlog (and has been for a couple months now).

Closing until it can be actioned, or if someone else wants to pick it up.

PR #9776, opened 2021-04-09
Add timeout for resolve_service

### LINKED ISSUE #9774: `synapse.http.federation.srv_resolver.SrvResolver.resolve_service` is able to "timeout" and stall federation
Originally discovered on `#synapse:matrix.org` by @LTangaF

On Joel's server, doing the following DNS query times out;

```
root@5d0681f56cda:/# dig _matrix._tcp.matrix.lion.fm SRV

; <<>> DiG 9.11.5-P4-5.1+deb10u3-Debian <<>> _matrix._tcp.matrix.lion.fm SRV
;; global options: +cmd
;; connection timed out; no servers could be reached
```

While a valid SRV record doesn't time out;

```
root@5d0681f56cda:/# dig _matrix._tcp.jboi.nl SRV

; <<>> DiG 9.11.5-P4-5.1+deb10u3-Debian <<>> _matrix._tcp.jboi.nl SRV
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 560
;; flags: qr rd ra; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 4096
;; QUESTION SECTION:
;_matrix._tcp.jboi.nl.          IN      SRV

;; ANSWER SECTION:
_matrix._tcp.jboi.nl.   120     IN      SRV     0 0 443 matrix.jboi.nl.

;; Query time: 40 msec
;; SERVER: 127.0.0.11#53(127.0.0.11)
;; WHEN: Thu Apr 08 20:50:06 UTC 2021
;; MSG SIZE  rcvd: 83
```

This is already odd, but synapse currently *doesn't specify a timeout when looking up SRV records*.

The offending snippet is this:

https://github.com/matrix-org/synapse/blob/c619253db80c8d1c606dc40756dd3c9e3a55a9fb/synapse/http/federation/srv_resolver.py#L137-L139

When the underlying DNS query times out, this does never complete, and it causes a federation transmission loop to "time out" the whole request, putting it on catchup.

`twisted` has the following interface for `lookupService`:
```py
    def lookupService(name: str, timeout: Sequence[int]) -> "Deferred":
        """
        Perform an SRV record lookup.

        @param name: DNS name to resolve.
        @param timeout: Number of seconds after which to reissue the query.
            When the last timeout expires, the query is considered failed.

        @return: A L{Deferred} which fires with a three-tuple of lists of
            L{twisted.names.dns.RRHeader} instances.  The first element of the
            tuple gives answers.  The second element of the tuple gives
            authorities.  The third element of the tuple gives additional
            information.  The L{Deferred} may instead fail with one of the
            exceptions defined in L{twisted.names.error} or with
            C{NotImplementedError}.
        """
```

The optional parameter `timeout` defines that timeout, however, synapse isn't giving it any, so it never times out. Or synapse doesn't give it a *strict* enough timeout.

I propose adding a 15 second timeout by adding `timeout=(15,)` to the `SrvResolver.resolve_service` snippet.

ShadowJonathan:
Fixes #9774

Adds a stricter timeout to `resolve_service` that returns in 50s (smaller than the previous 60s, and the overarching 60s of timeout a federation transaction has)

### Pull Request Checklist

### DIFF
--- a/synapse/http/federation/srv_resolver.py
+++ b/synapse/http/federation/srv_resolver.py
@@ -21,6 +21,7 @@ from typing import List
 
 import attr
 
+from twisted.internet import defer
 from twisted.internet.error import ConnectError
 from twisted.names import client, dns
 from twisted.names.error import DNSNameError, DomainError
@@ -135,7 +136,7 @@ class SrvResolver:
 
         try:
             answers, _, _ = await make_deferred_yieldable(
-                self._dns_client.lookupService(service_name)
+                self._dns_client.lookupService(service_name, timeout=(1, 3, 11, 35))
             )
         except DNSNameError:
             # TODO: cache this. We can get the SOA out of the exception, and use
@@ -152,6 +153,10 @@ class SrvResolver:
                 return list(cache_entry)
             else:
                 raise e
+        except defer.TimeoutError as e:
+            raise defer.TimeoutError(
+                f"Could not resolve DNS for SRV record {service_name!r} due to timeout (50s total)"
+            ) from e
 
         if (
             len(answers) == 1
--- a/changelog.d/9776.bugfix
+++ b/changelog.d/9776.bugfix
@@ -1,0 +1,1 @@
+Clarify SRV record DNS timeouts. (#9774)
--- a/changelog.d/9776.bugfix
+++ b/changelog.d/9776.bugfix
@@ -1,1 +1,1 @@
-Clarify SRV record DNS timeouts. (#9774)
+Clarify SRV record DNS timeouts (#9774).


synapse/http/federation/srv_resolver.py
```diff
@@ -135,7 +136,7 @@ async def resolve_service(self, service_name: bytes) -> List[Server]:
 
         try:
             answers, _, _ = await make_deferred_yieldable(
-                self._dns_client.lookupService(service_name)
+                self._dns_client.lookupService(service_name, timeout=(1, 3, 11, 35))
```
ShadowJonathan, line 139: 1 + 3 + 11 + 35 = 50
---
By the way, i saw somewhere another federation-esc timeout of 30 seconds, shouldn't this timeout sum become 25 seconds? Or is 50 seconds acceptable?

synapse/http/federation/srv_resolver.py
```diff
@@ -152,6 +153,10 @@ async def resolve_service(self, service_name: bytes) -> List[Server]:
                 return list(cache_entry)
             else:
                 raise e
+        except defer.TimeoutError as e:
+            raise defer.TimeoutError(
```
ShadowJonathan, line 157: I'm not sure if this exception should raise "all the way" here via this, but at least it has different wording.

synapse/http/federation/srv_resolver.py
richvdh: so what's going on here? we're trying to change the timeout so that it's less than the total federation request timeout, so that we get a useful error message?

if so, it would be useful to say that in a comment.

(looks like the defaults are `(1, 3, 11, 45)`. could you include that in the comment too)

richvdh (changes requested):
^

richvdh:
also: CI is failing.

richvdh:
closing this as it seems to be on hold for now. We can reopen if you want to pick it up again.

Issue #2853, opened 2018-02-06
No preview for meduza.io articles


rkfg:
### Description

There's no preview image for any article from meduza.io news site.

### Steps to reproduce

- post an url like https://meduza.io/slides/vokrug-shum-rekordnyy-snegopad-v-moskve
- only the title and text is shown in the preview without the image

Server drops a line in the log:

2018-02-06 15:31:22,234 - synapse.rest.media.v1.preview_url_resource - 238 - WARNING - GET-12191- Couldn't get dims for https://meduza.io/imgly/share/1517844117/slides/vokrug-shum-rekordnyy-snegopad-v-moskve

The image itself is quite fine, it's a PNG of size 1200x630.

### Version information

- **Homeserver**: own homeserver
- **Version**:  Synapse/0.26.0
- **Install method**: package manager
- **Platform**: set up on Debian 9.3 amd64.

neilisfragile:
$ curl https://meduza.io/imgly/share/1517844117/slides/vokrug-shum-rekordnyy-snegopad-v-moskve > image
$ identify image
image JPEG 1200x630 1200x630+0+0 8-bit sRGB 126632B 0.000u 0:00.000

Not clear on first pass why dims cannot be extracted

rkfg:
So for this article above I have a preview image now. But this one's broken (as well as other more recent articles): https://meduza.io/feature/2021/04/07/proekt-rasskazal-o-vtoroy-zhene-ramzana-kadyrova-i-novoy-lichnosti-ego-pervoy-suprugi

The issue is that their server sends content-type as `image/jpeg; charset=utf-8` which is weird but not against the spec I guess. It can be fixed [in this file](https://github.com/matrix-org/synapse/blob/0d87c6bd121d043d5474cc20e352794ff8cf5625/synapse/rest/media/v1/preview_url_resource.py#L514) like this:
```diff
                 if b"Content-Type" in headers:
                     media_type = headers[b"Content-Type"][0].decode("ascii")
+                    scpos = media_type.find(';')
+                    if scpos > 0:
+                        media_type = media_type[:scpos]
                 else:
                     media_type = "application/octet-stream"
```

clokep:
@rkfg Any chance you'd be willing to put up a PR? `tests/rest/media/v1/test_url_preview.py` has some tests which could be updated too.

rkfg:
[...]

PR #9788, opened 2021-04-11
More robust handling of the Content-Type header for thumbnail generation

rkfg:
### Pull Request Checklist

This PR solves thumbnail generation in previews for some sites that use not quite standard Content-Type headers as described in #2853.

### DIFF
--- a/synapse/rest/media/v1/preview_url_resource.py
+++ b/synapse/rest/media/v1/preview_url_resource.py
@@ -513,10 +513,6 @@ class PreviewUrlResource(DirectServeJsonResource):
 
                 if b"Content-Type" in headers:
                     media_type = headers[b"Content-Type"][0].decode("ascii")
-                    if media_type.startswith("image/"):
-                        scpos = media_type.find(";")
-                        if scpos > 0:
-                            media_type = media_type[:scpos]
                 else:
                     media_type = "application/octet-stream"
 
--- a/synapse/rest/media/v1/preview_url_resource.py
+++ b/synapse/rest/media/v1/preview_url_resource.py
@@ -513,6 +513,10 @@ class PreviewUrlResource(DirectServeJsonResource):
 
                 if b"Content-Type" in headers:
                     media_type = headers[b"Content-Type"][0].decode("ascii")
+                    if media_type.startswith("image/"):
+                        scpos = media_type.find(";")
+                        if scpos > 0:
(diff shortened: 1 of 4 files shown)


rkfg:
Not sure why some tests failed but it doesn't seem to be related to this PR. Some concurrent update in monthly active users.

anoadragon453:
Hey @rkfg, thanks for your contribution! Yes, I'm afraid we're currently dealing with some flaky tests.

I can confirm that the failed tests don't appear to relate to your code change. :slightly_smiling_face:

synapse/rest/media/v1/preview_url_resource.py
```diff
@@ -513,6 +513,10 @@ async def _download_url(self, url: str, user: str) -> Dict[str, Any]:
 
                 if b"Content-Type" in headers:
                     media_type = headers[b"Content-Type"][0].decode("ascii")
+                    if media_type.startswith("image/"):
+                        scpos = media_type.find(";")
+                        if scpos > 0:
+                            media_type = media_type[:scpos]
```
richvdh, line 519: this looks wrong. should we not preserve any mime-type parameters supplied by the origin server?

rkfg: Are there valid cases for images? Currently it breaks getting thumbnail requirements because they're retrieved from `ContentRepositoryConfig` by MIME type and all those types don't include parameters like charset. It will be preserved for HTML because its MIME type doesn't start with `image/`

richvdh: > Are there valid cases for images?

I don't know, but it doesn't seem crazy that there could be?

>  Currently it breaks getting thumbnail requirements because they're retrieved from ContentRepositoryConfig by MIME type and all those types don't include parameters like charset

ok, but we can also fix that by stripping of the parameters before we index into `ContentRepositoryConfig`, rather than stripping off the parameters before we store them in the database.

richvdh: (which also means we don't have to special-case by `media_type.startswith("image/")`)

rkfg: Okay, I moved it to a more appropriate place.

richvdh (approved):
lgtm. Thanks!

PR #9796, opened 2021-04-12
Move some replication processing out of generic_worker

erikjohnston:
This is in preparation for moving presence off master.

### DIFF
--- a/synapse/app/generic_worker.py
+++ b/synapse/app/generic_worker.py
@@ -17,7 +17,7 @@
 import contextlib
 import logging
 import sys
-from typing import Dict, Iterable, Optional, Set
+from typing import Dict, Iterable, Optional
 
 from typing_extensions import ContextManager
 
@@ -81,14 +81,9 @@ from synapse.replication.slave.storage.transactions import SlavedTransactionStor
 from synapse.replication.tcp.client import ReplicationDataHandler
 from synapse.replication.tcp.commands import ClearUserSyncsCommand
 from synapse.replication.tcp.streams import (
-    AccountDataStream,
     DeviceListsStream,
-    GroupServerStream,
     PresenceStream,
-    PushersStream,
-    PushRulesStream,
     ReceiptsStream,
-    TagAccountDataStream,
     ToDeviceStream,
 )
 from synapse.rest.admin import register_servlets_for_media_repo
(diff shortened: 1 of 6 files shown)


synapse/app/generic_worker.py
```diff
@@ -693,72 +688,11 @@ async def _process_and_notify(self, stream_name, instance_name, token, rows):
                     stream_name, token, rows
                 )
 
-            if stream_name == PushRulesStream.NAME:
-                self.notifier.on_new_event(
```
richvdh, line 697: can remove `self.notifier` from constructor

synapse/replication/tcp/client.py
```diff
@@ -126,11 +148,52 @@ async def on_rdata(
...
+        elif stream_name == PushersStream.NAME:
+            for row in rows:
+                if row.deleted:
+                    self.stop_pusher(row.user_id, row.app_id, row.pushkey)
+                else:
+                    await self.start_pusher(row.user_id, row.app_id, row.pushkey)
+        elif stream_name == PresenceStream.NAME:
+            await self._presence_handler.process_replication_rows(token, rows)
 
         if stream_name == EventsStream.NAME:
```
richvdh, line 198:
```suggestion
        elif stream_name == EventsStream.NAME:
```

synapse/replication/tcp/client.py
```diff
@@ -197,9 +260,16 @@ async def on_position(self, stream_name: str, instance_name: str, token: int):
         # may be streaming.
         self.notifier.notify_replication()
 
+        await self.on_rdata(stream_name, instance_name, token, [])
```
richvdh, line 263: can we get rid of the call to `self.store.process_replication_rows` above, given that is also called by `on_rdata` ?

synapse/replication/tcp/client.py
```diff
@@ -131,6 +144,42 @@ async def on_rdata(
...
+        elif stream_name == GroupServerStream.NAME:
+            self.notifier.on_new_event(
+                "groups_key", token, users=[row.user_id for row in rows]
+            )
+        elif stream_name == PushersStream.NAME:
+            for row in rows:
+                if row.deleted:
+                    self.stop_pusher(row.user_id, row.app_id, row.pushkey)
+                else:
+                    await self.start_pusher(row.user_id, row.app_id, row.pushkey)
```
richvdh, line 194: this means that all this stuff is going to happen on the master, where it previously did not (iirc each replication client receives its own replication data back). Is that not going to be a problem for any of these?

synapse/handlers/presence.py
```diff
@@ -209,6 +219,242 @@ async def bump_presence_active_time(self, user: UserID):
...
+
+
+class _NullContextManager(ContextManager[None]):
+    """A context manager which does nothing."""
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        pass
+
+
+class WorkerPresenceHandler(BasePresenceHandler):
```
richvdh, line 261: at some point in this project (not necessarily during this particular PR), could you add some docstrings to `WorkerPresenceHandler` and `PresenceHandler` to explain how they differ?

synapse/handlers/presence.py
```diff
@@ -209,6 +219,242 @@ async def bump_presence_active_time(self, user: UserID):
...
+
+    async def update_external_syncs_clear(self, process_id):
+        """Marks all users that had been marked as syncing by a given process
+        as offline.
+
+        Used when the process has stopped/disappeared.
+
+        This is a no-op when presence is handled by a different worker.
+        """
+        pass
```
richvdh, line 247: I'm a bit confused about why these are being added. They don't seem to be called or defined anywhere in this diff?

synapse/handlers/presence.py
richvdh, line 261: I'm assuming that this is a complete lift-and-shift of `GenericWorkerPresence`. Is there anything else I should know?

richvdh (commented):
generally seems fine. A few nits. Convince me that this is safe to do on the master?

synapse/replication/tcp/client.py
erikjohnston: > iirc each replication client receives its own replication data back

We filter out the echoes before we get here, as `RDATA` includes the sending instance name, so that shouldn't be a problem. Really, I think its a bug that this hasn't been happening on master, e.g. if we split out receipts etc but keep `/sync` on master then `/sync` requests won't get told about new receipts.

synapse/handlers/presence.py
erikjohnston: *nods* And type hints.

synapse/handlers/presence.py
erikjohnston: Err, sorry.  A commit in a different PR needed them to be defined on both to appease mypy, I'll back that out.

synapse/handlers/presence.py
erikjohnston: Nope, this is just a straight copy and paste.

synapse/replication/tcp/client.py
erikjohnston: Good point

synapse/replication/tcp/client.py
richvdh: > We filter out the echoes before we get here, 

I went looking for that code, but couldn't find it. Can you point me to it?

synapse/handlers/presence.py
erikjohnston: Oh actually, they are needed now. This is because `hs.get_presence_handler()` returns `BasePresenceHandler` now instead of `PresenceHandler`. So either we a) add a no-op impl here, or b) assert the type in the calling code. I vaguely prefer the former to the latter.

FTR without these stub impls:

```
synapse/replication/tcp/client.py:196: error: "BasePresenceHandler" has no attribute "process_replication_rows"  [attr-defined]
synapse/replication/tcp/handler.py:355: error: "BasePresenceHandler" has no attribute "update_external_syncs_row"  [attr-defined]
synapse/replication/tcp/handler.py:365: error: "BasePresenceHandler" has no attribute "update_external_syncs_clear"  [attr-defined]
```

synapse/replication/tcp/client.py
erikjohnston: https://github.com/matrix-org/synapse/blob/develop/synapse/replication/tcp/handler.py#L401-L403 and https://github.com/matrix-org/synapse/blob/develop/synapse/replication/tcp/handler.py#L492-L494

synapse/replication/tcp/client.py
erikjohnston: Done. I've moved the `self.notifier.notify_replication()` after the `on_rdata(..)` call in case anything listening was requiring the store to have processed the replication.

richvdh (approved):
lgtm

PR #9798, opened 2021-04-12
Experimental Sync Speedup

### LINKED ISSUE #7524: Only send devicelist updates for users with whom you share an encrypted room?

ShadowJonathan:
This PR adds a speedup to `/sync` when the user is in many rooms with many members, the devicelist checking part of these sync requests can take up a while (1 second flat per sync) on single-user homeservers / lower-end setups.

This PR makes it so that `/sync` will only check for devicelist changes for users in encrypted rooms, thus this fixes #7524.

### Pull Request Checklist

### DIFF
--- a/synapse/storage/databases/main/roommember.py
+++ b/synapse/storage/databases/main/roommember.py
@@ -488,12 +488,15 @@ class RoomMemberWorkerStore(EventsWorkerStore):
 
     @cached(max_entries=500000, cache_context=True, iterable=True)
     async def get_users_who_share_room_with_user(
-        self, user_id: str, cache_context: _CacheContext
+        self, user_id: str, encrypted_only: bool, cache_context: _CacheContext
     ) -> Set[str]:
         """Returns the set of users who share a room with `user_id`"""
         room_ids = await self.get_rooms_for_user(
             user_id, on_invalidate=cache_context.invalidate
         )
+
+        if encrypted_only:
+            room_ids = await self._filter_encrypted_rooms(room_ids)
 
         user_who_share_room = set()
         for room_id in room_ids:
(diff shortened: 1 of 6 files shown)


ShadowJonathan:
I tested this by applying the changes onto the `release/v1.31.0` branch and building a docker image from that, it works, i didn't experience any immediate breakages.

<details>
<summary>Performance comparisons</summary>
image
(Using the main process for a second, unpatched with this change.)

image
(Utilizing the patched synchotron, note the difference in database calls)
</details>

ShadowJonathan:
There are some things wrong with this change, which need scrutiny and review;

- [ ] I'm fairly sure I did caching incorrectly, or that this change causes caching to *go* incorrectly.
- [ ] I faintly remember an edgecase I inwardly noticed regarding state-setting `m.room.encrypted` in a room. Between the `from_token` and `now_token`.
- [ ] This needs a good test.
- [ ] Element's devs would need to be asked on how they currently rely on the devicelist updates, for anything outside of encryption, to see if this change will not silently break that functionality.
- [ ] A devicelist update needs to be fired when the user joins an encrypted room, when every user in that room did not have a devicelist update in-between the local user getting invited, and joining.

ShadowJonathan:
Also, a lot of sytest tests seem to fail because of this change, because sytest expects the behaviour that devicelist changes are still propagated even if you dont share an encrypted room with that user.

I can't do much about these tests, since this change directly either invalidates them, or they invalidate this change, so I need some input on this change, and a direction to go forward with.

erikjohnston:
If we're only sending devices down for users we share encrypted rooms with we need to ensure that we proactively send down the device list for a user if either a) a room gets upgraded to an encrypted room or b) we join an encrypted room with the user. I'm not sure if that's the case already?

synapse/storage/databases/main/roommember.py
```diff
@@ -488,13 +488,16 @@ async def get_rooms_for_user(self, user_id: str, on_invalidate=None):
...
-        self, user_id: str, cache_context: _CacheContext
+        self, user_id: str, encrypted_only: bool, cache_context: _CacheContext
     ) -> Set[str]:
         """Returns the set of users who share a room with `user_id`"""
         room_ids = await self.get_rooms_for_user(
             user_id, on_invalidate=cache_context.invalidate
         )
 
+        if encrypted_only:
+            room_ids = await self._filter_encrypted_rooms(room_ids)
```
erikjohnston, line 499: We'd need to invalidate the cached result if the encryption state of the room change. I think we'd probably want to change `_filter_encrypted_rooms` to be a `get_encryption_state_for_rooms` and add a cache on that, so that we can `on_invalidate=cache_context.invalidate` when we call it, and so it'll "do the right thing".

ShadowJonathan:
> If we're only sending devices down for users we share encrypted rooms with we need to ensure that we proactively send down the device list for a user if either a) a room gets upgraded to an encrypted room or b) we join an encrypted room with the user. I'm not sure if that's the case already?

That's not the case with the current code, but i have that on [this](https://github.com/matrix-org/synapse/pull/9798#issuecomment-818012246) checklist.

richvdh:
I'm going to close this as @ShadowJonathan says he has abandoned it.

ShadowJonathan:
Yeah, I don't think that I'll be working on this any time soon, the PR was more a reference on how-to-do this with some todo points for y'all.

PR #9800, opened 2021-04-12
Separate creating an event context and persisting the event in the fed handler

clokep:
This is split out of #9763 as a pure refactoring PR with no behavior changes. It attempts to allow a caller to create the event context without side effects.

* The bits of `_prep_event` related to auth checking is moved into `do_auth`. Note that some of it is moved to the top of `do_auth` and some to the bottom, to keep the overall flow the same.
* The call to `do_auth` is then moved into `_handle_new_event` and `_handle_new_events`.
* This leaves `_prep_event` as just a call to `StateHandler.compute_event_context` which is inlined each place that `_prep_event` was called.
* Finally, `_handle_new_event` is modified to take the context in, so any place where `_handle_new_event` was called also gets a call to `StateHandler.compute_event_context` before it.
* This also performs some renames and adds some docstrings:
    * `do_auth` -> `_check_event_auth`
    * `_handle_new_event`(s) -> `_auth_and_persist_event`(s)

The commits in here kind of became a disaster, but hopefully the overall diff is readable.


synapse/handlers/federation.py
```diff
@@ -2414,6 +2415,17 @@ async def do_auth(
...
+            await self._check_for_soft_fail(event, state, backfilled)
+
+        if event.type == EventTypes.GuestAccess and not context.rejected:
+            await self.maybe_kick_guest_users(event)
+
+        # If we are going to send this event over federation we precaclculate
+        # the joined hosts.
+        if event.internal_metadata.get_send_on_behalf_of():
+            await self.event_creation_handler.cache_joined_hosts_for_event(event)
+
```
richvdh, line 2462: I don't think this is really part of auth... I was expecting you to move the *other* half of `prep_event` in here :/

richvdh (commented):
some initial thoughts

richvdh:
Hum, it's possible I've failed to understand what you're doing here, having only looked at the first couple of commits. If the above makes no sense, ignore it. I'll have another look at the complete diff.

synapse/handlers/federation.py
```diff
@@ -2178,14 +2233,31 @@ async def _persist_auth_tree(
...
+                server.
+
+                Also NB that this function adds entries to it.
+
+        Returns:
+             The event context and auth events.
+        """
         context = await self.state_handler.compute_event_context(event, old_state=state)
 
         if not auth_events:
```
richvdh, line 2263: I don't think this stuff (up to [line 2278](https://github.com/matrix-org/synapse/pull/9800/files#diff-0bc92da3d703202f5b9be2d3f845e375f5b1a6bc6ba61705a8af9be1121f5e42R2278)) belongs here. It is magic that prepares `auth_events` for the calls to `_update_auth_events_and_context_for_auth` and `event_auth.check` and as such, belongs next to those calls in `_check_event_auth`.

(this leaves you with `_calculate_event_context` just being the call to `state_handler.compute_event_context`, so you can basically just replace it with inline calls to `state_handler.compute_event_context`)

richvdh (changes requested):
Sorry, I've had another look at this. It makes more sense now, but I'm not sure that the split is quite right.

synapse/handlers/federation.py
clokep: Thanks! This cleans it up quite a bit. I've made this change and updated the description to be clearer what this is doing.

synapse/handlers/federation.py
```diff
@@ -2331,19 +2332,28 @@ async def on_get_missing_events(
...
-            event:
+            origin: The host the event originates from.
+            event: The event itself.
             context:
+                The event context.
+
+                NB that this function potentially modifies it.
+            state:
+                The state events used to auth the event. If this is not provided
+                the current state events will be used.
```
richvdh, line 2356:
```suggestion
                The state events used to check the event for soft-fail. If this is not provided
                the current state events will be used.
```

(similarly for `_auth_and_persist_event`).

synapse/handlers/federation.py
```diff
@@ -1990,16 +1999,47 @@ async def get_persisted_pdu(
...
+            auth_events:
+                Map from (event_type, state_key) to event
+
+                Normally, our calculated auth_events based on the state of the room
+                at the event's position in the DAG, though occasionally (eg if the
+                event is an outlier), may be the auth events claimed by the remote
+                server.
```
richvdh, line 2030: what does it mean if it's None?

synapse/handlers/federation.py
richvdh, line 2356: In an ideal world we'd fish this out of the `context`, but I'm not sure I feel confident enough in the call paths to do that without changing behaviour, so it might be best parked for now.

richvdh (approved):
generally looks like a good clean-up, I think.

This code is gnarly :/

synapse/handlers/federation.py
clokep: I had wondered that and decided that made me a bit nervous. 😢

Issue #9764, opened 2021-04-08
Docker on Windows fails to resolve localhost, causing heath check to fail


brendan-mccoy:
### Description

I'm playing around with the latest docker image, and it keeps throwing a message that it is in an unhealthy state. I checked and it seems that this is because the configured health check is "curl -fSs http://localhost:8008/health || exit 1" and the /etc/hosts file dosen't exist and so localhost isn't able to be resolved. At the time of this posting the image has sha256:d0bca6bb6f4009297601c06a8e3c22e193fde993290e969677b3b30e70bc3286

### Steps to reproduce

- docker pull matrixdotorg/synapse:latest
- docker run -it --rm --mount type=volume,src=synapse-data,dst=/data -e SYNAPSE_SERVER_NAME=my.matrix.host -e SYNAPSE_REPORT_STATS=yes matrixdotorg/synapse:latest generate
- docker run -d --name synapse --mount type=volume,src=synapse-data,dst=/data -p 8008:8008 matrixdotorg/synapse:latest
- wait for health check
- enter container shell, enter "cat /etc/hosts"

<!--
Describe how what happens differs from what you expected.

If you can identify any relevant log snippets from _homeserver.log_, please include
those (please be careful to remove any personal or private data). Please surround them with
``` (three backticks, on a line on their own), so that they are formatted legibly.
-->
The health check fails, "curl: (6) Could not resolve host: localhost", the file "/etc/hosts" doesn't exist causing this.

### Version information

<!-- IMPORTANT: please answer the following questions, to help us narrow down the problem -->

<!--
 What version of Synapse is running?

You can find the Synapse version with this command:

$ curl http://localhost:8008/_synapse/admin/v1/server_version

(You may need to replace `localhost:8008` if Synapse is not configured to
listen on that port.)
-->
- **Version**: {"server_version":"1.31.0","python_version":"3.8.9"}

- **Install method**: Docker 20.10.0
<!-- examples: package manager/git clone/pip  -->

- **Platform**: Windows Server 2019 Standard (Core)
<!--
Tell us about the environment in which your homeserver is operating
distro, hardware, if it's running in a vm/container, etc.
-->

**Comment by anoadragon453** [comment]
Thanks for the detailed reproduction steps! Interestingly, following them I didn't end up with the same result. `exec`ing into the container brought me:

```
root@61f2bb34918e:/# curl http://localhost:8008/health && echo
OK
root@61f2bb34918e:/# 
```

[...]

PR #9801, opened 2021-04-13
Add note to docker docs explaining platform support

anoadragon453:
Context is in https://github.com/matrix-org/synapse/issues/9764#issuecomment-818615894.

I struggled to find a more official link for this. The problem occurs when using WSL1 instead of WSL2, which some Windows platforms (at least Server 2019) still don't have. Docker have updated their documentation to paint a much happier picture now given WSL2's support.

The last sentence here can probably be removed once WSL1 is no longer around... though that will likely not be for a very long time.

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -4,9 +4,10 @@ This Docker image will run Synapse as a single process. By default it uses a
 sqlite database; for production use you should connect it to a separate
 postgres database. The image also does *not* provide a TURN server.
 
-Running this image is supported on all platforms that are supported by Docker upstream.
-Note that as such, Docker's experimental WS1-backend Linux Containers on Windows
-platform is [not supported](https://github.com/docker/for-win/issues/6470).
+This image should work on all platforms that are supported by Docker upstream.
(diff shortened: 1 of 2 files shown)


docker/README.md
```diff
@@ -2,9 +2,11 @@
 
 This Docker image will run Synapse as a single process. By default it uses a
 sqlite database; for production use you should connect it to a separate
-postgres database.
+postgres database. The image also does *not* provide a TURN server.
 
-The image also does *not* provide a TURN server.
+Running this image is supported on all platforms that are supported by Docker upstream.
+Note that as such, Docker's experimental WS1-backend Linux Containers on Windows
+platform is [not supported](https://github.com/docker/for-win/issues/6470).
```
richvdh, line 9:
```suggestion
This image should work on all platforms that are supported by Docker upstream.
Note that Docker's WS1-backend Linux Containers on Windows
platform is [experimental](https://github.com/docker/for-win/issues/6470) and
and not supported by this image.
```

richvdh (approved):
some suggestions for wording tweaks; lgtm anyway.

PR #9803, opened 2021-04-13
Hardened systemd unit files

savyajha:
### Pull Request Checklist

Signed-off-by: Savyasachee Jha <savya.jha@hawkradius.com>


debian/matrix-synapse.service
```diff
@@ -13,5 +13,76 @@ Restart=always
...
+# - sysctl settings and loaded kernel modules
+# - No modifications allowed to Control Groups
+# - Hostname
+# - System Clock
+ProtectSystem=strict
+ProtectKernelTunables=true
+ProtectKernelModules=true
+ProtectControlGroups=true
+ProtectClock=true
+ProtectHostname=true
```
anoadragon453, line 49:
```suggestion
ProtectHostname=true
ProtectClock=true
```

I would just reorder these to match the documentation ordering.

richvdh (dismissed):
I'm unconvinced we should do this. It looks super-scary, and I've never seen any other systemd units to all this stuff. This is supposed to be a simple example for people getting synapse set up for the first time!

It might be more appropriate in a separate `contrib` directory or something.

anoadragon453:
@richvdh After going through each option in systemd's manpages, they all make sense and I think the comments presented here help explain what each option is doing. That being said, you have a point that the systemd services we ship out of the box should probably be simple, so one can quickly read and edit it.

Perhaps moving these to contrib or elsewhere and hyperlinking to them as "hardened" configs from the default service files is the best way to go about it?

richvdh:
> Perhaps moving these to contrib or elsewhere and hyperlinking to them as "hardened" configs from the default service files is the best way to go about it?

yup, sounds good.

callahad:
@richvdh Is there a path to including at least some of these directly? Systemd sandboxing feels like a really great way to add defense in depth. Perhaps just targeting `debian/matrix-synapse.service` and leaving the `docs/systemd-with-workers/...` files more bare?

richvdh:
I don't object to it going into the unit file in the debian package (though I would question why debian doesn't do this for all services in packages in its repo, if it's such a great idea).

richvdh:
Also: I'm not keen on us having multiple copies of this stuff that we have to remember to maintain; we are literally *bound* to forget to update at least one copy when there are changes. Can we do something like move the config options out to a `.conf` file, which is then copied into the debian package? Systemd has some nice support for sharing config files between similarly-named units.

savyajha:
@richvdh It's not completely uncommon. I got the idea from looking at the service files of [postgres](https://github.com/archlinux/svntogit-packages/blob/packages/postgresql/trunk/postgresql.service) and [unbound](https://github.com/NLnetLabs/unbound/blob/e9810e1154277dc5174c6051b6bba33560809bed/contrib/unbound.service.in) as shipped by Arch Linux and upstream respectively. I think there was the principle of maintaining equivalence between unit files and traditional sysv scripts in Debian till recently which might have been a reason for them to not do this.

Regardless, do let me know what to do. Should I keep the Debian changes? Should I remove the ones in `doc/` or make separate files with the `-hardened` suffix?

anoadragon453:
systemd does have support for adding options to a `/etc/systemd/system/your-project.d/override.conf` file, which would then override options set in `your-project.service`. I don't know if it's good practice to bundle that file with your package (it's intended for overriding whatever options come from your distro's package) but we could:

* Include an `override-hardened.conf` file in the repo which users can copy to `/etc/systemd/system/matrix-synapse.d/override.conf` and packagers can use if they want increased security.
* Append the contents of `override-hardened.conf` to the service file when building the debian packages for security by default.

savyajha:
@anoadragon453 Typically it's considered bad practice for packagers to put files in `/etc/systemd/system/<service-name>.service.d/`, iirc, that location is meant for end-users.

I have no issues separating the file out into an `override-hardened.conf` (put in `contrib`, I suppose?). However, given I'm not a programmer and further have no experience packaging for Debian, I would hope that appending the hardened config to the Debian service file during package-building could be done by someone else.

richvdh:
> @anoadragon453 Typically it's considered bad practice for packagers to put files in /etc/systemd/system/<service-name>.service.d/, iirc, that location is meant for end-users.

Indeed. `/lib/systemd/system/<unit>.d` exists for packagers.

savyajha:
I've made the changes as requested. Does this work? If it's acceptable, hopefully someone can put the file from `contrib` as desired into the Debian package.

docs/systemd-with-workers/README.md
```diff
@@ -65,3 +65,37 @@ systemctl restart matrix-synapse-worker@federation_reader.service
...
 systemctl restart matrix-synapse.target
 ```
+
+## Hardening
+
+**Optional:** If further hardening is desired, the file
+`override-hardened.conf` may be copied from
+`contrib/systemd/override-hardened.conf` in this repository to the location
+`/etc/systemd/system/matrix-synapse.service.d/override-hardened.conf` (the
+directory will have to be created). It enables certain sandboxing features in
```
anoadragon453, line 75:
```suggestion
directory may have to be created). It enables certain sandboxing features in
```

Something may have created it already ¯\\\_(ツ)_/¯

docs/systemd-with-workers/README.md
```diff
@@ -65,3 +65,37 @@ systemctl restart matrix-synapse-worker@federation_reader.service
...
+systemd to further secure the synapse service. You may read the comments to
+understand what the override file is doing. A similar override file at
+`contrib/systemd-with-workers/override-hardened-worker.conf` may be copied to
+`/etc/systemd/system/matrix-synapse-worker@.service.d/override-hardened-worker.conf`
+(this directory will also have to be created) in order to apply the same
```
anoadragon453, line 80:
```suggestion
(this directory may also have to be created) in order to apply the same
```

changelog.d/9802.doc
```diff
@@ -0,0 +1 @@
+Hardened systemd files as proposed in Issue 9760 and added them to contrib. Changed the docs to reflect the presence of these files in contrib.
```
anoadragon453, line 1:
```suggestion
Add hardened systemd files as proposed in [#9760](https://github.com/matrix-org/synapse/issues/9760) and added them to `contrib/`. Change the docs to reflect the presence of these files.
```

Just some minor changes. Also some contributors put a "Contributed by @abc." at the end, which you can do if you want :slightly_smiling_face:

contrib/systemd-with-workers/override-hardened-workers.conf
```diff
@@ -0,0 +1,71 @@
+[Service]
```
anoadragon453, line 1: Both `override-hardened-workers.conf` and `override-hardened.conf` appear to be the same files. This could cause issues when we decide to update these files (someone could accidentally updated one without copying the changes to the other).

I'd advise keeping just one of the files and instructing users to copy it to multiple places - unless we have good justification to have different security-related options specifically for workers. Which I'm not sure we do currently.

docs/systemd-with-workers/README.md
```diff
@@ -65,3 +65,37 @@ systemctl restart matrix-synapse-worker@federation_reader.service
...
+hardening options to any worker processes.
+
+Once these files have been copied to their appropriate locations, simply run
+the following commands to apply the hardening options. They will automatically
```
anoadragon453, line 84:
```suggestion
Once these files have been copied to their appropriate locations, simply reload
systemd's manager config files and restart all Synapse services to apply the hardening options. They will automatically
```

docs/systemd-with-workers/README.md
```diff
@@ -65,3 +65,37 @@ systemctl restart matrix-synapse-worker@federation_reader.service
...
+
+```sh
+systemctl daemon-reload
+
+# Restart the master
+systemctl restart matrix-synapse.service
+
+# If using a worker-based setup, restart the workers too.
+# To restart a specific worker (eg. federation_reader):
+systemctl restart matrix-synapse-worker@federation_reader.service
```
anoadragon453, line 96: People should be able to simply `systemctl restart synapse.target` to restart the master and all configured worker processes, rather than restarting each one individually.

anoadragon453 (changes requested):
Apologies for the delay in getting back to this. I've got a few suggestions below, but overall think this looks pretty solid!

contrib/systemd-with-workers/override-hardened-workers.conf
savyajha: This makes a lot of sense. I've made this change.

docs/systemd-with-workers/README.md
savyajha: I'll be honest I didn't know this could be done even though I run an instance of Synapse myself! It seems one learns something new at every turn. I mostly copied stuff from the section above mine in the document, as you might have been able to tell.

Regardless I've made the suggested change.

savyajha:
My turn to apologise for the delay: the situation with covid is frankly somewhat alarming in my city. I've made the requisite changes and I hope it looks good. :)

docs/systemd-with-workers/README.md
```diff
@@ -65,3 +65,33 @@ systemctl restart matrix-synapse-worker@federation_reader.service
...
+
+## Hardening
+
+**Optional:** If further hardening is desired, the file
+`override-hardened.conf` may be copied from
+`contrib/systemd/override-hardened.conf` in this repository to the location
+`/etc/systemd/system/matrix-synapse.service.d/override-hardened.conf` (the
+directory may have to be created). It enables certain sandboxing features in
+systemd to further secure the synapse service. You may read the comments to
+understand what the override file is doing. The same file at may also be copied
```
anoadragon453, line 77:
```suggestion
understand what the override file is doing. The same file will need to be copied
```

docs/systemd-with-workers/README.md
```diff
@@ -65,3 +65,33 @@ systemctl restart matrix-synapse-worker@federation_reader.service
...
+
+Once these files have been copied to their appropriate locations, simply reload
+systemd's manager config files and restart all Synapse services to apply the hardening options. They will automatically
+be applied at every restart as long as the override files are present at the
+specified locations.
+
+```sh
+systemctl daemon-reload
+
+# Restart the services
```
anoadragon453, line 91:
```suggestion
# Restart services
```

anoadragon453 (approved):
No worries - people's safety should always comes first. I do wish you and your family well in that regard :slightly_smiling_face: 

These changes look good to me, thank you so much!

schildbach:
I wonder how this sandboxing solution compares to running Synapse in Docker?

PR #9814, opened 2021-04-14
Check for space membership during a remote join of a restricted room

### LINKED ISSUE #9715: Experimental support for checking space membership when a remote user joins a restricted room

clokep:
This is a re-do of #9763 since I merged that prematurely. Conveniently this also means it is rebased on #9800 / the history is cleaned up into a single commit.

When receiving a /send_join request for a room with join rules set to 'restricted', check if the user is a member of the spaces defined in the 'allow' key of the join rules.

This now also returns the auth error properly back to the client (instead of a generic 400 error).

This only applies to an experimental room version, as defined in MSC3083.

See matrix-org/complement#99 for integration tests on this.

Fixes #9715


synapse/handlers/room_member.py
```diff
@@ -178,62 +178,6 @@ async def ratelimit_invite(
 
         await self._invites_per_user_limiter.ratelimit(requester, invitee_user_id)
 
-    async def _can_join_without_invite(
-        self, state_ids: StateMap[str], room_version: RoomVersion, user_id: str
-    ) -> bool:
```
clokep, line 183: This is moved to a separate handler to avoid an import loop between the `RoomMemberHandler` and `FederationHandler`.

synapse/handlers/federation.py
```diff
@@ -2006,7 +2037,7 @@ async def _auth_and_persist_event(
         state: Optional[Iterable[EventBase]] = None,
         auth_events: Optional[MutableStateMap[EventBase]] = None,
         backfilled: bool = False,
-    ) -> EventContext:
+    ) -> None:
```
clokep, line 2040: The return value here was now never used.

synapse/handlers/event_auth.py
```diff
@@ -0,0 +1,82 @@
...
+    from synapse.server import HomeServer
+
+
+class EventAuthHandler:
+    def __init__(self, hs: "HomeServer"):
+        self._store = hs.get_datastore()
+
+    async def can_join_without_invite(
+        self, state_ids: StateMap[str], room_version: RoomVersion, user_id: str
+    ) -> bool:
```
clokep, line 30: This is identical to the original version, except changing `self.store` to `self._store`.

clokep:
One last piece of this, which was mentioned on the complement PR would be to return a better error code over federation when this happens. I'm not sure if that's really part of this PR or a follow-up though.

synapse/federation/federation_client.py
```diff
@@ -524,7 +525,17 @@ async def _try_destination_list(
...
+                # If an auth error comes back from a federation endpoint, pass
+                # it back to the client.
+                if e.code == 403:
+                    raise
+
+                logger.warning(
+                    "Failed to %s via %s", description, destination, exc_info=True
+                )
             except Exception:
+                logger.warning("POMP POMP POMP POMP POMP")
```
richvdh, line 538: ETOOMUCHPOMP

synapse/federation/federation_client.py
```diff
@@ -524,7 +525,17 @@ async def _try_destination_list(
...
                 )
+            except ProxiedRequestError as e:
+                # If an auth error comes back from a federation endpoint, pass
+                # it back to the client.
+                if e.code == 403:
+                    raise
+
+                logger.warning(
+                    "Failed to %s via %s", description, destination, exc_info=True
+                )
```
richvdh, line 536: how does this work? AFAICT the ProxiedRequestError is raised at line 518, but that won't be caught here.

synapse/handlers/event_auth.py
```diff
@@ -0,0 +1,82 @@
...
+
+from synapse.api.constants import EventTypes, JoinRules
+from synapse.api.room_versions import RoomVersion
+from synapse.types import StateMap
+
+if TYPE_CHECKING:
+    from synapse.server import HomeServer
+
+
+class EventAuthHandler:
```
richvdh, line 24: can you give this a docstring describing what sort of thing it is supposed to contain? You could probably also update the FIXME on `synapse.api.Auth` to say that the eventy things should move here.

synapse/handlers/event_auth.py
```diff
@@ -0,0 +1,82 @@
...
+            user_id: The user joining the room.
+
+        Returns:
+            True if the user can join the room, false otherwise.
+        """
+        # This only applies to room versions which support the new join rule.
+        if not room_version.msc3083_join_rules:
+            return True
+
+        # If there's no join rule, then it defaults to public (so this doesn't apply).
```
richvdh, line 49: sorry I know this is just moving, and I failed to spot it before, but this isn't correct: https://github.com/matrix-org/matrix-doc/issues/2660.

```suggestion
        # If there's no join rule, then it defaults to invite (so this doesn't apply).
```

synapse/handlers/federation.py
```diff
@@ -146,6 +146,7 @@ def __init__(self, hs: "HomeServer"):
         self.is_mine_id = hs.is_mine_id
         self.spam_checker = hs.get_spam_checker()
         self.event_creation_handler = hs.get_event_creation_handler()
+        self.event_auth_handler = hs.get_event_auth_handler()
```
richvdh, line 149:
```suggestion
        self._event_auth_handler = hs.get_event_auth_handler()
```

synapse/handlers/federation.py
```diff
@@ -1673,17 +1674,48 @@ async def on_send_join_request(self, origin: str, pdu: EventBase) -> JsonDict:
...
-        context = await self._auth_and_persist_event(origin, event, context)
+
+        # Get the state before the new event.
+        prev_state_ids = await context.get_prev_state_ids()
+
+        # Check if the user is already in the room or invited to the room.
+        user_id = event.state_key
+        prev_member_event_id = prev_state_ids.get((EventTypes.Member, user_id), None)
+        newly_joined = True
+        is_invite = False
```
richvdh, line 1687:
```suggestion
        user_is_invited = False
```

synapse/handlers/federation.py
```diff
@@ -1673,17 +1674,48 @@ async def on_send_join_request(self, origin: str, pdu: EventBase) -> JsonDict:
...
+                prev_state_ids,
+                event.room_version,
+                user_id,
+            )
+        ):
+            raise SynapseError(
+                403,
+                "You do not belong to any of the required spaces to join this room.",
+                Codes.FORBIDDEN,
+            )
```
richvdh, line 1708: For consistency with the C-S code:

```suggestion
            raise AuthError(
                403,
                "You do not belong to any of the required spaces to join this room.",
            )
```

richvdh (commented):
a few nits, generally lgtm

synapse/federation/federation_client.py
clokep: Oh, oops. 😢 Must have been trying to track some code...

synapse/handlers/federation.py
clokep: Oops I forgot to bring this over from the previous PR. Thanks!

synapse/federation/federation_client.py
clokep: Without this change, the traceback looks like:

<details>

```
2021-04-20 18:22:30,115 - synapse.federation.federation_client - 536 - WARNING - POST-1 - Failed to send_join via hs1
  File "/usr/local/lib/python3.8/site-packages/synapse/federation/federation_client.py", line 754, in _do_send_join
    return await self.transport_layer.send_join_v2(
  File "/usr/local/lib/python3.8/site-packages/synapse/federation/transport/client.py", line 256, in send_join_v2
    response = await self.client.put_json(
  File "/usr/local/lib/python3.8/site-packages/synapse/http/matrixfederationclient.py", line 714, in put_json
    response = await self._send_request_with_optional_trailing_slash(
  File "/usr/local/lib/python3.8/site-packages/synapse/http/matrixfederationclient.py", line 296, in _send_request_with_optional_trailing_slash
    response = await self._send_request(request, **send_request_args)
  File "/usr/local/lib/python3.8/site-packages/synapse/http/matrixfederationclient.py", line 542, in _send_request
    raise exc
synapse.api.errors.HttpResponseException: 403: Forbidden

During handling of the above exception, another exception occurred:

  File "/usr/local/lib/python3.8/site-packages/synapse/federation/federation_client.py", line 496, in _try_destination_list
    res = await callback(destination)
  File "/usr/local/lib/python3.8/site-packages/synapse/federation/federation_client.py", line 663, in send_request
    content = await self._do_send_join(destination, pdu)
  File "/usr/local/lib/python3.8/site-packages/synapse/federation/federation_client.py", line 770, in _do_send_join
    raise e.to_synapse_error()
synapse.api.errors.ProxiedRequestError: 403: You do not belong to any of the required spaces to join this room.
```
</details>

This is not catching the exception raised on line 518, but one raised on line 770, i.e. `_do_send_join` has already converting the `HttpResponseException` into a `ProxiedRequestError`. I'm now realizing that this seems specific to the `_do_send_join` use-case, and not all callers of `_try_destination_list`, which is unfortunate. I'm not really sure if `_do_send_join` is doing the right thing or not though. 😢

clokep: I took a look at this (and put up a change at [`clokep/federation-client-failover`](https://github.com/matrix-org/synapse/compare/develop...clokep/federation-client-failover), but I don't think that's right. It is a bit complicated because there's a few cases here:

* Some of the functions try to support both v1 and v2 endpoints, but only treat some v2 errors as "real" errors, while others are assumed to be non-existent v2 endpoints (and fallback to v1). (The "real" errors get raised as `ProxiedRequestError`.)
* There's a couple of spots where we seem to just raise `SynapseError` directly.
* There seems to be `InvalidResponseError` which should be used in some places instead.
* The only times that an `HttpResponseException` is a 500 error (or if the endpoint doesn't exist) -- this seems to match what I'd expect (the 403 would get passed back to the client properly), but this isn't happening due to it actually be the `ProxiedRequestError` getting raised.

I guess what I'm saying is that this is becoming complicated and I'd like to fix it separately than holding up this change for it. I'm having trouble coming up with the expected behavior for some of these functions though. It might just be to ensure we always return the original `HttpResponseException`, but I suspect that will cause situations where we previously failed over to no longer failover.

richvdh: I think the long and the short of it is that we need to decide whether the conversion from `HttpResponseException` happens at the `do_send_join` level or the `_try_destination_list` level. At the moment, both levels are trying to do it, which leaves us with a mess.

> There's a couple of spots where we seem to just raise SynapseError directly.
> There seems to be InvalidResponseError which should be used in some places instead.

I think the general logic here is that `_try_destination_list` should retry for `InvalidResponseError`, but not for `SynapseError`. But it's very unclear.

> The only times that an HttpResponseException is a 500 error

this sentence seems to be missing some words: I'm unclear what you're saying here.

> I guess what I'm saying is that this is becoming complicated and I'd like to fix it separately than holding up this change for it.

That's fair. I think if you want to take that approach, I'd remove this extra code from `_try_destination_list` for now. The whole "joins don't return clear errors" thing is an existing problem, and we can consider it orthogonal.

clokep: Thanks! For now I've backed out the changes to `FederationClient` and I'm going to work on them in a separate branch.

clokep: This seems fairly well documented in the docstring of `_try_destination_list`, TBH I think the callbacks are just doing the wrong thing. But still, will fix it separately.

clokep: I think #9868 is the fix for this.

richvdh (approved):
lgtm

PR #9815, opened 2021-04-14
remove `HomeServer.get_config`

richvdh:
Every single time I want to access the config object, I have to remember
whether or not we use `get_config`. Let's just get rid of it.


clokep (approved):
Seems reasonable to me!

PR #9816, opened 2021-04-14
Rename handler modules to drop _handler and config modules to drop _config.

clokep:
This is a silly PR that we talked about a while ago. I find it annoying that we have files like `saml_handler` in a module called `handlers`. This renames them. There's a bit of a faff in that some configs might refer to mapping providers in those modules, we just silently update the module name in that case.


synapse/config/oidc.py
```diff
@@ -27,7 +27,8 @@
 
 from ._base import Config, ConfigError, read_file
 
-DEFAULT_USER_MAPPING_PROVIDER = "synapse.handlers.oidc_handler.JinjaOidcMappingProvider"
+DEFAULT_USER_MAPPING_PROVIDER = "synapse.handlers.oidc.JinjaOidcMappingProvider"
+LEGACY_USER_MAPPING_PROVIDER = "synapse.handlers.oidc_handler.JinjaOidcMappingProvider"
```
erikjohnston, line 31: Can you add a quick comment to say why this is a thing, i.e. "we renamed the file..."

erikjohnston (approved):

synapse/config/oidc.py
clokep: Sure! 👍

clokep:
Apparently I failed to merge this. 🤷

Issue #1610, opened 2016-10-08
File upload size limit logic is broken (SYN-785)


matrixbot:
Submitted by @&#8203;matthew:matrix.org
It checks only content-length, which may lie. And it does so after the file has been uploaded, which is too late.

(Imported from https://matrix.org/jira/browse/SYN-785)

PC-Admin:
I have been experiencing this bug since i first set up synapse.

Value is set to: max_upload_size: "10M"

Interestingly a @*:matrix.org account can post up to 10MB files on a #*:perthchat.org rooms, yet a perthchat user cannot post more then a 1MB file without experiencing:

> The file ‘2m’ exceeds this home server’s size limit for uploads.

Any help fixing this issue is greatly appreciated, Matrix sucks when you can't upload files to it, and worse, the entire file tries to upload itself BEFORE telling you it cannot be done.

grafenhofer:
I confirm this bug. I followed https://www.digitalocean.com/community/tutorials/how-to-install-matrix-synapse-on-ubuntu-16-04. I even tried to increase max_upload_size to  "100M" - no success.

grafenhofer:
Crap, I should have done more research: it is an nginx "issue". The server section of the nginx site config should in include the following (or a similar) line:

    client_max_body_size 20M;

PC-Admin:
/etc/nginx/nginx.conf

then add the above line under "Basic Settings", works well

pabelenda:
Is there any short/mid term plan for this issue?

lampholder:
Is the fix for this:
1. check `content-length` at the start of upload, assume it's not lying, and fail fast if `content-length` > `max_upload_size`
2. check the actual uploaded content size after upload has completed and before persisting it to the media repo
3. give better feedback to the user when their upload fails due to either 1. or 2.

richvdh:
> check the actual uploaded content size after upload has completed and before persisting it to the media repo

better to bail out during the upload rather than waiting for the upload to fill our disk

> give better feedback to the user when their upload fails due to either 1. or 2.

the problem is that it's hard to give good feedback over the API without accepting the entire body (which may take hours for a large file). Other problems include the fact that there may be things other than the synapse upload size limit (notably, the reverse-proxy's limit) which cause a failure, and I don't know how they behave.

lampholder:
> better to bail out during the upload rather than waiting for the upload to fill our disk

Makes sense

[...]

PR #9817, opened 2021-04-14
Improved validation for received requests

richvdh:
There's a couple of commits here. Hopefully each one is clear enough.

Fixes: #1610

### DIFF
--- a/synapse/api/constants.py
+++ b/synapse/api/constants.py
@@ -17,6 +17,9 @@
 
 """Contains constants from the specification."""
 
+# the max size of a (canonical-json-encoded) event
+MAX_PDU_SIZE = 65536
+
 # the "depth" field on events is limited to 2**63 - 1
 MAX_DEPTH = 2 ** 63 - 1
 
--- a/synapse/app/_base.py
+++ b/synapse/app/_base.py
@@ -30,8 +30,10 @@ from twisted.internet import defer, error, reactor
 from twisted.protocols.tls import TLSMemoryBIOFactory
 
 import synapse
+from synapse.api.constants import MAX_PDU_SIZE
 from synapse.app import check_bind_error
 from synapse.app.phone_stats_home import start_phone_stats_home
+from synapse.config.homeserver import HomeServerConfig
 from synapse.crypto import context_factory
 from synapse.logging.context import PreserveLoggingContext
 from synapse.metrics.background_process_metrics import wrap_as_background_process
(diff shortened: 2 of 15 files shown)


richvdh:
oh, hrmph, I need to check the max media upload size

synapse/app/_base.py
```diff
@@ -528,3 +530,25 @@ def sdnotify(state):
         # this is a bit surprising, since we don't expect to have a NOTIFY_SOCKET
         # unless systemd is expecting us to notify it.
         logger.warning("Unable to send notification to systemd: %s", e)
+
+
+def max_request_body_size(config: HomeServerConfig) -> int:
```
richvdh, line 535: I'm not terribly happy with this, but couldn't think of anywhere better to do it.

synapse/config/logger.py
```diff
@@ -318,6 +317,8 @@ def setup_logging(
     # Perform one-time logging configuration.
     _setup_stdlib_logging(config, log_config_path, logBeginner=logBeginner)
     # Add a SIGHUP handler to reload the logging configuration, if one is available.
+    from synapse.app import _base as appbase
+
```
richvdh, line 321: this is needed to avoid a circular import. `config` should not be depending on `app`.

synapse/api/constants.py
```diff
@@ -17,6 +17,9 @@
 
 """Contains constants from the specification."""
 
+# the max size of a (canonical-json-encoded) event
+MAX_PDU_SIZE = 65536
```
clokep, line 21: Below the comment says the max size is 65535, not 65536, doesn't really matter but is a bit confusing. 😄

richvdh: fixed the comment, thanks.

synapse/http/site.py
```diff
@@ -59,8 +61,9 @@ class SynapseRequest(Request):
         logcontext: the log context for this request
     """
 
-    def __init__(self, channel, *args, **kw):
+    def __init__(self, channel, *args, max_request_body_size=1024, **kw):
```
erikjohnston, line 64: This seems like a surprisingly low default? Can we not make it a required field?

synapse/server.py
```diff
@@ -286,6 +286,10 @@ def setup(self) -> None:
         if self.config.run_background_tasks:
             self.setup_background_tasks()
 
+    def start_listening(self) -> None:
+        """Start the HTTP, manhole, metrics, etc listeners"""
+        pass
```
erikjohnston, line 291: I find it a bit odd that the default implementation is a no-op, given the doc comment. What do you think about making this an abstract method and keeping the no-op version in admin cmd?

synapse/http/site.py
richvdh: the default was mostly to avoid messing around with the tests more than I had to. Happy to make it required if you prefer.

synapse/server.py
richvdh: personally, it seems fairly reasonable that the default be a no-op - I basically see it as a hook that the derived classes can override. I could make it clearer in the docstring?

don't feel that strongly though, so let me know what you think.

erikjohnston: I guess it just feels slightly odd that `start_listening()` doesn't do the right thing by default (which it feels like it could if we deduplicate the code between master and worker apps), the fact that the admin cmd server is no-oping it is a special thing that its doing that I'd expect to be specifically called out in its definition.

It doesn't really matter though, given we're not really adding new server objects.

erikjohnston (approved):

richvdh:
I suspect the test failures here are real.

richvdh:
I've rebased this on top of #9817, both to simplify the patches, and in the hope of making the tests pass.

richvdh:
FINALLY. The only change since erik's :heavy_check_mark: are https://github.com/matrix-org/synapse/pull/9817/commits/9849d60c8e825a81690242179a155fc2f1d6177f and a minor docstring tweak, so I'm going to take his approval and run with it.

PR #9819, opened 2021-04-15
Add presence federation stream

erikjohnston:
This is to allow moving presence off master. It turned out that making the federation send queue stuff work on arbitrary workers was fiddly (since not all workers have HTTP replication listeners), and so we replace it with a custom stream just for sending ad hoc presence states over federation. 

The new stream is only used for sending out federation pings and when a new remote host joins a room.  Normal presence updates are replicated via the existing presence stream. We just stream which user's presence we should send to which destinations, rather than the full presence states (as the remotes should already have that data via the presence stream).

Currently the presence federation stream always stores the last 5 minutes of data in memory, so that other workers can query recent data if they restart. The idea here is that a) this is just ephemeral data anyway, so safe to drop and b) there shouldn't be that much data going through anyway. It shouldn't be hard to add back in a periodic ACK to allow us to clear out the data.

The federation send queue is now unused, and a future PR will remove it (I can add it to this PR if that is preferable, I'm just aware that its already large).

Reviewable commit by commit.

---

Update: This is now based on #9828


synapse/federation/send_queue.py
```diff
@@ -559,9 +542,6 @@ def process_rows_for_federation(
         parsed_row = RowType.from_data(row.data)
         parsed_row.add_to_buffer(buff)
 
-    if buff.presence:
-        transaction_queue.send_presence(buff.presence)
```
richvdh, line 563: looks like we can remove `presence` from `ParsedFederationStreamData` ?

synapse/handlers/presence.py
```diff
@@ -245,9 +250,17 @@ async def update_external_syncs_clear(self, process_id):
         """
         pass
 
-    async def process_replication_rows(self, token, rows):
+    async def process_replication_rows(
+        self, stream_name: str, instance_name: str, token: int, rows: list
+    ):
         """Process presence stream rows received over replication."""
```
richvdh, line 256: I think this docstring is now incorrect (it gets all replication rows, not just presence)?

richvdh (changes requested):

synapse/handlers/presence.py
```diff
@@ -408,6 +433,20 @@ async def process_replication_rows(self, token, rows):
...
+
+        hosts_and_states = await get_interested_remotes(
+            self.store,
+            self.presence_router,
+            states,
+            self.state,
+        )
+
+        for destinations, states in hosts_and_states:
+            self._federation.send_presence_to_destinations(states, destinations)
```
richvdh, line 448: I can't help but feel that having four separate instances of the `get_interested_remotes; send_presence_to_destinations` pattern suggests that we should do something slightly different.

is there any milage in replacing `get_interested_remotes` with a `send_presence_to_interested_destinations`?

erikjohnston:
I've split out a bunch of stuff into #9828, which this PR is now based on.

synapse/handlers/presence.py
```diff
@@ -1859,3 +1868,185 @@ async def get_interested_remotes(
...
+
+        # Should we keep a queue of recent presence updates? We only bother if
+        # another process may be handling federation sending.
+        self._queue_presence_updates = True
+
+        # Whether this instance is a presence writer.
+        self._presence_writer = hs.config.worker.worker_app is None
+
+        # The federation sender if this instance is a federation sender.
+        self._federation = None
```
ShadowJonathan, line 1905: Why is it called `_federation` if its a federation *sender*?

Small nitpick, resolve at will, I'm just pointing at the naming

ShadowJonathan: Ah wait, i see, it is an interface to send to federation *through*

What is the actual type of this variable, though? Is there an interface type that describes this behaviour? (I'd like it to be annotated with it)

synapse/handlers/presence.py
```diff
@@ -724,12 +739,10 @@ async def _update_states(self, new_states: Iterable[UserPresenceState]) -> None:
...
 
-                # Since this is master we know that we have a federation sender or
-                # queue, and so this will be defined.
-                assert self._federation
-
                 for destinations, states in hosts_and_states:
-                    self._federation.send_presence_to_destinations(states, destinations)
+                    self._federation_queue.send_presence_to_destinations(
+                        states, destinations
+                    )
```
erikjohnston, line 745: @richvdh To continue the conversation from https://github.com/matrix-org/synapse/pull/9828#discussion_r614987677, this is a similar shape as in `maybe_send_presence_to_interested_destinations`, except it uses the federation queue instead of the federation sender. I don't know whether it is worth trying to merge them, but I think that gets a bit convoluted?

synapse/handlers/presence.py
```diff
@@ -127,10 +130,10 @@ def __init__(self, hs: "HomeServer"):
         self.state = hs.get_state_handler()
 
         self._federation = None
-        if hs.should_send_federation() or not hs.config.worker_app:
+        if hs.should_send_federation():
             self._federation = hs.get_federation_sender()
```
erikjohnston, line 134: When we remove the federation send queue I plan on making `hs.get_federation_sender()` return an optional federation sender, which will simplify this even more.

synapse/handlers/presence.py
```diff
@@ -1859,3 +1868,185 @@ async def get_interested_remotes(
...
+        self._federation = None
+
+        if hs.should_send_federation():
+            self._federation = hs.get_federation_sender()
+
+            # We don't bother queuing up presence states if only this instance
+            # is sending federation.
+            if hs.config.worker.federation_shard_config.instances == [
+                self._instance_name
+            ]:
```
richvdh, line 1914: does this work in a monolith? (I didn't think `hs.config.worker` existed, but icbw)

synapse/handlers/presence.py
```diff
@@ -1859,3 +1868,185 @@ async def get_interested_remotes(
...
+        upto_token: int,
+        target_row_count: int,
+    ) -> Tuple[List[Tuple[int, Tuple[str, str]]], int, bool]:
+        """Get all the updates between the two tokens.
+
+        We return rows in the form of `(destination, user_id)` to keep the size
+        of each row bounded (rather than returning the sets in a row).
+        """
+        if instance_name != self._instance_name:
+            # If not local we query over replication.
```
richvdh, line 1991:
```suggestion
            # If not local we query over http replication from the master
```

aiui?

TIL this was a thing

synapse/handlers/presence.py
```diff
@@ -1859,3 +1868,185 @@ async def get_interested_remotes(
...
+        #
+        # Since the start of the queue is periodically truncated we need to
+        # handle the case where `from_token` stream ID has already been dropped.
+        start_idx = max(from_token - self._next_id, -len(self._queue))
+
+        to_send = []  # type: List[Tuple[int, Tuple[str, str]]]
+        limited = False
+        new_id = upto_token
+        for _, stream_id, destinations, user_ids in self._queue[start_idx:]:
+            if stream_id > upto_token:
```
richvdh, line 2012: isn't `upto_token` meant to be *exclusive* ? 
```suggestion
            if stream_id >= upto_token:
```

synapse/handlers/presence.py
```diff
@@ -1859,3 +1868,185 @@ async def get_interested_remotes(
...
+        return to_send, new_id, limited
+
+    async def process_replication_rows(
+        self, stream_name: str, instance_name: str, token: int, rows: list
+    ):
+        if stream_name != PresenceFederationStream.NAME:
+            return
+
+        # We keep track of the current tokens
+        self._current_tokens[instance_name] = token
```
richvdh, line 2036: can't see this being used anywhere (yet?)

synapse/replication/tcp/streams/_base.py
```diff
@@ -290,6 +290,30 @@ def __init__(self, hs):
...
 
 
+class PresenceFederationStream(Stream):
+    """A stream used to send ad hoc presence updates over federation.
+
+    Streams the remote destination and the user ID of the presence state to
+    send.
+    """
+
+    @attr.s(slots=True, auto_attribs=True)
```
richvdh, line 300: ooooOOOOoooo pep526!

tests/handlers/test_presence.py
```diff
@@ -482,10 +645,17 @@ class PresenceJoinTestCase(unittest.HomeserverTestCase):
...
         hs = self.setup_test_homeserver(
-            "server", federation_http_client=None, federation_sender=Mock()
+            "server",
+            federation_http_client=None,
+            federation_sender=Mock(spec=FederationSender),
         )
         return hs
 
+    def default_config(self):
+        config = unittest.default_config("test")
```
richvdh, line 655:
```suggestion
        config = super().default_config()
```

richvdh (commented):
seems... plausible.

synapse/handlers/presence.py
```diff
@@ -253,9 +256,17 @@ async def update_external_syncs_clear(self, process_id):
...
+    async def process_replication_rows(
+        self, stream_name: str, instance_name: str, token: int, rows: list
+    ):
+        """Process streams received over replication."""
+        await self._federation_queue.process_replication_rows(
+            stream_name, instance_name, token, rows
+        )
+
+    def get_federation_queue(self) -> "PresenceFederationQueue":
+        """Get the presence federation queue, if any."""
```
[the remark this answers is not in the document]
erikjohnston: Woops, this used to be only instantiated on master, but that is no longer true. Will fix comment.

synapse/handlers/presence.py
```diff
@@ -1859,3 +1868,185 @@ async def get_interested_remotes(
...
+        self._instance_name = hs.get_instance_name()
+        self._presence_handler = presence_handler
+        self._repl_client = ReplicationGetStreamUpdates.make_client(hs)
+
+        # Should we keep a queue of recent presence updates? We only bother if
+        # another process may be handling federation sending.
+        self._queue_presence_updates = True
+
+        # Whether this instance is a presence writer.
+        self._presence_writer = hs.config.worker.worker_app is None
```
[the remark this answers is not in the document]
erikjohnston: Yup

synapse/handlers/presence.py
```diff
@@ -127,10 +130,10 @@ def __init__(self, hs: "HomeServer"):
         self.state = hs.get_state_handler()
 
         self._federation = None
-        if hs.should_send_federation() or not hs.config.worker_app:
+        if hs.should_send_federation():
             self._federation = hs.get_federation_sender()
 
-        self._send_federation = hs.should_send_federation()
+        self._federation_queue = PresenceFederationQueue(hs, self)
```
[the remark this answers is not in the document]
erikjohnston: Mainly just to bundle up all the logic into one place TBH. (In the first PoC we only instantiated this on the presence writer process, but that changed).

synapse/handlers/presence.py
erikjohnston: It does yeah, all the options exist and default to sane things. (All these paths are tested by the sytests fwiw, and I double checked that this does come to the right conclusion on monoliths)

synapse/handlers/presence.py
erikjohnston: Yeah, the typing and federation streams do this too.

synapse/handlers/presence.py
erikjohnston: It gets used in `def get_current_token`?

synapse/replication/tcp/streams/_base.py
erikjohnston: Oh yeah, I forgot I played with this. It does have a small foot gun where if you naively do `param: Dict[..] = {}` to have a default arg it reuses the same dict (in the same way as this is a foot gun for optional params).

synapse/handlers/presence.py
erikjohnston: I think the common pattern is that the lower bound is exclusive and the top bound is inclusive? At least the typing, events and presence streams all seem to work like that, e.g.:

https://github.com/matrix-org/synapse/blob/4b965c862dc66c0da5d3240add70e9b5f0aa720b/synapse/storage/databases/main/events_worker.py#L1142

synapse/handlers/presence.py
```diff
@@ -1859,3 +1868,185 @@ async def get_interested_remotes(
...
+
+class PresenceFederationQueue:
+    """Handles sending ad hoc presence updates over federation, which are *not*
+    due to state updates (that get handled via the presence stream), e.g.
+    federation pings and sending existing present states to newly joined hosts.
+
+    Only the last N minutes will be queued, so if a federation sender instance
+    is down for longer then some updates will be dropped. This is OK as presence
+    is ephemeral, and so it will self correct eventually.
+    """
```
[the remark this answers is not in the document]
erikjohnston: I've added some words

synapse/handlers/presence.py
```diff
@@ -1970,6 +1976,11 @@ def send_presence_to_destinations(
         self._notifier.notify_replication()
 
     def get_current_token(self, instance_name: str) -> int:
+        """Get the current position of the stream.
+
+        On workers this returns the last stream ID received from replication.
+        """
+
```
richvdh, line 1983:
```suggestion
```

synapse/handlers/presence.py
richvdh: yeah, I think you're right, on closer inspection.

synapse/handlers/presence.py
richvdh: oh, apparently I'm blind.

synapse/handlers/presence.py
```diff
@@ -1859,3 +1868,198 @@ async def get_interested_remotes(
...
+        return to_send, new_id, limited
+
+    async def process_replication_rows(
+        self, stream_name: str, instance_name: str, token: int, rows: list
+    ):
+        if stream_name != PresenceFederationStream.NAME:
+            return
+
+        # We keep track of the current tokens
```
richvdh, line 2048:
```suggestion
        # We keep track of the current tokens (so that we can catch up with anything we missed after a disconnect)
```

richvdh (approved):
lgtm!

PR #9825, opened 2021-04-15
Small speed up joining large remote rooms

erikjohnston:
There are a couple of points in `persist_events` where we are doing a query per event in series, which we can replace.

### DIFF
--- a/changelog.d/9825.misc
+++ b/changelog.d/9825.misc
@@ -1,1 +1,1 @@
-Small speed up joining large remote rooms.
+Small speed up for joining large remote rooms.
--- a/changelog.d/9825.misc
+++ b/changelog.d/9825.misc
@@ -1,0 +1,1 @@
+Small speed up joining large remote rooms.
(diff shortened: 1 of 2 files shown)


changelog.d/9825.misc
```diff
@@ -0,0 +1 @@
+Small speed up joining large remote rooms.
```
richvdh, line 1:
```suggestion
Small speed up for joining large remote rooms.
```

synapse/storage/databases/main/events.py
```diff
@@ -1881,20 +1885,26 @@ def _set_push_actions_for_event_and_users_txn(
...
+                    iterable=event_ids,
+                    keyvalues={},
+                    retcols=("user_id",),
                 )
 
+                for row in rows:
+                    txn.call_after(
+                        self.store.get_unread_event_push_actions_by_room_for_user.invalidate_many,
+                        (room_id, row["user_id"]),
+                    )
```
richvdh, line 1906: would it be worth deduping the user_ids, either with a `select distinct` or with a Python `set`? presumably there could be quite a few duplicates?

erikjohnston: Yeah, good point actually.

richvdh (approved):

PR #9827, opened 2021-04-15
Add a unique index on `federation_stream_position`

### LINKED ISSUE #4023: federation_stream_position table should have a unique index
The symptoms of this bug are:

* People on other servers cannot read your messages
* Typing notifications (and presence) are sent correctly
* Synapse logs contain the following exception:

```
CRITICAL - - Unhandled error in Deferred:
CRITICAL - - 
  File "env/lib/python2.7/site-packages/twisted/internet/defer.py", line 1416, in _inlineCallbacks
    result = result.throwExceptionIntoGenerator(g)
  File "env/lib/python2.7/site-packages/twisted/python/failure.py", line 491, in throwExceptionIntoGenerator
    return g.throw(self.type, self.value, self.tb)
  File "env/lib/python2.7/site-packages/synapse/metrics/background_process_metrics.py", line 193, in run
    yield func(*args, **kwargs)
  File "env/lib/python2.7/site-packages/twisted/internet/defer.py", line 1416, in _inlineCallbacks
    result = result.throwExceptionIntoGenerator(g)
  File "env/lib/python2.7/site-packages/twisted/python/failure.py", line 491, in throwExceptionIntoGenerator
    return g.throw(self.type, self.value, self.tb)
  File "env/lib/python2.7/site-packages/synapse/federation/transaction_queue.py", line 247, in _process_event_queue_loop
    "events", next_token
  File "env/lib/python2.7/site-packages/twisted/internet/defer.py", line 1416, in _inlineCallbacks
    result = result.throwExceptionIntoGenerator(g)
  File "env/lib/python2.7/site-packages/twisted/python/failure.py", line 491, in throwExceptionIntoGenerator
    return g.throw(self.type, self.value, self.tb)
  File "env/lib/python2.7/site-packages/synapse/storage/_base.py", line 325, in runInteraction
    *args, **kwargs
  File "env/lib/python2.7/site-packages/twisted/internet/defer.py", line 1416, in _inlineCallbacks
    result = result.throwExceptionIntoGenerator(g)
  File "env/lib/python2.7/site-packages/twisted/python/failure.py", line 491, in throwExceptionIntoGenerator
    return g.throw(self.type, self.value, self.tb)
  File "env/lib/python2.7/site-packages/synapse/storage/_base.py", line 374, in runWithConnection
    inner_func, *args, **kwargs
  File "env/lib/python2.7/site-packages/twisted/python/threadpool.py", line 250, in inContext
    result = inContext.theWork()
  File "env/lib/python2.7/site-packages/twisted/python/threadpool.py", line 266, in <lambda>
    inContext.theWork = lambda: context.call(ctx, func, *args, **kw)
  File "env/lib/python2.7/site-packages/twisted/python/context.py", line 122, in callWithContext
  File "env/lib/python2.7/site-packages/twisted/python/context.py", line 85, in callWithContext
    return func(*args,**kw)
  File "env/lib/python2.7/site-packages/twisted/enterprise/adbapi.py", line 306, in _runWithConnection
  File "env/lib/python2.7/site-packages/twisted/enterprise/adbapi.py", line 297, in _runWithConnection
  File "env/lib/python2.7/site-packages/synapse/storage/_base.py", line 370, in inner_func
    return func(conn, *args, **kwargs)
  File "env/lib/python2.7/site-packages/synapse/storage/_base.py", line 246, in _new_transaction
    r = func(txn, *args, **kwargs)
  File "env/lib/python2.7/site-packages/synapse/storage/_base.py", line 854, in _simple_update_one_txn
    raise StoreError(500, "More than one row matched")
StoreError: 500: More than one row matched
```

This has now been reported by two users, @TheForcer at https://github.com/matrix-org/synapse/issues/3750#issuecomment-424516412, and @SylvainCecchetto at #4018.

The basic problem is that `federation_stream_position` contains too many rows, for example:

```
sqlite> select * from federation_stream_position;
type        stream_id 
----------  ----------
federation  -1        
events      -1    
federation  -1        
events      -1    
federation  -1        
events      13943    
```

I have yet to understand how this happens. If anybody else sees it, it would be very helpful to get Synapse logs from the point at which the `More than one row matched` error starts.

djmitche:
### Pull Request Checklist

Fixes #4023.

### DIFF
--- a/synapse/storage/databases/main/schema/delta/60/federation_stream_position_unique_idx.sql
+++ b/synapse/storage/databases/main/schema/delta/60/federation_stream_position_unique_idx.sql
@@ -1,0 +1,4 @@
+-- Ensure that only one position can be tracked for each stream type.  This might
+-- be violated if, for exapmle, a backup is restored without first clearing the
+-- table's contents.
+CREATE INDEX federation_stream_position_idx ON federation_stream_position(type);
--- a/changelog.d/9827.bugfix
+++ b/changelog.d/9827.bugfix
@@ -1,0 +1,1 @@
+The `federation_stream_position` table now has a unique index, prohibiting multiple rows for the same type.  This might occur due to restores from backup without first truncating tables.
--- a/synapse/storage/prepare_database.py
+++ b/synapse/storage/prepare_database.py
@@ -34,7 +34,7 @@ logger = logging.getLogger(__name__)
 
 # Remember to update this number every time a change is made to database
 # schema files, so the users will be informed on server restarts.
-SCHEMA_VERSION = 59
+SCHEMA_VERSION = 60
 
 dir_path = os.path.abspath(os.path.dirname(__file__))
 


synapse/storage/databases/main/schema/delta/60/federation_stream_position_unique_idx.sql
```diff
@@ -0,0 +1,4 @@
+-- Ensure that only one position can be tracked for each stream type.  This might
+-- be violated if, for exapmle, a backup is restored without first clearing the
+-- table's contents.
+CREATE INDEX federation_stream_position_idx ON federation_stream_position(type);
```
erikjohnston, line 4:
```suggestion
CREATE UNIQUE INDEX federation_stream_position_idx ON federation_stream_position(type, instance_name);
```

For sharded federation senders we expect to have multiple rows, but with different instance_names

erikjohnston, line 4: We should also clear out any duplicates that exist, otherwise this will fail. We should be able to add a delete before the index creation that deletes duplicate rows (we want to keep the row with the min stream position).

erikjohnston (commented):
Thanks! I think this is a backwards compatible change (i.e older versions of synapse will still work after this delta has been applied), so we don't need to bump the schema version here.

richvdh:
As I wrote on #4023: I think the issue is already resolved, so there's no need for this PR :/

synapse/storage/databases/main/schema/delta/60/federation_stream_position_unique_idx.sql
djmitche: Ah, I made this mistake because I looked at the `full_schema` on the assumption it was the latest.  In other projects I work on, we use code generation to ensure that such things are always up-to-date, with a CI failure if that's forgotten.  Is that idea of interest?

Also, shame on me for omitting the critical "UNIQUE"!

PR #9832, opened 2021-04-16
Port "Allow users to click account renewal links multiple times without hitting an 'Invalid Token' page #74" from synapse-dinsic

anoadragon453:
This attempts to be a direct port of https://github.com/matrix-org/synapse-dinsic/pull/74 to mainline. There was some fiddling required to deal with the changes that have been made to mainline since (mainly dealing with the split of `RegistrationWorkerStore` from `RegistrationStore`, and the changes made to `self.make_request` in test code).

The original PR went through review already (hence not picking apart the commits here), but I've noted the changes that were made in order to get this working with mainline below.

### So what does this code actually do, and what does it mean for mainline?

As explained in the linked PR, it prevents account validity tokens from being deleted immediately after use, and instead just marks them as used (note that tokens are still expired according to the `expiration_ts_ms` column). If a token has been used we allow the endpoint to continue to succeed if called with the same token, but the user's expiration time is no longer updated.

A new template file was added, `account_previously_renewed.html`, which is what's sent to the user in this case. It leaves out the bit about their account renewal being updated, and instead just tells them how long their account is still valid for.

### What was changed?

I had to make some changes from the original PR in order to port it to mainline successfully:

1. The migration file which added a column tracking when a user had used a token to the `account_validity` table was in migration folder `58`. Since we're on `59` now, adding it to `58` would have no effect. https://github.com/matrix-org/synapse/commit/4caface312d40a744f6704872346c9bc817e8c1e moves it to `59`.
2. It appears we didn't actually add the new `account_validity.account_previously_renewed_html_path` to the sample config in the original PR. 69e4a359d1 adds it, and cleans up the wording in the surrounding template paths a little bit.
3. 7ce453f2e7 adds a note to UPGRADE.rst informing anyone that's created a custom `account_renewed.html` template that they can now show the user's expiration date and time. It also mentions the new `account_previously_renewed.html` template, and that it can accept the same.


UPGRADE.rst
```diff
@@ -85,6 +85,29 @@ for example:
...
+Account Validity HTML templates can now display a user's expiration date
+------------------------------------------------------------------------
+
+This may affect you if you have enabled the account validity feature, and have made use of a
+custom HTML template specified by the ``account_validity.template_dir`` and ``account_validity.account_renewed_html_path``
+Synapse config options.
+
+The template can now accept an ``expiration_ts`` variable, which represents the unix timestamp in milliseconds for the
+future date of which their account has been renewed until. See the
+`default template <https://github.com/matrix-org/synapse/blob/release-v1.33.0/synapse/res/templates/account_renewed.html>`
```
richvdh, line 100:
```suggestion
`default template <https://github.com/matrix-org/synapse/blob/release-v1.33.0/synapse/res/templates/account_renewed.html>`_
```

UPGRADE.rst
```diff
@@ -85,6 +85,29 @@ for example:
...
+for an example of usage.
+
+Relatedly, note that in this update a new HTML template has been added which is shown to users when they
+attempt to renew their account with a valid renewal token that has already been used before. The default template
+contents can been found
+`here <https://github.com/matrix-org/synapse/blob/release-v1.33.0/synapse/res/templates/account_previously_renewed.html>`, and
```
richvdh, line 106:
```suggestion
`here <https://github.com/matrix-org/synapse/blob/release-v1.33.0/synapse/res/templates/account_previously_renewed.html>`_, and
```

docs/sample_config.yaml
```diff
@@ -1432,6 +1369,83 @@ account_threepid_delegates:
...
+  #
+  #account_renewed_html_path: "account_renewed.html"
+
+  # File within 'template_dir' giving the HTML to be displayed to the user if
+  # they attempt to renew their account with a token that is valid, but that
+  # has already been used. The account is not renewed again in this case.
+  #
+  # If not set, the file is assumed to be named "account_previously_renewed.html".
+  #
+  #account_previously_renewed_html_path: "account_previously_renewed.html"
```
richvdh, line 1439: why do we have a configuration setting for this, rather than simply mandating the name of the file? Making it configurable seems overcomplicated.

synapse/config/account_validity.py
```diff
@@ -0,0 +1,157 @@
...
+                account_renewed_template_filename,
+                account_previously_renewed_template_filename,
+                invalid_token_template_filename,
+            ]
+        )
+
+    def generate_config_section(self, **kwargs):
+        return """\
+        ## Account Validity ##
+        #
```
richvdh, line 83:
```suggestion
        
```

synapse/api/auth.py
```diff
@@ -79,7 +79,7 @@ def __init__(self, hs):
 
         self._auth_blocking = AuthBlocking(self.hs)
 
-        self._account_validity = hs.config.account_validity
+        self._account_validity_enabled = hs.config.account_validity_enabled
```
richvdh, line 82: I won't demand you change this, since it's a backport, but we really ought to be using the fully-specified names for the config options rather than adding more places we fall back on the grandfathered magic:

```suggestion
        self._account_validity_enabled = hs.config.account_validity.account_validity_enabled
```

The idea is that we stop using the magic grandfathering, then remove the magic, then we can give these things less cumbersome names.

synapse/config/account_validity.py
```diff
@@ -0,0 +1,157 @@
...
+                self.account_validity_period * 10.0 / 100.0
+            )
+
+        if self.account_validity_renew_by_email_enabled:
+            if not self.public_baseurl:
+                raise ConfigError("Can't send renewal emails without 'public_baseurl'")
+
+        # Load account validity templates.
+        # We do this here instead of in AccountValidityConfig as read_templates
+        # relies on state that hasn't been initialised in AccountValidityConfig
```
richvdh, line 58: I don't understand this comment. We are in `AccountValidityConfig`.

synapse/config/account_validity.py
```diff
@@ -0,0 +1,157 @@
...
+            self.account_validity_account_renewed_template,
+            self.account_validity_account_previously_renewed_template,
+            self.account_validity_invalid_token_template,
+        ) = self.read_templates(
+            [
+                account_renewed_template_filename,
+                account_previously_renewed_template_filename,
+                invalid_token_template_filename,
+            ]
+        )
```
richvdh, line 78: this seems to ignore the `account_validity.template_dir` setting.

synapse/handlers/account_validity.py
```diff
@@ -220,50 +228,87 @@ async def _get_renewal_token(self, user_id: str) -> str:
...
+        token is considered stale. A token is stale if the 'token_used_ts_ms' db column
+        is non-null.
+
         Args:
             renewal_token: Token sent with the renewal request.
         Returns:
-            Whether the provided token is valid.
+            A tuple containing:
+              * A bool representing whether the token is valid and unused.
+              * A bool representing whether the token is stale.
```
richvdh, line 244:
```suggestion
              * A bool which is `true` if the token is valid, but stale.
```

richvdh (commented):
generally seems ok. A few nitpicks.

synapse/api/auth.py
anoadragon453: Oh, right! I think this was before I was really aware of it. This has a slight performance improvement as well, correct?

richvdh: correct.

docs/sample_config.yaml
anoadragon453: Mmm, yeah I agree. It was mostly for consistency with the existing options but we shouldn't add to the problem.

I'll give a short explanation in the upgrade notes.

synapse/api/auth.py
anoadragon453: Done in bd15afef24!

docs/sample_config.yaml
anoadragon453: This required a bit of rework in the sample config, but I think it's come out better overall.

richvdh (approved):
lgtm other than what looks like a bit of a thinko... (looks like it's breaking CI too)

docs/sample_config.yaml
```diff
@@ -1421,25 +1421,33 @@ account_validity:
...
-  # they attempt to renew their account with a token that is valid, but that
-  # has already been used. The account is not renewed again in this case.
-  #
-  # If not set, the file is assumed to be named "account_previously_renewed.html".
-  #
-  #account_previously_renewed_html_path: "account_previously_renewed.html"
-
-  # File within 'template_dir' giving the HTML to be displayed when the user
-  # tries to renew an account with an invalid renewal token.
+  # A custom file name for the 'invalid_token.html' template.
```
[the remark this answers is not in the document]
anoadragon453: Hmm, I'm not too opposed to that really. But I think we should do it in a separate PR anyhow.

synapse/config/account_validity.py
```diff
@@ -0,0 +1,164 @@
...
+            self.account_validity_account_renewed_template,
+            self.account_validity_account_previously_renewed_template,
+            self.account_validity_invalid_token_template,
+        ) = self.read_templates(
+            [
+                account_renewed_template_filename,
+                invalid_token_template_filename,
+            ],
+            account_validity_template_dir,
+        )
```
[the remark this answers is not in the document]
anoadragon453: Gah, no it certainly wouldn't work!

Issue #9846, opened 2021-04-19
Minor updates to UPGRADE.rst requested for 1.32.0


callahad:
1.32's UPGRADE.rst should probably:

1. Call out the Py36/Pg96 requirement
2. Call out the AS registration change (and maybe call out when we deprecated that in [1.30.0](https://github.com/matrix-org/synapse/releases/tag/v1.30.0)? FWIW, that probably should've gone in the UPGRADE notes then as well.)

Basically the whole body from the top of https://github.com/matrix-org/synapse/releases/tag/v1.32.0rc1 should probably be in there.

(Sorry for filing a GH Issue and not submitting a patch myself... currently on my phone)

PR #9849, opened 2021-04-20
Add Application Service registration type requirement + py35, pg95 deprecation notices to v1.32.0 upgrade notes

anoadragon453:
Fixes https://github.com/matrix-org/synapse/issues/9846.

Adds important removal information from the top of https://github.com/matrix-org/synapse/releases/tag/v1.32.0rc1 into UPGRADE.rst.

### DIFF
--- a/UPGRADE.rst
+++ b/UPGRADE.rst
@@ -106,8 +106,8 @@ has been available since Synapse 1.7.0 (2019-12-13), and is accessible under ``G
 
 The deprecation of the old endpoint was announced with Synapse 1.28.0 (released on 2021-02-25).
 
-Application Services must use type ``m.login.application`` when registering users
----------------------------------------------------------------------------------
+Application Services must use type ``m.login.application_service`` when registering users
+-----------------------------------------------------------------------------------------
 
 In compliance with the
 `Application Service spec <https://matrix.org/docs/spec/application_service/r0.1.2#server-admin-style-permissions>`_,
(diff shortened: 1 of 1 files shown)


UPGRADE.rst
```diff
@@ -88,6 +88,14 @@ for example:
 Upgrading to v1.32.0
 ====================
 
+Dropping support for end-of-life Python, Postgres and SQLite versions
+---------------------------------------------------------------------
+
+In line with our `deprecation policy <https://github.com/matrix-org/synapse/blob/v1.32.0rc1/docs/deprecation_policy.md>`_,
```
babolivier, line 94: Maybe want to change this link so it points to the stable release (or master) instead of the RC?

babolivier (approved):
lgtm otherwise (modulo missing newsfile)

anoadragon453:
I'm not sure if we want a newsfile for this change? It would be the only change since the RC.

UPGRADE.rst
anoadragon453: Probably best to keep it at the current stable just in case we change it significantly (or move it).

UPGRADE.rst
```diff
@@ -88,6 +88,14 @@ for example:
 Upgrading to v1.32.0
 ====================
 
+Dropping support for end-of-life Python, Postgres and SQLite versions
```
callahad, line 91: This is the most tiny nitpick, but I don't think minor versions of SQLite actually go EOL / have a defined support lifetimes.

Maybe just `s/end-of-life/old/` on the heading?

anoadragon453:
We've decided to merge without a changelog entry.

PR #9854, opened 2021-04-20
Stop BackgroundProcessLoggingContext making new prometheus timeseries

richvdh:
This undoes part of b076bc276e881b262048307b6a226061d96c4a8d.

### DIFF
--- a/synapse/metrics/background_process_metrics.py
+++ b/synapse/metrics/background_process_metrics.py
@@ -249,7 +249,7 @@ class BackgroundProcessLoggingContext(LoggingContext):
 
         Args:
             name: The name of the background process. Each distinct `name` gets a
-                separate prometheus data series.
+                separate prometheus time series.
 
             instance_id: an identifer to add to `name` to distinguish this instance of
                 the named background process, to distinguish them in the logs. If
--- a/synapse/metrics/background_process_metrics.py
+++ b/synapse/metrics/background_process_metrics.py
@@ -16,7 +16,7 @@
 import logging
 import threading
 from functools import wraps
-from typing import TYPE_CHECKING, Dict, Optional, Set
+from typing import TYPE_CHECKING, Dict, Optional, Set, Union
 
 from prometheus_client.core import REGISTRY, Counter, Gauge
 
@@ -199,7 +199,7 @@ def run_as_background_process(desc: str, func, *args, bg_start_span=True, **kwar
         _background_process_start_count.labels(desc).inc()
         _background_process_in_flight_count.labels(desc).inc()
 
-        with BackgroundProcessLoggingContext("%s-%s" % (desc, count)) as context:
+        with BackgroundProcessLoggingContext(desc, count) as context:
             try:
                 ctx = noop_context_manager()
                 if bg_start_span:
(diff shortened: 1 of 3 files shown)


richvdh:
It is likely that the bug will leave people's Prometheus instances full of invalid data. Some notes on resolving the problem:

 * The first job is to get Prometheus started. If it is stuck in a loop of replaying writeahead logs, you will see something like this in the logs:
    ```
    Apr 20 18:25:59 repo prometheus[4124]: level=info ts=2021-04-20T16:25:59.930Z caller=head.go:575 component=tsdb msg="replaying WAL, this may take awhile"
    Apr 20 18:26:00 repo prometheus[4124]: level=info ts=2021-04-20T16:26:00.125Z caller=head.go:600 component=tsdb msg="WAL checkpoint loaded"
    Apr 20 18:26:01 repo prometheus[4124]: level=info ts=2021-04-20T16:26:01.371Z caller=head.go:624 component=tsdb msg="WAL segment loaded" segment=4775 maxSegment=4839
   Apr 20 18:26:02 repo prometheus[4124]: level=info ts=2021-04-20T16:26:02.375Z caller=head.go:624 component=tsdb 
   ...
   Apr 20 18:26:24 repo prometheus[4124]: level=info ts=2021-04-20T16:26:24.894Z caller=head.go:624 component=tsdb msg="WAL segment loaded" segment=4817 maxSegment=4839
   Apr 20 18:26:33 repo prometheus[4124]: fatal error: runtime: out of memory
   ```
   ... which then repeats, without making much progress.

   The only solution we know of to this is to delete some of the recent writeahead logs, for example: `rm /var/lib/prometheus/metrics/wal/0*`

* Once prometheus is running, it may still have thousands of invalid dataseries, which will impact its performance. The bad dataseries can be cleared out with the following:
   ```sh
   for series in \
     synapse_background_process_db_sched_duration_seconds \
     synapse_background_process_db_sched_duration_seconds_created \
     synapse_background_process_db_sched_duration_seconds_total \
     synapse_background_process_db_txn_count \
     synapse_background_process_db_txn_count_created \
     synapse_background_process_db_txn_count_total \
     synapse_background_process_db_txn_duration_seconds \
     synapse_background_process_db_txn_duration_seconds_created \
     synapse_background_process_db_txn_duration_seconds_total \
     synapse_background_process_ru_stime_seconds \
     synapse_background_process_ru_stime_seconds_created \
     synapse_background_process_ru_stime_seconds_total \
     synapse_background_process_ru_utime_seconds \
     synapse_background_process_ru_utime_seconds_created \
     synapse_background_process_ru_utime_seconds_total \
   ; do
       echo "cleaning up $series"
       curl -g 'http://localhost:9090/api/v1/admin/tsdb/delete_series?match[]='$series'{name=~".*-[0-9][0-9]*"}' -d ''   
       curl 'http://localhost:9090/api/v2/admin/tsdb/clean_tombstones' -d ''
   done
   ```

   This requires the Prometheus admin API to be enabled with the `--web.enable-admin-api` commandline flag (see https://prometheus.io/docs/prometheus/latest/querying/api/#tsdb-admin-apis and https://www.robustperception.io/deleting-time-series-from-prometheus).

   Prometheus may OOM partway through the above, causing some of the requests to fail. If that happens, restart Prometheus, wait for it to start running again, and then restart the script.

synapse/metrics/background_process_metrics.py
```diff
@@ -244,8 +244,20 @@ class BackgroundProcessLoggingContext(LoggingContext):
...
+    def __init__(self, name: str, instance_id: Optional[Union[int, str]] = None):
+        """
+
+        Args:
+            name: The name of the background process. Each distinct `name` gets a
+                separate prometheus time series.
+
+            instance_id: an identifer to add to `name` to distinguish this instance of
+                the named background process, to distinguish them in the logs. If
+                this is `None`, one is made up based on id(self).
```
richvdh, line 256:
```suggestion
            instance_id: an identifer to add to `name` to distinguish this instance of
                the named background process in the logs. If this is `None`, one is
                made up based on id(self).
```

clokep (approved):
Makes sense. I wonder if the `BackgroundProcessLoggingContext` created in `RedisSubscriber` should also be provided an `instance_id`?

richvdh:
> Makes sense. I wonder if the BackgroundProcessLoggingContext created in RedisSubscriber should also be provided an instance_id?

possibly, but I couldn't see an obvious thing to name it, which was why I did the `id(self)` thing.

MTRNord:
@richvdh Just as a hint if some others get 404 errors for the second curl command:

I needed to run `curl -XPOST 'http://localhost:9999/api/v1/admin/tsdb/clean_tombstones'` instead with prometheus 2.26.0. The v2 endpoint just did a 404 error

PR #9829, opened 2021-04-16
Always use the name as the log ID.

clokep:
As far as I can tell our logging contexts are meant to log the request ID, or sometimes the request ID followed by a suffix (this is generally stored in the `name` field of `LoggingContext`). There's also code to log the name@memory location, but I'm not sure this is ever used.

This simplifies the code paths to require every logging context to have a name and use that in logging. For sub-contexts (created via `nested_logging_contexts`, `defer_to_threadpool`, `Measure`) we use the current context's str (which becomes their name or the string "sentinel") and then potentially modify that (e.g. add a suffix).

Fixes #9824

PR #9857, opened 2021-04-21
Make LoggingContext's name optional

richvdh:
Fixes https://github.com/matrix-org/synapse-s3-storage-provider/issues/55

### DIFF
--- a/synapse/logging/context.py
+++ b/synapse/logging/context.py
@@ -258,7 +258,8 @@ class LoggingContext:
           child to the parent
 
     Args:
-        name (str): Name for the context for debugging.
+        name: Name for the context for logging. If this is omitted, it is
+           inherited from the parent context.
         parent_context (LoggingContext|None): The parent of the new context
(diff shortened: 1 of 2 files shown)


synapse/logging/context.py
```diff
@@ -314,6 +314,15 @@ def __init__(
             # the request param overrides the request from the parent context
             self.request = request
 
+        # if we don't have a `name`, but do have a parent context, use its name.
+        if self.parent_context and not name:
+            name = str(self.parent_context)
+        if not name:
```
clokep, line 320: This will also catch empty strings, which should probably be OK, but not 100% sure. 😄

richvdh: hrm point. I hate it when we do that.

clokep (approved):

erikjohnston (approved):

clokep:
This seems to have been a regression from #9829.

PR #9858, opened 2021-04-21
Handle all new rate limits in demo scripts

manuroe:
The `--no-rate-limit` parameter should disable, or make useless, any rate limiting.
I updated start.sh with the list available in sample_config.yaml


erikjohnston (approved):
Sad that we do this manually, but ah well.

changelog.d/9858.misc
```diff
@@ -0,0 +1 @@
+Handle all new rate limits in demo scripts.
```
erikjohnston, line 1:
```suggestion
Handle recently added rate limits correctly when using `--no-rate-limit` with the demo scripts.
```

PR #9868, opened 2021-04-22
Pass back errors to the client when trying multiple federation destinations

clokep:
This is based off a discussion: https://github.com/matrix-org/synapse/pull/9814#discussion_r614735955

The goal of this is to pass back better error messages to the client when using the `_try_destination_list` helper in `FederationClient`. The overall approach is to ensure that the callbacks passed to that raise `HttpResponseException` when they receive an error that should *stop* trying not destinations and be returned to the client. (Previously each destination would be tried and the client gets a generic 502 error: "Failed to something via any server")

This is broken into small commits which should be reviewable separately.

I think it was likely an error previously that the callbacks raised `SynapseError` (which cause `_try_destination_list` to try the next server) as opposed to raising the `HttpResponseException`, which short circuits and kills the request.


synapse/federation/federation_client.py
```diff
@@ -744,15 +755,10 @@ async def _do_send_join(self, destination: str, pdu: EventBase) -> JsonDict:
...
-                # If we receive an error response that isn't a generic error, or an
-                # unrecognised endpoint error, we  assume that the remote understands
-                # the v2 invite API and this is a legitimate error.
-                if err.errcode not in [Codes.UNKNOWN, Codes.UNRECOGNIZED]:
-                    raise
-            else:
+            # If an error is received that is due to an unrecognised endpoint,
+            # fallback to the v1 endpoint. Otherwise consider it a legitmate error
+            # and raise.
+            if not self._is_unknown_endpoint(e, e.to_synapse_error()):
```
richvdh, line 761: it's tempting to move the `to_synapse_error` call into `_is_unknown_endpoint` even though that causes duplication in `_try_destination_list`. I feel like that makes `_is_unknown_endpoint` clearer.

synapse/federation/federation_client.py
```diff
@@ -821,28 +821,19 @@ async def _do_send_invite(
...
                     raise SynapseError(
                         400,
                         "User's homeserver does not support this room version",
                         Codes.UNSUPPORTED_ROOM_VERSION,
                     )
-            elif e.code in (403, 429):
-                raise e.to_synapse_error()
             else:
-                raise
+                raise err
```
richvdh, line 836: you might update the docstring to clarify the behaviour, while you're changing it.

changelog.d/9868.bugfix
```diff
@@ -0,0 +1 @@
+Fix a long-standing bug where errors from federation did not propagate to the client.
```
richvdh, line 1: The changes all look sensible, but I'm not entirely sure which errors which were previously swallowed are now propagated correctly. Do you have any examples?

richvdh (commented):
generally lgtm. a few little nits.

synapse/federation/federation_client.py
clokep: I also debated whether we could make it optional and generate it inside of `_is_unknown_endpoint` if it isn't provided?

changelog.d/9868.bugfix
clokep: Well the rationale for the work is the 403 error in #9814 is now passed back to the client as a 403, instead of a 400.

clokep: I should probably mention that this is tested by matrix-org/complement#99.

erikjohnston (approved):

PR #9875, opened 2021-04-23
Make DomainSpecificString an attrs class

erikjohnston:
Mostly because if you try and do `asizeof(user)` it'll try and iterate (since it has an `__iter__`), but then it'll fail as `__iter__` raises. Plus, this easily allows us to add type hints.

### DIFF
--- a/synapse/handlers/oidc.py
+++ b/synapse/handlers/oidc.py
@@ -957,6 +957,8 @@ class OidcProvider:
                 # and attempt to match it.
                 attributes = await oidc_response_to_user_attributes(failures=0)
 
+                assert attributes.localpart
+
                 user_id = UserID(attributes.localpart, self._server_name).to_string()
                 users = await self._store.get_users_by_id_case_insensitive(user_id)
                 if users:
--- a/synapse/handlers/oidc.py
+++ b/synapse/handlers/oidc.py
@@ -957,7 +957,10 @@ class OidcProvider:
                 # and attempt to match it.
                 attributes = await oidc_response_to_user_attributes(failures=0)
 
-                assert attributes.localpart
+                if attributes.localpart is None:
+                    # If no localpart is returned then we will generate one, so
+                    # there is no need to search for existing users.
+                    return None
 
                 user_id = UserID(attributes.localpart, self._server_name).to_string()
                 users = await self._store.get_users_by_id_case_insensitive(user_id)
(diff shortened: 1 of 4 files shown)


synapse/handlers/oidc.py
```diff
@@ -957,6 +957,8 @@ async def grandfather_existing_users() -> Optional[str]:
                 # and attempt to match it.
                 attributes = await oidc_response_to_user_attributes(failures=0)
 
+                assert attributes.localpart
```
erikjohnston, line 960: Does anyone know why this is (and the other place) is safe? It's not immediately obvious why this can't be `None` here (it's defined as `Optional[str]`)

richvdh: I suspect in this case it's not. It relies on `_allow_existing_users` not being set at the same time as having a `_user_mapping_provider` which returns `localpart is None`

On the other one, it's impossible to get to `NewUserConsentResource` without having first been through the pick-a-username step which ensures `chosen_localpart` gets set.

erikjohnston: Thanks!

Oh, and actually in the OIDC case we only use the generated user ID for comparison with existing users, so it doesn't matter if its None? In which case we should instead `return None` if localpart is None?

richvdh: yes.

synapse/rest/synapse/client/new_user_consent.py
```diff
@@ -61,6 +61,15 @@ async def _async_render_GET(self, request: Request) -> None:
...
             return
 
+        # It should be impossible to get here without having first been through
+        # the pick-a-username step, which ensures chosen_localpart gets set.
+        if not session.chosen_localpart:
+            logger.warning("Session has no user name selected")
+            self._sso_handler.render_error(
+                request, "no_user", "No user name has been selected.", code=400
+            )
+            return
```
erikjohnston, line 71: My thinking for not asserting here is that I assume its possible for a client that is being stupid to get here, and so we don't want to get stack traces/sentry alerts?

synapse/handlers/oidc.py
```diff
@@ -957,6 +957,11 @@ async def grandfather_existing_users() -> Optional[str]:
                 # and attempt to match it.
                 attributes = await oidc_response_to_user_attributes(failures=0)
 
+                if attributes.localpart is None:
+                    # If no localpart is returned then we will generate one, so
+                    # there is no need to search for existing users.
+                    return None
```
clokep, line 963: I guess this previously worked by there hopefully not being an `@None:foo` user?

clokep (approved):
Looks good. 👍

synapse/handlers/oidc.py
erikjohnston: Yup, I believe so

PR #9876, opened 2021-04-23
Add types to auth and auth_blocking.

clokep:
I was poking in this code and got annoyed with not having auto-complete.

I think it is all pretty straightforward, there's a second commit which removes an unused method.

### DIFF
--- a/synapse/event_auth.py
+++ b/synapse/event_auth.py
@@ -14,7 +14,7 @@
 # limitations under the License.
 
 import logging
-from typing import List, Optional, Set, Tuple
+from typing import Dict, List, Optional, Set, Tuple
 
 from canonicaljson import encode_canonical_json
 from signedjson.key import decode_verify_key_bytes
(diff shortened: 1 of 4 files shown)


synapse/event_auth.py
```diff
@@ -688,7 +688,7 @@ def _verify_third_party_invite(event: EventBase, auth_events: StateMap[EventBase
     return False
 
 
-def get_public_keys(invite_event):
+def get_public_keys(invite_event: EventBase) -> List[Dict[str, str]]:
```
clokep, line 691: I'm not 100% confident that the values here are strings, but I think they are?

erikjohnston: I think so too, but not 100% sure. I wonder if we want to be saying they are strings if we haven't actually asserted that? The event content is untrusted data after all

clokep: Would `Dict[str, Any]` make more sense?

erikjohnston: Yeah, I'd vote we either assert they're strings or say they're `Any`?

clokep: Done!

erikjohnston (approved):

PR #8013, opened 2020-07-31
Iteratively encode JSON responses

clokep:
This uses a producer to encode JSON in chunks to respond to requests. Fixes #6998.

This unfortunately rolls back a bit of the abstraction since `respond_with_json` can't call `respond_with_json_bytes` anymore... further it seems that `respond_with_json_bytes` is reasonable called in a couple of other places. 😢 

~~This also uses private members from `canonicaljson` which is kind of dirty, but wanted to put it up as a proof of concept.~~

PR #9878, opened 2021-04-23
Kill off `_PushHTTPChannel`.

richvdh:
Two commits here: suggest reviewing them separately.

First of all, a fixup to `FakeChannel` which is needed to make it work with the default HTTP channel implementation.

Secondly, it looks like we no longer need `_PushHTTPChannel`, because as of #8013, the producer that gets attached to the `HTTPChannel` is now an `IPushProducer`. This is good, because it means we can remove a whole load of test-specific boilerplate which causes variation between tests and production.

Based on #9874.


clokep (approved):
I'm a little surprised that mypy is OK with the request factory shenanigans, but seems to pass OK. 👍

PR #9879, opened 2021-04-23
Remove various bits of compatibility code for Python <3.6

anoadragon453:
I went through and removed a bunch of cruft that was lying around for compatibility with old Python versions. This PR also will now prevent Synapse from starting unless you're running Python 3.6+.

Best reviewed commit-by-commit.


synapse/secrets.py
```diff
@@ -15,30 +15,14 @@
...
-    import os
-
-    class Secrets:
-        def token_bytes(self, nbytes: int = 32) -> bytes:
-            return os.urandom(nbytes)
-
-        def token_hex(self, nbytes: int = 32) -> str:
-            return binascii.hexlify(self.token_bytes(nbytes)).decode("ascii")
+    def token_hex(self, nbytes: int = 32) -> str:
+        return secrets.token_hex(nbytes)
```
clokep, line 28: Looks like we can get rid of this and just directly use the `secrets` module?

synapse/storage/_base.py
```diff
@@ -114,7 +114,7 @@ def db_to_json(db_content: Union[memoryview, bytes, bytearray, str]) -> Any:
         db_content = db_content.tobytes()
 
     # Decode it to a Unicode string before feeding it to the JSON decoder, since
-    # Python 3.5 does not support deserializing bytes.
+    # it only supports handling strings
```
clokep, line 117: Isn't the opposite true now? That Python > 3.5 supports bytes or strings?

clokep (changes requested):
Looks good overall! Couple of comments.

synapse/storage/_base.py
anoadragon453: There's a bit of an explanation in the commit message of 0506cc9. Essentially we switched to an instance of `JSONDecoder` that looks to only support `str` arguments now. The comment was originally written when we were using `json.decode`, which did support different argument types based on the Python version.

clokep (approved):

clokep:
Looks great! Thanks for cleaning this up! 🎉

PR #9881, opened 2021-04-26
Optionally track memory usage of each LruCache

erikjohnston:
This will double count slightly in the presence of interned strings. It's off by default as it can consume a lot of resources.


richvdh (approved):
lgtm

PR #9882, opened 2021-04-26
Export jemalloc stats to prometheus when used

erikjohnston:
This is useful to see how much memory Synapse is actually using, as opposed to the memory reserved from the OS.

We do this by loading jemalloc lib and doing calls to `mallctl` to fetch the various stats we care about, such as `stats.allocated`.

**Note**: By default python has a small object allocator that sits on top of jemalloc, so unless that is disabled via `PYTHONMALLOC=malloc` the `allocated` stats won't be accurate.


synapse/metrics/__init__.py
```diff
@@ -597,6 +600,163 @@ def f(*args, **kwargs):
...
+    # path, as if we load a *different* jemalloc version things will seg fault.
+    pid = os.getpid()
+
+    # We're looking for a path at the end of the line that includes
+    # "libjemalloc".
+    regex = re.compile(r"/\S+/libjemalloc.*$")
+
+    jemalloc_path = None
+    with open(f"/proc/{pid}/maps") as f:
+        for line in f.readlines():
```
richvdh, line 619:
```suggestion
        for line in f:
```

synapse/metrics/__init__.py
```diff
@@ -597,6 +600,163 @@ def f(*args, **kwargs):
     return f
 
 
+def _setup_jemalloc_stats():
```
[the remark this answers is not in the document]
richvdh: of course, if you were really keen, you'd make a separate python package that can be reused by other jemalloc users...

synapse/metrics/__init__.py
```diff
@@ -597,6 +600,163 @@ def f(*args, **kwargs):
...
+            pass
+
+    class JemallocCollector:
+        """Metrics for internal jemalloc stats."""
+
+        def collect(self):
+            _jemalloc_refresh_stats()
+
+            g = GaugeMetricFamily(
+                "jemalloc_stats_app_memory",
```
richvdh, line 712:
```suggestion
                "jemalloc_stats_app_memory_bytes",
```

synapse/metrics/__init__.py
```diff
@@ -597,6 +600,163 @@ def f(*args, **kwargs):
...
+
+    def _jemalloc_refresh_stats() -> None:
+        """Request that jemalloc updates its internal statistics. This needs to
+        be called before querying for stats, otherwise it will return stale
+        values.
+        """
+        try:
+            _mallctl("epoch", read=False, write=1)
+        except Exception:
+            pass
```
richvdh, line 703: ...?

synapse/metrics/__init__.py
```diff
@@ -597,6 +600,163 @@ def f(*args, **kwargs):
...
+                "active",
+                "resident",
+                "mapped",
+                "metadata",
+            ):
+                try:
+                    value = _mallctl(f"stats.{t}")
+                except Exception:
+                    # There was an error fetching the value, skip.
+                    continue
```
richvdh, line 746: plsno

richvdh (commented):
looks generally good. Can you make it less exception-swallowy?

synapse/app/_base.py
```diff
@@ -115,6 +116,7 @@ def start_reactor(
 
     def run():
         logger.info("Running")
+        setup_jemalloc_stats()
```
erikjohnston, line 119: I changed it to get called here, mainly so that we try to set it up *after* we've got logging set up.

synapse/metrics/jemalloc.py
```diff
@@ -0,0 +1,191 @@
...
+
+    def _jemalloc_refresh_stats() -> None:
+        """Request that jemalloc updates its internal statistics. This needs to
+        be called before querying for stats, otherwise it will return stale
+        values.
+        """
+        try:
+            _mallctl("epoch", read=False, write=1)
+        except Exception:
+            pass
```
richvdh, line 131: I still feel like we shouldn't be completely dropping these exceptions. Why not log something at `warn`?

synapse/metrics/jemalloc.py
```diff
@@ -0,0 +1,191 @@
...
+    logger.debug("Added jemalloc stats")
+
+
+def setup_jemalloc_stats():
+    """Try to setup jemalloc stats, if jemalloc is loaded."""
+
+    try:
+        _setup_jemalloc_stats()
+    except Exception as e:
+        logger.info("Failed to setup collector to record jemalloc stats: %s", e)
```
richvdh, line 191: when do you expect this to be hit?

richvdh (changes requested):

synapse/metrics/jemalloc.py
erikjohnston: Oh, github obviously marked this as outdated. Will fix.

synapse/metrics/jemalloc.py
erikjohnston: This can happen if we choose the wrong jemalloc library mainly, though that shouldn't really happen if we're looking in `/proc/self/maps`

richvdh (approved):

PR #9884, opened 2021-04-26
Add a module type for account validity

babolivier:
This PR adds an API for third-party plugin modules to implement account validity, so they can provide this feature instead of Synapse. The module implementing the current behaviour for this feature can be found at https://github.com/matrix-org/synapse-email-account-validity.

To allow for a smooth transition between the current feature and the new module, hooks have been added to the existing account validity endpoints to allow their behaviours to be overridden by a module.

~~Best reviewed commit by commit - there are quite a few of them but only the 4 first ones are a bit large.~~

~~Requires #9977~~


babolivier:
I'd also appreciate if someone could sanity-check https://github.com/matrix-org/synapse-email-account-validity - it should be fairly OK since most of it is copy-pasted from the current account validity implem, but it'd be better if someone else has had a look at it.

UPGRADE.rst
```diff
@@ -85,6 +85,19 @@ for example:
...
+Upgrading to v1.34.0
+====================
+
+Deprecation of the built-in account validity feature
+----------------------------------------------------
+
+As of Synapse v1.34.0, the built-in account validity feature is deprecated in favour of
+plugin modules. If you are using this feature and want to keep its current behaviour,
+please use the `synapse-email-account-validity <https://github.com/matrix-org/synapse-email-account-validity>`_
+module.
```
richvdh, line 97: I think we'll need clearer instructions for admins here: how do they install the module, how should they update the config?

richvdh (changes requested):
looks generally good, but a few bits and bpbs to fix up.

Thank you for trying to make it reviewable by splitting it up into separate commits, though there are a couple of things I saw in early commits which got fixed later, so excuse the irrelevant comments.

synapse/module_api/__init__.py
```diff
@@ -439,6 +489,144 @@ async def send_local_online_presence_to(self, users: Iterable[str]) -> None:
...
+                self._raw_from,
+                raw_to,
+                multipart_msg.as_string().encode("utf8"),
+                reactor=self._hs.get_reactor(),
+                port=self._hs.config.email_smtp_port,
+                requireAuthentication=self._hs.config.email_smtp_user is not None,
+                username=self._hs.config.email_smtp_user,
+                password=self._hs.config.email_smtp_pass,
+                requireTransportSecurity=self._hs.config.require_transport_security,
+            )
```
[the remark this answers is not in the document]
richvdh: (splitting up Mailer is probably best done in a preparatory PR)

UPGRADE.rst
babolivier: These instructions are available at https://github.com/matrix-org/synapse-email-account-validity, my idea was that admins would follow the link to the module and follow the steps described there, but maybe that's not clear enough?

richvdh: wellll... I think on the one hand, we need some migration instructions rather than instructions that start from scratch.

Also, `pip install` doesn't work for people who didn't install with pip. (docker, debian, etc, etc)

UPGRADE.rst
```diff
@@ -85,6 +85,19 @@ for example:
...
+
+Synapse's built-in account validity feature will be removed in a future release.
```
callahad, line 99: Should we define this more explicitly?

(E.g., giving people three months to migrate would mean dropping in Synapse 1.40 on 2021-08-10)

UPGRADE.rst
babolivier: > I think on the one hand, we need some migration instructions rather than instructions that start from scratch.

The module attempts a migration when starting fwiw so just installing it from scratch and replacing the `account_validity` config with the module's should be enough, but fair enough I'll add more instructions on this doc as well.

> Also, `pip install` doesn't work for people who didn't install with pip. (docker, debian, etc, etc)

True, however I don't know how we support non-pip installs since none of the other modules we publish seem to do.

synapse/api/auth.py
```diff
@@ -224,12 +224,22 @@ async def get_user_by_req(
             shadow_banned = user_info.shadow_banned
 
             # Deny the request if the user account has expired.
-            if self._account_validity_enabled and not allow_expired:
-                if await self.store.is_account_expired(
-                    user_info.user_id, self.clock.time_msec()
+            if not allow_expired:
+                if await self.hs.get_account_validity().user_expired(
```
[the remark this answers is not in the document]
richvdh: ohh, or better:

we change round the way modules work, so we just have a single `modules` list at the top level of the homeserver config. Each module gets loaded at startup, but after all the handlers. Then the module registers for the hooks it needs (in this case, `account_validity` and additional http resources).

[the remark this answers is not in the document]
babolivier: > I'm guessing that we're doing this `hs.get_account_validity()` here rather than in the constructor to avoid a dependency loop?

Yes, iirc it's because `hs.get_auth()` is called in the module API's `__init__` function (or because a handler that's listed [here](https://github.com/matrix-org/synapse/blob/2dd3a354b1eb827c2ea54c075bfaa6aa942a1d45/synapse/server.py#L215-L223) does it).

> we change round the way modules work, so we just have a single `modules` list at the top level of the homeserver config. Each module gets loaded at startup, but after all the handlers. Then the module registers for the hooks it needs (in this case, `account_validity` and additional http resources).

I was thinking of that earlier, and I really like the idea. It would also have the added bonus of making documenting module development easier since you wouldn't need a different documentation for every different type of module.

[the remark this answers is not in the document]
richvdh: >  I was thinking of that earlier, and I really like the idea.

Yeah. If you fancy making that change here, I think it will be very worthwhile (and hopefully not too much work).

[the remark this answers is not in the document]
babolivier: I think I'm happy to do this change but given it changes some things about other modules as well it might be better to do it in a separate PR.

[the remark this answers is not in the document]
richvdh: I think splitting up this PR is a good idea - it's got a bit unweildly.

I wouldn't leap into changing the way other modules work right now. We'll need to keep the existing interfaces around for quite a while, even if they are deprecated.

synapse/module_api/__init__.py
```diff
@@ -46,11 +65,22 @@ def __init__(self, hs, auth_handler):
...
         self._state = hs.get_state_handler()
+        self._sendmail = hs.get_sendmail()
+        self._clock = hs.get_clock()  # type: Clock
+
+        try:
+            app_name = self._hs.config.email_app_name
+
+            self._from_string = self._hs.config.email_notif_from % {"app": app_name}
+        except Exception:
+            # If substitution failed, fall back to the bare strings.
```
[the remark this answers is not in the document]
babolivier: Substitution fails if there are placeholders in the string that don't match a key in the dictionary:

```python
>>> "Your Friendly %(hs)s homeserver" % {"app": "foo"}
KeyError: 'hs'
>>> 
```

But yeah limiting to `KeyError` sounds like a sensible thing to do.

synapse/module_api/__init__.py
```diff
@@ -439,6 +489,144 @@ async def send_local_online_presence_to(self, users: Iterable[str]) -> None:
...
+
+        @wrap_as_background_process(f.__name__)
+        async def background_call(*args, **kwargs):
+            await f(*args, **kwargs)
+
+        if self._hs.config.run_background_tasks:
+            self._clock.call_later(0.0, background_call, *args, **kwargs)
+        else:
+            logger.warning(
+                "Not running looping call %s as the configuration forbids it",
```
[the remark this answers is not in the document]
babolivier: (resolving as this got fixed in another commit, sorry for the faff)

synapse/module_api/__init__.py
```diff
@@ -439,6 +501,122 @@ async def send_local_online_presence_to(self, users: Iterable[str]) -> None:
...
+            custom_template_directory: An additional directory to look for the files in.
+
+        Returns:
+            A list containing the loaded templates, with the orders matching the one of
+            the filenames parameter.
+        """
+        return self._hs.config.read_templates(filenames, custom_template_directory)
+
+    async def get_profile_for_user(self, localpart: str) -> ProfileInfo:
+        """Lookup the profile info for the user with the given localpart.
```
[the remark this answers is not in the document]
babolivier: Thanks, I often get those confused

docs/sample_config.yaml
```diff
@@ -1371,87 +1371,28 @@ account_threepid_delegates:
...
-  #
-  # If not set, the file is assumed to be named "invalid_token.html".
-  #
-  #invalid_token_html_path: "invalid_token.html"
+   #- module: "my_custom_project.SomeAccountValidity"
+   #  config:
+   #    example_option: 'things'
+   #- module: "my_custom_project.SomeOtherAccountValidity"
+   #  config:
+   #    other_example_option: 'stuff'
```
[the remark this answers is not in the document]
babolivier: It's not 3 spaces though, it's 2 (4 total from the `#`)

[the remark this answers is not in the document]
richvdh: there are definitely 3 spaces before the `#`.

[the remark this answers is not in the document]
babolivier: Oh, I thought you meant the indent after the `#` :man_facepalming:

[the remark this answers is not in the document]
richvdh: my bad, cryptic comment was cryptic

synapse/account_validity/__init__.py
```diff
@@ -0,0 +1,109 @@
...
+# limitations under the License.
+
+from typing import TYPE_CHECKING, Any, List, Tuple
+
+if TYPE_CHECKING:
+    import synapse.events
+    import synapse.server
+
+
+class AccountValidity:
```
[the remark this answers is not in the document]
babolivier: I looked at other modules (namely spam checker and third party event rules) which have this code (instantiating + calling hooks on modules) in a dedicated class outside of handlers, and thought of the account validity handler as the legacy code that'll eventually go away.

docs/sample_config.yaml
```diff
@@ -1371,87 +1371,28 @@ account_threepid_delegates:
...
+# Server admins can (optionally) get Synapse to check the validity of all user
+# accounts against one or more custom modules.
+#
+# An account validity module must implement two APIs:
+#
+#  * `user_expired`, which takes a Matrix user ID and returns one boolean to
+#    indicate whether the provided account has expired, and one boolean to
+#    indicate whether it succeeded in figuring out this info. If this second
+#    boolean value is False, the `user_expired function of the next module
+#    (in the order modules are configured in this file) is called. If there
```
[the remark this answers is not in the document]
babolivier: Ah yes that makes more sense, thanks

synapse/module_api/__init__.py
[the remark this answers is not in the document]
babolivier: I've open https://github.com/matrix-org/synapse/pull/9977 for that.

synapse/api/auth.py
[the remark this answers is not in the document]
babolivier: Resolving the thread as I believe the conclusion is that this will be done in a future PR (and I have the conn on making that happen)

UPGRADE.rst
babolivier: I'm a bit uncertain what this PR should be doing to fix this issue. We had a discussion about it in #synapse-dev and (I think?) came to a conclusion (I don't remember the specifics, but iirc it was about telling users how to build custom Docker images based on the Synapse one with the modules they want in it) but it feels a bit out of scope for this PR to me.

UPGRADE.rst
babolivier: Agreed that it needs defining more precisely. Given the time this PR is taking to get through, I'll determine that once the review is done since we'll at least know which version this change will be shipped in.

richvdh: historically we've always ended up looking stupid when we say "we will do X on date Y", and Y then sails past. It might be possible to come up with some wording that allows us wiggle room.

UPGRADE.rst
```diff
@@ -85,6 +85,27 @@ for example:
...
+
+
+Removal of account validity template configuration settings
+-----------------------------------------------------------
+
+In line with previous changes around templates management in Synapse, support has been
+removed for the configuration settings ``template_dir``, ``account_renewed_html_path`` and
+``invalid_token_html_path`` in the ``account_validity`` section of the configuration file.
+
```
richvdh, line 108: ... which previous changes? I don't remember us doing this sort of thing before.

UPGRADE.rst
```diff
@@ -85,6 +85,27 @@ for example:
...
+please use the `synapse-email-account-validity <https://github.com/matrix-org/synapse-email-account-validity>`_
+module.
+
+Synapse's built-in account validity feature will be removed in a future release.
+
+
+Removal of account validity template configuration settings
+-----------------------------------------------------------
+
+In line with previous changes around templates management in Synapse, support has been
```
richvdh, line 105:
```suggestion
In line with previous changes around template management in Synapse, support has been
```

UPGRADE.rst
richvdh: I appreciate it feels like scope creep, but I think it's a prerequisite for moving functionality out of core Synapse into modules, which is why I questioned whether that was a good decision.

In short: if we're going down that path, I'd like us to have solutions in place for modules with docker *and debian* before this PR lands.

synapse/api/auth.py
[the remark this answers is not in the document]
richvdh: I'm reopening this. Whilst we should leave the *existing* module hooks in place for a future PR, I don't think we should add any *new* module hooks that assume single-function modules, so I'm afraid I'd like to see changes before this PR lands.

UPGRADE.rst
babolivier: I seemed to remember we had done work to hardcode the name of templates in the past, but looking at other places in the code it looks like I've imagined that.

synapse/module_api/__init__.py
```diff
@@ -439,6 +489,144 @@ async def send_local_online_presence_to(self, users: Iterable[str]) -> None:
...
+        async def background_call(*args, **kwargs):
+            await f(*args, **kwargs)
+
+        if self._hs.config.run_background_tasks:
+            self._clock.call_later(0.0, background_call, *args, **kwargs)
+        else:
+            logger.warning(
+                "Not running looping call %s as the configuration forbids it",
+                f,
+            )
```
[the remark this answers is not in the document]
richvdh: yeah fair enough, I don't really have any great suggestions here. it's probably best to expose a `looping_call` method on ModuleApi which calls `run_as_background_process` on the function (I think you may as well call `run_as_background_process` directly rather than messing with `wrap_as_background_process`).

It might be nice to make the "description" an optional param rather than hardcoding it to the name of the function.

synapse/account_validity/__init__.py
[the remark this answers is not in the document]
richvdh: hrm; I'm not convinced that's a great model. 

spam-checker seems to be an entire module to define three constants - I'm not really sure why we did that.

In my mind, the REST layer calls things in the handlers. It might be better to rename the existing `AccountValidityHandler` (to `LegacyAccountValidityImpl` or something) but keep it in the same file. Then you define a new `AccountValidityHandler` which is called by the REST layer, and manages the calls out to the module as well as the fallback to the `LegacyAccountValidityImpl`.

UPGRADE.rst
richvdh: as a general rule we have hardcoded names for new templates. Where we have existing config settings, we have removed them from the sample config, but they continue to work. I'd suggest doing the latter here.

likewise `template_dir`: I would remove it from the sample config, and have it default to the global `template_dir`, but keep it working if someone has deliberately set it.

UPGRADE.rst
babolivier: ftr we decided to ditch the deprecation issue for now, so this PR isn't blocked on finding a solution to this problem.

synapse/module_api/__init__.py
```diff
@@ -101,22 +158,53 @@ def public_room_list_manager(self):
...
-    def get_user_by_req(self, req, allow_guest=False):
+    @property
+    def public_baseurl(self):
+        """Allow accessing the configured public base URL for this homeserver."""
+        return self._hs.config.public_baseurl
+
+    @property
+    def email_app_name(self):
+        """Allow accessing the application name configured in the homeserver's
+        configuration.
```
richvdh, line 169:
```suggestion
        """The application name configured in the homeserver's configuration.
```

synapse/module_api/__init__.py
```diff
@@ -101,22 +158,53 @@ def public_room_list_manager(self):
...
         return self._public_room_list_manager
 
-    def get_user_by_req(self, req, allow_guest=False):
+    @property
+    def public_baseurl(self):
+        """Allow accessing the configured public base URL for this homeserver."""
+        return self._hs.config.public_baseurl
+
+    @property
+    def email_app_name(self):
```
richvdh, line 167:
```suggestion
    def email_app_name(self) -> str:
```

synapse/handlers/account_validity.py
```diff
@@ -27,6 +29,14 @@
 
 logger = logging.getLogger(__name__)
 
+IS_USER_EXPIRED_CALLBACK = Callable[[str], Awaitable[Optional[bool]]]
+ON_USER_REGISTRATION_CALLBACK = Callable[[str], Awaitable]
+# Legacy callbacks to allow for a smoother transition between the legacy native account
+# validity feature and synapse-email-account-validity.
```
richvdh: As I said on our call: I think it would be good to expand this a bit to explain the status of these "legacy" callbacks, why they are needed, and how/when we hope to get rid of them.

synapse/handlers/account_validity.py
```diff
@@ -70,6 +80,96 @@ def __init__(self, hs: "HomeServer"):
...
+        # /_synapse/client/account_validity, because:
+        #   * the feature isn't part of the Matrix spec thus shouldn't live under /_matrix
+        #   * because of the way we register servlets modules can't register resources
+        #     under /_matrix/client
+        # We need to allow for a transition period between the old and new endpoints
+        # in order to allow for clients to update (and for emails to be processed).
+        # Therefore we need to allow modules (in practice just the one implementing the
+        # email-based account validity) to temporarily hook into the legacy endpoint so we
+        # can route the traffic coming into the old endpoints into the module, which is
+        # why we have the following three temporary hooks.
```
richvdh, line 121: given that we're keeping the old code around for now anyway, do we really need all of these hooks? For instance, can't we have the core code handle requests to `/_synapse/admin/v1/account_validity/validity` and the module handle requests to `/_synapse/client/account_validity/validity`, thus making the `on_legacy_admin_request` hook redundant?

synapse/handlers/account_validity.py
```diff
@@ -27,6 +29,12 @@
 
 logger = logging.getLogger(__name__)
 
+IS_USER_EXPIRED_CALLBACK = Callable[[str], Awaitable[Optional[bool]]]
+ON_USER_REGISTRATION_CALLBACK = Callable[[str], Awaitable]
+ON_LEGACY_SEND_MAIL_CALLBACK = Callable[[str], Awaitable]
+ON_LEGACY_RENEW_CALLBACK = Callable[[str], Awaitable[Tuple[bool, bool, int]]]
+ON_LEGACY_ADMIN_REQUEST = Callable[[Request], Awaitable]
```
richvdh, line 36:
```suggestion
# Temporary hooks to allow for a transition from `/_matrix/client` endpoints
# to `/_synapse/client/account_validity`. See `register_account_validity_callbacks`.
ON_LEGACY_SEND_MAIL_CALLBACK = Callable[[str], Awaitable]
ON_LEGACY_RENEW_CALLBACK = Callable[[str], Awaitable[Tuple[bool, bool, int]]]
ON_LEGACY_ADMIN_REQUEST = Callable[[Request], Awaitable]
```

richvdh (changes requested):

docs/modules.md
```diff
@@ -188,6 +193,36 @@ async def check_media_file_for_spam(
...
+can return a `bool` to indicate whether the user has expired and should be locked out of
+their account, or `None` if the module wasn't able to figure it out. The user is
+represented by their Matrix user ID (e.g. `@alice:example.com`).
+
+If the module returns `True`, the current request will be denied with the error code
+`ORG_MATRIX_EXPIRED_ACCOUNT` and the HTTP status code 403. Note that this doesn't
+invalidate the user's access token.
+
+```python
+async def on_user_registration(user: str)
```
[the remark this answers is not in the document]
babolivier: I feel like this will mislead people into thinking the function _has_ to end with `return None`

synapse/api/auth.py
```diff
@@ -75,6 +75,7 @@ def __init__(self, hs: "HomeServer"):
         self.clock = hs.get_clock()
         self.store = hs.get_datastore()
         self.state = hs.get_state_handler()
+        self._account_validity_handler = hs.get_account_validity_handler()
```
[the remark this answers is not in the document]
babolivier: Also worth noting that this isn't a new issue since the line just above gets the state handler (and the account validity handler was already called in this file before this change, the PR is merely moving where we get it from the `hs` object).

synapse/handlers/account_validity.py
babolivier: I think it would be confusing from a user pov because e.g. if they receive an email but click on the link 1w later (e.g. because they were on holidays) and the switch from the core code to the module happens in between, the link wouldn't be doing anything (since it would only update the old database table). The only other way to mitigate this that I can think of would be to have both the core code and the module enabled at the same time, but that'd mean users would get 2 emails with 2 different links to click on at about the same time, which is probably quite confusing.

synapse/handlers/account_validity.py
babolivier: I assume this has already been fixed by https://github.com/matrix-org/synapse/pull/9884/commits/f8754fed56e93de206c561877b6fdfc60fa9191c

synapse/module_api/errors.py
```diff
@@ -14,5 +14,9 @@
 
 """Exception types which are exposed as part of the stable module API"""
 
-from synapse.api.errors import RedirectException, SynapseError  # noqa: F401
+from synapse.api.errors import (  # noqa: F401
+    InvalidClientCredentialsError,
```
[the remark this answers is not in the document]
babolivier: Because `get_user_by_req` raises it if the token is invalid or doesn't belong to an existing user. Which makes me realise the docstring for the `get_user_by_req` method on the module API is lying, will fix.

richvdh:
fixes #10139

docs/modules.md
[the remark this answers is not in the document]
richvdh: if they do think that, it isn't the end of the world.

Right now it just looks like we've forgotten to document the return type, when compared with the other callback functions.

synapse/api/auth.py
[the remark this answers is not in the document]
richvdh: I'm not seeing that, tbh. But it doesn't really matter.

synapse/handlers/account_validity.py
richvdh: I'm not following you. We can handle `/_synapse/admin/v1/account_validity/validity` in the core while still having emails link to `/_synapse/client/account_validity/validity` ?

synapse/module_api/__init__.py
```diff
@@ -159,18 +159,16 @@ def public_room_list_manager(self):
...
-    def email_app_name(self):
-        """Allow accessing the application name configured in the homeserver's
-        configuration.
-        """
+    def email_app_name(self) -> str:
+        """The application name configured in the homeserver's configuration."""
         return self._hs.config.email.email_app_name
 
-    def get_user_by_req(
+    async def get_user_by_req(
```
richvdh, line 171: it'd be nice to give this a return type annotation while you're changing it.

synapse/module_api/__init__.py
```diff
@@ -556,16 +583,19 @@ def looping_background_call(
...
 
         Args:
-            f: The function to call repeatedly.
+            f: The function to call repeatedly. f can be either synchronous or
+                asynchronous, and must follow Synapse's logcontext rules.
+                More info about logcontexts is available at
+                https://matrix-org.github.io/synapse/latest/log_contexts.html
             msec: How long to wait between calls in milliseconds.
             desc: The background task's description. Default to the function's name.
             *args: Positional arguments to pass to function.
```
richvdh:
```suggestion
            *args: Positional arguments to pass to function.
            desc: The background task's description. Defaults to the function's name.

```

richvdh (changes requested):
see the points above, also the tiny nit below.

synapse/api/auth.py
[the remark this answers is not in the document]
babolivier: Oh yeah I was mixing up things, mb

synapse/handlers/account_validity.py
babolivier: We can, but then we'd end up with two different tables (`account_validity`, and `email_account_validity` from https://github.com/matrix-org/synapse-email-account-validity) used for the same purpose but with divergent data.

richvdh: ugh, I hadn't realised the new module was using a different table. Can it not use the same table?

babolivier: I'm a bit wary of having an external module write to a table that belongs to Synapse's schema to be honest. Especially since it's tied to a feature we'll be getting rid of eventually; having an external module use it means we would need extra care deleting the feature and its storage component and imho it's very easy to just forget it and only remember when it breaks e.g. on a customer's infrastructure.

Plus the naming wouldn't be right, since it then wouldn't just be account validity, but rather one way of doing account validity (the email-based one), and other modules might implement other ways to verify the validity of an account (e.g. by checking with an auth backend).

richvdh: hmmm yes, fair enough.

synapse/handlers/account_validity.py
```diff
@@ -70,6 +80,96 @@ def __init__(self, hs: "HomeServer"):
...
+        if is_user_expired is not None:
+            self._is_user_expired_callbacks.append(is_user_expired)
+
+        if on_user_registration is not None:
+            self._on_user_registration_callbacks.append(on_user_registration)
+
+        # The builtin account validity feature exposes 3 endpoints (send_mail, renew, and
+        # an admin one). As part of moving the feature into a module, we need to change
+        # the path from /_matrix/client/unstable/account_validity/... to
+        # /_synapse/client/account_validity, because:
```
richvdh, line 113: let's add some space for readability:
```suggestion
        # /_synapse/client/account_validity, because:
        #
```

synapse/handlers/account_validity.py
```diff
@@ -70,6 +80,96 @@ def __init__(self, hs: "HomeServer"):
...
+        #   * the feature isn't part of the Matrix spec thus shouldn't live under /_matrix
+        #   * because of the way we register servlets modules can't register resources
```
richvdh, line 115:
```suggestion
        #   * the way we register servlets means that modules can't register resources
```

synapse/handlers/account_validity.py
```diff
@@ -70,6 +80,96 @@ def __init__(self, hs: "HomeServer"):
...
+        #     under /_matrix/client
```
richvdh, line 116:
```suggestion
        #     under /_matrix/client
        #
```

synapse/handlers/account_validity.py
```diff
@@ -70,6 +80,96 @@ def __init__(self, hs: "HomeServer"):
...
+        # We need to allow for a transition period between the old and new endpoints
+        # in order to allow for clients to update (and for emails to be processed).
+        # Therefore we need to allow modules (in practice just the one implementing the
+        # email-based account validity) to temporarily hook into the legacy endpoint so we
```
richvdh, line 120:
```suggestion
        # email-based account validity) to temporarily hook into the legacy endpoints so we
```

synapse/handlers/account_validity.py
```diff
@@ -218,6 +329,11 @@ async def renew_account(self, renewal_token: str) -> Tuple[bool, bool, int]:
               * An int representing the user's expiry timestamp as milliseconds since the
                 epoch, or 0 if the token was invalid.
         """
+        # If a module supports triggering a renew from here, do that, otherwise do the
```
richvdh, line 330: it'd be nice to update the docstring on this method to give some allusion to the fact this is to support the legacy renewal endpoint.

synapse/handlers/account_validity.py
```diff
@@ -70,6 +80,96 @@ def __init__(self, hs: "HomeServer"):
...
+        # The builtin account validity feature exposes 3 endpoints (send_mail, renew, and
+        # an admin one). As part of moving the feature into a module, we need to change
+        # the path from /_matrix/client/unstable/account_validity/... to
+        # /_synapse/client/account_validity, because:
+        #   * the feature isn't part of the Matrix spec thus shouldn't live under /_matrix
+        #   * because of the way we register servlets modules can't register resources
+        #     under /_matrix/client
+        # We need to allow for a transition period between the old and new endpoints
+        # in order to allow for clients to update (and for emails to be processed).
+        # Therefore we need to allow modules (in practice just the one implementing the
```
richvdh, line 119:
```suggestion
        #
        # Once the email-account-validity module is loaded, it will take control of account
        # validity by moving the rows from our `account_validity` table into its own table.
        # 
        # Therefore, we need to allow modules (in practice just the one implementing the
```

richvdh (approved):
lgtm otherwise

PR #9885, opened 2021-04-26
Add type hints to presence handler

erikjohnston:
Because type hints are Good™

### DIFF
--- a/synapse/handlers/presence.py
+++ b/synapse/handlers/presence.py
@@ -28,6 +28,7 @@ from bisect import bisect
 from contextlib import contextmanager
 from typing import (
     TYPE_CHECKING,
+    Callable,
     Collection,
     Dict,
     FrozenSet,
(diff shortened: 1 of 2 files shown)


synapse/handlers/presence.py
```diff
@@ -348,19 +349,19 @@ def __init__(self, hs):
...
+    def send_user_sync(self, user_id: str, is_syncing: bool, last_sync_ms: int) -> None:
         if self._presence_enabled:
             self.hs.get_tcp_replication().send_user_sync(
                 self.instance_id, user_id, is_syncing, last_sync_ms
             )
 
-    def mark_as_coming_online(self, user_id):
+    def mark_as_coming_online(self, user_id: str) -> None:
         """A user has started syncing. Send a UserSync to the presence writer,
         unless they had recently stopped syncing.
```
anoadragon453, line 367: There's still a `(str)` hanging around in the docstring here.

synapse/handlers/presence.py
```diff
@@ -373,7 +374,7 @@ def mark_as_coming_online(self, user_id):
             # were offline
             self.send_user_sync(user_id, True, self.clock.time_msec())
 
-    def mark_as_going_offline(self, user_id):
+    def mark_as_going_offline(self, user_id: str) -> None:
         """A user has stopped syncing. We wait before notifying the presence
         writer as its likely they'll come back soon. This allows us to avoid
         sending a stopped syncing immediately followed by a started syncing
```
anoadragon453, line 380: And here.

synapse/handlers/presence.py
```diff
@@ -1742,17 +1761,23 @@ def handle_timeout(state, is_mine, syncing_user_ids, now):
...
-        wheel_timer (WheelTimer)
-        now (int): Time now in ms
+        prev_state
+        new_state
+        is_mine: Whether the user is ours
+        wheel_timer
+        now: Time now in ms
 
     Returns:
         3-tuple: `(new_state, persist_and_notify, federation_ping)` where:
```
anoadragon453, line 1783: Some `(bool)`s still hanging around in the return docstring.

synapse/handlers/presence.py
```diff
@@ -329,7 +330,7 @@ def __init__(self, hs):
 
         # user_id -> last_sync_ms. Lists the users that have stopped syncing but
         # we haven't notified the presence writer of that yet
-        self.users_going_offline = {}
+        self.users_going_offline = {}  # type: Dict[str, int]
```
anoadragon453, line 333: Is there a reason not to write the type as `self.users_going_offline: Dict[str, int] = {}` other than consistency?

anoadragon453 (approved):
Minor things, but looks good overall. Thanks a lot for doing this!

synapse/handlers/presence.py
erikjohnston: Nope, I keep forgetting that is a thing we are allowed to do now

PR #9886, opened 2021-04-26
Reduce memory footprint of caches

erikjohnston:
This removes the set we store in each node, and replaces it with either `None` if empty or a list. Since we expect there to be only a couple of callbacks per entry the cost of deduplicating the list is likely the same (or faster!) than maintaining a set, and has a smaller memory footprint.

The size of storing `None` is basically 0, an empty list is 56 bytes, and an empty set is 248 bytes. 

It doesn't look like it's had much effect on jki.re, but that might be because my caches aren't that large :shrug:

### DIFF
--- a/synapse/util/caches/lrucache.py
+++ b/synapse/util/caches/lrucache.py
@@ -73,11 +73,13 @@ class _Node:
 
         # Set of callbacks to run when the node gets deleted. We store as a list
         # rather than a set to keep memory usage down (and since we expect few
-        # entries per node, the performance of checking for duplication in a list
-        # vs using a set is negligible).
+        # entries per node, the performance of checking for duplication in a
+        # list vs using a set is negligible).
         #
         # Note that we store this as an optional list to keep the memory
-        # footprint down. Empty lists are 56 bytes (and empty sets are 216 bytes).
+        # footprint down. Storing `None` is free as its a singleton, while empty
+        # lists are 56 bytes (and empty sets are 216 bytes, if we did the naive
+        # thing and used sets).
         self.callbacks = None  # type: Optional[List[Callable[[], None]]]
 
         self.add_callbacks(callbacks)
--- a/synapse/util/caches/lrucache.py
+++ b/synapse/util/caches/lrucache.py
@@ -73,7 +73,7 @@ class _Node:
 
         # Set of callbacks to run when the node gets deleted. We store as a list
         # rather than a set to keep memory usage down (and since we expect few
-        # entries per node the performance of checking for duplication in a list
+        # entries per node, the performance of checking for duplication in a list
         # vs using a set is negligible).
         #
         # Note that we store this as an optional list to keep the memory
@@ -96,7 +96,7 @@ class _Node:
                 self.callbacks.append(callback)
 
     def run_and_clear_callbacks(self) -> None:
-        """Run all callbacks and clear the stored set of callbacks. Used when
+        """Run all callbacks and clear the stored list of callbacks. Used when
         the node is being deleted.
         """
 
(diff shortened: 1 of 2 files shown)


callahad:
> It doesn't look like it's had much effect on jki.re

Do we have reason to believe it would have a meaningful effect on larger instances?

erikjohnston:
> > It doesn't look like it's had much effect on jki.re
> 
> Do we have reason to believe it would have a meaningful effect on larger instances?

I think it will a bit, but its hard to know really.

synapse/util/caches/lrucache.py
```diff
@@ -57,13 +59,54 @@ class _Node:
...
     ):
         self.prev_node = prev_node
         self.next_node = next_node
         self.key = key
         self.value = value
-        self.callbacks = callbacks or set()
+
+        # Set of callbacks to run when the node gets deleted. We store as a list
+        # rather than a set to keep memory usage down (and since we expect few
+        # entries per node the performance of checking for duplication in a list
```
anoadragon453, line 76:
```suggestion
        # entries per node, the performance of checking for duplication in a list
```

Took me a couple reads to grok this sentence :)

synapse/util/caches/lrucache.py
```diff
@@ -57,13 +59,54 @@ class _Node:
...
+        # vs using a set is negligible).
+        #
+        # Note that we store this as an optional list to keep the memory
+        # footprint down. Empty lists are 56 bytes (and empty sets are 216 bytes).
+        self.callbacks = None  # type: Optional[List[Callable[[], None]]]
```
anoadragon453, line 81: We should probably be talking about the memory usage of empty lists versus `None` here, instead of vs. Sets?

synapse/util/caches/lrucache.py
```diff
@@ -57,13 +59,54 @@ class _Node:
...
+
+        if not self.callbacks:
+            self.callbacks = []
+
+        for callback in callbacks:
+            if callback not in self.callbacks:
+                self.callbacks.append(callback)
+
+    def run_and_clear_callbacks(self) -> None:
+        """Run all callbacks and clear the stored set of callbacks. Used when
```
anoadragon453, line 99:
```suggestion
        """Run all callbacks and clear the stored list of callbacks. Used when
```

synapse/util/caches/lrucache.py
```diff
@@ -57,13 +59,54 @@ class _Node:
...
+        the node is being deleted.
+        """
+
+        if not self.callbacks:
+            return
```
anoadragon453, line 104: Is there a reason to keep an empty list versus going back to `None`?

anoadragon453 (changes requested):
Looks good for the most part. Thanks for adding some explanatory comments - they really make all the difference when you're not familiar with a class.

synapse/util/caches/lrucache.py
erikjohnston: I'll clarify, but `None` as a singleton is basically free. The comparison with an empty set is because using sets is the "obvious" way of doing it.

synapse/util/caches/lrucache.py
erikjohnston: At this point `self.callbacks` should either by `None` or a non empty list.

synapse/util/caches/lrucache.py
anoadragon453: Right, OK. I read it as a justification for why you're using an optional list vs. just a list, rather than an optional list vs. a set :)

anoadragon453 (approved):
lgtm!

PR #9887, opened 2021-04-26
Use current state table for `presence.get_interested_remotes`

erikjohnston:
This should be a lot quicker than asking the state handler.


anoadragon453 (approved):
lgtm!

I only now realise the terminology confusion between presence states and state events :D

PR #9889, opened 2021-04-26
Delete room endpoint

### LINKED ISSUE #9032: The Delete Room API should use the http DELETE method, not POST
With roots in [this suggestion](https://github.com/matrix-org/synapse/pull/7613#discussion_r438859896) by @richvdh, the current Delete Room API uses `POST /_synapse/admin/v1/rooms/<room_id>/delete` instead of `DELETE /_synapse/admin/v1/rooms/<room_id>`, on the logic that DELETE should not have a request body.

This logic is incorrect, as better detailed by [this stackoverflow answer](https://stackoverflow.com/a/299696), but in short, the spec supports it, and the matrix APIs already have other instances of `DELETE` being used with a request body.
https://matrix.org/docs/spec/client_server/latest#delete-matrix-client-r0-devices-deviceid

ThibF:
Fix #9032 by marking as deprecated the previous endpoint and add a new DELETE endpoint for a room

### DIFF
--- a/synapse/rest/admin/rooms.py
+++ b/synapse/rest/admin/rooms.py
@@ -282,7 +282,22 @@ class ListRoomRestServlet(RestServlet):
 
 
 class RoomRestServlet(RestServlet):
-    """Get room details.
+    """Manage a room.
+
+    On GET : Get details of a room.
+
+    On DELETE : Delete a room from server.
+
+    It is a combination and improvement of shutdown and purge room.
+
+    Shuts down a room by removing all local users from the room.
+    Blocking all future invites and joins to the room is optional.
+
+    If desired any local aliases will be repointed to a new room
+    created by `new_room_user_id` and kicked users will be auto-
+    joined to the new room.
+
+    If 'purge' is true, it will remove all traces of a room from the database.
 
     TODO: Add on_POST to allow room creation without joining the room
     """
@@ -293,6 +308,8 @@ class RoomRestServlet(RestServlet):
         self.hs = hs
         self.auth = hs.get_auth()
         self.store = hs.get_datastore()
+        self.room_shutdown_handler = hs.get_room_shutdown_handler()
+        self.pagination_handler = hs.get_pagination_handler()
 
     async def on_GET(
         self, request: SynapseRequest, room_id: str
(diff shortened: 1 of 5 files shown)


ThibF:
I'm not 100% convinced by the addition of the `_delete_room` function, but as there is 2 different endpoints with 2 different methods involved, it would cumbersome to group under the same Servlet.

synapse/rest/admin/rooms.py
```diff
@@ -694,3 +686,55 @@ async def on_GET(
...
 
         return 200, results
+
+
+async def _delete_room(
+    request: SynapseRequest,
+    room_id: str,
+    auth,
+    room_shutdown_handler,
+    pagination_handler,
```
erikjohnston, line 696: We may as well fully type this function (you'll also need to import things appropriately):

```suggestion
    auth: Auth,
    room_shutdown_handler: RoomShutdownHandler,
    pagination_handler: PaginationHandler,
```

FTR you can see what these types are by looking in `synapse/server.py`

tests/rest/admin/test_room.py
```diff
@@ -175,15 +175,15 @@ def prepare(self, reactor, clock, hs):
         self.room_id = self.helper.create_room_as(
             self.other_user, tok=self.other_user_tok
         )
-        self.url = "/_synapse/admin/v1/rooms/%s/delete" % self.room_id
```
erikjohnston, line 178: I think we should test both the deprecated and new methods, as we don't want to accidentally break the old endpoint.

I think we can use `@parameterized_class` decorator to avoid duplicating code, so something like:

```python
@parameterized_class(
    ("method", "url_template"),
    [("POST", "/_synapse/admin/v1/rooms/%s/delete"), ("DELETE", "/_synapse/admin/v1/rooms/%s")]
)
class DeleteRoomTestCase(unittest.HomeserverTestCase):
    ...
``` 

See https://github.com/wolever/parameterized for some docs and https://github.com/matrix-org/synapse/blob/develop/tests/rest/media/v1/test_media_storage.py#L130 as an example.

erikjohnston (changes requested):
Looks good, thanks! Just a couple of nits

erikjohnston:
Thanks for making the changes! Just as a small thing for next time: can you not rebase/force push changes please? That way it's easier to see the changes that were made since the last review, and saves us from having to re-review the entire thing :slightly_smiling_face:

erikjohnston (approved):
Awesome, thank you!

ThibF:
Will do, I have the habit to try to keep the git history minimal.
Thanks for the friendly/productive review !

shadowcat-mst:
Please see https://github.com/matrix-org/synapse/issues/11698 and any further discussion on https://github.com/matrix-org/synapse/issues/9032 for why this is dangerous in practice and the original API was both more robust and more in keeping with RESTful principles.

(sorry, nothing wrong with the code, but the original idea was arguably one that seemed like a good idea at the time but really isn't :/)

clokep:
@shadowcat-mst Thank you for opening #11698 -- please don't comment on every issue / PR that might have been related. Let's keep the conversation in a single place (#11698). Thank you!

shadowcat-mst:
@clokep sorry, was merely trying to make sure that anybody who was already subscribed to the other ones involved knew about #11698 specifically *so* we could centralise the conversation there.

So, uh, I was actually trying to achieve what you just asked me to do, I just apparently picked the wrong way to attempt to implement that :/

Issue #9641, opened 2021-03-17
Fix remaining uses of deprecated cmp argument in attrs decorators


ulope:
### Description

There are a handful of places left that use the deprecated `cmp` argument on the `attr.s` decorator:
- https://github.com/matrix-org/synapse/blob/0a00b7ff14890987f09112a2ae696c61001e6cf1/synapse/events/builder.py#L38
- https://github.com/matrix-org/synapse/blob/d29b71aa50f568c3b951b95f2ea3009292b95630/tests/server.py#L472
- https://github.com/matrix-org/synapse/blob/a7a379006651ea219c2618c0a47e8f4053a9e769/synapse/types.py#L421
- https://github.com/matrix-org/synapse/blob/31b1905e13f84d839f46990d7e63bafe0917dc75/synapse/crypto/keyring.py#L65

These should be fixed.

clokep:
Any idea what they should be replaced? Would you be willing to make a PR making this change?

lakshay1121:
Is this issue still open ?

clokep:
> Is this issue still open ?

It looks like some of the instances have been replaced, but not all. I still see:

```
synapse/crypto/keyring.py:@attr.s(slots=True, frozen=True, cmp=False, auto_attribs=True)
synapse/events/builder.py:@attr.s(slots=True, cmp=False, frozen=True, auto_attribs=True)
```

#9892 was a start on this, but was missing bumping the minimum version of attrs.

lakshay1121:
okk fine then.

PR #9892, opened 2021-04-27
Remove usage of deprecated attrs cmp argument

ulope:
The `cmp` argument to the `attr.s` decorator [is deprecated](https://www.attrs.org/en/stable/api.html#attr.s).
This replaces it with `eq` and `order` as suggested in the attrs docs.

Fixes: #9641
Signed-off-by: Ulrich Petri <github@ulo.pe>

### Pull Request Checklist

### DIFF
--- a/changelog.d/9892.misc
+++ b/changelog.d/9892.misc
@@ -1,0 +1,1 @@
+Remove usage of deprecated `cmp` argument in the `attr.s` decorator.
--- a/synapse/crypto/keyring.py
+++ b/synapse/crypto/keyring.py
@@ -61,7 +61,7 @@ if TYPE_CHECKING:
 logger = logging.getLogger(__name__)
 
 
-@attr.s(slots=True, cmp=False)
+@attr.s(slots=True, eq=False, order=False)
 class VerifyJsonRequest:
     """
     A request to verify a JSON object.
--- a/synapse/events/builder.py
+++ b/synapse/events/builder.py
@@ -34,7 +34,7 @@ from synapse.util import Clock
 from synapse.util.stringutils import random_string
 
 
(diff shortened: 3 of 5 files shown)


anoadragon453:
Looks like the version of `attrs` that we currently support back to (`attrs==19.1.0`) is too old to understand `eq`.

Flipping through the documentation, it looks like `eq` and `order` were added in 19.2.0. Do we have any objection to bumping `attrs` to 19.2.0 (or even 19.3.0) to support this?

@ulope Feel free to bump the version of `attrs` to check that it fixes CI:

https://github.com/matrix-org/synapse/blob/ad8690a26c891a8e3444a3e80bdcdb3ba0d6a02b/synapse/python_dependencies.py#L81

ulope:
There's a comment above the attrs require:
```
    # we use attr.validators.deep_iterable, which arrived in 19.1.0 (Note:
    # Fedora 31 only has 19.1, so if we want to upgrade we should wait until 33
    # is out in November.)
```

From the commit timestamp I assume this is talking about November 2020. 
So this means it should be fine to update?

anoadragon453:
Indeed that sounds promising @ulope. Let me just double-check with our distro packagers that bumping the dependency would be alright :slightly_smiling_face:

anoadragon453:
Looks like downstream packages should work with attrs 19.3.0! So feel free to update attrs to that in your PR to fix CI.

Thanks for your patience!

anoadragon453 (changes requested):
To upgrade attrs.

richvdh:
Worth noting that attr.s 21.1 *un*deprecates `cmp`, but this is probably still worthwhile

richvdh:
looks like this is starting to bitrot, so I'm going to close it.

PR #9895, opened 2021-04-27
Use the parent's logging context name for runWithConnection.

### LINKED ISSUE #9893: Transaction logging doesn't contain request ID
It instead logs with `runWithConnection`.

clokep:
Fixes #9893 hopefully. This should now log the connection ID for log lines.

I audited places where we create `LoggingContext` (again), but the rest seemed OK? Would appreciate a double checking though.

### DIFF
--- a/synapse/storage/database.py
+++ b/synapse/storage/database.py
@@ -718,7 +718,9 @@ class DatabasePool:
             # pool).
             assert not self.engine.in_transaction(conn)
 
-            with LoggingContext("runWithConnection", parent_context) as context:
+            with LoggingContext(
+                str(curr_context), parent_context=parent_context
+            ) as context:
                 sched_duration_sec = monotonic_time() - start_time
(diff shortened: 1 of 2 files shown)


synapse/storage/database.py
```diff
@@ -718,7 +718,9 @@ def inner_func(conn, *args, **kwargs):
             # pool).
             assert not self.engine.in_transaction(conn)
 
-            with LoggingContext("runWithConnection", parent_context) as context:
+            with LoggingContext(
+                str(curr_context), parent_context=parent_context
```
clokep, line 722: `str(curr_context)` will either be `parent_context.name` or `"sentinel"`. We do this in a few other spots so seems reasonable to do here too.

We could add a suffix, but in that case should be using `nested_logging_context`. 🤷

erikjohnston: I also couldn't see any other `LoggingContext` that need to be changed

PR #9896, opened 2021-04-27
Add missing type hints to handlers.

clokep:
With #9885 I think that all of `synapse.handlers` has type hints now.

Should be reviewable commit by commit, I'll leave some inline comments.

### DIFF
--- a/synapse/handlers/directory.py
+++ b/synapse/handlers/directory.py
@@ -14,7 +14,7 @@
 
 import logging
 import string
-from typing import Iterable, List, Optional
+from typing import TYPE_CHECKING, Iterable, List, Optional
 
 from synapse.api.constants import MAX_ALIAS_LENGTH, EventTypes
 from synapse.api.errors import (
@@ -27,15 +27,19 @@ from synapse.api.errors import (
     SynapseError,
 )
 from synapse.appservice import ApplicationService
-from synapse.types import Requester, RoomAlias, UserID, get_domain_from_id
+from synapse.storage.databases.main.directory import RoomAliasMapping
+from synapse.types import JsonDict, Requester, RoomAlias, UserID, get_domain_from_id
 
 from ._base import BaseHandler
+
+if TYPE_CHECKING:
+    from synapse.server import HomeServer
 
 logger = logging.getLogger(__name__)
 
 
 class DirectoryHandler(BaseHandler):
-    def __init__(self, hs):
+    def __init__(self, hs: "HomeServer"):
         super().__init__(hs)
 
         self.state = hs.get_state_handler()
@@ -60,7 +64,7 @@ class DirectoryHandler(BaseHandler):
         room_id: str,
         servers: Optional[Iterable[str]] = None,
         creator: Optional[str] = None,
-    ):
+    ) -> None:
         # general association creation for both human users and app services
 
         for wchar in string.whitespace:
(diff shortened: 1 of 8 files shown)


synapse/handlers/directory.py
```diff
@@ -133,12 +138,12 @@ async def create_association(
                     )
 
             if not await self.spam_checker.user_may_create_room_alias(
-                user_id, room_alias
+                user_id, room_alias_str
```
clokep, line 141: This was a real bug as far as I can tell. I think it has always existed?

synapse/handlers/message.py
```diff
@@ -91,7 +91,7 @@ async def get_room_data(
         room_id: str,
         event_type: str,
         state_key: str,
-    ) -> dict:
+    ) -> Optional[EventBase]:
```
clokep, line 94: This seems to return `EventBase`, and the caller calls methods on it that don't exist on `dict`. 🤷

synapse/handlers/ui_auth/checkers.py
```diff
@@ -57,10 +60,10 @@ async def check_auth(self, authdict: dict, clientip: str) -> Any:
 class DummyAuthChecker(UserInteractiveAuthChecker):
     AUTH_TYPE = LoginType.DUMMY
 
-    def is_enabled(self):
+    def is_enabled(self) -> bool:
         return True
 
-    async def check_auth(self, authdict, clientip):
+    async def check_auth(self, authdict: dict, clientip: str) -> Any:
```
clokep, line 66: These are somewhat lame, but match what `UserInteractiveAuthChecker` already has.

synapse/handlers/directory.py
erikjohnston: Hum, I wonder if any spam checkers in the wild assume its a `RoomAlias`?

erikjohnston (commented):
LGTM, other than we need to figure out what to do about passing the wrong type to the spam checker

synapse/handlers/directory.py
clokep: I wasn't sure what to do here, it's possible of course.

Mjolnir doesn't use this method: https://github.com/matrix-org/mjolnir/blob/master/synapse_antispam/mjolnir/antispam.py#L121-L122 (although there's a PR for a rewrite which does: https://github.com/matrix-org/mjolnir/pull/78/files#diff-abf5313147c56e3bf705f659a3b8535242f30744087be63343ee8281d43ded2bR131-R142).

🤷

erikjohnston: I'm hesitant to change behaviour here, in case it *is* used. I wonder if the less risky thing to do is to change the docs to say that its a string?

clokep: > I wonder if the less risky thing to do is to change the docs to say that its a string?

Do you mean to change the docs to say it is a `RoomAlias`?

The [docs](https://github.com/matrix-org/synapse/blob/develop/docs/spam_checker.md#python-spam-checker-class) don't really say anything, it would be updating the type hint on the spam checker, I think.

erikjohnston: Yes, sorry, I got them the wrong way round. If the docs don't say anything then I think we should just update the type hint. Either someone is using it already and changing it will explode, or no one is using it and there's no harm changing it?

clokep: That makes sense to me. 👍

erikjohnston (approved):

PR #9902, opened 2021-04-28
Limit how often GC happens by time.

### LINKED ISSUE #9890: Limit how often we do a GC.
We support configuring the GC thresholds via the [`gc_thresholds`](https://github.com/matrix-org/synapse/blob/develop/docs/sample_config.yaml#L151-L153) config option, however the default can lead to a lot of GC churn on busier instances. This can easily lead to a) a CPU usage by GC of ~20% and b) multiple multi second GC pauses in quick succession.

Since we manually run the GC we could add limits to ensure we don't run the GC e.g. gen 0 no more than once a second, gen 1 no more than once every 10s and gen 2 no more than once every 30s. This a) enforces that we don't repeatedly do gen2 really frequently and b) stops objects getting promoted to gen2 far too quickly.

erikjohnston:
Synapse can be quite memory intensive, and unless care is taken to tune
the GC thresholds it can end up thrashing, causing noticable performance
problems for large servers. We fix this by limiting how often we GC a
given generation, regardless of current counts/thresholds.

This does not help with the reverse problem where the thresholds are set
too high, but that should only happen in situations where they've been
manually configured.

Adds a `gc_min_seconds_between` config option to override the defaults.

Fixes #9890.

### DIFF
--- a/docs/sample_config.yaml
+++ b/docs/sample_config.yaml
@@ -152,6 +152,14 @@ presence:
 #
 #gc_thresholds: [700, 10, 10]
 
+# The minimum time in seconds between each GC for a generation, regardless of
+# the GC thresholds. This ensures that we don't do GC too frequently.
+#
+# A value of `[1, 10, 30]` indicates that a second must pass between consecutive
+# generation 0 GCs, etc.
+#
+# gc_min_seconds_between: [1, 10, 30]
+
 # Set the limit on the returned events in the timeline in the get
 # and sync operations. The default value is 100. -1 means no upper limit.
 #
--- a/docs/sample_config.yaml
+++ b/docs/sample_config.yaml
@@ -155,10 +155,10 @@ presence:
 # The minimum time in seconds between each GC for a generation, regardless of
 # the GC thresholds. This ensures that we don't do GC too frequently.
 #
-# A value of `[1, 10, 30]` indicates that a second must pass between consecutive
+# A value of `[1s, 10s, 30s]` indicates that a second must pass between consecutive
 # generation 0 GCs, etc.
 #
-# gc_min_seconds_between: [1, 10, 30]
+#gc_min_interval: [0.5s, 30s, 1m]
 
 # Set the limit on the returned events in the timeline in the get
 # and sync operations. The default value is 100. -1 means no upper limit.
--- a/docs/sample_config.yaml
+++ b/docs/sample_config.yaml
@@ -158,6 +158,8 @@ presence:
 # A value of `[1s, 10s, 30s]` indicates that a second must pass between consecutive
 # generation 0 GCs, etc.
 #
+# Defaults to `[1s, 10s, 30s]`.
+#
 #gc_min_interval: [0.5s, 30s, 1m]
 
 # Set the limit on the returned events in the timeline in the get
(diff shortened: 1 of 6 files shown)


synapse/metrics/__init__.py
```diff
@@ -575,11 +582,16 @@ def f(*args, **kwargs):
...
+        # promote an object into gen 2, and we don't want to handle the same
+        # object multiple times.
         threshold = gc.get_threshold()
         counts = gc.get_count()
         for i in (2, 1, 0):
-            if threshold[i] < counts[i]:
+            # We check if we need to do one based on a straightforward
+            # comparison between the threshold and count. We also do an extra
+            # check to make sure that we don't a GC too often.
+            if threshold[i] < counts[i] and MIN_TIME_BETWEEN_GCS[i] < end - _last_gc[i]:
```
ShadowJonathan, line 594: This can become out-of-bounds if someone specifies `[1, 10]` or similar accidentally.

```suggestion
            if threshold[i] < counts[i] and len(MIN_TIME_BETWEEN_GCS) => i and MIN_TIME_BETWEEN_GCS[i] < end - _last_gc[i]:
```

I'm also wary of mutating a "constant" like this, because it can give mixed signals in the code.

erikjohnston: Yeah, mutating is a bit icky, but I can't think of a better way of injecting the values at this point.

> This can become out-of-bounds if someone specifies [1, 10] or similar accidentally.

We check for the right format in the config, so I think its fine that this explodes if a developer messes up and sets it incorrectly.

docs/sample_config.yaml
```diff
@@ -155,10 +155,10 @@ presence:
 # The minimum time in seconds between each GC for a generation, regardless of
 # the GC thresholds. This ensures that we don't do GC too frequently.
 #
-# A value of `[1, 10, 30]` indicates that a second must pass between consecutive
+# A value of `[1s, 10s, 30s]` indicates that a second must pass between consecutive
 # generation 0 GCs, etc.
 #
-# gc_min_seconds_between: [1, 10, 30]
+#gc_min_interval: [0.5s, 30s, 1m]
```
richvdh: what's the default?

synapse/config/server.py
```diff
@@ -1314,6 +1314,23 @@ def add_arguments(parser):
             help="Turn on the twisted telnet manhole service on the given port.",
         )
 
+    def read_gc_intervals(self, durations):
```
richvdh, line 1317:
```suggestion
    def read_gc_intervals(self, durations) -> Tuple[int, int, int]:
```

synapse/config/server.py
```diff
@@ -572,6 +572,7 @@ def read_config(self, config, **kwargs):
             _warn_if_webclient_configured(self.listeners)
 
         self.gc_thresholds = read_gc_thresholds(config.get("gc_thresholds", None))
+        self.gc_seconds = self.read_gc_intervals(config.get("gc_min_interval", None))
```
richvdh, line 575: `parse_duration` returns milliseconds, so I think you need to do some renaming and stuff

richvdh (changes requested):

richvdh (approved):
I don't know why we're working in seconds here where we use ms literally everwhere else, but lgtm otherwise.

erikjohnston:
> I don't know why we're working in seconds here where we use ms literally everwhere else, but lgtm otherwise.

Where we using it everything is done in seconds (as we don't have access to the standard clocks), so using milliseconds would involve a lot of repeated conversion

PR #9904, opened 2021-04-28
Time external cache response time

erikjohnston:
It looks like on matrix.org this can take a surprisingly long time.

### DIFF
--- a/synapse/replication/tcp/external_cache.py
+++ b/synapse/replication/tcp/external_cache.py
@@ -36,7 +36,7 @@ get_counter = Counter(
 )
 
 response_timer = Histogram(
-    "synapse_external_cache_response_time",
+    "synapse_external_cache_response_time_seconds",
     "sec",
     labelnames=["method"],
     buckets=(
--- a/synapse/replication/tcp/external_cache.py
+++ b/synapse/replication/tcp/external_cache.py
@@ -15,7 +15,7 @@
 import logging
 from typing import TYPE_CHECKING, Any, Optional
 
-from prometheus_client import Counter
+from prometheus_client import Counter, Histogram
 
 from synapse.logging.context import make_deferred_yieldable
 from synapse.util import json_decoder, json_encoder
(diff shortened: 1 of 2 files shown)


synapse/replication/tcp/external_cache.py
```diff
@@ -72,23 +86,25 @@ async def set(self, cache_name: str, key: str, value: Any, expiry_ms: int) -> No
 
         logger.debug("Caching %s %s: %r", cache_name, key, encoded_value)
 
-        return await make_deferred_yieldable(
-            self._redis_connection.set(
-                self._get_redis_key(cache_name, key),
-                encoded_value,
-                pexpire=expiry_ms,
+        with response_timer.labels("set").time():
```
clokep, line 89: Would it make sense to add the `cache_name` as a label?

erikjohnston: I don't think so, the times shouldn't depend on what we're caching (and I'm a bit cautious about the fact it adds a bunch of extra metrics). Broadly these should be super quick, and if they're not then something is wrong/congested.

synapse/replication/tcp/external_cache.py
```diff
@@ -35,6 +35,20 @@
     labelnames=["cache_name", "hit"],
 )
 
+response_timer = Histogram(
+    "synapse_external_cache_response_time",
+    "sec",
```
richvdh, line 40: better documentation please

synapse/replication/tcp/external_cache.py
```diff
@@ -35,6 +35,20 @@
     labelnames=["cache_name", "hit"],
 )
 
+response_timer = Histogram(
+    "synapse_external_cache_response_time",
```
richvdh, line 39:
```suggestion
    "synapse_external_cache_response_time_seconds",
```

richvdh (changes requested):

synapse/replication/tcp/external_cache.py
```diff
@@ -36,8 +36,8 @@
 )
 
 response_timer = Histogram(
-    "synapse_external_cache_response_time",
-    "sec",
+    "synapse_external_cache_response_time_seconds",
+    "Time taken to get a response from Redis for a request",
```
richvdh, line 40:
```suggestion
    "Time taken to get a response from Redis for a cache get/set request",
```

or something

richvdh (approved):
lgtm

PR #9905, opened 2021-04-28
Don't set the external cache if its been done recently

erikjohnston:
This may shave some time off event sending and make Redis less busy.

### DIFF
--- a/changelog.d/9905.misc
+++ b/changelog.d/9905.misc
@@ -1,0 +1,1 @@
+Don't set the external cache for "joined hosts" if its been done recently.
--- a/changelog.d/9905.misc
+++ b/changelog.d/9905.misc
@@ -1,1 +1,0 @@
-Don't set the external cache for "joined hosts" if its been done recently.
--- a/synapse/handlers/message.py
+++ b/synapse/handlers/message.py
@@ -1022,6 +1022,9 @@ class EventCreationHandler:
         if not self._external_cache.is_enabled():
             return
 
+        # If external cache is enabled we should always have this.
+        assert self._external_cache_joined_hosts_updates
+
         # We actually store two mappings, event ID -> prev state group,
         # state group -> joined hosts, which is much more space efficient
         # than event ID -> joined hosts.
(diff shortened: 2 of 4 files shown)


synapse/handlers/message.py
```diff
@@ -1036,6 +1058,9 @@ async def cache_joined_hosts_for_event(self, event: EventBase) -> None:
                 expiry_ms=60 * 60 * 1000,
             )
 
+            if self._external_cache_joined_hosts_updates:
+                self._external_cache_joined_hosts_updates[context.prev_group] = None
```
richvdh, line 1062: why is this `context.prev_group` rather than `state_entry.state_group` ?

synapse/handlers/message.py
```diff
@@ -1013,16 +1029,22 @@ async def cache_joined_hosts_for_event(self, event: EventBase) -> None:
...
-        # we already set it, so that the expiry time is reset.
+        # Note: We set the state group -> joined hosts cache if it hasn't been
+        # set for a while, so that the expiry time is reset.
 
         state_entry = await self.state.resolve_state_groups_for_events(
             event.room_id, event_ids=event.prev_event_ids()
         )
 
         if state_entry.state_group:
+            if self._external_cache_joined_hosts_updates:
```
richvdh, line 1040: given that this whole method is conditional on `self._external_cache.is_enabled()`, can we just bung an `assert self._external_cache_joined_hosts_updates` up the top to save rechecking it twice?

richvdh (changes requested):

synapse/handlers/message.py
erikjohnston: They're (almost certainly) the same value, but we should use `state_entry.state_group`

changelog.d/9905.misc
```diff
@@ -0,0 +1 @@
+Don't set the external cache for "joined hosts" if its been done recently.
```
richvdh, line 1: Can we say "Improve performance of [operations] for worker-based deployments using Redis", or something?

richvdh (approved):

PR #9910, opened 2021-04-29
Use get_current_users_in_room from store and not StateHandler

erikjohnston:


changelog.d/9910.misc
```diff
@@ -0,0 +1 @@
+Improve performance after joining a large room when presence is enabled.
```
richvdh, line 1: .feature?

richvdh (approved):
lgtm. Should we get rid of `state.get_current_users_in_room`? Or at least add some warnings to it?

erikjohnston:
> lgtm. Should we get rid of `state.get_current_users_in_room`? Or at least add some warnings to it?

Yeah, we should add a warning. It is used by things that want to get the current users based on a particular set of extremities, but that is rare.

changelog.d/9910.misc
erikjohnston: Are performance improvements features? They're not new buttons per se.

richvdh: the line is obviously fuzzy, but to me the question is whether it is interesting to Synapse admins/users - in which case, it's a `.feature` - or just developers, in which case it's a `.misc`.

synapse/storage/_base.py
```diff
@@ -69,6 +69,7 @@ def _invalidate_state_caches(
             self._attempt_to_invalidate_cache("is_host_joined", (room_id, host))
 
         self._attempt_to_invalidate_cache("get_users_in_room", (room_id,))
+        self._attempt_to_invalidate_cache("get_users_in_room_with_profiles", (room_id,))
```
erikjohnston, line 72: So, err, we didn't ever invalidate this cache. I think we only ever used this function when handling visibility changes in the user directory, so I think its probably fairly unlikely that this bit anyone...

https://github.com/matrix-org/synapse/blob/c571736c6ca5d1d2d9bf7cd9b717465d446ac7b3/synapse/handlers/user_directory.py#L305

erikjohnston: I guess the question is whether we want to pull this into a separate PR?

richvdh: "meh". I woudn't bother.

You could always stick in a `.bugfix` as well as a `.feature` newsfile if you're feeling keen.

richvdh (approved):

PR #9911, opened 2021-04-30
Add port parameter to the sample config for psycopg2 args

anoadragon453:
Sometimes people run their Postgres on a non-standard port. This PR adds the `port` option with the default value to the sample config file.


changelog.d/9911.misc
```diff
@@ -0,0 +1 @@
+Add `port` argument to the Postgres database sample config section.
```
richvdh, line 1: .doc, maybe?

richvdh (approved):
lgtm

PR #8147, opened 2020-08-21
Add healthcheck for default localhost 8008 port on /health endpoint.

maquis196:
* Updated docker readme for healthcheck related options

### Pull Request Checklist

* [x] Pull request is based on the develop branch
* [x] Pull request includes a [changelog file](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md#changelog). The entry should:
  - Be a short description of your change which makes sense to users. "Fixed a bug that prevented receiving messages from other servers." instead of "Moved X method from `EventStore` to `EventWorkerStore`.".
  - Use markdown where necessary, mostly for `code blocks`.
  - End with either a period (.) or an exclamation mark (!).
  - Start with a capital letter.
* [x] Pull request includes a [sign off](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md#sign-off)
* [x] Code style is correct (run the [linters](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md#code-style))

PR #9913, opened 2021-04-30
Docker healthcheck timings - add startup delay and changed interval

maquis196:
### Pull Request Checklist

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -192,9 +192,9 @@ whilst running the above `docker run` commands.
    --no-healthcheck
 ```
 
-## Disabling the healthcheck in docker-compose
+## Disabling the healthcheck in docker-compose file
 
-If you wish to disable the healthcheck via docker-compose append your stanza with the follow.
+If you wish to disable the healthcheck via docker-compose append your stanza with the following.
 
 ```
   healthcheck:
@@ -212,7 +212,7 @@ If you wish to point the healthcheck at a different port with docker command, ad
 ## Setting the healthcheck in docker-compose file
 
 You can add the following to set a custom healthcheck in a docker compose file.
-You will need version >2.1 for this to work. 
+You will need docker-compose version >2.1 for this to work. 
 
 ```
 healthcheck:
@@ -222,7 +222,6 @@ healthcheck:
   retries: 3
   start_period: 5s
 ```
-
 
 ## Using jemalloc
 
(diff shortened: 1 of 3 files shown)


maquis196:
Signed-off-by: Christopher May-Townsend <chris@maytownsend.co.uk>

maquis196:
as per https://github.com/matrix-org/synapse/pull/8147 comments, changed the timings to help those on traefik without waiting a minute for nothing

anoadragon453:
The problem discussed in [the linked issue](https://github.com/matrix-org/synapse/pull/8147#issuecomment-830012399) is that the healthcheck will initially report `unhealthy` as it fires immediately after the container starts - while Synapse is still starting up.

Containers are [initially](https://docs.docker.com/engine/reference/builder/#healthcheck) in a `starting` state, until a healthcheck succeeds or fails. So just setting a delay on the first healthcheck (and setting a low retry period just in case we take longer than 5s to start) sounds like it would mostly address the problem to me (as this PR proposes).

That said, I do wonder if it's possible to delay the healthcheck (and consider the container actually started) until Synapse has itself started - however this is likely a much bigger task than what needs doing to address the original concern.

As for whether hitting `/health` 4x more is a concern, I don't really buy it as it's an [extremely lightweight](https://github.com/matrix-org/synapse/blob/7620912d84f6a8b24143f1340dd653f44b13bf30/synapse/rest/health.py#L19-L31) endpoint.

docker/README.md
```diff
@@ -191,6 +191,16 @@ whilst running the above `docker run` commands.
 ```
    --no-healthcheck
 ```
+
+## Disabling the healthcheck in docker-compose file
+
+If you wish to disable the healthcheck via docker-compose append your stanza with the following.
```
babolivier, line 197: Pardon my ignorance, but what's a stanza?

babolivier (changes requested):
Thanks for your contribution! This looks good to me, I just want to make sure the documentation is understandable enough before merging.

docker/README.md
maquis196: From wiki; "In computing, a stanza consists of a related group of lines in a script or configuration file.[1] Formats depend on context"

So with docker-compose the entire code block for a container would be the stanza.

service:
  nginx:
    image: "nginx:latest"
    restart: unless-stopped
    healthcheck:
      disabled: true

babolivier: Right, thanks for clarifying. I feel like that's a bit of an obscure term, and that it would be better to replace it to something that is easier to understand (e.g. "append the following to your service configuration" or something similar), since the goal of this documentation is to be easily to understand by as many people as possible.

maquis196: no problem, replaced.

babolivier: Brilliant thanks :)

babolivier (approved):

PR #9915, opened 2021-04-30
Support stable MSC1772 spaces identifiers.

clokep:
Blocked on FCP matrix-org/matrix-doc#1772.

Also see matrix-org/matrix-js-sdk#1679, matrix-org/complement#107.

### DIFF
--- a/synapse/handlers/space_summary.py
+++ b/synapse/handlers/space_summary.py
@@ -332,7 +332,9 @@ class SpaceSummaryHandler:
         )
 
         # TODO: update once MSC1772 lands
-        room_type = create_event.content.get(EventContentFields.MSC1772_ROOM_TYPE)
+        room_type = create_event.content.get(EventContentFields.ROOM_TYPE)
+        if not room_type:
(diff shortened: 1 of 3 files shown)


synapse/handlers/space_summary.py
```diff
@@ -360,8 +360,9 @@ async def _get_child_events(self, room_id: str) -> Iterable[EventBase]:
             [
                 event_id
                 for key, event_id in current_state_ids.items()
-                # TODO: update once MSC1772 lands
+                # TODO: update once MSC1772 has been FCP for a period of time.
```
clokep, line 363: It was suggested to do this when we stabilize the spaces summary MSC.

clokep:
Note that this is ready, but I've left it as draft until the MSC merges.

erikjohnston (approved):
SEEMS LEGIT

erikjohnston (approved):

clokep:
I pushed a follow-up to this in 70f0ffd2fcd815a065b4734ac606654a2e11dd28 which fixes the identifier for the room type, it's simply `type`, not `m.type`.

PR #9916, opened 2021-04-30
Increase perf of handling presence when joining large rooms.

erikjohnston:
We ended up doing a *lot* of duplicate work, and e.g. ended up doing n^2
worth of `is_mine_id(..)` checks across all joined users.

Based on #9910


synapse/handlers/presence.py
```diff
@@ -1231,72 +1234,55 @@ async def _handle_state_delta(self, deltas: List[JsonDict]) -> None:
...
+
+        # Construct sets for all the local users and remote hosts that were
+        # already in the room
+        prev_local_users = set()
+        prev_remote_hosts = set()
+        for user_id in prev_users:
+            if self.is_mine_id(user_id):
+                prev_local_users.add(user_id)
+            else:
+                prev_remote_hosts.add(get_domain_from_id(user_id))
```
richvdh, line 1265: I wonder if a general "partition a user list by homeserver" function would be generally useful. just a thought, no need to do anything about it.

synapse/handlers/presence.py
```diff
@@ -1231,72 +1234,55 @@ async def _handle_state_delta(self, deltas: List[JsonDict]) -> None:
...
+        # have them (because they already share rooms).
+
+        # Get all the users who were already in the room, by fetching the
+        # current users in the room and removing the newly joined users.
+        users = await self.store.get_users_in_room(room_id)
+        prev_users = set(users) - newly_joined_users
+
+        # Construct sets for all the local users and remote hosts that were
+        # already in the room
+        prev_local_users = set()
```
richvdh, line 1259: could make this one a list, since we don't rely on its settyness? Likewise newly_joined_local_users?

might be best to stay consistent though.

synapse/handlers/presence.py
```diff
@@ -1231,72 +1234,55 @@ async def _handle_state_delta(self, deltas: List[JsonDict]) -> None:
...
+        #      remote servers
+        #   2. presence states of newly joined users to all remote servers in
+        #      the room.
+        #
+        # TODO: Only send presence states to remote hosts that don't already
+        # have them (because they already share rooms).
+
+        # Get all the users who were already in the room, by fetching the
+        # current users in the room and removing the newly joined users.
+        users = await self.store.get_users_in_room(room_id)
```
richvdh, line 1254: can we rely on all deltas being for the same room id? and can we rely on there being at least one delta (so that `room_id` is set)?

richvdh (approved):
lgtm, modulo a couple of thoughts

synapse/handlers/presence.py
erikjohnston: Yeah, that would probably make a lot of sense

synapse/handlers/presence.py
erikjohnston: Oh, that is a good point. I don't think we can rely on there only being one room per call (though that is the common path).

richvdh (approved):

PR #9918, opened 2021-05-01
Support MSC2033: Device ID on whoami

turt2live:
MSC: matrix-org/matrix-doc#2033

The MSC has passed FCP, which means stable endpoints can be used.

This also includes a small change to the deactivation tests - see commits.


tests/rest/client/v2_alpha/test_account.py
```diff
@@ -458,6 +458,29 @@ def deactivate(self, user_id, tok):
...
+        register.register_servlets,
+    ]
+
+    def test_GET_whoami(self):
+        device_id = "wouldgohere"
+        user_id = self.register_user("kermit", "test")
+        tok = self.login("kermit", "test", device_id=device_id)
+
+        whoami = self.whoami(tok)
+        self.assertEqual(whoami, {"user_id": user_id, "device_id": device_id})
```
babolivier, line 476: I think this should be using `assertDictEqual` instead of `assertEqual`.

synapse/rest/client/v2_alpha/account.py
```diff
@@ -884,7 +884,10 @@ def __init__(self, hs):
     async def on_GET(self, request):
         requester = await self.auth.get_user_by_req(request)
 
-        return 200, {"user_id": requester.user.to_string()}
+        return 200, {
+            "user_id": requester.user.to_string(),
+            "device_id": requester.device_id,
```
babolivier, line 889: Have you tested what it does if no device ID is associated with the access token (e.g. if the user belongs to an appservice)? It looks to me like it would include the `device_id` property in the response but with a `null` value, is this compliant with MSC2033? It might also be worth adding a test for that to make sure it doesn't randomly explode in the future.

babolivier (changes requested):
A couple of concerns but lgtm otherwise.

tests/rest/client/v2_alpha/test_account.py
clokep: 'assertEqual' automatically uses that for dicts.

babolivier: Oh my bad, ignore me then, thanks for the tip!

synapse/rest/client/v2_alpha/account.py
turt2live: Do we have a test I can copy for being an appservice user?

turt2live: stole one from the appservice tests (which just makes sense, May 4th Travis)

turt2live:
it feels overly cruel to send this back to Brendan after a few months of radio silence, so sending back to the general queue.

tests/rest/client/v2_alpha/test_account.py
```diff
@@ -475,6 +476,23 @@ def test_GET_whoami(self):
...
 
+    def test_GET_whoami_appservices(self):
+        user_id = "@as_user_kermit:test"
+        as_token = "i_am_an_app_service"
+
+        appservice = ApplicationService(
+            as_token,
+            self.hs.config.server_name,
+            id="1234",
+            namespaces={"users": [{"regex": r"@as_user.*", "exclusive": True}]},
```
babolivier, line 487: Why do we need to treat `@as_user.*` as a raw string rather than a normal one?

babolivier (approved):

tests/rest/client/v2_alpha/test_account.py
turt2live: copy/paste? :P 

Can probably just use the same variable to do that though: will make the change and merge.

Issue #6579, opened 2019-12-20
Empty sync response could be more empty


bmarty:
Currently an empty sync response (i.e. without any change) is like that (with formatting added):

```
{
  "account_data": {
    "events": []
  },
  "to_device": {
    "events": []
  },
  "device_lists": {
    "changed": [],
    "left": []
  },
  "presence": {
    "events": []
  },
  "rooms": {
    "join": {},
    "invite": {},
    "leave": {}
  },
  "groups": {
    "join": {},
    "invite": {},
    "leave": {}
  },
  "device_one_time_keys_count": {
    "signed_curve25519": 50
  },
  "next_batch": "s1195808412_757269739_794431_REDACTED_1021253_37912961_19608850_66812"
}
```

According to the [spec](https://matrix.org/docs/spec/client_server/latest#get-matrix-client-r0-sync), only `next_batch` is required, so it could be acceptable to reduce it to:

```
{
  "next_batch": "s1195808412_757269739_794431_REDACTED_1021253_37912961_19608850_66812"
}
```

I do not know if it will solve the global climat changing problem, but if it's easy to fix, let's save some bytes!

bmarty:
I also observe empty elements in the `rooms.join` part, by instance:

```json
   "rooms": {
     "join": {
       "!gGkYcJLTzcNrfERksd:matrix.org": {
         "timeline": {
           "events": [], HERE
           "prev_batch": "s1222404111_757269739_4298397_414307685_257267759_1042954_39694860_20347120_70307",
           "limited": false
         },
         "state": { HERE
           "events": []
         },
         "account_data": { HERE
           "events": []
         },
         "ephemeral": {
           "events": [
             {
               "type": "m.receipt",
               "content": {
                 "$QIzSpVAHOcFGWWrVc6eXDc0wMBrAj87qtszSyO2TsEU": {
                   "m.read": {
                     "@kindly:naido.org": {
                       "ts": 1578909070063
                     }
                   }
                 }
               }
             }
           ]
         },
         "unread_notifications": {
           "notification_count": 27,
           "highlight_count": 0
         },
         "summary": {} HERE
       }
     }
   }
```

deepbluev7:
#9919 fixes the first part, but it doesn't go into rooms.join yet.

PR #9919, opened 2021-05-01
Leave out optional keys from /sync

deepbluev7:
This leaves out all optional keys from /sync. This should be fine for all clients tested against conduit already, but it may break some clients, as such we should check, that at least most of them don't break horribly and maybe back out some of the individual changes. (We can probably always leave out groups for example, while the others may cause more issues.)

Clients I have tested:

- [x] Nheko
- [x] Element (slightly outdated version)
- [x] FluffyChat (old version)

This also needs an updated sytest, see https://github.com/matrix-org/sytest/pull/1040

In my testing this decreases the sync json response to about half, which saves about 30% of data on every /sync. It also makes debugging a lot easier, since the response json is easier to read. :3

Tell me if the news entry is descriptive enough, since some people may want to be aware of this change, if they are client developers or so. Also, my python skills suck, I'm sure there is a tidier way to do this. Also, no idea how to run against the newer sytest version? Is that automatic, after it is merged?

### Pull Request Checklist

### DIFF
--- a/synapse/rest/client/v2_alpha/sync.py
+++ b/synapse/rest/client/v2_alpha/sync.py
@@ -14,6 +14,7 @@
 
 import itertools
 import logging
+from collections import defaultdict
 from typing import TYPE_CHECKING, Tuple
 
 from synapse.api.constants import PresenceState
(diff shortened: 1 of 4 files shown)


ShadowJonathan:
Please note this addresses https://github.com/matrix-org/synapse/issues/6579, see if it also fixes it.

synapse/rest/client/v2_alpha/sync.py
```diff
@@ -229,25 +229,54 @@ async def encode_response(self, time_now, sync_result, access_token_id, filter):
...
+            ] = sync_result.device_one_time_keys_count
+        if sync_result.device_unused_fallback_key_types:
+            response[
+                "org.matrix.msc2732.device_unused_fallback_key_types"
+            ] = sync_result.device_unused_fallback_key_types
+
+        if joined:
+            response.update({"rooms": {"join": joined}})
+        if invited:
+            response.update({"rooms": {"invite": invited}})
```
auscompgeek, line 267: Wouldn't this overwrite the existing `rooms` if you both joined a room and were invited to a room since last sync?

deepbluev7: I think you are right about that, I'll switch it to a defaultdict.

deepbluev7: Should be better now.

clokep:
I'm going to set the review flag on this even though it has failing tests, since I believe they should be fixed by matrix-org/sytest#1040.

deepbluev7:
Seems like some integration tests are failing, that I didn't see, when I ran them locally (probably overlooked them). I'll try to fix that soonish.

deepbluev7:
Should be fine now, although the tests would now fail, if you don't exclude the keys from /sync.

tests/server_notices/test_resource_limits_server_notices.py
```diff
@@ -306,8 +306,9 @@ def test_no_invite_without_notice(self):
 
         channel = self.make_request("GET", "/sync?timeout=0", access_token=tok)
 
-        invites = channel.json_body["rooms"]["invite"]
-        self.assertEqual(len(invites), 0, invites)
+        self.assertFalse(
+            "rooms" in channel.json_body, "Got invites without server notice"
+        )
```
richvdh, line 311:
```suggestion
        self.assertNotIn(
            "rooms", channel.json_body, "Got invites without server notice"
        )
```

richvdh (commented):
looks good! a couple of suggestions here.

tests/rest/client/v2_alpha/test_sync.py
```diff
@@ -40,11 +40,6 @@ def test_sync_argless(self):
         self.assertTrue(
             {
                 "next_batch",
-                "rooms",
-                "presence",
-                "account_data",
-                "to_device",
-                "device_lists",
             }.issubset(set(channel.json_body.keys()))
```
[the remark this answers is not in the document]
deepbluev7: Yes, probably, I just did the simplest transformation of the existing tests. I'll change it!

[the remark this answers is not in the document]
deepbluev7: Addtional question, do we even need the presence disabled test in that case?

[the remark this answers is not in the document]
richvdh: yeah, seems like it isn't doing much any more.

richvdh (approved):
lgtm

richvdh:
thanks very much!

deepbluev7:
Thank you for the review! <3

richvdh:
heads-up that this is being reverted in #9940

bmarty:
Hi,
You said that you've tested Element

> Element (slightly outdated version)

It would be nice to be more precise. Element Web, Element Android and Element iOS are three different implementations, so may have different bugs :/

deepbluev7:
@bmarty My bad, I missed to put Web. I only tested Element Web, since I have no Android or iOS device. That wasn't intentional, I just missed it! Sorry!

PR #9922, opened 2021-05-03
Allow a user who could join a restricted room to see it in spaces summary.

clokep:
Allows discovery of restricted rooms via [MSC3083](https://github.com/matrix-org/matrix-doc/pull/3083).

This is based on ~~#9974~~ #10009.

See matrix-org/complement#109 for complement tests.

### DIFF
--- a/synapse/handlers/event_auth.py
+++ b/synapse/handlers/event_auth.py
@@ -11,7 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Collection, Tuple
 
 from synapse.api.constants import EventTypes, JoinRules
 from synapse.api.room_versions import RoomVersion
@@ -28,6 +28,56 @@ class EventAuthHandler:
 
     def __init__(self, hs: "HomeServer"):
         self._store = hs.get_datastore()
+
+    async def get_allowed_spaces(
+        self, state_ids: StateMap[str], room_version: RoomVersion
+    ) -> Tuple[bool, Collection[str]]:
+        """
+        Generate a list of spaces allow access to a room.
+
+        Args:
+            state_ids: The state of the room as it currently is.
+            room_version: The room version of the room being joined.
+
+        Returns:
+            A tuple:
+                True if spaces give access to the room.
+                A collection of spaces (if any) which provide membership to the room.
+        """
+        # This only applies to room versions which support the new join rule.
+        if not room_version.msc3083_join_rules:
+            return False, ()
+
+        # If there's no join rule, then it defaults to invite (so this doesn't apply).
+        join_rules_event_id = state_ids.get((EventTypes.JoinRules, ""), None)
+        if not join_rules_event_id:
+            return False, ()
+
+        # If the join rule is not restricted, this doesn't apply.
+        join_rules_event = await self._store.get_event(join_rules_event_id)
+        if join_rules_event.content.get("join_rule") != JoinRules.MSC3083_RESTRICTED:
+            return False, ()
+
+        # If allowed is of the wrong form, then only allow invited users.
+        allowed_spaces = join_rules_event.content.get("allow", [])
+        if not isinstance(allowed_spaces, list):
+            return True, ()
+
+        # Pull out the room IDs, invalid data gets filtered.
+        result = []
+        for space in allowed_spaces:
+            if not isinstance(space, dict):
+                continue
+
+            space_id = space.get("space")
+            if not isinstance(space_id, str):
+                continue
+
+            # The user was joined to one of the spaces specified, they can join
+            # this room!
+            result.append(space_id)
+
+        return True, result
 
     async def can_join_without_invite(
         self, state_ids: StateMap[str], room_version: RoomVersion, user_id: str
@@ -46,39 +96,20 @@ class EventAuthHandler:
         Returns:
             True if the user can join the room, false otherwise.
         """
-        # This only applies to room versions which support the new join rule.
-        if not room_version.msc3083_join_rules:
+        # Get whether spaces allow access to the room and the allowed spaces.
+        allow_via_spaces, allowed_spaces = await self.get_allowed_spaces(
+            state_ids, room_version
+        )
+        # Access via spaces doesn't apply to this room.
+        if not allow_via_spaces:
             return True
-
-        # If there's no join rule, then it defaults to invite (so this doesn't apply).
-        join_rules_event_id = state_ids.get((EventTypes.JoinRules, ""), None)
-        if not join_rules_event_id:
-            return True
-
-        # If the join rule is not restricted, this doesn't apply.
-        join_rules_event = await self._store.get_event(join_rules_event_id)
-        if join_rules_event.content.get("join_rule") != JoinRules.MSC3083_RESTRICTED:
-            return True
-
-        # If allowed is of the wrong form, then only allow invited users.
-        allowed_spaces = join_rules_event.content.get("allow", [])
-        if not isinstance(allowed_spaces, list):
-            return False
 
         # Get the list of joined rooms and see if there's an overlap.
         joined_rooms = await self._store.get_rooms_for_user(user_id)
 
-        # Pull out the other room IDs, invalid data gets filtered.
-        for space in allowed_spaces:
-            if not isinstance(space, dict):
-                continue
-
-            space_id = space.get("space")
-            if not isinstance(space_id, str):
-                continue
-
-            # The user was joined to one of the spaces specified, they can join
-            # this room!
+        # If the user was joined to one of the spaces specified, they can join
+        # this room!
+        for space_id in allowed_spaces:
             if space_id in joined_rooms:
                 return True
 
--- a/synapse/handlers/event_auth.py
+++ b/synapse/handlers/event_auth.py
@@ -62,6 +62,11 @@ class EventAuthHandler:
         # Get the spaces which allow access to this room, if applicable.
         allowed_spaces = await self.get_spaces_that_allow_join(state_ids, room_version)
         if allowed_spaces is None:
+            # This is not a room with a restricted join rule, so we don't need to do the
+            # restricted room specific checks.
+            #
+            # Note: We'll be applying the standard join rule checks later, which will
+            # catch the cases of e.g. trying to join private rooms without an invite.
             return
 
         # The user was not in any of the required spaces.
--- a/synapse/handlers/event_auth.py
+++ b/synapse/handlers/event_auth.py
@@ -59,52 +59,68 @@ class EventAuthHandler:
         ):
             return
 
-        # Get the spaces which allow access to this room, if applicable.
-        allowed_spaces = await self.get_spaces_that_allow_join(state_ids, room_version)
-        if allowed_spaces is None:
-            # This is not a room with a restricted join rule, so we don't need to do the
-            # restricted room specific checks.
-            #
-            # Note: We'll be applying the standard join rule checks later, which will
-            # catch the cases of e.g. trying to join private rooms without an invite.
+        # This is not a room with a restricted join rule, so we don't need to do the
+        # restricted room specific checks.
+        #
+        # Note: We'll be applying the standard join rule checks later, which will
+        # catch the cases of e.g. trying to join private rooms without an invite.
+        if not await self.has_restricted_join_rules(state_ids, room_version):
             return
 
-        # The user was not in any of the required spaces.
+        # Get the spaces which allow access to this room and check if the user is
+        # in any of them.
+        allowed_spaces = await self.get_spaces_that_allow_join(state_ids)
         if not await self.is_user_in_rooms(allowed_spaces, user_id):
             raise AuthError(
                 403,
                 "You do not belong to any of the required spaces to join this room.",
             )
 
-    async def get_spaces_that_allow_join(
+    async def has_restricted_join_rules(
         self, state_ids: StateMap[str], room_version: RoomVersion
-    ) -> Optional[Collection[str]]:
+    ) -> bool:
         """
-        Generate a list of spaces which allow access to a room.
+        Return if the room has the proper join rules set for access via spaces.
 
         Args:
             state_ids: The state of the room as it currently is.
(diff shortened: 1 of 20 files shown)


synapse/handlers/space_summary.py
```diff
@@ -131,31 +140,30 @@ async def get_space_summary(
                 # as the requesting server, are allowed to see, but the requesting
                 # user is not permitted see. Handle that filtering.
                 room_ids = set()
-                rooms = []
+                rooms = []  # type: List[JsonDict]
```
erikjohnston, line 146: We don't seem to be using this anymore?

erikjohnston (commented):
I think this mostly looks sane?

synapse/handlers/space_summary.py
clokep: I think you're right, yes. 👍

synapse/handlers/event_auth.py
```diff
@@ -29,46 +29,42 @@ class EventAuthHandler:
...
-    ) -> bool:
+    async def get_allowed_spaces(
+        self, state_ids: StateMap[str], room_version: RoomVersion
+    ) -> Tuple[bool, Collection[str]]:
         """
-        Check whether a user can join a room without an invite.
-
-        When joining a room with restricted joined rules (as defined in MSC3083),
-        the membership of spaces must be checked during join.
+        Generate a list of spaces allow access to a room.
```
sumnerevans, line 36:
```suggestion
        Generate a list of spaces which allow access to a room.
```

sumnerevans (changes requested):

clokep:
This pretty much implements the logic in https://github.com/matrix-org/matrix-doc/blob/clokep/restricted-rooms/proposals/3083-restricted-rooms.md#discovery-of-restricted-rooms

The goal is that "restricted rooms" are able to show up in the space summary, but only for those who would be allowed to join them. This gets a bit hairy over federation.

erikjohnston (dismissed):

synapse/handlers/event_auth.py
```diff
@@ -29,46 +29,42 @@ class EventAuthHandler:
     def __init__(self, hs: "HomeServer"):
         self._store = hs.get_datastore()
 
-    async def can_join_without_invite(
-        self, state_ids: StateMap[str], room_version: RoomVersion, user_id: str
-    ) -> bool:
+    async def get_allowed_spaces(
```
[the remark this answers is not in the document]
clokep: The difference between `(True, [])` and `(False, [])` is that the former means "no one can access this room", while the latter means "spaces do not control access to this room"

Renaming this sounds good though.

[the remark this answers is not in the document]
erikjohnston: I just don't see anywhere where the callers care about the difference? They all seem to just iterate over the spaces returned, which is a no-op if an empty list is return?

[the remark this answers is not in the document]
erikjohnston: The reason for raising this is that I was thinking of this function as "list of spaces which allow access to a room", and couldn't figure out what we did differently between `(True, [])` and `(False, [])`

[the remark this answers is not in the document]
clokep: So this is differentiating between "these join rules don't apply to this room" and "here's the list of spaces that you must be in to join this room" -- I could split this into two separate checks if that's clearer? (Or raise an exception or something in one-case to change the control flow?)

> The reason for raising this is that I was thinking of this function as "list of spaces which allow access to a room"

The function is meant to be "check if spaces give access to this room, if so, get the list of spaces which allow access to this room"

synapse/handlers/event_auth.py
```diff
@@ -59,32 +59,55 @@ async def check_restricted_join_rules(
...
         join_rules_event_id = state_ids.get((EventTypes.JoinRules, ""), None)
         if not join_rules_event_id:
-            return
+            return None
 
         # If the join rule is not restricted, this doesn't apply.
         join_rules_event = await self._store.get_event(join_rules_event_id)
         if join_rules_event.content.get("join_rule") != JoinRules.MSC3083_RESTRICTED:
-            return
+            return None
```
erikjohnston, line 102: I'm wondering if splitting this out into `has_restricted_join_rules` would make sense? That way `get_spaces_that_allow_join` can always return a list and callers that care can do the explicit check? I think the only place the difference matters is `check_restricted_join_rules`, where being explicit probably helps? IDK.

synapse/handlers/space_summary.py
```diff
@@ -408,6 +521,7 @@ async def _build_room_entry(self, room_id: str) -> JsonDict:
             "guest_can_join": stats["guest_access"] == "can_join",
             "creation_ts": create_event.origin_server_ts,
             "room_type": room_type,
+            "allowed_spaces": allowed_spaces if allowed_spaces else None,
```
erikjohnston, line 524: I wonder if we should only be returning the list of spaces that the origin is in? I don't know to what extent we care about leaking the existence of spaces?

synapse/handlers/space_summary.py
```diff
@@ -111,28 +116,75 @@ async def get_space_summary(
...
+                # A room may be visible if it is world readable, or if it is accessible
+                # to the specific user.
+                room_ids = set()
+                events = []
+                for room in fed_rooms:
+                    fed_room_id = room.get("room_id")
+                    if not fed_room_id:
+                        continue
+
+                    # Check whether the room is world readable from the response.
```
erikjohnston, line 152:
```suggestion
                    # A user is only allowed to get a summary for the room if:
                    #     a. the user is in the room;
                    #     b. the room is world readable; or
                    #     c. the user is in a space that has been granted access to 
                    #        the room.
                    #
                    # The user isn't in the room (since the server isn't in the room), so
                    # we just need to check that one of the last two conditions hold. 
```

synapse/handlers/space_summary.py
```diff
@@ -111,28 +116,75 @@ async def get_space_summary(
...
+                        include_room = await self._event_auth_handler.is_user_in_rooms(
+                            allowed_spaces, requester
+                        )
+
+                    # Finally, check ourselves if we can access the room (this will
+                    # fail if we don't have any state for this room).
+                    if not include_room:
+                        include_room = await self._is_room_accessible(
+                            fed_room_id, requester, None
+                        )
```
erikjohnston, line 168: If we're not in the room this should always fail, no?

synapse/handlers/space_summary.py
```diff
@@ -111,28 +116,75 @@ async def get_space_summary(
...
-                    requester, room_id, suggested_only, max_children
+                room, events = await self._summarize_local_room(
+                    requester, None, room_id, suggested_only, max_children
                 )
+
+                logger.debug(
+                    "Query of local room %s returned events %s",
+                    room_id,
+                    ["%s->%s" % (ev["room_id"], ev["state_key"]) for ev in events],
+                )
```
erikjohnston, line 127: Do we still want this debug logging? (I just note we're creating a list here even when debugging is off)

synapse/handlers/space_summary.py
```diff
@@ -332,45 +397,88 @@ async def _summarize_remote_room(
...
+
+            # Otherwise, if they should be allowed access via membership in a space.
+            # Get the spaces which allow access to this room, if applicable.
+            allowed_spaces = await self._event_auth_handler.get_spaces_that_allow_join(
+                state_ids, room_version
+            )
+            if allowed_spaces and await self._event_auth_handler.is_user_in_rooms(
+                allowed_spaces, requester
+            ):
+                return True
```
erikjohnston, line 453: This is pretty much the same as `check_restricted_join_rules`? I wonder if should use that or add a helper function for it?

synapse/handlers/space_summary.py
erikjohnston: Is this trying to check for invites? I don't think that will work.

synapse/handlers/space_summary.py
```diff
@@ -111,28 +116,75 @@ async def get_space_summary(
...
+                            fed_room_id, requester, None
+                        )
+
+                    if include_room:
+                        rooms_result.append(room)
+                        room_ids.add(fed_room_id)
+
+                    # any rooms returned don't need visiting again (even if the user
+                    # didn't have access to them).
+                    processed_rooms.add(cast(str, room_id))
```
erikjohnston, line 176: Is this meant to be `fed_room_id`? If so, then we should probably explicitly check its a string rather than casting (and if not why do we need a cast as `room_id` is declared a `str`?).

erikjohnston (commented):
I think this is broadly looking good. I've made a couple of suggestions for things that might make this clearer, but feel free to push back if it just feels like rearranging deck chairs.

synapse/handlers/space_summary.py
clokep: Oh, interesting. The issue with that is we don't necessarily know if the origin is in them (if we're not in them ourselves), so I think that would make the edge-case at the bottom of https://github.com/matrix-org/matrix-doc/blob/clokep/restricted-rooms/proposals/3083-restricted-rooms.md#discovery-of-restricted-rooms worse:

> Note that there are still potential situations where each server individually doesn't have enough information to properly return the full summary, but these do not seem reasonable in what is considered a normal structure of spaces. (E.g. in the above example, if the remote server is not in the space and does not know whether the server is in the space or not it cannot return the room.)

So if we further filter this list down we might be making *more* situations where the servers cannot agree who is in the room. I'll need to think about it though...

synapse/handlers/space_summary.py
clokep: We call the federation API on a root room and it returns the root room + children rooms, so we *might* be in the room. (We only know we weren't in the "root" room, not the children room.) I think the comment suggested above is also incorrect for this reason, but I'm going to add one which explains this.

synapse/handlers/space_summary.py
clokep: It is, yes! Good catch. 👍 

I agree, I'll add it to the check where we ensure it exists.

synapse/handlers/space_summary.py
clokep: I'm not sure, this was originally added by @richvdh but does seem generally helpful. This could be a rather expensive call though I suppose. My inclination would be to leave it for now until we're sure this all works? (Or maybe guard it in if debug is enabled if you'd prefer?)

synapse/handlers/space_summary.py
clokep: I updated the comment above this to hopefully clarify what's going on.

synapse/handlers/event_auth.py
clokep: All the callers care if the join rules are active or not? But some need to differentiate between whether they are activate vs an empty list? Is that what you're getting at? I'll try it and see how it looks.

clokep: I'm not sure if this helps at all, the 4 callers end up just calling another method before it. It might be a bit clearer though about what's happening.

I pushed d7206152b6be446981beae4e83f863d9973d1294 with the change though.

synapse/handlers/space_summary.py
clokep: We could add a helper (`can_join_via_restricted_join_rules`) which returns a boolean which is called here and `check_restricted_join_rules`?

synapse/handlers/event_auth.py
```diff
@@ -59,32 +59,76 @@ async def check_restricted_join_rules(
...
+        # Get the spaces which allow access to this room and check if the user is
+        # in any of them.
+        allowed_spaces = await self.get_spaces_that_allow_join(state_ids)
+        if not await self.is_user_in_rooms(allowed_spaces, user_id):
+            raise AuthError(
+                403,
+                "You do not belong to any of the required spaces to join this room.",
+            )
+
+    async def has_restricted_join_rules(
```
erikjohnston, line 79: This doesn't need to be async?

synapse/handlers/event_auth.py
```diff
@@ -59,32 +59,76 @@ async def check_restricted_join_rules(
...
+        Args:
+            state_ids: The state of the room as it currently is.
+
+        Returns:
+            A collection of spaces which provide membership to the room.
+        """
+        # If there's no join rule, then it defaults to invite (so this doesn't apply).
+        join_rules_event_id = state_ids.get((EventTypes.JoinRules, ""), None)
+        if not join_rules_event_id:
+            return ()
```
erikjohnston, line 120: I'd be tempted to also do a check of `has_restricted_join_rules` here as well

synapse/handlers/space_summary.py
erikjohnston: Eh, its probably fine for now

synapse/handlers/space_summary.py
```diff
@@ -111,28 +116,88 @@ async def get_space_summary(
...
-
-            rooms_result.extend(rooms)
-
-            # any rooms returned don't need visiting again
-            processed_rooms.update(cast(str, room.get("room_id")) for room in rooms)
+                # The results over federation might include rooms that the we,
+                # as the requesting server, are allowed to see, but the requesting
+                # user is not permitted see.
+                #
+                # Filter the returned results to only what is accessible o the user.
```
erikjohnston, line 143:
```suggestion
                # Filter the returned results to only what is accessible to the user.
```

synapse/handlers/event_auth.py
clokep: Then we'll do the check twice for every caller? 😕

erikjohnston: Ugh, true. IDK.

synapse/handlers/event_auth.py
clokep: It has an await in it to get the event.

synapse/handlers/space_summary.py
clokep: I attempted this but it seemed more confusing in a way. I ended up doing a call `check_restricted_join_rules` and catching the `AuthError`.

synapse/handlers/event_auth.py
clokep: As of now we end up re-fetching the join rules event twice too which is unfortunate.

clokep:
@erikjohnston I think this is back to you. There's a couple of open threads above still, but I've taken into account most of your feedback (or replied to it).

erikjohnston (approved):
I think this now makes sense!

PR #9928, opened 2021-05-05
Include the origin_server_ts of the create event in Spaces Summary.

clokep:
Per updates the [MSC2946](https://github.com/matrix-org/matrix-doc/pull/2946), this is necessary for clients to sort rooms (as specified in MSC1772).

### DIFF
--- a/synapse/handlers/space_summary.py
+++ b/synapse/handlers/space_summary.py
@@ -344,7 +344,7 @@ class SpaceSummaryHandler:
                 stats["history_visibility"] == HistoryVisibility.WORLD_READABLE
             ),
             "guest_can_join": stats["guest_access"] == "can_join",
-            "origin_server_ts": create_event.origin_server_ts,
+            "creation_ts": create_event.origin_server_ts,
             "room_type": room_type,
         }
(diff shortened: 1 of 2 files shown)


synapse/handlers/space_summary.py
```diff
@@ -344,6 +344,7 @@ async def _build_room_entry(self, room_id: str) -> JsonDict:
                 stats["history_visibility"] == HistoryVisibility.WORLD_READABLE
             ),
             "guest_can_join": stats["guest_access"] == "can_join",
+            "origin_server_ts": create_event.origin_server_ts,
```
richvdh, line 347: I'd think `creation_ts` would be a better name here?

richvdh (approved):
lgtm modulo below

synapse/handlers/space_summary.py
clokep: Yeah, that's possible better although that sounds like a comment for the MSC. 😄

clokep: I've made this change on the MSC and here!

PR #9933, opened 2021-05-05
Support for database schema version ranges

richvdh:
This is essentially an implementation of the proposal made at https://hackmd.io/@richvdh/BJYXQMQHO, though the details have ended up looking slightly different.

I suggest reviewing commit-by-commit. In the second commit, I suggest starting with the README.md, which I hope will give a clear picture of how this is meant to work.

~~Based on #9932.~~


synapse/storage/schema/common/delta/59/13schema_compat_version.sql
```diff
@@ -0,0 +1,17 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+-- populate schema_compat_version for the first time.
+INSERT INTO schema_compat_version(compat_version) VALUES (59);
```
erikjohnston, line 17: Would it make sense to have this be defined in code next to `SCHEMA_VERSION` and updated when we check schema versions (ensuring it only increases)? It just feels a bit odd to me that we're defining these in two separate places, is all.

erikjohnston (commented):
I think this LGTM modulo maybe moving `COMPAT_VERSION` into code.

Non blocking thought: I wonder whether we can have any CI tests that check that we can safely rollback to a given version? I don't think using SyTest would work as it relies on having a clean DB (and generates new users anyway)?

synapse/storage/schema/common/delta/59/13schema_compat_version.sql
richvdh: ok, this is now done.

richvdh:
> Non blocking thought: I wonder whether we can have any CI tests that check that we can safely rollback to a given version? I don't think using SyTest would work as it relies on having a clean DB (and generates new users anyway)?

Yeah, I think that would be a nice thing to do, ultimately. As a starting point, we could maybe use the same database that is used for the port_db test - ie, take the test database, update it to the latest synapse, and then check that an older synapse will at least start against the updated database.

I think the hardest thing there would be finding a suitable version of synapse to test with.

docs/development/database_schema.md
```diff
@@ -0,0 +1,95 @@
+# Synapse database schema files
+
+Synapse's database schema is stored in the `synapse.storage.schema` module.
+
+## Logical databases
+
+Synapse supports splitting its datastore across multiple physical databases (which can
+be useful for large installations), and the schema files are therefore split according
+to the logical database they are apply to.
```
anoadragon453, line 9:
```suggestion
to the logical database they apply to.
```

Or alternatively "they are applied to".

docs/development/database_schema.md
```diff
@@ -0,0 +1,95 @@
...
+
+At the time of writing, the following "logical" databases are supported:
+
+* `state` - used to store Matrix room state (more specifically, `state_groups`,
+  their relationships and contents.)
```
anoadragon453, line 14:
```suggestion
  their relationships and contents).
```

docs/development/database_schema.md
```diff
@@ -0,0 +1,95 @@
...
+* `main` - stores everything else.
+
+Addionally, the `common` directory contains schema files for tables which must be
```
anoadragon453, line 17:
```suggestion
Additionally, the `common` directory contains schema files for tables which must be
```

anoadragon453 (commented):
I just came in to have a look at the docs, and found them very digestible!

synapse/storage/prepare_database.py
```diff
@@ -59,6 +58,28 @@ class UpgradeDatabaseException(PrepareDatabaseException):
...
+    """The SCHEMA_VERSION of the oldest version of Synapse for this database
+
+    If this is None, we have an old version of the database without the necessary
+    table.
+    """
+
+    applied_deltas: Collection[str] = attr.ib(factory=tuple)
+    """Any delta files for `current_version` which have already been applied"""
+
+    upgraded: bool = attr.ib(default=False)
```
erikjohnston, line 76: (Fwiw you could remove all the `attr.ib()` here and use `auto_attribs=True`)

erikjohnston (approved):
:tada:

synapse/storage/prepare_database.py
richvdh: yeah, I think I didn't know about `auto_attribs` when I first wrote this code...

PR #9935, opened 2021-05-06
Improve performance of backfilling in large rooms.

erikjohnston:
We were pulling the full auth chain for the room out of the DB each time
we backfilled, which can be *huge* for large rooms and is totally
unnecessary.

### DIFF
--- a/synapse/handlers/federation.py
+++ b/synapse/handlers/federation.py
@@ -552,8 +552,12 @@ class FederationHandler(BaseHandler):
         destination: str,
         room_id: str,
         event_id: str,
-    ) -> Tuple[List[EventBase], List[EventBase]]:
-        """Requests all of the room state at a given event from a remote homeserver.
+    ) -> List[EventBase]:
+        """Requests all of the room state at a given event from a remote
+        homeserver.
+
+        Will also fetch any missing events reported in the `auth_chain_ids`
+        section of `/state_ids`.
 
         Args:
             destination: The remote homeserver to query for the state.
@@ -561,8 +565,7 @@ class FederationHandler(BaseHandler):
             event_id: The id of the event we want the state at.
 
         Returns:
-            A list of events in the state, not including the event itself, and
-            a list of events in the auth chain for the given event.
+            A list of events in the state, not including the event itself.
         """
         (
             state_event_ids,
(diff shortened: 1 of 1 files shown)


synapse/handlers/federation.py
```diff
@@ -571,77 +574,64 @@ async def _get_state_for_room(
...
-            logger.warning(
-                "Failed to fetch missing state/auth events for %s %s",
-                event_id,
-                failed_to_fetch,
+        # Check for missing events. We handle state and auth event seperately,
+        # as we want to pull the state from the DB, but we don't for the auth
+        # events. (Note: we likely won't use the majority of the auth chain, and
+        # it can be *huge* for large rooms, so it's worth ensuring that we don't
+        # unnecessarily pull it from the DB).
+        missing_state_events = set(event_map) - set(state_event_ids)
```
clokep, line 586: Shouldn't this be `set(state_event_ids) - set(event_map)`? I think right now `missing_state_events` will always be empty (since `event_map` by definition could only possibly have things that were in `state_event_ids`).

synapse/handlers/federation.py
```diff
@@ -571,77 +574,64 @@ async def _get_state_for_room(
...
-
-        Persists any events we don't already have as outliers.
-
-        If we fail to fetch any of the events, a warning will be logged, and the event
-        will be omitted from the result. Likewise, any events which turn out not to
-        be in the given room.
+            if missing_state_events:
+                new_events = await self.store.get_events(
+                    state_event_ids, allow_rejected=True
+                )
```
clokep, line 598: Should this only fetch the new events instead of all events again?

synapse/handlers/federation.py
```diff
@@ -571,77 +574,64 @@ async def _get_state_for_room(
...
+                    )
 
-            # we need to make sure we re-load from the database to get the rejected
-            # state correct.
-            fetched_events.update(
-                (await self.store.get_events(missing_events, allow_rejected=True))
-            )
+        remote_state = [
+            event_map[e_id] for e_id in state_event_ids if e_id in event_map
+        ]
```
clokep, line 625: Isn't this pretty much `event_map.values()`?

synapse/handlers/federation.py
erikjohnston: Err, yes :facepalm:

clokep (approved):
I think this looks good.

PR #9939, opened 2021-05-06
Ensure we only have one copy of an event in memory at a time

erikjohnston:
This ensures that if the get event cache overflows we don't end up with
multiple copies of the event in RAM at the same time (which could lead
to memory bloat).

My hunch is that this would help with the federation readers on matrix.org using tonnes of memory due to them having undersized caches. But that is just a hunch with a tiny bit of circumstantial evidence to back it up.

### DIFF
--- a/synapse/storage/databases/main/events_worker.py
+++ b/synapse/storage/databases/main/events_worker.py
@@ -742,6 +742,8 @@ class EventsWorkerStore(SQLBaseStore):
             # that would require a larger refactor).
             cached_entry = self._in_memory_events.get(event_id)
             if cached_entry is not None:
+                # We need to add to the event_map as we read from it to fetch redactions.
+                event_map[event_id] = cached_entry.event
                 result_map[event_id] = cached_entry
                 self._get_event_cache.set((event_id,), cached_entry)
                 continue
--- a/synapse/storage/databases/main/events_worker.py
+++ b/synapse/storage/databases/main/events_worker.py
@@ -721,6 +721,9 @@ class EventsWorkerStore(SQLBaseStore):
             if events_to_fetch:
                 logger.debug("Also fetching redaction events %s", events_to_fetch)
 
+        # The events to return
+        result_map = {}  # type: Dict[str, _EventCacheEntry]
+
         # build a map from event_id to EventBase
         event_map = {}
         for event_id, row in fetched_events.items():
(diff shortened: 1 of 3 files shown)


synapse/storage/databases/main/events_worker.py
```diff
@@ -1056,7 +1070,11 @@ async def have_seen_events(self, event_ids):
             set[str]: The events we have already seen.
         """
         # if the event cache contains the event, obviously we've seen it.
-        results = {x for x in event_ids if self._get_event_cache.contains(x)}
+        results = {
+            x
+            for x in event_ids
+            if self._get_event_cache.contains((x,)) or x in self._in_memory_events
```
erikjohnston, line 1076: The change from `_get_event_cache.contains(x)` to `_get_event_cache.contains((x,))` is a bug fix I think, since the get event cache uses `(event_id,)` as its keys.

synapse/storage/databases/main/events_worker.py
```diff
@@ -73,7 +74,10 @@
 EVENT_QUEUE_TIMEOUT_S = 0.1  # Timeout when waiting for requests for events
 
 
-_EventCacheEntry = namedtuple("_EventCacheEntry", ("event", "redacted_event"))
+@attr.s(slots=True, frozen=True, auto_attribs=True)
+class _EventCacheEntry:
+    event: EventBase
+    redacted_event: Optional[EventBase]
```
erikjohnston, line 80: I needed to convert this as `namedtuple` cannot be a `weakref`.

richvdh (commented):
there's nothing that guarantees that we don't make multiple concurrent calls to `_get_events_from_cache_or_db` for the same event at once, so this is racy.

That's not the end of the world, but does mean that we can still end up with multiple copies of the same event in memory at once.

Maybe you also need to keep track of the events that we are in the process of (trying to) fetch?

erikjohnston:
> there's nothing that guarantees that we don't make multiple concurrent calls to `_get_events_from_cache_or_db` for the same event at once, so this is racy.

Argh, I thought we deduplicated requests for the same event (so there'd still be a race, just a small one), but apparently not. Let me have a look to see how tractable fixing that is.

erikjohnston:
> > there's nothing that guarantees that we don't make multiple concurrent calls to `_get_events_from_cache_or_db` for the same event at once, so this is racy.
> 
> Argh, I thought we deduplicated requests for the same event (so there'd still be a race, just a small one), but apparently not. Let me have a look to see how tractable fixing that is.

Ugh, I think fixing this properly requires more work than I'd otherwise like to start right now. We do deduplicate when pulling the events from the DB, but then we respond to each request with the event json, and then each request makes its own `Event` object. Unpicking that (and ensuring that we didn't have performance regressions) looks to be a bit of a PITA.

Instead I've pushed a change to check if we already have that event in memory before building another event object. That very much is plastering over the problem, but I *think* should work. I'm not entirely happy about it, but it's probably better than leaving the situation as is? Feel free to disagree though.

erikjohnston:
> > Ugh, I think fixing this properly requires more work than I'd otherwise like to start right now
> 
> silly question maybe, but is it not as simple as maintaining a map from `event_id` to `Deferred` in `_get_events_from_cache_or_db` ?

Oh hmm, for some reason I thought that was happening at a slightly different level. To make that work I think we'd need to hoist up the `allow_rejected` checks, but that is doable.

erikjohnston:
In other news: I don't think this works as intended? We're storing a weakref to the `_EventCacheEntry`, which gets dropped immediately upon leaving the cache? We *actually* want to store weakrefs to the events themselves?

erikjohnston:
This is going to take some more thinking about, and some rejigging, and ugh. So much for me thinking this would be a quick win :slightly_smiling_face:

PR #9940, opened 2021-05-06
Revert "Leave out optional keys from /sync (#9919)"

erikjohnston:
This reverts commit e9eb3549d32a6f93d07de8dbd5e1ebe54c8d8278 / #9919

The PR breaks Element iOS push, as its handling expects some of the keys to be there. We revert while they fix it and put out a new release.

### DIFF
--- a/changelog.d/9919.feature
+++ b/changelog.d/9919.feature
@@ -1,1 +1,0 @@
-Omit empty fields from the `/sync` response.  Contributed by @deepbluev7.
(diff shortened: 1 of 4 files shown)


manuroe:
iOS issue: https://github.com/vector-im/element-ios/issues/4309

richvdh (approved):

deepbluev7:
Do I need to do anything to get that back in in a few weeks then or will that be done automatically? :3

callahad:
> Do I need to do anything to get that back in in a few weeks then or will that be done automatically? :3

I've filed #9941 to track re-introducing that change. The `Z-Future-Maintenance` label ensures we'll look at it every two weeks during release planning, and, when things are ready, we'll pull it back in :)

deepbluev7:
Thank you!

Issue #9936, opened 2021-05-06
Failing to sync: attr._funcs.evolve() argument after ** must be a mapping, not RoomStreamToken


Half-Shot:
Seeing these errors on latest develop (merged with #9379 -- so far that branch has worked consistently). 

```
2021-05-06 10:37:24,144 - synapse.http.server - 93 - ERROR - GET-282- Failed handle request via 'SyncRestServlet': <XForwardedForRequest at 0x7fa7e5691910 method='GET' uri='/_matrix/client/r0/sync?filter=183&timeout=0&since=s45567119_293425298_1_29910218_595864_1764_1018470_21679023_52' clientproto='HTTP/1.0' site=8018>
Traceback (most recent call last):
  File "/opt/synapse/lib/python3.9/site-packages/twisted/internet/defer.py", line 1445, in _inlineCallbacks
    result = current_context.run(g.send, result)
StopIteration: ([<FrozenEventV3 event_id='$b1vB74hFd4li3crjazdhqyhysvEk6sr_dB85hCuWzK4', type='m.room.member', state_key='@freenode_Kryptron_:matrix.org'>, <FrozenEventV3 event_id='$4fSNY9DWOo8XOOR1TrLbev_S4OWjXFGGtkYlbFs9GWs', type='m.room.member', state_key='@greatudini:matrix.org'>, <FrozenEventV3 event_id='$XXjQ2trh4wxhW7BQ15UsTtgNRerkrSw5A7OTWM4OVTY', type='m.room.message', state_key=None>, <FrozenEventV3 event_id='$1ci12KD7g0aQowO7a7NbUM8ixYNt_NunMELTBUvau_k', type='m.room.message', state_key=None>, <FrozenEventV3 event_id='$Ux9DERofhO04Smj3PcVDAPw6w7s5ihNS0xrm1qpVY3w', type='m.room.message', state_key=None>, <FrozenEventV3 event_id='$bpVth3vcA6IMHZS2GnoVxhCMywgld9I-G0sIOsHIJdc', type='m.room.message', state_key=None>], RoomStreamToken(topological=None, stream=45567154, instance_map={}))

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/opt/synapse/lib/python3.9/site-packages/twisted/internet/defer.py", line 1445, in _inlineCallbacks
    result = current_context.run(g.send, result)
[...]

PR #9946, opened 2021-05-07
Unpin attrs dep after new version has been released

erikjohnston:
c.f. #9936

### DIFF
--- a/synapse/python_dependencies.py
+++ b/synapse/python_dependencies.py
@@ -79,7 +79,7 @@ REQUIREMENTS = [
     # Fedora 31 only has 19.1, so if we want to upgrade we should wait until 33
     # is out in November.)
     # Note: 21.1.0 broke `/sync`, see #9936
-    "attrs>=19.1.0,<21.1.0",
+    "attrs>=19.1.0,!=21.1.0",
     "netaddr>=0.7.18",
     "Jinja2>=2.9",
     "bleach>=1.4.3",


babolivier (approved):
Will we want to cherry-pick this into 1.33.2?

erikjohnston:
> Will we want to cherry-pick this into 1.33.2?

Yeah probably, I'll make a release-v1.33.2 branch and we can merge it in there?

babolivier:
> > Will we want to cherry-pick this into 1.33.2?
> 
> Yeah probably, I'll make a release-v1.33.2 branch and we can merge it in there?

sgtm

synapse/python_dependencies.py
```diff
@@ -79,7 +79,7 @@
     # Fedora 31 only has 19.1, so if we want to upgrade we should wait until 33
     # is out in November.)
     # Note: 21.1.0 broke `/sync`, see #9936
-    "attrs>=19.1.0,<21.1.0",
+    "attrs>=19.1.0,!=21.1.0",
```
callahad, line 82: Probably unnecessary; 21.1.0 was yanked from PyPI, so it effectively no longer exists. We can go back to just `attrs>=19.1.0`

erikjohnston: Mmm, true. Though I think since we know its bad we may as well leave it in there. I'm not sure if e.g. pypi mirrors and the like handle yanking etc.

PR #9947, opened 2021-05-07
Support fetching the spaces summary via GET over federation.

clokep:
Per updates to matrix-org/matrix-doc#2946.

I've left the POST endpoints (both C-S and S-S) with the intention that they would be removed when switching to the stable endpoint. For now I've left the federation queries going to the POST endpoint since that's more likely to exist (i.e. we don't need to have a chicken-egg problem with deploying this).


synapse/federation/transport/client.py
```diff
@@ -995,19 +995,20 @@ async def get_space_summary(
                returned per space
             exclude_rooms: a list of any rooms we can skip
         """
+        # TODO When switching to the stable endpoint, use GET instead of POST.
```
richvdh, line 998: we're already using GET? what's left to do?

richvdh (approved):
lgtm otherwise

synapse/federation/transport/client.py
clokep: Oh, hmm...I think I should not have included the changes to this file actually, as this will break federation of this feature until other users upgrade.

Do you think it is best to:

* Leave this as is and potentially break federation,
* Leave this as a `POST` for now and update it in the future

richvdh: yeah, I think it does, annoyingly.

clokep: I backed out the changes from here, minus the comment which now makes sense. 😄

PR #9949, opened 2021-05-07
Add room members admin API with power levels

awesome-manuel:
This commit adds two new endpoints to get power_levels for room members.
One is a V2 of the current room members endpoint, adding `membership` and `power_level` attributes for each user.
The other one exposes all power level attributes of the room using a new endpoint.
Both are missing pagination right now. Is this required for very large rooms?

Please comment how you would like to see such an API. @dklimpel @anoadragon453 

### Pull Request Checklist

### DIFF
--- a/changelog.d/9949.feature
+++ b/changelog.d/9949.feature
@@ -1,0 +1,1 @@
+Add room members admin API with membership and power levels. Contributed by Awesome Technologies Innovationslabor GmbH.
--- a/synapse/rest/admin/__init__.py
+++ b/synapse/rest/admin/__init__.py
@@ -43,6 +43,8 @@ from synapse.rest.admin.rooms import (
     MakeRoomAdminRestServlet,
     RoomEventContextServlet,
     RoomMembersRestServlet,
+    RoomMembersRestServletV2,
+    RoomPowerLevelsRestServlet,
     RoomRestServlet,
     RoomStateRestServlet,
     ShutdownRoomRestServlet,
@@ -217,6 +219,8 @@ def register_servlets(hs, http_server):
     RoomStateRestServlet(hs).register(http_server)
     RoomRestServlet(hs).register(http_server)
     RoomMembersRestServlet(hs).register(http_server)
+    RoomMembersRestServletV2(hs).register(http_server)
+    RoomPowerLevelsRestServlet(hs).register(http_server)
     DeleteRoomRestServlet(hs).register(http_server)
     JoinRoomAliasServlet(hs).register(http_server)
     PurgeRoomServlet(hs).register(http_server)
--- a/synapse/rest/admin/rooms.py
+++ b/synapse/rest/admin/rooms.py
@@ -330,6 +330,91 @@ class RoomMembersRestServlet(RestServlet):
         return 200, ret
 
 
+class RoomMembersRestServletV2(RestServlet):
+    """
+    Get members list of a room including power level for each user.
+    """
+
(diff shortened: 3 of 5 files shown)


awesome-manuel:
@dklimpel sorry for the missing doc. Nearly right.

The first API is `GET /_synapse/admin/v2/rooms/<room_id>/members` and includes also the "membership" attribute.
This means it returns more users than the old `members` endpoint, since also invited, banned, etc. members are listed.
Including the `displayname` would be possible, but in synapse-admin we typically have the user objects already in the cache and we can use a referenced attribute.

```json
{
  "members": [
    {
      "user_id": "@foo:matrix.org",
      "membership": "join",
      "power_level": 10
    },{
      "user_id": "@bar:matrix.org",
      "membership": "join",
      "power_level": 20
    },{
      "user_id": "@foobar:matrix.org",
      "membership": "invite",
      "power_level": 100
    }
  ],
  "total": 3
}
```

> The second API is `GET /_synapse/admin/v1/rooms/<room_id>/power_levels` with response of...

This is correct. It's simply what `synapse` has stored in it's database for that room. This means that not all members are listed, only those with non-default power level.

dklimpel:
IMO than the API needs a kind of filter. Perhaps something like `ban`, `kick`, `redact`, `invite`, `join`, `all` (default) and `active` (for all current members), and later potentially pagination.

awesome-manuel:
> IMO than the API needs a kind of filter. Perhaps something like `ban`, `kick`, `redact`, `invite`, `join`, `all` (default) and `active` (for all current members), and later potentially pagination.

That should be possible, though maybe combinations could be tricky, e.g. all invited AND joined members.
Maybe we could have both endpoints, remove the `power_level` from `members` and calculate the effective power level on client side.

richvdh:
> > The second API is `GET /_synapse/admin/v1/rooms/<room_id>/power_levels` with response of...
> 
> This is correct. It's simply what `synapse` has stored in it's database for that room. This means that not all members are listed, only those with non-default power level.

If you were to take this approach, I don't see any value in limiting it to `power_levels` - you may as well extend it so that you can ask for any state event - for instance by extending [`GET /_synapse/admin/v1/rooms/<room_id>/state`](https://github.com/matrix-org/synapse/blob/develop/docs/admin_api/rooms.md#room-state-api) to `GET /_synapse/admin/v1/rooms/<room_id>/state/<event_type>/<state_key>`.

I think that's probably easier than the v2 /members approach, but in case you want to pursue that instead/as well:

 * In terms of whether to expose `displayname`: if it's easy to do (and we already have the relevant information in the cache) then that's fine, but in general I would say that we should not expose additional attributes by default, since they will make the request more expensive to process. Instead, we could expose a parameter so that the caller can specify which attributes they require.

 * For consistency with the existing v1 /members API, I would be inclined to only include `joined` users by default - as @dklimpel says, a filter could be used to modify this if you need to.

 * Note that it's possible to assign non-default PLs to users who have never been a member of the room - so even if you include *all* users with an `m.room.membership` property, you might still miss some PL assignments - again this leads me to favour the other approach.

 * Ideally we'd at least consider pagination up-front, since retrofitting it is going to break compatibility. A shortcut might be just to return (say) the first 100 members for now: supporting more than 100 could happen in a future change.

clokep:
@awesome-manuel Do you have interest in continuing this PR?

richvdh:
Closing this as it appears to have been abandoned. If you're interested in finishing it up let us know and we can reopen.

PR #9951, opened 2021-05-07
Run cache_joined_hosts_for_event in background

erikjohnston:
This is because sometimes redis can take a 50ms to respond, and there's no point blocking the request waiting for the responses.

### DIFF
--- a/synapse/handlers/message.py
+++ b/synapse/handlers/message.py
@@ -19,6 +19,7 @@ from typing import TYPE_CHECKING, Any, Dict, List, Mapping, Optional, Tuple
 
 from canonicaljson import encode_canonical_json
 
+from twisted.internet import defer
 from twisted.internet.interfaces import IDelayedCall
 
 from synapse import event_auth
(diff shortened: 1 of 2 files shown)


richvdh:
> Can we run it in parallel with the rest of handle_new_client_event ?

Or is the whole point that we want to return to the client before the redis request completes?

synapse/handlers/message.py
```diff
@@ -979,9 +980,43 @@ async def handle_new_client_event(
...
+                        self._persist_event,
+                        requester=requester,
+                        event=event,
+                        context=context,
+                        ratelimit=ratelimit,
+                        extra_users=extra_users,
+                    ),
+                    run_in_background(
+                        self.cache_joined_hosts_for_event, event, context
+                    ),
```
richvdh, line 998: if this fails do we want to cancel `_persist_event` (and swallow the result if it completes successfully)? Maybe give it a `.errback` which just logs the error?

richvdh (approved):

PR #9953, opened 2021-05-07
add a cache to have_seen_event

richvdh:
Empirically, this helped my server considerably when handling gaps in Matrix HQ. The problem was that we would repeatedly call `have_seen_events` for the same set of (50K or so) `auth_events`, each of which would take many minutes to complete, even though it's only an index scan.

~~Now based on #9973.~~


synapse/storage/databases/main/events_worker.py
```diff
@@ -1055,23 +1056,46 @@ async def have_seen_events(self, event_ids):
...
 
-        for chunk in batch_iter((x for x in event_ids if x not in results), 100):
+        for chunk in batch_iter((x for x in event_ids if x not in cache_results), 100):
             await self.db_pool.runInteraction(
                 "have_seen_events", have_seen_events_txn, chunk
             )
+
         return results
 
+    @cached(max_entries=100000)
```
richvdh, line 1093: is this a sensible size? The entries are pretty tiny.

synapse/storage/databases/main/cache.py
```diff
@@ -168,6 +168,7 @@ def _invalidate_caches_for_event(
         backfilled,
     ):
         self._invalidate_get_event_cache(event_id)
+        self.have_seen_event.prefill((event_id,), True)
```
richvdh, line 171: we know what the value would be, so we can prefill - but would it be better just to `.invalidate()` ?

synapse/storage/databases/main/events_worker.py
erikjohnston: I think that's fine.

synapse/storage/databases/main/purge_events.py
```diff
@@ -283,6 +281,18 @@ def _purge_history_txn(
         # so make sure to keep this actually last.
         txn.execute("DROP TABLE events_to_purge")
 
+        for event_id, should_delete in event_rows:
+            self._invalidate_cache_and_stream(
+                txn, self._get_state_group_for_event, (event_id,)
+            )
+
+            # FIXME: this is racy - what if have_seen_event gets called between the
+            #    transaction completing and the invalidation running?
```
erikjohnston, line 290: I'm not sure there is much difference between that and calling `have_seen_event` just before we delete the event?

erikjohnston:
I think the idea is sane, have commented on the various `TODOs`

synapse/storage/databases/main/events_worker.py
```diff
@@ -1055,23 +1056,46 @@ async def have_seen_events(self, event_ids):
...
             sql = "SELECT event_id FROM events as e WHERE "
             clause, args = make_in_list_sql_clause(
                 txn.database_engine, "e.event_id", chunk
             )
             txn.execute(sql + clause, args)
-            results.update(row[0] for row in txn)
+            results.update({row[0]: True for row in txn})
 
-        for chunk in batch_iter((x for x in event_ids if x not in results), 100):
+        for chunk in batch_iter((x for x in event_ids if x not in cache_results), 100):
```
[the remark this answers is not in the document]
richvdh: Yeah, fair.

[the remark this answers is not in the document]
richvdh: Note to self: the main thing limiting this seems to be sqlite's [`SQLITE_LIMIT_VARIABLE_NUMBER`]( https://sqlite.org/c3ref/c_limit_attached.html#sqlitelimitvariablenumber) compile-time setting, which is 999 or so. In postgres, we build an array literal, which seems to be limited only by memory.

synapse/storage/databases/main/purge_events.py
richvdh: true. I've updated the comment accordingly.

synapse/storage/databases/main/purge_events.py
```diff
@@ -422,7 +432,14 @@ def _purge_room_txn(self, txn, room_id: str) -> List[int]:
         #       index on them. In any case we should be clearing out 'stream' tables
         #       periodically anyway (#5888)
 
-        # TODO: we could probably usefully do a bunch of cache invalidation here
+        # TODO: we could probably usefully do a bunch more cache invalidation here
```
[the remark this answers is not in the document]
richvdh: well, that sounded easy, and then led to a rabbithole of trying to `invalidate_many` over the cache invalidation stream, which is #9973.

richvdh:
ok this is finally ready for re-review, after a three-week hiatus while I sorted out invalidation of TreeCaches over replication.

The main difference is the use of a TreeCache, and hence invalidating by room when we purge a room.

synapse/storage/databases/main/events_worker.py
```diff
@@ -1045,32 +1046,74 @@ async def have_events_in_timeline(self, event_ids):
...
+        self, keys: Iterable[Tuple[str, str]]
+    ) -> Dict[Tuple[str, str], bool]:
+        """Helper for have_seen_events
+
+        Returns:
+             a dict {(room_id, event_id)-> bool}
+        """
+        # if the event cache contains the event, obviously we've seen it.
+        cache_results = {
+            eid for (_rid, eid) in keys if self._get_event_cache.contains(eid)
```
erikjohnston, line 1081:
```suggestion
            eid for (_rid, eid) in keys if self._get_event_cache.contains((eid,))
```

synapse/storage/databases/main/events_worker.py
```diff
@@ -1045,32 +1046,74 @@ async def have_events_in_timeline(self, event_ids):
...
+            #   index-only lookup on `events_event_id_key`.
+            sql = "SELECT event_id FROM events AS e WHERE"
             clause, args = make_in_list_sql_clause(
                 txn.database_engine, "e.event_id", chunk
             )
             txn.execute(sql + clause, args)
-            results.update(row[0] for row in txn)
-
-        for chunk in batch_iter((x for x in event_ids if x not in results), 100):
+            results.update({row[0]: True for row in txn})
```
erikjohnston, line 1097: Don't we need to include the room ID here?

synapse/storage/databases/main/events_worker.py
erikjohnston: Don't we also need to include the room ID in the result?

erikjohnston (changes requested):

synapse/storage/databases/main/events_worker.py
richvdh: yes we do.

synapse/storage/databases/main/events_worker.py
richvdh: yup.

erikjohnston (approved):

PR #9954, opened 2021-05-07
Sort child events according to MSC1772 rules.

clokep:
This (partially) implements the sorting rules from MSC1772 to hopefully ensure that different homeservers would return the same summary for a space.

The rules from MSC1772 say:

> The `order` key is a string which is used to provide a default ordering of siblings in the room list. (Rooms are sorted based on a lexicographic ordering of the Unicode codepoints of the characters in `order` values. Rooms with no `order` come last, in ascending numeric order of the `origin_server_ts` of their `m.room.create` events, or ascending lexicographic order of their `room_id`s in case of equal `origin_server_ts`.  `order`s which are not strings, or do not consist solely of ascii characters in the range `\x20` (space) to `\x7F` (`~`), or consist of more than 50 characters, are forbidden and the field should be ignored if received.)

Since we don't necessarily have the `origin_server_ts` of their `m.room.create` event (i.e. if it is a room from federation that we are not in) I've skipped that step (see https://github.com/matrix-org/matrix-doc/pull/2946#discussion_r627060830 for a conversation on this).


clokep:
> Since we don't necessarily have the `origin_server_ts` of their `m.room.create` event (i.e. if it is a room from federation that we are not in) I've skipped that step (see [matrix-org/matrix-doc#2946 (comment)](https://github.com/matrix-org/matrix-doc/pull/2946#discussion_r627060830)

I added some info about this to the MSC to bring this implementation inline with that.

richvdh (commented):
lgtm otherwise

PR #9958, opened 2021-05-10
Use ijson to parse the response to `/send_join`, reducing memory usage.

erikjohnston:
Instead of parsing the full response to `/send_join` into Python objects (which can be huge for large rooms) and *then* parsing that into events, we instead use ijson to stream parse the response directly into `EventBase` objects.

Question: Do we want to make this an optional dependency? It should be easy (just a case of adding a `JsonParser` subclass), though does mean we have more things to test

### DIFF
--- a/synapse/federation/transport/client.py
+++ b/synapse/federation/transport/client.py
@@ -17,13 +17,19 @@ import logging
 import urllib
 from typing import Any, Dict, List, Optional
 
+import attr
+import ijson
+
 from synapse.api.constants import Membership
 from synapse.api.errors import Codes, HttpResponseException, SynapseError
+from synapse.api.room_versions import RoomVersion
 from synapse.api.urls import (
     FEDERATION_UNSTABLE_PREFIX,
     FEDERATION_V1_PREFIX,
     FEDERATION_V2_PREFIX,
 )
+from synapse.events import EventBase, make_event_from_dict
+from synapse.http.matrixfederationclient import ByteParser
 from synapse.logging.utils import log_function
 from synapse.types import JsonDict
 
@@ -240,21 +246,36 @@ class TransportLayerClient:
         return content
 
     @log_function
-    async def send_join_v1(self, destination, room_id, event_id, content):
+    async def send_join_v1(
+        self,
+        room_version,
+        destination,
+        room_id,
+        event_id,
+        content,
+    ) -> "SendJoinResponse":
         path = _create_v1_path("/send_join/%s/%s", room_id, event_id)
 
         response = await self.client.put_json(
-            destination=destination, path=path, data=content
+            destination=destination,
+            path=path,
+            data=content,
+            parser=SendJoinParser(room_version, v1_api=True),
         )
 
         return response
 
     @log_function
-    async def send_join_v2(self, destination, room_id, event_id, content):
+    async def send_join_v2(
+        self, room_version, destination, room_id, event_id, content
+    ) -> "SendJoinResponse":
         path = _create_v2_path("/send_join/%s/%s", room_id, event_id)
 
         response = await self.client.put_json(
-            destination=destination, path=path, data=content
+            destination=destination,
+            path=path,
+            data=content,
+            parser=SendJoinParser(room_version, v1_api=False),
         )
 
         return response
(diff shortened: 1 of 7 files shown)


clokep:
A general question -- how strict is ijson at parsing JSON (does it bypass some of the strictness we do:

https://github.com/matrix-org/synapse/blob/03318a766cac9f8b053db2214d9c332a977d226c/synapse/util/__init__.py#L59-L60

synapse/http/matrixfederationclient.py
```diff
@@ -683,7 +756,41 @@ async def put_json(
...
+        path: str,
+        args: Optional[QueryArgs] = None,
+        data: Optional[JsonDict] = None,
+        json_data_callback: Optional[Callable[[], JsonDict]] = None,
+        long_retries: bool = False,
+        timeout: Optional[int] = None,
+        ignore_backoff: bool = False,
+        backoff_on_404: bool = False,
+        try_trailing_slash_on_400: bool = False,
+        parser: Optional[ByteParser] = None,
```
richvdh, line 792: docstring please

synapse/http/matrixfederationclient.py
```diff
@@ -148,13 +179,29 @@ def get_json(self) -> Optional[JsonDict]:
         return self.json
 
 
+class JsonParser(ByteParser[Union[JsonDict, list]]):
+    """A parser that buffers the response and tries and parses it as JSON."""
```
richvdh, line 183:
```suggestion
    """A parser that buffers the response and tries to parse it as JSON."""
```
```suggestion
    """A parser that buffers the response and tries and parses it as JSON."""
```

synapse/http/matrixfederationclient.py
```diff
@@ -168,19 +215,43 @@ async def _handle_json_response(
...
+async def _handle_response(
+    reactor: IReactorTime,
+    timeout_sec: float,
+    request: MatrixFederationRequest,
+    response: IResponse,
+    start_ms: int,
+    parser: ByteParser[T],
+) -> T:
+    """
+    Reads the JSON body of a response, with a timeout
```
richvdh, line 232:
```suggestion
    Reads the body of a response with a timeout and sends it to a parser
```
```suggestion
    Reads the JSON body of a response, with a timeout
```

synapse/http/matrixfederationclient.py
```diff
@@ -148,13 +179,29 @@ def get_json(self) -> Optional[JsonDict]:
...
+        self._binary_wrapper = BinaryIOWrapper(self._buffer)
+
+    def write(self, data: bytes) -> int:
+        return self._binary_wrapper.write(data)
+
+    def finish(self) -> Union[JsonDict, list]:
+        return json_decoder.decode(self._buffer.getvalue())
+
+
 async def _handle_json_response(
```
richvdh, line 198: given this is now only called in three places, maybe it would be better just to inline it (ie, replace it with `_handle_response`)

synapse/federation/transport/client.py
```diff
@@ -1052,3 +1073,67 @@ def _create_v2_path(path, *args):
...
+            )
+        else:
+            self._coro_state = ijson.items_coro(
+                _event_list_parser(room_version, self._response.state),
+                "state.item",
+            )
+            self._coro_auth = ijson.items_coro(
+                _event_list_parser(room_version, self._response.auth_events),
+                "auth_chain.item",
+            )
```
richvdh, line 1130:
```suggestion
        # The V1 API has the shape of `[200, {...}]`, which we handle by
        # prefixing with `item.*`.
        prefix = "item." if v1_api else ""
            
        self._coro_state = ijson.items_coro(
            _event_list_parser(room_version, self._response.state),
            prefix + "state.item",
        )
        self._coro_auth = ijson.items_coro(
            _event_list_parser(room_version, self._response.auth_events),
            prefix + "auth_chain.item",
        )
```

richvdh (commented):
generally lgtm, modulo a few thoughts.

As @clokep says: it would be good to check how it handles non-standard json (`inf`, `NaN`, etc).

erikjohnston:
> A general question -- how strict is ijson at parsing JSON (does it bypass some of the strictness we do:
> 
> https://github.com/matrix-org/synapse/blob/03318a766cac9f8b053db2214d9c332a977d226c/synapse/util/__init__.py#L59-L60

I tested this manually for all the ijson backends and they all reject `NaN`, `Infinity` and `-Infinity', so I think we're good on that front.

However, it will silently ignore if the `state_events` or `auth` keys are missing or have the wrong type, e.g. `list(ijson.items('{"foo": 1}', "foo.item"))` will produce `[]` rather than failing.

clokep (approved):
Seems reasonable. I wonder if there's other places we should be using this? (For some reason the receiving end of `/send` for federation jumps out to me.)

synapse/http/matrixfederationclient.py
```diff
@@ -88,6 +101,24 @@
...
 
+T = TypeVar("T")
+
+
+class ByteParser(ByteWriteable, Generic[T]):
+    """A `ByteWriteable` that has an additional `finish` function that returns
+    the parsed data.
+    """
+
+    CONTENT_TYPE = abc.abstractproperty()  # type: str  # type: ignore
```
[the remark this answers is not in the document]
erikjohnston: Yeah, mypy doesn't like abstractproperty :(

erikjohnston:
> Seems reasonable. I wonder if there's other places we should be using this? (For some reason the receiving end of `/send` for federation jumps out to me.)

Maybe, though I think `/send_join` is pretty much the only API (apart from deprecated `/state`) which is quite so big. `/send` only has a max of 50 events and 100 EDUs, so they're not *that* big :shrug:

PR #9965, opened 2021-05-11
Tests for to-device messages

richvdh:
Here are the tests that should have been in #9961.


clokep (approved):
LGTM. Thank you for writing tests. 🎉

PR #9966, opened 2021-05-11
Send the `m.room.create` stripped event with invites (support MSC1772).

clokep:
This was missed in #9915, MSC1772 did specify one change outside of the spaces summary endpoint, which is that the `m.room.create` event gets sent (as a stripped event) as part of `invite_state` (and `knock_state`, but #6739 isn't merged yet 🤷 ).

I think it is OK to just send it always now, so this simply removes the condition.

### DIFF
--- a/UPGRADE.rst
+++ b/UPGRADE.rst
@@ -92,7 +92,7 @@ Upgrading to v1.34.0
 -----------------------------------------------
 
 The ``room_invite_state_types`` configuration setting has been deprecated and
-replaced with ``disable_default_event_types`` and ``additional_event_types``.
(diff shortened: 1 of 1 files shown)


changelog.d/9966.feature
```diff
@@ -0,0 +1 @@
+Support [MSC1772](https://github.com/matrix-org/matrix-doc/pull/1772).
```
richvdh, line 1: I think we should be a bit more explicit about what this actually means for the average synapse sysadmin

richvdh (approved):
lgtm otherwise

PR #9968, opened 2021-05-11
Correctly ratelimit invites when creating a room

### LINKED ISSUE #9952: Appservices cannot create rooms with a large invite list
I thought this was fixed in #9663 but it's still a bug as of v1.33.0. Trying to create a room while inviting 7 users caused the homeserver to respond with `{ errcode: 'M_UNKNOWN', error: 'Cannot invite so many users at once' }`

babolivier:
Also allow ratelimiting for more than one action at a time.

Fixes #9952


synapse/api/ratelimiting.py
```diff
@@ -76,6 +77,7 @@ async def can_do_action(
             burst_count: How many actions that can be performed before being limited.
                 Overrides the value set during instantiation if set.
             update: Whether to count this check as performing the action
+            nb_actions: The number of actions the user wants to do.
```
richvdh, line 80: by the way I would just call this `n_actions`. `nb` isn't really an abbreviation I'm familiar with. It's not important though.

richvdh (commented):
`Ratelimiter` has some fairly detailed unit tests, so it would be good to extend them to cover the new parameter. OTOH that can certainly wait until after RC if you prefer

richvdh (approved):
lgtm, modulo the test failure

PR #9971, opened 2021-05-12
Updating Platform Prerequisites for macOS Big Sur (ARM Specific)

LukeWalsh:
I just installed synapse on a new Mac & didn't need brew for libffi, but, I did need brew for libjpeg. I'm wondering if libffi is now included in the Xcode developer tools? I'm also not sure why jpeg wasn't mentioned previously, but, perhaps this was simply missed.

Signed off by: Luke Walsh <me@lukewal.sh>

### Pull Request Checklist


babolivier:
Thanks for your contribution! Before we can review it, can you please have a look at the checklist bundled with the description? We require a few things before a PR can be considered which yours doesn't fulfill, namely:

* a [sign-off](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md#sign-off) in the pull-request description
* a [changelog file](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md#changelog)
* merging against the `develop` branch, not `master`

(the checklist also mentions running linters but I don't think this is necessary since this is not a code contribution)

LukeWalsh:
Done!

babolivier:
Thanks! However I'm afraid it's still not quite right:

* the sign-off needs to include an email address (as per the documentation linked above, please have a look through it for the exact format)
* the changelog file needs to end with a period (`.`) as per the checklist above

LukeWalsh:
maxresdefault

INSTALL.md
```diff
@@ -164,13 +164,26 @@ sudo dnf groupinstall "Development Tools"
...
+
+Install virtualenv with Pip:
+```sh
+sudo pip3 install virtualenv
+```
+
+Install Homebrew:
+```sh
+/bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
+```
```
clokep, line 182: We should probably link to Homebrew's install documentation instead of putting it inline. Putting a curl + `sh` execute is a bit meh.

clokep:
@LukeWalsh Can you include the error message you were seeing here? It would be helpful to verify the changes and in case someone else runs into the same issue and searches!

LukeWalsh:
It was an issue when running the main install with pip, Pillow was missing jpeg. There was of course a lot of console goop but I expect anyone would run into this, not sure whether the original author simply already had the jpeg library or if something changed with Big Sur, but, this was a fresh install on a new machine for me.

clokep:
> Where would you like me to include the reference to a failure in installing Pillow? Just here in the comments or also in the changelog file?

Just here in the PR was enough. 👍 

It sounds like you had to compile pillow yourself instead of installing a wheel? Overall I think this is OK, but a couple more questions:
* What version of Python are you using?
* What version of Pillow do you have installed in the virtualenv? (`pip freeze | grep -i pillow`)
* Do you have wheel installed in the virtualenv? (`pip freeze | grep -i wheel`)
* Are you on an ARM64 or x86-64 Mac?

LukeWalsh:
No, pPillow simply compiled recursively with the main package installation via pip (once jpeg was installed - prior to that the main pip installation excited with an error about Pillow). Pillow isn't able to include jpeg as it's a C library I suppose, but, once jpeg was installed everything just worked.

I'm using Python 3.8.2 in my virtualenv, Pillow 8.2.0, no wheel.

I'm using a Mac Mini with M1 chip (ARM64) running Mac OS Big Sur (11.2.3).

INSTALL.md
```diff
@@ -164,13 +164,21 @@ sudo dnf groupinstall "Development Tools"
 
 ##### macOS
 
-Installing prerequisites on macOS:
+Installing prerequisites on macOS (Big Sur):
```
clokep, line 167: It's confusing that this says "Big Sur" here, but then below it talks about Catalina.

clokep (commented):
I'm a bit confused by this -- I just created a new virtualenv, uninstalled libjpeg and had no problem. My guess is that since you're on an ARM Mac there's no wheel available for pillow so it has to built it from scratch, which is why you need libjpeg.

(I also did uninstall pkg-config and libffi and didn't have any issues.)

INSTALL.md
LukeWalsh: I didn't delete the instructions for Catalina and didn't check those, but, we could delete those entirely, or, keep them a bit until Big Sur is confidently free of security patches. Not sure how that would typically be handled in those docs, do you know?

LukeWalsh:
> I'm a bit confused by this -- I just created a new virtualenv, uninstalled libjpeg and had no problem. My guess is that since you're on an ARM Mac there's no wheel available for pillow so it has to built it from scratch, which is why you need libjpeg.
> 
> (I also did uninstall pkg-config and libffi and didn't have any issues.)

Did you do this on an Intel Mac? Perhaps the better solution here is for Pillow to handle the case where libjpeg isn't installed on ARM Macs?

INSTALL.md
clokep: Sorry for being unclear, my point was that I don't think we should add the ` (Big Sur)` here, we should just leave it as `macOS`.

clokep:
> > I'm a bit confused by this -- I just created a new virtualenv, uninstalled libjpeg and had no problem. My guess is that since you're on an ARM Mac there's no wheel available for pillow so it has to built it from scratch, which is why you need libjpeg.
> > (I also did uninstall pkg-config and libffi and didn't have any issues.)
> 
> Did you do this on an Intel Mac? Perhaps the better solution here is for Pillow to handle the case where libjpeg isn't installed on ARM Macs?

Sorry, yes I'm on an Intel Mac. So I think that's the main difference here, which makes me more OK with merging this. 👍

INSTALL.md
LukeWalsh: Fixed in the most recent commit.

LukeWalsh:
@clokep I moved the changes we were discussing in install.md to the new installation.md file. I also clarified that the section about libjpeg is only needed for ARM-based Macs & added more detail about the two lines I removed from the main section to 9971.doc. 

I believe both these removed references are only relevant for Mojave & older which seems unnecessary to mention as we are two OS versions ahead now: 1/ using easy_install to install pip as well as 2/ using brew to install libffi.

changelog.d/9971.doc
```diff
@@ -0,0 +1 @@
+Updated installation dependencies for ARM Macs & simplified overall Mac section by removing references to 1/ installing pip with easy_install (since pip3 comes with developer tools) & 2/ installing libffi with brew (since this only affects Mojave & older OS versions)!
```
clokep, line 1:
```suggestion
Updated installation dependencies for newer macOS versions and ARM Macs. Contributed by Luke Walsh.
```

clokep (approved):
Thanks for hanging in there!

PR #9973, opened 2021-05-12
Combine `LruCache.invalidate` and `invalidate_many`

richvdh:
I found myself wanting to stream an `invalidate_many` cache invalidation over the invalidation replication stream, and the easiest way seemed to be to make `invalidate` do the same thing as `invalidate_many`.


richvdh:
It might be helpful for the reviewer to remember that:
 * the top level of the caches is `DeferredCache`.
 * `DeferredCache.cache` is an `LruCache`
 * `LruCache.cache` is either a `dict` or a `TreeCache`
 * `DeferredCache` also contains a `_pending_deferred_cache` which will be the same type as `DeferredCache.cache.cache`

erikjohnston (approved):

PR #9974, opened 2021-05-12
Additional comments for space summary handler

clokep:
Just adds some docstrings / updates some comments. I pulled this out of #9922 to reduce that diff and highlight the changes from that PR better.

### DIFF
--- a/synapse/handlers/space_summary.py
+++ b/synapse/handlers/space_summary.py
@@ -231,11 +231,15 @@ class SpaceSummaryHandler:
         Generate a room entry and a list of event entries for a given room.
 
         Args:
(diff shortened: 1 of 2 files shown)


synapse/handlers/space_summary.py
```diff
@@ -32,7 +32,6 @@
 logger = logging.getLogger(__name__)
 
 # number of rooms to return. We'll stop once we hit this limit.
-# TODO: allow clients to reduce this with a request param.
```
clokep, line 35: The MSC has no mention of this, so I don't think mentioning it here makes sense.

erikjohnston (approved):

PR #9975, opened 2021-05-12
Minor `@cachedList` enhancements

richvdh:
- use a tuple rather than a list for the iterable that is passed into the
  wrapped function, for performance

- test that we can pass an iterable and that keys are correctly deduped.

### DIFF
--- a/synapse/storage/databases/main/user_erasure_store.py
+++ b/synapse/storage/databases/main/user_erasure_store.py
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from typing import Dict, Iterable
+
 from synapse.storage._base import SQLBaseStore
 from synapse.util.caches.descriptors import cached, cachedList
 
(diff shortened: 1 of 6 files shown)


synapse/storage/databases/main/user_erasure_store.py
```diff
@@ -37,21 +39,16 @@ async def is_user_erased(self, user_id: str) -> bool:
...
+            user_ids: full user ids to check
 
         Returns:
-            dict[str, bool]:
-                for each user, whether the user has requested erasure.
+            for each user, whether the user has requested erasure.
         """
-        # this serves the dual purpose of (a) making sure we can do len and
-        # iterate it multiple times, and (b) avoiding duplicates.
-        user_ids = tuple(set(user_ids))
```
anoadragon453, line 53: Just checking that this was OK to remove because `cachedList` effectively does these two things for us, correct?

anoadragon453 (approved):
lgtm, just had a quick question below. I suppose the name `cachedList` is now a bit misleading huh - not that it really matters too much.

synapse/storage/databases/main/user_erasure_store.py
richvdh: oh sorry, yes I meant to add a comment to say exactly that.

richvdh:
>  I suppose the name cachedList is now a bit misleading huh - not that it really matters too much.

I'm not entirely sure I follow you. I mean, it's not the *best* name ever, but there's nothing new about that, and I can't immediately come up with a hugely better one.

richvdh:
> I'm not entirely sure I follow you. I mean, it's not the best name ever, but there's nothing new about that, and I can't immediately come up with a hugely better one.

oh I guess your point is that it now passes a `tuple` into the cached function, where previously it passed a `list`. That's true as far as it goes, but in my head the `List` in `cachedList` didn't really refer to that, and was more about the fact that you could pass in a `list`. As it happens, you can pass in any `Iterable` (and always could). So yeah, not a great name, but that's not new.

PR #9977, opened 2021-05-13
Split multiplart email sending into a dedicated handler

babolivier:
In the context of making it available to the module API, cf #9884


synapse/handlers/send_email.py
```diff
@@ -0,0 +1,81 @@
...
+
+from synapse.logging.context import make_deferred_yieldable
+
+if TYPE_CHECKING:
+    from synapse.server import HomeServer
+
+logger = logging.getLogger(__name__)
+
+
+class SendEmailHandler:
```
anoadragon453, line 29: I initially wasn't sure about whether something that didn't immediately relate to processing based off an incoming HTTP request could be considered a `Handler`. But then I looked at `AuthHandler`, which is called from many random places, and realised the definition is definitely not as strict.

Besides that, I noticed that this function is explicitly handling sending Multipart emails. Should we update `send_email`/`SendEmailHandler` to include "Multipart" somewhere?

changelog.d/9977.misc
```diff
@@ -0,0 +1 @@
+Split multiplart email sending into a dedicated handler.
```
anoadragon453, line 1:
```suggestion
Split multipart email sending into a dedicated handler.
```

anoadragon453 (changes requested):
Some small notes below. My main question is: why can the module API not call `Mailer.send_email`? Perhaps because it would need to provide a text and HTML templates?

Essentially I don't currently know why this PR is necessary, even after looking at the comment chain in https://github.com/matrix-org/synapse/pull/9884#discussion_r626047364

synapse/handlers/send_email.py
babolivier: I'm not sure, since Synapse never sends non-multipart emails (and `Mailer` isn't named `MultipartMailer`).

babolivier:
> My main question is: why can the module API not call Mailer.send_email? Perhaps because it would need to provide a text and HTML templates?

I think there are two answers to this question. One is that `Mailer` is the push mailer, which is intended to only be used to send email push notifications (though I realise it's also used for 3pid management but I believe the right fix would be to move that code outside of the push notification code and to a more sensible location, not to pollute the push code even more with features that aren't related to push). The other is that the `Mailer` is created with the templates, but for modules it's more convenient to have the module load its templates, then in its own time do the substitutions it needs and then call `send_email`. I guess another solution could have been to split out `push.Mailer.send_mail` into two functions (still in the Mailer), and make the Mailer's `send_mail` function be a kind of proxy that adds a bunch of push logic and then call a second function to actually send the email, but I preferred killing two birds with one stone here.

babolivier:
> Essentially I don't currently know why this PR is necessary, even after looking at the comment chain in #9884 (comment)

The issue here, which this PR aims to fix, is that currently when we want to send non-push emails in Synapse we either pollute the push code with non-push stuff, or we end up copying the same snippet around.

anoadragon453 (approved):
Thank you for the explanation, that all makes sense! This lgtm then.

PR #9981, opened 2021-05-13
Run `black` on the scripts

richvdh:
Turns out these scripts weren't getting linted.


callahad (approved):
Feels like a step back in legibility, especially around the Docker invocations, but 🤷 better to be consistent?

PR #9982, opened 2021-05-13
Partially add type hints to synapse.util

clokep:
Some of the files in `synapse.util` had type hints, but weren't in `mypy.ini` so weren't being checked. 😢

### DIFF
--- a/synapse/util/iterutils.py
+++ b/synapse/util/iterutils.py
@@ -23,13 +23,14 @@ from typing import (
     Mapping,
     Sequence,
     Set,
+    Tuple,
     TypeVar,
 )
 
 T = TypeVar("T")
 
 
-def batch_iter(iterable: Iterable[T], size: int) -> Iterator[Collection[T]]:
+def batch_iter(iterable: Iterable[T], size: int) -> Iterator[Tuple[T, ...]]:
     """batch an iterable up into tuples with a maximum size
 
     Args:
--- a/synapse/util/iterutils.py
+++ b/synapse/util/iterutils.py
@@ -23,19 +23,18 @@ from typing import (
     Mapping,
     Sequence,
     Set,
-    Tuple,
     TypeVar,
 )
 
 T = TypeVar("T")
 
 
-def batch_iter(iterable: Iterable[T], size: int) -> Iterator[Tuple[T]]:
+def batch_iter(iterable: Iterable[T], size: int) -> Iterator[Collection[T]]:
     """batch an iterable up into tuples with a maximum size
 
     Args:
-        iterable (iterable): the iterable to slice
-        size (int): the maximum batch size
+        iterable: the iterable to slice
+        size: the maximum batch size
 
     Returns:
         an iterator over the chunks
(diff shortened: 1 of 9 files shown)


synapse/util/iterutils.py
```diff
@@ -23,19 +23,18 @@
...
     Set,
-    Tuple,
     TypeVar,
 )
 
 T = TypeVar("T")
 
 
-def batch_iter(iterable: Iterable[T], size: int) -> Iterator[Tuple[T]]:
+def batch_iter(iterable: Iterable[T], size: int) -> Iterator[Collection[T]]:
```
richvdh, line 32: this is fine, but I'm a bit curious about why it is needed. it definitely returns an iterator over `tuple`s.

synapse/util/iterutils.py
```diff
@@ -46,10 +45,10 @@ def batch_iter(iterable: Iterable[T], size: int) -> Iterator[Tuple[T]]:
     return iter(lambda: tuple(islice(sourceiter, size)), ())
 
 
-ISeq = TypeVar("ISeq", bound=Sequence, covariant=True)
+ISeq = TypeVar("ISeq", bound=Sequence)
 
 
-def chunk_seq(iseq: ISeq, maxlen: int) -> Iterable[ISeq]:
+def chunk_seq(iseq: ISeq, maxlen: int) -> Iterable[Iterable[ISeq]]:
```
richvdh, line 51: :confused: this doesn't look right

richvdh (commented):
sorry, I just realised you didn't flag this for review yet. I got overexcited.

synapse/util/iterutils.py
clokep: `Tuple[T]` means a tuple of a single element of type `T`, but the returned tuples return are (likely) longer than a single element.

synapse/util/iterutils.py
clokep: Oh, hmm...yeah it doesn't. I'm not really sure what the `bound` aspect of a `TypeVar` does. I'll have to look deeper into that.

Any particular reason this isn't using something similar to the above where it takes an `Iterable[T]`? (Well I guess it would need to be a `Collection` so you can call `len` on it.)

clokep: I updated this to not use `ISeq`, which I don't think is necessary.

synapse/util/iterutils.py
erikjohnston: FWIW `Tuple[T, ...]` means a variable length tuple of type `T`

clokep: I thought I tried that and it didn't work...I can try again though.

richvdh: hrrrm. I think we may have stuff that relies on it being specifically tuples, so I'm a bit hesitant to have an annotation that says the opposite, even if mypy is happy with it. As erik says, I'd expect `Iterator[Tuple[T...]]` to work.

clokep: Thanks for the feedback! It seems `Iteraor[Tuple[T, ...]]` worked!

richvdh (approved):
lgtm!

Issue #9794, opened 2021-04-12
default in encryption_enabled_by_default_for_room_type is misleading


Craeckie:
### Description

The option `encryption_enabled_by_default_for_room_type` as introduced in #7639 is misleading.

A default is something that has a value as long as it's not overwritten. In this case it would mean, that encryption is enabled for the specified room type as long as it is not disabled for a specific room/client/...
The current implementation prevents creation of unencrypted DMs/rooms. This is clearly not a _default_, but _forced_ encryption.

### Expected Naming

`force_encryption_enabled_for_room_type` or similar.

### Version information

`v1.31.0` and current `develop` branch.

anoadragon453:
Indeed, the description at least should be updated, if not the name as well.

A similar conclusion was reached at the end of https://github.com/matrix-org/synapse/issues/9246, but nobody ended up creating an issue off the back of it. Thanks for doing so!

FlorianJSa:
> force_encryption_enabled_for_room_type

Default is off so if you change this value it will overwrite the value

anoadragon453:
Heads up that #9794 is an on-going, initial attempt at addressing this issue.

richvdh:
itym #9983

PR #9983, opened 2021-05-13
#9794 default in encryption enabled by default for room type is misleading

jklippel:
Renamed configuration option `encryption_enabled_by_default_for_room_type` to `force_encryption_enabled_for_room_type` (#9794)


erikjohnston:
So we talked about this in `#synapse-dev:matrix.org` a bit, the problem with labelling it `force_*` is that its still possible to use the API to create rooms that aren't encrypted (by not specifying a `preset`), which would be surprising if the option `force_encryption_enabled_for_room_type` is enabled. 

I think the solution here is to change the behaviour so that specifying `ALL` actually does enable encryption for all rooms, ignoring the `preset`, i.e. instead of doing:

https://github.com/matrix-org/synapse/blob/4b965c862dc66c0da5d3240add70e9b5f0aa720b/synapse/config/room.py#L41-L46

we would instead always set the encrypted flag here (when the config says `ALL`):

https://github.com/matrix-org/synapse/blob/a683028d81606708f686b890c0a44f5a20b54798/synapse/handlers/room.py#L114-L119

richvdh:
I'm going to close this as it appears to have been abandoned. If you're interested in picking it up again, we can reopen it.

PR #9989, opened 2021-05-14
Update postgres docs

richvdh:


docs/postgres.md
```diff
@@ -33,28 +33,15 @@ Assuming your PostgreSQL database user is called `postgres`, first authenticate
...
-Then, create a user ``synapse_user`` with:
+Then, create a postgres user, and a database, with:
 
+    # this will prompt for a password for the new user
     createuser --pwprompt synapse_user
 
-Before you can authenticate with the `synapse_user`, you must create a
-database that it can access. To create a database, first connect to the
-database with your database user:
+    createdb --encoding=UTF8 --locale=C --template=template0 --owner=synapse_user synapse
```
richvdh, line 41: if we're using `createuser`, we may as well use `createdb` which is somewhat simpler to use on the commandline.

docs/postgres.md
```diff
@@ -63,79 +50,6 @@ You may need to enable password authentication so `synapse_user` can
 connect to the database. See
 <https://www.postgresql.org/docs/current/auth-pg-hba-conf.html>.
 
-If you get an error along the lines of `FATAL:  Ident authentication failed for
```
richvdh, line 66: I wanted to move this stuff out of the happy path, so it's now in "troubleshooting" below.

docs/postgres.md
```diff
@@ -63,79 +50,6 @@ You may need to enable password authentication so `synapse_user` can
...
-Synapse uses Postgres sequences to generate IDs for various tables. A sequence
-and associated table can get out of sync if, for example, Synapse has been
-downgraded and then upgraded again.
-
-To fix the issue shut down Synapse (including any and all workers) and run the
-SQL command included in the error message. Once done Synapse should start
-successfully.
-
-
-## Tuning Postgres
```
richvdh, line 120: likewise I've moved this down since it apparently it only affects "larger scale deployments".

docs/postgres.md
```diff
@@ -185,9 +123,8 @@ The script `synapse_port_db` allows porting an existing synapse server
 backed by SQLite to using PostgreSQL. This is done in as a two phase
 process:
 
-1.  Copy the existing SQLite database to a separate location (while the
-    server is down) and running the port script against that offline
-    database.
+1.  Copy the existing SQLite database to a separate location and run
+    the port script against that offline database.
```
richvdh, line 127: given this is just the overview, and the detailed steps below explain the process more clearly, I think we can do without the "while the server is down" bit here.

docs/postgres.md
```diff
@@ -33,28 +33,15 @@ Assuming your PostgreSQL database user is called `postgres`, first authenticate
...
-    CREATE DATABASE synapse
-     ENCODING 'UTF8'
-     LC_COLLATE='C'
-     LC_CTYPE='C'
-     template=template0
-     OWNER synapse_user;
-
-This would create an appropriate database named `synapse` owned by the
-`synapse_user` user (which must already have been created as above).
+The above will create a user called ``synapse_user`, and a database called
```
anoadragon453, line 43:
```suggestion
The above will create a user called `synapse_user`, and a database called
```

docs/postgres.md
```diff
@@ -33,28 +33,15 @@ Assuming your PostgreSQL database user is called `postgres`, first authenticate
     # Or, if your system uses sudo to get administrative rights
     sudo -u postgres bash
 
-Then, create a user ``synapse_user`` with:
+Then, create a postgres user, and a database, with:
```
anoadragon453, line 36:
```suggestion
Then, create a postgres user and a database with:
```

anoadragon453 (approved):
These look like excellent changes, thanks for taking the time to do them and to write out your thought process in the review comments.

docs/postgres.md
```diff
@@ -245,3 +182,60 @@ PostgreSQL database configuration file `homeserver-postgres.yaml`:
...
+
+### Fixing incorrect `COLLATE` or `CTYPE`
+
+Synapse will refuse to set up a new database if it has the wrong values of
+`COLLATE` and `CTYPE` set, and will log warnings on existing databases. Using
+different locales can cause issues if the locale library is updated from
+underneath the database, or if a different version of the locale is used on any
+replicas.
+
+The safest way to fix the issue is to dump the database and recreate it with
```
richvdh, line 225: this now includes the change from @babolivier in #9990

PR #9990, opened 2021-05-14
Specify which kind of dump admins should take when fixing a database

babolivier:
Because "taking a dump" has more than one meaning in English.


richvdh (commented):
this looks good, but is going to conflict with #9989, so I'll just merge it in there.

PR #9993, opened 2021-05-14
Remove `keylen` from `LruCache`.

richvdh:
`keylen` seems to be a thing that is frequently incorrectly set, and we don't really need it.

The only time it was used was to figure out if we had removed a subtree in `del_multi`, which we can do better by changing `TreeCache.pop` to return a different type (`TreeCacheNode`).

Commits should be independently reviewable.

### DIFF
--- a/synapse/storage/databases/main/client_ips.py
+++ b/synapse/storage/databases/main/client_ips.py
@@ -436,7 +436,7 @@ class ClientIpStore(ClientIpWorkerStore):
     def __init__(self, database: DatabasePool, db_conn, hs):
 
         self.client_ip_last_seen = LruCache(
-            cache_name="client_ip_last_seen", keylen=4, max_size=50000
+            cache_name="client_ip_last_seen", max_size=50000
         )
 
         super().__init__(database, db_conn, hs)
--- a/synapse/storage/databases/main/events_worker.py
+++ b/synapse/storage/databases/main/events_worker.py
@@ -157,7 +157,6 @@ class EventsWorkerStore(SQLBaseStore):
 
         self._get_event_cache = LruCache(
             cache_name="*getEvent*",
-            keylen=3,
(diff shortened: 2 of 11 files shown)


synapse/storage/databases/main/client_ips.py
```diff
@@ -436,7 +436,7 @@ class ClientIpStore(ClientIpWorkerStore):
     def __init__(self, database: DatabasePool, db_conn, hs):
 
         self.client_ip_last_seen = LruCache(
-            cache_name="client_ip_last_seen", keylen=4, max_size=50000
```
richvdh, line 439: should have been 3, ftr

synapse/storage/databases/main/events_worker.py
```diff
@@ -157,7 +157,6 @@ def __init__(self, database: DatabasePool, db_conn, hs):
 
         self._get_event_cache = LruCache(
             cache_name="*getEvent*",
-            keylen=3,
```
richvdh, line 160: and this one should have been 1.

synapse/util/caches/treecache.py
```diff
@@ -21,10 +46,23 @@ def __contains__(self, key):
...
+            # started.
+            raise ValueError("Cannot store TreeCacheNodes in a TreeCache")
+
         node = self.root
         for k in key[:-1]:
-            node = node.setdefault(k, {})
-        node[key[-1]] = _Entry(value)
+            next_node = node.get(k, SENTINEL)
+            if next_node is SENTINEL:
+                next_node = node[k] = TreeCacheNode()
```
richvdh, line 58: I've replaced `setdefault` with separate `get` and `set` to avoid making lots of redundant `TreeCacheNode` objects.

erikjohnston (approved):

PR #10002, opened 2021-05-17
Fix the allowed range of valid ordering characters for spaces.

clokep:
Per matrix-org/matrix-doc#3195.

### DIFF
--- a/synapse/handlers/space_summary.py
+++ b/synapse/handlers/space_summary.py
@@ -430,8 +430,8 @@ def _is_suggested_child_event(edge_event: EventBase) -> bool:
     return False
 
 
-# Order may only contain characters in the range of \x20 (space) to \x7F (~).
-_INVALID_ORDER_CHARS_RE = re.compile(r"[^\x20-\x7F]")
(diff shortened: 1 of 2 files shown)


synapse/handlers/space_summary.py
```diff
@@ -430,8 +430,8 @@ def _is_suggested_child_event(edge_event: EventBase) -> bool:
     return False
 
 
-# Order may only contain characters in the range of \x20 (space) to \x7F (~).
-_INVALID_ORDER_CHARS_RE = re.compile(r"[^\x20-\x7F]")
+# Order may only contain characters in the range of \x20 (space) to \x7E (~).
+_INVALID_ORDER_CHARS_RE = re.compile(r"[^\x20-\x7E]")
```
anoadragon453, line 434:
```suggestion
# Order may only contain characters in the range of \x20 (space) to \x7E (~) inclusive.
_INVALID_ORDER_CHARS_RE = re.compile(r"[^\x20-\x7E]")
```

This is implied in the MSC (and could probably use more clarification), but just while I have the ability to suggest it...

anoadragon453 (approved):

Issue #7737, opened 2020-06-23
Possible easy performance improvement: tuple/list literals -> set literals


joepie91:
### Description

In many places in the Synapse code, some variation of the following code exists:

```py
if foo in ("bar", "baz", "qux"):
# ... or ...
if foo in ["bar", "baz", "qux"]:
# ... or ...
STUFF = ["bar", "baz", "qux"]
if foo in STUFF:
```

These use tuple and list literals respectively, while the resulting tuples/lists are only ever used for presence checking of particular entries. Some quick microbenchmarking suggested, however, that set literals would be significantly faster here (relatively speaking), for both hits and misses:

```
$ python3 -m timeit '"nyet" in { "foo", "bar", "baz", 1, 2, 3, 4, 5, 1, 2, 3, 4, 5, 1, 2, 3, 4, 5 }'
10000000 loops, best of 5: 21.8 nsec per loop
$ python3 -m timeit '"baz" in { "foo", "bar", "baz", 1, 2, 3, 4, 5, 1, 2, 3, 4, 5, 1, 2, 3, 4, 5 }'
20000000 loops, best of 5: 18.1 nsec per loop

$ python3 -m timeit '"nyet" in [ "foo", "bar", "baz", 1, 2, 3, 4, 5, 1, 2, 3, 4, 5, 1, 2, 3, 4, 5 ]'
1000000 loops, best of 5: 276 nsec per loop
$ python3 -m timeit '"baz" in [ "foo", "bar", "baz", 1, 2, 3, 4, 5, 1, 2, 3, 4, 5, 1, 2, 3, 4, 5 ]'
5000000 loops, best of 5: 54 nsec per loop

$ python3 -m timeit '"nyet" in ( "foo", "bar", "baz", 1, 2, 3, 4, 5, 1, 2, 3, 4, 5, 1, 2, 3, 4, 5 )'
1000000 loops, best of 5: 264 nsec per loop
$ python3 -m timeit '"baz" in ( "foo", "bar", "baz", 1, 2, 3, 4, 5, 1, 2, 3, 4, 5, 1, 2, 3, 4, 5 )'
5000000 loops, best of 5: 53.8 nsec per loop
```

Even for very small collections:

```
$ python3 -m timeit '"bar" in { "foo", "bar" }'
20000000 loops, best of 5: 18.6 nsec per loop
$ python3 -m timeit '"bar" in [ "foo", "bar" ]'
10000000 loops, best of 5: 35.6 nsec per loop
[...]

PR #10007, opened 2021-05-18
Refactor checking restricted join rules

clokep:
Based on the conversation from https://github.com/matrix-org/synapse/pull/9922/files#r634207554 and a discussion between @erikjohnston and I.

This modifies the checks against the `restricted` join rules to raise an `AuthError` directly, instead of returning a boolean.

### DIFF
--- a/synapse/handlers/event_auth.py
+++ b/synapse/handlers/event_auth.py
@@ -14,6 +14,7 @@
 from typing import TYPE_CHECKING
 
 from synapse.api.constants import EventTypes, JoinRules
+from synapse.api.errors import AuthError
 from synapse.api.room_versions import RoomVersion
 from synapse.types import StateMap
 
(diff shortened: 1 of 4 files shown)


synapse/handlers/event_auth.py
```diff
@@ -29,44 +31,59 @@ class EventAuthHandler:
...
         Args:
             state_ids: The state of the room as it currently is.
             room_version: The room version of the room being joined.
             user_id: The user joining the room.
+            prev_member_event: The current membership event for this user.
 
-        Returns:
-            True if the user can join the room, false otherwise.
+        Raises:
+            AuthError if the user is cannot join the room.
```
sumnerevans, line 54:
```suggestion
            AuthError if the user cannot join the room.
```

synapse/handlers/event_auth.py
```diff
@@ -29,44 +31,59 @@ class EventAuthHandler:
...
         """
+        # If the member is invited or currently joined, then nothing to do.
+        if prev_member_event and (
+            prev_member_event.membership == Membership.JOIN
+            or prev_member_event.membership == Membership.INVITE
```
sumnerevans, line 59: maybe?

```suggestion
            prev_member_event.membership in (Membership.JOIN, Membership.INVITE)
```

sumnerevans (commented):
I'm really excited about this feature :)

erikjohnston (approved):
LGTM modulo @sumnerevans comments

synapse/handlers/event_auth.py
clokep: I purposefully didn't do this, but don't have a strong opinion on it. I'd really love to do it as a set comparison but that causes issues if `membership` is garbage data (and not hashable). 🤷

erikjohnston: Why would set comparison be better here? (I'm actually surprised to discover that a set comparison here *is* faster, given traditionally for small numbers of values lists are faster than sets :shrug: )

clokep: #7737 has some data.

clokep: (I didn't test which would be faster by the way! Just that we seem to be fairly inconsistent with how we do this...I also feel like we should pull out constant literals like this to avoid re-allocating them frequently...but maybe that's an over-optimization.)

PR #10009, opened 2021-05-18
Refactor the checking restricted join rules

clokep:
This is another bit split out of #9922 and builds on #10007. This performs the refactoring used by #9922 in the `EventAuthHandler` separately.

### DIFF
--- a/synapse/handlers/event_auth.py
+++ b/synapse/handlers/event_auth.py
@@ -11,7 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-from typing import TYPE_CHECKING, Optional
+from typing import TYPE_CHECKING, Collection, Optional
 
 from synapse.api.constants import EventTypes, JoinRules, Membership
 from synapse.api.errors import AuthError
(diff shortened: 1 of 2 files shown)


synapse/handlers/event_auth.py
```diff
@@ -93,13 +116,31 @@ async def check_restricted_join_rules(
...
+            result.append(space_id)
 
-        # The user was not in any of the required spaces.
-        raise AuthError(
-            403,
-            "You do not belong to any of the required spaces to join this room.",
-        )
+        return result
+
+    async def is_user_in_rooms(self, room_ids: Collection[str], user_id: str) -> bool:
```
clokep, line 123: I couldn't find a method on any of the handlers / stores that lets you check if a user is in a room already, arguably this might be better on the `RoomMemberHandler` or something, but it is convenient for the `EventAuthHandler` to have no dependencies.

clokep:
(I'm not 100% convinced this PR makes sense outside of #9922 now as it is hard to see that there will be other callers of these methods. We can abandon this and merge into #9922 if that's clearer.)

erikjohnston:
Thanks for this, it was actually helpful seeing it split up. I ended up referring back to #9922 to figure out how it was used in other places so I ended up just leaving comments there instead.

Issue #7083, opened 2020-03-16
Rooms with unsupported versions are shown as blank rooms


ara4n:
In the last ~4-5 days

`!VXChyxVNCBsWeSGcaN:sw1v.org`
`!eKTrlepAvBEhUIyoQV:jki.re`
`!bNAPsppHlSHCDvMRhh:abolivier.bzh`

ara4n:
(i haven't actually looked at my /sync response, but riot shows them to be entirely blank v1 rooms)

neilisfragile:
I also can't see ```!eKTrlepAvBEhUIyoQV:jki.re```

erikjohnston:
All three of those rooms use experimental room versions that have since been removed from synapse, I think its basically expected for those to go pop.

neilisfragile:
Right and it just about works okay when people have cached history, but switching over to a new client (like Riot electron) has this problem.

richvdh:
I'm going to suggest we triage this alongside other synapse bugs rather than treat it as a crisis.

clokep:
#7209 also points to some errors during initial sync for unknown room versions.

richvdh:
suggestion is that for now we exclude such rooms from initial /sync responses, thus hiding them on clients. (you'll have to flush your client cache, but we can live with that)

In future, if we drop production room versions (eg v1) we'll need to do better than this, but this will do as a first step.

ara4n:
how do I get back into these rooms, though? i can't seem to participate in them enough to follow the tombstone to the new version...

richvdh:
you don't. The whole point is that they were experimental room versions that are no longer supported. I guess you get someone to manually invite you to the new room.

anoadragon453:
Worth noting that deleting the room from `rooms` and the events from `events` and `events_json` clears them right up.

erikjohnston:
Currently `/purge_rooms` fails to delete these rooms, as users are in them. Might be worth fixing it so that it works?

jryans:
Personally, I would at least like to be able to leave these unsupported rooms, but currently even that is not possible... 😢 

Currently, the `/leave` request gives a 400 error with `M_UNSUPPORTED_ROOM_VERSION`. The unfortunate symptom of this is that the unleavable room is permanently at the top of my room list... 😭

clokep:
> Personally, I would at least like to be able to leave these unsupported rooms, but currently even that is not possible... 😢

[...]

PR #10011, opened 2021-05-18
Enable spaces by default.

clokep:
This enables the experimental spaces flag by default.

## Questions

* Do we want to do this for the multi-SSO providers and busy presence?
* Are there things we want to move behind a different flag (maybe the restricted rooms parts)?

### DIFF
--- a/changelog.d/10011.feature
+++ b/changelog.d/10011.feature
@@ -1,1 +1,1 @@
-Enable experimental support for [MSC2946](https://github.com/matrix-org/matrix-doc/pull/2946): Spaces Summary by default.
+Enable experimental support for [MSC2946](https://github.com/matrix-org/matrix-doc/pull/2946) (spaces summary API) and [MSC3083](https://github.com/matrix-org/matrix-doc/pull/3083) (restricted join rules) by default.
--- a/changelog.d/10011.feature
+++ b/changelog.d/10011.feature
@@ -1,0 +1,1 @@
+Enable experimental support for [MSC2946](https://github.com/matrix-org/matrix-doc/pull/2946): Spaces Summary by default.
--- a/docs/sample_config.yaml
+++ b/docs/sample_config.yaml
@@ -546,6 +546,21 @@ retention:
 #next_link_domain_whitelist: ["matrix.org"]
 
 
+# Enable experimental features in Synapse.
+#
+# Experimental features might break or be removed without a deprecation
+# period.
+#
+experimental_features:
+  # Support for Spaces (MSC1772), it enables the following:
+  #
+  # * The Spaces Summary API (MSC2946).
+  # * Restricting room membership based on space membership (MSC3083).
+  #
+  # Uncomment to disable support for Spaces.
+  #spaces_enabled: False
+
+
 ## TLS ##
 
 # PEM-encoded X509 certificate for TLS.
--- a/docs/sample_config.yaml
+++ b/docs/sample_config.yaml
@@ -558,7 +558,7 @@ experimental_features:
   # * Restricting room membership based on space membership (MSC3083).
   #
   # Uncomment to disable support for Spaces.
-  #spaces_enabled: False
+  #spaces_enabled: false
 
 
 ## TLS ##
--- a/docs/sample_config.yaml
+++ b/docs/sample_config.yaml
@@ -546,21 +546,6 @@ retention:
 #next_link_domain_whitelist: ["matrix.org"]
 
 
-# Enable experimental features in Synapse.
-#
-# Experimental features might break or be removed without a deprecation
-# period.
-#
-experimental_features:
-  # Support for Spaces (MSC1772), it enables the following:
-  #
-  # * The Spaces Summary API (MSC2946).
-  # * Restricting room membership based on space membership (MSC3083).
-  #
-  # Uncomment to disable support for Spaces.
-  #spaces_enabled: false
-
-
 ## TLS ##
 
 # PEM-encoded X509 certificate for TLS.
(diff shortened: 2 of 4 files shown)


clokep:
@callahad Is this what you had in mind from our conversation earlier?

callahad:
Sorry for the delay; needed to mull some things over.

Functionally, yes, I think this is what we want.

Since we're turning on `spaces_enabled` by default, I think we should also have it appear in the sample config (and docs, if there's an obvious place?) with an explanation of what it does. Which, to the best of my knowledge, is currently:

1. Enables the Spaces Summary API (MSC2946)
2. Enables restricting room membership based on space membership (MSC3083) (and thus the org.matrix.msc3083 experimental room version)

We should probably have a brief discussion before adding any additional MSCs under the `spaces_enabled` flag. I'm not sure _a priori_ what the right bias is there, so we can park it until there's a specific proposal.

clokep:
> Enables restricting room membership based on space membership (MSC3083) (and thus the org.matrix.msc3083 experimental room version)

This is maybe worth including, but I'm hesitant to talk about experimental room versions too loudly since they break in horrible ways when you remove them (see #7083).

> We should probably have a brief discussion before adding any additional MSCs under the spaces_enabled flag. I'm not sure *a priori* what the right bias is there, so we can park it until there's a specific proposal.

I don't think there's any plans to implement more at the moment?

I'll add some documentation.

Unfortunately neither of the questions from the description were answered. Could you take another look @callahad?

docs/sample_config.yaml
```diff
@@ -546,6 +546,21 @@ retention:
...
+# period.
+#
+experimental_features:
+  # Support for Spaces (MSC1772), it enables the following:
+  #
+  # * The Spaces Summary API (MSC2946).
+  # * Restricting room membership based on space membership (MSC3083).
+  #
+  # Uncomment to disable support for Spaces.
+  #spaces_enabled: False
```
clokep, line 561: This ended up higher up in the config file than might make sense? But the file is so big it is really only usable via searching anyway. So maybe this doesn't matter.

callahad:
> I don't think there's any plans to implement more [Spaces MSCs] at the moment?

None that I'm aware of, but I could imagine something coming down the pike like 
MSC2962: Managing power levels via Spaces or MSC3189: Per-room/per-space profile data and it seeming natural to toss it under `spaces_enabled`

> Unfortunately neither of the questions from the description were answered. Could you take another look @callahad?

I somehow skimmed over those, thanks for drawing attention back to those :)

> * Do we want to do this for the multi-SSO providers and busy presence?

I'll raise it with Product and have a concrete response by this time next week 

> * Are there things we want to move behind a different flag (maybe the restricted rooms parts)?

What do you think? _MSC2176: Update the redaction rules_ defines a `org.matrix.msc2176` room version which is currently [enabled by default](https://github.com/matrix-org/synapse/blob/v1.34.0/synapse/api/room_versions.py#L148-L149), but otherwise we just provide V1-6.

clokep:
> > * Do we want to do this for the multi-SSO providers and busy presence?
> 
> I'll raise it with Product and have a concrete response by this time next week

I think we can keep this PR just about spaces then, we can handle those separately.

> > * Are there things we want to move behind a different flag (maybe the restricted rooms parts)?
> 
> What do you think? _MSC2176: Update the redaction rules_ defines a `org.matrix.msc2176` room version which is currently [enabled by default](https://github.com/matrix-org/synapse/blob/v1.34.0/synapse/api/room_versions.py?rgh-link-date=2021-05-20T15%3A27%3A48Z#L148-L149), but otherwise we just provide V1-6.

I actually meant more specifically to spaces, if there's bits which we shouldn't expose by default yet. I think it is probably fine as is, since the restricted join rules stuff requires using an experimental room version.

docs/sample_config.yaml
anoadragon453: You can reorder where it appears by moving `ExperimentalConfig` around in this list:

https://github.com/matrix-org/synapse/blob/683d6f75af0e941e9ab3bc0a985aa6ed5cc7a238/synapse/config/homeserver.py#L58-L97

It probably should go somewhere on the bottom, but at the same time it probably doesn't matter too much.

clokep: I think it was originally this high up since it was used by `ApiConfig`, but that's no longer true (as of #9915). I'll give it a try!

synapse/config/experimental.py
```diff
@@ -29,9 +29,26 @@ def read_config(self, config: JsonDict, **kwargs):
...
-        self.spaces_enabled = experimental.get("spaces_enabled", False)  # type: bool
+        self.spaces_enabled = experimental.get("spaces_enabled", True)  # type: bool
         if self.spaces_enabled:
             KNOWN_ROOM_VERSIONS[RoomVersions.MSC3083.identifier] = RoomVersions.MSC3083
 
         # MSC3026 (busy presence state)
         self.msc3026_enabled = experimental.get("msc3026_enabled", False)  # type: bool
+
+    def generate_config_section(self, **kwargs):
+        return """\
```
anoadragon453, line 40: We may want to leave a note here stating that we only include experimental features that are on by default in the sample config.

changelog.d/10011.feature
```diff
@@ -0,0 +1 @@
+Enable experimental support for [MSC2946](https://github.com/matrix-org/matrix-doc/pull/2946): Spaces Summary by default.
```
anoadragon453, line 1: And MSC3083 as well?

anoadragon453, line 1: And MSC3083 as well?

anoadragon453 (approved):
Code looks good to me, and I think we've nailed down what other features we'd like to include here now.

synapse/config/experimental.py
clokep: Is that true though? This only includes the one enabled by default, but I'm unsure that's the *reason* it is the only one included.

clokep: I'm going to merge this as is for now, we can update it when (if?) we add more here.

PR #10013, opened 2021-05-18
Name release branches just after major.minor

callahad:
With the prior format, 1.33.0 / 1.33.1 / 1.33.2 got separate branches:

    release-v1.33.0
    release-v1.33.1
    release-v1.33.2

Under the new model, all three would share a common branch:

    release-v1.33

As before, RCs and actual releases exist as tags on these branches.

This better reflects our support model, e.g., that the "1.33" series had
a formal release followed by two patches / updates.

### Pull Request Checklist

### DIFF
--- a/docs/dev/git.md
+++ b/docs/dev/git.md
@@ -122,15 +122,15 @@ So, what counts as a more- or less-stable branch? A little reflection will show
 that our active branches are ordered thus, from more-stable to less-stable:
 
  * `master` (tracks our last release).
- * `release-vX.Y.Z` (the branch where we prepare the next release)<sup
+ * `release-vX.Y` (the branch where we prepare the next release)<sup
    id="a3">[3](#f3)</sup>.
  * PR branches which are targeting the release.
  * `develop` (our "mainline" branch containing our bleeding-edge).
  * regular PR branches.
 
 The corollary is: if you have a bugfix that needs to land in both
-`release-vX.Y.Z` *and* `develop`, then you should base your PR on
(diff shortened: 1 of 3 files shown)


callahad:
The internal release instructions have also been updated to reflect this change.

docs/dev/git.md
```diff
@@ -145,4 +145,4 @@ most intuitive name. [^](#a1)
 
 <b id="f3">[3]</b>: Very, very occasionally (I think this has happened once in
 the history of Synapse), we've had two releases in flight at once. Obviously,
-`release-v1.2.3` is more-stable than `release-v1.3.0`. [^](#a3)
+`release-v1.2` is more-stable than `release-v1.3`. [^](#a3)
```
clokep, line 148: I feel like this example doesn't really ring true anymore?

callahad: Agreed. I'm inclined to rewrite some other parts of that section (and regenerate the graph, etc.), but I didn't want this PR to fall victim to scope creep. Do you have any preferences for what should be in scope for this PR?

clokep: No, re-reading it I might rethink this though. With the caveat in the previous sentence of "we've had two release in flight" than it does make sense that `release-v1.2` would be more stable than `release-v1.3`. 

I guess I'm a bit nervous someone will read the second sentence and draw absolute claims...

callahad: Let's leave well enough alone and get back to this as part of the more general docs overhaul happening as part of the move to mdbook

erikjohnston (approved):

Issue #9962, opened 2021-05-11
presence_stream records get deleted


thermaq:
### Description

This may be me not understanding how presence_stream table is utilized, but I noticed dissapearing rows from presence_stream database table. More tangible effect may be that only one online user has a green dot showing online status in element and the rest doesn't get a dot at all.

My explanation:
- few people are online
- server wants to refresh their states invoking _update_presence_txn method in PresenceStore
- new online and offline states get inserted as rows to presence_stream
- we iterate through users we refreshed and delete old rows. The query in pseudocode look like `delete rows from presence_stream where stream_id < $stream_id and user_id is $user_id`. user_id is iterated over, but the stream_id is a remnant from previous loop
- so when we left previous loop the stream_id can be the maximum stream_id we inserted, causing every record we inserted to get deleted except the one with highest stream_id

I guess my question is: is it a bug or am I missing something? In case I'm right I have a pull request ready.

### Steps to reproduce
Turn on presence and look at presence_stream table table for few minutes with multiple online clients.

### Version information
Every version I used so far, develop too.

anoadragon453:
Yep, I believe you're correct! I wrote a quick test for this on a branch at https://github.com/matrix-org/synapse/commit/258bf0e722732ac53f10d65d4866bef0e69b2bfa and the results are that we only end up with one presence update in the database, rather than the 5 we desire.

We'd very happily accept a pull request that fixes the bug - thanks!

anoadragon453:
https://github.com/matrix-org/synapse/pull/10033 adds a regression test for the fix.

PR #10014, opened 2021-05-19
Fixed removal of new presence stream states

thermaq:
Signed-off-by: Marek Matys <themarcq@gmail.com>

https://github.com/matrix-org/synapse/issues/9962

This is a fix for above problem.

I fixed it by swaping the order of insertion of new records and deletion of old ones. This ensures that we don't delete fresh database records as we do deletes before inserts.

### DIFF
--- a/changelog.d/10014.bugfix
+++ b/changelog.d/10014.bugfix
@@ -1,1 +1,1 @@
-Fixed deletion of new presense stream states from database. 
+Fixed deletion of new presence stream states from database. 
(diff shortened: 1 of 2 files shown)


changelog.d/10014.bugfix
```diff
@@ -0,0 +1 @@
+Fixed deletion of new presense stream states from database.
```
anoadragon453, line 1:
```suggestion
Fixed deletion of new presence stream states from database. 
```

anoadragon453 (approved):
This passes the test I mentioned in #9962 :tada: 

I realise that we don't actually need `stream_id` to be the highest stream ID from `stream_orderings`. Really it just need to be 1 + the current stream ID, which everything in `stream_orderings` should at least be!

Thanks again!

PR #9945, opened 2021-05-06
Add config option to hide device names over federation

aaronraimist:
Now that cross signing exists there is much less of a need for other people to look at devices and verify them individually. This PR adds a config option to allow you to prevent device display names from being shared with other servers.

### Pull Request Checklist

* [x] Pull request is based on the develop branch
* [ ] Pull request includes a [changelog file](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md#changelog). The entry should:
  - Be a short description of your change which makes sense to users. "Fixed a bug that prevented receiving messages from other servers." instead of "Moved X method from `EventStore` to `EventWorkerStore`.".
  - Use markdown where necessary, mostly for `code blocks`.
  - End with either a period (.) or an exclamation mark (!).
  - Start with a capital letter.
* [x] Pull request includes a [sign off](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md#sign-off)
* [x] Code style is correct (run the [linters](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md#code-style))

PR #10015, opened 2021-05-19
Fix hiding devices names over federation

aaronraimist:
Follow up to https://github.com/matrix-org/synapse/pull/9945. Turns out, while that PR appears to work most of the time there is one more spot where device names can be sent over federation.

I also noticed all these blank `org.matrix.opentracing_context` lines while I was debugging why my first PR wasn't working. I assume those don't need to be there unless opentracing is enabled so I also prevented those from being sent if they were blank.

### Pull Request Checklist


erikjohnston:
Did you want review of this, or is it still a WIP?

aaronraimist:
It is still WIP but I think I'm going to need some help to finish it.

I tested this patch locally using the demo federation script and it seemed to work. But then I installed it on my server and device names were still being shared so this PR must not have fixed it. I haven't been able to figure out how the device names are being shared. I asked in `#synapse-dev:matrix.org` a few days ago but it got lost.

erikjohnston:
Shout if/when this is ready for review :slightly_smiling_face:

squahtx:
> I tested this patch locally using the demo federation script and it seemed to work. But then I installed it on my server and device names were still being shared so this PR must not have fixed it. I haven't been able to figure out how the device names are being shared. I asked in `#synapse-dev:matrix.org` a few days ago but it got lost.

Does the PR work now? I'm missing some context here.

aaronraimist:
@squahtx context is that I still have not had time to take another look and this PR still probably does not fix 100% of the places where devices names can leak. However this PR is better than nothing since it definitely leaks without this PR and [this feature is being turned on by default in 1.59.0rc1](https://github.com/matrix-org/synapse/releases/tag/v1.59.0rc1).

Unless someone else is able to put in some time to figure out all of the places where device names can leak then I would say just merge this and try to get it in to v1.59.0 so the feature at the top of the release notes is at least a bit more reliable.

Otherwise if this can't get in to v1.59.0, I would recommend postponing hiding device names over federation by default until someone has time to investigate it further.

Discussed at https://matrix.to/#/!XaqDhxuTIlvldquJaV:matrix.org/$A0Fo6Y5TQgS_mboU_QErLzehgW1xE9UE0hVcxQVeSUA?via=matrix.org&via=element.io&via=vector.modular.im

squahtx (approved):
Thanks for the explanation! I agree that it makes sense to land this fix as-is. I've filed #12750 to track the leakiness.

As part of future work, we could look into automated testing for the option. I also wonder if it would be more appropriate for the logic to sit in a higher layer than the data store.

LGTM

squahtx:
CI seems to be stuck from a year ago. Could you merge in the latest develop or a dummy commit to give it a kick?

squahtx (changes requested):
Merging in the latest develop branch has revealed a handful of test failures unfortunately.
We'll want to investigate and fix them before merging.

DMRobertson:
There is one failing sytest `Can query remote device keys using POST after notification` and one known flakey complement test. Otherwise no complaints.

DMRobertson:
I think this should pass (modulo complement flakes) with https://github.com/matrix-org/sytest/pull/1307.

anoadragon453:
Note that this does not look to solve https://github.com/matrix-org/synapse/issues/13114 (tested manually).

DMRobertson:
> Note that this does not look to solve #13114 (tested manually).

I'll update the changelog to reflect this.

PR #10017, opened 2021-05-19
Add a batching queue implementation.

erikjohnston:
This will be used during for the rewrite of the `KeyRing`.

### DIFF
--- a/synapse/util/batching_queue.py
+++ b/synapse/util/batching_queue.py
@@ -44,7 +44,9 @@ class BatchingQueue(Generic[V, R]):
     with all pending work (for a given key).
 
     The provided processing function will only be called once at a time for each
-    key.
+    key. It will be called the next reactor tick after `add_to_queue` has been
+    called, and will keep being called until the queue has been drained (for the
+    given key).
 
     Note that the return value of `add_to_queue` will be the return value of the
     processing function that processed the given item. This means that the
@@ -85,6 +87,11 @@ class BatchingQueue(Generic[V, R]):
     async def add_to_queue(self, value: V, key: Hashable = ()) -> R:
         """Adds the value to the queue with the given key, returning the result
         of the processing function for the batch that included the given value.
+
+        The optional `key` argument allows sharding the queue by some key. The
+        queues will then be processed in parallel, i.e. the process batch
+        function will be called in parallel with batched values from a single
+        key.
         """
 
         # First we create a defer and add it and the value to the list of
--- a/synapse/util/batching_queue.py
+++ b/synapse/util/batching_queue.py
@@ -1,0 +1,122 @@
+# Copyright 2021 The Matrix.org Foundation C.I.C.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import logging
+from typing import (
+    Awaitable,
+    Callable,
+    Dict,
+    Generic,
+    Hashable,
+    List,
+    Set,
+    Tuple,
+    TypeVar,
+)
+
+from twisted.internet import defer
+
+from synapse.logging.context import PreserveLoggingContext, make_deferred_yieldable
+from synapse.metrics.background_process_metrics import run_as_background_process
+from synapse.util import Clock
+
+logger = logging.getLogger(__name__)
+
+
+V = TypeVar("V")
+R = TypeVar("R")
+
+
+class BatchingQueue(Generic[V, R]):
+    """A queue that batches up work, calling the provided processing function
+    with all pending work (for a given key).
+
+    The provided processing function will only be called once at a time for each
+    key.
+
+    Note that the return value of `add_to_queue` will be the return value of the
+    processing function that processed the given item. This means that the
+    returned value will likely include data for other items that were in the
+    batch.
+    """
+
+    def __init__(
+        self, name: str, clock: Clock, process_items: Callable[[List[V]], Awaitable[R]]
+    ):
+        self._name = name
+        self._clock = clock
+
+        # The set of keys currently being processed.
+        self._processing_keys = set()  # type: Set[Hashable]
+
+        # The currently pending batch of values by key, with a Deferred to call
+        # with the result of the corresponding `process_items` call.
+        self._next_values = {}  # type: Dict[Hashable, List[Tuple[V, defer.Deferred]]]
+
+        # The function to call with batches of values.
+        self.process_items = process_items
+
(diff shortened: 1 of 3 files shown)


synapse/util/batching_queue.py
```diff
@@ -0,0 +1,137 @@
...
+    with all pending work (for a given key).
+
+    The provided processing function will only be called once at a time for each
+    key.
+
+    Note that the return value of `add_to_queue` will be the return value of the
+    processing function that processed the given item. This means that the
+    returned value will likely include data for other items that were in the
+    batch.
+    """
```
clokep, line 53: I feel like this is missing *when* the batch will be executed. Is it time-based? Item-based? Something else?

synapse/util/batching_queue.py
```diff
@@ -0,0 +1,137 @@
...
+        given key and call the `self.process_items` with the values.
+        """
+
+        try:
+            if key in self._processing_keys:
+                return
+
+            self._processing_keys.add(key)
+
+            while self._next_values:
```
clokep, line 113: Do we only care about `_next_values[key]` here? I'm a bit confused at when you want to loop here (I also think we're missing a test for incoming data while this is processing, or maybe that's what `test_queuing` is testing?)

synapse/util/batching_queue.py
```diff
@@ -0,0 +1,137 @@
...
+                    results = await self.process_items(values)
+
+                    for _, deferred in next_values:
+                        with PreserveLoggingContext():
+                            deferred.callback(results)
+
+                except Exception as e:
+                    for _, deferred in next_values:
+                        with PreserveLoggingContext():
+                            deferred.errback(e)
```
clokep, line 134: Could this cause a Deferred to have both `callback` and `errback` called on it if e.g. the second `deferred.callback` causes an error to be raised?

(I wonder if we should be guarding with `if not deferred.called`?)

synapse/util/batching_queue.py
```diff
@@ -0,0 +1,137 @@
...
+
+        # The set of keys currently being processed.
+        self._processing_keys = set()  # type: Set[Hashable]
+
+        # The currently pending batch of values by key, with a Deferred to call
+        # with the result of the corresponding `process_items` call.
+        self._next_values = {}  # type: Dict[Hashable, List[Tuple[V, defer.Deferred]]]
+
+        # The function to call with batches of values.
+        self.process_items = process_items
```
clokep, line 69: This can probably be private (prefixed with `_`) although I kind of feel like the name doesn't scream "this is a callable" to me. Maybe `items_processor` or `processing_callback` or something would be clearer?

synapse/util/batching_queue.py
```diff
@@ -0,0 +1,137 @@
...
+        )
+
+        LaterGauge(
+            "synapse_util_batching_queue_number_of_keys",
+            "The number of distinct keys that have items queued",
+            labelnames=("name",),
+            caller=lambda: len(self._next_values),
+        )
+
+    async def add_to_queue(self, value: V, key: Hashable = ()) -> R:
```
clokep, line 85: The default key here seems odd to me, is that so you don't have to provide a key (do we have a real use-case for that?)

synapse/util/batching_queue.py
```diff
@@ -0,0 +1,137 @@
...
+                # together requests that we're about to receive. A common
+                # pattern is to call `add_to_queue` multiple times at once, and
+                # deferring to the next reactor tick allows us to batch all of
+                # those up.
+                await self._clock.sleep(0)
+
+                next_values = self._next_values.pop(key, [])
+
+                try:
+                    values = [value for value, _ in next_values]
```
clokep, line 124: A fun trick you can do is `values, deferreds = zip(*next_values)`, doesn't really save anything here though unfortunately since you have to iterate all the `Deferreds` anyway.

clokep:
This seems to have some failing tests too?

synapse/util/batching_queue.py
erikjohnston: Woops. The class used to be implemented without the ability to shard by key, you see. I think that this technically still works, it'll just spin until we have exhausted all the queues.

synapse/util/batching_queue.py
erikjohnston: Yes, its so that you don't have to provide a key, and it'll degrade into a single queue. I'll added a comment. (We do have a use case for this.)

synapse/util/batching_queue.py
clokep: Yeah it should work, will just spin unnecessarily!

erikjohnston: ... I failed to finish my sentence :facepalm: 

> it'll just spin until we have exhausted all the queues *so I'm not sure how easy it is to write a test to catch this*.

synapse/util/batching_queue.py
erikjohnston: I don't think `deferred.callback` should ever explode, but may as well add the guard.

erikjohnston:
> This seems to have some failing tests too?

Serves me right for not running the tests after adding the metrics!

synapse/util/batching_queue.py
erikjohnston: I've added some words, not sure if they make sense.

synapse/util/batching_queue.py
clokep: I would hope not. Another option would be to move the callback calls to an `else` so that only calling the processing function is in the try clause?

synapse/util/batching_queue.py
```diff
@@ -0,0 +1,153 @@
...
+        # If we're not currently processing the key fire off a background
+        # process to start processing.
+        if key not in self._processing_keys:
+            run_as_background_process(self._name, self._process_queue, key)
+
+        return await make_deferred_yieldable(d)
+
+    async def _process_queue(self, key: Hashable) -> None:
+        """A background task to repeatedly pull things off the queue for the
+        given key and call the `self.process_batch_callback` with the values.
```
clokep, line 114:
```suggestion
        given key and call the `self._process_batch_callback` with the values.
```

clokep (approved):
LGTM!

synapse/util/batching_queue.py
erikjohnston: I think I prefer it as it is, as then we really should always call either `.errback` or `.callback` for each?

clokep: Right, isn't that what my suggestion guarantees though?

```python
try:
    result = processing_callback(values)
except Exception as e:
    for d in deferreds:
        d.errback(e)
else:
    for d in deferreds:
        d.callback(result)
```

erikjohnston: But if `d.callback` never throws then that's the same as

```python
try:
    result = processing_callback(values)

    for d in deferreds:
        d.callback(result)
except Exception as e:
    for d in deferreds:
        d.errback(e)
```

?

clokep: Yes. 😄 

If it *does* throw it has the benefit of causing a stack trace and letting us know about it.

clokep: OK, you can ignore me. It can't throw (for the reason I thought it could):

https://github.com/twisted/twisted/blob/cae365213868e8226a055f6b5e4d2c0c89c9ab3b/src/twisted/internet/defer.py#L851-L871

PR #10018, opened 2021-05-19
Add `Keyring.verify_events_for_server` and reduce memory usage

erikjohnston:
Also add support for giving a callback to generate the JSON object to
verify. This should reduce memory usage, as we no longer have the event
in memory in dict form (which has a large memory footprint) for extend
periods of time.

### DIFF
--- a/synapse/crypto/keyring.py
+++ b/synapse/crypto/keyring.py
@@ -229,11 +229,32 @@ class Keyring:
         )
 
     def verify_events_for_server(
-        self, server_and_json: Iterable[Tuple[str, EventBase, int]]
+        self, server_and_events: Iterable[Tuple[str, EventBase, int]]
     ) -> List[defer.Deferred]:
+        """Bulk verification of signatures on events.
+
+        Args:
+            server_and_events:
+                Iterable of `(server_name, event, validity_time)` tuples.
+
+                `server_name` is which server we are verifying the signature for
+                on the event.
+
+                `event` is the event that we'll verify the signatures of for
+                the given `server_name`.
+
+                `validity_time` is a timestamp at which the signing key must be
+                valid.
+
+        Returns:
+            List<Deferred[None]>: for each input triplet, a deferred indicating success
+                or failure to verify each event's signature for the given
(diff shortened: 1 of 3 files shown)


synapse/crypto/keyring.py
```diff
@@ -69,13 +71,15 @@ class VerifyJsonRequest:
...
+        json_object_callback: A callback to fetch the JSON object to verify.
+            A callback is used to allow deferring the creation of the JSON
+            object to verify until needed, e.g. for events we can defer
+            creating the redacted copy. This reduces the memory usage when
+            there are large numbers of in flight requests.
 
         minimum_valid_until_ts: time at which we require the signing key to
             be valid. (0 implies we don't care)
 
-        request_name: The name of the request.
```
clokep, line 77: This seems to still be an attribute?

synapse/crypto/keyring.py
```diff
@@ -175,10 +220,20 @@ def verify_json_objects_for_server(
...
-            VerifyJsonRequest(server_name, json_object, validity_time, request_name)
+            VerifyJsonRequest.from_json_object(
+                server_name, json_object, validity_time, request_name
+            )
             for server_name, json_object, validity_time, request_name in server_and_json
         )
 
+    def verify_events_for_server(
+        self, server_and_json: Iterable[Tuple[str, EventBase, int]]
+    ) -> List[defer.Deferred]:
```
clokep, line 231: Docstring please! (Especially for the `server_and_json` parameter).

synapse/crypto/keyring.py
```diff
@@ -69,13 +71,15 @@ class VerifyJsonRequest:
     Attributes:
         server_name: The name of the server to verify against.
 
-        json_object: The JSON object to verify.
+        json_object_callback: A callback to fetch the JSON object to verify.
```
clokep, line 74: Maybe `get_json_object` would be a clearer name?

clokep (commented):
Looks good overall, comments are all minor. :)

clokep (approved):
Looks good if tests pass!

PR #10035, opened 2021-05-21
Rewrite the KeyRing

### LINKED ISSUE #3825: server-key fetching logic is slow and queue-bound
There are a number of problems with the logic in `keyring.py` for fetching server keys:

* each request that needs a key for a given server gets queued up, and it's possible to end up with quite a long queue for a given server. If the lookup is successful, that's ok. However, if it fails (which may take many minutes while we wait for timeouts), then we try again for each request in the queue - so we can rapidly end up getting very badly behind. When we want key X for server Y, if there is already a request in the queue for that key, then we should just use the results from it, even if it fails.

* relatedly, the queueing logic might never complete. If a given request wants keys from server A and server B, and a lookup is already in progress for A, it waits for that to complete. By that time, another request might be doing a lookup for B, so it waits for that to complete. Then we might be waiting for A again. etc. We should immediately start lookups for those servers which aren't already in progress, rather than waiting for the complete set.

* see also #3819 and #3818

erikjohnston:
Sorry for a bit of a mammoth PR, but this is a total rewrite of the `KeyRing` class to use `BatchingQueue` rather than a mess of deferreds. The idea here is that we write the functions out in a way to handle each server key request separately, and then rely on `BatchingQueue` to merge requests that happen at the same when querying the keys from the database or perspective key servers.

It's probably best to review commit by commit, and for the first to ignore the diff of the `KeyRing` class and instead look at the end result by itself: https://github.com/matrix-org/synapse/blob/f941830e35db49395af224bd7c696435394a1490/synapse/crypto/keyring.py#L281-L447

Fixes #3825


synapse/crypto/keyring.py
```diff
@@ -193,40 +190,42 @@ def verify_json_for_server(
             server_name,
             json_object,
             validity_time,
-            request_name,
         )
-        requests = (request,)
-        return make_deferred_yieldable(self._verify_objects(requests)[0])
+        return defer.ensureDeferred(self.process_request(request))
```
richvdh, line 194: why not make this an async function, rather than messing with `ensureDeferred`? Afaik the only reason it was defined to return a Deferred previously was that the keyring code was too scary to turn into async (all the callers, with the exception of the tests, will accept an async)

synapse/crypto/keyring.py
```diff
@@ -193,40 +190,42 @@ def verify_json_for_server(
...
             List<Deferred[None]>: for each input triplet, a deferred indicating success
                 or failure to verify each json object's signature for the given
                 server_name. The deferreds run their callbacks in the sentinel
                 logcontext.
         """
-        return self._verify_objects(
-            VerifyJsonRequest.from_json_object(
-                server_name, json_object, validity_time, request_name
+        return [
+            defer.ensureDeferred(
```
richvdh, line 217: `run_in_background` returns a`Deferred`, and that's not likely to change in the short term. I'd skip this.

synapse/crypto/keyring.py
```diff
@@ -252,321 +251,225 @@ def verify_events_for_server(
...
-                verify_request.request_name,
-                verify_request.server_name,
-                verify_request.key_ids,
-                verify_request.minimum_valid_until_ts,
+        if not verify_request.key_ids:
+            raise SynapseError(
+                400,
+                f"Not signed by {verify_request.server_name}",
+                Codes.UNAUTHORIZED,
             )
```
richvdh, line 277: this is orthogonal, since I don't think you're changing it, but I hate the fact we are picking an HTTP response code this deep in library functions. Ideally we'd define a handful of Exception classes which can be raised by the keyring code, and turn them back into HTTP responses higher up the stack.

synapse/crypto/keyring.py
```diff
@@ -252,321 +251,225 @@ def verify_events_for_server(
...
 
-            # add the key request to the queue, but don't start it off yet.
-            key_lookups.append(verify_request)
+        # Add they keys we need to verify to the queue for retrieval. We queue
```
richvdh, line 279:
```suggestion
        # Add the keys we need to verify to the queue for retrieval. We queue
```

synapse/crypto/keyring.py
```diff
@@ -252,321 +251,225 @@ def verify_events_for_server(
...
+        found_keys = found_keys_by_server.get(verify_request.server_name, {})
+
+        # For each signature to check we ensure we have fetched the necessary
+        # keys and the signature matches.
+        for key_id in verify_request.key_ids:
+            key_result = found_keys.get(key_id)
+            if not key_result:
+                raise SynapseError(
+                    401,
+                    f"Missing keys for {verify_request.server_name}: {key_id}",
```
richvdh, line 297:
```suggestion
                    f"Failed to retrieve key {key_id} for {verify_request.server_name}",
```

synapse/crypto/keyring.py
```diff
@@ -252,321 +251,225 @@ def verify_events_for_server(
...
+                    401,
+                    f"Failed to find key with recent enough `valid_until_ts` for {verify_request.server_name}: {key_id}",
+                    Codes.UNAUTHORIZED,
+                )
 
-        if key_lookups:
-            run_in_background(self._start_key_lookups, key_lookups)
+            verify_key = key_result.verify_key
+            try:
+                json_object = verify_request.get_json_object()
```
richvdh, line 310: has this moved inside the `try/except` deliberately? Not that I think it makes a lot of difference.

synapse/crypto/keyring.py
```diff
@@ -252,321 +251,225 @@ def verify_events_for_server(
...
+        for request in requests:
+            by_server = server_to_key_to_ts.setdefault(request.server_name, {})
+            for key_id in request.key_ids:
+                existing_ts = by_server.get(key_id)
+
+                if existing_ts:
+                    by_server[key_id] = max(request.minimum_valid_until_ts, existing_ts)
+                else:
+                    by_server[key_id] = request.minimum_valid_until_ts
+
```
richvdh, line 357:
```suggestion
                existing_ts = by_server.get(key_id, 0)
                by_server[key_id] = max(request.minimum_valid_until_ts, existing_ts)

```

or, if you prefer:

```suggestion
                existing_ts = by_server.get(key_id)

                if existing_ts is not None:
                    by_server[key_id] = max(request.minimum_valid_until_ts, existing_ts)
                else:
                    by_server[key_id] = request.minimum_valid_until_ts

```

right now it's neither one thing nor the other

synapse/crypto/keyring.py
```diff
@@ -74,23 +74,19 @@ class VerifyJsonRequest:
         minimum_valid_until_ts: time at which we require the signing key to
             be valid. (0 implies we don't care)
 
-        request_name: The name of the request.
-
```
richvdh, line 84: hrm. this served a purpose: it was the thing that told us which event was failing verification.

richvdh (changes requested):
well, it's certainly much clearer now.

I've got a few concerns though. The big one is about what performance of concurrent fetches of keys when one key is quick and another is slow.

I'm also a bit concerned about dropping the event id from the logs when verification fails.

synapse/crypto/keyring.py
```diff
@@ -252,321 +251,225 @@ def verify_events_for_server(
...
+
+        logger.debug("Deduplicated key requests to %s", deduped_requests)
+
+        # For each key we call `_inner_verify_request` which will handle
+        # fetching each key. Note these shouldn't throw if we fail to contact
+        # other servers etc.
+        results_per_request = await yieldable_gather_results(
+            self._inner_fetch_key_request,
+            deduped_requests,
+        )
```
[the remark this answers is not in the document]
erikjohnston: I thought that was basically the case now, where we blocked waiting for previous look ups for keys for a server?

I think we could move the deduplication into the key fetchers themselves? But that doesn't really help much since we're batching/queuing up request for keys for a particular server anyway.

synapse/crypto/keyring.py
erikjohnston: Agreed

synapse/crypto/keyring.py
erikjohnston: It was only used for some debug logging AFAICT? I can add that back in if you think its useful

synapse/crypto/keyring.py
erikjohnston: No, and its a bit confusing that it is in there. I've moved it out

synapse/crypto/keyring.py
```diff
@@ -135,16 +115,38 @@ def from_event(
...
     pass
 
 
+@attr.s(slots=True)
+class _FetchKeyRequest:
+    """A request for keys for a given server."""
+
+    server_name = attr.ib(type=str)
+    minimum_valid_until_ts = attr.ib(type=int)
+    key_ids = attr.ib(type=List[str])
```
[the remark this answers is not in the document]
erikjohnston: This does feel wrong to me, and re-reading the previous code I think we actually did only require one signature to be valid?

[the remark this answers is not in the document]
erikjohnston: Could you have a quick look to see if you agree that that's the case, and then I'll fix this PR to do the same (and document it at the same time)

synapse/crypto/keyring.py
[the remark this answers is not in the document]
richvdh: > I thought that was basically the case now, where we blocked waiting for previous look ups for keys for a server?

huh. yes. Ok, ignore this then.

synapse/crypto/keyring.py
richvdh: eh, it probably doesn't matter. It looks like we stick the event id back into the error message further up the stack.

synapse/crypto/keyring.py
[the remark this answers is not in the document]
erikjohnston: Pushed the change. The code works slightly differently now where we try and find *all* the keys before running the verifications, and also bails if either a) any of the keys we find fail verification or b) we don't find any keys.

I don't think the change matters much, as I think its pretty rare for a server to sign the same event multiple times?

synapse/crypto/keyring.py
```diff
@@ -135,7 +135,18 @@ class KeyLookupError(ValueError):
...
+    """A request for keys for a given server.
+
+    We will continue to try and fetch until we have all the keys listed under
+    `key_ids` (with an appropriate `valid_until_ts` property) or we run out of
+    places to fetch keys from.
+
+    Attributes:
+        server_name: The name of the server that owns the keys.
+        minimum_valid_until_ts: The earliest timestamp at which we need the
+            keys to be valid at.
```
richvdh, line 147: this is a particularly mind-bending way of defining this, though I agree it's not easy.

```suggestion
        minimum_valid_until_ts: The timestamp which the keys must be valid until.
```

synapse/crypto/keyring.py
```diff
@@ -166,12 +177,12 @@ def __init__(
...
         )  # type: BatchingQueue[_FetchKeyRequest, Dict[str, Dict[str, FetchKeyResult]]]
 
-    def verify_json_for_server(
+    async def verify_json_for_server(
         self,
         server_name: str,
         json_object: JsonDict,
         validity_time: int,
-    ) -> defer.Deferred:
+    ):
```
richvdh, line 185:
```suggestion
    ) -> None:
```

synapse/crypto/keyring.py
```diff
@@ -191,7 +202,7 @@ def verify_json_for_server(
             json_object,
             validity_time,
         )
-        return defer.ensureDeferred(self.process_request(request))
+        return await self.process_request(request)
```
richvdh, line 205: docstring needs an update: no longer returns a `Deferred`

tests/crypto/test_keyring.py
```diff
@@ -611,21 +610,9 @@ def get_key_id(key):
...
-def run_in_context(f, *args, **kwargs):
-    with LoggingContext("testctx"):
-        rv = yield f(*args, **kwargs)
-    return rv
-
-
-def _verify_json_for_server(kr, *args):
+async def _verify_json_for_server(kr, *args):
     """thin wrapper around verify_json_for_server which makes sure it is wrapped
     with the patched defer.inlineCallbacks.
```
richvdh, line 615: it doesn't do this any more. It might be completely redundant?

richvdh (approved):
lgtm other than the nits below

PR #10036, opened 2021-05-21
Don't hammer the database for destination retry timings every ~5mins

erikjohnston:
We used an expiring cache to store the retry timings for remote servers as at the time we didn't have a reliable way of invalidating caches from workers, so we just expired the caches every 5mins. This has the disadvantage that every 5mins the server will do 100s of queries to get the retry timings of all servers its sending too.

Now that we have a reliable way of invalidating the caches we should use a standard cache.


synapse/util/retryutils.py
```diff
@@ -82,10 +82,10 @@ async def get_retry_limiter(destination, clock, store, ignore_backoff=False, **k
...
 
     if retry_timings:
-        failure_ts = retry_timings["failure_ts"]
+        failure_ts = retry_timings.failure_ts
         retry_last_ts, retry_interval = (
-            retry_timings["retry_last_ts"],
-            retry_timings["retry_interval"],
+            retry_timings.retry_last_ts,
+            retry_timings.retry_interval,
         )
```
clokep, line 89: I know this code was here before, but it seems simpler to write it as:

```python
retry_last_ts = retry_timings.retry_last_ts
retry_interval = retry_timings.retry_interval
```

I guess this is trying to match the code from above, but it seems more confusing to me.

synapse/storage/databases/main/transactions.py
```diff
@@ -145,7 +147,11 @@ async def set_received_txn_response(
             desc="set_received_txn_response",
         )
 
-    async def get_destination_retry_timings(self, destination):
+    @cached(max_entries=10000)
```
clokep, line 150: 10k destinations seems high, but maybe it isn't? (Did you just grab a random number or did you check the previous size of the cache?)

changelog.d/10036.misc
```diff
@@ -0,0 +1 @@
+Don't hammer the database for destination retry timings every ~5mins.
```
clokep, line 1: Maybe?

```suggestion
Properly invalidate caches for destination retry timings every (instead of expiring entries every 5 minutes).
```

Although still not descriptive to understand if it will fix anything. 🤷

clokep (approved):

synapse/storage/databases/main/transactions.py
erikjohnston: The flip side was that 1000 felt a bit small. I mostly just pulled it out of thin air, and it matches some other "small" caches.

clokep: Ah, yeah I guess 1000 doesn't feel big enough. 👍 Was mostly curious where it came from. Thin air seems OK to me. ✈️

PR #10038, opened 2021-05-21
Remove unused properties on SpaceSummaryHandler

clokep:
Follow-up to #9922 to remove (now) unused properties on the handler.


clokep:
(I wonder how many other handlers have unused properties. 😢 )

erikjohnston (approved):

PR #9086, opened 2021-01-12
Add an admin API endpoint to protect media.

clokep:
Fixes #9083. 

I didn't add an API to unprotect since there's also not one to unquarantine? It is easy enough to add if we want though.

PR #10040, opened 2021-05-21
Add an admin API for unprotecting local media from quarantine

dklimpel:
Related: #9086

### Pull Request Checklist

### DIFF
--- a/changelog.d/10040.feature
+++ b/changelog.d/10040.feature
@@ -1,0 +1,1 @@
+Add an admin API for unprotecting local media from quarantine.
--- a/changelog.d/10040.feature
+++ b/changelog.d/10040.feature
@@ -1,1 +1,1 @@
-Add an admin API for unprotecting local media from quarantine.
+Add an admin API for unprotecting local media from quarantine. Contributed by @dklimpel.
--- a/docs/admin_api/media_admin_api.md
+++ b/docs/admin_api/media_admin_api.md
@@ -7,6 +7,7 @@
   * [Quarantining media in a room](#quarantining-media-in-a-room)
   * [Quarantining all media of a user](#quarantining-all-media-of-a-user)
   * [Protecting media from being quarantined](#protecting-media-from-being-quarantined)
+  * [Unprotecting media from being quarantined](#unprotecting-media-from-being-quarantined)
 - [Delete local media](#delete-local-media)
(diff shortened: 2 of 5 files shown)


docs/admin_api/media_admin_api.md
```diff
@@ -159,6 +160,26 @@ Response:
...
 ```
 
+## Unprotecting media from being quarantined
+
+This API reverts the protection of a media.
+
+Request:
+
+```
+DELETE /_synapse/admin/v1/media/protect/<media_id>
```
richvdh, line 170: does `DELETE` feel scary to anyone else? I feel like I'd be worrying that I'm deleting the media itself.

I'd be inclined towards `POST /_synapse/admin/v1/media/unprotect/<media_id>`, but I don't have a very strong opinion.

richvdh (commented):
lgtm otherwise

docs/admin_api/media_admin_api.md
dklimpel: I do not like so many URI. When you something create with `PUT` or `PULL` you can revert it with `DELETE`. This ist HTTP.
We can add a sentence, that this do not delete a media. It only deletes the protection.
Do you have team call for discussing this? IMO there ist missin a kind of compliance for all oh the admin API.
Probably is need for an extensive API like for user.

clokep: I agree that `DELETE` seems like the wrong semantic since you're updating the media, not deleting it.

richvdh: @dklimpel so I did a straw poll among the team in #synapse-dev and the general feeling was that POST would be more intuitive than DELETE.

Although `DELETE` is in some ways the opposite of `POST`, I don't think that's always the case, and it's certainly not the case that there are any standards around that.

dklimpel: Ok. Now it is a `POST` request.

changelog.d/10040.feature
```diff
@@ -0,0 +1 @@
+Add an admin API for unprotecting local media from quarantine.
```
richvdh, line 1:
```suggestion
Add an admin API for unprotecting local media from quarantine. Contributed by @dklimpel.
```

richvdh (approved):
thank you!

PR #10045, opened 2021-05-22
Fix docker image to not log at `/homeserver.log`

### LINKED ISSUE #9970: Default generated logger config causes "Permission denied: '/homeserver.log'" for non-root containers
### Description

Using `generate` to create config files with the Docker image sets the value of `handlers.file.filename` inside the `*.log.config` file to `/homeserver.log`, causing the following error when running the container as non-root:

```
PermissionError: [Errno 13] Permission denied: '/homeserver.log'
```

Changing it to `/data/homeserver.log` fixes the issue since the `synapse` user has write access to that folder. 

Unfortunately it's not enough to disable the `file` log `handler`. Even when disabled an empty log file is initially created. 

### Steps to reproduce

- Create minimal config files with `generate`
- Run container in non-root mode

### Version information

If not matrix.org:

- **Version**: v1.33.1

- **Install method**: Docker Image

- **Platform**: Kubernetes

### Cause

The issue lies within how the log config is generated (https://github.com/matrix-org/synapse/blob/develop/synapse/config/logger.py#L171-L180):

```python
    def generate_files(self, config, config_dir_path):
        log_config = config.get("log_config")
        if log_config and not os.path.exists(log_config):
            log_file = self.abspath("homeserver.log")
            print(
                "Generating log config file %s which will log to %s"
                % (log_config, log_file)
            )
            with open(log_config, "w") as log_config_file:
                log_config_file.write(DEFAULT_LOG_CONFIG.substitute(log_file=log_file))
```

which calls:

```
@staticmethod
    def abspath(file_path):
        return os.path.abspath(file_path) if file_path else file_path
```
Since the container runs `start.py` in `/`, the above resolves to `/homeserver.log`

lonyelon:
Fixes #9970 

Signed-off-by: Sergio Miguéns Iglesias <lonyelon@lony.xyz>

### DIFF
--- a/changelog.d/10045.docker
+++ b/changelog.d/10045.docker
@@ -1,0 +1,1 @@
+Fixed server not having access to `homeserver.log` by moving the default path to `/data/homeserver.log`. Contributed by Sergio Miguéns Iglesias.
--- a/changelog.d/10045.docker
+++ b/changelog.d/10045.docker
@@ -1,1 +1,1 @@
-Fixed server not having access to `homeserver.log` by moving the default path to `/data/homeserver.log`. Contributed by Sergio Miguéns Iglesias.
+Fix bug introduced in Synapse 1.33.0 which caused a `Permission denied: '/homeserver.log'` error when starting Synapse with the generated log configuration. Contributed by Sergio Miguéns Iglesias.
--- a/docker/conf/log.config
+++ b/docker/conf/log.config
@@ -9,10 +9,11 @@ formatters:
 {% endif %}
 
 handlers:
+{% if LOG_FILE_PATH %}
   file:
     class: logging.handlers.TimedRotatingFileHandler
     formatter: precise
-    filename: {{ LOG_FILE_PATH or "/data/homeserver.log" }}
+    filename: {{ LOG_FILE_PATH }}
     when: "midnight"
     backupCount: 6  # Does not include the current log file.
     encoding: utf8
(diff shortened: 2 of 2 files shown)


docker/conf/log.config
```diff
@@ -12,7 +12,7 @@ handlers:
   file:
     class: logging.handlers.TimedRotatingFileHandler
     formatter: precise
-    filename: {{ LOG_FILE_PATH or "homeserver.log" }}
+    filename: {{ LOG_FILE_PATH or "/data/homeserver.log" }}
```
richvdh, line 15: the trouble with this is that it will still create an empty file in `/data`.

I think it would be better to surround the whole `file` and `buffer` configuration sections with `{% if LOG_FILE_PATH %}` (which means that the generated config will end up being similar to what we had before #9162).

richvdh (changes requested):
thanks for the contribution! I have a suggestion though.

lonyelon:
> thanks for the contribution! I have a suggestion though.

You mean like this?
And thanks for reviewing my small change : )

changelog.d/10045.docker
```diff
@@ -0,0 +1 @@
+Fixed server not having access to `homeserver.log` by moving the default path to `/data/homeserver.log`. Contributed by Sergio Miguéns Iglesias.
```
richvdh, line 1: this isn't really correct any more.

```suggestion
Fix bug introduced in Synapse 1.33.0 which caused a `Permission denied: '/homeserver.log'` error when starting Synapse with the generated log configuration. Contributed by Sergio Miguéns Iglesias.
```

richvdh (approved):
perfect, thank you!

PR #10046, opened 2021-05-23
Update CAPTCHA documentation to mention turning off verify origin feature

aaronraimist:
### Pull Request Checklist

### DIFF
--- a/changelog.d/10046.doc
+++ b/changelog.d/10046.doc
@@ -1,1 +1,1 @@
-Update CAPTCHA documentation to mention turning off the verify origin feature.
+Update CAPTCHA documentation to mention turning off the verify origin feature. Contributed by @aaronraimist.
--- a/changelog.d/10046.doc
+++ b/changelog.d/10046.doc
@@ -1,0 +1,1 @@
+Update CAPTCHA documentation to mention turning off the verify origin feature.
--- a/docs/CAPTCHA_SETUP.md
+++ b/docs/CAPTCHA_SETUP.md
@@ -10,7 +10,7 @@ from Google.
 1. Set the type to reCAPTCHA v2 using the "I'm not a robot" Checkbox option.
 This is the only type of captcha that works with Synapse.
 1. Add your server name to the list of authorized domains.
-1. Agree to the terms of service and submit
+1. Agree to the terms of service and submit.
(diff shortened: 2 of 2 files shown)


docs/CAPTCHA_SETUP.md
```diff
@@ -1,31 +1,35 @@
...
+accounts. Synapse currently uses Google's reCAPTCHA service which requires API keys
+from Google.
+
+## Getting API keys
+
+1. Create a new site at <https://www.google.com/recaptcha/admin/create>
+1. Set the label to anything you want
+1. Set the type to reCAPTCHA v2 using the "I'm not a robot" Checkbox option.
+This is the only type of captcha that works with Synapse.
+1. Add your server name to the list of authorized domains.
```
richvdh, line 12: is this server name as in `server_name` as in the thing that ends up in MXIDs, or the public hostname for the C-S API?

I've a feeling it's the latter, in which case it might be better to use words like "the public hostname for your server, as set in `public_baseurl` in `homeserver.yaml`".

docs/CAPTCHA_SETUP.md
```diff
@@ -1,31 +1,35 @@
...
+1. Agree to the terms of service and submit
```
richvdh, line 13:
```suggestion
1. Agree to the terms of service and submit.
```

changelog.d/10046.doc
```diff
@@ -0,0 +1 @@
+Update CAPTCHA documentation to mention turning off the verify origin feature.
```
richvdh, line 1:
```suggestion
Update CAPTCHA documentation to mention turning off the verify origin feature. Contributed by @aaronraimist.
```

richvdh (changes requested):
thanks very much for this! A couple of nits.

docs/CAPTCHA_SETUP.md
aaronraimist: I was just trying to have them put *some* domain in there because Google requires at least one domain to be added to the list. As long as they do step 8 then it doesn't matter what domains are in the list.

But yes that is probably more correct because someone may not do step 8 and may not have the `public_baseurl` as a subdomain of the server name domain.

richvdh (approved):
thank you!

PR #9247, opened 2021-01-28
Add endpoints for backfilling history (MSC2716)

MadLittleMods:
Implement [MSC2716](https://github.com/matrix-org/matrix-doc/pull/2716) to add endpoints for backfilling history. This PR does not support federation use cases with the "marker" and "insertion" events.

For reviewers, it's probably best to see this in action with the [associated Complement tests](https://github.com/matrix-org/complement/pull/68).

---

Complement MR: https://github.com/matrix-org/complement/pull/68

TARDIS visualization MR: https://github.com/matrix-org/tardis/pull/1

### Getting started

The PR adds the `POST /_matrix/client/unstable/org.matrix.msc2716/rooms/<roomID>/batch_send?prev_event=<eventID>&chunk_id=<chunkID>` endpoint which can insert a chunk of events historically back in time next to the given `prev_event`. `chunk_id` comes from `next_chunk_id` in the response of the batch send endpoint and is derived from the "insertion" events added to each chunk. It's not required for the first batch send.
```
{
    "events": [ ... ],
    "state_events_at_start": [ ... ]
}
```

The `/batchsend` endpoint is behind a feature flag: `experimental_features` -> `msc2716_enabled` (defined in `homseserver.yaml`). And is only available to application services so you will need to add one to your `homeserver.yml` and use the `as_token` defined to interact with the API (other tokens will 403).

`state_events_at_start` is used to define the historical state events needed to auth the `events` like join events. These events will float outside of the normal DAG as `outlier`'s and won't be visible in the chat history which also allows us to insert multiple chunks without having a bunch of `@mxid joined the room` noise between each chunk.

`events` is chronological chunk/list of events you want to insert. There is a reverse-chronological constraint on chunks so once you insert some messages, you can only insert older ones after that. tldr; Insert from your most recent history -> oldest history.

Why? `depth` is not re-calculated when historical messages are inserted into the DAG. This means we have to take care to insert in the right order. Events are sorted by `(topological_ordering, stream_ordering)` where `topological_ordering` is just `depth`. Normally, `stream_ordering` is an auto incrementing integer but for `backfilled=true` events, it decrements. Historical messages are inserted all at the same depth, and marked as backfilled so the `stream_ordering` decrements and each event is sorted behind the next. (from https://github.com/matrix-org/synapse/pull/9247#discussion_r588479201)

If you're curious to look at a known working example, the Complement tests have barebones test cases interacting with this API, https://github.com/matrix-org/complement/pull/68

#### Steps to reproduce:

[...]

PR #10048, opened 2021-05-24
Add `parse_strings_from_args` to get `prev_events` array

MadLittleMods:
Add `parse_strings_from_args(...)` to parse array of `?prev_event` query parameters from `POST /_matrix/client/r0/rooms/<roomID>/bulksend?prev_event=<eventID>&prev_event=<eventID>&chunk_id=<chunkID>`  (can be up to 20 `prev_event`)

Split out from https://github.com/matrix-org/synapse/pull/9247 for [MSC2716](https://github.com/matrix-org/matrix-doc/pull/2716).

Strings:

 - `parse_string`
 - `parse_string_from_args`
 - **`parse_strings_from_args`** (new in this PR)

For comparison with ints:

 - `parse_integer`
 - `parse_integer_from_args`

Previous discussions:

 - https://github.com/matrix-org/synapse/pull/9247#discussion_r573195687
 - https://github.com/matrix-org/synapse/pull/9247#discussion_r574214156
 - https://github.com/matrix-org/synapse/pull/9247#discussion_r573264791

### Pull Request Checklist


synapse/http/servlet.py
```diff
@@ -143,45 +143,81 @@ def parse_string(
             is not one of those allowed values.
     """
     return parse_string_from_args(
-        request.args, name, default, required, allowed_values, param_type, encoding
```
MadLittleMods, line 146: > Removed `param_type` from all of the string parsing functions. It's not used in the codebase and seems superfluous because it's always `string` anyway and just used for the error message.
>
> https://github.com/matrix-org/synapse/pull/9247/files#r573264791

synapse/http/servlet.py
```diff
@@ -131,7 +131,6 @@ def parse_string(
             string, or None if any value is allowed, defaults to None. Must be
             the same type as name, if given.
         encoding (str|None): The encoding to decode the string content with.
```
erikjohnston, line 133: Can you remove the types from the docstring please? (It stops them from getting out of sync)

synapse/http/servlet.py
```diff
@@ -142,45 +141,81 @@ def parse_string(
...
-    allowed_values=None,
-    param_type="string",
-    encoding="ascii",
-):
+def _parse_string_value(
+    value: Union[str, bytes],
+    allowed_values: Optional[Iterable[str]],
+    name: str,
+    encoding: Optional[str],
+) -> str:
```
erikjohnston, line 153:
```suggestion
) -> Union[str, bytes]:
```

synapse/http/servlet.py
```diff
@@ -142,45 +141,81 @@ def parse_string(
...
+
+
+def parse_strings_from_args(
+    args: List[str],
+    name: Union[bytes, str],
+    default: Optional[str] = None,
+    required: bool = False,
+    allowed_values: Optional[Iterable[str]] = None,
+    encoding: Optional[str] = "ascii",
+) -> Optional[List[Union[bytes, str]]]:
```
erikjohnston, line 177: If you're feeling mad keen then you can do a couple of `@overload` to say that this returns `bytes` if `encoding` is `Literal[None]` else `str` (c.f. https://mypy.readthedocs.io/en/stable/literal_types.html).

I don't think we ever return `None` here?

synapse/http/servlet.py
```diff
@@ -142,45 +141,81 @@ def parse_string(
...
+        )
+        raise SynapseError(400, message)
+    else:
+        return value
+
+
+def parse_strings_from_args(
+    args: List[str],
+    name: Union[bytes, str],
+    default: Optional[str] = None,
```
erikjohnston, line 173: We should probably say this is a `Optional[List[str]]` so that we have a consistent return type?

erikjohnston (commented):
Thanks for splitting out! While we're here I think we should fix up the types a bit.

synapse/http/servlet.py
MadLittleMods: It could be `None` if the default is returned.

MadLittleMods: Added a couple `@overload` which I think are correct. I still have `Optional` for the `default` return case

synapse/http/servlet.py
```diff
@@ -15,6 +15,7 @@
 """ This module contains base REST classes for constructing REST servlets. """
 
 import logging
+from typing import Iterable, List, Literal, Optional, Union, overload
```
erikjohnston, line 18: Ah, the test failures look to be due to old pythons not having `Literal` in `typing`.
```suggestion
from typing import Iterable, List, Optional, Union, overload
from typing_extensions import Literal
```

synapse/http/servlet.py
```diff
@@ -190,6 +249,59 @@ def parse_string_from_args(
...
+        args,
+        name,
+        default=default,
+        required=required,
+        allowed_values=allowed_values,
+        encoding=encoding,
+    )
+
+    if isinstance(strings, list):
+        return strings[0]
```
erikjohnston, line 299: I think it's probably clearer to pass in `default=[default]` and then always do `strings[0]`?

erikjohnston (commented):
I think this LGTM once we fix the tests

MadLittleMods:
I feel like the remaining failing Sytest tests are just flakey. They are different when I re-run them.

Trying to get it running locally to confirm, https://github.com/matrix-org/sytest/issues/1054

erikjohnston (approved):

MadLittleMods:
@erikjohnston Thanks for all of the review and merge 🐺

PR #10049, opened 2021-05-24
Add `depth` parameter and pipe through the event creation methods

MadLittleMods:
Add `depth` parameter and pipe through the event creation methods. Also add `auth_event_ids` and `prev_event_ids` in some places as well so I don't have to go through PR splitting conflict hell.

Split out from https://github.com/matrix-org/synapse/pull/9247

Call stacks:
```
create_event
create_new_client_event
builder.build
```

```
update_membership
update_membership_locked
_local_membership_update
create_event
```

### Pull Request Checklist


synapse/handlers/room_member.py
```diff
@@ -496,160 +561,170 @@ async def update_membership_locked(
             if block_invite:
                 raise SynapseError(403, "Invites have been disabled on this server")
 
-        latest_event_ids = await self.store.get_prev_events_for_room(room_id)
+        if prev_event_ids:
+            latest_event_ids = prev_event_ids
```
MadLittleMods, line 565: > New case is here where we override the prev_event_ids if passed in explicitly
>
> https://github.com/matrix-org/synapse/pull/9247/files#r610994535

synapse/handlers/room_member.py
```diff
@@ -496,160 +561,170 @@ async def update_membership_locked(
...
+        else:
+            latest_event_ids = await self.store.get_prev_events_for_room(room_id)
```
MadLittleMods, line 567: > All of this code is just nested under the `else` conditional and not changed.
>
> https://github.com/matrix-org/synapse/pull/9247/files#r610994518

synapse/storage/databases/main/event_federation.py
```diff
@@ -670,24 +671,61 @@ def get_oldest_events_with_depth_in_room_txn(self, txn, room_id):
 
         return dict(txn)
 
-    async def get_max_depth_of(self, event_ids: List[str]) -> int:
-        """Returns the max depth of a set of event IDs
-
+    async def get_max_depth_of(self, event_ids: List[str]) -> Tuple[str, int]:
```
MadLittleMods, line 674: A bit of new logic here. We also return the event ID for the event with the greatest depth

synapse/events/builder.py
```diff
@@ -131,8 +134,46 @@ async def build(
...
+            # to insert into.
+            if not successor_event_ids:
+                depth = most_recent_prev_event_depth
+            else:
+                (
+                    _,
+                    oldest_successor_depth,
+                ) = await self._store.get_min_depth_of(successor_event_ids)
+
+                depth = oldest_successor_depth
```
MadLittleMods, line 173: The smarts of `inherit_depth` are here 🎈

synapse/handlers/room_member.py
```diff
@@ -496,160 +561,170 @@ async def update_membership_locked(
...
-                if len(latest_event_ids) == 0:
-                    latest_event_ids = [invite.event_id]
+                    # thanks to overzealous cleaning up of event_forward_extremities in
+                    # `delete_old_current_state_events`, it's possible to end up with no
+                    # forward extremities here. If that happens, let's just hang the
+                    # rejection off the invite event.
+                    #
+                    # see: https://github.com/matrix-org/synapse/issues/7139
+                    if len(latest_event_ids) == 0:
+                        latest_event_ids = [invite.event_id]
```
erikjohnston, line 727: I wonder whether instead of indenting all this we should return early?

erikjohnston (commented):
These look like changes that would work :)

synapse/handlers/room_member.py
MadLittleMods: I've changed it to return early. It's pretty hard to spot the return in the middle of the function.

I'm thinking about moving all of this nested logic to its own function so it's easier to read here.

MadLittleMods: Looking closer, that bunch of nested logic has its own returns. Can leave it for now ⏩

MadLittleMods:
I feel like the remaining failing Sytest tests are just flakey. They are different when I re-run them.

Trying to get it running locally to confirm, https://github.com/matrix-org/sytest/issues/1054

synapse/handlers/message.py
```diff
@@ -776,8 +800,16 @@ async def create_and_send_nonmember_event(
                     assert event.internal_metadata.stream_ordering
                     return event, event.internal_metadata.stream_ordering
 
+            prev_events = event_dict.get("prev_events")
```
erikjohnston, line 803: Can we pass these in instead? I'm wary of accidentally letting clients set it.

erikjohnston (commented):
Thanks, this looks good other than tiny nit

synapse/handlers/message.py
MadLittleMods: Good idea :+1:

MadLittleMods:
@erikjohnston Thanks for all the review so far 🎢

erikjohnston:
I think this is good to merge, but I'm going to hold off a bit until we #9247 is a bit closer (in particular I'm wondering if we should pass in the depth explicitly, rather than have `inherit_depth`). Thanks for splitting it out from that PR though, it was helpful to be able to review it separately :)

MadLittleMods:
@erikjohnston I've updated this PR and the [base PR](https://github.com/matrix-org/synapse/pull/9247) to pass around `depth` directly :+1:

synapse/events/builder.py
```diff
@@ -131,8 +135,14 @@ async def build(
             auth_events = auth_event_ids
             prev_events = prev_event_ids
 
-        old_depth = await self._store.get_max_depth_of(prev_event_ids)
-        depth = old_depth + 1
+        (
+            _,
+            most_recent_prev_event_depth,
+        ) = await self._store.get_max_depth_of(prev_event_ids)
+
```
erikjohnston, line 142: May as well move this into the `if depth is None`?

synapse/storage/databases/main/event_federation.py
erikjohnston: Presumably this new logic will get used when we add the bulk insert API? (It's not used in this PR anyomre)

erikjohnston (approved):
This looks good, thanks, but again let's wait for the other PR before landing this

synapse/storage/databases/main/event_federation.py
MadLittleMods: It is still used in https://github.com/matrix-org/synapse/pull/9247. I can remove here if preferred.

erikjohnston: I think its ok to leave here.

tests/handlers/test_presence.py
```diff
@@ -863,7 +863,7 @@ def _add_new_user(self, room_id, user_id):
             self.store.get_latest_event_ids_in_room(room_id)
         )
 
-        event = self.get_success(builder.build(prev_event_ids, None))
+        event = self.get_success(builder.build(prev_event_ids, None, False))
```
MadLittleMods, line 866:
```suggestion
        event = self.get_success(builder.build(prev_event_ids, None, None))
```

PR #10062, opened 2021-05-25
Standardise the module interface

babolivier:
This PR adds a common configuration section for all modules (see [WIP docs](https://github.com/matrix-org/synapse/blob/babolivier/modules/docs/modules.md)). These modules are then loaded at startup by the homeserver.

Modules register their hooks and web resources using the new `register_[...]_callbacks` and `register_web_resource` methods of the module API. To give an example on how this fits together end to end, I've also upgraded the `SpamChecker` class to support these changes.

The next step once this PR is merged is to upgrade the other module APIs these changes.

## Example

Here's an example of a module that registers a web resource and implements the `user_may_create_room` method from the spam checker API:

```python
import json

from twisted.web.resource import Resource
from twisted.web.server import Request

from synapse.module_api import ModuleApi


class DemoResource(Resource):
    def __init__(self, config):
        super(DemoResource, self).__init__()
        self.config = config

    def render_GET(self, request: Request):
        name = request.args.get(b"name")[0]
        request.setHeader(b"Content-Type", b"application/json")
        return json.dumps({"hello": name})


class DemoModule:
    def __init__(self, config: dict, api: ModuleApi):
        self.config = config
        self.api = api

        self.api.register_web_resource(
            path="/_synapse/client/demo/hello",
            resource=DemoResource(self.config),
        )

        self.api.register_spam_checker_callbacks(
            user_may_create_room=self.user_may_create_room,
        )

    @staticmethod
    def parse_config(config):
        return config

    async def user_may_create_room(self, user: str) -> bool:
        if user == "@evilguy:example.com":
            return False

        return True
```


synapse/config/modules.py
```diff
@@ -0,0 +1,51 @@
...
+            #
+            # See https://github.com/matrix-org/synapse/tree/master/docs/modules.md for
+            # more documentation on how to configure or create custom modules for Synapse.
+            #
+            modules:
+                # - module: my_super_module.MySuperClass
+                #   config:
+                #       do_thing: true
+                # - module: my_other_super_module.SomeClass
+                #   config: {}
```
richvdh, line 49: I've never really liked this `module`, `config` structure. I wonder if we should take the opportunity to reshuffle it. For example:

```yaml
modules:
    my_super_module.MySuperClass:
        do_thing: true
    my_other_super_module.SomeClass: {}
```

I'm not sure I love that either, though. Alternative suggestions welcome.

richvdh: (there must be other systems that have configurable modules that we can draw inspiration from?)

synapse/module_api/__init__.py
```diff
@@ -73,6 +87,62 @@ def public_room_list_manager(self):
...
+            fn_name: The hook name.
+            callback: The function to call for this hook.
+        """
+        if fn_name not in self._callbacks.keys():
+            self._callbacks[fn_name] = [callback]
+            return
+
+        self._callbacks[fn_name].append(callback)
+
+    def get_registered_hook(self, fn_name: str) -> List[Callable]:
```
babolivier, line 108: Currently these look like they could be used by modules, even though the intent is very much that they're called from within Synapse, I'm not sure how to make this clear enough (maybe changing them so they're private methods? feels a bit hacky since we'll need to call them from outside the `ModuleApi`).

synapse/module_api/__init__.py
```diff
@@ -73,6 +87,62 @@ def public_room_list_manager(self):
         """
         return self._public_room_list_manager
 
+    def register_hook(self, fn_name: str, callback: Callable):
+        """Registers the given callback as a hook with the given name.
+
+        This function must be called in the __init__ method of the modules using it.
```
babolivier, line 93: It feels like we might want to error if this function isn't called in the `__init__` method of a module, but I'm not sure how to check that nor whether it's worth the pain.

richvdh: I don't think it's necessary tbh. Why must it be called in the __init__ method specifically? It seems the obvious way, but ultimately it's up for the module to ensure that its hooks are registered during startup. If it wants to do it some other way, that's ok by us.

synapse/module_api/__init__.py
richvdh: well the obvious answer is to move `_callbacks` and `get_registered_hook` somewhere else, and have `register_hook` proxy to that place.

However, I'm dubious about having a single function for all hooks and callbacks. Better to have individual functions for each type of hook - for example you might have `ModuleApi.register_spamcheck_callback`, which proxies to a method on `hs.get_spam_checker()`, and the `SpamChecker` is responsible for maintaining its own list of callbacks.

A big advantage of doing so is that you can make the definition more typesafe - ie, declare it to take a `Callable[[Event], bool]` rather than just a `Callable`.

babolivier: > A big advantage of doing so is that you can make the definition more typesafe - ie, declare it to take a Callable[[Event], bool] rather than just a Callable.

Not necessarily because all e.g. spamchecker callbacks don't have the same signature, unless you're proposing we create a `register_` method for each hook which doesn't sound like the right solution.

Overall it does sound like a better solution though, thanks.

synapse/events/spamcheck.py
```diff
@@ -30,19 +40,116 @@
...
+            self._register_callback(
+                "check_registration_for_spam",
+                check_registration_for_spam,
+            )
+
+        if check_media_file_for_spam is not None:
+            self._register_callback(
+                "check_media_file_for_spam",
+                check_media_file_for_spam,
+            )
```
babolivier, line 140: This function feels a bit meh to me because it looks like a big function with a lot of copy/paste but it was the only way I could think of to make it work with type hints on the callbacks signatures (and the method itself is relatively simple so *shrug*)

synapse/module_api/__init__.py
babolivier: Implemented in https://github.com/matrix-org/synapse/pull/10062/commits/f5098c9113e5d20ffdc8ef0abe3a09f33795773c

synapse/events/spamcheck.py
```diff
@@ -30,19 +40,120 @@
...
-
-        for module, config in hs.config.spam_checkers:
-            # Older spam checkers don't accept the `api` argument, so we
-            # try and detect support.
-            spam_args = inspect.getfullargspec(module)
-            if "api" in spam_args.args:
-                self.spam_checkers.append(module(config=config, api=api))
-            else:
-                self.spam_checkers.append(module(config=config))
+        self._callbacks: Dict[str, List[Callable]] = {}
```
richvdh, line 85: I would honestly go with a separate list for each callback. I realise it'll make this longer, but I don't think it'll be too bad, and it'll mean you can replace `self._callbacks.get("user_may_create_room", [])` with `self._user_may_create_room_callbacks`.

synapse/events/spamcheck.py
```diff
@@ -30,19 +40,120 @@
...
+        check_username_for_spam: Optional[Callable[[Dict[str, str]], bool]] = None,
+        check_registration_for_spam: Optional[
+            Callable[
+                [
+                    Optional[dict],
+                    Optional[str],
+                    Collection[Tuple[str, str]],
+                    Optional[str],
+                ],
+                bool,
```
richvdh, line 105: consider defining a type for this:

```
CHECK_REGISTRATION_FOR_SPAM_CALLBACK = Callable[ ... etc ... ]
```

synapse/events/spamcheck.py
```diff
@@ -211,33 +301,28 @@ async def check_registration_for_spam(
...
+                    request_info,
+                    auth_provider_id,
+                )
+            elif len(checker_args.parameters) == 3:
+                d = checker(email_threepid, username, request_info)
+            else:
+                logger.error(
+                    "Invalid signature for check_registration_for_spam. Denying registration",
+                )
+                return RegistrationBehaviour.DENY
```
richvdh, line 320: you could sort this out in `register_callback`: if the callback only takes three params, wrap it with a function which takes 4.

synapse/module_api/__init__.py
```diff
@@ -73,6 +87,62 @@ def public_room_list_manager(self):
...
+        Args:
+            fn_name: The name of the hook.
+
+        Returns:
+            The callbacks implementing the hook, or an empty list if there is no callback
+            for this hook.
+        """
+        return self._callbacks.get(fn_name, [])
+
+    def register_web_resource(self, path: str, resource: IResource):
```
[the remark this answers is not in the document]
richvdh: ideally, I guess it should.

synapse/events/spamcheck.py
richvdh: I think it's fine. It's a bit boilerplatey but as you say it's simple enough. Sometimes it's better just to accept the boilerplate and make everything explicit than it is to do anything too clever.

tests/rest/media/v1/test_media_storage.py
```diff
@@ -535,6 +536,8 @@ def prepare(self, reactor, clock, hs):
         self.download_resource = self.media_repo.children[b"download"]
         self.upload_resource = self.media_repo.children[b"upload"]
 
+        load_legacy_spam_checkers(hs)
```
babolivier, line 539: I've fixed this test with the legacy spam checker code so at least this path is evaluated.

synapse/config/modules.py
callahad: Django typically puts everything (core and add-on settings) into the same settings namespace, for what that's worth. For example, see the Installation and Configuration sections for the Allauth module https://django-allauth.readthedocs.io/en/latest/installation.html

clokep: I'm unsure if it matters, but it is probably worth considering whether we care about the order that modules run in, e.g. I'm unsure if the example above is ordered:

```yaml
modules:
    my_super_module.MySuperClass:
        do_thing: true
    my_other_super_module.SomeClass: {}
```

(Implementation-wise it is probably fine, but I don't know what YAML says about mappings.)

callahad: Looks like the actual PR uses an array of objects (hence the `- ` prefixes), while Rich's snippet above is just a mapping, and as you pointed out, lord only knows what ordering we'll get out of that. :)

I'd argue in favor of the ordered array approach, which sidesteps this issue and potentially allows for multiple instances of the same module which vary only in their configuration.

babolivier: Thanks all, I'll keep it as is then.

docs/modules.md
```diff
@@ -0,0 +1,229 @@
...
+
+from synapse.module_api import ModuleApi
+
+
+class DemoResource(Resource):
+    def __init__(self, config):
+        super(DemoResource, self).__init__()
+        self.config = config
+
+    async def render_GET(self, request: Request):
```
twm, line 200: `t.w.r.Resource` doesn't allow coroutines when it delegates to `render_*`. Is the `async` misplaced?

richvdh (commented):
generally, this is looking great. A few bits and pieces..

docs/modules.md
babolivier: Ah yes this is bad copy-paste from my part, will fix

synapse/server.py
```diff
@@ -259,6 +259,33 @@ def __init__(
...
+                listeners have been started.
+        """
+        if self._module_web_resources_consumed:
+            raise SynapseError(
+                500,
+                "Tried to register a web resource from a module after startup",
+            )
+
+        # Don't register a resource that's already been registered.
+        if path not in self._module_web_resources.keys():
```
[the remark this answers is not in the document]
babolivier: I think raising an exception might be a bit too much since two modules conflicting on a path shouldn't be expected to be the user's fault, but having a warning log sounds like a good idea.

docs/spam_checker.md
```diff
@@ -1,3 +1,7 @@
+**Note: this page of the Synapse documentation is now deprecated. For up to date
+documentation on setting up or writing a spam checker module, please see
+[this page](https://matrix-org.github.io/synapse/develop/modules.html).**
+
```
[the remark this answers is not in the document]
babolivier: I've realised I forgot to include a deprecation notice to the legacy spam checker interface in the upgrade notes and the changelog, which means setting a time for dropping support for it. So I suggest we pair removing this doc with removing the code to support the legacy interface.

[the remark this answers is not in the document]
richvdh: fair enough!

richvdh (changes requested):
I am super excited by this. I think it opens the path to much cleaner extensibility of synapse.

synapse/events/spamcheck.py
```diff
@@ -96,19 +102,62 @@ def load_legacy_spam_checkers(hs: "synapse.server.HomeServer"):
...
+                return None
+
+            def run(*args, **kwargs):
+                return maybe_awaitable(f(*args, **kwargs))
+
+            wrapper = run
+
+            # register_spam_checker_callbacks does additional checks on the number of
+            # arguments for check_registration_for_spam, so we can't just pass run,
+            # instead we use another wrapper with the right number of arguments.
```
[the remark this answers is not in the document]
babolivier: Oh I didn't know about that, thanks! Though I agree, it would be better to move that logic to the legacy code.

babolivier:
> I am super excited by this. I think it opens the path to much cleaner extensibility of synapse.

I 100% agree, it's going to make it so much easier to allow modules to implement new features going forward.

CHANGES.md
```diff
@@ -1,3 +1,11 @@
+Unreleased
+==========
+
+The spam checker interface is now deprecated in favour of a new generic module system and
+will be removed in Synapse 1.40 (planned for mid-August 2021). See the
```
richvdh, line 5:
```suggestion
is planned for removal in August 2021. See the
```

UPGRADE.rst
```diff
@@ -85,6 +85,23 @@ for example:
...
+
+Deprecation of the current spam checker interface
+-------------------------------------------------
+
+The current spam checker interface is deprecated in favour of a new generic modules system.
+Authors of spam checker modules can refer to `this documentation <https://matrix-org.github.io/synapse/develop/modules.html#porting-an-existing-module-that-uses-the-old-interface>`_
+to update their modules. Synapse administrators can refer to `this documentation <https://matrix-org.github.io/synapse/develop/modules.html#using-modules>`_
+to update their configuration once the modules they are using have been updated.
+
+We will remove support for the current spam checker interface in August 2021.
```
richvdh, line 99:
```suggestion
We plan to remove support for the current spam checker interface in August 2021.
```

richvdh (approved):
lgtm otherwise!

PR #10063, opened 2021-05-25
Remove the experimental spaces flag.

### LINKED ISSUE #10061: remove the `spaces_enabled` flag
there's no point having a feature flag which is on by default

clokep:
...to always enable the Spaces features (under unstable endpoints).

Fixes #10061

### DIFF
--- a/synapse/api/room_versions.py
+++ b/synapse/api/room_versions.py
@@ -183,5 +183,4 @@ KNOWN_ROOM_VERSIONS = {
         RoomVersions.MSC2176,
         RoomVersions.MSC3083,
     )
-    # Note that we do not include MSC3083 here unless it is enabled in the config.
 }  # type: Dict[str, RoomVersion]
--- a/synapse/api/room_versions.py
+++ b/synapse/api/room_versions.py
@@ -181,6 +181,7 @@ KNOWN_ROOM_VERSIONS = {
         RoomVersions.V5,
         RoomVersions.V6,
         RoomVersions.MSC2176,
+        RoomVersions.MSC3083,
     )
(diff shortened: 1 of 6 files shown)


synapse/federation/transport/server.py
```diff
@@ -1562,13 +1562,12 @@ def register_servlets(
...
 
-        if hs.config.experimental.spaces_enabled:
-            FederationSpaceSummaryServlet(
-                handler=hs.get_space_summary_handler(),
-                authenticator=authenticator,
-                ratelimiter=ratelimiter,
-                server_name=hs.hostname,
-            ).register(resource)
+        FederationSpaceSummaryServlet(
+            handler=hs.get_space_summary_handler(),
```
clokep, line 1566: Note that this gets a different `handler` than the above loop. I have a branch somewhere that changes all of these endpoints to accept a `HomeServer` (like the normal C-S API endpoints) which would probably clean this all up.

clokep:
I've again punted what we want to do with the other experimental flags since those default to off. I'm willing to include removing them here though if we'd like. 🤷

synapse/api/room_versions.py
```diff
@@ -181,6 +181,7 @@ class RoomVersions:
         RoomVersions.V5,
         RoomVersions.V6,
         RoomVersions.MSC2176,
+        RoomVersions.MSC3083,
     )
     # Note that we do not include MSC3083 here unless it is enabled in the config.
```
richvdh, line 186:
```suggestion
```

richvdh (approved):
mmkay.

synapse/api/room_versions.py
clokep: Ugh, thanks.

Issue #8551, opened 2020-10-15
Implement MSC2414: Make reason and score parameters optional for reporting content


anoadragon453:
[MSC2414](https://github.com/matrix-org/matrix-doc/pull/2414) made the `reason` and `score` parameters for the [POST /_matrix/client/r0/rooms/{roomId}/report/{eventId}](https://matrix.org/docs/spec/client_server/r0.6.1#post-matrix-client-r0-rooms-roomid-report-eventid) endpoint.

That endpoint is implemented in Synapse here:

https://github.com/matrix-org/synapse/blob/8a4a4186ded34bab1ffb4ee1cebcb476890da207/synapse/rest/client/v2_alpha/report_event.py#L31-L46

The [Event Reports Admin API](https://github.com/matrix-org/synapse/blob/develop/docs/admin_api/event_reports.rst) should also be updated to state that the `reason` and `content` fields may be both blank and `null`, and the Admin API endpoint should be updated to expect `None` values for these keys as well:

https://github.com/matrix-org/synapse/blob/cbabb312e0b59090e5a8cf9e7e016a8618e62867/synapse/storage/databases/main/room.py#L1388-L1389

(that function may not actually need updating, but it would be good to check that the API still behaves as expected).

Some tests for clients sending reports without `reason` and `score` would be great as well.

This will need to be implemented in order to eventually advertise support for the next Client-Server API version.

clokep:
Should this be optional for all "versions" of the endpoint or only if accessed via `/vNext/rooms/....`? Not sure how we usually do that?

anoadragon453:
@clokep The MSC states that it is a change to the existing endpoint (however clients should check whether CS versions the homeserver advertises before potentially omitting a parameter).

As long as we don't advertise support for `$next_cs_version` before making this change, all should be fine.

lukasdenk:
@anoadragon453 Is this issue still relevant? Or should it be closed?

PR #10077, opened 2021-05-27
Make reason and score optional for report_event

govynnus:
Implements MSC2414: https://github.com/matrix-org/matrix-doc/pull/2414
See #8551 

### DIFF
--- a/changelog.d/10077.feature
+++ b/changelog.d/10077.feature
@@ -1,0 +1,1 @@
+Make reason and score parameters optional for reporting content. Implements MSC2414. Contributed by Callum Brown.
--- a/changelog.d/10077.feature
+++ b/changelog.d/10077.feature
@@ -1,1 +1,1 @@
-Make reason and score parameters optional for reporting content. Implements MSC2414. Contributed by Callum Brown.
+Make reason and score parameters optional for reporting content. Implements [MSC2414](https://github.com/matrix-org/matrix-doc/pull/2414). Contributed by Callum Brown.
--- a/synapse/rest/client/v2_alpha/report_event.py
+++ b/synapse/rest/client/v2_alpha/report_event.py
@@ -16,11 +16,7 @@ import logging
 from http import HTTPStatus
 
 from synapse.api.errors import Codes, SynapseError
-from synapse.http.servlet import (
-    RestServlet,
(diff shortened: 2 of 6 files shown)


synapse/rest/client/v2_alpha/report_event.py
```diff
@@ -42,15 +38,14 @@ async def on_POST(self, request, room_id, event_id):
         user_id = requester.user.to_string()
 
         body = parse_json_object_from_request(request)
-        assert_params_in_dict(body, ("reason", "score"))
 
-        if not isinstance(body["reason"], str):
+        if not isinstance(body.get("reason", ""), str):
```
babolivier, line 42: Note that this will fail if it's `None` (if you want to support it you probably want to do `body.get("reason") or ""` instead). Same comment for `score` a few lines down.

changelog.d/10077.feature
```diff
@@ -0,0 +1 @@
+Make reason and score parameters optional for reporting content. Implements MSC2414. Contributed by Callum Brown.
```
babolivier, line 1: Can you link to the MSC in there? Changelog files will be compiled into a markdown document on release so feel free to make it a markdown link.

babolivier (changes requested):
Thanks! This looks good to me apart from a few quirks :)

synapse/rest/client/v2_alpha/report_event.py
govynnus: Ah ok. So this is an acceptable request?
```
{
    "reason": null,
    "score": -100
}
```

changelog.d/10077.feature
govynnus: Sure, didn't realise it was markdown.

synapse/rest/client/v2_alpha/report_event.py
babolivier: Sorry, I got things mixed up. Your solution is correct.

changelog.d/10077.feature
babolivier: To be fair we should probably spell it out better.

babolivier (approved):

anoadragon453:
CI looks to be passing; thank you very much!

PR #10078, opened 2021-05-27
Fix up `BatchingQueue`

### LINKED ISSUE #10068: Problems with `BatchingQueue`
#10017 introduced a `BatchingQueue` implementation. While reviewing #10035 I noticed a few oddities that could do with cleaning up:

* The params to `__init__` could do with docstrings, especially `name` - what is it used for?
* The prometheus metrics (`synapse_util_batching_queue_number_queued` etc) don't appear to include the `name` label in their result.
* If the key is already being processed when `_process_queue` starts, then `_process_queue` immediately returns (https://github.com/matrix-org/synapse/blob/v1.35.0rc1/synapse/util/batching_queue.py#L118-L119). However, it will pass through the `finally` block which will [remove the key from `self._processing_keys`](https://github.com/matrix-org/synapse/blob/v1.35.0rc1/synapse/util/batching_queue.py#L153). This looks incorrect to me.
* `PreserveLoggingContext` isn't free. consider moving it outside the loops?
* It would be nice to document the behaviour on exceptions.

erikjohnston:
Fixes #10068

### DIFF
--- a/synapse/util/batching_queue.py
+++ b/synapse/util/batching_queue.py
@@ -120,10 +120,10 @@ class BatchingQueue(Generic[V, R]):
         given key and call the `self._process_batch_callback` with the values.
         """
 
-        try:
-            if key in self._processing_keys:
-                return
+        if key in self._processing_keys:
+            return
 
+        try:
             self._processing_keys.add(key)
 
             while True:
--- a/synapse/util/batching_queue.py
+++ b/synapse/util/batching_queue.py
@@ -25,10 +25,11 @@ from typing import (
     TypeVar,
 )
 
+from prometheus_client import Gauge
+
 from twisted.internet import defer
 
 from synapse.logging.context import PreserveLoggingContext, make_deferred_yieldable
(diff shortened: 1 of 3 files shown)


synapse/util/batching_queue.py
```diff
@@ -38,6 +39,24 @@
...
 R = TypeVar("R")
 
+number_queued = Gauge(
+    "synapse_util_batching_queue_number_queued",
+    "The number of items waiting in the queue across all keys",
+    labelnames=("name",),
+)
+
+number_in_flight = Gauge(
+    "synapse_util_batching_queue_in_flight",
```
richvdh, line 49: To me, `in_flight` would be the ones actually being processed right now

```suggestion
    "synapse_util_batching_queue_number_pending",
```

richvdh (approved):
lgtm otherwise

synapse/util/batching_queue.py
erikjohnston: That is a better name.

PR #10079, opened 2021-05-27
Pass the origin when calculating the spaces summary over GET.

clokep:
This fixes a bug caused by the combination of #9922 and #9947 merging separately. The former added an additional parameter to `federation_space_summary` while the latter added an additional call to `federation_space_summary`. 😢 

I found this while brushing off a branch to add some type hints to the `synapse/federation/transport/server.py`, but those were too invasive to put into a release branch.

### DIFF
--- a/changelog.d/10079.bugfix
+++ b/changelog.d/10079.bugfix
@@ -1,0 +1,1 @@
+Fix a bug introduced in v1.35.0rc1 when calling the spaces summary API via a GET request.
--- a/synapse/federation/transport/server.py
+++ b/synapse/federation/transport/server.py
@@ -1398,7 +1398,7 @@ class FederationSpaceSummaryServlet(BaseFederationServlet):
                 )
 
         return 200, await self.handler.federation_space_summary(
-            room_id, suggested_only, max_rooms_per_space, exclude_rooms
+            origin, room_id, suggested_only, max_rooms_per_space, exclude_rooms
         )
 
     # TODO When switching to the stable endpoint, remove the POST handler.


clokep:
The reasoning this is not caught by tests is that Synapse currently uses the `POST` request when requesting over federation (due to backwards compatibility concerns), so this would have been picked up by the complement tests if we updated that to use the `GET` API.

clokep:
> I find it strange/worrying that the CI didn't pick up on the fact that we were calling `federation_space_summary` with the wrong number of arguments (especially seeing as all arguments are required since none of them has a default value). Or is this what #10080 is meant to fix by adding type hints to it (and therefore enabling mypy etc to look at it)?

Yes, this is pretty bad. I attempted to explain this in https://github.com/matrix-org/synapse/pull/10079#issuecomment-849556963, but I can elaborate.

1. The federation was originally added as `POST` only.
2. #9947 added support for `GET` (which is the only thing supported, according to the MSC). In this PR we backed out the changes to *call* the `GET` version since pre-existing servers would only have the `POST` version.
3. Due to the above, this is essentially unused code until we start calling the `GET` version.
4. Complement *does* have tests for this, but it doesn't call the federation APIs directly (it calls the local HS, which then calls the federation API, which currently uses `POST`.)
5. In the future we'll switch the federation client to use `GET`, which would have picked this up.

I'm not sure if there's a better way of testing this right now unfortunately. 😢 

#10080 would have found this since it lets mypy know what the types are, but still doesn't really "test" it per-say.

PR #10080, opened 2021-05-27
Add type hints to the federation server transport.

clokep:
This cleans-up some type hints in the federation server transport code, but it got a bit invasive and has a real change in it:

Each servlet now takes a `HomeServer` and pulls out their handler manually (this allows mypy to figure out the type of `self.handler`). In order to do this without duplicating tons of code a few other "base" classes are created.

After this was done there's a few clean-up commits for real bugs / incorrect type hints. Includes the fix from #10079 in it. (Since this is how I found it.)

The other classes in that module need types too, but the PR started getting too large so I can do that after.


erikjohnston (approved):
Can you add some docstrings to the various base classes please?

clokep:
@erikjohnston I had to make a few more changes once develop was merged in, can you take a quick look?

synapse/http/servlet.py
```diff
@@ -295,6 +295,30 @@ def parse_strings_from_args(
...
+
+@overload
+def parse_string_from_args(
+    args: Dict[bytes, List[bytes]],
+    name: str,
+    default: Optional[str] = None,
+    required: bool = False,
+    allowed_values: Optional[Iterable[str]] = None,
+    encoding: str = "ascii",
+) -> Optional[str]:
```
erikjohnston, line 318: Is this meant to be in here?

clokep: These changes were necessary since previously the only caller of `parse_string_from_args` that was covered by mypy was `parse_string`, but with the additional type hints being added to the `synapse/federation/transport/server.py` file it seems to have shook out that we need an `overload` for this.

erikjohnston (approved):

Issue #9956, opened 2021-05-07
Replication request exceeds max size when trying to join room


deepbluev7:
### Description

I'm getting the following error when trying to join #irc:matrix.org:

```
2021-05-07 20:16:08,373 - synapse.http.site - 107 - WARNING - sentinel - Aborting connection from IPv4Address(type='TCP', host='127.0.0.1', port=44100) because the request exceeds maximum size
2021-05-07 20:16:08,375 - synapse.http.site - 312 - INFO - sentinel - Connection from IPv4Address(type='TCP', host='127.0.0.1', port=44100) lost before request headers were read
2021-05-07 20:16:08,376 - synapse.http.client - 440 - INFO - POST-3 - Received response to POST http://127.0.0.1:9893/_synapse/replication/remote_join/%21BAXLHOFjvDKUeLafmO%3Amatrix.org/%40deepbluev7%3Aneko.dev/qjkxpFkuVk: 502
2021-05-07 20:16:08,376 - synapse.http.server - 88 - INFO - POST-3 - <SynapseRequest at 0x7f0f95586d00 method='POST' uri='/_matrix/client/r0/join/%23irc%3Amatrix.org' clientproto='HTTP/1.0' site=8086> SynapseError: 502 - Failed to talk to 
main process
2021-05-07 20:16:08,377 - synapse.access.http.8086 - 387 - INFO - POST-3 - 127.0.0.1 - 8086 - {@deepbluev7:neko.dev} Processed request: 27.573sec/-0.000sec (0.005sec, 0.001sec) (0.002sec/0.008sec/10) 64B 502 "POST /_matrix/client/r0/join/%
23irc%3Amatrix.org HTTP/1.0" "mtxclient v0.5.1" [0 dbevts]
```

I have a few workers with an event_creator and persister as well as a master and a few federation workers. This is from the event_creator log. Increasing the maximum size in https://github.com/matrix-org/synapse/blob/3ff225175462dde8376aa584e3a47c43b1f0e790/synapse/http/site.py#L106 made the join work.

### Steps to reproduce

- Setup synapse with event_creator
- Try to join #irc:matrix.org

<!--
Describe how what happens differs from what you expected.

If you can identify any relevant log snippets from _homeserver.log_, please include
those (please be careful to remove any personal or private data). Please surround them with
``` (three backticks, on a line on their own), so that they are formatted legibly.
-->

### Version information

<!-- IMPORTANT: please answer the following questions, to help us narrow down the problem -->

[...]

PR #10082, opened 2021-05-27
Limit the number of events sent over replication when persisting events.

babolivier:


synapse/handlers/federation.py
```diff
@@ -3053,13 +3054,15 @@ async def persist_events_and_notify(
...
         if instance != self._instance_name:
-            result = await self._send_events(
-                instance_name=instance,
-                store=self.store,
-                room_id=room_id,
-                event_and_contexts=event_and_contexts,
-                backfilled=backfilled,
-            )
+            # Limit the number of events sent over federation.
+            for batch in batch_iter(event_and_contexts, 50):
```
erikjohnston, line 3058:
```suggestion
            for batch in batch_iter(event_and_contexts, 1000):
```

We want to batch things up as match as possible, and I think 1000 should still be under the size limit

erikjohnston (approved):

synapse/handlers/federation.py
clokep: Yeah I had no idea what to make this. I chose 50 as that's what comes in a transaction... wasn't sure what was reasonable though!

richvdh:
this was an attempt to fix #9956, but it was incorrect. #10118 is a further attempt.

PR #10084, opened 2021-05-27
Move some more endpoints off master

erikjohnston:
Tests at https://github.com/matrix-org/sytest/pull/1053

### DIFF
--- a/changelog.d/10084.feature
+++ b/changelog.d/10084.feature
@@ -1,0 +1,1 @@
+Add support for routing more requests to workers.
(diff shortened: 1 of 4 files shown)


clokep (approved):
Should be OK (if CI passes) since these are all read-only. I suspect the `SearchWorkerStore` isn't super useful and could be moved into `SearchBackgroundUpdateStore`, but doesn't really matter.

changelog.d/10084.feature
```diff
@@ -0,0 +1 @@
+Add support for routing more requests to workers.
```
clokep, line 1: I do wonder if we should do a better job of helping people upgrade their reverse proxies by giving more info here. 🤷

erikjohnston: Probably :shrug:

PR #10085, opened 2021-05-27
Avoid unnecessary work in the spaces summary.

clokep:
This has a few different changes in separate commits:

* Only search for `m.child.space` events in spaces, not all rooms.
* Do not track event pairs -- I think this will break some behavior with Dendrite, but Dendrite doesn't return the proper result at the moment anyway, so I'm not too concerned about this.
* Do not include `events` which point to inaccessible rooms (or are otherwise not part of the graph).

The first and the third bullets are driven by doing work that the clients drop on the floor (they do not show a normal room as a space and do not care about events with not matching room). The [MSC was updated](https://github.com/matrix-org/matrix-doc/pull/2946/commits/797dda44c4158e83dcff017f1ef2f36d7c2c4f3c) with these changes.

My hope is that this will help servers which have a slow database (or are starved for CPU, maybe). There's been some reports of the endpoint being very slow.

There is a corresponding complement PR: matrix-org/complement#123.

This depends on #10160 being merged first.

This is also some prep work for handling pagination in a roundabout way.


clokep:
This is going to fail complement tests since that depends on the old behavior, I'm not really sure how to get complement and synapse on the same page in this instance. 😢

synapse/api/constants.py
```diff
@@ -183,6 +183,11 @@ class EventContentFields:
     MSC1772_ROOM_TYPE = "org.matrix.msc1772.type"
 
 
+class RoomTypes:
+    SPACE = "m.space"
+    MSC1772_SPACE = "org.matrix.msc1772.space"
```
clokep, line 188: Note that #10161 removes support for the unstable identifiers, so depending on when these land, will probably want to remove this first.

synapse/handlers/space_summary.py
```diff
@@ -181,34 +190,42 @@ async def get_space_summary(
...
+                try:
+                    if (
+                        processed_rooms[parent_room_id]
+                        and processed_rooms[child_room_id]
+                    ):
+                        events_result.append(ev)
+                except KeyError:
+                    # Note that events are pending of the child event since if
+                    # the parent event was not accessible it shouldn't be included
+                    # at all.
```
richvdh, line 223: I don't really follow this logic. Why do we only add the event to the pending list for the child room?

richvdh (changes requested):

synapse/handlers/space_summary.py
```diff
@@ -123,21 +126,27 @@ async def get_space_summary(
 
                 if room:
                     rooms_result.append(room)
+
+                # Mark whether this room was accessible or not.
+                processed_rooms[room_id] = bool(room)
```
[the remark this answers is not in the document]
clokep: Ugh, the return type got updated in a previous PR and I failed to update the comment. Thanks!

synapse/handlers/space_summary.py
clokep: The thought process was that we only want to send the event in the case of both the parent and child room being accessible to the requester.

I think the assumption here is that if the parent room isn't in `processed_rooms` something terrible has happened? There are two sources of data we're iterating over:

* `events` - `m.space.child` events from the current room.
	* For local rooms, the parent will be the current room, the child will be a room that may or may not have been processed.
	* For remote rooms, this includes the above, as well as additional events that might be children of children, etc. (Or even events that aren't part of the graph.)
* `room_pending_events` - `m.space.child` events which were previously found to be pointing to the current room, i.e. the child room is the current room and the parent room is a previously processed room.

I'm now having trouble convincing myself that parents must be processed before children and I think that assumption might be bogus actually.

clokep:
I'm going to merge #10208 into this and build out some more tests to ensure this is working as I expect.

clokep:
I'm going to abandon this and re-do this work.

PR #10086, opened 2021-05-27
Compile and render Synapse's docs into a browsable, mobile-friendly and searchable website

anoadragon453:
**View the demo!** [https://anoadragon453.github.io/synapse/](https://anoadragon453.github.io/synapse/latest/index.html)

Synapse's documentation has traditionally existed as a collection of notes inside our `docs/` directory. While this has the benefit of being close to the code (such that it can be updated alongside code changes), it suffers from a lack of discoverability, searchability and presentation in general.

This PR attempts to organise our existing documentation into a searchable and user-friendly website for the benefit of both users of Synapse (as they can quickly browse and search for questions about running the homeserver), as well as Synapse developers in terms of quickly seeing what components of the system are documented, and which aren't.

For this project I chose [mdbook](https://github.com/rust-lang/mdBook/). Partly because we're [already using it](https://matrix-org.github.io/matrix-appservice-irc/latest/index.html) for other Matrix-related projects, but also partly because of the extensibility of the UI and rendering, quick build times and great UX.

I've done by best to organise the changes into logical commits, with useful commit messages. As a basic overview of how mdbook works however:

* `book.toml` is a new file at the root of the repository that configures how mdbook behaves, the features of the site and where to look for the content.
* `docs/SUMMARY.md` defines the outline of the site (and is what is rendered into the left sidebar). When you add a new page, it needs to be linked to from this file.
* `docs/website_files` contains custom HTML, CSS and JS that's injected into the resulting static site files at build time. These files are defined in `book.toml`, and we utilise this extensibility to add a floating table of contents plugin, and to tweak some of the default UI.

This PR strictly keeps the old documentation where it was so that historical links don't break. That being said, ideally new documentation pages would land in a sensible location rather than the root of the `docs/` directory. I've explained a little bit about the new structure (and what to do when adding a new page) in [docs/README.md](https://github.com/matrix-org/synapse/blob/anoa/mdbook/docs/README.md#adding-to-the-documentation).

All that being said, there's still a few open-ended questions that I'd like to discuss on this pull request:

1. I've configured the CI through Github Actions to push the built documentation to the `gh-pages` branch via Github Actions. This means that the documentation will initially be available at https://matrix-org.github.io/synapse, and we could configure our DNS to point a more official-looking URL at it for the long term.

    The Github Actions configuration (located [here](https://github.com/matrix-org/synapse/blob/anoa/mdbook/.github/workflows/docs.yaml)) is currently configured to push to the `latest` version of the website on every new commit on `develop`.

[...]

PR #10044, opened 2021-05-22
Add new admin APIs to remove media by media ID from quarantine.

dklimpel:
Related to: #6681, #5956, #10040

### Pull Request Checklist

* [x] Pull request is based on the develop branch
* [x] Pull request includes a [changelog file](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md#changelog). The entry should:
  - Be a short description of your change which makes sense to users. "Fixed a bug that prevented receiving messages from other servers." instead of "Moved X method from `EventStore` to `EventWorkerStore`.".
  - Use markdown where necessary, mostly for `code blocks`.
  - End with either a period (.) or an exclamation mark (!).
  - Start with a capital letter.
* [x] Pull request includes a [sign off](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md#sign-off)
* [x] Code style is correct (run the [linters](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md#code-style))

Signed-off-by: Dirk Klimpel dirk@klimpel.org

PR #10089, opened 2021-05-28
Convert admin api docs to markdown

anoadragon453:
So that they render nicely in mdbook (see #10086), and so that we no longer have a mix of structured text languages in our documentation (excluding files outside of `docs/`).

The largest change in here is definitely the `user_admin_api` docs.


docs/admin_api/purge_history_api.md
```diff
@@ -13,65 +12,67 @@ delete the last message in a room.
...
 
-``POST /_synapse/admin/v1/purge_history/<room_id>[/<event_id>]``
+```
+POST /_synapse/admin/v1/purge_history/<room_id>[/<event_id>]
+```
 
-To use it, you will need to authenticate by providing an ``access_token`` for a
-server admin: see `README.rst <README.rst>`_.
+To use it, you will need to authenticate by providing an `access_token` for a
+server admin: see [Admin API](../../usage/administration/admin_api/index.html).
```
richvdh, line 20: is this link expected to work? It's a 404 right now.

richvdh: likewise a bunch of similar links below.

richvdh (approved):
I haven't reviewed every line of `user_admin_api`, but it looks good other than the below.

richvdh:
It's going to conflict with #10044 I'm afraid.

docs/admin_api/purge_history_api.md
anoadragon453: Ah, there's two things here:

* That link only works in the context of https://github.com/matrix-org/synapse/pull/10086 which this commit was pulled out of.
* We shouldn't link to `...index.html` as that won't work when viewing the docs on GitHub. It should have been `../../usage/administration/admin_api/README.md`. I'll keep this in mind when updating https://github.com/matrix-org/synapse/pull/10086 once this has merged.

anoadragon453:
> It's going to conflict with #10044 I'm afraid.

Actually, this ended up not doing so as only a markdown file was changed :slightly_smiling_face:

PR #10094, opened 2021-05-28
github actions: summarize Sytest results in an easy-to-read format

richvdh:
... using the script from https://github.com/matrix-org/sytest/pull/1052.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -234,7 +234,7 @@ jobs:
         working-directory: /src
       - name: Summarise results.tap
         if: ${{ always() }}
-        run: /scripts/tap_to_gha.pl /logs/results.tap
+        run: /sytest/scripts/tap_to_gha.pl /logs/results.tap
       - name: Upload SyTest logs
         uses: actions/upload-artifact@v2
         if: ${{ always() }}
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -232,9 +232,9 @@ jobs:
       - name: Run SyTest
         run: /bootstrap.sh synapse
         working-directory: /src
-      - name: Dump results.tap
+      - name: Summarise results.tap
         if: ${{ always() }}
-        run: cat /logs/results.tap
+        run: /scripts/tap_to_gha.pl /logs/results.tap
       - name: Upload SyTest logs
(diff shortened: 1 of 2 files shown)


richvdh:
Conveniently, the tests failed so we can see what it looks like:

image

I'm not entirely convinced by having the script exit with a non-zero exit code - it feels like it just adds to the noise. But ymmv.

callahad:
> I'm not entirely convinced by having the script exit with a non-zero exit code - it feels like it just adds to the noise. But ymmv.

It ensures that clicking on `Details` in the CI summary actually takes you to that part of the logs, with the relevant section expanded. Otherwise, you end up in the preceding section (the actual SyTest run) and you have to manually scroll down, expand things, etc.

Example links:

- [Without the exit status](https://github.com/matrix-org/synapse/pull/10090/checks?check_run_id=2693895651)
- [With it (this run)](https://github.com/matrix-org/synapse/pull/10094/checks?check_run_id=2697886815)

callahad:
Alternatively, we could merge the SyTest run and Results summary, e.g.

```yaml
- name: Run SyTest
  working-directory: /src
    run: |
      /bootstrap.sh synapse
      /sytest/scripts/tap_to_gha.pl /logs/results.tap
```

...would still want `tap_to_gha.pl` to set an exit code, though. And may need to put an `|| true` on that first line to ensure that the summarizer always runs, even if bootstrap.sh reports a failure

richvdh:
ah fair enough. Let's just go with this then?

callahad (dismissed):
Looks like `git commit` on c77a9c725ab064bebe9b128b1f3a174e3425dc98 caught some unrelated work. r+ once that history gets rewritten :)

richvdh:
> Looks like `git commit` on [c77a9c7](https://github.com/matrix-org/synapse/commit/c77a9c725ab064bebe9b128b1f3a174e3425dc98) caught some unrelated work. r+ once that history gets rewritten :)

oops, thanks.

richvdh:
> r+ once that history gets rewritten :)

I'm interpreting this to mean that you don't want to re-review this :)

callahad:
> > r+ once that history gets rewritten :)
> 
> I'm interpreting this to mean that you don't want to re-review this :)

Yep :)

callahad (approved):

PR #10101, opened 2021-06-01
Announce deprecation of experimental `msc2858_enabled` option.

### LINKED ISSUE #9627: deprecate unstable MSC2858 APIs
As of #9617, Synapse now supports the final stable API interface for MSC2858 (https://github.com/matrix-org/matrix-doc/pull/2858). We should define a deprecation period for the unstable API and eventually remove it.

erikjohnston:
c.f. https://github.com/matrix-org/synapse/pull/9617 and https://github.com/matrix-org/matrix-doc/blob/master/proposals/2858-Multiple-SSO-Identity-Providers.md

Fixes #9627.


richvdh (changes requested):
I would replace all of this with a simple `.removal` newsfile, along the lines of:

> The core Synapse development team plan to drop support for the [unstable API of MSC2858](https://github.com/matrix-org/matrix-doc/blob/master/proposals/2858-Multiple-SSO-Identity-Providers.md#unstable-prefix), including the undocumented `experimental.msc2858_enabled` config option, in August 2021. Client authors should ensure that their clients are updated to use the stable API (which has been supported since Synapse 1.30) well before that time, to give their users time to upgrade.

anoadragon453 (approved):

callahad:
I've filed #10103 to track the actual removal

PR #10102, opened 2021-06-01
Make /sync do less state res

erikjohnston:


richvdh (approved):
lgtm

should we go through the other places `StateHandler.get_current_state_ids` is called and see how many we can fix?

erikjohnston:
> should we go through the other places `StateHandler.get_current_state_ids` is called and see how many we can fix?

I thought I had TBH. I don't see any other instances that aren't passing in `prev_events`

PR #10107, opened 2021-06-02
Always update AS last_pos, even on no events

### LINKED ISSUE #1834: appservice_stream_position can get into such a state that ASes break silently
I wasted 3h on arasphere last night trying to work out why it wouldn't send stuff to ASes. turns out the value in that table was somehow mangled. deleting it didn't help, but manually reinserting max streampos did. Plus it's inconsistently named wrt all the other AS stuff :(

14mRh4X0r:
Fixes #1834.

`get_new_events_for_appservice` internally calls `get_events_as_list`, which will filter out any rejected events. If all returned events are filtered out, `_notify_interested_services` will return without updating the last handled stream position. If there are 100 consecutive such events, processing will halt altogether.

Breaking the loop is now done by checking whether we're up-to-date with `current_max` in the loop condition, instead of relying on an empty `events` list.

### Pull Request Checklist

### DIFF
--- a/changelog.d/10107.bugfix
+++ b/changelog.d/10107.bugfix
@@ -1,0 +1,1 @@
+Fixed a bug that could cause Synapse to stop notifying application services. Contributed by Willem Mulder.
--- a/synapse/handlers/appservice.py
+++ b/synapse/handlers/appservice.py
@@ -87,16 +87,14 @@ class ApplicationServicesHandler:
             self.is_processing = True
             try:
                 limit = 100
-                while True:
+                upper_bound = -1
+                while upper_bound < self.current_max:
                     (
                         upper_bound,
                         events,
                     ) = await self.store.get_new_events_for_appservice(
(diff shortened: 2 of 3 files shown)


erikjohnston (commented):
I think this looks good. Would you be happy taking a stab at adding a regression test? I think that sending, say, 200 rejected events and then a valid event should be enough?

14mRh4X0r:
Sure, I could do that, I'll need some hints on where to start though. I also tried to cover that scenario in the `test_notify_interested_services` unit test, by first returning an empty array before returning the actual event.

erikjohnston:
> Sure, I could do that, I'll need some hints on where to start though. I also tried to cover that scenario in the `test_notify_interested_services` unit test, by first returning an empty array before returning the actual event.

Oh, so you did. I completely blanked that you switched the order of the empty lists! I think that's good enough for now :+1:

PR #10108, opened 2021-06-02
Implement config option `sso.update_profile_information`

jkanefendt:
Implemented config option sso.update_display_name to keep user's display name in sync with the SSO display name

Signed-off-by: Johannes Kanefendt <johannes.kanefendt@krzn.de>


anoadragon453:
Hi @jkanefendt, if you run `scripts-dev/generate_sample_config` that should regenerate a new `docs/sample_config.yaml` file. The currently proposed file looks like it contains some unintended changes.

synapse/config/sso.py
```diff
@@ -110,7 +112,10 @@ def generate_config_section(self, **kwargs):
             #client_whitelist:
             #  - https://riot.im/develop
             #  - https://my.custom.client/
-
+            #
+            # Keep a user's display name in sync with the SSO display name
```
anoadragon453, line 116:
```suggestion
            # Keep a user's display name in sync with the SSO display name
            #
```

It'd also be good to mention that a user's SSO profile information will only be updated after each log in to Synapse.

synapse/config/sso.py
```diff
@@ -74,6 +74,8 @@ def read_config(self, config, **kwargs):
 
         self.sso_client_whitelist = sso_config.get("client_whitelist") or []
 
+        self.sso_update_display_name = sso_config.get("update_display_name") or False
+
```
anoadragon453, line 78: While having `False` here does preserve existing behaviour, I almost feel like this option should be enabled by default, as I imagine it's expected for user details on SSO-enabled systems to update automatically.

Do you have any thoughts @clokep?

anoadragon453, line 78: Additionally, I wonder if it's likely that we'll ever add any more auto-updating information, such as user avatars. It'd be ideal to have both under the same option. Perhaps `update_profile_information` may be more future-proof?

changelog.d/10108.feature
```diff
@@ -0,0 +1 @@
+Implemented config option sso.update_display_name to keep user's display name in sync with the SSO display name.
```
anoadragon453, line 1:
```suggestion
Implement config option `sso.update_display_name` to keep user's display name in sync with the SSO display name.
```

Just to keep in line with existing changelog entries, which are mostly in the present tense :slightly_smiling_face:

docs/sample_config.yaml
```diff
@@ -67,12 +67,11 @@ pid_file: DATADIR/homeserver.pid
 #
```
anoadragon453, line 67: Please run `scripts-dev/generate_sample_config` to fix this file and the logging sample config.

anoadragon453 (changes requested):
Thank you for your work!

The code where we actually update the user's display name looks good - I have just a few questions surrounding that.

synapse/config/sso.py
clokep: It does seem like people would want this automatically, we should probably just make sure we document it. 🤷 

It probably depends on if people are using corporate-type SSO vs. social login sort of SSO though...

clokep: There's also e-mails...

clokep:
A thing that is a bit odd about this is that we have a page which asks people if they want to use their display name / avatar / which emails on their Matrix account. If someone chooses not to import their display name, but then logs in after it would overwrite it automatically. It almost seems like this should be a "force_sso_display_name" or something to make those features interact reasonably.

anoadragon453:
> It almost seems like this should be a "force_sso_display_name" or something to make those features interact reasonably.

That would also be useful if the organisation wants user details to stay in sync on all platforms regardless of user input.

We could eventually extend that even further, where if that option were False, there would be a page that asks the user whether they want to update their details after a change on the IdP. That would require keeping some information about those details and whether they user has already chosen not to transfer them, but would be nice UX.

synapse/config/sso.py
anoadragon453: Indeed, which are already implemented. Seems best to generalise the option name then, but to document that only display names are transferred for now.

How about `sync_sso_profile_information`?

anoadragon453: > It does seem like people would want this automatically, we should probably just make sure we document it. :shrug:

This sounds reasonable.

> It probably depends on if people are using corporate-type SSO vs. social login sort of SSO though...

I think the majority of people would expect things to sync by default. I can also see the case where I might change my social media handle frequently, but not want my displayname in all rooms to update every time. So the option is handy.

docs/sample_config.yaml
anoadragon453: Looks correct to me! Also thank you for fixing a typo in the log config :)

synapse/config/sso.py
```diff
@@ -110,6 +114,12 @@ def generate_config_section(self, **kwargs):
             #client_whitelist:
             #  - https://riot.im/develop
             #  - https://my.custom.client/
+            #
+            # Keep a user's profile fields in sync with the values from the SSO ticket
+            # (currently display_name). The update is performed on every SSO login.
```
anoadragon453, line 119: In line with what @clokep [noted above](https://github.com/matrix-org/synapse/pull/10108#issuecomment-853261726), we should inform the sysadmin that setting this option to `true` will overwrite profile information of users from the SSO ticket, regardless of whether they've chosen to opt out of importing that information when first logging in or not.

```suggestion
            # Keep a user's profile fields in sync with the values from the SSO ticket
            # (currently display_name). The update is performed on every SSO login.
            #
            # Note that enabling this option will override user profile information from
            # the SSO ticket, regardless of whether users have opted-out of syncing
            # that information when first signing in. Defaults to false.
```

anoadragon453 (changes requested):
This is looking pretty close. I'd just like to document the potential pitfall with syncing user information automatically, but otherwise I think this is fine for now.

Note that you'll need to regenerate the sample config again after changing `synapse/config/sso.py`

changelog.d/10108.feature
```diff
@@ -0,0 +1 @@
+Implement config option sso.update_profile_information to keep a user's profile fields in sync with the values from the SSO ticket (currently display_name).
```
anoadragon453, line 1:
```suggestion
Implement config option `sso.update_profile_information` to sync SSO users' profile information with the identity provider each time they login. Currently only displayname is supported.
```

As "SSO ticket" might not be a term most sysadmins are immediately familiar with.

synapse/config/sso.py
```diff
@@ -110,6 +114,14 @@ def generate_config_section(self, **kwargs):
             #client_whitelist:
             #  - https://riot.im/develop
             #  - https://my.custom.client/
```
anoadragon453, line 116:
```suggestion
            #  - https://my.custom.client/

```

Heads up options should have a blank link separating them from each other for readability. See the [configuration file format guide](https://matrix-org.github.io/synapse/develop/code_style.html#configuration-file-format) for more details.

synapse/config/sso.py
```diff
@@ -110,6 +114,14 @@ def generate_config_section(self, **kwargs):
...
+            # Keep a user's profile fields in sync with the values from the SSO ticket
+            # (currently display_name). The update is performed on every SSO login.
```
anoadragon453, line 118:
```suggestion
            # Keep a user's profile fields in sync with information from the identity provider. Currently only
            # overriding the displayname is supported. Fields are checked on every SSO login, and are updated
            # if necessary.
```

synapse/handlers/sso.py
```diff
@@ -458,6 +466,24 @@ async def complete_sso_login_request(
                     request.getClientIP(),
                 )
                 new_user = True
+            else:
+                if self._sso_update_profile_information:
```
anoadragon453, line 470: This can be `elif self._sso_update_profile_information:`.

anoadragon453 (commented):
Sorry for the long RTT. Some final thoughts, but otherwise this looks good to me.

synapse/handlers/sso.py
jkanefendt: See https://github.com/matrix-org/synapse/pull/10108/commits/748fcf5429eb63430a8a2eb8f2caad8529f66f86

anoadragon453 (approved):
I think this looks good to me now - thanks for the going through the iterations with us!

PR #10109, opened 2021-06-02
Do not show invite-only rooms in spaces summary (unless joined/invited).

### LINKED ISSUE #10104: all space members are able to see invite only rooms in spaces
### Description
After upgrading synapse from 1.34 to 1.35 (debian testing), space members with default power level in the space are able to see invite only rooms in the space's "Explore rooms"-view that they are not allowed to join.

### Steps to reproduce

- create a private space S with account A
- add/create a public room P in space S with account A
- add/create a room R in space S with account A set to:
  - "Who can access this room?": "Only people who have been invited"
  - "Who can read history?": "Members only (since they joined)"
- invite account B to space S

### Expected outcome
- account B is able to see room P in the "Explore rooms"-view of the space
- account B is *not* able to see room R in the "Explore rooms"-view

### Actual outcome
- account B is able to see room P in the "Explore rooms"-view and is able to join it.
- account B *is* able to see room R in the "Explore rooms"-view. However, when trying to join, account b is presented with the error message:
Failed to join room
You are not invited to this room.

As per https://gist.github.com/t3chguy/fdf541446d992d1fc96b8089b342fe7b#gistcomment-3755026, this does not seem to be the intended behavior.

### Version information
- **Homeserver**: matrix.pack.rocks (ran by me)
- **Version**:  1.35.0+bullseye1
- **Install method**: debs from https://matrix.org/packages/debian/
- **Platform**: qemu-VM on Debian 11, no containers involved
- **Default Room Version**: 6
I did set `experimental_features: { spaces_enabled: true }` in /etc/matrix-synapse/homeserver.yaml in 1.34 which is still present. Removing it does not change the behavior described above.

clokep:
This fixes #10104 to not show invite-only rooms to everyone in a space.

This was due to a bug in the refactoring done in #9922 where we called a method expecting an exception to be raised, but the function bailed early if the restricted join rules weren't used. To fix it, I called some lower-level functions which seems clearer.

### DIFF
--- a/changelog.d/10109.bugfix
+++ b/changelog.d/10109.bugfix
@@ -1,0 +1,1 @@
+Fix a bug introduced in v1.35.0 where invite-only rooms would be shown to users in a space who were not invited.
--- a/synapse/handlers/space_summary.py
+++ b/synapse/handlers/space_summary.py
@@ -26,7 +26,6 @@ from synapse.api.constants import (
     HistoryVisibility,
     Membership,
 )
-from synapse.api.errors import AuthError
 from synapse.events import EventBase
 from synapse.events.utils import format_event_for_client_v2
 from synapse.types import JsonDict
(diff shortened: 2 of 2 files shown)


herbetom:
I applied it manually on top of `1.35.0+buster1` and it seems to work :+1: 

With this Spaces are empty and only the rooms a user has joined are shown (not even public rooms).

andir:
Pardon me commenting here (I am just someone watching from the sidelines): Shouldn't this have a regression test? This can be kind of a privacy/security issue when topics leak into the public.

alturiak:
> I applied it manually on top of `1.35.0+buster1` and it seems to work 👍
> 
> With this Spaces are empty and only the rooms a user has joined are shown (not even public rooms).

Hm, public, peekable (history visibility set to "Everyone") rooms should be displayed, though.
This might be worth testing explicitly again.

herbetom:
> (history visibility set to "Everyone")

Okay, it was a public room, but history was set to "Members only (since they were invited)". If i set it to "Everyone" it works. :+1:

clokep:
> Pardon me commenting here (I am just someone watching from the sidelines): Shouldn't this have a regression test? This can be kind of a privacy/security issue when topics leak into the public.

Of course. I'm writing those as we speak.

clokep:
Also see matrix-org/complement#118.

erikjohnston (approved):

PR #10115, opened 2021-06-03
Fix 'ip_range_whitelist' not working for federation servers

### LINKED ISSUE #9569: ip_range_whitelist has no impact on federation client
### Description
`ip_range_whitelist` does not affect federation ip blacklisting. I have an internal matrix server which cannot federate with other servers on the same network. 

### Steps to reproduce

- Setup a homeserver using an internal ip address
- Set `ip_range_whitelist` to include the other servers ip address
- Attempt to federate with it

Relevant log entry

```
INFO - None - Dropped 10.x.x.x from DNS resolution to hs-name due to blacklist
```

Looking at the code the federation client is setup without passing an ip_whitelist to its DNS resolver. Only application services appears to at present use the `ip_range_whitelist` attribute.

### Version information

- **Version**: 1.29.0

- **Install method**: debian package

- **Platform**: debian

mikure:
This fixes #9569, which prevented 'ip_range_whitelist' from working for federation servers.

This implements the changes already suggested by @clokep to still provide backwards-compatibility, if 'federation_ip_range_blacklist' is set. Otherwise 'ip_range_whitelist' will be used.
I hopefully got the idea right. But some tests showed the expected results.


mikure:
I just realized, that I the wrong named changelog file was accidentally added. I hope, it is correct now.

changelog.d/10115.bugfix
```diff
@@ -0,0 +1,2 @@
+Fix a bug that prevented the 'ip_range_whitelist' configuration to work for federation and identity servers.
```
richvdh, line 1:
```suggestion
Fix a bug introduced in Synapse v1.25.0 that prevented the `ip_range_whitelist` configuration option from working for federation and identity servers. Contributed by @mikure.
```

changelog.d/10115.bugfix
```diff
@@ -0,0 +1,2 @@
...
+Notice that this change will not be applied, if your configuration still uses 'federation_ip_range_blacklist', to allow backwards-compatibility.
```
richvdh, line 2: we wouldn't normally include this level of detail in the changelog. I think you can safely omit it.
```suggestion
```

synapse/config/server.py
```diff
@@ -410,6 +410,11 @@ def read_config(self, config, **kwargs):
             ["0.0.0.0", "::"],
             config_path=("federation_ip_range_blacklist",),
         )
+        # The federation_ip_range_whitelist is used to provide
+        # backwards-compatibility, if federation_ip_range_blacklist is set.
```
richvdh, line 414: please could you add a comment explaining why this is a sensible thing to do? It's not obvious at first glance why we would check `federation_ip_range_blacklist` when setting `federation_ip_range_whitelist`.

richvdh (changes requested):
Thanks for the contribution! A few suggestions below.

synapse/config/server.py
clokep: It's because there never was an `federation_ip_range_whitelist` before, so if you're using the old `federation_ip_range_blacklist` config to ensure backwards compatibility you should *not* apply `ip_range_whitelist` against it.

changelog.d/10115.bugfix
clokep: This feels a bit odd to me since this wasn't a thing that ever worked -- I don't think it is really a bug, but a new feature? I guess it depends on how you look at it.

richvdh: I had the same thought, but the sample config says of `ip_range_whitelist`:

> # List of IP address CIDR ranges that should be allowed for federation,
> # identity servers, push servers, ...

which is just wrong. So whether it be in the docs or the impl, it seems there was a bug.

synapse/config/server.py
richvdh: sorry, yes, having stared at #9569 I figured it out. What I'm saying is that this comment needs to say it, to save everyone who ever reads that bit of code having to do the same detective work.

changelog.d/10115.bugfix
clokep: Ah, fair enough! I didn't realize the docs were that specific. 👍

mikure: Thank you for your feedback.
For me it felt like a bug, as the `ip_range_whitelist` did not work for my test setup, although the comment claimed otherwise.

changelog.d/10115.bugfix
mikure: Thank you. I thought, it might be interesting for people, who still use `federation_ip_range_blacklist`. So I preferred to add some more details, just in case.
I'll remove that line.

synapse/config/server.py
mikure: The comment has been updated. However, the explanation still feels a bit complex and confusing.

synapse/config/server.py
```diff
@@ -410,6 +410,13 @@ def read_config(self, config, **kwargs):
             ["0.0.0.0", "::"],
             config_path=("federation_ip_range_blacklist",),
         )
+        # The federation_ip_range_whitelist is used for backwards-compatibility
+        # and will always be None, as it was never set in any configuration files.
+        # If no backwards-compatibility is required, i.e. the configuration file
+        # doesn't set federation_ip_range_blacklist, use ip_range_whitelist instead.
```
richvdh, line 416: Hrm yes, this is still a bit confusing.

How about something like this:

```python
        # federation_ip_range_blacklist is used for backwards-compatibility
        # and only applies to federation and identity servers.
        if "federation_ip_range_blacklist" in config:
            # Always blacklist 0.0.0.0, ::
            self.federation_ip_range_blacklist = generate_ip_set(
                config["federation_ip_range_blacklist"],
                ["0.0.0.0", "::"],
                config_path=("federation_ip_range_blacklist",),
            )
            # 'federation_ip_range_whitelist' was never a supported configuration option.
            self.federation_ip_range_whitelist = None
        else:
            # federation_ip_range_blacklist is not given.
            # Default to ip_range_blacklist and ip_range_whitelist.
            self.federation_ip_range_blacklist = self.ip_range_blacklist
            self.federation_ip_range_whitelist = self.ip_range_whitelist
```

To me, that feels like makes the logic clear without needing any extra explanation. @mikure , @clokep wdyt?

clokep: That seems quite a bit more explicit, I think it is clearer. 👍

mikure: I agree. That makes is much clearer. I'll test and update later, when I have more time.

mikure: I just finished testing and I couldn't find any combinations, that didn't work as I expected.
It's also easier to not miss any combinations, as the code is much clearer now. Thank you for this suggestion, @richvdh.

richvdh (approved):
looks great. thanks very much!

richvdh (changes requested):
sorry, one last thing. We need your agreement to include this in Synapse.

Please could you add a comment to the PR with the contents:

> ```
> Signed-off-by: Your Name <your@email.example.org>
> ```

... as per https://github.com/matrix-org/synapse/blob/develop/CONTRIBUTING.md#sign-off ?

mikure:
I'm sorry. I think, I misread that section and assumed it would be added automatically.

Signed-off-by: Michael Kutzner <1mikure@gmail.com>

richvdh:
great, thank you!

PR #10117, opened 2021-06-03
When joining a remote room limit the number of events we concurrently check signatures/hashes for

erikjohnston:
If we do hundreds of thousands at once the memory overhead can easily reach 500+ MB.

I also took the opportunity to change `_check_sigs_and_hashes` to use async/await.


synapse/federation/federation_base.py
```diff
@@ -48,22 +38,14 @@ def __init__(self, hs):
...
+        """Checks that event is correctly signed by the sending server.
 
         Args:
-            room_version: The room version of the PDUs
-            pdus: the events to be checked
+            room_version: The room version of the PDU
+            pdu: the event to be checked
 
         Returns:
             For each input event, a deferred which:
```
richvdh, line 51: this needs an update

synapse/federation/federation_base.py
```diff
@@ -73,77 +55,60 @@ def _check_sigs_and_hashes(
...
-                            "Event %s content has been tampered, redacting",
-                            pdu.event_id,
-                        )
-                    return redacted_event
-
-                result = yield defer.ensureDeferred(
-                    self.spam_checker.check_event_for_spam(pdu)
+        try:
+            await _check_sigs_on_pdu(self.keyring, room_version, pdu)
+        except Exception as e:
```
richvdh, line 60: was changing this from just SynapseError deliberate?

synapse/federation/federation_base.py
```diff
@@ -73,77 +55,60 @@ def _check_sigs_and_hashes(
...
 class PduToCheckSig(namedtuple("PduToCheckSig", ["pdu", "sender_domain", "deferreds"])):
     pass
 
 
-def _check_sigs_on_pdus(
-    keyring: Keyring, room_version: RoomVersion, pdus: Iterable[EventBase]
-) -> List[Deferred]:
+async def _check_sigs_on_pdu(
+    keyring: Keyring, room_version: RoomVersion, pdu: EventBase
+):
```
richvdh, line 111:
```suggestion
) -> None:
```

synapse/federation/federation_base.py
```diff
@@ -73,77 +55,60 @@ def _check_sigs_and_hashes(
...
     """Check that the given events are correctly signed
```
richvdh, line 112: docstring needs an update

synapse/federation/federation_base.py
```diff
@@ -177,90 +142,47 @@ def _check_sigs_on_pdus(
...
-                    p.pdu.origin_server_ts if room_version.enforce_key_validity else 0,
-                )
-                for p in pdus_to_check_event_id
-            ]
-        )
-
-        def event_err(e, pdu_to_check):
+    if room_version.event_format == EventFormatVersions.V1 and get_domain_from_id(
+        pdu.event_id
+    ):
```
richvdh, line 168: this seems to have lost the comparison with `pdu.sender_domain`

synapse/federation/federation_client.py
```diff
@@ -696,14 +715,29 @@ async def send_request(destination) -> Dict[str, Any]:
...
+                    outlier=True,
+                    room_version=room_version,
+                )
+
+                if valid_pdu:
+                    valid_pdus_map[valid_pdu.event_id] = valid_pdu
+
+            await concurrently_execute(
+                _execute, itertools.chain(state, auth_chain), 10000
+            )
```
richvdh, line 740: I don't think `concurrently_execute` handles a limit of 10000 elegantly.

synapse/federation/federation_client.py
```diff
@@ -386,51 +383,75 @@ async def _check_sigs_and_hash_and_fetch(
...
             A list of PDUs that have valid signatures and hashes.
         """
-        deferreds = self._check_sigs_and_hashes(room_version, pdus)
+        valid_pdus = await yieldable_gather_results(
+            self._check_sigs_and_hash_and_fetch_one,
+            pdus,
+            origin=origin,
+            room_version=room_version,
+            outlier=outlier,
+        )
```
richvdh, line 392: does this need a concurrency limit?

synapse/federation/federation_base.py
erikjohnston: Woops, good spot.

synapse/federation/federation_client.py
erikjohnston: Ah, yeah, may as well actually. I've left it so we still do the limiting manually in the `send_join` function, as that we we don't need to convert from list to map after the fact.

synapse/federation/federation_client.py
erikjohnston: I've fixed it up to handle large limits a bit better.

erikjohnston:
The tests were failing due to the change of `concurrently_execute`, which I think is just a race fixed by: https://github.com/matrix-org/sytest/pull/1057

synapse/util/async_helpers.py
```diff
@@ -173,20 +176,26 @@ def concurrently_execute(
...
             pass
 
+    # We use `zip` to handle the case where the number of args is less than the
+    # limit, avoiding needlessly spawning unnecessary background tasks.
     return make_deferred_yieldable(
         defer.gatherResults(
-            [run_in_background(_concurrently_execute_inner) for _ in range(limit)],
+            [
+                run_in_background(_concurrently_execute_inner, value)
+                for value, _ in zip(it, range(limit))
```
richvdh, line 197: are you sure `zip` is guaranteed not to drop values when one of the iterators runs out of inputs?

https://docs.python.org/3/library/functions.html#zip says:

> `zip()` should only be used with unequal length inputs when you don’t care about trailing, unmatched values from the longer iterables.

indeed:

```
>>> a = [1,2,3,4]
>>> b = [1,2]
>>> 
>>> a_iter = iter(a)
>>> b_iter = iter(b)
>>> list(zip(a_iter, b_iter))
[(1, 1), (2, 2)]
>>> next(a_iter)
4
```

3 went missing :/

richvdh (changes requested):
^

synapse/util/async_helpers.py
erikjohnston: Oh ugh. Good spot.

erikjohnston: Hmm, I wonder if:

> Returns an iterator of tuples, where the i-th tuple contains the i-th element from each of the argument sequences or iterables.

means that its safe if we do `zip(range(n), it)`? We probably just want to do it in a different way...

erikjohnston: Oh, using `itertools.islice(it, limit)` is probably the right way of doing this, and cleaner...

richvdh (approved):
lgtm

PR #10119, opened 2021-06-03
Improve event caching code

erikjohnston:
This PR mainly does two things:
1. Ensure we only load an event from the DB once when the same event is requested multiple times at once.
2. Ensure that there is only one copy of any given event in memory at a time by using weak references.

### DIFF
--- a/changelog.d/10119.misc
+++ b/changelog.d/10119.misc
@@ -1,0 +1,1 @@
+Try and ensure we only have one copy of an event in memory at a time.
--- a/changelog.d/10119.misc
+++ b/changelog.d/10119.misc
@@ -1,1 +1,1 @@
-Try and ensure we only have one copy of an event in memory at a time.
+Improve event caching mechanism to avoid having multiple copies of an event in memory at a time.
--- a/synapse/events/__init__.py
+++ b/synapse/events/__init__.py
@@ -97,7 +97,7 @@ class DefaultDictProperty(DictProperty):
 
 
 class _EventInternalMetadata:
-    __slots__ = ["_dict", "stream_ordering", "outlier", "redacted_by"]
+    __slots__ = ["_dict", "stream_ordering", "outlier"]
 
     def __init__(self, internal_metadata_dict: JsonDict):
         # we have to copy the dict, because it turns out that the same dict is
@@ -110,10 +110,6 @@ class _EventInternalMetadata:
         # whether this event is an outlier (ie, whether we have the state at that point
         # in the DAG)
         self.outlier = False
-
-        # Whether this event has a valid redaction event pointing at it (i.e.
-        # whether it should be redacted before giving to clients).
-        self.redacted_by: Optional[str] = None
 
     out_of_band_membership: bool = DictProperty("out_of_band_membership")
     send_on_behalf_of: str = DictProperty("send_on_behalf_of")
--- a/synapse/events/__init__.py
+++ b/synapse/events/__init__.py
@@ -97,7 +97,7 @@ class DefaultDictProperty(DictProperty):
 
 
 class _EventInternalMetadata:
-    __slots__ = ["_dict", "stream_ordering", "outlier"]
+    __slots__ = ["_dict", "stream_ordering", "outlier", "redacted_by"]
 
     def __init__(self, internal_metadata_dict: JsonDict):
         # we have to copy the dict, because it turns out that the same dict is
@@ -110,6 +110,10 @@ class _EventInternalMetadata:
         # whether this event is an outlier (ie, whether we have the state at that point
         # in the DAG)
         self.outlier = False
+
+        # Whether this event has a valid redaction event pointing at it (i.e.
+        # whether it should be redacted before giving to clients).
+        self.redacted_by: Optional[str] = None
 
     out_of_band_membership: bool = DictProperty("out_of_band_membership")
     send_on_behalf_of: str = DictProperty("send_on_behalf_of")
--- a/synapse/storage/databases/main/events_worker.py
+++ b/synapse/storage/databases/main/events_worker.py
@@ -14,7 +14,6 @@
 
 import logging
 import threading
-from collections import namedtuple
 from typing import (
     Collection,
     Container,
@@ -27,6 +26,7 @@ from typing import (
     overload,
 )
 
+import attr
 from constantly import NamedConstant, Names
 from typing_extensions import Literal
 
@@ -74,7 +74,10 @@ EVENT_QUEUE_ITERATIONS = 3  # No. times we block waiting for requests for events
 EVENT_QUEUE_TIMEOUT_S = 0.1  # Timeout when waiting for requests for events
 
 
-_EventCacheEntry = namedtuple("_EventCacheEntry", ("event", "redacted_event"))
+@attr.s(slots=True, auto_attribs=True)
+class _EventCacheEntry:
+    event: EventBase
+    redacted_event: Optional[EventBase]
 
 
 class EventRedactBehaviour(Names):
@@ -476,7 +479,9 @@ class EventsWorkerStore(SQLBaseStore):
 
         return events
 
-    async def _get_events_from_cache_or_db(self, event_ids, allow_rejected=False):
+    async def _get_events_from_cache_or_db(
+        self, event_ids: Iterable[str], allow_rejected: bool = False
+    ) -> Dict[str, _EventCacheEntry]:
         """Fetch a bunch of events from the cache or the database.
 
         If events are pulled from the database, they will be cached for future lookups.
@@ -485,14 +490,13 @@ class EventsWorkerStore(SQLBaseStore):
 
         Args:
 
-            event_ids (Iterable[str]): The event_ids of the events to fetch
+            event_ids: The event_ids of the events to fetch
 
-            allow_rejected (bool): Whether to include rejected events. If False,
+            allow_rejected: Whether to include rejected events. If False,
                 rejected events are omitted from the response.
 
         Returns:
-            Dict[str, _EventCacheEntry]:
-                map from event id to result
+            map from event id to result
         """
         event_entry_map = self._get_events_from_cache(
             event_ids,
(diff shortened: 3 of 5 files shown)


synapse/storage/databases/main/events_worker.py
```diff
@@ -485,20 +505,48 @@ async def _get_events_from_cache_or_db(self, event_ids, allow_rejected=False):
...
+        # we're going to pull from the DB. We use a single deferred that
+        # resolves to all the events we pulled from the DB (this will result in
+        # this function returning more events than requested, but that can
+        # happen already due to `_get_events_from_db`).
+        fetching_deferred = ObservableDeferred(defer.Deferred())
+
+        for event_id in missing_events_ids:
+            deferred = self._current_event_fetches.get(event_id)
+            if deferred is not None:
+                # We're already pulling the event out of the DB, ad the deferred
```
richvdh, line 537:
```suggestion
                # We're already pulling the event out of the DB. Add the deferred
```

synapse/storage/databases/main/events_worker.py
```diff
@@ -485,20 +505,48 @@ async def _get_events_from_cache_or_db(self, event_ids, allow_rejected=False):
...
+                # to the collection of deferreds to wait on.
+                already_fetching[event_id] = deferred.observe()
+            else:
+                # We're not already pulling the event from the DB, so add our
+                # deferred to the the map of events that are being fetched.
+                self._current_event_fetches[event_id] = fetching_deferred
+                fetching_deferred.observe().addBoth(
+                    lambda _, event_id: self._current_event_fetches.pop(event_id, None),
```
richvdh, line 545: why not do this for all the `missing_events_ids` before calling `fetching_deferred.callback` at line 570?

synapse/storage/databases/main/events_worker.py
```diff
@@ -485,20 +505,48 @@ async def _get_events_from_cache_or_db(self, event_ids, allow_rejected=False):
...
+        # if so we wait for those lookups to finish instead of pulling the same
+        # events out of the DB multiple times.
+        already_fetching: Dict[str, defer.Deferred] = {}
+
+        # We also add entries to `self._current_event_fetches` for each event
+        # we're going to pull from the DB. We use a single deferred that
+        # resolves to all the events we pulled from the DB (this will result in
+        # this function returning more events than requested, but that can
+        # happen already due to `_get_events_from_db`).
+        fetching_deferred = ObservableDeferred(defer.Deferred())
```
richvdh, line 532: we could skip building this in the case that all fetches are in flight by moving this line, and `self._current_event_fetches[event_id] = fetching_deferred`, into the `if missing_events_ids` block.

synapse/storage/databases/main/events_worker.py
```diff
@@ -509,43 +557,85 @@ async def _get_events_from_cache_or_db(self, event_ids, allow_rejected=False):
...
+                missing_events = await self._get_events_from_db(
+                    missing_events_ids,
+                )
+
+                event_entry_map.update(missing_events)
+            except Exception as e:
+                fetching_deferred.errback(e)
+                raise e
+
+            fetching_deferred.callback(missing_events)
```
richvdh, line 570: I think you need to drop the logcontext before doing .callback or .errback on observabledeferreds?

synapse/storage/databases/main/events_worker.py
```diff
@@ -509,43 +557,85 @@ async def _get_events_from_cache_or_db(self, event_ids, allow_rejected=False):
...
-            dict of event_id -> _EventCacheEntry for each event_id in cache. If
-            allow_rejected is `False` then there will still be an entry but it
-            will be `None`
+            events: list of event_ids to fetch
+            update_metrics: Whether to update the cache hit ratio metrics
         """
         event_map = {}
 
         for event_id in events:
+            # First check if its in the event cache
```
richvdh, line 611:
```suggestion
            # First check if it's in the event cache
```

synapse/storage/databases/main/events_worker.py
```diff
@@ -509,43 +557,85 @@ async def _get_events_from_cache_or_db(self, event_ids, allow_rejected=False):
...
-            if allow_rejected or not ret.event.rejected_reason:
+            if ret:
                 event_map[event_id] = ret
-            else:
-                event_map[event_id] = None
+
+            # Otherwise check if we still have the event in memory.
+            event = self._event_ref.get(event_id)
+            if event:
+                redacted_event = None
```
richvdh, line 621:
```suggestion
                # reconstruct an event cache entry
                
                redacted_event = None
```

synapse/storage/databases/main/events_worker.py
```diff
@@ -601,13 +608,34 @@ def _get_events_from_cache(
...
+                if event.internal_metadata.redacted_by is not None:
+                    # The event has been redacted, so we generate a redacted
+                    # version.
+                    redacted_event = prune_event(event)
+                    redacted_event.unsigned[
+                        "redacted_by"
+                    ] = event.internal_metadata.redacted_by
```
richvdh, line 628: do we not need `redacted_because` here?

synapse/storage/databases/main/events_worker.py
```diff
@@ -601,13 +608,34 @@ def _get_events_from_cache(
...
+                cache_entry = _EventCacheEntry(
+                    event=event,
+                    redacted_event=redacted_event,
+                )
+                event_map[event_id] = cache_entry
 
-            event_map[event_id] = ret
+                # We add the entry back into the cache as we want to keep
+                # recently queried events in the cache.
+                self._get_event_cache.set((event_id,), cache_entry)
```
richvdh: could we refactor the similar code in `_get_events_from_db`/`_maybe_redact_event_row` so that we can reuse some common code rather than have two similar but slightly different impls?


For example: have `_maybe_redact_event_row` just return a boolean rather than the redaction; have `_get_events_from_db` poke `internal_metadata.redacted_by` based on that boolean; then move these lines here out to a `build_cache_entry` which can be called by `_get_events_from_db`.

richvdh (changes requested):
looks like it should be a significant improvement.

I feel a bit like there are two PRs' worth of changes here: one to dedupe concurrent lookups, and one to use the weakref magic. ymmv though.

erikjohnston:
> I feel a bit like there are two PRs' worth of changes here: one to dedupe concurrent lookups, and one to use the weakref magic. ymmv though.

Err, yes indeed. I have done so.

synapse/events/__init__.py
```diff
@@ -111,6 +111,10 @@ def __init__(self, internal_metadata_dict: JsonDict):
         # in the DAG)
         self.outlier = False
 
+        # Whether this event has a valid redaction event pointing at it (i.e.
+        # whether it should be redacted before giving to clients).
+        self.redacted_by: Optional[str] = None
```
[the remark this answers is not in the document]
erikjohnston: This code will be split out into its own PR.

synapse/storage/databases/main/events_worker.py
erikjohnston: This code will be split out into its own PR.

synapse/storage/databases/main/events_worker.py
erikjohnston: This code will be split out into its own PR.

synapse/storage/databases/main/events_worker.py
erikjohnston: This code will be split out into its own PR.

synapse/storage/databases/main/events_worker.py
```diff
@@ -564,10 +558,16 @@ async def _get_events_from_cache_or_db(
...
                 event_entry_map.update(missing_events)
             except Exception as e:
-                fetching_deferred.errback(e)
+                with PreserveLoggingContext():
+                    fetching_deferred.errback(e)
                 raise e
+            finally:
+                # Ensure that we mark these events as no longer being fetched.
+                for event_id in missing_events_ids:
+                    self._current_event_fetches.pop(event_id, None)
```
richvdh, line 567: I wonder if doing this after  `fetching_deferred.errback` could cause races. I can't really think how it could, though.

synapse/storage/databases/main/events_worker.py
```diff
@@ -593,13 +593,11 @@ async def _get_events_from_cache_or_db(
...
     def _invalidate_get_event_cache(self, event_id):
         self._get_event_cache.invalidate((event_id,))
-        self._event_ref.pop(event_id, None)
-        self._current_event_fetches.pop(event_id, None)
 
     def _get_events_from_cache(
         self, events: Iterable[str], update_metrics: bool = True
     ) -> Dict[str, _EventCacheEntry]:
-        """Fetch events from the caches
+        """Fetch events from the caches, may return rejected events.
```
richvdh, line 600:
```suggestion
        """Fetch events from the caches.
        
        May return rejected events.
```

synapse/storage/databases/main/events_worker.py
```diff
@@ -672,23 +739,22 @@ def fire(evs, exc):
                 with PreserveLoggingContext():
                     self.hs.get_reactor().callFromThread(fire, event_list, e)
 
-    async def _get_events_from_db(self, event_ids, allow_rejected=False):
-        """Fetch a bunch of events from the database.
+    async def _get_events_from_db(
+        self, event_ids: Iterable[str]
+    ) -> Dict[str, _EventCacheEntry]:
+        """Fetch a bunch of events from the database, may return rejected
+        events.
```
richvdh, line 746:
```suggestion
        """Fetch a bunch of events from the database.
        
        May return rejected events.
```

changelog.d/10119.misc
```diff
@@ -0,0 +1 @@
+Try and ensure we only have one copy of an event in memory at a time.
```
richvdh, line 1:
```suggestion
Improve event caching mechanism to avoid having multiple copies of an event in memory at a time.
```

richvdh (approved):
lgtm

synapse/storage/databases/main/events_worker.py
erikjohnston: I think the only thing that can happen is that another request waits on the deferred after its already been resolved, which shouldn't be an issue as awaiting on the deferred will just return immediately?

Issue #6485, opened 2019-12-06
Document that you should increase the max client file upload size in reverse proxy docs


anoadragon453:
We should specify the `client_max_body_size` nginx option and equivalent for other reverse proxies in https://github.com/matrix-org/synapse/blob/master/docs/reverse_proxy.md so people don't get caught up when they can't upload large images (and Synapse doesn't report an error).

Also client developers should interpret the 413 from the reverse proxy as the image is too large...

Graphicscore:
I second this, this was exactly what I needed +1

dklimpel:
This issue can be closed?
See https://github.com/matrix-org/synapse/pull/10122#issuecomment-854982727

There is a note in reverse proxy doc and [config settings](https://github.com/matrix-org/synapse/blob/develop/docs/usage/configuration/config_documentation.md#max_upload_size).

clokep:
Seems like all the relevant info is there. Thanks!

PR #10122, opened 2021-06-04
Mention that you need to configure max upload size in reverse proxy as well

aaronraimist:
### Pull Request Checklist

### DIFF
--- a/changelog.d/10122.doc
+++ b/changelog.d/10122.doc
@@ -1,1 +1,1 @@
-Mention in `homeserver.yaml` that you may need to configure max upload size in your reverse proxy. Contributed by @aaronraimist.
+Mention in the sample homeserver config that you may need to configure max upload size in your reverse proxy. Contributed by @aaronraimist.
--- a/changelog.d/10122.doc
+++ b/changelog.d/10122.doc
@@ -1,0 +1,1 @@
+Mention in `homeserver.yaml` that you may need to configure max upload size in your reverse proxy. Contributed by @aaronraimist.
--- a/synapse/config/repository.py
+++ b/synapse/config/repository.py
@@ -248,6 +248,10 @@ class ContentRepositoryConfig(Config):
 
         # The largest allowed upload size in bytes
         #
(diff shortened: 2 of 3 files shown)


changelog.d/10122.doc
```diff
@@ -0,0 +1 @@
+Mention in `homeserver.yaml` that you may need to configure max upload size in your reverse proxy. Contributed by @aaronraimist.
```
anoadragon453, line 1:
```suggestion
Mention in the sample homeserver config that you may need to configure max upload size in your reverse proxy. Contributed by @aaronraimist.
```

synapse/config/repository.py
```diff
@@ -248,6 +248,10 @@ def generate_config_section(self, data_dir_path, **kwargs):
 
         # The largest allowed upload size in bytes
         #
+        # If you are using a reverse proxy you will also need to set this value in
+        # your reverse proxy's config. See
```
anoadragon453, line 252:
```suggestion
        # If you are using a reverse proxy you may also need to set this value in
        # your reverse proxy's config. See
```

anoadragon453 (commented):
Definitely something a lot of people forget, although I realise that the reverse proxy documentation only specifies instructions on upping the max for nginx. Looking around, I think nginx may be the only webserver which actually limits upload size out of the box...?

I wonder if we should say "notably nginx" here, or something along those lines.

aaronraimist:
According to https://httpd.apache.org/docs/2.0/mod/core.html#limitrequestbody Apache seems to default to unlimited. Caddy 1 seems to default to unlimited https://caddy.its-em.ma/v1/docs/limits though I can't find any documentation confirming that is also true for Caddy 2. Not sure on any others.

aaronraimist:
If Nginx is the only one then https://github.com/matrix-org/synapse/issues/6485 can probably be closed

anoadragon453 (approved):
Thanks for checking -- the current wording sounds fine to me!

Issue #9564, opened 2021-03-08
toDevice messages on matrix.org can take 14.1s to propagate.


ara4n:
https://github.com/vector-im/element-web/issues/8376

This has been around for years, but bites me surprisingly frequently (even on test accounts!) making me assume it might be quite a widespread problem (and perhaps contribute to UISIs and encryption delays).

anoadragon453:
Note that the linked Element Web issue has Synapse logs as well. At the moment it looks to have been tracked down to the toDevice message not waking up the synchrotron for ~14s. We do at least *try* to do so:

https://github.com/matrix-org/synapse/blob/fc8b3d88097d2c985e8eae9779cdb4b23e1a8ef6/synapse/handlers/devicemessage.py#L237-L239

We also talked about increasing Jaeger's monitoring on matrix.org to try and catch the full path of the request as it goes through the system (it's currently only at 10%). We also don't have the encryption worker in Jaeger's services at the moment, which would be a good initial fix to begin with.

callahad:
From the team sync today, we think getting our Jaeger deployment into a place where it's useful for solving this could turn into a pretty significant rabbit hole, and we're at capacity for rabbit holes at the moment.

The hope is that we'll be able to come up for air in the next ~1-2 weeks and consider digging in more deeply, but in the meantime ara4n is going to do some manual digging and see if he turns up anything useful.

richvdh:
related: #9533

richvdh:
@ara4n reported another delayed message:
```
Wed May 12 16:43:29 BST 2021
1620834209.501334000
{}
->

16:43:35.609 rageshake.js:65 received to_device net.arasphere.test from: @matthewtest7:matrix.org id: undefined
```

On the encryption worker:
```
[...]

PR #10124, opened 2021-06-04
Delete delivered to-device messages earlier in /sync

richvdh:
I hope this will improve #9564.

The current impl doesn't seem optimal. Here's what happens:

 * sync arrives, we wait for a wakeup
 * we get woken up by something random
 * we run the to-device deletion.
 * we ultimately decide it's not time to return yet
 * we wait for another wakeup
 * a to-device message arrives, which wakes us up
 * we run the to-device deletion *again*.
   * There cannot be anything to delete which we haven't already deleted, but the `last_deleted_stream_id` magic in `delete_messages_for_device` is confused by the new to-device message
   * so we run a DELETE operation which will do nothing, but is surprisingly slow (something something handwave postgres deleted rows)

Hopefully, by doing the deletion right at the start, rather than after we get woken up, we can improve the latency between wakeup and response.

### DIFF
--- a/changelog.d/10124.misc
+++ b/changelog.d/10124.misc
@@ -1,0 +1,1 @@
+Work to improve the responsiveness of `/sync` requests.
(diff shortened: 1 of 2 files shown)


erikjohnston (approved):
Makes sense to me

Issue #9812, opened 2021-04-14
The interaction between `url_preview_ip_range_blacklist` and outbound HTTP proxies is confusing


anoadragon453:
Context: https://github.com/matrix-org/synapse/issues/9417#issuecomment-816262975

The Synapse config option `url_preview_ip_range_blacklist` is intended to restrict access to previewing certain URLs based on their resolved IP address. However, when using a proxy with Synapse (via the `HTTP_PROXY` and `HTTPS_PROXY` environment variables), url preview requests are forwarded to the proxy, which then resolves the IP address of the URL. This completely bypasses any IP ranges set in the `url_preview_ip_range_blacklist` option.

Ideally we'd do a couple things that would help reduce confusion for sysadmins:

* We should document in the [sample config file](https://github.com/matrix-org/synapse/blob/develop/docs/sample_config.yaml) that the value of `url_preview_ip_range_blacklist` is ignored when using an HTTP proxy, as resolving URLs to preview is done by the proxy.
* Currently it is [required](https://github.com/matrix-org/synapse/blob/04819239bae2b39ee42bfdb6f9b83c6d9fe34169/docs/sample_config.yaml#L1008-L1009) to set `url_preview_ip_range_blacklist` if URL previews are enabled. We shouldn't require `url_preview_ip_range_blacklist` to be set when using a proxy, and we should log a warning when `url_preview_ip_range_blacklist` is set and a proxy is in use so that the sysadmin isn't under the wrong impression regarding security of the system.

To be clear, if you're using a proxy for URL previews, you should do your ip blacklisting in the proxy config. Also note that `url_preview_url_blacklist` is still enforced whether a proxy is in use or not.

clokep:
We should also ensure that `ip_blacklist` says that it is ignored when a proxy is in use too. 👍 I'm not sure if it already has docs around that or not.

anoadragon453:
It does not! https://github.com/matrix-org/synapse/blob/04819239bae2b39ee42bfdb6f9b83c6d9fe34169/docs/sample_config.yaml#L174-L186

richvdh:
> We should document in the sample config file that the value of url_preview_ip_range_blacklist is ignored when using an HTTP proxy, as resolving URLs to preview is done by the proxy.

> We should also ensure that `ip_blacklist` says that it is ignored when a proxy is in use too.

I don't think these are quite accurate, are they? IIRC they are applied to the connection to the proxy. If we say that they are ignored, that is also confusing.

[...]

PR #10129, opened 2021-06-05
Add warnings to ip_range_blacklist usage with proxies

Kentokamoto:
Per issue #9812 using `url_preview_ip_range_blacklist` with a proxy via `HTTPS_PROXY` or `HTTP_PROXY` environment variables has some inconsistent bahavior than mentioned. This PR changes the following:

- Changes the Sample Config file to include a note mentioning that `url_preview_ip_range_blacklist` and `ip_range_blacklist` is ignored when using a proxy
- Changes some logic in synapse/config/repository.py to send a warning when both `*ip_range_blacklist` configs and a proxy environment variable are set and but no longer throws an error.

Signed-off-by: Kento Okamoto <kentokamoto@protonmail.com>


richvdh (changes requested):
Thank you for the contribution!

However, as I wrote on https://github.com/matrix-org/synapse/issues/9812, I'm not convinced that the comments here are quite accurate. I suggest we pursue the conversation there.

richvdh (commented):
well, apparently I was wrong about that, so ignore that!

synapse/config/repository.py
```diff
@@ -181,11 +187,17 @@ def read_config(self, config, **kwargs):
...
+                    raise ConfigError(
+                        "For security, you must specify an explicit target IP address "
+                        "blacklist in url_preview_ip_range_blacklist for url previewing "
+                        "to work"
+                    )
+                else:
+                    proxy_var = (
+                        "HTTP_PROXY" if "HTTP_PROXY" in os.environ else "HTTPS_PROXY"
+                    )
+                    logger.warning("".join(HTTP_PROXY_SET_WARNING % proxy_var))
```
richvdh, line 200: this warning will be written when a `url_preview_ip_range_blacklist` is *not* set. That seems surprising?

synapse/config/repository.py
```diff
@@ -36,6 +39,9 @@
         #    method: %(method)s
 """
 
+HTTP_PROXY_SET_WARNING = """\
+The Synapse config url_preview_ip_range_blacklist will be ignored as %s set."""
+
```
richvdh, line 44: I think you could just say:

```suggestion
HTTP_PROXY_SET_WARNING = """\
The Synapse config url_preview_ip_range_blacklist will be ignored as an HTTP(s) proxy is configured."""

```

changelog.d/10129.bugfix
```diff
@@ -0,0 +1 @@
+Fixed the interaction between `*ip_range_blacklist` setting and outbound HTTP proxies to reduce confusion.
```
richvdh, line 1:
```suggestion
Add some clarification to the sample config file. Contributed by @Kentokamoto.
```

richvdh (changes requested):

synapse/config/repository.py
Kentokamoto: Ah. Thank you for the catch. Will go ahead and edit this so the warning is written when `url_preview_ip_range_blacklist` _is_ set AND a proxy is set.  

In addition to this, would it be beneficial to have a separate info log generated when `url_preview_ip_range_blacklist` _is not_ set AND a proxy is set?

richvdh: > In addition to this, would it be beneficial to have a separate info log generated when `url_preview_ip_range_blacklist` _is not_ set AND a proxy is set?

I don't really think that is necessary?

changelog.d/10129.bugfix
```diff
@@ -0,0 +1 @@
+Add some clarification to the sample config file. Contributed by @Kentokamoto
```
erikjohnston, line 1:
```suggestion
Add some clarification to the sample config file. Contributed by @Kentokamoto.
```

synapse/config/repository.py
```diff
@@ -22,6 +23,8 @@
 
 from ._base import Config, ConfigError
 
+logger = logging.Logger(__name__)
```
richvdh, line 26: I don't know if there's an actual difference, but for consistency with the rest of the codebase:

```suggestion
logger = logging.getLogger(__name__)
```

synapse/config/repository.py
```diff
@@ -180,12 +186,17 @@ def read_config(self, config, **kwargs):
                     e.message  # noqa: B306, DependencyException.message is a property
                 )
 
+            proxies = ["HTTP_PROXY", "HTTPS_PROXY"]
```
richvdh, line 189: this feels a bit brittle. It will ignore a lower-case `http_proxy`, for example.

I suggest you import `getproxies_environment` from `urllib.request` and then check the result for `http` and `https`.

synapse/config/repository.py
```diff
@@ -180,12 +186,17 @@ def read_config(self, config, **kwargs):
...
             if "url_preview_ip_range_blacklist" not in config:
-                raise ConfigError(
-                    "For security, you must specify an explicit target IP address "
-                    "blacklist in url_preview_ip_range_blacklist for url previewing "
-                    "to work"
-                )
+                if not any(proxy in os.environ for proxy in proxies):
```
richvdh, line 191: this will suppress the warning if *either* `HTTP_PROXY` *or* `HTTPS_PROXY` is set.

I think we should show the warning unless *both* are set.

richvdh (changes requested):

synapse/config/repository.py
clokep: Ideally we should do whatever we do in the `ProxyAgent`.

richvdh: yup. `ProxyAgent` uses `getproxies_environment`.

synapse/config/repository.py
Kentokamoto: Just to be clear, this doesn't have to do with the ConfigError correct? Or did you want me to change both proxy conditional statements so it raises the error only if HTTP_PROXY AND HTTPS_PROXY are both set as well as have the logger warning display when HTTP_PROXY AND HTTPS_PROXY are both set.
So something like:
```
if "url_preview_ip_range_blacklist" not in config:
    if all(proxy not in proxy_env for proxy in proxies):
        raise ConfigError(...)
else:
    if all(proxy in proxy_env for proxy in proxies):
        logger.warning("".join(HTTP_PROXY_SET_WARNING))
```
What would the expected behavior be if either is set? Do we want to raise another error mentioning that both need to be set if they'd like to use the proxy option?

richvdh: I don't quite follow your question, but my point is this: if `HTTP_PROXY` is set, and `HTTPS_PROXY` is not set, then Synapse will make requests for `https` urls itself. So, the user should set a blacklist for security, and should see a warning if they do not set it.

richvdh: As an aside: I think the `if not any` syntax is probably overcomplicated when you only have two things to worry about. Why not just write:

```
if 'http' not in proxies or 'https' not in proxies
```

It's shorter and easier to understand, imho.

Kentokamoto: Sounds good. I'll change it to just be a boolean statement for http/https then. I wasn't sure if it would be better to have it in a list incase  list does end up growing in the future to accommodate for other proxies but I guess the change is easy enough that we'll cross that bridge if it ever were to come up.

Also just to clarify you mean a boolean AND correct not a boolean OR since you mentioned showing the warning unless both are set.

> this will suppress the warning if _either_ `HTTP_PROXY` _or_ `HTTPS_PROXY` is set.
> 
> I think we should show the warning unless _both_ are set.

richvdh: I'm pretty sure I mean a boolean OR.

We should show the warning if *any* of the following are true:
 * `HTTP_PROXY` is not set
 * `HTTPS_PROXY` is not set
 * neither is set

Do you disagree?

Kentokamoto: No I agree. Just wanted to make sure because that was my original intent with the `any` syntax.

synapse/config/repository.py
Kentokamoto: using `getproxies_environment` seems to be failing the linter with:

```
synapse/config/repository.py:19: error: Module 'urllib.request' has no attribute 'getproxies_environment'
```
Are there any additional changes I need to make to fix this? The other option is to use `getproxies` instead which appears to do the same thing.

richvdh: yeah I don't know why we use `getproxies_environment`. Suggest you just use `getproxies` (but please also update `ProxyAgent` to keep them in sync.)

clokep: See https://github.com/matrix-org/synapse/pull/9372#issuecomment-777015781 for why we use `getproxies_environment`.

I think we should be consistent (or ideally only call this once and store it on the HS object).

dklimpel: @kentokamoto are you still working on this PR?

dklimpel: > I think we should be consistent (or ideally only call this once and store it on the HS object).

@clokep we need this information for ProxyAgent and Federation. I am agree with you. In which file / config file should this?

Kentokamoto: Hi @dklimpel Yes I am still working on this. I apologize for taking so long to get this change out. I am keeping my change with `getproxies_environment` as requested for consistency but the linter is failing because it does not recognize the function name. I'm trying to avoid a change where I remove `/synapse/config/repository.py` from the mypy list.

dklimpel: Can you push your changes? Then we have a look on it. At the moment I can not see something with `getproxies_environment`.

Kentokamoto: Just pushed the changes. Should be under the hash `3de1c3050a9637c787a4e1868374c95e7024fdce`

dklimpel: Ok. I am understand the problem now.
`mypy`  does check `/synapse/config/repository.py` an we get `synapse/config/repository.py:19: error: Module 'urllib.request' has no attribute 'getproxies_environment'`
And `mypy` does **NOT** check `synapse/http/proxyagent.py`. That is the reason why this problem is new.

dklimpel: This is a solution
`from urllib.request import getproxies_environment # type: ignore`

richvdh (approved):
looks good to me. thanks very much!

reivilibre:
The complement failures might be spurious; try merging develop into this branch and they might just go away

Kentokamoto:
@reivilibre I just did a git rebase and that pulled a bunch of changes in hence the large number of commits...Hopefully that didn't mess anything up.

richvdh (changes requested):
> @reivilibre I just did a git rebase and that pulled a bunch of changes in hence the large number of commits...Hopefully that didn't mess anything up.

@Kentokamoto I'm afraid it does seem to have messed things up, because it now shows "Files changed: 438". In general, rebasing isn't what you want to do. Indeed, git even shows you this by not allowing you to push after a rebase unless you `--force`.

Please could you put this back how it was, with:

```
git checkout kokamoto/proxyIpRange
git reset --hard d05946adf3cb89b987853cb9d6d3547859c9589f
git push --force
```

and then *merge*, not rebase, from develop:

```
git merge origin/develop
git push
```

richvdh (approved):

PR #10134, opened 2021-06-07
Make opentracing trace into event persistence

richvdh:
Previously, because events were persisted by a separate background process, we wouldn't be able to see what was happening when we traced an event send request.

~~ Let's treat the persist operation as a child of the send request~~

~~There are a couple of preparatory refactors here: suggest reviewing commit-by-commit.~~ 
~~Now based on #10145.~~

Use the span references feature to link from the persist operation to the request, and back again.

### DIFF
--- a/synapse/storage/persist_events.py
+++ b/synapse/storage/persist_events.py
@@ -40,6 +40,7 @@ from twisted.internet import defer
 from synapse.api.constants import EventTypes, Membership
 from synapse.events import EventBase
 from synapse.events.snapshot import EventContext
+from synapse.logging import opentracing
 from synapse.logging.context import PreserveLoggingContext, make_deferred_yieldable
 from synapse.metrics.background_process_metrics import run_as_background_process
 from synapse.storage.databases import Databases
@@ -108,6 +109,7 @@ class _EventPersistQueueItem:
     events_and_contexts: List[Tuple[EventBase, EventContext]]
     backfilled: bool
     deferred: ObservableDeferred
+    opentracing_context: Optional[object]
 
 
 _PersistResult = TypeVar("_PersistResult")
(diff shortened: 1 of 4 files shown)


synapse/storage/persist_events.py
```diff
@@ -131,22 +169,26 @@ def add_to_queue(self, room_id, events_and_contexts, backfilled):
...
         deferred = ObservableDeferred(defer.Deferred(), consumeErrors=True)
 
+        span = opentracing.active_span()
         queue.append(
-            self._EventPersistQueueItem(
+            _EventPersistQueueItem(
                 events_and_contexts=events_and_contexts,
                 backfilled=backfilled,
                 deferred=deferred,
+                opentracing_context=span.context if span else None,
```
erikjohnston, line 178: I'm a bit concerned that this is going to cause confusion, given that we'll only be linking together the persist span with the *first* request that added events into this batch (rather than all of them). It'll likely work 99% of the time as usually there will only be events from one request in the queue at any given time.

Ideally we'd link the persist span to all the request spans, which can be done using `follows_from` refs, but a) I have a suspicion that the debug sampling does the right thing there (in that iirc it only looks at the first parent?), and b) the UI doesn't really support `follows_from` with mutliple parents :/

richvdh: oh. yes. I'd forgotten that we could extend batches.

And yes, I spent a while experimenting with multiple parents/follows_from. The UI only seems to let you go from the child to the parent, not the parent to the child. Which might be enough, I suppose, if we create a `event_persist_result` span for each request...

richvdh:
ok, here's another go at this. We now let the "child of" reference cascade as normal, and use the "follows_from" reference to link in both directions, from the persist operation to the request, and back again.

We also make sure that if any of the requests have tracing forced, we also do so in the persist operation

erikjohnston (approved):

PR #10136, opened 2021-06-07
More database opentracing

richvdh:
Add a couple of extra logs/spans, to give a bit of a better idea.

follow-up to #10113.

### DIFF
--- a/synapse/storage/database.py
+++ b/synapse/storage/database.py
@@ -541,6 +541,7 @@ class DatabasePool:
                         },
                     ):
                         r = func(cursor, *args, **kwargs)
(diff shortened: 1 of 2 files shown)


synapse/storage/database.py
```diff
@@ -541,6 +541,7 @@ def new_transaction(
                         },
                     ):
                         r = func(cursor, *args, **kwargs)
+                        opentracing.log_kv({"message": "commit"})
```
erikjohnston, line 544: Is the idea that we can now time how long the function vs commit takes? I think `start_active` span should tag the span with `error` if any of this fails?

erikjohnston (approved):

PR #10137, opened 2021-06-07
Correct type hints for parse_string(s)_from_args.

clokep:
Does a few things:

* Fixes the type hints added in #10048 which were incorrect in some situations.
* Adds a seperate `parse_bytes` method (which simplifies the above type hints dramatically).
* Enables mypy for the `servlet` file (which would have found the issues with #10048).

Each commit should be reviewable separately.

This is blocking #10080, as the type hints that are currently there conflict with that PR.

### DIFF
--- a/synapse/http/servlet.py
+++ b/synapse/http/servlet.py
@@ -253,11 +253,10 @@ def parse_strings_from_args(
         args: the twisted HTTP request.args list.
         name: the name of the query parameter.
         default: value to use if the parameter is absent, defaults to None.
-        required : whether to raise a 400 SynapseError if the
+        required: whether to raise a 400 SynapseError if the
             parameter is absent, defaults to False.
-        allowed_values (list[bytes|unicode]): List of allowed values for the
-            string, or None if any value is allowed, defaults to None. Must be
-            the same type as name, if given.
+        allowed_values: List of allowed values for the
+            string, or None if any value is allowed, defaults to None.
         encoding: The encoding to decode the string content with.
 
     Returns:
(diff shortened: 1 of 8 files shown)


synapse/http/servlet.py
```diff
@@ -108,13 +110,67 @@ def parse_boolean_from_args(args, name, default=None, required=False):
...
-    request,
-    name: Union[bytes, str],
+    request: Request,
+    name: str,
     default: Optional[str] = None,
     required: bool = False,
     allowed_values: Optional[Iterable[str]] = None,
-    encoding: Optional[str] = "ascii",
+    encoding: str = "ascii",
 ):
```
clokep, line 174: Adding a return value here cascades to a lot of changes. We should do it, but I'd prefer to not do it in this PR.

synapse/http/servlet.py
```diff
@@ -108,6 +110,60 @@ def parse_boolean_from_args(args, name, default=None, required=False):
...
+def parse_bytes(
+    request: Request,
+    name: str,
+    default: Optional[bytes] = None,
+    required: bool = False,
+) -> Optional[bytes]:
+    ...
+
+
+def parse_bytes(
```
richvdh, line 133: `parse_bytes` seems like a rubbish name to me (as does `parse_string`).

Suggestion would be that you replace it with `parse_bytes_from_args` and make it take an `args` dict (and do `request.args` in the callers).

clokep: That seems fine with me! I was mostly matching `parse_string` (which should maybe be `parse_string_from_request`, but it really is just `parse_string_from_args`...)

clokep: I pushed a commit, it does need some typing foo since `request.args` pops up as `Optional[Any]`, I think this is probably OK though?

Another option would be to do `parse_bytes_from_request` (or `parse_bytes_from_request_args`) and have it still take the `Request` object.

richvdh (approved):
that's fiddlier than I hoped, but lgtm

synapse/http/servlet.py
richvdh: yet another option might be to allow `parse_bytes_from_request` to take `None` and do something plausible?

PR #6105, opened 2019-09-25
Refactor the user-interactive auth handling

richvdh:
Pull the checkers out to their own classes, rather than having them lost in a
massive 1000-line class which does everything.

This is also preparation for some more intelligent advertising of flows, as per #6100

PR #10142, opened 2021-06-08
Implement MSC3231: Token authenticated registration

govynnus:

This is part of my GSoC project implementing [MSC3231](https://github.com/matrix-org/matrix-doc/pull/3231).

Status:
- [X] Works with a list of hard-coded tokens
- [x] Checks against tokens stored in the database
- [x] Validity checking endpoint
- [x] Fallback
- [x] Admin API for managing tokens


synapse/handlers/ui_auth/checkers.py
```diff
@@ -237,11 +237,65 @@ async def check_auth(self, authdict: dict, clientip: str) -> Any:
...
+            raise LoginError(400, "Missing token", Codes.MISSING_PARAM)
+        if not isinstance(authdict["token"], str):
+            raise LoginError(400, "Token must be a string", Codes.INVALID_PARAM)
+        if "session" not in authdict:
+            raise SynapseError(400, "Missing UIA session", Codes.MISSING_PARAM)
+        session = authdict["session"]
+
+        # If the LoginType.REGISTRATION_TOKEN stage has already been completed,
+        # return early to avoid incrementing `pending` again.
+        # TODO: check stored token matches provided token?
```
anoadragon453, line 262: I think it'd be confusing if a client presented two different tokens, both with a successful result, and upon completing registration finding that the first token was taken.

Erroring on switching tokens (or overwriting the old token, though this requires some extra book-keeping) is probably clearer from the client's perspective.

synapse/handlers/ui_auth/checkers.py
```diff
@@ -237,11 +237,65 @@ async def check_auth(self, authdict: dict, clientip: str) -> Any:
...
+        except StoreError:
+            raise SynapseError(400, "Unknown session ID: {}".format(session))
+        if stored_token:
+            return True
+
+        token = authdict["token"]
+        if await self.store.registration_token_is_valid(token):
+            # Increment pending counter, so that if token has limited uses it
+            # can't be used up by someone else in the meantime.
+            await self.store.set_registration_token_pending(token)
```
anoadragon453, line 278: We'll need to do something about expiring UIAA sessions at some point, though I don't think there's currently a method of doing so in Synapse...

synapse/storage/schema/main/delta/59/999create_registration_tokens.sql
```diff
@@ -0,0 +1,23 @@
+/* Copyright 2021 Callum Brown
```
anoadragon453, line 1: I assume the `999` in the filename is just a placeholder for now? :P

anoadragon453 (commented):
Looking excellent so far, nice work! I've dropped a myriad of comments all over the place, and I realise the PR is still in a draft state. In any case, let me know if any of them are unclear or don't make sense at this point :slightly_smiling_face:

synapse/config/registration.py
```diff
@@ -178,6 +181,10 @@ def generate_config_section(self, generate_secrets=False, **kwargs):
         #
         #enable_3pid_lookup: true
 
+        # Require users to submit a token during registration
+        #
+        #registrations_require_token: false
```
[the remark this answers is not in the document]
govynnus: Yep, good point. I'll link to the admin API docs once they exist :-)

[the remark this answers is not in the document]
govynnus: I agree. I was copying `registrations_require_3pid`, but I'll change it.

synapse/handlers/ui_auth/checkers.py
```diff
@@ -237,11 +237,65 @@ async def check_auth(self, authdict: dict, clientip: str) -> Any:
...
+        # return early to avoid incrementing `pending` again.
+        # TODO: check stored token matches provided token?
+        try:
+            stored_token = await self.store.get_ui_auth_session_data(
+                session,
+                "registration_token",
+                None,
+            )
+        except StoreError:
+            raise SynapseError(400, "Unknown session ID: {}".format(session))
```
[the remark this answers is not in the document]
govynnus: I get a circular dependency when I do that, because `synapse/handlers/auth.py` (where `AuthHandler` is defined), imports `synapse.handlers.ui_auth.checkers`. Am I missing something?

synapse/storage/databases/main/registration.py
```diff
@@ -1760,6 +1760,100 @@ def start_or_continue_validation_session_txn(txn):
...
+            return False
+        # Check if the token has been used up
+        if (
+            res["uses_allowed"]
+            and res["pending"] + res["completed"] >= res["uses_allowed"]
+        ):
+            # Don't delete the token because there could be pending uses
+            return False
+        # Otherwise, the token is valid
+        return True
```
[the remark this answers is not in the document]
govynnus: yep :grimacing:

synapse/storage/databases/main/registration.py
```diff
@@ -1760,6 +1760,100 @@ def start_or_continue_validation_session_txn(txn):
...
+            keyvalues={"token": token},
+            retcol="pending",
+            desc="set_registration_token_pending",
+        )
+        await self.db_pool.simple_update_one(
+            "registration_tokens",
+            keyvalues={"token": token},
+            updatevalues={"pending": pending + 1},
+            desc="set_registration_token_pending",
+        )
```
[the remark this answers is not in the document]
govynnus: I did wonder about that, thanks.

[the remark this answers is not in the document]
govynnus: Just checking, running in the same transaction isn't needed for `registration_token_is_valid`?

synapse/storage/databases/main/registration.py
```diff
@@ -1760,6 +1760,100 @@ def start_or_continue_validation_session_txn(txn):
...
+            return False
+        # Check if the token has expired
+        now = self._clock.time_msec()
+        if res["expiry_time"] and res["expiry_time"] < now:
+            # The token is invalid, so delete it
+            await self.db_pool.simple_delete_one(
+                "registration_tokens",
+                keyvalues={"token": token},
+                desc="registration_token_is_valid",
+            )
```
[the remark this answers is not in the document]
govynnus: That was what I thought, but then you said to delete them in the Matrix room :grin:

synapse/storage/schema/main/delta/59/999create_registration_tokens.sql
```diff
@@ -0,0 +1,23 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+CREATE TABLE IF NOT EXISTS registration_tokens(
+    token TEXT NOT NULL,  -- The token that can be used for authentication.
+    uses_allowed INT,  -- The total number of times this token can be used. NULL if no limit.
+	pending INT NOT NULL, -- The number of in progress registrations using this token.
+	completed INT NOT NULL, -- The number of times this token has been used to complete a registration.
```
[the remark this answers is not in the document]
govynnus: Oops, that was tabs and spaces problem

synapse/storage/schema/main/delta/59/999create_registration_tokens.sql
govynnus: yep :-)

tests/rest/client/v2_alpha/test_register.py
```diff
@@ -205,6 +205,248 @@ def test_POST_ratelimiting(self):
 
         self.assertEquals(channel.result["code"], b"200", channel.result)
 
+    @unittest.override_config({"registrations_require_token": True})
```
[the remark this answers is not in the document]
govynnus: Yep, I blindly copied that from below...

synapse/storage/databases/main/registration.py
```diff
@@ -1760,6 +1760,96 @@ def start_or_continue_validation_session_txn(txn):
...
+                retcols=["pending", "completed"],
+            )
+
+            # Decrement pending and increment completed
+            self.db_pool.simple_update_one_txn(
+                txn,
+                "registration_tokens",
+                keyvalues={"token": token},
+                updatevalues={
+                    "completed": res["completed"] + 1,
```
govynnus, line 1844: mypy is giving me `error: Value of type "Optional[Dict[str, Any]]" is not indexable  [index]` here and I have no idea what that means :upside_down_face: 

https://github.com/matrix-org/synapse/pull/10142/checks?check_run_id=2847878288#step:5:14

synapse/storage/databases/main/registration.py
```diff
@@ -1760,6 +1760,100 @@ def start_or_continue_validation_session_txn(txn):
...
+            updatevalues={"pending": pending + 1},
+            desc="set_registration_token_pending",
+        )
+
+    async def use_registration_token(self, token: str) -> None:
+        """Increment the completed registrations counter for a token.
+
+        If the token is now invalid, it will be deleted, otherwise the
+        pending counter will be decremented and the completed counter
+        will be incremented.
```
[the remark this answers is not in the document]
govynnus: I've stopped deleting tokens now, so the comment has changed

synapse/handlers/ui_auth/checkers.py
govynnus: I completely forgot about that...
I think UIA sessions expire after 2 days, but that might be a bit long.

https://github.com/matrix-org/synapse/blob/52c60bd0a96cd61583209a9ef6d8270425e8a902/synapse/handlers/auth.py#L187

synapse/storage/databases/main/registration.py
anoadragon453: Ah, because `res` is `Optional[Dict[str, Any]]`, which may be either `Dict[str, Any]` or `None`, and you can't do `None[...]`.

According to `simple_select_one_txn`s docstring, this is a known bug:

https://github.com/matrix-org/synapse/blob/1092718cac3800080bb766b251ae472282aef751/synapse/storage/database.py#L1711-L1714

So I would just override `res`s type to be `Dict[str, Any]` and leave a comment explaining why doing so is necessary.

```python
            res: Dict[str, Any] = self.db_pool.simple_select_one_txn(
                txn,
                "registration_tokens",
                keyvalues={"token": token},
                retcols=["pending", "completed"],
            )  # type: ignore
```

The `: Dict[str, Any]` overrides the type of `res`, and `type: ignore` tells mypy to ignore the otherwise dangerous type re-assignment.

tests/rest/client/v2_alpha/test_register.py
```diff
@@ -205,6 +205,248 @@ def test_POST_ratelimiting(self):
...
+
+        # Check auth fails when using token with session2
+        params2["auth"] = {
+            "type": LoginType.REGISTRATION_TOKEN,
+            "token": token,
+            "session": session2,
+        }
+        channel = self.make_request(b"POST", self.url, json.dumps(params2))
+        self.assertEquals(channel.result["code"], b"401", channel.result)
+        self.assertEquals(channel.json_body["errcode"], Codes.UNAUTHORIZED)
```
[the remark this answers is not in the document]
govynnus: Yeah, but shouldn't the error code check tell you it's failed?
I'll check that `completed` is empty too.

synapse/storage/databases/main/registration.py
govynnus: Ok, thanks.

synapse/rest/client/v2_alpha/register.py
```diff
@@ -669,6 +669,15 @@ async def on_POST(self, request):
         )
 
         if registered:
+            # Check if a token was used to authenticate registration
+            registration_token = await self.auth_handler.get_session_data(
+                session_id,
+                "registration_token",
```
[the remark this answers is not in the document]
govynnus: So `REGISTRATION_TOKEN = "org.matrix.msc3231.login.registration_token"`?
But then there's another circular import problem when you want to use that constant in `checkers.py`...

synapse/handlers/ui_auth/checkers.py
govynnus: Yep, I've now got it raising a `400` if the token changes. I think it's probably not worth overwriting old tokens (involving changing pending) given that clients are (hopefully) unlikely to retry a stage they have already successfully completed.

synapse/handlers/ui_auth/checkers.py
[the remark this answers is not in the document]
anoadragon453: Aha I see, that is a pain. Well, in that case it's probably fine to use the storage methods directly (it'd be tricky to get around a 2-node import cycle).

But best to leave a comment stating that a cyclic dependency prevents us from using `AuthHandler.get_session_data`, so that a later refactor doesn't fall into the same trap :)

synapse/rest/client/v2_alpha/register.py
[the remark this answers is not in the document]
anoadragon453: Aha, indeed... We can get around this though by simply doing:

```
from synapse.handlers.ui_auth import UIAuthSessionDataConstants
```

inside of `check_auth`. Another comment as to why the buried import is necessary would be good to add though.

synapse/handlers/ui_auth/checkers.py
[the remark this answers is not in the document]
anoadragon453: Actually, a buried import may work here as well (see https://github.com/matrix-org/synapse/pull/10142#discussion_r654625680). As an example elsewhere: https://github.com/matrix-org/synapse/blob/a0cd8ae8cbe14d2821cbe8fd6b011c4ddc729344/synapse/handlers/federation.py#L1531-L1532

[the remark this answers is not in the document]
govynnus: Yep, that works

synapse/rest/client/v2_alpha/register.py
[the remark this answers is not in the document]
govynnus: Done, thanks :-)

synapse/rest/client/v2_alpha/register.py
```diff
@@ -377,6 +377,33 @@ async def on_GET(self, request):
             return 200, {"available": True}
 
 
+class RegistrationTokenValidityRestServlet(RestServlet):
```
govynnus, line 380: I'm a bit confused by all the rate limiting stuff, do I need to add that here somewhere?

anoadragon453: Yep, you'll want to do so (otherwise people may be tempted to try and brute force a token).

The `Ratelimiter` class is indeed a bit complicated, but calling it should be fairly straight-forward. Since we don't have a user account to rate-limit on yet, we'll probably need to rate-limit by IP address. Thus, we just need to grab the IP address of the requester and use that as the ratelimit key:

```python
        # Somewhere in __init__()
        self._my_ratelimiter = Ratelimiter(
            store=hs.get_datastore(),
            clock=hs.get_clock(),
            rate_hz=hs.config.ratelimiting.rc_your_ratelimiter.per_second,
            burst_count=hs.config.ratelimiting.rc_your_ratelimiter.burst_count,
        )

        # In your on_GET
        await self._my_ratelimier.ratelimit(
            None, (request.getClientIP(),)
        )
```

`ratelimit` will raise a `LimitExceededError` if the limit has been reached, and will inform the client on the `ms` they need to wait before they can try again.

The other question is whether to create a new ratelimiting config option for this endpoint, or to just reuse `rc_registration`. Personally I think it's always best to separate config options out in case people want to tweak them individually (plus, it would be annoying to get your token wrong a few times before eventually coming to hit `/register` and being denied because of the previous token checks).

Ratelimiting configs are managed by the [RatelimitConfig](https://github.com/matrix-org/synapse/blob/2ca4e349e9d0c606d802ae15c06089080fa4f27e/synapse/config/ratelimiting.py#L46) class.

govynnus: Thanks!

govynnus: Ok, I've added that. Not sure if I got the defaults right.
https://github.com/matrix-org/synapse/pull/10142/commits/003e67d1ab5d64ea54f3c125ccd2b7a03e18cddc

anoadragon453: I think the current defaults are fine, and the commit looks good otherwise as well!

It's good to leave a little bit of a burst count on IP rate-limiting, in case you have a situation where a school tries to sign up a whole class at once or something.

synapse/handlers/ui_auth/checkers.py
anoadragon453: I wonder how problematic it would be to expire a UIAA stage before expiring the whole session.

We check for expired sessions every 5 minutes here:

https://github.com/matrix-org/synapse/blob/bd4919fb72b2a75f1c0a7f0c78bd619fd2ae30e8/synapse/handlers/auth.py#L264-L271

The registration flow on clients may become a bit confused if they note a completed token stage, only for it to be no longer completed later on...

Definitely not a problem we need to first in the first pass anyhow.

govynnus: Yeah, it looks a bit tricky...
I guess first thing is to make sure pending is actually decreased when the whole session expires at least

synapse/rest/client/v2_alpha/auth.py
```diff
@@ -128,9 +138,29 @@ async def on_POST(self, request, stagetype):
...
+            token = parse_string(request, "token")
+            authdict = {"session": session, "token": token}
+
+            success = await self.auth_handler.add_oob_auth(
+                LoginType.REGISTRATION_TOKEN, authdict, request.getClientIP()
+            )
+
+            if success:
+                html = self.success_template.render()
+            else:
```
govynnus, line 156: The auth checker in `synapse/handlers/ui_auth/checkers.py` either returns `True` or raises a `LoginError`, so this `else:` is never executed.

Basically if you put in an invalid token you get shown some JSON rather than a pretty message and an option to try again. This could maybe be fixed by capturing `LoginError`s in `add_oob_auth` like the normal authentication code does just above it.

The other fallbacks don't really get this problem because terms always succeeds and the recaptcha just needs to be filled out (not tested). Although if you go to https://example.org/_matrix/client/r0/auth/m.login.recaptcha/fallback/web?session=xxxxx and disable javascript, then you get a similar thing.

synapse/rest/client/v2_alpha/auth.py
```diff
@@ -74,6 +75,14 @@ async def on_GET(self, request, stagetype):
             # re-authenticate with their SSO provider.
             html = await self.auth_handler.start_sso_ui_auth(request, session)
 
+        elif stagetype == LoginType.REGISTRATION_TOKEN:
+            html = self.registration_token_template.render(
+                session=session,
+                myurl="{}/r0/auth/{}/fallback/web".format(
```
govynnus, line 81: This hasn't got an unstable prefix or anything, but I don't think it can because clients use a standard form and just insert the auth type?

anoadragon453 (commented):
Apologies for taking a while to get back to this. I left a review mid-way and only now realised.

synapse/rest/client/v2_alpha/auth.py
govynnus: That would work fine for the fallback, but then if you get an unknown token during UIAA no error messages will be returned.

The docstring for check_auth():

https://github.com/matrix-org/synapse/blob/c2c364f27f61bece85dc7fd17cdedc4b60b9f7af/synapse/handlers/ui_auth/checkers.py#L45

Although I think `SynapseError` should be `LoginError` as that is what's excepted here:

https://github.com/matrix-org/synapse/blob/c2c364f27f61bece85dc7fd17cdedc4b60b9f7af/synapse/handlers/auth.py#L600

docs/SUMMARY.md
```diff
@@ -52,6 +52,7 @@
       - [Purge History](admin_api/purge_history_api.md)
       - [Purge Rooms](admin_api/purge_room.md)
       - [Register Users](admin_api/register_api.md)
+      - [Registration Tokens](admin_api/registration_tokens.md)
```
dklimpel, line 55: I think the location of new docs should be there.

```suggestion
      - [Registration Tokens](usage/administration/admin_api/registration_tokens.md)
```

See there: https://github.com/matrix-org/synapse/blob/develop/docs/admin_api/README.rst

govynnus: Thanks

anoadragon453 (changes requested):
Broadly this is looking excellent. I've got just a few notes below, but overall this looks solid.

synapse/rest/admin/registration_tokens.py
```diff
@@ -0,0 +1,195 @@
...
+from synapse.types import JsonDict
+
+if TYPE_CHECKING:
+    from synapse.server import HomeServer
+
+logger = logging.getLogger(__name__)
+
+
+class ListRegistrationTokensRestServelet(RestServlet):
+    """List registration tokens."""
```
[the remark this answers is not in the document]
govynnus: Ok, I've copied a couple examples over from the docs, is that alright?

synapse/rest/admin/registration_tokens.py
```diff
@@ -0,0 +1,195 @@
...
+        if not isinstance(uses_allowed, (int, type(None))):
+            raise SynapseError(
+                400, "uses_allowed must be an integer or null", Codes.INVALID_PARAM
+            )
+
+        expiry_time = body.get("expiry_time", None)
+        if not isinstance(expiry_time, (int, type(None))):
+            raise SynapseError(
+                400, "expiry_time must be an integer or null", Codes.INVALID_PARAM
+            )
```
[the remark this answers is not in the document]
govynnus: For `uses_allowed` it could be useful to allow zero as a way of invalidating a token without deleting it, but otherwise yes :-)

synapse/rest/admin/registration_tokens.py
```diff
@@ -0,0 +1,195 @@
...
+            # Generate token
+            token = "".join(random.choices(self.allowed_chars, k=length))
+
+        uses_allowed = body.get("uses_allowed", None)
+        if not isinstance(uses_allowed, (int, type(None))):
+            raise SynapseError(
+                400, "uses_allowed must be an integer or null", Codes.INVALID_PARAM
+            )
+
+        expiry_time = body.get("expiry_time", None)
```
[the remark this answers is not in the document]
govynnus: Yep I've added some extra checks and tests for them

tests/rest/admin/test_registration_tokens.py
```diff
@@ -0,0 +1,517 @@
...
+                "registration_tokens",
+                {
+                    "token": "invalid_2",
+                    "uses_allowed": 2,
+                    "pending": 1,
+                    "completed": 1,
+                    "expiry_time": now + 1000000,
+                },
+            )
+        )
```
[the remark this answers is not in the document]
govynnus: Just want to check what I've done here is acceptable?

synapse/storage/databases/main/registration.py
```diff
@@ -1854,6 +1854,166 @@ def _use_registration_token_txn(txn):
...
+        Returns:
+            A dict, or None if token doesn't exist.
+        """
+        return await self.db_pool.simple_select_one(
+            "registration_tokens",
+            keyvalues={"token": token},
+            retcols=["token", "uses_allowed", "pending", "completed", "expiry_time"],
+            allow_none=True,
+            desc="get_one_registration_token",
+        )
```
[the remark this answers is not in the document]
govynnus: I don't really understand...

tests/rest/client/v2_alpha/test_register.py
```diff
@@ -435,6 +436,91 @@ def test_POST_registration_token_expiry(self):
...
+        # Check pending is now 0
+        pending = self.get_success(
+            store.db_pool.simple_select_one_onecol(
+                "registration_tokens",
+                keyvalues={"token": token},
+                retcol="pending",
+            )
+        )
+        self.assertEquals(pending, 0)
+
```
anoadragon453, line 523: Your code handles it fine, but it may be worth adding a test for:

1. Start but never complete a session.
2. Delete the token from the db.
3. Expire the session.

Just as that may be an edge case that's easy to miss and could break in the future.

synapse/storage/databases/main/registration.py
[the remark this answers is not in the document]
anoadragon453: So we're returning a dict here, and if we wanted to access a value in this dict later then we'd need to do `returned_dict["value_name"]`. We don't currently as we're just sending this dict over the wire (up to where I've reviewed), but if we start using this method in other bits of the codebase it can be a good idea to make an object type that encapsulates the data instead. Both for convenience (auto-completion, type-checking) and flexibility (attaching methods).

And if you set `__slots__` for the object, you'll have essentially no performance difference vs. just the bare dictionary.

tests/rest/admin/test_registration_tokens.py
```diff
@@ -0,0 +1,679 @@
...
+        channel = self.make_request(
+            "GET",
+            self.url + "?valid=x",
+            {},
+            access_token=self.admin_user_tok,
+        )
+
+        self.assertEqual(400, int(channel.result["code"]), msg=channel.result["body"])
+
+    def _test_list_query_parameter(self, valid):
```
anoadragon453, line 636:
```suggestion
    def _test_list_query_parameter(self, valid: str):
```

especially as one might look at the docstring and think to use a bool accidentally.

tests/rest/admin/test_registration_tokens.py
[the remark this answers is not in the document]
anoadragon453: I'd say so, thanks!

synapse/storage/databases/main/registration.py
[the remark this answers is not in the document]
govynnus: Ah ok. I think I'll leave it for now, but if I add the token's validity as a new field at some point a class would be good.

tests/rest/client/v2_alpha/test_register.py
govynnus: Have done

synapse/handlers/ui_auth/checkers.py
govynnus: Pending is decreased when uncompleted sessions expire in e7495e61a9a50ee49ac9e8259094444b1bf05b55, but that does take 2 days.

synapse/rest/admin/registration_tokens.py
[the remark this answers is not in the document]
anoadragon453: Yep, that's fine!

synapse/rest/admin/registration_tokens.py
```diff
@@ -0,0 +1,195 @@
...
+        if token_info is None:
+            raise NotFoundError("No such registration token: {}".format(token))
+
+        return 200, token_info
+
+    async def on_PUT(self, request: SynapseRequest, token: str) -> Tuple[int, JsonDict]:
+        """Update a registration token."""
+        await assert_requester_is_admin(self.auth, request)
+        body = parse_json_object_from_request(request)
+        updatevalues = {}
```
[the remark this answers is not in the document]
anoadragon453: Ah, this is just because we're passing to the `updatevalues` arg eventually.

Still, I'd suggest using something like `token_attributes` which may be a bit more descriptive.

synapse/rest/admin/registration_tokens.py
```diff
@@ -0,0 +1,195 @@
...
+
+        if "expiry_time" in body:
+            if not isinstance(body["expiry_time"], (int, type(None))):
+                raise SynapseError(
+                    400, "expiry_time must be an integer or null", Codes.INVALID_PARAM
+                )
+            updatevalues["expiry_time"] = body["expiry_time"]
+
+        if len(updatevalues) == 0:
+            raise SynapseError(400, "Nothing to update", Codes.MISSING_PARAM)
```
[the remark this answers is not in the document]
govynnus: Makes sense

synapse/rest/client/v2_alpha/auth.py
anoadragon453: > Although I think `SynapseError` should be `LoginError` as that is what's excepted here.

I think you're right. The [PR that docstring was introduced with](https://github.com/matrix-org/synapse/pull/6105/files) also extracts the each UIAA checker into a separate class, where `check_auth` routinely raises `LoginError`.

I think I understand then. So the problem is that when calling `AuthHandler.check_ui_auth`, which in turn calls `RegistrationTokenAuthChecker.check_auth` during UIAA, we expect a `LoginError` if an invalid token was passed.

Conversely, when we hit `POST /_matrix/client/r0/auth/org.matrix.msc3231.login.registration_token/fallback/web` with an invalid token, we call `AuthHandler.add_oob_auth`, which in turn calls `RegistrationTokenAuthChecker.check_auth`. The call in `add_oob_auth` is  expecting either *something* or `None` from `check_auth`. Taking a look, it doesn't look like any auth checkers ever return `None` - so perhaps this is an invalid assumption, and `add_oob_auth` should be refactored to either return a result or raise `LoginError` / `SynapseError`, rather than `True` / `False`.

Then we can just wrap calls to `add_oob_auth` in `AuthRestServlet` in a try/catch for `LoginError`s that have a `401` error code. If so, return the respective template to the client indicating that the auth credentials provided (i.e a token) was invalid. Otherwise, just let the error get raised - and thus the client will receive JSON (which is expected as the request was malformed).

As an aside; it doesn't appear that [the spec](https://spec.matrix.org/unstable/client-server-api/#fallback) specifies what `POST /_matrix/client/r0/auth/<auth type>/fallback/web` needs to return - or even that that endpoint needs to be called. It looks to me as if Synapse just chose that as a sensible path - though it probably shouldn't as the `POST` variant isn't specced.

govynnus: Should I fix that up in a separate PR then, or just bundle it in here?

I think that POST endpoint is just used by synapse internally when submitting UIA creds from the web fallback page.

anoadragon453: > Should I fix that up in a separate PR then, or just bundle it in here?

A separate PR sounds good as it seems like a reasonably encapsulated bit of work. Simply merge `develop` into this branch once that PR lands.

> I think that POST endpoint is just used by synapse internally when submitting UIA creds from the web fallback page.

Yeah, it should just live on the `/_synapse/client` resource rather than `/_matrix/client` is all.

govynnus: > A separate PR sounds good as it seems like a reasonably encapsulated bit of work. Simply merge `develop` into this branch once that PR lands.

How does #10561 look?

> Yeah, it should just live on the `/_synapse/client` resource rather than `/_matrix/client` is all.

Oh ok

synapse/rest/client/v2_alpha/register.py
```diff
@@ -377,6 +378,41 @@ async def on_GET(self, request):
...
+    PATTERNS = client_patterns(
+        f"/org.matrix.msc3231/register/{LoginType.REGISTRATION_TOKEN}/validity",
+        unstable=True,
+    )
+
+    def __init__(self, hs):
```
anoadragon453, line 387: Also, could you add a docstring with an example request/response to this servlet as well?

synapse/storage/databases/main/registration.py
```diff
@@ -1760,6 +1760,260 @@ def start_or_continue_validation_session_txn(txn):
             start_or_continue_validation_session_txn,
         )
 
+    async def registration_token_is_valid(self, token: str) -> bool:
```
anoadragon453, line 1763: Note that these methods being in `RegistrationStore` rather than `RegistrationWorkerStore` will mean that worker processes cannot call them. Since `^/_matrix/client/(r0|unstable)/register$` is already supported by workers - we should make sure that this change doesn't drop that support.

All that's necessary should be just moving these methods into `RegistrationWorkerStore` instead - which worker processes make use of, rather than `RegistrationStore` (which inherits from `RegistrationWorkerStore` and is used by the main process only).

anoadragon453 (changes requested):
Some small comments, but overall looking rather solid!

Did you want to add Complement tests as a stretch goal btw?

synapse/handlers/ui_auth/checkers.py
govynnus: Ok, I'll leave it like it is then and if it turns out to be a problem I can try figure something else out.

synapse/rest/client/v2_alpha/register.py
govynnus: Sure

synapse/storage/databases/main/registration.py
govynnus: Ok, I've moved them but haven't tested with workers at all

anoadragon453: That's alright. That would be done in Complement CI sometime in the future when Complement worker infrastructure is up and running.

synapse/storage/databases/main/registration.py
```diff
@@ -1343,18 +1343,22 @@ async def create_registration_token(
...
+                values={
+                    "token": token,
+                    "uses_allowed": uses_allowed,
+                    "pending": 0,
+                    "completed": 0,
+                    "expiry_time": expiry_time,
+                },
+                desc="create_registration_token",
+            )
+        except self.database_engine.module.IntegrityError:
```
anoadragon453, line 1358: Note that raising integrity errors will still appear in postgres' logs, so it's best to avoid them where we can. See https://github.com/matrix-org/synapse/pull/10442.

While it would cost another database query, it may be a good idea to check for existing tokens with a `simple_select` first instead.

anoadragon453 (changes requested):
Sorry, note that `63` is actually the latest schema delta, rather than 62. It was raised in [this commit](https://github.com/matrix-org/synapse/commit/84469bdac773ddb79cfc99f31bbac78d27450682#diff-11187aa9d2e4ce23bd5c39e3e7f7f04fd93989a1c804edd249cf554de41bdf84R15-R32). It didn't involve a migration though, which could be confusing if one is just looking at the list of schema delta directories.

I'll need to take another overarching look at this tomorrow - but it's looking very close! I'd also at least try using it from a [worker setup manually](https://github.com/matrix-org/synapse/blob/master/docs/workers.md), by creating a generic worker alongside the main process that handles client traffic. Then, attempt to register through it using a token.

I need to try it again, but during my last review I was unable to call the verify endpoint on the worker (I got back an "Unrecognised request" error). So it looks like the endpoint wasn't mounted on the worker, but didn't look at deeper than that yet.

synapse/rest/admin/registration_tokens.py
```diff
@@ -0,0 +1,309 @@
...
+                )
+            if not (0 < length <= 64):
+                raise SynapseError(
+                    400,
+                    "length must be greater than zero and not greater than 64",
+                    Codes.INVALID_PARAM,
+                )
+
+            # Generate token
+            token = "".join(random.choices(self.allowed_chars, k=length))
```
anoadragon453, line 154: While not likely, it is possible for this to generate a token that already exists. The user will then be presented with:

```
{
  "errcode": "M_INVALID_PARAM",
  "error": "Token already exists: L"
}
```

later on when we call `create_registration_token`. Perhaps we should check whether the token already exists, and try generating a new one if so, before continuing on?

We can bail out with an exception after a few unsuccessful tries.

synapse/rest/client/register.py
```diff
@@ -379,6 +380,55 @@ async def on_GET(self, request):
             return 200, {"available": True}
 
 
+class RegistrationTokenValidityRestServlet(RestServlet):
+    """Check the validity of a registration token.
```
anoadragon453, line 384: I've realised that to allow this endpoint to be mounted on a client worker, you simply need to register it here:

https://github.com/matrix-org/synapse/blob/84469bdac773ddb79cfc99f31bbac78d27450682/synapse/app/generic_worker.py#L282-L285

Let's do that (and update the worker docs) afterwards in a separate PR though, in order to land this one quicker. :slightly_smiling_face:

synapse/handlers/ui_auth/__init__.py
```diff
@@ -34,3 +34,8 @@ class UIAuthSessionDataConstants:
     # used by validate_user_via_ui_auth to store the mxid of the user we are validating
     # for.
     REQUEST_USER_ID = "request_user_id"
+
+    # used during registration to store the registration used (if required) so that
```
anoadragon453, line 38:
```suggestion
    # used during registration to store the registration token used (if required) so that:
```

synapse/handlers/ui_auth/checkers.py
```diff
@@ -241,11 +241,76 @@ async def check_auth(self, authdict: dict, clientip: str) -> Any:
...
+                400, "Registration token must be a string", Codes.INVALID_PARAM
+            )
+        if "session" not in authdict:
+            raise LoginError(400, "Missing UIA session", Codes.MISSING_PARAM)
+
+        # Import here to avoid a cyclic dependency
+        from synapse.handlers.ui_auth import UIAuthSessionDataConstants
+
+        # Retrieve the auth handler here to avoid a cyclic dependency
+        auth_handler = self.hs.get_auth_handler()
```
anoadragon453, line 270: We can probably combine these two comments.

synapse/rest/admin/registration_tokens.py
```diff
@@ -0,0 +1,309 @@
...
+                    "uses_allowed": null,
+                    "pending": 0,
+                    "completed": 9,
+                    "expiry_time": 1625394937000
+                }
+            ]
+        }
+
+    The optional query parameter `valid` is used to the response to only
+    valid or invalid tokens. It can be `true` or `false`.
```
anoadragon453, line 65:
```suggestion
    The optional query parameter `valid` can used to filter the response to contain only
    valid or invalid tokens. It can be `true` or `false`.
```

anoadragon453: What is defined as an invalid token here? Perhaps instead we should say "tokens that have expired or exhausted all uses"?

synapse/rest/admin/registration_tokens.py
```diff
@@ -0,0 +1,309 @@
...
+
+    async def on_PUT(self, request: SynapseRequest, token: str) -> Tuple[int, JsonDict]:
+        """Update a registration token."""
+        await assert_requester_is_admin(self.auth, request)
+        body = parse_json_object_from_request(request)
+        new_attributes = {}
+
+        # Only add uses_allowed to new_attributes if it is present and valid
+        if "uses_allowed" in body:
+            ua = body["uses_allowed"]
```
anoadragon453, line 265: It's usually clearer to write out words in variable names rather than abbreviating them. Same for `et` below.

anoadragon453 (changes requested):
A few remaining bits and pieces, but otherwise this is looking ready to go!

synapse/storage/databases/main/registration.py
govynnus: Done in 01a74da6f7b42a1b146210c51aa0928602d4e85f

synapse/rest/admin/registration_tokens.py
```diff
@@ -0,0 +1,309 @@
...
+            if not (ua is None or (isinstance(ua, int) and ua >= 0)):
+                raise SynapseError(
+                    400,
+                    "uses_allowed must be a non-negative integer or null",
```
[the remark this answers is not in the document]
govynnus: I think it could be useful to let uses_allowed be zero as a way of making the token invalid without deleting it.

docs/usage/administration/admin_api/registration_tokens.md
```diff
@@ -0,0 +1,292 @@
...
+- `pending`: The number of pending uses the token has. When someone uses
+  the token to authenticate themselves, the pending counter is incremented
+  so that the token is not used more than the permitted number of times.
+  When the person completes registration the pending counter is decremented,
+  and the completed counter is incremented.
+- `completed`: The number of times the token has been used to successfully
+  complete a registration.
+- `expiry_time`: The latest time the token is valid. Given as the number of
+  milliseconds since 1970-01-01 00:00:00 UTC (the start of the Unix epoch).
+  A more human friendly format will be provided at some point, but in the
```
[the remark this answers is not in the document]
govynnus: Makes sense. I started adding support in synadm actually, so I'll try something there.

synapse/rest/admin/registration_tokens.py
govynnus: Yep, I've made that clearer in the comment and the docs

synapse/rest/admin/registration_tokens.py
govynnus: How does b5608c37944ef69d2a63288a97718b72880206a7 look?
I've tested it manually, but not sure how a unit test for that case would work.

synapse/storage/databases/main/registration.py
govynnus: Tested with a generic worker and it works :-)

synapse/rest/client/register.py
govynnus: Done in bf28876f6764e9ac820ae61cf0f7436b413cb19c

synapse/rest/admin/registration_tokens.py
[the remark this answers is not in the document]
anoadragon453: Oh, great idea actually. I would probably start with throwing that as an idea into the MSC, and then modify the API shape in a separate PR?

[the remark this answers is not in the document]
govynnus: Well token management is up to the server implementation so it doesn't matter to the MSC?
The admin API currently does allow uses_allowed to be zero, that's why the message says non-negative rather than positive :D

synapse/rest/admin/registration_tokens.py
anoadragon453: Thanks - I left some comments.

As for a unit test, I would probably just mock the suggested `generate_registration_token` to return `None`.

docs/usage/administration/admin_api/registration_tokens.md
```diff
@@ -23,9 +24,6 @@ These objects have the following fields:
   complete a registration.
 - `expiry_time`: The latest time the token is valid. Given as the number of
   milliseconds since 1970-01-01 00:00:00 UTC (the start of the Unix epoch).
-  A more human friendly format will be provided at some point, but in the
-  meantime you can remove the milliseconds and use the `date` command.
-  For example, `date -d '@1625394937'`.
```
anoadragon453: Oh, I was just commenting on the "A more human friendly format will be provided". I would still keep the `date` example though. That was handy!  :smiley:

synapse/rest/admin/registration_tokens.py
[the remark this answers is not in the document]
anoadragon453: Oh! Sorry, I got confused with the `if not`. This is currently only disallowing `<`, rather than `<=`.

In that case - cool! I suppose we should just drop a line in the documentation to suggest disabling tokens in that manner? Otherwise no more qualms from me here!

anoadragon453 (commented):
lgtm other than the outstanding comments!

synapse/rest/admin/registration_tokens.py
govynnus: Ok. I think 54867efb3a404042140f88bbba1c73055d4787b8 looks better?

synapse/storage/databases/main/registration.py
```diff
@@ -1326,6 +1326,47 @@ async def get_one_registration_token(self, token: str) -> Optional[Dict[str, Any
...
+                desc="check_if_registration_token_exists",
+            )
+
+            if existing_token is None:
+                # The generated token doesn't exist yet, return it
+                return token
+
+        raise SynapseError(
+            500,
+            "Unable to generate a registration token. Try again with a greater length",
```
anoadragon453, line 1366:
```suggestion
            "Unable to generate a unique registration token. Try again with a greater length",
```

anoadragon453 (approved):
Nice method of testing token generation collision.

This LGTM!

PR #10148, opened 2021-06-09
Document how to see logger output when running the twisted tests

MadLittleMods:
Document how to see logger output when running the twisted tests

As shared by @erikjohnston ❤
https://github.com/matrix-org/synapse/pull/9247#discussion_r647237778

### Pull Request Checklist


changelog.d/10148.doc
```diff
@@ -0,0 +1 @@
+Document `SYNAPSE_TEST_LOG_LEVEL` to see the logger output when running tests.
```
richvdh, line 1: this is probably more of a .misc thing since it's not of interest to users.

richvdh (commented):
Running the tests is also covered in CONTRIBUTING.md, where I think this fits more naturally. Possibly put the extra info there with a "see [contributing] for more info" link?

README.rst
```diff
@@ -324,6 +312,23 @@ If you just want to start a single instance of the app and run it directly::
...
+
+This should end with a 'PASSED' result (note that exact numbers will
+differ)::
+
+    Ran 1337 tests in 716.064s
+
+    PASSED (skips=15, successes=1322)
+
+For more tips on running the unit tests, like running a specific test or
+show the logger output of the app, see the `CONTRIBUTING doc <CONTRIBUTING.md#run-the-unit-tests>`_.
```
MadLittleMods, line 331: > Running the tests is also covered in CONTRIBUTING.md, where I think this fits more naturally. Possibly put the extra info there with a "see [contributing] for more info" link?

For reference, I never checked out the contributing doc for tests because the readme already had content in it. I was only linked to it from the merge request template for changelog and linting stuff.

Hopefully the "see contributing doc" link is obvious enough. Good as a first iteration to have this documented somewhere ⏩

CONTRIBUTING.md
```diff
@@ -163,6 +163,8 @@ was broken. They are slower than the linters but will typically catch more error
 ```sh
 source ./env/bin/activate
 trial tests
+# The tests can also be run with this longer command
+python -m twisted.trial tests
```
MadLittleMods, line 167: I think it's good to call this out because the readme uses this syntax and people might be Ctrl+F finding where to jump off from.

richvdh: I see your point, though I'm bound to say that I think this just makes it confusing for readers of CONTRIBUTING.md.

A better solution would be to use one command or the other consistently. (I'd suggest `trial` since it's shorter, but ymmv)

CONTRIBUTING.md
```diff
@@ -179,6 +181,14 @@ If your tests fail, you may wish to look at the logs:
 less _trial_temp/test.log
 ```
 
+To see always see logger output from the app while the tests run,
+set `SYNAPSE_TEST_LOG_LEVEL` and the resulting logs will be in `_trial_temp/test.log`:
```
richvdh, line 185: This sentence doesn't make any sense. (Also, we've just mentioned `_trial_temp` - do we really need it again? Leave it in if you prefer, though).

```suggestion
To increase the log level for the tests,  set `SYNAPSE_TEST_LOG_LEVEL`:
```

README.rst
richvdh, line 331:
```suggestion
to see the logging output, see the `CONTRIBUTING doc <CONTRIBUTING.md#run-the-unit-tests>`_.
```

richvdh: :+1:

richvdh (commented):
fyi, `SYNAPSE_TEST_LOG_LEVEL` defaults to `ERROR`. So it's not that there won't be any logs at all, just that most of it will be dropped due to the log level.

CONTRIBUTING.md
MadLittleMods: I've updated the readme to use `trial tests` 🤷

MadLittleMods:
> fyi, `SYNAPSE_TEST_LOG_LEVEL` defaults to `ERROR`. So it's not that there won't be any logs at all, just that most of it will be dropped due to the log level.

Ahh, that makes more sense. I've added that detail to the docs as well :+1:

MadLittleMods:
Thanks for the review pass @richvdh! Updated with your suggestions :+1:

CONTRIBUTING.md
```diff
@@ -173,12 +173,19 @@ source ./env/bin/activate
 trial tests.rest.admin.test_room tests.handlers.test_admin.ExfiltrateData.test_invite
 ```
 
-If your tests fail, you may wish to look at the logs:
+If your tests fail, you may wish to look at the logs (the default log level is `ERROR`):
 
 ```sh
 less _trial_temp/test.log
```
MadLittleMods, line 179: It would be interesting to add a note to the `trial tests` output when the tests fail to go checkout the `_trial_temp/test.log` file. I wonder if we can hook in at all 🤔

MadLittleMods: It looks possible if we defined our own reporter but then that requires adding `--reporter=ourcustomreporter` to the test command.

Anyway, another possible PR ⏩

richvdh (approved):
looks great now, thanks for your patience!

MadLittleMods:
Thanks for the review and merge @richvdh 🐓

PR #10149, opened 2021-06-09
Fix bug when running presence off master

### LINKED ISSUE #10027: Own presence shown as offline after a while with presence writer worker enabled
### Description

We're trialing presence workers on a few EMS hosts and for a while the new presence writer worker worked fine. After a few days, both of the hosts stopped reporting presence online for the user themselves. A restart of Synapse made presence show correctly for self again.

### Version information

- **Version**: v1.34.0
- **Platform**: EMS/Docker

Please ping for host names for logs access.

erikjohnston:
Hopefully fixes #10027.

### DIFF
--- a/changelog.d/10149.bugfix
+++ b/changelog.d/10149.bugfix
@@ -1,1 +1,1 @@
-Fix bug where presence updates would stop working when using a presence writer worker.
+Fix a bug which caused  presence updates to stop working some time after restart, when using a presence writer worker.
--- a/changelog.d/10149.bugfix
+++ b/changelog.d/10149.bugfix
@@ -1,0 +1,1 @@
+Fix bug where presence updates would stop working when using a presence writer worker.
--- a/synapse/storage/databases/main/presence.py
+++ b/synapse/storage/databases/main/presence.py
@@ -50,7 +50,7 @@ class PresenceStore(SQLBaseStore):
                 instance_name=self._instance_name,
                 tables=[("presence_stream", "instance_name", "stream_id")],
                 sequence_name="presence_stream_sequence",
-                writers=hs.config.worker.writers.to_device,
+                writers=hs.config.worker.writers.presence,
             )
         else:
             self._presence_id_gen = StreamIdGenerator(
--- a/synapse/storage/util/id_generators.py
+++ b/synapse/storage/util/id_generators.py
@@ -397,7 +397,7 @@ class MultiWriterIdGenerator:
                 # ... persist event ...
         """
 
-        # If we have a list of instance that are allowed to write to this
+        # If we have a list of instances that are allowed to write to this
         # stream, make sure we're in it.
         if self._writers and self._instance_name not in self._writers:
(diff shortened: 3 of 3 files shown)


callahad (commented):
Seems reasonable. Two completely subjective questions:

- How do we feel about `assert`, versus code which is guaranteed to execute at runtime?

- Given the duplication, would it make sense to throw the assertion in a decorator and slap that on the `get_next` functions?

anoadragon453:
> How do we feel about assert, versus code which is guaranteed to execute at runtime?

I've always been a bit wary about it, as its possible to strip `assert`s out when running Python with `-O` and `-OO`, which some people may be doing. Alternatively, I think using them only in the context of when a programming error was made (assert that a str was passed to this function) is fine.

I'm not sure if this change falls into the latter bracket though, so perhaps raising an Exception would be better?

synapse/storage/util/id_generators.py
```diff
@@ -397,6 +397,11 @@ def get_next(self):
                 # ... persist event ...
         """
 
+        # If we have a list of instance that are allowed to write to this
```
richvdh, line 400:
```suggestion
        # If we have a list of instances that are allowed to write to this
```

synapse/storage/util/id_generators.py
```diff
@@ -406,6 +411,11 @@ def get_next_mult(self, n: int):
                 # ... persist events ...
         """
 
+        # If we have a list of instance that are allowed to write to this
```
richvdh, line 414:
```suggestion
        # If we have a list of instances that are allowed to write to this
```

synapse/storage/util/id_generators.py
```diff
@@ -416,6 +426,11 @@ def get_next_txn(self, txn: LoggingTransaction):
             # ... persist event ...
         """
 
+        # If we have a list of instance that are allowed to write to this
```
richvdh, line 429:
```suggestion
        # If we have a list of instances that are allowed to write to this
```

synapse/storage/databases/main/presence.py
```diff
@@ -50,7 +50,7 @@ def __init__(
                 instance_name=self._instance_name,
                 tables=[("presence_stream", "instance_name", "stream_id")],
                 sequence_name="presence_stream_sequence",
-                writers=hs.config.worker.writers.to_device,
+                writers=hs.config.worker.writers.presence,
```
richvdh, line 53: hawkward!

changelog.d/10149.bugfix
```diff
@@ -0,0 +1 @@
+Fix bug where presence updates would stop working when using a presence writer worker.
```
richvdh, line 1:
```suggestion
Fix a bug which caused  presence updates to stop working some time after restart, when using a presence writer worker.
```

richvdh (approved):
lgtm

synapse/storage/databases/main/presence.py
erikjohnston: inorite

PR #10150, opened 2021-06-09
Avoid sending lots of federation transactions all at once due to presence.

erikjohnston:
One of the really annoying things with presence is that every time your state changes it will send a transaction to all servers you share a room with all at once. This can cause huge spikes in CPU (and I think memory), so ideally we'd avoid that.

This PR adds a random delay between 0 and 5s before poking the PerDestinationQueue to start sending transactions. The idea here is that a) a 5s delay to presence is fine most of the time, and b) we mainly care about presence being fast when a user is already interacting with a room, e.g. they've started typing, at which point we'll already be sending transactions to those servers which will pull in the relevant presence updates.

The PR needs clean up, but I mostly want to check whether this seems like a sane thing to do? It has made things noticeably better on jki.re.

### DIFF
--- a/synapse/federation/sender/per_destination_queue.py
+++ b/synapse/federation/sender/per_destination_queue.py
@@ -32,7 +32,9 @@ from synapse.handlers.presence import format_user_presence_state
 from synapse.logging import issue9533_logger
 from synapse.logging.opentracing import SynapseTags, set_tag
 from synapse.metrics import sent_transactions_counter
-from synapse.metrics.background_process_metrics import run_as_background_process
+from synapse.metrics.background_process_metrics import (
+    wrap_as_background_process,
+)
 from synapse.types import ReadReceipt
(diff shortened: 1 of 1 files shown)


richvdh:
In principle I think that delaying presence updates is fine, and that batching them up is a good idea.

The idea of doing it randomly makes me feel a bit funny, but I guess it's quick, easy and effective.

anoadragon453:
I assumed we'd be removing the ability for presence changes to wake up the destination queue entirely, but I suppose that putting a cap of 5s on it is a good idea to prevent stale presence from being sent out. The randomness factor ensure that we don't just end up with the same problem 5s down the line. So this seems like a reasonable approach - interested to see how much it helps on large deployments.

erikjohnston:
Thanks for the responses! I need to rework this a bit, so closing it for now

Issue #10128, opened 2021-06-04
synapse/app/admin_cmd.py still using removed storage.presence


callahad:
synapse/app/admin_cmd.py still using removed storage.presence:
`from synapse.replication.slave.storage.presence import SlavedPresenceStore`

_Originally posted by @vitlav in https://github.com/matrix-org/synapse/issues/9820#issuecomment-854150858_

anoadragon453:
This file is used for the exfiltration of historical data in case of GDPR requests. Definitely something we'd like to have working when required, and we should probably fix it for the next release.

PR #10154, opened 2021-06-09
Clean up a broken import in admin_cmd.py

anoadragon453:
Fixes: https://github.com/matrix-org/synapse/issues/10128

This PR fixes an import in `synapse/app/admin_cmd.py` that broke in #9820. Interestingly none of our linting caught this, even though our linting tools pass over this file. I'm not quite sure why, but my thought is that as this file is technically a new entrypoint to the codebase, `flake8` (which would be the tool to catch this), isn't able to check `synapse.replication.slave.storage.presence.SlavedPresenceStore` actually exists.

While this PR does remove the import, nowhere do I see in `admin_cmd.py` nor in `hs.get_admin_handler().export_user_data()`, which this function calls, the usage of any methods from `SlavedPresenceStore`, now `PresenceStore`. So I'm not sure if it's necessary to replace the import with anything.

I'm also not really sure how to use this tool in general. I got as far as running `python -m synapse.app.admin_cmd export-data "@userid:localhost"`, though it seems that this spins up a worker and requires a worker config file to be available.

We also don't use this tool internally either at matrix.org, so it's also worth asking whether anyone actually uses this tool...

### DIFF
--- a/changelog.d/10154.bugfix
+++ b/changelog.d/10154.bugfix
@@ -1,0 +1,1 @@
+Remove a broken import line in Synapse's admin_cmd worker. Broke in 1.33.0.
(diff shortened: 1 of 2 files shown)


richvdh (approved):
:man_shrugging:

PR #10155, opened 2021-06-09
Update Complement run with Synapse-supported MSC-related build tags

anoadragon453:
This PR updates the build tags that we perform Complement runs with to match our [buildkite pipeline](https://github.com/matrix-org/pipelines/blob/618b3e90bcae8efd1a71502ae95b7913e6e24665/synapse/pipeline.yml#L570), as well as adding `msc2403` (as it will be required once #9359 is merged). Build tags are what we use to determine which tests to run in Complement (really it determines which test files are compiled into the final binary).

I haven't put in a comment about updating the buildkite side here, as we've decided to migrate fully to GitHub Actions anyhow.

Requires: https://github.com/matrix-org/synapse/pull/9359

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -322,7 +322,7 @@ jobs:
         working-directory: complement/dockerfiles
 
       # Run Complement
-      - run: go test -v -tags synapse_blacklist ./tests
+      - run: go test -v -tags synapse_blacklist,msc2946,msc3083,msc2403 ./tests
         env:
           COMPLEMENT_BASE_IMAGE: complement-synapse:latest
         working-directory: complement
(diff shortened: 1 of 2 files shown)


.github/workflows/tests.yml
```diff
@@ -322,7 +322,7 @@ jobs:
         working-directory: complement/dockerfiles
 
       # Run Complement
-      - run: go test -v -tags synapse_blacklist ./tests
+      - run: go test -v -tags synapse_blacklist,msc2946,msc3083,msc2403 ./tests
```
clokep, line 325: For easier scanning of the list:

```suggestion
      - run: go test -v -tags synapse_blacklist,msc2403,msc2946,msc3083 ./tests
```

clokep (approved):
Should be OK if CI passes.

Issue #10066, opened 2021-05-25
Federation race condition when joining and sending a message from bridge appservice


MadLittleMods:
### Description

The Gitter bridge seems to have reproduced a race condition in federation where because the appservice joins and sends a message in quick succession, the message can out run the `join` which causes the message to `soft_fail`. 

To see an actual reproduction, there is a missing message in the `!SrkiFczPSWmYrlSNYF:matrix.org` room for `matrix.org` homeserver users (`$mMJdjGPC2GaoWK3pLApa_O2vBoeuQwYtP6iY6XzbrfY`). This is tracked on Gitter by https://gitlab.com/gitterHQ/webapp/-/issues/2770#note_583925093

The event exists in the `gitter.im` homeserver database and is visible to the `gitter.im` homeserver where the Gitter bridge appservice operates. The message is also visible to any new homeserver that comes in the room and backfills the messages. It's only not visible on the `matrix.org` homeserver.

![](https://user-images.githubusercontent.com/558581/119552215-144c8000-bd60-11eb-8e28-62c2d404e713.png)

It would be good to check whether the message exists on the `matrix.org` homeserver database but is `soft_failed` to confirm this suspicion. Can someone with access check?

@Half-Shot has also seen this happen a few times with the IRC bridge.

As @leonerd from https://github.com/matrix-org/synapse/issues/1444 and @Half-Shot mentioned, this is probably not a problem for normal users because the time taken to join and then send a message is some number of seconds. Whereas with appservices, can join and send an event in quick succession (almost instantly).

The related issue https://github.com/matrix-org/synapse/issues/1444 reproduces with a `join`, set profile, `membership` event.

### Steps to reproduce

 - Create a room on HS1, `!my-room:hs1`
 - Setup HS2 with an appservice
 - Join and send a message through the appservice to `!my-room:hs1`
    - The appservice is probably not necessary to reproduce. Just send a `join` and message event from HS2
 - Check whether the message is visible on hs1
 - Since it's a race condition, this probably does not always reproduce

<!--
Describe how what happens differs from what you expected.

If you can identify any relevant log snippets from _homeserver.log_, please include
those (please be careful to remove any personal or private data). Please surround them with
``` (three backticks, on a line on their own), so that they are formatted legibly.
-->

### Version information

<!-- IMPORTANT: please answer the following questions, to help us narrow down the problem -->

<!-- Was this issue identified on matrix.org or another homeserver? -->
[...]

PR #8607, opened 2020-10-20
Support outputting structured logs in addition to standard logs

clokep:
This supports outputting structured logs at the same time as "standard" logs by reworking how the structured logging system works to be configurable as part of the standard Python logging config. This unfortunately got very large, but I've been unable to come up with a reasonable plan to split it up. It should be reviewable commit-by-commit, although a couple are still rather large! I can walk through it with someone if that's easiest!

The first commit removes the custom configuration code (`synapse.logging._structured`) for structured logging, which is no longer necessary.

The main commit converts the JSON log observers into a `Formatter` and the TCP log observer into a `Handler`. This is essentially converting from Twisted logging infrastructure to the Python standard library infrastructure. Due to this, the terse JSON code gets simplified since a `LogRecord` is simpler than the Twisted log events.

After doing that we need to add back support for passing the `server_name` through to the output via a `Filter`.

Finally, we update synmark to use the new code, which is fairly invasive. I ended up modifying it to avoid creating a homeserver and *only instantiating the logging framework*, this should be more realistic for benchmarking.

There's also a couple of clean-up commits, but hopefully those are self-explanatory.

This builds on #8587 

Fixes #8588

PR #10156, opened 2021-06-09
Add metrics to track how often events are `soft_failed`

MadLittleMods:
Add metrics to track how often events are `soft_failed`

Spawned from missing messages we were seeing on `matrix.org` from a
federated Gitter bridged room, https://gitlab.com/gitterHQ/webapp/-/issues/2770.
The underlying issue in Synapse is tracked by https://github.com/matrix-org/synapse/issues/10066
where the message and join event race and the message is `soft_failed` before the
`join` event reaches the remote federated server.

Less soft_failed events = better and usually this should only trigger for events
where people are doing bad things and trying to fuzz and fake everything.

This metric does not track what situation causes the event to be `soft_failed` but could be used to know how many people roughly are potentially running into https://github.com/matrix-org/synapse/issues/10066.

### Pull Request Checklist

### DIFF
--- a/changelog.d/10156.misc
+++ b/changelog.d/10156.misc
@@ -1,0 +1,1 @@
+Add `synapse_handlers_federation_soft_failed_event` metric to track how often events are `soft_failed`.
--- a/changelog.d/10156.misc
+++ b/changelog.d/10156.misc
@@ -1,1 +1,1 @@
-Add `synapse_federation_soft_failed_events_total` metric to track how often events are `soft_failed`.
+Add `synapse_federation_soft_failed_events_total` metric to track how often events are soft failed.
--- a/changelog.d/10156.misc
+++ b/changelog.d/10156.misc
@@ -1,1 +1,1 @@
-Add `synapse_handlers_federation_soft_failed_event` metric to track how often events are `soft_failed`.
+Add `synapse_federation_soft_failed_events_total` metric to track how often events are `soft_failed`.
--- a/synapse/handlers/federation.py
+++ b/synapse/handlers/federation.py
@@ -33,6 +33,7 @@ from typing import (
 )
 
 import attr
+from prometheus_client import Counter
 from signedjson.key import decode_verify_key_bytes
 from signedjson.sign import verify_signed_json
 from unpaddedbase64 import decode_base64
@@ -100,6 +101,8 @@ if TYPE_CHECKING:
     from synapse.server import HomeServer
 
 logger = logging.getLogger(__name__)
+
+soft_failed_event_counter = Counter("synapse_handlers_federation_soft_failed_event", "")
 
 
 @attr.s(slots=True)
@@ -2498,6 +2501,7 @@ class FederationHandler(BaseHandler):
             event_auth.check(room_version_obj, event, auth_events=current_auth_events)
         except AuthError as e:
             logger.warning("Soft-failing %r because %s", event, e)
+            soft_failed_event_counter.inc()
             event.internal_metadata.soft_failed = True
 
     async def on_get_missing_events(
--- a/synapse/handlers/federation.py
+++ b/synapse/handlers/federation.py
@@ -102,7 +102,10 @@ if TYPE_CHECKING:
 
 logger = logging.getLogger(__name__)
 
-soft_failed_event_counter = Counter("synapse_federation_soft_failed_events_total", "")
+soft_failed_event_counter = Counter(
+    "synapse_federation_soft_failed_events_total",
+    "Events received over federation that we marked as soft_failed",
+)
 
 
 @attr.s(slots=True)
--- a/synapse/handlers/federation.py
+++ b/synapse/handlers/federation.py
@@ -102,7 +102,7 @@ if TYPE_CHECKING:
 
 logger = logging.getLogger(__name__)
 
-soft_failed_event_counter = Counter("synapse_handlers_federation_soft_failed_event", "")
+soft_failed_event_counter = Counter("synapse_federation_soft_failed_events_total", "")
 
 
 @attr.s(slots=True)


synapse/handlers/federation.py
```diff
@@ -101,6 +102,10 @@
 
 logger = logging.getLogger(__name__)
 
+soft_failed_event_counter = Counter(
+    "synapse_handlers_federation_events_soft_failed", ""
```
MadLittleMods, line 106: This metric name is based off of the following counters already in the codebase. It looks like there is already inconsistency in the plurality

 - [`synapse_handlers_appservice_events_processed`](https://github.com/matrix-org/synapse/blob/aec2cf1c9832eac46e1d94a39e75490b02320555/synapse/handlers/appservice.py#L44)
 - [`synapse_handler_presence_notified_presence`](https://github.com/matrix-org/synapse/blob/aec2cf1c9832eac46e1d94a39e75490b02320555/synapse/handlers/presence.py#L75)
  - [`synapse_storage_events_persisted_events`](https://github.com/matrix-org/synapse/blob/aec2cf1c9832eac46e1d94a39e75490b02320555/synapse/storage/databases/main/events.py#L58)
   - [`synapse_storage_events_state_delta_single_event`](https://github.com/matrix-org/synapse/blob/aec2cf1c9832eac46e1d94a39e75490b02320555/synapse/storage/persist_events.py#L51)

synapse/handlers/federation.py
```diff
@@ -101,6 +102,8 @@
 
 logger = logging.getLogger(__name__)
 
+soft_failed_event_counter = Counter("synapse_handler_federation_soft_failed_event", "")
```
MadLittleMods, line 105: This metric name is based off of the following counters already in the codebase. It looks like there is already inconsistency in the plurality

 - [`synapse_handlers_appservice_events_processed`](https://github.com/matrix-org/synapse/blob/aec2cf1c9832eac46e1d94a39e75490b02320555/synapse/handlers/appservice.py#L44)
 - [`synapse_handler_presence_notified_presence`](https://github.com/matrix-org/synapse/blob/aec2cf1c9832eac46e1d94a39e75490b02320555/synapse/handlers/presence.py#L75)
  - [`synapse_storage_events_persisted_events`](https://github.com/matrix-org/synapse/blob/aec2cf1c9832eac46e1d94a39e75490b02320555/synapse/storage/databases/main/events.py#L58)
   - [`synapse_storage_events_state_delta_single_event`](https://github.com/matrix-org/synapse/blob/aec2cf1c9832eac46e1d94a39e75490b02320555/synapse/storage/persist_events.py#L51)

synapse/handlers/federation.py
```diff
@@ -101,6 +102,8 @@
 
 logger = logging.getLogger(__name__)
 
+soft_failed_event_counter = Counter("synapse_handlers_federation_soft_failed_event", "")
```
MadLittleMods, line 105: This metric name is based off of the following counters already in the codebase. It looks like there is already inconsistency in the plurality. It makes sense that it matches the file name `synapse/handlers/federation.py`

 - [`synapse_handlers_appservice_events_processed`](https://github.com/matrix-org/synapse/blob/aec2cf1c9832eac46e1d94a39e75490b02320555/synapse/handlers/appservice.py#L44)
 - [`synapse_handler_presence_notified_presence`](https://github.com/matrix-org/synapse/blob/aec2cf1c9832eac46e1d94a39e75490b02320555/synapse/handlers/presence.py#L75)
  - [`synapse_storage_events_persisted_events`](https://github.com/matrix-org/synapse/blob/aec2cf1c9832eac46e1d94a39e75490b02320555/synapse/storage/databases/main/events.py#L58)
   - [`synapse_storage_events_state_delta_single_event`](https://github.com/matrix-org/synapse/blob/aec2cf1c9832eac46e1d94a39e75490b02320555/synapse/storage/persist_events.py#L51)

anoadragon453 (commented):
Thanks for the detailed explanation. Code looks good, though I wonder if a simple counter is the most useful metric here versus something that may give us an idea of how many rooms soft failed events are appearing across.

I'm not sure of a metric type that could be used for that though, so leaving this open for someone to have a second look.

synapse/handlers/federation.py
richvdh: > synapse_handler_presence_notified_presence (plurality different)

I'm not sure what the plural of that would be. `synapse_handler_presence_notified_presences`, maybe.

Counters should generally be plural, and end in `_total`, per https://prometheus.io/docs/practices/naming - `synapse_handlers_federation_soft_failed_events_total` in this case. (Actually the prometheus library will add `_total` for you if you don't add it.)

Some background on the naming of our existing metrics: once upon a time, someone thought it would be a good idea to name them according to the python module they were defined in. I don't know why they thought that, but I don't think it was a good idea, and I don't necessarily think we need to stick slavishly to it. 

I think I would go with `synapse_federation_soft_failed_events_total` here.

richvdh, line 105: also, please populate the help string (the second arg to `Counter`). I know not all of our existing code sets a good example here, but let's get it right for new code.

changelog.d/10156.misc
```diff
@@ -0,0 +1 @@
+Add `synapse_handlers_federation_soft_failed_event` metric to track how often events are `soft_failed`.
```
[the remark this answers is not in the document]
richvdh: remember to update this to match the name of the metric.

richvdh (commented):
As @anoadragon453 says, I'm not sure if this will help a great deal, because you don't know where the failures are happening. That said, I don't really think there's a better way to do this, and at least if the soft-failures counter is zero, that gives you an easy way to rule out soft-failure as a problem.

synapse/handlers/federation.py
MadLittleMods: Thanks for the docs and context here! Updated to `synapse_federation_soft_failed_events_total`

MadLittleMods:
Thanks for the review passes @anoadragon453 and @richvdh 🕺

I think it's useful to know whether it's happening or not regardless of how.

---

For the future:

If we want to dig deeper into what rooms are soft_failing messages, we could use the Elasticsearch logs and some aggregations (Kibana or raw ES queries). We can add a few fields to the Elasticsearch mapping, `room_id`, `mxid`, `event_id` which should handle the high cardinality(lots of different values) just fine.

It looks like we use [Logstash](https://github.com/matrix-org/matrix-ansible-private/tree/main/roles/logstash) or [Filebeat](https://github.com/matrix-org/matrix-ansible-private/tree/main/roles/filebeat) and I assume there is a way we can parse another field out besides `message`? Maybe https://stackoverflow.com/q/40460830/796832

synapse/handlers/federation.py
```diff
@@ -2498,6 +2504,7 @@ async def _check_for_soft_fail(
             event_auth.check(room_version_obj, event, auth_events=current_auth_events)
         except AuthError as e:
             logger.warning("Soft-failing %r because %s", event, e)
```
MadLittleMods, line 2506: > For the future:
> 
> If we want to dig deeper into what rooms are soft_failing messages, we could use the Elasticsearch logs and some aggregations (Kibana or raw ES queries). We can add a few fields to the Elasticsearch mapping, room_id, mxid, event_id which should handle the high cardinality(lots of different values) just fine.
> 
> It looks like we use Logstash or Filebeat and I assume there is a way we can parse another field out besides message? Maybe https://stackoverflow.com/q/40460830/796832
> 
> *-- https://github.com/matrix-org/synapse/pull/10156#issuecomment-859287143*


To document my additional findings here;

We can already use `logger.info('foo', extra={ "foo": "bar"}` as shown in [`tests/logging/test_terse_json.py#L64-L74`](https://github.com/matrix-org/synapse/blob/13577aa55ebe6087e8b813c0643bbb53148e9510/tests/logging/test_terse_json.py#L64-L74) to add extra fields to the structured logging.

 - Docs on structured logging in Synapse: https://github.com/matrix-org/synapse/blob/develop/docs/structured_logging.md
 - Improve structured logging: https://github.com/matrix-org/synapse/issues/8683
    - Record more information into structured logs: https://github.com/matrix-org/synapse/pull/9654
 - Use structured logging for EMS: https://github.com/matrix-org/matrix-hosted/issues/1229
 - Use structured logging for `matrix.org`: https://github.com/matrix-org/internal-config/issues/889
 - [Allow simultaneously enabling structured logging and "standard" logging](https://github.com/matrix-org/synapse/issues/8588) which was implemented in https://github.com/matrix-org/synapse/pull/8607
 - List of issues with the `z-logging` label: https://github.com/matrix-org/synapse/labels/z-logging

MadLittleMods: Created https://github.com/matrix-org/synapse/pull/10168 to add these fields

PR #10157, opened 2021-06-09
Extend `ResponseCache` to pass a context object into the callback

### LINKED ISSUE #8518: Sync requests immediately return with empty payload

### LINKED ISSUE #9738: class in `test_responsecache.py` has the wrong name (`DeferredCacheTestCase` instead of `ResponseCacheTestCase`)
This is my own fault, but i'm putting it here because i noticed it while working on something else.

The file under `tests/util/caches/test_responsecache.py` uses `DeferredCacheTestCase`, `test_deferred_cache.py` also uses this classname, so it has to be `ResponseCacheTestCase` instead.

richvdh:
This is the first of two PRs which seek to address #8518. This first PR lays the groundwork by extending `ResponseCache`; a second PR (#10158) will update the `SyncHandler` to actually use it, and fix the bug. The two PRs form an alternative to #9739.

The idea here is that we allow the callback given to `ResponseCache.wrap` to decide whether its result should be cached or not. We do that by (optionally) passing a `ResponseCacheContext` into it, which it can modify.

I think this is considerably simpler than the solution in #9739.

(The commits in this PR should be reviewable individually - the first three are just minor cleanups and refactoring; the meat happens in 63e6c20.)

Also fixes #9738

### DIFF
--- a/tests/util/caches/test_response_cache.py
+++ b/tests/util/caches/test_response_cache.py
@@ -11,10 +11,11 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+from parameterized import parameterized
 
 from twisted.internet import defer
 
-from synapse.util.caches.response_cache import ResponseCache
+from synapse.util.caches.response_cache import ResponseCache, ResponseCacheContext
 
 from tests.server import get_clock
 from tests.unittest import TestCase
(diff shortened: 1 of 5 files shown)


tests/util/caches/test_response_cache.py
```diff
@@ -129,3 +143,49 @@ def test_cache_wait_expire(self):
...
+    def test_cache_context_nocache(self, should_cache: bool):
+        """If the callback clears the should_cache bit, the result should not be cached"""
+        cache = self.with_cache("medium_cache", ms=3000)
+
+        expected_result = "howdy"
+
+        call_count = [0]
+
+        async def non_caching(o: str, cache_context: ResponseCacheContext[int]):
+            call_count[0] += 1
```
ShadowJonathan, line 157: I understand what is happening here, the actual variable needs to be referenced, and that is done by putting it inside a list.

You can make nonlocal variable names mutable by using `nonlocal`, so this can be the following instead (also changing above to `call_count = 0`, and below to `self.assertEqual(call_count, 1)`);

```suggestion
            nonlocal call_count
            call_count += 1
```

ShadowJonathan (approved):
Thank you for finally making a PR that can do what #9739 was supposed to do.

I also see you renamed the TestCase class, that technically fixes #9738, so maybe add it to the description.

tests/util/caches/test_response_cache.py
richvdh: TIL

ShadowJonathan: FWIW the same is possible with `global`

erikjohnston (approved):

PR #10158, opened 2021-06-09
Ensure that we do not cache empty sync responses after a timeout

### LINKED ISSUE #8518: Sync requests immediately return with empty payload

richvdh:
Fixes #8518 by telling the `ResponseCache` not to cache the `/sync` response if the next_batch param is the same as the `since` token.

~~Based on #10157.~~

Note that actually enabling the sync cache timeout (ie, #3880) is left for a future task.

### DIFF
--- a/synapse/types.py
+++ b/synapse/types.py
@@ -404,7 +404,7 @@ def map_username_to_mxid_localpart(
     return username.decode("ascii")
 
 
-@attr.s(frozen=True, slots=True, cmp=False)
+@attr.s(frozen=True, slots=True, order=False)
 class RoomStreamToken:
     """Tokens are positions between events. The token "s1" comes after event 1.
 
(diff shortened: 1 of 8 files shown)


synapse/types.py
```diff
@@ -404,7 +404,7 @@ def f2(m):
     return username.decode("ascii")
 
 
-@attr.s(frozen=True, slots=True, cmp=False)
+@attr.s(frozen=True, slots=True, order=False)
```
richvdh, line 407: I had to fix this to make the `if result.next_batch == since_token` conditional work. As I wrote in #8518, I think it might have been exacerbating the situation.

ShadowJonathan (approved):
Looks good, thanks for helping out!

erikjohnston (approved):

PR #10160, opened 2021-06-10
Manually pull the complement branch matching the current branch name.

clokep:
This is akin to matrix-org/pipelines#157, but using GitHub Actions.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -326,7 +326,7 @@ jobs:
               continue
             fi
 
-            (wget -O - https://github.com/matrix-org/complement/archive/$BRANCH_NAME.tar.gz | tar -xz --strip-components=1 -C complement) && break
+            (wget -O - "https://github.com/matrix-org/complement/archive/$BRANCH_NAME.tar.gz" | tar -xz --strip-components=1 -C complement) && break
           done
 
       # Build initial Synapse image
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -308,9 +308,16 @@ jobs:
       # Attempt to check out the same branch of Complement as the PR. If it
       # doesn't exist, fallback to master.
       - name: Checkout complement
+        shell: bash
         run: |
+          # For pull requests we want to use GITHUB_HEAD_REF (the pull request
+          # head branch). Otherwise, we use GITHUB_REF (refs/heads/<branch>).
+          BRANCH_NAME="$GITHUB_HEAD_REF"
+          if [[ -z "$BRANCH_NAME" ]]; then
+            BRANCH_NAME="${GITHUB_REF#refs/heads/}"
+          fi
           mkdir -p complement
-          (wget -O - https://github.com/matrix-org/complement/archive/$GITHUB_HEAD_REF.tar.gz || wget -O - https://github.com/matrix-org/complement/archive/master.tar.gz) | tar -xz --strip-components=1 -C complement
+          (wget -O - https://github.com/matrix-org/complement/archive/$BRANCH_NAME.tar.gz || wget -O - https://github.com/matrix-org/complement/archive/master.tar.gz) | tar -xz --strip-components=1 -C complement
 
       # Build initial Synapse image
       - run: docker build -t matrixdotorg/synapse:latest -f docker/Dockerfile .
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -310,14 +310,24 @@ jobs:
       - name: Checkout complement
         shell: bash
         run: |
-          # For pull requests we want to use GITHUB_HEAD_REF (the pull request
-          # head branch). Otherwise, we use GITHUB_REF (refs/heads/<branch>).
-          BRANCH_NAME="$GITHUB_HEAD_REF"
-          if [[ -z "$BRANCH_NAME" ]]; then
-            BRANCH_NAME="${GITHUB_REF#refs/heads/}"
-          fi
           mkdir -p complement
-          (wget -O - https://github.com/matrix-org/complement/archive/$BRANCH_NAME.tar.gz || wget -O - https://github.com/matrix-org/complement/archive/master.tar.gz) | tar -xz --strip-components=1 -C complement
+          # Attempt to use the version of complement which best matches the current
+          # build. Depending on whether this is a PR or release, etc. we need to
+          # use different fallbacks.
+          #
+          # 1. First check if there's a similarly named branch (GITHUB_HEAD_REF
+          #    for pull requests, otherwise GITHUB_REF).
+          # 2. Attempt to use the base branch, e.g. when merging into release-vX.Y
+          #    (GITHUB_BASE_REF for pull requests).
+          # 3. Use the default complement branch ("master").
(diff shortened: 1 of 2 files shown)


clokep:
I'm fairly certain we cannot use the action here since we need the fallback behavior?

anoadragon453 (approved):
The fallback looks to work from the CI output. From the code it appears the same-named branch stuff should work as well.

As you said, I don't think fallback behaviour is possible with `actions/checkout`.

.github/workflows/tests.yml
```diff
@@ -305,11 +305,12 @@ jobs:
...
-        uses: actions/checkout@v2
-        with:
-          repository: "matrix-org/complement"
-          path: complement
+      # Attempt to check out the same branch of Complement as the PR. If it
+      # doesn't exist, fallback to master.
+      - name: Checkout complement
+        run: |
+          mkdir -p complement
+          (wget -O - https://github.com/matrix-org/complement/archive/$GITHUB_HEAD_REF.tar.gz || wget -O - https://github.com/matrix-org/complement/archive/master.tar.gz) | tar -xz --strip-components=1 -C complement
```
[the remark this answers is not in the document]
clokep: I saw that, but don't we only run builds for PRs anyway?

There's also `$GITHUB_REF`, but I don't think that returns the information I need.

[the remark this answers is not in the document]
anoadragon453: If you check the top of the workflow file you'll see that we run it whenever `develop` or `release-*` branches are pushed to: https://github.com/matrix-org/synapse/blob/2df7196c00856df6b7e25e8ec181e37e8eba28de/.github/workflows/tests.yml#L3-L6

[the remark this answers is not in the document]
callahad: I may be cleaner to keep using `actions/checkout@v2` here and just do a full clone (`fetch-depth: 0`) followed by opportunistically trying to `git checkout` the matching branch?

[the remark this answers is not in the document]
clokep: @anoadragon453 Thanks! I believe I've updated this to take this into account, it now:

1. Uses `GITHUB_HEAD_REF` if it exists.
2. Falls back to `GITHUB_REF` (and cuts off a `refs/head/` prefix)
3. Falls back to `master`

callahad:
A few options here:

- Complement is tiny; cloning the whole history is less than 1 MiB. You could grab the whole history (`fetch-depth: 0`) and then try to `git checkout` the target branch, ignoring (and thus staying on `master`) if that fails.

- You could stick with the `wget` logic you have, but might be slightly easier to instead try to call `git clone --shallow --branch $whatever https://github.com/matrix-org/complement`; then you don't have to directly muck with wget and tar.

- If you can somehow easily determine whether the branch exists before actually cloning the repo, you could do that in one step and pass it along as the `ref` argument to the `actions/checkouts` step as an `output`. E.g., if a step echoes `::set-output name=branch::$whatever`,  then in the following steps can access that as `${{ steps.STEP_ID.outputs.branch }}`

To find the right reference from either a `push` or a `pull_request` you might need to dig into [events that trigger workflows](https://docs.github.com/en/actions/reference/events-that-trigger-workflows), [the `github` context](https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#github-context), and [default environment variables](https://docs.github.com/en/actions/reference/environment-variables#default-environment-variables). E.g., for pull requests you probably want to look at `github.head_ref` or `github.base_ref`, but for pushes you probably want`github.ref`.

All of those are also available as envvars: `$GITHUB_HEAD_REF`, `$GITHUB_BASE_REF`, and `$GITHUB_REF`.

.github/workflows/tests.yml
```diff
@@ -305,11 +305,19 @@ jobs:
...
-          repository: "matrix-org/complement"
-          path: complement
+      # Attempt to check out the same branch of Complement as the PR. If it
+      # doesn't exist, fallback to master.
+      - name: Checkout complement
+        shell: bash
+        run: |
+          # For pull requests we want to use GITHUB_HEAD_REF (the pull request
+          # head branch). Otherwise, we use GITHUB_REF (refs/heads/<branch>).
+          BRANCH_NAME="$GITHUB_HEAD_REF"
```
callahad, line 315: Is `$GITHUB_HEAD_REF` right, here? Should we instead (or also?) look at `$GITHUB_BASE_REF`?

E.g., if I have a branch called `issue-451` and make a PR against Synapse's `release-v1.34` branch, then I'm looking at:

    GITHUB_HEAD_REF="issue-451"
    GITHUB_BASE_REF="release-v1.34"
    GITHUB_REF="refs/pull/10700/merge"

We probably want to look for matching branches in complement for _both_ the `HEAD` and `BASE`, before falling back to `master`, right?

anoadragon453: During the release process for Synapse, we'll typically create and push a matching branch on Sytest. If we started doing the same for Complement, then falling back to the base ref instead of master (which might be ahead of the release branch at that point) is a good idea.

So I think `GITHUB_HEAD_REF` -> `GITHUB_BASE_REF` -> `master` is a sound idea.

callahad: Concretely, it may be cleaner to do something like

```bash
# Attempt to check out the most relevant branch of Complement, if it exists
#   GITHUB_HEAD_REF and GITHUB_BASE_REF are only set on pulls; otherwise empty
#   GITHUB_REF is /refs/pull/BRANCH on pushes, but refs/pull/123/merge on pulls
# Falls back to master if no more specific branch is found
for branch in "$GITHUB_HEAD_REF" "$GITHUB_BASE_REF" "${GITHUB_REF#refs/heads/}" "master"; do
    if [ -z "$branch" ] || [[ $branch =~ ^refs/pull/.* ]]; then
        continue
    fi

    git clone --shallow --branch "$branch" https://github.com/matrix-org/complement && break
done
```

clokep:
@anoadragon453 Please take another look

.github/workflows/tests.yml
```diff
@@ -305,11 +305,29 @@ jobs:
...
+          # 2. Attempt to use the base branch, e.g. when merging into release-vX.Y
+          #    (GITHUB_BASE_REF for pull requests).
+          # 3. Use the default complement branch ("master").
+          for BRANCH_NAME in "$GITHUB_HEAD_REF" "$GITHUB_BASE_REF" "${GITHUB_REF#refs/heads/}" "master"; do
+            # Skip empty branch names and merge commits.
+            if [[ -z "$BRANCH_NAME" || $BRANCH_NAME =~ ^refs/pull/.* ]]; then
+              continue
+            fi
+
+            (wget -O - https://github.com/matrix-org/complement/archive/$BRANCH_NAME.tar.gz | tar -xz --strip-components=1 -C complement) && break
```
callahad, line 329: The URL needs to be quoted per shellcheck to avoid errors if the branch name somehow ends up containing a space

callahad: Is there a reason to prefer wget over a shallow clone of the repo?

clokep: I lost that somehow when refactoring... 😢

clokep: I also don't know what you mean by "per spellcheck".

callahad: Oh, sorry 🙂 Shellcheck is a really, really excellent static analysis tool for shell scripts: http://shellcheck.net/

anoadragon453 (approved):
lgtm :slightly_smiling_face:

PR #10161, opened 2021-06-10
Remove support for unstable MSC1772 prefixes.

### LINKED ISSUE #9929: Remove unstable MSC1772 identifiers
Now that we support the stable identifiers (#9915) we can at some point in the future remove support for the unstable ones. These should all appear prefixed with `org.matrix.msc1772`.

clokep:
This drops support for unstable prefixes for MSC1772 (`org.org.matrix.msc1772.space.*`). This will change to only support the `m.space.*` versions.

I think this is fine since current Element clients (which I believe are the only things to support spaces) only use the stable prefixes and we said very loudly from the beginning that the unstable prefixes would break.

The reality of this is that you would not be able to use the spaces summary on those anymore, but current Elements treat them as normal rooms anyway so it shouldn't matter.

Fixes #9929

### DIFF
--- a/changelog.d/10161.removal
+++ b/changelog.d/10161.removal
@@ -1,0 +1,1 @@
+Stop supporting the unstable spaces prefixes from MSC1772.
--- a/synapse/api/constants.py
+++ b/synapse/api/constants.py
@@ -112,8 +112,6 @@ class EventTypes:
 
     SpaceChild = "m.space.child"
     SpaceParent = "m.space.parent"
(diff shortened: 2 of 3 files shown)


clokep:
I seem to have missed a change in complement at some point...

clokep:
> I seem to have missed a change in complement at some point...

See matrix-org/complement#124.

richvdh (approved):
lgtm

PR #10163, opened 2021-06-11
Fix sending presence over federation when using workers

### LINKED ISSUE #10153: Outgoing presence exploded in 1.35.0rc1
### Description

There seems to be a huge spike in outgoing presence right after 1.35.0rc1 was released. Incoming presence also went up, but not as much.

image

My federation sender is also having trouble with metrics after the 1.36.0rc1 upgrade (which is what I was looking into when I noticed the presence problem), but that might be a separate issue:

image

### Version information

- **Version**: 1.36.0rc1
- **Install method**: Docker
- **Workers**: federation sender, 3 generic workers (federation reader, synchrotron, `events` stream writer), appservice sender, media repo

erikjohnston:
When using a federation sender we'd send out all local presence updates over
federation even when they shouldn't be.

Fixes #10153.

Introduced in #9819

### DIFF
--- a/changelog.d/10163.bugfix
+++ b/changelog.d/10163.bugfix
@@ -1,0 +1,1 @@
+Fix a bug when using federation sender worker where it would send out more presence updates than necessary, leading to high resource usage. Broke in v1.33.0.
--- a/synapse/handlers/presence.py
+++ b/synapse/handlers/presence.py
@@ -495,9 +495,6 @@ class WorkerPresenceHandler(BasePresenceHandler):
             users=users_to_states.keys(),
         )
 
-        # If this is a federation sender, notify about presence updates.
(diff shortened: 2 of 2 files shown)


anoadragon453 (approved):
Good catch! I assume the presence writer sending non-notify presence states over replication is necessary (and we can't just prevent that behaviour)?

erikjohnston:
> Good catch! I assume the presence writer sending non-notify presence states over replication is necessary (and we can't just prevent that behaviour)?

The alternative approach here is when writing to the DB add some sort of flag to say whether they should be streamed over replication or not, but that becomes tricky as you need to handle replacing existing rows that may or may not need to be notified, etc.

erikjohnston:
We deployed this on vector.modular.im and it seems to have worked \o/

PR #10165, opened 2021-06-11
Only send a presence state to a destination once

erikjohnston:
It turns out that we were sending the same presence state to a remote potentially multiple times.

In practice this is probably isn't causing too many issues, as each federation destination queue will dedupe presence updates. However, depending on caching, it is possible that we'd start a new transaction the first time the destination queue gets told about an update. This would cause us to send the presence update twice.


anoadragon453 (approved):

erikjohnston:
Complement is failing as this is based on the release branch

PR #10167, opened 2021-06-11
Remove the experimental flag for knocking and use stable prefixes / endpoints.

clokep:
Should be reviewable commit-by-commit:

* Removes the experimental config flag.
* Uses stable:
	* Room version
	* Endpoints
	* Prefixes

See matrix-org/complement#125 for complement updates.


anoadragon453 (approved):
This looks like it covers everything to me. Thank you very much for doing this!

PR #10175, opened 2021-06-14
Fix requestOpenIdToken response: integer expires_in

lukaslihotzki:
`expires_in` must be an integer according to the OpenAPI spec:
https://github.com/matrix-org/matrix-doc/blob/master/data/api/client-server/definitions/openid_token.yaml#L32

True division (`/`) returns a float instead (`"expires_in": 3600.0`).
Floor division (`//`) returns an integer, so the response is spec compliant.

### DIFF
--- a/changelog.d/10175.bugfix
+++ b/changelog.d/10175.bugfix
@@ -1,1 +1,1 @@
-Return an integer `expires_in` in requestOpenIdToken response.
+Fix a minor bug in the response to `/_matrix/client/r0/user/{user}/openid/request_token`. Contributed by @lukaslihotzki.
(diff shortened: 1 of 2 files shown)


lukaslihotzki:
It seems like the test fail wasn't caused by the change in this PR, because the change should be database-independent. Can the failed test be retried somehow?

changelog.d/10175.bugfix
```diff
@@ -0,0 +1 @@
+Return an integer `expires_in` in requestOpenIdToken response.
```
richvdh, line 1:
```suggestion
Fix a minor bug in the response to `/_matrix/client/r0/user/{user}/openid/request_token`. Contributed by @lukaslihotzki.
```

richvdh (approved):
thank you!

Issue #9253, opened 2021-01-28
Mitigate Read Receipts by overwriting them


peete1311:
Hello,

I'm implementing Synapse for a public sector entity where the worker's council is likely to enforce enhanced privacy for message read receipts. As far as I understood Synapse's event handling, this isn't easy to accomplish in the current state of implementation.

I'm currently looking at pseudonymizing read receipt events for a public room as a workaround (others, like private rooms, might be harder to implement and will likely not be requested here in the first place). Is it possible to overwrite events in the database, i.e. so that every m.read or m.fully_read is then related to some dummy user (e.g. a service account)? People would see 'someone' read the message but can't tell who it was anymore.

erikjohnston:
Read receipts happen outside of the event graph, so they can easily be dropped by the server. Would adding a switch to the config to drop all read receipts address your needs?

peete1311:
Yep, I think it would. What about already existing read receipts though?

EdGeraghty:
> Read receipts happen oureadHey e of the event graph, so they can easily be dropped by the server. Would adding a switch to the config to drop all read receipts address your needs?

Hi - I'm really pleased that Matrix have finally managed to decouple read receipts from the legacy federation and multidevice problems which acted as blockers until now! 🎉

Could we please get this config switch added?

In the time between me and my colleagues testing and deploying Matrix at our (Privacy) org, the labs option to switch off read receipts from the client had been disabled. Several members of staff have (quite rightly) questioned us on this.

Please implement this switch, it's a humongous blocker for a lot of organisations, and has been for years!

EdGeraghty:
Hey!

Is this in the Synapse dev timeline? 🤞

erikjohnston:
I think we still need to support receiving read receipts from local users, so that the notification counts are correct, but we could simple not relay those receipts to local or remote users.

erikjohnston:
> Hey!
> 
> Is this in the Synapse dev timeline?

I'm afraid this isn't something we'd likely get around to anytime soon, but we'd happily accept a PR to add support for it.

EdGeraghty:
Something like https://github.com/matrix-org/synapse/pull/10177/ ?

PR #10177, opened 2021-06-14
Add configuration for dropping read receipts

EdGeraghty:
### Pull Request Checklist

### DIFF
--- a/changelog.d/10177.feature
+++ b/changelog.d/10177.feature
@@ -1,1 +1,1 @@
-Add new configuration option for dropping read receipts at the server level.  [#10177](https://github.com/matrix-org/matrix-doc/pull/10177). Contributed by Ed Geraghty
+Add new configuration option for dropping read receipts at the server level.  [#10177](https://github.com/matrix-org/matrix-doc/pull/10177). Contributed by Ed Geraghty.
--- a/changelog.d/10177.feature
+++ b/changelog.d/10177.feature
@@ -1,1 +1,1 @@
-Add new configuration option for dropping read receipts at the server level.  [#10177](https://github.com/matrix-org/matrix-doc/pull/10177). Contributed by Ed Geraghty.
+Add new configuration option for dropping read receipts at the server level. Contributed by Ed Geraghty.
--- a/changelog.d/10177.feature
+++ b/changelog.d/10177.feature
@@ -1,0 +1,1 @@
+Add new configuration option for dropping read receipts at the server level.  [#10177](https://github.com/matrix-org/matrix-doc/pull/10177). Contributed by Ed Geraghty
--- a/synapse/config/server.py
+++ b/synapse/config/server.py
@@ -1192,7 +1192,11 @@ class ServerConfig(Config):
 
         # Read receipt sending policy at the server level.
         #
-        # Defaults to `true`. Uncomment to disable sending read receipts.
+        # Homeserver administrators may wish to disable the sending and receiving of read
+        # receipts to and from local users for organisational policy, local law, or just
+        # for privacy reasons.
+        #
+        # Defaults to `true`. Uncomment to disable sending and receiving read receipts.
         #
         #send_read_receipts: false
 
--- a/synapse/config/server.py
+++ b/synapse/config/server.py
@@ -684,6 +684,9 @@ class ServerConfig(Config):
             False,
         )
 
+        # Disable sending read receipts from this homeserver
+        self.send_read_receipts = config.get("send_read_receipts", True)
+
         # List of users trialing the new experimental default push rules. This setting is
         # not included in the sample configuration file on purpose as it's a temporary
         # hack, so that some users can trial the new defaults without impacting every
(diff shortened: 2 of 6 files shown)


EdGeraghty:
Signed-off-by: Ed Geraghty <ed@geraghty.london>

changelog.d/10177.feature
```diff
@@ -0,0 +1 @@
+Add new configuration option for dropping read receipts at the server level.  [#10177](https://github.com/matrix-org/matrix-doc/pull/10177). Contributed by Ed Geraghty.
```
richvdh, line 1: the PR number is automatically included, so no need to include it explicitly.

```suggestion
Add new configuration option for dropping read receipts at the server level. Contributed by Ed Geraghty.
```

synapse/server.py
```diff
@@ -789,3 +789,7 @@ def get_outbound_redis_connection(self) -> Optional["RedisProtocol"]:
     def should_send_federation(self) -> bool:
         "Should this server be sending federation traffic directly?"
         return self.config.send_federation
+
+    def should_send_read_receipts(self) -> bool:
+        "Should this server be sending read receipts onwards?"
+        return self.config.send_read_receipts
```
richvdh, line 795: I don't think we need an accessor here (`should_send_federation` is overridden in a subclass, so is special). you can just do `hs.config.send_read_receipts` at the call site.

synapse/handlers/receipts.py
```diff
@@ -139,13 +142,17 @@ async def received_client_receipt(
         if not is_new:
             return
 
+        if not self.hs.should_send_read_receipts():
+            return
+
```
richvdh, line 147: I think `_handle_new_receipts` returns `False` in this case, so this is unreachable and redundant?

synapse/config/server.py
```diff
@@ -1184,6 +1187,12 @@ def generate_config_section(
         #
         #user_ips_max_age: 14d
 
+        # Read receipt sending policy at the server level.
```
richvdh, line 1190: Could you add some more words about what this actually means?

synapse/handlers/read_marker.py
```diff
@@ -44,6 +44,9 @@ async def received_client_read_marker(
         """
 
         with await self.read_marker_linearizer.queue((room_id, user_id)):
+            if not self.hs.should_send_read_receipts():
+                return
+
```
richvdh, line 49: this one seems to be blocking requests from the local user to send receipts. As noted on https://github.com/matrix-org/synapse/issues/9253, that's risky - can you clarify the intention here?

(also, this can safely move outside the `read_marker_linearizer` block)

synapse/handlers/receipts.py
```diff
@@ -111,6 +111,9 @@ async def _handle_new_receipts(self, receipts: List[ReadReceipt]) -> bool:
             # no new receipts
             return False
 
+        if not self.hs.should_send_read_receipts():
```
richvdh, line 114: I think this is bailing out too early. We need to make sure that the pushers are notified of new receipts, so that notifications can be removed when messages are read on other servers.

richvdh (changes requested):

synapse/handlers/receipts.py
EdGeraghty: Yep, think you're correct.

EdGeraghty: I think you're right...

synapse/handlers/receipts.py
EdGeraghty: You're right. I don't think it should be bailing out at all, should it...

synapse/handlers/read_marker.py
EdGeraghty: Yep, shouldn't have put this here.

ShadowJonathan:
This PR doesn't look to be dropping read receipts as they arrive, so they're still processed and stored in the database, is that intentional?

EdGeraghty:
Is that not required in order to not make sure we don't end up with sticky "unread" notifications?

richvdh (commented):
Am I right in understanding that this means we will still send read receipts over federation? Is that deliberate? It seems a very curious choice that we would share read receipts with users on other servers, but not our own users.

Another thing: it looks like this means that we won't share read receipts between a user's own devices, so when the user switches device, he won't know where he's read up to.

EdGeraghty:
> Am I right in understanding that this means we will still send read receipts over federation? Is that deliberate?

No it's not. I'm guessing I've removed one too many returns?

> Another thing: it looks like this means that we won't share read receipts between a user's own devices, so when the user switches device, he won't know where he's read up to.

Which is literally the one behaviour I wasn't trying to stop.

I'll keep headbutting this wall for a little while...

anoadragon453:
Note that support for read receipt privacy at the per-account level in Matrix is currently being added: https://github.com/matrix-org/matrix-doc/pull/2285, and an implementation has already landed in Synapse: https://github.com/matrix-org/synapse/pull/10413

Which may be enough to solve the original problem proposed here?

EdGeraghty:
yep! Glad someone's done the hard bit for me :)

PR #10183, opened 2021-06-16
Add debug logging when we enter/exit Measure block

erikjohnston:
It can be helpful to know when trying to track down slow requests.

### DIFF
--- a/synapse/util/metrics.py
+++ b/synapse/util/metrics.py
@@ -133,11 +133,16 @@ class Measure:
         self.start = self.clock.time()
         self._logging_context.__enter__()
         in_flight.register((self.name,), self._update_in_flight)
+
+        logger.debug("Entering block %s", self.name)
+
         return self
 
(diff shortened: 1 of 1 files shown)


richvdh (approved):
I guess.

In the past DEBUG is a bit firehosey but still just about plausible - I worry that this takes it to the next level. it's possible to log at even lower levels than DEBUG (eg set `TRACE=5` and `logger.log(TRACE, "msg")` - is that worth considering?

erikjohnston:
> In the past DEBUG is a bit firehosey but still just about plausible - I worry that this takes it to the next level. it's possible to log at even lower levels than DEBUG (eg set `TRACE=5` and `logger.log(TRACE, "msg")` - is that worth considering?

Mm, possibly, though in some Measure blocks are kinda specifically saying we *do* care about resource usage for that block of code, so it sort of makes sense in my mind to log them. (I also don't think that we have that many Measure blocks compared with things like DB transactions and the like).

PR #8970, opened 2020-12-17
Allow re-using a UI auth validation for a period of time

clokep:
This should be reviewable commit-by-commit:

1. Start tracking the last time that an access token was "validated" (defined as logged in or performed UI auth).
2. Rename some config files.
3. Add a configuration option for re-using UI auth sessions.

I went with the configuration option since it seemed straightforward enough to implement. If we don't like that I can back this out to just the first commit.

Questions:
* Should we warn the user about this in some way? (I'm not sure if there's a way to, but this allows the client to do any operation for a period of time.)
* Is it worth trying to backfill the validation time for access tokens? (I don't think so, but let me know if you disagree.)

I also came across #8968 while doing this, but it isn't directly related.

Fixes #8912

PR #10184, opened 2021-06-16
Only allow skipping UI auth for certain actions.

### LINKED ISSUE #9754: Make UIA grace period only apply to cross-signing requests
Matrix.org's Synapse now applies a grace period to user-interactive-authenticated requests (the logic is that you do not need to confirm that you are the legitimate owner of an access token if you have literally just been given that access token).

However, various people have angrily reported that it is now too easy to deactivate your account by accident, since clients incorrectly assume that there will be a UIA step before the deactivation takes place.

As a workaround, it is suggested that the UIA grace period should only apply to E2E key uploads.

clokep:
This limits the UI auth grace period to:

* Deleting devices
* Uploading cross-signing keys

So it stops using it for (and thus always requires UI auth):

* Deactivating accounts
* Resetting passwords
* Adding 3PIDs

It should be easy to change the above list if people want to bikeshed it. (In particular adding 3PIDs I went back and forth on.)

This modifies behavior added in #8970 in a backwards-incompatible manner, but I think it is likely OK in this case as the security added outweighs backwards compatibility concerns.

Fixes #9754

### DIFF
--- a/changelog.d/10184.bugfix
+++ b/changelog.d/10184.bugfix
@@ -1,1 +1,1 @@
-Do not apply the `ui_auth.session_timeout` to dangerous operations: deactivating an account, modifying an account password, and adding 3PIDs.
+Always require users to re-authenticate for dangerous operations: deactivating an account, modifying an account password, and adding 3PIDs.
--- a/changelog.d/10184.bugfix
+++ b/changelog.d/10184.bugfix
@@ -1,0 +1,1 @@
+Do not apply the `ui_auth.session_timeout` to dangerous operations: deactivating an account, modifying an account password, and adding 3PIDs.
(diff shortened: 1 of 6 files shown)


changelog.d/10184.bugfix
```diff
@@ -0,0 +1 @@
+Do not apply the `ui_auth.session_timeout` to dangerous operations: deactivating an account, modifying an account password, and adding 3PIDs.
```
richvdh, line 1: I wonder if we can make this more focussed on the user-visible behaviour:

```suggestion
Always require users to re-authenticate for dangerous operations: deactivating an account, modifying an account password, and adding 3PIDs.
```

richvdh (approved):
lgtm other than nitpicking about the changelog.

richvdh:
for future reference: I believe the correct long-term fix is to focus our efforts on switching to OAuth2-based authentication ([MSC2964](https://github.com/matrix-org/matrix-doc/pull/2964) etc); this is a bit of a fudge.

Issue #9865, opened 2021-04-22
DNS failures connecting to IP address (!) for replication


richvdh:
We had a slew of this sort of error:

```
2021-04-22 10:12:31,676 - synapse.http.client - 453 - INFO - PUT-1883814-- - Error sending request to  POST http://10.103.0.1:8135/_synapse/replication/fed_send_edu/m.direct_to_device/cYKChokZoY: DNSLookupError Couldn't find the hostname '10.103.0.1'
```

There are two things of concern here:
 * we are doing DNS lookups for IP addresses. This is somewhat known, though highly unsatisfactory. It happens within `HostnameEndpoint` in twisted, which is used fairly widely, including the HTTP Agent.
 * the fact the DNS lookup failed.

richvdh:
Ah, I think what probably happened was that the request was *cancelled* after 60 seconds. `HostnameEndpoint` will, misleadingly, report a `DNSLookupError` if its `connect` method is cancelled before the DNS lookup completes (this is https://twistedmatrix.com/trac/ticket/9696).

Which leads us to... why did the DNS lookup take 60 seconds? I suspect the answer to this is that we got a bunch of other DNS lookups stacked up, to the point that the `_LimitedHostnameResolver` ended up taking more than 60seconds to get through its queue. In other words, this is basically the same as #7113.

richvdh:
I think we should just rip out `_LimitedHostnameResolver` and hope it (a) doesn't break anything and (b) fixes this and #7113.

richvdh:
> I think we should just rip out `_LimitedHostnameResolver` and hope it (a) doesn't break anything and (b) fixes this and #7113.

I'm not actually sure that it would. We'd still be subject to the concurrency limits of `GAIResolver` (which is backed by the reactor's default threadpool, which is limited to 10 threads.)

richvdh:
So I suspect the real solution is to fix https://twistedmatrix.com/trac/ticket/10181, and/or to install a `nameResolver` which spots literal IP addresses and fast-paths them.

clokep:
> So I suspect the real solution is to fix [twistedmatrix.com/trac/ticket/10181](https://twistedmatrix.com/trac/ticket/10181), and/or to install a `nameResolver` which spots literal IP addresses and fast-paths them.

We should be sure to double check this doesn't skip the IP address blacklists if we do this...although that's why we have the `BlacklsistingReactor`, I think. 😄

erikjohnston:
[...]

Issue #7113, opened 2020-03-20
Pusher stops pushing, with lots of DNSLookupFailed errors


richvdh:
Our pusher process got behind. A couple of hours after it started to catch up, almost all pushes started failing:

![image](https://user-images.githubusercontent.com/1389908/77156873-31206000-6a98-11ea-8855-6102545355fa.png)

Inspection of logs showed that they were failing with DNSLookupFailed errors:

```
2020-03-19 11:11:27,083 - synapse.http.client - 281 - INFO - httppush.process-57923 - Sending request POST http://10.103.0.7/_matrix/push/v1/notify
2020-03-19 11:12:28,647 - synapse.http.client - 328 - INFO - httppush.process-57923 - Error sending request to  POST http://10.103.0.7/_matrix/push/v1/notify: DNSLookupError Couldn't find the hostname '10.103.0.7'
2020-03-19 11:12:28,647 - synapse.push.httppusher - 391 - WARNING - httppush.process-57923 - Failed to push event <redacted> to  <redacted>: <class 'twisted.internet.error.DNSLookupError'> DNS lookup failed: Couldn't find the hostname '10.103.0.7'.
2020-03-19 11:12:32,225 - synapse.push.httppusher - 276 - INFO - httppush.process-57923 - Push failed: delaying for 1s
```

richvdh:
(it's worth pointing out that the matrix-org-hotfixes branch has this substitution to make it talk over http: https://github.com/matrix-org/synapse/blob/bc936b56572aef4fc16fffaa44a75bfdf99eeec4/synapse/push/httppusher.py#L106-L109)

richvdh:
so here's what's going on

* each HTTP hit has a timeout of 60seconds
* we limit the number of concurrent DNS lookups to 100
* we end up doing so many DNS lookups that the queue grows so long that the DNS lookup does not complete within 60 seconds
* so the request gets cancelled
* but the DNS lookup does not
* pretty soon we end up spending all our time doing DNS lookups whose results will get thrown away

and yes, we're doing DNS lookups despite the fact it's an IP address

It appears that we're using Twisted's [GAIResolver](https://github.com/twisted/twisted/blob/trunk/src/twisted/internet/_resolver.py#L71), which works via `getaddrinfo(3)` calls from a threadpool. (It remains a mystery to me when we use the GAIResolver and when we use `twisted.names`).

richvdh:
Some thoughts on what we could do about this:
[...]

PR #10190, opened 2021-06-16
Rip out the DNS lookup limiter

richvdh:
As I've written in various places in the past (#7113, #9865) I'm pretty sure this is doing nothing useful at all.

### DIFF
--- a/changelog.d/10190.misc
+++ b/changelog.d/10190.misc
@@ -1,0 +1,1 @@
+Remove redundant DNS lookup limiter.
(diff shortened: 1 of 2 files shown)


erikjohnston:
Can you include some more words why you don't think this is useful? It looks to me like using GAIResolver still causes FDs to open (to look up resolution config and then talk to the kernel via a socket, based on doing `socket.getaddrinfo` and stracing it).

richvdh:
GAIResolver does its work in the reactor's threadpool which is limited to 10 threads, so will never open more than 10 FDs at once.

(for background, GAIResolver calls [`socket.getaddrinfo`](https://docs.python.org/3/library/socket.html#socket.getaddrinfo), which calls glibc's `getaddrinfo`, which is [complicated](http://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/posix/getaddrinfo.c;h=b7e1aee80f68194a80fc8039d10b01e89d395a33;hb=HEAD) but takes care of reading `resolv.conf`, `hosts`, and whathaveyou. The point being that all of that happens within the constraints of the reactor's threadpool.)

erikjohnston (approved):

Issue #9778, opened 2021-04-09
Remove ACME support


anoadragon453:
Synapse currently only supports the ACMEv1 protocol, which has been deprecated by Let's Encrypt (the usual place one gets certs via ACME) [since 2020](https://community.letsencrypt.org/t/acme-v2-scheduled-deprecation-of-unauthenticated-resource-gets/74380). ACMEv2 is now required to request new certs and renew existing ones.

The library we use for ACME support, [txacme](https://github.com/twisted/txacme/issues/151), had implemented some promising work to support v2, but it is currently unfinished.

Advertising ACME support in Synapse currently is confusing to users. We should remove it for now (and potentially re-add it again in the future if a suitable ACMEv2 library surfaces).

babolivier:
Related: https://github.com/matrix-org/synapse/issues/6397

> ACMEv2 is now required to request new certs and renew existing ones.

This wasn't true at the date this comment was written. At the time, it was still possible for existing installs to use ACME v1 to get certificates, which is why it hasn't entirely been removed from Synapse yet. However it has been fully turned off a couple of weeks ago: https://community.letsencrypt.org/t/end-of-life-plan-for-acmev1/88430/27

> The library we use for ACME support, txacme, had implemented some promising work to support v2, but it is currently unfinished.

There still hasn't been any progress on txacme supporting ACMEv2 for over a year so I don't think this is likely to happen anytime soon. Plus there's a question of whether this is something we want to continue supporting as a core Synapse feature, to which our answer would lean towards the negative.

So all should be clear for killing it off. I'll open a PR to do this.

anoadragon453:
> Plus there's a question of whether this is something we want to continue supporting as a core Synapse feature, to which our answer would lean towards the negative.

I wasn't aware that we were against having it as a feature if txacme implemented support. Though I do agree that it's best to remove it given txacme's current state.

babolivier:
> I wasn't aware that we were against having it as a feature if txacme implemented support. Though I do agree that it's best to remove it given txacme's current state.

In general I think we want to lean more towards telling users to use a reverse proxy for this kind of things, rather than clobbering Synapse with features like this. iirc that's the reason we decided to remove the feature once ACMEv1 fully reaches EOL, back when we were investigating what to do about its deprecation.

PR #10194, opened 2021-06-17
Remove support for ACME v1

babolivier:
Fixes https://github.com/matrix-org/synapse/issues/9778

ACME v1 has been fully decommissioned for existing installs on June 1st 2021(see https://community.letsencrypt.org/t/end-of-life-plan-for-acmev1/88430/27), so we can now safely remove it from Synapse.


synapse/config/tls.py
```diff
@@ -428,11 +302,7 @@ def read_tls_certificate(self) -> crypto.X509:
         # Check if it is self-signed, and issue a warning if so.
         if cert.get_issuer() == cert.get_subject():
             warnings.warn(
-                (
-                    "Self-signed TLS certificates will not be accepted by Synapse 1.0. "
-                    "Please either provide a valid certificate, or use Synapse's ACME "
-                    "support to provision one."
-                )
+                "Self-signed TLS certificates will not be accepted by Synapse 1.0."
             )
```
richvdh, line 306: I think I'd get rid of all this. It was mostly there to warn people before 1.0 was released.

richvdh (approved):
feels good!

synapse/config/tls.py
```diff
@@ -299,12 +299,6 @@ def read_tls_certificate(self) -> crypto.X509:
         cert_pem = self.read_file(cert_path, "tls_certificate_path")
         cert = crypto.load_certificate(crypto.FILETYPE_PEM, cert_pem)
 
-        # Check if it is self-signed, and issue a warning if so.
```
richvdh, line 428: there's some incorrect text in the docstring which needs an update now.

babolivier: Oh oops

PR #10198, opened 2021-06-17
Deploy a documentation version for each new Synapse release

anoadragon453:
This PR will run a new "Deploy release-specific documentation" job whenever a push to a branch name matching `release-v*` occurs. Doing so will create/add to a folder named `vX.Y` on the `gh-pages` branch. Doing so will allow us to build up `major.minor` releases of the docs as we release Synapse.

This is especially useful for having a mechanism for keeping around documentation of old/removed features (for those running older versions of Synapse), without needing to clutter the latest copy of the docs.

After a [discussion](https://matrix.to/#/!XaqDhxuTIlvldquJaV:matrix.org/$rKmkBmQle8OwTlGcoyu0BkcWXdnHW3_oap8BMgclwIY?via=matrix.org&via=vector.modular.im&via=envs.net) in #synapse-dev, we wanted to use tags to trigger the documentation deployments, which I agreed with. However, I soon realised that the bash-foo required to turn a tag of `v1.2.3rc1` into `1.2` was a lot more complex than the branch's `release-v1.2`. So, I've gone with the latter for simplicity.

In the future we'll have some UI on the website to switch between versions, but for now you can simply just change 'develop' to 'v1.2' in the URL.

You can see the workflow in action here: https://github.com/anoadragon453/synapse/runs/2850612850

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -3,7 +3,10 @@ name: Deploy the documentation
 on:
   push:
     branches:
+      # For bleeding-edge documentation
       - develop
+      # For documentation specific to a release
+      - 'release-v*'
 
   workflow_dispatch:
 
@@ -22,6 +25,7 @@ jobs:
       - name: Build the documentation
         run: mdbook build
 
+      # Deploy to the latest documentation directories
       - name: Deploy latest documentation
         uses: peaceiris/actions-gh-pages@068dc23d9710f1ba62e86896f84735d869951305 # v3.8.0
         with:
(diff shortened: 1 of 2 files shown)


richvdh (approved):
lgtm

anoadragon453:
After a look-over, it turned out that the workflow failed to properly check for release candidates versus full releases (we were checking for the presence of `rc` in a branch name; something we never have).

The latest commit fixes this by once again checking Synapse's `__version__` str.

.github/workflows/docs.yaml
```diff
@@ -37,9 +37,15 @@ jobs:
       - name: Get the current Synapse version
         id: vars
```
clokep, line 38: I think you'll need a `shell: bash` to do the bash magic?

anoadragon453: Hmm, good point. Turns out this works in POSIX shells as well. Will edit the wording.

PR #10205, opened 2021-06-18
Add support for evicting cache entries based on last access time.

erikjohnston:
This adds ~100-150B memory overhead per cache entry, which is annoying. We could reduce that a little bit by removing the `last_access_ts` field and instead periodically inserting a dummy list node that *does* include the the ts it was inserted at, which would allow us to just evict any entries after the dummy list node. I don't think the complexity there is particularly warranted though.

~~Unit tests are incoming.~~

Reviewable commit-by-commit, hopefully.


erikjohnston:
(Sorry, implementing the unit tests makes me realise I need to rejig this a bit to make it use a reactor clock)

erikjohnston:
The performance impact of this change, compared with develop:

| Operation | Develop | Time eviction disabled | Time eviction enabled |
|---------------|------------------------|----------------------------|--|
| Set | 1.5 µs | 2.9 µs | 5.8µs |
| Set (causing eviction) | 1.7 µs | 3.2 µs | 5.6 µs |
| Get an existing item | 358 ns | 365 ns | 364 ns |

So there is definitely some overhead, particularly when inserting items.

synapse/util/caches/lrucache.py
```diff
@@ -484,3 +750,8 @@ def set_cache_factor(self, factor: float) -> bool:
                 self._on_resize()
             return True
         return False
+
+    def __del__(self) -> None:
+        # We're about to be deleted, so we make sure to clear up all the nodes
+        # and run callbacks, etc.
+        self.clear()
```
richvdh, line 757: not saying this is a bad thing, but what is the thinking here? Do we expect LruCache objects to get deleted in operation, or is this just for tests?

synapse/util/caches/lrucache.py
```diff
@@ -82,19 +93,243 @@ def enumerate_leaves(node, depth):
...
+
+    def __init__(self, cache_entry: Optional[P] = None) -> None:
+        self.cache_entry = cache_entry
+        self.prev_node: Optional[_ListNode[P]] = self
+        self.next_node: Optional[_ListNode[P]] = self
+
+    @staticmethod
+    def insert_after(
+        cache_entry: P, root: "_ListNode[P]", clock: Clock
+    ) -> "_ListNode[P]":
```
richvdh, line 136: I find the parameter ordering here a bit unintuitive. I'd expect `root` and `cache_entry` to be the other way around.

synapse/util/caches/lrucache.py
```diff
@@ -82,19 +93,243 @@ def enumerate_leaves(node, depth):
...
+    def drop_from_cache(self) -> None:
+        """Request the entry be dropped from the cache.
+
+        Note: This should call `.remove_from_list()` on all `_ListNodes`
+        referencing it.
+        """
+        ...
+
+
+P = TypeVar("P", bound=_CacheEntry)
```
richvdh, line 110: given that `_ListNode` doesn't call `drop_from_cache`, why is this bound necessary?

synapse/util/caches/lrucache.py
```diff
@@ -82,19 +93,243 @@ def enumerate_leaves(node, depth):
...
+    while node is not GLOBAL_ROOT:
+        # Only the root node isn't a `_TimedListNode`.
+        assert isinstance(node, _TimedListNode)
+
+        if node.last_access_ts_secs > now - expiry_seconds:
+            break
+
+        cache_entry = node.get_cache_entry()
+        current_node = node
+        node = node.prev_node
```
richvdh, line 250: these are confusing names. I suggest:
```suggestion
        next_node = node.prev_node
```

... and then do `node=next_node` at the bottom of the loop.

synapse/util/caches/lrucache.py
```diff
@@ -82,19 +93,243 @@ def enumerate_leaves(node, depth):
...
+        if cache_entry:
+            cache_entry.drop_from_cache()
+        else:
+            # The cache entry has been dropped without being cleared out of this
+            # list. This can happen if the `LruCache` has been dropped without
+            # being cleared up properly.
```
richvdh, line 256: ... and how does that happen?

synapse/util/caches/lrucache.py
```diff
@@ -82,19 +93,243 @@ def enumerate_leaves(node, depth):
...
+        """Create a new list node that is placed after the given root node."""
+        node = _ListNode(cache_entry)
+        node.move_after(root, clock)
+        return node
+
+    def remove_from_list(self):
+        """Remove this node from the list."""
+        if self.prev_node is None or self.next_node is None:
+            # We've already been removed from the list.
+            return
```
richvdh, line 146: I'd kinda be more comfortable if this raised an exception; it seems like an error condition to me.

synapse/util/caches/lrucache.py
```diff
@@ -219,20 +482,34 @@ def __init__(
...
         def evict():
+            orphaned = 0
             while cache_len() > self.max_size:
                 todelete = list_root.prev_node
-                evicted_len = delete_node(todelete)
-                cache.pop(todelete.key, None)
+                assert todelete is not None
+
+                node = todelete.get_cache_entry()
+                if not node:
```
richvdh, line 500: comments about how this happens might be good.

synapse/util/caches/lrucache.py
```diff
@@ -82,19 +93,243 @@ def enumerate_leaves(node, depth):
...
+        next_node.prev_node = self
+
+    def get_cache_entry(self) -> Optional[P]:
+        """Get the cache entry, returns None if this is the root node (i.e.
+        cache_entry is None) or if the entry has been dropped.
+        """
+        return self.cache_entry
+
+
+class _TimedListNode(_ListNode[P]):
```
richvdh, line 195: Could this *contain* a `_ListNode` rather than extend it? Having `_ListNode.move_after` take a `clock` is pretty ugly, and afaict the only places you call `move_after` or `insert_after` you know whether you have a `_ListNode` or a `_TimedListNode`.

richvdh (changes requested):
Seems good in principle, but I have some opinions.

synapse/util/caches/lrucache.py
```diff
@@ -219,20 +482,34 @@ def __init__(
         # this is exposed for access from outside this class
         self.metrics = metrics
 
-        list_root = _Node(None, None, None, None)
-        list_root.next_node = list_root
-        list_root.prev_node = list_root
+        # We create a single weakref to self here, so that all `_Node` can share
+        # a single reference (as weakrefs are surprisingly large).
```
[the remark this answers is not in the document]
erikjohnston: Ooooh, interesting. I did see memory usage go down when removing repeated weakrefs, but I guess that was actually because we kept adding new finalizers every time we took a new ref (originally I added a new finalizer to call `_Node.remove_from_list` for each node).

synapse/util/caches/lrucache.py
```diff
@@ -82,19 +93,243 @@ def enumerate_leaves(node, depth):
...
-    __slots__ = ["prev_node", "next_node", "key", "value", "callbacks", "memory"]
+    __slots__ = [
+        "list_node",
+        "global_list_node",
+        "cache",
+        "key",
+        "value",
+        "callbacks",
+        "memory",
+        "__weakref__",
```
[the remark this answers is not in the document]
erikjohnston: Ah, this is now redundant as I stopped using a weakref in `_ListNode` for `_Node`

synapse/util/caches/lrucache.py
erikjohnston: OMG this took ages to track down, but in the sync code we have an expiring cache of lru caches: 

https://github.com/matrix-org/synapse/blob/fcf3c7032b96ab454120f86f3f070160c409d599/synapse/handlers/sync.py#L280-L286

which led to all sorts of fun and games with orphaned nodes.

synapse/util/caches/lrucache.py
erikjohnston: It's called in `_expire_old_entries`, which iterates over the `_ListNode`?

synapse/util/caches/lrucache.py
erikjohnston: Hmm, I guess now we're not using weakrefs it shouldn't actually happen. I guess we should still do the same thing but maybe log an error?

synapse/util/caches/lrucache.py
erikjohnston: Mmm, maybe. My thinking was that a) calling `remove_from_list` multiple times doesn't seem wrong if we're trying to make damn sure that we remove list node and b) I could see a situation where there's a race between the time based eviction and standard eviction that might cause this.

synapse/util/caches/lrucache.py
erikjohnston: Mmm, true. Though I think extending it has a lower memory footprint:

```python
In [27]: class A(): 
    ...:     __slots__ = [] 
    ...:                                                                                                                                                                                                          

In [28]: class B(A): 
    ...:     __slots__ = [] 
    ...:                                                                                                                                                                                                          

In [29]: class C: 
    ...:     __slots__ = ["a"] 
    ...:                                                                                                                                                                                                          

In [30]: asizeof(A())                                                                                                                                                                                             
Out[30]: 16

In [31]: asizeof(B())                                                                                                                                                                                             
Out[31]: 16

In [32]: c = C()                                                                                                                                                                                                  

In [33]: c.a = A()                                                                                                                                                                                                

In [34]: asizeof(c)                                                                                                                                                                                               
Out[34]: 56
```

synapse/config/cache.py
```diff
@@ -145,6 +145,12 @@ def generate_config_section(self, **kwargs):
            #
            per_cache_factors:
              #get_users_who_share_room_with_user: 2.0
+
+            # Controls how long an entry can be in a cache without having been
+            # accessed before being evicted. Defaults to None, which means
+            # entries are never evicted based on time.
+            #
+            #expiry_time: 30m
```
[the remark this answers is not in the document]
erikjohnston: Oh, eww, the cache config is using a three space indent.

synapse/util/caches/lrucache.py
```diff
@@ -116,11 +351,16 @@ def __init__(
             self.memory = (
                 _get_size_of(key)
                 + _get_size_of(value)
+                + _get_size_of(self.list_node, recurse=False)
                 + _get_size_of(self.callbacks, recurse=False)
                 + _get_size_of(self, recurse=False)
```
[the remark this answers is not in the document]
erikjohnston: The cache weakref should be shared across all the nodes, so I don't think we should? The extra attribute in `_ListNode` gets counted as part of `_get_size_of(self)`?

synapse/util/caches/lrucache.py
richvdh: riiight, but doesn't `_expire_old_entries` know it's a `_ListNode[_CacheEntry]` ?

erikjohnston: Oh I see what you mean, yes you're right.

erikjohnston: Turns out we can get rid of `_CacheEntry` entirely.

synapse/util/caches/lrucache.py
richvdh: True enough I guess. Does 40 bytes per entry matter that much (it seems like this is the only thing stopping `_ListNode` being a generic DLL rather than anything specific to caches)?

erikjohnston: I don't think its the end of the world, though I'm a bit wary we have quite a few caches that are e.g. `str -> bool`, where the overhead of metadata really starts adding up.


> Could this _contain_ a `_ListNode` rather than extend it? Having `_ListNode.move_after` take a `clock` is pretty ugly, and afaict the only places you call `move_after` or `insert_after` you know whether you have a `_ListNode` or a `_TimedListNode`.

FWIW I shuffled things around such that `_ListNode` no longer takes a clock.

richvdh: I got thoroughly nerd-sniped by this.

An object with `__slots__` needs:
 * 8 bytes for a refcount
 * 8 bytes for a pointer to the type def
 * 8 bytes for each slot
 * 16 bytes for the garbage collector (GC-able objects form a double-linked-list, so this is a pointer in each direction)

- total 40 bytes for an object with one slot.

An object with *no* slots is special-cased: it does not need to partake in the GC protocol (since it can never reference other objects), hence 16 bytes total.

A `__slots__` object which inherits from a single other `__slots__` object can do it by just adding its own slots to the end of the list of its parents slots - hence just 8 bytes for each extra slot.

synapse/config/cache.py
```diff
@@ -116,35 +116,35 @@ def generate_config_section(self, **kwargs):
...
-           #
-           # Some caches have '*' and other characters that are not
-           # alphanumeric or underscores. These caches can be named with or
-           # without the special characters stripped. For example, to specify
-           # the cache factor for `*stateGroupCache*` via an environment
-           # variable would be `SYNAPSE_CACHE_FACTOR_STATEGROUPCACHE=2.0`.
-           #
-           per_cache_factors:
-             #get_users_who_share_room_with_user: 2.0
+            # Controls the global cache factor, which is the default cache factor
```
richvdh, line 119: style guide says "use two-space indents".

synapse/util/caches/lrucache.py
```diff
@@ -127,20 +118,56 @@ class _ListNode(Generic[P]):
...
-    ) -> "_ListNode[P]":
-        """Create a new list node that is placed after the given root node."""
-        node = _ListNode(cache_entry)
-        node.move_after(root, clock)
-        return node
+        cls: Type[LN],
+        cache_entry: P,
+        node: "_ListNode[P]",
+    ) -> LN:
+        """Create a new list node that is placed after the given node."""
```
richvdh, line 137: some docstrings for the params would be useful, I think.

synapse/util/caches/lrucache.py
```diff
@@ -82,19 +93,241 @@ def enumerate_leaves(node, depth):
...
+        "next_node",
+    ]
+
+    def __init__(self, cache_entry: Optional[P] = None) -> None:
+        self.cache_entry = cache_entry
+        self.prev_node: Optional[_ListNode[P]] = None
+        self.next_node: Optional[_ListNode[P]] = None
+
+    @classmethod
+    def create_root_node(cls: Type["_ListNode[P]"]) -> "_ListNode[P]":
```
richvdh, line 125: docstring please?

synapse/util/caches/lrucache.py
```diff
@@ -82,19 +93,241 @@ def enumerate_leaves(node, depth):
...
+        return new_node
+
+    def remove_from_list(self):
+        """Remove this node from the list."""
+        with self._LOCK:
+            self._refs_remove_node_from_list()
+
+        self.cache_entry = None
+
+    def move_after(self, root: "_ListNode"):
```
richvdh, line 150: can we rename `root` to `node` or `other` or `target` or something? It's not limited to being the root node afaict.

synapse/util/caches/lrucache.py
```diff
@@ -82,19 +93,241 @@ def enumerate_leaves(node, depth):
...
+        with cls._LOCK:
+            new_node._refs_insert_after(node)
+        return new_node
+
+    def remove_from_list(self):
+        """Remove this node from the list."""
+        with self._LOCK:
+            self._refs_remove_node_from_list()
+
+        self.cache_entry = None
```
richvdh, line 148: a comment to explain why we do this might be good.

synapse/util/caches/lrucache.py
```diff
@@ -82,19 +93,241 @@ def enumerate_leaves(node, depth):
...
+        """Move this node from its current location in the list to after the
+        given node.
+        """
+        with self._LOCK:
+            # We assert that both this node and the root node is still "alive".
+            assert self.prev_node
+            assert self.next_node
+            assert root.prev_node
+            assert root.next_node
+
```
richvdh, line 160:
```suggestion
            assert self is not root

```

synapse/util/caches/lrucache.py
```diff
@@ -82,19 +93,241 @@ def enumerate_leaves(node, depth):
                 yield m
 
 
+P = TypeVar("P")
+LN = TypeVar("LN", bound="_ListNode")
+
+
+class _ListNode(Generic[P]):
```
richvdh, line 100: This thing is now a completely generic DLL implementation. I would argue strongly in favour of moving it out of the `caches` package.

synapse/util/caches/lrucache.py
```diff
@@ -82,19 +93,241 @@ def enumerate_leaves(node, depth):
...
+    logger.info(
+        "Expiring LRU caches after %d seconds", hs.config.caches.expiry_time_msec / 1000
+    )
+
+    global USE_GLOBAL_LIST
+    USE_GLOBAL_LIST = True
+
+    clock = hs.get_clock()
+    clock.looping_call(
+        _expire_old_entries, 30 * 1000, clock, hs.config.caches.expiry_time_msec / 1000
```
richvdh, line 296: is the initial 30 second wait arbitrary?

synapse/util/caches/lrucache.py
```diff
@@ -82,19 +93,241 @@ def enumerate_leaves(node, depth):
...
+    )
+
+
 class _Node:
-    __slots__ = ["prev_node", "next_node", "key", "value", "callbacks", "memory"]
+    __slots__ = [
+        "_list_node",
```
richvdh, line 302: given that this is no longer a linked list node, I wonder if we should consider renaming to `_LruCacheEntry` or something.

That said, it is referred to as `node` in about 1567 places, and you'd want to change all of those too, so maybe best leave it for now.

richvdh (changes requested):

synapse/util/caches/lrucache.py
erikjohnston: It's every 30 seconds, but yes its just an arbitrary period I picked. We want it to happen relatively frequently so we don't expire lots all at once, but not too frequently that there is wasted cycles.

synapse/util/caches/lrucache.py
erikjohnston: Err, true, though yeah I'd rather punt that for now.

synapse/util/linked_list.py
```diff
@@ -0,0 +1,150 @@
...
+        # between the list node and cache entry, allowing the two to be dropped
+        # immediately rather than at the next GC.
+        self.cache_entry = None
+
+    def move_after(self, node: "ListNode"):
+        """Move this node from its current location in the list to after the
+        given node.
+        """
+        with self._LOCK:
+            # We assert that both this node and the root node is still "alive".
```
richvdh, line 92:
```suggestion
            # We assert that both this node and the target node is still "alive".
```

synapse/util/linked_list.py
```diff
@@ -0,0 +1,150 @@
...
+            assert self.next_node
+            assert node.prev_node
+            assert node.next_node
+
+            assert self is not node
+
+            # Remove self from the list
+            self._refs_remove_node_from_list()
+
+            # Insert self back into the list, after root
```
richvdh, line 103:
```suggestion
            # Insert self back into the list, after target node
```

richvdh (approved):
lgtm otherwise

PR #10206, opened 2021-06-18
Describe callbacks signatures as async in new modules doc

babolivier:
I've just realised that after merging #10062. The doc already says callbacks must be async but it's probably better to make it as obvious as possible.


clokep (approved):
Definitely good to be explicit!

PR #10208, opened 2021-06-18
Fix a missing await when in the spaces summary

clokep:
* Adds a couple of unit tests.
* Fixes a missing `await` in the spaces summary API. (This was introduced in #10109.)

### DIFF
--- a/changelog.d/10208.bugfix
+++ b/changelog.d/10208.bugfix
@@ -1,0 +1,1 @@
+Fix a bug introduced in v1.35.1 where an `allow` key of a `m.room.join_rules` event could be applied for incorrect room versions and configurations.
--- a/synapse/handlers/space_summary.py
+++ b/synapse/handlers/space_summary.py
@@ -452,7 +452,7 @@ class SpaceSummaryHandler:
                     return True
 
             # Otherwise, check if they should be allowed access via membership in a space.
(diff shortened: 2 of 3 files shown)


changelog.d/10208.bugfix
```diff
@@ -0,0 +1 @@
+Fix a bug introduced in v1.35.1 where an `allow` key of a `m.room.join_rules` event could be applied for incorrect room versions and configurations.
```
clokep, line 1: I'm not sure this explains it well enough, but not awaiting there it means that you could have a room with `allow` set in the join rules, but a join rule of *not* `restricted` and those allow rules would still be applied.

Realistically I don't think this is user visible unless someone was messing with join rules in a weird way, although it means you could slightly leak data if you did this.

anoadragon453: If this was more user-visible, I'd probably suggest rephrasing "an `allow` key of a `m.room.join_rules` event could be applied" with something more user-digestible, but as it's not it's likely more useful to be clear about what exactly the missing await affected. In which case, to a developer, this is very clear :slightly_smiling_face:

anoadragon453 (approved):
lgtm, and thanks for adding a test. It was also very useful to see the test built up over a series of commits :slightly_smiling_face:

clokep:
Thanks! 🎉

PR #10210, opened 2021-06-18
Warn users trying to use the deprecated spam checker interface

babolivier:
I forgot to do that in #10062 

So admins aren't surprised if things break when we remove this code in a couple of months.


babolivier:
(adding the release blocker label since this is a small change that's an omission from a bigger one that's supposed to go into the next release)

erikjohnston (approved):

PR #10213, opened 2021-06-18
Add additional types to the federation transport server

clokep:
This is a follow-up to #10080 which adds additional type hints.


clokep:
This has two mypy errors:

```
synapse/federation/transport/server.py:584: error: Statement is unreachable  [unreachable]
synapse/federation/transport/server.py:915: error: Statement is unreachable  [unreachable]
```

Which seem to stem from changes made in #10080, but I'm unsure why the overload there isn't working (it seems that mypy thinks `parse_string_from_args` always returns a `str`). Any help would be appreciated!

erikjohnston:
So, this was due to: 

https://github.com/matrix-org/synapse/blob/60336cd440da587419004c9212c166451a9aa7ca/synapse/http/servlet.py#L302

where in one overload we specified the default of required to be `True`, when the default is actually `False`. Unfortunately you can't just do `required: Literal[True] = False` as mypy complains that True isn't False. To fix this we basically need to create a new overload where the `required` isn't specified and annotate that correctly, then remove the default values from the other overrides, which in turn requires adding `*` to make it work. c.f. https://github.com/python/mypy/issues/7333

All in all its a bit of a mess but I think it works now.

synapse/federation/transport/server.py
```diff
@@ -649,7 +649,9 @@ async def on_GET(
     ) -> Tuple[int, JsonDict]:
         try:
             # Retrieve the room versions the remote homeserver claims to support
-            supported_versions = parse_strings_from_args(query, "ver", encoding="utf-8")
+            supported_versions = parse_strings_from_args(
+                query, "ver", required=True, encoding="utf-8"
+            )
```
erikjohnston, line 654: This is meant to be a required parameter, and if its not supplied then the code will explode later anyway.

clokep:
Thanks @erikjohnston! 🎉

synapse/federation/transport/server.py
clokep: Good shout. We can actually remove the try-except in this case too!

erikjohnston (approved):

Issue #9941, opened 2021-05-06
Re-introduce "Leave out optional keys from /sync"


callahad:
The changes from #9919 were backed out in #9940 as it broke notifications on Element iOS.

We should re-introduce #9919 as soon as we're able.

Blockers:

- [x] Fix Element iOS vector-im/element-ios#4309
- [x] Wait for a fixed Element iOS release to gain sufficient adoption

richvdh:
I think we should set a target on doing this work, otherwise it's never going to happen.

@manuroe: I'm going to assume it is realistic that we will be able to do this in three months (ie, by 6th August). Please let us know if you think that is too ambitious.

manuroe:
The fix will be available in the release we will make in 2 weeks. 95% of people update their app within 1 week.

deepbluev7:
Also related: https://github.com/vector-im/hydrogen-web/pull/343 https://github.com/poljar/matrix-nio/commit/6097f2c48fca0c3802199c4998dc4883b513e758

Although those are merged already at least.

deepbluev7:
FTR, this is now fixed in iOS. Not released yet though.

https://github.com/matrix-org/matrix-ios-sdk/pull/1100

tulir:
The release seems to have hit app store 9 days ago, so I guess 95% of people should have updated by now. Can we get this into 1.37? :3

PR #10214, opened 2021-06-18
Re-introduce "Leave out optional keys from /sync" change

anoadragon453:
Closes: https://github.com/matrix-org/synapse/issues/9941

Required some fixes due to merge conflicts with #6739, but nothing too hairy. The first commit is the same as the original (after merge conflict resolution) then two more for compatibility with the latest sync code.

cc @callahad

### DIFF
--- a/changelog.d/9919.feature
+++ b/changelog.d/9919.feature
@@ -1,1 +1,0 @@
-Omit empty fields from the `/sync` response.  Contributed by @deepbluev7.
--- a/changelog.d/9919.feature
+++ b/changelog.d/9919.feature
@@ -1,0 +1,1 @@
+Omit empty fields from the `/sync` response.  Contributed by @deepbluev7.
--- a/tests/server_notices/test_resource_limits_server_notices.py
+++ b/tests/server_notices/test_resource_limits_server_notices.py
@@ -306,8 +306,9 @@ class TestResourceLimitsServerNoticesWithRealRooms(unittest.HomeserverTestCase):
 
         channel = self.make_request("GET", "/sync?timeout=0", access_token=tok)
 
-        invites = channel.json_body["rooms"]["invite"]
-        self.assertEqual(len(invites), 0, invites)
+        self.assertNotIn(
+            "rooms", channel.json_body, "Got invites without server notice"
+        )
 
     def test_invite_with_notice(self):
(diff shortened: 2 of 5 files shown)


changelog.d/9919.feature
```diff
@@ -0,0 +1 @@
+Omit empty fields from the `/sync` response.  Contributed by @deepbluev7.
```
richvdh, line 1: I think we should omit this.

tests/server_notices/test_resource_limits_server_notices.py
```diff
@@ -321,7 +327,9 @@ def test_invite_with_notice(self):
 
         # Scan the events in the room to search for a message from the server notices
         # user.
-        events = channel.json_body["rooms"]["join"][room_id]["timeline"]["events"]
+        events = channel.json_body["rooms"][Membership.JOIN][room_id]["timeline"][
```
richvdh, line 330: as a matter of general preference, I would avoid using symbolic constants like this in tests where possible. It's a chance to spot typos in the definition of the constant!

richvdh (approved):
lgtm modulo the duplicate changelog

tests/server_notices/test_resource_limits_server_notices.py
anoadragon453: That's a good point! I also did so when adding the tests for knocking, though that made a bit more sense when we needed to switch the prefix from unstable to stable.

But a good tactic to carry forward.

erdnaxeli:
I just want you to notice that this can broke lib :grimacing: 

Actually the spec is not helpfull to know which fields will always be there and which are optional. Sometimes the field is required and this is specified, like in the response of `/login`:
>  m.homeserver | Homeserver Information | Required. Used by clients to discover homeserver information.

Sometimes the field is optional and this is specified, like in the description on `UnsignedData`:
> redacted_because | Event | Optional. The event that redacted this event, if any.

But often there is just nothing, and we don't know. The specs should specify at the start of the document a default behavior, and then remove the useless precision ("required" or "optional", depending on the default).

ShadowJonathan:
All fields in `/sync` are optional, except next_batch, it's just that synapse has always returned them by default.

erdnaxeli:
Well, as I was saying, the spec does not make it clear.

PR #10221, opened 2021-06-21
Fix performance of responding to user key requests over federation

erikjohnston:
We were repeatedly looking up a config option in a loop (using the
unclassed config style), which is expensive enough that it can cause
large CPU usage.

### DIFF
--- a/changelog.d/10221.bugfix
+++ b/changelog.d/10221.bugfix
@@ -1,1 +1,1 @@
-Fix performance of responding to user key requests over federation. Introduced in v1.34.0.
+Fix performance of responding to user key requests over federation. Introduced in v1.34.0rc1.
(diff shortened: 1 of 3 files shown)


richvdh (commented):
looks like this was introduced in https://github.com/matrix-org/synapse/pull/9945 ? can you put that in the changelog pls

changelog.d/10221.bugfix
```diff
@@ -0,0 +1 @@
+Fix performance of responding to user key requests over federation. Introduced in v1.34.0rc1.
```
richvdh, line 1:
```suggestion
Fix performance regression in responding to user key requests over federation. Introduced in v1.34.0rc1.
```

richvdh (approved):

PR #10225, opened 2021-06-21
Improve validation for `send_{join,leave,knock}`

richvdh:
The idea here is to stop people sending things that aren't joins/leaves/knocks through these endpoints: previously you could send anything you liked through them. I wasn't able to find any security holes from doing so, but it doesn't sound like a good thing.

I've taken the opportunity to merge the three almost identical codepaths. Commits best reviewed separately.

### DIFF
--- a/changelog.d/10225.feature
+++ b/changelog.d/10225.feature
@@ -1,0 +1,1 @@
+Improve validation on federation `send_{join,leave,knock}` endpoints.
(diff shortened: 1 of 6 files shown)


richvdh:
tests in https://github.com/matrix-org/complement/pull/131

erikjohnston (approved):

Issue #10233, opened 2021-06-23
Auto-joining a remote autojoin room doesn't work reliably


benbz:
### Description

Users registering on HS B can't auto-join a room on HS A that's listed in `auto_join_rooms` if they are either the 1st user on HS B or if there are no users on HS B that are in the room on HS A.

### Steps to reproduce

- HS A has a public, federatable room with an alias
- HS B sets `auto_join_rooms` to include the room on HS A
- If there are no existing users on HS B:
  - 1st user registers
  - Auto-joining skipped with `Cannot create room alias #HS-A:room-alias it does not match server domain`
- If there are existing users on HS B and no-one on HS B is in this room
  - Registrations after setting `auto_join_rooms` as per the above fail with `Failed to join new user to '#HS-A:room-alias': AssertionError()`
  - This repeats until someone manually joins the room on HS A
  - I believe this is because if there's no local copy of the state, then an invite is assumed - https://github.com/matrix-org/synapse/blob/1c8045f67477599fabc5759205c018e44d770078/synapse/handlers/register.py (Lines 453-471)

The expectation is that `auto_join_rooms` works with any public room local or federated.

### Version information
- **Homeserver**: Any 2 publicly federated servers
- **Version**: 1.36.0
- **Install method**: Docker
- **Platform**: Linux

erikjohnston:
I think https://github.com/matrix-org/synapse/blob/1c8045f67477599fabc5759205c018e44d770078/synapse/handlers/register.py#L455-L469 is the problematic logic here. I think we can fix it by adding a check to see if the host is in the room and the alias is remote.

We also need to handle the case where auto creating auto join is enabled but the alias is remote:
https://github.com/matrix-org/synapse/blob/1c8045f67477599fabc5759205c018e44d770078/synapse/handlers/register.py#L388-L393

PR #10237, opened 2021-06-23
Improve the reliability of auto-joining remote rooms

babolivier:
If a room is remote and we don't have a user in it, always try to join it. It might fail if the room is invite-only, but we don't have a user to invite with, so at this point it's the best we can do.

Fixes https://github.com/matrix-org/synapse/issues/10233 (at least to some extent)


synapse/handlers/register.py
```diff
@@ -448,9 +463,15 @@ async def _join_rooms(self, user_id: str) -> None:
...
-                requires_invite = True
+                # joined directly. By default, we consider the room as requiring an
+                # invite if the homeserver is in the room (unless told otherwise by the
+                # join rules). Otherwise we consider it as being joinable, at the risk of
+                # failing to join, but in this case there's little more we can do since
+                # we don't have a local user in the room to craft up an invite with.
+                hosts_in_room = await self.state_handler.get_current_hosts_in_room(
+                    room_id
+                )
+                requires_invite = self.server_name in hosts_in_room
```
erikjohnston, line 480: I'd use `store.is_host_joined` to avoid have to do state res here (in general using store is better than state handler), plus its got a cache on it.

synapse/handlers/register.py
```diff
@@ -448,9 +463,15 @@ async def _join_rooms(self, user_id: str) -> None:
...
 
                 state = await self.store.get_filtered_current_state_ids(
                     room_id, StateFilter.from_types([(EventTypes.JoinRules, "")])
```
erikjohnston, line 483: If we're not in the room then we may as well not do this check, so I'd probably put it in an `if`

synapse/handlers/register.py
```diff
@@ -386,11 +386,32 @@ async def _create_and_join_rooms(self, user_id: str) -> None:
                 room_alias = RoomAlias.from_string(r)
 
                 if self.hs.hostname != room_alias.domain:
+                    # If the alias is remote, try to join the room. This might fail
+                    # because the room might be invite only, but we don't have any local
+                    # user in the room to invite this one with, so at this point that's
+                    # the best we can do.
                     logger.warning(
```
anoadragon453, line 393:
```suggestion
                    logger.info(
```

Since it's no longer a failure mode.

synapse/handlers/register.py
```diff
@@ -386,11 +386,32 @@ async def _create_and_join_rooms(self, user_id: str) -> None:
...
-                        "it does not match server domain",
+                        "Cannot automatically create room with alias %s as it isn't"
+                        " local, trying to join the room instead",
                         r,
                     )
+
+                    (
+                        room,
+                        remote_room_hosts,
+                    ) = await room_member_handler.lookup_room_alias(room_alias)
```
anoadragon453, line 402: This method raising a `SynapseError` during registration would probably be quite bad. We probably want to wrap it in a `try/except`, and log on failure to join.

erikjohnston (changes requested):
I think this works as is though

erikjohnston:
(Tests would also be nice :innocent: )

synapse/handlers/register.py
```diff
@@ -465,6 +492,13 @@ async def _join_rooms(self, user_id: str) -> None:
                         join_rule = join_rules_event.content.get("join_rule", None)
                         requires_invite = join_rule and join_rule != JoinRules.PUBLIC
 
+                        if requires_invite and self.server_name not in hosts_in_room:
+                            raise SynapseError(
+                                400,
+                                "Auto-join room is invite only but there are no local"
+                                " user to invite with",
```
anoadragon453, line 499:
```suggestion
                                " users to invite with",
```

synapse/handlers/register.py
babolivier: Hmm, not sure how I've missed this method when I was looking for one to use. Thanks!

synapse/handlers/register.py
```diff
@@ -465,6 +492,13 @@ async def _join_rooms(self, user_id: str) -> None:
                         join_rule = join_rules_event.content.get("join_rule", None)
                         requires_invite = join_rule and join_rule != JoinRules.PUBLIC
 
+                        if requires_invite and self.server_name not in hosts_in_room:
+                            raise SynapseError(
```
anoadragon453, line 496: We probably don't want to raise in the middle of registration.

anoadragon453 (changes requested):
Looks like it would work! I mostly just had some comments on handling failure modes.

synapse/handlers/register.py
babolivier: It's already inside a try/except.

synapse/handlers/register.py
babolivier: It's already inside a try/except.

synapse/handlers/register.py
babolivier: It is still a misconfiguration, so imho something we should warn the user about.

anoadragon453: Oh, is it not a feature that you can auto-join remote rooms?

erikjohnston: I'm a bit conflicted about this, as this should only happen if you've set "auto create rooms" on *and* specified a remote alias, which obviously won't work. However, its not totally unreasonable to configure a local and remote room to auto join, and you want to auto create the local one.

babolivier: > Oh, is it not a feature that you can auto-join remote rooms?

It is, but my point was that this code is for creating and joining the room, not just joining them (hence the warning about _creating_ a room).

> However, its not totally unreasonable to configure a local and remote room to auto join, and you want to auto create the local one.

That's a fair point. I'll downgrade to info.

erikjohnston (approved):

babolivier:
The complement failure is because this PR is based off 1.36 to help a deployment so I'm going to ignore it. The sytest one seems to be a fluke given it run for a pretty long time and looks like it's just timed out on a test, so I'm going to ignore it as well and merge this PR.

PR #10239, opened 2021-06-23
Fixes to the release script

babolivier:
* rename major/minor into the right semver terminology minor/patch (since this was something that got me very confused the first couple of times I've used the script)
* name the release branch based on the new version, not the previous one

### DIFF
--- a/changelog.d/10239.misc
+++ b/changelog.d/10239.misc
@@ -1,0 +1,1 @@
+Fixes to the release script.
--- a/changelog.d/10239.misc
+++ b/changelog.d/10239.misc
@@ -1,1 +1,1 @@
-Fixes to the release script.
+Update the release script to use the semver terminology and determine the release branch based on the next version.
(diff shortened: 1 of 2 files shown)


scripts-dev/release.py
```diff
@@ -97,49 +91,57 @@ def run():
...
         else:
-            new_version = base_version
+            new_version = "{}.{}.{}".format(
+                current_version.major,
+                current_version.minor,
+                current_version.micro,
+            )
     else:
         # If this is a new release cycle then we need to know if its a major
         # version bump or a hotfix.
```
anoadragon453, line 101:
```suggestion
        # If this is a new release cycle then we need to know if it's a minor
        # or a patch version bump.
```

changelog.d/10239.misc
```diff
@@ -0,0 +1 @@
+Fixes to the release script.
```
anoadragon453, line 1: This lacks some description of what the changes actually were..

anoadragon453 (approved):
Thanks for doing this!

scripts-dev/release.py
babolivier: Oh, good catch!

anoadragon453 (approved):
lgtm!

PR #10241, opened 2021-06-23
Modify docs root redirect to develop/index.html instead of a specific page

anoadragon453:
Context: https://github.com/matrix-org/synapse/pull/10242

`index.html` will still be the welcome page (for now), but if we ever decide to change that in the future, we won't need to update this redirect page again.

Note that CI is not expected to pass.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,6 +1,6 @@
 <title> Synapse Documentation </title>
-<meta http-equiv="Refresh" content="0; url='https://matrix-org.github.io/synapse/develop/welcome_and_overview.html'" />
+<meta http-equiv="Refresh" content="0; url='https://matrix-org.github.io/synapse/develop" />
 <p>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1,6 +1,6 @@
 <title> Synapse Documentation </title>
-<meta http-equiv="Refresh" content="0; url='https://matrix-org.github.io/synapse/develop/welcome_and_overview.html'" />
+<meta http-equiv="Refresh" content="0; url='https://matrix-org.github.io/synapse/develop" />
```
anoadragon453, line 2: Note that GitHub Pages will serve `develop/index.html` if we navigate to `develop/`.

anoadragon453:
CI is not expected to pass against the `gh-pages` branch.

babolivier (approved):

PR #10243, opened 2021-06-23
Return errors from `send_join` etc if the event is rejected

richvdh:
Rather than persisting rejected events via `send_join` and friends, raise a 403 if someone tries to pull a fast one.

(~~Based on #10225~~)

### DIFF
--- a/synapse/handlers/federation.py
+++ b/synapse/handlers/federation.py
@@ -1953,9 +1953,21 @@ class FederationHandler(BaseHandler):
         self, origin: str, event: EventBase
     ) -> EventContext:
         """
-        We have received a join/leave/knock event for a room.
+        We have received a join/leave/knock event for a room via send_join/leave/knock.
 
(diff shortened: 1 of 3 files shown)


richvdh:
tests in https://github.com/matrix-org/complement/pull/132

synapse/handlers/federation.py
```diff
@@ -2061,35 +1949,54 @@ async def on_make_knock_request(
...
                 event.sender,
                 origin,
             )
             raise SynapseError(403, "User not from origin", Codes.FORBIDDEN)
 
-        event.internal_metadata.outlier = False
+        if event.sender != event.state_key:
+            raise SynapseError(400, "state_key and sender must match", Codes.BAD_JSON)
+
+        assert not event.internal_metadata.outlier
```
richvdh, line 1999: there's no way that we can have an outlier here.

erikjohnston (approved):

PR #10245, opened 2021-06-24
Make historical events discoverable from backfill for servers without any scrollback history (MSC2716)

MadLittleMods:
Make historical events discoverable from backfill for servers without any scrollback history.

 - This does not cover "marker" events or servers which already have all of the scrollback(no backwards extremeties to pull from) 
 - As the `/backfill` function iterates up the DAG, we also check for any insertion events which have prev_events pointing to the current backfill event. If we find one, we pull in the insertion event and start of the chunk. Then keep iterating up the chunk of history.

Follow-up to https://github.com/matrix-org/synapse/pull/9247

[Mermaid live editor playground link](https://mermaid-js.github.io/mermaid-live-editor/edit/#eyJjb2RlIjoiZmxvd2NoYXJ0IEJUXG4gICAgc3ViZ3JhcGggbGl2ZVxuICAgICAgICBCIC0tLS0tLS0tLS0tLS0tLS0tPiBBXG4gICAgZW5kXG4gICAgXG4gICAgc3ViZ3JhcGggY2h1bmswXG4gICAgICAgIGNodW5rMC1jaHVua1tbXCJjaHVua1wiXV0gLS0-IGNodW5rMC0yKChcIjJcIikpIC0tPiBjaHVuazAtMSgoMSkpIC0tPiBjaHVuazAtMCgoMCkpIC0tPiBjaHVuazAtaW5zZXJ0aW9uWy9pbnNlcnRpb25cXF1cbiAgICBlbmRcblxuICAgIHN1YmdyYXBoIGNodW5rMVxuICAgICAgICBjaHVuazEtY2h1bmtbW1wiY2h1bmtcIl1dIC0tPiBjaHVuazEtMigoXCIyXCIpKSAtLT4gY2h1bmsxLTEoKDEpKSAtLT4gY2h1bmsxLTAoKDApKSAtLT4gY2h1bmsxLWluc2VydGlvblsvaW5zZXJ0aW9uXFxdXG4gICAgZW5kXG4gICAgXG4gICAgc3ViZ3JhcGggY2h1bmsyXG4gICAgICAgIGNodW5rMi1jaHVua1tbXCJjaHVua1wiXV0gLS0-IGNodW5rMi0yKChcIjJcIikpIC0tPiBjaHVuazItMSgoMSkpIC0tPiBjaHVuazItMCgoMCkpIC0tPiBjaHVuazItaW5zZXJ0aW9uWy9pbnNlcnRpb25cXF1cbiAgICBlbmRcblxuICAgIFxuICAgIGNodW5rMC1pbnNlcnRpb25CYXNlWy9pbnNlcnRpb25cXF0gLS0tLS0tLS0tLS0tLT4gQVxuICAgIGNodW5rMC1jaHVuayAtLi0-IGNodW5rMC1pbnNlcnRpb25CYXNlWy9pbnNlcnRpb25cXF1cbiAgICBjaHVuazEtY2h1bmsgLS4tPiBjaHVuazAtaW5zZXJ0aW9uXG4gICAgY2h1bmsyLWNodW5rIC0uLT4gY2h1bmsxLWluc2VydGlvblxuIiwibWVybWFpZCI6Int9IiwidXBkYXRlRWRpdG9yIjpmYWxzZSwiYXV0b1N5bmMiOnRydWUsInVwZGF0ZURpYWdyYW0iOmZhbHNlfQ)

<details>
<summary>mermaid graph syntax</summary>

```mermaid
flowchart BT
    subgraph live
        B -----------------> A
    end
    
    subgraph chunk0
        chunk0-chunk[["chunk"]] --> chunk0-2(("2")) --> chunk0-1((1)) --> chunk0-0((0)) --> chunk0-insertion[/insertion\]
    end

    subgraph chunk1
        chunk1-chunk[["chunk"]] --> chunk1-2(("2")) --> chunk1-1((1)) --> chunk1-0((0)) --> chunk1-insertion[/insertion\]
    end
    
    subgraph chunk2
        chunk2-chunk[["chunk"]] --> chunk2-2(("2")) --> chunk2-1((1)) --> chunk2-0((0)) --> chunk2-insertion[/insertion\]
    end

    
    chunk0-insertionBase[/insertion\] -------------> A
    chunk0-chunk -.-> chunk0-insertionBase[/insertion\]
    chunk1-chunk -.-> chunk0-insertion
    chunk2-chunk -.-> chunk1-insertion
```

</details>

### Dev notes

 - https://github.com/matrix-org/synapse/blob/develop/docs/development/database_schema.md
 - https://github.com/matrix-org/synapse/blob/develop/docs/postgres.md
 - `psql synapse`: Open the postgres CLI against the `synapse` database
 - `\q` -> `enter`: To quit/exit
 - `\dt`: list tables
 - `\d+ issues`: list columns and indexes for issues table
 - `CREATE TABLE board_labels();`: Create a table called board_labels
 - `DROP TABLE insertion_event_extremeties`
 - `explain SELECT * FROM events WHERE depth = 1;`
 - `explain analyze SELECT * FROM events WHERE depth = 1;`
 
---

#### Access database in Complement test Docker container:

```
$ docker exec -it complement_1_hs_with_application_service.hs1_2 /bin/bash

$ apt-get update && apt-get install -y sqlite3

$ sqlite3
> .open /conf/homeserver.db
```

 
---

 - `synapse/storage/schema/__init__.py` (`synapse.storage.schema.SCHEMA_VERSION` and `synapse.storage.schema.SCHEMA_COMPAT_VERSION`)
 - `synapse/storage/prepare_database.py` ->`_setup_new_database` and `_upgrade_existing_database`
    > Sets up the physical database by finding a base set of "full schemas" and
    > then applying any necessary deltas, including schemas from the given data
    > stores.
    > 
    > The "full_schemas" directory has subdirectories named after versions. This
    > function searches for the highest version less than or equal to
    > `SCHEMA_VERSION` and executes all .sql files in that directory.
  - asdf

---

 - `event_backward_extremities`
 - `event_relations`
 - [`synapse/handlers/federation.py` -> `maybe_backfill`](https://github.com/matrix-org/synapse/blob/7f25d7385909ace3a84ee621f014d56734fecd44/synapse/handlers/federation.py#L1037) 
 - [`synapse/handlers/federation.py` -> `on_backfill_request`](https://github.com/matrix-org/synapse/blob/f0e02f5df2bc3ae779ac8c18578deebdfecc7e97/synapse/handlers/federation.py#L2111) -> [`synapse/storage/databases/main/event_federation.py` -> `get_backfill_events`](https://github.com/matrix-org/synapse/blob/f0e02f5df2bc3ae779ac8c18578deebdfecc7e97/synapse/storage/databases/main/event_federation.py#L896)

---

#### Call stacks for backfilling

**From the remote server**

 - `RoomMessageListRestServlet.on_GET`
 - `get_messages`
 - `maybe_backfill`
 - `_maybe_backfill_inner`
 - `try_backfill`
 - `backfill`

**From the origin server**

 - `FederationBackfillServlet.on_GET`
 - `on_backfill_request`
 - `get_backfill_events`

---

```
get_events
get_events_as_list
_get_events_from_cache_or_db
_get_events_from_db
_enqueue_events
_do_fetch
_fetch_event_list
_fetch_event_rows
```

---

Previous federation attempt at changing backfill looking at successors, https://github.com/matrix-org/synapse/pull/9247/files/13b18a8ec5cf03e3f46cc730464cfbfd10c82664#diff-1f5d8bffadd3271e42c2f6a66474bbf5e3e6694b009aa616b5f5433506217ff1R965

---

```
go test -v -tags synapse_blacklist,msc2946,msc3083,msc2716,msc2403 -count=1 $EXTRA_COMPLEMENT_ARGS ./tests -run TestBackfillingHistory/parallel/Historical_messages_are_visible_when_joining_on_federated_server

go test -v -tags synapse_blacklist,msc2946,msc3083,msc2716,msc2403 -count=1 $EXTRA_COMPLEMENT_ARGS ./tests -run TestBackfillingHistory/parallel/Historical_messages_are_visible_when_joining_on_federated_server_-_pre-made_insertion_event
```

### Todo

 - [x] Wait for https://github.com/matrix-org/synapse/pull/10250 to merge
 - [x] Remove need for `?user_id` query parameter which stops us from importing from multiple users in the chunk -> https://github.com/matrix-org/synapse/pull/10276
 - [x] Update `/backfill` to traverse ~~"markers"~~, "insertion" points, and chunks
 - [ ] A new "marker" event will allow server already in room to discover the history
    - Covered by https://github.com/matrix-org/synapse/pull/10420
 - [x] Add test case for base insertion event from the "live" timeline
 - [x] Add test for using the generated base insertion event part of the chunk
 - [x] Limit who can send insertion events and chunk ID content field on events (reject in auth code)
 - [ ] Assert `(room_id, next_chunk_id)` is unique in the app code. Insertion events should have unique `next_chunk_id` per room.

### Pull Request Checklist


synapse/storage/schema/main/delta/61/01insertion_event_lookups.sql
```diff
@@ -0,0 +1,50 @@
...
+CREATE INDEX IF NOT EXISTS insertion_event_edges_insertion_prev_event_id ON insertion_event_edges(insertion_prev_event_id);
+
+CREATE TABLE IF NOT EXISTS chunk_edges(
+    event_id TEXT NOT NULL,
+    room_id TEXT NOT NULL,
+    chunk_id TEXT NOT NULL,
+    UNIQUE (event_id, room_id)
+);
+
+CREATE INDEX IF NOT EXISTS chunk_edges_chunk_id ON chunk_edges(chunk_id);
```
MadLittleMods, line 50: I saw the following error once when running the Complement tests and just took note of it. Not quite sure if flakey or random intermittent state of my code while I was debugging stuff.

```
2021-07-13 02:27:57,810 - synapse.handlers.federation - 1248 - ERROR - GET-4 - Failed to backfill from hs1 because NOT NULL constraint failed: event_to_state_groups.state_group
  File "/usr/local/lib/python3.8/site-packages/synapse/handlers/federation.py", line 1216, in try_backfill
    await self.backfill(
  File "/usr/local/lib/python3.8/site-packages/synapse/handlers/federation.py", line 1035, in backfill
    await self._auth_and_persist_event(dest, event, context, backfilled=True)
  File "/usr/local/lib/python3.8/site-packages/synapse/handlers/federation.py", line 2222, in _auth_and_persist_event
    await self._run_push_actions_and_persist_event(event, context, backfilled)
  File "/usr/local/lib/python3.8/site-packages/synapse/handlers/federation.py", line 2244, in _run_push_actions_and_persist_event
    await self.persist_events_and_notify(
  File "/usr/local/lib/python3.8/site-packages/synapse/handlers/federation.py", line 3290, in persist_events_and_notify
    events, max_stream_token = await self.storage.persistence.persist_events(
  File "/usr/local/lib/python3.8/site-packages/synapse/logging/opentracing.py", line 774, in _trace_inner
    return await func(*args, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/synapse/storage/persist_events.py", line 320, in persist_events
    ret_vals = await yieldable_gather_results(enqueue, partitioned.items())
  File "/usr/local/lib/python3.8/site-packages/synapse/storage/persist_events.py", line 237, in handle_queue_loop
    ret = await self._per_item_callback(
  File "/usr/local/lib/python3.8/site-packages/synapse/storage/persist_events.py", line 577, in _persist_event_batch
    await self.persist_events_store._persist_events_and_state_updates(
  File "/usr/local/lib/python3.8/site-packages/synapse/storage/databases/main/events.py", line 176, in _persist_events_and_state_updates
  File "/usr/local/lib/python3.8/site-packages/synapse/storage/database.py", line 681, in runInteraction
    result = await self.runWithConnection(
  File "/usr/local/lib/python3.8/site-packages/synapse/storage/database.py", line 770, in runWithConnection
    return await make_deferred_yieldable(
  File "/usr/local/lib/python3.8/site-packages/twisted/python/threadpool.py", line 238, in inContext
    result = inContext.theWork()  # type: ignore[attr-defined]
  File "/usr/local/lib/python3.8/site-packages/twisted/python/threadpool.py", line 254, in <lambda>
    inContext.theWork = lambda: context.call(  # type: ignore[attr-defined]
  File "/usr/local/lib/python3.8/site-packages/twisted/python/context.py", line 118, in callWithContext
  File "/usr/local/lib/python3.8/site-packages/twisted/python/context.py", line 83, in callWithContext
    return func(*args, **kw)
  File "/usr/local/lib/python3.8/site-packages/twisted/enterprise/adbapi.py", line 293, in _runWithConnection
  File "/usr/local/lib/python3.8/site-packages/twisted/python/deprecate.py", line 298, in deprecatedFunction
    return function(*args, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/twisted/python/compat.py", line 403, in reraise
    raise exception.with_traceback(traceback)
  File "/usr/local/lib/python3.8/site-packages/twisted/enterprise/adbapi.py", line 284, in _runWithConnection
  File "/usr/local/lib/python3.8/site-packages/synapse/storage/database.py", line 765, in inner_func
    return func(db_conn, *args, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/synapse/storage/database.py", line 549, in new_transaction
  File "/usr/local/lib/python3.8/site-packages/synapse/logging/utils.py", line 69, in wrapped
    return f(*args, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/synapse/storage/databases/main/events.py", line 385, in _persist_events_txn
    self._store_event_state_mappings_txn(txn, events_and_contexts)
  File "/usr/local/lib/python3.8/site-packages/synapse/storage/databases/main/events.py", line 2065, in _store_event_state_mappings_txn
    self.db_pool.simple_insert_many_txn(
  File "/usr/local/lib/python3.8/site-packages/synapse/storage/database.py", line 923, in simple_insert_many_txn
    txn.execute_batch(sql, vals)
  File "/usr/local/lib/python3.8/site-packages/synapse/storage/database.py", line 280, in execute_batch
    self.executemany(sql, args)
  File "/usr/local/lib/python3.8/site-packages/synapse/storage/database.py", line 300, in executemany
    self._do_execute(self.txn.executemany, sql, *args)
  File "/usr/local/lib/python3.8/site-packages/synapse/storage/database.py", line 330, in _do_execute
    return func(sql, *args)
sqlite3.IntegrityError: NOT NULL constraint failed: event_to_state_groups.state_group
```

synapse/storage/databases/main/events.py
```diff
@@ -1504,6 +1504,9 @@ def _update_metadata_tables_txn(
 
             self._handle_event_relations(txn, event)
 
+            self._handle_insertion_event(txn, event)
+            self._handle_chunk_id(txn, event)
```
MadLittleMods, line 1508: Where is the best place to limit who can add an `insertion` event or a chunk connection content field?

synapse/storage/databases/main/event_federation.py
```diff
@@ -936,15 +936,50 @@ def _get_backfill_events(self, txn, room_id, event_list, limit):
...
             " INNER JOIN events"
             " ON prev_event_id = events.event_id"
+            # Find an event which matches the given event_id
             " WHERE event_edges.event_id = ?"
             " AND event_edges.is_state = ?"
             " LIMIT ?"
         )
 
+        # Look for the "insertion" events connected to the given event_id
+        # TODO: Do we need to worry about selecting only from the given room_id? The other query above doesn't
```
MadLittleMods, line 952: TODO: Do we need to worry about selecting only from the given room_id? The other query above doesn't

synapse/api/constants.py
```diff
@@ -196,9 +196,6 @@ class EventContentFields:
     MSC2716_CHUNK_ID = "org.matrix.msc2716.chunk_id"
     # For "marker" events
     MSC2716_MARKER_INSERTION = "org.matrix.msc2716.marker.insertion"
-    MSC2716_MARKER_INSERTION_PREV_EVENTS = (
-        "org.matrix.msc2716.marker.insertion_prev_events"
-    )
```
MadLittleMods, line 201: I don't think we will use this 🤷

synapse/storage/schema/main/delta/61/01insertion_event_lookups.sql
```diff
@@ -0,0 +1,50 @@
...
+-- Add a table that keeps track of "insertion" events back in the history
+-- when we get a "marker" event over the "live" timeline. When navigating the DAG
+-- and we hit an event which matches `insertion_prev_event_id`, it should backfill 
+-- the "insertion" event and start navigating from there.
+
+CREATE TABLE IF NOT EXISTS insertion_events(
+    insertion_event_id TEXT NOT NULL,
+    room_id TEXT NOT NULL,
+    next_chunk_id TEXT NOT NULL,
+    UNIQUE (insertion_event_id, room_id, next_chunk_id)
```
MadLittleMods: This needs to be unique to at least `(room_id, next_chunk_id)` 🤔

synapse/storage/schema/main/delta/61/01insertion_event_lookups.sql
MadLittleMods: This probably has something to do with how we are pulling in the historical messages and the auth events not being there 🤔

changelog.d/10245.feature
```diff
@@ -0,0 +1 @@
+Make historical events discoverable from backfill for servers without any scrollback history (part of MSC2716).
```
MadLittleMods, line 1: @erikjohnston I'm curious about your general sense for how I am tackling this bit so far?

For easy reference, here is what this PR does so far:

> Make historical events discoverable from backfill for servers without any scrollback history. This does not cover "marker" events or servers which already have all of the scrollback(no backwards extremeties to pull from)

I've got it working for auto-generated insertion event scenario. But still need to get the [Complement test](https://github.com/matrix-org/complement/pull/68) passing for when the insertion event is already in the "live" DAG and we add history onto that.

synapse/storage/databases/main/event_federation.py
```diff
@@ -936,15 +936,50 @@ def _get_backfill_events(self, txn, room_id, event_list, limit):
...
+        connected_insertion_event_query = (
+            "SELECT e.depth, i.event_id FROM insertion_event_edges AS i"
+            # Get the depth of the insertion event from the events table
+            " INNER JOIN events AS e"
+            " ON e.event_id = i.event_id"
```
erikjohnston, line 957:
```suggestion
            " INNER JOIN events AS e USING (event_id)"
```

That way you don't need to specify the table when referecing `event_id`

synapse/storage/databases/main/event_federation.py
```diff
@@ -970,9 +1005,38 @@ def _get_backfill_events(self, txn, room_id, event_list, limit):
 
             event_results.add(event_id)
 
+            txn.execute(
+                connected_insertion_event_query, (event_id, limit - len(event_results))
+            )
+            connected_insertion_event_id_results = list(txn)
```
erikjohnston, line 1011:
```suggestion
            connected_insertion_event_id_results = txn.fetchall()
```

Does the same thing, but I find it a bit more obvious

synapse/storage/databases/main/events.py
erikjohnston: We probably want to reject such events, so we probably want to add some stuff to the event authorization code.

synapse/storage/databases/main/event_federation.py
MadLittleMods: We don't need to worry about it because ["event IDs are globally unique"](https://github.com/matrix-org/synapse/pull/10245/files#r669641879) 👍

synapse/storage/databases/main/event_federation.py
MadLittleMods: Updated ⏩ 

Is this the preferred way? To me, it's a bit more confusing to keep track of which `event_id` it is.

synapse/storage/databases/main/event_federation.py
```diff
@@ -936,15 +936,50 @@ def _get_backfill_events(self, txn, room_id, event_list, limit):
...
+        # TODO: Do we need to worry about selecting only from the given room_id? The other query above doesn't
+        connected_insertion_event_query = (
+            "SELECT e.depth, i.event_id FROM insertion_event_edges AS i"
+            # Get the depth of the insertion event from the events table
+            " INNER JOIN events AS e"
+            " ON e.event_id = i.event_id"
+            # Find an insertion event which points via prev_events to the given event_id
+            " WHERE i.insertion_prev_event_id = ?"
+            " LIMIT ?"
+        )
```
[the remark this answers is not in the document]
MadLittleMods: 👍 

Ran into the `_make_sql_one_line` problem with the comments when I tried to use the `-- comment` syntax because it was commenting the rest of the query after the comment. Switched to `/* comment */` 😁

synapse/storage/schema/main/delta/61/01insertion_event_lookups.sql
```diff
@@ -0,0 +1,53 @@
...
+ * limitations under the License.
+ */
+
+-- Add a table that keeps track of "insertion" events and
+-- their next_chunk_id's so we can navigate to the next chunk of history.
+CREATE TABLE IF NOT EXISTS insertion_events(
+    event_id TEXT NOT NULL,
+    room_id TEXT NOT NULL,
+    next_chunk_id TEXT NOT NULL,
+    UNIQUE (event_id, room_id, next_chunk_id)
```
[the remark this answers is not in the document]
MadLittleMods: > Note that event IDs are globally unique

Good to know! Switched to just `event_id` unique.

> I could imagine that one day a remote server gets confused and reuses a chunk ID or something, at which point we probably don't want to explode on insertion of the event.

I think we probably want to reject this somewhere. If the `next_chunk_id` is the same as another in the same room, we will run into weird cross-talk between chunks.

synapse/storage/schema/main/delta/61/01insertion_event_lookups.sql
```diff
@@ -0,0 +1,53 @@
...
+);
+
+CREATE INDEX IF NOT EXISTS insertion_events_insertion_room_id ON insertion_events(room_id);
```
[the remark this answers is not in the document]
MadLittleMods: Removed as it's not used now :+1:

synapse/storage/databases/main/events.py
MadLittleMods: I am having trouble fitting this into the existing code.

 - I need to make sure that only application services can send insertion, marker, and events with chunk ID's
 - Other homeservers need to be able to accept these events over federation. But probably should only trust the originating homeserver where the room is. How does originating homeserver even work when future room ID's don't include the homeserver?
    - Maybe it should be some power level where we can add the application service user ID?

---

 - [`synapse/event_auth.py` -> check(...)](https://github.com/matrix-org/synapse/blob/0ae95b38474a4d64a4d5057499e645a3b81e3736/synapse/event_auth.py):
    - It feels like the checks should go here next to `check_redaction`
    - That class doesn't have access to the `store` so I can't use `store.get_app_service_by_user_id(...)` to make sure the sender is an application service. Plus this code is probably run for events coming over federation which don't know which `sender` is an application service.
    - It could fit here if it was a new power level since we have the auth events
 - [`synapse/handlers/event_auth.py` -> `check_from_context(...)`](https://github.com/matrix-org/synapse/blob/0ae95b38474a4d64a4d5057499e645a3b81e3736/synapse/handlers/event_auth.py)
    - I can use `store.get_app_service_by_user_id(...)` but other code is still only protected by the raw `check(...)`

---

<details>
<summary>def _check_if_allowed_to_send_historical_events</summary>

```python
    def _check_if_allowed_to_send_historical_events(
        self,
        event: EventBase,
    ):
        """Check whether the event sender is allowed to add insertion, marker,
        or events with a chunk ID

        Raises:
            AuthError if the event sender is not an application service
        """

        if (
            event.type != EventTypes.MSC2716_INSERTION
            and event.type != EventTypes.MSC2716_MARKER
            and event.content.get(EventContentFields.MSC2716_CHUNK_ID) is None
        ):
            pass

        app_service = self._store.get_app_service_by_user_id(event.sender)

        if app_service is None:
            raise AuthError(403, "Only application services can send insertion events")
```

</details>

MadLittleMods: Using the [` m.room.power_levels` -> `events`](https://matrix.org/docs/spec/client_server/r0.6.1#m-room-power-levels) ("The level required to send specific event types. This is a mapping from event type to power level required.") field seems perfect 🎉

synapse/storage/schema/main/delta/61/01insertion_event_lookups.sql
MadLittleMods: Continued in https://github.com/matrix-org/synapse/pull/10245#discussion_r669641879

synapse/storage/schema/main/delta/61/01insertion_event_lookups.sql
MadLittleMods: This is definitely a problem but it's flakey and doesn't happen every time which makes it harder to debug.

It looks like it's failing for one of the historical messages occasionally:

Logged from [`synapse/storage/databases/main/events.py#L1942-L1962`](https://github.com/matrix-org/synapse/blob/0ae95b38474a4d64a4d5057499e645a3b81e3736/synapse/storage/databases/main/events.py#L1942-L1962)
```
state_groups dict_items([('$SKA-gTylAHOHjmy5yScvHvPcTr7FikhGxVMmCVR6ufw', None)])
```

And the full event for context:
```
{
  "type": "m.room.message",
  "room_id": "!WLfsCaykamkOnZMHxt:hs1",
  "sender": "@maria:hs1",
  "content": {
    "body": "Historical 1 (chunk=0)",
    "msgtype": "m.text",
    "org.matrix.msc2716.historical": true,
    "org.matrix.msc2716.chunk_id": "ndocOxvR"
  },
  "origin_server_ts": 1626324541183,
  "unsigned": {
    "age": 1360
  },
  "event_id": "$SKA-gTylAHOHjmy5yScvHvPcTr7FikhGxVMmCVR6ufw",
  "user_id": "@maria:hs1",
  "age": 1360
},
```

---

I can see that hs2 does a `hs1:event_auth/` check for that event but then later says `Skipping auth_event fetch for outlier` which is expected. Then my guess is somehow we are getting to the `state_groups` code which doesn't like `outliers` with no state 🤷‍♀️

```
2021-07-15 04:49:03,643 - synapse.http.matrixfederationclient - 278 - INFO - GET-7 - {GET-O-17} [hs1] Completed request: 200 OK in 0.01 secs, got 3359 bytes - GET matrix://hs1/_matrix/federation/v1/event_auth/%21WLfsCaykamkOnZMHxt%3Ahs1/%24SKA-gTylAHOHjmy5yScvHvPcTr7FikhGxVMmCVR6ufw
2021-07-15 04:49:03,648 - synapse.handlers.federation - 2733 - INFO - GET-7 - Skipping auth_event fetch for outlier
2021-07-15 04:49:03,650 - synapse.storage.databases.main.events - 2030 - INFO - persist_events-27 - state_groups asdfasdf dict_items([])
2021-07-15 04:49:03,653 - synapse.storage.databases.main.events - 2030 - INFO - persist_events-27 - state_groups asdfasdf dict_items([('$SKA-gTylAHOHjmy5yScvHvPcTr7FikhGxVMmCVR6ufw', None)])
2021-07-15 04:49:03,655 - synapse.handlers.federation - 1244 - ERROR - GET-7 - Failed to backfill from hs1 because NOT NULL constraint failed: event_to_state_groups.state_group
```

MadLittleMods: @erikjohnston Does every `event` need a `state_group`? I'm curious if I can just add a condition that if the `event` doesn't have a `state_group`, then just skip trying to [persist it to `event_to_state_groups`](https://github.com/matrix-org/synapse/blob/0ae95b38474a4d64a4d5057499e645a3b81e3736/synapse/storage/databases/main/events.py#L1939).

changelog.d/10245.feature
MadLittleMods: Thanks for the review pass @erikjohnston 🦘. I've run into a few head scratchers around [securing who can send these new events](https://github.com/matrix-org/synapse/pull/10245#discussion_r669979503) and the [`state_groups` bug](https://github.com/matrix-org/synapse/pull/10245#discussion_r670145534) and could use some thoughts!

synapse/storage/schema/main/delta/61/01insertion_event_lookups.sql
MadLittleMods: It looks like a `state_group` isn't necessary for outliers because we don't have the state for them. Our historical state to auth are outliers but the actual historical messages are just backfilled events.

> _state_group:
> [... ]
> For an outlier, where we don't have the state at the event, this will be None.
>
> [`synapse/events/snapshot.py#L45`](https://github.com/matrix-org/synapse/blob/323452944e13114264b0c645db875dc6950315c5/synapse/events/snapshot.py#L45)

synapse/storage/databases/main/events.py
MadLittleMods: Added unstable room version `org.matrix.msc2716` which adds the `historical` power level that controls whether you can send `insertion`, `chunk`, and `marker` events ✅

I switched to `chunk` events so we can easily auth them against the PL level because they are sent by the application service user ID with the proper PL level for the room. See https://github.com/matrix-org/synapse/pull/10432

synapse/storage/schema/main/delta/61/01insertion_event_lookups.sql
erikjohnston: We should really document this somewhere, but: for every non-outlier event we need to know the state at that event. Instead of storing the full state for each event in the DB (i.e. a `event_id -> state` mapping), which is *very* space inefficient when state doesn't change, we instead assign each different set of state a "state group" and then have mappings of `event_id -> state_group` and `state_group -> state`

MadLittleMods: Thanks for the clarification @erikjohnston, I've fixed this up in https://github.com/matrix-org/synapse/pull/10439

synapse/api/constants.py
MadLittleMods: Switched to backfilling the insertion event in https://github.com/matrix-org/synapse/pull/10420

synapse/handlers/federation.py
```diff
@@ -2714,9 +2714,11 @@ async def _update_auth_events_and_context_for_auth(
                             event.event_id,
                             e.event_id,
                         )
-                        context = await self.state_handler.compute_event_context(e)
+                        missing_auth_event_context = (
+                            await self.state_handler.compute_event_context(e)
+                        )
                         await self._auth_and_persist_event(
-                            origin, e, context, auth_events=auth
+                            origin, e, missing_auth_event_context, auth_events=auth
```
MadLittleMods, line 2721: This will merged separately in https://github.com/matrix-org/synapse/pull/10439 but need the fix here as well

synapse/storage/schema/main/delta/61/01insertion_event_lookups.sql
MadLittleMods: Documented in https://github.com/matrix-org/synapse/pull/10464 🎉

synapse/storage/databases/main/event_federation.py
```diff
@@ -936,15 +936,46 @@ def _get_backfill_events(self, txn, room_id, event_list, limit):
...
+        query = """
+            SELECT depth, prev_event_id FROM event_edges
+            /* Get the depth of the prev_event_id from the events table */
+            INNER JOIN events
+            ON prev_event_id = events.event_id
+            /* Find an event which matches the given event_id */
+            WHERE event_edges.event_id = ?
+            AND event_edges.is_state = ?
+            LIMIT ?
+        """
```
MadLittleMods, line 949: Same as the previous query just in this new format with comments

synapse/storage/schema/main/delta/61/01insertion_event_lookups.sql
```diff
@@ -0,0 +1,52 @@
...
+-- Add a table that keeps track of "insertion" events and
+-- their next_chunk_id's so we can navigate to the next chunk of history.
+CREATE TABLE IF NOT EXISTS insertion_events(
+    event_id TEXT NOT NULL,
+    room_id TEXT NOT NULL,
+    next_chunk_id TEXT NOT NULL,
+    UNIQUE (event_id)
+);
+
+CREATE INDEX IF NOT EXISTS insertion_events_event_id ON insertion_events(event_id);
```
erikjohnston, line 25:
```suggestion
    next_chunk_id TEXT NOT NULL
);

CREATE UNIQUE INDEX IF NOT EXISTS insertion_events_event_id ON insertion_events(event_id);
```

I find that less confusing (and is generally what we do elsewhere)

synapse/storage/databases/main/event_federation.py
```diff
@@ -970,9 +1001,40 @@ def _get_backfill_events(self, txn, room_id, event_list, limit):
...
             event_results.add(event_id)
 
+            txn.execute(
+                connected_insertion_event_query, (event_id, limit - len(event_results))
+            )
+            connected_insertion_event_id_results = txn.fetchall()
+            logger.debug(
+                "_get_backfill_events: connected_insertion_event_query %s",
+                connected_insertion_event_id_results,
+            )
```
erikjohnston, line 1011: Can you add a comment about what this is all doing at a high level? i.e. something like "if we have an insertion event pointing at this event then we need to go and pull in all connected chunks", but with better words

synapse/event_auth.py
```diff
@@ -504,6 +511,38 @@ def check_redaction(
...
+
+    Raises:
+        AuthError if the event sender is not allowed to send historical related events
+        ("insertion", "chunk", and "marker").
+    """
+    if not room_version_obj.msc2716_historical:
+        raise AuthError(
+            403,
+            "Historical events not supported with your room version",
+        )
```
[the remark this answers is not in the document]
MadLittleMods: I think I know what you mean. We can't reject historical events on unsupported room versions because homeservers without knowledge of MSC2716 or the new room version don't reject historical events either.

Since we can't rely on the auth check here to stop historical events on unsupported room versions, I've added some additional checks in the processing/persisting code (`synapse/storage/databases/main/events.py` ->  `_handle_insertion_event` and `_handle_chunk_event`). I've had to do some refactoring so there is method to fetch the room version by `txn`.

synapse/storage/databases/main/events_worker.py
```diff
@@ -207,6 +207,82 @@ async def get_received_ts(self, event_id: str) -> Optional[int]:
...
+        room_version = KNOWN_ROOM_VERSIONS.get(room_version_id)
+
+        event = make_event_from_dict(
+            event_dict=d,
+            room_version=room_version,
+            internal_metadata_dict=internal_metadata,
+            rejected_reason=rejected_reason,
+        )
+
+        return event
```
erikjohnston, line 284: I'm a little bit cautious about adding this, so far we've managed to ensure we pull out events outside of transactions. Mainly because a) its expensive and b) there are vague thoughts about splitting the event_json table out of the main DB.

However, I think https://github.com/matrix-org/synapse/blob/master/synapse/storage/schema/main/delta/57/rooms_version_column_3.sql.postgres means we can rip out the fallback check where we look at the create event. Either way, we know that all `msc2716_historical` rooms will have been created after we populate the room version column, so we will never need to check the create event?

erikjohnston (commented):
I think this looks good, modulo adding a `get_event_txn` function. Have asked in #synapse-dev:matrix.org if its safe to rip the fallback code out

synapse/storage/databases/main/events_worker.py
MadLittleMods: I've removed the create event lookup for `room_version` 👍

erikjohnston (approved):
:tada:

MadLittleMods:
Woot! Thanks for all of the review @erikjohnston 🐣

PR #10247, opened 2021-06-24
build debs in GHA

### LINKED ISSUE #8555: Add debian packaging to CI
We keep breaking the debian package builds, which is annoying when trying to release.

We may want to consider just running this on develop/release branches to save resource?

Presumably we don't want to use the debs built in CI, in case they get compromised?

richvdh:
fixes #8555


.github/workflows/debs.yml
```diff
@@ -0,0 +1,44 @@
+# GitHub actions workflow which builds the debian packages.
+
+name: Debs
+
+on:
+  push:
```
babolivier, line 6: Do we want to run this on pull requests as well? I'm sometimes quick to dismiss a :x: on a branch because it happens sometimes due to flaky sytests and I don't necessarily bother looking further (because I assume everything that's been PR'd has been :heavy_check_mark:'d on the CI), but I think personally I'd pay more attention if the failure happens on a PR.

scripts-dev/build_debian_packages
```diff
@@ -179,11 +179,20 @@ if __name__ == "__main__":
...
         "dist",
         nargs="*",
         default=DISTS,
         help="a list of distributions to build for. Default: %(default)s",
     )
     args = parser.parse_args()
-    run_builds(dists=args.dist, jobs=args.jobs, skip_tests=args.no_check)
+    if args.show_dists_json:
+        json.dump(DISTS, sys.stdout)
+        print()
```
babolivier, line 196: I find `print(json.dumps(DISTS))` easier to read but ymmv

babolivier (approved):
lgtm otherwise

scripts-dev/build_debian_packages
richvdh: fair

.github/workflows/debs.yml
richvdh: yeah, I don't really want to tie up 7 GHA workers for 10-12 minutes every time someone pushes to a PR branch. #8555 suggested an approach of building for oldest and newest supported releases, which might make sense, but I'd rather have that as a separate PR.

PR #10250, opened 2021-06-25
Add base starting insertion event when no chunk ID is provided (MSC2716)

MadLittleMods:
Add base starting insertion event when no chunk ID is provided.

This is spawning from https://github.com/matrix-org/synapse/pull/10245 (follow-up to https://github.com/matrix-org/synapse/pull/9247) where I am actually trying to traverse the historical events. I realized we would skip the first chunk of messages if the marker event was pointing to the insertion point at the top of the chunk. Now the marker event can point to this base insertion event and we can traverse up the events and over to the next chunk.

Before | After
--- | ---
 | 

[TARDIS visualization](https://github.com/matrix-org/tardis/pull/1):

<details>
<summary>Previous diagrams (backup)</summary>

Before | After
--- | ---
 | 

[TARDIS visualization](https://github.com/matrix-org/tardis/pull/1):
Screen Shot 2021-06-25 at 12 35 11 AM

</details>

### Pull Request Checklist


synapse/rest/client/v1/room.py
```diff
@@ -449,30 +473,40 @@ async def on_POST(self, request, room_id):
...
+        # from the chunk_id in the "insertion" event from the previous chunk.
+        last_event_in_chunk = events_to_create[-1]
+        chunk_id_to_connect_to = chunk_id_from_query
         if chunk_id_from_query:
-            last_event_in_chunk = events_to_create[-1]
-            last_event_in_chunk["content"][
-                EventContentFields.MSC2716_CHUNK_ID
-            ] = chunk_id_from_query
+            # TODO: Verify the chunk_id_from_query corresponds to an insertion event
+            pass
```
MadLittleMods, line 483: We need a database index to make this lookup fast. Going to add this in https://github.com/matrix-org/synapse/pull/10245

synapse/rest/client/v1/room.py
```diff
@@ -449,30 +473,40 @@ async def on_POST(self, request, room_id):
...
+        # to as a starting point.
+        else:
+            base_insertion_event = self._create_insertion_event_dict(
+                sender=requester.user.to_string(),
+                origin_server_ts=last_event_in_chunk["origin_server_ts"],
+            )
+            events_to_create.append(base_insertion_event)
+            chunk_id_to_connect_to = base_insertion_event["content"][
+                EventContentFields.MSC2716_NEXT_CHUNK_ID
+            ]
```
MadLittleMods, line 494: Here is where we are creating the base insertion event

erikjohnston:
I'm a little bit confused. Shouldn't the marker event be pointing at the next chunk? The same way that the insertion events are pointing at the next chunk?

erikjohnston:
> I'm a little bit confused. Shouldn't the marker event be pointing at the next chunk? The same way that the insertion events are pointing at the next chunk?

So the way to think about this is that the base insertion event isn't in the first chunk per se, as ideally it would be part of the original DAG when we first created the room. However, we want to support adding historical events even where we don't have insertion events, and so we create a new base insertion event in an ad hoc manner.

synapse/rest/client/v1/room.py
```diff
@@ -449,30 +473,40 @@ async def on_POST(self, request, room_id):
...
+        chunk_id_to_connect_to = chunk_id_from_query
         if chunk_id_from_query:
-            last_event_in_chunk = events_to_create[-1]
-            last_event_in_chunk["content"][
-                EventContentFields.MSC2716_CHUNK_ID
-            ] = chunk_id_from_query
+            # TODO: Verify the chunk_id_from_query corresponds to an insertion event
+            pass
+        # Otherwise, create an insertion event to be based off of and connect
+        # to as a starting point.
```
erikjohnston, line 485:
```suggestion
        # We don't always have an insertion event to hang more history off of
        # (ideally they'd be one in the main DAG, but that's not the case if
        # we're wanting to add history to e.g. existing rooms without such an
        # event), in which case we just create a new insertion event that will
        # then get pointed to by a "marker" event.
```

Or something to that effect, I'm terrible with words.

synapse/rest/client/v1/room.py
```diff
@@ -349,6 +349,30 @@ async def inherit_depth_from_prev_ids(self, prev_event_ids) -> int:
...
 
+    def _create_insertion_event_dict(self, sender: str, origin_server_ts: int):
+        """
+        Creates an event dict for an "insertion" event with the proper fields
+        and a random chunk ID.
+        Args:
+            sender: The event author MXID
+            origin_server_ts: Timestamp when the event was sent
+        Returns:
+            Tuple of event ID and stream ordering position
```
erikjohnston, line 360:
```suggestion
        """Creates an event dict for an "insertion" event with the proper fields
        and a random chunk ID.
        
        Args:
            sender: The event author MXID
            origin_server_ts: Timestamp when the event was sent
            
        Returns:
            Tuple of event ID and stream ordering position
```

MadLittleMods:
> So the way to think about this is that the base insertion event isn't in the first chunk per se, as ideally it would be part of the original DAG when we first created the room. However, we want to support adding historical events even where we don't have insertion events, and so we create a new base insertion event in an ad hoc manner.

It would probably be more accurate if showed the chunk connection from the base insertion point. I can't make the graph align better but here is how it would look like. The insertion point is part of the first chunk currently which made it easy to insert and add.




Perhaps it should float more as you're thinking about:

synapse/rest/client/v1/room.py
```diff
@@ -349,6 +349,30 @@ async def inherit_depth_from_prev_ids(self, prev_event_ids) -> int:
...
+        """
+
+        next_chunk_id = random_string(64)
```
erikjohnston, line 363:
```suggestion
        next_chunk_id = random_string(8)
```

These will be scoped per room, so eight characters feels like more than enough? Which I think is equivalent to a ~50bit integer. 64 characters just seems needlessly long.

erikjohnston (changes requested):
I think this all makes sense

synapse/rest/client/v1/room.py
```diff
@@ -449,37 +478,68 @@ async def on_POST(self, request, room_id):
...
+                base_insertion_event,
+                _,
+            ) = await self.event_creation_handler.create_and_send_nonmember_event(
+                requester,
+                base_insertion_event_dict,
+                prev_event_ids=base_insertion_event_dict.get("prev_events"),
+                auth_event_ids=auth_event_ids,
+                historical=True,
+                depth=inherited_depth,
+            )
```
MadLittleMods, line 518: @erikjohnston I've decided to make the base insertion event float off on its own so its easier to reason about in my head

erikjohnston (approved):

MadLittleMods:
@erikjohnston Thanks for the review on this one! Really valuable to talk through it to see how we align 🦃

Issue #10240, opened 2021-06-23
certain users can only be deleted with erase: false


ajs124:
### Description
I was purging a few thousand users an appbridge left behind, all with `{ "erase": true }`, but got `{"errcode":"M_UNKNOWN","error":"No row found (profiles)"}` for a few dozen of them.
Without `erase: true`, I could successfully deactivate them.

### Steps to reproduce
- Try to delete users with erase: true
- Observe failure
- Try to delete same users without erase: true
- Observe success

### Version information
- **Homeserver**:
synapse @ helsinki-systems.de
- **Version**:
`{"server_version":"1.36.0","python_version":"3.8.9"}`
- **Install method**:
NixOS module
- **Platform**:
NixOS 21.05

dklimpel:
Do you have informations about this issue from your log file (`homeserver.log`)?

ajs124:
I got this from the journal, for the failed request:
```
synapse.http.server: [POST-199343] <XForwardedForRequest at 0x76f9bed0c6d0 method='POST' uri='/_synapse/admin/v1/deactivate/@[bridged user]' clientproto='HTTP/1.1' site='8080'> SynapseError: 404 - No row found (profiles)
synapse.access.http.8080: [POST-199343] ::ffff:127.0.0.1 - 8080 - {@[me]} Processed request: 0.072sec/-0.000sec (0.016sec, 0.007sec) (0.009sec/0.031sec/14) 57B 404 "POST /_synapse/admin/v1/deactivate/@[bridged user] HTTP/1.1" "curl/7.76.1" [0 dbevts]
```
For the successful request:
```
synapse.handlers.deactivate_account: [user_parter_loop-30710] User parter parting '@[bridged user]'
synapse.handlers.deactivate_account: [user_parter_loop-30710] User parter finished parting '@[bridged user]'
synapse.access.http.8080: [POST-199767] ::ffff:127.0.0.1 - 8080 - {@[me]} Processed request: 0.078sec/-0.000sec (0.024sec, 0.008sec) (0.010sec/0.038sec/16) 37B 200 "POST /_synapse/admin/v1/deactivate/@[bridged user] HTTP/1.1" "curl/7.76.1" [0 dbevts]
```

dklimpel:
It seems that your users have no profile in the database table `profiles`. For whatever reason this is so.

ajs124:
Right. I have a database dump that I could verify this with, if needed, but I'd rather not have to, because it's >50GB.

[...]

PR #8932, opened 2020-12-11
Remove user's avatar URL and displayname when deactivate an user with Admin API

dklimpel:
Related to: #7005, #7984
When a user gets deactivated this infomation is not needed anymore.

### Pull Request Checklist

* [x] Pull request is based on the develop branch
* [x] Pull request includes a [changelog file](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md#changelog). The entry should:
  - Be a short description of your change which makes sense to users. "Fixed a bug that prevented receiving messages from other servers." instead of "Moved X method from `EventStore` to `EventWorkerStore`.".
  - Use markdown where necessary, mostly for `code blocks`.
  - End with either a period (.) or an exclamation mark (!).
  - Start with a capital letter.
* [x] Pull request includes a [sign off](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md#sign-off)
* [x] Code style is correct (run the [linters](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md#code-style))

Signed-off-by: Dirk Klimpel dirk@klimpel.org

PR #10252, opened 2021-06-25
Fix deactivate a user if he does not have a profile

dklimpel:
Fixes: #10240
A user should also be deactivated and mark as erased if he has no profile.

### Pull Request Checklist

### DIFF
--- a/synapse/handlers/deactivate_account.py
+++ b/synapse/handlers/deactivate_account.py
@@ -15,7 +15,7 @@
 import logging
 from typing import TYPE_CHECKING, Optional
 
-from synapse.api.errors import SynapseError
+from synapse.api.errors import StoreError, SynapseError
 from synapse.metrics.background_process_metrics import run_as_background_process
 from synapse.types import Requester, UserID, create_requester
 
@@ -136,10 +136,21 @@ class DeactivateAccountHandler(BaseHandler):
         # Mark the user as erased, if they asked for that
         if erase_data:
             user = UserID.from_string(user_id)
-            # Remove avatar URL from this user
-            await self._profile_handler.set_avatar_url(user, requester, "", by_admin)
-            # Remove displayname from this user
-            await self._profile_handler.set_displayname(user, requester, "", by_admin)
+            try:
+                # Remove avatar URL from this user
+                await self._profile_handler.set_avatar_url(
+                    user, requester, "", by_admin
+                )
+                # Remove displayname from this user
+                await self._profile_handler.set_displayname(
+                    user, requester, "", by_admin
+                )
+            except StoreError as e:
+                # catch only 404 if user does not have a profile
+                if e.code == 404:
+                    pass
+                else:
+                    raise e
 
             logger.info("Marking %s as erased", user_id)
             await self.store.mark_user_erased(user_id)
--- a/synapse/handlers/deactivate_account.py
+++ b/synapse/handlers/deactivate_account.py
@@ -15,7 +15,7 @@
 import logging
 from typing import TYPE_CHECKING, Optional
 
-from synapse.api.errors import StoreError, SynapseError
+from synapse.api.errors import SynapseError
 from synapse.metrics.background_process_metrics import run_as_background_process
 from synapse.types import Requester, UserID, create_requester
 
@@ -136,19 +136,10 @@ class DeactivateAccountHandler(BaseHandler):
         # Mark the user as erased, if they asked for that
         if erase_data:
             user = UserID.from_string(user_id)
-            try:
-                # Remove avatar URL from this user
-                await self._profile_handler.set_avatar_url(
-                    user, requester, "", by_admin
-                )
-                # Remove displayname from this user
-                await self._profile_handler.set_displayname(
-                    user, requester, "", by_admin
-                )
-            except StoreError as e:
-                # bypass only 404 if user does not have a profile
-                if e.code != 404:
-                    raise e
+            # Remove avatar URL from this user
+            await self._profile_handler.set_avatar_url(user, requester, "", by_admin)
+            # Remove displayname from this user
+            await self._profile_handler.set_displayname(user, requester, "", by_admin)
 
             logger.info("Marking %s as erased", user_id)
             await self.store.mark_user_erased(user_id)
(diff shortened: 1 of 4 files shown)


synapse/handlers/deactivate_account.py
```diff
@@ -136,10 +136,19 @@ async def deactivate_account(
...
+                    user, requester, "", by_admin
+                )
+                # Remove displayname from this user
+                await self._profile_handler.set_displayname(
+                    user, requester, "", by_admin
+                )
+            except StoreError as e:
+                # bypass only 404 if user does not have a profile
+                if e.code != 404:
+                    raise e
```
anoadragon453, line 151: So the reason a `StoreError` is raised here is due to `set_profile_avatar_url` performing an `update` on `profiles` rather than an `upsert`:

https://github.com/matrix-org/synapse/blob/4b965c862dc66c0da5d3240add70e9b5f0aa720b/synapse/storage/databases/main/profile.py#L83-L91

I think a more fundamental fix for this whole problem would actually just be to convert `set_profile_avatar_url` to use `self.db_pool.simple_upsert`, like so:

```python
        await self.db_pool.simple_upsert(
            table="profiles",
            keyvalues={"user_id": user_localpart},
            values={},
            insertion_values={"avatar_url": new_avatar_url},
            desc="set_profile_avatar_url",
        )
```

it'd probably be a good idea to do the same for `set_profile_displayname` as well, just to avoid the same issue in the future.

tests/rest/admin/test_user.py
```diff
@@ -1098,6 +1098,61 @@ def test_deactivate_user_erase_false(self):
 
         self._is_erased("@user:test", False)
 
+    def test_deactivate_user_erase_true_no_profile(self):
+        """
+        Test deactivating a user and set `erase` to `true`
+        if user has no profile information in database table `profile`
```
anoadragon453, line 1104:
```suggestion
        if user has no profile information (stored in the database table `profiles`).
```

tests/rest/admin/test_user.py
```diff
@@ -1098,6 +1098,61 @@ def test_deactivate_user_erase_false(self):
...
+        )
+
+        self.assertEqual(200, int(channel.result["code"]), msg=channel.result["body"])
+        self.assertEqual("@user:test", channel.json_body["name"])
+        self.assertEqual(False, channel.json_body["deactivated"])
+        self.assertEqual("foo@bar.com", channel.json_body["threepids"][0]["address"])
+        self.assertIsNone(channel.json_body["avatar_url"])
+        self.assertIsNone(channel.json_body["displayname"])
+
+        # Deactivate user
```
anoadragon453, line 1128:
```suggestion
        # Deactivate and erase user
```

tests/rest/admin/test_user.py
```diff
@@ -1098,6 +1098,61 @@ def test_deactivate_user_erase_false(self):
...
+        body = json.dumps({"erase": True})
+
+        channel = self.make_request(
+            "POST",
+            self.url,
+            access_token=self.admin_user_tok,
+            content=body.encode(encoding="utf_8"),
```
anoadragon453, line 1135: You can actually just pass the Python dict directory to `make_request` these days.

tests/rest/admin/test_user.py
```diff
@@ -1098,6 +1098,61 @@ def test_deactivate_user_erase_false(self):
...
+        )
+
+        self.assertEqual(200, int(channel.result["code"]), msg=channel.result["body"])
```
anoadragon453, line 1138:
```suggestion
        self.assertEqual(200, channel.code, msg=channel.json_body)
```

tests/rest/admin/test_user.py
```diff
@@ -1098,6 +1098,61 @@ def test_deactivate_user_erase_false(self):
...
+
+        # Get user
+        channel = self.make_request(
+            "GET",
+            self.url_other_user,
+            access_token=self.admin_user_tok,
+        )
+
+        self.assertEqual(200, int(channel.result["code"]), msg=channel.result["body"])
```
anoadragon453, line 1147:
```suggestion
        self.assertEqual(200, channel.code, msg=channel.json_body)
```

anoadragon453:
Looks like this was introduced in https://github.com/matrix-org/synapse/pull/8932.

anoadragon453 (changes requested):
Thanks for the test! I have a suggestion below for an alternative fix that may prevent other edge cases.

dklimpel:
> Looks like this was introduced in #8932.

Yes. I didn't realise that there were users without profiles.

synapse/handlers/deactivate_account.py
dklimpel: Ok. I will make an update.

richvdh:
why *do* we have users without profiles?

dklimpel:
> why _do_ we have users without profiles?

If you create a user without displayname
https://github.com/matrix-org/synapse/blob/f0e02f5df2bc3ae779ac8c18578deebdfecc7e97/synapse/storage/databases/main/registration.py#L1563-L1575

no profile will be created
https://github.com/matrix-org/synapse/blob/f0e02f5df2bc3ae779ac8c18578deebdfecc7e97/synapse/storage/databases/main/registration.py#L1629-L1638

dklimpel:
> why _do_ we have users without profiles?

https://github.com/matrix-org/synapse/issues/10240#issuecomment-868986543

richvdh:
> > why _do_ we have users without profiles?
> 
> If you create a user without displayname ... no profile will be created

ugh.

tests/rest/admin/test_user.py
```diff
@@ -1098,6 +1094,59 @@ def test_deactivate_user_erase_false(self):
 
         self._is_erased("@user:test", False)
 
+    def test_deactivate_user_erase_true_no_profile(self):
+        """
+        Test deactivating a user and set `erase` to `true`
+        if user has no profile information (stored in the database table `profiles`).
+        """
+
+        # Delete profile information
```
richvdh, line 1103:
```suggestion
        # Users normally have an entry in `profiles`, but occasionally they are created without one.
        # To test deactivation for users without a profile, we delete the profile information for our user.
```

richvdh (approved):
lgtm, thanks!

PR #10253, opened 2021-06-25
Correct type hints for computing auth events.

clokep:
Found this while updating the implementation for MSC3083. Seemed easiest to put up a separate PR for it.

In order to not have an "unused" attribute, I need to change from using `blah = attr.ib(type=Foo` to `blah: "Foo"`. I took it as an opportunity to use `auto_attribs`.

### DIFF
--- a/synapse/handlers/message.py
+++ b/synapse/handlers/message.py
@@ -509,6 +509,8 @@ class EventCreationHandler:
                 Should normally be left as None, which will cause them to be calculated
                 based on the room state at the prev_events.
 
+                If non-None, prev_event_ids must also be provided.
+
             require_consent: Whether to check if the requester has
                 consented to the privacy policy.
 
@@ -581,6 +583,9 @@ class EventCreationHandler:
         # Strip down the auth_event_ids to only what we need to auth the event.
         # For example, we don't need extra m.room.member that don't match event.sender
         if auth_event_ids is not None:
+            # If auth events are provided, prev events must be also.
+            assert prev_event_ids is not None
+
             temp_event = await builder.build(
                 prev_event_ids=prev_event_ids,
                 auth_event_ids=auth_event_ids,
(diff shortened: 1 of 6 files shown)


erikjohnston:
> In order to not have an "unused" attribute, I need to change from using `blah = attr.ib(type=Foo` to `blah: "Foo"`. I took it as an opportunity to use `auto_attribs`.

I'm confused by this, why would using `auto_attribs` make a difference?

clokep:
> > In order to not have an "unused" attribute, I need to change from using `blah = attr.ib(type=Foo` to `blah: "Foo"`. I took it as an opportunity to use `auto_attribs`.
> 
> I'm confused by this, why would using `auto_attribs` make a difference?

Changing from declaring the type using `type=blah` to the native type hint syntax (`foo: blah`) is what fixed it, but I figured I should just switch to `auto_attrib` while I was doing it (to make the class consistent).

Otherwise you'd end up with:

```python
class Klass:
    foo: int = attr.ib()
    bar = attr.ib(type=str)
```

clokep:
> In order to not have an "unused" attribute,

Ah -- this should say "unused **import**" not attribute. 😄

synapse/handlers/message.py
```diff
@@ -581,6 +583,9 @@ async def create_event(
         # Strip down the auth_event_ids to only what we need to auth the event.
         # For example, we don't need extra m.room.member that don't match event.sender
         if auth_event_ids is not None:
+            # If auth events are provided, prev events must be also.
+            assert prev_event_ids is not None
```
clokep, line 587: I discussed this with @MadLittleMods, although the docstring says that `prev_event_ids` will be computed if not given this happens in the call to `create_new_client_event` ~20 lines below this.

MadLittleMods: As discussed at https://matrix.to/#/!XaqDhxuTIlvldquJaV:matrix.org/$J_SAM4WacKXqTfcSwGfBIcos5vEJO1i6x9L38oyKPY0?via=matrix.org&via=vector.modular.im&via=envs.net

babolivier (approved):

PR #10254, opened 2021-06-25
Update the MSC3083 support to verify if joins are from an authorized server

clokep:
[MSC3083](https://github.com/matrix-org/matrix-doc/pull/3083) was updated to only allow `join` events to the room if one of the following is true:

* The user is already joined or invited to the room.
* The join event is signed by a server whose users are able to issue invites for that room.

In order to do this we need to:

* Have the `join` event depend on a membership event in the room (as proof the user can sign the event).
* Add our signature when an event comes over federation to `send_join`.
* Check whether a remote join needs to be done, even if the server is in the room already.
* Update the auth rules to check the above signature.

Depends on #10268

Complement tests at matrix-org/complement#145


clokep:
(Complement tests failing here are expected.)

synapse/handlers/federation.py
```diff
@@ -1695,22 +1695,84 @@ async def on_make_join_request(
...
         # The remote hasn't signed it yet, obviously. We'll do the full checks
         # when we get the event back in `on_send_join_request`
         await self.auth.check_from_context(
-            room_version, event, context, do_sig_check=False
+            room_version.identifier, event, context, do_sig_check=False
         )
 
         return event
 
+    async def _get_user_event_which_could_invite(
```
clokep, line 1730: I'm surprised I don't seem to need to call this in `send_join`, but I might be getting "lucky" in my testing that the room state hasn't changed between `make_join` and `send_join`?

synapse/handlers/federation.py
```diff
@@ -1695,22 +1695,84 @@ async def on_make_join_request(
...
+        additional_auth_ids = None
+        current_state_ids = await self.store.get_current_state_ids(room_id)
+        if await self._event_auth_handler.has_restricted_join_rules(
+            current_state_ids, room_version
+        ):
+            additional_auth_ids = [
+                await self._get_user_event_which_could_invite(
+                    room_id, current_state_ids
+                )
+            ]
```
clokep, line 1709: This same thing needs to be done for a locally issued join.

clokep: I'm unsure this is going to work properly, it kind of moves some of the logic about which events to consider in the  auth chain to outside of that calculation (see `auth_types_for_event`). Unfortunately this change means that the events to include in the auth chain depend on the current state of the room (since you need a current membership event in the room to invite).

I'm unsure if this is just some refactoring, or if this breaks an assumption somewhere.

clokep:
I'm going to need to rebase this on #10253 and #10268.

synapse/event_auth.py
```diff
@@ -342,16 +351,41 @@ def _is_membership_change_allowed(
...
-            pass
+            # This is the same as public, but the event must be signed by a server
+            # whose users could issue invites.
+            #
+            # Signatures are only checked once the event is fully created, e.g.
+            # not during make_join,
+            #
+            # Note that if the caller is in the room or invited, then they do
+            # not need to meet the allow rules.
+            if do_sig_check and not caller_in_room and not caller_invited:
```
clokep, line 374: I'm not 100% confident that we should be skipping this when `do_sig_check` is `False`, but it was a change I need to make in order for `/make_join` to work.

synapse/event_auth.py
```diff
@@ -637,6 +671,66 @@ def get_user_power_level(user_id: str, auth_events: StateMap[EventBase]) -> int:
...
+    This is done by exploring the joined users and comparing their power levels
+    to the necessyar power level to issue an invite.
+
+    Args:
+        auth_events: state in force at this point in the room
+
+    Returns:
+        The users which can issue invites.
+    """
+    invite_level = _get_named_level(auth_events, "invite", 50)
```
clokep, line 687: Other places seem to default this to `0` if it doesn't exist, but the spec says `50`.

synapse/handlers/event_auth.py
```diff
@@ -52,7 +52,7 @@ async def check_from_context(
             room_version_obj, event, auth_events=auth_events, do_sig_check=do_sig_check
         )
 
-    def compute_auth_events(
+    async def compute_auth_events(
```
clokep, line 55: There's a few places that directly call `auth_types_for_event` instead of this:

* `FederationHandler._check_for_soft_fail`
* State resolution:
    * `v1._create_auth_events_from_maps`
    * `v1._resolve_auth_events`
    * `v2._iterative_auth_checks`

I haven't yet checked what (if any) changes need to be made to those.

synapse/handlers/federation.py
```diff
@@ -2009,6 +2013,17 @@ async def on_send_membership_event(
...
+        # Sign the event since we're vouching on behalf of the remote server that
+        # the event is valid to be sent into the room.
+        event.signatures.update(
+            compute_event_signature(
+                room_version,
+                event.get_pdu_json(),
+                self.hs.hostname,
+                self.hs.signing_key,
+            )
+        )
```
clokep, line 2025: We discussed this in #synapse-dev:matrix.org and the decision was to unconditionally do this (for `send_{join,leave,knock}`) since it sounds like a good idea. I don't know if we should only do it for the new room version though? The MSC doesn't assert it must always be sent, just that it needs to be done for join events in restricted rooms.

clokep:
This is ready for an initial review. I've left a few open questions that I would appreciate people's thoughts on!

synapse/event_auth.py
erikjohnston: I think that's a spec bug.

synapse/event_auth.py
```diff
@@ -637,6 +671,66 @@ def get_user_power_level(user_id: str, auth_events: StateMap[EventBase]) -> int:
...
+    Args:
+        users: A list of users.
+
+    Returns:
+        A set of servers.
+    """
+    servers = set()
+    for user in users:
+        try:
+            servers.add(UserID.from_string(user).domain)
```
erikjohnston, line 728: Can we not use `get_domain_from_id` here

synapse/handlers/event_auth.py
```diff
@@ -88,8 +88,69 @@ def compute_auth_events(
             if auth_ev_id:
                 auth_ids.append(auth_ev_id)
 
+        # If the current room is using restricted join rules, an additional event
+        # must be included to assert that the server has the right to authorise
+        # a join event.
+        if event.type == EventTypes.Member:
+            if await self.has_restricted_join_rules(
+                current_state_ids, event.room_version
+            ):
```
erikjohnston, line 97: We'll also need to update `auth_types_for_event` somehow. I'm not really sure how best to do that though.

synapse/handlers/federation.py
erikjohnston: I'm a little bit nervous about adding signatures that we then don't check? The problem being if in future these signatures can no longer be verified, what do servers do?

synapse/handlers/room_member.py
```diff
@@ -701,7 +702,36 @@ async def update_membership_locked(
...
-            if not is_host_in_room:
+            # Check if a remote join should be performed.
+            remote_join = not is_host_in_room
+            if not remote_join:
+                # If the host is in the room, but not one of the authorised hosts
+                # for restricted join rules, a remote join must be used.
+                room_version = await self.store.get_room_version(room_id)
+                current_state_ids = await self.store.get_current_state_ids(room_id)
+
+                # Otherwise, check if they should be allowed access via membership in a space.
```
erikjohnston, line 713: Otherwise? We haven't done a check yet?

synapse/handlers/room_member.py
```diff
@@ -701,7 +702,36 @@ async def update_membership_locked(
...
+                if await self.event_auth_handler.has_restricted_join_rules(
+                    current_state_ids, room_version
+                ):
+                    current_state_map = await self.store.get_events(
```
erikjohnston, line 717:
```suggestion
                    event_map = await self.store.get_events(
```

? `current_state_map` to me makes me think of a `StateMap` which have keys of `(type, state_key)`, rather than event IDs

synapse/event_auth.py
```diff
@@ -342,16 +351,41 @@ def _is_membership_change_allowed(
...
+                authorised_servers = get_servers_from_users(
+                    get_users_which_can_issue_invite(auth_events)
+                )
+
+                # Ensure one of the signatures is from one of the authorised servers.
+                # Note that it was previously checked that the signatures are
+                # valid.
+                for signing_server in event.signatures:
+                    if signing_server in authorised_servers:
+                        break
```
erikjohnston, line 385: We need to add a check in `federation_base._check_sigs_on_pdu` for a valid sig, we don't automatically check all signatures on events.

synapse/event_auth.py
```diff
@@ -637,6 +671,66 @@ def get_user_power_level(user_id: str, auth_events: StateMap[EventBase]) -> int:
             return 0
 
 
+def get_users_which_can_issue_invite(auth_events: StateMap[EventBase]) -> List[str]:
```
erikjohnston, line 674: I'm a bit worried that this is returning a list of potential servers? It feels like a recipe for different implementations to do signature checks on different servers and thus potentially come to different conclusions (though it seems fairly unlikely).

synapse/handlers/event_auth.py
erikjohnston: I wonder if we should add a field to the join event of `joined_via_user: <user_id>`, which corresponds to the user ID in the auth events. That way the set of auth types for a given event is still well specified.

synapse/event_auth.py
erikjohnston: And I don't think we should be using different defaults to other places in this file, as that feels like its going to be *very* confusing

clokep: I think we already might use different defaults in different places? I'll file a spec bug I suppose. 😄 

(This is partially why I wanted to abstract the power-level stuff out so we only specify the default once...)

synapse/handlers/federation.py
clokep: They would do the same thing as if the current signature is not able to be verified, I think?

synapse/event_auth.py
clokep: I don't think I realized that existed! 😄

synapse/handlers/event_auth.py
clokep: Let's leave this conversation to the MSC?

synapse/event_auth.py
clokep: Do we only require a single valid signature?

synapse/event_auth.py
erikjohnston: I don't think we do use different defaults in the auth rules check? We may as where (let's not confuse the default power levels sent when creating the room).

synapse/handlers/federation.py
erikjohnston: Synapse doesn't check extraneous signatures, so in some ways that's fine, but its a bit odd for Synapse to add extra signatures that aren't needed and I do think will likely cause problems where new servers think they should be verifying those sigs :shrug:

synapse/event_auth.py
clokep: Oh, we don't seem to...except here (which is the one I added).

clokep: I updated this and filed matrix-org/matrix-doc#3281.

synapse/event_auth.py
clokep: I think that means we *don't* need to check the signature here then, does that seem correct?

synapse/handlers/room_member.py
clokep: I think this comment got pulled around from copying and pasting. It doesn't seem to make sense anymore.

synapse/event_auth.py
clokep: Actually this check should morph into ensuring that the user provided has the power to invite, I think.

erikjohnston: Yup

synapse/event_auth.py
```diff
@@ -106,6 +106,18 @@ def check(
...
+            event.type == EventTypes.Member
+            and event.membership == Membership.JOIN
+            and "join_authorised_via_users_server" in event.content
+        )
+        if is_invite_via_allow_rule:
+            authoriser_domain = get_domain_from_id(
+                event.content["join_authorised_via_users_server"]
+            )
+            if not event.signatures.get(authoriser_domain):
+                raise AuthError(403, "Event not signed by authorising server")
```
clokep, line 119: I'm pretty sure we need to do this here and in `_check_sigs_on_pdu` (since that checks the signature is *valid* while this just checks that a *signature exists*). Anyway I matched what the 3pid code did.

synapse/event_auth.py
clokep: This is now only used to find a list of candidate servers to attempt a remote join to.

synapse/event_auth.py
```diff
@@ -760,4 +855,13 @@ def auth_types_for_event(event: Union[EventBase, EventBuilder]) -> Set[Tuple[str
                 )
                 auth_types.add(key)
 
+        # TODO Should this be limited to only MSC3083 rooms.
```
clokep, line 858: I'm thinking yes, which means we'll need to pipe the room version in here.

synapse/handlers/room_member.py
```diff
@@ -340,16 +341,10 @@ async def _local_membership_update(
...
             newly_joined = True
-            prev_member_event = None
             if prev_member_event_id:
                 prev_member_event = await self.store.get_event(prev_member_event_id)
                 newly_joined = prev_member_event.membership != Membership.JOIN
 
-            # Check if the member should be allowed access via membership in a space.
-            await self.event_auth_handler.check_restricted_join_rules(
-                prev_state_ids, event.room_version, user_id, prev_member_event
-            )
```
clokep, line 351: This is moved further up the call chain (into `_should_perform_remote_join`).

synapse/federation/federation_server.py
```diff
@@ -597,6 +598,7 @@ async def on_send_join_request(
 
         time_now = self._clock.time_msec()
         return {
+            "event": event.get_pdu_json(),
```
clokep, line 601: Should we only include this on certain room versions? Only if we've modified it?

clokep:
@erikjohnston Would appreciate another review on this! I've also made some updates to the MSC.

The thing this is still missing is better error handling (so that a server knows whether it should try another server or not when attempting to join a room). I'm hoping to do that as a follow-on to this as this is already much larger than I wanted it to be.

The commits became somewhat of a disaster. It can rebase this / attempt to split it up if you would like, but as you've already reviewed it I'm unsure if that would be clearer.

synapse/handlers/room_member.py
```diff
@@ -826,6 +825,107 @@ async def update_membership_locked(
...
+                Membership.INVITE,
+            ):
+                return False, []
+
+        # If the local host has a user who can issue invites, then a local
+        # join can be done.
+        #
+        # If not, generate a new list of remote hosts based on which
+        # can issue invites.
+        event_map = await self.store.get_events(current_state_ids.values())
```
erikjohnston, line 896: Again, its a bit of a shame we're pulling out the full state here preemptively :(

synapse/handlers/room_member.py
```diff
@@ -826,6 +825,107 @@ async def update_membership_locked(
...
+        # If this is going to be a local join, additional information must
+        # be included in the event content in order to efficiently validate
+        # the event.
+        authorised_user_id = await self.event_auth_handler.get_user_which_could_invite(
+            room_id,
+            current_state_ids,
+        )
+        if authorised_user_id:
+            content["join_authorised_via_users_server"] = authorised_user_id
+        # TODO If no authorised user is found then something bad happened.
```
erikjohnston, line 925: TODO?

synapse/federation/federation_server.py
```diff
@@ -748,6 +750,34 @@ async def _on_send_membership_event(
...
+        ):
+            # We can only authorise our own users.
+            authorising_server = get_domain_from_id(
+                event.content["join_authorised_via_users_server"]
+            )
+            if authorising_server != self.server_name:
+                raise SynapseError(
+                    400,
+                    "Cannot authorise request from resident server: %s"
+                    % (authorising_server,),
```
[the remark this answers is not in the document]
clokep: I don't think all that has landed yet, but I can update this.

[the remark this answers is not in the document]
erikjohnston: I thought I'd merged that....

synapse/federation/federation_base.py
```diff
@@ -178,6 +178,29 @@ async def _check_sigs_on_pdu(
             )
             raise SynapseError(403, errmsg, Codes.FORBIDDEN)
 
+    # If this is a join event for a restricted room it may have been authorised
+    # via a different server from the sending server. Check those signatures.
+    if room_version.msc3083_join_rules and _is_invite_via_allow_rule(pdu):
+        authorising_server = get_domain_from_id(
+            pdu.content["join_authorised_via_users_server"]
```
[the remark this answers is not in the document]
clokep: Sure! I was trying to match the logic of the 3pid check above, but I guess that uses a field from the `EventBase` object, not from the content so it is a bit less confusing.

synapse/handlers/federation.py
```diff
@@ -1691,8 +1692,44 @@ async def on_make_join_request(
...
         event_content = {"membership": Membership.JOIN}
 
+        # If the current room is using restricted join rules, additional information
+        # may need to be included in the event content in order to efficiently
+        # validate the event.
+        #
+        # Note that this requires the /send_join request to come back to the
+        # same server.
+        if room_version.msc3083_join_rules:
+            state_ids = await self.store.get_current_state_ids(room_id)
```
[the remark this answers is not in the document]
clokep: That's definitely true. Do you think it makes more sense to pull just the join rules, then re-pull more if we need it?

synapse/federation/federation_client.py
```diff
@@ -691,18 +696,41 @@ async def send_join(
...
                 no servers successfully handle the request.
         """
 
-        async def send_request(destination) -> Dict[str, Any]:
+        async def send_request(destination) -> SendJoinResult:
             response = await self._do_send_join(room_version, destination, pdu)
 
+            # If an event was returned:
+            #
+            # * Ensure it has the same event ID.
```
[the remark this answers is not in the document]
clokep: This is only true for newer room versions, but I guess no one is going to propose basing this off room version 1 or something. 👍

synapse/handlers/federation.py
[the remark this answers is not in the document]
erikjohnston: Yeah, I'd be tempted to only pull out what you need when you need it, and rely on caching to protect from duplicate work :shrug:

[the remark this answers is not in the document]
clokep: We talked a bit about this and decided to keep the current version as `get_filtered_current_state_ids` is uncached.

synapse/handlers/room_member.py
clokep: From above -- we discussed this and decided to keep the current version since the alternative (`get_filtered_current_state_ids`) is not cached.

clokep:
Things this needs to do still:

* [ ] Update error codes to match MSC.
* [ ] Ensure a send_join goes back to the same server as make_join.

synapse/event_auth.py
erikjohnston: I'd be tempted to lift these two functions out of here, as this file is otherwise stuff purely related to the auth rules check from the spec?

synapse/event_auth.py
erikjohnston: Yeah, I think it makes sense for `auth_types_for_event` to be room version specific.

synapse/federation/federation_server.py
erikjohnston: Maybe we should namespace it for now though, i.e. `org.matrix.msc3083.v2.event` or something?

synapse/handlers/event_auth.py
```diff
@@ -90,6 +94,66 @@ def compute_auth_events(
...
+        chosen_user = max(
+            local_users_in_room,
+            key=lambda user: users.get(user, users_default_level),
+            default=None,
+        )
+
+        # Return the chosen if they can issue invites.
+        user_power_level = users.get(chosen_user, users_default_level)
+        if chosen_user and user_power_level >= invite_level:
+            logger.error(
```
erikjohnston, line 142: Error?

synapse/handlers/room_member.py
```diff
@@ -826,6 +825,106 @@ async def update_membership_locked(
...
+        event_map = await self.store.get_events(current_state_ids.values())
+        current_state = {
+            state_key: event_map[event_id]
+            for state_key, event_id in current_state_ids.items()
+        }
+        allowed_servers = get_servers_from_users(
+            get_users_which_can_issue_invite(current_state)
+        )
+
+        # If the local server isnot  one of allowed servers, then a remote
```
erikjohnston, line 905:
```suggestion
        # If the local server is not one of allowed servers, then a remote
```

erikjohnston (commented):
I think otherwise it looks good

synapse/handlers/event_auth.py
clokep: Looks like this was debug code I left in. 😄

synapse/federation/federation_server.py
clokep: Sounds like a good idea.

erikjohnston (approved):
SHIP IT. SHIP IT NOW.

PR #10256, opened 2021-06-25
Do not recurse into non-spaces in the spaces summary.

clokep:
This is the straightforward part of #10085 split out into a separate PR.

This only searches for `m.child.space` events in spaces, not all rooms.

Clients do not use this information anyway (since they do not show a normal room as a space) -- the [MSC was updated](https://github.com/matrix-org/matrix-doc/pull/2946/commits/797dda44c4158e83dcff017f1ef2f36d7c2c4f3c) with these changes.

My hope is that this will help servers which have a slow database (or are starved for CPU, maybe). There's been some reports of the endpoint being very slow.

### DIFF
--- a/synapse/handlers/space_summary.py
+++ b/synapse/handlers/space_summary.py
@@ -319,7 +319,7 @@ class SpaceSummaryHandler:
 
         Returns:
             A tuple of:
-                The room information, if it the room should be returned to the
+                The room information, if the room should be returned to the
                 user. None, otherwise.
 
                 An iterable of the sorted children events. This may be limited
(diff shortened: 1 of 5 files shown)


clokep:
Corresponding complement tests: matrix-org/complement#138

synapse/handlers/space_summary.py
```diff
@@ -328,7 +330,11 @@ async def _summarize_local_room(
 
         room_entry = await self._build_room_entry(room_id)
 
-        # look for child rooms/spaces.
+        # If the the room is not a space, return just the room information.
```
dkasak, line 333:
```suggestion
        # If the room is not a space, return just the room information.
```

dkasak (approved):
Some minor things but looks good to me otherwise.

babolivier (approved):

PR #10264, opened 2021-06-28
Migrate stream_ordering to a bigint

### LINKED ISSUE #8255: events table uses a 32-bit integer for stream_ordering
2 billion events isn't actually that many; current estimate for exhaustion on matrix.org is Q3 2021, but that's an optimistic estimate as the rate of events is increasing.

nb there are other tables that use a standard int too.

richvdh:
A set of background updates (mostly) to replace `events.stream_ordering` with a `BIGINT`.

Fixes at least part of #8255.

The general approach here is:

1. Add a new `BIGINT` column called `stream_ordering2`.
2. Add a Postgres rule to populate that column with the same value as `stream_ordering` for any new rows.
3. Run a bg update to back-populate the column for all *existing* rows.
4. Add an index on the new column.
5. In a single transaction:
    * Double-check that we didn't miss any rows with a final `UPDATE`
    * Drop the rule
    * Drop the old column
    * Rename the new column into place

I still need to check that we don't have any ints in other tables.

### DIFF
--- a/synapse/storage/databases/main/events_bg_updates.py
+++ b/synapse/storage/databases/main/events_bg_updates.py
@@ -29,6 +29,12 @@ from synapse.types import JsonDict
 logger = logging.getLogger(__name__)
 
 
+class _BackgroundUpdates:
+    EVENT_ORIGIN_SERVER_TS_NAME = "event_origin_server_ts"
+    EVENT_FIELDS_SENDER_URL_UPDATE_NAME = "event_fields_sender_url"
+    DELETE_SOFT_FAILED_EXTREMITIES = "delete_soft_failed_extremities"
+
+
 @attr.s(slots=True, frozen=True)
 class _CalculateChainCover:
     """Return value for _calculate_chain_cover_txn."""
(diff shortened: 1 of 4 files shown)


callahad:
Do we also need to worry about `last_stream_ordering` on the `pushers` table?

richvdh:
> Do we also need to worry about `last_stream_ordering` on the `pushers` table?

Probably! I have yet to work through other tables to see what other problems we have. I'm assuming they are all going to be easier than `events`.

synapse/storage/databases/main/events_bg_updates.py
```diff
@@ -1009,3 +1043,75 @@ def purged_chain_cover_txn(txn) -> int:
...
+
+    async def _background_populate_stream_ordering2(
+        self, progress: JsonDict, batch_size: int
+    ) -> int:
+        """Populate events.stream_ordering2, then replace stream_ordering
+
+        This is to deal with the fact that stream_ordering was initially created as a
+        32-bit integer field.
+        """
+        batch_size = max(batch_size, 1)
```
erikjohnston, line 1055: Why is this an issue here? I'd expect setting batch_size to zero/negative anywhere is going to cause issues?

erikjohnston (commented):
At a *very* cursory glance this looks sane

anoadragon453:
> Add an index on the new column.

Is this expected to be a long-running task?

richvdh:
> > Add an index on the new column.
> 
> Is this expected to be a long-running task?

well yes. But the creation runs concurrently so shouldn't block other operations.

synapse/storage/databases/main/events_bg_updates.py
richvdh: true. It's just defensive here.

babolivier (approved):

PR #10268, opened 2021-06-29
Move methods involving event authentication to EventAuthHandler.

clokep:
Instead of being on the `Auth` object, which deals with user authentication.

This finishes some refactoring started in #9814.

~~Currently based on #10253.~~

Each commit generally makes a single change (moving a single method) although a couple had inner dependencies that I moved together.

### DIFF
--- a/synapse/handlers/event_auth.py
+++ b/synapse/handlers/event_auth.py
@@ -25,6 +25,7 @@ from synapse.api.room_versions import KNOWN_ROOM_VERSIONS, RoomVersion
 from synapse.events import EventBase
 from synapse.events.builder import EventBuilder
 from synapse.types import StateMap
+from synapse.util.metrics import Measure
 
 if TYPE_CHECKING:
     from synapse.server import HomeServer
@@ -36,6 +37,7 @@ class EventAuthHandler:
     """
 
     def __init__(self, hs: "HomeServer"):
+        self._clock = hs.get_clock()
         self._store = hs.get_datastore()
 
     async def check_from_context(
@@ -87,6 +89,10 @@ class EventAuthHandler:
                 auth_ids.append(auth_ev_id)
 
         return auth_ids
+
+    async def check_host_in_room(self, room_id: str, host: str) -> bool:
+        with Measure(self._clock, "check_host_in_room"):
+            return await self._store.is_host_joined(room_id, host)
 
     async def check_restricted_join_rules(
         self,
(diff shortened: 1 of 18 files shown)


synapse/handlers/event_auth.py
```diff
@@ -34,7 +38,191 @@ class EventAuthHandler:
...
+    async def check_can_change_room_list(self, room_id: str, user: UserID) -> bool:
+        """Determine whether the user is allowed to edit the room's entry in the
+        published room list.
+
+        Args:
+            room_id
+            user
+        """
+
+        is_admin = await self._store.is_server_admin(user)
```
clokep, line 202: This was previously calling `self.is_server_admin` when it was on `Auth`, but I don't want a dependency between `EventAuthHandler` and `Auth`. Frankly I see no point in `Auth.is_server_admin` since it just calls `store.is_server_admin`, but it gets called in many places so I didn't want to mess with that in this PR. It seemed clear enough to just directly call the store method.

clokep:
Failing tests are #10275.

synapse/handlers/event_auth.py
```diff
@@ -90,10 +92,138 @@ def compute_auth_events(
...
+                403,
+                "User %s not in room %s, and room previews are disabled"
+                % (user_id, room_id),
+            )
+
     async def check_host_in_room(self, room_id: str, host: str) -> bool:
         with Measure(self._clock, "check_host_in_room"):
             return await self._store.is_host_joined(room_id, host)
 
+    async def check_can_change_room_list(self, room_id: str, user: UserID) -> bool:
```
richvdh, line 193: > This class contains methods for authenticating events added to room graphs.

I'm not really convinced this fits here, to be honest (likewise `check_user_in_room_or_world_readable` and `check_user_in_room`). They are more abount restricting access for the C-S API.

richvdh (commented):
looks fine, but I feel like you might have gone a bit overboard. wdyt?

synapse/handlers/event_auth.py
clokep: I wasn't sure where the draw the line, but those seemed more about restricting access to adding events to the room graph, not about access to the C-S API, but maybe I missed the mark on those two. I'm happy to move those back.

richvdh: I'd vote for leaving them where they were, for now at least.

clokep: I reverted 3b07593f2e793501da7b9053616182db4da31a4c which also included `check_can_change_room_list`. There was one minor conflict in the imports.

PR #10269, opened 2021-06-29
Add a distributed lock

erikjohnston:
This adds a simple best effort locking mechanism that works cross workers.

### DIFF
--- a/changelog.d/10269.misc
+++ b/changelog.d/10269.misc
@@ -1,0 +1,1 @@
+Add a distributed lock.
--- a/changelog.d/10269.misc
+++ b/changelog.d/10269.misc
@@ -1,1 +1,1 @@
-Add a distributed lock.
+Add a distributed lock implementation.
--- a/synapse/storage/databases/main/lock.py
+++ b/synapse/storage/databases/main/lock.py
@@ -238,9 +238,8 @@ class Lock:
     """An async context manager that manages an acquired lock, ensuring it is
     regularly renewed and dropping it when the context manager exits.
 
-    The context manager returns an async function that can be called to check if
-    the lock is still valid, which can be used to bail out of processing if e.g.
-    processing work in a loop.
+    The lock object has an `is_still_valid` method which can be used to
+    double-check the lock is still valid, if e.g. processing work in a loop.
 
     For example:
 
@@ -248,11 +247,11 @@ class Lock:
         if not lock:
             return
 
-        async with lock as still_valid:
+        async with lock:
             for item in work:
                 await process(item)
 
-                if not await still_valid():
+                if not await is_still_valid():
                     break
     """
 
@@ -295,14 +294,15 @@ class Lock:
         """
         await store._renew_lock(lock_name, lock_key, token)
 
-    async def is_lock_still_valid(self) -> bool:
+    async def is_still_valid(self) -> bool:
         """Check if the lock is still held by us"""
         return await self._store._is_lock_still_valid(
             self._lock_name, self._lock_key, self._token
         )
 
-    async def __aenter__(self) -> Callable[[], Awaitable[bool]]:
-        return self.is_lock_still_valid
+    async def __aenter__(self) -> None:
+        if self._dropped:
+            raise Exception("Cannot reuse a Lock object")
 
     async def __aexit__(
         self,
--- a/synapse/storage/databases/main/lock.py
+++ b/synapse/storage/databases/main/lock.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 import logging
 from types import TracebackType
-from typing import TYPE_CHECKING, Awaitable, Callable, Dict, Optional, Tuple, Type
+from typing import TYPE_CHECKING, Dict, Optional, Tuple, Type
 
 from twisted.internet.interfaces import IReactorCore
 
--- a/synapse/storage/databases/main/lock.py
+++ b/synapse/storage/databases/main/lock.py
@@ -251,7 +251,7 @@ class Lock:
             for item in work:
                 await process(item)
 
-                if not await is_still_valid():
+                if not await lock.is_still_valid():
                     break
     """
 
--- a/synapse/storage/databases/main/lock.py
+++ b/synapse/storage/databases/main/lock.py
@@ -200,14 +200,10 @@ class LockStore(SQLBaseStore):
             allow_none=True,
             desc="is_lock_still_valid",
         )
-        # If there is an entry and it has not timed out, then we still hold the lock
-        if (
+        return (
             last_renewed_ts is not None
             and self._clock.time_msec() - _LOCK_TIMEOUT_MS < last_renewed_ts
-        ):
-            return True
-
-        return False
+        )
 
     async def _renew_lock(self, lock_name: str, lock_key: str, token: str) -> None:
         """Attempt to renew the lock if we still hold it."""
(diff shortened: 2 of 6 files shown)


synapse/storage/databases/main/lock.py
```diff
@@ -0,0 +1,322 @@
...
+                    keyvalues={
+                        "lock_name": lock_name,
+                        "lock_key": lock_key,
+                    },
+                    values={},
+                    insertion_values={
+                        "token": token,
+                        "last_renewed_ts": self._clock.time_msec(),
+                    },
+                )
```
richvdh, line 165: this is missing a value for `instance_name` I think (and hence CI is failing)

changelog.d/10269.misc
```diff
@@ -0,0 +1 @@
+Add a distributed lock.
```
richvdh, line 1:
```suggestion
Add a distributed lock implementation.
```

synapse/storage/databases/main/lock.py
```diff
@@ -0,0 +1,323 @@
...
+        self._reactor = hs.get_reactor()
+        self._instance_name = hs.get_instance_id()
+
+        # A map from `(lock_name, lock_key)` to the token of any locks that we
+        # think we currently hold.
+        self._live_tokens: Dict[Tuple[str, str], str] = {}
+
+        # When we shut down we want to remove the locks. Technically this can
+        # lead to a race, as we may drop the lock while we are still processing.
+        # However, a) it should be a small windown, b) the lock is best effort
```
richvdh, line 68:
```suggestion
        # However, a) it should be a small window, b) the lock is best effort
```

synapse/storage/databases/main/lock.py
```diff
@@ -0,0 +1,323 @@
...
+                sql = """
+                    INSERT INTO worker_locks (lock_name, lock_key, instance_name, token, last_renewed_ts)
+                    VALUES (?, ?, ?, ?, ?)
+                    ON CONFLICT (lock_name, lock_key)
+                    DO UPDATE SET
+                        token = EXCLUDED.token,
+                        instance_name = EXCLUDED.instance_name,
+                        last_renewed_ts = EXCLUDED.last_renewed_ts
+                    WHERE
+                        worker_locks.last_renewed_ts < ?
```
richvdh, line 109: I had to stare at this for a while; I failed to grok the WHERE applied only to the UPDATE.

```suggestion
                    DO UPDATE 
                        SET
                            token = EXCLUDED.token,
                            instance_name = EXCLUDED.instance_name,
                            last_renewed_ts = EXCLUDED.last_renewed_ts
                        WHERE
                            worker_locks.last_renewed_ts < ?
```

synapse/storage/databases/main/lock.py
```diff
@@ -0,0 +1,323 @@
...
+            desc="is_lock_still_valid",
+        )
+        # If there is an entry and it has not timed out, then we still hold the lock
+        if (
+            last_renewed_ts is not None
+            and self._clock.time_msec() - _LOCK_TIMEOUT_MS < last_renewed_ts
+        ):
+            return True
+
+        return False
```
clokep, line 210:
```suggestion
        return (
            last_renewed_ts is not None
            and self._clock.time_msec() - _LOCK_TIMEOUT_MS < last_renewed_ts
        )
```

synapse/storage/databases/main/lock.py
```diff
@@ -0,0 +1,323 @@
...
+            keyvalues={
+                "lock_name": lock_name,
+                "lock_key": lock_key,
+                "token": token,
+            },
+            updatevalues={"last_renewed_ts": self._clock.time_msec()},
+            desc="renew_lock",
+        )
+
+        self._live_tokens.pop((lock_name, lock_key), None)
```
richvdh, line 221: this looks wrong? should be in `_drop_lock`, no?

tests/storage/databases/main/test_lock.py
```diff
@@ -0,0 +1,84 @@
...
+
+        # We simulate the process getting stuck by cancelling the looping call
+        # that keeps the lock active.
+        lock._looping_call.stop()
+
+        # Wait for the lock to timeout.
+        self.reactor.advance(2 * _LOCK_TIMEOUT_MS / 1000)
+
+        lock2 = self.get_success(self.store.try_acquire_lock("name", "key"))
+        self.assertIsNotNone(lock2)
```
clokep, line 84: We should assert that the callback from `lock` now returns `False`, I think.

synapse/storage/databases/main/lock.py
```diff
@@ -0,0 +1,323 @@
...
+    the lock is still valid, which can be used to bail out of processing if e.g.
+    processing work in a loop.
+
+    For example:
+
+        lock = await self.store.try_acquire_lock(...)
+        if not lock:
+            return
+
+        async with lock() as still_valid:
```
richvdh, line 250: is this correct? should it be:

```suggestion
        async with lock as still_valid:
```

richvdh: also, this business of returning a callback from the context manager feels pretty confusing. Maybe just call `lock.is_lock_still_valid` from inside the loop.

synapse/storage/databases/main/lock.py
```diff
@@ -0,0 +1,323 @@
...
+        await self._store._renew_lock(self._lock_name, self._lock_key, self._token)
+
+    async def is_lock_still_valid(self) -> bool:
+        """Check if the lock is still held by us"""
+        return await self._store._is_lock_still_valid(
+            self._lock_name, self._lock_key, self._token
+        )
+
+    async def __aenter__(self) -> Callable[[], Awaitable[bool]]:
+        return self.is_lock_still_valid
```
richvdh, line 290: maybe assert that we have not dropped the lock? (ie, that we are not reusing the context manager)

synapse/storage/databases/main/lock.py
```diff
@@ -200,14 +200,10 @@ async def _is_lock_still_valid(
             allow_none=True,
             desc="is_lock_still_valid",
         )
-        # If there is an entry and it has not timed out, then we still hold the lock
```
clokep: Oops I don't think I meant to delete this comment in my suggestion.

synapse/storage/databases/main/lock.py
```diff
@@ -0,0 +1,323 @@
...
+
+        return False
+
+    def __del__(self) -> None:
+        if not self._dropped:
+            # We should not be dropped without the lock being released (unless
+            # we're shutting down), but if we are then let's at least stop
+            # renewing the lock.
+            if self._looping_call.running:
+                self._looping_call.stop()
```
richvdh, line 312: given the timer holds a reference to `renew`, and `renew` holds a reference to `self`... I suspect this is never going to be reached. You could maybe make the renew task a function with minimal closure.

richvdh (changes requested):

synapse/storage/databases/main/lock.py
erikjohnston: oh lord, yes

synapse/storage/databases/main/lock.py
erikjohnston: good point

richvdh (approved):
lgtm

synapse/storage/databases/main/lock.py
```diff
@@ -0,0 +1,334 @@
...
+
+        lock = await self.store.try_acquire_lock(...)
+        if not lock:
+            return
+
+        async with lock:
+            for item in work:
+                await process(item)
+
+                if not await is_still_valid():
```
clokep, line 254: I think?

```suggestion
                if not await lock.is_still_valid():
```

richvdh: yup, thanks

PR #10272, opened 2021-06-29
Handle inbound events from federation asynchronously

### LINKED ISSUE #9490: _room_pdu_linearizer starves other incoming transactions

erikjohnston:
Fixes #9490

Based on #10269

This will break a couple of SyTest that are expecting failures to be added to the response of a federation `/send`, which obviously doesn't happen now that things are asynchronous.

Two drawbacks:
1. Currently there is no logic to handle any events left in the staging area after restart, and so they'll only be handled on the next incoming event in that room. That can be fixed separately.
2. We now only process one event per room at a time. This can be fixed up further down the line.

### DIFF
--- a/synapse/federation/federation_server.py
+++ b/synapse/federation/federation_server.py
@@ -100,7 +100,7 @@ last_pdu_ts_metric = Gauge(
 
 # The name of the lock to use when process events in a room received over
 # federation.
-_INBOUND_EVENT_HANDLING_LOCK_NAME = "federtion_inbound_pdu"
+_INBOUND_EVENT_HANDLING_LOCK_NAME = "federation_inbound_pdu"
 
 
 class FederationServer(FederationBase):
@@ -904,7 +904,7 @@ class FederationServer(FederationBase):
                         exc_info=(f.type, f.value, f.getTracebackObject()),  # type: ignore
                     )
 
-                await self.store.remove_received_event_to_staging(
+                await self.store.remove_received_event_from_staging(
                     origin, event.event_id
                 )
 
--- a/synapse/federation/federation_server.py
+++ b/synapse/federation/federation_server.py
@@ -57,10 +57,12 @@ from synapse.logging.context import (
 )
 from synapse.logging.opentracing import log_kv, start_active_span_from_edu, trace
 from synapse.logging.utils import log_function
+from synapse.metrics.background_process_metrics import wrap_as_background_process
 from synapse.replication.http.federation import (
     ReplicationFederationSendEduRestServlet,
     ReplicationGetQueryRestServlet,
 )
+from synapse.storage.databases.main.lock import Lock
 from synapse.types import JsonDict
 from synapse.util import glob_to_regex, json_decoder, unwrapFirstError
 from synapse.util.async_helpers import Linearizer, concurrently_execute
@@ -94,6 +96,11 @@ last_pdu_ts_metric = Gauge(
     "The timestamp of the last PDU which was successfully received from the given domain",
     labelnames=("server_name",),
 )
+
+
+# The name of the lock to use when process events in a room received over
+# federation.
+_INBOUND_EVENT_HANDLING_LOCK_NAME = "federtion_inbound_pdu"
 
 
 class FederationServer(FederationBase):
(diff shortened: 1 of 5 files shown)


synapse/federation/federation_server.py
```diff
@@ -834,7 +841,87 @@ async def _handle_received_pdu(self, origin: str, pdu: EventBase) -> None:
...
+    ) -> None:
+        """Process events in the staging area for the given room.
+
+        The latest_origin and latest_event args are the latest origin and event
+        received.
+        """
+
+        # The common path is for the event we just received be the only event in
+        # the room, so instead of pulling the event out of the DB and parsing
+        # the event we just pull out the next event ID and check if that matches.
```
erikjohnston, line 871: FWIW I don't know whether the extra complexity is worth saving fetching a single event from the DB, happy to delete this logic :shrug:

synapse/storage/schema/main/delta/59/16federation_inbound_staging.sql
```diff
@@ -0,0 +1,33 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+
+-- A staging area for newly received events over federation.
+--
+-- Note we may store the same event multiple times if it comes from different
+-- servers, this is to handle the case if we get a redacted and non-redacted
```
richvdh, line 20:
```suggestion
-- servers; this is to handle the case if we get a redacted and non-redacted
```

synapse/storage/schema/main/delta/59/16federation_inbound_staging.sql
```diff
@@ -0,0 +1,33 @@
...
+-- versions of the event.
+CREATE TABLE federation_inbound_events_staging (
+    origin TEXT NOT NULL,
+    room_id TEXT NOT NULL,
+    event_id TEXT NOT NULL,
+    room_version TEXT NOT NULL,
```
richvdh, line 26: why not just re-query for this?

synapse/federation/federation_server.py
```diff
@@ -96,6 +98,11 @@
 )
 
 
+# The name of the lock to use when process events in a room received over
+# federation.
+_INBOUND_EVENT_HANDLING_LOCK_NAME = "federtion_inbound_pdu"
```
clokep, line 103:
```suggestion
_INBOUND_EVENT_HANDLING_LOCK_NAME = "federation_inbound_pdu"
```

synapse/federation/federation_server.py
```diff
@@ -834,7 +841,87 @@ async def _handle_received_pdu(self, origin: str, pdu: EventBase) -> None:
...
+                    logger.warning("Error handling PDU %s: %s", event.event_id, e)
+                except Exception:
+                    f = failure.Failure()
+                    logger.error(
+                        "Failed to handle PDU %s",
+                        event.event_id,
+                        exc_info=(f.type, f.value, f.getTracebackObject()),  # type: ignore
+                    )
+
+                await self.store.remove_received_event_to_staging(
```
richvdh, line 907:
```suggestion
                await self.store.remove_received_event_from_staging(
```

synapse/storage/databases/main/event_federation.py
```diff
@@ -1044,6 +1046,124 @@ def _delete_old_forward_extrem_cache_txn(txn):
...
+                "received_ts": self._clock.time_msec(),
+                "event_json": json_encoder.encode(event.get_dict()),
+                "internal_metadata": json_encoder.encode(
+                    event.internal_metadata.get_dict()
+                ),
+            },
+            desc="insert_received_event_to_staging",
+        )
+
+    async def remove_received_event_to_staging(
```
richvdh, line 1075:
```suggestion
    async def remove_received_event_from_staging(
```

synapse/storage/databases/main/event_federation.py
```diff
@@ -1044,6 +1046,124 @@ def _delete_old_forward_extrem_cache_txn(txn):
...
+        event_id: str,
+    ) -> None:
+        """Remove the given event from the staging area"""
+        await self.db_pool.simple_delete(
+            table="federation_inbound_events_staging",
+            keyvalues={
+                "origin": origin,
+                "event_id": event_id,
+            },
+            desc="remove_received_event_to_staging",
```
richvdh, line 1087:
```suggestion
            desc="remove_received_event_from_staging",
```

synapse/storage/databases/main/event_federation.py
```diff
@@ -1044,6 +1046,124 @@ def _delete_old_forward_extrem_cache_txn(txn):
...
+        room_version = KNOWN_ROOM_VERSIONS.get(room_version_str)
+
+        if not room_version:
+            logger.warning(
+                "Event %s in room %s has unknown room version %s",
+                event_id,
+                room_id,
+                room_version_str,
+            )
+            await self.remove_received_event_to_staging(
```
richvdh, line 1153:
```suggestion
            await self.remove_received_event_from_staging(
```

richvdh (commented):
looks generally about right

synapse/federation/federation_server.py
richvdh: :shrug:

synapse/storage/schema/main/delta/59/16federation_inbound_staging.sql
erikjohnston: Oh, d'oh, I forgot we must already have it in the DB

PR #10276, opened 2021-06-29
Fix messages from multiple senders in historical chunk (MSC2716)

MadLittleMods:
Fix messages from multiple senders in historical chunk. This also means that an app service does not need to define `?user_id` when using this endpoint.

Follow-up to https://github.com/matrix-org/synapse/pull/9247

Complement test for this case added by https://github.com/matrix-org/complement/pull/68/commits/e56c87281d3b1ac7ac9edd74cc46ea4f620183b0

---

Previously, Synapse would throw a [403, `Cannot force another user to join.`](https://github.com/matrix-org/synapse/blob/bc5589a1bb5fc9fbbd64f2dec3335e8eccd43640/synapse/event_auth.py#L346), because we were trying to use `?user_id` from a single virtual user which did not match with messages from other users in the chunk.

### Pull Request Checklist


synapse/rest/client/v1/room.py
```diff
@@ -510,7 +523,9 @@ async def on_POST(self, request, room_id):
         # Prepend the insertion event to the start of the chunk
         events_to_create = [insertion_event] + events_to_create
 
-        inherited_depth = await self.inherit_depth_from_prev_ids(prev_events_from_query)
+        inherited_depth = await self._inherit_depth_from_prev_ids(
```
MadLittleMods, line 526: Just renaming this with an underscore to align with the other internal functions

synapse/rest/client/v1/room.py
```diff
@@ -309,7 +311,14 @@ def __init__(self, hs):
         self.room_member_handler = hs.get_room_member_handler()
         self.auth = hs.get_auth()
 
-    async def inherit_depth_from_prev_ids(self, prev_event_ids) -> int:
+    def _copy_requester_and_override_user_id(self, requester, new_user_id):
+        serialized_requester = requester.serialize()
+        serialized_requester["user_id"] = new_user_id
+        new_requester = Requester.deserialize(self.store, serialized_requester)
+
+        return new_requester
```
MadLittleMods, line 319: Is there a better way to copy this efficiently?

erikjohnston: TBH I think in this case we probably explicitly want to set the fields, as I don't think we want to copy over e.g. `device_id` and the like? So I think we can just do `create_requester(user_id, app_server_id=requester.app_server_id)`?

erikjohnston:
It's just occurred to me that we do some checks when an appservice tries to send an event on behalf of a user that we should ensure we still do here:

https://github.com/matrix-org/synapse/blob/8d609435c0053fc4decbc3f9c3603e728912749c/synapse/api/auth.py#L267-L270

erikjohnston:
(Otherwise the code looks fine)

erikjohnston (changes requested):
As per comment

MadLittleMods:
> It's just occurred to me that we do some checks when an appservice tries to send an event on behalf of a user that we should ensure we still do here:
> 
> https://github.com/matrix-org/synapse/blob/8d609435c0053fc4decbc3f9c3603e728912749c/synapse/api/auth.py#L267-L270

Good shout! I found some bugs in my Complement tests around this after adding the validation 🙌

synapse/rest/client/v1/room.py
```diff
@@ -378,6 +381,38 @@ def _create_insertion_event_dict(
...
+        elif not app_service.is_interested_in_user(user_id):
+            raise AuthError(
+                403,
+                "Application service cannot masquerade as this user (%s)." % user_id,
+            )
+        # Check to make sure the user is already registered on the homeserver
+        elif not (await self.store.get_user_by_id(user_id)):
+            raise AuthError(
+                403, "Application service has not registered this user (%s)" % user_id
+            )
```
MadLittleMods, line 412: This logic is based off of 

https://github.com/matrix-org/synapse/blob/d26094e92cace20525552e5a0c8b21ff9ce53f11/synapse/api/auth.py#L264-L270

erikjohnston (approved):

MadLittleMods:
Thanks for the review @erikjohnston 🐷

PR #10286, opened 2021-06-30
Change more stream_ordering columns to BIGINT

richvdh:
Another part of #8255.

### DIFF
--- a/changelog.d/10286.bugfix
+++ b/changelog.d/10286.bugfix
@@ -1,0 +1,1 @@
+Fix a long-standing bug where Synapse would return errors after 2<sup>31</sup> events were handled by the server.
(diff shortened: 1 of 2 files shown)


anoadragon453 (approved):
Most of these tables are ~20 rows on matrix.org, with the only outlier being pushers at under ~1 mil.

Thus these seem fine to run, and I don't see any other relevant rows to change in our most recent full.postgres snapshot.

callahad:
Two concerns:

- If I need to roll back Synapse to 1.37.1 for some reason, does this break me?

    I don't _think_ so, since anything we expect to touch this data should work equally well with any integral numeric type?

- Do we have any indexes here that need special handling?

    I don't _see_ any, but that's only from a cursory grep of the schema directory.

richvdh:
> Two concerns:
> 
>     * If I need to roll back Synapse to 1.37.1 for some reason, does this break me?
>       I don't _think_ so, since anything we expect to touch this data should work equally well with any integral numeric type?

correct.

>     * Do we have any indexes here that need special handling?
>       I don't _see_ any, but that's only from a cursory grep of the schema directory.

no, there are no indexes on these columns.

PR #10289, opened 2021-06-30
Replace `room_depth.min_depth` with a BIGINT

richvdh:
while I'm dealing with INTEGERs and BIGINTs, let's replace room_depth.min_depth with a BIGINT.


synapse/storage/schema/main/delta/61/03recreate_min_depth.py
```diff
@@ -0,0 +1,70 @@
...
+       """
+    )
+
+    # and another to switch them over once it completes.
+    cur.execute(
+        """
+        INSERT INTO background_updates (ordering, update_name, progress_json, depends_on) VALUES
+            (6103, 'replace_room_depth_min_depth', '{}', 'populate_room_depth2')
+        """
+    )
```
erikjohnston, line 66: Why does this need to be in a python delta rather than a `.sql.postgers` delta?

erikjohnston (approved):
lgtm otherwise

synapse/storage/schema/main/delta/61/03recreate_min_depth.py
erikjohnston: Ah. That's sad but fair

PR #10290, opened 2021-06-30
Add script for getting info about recently registered users

erikjohnston:
Usage:

```
$ ./scripts/synapse_review_recent_signups  --help
usage: synapse_review_recent_signups [-h] [-c CONFIG_FILE] [-s duration] [-e]

optional arguments:
  -h, --help            show this help message and exit
  -c CONFIG_FILE, --config-path CONFIG_FILE
                        The config files for Synapse.
  -s duration, --since duration
                        Specify how far back to review user registrations for, defaults to 7d (i.e. 7 days).
  -e, --exclude-emails  Exclude users that have validated email addresses
```

Sample output:

```
User ID: @anon-20210630_165107-115:localhost:8800
Created: 2021-06-30 17:52:05
Email: None
IPs: 127.0.0.1
Number joined public rooms: 0 
Number joined private rooms: 0

User ID: @anon-20210630_165107-122:localhost:8800
Created: 2021-06-30 17:52:09
Email: None
IPs: 127.0.0.1
Number joined public rooms: 1 (#test-20210630_165107-10:localhost:8800)
Number joined private rooms: 0

User ID: @anon-20210630_165107-124:localhost:8800
Created: 2021-06-30 17:52:11
Email: None
IPs: 127.0.0.1
Number joined public rooms: 8 (#test-20210630_165107-15:localhost:8800, #test-20210630_165107-12:localhost:8800, #test-20210630_165107-14:localhost:8800, ...)
Number joined private rooms: 0
```

### DIFF
--- a/debian/hash_password.1
+++ b/debian/hash_password.1
@@ -1,90 +1,58 @@
-.\" generated with Ronn/v0.7.3
-.\" http://github.com/rtomayko/ronn/tree/0.7.3
-.
-.TH "HASH_PASSWORD" "1" "February 2017" "" ""
-.
+.\" generated with Ronn-NG/v0.8.0
+.\" http://github.com/apjanke/ronn-ng/tree/0.8.0
+.TH "HASH_PASSWORD" "1" "July 2021" "" ""
 .SH "NAME"
 \fBhash_password\fR \- Calculate the hash of a new password, so that passwords can be reset
-.
 .SH "SYNOPSIS"
 \fBhash_password\fR [\fB\-p\fR|\fB\-\-password\fR [password]] [\fB\-c\fR|\fB\-\-config\fR \fIfile\fR]
-.
 .SH "DESCRIPTION"
 \fBhash_password\fR calculates the hash of a supplied password using bcrypt\.
-.
 .P
 \fBhash_password\fR takes a password as an parameter either on the command line or the \fBSTDIN\fR if not supplied\.
-.
 .P
 It accepts an YAML file which can be used to specify parameters like the number of rounds for bcrypt and password_config section having the pepper value used for the hashing\. By default \fBbcrypt_rounds\fR is set to \fB10\fR\.
-.
 .P
 The hashed password is written on the \fBSTDOUT\fR\.
-.
 .SH "FILES"
 A sample YAML file accepted by \fBhash_password\fR is described below:
-.
 .P
 bcrypt_rounds: 17 password_config: pepper: "random hashing pepper"
-.
 .SH "OPTIONS"
-.
 .TP
 \fB\-p\fR, \fB\-\-password\fR
 Read the password form the command line if [password] is supplied\. If not, prompt the user and read the password form the \fBSTDIN\fR\. It is not recommended to type the password on the command line directly\. Use the STDIN instead\.
-.
 .TP
 \fB\-c\fR, \fB\-\-config\fR
 Read the supplied YAML \fIfile\fR containing the options \fBbcrypt_rounds\fR and the \fBpassword_config\fR section containing the \fBpepper\fR value\.
-.
 .SH "EXAMPLES"
 Hash from the command line:
-.
 .IP "" 4
-.
 .nf
-
 $ hash_password \-p "p@ssw0rd"
 $2b$12$VJNqWQYfsWTEwcELfoSi4Oa8eA17movHqqi8\.X8fWFpum7SxZ9MFe
-.
 .fi
-.
 .IP "" 0
-.
 .P
 Hash from the STDIN:
-.
 .IP "" 4
-.
 .nf
-
 $ hash_password
 Password:
 Confirm password:
 $2b$12$AszlvfmJl2esnyhmn8m/kuR2tdXgROWtWxnX\.rcuAbM8ErLoUhybG
-.
 .fi
-.
 .IP "" 0
-.
 .P
 Using a config file:
-.
 .IP "" 4
-.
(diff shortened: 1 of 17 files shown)


dklimpel:
I am think about what is the preferred way to do this action. Shell vs. admin API.
Perhaps it is a good idea to move part of the code to a handler to reuse the code for an admin API in the future.

synapse/_scripts/review_recent_signups.py
```diff
@@ -0,0 +1,163 @@
...
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "-c",
+        "--config-path",
+        action="append",
+        metavar="CONFIG_FILE",
+        help="The config files for Synapse.",
+    )
```
richvdh, line 99: this doesn't seem to have a default, and yet we assume it will exist, below.

It'll need to support split config files to work with the Debian packaging.

For bonus points, it could do with finding the right config file on Debian (this might be a matter of tweaking the defaults in the Debian packaging somehow)

scripts/synapse_review_recent_signups
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env python
```
richvdh, line 1: does this need adding to setup.py or some bit of debian arcana to end up in the right place?

callahad:
Suggestion: Could we add an `--interactive` flag which allows admins to `[y/n]` their way through whether to deactivate users? Would probably do it in two phases: `[y/n]` everyone, then print out a list for review and final confirmation before _actually_ deactivating?

synapse/_scripts/review_recent_signups.py
```diff
@@ -0,0 +1,163 @@
...
+        "-s",
+        "--since",
+        metavar="duration",
+        help="Specify how far back to review user registrations for, defaults to 7d (i.e. 7 days).",
+        default="7d",
+    )
+    parser.add_argument(
+        "-e",
+        "--exclude-emails",
+        action="store_false",
```
richvdh, line 110: omg this is a headf*ck. Can we do `store_true` and have `include_users_with_email = not config_args.exclude_emails` below?

richvdh (commented):
a few drive-by comments

scripts/synapse_review_recent_signups
erikjohnston: Oh lord looks like there is a bunch of stuff, including man pages, for debian packaging.

synapse/_scripts/review_recent_signups.py
erikjohnston: > this doesn't seem to have a default, and yet we assume it will exist, below.

Whoops (I cribbed it from the arguments for the main synapse process, but turns out we manually check in that case)

> It'll need to support split config files to work with the Debian packaging.

Yup, it does thanks to `action="append"` and `find_config_files(..)`.

> For bonus points, it could do with finding the right config file on Debian (this might be a matter of tweaking the defaults in the Debian packaging somehow)

Good idea

erikjohnston:
The format is now:

```
# Created: 2021-07-01 11:12:15
# Email: None
# IPs: 127.0.0.1
# Number joined public rooms: 1 (!SatAVWGFmetsyvcIau:localhost:8800)
# Number joined private rooms: 0
#
@anon-20210701_101211-1:localhost:8800

# Created: 2021-07-01 11:12:15
# Email: None
# IPs: 127.0.0.1
# Number joined public rooms: 1 (!yOyZKGtuHsXMQNJyNB:localhost:8800)
# Number joined private rooms: 0
#
@anon-20210701_101211-2:localhost:8800
```

Which allows easier filtering out of the context.

I've also added a `-u/--only-users` option to only print the user IDs.

erikjohnston:
> Suggestion: Could we add an `--interactive` flag which allows admins to `[y/n]` their way through whether to deactivate users? Would probably do it in two phases: `[y/n]` everyone, then print out a list for review and final confirmation before _actually_ deactivating?

We decided not to do this for now.

MurzNN:
Will be good to show also number of total sent messages in interval, or even better - grouped per day... Is it hard to add?

MurzNN:
Also useful metric to identify spammers can be number of sent messages per room.

debian/hash_password.1
```diff
@@ -1,90 +1,58 @@
-.\" generated with Ronn/v0.7.3
```
richvdh, line 1: random thought: I wonder if we should generate these at build time rather than checking them in...

synapse/_scripts/review_recent_signups.py
```diff
@@ -0,0 +1,175 @@
...
```
richvdh, line 111: this is still super confusing, and I *think* the logic is now backwards.

richvdh (commented):
lgtm other than `exclude_emails`.

synapse/_scripts/review_recent_signups.py
erikjohnston: Bah, I did mean to change that to `store_true`. Manually tested and fixed

debian/hash_password.1
erikjohnston: That probably makes sense

richvdh (approved):
lgtm

erikjohnston:
> Will be good to show also number of total sent messages (and invites separately) in interval, or even better - grouped per day... Is it hard to add?

Those would be useful metrics, but I think those can be added as later PRs :slightly_smiling_face:

PR #10298, opened 2021-07-01
Show joinable rooms in the spaces summary.

### LINKED ISSUE #10255: Update spaces summary to use join rules, not history visibility
Now that [MSC3173](https://github.com/matrix-org/matrix-doc/pull/3173) has merged we've updated [MSC2946](https://github.com/matrix-org/matrix-doc/pull/2946) to mention that join rules, not history visibility, should be used when returning results in the spaces summary.

We need to update the spaces summary to include any rooms that are *joinable* as opposed to those with world readable history visibility.

clokep:
Per [MSC3173](https://github.com/matrix-org/matrix-doc/pull/3173) we can expose the stripped state (which is what the [spaces summary](https://github.com/matrix-org/matrix-doc/pull/2946) returns).

This updates the "is room accessible" check to also check join rules (in addition to history visibility).

Fixes #10255


clokep:
This does pass the current complement tests, but I should update them be more thorough.

synapse/handlers/space_summary.py
```diff
@@ -420,9 +421,8 @@ async def _is_room_accessible(
 
         It should be included if:
 
-        * The requester is joined or invited to the room.
-        * The requester can join without an invite (per MSC3083).
-        * The origin server has any user that is joined or invited to the room.
+        * The requester is joined or can join the room (per MSC3173).
+        * The origin server has any user that is joined or can join the room.
         * The history visibility is set to world readable.
```
clokep, line 426: I debated making all of this a method on `Auth`: `can_user_view_stripped_state` or something, but seemed a bit silly until we want to use it in another place.

synapse/handlers/space_summary.py
```diff
@@ -446,8 +446,30 @@ async def _is_room_accessible(
...
-        # stripped state in the room.
+        # Include the room if it has join rules of public or knock.
+        join_rules_event_id = state_ids.get((EventTypes.JoinRules, ""))
+        if join_rules_event_id:
+            join_rules_event = await self._store.get_event(join_rules_event_id)
+            join_rule = join_rules_event.content.get("join_rule")
+            if (
+                join_rule == JoinRules.PUBLIC
+                or room_version.msc2403_knocking
+                and join_rule == JoinRules.KNOCK
```
erikjohnston, line 457:
```suggestion
                or (room_version.msc2403_knocking
                and join_rule == JoinRules.KNOCK)
```

otherwise I start worrying about operator precedence, which is something I'd rather not think about :slightly_smiling_face:

clokep:
I realized this doesn't do the filtering when receiving data back over federation so I'll need to add that.

clokep:
> I realized this doesn't do the filtering when receiving data back over federation so I'll need to add that.

This has now been added and additional tests are included.

erikjohnston (approved):

PR #10303, opened 2021-07-02
Handle old staged inbound events

### LINKED ISSUE #10295: Handle any events left in the staging area after restart

erikjohnston:
We might have events in the staging area if the service was restarted while there were unhandled events in the staging area.

Fixes #10295

---

Currently this works by starting up a background task when we see an incoming transaction. The reason for doing it that way is that we don't know what know which workers are handling the incoming federation traffic. The downside to this is that there will be a delay in beginning to process old events until we see a new incoming transaction.

An alternative would be to always handle on e.g. the worker configured to run background tasks?

### DIFF
--- a/synapse/federation/federation_server.py
+++ b/synapse/federation/federation_server.py
@@ -915,6 +915,13 @@ class FederationServer(FederationBase):
             self._process_incoming_pdus_in_room_inner(
                 pdu.room_id, room_version, lock, origin, pdu
             )
+        else:
+            logger.info(
+                "Previously received events in the room are already being"
+                " processed; event has been added to the queue: %s in %s",
+                pdu.event_id,
+                pdu.room_id,
+            )
 
     @wrap_as_background_process("_process_incoming_pdus_in_room_inner")
     async def _process_incoming_pdus_in_room_inner(
--- a/synapse/federation/federation_server.py
+++ b/synapse/federation/federation_server.py
@@ -951,6 +951,7 @@ class FederationServer(FederationBase):
             ) = await self.store.get_next_staged_event_id_for_room(room_id)
             if next_origin != latest_origin or next_event_id != latest_event.event_id:
                 latest_origin = None
+                latest_event = None
 
         if latest_origin is None or latest_event is None:
             next = await self.store.get_next_staged_event_for_room(
--- a/synapse/federation/federation_server.py
+++ b/synapse/federation/federation_server.py
@@ -148,6 +148,52 @@ class FederationServer(FederationBase):
 
         self._room_prejoin_state_types = hs.config.api.room_prejoin_state
 
+        # Whether we have started handling old events in the staging area.
+        self._started_handling_of_staged_events = False
+
+    def _start_handling_old_staged_events(self) -> None:
+        """Start handling old events in the staging area. This should be called
+        when we receive a new incoming transaction (and are thus an instance
+        that handles incoming federation).
+        """
+        if self._started_handling_of_staged_events:
+            return
+
+        self._started_handling_of_staged_events = True
+
+        self._handle_old_staged_events()
+
+    @wrap_as_background_process("_handle_old_staged_events")
+    async def _handle_old_staged_events(self) -> None:
+        """Handle old staged events by fetching all rooms that have staged
+        events and start the processing of each of those rooms.
+        """
+
+        # Get all the rooms IDs with staged events.
+        room_ids = await self.store.get_all_staged_rooms()
+
+        # We then shuffle them so that if there are multiple instances doing
+        # this work they're less likely to collide.
+        random.shuffle(room_ids)
+
+        for room_id in room_ids:
+            room_version = await self.store.get_room_version(room_id)
+
+            # Try and acquire the processing lock for the room, if we get it start a
+            # background process for handling the events in the room.
+            lock = await self.store.try_acquire_lock(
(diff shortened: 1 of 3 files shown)


synapse/federation/federation_server.py
```diff
@@ -148,6 +148,52 @@ def __init__(self, hs: "HomeServer"):
...
+        for room_id in room_ids:
+            room_version = await self.store.get_room_version(room_id)
+
+            # Try and acquire the processing lock for the room, if we get it start a
+            # background process for handling the events in the room.
+            lock = await self.store.try_acquire_lock(
+                _INBOUND_EVENT_HANDLING_LOCK_NAME, room_id
+            )
+            if lock:
+                self._process_incoming_pdus_in_room_inner(
```
richvdh, line 188: probably worth logging something here.

synapse/federation/federation_server.py
```diff
@@ -148,6 +148,52 @@ def __init__(self, hs: "HomeServer"):
...
+        """Start handling old events in the staging area. This should be called
+        when we receive a new incoming transaction (and are thus an instance
+        that handles incoming federation).
+        """
+        if self._started_handling_of_staged_events:
+            return
+
+        self._started_handling_of_staged_events = True
+
+        self._handle_old_staged_events()
```
richvdh, line 164: given this is only used in one place, I'd be tempted to inline it.

synapse/federation/federation_server.py
```diff
@@ -882,32 +932,40 @@ async def _process_incoming_pdus_in_room_inner(
...
+        if latest_event is not None and latest_origin is not None:
+            (
+                next_origin,
+                next_event_id,
+            ) = await self.store.get_next_staged_event_id_for_room(room_id)
+            if next_origin == latest_origin and next_event_id == latest_event.event_id:
+                found_origin = latest_origin
+                found_event = latest_event
+
+        if found_origin is None or found_event is None:
```
richvdh, line 958: I would simplify this by just clearing `latest_event` if it is found unacceptable:

```suggestion
            if next_origin != latest_origin or next_event_id != latest_event.event_id:
                latest_origin = None

        if latest_origin is None or latest_event is None:
```

(and then `s/found_/latest_/` below)

ymmv though

synapse/federation/federation_server.py
```diff
@@ -882,32 +932,40 @@ async def _process_incoming_pdus_in_room_inner(
...
             next = await self.store.get_next_staged_event_for_room(
                 room_id, room_version
             )
             if not next:
                 return
 
             origin, event = next
+        else:
+            origin = found_origin
+            event = found_event
```
richvdh, line 968: I would simplify this by just clearing latest_event if it is found unacceptable:

```suggestion
        if latest_event is not None and latest_origin is not None:
            (
                next_origin,
                next_event_id,
            ) = await self.store.get_next_staged_event_id_for_room(room_id)
            if next_origin != latest_origin or next_event_id != latest_event.event_id:
                latest_origin = None

        if latest_origin is None or latest_event is None:
            next = await self.store.get_next_staged_event_for_room(
                room_id, room_version
            )
            if not next:
                return

            origin, event = next
        else:
            origin = latest_origin
            event = latest_event
```
ymmv though.

synapse/storage/databases/main/event_federation.py
```diff
@@ -1207,6 +1207,15 @@ def _get_next_staged_event_for_room_txn(txn):
 
         return origin, event
 
+    async def get_all_staged_rooms(self) -> List[str]:
```
richvdh, line 1210: given that our database methods aren't really scoped, I think this needs a better name:

```suggestion
    async def get_all_rooms_with_staged_incoming_events(self) -> List[str]:
```

richvdh (commented):
gneerally fine, a few comments

synapse/federation/federation_server.py
erikjohnston: I think the common case is that we get the lock, so I wonder if it'd be better to log when we *don't* acquire the lock? (From reducing our log lines perspective?)

erikjohnston: I've done that for now, but I'm not particularly happy about the wording.

erikjohnston: (Also note we log `handling received PDU` when we come round to actually handling an event)

richvdh: hrmmm. given this should only happen once (per room with staged events) at startup -  I don't feel like the logging would be too onerous?

synapse/federation/federation_server.py
```diff
@@ -925,6 +915,13 @@ async def _handle_received_pdu(self, origin: str, pdu: EventBase) -> None:
             self._process_incoming_pdus_in_room_inner(
                 pdu.room_id, room_version, lock, origin, pdu
             )
+        else:
+            logger.info(
+                "Previously received events in the room are already being"
+                " processed; event has been added to the queue: %s in %s",
+                pdu.event_id,
+                pdu.room_id,
+            )
```
richvdh, line 924: I'm not sure if this is needed, but :shrug:

richvdh (commented):
looks good other than the logging

synapse/federation/federation_server.py
erikjohnston: I completely misread which this was. Sorry.

richvdh (approved):
lgtm

PR #10315, opened 2021-07-05
Add tests to characterise the current behaviour of R30 phone-home metrics

reivilibre:
I'm working on adding a new variant of R30.

This PR first characterises the current behaviour of the current R30 metric, by adding unit tests.

### DIFF
--- a/tests/app/test_phone_stats_home.py
+++ b/tests/app/test_phone_stats_home.py
@@ -1,6 +1,7 @@
 import synapse
 from synapse.rest.client.v1 import login, room
 
+from tests import unittest
 from tests.unittest import HomeserverTestCase
 
 DAY = 86400
@@ -13,6 +14,9 @@ class PhoneHomeTestCase(HomeserverTestCase):
         login.register_servlets,
     ]
 
+    # Override the retention time for the user_ips table because otherwise it
+    # gets pruned too aggressively for our R30 test.
+    @unittest.override_config({"user_ips_max_age": "365d"})
     def test_r30_minimum_usage(self):
         """
         Tests the minimum amount of interaction necessary for the R30 metric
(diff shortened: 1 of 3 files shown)


tests/app/test_phone_stats_home.py
```diff
@@ -0,0 +1,99 @@
+import synapse
+from synapse.rest.client.v1 import login, room
+
+from tests import unittest
+from tests.unittest import HomeserverTestCase
+
+DAY = 86400
```
erikjohnston, line 7:
```suggestion
ONE_DAY_IN_SECONDS = 86400
```

?

tests/app/test_phone_stats_home.py
```diff
@@ -0,0 +1,99 @@
...
+class PhoneHomeTestCase(HomeserverTestCase):
+    servlets = [
+        synapse.rest.admin.register_servlets_for_client_rest_resource,
+        room.register_servlets,
+        login.register_servlets,
+    ]
+
+    # Override the retention time for the user_ips table because otherwise it
+    # gets pruned too aggressively for our R30 test.
+    @unittest.override_config({"user_ips_max_age": "365d"})
```
erikjohnston, line 19: Does this mean that R30 doesn't work out of the box?

tests/app/test_phone_stats_home.py
```diff
@@ -0,0 +1,99 @@
...
+        # (Make sure the user isn't somehow counted by this point.)
+        r30_results = self.get_success(self.hs.get_datastore().count_r30_users())
+        self.assertEqual(r30_results, {"all": 0})
+
+        # Send a message (this counts as activity)
+        self.helper.send(room_id, "message2", tok=access_token)
+
+        # We have to wait up to 5 seconds for _update_client_ips_batch to get
+        # called and update the user_ips table in a batch.
+        self.reactor.advance(5.1)
```
erikjohnston, line 49: I'd advance by a couple of hours, as we don't really care if it takes 5s vs 1hr for the stats to update. so no point asserting that it takes less than 5s?

erikjohnston (changes requested):

tests/app/test_phone_stats_home.py
reivilibre: The default is 28d, so out of the box, it does seem like it's slightly broken, yes...

Out of the box, it can only count users who had activity in the last **28** days, where that activity occurred at least 30 days after registration.

The intention would be to count users who had activity in the last **30** days, where that activity occurred at least 30 days after registration, so it's not a MASSIVE difference.

erikjohnston: I'm just wary of testing a feature by first changing the default config? Can we update the test to work with the default, or do we need to update the default?

reivilibre: In my opinion, the default should be changed. (However, I'm not sure how worthwhile that is since the idea is to add R30v2 and eventually remove R30 [I expect]...).
Updating the test to work with the default wouldn't really show what R30 actually means, it'd instead be R28 or something ;p.
I was hoping to basically just add a test to characterise the current behaviour and wasn't intending to change anything, though...

Suppose I could make a copy of the test that uses the default settings? What do you think is preferable?

erikjohnston: Yeah, I guess its fair to merge as is.

erikjohnston: Actually, yes, I think we should have a test for the default, since this is for phone home stats.

erikjohnston (approved):
Can you make an issue for the whole R30 vs 28 days thing?

PR #10316, opened 2021-07-05
Rebuild event auth when rebuilding an event after a call to a `ThirdPartyEventRules` module

babolivier:
Because modules might send extra state events when processing an event (e.g. https://github.com/matrix-org/synapse-dinsic/pull/100), and in some cases these extra events might get dropped if we don't recalculate the initial event's auth.

### DIFF
--- a/changelog.d/10316.misc
+++ b/changelog.d/10316.misc
@@ -1,0 +1,1 @@
+Rebuild event context and auth when processing specific results from `ThirdPartyEventRules` modules.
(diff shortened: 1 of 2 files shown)


changelog.d/10316.misc
```diff
@@ -0,0 +1 @@
+Rebuild event context and auth when processing specific results from `ThirdPartyEventRules` modules.
```
anoadragon453, line 1: > when processing specific results

This is a little vague. How about "Rebuild context for events that have been modified by `ThirdPartEventRules` modules"?

anoadragon453 (approved):

changelog.d/10316.misc
babolivier: That doesn't work either, because modules can just return the event dict without modifying it to trick Synapse into rebuilding the context and auth (which is what we do in https://github.com/matrix-org/synapse-dinsic/pull/100)

PR #10323, opened 2021-07-06
build the docs for master

richvdh:
as well as running the job for master, I've reworked the logic for where we deploy to.

Firstly, we shouldn't deploy to `develop` unless we're actually building `develop` (as otherwise we can end up rewinding the develop docs to a release build).

Secondly, I've ripped out the RC logic. I wasn't at all convinced it was helpful. Consider a release cycle:
 1. fork the release branch
 2. land a release-blocking bugfix or two
 3. cut an RC: bump the synapse version
 4. cut the final release: bump the synapse version

It is odd to me that we will deploy the docs for the release branch ant steps 1, 2 and 4 but not step 3.

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -27,40 +27,28 @@ jobs:
       - name: Build the documentation
         run: mdbook build
 
-      # Deploy to the latest documentation directories
-      - name: Deploy latest documentation
+      # Figure out the target directory.
+      #
+      # The target directory is normally the same as the branch name, but we
+      # strip 'release-' from the name for release branches.
+      #
+      - name: Get the target directory name
+        id: vars
+        run: |
+          # first strip the 'refs/heads/' prefix with some shell foo
+          branch="${GITHUB_REF#refs/heads/}"
+
+          # now strip 'release-'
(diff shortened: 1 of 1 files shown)


babolivier (approved):

anoadragon453 (approved):
Looks excellent - thank you.

Note that this *may* conflict with https://github.com/matrix-org/synapse/pull/10242/files while merging master -> develop.

anoadragon453:
> It is odd to me that we will deploy the docs for the release branch ant steps 1, 2 and 4 but not step 3.

Oh, however note that this will show `vX.Y` in the eventual version dropdown on the docs site before that version of Synapse is actually released.

richvdh:
> Oh, however note that this will show vX.Y in the eventual version dropdown on the docs site before that version of Synapse is actually released. I'm not sure if that's something we want?

well right, but again, filtering out the RCs doesn't help with that.

I think we probably don't care too much about docs leaking out early.

callahad:
Bikeshed: Do we want to go with `master` here, or should we map it to `latest` (or `stable`) in our URLs?

richvdh:
yeah maybe `latest` would be better

PR #10324, opened 2021-07-06
Minor changes to `user_daily_visits`

reivilibre:
Change 1: we use the clock as the source of time in _get_start_of_day, because otherwise in (WIP) tests, we end up in a situation where we're searching (real time) < ... <= (fake time), which typically yields no results, as real time is much bigger than fake time.

Change 2: Use `? <= last_seen AND last_seen < ?` instead of `? < last_seen AND last_seen <= ?`, because the latter ignores changes that occur in the same millisecond after the query is run (which is probably a bit pedantic for real life, but in tests, causes trouble).

### DIFF
--- a/synapse/storage/databases/main/metrics.py
+++ b/synapse/storage/databases/main/metrics.py
@@ -352,7 +352,7 @@ class ServerMetricsStore(EventPushActionsWorkerStore, SQLBaseStore):
                     ) udv
                     ON u.user_id = udv.user_id AND u.device_id=udv.device_id
                     INNER JOIN users ON users.name=u.user_id
-                    WHERE last_seen > ? AND last_seen <= ?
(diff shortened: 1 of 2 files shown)


synapse/storage/databases/main/metrics.py
```diff
@@ -320,7 +320,7 @@ def _get_start_of_day(self):
         """
         Returns millisecond unixtime for start of UTC day.
         """
-        now = time.gmtime()
+        now = time.gmtime(self.hs.get_clock().time())
```
erikjohnston, line 323: We try and avoid using `self.hs` directly where possible (it can hide problems that should be caught at start up until later).

```suggestion
        now = time.gmtime(self._clock.time())
```

erikjohnston (approved):

Issue #10330, opened 2021-07-06
Install instructions point to 404 for postgresql


turt2live:
https://matrix-org.github.io/synapse/develop/setup/installation.html#using-postgresql links to https://matrix-org.github.io/synapse/develop/setup/docs/postgres.html which doesn't go anywhere.

dklimpel:
This issue is solved.

richvdh:
well it will be as soon as we merge release-v1.38 to develop. Thanks @dklimpel !

PR #10331, opened 2021-07-07
Fix broken links in INSTALL.md

dklimpel:
Fixes: #10330

In this case use absolute URL like `https://matrix-org.github.io/synapse/develop/federate.html`.
The alternative is to use relative URL like `../federate.md`

Reason is the import in:
https://github.com/matrix-org/synapse/blob/bc60f999e866bef27e97a53f70ca0fc87908697e/docs/setup/installation.md#L7

Absolute URL are at a disadvantage when there is a version picker in the documentation.
Relative URL let break the native use of https://github.com/matrix-org/synapse/blob/develop/INSTALL.md

### Pull Request Checklist


richvdh:
It feels like a mistake to use absolute links here. 

I think maybe we should just do something similar to what we did in https://github.com/matrix-org/synapse/blob/develop/UPGRADE.rst and move the install docs into the main docs tree.

richvdh (approved):
thanks! once again I am going to merge this into the release branch.

PR #10332, opened 2021-07-07
Add a new version of the R30 phone-home metric, which removes a false impression of retention given by the old R30 metric

reivilibre:
The old R30 metric can be tricked into thinking a user is 'retained' if, any point after 30 days of registration, they return even for just one 'day'. That said, a user which has been inactive for (e.g.) 6 months and opens up (e.g.) Element briefly (e.g. to check a new feature they heard about) can hardly be called 'retained'.

R30v2 instead counts users which have had at least 2 active days in the past 60 days, and where the earliest of these is at least 30 days earlier than the last. This more closely matches a sensible definition of 30-day retention.

This PR adds the R30v2 metric, some tests which characterise it, and plugs it into the phone-home statistics (which will be received by Panopticon).

### DIFF
--- a/synapse/app/phone_stats_home.py
+++ b/synapse/app/phone_stats_home.py
@@ -107,6 +107,10 @@ async def phone_stats_home(hs, stats, stats_process=_stats_process):
     for name, count in r30_results.items():
         stats["r30_users_" + name] = count
 
+    r30v2_results = await hs.get_datastore().count_r30_users()
+    for name, count in r30v2_results.items():
+        stats["r30v2_users_" + name] = count
+
     stats["cache_factor"] = hs.config.caches.global_factor
     stats["event_cache_size"] = hs.config.caches.event_cache_size
 
--- a/synapse/storage/databases/main/metrics.py
+++ b/synapse/storage/databases/main/metrics.py
@@ -404,19 +404,20 @@ class ServerMetricsStore(EventPushActionsWorkerStore, SQLBaseStore):
 
             # This is the 'all users' count.
             sql = """
-                SELECT
-                    DISTINCT COUNT(*) OVER ()
-                FROM
-                    user_daily_visits
-                WHERE
-                    timestamp > (? * 1000)
-                    AND
-                    timestamp < (? * 1000)
-                GROUP BY
-                    user_id
-                HAVING
-                    max(timestamp) - min(timestamp) > (? * 1000)
-                ;
+                SELECT COUNT(*) FROM (
+                    SELECT
+                        1
+                    FROM
+                        user_daily_visits
+                    WHERE
+                        timestamp > (? * 1000)
+                        AND
+                        timestamp < (? * 1000)
+                    GROUP BY
+                        user_id
+                    HAVING
+                        max(timestamp) - min(timestamp) > (? * 1000)
+                )
             """
 
             txn.execute(
--- a/synapse/storage/databases/main/metrics.py
+++ b/synapse/storage/databases/main/metrics.py
@@ -373,14 +373,14 @@ class ServerMetricsStore(EventPushActionsWorkerStore, SQLBaseStore):
                         FROM
                             user_daily_visits
                         WHERE
-                            timestamp > (? * 1000)
+                            timestamp > (CAST(? AS BIGINT) * 1000)
                             AND
-                            timestamp < (? * 1000)
+                            timestamp < (CAST(? AS BIGINT) * 1000)
                         GROUP BY
                             user_id,
                             client_type
                         HAVING
-                            max(timestamp) - min(timestamp) > (? * 1000)
+                            max(timestamp) - min(timestamp) > (CAST(? AS BIGINT) * 1000)
                     ) AS temp
                 GROUP BY
                     client_type
@@ -410,13 +410,13 @@ class ServerMetricsStore(EventPushActionsWorkerStore, SQLBaseStore):
                     FROM
                         user_daily_visits
                     WHERE
-                        timestamp > (? * 1000)
+                        timestamp > (CAST(? AS BIGINT) * 1000)
                         AND
-                        timestamp < (? * 1000)
+                        timestamp < (CAST(? AS BIGINT) * 1000)
                     GROUP BY
                         user_id
                     HAVING
-                        max(timestamp) - min(timestamp) > (? * 1000)
+                        max(timestamp) - min(timestamp) > (CAST(? AS BIGINT) * 1000)
                 )
             """
 
--- a/synapse/storage/databases/main/metrics.py
+++ b/synapse/storage/databases/main/metrics.py
@@ -389,9 +389,9 @@ class ServerMetricsStore(EventPushActionsWorkerStore, SQLBaseStore):
                 ;
             """
 
-            # REVIEW: do we want to initialise the counts to zero so that we
-            # get an explicit zero for clients that don't appear in our tables?
-            results = {}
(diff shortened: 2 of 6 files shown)


tests/app/test_phone_stats_home.py
```diff
@@ -151,3 +152,234 @@ def test_r30_user_must_be_retained_for_at_least_a_month(self):
...
+
+
+class PhoneHomeR30V2TestCase(HomeserverTestCase):
+    servlets = [
+        synapse.rest.admin.register_servlets_for_client_rest_resource,
+        room.register_servlets,
+        login.register_servlets,
+    ]
+
+    def _advance_to(self, desired_time: float):
```
richvdh, line 164: suggest calling this `desired_time_secs` or something, since it's super unusual for us to use seconds (rather than ms) in Synapse code.

tests/app/test_phone_stats_home.py
```diff
@@ -151,3 +152,234 @@ def test_r30_user_must_be_retained_for_at_least_a_month(self):
...
+
+        # Register a user, log it in, create a room and send a message
+        user_id = self.register_user("u1", "secret!")
+        access_token = self.login("u1", "secret!")
+        room_id = self.helper.create_room_as(room_creator=user_id, tok=access_token)
+        self.helper.send(room_id, "message", tok=access_token)
+        first_post_at = self.hs.get_clock().time()
+
+        # (give time for tables to be updated)
+        self.reactor.advance(300)
```
richvdh, line 197: why 300? maybe add a comment?

tests/app/test_phone_stats_home.py
```diff
@@ -151,3 +152,234 @@ def test_r30_user_must_be_retained_for_at_least_a_month(self):
...
+
+        # Check the R30 results do not count that user.
+        r30_results = self.get_success(self.hs.get_datastore().count_r30v2_users())
```
richvdh, line 200: can you factor out `self.hs.get_datastore()` to a local variable? (`store` is a common name)

tests/app/test_phone_stats_home.py
```diff
@@ -151,3 +152,234 @@ def test_r30_user_must_be_retained_for_at_least_a_month(self):
...
+
+        # *Now* the user is counted.
+        r30_results = self.get_success(self.hs.get_datastore().count_r30v2_users())
+        self.assertEqual(
+            r30_results, {"all": 1, "android": 0, "electron": 0, "ios": 0, "web": 0}
+        )
+
+        # Advance to JUST under 60 days after the user's first post
+        self._advance_to(first_post_at + 60 * ONE_DAY_IN_SECONDS - 5)
+        self.reactor.advance(1)
```
richvdh, line 233: why this?

tests/app/test_phone_stats_home.py
```diff
@@ -151,3 +152,234 @@ def test_r30_user_must_be_retained_for_at_least_a_month(self):
...
+
+        # Register a user and send a message
+        user_id = self.register_user("u1", "secret!")
+        access_token = self.login("u1", "secret!", custom_headers=headers)
+        room_id = self.helper.create_room_as(
+            room_creator=user_id, tok=access_token, custom_headers=headers
+        )
+        self.helper.send(room_id, "message", tok=access_token, custom_headers=headers)
+
+        self.reactor.advance(400)
```
richvdh, line 272: why 400?

tests/app/test_phone_stats_home.py
```diff
@@ -151,3 +152,234 @@ def test_r30_user_must_be_retained_for_at_least_a_month(self):
...
+
+        # Check the user does not contribute to R30 yet.
+        r30_results = self.get_success(self.hs.get_datastore().count_r30v2_users())
```
richvdh, line 275: again, factor out `self.hs.get_datastore()` please

richvdh (changes requested):
generally great - a few comments below!

synapse/storage/databases/main/metrics.py
```diff
@@ -316,6 +316,126 @@ def _count_r30_users(txn):
...
+            sql = """
+                SELECT COUNT(*) FROM (
+                    SELECT
+                        1
+                    FROM
+                        user_daily_visits
+                    WHERE
+                        timestamp > (CAST(? AS BIGINT) * 1000)
+                        AND
+                        timestamp < (CAST(? AS BIGINT) * 1000)
```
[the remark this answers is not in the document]
reivilibre: If you don't do them, Postgres tells you off for out of range integers. Yes, good point about doing the *1000 in Python

synapse/storage/databases/main/metrics.py
```diff
@@ -316,6 +316,126 @@ def _count_r30_users(txn):
...
+                                        WHEN
+                                            LOWER(user_agent) LIKE '%%ios%%'
+                                            THEN 'ios'
+                                        ELSE 'unknown'
+                                    END
+                                WHEN
+                                    LOWER(user_agent) LIKE '%%mozilla%%' OR
+                                    LOWER(user_agent) LIKE '%%gecko%%'
+                                    THEN 'web'
+                                ELSE 'unknown'
```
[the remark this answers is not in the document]
reivilibre: I'm hesitant to change the pre-existing R30 mapping, but as far as I know, we only want to count Element clients for R30v2.

synapse/storage/databases/main/metrics.py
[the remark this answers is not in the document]
reivilibre: actually, think the error only happens because of the multiplication, so shifting that into Python might make the casts redundant :)

tests/app/test_phone_stats_home.py
reivilibre: For context, this was needed earlier before #10324 (strict inequality in some of the queries), but now it's not, so I got rid of it.

synapse/app/phone_stats_home.py
```diff
@@ -107,6 +107,10 @@ async def phone_stats_home(hs, stats, stats_process=_stats_process):
     for name, count in r30_results.items():
         stats["r30_users_" + name] = count
 
+    r30v2_results = await hs.get_datastore().count_r30_users()
```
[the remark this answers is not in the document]
reivilibre: have a follow-up PR ready to go after this one

synapse/storage/databases/main/metrics.py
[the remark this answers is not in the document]
richvdh: :)

tests/app/test_phone_stats_home.py
reivilibre: is the comment now OK?

tests/app/test_phone_stats_home.py
```diff
@@ -151,3 +153,241 @@ def test_r30_user_must_be_retained_for_at_least_a_month(self):
...
+        # the user returns for one day, perhaps just to check out a new feature
+        self.helper.send(room_id, "message", tok=access_token, custom_headers=headers)
+
+        # Give time for user_daily_visits table to be updated.
+        # (user_daily_visits is updated every 5 minutes using a looping call.)
+        self.reactor.advance(FIVE_MINUTES_IN_SECONDS)
+
+        # Check that the user does not contribute to R30v2, even though it's been
+        # more than 30 days since registration.
+        r30_results = self.get_success(self.hs.get_datastore().count_r30v2_users())
```
richvdh, line 352: can you do these `hs.get_datastore`s too, please?

PR #10336, opened 2021-07-08
Ensure we always drop the federation inbound lock

erikjohnston:
Hopefully fixes errors like: `Lock for (federation_inbound_pdu, <room_id>) dropped without being released`

(Sentry link: https://sentry.matrix.org/sentry/synapse-matrixorg/issues/224832/)

### DIFF
--- a/synapse/federation/federation_server.py
+++ b/synapse/federation/federation_server.py
@@ -949,10 +949,8 @@ class FederationServer(FederationBase):
                 room_id, room_version
             )
             if not next:
-                # We need to ensure that we always use the lock, so that it gets
-                # correctly dropped.
-                async with lock:
-                    return
+                await lock.release()
+                return
 
             origin, event = next
         else:
--- a/synapse/federation/federation_server.py
+++ b/synapse/federation/federation_server.py
@@ -949,7 +949,10 @@ class FederationServer(FederationBase):
                 room_id, room_version
             )
             if not next:
-                return
+                # We need to ensure that we always use the lock, so that it gets
+                # correctly dropped.
+                async with lock:
+                    return
 
             origin, event = next
         else:
--- a/changelog.d/10336.bugfix
+++ b/changelog.d/10336.bugfix
@@ -1,0 +1,1 @@
+Fix bug where inbound federation in a room could be delayed due to not correctly dropping a lock. Introduced in v1.37.1.
--- a/synapse/storage/databases/main/lock.py
+++ b/synapse/storage/databases/main/lock.py
@@ -310,13 +310,24 @@ class Lock:
         _excinst: Optional[BaseException],
         _exctb: Optional[TracebackType],
     ) -> bool:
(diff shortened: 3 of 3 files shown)


synapse/federation/federation_server.py
```diff
@@ -949,7 +949,10 @@ async def _process_incoming_pdus_in_room_inner(
                 room_id, room_version
             )
             if not next:
-                return
+                # We need to ensure that we always use the lock, so that it gets
+                # correctly dropped.
+                async with lock:
+                    return
```
richvdh, line 955: I can't help wondering if `Lock` should really be implementing the context manager interface. What's the advantage over just mandating that people call `lock.release()` ?

erikjohnston: Personally I find using a context manager interface clearer than using a `try/finally`. We could also have an explicit `lock.release()` to use here

callahad:
Let's get this in on 1.38 and do an RC2

synapse/federation/federation_server.py
richvdh: > Personally I find using a context manager interface clearer than using a try/finally

well maybe, but that seems a bit irrelevant.

basically:

```
lock.release()
```

seems a more intuitive way of dropping a lock than 

```
async with lock():
    pass
```

I think it's basically a bit silly to have an `aenter` which does nothing.

erikjohnston: I've added a `lock.release()`  method



> > Personally I find using a context manager interface clearer than using a try/finally
> 
> well maybe, but that seems a bit irrelevant.

How so? The other place we use the `async with lock` we'd need to convert to a `try/finally` if we got rid of the context manager.

richvdh (approved):
lgtm

synapse/federation/federation_server.py
richvdh: > How so? The other place we use the `async with lock` we'd need to convert to a `try/finally` if we got rid of the context manager.

Oh I see. Well, yes, we would, and tbh I think that would be an improvement. Currently it looks like something magical happens at the start of that `async with` block, which is misleading. The `with` mechanism doesn't really work very well here, because in each case we already hold the resource (ie, the lock) when we enter the `with` block.

Instead we could put the whole body of `_process_incoming_pdus_in_room_inner` inside a `try` block:

```python
async def _process_incoming_pdus_in_room_inner(...):
    try:
        # ...
        while True:
            await self.handler.on_receive_pdu(event)

            # drop the lock before checking for more events
            await lock.release()
            lock = None

            next = await self.store.get_next_staged_event_for_room()
            if not next:
                break
            
            lock = await self.store.try_acquire_lock()
            if not lock:
                break
    finally:
        if lock:
            lock.release()
```

... thus ensuring that we always drop the lock even if we end up bailing out at odd points due to exceptions or whatever.


Anyway, what you have now is fine.

PR #10353, opened 2021-07-09
Update the logcontext doc

babolivier:
By referring to awaitables instead of deferreds.

### DIFF
--- a/docs/log_contexts.md
+++ b/docs/log_contexts.md
@@ -14,12 +14,16 @@ The `synapse.logging.context` module provides a facilities for managing
 the current log context (as well as providing the `LoggingContextFilter`
 class).
 
-Deferreds make the whole thing complicated, so this document describes
+Awaitables make the whole thing complicated, so this document describes
 how it all works, and how to write code which follows the rules.
 
-## Logcontexts without Deferreds
+In this document, "awaitable" refers to both Python's
+[native awaitables types](https://docs.python.org/3/library/asyncio-task.html#awaitables)
+and Twisted's [Deferreds](https://twistedmatrix.com/documents/current/core/howto/defer.html).
 
-In the absence of any Deferred voodoo, things are simple enough. As with
+## Logcontexts without awaitables
+
+In the absence of any awaitable voodoo, things are simple enough. As with
 any code of this nature, the rule is that our function should leave
 things as it found them:
 
@@ -55,31 +59,28 @@ def do_request_handling():
     logger.debug("phew")
 ```
 
-## Using logcontexts with Deferreds
+## Using logcontexts with awaitables
 
-Deferreds --- and in particular, `defer.inlineCallbacks` --- break the
-linear flow of code so that there is no longer a single entry point
-where we should set the logcontext and a single exit point where we
-should remove it.
+Awaitables break the linear flow of code so that there is no longer a single entry point
+where we should set the logcontext and a single exit point where we should remove it.
 
 Consider the example above, where `do_request_handling` needs to do some
-blocking operation, and returns a deferred:
+blocking operation, and returns an awaitable:
 
 ```python
-@defer.inlineCallbacks
-def handle_request(request_id):
+async def handle_request(request_id):
     with context.LoggingContext() as request_context:
         request_context.request = request_id
-        yield do_request_handling()
+        await do_request_handling()
         logger.debug("finished")
 ```
 
 In the above flow:
 
 -   The logcontext is set
--   `do_request_handling` is called, and returns a deferred
+-   `do_request_handling` is called, and returns an awaitable
 -   `handle_request` yields the deferred
--   The `inlineCallbacks` wrapper of `handle_request` returns a deferred
+-   The `async` syntax on `handle_request` returns an awaitable
 
 So we have stopped processing the request (and will probably go on to
 start processing the next), without clearing the logcontext.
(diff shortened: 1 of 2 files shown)


babolivier:
I didn't change the last section (A note on garbage-collection of Deferred chains) because I don't know if it applies to any awaitable or just to deferreds.

docs/log_contexts.md
```diff
@@ -385,29 +372,27 @@ def handle_request(request_id):
...
---- and provided `do_some_stuff` follows the rules of returning a
-Deferred which runs its callbacks in the original logcontext, all is
-happy.
+--- and provided `do_some_stuff` follows the rules of returning an awaitable
+which runs its callbacks in the original logcontext, all is happy.
 
-The business of a Deferred which runs its callbacks in the original
+The business of an awaitable which runs its callbacks in the original
 logcontext isn't hard to achieve --- we have it today, in the shape of
 `context._PreservingContextDeferred`:
```
richvdh, line 397: _PreservingContextDeferred was removed back in #2675.

I kinda think this whole section is just a source of confusion rather than providing any useful data now. I don't think the approach it's trying to outline carries over particularly well to `async`/`await`, so I don't think we'd take this approach today. WDYT about just removing it?

docs/log_contexts.md
```diff
@@ -419,16 +404,15 @@ def do_some_stuff():
...
-@defer.inlineCallbacks
-def handle_request():
-    yield do_some_stuff()
-    yield do_some_more_stuff()
+async def handle_request():
+    await do_some_stuff()
+    await do_some_more_stuff()
 ```
 
 To conclude: I think this scheme would have worked equally well, with
```
richvdh, line 418: can you `s/yield/await` and `s/inlineCallbacks/async` in the "garbage-collection" section below? afaik it otherwise all still applies.

richvdh (changes requested):
looks like a great improvement, thanks!

docs/log_contexts.md
babolivier: > _PreservingContextDeferred was removed back in #2675.

Oh, good to know

> WDYT about just removing it?

Yeah agreed, let's do it

docs/log_contexts.md
babolivier: Are you sure? It says it's because "`await_something_interesting` is a generator function" which it doesn't seem to be? Or is an awaitable just a fancy generator?

richvdh: A coroutine is indeed, more-or-less, a fancy generator. Indeed, until Python 3.5, that's exactly what it was. The `async` syntax in Python 3.5 means that they aren't quite the same thing, but the mechanism is still basically the same.

After all, they amount to the same thing: a function gets halfway through, and is then suspended.

docs/log_contexts.md
```diff
@@ -55,156 +59,127 @@ def do_request_handling():
...
 
-The easy way to do it is with a combination of `defer.inlineCallbacks`,
-and `context.PreserveLoggingContext`. Suppose we want to implement
+The easy way to do it is by using `context.make_deferred_yieldable`. Suppose we want to implement
 `sleep`, which returns a deferred which will run its callbacks after a
 given number of seconds. That might look like:
 
 ```python
 # not a logcontext-rules-compliant function
+
```
richvdh, line 170:
```suggestion
```

docs/log_contexts.md
```diff
@@ -294,152 +265,54 @@ def do_request_handling():
...
-more deferred via `defer.gatherResults`:
+more awaitables via `defer.gatherResults`:
 
 ```python
-d1 = operation1()
-d2 = operation2()
-d3 = defer.gatherResults([d1, d2])
+a1 = operation1()
+a2 = operation2()
+a3 = defer.gatherResults(a1, a2)
```
richvdh, line 273:
```suggestion
a3 = defer.gatherResults([a1, a2])
```

docs/log_contexts.md
```diff
@@ -294,152 +265,54 @@ def do_request_handling():
...
 ```python
-@defer.inlineCallbacks
-def do_request_handling():
-    d1 = context.preserve_fn(operation1)()
-    d2 = context.preserve_fn(operation2)()
+async def do_request_handling():
+    a1 = context.run_in_background(operation1)
+    a2 = context.run_in_background(operation2)
 
     with PreserveLoggingContext():
```
richvdh, line 309: normally we'd use `make_deferred_yieldable` rather than `PreserveLoggingContext` here.

docs/log_contexts.md
```diff
@@ -476,8 +348,8 @@ def reset_listener_queue():
     listener_queue.clear()
 ```
 
-So, both ends of the deferred chain have now dropped their references,
-and the deferred chain is now orphaned, and will be garbage-collected at
+So, both ends of the awaitable chain have now dropped their references,
+and the awaitable chain is now orphaned, and will be garbage-collected at
 some point. Note that `await_something_interesting` is a generator
 function, and when Python garbage-collects generator functions, it gives
 them a chance to clean up by making the `yield` raise a `GeneratorExit`
```
richvdh, line 355:
```suggestion
some point. Note that `await_something_interesting` is a coroutine, 
which Python implements as a generator function.  When Python
garbage-collects generator functions, it gives them a chance to 
clean up by making the `async` (or `yield`) raise a `GeneratorExit`
```

docs/log_contexts.md
```diff
@@ -486,8 +358,8 @@ exception. In our case, that means that the `__exit__` handler of
 there is now nothing waiting for its return, so the request context is
 never cleared.
 
-To reiterate, this problem only arises when *both* ends of a deferred
-chain are dropped. Dropping the the reference to a deferred you're
+To reiterate, this problem only arises when *both* ends of a awaitable
+chain are dropped. Dropping the the reference to an awaitable you're
 supposed to be calling is probably bad practice, so this doesn't
```
richvdh, line 363:
```suggestion
supposed to be awaiting is bad practice, so this doesn't
```

richvdh (approved):
a few more nits. lgtm otherwise!

docs/log_contexts.md
```diff
@@ -294,152 +265,54 @@ def do_request_handling():
...
-@defer.inlineCallbacks
-def do_request_handling():
+async def do_request_handling():
     with PreserveLoggingContext():
-        d1 = operation1()
-        d2 = operation2()
-        result = yield defer.gatherResults([d1, d2])
+        a1 = operation1()
+        a2 = operation2()
+        result = await defer.gatherResults(a1, a2)
```
babolivier, line 296:
```suggestion
        result = await defer.gatherResults([a1, a2])
```

Issue #10275, opened 2021-06-29
followup on asynchronous-federation work


richvdh:
A few things it might be good to think about having landed #10272:

* [x] we disabled a few tests. We should fix and re-enable them.
   * [x] sytest: `Outbound federation will ignore a missing event with bad JSON for room version 6`
     #10345
   * [x] sytest: `Backfilled events whose prev_events are in a different room do not allow cross-room back-pagination` 
     #10292
   * [x] sytest: `Federation rejects inbound events where the prev_events cannot be found`
     #10357
   * [x] complement: `Outbound federation will ignore a missing event with bad JSON for room version 6`
     https://github.com/matrix-org/complement/pull/152
* [x] There's nothing stopping the queues building up almost indefinitely. We need to figure out what to do if a server can't keep up with the inbound traffic. https://github.com/matrix-org/synapse/pull/10390
* We previously had a rudimentary form of this queue which kicked in while we were joining a room over federation (see `room_queues` in `FederationHandler`. I think we should update that code to use the same logic. (parking this for now - we'll revisit it as part of future join work)
* [x] #10295
* [x] The new queues need some monitoring.
* [x] ~~We now only process one event per room at a time.~~ Not expected to be significant

erikjohnston:
https://github.com/matrix-org/sytest/pull/1059 fixes one of the tests.

clokep:
This also seems to be causing a complement test to fail: `TestOutboundFederationIgnoresMissingEventWithBadJSONForRoomVersion6`

ShadowJonathan:
...now *that* is a java-worthy name

clokep:
matrix-org/complement#142 updates complement to match the sytest blacklist applied in #10272.

erikjohnston:
Some more monitoring has been added at https://github.com/matrix-org/synapse/pull/10284 and https://github.com/matrix-org/synapse/pull/10279

PR #10357, opened 2021-07-09
Unblacklist fixed tests

erikjohnston:
Fixed in https://github.com/matrix-org/sytest/pull/1061 and https://github.com/matrix-org/sytest/pull/1067

c.f. #10275


richvdh (approved):
looks fine, modulo the conflict.

you might want to merge with the newsfile from #10345 :shrug:

Issue #5064, opened 2019-04-15
state_groups_state uses a large sequential scan


grinapo:
### Description
I have caught a query which runs several times for room history, recursively, and therefore take a significant amount of time (tens of minutes) to scroll up.

```SQL
WITH RECURSIVE state(state_group) AS ( 
	VALUES(64193::bigint) 
		UNION ALL 
	SELECT prev_state_group FROM state_group_edges e, state s WHERE s.state_group = e.state_group ) 
	
SELECT DISTINCT type, state_key, last_value(event_id) 
	OVER ( 
		PARTITION BY type, state_key ORDER BY state_group ASC ROWS BETWEEN UNBOUNDED PRECEDING AND UNBOUNDED FOLLOWING 
	) AS event_id FROM state_groups_state WHERE state_group IN ( 
		SELECT state_group FROM state ) AND ((type = 'm.room.history_visibility' AND state_key = '') OR 
			(type = 'm.room.member' AND state_key = '@grin:grin.hu'))
```

And the query can be speeded up by
```SQL
CREATE INDEX state_groups_state_state_key_type ON state_groups_state (state_key, type);
```

Without it the planner say:
```
                                 ->  Parallel Seq Scan on state_groups_state  (cost=0.00..284361.03 rows=82 width=83)                                                                                                                                                                                        
                                       Filter: (((type = 'm.room.history_visibility'::text) AND (state_key = ''::text)) OR ((type = 'm.room.member'::text) AND (state_key = '@grin:grin.hu'::text)))

```

And with it it's
```
                           ->  Bitmap Heap Scan on state_groups_state  (cost=15.20..792.20 rows=198 width=83)
                                 Recheck Cond: (((state_key = ''::text) AND (type = 'm.room.history_visibility'::text)) OR ((state_key = '@grin:grin.hu'::text) AND (type = 'm.room.member'::text)))
                                 ->  BitmapOr  (cost=15.20..15.20 rows=198 width=0)
                                       ->  Bitmap Index Scan on state_groups_state_state_key  (cost=0.00..4.60 rows=4 width=0)
                                             Index Cond: ((state_key = ''::text) AND (type = 'm.room.history_visibility'::text))
                                       ->  Bitmap Index Scan on state_groups_state_state_key  (cost=0.00..10.50 rows=194 width=0)
                                             Index Cond: ((state_key = '@grin:grin.hu'::text) AND (type = 'm.room.member'::text))
```
which is 175 times faster. ;-)


### Version information
[...]

PR #10359, opened 2021-07-09
Reduce likelihood of Postgres table scanning `state_groups_state`.

erikjohnston:
The postgres statistics collector sometimes massively underestimates the
number of distinct state groups are in the `state_groups_state`, which
can cause postgres to use table scans for queries for multiple state
groups.

We fix this by manually setting `n_distinct` on the column.

Note that the value `-0.02` was taken by looking at jki.re's database.

Fixes https://github.com/matrix-org/synapse/issues/5064


synapse/storage/schema/main/delta/61/02state_groups_state_n_distinct.sql.postgres
```diff
@@ -0,0 +1,34 @@
...
+--
+-- To work around this we can manually tell postgres the number of distint state
+-- groups there are by setting `n_distinct` (a negative value here is the number
+-- of distinct values divided by the number of rows, so -0.02 means on average
+-- there are 50 rows per disinct value). We don't need a particularly
+-- accurate number here, as a) we just want it to always use index scans and b)
+-- our estimate is going to be better than the one made by the statistics
+-- collector.
+
+ALTER TABLE state_groups_state ALTER COLUMN state_group SET (n_dinstict = -0.02)
```
Half-Shot, line 29: itym
```suggestion
ALTER TABLE state_groups_state ALTER COLUMN state_group SET (n_distinct = -0.02)
```

erikjohnston: Thanks! I was confused why this didn't error, but it turns out if you forget the semicolon on the end we don't even run the command...

synapse/storage/schema/main/delta/61/02state_groups_state_n_distinct.sql.postgres
```diff
@@ -0,0 +1,34 @@
...
+-- of distinct values divided by the number of rows, so -0.02 means on average
+-- there are 50 rows per disinct value). We don't need a particularly
+-- accurate number here, as a) we just want it to always use index scans and b)
+-- our estimate is going to be better than the one made by the statistics
+-- collector.
+
+ALTER TABLE state_groups_state ALTER COLUMN state_group SET (n_distinct = -0.02);
+
+-- Ideally we'd do an `ANALYZE state_groups_state (state_group)` here so that
+-- the above gets picked up immediately, but that can take a bit of time so we
```
grinapo, line 32: As a sidenote: on my 72GB sized db the ANALYZE above it took about 1500 ms.

MTRNord: For my db ist was around 2.5s but the result of analyze caused it to be noticeable slower than with the ALTER TABLE.

synapse/storage/schema/main/delta/61/02state_groups_state_n_distinct.sql.postgres
```diff
@@ -0,0 +1,34 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+
+-- By default the postgres statistics collector massively underestimates the
+-- number of distinct state groups are in the `state_groups_state`, which can
+-- cause postgres to use table scans for queries for multiple state groups.
+--
+-- To work around this we can manually tell postgres the number of distint state
```
richvdh, line 21:
```suggestion
-- To work around this we can manually tell postgres the number of distinct state
```

synapse/storage/schema/main/delta/61/02state_groups_state_n_distinct.sql.postgres
```diff
@@ -0,0 +1,34 @@
...
+-- groups there are by setting `n_distinct` (a negative value here is the number
+-- of distinct values divided by the number of rows, so -0.02 means on average
+-- there are 50 rows per disinct value). We don't need a particularly
```
richvdh, line 24:
```suggestion
-- there are 50 rows per distinct value). We don't need a particularly
```

richvdh (approved):

PR #10368, opened 2021-07-10
doc: Add delegation example to the caddy reverse proxy section

moritzdietz:
This PR can be seen as a mere suggestion. I can see the documentation getting a bit out of hand if one adds full examples of all of the reverse proxies out there. Dedicated guides on how to set up a specific reverse proxy could be better maybe.
But I also thought that it couldn't hurt to expand the bit of documentation and give admins looking into getting all information from one source the opportunity to not just choose nginx because it's _just_ all there for it already.

Perhaps there is a better place within the repository to document full example of a specific reverse proxy flavor which?

The documentation consists of information I gathered from these sources:

- CORS headers: https://matrix.org/docs/spec/client_server/latest#cors
- `Content-Type "application/json"`: https://matrix.org/docs/spec/server_server/r0.1.4#get-well-known-matrix-server
- Caddy: [concpets/addresses](https://caddyserver.com/docs/caddyfile/concepts#addresses)
- Caddy: [concepts/snippets](https://caddyserver.com/docs/caddyfile/concepts#snippets)
- Caddy: [concepts/tokens-and-quotes](https://caddyserver.com/docs/caddyfile/concepts#tokens-and-quotes)
- Caddy: [directives/handle](https://caddyserver.com/docs/caddyfile/directives/handle)
- Caddy: [directives/header](https://caddyserver.com/docs/caddyfile/directives/header)
- Caddy: [directives/respond](https://caddyserver.com/docs/caddyfile/directives/respond)
- Caddy: [directives/reverse_proxy](https://caddyserver.com/docs/caddyfile/directives/reverse_proxy)

I used a flattened JSON body for the respond body, as Caddy will throw a format error for the Caddyfile configuration file otherwise.

### Pull Request Checklist

Signed-off-by: Moritz Dietz moritz@moritzdietz.com

### DIFF
--- a/docs/reverse_proxy.md
+++ b/docs/reverse_proxy.md
@@ -116,7 +116,7 @@ example.com {
 
     handle /.well-known/matrix/client {
         import matrix-well-known-header
-        respond `{"m.homeserver":{"base_url":"https://matrix.example.com"},"m.identity_server":{"base_url":"https://identity.example.com}}`
+        respond `{"m.homeserver":{"base_url":"https://matrix.example.com"},"m.identity_server":{"base_url":"https://identity.example.com"}}`
     }
 }
 
--- a/docs/reverse_proxy.md
+++ b/docs/reverse_proxy.md
@@ -98,6 +98,37 @@ example.com:8448 {
   reverse_proxy http://localhost:8008
 }
 ```
+[Delegation](delegate.md) example:
+```
+(matrix-well-known-header) {
+    # Headers
+    header Access-Control-Allow-Origin "*"
+    header Access-Control-Allow-Methods "GET, POST, PUT, DELETE, OPTIONS"
+    header Access-Control-Allow-Headers "Origin, X-Requested-With, Content-Type, Accept, Authorization"
+    header Content-Type "application/json"
+}
+
+example.com {
+    handle /.well-known/matrix/server {
+        import matrix-well-known-header
+        respond `{"m.server":"matrix.example.com:8448"}`
+    }
+
+    handle /.well-known/matrix/client {
+        import matrix-well-known-header
+        respond `{"m.homeserver":{"base_url":"https://matrix.example.com"},"m.identity_server":{"base_url":"https://identity.example.com}}`
+    }
+}
+
+matrix.example.com {
+    reverse_proxy /_matrix/* http://localhost:8008
+    reverse_proxy /_synapse/client/* http://localhost:8008
+}
+
+matrix.example.com:8448 {
+    reverse_proxy http://localhost:8008
+}
+```
 
 ### Apache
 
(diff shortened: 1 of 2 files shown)


docs/reverse_proxy.md
```diff
@@ -98,6 +98,37 @@ example.com:8448 {
...
+    header Access-Control-Allow-Origin "*"
+    header Access-Control-Allow-Methods "GET, POST, PUT, DELETE, OPTIONS"
+    header Access-Control-Allow-Headers "Origin, X-Requested-With, Content-Type, Accept, Authorization"
+    header Content-Type "application/json"
+}
+
+example.com {
+    handle /.well-known/matrix/server {
+        import matrix-well-known-header
+        respond `{"m.server":"matrix.example.com:8448"}`
```
tulir, line 114: Why not just delegate to 443 and remove the `matrix.example.com:8448` handler entirely?

moritzdietz: Because I didn't want to deviate from the current docs as they were. Although I myself have this setup as you suggested myself, so I am definitely willing to update the docs on that too.
Not sure if I should add it in this PR too - if that's fine, then I'll gladly make those changes :)

docs/reverse_proxy.md
```diff
@@ -98,6 +98,37 @@ example.com:8448 {
...
+        respond `{"m.homeserver":{"base_url":"https://matrix.example.com"},"m.identity_server":{"base_url":"https://identity.example.com}}`
+    }
+}
+
+matrix.example.com {
+    reverse_proxy /_matrix/* http://localhost:8008
+    reverse_proxy /_synapse/client/* http://localhost:8008
+}
+
+matrix.example.com:8448 {
```
ShadowJonathan, line 128: I'm pretty sure you have to prefix this with `https://`? Can you test this and confirm it works?

Also, making sure that it only sends federation-related paths to this handler, and only client-related paths to the handler above, would help with illustrating the difference between these.

ShadowJonathan (approved):
Looks good, some notes to look into though.

docs/reverse_proxy.md
moritzdietz: > I'm pretty sure you have to prefix this with `https://`? Can you test this and confirm it works?

You don't have to specify `https://` as that's implicit. As per https://caddyserver.com/docs/caddyfile/concepts#addresses which I linked in the description as well (for this very reason actually so my reasoning for choosing this configuration is explained).

moritzdietz: > Also, making sure that it only sends federation-related paths to this handler, and only client-related paths to the handler above, would help with illustrating the difference between these.

Yes, I agree. I'll add it.

moritzdietz:
If I combine the suggestions from the both of you  I can actually trim it down quite a bit.
I'll make those changes and  push them  to this PR (knowing that they work) and we can see if we like it.

docs/reverse_proxy.md
moritzdietz: I've applied your suggestion just now.

ShadowJonathan:
Looking good now, thanks!

babolivier (approved):
Thanks! :)

PR #10379, opened 2021-07-13
Github Actions workflow to attach release artifacts to release

richvdh:


erikjohnston (approved):
Shiny

.github/workflows/release-artifacts.yml
```diff
@@ -62,3 +62,29 @@ jobs:
...
+    needs:
+      - build-debs
+      - build-sdist
+    runs-on: ubuntu-latest
+    steps:
+      - name: Download all workflow run artifacts
+        uses: actions/download-artifact@v2
+      - name: Build a tarball for the debs
+        run: tar -cvJf debs.tar.xz debs
+      - name: Attach to release
```
babolivier, line 79: This will also create the release if it doesn't exist, right?

babolivier (approved):

.github/workflows/release-artifacts.yml
richvdh: yes, it will. per the comments: it will leave it as a "draft", so we can go and fix it up before it gets published.

Issue #8351, opened 2020-09-18
Type-annotate all function arguments and internal variables


ShadowJonathan:
**Description:**

Related to #8347, ~~(partially) blocked till 3.5 is deprecated~~

The goal is to;
- [x] Convert existing `name = ... # type:` comments to `name: type = ...`. (see "`com2ann` progress" below)
- [ ] Put type annotations on every function's parameters and return type.
- [ ] Put `name: type` annotations on;
  - [ ] Every class variable.
  - [ ] Every instance variable (declared in `__init__()` functions), unless;
    - They're constructing another instance (`self.thing = Thing()`) (dubious, see https://github.com/matrix-org/synapse/pull/10382#discussion_r671357795)
    - They're getting assigned a non-container literal (`1`, `b""`, `""`, etc. but not `None`)
    - The assignment comes from a function/method call (`self.thing = hs.get_another_thing()`), given that the function signature is known and inferrable.

---

<details>
<summary><code>com2ann</code> progress</summary>

- [x] `tests/` (#10350)
- [x] `synapse/`
	- [x] #10382
	- [x] #10381
	- [x] #10380
- [x] Other (#10418)

(`com2ann` changes comment type annotations to inline type annotations, `name = ... # type:` to `name: type = ...`, this is possible since python 3.6, and standard across the python ecosystem.)

</details>

clokep:
> This could probably finally allow a `mypy` linter to run with CI, to catch future inconsistencies.

`mypy` runs in CI on a subset of the code-base, see https://github.com/matrix-org/synapse/blob/develop/mypy.ini#L9

clokep:
> blocked till 3.5 is deprecated

This isn't blocked? Python 3.5 supports type annotations.

ShadowJonathan:
> > blocked till 3.5 is deprecated
> 
> This isn't blocked? Python 3.5 supports type annotations.

not variable type annotations, but you're right about it not being completely blocked

ShadowJonathan:
A large portion of this can be fixed with `com2ann`, see https://github.com/matrix-org/synapse/issues/9131#issuecomment-776568681

ShadowJonathan:
Updated description, the previous one was vague and also implied setting type annotations for every variable inside non-init functions, which i think is overdoing it a little bit, type-annotating all other places noted should already correctly resolve those spots.

ShadowJonathan:
I found some interesting `mypy` optional errors that could be enabled to enforce type annotations;

[...]

PR #10381, opened 2021-07-13
Use inline type hints in `http/federation/`, `storage/` and `util/`

ShadowJonathan:
Part of #8351

This PR is tantamount to running

```
python3.8 -m com2ann -v 6 http/federation/
python3.8 -m com2ann -v 6 storage/
python3.8 -m com2ann -v 6 util/
```

(com2ann requires python 3.8 to run)

### Pull Request Checklist

---

### DIFF
--- a/synapse/http/federation/well_known_resolver.py
+++ b/synapse/http/federation/well_known_resolver.py
@@ -70,10 +70,8 @@ WELL_KNOWN_RETRY_ATTEMPTS = 3
 logger = logging.getLogger(__name__)
 
 
-_well_known_cache = TTLCache("well-known")  # type: TTLCache[bytes, Optional[bytes]]
-_had_valid_well_known_cache = TTLCache(
-    "had-valid-well-known"
-)  # type: TTLCache[bytes, bool]
+_well_known_cache: TTLCache[bytes, Optional[bytes]] = TTLCache("well-known")
+_had_valid_well_known_cache: TTLCache[bytes, bool] = TTLCache("had-valid-well-known")
 
 
 @attr.s(slots=True, frozen=True)
@@ -130,9 +128,10 @@ class WellKnownResolver:
         # requests for the same server in parallel?
         try:
             with Measure(self._clock, "get_well_known"):
-                result, cache_period = await self._fetch_well_known(
-                    server_name
-                )  # type: Optional[bytes], float
+                result: Optional[bytes]
+                cache_period: float
+
+                result, cache_period = await self._fetch_well_known(server_name)
 
         except _FetchWellKnownFailure as e:
             if prev_result and e.temporary:
--- a/synapse/util/caches/descriptors.py
+++ b/synapse/util/caches/descriptors.py
@@ -46,17 +46,17 @@ F = TypeVar("F", bound=Callable[..., Any])
 
 
 class _CachedFunction(Generic[F]):
-    invalidate = None  # type: Any
-    invalidate_all = None  # type: Any
-    prefill = None  # type: Any
-    cache = None  # type: Any
-    num_args = None  # type: Any
+    invalidate: Any = None
+    invalidate_all: Any = None
+    prefill: Any = None
+    cache: Any = None
+    num_args: Any = None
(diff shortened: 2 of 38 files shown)


synapse/util/caches/descriptors.py
```diff
@@ -46,17 +46,17 @@
...
-    cache = None  # type: Any
-    num_args = None  # type: Any
+    invalidate: Any = None
+    invalidate_all: Any = None
+    prefill: Any = None
+    cache: Any = None
+    num_args: Any = None
 
-    __name__ = None  # type: str
+    __name__: str
```
clokep, line 55: Looks like the defaults of `None` are removed in a bunch of places in this file. I'm not sure if that's really correct or not?

I guess it is OK since this is a generic, not an implementation, but I find it a bit odd.

synapse/http/federation/well_known_resolver.py
```diff
@@ -130,9 +128,10 @@ async def get_well_known(self, server_name: bytes) -> WellKnownLookupResult:
         # requests for the same server in parallel?
         try:
             with Measure(self._clock, "get_well_known"):
-                result, cache_period = await self._fetch_well_known(
-                    server_name
-                )  # type: Optional[bytes], float
+                result: Optional[bytes]
+                cache_period: float
+
+                result, cache_period = await self._fetch_well_known(server_name)
```
clokep, line 134: I find this fairly unreadable and would prefer the old way, but 🤷

ShadowJonathan: Some SO answers say this is the way to do it, i think this is the most pythonic you're gonna get, an alternative would be
```py
res: Tuple[Optional[bytes], float] = await self._fetch_well_known(server_name)

result: Optional[bytes] = res[0]
cache_period: float = res[1]
```

synapse/util/caches/descriptors.py
ShadowJonathan: I believe it's more correct than before, though i don't know for sure, i'm mainly afraid of the `None`s leaking through to the subclasses, becoming defaults on the class/instance variables there

reivilibre: This transformation seems a bit weird in general, but it *is* a generic and I guess `None` isn't of type `str` either, so maybe it's no more nonsensical than before.

I would be careful about similar transformations elsewhere though, since it hasn't preserved the exact same semantics:

```python
>>> xyz: str
>>> xyz
NameError: name 'xyz' is not defined
```

synapse/http/federation/well_known_resolver.py
reivilibre: In my opinion, neither before or after is particularly pretty, but I guess there's something to be said about adhering to the 'Pythonic'/standard way of doing things, and if this is it, then why not?

synapse/util/caches/descriptors.py
clokep: Looks like we already do this in some other places, so you can ignore this. 😄

synapse/http/federation/well_known_resolver.py
clokep: Yeah I can't say the alternatives are very pretty either. I don't think we'd even need them except we have to mark `result` as optional. 😢

ShadowJonathan: I think that explicitly annotating unpacked variables wont be needed if we have enough function signatures in place, after which we can just drop the inline type arguments (unless its needed for some explicit reason) and let mypy infer from the methods.

So i think this is okay coupled with the reason to please mypy for now, with the intention to remove it later, does that sound good?

clokep: > I think that explicitly annotating unpacked variables wont be needed if we have enough function signatures in place

`_fetch_well_known` *does* have proper types, it just that it is `Tuple[bytes, float]`, but the caller might have the `result` value be `Optional[bytes]` so it gets a bit awkward. 😄 

Anyway, I think this is fine.

PR #10386, opened 2021-07-13
Port the ThirdPartyEventRules module interface to the new generic interface

babolivier:
Port the third-party event rules interface to the generic module interface introduced in v1.37.0


babolivier:
Third-party rules tests are failing because I haven't ported them yet, the code in the PR should be fine to run/review.

synapse/events/third_party_rules.py
```diff
@@ -46,6 +125,43 @@ def __init__(self, hs: "HomeServer"):
...
+    def register_third_party_rules_callbacks(
+        self,
+        check_event_allowed: Optional[CHECK_EVENT_ALLOWED_CALLBACK] = None,
+        on_create_room: Optional[ON_CREATE_ROOM_CALLBACK] = None,
+        check_threepid_can_be_invited: Optional[
+            CHECK_THREEPID_CAN_BE_INVITED_CALLBACK
+        ] = None,
+        check_visibility_can_be_modified: Optional[
+            CHECK_VISIBILITY_CAN_BE_MODIFIED_CALLBACK
+        ] = None,
```
anoadragon453, line 166: This list might grow very large at some point, and I'm aware that there's a few different places one needs to edit to add a new method to the interface.

I think this could all be replaced by a `Dict[str, List[Callable]]` that looks like:

```jsonc
{
    "check_event_allowed": [callback1, callback2]
    "on_create_room": [callback3, callback4],
    // ...
}
```

Not that this needs to be done now, but it might make things simpler for the future. I'm also not sure whether this setup would preserve the same type-checking we currently have though.

It would also allow for a simpler implementation of this method:

```python
def register_third_party_rules_callbacks(callback_names_to_functions):
    for name, f in callback_names_to_functions:
        if name not in self._callbacks:
            raise Exception("Unknown callback")

        self._callbacks[name].append(f)
```

synapse/events/third_party_rules.py
```diff
@@ -63,7 +179,8 @@ async def check_event_allowed(
         Returns:
             The result from the ThirdPartyRules module, as above
         """
-        if self.third_party_rules is None:
+        # Bail out early without hitting the store if we don't have any callback
```
anoadragon453, line 204:
```suggestion
        # Bail out early without hitting the store if we don't have any callbacks to run.
```

synapse/events/third_party_rules.py
```diff
@@ -77,29 +194,31 @@ async def check_event_allowed(
...
         """
-
-        if self.third_party_rules is None:
-            return True
-
-        return await self.third_party_rules.on_create_room(
-            requester, config, is_requester_admin
-        )
+        for callback in self._on_create_room_callbacks:
+            await callback(requester, config, is_requester_admin)
```
anoadragon453, line 242: Should we add some error handling so that one broken module doesn't prevent others from executing this method?

synapse/events/third_party_rules.py
```diff
@@ -114,15 +233,17 @@ async def check_threepid_can_be_invited(
         Returns:
             True if the 3PID can be invited, False if not.
         """
-
-        if self.third_party_rules is None:
+        # Bail out early without hitting the store if we don't have any callback
```
anoadragon453, line 257:
```suggestion
        # Bail out early without hitting the store if we don't have any callbacks to run.
```

synapse/events/third_party_rules.py
```diff
@@ -137,18 +258,17 @@ async def check_visibility_can_be_modified(
         Returns:
```
anoadragon453, line 279: Could you mention in the docstring that `False` will be returned as long as one `ThirdPartyRulesModule` denies the request?

synapse/events/third_party_rules.py
```diff
@@ -164,20 +183,22 @@ def register_third_party_rules_callbacks(
...
+    ) -> Tuple[bool, dict]:
         """Check if a provided event should be allowed in the given context.
 
         The module can return:
             * True: the event is allowed.
             * False: the event is not allowed, and should be rejected with M_FORBIDDEN.
-            * a dict: replacement event data.
+
+        If the event is allowed, the module can also return a dictionary to use as a
+        replacement for the event.
```
anoadragon453, line 195:
```suggestion
        replacement for the event's content.
```
?

synapse/handlers/federation.py
```diff
@@ -1931,7 +1931,7 @@ async def on_make_knock_request(
             builder=builder
         )
 
-        event_allowed = await self.third_party_event_rules.check_event_allowed(
+        event_allowed, _ = await self.third_party_event_rules.check_event_allowed(
```
anoadragon453, line 1934: Do we not allow modification of events received over federation?

docs/modules.md
```diff
@@ -181,13 +181,68 @@ The arguments passed to this callback are:
...
+    file_info: "synapse.rest.media.v1._base.FileInfo",
 ) -> bool
 ```
 
 Called when storing a local or remote file. The module must return a boolean indicating
 whether the given file can be stored in the homeserver's media store.
 
+#### Third party rules callbacks
+
+Third party rules callbacks allow modules developers to add extra checks to verify the
```
anoadragon453, line 193:
```suggestion
Third party rules callbacks allow module developers to add extra checks to verify the
```

docs/modules.md
```diff
@@ -181,13 +181,68 @@ The arguments passed to this callback are:
...
+The available third party rules callbacks are:
+
+```python
+async def check_event_allowed(
+    event: "synapse.events.EventBase",
+    state_events: "synapse.types.StateMap",
+) -> Tuple[bool, Optional[dict]]
+```
+
+Called when processing any incoming event, with the event and the a `StateMap`
```
anoadragon453, line 206:
```suggestion
Called when processing any incoming event, with the event and the `StateMap`
```

docs/modules.md
```diff
@@ -181,13 +181,68 @@ The arguments passed to this callback are:
...
+representing the current state of the room the event is being sent into. A `StateMap` is
+a dictionary indexed on tuples containing an event type and a state key; for example
```
anoadragon453, line 208: "indexed on" sounds odd... what about "a dictionary that maps from a two-tuple containing event type and state key strings to an arbitrary value"?

docs/modules.md
```diff
@@ -181,13 +181,68 @@ The arguments passed to this callback are:
...
+retrieving the room's `m.room.create` event from the `state_events` argument looks like
+this: `state_events.get(("m.room.create", ""))`. The module must return a boolean
+indicating whether the event can be allowed.
+
+Note that this callback function processes incoming events coming via federation
+traffic (on top of client traffic). This means denying an event might cause the local
+copy of the room's history to diverge from the ones of remote servers. This may cause
```
anoadragon453, line 215:
```suggestion
copy of the room's history to diverge from that of remote servers. This may cause
```

docs/modules.md
```diff
@@ -181,13 +181,68 @@ The arguments passed to this callback are:
...
+federation issues in the room. It is strongly recommended to only deny events using this
+callback function if the sender is a local user, or in a private federation in which all
+servers are using the same module, with the same configuration.
```
anoadragon453, line 218: Hmm, why do we even have this functionality then?

Additionally, it might be worth calling out that the spam checker callback may be better suited for handling unwanted events over federation.

docs/modules.md
```diff
@@ -181,13 +181,68 @@ The arguments passed to this callback are:
...
+
+If the boolean returned by the module is `True`, it may also tell Synapse to replace the
+event with new data by returning the new event's data as a dictionary. In order to do
```
anoadragon453, line 221: Again, is this only the event's content?

docs/upgrade.md
```diff
@@ -86,6 +86,19 @@ process, for example:
...
+
+## Deprecation of the current third-party rules module interface
+
+The current third-party rules module interface is deprecated in favour of the new generic
+modules system introduced in Synapse v1.37.0. Authors of third-party rules modules can refer
+to [this documentation](modules.md#porting-an-existing-module-that-uses-the-old-interface)
+to update their modules. Synapse administrators can refer to [this documentation](modules.md#using-modules)
+to update their configuration once the modules they are using have been updated.
+
+We plan to remove support for the current spam checker interface in September 2021.
```
anoadragon453, line 99:
```suggestion
We plan to remove support for the current third-party rules interface in September 2021.
```

anoadragon453 (changes requested):
Looking good overall, happy to see more modules being migrated!

Bunch of tiny things below.

synapse/events/third_party_rules.py
babolivier: This reproducing the way spam checker callbacks are registered. Since we're trying to standardise the module interface, I'd like it if we can avoid having different ways to register callbacks depending on the module type. There's more context in https://github.com/matrix-org/synapse/pull/10062 on why we're doing it like that, but the gist of it is that it's the only way to have good enough typing of the callbacks without going for something over-engineered.

synapse/handlers/federation.py
babolivier: No, to avoid getting splitbrained on what a given event looks like, there's some words on that in the doc.

docs/modules.md
babolivier: > Hmm, why do we even have this functionality then?

Because it's useful to have in private federations, which is what the doc says.

> Additionally, it might be worth calling out that the spam checker callback may be better suited for handling unwanted events over federation.

Yeah good idea. There are some other issues with this method which would justify even more recommending using the spam checker callback instead.

synapse/events/third_party_rules.py
babolivier: No, it's the event, not just its content.

docs/modules.md
babolivier: Nope.

synapse/events/third_party_rules.py
babolivier: This is more following the rule of thumb of running several modules, so the right fix would be adding more documentation in `modules.md`, I'll do that in a separate PR.

babolivier:
Flagging as release blockers since we have commitments to have this released by the end of the month

synapse/events/third_party_rules.py
```diff
@@ -217,7 +215,12 @@ async def check_event_allowed(
         event.freeze()
 
         for callback in self._check_event_allowed_callbacks:
-            res, replacement_data = await callback(event, state_events)
+            try:
+                res, replacement_data = await callback(event, state_events)
+            except Exception as e:
+                logger.warning("Failed to run callback %s: %s", callback, e)
```
anoadragon453, line 221:
```suggestion
                logger.warning("Failed to run module API callback %s: %s", callback, e)
```

Just because "Failed to run callback" is pretty generic.

synapse/events/third_party_rules.py
```diff
@@ -254,15 +268,18 @@ async def check_threepid_can_be_invited(
...
         state_events = await self._get_state_map_for_room(room_id)
 
         for callback in self._check_threepid_can_be_invited_callbacks:
-            if await callback(medium, address, state_events) is False:
-                return False
+            try:
+                if await callback(medium, address, state_events) is False:
+                    return False
+            except Exception as e:
+                logger.warning("Failed to run callback %s: %s", callback, e)
```
anoadragon453, line 282: Likewise.

synapse/events/third_party_rules.py
```diff
@@ -286,8 +303,11 @@ async def check_visibility_can_be_modified(
         state_events = await self._get_state_map_for_room(room_id)
 
         for callback in self._check_visibility_can_be_modified_callbacks:
-            if await callback(room_id, state_events, new_visibility) is False:
-                return False
+            try:
+                if await callback(room_id, state_events, new_visibility) is False:
+                    return False
+            except Exception as e:
+                logger.warning("Failed to run callback %s: %s", callback, e)
```
anoadragon453, line 310: Likewise.

anoadragon453 (approved):
One very minor nit, but otherwise looks good to me!

synapse/events/third_party_rules.py
babolivier:
```suggestion
                logger.warning("Failed to run module API callback %s: %s", callback, e)
```

synapse/events/third_party_rules.py
babolivier:
```suggestion
                logger.warning("Failed to run module API callback %s: %s", callback, e)
```

PR #10390, opened 2021-07-14
Prune inbound federation queues if they get too long

erikjohnston:
c.f. #10275

### DIFF
--- a/synapse/federation/federation_server.py
+++ b/synapse/federation/federation_server.py
@@ -998,7 +998,7 @@ class FederationServer(FederationBase):
 
             origin, event = next
 
-            # Prune the event queue if its getting large.
+            # Prune the event queue if it's getting large.
             #
             # We do this *after* handling the first event as the common case is
             # that the queue is empty (/has the single event in), and so there's
--- a/synapse/federation/federation_server.py
+++ b/synapse/federation/federation_server.py
@@ -998,6 +998,23 @@ class FederationServer(FederationBase):
 
             origin, event = next
 
+            # Prune the event queue if its getting large.
+            #
+            # We do this *after* handling the first event as the common case is
+            # that the queue is empty (/has the single event in), and so there's
+            # no need to do this check.
+            pruned = await self.store.prune_staged_events_in_room(room_id, room_version)
+            if pruned:
+                # If we have pruned the queue check we need to refetch the next
+                # event to handle.
+                next = await self.store.get_next_staged_event_for_room(
+                    room_id, room_version
+                )
+                if not next:
+                    break
+
+                origin, event = next
+
             lock = await self.store.try_acquire_lock(
                 _INBOUND_EVENT_HANDLING_LOCK_NAME, room_id
             )
--- a/synapse/storage/databases/main/event_federation.py
+++ b/synapse/storage/databases/main/event_federation.py
@@ -1252,17 +1252,39 @@ class EventFederationWorkerStore(EventsWorkerStore, SignatureWorkerStore, SQLBas
         referenced_events: Set[str] = set()
         seen_events: Set[str] = set()
         for row in rows:
-            seen_events.add(row["event_id"])
+            event_id = row["event_id"]
+            seen_events.add(event_id)
             event_d = db_to_json(row["event_json"])
 
             # We don't bother parsing the dicts into full blown event objects,
             # as that is needlessly expensive.
+
+            # We haven't checked that the `prev_events` have the right format
+            # yet, so we check as we go.
+            prev_events = event_d.get("prev_events", [])
+            if not isinstance(prev_events, list):
+                logger.info("Invalid prev_events for %s", event_id)
+                continue
+
             if room_version.event_format == EventFormatVersions.V1:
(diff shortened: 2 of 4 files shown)


synapse/storage/databases/main/event_federation.py
```diff
@@ -1207,6 +1213,78 @@ def _get_next_staged_event_for_room_txn(txn):
...
+
+        # First check the size of the queue.
+        count = await self.db_pool.simple_select_one_onecol(
+            table="federation_inbound_events_staging",
+            keyvalues={"room_id": room_id},
+            retcol="COALESCE(COUNT(*), 0)",
+            desc="prune_staged_events_in_room_count",
+        )
+
+        if count < 100:
```
ShadowJonathan, line 1236: Should we make this configurable?

erikjohnston: Could do, though I don't really see it being very useful :shrug:

synapse/storage/databases/main/event_federation.py
```diff
@@ -1207,6 +1213,78 @@ def _get_next_staged_event_for_room_txn(txn):
...
+            return False
+
+        # If the queue is too large, then we want clear the entire queue,
+        # keeping only the forward extremities (i.e. the events not referenced
+        # by other events in the queue). We do this as that we can always
```
richvdh, line 1241:
```suggestion
        # by other events in the queue). We do this so that we can always
```

synapse/storage/databases/main/event_federation.py
```diff
@@ -1207,6 +1213,78 @@ def _get_next_staged_event_for_room_txn(txn):
...
+            event_d = db_to_json(row["event_json"])
+
+            # We don't bother parsing the dicts into full blown event objects,
+            # as that is needlessly expensive.
+            if room_version.event_format == EventFormatVersions.V1:
+                referenced_events.update(
+                    event_id for event_id, _ in event_d.get("prev_events", [])
+                )
+            else:
+                referenced_events.update(event_d.get("prev_events", []))
```
richvdh, line 1265: I don't think we validate that `prev_events` has the right shape when we receive the event, so you'll need to add some validation here otherwise we'll end up exploding with malformed events.

richvdh (changes requested):

synapse/storage/databases/main/event_federation.py
```diff
@@ -1207,6 +1213,100 @@ def _get_next_staged_event_for_room_txn(txn):
...
+                        break
+
+                    referenced_events.add(prev_event_id)
+            else:
+                for prev_event_id in prev_events:
+                    if not isinstance(prev_event_id, str):
+                        logger.info("Invalid prev_events for %s", event_id)
+                        break
+
+                    referenced_events.add(prev_event_id)
```
richvdh, line 1287: Could use a generator to reduce duplication?

```suggestion
            if room_version.event_format == EventFormatVersions.V1:
                def get_prev_events():                    
                    for prev_event_tuple in prev_events:
                        if not isinstance(prev_event_tuple, list) or len(prev_events) != 2:
                            logger.info("Invalid prev_events for %s", event_id)
                            break
                        yield prev_event_tuple[0]
                prev_event_ids = get_prev_events()
            else:
                prev_event_ids = prev_events
          
            for prev_event_id in prev_event_ids:
                  if not isinstance(prev_event_id, str):
                      logger.info("Invalid prev_events for %s", event_id)
                      break
                  referenced_events.add(prev_event_id)
```

synapse/storage/databases/main/event_federation.py
```diff
@@ -1252,17 +1252,39 @@ async def prune_staged_events_in_room(
...
 
             # We don't bother parsing the dicts into full blown event objects,
             # as that is needlessly expensive.
+
+            # We haven't checked that the `prev_events` have the right format
+            # yet, so we check as we go.
+            prev_events = event_d.get("prev_events", [])
+            if not isinstance(prev_events, list):
+                logger.info("Invalid prev_events for %s", event_id)
+                continue
```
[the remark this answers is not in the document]
erikjohnston: Yeeees, ideally we'd have checked long before here that the `Event` was valid (given we'd already have parsed the JSON into an event object before persistence). I'm slightly hesitant about dropping the "invalid" events here in case we update the format of events and forget to change it here, which would cause us to *sometimes* drop those events (as this function does not get called for each inbound event).

synapse/storage/databases/main/event_federation.py
erikjohnston: TBH I find the structure there harder to follow :shrug:

synapse/federation/federation_server.py
```diff
@@ -998,6 +998,23 @@ async def _process_incoming_pdus_in_room_inner(
 
             origin, event = next
 
+            # Prune the event queue if its getting large.
```
richvdh, line 1001:
```suggestion
            # Prune the event queue if it's getting large.
```

richvdh (approved):
fair enough then.

Issue #6116, opened 2019-09-26
APT repository with RC releases


fjerhammer:
**Description:**
I think it would be nice to have an additional APT repository containing both the stable releases from the "main" repository and the current RC releases, if any.
It would be an easy and non-intrusive way for users to test out the new RCs, and they would probably be more inclined to do it if they don't have to set up a Pip environment, manually create a systemd service for starting it automatically and all that stuff.

If this repository contains the same packages as the stable repository, I suppose they would be functionally identical when no RC is available, and at those times you could easily switch back to "main" if you decide that you don't want to be an RC guinea pig anymore. All of this without leaving your trusty old APT behind. :)

Being a Python developer I know Pip and use it on a daily basis, but if that wasn't the case, I think I would be hesitant to maintain a piece of (fairly important) software using a tool that I haven't used before and isn't fully comfortable with.
You know, if it suddenly throws a weird error at an unfitting time that I would have to spend time investigating - compared to just doing a simple "apt upgrade" where everything usually has been taken care of.

I don't know much about building Debian packages, but since you already have a package building infrastructure in place, I guess it would be possible to create such a repository without too much work.
And if it can help you catch just a few bugs that wouldn't have been reported otherwise, it might be worth it. :)

I, for one, would use it - I would like to test the RCs, but haven't found the time to switch away from the matrix-synapse-py3 package yet.
And honestly I would prefer to keep everything in APT instead, seeing that my Matrix server is already configured to alert me about available packages upgrades and the like.

richvdh:
It would also be nice to have a way to test our `.deb`s before we push them out to the whole world.

richvdh:
This is now available: add the `prerelease` component to the apt-sources line to get access to future RCs:

```
deb [signed-by=/usr/share/keyrings/matrix-org-archive-keyring.gpg] https://packages.matrix.org/debian/ <os_release> main prerelease
```

PR #10391, opened 2021-07-14
Set section for prerelease debs

richvdh:
This is part of fixing https://github.com/matrix-org/synapse/issues/6116: we want to put RC debs into a different place than release debs, so reprepro has to be able to tell them apart.

### DIFF
--- a/changelog.d/10391.misc
+++ b/changelog.d/10391.misc
@@ -1,0 +1,1 @@
+When building Debian packages for prerelease versions, set the Section accordingly.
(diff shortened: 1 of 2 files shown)


richvdh:
(see also https://github.com/matrix-org/pkg-repo-configs/commit/96b5f0d76822233900d5501d845e8daf9c5f3f52 which configures reprepro to accept this)

erikjohnston (approved):
Sure

Issue #9733, opened 2021-04-02
Youtube captions are useless


eras:
### Description

At some point Youtube has updated the site and now all (?) captions generated by Synapse for the site are:

Before you continue to YouTube
Sign in a Google company Before you continue to YouTube Google uses cookies and data to: Deliver and maintain services, like tracking outages and protecting against spam, fraud, and abuse Measure audience engagement and site statistics to understand how our services are used

This is basically useless considering the primary point of the function, in particular in the case of a very popular website.

### Steps to reproduce

- send a Youtube URL `m.room.message` into a room, e.g. https://www.youtube.com/watch?v=RzJf02TIqxk
- wait for Synapse to produce a caption for the link
- witness the caption to contain no information about the actual link :)

Expected results:

- A descriptive message about the contents, such as the one produced by an up-to-date `youtube-dl --get-description`:

Authentic recordings from inside Hetzner Online's data center park
Just like birds and insects, each server sings its own unique song.

### Version information

- **Homeserver**: matrix.org

ShadowJonathan:
(FTR: This is about link previews)

This is not neccecarily a problem with synapse, synapse is doing it's job perfectly by previewing the url as-is fetched, because `matrix.org`'s server is located within the EU, Google has a tendency (heh) to present users with the cookie page before letting them access any part of the site, by law.

ShadowJonathan:
https://meta.discourse.org/t/youtube-embeddings-have-stopped-working-for-servers-in-europe/185128

:thinking:

eras:
I agree that it's not particularly a bug in Synapse; however the only parties able to resolve this issue are Google and Synapse (or the 3rd party component it's using), and I have my doubts about Google doing anything about it :).

IIRC e.g. Slack doesn't have this issue, so it's resolvable; even if with special handling.

eras:
For one plausible solution consider the following session:

```
% curl -s -A Mozilla -I https://www.youtube.com/watch?v=RzJf02TIqxk | grep -e '^HTTP' -e '^location'
HTTP/2 302 
location: https://consent.youtube.com/m?continue=https%3A%2F%2Fwww.youtube.com%2Fwatch%3Fv%3DRzJf02TIqxk&gl=FI&m=0&pc=yt&uxe=23983172&hl=fi&src=1

% curl -s -I https://www.youtube.com/watch?v=RzJf02TIqxk | grep -e '^HTTP' -e '^location'        
HTTP/2 200 
```

[...]

Issue #2752, opened 2018-01-04
Use oembed for url previews?


fuzzy76:
Wouldn't it be an idea to use oembed for url previews if the url announces support for it?

https://oembed.com

fuzzy76:
Also suggesting https://noembed.com :)

ZerataX:
very interested in this, just want to notice that this definitely needs to be optional per client, since I'm sure there are lots of people that don't want iframes to arbitrary sites even if they are okay with url previews

richvdh:
Using oembed more widely would be a good fix for https://github.com/matrix-org/synapse/issues/9733.

richvdh:
We have both https://github.com/matrix-org/synapse/pull/10392 and https://github.com/matrix-org/synapse/pull/10536 attempting to do this. Both do so by adding new configuration options so that people can provide their own list of URL/provider mappings.

I think that's problematic, because (a) it requires everyone to configure it themselves; (b) we have more than enough config options as it is, and we should decide if we really need more. It would be *much* nicer to automatically base use the list at https://oembed.com/providers.json.

So here's what I think we should do.

**Step one**: Embed https://oembed.com/providers.json in the synapse source code repo (it's MIT-licensed, so I think we're ok on that front), and use that to build the `_oembed_patterns` map at startup.

Everything else is then an optional extension that can come later:

* Consider if we need to keep our copy of `providers.json` up to date, and if so, how.
* Use the [autodiscovery mechanism](https://oembed.com/#section4) for those sites that support it, so that we can reduce `providers.json` to those with `discovery != true`.
* Allow specification of a custom `providers.json` file for people with exotic requirements.

fuzzy76:
Agreed. To quote [oEmbed.com](https://oembed.com/#section7):
> Providers and consumers are **strongly encouraged** to use the discovery mechanism, rather than the registry.

So this seems to be the recommended path.

srdjan-catalyst:
I have no problems implementing sourcing providers.json file, however we are those people with exotic requirements and will have to do a custom-providers.json or something too. Any suggestions where to place those files?

richvdh:
>  Any suggestions where to place those files?

I'd recommend a config option which can be used to specify the path for a custom `providers.json` file; if the config option is left unset, the embedded `providers.json` is used.

ShadowJonathan:
#9877 is also relevant here, though idk if thats a dup of this

srdjan-catalyst:
[...]

PR #10392, opened 2021-07-14
Make oEmbed globs configurable and extract more info from the response

t3chguy:
Fixes https://github.com/matrix-org/synapse/issues/9733

Needs moar tests, some refactoring to tidy up the code, to load the providers.json from upstream automagically

### Pull Request Checklist

### DIFF
--- a/synapse/config/repository.py
+++ b/synapse/config/repository.py
@@ -206,6 +206,8 @@ class ContentRepositoryConfig(Config):
                 "url_preview_accept_language"
             ) or ["en"]
 
+            self.oembed_globs = config.get("oembed_globs", {})
+
     def generate_config_section(self, data_dir_path, **kwargs):
         media_store = os.path.join(data_dir_path, "media_store")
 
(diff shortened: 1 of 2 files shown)


richvdh:
A thought occurs to me (sorry for moving the goalposts): rather than embedding a list of oEmbed provider URLs, could we instead pull the `<link rel="alternate" type="application/json+oembed">` value from the response?

Youtube's `watch` links include content such as the following:

```
<link rel="alternate" type="application/json+oembed" href="https://www.youtube.com/oembed?format=json&amp;url=https%3A%2F%2Fwww.youtube.com%2Fwatch%3Fv%3DRzJf02TIqxk" title="PURE RELAXATION - SERVER SOUNDS by Hetzner">
```

https://oembed.com/#section7 says:

> There are currently 290 providers in the registry. Providers and consumers are strongly encouraged to use the discovery mechanism, rather than the registry.


... of course that doesn't work for Twitter, so maybe we need to support a static list as well. Gah.

ajs124:
Instead of maintaining our own static list, why can't we just pull in the registry from https://oembed.com/providers.json?

clokep:
> Instead of maintaining our own static list, why can't we just pull in the registry from [oembed.com/providers.json](https://oembed.com/providers.json)?

I had recommended this to @t3chguy as a follow-up to keep this PR smaller. It is a bit unclear though if we need to support custom / configurable values. I can come up with use-cases for it, but I'm unsure if they're realistic or not.

richvdh (changes requested):
I've written some words about how I think we should approach this over at https://github.com/matrix-org/synapse/issues/2752#issuecomment-894273747. TL;DR: let's use a `providers.json` file rather than config.

Also: please could you separate the bits of this PR that aren't to do with updating the glob list out to a separate PR?

clokep:
Hello! The main gist of this PR was implemented by #10714. I think the "extra more info" bits could be moved to a separate PR, but will likely be difficult to rebase on the changes.

clokep:
I'm going to close this in-lieu of #10814, which I *think* makes the same improvements.

PR #10394, opened 2021-07-14
Experimental support for MSC3266 Room Summary API

t3chguy:
See [MSC3266](https://github.com/matrix-org/matrix-doc/pull/3266).

### Pull Request Checklist


synapse/handlers/room_summary.py
```diff
@@ -0,0 +1,285 @@
...
+            remote_room_hosts: List of homeservers to attempt to fetch the data from.
+
+        Returns:
+            summary dict to return
+        """
+        logger.info("Requesting summary for %s via %s", room_id, remote_room_hosts)
+
+        return None  # TODO federation API
+
+    # TODO extract into mixin/helper method
```
clokep, line 128: Might make sense to go on `synapse.api.auth.Auth`? Or could leave it here and make it public and have the spaces summary use the `RoomSummaryHandler`?

synapse/handlers/room_summary.py
```diff
@@ -0,0 +1,285 @@
...
+            "num_joined_members": stats["joined_members"],
+            "avatar_url": stats["avatar"],
+            "join_rules": stats["join_rules"],
+            "world_readable": (
+                stats["history_visibility"] == HistoryVisibility.WORLD_READABLE
+            ),
+            "guest_can_join": stats["guest_access"] == "can_join",
+            "creation_ts": create_event.origin_server_ts,
+            "room_type": create_event.content.get(EventContentFields.ROOM_TYPE),
+            "allowed_spaces": allowed_rooms,  # TODO unspecced for MSC3266
```
clokep, line 278: This is also only returned over federation for MSC2946.

synapse/handlers/room_summary.py
```diff
@@ -0,0 +1,285 @@
...
+                        return True
+
+        logger.info(
+            "room %s is unpeekable and requester %s is not a member / not allowed to join, omitting from summary",
+            room_id,
+            requester or origin,
+        )
+        return False
+
+    async def _build_room_entry(self, room_id: str) -> JsonDict:
```
clokep, line 241: Could probably also make this public and have the `SpaceSummaryHandler` call it?

synapse/handlers/room_summary.py
t3chguy: Hmm its listed in the example response under Client-server API

https://github.com/matrix-org/matrix-doc/blob/kegan/spaces-summary/proposals/2946-spaces-summary.md#client-server-api

clokep: I think that's incorrect, see https://github.com/matrix-org/synapse/blob/2d16e69b4bf09b5274a8fa15c8ca4719db8366c1/synapse/handlers/space_summary.py#L233-L236

Also under the [server-server API section](https://github.com/matrix-org/matrix-doc/blob/kegan/spaces-summary/proposals/2946-spaces-summary.md#server-server-api):

> Response fields are the same as the Client-Server API, with the addition of:
>
> * rooms: Each room contains an additional field:
>    * allowed_room_ids: A list of room IDs which give access to this room per MSC3083

(I haven't yet updated the implementation to refer to rooms instead of spaces in this code.)

t3chguy: Indeed, well we only need it over federation anyhow to do the local filtering based on requester, same as the space summary API

synapse/handlers/room_summary.py
t3chguy: If we're happy for the space summary API to include `is_encrypted` then it'd be the same identical method, yes

t3chguy: By call it here do you mean `SpaceSummaryHandler.build_room_entry(self, room_id)` and knowingly passing the wrong self? Is that pythonic?

t3chguy: Have extracted into a mixin

synapse/config/experimental.py
```diff
@@ -32,3 +32,6 @@ def read_config(self, config: JsonDict, **kwargs):
 
         # MSC2716 (backfill existing history)
         self.msc2716_enabled = experimental.get("msc2716_enabled", False)  # type: bool
+
+        # MSC3266 (room summary api)
+        self.mxc3266_enabled = experimental.get("mxc3266_enabled", False)  # type: bool
```
jryans, line 37:
```suggestion
        self.msc3266_enabled = experimental.get("msc3266_enabled", False)  # type: bool
```

t3chguy: Yeah hah I hit this locally and just forgot to push, thanks!

t3chguy:
blocking this for now as it'll conflict with @clokep's ongoing work in the area which has much higher priority

synapse/api/auth.py
```diff
@@ -620,3 +621,115 @@ async def check_user_in_room_or_world_readable(
 
     async def check_auth_blocking(self, *args, **kwargs) -> None:
         await self._auth_blocking.check_auth_blocking(*args, **kwargs)
+
+    async def is_room_visible(
```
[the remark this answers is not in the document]
t3chguy: > Curious why this was changed from accessible to visible?

I felt like accessible suggested joinable, which may not be the case if it is only history_visisbility allowing you to see it but it is otherwise invite-only.

> Is there any changes from SpaceSummary._is_local_room_accessible or was it copied verbatim?

Verbatim other than changing the docstring/comment to not talk about summaries

synapse/rest/client/v1/room.py
```diff
@@ -1487,8 +1476,46 @@ async def on_GET(
...
+        self._room_summary_handler = hs.get_room_summary_handler()
+
+    async def on_GET(
+        self, request: SynapseRequest, room_identifier: str
+    ) -> Tuple[int, JsonDict]:
+        try:
+            requester = await self._auth.get_user_by_req(request, allow_guest=True)
+            requester_user_id: Optional[str] = requester.user.to_string()
+        except MissingClientTokenError:
+            requester_user_id = None
```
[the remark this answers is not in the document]
t3chguy: I copied this from publicRooms iirc, I'll double check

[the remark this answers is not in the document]
clokep: Ah, it's quite possible this is the cleanest way! Our auth code is a bit complicated.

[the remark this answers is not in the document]
t3chguy: I might abstract it behind api.auth and reuse in both places for our sanity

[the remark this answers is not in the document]
t3chguy: Actually, probably sane to not abstract it away, that one is slightly differing

https://github.com/matrix-org/synapse/blob/faa8bf1ca87380ebe70c9235a130d620eb473b6c/synapse/rest/client/v1/room.py#L740-L755

We're also excepting on a subtype of the error

synapse/api/auth.py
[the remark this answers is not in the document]
t3chguy: It feels like a fundamental & reusable implementation of MSC3173 which can be reused anywhere, so will keep it extracted to api.auth

t3chguy:
@clokep happy to break the refactoring out into a separate PR if that helps review

synapse/handlers/room_summary.py
```diff
@@ -112,6 +113,51 @@ def __init__(self, hs: "HomeServer"):
             "get_room_hierarchy",
         )
 
+    async def _is_remote_room_accessible(
+        self, requester: Optional[str], room_id: str, room: JsonDict
+    ) -> bool:
```
clokep, line 118: Please revert the change to move this.

synapse/handlers/room_summary.py
```diff
@@ -150,10 +196,10 @@ async def get_space_summary(
                 is overridden by MAX_ROOMS_PER_SPACE.
 
         Returns:
-            summary dict to return
+            hierarchy dict to return
```
clokep, line 199: A bunch of changes from `summary` -> `hierarchy` seem incorrect here.

tests/test_state.py
```diff
@@ -175,6 +176,7 @@ def setUp(self):
         hs.config = default_config("tesths", True)
         hs.get_datastore.return_value = self.store
         hs.get_state_handler.return_value = None
+        hs.get_event_auth_handler.return_value = None
```
clokep, line 179: I'm not sure why we need this change? It doesn't seem like any event auth code is touched in this PR?

synapse/api/auth.py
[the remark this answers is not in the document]
clokep: I think I'd be keen to keep this where it was to simplify this PR.

tests/test_state.py
t3chguy: Strange, had test failures without this at one point but now is fine without

synapse/handlers/room_summary.py
```diff
@@ -519,8 +519,9 @@ async def _summarize_local_room(
 
         room_entry = await self._build_room_entry(room_id, for_federation=bool(origin))
 
-        # If the room is not a space, return just the room information.
-        if room_entry.get("room_type") != RoomTypes.SPACE:
+        # If the room is not a space or the children don't matter, return just
+        # the room information.
+        if room_entry.get("room_type") != RoomTypes.SPACE or max_children == 0:
```
clokep, line 524: This change is from #10569 and means we don't need to use a separate method to summarize a single room.

clokep:
I think this is good to go now. I ended up restructuring it a bit to fit in better with changes in #10569. I also stripped out a layer of unsupported federation code. I think it makes sense to add that all at once instead of having some methods which just always raise an exception.

clokep:
I had to resolve some conflicts from develop, but I think this looks good.

clokep (approved):

PR #10395, opened 2021-07-14
Make deprecation notice of the spam checker doc more obvious

### LINKED ISSUE #10333: Deprecation notice on the spam checker doc isn't obvious enough
Looks like the deprecation notice on https://matrix-org.github.io/synapse/latest/spam_checker.html is easier to miss than I thought. @anoadragon453 suggested:

> as far as the warning goes, I would start by putting it below "Handling spam in Synapse", then probably bumping it to a higher header number (maybe even just h1).
>
> we can do red - none of the themes have red text as the default (yet)

babolivier:
Old:

image

New:

image

Fixes #10333

### DIFF
--- a/docs/spam_checker.md
+++ b/docs/spam_checker.md
@@ -1,7 +1,7 @@
 <h2 style="color:red">
 This page of the Synapse documentation is now deprecated. For up to date
 documentation on setting up or writing a spam checker module, please see
-<a href="https://matrix-org.github.io/synapse/develop/modules.html">this page</a>.
+<a href="modules.md">this page</a>.
 </h2>
 
 # Handling spam in Synapse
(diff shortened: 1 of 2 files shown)


babolivier:
PRing against master on purpose so it reaches the stable docs without having to wait for a full release cycle

richvdh:
(please don't forget to merge back into develop after merging to master)

docs/spam_checker.md
```diff
@@ -1,6 +1,8 @@
-**Note: this page of the Synapse documentation is now deprecated. For up to date
+<h2 style="color:red">
+This page of the Synapse documentation is now deprecated. For up to date
 documentation on setting up or writing a spam checker module, please see
-[this page](https://matrix-org.github.io/synapse/develop/modules.html).**
+<a href="modules.md">this page</a>.
```
anoadragon453, line 4: Curious that markdown links don't work within custom HTML tags - but I guess that's probably a markdown thing.

anoadragon453 (approved):
Looking good!

PR #10398, opened 2021-07-14
Stagger send presence to remotes

erikjohnston:
This is to help with performance, where trying to connect to thousands of hosts at once can consume a lot of CPU (due to TLS etc).

### DIFF
--- a/synapse/federation/sender/__init__.py
+++ b/synapse/federation/sender/__init__.py
@@ -13,10 +13,14 @@
 # limitations under the License.
 
 import abc
+import collections
 import logging
+import typing
 from typing import TYPE_CHECKING, Dict, Hashable, Iterable, List, Optional, Set, Tuple
 
+import attr
 from prometheus_client import Counter
+from typing_extensions import Literal
 
 from twisted.internet import defer
 
@@ -33,8 +37,12 @@ from synapse.metrics import (
     event_processing_loop_room_count,
     events_processed_counter,
 )
-from synapse.metrics.background_process_metrics import run_as_background_process
+from synapse.metrics.background_process_metrics import (
+    run_as_background_process,
+    wrap_as_background_process,
+)
 from synapse.types import JsonDict, ReadReceipt, RoomStreamToken
+from synapse.util import Clock
 from synapse.util.metrics import Measure
 
 if TYPE_CHECKING:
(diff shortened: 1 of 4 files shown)


synapse/federation/sender/__init__.py
```diff
@@ -137,6 +144,84 @@ async def get_replication_rows(
...
+    # being woken up.
+    _MAX_TIME_IN_QUEUE = 30.0
+
+    # The maximum duration in seconds between waking up consecutive destination
+    # queues.
+    _MAX_DELAY = 0.1
+
+    sender: "FederationSender" = attr.ib()
+    clock: Clock = attr.ib()
+    queue: "OrderedDict[str, Literal[None]]" = attr.ib(factory=OrderedDict)
```
babolivier, line 166:
```suggestion
    queue: OrderedDict[str, Literal[None]] = attr.ib(factory=OrderedDict)
```

Since none of the types in here have their imports conditioned on `TYPE_CHECKING`

synapse/federation/sender/__init__.py
```diff
@@ -137,6 +144,84 @@ async def get_replication_rows(
...
+        """Background process to drain the queue."""
+
+        if not self.queue:
+            return
+
+        assert not self.processing
+        self.processing = True
+
+        try:
+            # We start with a delay that should drain queue quickly enough that
```
babolivier, line 188:
```suggestion
            # We start with a delay that should drain the queue quickly enough that
```

synapse/federation/sender/__init__.py
```diff
@@ -137,6 +144,84 @@ async def get_replication_rows(
...
+                queue.attempt_new_transaction()
+
+                await self.clock.sleep(current_sleep_seconds)
+
+                if not self.queue:
+                    break
+
+                # More destinations may have been added to the queue, so we may
+                # need to reduce the delay to ensure everything gets processed
+                # with _MAX_TIME_IN_QUEUE seconds.
```
babolivier, line 216:
```suggestion
                # within _MAX_TIME_IN_QUEUE seconds.
```

babolivier (approved):

synapse/federation/sender/__init__.py
erikjohnston: FWIW I thought this was needed as `OrderdDict` wasn't generic in 3.6, but that doesn't look to be true.

erikjohnston: (Just saying that as that is the other reason to use quoted types)

PR #10399, opened 2021-07-14
Docs: Use something other than the document name to describe a page

anoadragon453:
Our documentation has a history of using a document's name as a way to link to it, such as "See [workers.md]() for details". This makes sense when you're traversing a directory of files, but less sense when the files are abstracted away - as they are on the documentation website.

This PR changes the links to various documentation pages to something that fits better into the surrounding sentence, as you would when making any hyperlink on the web.

### DIFF
--- a/docs/MSC1711_certificates_FAQ.md
+++ b/docs/MSC1711_certificates_FAQ.md
@@ -132,7 +132,7 @@ your domain, you can simply route all traffic through the reverse proxy by
 updating the SRV record appropriately (or removing it, if the proxy listens on
 8448).
 
-See [reverse_proxy.md](reverse_proxy.md) for information on setting up a
+See [the reverse proxy page](reverse_proxy.md) for information on setting up a
 reverse proxy.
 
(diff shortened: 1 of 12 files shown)


docs/MSC1711_certificates_FAQ.md
```diff
@@ -132,7 +132,7 @@ your domain, you can simply route all traffic through the reverse proxy by
 updating the SRV record appropriately (or removing it, if the proxy listens on
 8448).
 
-See [reverse_proxy.md](reverse_proxy.md) for information on setting up a
+See [the reverse proxy page](reverse_proxy.md) for information on setting up a
```
clokep, line 135: Would it make more sense to use "the reverse proxy documentation" for each of these? I feel like referring them as a "page" isn't really useful. (What if it is a different part of the same page?!)

anoadragon453: Works for me!

clokep (approved):

PR #10400, opened 2021-07-14
Fix a number of logged errors caused by remote servers being down.

erikjohnston:
Each commit is reviewable independently.

### DIFF
--- a/synapse/handlers/directory.py
+++ b/synapse/handlers/directory.py
@@ -22,6 +22,7 @@ from synapse.api.errors import (
     CodeMessageException,
     Codes,
     NotFoundError,
+    RequestSendFailed,
     ShadowBanError,
     StoreError,
     SynapseError,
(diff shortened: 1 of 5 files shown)


synapse/handlers/directory.py
```diff
@@ -252,12 +253,14 @@ async def get_association(self, room_alias: RoomAlias) -> JsonDict:
...
                 )
+            except RequestSendFailed:
+                raise SynapseError(502, "Failed to fetch alias")
             except CodeMessageException as e:
                 logging.warning("Error retrieving alias")
                 if e.code == 404:
                     fed_result = None
                 else:
-                    raise
+                    raise SynapseError(502, "Failed to fetch alias")
```
babolivier, line 263: Shouldn't we log the original error as well? (or maybe it's already logged as part of the HTTP client logs, I can't remember)

babolivier (approved):

synapse/handlers/directory.py
erikjohnston: Good Q. We should log response success/errors in the http clients, so I don't think its necessary here.

ShadowJonathan: FWIW, re-raising here implicitly adds `from e`, so the traceback of the original exception will be available to anyone who catches it.

PR #10402, opened 2021-07-15
test twisted 21.7

richvdh:
We shouldn't merge this at least until 21.7 is properly released, since the type annotation changes will cause mypy to fail on twisted 21.2. For now it just serves as a way to run CI against 21.7.

Based on #10446 and #10450.


ShadowJonathan:
Could you maybe change the type hints to inline type hints as per [this effort](https://github.com/matrix-org/synapse/issues/8351)? If deferreds give errors on "not subscribable", wrap them in string quotations.

richvdh:
> Could you maybe change the type hints to inline type hints as per [this effort](https://github.com/matrix-org/synapse/issues/8351)? If deferreds give errors on "not subscribable", wrap them in string quotations.

done. Thanks!

richvdh:
superceded by #10490/#10491

PR #10404, opened 2021-07-15
Do not include signatures/hashes in make_{join,leave,knock} responses.

clokep:
Currently the responses to `/make_{join,leave,knock}` requests include the `signatures` and `hashes` keys, these will become invalid once the joining/leaving/knocking server modifies the event (before calling `/send_{join,leave,knock}`).

In order to avoid invalid signatures being added to events, we should not return them. This seems to match the [concept of templated events in the spec](https://matrix.org/docs/spec/server_server/latest#get-matrix-federation-v1-make-join-roomid-userid), although the examples show a much smaller subset of returned fields.

I originally tried to have the `EventBuilder` know about whether the event is a "full" event or not, but then you cannot properly instantiate an `EventBase` object. It seemed simpler to strip out the unnecessary fields at the end, although they'll still be calculated.

### DIFF
--- a/changelog.d/10404.bugfix
+++ b/changelog.d/10404.bugfix
@@ -1,0 +1,1 @@
+Responses from `/make_{join,leave,knock}` no longer include signatures, which will turn out to be invalid after events are returned to `/send_{join,leave,knock}`.
--- a/synapse/events/__init__.py
+++ b/synapse/events/__init__.py
@@ -291,6 +291,20 @@ class EventBase(metaclass=abc.ABCMeta):
 
         return pdu_json
 
+    def get_templated_pdu_json(self) -> JsonDict:
+        """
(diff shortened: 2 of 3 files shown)


changelog.d/10404.bugfix
```diff
@@ -0,0 +1 @@
+Responses from `/make_{join,leave,knock}` no longer include signatures, which will turn out to be invalid after events are returned to `/send_{join,leave,knock}`.
```
clokep, line 1: I'm unsure if there's a clearer changelog to give people here?

clokep:
I suppose I should add a test for this too, I'm unsure if we have any unit tests for `make_join` or if this would be more appropriate in sytest/complement.

erikjohnston (approved):

PR #10407, opened 2021-07-15
Add a PeriodicallyFlushingMemoryHandler to prevent logging silence

### LINKED ISSUE #10244: No logging during startup
Because we write logs to a buffer, and there isn't much logging during startup, Synapse tends to be ominously silent during startup. Worse, if there is a problem, and you kill synapse, you have no way of telling where it got stuck.

It is suggested that we could extend `MemoryHandler` to flush its logs every few seconds to alleviate this problem.

reivilibre:
This is intended to resolve #10244 .

It spins up a thread to do the flushing.

Alternative ideas might be to just keep a timer and flush when it gets logged to (but if nothing logs for a while, this won't help) or to somehow use the Twisted reactor and plug a timer into the reactor.


erikjohnston:
Since we're here I wonder if we should also override `shouldFlush` to always return True at first, and then fall back to default behaviour after Synapse has started (I think registering a callback with `reactor.callWhenRunning` to set a flag should work)?

synapse/logging/handlers.py
```diff
@@ -0,0 +1,47 @@
...
+        )
+        self._flushing_thread.start()
+
+    def _flush_periodically(self):
+        """
+        Whilst this handler is active, flush the handler periodically.
+        """
+
+        while self._active:
+            self.flush()
```
erikjohnston, line 42: We should add a comment here that `self.flush()` is thread safe

reivilibre:
Thanks to Rich and Patrick for their help in channel :).

I'll be away from tomorrow, so if it turns out to be good and you want it in the RC, you'll probably have to merge it in yourself...

See you soon!

erikjohnston (approved):

richvdh:
we should make sure we test this on a real server (matrix.org?) before pushing this out as the default option everywhere.

PR #10408, opened 2021-07-15
Add type hints to synapse.federation.transport.client.

clokep:
I wanted to expand on some of these and got carried away with fixing what we have.

* Adds type hints.
* Converts methods to be async instead of returning awaitables.
* Removes some useless `result` variables.


synapse/federation/transport/client.py
```diff
@@ -149,21 +160,21 @@ async def send_transaction(self, transaction, json_data_callback=None):
         """
         logger.debug(
             "send_data dest=%s, txid=%s",
-            transaction.destination,
-            transaction.transaction_id,
+            transaction.destination,  # type: ignore
+            transaction.transaction_id,  # type: ignore
```
erikjohnston, line 164: Ugh, can we do something to `Transaction` or `JsonEncodedobject` so we don't have to do this? I'd almost prefer *not* adding type hints to the transaction object...

clokep: I'll take a look, we do this in a couple of other spots too, but it is far from ideal.

clokep: I think the solution is to make these real objects. I have a WIP branch that does this, but it needs a bit more and I'd rather not let this bitrot. I think we can:

* Merge this as is and fix-up `Transaction` in the future.
* Remove the type hints for `Transaction` (although this ironically makes it harder to fix this in the future...)
* Leave this PR as is until we can fix `Transaction`.

I'd prefer the first option.

erikjohnston (approved):
On the assumption that fixing up the `Transaction` object is a thing that will happen soon

synapse/federation/transport/client.py
clokep: See #10542 for the eventual follow-up here.

PR #10411, opened 2021-07-16
allow specifying https:// proxy

dklimpel:
Replace and follow up: #9119
Fixes: #9090

I have tried to do some smaller commits for better reading.

### Pull Request Checklist


tests/http/test_proxyagent.py
```diff
@@ -218,29 +373,69 @@ def test_http_request_via_proxy_with_auth(self):
...
+        """Tests that TLS-encrypted requests can be made through a proxy"""
+        self._do_https_request_via_proxy(ssl=True, auth_credentials=None)
+
+    @patch.dict(
+        os.environ,
+        {"https_proxy": "https://bob:pinkponies@proxy.com", "no_proxy": "unused.com"},
+    )
+    def test_https_request_via_https_proxy_with_auth(self):
+        """Tests that authenticated, TLS-encrypted requests can be made through a proxy"""
+        self._do_https_request_via_proxy(ssl=True, auth_credentials=b"bob:pinkponies")
```
dklimpel, line 420: I have some trouble with this new tests.
I get always an failure with the https request in https proxy tunnel.
https://github.com/matrix-org/synapse/blob/292bdf076570702d19bea35db11205a8ebbc429c/tests/http/test_proxyagent.py#L569-L581

I do not know how to solve it. Or is there no solution because of https in https?

dklimpel: I have tested it in a live system. It works there. https over https proxy without problems.

dklimpel: I suspect that it is a mismatch with the TLS certificate.

synapse/http/proxyagent.py
```diff
@@ -63,35 +71,38 @@ class ProxyAgent(_AgentBase):
...
+        bindAddress: The local address for client sockets to bind to.
 
-        pool (HTTPConnectionPool|None): connection pool to be used. If None, a
+        pool: connection pool to be used. If None, a
             non-persistent pool instance will be created.
 
-        use_proxy (bool): Whether proxy settings should be discovered and used
+        use_proxy: Whether proxy settings should be discovered and used
             from conventional environment variables.
+            A proxy without scheme is assumed to be http.
```
richvdh, line 91: I think this line is still in the wrong place.

```suggestion
```

synapse/http/proxyagent.py
```diff
@@ -63,35 +71,38 @@ class ProxyAgent(_AgentBase):
...
+
+    Raises:
+        ValueError if given a proxy with a scheme we don't support.
```
richvdh, line 94:
```suggestion
        ValueError if use_proxy is set and the environment variables contain an invalid proxy specification.
```

synapse/http/proxyagent.py
```diff
@@ -263,93 +270,79 @@ def _http_proxy_endpoint(
...
 def parse_proxy(
     proxy: bytes, default_scheme: bytes = b"http", default_port: int = 1080
-) -> Tuple[bytes, bytes, int]:
+) -> Tuple[bytes, bytes, int, Optional[ProxyCredentials]]:
     """
-    Parse the scheme, hostname and port from a proxy connection byte string.
+    Given a HTTP proxy URL, breaks it down into components and checks that it
+    has a hostname (otherwise it is not right useful to us trying to find a
+    proxy) and asserts that the URL has the scheme as that is all we support.
+    Parse the scheme, username, password, hostname and port from a proxy connection byte string.
```
richvdh, line 315: docstrings should start with a one-line summary:

```suggestion
    Parse a proxy connection string.
    
    Given a HTTP proxy URL, breaks it down into components and checks that it
    has a hostname (otherwise it is not useful to us when trying to find a
    proxy) and asserts that the URL has a scheme we support.

```

synapse/http/proxyagent.py
```diff
@@ -263,93 +270,79 @@ def _http_proxy_endpoint(
...
         default_scheme: The default scheme to return if one is not found in `proxy`. Defaults to http
         default_port: The default port to return if one is not found in `proxy`. Defaults to 1080
 
     Returns:
-        A tuple containing the scheme, hostname and port.
+        A tuple containing the scheme, hostname, port and ProxyCredentials.
+            If no credentials were found, the ProxyCredentials instance is replaced with None.
+
+    Raise:
+        RuntimeError if proxy has no hostname or unsupported scheme.
```
richvdh, line 327: why RuntimeError rather than ValueError?

synapse/http/proxyagent.py
```diff
@@ -263,93 +270,79 @@ def _http_proxy_endpoint(
...
 
         tls_options_factory: the TLS options to use when connecting through a https proxy
+
         kwargs: other args to be passed to HostnameEndpoint
 
     Returns:
-        interfaces.IStreamClientEndpoint|None: endpoint to use to connect to the proxy,
-            or None
-
-    Raises: ValueError if given a proxy with a scheme we don't support.
```
richvdh: We should specify the behaviour of this function when the proxy string is invalid.

synapse/http/proxyagent.py
```diff
@@ -237,7 +244,7 @@ def request(self, method, uri, headers=None, bodyProducer=None):
                 self._reactor, parsed_uri.host, parsed_uri.port, **self._endpoint_kwargs
             )
 
-        logger.debug("Requesting %s via %s", uri, endpoint)
+        logger.debug(f"Requesting {uri !s} via {endpoint !s}")
```
richvdh, line 247: `logger.debug` (and other logger.*` methods) does its string interpolation lazily: ie, it will not interpolate unless the line is actually going to be logged. It's best to leave this as `%s` etc.

tests/http/test_proxyagent.py
```diff
@@ -37,33 +40,179 @@
...
+        ]
+    )
+    def test_parse_proxy(
+        self,
+        proxy: bytes,
+        scheme: bytes,
+        hostname: bytes,
+        port: int,
+        credentials: Optional[bytes],
+    ):
```
richvdh, line 174: this is a bit hard to follow: which arguments are inputs, and which are expected outputs?

I think you should `s/scheme/expected_scheme/` (and likewise for hostname, port, credentials), and `s/proxy/proxy_string/`.

A docstring wouldn't go amiss either.

richvdh (changes requested):
this looks excellent. A couple of nits. I'll have a look to see what the problem with the tests is.

synapse/http/proxyagent.py
dklimpel: Ok. I would also like to write a documentation.

synapse/http/proxyagent.py
dklimpel: Same code in sygnal.
https://github.com/matrix-org/sygnal/blob/8cbe3d6b21e60987beb83ca21db5675fbdcc3e14/sygnal/helper/proxy/__init__.py#L31-L64

tests/http/test_proxyagent.py
dklimpel: Hopefully I have understood you correctly. I did an update.

synapse/http/proxyagent.py
richvdh: ok, but I see that as a bug in sygnal rather than something we ought to replicate here.

richvdh (changes requested):
@dklimpel I've pushed a change which fixes up the tests. The problem was that the way we were setting up the connections meant that the traffic from the client was going straight to the server, rather than first being un-SSLed by the proxy.

Could you switch the exceptions back to `ValueError` and then I think we can merge this?

dklimpel:
> @dklimpel I've pushed a change which fixes up the tests. The problem was that the way we were setting up the connections meant that the traffic from the client was going straight to the server, rather than first being un-SSLed by the proxy.
> 
> Could you switch the exceptions back to `ValueError` and then I think we can merge this?

Great.
This test is a little bit to crazy for me. :)

richvdh (approved):
lgtm. thanks both @dklimpel and @Bubu !

PR #10414, opened 2021-07-16
Fix exception when failing to get remote room list

erikjohnston:
I missed one in https://github.com/matrix-org/synapse/pull/10400

### DIFF
--- a/synapse/handlers/room_list.py
+++ b/synapse/handlers/room_list.py
@@ -381,12 +381,12 @@ class RoomListHandler(BaseHandler):
                     Codes.UNRECOGNIZED,
                     Codes.NOT_FOUND,
                 ):
-                    # Not an error that should trigger a fallback.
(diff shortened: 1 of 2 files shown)


synapse/handlers/room_list.py
```diff
@@ -381,9 +381,13 @@ async def get_remote_public_room_list(
                     Codes.UNRECOGNIZED,
                     Codes.NOT_FOUND,
                 ):
+                    # Not an error that should trigger a fallback.
                     logger.debug("Falling back to locally-filtered /publicRooms")
```
babolivier, line 385: Is this right? The comment says the error shouldn't trigger a fallback but the following log line says it's triggering a fallback.

babolivier (approved):
lgtm otherwise

synapse/handlers/room_list.py
erikjohnston: I put this comment in the wrong place...

PR #10415, opened 2021-07-16
Don’t have Scripts in Modules

V02460:
*This exists because rpmlint complained while packaging Synapse for Fedora.*

This PR removes the shebang lines from all files belonging to Python modules.

Afaict this does not remove functionality:
- For the files under `_script` the shebang can be removed as they are wrapped by a *real* script (with shebang).
- The files under `app` are actually meant to be called directly, but should not occur as cli tools. Without a shebang, they can still be called by their module name, e.g. `python3 -m synapse.app.homeserver`.

Tangential Suggestion: After reading into Python and scripts, using `entry_point` in `setup.py` would make those wrappers scripts unnecessary.

### Pull Request Checklist

### DIFF
--- a/synapse/app/admin_cmd.py
+++ b/synapse/app/admin_cmd.py
@@ -1,4 +1,3 @@
-#!/usr/bin/env python
 # Copyright 2019 Matrix.org Foundation C.I.C.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
--- a/changelog.d/10415.misc
+++ b/changelog.d/10415.misc
@@ -1,0 +1,1 @@
+Remove shebang line from module files.
--- a/synapse/_scripts/review_recent_signups.py
+++ b/synapse/_scripts/review_recent_signups.py
@@ -1,4 +1,3 @@
-#!/usr/bin/env python
 # Copyright 2021 The Matrix.org Foundation C.I.C.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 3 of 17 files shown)


babolivier:
> This exists because rpmlint complained while packaging Synapse for Fedora.

Is passing rpmlint a required step when packaging for Fedora? If not I'd be in favour of keeping these around since they're not doing any harm.

V02460:
Yes, passing rpmlint is part of the [Fedora Packaging Guidelines](https://docs.fedoraproject.org/en-US/packaging-guidelines/#_use_rpmlint) and when run it errors for the given files. It would be preferable to keep downstream changes to a minimum and this would help.

synapse/app/admin_cmd.py
```diff
@@ -1,4 +1,3 @@
-#!/usr/bin/env python
```
richvdh, line 1: looks like you're setting the `+x` bit on this? That seems to be in opposition to the removal of the shebang?

richvdh (commented):
looks good otherwise

synapse/app/admin_cmd.py
V02460: Good spot, that wasn’t my intention. Fixed it.

richvdh (approved):
lgtm, thank you for the contribution!

For future reference: please don't rebase after a review has taken place - it makes it hard to see what has changed since the previous review.

PR #10416, opened 2021-07-16
Move the docker image build to Github Actions

richvdh:
it's flaky on circleCI, and having to manage multiple CI providers is painful.


.github/workflows/docker.yml
```diff
@@ -0,0 +1,72 @@
...
+                  tag=${GITHUB_REF#refs/tags/}
+                  ;;
+              *)
+                  tag=${GITHUB_SHA}
+                  ;;
+          esac
+          echo "::set-output name=tag::$tag"
+
+        # for release builds, we want to get the amd64 image out asap, so first
+        # we do an amd64-only build, before following up with a multiarch build.
```
babolivier, line 54: Can't we do both in parallel (which CircleCI was doing)?

babolivier (approved):

.github/workflows/docker.yml
richvdh: no it wasn't. It was building an amd64-only build first, before kicking off a parallel build of amd64+arm64.

PR #10426, opened 2021-07-19
Update the notification email subject when invited to a space.

clokep:
I debated about re-using `invite_from_person_to_room`, but passing in whether it was a room or a space, but that seemed less flexible.

### DIFF
--- a/changelog.d/10426.feature
+++ b/changelog.d/10426.feature
@@ -1,0 +1,1 @@
+Email notifications now state whether an invitation is to a room or a space.
--- a/synapse/config/emailconfig.py
+++ b/synapse/config/emailconfig.py
@@ -39,12 +39,13 @@ DEFAULT_SUBJECTS = {
     "messages_from_person_and_others": "[%(app)s] You have messages on %(app)s from %(person)s and others...",
     "invite_from_person": "[%(app)s] %(person)s has invited you to chat on %(app)s...",
     "invite_from_person_to_room": "[%(app)s] %(person)s has invited you to join the %(room)s room on %(app)s...",
+    "invite_from_person_to_space": "[%(app)s] %(person)s has invited you to join the %(space)s space on %(app)s...",
     "password_reset": "[%(server_name)s] Password reset",
     "email_validation": "[%(server_name)s] Validate your email",
 }
 
 
-@attr.s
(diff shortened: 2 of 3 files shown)


changelog.d/10426.feature
```diff
@@ -0,0 +1 @@
+Email notifications now state whether an invitation is to a room or a space.
```
clokep, line 1: Not sure this is really "feature"-worthy?

babolivier: I'd say it is since it's a user-facing change (that's not a bugfix)

babolivier (approved):
I assume we don't need to change the email templates themselves? I've had a quick look and think not but would like to be sure.

clokep:
> I assume we don't need to change the email templates themselves? I've had a quick look and think not but would like to be sure.

Ah sorry, was in a rush to get this up and should have mentioned this. The actual template says:

https://github.com/matrix-org/synapse/blob/c356b4bf422430cd5769c9bf90756fca2efd8451/synapse/res/templates/room.html#L24

https://github.com/matrix-org/synapse/blob/c356b4bf422430cd5769c9bf90756fca2efd8451/synapse/res/templates/room.txt#L4

----

Maybe the HTML version should be updated? What do you think?

babolivier:
> Maybe the HTML version should be updated? What do you think?

Yeah I think it should, at least the html version. It sounds confusing to me to get an email about a new conversation when you're invited into a space.

Not sure what the right fix is here, either having two templates or adding a variable to the existing one that says whether the room is a space or a "normal" one? I'd say the latter but ymmv.

nadonomy:
@babolivier to add some more context. We're improving this in 2 different streams of effort:

1. The bare minimum for email invites to not be inaccurate (i.e. don't use the word room when it's a Space)
2. Tight loop cross-functional iteration on the email journeys holistically: Sending email invites, email templates, clicking through to the app, etc

In this PR/sprint we're focused on the former and will return to the latter in the near future/August. For this, I think 'Join the conversation' is sufficiently generic enough it's correct for both rooms and spaces.

I see this already approved so assume that works for both yourself & @clokep, but hopefully gives more context and confidence in the changes!

clokep:
Thanks! I'm going to merge this then!

PR #10428, opened 2021-07-19
Remove old R30 because R30v2 supercedes it

reivilibre:
The plan is to remove the old R30 some time after R30v2 has been out a while.

This PR is **not** to be merged immediately.

new R30 was introduced by #10332


synapse/app/phone_stats_home.py
```diff
@@ -103,10 +103,6 @@ async def phone_stats_home(hs, stats, stats_process=_stats_process):
     daily_sent_messages = await hs.get_datastore().count_daily_sent_messages()
     stats["daily_sent_messages"] = daily_sent_messages
 
-    r30_results = await hs.get_datastore().count_r30_users()
-    for name, count in r30_results.items():
-        stats["r30_users_" + name] = count
-
     r30v2_results = await hs.get_datastore().count_r30_users()
     for name, count in r30v2_results.items():
         stats["r30v2_users_" + name] = count
```
babolivier, line 108: Won't we want to rename the metric then? It feels a bit weird to me to keep a v2 around with the v1 gone. Though I don't know how much of our tooling would be confused by such a change.

babolivier (approved):

synapse/app/phone_stats_home.py
clokep: I think we'll want to keep it since you want to keep the data separate (forever).

reivilibre: Indeed (what Patrick said) — old R30 data will still live on in Panopticon (and old servers that don't get upgraded will continue to send old R30). The two are not compatible, so it's much like a version number for a library, or how `/v2` endpoints in the Matrix APIs won't be renamed to `v1` again in the future just because the `v1` is obsolete (to us).

babolivier: Fair enough.

clokep:
> The plan is to remove the old R30 some time after R30v2 has been out a while.

Any idea what "a while" means? 😄 It looks like v1.39.0 introduced the new R30v2 for reference.

reivilibre:
> Any idea what "a while" means? smile It looks like v1.39.0 introduced the new R30v2 for reference.

Thanks for bringing this back up. I think we originally said 4 weeks, but Panopticon wasn't set up to receive the data in time.

We started collecting R30v2 since 2021-08-16 (as far as I remember). v1.43 should be out on 2021-09-21, so I think it would be sensible to include this in the next release.

callahad:
We want historic data to pile up for a while for R30v2 before dropping v1... let's revisit this next year.

callahad:
Still useful for Element's reporting (trying to keep apples-to-apples comparisons of metrics for at least a year) so punting this into 2023.

erikjohnston:
This can be resurrected!

docs/usage/administration/monitoring/reporting_homeserver_usage_statistics.md
```diff
@@ -42,11 +42,6 @@ The following statistics are sent to the configured reporting endpoint:
...
-| `r30_users_ios`            | int    | The number of 30 day retained users, as defined above. Filtered only to clients with "iOS" in the user agent string.                                                                                                                                                                            |
-| `r30_users_electron`       | int    | The number of 30 day retained users, as defined above. Filtered only to clients with "Electron" in the user agent string.                                                                                                                                                                       |
-| `r30_users_web`            | int    | The number of 30 day retained users, as defined above. Filtered only to clients with "Mozilla" or "Gecko" in the user agent string.                                                                                                                                                             |
```
MadLittleMods, line 49: Change since last review. I removed the stat docs for the old R30 metric as well. That was only dangling reference to the old stat I could still find.

MadLittleMods:
*Discussed in the triage meeting today.* Revived this and put it in the review queue now :syringe:

clokep (approved):

MadLittleMods:
Thanks for the review @clokep and nice evergreen PR @reivilibre!

PR #10431, opened 2021-07-19
Enable docker image caching for the deb build

richvdh:
This reduced the time spent building the packages from about 13 minutes to about 10 minutes (most of the rest is spent running the unit tests).

I'm not entirely sure this is the best approach. The images that we use to build the packages are pretty static, so another option would be just to build them once and host them on dockerhub or a private docker repository. Thoughts or comments welcome.

~~Based on #10429~~.


erikjohnston (approved):
I think this makes sense

PR #10435, opened 2021-07-20
Send unstable-prefixed room_type in store-invite IS API requests

t3chguy:
Related https://github.com/matrix-org/sydent/pull/375
Implementation of https://github.com/matrix-org/matrix-doc/pull/3288

### Pull Request Checklist

There look to be no tests exercising this path so couldn't extend any existing tests and it seems simple enough that it doesn't warrant a lot of testing effort at this time.


synapse/handlers/room_member.py
```diff
@@ -1138,6 +1138,11 @@ async def _make_and_store_3pid_invite(
         if room_name_event:
             room_name = room_name_event.content.get("name", "")
 
+        room_type = ""
+        room_create_event = room_state.get((EventTypes.Create, ""))
+        if room_create_event:
+            room_type = room_create_event.content.get("type", "")
```
clokep, line 1144: I kept using the wrong field here so I ended up making a constant...

```suggestion
            room_type = room_create_event.content.get(EventContentFields.ROOM_TYPE, "")
```

t3chguy: Great, thanks!

changelog.d/10435.feature
```diff
@@ -0,0 +1 @@
+Initial support for MSC3288, sending room_type to the identity server for 3pid invites over the /store-invite API.
```
clokep, line 1:
```suggestion
Experimental support for [MSC3288](https://github.com/matrix-org/matrix-doc/pull/3288), sending `room_type` to the identity server for 3pid invites over the `/store-invite` API.
```

synapse/handlers/identity.py
```diff
@@ -868,6 +870,7 @@ async def ask_id_server_for_third_party_invite(
             "sender": inviter_user_id,
             "sender_display_name": inviter_display_name,
             "sender_avatar_url": inviter_avatar_url,
+            "org.matrix.msc3288.room_type": room_type,  # MSC3288
```
clokep, line 873: Not sure if the extra `MSC3288` comment here adds much?

synapse/handlers/room_member.py
```diff
@@ -1138,6 +1143,11 @@ async def _make_and_store_3pid_invite(
         if room_name_event:
             room_name = room_name_event.content.get("name", "")
 
+        room_type = ""
+        room_create_event = room_state.get((EventTypes.Create, ""))
+        if room_create_event:
+            room_type = room_create_event.content.get(EventContentFields.ROOM_TYPE, "")
```
clokep, line 1149: I left an open question on the MSC about what to do if the field doesn't exist.

clokep (commented):
Looks reasonable!

synapse/handlers/identity.py
t3chguy: Probably not, habit of consistency due to not all experimental prefixes including the msc number, e.g the room summary one

erikjohnston:
My understanding is this is blocked on the MSC progressing?

t3chguy:
Given that its an unstable-prefixed field it shouldn't be, MSCs need a POC implementation to be considered and its barely a proof of concept if no one runs it

t3chguy:
Have tested this when running with https://github.com/matrix-org/sydent/pull/375

image
image

synapse/handlers/room_member.py
clokep: From the MSC it says to not include this field if it doesn't exist, so I think we want to default this to `None` if the create event doesn't exist or if the event doesn't have the field.

And then update the identity handler to only send the field if it is none-`None`.

clokep (changes requested):
Changes about including empty fields.

synapse/handlers/room_member.py
```diff
@@ -1237,6 +1242,13 @@ async def _make_and_store_3pid_invite(
         if room_name_event:
             room_name = room_name_event.content.get("name", "")
 
+        room_type = None
+        room_create_event = room_state.get((EventTypes.Create, ""))
+        if room_create_event:
+            room_type = room_create_event.content.get(
+                EventContentFields.ROOM_TYPE, None
```
novocaine, line 1249: for a `dict`, the second arg to `.get` defaults to `None` already

```suggestion
                EventContentFields.ROOM_TYPE
```

t3chguy: Shows how much (little) I do Python :P
I wonder if this is going to trip the linter into wanting it back on L1248

clokep: I find this to be more of a style thing, some people like to be explicit that it will return `None` in the default case. I was actually going to comment and then deleted it. 😄

synapse/handlers/room_member.py
```diff
@@ -1237,6 +1242,13 @@ async def _make_and_store_3pid_invite(
         if room_name_event:
             room_name = room_name_event.content.get("name", "")
 
+        room_type = None
+        room_create_event = room_state.get((EventTypes.Create, ""))
+        if room_create_event:
+            room_type = room_create_event.content.get(
+                EventContentFields.ROOM_TYPE
+            )
```
clokep, line 1250:
```suggestion
            room_type = room_create_event.content.get(EventContentFields.ROOM_TYPE)
```

t3chguy: Thanks

clokep (approved):
LGTM! Unfortunately this code doesn't seem to have any tests, but that's fine.

clokep:
I'm fairly confident this change will be enough for sydent so I'm going to merge this.

PR #10437, opened 2021-07-20
Add type hints to additional servlet functions

clokep:
Should be reviewable commit-by-commit. This adds type hints to `parse_{boolean,integer}`, `parse_{boolean,integer}_from_args`, and `parse_json_{value,object}_from_request`.

From this it fixes a few incorrect type hints.

This also updates all of the related functions to be more generic (`Mapping[bytes, Sequence[bytes]]` instead of `Dict[bytes, List[bytes]]`).


synapse/rest/client/v1/room.py
```diff
@@ -745,7 +745,7 @@ async def on_GET(self, request):
             if server:
                 raise e
 
-        limit = parse_integer(request, "limit", 0)
+        limit: Optional[int] = parse_integer(request, "limit", 0)
```
clokep, line 748: We need to specify now that this can be `Optional` since the `parse_integer` call always returns an `int`, which mypy now realizes. We manually set it to `None` below if `limit == 0`.

synapse/http/servlet.py
```diff
@@ -66,36 +66,102 @@ def parse_integer_from_args(args, name, default=None, required=False):
...
 
     Returns:
-        bool|None: A bool value or the default.
+        A bool value or the default.
 
     Raises:
         SynapseError: if the parameter is absent and required, or if the
             parameter is present and not one of "true" or "false".
     """
+    args: Mapping[bytes, Sequence[bytes]] = request.args  # type: ignore
```
babolivier, line 144: Why are we ignoring type checks here? Because Twisted's type on `request.args` differ?

babolivier (approved):

synapse/http/servlet.py
clokep: Yes, Twisted don't have the proper type here (I think it thinks it is `None` since that's what it is initialized to?). We do similar things for the other `parse_foo` functions.

PR #10438, opened 2021-07-20
Add a return type to parse_string.

clokep:
Related to #10437, this adds a return type to `parse_string` which had a lot of fallout of mypy being able to tell types of particular variables.

### DIFF
--- a/synapse/http/servlet.py
+++ b/synapse/http/servlet.py
@@ -172,6 +172,42 @@ def parse_bytes_from_args(
     return default
 
 
+@overload
+def parse_string(
+    request: Request,
+    name: str,
+    default: str,
+    *,
+    allowed_values: Optional[Iterable[str]] = None,
+    encoding: str = "ascii",
+) -> str:
+    ...
+
+
+@overload
+def parse_string(
+    request: Request,
+    name: str,
+    *,
+    required: Literal[True],
+    allowed_values: Optional[Iterable[str]] = None,
+    encoding: str = "ascii",
+) -> str:
+    ...
+
+
+@overload
+def parse_string(
+    request: Request,
+    name: str,
+    *,
+    required: bool = False,
+    allowed_values: Optional[Iterable[str]] = None,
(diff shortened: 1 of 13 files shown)


synapse/rest/client/v2_alpha/keys.py
```diff
@@ -194,7 +194,7 @@ def __init__(self, hs):
     async def on_GET(self, request):
         requester = await self.auth.get_user_by_req(request, allow_guest=True)
 
-        from_token_string = parse_string(request, "from")
+        from_token_string = parse_string(request, "from", required=True)
```
clokep, line 197: This seems to be an actual bugfix.

synapse/rest/media/v1/preview_url_resource.py
```diff
@@ -186,15 +186,11 @@ async def _async_render_OPTIONS(self, request: Request) -> None:
...
-        # This will always be set by the time Twisted calls us.
-        assert request.args is not None
-
         # XXX: if get_user_by_req fails, what should we do in an async render?
         requester = await self.auth.get_user_by_req(request)
-        url = parse_string(request, "url")
-        if b"ts" in request.args:
-            ts = parse_integer(request, "ts")
-        else:
+        url = parse_string(request, "url", required=True)
```
clokep, line 191: This seems to be an actual bugfix.

synapse/http/servlet.py
```diff
@@ -172,14 +172,50 @@ def parse_bytes_from_args(
...
+) -> str:
+    ...
+
+
+@overload
+def parse_string(
+    request: Request,
+    name: str,
+    *,
+    required: bool = False,
```
babolivier, line 204: Why not `Literal[False]`?

babolivier (approved):

synapse/http/servlet.py
clokep: Because it doesn't matter if it is provided or not, the default is `False`.

Issue #5961, opened 2019-09-02
Lots of psql duplicate key errors for `device_uniqueness` and `users_name_key`


richvdh:
We get thousands of these a day:

```
2019-09-02 00:00:03.009 UTC [matrix main] ERROR:  duplicate key value violates unique constraint "device_uniqueness"
2019-09-02 00:00:04.622 UTC [matrix main] ERROR:  duplicate key value violates unique constraint "users_name_key"
```

They are due (I think) to our poor man's upsert impl, which we should replace with a proper upsert.

Valodim:
Just a heads up for folks who find this issue because they are facing weird unique constraint violation errors: I've had similar errors after an upgrade to debian buster, due to https://postgresql.verite.pro/blog/2018/08/27/glibc-upgrade.html. This happens when you don't correctly set the collation of all tables to "C" [as requested in the documentation](https://github.com/matrix-org/synapse/blob/master/docs/postgres.rst).

Ezwen:
We have this problem on our homeserver but we did not update to debian buster, any idea?

richvdh:
> Should we be worried?

No. The errors are annoying but not otherwise a problem.

MurzNN:
See in logs same errors when configuring application services:
```
2021-07-21 18:15:53.321 MSK [2739047] matrix_synapse@matrix_synapse ERROR:  duplicate key value violates unique constraint "users_name_key"
2021-07-21 18:15:53.321 MSK [2739047] matrix_synapse@matrix_synapse DETAIL:  Key (name)=(@facebook:homeserver.org) already exists.
2021-07-21 18:15:53.321 MSK [2739047] matrix_synapse@matrix_synapse STATEMENT:  INSERT INTO users (name, password_hash, creation_ts, is_guest, appservice_id, admin, user_type, shadow_banned) V
ALUES('@facebook:homeserver.org', '', 1626880553, 0, 'mautrix_facebook', 0, NULL, false)
```

PR #10442, opened 2021-07-21
Replace `or_ignore` in `simple_insert` with `simple_upsert`

erikjohnston:
Now that we have `simple_upsert` that should be used in preference to
trying to insert and looking for an exception. The main benefit is that
we ERROR message don't get written to postgres logs.

We also have tidy up the return value on `simple_upsert`, rather than having a tri-state of inserted/not-inserted/unknown.

### DIFF
--- a/synapse/storage/database.py
+++ b/synapse/storage/database.py
@@ -930,7 +930,7 @@ class DatabasePool:
         insertion_values: Optional[Dict[str, Any]] = None,
         desc: str = "simple_upsert",
         lock: bool = True,
-    ) -> Optional[bool]:
+    ) -> bool:
         """
 
         `lock` should generally be set to True (the default), but can be set
@@ -951,8 +951,8 @@ class DatabasePool:
             desc: description of the transaction, for logging and metrics
             lock: True to lock the table when doing the upsert.
         Returns:
-            Native upserts always return None. Emulated upserts return True if a
-            new entry was created, False if an existing one was updated.
+            Returns True if a row was inserted or updated (i.e. if `values` is
+            not empty then this always returns True)
         """
         insertion_values = insertion_values or {}
 
@@ -995,7 +995,7 @@ class DatabasePool:
         values: Dict[str, Any],
         insertion_values: Optional[Dict[str, Any]] = None,
         lock: bool = True,
-    ) -> Optional[bool]:
+    ) -> bool:
         """
         Pick the UPSERT method which works best on the platform. Either the
         native one (Pg9.5+, recent SQLites), or fall back to an emulated method.
@@ -1008,16 +1008,15 @@ class DatabasePool:
             insertion_values: additional key/values to use only when inserting
             lock: True to lock the table when doing the upsert.
         Returns:
-            Native upserts always return None. Emulated upserts return True if a
-            new entry was created, False if an existing one was updated.
+            Returns True if a row was inserted or updated (i.e. if `values` is
+            not empty then this always returns True)
         """
         insertion_values = insertion_values or {}
 
         if self.engine.can_native_upsert and table not in self._unsafe_to_upsert_tables:
-            self.simple_upsert_txn_native_upsert(
+            return self.simple_upsert_txn_native_upsert(
                 txn, table, keyvalues, values, insertion_values=insertion_values
             )
-            return None
         else:
             return self.simple_upsert_txn_emulated(
                 txn,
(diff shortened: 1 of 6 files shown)


synapse/storage/databases/main/monthly_active_users.py
```diff
@@ -307,7 +307,7 @@ def upsert_monthly_active_user_txn(self, txn, user_id):
         # never be a big table and alternative approaches (batching multiple
         # upserts into a single txn) introduced a lot of extra complexity.
         # See https://github.com/matrix-org/synapse/issues/3854 for more
-        is_insert = self.db_pool.simple_upsert_txn(
+        self.db_pool.simple_upsert_txn(
```
clokep, line 310: I think the docstring needs to be updated with this change.

synapse/storage/database.py
```diff
@@ -832,31 +832,16 @@ async def simple_insert(
         self,
         table: str,
         values: Dict[str, Any],
-        or_ignore: bool = False,
         desc: str = "simple_insert",
-    ) -> bool:
+    ):
```
clokep, line 836: Can we keep a type hint here? I think it returns `None` now?

synapse/storage/databases/main/user_directory.py
```diff
@@ -388,67 +388,23 @@ def _update_profile_in_user_dir_txn(txn):
...
-                            user_id,
-                            get_localpart_from_id(user_id),
-                            get_domain_from_id(user_id),
-                            display_name,
-                        ),
-                    )
-                else:
-                    # TODO: Remove this code after we've bumped the minimum version
-                    # of postgres to always support upserts, so we can get rid of
-                    # `new_entry` usage
```
clokep, line 412: I'm trusting that our minimum now supports upserts, I haven't double checked that.

erikjohnston: We now always return true for postgres \o/

https://github.com/matrix-org/synapse/blob/329ef5c715d81b538e8b071de046c698a82eae10/synapse/storage/engines/postgres.py#L125-L129

synapse/storage/databases/main/transactions.py
```diff
@@ -134,16 +134,17 @@ async def set_received_txn_response(
...
 
-        await self.db_pool.simple_insert(
+        await self.db_pool.simple_upsert(
             table="received_transactions",
-            values={
+            keyvalues={
                 "transaction_id": transaction_id,
                 "origin": origin,
+            },
+            values={
```
clokep, line 143: Should this be `values` or `insertion_values`? It seems like `insertion_values` would match the previous behavior better?

synapse/storage/databases/main/devices.py
```diff
@@ -1078,16 +1078,18 @@ async def store_device(
...
+                },
+                values={},
+                insertion_values={
                     "display_name": initial_device_display_name,
                     "hidden": False,
                 },
                 desc="store_device",
-                or_ignore=True,
             )
             if not inserted:
```
clokep, line 1094: This block will now never run for postgresql, this seems like a change in behavior, but I'm not sure if it matters or not.

synapse/storage/database.py
```diff
@@ -1111,7 +1094,7 @@ def simple_upsert_txn_native_upsert(
         keyvalues: Dict[str, Any],
         values: Dict[str, Any],
         insertion_values: Optional[Dict[str, Any]] = None,
-    ) -> None:
+    ) -> bool:
         """
         Use the native UPSERT functionality in recent PostgreSQL versions.
```
clokep, line 1099: While we're here?

```suggestion
        Use the native UPSERT functionality in PostgreSQL.
```

clokep (commented):
Overall looks fine! I'm a bit nervous we're changing subtle behavior from the callers, but I'm fairly certain they're fine.

synapse/storage/databases/main/devices.py
erikjohnston: It should? The idea is that `simple_upsert` will return False if we didn't insert or update a new row, and since `values={}` we'll not update rows and so we'll hit this condition if we didn't insert a new entry?

synapse/storage/databases/main/transactions.py
erikjohnston:
```suggestion
            values={},
            insertion_values={
```

synapse/storage/databases/main/devices.py
clokep: I was thinking `inserted` would also be true, but yeah I see if `values` is `{}` then this isn't true.

clokep (approved):

richvdh:
does this fix https://github.com/matrix-org/synapse/issues/5961 (or at least part of it)?

PR #10444, opened 2021-07-21
Fix tests-done github actions status

richvdh:
Set the build to failed if any of the other jobs haven't succeeded. (previously it would be 'skipped' if they failed, which was apparently enough to make github's auto-merge happy.)


erikjohnston (approved):
Suuuuuuure

PR #10447, opened 2021-07-21
Improve failover logic for MSC3083 restricted rooms

clokep:
Based on the error handling in [MSC3083](https://github.com/matrix-org/matrix-doc/pull/3083).

~~Based on #10254.~~

This improves the failover logic for `/make_join` and `/send_join` such that errors of type `M_UNABLE_TO_AUTHORISE_JOIN` and `M_UNABLE_TO_GRANT_JOIN` cause trying another server instead of erroring back to the client.

See complement tests: matrix-org/complement#172.

### DIFF
--- a/synapse/federation/federation_client.py
+++ b/synapse/federation/federation_client.py
@@ -22,6 +22,7 @@ from typing import (
     Awaitable,
     Callable,
     Collection,
+    Container,
     Dict,
     Iterable,
     List,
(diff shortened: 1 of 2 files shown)


synapse/federation/federation_client.py
```diff
@@ -533,6 +535,9 @@ async def _try_destination_list(
                 next server tried. Normally the stacktrace is logged but this is
                 suppressed if the exception is an InvalidResponseError.
 
+            failover_errcodes: Error codes (specific to this endpoint) which should
+                cause a failover.
```
richvdh, line 539: we should probably document that they only cause a failover when in companionship with a 400 HTTP error.

richvdh (approved):
lgtm otherwise

synapse/federation/federation_client.py
clokep: Good call. Thanks!

Issue #10434, opened 2021-07-20
Imrpove format of synapse_review_recent_signups output to easier parsing as JSON or XML


MurzNN:
**Description:**
The recently added `synapse_review_recent_signups` is great, but my homeserver I have use bridges, so they constantly creates a lot of virtual accounts such as `@_telegram_123456:example.com` and `@_slackpuppet___t02_f_l_v1_a_n=2d_u027_h616_b41:example.com`, that shown as long list in output of `synapse_review_recent_signups` script, and I can't find the real registered users between that accounts.

And the output format is not so easy to parse, for example, via `grep`, `jq`, or any other cli tool:
```
# Created: 2021-07-13 18:55:08
# Email: None
# IPs: 
# Number joined public rooms: 0 
# Number joined private rooms: 1
#
@_slackpuppet___t02_f_l_v1_h_n=2d_u028_a9_e2_k_p_b:example.com

# Created: 2021-07-13 19:47:54
# Email: None
# IPs: 
# Number joined public rooms: 0 
# Number joined private rooms: 0
#
@_slackpuppet___t02_f_l_v1_h_n=2d_u027_h616_b47:example.com

# Created: 2021-07-13 20:17:02
# Email: None
# IPs: 
# Number joined public rooms: 0 
# Number joined private rooms: 1
#
@_telegram_1234567:example.com
```

To solve this problem will be good to add some parseable formatters to output, for example - JSON or XML format.

Alternative can be adding support for filtering account name by regex into the script, but this will add more complexity to it and anyway don't give all flexibility like custom output parsers.

anoadragon453:
Generally bridges tend to place a `_` directly after the `@` in their matrix ID in order to differentiate bridged users from native Matrix users -- though this is not a requirement.

Agreed on the formatting, though I believe this script was initially designed solely for admins to use to check for spam bots, rather than something to be consumed by a further processing pipeline. For that purpose it may be a better idea to extend the [List Users Admin API endpoint](https://matrix-org.github.io/synapse/latest/admin_api/user_admin_api.html#list-accounts) to return (and allow sorting by) user creation time.

babolivier:
> For that purpose it may be a better idea to extend the List Users Admin API endpoint to return (and allow sorting by) user creation time.

Wouldn't it be easier to add the appservice_id to the admin API and allow filtering on that?

MurzNN:
[...]

PR #10448, opened 2021-07-21
Add `creation_ts` to list users admin API

dklimpel:
### Pull Request Checklist

### DIFF
--- a/docs/admin_api/user_admin_api.md
+++ b/docs/admin_api/user_admin_api.md
@@ -145,7 +145,7 @@ A response body like the following is returned:
             "shadow_banned": 0,
             "displayname": "<User One>",
             "avatar_url": null,
-            "creation_ts": 1560432668
+            "creation_ts": 1560432668000
         }, {
             "name": "<user_id2>",
             "is_guest": 0,
@@ -155,7 +155,7 @@ A response body like the following is returned:
             "shadow_banned": 0,
             "displayname": "<User Two>",
             "avatar_url": "<avatar_url>",
-            "creation_ts": 1561550621
+            "creation_ts": 1561550621000
         }
     ],
     "next_token": "100",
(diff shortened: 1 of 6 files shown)


babolivier:
I've removed the link to https://github.com/matrix-org/synapse/issues/10434 since it's not clear that this is the right fix for it (we can always close it separately later).

tests/rest/admin/test_user.py
```diff
@@ -781,7 +781,10 @@ def test_order_by(self):
         Testing order list with parameter `order_by`
         """
 
+        # make sure that the users do not have the same timestamps
+        self.pump(1.0)
         user1 = self.register_user("user1", "pass1", admin=False, displayname="Name Z")
+        self.pump(1.0)
```
babolivier, line 787: You probably want to use `self.reactor.advance` here instead

docs/admin_api/user_admin_api.md
```diff
@@ -197,11 +199,12 @@ The following parameters should be set in the URL:
   - `shadow_banned` - Users are ordered by `shadow_banned` status.
   - `displayname` - Users are ordered alphabetically by `displayname`.
   - `avatar_url` - Users are ordered alphabetically by avatar URL.
+  - `creation_ts` - Users are ordered by when the users was created in s (Unix timestamp).
```
babolivier, line 202: Could you make it return a value in milliseconds instead of seconds please? The `users` table is an old one and as such as a few inconsistencies with the rest of Synapse, one of them being using seconds instead of milliseconds. The rest of Synapse uses milliseconds, so I think it'd be nice if we could try to head towards consistency here.

babolivier (changes requested):
Looks broadly good to me, a few changes needed though

docs/admin_api/user_admin_api.md
dklimpel: Yes I can do that, but then there is a mismatch to [Query user API](https://github.com/matrix-org/synapse/blob/develop/docs/admin_api/user_admin_api.md#query-user-account)
What do we do with the Query user API? Also change to `ms`? With a version increment?

I would like to prefer that both APIs returns the same values.

babolivier: Ah yeah somehow I've missed that the query user API also returns a value in seconds. Ideally I'd like it to change to ms too (with a v3 of the endpoint in order not to confuse users). I'd suggest doing that in a separate PR though.

dklimpel: I do not like a `v3` for this small change.
But we can start here to use `ms`.

babolivier: Yes definitely. And agreed that v3 is maybe a bit overkill for such a small change, but I also don't think we should change the unit in the v2 endpoint, lest we break existing scripts using it.

babolivier (approved):
Thanks!

PR #10449, opened 2021-07-21
Support MSC3289: Room version 8

clokep:
This adds support for [MSC3289](https://github.com/matrix-org/matrix-doc/pull/3289): room version 8. This is room version 7 + [MSC3083](https://github.com/matrix-org/matrix-doc/pull/3083).

~~Based on #10447. (Hence the draft.)~~

This should not be merged until MSC3289 is merged.

Also see matrix-org/sytest#1100 and matrix-org/complement#181.

### DIFF
--- a/synapse/api/room_versions.py
+++ b/synapse/api/room_versions.py
@@ -276,12 +276,12 @@ MSC3244_CAPABILITIES = {
     for cap in (
         RoomVersionCapability(
             "knock",
-            RoomVersions.V8,
+            RoomVersions.V7,
             lambda room_version: room_version.msc2403_knocking,
         ),
         RoomVersionCapability(
             "restricted",
-            RoomVersions.V8,
+            None,
             lambda room_version: room_version.msc3083_join_rules,
         ),
     )
--- a/synapse/api/room_versions.py
+++ b/synapse/api/room_versions.py
@@ -193,7 +193,7 @@ class RoomVersions:
     )
     MSC2716 = RoomVersion(
         "org.matrix.msc2716",
-        RoomDisposition.STABLE,
+        RoomDisposition.UNSTABLE,
         EventFormatVersions.V3,
         StateResolutionVersions.V2,
         enforce_key_validity=True,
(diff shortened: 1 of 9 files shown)


synapse/api/room_versions.py
```diff
@@ -234,12 +248,12 @@ class RoomVersionCapability:
...
-            RoomVersions.V7,
+            RoomVersions.V8,
             lambda room_version: room_version.msc2403_knocking,
         ),
         RoomVersionCapability(
             "restricted",
-            None,
+            RoomVersions.V8,
             lambda room_version: room_version.msc3083_join_rules,
         ),
```
t3chguy, line 258: this probably wants to be done at in a subsequent release otherwise clients will advertise this new feature to users with high likelihood that an upgrade to v8 will block existing members from joining the upgraded room until a decent chunk of Synapses support v8

synapse/api/room_versions.py
```diff
@@ -193,6 +193,19 @@ class RoomVersions:
         msc3083_join_rules=False,
         msc2403_knocking=True,
     )
+    V8 = RoomVersion(
```
clokep, line 196: Or maybe we should just update the `MSC3083` room version?

synapse/api/room_versions.py
clokep: Hmm...yeah I guess that makes sense. I guess there isn't a better way to do this, it is similar to updating the default version but we can probably be a bit more aggressive.

synapse/api/room_versions.py
clokep: Seems like for both room version 6 & 7 we removed the experimental ones at the same time.

clokep:
Requesting review on this since it is just pending the MSC.

erikjohnston (approved):

callahad:
@clokep Please retarget against release-v1.40

clokep:
> @clokep Please retarget against release-v1.40

MSC3289 will not merge for 3 more days, so unless we're going to do another RC after that I don't see why we would re-target this. I was assuming this would be in 1.41.

babolivier:
The current plan is to make an rc3 of 1.40 with this PR on Monday (since the MSC will have merged by then) and then release 1.40.0 on Tuesday.

clokep:
I cheekily snuck a commit in here to mark the `org.matrix.msc2716` room version added in #10245 as **unstable**.

PR #10450, opened 2021-07-21
improve typing annotations in CachedCall

richvdh:
tighten up some of the typing in CachedCall, which is going to be needed when
Twisted 21.7 brings better typing on Deferred.

### DIFF
--- a/synapse/util/caches/cached_call.py
+++ b/synapse/util/caches/cached_call.py
@@ -11,15 +11,20 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
+import enum
 from typing import Awaitable, Callable, Generic, Optional, TypeVar, Union
 
+from twisted.internet import defer
 from twisted.internet.defer import Deferred
 from twisted.python.failure import Failure
 
 from synapse.logging.context import make_deferred_yieldable, run_in_background
 
 TV = TypeVar("TV")
+
+
+class _Sentinel(enum.Enum):
+    sentinel = object()
 
 
 class CachedCall(Generic[TV]):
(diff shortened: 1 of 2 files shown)


synapse/util/caches/cached_call.py
```diff
@@ -92,13 +98,19 @@ def got_result(r):
...
-        d.callback(self._result)
-        return await d
+        if isinstance(self._result, _Sentinel):
+            await make_deferred_yieldable(self._deferred)
+            assert not isinstance(self._result, _Sentinel)
+
+        if isinstance(self._result, Failure):
+            # I *think* awaiting a failed Deferred is the easiest way to correctly raise
+            # the right exception.
+            d = defer.fail(self._result)
```
erikjohnston, line 108: Can we not use `Failure.raiseException`?

richvdh: yes :facepalm:

synapse/util/caches/cached_call.py
```diff
@@ -65,7 +69,7 @@ def __init__(self, f: Callable[[], Awaitable[TV]]):
         """
         self._callable: Optional[Callable[[], Awaitable[TV]]] = f
         self._deferred: Optional[Deferred] = None
-        self._result: Union[None, Failure, TV] = None
+        self._result: Union[_Sentinel, TV, Failure] = _Sentinel.sentinel
```
babolivier, line 72: Looks like `_result` is an `Union[_Sentinel, TV, Failure]` but its value is an `object`? Or am I misunderstanding something here?

richvdh: I think you're missing something. Or I don't understand the question.

`_result` starts out as `_Sentinel.sentinel`, which (via the magic of Enums) is an instance of `_Sentinel`. It is later set to either a `TV` or a `Failure`.

clokep (approved):
Looks reasonable to me.

I think some of this could be replaced with non-callback logic, but maybe that's out of scope for improving type annotations. 😄

PR #10451, opened 2021-07-21
Cancel redundant GHA workflows

richvdh:
The idea here is that we only need the most recent `tests` or `release artifacts` workflow for each branch, and all the rest are just sucking up GHA worker time redundantly, so we may as well cancel them.

(docs at https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#concurrency)

### DIFF
--- a/.github/workflows/release-artifacts.yml
+++ b/.github/workflows/release-artifacts.yml
@@ -12,6 +12,10 @@ on:
     # we do the full build on tags.
     tags: ["v*"]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+  
 permissions:
   contents: write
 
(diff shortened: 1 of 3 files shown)


callahad:
Are you sure you're OK cancelling test workflows triggered by pushes to develop? Thinking that might make it harder for us to identify which in a series of pushes caused tests to begin failing...

erikjohnston (approved):
This is what we did for buildkite I believe

richvdh:
> Are you sure you're OK cancelling test workflows triggered by pushes to develop? Thinking that might make it harder for us to identify which in a series of pushes caused tests to begin failing...

Yeah I might have preferred not to do this for develop, but it's such a huge pita to avoid it that I don't think it's worth the cost.

Issue #7164, opened 2020-03-28
Synapse returned event in backfill that was already returned during sync


bwindels:
### Steps to reproduce

- request https://matrix.org/_matrix/client/r0/rooms/!WCHwNGGsrVwPeRtAnq%3Amatrix.org/messages?from=t116230-1314852630_757269739_1037333_466964661_301544009_1144442_46059099_22957226_86880&dir=b&limit=10&filter=%7B%22lazy_load_members%22%3Atrue%7D with an access token that has access to the room "Matrix & ex-Matrix" (!WCHwNGGsrVwPeRtAnq:matrix.org)
- The second event is `$1585350022333PFkGE:lant.uk`,  before `$158535013295770ZmONI:matrix.org` (the 1st event in the backfill)
- This event was also returned during a sync to be between events `$1585352285101062xfUvo:matrix.org` and `$1585352478298tKJwR:lant.uk`

The same event is thus returned to be at 2 locations in the timeline.

The event came from another server, at a time that the federation was lagging considerably.

bwindels:
I don't have the request url for the sync that returned the event initially. I did try to do a new initial sync and the event wasn't there anymore between `$1585352285101062xfUvo:matrix.org` and `$1585352478298tKJwR:lant.uk` and appears indeed before `$158535013295770ZmONI:matrix.org` as the backfill above returns.

Both my riot-web and brawl are showing the event between `$1585352285101062xfUvo:matrix.org` and `$1585352478298tKJwR:lant.uk` after incrementally syncing though, hence my suspicion that it was indeed returned during a sync at that location.

So possibly synapse changed it mind about where the event should appear in the timeline after having synced it to some clients?

richvdh:
This is somewhat expected. Sync returns things in the order they arrive at the server; backfill returns them in the order determined by the event graph.

bwindels:
Hmm, would be good to document this, as it's a pitfall for client implementers. Is this more of a protocol implication or rather a synapse implementation detail?

Also, any idea which location should clients choose for the event? The sync location or the backfill location?

richvdh:
[...]

PR #10464, opened 2021-07-23
Add developer documentation to explain room DAG concepts like `outliers` and `state_groups`

MadLittleMods:
Add developer FAQ to explain room DAG concepts like `outliers` and `state_groups`

 - `outlier` explanation pulled from my understanding working on MSC2716 and various comments throughout the code:
    - [`synapse/handlers/federation.py#L2196-L2199`](https://github.com/matrix-org/synapse/blob/cd5fcd2731182aa48aa7c0a44c7e547681021296/synapse/handlers/federation.py#L2196-L2199)
    - [`synapse/handlers/federation.py#L2731-L2735`](https://github.com/matrix-org/synapse/blob/cd5fcd2731182aa48aa7c0a44c7e547681021296/synapse/handlers/federation.py#L2731-L2735)
     - [`synapse/federation/federation_client.py#L259-L261`](https://github.com/matrix-org/synapse/blob/cd5fcd2731182aa48aa7c0a44c7e547681021296/synapse/federation/federation_client.py#L259-L261)
 - `state_group` explanation pulled from https://github.com/matrix-org/synapse/pull/10245#discussion_r673375767 by @erikjohnston
 - Depth and stream ordering from the following comments:
    - @erikjohnston, https://github.com/matrix-org/synapse/pull/9247#discussion_r588479201
    - @richvdh, https://github.com/matrix-org/synapse/issues/7164#issuecomment-605877176
    - @richvdh, https://github.com/matrix-org/synapse/issues/7164#issuecomment-605953296

### Pull Request Checklist


docs/development/faq.md
```diff
@@ -0,0 +1,15 @@
+# Developer FAQ
+
+## What is an `outlier`?
+
+An `outlier` is an arbitrary floating event in the DAG (as opposed to being
+inline with the current DAG). It also means that we don't have the state events
+backfilled on the homeserver and we trust the events *claimed* auth events rather
+than those we calculate and verify to be correct.
+
```
MadLittleMods, line 9: We can also talk about:

```
An event can be unmarked as an `outlier` once we fetch all of its `prev_events` (you will see some `ex_outlier` code around this)
```

Or

```
After we fetch all of the `prev_events` of the `outlier` we can unmark the event as an outlier (an `ex_outlier`)
```

dklimpel:
Also the summary needs an update / entry.
https://github.com/matrix-org/synapse/blob/develop/docs/SUMMARY.md

MadLittleMods:
Thanks for the shout @dklimpel! Added 👍

docs/development/faq.md
```diff
@@ -0,0 +1,17 @@
+# Developer FAQ
+
+## What is an `outlier`?
+
+An `outlier` is an arbitrary floating event in the DAG (as opposed to being
+inline with the current DAG). It also means that we don't have the state events
+backfilled on the homeserver and we trust the events *claimed* auth events rather
+than those we calculate and verify to be correct. 
+
+An event can be unmarked as an `outlier` once we fetch all of its `prev_events` (you will see some `ex_outlier` code around this).
```
richvdh, line 10: I think it's slightly different. The thing that makes an `outlier` an outlier is that we haven't figured out the state for the room at that point in the DAG. 

We won't *necessarily* have its prev_events in the database, but it's entirely possible that we *will*. Conversely, it's possible for us not to have all the prev_events for non-outlier event (in which case, that event would be a "backward extremity").

The process of transforming an outlier to a non-outlier involves figuring out the room state at that point in the DAG - *either* by having the state at all the `prev_events` and resolving the state between them, or simply by asking another server.

auth events are a whole extra level of complexity, but even for an outlier we should have all the events in the auth chain. https://github.com/matrix-org/synapse/blob/cd5fcd2731182aa48aa7c0a44c7e547681021296/synapse/handlers/federation.py#L2196-L2199 is quite a specific bit of code relating to one part of processing incoming events (and to be honest, I'm not convinced it does the right thing).  I would leave auth events out of this explanation, tbh.

docs/development/faq.md
```diff
@@ -0,0 +1,17 @@
...
+
+
+## What is a `state_group`?
+
+For every non-outlier event we need to know the state at that event. Instead of storing the full state for each event in the DB (i.e. a `event_id -> state` mapping), which is *very* space inefficient when state doesn't change, we instead assign each different set of state a "state group" and then have mappings of `event_id -> state_group` and `state_group -> state`.
+
```
richvdh, line 16: yes, though has anyone explained `state_group_edges` to you yet? (which is a further optimisation...)

could you wrap this text at about 80 characters for ease of reading/commenting?

docs/development/faq.md
```diff
@@ -0,0 +1,17 @@
+# Developer FAQ
```
richvdh, line 1: I'd find this better presented as a document describing "Room DAG concepts" with headings for "Outliers", "State groups", etc, rather than a FAQ, but ymmv.

richvdh (changes requested):
thanks very much for this, @MadLittleMods - we're definitely lacking some useful documentation in this area!

docs/development/faq.md
clokep: I think I would +1 this and place it in the docs hierarchy above https://matrix-org.github.io/synapse/latest/auth_chain_difference_algorithm.html

docs/development/faq.md
MadLittleMods: Thanks for all info @richvdh 😀

These seem so closely coupled and aren't mutually exclusive in real life situations so it's hard for me to make a clear description 🤯

I've updated with something more inline with your description but feel free to rewrite.

---

I feel like I don't understand where an `outlier` becomes an `ex_outlier`.

Say we fetch a missing auth event for an event, we mark that auth event as an `outlier`. Then at what point later, do we get the event again as a non-outlier? From your statement, it's not when we have fetched all of the `prev_events` as those could already be all persisted in the db.

There is [`_update_outliers_txn`](https://github.com/matrix-org/synapse/blob/858363d0b7e58fd71875b25d183537bb3b5a397f/synapse/storage/databases/main/events.py#L1225) which handles `ex_outlier` stuff but it's confusing where/when we will come across the same `outlier` event again while persisting.

Based on the DAG below, as a remote federating server, my assumption is that we process event `D` which fetches the `state_event` as an `outlier`. Then as we backfill more, and process `state_event` directly, it becomes an `ex_outlier`.

[Mermaid live editor playground link](https://mermaid-js.github.io/mermaid-live-editor/edit/#eyJjb2RlIjoiZmxvd2NoYXJ0IEJUXG4gICAgc3ViZ3JhcGggbGl2ZVxuICAgICAgICBEIC0tLS0-IEMgLS0tLT4gQiAtLS0tPiBzdGF0ZV9ldmVudCAtLS0tPiBBXG4gICAgZW5kXG4gICAgXG5cbiAgICBEIC0uLT4gc3RhdGVfZXZlbnRcblxuICAgIFxuICAgICUlIGFsaWdubWVudCBsaW5rcyB0byBtYWtlIHRoZSBncmFwaCBwb2ludCB1cHdhcmRcbiAgICBudWxsWyBdIC0tLSBBXG4gICAgJSUgbWFrZSB0aGUgbGlua3MgYW5kIG5vZGVzIGludmlzaWJsZVxuICAgIHN0eWxlIG51bGwgZGlzcGxheTpub25lO1xuICAgIGxpbmtTdHlsZSA1IHN0cm9rZS13aWR0aDoycHgsZmlsbDpub25lLHN0cm9rZTpub25lO1xuXG4gICAgIiwibWVybWFpZCI6Int9IiwidXBkYXRlRWRpdG9yIjpmYWxzZSwiYXV0b1N5bmMiOnRydWUsInVwZGF0ZURpYWdyYW0iOmZhbHNlfQ)




I'm still a bit wary on a situation where we could have all of the `prev_events` for the `state_event` but still mark it as an `outlier`. Perhaps with gaps in the DAG and the event hits it just right 🤷

docs/development/faq.md
MadLittleMods: I wasn't aware of `state_group_edges` yet 😁

I've added a TODO section to these docs so we can fill it in in a future PR or pop it in if you want to explain now 🤑

docs/development/room-dag-concepts.md
```diff
@@ -0,0 +1,83 @@
...
+incrementing integer but for `backfilled=true` events, it decrements.
+
+`depth` is not re-calculated if a message is inserted into the middle of the DAG.
+
+---
+
+ - `/sync` returns things in the order they arrive at the server (`stream_ordering`).
+ - `/backfill` returns them in the order determined by the event graph `(topological_ordering, stream_ordering)`.
+
+The general idea is that, if you're following a room in real-time (i.e. `/sync`), you probably want to see the messages as they arrive at your server, rather than skipping any that arrived late; whereas if you're looking at a historical section of timeline (i.e. `/messages`), you want to see the best representation of the state of the room as others were seeing it at the time.
```
MadLittleMods, line 29: Lifted this text from @richvdh,

 - https://github.com/matrix-org/synapse/issues/7164#issuecomment-605877176
 - https://github.com/matrix-org/synapse/issues/7164#issuecomment-605953296

docs/development/faq.md
MadLittleMods: Thanks for the shout on a better spot to put this 🏴‍☠️

docs/SUMMARY.md
```diff
@@ -80,6 +80,7 @@
       - [SAML](development/saml.md)
       - [CAS](development/cas.md)
     - [State Resolution]()
+      - [Room DAG concepts](development/room-dag-concepts.md)
```
richvdh, line 83: most of this isn't really part of state resolution, and I don't think I agree with @clokep's recommendation to put it just here. what do you both think?

docs/development/room-dag-concepts.md
```diff
@@ -0,0 +1,83 @@
...
+```
+
+
+## Depth and stream ordering
+
+Events are sorted by `(topological_ordering, stream_ordering)` where
+`topological_ordering` is just `depth`. In other words, we first sort by `depth`
+and then tie-break based on `stream_ordering`. `depth` is incremented as new
+messages are added to the DAG. Normally, `stream_ordering` is an auto
+incrementing integer but for `backfilled=true` events, it decrements.
```
richvdh, line 20:
```suggestion
incrementing integer, but backfilled events start with `stream_ordering=-1` and decrement.
```

docs/development/room-dag-concepts.md
```diff
@@ -0,0 +1,83 @@
...
+
+`depth` is not re-calculated if a message is inserted into the middle of the DAG.
```
richvdh, line 22: `depth` is an inherent property of the event - it's calculated by the originating server as `max(e.depth for e in prev_events)+1`, and it is then covered by the signature (and therefore set in stone).

docs/development/room-dag-concepts.md
richvdh, line 29: wrap at 80 chars please.

docs/development/room-dag-concepts.md
```diff
@@ -0,0 +1,83 @@
...
+
+ - `/sync` returns things in the order they arrive at the server (`stream_ordering`).
+ - `/backfill` returns them in the order determined by the event graph `(topological_ordering, stream_ordering)`.
+
+The general idea is that, if you're following a room in real-time (i.e. `/sync`), you probably want to see the messages as they arrive at your server, rather than skipping any that arrived late; whereas if you're looking at a historical section of timeline (i.e. `/messages`), you want to see the best representation of the state of the room as others were seeing it at the time.
+
+
+## Forward extremity
+
+Most-recent-in-time events in the DAG which are not referenced by any `prev_events` yet.
```
richvdh, line 34:
```suggestion
Most-recent-in-time events in the DAG which are not referenced by any other events' `prev_events` yet.
```

docs/development/room-dag-concepts.md
```diff
@@ -0,0 +1,83 @@
...
+
+The forward extremities of a room are used as the `prev_events` when the next event is sent.
+
+
+## Backwards extremity
+
+The current marker of where we have backfilled up to.
+
+A backwards extremity is a place where the oldest-in-time events of the DAG
```
richvdh, line 43: I think this sentence has got lost and can be removed?

docs/development/room-dag-concepts.md
```diff
@@ -0,0 +1,83 @@
...
+
+This is an event where we haven't fetched all of the `prev_events` for.
+
+Once we have fetched all of it's `prev_events`, it's unmarked as backwards
```
richvdh, line 47:
```suggestion
Once we have fetched all of its `prev_events`, it's unmarked as a backwards
```

docs/development/room-dag-concepts.md
```diff
@@ -0,0 +1,83 @@
...
+extremity and those `prev_events` become the new backwards extremities.
```
richvdh, line 48:
```suggestion
extremity.
```

The `prev_events` probably don't themselves become backwards extremities, for two reasons:

1. normally we backfill in batches (of 20 events or so, iirc?) so, assuming a linear timeline, the new backwards extremity will be 20 events back up the timeline.

2. More subtly, imagine a DAG like this (sorry for terrible ascii art, hopefully you get the idea):

```
A <--  B <-- C <-- E
^                /
 \--- ( D ) <---
```

To start with, `D` is missing, so `E` is a backwards extremity. This isn't that unusual - it can happen whenever there's a fork in the DAG and we can't talk to D's server for some reason. Then we backfill, retrieving `D`. *Neither* of `E`'s prev-events (`C` and `D`) become backwards extremities, because we already have all their prev-events.

docs/development/faq.md
richvdh: @Azrenbeth has written some excellent notes on this at https://pastebin.com/seUGVGeT which we should consider incorporating into this document - but honestly I would focus on landing what you've got rather than expanding it to cover everything. Let's keep it as a TODO for now.

docs/development/faq.md
richvdh: > I've mainly been working with floating outliers so I don't think I had the complete picture.

I think I might have misled you here. From Synapse's point of view, all outliers are considered to be floating, and in the majority of cases we indeed won't have their prev events. However, it's possible for the stars to align in a way that we do actually have their prev events in the database - at least as outliers themselves.

To take a simple example:

Two of the auth events for a `join` event in an invite-only room are the `join_rules` event and an invite event for the joiner. So let's suppose we fetch both of those events as outliers. Let's also suppose that the invite was issued immediately after changing the join rules - so the `join_rules` is the only prev-event of the `invite` event.

```
join_rules <-  invite
   ^             ^
   ............................. join
```

In that case, it so happens that we have all the prev-events of the invite event - but we're still considering it an outlier.

So I don't want you to imagine there is a distinction between "floating outliers" and "regular outliers" - they are all just outliers in terms of how we handle them in Synapse.

> Say we fetch a missing auth event for an event, we mark that auth event as an outlier. Then at what point later, do we get the event again as a non-outlier? From your statement, it's not when we have fetched all of the prev_events as those could already be all persisted in the db.

right. It becomes a non-outlier in the situation where we process it as a regular event as part of the timeline - normally via backfill.

> Based on the DAG below, as a remote federating server, my assumption is that we process event D which fetches the state_event as an outlier. Then as we backfill more, and process state_event directly, it becomes an ex_outlier.

exactly so, yes.

> I'm still a bit wary on a situation where we could have all of the `prev_events` for the `state_event` but still mark it as an `outlier`. Perhaps with gaps in the DAG and the event hits it just right :shrug:

yup. Imagine there is another fork in the DAG pointing to A:

[](https://mermaid-js.github.io/mermaid-live-editor/edit/##eyJjb2RlIjoiZmxvd2NoYXJ0IEJUXG4gICAgJSUgYWxpZ25tZW50IGxpbmtzIHRvIG1ha2UgdGhlIGdyYXBoIHBvaW50IHVwd2FyZFxuICAgIG51bGxbIF0gLS0tIEFcblxuICAgIHN1YmdyYXBoIGxpdmVcbiAgICAgICAgRCAtLS0tPiBDIC0tLS0-IEIgLS0tLT4gc3RhdGVfZXZlbnQgLS0tLT4gQVxuICAgICAgICBaIC0tPiBZIC0tLS0-IEFcbiAgICBlbmRcbiAgICBcblxuICAgIEQgLS4tPiBzdGF0ZV9ldmVudFxuXG4gICAgXG5cbiAgICAlJSBtYWtlIHRoZSBsaW5rcyBhbmQgbm9kZXMgaW52aXNpYmxlXG4gICAgc3R5bGUgbnVsbCBkaXNwbGF5Om5vbmU7XG4gICAgbGlua1N0eWxlIDAgc3Ryb2tlLXdpZHRoOjJweCxmaWxsOm5vbmUsc3Ryb2tlOm5vbmU7XG5cbiAgICAiLCJtZXJtYWlkIjoie30iLCJ1cGRhdGVFZGl0b3IiOmZhbHNlLCJhdXRvU3luYyI6dHJ1ZSwidXBkYXRlRGlhZ3JhbSI6ZmFsc2V9)

When we receive Y, we'll backfill A. So then we have `state_event`'s `prev_events` - but `state_event` is still an outlier at this point.

docs/development/room-dag-concepts.md
MadLittleMods: Woah! I was under the impression that it was a just a field that Synapse calculated to make the slicing and dicing of all events easier.

From the client perspective it doesn't even exist: https://spec.matrix.org/unstable/client-server-api/ 

But I do see some mentions from the server API: https://spec.matrix.org/unstable/server-server-api/ 

Why don't I see when I inspect the original source of an event in Element for example? Where is it stored on the event? Any docs?

---

The reason I mentioned that line in the first place is because during the MSC2716 talks, there was some discussion of doing some major refactoring so that the `depth` is calculated. Maybe this was meant for a future MSC vs a homeserver implementation 😕

> So having rediscussed this IRL: Dendrite (and Synapse) currently get their `depth` parameters used for ordering from the wire. Ideally, we'd calculate the depth parameter instead
>
> https://github.com/matrix-org/matrix-doc/pull/2716#discussion_r559828899

Relates to https://github.com/matrix-org/matrix-doc/issues/1229

docs/SUMMARY.md
MadLittleMods: It's in the same ballpark 🤷. I was thinking of renaming the category overall too but didn't come up with something better.

docs/development/room-dag-concepts.md
MadLittleMods: Good spot! I've combined it with the sentence above ⏩

docs/development/room-dag-concepts.md
MadLittleMods: Thanks for the good context here! I think this would be good to explain. I've added a `(unless we have already persisted them)` qualifying statement to it and also an "in reality" sentence to describe the batched backfills ⏩

docs/development/faq.md
MadLittleMods: Thanks for the clarification here. I've added a paragraph explaining that there is no distinction.

I like these examples and edge cases. I'm tempted to add them into the docs but I think I'll save that for another iteration.

docs/development/faq.md
MadLittleMods: Added the link to the TODO comment ⏩

MadLittleMods:
Thanks for all of the review so far @richvdh. My mind is being blown in multiple ways! 🤯

docs/SUMMARY.md
richvdh: I think just stick it one level up the tree.

docs/development/room-dag-concepts.md
richvdh: > Why don't I see when I inspect the original source of an event in Element for example? 

Well, therein lies a complication. The format of events served up to clients is not quite the same as those that get passed around between homeservers. I opened https://github.com/matrix-org/matrix-doc/issues/3305 to clarify this.



> The reason I mentioned that line in the first place is because during the MSC2716 talks, there was some discussion of doing some major refactoring so that the `depth` is calculated.

Well yes, it would be nice if we didn't have to trust remote homeservers to tell us the depth of an event, either by calculating it ourselves (somehow) or, more likely, avoiding our dependence on it. But that's all a bit sci-fi at the moment, and if you're writing Synapse developer docs, I think it's better to make them reflect current reality than future aspirations.

docs/development/room-dag-concepts.md
```diff
@@ -13,39 +13,43 @@ A (oldest) <---- B <---- C (most recent)
...
-incrementing integer but for `backfilled=true` events, it decrements.
-
-`depth` is not re-calculated if a message is inserted into the middle of the DAG.
+incrementing integer, but backfilled events start with `stream_ordering=-1` and decrement.
 
 ---
 
  - `/sync` returns things in the order they arrive at the server (`stream_ordering`).
- - `/backfill` returns them in the order determined by the event graph `(topological_ordering, stream_ordering)`.
+ - `/messages` (and `/backfill in the federation API) return them in the order determined by the event graph `(topological_ordering, stream_ordering)`.
```
richvdh, line 25:
```suggestion
 - `/messages` (and `/backfill` in the federation API) return them in the order determined by the event graph `(topological_ordering, stream_ordering)`.
```

docs/development/room-dag-concepts.md
```diff
@@ -13,39 +13,43 @@ A (oldest) <---- B <---- C (most recent)
...
 
 This is an event where we haven't fetched all of the `prev_events` for.
 
-Once we have fetched all of it's `prev_events`, it's unmarked as backwards
-extremity and those `prev_events` become the new backwards extremities.
+Once we have fetched all of its `prev_events`, it's unmarked as a backwards
+extremity and those `prev_events` become the new backwards extremities (unless
+we have already persisted them). Also in reality, we backfill in batches of 20
+events or so, so only the `prev_events` of the last oldest-in-time event will
+become the backwards extremeties.
```
richvdh, line 52: I think this is a bit confusing. How about:

```suggestion
extremity (although we may have formed new backwards extremities during the backfilling process).
```

docs/development/room-dag-concepts.md
```diff
@@ -61,12 +65,11 @@ For example, when we fetch the event auth chain or state for a given event, we
 mark all of those claimed auth events as outliers because we haven't done the
 state calculation ourself.
 
+Outliers are sometimes referred to as floating outliers but there is no
```
richvdh, line 68: I've never heard them referred to as that...

MadLittleMods: Maybe it was just a descriptor that came out of the intro discussions for discussing how MSC2716 could work. Can remove ⏩

docs/development/room-dag-concepts.md
MadLittleMods: 👍 I tweaked it slightly to include the prev event chain,

> Once we have fetched all of its `prev_events`, it's unmarked as a backwards
extremity (although we may have formed new backwards extremities from the prev
events during the backfilling process).

richvdh (approved):
lgtm!

richvdh:
thanks for your perseverence, @MadLittleMods!

MadLittleMods:
Thank you for the meaty review @richvdh to get this factually correct! Lots of great tidbits shared 🦌

PR #10465, opened 2021-07-23
Change release script to update debian changelog for RCs

erikjohnston:
We used to only do debs for full releases, but we're now doing them for RCs too, so we need to update the changelog.

### DIFF
--- a/scripts-dev/release.py
+++ b/scripts-dev/release.py
@@ -190,12 +190,24 @@ def run():
     # Generate changelogs
     subprocess.run("python3 -m towncrier", shell=True)
 
-    # Generate debian changelogs if its not an RC.
-    if not rc:
-        subprocess.run(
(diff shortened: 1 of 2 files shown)


scripts-dev/release.py
```diff
@@ -190,12 +190,24 @@ def run():
...
 
-    # Generate debian changelogs if its not an RC.
-    if not rc:
-        subprocess.run(
-            f'dch -M -v {new_version} "New synapse release {new_version}."', shell=True
-        )
-        subprocess.run('dch -M -r -D stable ""', shell=True)
+    # Generate debian changelogs
+    assert not parsed_new_version.is_devrelease
+    assert not parsed_new_version.is_postrelease
```
babolivier, line 195: Looks like these checks should be higher up, since we also probably don't want to go through all of the previous steps if the release isn't either an RC or a stable release.

babolivier (approved):
lgtm modulo CI complaints

Issue #10469, opened 2021-07-23
Port the remaining module interfaces to the new generic one


babolivier:
* [ ] Auth provider
* [ ] Storage provider
* [ ] OIDC user mapping provider
* [ ] SAML2 user mapping provider
* [ ] Presence router

richvdh:
The initial intention (as noted in #10062) was to start removing the old interfaces in August 2021 - but I think it might be less confusing to remove them all at once.

babolivier:
Agreed.

PR #10472, opened 2021-07-23
Deprecate `additional_resources`

babolivier:


docs/upgrade.md
```diff
@@ -88,6 +88,15 @@ process, for example:
 
 # Upgrading to v1.39.0
 
+## Deprecation of `additional_resources`
+
+The `additional_resources` section of the listener configuration is deprecated in favour
+of the new modules system introduced in Synapse v1.37.0. Authors of modules using this
```
anoadragon453, line 94: It may be worth mentioning that modules can migrate away from `additional_resources` without needing to use the new modules system? For instance, if I were making use of `additional_resources` for a `PresenceRouter` module, I shouldn't think that I need to convert my module first (which currently isn't possible).

anoadragon453 (changes requested):
lgtm otherwise

docs/upgrade.md
babolivier: > It may be worth mentioning that modules can migrate away from `additional_resources` without needing to use the new modules system?

They do need to use `register_web_resource` which is part of the new system though. Though yeah might be worth rewording so maintainers don't think they need to port their entire module first.

babolivier:
Closing this PR since, as mentioned in https://github.com/matrix-org/synapse/issues/10469#issuecomment-888213232, we'll be removing all legacy interfaces at once.

PR #10481, opened 2021-07-26
Fix debian package triggers

richvdh:
Replace the outdated list of dpkg triggers with an autogenerated one.

### DIFF
--- a/debian/build_virtualenv
+++ b/debian/build_virtualenv
@@ -100,3 +100,18 @@ esac
 # add a dependency on the right version of python to substvars.
 PYPKG=`basename $SNAKE`
 echo "synapse:pydepends=$PYPKG" >> debian/matrix-synapse-py3.substvars
+
+
+# add a couple of triggers.  This is needed so that dh-virtualenv can rebuild
+# the venv when the system python changes (see
+# https://dh-virtualenv.readthedocs.io/en/latest/tutorial.html?highlight=triggers#step-2-set-up-packaging-for-your-project)
+#
+# we do it here rather then the more conventional way of just adding it to
+# debian/matrix-synapse-py3.triggers, because we need to add a trigger on the
+# right version of python.
+cat >>"debian/.debhelper/generated/matrix-synapse-py3/triggers" <<EOF
+# triggers for dh-virtualenv
+interest-noawait $SNAKE
+interest dh-virtualenv-interpreter-update
+
+EOF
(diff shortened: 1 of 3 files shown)


debian/build_virtualenv
```diff
@@ -100,3 +100,18 @@ esac
 # add a dependency on the right version of python to substvars.
 PYPKG=`basename $SNAKE`
 echo "synapse:pydepends=$PYPKG" >> debian/matrix-synapse-py3.substvars
+
+
+# add a couple of triggers.  This is needed so that dh-virtualenv can rebuild
+# the venv when the system python changes (see
+# https://dh-virtualenv.readthedocs.io/en/latest/tutorial.html?highlight=triggers#step-2-set-up-packaging-for-your-project)
+#
+# we do it here rather then the more conventional way of just adding it to
```
babolivier, line 109:
```suggestion
# we do it here rather than the more conventional way of just adding it to
```

debian/build_virtualenv
```diff
@@ -100,3 +100,18 @@ esac
 # add a dependency on the right version of python to substvars.
 PYPKG=`basename $SNAKE`
 echo "synapse:pydepends=$PYPKG" >> debian/matrix-synapse-py3.substvars
+
+
+# add a couple of triggers.  This is needed so that dh-virtualenv can rebuild
+# the venv when the system python changes (see
+# https://dh-virtualenv.readthedocs.io/en/latest/tutorial.html?highlight=triggers#step-2-set-up-packaging-for-your-project)
```
babolivier, line 107: Is the highlight on "trigger" intentional?

debian/changelog
```diff
@@ -1,6 +1,8 @@
 matrix-synapse-py3 (1.39.0ubuntu1) UNRELEASED; urgency=medium
 
   * Drop backwards-compatibility code that was required to support Ubuntu Xenial.
+  * Update package triggers so that the virtualenv is correctly rebuild
+    when the system python is rebuilt, on recent Python versions.
```
babolivier, line 5: It feels a bit weird to me that this change is listed under 1.39.0 but the PR's target branch isn't release-1.39

debian/changelog
```diff
@@ -1,6 +1,8 @@
 matrix-synapse-py3 (1.39.0ubuntu1) UNRELEASED; urgency=medium
 
   * Drop backwards-compatibility code that was required to support Ubuntu Xenial.
+  * Update package triggers so that the virtualenv is correctly rebuild
```
babolivier, line 4:
```suggestion
  * Update package triggers so that the virtualenv is correctly rebuilt
```

babolivier (changes requested):

debian/changelog
richvdh: in the debian changelog, the version number is pretty meaningless while it is UNRELEASED. when we go through the release process, whatever is UNRELEASED gets turned into the version we're releasing.

erikjohnston (approved):

Issue #10484, opened 2021-07-27
Synapse does not tell the clients when it gets out of OTK


bmarty:
Following discussion in #10479 (and especially https://github.com/matrix-org/synapse/issues/10479#issuecomment-887001086), I've checked the sync response when Synapse does not have OTK anymore. So when the number of OTK is changing from N to 0.

Instead of receiving 

```json
"device_one_time_keys_count":{"signed_curve25519":0}
```

The client receives

```json
"device_one_time_keys_count":{}
```

and so the client cannot distinguish "no change" and "no OTK left", and do not upload new OTK, which at the end breaks decryption of some messages.

richvdh:
It seems like this needs clarification in the spec too. Is a server obliged to know about all potential one-time-key algorithms so that it can fully populate this field? What happens if/when we introduce a new algorithm?

Maybe `"device_one_time_keys_count":{}` should mean "I don't have any one-time keys", and the complete absence of the field should mean "no change".

anoadragon453:
@richvdh I too ran into this line of thinking, though the possible algorithm names are specced.

> Maybe "device_one_time_keys_count":{} should mean "I don't have any one-time keys", and the complete absence of the field should mean "no change".

What if there is no change in algorithm A, but there is a change in algorithm B? You can't have the absence of `"device_one_time_keys_count"` in that scenario.

richvdh:
> @richvdh I too ran into this line of thinking, though the possible algorithm names are specced.

well they are now, but again: what if we introduce a new algorithm? Clients using the new algorithm will be back where we are now: unable to tell if the server is out of keys or if there is no change.

> What if there is no change in algorithm A, but there is a change in algorithm B? You can't have the absence of `"device_one_time_keys_count"` in that scenario.

I was thinking you'd populate the whole field anyway, including the (redundant) count of algorithm A. But you're right, it's not obvious. We should probably take this over to matrix-doc.

uhoreg:
The spec has been [updated](https://github.com/matrix-org/matrix-doc/pull/3636) to clarify that a missing algorithm means zero one-time keys are available.

PR #10485, opened 2021-07-27
Always communicate device OTK counts to clients

anoadragon453:
Fixes https://github.com/matrix-org/synapse/issues/10484.

As far as I can tell from [the spec](https://spec.matrix.org/unstable/client-server-api/#device-keys), `signed_curve25519` is the only key algorithm employed for one-time keys at the moment. Thus this PR will by default return `{"signed_curve25519": 0}` for key counts. Only if there are any keys available will this count be something other than 0.

This affects both [`/sync` responses](https://matrix.org/docs/spec/client_server/r0.6.1#get-matrix-client-r0-sync), as well as the response to [`POST /_matrix/client/r0/keys/upload`](https://matrix.org/docs/spec/client_server/r0.6.1#post-matrix-client-r0-keys-upload) requests.

This isn't a complete solution to the problem (hence the included TODO) as we now over-communicates to the client rather than omit the information if there have been no changes. However, that's wasting a few bytes, whereas before this code could cause encryption issues.

### DIFF
--- a/changelog.d/10485.bugfix
+++ b/changelog.d/10485.bugfix
@@ -1,1 +1,1 @@
-Fix a long-standing bug where Synapse would signal "no change" in device one-time key counts, whereas in reality the number of keys had been exhausted.
+Fix a long-standing bug where Synapse would not inform clients that a device had exhausted its one-time-key pool, potentially causing problems decrypting events.
--- a/changelog.d/10485.bugfix
+++ b/changelog.d/10485.bugfix
@@ -1,0 +1,1 @@
+Fix a long-standing bug where Synapse would signal "no change" in device one-time key counts, whereas in reality the number of keys had been exhausted.
--- a/synapse/handlers/sync.py
+++ b/synapse/handlers/sync.py
@@ -1093,9 +1093,10 @@ class SyncHandler:
         one_time_key_counts: JsonDict = {}
         unused_fallback_key_types: List[str] = []
         if device_id:
-            # TODO: If this is an incremental sync and there's been no change in a count
-            # since the provided since token, then simply omit that algorithm's entry
-            # altogether.
+            # TODO: We should have a way to let clients differentiate between the states of:
+            #   * no change in OTK count since the provided since token
+            #   * the server has zero OTKs left for this device
+            #  Spec issue: https://github.com/matrix-org/matrix-doc/issues/3298
             one_time_key_counts = await self.store.count_e2e_one_time_keys(
                 user_id, device_id
             )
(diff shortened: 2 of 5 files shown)


changelog.d/10485.bugfix
```diff
@@ -0,0 +1 @@
+Fix a long-standing bug where Synapse would signal "no change" in device one-time key counts, whereas in reality the number of keys had been exhausted.
```
anoadragon453, line 1: Feedback on whether this is digestible enough is welcome.

babolivier: Maybe

> Fix a long-standing bug where Synapse would mistakenly tell clients one-time key counts hadn't changed instead of telling them no key is available.

Would be easier to read?

babolivier (approved):
lgtm otherwise

changelog.d/10485.bugfix
richvdh: I think it would be nice to express this in terms of user-visible impact. 
```suggestion
Fix a long-standing bug where Synapse would not inform clients that a device had exhausted its one-time-key pool, potentially causing problems decrypting events.
```

babolivier: Ah yeah that sounds better

anoadragon453: Thanks both!

synapse/handlers/sync.py
```diff
@@ -1093,6 +1093,9 @@ async def generate_sync_result(
         one_time_key_counts: JsonDict = {}
         unused_fallback_key_types: List[str] = []
         if device_id:
+            # TODO: If this is an incremental sync and there's been no change in a count
+            # since the provided since token, then simply omit that algorithm's entry
+            # altogether.
```
richvdh, line 1098: I'm not 100% convinced this is a thing we want to do, per https://github.com/matrix-org/synapse/issues/10484#issuecomment-887415854.

anoadragon453: Thanks for calling this out. Shall I remove this TODO for now as we design a more robust signalling solution on the spec side?

richvdh: I dunno, maybe soften the phrasing a bit: "consider omitting the entry altogether". possibly also link to the discussion in #10484.

richvdh: (or the spec issue)

anoadragon453: Out of that discussion came an issue on matrix-doc. I've changed the comment here to link to it in the meantime.

anoadragon453:
The unit tests needed to be updated in order to expect `signed_curve25519` in each response (though this is a little bit meh feeling).

Complement is expected to fail restricted room tests as we're based on the `release-v1.39` branch.

babolivier (approved):

richvdh (approved):
lgtm

PR #10488, opened 2021-07-27
Use the new test running syntax for complement.

clokep:
CC @anoadragon453 

This is to match matrix-org/complement#171.


anoadragon453 (approved):
Thanks!

PR #10491, opened 2021-07-28
Generics for `ObservableDeferred`

richvdh:
Now that `Deferred` is a generic class, let's update `ObservableDeferred` to follow suit.

Based on #10490.

### DIFF
--- a/synapse/storage/persist_events.py
+++ b/synapse/storage/persist_events.py
@@ -170,7 +170,7 @@ class _EventPeristenceQueue(Generic[_PersistResult]):
             end_item = queue[-1]
         else:
             # need to make a new queue item
-            deferred = ObservableDeferred[_PersistResult](
+            deferred: ObservableDeferred[_PersistResult] = ObservableDeferred(
                 defer.Deferred(), consumeErrors=True
             )
 
--- a/synapse/storage/persist_events.py
+++ b/synapse/storage/persist_events.py
@@ -170,7 +170,9 @@ class _EventPeristenceQueue(Generic[_PersistResult]):
             end_item = queue[-1]
         else:
             # need to make a new queue item
-            deferred = ObservableDeferred(defer.Deferred(), consumeErrors=True)
+            deferred: ObservableDeferred[_PersistResult] = ObservableDeferred(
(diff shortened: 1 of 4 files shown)


synapse/storage/persist_events.py
```diff
@@ -170,7 +170,9 @@ async def add_to_queue(
             end_item = queue[-1]
         else:
             # need to make a new queue item
-            deferred = ObservableDeferred(defer.Deferred(), consumeErrors=True)
+            deferred: ObservableDeferred[_PersistResult] = ObservableDeferred(
```
erikjohnston, line 173:
```suggestion
            deferred = ObservableDeferred[_PersistResult](
```

Is that a thing we can do now?

erikjohnston (approved):

synapse/storage/persist_events.py
ShadowJonathan: First instinct tells me no (though not sure)

richvdh: It seems we can't, at least under Python 3.6. It produces an infinite recursion:

```
  File "/home/rav/work/synapse/synapse/storage/persist_events.py", line 174, in add_to_queue
    defer.Deferred(), consumeErrors=True
  File "/home/rav/.pyenv/versions/3.6.3/lib/python3.6/typing.py", line 1227, in __new__
    return _generic_new(cls.__next_in_mro__, cls, *args, **kwds)
  File "/home/rav/.pyenv/versions/3.6.3/lib/python3.6/typing.py", line 1193, in _generic_new
    obj.__orig_class__ = cls
  File "/home/rav/work/synapse/synapse/util/async_helpers.py", line 158, in __setattr__
    setattr(self._deferred, name, value)
  File "/home/rav/work/synapse/synapse/util/async_helpers.py", line 155, in __getattr__
    return getattr(self._deferred, name)
  File "/home/rav/work/synapse/synapse/util/async_helpers.py", line 155, in __getattr__
    return getattr(self._deferred, name)
```

No doubt this is fixable (probably by removing the `getattr`/`setattr` magic which tries to make ObservableDeferred a proxy for the original Deferred, for no good reason), but I can't be doing with that right now.

PR #10493, opened 2021-07-28
Remove the deprecation notice for the Third-Party Rules module interface

babolivier:
As per https://github.com/matrix-org/synapse/issues/10469#issuecomment-888213232, we don't want to deprecate/remove interfaces one by one anymore, but rather deprecate and remove them all at once.

CI failing is expected since this PR is missing a changelog entry on purpose.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -43,7 +43,6 @@ Bugfixes
 Synapse 1.39.0rc1 (2021-07-20)
 ==============================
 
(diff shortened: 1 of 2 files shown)


richvdh (commented):
we can still deprecate it, even if we don't remove the old one for a while?

babolivier:
Hmm fair enough.

richvdh:
well, my point was we could keep the deprecation notice, and remove the statement about removing it in September?

PR #10496, opened 2021-07-28
Extend the release script to tag and create the releases.

erikjohnston:
This splits the release script into three subcommands:

```
./scripts-dev/release.py prepare

# ... ask others to look at the changelog ...

./scripts-dev/release.py tag

# ... wait for asssets to build ...

./scripts-dev/release.py publish
```

Reviewable commit by commit.

### DIFF
--- a/scripts-dev/release.py
+++ b/scripts-dev/release.py
@@ -19,7 +19,7 @@
 
 import subprocess
 import sys
-from typing import Optional
+from typing import Optional, Tuple
 
 import click
 import git
@@ -56,32 +56,8 @@ def prepare():
     click.secho("Updating git repo...")
     repo.remote().fetch()
 
-    # Parse the AST and load the `__version__` node so that we can edit it
-    # later.
-    with open("synapse/__init__.py") as f:
-        red = RedBaron(f.read())
-
-    version_node = None
-    for node in red:
-        if node.type != "assignment":
-            continue
-
-        if node.target.type != "name":
-            continue
-
-        if node.target.value != "__version__":
-            continue
-
-        version_node = node
-        break
-
-    if not version_node:
-        print("Failed to find '__version__' definition in synapse/__init__.py")
-        sys.exit(1)
-
-    # Parse the current version.
-    current_version = version.parse(version_node.value.value.strip('"'))
-    assert isinstance(current_version, version.Version)
+    # Get the current version and AST from root Synapse module.
+    current_version, parsed_synapse_ast, version_node = parse_version_from_module()
 
     # Figure out what sort of release we're doing and calcuate the new version.
     rc = click.confirm("RC", default=True)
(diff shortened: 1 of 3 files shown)


scripts-dev/release.py
```diff
@@ -240,6 +244,171 @@ def run():
...
+        raise click.ClickException("Uncommitted changes exist.")
+
+    current_version, _, _ = parse_version_from_module()
+    tag_name = f"v{current_version}"
+
+    if not click.confirm(f"Publish {tag_name}?", default=True):
+        return
+
+    if gh_token:
+        # Create a new draft release
```
richvdh, line 337:
```suggestion
        # Publish the draft release
```

scripts-dev/release.py
```diff
@@ -240,6 +244,171 @@ def run():
...
+        f"https://github.com/matrix-org/synapse/actions?query=branch%3A{tag_name}"
+    )
+
+    click.echo("Wait for release assets to be built")
+
+
+@cli.command()
+@click.option("--gh-token", envvar=["GH_TOKEN", "GITHUB_TOKEN"])
+def publish(gh_token: Optional[str]):
+    """Publish release and upload to PyPI."""
```
richvdh, line 319: I can't help but feel these should be two separate commands, for clarity.

scripts-dev/release.py
```diff
@@ -334,7 +334,7 @@ def publish(gh_token: Optional[str]):
         return
 
     if gh_token:
```
richvdh, line 336: given that this now does nothing without a gh_token, I think it should error out rather than fail silently.

richvdh (commented):
lgtm otherwise

scripts-dev/release.py
```diff
@@ -240,6 +244,180 @@ def run():
...
+    for release in gh_repo.get_releases():
+        if release.title == tag_name:
+            break
+    else:
+        raise ClickException(f"Failed to find GitHub release for {tag_name}")
+
+    assert release.title == tag_name
+
+    if not release.draft:
+        if not click.confirm("Release already published. Continue?", default=True):
```
richvdh, line 348: continue to... do nothing?

erikjohnston: oh bah, less haste more speed or something

scripts-dev/release.py
```diff
@@ -240,6 +244,180 @@ def run():
...
+        raise click.ClickException("Not in Synapse repo.")
+
+    if repo.is_dirty():
+        raise click.ClickException("Uncommitted changes exist.")
+
+    current_version, _, _ = parse_version_from_module()
+    tag_name = f"v{current_version}"
+
+    if not click.confirm(f"Publish {tag_name}?", default=True):
+        return
```
richvdh, line 334: feel like this should be at the end, just before we actually publish it. but I'm nitpicking now.

richvdh (approved):
lgtm

PR #10498, opened 2021-07-28
Add support for MSC2716 marker events

MadLittleMods:
Add support for "marker" events which makes historical events discoverable for servers that already have all of the scrollback history.

Follow-up to https://github.com/matrix-org/synapse/pull/9247 and https://github.com/matrix-org/synapse/pull/10245

[Mermaid live editor playground link](https://mermaid-js.github.io/mermaid-live-editor/edit/#eyJjb2RlIjoiZmxvd2NoYXJ0IEJUXG4gICAgc3ViZ3JhcGggbGl2ZVxuICAgICAgICBtYXJrZXIxPlwibWFya2VyXCJdIC0tLS0-IEIgLS0tLS0tLS0tLS0tLS0tLS0-IEFcbiAgICBlbmRcbiAgICBcbiAgICBzdWJncmFwaCBjaHVuazBcbiAgICAgICAgY2h1bmswLWNodW5rW1tcImNodW5rXCJdXSAtLT4gY2h1bmswLTIoKFwiMlwiKSkgLS0-IGNodW5rMC0xKCgxKSkgLS0-IGNodW5rMC0wKCgwKSkgLS0-IGNodW5rMC1pbnNlcnRpb25bL2luc2VydGlvblxcXVxuICAgIGVuZFxuXG4gICAgc3ViZ3JhcGggY2h1bmsxXG4gICAgICAgIGNodW5rMS1jaHVua1tbXCJjaHVua1wiXV0gLS0-IGNodW5rMS0yKChcIjJcIikpIC0tPiBjaHVuazEtMSgoMSkpIC0tPiBjaHVuazEtMCgoMCkpIC0tPiBjaHVuazEtaW5zZXJ0aW9uWy9pbnNlcnRpb25cXF1cbiAgICBlbmRcbiAgICBcbiAgICBzdWJncmFwaCBjaHVuazJcbiAgICAgICAgY2h1bmsyLWNodW5rW1tcImNodW5rXCJdXSAtLT4gY2h1bmsyLTIoKFwiMlwiKSkgLS0-IGNodW5rMi0xKCgxKSkgLS0-IGNodW5rMi0wKCgwKSkgLS0-IGNodW5rMi1pbnNlcnRpb25bL2luc2VydGlvblxcXVxuICAgIGVuZFxuXG4gICAgXG4gICAgbWFya2VyMSAtLi0-IGNodW5rMC1pbnNlcnRpb25CYXNlXG4gICAgY2h1bmswLWluc2VydGlvbkJhc2VbL2luc2VydGlvblxcXSAtLS0tLS0tLS0tLS0tPiBBXG4gICAgY2h1bmswLWNodW5rIC0uLT4gY2h1bmswLWluc2VydGlvbkJhc2VbL2luc2VydGlvblxcXVxuICAgIGNodW5rMS1jaHVuayAtLi0-IGNodW5rMC1pbnNlcnRpb25cbiAgICBjaHVuazItY2h1bmsgLS4tPiBjaHVuazEtaW5zZXJ0aW9uXG4iLCJtZXJtYWlkIjoie30iLCJ1cGRhdGVFZGl0b3IiOmZhbHNlLCJhdXRvU3luYyI6dHJ1ZSwidXBkYXRlRGlhZ3JhbSI6ZmFsc2V9)

<details>
<summary>mermaid graph syntax</summary>

```mermaid
flowchart BT
    subgraph live
        marker1>"marker"] ----> B -----------------> A
    end
    
    subgraph chunk0
        chunk0-chunk[["chunk"]] --> chunk0-2(("2")) --> chunk0-1((1)) --> chunk0-0((0)) --> chunk0-insertion[/insertion\]
    end

    subgraph chunk1
        chunk1-chunk[["chunk"]] --> chunk1-2(("2")) --> chunk1-1((1)) --> chunk1-0((0)) --> chunk1-insertion[/insertion\]
    end
    
    subgraph chunk2
        chunk2-chunk[["chunk"]] --> chunk2-2(("2")) --> chunk2-1((1)) --> chunk2-0((0)) --> chunk2-insertion[/insertion\]
    end

    
    marker1 -.-> chunk0-insertionBase
    chunk0-insertionBase[/insertion\] -------------> A
    chunk0-chunk -.-> chunk0-insertionBase[/insertion\]
    chunk1-chunk -.-> chunk0-insertion
    chunk2-chunk -.-> chunk1-insertion
```

</details>

---

TARDIS visualization: https://github.com/matrix-org/tardis/pull/1

Screen Shot 2021-07-28 at 11 27 22 PM

### Dev notes

 - `event_backward_extremities`

#### Call stacks for backfilling

**From the remote server**

 - `RoomMessageListRestServlet.on_GET`: `/rooms/(?P<room_id>[^/]*)/messages$` (`hs2` asks for more messages)
 - `get_messages`
 - `maybe_backfill`
 - `_maybe_backfill_inner`
    - `get_oldest_events_with_depth_in_room`
    - `try_backfill`
 - `backfill`
 - `federation_client.backfill`
 - `transport_layer.backfill` -> calls `/backfill`

**From the origin server**

 - `FederationBackfillServlet.on_GET`: `/backfill/(?P<room_id>[^/]*)/?`
 - `on_backfill_request`
 - `get_backfill_events`

#### Where to fetch the missing insertion event?

 - `_persist_events_and_state_updates` -> `_persist_events_txn` -> `_update_metadata_tables_txn` ([`synapse/storage/databases/main/events.py#L1442`](https://github.com/matrix-org/synapse/blob/323452944e13114264b0c645db875dc6950315c5/synapse/storage/databases/main/events.py#L1442))
    - Where we handle insertion/chunk ID's in https://github.com/matrix-org/synapse/pull/10245
    - We don't have access to `async`/`await` stuff and probably not best to call federation code from there
 - `on_receive_pdu` [`synapse/handlers/federation.py#L193`](https://github.com/matrix-org/synapse/blob/323452944e13114264b0c645db875dc6950315c5/synapse/handlers/federation.py#L193)

#### How to fetch the missing insertion event?

 - `federation_client.backfill` ([`synapse/federation/federation_client.py#L203`](https://github.com/matrix-org/synapse/blob/323452944e13114264b0c645db875dc6950315c5/synapse/federation/federation_client.py#L203)): Not for fetching a specific event. Only goes off of extremities and goes backwards from there.

---

 - `_get_events_and_persist`: [`synapse/handlers/federation.py#L1303`](https://github.com/matrix-org/synapse/blob/323452944e13114264b0c645db875dc6950315c5/synapse/handlers/federation.py#L1303)
 - `federation_client.get_pdu`: [`synapse/federation/federation_client.py#L241`](https://github.com/matrix-org/synapse/blob/323452944e13114264b0c645db875dc6950315c5/synapse/federation/federation_client.py#L241)
 - `federation_client.transport_layer.get_event`: [`synapse/federation/transport/client.py#L76`](https://github.com/matrix-org/synapse/blob/323452944e13114264b0c645db875dc6950315c5/synapse/federation/transport/client.py#L76)

---

```
on_receive_pdu
_get_state_after_missing_prev_event
_get_events_and_persist
get_pdu
```

### Pull Request Checklist


synapse/visibility.py
```diff
@@ -366,7 +366,8 @@ def check_event_is_visible(event: EventBase, state: StateMap[EventBase]) -> bool
         if erased_senders:
             to_return = []
             for e in events:
-                if not is_sender_erased(e, erased_senders):
+                erased = is_sender_erased(e, erased_senders)
+                if not erased:
```
MadLittleMods, line 370: TODO: can revert

synapse/storage/databases/main/events.py
```diff
@@ -1845,6 +1845,18 @@ def _handle_chunk_event(self, txn: LoggingTransaction, event: EventBase):
...
 
+        # When we receive an event with a `chunk_id` referencing the
+        # `next_chunk_id` of the insertion event, we can remove it from the
+        # `insertion_event_extremities` table.
+        sql = """
+            DELETE FROM insertion_event_extremities WHERE event_id IN (
+                SELECT event_id FROM insertion_events
+                WHERE next_chunk_id = ?
+            )
+        """
```
MadLittleMods, line 1856: I'm unsure if this query is correct 🎯

---

```
# Doesn't seem to be supported in postgres
sql = """
    DELETE extrem FROM insertion_event_extremities as extrem
    INNER JOIN insertion_events AS i USING (event_id)
    WHERE i.next_chunk_id = ?
"""
```

```
sql = """
    DELETE FROM insertion_event_extremities WHERE event_id IN (
        SELECT event_id FROM insertion_events
        WHERE next_chunk_id = ?
    )
"""
```

synapse/handlers/federation.py
```diff
@@ -263,6 +264,7 @@ async def on_receive_pdu(
         state = None
 
         # Get missing pdus if necessary.
+        # We don't need to worry about outliers because TODO!
```
MadLittleMods, line 267: Unrelated to the PR but just want to better clarify why we skip outliers in this step. Any insight here?

 - Outliers are outside of the normal DAG
 - We don't have state for outliers
 - We trust their auth events over calculating them

synapse/handlers/federation.py
```diff
@@ -1106,10 +1191,19 @@ async def _maybe_backfill_inner(
...
+            "_maybe_backfill_inner: filtered_extremities %s", filtered_extremities
+        )
 
-        if not filtered_extremities:
+        if not filtered_extremities and not insertion_events_to_be_backfilled:
             return False
 
+        extremities = {
+            **oldest_events_with_depth,
+            # TODO: insertion_events_to_be_backfilled is currently skipping the filtered_extremities checks
```
MadLittleMods, line 1203: Does it even matter that `insertion_events_to_be_backfilled` is not going through the same `filtered_extremities` checks?

We don't even use `filtered_extremities` after we calculate it. Maybe that is a bug though 🤔. Currently, we filter the `extremities`, check if there is any left, and if there is, go back to using the full `extremities` in the rest of the code.

synapse/handlers/federation.py
```diff
@@ -1089,10 +1170,12 @@ async def _maybe_backfill_inner(
         #   state *before* the event, ignoring the special casing certain event
         #   types have.
 
-        forward_events = await self.store.get_successor_events(list(extremities))
+        forward_event_ids = await self.store.get_successor_events(
+            list(oldest_events_with_depth)
+        )
```
MadLittleMods, line 1175: Why do we get successor events here?

This seems to be explained by the comment:

> Given we don't have the extremity events themselves, we need to actually check the events that reference them.

Butttt..... we are already getting events that have `depth`. That's the whole point of `get_oldest_events_with_depth_in_room`.

This would only make sense if we returned the backward extremities.

This visibility filtering was added in https://github.com/matrix-org/synapse/commit/71b625d80806886794c5e72f7ff11432e99b736c but the code was still `get_oldest_events_with_depth_in_room` back then.

---

I feel like this is a bug and we don't need to get the forward events first.

synapse/storage/databases/main/event_federation.py
```diff
@@ -665,27 +665,95 @@ def _get_auth_chain_difference_txn(
...
+                /**
+                 * We find the "oldest" events in the room by looking for
+                 * events connected to backwards extremeties (oldest events
+                 * in the room that we know of so far).
+                 */
+                INNER JOIN event_backward_extremities as b
+                ON g.prev_event_id = b.event_id
+                WHERE b.room_id = ? AND g.is_state is ?
+                GROUP BY b.event_id
+            """
```
MadLittleMods, line 708: No change to the query, just new `"""` syntax and comments

synapse/handlers/federation.py
MadLittleMods: Changing this code does make some of my federation backfill Complement tests fail which is making it seem like it serves a purpose. But it's not clear to me 😕

synapse/handlers/federation.py
MadLittleMods: I'm open to filtering the `insertion_events_to_be_backfilled`, just give the nudge ⏩

synapse/storage/databases/main/events.py
erikjohnston: I think the query is fine. With postgres you can use a [`USING`](https://www.postgresql.org/docs/10/sql-delete.html) clause to join against tables, but that isn't supported in SQLite.

synapse/handlers/federation.py
erikjohnston: This bit of code is basically implementing the additional checks on events that are part of the "timeline" (on top of the checks we'll do anyway for outliers). Specifically: a) fetching any missing prev events to fill in gaps in the graph, and b) fetching state if we have a hole in the graph.

> Outliers are outside of the normal DAG

More accurately, they're outside the chunk of the DAG we currently have. 

> We trust their auth events over calculating them

There are two level of checks:

1. Check the event passes auth based on `auth_events`, this allows us to assert that the event would have been allowed at some point. If an event passes this check its OK for it to be used as part of a returned `/state` request, as either a) we received the event as part of the original join and so trust it, or b) we'll do a state resolution with existing state before it becomes part of the "current state", which adds more protection.
2. Check the event passes auth based on the state at the event. This is done for events that are to be added to the timeline.

synapse/handlers/federation.py
erikjohnston: `get_oldest_events_with_depth_in_room` I *think* returns the event ID of the extremity plus the max depth of all events that reference it? https://github.com/matrix-org/synapse/blob/c167e09fe58d3a256fb1c763b391ad6633d2507d/synapse/storage/databases/main/event_federation.py#L677-L683

synapse/handlers/federation.py
erikjohnston: I think the logic behind using `extremities` rather than `filtered_extremities` is that if we are making a `/backfill` request we may as well send over all the information we have. We're not really trying to filter the extremities but instead trying to answer the question of "should we even bother calling `/backfill`? There's an argument that it might be better to only use the filtered extremities, but ideally the remote will do the right thing.

> I'm open to filtering the `insertion_events_to_be_backfilled`, just give the nudge

I think it probably doesn't make sense to filter it, as the logic of checking the history visibility doesn't really apply? The worst case here is that we backfill, get a bunch of redacted events, and then next time we hit this code path we'll likely not have an insertion event and so the filtering logic will work.

synapse/handlers/federation.py
```diff
@@ -889,6 +891,75 @@ async def _process_received_pdu(
...
+    async def _handle_marker_event(self, origin: str, marker_event: EventBase):
+        """Handles backfilling the insertion event when we receive a marker
+        event that points to one.
+
+        Args:
+            origin: Origin of the event. Will be called to get the insertion event
+            marker_event: The event to process
+        """
+
+        if marker_event.type != EventTypes.MSC2716_MARKER:
```
erikjohnston, line 905: We should also check the event hasn't been rejected via `marker_event.rejected_reason is None`

synapse/storage/databases/main/event_federation.py
```diff
@@ -1130,6 +1197,21 @@ def _delete_old_forward_extrem_cache_txn(txn):
             _delete_old_forward_extrem_cache_txn,
         )
 
+    async def insert_insertion_extremity(self, event_id: str, room_id: str) -> None:
+        def _insert_insertion_extremity_txn(txn):
+            self.db_pool.simple_insert_txn(
```
erikjohnston, line 1202: You can use `await self.db_pool.simple_insert(...)` helper, to avoid this boilerplate

erikjohnston: Also, we probably want to use a `simple_upsert` here, to handle the case where we've already processed the event somehow, I think?

synapse/handlers/federation.py
MadLittleMods: Can you check `get_oldest_event_ids_with_depth_in_room` in this PR? It's the same function but with comments and is renamed to show that it returns `event_ids` (even though also includes `depth`)

I feel like it returns the closest events to the extremeties because the extremeties aren't persisted in our DB (and have no depth to reference).

synapse/handlers/federation.py
MadLittleMods: Added 👍

synapse/storage/databases/main/event_federation.py
MadLittleMods: Should I be preferring the `async` helpers over the `txn` stuff? Any context behind why both?

It seems some of the code isn't `async` yet and kinda forces into a `txn` corner.

synapse/storage/databases/main/event_federation.py
```diff
@@ -1130,6 +1197,18 @@ def _delete_old_forward_extrem_cache_txn(txn):
...
+        await self.db_pool.simple_upsert(
+            table="insertion_event_extremities",
+            keyvalues={"event_id": event_id},
+            values={
+                "event_id": event_id,
+                "room_id": room_id,
+            },
+            insertion_values={},
+            desc="insert_insertion_extremity",
+        )
```
MadLittleMods, line 1210: Can we set `lock=false` on this? We're using `event_id` as the `keyvalues` which has a unique index constraint.

> `lock` should generally be set to True (the default), but can be set
>
> there is a UNIQUE INDEX on the key columns. In this case a conflict
          will cause an IntegrityError in which case this function will retry
          the update.

synapse/handlers/federation.py
MadLittleMods: Added this info to a comment there 🤔

 1. For 1, I think this is in `_process_received_pdu`
 1. For 2, I think this is the current position of this comment in `on_receive_pdu`

synapse/handlers/federation.py
erikjohnston: The query is returning the column `event_backward_extremities.event_id` (in the query the table is aliased as `b`), where the events listed in  `event_backward_extremities` have *not* been persisted yet. There is a bunch of joining to figure out what events point to those and what their max depth is, which is a bit confusing, but doesn't affect the returned event IDa

synapse/storage/databases/main/event_federation.py
erikjohnston: Yes we can. Though the docstring is a bit misleading as we'll use native upserts (which ignores the `lock` param) for everything bar old SQLite versions, so it doesn't really matter in practice.

synapse/storage/schema/main/delta/61/01insertion_event_lookups.sql
```diff
@@ -37,6 +37,15 @@ CREATE UNIQUE INDEX IF NOT EXISTS insertion_event_edges_event_id ON insertion_ev
...
 
+-- Add a table that keeps track of which "insertion" events need to be backfilled
+CREATE TABLE IF NOT EXISTS insertion_event_extremities(
+    event_id TEXT NOT NULL,
+    room_id TEXT NOT NULL
+);
+
+CREATE UNIQUE INDEX IF NOT EXISTS insertion_event_extremities_event_id ON insertion_event_extremities(event_id);
+CREATE INDEX IF NOT EXISTS insertion_event_extremities_room_id ON insertion_event_extremities(room_id);
+
```
erikjohnston, line 48: Bah, I don't know how I missed this, but this needs to be done in a new delta file, as we don't rerun delta files that have been run previously. (The only reason we tend to use `IF NOT EXISTS` is that it makes it easy to revert and rerun the delta when testing).

synapse/storage/databases/main/event_federation.py
erikjohnston: > Should I be preferring the `async` helpers over the `txn` stuff? Any context behind why both?

Yes, as it reduces the boiler plate. We have `_txn` versions for when we want to run multiple queries in a single transaction (for atomicity).

> It seems some of the code isn't `async` yet and kinda forces into a `txn` corner.

The async helpers should  always be usable wherever we're doing this currently, as `runInteraction` is also `async`. What you may be seeing is that if you have a function that calls a single async function then technically it can just be a normal function that returns the result of the call to the async function (i.e. just return the coroutine, rather than awaiting it and returning the result). That's not a great thing to do though, as it messes up stack traces, but they crept in as we converted stuff from `@defer.inlineCallbacks` to async/await.

synapse/handlers/federation.py
MadLittleMods: Ahhh, thanks for this! I assumed the `depth` would be associated with the event chosen to return but makes sense that it wouldn't if it was `b.event_id`(`event_backward_extremities.event_id`) being returned.

I've updated the comments to point out that it's an approximate depth based the persisted events connected to the extremity 👍

synapse/storage/databases/main/event_federation.py
MadLittleMods: Thanks for the info 😀

I'm thinking specifically of [`synapse/storage/databases/main/events.py`](https://github.com/matrix-org/synapse/blob/72935b7c5053af122c8cb5767a3e85a3a0f3a20c/synapse/storage/databases/main/events.py#L1826) where I had to refactor `get_room_version` to be available as `get_room_version_txn`. But it seems like "We have _txn versions for when we want to run multiple queries in a single transaction (for atomicity)." applies in this case so we can persist multiple events at once (`_persist_events_txn`).

synapse/storage/databases/main/event_federation.py
MadLittleMods: Added and updated the docstring there to mention it only matters for SQLite 👍

synapse/storage/schema/main/delta/61/01insertion_event_lookups.sql
MadLittleMods: Moved to `62` 👍

synapse/storage/databases/main/event_federation.py
erikjohnston: > But it seems like "We have _txn versions for when we want to run multiple queries in a single transaction (for atomicity)." applies in this case so we can persist multiple events at once (`_persist_events_txn`).

fwiw, it also applies when persisting only one event, as we want to either add the event to all the appropriate tables or none of them.

synapse/storage/schema/main/delta/61/01insertion_event_lookups.sql
erikjohnston: Moving to `62` absolutely works, but just to note that you could have also moved it to a different file in `61/` and that would have also worked fine. (It's a bit handwavey when we should cut a new schema version, but 61 is getting a bit big so may as well do one now)

erikjohnston (approved):

MadLittleMods:
Woot! All of the major pieces are merged now 🙌 Full-fat federated historical messages!

@erikjohnston Thanks for the tremendous efforts reviewing each of these big pieces and onboarding me onto all of the Synapse/Matrix concepts to be able to accomplish this 🤠❤

synapse/storage/schema/main/delta/61/01insertion_event_lookups.sql
MadLittleMods: Ahhh, I assumed we already shipped a Synapse version with `61` so people who already ran the upgrades wouldn't run deltas in that directory again. (probably what I just said is even wrong on how it works) ([docs](https://github.com/matrix-org/synapse/blob/develop/docs/development/database_schema.md))

erikjohnston:
Woo! :tada:  Thanks for bearing with all the back and forth! :heart:

PR #10517, opened 2021-08-03
Fix PeriodicallyFlushingMemoryHandler inhibiting application shutdown

reivilibre:
not tested (not immediately sure how to -- `synctl stop` used to already work)


richvdh (approved):
I kinda wonder if it would be better to find a way to wake up the sleeping thread, but I think this is fine.

reivilibre:
this has now been tested by forcing the failure conditions to happen and trying locally... `daemon=True` makes all the difference.

PR #10518, opened 2021-08-03
Make `PeriodicallyFlushingMemoryHandler` (buffered logging handler which periodically flushes itself) the default logging handler.

reivilibre:
merge blocked:
- [x] check that #10517 is successful
- [x] check that it doesn't explode matrix.org

### DIFF
--- a/changelog.d/10518.feature
+++ b/changelog.d/10518.feature
@@ -1,0 +1,1 @@
+Make `PeriodicallyFlushingMemoryHandler` (buffered logging handler which periodically flushes itself) the default logging handler.
--- a/changelog.d/10518.feature
+++ b/changelog.d/10518.feature
@@ -1,1 +1,1 @@
-Make `PeriodicallyFlushingMemoryHandler` (buffered logging handler which periodically flushes itself) the default logging handler.
+The default logging handler for new installations is now `PeriodicallyFlushingMemoryHandler`, a buffered logging handler which periodically flushes itself.
--- a/docker/conf/log.config
+++ b/docker/conf/log.config
@@ -25,19 +25,23 @@ handlers:
   buffer:
     class: synapse.logging.handlers.PeriodicallyFlushingMemoryHandler
     target: file
+
     # The capacity is the maximum number of log lines that are buffered
     # before being written to disk. Increasing this will lead to better
     # performance, at the expensive of it taking longer for log lines to
     # be written to disk.
     # This parameter is required.
     capacity: 10
+
     # Logs with a level at or above the flush level will cause the buffer to
     # be flushed immediately.
     # Default value: 40 (ERROR)
     # Other values: 50 (CRITICAL), 30 (WARNING), 20 (INFO), 10 (DEBUG)
     flushLevel: 30  # Flush immediately for WARNING logs and higher
+
     # The period of time, in seconds, between forced flushes.
     # Messages will not be delayed for longer than this time.
+    # Default value: 5 seconds
     period: 5
 {% endif %}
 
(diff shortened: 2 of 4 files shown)


changelog.d/10518.feature
```diff
@@ -0,0 +1 @@
+Make `PeriodicallyFlushingMemoryHandler` (buffered logging handler which periodically flushes itself) the default logging handler.
```
richvdh, line 1:
```suggestion
The default logging handler for new installations is now `PeriodicallyFlushingMemoryHandler`, a buffered logging handler which periodically flushes itself.
```

richvdh (changes requested):
https://github.com/matrix-org/synapse/blob/develop/docker/conf/log.config needs an update too.

docs/sample_log_config.yaml
```diff
@@ -28,14 +28,17 @@ handlers:
     # will be a delay for INFO/DEBUG logs to get written, but WARNING/ERROR
```
richvdh, line 28: this isn't quite accurate any more.

richvdh (approved):
looks fine, though given that `sample_log_config` is the only place that this thing is documented, I wonder if we could do a bit better with it. How do `capacity`, `flushLevel` and `period` interact? What are their default values if left unset?

docs/sample_log_config.yaml
reivilibre: I see what you mean, but it also *kind* of is. I have updated the wording now, mind letting me know if you think it's better?

synapse/config/logger.py
```diff
@@ -67,18 +67,27 @@
...
+    # WARNING/ERROR logs will still be flushed immediately, but there will be a
+    # delay (of up to `period` seconds, or until the buffer is full with
+    # `capacity` messages) before INFO/DEBUG logs get written.
     buffer:
-        class: logging.handlers.MemoryHandler
+        class: synapse.logging.handlers.PeriodicallyFlushingMemoryHandler
         target: file
-        # The capacity is the number of log lines that are buffered before
-        # being written to disk. Increasing this will lead to better
+        # The capacity is the maximum number of log lines that are buffered
```
richvdh, line 77: probably worth adding some whitespace between these options:
```suggestion

        # The capacity is the maximum number of log lines that are buffered
```

synapse/config/logger.py
```diff
@@ -67,18 +67,27 @@
...
         capacity: 10
-        flushLevel: 30  # Flush for WARNING logs as well
+        # Logs with a level at or above the flush level will cause the buffer to
+        # be flushed immediately.
+        # Default value: 40 (ERROR)
+        # Other values: 50 (CRITICAL), 30 (WARNING), 20 (INFO), 10 (DEBUG)
+        flushLevel: 30  # Flush immediately for WARNING logs and higher
+        # The period of time, in seconds, between forced flushes.
+        # Messages will not be delayed for longer than this time.
+        period: 5
```
richvdh, line 90: what's the default for this? what happens if it is omitted?

docker/conf/log.config
```diff
@@ -18,18 +18,27 @@ handlers:
...
-    flushLevel: 30  # Flush for WARNING logs as well
+    # Logs with a level at or above the flush level will cause the buffer to
+    # be flushed immediately.
+    # Default value: 40 (ERROR)
+    # Other values: 50 (CRITICAL), 30 (WARNING), 20 (INFO), 10 (DEBUG)
+    flushLevel: 30  # Flush immediately for WARNING logs and higher
+    # The period of time, in seconds, between forced flushes.
+    # Messages will not be delayed for longer than this time.
+    period: 5
 {% endif %}
```
richvdh, line 42: I'd actually be quite happy for this to be very basic, and just link to the sample_log_config for docs. But this is fine too.

richvdh (commented):
much better! A couple of little nits.

richvdh (approved):

PR #10520, opened 2021-08-04
Send and handle cross-signing messages using the stable prefix.

uhoreg:


uhoreg:
I suspect that the test failure is just a test being flaky, since it doesn't seem to be related to my change, and other tests succeed.

richvdh:
> I suspect that the test failure is just a test being flaky, since it doesn't seem to be related to my change, and other tests succeed.

indeed: https://github.com/matrix-org/sytest/issues/1037

richvdh (commented):
I'm a bit worried that the fact we send both types of EDU means the target server will store the updates twice and wake up all the clients twice. Can you convince me this isn't going to be significant? Or can we find a different solution?

uhoreg:
If you're concerned about waking up the clients twice, we can probably change it so that it only wakes up clients if the keys were actually changed (i.e. compare with the old key, and do nothing if it's the same).  But I think it shouldn't happen too often, so it shouldn't be a problem.

richvdh (approved):
Ah, I hadn't realised this was only relevant to cross-signing keys.

seems fair to me. The sooner that we can get rid of the unstable EDUs the better, though.

richvdh:
(could you merge latest develop to check the tests pass?)

uhoreg:
> (could you merge latest develop to check the tests pass?)

Done

Issue #10495, opened 2021-07-28
Implement updated spaces summary API


clokep:
[MSC2946](https://github.com/matrix-org/matrix-doc/pull/2946) was recently rewritten based on some changes in design and lessons learned during initial implementation. We need to implement the changes and test them, etc.

The main design changes are:

* The API now has a `max_depth` instead of a `max_rooms_per_space`.
* The API now has pagination (controlled via `limit` and `from` in the requests and `next_token` in the response).
* The responses now embed the child events with the room.
* The server-server API is significantly different.

This is enough changes that I think we'll need to add new API endpoints / handlers for this code. I think we're going to want to keep backwards compatibility (for at least a few releases) so can't quite get rid of the old code yet. This will be a bit annoying, but should get cleaned up when we switch to the stable version.

I think some rough steps of what we'll want to do are:

1. [x] Refactor some of the existing code.
2. [x] Implement a version without federation (but with pagination) using in-memory storage.
3. [x] Hook up federation APIs to fetch unknown rooms.
4. [x] Handle any pending TODOs.
4. [ ] Handle persisting to the database (for multiple workers).
5. [x] Expire old persisted data.
5. [ ] Explore caching of requests made over federation.
6. [x] Attempt to fetch from the old federation APIs and translate.
6. [x] ~~Explore whether pagination tokens can be shared between users (with results cached / reused).~~

clokep:
We also need to update `allowed_spaces` -> `allowed_room_ids`.

clokep:
The local version of this is done now, with the federation APIs up for review.

I'm wondering about caching the requests made out over federation -- is this something that we do anywhere else? Ideally we would cache them so that they could be used by multiple workers...but again I'm not sure this is something we do? Maybe if the request was routed via the `federation_reader` or something that would make sense?

clokep:
Also see matrix-org/matrix-js-sdk#1818 for the Element Web side.

clokep:
I'm going to call this "done" and handle the follow-ups as part of #10523, which are all to do with optimizations / performance.

This feature will be coming out in Synapse 1.41.0, which should have an RC available this week.

PR #10527, opened 2021-08-04
Prep work for the updated spaces summary

clokep:
This is prep work related to #10495. I want to be able to use some of the same data structures, but the eventual format of the data returned is a bit different. Grouping the room + the children rooms together into the same data structure allows for this a bit more easily.

This also switches to using auto-attribs while I'm here.

Reviewable commit-by-commit.

### DIFF
--- a/synapse/federation/federation_client.py
+++ b/synapse/federation/federation_client.py
@@ -1290,7 +1290,7 @@ class FederationClient(FederationBase):
         )
 
 
-@attr.s(frozen=True, slots=True)
+@attr.s(frozen=True, slots=True, auto_attribs=True)
 class FederationSpaceSummaryEventResult:
     """Represents a single event in the result of a successful get_space_summary call.
 
@@ -1299,12 +1299,12 @@ class FederationSpaceSummaryEventResult:
     object attributes.
     """
 
-    event_type = attr.ib(type=str)
-    state_key = attr.ib(type=str)
-    via = attr.ib(type=Sequence[str])
+    event_type: str
+    state_key: str
+    via: Sequence[str]
 
     # the raw data, including the above keys
-    data = attr.ib(type=JsonDict)
+    data: JsonDict
 
     @classmethod
     def from_json_dict(cls, d: JsonDict) -> "FederationSpaceSummaryEventResult":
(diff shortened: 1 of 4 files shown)


synapse/handlers/space_summary.py
```diff
@@ -141,12 +143,10 @@ async def get_space_summary(
                 # user is not permitted see.
                 #
                 # Filter the returned results to only what is accessible to the user.
-                room_ids = set()
                 events = []
-                for room in fed_rooms:
-                    fed_room_id = room.get("room_id")
-                    if not fed_room_id or not isinstance(fed_room_id, str):
-                        continue
```
clokep, line 149: These checks were moved to `_summarize_remote_room` (since the `_RoomEntry` has a `room_id` attribute we need to check it before creating the object).

synapse/federation/federation_client.py
```diff
@@ -1335,15 +1335,15 @@ def from_json_dict(cls, d: JsonDict) -> "FederationSpaceSummaryEventResult":
         if any(not isinstance(v, str) for v in via):
             raise ValueError("Invalid event: 'via' must be a list of strings")
 
-        return cls(event_type, state_key, via, d)
+        return cls(event_type, room_id, state_key, via, d)
```
richvdh, line 1343: this has ended up in the wrong commit, but nm

richvdh (approved):
lgtm

synapse/handlers/space_summary.py
```diff
@@ -612,6 +618,17 @@ class _RoomQueueEntry:
...
 
+@attr.s(frozen=True, slots=True, auto_attribs=True)
+class _RoomEntry:
+    room_id: str
+    # The room summary for this room.
+    room: JsonDict
+    # An iterable of the sorted, stripped children events for children of this room.
+    #
+    # This may not include all children.
+    children: List[JsonDict] = attr.ib(factory=list)
```
richvdh, line 629: wondering if this should be a Collection rater than necessarily a List (and the default could be the empty tuple rather than a new list).

synapse/federation/federation_client.py
clokep: Good shout, sorry about that. I had much rebasing in these branches... 😢

synapse/handlers/space_summary.py
clokep: I originally had it as a `Collection`, this was causing issues though since in `get_space_summary` (for the local room case) we directly use it as `events`, but for the remote case we call `append()` on `events`.

I could probably restructure that code, but it seemed a bit silly to do that just to get a more generic type (when all the callers *do* actually set it to a list).

I can take another look if you'd like?

clokep: Seems like it was easy enough actually!

PR #10529, opened 2021-08-04
Don't fail CI when a job was skipped

erikjohnston:
We don't always run all jobs, e.g. we only run the newsfile lint on PRs,
which means that currently CI always fails on the develop branch.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -367,12 +367,14 @@ jobs:
       - name: Set build result
         env:
           NEEDS_CONTEXT: ${{ toJSON(needs) }}
+          JOB_STATUS: ${{ job.status }}
         # the `jq` incantation dumps out a series of "<job> <result>" lines.
         # we set it to an intermediate variable to avoid a pipe, which makes it
         # hard to set $rc.
         run: |
           rc=0
           results=$(jq -r 'to_entries[] | [.key,.value.result] | join(" ")' <<< $NEEDS_CONTEXT)
+          echo "Current job status: $JOB_STATUS"
           while read job result ; do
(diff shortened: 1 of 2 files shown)


.github/workflows/tests.yml
```diff
@@ -374,7 +374,7 @@ jobs:
           rc=0
           results=$(jq -r 'to_entries[] | [.key,.value.result] | join(" ")' <<< $NEEDS_CONTEXT)
           while read job result ; do
-              if [ "$result" != "success" ]; then
+              if [ "$result" != "success" ] || [ "$result" != "skipped" ]; then
```
babolivier, line 377: Shouldn't this be an `&&`?

erikjohnston: oh balls

richvdh: hrm, this looks risky.

If a job is cancelled, all the steps will be skipped. We don't really want to accept such test runs.

richvdh (changes requested):

.github/workflows/tests.yml
erikjohnston: Hmm, looks like some stuff returns cancelled, and others skipped? I guess that's the difference between those that were already running and those that hadn't? c.f. https://github.com/matrix-org/synapse/runs/3241552601

richvdh: yeah. I'm a bit worries about a race condition where the run is cancelled at just the right time so that the first set of jobs run to completion, but the second set all fail.

Could we only allow `lint-newsfile` to be skipped for now?

erikjohnston: > Could we only allow `lint-newsfile` to be skipped for now?

I have done this because all other options I can see are *awful*

richvdh (approved):

PR #10413, opened 2021-07-16
Support for MSC2285 (hidden read receipts)

SimonBrandner:
* [x] Pull request is based on the develop branch
* [x] Pull request includes a [changelog file](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md#changelog).
* [x] Pull request includes a [sign off](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md#sign-off)
* [x] Code style is correct (run the [linters](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md#code-style))

<hr>

Based on https://github.com/matrix-org/synapse/pull/5990
Implementation of https://github.com/matrix-org/matrix-doc/pull/2285

<hr>

To be honest, this PR doesn't fill me with feelings of confidence as I don't have much backend and Python development experience :sweat_smile: In case it's bad, sorry and feel free to reject it

PR #10531, opened 2021-08-04
Don't fail on empty bodies when sending out read receipts

babolivier:
Fixes a bug introduced in rc1 that would cause Synapse to 400 on read receipts requests with empty bodies.

Broken in https://github.com/matrix-org/synapse/pull/10413

### DIFF
--- a/changelog.d/10531.bugfix
+++ b/changelog.d/10531.bugfix
@@ -1,0 +1,1 @@
+Fix a bug introduced in Synapse v1.40.0rc1 that would cause Synapse to respond with an error when clients would update their read receipts.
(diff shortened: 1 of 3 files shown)


reivilibre:
Does this mean we don't have tests that check this worked?

babolivier:
> Does this mean we don't have tests that check this worked?

We have tests with a body that's an empty JSON object (`{}`) but none with no body at all, I'm looking at adding one right now.

erikjohnston (approved):

kegsay:
Please add these tests to Complement so all homeservers can benefit.

richvdh:
> Please add these tests to Complement so all homeservers can benefit.

that presumes that we'll be making the body optional in the spec. Again, please see https://github.com/matrix-org/matrix-doc/issues/3237.

PR #10532, opened 2021-08-04
Check that an appservice protocol has at least one service providing it

### LINKED ISSUE #9694: /thirdparty/protocols API returned an empty object for an appservice that wasn't ready
### Description
`/thirdparty/protocols` returned 
```
{ irc: {} }
```

The `irc` object is missing a few required values.

https://matrix.org/docs/spec/client_server/r0.6.1#get-matrix-client-r0-thirdparty-protocols

### Steps to reproduce

- I set up my IRC bridge but the pingback port wasn't working yet (I had set it to the wrong port in the appservice file) and the IRC server on localhost wasn't connecting yet.

Synapse-config, Bridge config and logs (look around the time 2021-03-25 20:44):
https://gateway.pinata.cloud/ipfs/QmTbjP9PwfaTELkYQQ1HmznR4n8GNusveE6BwpZXw5tjk4

<!--

``` (three backticks, on a line on their own), so that they are formatted legibly.
-->

### Version information

<!--

listen on that port.)
-->
- **Version**:
- Synapse 1.29.0 with Python 3.8.8
- https://registry.hub.docker.com/layers/matrixdotorg/synapse/latest/images/sha256-67a507a29e0406a0e93a9040617380f7310f625e86adb5652ef24cd1ce39db48

- **Install method**:
- DockerHub's `matrixdotorg/synapse:latest`

- **Platform**:
- Manjaro Linux (based on the repo of Arch Linux)

### Additional notes:

The API response caused an issue in Element Web because it's not spec-compliant.

https://github.com/vector-im/element-web/issues/16789

https://user-images.githubusercontent.com/10872136/112535146-3a7b9100-8dac-11eb-9270-a7f4746d6300.mp4

### PR DESCRIPTION
**Pull request by Half-Shot**
Fixes #9694 

If there is no information to serve for a given protocol, we should prefer to omit that protocol rather than returning an empty object. The spec https://matrix.org/docs/spec/client_server/r0.6.0#get-matrix-client-r0-thirdparty-protocols mandates that protocols must contain at least a few fields to be valid.

### DIFF
--- a/changelog.d/10532.bugfix
+++ b/changelog.d/10532.bugfix
@@ -1,0 +1,1 @@
+Only report protocols that are implemented by at least one service, as required by the spec when handling `GET /_matrix/client/r0/thirdparty/protocols`.
--- a/changelog.d/10532.bugfix
+++ b/changelog.d/10532.bugfix
@@ -1,1 +1,1 @@
-Only report protocols that are implemented by at least one service, as required by the spec when handling `GET /_matrix/client/r0/thirdparty/protocols`.
+Fix a long-standing bug where protocols which are not implemented by any appservices were incorrectly returned via `GET /_matrix/client/r0/thirdparty/protocols`.
--- a/synapse/handlers/appservice.py
+++ b/synapse/handlers/appservice.py
@@ -392,9 +392,6 @@ class ApplicationServicesHandler:
                     protocols[p].append(info)
 
         def _merge_instances(infos: List[JsonDict]) -> JsonDict:
-            if not infos:
-                return None
-
             # Merge the 'instances' lists of multiple results, but just take
             # the other fields from the first as they ought to be identical
             # copy the result so as not to corrupt the cached one
--- a/synapse/handlers/appservice.py
+++ b/synapse/handlers/appservice.py
@@ -406,7 +406,11 @@ class ApplicationServicesHandler:
 
             return combined
 
-        return {p: _merge_instances(protocols[p]) for p in protocols.keys()}
+        return {
+            p: _merge_instances(protocols[p])
+            for p in protocols.keys()
+            if len(protocols[p]) > 0
+        }
 
     async def _get_services_for_event(
         self, event: EventBase
--- a/synapse/handlers/appservice.py
+++ b/synapse/handlers/appservice.py
@@ -393,7 +393,7 @@ class ApplicationServicesHandler:
 
         def _merge_instances(infos: List[JsonDict]) -> JsonDict:
             if not infos:
-                return {}
+                return None
 
             # Merge the 'instances' lists of multiple results, but just take
             # the other fields from the first as they ought to be identical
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: synapse/handlers/appservice.py [hunk 1]
```diff
@@ -406,7 +406,11 @@ def _merge_instances(infos: List[JsonDict]) -> JsonDict:
 
             return combined
 
-        return {p: _merge_instances(protocols[p]) for p in protocols.keys()}
+        return {
+            p: _merge_instances(protocols[p])
+            for p in protocols.keys()
+            if len(protocols[p]) > 0
```
clokep, line 412: This can probably just be:

```suggestion
            if protocols[p]
```

But something else seems bogus here -- `_merge_instances` will return an empty dict if there's no protocols passed in. Should that throw an exception or something if we're explicitly filtering those out?

Half-Shot: I don't think we want the request to fail so throwing seems off. Returning None will return `{"my-protocol": None}` which isn't the same as `{}`, which is what the spec expects us to return if the protocol is incomplete (i.e. it has no appservices implementing it).

Removing the `len` seems sane however.

clokep: Right, my point was mostly that it was redundant Removing it (like you did) seems reasonable.

synapse/handlers/appservice.py
```diff
@@ -406,7 +403,11 @@ def _merge_instances(infos: List[JsonDict]) -> JsonDict:
 
             return combined
 
-        return {p: _merge_instances(protocols[p]) for p in protocols.keys()}
+        return {
+            p: _merge_instances(protocols[p])
+            for p in protocols.keys()
+            if protocols[p]
```
clokep, line 409:
```suggestion
            p: _merge_instances(protocols[p]) for p in protocols.keys() if protocols[p]
```

synapse/handlers/appservice.py
Half-Shot: Ah, I see. Good good.

clokep:
Are there any tests for this code (in sytest or unit tests)? It would be beneficial to add some for edge-cases like this.

Half-Shot:
Complement looked too scary / time consuming for trying to mock out appservice responses, so I've settled for some classic unit tests.

changelog.d/10532.bugfix
```diff
@@ -0,0 +1 @@
+Only report protocols that are implemented by at least one service, as required by the spec when handling `GET /_matrix/client/r0/thirdparty/protocols`.
```
clokep, line 1: I assume this has always been broken?

```suggestion
Fix a long-standing bug where protocols which are not implemented by any appservices were incorrectly returned via `GET /_matrix/client/r0/thirdparty/protocols`.
```

clokep (approved):
LGTM! Thanks for adding tests!

changelog.d/10532.bugfix
Half-Shot: Well put :+1:

PR #10533, opened 2021-08-04
Track in memory events using weakrefs

erikjohnston:
Pulled out form #10119

### DIFF
--- a/synapse/events/__init__.py
+++ b/synapse/events/__init__.py
@@ -97,7 +97,7 @@ class DefaultDictProperty(DictProperty):
 
 
 class _EventInternalMetadata:
-    __slots__ = ["_dict", "stream_ordering", "outlier"]
+    __slots__ = ["_dict", "stream_ordering", "outlier", "redacted_by"]
 
     def __init__(self, internal_metadata_dict: JsonDict):
         # we have to copy the dict, because it turns out that the same dict is
@@ -110,6 +110,10 @@ class _EventInternalMetadata:
         # whether this event is an outlier (ie, whether we have the state at that point
         # in the DAG)
         self.outlier = False
+
+        # Whether this event has a valid redaction event pointing at it (i.e.
+        # whether it should be redacted before giving to clients).
+        self.redacted_by: Optional[str] = None
 
     out_of_band_membership: bool = DictProperty("out_of_band_membership")
     send_on_behalf_of: str = DictProperty("send_on_behalf_of")
--- a/synapse/storage/databases/main/events_worker.py
+++ b/synapse/storage/databases/main/events_worker.py
@@ -625,6 +625,8 @@ class EventsWorkerStore(SQLBaseStore):
             # Otherwise check if we still have the event in memory.
             event = self._event_ref.get(event_id)
             if event:
+                # Reconstruct an event cache entry
+
                 redacted_event = None
                 if event.internal_metadata.redacted_by is not None:
                     # The event has been redacted, so we generate a redacted
--- a/synapse/storage/databases/main/events_worker.py
+++ b/synapse/storage/databases/main/events_worker.py
@@ -14,6 +14,7 @@
 
 import logging
 import threading
+import weakref
 from typing import (
     Collection,
     Container,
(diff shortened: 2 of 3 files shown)


synapse/events/__init__.py
```diff
@@ -111,6 +111,10 @@ def __init__(self, internal_metadata_dict: JsonDict):
         # in the DAG)
         self.outlier = False
 
+        # Whether this event has a valid redaction event pointing at it (i.e.
+        # whether it should be redacted before giving to clients).
+        self.redacted_by: Optional[str] = None
```
erikjohnston, line 116: From https://github.com/matrix-org/synapse/pull/10119#discussion_r681126225:

> I'm super dubious about stuffing more things in here that don't form part of the `internal_metadata` column in the database.
> 
> Should we start a new attrs class for things like this, `stream_ordering` and `outlier`? (it also looks like `before`, `after`, `order` might fall into the same category).

> also: it seems confusing to have both a `redacted_by` and a `redacted` property.

synapse/storage/databases/main/events_worker.py
```diff
@@ -608,13 +615,36 @@ def _get_events_from_cache(
...
+                # Reconstruct an event cache entry
+
+                redacted_event = None
+                if event.internal_metadata.redacted_by is not None:
+                    # The event has been redacted, so we generate a redacted
+                    # version.
+                    redacted_event = prune_event(event)
+                    redacted_event.unsigned[
+                        "redacted_by"
+                    ] = event.internal_metadata.redacted_by
```
erikjohnston, line 637: From https://github.com/matrix-org/synapse/pull/10119#discussion_r681145699:

> do we not need `redacted_because` here?

synapse/storage/databases/main/events_worker.py
```diff
@@ -608,13 +615,36 @@ def _get_events_from_cache(
...
+                cache_entry = _EventCacheEntry(
+                    event=event,
+                    redacted_event=redacted_event,
+                )
+                event_map[event_id] = cache_entry
 
-            event_map[event_id] = ret
+                # We add the entry back into the cache as we want to keep
+                # recently queried events in the cache.
+                self._get_event_cache.set((event_id,), cache_entry)
```
erikjohnston, line 647: From https://github.com/matrix-org/synapse/pull/10119#discussion_r681146647:

> could we refactor the similar code in `_get_events_from_db`/`_maybe_redact_event_row` so that we can reuse some common code rather than have two similar but slightly different impls?
> 
> For example: have `_maybe_redact_event_row` just return a boolean rather than the redaction; have `_get_events_from_db` poke `internal_metadata.redacted_by` based on that boolean; then move these lines here out to a `build_cache_entry` which can be called by `_get_events_from_db`.

synapse/events/__init__.py
richvdh: the Event class, I think.

synapse/storage/databases/main/events_worker.py
```diff
@@ -738,13 +748,29 @@ def _get_events_from_cache(
         event_map = {}
 
         for event_id in events:
+            # First check if its in the event cache
```
DMRobertson, line 751:
```suggestion
            # First check if it's in the event cache
```

synapse/storage/databases/main/events_worker.py
```diff
@@ -738,13 +748,29 @@ def _get_events_from_cache(
...
             ret = self._get_event_cache.get(
                 (event_id,), None, update_metrics=update_metrics
             )
-            if not ret:
-                continue
+            if ret:
+                event_map[event_id] = ret
+
+            # Otherwise check if we still have the event in memory.
```
DMRobertson, line 758: "Otherwise" is confusing here. We run through all of lines 758--771 even if there was a (strong) cache hit on lines 752--755. Isn't that redundant if `ret` was truthy?

Do we want a `continue` on 757 or similar bit of control flow?

DMRobertson (changes requested):
Otherwise LGTM

synapse/storage/databases/main/events_worker.py
erikjohnston: Oh, yes, you're quite right.

DMRobertson (approved):

PR #10536, opened 2021-08-05
Oembed config - oembed_endpoints (#2752)

srdjan-catalyst:
Moved twitter patterns from preview_url_resource to oembed_endpoints

Signed-off-by: Srdjan <srdjan@catalyst.net.nz>

### Pull Request Checklist


clokep:
See also #10392.

srdjan-catalyst:
> See also #10392.

Yes, it's a bit unfortunate. I made the change some time ago and did not have time to do the PR. Pick one that you like better I guess.

richvdh:
I think this is meant to fix #2752.

richvdh (changes requested):
Thanks for the contribution!

I've written some words about how I think we should approach this over at #2752 (comment). TL;DR: let's use a `providers.json` file rather than config.

Also: note that CI is failing; we'll need to fix the failures before we can merge anything.

srdjan-catalyst:
@richvdh I'm about to commit again. Would you prefer a separate commit, or squashed force push this time please?

richvdh:
a new commit please. there is never any need to squash.

synapse/config/oembed.py
```diff
@@ -0,0 +1,155 @@
...
+                            results.scheme,
+                            re.escape(results.netloc).replace("\\*", "[a-zA-Z0-9_-]+"),
+                        ]
+                        + [re.escape(part).replace("\\*", ".+") for part in results[2:]]
+                    )
+                    patterns.append(re.compile(pattern))
+                endpoint["patterns"] = patterns
+
+            parsed = urlparse(provider_url)
+            self.oembed_providers[re.sub(r"^www\.", "", parsed.netloc)] = provider
```
richvdh, line 84: why do we use `provider_url` as the key rather than `schemes`?

synapse/config/oembed.py
```diff
@@ -0,0 +1,155 @@
...
+                    results = urlparse(s)
+                    pattern = urlunparse(
+                        [
+                            results.scheme,
+                            re.escape(results.netloc).replace("\\*", "[a-zA-Z0-9_-]+"),
+                        ]
+                        + [re.escape(part).replace("\\*", ".+") for part in results[2:]]
+                    )
+                    patterns.append(re.compile(pattern))
+                endpoint["patterns"] = patterns
```
richvdh, line 81: it seems poor style to modify the original `oembed_providers` data structure, as well as being wasteful (we will keep a lot of information in memory that will never be used). I suggest building a completely new structure.

synapse/rest/media/v1/preview_url_resource.py
```diff
@@ -144,6 +101,7 @@ def __init__(
         self.clock = hs.get_clock()
         self.filepaths = media_repo.filepaths
         self.max_spider_size = hs.config.max_spider_size
+        self.oembed = hs.config.oembed
```
richvdh, line 104: Please can you give this a more descriptive name?

```suggestion
        self.oembed_config = hs.config.oembed
```

tests/rest/media/v1/test_url_preview.py
```diff
@@ -79,6 +79,9 @@ def make_homeserver(self, reactor, clock):
 
         config["media_storage_providers"] = [provider_config]
 
+        # Route the HTTP version to an HTTP endpoint so that the tests work.
```
richvdh, line 82: I don't understand this comment?

tests/rest/media/v1/test_url_preview.py
```diff
@@ -544,123 +547,99 @@ def test_accept_language_config_option(self):
...
+            output_stream.write(end_content)
+
+            return (end_content_len, end_headers, url, 200)
+
+        mock_get_file = mock_client.get_file
+        mock_get_file.side_effect = download
+
+        channel = self.make_request(
+            "GET",
+            "preview_url?url={}".format(preview_url),
```
richvdh, line 588:
```suggestion
            f"preview_url?url={preview_url}",
```

tests/rest/media/v1/test_url_preview.py
```diff
@@ -544,123 +547,99 @@ def test_accept_language_config_option(self):
...
-            server = AccumulatingProtocol()
-            server.makeConnection(FakeTransport(client, self.reactor))
-            client.makeConnection(FakeTransport(server, self.reactor))
-            client.dataReceived(
-                (
-                    b"HTTP/1.0 200 OK\r\nContent-Length: %d\r\n"
-                    b'Content-Type: application/json; charset="utf8"\r\n\r\n'
-                )
-                % (len(oembed_content),)
-                + oembed_content
```
richvdh, line 593: I think mocking out the HTTP client is fine, but out of interest, was there a reason you chose to replace this logic with mocks?

richvdh (changes requested):
Thanks very much for your work on this! It's definitely progressing in the right direction, but also I have a few comments below.

tests/rest/media/v1/test_url_preview.py
clokep: This looks like it was my original comment, it was trying to avoid requiring HTTPS.

richvdh: ok, but I don't think it makes sense in its new context.

tests/rest/media/v1/test_url_preview.py
srdjan-catalyst: I could not make FakeTransport work with SSL at all with tests. I spent a day trying to make it work with no success. Other tests are passing because they are checking that certain requesta are attempted, and the fact that they are breaking at the SSL level does not matter for the outcome. In case of oembed we have two consecutive calls, and the second call never happens because the first one breaks.
Urls that come from oembed.com are (almost) all https, so I wanted to keep that.

oembed/providers.json
```diff
@@ -0,0 +1,4130 @@
+[
```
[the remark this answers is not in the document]
srdjan-catalyst: Definitely. Was not aware of the dir.

changelog.d/10536.misc
```diff
@@ -0,0 +1 @@
+Added `oembed_providers_dir` stanza to `homeserver.yaml`, supprted with `OembedConfig` class.
```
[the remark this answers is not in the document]
srdjan-catalyst: I hope new wording is a bit better.

synapse/config/oembed.py
srdjan-catalyst: I have replaced oembed_providers with oembed_endpoints, should be a bit leaner.

clokep:
Hello! I implemented something similar in #10714, I think that meets all the requirements that you were going for, but please shout if not. The changes from that should be in Synapse 1.43.0 (which won't be released for several weeks). Thanks for your contribution!

Issue #10521, opened 2021-08-04
Redacting MSC2716 related events


MadLittleMods:
What should happen when you redact MSC2716 related events? @erikjohnston does this look sane?

 - "insertion" event
    - Undo the steps we take in [`_handle_insertion_event`](https://github.com/matrix-org/synapse/blob/72935b7c5053af122c8cb5767a3e85a3a0f3a20c/synapse/storage/databases/main/events.py#L1760-L1809)
    - Remove from the `insertion_events` table
    - Remove all of the `insertion_event_edges`
 - "chunk" event
    - Undo the steps we take in [`_handle_chunk_event`](https://github.com/matrix-org/synapse/blob/72935b7c5053af122c8cb5767a3e85a3a0f3a20c/synapse/storage/databases/main/events.py#L1811-L1846)
    - Remove from `chunk_events`
 - "marker" event
    - not merged yet, see https://github.com/matrix-org/synapse/pull/10498
    - Undo the steps we take in `_handle_marker_event`
    - We backfill the insertion event in `_handle_marker_event` but I don't think we need to undo that. If they want to redact the insertion event, they can do that separately. We can just redact the "marker" event as normal with no special logic.

---
  
A more general question but what happens when you redact a redaction? Does it cancel the event being redacted?

erikjohnston:
I think we should probably add those fields to be included in the redacted event, so that the structure of the imported history isn't lost. The idea of redaction is that you can use it to strip out fields from events without affecting anything at the protocol level.

> A more general question but what happens when you redact a redaction? Does it cancel the event being redacted?

I don't think anything happens, as it won't strip out the `redacts` field

MadLittleMods:
> The idea of redaction is that you can use it to strip out fields from events without affecting anything at the protocol level.

Interesting. Didn't realize that was the goal of redacting 👨‍🌾 

Created https://github.com/matrix-org/synapse/pull/10538 to maintain the structural fields around 👍

PR #10538, opened 2021-08-05
When redacting, keep event fields around that maintain the historical event structure intact (MSC2716)

MadLittleMods:
When redacting, keep event fields around that maintain the historical event structure intact.

Fix https://github.com/matrix-org/synapse/issues/10521

### Pull Request Checklist


erikjohnston:
This looks good, though does mean that existing events will end up with different event IDs. I think we should probably update the room version ID to avoid mass confusion?

MadLittleMods:
@erikjohnston I'm not following 😣. Does redacting an event give it a different event ID?

These redaction checks are already under the experimental MSC2716 room version? Do we need to bump that version again?

erikjohnston:
> @erikjohnston I'm not following . Does redacting an event give it a different event ID?
> 
> These redaction checks are already under the experimental MSC2716 room version? Do we need to bump that version again?

Ah, sorry, its non obvious how this relates to event IDs. In non-v1 room versions event IDs are the [reference hashes](https://spec.matrix.org/unstable/server-server-api/#calculating-the-reference-hash-for-an-event) for the event, which are calculated by redacting the event, serialising it to canonical JSON and then calculating the SHA256 hash of those bytes. Hence, if you change the redaction algorithm then the canonical JSON bytes will be different and therefore so would the hash.<sup>1</sup>.



> These redaction checks are already under the experimental MSC2716 room version? Do we need to bump that version again?

I would, just to prevent confusion with any existing test rooms using that experimental version. I hope that's fine and easy?





---
<sup>1</sup> Note that an event also contains a [content hash](https://spec.matrix.org/unstable/server-server-api/#calculating-the-content-hash-for-an-event), which is a hash over the *un*redacted event and is included in the redacted event. This is what protects the unredacted event fields from being changed.

MadLittleMods:
> Ah, sorry, its non obvious how this relates to event IDs. In non-v1 room versions event IDs are the [reference hashes](https://spec.matrix.org/unstable/server-server-api/#calculating-the-reference-hash-for-an-event) for the event, which are calculated by redacting the event, serialising it to canonical JSON and then calculating the SHA256 hash of those bytes. Hence, if you change the redaction algorithm then the canonical JSON bytes will be different and therefore so would the hash1.

Thanks for that context! That makes sense 👍 

---

Do we actually verify the `event_id` hashes after it's created? I assume this only matters for remote federated homeservers trying to verify a received event?

Just as a thought, if redacting an event messes up the event verification, that essentially `soft_fails` the event which has the same effect as redacting 🤔 (not visible for client)

I understand why it matters, just asking a "technically, does it work? 🤓" kind of question 😁

erikjohnston:
> Do we actually verify the `event_id` hashes after it's created? I assume this only matters for remote federated homeservers trying to verify a received event that was redacted?

Amusingly, yes, as the Event object calculates the `event_id` on creation, rather than reading it from the DB. The `get_events(..)` function returns the event as part of an `event.event_id -> event` mapping, so if you ask for event `$A` you'd get back a dict with `$B` in instead, which'd then cause the code to assume the server didn't have `$A`

> Just as a thought, if redacting an event messes up the event verification, that essentially `soft_fails` the event which has the same effect as redacting  (not visible for client)
> 
> I understand why it matters, just asking a "technically, does it work? " kind of question

It will reject the event rather than "soft fail", but yes it'd have the same impact as redacting the event. I think not changing the redaction rules would mostly be fine in practice (we have in the past forgotten to change the redaction algorithm when we've changed the auth rules).

erikjohnston (approved):

changelog.d/10538.feature
```diff
@@ -0,0 +1 @@
+When redacting, keep event fields around that maintain the MSC2716 historical event structure intact.
```
erikjohnston, line 1:
```suggestion
Add support for new redaction rules for historical events specified in [MSC2716](https://github.com/matrix-org/matrix-doc/pull/2716).
```

MadLittleMods:
> Amusingly, yes, as the Event object calculates the `event_id` on creation, rather than reading it from the DB. The `get_events(..)` function returns the event as part of an `event.event_id -> event` mapping, so if you ask for event `$A` you'd get back a dict with `$B` in instead, which'd then cause the code to assume the server didn't have `$A`

😁 That would be an interesting scenario

---

Thanks for the review @erikjohnston 🐘

Issue #9595, opened 2021-03-11
rejected event used as an auth event


richvdh:
in my database, `$h+BUjV0LuuSRmFh5ZFeSVZkP+oo6v7bhyJ+1wYjpLb4` (a regular membership event) uses `$NncVOvrPzkKl0u1Q8FIJ7Pz1kZH1iF2lwWj6pZ6+Kkg` (a membership event which was *rejected* due to missing auth events) as an auth event. This seems wrong: `$h+BUjV0LuuSRmFh5ZFeSVZkP+oo6v7bhyJ+1wYjpLb4` should have been rejected too.

Logs from the arrival of that event:

```
2021-03-08 16:00:19,930 - synapse.handlers.federation - 189 - INFO - PUT-2238636-$h+BUjV0LuuSRmFh5ZFeSVZkP+oo6v7bhyJ+1wYjpLb4 - handling received PDU: <FrozenEventV2 event_id='$h+BUjV0LuuSRmFh5ZFeSVZkP+oo6v7bhyJ+1wYjpLb4', type='m.room.member', state_key='@yan:yetanothernerd.xyz'>
2021-03-08 16:00:20,090 - synapse.handlers.federation - 2406 - INFO - PUT-2238636-$h+BUjV0LuuSRmFh5ZFeSVZkP+oo6v7bhyJ+1wYjpLb4 - auth_events refers to events which are not in our calculated auth chain: {'$NncVOvrPzkKl0u1Q8FIJ7Pz1kZH1iF2lwWj6pZ6+Kkg'}
2021-03-08 16:00:20,090 - synapse.state - 452 - INFO - PUT-2238636-$h+BUjV0LuuSRmFh5ZFeSVZkP+oo6v7bhyJ+1wYjpLb4 - Resolving state for !BAXLHOFjvDKUeLafmO:matrix.org with 2 groups
2021-03-08 16:00:20,091 - synapse.handlers.federation - 2446 - INFO - PUT-2238636-$h+BUjV0LuuSRmFh5ZFeSVZkP+oo6v7bhyJ+1wYjpLb4 - After state res: updating auth_events with new state {}
[...]

PR #10539, opened 2021-08-05
Clean up federation event auth code

richvdh:
This is part of my work trying to sort out #9595. As a first step, this should be an (almost) non-functional change to try to make these code paths a little more understandable.

Reviewable commit by commit.

### DIFF
--- a/synapse/handlers/federation.py
+++ b/synapse/handlers/federation.py
@@ -2208,7 +2208,7 @@ class FederationHandler(BaseHandler):
         event: EventBase,
         context: EventContext,
         state: Optional[Iterable[EventBase]] = None,
-        auth_events: Optional[MutableStateMap[EventBase]] = None,
+        auth_events: Optional[StateMap[EventBase]] = None,
         backfilled: bool = False,
     ) -> None:
         """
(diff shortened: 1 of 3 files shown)


synapse/handlers/federation.py
```diff
@@ -2611,16 +2611,6 @@ async def _check_event_auth(
...
 
-        # This is a hack to fix some old rooms where the initial join event
-        # didn't reference the create event in its auth events.
-        if event.type == EventTypes.Member and not event.auth_event_ids():
-            if len(event.prev_event_ids()) == 1 and event.depth < 5:
-                c = await self.store.get_event(
-                    event.prev_event_ids()[0], allow_none=True
-                )
-                if c and c.type == EventTypes.Create:
-                    auth_events[(c.type, c.state_key)] = c
```
erikjohnston, line 2622: What's the reasoning for not needing this anymore? Such old rooms may still be floating around? Possibly an alternative is to only do this for v1 rooms?

erikjohnston (commented):
Otherwise LGTM!

synapse/handlers/federation.py
richvdh: This workaround was added in https://github.com/matrix-org/synapse/commit/858e87ab0d1bd88fcd5c819c482a40130e571874, from which I infer it was fixed before then. This is only a week after ae9c2ab1652511b36dc594882e88152b6067c816, which we know was a critical security fix. I'm pretty sure we previously checked that there weren't any remaining rooms affected by the SYN-149 bug, but if there are any other rooms that predate that fix... I don't think I want to join them.

erikjohnston: Hmm, OK fair

erikjohnston (approved):

PR #10542, opened 2021-08-05
Convert Transation and Edu object to attrs

clokep:
This was requested at https://github.com/matrix-org/synapse/pull/10408#discussion_r671092527, instead of adding many `ignore` comments for the type of `Transaction` and `Edu` properties this converts them to attrs classes and adds a bit more structure around them.

I think this does have a minor change in that if you previously could create a `Transaction` object and then call `get_dict()` on it to get back arbitrary keys. The fields of `Transaction` are now limited.

This should be reviewable commit-by-commit and includes some clean-up first.


synapse/federation/federation_server.py
```diff
@@ -212,18 +216,22 @@ async def on_incoming_transaction(
         # accurate as possible.
         request_time = self._clock.time_msec()
 
-        transaction = Transaction(**transaction_data)
-        transaction_id = transaction.transaction_id  # type: ignore
+        transaction = Transaction(
+            transaction_id=transaction_id,
+            destination=destination,
+            origin=origin,
```
clokep, line 222: Previously we pulled `origin` from `transaction_data`, but the comment in `synapse.federation.transport.server` makes me think we should be using the origin we've calculated instead.

synapse/federation/sender/transaction_manager.py
```diff
@@ -104,13 +104,13 @@ async def send_new_transaction(
...
 
-            transaction = Transaction.create_new(
+            transaction = Transaction(
                 origin_server_ts=int(self.clock.time_msec()),
                 transaction_id=txn_id,
                 origin=self._server_name,
                 destination=destination,
-                pdus=pdus,
-                edus=edus,
+                pdus=[p.get_pdu_json() for p in pdus],
```
erikjohnston, line 113: This will interfere with magic below where we serialise the events *just* before sending out the request, this is for the `age` stuff.

clokep: I'm not sure I follow, this moved logic out of `create_new` into the caller, so it should be identical.

erikjohnston: Oh, hmm, you're right. I'm just confused then. Ignore me. Nothing to see here. Lalalalala.

erikjohnston (approved):

PR #10547, opened 2021-08-06
PreviewUrlResource: set `od:iframe` in json (#2752)

srdjan-catalyst:
Set `od:iframe` in json when oembed returns `<iframe/>`
    
    Signed-off-by: Srdjan <srdjan@catalyst.net.nz>

### Pull Request Checklist


clokep:
@srdjan-catalyst Can you expand upon why this is necessary / what this does?

srdjan-catalyst:
Currently, when oembed api returns html, it is parsed and as far as content is concerned, only images are being looked for. We have a case when an iframe is sent back to display content. That is because of the content being available to authenticated users only so it has to be dealt with on the client side.

synapse/rest/media/v1/preview_url_resource.py
```diff
@@ -876,6 +876,12 @@ def _calc_og(tree: "etree.Element", media_uri: str) -> Dict[str, Optional[str]]:
         assert isinstance(og["og:description"], str)
         og["og:description"] = summarize_paragraphs([og["og:description"]])
 
+    iframe = tree.xpath("//iframe")
+    if iframe:
+        og["og:iframe"] = iframe[0].attrib["src"]
+        og["og:iframe:width"] = iframe[0].attrib.get("width", "100%")
+        og["og:iframe:height"] = iframe[0].attrib.get("height")
```
richvdh, line 883: could you give me a link to where these opengraph properties are specified?

srdjan-catalyst: They are not. Here we have oembed translated to og. But there's no og for iframe. I could set type to iframe and set url. I did it this way because it makes it obviously bogus. Setting type and url may give a false impression that it is something universally supported, whereas I just want to signal an iframe to the front end.

richvdh: I'm afraid I don't think that Synapse ought to be generating custom properties here which will only be meaningful to your client, and in your specific usecase.

srdjan-catalyst: Well I was thinking of supporting that in element. Any suggestions how to push oembed iframes to the client please?

richvdh: Such an extension would need to be part of the spec, so that it can be implemented by other servers and clients - see https://matrix.org/docs/spec/client_server/r0.6.1#get-matrix-media-r0-preview-url where we already have a `matrix:image:size` custom property.

Changes to the spec generally need a proposal - https://spec.matrix.org/unstable/proposals/.

srdjan-catalyst: https://github.com/matrix-org/matrix-doc/pull/3338

I hope I didn't go horribly wrong. Is there anything else I need to do?
I saw something about `/unstable` endpoints, but didn't quite get if I can use it for this.

erikjohnston:
This appears blocked on https://github.com/matrix-org/matrix-doc/pull/3338 now?

PR #10548, opened 2021-08-06
Port the Password Auth Providers module interface to the new generic interface

Azrenbeth:
Port the Password Auth Providers module interface to the new generic interface introduced in v1.37.0


babolivier (changes requested):
FTR I haven't looked at the tests and docs yet, since I feel like there are some changes to be made in the API that need to be addressed before I can look at them.

synapse/handlers/auth.py
```diff
@@ -1214,33 +1198,43 @@ async def _validate_userid_login(
...
-            supported_login_types = provider.get_supported_login_types()
-            if login_type not in supported_login_types:
-                # this password provider doesn't understand this login type
-                continue
-
+        # get all of the login_types registered by modules
+        supported_login_types = self.password_auth_provider.get_supported_login_types()
+        # check if the login type being used is supported by a module
+        # note that if LoginType.PASSWORD has been disabled then login_type has already been
+        # checked for that by this point
```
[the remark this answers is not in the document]
Azrenbeth: So there's a config option that disables password authentication. I mean that we don't need to remove LoginType.PASSWORD from the list types that modules support because by this point in the code, we know that the variable `login_type` is not LoginType.PASSWORD

What would be a better way of phrasing this?

synapse/handlers/auth.py
```diff
@@ -1871,63 +2050,130 @@ async def check_auth(
...
-        # Check if the return value is a str or a tuple
-        if isinstance(result, str):
-            # If it's a str, set callback function to None
-            return result, None
+            if result is not None:
+                # Check that the callback returned a Tuple[str, Optional[Callable]]
+
+                # "type: ignore" is used on the isinstance checks because mypy thinks
+                # result is always the right type, but as it is 3rd party code it might not be
+                if not isinstance(result, Tuple) or len(result) != 2:  # type: ignore
```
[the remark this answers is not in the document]
Azrenbeth: These are all arg-type errors but yes have changed them all to`# type: ingore[arg-type]`

synapse/handlers/auth.py
```diff
@@ -1805,59 +1801,242 @@ def _generate_base_macaroon(self, user_id: str) -> pymacaroons.Macaroon:
...
+
+        # Mapping from login type to auth checker callbacks
+        self.auth_checker_callbacks: Dict[str, List[CHECK_AUTH_CALLBACK]] = {}
+
+    def register_password_auth_provider_callbacks(
+        self,
+        check_3pid_auth: Optional[CHECK_3PID_AUTH_CALLBACK] = None,
+        on_logged_out: Optional[ON_LOGGED_OUT_CALLBACK] = None,
+        supported_login_types: Optional[Dict[str, Iterable[str]]] = None,
+        auth_checkers: Optional[Dict[str, CHECK_AUTH_CALLBACK]] = None,
```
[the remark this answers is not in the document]
Azrenbeth: Yes I think it would make more sense to register them as `Dict[Tuple[str, Iterable[str]], CHECK_AUTH_CALLBACK]` 

However I'm not really sure about splitting them up into two different register methods: for example in which one do you put your callback for `check_3pid_auth`? It is an *auth checker* but it's meant to handle just 3pid authentication and so is the same format...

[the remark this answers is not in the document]
babolivier: > However I'm not really sure about splitting them up into two different register methods: for example in which one do you put your callback for `check_3pid_auth`? It is an auth checker but it's meant to handle just 3pid authentication and so is the same format...

That's a good point. Maybe we should change how the call stack around auth checkers work to allow 3pid auth to be checked in the same callback format as other checkers, but that's definitely out of scope for this PR.

docs/modules.md
```diff
@@ -329,6 +329,96 @@ For example, if the user `@alice:example.org` is passed to this method, and the
...
+) -> Optional[Tuple[str, Optional[Callable]]]
+```
+
+It is passed the user field provided by the client (which might not be in `@username:server` form), 
+the login type, and a dictionary of login secrets passed by the client.
+
+On a failure it should return None. On a success it should return a (user_id, callback) tuple
+where user_id is a str containing the user's matrix id (i.e. in `@username:server` form) and the 
+callback is a method to be called with the result from the `/login` call (see the 
+<a href="https://spec.matrix.org/unstable/client-server-api/#login">spec</a>
```
babolivier, line 384: I think that the return type could be described in a less confusing way, eg:

> If the authentication is successful, the module must return the user's Matrix ID (e.g. `@alice:example.com`) and optionally a callback to be called with the response to the <a href="https://spec.matrix.org/unstable/client-server-api/#login">`/login`</a> request. If the module doesn't wish to return a callback, it must return `None` instead.

It would also be good to document what these callbacks should look like.

docs/modules.md
```diff
@@ -329,6 +329,96 @@ For example, if the user `@alice:example.org` is passed to this method, and the
...
+
+```python
+async def check_3pid_auth(
+    medium: str, 
+    address: str,
+    password: str,
+) -> Optional[Tuple[str, Optional[Callable]]]
+```
+
+This  is called when a user attempts to register or log in with a third party identifier,
```
babolivier, line 397:
```suggestion
Called when a user attempts to register or log in with a third party identifier,
```

docs/modules.md
```diff
@@ -329,6 +329,96 @@ For example, if the user `@alice:example.org` is passed to this method, and the
...
+such as email. It is passed the medium (eg. "email"), an address (eg. "jdoe@example.com")
+and the user's password.
+
+On a failure it should return None. On a success it should return a (user_id, callback) tuple
+where user_id is a str containing the user's matrix id (i.e. in `@username:server` form) and the 
+callback is a method to be called with the result from the `/login` call (see the 
+<a href="https://spec.matrix.org/unstable/client-server-api/#login">spec</a> 
+for details). This callback can be None.
```
babolivier, line 405: Same comment.

docs/modules.md
```diff
@@ -329,6 +329,96 @@ For example, if the user `@alice:example.org` is passed to this method, and the
...
+
+```python
+async def on_logged_out(
+        user_id: str,
+        device_id: str,
+        access_token: str,
+) -> None
```
babolivier, line 412: The indentation looks off. Also it looks like this should be able to fit on a single line.

docs/modules.md
```diff
@@ -329,6 +329,96 @@ For example, if the user `@alice:example.org` is passed to this method, and the
...
+``` 
+This is called when a user logs out. It is passed the qualified user ID, the ID of the
```
babolivier, line 414:
```suggestion
Called when a user logs out. It is passed the qualified user ID, the ID of the
```

docs/modules.md
```diff
@@ -329,6 +329,96 @@ For example, if the user `@alice:example.org` is passed to this method, and the
...
+deactivated device (if any: access tokens are occasionally created without an associated
+device ID), and the (now deactivated) access token.
+
+The callback must be asynchronous but the logout request will wait for the callback
+to complete.
```
babolivier, line 419: Not sure that's worth mentioning, since that's the case for all callbacks. The point of having callbacks being async isn't for Synapse to fire and forget them, but rather that if they end up waiting for a resource they don't block other coroutines (e.g. other requests being handled).

docs/modules.md
```diff
@@ -347,6 +437,22 @@ The module's author should also update any example in the module's configuration
...
 for more info).
 
+#### Porting password auth provider modules
+
+There is no longer a `get_db_schema_files` callback provided, any changes to the database
+should now be made by the module using the module API class.
+
+To port a module that has a `check_password` method: 
+- Register `("m.login.password", ("password",): self.check_password` as an auth checker
+- Set `self.module_api` to point to the `ModuleApi` object given in `__init__`
```
babolivier, line 447: Why is this step necessary? It shouldn't matter how a module calls the module API instance it's given.

docs/modules.md
```diff
@@ -347,6 +437,22 @@ The module's author should also update any example in the module's configuration
...
+- Change the arguments of `check_password` to
+  `(username: str, login_type: str, login_dict: JsonDict)`
+- Add the following lines to the top of the `check_password` method:
+  - `user_id = self.module_api.get_qualified_user_id(username)`
+  - `password = login_dict["password"]`
+- Alter `check_password` so that it returns `None` on an authentication failure, and `user_id`
+  on a success
```
babolivier, line 454: I think this, and the bullet points above, can be summed up into:

> If the module implements the `check_password` callback, modify it so it complies with the auth checker signature, and register it as an auth checker with `("m.login.password", ("password",))` as the login type and fields. The callback can access the user's full Matrix ID using the `ModuleApi`'s `get_qualified_user_id` method, and can access the provided password by looking for the `password` key in the provided `login_dict` argument.

babolivier (changes requested):
Getting there!

docs/modules/password_auth_provider_callbacks.md
```diff
@@ -0,0 +1,155 @@
...
+Password auth providers offer a way for server administrators to integrate
+their Synapse installation with an external authentication system. The callbacks can be
+registered by using the Module API's `register_password_auth_provider_callbacks` method.
+
+## Callbacks
+
+### `auth_checkers`
+
+```
+ auth_checkers: Dict[Tuple[str,Tuple], CHECK_AUTH_CALLBACK]
```
babolivier, line 12: I think this line is confusing to people not familiar with this portage work.

I would suggest:

* replacing `CHECK_AUTH_CALLBACK` with `Callable` (which is the generic Python type for a function)
* moving the example at line 41 (and the few paragraphs below it) up to ~line 18 (and having that paragraph end with something like "An authentication checking callback should be of the following form:")

docs/modules/password_auth_provider_callbacks.md
```diff
@@ -0,0 +1,155 @@
...
+    - Is checked by the method: `self.password_checker` 
+    
+Then it should register the following dict:
+
+```python
+{ 
+    ("com.example.custom_login", ("secret1", "secret2")): self.check_custom_login,
+    ("m.login.password", ("password",)): self.password_checker,
+}
+```
```
babolivier, line 37: I feel like this could be all replaced with a link to https://matrix.org/docs/spec/client_server/latest#user-interactive-authentication-api (especially since we've got an example below on how it translates into code, so the additional example isn't strictly necessary)

docs/modules/password_auth_provider_callbacks.md
```diff
@@ -0,0 +1,155 @@
...
+        Optional[Callable[["synapse.module_api.LoginResponse"], Awaitable[None]]]
+    ]
+]
+```
+
+Called when a user attempts to register or log in with a third party identifier,
+such as email. It is passed the medium (eg. "email"), an address (eg. "jdoe@example.com")
+and the user's password.
+
+If the authentication is successful, the module must return the user's Matrix ID (e.g. @alice:example.com) and optionally a callback to be called with the response to the `/login` request. If the module doesn't wish to return a callback, it must return None instead.
```
babolivier, line 80: Line wrap please

babolivier: Also:

```suggestion
If the authentication is successful, the module must return the user's Matrix ID (e.g. `@alice:example.com`) and optionally a callback to be called with the response to the `/login` request. If the module doesn't wish to return a callback, it must return None instead.
```

docs/modules/password_auth_provider_callbacks.md
```diff
@@ -0,0 +1,155 @@
...
+) -> Optional[
+    Tuple[
+        str, 
+        Optional[Callable[["synapse.module_api.LoginResponse"], Awaitable[None]]]
+    ]
+]
+```
+
+Called when a user attempts to register or log in with a third party identifier,
+such as email. It is passed the medium (eg. "email"), an address (eg. "jdoe@example.com")
```
babolivier, line 77:
```suggestion
such as email. It is passed the medium (eg. `email`), an address (eg. `jdoe@example.com`)
```

docs/modules/password_auth_provider_callbacks.md
```diff
@@ -0,0 +1,155 @@
...
+    user_id: str,
+    device_id: Optional[str],
+    access_token: str
+) -> None
+``` 
+Called during a logout request for a user. It is passed the qualified user ID, the ID of the
+deactivated device (if any: access tokens are occasionally created without an associated
+device ID), and the (now deactivated) access token.
+
+## Example
```
babolivier, line 97: Could you add a few lines to explain what this example does please? So the reader knows what's going on, like eg https://matrix-org.github.io/synapse/develop/modules/spam_checker_callbacks.html#example

synapse/handlers/auth.py
```diff
@@ -1808,59 +1806,244 @@ def _generate_base_macaroon(self, user_id: str) -> pymacaroons.Macaroon:
...
+            self.on_logged_out_callbacks.append(on_logged_out)
+
+        if auth_checkers is not None:
+            # register a new supported login_type
+            # Iterate through all of the types being registered
+            for (login_type, fields), callback in auth_checkers.items():
+                # Note: fields may be empty here. This would allow a modules auth checker to
+                # be called with just 'login_type' and no password or other secrets
+
+                # Need to check that all the field names are strings or will get nasty error later
```
babolivier, line 2002:
```suggestion
                # Need to check that all the field names are strings or will get nasty errors later
```

synapse/handlers/auth.py
```diff
@@ -1808,59 +1806,244 @@ def _generate_base_macaroon(self, user_id: str) -> pymacaroons.Macaroon:
...
+                    if fields_currently_supported != fields:
+                        raise RuntimeError(
+                            "A module tried to register support for login type: %s with parameters %s"
+                            " but another module had already registered support for that type with parameters %s"
+                            % (login_type, fields, fields_currently_supported)
+                        )
+
+                # Add the new method to the list of auth_checker_callbacks for this login type
+                callback_list = self.auth_checker_callbacks.get(login_type, [])
+                callback_list.append(callback)
```
babolivier, line 2029: Bit of a nitpick, but I think you can do this as a one-liner:

```suggestion
                self.auth_checker_callbacks.setdefault(login_type, []).append(callback)
```

babolivier (changes requested):
Code looks good, just needs to get the documentation in shape!

docs/modules/password_auth_provider_callbacks.md
Azrenbeth: email and jdoe@example.com are strings - why should it use `...` instead of "..."?

babolivier: It wasn't but it works too :)

docs/modules/password_auth_provider_callbacks.md
```diff
@@ -0,0 +1,151 @@
...
+defines some types, however user defined ones are also allowed.
+
+It is passed the user field provided by the client (which might not be in `@username:server` form), 
+the login type, and a dictionary of login secrets passed by the client.
+
+If the authentication is successful, the module must return the user's Matrix ID (e.g. 
+`@alice:example.com`) and optionally a callback to be called with the response to the `/login` request.
+If the module doesn't wish to return a callback, it must return None instead.
+
+If the authentication is unsuccessful, the module must return None.
```
babolivier, line 43:
```suggestion
If the authentication is unsuccessful, the module must return `None`.
```

PR #10549, opened 2021-08-06
Add local support for the new spaces summary endpoint

clokep:
Adds updated support for [MSC2946](https://github.com/matrix-org/matrix-doc/pull/2946) -- the `/hierarchy` endpoint and the various parameters for it. This does not yet reach out over federation or support workers.

Step 2 of #10495

See matrix-org/complement#187 for tests.


synapse/handlers/space_summary.py
```diff
@@ -67,6 +69,12 @@ def __init__(self, hs: "HomeServer"):
         self._server_name = hs.hostname
         self._federation_client = hs.get_federation_client()
 
+        # TODO Allow for multiple workers to share this data.
+        # TODO Expire pagination tokens.
+        self._pagination_sessions: Dict[
+            str, Tuple[Deque[_RoomQueueEntry], Set[str]]
+        ] = {}
```
richvdh: could do with a comment describing the contents? (Or possibly a proper attr.s class rather than a tuple?)

synapse/handlers/space_summary.py
```diff
@@ -278,13 +286,19 @@ async def get_room_hierarchy(
...
-        # TODO Handle pulling previously persisted state by the pagination token.
+        # If this is continuing a previous session, pull the persisted data.
+        if from_token:
+            if from_token not in self._pagination_sessions:
+                raise SynapseError(400, "Unknown pagination token", Codes.INVALID_PARAM)
 
-        # the queue of rooms to process
-        room_queue = deque((_RoomQueueEntry(room_id, ()),))
+            # TODO Assert that other parameters have not changed.
+            (room_queue, processed_rooms) = self._pagination_sessions[from_token]
```
richvdh: I think this is going to go fairly badly wrong when we get two concurrent requests with the same `from_token`, and both end up popping entries from the same Deque. We might need a ResponseCache to provide locking? Or to copy the deque so that we can safely process both requests?

synapse/handlers/space_summary.py
```diff
@@ -69,10 +69,16 @@ def __init__(self, hs: "HomeServer"):
...
+        # A map of query information to the current state, for pagination.
+        #
+        # The query information is a tuple of the room ID, suggested-only, and
+        # max-depth fields, followed by the pagination token.
+        #
         # TODO Allow for multiple workers to share this data.
         # TODO Expire pagination tokens.
         self._pagination_sessions: Dict[
-            str, Tuple[Deque[_RoomQueueEntry], Set[str]]
+            Tuple[str, bool, Optional[int], str],
```
richvdh, line 80: probably ought to consider an attr.s for the session key, too.

richvdh (approved):
looks grand. A few suggestions here but no showstoppers.

synapse/handlers/space_summary.py
clokep: A response cache sounds like it would be good regardless, since this could be a somewhat heavy calculation?

PR #10552, opened 2021-08-07
Only return state events that the AS passed in via `state_events_at_start` (MSC2716)

MadLittleMods:
Update `/batch_send` endpoint to only return `state_events` created by the `state_events_from_before` passed in.

As discovered by @Half-Shot in
https://github.com/matrix-org/matrix-doc/pull/2716#discussion_r684158448

Part of [MSC2716](https://github.com/matrix-org/matrix-doc/pull/2716)

### Pull Request Checklist


synapse/rest/client/v1/room.py
```diff
@@ -438,6 +438,7 @@ async def on_POST(self, request, room_id):
         prev_state_ids = list(prev_state_map.values())
```
erikjohnston, line 438: I've just realised that we should probably not be adding this class to `v1/room.py`, but instead somewhere in `v2/`. If only someone had reviewed the PR that added this endpoint.................. :facepalm:

erikjohnston (approved):

synapse/rest/client/v1/room.py
MadLittleMods: I can switch it over in another PR 😁

MadLittleMods:
@erikjohnston Thanks for the review and noticing the endpoint being organized in the wrong area 🦑

synapse/rest/client/v1/room.py
MadLittleMods: https://github.com/matrix-org/synapse/pull/10576 ⏩

PR #10556, opened 2021-08-09
Fix documentation of directory name for remote thumbnails

cuttingedge1109:
The directory name for remote thumbnails are specified as `/remote_thumbnails` in Readme but `/remote_thumbnail` in the code. It seems a typo.

### DIFF
--- a/changelog.d/10556.doc
+++ b/changelog.d/10556.doc
@@ -1,1 +1,1 @@
-Change the directory name for remote thumbnails from `remote_thumbnails` to `remote_thumbnail` in  media_repository.md file to match the code.
+Minor fix to the `media_repository` developer documentation. Contributed by @cuttingedge1109.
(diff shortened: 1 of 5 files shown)


clokep:
I think we probably want to update the README if it is wrong, as changing the code will likely break any existing deploys.

richvdh:
@cuttingedge1109 are you still interested in working on this?

cuttingedge1109:
I wanted to get feedbacks from other guys also.
I can change the README but I prefer to use the same naming convention.

richvdh:
consistency would be nice, but breaking existing deployments is a greater concern.

cuttingedge1109:
Ok, then let me change the readme.

cuttingedge1109:
Ok, reverted the code change.
I suggest that we can change the code for consistent naming in the major or minor version upgrade.

changelog.d/10556.doc
```diff
@@ -0,0 +1 @@
+Change the directory name for remote thumbnails from `remote_thumbnails` to `remote_thumbnail` in  media_repository.md file to match the code.
```
richvdh, line 1:
```suggestion
Minor fix to the `media_repository` developer documentation. Contributed by @cuttingedge1109.
```

richvdh (approved):
thank you!

PR #10560, opened 2021-08-09
Fix an edge-case for being invited to a room for the spaces summary.

clokep:
I split this out from the (upcoming) federation PR since it is a refactoring + unrelated bugfix.

* It is moved from inline to a separate method since it will be re-used with the updates to the spaces summary API (#10495).
* It fixes an edge-case when a room you're invited to is directly returned over federation.

### DIFF
--- a/tests/handlers/test_space_summary.py
+++ b/tests/handlers/test_space_summary.py
@@ -586,8 +586,11 @@ class SpaceSummaryTestCase(unittest.HomeserverTestCase):
 
     def test_fed_invited(self):
         """
-        Rooms returned over federation should be properly filtered to only include
-        rooms the user has access to.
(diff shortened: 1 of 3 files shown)


tests/handlers/test_space_summary.py
```diff
@@ -570,3 +583,55 @@ async def summarize_remote_room(
                 (subspace, joined_room),
             ],
         )
+
+    def test_fed_invited(self):
+        """
+        Rooms returned over federation should be properly filtered to only include
+        rooms the user has access to.
```
richvdh, line 590: this is the same as the docstring on `test_fed_filtering` - I suspect it has been c&p'ed and could do with an update?

richvdh (approved):
lgtm otherwise!

PR #10561, opened 2021-08-09
Display an error during failure of fallback UIA

govynnus:

See discussion in https://github.com/matrix-org/synapse/pull/10142#discussion_r661613726
This doesn't fix anything existing because none of the current fallbacks really fail, but is required for decent UX with the registration token fallback in #10142


synapse/handlers/auth.py
```diff
@@ -627,23 +627,28 @@ async def check_ui_auth(
 
     async def add_oob_auth(
         self, stagetype: str, authdict: Dict[str, Any], clientip: str
-    ) -> bool:
+    ) -> Any:
```
clokep, line 630: It looks like the result of this isn't even used anymore after this?

If it is, I'd like to see a more specific type here than `Any`. If it isn't, we should have it return `None` and properly document what it raises.

govynnus: Yep, I'll do that

govynnus: How's that?

synapse/handlers/auth.py
```diff
@@ -627,23 +627,28 @@ async def check_ui_auth(
...
-            raise LoginError(400, "", Codes.MISSING_PARAM)
+            raise LoginError(
+                400, f"Unknown UIA stage type: {stagetype}", Codes.INVALID_PARAM
+            )
         if "session" not in authdict:
-            raise LoginError(400, "", Codes.MISSING_PARAM)
+            raise LoginError(400, "Missing session ID", Codes.MISSING_PARAM)
 
+        # If authentication fails a LoginError is raised, so if we do get a
+        # result we know authentication has succeeded.
```
clokep, line 647:
```suggestion
        # If authentication fails a LoginError is raised. Otherwise, store
        # the successful result.
```

synapse/handlers/ui_auth/checkers.py
```diff
@@ -49,7 +49,7 @@ async def check_auth(self, authdict: dict, clientip: str) -> Any:
             clientip: The IP address of the client.
 
         Raises:
-            SynapseError if authentication failed
+            LoginError if authentication failed.
 
         Returns:
             The result of authentication (to pass back to the client?)
```
clokep, line 55: I don't suppose we can clarify this comment while we're looking at all of this!

synapse/rest/client/v2_alpha/auth.py
```diff
@@ -95,29 +95,32 @@ async def on_POST(self, request, stagetype):
...
+                    LoginType.RECAPTCHA, authdict, request.getClientIP()
+                )
+            except LoginError as e:
+                # Authentication failed, let user try again
                 html = self.recaptcha_template.render(
                     session=session,
                     myurl="%s/r0/auth/%s/fallback/web"
                     % (CLIENT_API_PREFIX, LoginType.RECAPTCHA),
                     sitekey=self.hs.config.recaptcha_public_key,
+                    error=e.msg,
```
clokep, line 109: Are we sure that these error messages are suitable for displaying to the user? Looking through the checkers I see a few that raise `LoginError` with a blank message.

clokep (changes requested):
Looks better overall! Anything we can do to make this code clearer is great!

clokep:
>  This doesn't fix anything existing because none of the current fallbacks really fail, but is required for decent UX with the registration token fallback in #10142

I'm not sure I agree. It looks like a broken client (sending the wrong parameters) or if the recaptcha endpoint was down could cause ugly errors to pop back into the user's face. Is that what this is improving?

Could you include a screenshot of how the error renders, by the way?

synapse/handlers/ui_auth/checkers.py
govynnus: Clarify in what way? Say why/how LoginError is used?

clokep: > (to pass back to the client?)

It seems whoever wrote this comment wasn't sure why we care about this. If you've groked this and know the answer it would be great to say *why* we care about the result of the authentication.

govynnus: The only checkers that return something other than just `True` are the threepid ones:
https://github.com/matrix-org/synapse/blob/8c654b73095a594b36101aa81cf91a8e1bebc29f/synapse/handlers/ui_auth/checkers.py#L206

And the only place I can see the result is accessed is here, ~~but I have no clue where the user ID would get into the result field of a UIA stage.~~ It looks like the [password provider docs](https://matrix-org.github.io/synapse/latest/password_auth_providers.html#optional-methods) (which you wrote?) mention that 
https://github.com/matrix-org/synapse/blob/8c654b73095a594b36101aa81cf91a8e1bebc29f/synapse/handlers/auth.py#L377-L405

Where `result` comes from the only ever use of `get_completed_ui_auth_stages`:
https://github.com/matrix-org/synapse/blob/8c654b73095a594b36101aa81cf91a8e1bebc29f/synapse/handlers/auth.py#L605

There's also a section in the SSO handler which sets the result field, but I'm not sure if that's ever looked at later:
https://github.com/matrix-org/synapse/blob/8c654b73095a594b36101aa81cf91a8e1bebc29f/synapse/handlers/sso.py#L690-L704

I am currently using the fact that the result is stored for expiring registration tokens, but I think I could use the UIA session data instead.
https://github.com/govynnus/synapse/blob/e9435f8f6c9ba1497f689e3963c9c005c1d71dfc/synapse/handlers/ui_auth/checkers.py#L294
https://github.com/govynnus/synapse/blob/e9435f8f6c9ba1497f689e3963c9c005c1d71dfc/synapse/storage/databases/main/ui_auth.py#L333-L373

So basically no, I don't grok it.
Although I'm pretty sure it's not passed back to the client :-)

clokep: Hah, thanks for taking another look! We can just leave it as is then!

clokep:
@govynnus Is this ready for another round of review? (Also please take a look at https://github.com/matrix-org/synapse/pull/10561#issuecomment-896770861)

synapse/rest/client/v2_alpha/auth.py
govynnus: Yeah you're right, the errors are more for devs and people troubleshooting than users I think. From what I can tell the errors would be caused by a broken client rather than the user.

clokep: I think we should at least give some string back for those cases so that people can ask a server admin about it. Right now some of them will end up just saying `Error: `.

govynnus: I added a message to the couple LoginErrors that were missing one https://github.com/matrix-org/synapse/pull/10561/commits/8d6a469a43e276a40862eb8a1f2993717572320e

docs/upgrade.md
```diff
@@ -112,6 +112,16 @@ environment variable.
 See [using a forward proxy with Synapse documentation](setup/forward_proxy.md) for
 details.
 
+## UIA fallback templates can display errors
```
clokep, line 115: Let's be explicit:

```suggestion
## User-interactive authentication fallback templates can now display errors
```

changelog.d/10561.bugfix
```diff
@@ -0,0 +1 @@
+Handle failure of User-Interactive Authentication when using a fallback. Contributed by Callum Brown.
```
clokep, line 1: Is this handling the failure or displaying an error? It seems it was previously "handled", but by displaying a JSON error message to the user?

docs/upgrade.md
```diff
@@ -112,6 +112,16 @@ environment variable.
...
+
+This may affect you if you make use of custom HTML templates for the
+[reCAPTCHA](../synapse/res/templates/recaptcha.html) and
+[terms](../synapse/res/templates/terms.html) fallback pages.
+There is no configuration option for these templates.
+
+The template can now use an optional `error` variable, which is defined if
+the authentication process failed. See the default templates linked above for
+an example.
```
clokep, line 124:
```suggestion
The template is now provided an `error` variable if the authentication 
process failed. See the default templates linked above for an example.
```

docs/upgrade.md
```diff
@@ -112,6 +112,16 @@ environment variable.
 See [using a forward proxy with Synapse documentation](setup/forward_proxy.md) for
 details.
 
+## UIA fallback templates can display errors
+
+This may affect you if you make use of custom HTML templates for the
+[reCAPTCHA](../synapse/res/templates/recaptcha.html) and
+[terms](../synapse/res/templates/terms.html) fallback pages.
```
clokep, line 119:
```suggestion
This may affect you if you make use of custom HTML templates for the
[reCAPTCHA](../synapse/res/templates/recaptcha.html) or
[terms](../synapse/res/templates/terms.html) fallback pages.
```

docs/upgrade.md
```diff
@@ -112,6 +112,16 @@ environment variable.
...
+There is no configuration option for these templates.
```
clokep, line 120: I'm not sure what this means -- what would the configuration option do?

clokep:
Looks close! I left some comments on the upgrade notes, but the code is solid.

docs/upgrade.md
govynnus: Well for other templates (e.g. account validity) there are configuration options so you can give a path to custom templates. I'll just remove that line though because it's probably just confusing.

govynnus:
> Looks close! I left some comments on the upgrade notes, but the code is solid.

Thanks, I fixed the upgrade notes in 70e653e0827ee563420e57c9f821c2a397da950f

clokep (approved):

PR #10563, opened 2021-08-09
Update contributing.md to warn against rebasing an open PR.

H-Shay:
Update contributing.md to warn against rebasing an open PR.

### DIFF
--- a/changelog.d/10563.misc
+++ b/changelog.d/10563.misc
@@ -1,0 +1,1 @@
+Update contributing.md to warn against rebasing an open PR.
--- a/10562.misc
+++ b/10562.misc
@@ -1,0 +1,1 @@
+Update contributing.md to warn against rebasing an open PR.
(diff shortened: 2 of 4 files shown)


richvdh (approved):
Thanks @H-Shay!

(On closer inspection, this is also mentioned in https://github.com/matrix-org/synapse/blob/a054cd7e05667a353ed5b28c46d489ac955b33fd/CONTRIBUTING.md#10-turn-feedback-into-better-code, but I don't think there's any harm having it here too.)

changelog.d/10563.misc
```diff
@@ -0,0 +1 @@
+Update contributing.md to warn against rebasing an open PR.
```
clokep, line 1: Maybe rename this to a `.doc`?

richvdh: given it's developer rather than admin doc, I don't think it matters too much :man_shrugging:

Issue #7867, opened 2020-07-16
Allow `/createRoom` to be run against event creator


erikjohnston:
Currently it has to point at master, purely because `store_room` calls `self._public_room_id_gen.get_next()`. We should try and figure out a way that we can run `/createRoom` without interacting with master if possible.

anoadragon453:
Would this involve modifying/splitting out `self._public_room_id_gen` into a `MultiWriterIdGenerator` instance, similar to what was done for presence's ID generator [here](https://github.com/matrix-org/synapse/commit/9d25a0ae65ce8728d0fda1eebaf0b469316f84d7#diff-0de9c0df728ed72dd61e651b98052550076feb1b8947f95e5c75aeea869af034R45-R58)?

anoadragon453:
> Would this involve modifying/splitting out self._public_room_id_gen into a MultiWriterIdGenerator instance, similar to what was done for presence's ID generator here?

I believe so. One needs to:

1. Convert `self._public_room_id_gen` to a `MultiWriterIdGenerator` such that it can be incremented by more than one process.
2. Move the `store_room` storage method to `RoomWorkerStore`.

With that, we could also explore moving room upgrade tasks to workers, as they also rely on being able to create a room.

anoadragon453:
https://github.com/matrix-org/synapse/pull/9044 is another good example of a PR moving a `StreamIdGenerator` / `SlavedIdTracker` pair to a `MultiWriterIdGenerator`.

However in the process of investigating what would be involved, @erikjohnston noticed that this stream doesn't actually seem to be utilised anywhere anymore. It likely stemmed from the days of when we used to paginate the public rooms list with stream IDs, instead of an arbitrary pagination today - which we have today.

As such, the solution here is simply to just remove `public_room_list_stream`, rather than needing to convert it. That will then unblock moving `/createRoom` off of the main process.

As with all potentially risky database changes, the plan is to:

1. Remove reading and writing to `public_room_list_stream`, but keep the table around just in case.
2. Ship a release of Synapse and let that change sit for a bit. Move `/createRoom` off the main process at the same time.
3. Create a migration to remove the table and its contents.
4. Ship another release.

anoadragon453:
Fixed by https://github.com/matrix-org/synapse/pull/10564.

PR #10564, opened 2021-08-09
Allow /createRoom to be run on workers

anoadragon453:
Fixes https://github.com/matrix-org/synapse/issues/7867

As stated in the issue, `/createRoom` could only be run on the main process due to the `public_room_list_stream` only being writable by the main process. It turns out however that this stream was unused - meaning that we could simply rip it out. This unblocked moving `/createRoom` off the main process :tada: It potentially also unblocks moving the Delete Room Admin API - but that's the subject of another PR.

**This PR is based on https://github.com/matrix-org/synapse/pull/10565**, which removes `public_room_list_stream` (just reading/writing to the table - not the table itself, for now). This PR does the job of mounting the endpoint to workers and moving the database method `store_room` to the `RoomWorkerStore`.

### DIFF
--- a/changelog.d/10564.feature
+++ b/changelog.d/10564.feature
@@ -1,0 +1,1 @@
+Allow creating a room to be handled by worker processes.
--- a/changelog.d/10564.feature
+++ b/changelog.d/10564.feature
@@ -1,1 +1,1 @@
-Allow creating a room to be handled by worker processes.
+Add support for routing `/createRoom` to workers.
--- a/docs/workers.md
+++ b/docs/workers.md
@@ -214,6 +214,7 @@ expressions:
     ^/_matrix/federation/v1/send/
 
     # Client API requests
+    ^/_matrix/client/(api/v1|r0|unstable)/createRoom$
     ^/_matrix/client/(api/v1|r0|unstable)/publicRooms$
     ^/_matrix/client/(api/v1|r0|unstable)/rooms/.*/joined_members$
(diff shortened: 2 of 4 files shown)


synapse/replication/slave/storage/room.py
```diff
@@ -12,26 +12,15 @@
...
-        )
-
-    def get_current_public_room_stream_id(self):
-        return self._public_room_id_gen.get_current_token()
 
     def process_replication_rows(self, stream_name, instance_name, token, rows):
-        if stream_name == PublicRoomsStream.NAME:
-            self._public_room_id_gen.advance(instance_name, token)
-
         return super().process_replication_rows(stream_name, instance_name, token, rows)
```
anoadragon453, line 26: I'm unsure whether this method can be removed entirely now.

changelog.d/10564.feature
```diff
@@ -0,0 +1 @@
+Allow creating a room to be handled by worker processes.
```
erikjohnston, line 1:
```suggestion
Add support for routing `/createRoom` to workers.
```

erikjohnston (changes requested):
Can you update `docs/workers.md` please?

synapse/replication/slave/storage/room.py
anoadragon453: Removed in #10565.

erikjohnston (approved):
Once the other PR has landed

PR #10565, opened 2021-08-09
Remove the unused public_room_list_stream

anoadragon453:
Helps address https://github.com/matrix-org/synapse/issues/7867

Modifications to this stream were only possible on the main process, which had the knock-on effect of other tasks, mainly creating a room, only being possible on the main process as well.

Instead of rearchitecting this stream to allow multiple writers though, we can actually just remove it entirely - it's not used.

The actual `public_room_list_stream` database table will be dropped in a subsequent release of Synapse. Thus we PR bump the `SCHEMA_VERSION` to 63, as demonstrated in https://matrix-org.github.io/synapse/latest/development/database_schema.html#synapse-schema-versions.


synapse/replication/tcp/streams/_base.py
```diff
@@ -447,31 +447,6 @@ def __init__(self, hs):
...
-
-    NAME = "public_rooms"
-    ROW_TYPE = PublicRoomsStreamRow
-
-    def __init__(self, hs):
-        store = hs.get_datastore()
-        super().__init__(
-            hs.get_instance_name(),
-            current_token_without_instance(store.get_current_public_room_stream_id),
-            store.get_all_new_public_rooms,
```
erikjohnston, line 471: We should also remove `get_all_new_public_rooms`?

erikjohnston (changes requested):

synapse/replication/slave/storage/room.py
```diff
@@ -12,26 +12,15 @@
...
 from synapse.storage.database import DatabasePool
 from synapse.storage.databases.main.room import RoomWorkerStore
 
 from ._base import BaseSlavedStore
-from ._slaved_id_tracker import SlavedIdTracker
 
 
 class RoomStore(RoomWorkerStore, BaseSlavedStore):
     def __init__(self, database: DatabasePool, db_conn, hs):
         super().__init__(database, db_conn, hs)
```
clokep, line 23: The `__init__` can be fully removed if it just calls `super(...)`. 🚗 

Same seems to be true of `process_replication_rows`.

synapse/storage/databases/main/room.py
```diff
@@ -1423,21 +1423,8 @@ def store_room_txn(txn, next_id):
...
-                            "room_id": room_id,
-                            "visibility": is_public,
-                        },
-                    )
 
-            async with self._public_room_id_gen.get_next() as next_id:
-                await self.db_pool.runInteraction(
-                    "store_room_txn", store_room_txn, next_id
-                )
+            await self.db_pool.runInteraction("store_room_txn", store_room_txn)
```
clokep, line 1427: I think this can be replaced with calling `simple_insert` directly instead of the `txn` version?

synapse/replication/tcp/streams/_base.py
anoadragon453: Indeed - it's now completely unused!

synapse/storage/databases/main/room.py
anoadragon453: That was my thought as well - though is there any downside to losing the `store_room_txn` description information?

synapse/replication/slave/storage/room.py
```diff
@@ -12,26 +12,14 @@
...
 
-    def get_current_public_room_stream_id(self):
-        return self._public_room_id_gen.get_current_token()
-
-    def process_replication_rows(self, stream_name, instance_name, token, rows):
-        if stream_name == PublicRoomsStream.NAME:
-            self._public_room_id_gen.advance(instance_name, token)
-
-        return super().process_replication_rows(stream_name, instance_name, token, rows)
+    pass
```
clokep, line 25: Just an FYI that docstrings count as a statement so you don't need the `pass` here! 🌮

anoadragon453: Good shout!

synapse/storage/databases/main/room.py
clokep: `simple_insert` (and all the `simple_*` functions take a `desc` attribute, which I think is passed as the same parameter to `runInteraction`?

anoadragon453: I can't believe I missed that - thanks!

synapse/replication/slave/storage/room.py
```diff
@@ -12,26 +12,12 @@
...
 
-from synapse.replication.tcp.streams import PublicRoomsStream
-from synapse.storage.database import DatabasePool
 from synapse.storage.databases.main.room import RoomWorkerStore
 
 from ._base import BaseSlavedStore
-from ._slaved_id_tracker import SlavedIdTracker
 
 
 class RoomStore(RoomWorkerStore, BaseSlavedStore):
```
clokep, line 20: I wonder if we should just directly use `RoomWorkerStore` in `generic_worker`. We seem to do a mixture though so I wouldn't block on that.

synapse/storage/databases/main/room.py
```diff
@@ -1471,48 +1405,13 @@ async def maybe_store_room_on_outlier_membership(
         )
 
     async def set_room_is_public(self, room_id, is_public):
```
clokep, line 1407: While we're here?

```suggestion
    async def set_room_is_public(self, room_id: str, is_public: bool) -> None:
```

synapse/storage/databases/main/room.py
```diff
@@ -1533,68 +1432,31 @@ async def set_room_is_public_appservice(
...
                     },
+                    desc="set_room_is_public_appservice_true",
                 )
-
-            entries = self.db_pool.simple_select_list_txn(
-                txn,
-                table="public_room_list_stream",
+            except self.database_engine.module.IntegrityError:
+                # We've already inserted, nothing to do.
+                return
```
clokep, line 1448: I think this early return will now cause it to skip the notifier. I think you want to `pass` here.

synapse/storage/databases/main/room.py
```diff
@@ -1533,68 +1432,31 @@ async def set_room_is_public_appservice(
...
                         "network_id": network_id,
                         "room_id": room_id,
                     },
+                    desc="set_room_is_public_appservice_true",
                 )
-
-            entries = self.db_pool.simple_select_list_txn(
-                txn,
-                table="public_room_list_stream",
+            except self.database_engine.module.IntegrityError:
```
erikjohnston, line 1446: This works, but I'd rather we avoided it as integrity errors cause postgres to log errors to its logs, which is a bit annoying. Instead can you use `simple_upsert` with `values={}` please? That will do a `INSERT ... ON CONFLICT DO NOTHING`, which is what we probably want here.

synapse/replication/slave/storage/room.py
anoadragon453: We do import the `*WorkerStore` variant in `GenericWorkerSlavedStore` in some cases, so doing so here probably isn't too contentious.

Where is this mixture you mentioned?

synapse/storage/databases/main/room.py
anoadragon453: Good shout, though the refactor took care of this as well.

synapse/replication/slave/storage/room.py
anoadragon453: I see, OK. I've removed the class for now as it's probably best to not keep it around unless needed.

Tests pass, though it probably warrants a double-check to make sure I did the right thing :)

clokep (approved):
I think this looks OK, not sure if @erikjohnston wants to take a quick look too.

erikjohnston (approved):

PR #10570, opened 2021-08-10
Update the Synapse Grafana dashboard

babolivier:
With shiny new panels and stuff

### DIFF
--- a/changelog.d/10570.feature
+++ b/changelog.d/10570.feature
@@ -1,0 +1,1 @@
+Update the Synapse Grafana dashboard.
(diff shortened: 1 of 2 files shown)


babolivier:
I'm not sure whether this PR needs a review since it was lifted as is from our internal Grafana. So I'm also happy if someone decides to approve it without looking at the changes too much.

anoadragon453 (approved):
It'd be helpful to quickly note the names of the panels being added for historical sake - but otherwise looks good to me.

babolivier:
> It'd be helpful to quickly note the names of the panels being added for historical sake

I thought the same, but it looks like more got added since the last snapshot than just the one I added recently.

PR #10571, opened 2021-08-10
Set room version 8 as preferred for restricted rooms.

clokep:
And enable room capabilities (MSC3244) by default.

TBD when we want to merge this, but hopefully by Synapse 1.42.


clokep:
Unsure when we want to do this (targeting either 1.41 or 1.42), but it would be nice to get review on this so it can be merged once there's enough Synapse uptake.

babolivier (approved):

clokep:
> Unsure when we want to do this (targeting either 1.41 or 1.42), but it would be nice to get review on this so it is ready to be merged once there's enough Synapse uptake.

Only about ~25% of Synapse in the federation have updated to 1.40.0+ (which supports room version 8). We're not going to include this in 1.41.0 and will re-evaluate for 1.42.0.

clokep:
> > Unsure when we want to do this (targeting either 1.41 or 1.42), but it would be nice to get review on this so it is ready to be merged once there's enough Synapse uptake.
> 
> Only about ~25% of Synapse in the federation have updated to 1.40.0+ (which supports room version 8). We're not going to include this in 1.41.0 and will re-evaluate for 1.42.0.

We re-evaluated this data and although only about ~35% of Synapses have upgraded, this represents the vast majority of *users*. So we're going to include this in 1.41.0.

PR #10572, opened 2021-08-10
Clarify error message when joining a restricted room.

clokep:
This clarifies the error message when you attempt to join a restricted room (from MSC3083) that you do not have permission to join. It now mentions both "rooms" and "spaces".

Note that this is shown to users:

image


ara4n (approved):
lgtm, thanks

richvdh (approved):

PR #10574, opened 2021-08-10
Expire old spaces summary pagination sessions.

clokep:
Part of step 4 of #10495.

This expires spaces summary pagination sessions after a period of time. This arbitrarily uses 5 minutes -- if a client hasn't come back to keep paginating after that I think we can safely get rid of the session.

There might be more clever things to do here (reusing sessions and such), but I think we at least want to ensure this doesn't grow forever before releasing.

### DIFF
--- a/synapse/handlers/space_summary.py
+++ b/synapse/handlers/space_summary.py
@@ -78,7 +78,7 @@ class _PaginationSession:
     """The information that is stored for pagination."""
 
     # The time the pagination session was created, in milliseconds.
-    creatione_time_ms: int
+    creation_time_ms: int
     # The queue of rooms which are still to process.
     room_queue: Deque["_RoomQueueEntry"]
     # A set of rooms which have been processed.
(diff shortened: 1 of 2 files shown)


synapse/handlers/space_summary.py
```diff
@@ -77,13 +77,18 @@ class _PaginationKey:
 class _PaginationSession:
     """The information that is stored for pagination."""
 
+    # The time the pagination session was created, in milliseconds.
+    creatione_time_ms: int
```
richvdh, line 81: creatione! molto bene!

richvdh (commented):
Looks good except for the unexpected switch into Italian.

richvdh (approved):

synapse/handlers/space_summary.py
clokep: Must have been my Duolingo kicking in...

PR #10575, opened 2021-08-10
Fix types in spaces tests.

clokep:
...and actually check the file in `mypy.ini`.


anoadragon453 (approved):
Looks fine to me! Uses the same changelog as #10560.

clokep:
> Uses the same changelog as #10560.

Oops, yes. I didn't see a reason to split this change out to a separate changelog. 🤷 I found it by splitting out the work anyway. 😄

PR #10578, opened 2021-08-11
Add an admin API to check if a username is available

### LINKED ISSUE #10577: Ability to check if a username is available as an admin
As part of a project I'm working on, I'd like to see if a given localpart is available in synapse *without* having enable registration. The details are a bit complex but:
- We don't want to enable registration because we're using SSO (and don't want users to simply stroll up and register an account directly).
- We have appservices on the host, which means we cannot simply use the SSO provider as a means to check if a localpart is avaliable. Synapse is the best place to be checking if a username is available because it's aware of these.

Essentially, I'd like for admins to still be able to call https://matrix.org/docs/spec/client_server/r0.6.0#get-matrix-client-r0-register-available regardless of the config setting.

Half-Shot:
Fixes #10577 

This adds a new API `GET /_synapse/admin/v1/username_available?username=foo` to check if a username is available. It is the counterpart to https://matrix.org/docs/spec/client_server/r0.6.0#get-matrix-client-r0-register-available, except that it works even if registration is disabled.


Half-Shot:
Tests should be golden now.

There was a conversation in #synapse-dev:matrix.org about whether we should have this endpoint in Synapse, given the expectation is that SSO is the source of truth for users. There are a few reasons why I'd like this API to still be considered valid however:

I believe there are other use cases outside of an SSO use case where you'd still like this API to be available to admins without enabling the registration bit. Administrators may wish to disable public registration, but provide an alternative registration service which then uses the admin API to register users (such as https://github.com/ZerataX/matrix-registration/) where that service would still like to check for username availability.

On the subject of why we can't use the existing admin api to check for a user's existence, we need to be able to check for appservice namespace clashes or any additional synapse module which might block registering a localpart (such as anti-spam).

On the subject of why we can't tell the SSO provider about the appservice/module namespaces, we could...but it's difficult:

- It would first require maintaining two sources of truth for the same information, which isn't a problem in itself but it is more maintenance burden for the administrator
- If there are any *dynamic* modules involved which block usernames *sometimes*, we would need to somehow inform the SSO provider of those states as well.
- Currently without modification, our chosen provider KeyCloak doesn't allow you to namespace away attributes (which is what we use for localparts). While we could indeed write some Java code to solve this, or use a different provider, it becomes an order of magnitude more work to maintain than this API.
- There could be a valid use case for the SSO provider to use **this** API as a validation source when registering new users in it's system, rather than maintaining it's own config.

docs/admin_api/user_admin_api.md
```diff
@@ -1013,3 +1013,16 @@ The following parameters should be set in the URL:
 - `user_id` - The fully qualified MXID: for example, `@user:server.com`. The user must
   be local.
 
+### Check username availability
```
richvdh, line 1016: please can you define what "available" actually means here?

docs/admin_api/user_admin_api.md
```diff
@@ -1013,3 +1013,16 @@ The following parameters should be set in the URL:
...
+
+```
+POST /_synapse/admin/v1/username_availabile?username=$localpart
+```
+
+To use it, you will need to authenticate by providing an `access_token` for a
+server admin: [Admin API](../usage/administration/admin_api)
+
+
+The request and response format is the same as the [/_matrix/client/r0/register/available](https://matrix.org/docs/spec/client_server/r0.6.0#get-matrix-client-r0-register-available) API
```
richvdh, line 1028:
```suggestion
The request and response format is the same as the [/_matrix/client/r0/register/available](https://matrix.org/docs/spec/client_server/r0.6.0#get-matrix-client-r0-register-available)
API.
```

docs/admin_api/user_admin_api.md
richvdh: Or rather, I guess, give a short summary and link to the `/register/available` API for more details.

synapse/rest/admin/users.py
```diff
@@ -1097,3 +1097,28 @@ async def on_DELETE(
         await self.store.delete_ratelimit_for_user(user_id)
 
         return 200, {}
+
+
+class UsernameAvailableRestServlet(RestServlet):
```
richvdh, line 1102: we ought to break up this file into smaller modules. Also, there ought to be a "register servlets" function in here so that they don't all have to be imported individually into `synapse.rest.admin.__init__`.

Still, I guess this straw isn't going to break the camel's back.

tests/rest/admin/test_user.py
```diff
@@ -3406,3 +3411,54 @@ def test_success(self):
...
+
+        self.assertEqual(200, int(channel.result["code"]), msg=channel.result["body"])
+        self.assertTrue(channel.json_body["available"])
+
+    def test_username_unavailable(self):
+        """
+        The endpoint should return a 200 response if the username does not exist
+        """
+
+        async def check_username(username):
```
richvdh, line 3455: this seems unused?

richvdh (changes requested):

docs/admin_api/user_admin_api.md
Half-Shot: I've added a short description, there is a link already here to the existing docs further down.

synapse/rest/admin/users.py
Half-Shot: I've broken out both the tests and the rest handler, as both were easy to move :)

tests/rest/admin/test_user.py
Half-Shot: It was, and gone it is.

docs/admin_api/user_admin_api.md
```diff
@@ -1057,3 +1057,19 @@ The following parameters should be set in the URL:
 
 - `user_id` - The fully qualified MXID: for example, `@user:server.com`. The user must
   be local.
+
+### Check username availability
+
+Checks to see if a username is available, and valid, for the server. This will
+work even if registration is disabled on the server, unlike the Client-Server endpoint counterpart.
```
richvdh, line 1064: Again: which C-S endpoint counterpart?

```suggestion
Checks to see if a username is available, and valid, for the server. See [the client-server 
API](https://matrix.org/docs/spec/client_server/r0.6.0#get-matrix-client-r0-register-available)
for more information.

This endpoint will work even if registration is disabled on the server, unlike 
`/_matrix/client/r0/register/available`.
```

docs/admin_api/user_admin_api.md
richvdh: the link further down just talks about the request/response format shape. It doesn't say anything about defining what "available" actually means. I'm just talking about a "see ... for more information" link. See the suggestion below.

tests/rest/admin/test_username_available.py
```diff
@@ -0,0 +1,66 @@
...
+            if username == "allowed":
+                return True
+            raise SynapseError(400, "User ID already taken.", errcode=Codes.USER_IN_USE)
+
+        handler = self.hs.get_registration_handler()
+        handler.check_username = check_username
+
+    def make_homeserver(self, reactor, clock):
+        self.hs = self.setup_test_homeserver()
+        return self.hs
```
richvdh, line 43: I think this the default behaviour?

tests/rest/admin/test_username_available.py
```diff
@@ -0,0 +1,66 @@
...
+class UsernameAvailableTestCase(unittest.HomeserverTestCase):
+    servlets = [
+        synapse.rest.admin.register_servlets,
+        login.register_servlets,
+    ]
+
+    def prepare(self, reactor, clock, hs):
+        self.register_user("admin", "pass", admin=True)
+        self.admin_user_tok = self.login("admin", "pass")
+        self.url = "/_synapse/admin/v1/username_available"
```
richvdh, line 31: this should probably be a module-level constant?

richvdh (changes requested):

richvdh (approved):
lgtm!

PR #10580, opened 2021-08-11
Allow public spaces to be returned via the summary APIs.

clokep:
I'm pretty sure that we should be allowing viewing of a `public` room when requested over the spaces summary. The MSC states:

> An HTTP response with a status code of 403 and an error code of `M_FORBIDDEN` should be returned if the user doesn't have permission to view/peek the root room. This should also be returned if that room does not exist, which matches the behavior of other room endpoints

This description essentially matches what MSC3173 says (stripped state is viewable by any potential joiner). The current checks ensure the user is in the room or it is world-readable, which does not take into account cases such as:

* A public room
* An invite-only room where the user is invited

CC @t3chguy

### DIFF
--- a/changelog.d/10580.bugfix
+++ b/changelog.d/10580.bugfix
@@ -1,0 +1,1 @@
+Allow public rooms to be previewed in the spaces summary APIs from [MSC2946](https://github.com/matrix-org/matrix-doc/pull/2946).
--- a/synapse/handlers/space_summary.py
+++ b/synapse/handlers/space_summary.py
@@ -86,7 +86,6 @@ class _PaginationSession:
 class SpaceSummaryHandler:
     def __init__(self, hs: "HomeServer"):
         self._clock = hs.get_clock()
-        self._auth = hs.get_auth()
         self._event_auth_handler = hs.get_event_auth_handler()
         self._store = hs.get_datastore()
         self._event_serializer = hs.get_event_client_serializer()
--- a/synapse/handlers/space_summary.py
+++ b/synapse/handlers/space_summary.py
@@ -38,7 +38,7 @@ from synapse.api.constants import (
     Membership,
     RoomTypes,
 )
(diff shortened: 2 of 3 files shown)


richvdh (approved):
The change looks plausible, but please could you write down, for the record, exactly what you mean by a `public` room? It's something that is prone to interpretation, so it would be good to record exactly what states are affected here.

clokep:
> The change looks plausible, but please could you write down, for the record, exactly what you mean by a `public` room? It's something that is prone to interpretation, so it would be good to record exactly what states are affected here.

Definitely! Would you like it in comments in the code? (I think `_is_local_room_accessible` is documented decently.)

For the record, we should be allowing access in the spaces summary to rooms:

* Which a user has joined or is invited to or is joinable (`public` or `knock` join rules or joinable via restricted rooms).
* Which has world-readable history visibility.

This already matches the permissions we checked for sub-rooms/sub-spaces, but this PR updates the check for the *root* room to be consistent.

All of this is considered OK based on the logic given in [MSC3173](https://github.com/matrix-org/matrix-doc/pull/3173), with the caveat that it essentially means we're considering `m.space.child` state events as part of the stripped state (something along those lines should probably be added to [MSC2946](https://github.com/matrix-org/matrix-doc/pull/2946)).

richvdh:
yeah, I think the code is probably clear enough. Just wanted the PR to record exactly what was changing.

For clarity: before this PR we only allowed the user to start the summary at a room they were already in (or had world-readable history vis), and not those they could *potentially* join?

clokep:
> For clarity: before this PR we only allowed the user to start the summary at a room they were already in (or had world-readable history vis), and not those they could _potentially_ join?

Correct.

Once the above was met, a user was allowed to see any rooms in that summary which they user was in, had world readable history vis, or could _potentially_ join. Which is rather confusing from a user's POV. This now consistently applies the same checks in both cases.

PR #10581, opened 2021-08-11
Remove pushers when deleting 3pid from account

### LINKED ISSUE #10568: Removing an email does not remove the notifications
Hello

When you add an email, then activate notifications by email at that address, and then remove the email from your profile, you still get notifications.

I believe removing the email (threepid) from your profile should remove the notification.
Alternatively, the email notification block should allow you to remove it. 

The work around is to add the email back, disable the notification, then remove the email again.

However, if you lost access to an email account, you can find yourself in a position when you have absolutely not way of removing the notifications by email. This is really a problem.

Fixing that issue would probably fix bug #4958 too.

Azrenbeth:
When a user deletes an email from their account it will now also remove all pushers for that email and that user (even if these pushers were created by a different client)

Fixes #10568


synapse/storage/databases/main/pusher.py
```diff
@@ -478,6 +478,53 @@ def delete_pusher_txn(txn, stream_id):
                 "delete_pusher", delete_pusher_txn, stream_id
             )
 
+    async def delete_all_pushers_with_pushkey_and_user_id(
```
anoadragon453, line 481: I realise in the linked issue that we should add another method for deleting pushers, but I've just thought that since there's a unique constraint on the `pushers` table for `(app_id, push_key, user_name)`, we may be able to get away with just using `delete_pusher_by_app_id_pushkey_user_id`, as app_id should always be `m.email` in this context.

anoadragon453: We should also probably make sure we remove only entries with `app_id` = `m.email` anyhow I suppose. Note that I don't believe msisdns (phone numbers) are specced here at all yet: https://spec.matrix.org/unstable/client-server-api/#post_matrixclientr0pushersset

anoadragon453 (changes requested):

clokep:
Do we have any sort of clean-up to do with old data or do we not know enough information in order to do that?

Azrenbeth:
> Do we have any sort of clean-up to do with old data or do we not know enough information in order to do that?

What do you mean by that?

anoadragon453:
@clokep Good point! I think that would be ideal, yes.

This would effective involve writing a background job that removes any `m.email` entries in `pushers` with emails that are not currently linked to a user ID.

Note that people *may* be relying on those notifications though - so we'd need to shout that if you stop receiving notifications unexpectedly - you should make sure your email is registered to your user account.

Azrenbeth:
> Up to @Azrenbeth whether that would be in this PR, a separate one or whether to delegate it to someone else though :)

I think maybe that could be a seperate PR? Just because it might take me a while to work out how to do that!

synapse/storage/databases/main/pusher.py
```diff
@@ -48,6 +48,11 @@ def __init__(self, database: DatabasePool, db_conn: Connection, hs: "HomeServer"
             self._remove_stale_pushers,
         )
 
+        self.db_pool.updates.register_background_update_handler(
+            "remove_delete_email_pushers",
```
nirgal, line 52: wouldn't remove_delete**d**_email_pushers be better?

synapse/storage/databases/main/pusher.py
```diff
@@ -388,6 +393,57 @@ def _delete_pushers(txn) -> int:
 
         return number_deleted
 
+    async def _remove_deleted_email_pushers(
+        self, progress: dict, batch_size: int
+    ) -> int:
+        """A background update that deletes all pushers for email addresses no longer
+        associated with a user.
```
anoadragon453, line 400: Could you note here in [Google python docstring notation](https://google.github.io/styleguide/pyguide.html#383-functions-and-methods) a description for each argument and the return value? Something analogous to: https://github.com/matrix-org/synapse/blob/684d19a11c3b93c9dd5fb90f43d38aa7e8c6005f/synapse/storage/database.py#L1795-L1805

changelog.d/10581.bugfix
```diff
@@ -0,0 +1 @@
+Remove pushers when deleting 3pid from account.
```
anoadragon453, line 1: Best to update this to mention that old, unlinked emails will also be deleted.

anoadragon453, line 1: Additionally, could you add an entry to [CHANGES.md](https://github.com/matrix-org/synapse/blob/develop/CHANGES.md) and [docs/upgrade.md](https://github.com/matrix-org/synapse/blob/develop/docs/upgrade.md) just informing homeserver operators that users may stop receiving message updates via email for addresses that were once, but not longer linked to their account?

Just sticking it at the top of CHANGES.md should be fine, we'll just need to remember to hoist it up to the right section come release day. For docs/upgrade.md, simply make a new section for the next release, likely 1.42.0, and put the notice in there under an appropriate heading.

synapse/storage/databases/main/pusher.py
```diff
@@ -388,6 +393,57 @@ def _delete_pushers(txn) -> int:
...
+            sql = """
+                SELECT p.id, p.user_name, p.add_id, p.pushkey
+                FROM pushers AS p
+                    JOIN user_threepids AS t ON t.user_id=p.user_name
+                WHERE p.app_id = 'm.email'
+                    AND t.medium = 'email'
+                    AND t.address = p.pushkey
+                    AND p.id > ?
+                ORDER BY p.id ASC
+                LIMIT ?
```
anoadragon453, line 416: I may be missing something, but won't this only delete pushers that *are* associated with a user account; whereas we want to be doing the opposite?

anoadragon453 (changes requested):

synapse/storage/databases/main/pusher.py
Azrenbeth: Eek I forgot to test that! At least I spelled `p.app_id` wrong too!

This query should actually do what we want:
```sql
SELECT p.id, p.user_name, p.app_id, p.pushkey
FROM pushers AS p
    LEFT JOIN user_threepids AS t 
        ON t.user_id=p.user_name
        AND t.medium = 'email'
        AND t.address = p.pushkey
WHERE t.user_id is NULL
    AND p.app_id = 'm.email'
    AND p.id > ?
ORDER BY p.id ASC
LIMIT ?
```

CHANGES.md
```diff
@@ -1,3 +1,5 @@
+Users will stop receiving message updates via email for addresses that were once, but not still, linked to their account
```
anoadragon453, line 1:
```suggestion
Users will stop receiving message updates via email for addresses that were previously linked to their account.
```

Might be slightly simpler wording.

changelog.d/10581.bugfix
```diff
@@ -1 +1 @@
-Remove pushers when deleting 3pid from account.
\ No newline at end of file
+Remove pushers when deleting 3pid from account. Pushers for old unlinked emails will also be deleted.
```
anoadragon453, line 1:
```suggestion
Remove pushers when deleting a 3pid from an account. Pushers for old unlinked emails will also be deleted.
```

synapse/storage/databases/main/pusher.py
```diff
@@ -396,42 +396,57 @@ def _delete_pushers(txn) -> int:
     async def _remove_deleted_email_pushers(
         self, progress: dict, batch_size: int
     ) -> int:
-        """A background update that deletes all pushers for email addresses no longer
-        associated with a user.
+        """A background update that deletes all pushers for deleted email addresses.
+
+        In previous versions of synapse, when users deleted their email address, it didn't
+        also delete all the pushers for that email address. This background update removes
+        those to prevent unwanted emails
```
anoadragon453, line 403: You may wish to mention that this should only need to be run once when users upgrade to v1.42.0. Otherwise it may be easy to assume this background job is run every so often to clean up old pushers.

synapse/storage/databases/main/pusher.py
```diff
@@ -396,42 +396,57 @@ def _delete_pushers(txn) -> int:
...
             sql = """
-                SELECT p.id, p.user_name, p.add_id, p.pushkey
+                SELECT p.id, p.user_name, p.app_id, p.pushkey
                 FROM pushers AS p
-                    JOIN user_threepids AS t ON t.user_id=p.user_name
-                WHERE p.app_id = 'm.email'
-                    AND t.medium = 'email'
-                    AND t.address = p.pushkey
+                    LEFT JOIN user_threepids AS t
+                        ON t.user_id=p.user_name
```
anoadragon453, line 421:
```suggestion
                        ON t.user_id = p.user_name
```

just for consistency with the below.

synapse/storage/databases/main/pusher.py
```diff
@@ -396,42 +396,57 @@ def _delete_pushers(txn) -> int:
...
+        Returns:
+            The number of deleted rows
         """
 
         last_pusher = progress.get("last_pusher", 0)
 
         def _delete_pushers(txn) -> int:
 
             sql = """
-                SELECT p.id, p.user_name, p.add_id, p.pushkey
```
anoadragon453: Oop, good catch!

anoadragon453 (commented):
This is looking really good! Last request (sorry) is to add a test for the pusher getting deleted. Essentially just:

* Link an email and create a pusher for it.
* Deactivate the account.
* Check whether a pusher still exists.

tests/push/test_email.py
```diff
@@ -305,6 +307,36 @@ def test_encrypted_message(self):
         # We should get emailed about that message
         self._check_for_mail()
 
+    def test_no_email_sent_after_removed(self):
+        # Create a simple room with two users
+        room = self.helper.create_room_as(self.user_id, tok=self.access_token)
+        self.helper.invite(
+            room=room, src=self.user_id, tok=self.access_token, targ=self.others[0].id
+        )
+        self.helper.join(room=room, user=self.others[0].id, tok=self.others[0].token)
```
anoadragon453, line 316: Note that when specifying argument names for all arguments, it can often be more readable to write each on a separate line. I found this rather tricky to scan through.

Unless black is forcing it to be on one line, then :shrug:

anoadragon453 (approved):
This LGTM! Thanks for adding a test :tada:

PR #10583, opened 2021-08-11
Query legacy spaces federation API if the hierarchy API is unavailable.

clokep:
Adds a backwards compatibility path to the room hierarchy federation endpoint by using the `/spaces` endpoint and translating the results (and throwing some of it away).

The reasoning for wanting this is that the new `/hierarchy` federation endpoint will not be available on all servers initially, but we want the C-S endpoint to work as expected. This should allow clients to adopt the new endpoint quickly.

~~Builds on #10569.~~

Part of #10495.

I did some manual testing of this, but I'm not really sure how to make automated tests for this (as I had to manually run two different versions of Synapse, one which knew of the `/hierarchy` endpoint and one which did not).

### DIFF
--- a/synapse/federation/federation_client.py
+++ b/synapse/federation/federation_client.py
@@ -1364,13 +1364,59 @@ class FederationClient(FederationBase):
 
             return room, children, inaccessible_children
 
-        # TODO Fallback to the old federation API and translate the results.
-        return await self._try_destination_list(
-            "fetch room hierarchy",
-            destinations,
-            send_request,
-            failover_on_unknown_endpoint=True,
-        )
+        try:
+            return await self._try_destination_list(
+                "fetch room hierarchy",
+                destinations,
+                send_request,
+                failover_on_unknown_endpoint=True,
+            )
+        except SynapseError as e:
+            # Fallback to the old federation API and translate the results if
+            # no servers implement the new API.
+            #
+            # The algorithm below is a bit inefficient as it only attempts to
+            # get information for the requested room, but the legacy API may
+            # return additional layers.
+            if e.code == 502:
+                legacy_result = await self.get_space_summary(
+                    destinations,
+                    room_id,
+                    suggested_only,
+                    max_rooms_per_space=None,
+                    exclude_rooms=[],
+                )
+
+                # Find the requested room in the response (and remove it).
+                for _i, room in enumerate(legacy_result.rooms):
+                    if room.get("room_id") == room_id:
+                        break
+                else:
+                    # The requested room was not returned, nothing we can do.
+                    raise
+                requested_room = legacy_result.rooms.pop(_i)
+
+                # Find any children events of the requested room.
+                children_events = []
+                children_room_ids = set()
+                for event in legacy_result.events:
+                    if event.room_id == room_id:
+                        children_events.append(event.data)
+                        children_room_ids.add(event.state_key)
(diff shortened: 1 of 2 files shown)


synapse/federation/federation_client.py
```diff
@@ -1364,13 +1364,59 @@ async def send_request(
...
+
+                # Find the children rooms.
+                children = []
+                for room in legacy_result.rooms:
+                    if room.get("room_id") in children_room_ids:
+                        children.append(room)
+
+                # It isn't clear from the response whether some of the rooms are
+                # not accessible.
+                return requested_room, children, ()
```
anoadragon453, line 1417: Will this produce odd results on the client-side where one may believe they can access a room, but will find they are unable to?

Mostly a question out of curiosity, as I don't know any way that the situation could be improved (and this is a legacy API so meh).

synapse/federation/federation_client.py
```diff
@@ -1364,13 +1364,59 @@ async def send_request(
...
+                failover_on_unknown_endpoint=True,
+            )
+        except SynapseError as e:
+            # Fallback to the old federation API and translate the results if
+            # no servers implement the new API.
+            #
+            # The algorithm below is a bit inefficient as it only attempts to
+            # get information for the requested room, but the legacy API may
+            # return additional layers.
+            if e.code == 502:
```
anoadragon453, line 1381: Does Synapse not return a `400, M_UNRECOGNIZED` for endpoints it fails to understand, rather than a 502?

anoadragon453 (commented):
Also noticed a typo in this method's docstring: "Performs simple data **validates** ...".

synapse/federation/federation_client.py
clokep: It shouldn't, this is before any filtering is done on whether a room is accessible or not, it pretty much is just about changing API shapes from the old `/spaces` endpoint to the new `/hierarchy`. Part of that is that the `/spaces` endpoint returned multiple layers deep (over federation) while the `/hierarchy` one does *not*. So we throw away a bunch of data.

The overall algorithm is:

1. Find the requested room (which is returned in `rooms` for `/spaces`, but as a separate field for `/hierarchy`)
2. Attach the children events to it (which is returned in a list including children for other rooms in `/spaces` and as part of the `room` object for `/hierarchy`).
3. Get the summaries for any of those children room (which are returned as `rooms` for `/spaces`, but those will include additional layers).

synapse/federation/federation_client.py
clokep: This is from https://github.com/matrix-org/synapse/blob/4b1722a32e9d6c60cee83c6d9ed769689e493ce8/synapse/federation/federation_client.py#L600

anoadragon453: aha, thanks!

clokep: The goal is only to run this if we were unable to fetch data from *all* destinations, not just if we were unable to fetch it from the first destination. (This is different than our fallback of v2 -> v1 for e.g. `send_join`, which we do on a per-destination basis).

synapse/federation/federation_client.py
anoadragon453: Aha, I see. Thanks for explaining it so clearly.

anoadragon453 (approved):
My questions have been answered, and I have no more.

PR #10587, opened 2021-08-12
Allow using several custom template directories

babolivier:
Allow using several directories in `read_templates`. Useful for #10596, and also allowing modules to specify their own module directory _on top_ of the configured default one(s).


synapse/config/_base.py
```diff
@@ -262,18 +262,23 @@ def read_templates(
...
+                    if not self.path_exists(custom_template_directory):
+                        raise ConfigError(
+                            "Configured template directory does not exist: %s"
+                            % (custom_template_directory,)
+                        )
 
-            # Search the custom template directory as well
-            search_directories.insert(0, custom_template_directory)
+                    # Search the custom template directory as well
+                    search_directories.insert(0, custom_template_directory)
```
richvdh, line 281: this means we'll search the directories in reverse order to that given in the arguments, I think? Is that deliberate? It should probably be documented if so? If not, it suggests we need more tests?

babolivier: It was not deliberate, thanks for pointing it out! I've added a test that checks the order we search directories in.

tests/config/test_base.py
```diff
@@ -74,8 +74,66 @@ def test_loading_custom_templates(self):
...
+        """Tests that directories are searched in the right order if multiple custom
+        template directories are provided.
+        """
+        # Create two temporary directories on the filesystem.
+        tempdirs = [
+            tempfile.TemporaryDirectory(),
+            tempfile.TemporaryDirectory(),
+        ]
+
+        # Create one template in each directory, which content is the index of the
```
richvdh, line 87:
```suggestion
        # Create one template in each directory, whose content is the index of the
```

richvdh (approved):
lgtm!

PR #10590, opened 2021-08-12
Split `synapse.federation.transport.server` into multiple files.

clokep:
Done as a few commits, but pretty much copies & pastes code between files.


richvdh (commented):
looks generally good to me (though I'm taking it on trust that you're just moving things around :)

However: should we move the `GROUP_SERVER_SERVLET_CLASSES` to `groups_server` and import that, rather than importing each individual servlet class? Similarly for the other files?

(Alternatively: should each module expose its own `register_servlets` method?)

clokep:
> However: should we move the `GROUP_SERVER_SERVLET_CLASSES` to `groups_server` and import that, rather than importing each individual servlet class? Similarly for the other files?
> 
> (Alternatively: should each module expose its own `register_servlets` method?)

I had wondered about moving the serlet constants in there. I'm not sure if moving the `register_servlets` in makes sense too, but I'll play with it a bit.

clokep:
> > However: should we move the `GROUP_SERVER_SERVLET_CLASSES` to `groups_server` and import that, rather than importing each individual servlet class? Similarly for the other files?
> > (Alternatively: should each module expose its own `register_servlets` method?)
> 
> I had wondered about moving the servlet constants in there. I'm not sure if moving the `register_servlets` in makes sense too, but I'll play with it a bit.

What I ended up doing was:

* Moving the long lists of servlets into the sub-modules. (Note that the `OpenIdUserInfo` was incorrectly moved in into one of the sub-modules.)
* Reworking the registration code to be less repetitive. This is done as two commits.

synapse/federation/transport/server/__init__.py
```diff
@@ -0,0 +1,333 @@
...
+    if not servlet_groups:
+        servlet_groups = DEFAULT_SERVLET_GROUPS.keys()
+
+    for servlet_group in servlet_groups:
+        # Skip unknown servlet groups.
+        if servlet_group not in DEFAULT_SERVLET_GROUPS:
+            logger.warning(
+                f"Attempting to register unknown federation servlet: '{servlet_group}'"
+            )
+            continue
```
clokep, line 325: As far as I can tell, `register_servlets` is only ever called `servlet_groups` of `None` or `["openid"]` so this check might not really be needed (or maybe it should raise an exception?)

richvdh: +1 to an exception

richvdh (approved):
lgtm!

PR #10592, opened 2021-08-12
Handle the staged events queue being empty.

clokep:
(I'm not 100% sure the title is correct.)

This attempts to fix https://sentry.matrix.org/sentry/synapse-matrixorg/issues/225383/ by properly handling the return value from `get_next_staged_event_id_for_room`, but I'm not 100% sure it is doing the right thing. :)

### DIFF
--- a/synapse/federation/federation_server.py
+++ b/synapse/federation/federation_server.py
@@ -977,7 +977,7 @@ class FederationServer(FederationBase):
                 latest_origin = None
                 latest_event = None
             else:
-                next_origin, next_event_id  = result
+                next_origin, next_event_id = result
                 if (
                     next_origin != latest_origin
                     or next_event_id != latest_event.event_id
--- a/synapse/federation/federation_server.py
+++ b/synapse/federation/federation_server.py
@@ -972,13 +972,21 @@ class FederationServer(FederationBase):
         # the room, so instead of pulling the event out of the DB and parsing
         # the event we just pull out the next event ID and check if that matches.
         if latest_event is not None and latest_origin is not None:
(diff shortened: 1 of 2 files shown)


synapse/federation/federation_server.py
```diff
@@ -972,13 +972,21 @@ async def _process_incoming_pdus_in_room_inner(
...
-            if next_origin != latest_origin or next_event_id != latest_event.event_id:
+            result = await self.store.get_next_staged_event_id_for_room(room_id)
+            if result is None:
                 latest_origin = None
                 latest_event = None
+            else:
+                (
+                    next_origin,
+                    next_event_id,
+                ) = result
```
erikjohnston, line 983:
```suggestion
                next_origin, next_event_id  = result
```

erikjohnston (approved):
Yeah that looks about right! I wonder if we should just return immediately if we get a `None` back, rather than doing `get_next_staged_event_for_room` (which does the same)?

synapse/federation/federation_server.py
clokep: I'm surprised lint was happy with this. I think I assumed black would just reformat it. Oops!

clokep:
> Yeah that looks about right! I wonder if we should just return immediately if we get a `None` back, rather than doing `get_next_staged_event_for_room` (which does the same)?

I don't know, I just kind of followed the code that was there (I found this function quite hard to follow TBH).

I should probably have mentioned that `mypy` would have caught this if we were checking the corresponding storage file with mypy, but that has many errors when I added it.

Issue #8445, opened 2020-10-01
Improve API/event validation in synapse


richvdh:
# Background

We've recently encountered a number of bugs in which malformed (or at least, unexpected) data has caused Synapse or clients to misbehave in some way. These bugs stem from the fact that, faced with a given datastructure, you cannot rely on it having the expected format. For example: 
 - What happens if the `displayname` in an `m.room.member` event is not a string (#8220, #8340)?
 - What happens if an event object is missing its `origin` field (#8319)?

Such bugs are disruptive, and in extreme cases could progress beyond "denial of service" to "security threat", and it might be possible to avoid them by validating data at the point of entry to Synapse. We've recently discussed this in some depth within the core Synapse team; this issue serves to record some of our thoughts on the topic, including promising areas for further development.

# Introduction

There are actually multiple reasons why it would be useful to improve validation of data within Synapse. These include:

 * As above, avoiding potential bugs
 * Simplifying code by removing `isinstance` checks everywhere
 * Forcing clients to follow the specification rather than using "whatever works in Synapse", thus improving compatibility with other server implementations which *do* enforce the spec (or conversely: not forcing other implementations to grandfather in Synapse's misfeatures for compatibility with clients)
 * Making synapse return 400 errors rather than "Internal Server Errors" if clients pass invalid parameters would:
    * Give better feedback to client developers: the 500 error gives the developer little information about their error.)
    * Allow server administrators to distinguish between "my server is having problems" and "somebody passed invalid parameters to my server"
    * Allow load-balancers to distinguish between "this process is having problems" and "somebody passed invalid parameters". Notably, Cloudflare will stop sending requests to an origin which returns too many 500 errors.

There are multiple related, but different, things we mean when we talk about validation. At a high level, these can be broken down into:

 * validating API request parameters (both CS and SS API); and
 * validating events (and other such data types, like EDUs etc).

Let's consider these in turn.

# API validation

Given that we already have JSON schema specifications for our APIs, this is theoretically relatively straightforward (see, for example https://github.com/matrix-org/synapse/commit/c39941cd6e3727d5571d01bc6f72d7d439f41d13, which is a proof-of-concept applying this to a single endpoint), though it's certain to uncover a large number of places where clients are relying on non-spec-compliant behaviour.  We also have to be wary, especially on the SS API, to ensure backwards compatibility. 

[...]

PR #10593, opened 2021-08-12
Validate device_keys for C-S /keys/query requests

### LINKED ISSUE #10354: Synapse accepts objects in place of device ids for `/_matrix/client/r0/keys/query`
https://matrix.org/docs/spec/client_server/r0.6.1#post-matrix-client-r0-keys-query says that `device_keys` must be a map from string to a list of strings (the device IDs). 

However, Synapse erroneously accepts other types in place of that list of strings (in particular, it accepts other maps).

In short: this is accepted as a valid request, where it should be rejected with a 400:

```
POST /_matrix/client/r0/keys/query HTTP/1.1

{
    "device_keys": {
        "@bob:hs1": {
            "device_id1": true,
        }
    }
}
```

DMRobertson:
Closes #10354

A small, not particularly critical fix.

I'm interested in seeing if we can find a more systematic approach though. #8445 is the place for any discussion.

Wishlist:
1. Declaratively specify the data we expect
2. Automatic validation
3. Gradually roll this out across synapse
4. don't eat too many CPU cycles
5. type hints available to Python to mypy/PyCharm etc can make use of the data

A quick search mentions jsonschema, fastjsonschema, pydantic. attrs has
validation but I think it's quite verbose to specify it?


synapse/rest/client/v2_alpha/keys.py
```diff
@@ -163,6 +163,18 @@ async def on_POST(self, request):
         device_id = requester.device_id
         timeout = parse_integer(request, "timeout", 10 * 1000)
         body = parse_json_object_from_request(request)
+
+        try:
+            device_keys = body["device_keys"]
+        except KeyError:
+            raise InvalidAPICallError("device_keys is a required body parameter")
+
+        for mxid, device_ids in device_keys.items():
```
clokep, line 172: Should we first confirm that `device_keys` is a `dict`?

It would also be nice to add a check that items in the list are of the proper form.

We do something similar to this:

https://github.com/matrix-org/synapse/blob/ef4329f8daa25079beba0894a3e46015eab2cbd7/synapse/federation/federation_client.py#L1420-L1424

DMRobertson: I was erring on the side of not doing too much work here, in lieu of looking for a more systematic approach. Happy to do this though

DMRobertson: Also, in the quoted code sample we will happily accept the case when `via` is a string:

```python3
Python 3.9.6 (default, Jul 16 2021, 00:00:00) 
[GCC 11.1.1 20210531 (Red Hat 11.1.1-3)] on linux
>>> x = "hello, world"
>>> from collections import Sequence
<stdin>:1: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.10 it will stop working
>>> isinstance(x, Sequence)
True
>>> all(isinstance(c, str) for c in x)
True
>>> 
```

clokep:
@DMRobertson Looks like this has a conflict!

tests/rest/client/v2_alpha/test_keys.py
```diff
@@ -0,0 +1,77 @@
+from http import HTTPStatus
```
clokep, line 1: This needs a license header.

synapse/api/errors.py
```diff
@@ -147,6 +147,14 @@ def error_dict(self):
         return cs_error(self.msg, self.errcode)
 
 
+class InvalidAPICallError(SynapseError):
```
clokep, line 150: I feel like if we're going to introduce a new subclass here we should consider using it more widely. (Or at least file an issue about it.)

changelog.d/10593.bugfix
```diff
@@ -0,0 +1 @@
+Reject Client-Server /keys/query requests which provide device_ids incorrectly.
```
clokep, line 1:
```suggestion
Reject Client-Server `/keys/query` requests which provide `device_ids` incorrectly.
```

synapse/api/errors.py
DMRobertson: That's fair. I had in mind #8445 here---e.g. that any systematic approach to validation could make use of this subclass.

clokep: Yeah that makes sense as doing it as part of that issue!

DMRobertson:
Apologies again. Those changes are made in 7862d704f.

DMRobertson:
Sanity check: ran my changes locally with 6e6398609, which has the changes in ee3b2ac because

```
$ git merge-base --is-ancestor ee3b2ac 6e6398609 && echo yes || echo no
yes
```

Could register a new client `C` on the command line and login as them in element-web 1.7.34. From another instance of the same element-web, I could invite user `C` to a group. I could accept the invite as `C`. I saw a successful request to `/keys/query` in the homserver log:

```
  2021-08-23 17:54:39,952 - synapse.access.http.8888 - 389 - INFO - POST-341 - 127.0.0.1 - 8888 - {@a:localhost} Processed request: 0.013sec/-* 0.000sec (0.005sec, 0.000sec) (0.004sec/0.005sec/3) 1412B 200 "POST /_matrix/client/r0/keys/query HTTP/1.1" "Mozilla/5.0 (X11; Fedora; Linux*  x86_64; rv:91.0) Gecko/20100101 Firefox/91.0" [0 dbevts]
```

This suggests I haven't horribly broken element-web, but I'd feel better having tested with a wider range of clients.

mcg-matrix:
> This suggests I haven't horribly broken element-web, but I'd feel better having tested with a wider range of clients.

As @aaronraimist helpfully explained in https://github.com/vector-im/element-web/issues/18992#issuecomment-918469788, the change probably makes Synapse unusable for the final release of Riot-Android (0.9.13).

If anyone can suggest a decent Matrix client for old Android environments that will continue to work, my users and me will be enormously thankful for a hint.

aaronraimist:
@mcg-matrix from what I can tell Element, FluffyChat, and SchildiChat all target Android 5.0 or higher. [Syphon](https://play.google.com/store/apps/details?id=org.tether.tether) is the only actively maintained Matrix client that says it supports Android versions below 5.0. Looks like it supports Android 4.3, not sure what your requirements are. Google Play doesn't seem to say what Android versions Ditto supports.

You mentioned Sailfish OS in the other thread. There is also [Sailtrix](https://gitlab.com/HengYeDev/harbour-sailtrix) which is a Matrix client specific for Sailfish OS.

If 4.3 isn't low enough then depending on how important it is to you you could
1. Run your own matrix server and keep it up to date but remove this specific patch
2. Fork Riot Android and fix the bug with invalid requests to this API (probably not too difficult)
3. Fork Element Android and remove the things that need new APIs (probably pretty difficult)
4. Or buy your users new devices :)

Issue #7280, opened 2020-04-15
Improve customizability and consistency of templates


clokep:
Some new templates were added recently that dealt with SSO (#7102, #7240, #7279, #7067), but I think this is a wider issue.

There's a few oddities with the templates used in Synapse:
* [Some are hard-coded HTML embedded in Python](https://github.com/matrix-org/synapse/blob/6fb63d6426b87e989143b6ff2679f5ae4a8fd22b/synapse/rest/client/v2_alpha/auth.py#L28-L114) (this is non-ideal due to no syntax highlighting in editors, etc as well as not being customizable).
* Some "templates" are loaded [as raw HTML files](https://github.com/matrix-org/synapse/blob/229eb81498b0fe1da81e9b5b333a0285acde9446/synapse/config/emailconfig.py#L227-L245), but some are [treated as Jinja2](https://github.com/matrix-org/synapse/blob/229eb81498b0fe1da81e9b5b333a0285acde9446/synapse/handlers/auth.py#L116-L119). (Note also that some of those are loaded during the configuration step, while some are loaded while handlers are initialized.)
* There's a variety of template directories that must be configured: `account_validity`, `saml2_config`, `sso`, `email`, `user_consent`.
* Some of those template directories also allow you to customize the names of templates, while others use hard-coded names (e.g. the `account_renewed_html_path` setting vs. SSO uses hard-coded names).

I think there's a few concrete things we could do to improve this situation:
* Use a single template directory and allow the user to configure it in a single place.
* Treat all template files as Jinja (even if they don't have variable expressions there could be other reasons you'd want it to be Jinja).
* Hard-code all template names and document them well.
* (Somewhat unrelated, but...) Fix #6583 so that missing templates don't break Synapse.

richvdh:
> Some of those template directories also allow you to customize the names of templates, while others use hard-coded names (e.g. the account_renewed_html_path setting vs. SSO uses hard-coded names).

I think the settings that allow you to customise the template names are now deprecated?

babolivier:
I'm not aware of deprecation for `account_renewed_html_path`, but it should probably be deprecated.

clokep:
> I think the settings that allow you to customise the template names are now deprecated?

Then this would be a good opportunity to clean them up! 😄

clokep:
Another idea, per https://github.com/matrix-org/synapse/pull/8444#issuecomment-702642606:

[...]

PR #10596, opened 2021-08-13
Centralise the custom template directory

babolivier:
Several configuration sections are using separate settings for custom template directories, which can be confusing. This PR adds a new top-level configuration for a custom template directory which is then used for every module. The only exception is the consent templates, since the consent template directory require a specific hierarchy, so it's probably better that it stays separate from everything else.

TODO:

* [x] ~Deprecate the existing `template_dir` settings~

Based on https://github.com/matrix-org/synapse/pull/10587

Related: https://github.com/matrix-org/synapse/issues/7280


synapse/config/server.py
```diff
@@ -710,6 +710,11 @@ class LimitRemoteRoomsConfig:
             # Turn the list into a set to improve lookup speed.
             self.next_link_domain_whitelist = set(next_link_domain_whitelist)
 
+        templates_config = config.get("templates") or {}
+        self.custom_template_directory = templates_config.get(
+            "custom_template_directory"
+        )
```
richvdh, line 716: we should do some type-checking here: that `templates` is a `dict` and `custom_template_directory` is a string.

synapse/config/account_validity.py
```diff
@@ -78,6 +78,8 @@ def read_config(self, config, **kwargs):
         )
 
         # Read and store template content
+        templates_config = config.get("templates") or {}
+        custom_template_directory = templates_config.get("custom_template_directory")
```
richvdh, line 82: having five copies of this code is unsatisfactory (particularly since we ought to type-check it before using it). Since `ServerConfig` is instantiated first, can we just do `self.root.server.custom_template_directory` ?

richvdh (changes requested):

synapse/config/account_validity.py
babolivier: Oh TIL the root config can be accessed like this. I agree it's a much nicer solution.

richvdh (approved):
lgtm!

PR #10599, opened 2021-08-13
Update CONTRIBUTING.md to fix index links and SyTest instructions

reivilibre:
Signed-off-by: Olivier Wilkinson (reivilibre) <olivier@librepush.net>


reivilibre:
rendered, so you can witness that the links work with your own eyes: https://github.com/matrix-org/synapse/blob/81df7b6cf66d8571ea57326c8764fc69332ea4b0/CONTRIBUTING.md

richvdh (approved):
lgtm.

Probably we should move this into the `docs` directory, so that it can be rendered at https://matrix-org.github.io/synapse/develop, including an *auto-geneated* ToC.

PR #10576, opened 2021-08-11
Move /batch_send to /v2_alpha directory (MSC2716)

MadLittleMods:
Move `/batch_send` endpoint to `/v2_alpha` directory

As pointed out by @erikjohnston, https://github.com/matrix-org/synapse/pull/10552#discussion_r685836624

Part of MSC2716: https://github.com/matrix-org/matrix-doc/pull/2716

### Pull Request Checklist

* [x] Pull request is based on the develop branch
* [x] Pull request includes a [changelog file](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md#changelog). The entry should:
  - Be a short description of your change which makes sense to users. "Fixed a bug that prevented receiving messages from other servers." instead of "Moved X method from `EventStore` to `EventWorkerStore`.".
  - Use markdown where necessary, mostly for `code blocks`.
  - End with either a period (.) or an exclamation mark (!).
  - Start with a capital letter.
* [ ] ~~Pull request includes a [sign off](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md#sign-off)~~
* [x] Code style is correct (run the [linters](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md#code-style))

PR #10600, opened 2021-08-13
Flatten the synapse.rest.client package

reivilibre:
Because the `v1` and `v2_alpha` subpackages are downright LIES.

This will probably cause a couple of conflicts in the import lines, but otherwise should be fairly conflict-free (because I structured it as a `git mv` (pure rename) to start with, before going back and doing the move-with-updated-references).

This should be a commit-by-commit-reviewable PR (since each commit tries to be one mechanical step)


synapse/rest/client/__init__.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2015, 2016 OpenMarket Ltd
```
clokep, line 1: Looks like we gained a year?

Anyway this should get updated to the m.org foundation.

clokep (approved):
LGTM if CI passes.

synapse/rest/client/__init__.py
reivilibre: the __init__ in the subpackages said 2014-2016, so I widened it.
Should this be updated? I don't know the legal interpretation of that, but I thought we kept that around and M.org came in after (so the old copyright still exists).

Not that any of this makes much sense on an empty file, though...

clokep: We update them to the matrix.org foundation for openmarket and new vector.

reivilibre:
Uh oh, a juicy conflict popped up (someone created `v2_alpha/room.py` which conflicts with `v1/room.py` as far as renames go.

clokep:
> Should I merge the files together or leave the v2 one renamed as `roomv2` (this is the name it was imported with anyway)?

`synapse/rest/client/v1/room.py` is already very large (~1.1k LOC), I think renaming `synapse/rest/client/v2_alpha/room.py` makes the most sense since that file is large as well.  (Maybe something like `room_batch.py` or something. @MadLittleMods might have ideas.)

synapse/rest/__init__.py
```diff
@@ -14,49 +14,47 @@
...
+    presence,
+    profile,
+    push_rule,
+    pusher,
     read_marker,
     receipts,
     register,
     relations,
     report_event,
-    room as roomv2,
```
MadLittleMods, line 50: For reference, this change came from https://github.com/matrix-org/synapse/pull/10576

MadLittleMods:
> I think renaming synapse/rest/client/v2_alpha/room.py makes the most sense since that file is large as well. (Maybe something like room_batch.py or something.
>
> -- https://github.com/matrix-org/synapse/pull/10600#issuecomment-899460109

Either `room_historical_batch.py` or `room_batch.py` sound good to me 👍

clokep (approved):
LGTM if CI passes.

PR #10601, opened 2021-08-13
Add types to synapse.util.

reivilibre:


synapse/util/manhole.py
```diff
@@ -135,9 +139,8 @@ def showtraceback(self):
         """
         sys.last_type, sys.last_value, last_tb = ei = sys.exc_info()
         sys.last_traceback = last_tb
-        try:
-            # We remove the first stack item because it is our own code.
-            lines = traceback.format_exception(ei[0], ei[1], last_tb.tb_next)
-            self.write("".join(lines))
-        finally:
-            last_tb = ei = None
```
erikjohnston, line 143: Why do you think it is safe to remove this?

synapse/util/ratelimitutils.py
```diff
@@ -15,33 +15,38 @@
...
         def new_limiter():
             return _PerHostRatelimiter(clock=clock, config=config)
 
-        self.ratelimiters = collections.defaultdict(new_limiter)
+        self.ratelimiters: DefaultDict[
+            str, "_PerHostRatelimiter"
+        ] = collections.defaultdict(new_limiter)
 
-    def ratelimit(self, host):
+    def ratelimit(self, host: str):
```
erikjohnston, line 49:
```suggestion
    def ratelimit(self, host: str) -> None:
```

synapse/util/ratelimitutils.py
```diff
@@ -122,7 +129,7 @@ def _on_enter(self, request_id):
 
         def queue_request():
             if len(self.current_processing) >= self.concurrent_requests:
-                queue_defer = defer.Deferred()
+                queue_defer: defer.Deferred = defer.Deferred()
```
erikjohnston, line 132: It'd be good to give this deferred a type

synapse/util/manhole.py
reivilibre: `last_tb` and `ei` are dead.

can you see a reason why it's like that? It really needs a comment if there's a reason :/

richvdh: the original was cargo-culted from https://github.com/python/cpython/blob/main/Lib/code.py#L150, if that helps at all.

erikjohnston: > `last_tb` and `ei` are dead.

Technically not, an exception raised here keeps the local frame around to allow inspection (which is how things like Sentry can report the local variables of each frame in the stack). 

I don't know if or why that is an important detail (maybe to prevent recursion if the exception thrown goes back via this function? Or just preventing circular references?), but I'm loathed to remove something like that unless we know why it was put there. (Obligatory reference to [Chesterton's fence](https://en.wikipedia.org/wiki/G._K._Chesterton#Chesterton's_fence))

richvdh: > Technically not, an exception raised here keeps the local frame around to allow inspection (which is how things like Sentry can report the local variables of each frame in the stack).

I don't follow this. This method doesn't raise an exception, so how would `last_db` and `ei` be retained in a stack frame?

I think it's worth digging into the history of the cpython code. It looks to me like those lines were introduced before @reivilibre was born, and at the time there was more code where it might have been useful that the traceback be gc-able.

erikjohnston: I was assuming `self.write` could throw, otherwise there's also a question over why it uses `try/finally`

synapse/util/__init__.py
```diff
@@ -133,7 +135,7 @@ def wrapped_callback(*args, **kwargs):
         with context.PreserveLoggingContext():
             return self._reactor.callLater(delay, wrapped_callback, *args, **kwargs)
 
-    def cancel_call_later(self, timer, ignore_errs=False):
+    def cancel_call_later(self, timer: IDelayedCall, ignore_errs=False) -> None:
```
erikjohnston, line 138:
```suggestion
    def cancel_call_later(self, timer: IDelayedCall, ignore_errs: bool = False) -> None:
```

synapse/util/__init__.py
```diff
@@ -78,22 +80,22 @@ class Clock:
...
         """Returns the current system time in seconds since epoch."""
         return self._reactor.seconds()
 
-    def time_msec(self):
+    def time_msec(self) -> int:
         """Returns the current system time in milliseconds since epoch."""
         return int(self.time() * 1000)
 
-    def looping_call(self, f, msec, *args, **kwargs):
+    def looping_call(self, f, msec, *args, **kwargs) -> LoopingCall:
```
erikjohnston, line 98:
```suggestion
    def looping_call(self, f: Callable, msec: int, *args, **kwargs) -> LoopingCall:
```

I think its worth at least giving some of the types here. When 3.10 is released we might be able to use `ParamSpec` for the function bounds here.

synapse/util/manhole.py
```diff
@@ -81,7 +81,10 @@ def manhole(username, password, globals):
     checker = checkers.InMemoryUsernamePasswordDatabaseDontUse(**{username: password})
 
     rlm = manhole_ssh.TerminalRealm()
-    rlm.chainedProtocolFactory = lambda: insults.ServerProtocol(
+    # mypy ignored here because:
+    # - can't deduce types of lambdas
+    # - variable is Type[ServerProtocol], expr is Callable[[], ServerProtocol]
+    rlm.chainedProtocolFactory = lambda: insults.ServerProtocol(  # type: ignore[misc,assignment]
```
erikjohnston, line 87: It's probably better to create a real temporary class here:

```python
class SynapseServerProtocol(insults.ServerProtocol):
    def __init__(self):
        super().__init__(SynapseManhole, dict(globals, __name__="__console__"))
```

I think should work?

synapse/util/ratelimitutils.py
```diff
@@ -177,7 +182,7 @@ def on_both(r):
         ret_defer.addBoth(on_both)
         return make_deferred_yieldable(ret_defer)
 
-    def _on_exit(self, request_id):
+    def _on_exit(self, request_id) -> None:
```
erikjohnston, line 185:
```suggestion
    def _on_exit(self, request_id: object) -> None:
```

synapse/util/retryutils.py
```diff
@@ -51,7 +51,9 @@ def __init__(self, retry_last_ts, retry_interval, destination):
         self.destination = destination
 
 
-async def get_retry_limiter(destination, clock, store, ignore_backoff=False, **kwargs):
+async def get_retry_limiter(
+    destination, clock, store, ignore_backoff=False, **kwargs
```
erikjohnston, line 55: Types for these?

erikjohnston (commented):
Looks good, just a few places where we haven't also added types to the args (we try and type both the args and return types at the same time, rather than having to do two passes of the same functions)

synapse/util/manhole.py
reivilibre: I tried something of that sort, it then said (paraphrased): found `Type[SynapseServerProtocol]` expected `Type[ServerProtocol]`.

reivilibre:
[`cd9a68d` (#10601)](https://github.com/matrix-org/synapse/pull/10601/commits/cd9a68de5a842e922cf34313e0a51aac17a9b804) is a functional change but it's only in the unit tests so hopefully not a problem (before, we ignored wrongly-named arguments!).

synapse/rest/client/register.py
```diff
@@ -330,11 +330,11 @@ def __init__(self, hs: "HomeServer"):
                 # Artificially delay requests if rate > sleep_limit/window_size
                 sleep_limit=1,
                 # Amount of artificial delay to apply
-                sleep_msec=1000,
+                sleep_delay=1000,
                 # Error with 429 if more than reject_limit requests are queued
                 reject_limit=1,
                 # Allow 1 request at a time
-                concurrent_requests=1,
+                concurrent=1,
```
erikjohnston, line 337: This is a functional change too? I think its fine though

synapse/rest/client/register.py
```diff
@@ -763,7 +763,10 @@ async def _create_registration_details(
         Returns:
              dictionary for response from /register
         """
-        result = {"user_id": user_id, "home_server": self.hs.hostname}
+        result: Dict[str, Union[str, int]] = {
```
erikjohnston, line 766:
```suggestion
        result: JsonDict = {
```

May as well label it a `JsonDict` as that is what we're returning

synapse/rest/client/register.py
```diff
@@ -814,7 +817,7 @@ async def _do_guest_registration(
             user_id, device_id, initial_display_name, is_guest=True
         )
 
-        result = {
+        result: Dict[str, Union[int, str]] = {
```
erikjohnston, line 820:
```suggestion
        result: JsonDict = {
```

synapse/util/async_helpers.py
```diff
@@ -550,19 +550,21 @@ def failure_cb(val):
     return new_d
 
 
+# This class can't be generic because it uses slots with attrs.
+# See: https://github.com/python-attrs/attrs/issues/313
```
erikjohnston, line 554: Looks like rich worked around this in https://github.com/matrix-org/synapse/commit/d15d241f5ebc978328494ee350d4a0ef8e0070f5

synapse/util/caches/__init__.py
```diff
@@ -64,32 +64,32 @@ class CacheMetric:
...
-    def inc_hits(self):
+    def inc_hits(self) -> None:
         self.hits += 1
 
-    def inc_misses(self):
+    def inc_misses(self) -> None:
         self.misses += 1
 
-    def inc_evictions(self, size=1):
+    def inc_evictions(self, size=1) -> None:
```
erikjohnston, line 73:
```suggestion
    def inc_evictions(self, size: int = 1) -> None:
```

synapse/util/caches/lrucache.py
```diff
@@ -341,7 +341,7 @@ def __init__(
         # Default `clock` to something sensible. Note that we rename it to
         # `real_clock` so that mypy doesn't think its still `Optional`.
         if clock is None:
-            real_clock = Clock(reactor)
+            real_clock = Clock(reactor)  # type: ignore[arg-type]
```
erikjohnston, line 344: Why does this complain?

synapse/util/manhole.py
```diff
@@ -129,20 +133,26 @@ def showsyntaxerror(self, filename=None):
...
         try:
             # We remove the first stack item because it is our own code.
             lines = traceback.format_exception(ei[0], ei[1], last_tb.tb_next)
             self.write("".join(lines))
         finally:
-            last_tb = ei = None
+            # On the line below, last_tb and ei appear to be dead.
+            # It's unclear whether there is a reason behind this line.
+            # It conceivably could be because an exception raised in this block
+            # will keep the local frame (containing these local variables) around.
```
erikjohnston, line 154: Can you note this implementation is lifted straight from the python source code? Otherwise it makes it sound like you're writing code you don't understand :)

synapse/util/retryutils.py
```diff
@@ -150,10 +160,10 @@ def __init__(
         self.backoff_on_404 = backoff_on_404
         self.backoff_on_failure = backoff_on_failure
 
-    def __enter__(self):
+    def __enter__(self) -> None:
         pass
 
-    def __exit__(self, exc_type, exc_val, exc_tb):
+    def __exit__(self, exc_type, exc_val, exc_tb: TracebackType) -> None:
```
erikjohnston, line 166: Aaaaaand the other args?

synapse/util/templates.py
```diff
@@ -56,7 +56,8 @@ def build_jinja_env(
     if autoescape is None:
         autoescape = jinja2.select_autoescape()
 
-    loader = jinja2.FileSystemLoader(template_search_directories)
+    # the type signature of this is wrong
+    loader = jinja2.FileSystemLoader(template_search_directories)  # type: ignore[arg-type]
```
erikjohnston, line 60: Is it? It looks like it should accept a `Sequence[str]`? Which suggests *our* types our wrong

synapse/util/wheel_timer.py
```diff
@@ -11,17 +11,20 @@
...
 
+T = TypeVar("T")
 
-class _Entry:
+
+class _Entry(Generic[T]):
     __slots__ = ["end_key", "queue"]
 
-    def __init__(self, end_key):
+    def __init__(self, end_key) -> None:
```
erikjohnston, line 22:
```suggestion
    def __init__(self, end_key: T) -> None:
```

?

synapse/util/async_helpers.py
```diff
@@ -268,7 +268,7 @@ def __init__(
         if not clock:
             from twisted.internet import reactor
 
-            clock = Clock(reactor)
+            clock = Clock(reactor)  # type: ignore[arg-type]
```
erikjohnston, line 271: Why?

synapse/util/caches/treecache.py
```diff
@@ -35,17 +35,17 @@ class TreeCache:
...
 
-    def __init__(self):
-        self.size = 0
+    def __init__(self) -> None:
+        self.size: int = 0
         self.root = TreeCacheNode()
 
-    def __setitem__(self, key, value):
-        return self.set(key, value)
+    def __setitem__(self, key, value) -> None:
```
erikjohnston, line 42: It'd be nice if we could have typed this while we're here, but this is already a *huge* PR.

erikjohnston (commented):
Well this has turned into a bit of a mammoth of a PR, I probably should've insisted on not doing the entire utils package at once. Ah well.

synapse/rest/client/register.py
reivilibre: Sorry, forgot to point it out

synapse/rest/client/register.py
reivilibre: TIL :)

synapse/util/async_helpers.py
reivilibre: NOPE, that's a trap by Rich! I found that on the attrs bug tracker, did the same thing, only to realise that Rich didn't actually get it to work either.

(GitHub does say that it's not part of any branch .... I should have thought more deeply about what that meant :P)

erikjohnston: Oh, hahaha. I just assumed it was on a PR that got squash merged and then the branch got deleted.

richvdh: :-S

synapse/util/caches/lrucache.py
reivilibre: the import type is `module`, not `IReactorTime`. I suppose this might be better as a `cast`, so I'll do that.

synapse/util/retryutils.py
reivilibre: I didn't know how to type these, but actually I think I can do it with `cast`

synapse/util/templates.py
reivilibre: True. I suppose the types are an interface detail, and the actual implementation-accepted types can change. Will address.

synapse/util/wheel_timer.py
reivilibre: think this is an `int` index actually

synapse/util/caches/treecache.py
reivilibre: will come back in another PR to pick this up :)

reivilibre: will come back in another PR to pick this one up :)

synapse/util/templates.py
erikjohnston: I'm don't understand what you mean there, sorry?

synapse/util/retryutils.py
erikjohnston: The types are `exc_type: Optional[Type[BaseException]], exc_val: Optional[BaseException],  exc_tb: Optional[TracebackType]` by the sounds of things:  https://github.com/python/typeshed/blob/master/stdlib/typing.pyi#L426

synapse/util/templates.py
reivilibre: Jinja2 currently accepts `Iterable[str]` but says it will accept `Sequence[str]`.
So morally speaking, they might change the implementation that breaks something so that `Iterable[str]` doesn't work anymore, and it would be silly us for not trusting their type annotations.
(Hence I've converted our iterator to a `List`.)

synapse/util/async_helpers.py
```diff
@@ -268,7 +269,7 @@ def __init__(
         if not clock:
             from twisted.internet import reactor
 
-            clock = Clock(reactor)  # type: ignore[arg-type]
+            clock = Clock(cast(IReactorTime, reactor))
```
erikjohnston, line 272: Does it work if you add a `assert isinstance(reactor, IReactorTime)` above?

synapse/util/retryutils.py
```diff
@@ -172,6 +177,7 @@ def __exit__(self, exc_type, exc_val, exc_tb: TracebackType) -> None:
             # failures; this is mostly so as not to catch defer._DefGen.
             valid_err_code = True
         elif issubclass(exc_type, CodeMessageException):
+            exc_val_cme = cast(CodeMessageException, exc_val)
```
erikjohnston, line 180: We may as well replace the `issubclass` with an `isinstance(exc_val, CodeMessageException)` rather than casting?

erikjohnston (commented):
Casting is a bit evil, its better to assert where possible, though not sure if that works for the reactor stuff.

synapse/util/async_helpers.py
reivilibre: Mypy likes it, but `assert isinstance(reactor, IReactorTime)` gives an AssertionError at runtime :(.

EPollReactor has `ReactorBase` in its ancestry which implements `IReactorTime`, so it seems like doing `assert isinstance(reactor, ReactorBase)` is actually what we want (and appears to work).

erikjohnston (approved):

PR #10604, opened 2021-08-14
Better syntax in the shell script

joao-vitor-sr:
Now the shell script avoid errors like word splitting, gobbling, and the Unicode was disabled to improve the performance

### DIFF
--- a/changelog.d/10588.removal
+++ b/changelog.d/10588.removal
@@ -1,0 +1,1 @@
+No longer build `.dev` packages for Ubuntu 20.10 LTS Groovy Gorilla, which has now EOLed.
--- a/scripts-dev/build_debian_packages
+++ b/scripts-dev/build_debian_packages
@@ -25,7 +25,6 @@ DISTS = (
     "debian:sid",
     "ubuntu:bionic",  # 18.04 LTS (our EOL forced by Py36 on 2021-12-23)
     "ubuntu:focal",  # 20.04 LTS (our EOL forced by Py38 on 2024-10-14)
-    "ubuntu:groovy",  # 20.10 (EOL 2021-07-07)
     "ubuntu:hirsute",  # 21.04 (EOL 2022-01-05)
 )
 


demo/start.sh
```diff
@@ -1,24 +1,29 @@
 #!/usr/bin/env bash
 
+# Disable unicode.
+LC_ALL=C
+LANG=C
```
richvdh, line 5: I've never heard of needing to do this before. Can you provide any references - what does this change, why is it a good thing to do, etc?

demo/start.sh
```diff
@@ -1,24 +1,29 @@
...
+
 DIR="$( cd "$( dirname "$0" )" && pwd )"
 
 CWD=$(pwd)
 
-cd "$DIR/.."
+cd "$DIR/.." || exit
```
richvdh, line 11: I note that you have gone through and added `|| exit` to commands that change directory (`cd`, `pushd`, `popd`). It's true that these scripts could do with more error-checking, but this approach seems strange. These are far from the only commands that can fail - indeed, they are generally the least likely to fail. (`cd $(pwd)/..` can only fail in the most unusual circumstances). Why did you pick out these particular commands for error-checking? Wouldn't it make more sense just to put `set -e` at the top of the scripts?

demo/start.sh
```diff
@@ -1,24 +1,29 @@
...
 
 mkdir -p demo/etc
 
-export PYTHONPATH=$(readlink -f $(pwd))
+export PYTHONPATH
+PYTHONPATH=$(readlink -f "$(pwd)")
```
richvdh, line 16: these two are equivalent, as far as I can tell.

demo/start.sh
```diff
@@ -27,12 +32,13 @@ for port in 8080 8081 8082; do
         --config-path "$DIR/etc/$port.config" \
         --report-stats no
 
-    if ! grep -F "Customisation made by demo/start.sh" -q  $DIR/etc/$port.config; then
-        printf '\n\n# Customisation made by demo/start.sh\n' >> $DIR/etc/$port.config
-
-        echo "public_baseurl: http://localhost:$port/" >> $DIR/etc/$port.config
+    if ! grep -F "Customisation made by demo/start.sh" -q  "$DIR"/etc/$port.config; then
```
richvdh, line 36:
```suggestion
```

demo/start.sh
```diff
@@ -27,12 +32,13 @@ for port in 8080 8081 8082; do
...
 
-        echo 'enable_registration: true' >> $DIR/etc/$port.config
+        {
+          printf '\n\n# Customisation made by demo/start.sh\n'
+          echo "public_baseurl: http://localhost:$port/"
+          echo 'enable_registration: true'
```
richvdh, line 40:
```suggestion
            printf '\n\n# Customisation made by demo/start.sh\n'
            echo "public_baseurl: http://localhost:$port/"
            echo 'enable_registration: true'
```

demo/start.sh
```diff
@@ -56,26 +62,32 @@ for port in 8080 8081 8082; do
...
+          echo 'federation_verify_certificates: false'
+        } >> "$DIR"/etc/$port.config
 
         # Set tls paths
-        echo "tls_certificate_path: \"$DIR/etc/localhost:$https_port.tls.crt\"" >> $DIR/etc/$port.config
-        echo "tls_private_key_path: \"$DIR/etc/localhost:$https_port.tls.key\"" >> $DIR/etc/$port.config
+        {
+          echo "tls_certificate_path: \"$DIR/etc/localhost:$https_port.tls.crt\""
+          echo "tls_private_key_path: \"$DIR/etc/localhost:$https_port.tls.key\""
+        } >> "$DIR"/etc/$port.config
```
richvdh, line 79: what is the logic behind merging *some* of the `printf`/`echo` commands into `{ ... } ` blocks? If you're going to do that, why not just make one big `{ ... }` block?

demo/start.sh
```diff
@@ -56,26 +62,32 @@ for port in 8080 8081 8082; do
...
 
         # Generate tls keys
-        openssl req -x509 -newkey rsa:4096 -keyout $DIR/etc/localhost\:$https_port.tls.key -out $DIR/etc/localhost\:$https_port.tls.crt -days 365 -nodes -subj "/O=matrix"
+        openssl req -x509 -newkey rsa:4096 -keyout "$DIR"/etc/localhost"$(printf '\:')"$https_port.tls.key -out "$DIR"/etc/localhost"$(printf '\:')"$https_port.tls.crt -days 365 -nodes -subj "/O=matrix"
```
richvdh, line 82: I assume you've tested this? It doesn't look like it will do the right thing to me.

scripts-dev/complement.sh
```diff
@@ -65,4 +70,4 @@ if [[ -n "$1" ]]; then
 fi
 
 # Run the tests!
-go test -v -tags synapse_blacklist,msc2946,msc3083,msc2403,msc2716 -count=1 $EXTRA_COMPLEMENT_ARGS ./tests/...
+go test -v -tags synapse_blacklist,msc2946,msc3083,msc2403,msc2716 -count=1 "$EXTRA_COMPLEMENT_ARGS" ./tests/...
```
richvdh, line 73: this is incorrect. `EXTRA_COMPLEMENT_ARGS` contains multiple words.

scripts-dev/next_github_number.sh
```diff
@@ -4,6 +4,6 @@ set -e
 
 # Fetch the current GitHub issue number, add one to it -- presto! The likely
 # next PR number.
-CURRENT_NUMBER=`curl -s "https://api.github.com/repos/matrix-org/synapse/issues?state=all&per_page=1" | jq -r ".[0].number"`
+CURRENT_NUMBER=$(curl -s "https://api.github.com/repos/matrix-org/synapse/issues?state=all&per_page=1" | jq -r ".[0].number")
```
richvdh, line 7: these are equivalent, afaict.

test_postgresql.sh
```diff
@@ -3,10 +3,14 @@
...
+LANG=C
+
 set -e
 
 # Build, and tag
 docker build docker/ -f docker/Dockerfile-pgtests -t synapsepgtests
 
 # Run, mounting the current directory into /src
-docker run --rm -it -v $(pwd)\:/src synapsepgtests
+docker run --rm -it -v "$(pwd)""$(printf '\:')"/src synapsepgtests
```
richvdh, line 16: again, have you tested this?

richvdh (changes requested):
Thank you for the contribution!

There are some valid changes here - but there are also some things that seem to be unnecessary.  There are also some things that are definitely *wrong*, so I'm worried that you haven't tested your changes.

richvdh:
By the way, we also require sign-off from contributors before we can accept PRs - please see https://github.com/matrix-org/synapse/blob/develop/CONTRIBUTING.md#sign-off and add a `Signed-off-by` comment to this PR.

Finally, when you update this PR, please remember *not* to rebase your changes but just add new commits to your branch.

DMRobertson:
Could think about running shellcheck (https://www.shellcheck.net/) as part of the linting steps --- perhaps it can catch other things like this?

dkasak:
Indeed. The changes above sound exactly as the list shellcheck would produce.

demo/start.sh
joao-vitor-sr: this section is good for optimization,
in this script the Unicode is not necessary anywhere so is useful to disable the Unicode for performance optimization
## Some references
[https://tldp.org/LDP/abs/html/optimizations.html
](https://tldp.org/LDP/abs/html/optimizations.html)
[https://github.com/dylanaraps/pure-bash-bible#disable-unicode](https://github.com/dylanaraps/pure-bash-bible#disable-unicode)

demo/start.sh
joao-vitor-sr: Is good to declare and assign separately to avoid masking return values
# References
[https://github.com/koalaman/shellcheck/wiki/SC2155](https://github.com/koalaman/shellcheck/wiki/SC2155)

richvdh: TIL, thanks!

demo/start.sh
richvdh: It seems like it only makes a difference if the script is doing repeated string matching. I'm not in favour of cluttering these scripts like this unless it actually makes any difference.

Issue #10603, opened 2021-08-13
Unable to login after enabling hidden read receipts (MSC2285)


aaronraimist:
### Description

I enabled MSC2285 in Synapse and Element yesterday and everything seemed to be working properly. Then today I randomly got logged out of Element (unsure if browser issue or Synapse/Element bug) and when I tried to sign back in /sync kept returning errors.

```
Aug 13 20:35:51 michaelpanderson matrix-synapse-worker-generic_worker-18111[2353406]: During handling of the above exception, another exception occurred:
Aug 13 20:35:51 michaelpanderson matrix-synapse-worker-generic_worker-18111[2353406]: Traceback (most recent call last):
Aug 13 20:35:51 michaelpanderson matrix-synapse-worker-generic_worker-18111[2353406]:   File "/usr/local/lib/python3.8/site-packages/twisted/internet/defer.py", line 1661, in _inlineCallbacks
Aug 13 20:35:51 michaelpanderson matrix-synapse-worker-generic_worker-18111[2353406]:     result = current_context.run(gen.send, result)
Aug 13 20:35:51 michaelpanderson matrix-synapse-worker-generic_worker-18111[2353406]:   File "/usr/local/lib/python3.8/site-packages/synapse/handlers/sync.py", line 353, in _wait_for_sync_for_user
Aug 13 20:35:51 michaelpanderson matrix-synapse-worker-generic_worker-18111[2353406]:     result: SyncResult = await self.current_sync_for_user(
Aug 13 20:35:51 michaelpanderson matrix-synapse-worker-generic_worker-18111[2353406]:   File "/usr/local/lib/python3.8/site-packages/synapse/handlers/sync.py", line 397, in current_sync_for_user
Aug 13 20:35:51 michaelpanderson matrix-synapse-worker-generic_worker-18111[2353406]:     sync_result = await self.generate_sync_result(
Aug 13 20:35:51 michaelpanderson matrix-synapse-worker-generic_worker-18111[2353406]:   File "/usr/local/lib/python3.8/site-packages/synapse/handlers/sync.py", line 1063, in generate_sync_result
Aug 13 20:35:51 michaelpanderson matrix-synapse-worker-generic_worker-18111[2353406]:     res = await self._generate_sync_entry_for_rooms(
[...]

PR #10606, opened 2021-08-14
Handle string read receipt data

SimonBrandner:
Fixes https://github.com/matrix-org/synapse/issues/10603

### DIFF
--- a/changelog.d/10606.bugfix
+++ b/changelog.d/10606.bugfix
@@ -1,1 +1,1 @@
-Fix errors on /sync when read receipt data is a string with MSC2285 enabled. Contributed by @SimonBrandner.
+Fix errors on /sync when read receipt data is a string. Only affects homeservers with the experimental flag for [MSC2285](https://github.com/matrix-org/matrix-doc/pull/2285) enabled. Contributed by @SimonBrandner.
--- a/changelog.d/10606.bugfix
+++ b/changelog.d/10606.bugfix
@@ -1,0 +1,1 @@
+Fix errors on /sync when read receipt data is a string with MSC2285 enabled. Contributed by @SimonBrandner.
--- a/synapse/handlers/receipts.py
+++ b/synapse/handlers/receipts.py
@@ -190,19 +190,20 @@ class ReceiptEventSource:
                 for rr_user_id, user_rr in m_read.items():
                     try:
                         hidden = user_rr.get("hidden", None)
-                        if hidden is not True or rr_user_id == user_id:
-                            new_users[rr_user_id] = user_rr.copy()
-                            # If hidden has a value replace hidden with the correct prefixed key
-                            if hidden is not None:
-                                new_users[rr_user_id].pop("hidden")
-                                new_users[rr_user_id][
-                                    ReadReceiptEventFields.MSC2285_HIDDEN
-                                ] = hidden
                     except AttributeError:
                         # Due to https://github.com/matrix-org/synapse/issues/10376
                         # there are cases where user_rr is a string, in those cases
                         # we just ignore the read receipt
-                        pass
+                        continue
+
+                    if hidden is not True or rr_user_id == user_id:
+                        new_users[rr_user_id] = user_rr.copy()
+                        # If hidden has a value replace hidden with the correct prefixed key
+                        if hidden is not None:
+                            new_users[rr_user_id].pop("hidden")
+                            new_users[rr_user_id][
+                                ReadReceiptEventFields.MSC2285_HIDDEN
+                            ] = hidden
 
                 # Set new users unless empty
                 if len(new_users.keys()) > 0:
--- a/synapse/handlers/receipts.py
+++ b/synapse/handlers/receipts.py
@@ -188,15 +188,22 @@ class ReceiptEventSource:
 
                 new_users = {}
                 for rr_user_id, user_rr in m_read.items():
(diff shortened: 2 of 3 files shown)


synapse/handlers/receipts.py
```diff
@@ -188,15 +188,22 @@ def filter_out_hidden(events: List[JsonDict], user_id: str) -> List[JsonDict]:
...
+                            if hidden is not None:
+                                new_users[rr_user_id].pop("hidden")
+                                new_users[rr_user_id][
+                                    ReadReceiptEventFields.MSC2285_HIDDEN
+                                ] = hidden
+                    except AttributeError:
+                        # Due to https://github.com/matrix-org/synapse/issues/10376
+                        # there are cases where user_rr is a string, in those cases
+                        # we just copy the user_rr
                         new_users[rr_user_id] = user_rr.copy()
```
SimonBrandner, line 205: Perhaps we should just ignore these read receipts or delete their data? I am not completely sure what is the best thing to do here

synapse/handlers/receipts.py
```diff
@@ -188,15 +188,22 @@ def filter_out_hidden(events: List[JsonDict], user_id: str) -> List[JsonDict]:
...
+                                new_users[rr_user_id].pop("hidden")
+                                new_users[rr_user_id][
+                                    ReadReceiptEventFields.MSC2285_HIDDEN
+                                ] = hidden
+                    except AttributeError:
+                        # Due to https://github.com/matrix-org/synapse/issues/10376
+                        # there are cases where user_rr is a string, in those cases
+                        # we just copy the user_rr
+                        new_users[rr_user_id] = user_rr
+                        pass
```
SimonBrandner, line 206: Perhaps we should just ignore these read receipts or delete their data? I am not completely sure what is the best thing to do here

aaronraimist:
I have tested this on my server and can confirm it fixes the issue. Thanks!

synapse/handlers/receipts.py
```diff
@@ -188,15 +188,22 @@ def filter_out_hidden(events: List[JsonDict], user_id: str) -> List[JsonDict]:
...
-                    if hidden is not True or rr_user_id == user_id:
-                        new_users[rr_user_id] = user_rr.copy()
-                        # If hidden has a value replace hidden with the correct prefixed key
-                        if hidden is not None:
-                            new_users[rr_user_id].pop("hidden")
-                            new_users[rr_user_id][
-                                ReadReceiptEventFields.MSC2285_HIDDEN
-                            ] = hidden
+                    try:
+                        hidden = user_rr.get("hidden", None)
```
anoadragon453, line 192:
```suggestion
                        hidden = user_rr.get("hidden")
```

`None` is the default value for the second argument of `get` (might as well fix while we're here).

synapse/handlers/receipts.py
anoadragon453, line 206: We should probably only surround `hidden = user_rr.get("hidden", None)` in the try/except.

anoadragon453: It's probably best to avoid sending them to clients. We can't expect clients to handle this particularly invalid event shape.

I would just ignore the receipt.

tests/handlers/test_receipts.py
```diff
@@ -286,6 +286,36 @@ def test_filters_out_receipt_event_with_only_hidden_receipt_and_ignores_rest(sel
             ],
         )
 
+    def test_handles_string_data(self):
```
anoadragon453, line 289: Could you add a docstring to this function that notes that this is unintended behaviour? i.e:

```python
def test_handles_string_data(self):
    """
    Tests that an invalid shape for read-receipts is handled.
    Context: https://github.com/matrix-org/synapse/issues/10603
```

anoadragon453 (changes requested):
Thank you for this, and for adding tests!

Overall, as read receipts aren't part of the room DAG, my thought is that this fix is great for now (let's get people's sync working again ASAP!). But in the long term, we should add structural validation of read receipts on both the client and federation side, and drop the bad data from the database.

This likely won't be the last time this particular invalid data data will break a change to the way we handle read receipts.

synapse/handlers/receipts.py
SimonBrandner: Yeah, makes senes :+1:

SimonBrandner:
> Overall, as read receipts aren't part of the room DAG, my thought is that this fix is great for now (let's get people's sync working again ASAP!). But in the long term, we should add structural validation of read receipts on both the client and federation side, and drop the bad data from the database.

Yeah, I definitely agree, this should get fixed properly but I wanted a quick fix to land. And I don't think I know the codebase well enough to do this the proper way :sweat_smile:

anoadragon453 (approved):
> Yeah, I definitely agree, this should get fixed properly but I wanted a quick fix to land.

Yep, definitely the way to do it!

This lgtm now, thanks!

changelog.d/10606.bugfix
```diff
@@ -0,0 +1 @@
+Fix errors on /sync when read receipt data is a string with MSC2285 enabled. Contributed by @SimonBrandner.
```
anoadragon453, line 1:
```suggestion
Fix errors on /sync when read receipt data is a string. Only affects homeservers with the experimental flag for [MSC2285](https://github.com/matrix-org/matrix-doc/pull/2285) enabled. Contributed by @SimonBrandner.
```

PR #10608, opened 2021-08-15
Type hints for proxyagent

dklimpel:
Simplified `pool_key` to a string instead a `Tuple`.

Information about the key in twisted are there:
https://twistedmatrix.com/documents/current/api/twisted.web.client.HTTPConnectionPool.html#getConnection
`A unique key identifying connections that can be used interchangeably.`

https://github.com/twisted/twisted/blob/fcd8f3cac8c349cfee3f3068e45d8b4f810394b7/src/twisted/web/client.py#L1380-L1381
```
@param key: A unique key identifying connections that can be used
    interchangeably.
```

### Pull Request Checklist


synapse/http/federation/srv_resolver.py
```diff
@@ -103,11 +103,16 @@ class SrvResolver:
...
-        cache (dict): cache object
-        get_time (callable): clock implementation. Should return seconds since the epoch
+        cache: cache object
+        get_time: clock implementation. Should return seconds since the epoch
     """
 
-    def __init__(self, dns_client=client, cache=SERVER_CACHE, get_time=time.time):
+    def __init__(
+        self,
+        dns_client=client,
```
clokep, line 112: You'll probably need to import this.

```suggestion
        dns_client: IResolver = client,
```

(Also please remove it from the docstring.)

synapse/http/federation/srv_resolver.py
```diff
@@ -28,7 +28,7 @@
 
 logger = logging.getLogger(__name__)
 
-SERVER_CACHE = {}
+SERVER_CACHE: Dict = {}
 
 
 @attr.s(slots=True, frozen=True)
```
clokep, line 34: Maybe make this `auto_attribs=True` while we're at it. We could do something like the following for attributes without a default:

```python
host: bytes
```

And for ones with a default:

```python
expires: int = 0
```

synapse/http/federation/srv_resolver.py
```diff
@@ -28,7 +28,7 @@
 
 logger = logging.getLogger(__name__)
 
-SERVER_CACHE = {}
+SERVER_CACHE: Dict = {}
```
clokep, line 31: Can we make this more specific? (What is this a mapping of?)

synapse/http/federation/srv_resolver.py
```diff
@@ -103,11 +103,16 @@ class SrvResolver:
...
+        cache: cache object
+        get_time: clock implementation. Should return seconds since the epoch
     """
 
-    def __init__(self, dns_client=client, cache=SERVER_CACHE, get_time=time.time):
+    def __init__(
+        self,
+        dns_client=client,
+        cache: Dict = SERVER_CACHE,
+        get_time: Callable = time.time,
```
clokep, line 114: I think?

```suggestion
        get_time: Callable[[], int] = time.time,
```

synapse/http/proxyagent.py
```diff
@@ -198,8 +198,9 @@ def request(
                     self.http_proxy_creds.as_proxy_authorization_value(),
                 )
             # Cache *all* connections under the same key, since we are only
-            # connecting to a single destination, the proxy:
-            pool_key = ("http-proxy", self.http_proxy_endpoint)
+            # connecting to a single destination, the proxy
+            # The URL of proxy is not important for the key
```
clokep, line 202: I think the original comment was clear enough.

```suggestion
            # connecting to a single destination, the proxy.
```

clokep (changes requested):
Looks good overall, but I think some of the type hints can be made more specific.

synapse/http/federation/srv_resolver.py
dklimpel: I tried this and got new errors. I will reproduce it an post it here.

dklimpel: With `dns_client: IResolver = client,`

I get this:
```
synapse/http/federation/srv_resolver.py:113: error: Incompatible default for argument "dns_client" (default has type Module, argument has type "IResolver")  [assignment]
synapse/http/federation/srv_resolver.py:143: error: Missing positional argument "timeout" in call to "lookupService" of "IResolver"  [call-arg]
synapse/http/federation/srv_resolver.py:143: error: Argument 1 to "lookupService" of "IResolver" has incompatible type "bytes"; expected "str"  [arg-type]
```

synapse/http/federation/srv_resolver.py
dklimpel: An example
https://github.com/matrix-org/synapse/blob/4b965c862dc66c0da5d3240add70e9b5f0aa720b/tests/http/federation/test_srv_resolver.py#L77-L85

synapse/http/federation/srv_resolver.py
```diff
@@ -28,35 +28,35 @@
 
 logger = logging.getLogger(__name__)
 
-SERVER_CACHE = {}
+SERVER_CACHE: Dict[bytes, List] = {}
```
clokep, line 31:
```suggestion
SERVER_CACHE: Dict[bytes, List["Server"]] = {}
```

synapse/http/federation/srv_resolver.py
```diff
@@ -28,35 +28,35 @@
...
-    weight = attr.ib(default=0)
-    expires = attr.ib(default=0)
+    host: bytes
+    port: int
+    priority: int = 0
+    weight: int = 0
+    expires: int = 0
 
 
 def _sort_server_list(server_list):
```
clokep, line 55: I think?

```suggestion
def _sort_server_list(server_list: List[Server]) -> List[Server]:
```

synapse/http/federation/srv_resolver.py
clokep: That's unfortunate, it seems that some of the Twisted types around DNS are odd. You can leave it for now and we can take another look in the future.

clokep:
Sorry for the delay! I think this looks good minus the last two suggestions I made!

changelog.d/10608.misc
```diff
@@ -0,0 +1 @@
+Type hints for proxyagent and srv_resolver.
```
clokep, line 1:
```suggestion
Improve type hints for the proxy agent and SRV resolver modules. Contributed by @dklimpel.
```

clokep (approved):

PR #10612, opened 2021-08-16
Build debs for bookworm

### LINKED ISSUE #10609: build packages for debian bookworm
bullseye is now released as debian stable, so we should start building debs for new testing

richvdh:
Fixes #10609


DMRobertson:
Do we need to update CI to be bookworm-aware? Didn't fully follow all of the details, but I think this will happen for free once `debian:testing` points to bookworm and we run sytest's build.sh/push.sh?

richvdh:
hum; https://hub.docker.com/_/debian does not yet include `bookworm`, so I think this will need to be put on hold.

richvdh:
> Do we need to update CI to be bookworm-aware?

I don't think there's any particular rush on that. We already run sytest against a synapse running on `testing`. The main thing is that we cover a range of python and postgres versions. If testing/bookworm starts to diverge from bullseye, we might want to consider adding extra tests against bullseye, but that can happen later.

richvdh:
> hum; https://hub.docker.com/_/debian does not yet include `bookworm`, so I think this will need to be put on hold.

there is a `testing`, of course, but using it would require some hoop-jumping. I think we'll just wait.

richvdh:
there's now a `bookworm` image, so this is unblocked.

erikjohnston (approved):

Issue #10523, opened 2021-08-04
Spaces summary API sometimes has poor performance


grinapo:
### Description
Spaces room list (`_matrix/client/unstable/org.matrix.msc2946/rooms/<room_id>`) takes considerable amount of time to serve (20-60 seconds), which does not seem to be either justified or desirable. (A good example is #osm-space:matrix.org, which contains rooms and spaces as well and takes 20-30 seconds.)

First, room list is quite static most of the time, so caching it (and invalidate cache on change) seems to be desirable, especially if the long query time is justified.

However I do not see any real resource usage there (admittedly without analysing the code): both cpu and psql seems not to be busy.

### Steps to reproduce

- see space room list
- observe the spinner

### Version information
- **Homeserver**: grin.hu
- **Version**: 1.38.1
-

clokep:
There's been reports of this, but I haven't yet had time to look at why it is slow. I don't plan to look at this until after #10495 since the API is changing anyway.

clokep:
Now that #10495 has landed, we can take a look at this. I'll likely only handle this for the new `/hierarchy` endpoint, not the original `/spaces` endpoint, but overall that should make things better moving forward.

My plan here is two pronged:

* [ ] Support running the `/hierarchy` endpoint on multiple workers:
    * [ ] Persist the pagination sessions to the database.
    * [ ] (Maybe?) Update something about fetching data over federation (i.e. do we care about the current sharding?)
    * [ ] Update the worker documentation.
* [ ] Explore caching of requests made over federation.

clokep:
Another thing to do here, if necessary, might be to store the space relationships in a dedicated table. I'd like to see all of this running for a bit before doing things such as that!

clokep:
The last remaining bit of this (for now) is to test that running this on workers is fine (and document that). Once this is in the wild for a bit we might see unexpected performance issues though.

clokep:
Most of the changes discussed in here are in Synapse 1.42.0, although 1.43.0 will include some more changes (worker support, properly stripping the `m.space.child` state events).

As of now, I think I'd like to see if those have good impact on the endpoint before trying to do more optimization. I also might need additional help getting profiles from smaller servers as I *think* the performance depends quite a bit on what data is being fetched over federation.

PR #10613, opened 2021-08-16
Persist room hierarchy pagination sessions to the database.

clokep:
Part of #10523.

This starts to support running the `/hierarchy` endpoint on multiple workers by persisting pagination state to the database.

## Open questions:

* Does persisting this to the database make sense? (Would it make more sense to persist it to Redis?)
* How do we reasonably share responses over federation among workers?
	* Can we put the responses in Redis/database?
	* Should we be routing those to the proper federation worker and let that cache it in memory?

This somewhat requires fixing matrix-org/complement#194 before I have high confidence in actually saying we can run this as workers, but this is definitely part of the puzzle. (This change also helps with losing state across restart, which is desirable regardless.)


clokep:
I'm requesting review on this more for discussion of the open questions than an actual code review (although someone taking a look and making sure I'm not doing something crazy would be nice too!)

richvdh:
Inst

Regarding "How do we reasonably share responses over federation among workers?" - I'm not sure I quite understand the question. Are you talking about a 

>     * Does persisting this to the database make sense? (Would it make more sense to persist it to Redis?)

Instinctively, my first reaction to this is "yes, this should be done in Redis rather than the database". Given we already have an `ExternalCache` class, I'd hope this wouldn't be too hard. However, perhaps we need to consider the behaviour when Synapse/Redis are restarted at the same time - is it acceptable for clients to see pagination errors in such a case?

>     * How do we reasonably share responses over federation among workers?
>       
>       * Can we put the responses in Redis/database?
>       * Should we be routing those to the proper federation worker and let that cache it in memory?

I think I need more context here. Are you talking about caching the response from an external server after we make a callout for information on a room? Do we really *need* to share such a cache between workers?

It might be worth taking these questions to the mailing list, for more detailed discussion?

clokep:
> > * How do we reasonably share responses over federation among workers?
> >   
> >   * Can we put the responses in Redis/database?
> >   * Should we be routing those to the proper federation worker and let that cache it in memory?
> 
> I think I need more context here. Are you talking about caching the response from an external server after we make a callout for information on a room? Do we really _need_ to share such a cache between workers?

I think I might be over-engineering this bit. We don't seem to do this in other places and just let each worker call out over federation (e.g. `get_pdu` has a cache that isn't attempted to be shared. See #10647). I'm just going to punt this for now and say "Maybe in the future".

clokep:
> > * Does persisting this to the database make sense? (Would it make more sense to persist it to Redis?)
> 
> Instinctively, my first reaction to this is "yes, this should be done in Redis rather than the database". Given we already have an `ExternalCache` class, I'd hope this wouldn't be too hard. However, perhaps we need to consider the behaviour when Synapse/Redis are restarted at the same time - is it acceptable for clients to see pagination errors in such a case?

I somewhat had the same thought (that this would be better to put into an `ExternalCache`), but it doesn't seem to match what we do now for similar things (UI auth sessions?). In a way this isn't really *cache* data in that it can't be rebuilt. Which makes me lean towards putting it into the database?

richvdh:
> it doesn't seem to match what we do now for similar things (UI auth sessions?)

yeeeees, but it's worth noting that (if memory serves) UI auth session persistence was invented before we used redis for replication, and certainly before `ExternalCache` was created, so that's not necessarily a good guide to the future.

However:

> In a way this isn't really cache data in that it can't be rebuilt.

I suspect this is probably a deal breaker really.

If we're going down the path of using the database, it might still be better just to make it a table with abstract `key` and `value` columns (and possibly `expiry`), so that we don't need to create a new table and new storage methods each time we encounter something that fits this pattern?

(In an ideal world, maybe we'd have an abstract `SessionCache` which is currently backed by the database, but could also be backed by something else in future. But perhaps that's overengineering it.)

clokep:
> If we're going down the path of using the database, it might still be better just to make it a table with abstract `key` and `value` columns (and possibly `expiry`), so that we don't need to create a new table and new storage methods each time we encounter something that fits this pattern?

I did this, I think it is generic enough that it is usable! I'm pretty sure it could replace:

* `threepid_validation_*`
* `ui_auth_sessions`
* Potential solution for #7530 (and maybe a few other spots we keep state in memory).

I'm not sure if this is really an improvement though since you're now serializing arbitrary data to JSON. 🤷 Let me know what you think!

> (In an ideal world, maybe we'd have an abstract `SessionCache` which is currently backed by the database, but could also be backed by something else in future. But perhaps that's overengineering it.)

I think you could probably add a facade in front of what I have here to do this in the future? I still hesitate on calling it a cache though!

richvdh (commented):
I'm generally finding the concept of `segment` slippery to get my head around: it feels like a detail that the `SessionStore` shouldn't need to concern itself with. Looking at how it is used in the
`RoomSummaryHandler`, could it not just be part of the `value` of the session?

I also feel like what you are currently calling `key` should be a `session_type`.

clokep:
> I'm generally finding the concept of `segment` slippery to get my head around: it feels like a detail that the `SessionStore` shouldn't need to concern itself with. Looking at how it is used in the
> `RoomSummaryHandler`, could it not just be part of the `value` of the session?

It could be specific to the implementation needed by `RoomSummaryHandler` instead of generic. I'll take a look at that.

> I also feel like what you are currently calling `key` should be a `session_type`.

This is what I originally had it called, but changed it.

clokep:
I updated the code a bit, as requested, to remove the `segment` idea (I replaced it with explicit checks in the `RoomSummaryHandler`). I also renamed `key` -> `session_type`.

richvdh (approved):
lgtm!

(I slightly wonder if it would be better to use an `attr.s` class to hold the session data, converting to and from a dict when calling the storage layer. But we should probably just get on and land this rather than keep polishing it forever.)

clokep:
> (I slightly wonder if it would be better to use an `attr.s` class to hold the session data, converting to and from a dict when calling the storage layer. But we should probably just get on and land this rather than keep polishing it forever.)

I'm not sure what benefit this would offer? If it is meant to be generic I'm not really sure how the storage layer would know what class to turn it into?

I think I'll land this and we can improve it in the future if necessary.

richvdh:
> I'm not sure what benefit this would offer? If it is meant to be generic I'm not really sure how the storage layer would know what class to turn it into?

to be clear, I'd expect the `RoomSummaryHandler` to do the translation between attr.s and dicts. 

The benefit would essentially be to help with type-safety by collecting all the code that does the `dict` referencing into one place. Quite happy for this to land as it is though.

PR #10615, opened 2021-08-16
Refactor `on_receive_pdu` code

richvdh:
This should be a completely non-functional refactor. I've constructed it as a series of tiny commits, to help show the logic behind it.

~~Based on #10614~~


richvdh:
By the way, I'm aware this code is unwieldy and needs extracting to separate methods. I have a plan.

erikjohnston (approved):

Issue #10441, opened 2021-07-21
Remove the concept of worker "types"


anoadragon453:
We're currently midway through the process of fading out the concept of different types for a Synapse worker process. For example `synapse.app.pusher`, `synapse.app.federation_sender` etc. While initially this provided a quick way to configure a worker to perform certain actions, it is also rather confusing and the inability to specify multiple types per worker can limit the flexibility of deployment.

For instance, it's currently not possible to configure a worker to both send events to appservices (`synapse.app.appservice`) and update the user directory tables (`synapse.app.user_dir`), as the functionality is forcefully disabled unless the worker is specifically marked as each respective type:

https://github.com/matrix-org/synapse/blob/85d237eba789a667109ced140026d2494b210310/synapse/app/generic_worker.py#L433-L463

Instead, workers should be configured via worker file config options, so that one could simply set `notify_appservices: true` and `update_user_directory: true` in the worker config to enable the worker to handle both sets of tasks.

As many worker deployments currently rely on the `worker_type` configuration, a deprecation period is necessary. A rough plan for carrying this out could look like:

* Allow all current worker configurations through config file values alone.
* Announce a deprecation period for the `worker_type` config option.
* After some time, remove the `worker_type` config option. All workers are now a "generic worker", though this term is no longer a necessary distinction.

In the short term, the first step would allow for more immediate flexibility for deployments medium-sized deployments - which need to move tasks off the main process, but don't have the resources to spin up a separate worker for each type.

anoadragon453:
One small thinko for the first step was how we'd be able to keep the current functionality of warning the user that both the worker and main process was configured for the same purpose. This is necessary if you need a task to run on at max one worker process. We currently use a combination of the worker type (i.e `synapse.app.appservice`) and config options (`notify_appservices`), and we'd like to remove the former.

Well, we already have a solution to this: simply add a config option for specify the name of the worker that should handle the task. As an example, [`run_background_tasks_on: <worker_name>`](https://github.com/matrix-org/synapse/blob/a743bf46949e851c9a10d8e01a138659f3af2484/docs/sample_config.yaml#L2739-L2742).

richvdh:
`frontend_proxy` is already totally redundant: there is no difference between it and `generic_worker`. The only thing left is to remove it from the documentation, which I hope #13451 will do.

I also think `synapse.app.pusher` is superceded by `pusher_instances` (#9466), but again it needs documenting properly.

PR #10616, opened 2021-08-16
Replace the `synapse.app.appservice` and `synapse.app.user_dir` worker types with more flexible config options

anoadragon453:
First phase of https://github.com/matrix-org/synapse/issues/10441.

This PR replaces the logic that decides which worker to notify appservices or update the user directory. Originally, you would have to set `notify_appservices` / `update_user_directory` to `false`, as well as setting the `worker_type` config option for each worker.

Now we follow the same format as `run_background_tasks_on`, where we have a single config option which both the main process and worker processes read from. If the value matches the worker's current name, then that worker will carry out that task.

TODO:

- [ ] Update Complement worker config generator script
- [ ] Make a note in UPGRADE.rst - **this is currently a backwards-incompatible change**
- [ ] Update worker documentation

### DIFF
--- a/changelog.d/10616.feature
+++ b/changelog.d/10616.feature
@@ -1,0 +1,1 @@
+Allow a single worker to both notify appservices and update the user directory by removing the dependence on the `worker_type` worker config option.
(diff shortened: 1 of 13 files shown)


richvdh:
this looks very similar to #11450... is it still alive?

dklimpel:
Is this replaced by #12446 and #12654 ?

anoadragon453:
Oh! Yes, this is old, old, old and can be closed.

PR #10619, opened 2021-08-16
Improve the syntax of various utility shell scripts across the codebase

joao-vitor-sr:
Now the shell script avoid errors like word splitting, gobbling, and the Unicode was disabled to improve the performance

## Comments
I had created a pull request before, but I had some issues with git (conflict and branch issues), so I just created a new pull request.

### DIFF
--- a/changelog.d/10616.misc
+++ b/changelog.d/10616.misc
@@ -1,1 +1,1 @@
-Better syntax at syntax at the shell script.
+Better syntax at the shell script.
--- a/changelog.d/10616.misc
+++ b/changelog.d/10616.misc
@@ -1,1 +1,0 @@
-Better syntax at the shell script.
--- a/changelog.d/10616.misc
+++ b/changelog.d/10616.misc
@@ -1,0 +1,1 @@
+Better syntax at syntax at the shell script.
--- a/demo/start.sh
+++ b/demo/start.sh
@@ -1,9 +1,5 @@
 #!/usr/bin/env bash
 
-# Speed up script by not using unicode.
-LC_ALL=C
-LANG=C
-
 set -e
 
 DIR="$( cd "$( dirname "$0" )" && pwd )"
--- a/demo/start.sh
+++ b/demo/start.sh
@@ -1,5 +1,9 @@
 #!/usr/bin/env bash
 
+# Speed up script by not using unicode.
+LC_ALL=C
+LANG=C
+
 set -e
 
 DIR="$( cd "$( dirname "$0" )" && pwd )"
--- a/demo/start.sh
+++ b/demo/start.sh
@@ -80,7 +80,7 @@ for port in 8080 8081 8082; do
         } >> "$DIR"/etc/$port.config
 
         # Generate tls keys
-        openssl req -x509 -newkey rsa:4096 -keyout "$DIR"/etc/localhost"$(printf '\:')"$https_port.tls.key -out "$DIR"/etc/localhost"$(printf '\:')"$https_port.tls.crt -days 365 -nodes -subj "/O=matrix"
+        openssl req -x509 -newkey rsa:4096 -keyout $DIR/etc/localhost\:$https_port.tls.key -out $DIR/etc/localhost\:$https_port.tls.crt -days 365 -nodes -subj "/O=matrix"
 
         # Reduce the blacklist
         blacklist=$(cat <<-BLACK
--- a/demo/start.sh
+++ b/demo/start.sh
@@ -32,58 +32,54 @@ for port in 8080 8081 8082; do
 
     if ! grep -F "Customisation made by demo/start.sh" -q  "$DIR"/etc/$port.config; then
         {
-            printf '\n\n# Customisation made by demo/start.sh\n'
-            echo "public_baseurl: http://localhost:$port/"
-            echo 'enable_registration: true'
-        } >> "$DIR"/etc/$port.config
+					printf '\n\n# Customisation made by demo/start.sh\n'
+					echo "public_baseurl: http://localhost:$port/"
+					echo 'enable_registration: true'
 
-        # Warning, this heredoc depends on the interaction of tabs and spaces. Please don't
-        # accidentaly bork me with your fancy settings.
-		listeners=$(cat <<-PORTLISTENERS
-		# Configure server to listen on both $https_port and $port
-		# This overides some of the default settings above
-		listeners:
-		  - port: $https_port
-		    type: http
-		    tls: true
-		    resources:
-		      - names: [client, federation]
+		      # Warning, this heredoc depends on the interaction of tabs and spaces. Please don't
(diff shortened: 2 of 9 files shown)


changelog.d/10616.misc
```diff
@@ -0,0 +1 @@
+Better syntax at the shell script.
```
anoadragon453, line 1:
```suggestion
Improve the syntax of various utility shell scripts across the codebase.
```

may be a bit clearer.

anoadragon453 (commented):
One minor nit. At a glance this looks reasonable to me, but I'd like to get a look over from @richvdh as well.

demo/start.sh
```diff
@@ -1,5 +1,11 @@
 #!/usr/bin/env bash
 
+# Speed up script by not using unicode.
+LC_ALL=C
+LANG=C
```
richvdh, line 5: per my other review, let's not do this.

demo/start.sh
```diff
@@ -56,26 +63,28 @@ for port in 8080 8081 8082; do
...
-        printf '\n\n# Disable tls on the servers.' >> $DIR/etc/$port.config
-        echo '# DO NOT USE IN PRODUCTION' >> $DIR/etc/$port.config
-        echo 'use_insecure_ssl_client_just_for_testing_do_not_use: true' >> $DIR/etc/$port.config
-        echo 'federation_verify_certificates: false' >> $DIR/etc/$port.config
-
-        # Set tls paths
-        echo "tls_certificate_path: \"$DIR/etc/localhost:$https_port.tls.crt\"" >> $DIR/etc/$port.config
-        echo "tls_private_key_path: \"$DIR/etc/localhost:$https_port.tls.key\"" >> $DIR/etc/$port.config
+        {
+            echo "${listeners}"
```
richvdh, line 67: why not replace this with a simple `cat` ?

demo/start.sh
```diff
@@ -56,26 +63,28 @@ for port in 8080 8081 8082; do
...
 
         # Generate tls keys
-        openssl req -x509 -newkey rsa:4096 -keyout $DIR/etc/localhost\:$https_port.tls.key -out $DIR/etc/localhost\:$https_port.tls.crt -days 365 -nodes -subj "/O=matrix"
-
-        # Ignore keys from the trusted keys server
-        echo '# Ignore keys from the trusted keys server' >> $DIR/etc/$port.config
-        echo 'trusted_key_servers:' >> $DIR/etc/$port.config
-        echo '  - server_name: "matrix.org"' >> $DIR/etc/$port.config
-        echo '    accept_keys_insecurely: true' >> $DIR/etc/$port.config
+        openssl req -x509 -newkey rsa:4096 -keyout "$DIR"/etc/localhost"$(printf '\:')"$https_port.tls.key -out "$DIR"/etc/localhost"$(printf '\:')"$https_port.tls.crt -days 365 -nodes -subj "/O=matrix"
```
richvdh, line 87: per my other review: have you tested this? are you sure it is correct?

scripts-dev/next_github_number.sh
```diff
@@ -4,6 +4,6 @@ set -e
 
 # Fetch the current GitHub issue number, add one to it -- presto! The likely
 # next PR number.
-CURRENT_NUMBER=`curl -s "https://api.github.com/repos/matrix-org/synapse/issues?state=all&per_page=1" | jq -r ".[0].number"`
+CURRENT_NUMBER=$(curl -s "https://api.github.com/repos/matrix-org/synapse/issues?state=all&per_page=1" | jq -r ".[0].number")
```
richvdh, line 7: these are identical, I think?

test_postgresql.sh
```diff
@@ -3,10 +3,14 @@
...
+LANG=C
+
 set -e
 
 # Build, and tag
 docker build docker/ -f docker/Dockerfile-pgtests -t synapsepgtests
 
 # Run, mounting the current directory into /src
-docker run --rm -it -v $(pwd)\:/src synapsepgtests
+docker run --rm -it -v "$(pwd)$(printf '\:')"/src synapsepgtests
```
richvdh, line 16: this still looks wrong

richvdh (changes requested):
> I just created a new pull request.

Please *don't* do this. It makes it impossible to see what is different since the previous review, and means that the context from the previous PR is lost.

richvdh:
for context, the previous pr was #10604.

richvdh:
> joao-vitor-sr force-pushed

@joao-vitor-sr : please. I've asked you twice *not* to do this. It makes it much harder to review the PR. Please put it back to 54814ba.

joao-vitor-sr:
> > joao-vitor-sr force-pushed
> 
> @joao-vitor-sr : please. I've asked you twice _not_ to do this. It makes it much harder to review the PR. Please put it back to [54814ba](https://github.com/matrix-org/synapse/commit/54814ba0524f67f069cbcf41899b329fbc4c4e55).
I'm going to reverse, to reverse it will take one more push --force, I won't do that later. *sorry for the amateur mistake*

demo/start.sh
joao-vitor-sr: yes I've already tested it, and it's working
## References
https://github.com/koalaman/shellcheck/wiki/SC1001

richvdh: > yes I've already tested it, and it's working

Hrm. When I run the modified version of your script, the instances fail to start, with errors like this:

```
2021-08-18 17:10:14,801 - synapse.app._base - 208 - CRITICAL - sentinel - Error during startup
  File "/home/rav/work/synapse/synapse/config/_base.py", line 908, in read_file
    os.stat(file_path)
FileNotFoundError: [Errno 2] No such file or directory: '/home/rav/work/synapse/demo/etc/localhost:8480.tls.key'
```

demo/start.sh
joao-vitor-sr: the reason to use the group command instead of cat is that the 
cat is a binary written in c even with a great performance the performance and readability of the group command is better, in this situation the bash group all commands and redirect the result to the file, and the cat is a binary written for the manipulation of files.

## References
[https://mywiki.wooledge.org/BashFAQ/014](https://mywiki.wooledge.org/BashFAQ/014)
[https://www.man7.org/linux/man-pages/man1/cat.1.html](https://www.man7.org/linux/man-pages/man1/cat.1.html)

richvdh: I think you misunderstand me.

Currently, there is a `cat` command (`cat <<-PORTLISTENERS`) whose output is written to a bash variable, which is then `echo`ed again. I'm asking why not skip capturing the output, and run the `cat` from inside the command group so that its output goes straight to  the output file.

joao-vitor-sr: sorry for my mistake I was thinking of using the cat to redirect the output to the file

joao-vitor-sr:
so the pull request is taking a long time, I will cancel the pull request, I will study more.

PR #10621, opened 2021-08-17
Make a note to leave a summary when one is bumping the schema version

anoadragon453:
I found this easy to miss (and evidently, it looks like it was missed for schema version 62).


synapse/storage/schema/__init__.py
```diff
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# When updating this value, please leave a short summary of the changes below.
```
anoadragon453, line 15: Hmm, should we be asking the same for `SCHEMA_COMPAT_VERSION` perhaps?

richvdh: I wouldn't really think so. it's somewhat obvious?

richvdh (approved):

synapse/storage/schema/__init__.py
anoadragon453: The same could be said about this value :stuck_out_tongue:

I might just tweak the wording slightly to apply to both values.

Issue #10618, opened 2021-08-16
`device_unused_fallback_key_types` doesn't appear in `/sync` until after a fallback key is uploaded


turt2live:
### Description

https://github.com/matrix-org/matrix-doc/pull/2732 states that a client can detect server support by looking for the `device_unused_fallback_key_types` field in `/sync`, however this field is only populated by Synapse once a fallback key is uploaded. This can leave clients looping a fallback key upload as they'll be forced to assume the fallback key support is there and that all their keys have been used, thus upload a new key, which leads to the server (possibly) not supporting it and ignoring the unknown fields, so the client syncs, doesn't see the field, uploads, repeat.

Further, clients like Element Web won't upload fallback keys for users if the server doesn't advertise this field on `/sync`, thus there are likely users out there missing the value of fallback keys.

### Steps to reproduce

- Be on a fresh account
- Look for `device_unused_fallback_key_types` in the `/sync` response

<!--
Describe how what happens differs from what you expected.

If you can identify any relevant log snippets from _homeserver.log_, please include
those (please be careful to remove any personal or private data). Please surround them with
``` (three backticks, on a line on their own), so that they are formatted legibly.
-->

### Version information

<!-- IMPORTANT: please answer the following questions, to help us narrow down the problem -->

<!-- Was this issue identified on matrix.org or another homeserver? -->
- **Homeserver**: localhost

If not matrix.org:

<!--
 What version of Synapse is running?

You can find the Synapse version with this command:

$ curl http://localhost:8008/_synapse/admin/v1/server_version

(You may need to replace `localhost:8008` if Synapse is not configured to
listen on that port.)
-->
- **Version**: 1.40.0

- **Install method**:
<!-- examples: package manager/git clone/pip  -->

- **Platform**: Windows / Docker
<!--
Tell us about the environment in which your homeserver is operating
distro, hardware, if it's running in a vm/container, etc.
-->

anoadragon453:
The MSC states:

> The `device_unused_fallback_key_types` parameter must be present if the server supports fallback keys. Clients can thus treat this field as an indication that the server supports fallback keys, and so only upload fallback keys to servers that support them.

Apologies for the breakage here. The relevant code is:

https://github.com/matrix-org/synapse/blob/683deee9a4be8400e72bb2de87b82912656597ac/synapse/rest/client/v2_alpha/sync.py#L262

And looks like the fix is just to remove the conditional.

PR #10623, opened 2021-08-17
Always list fallback key types in /sync

anoadragon453:
Fixes https://github.com/matrix-org/synapse/issues/10618

https://github.com/matrix-org/synapse/pull/9919 removed optional keys from `/sync`, though it was missed that `org.matrix.msc2732.device_unused_fallback_key_types` should always be included down `/sync` in accordance with [MSC2732](https://github.com/matrix-org/matrix-doc/blob/54255851f642f84a4f1aaf7bc063eebe3d76752b/proposals/2732-olm-fallback-keys.md).

This PR reintroduces the behaviour of always including the key.


richvdh (approved):
lgtm

Issue #10626, opened 2021-08-17
Error on startup: AttributeError: ('max_upload_size', "not found in [...]")


anoadragon453:
We received the following traceback on the main process when attempting to deploy v1.41.0rc1 (18572fddcc5bbcecabdcfbc69c8e8d4a1e9a1a1b) to matrix.org:

```
2021-08-17 13:47:11,357 - synapse.federation.federation_server - 1204 - INFO - sentinel - Registering federation EDU handler for 'org.matrix.signing_key_update'
2021-08-17 13:47:11,357 - synapse.federation.federation_server - 1224 - INFO - sentinel - Registering federation query handler for 'client_keys'
2021-08-17 13:47:11,368 - synapse.app._base - 208 - CRITICAL - sentinel - Error during startup
Capture point (most recent call last):
  File "/usr/local/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/local/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/synapse/src/synapse/app/homeserver.py", line 462, in <module>
    main()
  File "/home/synapse/src/synapse/app/homeserver.py", line 458, in main
    run(hs)
  File "/home/synapse/src/synapse/app/homeserver.py", line 444, in run
    logger=logger,
  File "/home/synapse/src/synapse/app/_base.py", line 144, in start_reactor
    run()
  File "/home/synapse/src/synapse/app/_base.py", line 128, in run
    run_command()
  File "/home/synapse/env-py37/lib/python3.7/site-packages/twisted/internet/base.py", line 1318, in run
    self.mainLoop()
  File "/home/synapse/env-py37/lib/python3.7/site-packages/twisted/internet/base.py", line 1331, in mainLoop
    reactorBaseSelf.doIteration(t)
  File "/home/synapse/env-py37/lib/python3.7/site-packages/twisted/internet/epollreactor.py", line 244, in doPoll
    log.callWithLogger(selectable, _drdw, selectable, fd, event)
  File "/home/synapse/env-py37/lib/python3.7/site-packages/twisted/python/log.py", line 101, in callWithLogger
    return callWithContext({"system": lp}, func, *args, **kw)
  File "/home/synapse/env-py37/lib/python3.7/site-packages/twisted/python/log.py", line 85, in callWithContext
    return context.call({ILogContext: newCtx}, func, *args, **kw)
[...]

PR #10558, opened 2021-08-09
Admin API to delete media for a specific user

dklimpel:
Fixes: #9565

API to delete more than one media for a specific user
`DELETE /_synapse/admin/v1/users/<user_id>/media`

### Pull Request Checklist

* [x] Pull request is based on the develop branch
* [x] Pull request includes a [changelog file](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md#changelog). The entry should:
  - Be a short description of your change which makes sense to users. "Fixed a bug that prevented receiving messages from other servers." instead of "Moved X method from `EventStore` to `EventWorkerStore`.".
  - Use markdown where necessary, mostly for `code blocks`.
  - End with either a period (.) or an exclamation mark (!).
  - Start with a capital letter.
* [x] Pull request includes a [sign off](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md#sign-off)
* [x] Code style is correct (run the [linters](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md#code-style))

Signed-off-by: Dirk Klimpel dirk@klimpel.org

PR #10628, opened 2021-08-17
Mount /_synapse/admin/v1/users/{userId}/media admin API on media workers only

anoadragon453:
Fixes https://github.com/matrix-org/synapse/issues/10626

https://github.com/matrix-org/synapse/pull/10558 makes use of `synapse.rest.media.v1.MediaRepository`, however this should only be done on servlets registered by `synapse.rest.admin.media.register_servlets_for_media_repo`.

This PR moves the servlet over and updates the appropriate documentation.


synapse/rest/admin/media.py
```diff
@@ -328,3 +488,4 @@ def register_servlets_for_media_repo(hs: "HomeServer", http_server: HttpServer)
     ListMediaInRoom(hs).register(http_server)
     DeleteMediaByID(hs).register(http_server)
     DeleteMediaByDateSize(hs).register(http_server)
+    UserMediaRestServlet(hs).register(http_server)
```
clokep, line 491: Weirdly I think this means this will be available when a `client` listener is configured since `register_servlets_for_media_repo` is called via `register_servlets_for_client_rest_resource`.

I don't think that's a huge deal, although we should probably clean all that up at some point...

docs/workers.md
```diff
@@ -426,6 +426,7 @@ Handles the media repository. It can handle all endpoints starting with:
     ^/_synapse/admin/v1/user/.*/media.*$
     ^/_synapse/admin/v1/media/.*$
     ^/_synapse/admin/v1/quarantine_media/.*$
+    ^/_synapse/admin/v1/users/.*/media$
```
clokep, line 429: (I actually wanted to comment on line 418...)

I think this change is saying that this endpoint *must* be handled by the media repo? "It *can* handle all endpoints start with..." makes it sound like an optional thing.

clokep:
Overall I think this looks fine!

docs/workers.md
anoadragon453: Good spot! Well, the main process can also handle it; but not when media workers are in use, as media repository functionality is typically disabled on the main process in that case.

I've added a line to help clarify things.

clokep (approved):

PR #10629, opened 2021-08-17
Convert room members storage tuples to attrs.

clokep:
This converts the `namedtuples` to `attrs` classes for the storage classes used by the `RoomMemberStore`.

Each is converted as a separate commit.

The reasoning for this is:

* Better type hints
* Will be expanding one of these in the future

### DIFF
--- a/synapse/storage/roommember.py
+++ b/synapse/storage/roommember.py
@@ -14,8 +14,7 @@
 # limitations under the License.
 
 import logging
-from collections import namedtuple
-from typing import Optional
+from typing import List, Optional, Tuple
 
 import attr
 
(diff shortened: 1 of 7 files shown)


synapse/storage/roommember.py
```diff
@@ -16,16 +16,22 @@
...
 
 RoomsForUser = namedtuple(
     "RoomsForUser", ("room_id", "sender", "membership", "event_id", "stream_ordering")
 )
 
-GetRoomsForUserWithStreamOrdering = namedtuple(
-    "GetRoomsForUserWithStreamOrdering", ("room_id", "event_pos")
-)
+
+@attr.s(slots=True, frozen=True, auto_attribs=True)
```
erikjohnston, line 35: Could we add `weakref_slot=False` for things that are going to be cached a lot? That cuts out quite a few bytes.

erikjohnston (approved):

synapse/storage/roommember.py
clokep: I put it on all of them since they're all cached. Hopefully that seems OK?

erikjohnston: Sounds good, there's no downside to it since we don't use weakrefs

DMRobertson: I think e97ac25282d763c438221d462ece249f0e21a757 added `weakref_slot=True` rather than `=False` --- don't follow that given the above

clokep: Oh, you're right. I messed that up. 🤦 Thanks for catching it!

clokep: See #10642

erikjohnston: I'm glad someone is on the ball today :slightly_smiling_face:

PR #10630, opened 2021-08-17
Use auto-attribs for attrs classes for sync.

clokep:
Converts the `attrs` classes used by the `SyncHandler` to use `auto_attribs`.

### DIFF
--- a/changelog.d/10630.misc
+++ b/changelog.d/10630.misc
@@ -1,0 +1,1 @@
+Use auto-attribs for the attrs classes used in sync.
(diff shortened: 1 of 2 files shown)


DMRobertson (approved):
Didn't realise attrs supported this dataclasses-like syntax---neat!

clokep:
> Didn't realise attrs supported this dataclasses-like syntax---neat!

We've only started switching to it recently, but I was going to touch this code anyway so figured I should update. Thanks for the review!

PR #10638, opened 2021-08-18
Allow modules to run looping call on all instances

### LINKED ISSUE #10635: Module API does not permit running looping jobs on worker processes
`ModuleApi.looping_background_call` does nothing on worker processes, which is problematic if your module needs to run any sort of background job.

erikjohnston:
By default the calls only ran on the worker configured to run background tasks.

Fixes #10635

### DIFF
--- a/synapse/module_api/__init__.py
+++ b/synapse/module_api/__init__.py
@@ -623,7 +623,7 @@ class ModuleApi:
             msec: How long to wait between calls in milliseconds.
             *args: Positional arguments to pass to function.
             desc: The background task's description. Default to the function's name.
-            run_on_all_workers: Whether to run this on all instances, rather
+            run_on_all_instances: Whether to run this on all instances, rather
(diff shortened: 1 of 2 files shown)


synapse/module_api/__init__.py
```diff
@@ -618,12 +623,14 @@ def looping_background_call(
             msec: How long to wait between calls in milliseconds.
             *args: Positional arguments to pass to function.
             desc: The background task's description. Default to the function's name.
+            run_on_all_workers: Whether to run this on all instances, rather
```
babolivier, line 626:
```suggestion
            run_on_all_instances: Whether to run this on all instances, rather
```

babolivier (approved):
lgtm otherwise

synapse/module_api/__init__.py
erikjohnston: Stupid replace all not working grumble

PR #10639, opened 2021-08-18
Fix the titles in the OIDC documentation

anoadragon453:
Having them as links broke the table-of-contents rendering in mdbook:

image

Plus there's not really much reason for only some of the provider titles to be links.

### DIFF
--- a/docs/openid.md
+++ b/docs/openid.md
@@ -246,9 +246,11 @@ oidc_providers:
 
 ### Google
 
+[Google][google-idp] is an OpenID certified authentication and authorisation provider.
+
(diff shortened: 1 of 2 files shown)


docs/openid.md
```diff
@@ -242,7 +244,7 @@ oidc_providers:
         display_name_template: "{{ user.name }}"
 ```
 
-### [Google][google-idp]
+### Google
```
richvdh, line 247: add a link in the text, maybe?

richvdh (approved):
lgtm otherwise

PR #10644, opened 2021-08-18
Do not include rooms with an unknown room version in a sync response.

### LINKED ISSUE #7083: Rooms with unsupported versions are shown as blank rooms

### LINKED ISSUE #10525: Sync failing with 404 M_NOT_FOUND when the user is invited to a room with an unknown version
### Description

As per the title; tried logging into my matrix.org test account and /sync responded with a 404 over and over again

image

### Version information

- **Homeserver**: matrix.org

I believe the account was `@webdevguru_test17:matrix.org` in case that helps with log spelunking

It might be because the user was in an msc3083 unstable room the server dropped support for. 🔥

clokep:
This simply leaves the rooms out of the `/sync` response. The user needs to clear cache & reload (for Element Web at least) in order to get rid of the rooms locally.

I debated about trying to fake a leave or something, but that seemed more complicated than worth it.

Note that for incremental syncs the room is already ignored (since the events cannot be created due to having unknown room versions). The fix is really just for full syncs where the room is now completely ignored.

I tested this manually with the following:

1. Created three version 8 rooms with alice.
2. Joined one room with bob.
3. Invited bob to one room from alice.
4. Updated the join rules of the last room to `knock` as alice, then had bob knock on the room.
5. Shutdown the server
6. Comment out `RoomVersions.V8` from `synapse.api.room_versions`.
7. Restart the server.
8. Clear cache & reload
9. Check that the rooms aren't there and no errors in the logs.

Fixes #7083 and fixes #10525

### DIFF
--- a/changelog.d/10644.bugfix
+++ b/changelog.d/10644.bugfix
@@ -1,0 +1,1 @@
+Rooms with unsupported room versions are no longer returned via `/sync`.
--- a/mypy.ini
+++ b/mypy.ini
@@ -87,6 +87,7 @@ files =
   tests/test_utils,
   tests/handlers/test_password_providers.py,
   tests/handlers/test_room_summary.py,
+  tests/handlers/test_sync.py,
   tests/rest/client/v1/test_login.py,
   tests/rest/client/v2_alpha/test_auth.py,
   tests/util/test_itertools.py,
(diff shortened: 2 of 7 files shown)


clokep:
It occurs to me that maybe I should have looked into initial sync too? I'm unsure how deprecated that is and if we should think about that at all.

erikjohnston (approved):

erikjohnston:
> It occurs to me that maybe I should have looked into initial sync too? I'm unsure how deprecated that is and if we should think about that at all.

That should only be used for peeking now, so I don't think its worth it

PR #10645, opened 2021-08-18
Make `backfill` and `get_missing_events` use the same codepath

### LINKED ISSUE #3124: We accept stupid events over backfill, causing state res storms
If somebody creates an event with hundreds of prev_events, we will (since #3118) reject it when it arrives over federation. However, we may then try to backfill it, and will accept it and do an awful state resolution.

Fixing this is tricky as simply rejecting the event would break back-pagination for the room in question, in the same way that it is currently broken by events which don't have the correct signature (#3121). A solution to #3121 would probably mean that we could also fix this.

richvdh:
Given that `backfill` and `get_missing_events` are basically the same thing, it's somewhat crazy that we have entirely separate code paths for them. This makes `backfill` use the existing `get_missing_events` code, and then clears up all the unused code.

Requires https://github.com/matrix-org/sytest/pull/1121.

Fixes #3124.

Based on #10640.


richvdh:
CI is currently failing pending matrix-org/sytest#1121.

PR #10648, opened 2021-08-18
Note that /spaces & /hierarchy can be routed to workers

clokep:
These are read-only APIs and can be routed to workers.

This also cleans-up some of the registering of rest servlets for the worker code. Somewhat unrelated, I can remove it if we would like.

Part of #10523


clokep:
I'm not 100% sure if we should actually merge this since we have no tests for worker mode (see matrix-org/complement#194).

docs/workers.md
```diff
@@ -256,6 +259,7 @@ ensure that the purge history admin API is not used while pagination requests
 for the room are in flight:
 
     ^/_matrix/client/(api/v1|r0|unstable)/rooms/.*/messages$
+    ^/_matrix/client/unstable/org.matrix.msc2946/rooms/.*/hierarchy$
```
clokep, line 262: I think ensuring all requests go to the same worker would be fixed by #10613, but it is unclear if that's the proper approach.

clokep:
I think we'll want to wait for #10613 before doing this.

clokep:
This has been running on matrix.org for a couple of days without issue.

DMRobertson (commented):
Am I right in thinking that the changes to generic_worker.py are just tidy-ups, i.e. not a functional change?

docs/upgrade.md
```diff
@@ -85,6 +85,15 @@ process, for example:
...
     ```
 
+# Upgrading to v1.43.0
+
+## The spaces summary APIs can now be handled by workers
+
+The [available worker applications documentation](https://matrix-org.github.io/synapse/latest/workers.html#available-worker-applications)
+has been updated to reflect that calls to the `/spaces`, `/hierarchy`, and
+`/summary` endpoints can now be routed to workers for both client API and
+federation requests.
```
DMRobertson, line 95: What's the intented action for server owners here---copy and and paste the new list of regexes into their config if they want to use workers?

clokep: The intended action is to go look at the worker document and potentially update their config, yes.

We seem to have sometimes included upgrade notes about this in the past (we did for moving `createRoom`), but don't always seem to. It's possible this document is really only meant for things which *must* be done.

clokep:
> Am I right in thinking that the changes to generic_worker.py are just tidy-ups, i.e. not a functional change?

Yep, sorry if the description wasn't clear:

> This also cleans-up some of the registering of rest servlets for the worker code. Somewhat unrelated, I can remove it if we would like.

I initially thought I was going to have to register the endpoint so did some clean-up there...then realized it was actually registered already. 🤦

DMRobertson:
> Yep, sorry if the description wasn't clear

Ahh yes, right you are. I missed the description completely. :facepalm:

DMRobertson (approved):

docs/upgrade.md
DMRobertson: Sounds good! Just wanted to check the thinking.

Issue #10340, opened 2021-07-08
Run CI against Twisted's "trunk" branch


callahad:
We're a big / complex enough project to where testing against trunk would be genuinely valuable for the broader Twisted ecosystem.

This would likely take the form of a scheduled GitHub Action which runs our test suite once a day, using Twisted's `trunk` branch instead of whatever version is specified in our dependencies.

ShadowJonathan:
How would the signalling of a failed job for this go? A matrix room?

callahad:
I believe GitHub Actions sends repo owners a notification when a scheduled job fails, similar to how it notifies you when your own PR fails.

richvdh:
> We're a big / complex enough project to where testing against trunk would be genuinely valuable for the broader Twisted ecosystem.

This will also bring a more direct benefit to the Synapse project: it will give us early warning of any compatibility problems in the next version of Twisted.

richvdh:
I've been wondering how we could best approach this. I'm assuming we want to run at least `{mypy, trial, sytest, complement}` on Synapse with `pip install https://github.com/twisted/twisted/archive/trunk.zip`.

Of those, I think the hardest is going to be sytest, because the incantations for setting up and running Synapse are all embedded within the [sytest-synapse](https://hub.docker.com/r/matrixdotorg/sytest-synapse) docker image.

Possibly we could do some `sed` on `python-dependencies` before we start the test runs, to replace the Twisted dependency with trunk instead of 18.whatever?

PR #10651, opened 2021-08-18
Run a nightly CI build against Twisted trunk.

richvdh:
This creates a GHA workflow which runs at 8am every day, and runs `mypy`, `trial` and `sytest` against Twisted's current trunk. If any of the jobs fail, it opens an issue.

See https://github.com/matrix-org/synapse/actions/runs/1144688011 for an example of a successful run, and https://github.com/matrix-org/synapse/issues/10650 for what happens when the build fails.

Fixes: #10340.


erikjohnston (approved):
Sure?

richvdh:
I'll merge this next week. Thanks for the review!

Issue #9236, opened 2021-01-27
Display name limit doesn't apply to per-room display names


babolivier:
As per [discussion](https://matrix.to/#/!OGEhHVWSdvArJzumhm:matrix.org/$iPQbTIcztGWabBc-pibIGP4E-6tFtN4Zcmk-dRYsQUI?via=matrix.org&via=privacytools.io&via=dendrite.matrix.org) in Matrix HQ, it seems like [Synapse's limit to display name length](https://github.com/matrix-org/synapse/blob/7a2e9b549defe3f55531711a863183a33e7af83c/synapse/handlers/profile.py#L43) doesn't apply to per-room display names.

anoadragon453:
This would just be adding a length check to `update_membership_locked` that would fail the request if the length was too long. Similar to where we prevent per-room profile changes entirely:

https://github.com/matrix-org/synapse/blob/6633a4015a7b4ba60f87c5e6f979a9c9d8f9d8fe/synapse/handlers/room_member.py#L358-L365

b-thebest:
Can I work on  this issue ??

anoadragon453:
@b-thebest Of course!

b-thebest:
Please assign this task to me

arya2331:
is someone still working on it? or I can do this?

b-thebest:
I am working on it

richvdh:
@b-thebest do you have any idea of when you expect to make some progress on it? I don't think it's a big job and if others are ready to pick it up it might be nice to let them do so.

b-thebest:
It was not in list of task assigned to me, that's why I was unable to do
it. I will do it now asap

anoadragon453:
Hey @b-thebest, have you been able to make any progress on this?

b-thebest:
Yes I was working on it. Sorry for being late, during past days my health conditions were not good that's why I was unable to complete it. From today I will take it on and try to complete it as soon as possible

richvdh:
As is normal in open-source projects, we don't generally "assign" issues to individual contributors, and you shouldn't consider this a prerequisite to start work on it.

lukasdenk:
@Azrenbeth Should this issue still be open? According to #10654 the issue is solved.

PR #10654, opened 2021-08-19
Apply limit to per-room display names when setting them

Azrenbeth:
Apply limit to per-room display names when setting them

Fixes issue #9236

### DIFF
--- a/changelog.d/10654.bugfix
+++ b/changelog.d/10654.bugfix
@@ -1,0 +1,1 @@
+Apply display name length limit to per-room display names.
--- a/changelog.d/10654.bugfix
+++ b/changelog.d/10654.bugfix
@@ -1,1 +1,1 @@
-Apply display name length limit to per-room display names.
+Enforce the maximum length for per-room display names and avatar URLs.
--- a/synapse/handlers/room_member.py
+++ b/synapse/handlers/room_member.py
@@ -49,6 +49,7 @@ from synapse.util.async_helpers import Linearizer
 from synapse.util.distributor import user_left_room
 
 from ._base import BaseHandler
+from .profile import MAX_DISPLAYNAME_LEN
 
 if TYPE_CHECKING:
     from synapse.server import HomeServer
@@ -548,8 +549,10 @@ class RoomMemberHandler(metaclass=abc.ABCMeta):
         )
 
         if (
-            not self.allow_per_room_profiles and not is_requester_server_notices_user
-        ) or requester.shadow_banned:
+            (not self.allow_per_room_profiles and not is_requester_server_notices_user)
+            or requester.shadow_banned
+            or len(content.get("displayname", "")) > MAX_DISPLAYNAME_LEN
+        ):
             # Strip profile data, knowing that new profile data will be added to the
             # event's content in event_creation_handler.create_event() using the target's
             # global profile.
--- a/synapse/handlers/room_member.py
+++ b/synapse/handlers/room_member.py
@@ -80,7 +80,7 @@ class RoomMemberHandler(metaclass=abc.ABCMeta):
         self.account_data_handler = hs.get_account_data_handler()
         self.event_auth_handler = hs.get_event_auth_handler()
 
-        self.member_linearizer = Linearizer(name="member")
+        self.member_linearizer: Linearizer = Linearizer(name="member")
 
         self.clock = hs.get_clock()
         self.spam_checker = hs.get_spam_checker()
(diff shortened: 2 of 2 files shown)


synapse/handlers/room_member.py
```diff
@@ -49,6 +49,7 @@
 from synapse.util.distributor import user_left_room
 
 from ._base import BaseHandler
+from .profile import MAX_DISPLAYNAME_LEN
```
clokep, line 52: I think we generally prefer absolute imports now.

synapse/handlers/room_member.py
```diff
@@ -79,7 +80,7 @@ def __init__(self, hs: "HomeServer"):
         self.account_data_handler = hs.get_account_data_handler()
         self.event_auth_handler = hs.get_event_auth_handler()
 
-        self.member_linearizer = Linearizer(name="member")
+        self.member_linearizer: Linearizer = Linearizer(name="member")
```
clokep, line 83: This type hint shouldn't be necessary since it is declared as a `Linearizer`. 😕

synapse/handlers/room_member.py
```diff
@@ -548,13 +549,17 @@ async def update_membership_locked(
         )
 
         if (
-            not self.allow_per_room_profiles and not is_requester_server_notices_user
-        ) or requester.shadow_banned:
+            (not self.allow_per_room_profiles and not is_requester_server_notices_user)
+            or requester.shadow_banned
+            or len(content.get("displayname", "")) > MAX_DISPLAYNAME_LEN
```
clokep, line 554: We should probably check the avatar URL while we're here.

I also don't think this check is correct -- if the display name is too long, this shouldn't mean that we also strip off the avatar URL! (And vice versa.)

@anoadragon453 asserted in #9236 that we should "fail the request". This is silently stripping it, which I'm unsure is the right thing to do.

synapse/handlers/room_member.py
```diff
@@ -548,13 +549,17 @@ async def update_membership_locked(
...
+        ):
             # Strip profile data, knowing that new profile data will be added to the
             # event's content in event_creation_handler.create_event() using the target's
             # global profile.
             content.pop("displayname", None)
             content.pop("avatar_url", None)
+            # content may now be emtpy
+            content_specified = len(content) == 0
```
clokep, line 562: Why is this change necessary?

clokep (changes requested):
I think this is on the right track, but I'm not sure about a couple of the changes.

synapse/handlers/room_member.py
Azrenbeth: Not having that gave mypy errors about not being able to determine the type of `member_linearizer`. I wasn't able to track down what was making mypy unhappy so thought adding in the type hint was a good enough solution

synapse/handlers/room_member.py
Azrenbeth: Have added the avatar URL length check

Was unsure if it was safe to "fail the request" - would that essentially mean having synapse prevent the event from ever having happened? What would happen if other servers did accept it?

synapse/handlers/room_member.py
Azrenbeth: It wasn't necessary, I'd just added it as I thought it was weird that content_specified might be incorrect. I've deleted it though as it wasn't really relevant to this

changelog.d/10654.bugfix
```diff
@@ -0,0 +1 @@
+Apply display name length limit to per-room display names.
```
clokep, line 1:
```suggestion
Enforce the maximum length for per-room display names and avatar URLs.
```

synapse/handlers/room_member.py
```diff
@@ -556,6 +557,12 @@ async def update_membership_locked(
             content.pop("displayname", None)
             content.pop("avatar_url", None)
 
+        if len(content.get("displayname") or "") > MAX_DISPLAYNAME_LEN:
+            raise SynapseError(403, "Displayname is too long (max 256)")
```
clokep, line 561: For each of these, please use a `400` error with a code of `Codes.BAD_JSON`.

synapse/handlers/room_member.py
clokep: I *think* there's something weird going on here because it is an abstract class. Seems like a reasonable enough solution.

clokep (approved):

clokep:
@Azrenbeth Looks good! Any idea if it would easy to add tests for this?

Azrenbeth:
> @Azrenbeth Looks good! Any idea if it would easy to add tests for this?

I don't really know how I could do that sorry!

clokep:
> > @Azrenbeth Looks good! Any idea if it would easy to add tests for this?
> 
> I don't really know how I could do that sorry!

Taking a look I don't see a straightforward way to do this unfortunately. 😢

PR #8051, opened 2020-08-07
Auto set logging filter

erikjohnston:
Instead of requiring people to explicitly set the filter correctly in the config, instead let's do it automatically for them. It's too easy at the moment to mess it up by e.g. setting the filter on the wrong handler when using buffering and get the wrong contexts logged.

This is backwards compatible, manually setting the filter will just result in the context being clobbered. This does mean that if you set the filter in the wrong place you'll get incorrect contexts still.

Based on https://github.com/matrix-org/synapse/pull/8040

PR #10657, opened 2021-08-19
Log exceptions in tests to stderr

DMRobertson:
This makes it easy to see errors in background processes when you're
writing unit tests. Otherwise, you have to know/remember to look at the
test logs in `_trial_temp`.

### DIFF
--- a/tests/test_utils/logging_setup.py
+++ b/tests/test_utils/logging_setup.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 import logging
 import os
+import sys
 
 import twisted.logger
 
@@ -35,7 +36,8 @@ class ToTwistedHandler(logging.Handler):
 def setup_logging():
     """Configure the python logging appropriately for the tests.
 
-    (Logs will end up in _trial_temp.)
+    Logs will end up in _trial_temp. Exceptions are additionally
+    logged to stderr.
     """
     root_logger = logging.getLogger()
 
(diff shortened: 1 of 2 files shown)


tests/test_utils/logging_setup.py
```diff
@@ -35,7 +36,8 @@ def emit(self, record):
 def setup_logging():
     """Configure the python logging appropriately for the tests.
 
-    (Logs will end up in _trial_temp.)
+    Logs will end up in _trial_temp. Exceptions are additionally
```
richvdh, line 39: Not everything logged at `logger.error` is an exception! `s/exceptions/errors/`.

Likewise in the changelog, and the name of the handler.

tests/test_utils/logging_setup.py
```diff
@@ -44,11 +46,16 @@ def setup_logging():
...
     )
 
-    handler = ToTwistedHandler()
+    to_twisted_handler = ToTwistedHandler()
     formatter = logging.Formatter(log_format)
-    handler.setFormatter(formatter)
-    handler.addFilter(LoggingContextFilter())
-    root_logger.addHandler(handler)
+    to_twisted_handler.setFormatter(formatter)
+    to_twisted_handler.addFilter(LoggingContextFilter())
```
richvdh, line 52: I *think* this is redundant, as of #8051. (If it's not redundant, we need to do it for the other handler too.)

richvdh (changes requested):

tests/test_utils/logging_setup.py
DMRobertson: Alright. I propose to test if this is redundant by
* removing filter here
* changing LoggingContextFilter to have some obvious side effect
* trigger the new log-to-stderr behaviour, and look for the side effect

richvdh:
I'm a bit concerned that this makes the test output hard to parse - there's actually quite a lot of stuff logged at ERROR which is expected in the tests, and this makes it look like things are failing when they're not.

Might be worth getting opinions from the rest of the team, in the room?

richvdh:
(on the other hand, possibly a bunch of those things logged at ERROR shouldn't be...)

tests/test_utils/logging_setup.py
DMRobertson: Appears to be needed. Otherwise the call to `debug` in synapse/metrics/__init__.py:596 will fail because the record doesn't have a "request" field.

DMRobertson:
Don't think anyone was super keen on this, so I'll close this down.

Maybe something simpler would suffice. (If a test fails, print out "check homeserver.log in _trial_temp" in bright yellow)

DMRobertson:
Argh. Just lost about an hour to this again.

Can we do better? I'm guessing somewhere high up we catch any remaining uncaught exceptions and `logger.exception` them. Can we stop that during test runs so that they bubble up immediately?

In the past I've used [pytest.raises](https://docs.pytest.org/en/latest/reference/reference.html#pytest.raises) to handle tests that are expected to raise exceptions.

richvdh:
uncaught exceptions will cause a test to fail.

richvdh:
(I think the problem is that the exceptions being logged *are* caught)

It would be instructive to see the sorts of errors that you are seeing being logged that would have been useful to make a test fail. Then we can see where it is being caught, and whether it would make sense to stop catching it during tests.

PR #10658, opened 2021-08-19
 Include room avatars in email notifications 

### LINKED ISSUE #1546: Room avatars are broken on email notifications
Submitted by @&#8203;matthew:matrix.org

(Imported from https://matrix.org/jira/browse/SYN-695)

### Relevant code

 - [`synapse/res/templates/room.html#L4-L14`](https://github.com/matrix-org/synapse/blob/430241a1e9fd4fbb82d83958b61bbd66c9ba3505/synapse/res/templates/room.html#L4-L14)

DMRobertson:
Judging by the template, this was intended ages ago, but we never
actually passed an avatar URL to the template.

I'm not sure what the best way to test this is. I've written a unit test
of sorts, but I'd like to manually verify it by actually seeing an
email. Maybe I should setup my own installation of synapse?

Closes #1546.


synapse/push/mailer.py
```diff
@@ -399,6 +394,16 @@ async def _get_room_vars(
...
         return room_vars
 
+    async def _get_room_avatar(
+        self,
+        room_state_ids: StateMap[str],
+    ) -> Optional[str]:
+        event_id = room_state_ids.get((EventTypes.RoomAvatar, ""))
+        if event_id:
+            ev = await self.store.get_event(event_id)
+            return ev.content["url"]
```
clokep, line 404: This is untrusted, so there's no guarantee that this will have a `url` (or that it will be a string).

I also wonder if it makes sense to use the `thumbnail_url` instead?

Regardless, we'll need to convert the MXC URL to an HTTP one, I believe.

DMRobertson: > Regardless, we'll need to convert the MXC URL to an HTTP one, I believe.

I think this is already done in the jinja template room.html:
```jinja2
                <img alt="" src="{{ room.avatar_url|mxc_to_http(48,48) }}" />
```
where I'm guessing that pipe will feed the LHS into the `mxc_to_http_filter` defined in synapse/util/templates.py:85

DMRobertson: > This is untrusted, so there's no guarantee that this will have a url (or that it will be a string).

The spec says that "url" is required---but I guess we don't enforce that at present? Can do an `isintance(..., str)` check on the URL though.

> I also wonder if it makes sense to use the thumbnail_url instead?

Spec says that this is "Only present if the thumbnail is unencrypted.". But I'm not sure if the main avatar could be encrypted too?

reivilibre:
If you want to test it, you could probably test it with a local test homeserver, perhaps using https://mailtrap.io/ as your SMTP server (or a real one if you have access to one that you don't mind using — but I think I used mailtrap a long time ago for this kind of thing)

synapse/push/mailer.py
clokep: > > This is untrusted, so there's no guarantee that this will have a url (or that it will be a string).
> 
> The spec says that "url" is required---but I guess we don't enforce that at present? Can do an `isintance(..., str)` check on the URL though.

In general, event content can be whatever. So I don't think we can assume it is there / the right type. Even if we enforce it, an event could come over federation that has the incorrect shape. (Or it could have been redacted so all event content was removed.)

DMRobertson:
> If you want to test it, you could probably test it with a local test homeserver, perhaps using https://mailtrap.io/ as your SMTP server (or a real one if you have access to one that you don't mind using — but I think I used mailtrap a long time ago for this kind of thing — basically it's a test SMTP server that captures all outgoing e-mail to your account for you to peruse in the web interface instead of sending them to someone).

Mailtrap seems very useful---thanks.

I didn't get the obvious success I was hoping for, but I think there's evidence that this works. In the screenshot below I'm looking at the `` for the potato group in Firefox's web inspector. Its `src` can't be loaded into Mailtrap because that's served over https, but my local homeserver is serving on `http://localhost:8008` so the cross-origin request is refused by the browser. But navigating to the source directly (rhs, chrome) shows a small potato avatar.

Screenshot from 2021-08-25 12-01-23

tests/push/test_email.py
```diff
@@ -253,6 +261,47 @@ def test_multiple_rooms(self):
...
+            email.message_from_bytes(msg)
+            .get_payload()[1]
+            .get_payload(decode=True)
+            .decode()
+        )
+        root: lxml.etree.Element = html5_parser.parse(html)
+        images = root.xpath('.//td[@class="room_avatar"]//img')
+        self.assertEqual(len(images), 1)
+        source = images[0].attrib["src"]
+        self.assertIn("_matrix/media/v1/thumbnail/DUMMY_MEDIA_ID", source)
```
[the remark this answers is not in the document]
DMRobertson: I had a quick go with using [html.parser](https://docs.python.org/3/library/html.parser.html) but found it a little awkward to use if I wanted to make a specific assertion like the above. On reflection, that specific assertion is coupled to the HTML template that gets rendered, so maybe it's not worth it as written.

>  (Or just assert that the media ID is somewhere in the content?)

Yeah, let's do this.

clokep:
@DMRobertson Is this ready for review again?

DMRobertson:
@clokep yes please! (Sorry, was just trying to get CI to run overnight).

clokep (approved):
LGTM! 🎉

PR #10661, opened 2021-08-19
Add further logging for #9533

DMRobertson:
DRAFT so I can run CI and sanity check I haven't done anything horrible

If in doubt, print it out. Thanks Erik for advising.


DMRobertson:
Gonna close this down because the most recent case of #9533 was a configuration problem, not an application problem.

PR #10667, opened 2021-08-20
Flatten tests/rest/client/{v1,v2_alpha} too

DMRobertson:
Follow-up to  #10600.


clokep (approved):
LGTM if CI passes! Can you also update mypy.ini with this change?

changelog.d/10667.misc
```diff
@@ -0,0 +1 @@
+Flatten the `tests.synapse.rests` package by moving the contents of `v1` and `v2_alpha` into the parent.
```
clokep, line 1: Could use the same changelog as #10600 if you wanted to combine them, but they're both `misc` so probably doesn't matter.

DMRobertson: What's the mechanism there, sorry? Will towncrier spot that two changelogs have the same contents and report them only once?

DMRobertson:
> LGTM if CI passes! Can you also update mypy.ini with this change?

Happy to. Bit surprised that we have to opt-in to checking files though, as opposed to excluding known-bad ones

changelog.d/10667.misc
clokep: > What's the mechanism there, sorry? Will towncrier spot that two changelogs have the same contents and report them only once?

Yes, it will combine them onto a single line and link to all the PRs. Something like "My awesome change. (123, 456)"

clokep:
> > LGTM if CI passes! Can you also update mypy.ini with this change?
> 
> Happy to. Bit surprised that we have to opt-in to checking files though, as opposed to excluding known-bad ones

Well we used to have very few files checked and have grown over time. I think in general we've just been adding to it. I'm not sure if you can opt-out.

PR #8455, opened 2020-10-02
e2e: ensure we have both master and self-signing key

TheJJ:
it seems to be possible that only one of them ends up to be cached.
when this was the case, the missing one was not fetched via federation,
and clients then failed to validate cross-signed devices.

PR #10668, opened 2021-08-20
e2e: fetch cross-signing keys when device is signed but no key cached

TheJJ:
when for a remote user we don't have cached `e2e_cross_signing_keys`
but we do have cached the devices in `device_lists_remote_cache`
and the user has cross-signing set up, missing keys are not fetched
and cross-signing verification will fail.

with this patch, the missing cross-signing keys are requested
if the cached devices list does have cross-signing signatures.

this is a follow-up for #8455.

### DIFF
--- a/changelog.d/10668.bugfix
+++ b/changelog.d/10668.bugfix
@@ -1,0 +1,1 @@
+Fix fetching of E2E cross signing keys over federation when devices are cached and have cross-signing-signature, but we don't have the cross-signing keys cached.
--- a/synapse/handlers/e2e_keys.py
+++ b/synapse/handlers/e2e_keys.py
@@ -168,6 +168,24 @@ class E2eKeysHandler:
                 ) = await self.store.get_user_devices_from_cache(query_list)
                 for user_id, devices in remote_results.items():
                     user_devices = results.setdefault(user_id, {})
+
+                    cached_cross_master = user_id in cross_signing_keys["master_keys"]
+                    self_sig_key = cross_signing_keys["self_signing_keys"].get(
+                        user_id, {}
+                    )
+                    cached_cross_selfsigning = bool(self_sig_key)
+
+                    have_xsign_keys = cached_cross_master and cached_cross_selfsigning
+
+                    # check if we are missing only one of cross-signing master or
+                    # self-signing key, but the other one is cached.
+                    # as we need both, this will issue a federation request.
+                    # if we don't have any of the keys, either the user doesn't have
+                    # cross-signing set up, or we did not fetch the
+                    # cross-signing keys yet since the device list is not (yet) updated.
+                    if cached_cross_master ^ cached_cross_selfsigning:
+                        user_ids_not_in_cache.add(user_id)
+
                     for device_id, device in devices.items():
                         keys = device.get("keys", None)
                         device_display_name = device.get("device_display_name", None)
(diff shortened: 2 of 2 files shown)


TheJJ:
I'm not sure if `if sig_device != device_id:` is the best and correct way to check for a signature being from a cross-signing key, but it was the most obvious way. Please advise :)

TheJJ:
So what is your suggestion on how we solve this bug? :) I know at least 5 users affected by this, so I guess it's plenty more.

richvdh:
> The spec doesn't say that implementations should parse the signatures and see there's a cross-signing key, but I think it's allowable for servers to do so.

@uhoreg: I'm not really following this, I'm afraid. Surely it must be up to either the client or the server to decide if it is missing a cross-signing key. If the spec doesn't say that the server needs to do it, then the clients can't rely on the server doing it, so the clients have to do it?

richvdh:
>  IIRC, even if clients request the cross-signing keys, the server won't re-fetch them if it thinks that it's already up to date.

If it's already up to date, why would the server need to re-fetch them?

uhoreg:
> > IIRC, even if clients request the cross-signing keys, the server won't re-fetch them if it thinks that it's already up to date.
> 
> If it's already up to date, why would the server need to re-fetch them?

It may not be up to date, but it may think that it's up to date if it missed a message notifying it of new keys.

> > The spec doesn't say that implementations should parse the signatures and see there's a cross-signing key, but I think it's allowable for servers to do so.
> 
> @uhoreg: I'm not really following this, I'm afraid. Surely it must be up to either the client or the server to decide if it is missing a cross-signing key. If the spec doesn't say that the server needs to do it, then the clients can't rely on the server doing it, so the clients have to do it?

The current spec kind of assumes that EDUs don't get dropped, so it doesn't really define a way of dealing with that situation.  (It has a way of dealing with dropped EDUs, but only after you've received another key update message -- it doesn't help in the situation where the dropped message is the last one that was sent.)  This is similar to the situation where the device list gets out of sync and you fail to encrypt to new devices because you don't know they exist.  IIRC Erik added a thing a while back where if you get a message claiming to be from a device you don't know about, Synapse will automagically re-fetch the user's devices from the remote server.

richvdh:
> The current spec kind of assumes that EDUs don't get dropped, so it doesn't really define a way of dealing with that situation

wait? what? Which EDUs are getting dropped? Surely that's not right? And we should be fixing that rather than layering more and more unspecified hacks on top of it?

I still don't think we should be doing this unless it's specced, so if we *should* be doing it, then it needs to be specced.

uhoreg:
> > The current spec kind of assumes that EDUs don't get dropped, so it doesn't really define a way of dealing with that situation
> 
> wait? what? Which EDUs are getting dropped? Surely that's not right? And we should be fixing that rather than layering more and more unspecified hacks on top of it?

I don't know the details of how EDUs can get dropped.  Maybe Erik remembers something from investigating the device list?

callahad:
@erikjohnston Could you please weigh in on this?

erikjohnston:
> > > The current spec kind of assumes that EDUs don't get dropped, so it doesn't really define a way of dealing with that situation
> > 
> > 
> > wait? what? Which EDUs are getting dropped? Surely that's not right? And we should be fixing that rather than layering more and more unspecified hacks on top of it?
> 
> I don't know the details of how EDUs can get dropped. Maybe Erik remembers something from investigating the device list?

We have a bunch of logic to try and ensure that device list update EDUs *don't* get dropped and will be retried, though the cross signing stuff was tacked on at a later date so the logic may be subtly broken somehow.

It's probably worth noting we do have some paranoia code when we receive encrypted events over federation to check if our cache for the sending device is stale: https://github.com/matrix-org/synapse/blob/develop/synapse/handlers/federation_event.py#L918

richvdh:
@erikjohnston 

> I.e. if we're seeing cross signing keys not being correctly cached then that's a bug we should investigate.

so to be clear - you don't think we should be taking this approach of inspecting the signatures, and instead should find out why the cross-signing keys are going missing?

erikjohnston:
> @erikjohnston
> 
> > I.e. if we're seeing cross signing keys not being correctly cached then that's a bug we should investigate.
> 
> so to be clear - you don't think we should be taking this approach of inspecting the signatures, and instead should find out why the cross-signing keys are going missing?

Yes. I'm not super against having extra checks to try and detect when things go wrong, but if we do that we should still investigate occurrences of it happening.

richvdh:
Thanks. I'm mostly opposed to adding yet more complication and magic to this already complicated, magical bit of code. If there is any way we can avoid doing so I think we should take it.

TheJJ:
Sure, we should find the root cause, but still we need a way to clean up the out-of-sync state.

And my signature check was the most obvious thing to me :) But i agree that doesn't seem like a good solution.

What we should keep though is the bottom code hunk where newly fetched xsignkeys are actually returned, because currently only cached keys make it to the `return`.

TheJJ:
So how should we proceed here? If we're sure now state is synched properly, just flushing the cache once in some db migration might do the trick already?

erikjohnston:
> So how should we proceed here? If we're sure now state is synched properly, just flushing the cache once in some db migration might do the trick already?

In terms of hunting down the root cause: step 1 would be to try and reproduce this reliably (and produce a test case). If that doesn't work out then we resort to trying to add enough logging such that if it happens again we have some hope of at least narrowing down what's going on. Unfortunately, we're a bit swamped at the minute so we likely won't have a chance to work on that ourselves right now, even though flakey cross signing is a big concern. 


> What we should keep though is the bottom code hunk where newly fetched xsignkeys are actually returned, because currently only cached keys make it to the `return`.


If there are things in this PR that would be useful besides the inspection of signature stuff then it's probably easiest to open them in a separate PR? Naively that sounds like just a good bug fix that we should land regardless of the anything else (which I'd missed) :slightly_smiling_face:

TheJJ:
Ok, I've submitted the return-fetched-keys pull request in #10912.

About finding the root cause: I'm not sure this is the right approach. It may be that this bug may no longer occur, but there _is_ said cache corruption for the keys in 3 homeservers i administer. It may be from the early days of the cross-signing implementation. Which is why we should try to detect the corruption and refetch the correct state, and that was my attempt in the patch.

erikjohnston:
I think all the changes we wanted to land in this PR have been landed separately, so I'm going to go ahead and close this (shout if that's wrong!). Thanks @TheJJ for spending the time investigating this and coming up with the various solutions :+1:

TheJJ:
Thanks! I'll test if it now works :)

TheJJ:
No sorry, my cross-signing-keys have still not been stored in the other homeserver:
- `select * from e2e_cross_signing_keys where user_id ~* 'myyuser';` -> empty
the e2exsignkeys of another user on my homeserver are cached there.

PR #10672, opened 2021-08-23
Avoid duplicate issues from Twisted trunk failures

callahad:
Setting `update_existing: true` in the `create-an-issue` GitHub Action
will avoid opening duplicate issues if an open issue already exists with
an identical title.

If no open issues match the title, then a new issue will be created.

This helps avoid spamming our issue tracker should there be a failure
when testing against Twisted's trunk.

This PR also pins the SHA of the `create-an-issue` action to mitigate
the risk of a malicious actor gaining access to JasonEtco's account.

See GitHub's page on security hardening third party actions for more:
https://docs.github.com/en/actions/learn-github-actions/security-hardening-for-github-actions#using-third-party-actions

### Pull Request Checklist

### DIFF
--- a/changelog.d/10672.misc
+++ b/changelog.d/10672.misc
@@ -1,0 +1,1 @@
+Update our periodic test of Synapse against Twisted trunk to avoid opening duplicate issues.
--- a/changelog.d/10672.misc
+++ b/changelog.d/10672.misc
@@ -1,1 +1,1 @@
-Update our periodic test of Synapse against Twisted trunk to avoid opening duplicate issues.
+Run a nightly CI build against Twisted trunk.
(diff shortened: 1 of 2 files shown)


changelog.d/10672.misc
```diff
@@ -0,0 +1 @@
+Update our periodic test of Synapse against Twisted trunk to avoid opening duplicate issues.
```
richvdh, line 1: probably just set this to the same as #10651, rather than having two separate changelog entries?

richvdh (approved):
lgtm otherwise. Thanks!

changelog.d/10672.misc
callahad: Ah right, I keep forgetting about how Towncrier does that

PR #10677, opened 2021-08-23
Correctly initialise the `synapse_user_logins` metric.

richvdh:
Fix a bug where the prometheus metrics for SSO logins wouldn't be initialised
until the first user logged in with a given auth provider.

### DIFF
--- a/synapse/rest/client/login.py
+++ b/synapse/rest/client/login.py
@@ -602,6 +602,13 @@ def register_servlets(hs, http_server):
 
 
 def _load_sso_handlers(hs: "HomeServer"):
+    """Ensure that the SSO handlers are loaded, if they are enabled by configuration.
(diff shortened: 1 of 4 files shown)


synapse/rest/client/login.py
```diff
@@ -598,3 +599,12 @@ def register_servlets(hs, http_server):
     SsoRedirectServlet(hs).register(http_server)
     if hs.config.cas_enabled:
         CasTicketServlet(hs).register(http_server)
+
+
+def _load_sso_handlers(hs: "HomeServer"):
```
erikjohnston, line 604: Can you add a quick docstring, mainly its useful to know this is safe to call multiple times.

erikjohnston (approved):

PR #10681, opened 2021-08-24
Deduplicate concurrent requests to get_state_for_groups

### LINKED ISSUE #10301: many duplicate `_get_state_groups_from_groups` queries leading to OOMs
We had a federation sender instance which died. On restart, it rapidly consumed all available ram and OOMed again.

Inspection from postgres side shows it is doing many duplicate queries of the form

```sql
WITH RECURSIVE state(state_group) AS ( VALUES(3405820::bigint) UNION ALL SELECT prev_state_group FROM state_group_edges e, state s WHERE s.state_group = e.state_group ) SELECT DISTINCT ON (type, state_key) type, state_key, event_id FROM state_groups_state WHERE state_group IN ( SELECT state_group FROM state ) ORDER BY type, state_key, state_group DESC
```

This query is `_get_state_groups_from_groups`, which is called from `_get_state_for_groups`. Although the latter has a cache, if many threads hit it at the same time, all will find the cache empty and go on to hit the database. I think we need a zero-timeout `ResponseCache` on `_get_state_groups_from_groups`.

reivilibre:
Closes #10510. Closes #10301.

### DIFF
--- a/changelog.d/10681.misc
+++ b/changelog.d/10681.misc
@@ -1,0 +1,1 @@
+Deduplicate requests in `_get_state_for_groups`.
--- a/mypy.ini
+++ b/mypy.ini
@@ -96,6 +96,7 @@ files =
   tests/handlers/test_sync.py,
   tests/rest/client/test_login.py,
   tests/rest/client/test_auth.py,
+  tests/storage/test_state.py,
   tests/util/test_itertools.py,
   tests/util/test_stream_change_cache.py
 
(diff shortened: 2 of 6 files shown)


reivilibre:
This is a draft reincarnation of #10510.

I don't fully like it. It feels like it should be possible to do something simpler, but there are quite a few cases to consider (mostly due to the `include_others` flag on `StateFilter`s). That said, I haven't come up with anything I like more so far.

I will write tests for this, because I feel like it really needs them to exercise all the conditions, but I was hoping to ask for advice on this before I sink the effort to write some tests in (though maybe that's not such a concern since they might be reusable?).

erikjohnston:
I think @reivilibre was going to try and tidy up the commits a bit and then re-request review

erikjohnston:
@reivilibre and I had a chat about this and we're changing the way we cache things a bit to avoid having a new type of cache.

Issue #10406, opened 2021-07-15
Rework how `ThirdPartyEventRules`'s `check_event_allowed` works for replacing events


babolivier:
Currently, `check_event_allowed` can return either a boolean (indicating whether an event can be allowed) or a dict (indicating that the event can be allowed, but be replaced with specific data). If it returns a dict, its content is only considered if the event is local (it's ignored in the federation code).

Making this callback return replacement data for the event feels like the wrong way to allow modules to change the content of existing events, since the function was only intended for figuring out whether an event could be allowed onto the homeserver. It's also an issue because replacing the event's data means rebuilding it with brand new auth events, which might not be possible in some cases (e.g. rejecting a local invite).

It's also an issue since the event is frozen before calling the callback, so `get_dict` will return a dict but any nested dicts (e.g. `content`) will be `frozendict`s, which isn't easily modified by a module.

For context, this feature was initially patched onto this callback to allow for a customer to build a module that hits their internal API with the json of an event and receives back whether it should allow this event, deny it, or allow it but with a different content.

anoadragon453:
Thanks for recording this. Do you have an idea for a better API shape that may sidestep these issues?

babolivier:
@anoadragon453 The new version of this callback in the [new module interface](https://matrix-org.github.io/synapse/develop/modules.html#third-party-rules-callbacks) separates the allowed/denied information from the "this event needs to be replaced" one (by returning two separate return values), but I'm still not happy with this shape. I wrote it down as an issue because I think getting to a satisfying and fully working result needs putting some thought into it, which I didn't have time for then.

richvdh:
> The new version of this callback in the new module interface separates

I think this refers to [this](https://matrix-org.github.io/synapse/latest/modules/third_party_rules_callbacks.html#check_event_allowed).

PR #10682, opened 2021-08-24
Only process replacement from `check_event_allowed` module callbacks after calling every callback

babolivier:
I haven't updated the documentation on this callback because I believe this would be the most logical way to handle this, but I'm happy to do it if the reviewer disagree.

### DIFF
--- a/changelog.d/10682.misc
+++ b/changelog.d/10682.misc
@@ -1,0 +1,1 @@
+Only process replacement from `check_event_allowed` module callbacks after calling every callback.
--- a/synapse/events/third_party_rules.py
+++ b/synapse/events/third_party_rules.py
@@ -214,6 +214,7 @@ class ThirdPartyEventRules:
         # the hashes and signatures.
         event.freeze()
 
+        final_replacement_data = None
         for callback in self._check_event_allowed_callbacks:
             try:
                 res, replacement_data = await callback(event, state_events)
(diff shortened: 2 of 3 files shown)


clokep:
This still seems to return early in the failure case though? I'm a bit confused at why this change is wanted? Is it in-case one of the callbacks stores something externally or something like that?

We should likely enforce whatever is wanted here via updates to the tests, as this seems likely to be optimized out next time someone looks at the code.

babolivier:
> This still seems to return early in the failure case though?

It only returns early if a module forbids the event from being sent, at which point it's not worth continuing processing it - which is what we also do in other similar callbacks such as `check_event_for_spam` fwiw.

> I'm a bit confused at why this change is wanted?

Because if one module says an event can go through with some changes and another says it should be blocked, the event should be blocked. At least that's how we've done it with other callbacks.

I'm also considering a more concrete use case (related to a customer) with a server uses two modules: one that demotes users with non-default PL when they leave and https://github.com/matrix-org/synapse-freeze-room which might also do stuff when a user with non-default PL leaves the room. Some callbacks may return a dummy replacement dict (i.e. just `event.get_dict()`) to force the homeserver to recalculate its context after updating the state of the room (this is to work around a bug I need to investigate a bit more where some state changes might get instantly reverted if the context isn't rebuilt, which might have some roots in the state res implem), and this would be the case for this one. We don't want the second module to be bypassed because the first one let the event through but returned a dict, because otherwise the room might not end up being frozen when it should.

It's worth noting there already are a few issues with this callback, namely with the way it replaces events, e.g. https://github.com/matrix-org/synapse/issues/10406 - as the doc says it's an experimental one that module devs shouldn't use until we figure out what to do about them.

> We should likely enforce whatever is wanted here via updates to the tests

Gah, indeed, I wanted to write a test for that but forgot :facepalm:

clokep:
> > I'm a bit confused at why this change is wanted?
> 
> Because if one module says an event can go through with some changes and another says it should be blocked, the event should be blocked. At least that's how we've done it with other callbacks.

Oh, duh! I didn't quite grok this from the description. This is fixing a bug, not changing behavior. Carry on! 👍

babolivier:
Yeah my reasoning was that we usually consider the highest module in the config to be the source of truth, though I can see how it doesn't apply here. The right solution would probably be to rebuild the event before moving onto the next module if a replacement was provided. My main motivation for getting this over the line was considering https://github.com/matrix-org/synapse/issues/10830 but as it says I'm not seeing it happen anymore, which https://github.com/matrix-org/synapse/pull/10835 seems to confirm, so I'm tempted to park this until we figure out what to do about https://github.com/matrix-org/synapse/issues/10406, which looks like a superset of this issue.

PR #10683, opened 2021-08-24
Port the SAML user mapping providers module interface to the new generic interface.

Azrenbeth:
Port the SAML user mapping providers module interface to the new generic interface.

### DIFF
--- a/changelog.d/10683.feature
+++ b/changelog.d/10683.feature
@@ -1,0 +1,1 @@
+Port the SAML user mapping providers module interface to the new generic interface.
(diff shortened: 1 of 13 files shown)


erikjohnston:
@babolivier is this something that should be resurrected, or?

babolivier:
> @babolivier is this something that should be resurrected, or?

I don't think so. There were discussions a few months ago going in the general direction that OIDC and SAML mapping providers should be merged to reduce code duplication. But regardless of whether that's what we're going to end up doing, this PR is probably getting a bit too big to be properly reviewed (and doesn't seem reviewable commit by commit), so while it's probably a good source of inspiration for porting the interface in the future (if we don't decide to go ahead with keeping one type of mapping providers) it should probably be closed.

Issue #10065, opened 2021-05-25
Synapse workers Complement image results in flaky tests due to inconsistent worker process init


anoadragon453:
The worker-version of Synapse running in Complement (as described [here](https://github.com/matrix-org/synapse/blob/develop/docker/README-testing.md#testing-with-postgresql-and-single-or-multi-process-synapse)) currently uses [Supervisor](https://supervisord.org) as an init system to start all worker processes in the container. See the current [config template](https://github.com/matrix-org/synapse/blob/557635f69ab734142ae5889e215ea512f7678f21/docker/conf-workers/supervisord.conf.j2) we're using.

This works, and eventually all processes start up. However, Complement checks whether a homeserver is ready to start testing by the fact that it responds successfully to a `GET /_matrix/client/versions` call. This endpoint may be successfully responded to by a worker that has started, while other workers are still starting up. This inconsistency can lead to test failures, where Complement finds a 502 from a call to a different endpoint that should be handled by a different worker. Since that worker hasn't started yet, nginx returns a 502, and the test fails.

The result of this is flaky Complement tests - which nobody wants.

I believe the solution is to start groups of processes in the container through a priority system. Only should the next group be started once the previous has successfully responded to healthchecks (indicating the process is ready to receive connections):

1. Redis
1. the main Synapse process
    * thus all database migrations are handled before any workers start up.
1. all worker Synapse processes
1. nginx

(Note that caddy [just used for custom CA stuff] and Postgres are started before even Supervisor is.) By starting nginx at the very end, which is the reverse proxy that actually routes matrix requests to the appropriate Synapse process, Complement will not receive a successful response to `/_matrix/client/versions` until everything else has started.

[...]

PR #10039, opened 2021-05-21
Run complement with Synapse workers manually.

clokep:
This attempts to be able to run complement with Synapse worker mode via the `complement.sh` script. Adding a `WORKERS=1` will cause them to be used.

It also fixes a couple of minor things necessary to get this to work.

Corresponding complement PR: matrix-org/complement#116.

PR #10686, opened 2021-08-24
Ensure the base image is rebuilt when running complement with workers.

clokep:
This fixes a bug in the changes to `complement.sh` from #10039 -- currently either the `matrixdotorg/synapse` Docker image is built *or* the `matrixdotorg/synapse-workers` image is built via `complement.sh`, it should be that the `matrixdotorg/synapse` image is *always* built and we either build complement directly on it, or first build `matrixdotorg/synapse-workers` on it and then complement on that.

Vaguely related to matrix-org/synapse#10065.

### DIFF
--- a/changelog.d/10686.misc
+++ b/changelog.d/10686.misc
@@ -1,0 +1,1 @@
+Update `complement.sh` to rebuild the base Docker image when run with workers.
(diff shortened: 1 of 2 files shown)


clokep:
Essentially `complement.sh` now properly follows the steps in the docs: https://github.com/matrix-org/synapse/blob/develop/docker/README-testing.md#testing-with-postgresql-and-single-or-multi-process-synapse

squahtx (approved):

Issue #5677, opened 2019-07-12
User Directory leaks Per-room Nicknames and Avatars


reivilibre:
### Description

The User Directory leaks display names and avatars for a user that are sent in only one room.
For example, by manually crafting a `m.room.member` state event – or recently using the `/myroomnick` command in Riot/Web, even if the state event is sent in a private room.

### Steps to reproduce

- Using Riot/Web with account `@alice:example.org`, open a private chat (such as a direct chat with a close friend)
- Issue the command `/myroomnick Freddy`, which sends a `m.room.member` state event into only that room with a custom nickname.
- From another account, say `@bob:example.org`, open up the User Search
- Search for 'Freddy' or 'alice' — `@alice:example.org` will be listed with the name 'Freddy'
    - Note: this assumes that alice is visible to bob in the user directory – i.e. alice is in a public room known to the homeserver AND/OR alice and bob share a private room together.
- (Note that Synapse's user_directory table also reflects the change)

#### Expected Behaviour

alice's original display name should be shown in the user search.

#### Implications

This has privacy implications – a nickname set in a private room with a close friend may be quite personal and perhaps embarrassing if seen by other users.

### Version information

- **Homeserver**: librepush.net

If not matrix.org:

- **Version**: 1.1.0+bionic1

not really relevant, I suspect:

- **Install method**: Debian packages

- **Platform**: Ubuntu 18.04 in an LXC container on NixOS

reivilibre:
Note that I have not tested this interaction over federation – I suspect that only servers receiving the `m.room.member` state event will update their user directories with the private nickname, but have not tested this.

Mikaela:
Has this issue or it's implications been thought about in further detail than discussed in this issue? I recently blogged about [inconsistencies with Element Web/iOS and Matrix/Synapse privacy issues](https://mikaela.info/blog/english/2021/08/03/matrix-perfect-privacy-not.html) and this has been the most surprising to many people who have commented on it, and these scenarios have been offered to me:

* Youth belonging to [gender](https://en.wikipedia.org/wiki/Transgender_Day_of_Remembrance) (or [sexual minority](https://en.wikipedia.org/wiki/Violence_against_LGBT_people)) in less tolerant household uses Matrix for peer support and in a peer support group sets a name and avatar matching their identity. Their information is accidentally leaked to someone they know on Matrix due to this issue and they get thrown out, face violence or killed.
[...]

PR #10695, opened 2021-08-25
Prevent local user profiles leaking when a "per-room profile" is changed

DMRobertson:
The plan is described [here](https://gist.github.com/DMRobertson/b63d7caa1c7c1e39d88d98f27b233bc2). In a nutshell, I want to make it so that local users' entries in the user directory only ever refer to their "public" profile, and completely ignore any per-room nicknames.

This goes some way towards fixing #5677, but we still have the problem of how to handle remote users' profiles.

### DIFF
--- a/changelog.d/10695.bugfix
+++ b/changelog.d/10695.bugfix
@@ -1,1 +1,1 @@
-Stop leaking per-room nicknames and avatars for users on this homeserver. NB: synapse can still leak nicknames and avatars of users from *other* homeservers.
+Fixed one case of leaking per-room nicknames and avatar to the user directory, pending further work to tackle other cases.
--- a/changelog.d/10695.bugfix
+++ b/changelog.d/10695.bugfix
@@ -1,1 +1,1 @@
-TODO write a proper changelog dmr.
+Stop leaking per-room nicknames and avatars for users on this homeserver. NB: synapse can still leak nicknames and avatars of users from *other* homeservers.
--- a/changelog.d/10695.bugfix
+++ b/changelog.d/10695.bugfix
@@ -1,1 +1,1 @@
-Fixed one case of leaking per-room nicknames and avatar to the user directory, pending further work to tackle other cases.
+Fix leaking per-room nicknames and avatars to the user directory for local users when they update their per-room nickname or avatar.
--- a/changelog.d/10695.bugfix
+++ b/changelog.d/10695.bugfix
@@ -1,0 +1,1 @@
+TODO write a proper changelog dmr.
--- a/docs/user_directory.md
+++ b/docs/user_directory.md
@@ -17,8 +17,7 @@ Data model
 There are five relevant tables:
 
 * `user_directory`. This contains the user_id, display name and avatar we'll
-  return when you search the directory. For some reason the `user_directory`
-  also tracks a room_id.
+  return when you search the directory.
   - Because there's only one directory entry per user, it's important that we only
     ever put publicly visible names here. Otherwise we might leak a private
     nickname or avatar used in a private room.
@@ -37,6 +36,6 @@ There are five relevant tables:
 * `users_in_public_rooms`. Contains associations between users and the public rooms they're in.
   Used to determine which users are in public rooms and should be publicly visible in the directory.
 
-* `users_who_share_private_rooms`. Rows are triples `(L, U, room id)` where `L`
-   is a local user and `R` is a local or remote user. `L` and `R` should be
+* `users_who_share_private_rooms`. Rows are triples `(L, M, room id)` where `L`
+   is a local user and `M` is a local or remote user. `L` and `M` should be
    different, but this isn't enforced by a constraint.
--- a/docs/user_directory.md
+++ b/docs/user_directory.md
@@ -10,3 +10,32 @@ DB corruption) get stale or out of sync.  If this happens, for now the
 solution to fix it is to execute the SQL [here](https://github.com/matrix-org/synapse/blob/master/synapse/storage/schema/main/delta/53/user_dir_populate.sql)
 and then restart synapse. This should then start a background task to
 flush the current tables and regenerate the directory.
+
+Data model
+----------
+
+There are five relevant tables:
+
+* `user_directory`. This contains the user_id, display name and avatar we'll
+  return when you search the directory. For some reason the `user_directory`
+  also tracks a room_id.
+  - Because there's only one directory entry per user, it's important that we only
+    ever put publicly visible names here. Otherwise we might leak a private
+    nickname or avatar used in a private room.
+  - Indexed on rooms. Indexed on users.
+
+* `user_directory_search`. To be joined to `user_directory`. It contains an extra
+  column that enables full text search based on user ids and display names.
+  Different schemas for sqlite and postgres with different code paths to match.
+  - Indexed on the full text search data. Indexed on users.
+
+* `user_directory_stream_pos`. When the initial background update to populate
+  the directory is complete, we record a stream position here. This indicates
+  that synapse should now listen for room changes and incrementally update
+  the directory where necessary.
+
+* `users_in_public_rooms`. Tracks both users and which rooms they're in.
+
+* `users_who_share_private_rooms`. Rows are triples `(L, U, room id)` where `L`
+   is a local user and `R` is a local or remote user. `L` and `R` should be
+   different, but this isn't enforced by a constraint.
--- a/docs/user_directory.md
+++ b/docs/user_directory.md
@@ -26,7 +26,7 @@ There are five relevant tables:
 
 * `user_directory_search`. To be joined to `user_directory`. It contains an extra
   column that enables full text search based on user ids and display names.
-  Different schemas for sqlite and postgres with different code paths to match.
+  Different schemas for SQLite and Postgres with different code paths to match.
   - Indexed on the full text search data. Indexed on users.
 
 * `user_directory_stream_pos`. When the initial background update to populate
@@ -34,7 +34,8 @@ There are five relevant tables:
   that synapse should now listen for room changes and incrementally update
   the directory where necessary.
 
-* `users_in_public_rooms`. Tracks both users and which rooms they're in.
+* `users_in_public_rooms`. Contains associations between users and the public rooms they're in.
+  Used to determine which users are in public rooms and should be publicly visible in the directory.
 
 * `users_who_share_private_rooms`. Rows are triples `(L, U, room id)` where `L`
    is a local user and `R` is a local or remote user. `L` and `R` should be
--- a/synapse/handlers/user_directory.py
+++ b/synapse/handlers/user_directory.py
@@ -201,8 +201,7 @@ class UserDirectoryHandler(StateDeltasHandler):
                 room_id, prev_event_id, event_id, typ
             )
         elif typ == EventTypes.Member:
-            is_support_user = await self.store.is_support_user(state_key)
-            if is_support_user:
+            if await self._user_omitted_from_directory(state_key):
                 return
 
             joined = await self._get_key_change(
@@ -257,6 +256,16 @@ class UserDirectoryHandler(StateDeltasHandler):
                 await self._handle_remove_user(room_id, state_key)
         else:
             logger.debug("Ignoring irrelevant type: %r", typ)
+
+    async def _user_omitted_from_directory(self, user_id: str) -> bool:
+        """We want to ignore events from "hidden" users who shouldn't be exposed
+        to real users."""
+        if await self.store.is_support_user(user_id):
+            return True
+        if self.store.get_if_app_services_interested_in_user(user_id):
+            return True
+
+        return False
 
     async def _handle_room_publicity_change(
         self,
@@ -352,38 +361,31 @@ class UserDirectoryHandler(StateDeltasHandler):
         is_public = await self.store.is_room_world_readable_or_publicly_joinable(
             room_id
         )
-        # Now we update users who share rooms with users.
-        other_users_in_room = await self.store.get_users_in_room(room_id)
-
         if is_public:
             await self.store.add_users_in_public_rooms(room_id, (user_id,))
         else:
+            other_users_in_room = await self.store.get_users_in_room(room_id)
+            other_users_in_room = [
+                other_user
+                for other_user in other_users_in_room
+                if not await self._user_omitted_from_directory(other_user)
+            ]
+
             to_insert = set()
 
             # First, if they're our user then we need to update for every user
             if self.is_mine_id(user_id):
+                for other_user_id in other_users_in_room:
+                    if user_id == other_user_id:
+                        continue
 
-                is_appservice = self.store.get_if_app_services_interested_in_user(
-                    user_id
-                )
-
-                # We don't care about appservice users.
-                if not is_appservice:
-                    for other_user_id in other_users_in_room:
-                        if user_id == other_user_id:
-                            continue
-
-                        to_insert.add((user_id, other_user_id))
+                    to_insert.add((user_id, other_user_id))
 
             # Next we need to update for every local user in the room
             for other_user_id in other_users_in_room:
                 if user_id == other_user_id:
                     continue
-
-                is_appservice = self.store.get_if_app_services_interested_in_user(
-                    other_user_id
-                )
-                if self.is_mine_id(other_user_id) and not is_appservice:
+                if self.is_mine_id(other_user_id):
                     to_insert.add((other_user_id, user_id))
 
             if to_insert:
--- a/synapse/handlers/user_directory.py
+++ b/synapse/handlers/user_directory.py
@@ -376,11 +376,14 @@ class UserDirectoryHandler(StateDeltasHandler):
 
     async def _handle_remove_user(self, room_id: str, user_id: str) -> None:
         """Called in two cases:
-        - when the given user has left the room
-        - we can no longer see that the given user is in this room.
+        1. The given user has left the given room.
+        2. The last local user (someone else) just left the given room. So
+           so everyone else in the room is remote. We may no longer need to
+           include the given user (`user_id`) in the directory.
 
         Args:
-            room_id: The room ID which the given user was in
+            user_id: The user we may need to remove from the directory
+            room_id: The room which someone just left
         """
         logger.debug("Removing user %r", user_id)
 
--- a/synapse/handlers/user_directory.py
+++ b/synapse/handlers/user_directory.py
@@ -172,7 +172,8 @@ class UserDirectoryHandler(StateDeltasHandler):
                 )
 
                 logger.debug("Handling %d state deltas", len(deltas))
-                await self._handle_deltas(deltas)
+                for delta in deltas:
+                    await self._handle_delta(delta)
 
                 self.pos = max_pos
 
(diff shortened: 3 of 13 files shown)


DMRobertson:
Tested this locally against element-web. It offered some kind of client-side suggestions which seemed to take into account per-room nicknames. However I confirmed the underlying response to `/user_directory/search` was returning the expected information: it did not yield results when searching for the per-room name, and it yielded the public name when searching by mxid.

DMRobertson:
There is a bit of a hole here. We still use the event displayname and avatar when handling someone joining a room. I don't think many clients will offer the chance to set a per-room nickname before you join, but it's possible from the protocol's POV.

reivilibre:
> There is a bit of a hole here. We still use the event displayname and avatar when handling someone joining a room. I don't think many clients will offer the chance to set a per-room nickname before you join, but it's possible from the protocol's POV.

out of interest, why is this the case? Is it difficult for some reason?

changelog.d/10695.bugfix
```diff
@@ -0,0 +1 @@
+TODO write a proper changelog dmr.
```
reivilibre, line 1: ... :D

DMRobertson:
> > There is a bit of a hole here. We still use the event displayname and avatar when handling someone joining a room.
> 
> out of interest, why is this the case? Is it difficult for some reason?

Just an oversight on my part. Until I started looking at this for remote users, I hadn't properly considered that local users could go down this code path (the branch with ` elif joined is MatchChange.now_true`) with an event whose name didn't match their profile.

synapse/handlers/user_directory.py
```diff
@@ -217,30 +229,32 @@ async def _handle_deltas(self, deltas: List[Dict[str, Any]]) -> None:
                     else:
                         logger.debug("Server is still in room: %r", room_id)
 
-                is_support = await self.store.is_support_user(state_key)
-                if not is_support:
-                    if change is None:
-                        # Handle any profile changes
+                is_support_user = await self.store.is_support_user(state_key)
+                if is_support_user:
+                    continue
```
reivilibre, line 234:
```suggestion
                if is_support_user:
                    # support users should not be added to the directory
                    continue
```

just in case we wonder what this is for (since it won't be long before we all forget what support users are again..)

synapse/handlers/user_directory.py
```diff
@@ -217,30 +229,32 @@ async def _handle_deltas(self, deltas: List[Dict[str, Any]]) -> None:
...
-                        )
+                    profile = ProfileInfo(
+                        avatar_url=event.content.get("avatar_url"),
+                        display_name=event.content.get("displayname"),
+                    )
 
-                        await self._handle_new_user(room_id, state_key, profile)
-                    else:  # The user left
-                        await self._handle_remove_user(room_id, state_key)
+                    await self._handle_new_user(room_id, state_key, profile)
```
reivilibre, line 255: As you mentioned, this has the hole of 'new joins will still trust the event content'.

Since we're expecting local users to populate themselves into the directory as soon as they update their profile, there's no need to run the risk of overwriting that with the content from the event (which feasibly may allow per-room information in the future).

I think it might be worth splitting up `handle_new_user`, which is currently these two chunks, into two separate things:

* updating the profile itself (is that just `update_profile_in_user_dir`?)
  * **now**: don't trigger on state events in rooms for local users (but continue doing so for remote users until we get the federation requests sorted out)
  * **future**: only trigger on local user profile updates and when we get responses for our federation profile queries
* updating the 'users in public rooms' and 'users that share private rooms' tables — never from state events
  * carry on triggering when you notice a new join in any room

shout if this seems silly

reivilibre (changes requested):

synapse/handlers/user_directory.py
```diff
@@ -341,11 +341,13 @@ async def _handle_new_user(
...
-        await self.store.update_profile_in_user_dir(
-            user_id, profile.display_name, profile.avatar_url
-        )
+        # Local users' directory entries are created and updated in tandem with
+        # their global profile, not in response to room events.
+        if not self.is_mine_id(user_id):
+            logger.debug("Adding new user to dir, %r", user_id)
+            await self.store.update_profile_in_user_dir(
+                user_id, profile.display_name, profile.avatar_url
+            )
```
reivilibre, line 350: that seems reasonable too, suspect you'll rip it out when we do remote users properly but LGTM for now

reivilibre:
my review got ninja'd so hopefully things didn't get all tangled up and stop making sense

reivilibre:
Worth remarking that this PR makes a decision that the user directory table should contain rows for local users, even if they aren't in any rooms (and thus won't appear in the directory) — the search code does the right thing and we rely on that.

synapse/handlers/user_directory.py
DMRobertson: > As you mentioned, this has the hole of 'new joins will still trust the event content'.

Where does the event content actually come from? It seems to come from the homeserver of the person joining; can't see anything in the CSAPI spec suggesting clients can customise this.

> updating the 'users in public rooms' and 'users that share private rooms' tables — never from state events

Erm, what does state event mean here? I think you're getting at "we don't need to do this when someone's name/avatar changes", just on joining or leaving?

synapse/handlers/user_directory.py
reivilibre: They're users that really shouldn't be counted as users and shouldn't be exposed to real users on the server.

The common use case is support: EMS (presumably) might have these accounts on a customer's homeserver so that they can log in, check that things are working (do messages send? how long does it take? It's possible [matrixmon](https://github.com/matrix-org/matrixmon/) uses a support user on these servers for automatic testing of 'Can I send an event and receive it?'), etc

DMRobertson: Maybe the right thing to do (analogous to your suggestion elsewhere) is to separate the management of `user_directory` from the management of the `users_in_public_rooms` and `users_who_share_private_rooms`. The former can skip support users fairly safely I think?

synapse/handlers/user_directory.py
reivilibre: > Where does the event content actually come from? It seems to come from the homeserver of the person joining; can't see anything in the CSAPI spec suggesting clients can customise this.

Yes, that's right. Maybe the CS API doesn't support it today, but it's not unreasonable to think it would be possible one day (and in our case, I think it'd be more consistent if we didn't use the nicknames from state events in some cases and not others). -- Or I can imagine your name being inherited from a space, when you join a room that's in that space. I'm aware some people are quite enthusiastic about using spaces for segregating their (work, play/gaming, personal, ...) 'lives' and I believe this kind of thing has already been requested (since people playing games put 'gamer tags' in their name or something, but I can also see it being useful for having your real name in some contexts and not others, for example).

> Erm, what does state event mean here? I think you're getting at "we don't need to do this when someone's name/avatar changes", just on joining or leaving?

I'm specifically talking about the `m.room.member` state event, which (if you know what state events are, if not ask!) form a key-value mapping {user_id → (left/joined/kicked/banned/knocking, display name, avatar)}.

Type `/devtools` in Element Web and press 'Explore room state', if that helps you see things.

I'm trying to get at: we shouldn't be using `m.room.member` state event displaynames in the user directory *ever* (except for remote users for now, until we can properly accommodate those).

Crucially, the fact that a state event corresponds to a new user joining a room or them 'just' updating their profile information, should not influence that we use/don't use that information.

synapse/handlers/user_directory.py
reivilibre: Without being able to fully imagine the difficulty of doing that cleanly, sounds sensible.

That said, `users_in_public_rooms` and `users_who_share_private_rooms` are *purely* for the user directory, so support users can be skipped from those as well.

I think, for your sanity, it should be sensible to just try and forget about support users as soon as you can. `continue` as you have it seems reasonable.

(do we even need this check here though? We should be preventing support users from being added to the directory already, and the plan is to not add local users to the directory from this background process, so thinking further, I think you can just forget about special casing them now?)

DMRobertson: > That said, users_in_public_rooms and users_who_share_private_rooms are purely for the user directory, so support users can be skipped from those as well.

Slightly anxious that this isn't obvious from the table names. But I'm struggling to think of a reason where you would actively want to count or search for these support users.

Yeah. Let's suppress them!

DMRobertson: Further to this: later on we skip adding users to the `users_who_share_private_room` and `users_in_public_room` tables. I think we can exclude these at the same time that we exclude support users. Do you agree?

reivilibre: sounds reasonable

synapse/handlers/user_directory.py
```diff
@@ -290,13 +304,13 @@ async def _handle_room_publicity_change(
...
-        if change and not is_public:
+        if publicity is MatchChange.now_true and not is_public:
             # If we became world readable but room isn't currently public then
             # we ignore the change
             return
-        elif not change and is_public:
+        elif publicity is MatchChange.now_false and is_public:
             # If we stopped being world readable but are still public,
             # ignore the change
             return
```
DMRobertson, line 316: I think I've spotted another problem.

If a room becomes public, within `_handle_room_publicity_change` we'll fetch `other_users_in_room_with_profiles` by calling `get_users_in_room_with_profiles`. This will read from `room_memberships`, which will contain any per-room nicknames and avatars. We'll then shove those back into the directory by calling `_handle_new_user`.

(The relevant bit of code is just below here, but I couldn't add a comment because it's not part of the diff. GRR)

DMRobertson: (The above is okay from a privacy point of view, but breaks the desired invariant "the directory only contains public profile information)

reivilibre: > If a room becomes public, within _handle_room_publicity_change we'll fetch other_users_in_room_with_profiles by calling get_users_in_room_with_profiles. This will read from room_memberships, which will contain any per-room nicknames and avatars. We'll then shove those back into the directory by calling _handle_new_user.

Yup, so I think `_handle_new_user` is basically doing too much (it shouldn't be involved with display names or avatars).

A room becoming public should have no effect upon the `user_directory` table anymore (all the users will still/already be in there!).

DMRobertson:
@reivilibre  many thanks for your eyes on this one. I'm about to push another batch of changes. This one had a fair bit of movement in due to refactoring, so it might be easiest to follow commit-by-commit. Thanks again!

DMRobertson:
There's a lot of duplication between `UserDirectoryBackgroundUpdateStore.populate_user_directory_process_rooms` and `UserDirectoryHandler._track_user_joined_room`.

I'd like to introduce free functions in the handler's module that I can import in the store. (Or maybe the other way round?). I was going to say "this can be a separate PR", but...

I've just noticed that `_populate_user_directory_process_rooms` in `UserDirectoryBackgroundUpdateStore` will read from room_memberships and so will populate the directory with per-room data. So maybe it is worth introducing those free functions now.

DMRobertson:
> I've just noticed that `_populate_user_directory_process_rooms` in `UserDirectoryBackgroundUpdateStore` will read from room_memberships and so will populate the directory with per-room data. So maybe it is worth introducing those free functions now.

Nah, this PR is long enough as it is. Let's consider this "don't leak local nicknames  on updates". I'll open a separate PR for "don't leak local nicknames  on rebuild".

tests/handlers/test_user_directory.py
```diff
@@ -639,6 +640,64 @@ def _add_user_to_room(
             self.hs.get_storage().persistence.persist_event(event, context)
         )
 
+    def test_making_room_public_doesnt_alter_directory_entry(self):
+        """Per-room names shouldn't go to the directory when the room becomes public.
+
+        I made this a synapse test case rather than a complement one because
```
reivilibre, line 646:
```suggestion
        I made this a Synapse test case rather than a Complement one because
```

tests/handlers/test_user_directory.py
```diff
@@ -639,6 +640,64 @@ def _add_user_to_room(
...
+        I think this is (strictly speaking) an implementation choice. Synapse
+        has chosen to only ever use the public profile when responding to a user
+        directory search. There's no privacy leak here, because making the room
+        public discloses the per-room name.
```
reivilibre, line 650: yeah. This one is a little bit loose. (Maybe we should get spec involved?)

I think Synapse is aiming to do the right thing: just because a room is public (either from history readability or relaxed join rules), doesn't mean you want your name in there to leak (it might be related to the community in use, but unprofessional, personal or weird for use outside of that community* — and sure, a motivated enemy can go looking in all the public rooms everywhere, but we don't necessarily want to make it easy for them).

*a project I know has a public Matrix room where the devs have `(frontend)` or similar labels on their names. But you don't necessarily want that to show up in a personal context.

I have no objections to leaving this as a Synapse test (especially because the knowledge that the room dir changes when a room goes public is kind of Synapse-specific), but really just thinking about preventing this issue for future homeserver implementors.

tests/handlers/test_user_directory.py
```diff
@@ -639,6 +640,64 @@ def _add_user_to_room(
...
+
+        The spec doesn't mandate anything about _how_ a user
+        should appear in a /user_directory/search result. Hypothetical example:
+        suppose Bob searches for Alice. When representing Alice in a search
+        result, it's reasonable to use any of Alice's nicknames that Bob is
+        aware of. Heck, maybe we even want to use lots of them in a combined
+        displayname like `Alice (aka "ali", "ally", "41iC3")`.
```
reivilibre, line 657: There are localisation & context issues with that idea, I really think this needs spec work or being left to clients to decide. (But really, Synapse should avoid making the decision to output hardcoded English text.)

Having one canonical name — the profile name — should be understandable and clear to users about what is exposed. It also means that you don't have to fear using one single Matrix account in multiple contexts (even if the current support for having a split-life kind of account is not fully there).

tests/handlers/test_user_directory.py
```diff
@@ -639,6 +640,64 @@ def _add_user_to_room(
...
+        )
+
+        # One of them makes the room public.
+        self.helper.send_state(
+            room,
+            "m.room.join_rules",
+            {"join_rule": "public"},
+            alice_token,
+        )
+        # Wait for the handler to process the event
```
reivilibre, line 689: we're not actually waiting, we're just checking.

```suggestion
        # Check that Alice is now recorded as being in a public room
```

further, it'd be worth checking that this does NOT hold — before you send the join rule event. (So we know the test is actually testing something)

tests/handlers/test_user_directory.py
```diff
@@ -639,6 +640,64 @@ def _add_user_to_room(
...
+        self.assertIn((alice, room), self.get_users_in_public_rooms())
+
+        # Alice's nickname is now publicly visible, but this shouldn't change
+        # her entry in the user directory.
```
reivilibre, line 693: this comment is likely to be confusing when all the context has left us.

I don't think it really needs much other than just remarking it stayed the same.

```suggestion
        # Alice's display name remains the same in the user directory.
```

tests/handlers/test_user_directory.py
```diff
@@ -639,6 +640,64 @@ def _add_user_to_room(
             self.hs.get_storage().persistence.persist_event(event, context)
         )
 
+    def test_making_room_public_doesnt_alter_directory_entry(self):
+        """Per-room names shouldn't go to the directory when the room becomes public.
```
reivilibre, line 644: (feels a little bit funny testing that something *doesn't* happen :smile:, but I guess since it historically did, it makes sense)

changelog.d/10695.bugfix
```diff
@@ -1 +1 @@
-TODO write a proper changelog dmr.
\ No newline at end of file
+Fixed one case of leaking per-room nicknames and avatar to the user directory, pending further work to tackle other cases.
```
reivilibre, line 1:
```suggestion
Fix leaking per-room nicknames and avatars to the user directory for local users when they update their per-room nickname or avatar.
```

We usually write these as 'Fix' rather than 'Fixed'.

When you come and make another PR, rewrite both changelogs so that they're the same (but also corrected to be what you've actually done in total) — they'll get merged in the final changelog when we do a release.

Equally, if a release goes out before you finish everything else, you still have a sensible log entry that says what you've done.

reivilibre (changes requested):
Some substantial work has been done here; thanks for taking it on.

Happy that you're cleaning it up as you go!

docs/user_directory.md
```diff
@@ -10,3 +10,32 @@ DB corruption) get stale or out of sync.  If this happens, for now the
...
+  - Indexed on the full text search data. Indexed on users.
+
+* `user_directory_stream_pos`. When the initial background update to populate
+  the directory is complete, we record a stream position here. This indicates
+  that synapse should now listen for room changes and incrementally update
+  the directory where necessary.
+
+* `users_in_public_rooms`. Tracks both users and which rooms they're in.
+
+* `users_who_share_private_rooms`. Rows are triples `(L, U, room id)` where `L`
```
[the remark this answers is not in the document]
DMRobertson: Hmm. Didn't want to suggest that `R` must be remote. Might change this to `L` and `M`.

synapse/handlers/user_directory.py
```diff
@@ -315,102 +323,81 @@ async def _handle_room_publicity_change(
...
                     to_insert.add((other_user_id, user_id))
 
             if to_insert:
                 await self.store.add_users_who_share_private_room(room_id, to_insert)
 
     async def _handle_remove_user(self, room_id: str, user_id: str) -> None:
-        """Called when we might need to remove user from directory
+        """Called in two cases:
+        - when the given user has left the room
+        - we can no longer see that the given user is in this room.
```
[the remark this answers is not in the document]
DMRobertson: We call this function in two cases:
1. `user_id` leaves the room
2. the last local user left this room, so everyone else in the room is remote. We no longer want may no longer want to show `user_id` in the directory.

... maybe that's just a better description

tests/handlers/test_user_directory.py
DMRobertson: Possibly overkill. I'm just in the habit of reproducing bugs with a test case.

tests/handlers/test_user_directory.py
DMRobertson: > There are localisation & context issues with that idea, I really think this needs spec work or being left to clients to decide.

Entirely agreed and not proposing this; just trying to rationalise Synapse versus Complement.

docs/user_directory.md
```diff
@@ -10,3 +10,32 @@ DB corruption) get stale or out of sync.  If this happens, for now the
...
 flush the current tables and regenerate the directory.
+
+Data model
+----------
+
+There are five relevant tables:
+
+* `user_directory`. This contains the user_id, display name and avatar we'll
+  return when you search the directory. For some reason the `user_directory`
+  also tracks a room_id.
```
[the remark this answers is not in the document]
DMRobertson: Suggest we leave it be for now and come back to it after this leaking business is fixed.

tests/handlers/test_user_directory.py
reivilibre: I agree, though this one feels funny in a way I can't quite describe. :)

synapse/handlers/user_directory.py
```diff
@@ -376,11 +376,14 @@ async def _track_user_joined_room(self, room_id: str, user_id: str) -> None:
...
+        1. The given user has left the given room.
+        2. The last local user (someone else) just left the given room. So
+           so everyone else in the room is remote. We may no longer need to
+           include the given user (`user_id`) in the directory.
 
         Args:
-            room_id: The room ID which the given user was in
+            user_id: The user we may need to remove from the directory
+            room_id: The room which someone just left
         """
```
reivilibre: thanks for this!

reivilibre (commented):
LGTM, though I'm still new to reviewing so would like to get another pair of eyes on this (both to check that I haven't missed anything, and so that I can see what else I should be looking out for as a reviewer).

docs/user_directory.md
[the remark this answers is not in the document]
reivilibre: yeah, leave for now, not your problem.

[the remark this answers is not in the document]
reivilibre: #10770 — marking this review point as resolved

DMRobertson:
@reivilibre  thank you for reviewing this. @richvdh took a look at this too but struggled to follow. I offered to chop this up into lots of smaller PRs to make it easier to follow. With that in mind, I'll close this.

PR #10566, opened 2021-08-10
Allow room creator to send MSC2716 related events in existing room versions

MadLittleMods:
Allow room creator to send MSC2716 related events in existing room versions.

> One way of getting this feature into existing room versions is instead of changing the auth rules we'd instead only process the insertion/marker events that are sent by the creator of the room? [...] That hopefully should be enough to support the basic use cases for existing rooms, while doing it "properly" for newer rooms?
>
 > *-- discussed at https://github.com/matrix-org/matrix-doc/pull/2716/#discussion_r682474869*

Complement tests added in https://github.com/matrix-org/complement/pull/192

~~Restoring `get_create_event_for_room_txn` from, https://github.com/matrix-org/synapse/pull/10245/commits/44bb3f0cf5cb365ef9281554daceeecfb17cc94d~~

Part of MSC2716: https://github.com/matrix-org/matrix-doc/pull/2716

### Dev notes

```sh
$ psql synapse

# \d+ rooms
                                          Table "public.rooms"
        Column        |  Type   | Collation | Nullable | Default | Storage  | Stats target | Description
----------------------+---------+-----------+----------+---------+----------+--------------+-------------
 room_id              | text    |           | not null |         | extended |              |
 is_public            | boolean |           |          |         | plain    |              |
 creator              | text    |           |          |         | extended |              |
 room_version         | text    |           |          |         | extended |              |
 has_auth_chain_index | boolean |           |          |         | plain    |              |
Indexes:
    "rooms_pkey" PRIMARY KEY, btree (room_id)
    "public_room_index" btree (is_public)
Referenced by:
    TABLE "destination_rooms" CONSTRAINT "destination_rooms_room_id_fkey" FOREIGN KEY (room_id) REFERENCES rooms(room_id)
Access method: heap

# \q
```

```sh
$ psql synapse

# Show syntax errors
> \set VERBOSITY verbose

> SELECT * FROM events WHERE event_id = '$zpFflMEBtZdgcMQWTakaVItTLMjLFdKcRWUPHbbSZJl';
> SELECT * FROM event_json WHERE event_id = '$zpFflMEBtZdgcMQWTakaVItTLMjLFdKcRWUPHbbSZJl';
```

```sh
$ sqlite3
> .open ./scripts/docker/matrix/synapse/data/homeserver.db

# List databases
> .database

# List tables
> .tables

# Show schema of table
> .schema events

[...]

PR #10697, opened 2021-08-25
Populate `rooms.creator` field for easy lookup

MadLittleMods:
Populate `rooms.creator` field for easy lookup

Part of https://github.com/matrix-org/synapse/pull/10566 for [MSC2716](https://github.com/matrix-org/matrix-doc/pull/2716)

 - Fill in creator whenever we insert into the rooms table
 - Add background update to backfill any missing creator values
 
 
## Dev notes

```sh
$ psql synapse

# \d+ rooms
                                          Table "public.rooms"
        Column        |  Type   | Collation | Nullable | Default | Storage  | Stats target | Description
----------------------+---------+-----------+----------+---------+----------+--------------+-------------
 room_id              | text    |           | not null |         | extended |              |
 is_public            | boolean |           |          |         | plain    |              |
 creator              | text    |           |          |         | extended |              |
 room_version         | text    |           |          |         | extended |              |
 has_auth_chain_index | boolean |           |          |         | plain    |              |
Indexes:
    "rooms_pkey" PRIMARY KEY, btree (room_id)
    "public_room_index" btree (is_public)
Referenced by:
    TABLE "destination_rooms" CONSTRAINT "destination_rooms_room_id_fkey" FOREIGN KEY (room_id) REFERENCES rooms(room_id)
Access method: heap

# \d+ current_state_events
                            Table "public.current_state_events"
   Column   | Type | Collation | Nullable | Default | Storage  | Stats target | Description
------------+------+-----------+----------+---------+----------+--------------+-------------
 event_id   | text |           | not null |         | extended |              |
 room_id    | text |           | not null |         | extended |              |
 type       | text |           | not null |         | extended |              |
 state_key  | text |           | not null |         | extended |              |
 membership | text |           |          |         | extended |              |
Indexes:
    "current_state_events_event_id_key" UNIQUE CONSTRAINT, btree (event_id)
    "current_state_events_member_index" btree (state_key) WHERE type = 'm.room.member'::text
    "current_state_events_room_id_type_state_key_key" UNIQUE CONSTRAINT, btree (room_id, type, state_key)
Access method: heap

# \d+ event_json
                                      Table "public.event_json"
      Column       |  Type   | Collation | Nullable | Default | Storage  | Stats target | Description
-------------------+---------+-----------+----------+---------+----------+--------------+-------------
 event_id          | text    |           | not null |         | extended |              |
 room_id           | text    |           | not null |         | extended |              |
 internal_metadata | text    |           | not null |         | extended |              |
 json              | text    |           | not null |         | extended |              |
 format_version    | integer |           |          |         | plain    |              |
Indexes:
    "event_json_event_id_key" UNIQUE CONSTRAINT, btree (event_id)
Access method: heap

# \q
```

#### Background update

 - `register_background_update_handler`

See [`_background_add_rooms_room_version_column`](https://github.com/matrix-org/synapse/blob/e3abc0a5cc0669b76e5bdf2e1539645d790a7a3d/synapse/storage/databases/main/room.py#L1125-L1192) as a reference background update

---

Find where rows are inserted into `rooms`

 - `table="rooms",` (`simple_upsert_txn`)
 - `"rooms",` (`simple_insert`) 

---

```
SYNAPSE_TEST_LOG_LEVEL=INFO python -m twisted.trial tests.storage.databases.main.test_room

SYNAPSE_POSTGRES=1 SYNAPSE_TEST_LOG_LEVEL=INFO python -m twisted.trial tests.storage.databases.main.test_room

SYNAPSE_TEST_LOG_LEVEL=INFO python -m twisted.trial tests.handlers.test_room_summary.SpaceSummaryTestCase.test_fed_invited

SYNAPSE_TEST_LOG_LEVEL=INFO python -m twisted.trial tests.handlers.test_room_summary.SpaceSummaryTestCase.test_fed_filtering
```

### Pull Request Checklist

### DIFF
--- a/synapse/api/constants.py
+++ b/synapse/api/constants.py
@@ -198,6 +198,7 @@ class EventContentFields:
     # cf https://github.com/matrix-org/matrix-doc/pull/1772
     ROOM_TYPE = "type"
 
+    # The creator of the room, as used in `m.room.create` events.
     ROOM_CREATOR = "creator"
 
     # Used on normal messages to indicate they were historically imported after the fact
--- a/synapse/api/constants.py
+++ b/synapse/api/constants.py
@@ -198,6 +198,8 @@ class EventContentFields:
     # cf https://github.com/matrix-org/matrix-doc/pull/1772
     ROOM_TYPE = "type"
 
+    ROOM_CREATOR = "creator"
+
     # Used on normal messages to indicate they were historically imported after the fact
     MSC2716_HISTORICAL = "org.matrix.msc2716.historical"
     # For "insertion" events to indicate what the next chunk ID should be in
--- a/synapse/storage/databases/main/room.py
+++ b/synapse/storage/databases/main/room.py
@@ -19,7 +19,7 @@ from abc import abstractmethod
 from enum import Enum
 from typing import Any, Dict, List, Optional, Tuple
 
-from synapse.api.constants import EventTypes, JoinRules
+from synapse.api.constants import EventTypes, EventContentFields, JoinRules
 from synapse.api.errors import StoreError
 from synapse.api.room_versions import RoomVersion, RoomVersions
 from synapse.events import EventBase
@@ -1223,7 +1223,7 @@ class RoomBackgroundUpdateStore(SQLBaseStore):
             for room_id, event_json in txn:
                 event_dict = db_to_json(event_json)
 
-                creator = event_dict.get("content").get("creator")
+                creator = event_dict.get("content").get(EventContentFields.ROOM_CREATOR)
 
                 self.db_pool.simple_update_txn(
                     txn,
@@ -1439,7 +1439,7 @@ class RoomStore(RoomBackgroundUpdateStore, RoomWorkerStore, SearchStore):
             # invalid, and it would fail auth checks anyway.
             raise StoreError(400, "No create event in state")
 
-        room_creator = create_event.content.get("creator", None)
+        room_creator = create_event.content.get(EventContentFields.ROOM_CREATOR)
 
         if room_creator is None:
             # If the create event does not have a creator then the room is
--- a/synapse/storage/databases/main/room.py
+++ b/synapse/storage/databases/main/room.py
@@ -1370,9 +1370,10 @@ class RoomBackgroundUpdateStore(SQLBaseStore):
             """
 
             txn.execute(sql, (last_room_id, batch_size))
+            room_id_to_create_event_results = txn.fetchall()
 
             new_last_room_id = ""
-            for room_id, event_json in txn:
+            for room_id, event_json in room_id_to_create_event_results:
                 event_dict = db_to_json(event_json)
 
                 creator = event_dict.get("content").get(EventContentFields.ROOM_CREATOR)
--- a/synapse/storage/databases/main/room.py
+++ b/synapse/storage/databases/main/room.py
@@ -1012,9 +1012,9 @@ class RoomWorkerStore(SQLBaseStore):
 class _BackgroundUpdates:
     REMOVE_TOMESTONED_ROOMS_BG_UPDATE = "remove_tombstoned_rooms_from_directory"
     ADD_ROOMS_ROOM_VERSION_COLUMN = "add_rooms_room_version_column"
-    POPULATE_ROOMS_CREATOR_COLUMN = "populate_rooms_creator_column"
     POPULATE_ROOM_DEPTH_MIN_DEPTH2 = "populate_room_depth_min_depth2"
     REPLACE_ROOM_DEPTH_MIN_DEPTH = "replace_room_depth_min_depth"
+    POPULATE_ROOMS_CREATOR_COLUMN = "populate_rooms_creator_column"
 
 
 _REPLACE_ROOM_DEPTH_SQL_COMMANDS = (
@@ -1046,11 +1046,6 @@ class RoomBackgroundUpdateStore(SQLBaseStore):
             self._background_add_rooms_room_version_column,
         )
 
-        self.db_pool.updates.register_background_update_handler(
-            _BackgroundUpdates.POPULATE_ROOMS_CREATOR_COLUMN,
-            self._background_populate_rooms_creator_column,
-        )
-
         # BG updates to change the type of room_depth.min_depth
         self.db_pool.updates.register_background_update_handler(
             _BackgroundUpdates.POPULATE_ROOM_DEPTH_MIN_DEPTH2,
@@ -1059,6 +1054,11 @@ class RoomBackgroundUpdateStore(SQLBaseStore):
         self.db_pool.updates.register_background_update_handler(
             _BackgroundUpdates.REPLACE_ROOM_DEPTH_MIN_DEPTH,
             self._background_replace_room_depth_min_depth,
+        )
+
+        self.db_pool.updates.register_background_update_handler(
+            _BackgroundUpdates.POPULATE_ROOMS_CREATOR_COLUMN,
+            self._background_populate_rooms_creator_column,
         )
 
     async def _background_insert_retention(self, progress, batch_size):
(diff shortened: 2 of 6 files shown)


synapse/storage/databases/main/room.py
```diff
@@ -1191,6 +1198,63 @@ def _background_add_rooms_room_version_column_txn(txn: LoggingTransaction):
 
         return batch_size
 
+    async def _background_populate_rooms_creator_column(
+        self, progress: dict, batch_size: int
+    ):
```
MadLittleMods, line 1203: This is not too dissimilar from `_background_add_rooms_room_version_column` which also sets `creator` 🤔 

Does a background update iterate over every single row? I was expecting it to fetch only the rows missing `creator` and fill them in.

synapse/storage/databases/main/room.py
```diff
@@ -1273,7 +1337,7 @@ async def has_auth_chain_index(self, room_id: str) -> bool:
             keyvalues={"room_id": room_id},
             retcol="MAX(stream_ordering)",
             allow_none=True,
-            desc="upsert_room_on_join",
+            desc="has_auth_chain_index_fallback",
```
MadLittleMods, line 1340: Fixing a copy-paste error (description does not match method)

synapse/storage/databases/main/room.py
```diff
@@ -1361,14 +1427,27 @@ async def upsert_room_on_join(self, room_id: str, room_version: RoomVersion):
         # mark the room as having an auth chain cover index.
         has_auth_chain_index = await self.has_auth_chain_index(room_id)
 
+        create_event = None
+        for e in auth_events:
```
MadLittleMods, line 1431: I originally had this using `self.get_create_event_for_room(room_id)` but changed it to passing in the `auth_events` because the current state ID's weren't persisted yet when this is called in [`do_invite_join`](https://github.com/matrix-org/synapse/blob/ecd823d766fecdd7e1c7163073c097a0084122e2/synapse/handlers/federation.py#L1681-L1693).

We could [call `_persist_auth_tree` before we call `upsert_room_on_join`](https://github.com/matrix-org/synapse/blob/ecd823d766fecdd7e1c7163073c097a0084122e2/synapse/handlers/federation.py#L1690-L1697) to make them available AFAICT 🤷  but that order may be more fragile than I am assuming.

synapse/storage/databases/main/room.py
```diff
@@ -1361,14 +1427,27 @@ async def upsert_room_on_join(self, room_id: str, room_version: RoomVersion):
...
+
         await self.db_pool.simple_upsert(
             desc="upsert_room_on_join",
             table="rooms",
             keyvalues={"room_id": room_id},
             values={"room_version": room_version.identifier},
             insertion_values={
                 "is_public": False,
-                "creator": "",
+                "creator": room_creator,
```
MadLittleMods, line 1450: Instance #1 where we insert into the `rooms` table and need to make sure `creator` is also populated

synapse/storage/databases/main/room.py
```diff
@@ -1388,6 +1467,9 @@ async def maybe_store_room_on_outlier_membership(
         # mark the room as having an auth chain cover index.
         has_auth_chain_index = await self.has_auth_chain_index(room_id)
 
+        create_event = await self.get_create_event_for_room(room_id)
+        room_creator = create_event.content.get("creator", None)
```
MadLittleMods, line 1471: I'm not so sure about this case being totally right. The MSC2716 stuff (Complement test suite) doesn't touch this code at all to be sure. And there are many usages of this function to check out.

I don't know if all of the usages of `maybe_store_room_on_outlier_membership`  have room state event ID's in place when we call it to rely on `get_create_event_for_room(room_id)`

For example with the `upsert_room_on_join` case above, I had to change it to pass in the `auth_events` because no state events were persisted at that time to use `get_create_event_for_room` which derives from them. But `upsert_room_on_join` also only has one usage so it's easy to reason about and refactor.

synapse/storage/databases/main/room.py
```diff
@@ -1396,7 +1478,7 @@ async def maybe_store_room_on_outlier_membership(
             insertion_values={
                 "room_version": room_version.identifier,
                 "is_public": False,
-                "creator": "",
+                "creator": room_creator,
```
MadLittleMods, line 1481: Instance #2 where we insert into the `rooms` table and need to make sure `creator` is also populated

synapse/storage/databases/main/room.py
MadLittleMods: > **I think what this really comes down to is when a user gets a federated invitation to the room, does the homeserver start receiving messages in the room before the user joins (while in the invited state)?** Our goal is to process MSC2716 related events consistently across homeservers which are restricted to the room creator.

@richvdh Any insight here?

synapse/storage/databases/main/room.py
```diff
@@ -1396,6 +1480,9 @@ async def maybe_store_room_on_outlier_membership(
             insertion_values={
                 "room_version": room_version.identifier,
                 "is_public": False,
+                # We don't worry about setting the `creator` here because
+                # we don't process any messages in a room while a user is
+                # invited (only after the join).
                 "creator": "",
```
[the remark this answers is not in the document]
MadLittleMods: > Do you need to worry about it? Depends what you want to use the field for, I guess :)

I think we do need to consider it. I've asked for specific clarification on [what I don't know there](https://github.com/matrix-org/synapse/pull/10697#discussion_r698924269).

---

> > Should this be `None` over `""`?
> 
> :man_shrugging: Maybe, but I'm not sure I'd bother changing it.

Leaving it as-is ⏩ Just need to consider it could be an empty string in the queries

synapse/storage/databases/main/room.py
```diff
@@ -1011,6 +1012,7 @@ def get_rooms_for_retention_period_in_range_txn(txn):
 class _BackgroundUpdates:
     REMOVE_TOMESTONED_ROOMS_BG_UPDATE = "remove_tombstoned_rooms_from_directory"
     ADD_ROOMS_ROOM_VERSION_COLUMN = "add_rooms_room_version_column"
+    POPULATE_ROOMS_CREATOR_COLUMN = "populate_rooms_creator_column"
```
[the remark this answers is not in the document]
MadLittleMods: Moved to end. Does every new background update have to be added to the end of the list?

synapse/storage/schema/main/delta/63/02populate-rooms-creator.sql
```diff
@@ -0,0 +1,17 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+INSERT into background_updates (update_name, progress_json)
+    VALUES ('populate_rooms_creator_column', '{}');
```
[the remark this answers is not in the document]
MadLittleMods: I can't find an example that uses the `order` sequence number. What should the number be?

tests/storage/databases/main/test_room.py
```diff
@@ -0,0 +1,98 @@
...
+        # Make sure the test is starting out with a room without a creator
+        room_creator_before = self.get_success(
+            self.store.db_pool.simple_select_one_onecol(
+                table="rooms",
+                keyvalues={"room_id": room_id},
+                retcol="creator",
+                allow_none=True,
+            )
+        )
+        self.assertEqual(room_creator_before, None)
```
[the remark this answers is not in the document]
MadLittleMods: I feel like the way we create a room could change in the future and could easily get into a situation where we are testing nothing because it's already in the correct state.

synapse/storage/databases/main/room.py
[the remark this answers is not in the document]
richvdh: well, that would seem logical?

synapse/storage/schema/main/delta/63/02populate-rooms-creator.sql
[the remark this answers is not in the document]
richvdh: See for example https://github.com/matrix-org/synapse/blob/develop/synapse/storage/schema/main/delta/59/10delete_purged_chain_cover.sql#L16: the column in question is called `ordering`. We normally just set it to the numbers of the schema version and the delta file - `6302` in this case.

synapse/storage/databases/main/room.py
richvdh: > I think what this really comes down to is when a user gets a federated invitation to the room, does the homeserver start receiving messages in the room before the user joins (while in the invited state)? 

No, the homeserver should not receive any regular events in a room until the user has properly joined. You can see this at https://github.com/matrix-org/synapse/blob/develop/synapse/handlers/federation_event.py#L223 - any incoming events for rooms we are not in get dropped. The exception to that is "out-of-band" membership events - such as further invite events.

So, yes, I think you're safe not knowing about the creator while you're in the "invited" state.

synapse/api/constants.py
```diff
@@ -198,6 +198,8 @@ class EventContentFields:
     # cf https://github.com/matrix-org/matrix-doc/pull/1772
     ROOM_TYPE = "type"
 
+    ROOM_CREATOR = "creator"
```
richvdh, line 201: It might be obvious, but some comment as to where this is used would be good:

```suggestion
    # The creator of the room, as used in `m.room.create` events.
    ROOM_CREATOR = "creator"
```

synapse/storage/databases/main/room.py
MadLittleMods: Thanks for for the info and source 😀 glad we don't have to go down that rabbit hole of hooking everything up there.

synapse/storage/databases/main/room.py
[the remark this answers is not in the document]
MadLittleMods: Doesn't seem strictly necessary 🤔 but whatever the convention is works ⏩ - I could see this background update working in whatever position it runs at.

MadLittleMods:
Thanks for all of the review so far @richvdh 🦝

richvdh (approved):
lgtm!

synapse/storage/databases/main/room.py
```diff
@@ -1370,9 +1370,10 @@ def _background_populate_rooms_creator_column_txn(txn: LoggingTransaction):
             """
 
             txn.execute(sql, (last_room_id, batch_size))
+            room_id_to_create_event_results = txn.fetchall()
 
             new_last_room_id = ""
-            for room_id, event_json in txn:
```
MadLittleMods: Maybe I was also running into something similar to https://github.com/matrix-org/synapse/pull/10734

> We were originally iterating over `txn`, which can work, but as `txn` was being used to run a
>
> https://github.com/matrix-org/synapse/pull/10734

And in this case, we're iterating over `txn` and doing a `simple_update_txn` 🤔

PR #10705, opened 2021-08-26
Directly start Synapse processes instead of using supervisor for Docker image with workers

### LINKED ISSUE #10065: Synapse workers Complement image results in flaky tests due to inconsistent worker process init

clokep:
This attempts to fix #10065.

This takes the suggestion in there to manually start the Synapse processes and wait for it to be ready via the `SDNOTIFY` stuff (instead of depending on supervisor to start the processes).

Unfortunately at the end of this (while using complement) I seem to end up at the same result as when using supervisor. Complement intermittently cannot get a `/versions` response.

I'm also unsure if this makes sense if we're asking people to use this image *outside of complement* since this has no process monitoring.

The core of the change is in the second commit, but I'm not really sure where to go from here. 😢 

I was testing this with: `COMPLEMENT_DIR=<...> WORKERS=1 ./scripts-dev/complement.sh TestVersionStructure` from the synapse root


clokep:
I'm going to close this, but if someone wants to pick this open do feel free to reuse this branch / PR!

PR #10707, opened 2021-08-26
Additional type hints for client REST servlets (part 3)

clokep:
Follow-up to #10665/#10674 with more modules.

~~Built-on #10706.~~

### DIFF
--- a/synapse/rest/client/register.py
+++ b/synapse/rest/client/register.py
@@ -14,7 +14,9 @@
 # limitations under the License.
 import logging
 import random
-from typing import List, Union
+from typing import TYPE_CHECKING, List, Optional, Tuple
+
+from twisted.web.server import Request
 
 import synapse
 import synapse.api.auth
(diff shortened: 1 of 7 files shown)


synapse/rest/client/register.py
```diff
@@ -445,23 +429,21 @@ def __init__(self, hs):
...
 
-        if kind == b"guest":
+        if kind == "guest":
             ret = await self._do_guest_registration(body, address=client_addr)
             return ret
-        elif kind != b"user":
+        elif kind != "user":
             raise UnrecognizedRequestError(
-                "Do not understand membership kind: %s" % (kind.decode("utf8"),)
+                "Do not understand membership kind: %s" % (kind,)
```
erikjohnston, line 446:
```suggestion
                f"Do not understand membership kind: {kind}",
```

while we're here

synapse/rest/client/room.py
```diff
@@ -929,48 +1004,58 @@ async def on_POST(self, request):
...
-    if with_get:
+    on_POST = getattr(servlet, "on_POST", None)
+    if on_POST:
+        http_server.register_paths(
+            "POST",
+            client_patterns(regex_string + "$", v1=True),
+            on_POST,
+            servlet.__class__.__name__,
+        )
+    on_PUT = getattr(servlet, "on_PUT", None)
```
erikjohnston, line 1045: Do we want to allow servlet classes to not have a `PUT` here? This feels more like we should be asserting here instead (which will cause Synapse not to start)

erikjohnston:
LGTM otherwise

synapse/rest/client/room.py
clokep: That's fine. I figured we can go either way. Annoyingly an mypy doesn't seem to work great with `hasattr`. 😢 I'll take another look!

erikjohnston (approved):

Issue #7968, opened 2020-07-28
master blocks for a long time during startup


richvdh:
Why does it take 70 seconds to initialise the database? It might be some sort of database connection contention but it would be nice if there were some log messages or something.

```
2020-07-28 09:56:32,141 - synapse.storage.data_stores - 50 - INFO - None - Preparing database 'master'...
2020-07-28 09:56:56,412 - synapse.storage.prepare_database - 325 - INFO - None - Upgrading schema to v58
2020-07-28 09:56:56,421 - synapse.storage.data_stores - 60 - INFO - None - Starting 'main' data store
2020-07-28 09:57:42,940 - synapse.config.appservice - 87 - INFO - None - Loaded application service: ApplicationService: ...
```

what's going on there?

richvdh:
keywords: restart

richvdh:
A lot of the problem here is this query:

```
SELECT instance_name, MAX(stream_id) FROM cache_invalidation_stream_by_instance GROUP BY instance_name
```

... which takes six seconds on a relatively idle database.

richvdh:
this one takes 11 seconds:

```
SELECT COUNT(*) FROM users WHERE name NOT LIKE '%:matrix.org'
```

richvdh:
hopefully the latter will be fixed by #8271.

richvdh:
I've also manually cleared out `cache_invalidation_stream_by_instance`, which seems to have mostly resolved this problem, so I'm going to close this and leave #8269 to track a more permanent fix.

richvdh:
it's still pretty slow, so I'm not sure it's fixed.

richvdh:
shutting down also takes ages: that seems to be mostly because python runs a final GC at shutdown, which can take many seconds. Maybe we can skip that GC. (or maybe it improves if we fix https://github.com/matrix-org/synapse/issues/7176)

richvdh:
> (or maybe it improves if we fix #7176)

given we're running zope.interface-4.7.1 on matrix.org, that's sadly unlikely.

richvdh:
> > (or maybe it improves if we fix #7176)
> 
> given we're running zope.interface-4.7.1 on matrix.org, that's sadly unlikely.

Per https://instagram-engineering.com/dismissing-python-garbage-collection-at-instagram-4dca40b29172, we could skip the final GC with an early `atexit.register(os._exit, 0)`.

richvdh:
[...]

PR #10712, opened 2021-08-27
Skip the final GC on shutdown to improve restart times

squahtx:
As suggested by @richvdh in #7968 

Partially addresses #7968

~~Installs an `atexit` handler that directly calls `os._exit`:~~
* ~~From testing, all the reactor shutdown hooks get run before any `atexit` handlers.~~
  ```
  2021-08-27 12:07:28,476 - synapse.storage.databases.main.lock - 79 - INFO - LockStore._on_shutdown-0 - Dropping held locks due to shutdown
  2021-08-27 12:07:28,476 - synapse.storage.databases.main.lock - 88 - INFO - LockStore._on_shutdown-0 - Dropped locks due to shutdown
  2021-08-27 12:07:28,477 - synapse.handlers.presence - 737 - INFO - presence.on_shutdown-0 - Performing _on_shutdown. Persisting 151 unpersisted changes
  2021-08-27 12:07:28,477 - synapse.app._base - 495 - INFO - sentinel - Shutting down...
  2021-08-27 12:07:28,490 - synapse.handlers.presence - 750 - INFO - presence.on_shutdown-0 - Finished _on_shutdown
  2021-08-27 12:07:28,491 - synapse.http.site - 325 - INFO - GET-1570 - Connection from client lost before response was sent
  2021-08-27 12:07:28,492 - twisted - 271 - INFO - sentinel - (TCP Port 8008 Closed)
  2021-08-27 12:07:28,492 - twisted - 271 - INFO - sentinel - (TCP Port 8008 Closed)
  2021-08-27 12:07:28,493 - twisted - 271 - INFO - sentinel - (TCP Port 8008 Closed)
  2021-08-27 12:07:28,494 - twisted - 271 - INFO - sentinel - Main loop terminated.
  2021-08-27 12:07:28,495 - synapse.app._skip_final_gc - 38 - INFO - sentinel - Skipping final GC and further exit handlers...
  ```
* ~~The `logging` module registers an `atexit` handler to flush buffers, which we call manually~~
* ~~There's no nice, sensible way to determine the right exit code to use in the `atexit` handler, so we allow shutdown to run normally when we detect an unclean shutdown.~~


richvdh:
I was interested to see if `gc.freeze()` was `O(n)` in the number of allocated objects. Seems like no, it's `O(1)`: it uses [`gc_list_merge`](https://github.com/python/cpython/blob/v3.7.11/Modules/gcmodule.c#L187) (https://github.com/python/cpython/blob/v3.7.11/Modules/gcmodule.c#L1428) which just does some pointer manipulation on head and tail of the linked lists.

Issue #10694, opened 2021-08-25
DNS problems in IPv6-only environment


xfk:
### Description

My redis instance has only an IPv6 address. The matrix-synapse container is throwing up a DNS lookup error:
`
2021-08-24 13:30:55,585 - twisted - 258 - ERROR - sentinel- Message: 'Connection to redis server %s failed: %s'
2021-08-24 13:30:55,585 - twisted - 258 - ERROR - sentinel- Arguments: ("b'matrix-test-redis-master':6379", DNSLookupError(b'matrix-test-redis-master'))
`
It also doesn't work with a FQDN IPv6-only address.

I did a tcpdump of DNS traffic in the container, it's only searching for A records (not AAAA).

- **Platform**:

It's running in a IPv6-only Kubernetes cluster.

glyph:
Anywhere you need to be looking up A or AAAA, you want to be using [`HostnameEndpoint`](https://twistedmatrix.com/documents/20.3.0/api/twisted.internet.endpoints.HostnameEndpoint.html). There's a rules-lawyer-y answer to the `IResolverSimple.getHostByName` dilemma, but it's going to be long and drawn out and complicated and at the end of the day, this API is fundamentally too narrow for the use-case of actually connecting to a TCP endpoint.

xfk:
@glyph I tried it out and everything seems to be working.

https://github.com/xfk/synapse/commit/4d7037a26f6ef3b34a5a41a357844583f913b88c

clokep:
@xfk Would you mind making a PR? Seems like it is overall the right fix.

xfk:
@clokep I created a PR.

https://github.com/matrix-org/synapse/pull/10717

a-0-dev:
I just ran into this exact problem, are there any issues holding the PR up? I could work around it by querying the IPv6 at install time, but that is very hacky... :/

richvdh:
@a-0-dev: I don't think anyone is working on this currently.

clokep:
Note that #7720 is now fixed, we can likely apply the same fixes for Redis.

Issue #6895, opened 2020-02-12
Bogus DNS data leads to poor performance


evilham:
### Preamble
This is similar to #5373 and I *think* it  would get very bad if #3765 is not "properly dealt with" and lingering homeservers start being widespread (with e.g. "this domain is for sale" wildcard DNS that *drop* connections on 8448 or something), or if network splits start happening for whatever reason.
But it is particular enough to be worthy of explaining in its own issue, since this instance may be mitigated somewhat easily.

### Description
There is a high number of Matrix/Synapse servers with bogus DNS data. By bogus DNS data in this case, I mean something isomorphic to(*):
Matrix domain: example.org
Delegating to: https://matrix.example.org:443
Resolving matrix.example.org as:
- A record for an IPv4
- AAAA record for an IPv6

Where actually that IPv6 is not listening on port 443, or the port is being blocked by the firewall or they have networking issues, .... (insert plethora of reasons to get IPv6 wrong here).

(*): As a note, this is not exclusive to IPv6, I have recently worked with CISCO's top 1M websites dataset and checked their DNS and have found... interesting things like A records in the `127.0.0.0/8` subnet or AAAA records with `::1` or `fe80::`.
Another way to reach this could be with a network split (e.g. country-wide bans to certain servers, ...)

With this in mind, I have come across an issue where:
A sufficient number of unreachable matrix servers will degrade performance critically. I believe this is because the server will see events from those "unreachable" servers through other servers sharing a room and start trying to reach out to them (fetch events, send events, ...).

### Steps to reproduce

- Setup an synapse in an IPv6-only network (yes, it's 2020!)
- Use DNS64 (**), so that the v4 world is reachable form synapse
- Use a dual-stack server to proxy incoming v4 connections to synapse
- Join e.g. #synapse-admins:matrix.org (modulo typos)
- Check the logs and see increasingly federationclient messages about timeouts and unreachable networks
- Wait for a few hours and see performance degrade significantly (sending of events --> 10-15 seconds)
- Leave the afore mentioned room and see performance be restored over a few minutes.

This by the way describes real-life set ups, maybe not numerous, but I know different people that have reached a similar setup independently.

(**): DNS64 works just like regular DNS, except, it adds AAAA records for sites that only have A records. This works by embedding the IPv4 addresses in the remaining bits of a `/64` subnet, then doing stateful NAT-ing.
[...]

PR #10717, opened 2021-08-30
 Fix resolution of domains with only AAAA records (#10694)

xfk:


clokep:
@xfk Could you please [sign-off](https://github.com/matrix-org/synapse/blob/develop/CONTRIBUTING.md#sign-off) on your contribution? You can just add it as a comment on the PR. 👍

xfk:
DCO (Developer Certificate of Origin: http://developercertificate.org/)
Signed-off-by: Fran Kristo <fran.kristo42@gmail.com>

ShadowJonathan:
Could you maybe add the following to the PR description? Then the issues will be automatically closed;
```
Fixes #10694
Fixes #7720
```

ShadowJonathan:
I wonder if this also addresses https://github.com/matrix-org/synapse/issues/6895

clokep:
The tests are failing due to this change, I think something like the following will fix it:

```diff
diff --git a/tests/handlers/test_send_email.py b/tests/handlers/test_send_email.py
index 6f77b1237..10a1c4a5f 100644
--- a/tests/handlers/test_send_email.py
+++ b/tests/handlers/test_send_email.py
@@ -71,6 +71,8 @@ class _DummyMessage:
 class SendEmailHandlerTestCase(HomeserverTestCase):
     def test_send_email(self):
         """Happy-path test that we can send email to a non-TLS server."""
+        self.reactor.lookups["localhost"] = "1.2.3.4"
+
         h = self.hs.get_send_email_handler()
         d = ensureDeferred(
             h.send_email(
@@ -82,7 +84,7 @@ class SendEmailHandlerTestCase(HomeserverTestCase):
         (host, port, client_factory, _timeout, _bindAddress) = self.reactor.tcpClients[
             0
         ]
-        self.assertEqual(host, "localhost")
+        self.assertEqual(host, "1.2.3.4")
         self.assertEqual(port, 25)
 
         # wire it up to an SMTP server
```

I also think we need to `await` each connect call since that returns a `Deferred`?

synapse/replication/tcp/handler.py
```diff
@@ -314,17 +315,19 @@ def start_replication(self, hs):
...
             )
+            endpoint.connect(self._factory)
         else:
             client_name = hs.get_instance_name()
             self._factory = DirectTcpReplicationClientFactory(hs, client_name, self)
             host = hs.config.worker_replication_host
             port = hs.config.worker_replication_port
-            hs.get_reactor().connectTCP(host.encode(), port, self._factory)
+            endpoint = HostnameEndpoint(hs.get_reactor(), host.encode(), port)
+            endpoint.connect(self._factory)
```
richvdh, line 330: Unfortunately, I think fixing this is more involved. It's to do with how we handle connection failures, and retries.

`DirectTcpReplicationClientFactory` and `RedisDirectTcpReplicationClientFactory` are both based on `twisted.internet.protocol.ReconnectingClientFactory`. If `reactor.connectTCP` is unable to connect to the hostname, it will call `ReconnectingClientFactory.clientConnectionFailed`, which will schedule a retry a few seconds later.

`HostnameEndpoint.connect` does its own set of retries - one for each IP address returned by the hostname lookup - and then calls `reactor.connectTCP` for each.

So, if we use `HostnameEndpoint` *and* `ReconnectingClientFactory`, we can end up with two different, conflicting, retry systems going on at the same time, leaving us in a terrible mess.

I gather (h/t @glyph) that `ReconnectingClientFactory` is softly deprecated, and instead we want to use https://twistedmatrix.com/documents/current/api/twisted.application.internet.ClientService.html, which will compose properly with a `HostnameEndpoint`.

So, I think what we want to do is create a `ClientService` in the constructor of `ReplicationCommandHandler`, and then we can just call `ClientService.startService` in `start_replication`.

---

Now, all of this is getting rather complicated. I think you should split up this PR - let's just fix the SMTP sender for now, and then consider fixing the replication connections in a later PR?

synapse/replication/tcp/redis.py
```diff
@@ -364,6 +365,9 @@ def lazyConnection(
     factory.continueTrying = reconnect
 
     reactor = hs.get_reactor()
-    reactor.connectTCP(host.encode(), port, factory, timeout=30, bindAddress=None)
+    endpoint = HostnameEndpoint(
+        reactor, host.encode(), port, timeout=30, bindAddress=None
+    )
+    endpoint.connect(factory)
```
richvdh, line 371: this has the same problem as `ReplicationCommandHandler`.

synapse/handlers/send_email.py
```diff
@@ -106,7 +107,10 @@ def build_sender_factory(**kwargs) -> ESMTPSenderFactory:
         factory = build_sender_factory(hostname=smtphost if enable_tls else None)
 
     # the IReactorTCP interface claims host has to be a bytes, which seems to be wrong
-    reactor.connectTCP(smtphost, smtpport, factory, timeout=30, bindAddress=None)  # type: ignore[arg-type]
+    endpoint = HostnameEndpoint(
+        reactor, smtphost, smtpport, timeout=30, bindAddress=None
+    )
+    endpoint.connect(factory)
```
richvdh, line 113: As @clokep said, I think we need to `await` this. In fact, for logcontext-related reasons too annoying to explain here, we need to:

```suggestion
    await make_deferred_yieldable(endpoint.connect(factory))
```

richvdh (changes requested):
Thank you for the contribution! Some comments below though.

anoadragon453:
Hello @xfk, just checking in to see whether you're around to tackle the requested changes here?

xfk:
@anoadragon453

I made some progress, but didn't get a chance to finish.  They are a bit more involved than anticipated.

anoadragon453:
> @anoadragon453
> 
> I made some progress, but didn't get a chance to finish. They are a bit more involved than anticipated.

No worries. Feel free to ask questions if you hit a roadblock. We're happy to help!

telmich:
ping - what's the best way to move forward with this one?

anoadragon453:
Hello @xfk! Are you around to continue working on this PR?

If fixing the full problem is too daunting for now, you may want to split up the PR into more manageable chunks, landing what's figured out so far for now, as suggested in https://github.com/matrix-org/synapse/pull/10717/files#r699626644?

richvdh:
Sadly it looks like @xfk no longer has time to work on this, so I'm going to close it.

Issue #10718, opened 2021-08-30
deal with an event that is in event_json but not events


JanZerebecki:
### Description

When an event comes in that is not in events, when trying to persist it in event_json a possible duplicate should be handled. Currently https://github.com/matrix-org/synapse/blob/f03cafb50c49a1569f1f99485f9cc42abfdc7b21/synapse/storage/databases/main/events.py#L1337-L1339 fails (see exception below). I do not have enough domain knowledge to say whether to leave the existing entry unchanged or to update the existing entry. I assume the better choice is to replace it as then the data comes from the same input. However that might be a vector for maliciously overwriting data and breaking history integrity.

The same issue was also seen in https://github.com/matrix-org/synapse/issues/8889 but not further discussed (besides the manually deleting the event from event_json).

```
Aug 27 11:43:51 matrix synapse[27658]: synapse.handlers.federation: [_process_incoming_pdus_in_room_inner-13-$REDACTED1] Error attempting to resolve state at missing prev_events
                                       Traceback (most recent call last):
                                         File "/usr/lib/python3.6/site-packages/synapse/handlers/federation.py", line 387, in on_receive_pdu
                                           origin, room_id, p
                                         File "/usr/lib/python3.6/site-packages/synapse/handlers/federation.py", line 732, in _get_state_after_missing_prev_event
                                           destination=destination, room_id=room_id, events=missing_events
                                         File "/usr/lib/python3.6/site-packages/synapse/handlers/federation.py", line 1379, in _get_events_and_persist
                                           event_infos,
                                         File "/usr/lib/python3.6/site-packages/synapse/handlers/federation.py", line 2323, in _auth_and_persist_events
                                           backfilled=backfilled,
                                         File "/usr/lib/python3.6/site-packages/synapse/handlers/federation.py", line 3320, in persist_events_and_notify
                                           event_and_contexts, backfilled=backfilled
                                       File "/usr/lib/python3.6/site-packages/synapse/storage/persist_events.py", line 322, in persist_events
                                           ret_vals = await yieldable_gather_results(enqueue, partitioned.items())
                                         File "/usr/lib/python3.6/site-packages/synapse/storage/persist_events.py", line 240, in handle_queue_loop
                                           item.events_and_contexts, item.backfilled
                                         File "/usr/lib/python3.6/site-packages/synapse/storage/persist_events.py", line 582, in _persist_event_batch
                                           backfilled=backfilled,
[...]

PR #10719, opened 2021-08-30
Fix instert of duplicate key into event_json

JanZerebecki:
When an incoming event id is present in event_json but not in events
synapse fails trying to insert it with "psycopg2.errors.UniqueViolation:
duplicate key value violates unique constraints", because it is only
filtered based on those that are in events.

I don't know why those become out of sync, but this happening was
reported by others before.

Fix this by using an upsert (which inserts or updates existing records)
instead of a normal insert.

Please verify that this is the safe and correct thing to do before
merging this. Verify e.g. that it doesn't allow breaking history
integrity or something like it. As I don't know enough to understand
what this change entails.

Fixes: https://github.com/matrix-org/synapse/issues/10718
Signed-off-by: Jan Zerebecki <jan.suse@zerebecki.de>

### Pull Request Checklist


richvdh (changes requested):
per the discussion on #10718, we're not (yet) convince that this is a change that we should be making.

PR #10727, opened 2021-08-31
Ignore rooms with unknown room versions in the spaces summary.

### LINKED ISSUE #10724: A room with an unknown room version breaks the spaces summary endpoint
From @BillCarsonFr:

> it looks like I have a space that contains a room with a deleted room version, and in this case the pagination breaks
>
> not sure what would be expected
>
> image

clokep:
Fixes #10724

This is the same strategy we used for `/sync` where we just ignore the room and do not return it.

### DIFF
--- a/synapse/handlers/room_summary.py
+++ b/synapse/handlers/room_summary.py
@@ -28,7 +28,14 @@ from synapse.api.constants import (
     Membership,
     RoomTypes,
 )
-from synapse.api.errors import AuthError, Codes, NotFoundError, StoreError, SynapseError
+from synapse.api.errors import (
+    AuthError,
+    Codes,
+    NotFoundError,
+    StoreError,
+    SynapseError,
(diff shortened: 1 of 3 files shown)


synapse/handlers/room_summary.py
```diff
@@ -814,7 +821,10 @@ async def _is_local_room_accessible(
             logger.info("room %s is unknown, omitting from summary", room_id)
             return False
 
-        room_version = await self._store.get_room_version(room_id)
+        try:
+            room_version = await self._store.get_room_version(room_id)
+        except UnsupportedRoomVersionError:
+            return False
```
erikjohnston, line 827: Can you add a bit of a comment as to why this is OK, ie "this stops us breaking the summary in the case where we drop room version support", or something?

erikjohnston (approved):

Issue #10729, opened 2021-08-31
remove_deleted_email_pushers background job fails with psycopg2.ProgrammingError when unlinked addresses exist


anoadragon453:
We found the following traceback when testing develop this morning. This is related to #10581:

<details>
<summary>Synapse traceback logs</summary>

```
2021-08-31 13:39:38,502 - synapse.storage.background_updates - 117 - ERROR - background_updates-0 - Error doing update
Capture point (most recent call last):
  File "/usr/local/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/local/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/synapse/src/synapse/app/homeserver.py", line 462, in <module>
    main()
  File "/home/synapse/src/synapse/app/homeserver.py", line 458, in main
    run(hs)
  File "/home/synapse/src/synapse/app/homeserver.py", line 444, in run
    logger=logger,
  File "/home/synapse/src/synapse/app/_base.py", line 145, in start_reactor
    run()
  File "/home/synapse/src/synapse/app/_base.py", line 129, in run
    run_command()
  File "/home/synapse/env-py37/lib/python3.7/site-packages/twisted/internet/base.py", line 1318, in run
    self.mainLoop()
  File "/home/synapse/env-py37/lib/python3.7/site-packages/twisted/internet/base.py", line 1328, in mainLoop
    reactorBaseSelf.runUntilCurrent()
  File "/home/synapse/src/synapse/metrics/__init__.py", line 565, in f
    ret = func(*args, **kwargs)
  File "/home/synapse/env-py37/lib/python3.7/site-packages/twisted/internet/base.py", line 967, in runUntilCurrent
    f(*a, **kw)
  File "/home/synapse/env-py37/lib/python3.7/site-packages/twisted/internet/defer.py", line 701, in errback
    self._startRunCallbacks(fail)
  File "/home/synapse/env-py37/lib/python3.7/site-packages/twisted/internet/defer.py", line 764, in _startRunCallbacks
    self._runCallbacks()
  File "/home/synapse/env-py37/lib/python3.7/site-packages/twisted/internet/defer.py", line 859, in _runCallbacks
    current.result, *args, **kwargs
  File "/home/synapse/env-py37/lib/python3.7/site-packages/twisted/internet/defer.py", line 1751, in gotResult
[...]

PR #10734, opened 2021-09-01
Fix iteration in _remove_deleted_email_pushers background job.

anoadragon453:
Fixes https://github.com/matrix-org/synapse/issues/10729, see issue for context.

We were originally iterating over `txn`, which can work, but as `txn` was being used to run a

Thus, this caused the background job that was intended to remove old pushers that were
associated with an unlinked email address to fail if there were any to remove.We were originally iterating over 'txn', which can work, but as 'txn' was being used to run a

Thus, this caused the background job that was intended to remove old pushers that were
associated with an unlinked email address to fail if there were any to remove.

### DIFF
--- a/tests/push/test_email.py
+++ b/tests/push/test_email.py
@@ -67,7 +67,7 @@ class EmailPusherTests(HomeserverTestCase):
         config["public_baseurl"] = "aaa"
         config["start_pushers"] = True
 
-        hs = self.setup_test_homeserver(config=config)
+        self.hs = self.setup_test_homeserver(config=config)
 
         # List[Tuple[Deferred, args, kwargs]]
         self.email_attempts = []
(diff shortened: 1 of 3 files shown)


tests/push/test_email.py
```diff
@@ -67,7 +67,7 @@ def make_homeserver(self, reactor, clock):
         config["public_baseurl"] = "aaa"
         config["start_pushers"] = True
 
-        hs = self.setup_test_homeserver(config=config)
+        self.hs = self.setup_test_homeserver(config=config)
```
richvdh, line 70: `setUp` (the thing that calls `make_homeserver` sets `self.hs`, so this is unnecessary.

tests/push/test_email.py
```diff
@@ -344,6 +344,50 @@ def test_no_email_sent_after_removed(self):
...
+        self.hs.get_datastore().db_pool.updates._all_done = False
+
+        # Now let's actually drive the updates to completion
+        while not self.get_success(
+            self.hs.get_datastore().db_pool.updates.has_completed_background_updates()
+        ):
+            self.get_success(
+                self.hs.get_datastore().db_pool.updates.do_next_background_update(100),
+                by=0.1,
+            )
```
richvdh, line 382: We should probably extract all this to a utility function. Maybe next time.

richvdh (commented):
needs a newsfile. looks sensible otherwise.

tests/push/test_email.py
anoadragon453: Aha, indeed, good point!

richvdh (approved):
lgtm

PR #10736, opened 2021-09-01
Additional type hints for client REST servlets (part 5)

clokep:
Follow-up to #10665/#10674/#10707/#10728 to finish of `synapse.rest.client` and enforce that all method signatures need type hints.

This is mostly the same, except the initial PR converts a freeform dictionary to attrs. Some of this is also a bit more abstract (and touches the decorators). I didn't think we needed to go crazy with those type hints, but we can try to make them stricter if people like.

### DIFF
--- a/synapse/http/servlet.py
+++ b/synapse/http/servlet.py
@@ -579,14 +579,14 @@ def parse_json_value_from_request(request: Request) -> JsonDict:
 
 @overload
 def parse_json_value_from_request(
-    request: Request, *, allow_empty_body: Literal[True]
-) -> Optional[JsonDict]:
+    request: Request, allow_empty_body: Literal[False]
+) -> JsonDict:
     ...
 
 
 @overload
 def parse_json_value_from_request(
-    request: Request, *, allow_empty_body: bool = False
+    request: Request, allow_empty_body: bool = False
 ) -> Optional[JsonDict]:
     ...
 
--- a/synapse/http/servlet.py
+++ b/synapse/http/servlet.py
@@ -572,6 +572,25 @@ def parse_string_from_args(
     return strings[0]
 
 
+@overload
+def parse_json_value_from_request(request: Request) -> JsonDict:
+    ...
+
+
+@overload
+def parse_json_value_from_request(
+    request: Request, *, allow_empty_body: Literal[True]
+) -> Optional[JsonDict]:
+    ...
+
+
+@overload
+def parse_json_value_from_request(
+    request: Request, *, allow_empty_body: bool = False
+) -> Optional[JsonDict]:
+    ...
+
+
 def parse_json_value_from_request(
     request: Request, allow_empty_body: bool = False
 ) -> Optional[JsonDict]:
--- a/synapse/rest/client/_base.py
+++ b/synapse/rest/client/_base.py
@@ -16,7 +16,7 @@
 """
 import logging
 import re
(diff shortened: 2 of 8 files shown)


synapse/http/servlet.py
```diff
@@ -572,6 +572,25 @@ def parse_string_from_args(
...
+def parse_json_value_from_request(
+    request: Request, *, allow_empty_body: Literal[True]
+) -> Optional[JsonDict]:
+    ...
+
+
+@overload
+def parse_json_value_from_request(
+    request: Request, *, allow_empty_body: bool = False
+) -> Optional[JsonDict]:
```
erikjohnston, line 590: Shouldn't this be 

```suggestion
) -> JsonDict:
```

?

synapse/rest/client/_base.py
```diff
@@ -76,7 +76,9 @@ def set_timeline_upper_limit(filter_json: JsonDict, filter_timeline_limit: int)
         )
 
 
-def interactive_auth_handler(orig):
+def interactive_auth_handler(
+    orig: Callable[..., Awaitable[Tuple[int, JsonDict]]]
+) -> Callable[..., Awaitable[Tuple[int, JsonDict]]]:
```
erikjohnston, line 81: Since we're here we may as well factor out a `C = TypeVar("C", bound=Callable[..., Awaitable[Tuple[int, JsonDict]]])`? Otherwise using `@interactive_auth_handler` will strip out a bunch of the typing?

clokep: Maybe. I haven't been able to figure out what using the `bound` keyword actually does for typevars, but I'll take another look.

clokep: I tried this (and changed `def interactive_auth_handler(orgi: C) -> C`), but it then gave an error about the return value not equaling the expected value.

erikjohnston: > Maybe. I haven't been able to figure out what using the `bound` keyword actually does for typevars, but I'll take another look.

The bound basically means that you can only use the `@interactive_auth_handler` on things that are `Callable[..., Awaitable[Tuple[int, JsonDict]]]`



> I tried this (and changed `def interactive_auth_handler(orgi: C) -> C`), but it then gave an error about the return value not equaling the expected value.

Yeah, annoyingly you need to add a `cast(C, warpped)` when returning, an example of this can be found at e.g. https://github.com/matrix-org/synapse/blob/e2481dbe9325321d460037a2efe9b9ea2ac78057/synapse/util/caches/lrucache.py#L407-L413

synapse/http/servlet.py
erikjohnston: The thing is that none of these overrides say that when `allow_empty_body=False` then the return value is `JsonDict`. I think we actually want to change the variant above with `Literal[True]` to be `Literal[False]` and return a `JsonDict`. (You could specify both a `Literal[False]` and `Literal[True]` variant, but the `Literal[True]` variant is the same as the non literal variant and so its redundant).

clokep: Ah, maybe that's where I messed up then. The suggested change causes mypy to complain about overlapping definitions.

erikjohnston: I get very confused at the correct way of doing these sort of overrides every single time tbf

erikjohnston (approved):

PR #10743, opened 2021-09-02
Fix bug with reusing 'txn' when persisting event.

### LINKED ISSUE #10739: `txn` being redefined in for loop as we are iterating over it
Spurred on by the note in https://github.com/matrix-org/synapse/pull/10734 about `txn` being redefined in for loop as we are iterating over it.

> We were originally iterating over `txn`, which can work, but as `txn` was being used to run a
>
> https://github.com/matrix-org/synapse/pull/10734

I was curious if there were other cases in the codebase, so I searched for `in txn:` and found one more case. It looks like this would cause a problem if there was an event where the `event_auth_chains` -> `chain_id` was not defined.

https://github.com/matrix-org/synapse/blob/c586d6803a2adebcdd486be2d9eac1f62fd7d4ab/synapse/storage/databases/main/events.py#L578-L588

## Linting

Maybe we should disallow iterating over `txn` directly in favor of using `txn.fetchall()`

erikjohnston:
I don't think its possible for this to happen unless there are multiple out of band memberships in the room that were persisted in an older version of Synapse that didn't do auth chains.

Fixes #10739

### DIFF
--- a/changelog.d/10743.bugfix
+++ b/changelog.d/10743.bugfix
@@ -1,1 +1,1 @@
-Fix joining room when the server has received multiple invites for different users in the room.
+Fix edge case when persisting events into a room where there are multiple events we previously hadn't calculated auth chains for (and hadn't marked as needing to be calculated).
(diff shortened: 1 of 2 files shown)


reivilibre (approved):
Interesting, subtle problem!

Fix LGTM

PR #10744, opened 2021-09-02
Move `maybe_kick_guest_users` out of `BaseHandler`

richvdh:
This is part of my ongoing war against `BaseHandler`. I've moved `kick_guest_users` into `RoomMemberHandler` (since it calls out to that handler anyway), and split `maybe_kick_guest_users` into the two places it is called.

### DIFF
--- a/synapse/handlers/room_member.py
+++ b/synapse/handlers/room_member.py
@@ -44,6 +44,7 @@ from synapse.types import (
     RoomID,
     StateMap,
     UserID,
+    create_requester,
     get_domain_from_id,
 )
 from synapse.util.async_helpers import Linearizer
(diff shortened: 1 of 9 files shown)


synapse/handlers/room_member.py
```diff
@@ -1352,7 +1405,6 @@ def __init__(self, hs: "HomeServer"):
 
         self.distributor = hs.get_distributor()
         self.distributor.declare("user_left_room")
-        self._server_name = hs.hostname
```
richvdh, line 1355: this is now redundant, since `RoomMemberHandler` has its own `_server_name`.

erikjohnston (approved):

PR #10747, opened 2021-09-02
Support MSC3375: room version 9.

clokep:
Per [MSC3375](https://github.com/matrix-org/matrix-doc/pull/3375).

This needs to wait for that the MSC to be merged.

### DIFF
--- a/synapse/api/room_versions.py
+++ b/synapse/api/room_versions.py
@@ -214,7 +214,7 @@ class RoomVersions:
     )
     V8 = RoomVersion(
         "8",
-        RoomDisposition.UNSTABLE,
+        RoomDisposition.STABLE,
         EventFormatVersions.V3,
         StateResolutionVersions.V2,
         enforce_key_validity=True,
--- a/synapse/api/room_versions.py
+++ b/synapse/api/room_versions.py
@@ -70,6 +70,9 @@ class RoomVersion:
     msc2176_redaction_rules = attr.ib(type=bool)
     # MSC3083: Support the 'restricted' join_rule.
     msc3083_join_rules = attr.ib(type=bool)
+    # MSC3375: Support for the proper redaction rules for MSC3083. This mustn't
+    #          be enabled if MSC3083 is not.
+    msc3375_redaction_rules = attr.ib(type=bool)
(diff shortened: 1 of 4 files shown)


synapse/api/room_versions.py
```diff
@@ -197,12 +207,29 @@ class RoomVersions:
...
         msc2176_redaction_rules=False,
         msc3083_join_rules=False,
+        msc3375_redaction_rules=False,
         msc2403_knocking=True,
         msc2716_historical=False,
         msc2716_redactions=False,
     )
     V8 = RoomVersion(
         "8",
+        RoomDisposition.UNSTABLE,
```
clokep, line 217:
```suggestion
        RoomDisposition.STABLE,
```

synapse/api/room_versions.py
```diff
@@ -293,7 +324,7 @@ class RoomVersionCapability:
         ),
         RoomVersionCapability(
             "restricted",
-            RoomVersions.V8,
+            RoomVersions.V9,
```
clokep, line 327: We probably don't want to do this at the same time, we might want to ensure we get some people upgraded first.

clokep: My initial feeling is we should probably wait 1 version before doing this, but would be curious of other's thoughts.

erikjohnston: I think it depends on how the clients are representing this flag to the user. If we enable it then obviously at first very few other servers will be able to join the newly created rooms, which might be fine if the UI makes it clear its experimental?

clokep: It uses this to determine what room version to upgrade to if attempting to change a room to use restricted join rules for an old room version. This leads me to think we should wait until there's at least some uptake before doing this? 🤷 

image

erikjohnston (approved):

synapse/api/room_versions.py
clokep: After some discussion the decision was to change the preferred version 1 release after v9 is available.

clokep:
```suggestion
            RoomVersions.V8,
```

PR #10750, opened 2021-09-03
Ease searching for M_TOO_LARGE-related error codes

anoadragon453:
I recall having a difficult time searching through our GitHub org and Synapse's codebase directly in order to find the origin of an 'event_id too large' error message. I eventually found it here are asking around.

Then today someone else ran through the exact same failure mode, and I had to help them out.

This small commit should put an end to that needless cycle, and is a good lesson on making error messages easy to search for in a codebase.

### DIFF
--- a/synapse/event_auth.py
+++ b/synapse/event_auth.py
@@ -217,15 +217,15 @@ def check(
 
 def _check_size_limits(event: EventBase) -> None:
     if len(event.user_id) > 255:
-        EventSizeError("user_id too large")
+        EventSizeError("'user_id' too large")
     if len(event.room_id) > 255:
-        EventSizeError("room_id too large")
+        EventSizeError("'room_id' too large")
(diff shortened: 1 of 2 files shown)


synapse/event_auth.py
```diff
@@ -216,21 +216,18 @@ def check(
...
+        EventSizeError("room_id too large")
     if event.is_state() and len(event.state_key) > 255:
-        too_big("state_key")
+        EventSizeError("state_key too large")
     if len(event.type) > 255:
-        too_big("type")
+        EventSizeError("type too large")
     if len(event.event_id) > 255:
-        too_big("event_id")
+        EventSizeError("event_id too large")
```
erikjohnston, line 228: While we're here we should quote the key names

erikjohnston (approved):

PR #10752, opened 2021-09-03
Pull out encrypted_by_default tests

DMRobertson:
Don't see why they're in user_directory?


anoadragon453 (approved):
Good question! I have no idea who could have done that :eyes: :sweat_drops:

PR #10753, opened 2021-09-03
Use attrs internally for the URL preview code & add documentation.

clokep:
This does two unrelated-ish things to the URL preview code:

1. Uses `attrs` internally to pass data between methods (instead of a `dict`).
2. Documents what happens when a URL preview is generated. This was mostly for me trying to figure stuff out, but I wanted to write it down.
3. Deletes the old design doc for the API, which was moved to matrix-doc: matrix-org/matrix-doc#3378. These are interesting, but aren't really Synapse specific so shouldn't be in this repo.


synapse/rest/media/v1/preview_url_resource.py
```diff
@@ -71,6 +74,20 @@
 ONE_HOUR = 60 * 60 * 1000
 
 
+@attr.s(slots=True, frozen=True, auto_attribs=True)
+class MediaInfo:
```
erikjohnston, line 78: Docstring?

erikjohnston (approved):

PR #10754, opened 2021-09-03
Use `execute_values` more in PostgreSQL

erikjohnston:
`execute_values` is a faster version of `execute_batch`.

### DIFF
--- a/synapse/storage/database.py
+++ b/synapse/storage/database.py
@@ -1297,8 +1297,8 @@ class DatabasePool:
             args.append(tuple(x) + tuple(y))
 
         if isinstance(txn.database_engine, PostgresEngine):
-            # We use `execute_values` for postgres as it can be a lot faster
-            # than `execute_batch`, but it's only available on postgres.
+            # We use `execute_values` as it can be a lot faster than `execute_batch`,
+            # but it's only available on postgres.
             sql = "INSERT INTO %s (%s) VALUES ? ON CONFLICT (%s) DO %s" % (
                 table,
                 ", ".join(k for k in allnames),
--- a/synapse/storage/database.py
+++ b/synapse/storage/database.py
@@ -280,18 +280,18 @@ class LoggingTransaction:
         else:
             self.executemany(sql, args)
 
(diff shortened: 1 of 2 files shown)


synapse/storage/database.py
```diff
@@ -920,13 +920,23 @@ def simple_insert_many_txn(
...
                 raise RuntimeError("All items must have the same keys")
 
-        sql = "INSERT INTO %s (%s) VALUES(%s)" % (
-            table,
-            ", ".join(k for k in keys[0]),
-            ", ".join("?" for _ in keys[0]),
-        )
+        if isinstance(txn.database_engine, PostgresEngine):
+            # We use `execute_values` for postgres as it can be a lot faster
+            # than `execute_batch`, but it's only available on postgres.
```
clokep, line 925: Seems we got a bit aggressive with the "is postgres" :) (I think this was copied below?)

```suggestion
            # We use `execute_values` as it can be a lot faster than `execute_batch`,
            # but it's only available on postgres.
```

synapse/storage/database.py
```diff
@@ -1281,20 +1291,33 @@ def simple_upsert_many_txn_native_upsert(
...
+            sql = "INSERT INTO %s (%s) VALUES ? ON CONFLICT (%s) DO %s" % (
+                table,
+                ", ".join(k for k in allnames),
+                ", ".join(key_names),
+                latter,
+            )
+
+            txn.execute_values(sql, args, fetch=False)
+
+        else:
```
clokep, line 1311: I thought sqlite doesn't have native upsert? So will this ever get called?

clokep (approved):
Should be fine if tests pass.

synapse/storage/database.py
erikjohnston: It can on sqlite 3.24+ :D

https://github.com/matrix-org/synapse/blob/00640ee71affeab0f6ed7e784925522437f076c5/synapse/storage/engines/sqlite.py#L50-L56

clokep: Oh, didn't realize that! 🎉

PR #10755, opened 2021-09-03
Speed up MultiWriterIdGenerator when lots of IDs are in flight.

erikjohnston:
If we have lots of `_unfinished_ids` in flight (which happens when persisting large numbers of events at once, e.g. during room joins) then its more efficient to keep the set sorted than repeatedly calling `min(_unfinished_ids)`


synapse/storage/util/id_generators.py
```diff
@@ -473,7 +474,7 @@ def _mark_id_as_finished(self, next_id: int):
 
                 finished = set()
 
-                min_unfinshed = min(self._unfinished_ids)
+                min_unfinshed = self._unfinished_ids[0]
```
clokep, line 477: If we only care about the minimum one, would it make more sense to track that separately? (Does `SortedSet` have any overhead compared to `set`?)

erikjohnston: Potentially, but in practice this change stopped the ID generators from appearing in the profiling, so :shrug:

stubs/sortedcontainers/sortedset.pyi
```diff
@@ -0,0 +1,118 @@
+# stub for SortedDict. This is a lightly edited copy of
```
clokep, line 1:
```suggestion
# stub for SortedSet. This is a lightly edited copy of
```

What was edited? 😄 I didn't look super closely at these, but they look reasonable overall.

clokep (approved):

stubs/sortedcontainers/sortedset.pyi
erikjohnston: I just isort/blacked it and copied that stub from the other files :)

PR #10756, opened 2021-09-03
Speed up persisting redacted events

erikjohnston:


synapse/storage/databases/main/events.py
```diff
@@ -1379,18 +1379,18 @@ def get_internal_metadata(event):
         # If we're persisting an unredacted event we go and ensure
         # that we mark any redactions that reference this event as
         # requiring censoring.
-        sql = "UPDATE redactions SET have_censored = ? WHERE redacts = ?"
-        txn.execute_batch(
```
clokep, line 1383: Why not `execute_values` here? I guess it's even faster to just do a single query?

erikjohnston: Yeah, so the rough order of speed I believe is `make_in_list_sql_clause > execute_values > execute_batch`. `execute_values` is just a helper function that internally does a bunch of sql building to turn a list of rows into something that postgres will understand, so just passing in the array bypasses all of that nonsense.

clokep (approved):

Issue #10716, opened 2021-08-28
can't create rooms with with an alias when createRoom is on a worker


cremesk:
### Description

Hello,

i can not create new rooms with the new generic_worker location `^/_matrix/client/(api/v1|r0|unstable)/createRoom$`.

### Steps to reproduce

- add the new location to the reserve proxy config:  
https://git.envs.net/envs/matrix-conf/src/branch/master/etc/nginx/include.d/generic_worker.conf#L145-L149

### Version information

- **Homeserver**: envs.net
- **Version**: 1.41.0
- **Install method**: apt
- **Platform**: Debian 11 / VM

### Log:
```
2021-08-28 18:31:05,066 - synapse.http.server - 93 - ERROR - POST-192338 - Failed handle request via 'RoomCreateRestServlet': <XForwardedForRequest at 0x7f907639c550 method='POST' uri='/_matrix/client/r0/createRoom' clientproto='HTTP/1.0' site=8513>
Traceback (most recent call last):
  File "/opt/venvs/matrix-synapse/lib/python3.9/site-packages/synapse/http/server.py", line 258, in _async_render_wrapper
    callback_return = await self._async_render(request)
  File "/opt/venvs/matrix-synapse/lib/python3.9/site-packages/synapse/http/server.py", line 446, in _async_render
    callback_return = await raw_callback_return
  File "/opt/venvs/matrix-synapse/lib/python3.9/site-packages/synapse/rest/client/room.py", line 84, in on_POST
    info, _ = await self._room_creation_handler.create_room(
  File "/opt/venvs/matrix-synapse/lib/python3.9/site-packages/synapse/handlers/room.py", line 726, in create_room
    await directory_handler.create_association(
  File "/opt/venvs/matrix-synapse/lib/python3.9/site-packages/synapse/handlers/directory.py", line 162, in create_association
    await self._create_association(room_alias, room_id, servers, creator=user_id)
  File "/opt/venvs/matrix-synapse/lib/python3.9/site-packages/synapse/handlers/directory.py", line 88, in _create_association
    await self.store.create_room_alias_association(
AttributeError: 'GenericWorkerSlavedStore' object has no attribute 'create_room_alias_association'
```

clokep:
Seems this was added in #10564, I wonder if creating a room with an alias wasn't taken into account. (Does it work if you don't create an alias at the same time?)

cremesk:
yes i had tried to create a public room.

cremesk:
[...]

PR #10757, opened 2021-09-03
Allow `room_alias_name` parameter to be handled by /createRoom calls on workers

anoadragon453:
Fixes https://github.com/matrix-org/synapse/issues/10716

[/createRoom](https://spec.matrix.org/unstable/client-server-api/#post_matrixclientr0createroom) recently gained support for being [run on worker processes](https://github.com/matrix-org/synapse/pull/10564). Unfortunately, it went unnoticed that passing the `room_alias_name` parameter to `/createRoom` fails on workers.

The reason for this not being caught beforehand was that the endpoint was never routed to workers in [sytest](https://github.com/matrix-org/sytest), thus the tests never raised a failure. Thus a PR for Sytest ~~will be coming along soon~~ is at https://github.com/matrix-org/sytest/pull/1128 which tests both this PR, and raises any other errors which may be lurking.

### DIFF
--- a/changelog.d/10757.bugfix
+++ b/changelog.d/10757.bugfix
@@ -1,0 +1,1 @@
+Fix a bug which prevented calls to `/createRoom` that included the `room_alias_name` parameter from being handled by worker processes.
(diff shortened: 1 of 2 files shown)


erikjohnston (approved):

wol-win:
Quick question: is this problem resulting in an error message / failure when creating a new public room on a HS version 1.42.0 which I currently observe on my installation?

clokep:
@wol-win See the stack trace in #10716

PR #10759, opened 2021-09-03
Request JSON for oEmbed requests.

clokep:
This builds on #10714 to do a few odds and ends related to oEmbed to ensure we have a good path forward:

* A bit more error checking of `providers.json`.
* Provides the proper `format=json` parameter on the URL. (At least Flickr requires this, otherwise it returns XML.)
* Ignores XML-only endpoints (since these aren't supported).

Nothing major here, but all of these things prevent you from just dropping the full `providers.json` in at the moment.


synapse/config/oembed.py
```diff
@@ -93,11 +95,22 @@ def _parse_and_validate_provider(
             # might have multiple patterns to match.
             for endpoint in provider["endpoints"]:
                 api_endpoint = endpoint["url"]
+
+                # The API endpoint must be an HTTP(S) URL.
+                results = urlparse.urlparse(api_endpoint)
+                if results.scheme not in {"http", "https"}:
+                    raise ConfigError(
+                        f"Insecure oEmbed scheme ({results.scheme}) for endpoint {api_endpoint}",
```
erikjohnston, line 103:
```suggestion
                        f"Unsupported oEmbed scheme ({results.scheme}) for endpoint {api_endpoint}",
```

Otherwise I assumed it meant we were forbidding `http` schemes

synapse/config/oembed.py
```diff
@@ -114,9 +127,12 @@ def _glob_to_pattern(self, glob: str, config_path: Iterable[str]) -> Pattern:
         """
         results = urlparse.urlparse(glob)
 
-        # Ensure the scheme does not have wildcards (and is a sane scheme).
+        # The scheme must be HTTP(S) (and cannot contain wildcards).
         if results.scheme not in {"http", "https"}:
-            raise ConfigError(f"Insecure oEmbed scheme: {results.scheme}", config_path)
+            raise ConfigError(
+                f"Insecure oEmbed scheme ({results.scheme}) for pattern: {glob}",
```
erikjohnston, line 133: Ditto

synapse/rest/media/v1/oembed.py
```diff
@@ -49,8 +49,23 @@ class OEmbedProvider:
...
+
+            # Only JSON is supported at the moment. This could be declared in
+            # the formats field. Otherwise, if the endpoint ends in .xml assume
+            # it doesn't support JSON.
+            if (
+                oembed_endpoint.formats is not None
+                and "json" not in oembed_endpoint.formats
+            ) or api_endpoint.endswith(".xml"):
+                logger.info(
+                    f"Ignoring oEmbed endpoint due to not supporting JSON: {api_endpoint}"
```
erikjohnston, line 62:
```suggestion
                    "Ignoring oEmbed endpoint due to not supporting JSON: %s", api_endpoint
```

clokep: Why not an f-string here?

erikjohnston: Because its logging, and using f-strings means that we unconditionally do the string interpolation rather than it being gated behind the log levels. Its not a *huge* deal, but it can add up (especially when we're interpolating variables with more expensive `str` functions)

erikjohnston (approved):

PR #10760, opened 2021-09-03
Return stripped state events for the room summary.

clokep:
The spaces summary / hierarchy endpoints should only return the stripped state events (+ `origin_server_ts`), not the entire event, for the children events of a room.

This is somewhat related to #10523 in that we'll be doing a decent chunk less work here.

### DIFF
--- a/changelog.d/10760.bugfix
+++ b/changelog.d/10760.bugfix
@@ -1,0 +1,1 @@
+Only return the stripped state events for the `m.space.child` events in a room for the spaces summary from [MSC2946](https://github.com/matrix-org/matrix-doc/pull/2946).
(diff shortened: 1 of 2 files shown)


erikjohnston (approved):

clokep:
I should note that this is similar to the `get_stripped_room_state_from_event_context` method used by invites and knocking:

https://github.com/matrix-org/synapse/blob/c4fa4f37cbc734f9cd6354a5f2661efc30d73cac/synapse/storage/databases/main/events_worker.py#L638-L687

That does a bunch of additional work (e.g. pulling state based on the pre-join state, etc.) so didn't seem applicable. I could have created a method to pull out the keys for the stripped state, but since it isn't identical it seemed better to inline it.

PR #10762, opened 2021-09-04
Fix reactivated users not being added to directory

DMRobertson:
Pulled this out of larger WIP changes because it's small and relatively isolated from the other changes.

The problem here is that `handle_local_profile_change` checks the DB to ensure the user isn't deactivated before upserting a directory entry.. So we need to reactivate them before calling this function, otherwise it'll do nothing.

### DIFF
--- a/synapse/handlers/deactivate_account.py
+++ b/synapse/handlers/deactivate_account.py
@@ -263,6 +263,8 @@ class DeactivateAccountHandler(BaseHandler):
         # Mark the user as active.
         await self.store.set_user_deactivated_status(user_id, False)
 
-        # Add the user to the directory, if necessary.
+        # Add the user to the directory, if necessary. Note that
+        # this must be done after the user is re-activated, because
+        # deactivated users are excluded from the user directory.
         profile = await self.store.get_profileinfo(user.localpart)
         await self.user_directory_handler.handle_local_profile_change(user_id, profile)
--- a/synapse/handlers/deactivate_account.py
+++ b/synapse/handlers/deactivate_account.py
@@ -257,11 +257,12 @@ class DeactivateAccountHandler(BaseHandler):
         """
         # Add the user to the directory, if necessary.
         user = UserID.from_string(user_id)
-        profile = await self.store.get_profileinfo(user.localpart)
-        await self.user_directory_handler.handle_local_profile_change(user_id, profile)
 
         # Ensure the user is not marked as erased.
         await self.store.mark_user_not_erased(user_id)
 
         # Mark the user as active.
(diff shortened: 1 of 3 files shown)


synapse/handlers/deactivate_account.py
```diff
@@ -257,11 +257,12 @@ async def activate_account(self, user_id: str) -> None:
         """
         # Add the user to the directory, if necessary.
```
dklimpel, line 258:
```suggestion

```

synapse/handlers/deactivate_account.py
```diff
@@ -257,11 +257,12 @@ async def activate_account(self, user_id: str) -> None:
...
         user = UserID.from_string(user_id)
-        profile = await self.store.get_profileinfo(user.localpart)
-        await self.user_directory_handler.handle_local_profile_change(user_id, profile)
 
         # Ensure the user is not marked as erased.
         await self.store.mark_user_not_erased(user_id)
 
         # Mark the user as active.
         await self.store.set_user_deactivated_status(user_id, False)
+
```
dklimpel, line 266:
```suggestion

        # Add the user to the directory, if necessary.
```

synapse/handlers/deactivate_account.py
```diff
@@ -255,13 +255,14 @@ async def activate_account(self, user_id: str) -> None:
...
+        # Add the user to the directory, if necessary.
```
clokep, line 266: Maybe to help us avoid re-introducing this in the future?

```suggestion
        # Add the user to the directory, if necessary. Note that this must be done
        # after the user is re-activated.
```

tests/handlers/test_user_directory.py
```diff
@@ -131,6 +132,52 @@ def test_handle_user_deactivated_regular_user(self):
...
+            deactivate_handler.deactivate_account(
+                user, erase_data=False, requester=create_requester(admin_user)
+            )
+        )
+        s = self.get_success(self.handler.search_users(admin_user, user, 10))
+        self.assertEqual(s["results"], [])
+
+        # Reactivate the user
+        self.get_success(deactivate_handler.activate_account(user))
+        # Hackily reset password by restoring the old pw hash.
```
clokep, line 165: Could we use the `SetPasswordHandler.set_password` API instead?

clokep (commented):
Looks good overall! Thanks for adding tests!

clokep:
For cross-referencing, I'm assuming this was split out of #10761?

DMRobertson:
> For cross-referencing, I'm assuming this was split out of #10761?

That's right, I spotted it as part of that work.

clokep (approved):
Looks great!

PR #10768, opened 2021-09-06
Add a warning when using deprecated template_dir settings

babolivier:
The deprecation itself happened in https://github.com/matrix-org/synapse/pull/10596 which shipped with Synapse v1.41.0. However, it doesn't seem fair to suddenly drop support for these settings in ~4-6w without being more vocal about said deprecation.

### DIFF
--- a/synapse/config/sso.py
+++ b/synapse/config/sso.py
@@ -11,11 +11,22 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import logging
 from typing import Any, Dict, Optional
 
 import attr
 
 from ._base import Config
+
+logger = logging.getLogger(__name__)
+
+LEGACY_TEMPLATE_DIR_WARNING = """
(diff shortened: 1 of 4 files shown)


synapse/config/sso.py
```diff
@@ -11,12 +11,23 @@
...
 
+logger = logging.getLogger(__name__)
+
+LEGACY_TEMPLATE_DIR_WARNING = """
+This server's configuration file is using the deprecated 'template_dir' setting in the
+'sso' section. Support for this setting has been deprecated and will be removed in a
+future version of Synapse. Server admins should instead use the new
+'custom_templates_directory' setting documented here:
+https://matrix-org.github.io/synapse/latest/templates.html
+---------------------------------------------------------------------------------------"""
```
erikjohnston, line 29: Hmm, I guess this could be printed out twice, once for `sso` and `email`? We could be more clever but probably not worth it?

babolivier: Yeah plus I think it's better that the user sees as many warnings as they have obsolete settings in use.

anoadragon453 (approved):

babolivier:
Ignoring Sytest failures as they seem unrelated to this change.

PR #10771, opened 2021-09-07
Persist auth events before the events that rely on them

richvdh:
If we're persisting an event `E` which has `auth_event`s `A1`, `A2`, then we ought to make sure that we correctly auth and persist `A1` and `A2`, before we blindly accept `E`.

This PR does *part* of that - it persists the auth events first - but it does not fully solve the problem, because we still don't check that the auth events weren't rejected.

It's a bit of a refactorathon, but it should be reviewable commit-by-commit.

### DIFF
--- a/changelog.d/10771.misc
+++ b/changelog.d/10771.misc
@@ -1,0 +1,1 @@
+Clean up some of the federation event authentication code for clarity.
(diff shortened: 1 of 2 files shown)


clokep (approved):
Seems reasonable to me.

PR #10773, opened 2021-09-07
Add some assertions about outliers

richvdh:
I think I have finally teased apart the codepaths which handle outliers, and those that handle non-outliers. Let's add some assertions to demonstrate my newfound knowledge.

### DIFF
--- a/synapse/handlers/federation_event.py
+++ b/synapse/handlers/federation_event.py
@@ -1282,15 +1282,16 @@ class FederationEventHandler:
                 Possibly incomplete, and possibly including events that are not yet
                 persisted, or authed, or in the right room.
 
-                Only populated where we may not already have persisted these events -
-                for example, when populating outliers, or the state for a backwards
-                extremity.
+                Only populated when populating outliers.
 
             backfilled: True if the event was backfilled.
 
         Returns:
             The updated context object.
         """
+        # claimed_auth_event_map should be given iff the event is an outlier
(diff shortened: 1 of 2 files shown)


synapse/handlers/federation_event.py
```diff
@@ -232,77 +235,71 @@ async def on_receive_pdu(self, origin: str, pdu: EventBase) -> None:
             )
             return None
 
-        # Check that the event passes auth based on the state at the event. This is
-        # done for events that are to be added to the timeline (non-outliers).
-        #
-        # Get missing pdus if necessary:
-        #  - Fetching any missing prev events to fill in gaps in the graph
-        #  - Fetching state if we have a hole in the graph
-        if not pdu.internal_metadata.is_outlier():
```
richvdh, line 241: this became a tautology after https://github.com/matrix-org/synapse/pull/10614/files#diff-0bc92da3d703202f5b9be2d3f845e375f5b1a6bc6ba61705a8af9be1121f5e42L301, for the record.

clokep (approved):
LGTM!

I did wonder if we should be using non-`assert` checks here, but we seem to go back and forth on when to do that. My understanding is that if these asserts fail it is due to a programming error? If so, I think this is OK.

richvdh:
> My understanding is that if these asserts fail it is due to a programming error? 

Yes, exactly. Also they're really there to demonstrate intention - a form of documentation to say "this thing should never be called with outliers". I'd be pretty happy for them to be optimised out if someone wanted to run with `-O`.

PR #10774, opened 2021-09-07
Handle room upgrades for spaces

### LINKED ISSUE #10207: Performing a room upgrade on a space will not transition the type
If a room upgrade is performed on a space then the `type` field from the content will be lost.

We seem to pull across the `m.federate` content:

https://github.com/matrix-org/synapse/blob/a683028d81606708f686b890c0a44f5a20b54798/synapse/handlers/room.py#L386-L392

But I wonder if we should start with the old event's content and then splat `room_version` and `predecessor` on top of it?

The [spec is a bit vague](https://matrix.org/docs/spec/client_server/r0.6.1#id163), but seems to imply we shouldn't even be copying `m.federate`:

> Creates a replacement room with a `m.room.create` event containing a `predecessor` field and the applicable `room_version`.

We also will want to copy the `m.space.child` and `m.space.parent` events from the space. It is unclear what we would do with the `m.space.parent` events in rooms that point to the space.

This might need an MSC for discussion, or might just be an implementation detail.

clokep:
This handles room upgrades for spaces by transferring the `room_type` field of the create event and copying any `m.space.child` events from the previous room.

Fixes #10207, per the comment in there the upgrade process is somewhat implementation specific, so I think this change is fine to make without corresponding spec changes.

Open questions:

* Should we be ignoring `m.space.child` events which have no content (and thus were rooms that were removed from the space)?

### DIFF
--- a/synapse/handlers/room.py
+++ b/synapse/handlers/room.py
@@ -33,6 +33,7 @@ from synapse.api.constants import (
     Membership,
     RoomCreationPreset,
     RoomEncryptionAlgorithms,
+    RoomTypes,
 )
 from synapse.api.errors import (
     AuthError,
@@ -397,7 +398,7 @@ class RoomCreationHandler(BaseHandler):
         initial_state = {}
 
         # Replicate relevant room events
-        types_to_copy = (
+        types_to_copy: List[Tuple[str, Optional[str]]] = [
             (EventTypes.JoinRules, ""),
             (EventTypes.Name, ""),
             (EventTypes.Topic, ""),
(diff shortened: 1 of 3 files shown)


richvdh (approved):

synapse/handlers/room.py
```diff
@@ -408,7 +409,16 @@ async def clone_existing_room(
...
-        )
+        ]
+
+        # If the old room was a space, copy over the room type and the rooms in
+        # the space.
+        if (
+            old_room_create_event.content.get(EventContentFields.ROOM_TYPE)
+            == RoomTypes.SPACE
+        ):
+            creation_content[EventContentFields.ROOM_TYPE] = RoomTypes.SPACE
```
clokep, line 420: Note that [MSC3385](https://github.com/matrix-org/matrix-doc/pull/3385) (made after this PR) suggests using the full create content. I think we can leave that to a follow-up depending on how that MSC goes...

clokep:
@richvdh I added support to ignore any `m.space.child` events with empty content when upgrading (as those are children which were removed). Could you please take another look!

richvdh (approved):
lgtm!

PR #10775, opened 2021-09-07
Add a constant for m.federate.

clokep:
Adds a constant for `m.federate` to `EventContentFields` and uses it through our code.

### DIFF
--- a/synapse/handlers/room_list.py
+++ b/synapse/handlers/room_list.py
@@ -313,7 +313,9 @@ class RoomListHandler(BaseHandler):
 
         # Return whether this room is open to federation users or not
         create_event = current_state[EventTypes.Create, ""]
-        result["m.federate"] = create_event.content.get("m.federate", True)
+        result["m.federate"] = create_event.content.get(
+            EventContentFields.FEDERATE, True
+        )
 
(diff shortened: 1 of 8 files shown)


synapse/handlers/room_list.py
```diff
@@ -313,7 +313,9 @@ async def generate_room_entry(
 
         # Return whether this room is open to federation users or not
         create_event = current_state[EventTypes.Create, ""]
-        result["m.federate"] = create_event.content.get("m.federate", True)
+        result["m.federate"] = create_event.content.get(
```
squahtx, line 316: Is this `"m.federate"` intentionally left in?

squahtx (approved):
LGTM, 1 small question

synapse/handlers/room_list.py
clokep: I left it in because it describes the API shape (as opposed to event content). They *are* the same (and I assume they're the same because it is the name of the event field)... Do you think I should update it? I could go either way.

squahtx: I've got no strong opinion on this one. It's up to you :)

PR #10778, opened 2021-09-08
Add a script to update the debian changelog for non-Debian systems

babolivier:
When releasing 1.42.0 with @Azrenbeth and talking with @clokep yesterday I realised doing the `dch` incantations related to releasing Synapse wasn't trivial, so I thought I'd write a script to run in a Debian container to make things a bit easier.

### DIFF
--- a/scripts-dev/docker_update_debian_changelog.sh
+++ b/scripts-dev/docker_update_debian_changelog.sh
@@ -1,4 +1,4 @@
-#!/bin/bash -e
+#!/bin/bash
 # Copyright 2021 The Matrix.org Foundation C.I.C.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
@@ -47,7 +47,9 @@ fi
 
 # We set -x here rather than in the shebang so that if we need to exit early because no
 # version was provided, the message doesn't get drowned in useless output.
-set -x
+# We also set -e here rather than in the shebang, because we don't want the script to exit
+# without any error message if one of the checks before (which and grep) returned 1.
+set -xe
 
 # Make the root of the Synapse checkout the current working directory.
 cd /synapse
--- a/scripts-dev/docker_update_debian_changelog.sh
+++ b/scripts-dev/docker_update_debian_changelog.sh
@@ -18,7 +18,7 @@
 # not easily accessible.
 #
 # Running it (when if the current working directory is the root of the Synapse checkout):
-#   docker run -v $PWD:/synapse ubuntu:latest /synapse/scripts-dev/update_debian_changelog.sh VERSION
+#   docker run -v $PWD:/synapse ubuntu:latest /synapse/scripts-dev/docker_update_debian_changelog.sh VERSION
 #
 # The image can be replaced by any other Debian-based image (as long as the `devscripts`
 # package exists in the default repository).
@@ -28,6 +28,20 @@
 if [ "$#" -ne 1 ]; then
   echo "Usage: update_debian_changelog.sh VERSION"
   echo "VERSION is the version of Synapse being released in the form 1.42.0 (without the leading \"v\")"
+  exit 1
+fi
+
+# Check that apt-get is available on the system.
+which apt-get > /dev/null 2>&1
+if [ "$?" -ne 0 ]; then
+  echo "\"apt-get\" isn't available on this system. This script needs to be run in a Docker container using a Debian-based image."
+  exit 1
+fi
+
+# Check if devscripts is available in the default repos for this distro.
+apt-cache search devscripts | grep -E "^devscripts \-" > /dev/null
+if [ "$?" -ne 0 ]; then
+  echo "The package \"devscripts\" needs to exist in the default repositories for this distribution."
   exit 1
 fi
 
--- a/scripts-dev/docker_update_debian_changelog.sh
+++ b/scripts-dev/docker_update_debian_changelog.sh
@@ -54,8 +54,8 @@ set -x
 # Make the root of the Synapse checkout the current working directory.
 cd /synapse
 
-# Install devscripts (which provides dch). We need to make the Debian frontent noninteractive
-# because installing devscripts otherwise asks for the machine's location.
+# Install devscripts (which provides dch). We need to make the Debian frontend
+# noninteractive because installing devscripts otherwise asks for the machine's location.
 DEBIAN_FRONTEND=noninteractive apt-get install -y devscripts
 
 # Update the Debian changelog.
--- a/scripts-dev/docker_update_debian_changelog.sh
+++ b/scripts-dev/docker_update_debian_changelog.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/bash -e
 # Copyright 2021 The Matrix.org Foundation C.I.C.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
@@ -49,9 +49,7 @@ fi
 
 # We set -x here rather than in the shebang so that if we need to exit early because no
 # version was provided, the message doesn't get drowned in useless output.
-# We also set -e here rather than in the shebang, because we don't want the script to exit
-# without any error message if one of the checks before (which and grep) returned 1.
-set -xe
+set -x
 
 # Make the root of the Synapse checkout the current working directory.
 cd /synapse
--- a/scripts-dev/docker_update_debian_changelog.sh
+++ b/scripts-dev/docker_update_debian_changelog.sh
@@ -32,15 +32,17 @@ if [ "$#" -ne 1 ]; then
 fi
 
 # Check that apt-get is available on the system.
-which apt-get > /dev/null 2>&1
-if [ "$?" -ne 0 ]; then
+if ! which apt-get > /dev/null 2>&1; then
   echo "\"apt-get\" isn't available on this system. This script needs to be run in a Docker container using a Debian-based image."
   exit 1
 fi
 
 # Check if devscripts is available in the default repos for this distro.
-apt-cache search devscripts | grep -E "^devscripts \-" > /dev/null
-if [ "$?" -ne 0 ]; then
+# Update the apt package list cache.
+# We need to do this before we can search the apt cache or install devscripts.
+apt-get update || exit 1
+
+if ! apt-cache search devscripts | grep -E "^devscripts \-" > /dev/null; then
   echo "The package \"devscripts\" needs to exist in the default repositories for this distribution."
   exit 1
 fi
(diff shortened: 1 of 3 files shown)


scripts-dev/update_debian_changelog.sh
```diff
@@ -0,0 +1,50 @@
+#!/bin/bash -e
```
reivilibre, line 1: should we make the title of this script a bit more indicative that it's specific to Docker?

scripts-dev/update_debian_changelog.sh
```diff
@@ -0,0 +1,50 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This script is meant to be used inside a Docker container to run the `dch` incantations
+# needed to release Synapse. This is useful on systems like macOS where such scripts are
+# not easily accessible.
+#
+# Running it (when if the current working directory is the root of the Synapse checkout):
```
reivilibre, line 20:
```suggestion
# Running it (when the current working directory is the root of the Synapse checkout):
```

scripts-dev/update_debian_changelog.sh
```diff
@@ -0,0 +1,50 @@
...
+#   docker run -v $PWD:/synapse ubuntu:latest /synapse/scripts-dev/update_debian_changelog.sh VERSION
```
reivilibre, line 21:
```suggestion
#   docker run --rm -v $PWD:/synapse ubuntu:latest /synapse/scripts-dev/update_debian_changelog.sh VERSION
```

otherwise you're spinning up a new one every time

scripts-dev/update_debian_changelog.sh
```diff
@@ -0,0 +1,50 @@
...
+#
+# The image can be replaced by any other Debian-based image (as long as the `devscripts`
+# package exists in the default repository).
+# `VERSION` is the version of Synapse being released without the leading "v" (e.g. 1.42.0).
+
+# Check if a version was provided.
+if [ "$#" -ne 1 ]; then
+  echo "Usage: update_debian_changelog.sh VERSION"
+  echo "VERSION is the version of Synapse being released in the form 1.42.0 (without the leading \"v\")"
```
reivilibre, line 30:
```suggestion
  echo "VERSION is the version of Synapse being released in the form 1.42.0 (without the leading \"v\")"
  echo "This script must be executed in a container where Synapse is at /synapse"
  echo "and apt-get is available to install the `devscripts` package."
```
?

scripts-dev/update_debian_changelog.sh
babolivier: The instructions on how the script is meant to be run are in L16, which starts off by saying it's meant to be run in a container. Not sure how to be more verbose here.

scripts-dev/update_debian_changelog.sh
babolivier: That's not what this statement checks. I assume the user has read the instructions on how to run the script, which already mention the image needs to be Debian-based and have the `devscripts` package available.

scripts-dev/update_debian_changelog.sh
babolivier: Oooh TIL, thanks!
(technically it _is_ spinning up a new one every time, this removes the container and its volume after it runs according to the docs)

scripts-dev/update_debian_changelog.sh
reivilibre: The name of the script seems very plain for it to actually turn out to require Docker

reivilibre: (i.e. I would suggest `docker_update_debian_changelog.sh` but not sure if you can do better)

babolivier:
fwiw @Azrenbeth has run this script on his macOS laptop and it seems to be working perfectly.

scripts-dev/update_debian_changelog.sh
reivilibre: Yes, but they'll linger and eat your disk space :(

scripts-dev/update_debian_changelog.sh
reivilibre: If you assume the user has read the instructions, why are you printing the usage? I'm not sure that's consistent.

scripts-dev/update_debian_changelog.sh
babolivier: Yeah I was thinking about doing something like that but didn't want people to misunderstand it as a script that is only needed when eg building docker images. But if you also think it could be worth renaming it that way then I'll do it.

scripts-dev/update_debian_changelog.sh
babolivier: Because it's imho more likely that the user forgets an argument (or eg passes in the `$ver` variable from the release process but hasn't initialised it correctly) than that they run the script in the wrong context (especially since it'll eventually be just a line to copy/paste from the release instructions).
I'll add extra checks though to make sure apt-get is available and the devscripts package is too.

scripts-dev/update_debian_changelog.sh
reivilibre: Hmm, I see your point. I think *something* would be nice though. 

I would probably be happy to call it `update_debian_changelog_in_docker.sh`. I also guess that docker image-specific scripts would live in `synapse/docker` so hopefully it's enough; thoughts?

It also means we can have a non-docker equivalent in the future if we so wished, without awkward naming issues.

scripts-dev/docker_update_debian_changelog.sh
```diff
@@ -0,0 +1,66 @@
...
+
+# Check that apt-get is available on the system.
+which apt-get > /dev/null 2>&1
+if [ "$?" -ne 0 ]; then
+  echo "\"apt-get\" isn't available on this system. This script needs to be run in a Docker container using a Debian-based image."
+  exit 1
+fi
+
+# Check if devscripts is available in the default repos for this distro.
+apt-cache search devscripts | grep -E "^devscripts \-" > /dev/null
```
reivilibre, line 42: will this fail on a new Debian container because you haven't updated your cache with `apt update`?

scripts-dev/docker_update_debian_changelog.sh
```diff
@@ -0,0 +1,66 @@
...
+fi
+
+# We set -x here rather than in the shebang so that if we need to exit early because no
+# version was provided, the message doesn't get drowned in useless output.
+# We also set -e here rather than in the shebang, because we don't want the script to exit
+# without any error message if one of the checks before (which and grep) returned 1.
+set -xe
+
+# Make the root of the Synapse checkout the current working directory.
+cd /synapse
```
reivilibre, line 55: I guess this will already make complaining noises if it's misused (by not running in Docker).

Up to your opinion, but should we have something like the following?

```
if [ ! -d /synapse ]; then
  echo "Synapse's source has not been mounted to /synapse. This script needs to be run in a Docker container."
  exit 1
fi

```

scripts-dev/docker_update_debian_changelog.sh
```diff
@@ -0,0 +1,66 @@
...
+
+# Update the packages list and install devscripts (which provides dch). We need to make
+# the Debian frontent noninteractive because installing devscripts otherwise asks for the
+# machine's location.
+apt-get update
+DEBIAN_FRONTEND=noninteractive apt-get install -y devscripts
```
reivilibre, line 61: btw woo on `DEBIAN_FRONTEND=noninteractive`, it took me a long time to find out that `-yqq` wasn't good enough and you need this env var to stop apt blocking for input (in the case of some packages; devscripts is probably fine but better safe than sorry)

scripts-dev/docker_update_debian_changelog.sh
reivilibre: yes

scripts-dev/docker_update_debian_changelog.sh
babolivier: Not sure it's worth it tbh, I feel like the error bash shows and the documentation should be enough.

scripts-dev/docker_update_debian_changelog.sh
```diff
@@ -0,0 +1,66 @@
...
+  echo "\"apt-get\" isn't available on this system. This script needs to be run in a Docker container using a Debian-based image."
+  exit 1
+fi
+
+# Check if devscripts is available in the default repos for this distro.
+apt-cache search devscripts | grep -E "^devscripts \-" > /dev/null
+if [ "$?" -ne 0 ]; then
+  echo "The package \"devscripts\" needs to exist in the default repositories for this distribution."
+  exit 1
+fi
```
reivilibre, line 46: (the if style is up to your taste)

```suggestion
# Update the apt package list cache.
# We need to do this before we can search the apt cache or install devscripts.
apt-get update || exit 1

if ! apt-cache search devscripts | grep -E "^devscripts \-" > /dev/null; then
  echo "The package \"devscripts\" needs to exist in the default repositories for this distribution."
  exit 1
fi
```

scripts-dev/docker_update_debian_changelog.sh
```diff
@@ -0,0 +1,66 @@
...
+# without any error message if one of the checks before (which and grep) returned 1.
+set -xe
+
+# Make the root of the Synapse checkout the current working directory.
+cd /synapse
+
+# Update the packages list and install devscripts (which provides dch). We need to make
+# the Debian frontent noninteractive because installing devscripts otherwise asks for the
+# machine's location.
+apt-get update
```
reivilibre, line 60:
```suggestion
# Install devscripts (which provides dch). We need to make the Debian frontent noninteractive
# because installing devscripts otherwise asks for the machine's location.
```

oh apparently you do need noninteractive, heh :)

reivilibre (changes requested):

scripts-dev/docker_update_debian_changelog.sh
reivilibre: (addressed in another review suggestion)

scripts-dev/docker_update_debian_changelog.sh
babolivier: devscripts does block for input - see the comment above that line. This is why we're using this frontend :)

scripts-dev/docker_update_debian_changelog.sh
```diff
@@ -0,0 +1,66 @@
...
+# Check if a version was provided.
+if [ "$#" -ne 1 ]; then
+  echo "Usage: update_debian_changelog.sh VERSION"
+  echo "VERSION is the version of Synapse being released in the form 1.42.0 (without the leading \"v\")"
+  exit 1
+fi
+
+# Check that apt-get is available on the system.
+which apt-get > /dev/null 2>&1
+if [ "$?" -ne 0 ]; then
```
babolivier, line 36:
```suggestion
if ! which apt-get > /dev/null 2>&1; then
```

reivilibre (approved):
LGTM!

scripts-dev/docker_update_debian_changelog.sh
```diff
@@ -0,0 +1,64 @@
+#!/bin/bash -e
```
richvdh, line 1: I'd say putting `-e` in the shebang isn't great style. If your script relies on `-e` being set, why not do so with `set -e` - which means it doesn't matter if it gets run as `./docker_update_debian_changelog.sh` or `bash docker_update_debian_changelog.sh`.

Not a big deal, just seems an odd choice to me.

PR #10779, opened 2021-09-08
Change logging of puppeted requests to better differentiate users

erikjohnston:
This used to be a comma and got accidentally changed to a period in https://github.com/matrix-org/synapse/pull/9654, but a pipe character is more easier to parse visually

```
{@ems_administrator:one-staging.ems.host,@eotestingcheckboxtest:one-staging.ems.host}
```

vs
```
{@ems_administrator:one-staging.ems.host|@eotestingcheckboxtest:one-staging.ems.host}
```

### DIFF
--- a/changelog.d/10779.misc
+++ b/changelog.d/10779.misc
@@ -1,0 +1,1 @@
+Change the format of authenticated users in logs when a user is being puppeted by and admin user.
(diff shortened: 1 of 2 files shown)


clokep (approved):
LGTM if tests pass.

PR #10780, opened 2021-09-08
Skip handling of push actions for outlier events

erikjohnston:
Outlier events don't ever have push actions associated with them, so we
can skip some expensive queries during event persistence.

### DIFF
--- a/synapse/storage/databases/main/events.py
+++ b/synapse/storage/databases/main/events.py
@@ -1990,6 +1990,15 @@ class PersistEventsStore:
                 events_and_context.
         """
 
+        # Only non outlier events will have push actions associated with them,
+        # so let's filter them out. (This makes joining large rooms faster, as
+        # these queries took seconds to process all the state events).
+        non_outlier_events = [
+            event
+            for event, _ in events_and_contexts
+            if not event.internal_metadata.is_outlier()
+        ]
+
         sql = """
             INSERT INTO event_push_actions (
                 room_id, event_id, user_id, actions, stream_ordering,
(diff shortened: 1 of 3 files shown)


clokep (approved):
Could be helpful to add type-hints to these methods as we go along, but I won't block this PR on it!

synapse/storage/databases/main/events.py
```diff
@@ -1990,6 +1990,15 @@ def _set_push_actions_for_event_and_users_txn(
                 events_and_context.
         """
 
+        # Only non outlier events will have push actions associated with them,
+        # so let's filter them out. (This makes joining large rooms faster, as
+        # these queries took seconds to process all the state events).
```
DMRobertson, line 1995: Let's see if I can reason through this.

* An event in the DAG is an outlier if we don't have the state at that point in the DAG
* "Don't have the state" meaning "hat state could always be computed if needed, but we don't have easy access to it"?

> Only non outlier events will have push actions associated with them

Not sure why this follows. What's a push action?

erikjohnston: Push actions is the output of running a user's push rules on an event and state and basically says if and how the user should be notified about the event (e.g. should they be bushed? Should it be highlighted? etc).

An `outlier` really means that its not part of the "timeline", i.e. the connected chunk(s) of the rooms graph that makes up the history/timeline we send down to the clients. Think timeline events as being a sequence of messages, whereas outliers being state that was set (e.g. room name) before you joined the room. You're right that we track state for events in the timeline and not for outliers, but the reason we won't have push actions for outliers is that we only calculate them for timeline events, as we only notify for events that have just been added to the timeline rather than for old state events.

PR #10782, opened 2021-09-08
Fix reactivated users not being added to the user directory

DMRobertson:
This is #10762, but targeted directly at develop.

When cherry-picking the commits, there was on merge conflict. See https://github.com/matrix-org/synapse/commit/ad6eab0c2002e69cdd3996bb3bb89baa124262fd 's commit message for details.


changelog.d/10762.bugfix
```diff
@@ -0,0 +1 @@
+Fix a bug where reactivated users would be missing from the user directory.
```
clokep, line 1: We can probably just remove this one (but also be sure to remove it from the other branch).

clokep (approved):

changelog.d/10762.bugfix
DMRobertson: Not sure I follow you here (given that towncrier coalesces identical fragments). Is there a specific gain by tidying this up, or is it just nicer to have?

(Happy to clean up, but just anxious that it's further git-churn)

clokep: It doesn't seem useful is all, it would end up pointing people to two identical PRs, which seems confusing.

I don't feel strongly either way, however.

DMRobertson: Oh, looking at the review page properly, I see what you mean: this PR adds two changelog files!!! (I thought you were referring to the two different files on two different PRs)

DMRobertson:
@richvdh  and @reivilibre, I propose this as the next part of the user_directory changes.

tests/handlers/test_user_directory.py
```diff
@@ -130,6 +131,50 @@ def test_handle_user_deactivated_regular_user(self):
...
+        admin_user = self.register_user("admin", "pass", admin=True)
+
+        # Ensure the regular user is publicly visible and searchable.
+        self.helper.create_room_as(user, is_public=True, tok=user_token)
+        s = self.get_success(self.handler.search_users(admin_user, user, 10))
+        self.assertEqual(len(s["results"]), 1)
+        self.assertEqual(s["results"][0]["user_id"], user)
+
+        # Deactivate the user and check they're not searchable.
+        deactivate_handler = self.hs._deactivate_account_handler
```
reivilibre, line 153:
```suggestion
        deactivate_handler = self.hs.get_deactivate_account_handler()
```

tests/handlers/test_user_directory.py
```diff
@@ -130,6 +131,50 @@ def test_handle_user_deactivated_regular_user(self):
...
+        # Reactivate the user
+        self.get_success(deactivate_handler.activate_account(user))
+        # Hackily reset password by restoring the old pw hash.
+        self.get_success(
+            self.hs.get_set_password_handler().set_password(
+                user, password_hash, logout_devices=False
+            )
+        )
+        user_token = self.login(user, "pass")
+        self.helper.create_room_as(user, is_public=True, tok=user_token)
```
reivilibre, line 171: No strong feeling either way, but could have used the reactivation endpoint which will handle the password hash for you.

https://github.com/matrix-org/synapse/blob/bfb4b858a999684ba2459ee4c3aa20270d13062d/tests/rest/admin/test_user.py#L1975-L1986

reivilibre (commented):
Seems good to me.

Has this 'bug' made it out into a release, or is it purely because of your recent work? If the latter, would be tempted to coalesce the changelog with your recent work.

DMRobertson:
> Has this 'bug' made it out into a release, or is it purely because of your recent work? If the latter, would be tempted to coalesce the changelog with your recent work.

AFAICS this has has been around for a while; I haven't committed anything to develop that would affect this.

tests/handlers/test_user_directory.py
DMRobertson: Let's use the endpoint proper.

reivilibre (approved):
Thanks"!

changelog.d/10762.bugfix
richvdh: +1 please remove the duplicate newsfile!

changelog.d/10782.bugfix
```diff
@@ -0,0 +1 @@
+Fix a bug where reactivated users would be missing from the user directory.
```
richvdh, line 1: I'm assuming this has never worked?

```suggestion
Fix a long-standing bug which caused deactivated users that were later reactivated to be missing from the user directory.
```

richvdh (approved):
lgtm!

PR #10784, opened 2021-09-08
Don't needlessly batch in `add_event_to_cache`

erikjohnston:
We've already batched up the events previously, and assume in other
places in the events.py file that we have. Removing this makes it easier
to adjust the batch sizes in one place.


richvdh (approved):
sure

PR #10785, opened 2021-09-08
Convert FileInfo to attrs.

clokep:
This converts the `FileInfo` class to `attrs` and adds type hints and fixes up the usage around them.


synapse/rest/media/v1/_base.py
```diff
@@ -287,44 +289,25 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+    url_cache: Optional[str] = None
+    # Whether the file is a thumbnail or not.
+    thumbnail: bool = False
+    thumbnail_width: Optional[int] = None
+    thumbnail_height: Optional[int] = None
+    thumbnail_method: Optional[str] = None
+    # Content type of thumbnail, e.g. image/png
+    thumbnail_type: Optional[str] = None
+    # The size of the media file, in bytes.
+    thumbnail_length: Optional[int] = None
```
clokep, line 310: It would be much cleaner to group these attributes into a separate class that is optional on `FileInfo`, but this class is available to the spam checker and storage providers so we can't just change it willy-nilly.

synapse/rest/media/v1/_base.py
```diff
@@ -287,44 +289,25 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
-        self.thumbnail_type = thumbnail_type
-        self.thumbnail_length = thumbnail_length
+    """Details about a requested/uploaded file."""
+
+    # The server name where the media originated from, or None if local.
+    server_name: Optional[str]
+    # The local ID of the file. For local files this is the same as the media_id
+    file_id: str
+    # If the file is for the url preview cache
+    url_cache: Optional[str] = None
```
clokep, line 301: As far as I can tell `url_cache` is supposed to be a string. It is saved in the tables as `TEXT`, but we were previously sometimes setting it to a boolean. 🤷

synapse/rest/media/v1/_base.py
```diff
@@ -287,44 +289,25 @@ def __exit__(self, exc_type, exc_val, exc_tb):
         pass
 
 
+@attr.s(slots=True, frozen=True, auto_attribs=True)
```
clokep, line 292: I'm *fairly* certain this can be frozen.

synapse/rest/media/v1/_base.py
clokep: (Alternately if there's a way to tell attrs "if this attribute exists, so will these other ones" that would work better...)

synapse/rest/media/v1/_base.py
richvdh: I'm not seeing anywhere it is used as a string - am I blind? Indeed, as far as I can tell, the only place it is read is in `MediaStorage._file_info_to_path`, which just uses its boolean value.

There's a `url_cache` in `local_media_repository`, but it doesn't seem to be populated from a `FileInfo`.

synapse/rest/media/v1/_base.py
richvdh: well... you could use some @property methods to maintain compatibility for external modules, and raise an Exception if it's not a thumbnail.

Not sure it's worth it though.

richvdh (changes requested):

synapse/rest/media/v1/_base.py
clokep: We pull info from `local_media_repository` and then create a `FileInfo` in one spot:

https://github.com/matrix-org/synapse/blob/63f28e4a0c22c8133a6da175c7986e389fc781e2/synapse/rest/media/v1/media_repository.py#L213

Maybe that should be turning the `str` into a `bool` and the rest should be left alone?

clokep: It also seems to happen when responding with local thumbnails:

https://github.com/matrix-org/synapse/blob/63f28e4a0c22c8133a6da175c7986e389fc781e2/synapse/rest/media/v1/thumbnail_resource.py#L117

Which calls into:

https://github.com/matrix-org/synapse/blob/63f28e4a0c22c8133a6da175c7986e389fc781e2/synapse/rest/media/v1/thumbnail_resource.py#L346

richvdh: right yes, I think we should be casting to a bool in those places.

erikjohnston (approved):

Issue #1531, opened 2016-04-11
Enabling/disabling a unknown default push rule should 404. (SYN-676)


matrixbot:
```
PUT /_matrix/client/r0/pushrules/global/override/.not.a.default.rule/enabled

{"enabled":true}
```

The above request should fail with a 404.

(Imported from https://matrix.org/jira/browse/SYN-676)

(Reported by @NegativeMjark)

PR #10788, opened 2021-09-09
Remove fixed and flakey tests from the sytest blacklist

DMRobertson:
Method:

- [x] Comment out all blacklisted tests
- [x] Run tests
- [x] Comment out that ones that fail
- [x] Run tests
- [x] Check that everything passes
- [ ] Check the tests which are still commented out are genuinely fixed (i.e. corresponding issues are closed)
- [ ] Delete those tests
- [ ] ???
- [ ] Profit

### DIFF
--- a/.ci/worker-blacklist
+++ b/.ci/worker-blacklist
@@ -1,10 +1,10 @@
 # This file serves as a blacklist for SyTest tests that we expect will fail in
 # Synapse when run under worker mode. For more details, see sytest-blacklist.
 
-Can re-join room if re-invited
+# Can re-join room if re-invited
 
 # new failures as of https://github.com/matrix-org/sytest/pull/732
-Device list doesn't change if remote server is down
+# Device list doesn't change if remote server is down
 
 # https://buildkite.com/matrix-dot-org/synapse/builds/6134#6f67bf47-e234-474d-80e8-c6e1868b15c5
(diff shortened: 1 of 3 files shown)


DMRobertson:
I was quite optimistic, because many of these seemed not to fail. But reading around on github, I think the only one that has actually had a specific fix put in is `Enabling an unknown default rule fails with 404` from https://github.com/matrix-org/synapse/issues/1531. Perhaps many others are flakey?

DMRobertson:
The worker-specific blacklist history is [here](https://github.com/matrix-org/synapse/blame/05111f8f26252cc936fce685846322289039128d/.buildkite/worker-blacklist). (My rename of .buildkite to .ci obscured that.)

These all succeeded so I'm guessing they're just prone to flakiness?

DMRobertson:
> The worker-specific blacklist history is [here](https://github.com/matrix-org/synapse/blame/05111f8f26252cc936fce685846322289039128d/.buildkite/worker-blacklist). (My rename of .buildkite to .ci obscured that.)
> 
> These all succeeded so I'm guessing they're just prone to flakiness?

https://github.com/matrix-org/synapse/runs/3558256491 saw `Device list doesn't change if remote server is down` fail on 2/3 worker runs.

richvdh (approved):
thank you!

PR #10789, opened 2021-09-09
Easy refactors of the user directory

DMRobertson:
PRs #10695 and #10761 became unwieldy. It was particularly hard to review because they both mix functional changes with refactorings.

Let's take a slower approach. I've started with a bunch of straightfoward refactorings.

### DIFF
--- a/synapse/handlers/user_directory.py
+++ b/synapse/handlers/user_directory.py
@@ -142,7 +142,7 @@ class UserDirectoryHandler(StateDeltasHandler):
                 user_id, profile.display_name, profile.avatar_url
             )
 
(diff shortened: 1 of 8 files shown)


synapse/handlers/user_directory.py
```diff
@@ -292,7 +292,7 @@ async def _handle_room_publicity_change(
             raise Exception("Invalid event type")
         # If change is None, no change. True => become world_readable/public,
         # False => was world_readable/public
```
richvdh, line 294: this comment is no longer relevant.

richvdh (approved):
lgtm otherwise!

PR #10796, opened 2021-09-10
Always add local users to the user directory

DMRobertson:
This means the config option `search_all_users` ends up controlling the
search query only, and not the data we store. The cost of doing so is an
extra row in the `user_directory` and `user_directory_search` tables for
each local user in zero rooms. I think the cost of this will be
marginal (since they'll already have entries in `users` and `profiles`
anyway).

As a small upside, a homeserver whose directory was built with this
change can toggle `search_all_users` without having to rebuild their
directory.

### DIFF
--- a/synapse/storage/databases/main/user_directory.py
+++ b/synapse/storage/databases/main/user_directory.py
@@ -85,19 +85,17 @@ class UserDirectoryBackgroundUpdateStore(StateDeltasStore):
             self.db_pool.simple_insert_many_txn(txn, TEMP_TABLE + "_rooms", rooms)
             del rooms
 
-            # If search all users is on, get all the users we want to add.
-            if self.hs.config.user_directory_search_all_users:
-                sql = (
-                    "CREATE TABLE IF NOT EXISTS "
-                    + TEMP_TABLE
-                    + "_users(user_id TEXT NOT NULL)"
-                )
-                txn.execute(sql)
+            sql = (
+                "CREATE TABLE IF NOT EXISTS "
+                + TEMP_TABLE
+                + "_users(user_id TEXT NOT NULL)"
+            )
+            txn.execute(sql)
 
-                txn.execute("SELECT name FROM users")
-                users = [{"user_id": x[0]} for x in txn.fetchall()]
+            txn.execute("SELECT name FROM users")
+            users = [{"user_id": x[0]} for x in txn.fetchall()]
 
-                self.db_pool.simple_insert_many_txn(txn, TEMP_TABLE + "_users", users)
+            self.db_pool.simple_insert_many_txn(txn, TEMP_TABLE + "_users", users)
 
         new_pos = await self.get_max_stream_id_in_current_state_deltas()
         await self.db_pool.runInteraction(
@@ -265,13 +263,8 @@ class UserDirectoryBackgroundUpdateStore(StateDeltasStore):
 
     async def _populate_user_directory_process_users(self, progress, batch_size):
         """
-        If search_all_users is enabled, add all of the users to the user directory.
+        Add all local users to the user directory.
         """
-        if not self.hs.config.user_directory_search_all_users:
-            await self.db_pool.updates._end_background_update(
-                "populate_user_directory_process_users"
-            )
-            return 1
 
         def _get_next_batch(txn):
             sql = "SELECT user_id FROM %s LIMIT %s" % (
--- a/synapse/handlers/deactivate_account.py
+++ b/synapse/handlers/deactivate_account.py
@@ -257,11 +257,8 @@ class DeactivateAccountHandler(BaseHandler):
         """
         # Add the user to the directory, if necessary.
         user = UserID.from_string(user_id)
-        if self.hs.config.user_directory_search_all_users:
-            profile = await self.store.get_profileinfo(user.localpart)
-            await self.user_directory_handler.handle_local_profile_change(
-                user_id, profile
-            )
+        profile = await self.store.get_profileinfo(user.localpart)
+        await self.user_directory_handler.handle_local_profile_change(user_id, profile)
 
         # Ensure the user is not marked as erased.
         await self.store.mark_user_not_erased(user_id)
--- a/synapse/handlers/profile.py
+++ b/synapse/handlers/profile.py
@@ -214,11 +214,10 @@ class ProfileHandler(BaseHandler):
             target_user.localpart, displayname_to_set
         )
 
-        if self.hs.config.user_directory_search_all_users:
-            profile = await self.store.get_profileinfo(target_user.localpart)
-            await self.user_directory_handler.handle_local_profile_change(
-                target_user.to_string(), profile
-            )
+        profile = await self.store.get_profileinfo(target_user.localpart)
+        await self.user_directory_handler.handle_local_profile_change(
+            target_user.to_string(), profile
+        )
 
         await self._update_join_states(requester, target_user)
 
@@ -300,11 +299,10 @@ class ProfileHandler(BaseHandler):
             target_user.localpart, avatar_url_to_set
         )
 
-        if self.hs.config.user_directory_search_all_users:
-            profile = await self.store.get_profileinfo(target_user.localpart)
-            await self.user_directory_handler.handle_local_profile_change(
-                target_user.to_string(), profile
-            )
+        profile = await self.store.get_profileinfo(target_user.localpart)
+        await self.user_directory_handler.handle_local_profile_change(
+            target_user.to_string(), profile
+        )
 
         await self._update_join_states(requester, target_user)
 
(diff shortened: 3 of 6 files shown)


DMRobertson:
Worth adding: this changes the implementation but shouldn't have any user-visible effects. The intent is to simplify the logic which creates and maintains the user directory tables.

synapse/storage/databases/main/user_directory.py
```diff
@@ -85,19 +85,17 @@ def _make_staging_area(txn):
...
+            )
+            txn.execute(sql)
 
-                txn.execute("SELECT name FROM users")
-                users = [{"user_id": x[0]} for x in txn.fetchall()]
+            txn.execute("SELECT name FROM users")
+            users = [{"user_id": x[0]} for x in txn.fetchall()]
 
-                self.db_pool.simple_insert_many_txn(txn, TEMP_TABLE + "_users", users)
+            self.db_pool.simple_insert_many_txn(txn, TEMP_TABLE + "_users", users)
```
reivilibre, line 98: This isn't your fault, but since we're now going to start doing this for all servers: Is this going to be OK on big servers (like matrix.org?)?

Maybe we can use `INSERT INTO temp_users (user_id) SELECT name FROM users` to avoid having to pull it into Synapse only to shove it back out? (I checked this works on (recent) SQLite and Postgres. Hopefully we're not tripped up by older SQLites, but guess we could only do this for Postgres if there's a problem.)

I imagine the database would be smart enough to stream the data directly from one table to another (and save us buffering it into memory).

reivilibre (changes requested):
Looks good to me, *may* be worth checking that e.g. @richvdh thinks this is sensible, since now this is going straight to develop

synapse/storage/databases/main/user_directory.py
DMRobertson: I'm not sure how to answer that question without a big server to try it on. :( Worth noting that this is only at the point of rebuilding the directory, which is apparently a slow process anyway.

Another thought: we could `INSERT INTO user_directory` directly `FROM profiles`. We'd also need to `INSERT INTO user_directory_search`. And we'd need to consider what happens if someone adds to the profiles table in between the insertions. (Lock the profiles table?)

reivilibre: > And we'd need to consider what happens if someone adds to the profiles table in between the insertions. (Lock the profiles table?)

Locking sounds slightly no-no-y (depending on how long we're talking).

But when someone updates their profile, it goes into the directory anyway (or will when you add that change in)! :D

> Another thought: we could INSERT INTO user_directory directly FROM profiles

I do like the sound of that. I don't know how much overhaul it'd want from you and whether you'd want to do it or not.
(Just be careful: I don't think it's wise to try and do that in one go, because with concurrent writers to the `user_directory` table, it might happen that you get a serialisation failure in Postgres and have to start the transaction all over again.)

DMRobertson: > I don't know how much overhaul it'd want from you and whether you'd want to do it or not.

Not against doing it, but I'm not chomping at the bit to get that started now.

> But when someone updates their profile, it goes into the directory anyway (or will when you add that change in)! :D

Oh yeah, that's right.

> it might happen that you get a serialisation failure in Postgres and have to start the transaction all over again.

How would that manifest? A certain kind of error from psycopg2 that we'd need to catch and retry within the application? Do we have to handle that generally across synapse?

> with concurrent writers to the user_directory table, 

Erm, don't we already have concurrent writers to the `user_directory` table? I think the things competing for writes are:

- all of the stuff in the `UserDirectoryHandler`, which only ever runs on one worker
   - there's a scary looking comment in its docstring: `N.B.: ASSUMES IT IS THE ONLY THING THAT MODIFIES THE USER DIRECTORY`
- the background process in the `UserDirectoryBackgroundUpdateStore`
- anything that calls `UserDirectoryHandler.handle_local_profile_change`. (Register, Activate, Deactivate, Update Profile).

reivilibre: > How would that manifest? A certain kind of error from psycopg2 that we'd need to catch and retry within the application? Do we have to handle that generally across synapse?

Yup! `runInteraction` (or one of the things it calls) will automatically retry those up to 5 times (iirc), so your transaction might actually get run multiple times!

> Erm, don't we already have concurrent writers to the user_directory table?

Yes, and it's not too much hassle if they get restarted, because they're small transactions that only rarely collide. But if you have a single transaction writing (and potentially colliding with any one of) 100k+ rows, that's a *big* transaction to have to restart (and if it's slow enough and collision-y enough, might not succeed within 5 retries).

NB this is my understanding; I don't know all the subtleties about Postgres.
The relevant section of the manual is (I think we use this level) [REPEATABLE READ transaction isolation](https://www.postgresql.org/docs/13/transaction-iso.html#XACT-REPEATABLE-READ).

richvdh: > (I think we use this level) REPEATABLE READ transaction isolation.

That's correct, for the record. It gets set at https://github.com/matrix-org/synapse/blob/v1.42.0/synapse/storage/engines/postgres.py#L107-L109.

On the topic of this thread: slurping all the users into synapse just to spit them out again does seem *particularly* stupid and replacing it with a `INSERT INTO temp_users (user_id) SELECT name FROM users` might be a quick win.  And yes, no doubt you could go further in optimising this code too.

Still, I'd say that any such optimisations are orthogonal to this PR, and my inclination would be to leave it as is. Particularly given this code only runs when the directory is first populated (at which point it is reasonable to expect a small `users` table), I don't think we need to worry too much.

richvdh (approved):
seems fine to me.

One thing to note is that the text in the sample config (https://github.com/matrix-org/synapse/blob/v1.42.0/docs/sample_config.yaml#L2329-L2331) is no longer quite correct, so will need an update. I guess we'll have to say something like "Until Synapse 1.44, the indexes were not updated if this setting was False, so you may need to recreate the indexes".

Or maybe we should add a db migration to automatically rebuild the indexes?

synapse/storage/databases/main/user_directory.py
reivilibre: @richvdh I think we will wind up wanting to do a rebuild once all of David's work lands, to make sure that the user directory is scrubbed of private profiles. In that case, the fact that big servers such as `matrix.org` now need to hit this code, I think it would be worthwhile doing at least that.
(I agree that this could easily be a tiny separate PR, though.)

DMRobertson: I'll make a note to revisit this in a future PR.

synapse/config/user_directory.py
```diff
@@ -45,12 +45,16 @@ def generate_config_section(self, config_dir_path, server_name, **kwargs):
...
+            # visible in public rooms and users sharing a room with the requester.
+            # Defaults to false.
             #
-            # If you set it true, you'll have to rebuild the user_directory search
-            # indexes, see:
-            # https://matrix-org.github.io/synapse/latest/user_directory.html
+            # NB. If you set this to true, and the last time the user_directory search
+            # indexes were (re)built was before Synapse 1.44, you'll have to
+            # rebuild the indexes in order to search through all known users.
+            # These indexes are build the first time Synapse starts; admins can
```
richvdh, line 55:
```suggestion
            # These indexes are built the first time Synapse starts; admins can
```

reivilibre, line 55:
```suggestion
            # indices were (re)built was before Synapse 1.44, you'll have to
            # rebuild the indices in order to search through all known users.
            # These indices are built the first time Synapse starts; admins can
```

richvdh (approved):

reivilibre (approved):
looks fine and dandy to me

docs/sample_config.yaml
```diff
@@ -2362,12 +2362,16 @@ user_directory:
...
+    # visible in public rooms and users sharing a room with the requester.
+    # Defaults to false.
     #
-    # If you set it true, you'll have to rebuild the user_directory search
-    # indexes, see:
-    # https://matrix-org.github.io/synapse/latest/user_directory.html
+    # NB. If you set this to true, and the last time the user_directory search
+    # indexes were (re)built was before Synapse 1.44, you'll have to
+    # rebuild the indexes in order to search through all known users.
+    # These indexes are build the first time Synapse starts; admins can
```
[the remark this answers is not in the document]
DMRobertson: `indices` was my instinct too, but "indexes" seems to be slightly more common. No strong opinions personally, but FWIW

- https://www.postgresql.org/docs/current/indexes.html
- https://www.postgresql.org/message-id/8182.986230075@sss.pgh.pa.us
- [Old versions of postgres seemed unable to make their mind up](https://www.postgresql.org/docs/7.0/indices.htm).

(of course, postgres doesn't have a monopoly on the concept of a db index)

PR #10798, opened 2021-09-10
Use direct references for some configuration variables

clokep:
This uses direct references to config flags instead of letting the `Config` object resolve it.

I have a cunning plan of getting rid of the magic in that class...

The changes the flags that were fairly prevalent and easy to change with sed (+ black).

I don't think any of these are in hot enough paths that they should matter much, but theoretically it is more performant.

Should be reviewable commit-by-commit (although it isn't bad to just look at it all either...)

### DIFF
--- a/synapse/python_dependencies.py
+++ b/synapse/python_dependencies.py
@@ -107,7 +107,7 @@ CONDITIONAL_REQUIREMENTS = {
     "systemd": ["systemd-python>=231"],
     "url_preview": ["lxml>=3.5.0"],
     "sentry": ["sentry-sdk>=0.7.2"],
-    "opentracing": ["jaeger-client>=4.0.0,<4.7.0", "opentracing>=2.2.0"],
+    "opentracing": ["jaeger-client>=4.0.0", "opentracing>=2.2.0"],
     "jwt": ["pyjwt>=1.6.4"],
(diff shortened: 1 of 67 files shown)


synapse/python_dependencies.py
```diff
@@ -107,7 +107,7 @@
     "systemd": ["systemd-python>=231"],
     "url_preview": ["lxml>=3.5.0"],
     "sentry": ["sentry-sdk>=0.7.2"],
-    "opentracing": ["jaeger-client>=4.0.0", "opentracing>=2.2.0"],
+    "opentracing": ["jaeger-client>=4.0.0,<4.7.0", "opentracing>=2.2.0"],
```
anoadragon453, line 110: Looks like this change was accidentally included?

anoadragon453 (commented):
Thank you for this! I've been wanting to do this for so long...

> I have a cunning plan of getting rid of the magic in that class...

Note that this will probably break a bunch of third-party libraries that reach in to the depths of Synapse, but on the other hand it may be a good way to root out misbehaving modules :)

synapse/python_dependencies.py
clokep: Oh, I needed to back this out. Good call.

anoadragon453 (approved):

PR #10799, opened 2021-09-10
Use  the BaseReporter super-class for _WrappedRustReporter.

clokep:
Add a max version for the `jaeger-client` dependency for an incompatibility with the rust reporter.

Without this you get an error:

```
synapse/logging/opentracing.py:385: error: Argument 1 to "create_tracer" of "Config" has incompatible type "_WrappedRustReporter"; expected "BaseReporter"  [arg-type]
synapse/logging/opentracing.py:385: error: Argument 2 to "create_tracer" of "Config" has incompatible type "Optional[Sampler]"; expected "Sampler"  [arg-type]
```

### DIFF
--- a/synapse/logging/opentracing.py
+++ b/synapse/logging/opentracing.py
@@ -236,8 +236,16 @@ except ImportError:
 try:
     from rust_python_jaeger_reporter import Reporter
 
+    # jaeger-client 4.7.0 requires that reporters inherit from BaseReporter, which
+    # didn't exist before that version.
+    try:
+        from jaeger_client.reporter import BaseReporter
+    except ImportError:
+        class BaseReporter:  # type: ignore[no-redef]
+            pass
+
     @attr.s(slots=True, frozen=True)
-    class _WrappedRustReporter:
+    class _WrappedRustReporter(BaseReporter):
         """Wrap the reporter to ensure `report_span` never throws."""
 
         _reporter = attr.ib(type=Reporter, default=attr.Factory(Reporter))
@@ -382,6 +390,7 @@ def init_tracer(hs: "HomeServer"):
     # If we have the rust jaeger reporter available let's use that.
     if RustReporter:
         logger.info("Using rust_python_jaeger_reporter library")
+        assert config.sampler is not None
         tracer = config.create_tracer(RustReporter(), config.sampler)
         opentracing.set_global_tracer(tracer)
     else:
--- a/synapse/logging/opentracing.py
+++ b/synapse/logging/opentracing.py
@@ -241,6 +241,7 @@ try:
     try:
         from jaeger_client.reporter import BaseReporter
     except ImportError:
+
         class BaseReporter:  # type: ignore[no-redef]
             pass
 
--- a/synapse/python_dependencies.py
+++ b/synapse/python_dependencies.py
@@ -107,7 +107,7 @@ CONDITIONAL_REQUIREMENTS = {
     "systemd": ["systemd-python>=231"],
     "url_preview": ["lxml>=3.5.0"],
     "sentry": ["sentry-sdk>=0.7.2"],
-    "opentracing": ["jaeger-client>=4.0.0,<4.7.0", "opentracing>=2.2.0"],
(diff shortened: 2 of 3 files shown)


clokep:
A failing build: https://github.com/matrix-org/synapse/runs/3569940997

I have no idea if this is the "right" fix, but 🤷

synapse/python_dependencies.py
```diff
@@ -107,7 +107,7 @@
     "systemd": ["systemd-python>=231"],
     "url_preview": ["lxml>=3.5.0"],
     "sentry": ["sentry-sdk>=0.7.2"],
-    "opentracing": ["jaeger-client>=4.0.0", "opentracing>=2.2.0"],
+    "opentracing": ["jaeger-client>=4.0.0,<4.7.0", "opentracing>=2.2.0"],
```
DMRobertson, line 110: [The 4.7 release](https://github.com/jaegertracing/jaeger-client-python/releases/tag/4.7.0) [requires Python 3.7+](https://github.com/jaegertracing/jaeger-client-python/pull/325), and I think we're still supporting 3.6, so this is a no-go anyway. Might be worth leaving a comment to explain why we need that max version?

As for whypy mypy fails, I'm not sure. [They run mypy as part of `make lint`](https://github.com/jaegertracing/jaeger-client-python/blob/34922a8546620045151756b583b3c201e4929601/Makefile#L72), which in turn gets run by `make test_ci` and that is run by GHA.

Maybe the two projects have different mypy config?

DMRobertson (approved):

synapse/python_dependencies.py
clokep: > [The 4.7 release](https://github.com/jaegertracing/jaeger-client-python/releases/tag/4.7.0) [requires Python 3.7+](https://github.com/jaegertracing/jaeger-client-python/pull/325)

If they have their package setup correctly Python 3.6 should install <4.7.0 anyway...

I'll add a comment though.

clokep: > As for whypy mypy fails, I'm not sure. [They run mypy as part of `make lint`](https://github.com/jaegertracing/jaeger-client-python/blob/34922a8546620045151756b583b3c201e4929601/Makefile#L72), which in turn gets run by `make test_ci` and that is run by GHA.
> 
> Maybe the two projects have different mypy config?

I'm confused by this -- I suspect our mypy is failing since they added type hints which mypy is now finding? But I'm not 100% sure.

DMRobertson: Ahh, I didn't realise the rust reporter was our own thing!

clokep:
@DMRobertson I took a different approach, which I think will be a bit more flexible. What do you think?

synapse/logging/opentracing.py
```diff
@@ -236,8 +236,16 @@ class _DummyTagNames:
 try:
     from rust_python_jaeger_reporter import Reporter
 
+    # jaeger-client 4.7.0 requires that reporters inherit from BaseReporter, which
+    # didn't exist before that version.
```
DMRobertson, line 240: Ahh, I didn't realise this was new in 4.7.

(Is the inheritance you describe a runtime requirement, or just part of the type annotations added in 4.7?)

DMRobertson (approved):
That feels neater! lgtm assuming CI is happy.

General thought: we saw this because we were running mypy with a new enough python version to support jaeger-client 4.7. Should we be running mypy against a variety of Python versions?

synapse/logging/opentracing.py
clokep: Based on jaegertracing/jaeger-client-python#329 I think it is really just there for type-annotations.

It does seem that they consolidated some logic into it. (The `close` method as you noticed.) But it seems to essentially be a no-op so I don't think this is huge.

clokep:
> General thought: we saw this because we were running mypy with a new enough python version to support jaeger-client 4.7. Should we be running mypy against a variety of Python versions?

IIRC our previous thought on this was to run the latest Python we support (+ the latest of each package), but I'm not sure that has been thoroughly thought through. 😄

Issue #10270, opened 2021-06-29
Sending a membership event to unban a user results in "Cannot leave user who was banned"


turt2live:
instead, the caller needs to use the `/unban` API. Sending `m.room.member` directly does not work, when it should.

PR #10807, opened 2021-09-11
Allow sending a membership event to unban a user 

aaronraimist:
Fixes https://github.com/matrix-org/synapse/issues/10270

### Pull Request Checklist


aaronraimist:
This is how the spec says it should work. However this would reintroduce https://github.com/vector-im/element-web/issues/1961 so maybe needs some thought at the spec level.

anoadragon453:
Note that the Spec offers two APIs that are relevant here:

* `/_matrix/client/r0/rooms/{roomId}/kick`
* `/_matrix/client/r0/rooms/{roomId}/unban`

The Spec doesn't seem to say anywhere that calling `/kick` should necessarily unban a user. However it does make sense that manually sending a `m.room.member` event with membership `leave` for the user should unban them.

To fix both [the issue mentioned on the Element Web side](https://github.com/matrix-org/matrix-js-sdk/pull/364/files#diff-627fe08d8bba6a4fa76e7c5ed36ef6f16d126733e65e19236b391de0d34f1fe0R1387-R1391) (where one user kicking a user and another banning them may race and result in a ban -> leave [unbanned] state), and allow more advanced use-cases, perhaps we could modify Synapse such that:

* `/kick` on a banned user does not work.
* `/unban` on a banned user will unban them.
* Sending a custom `m.room.members` with membership `leave` for a banned user will unban them.

That's effectively what this PR does, besides excluding `"kick"` from the list of actions that won't raise a `SynapseError`.

anoadragon453:
@aaronraimist I think you may need to merge `develop` into your branch in order to fix the CI errors.

aaronraimist:
Okay I'll look in to that soon.

Element will need a change as well though. Kicking in Element doesn't use the /kick API, it just sends a membership event.

richvdh (approved):
lgtm. Thanks!

[it'd be nice to have system tests for this in Complement, to make sure other server implementations don't fall into the same trap]

Issue #10702, opened 2021-08-26
matrix.org's sentry is full of logcontext errors


richvdh:
See https://sentry.matrix.org/sentry/synapse-matrixorg/issues/15805, for example. They mostly seem to be due to varanustra, though that should have been fixed in https://gitlab.matrix.org/new-vector/varunastra/-/merge_requests/20, so we need to investigate why it's still full of spam.

callahad:
Link to contextmanager issues in Sentry: https://sentry.matrix.org/sentry/synapse-matrixorg/issues/?query=logger%3Asynapse.logging.scopecontextmanager

Seems to have changed profile recently; needs further investigation

squahtx:
The errors we're seeing in Sentry are:
* "Tried to activate scope outside of loggingcontext", which stopped on September 1st
* "Tried to close a non-active scope!", which started on September 1st

These could very well be caused by logging contexts being lost / not being correctly restored.

@richvdh has pointed out that we have a known issue where this can happen during federation send timeouts: https://github.com/matrix-org/synapse/issues/5604

@anoadragon453 has found that synapse-spamcheck-badlist causes logging contexts to be lost, which we have a fix for: https://github.com/matrix-org/synapse-spamcheck-badlist/pull/24

I've opened a PR to try to prevent logging contexts going missing during federation send timeouts: https://github.com/matrix-org/synapse/pull/10810

squahtx:
The spam has disappeared now that https://github.com/matrix-org/synapse-spamcheck-badlist/pull/24 and #10810 have been deployed.

Thanks to @anoadragon453 and @reivilibre for pointing me in the right direction!

richvdh:
yay, thanks for sorting this!

PR #10810, opened 2021-09-13
Prevent logging context going missing on federation request timeout

squahtx:
In `MatrixFederationHttpClient._send_request()`, we make a HTTP request
using an `Agent`, wrap that request in a timeout and await the resulting
`Deferred`. On its own, the `Agent` performing the HTTP request
correctly stashes and restores the logging context while waiting.
The addition of the timeout introduces a path where the logging context
is not restored when execution resumes.

To address this, we wrap the timeout `Deferred` in a
`make_deferred_yieldable()` to stash the logging context and restore it
on completion of the `await`. However this is not sufficient, since by
the time we construct the timeout `Deferred`, the `Agent` has already
stashed and cleared the logging context when using
`make_deferred_yieldable()` to produce its `Deferred` for the request.

Hence, we wrap the `Agent` request in a `run_in_background()` to "fork"
and preserve the logging context so that we can stash and restore it
when `await`ing the timeout `Deferred`.

Note that the code is still not fully correct. When a timeout occurs,
the request remains running in the background (existing behavior which
is nothing to do with the new call to `run_in_background`) and may
re-start the logging context after it has finished.

### DIFF
--- a/synapse/http/matrixfederationclient.py
+++ b/synapse/http/matrixfederationclient.py
@@ -553,6 +553,15 @@ class MatrixFederationHttpClient:
                         with Measure(self.clock, "outbound_request"):
                             # we don't want all the fancy cookie and redirect handling
                             # that treq.request gives: just use the raw Agent.
+
+                            # To preserve the logging context, the timeout is treated
+                            # in a similar way to `defer.gatherResults`:
+                            # * Each logging context-preserving fork is wrapped in
+                            #   `run_in_background`. In this case there is only one,
+                            #   since the timeout fork is not logging-context aware.
+                            # * The `Deferred` that joins the forks back together is
+                            #   wrapped in `make_deferred_yieldable` to restore the
+                            #   logging context regardless of the path taken.
                             request_deferred = run_in_background(
                                 self.agent.request,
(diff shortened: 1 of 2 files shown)


synapse/http/matrixfederationclient.py
```diff
@@ -553,20 +553,20 @@ async def _send_request(
                         with Measure(self.clock, "outbound_request"):
                             # we don't want all the fancy cookie and redirect handling
                             # that treq.request gives: just use the raw Agent.
-                            request_deferred = self.agent.request(
+                            request_deferred = run_in_background(
```
anoadragon453, line 556: It may be worth adding a comment here and below as to why `run_in_background` and `make_deferred_yieldable` are necessary.

anoadragon453 (approved):
Lgtm otherwise, thanks for finding an fixing these!

anoadragon453 (approved):
lgtm! thanks!

squahtx:
CI seems to be failing because of https://github.com/matrix-org/sytest/issues/1136 , merging anyway

squahtx:
This PR appears to have resolved #10702.

PR #10812, opened 2021-09-13
Use direct references for some configuration variables (part 2)

clokep:
This builds on #10798 to move more references over.

The first commit is done with sed + black, the rest were done by inspection / grep.

### DIFF
--- a/synapse/push/pusherpool.py
+++ b/synapse/push/pusherpool.py
@@ -62,7 +62,7 @@ class PusherPool:
         self.clock = self.hs.get_clock()
 
         # We shard the handling of push notifications by user ID.
-        self._pusher_shard_config = hs.config.push.pusher_shard_config
+        self._pusher_shard_config = hs.config.worker.pusher_shard_config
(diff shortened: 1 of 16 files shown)


erikjohnston (approved):
I'm trusting that you've got the write section for each config, but they all seem plausible

synapse/push/pusherpool.py
```diff
@@ -62,7 +62,7 @@ def __init__(self, hs: "HomeServer"):
         self.clock = self.hs.get_clock()
 
         # We shard the handling of push notifications by user ID.
-        self._pusher_shard_config = hs.config.push.pusher_shard_config
+        self._pusher_shard_config = hs.config.worker.pusher_shard_config
```
clokep, line 65: I'm surprised this one used to work since it was specific, but pointing to the *wrong* one?

clokep:
> I'm trusting that you've got the write section for each config, but they all seem plausible

I've double checked them. :)

PR #10814, opened 2021-09-13
Refactor oEmbed previews

clokep:
This reworks how we do oEmbed quite a bit, with an eye towards being able to support auto-discovery and other features. It attempts to fit the oEmbed previews a bit better into the preview framework by moving it to a higher level. This allows for us re-using more of the framework we use for generating previews.

* There's a few commits that refactor things to split out methods / move some code around.
* Move oEmbed higher up in the process chain -- use the standard `_download_file` method and then post-process the result (like we do for HTML).

This also updates the tests to be a bit more reasonable.

Related to #2752.


clokep:
I think this is ready for review. There's currently two other PRs built on this that might be useful to look at where this is heading (and why we want to refactor this code).

synapse/rest/media/v1/oembed.py
```diff
@@ -12,30 +12,26 @@
...
 
 
-@attr.s(slots=True, auto_attribs=True)
+@attr.s(slots=True, frozen=True, auto_attribs=True)
 class OEmbedResult:
-    # Either HTML content or URL must be provided.
-    html: Optional[str]
-    url: Optional[str]
-    title: Optional[str]
+    og: JsonDict
```
reivilibre, line 32: this name is a little bit cryptic; suggest `opengraph` or ideally something that actually says what it is?

`opengraph_metadata`? I don't really have a good handle of what opengraph is, so you may know better.

tests/rest/media/v1/test_url_preview.py
```diff
@@ -576,11 +577,10 @@ def test_oembed_photo(self):
...
 
-        end_content = (
-            b"<html><head>"
-            b"<title>Some Title</title>"
-            b'<meta property="og:description" content="hi" />'
-            b"</head></html>"
+        end_content = unhexlify(
+            b"89504e470d0a1a0a0000000d4948445200000001000000010806"
+            b"0000001f15c4890000000a49444154789c63000100000500010d"
+            b"0a2db40000000049454e44ae426082"
```
reivilibre, line 583: ?
I feel as though it was clearer before.
I guess you had a reason to do this?
If sticking with the hex-encoded gobbledygook, please can we have a comment saying what's inside it?

reivilibre: Oh actually I guess this is an image with binary content, judging by the null bytes. Mind popping a comment on saying what it is (e.g.: small GIF image, small PNG image, whatever it is?). :)

reivilibre (commented):
I don't think I know enough to review this properly, but a couple of comments

synapse/rest/media/v1/oembed.py
clokep: I agree, but this is what is used throughout this code (which is why I used it here).

tests/rest/media/v1/test_url_preview.py
clokep: Yeah I copied it from somewhere else in the tests, I should probably make a constant for it and re-use it.

clokep: See #10834 for factoring this out. I did a separate PR since it seem unrelated enough.

synapse/rest/media/v1/oembed.py
clokep: Oh, also the documentation (which is updated by this PR) does include some information on Open Graph, but pretty much it is a JSON representation of metadata about a resource.

richvdh (commented):
looks generally great! a few comments.

synapse/rest/media/v1/preview_url_resource.py
```diff
@@ -292,6 +298,22 @@ async def _do_preview(self, url: str, user: str, ts: int) -> bytes:
...
+                oembed_body = utf8_file.read()
+
+            oembed_response = self._oembed.parse_oembed_response(
+                media_info.uri, oembed_body
+            )
+            og = oembed_response.og
+
+            # Use the cache age from the oEmbed result, instead of the HTTP response.
+            if oembed_response.cache_age is not None:
+                expiration_ts_ms = oembed_response.cache_age + media_info.created_ts_ms
```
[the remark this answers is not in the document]
clokep: Yes, we definitely should. Do you have any thoughts for a max-age? 24 hours? A week?

synapse/rest/media/v1/preview_url_resource.py
```diff
@@ -292,6 +298,22 @@ async def _do_preview(self, url: str, user: str, ts: int) -> bytes:
 
             await self._precache_image_url(user, media_info, og)
 
+        elif oembed_url and _is_json(media_info.media_type):
+            # Handle an oEmbed response.
+            with open(media_info.filename, "r") as utf8_file:
+                oembed_body = utf8_file.read()
```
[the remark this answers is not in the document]
clokep: It *should* be UTF-8 encoded JSON, I guess the headers could say they're something else, but that seems like it would be extremely rare? Maybe that's an invalid assumption though!

synapse/rest/media/v1/preview_url_resource.py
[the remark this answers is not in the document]
clokep: I went with 24 hours, but should be easy enough to change. I also chose to apply this to all previews, not just ones from oEmbed. Although other previews currently always use 1 hour:

https://github.com/matrix-org/synapse/blob/9ffa787eb243c98a6ca1ecd9eac4a6b5dac2bef0/synapse/rest/media/v1/preview_url_resource.py#L418-L420

synapse/rest/media/v1/preview_url_resource.py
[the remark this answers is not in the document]
clokep: I could also load this as bytes and push the `UTF-8` conversion into the oEmbed code, if you'd like? At least that way the exception should be caught more nicely.

clokep:
A couple of outstanding questions so I put it back on your review queue @richvdh!

synapse/rest/media/v1/oembed.py
```diff
@@ -32,6 +32,9 @@ class OEmbedResult:
     # The Open Graph result (converted from the oEmbed result).
     open_graph_result: JsonDict
     # Number of seconds to cache the content.
```
richvdh, line 34:
```suggestion
    # Number of seconds to cache the content, according to the oEmbed response
```

richvdh (approved):
lgtm

PR #10815, opened 2021-09-14
Name the type of token in "Invalid token" messages

DMRobertson:
I had one of these error messages yesterday and assumed it was an
invalid auth token (because that was an HTTP query parameter in the
test) I was working on. In fact, it was an invalid next batch token for
syncing.


babolivier (approved):

DMRobertson:
(CI failed some sytests only, known flakyness)

PR #10817, opened 2021-09-14
Add missing type hints to rest servlets.

clokep:
Expands the work done in  #10665 and follow-ons to do all of `synapse.rest` package. This now enforces type hints in this package.

Builds on #10785, should be reviewable commit-by-commit.

### DIFF
--- a/synapse/rest/consent/consent_resource.py
+++ b/synapse/rest/consent/consent_resource.py
@@ -17,16 +17,21 @@ import logging
 from hashlib import sha256
 from http import HTTPStatus
 from os import path
-from typing import Dict, List
+from typing import TYPE_CHECKING, Any, Dict, List
 
 import jinja2
 from jinja2 import TemplateNotFound
+
+from twisted.web.server import Request
 
 from synapse.api.errors import NotFoundError, StoreError, SynapseError
 from synapse.config import ConfigError
 from synapse.http.server import DirectServeHtmlResource, respond_with_html
 from synapse.http.servlet import parse_bytes_from_args, parse_string
 from synapse.types import UserID
+
+if TYPE_CHECKING:
+    from synapse.server import HomeServer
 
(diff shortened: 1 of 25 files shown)


synapse/rest/consent/consent_resource.py
```diff
@@ -106,18 +107,14 @@ def __init__(self, hs):
...
-        """
+    async def _async_render_GET(self, request: Request) -> None:
         version = parse_string(request, "v", default=self._default_consent_version)
         username = parse_string(request, "u", default="")
         userhmac = None
         has_consented = False
         public_version = username == ""
         if not public_version:
-            args: Dict[bytes, List[bytes]] = request.args
+            args: Dict[bytes, List[bytes]] = request.args  # type: ignore
```
erikjohnston, line 117: Why do we need to ignore the type here ooi?

clokep: We do this in many places -- Twisted has the wrong type for `request.args`.

clokep: Ah, actually it isn't wrong, it's just missing. Without this we get:

```
synapse/rest/consent/consent_resource.py:118: error: Argument 1 to "parse_bytes_from_args" has incompatible type "Optional[Any]"; expected "Mapping[bytes, Sequence[bytes]]"  [arg-type]
```

erikjohnston: I have a feeling this might be the 154312th time I've asked that Q....

erikjohnston (approved):
I thought I had hit the approve button last time...

synapse/rest/consent/consent_resource.py
clokep: > I have a feeling this might be the 154312th time I've asked that Q....

That's OK! I think it might be fixed in the next Twisted. I should double check / fix it.

PR #10819, opened 2021-09-14
Improve oEmbed previews

clokep:
* Better titles (fall back to the author name if there's not title).
* Parse photo/video payloads.
* Include the original URL in the Open Graph response.
* Uses the proper expiration time (the cache age was not being converted from seconds to milliseconds).

~~Builds on #10814~~

Related to #2752, #9733

## Open questions:

* Does it make sense to include YouTube in our list of default oEmbed providers included with Synapse?
		* With this change, YouTube previews work when using oEmbed, but they aren't as good as parsing the HTML directly (you end up with no description). This is better than having no preview, as reported in #9733, but I've been unable to reproduce that.
		* Based on response in #9733, I don't think it makes sense to.

### Before this PR

Screen Shot 2021-09-21 at 1 00 30 PM

### After this PR

Screen Shot 2021-09-21 at 12 59 10 PM

Note the added title (`Matrix - Twitter`).

### Without oEmbed

Screen Shot 2021-09-21 at 12 56 47 PM

### After this PR

Screen Shot 2021-09-21 at 12 55 41 PM

### DIFF
--- a/synapse/rest/media/v1/oembed.py
+++ b/synapse/rest/media/v1/oembed.py
@@ -33,7 +33,7 @@ logger = logging.getLogger(__name__)
 class OEmbedResult:
     # The Open Graph result (converted from the oEmbed result).
     open_graph_result: JsonDict
-    # Number of seconds to cache the content, according to the oEmbed response.
+    # Number of milliseconds to cache the content, according to the oEmbed response.
     #
     # This will be None if no cache-age is provided in the oEmbed response (or
     # if the oEmbed response cannot be turned into an Open Graph response).
@@ -121,7 +121,7 @@ class OEmbedProvider:
             # Ensure the cache age is None or an int.
             cache_age = oembed.get("cache_age")
             if cache_age:
-                cache_age = int(cache_age)
+                cache_age = int(cache_age) * 1000
 
             # The results.
             open_graph_response = {
--- a/synapse/rest/media/v1/oembed.py
+++ b/synapse/rest/media/v1/oembed.py
@@ -13,7 +13,7 @@
 #  limitations under the License.
 import logging
 import urllib.parse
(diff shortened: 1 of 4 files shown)


synapse/rest/media/v1/oembed.py
```diff
@@ -119,10 +121,22 @@ def parse_oembed_response(self, url: str, raw_body: bytes) -> OEmbedResult:
...
             cache_age = oembed.get("cache_age")
             if cache_age:
-                cache_age = int(cache_age)
+                cache_age = int(cache_age) * 1000
 
             # The results.
-            open_graph_response = {"og:title": oembed.get("title")}
+            open_graph_response = {
+                "og:url": url,
+            }
```
squahtx, line 129: Won't this be the oEmbed URL, rather than the original? We seem to be passing the wrong URL in `_do_preview`

clokep: Quite possibly.

clokep: Thanks! I've fixed that and updated the tests!

squahtx (approved):
LGTM!

It's a shame that youtube previews look so sparse on the official client. It might be good to check how other messaging clients deal with youtube links. IIRC some don't include the description at all and let you play the video inline.

clokep:
> It's a shame that youtube previews look so sparse on the official client. It might be good to check how other messaging clients deal with youtube links. IIRC some don't include the description at all and let you play the video inline.

I agree, but from the discussion in #9733 it seems better to use the HTML preview now anyway.

This is an improvement, regardless, as it properly pulls the title / site name and some other bits and pieces. I'm also sure there are other sites which do `video` oEmbed responses which are good to support!

Thank you for the review! 🎉

PR #10820, opened 2021-09-14
Replace "\u0000" with "\u0020" before inserting into database

### LINKED ISSUE #9341: Internal error when posting an event with \u0000
### Description

Although `"\u0000"` is a perfectly valid JSON string, it seems to fail when storing in the database.

### Steps to reproduce

```
HS_URL=
ROOM_ID=
ACCESS_TOKEN=
curl "https://$HS_URL/_matrix/client/r0/rooms/$ROOM_ID/send/m.room.message" -X POST H 'Content-Type: application/json' -H "Authorization: Bearer $ACCESS_TOKEN" --data-raw '{"msgtype":"m.text","body":"\u0000"}'
```

Error seen in the log file:
```
  File "/opt/venvs/matrix-synapse/lib/python3.7/site-packages/synapse/storage/persist_events.py", line 172, in handle_queue_loop
    ret = await per_item_callback(item)
  File "/opt/venvs/matrix-synapse/lib/python3.7/site-packages/synapse/storage/persist_events.py", line 325, in persisting_queue
    item.events_and_contexts, backfilled=item.backfilled
  File "/opt/venvs/matrix-synapse/lib/python3.7/site-packages/synapse/storage/persist_events.py", line 537, in _persist_events
    backfilled=backfilled,
  File "/opt/venvs/matrix-synapse/lib/python3.7/site-packages/synapse/storage/databases/main/events.py", line 177, in _persist_events_and_state_updates
  File "/opt/venvs/matrix-synapse/lib/python3.7/site-packages/synapse/storage/database.py", line 664, in runInteraction
    **kwargs,
  File "/opt/venvs/matrix-synapse/lib/python3.7/site-packages/synapse/storage/database.py", line 740, in runWithConnection
    self._db_pool.runWithConnection(inner_func, *args, **kwargs)
  File "/opt/venvs/matrix-synapse/lib/python3.7/site-packages/twisted/python/threadpool.py", line 250, in inContext
    result = inContext.theWork()
  File "/opt/venvs/matrix-synapse/lib/python3.7/site-packages/twisted/python/threadpool.py", line 266, in <lambda>
    inContext.theWork = lambda: context.call(ctx, func, *args, **kw)
  File "/opt/venvs/matrix-synapse/lib/python3.7/site-packages/twisted/python/context.py", line 122, in callWithContext
  File "/opt/venvs/matrix-synapse/lib/python3.7/site-packages/twisted/python/context.py", line 85, in callWithContext
    return func(*args,**kw)
  File "/opt/venvs/matrix-synapse/lib/python3.7/site-packages/twisted/enterprise/adbapi.py", line 306, in _runWithConnection
  File "/opt/venvs/matrix-synapse/lib/python3.7/site-packages/twisted/python/compat.py", line 464, in reraise
    raise exception.with_traceback(traceback)
  File "/opt/venvs/matrix-synapse/lib/python3.7/site-packages/twisted/enterprise/adbapi.py", line 297, in _runWithConnection
  File "/opt/venvs/matrix-synapse/lib/python3.7/site-packages/synapse/storage/database.py", line 734, in inner_func
    return func(db_conn, *args, **kwargs)
  File "/opt/venvs/matrix-synapse/lib/python3.7/site-packages/synapse/storage/database.py", line 534, in new_transaction
  File "/opt/venvs/matrix-synapse/lib/python3.7/site-packages/synapse/logging/utils.py", line 71, in wrapped
    return f(*args, **kwargs)
  File "/opt/venvs/matrix-synapse/lib/python3.7/site-packages/synapse/storage/databases/main/events.py", line 394, in _persist_events_txn
    backfilled=backfilled,
  File "/opt/venvs/matrix-synapse/lib/python3.7/site-packages/synapse/storage/databases/main/events.py", line 1362, in _update_metadata_tables_txn
    self._store_room_message_txn(txn, event)
  File "/opt/venvs/matrix-synapse/lib/python3.7/site-packages/synapse/storage/databases/main/events.py", line 1653, in _store_room_message_txn
    txn, event, "content.body", event.content["body"]
  File "/opt/venvs/matrix-synapse/lib/python3.7/site-packages/synapse/storage/databases/main/events.py", line 1707, in store_event_search_txn
    origin_server_ts=event.origin_server_ts,
  File "/opt/venvs/matrix-synapse/lib/python3.7/site-packages/synapse/storage/databases/main/search.py", line 66, in store_search_entries_txn
    txn.executemany(sql, args)
  File "/opt/venvs/matrix-synapse/lib/python3.7/site-packages/synapse/storage/database.py", line 291, in executemany
    self._do_execute(self.txn.executemany, sql, *args)
  File "/opt/venvs/matrix-synapse/lib/python3.7/site-packages/synapse/storage/database.py", line 314, in _do_execute
    return func(sql, *args)
ValueError: A string literal cannot contain NUL (0x00) characters.
```

<!--

``` (three backticks, on a line on their own), so that they are formatted legibly.
-->

### Version information

- **Homeserver**: cloud.cervoi.se

If not matrix.org:

<!--

listen on that port.)
-->
- **Version**: 1.26.0

- **Install method**: apt

- **Platform**: debian 10
<!--
-->

### PR DESCRIPTION
**Pull request by H-Shay**
This PR intercepts message bodies with "\u0000" in them and replaces them with "\u0020" before inserting into the database, which causes an error in Postgres databases. Fixes #9341. 

### Pull Request Checklist

* [ x] Pull request is based on the develop branch
* [ x] Pull request includes a [changelog file](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md#changelog). The entry should:
* [ x] Pull request includes a [sign off](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md#sign-off)
* [ x] Code style is correct (run the [linters](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md#code-style))

### CODE REVIEW

**Comment by ShadowJonathan** [comment]
I think this is a *very* bad idea, as it can corrupt data that legitimately contain `\u0000` for string data, which is then malformed after it gets sent to synapse, leading to untracable errors, and most importantly, non-compliance with the spec.

I think that all other avenues, such as making sure postgres can properly ingest the data, should be exhausted, before going to this end.

**Comment by clokep** [comment]
> This PR intercepts message bodies with "\u0000" in them and replaces them with "\u0020" before inserting into the database, which causes an error in Postgres databases. Fixes #9341

To be clear -- this *only* does it for the search tables, correct? Not the original event content?

**Comment by ShadowJonathan** [comment]
Ah, yeah, it does. Sorry, i was mistaken and misled that this was about storing event JSON, if it's about NULL character bytes that have to be parsed (only) for searching, this is a sensible and correct solution.

**Comment by H-Shay** [comment]
> > This PR intercepts message bodies with "\u0000" in them and replaces them with "\u0020" before inserting into the database, which causes an error in Postgres databases. Fixes #9341
> 
> To be clear -- this _only_ does it for the search tables, correct? Not the original event content?

I think the answer to this is yes?

**Comment by clokep** [comment]
> Ah, yeah, it does. Sorry, i was mistaken and misled that this was about storing event JSON, if it's about NULL character bytes that have to be parsed (only) for searching, this is a sensible and correct solution.

That's fair, I made the same mistake on the issue when discussing this!

File: synapse/storage/databases/main/events.py [hunk 1]
```diff
@@ -1955,6 +1955,7 @@ def store_event_search_txn(self, txn, event, key, value):
             key (str):
             value (str):
         """
+
```
clokep, line 1958: Since we didn't actually change this file, let's backout this change (it makes it clearer where the real changes are)!

```suggestion
```

synapse/storage/databases/main/search.py
```diff
@@ -33,7 +33,7 @@
 
 
 class SearchWorkerStore(SQLBaseStore):
-    def store_search_entries_txn(self, txn, entries):
+    def store_search_entries_txn(self, txn, entries: Iterable[SearchEntry]):
```
clokep, line 36:
```suggestion
    def store_search_entries_txn(self, txn, entries: Iterable[SearchEntry]) -> None:
```

And please remove the type for `entries` in the docstring (and probably move up the description to a single line).

synapse/storage/databases/main/search.py
```diff
@@ -55,7 +55,7 @@ def store_search_entries_txn(self, txn, entries):
                     entry.event_id,
                     entry.room_id,
                     entry.key,
-                    entry.value,
+                    entry.value.replace("\u0000", "\u0020"),
```
clokep, line 58: I think this is just clearer as a space, what do you think?

```suggestion
                    entry.value.replace("\u0000", " "),
```

tests/test_null_byte_insertion.py
```diff
@@ -0,0 +1,43 @@
...
+        synapse.rest.admin.register_servlets_for_client_rest_resource,
+        login.register_servlets,
+        account.register_servlets,
+        register.register_servlets,
+        room.register_servlets,
+    ]
+
+    def setUp(self):
+        super().setUp()
+
```
clokep, line 33: A missing method automatically goes up to the super-class, so no reason to include it.

```suggestion
    def setUp(self):
        super().setUp()

```

tests/test_null_byte_insertion.py
```diff
@@ -0,0 +1,43 @@
...
+from tests.unittest import HomeserverTestCase
+
+
+class NullByteInsertionTest(HomeserverTestCase):
+    servlets = [
+        synapse.rest.admin.register_servlets_for_client_rest_resource,
+        login.register_servlets,
+        account.register_servlets,
+        register.register_servlets,
```
clokep, line 27: I don't think we need the `account` or `register` ones, but we might... it's hard to tell which are necessary.

tests/test_null_byte_insertion.py
```diff
@@ -0,0 +1,43 @@
...
+        room.register_servlets,
+    ]
+
+    def setUp(self):
+        super().setUp()
+
+    # Note that this test must be run with postgres or else is meaningless,
+    # as sqlite will accept insertion of null code points
+    def test_null_byte(self):
```
clokep, line 36: I'm not sure it is meaningless, it's still asserting that sqlite works!

Anyway, it would be good to move this into a docstring and flesh it out a bit.

```suggestion
    def test_null_byte(self):
        """
        Postgres doesn't like null bytes going into the search tables. Internally
        we replace those with a space.
        
        Ensure this doesn't break anything.
        """
```

tests/test_null_byte_insertion.py
```diff
@@ -0,0 +1,43 @@
...
+        self.register_user("alice", "password")
+        access_token = self.login("alice", "password")
+        room_id = self.helper.create_room_as("alice", True, "1", access_token)
+        body = '{"body":"\u0000", "msgtype":"m.text"}'
```
clokep, line 40: I think this is meant to be a dictionary, not a string?

tests/test_null_byte_insertion.py
```diff
@@ -0,0 +1,43 @@
...
+
+        resp = self.helper.send(room_id, body, "1", access_token)
+        self.assertTrue("event_id" in resp)
```
clokep, line 43: We should probably test searching for this content afterward to ensure that everything is working as we expect. (That probably means we'll need to expand the body to something that has words we can search for!)

clokep (commented):
I think this is the right direction!

tests/test_null_byte_insertion.py
ShadowJonathan: (uhh, the suggested change is one of adding a newline?)

synapse/storage/databases/main/search.py
H-Shay: Yes I think that makes sense!

tests/test_null_byte_insertion.py
H-Shay: The test runs fine without them so they probably are not needed!

tests/test_null_byte_insertion.py
H-Shay: Setting it as a dict results in the error `Expected: 200, got: 400, resp: b'{"errcode":"M_UNKNOWN","error":"\'body\' not a string type"}'` so I assumed it needed to be a string?

tests/test_null_byte_insertion.py
clokep: Yeah, then they're not needed and can be removed. I think the others are needed, but probably worth double checking.

tests/test_null_byte_insertion.py
H-Shay: So I tried to test search using the `search_msgs` function https://github.com/matrix-org/synapse/blob/3eba047d388fd0d798229a0779f343dbda8a2887/synapse/storage/databases/main/search.py#L346 to search for text that was inserted (I changed the body to 'hi\u0000bob' to test that) but ran into problems because `search_msg` calls `_find_highlights_in_postgres` https://github.com/matrix-org/synapse/blob/3eba047d388fd0d798229a0779f343dbda8a2887/synapse/storage/databases/main/search.py#L623 which afaict pulls message bodies from different tables than the one this PR alters (`event_search`), and as such those message bodies still contain the nefarious `\u0000`, which causes `_find_highlights_in_postgres` to fail at this line: https://github.com/matrix-org/synapse/blob/3eba047d388fd0d798229a0779f343dbda8a2887/synapse/storage/databases/main/search.py#L675 causing the whole search to fail. This, however, seems out of the scope of this PR, since the idea was to specifically only alter the `event_search` table.
I did write a test that manually queries the `event_search` table to be sure that the string was still being stored and could be retrieved, would this be sufficient or should I take another approach?

reivilibre:
I have changed the headline just to be a little bit less alarming :).

tests/test_null_byte_insertion.py
reivilibre: I think we ought to do something about `_find_highlights_in_postgres` — does this mean that if someone inserted a message containing a null byte, that they'd be able to prevent other users from performing any room searches that match the null-containing message? That sounds like a nasty bug, sort of a security issue.

I think replacing null bytes with spaces in `values` there also would make sense.

It might stop matching parts of that specific message (as in, it might not yield a result if the message says `"the\u0000mat"` and you search for `"the mat"`), but at least it won't break the search functionality entirely whenever a potential result contains a NULL and causes an exception that bubbles all the way back up to cause a 500 internal server error (if that makes sense?).

tests/test_null_byte_insertion.py
```diff
@@ -0,0 +1,43 @@
+# Copyright 2021 The Matrix.org Foundation C.I.C.
```
reivilibre, line 1: This file probably wants to be in a more organised directory (and named something a bit more specific to *where* the null byte is being inserted).

There's already `RoomSearchTestCase` in `tests/rest/client/test_rooms.py`, but 1) that file is already big and 2) I'm not sure you're so interested in testing the behaviour from a REST point of view.

My personal feeling would be perhaps `tests/storage/test_room_search.py` (especially if you're going to e.g. call `search_msgs` which is in the `storage` component of Synapse), but I would like to ask @clokep for his opinion.

*Maybe you would* want to test this from the client API point of view and perhaps `tests/rest/client/test_room_search.py` would make sense in that case.

tests/test_null_byte_insertion.py
reivilibre: I think `send` sends an `m.text` message, so you would just write `"\u0000"` rather than specify any JSON.

tests/test_null_byte_insertion.py
reivilibre: (or in other words,

> causing the whole search to fail

is exactly the kind of thing we're interested in testing *doesn't* happen! :))

tests/test_null_byte_insertion.py
clokep: > My personal feeling would be perhaps `tests/storage/test_room_search.py` (especially if you're going to e.g. call `search_msgs` which is in the `storage` component of Synapse), but I would like to ask @clokep for his opinion.

This seems reasonable to me. I did ask @H-Shay to look if there are other tests this could be grouped with, but if those are only rest tests then I'm not sure it makes sense to put it with those.

tests/test_null_byte_insertion.py
H-Shay: > — does this mean that if someone inserted a message containing a null byte, that they'd be able to prevent other users from performing any room searches that match the null-containing message?

It does exactly mean that. It could also have been happening this whole time: if someone sent a "hi\u0000bob" and someone else sent "hi bob" then theoretically the first message body could be pulled up and cause things to fall down even though technically it was never stored in the `event_search` table because now the same "hi" and "bob" keywords have been stored. So I have been trying to work out whether that behavior predates/will persist after we fix search. Afaict just because message bodies were getting bounced out of the `events_search` table doesn't necessarily mean they were getting bounced out of cache, etc (as the testing shows)-that really depends on the order in which things are stored as the message comes in, doesn't it?

synapse/storage/databases/main/search.py
clokep: Yeah I probably would have used `\x00` personally, but they're all equivalent.

tests/test_null_byte_insertion.py
reivilibre: > if someone sent a "hi\u0000bob" and someone else sent "hi bob" then theoretically the first message body could be pulled up

This doesn't match my understanding, because as you say, `"hi\u0000bob"` won't make it to the `event_search` table (due to the exact bug you're fixing in this PR).
If it doesn't make it into that table, then a search is not able to retrieve it.
If a search doesn't retrieve it, then that search won't ask Postgres to highlight it.

If you see some other way that this could happen, you may well be right — I'm not all that familiar! (so do let me know if this doesn't match up with what you expect)

(BTW, I don't think the 'highlighting' thing is obvious unless you have a clear picture of what the message search looks like — try searching a room if you haven't seen it yet. The only point of it seems to be to figure out *which* part of the message caused Postgres to match it.)

reivilibre:
Remark: There is also at least another good reason to make sure that you treat Postgres and SQLite the same: sometimes users will convert their SQLite databases to Postgres ones (see: the portdb script), and it would be good to avoid inserting null bytes into SQLite that would prevent someone from making that conversion.

tests/storage/test_room_search.py
```diff
@@ -0,0 +1,82 @@
...
+    ]
+
+    def test_null_byte(self):
+        """
+        Postgres/SQLite don't like null bytes going into the search tables. Internally
+        we replace those with a space.
+
+        Ensure this doesn't break anything.q
+
+        """
```
reivilibre, line 37:
```suggestion
        Ensure this doesn't break anything.
        """
```

tests/storage/test_room_search.py
```diff
@@ -0,0 +1,82 @@
...
+        # check that search still works when given a search term that overlaps
+        # with the message that we replaced the null byte in and an unrelated one
+        res3 = self.get_success(store.search_msgs([room_id], "hi", ["content.body"]))
+        self.assertEquals(res3.get("count"), 2)
+        res4 = self.get_success(
+            store.search_msgs([room_id], "hi alice", ["content.body"])
+        )
+
+        # check content of highlights if we are using postgres
+        if "SYNAPSE_POSTGRES" in os.environ and os.environ["SYNAPSE_POSTGRES"] == "1":
```
reivilibre, line 78: I would probably prefer to check that our store engine is Postgres directly, rather than rely on these env vars being present — *just in case* we ever change these around and don't remember to fix it here.

I think this will do it, but may be wrong:

```suggestion
        if isinstance(store.database_engine, PostgresEngine):
```

changelog.d/10820.misc
```diff
@@ -0,0 +1 @@
+Made a change so that posting an event with "\u0000" no longer causes an internal server error.
```
reivilibre, line 1: 'Made a change' sounds a bit soft (you fixed a bug — don't understate it :)) and I would probably say a 'null byte' rather than `"\u0000"`. I am probably not the best wordsmith around, however :innocent:.

```suggestion
Fix a bug where sending an `m.room.text` event containing a null byte would cause an internal server error.
```

synapse/storage/databases/main/search.py
reivilibre: I think you want to apply ^

reivilibre: also annotate `txn: LoggingTransaction` whilst we're here, if you don't mind :)

reivilibre: also remove the types from the docstring `txn (cursor):` → `txn:` etc — we tend to do that to make sure they don't fall out of sync with the signature that is checked by Mypy

reivilibre (changes requested):

H-Shay:
So I am noting that there were some Sytest failures on the last push, which had very few changes and none to any real logic-also the previous version of this commit passed these tests and nothing significant changed, and the tests seem to pass locally. Wondering if this is a flaky Sytest?

reivilibre:
> So I am noting that there were some Sytest failures on the last push, which had very few changes and none to any real logic-also the previous version of this commit passed these tests and nothing significant changed, and the tests seem to pass locally. Wondering if this is a flaky Sytest issue?

I would have thought so, but given there were about 5~10, I've given it a restart to see if that will shake them off. :(

reivilibre (commented):
I think this is good, let's wait and see if the SyTests get shaken out (and I would like to see if @clokep has anything to add).

synapse/storage/databases/main/search.py
```diff
@@ -55,7 +56,7 @@ def store_search_entries_txn(self, txn, entries):
                     entry.event_id,
                     entry.room_id,
                     entry.key,
-                    entry.value,
+                    entry.value.replace("\u0000", " "),
```
clokep, line 59: Probably best to do this above the if-statement with a comment about why we're doing it.

```
# ...
value = entry.value.replace("\u0000", " ")
```

tests/storage/test_room_search.py
```diff
@@ -0,0 +1,80 @@
...
+    servlets = [
+        synapse.rest.admin.register_servlets_for_client_rest_resource,
+        login.register_servlets,
+        room.register_servlets,
+    ]
+
+    def test_null_byte(self):
+        """
+        Postgres/SQLite don't like null bytes going into the search tables. Internally
+        we replace those with a space.
```
clokep, line 32: I thought this was only a Postgres problem?

tests/storage/test_room_search.py
```diff
@@ -0,0 +1,80 @@
...
+        body3 = "hi alice"
+
+        # send messages and ensure they don't cause an internal server
+        # error
+        resp1 = self.helper.send(room_id, body1, "1", access_token)
+        resp2 = self.helper.send(room_id, body2, "2", access_token)
+        resp3 = self.helper.send(room_id, body3, "3", access_token)
+        self.assertTrue("event_id" in resp1)
+        self.assertTrue("event_id" in resp2)
+        self.assertTrue("event_id" in resp3)
```
clokep, line 52: We can simplify this a bit:

1. Remove some duplicated code.
2. Use longer variable names.
3. User `assertIn`.

```suggestion
        # Send messages and ensure they don't cause an error.
        for body in ["hi\u0000bob", "another message", "hi alice"]:
            response = self.helper.send(room_id, body, tok=access_token)
            self.assertIn("event_id", response)
```

tests/storage/test_room_search.py
```diff
@@ -0,0 +1,80 @@
...
+
+        # check that search still works with the message where the null byte was replaced
+        store = self.hs.get_datastore()
+        res1 = self.get_success(
+            store.search_msgs([room_id], "hi bob", ["content.body"])
+        )
+        self.assertEquals(res1.get("count"), 1)
```
clokep, line 59: I think it would be clearer to put the highlights checks here after the request. We won't need to keep as much state in our head as we move down through the test then.

(And we can use a single `response` variable that gets re-used instead of storing them separately).

clokep (commented):
Thanks for expanding the tests! I think we're almost there!

synapse/storage/databases/main/search.py
H-Shay: So I actually tried to do this at first and then could not figure out how to do it, since `entries` is of type `SearchEntry` which is a named tuple, and then each `entry` in `entries` is a tuple itself-I was having a very hard time figuring out how to replace a value in those immutable structures. What would be the most graceful way of achieving this?

clokep: Right, I think I completely ignored that there can be more than one entry!

Maybe a better thing to do would be to make a function which does this cleaning so we can consolidate the logic into one place and put a nice comment on it? This would also let us use it in `_find_highlights_in_postgres` which is nice!

Something like:

```python
def _clean_value_for_search(value: str) -> str:
    """..."""
    return value.replace("\u0000", " ")
```

I'm not 100% sure what to put in the docstring, but something about why we want to remove nulls as postgres breaks on those.

tests/storage/test_room_search.py
H-Shay: see my comment above: 
> -testing revealed that while SQLite doesn't complain when you insert a text string with "\u0000", it will stop processing that string at null byte when queried later. This causes search to fail silently when searching for the text that was inserted into the `events_search` table containing the null byte. While minor, I figured this was not ideal so I modified the code so that we are also substituting space for "\u0000" when inserting into a SQLite version of `events_search`.

clokep: Sorry, I missed that!

changelog.d/10820.misc
```diff
@@ -0,0 +1 @@
+Fix a bug where sending an `m.room.text` event containing a null byte would cause an internal server error.
```
clokep, line 1:
```suggestion
Fix a long-standing bug where an `m.room.text` event containing a null byte would cause an internal server error.
```

synapse/storage/databases/main/search.py
```diff
@@ -33,13 +33,19 @@
 
 
 class SearchWorkerStore(SQLBaseStore):
-    def store_search_entries_txn(self, txn, entries):
+    def _clean_value_for_search(self, value: str) -> str:
+        """Replaces null code points in string as postgres/sql do not
+        like the insertion of strings with null code points"""
+        return value.replace("\u0000", " ")
```
clokep, line 39: I think we can move this outside the class since it isn't really related to the store (we don't need `self`).

tests/storage/test_room_search.py
```diff
@@ -0,0 +1,74 @@
...
+        Postgres/SQLite don't like null bytes going into the search tables. Internally
+        we replace those with a space.
+
+        Ensure this doesn't break anything.
+        """
+
+        # register a user and create a room, create some messages
+        self.register_user("alice", "password")
+        access_token = self.login("alice", "password")
+        room_id = self.helper.create_room_as("alice", True, "1", access_token)
```
clokep, line 40: I usually try to keep these calls as small as possible to make it clear what's important to the tests and what we're just using defaults from.

```suggestion
        room_id = self.helper.create_room_as("alice", tok=access_token)
```

tests/storage/test_room_search.py
```diff
@@ -0,0 +1,74 @@
...
+
+        # send messages and ensure they don't cause an internal server
+        # error
+        for body in ["hi\u0000bob", "another message", "hi alice"]:
+            response = self.helper.send(room_id, body, tok=access_token)
+            self.assertIn("event_id", response)
+
+        # check that search still works with the message where the null byte was replaced
```
clokep, line 48:
```suggestion
        # Check that search works for the message where the null byte was replaced.
```

tests/storage/test_room_search.py
```diff
@@ -0,0 +1,74 @@
...
+        store = self.hs.get_datastore()
+        result = self.get_success(
+            store.search_msgs([room_id], "hi bob", ["content.body"])
+        )
+        self.assertEquals(result.get("count"), 1)
+        if isinstance(store.database_engine, PostgresEngine):
+            self.assertIn("hi", result.get("highlights"))
+            self.assertIn("bob", result.get("highlights"))
+
+        # check that search still works with another unrelated message
```
clokep, line 58:
```suggestion
        # Check that search works for an unrelated message.
```

tests/storage/test_room_search.py
```diff
@@ -0,0 +1,74 @@
...
+        result = self.get_success(
+            store.search_msgs([room_id], "another", ["content.body"])
+        )
+        self.assertEquals(result.get("count"), 1)
+        if isinstance(store.database_engine, PostgresEngine):
+            self.assertIn("another", result.get("highlights"))
+
+        # check that search still works when given a search term that overlaps
+        # with the message that we replaced the null byte in and an unrelated one
```
clokep, line 67:
```suggestion
        # Check that search works for a search term that overlaps with the message
        # containing a null byte and an unrelated message.
```

clokep (commented):
Looks good! Just a few nits left.

synapse/storage/databases/main/search.py
```diff
@@ -32,14 +32,21 @@
 )
 
 
+def _clean_value_for_search(value: str) -> str:
+    """Replaces null code points in string as postgres/sql do not
```
clokep, line 36:
```suggestion
    """Replaces null code points in string as postgres/sqlite do not
```

synapse/storage/databases/main/search.py
```diff
@@ -32,6 +32,12 @@
 )
 
 
+def _clean_value_for_search(value: str) -> str:
+    """Replaces null code points in string as postgres/sqlite do not
+    like the insertion of strings with null code points"""
```
reivilibre, line 37:
```suggestion
    """
    Replaces any null code points in the string with spaces as
    Postgres and SQLite do not like the insertion of strings with
    null code points into the full-text search tables.
    """
```

reivilibre (approved):
Looks good to me, just one pedantic tweak :)

changelog.d/10820.misc
```diff
@@ -0,0 +1 @@
+Fix a long-standing bug where an `m.room.text` event containing a null byte would cause an internal server error.
```
aaronraimist, line 1:
```suggestion
Fix a long-standing bug where an `m.room.message` event containing a null byte would cause an internal server error.
```

I don't think m.room.text is a thing

reivilibre: Ah sugar, excellent spot! `m.text` is a thing so I got confused.

H-Shay:
Merged.

PR #10823, opened 2021-09-14
Add type hints to state database

clokep:
Adds type hints to `synapse.storage.databases.state`.

This is an old branch I had started at some point... (hence the `-2`...)

### DIFF
--- a/synapse/storage/databases/state/bg_updates.py
+++ b/synapse/storage/databases/state/bg_updates.py
@@ -13,12 +13,19 @@
 # limitations under the License.
 
 import logging
-from typing import Optional
+from typing import TYPE_CHECKING, Dict, List, Optional, Tuple, Union
 
 from synapse.storage._base import SQLBaseStore
-from synapse.storage.database import DatabasePool
+from synapse.storage.database import (
+    DatabasePool,
+    LoggingDatabaseConnection,
+    LoggingTransaction,
+)
 from synapse.storage.engines import PostgresEngine
 from synapse.storage.state import StateFilter
+
+if TYPE_CHECKING:
+    from synapse.server import HomeServer
 
 logger = logging.getLogger(__name__)
 
@@ -31,7 +38,9 @@ class StateGroupBackgroundUpdateStore(SQLBaseStore):
     updates.
     """
 
-    def _count_state_group_hops_txn(self, txn, state_group):
+    def _count_state_group_hops_txn(
+        self, txn: LoggingTransaction, state_group: int
+    ) -> int:
         """Given a state group, count how many hops there are in the tree.
 
         This is used to ensure the delta chains don't get too long.
@@ -56,7 +65,7 @@ class StateGroupBackgroundUpdateStore(SQLBaseStore):
         else:
             # We don't use WITH RECURSIVE on sqlite3 as there are distributions
             # that ship with an sqlite3 version that doesn't support it (e.g. wheezy)
-            next_group = state_group
+            next_group: Optional[int] = state_group
             count = 0
 
             while next_group:
@@ -73,11 +82,14 @@ class StateGroupBackgroundUpdateStore(SQLBaseStore):
             return count
 
     def _get_state_groups_from_groups_txn(
-        self, txn, groups, state_filter: Optional[StateFilter] = None
+        self,
+        txn: LoggingTransaction,
+        groups: List[int],
+        state_filter: Optional[StateFilter] = None,
     ):
         state_filter = state_filter or StateFilter.all()
 
-        results = {group: {} for group in groups}
+        results: Dict[int, Dict[Tuple[str, str], str]] = {group: {} for group in groups}
 
         where_clause, where_args = state_filter.make_sql_filter_clause()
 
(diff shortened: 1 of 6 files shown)


synapse/storage/databases/state/bg_updates.py
```diff
@@ -73,11 +82,14 @@ def _count_state_group_hops_txn(self, txn, state_group):
             return count
 
     def _get_state_groups_from_groups_txn(
-        self, txn, groups, state_filter: Optional[StateFilter] = None
+        self,
+        txn: LoggingTransaction,
+        groups: List[int],
+        state_filter: Optional[StateFilter] = None,
     ):
```
erikjohnston, line 89:
```suggestion
    ) -> Dict[int, Dict[Tuple[str, str], str]]:
```

?

synapse/storage/databases/state/bg_updates.py
```diff
@@ -340,14 +360,14 @@ def reindex_txn(txn):
...
-        def reindex_txn(conn):
+    async def _background_index_state(self, progress: dict, batch_size: int) -> int:
+        def reindex_txn(conn: LoggingDatabaseConnection) -> None:
             conn.rollback()
             if isinstance(self.database_engine, PostgresEngine):
                 # postgres insists on autocommit for the index
                 conn.set_session(autocommit=True)
                 try:
-                    txn = conn.cursor()
+                    txn = conn.LoggingTransaction()
```
erikjohnston, line 370: Huh? `LoggingDatabaseConnection` doesn't have such a method, and `cursor()` should be correct anyway?

clokep: I have no idea what happened here...

synapse/storage/databases/state/bg_updates.py
```diff
@@ -173,7 +186,8 @@ def _get_state_groups_from_groups_txn(
                         allow_none=True,
                     )
 
-        return results
+        # The results shouldn't be considered mutable.
+        return cast(Dict[int, StateMap[str]], results)
```
clokep, line 190: This is pretty dirty, but I couldn't figure out a way for it to be mutable in this method, but not mutable on return...

synapse/storage/databases/state/bg_updates.py
```diff
@@ -198,7 +217,9 @@ def __init__(self, database: DatabasePool, db_conn, hs):
             columns=["room_id"],
         )
 
-    async def _background_deduplicate_state(self, progress, batch_size):
+    async def _background_deduplicate_state(
+        self, progress: dict, batch_size: int
```
clokep, line 221: I'm not sure if `progress` can be more specific? Maybe `JsonDict`?

clokep:
Thanks @erikjohnston! I double checked and the rest of the PR didn't seem to have anything out of place in it.

synapse/storage/databases/state/bg_updates.py
reivilibre: Not easily. Me and David looked at a similar case. (Though actually it seems reasonable to believe it might be doable as long as you update all background processes at the same time.)

synapse/storage/databases/state/bg_updates.py
reivilibre: I typed this function in one of my abandoned PRs and don't recall needing any casts.

`MutableStateMap[X]` is a subtype of `StateMap[X]`, so this will work as long as you say this is the return type: `Mapping[int, StateMap[str]]` (iirc).

clokep: That's what i had originally and it didn't like it.

reivilibre: > That's what i had originally and it didn't like it.

Why not?

reivilibre: Oh I thought you said **you** didn't like it. I tried this just now on a copy of your branch and it worked fine

clokep: Ah, the trick seems to be to use `Mapping`, not `Dict` as the outer container. (o_O)

reivilibre: yes, because `Dict` can't be covariant in its value type (due to its mutability)

erikjohnston (approved):

synapse/storage/databases/state/bg_updates.py
DMRobertson: I hoped I could define my own `TypedDict`, but this ended up clashing with the background process stuff. Specifically:

- `register_background_update_handler`'s `update_handler` has type `Callable[[JsonDict, int], Awaitable[int]]`
- this means that the `progress` argument type ends up being compared against `JsonDict = Dict[str, Any]`
- `TypedDict` is not a subtype of `JsonDict` because you can add new keys or remove existing keys from a `Dict`
   - so passing `t: TypedDict` to a function `process_dict(d: Dict)` might alter `t`'s keys. This would be bad.
- `TypedDict` is a subtype of `Mapping[str, Any]` because the latter is immutable
- More details here: https://www.python.org/dev/peps/pep-0589/#id20

An `update_handler` itself is called here:

```python
        progress = db_to_json(progress_json)

        time_start = self._clock.time_msec()
        items_updated = await update_handler(progress, batch_size)
```
and `db_to_json` returns a `JsonDict`.

We didn't find a nice way of joining this all up that didn't involve a sea of `cast`s.

PR #10825, opened 2021-09-15
Add an approximate difference method to StateFilters

reivilibre:


reivilibre:
The SyTest failure is known to be very flakey.

synapse/storage/state.py
```diff
@@ -356,6 +356,60 @@ def get_member_split(self) -> Tuple["StateFilter", "StateFilter"]:
...
+        """
+        Returns a state filter which represents self - subtrahend;
+        if the set of state events given by a state filter F are represented as
+        E(F), then the resultant state filter bears this property:
+
+            E(difference(self, subtrahend)) ⊇ E(self) ∖ E(subtrahend)
+
+        Ideally, this should be the narrowest such state filter, but this
+        function returns an approximation (since, for example, the set of
+        possible state keys is infinite).
```
DMRobertson, line 369: Much as I'm keen for any excuse to use `$\setminus$`... want some bikeshedding?

```suggestion
        Creates an approximate copy of `self` which additionally filters out `subtrahend`.

        This is "approximate" because the resulting filter might accept ..... / reject ...
                
        In symbols, if E(F) denotes the set of state events accepted by a state filter F,
        the new filter we create satisfies
        
            E(approx_difference(self, subtrahend)) ⊇ E(self) ∖ E(subtrahend) .
```

reivilibre: what do you think of the new description?

synapse/storage/state.py
```diff
@@ -356,6 +356,85 @@ def get_member_split(self) -> Tuple["StateFilter", "StateFilter"]:
...
+        The resultant state filter MUST admit all state events that are admitted
+        by only this filter (`self`) and not `subtrahend`. (1)
+        The resultant filter MAY be an over-approximation: the resultant state
+        filter MAY additionally admit other state events.
+
+
+        Formally, if the set of state events admitted by a state filter F are
+        written as E(F), then the resultant state filter bears this property:
+
+            E(difference(self, subtrahend)) ⊇ E(self) ∖ E(subtrahend)
```
DMRobertson, line 372: I can make the left hand side bigger while still preserving this property in two ways:

1. Add elements in neither E(self) nor E(subtrahend)
2. Add elements in E(subtrahend)

I think the prose below suggests that the approximation might be inexact due to 2 but not due to 1. Is that correct?

If so, I suggest replacing this with 

```suggestion
            E(self) \ E(subtrahend) ⊆ E(approx_difference) ⊆ E(self)
```

synapse/storage/state.py
```diff
@@ -356,6 +356,85 @@ def get_member_split(self) -> Tuple["StateFilter", "StateFilter"]:
 
         return member_filter, non_member_filter
 
+    def approx_difference(self, subtrahend: "StateFilter") -> "StateFilter":
+        """
+        Returns a state filter which represents `self - subtrahend`.
+
+        The resultant state filter MUST admit all state events that are admitted
+        by only this filter (`self`) and not `subtrahend`. (1)
```
DMRobertson, line 364: Not sure what the `(1)` refers to here.

synapse/storage/state.py
DMRobertson: And again, if so, suggest "MAY additionally admit other state events" be altered, e.g. to "MAY admit some events from `subtrahend`"

reivilibre: You're right.

I'm not actually 'interested' in the stronger property that you write — the code using this function does not depend on it and is happy with `StateFilter.all()` as a return value (for example) — though equally just returning `self` is also a satisfactory 'lazy, technically correct' answer.

For the sake of potential re-use in the future, I think what you are writing is sensible, so I'll try and incorporate it.

synapse/storage/state.py
reivilibre: thanks, it was left over from a prior draft where I referred to that property as (1)

synapse/storage/state.py
```diff
@@ -361,18 +361,20 @@ def approx_difference(self, subtrahend: "StateFilter") -> "StateFilter":
         Returns a state filter which represents `self - subtrahend`.
 
         The resultant state filter MUST admit all state events that are admitted
```
DMRobertson, line 363: I think we probably want something in prose that explicitly describes the `⊆ E(self)` bit. How about this?

```suggestion
The resultant filter

- MUST admit all state events that are admitted by only this filter (`self`) and not `subtrahend`;
- MUST NOT admit state events rejected by this filter (`self`); and
- MAY be an over-approximation: it may additionally admit some state events from `subtrahend`.
```

tests/storage/test_state.py
```diff
@@ -483,3 +482,119 @@ def test_get_state_for_event(self):
...
+            StateFilter.all().approx_difference(
+                StateFilter.from_types(
+                    (
+                        (EventTypes.Member, "@wombat:hs1"),
+                        (EventTypes.Member, "@spqr:hs1"),
+                    )
+                )
+            ),
+            StateFilter.all(),
+        )
```
DMRobertson, line 515: I think this is repeating the same assertion twice in different ways?

tests/storage/test_state.py
```diff
@@ -483,3 +482,119 @@ def test_get_state_for_event(self):
...
+
+        # we can subtract wildcards from wildcards
+        assert_difference(StateFilter.all(), StateFilter.all(), StateFilter.none())
+        assert_difference(
+            StateFilter(
+                types=frozendict(
+                    {
+                        EventTypes.Member: frozenset(),
+                        EventTypes.CanonicalAlias: None,
```
DMRobertson, line 524: Not really relevant to this change, but: it wasn't obvious to me what an empty set and None mean here (and if they ought to have different meanings).

I don't know if there's a nice way. If we had sum types I'd suggest
```rust
enum Fetch {
    All,
    None,
    WithStateKey(set_of_strings),
}
```
But that's slightly awkward because `None` and `WithStateKey(empty_set)` represent the same thing.  /shrug

tests/storage/test_state.py
```diff
@@ -483,3 +482,119 @@ def test_get_state_for_event(self):
...
+            StateFilter(
+                types=frozendict({EventTypes.Member: None}),
+                include_others=True,
+            ),
+            StateFilter(
+                types=frozendict({EventTypes.CanonicalAlias: frozenset({""})}),
+                include_others=True,
+            ),
+            StateFilter(
+                types=frozendict({EventTypes.CanonicalAlias: None}),
```
DMRobertson, line 551: Because as you say elsewhere, "all canonical aliases except for `""`" isn't currently expressible as a state filter.

DMRobertson (commented):
The tests and comment lgtm!

I'm a little scared about the implementation itself. How well do the tests exercise the different code paths and possibilities?

tests/storage/test_state.py
reivilibre: huh, you're right. I wonder why that got left in.

tests/storage/test_state.py
reivilibre: The docstring for StateFilter specifies what they mean, though that's not to say I don't agree with you — and yeah, I think empty set makes sense, but 'None' very much reads as, well, none. It's half tempting to have a constant `WILDCARD` for `None` ;p, but I feel like that might just hide things more.

reivilibre:
You're right to think about whether the tests exercise all the corners; I do have it in mind to more carefully comb through and write more systematic tests.

I was more hoping for a comment about whether the whole approach seemed sensible, though, to try and avoid writing loads of stuff for it to turn out to be a bit lacking.

erikjohnston:
FTR I think the idea is sane. We just had a call about a potential alternative implementation that will hopefully be easier to read about, @reivilibre will give it a (time boxed) go and see if it is in fact easier or not.

(The idea is broadly: you can split up a `StateFilter` into combinations of the all set, minus wildcards, plus wildcards and specific types, doing a difference based on the split out `StateFilter`s will hopefully be easier to reason about as its now just a difference of each set, and then at the end we can turn the combinations of sets back into a `StateFilter`

synapse/storage/state.py
```diff
@@ -356,6 +375,166 @@ def get_member_split(self) -> Tuple["StateFilter", "StateFilter"]:
 
         return member_filter, non_member_filter
 
+    def decompose_into_four_parts(
```
erikjohnston, line 378: I think these should be internal functions

synapse/storage/state.py
```diff
@@ -356,6 +375,166 @@ def get_member_split(self) -> Tuple["StateFilter", "StateFilter"]:
...
+        if self_all:
+            self_concretes_or_wild_types = self_wildcards.union(
+                typ for typ, _key in self_concrete_keys
+            )
+            # If self starts with all, then we add as wildcards any
+            # types which appear in the subtrahend's exclusion filter but do
+            # not already appear in self's concrete or wild types.
+            # This is needed to handle the fact that `include_others` includes
+            # all unspecified state types.
+            new_wildcards |= sub_excludes.difference(self_concretes_or_wild_types)
```
erikjohnston, line 514:
```suggestion
            # If self starts with all, then we add as wildcards any
            # types which appear in the subtrahend's exclusion filter (but
            # aren't in the self exclusion filter). This is as the subtrahend
            # filter will return everything BUT the types in its exclusion, so
            # we need to add those excluded types that also match the self
            # filter as wildcard types in the new filter.
            new_wildcards |= sub_excludes.difference(self_exclusion)
```

(Note that `self_exclusion` is the same as the union of wildcard types and concrete types in our case, due to the way `include_others` works)

I'd be also tempted to move this into the `if sub_all` section, as it really only makes sense if both sides are `ALL`, though it doesn't functionally matter as if `sub_all` is false then `sub_excludes` is empty.

erikjohnston (commented):
I think this is easier to follow! Thanks! Having come up with the idea of this approach I'd be keen to hear whether other people on the team follow it

synapse/storage/state.py
```diff
@@ -356,6 +356,85 @@ def get_member_split(self) -> Tuple["StateFilter", "StateFilter"]:
...
+            None if subtrahend.include_others else frozenset()
+        )
+        current_default_for_unspecified: Optional[FrozenSet[str]] = (
+            None if self.include_others else frozenset()
+        )
+        new_default_for_unspecified: Optional[FrozenSet[str]] = (
+            None if new_include_others else frozenset()
+        )
+        state_types = set(subtrahend.types.keys())
+        state_types.update(set(self.types.keys()))
```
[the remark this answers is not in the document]
reivilibre: urrr. This comment is made on an outdated diff even though I haven't made it outdated since your review :s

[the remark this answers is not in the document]
erikjohnston: Sooooooorry, this was a left over comment from when I looked at it previously :facepalm:

reivilibre:
Opening up for further feedback.

clokep:
> Opening up for further feedback.

Can you add a description to the PR (or link to a bug). It would help before diving into the code to know *why* we need this. :)

reivilibre:
> Can you add a description to the PR (or link to a bug). It would help before diving into the code to know _why_ we need this. :)

I've added a brief description, but wondering if you have thoughts on anything else I add. I have a tendency to leave the description blank because it seems silly to put text into the (ephemeral) PR when it could go into the docstring description itself.

But I guess this is breaking things up; now you can't see the entirety of the work so this comes a bit out of a blue and you can't look at the entirety of the changeset as a whole. Struggling a bit with interdependent PRs :/.

clokep:
> it seems silly to put text into the (ephemeral) PR when it could go into the docstring description itself.

I don't think of PRs as ephemeral. I quite often reference old ones when trying to figure out why something is the way it is.

reivilibre:
I should add that it's used in #10870 (currently there's a fake implementation which just returns the empty StateFilter if you do `self - self`, or `self` if you do `self - other` — it's a legal but not very useful over-approximation.

synapse/storage/state.py
```diff
@@ -356,6 +375,164 @@ def get_member_split(self) -> Tuple["StateFilter", "StateFilter"]:
...
+
+        return StateFilter.freeze(new_types, include_others=all_part)
+
+    def approx_difference(self, subtrahend: "StateFilter") -> "StateFilter":
+        """
+        Returns a state filter which represents `self - subtrahend`.
+
+        The resultant state filter
+        - MUST admit all state events that are admitted
+          by only this filter (`self`) and not `subtrahend`;
```
clokep, line 448: I'm having trouble understanding this, if it admits all state events from `self`, then isn't that not removing anything from `subtrahend`?

reivilibre: p'**only** from self and not subtrahend' — if you can think of a clearer way to say this, would appreciate (did think this was a rough part myself)

DMRobertson: perhaps "MUST admin all state events which are both admitted by self and rejected by subtrahend?"

clokep: I'm going to resolve this since we had a long conversation about this.

synapse/storage/state.py
```diff
@@ -134,6 +136,23 @@ def from_lazy_load_member_list(members: Iterable[str]) -> "StateFilter":
             include_others=True,
         )
 
+    @staticmethod
+    def freeze(types: Mapping[str, Optional[Collection[str]]], include_others: bool):
```
clokep, line 140: Somewhat feels like this could be replaced with `return StateFilter(synapse.util.frozenutil.freeze(types), include_others)`? I'm not sure if that's clearer or not though.

clokep (commented):
I didn't check over the implementation, I think you mostly wanted my thoughts on the docstrings? I find them much clearer now. Thanks for updating them!

erikjohnston (approved):

Issue #10786, opened 2021-09-08
anecdata says gappy syncs have slowed down by 3-4x at some point in the last N months.


ara4n:
empirically my ios app takes aaaages to incremental sync in the morning or after any significant time spent offline. manu and others have spotted it too. need to check average incr sync times to see when the step change was?

anoadragon453:
I don't really see an obvious step change across all synchrotrons in the last 50 days on matrix.org. I do however see one synchrotron hitting 100% CPU, before it relaxes and another one jumps to 100% CPU.

![image](https://user-images.githubusercontent.com/1342360/132664402-80a8efe8-1246-46ab-9bea-8a6c95d4b466.png)

It looks as if one account is causing one synchrotron to become slow. However, people have mentioned incremental syncs becoming slow across more homeservers than just matrix.org. So I think this is just a red herring (although one still worth investigating separately).

ara4n:
the timeframe of change could easily be >50 days here.

anoadragon453:
Result of the discussion is that it would be helpful to have detailed client logs of `/sync` timings before digging too far into this on the server side. Is the problem purely waiting for the server to return `/sync` data, or is there an increase in local processing requirements? Has the size of `/sync` data increased or become more complex? Are `/sync` requests timing out due to long server-side processing times or simply because there are no messages?

I see a few mentions in features/bugfixes over the past few Synapse releases regarding `/sync`, though it's not entirely clear where to start looking.

ara4n:
It might be worth profiling some of my catchup syncs via jaeger (which I think is enabled for my account). 

For instance, I just got a great instance of this on iOS after an hour of being offline - see the linked rageshake.

I closed the app at `2021-09-09 20:54:16.755`, and launched it again at `2021-09-09 21:41:07.810` (console.log.gz). Then, it starts an incremental sync at:

```
2021-09-09 21:41:12.418 Riot[19788:9287787]  [MXSession] Do a server sync (catching up) from token: s2297526719_757284961_9800653_1015632397_934029780_2607818_323407029_1852203500_207666
2021-09-09 21:41:12.418 Riot[19788:9287787]  [MXHTTPClient] #6 - GET _matrix/client/r0/sync
```

...which completes after 36 seconds?!

```
2021-09-09 21:41:48.399 Riot[19788:9287787]  [MXHTTPClient] #6 - GET _matrix/client/r0/sync completed in 35980ms
[...]

PR #10826, opened 2021-09-15
Opt out of cache expiry for `get_users_who_share_room_with_user`

DMRobertson:
We'll still invalidate the cache as usual, and we'll still throw away entries if the cache gets too big. But we won't prematurely throw away entries just because they haven't been used in a while.

Might help with #10786.

### DIFF
--- a/changelog.d/10826.misc
+++ b/changelog.d/10826.misc
@@ -1,0 +1,2 @@
+Opt out of cache expiry for `get_users_who_share_room_with_user`, to hopefully improve `/sync` performance when you
+haven't synced recently.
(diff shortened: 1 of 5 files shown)


erikjohnston (approved):
I'm going to manually merge this onto the hotfixes branch while we decide if we want to make it part of the RC or not

DMRobertson:
> I'm going to manually merge this onto the hotfixes branch while we decide if we want to make it part of the RC or not

We didn't end up adding it to the RC. I guess this is best pointed at develop now?

erikjohnston (approved):
I think people made noises that this helped?

DMRobertson:
Sytest failures are in flakes that should be dealt with by matrix-org/sytest#1142.

PR #10827, opened 2021-09-15
Fix /initialSync error due to unhashable `RoomStreamToken`

squahtx:
Thanks to @DMRobertson for spotting the error.

It's triggered by Sytest's tests/90jira/SYN-115.pl, though the test does not fail.

The full stack trace looks like:
```
   File "/home/squah/repos/synapse/synapse/http/server.py", line 258, in _async_render_wrapper
  File "/home/squah/repos/synapse/synapse/http/server.py", line 446, in _async_render
  File "/home/squah/repos/synapse/synapse/rest/client/initial_sync.py", line 44, in on_GET
    content = await self.initial_sync_handler.snapshot_all_rooms(
  File "/home/squah/repos/synapse/synapse/handlers/initial_sync.py", line 97, in snapshot_all_rooms
    return await self.snapshot_cache.wrap(
  File "/home/squah/repos/synapse/synapse/util/caches/response_cache.py", line 192, in wrap
    result = self.get(key)
  File "/home/squah/repos/synapse/synapse/util/caches/response_cache.py", line 98, in get
    result = self.pending_result_cache.get(key)
  File "<attrs generated hash synapse.types.StreamToken>", line 2, in __hash__
    return hash((
  File "<attrs generated hash synapse.types.RoomStreamToken>", line 2, in __hash__
    return hash((
TypeError: unhashable type: 'dict'
```


clokep:
Should we update the documentation of `RoomStreamToken` to mention it must be hashable? I guess that would also be true of `PaginationConfig`.

synapse/types.py
```diff
@@ -466,12 +467,12 @@ class RoomStreamToken:
     stream = attr.ib(type=int, validator=attr.validators.instance_of(int))
 
     instance_map = attr.ib(
-        type=Dict[str, int],
-        factory=dict,
+        type=Mapping[str, int],
```
reivilibre, line 470: I was about to say 'huh, I never thought of just calling frozendict a Mapping!', but I'm not sure this type annotation tells the whole story.

`dict` is a subclass of `Mapping`, yet this field very clearly needs the `Mapping` to be hashable, so the type annotation is too loose here.

I think this really does need `frozendict[str, int]` — which I think you have to put in double quotes (:sob: — there is a fork of frozendict which came in after generics emerged and fixes that, but we don't use it in Synapse).

synapse/types.py
```diff
@@ -466,12 +467,12 @@ class RoomStreamToken:
...
+        factory=frozendict,
         validator=attr.validators.deep_mapping(
             key_validator=attr.validators.instance_of(str),
             value_validator=attr.validators.instance_of(int),
-            mapping_validator=attr.validators.instance_of(dict),
+            mapping_validator=attr.validators.instance_of(frozendict),
```
reivilibre, line 475: oh but you have this validator here ... well. Now I am wondering if this is better or worse. I'm interested to see what others of the team think (or is this something we already do elsewhere?)

reivilibre: I guess this is a runtime check rather than something Mypy can check, though

synapse/types.py
clokep: > I think this really does need `frozendict[str, int]` — which I think you have to put in double quotes (😭 — there is a fork of frozendict which came in after generics emerged and fixes that, but we don't use it in Synapse).

We have a stub for this in our repo.

squahtx: I've updated the type to be `"frozendict[str, int]"` with the quotes. It does feel slightly off since the `frozendict`-ness is really an implementation detail. The type required is `Mapping[str, int] & Hashable` except there's no way to express that in Python that I know of.

squahtx:
> Should we update the documentation of `RoomStreamToken` to mention it must be hashable? I guess that would also be true of `PaginationConfig`.

Good point, I've updated the docstrings of `RoomStreamToken` and `StreamToken`. `PaginationConfig` technically doesn't need updating since its attributes gets unpacked somewhere before the hashing.

synapse/types.py
reivilibre: > > I think this really does need `frozendict[str, int]` — which I think you have to put in double quotes (sob — there is a fork of frozendict which came in after generics emerged and fixes that, but we don't use it in Synapse).
> 
> We have a stub for this in our repo.

I don't think stubs are used at runtime, so you still have to put in the quotes otherwise you get the type subscripting error. (am I missing something? I hope so :))

squahtx: > I don't think stubs are used at runtime, so you still have to put in the quotes otherwise you get the type subscripting error. (am I missing something? I hope so :))

Yep, without the quotes I get a type subscripting error.

PR #10828, opened 2021-09-15
Cleanup opentracing logging for syncs

erikjohnston:
We added a bunch of spans in #10704, but this ended up adding a *lot* of
redundant spans for rooms where nothing changed, so instead we only
start the span if there might be something interesting going on.

Pro tip: Probably easiest to review by disabling showing of whitespace changes :)

### DIFF
--- a/synapse/handlers/sync.py
+++ b/synapse/handlers/sync.py
@@ -1960,7 +1960,7 @@ class SyncHandler:
 
         with start_active_span("generate_room_entry"):
             set_tag("room_id", room_id)
-            log_kv({"events": len(events or [])})
+            log_kv({"events": len(events or ())})
 
             log_kv(
                 {
--- a/synapse/handlers/sync.py
+++ b/synapse/handlers/sync.py
@@ -1533,21 +1533,18 @@ class SyncHandler:
         newly_left_rooms = room_changes.newly_left_rooms
 
         async def handle_room_entries(room_entry: "RoomSyncResultBuilder"):
-            with start_active_span("generate_room_entry"):
-                set_tag("room_id", room_entry.room_id)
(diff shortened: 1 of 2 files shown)


synapse/handlers/sync.py
```diff
@@ -1960,139 +1957,157 @@ async def _generate_room_entry(
...
-        log_kv(
-            {
-                "batch_events": len(batch.events),
-                "prev_batch": batch.prev_batch,
-                "batch_limited": batch.limited,
-            }
-        )
+        with start_active_span("generate_room_entry"):
+            set_tag("room_id", room_id)
+            log_kv({"events": len(events or [])})
```
erikjohnston, line 1963: Everything after this is just a change in whitespace

synapse/handlers/sync.py
```diff
@@ -1960,139 +1957,157 @@ async def _generate_room_entry(
...
-            newly_joined_room=newly_joined,
-        )
-        log_kv(
-            {
-                "batch_events": len(batch.events),
-                "prev_batch": batch.prev_batch,
-                "batch_limited": batch.limited,
-            }
-        )
+        with start_active_span("generate_room_entry"):
```
clokep, line 1961: Do we want to change the span name?

synapse/handlers/sync.py
clokep: A micro-optimization would be to use a tuple instead of a list, I think. (since an empty tuple is only created once).

clokep (approved):

synapse/handlers/sync.py
erikjohnston: I'm OK with it, but open to other suggestions :shrug:

PR #10829, opened 2021-09-15
Track why we're evicting from caches

DMRobertson:
So we can see how often we're evicting because the cache is too big or because
the cache entries haven't been recently used. Previously these weren't
distinguishable on Grafana.

### DIFF
--- a/changelog.d/10829.misc
+++ b/changelog.d/10829.misc
@@ -1,0 +1,1 @@
+Track cache eviction rates more finely in Prometheus' monitoring.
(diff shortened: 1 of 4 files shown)


erikjohnston (approved):
We'll need to remember to update the synapsh grafana dashboard (and export it into `contrib/`)

DMRobertson:
> We'll need to remember to update the synapsh grafana dashboard (and export it into `contrib/`)

Thanks. I guess it's best to wait for a deployment on matrix.org for this?

erikjohnston:
Yup, though feel free to redeploy matrix.org whenever :)

DMRobertson:
> Yup, though feel free to redeploy matrix.org whenever :)

For completeness: this is done now.

PR #10831, opened 2021-09-15
Require type hints in the handlers module

clokep:
Similar to what I did in #10736 and then #10817, this requires all methods to be typed in the `synapse.handlers` module.

In order to do that I had to add a bunch of type hints.

This is reviewable commit-by-commit.

### DIFF
--- a/synapse/handlers/account_data.py
+++ b/synapse/handlers/account_data.py
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import random
-from typing import TYPE_CHECKING, List, Tuple
+from typing import TYPE_CHECKING, Any, List, Tuple
 
 from synapse.replication.http.account_data import (
     ReplicationAddTagRestServlet,
@@ -171,7 +171,7 @@ class AccountDataEventSource:
         return self.store.get_max_account_data_stream_id()
 
     async def get_new_events(
-        self, user: UserID, from_key: int, **kwargs
+        self, user: UserID, from_key: int, **kwargs: Any
     ) -> Tuple[List[JsonDict], int]:
         user_id = user.to_string()
         last_stream_id = from_key
--- a/synapse/handlers/appservice.py
+++ b/synapse/handlers/appservice.py
@@ -116,9 +116,9 @@ class ApplicationServicesHandler:
 
                         if not self.started_scheduler:
 
-                            async def start_scheduler():
+                            async def start_scheduler() -> None:
                                 try:
-                                    return await self.scheduler.start()
+                                    await self.scheduler.start()
                                 except Exception:
                                     logger.error("Application Services Failure")
 
--- a/synapse/handlers/appservice.py
+++ b/synapse/handlers/appservice.py
@@ -58,7 +58,7 @@ class ApplicationServicesHandler:
         self.current_max = 0
         self.is_processing = False
 
-    def notify_interested_services(self, max_token: RoomStreamToken):
+    def notify_interested_services(self, max_token: RoomStreamToken) -> None:
         """Notifies (pushes) all application services interested in this event.
 
         Pushing is done asynchronously, so this method won't block for any
@@ -82,7 +82,7 @@ class ApplicationServicesHandler:
         self._notify_interested_services(max_token)
 
     @wrap_as_background_process("notify_interested_services")
-    async def _notify_interested_services(self, max_token: RoomStreamToken):
+    async def _notify_interested_services(self, max_token: RoomStreamToken) -> None:
         with Measure(self.clock, "notify_interested_services"):
             self.is_processing = True
             try:
@@ -184,7 +184,7 @@ class ApplicationServicesHandler:
         stream_key: str,
         new_token: Optional[int],
         users: Optional[Collection[Union[str, UserID]]] = None,
-    ):
+    ) -> None:
         """This is called by the notifier in the background
         when a ephemeral event handled by the homeserver.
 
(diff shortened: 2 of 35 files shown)


synapse/handlers/cas.py
```diff
@@ -133,11 +133,9 @@ async def _validate_ticket(
...
         except HttpResponseException as e:
             description = (
-                (
-                    'Authorization server responded with a "{status}" error '
-                    "while exchanging the authorization code."
-                ).format(status=e.code),
-            )
+                'Authorization server responded with a "{status}" error '
+                "while exchanging the authorization code."
+            ).format(status=e.code)
```
clokep, line 138: This was a real bug, it was passing a tuple instead of a string.

synapse/handlers/appservice.py
```diff
@@ -116,9 +116,9 @@ async def handle_event(event: EventBase) -> None:
 
                         if not self.started_scheduler:
 
-                            async def start_scheduler():
+                            async def start_scheduler() -> None:
                                 try:
-                                    return await self.scheduler.start()
+                                    await self.scheduler.start()
```
clokep, line 121: `Scheduler.start` doesn't return anything so no need to pass that back.

synapse/handlers/sync.py
```diff
@@ -1534,12 +1534,12 @@ async def _generate_sync_entry_for_rooms(
...
         newly_left_rooms = room_changes.newly_left_rooms
 
-        async def handle_room_entries(room_entry: "RoomSyncResultBuilder"):
+        async def handle_room_entries(room_entry: "RoomSyncResultBuilder") -> None:
             with start_active_span("generate_room_entry"):
                 set_tag("room_id", room_entry.room_id)
                 log_kv({"events": len(room_entry.events or [])})
                 logger.debug("Generating room entry for %s", room_entry.room_id)
-                res = await self._generate_room_entry(
+                await self._generate_room_entry(
```
clokep, line 1542: _generate_room_entry modifies the RoomSyncResultBuilder and doesn't return anything.

synapse/handlers/federation.py
```diff
@@ -1262,11 +1262,11 @@ def sort_fun(ev: EventBase) -> Tuple[int, str]:
 
         logger.debug("construct_auth_difference before get_next!")
 
-        def get_next(it, opt=None):
```
clokep, line 1265: `opt` was never passed, it seemed clearer to just inline it below.

synapse/handlers/federation.py
```diff
@@ -1305,6 +1308,7 @@ def get_next(it, opt=None):
             if current_local.event_id < current_remote.event_id:
                 missing_locals.append(current_local)
                 current_local = get_next(local_iter)
+                continue
```
clokep, line 1311: I believe this is a real bug, without this the below code can run, but with a new `current_local` (which could be `None`).

synapse/handlers/account_data.py
```diff
@@ -171,7 +171,7 @@ def get_current_key(self, direction: str = "f") -> int:
         return self.store.get_max_account_data_stream_id()
 
     async def get_new_events(
-        self, user: UserID, from_key: int, **kwargs
+        self, user: UserID, from_key: int, **kwargs: Any
```
reivilibre, line 174: what's this `kwargs` for?

synapse/handlers/federation.py
reivilibre: unchecked since I don't know the code well

synapse/handlers/receipts.py
```diff
@@ -216,7 +216,7 @@ def filter_out_hidden(events: List[JsonDict], user_id: str) -> List[JsonDict]:
         return visible_events
 
     async def get_new_events(
-        self, from_key: int, room_ids: List[str], user: UserID, **kwargs
+        self, from_key: int, room_ids: List[str], user: UserID, **kwargs: Any
```
reivilibre, line 219: also the `kwargs` here

synapse/handlers/typing.py
```diff
@@ -485,7 +485,7 @@ async def get_new_events_as(
             return (events, handler._latest_room_serial)
 
     async def get_new_events(
-        self, from_key: int, room_ids: Iterable[str], **kwargs
+        self, from_key: int, room_ids: Iterable[str], **kwargs: Any
```
reivilibre, line 488: also the kwargs here

reivilibre (commented):
Everything looks good except:

- Can't verify the bug — would want someone who knew what they were looking at to confirm (though should we consider a test case to confirm?)
- these random `kwargs` — what are they doing there? Can we get rid of them? It seems like they might hide passed-but-unused arguments in the rest of the code, and since they're not being used anyway (as far as I can tell)…

synapse/handlers/account_data.py
clokep: I think that `get_new_events` generally takes some random arguments which most things don't care about. I did verify that there *are* other arguments going into this call, I'll double check if we can tighten this up.

clokep: I think this can be cleaned up, but I'd prefer to do it in a separate PR if that's OK!

clokep: I resolved the other threads about this to keep the conversation in one spot.

reivilibre: Sure, take this to another PR if you prefer.

erikjohnston:
FYI this now has conflicts

synapse/handlers/federation.py
erikjohnston: `construct_auth_difference` doesn't seem to be used by anything anymore, so let's just delete this entire function?

erikjohnston (approved):
Otherwise lgtm

synapse/handlers/federation.py
clokep: Good catch!

PR #10833, opened 2021-09-16
Extend ModuleApi with the methods we'll need to reject spam based on …IP - resolves #10832

### LINKED ISSUE #10832: An antispam module should be able to reject messages, invites, etc. by IP
**Description:**

An antispam module can block registration from a known-to-be-bad IP.

However, in most cases, we can only identify the IPs used by spambots after the spam has started, by which time it is often too late to block registration.

In such cases, when the sender is local, it would be useful to be able to block a message, etc. based on the current IPs of the sender.

Yoric:
### Pull Request Checklist

### DIFF
--- a/synapse/module_api/__init__.py
+++ b/synapse/module_api/__init__.py
@@ -724,7 +724,7 @@ class ModuleApi:
     def is_mine(self, id: Union[str, DomainSpecificString]) -> bool:
         """
         Checks whether an ID comes from this homeserver.
-        
+
         Added in Synapse v1.44.0.
         """
         if isinstance(id, DomainSpecificString):
@@ -733,12 +733,13 @@ class ModuleApi:
             return self._hs.is_mine_id(id)
 
     async def get_user_ip_and_agents(
-        self, user_id: str, since_ts: Optional[float] = None
+        self, user_id: str, since_ts: Optional[float] = 0
     ) -> List[UserIpAndAgent]:
         """
         Return the list of user IPs and agents for a user.
 
-        Only useful for local users.
+        Only useful for local users. If since_ts is not specified,
+        return the list since the epoch.
 
         Added in Synapse v1.44.0.
         """
@@ -752,7 +753,7 @@ class ModuleApi:
             pass
         if is_mine:
             raw_data = await self._store.get_user_ip_and_agents(
-                UserID.from_string(user_id)
+                UserID.from_string(user_id), since_ts
             )
             # Sanitize some of the data. We don't want to return tokens.
             return [
@@ -762,7 +763,6 @@ class ModuleApi:
                     last_seen=int(data["last_seen"]),
                 )
                 for data in raw_data
-                if since_ts is None or int(data["last_seen"]) >= since_ts
             ]
         else:
             return []
--- a/synapse/module_api/__init__.py
+++ b/synapse/module_api/__init__.py
@@ -733,7 +733,7 @@ class ModuleApi:
             return self._hs.is_mine_id(id)
 
     async def get_user_ip_and_agents(
-        self, user_id: str, since_ts: Optional[float] = 0
+        self, user_id: str, since_ts: int = 0
     ) -> List[UserIpAndAgent]:
         """
         Return the list of user IPs and agents for a user.
@@ -751,6 +751,7 @@ class ModuleApi:
                 is_mine = True
         except Exception:
             pass
+
         if is_mine:
             raw_data = await self._store.get_user_ip_and_agents(
                 UserID.from_string(user_id), since_ts
--- a/synapse/module_api/__init__.py
+++ b/synapse/module_api/__init__.py
@@ -27,6 +27,7 @@ from typing import (
     Union,
 )
 
+import attr
 import jinja2
 
 from twisted.internet import defer
@@ -87,20 +88,16 @@ __all__ = [
 logger = logging.getLogger(__name__)
 
 
+@attr.s(auto_attribs=True)
 class UserIpAndAgent:
     """
-    A ip, user_agent pair used by a user to connect to this homeserver.
+    An IP address and user agent used by a user to connect to this homeserver.
     """
 
     ip: str
     user_agent: str
     # The time at which this user agent/ip was last seen.
     last_seen: int
-
-    def __init__(self, ip: str, user_agent: str, last_seen: int):
-        self.ip = ip
-        self.user_agent = user_agent
-        self.last_seen = last_seen
 
 
 class ModuleApi:
@@ -725,7 +722,11 @@ class ModuleApi:
         )
 
     def is_mine(self, id: Union[str, DomainSpecificString]) -> bool:
-        """Checks whether an ID comes from this homeserver."""
+        """
+        Checks whether an ID comes from this homeserver.
+        
+        Added in Synapse v1.44.0.
+        """
         if isinstance(id, DomainSpecificString):
             return self._hs.is_mine(id)
         else:
@@ -738,6 +739,8 @@ class ModuleApi:
         Return the list of user IPs and agents for a user.
 
         Only useful for local users.
+
+        Added in Synapse v1.44.0.
         """
         # Don't hit the db if this is not a local user.
         is_mine = False
--- a/synapse/module_api/__init__.py
+++ b/synapse/module_api/__init__.py
@@ -723,7 +723,13 @@ class ModuleApi:
 
     def is_mine(self, id: Union[str, DomainSpecificString]) -> bool:
         """
-        Checks whether an ID comes from this homeserver.
+        Checks whether an ID (user id, room, ...) comes from this homeserver.
+
+        Args:
+            id: any Matrix id (e.g. user id, room id, ...), either as a raw id,
+                e.g. string "@user:example.com" or as a parsed UserID, RoomID, ...
+        Returns:
+            True if id comes from this homeserver, False otherwise.
 
         Added in Synapse v1.44.0.
         """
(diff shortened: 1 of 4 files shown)


synapse/module_api/__init__.py
```diff
@@ -700,6 +715,38 @@ def read_templates(
...
+        # Sanitize some of the data. We don't want to return tokens.
+        return [
+            {
+                "ip": str(data["ip"]),
+                "user_agent": str(data["user_agent"]),
+                "last_seen": int(data["last_seen"]),
+            }
+            for data in raw_data
+            if since_ts is None or int(data["last_seen"]) >= since_ts
+        ]
```
babolivier, line 748: This returns a list of dicts, not a list of `UserIpAndAgent`

synapse/module_api/__init__.py
```diff
@@ -700,6 +715,38 @@ def read_templates(
...
+        """Return a timestamp for the current time, in milliseconds since the epoch"""
+        return self._hs.get_clock().time_msec()
+
+    def is_mine(self, domain_specific_string: DomainSpecificString) -> bool:
+        """Checks whether an ID comes from this homeserver."""
+        return self._hs.is_mine(domain_specific_string)
+
+    def is_mine_id(self, string: str) -> bool:
+        """Checks whether an ID comes from this homeserver."""
+        return string.split(":", 1)[1] == self._server_name
```
babolivier, line 728: Could you please rename this into `is_mine_str` or similar and extend the docstring to explain how it differs from `is_mine`? As it stands it feels a bit confusing to the external module developer.

synapse/module_api/__init__.py
```diff
@@ -700,6 +715,38 @@ def read_templates(
             (td for td in (self.custom_template_dir, custom_template_directory) if td),
         )
 
+    def ts(self) -> int:
+        """Return a timestamp for the current time, in milliseconds since the epoch"""
+        return self._hs.get_clock().time_msec()
```
babolivier, line 720: We've usually been biased towards telling people to just use `time.time()` rather than include a module API that does the exact same thing, so I'd suggest dropping this method.

synapse/module_api/__init__.py
```diff
@@ -79,6 +87,13 @@
 logger = logging.getLogger(__name__)
 
 
+class UserIpAndAgent(TypedDict):
```
babolivier, line 90: Could you add a docstring to this class please

tests/module_api/test_api.py
```diff
@@ -89,6 +89,19 @@ def test_get_userinfo_by_id__no_user_found(self):
...
+        info = self.get_success(self.module_api.get_user_ip_and_agents(user_id))
+        self.assertIdentical(info, [])
+
+    def test_get_user_ip_and_agents__no_user_found(self):
+        info = self.get_success(
+            self.module_api.get_user_ip_and_agents(
+                "@test_get_user_ip_and_agents_user_nonexistent"
+            )
+        )
+        self.assertIdentical(info, [])
```
babolivier, line 103: It looks like these only test that the module API returns nothing when it should, could you also add tests to make sure it returns the right data when it should return data please?

babolivier (changes requested):

synapse/module_api/__init__.py
anoadragon453: Or extend `is_mine` to allow passing user IDs are `str`s as well?

babolivier: Ah yeah that'd probably be a better way of doing it.

erikjohnston: TBF `is_mine_id` is consistent with the function in `synapse.types`, and the convention where e.g. `user` is an object and `user_id` is a string. :shrug:

synapse/module_api/__init__.py
Yoric: Almost sure it does :)

synapse/module_api/__init__.py
Yoric: Ah, I hadn't realized, thanks.

Yoric:
Note: Not clear to me why  `Tests / sytest (buster, postgres, workers, redis) (pull_request)` [failed](https://github.com/matrix-org/synapse/pull/10833/checks?check_run_id=3622768018). Not sure it's something I can action.

DMRobertson:
> Note: Not clear to me why `Tests / sytest (buster, postgres, workers, redis) (pull_request)` [failed](https://github.com/matrix-org/synapse/pull/10833/checks?check_run_id=3622768018). Not sure it's something I can action.

Looks like a flakey test, see #1116

babolivier (changes requested):

synapse/module_api/__init__.py
```diff
@@ -700,6 +724,46 @@ def read_templates(
...
+                UserIpAndAgent(
+                    ip=str(data["ip"]),
+                    user_agent=str(data["user_agent"]),
+                    last_seen=int(data["last_seen"]),
+                )
+                for data in raw_data
+                if since_ts is None or int(data["last_seen"]) >= since_ts
+            ]
+        else:
+            return []
```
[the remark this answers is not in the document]
Yoric: Yeah, that was my initial version. Then I rewrote it because I don't want to accidentally catch everything.

synapse/module_api/__init__.py
babolivier: Yeah tbh I find how it's done in `synapse.types` confusing, so I think it makes sense to make it not confusing when the target audience isn't Synapse devs.

synapse/module_api/__init__.py
babolivier: Well sure it does now :p

Yoric: I'm almost sure that you can coerce a dictionary into a subclass of `TypedDict`.

Yoric:
I have just modified things to decrease the amount of data we'll request when a user has many IPs/Agents.

synapse/storage/databases/main/client_ips.py
```diff
@@ -555,8 +555,11 @@ async def get_last_client_ip_by_device(
         return ret
 
     async def get_user_ip_and_agents(
-        self, user: UserID
+        self, user: UserID, since_ts: Optional[float] = 0
```
babolivier, line 558: If it's an `Optional`, surely the default should be `None`? And why a `float` and not an `int`? `last_seen` is a `BIGINT`.

synapse/storage/databases/main/client_ips.py
```diff
@@ -568,13 +571,23 @@ async def get_user_ip_and_agents(
...
+            txn.execute(
+                """
+                SELECT access_token, ip, user_agent, last_seen FROM user_ips
+                WHERE last_seen >= ? AND user_id = ?
+                ORDER BY last_seen
+                DESC
+                """,
+                (since_ts, user_id),
+            )
+            return txn.fetchall()
```
babolivier, line 587: This doesn't seem to support the case where `since_ts` is `None`.

synapse/module_api/__init__.py
```diff
@@ -700,6 +721,52 @@ def read_templates(
...
+        """
+        # Don't hit the db if this is not a local user.
+        is_mine = False
+        try:
+            # Let's be defensive against ill-formed strings.
+            if self.is_mine(user_id):
+                is_mine = True
+        except Exception:
+            pass
+        if is_mine:
```
babolivier, line 754:
```suggestion
            pass

        if is_mine:
```

(for readability)

synapse/module_api/__init__.py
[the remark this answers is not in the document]
babolivier: Hmm, fair.

babolivier (changes requested):
Apparently GitHub glitched and sent my review without letting me label it as "Changes requested" :slightly_frowning_face:

synapse/storage/databases/main/client_ips.py
Yoric: Sorry, got confused when someone (I forgot whom) told me to "just use `time.time()`" for timestamps, which returns floating-point seconds instead of integer milliseconds.

babolivier:
(setting the PR as ready for review since it's already got a few rounds of review)

babolivier (approved):
LGTM!

synapse/module_api/__init__.py
```diff
@@ -700,6 +721,53 @@ def read_templates(
             (td for td in (self.custom_template_dir, custom_template_directory) if td),
         )
 
+    def is_mine(self, id: Union[str, DomainSpecificString]) -> bool:
+        """
+        Checks whether an ID comes from this homeserver.
```
anoadragon453, line 726:
```suggestion
        Checks whether an ID (user, room, etc.) comes from this homeserver.
```

synapse/module_api/__init__.py
```diff
@@ -700,6 +721,53 @@ def read_templates(
...
+
+        Added in Synapse v1.44.0.
```
anoadragon453, line 728: Could you add `Args` and `Returns` sections to the docstring of this function and the one below please?

I was initially confused that this method wasn't just for user IDs :slightly_smiling_face:

synapse/module_api/__init__.py
```diff
@@ -700,6 +721,53 @@ def read_templates(
...
+            return self._hs.is_mine_id(id)
+
+    async def get_user_ip_and_agents(
+        self, user_id: str, since_ts: int = 0
+    ) -> List[UserIpAndAgent]:
+        """
+        Return the list of user IPs and agents for a user.
+
+        Only useful for local users. If since_ts is not specified,
+        return the list since the epoch.
```
anoadragon453, line 742: It may be worth mentioning that if a non-local user is provided, only an empty list will be returned.

synapse/storage/databases/main/client_ips.py
```diff
@@ -555,8 +555,11 @@ async def get_last_client_ip_by_device(
         return ret
 
     async def get_user_ip_and_agents(
-        self, user: UserID
+        self, user: UserID, since_ts: int = 0
     ) -> List[Dict[str, Union[str, int]]]:
+        """
+        Fetch IP/User Agent connection since a given timestamp.
```
anoadragon453, line 561: `Args`/`Returns` sections here too please.

anoadragon453 (commented):
Poking my head in with some documentation concerns, but otherwise the logic looks sound.

PR #10834, opened 2021-09-16
Create a constant for a small png image in tests.

clokep:
From a review comment on https://github.com/matrix-org/synapse/pull/10814#discussion_r710113115, but seemed worth factoring out and using it in other places too.

This only touches tests so should be good to go if unit tests pass.

### DIFF
--- a/tests/rest/admin/test_user.py
+++ b/tests/rest/admin/test_user.py
@@ -1,4 +1,4 @@
-# Copyright 2018 New Vector Ltd
+# Copyright 2018-2021 The Matrix.org Foundation C.I.C.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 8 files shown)


reivilibre (approved):
Thanks!

If tests are good, good with me.

Not sure why Portdb is sad, but probably unrelated...?

tests/rest/admin/test_user.py
```diff
@@ -2835,11 +2835,7 @@ def test_order_by(self):
...
 
         # Resolution: 1×1, MIME type: image/png, Extension: png, Size: 67 B
-        image_data1 = unhexlify(
-            b"89504e470d0a1a0a0000000d4948445200000001000000010806"
-            b"0000001f15c4890000000a49444154789c63000100000500010d"
-            b"0a2db40000000049454e44ae426082"
-        )
+        image_data1 = SMALL_PNG
         # Resolution: 1×1, MIME type: image/gif, Extension: gif, Size: 35 B
         image_data2 = unhexlify(
```
clokep, line 2840: I should mention that I checked other images in the tests and all of them are unique, so nothing to factor out.

reivilibre:
> I seem to have made all of CI unhappy. I'm going to let it churn for a bit and then restart.

Probably got cancelled because Portdb failed (looked like starting Postgres timed out, think that happened 'randomly' to me before). I'll give it a ~~kick~~ _restart_.

PR #3701, opened 2018-08-15
Use a producer to stream back responses

richvdh:
The problem with dumping all of the json response into the Request object at once is that doing so starts the timeout for the next request to be received: so if it takes longer than 60s to stream back the response to the client, the client never gets it.

The correct solution is to use a Producer; then the timeout is only started once all of the content is sent over the TCP connection.

(This PR doesn't strictly depend on #3700, but without it, the logging will be a bit messed up.)

PR #10844, opened 2021-09-17
Encode JSON responses on a thread in C

erikjohnston:
Currently we use `JsonEncoder.iterencode` to write JSON responses, which ensures that we don't block the main reactor thread when encoding huge objects. The downside to this is that `iterencode` falls back to using a pure Python encoder that is *much* less efficient and can easily burn a lot of CPU for huge responses. To fix this, while still ensuring we don't block the reactor loop, we encode the JSON on a threadpool using the standard `JsonEncoder.encode` functions, which is backed by a C library.

Doing so, however, requires `respond_with_json` to have access to the reactor, which it previously didn't. There are two ways of doing this:

1. threading through the reactor object, which is a bit fiddly as e.g. `DirectServeJsonResource` doesn't currently take a reactor, but is exposed to modules and so is a PITA to change; or
2. expose the reactor in `SynapseRequest`, which requires updating a bunch of servlet types.

I went with the latter as that is just a mechanical change, and I think makes sense as a request already has a reactor associated with it (via its http channel).

### DIFF
--- a/synapse/http/server.py
+++ b/synapse/http/server.py
@@ -320,7 +320,7 @@ class DirectServeJsonResource(_AsyncResource):
 
     def _send_response(
         self,
-        request: Request,
+        request: SynapseRequest,
         code: int,
         response_object: Any,
     ):
@@ -629,7 +629,7 @@ def _encode_json_bytes(json_object: Any) -> Iterator[bytes]:
 
 
 def respond_with_json(
-    request: Request,
+    request: SynapseRequest,
     code: int,
     json_object: Any,
     send_cors: bool = False,
--- a/synapse/http/server.py
+++ b/synapse/http/server.py
@@ -37,7 +37,7 @@ from typing import (
 )
 
 import jinja2
-from canonicaljson import iterencode_canonical_json
+from canonicaljson import encode_canonical_json
 from typing_extensions import Protocol
 from zope.interface import implementer
 
@@ -620,12 +620,11 @@ class _ByteProducer:
         self._request = None
 
 
-def _encode_json_bytes(json_object: Any) -> Iterator[bytes]:
+def _encode_json_bytes(json_object: Any) -> bytes:
     """
     Encode an object into JSON. Returns an iterator of bytes.
     """
-    for chunk in json_encoder.iterencode(json_object):
-        yield chunk.encode("utf-8")
+    return json_encoder.encode(json_object).encode("utf-8")
 
 
 def respond_with_json(
@@ -659,7 +658,7 @@ def respond_with_json(
         return None
 
     if canonical_json:
-        encoder = iterencode_canonical_json
+        encoder = encode_canonical_json
     else:
         encoder = _encode_json_bytes
 
@@ -670,7 +669,7 @@ def respond_with_json(
     if send_cors:
         set_cors_headers(request)
 
-    _ByteProducer(request, encoder(json_object))
+    _write_json_to_request_in_thread(request, encoder, json_object)
     return NOT_DONE_YET
 
 
--- a/synapse/http/server.py
+++ b/synapse/http/server.py
@@ -56,7 +56,7 @@ from synapse.api.errors import (
     UnrecognizedRequestError,
 )
 from synapse.http.site import SynapseRequest
-from synapse.logging.context import preserve_fn
+from synapse.logging.context import defer_to_thread, preserve_fn, run_in_background
 from synapse.logging.opentracing import trace_servlet
 from synapse.util import json_encoder
 from synapse.util.caches import intern_dict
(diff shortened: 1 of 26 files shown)


clokep:
Is the slow bit iterating a list of events? I wonder if we could iterate that manually and spit out the [ / ] manually as well, encoding each event with the fast C encoder. (Essentially a custom producer for this case.)

erikjohnston:
> Is the slow bit iterating a list of events? I wonder if we could iterate that manually and spit out the [ / ] manually as well, encoding each event with the fast C encoder. (Essentially a custom producer for this case.)

This isn't specific to events, any large response will suffer from the same problem. Most of the time this will be due to events, but special casing things like `/send_join` and initial syncs just feels a bit tedious, though would work :shrug:

synapse/push/emailpusher.py
```diff
@@ -184,7 +184,7 @@ async def _unsafe_process(self) -> None:
 
             should_notify_at = max(notif_ready_at, room_ready_at)
 
-            if should_notify_at < self.clock.time_msec():
+            if should_notify_at <= self.clock.time_msec():
```
erikjohnston, line 187: This looks like it has nothing to do with the changes in the rest of the PR right?? WRONG. Without this change the tests would tight loop and then OOM. What I think is going on is that moving the encoding of JSON to a thread has subtly changed the timings of the tests, causing us to hit this line such that `should_notify_at` is *now*, causing us to not send the response but then schedule the function to be rerun in 0 seconds, causing the tests to enter a tight loop and never make any progress.

synapse/http/server.py
```diff
@@ -809,3 +810,28 @@ def finish_request(request: Request):
...
+):
+    """Encodes the given JSON object on a thread and then writes it to the
+    request.
+
+    This is done so that encoding large JSON objects doesn't block the reactor
+    thread.
+
+    Note: We don't use JsonEncoder.iterencode here as that falls back to the
+    Python implementation (rather than the C backend), which is *much* more
+    expensive.
```
reivilibre, line 828: and presumably holds the global interpreter lock, making thread pooling it useless?

reivilibre (approved):
Seems reasonable to me

synapse/http/server.py
erikjohnston: ja

erikjohnston:
Note to self I should probably change this to use byte producer rather than `request.write` all at once

richvdh:
> The downside to this is that iterencode falls back to using a pure Python encoder that is much less efficient and can easily burn a lot of CPU for huge responses.

I'm surprised about that. I'd have thought we'd have checked it before switching to iterencode, and it doesn't fit with my memory of how the json encoder works. I'll assume you've double-checked, though!

erikjohnston:
> > The downside to this is that iterencode falls back to using a pure Python encoder that is much less efficient and can easily burn a lot of CPU for huge responses.
> 
> I'm surprised about that. I'd have thought we'd have checked it before switching to iterencode, and it doesn't fit with my memory of how the json encoder works. I'll assume you've double-checked, though!

Yeah, it is surprising. FWIW the key bit here is, where `_one_shot` is always `False` for `iterencode` (I *think* because the C lib does the encoding all at once):

https://github.com/python/cpython/blob/5822ab672a1d26ff1837103c1ed8e4c3c2a42b87/Lib/json/encoder.py#L246-L256

tests/test_server.py
```diff
@@ -105,7 +105,7 @@ def _throw(*args):
         def _callback(request, **kwargs):
             d = Deferred()
             d.addCallback(_throw)
-            self.reactor.callLater(1, d.callback, True)
+            self.reactor.callLater(0.5, d.callback, True)
```
erikjohnston, line 108: This is needed because we time out responses that take 1s or longer to process. I'm not 100% why the change in this PR would make this suddenly start failing, but :shrug:

erikjohnston:
I've updated this with a commit to use a producer rather than writing all the content at once, for the reasons mentioned in #3701

synapse/http/server.py
```diff
@@ -809,3 +809,31 @@ def finish_request(request: Request):
...
+    """
+
+    async def _inner_encode_in_thread():
+        json_str = await defer_to_thread(request.reactor, json_encoder, json_object)
+
+        try:
+            request.write(json_str)
+            request.finish()
+        except RuntimeError as e:
+            logger.info("Connection disconnected before response was written: %r", e)
```
richvdh: is this really the only possible reason that a RuntimeError can be raised?

synapse/http/server.py
```diff
@@ -706,15 +707,35 @@ def respond_with_json_bytes(
...
-    # the original `bytes`).
-    bytes_io = BytesIO(json_bytes)
-
-    producer = NoRangeStaticProducer(request, bytes_io)
-    producer.start()
+    _write_json_bytes_to_request(request, json_bytes)
     return NOT_DONE_YET
 
 
+def _write_json_bytes_to_request(request: Request, json_bytes: bytes) -> None:
```
richvdh, line 714: is this specific to being json bytes, or could it be used for any byte sequence?

synapse/http/server.py
```diff
@@ -706,15 +707,35 @@ def respond_with_json_bytes(
     if send_cors:
         set_cors_headers(request)
 
-    # note that this is zero-copy (the bytesio shares a copy-on-write buffer with
-    # the original `bytes`).
-    bytes_io = BytesIO(json_bytes)
-
-    producer = NoRangeStaticProducer(request, bytes_io)
-    producer.start()
+    _write_json_bytes_to_request(request, json_bytes)
```
richvdh, line 710: why is this changing in this PR? It seems unrelated to how we encode JSON responses?

richvdh (changes requested):
I think I need to ask you to break this up. There's a lot changing, including a bunch of prep work, and going through commit-by-commit isn't working, as the implementation has obviously evolved a bit as you've been working on it. sorry.

erikjohnston:
I've split this up into separate PRs, with the final one being: https://github.com/matrix-org/synapse/pull/10905

PR #10845, opened 2021-09-17
Fix #10837 by adding JSON encoding/decoding to the Module API example…

### LINKED ISSUE #10837: The Module API docs' DemoResource crashes
### Description

The Module API docs include an example module, which is extremely helpful for a new module developer.

However, the DemoResource's `render_GET` function causes a `TypeError` during JSON encoding and fails to return the intended response.

### Steps to reproduce

- Install the DemoModule.
  - I put the example code in `synapse/demo_module/__init__.py`
  - Copy the example config verbatim into `homeserver.yaml`
- Start Synapse: 
```console
$ python3 -m synapse.app.homeserver --config-path homeserver.yaml
```
- Load the demo URL `_synapse/client/demo/hello?name=world` in a browser or use cURL/wget/etc

- **Homeserver**: My own development server

If not matrix.org:

- **Version**: 1.42.0

- **Install method**: Git clone

- **Platform**: Debian 10, x86, Python 3.7.3

### Proposed Fix

Since the problem seems to be with the JSON encoder failing to handle bytes, we can simply decode the string before we put it into the JSON.  Also, it seems that we need to encode the JSON string back to bytes before we return.

```python
    def render_GET(self, request: Request):
        name = request.args.get(b"name")[0].decode() 
        request.setHeader(b"Content-Type", b"application/json") 
        return json.dumps({"hello": name}).encode()
```

cvwright:
This PR fixes crashes in the Module API example code.  All that was missing was a little JSON encoding/decoding.

Signed-off-by: Charles Wright <cvwright@kombucha.systems>

### Pull Request Checklist

### DIFF
--- a/changelog.d/10845.doc
+++ b/changelog.d/10845.doc
@@ -1,0 +1,1 @@
+Fix some crashes in the Module API example code, by adding JSON encoding/decoding.
--- a/docs/modules/spam_checker_callbacks.md
+++ b/docs/modules/spam_checker_callbacks.md
@@ -136,9 +136,9 @@ class IsUserEvilResource(Resource):
         self.evil_users = config.get("evil_users") or []
 
     def render_GET(self, request: Request):
-        user = request.args.get(b"user")[0]
+        user = request.args.get(b"user")[0].decode()
         request.setHeader(b"Content-Type", b"application/json")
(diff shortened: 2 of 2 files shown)


anoadragon453 (approved):
Looks good to me. Thanks for spotting and sending a fix!

PR #10847, opened 2021-09-17
Add type annotations to `synapse.metrics`

squahtx:


synapse/metrics/background_process_metrics.py
```diff
@@ -169,7 +188,19 @@ def update_metrics(self):
...
 
-def run_as_background_process(desc: str, func, *args, bg_start_span=True, **kwargs):
+R = TypeVar("R")
+
+
+def run_as_background_process(
+    desc: str,
+    func: Union[
+        Callable[..., Awaitable[R]],
+        # Callable[..., R],
```
squahtx, line 198: Uncommenting this results in
```
synapse/util/distributor.py:74: error: Argument 2 to "run_as_background_process" has incompatible type "Callable[[VarArg(Any), KwArg(Any)], Deferred[List[Any]]]"; expected "Callable[..., Awaitable[<nothing>]]"  [arg-type]
```
and a bunch more errors.

On the other hand, omitting it from the union results in
```
synapse/util/caches/expiringcache.py:86: error: Argument 2 to "run_as_background_process" has incompatible type "Callable[[], None]"; expected "Callable[..., Awaitable[Any]]"  [arg-type]
```

synapse/storage/util/id_generators.py
```diff
@@ -448,6 +448,9 @@ def get_next_txn(self, txn: LoggingTransaction):
         # position points to a persisted row with the correct instance name.
         if self._writers:
             txn.call_after(
+                # Argument 1 to "call_after" of "LoggingTransaction" has incompatible
+                # type "Callable[[str, Callable[..., Awaitable[R]], VarArg(Any), DefaultNamedArg(bool, 'bg_start_span'), KwArg(Any)], Deferred[R]]";
+                # expected "Callable[..., None]"  [arg-type]
```
squahtx, line 453: `run_as_background_process` returns a `Deferred[None]`, but `call_after` expects a function returning `None`

Shall we update the types of `call_after` to allow functions returning anything?
Or wrap this in a function returning `None`?

synapse/metrics/background_process_metrics.py
```diff
@@ -216,6 +247,7 @@ async def run():
                     "Background process '%s' threw an exception",
                     desc,
                 )
+                raise
```
squahtx, line 250: We're missing a `return` or a `raise` in this branch.
raising here makes looping calls stop, so we should return `None` instead.

synapse/metrics/background_process_metrics.py
squahtx: Fixed by using overloads and suppressing the error about overlapping signatures.

synapse/groups/attestations.py
```diff
@@ -166,7 +168,7 @@ async def on_renew_attestation(
 
         return {}
 
-    def _start_renew_attestations(self) -> None:
+    def _start_renew_attestations(self) -> Deferred[None]:
```
clokep, line 171: I think you need quotes for old-Twisted.

```suggestion
    def _start_renew_attestations(self) -> "Deferred[None]":
```

synapse/metrics/__init__.py
```diff
@@ -53,7 +66,7 @@
 
 class RegistryProxy:
     @staticmethod
-    def collect():
+    def collect() -> Iterable[Metric]:
```
clokep, line 69: I would think this needs to be a `Generator`, but maybe not?

This is true for a variety of spots we use `yield`.

synapse/storage/database.py
```diff
@@ -179,7 +179,7 @@ def __getattr__(self, name):
 
 
 # The type of entry which goes on our after_callbacks and exception_callbacks lists.
-_CallbackListEntry = Tuple[Callable[..., None], Iterable[Any], Dict[str, Any]]
+_CallbackListEntry = Tuple[Callable[..., object], Iterable[Any], Dict[str, Any]]
```
clokep, line 182: Curious why this is `object` and not `Any`?

synapse/metrics/background_process_metrics.py
```diff
@@ -225,19 +281,25 @@ async def run():
...
+
+def wrap_as_background_process(desc: str) -> Callable[[F], F]:
     """Decorator that wraps a function that gets called as a background
     process.
 
     Equivalent of calling the function with `run_as_background_process`
     """
 
-    def wrap_as_background_process_inner(func):
+    # NB: Return type is incorrect and should be F with a Deferred[Optional[R]] return
```
clokep, line 294: I think this should work, but maybe not? This makes me a bit nervous that we're not mentioning this returns a `Deferred`.

synapse/metrics/background_process_metrics.py
```diff
@@ -189,11 +243,12 @@ def run_as_background_process(desc: str, func, *args, bg_start_span=True, **kwar
         args: positional args for func
         kwargs: keyword args for func
 
-    Returns: Deferred which returns the result of func, but note that it does not
-        follow the synapse logcontext rules.
+    Returns: Deferred which returns the result of func, or `None` if func raises.
+        Note that the returned Deferred does not follow the synapse logcontext
+        rules.
```
clokep, line 248:
```suggestion
    Returns:
        Deferred which returns the result of func, or `None` if func raises.
        Note that the returned Deferred does not follow the synapse logcontext
        rules.
```

synapse/metrics/background_process_metrics.py
```diff
@@ -169,7 +189,41 @@ def update_metrics(self):
         )
 
 
-def run_as_background_process(desc: str, func, *args, bg_start_span=True, **kwargs):
+R = TypeVar("R")
+
+
+@overload
+def run_as_background_process(  # type: ignore[misc]
```
clokep, line 196: What's being ignored here?

squahtx: It's to suppress `error: Overloaded function signatures 1 and 2 overlap with incompatible return types  [misc]`. I shall add a comment

I initially tried a `Union` but ran into a bug/deficiency in mypy where it couldn't figure out what `R` was at every call site.
https://github.com/matrix-org/synapse/pull/10847#discussion_r711249811

synapse/metrics/__init__.py
clokep: (Or maybe I can significantly simplify type hints in other spots!)

squahtx: I opted for using the most basic type that callers want, to minimize assumptions, rather than documenting the concrete return type.

synapse/storage/database.py
squahtx: It's to make mypy complain if we ever try to use the return values of the callbacks.

Everything is assignable to `object`, but not the other way around, whereas `Any` is assignable to everything without any checks.

synapse/metrics/background_process_metrics.py
squahtx: Not sure what to do about this. I updated the docstring above to call the change of return type out.

We can either annotate the return type correctly and lose the types of the arguments for the decorated function,  
or annotate the return type incorrectly and preserve the types of the arguments.

synapse/groups/attestations.py
squahtx: Thanks, I didn't realise that that was the cause of the slew of duplicate-metric CI test failures.

synapse/storage/database.py
reivilibre: `object` is the supertype of everything is what I think you're saying, so we'd get shouted at if we tried to do `x: int = (something returning object)` whereas `Any` is entirely untyped & unchecked.

I think this makes sense.

reivilibre (commented):
This all seems to make sense. I'm a little bit concerned about the lie and would probably want to ask someone else on the team about that ...

synapse/metrics/_exposition.py
```diff
@@ -41,7 +43,7 @@
 MINUS_INF = float("-inf")
 
 
-def floatToGoString(d):
+def floatToGoString(d: Any) -> str:
```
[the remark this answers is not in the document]
squahtx: It's `typeof(Sample.value)` and I had a really hard time figuring out what that was, since `prometheus_client` doesn't have typings.

Though it seems that `prometheus_client` pinky promises that it's an `int` or `float`:
https://github.com/prometheus/client_python/blob/master/prometheus_client/samples.py#L35-L39
I shall update the function!

synapse/metrics/__init__.py
```diff
@@ -100,6 +113,10 @@ def _register(self):
         all_gauges[self.name] = self
 
 
+# Placeholder for `InFlightGauge._metrics_class`, which is created dynamically
+_MetricsEntry = Any
```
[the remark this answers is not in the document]
DMRobertson: I'm taking a look at adding more typing to `synapse.utils` and got stuck on annotating [`metrics` here](https://github.com/matrix-org/synapse/blob/bdfde6dca11a9468372b3c9b327ad3327cbdbe4a/synapse/util/metrics.py#L171). I think it should be one of these dynamically-created types but can't annotate it meaningfully.

So +1 to any other smart approaches!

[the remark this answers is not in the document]
squahtx: Python typings are really frustrating. The equivalent in typescript would be something like: [this](https://www.typescriptlang.org/play?#code/MYGwhgzhAECSB2AxEBLA5gCwC4HEwFc0BTAHgGV8AjAWSKwCcVgBpIgTxiIA8sj4ATGBAYp4aAHwAoAN6To86MAD28YfXzAsS+gAoAtnUbAIALmgUahpqw4BtALoBKOQtkLoAX0kv59ImhRhIl1gMBAQSjBgAGszfStjM2lbaPZoUXMqWhEWdgh7M3h8PUpgj0doAF5xaAA3JRR+Z3c3BS8vSWVVLHT4AH0AM1RMHsroeCIAdzgkYew8QiIdWwAiPzC+rBQDPr0wLhWAGmg1og2tnYhilacAbm9JAHpH6ABhbT9NRTCIqOjoCDoeBgLD4PySAb4eCaFAqaD4AAO-BBRD6okGcyw8Rypmg0hO6xAm22qL2B0KxVK9GOp3OJL6Vz0KwpJWCnkcZnqjRkPmgBhxADpCcSdmToABqMYAJnu7n5RggQrORIuqMZEul9y86KG6GwSoCQV0iORvDR-V1I0c928zzeP0iMWgWAwIOgKhAbHhSJRMAAjNABigiCB+BCoTC4SaUeaMXqsH0EWB6FswtiFUkCcqRaT9szxpS2eVOQ0w615PKmIrhardvsNdAZZJtRbMQbArxjT6zTrMYnk6mQNaHnaEMp6J8eqFwo7-oC0MDQeDIdCtlHu6je-HzbUwo101XM7SVfS9KJ80VWfR2SXueW+Qlq9na2f4A2my24yN20adNGe6226iLuqBNPcQA)

Perhaps we can add a type parameter to `InFlightGauge` and make the calling code provide an appropriate `Protocol`?

synapse/metrics/background_process_metrics.py
squahtx: I tried to improve the comments further

synapse/metrics/background_process_metrics.py
squahtx: I tried to improve the docstring further

synapse/metrics/__init__.py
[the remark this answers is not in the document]
squahtx: Ok. `InFlightGauge` now takes a type parameter, so now those callbacks can have proper type annotations.  
mypy forced me to annotate `in_flight` in synapse/util/metrics.py (sorry!).

Two caveats:
  * mypy won't allow the type parameter to be constrained to a `Protocol`
  * there's no check that the names passed to the `InFlightGauge` constructor match the attributes in the protocol

synapse/metrics/_exposition.py
```diff
@@ -39,15 +39,11 @@
 CONTENT_TYPE_LATEST = "text/plain; version=0.0.4; charset=utf-8"
 
 
-INF = float("inf")
-MINUS_INF = float("-inf")
-
-
-def floatToGoString(d: Any) -> str:
+def floatToGoString(d: Union[int, float]) -> str:
```
reivilibre, line 42:
```suggestion
def floatToGoString(d: float) -> str:
```

I think `int` is a subtype of `float`

reivilibre: hm, actually, I'm pretty minded to leave it as is to signal that we explicitly thought about both ints and floats.

Python ints are arbitrary-sized, so it'd feel funny to have it accept float and then still have to call `float(d)` (which is understandable because presumably it will cut off very big ints as 'infinity'?)

reivilibre: Never mind:

```python
>>> float(2 ** 3000)
OverflowError: int too large to convert to float
```

synapse/metrics/__init__.py
```diff
@@ -113,11 +115,12 @@ def _register(self) -> None:
...
 
-# Placeholder for `InFlightGauge._metrics_class`, which is created dynamically
-_MetricsEntry = Any
+# `MetricsEntry` only makes sense when it is a `Protocol`,
+# but `Protocol` can't be used as a `TypeVar` bound.
+MetricsEntry = TypeVar("MetricsEntry")
 
 
-class InFlightGauge:
+class InFlightGauge(Generic[MetricsEntry]):
```
reivilibre, line 123: well done, that seems pretty elegant, considering how *untypable* this stuff is

reivilibre (commented):
going to collect some thoughts about whether we feel that it's OK to lie; otherwise LGTM

synapse/metrics/background_process_metrics.py
```diff
@@ -225,19 +287,30 @@ async def run():
...
+    Equivalent to calling the function with `run_as_background_process`.
+
+    Note that the annotated return type of this function is incorrect.
+    The return type of the function, once wrapped, is actually a
+    `Deferred[Optional[T]]`.
     """
 
-    def wrap_as_background_process_inner(func):
+    # NB: Return type is incorrect and should be F with a Deferred[Optional[R]] return
+    #     We can't expression the return types correctly until Python 3.10.
```
DMRobertson, line 305: I'm guessing this refers to ParamSpec in https://www.python.org/dev/peps/pep-0612/ ?

If so, could we use the backport in (a sufficiently new version of) [typing_extensions](https://github.com/python/typing/blob/master/typing_extensions/README.rst) now?

synapse/storage/database.py
DMRobertson: Does this mean we should generally prefer `object` over `Any`?

synapse/metrics/background_process_metrics.py
squahtx: Yep! The `ParamSpec` classes are available in `typing_extensions`, but mypy doesn't actually support them yet: https://github.com/python/mypy/issues/8645

synapse/metrics/_exposition.py
squahtx: Normally I'd go for `float` over `Union[int, float]`, but then the `float()` cast looks silly.  Without the cast, the function changes behaviour and omits the decimal point for integer values (making some tests fail).

synapse/storage/database.py
squahtx: I haven't a clue. I imagine it's on a case by case basis. `Any` is certainly a last resort for me, since it's equivalent to not annotating types at all.

synapse/metrics/_exposition.py
reivilibre: Ah I see — in that case, LGTM leaving it as a union

synapse/metrics/background_process_metrics.py
squahtx: > mypy doesn't actually support them yet

mypy rejects usage of `ParamSpec`s with `error: The first argument to Callable must be a list of types or "..."  [misc]`

DMRobertson:
What was the status of this change? I think we were keen to see it done and wouldn't want it to fall through the cracks!

richvdh:
> Edit: I guess Olivier wanted to think things over in [#10847 (review)](https://github.com/matrix-org/synapse/pull/10847#pullrequestreview-761842895)

NB that edits don't seen by those of us that read our GH notifications via email.

I think this is somewhere between @squahtx and @reivilibre to sort out currently.

reivilibre:
> NB that edits don't seen by those of us that read our GH notifications via email.
> 
> I think this is somewhere between @squahtx and @reivilibre to sort out currently.

Thanks for pointing this out to me! I didn't realise I was holding it up.

I think the only thing on my mind was https://github.com/matrix-org/synapse/pull/10847#discussion_r713866701 really — if other people are happy with an innocent lie, then I'm happy to roll it through (and it sounds like, bear in mind that we can one day use `ParamSpec`?)

synapse/metrics/background_process_metrics.py
richvdh: I think it's fine as - is.

question though, what happens if you bound `F` to return an Awaitable?

`F = TypeVar("F", bound=Callable[..., Awaitable])`

(or something similar?)

richvdh: > I think it's fine as - is.

to be clear: I'm saying I'm happy with the PR as it stands.

squahtx: >question though, what happens if you bound F to return an Awaitable?
> `F = TypeVar("F", bound=Callable[..., Awaitable])`

Ooh. mypy throws up 3 errors which all look trivial to fix:
```
synapse/util/caches/expiringcache.py:86: error: Argument 2 to "run_as_background_process" has incompatible type "Callable[[], None]"; expected "Callable[..., Awaitable[Any]]"  [arg-type]
synapse/app/_base.py:404: error: Value of type variable "F" of function cannot be "Callable[[VarArg(Any), KwArg(Any)], None]"  [type-var]
synapse/handlers/typing.py:92: error: Value of type variable "F" of function cannot be "Callable[[FollowerTypingHandler], None]"  [type-var]
```

So it sounds like the cleanest thing to do is to restrict `run_as_background_process` to only accept `Awaitable`-returning functions and drop the whole `maybe_awaitable` thing. I'll make the change when I find some time.

squahtx:
Awaiting changes from myself, to restrict `run_as_background_process` to `Awaitable`-returning functions only.

synapse/metrics/background_process_metrics.py
squahtx: I've gotten rid of the `maybe_awaitable`.

reivilibre (approved):
LGTM

synapse/metrics/background_process_metrics.py
```diff
@@ -192,39 +190,9 @@ def update_metrics(self) -> None:
...
-    ...
-
-
 def run_as_background_process(
     desc: str,
-    func: Union[
-        Callable[..., Awaitable[R]],
-        Callable[..., R],
-    ],
+    func: Callable[..., Awaitable[Optional[R]]],
```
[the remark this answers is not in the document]
squahtx: You're right, it ought to be `Awaitable[R]`

mypy gets unhappy elsewhere if I change it:
```
synapse/metrics/background_process_metrics.py:273: error: Argument 2 to "run_as_background_process" has incompatible type "F"; expected "Callable[..., Awaitable[R]]"  [arg-type]
```

PR #10199, opened 2021-06-17
Expose opentracing trace id in response headers

richvdh:
Fixes: #9480

PR #10854, opened 2021-09-20
Return request ID in HTTP response headers

DMRobertson:
We already have something like this when opentracing is enabled, see #10199. But that's only across federation.

When investigating test failures it's really useful to cross-reference HTTP responses with synapse logs. I propose exposing the request id to facilitate that. I've tried to change the request ids to be unique across all workers now, by including an instance name.

See also matrix-org/sytest#1144.

### DIFF
--- a/synapse/http/site.py
+++ b/synapse/http/site.py
@@ -61,9 +61,17 @@ class SynapseRequest(Request):
         logcontext: the log context for this request
     """
 
-    def __init__(self, channel, *args, max_request_body_size=1024, **kw):
+    def __init__(
+        self,
+        channel,
+        *args,
+        instance_name: str,
+        max_request_body_size: int = 1024,
+        **kw,
+    ):
         Request.__init__(self, channel, *args, **kw)
         self._max_request_body_size = max_request_body_size
+        self._instance_name = instance_name
         self.site: SynapseSite = channel.site
         self._channel = channel  # this is used by the tests
         self.start_time = 0.0
@@ -139,8 +147,8 @@ class SynapseRequest(Request):
         # If there's no authenticated entity, it was the requester.
         self.logcontext.request.authenticated_entity = authenticated_entity or requester
 
-    def get_request_id(self):
-        return "%s-%i" % (self.get_method(), self.request_seq)
+    def get_request_id(self) -> str:
+        return f"{self._instance_name}/{self.get_method()}-{self.request_seq}"
 
     def get_redacted_uri(self) -> str:
         """Gets the redacted URI associated with the request (or placeholder if the URI
@@ -208,9 +216,10 @@ class SynapseRequest(Request):
     def render(self, resrc):
         # this is called once a Resource has been found to serve the request; in our
         # case the Resource in question will normally be a JsonResource.
+        request_id = self.get_request_id()
+        self.setHeader("X-Request-ID", request_id)
 
         # create a LogContext for this request
-        request_id = self.get_request_id()
         self.logcontext = LoggingContext(
             request_id,
             request=ContextRequest(
@@ -523,6 +532,7 @@ class SynapseSite(Site):
         server_version_string,
         max_request_body_size: int,
         reactor: IReactorTime,
+        instance_name: str,
     ):
         """
 
@@ -547,7 +557,10 @@ class SynapseSite(Site):
 
         def request_factory(channel, queued) -> Request:
             return request_class(
-                channel, max_request_body_size=max_request_body_size, queued=queued
+                channel,
+                max_request_body_size=max_request_body_size,
+                queued=queued,
+                instance_name=instance_name,
             )
 
         self.requestFactory = request_factory  # type: ignore
--- a/synapse/http/site.py
+++ b/synapse/http/site.py
@@ -14,7 +14,7 @@
 import contextlib
 import logging
 import time
-from typing import Optional, Tuple, Union
+from typing import Optional, Tuple, Union, cast
 
 import attr
 from zope.interface import implementer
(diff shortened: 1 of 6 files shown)


synapse/http/site.py
```diff
@@ -83,11 +91,11 @@ def __init__(self, channel, *args, max_request_body_size=1024, **kw):
         self._is_processing = False
 
         # the time when the asynchronous request handler completed its processing
-        self._processing_finished_time = None
+        self._processing_finished_time: Optional[float] = None
 
         # what time we finished sending the response to the client (or the connection
         # dropped)
-        self.finish_time = None
+        self.finish_time: Optional[float] = None
```
DMRobertson, line 98: Not sure why mypy started being unhappy, but it was interpreting these as being of type `NoneType` throughout the program, causing problems later.

clokep:
> But that's only across federation.

I don't think this is true. `inject_response_headers` gets called from two places:

* `BaseFederationServlet` -- used by federation code as you mention.
* `trace_servlet` -- which is called by `_AsyncResource` which is how most of the C-S API is implemented.

It is only used if opentracing is enabled (and I'm not 100% sure what is included as the header).

changelog.d/10854.misc
```diff
@@ -0,0 +1 @@
+Include a request id in Synapse's HTTP responses to aid debugging.
```
reivilibre, line 1:
```suggestion
Include a request ID in Synapse's HTTP responses to aid debugging.
```

synapse/http/site.py
```diff
@@ -139,8 +147,8 @@ def requester(self, value: Union[Requester, str]) -> None:
         # If there's no authenticated entity, it was the requester.
         self.logcontext.request.authenticated_entity = authenticated_entity or requester
 
-    def get_request_id(self):
-        return "%s-%i" % (self.get_method(), self.request_seq)
+    def get_request_id(self) -> str:
+        return f"{self._instance_name}/{self.get_method()}-{self.request_seq}"
```
reivilibre, line 151: Is this the same thing that winds up in the logs?
I can see this perhaps making things a bit noisier, but not sure.

synapse/http/site.py
```diff
@@ -368,7 +381,10 @@ def _finished_processing(self):
 
         # the time between the request handler finishing and the response being sent
         # to the client (nb may be negative)
-        response_send_time = self.finish_time - self._processing_finished_time
+        # TODO can we avoid the cast? Maybe take finish_time as an explicit float param?
+        response_send_time = (
+            cast(float, self.finish_time) - self._processing_finished_time
```
reivilibre, line 386: I think this is slightly less evil.

```suggestion
        assert self.finish_time is not None
        response_send_time = (
            self.finish_time - self._processing_finished_time
```

tests/logging/test_terse_json.py
```diff
@@ -188,7 +188,7 @@ def test_with_request_context(self):
         ]
         self.assertCountEqual(log.keys(), expected_log_keys)
         self.assertEqual(log["log"], "Hello there, wally!")
-        self.assertTrue(log["request"].startswith("POST-"))
+        self.assertIn("POST-", log["request"])
```
reivilibre, line 191: or vs:

```suggestion
        self.assertTrue(log["request"].startswith("worker-test/POST-"))
```

reivilibre (commented):
I don't know if Patrick's comment above invalidates this PR or not, but I took a look.

Not entirely sure about the potential for log pollution; compare `POST-424242` vs `worker-event_persister18/POST-424242`, especially when the worker ID is redundant/implied in the log file itself (as far as I know?).

DMRobertson:
> I don't know if Patrick's comment above invalidates this PR or not, but I took a look.

It sounds like it does.

> Not entirely sure about the potential for log pollution; compare `POST-424242` vs `worker-event_persister18/POST-424242`, especially when the worker ID is redundant/implied in the log file itself (as far as I know?).

What I want here is to be able quickly backtrack from a client HTTP response to the corresponding server-side processing. I've found myself wanting this when investigating sytests. A sytest run makes thousands of requests and it's a pain to have to cross-reference timestamps---never mind working out which workers' log file contains the relevant logs. I want a unique string I can bang into `grep -r`  or `rg` and just find the logs I'm after. 

I think this is useful to have, even in the absence of full opentracing telemetry. It might be that

reivilibre:
> I want a unique string I can bang into `grep -r` or `rg` and just find the logs I'm after.

I think this is a fair use-case. I personally feel somewhat inclined to say that the request ID in the logs themselves shouldn't make mention of the worker ID; but the header should.

If you make the header format `X-Request-Id: POST-424242 event_persister18`, then a `rg POST-424242 event_persister18.log` doesn't seem so bad (assuming the sytest log file names follow a sensible pattern like that).

I think thoughts from someone on the team who does a bit more of that might be useful.

richvdh:
I'm afraid I'm a little unconvinced that we want to be adding this header as well as the existing `Synapse-Trace-Id` header, which, as @clokep has mentioned, *is* returned for C-S requests provided `opentracing` is enabled (even if the opentracing span doesn't end up being sent on to the Jaeger collector).

In order for `Synapse-Trace-Id` to be useful for your usecase, I think two things are needed:
 * We should log the value of the `Synapse-Trace-Id` header in the final "Processed" line for each request. I think you can just fish it out of `self.responseHeaders` in `_finished_processing`
 * We need to actually give requests a trace-id when running in sytest. One way to do this is to enable opentracing under sytest; another is to update `synapse.logging.opentracing.trace_servlet` and `BaseFederationServlet` to generate a trace ID internally even if opentracing isn't enabled. I think the latter would be generally useful and would favour that?

That still leaves you with the problem of figuring out which worker handled a particular request. A few thoughts here:
 * can you just grep across the likely logfiles for the right `Trace-Id`? Generally it shouldn't be too hard to find once you figure out how your Synapse instance is routing the different types of request?
 * we could include the worker name in another response header?

DMRobertson:
Many thanks all.

> another is to update `synapse.logging.opentracing.trace_servlet` and `BaseFederationServlet` to generate a trace ID internally even if opentracing isn't enabled. I think the latter would be generally useful and would favour that?

Absolutely, this sounds ideal.

>     * can you just grep across the likely logfiles for the right `Trace-Id`? Generally it shouldn't be too hard to find once you figure out how your Synapse instance is routing the different types of request?

That's fine by me, assuming the trace IDs are unique across workers (guessing they're a uuid of some kind?).

PR #10856, opened 2021-09-20
Add type hints to event streams.

clokep:
This is a follow-on to #10831 (based on a request https://github.com/matrix-org/synapse/pull/10831#discussion_r709953784 to not use `kwargs: Any`).

It clarifies the type hints for the event streams so that mypy can tell what objects they are. From there we then fix the various type hint errors. Finally, we clarify the type hints for `get_new_events`.

### DIFF
--- a/synapse/handlers/account_data.py
+++ b/synapse/handlers/account_data.py
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import random
-from typing import TYPE_CHECKING, Any, List, Tuple
+from typing import TYPE_CHECKING, Collection, List, Optional, Tuple
 
 from synapse.replication.http.account_data import (
     ReplicationAddTagRestServlet,
@@ -174,7 +174,10 @@ class AccountDataEventSource:
         self,
         user: UserID,
         from_key: int,
-        **kwargs: Any,
+        limit: Optional[int],
+        room_ids: Collection[str],
+        is_guest: bool,
+        explicit_room_id: Optional[str] = None,
     ) -> Tuple[List[JsonDict], int]:
         user_id = user.to_string()
         last_stream_id = from_key
--- a/synapse/handlers/account_data.py
+++ b/synapse/handlers/account_data.py
@@ -171,7 +171,10 @@ class AccountDataEventSource:
         return self.store.get_max_account_data_stream_id()
 
     async def get_new_events(
-        self, user: UserID, from_key: int, **kwargs: Any
+        self,
+        user: UserID,
+        from_key: int,
+        **kwargs: Any,
     ) -> Tuple[List[JsonDict], int]:
         user_id = user.to_string()
         last_stream_id = from_key
--- a/synapse/handlers/account_data.py
+++ b/synapse/handlers/account_data.py
@@ -21,6 +21,7 @@ from synapse.replication.http.account_data import (
     ReplicationRoomAccountDataRestServlet,
     ReplicationUserAccountDataRestServlet,
 )
+from synapse.streams import EventSource
 from synapse.types import JsonDict, UserID
 
 if TYPE_CHECKING:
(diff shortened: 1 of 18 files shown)


synapse/handlers/account_data.py
```diff
@@ -171,7 +171,13 @@ def get_current_key(self, direction: str = "f") -> int:
...
 
     async def get_new_events(
-        self, user: UserID, from_key: int, **kwargs: Any
+        self,
+        user: UserID,
+        from_key: int,
+        limit: Optional[int],
+        room_ids: Collection[str],
+        is_guest: bool,
+        explicit_room_id: Optional[str] = None,
```
reivilibre, line 180: What are these parameters? They seem to my eye unused. Is it not possible to remove them?

reivilibre: Ah, I see, these sources are passed to a for loop in:

https://github.com/matrix-org/synapse/blob/858363d0b7e58fd71875b25d183537bb3b5a397f/synapse/notifier.py#L594-L601

In the `EventSources` class, we seem to keep a list of these but they're typed as `Any`. I think it would be nice to give these a common supertype that specifies the signature of `get_new_event` — I'm happy to take this up in a separate PR though.

reivilibre: (actually, I see you already reworked `EventSources` quite a bit)

synapse/storage/databases/main/receipts.py
```diff
@@ -153,7 +153,7 @@ def f(txn):
         }
 
     async def get_linearized_receipts_for_rooms(
-        self, room_ids: List[str], to_key: int, from_key: Optional[int] = None
+        self, room_ids: Iterable[str], to_key: int, from_key: Optional[int] = None
```
reivilibre, line 156: the docstring is out of sync; should be `room_ids` in the docstring (and perhaps say something other than 'List', though that's not too critical as I think 'list' is still an intuitive word)

synapse/streams/events.py
```diff
@@ -21,20 +23,43 @@
...
+                RoomEventSource,
+                PresenceEventSource,
+                TypingNotificationEventSource,
+                ReceiptEventSource,
+                AccountDataEventSource,
+            ],
+        ],
+        None,
+        None,
+    ]:
```
reivilibre, line 53: I think this below is valid, and a little bit less intimidating (especially since we're not making use of being able to pass values into the generator, or returning from it):

```suggestion
    ) -> Iterator[
        Tuple[
            str,
            Union[
                RoomEventSource,
                PresenceEventSource,
                TypingNotificationEventSource,
                ReceiptEventSource,
                AccountDataEventSource,
            ],
        ],
    ]:
```

reivilibre (approved):
smashing

synapse/streams/events.py
clokep: Yes, I recently learned you can use `Iterator` here.

synapse/handlers/account_data.py
reivilibre: > We keep a list of what?

Event sources

> mypy should know at this point it is one of the sub-types since get_sources returns a Union.

Yes, I should have read on further before commenting since you changed this at the same time :p

> I thought about having a Protocol for this, yes.

This is just my personal preference, but a `Protocol` (i.e. structural subtyping, essentially 'duck typing') seems too weak for this case when we could use 'nominal subtyping' (i.e. actually extend some base class `EventSource`).
I think this would also help IDE completion of 'Find Usages' etc, which is one of the most useful tools (afaik) for a newcomer like me — when it works.

One thing I wonder is whether mypy will complain if you override a method in an incompatible way? I think this would perhaps lead to clearer messages than if you messed up the method signature and it suddenly no longer counts as a subtype of the Protocol.

clokep:
@reivilibre would appreciate another look before merging!

reivilibre (approved):
looks great to me! Thanks for doing this :)

PR #70, opened 2015-02-12
Exception fixes

erikjohnston:

PR #10857, opened 2021-09-20
Integrate the rust state compressor with Synapse.

Azrenbeth:
Add config options to control the run of the automatic tool from the [state compressor](https://github.com/matrix-org/rust-synapse-compress-state/)

Not ready for Synapse until [PR #70](https://github.com/matrix-org/rust-synapse-compress-state/pull/70) merges in that repo

### DIFF
--- a/docs/state_compressor.md
+++ b/docs/state_compressor.md
@@ -1,0 +1,154 @@
+# State compressor
+
+The state compressor is an **experimental** tool that attempts to reduce the number of rows 
+in the `state_groups_state` table inside of a postgres database.
+
+## Introduction to the state tables and compression
+### What is state?
+State is things like who is in a room, what the room topic/name is, who has
+what privilege levels etc. Synapse keeps track of it so that it can spot invalid
+events (e.g. ones sent by banned users, or by people with insufficient privilege).
+
+### What is a state group?
+
+Synapse needs to keep track of the state at the moment of each event. A state group
+corresponds to a unique state. The database table `event_to_state_groups` keeps track
+of the mapping from event ids to state group ids.
+
+Consider the following simplified example:
+```
+State group id   |          State
+_____________________________________________
+       1         |      Alice in room
+       2         | Alice in room, Bob in room
+       3         |        Bob in room
+
+
+Event id |     What the event was
+______________________________________
+    1    |    Alice sends a message
+    3    |     Bob joins the room
+    4    |     Bob sends a message
+    5    |    Alice leaves the room
+    6    |     Bob sends a message
+
+
+Event id | State group id
+_________________________
+    1    |       1
+    2    |       1
+    3    |       2
+    4    |       2
+    5    |       3
+    6    |       3
+```
+### What are deltas and predecessors?
+When a new state event happens (e.g. Bob joins the room) a new state group is created.
+BUT instead of copying all of the state from the previous state group, we just store
+the change from the previous group (saving on lots of storage space!). The difference
+from the previous state group is called the "delta"
+
+So for the previous example we would have the following (Note only rows 1 and 2 will
+make sense at this point):
+
+```
+State group id | Previous state group id |      Delta
+____________________________________________________________
+       1       |          NONE           |   Alice in room
+       2       |           1             |    Bob in room
+       3       |          NONE           |    Bob in room
+```
+So why is state group 3's previous state group NONE and not 2? Well the way that deltas 
+work in synapse is that they can only add in new state or overwrite old state, but they
+cannot remove it. (So if the room topic is changed then that is just overwriting state,
+but removing alice from the room is neither an addition or an overwriting). If it is 
+impossible to find a delta, then you just start from scratch again with a "snapshot" of
+the entire state. 
+
+(NOTE this is not documentation on how synapse handles leaving rooms but is purely for illustrative
+purposes)
+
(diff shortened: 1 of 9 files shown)


synapse/util/state_compressor.py
```diff
@@ -0,0 +1,68 @@
...
+    """Schedules the state compressor to run regularly"""
+
+    # Return if cannot import auto_compressor
+    if not state_compressor:
+        return
+
+    # Return if compressor isn't enabled
+    compressor_config = hs.config.statecompressor
+    if not compressor_config.compressor_enabled:
+        return
```
clokep, line 33: We probably want to only run this only if `hs.config.workers.run_background_tasks` is true so that we don't run it on every worker?

docs/state_compressor.md
```diff
@@ -0,0 +1,156 @@
+TODO: Update with final contents of README after PR #70 merged in rust-synapse-compress-state repo
```
clokep, line 1: If we duplicate the documentation I'm rather afraid it will get out of date. Maybe we can link out or copy it as part of the build?

Azrenbeth: Yeah I wasn't really sure what to do about the docs! I think linking them somehow would be more sensible than duplicating. I was leaving this as undone until the docs from the compressor get finalised so we know what they'll actually say!

clokep: Makes sense! I mostly took a look cause I was curious how the rust was integrated! Figured I'd leave a couple comments while I was there!

synapse/util/state_compressor.py
```diff
@@ -0,0 +1,73 @@
...
+
+    # construct the database URL from the database config
+    db_args = db_config["args"]
+    db_url = "postgresql://{username}:{password}@{host}:{port}/{database}".format(
+        username=db_args["user"],
+        password=db_args["password"],
+        host=db_args["host"],
+        port=db_args["port"],
+        database=db_args["database"],
```
erikjohnston, line 52: Bleurgh, a bunch of these are actually optional values. Annoyingly host is optional in Synapse (which uses psycopg2 which uses libpq) but not in Rust, and its Hard to know what the default host is in libpq without asking it. If we want this to work I think we need to explicitly pull out `db_conn.info.host` of a real connection to the state database.

erikjohnston:
@Azrenbeth I've pushed a couple of commits that attempt to reuse the postgres connection parameters from an existing connection. I haven't had a chance to test it though

Azrenbeth:
> @Azrenbeth I've pushed a couple of commits that attempt to reuse the postgres connection parameters from an existing connection. I haven't had a chance to test it though

That should now be working again!

docs/state_compressor.md
erikjohnston: That PR has now merged

Azrenbeth: I've replaced most of that with a link to the compressor documentation

synapse/util/state_compressor.py
Azrenbeth: I think this now works with only password still being taken from db_args (since psycopg2 doesn't provide that in it's dsn_parameters field - see [here](https://www.psycopg.org/docs/extensions.html#psycopg2.extensions.ConnectionInfo.dsn_parameters))

erikjohnston:
It looks like we also block shutdown until the compression run has finished, which isn't ideal

Azrenbeth:
> It looks like we also block shutdown until the compression run has finished, which isn't ideal

Is there a way to interupt it? There shouldn't be any problem if the compressor gets stopped mid way through.

erikjohnston:
> > It looks like we also block shutdown until the compression run has finished, which isn't ideal
> 
> Is there a way to interupt it? There shouldn't be any problem if the compressor gets stopped mid way through.

There should be a way of doing it, you can mark threads as a "daemon" so they don't block shutdown, but can't remember quite how to do it off the top of my head

Azrenbeth:
> > > It looks like we also block shutdown until the compression run has finished, which isn't ideal
> > 
> > 
> > Is there a way to interupt it? There shouldn't be any problem if the compressor gets stopped mid way through.
> 
> There should be a way of doing it, you can mark threads as a "daemon" so they don't block shutdown, but can't remember quite how to do it off the top of my head

Ran out of time to work out how Twisted shuts down, but we'd want it to stop the compressor, not let it keep running after Synapse stops (to prevent accidentally having two compressor runs happen simultaneously)

erikjohnston:
We need to figure out exactly what to do here, as things stands this makes our CPU metrics useless. One thing we could do is run it in a child process?

anoadragon453:
> We need to figure out exactly what to do here, as things stands this makes our CPU metrics useless. One thing we could do is run it in a child process?

That sounds sensible. Is there any reason not to?

erikjohnston:
> > We need to figure out exactly what to do here, as things stands this makes our CPU metrics useless. One thing we could do is run it in a child process?
> 
> That sounds sensible. Is there any reason not to?

I don't think so, just time.

reivilibre:
> We need to figure out exactly what to do here, as things stands this makes our CPU metrics useless. One thing we could do is run it in a child process?

As a clarification for future context: the issue is that it would frequently spike over 100% cpu, so you could no longer tell how close the python code was to saturating the cpu. In other words, we'd like it to be separate from the normal CPU metrics.

PR #10859, opened 2021-09-20
Allow Synapse Admin API's Room Search to accept non-ASCII characters

### LINKED ISSUE #10841: In Synapse Admin API's Rooms query seach_term don't accept non-ASCII characters
### Description

Not all rooms use English characters, but In Synapse Admin API's Rooms query seach_term don't accept non-ASCII characters.

### Steps to reproduce

For example query to search rooms with Russian symbol "ж" using path `/_synapse/admin/v1/rooms?dir=f&from=0&limit=10&order_by=name&search_term=%D0%B6` shows the error:
```json
{
  "errcode":"M_UNKNOWN",
  "error":"Query parameter 'search_term' must be ascii"
}
```
### Version information
- **Synapse Version**:
    "server_version": "1.42.0",
    "python_version": "3.8.10"

H-Shay:
As per the suggestion here https://github.com/matrix-org/synapse/issues/10841#issuecomment-921649102 made a change to `parse_string` which changes the default encoding from ascii to utf-8, and added a test to check that the admin api room search accepts non-ascii characters and returns the correct room. Fixes #10841.
### Pull Request Checklist

* [ x] Pull request is based on the develop branch
* [ x] Pull request includes a [changelog file](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md#changelog). The entry should:
* [ x] Code style is correct (run the [linters](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md#code-style))


synapse/http/servlet.py
```diff
@@ -331,7 +331,7 @@ def parse_string(
     default: Optional[str] = None,
     required: bool = False,
     allowed_values: Optional[Iterable[str]] = None,
-    encoding: str = "ascii",
+    encoding: str = "UTF-8",
```
clokep, line 334: I don't think we want to do this for all callers of `parse_string`. I think we want to change the call in the admin API to include `encoding="utf-8"`.

changelog.d/10859.misc
```diff
@@ -0,0 +1,2 @@
+Made a change that Synapse Admin API's Rooms Search so that it can now accept non-ASCII characters, allowing admins
+to find rooms with non-ASCII characters.
```
clokep, line 2: This is a bug fix, not a misc change! 👍 

```suggestion
Fix a bug in Unicode support of the room search admin API. It is now possible to search for rooms with non-ASCII characters.
```

tests/rest/admin/test_room.py
```diff
@@ -941,6 +941,52 @@ def _search_test(
         _search_test(None, "bar")
         _search_test(None, "", expected_http_code=400)
 
+    def test_search_term_non_ASCII(self):
```
clokep, line 944: We usually use all lower-case for method names.

```suggestion
    def test_search_term_non_ascii(self):
```

tests/rest/admin/test_room.py
```diff
@@ -941,6 +941,52 @@ def _search_test(
...
+
+        # Set the name for the room
+        self.helper.send_state(
+            room_id,
+            "m.room.name",
+            {"name": room_name},
+            tok=self.admin_user_tok,
+        )
+
+        def _search_test_utf8(
```
clokep, line 960: Since we're only calling this once, I don't think there's much value in it being an inner-function. It hink we can just inline the code.

clokep (commented):
Looks to be on the right path!

synapse/http/servlet.py
clokep: I suspect there's other calls to this that *really* do actually want only ASCII strings. 😄

H-Shay: Yes that makes way more sense than what I did.

tests/rest/admin/test_room.py
```diff
@@ -941,6 +941,33 @@ def _search_test(
...
+        search_term = urllib.parse.quote("ж", "utf-8")
+        url = "/_synapse/admin/v1/rooms?search_term=%s" % (search_term,)
+        channel = self.make_request(
+            "GET",
+            url.encode("ascii"),
+            access_token=self.admin_user_tok,
+        )
+        self.assertEqual(200, channel.code, msg=channel.json_body)
+        self.assertIn(room_id, channel.json_body.get("rooms")[0].get("room_id"))
+        self.assertIn("ж", channel.json_body.get("rooms")[0].get("name"))
```
clokep, line 969: I think both of these we can search for exact equality.

```suggestion
        self.assertEqual(room_id, channel.json_body.get("rooms")[0].get("room_id"))
        self.assertEqual("ж", channel.json_body.get("rooms")[0].get("name"))
```

clokep (approved):
Looks good, minus the equality checks!

DMRobertson:
I think it's worth leaving a note about this in the [admin api documentation](https://matrix-org.github.io/synapse/latest/admin_api/rooms.html). Where we define `search_term`, how about adding something like this?

> The `search_term` may be any Unicode string encoded in UTF-8. To include this in the query URL, any non-ASCII bytes in the UTF-8 bytestream must be escaped using [percent-encoding](https://en.wikipedia.org/wiki/Percent-encoding).

(Possibly overkill?)

clokep:
> I think it's worth leaving a note about this in the [admin api documentation](https://matrix-org.github.io/synapse/latest/admin_api/rooms.html). Where we define `search_term`, how about adding something like this?
> 
> > The `search_term` may be any Unicode string encoded in UTF-8. To include this in the query URL, any non-ASCII bytes in the UTF-8 bytestream must be escaped using [percent-encoding](https://en.wikipedia.org/wiki/Percent-encoding).
> 
> (Possibly overkill?)

I wonder if there's a reference in the spec we could use instead?

H-Shay:
merged

PR #10865, opened 2021-09-21
Add documentation for experimental features.

clokep:
Per our discussions of when to use experimental config flags.

### DIFF
--- a/docs/development/experimental_features.md
+++ b/docs/development/experimental_features.md
@@ -9,8 +9,8 @@ applicable for features which:
   (e.g. a feature is incomplete, buggy, performs poorly, or needs further testing).
 
 Note that this only really applies to features which are expected to be desirable
-to a broad audience. The [module infrastructure](../modules/index.md) should be
-investigated for non-standard features.
+to a broad audience. The [module infrastructure](../modules/index.md) should
+instead be investigated for non-standard features.
 
 Guarding experimental features behind configuration flags should help with some
 of the following scenarios:
--- a/docs/development/experimental_features.md
+++ b/docs/development/experimental_features.md
@@ -4,11 +4,11 @@ It can be desirable to implement "experimental" features which are disabled by
 default and must be explicitly enabled via the Synapse configuration. This is
 applicable for features which:
 
-* Are unstable in the Matrix spec (e.g. those defined by an MSC).
-* Developers are not confident are ready for use by general Synapse administrators/users
+* Are unstable in the Matrix spec (e.g. those defined by an MSC that has not yet been merged).
+* Developers are not confident in their use by general Synapse administrators/users
   (e.g. a feature is incomplete, buggy, performs poorly, or needs further testing).
 
-Note that this does *not* include features which are not expected to be desirable
+Note that this only really applies to features which are expected to be desirable
 to a broad audience. The [module infrastructure](../modules/index.md) should be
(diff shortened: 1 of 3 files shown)


docs/development/experimental_features.md
```diff
@@ -0,0 +1,37 @@
+# Implementing experimental features in Synapse
+
+It can be desirable to implement "experimental" features which are disabled by
+default and must be explicitly enabled via the Synapse configuration. This is
+applicable for features which:
+
+* Are unstable in the Matrix spec (e.g. those defined by an MSC).
```
anoadragon453, line 7:
```suggestion
* Are unstable in the Matrix spec (e.g. those defined by an MSC that has not yet been merged).
```

docs/development/experimental_features.md
```diff
@@ -0,0 +1,37 @@
...
+* Developers are not confident are ready for use by general Synapse administrators/users
```
anoadragon453, line 8:
```suggestion
* Developers are not confident in their use by general Synapse administrators/users
```

Possibly?

docs/development/experimental_features.md
```diff
@@ -0,0 +1,37 @@
...
+  (e.g. a feature is incomplete, buggy, performs poorly, or needs further testing).
+
+Note that this does *not* include features which are not expected to be desirable
```
anoadragon453, line 11:
```suggestion
Note that this only really applies to features which are expected to be desirable
```

An attempt at eliminating double negatives.

docs/development/experimental_features.md
```diff
@@ -0,0 +1,37 @@
...
+to a broad audience. The [module infrastructure](../modules/index.md) should be
+investigated for non-standard features.
```
anoadragon453, line 13: Is the suggestion here that some module features should be hidden behind experimental config flags, or?

docs/development/experimental_features.md
```diff
@@ -0,0 +1,37 @@
...
+* Ease testing a feature (or removal of feature) due to enabling/disabling without
+  code changes. It also becomes possible to ask for wider testing, if desired.
+
+Experimental configuration flags should be disabled by default (requiring Synapse
+administrators to explicitly opt-in), although there are situations where it makes
+sense (from a product point-of-view) to enable features by default. This is
+expected and not an issue.
+
+It is not a requirement that experimental features require a configuration flag,
+but if one should be included if unsure.
```
anoadragon453, line 33:
```suggestion
It is not a requirement for experimental features to be behind a configuration flag,
but one should be used if unsure.
```

anoadragon453 (approved):
Thank you so much for documenting this! I just had a few wording nits below, but otherwise this looks like it covers all the usual questions (other than documenting them separately).

docs/development/experimental_features.md
clokep: The suggestion is that some features should be built as modules, not as things behind experimental config flags.

clokep: Maybe it is clearer to just remove this? Not sure, was trying to give people other options though.

anoadragon453: Aha OK! In that case, I might just slightly reword this to:

```suggestion
to a broad audience. The [module infrastructure](../modules/index.md) should
instead be investigated for non-standard features.
```

PR #10867, opened 2021-09-21
Add types to http.site

erikjohnston:
Split out from #10844


clokep (approved):
Seems reasonable!

synapse/http/site.py
```diff
@@ -547,12 +549,14 @@ def __init__(
 
         def request_factory(channel, queued) -> Request:
```
erikjohnston, line 550:
```suggestion
        def request_factory(channel, queued: bool) -> Request:
```

PR #10868, opened 2021-09-21
Add reactor to `SynapseRequest` and fix up types.

erikjohnston:
It's useful to be able to get the reactor from a request, c.f. #10844

Split out from #10844

### DIFF
--- a/synapse/http/server.py
+++ b/synapse/http/server.py
@@ -320,7 +320,7 @@ class DirectServeJsonResource(_AsyncResource):
 
     def _send_response(
         self,
-        request: Request,
+        request: SynapseRequest,
         code: int,
         response_object: Any,
     ):
@@ -629,7 +629,7 @@ def _encode_json_bytes(json_object: Any) -> Iterator[bytes]:
 
 
 def respond_with_json(
-    request: Request,
+    request: SynapseRequest,
     code: int,
     json_object: Any,
     send_cors: bool = False,
--- a/synapse/rest/key/v2/remote_key_resource.py
+++ b/synapse/rest/key/v2/remote_key_resource.py
@@ -17,12 +17,11 @@ from typing import TYPE_CHECKING, Dict
 
 from signedjson.sign import sign_json
 
-from twisted.web.server import Request
-
 from synapse.api.errors import Codes, SynapseError
 from synapse.crypto.keyring import ServerKeyFetcher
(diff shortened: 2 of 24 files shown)


synapse/http/server.py
```diff
@@ -629,7 +629,7 @@ def _encode_json_bytes(json_object: Any) -> Iterator[bytes]:
 
 
 def respond_with_json(
-    request: Request,
+    request: SynapseRequest,
```
clokep, line 632: This doesn't look like it needs a `SynapseRequest`? I'm guessing that's coming in a different PR?

erikjohnston: Ah, yes it is.

synapse/rest/key/v2/remote_key_resource.py
```diff
@@ -126,7 +125,7 @@ async def _async_render_POST(self, request: Request) -> None:
 
     async def query_keys(
         self,
-        request: Request,
+        request: SynapseRequest,
```
squahtx, line 128: I can't see where we use this `SynapseRequest`?

erikjohnston: It gets fed into `respond_with_json` which now requires a `SynapseRequest`

richvdh (approved):
tying all this stuff to `SynapseRequest` rather than `Request` makes me very sad, because `SynapseRequest` is conflating a bunch of different things. It's not really clear what aspect of `SynapseRequest` these things care about (ie, what the data dependencies are).

Part of me wonders if we could extract some sort of `Protocol` definition that we could use for these things instead (`RequestWithReactor` or something), but I suspect that's going to end up being fiddly and ugly, and we've probably spent enough time messing about here.

ut I don't have any other great ideas.

erikjohnston:
> tying all this stuff to `SynapseRequest` rather than `Request` makes me very sad, because `SynapseRequest` is conflating a bunch of different things. It's not really clear what aspect of `SynapseRequest` these things care about (ie, what the data dependencies are).


Yeah, I'm not super happy about it, but I've vaguely convinced myself that being able to get at the reactor being used by the request is a valid thing :shrug:


> Part of me wonders if we could extract some sort of `Protocol` definition that we could use for these things instead (`RequestWithReactor` or something), but I suspect that's going to end up being fiddly and ugly, and we've probably spent enough time messing about here.

That could work I guess, though I have a vague suspicion that in our case it'll probably lead to more noise than help.

PR #10869, opened 2021-09-21
Clear our destination directories before copying files to GitHub pages.

clokep:
This properly removes stale files, e.g. https://matrix-org.github.io/synapse/latest/modules.html (which should 404 after this).


clokep:
According to @anoadragon453:

> Indeed! Looks like only the files in `destination_dir` will be purged (rather than the whole branch), so yep, just need to remove that line.

changelog.d/10869.docs
```diff
@@ -0,0 +1 @@
+Properly remove stale files from GitHub pages.
```
anoadragon453, line 1:
```suggestion
Properly remove stale files from GitHub pages when generating the documentation.
```

Just for some more clarity around GitHub pages is used for.

anoadragon453 (approved):

PR #10870, opened 2021-09-21
Track and deduplicate in-flight requests to `_get_state_for_groups`.

reivilibre:
This is the 2nd PR of a series (expected to have 4 PRs) to address #10301 .

Depends on #10825.

The general idea is that when requesting some state, we make use of existing queries that are already in-flight.
Those queries might only give us partial information, so we may have to use multiple, and we may even have to spawn our own (smaller) query for the leftover state.

Coming up later:
- ordering the in-flight request cache so that 'wider' StateFilters get gathered first
- capping the number of in-flight requests for a group, before we just grab all the state at once and satisfy everyone.


reivilibre:
I'm going to add some more tests when #10825 is available so I can actually prove that things get deduplicated, but would like some thoughts on if this seems sensible.

erikjohnston (dismissed):

tests/storage/databases/test_state_store.py
```diff
@@ -0,0 +1,109 @@
+from typing import Dict, List, Sequence, Tuple
```
clokep, line 1: License header please!

tests/storage/databases/test_state_store.py
```diff
@@ -0,0 +1,109 @@
...
+
+from synapse.storage.state import StateFilter
+from synapse.types import MutableStateMap, StateMap
+
+from tests.unittest import HomeserverTestCase
+
+
+class StateGroupInflightCachingTestCase(HomeserverTestCase):
+    def setUp(self) -> None:
+        super(StateGroupInflightCachingTestCase, self).setUp()
```
clokep, line 14:
```suggestion
        super().setUp()
```

tests/storage/databases/test_state_store.py
```diff
@@ -0,0 +1,109 @@
...
+        # Patch out the `_get_state_groups_from_groups`.
+        # This is useful because it lets us pretend we have a slow database.
+        gsgfg_patch = patch(
```
clokep, line 17: `gsgfg` looks like a cat walked on your keyboard! Maybe we can use `get_state_groups` instead?

tests/storage/databases/test_state_store.py
```diff
@@ -0,0 +1,109 @@
...
+        gsgfg_patch.start()
+        self.addCleanup(gsgfg_patch.stop)
+        self.gsgfg_calls: List[
+            Tuple[Tuple[int, ...], StateFilter, Deferred[Dict[int, StateMap[str]]]]
+        ] = []
+
+    def prepare(self, reactor, clock, homeserver) -> None:
+        super(StateGroupInflightCachingTestCase, self).prepare(
+            reactor, clock, homeserver
+        )
```
clokep, line 30: No reason to call super here (we don't usually, this method is meant to be overridden).

```suggestion
```

tests/storage/databases/test_state_store.py
```diff
@@ -0,0 +1,109 @@
...
+        self.state_storage = homeserver.get_storage().state
+        self.state_datastore = homeserver.get_datastores().state
+
+    def _fake_get_state_groups_from_groups(
+        self, groups: Sequence[int], state_filter: StateFilter
+    ) -> "Deferred[Dict[int, StateMap[str]]]":
+        print("hi", groups, state_filter)
```
clokep, line 37: Please remove debug code!

tests/storage/databases/test_state_store.py
```diff
@@ -0,0 +1,109 @@
...
+
+        for group in groups:
+            group_result: MutableStateMap[str] = {}
+            result[group] = group_result
+
+            for state_type, state_keys in state_filter.types.items():
+                if state_keys is None:
+                    group_result[
+                        (state_type, "wild wombat")
+                    ] = f"{group} {state_type} wild wombat"
```
clokep, line 62: Are the keys/values here important?

tests/storage/databases/test_state_store.py
```diff
@@ -0,0 +1,109 @@
...
+        ] = []
+
+    def prepare(self, reactor, clock, homeserver) -> None:
+        super(StateGroupInflightCachingTestCase, self).prepare(
+            reactor, clock, homeserver
+        )
+        self.state_storage = homeserver.get_storage().state
+        self.state_datastore = homeserver.get_datastores().state
+
+    def _fake_get_state_groups_from_groups(
```
clokep, line 34: This looks like it could be replaced by a `side_effect=Deferred` when mocking the call and then using the standard methods on `Mock` to get back the arguments / results of calls...but maybe not.

synapse/storage/databases/state/store.py
```diff
@@ -228,6 +247,144 @@ def _get_state_for_group_using_cache(
...
+        """
+        Attempts to gather in-flight requests and re-use them to retrieve state
+        for the given state group, filtered with the given state filter.
+
+        Returns:
+            tuple (
+                sequence of ObservableDeferreds to observe,
+                StateFilter representing what is left over (what else needs
+                to be requested to fulfil the request)
+            )
```
clokep, line 262: I tried to somehow make this so it looks like two items, not three:

```suggestion
            Tuple of two values:
                A sequence of ObservableDeferreds to observe
                A StateFilter representing what else needs to be requested to fulfill the request
```

synapse/storage/databases/state/store.py
```diff
@@ -228,6 +247,144 @@ def _get_state_for_group_using_cache(
...
+        (
+            reusable_requests,
+            state_filter_left_over,
+        ) = self._get_state_for_group_gather_inflight_requests(group, state_filter)
+
+        if state_filter_left_over != StateFilter.none():
+            # Fetch remaining state
+            remaining = await self._get_state_for_group_fire_request(
+                group, state_filter_left_over
+            )
```
clokep, line 370: I assume that `reusable_requests` won't become stale in the time that we're awaiting?

synapse/storage/databases/state/store.py
```diff
@@ -228,6 +247,144 @@ def _get_state_for_group_using_cache(
...
+            assembled_state: MutableStateMap[str] = dict(remaining)
+        else:
+            assembled_state = {}
+
+        gathered = await make_deferred_yieldable(
+            defer.gatherResults(
+                (r.observe() for r in reusable_requests), consumeErrors=True
+            )
+        ).addErrback(unwrapFirstError)
```
clokep, line 379: Does it make sense to both consume errors and unwrap the first one?

clokep (commented):
I took a go at reviewing this, I'm not sure I fully understand it yet, but found some bits to comment on.

tests/storage/databases/test_state_store.py
reivilibre: hah! Yes that sounds like it'd be much better.

tests/storage/databases/test_state_store.py
reivilibre: It's been so long, but I think I was intending to write some more tests that check the behaviour of how subsets of state fit in, if the idea so far looked OK.

As it stands this could be really simplified, but instead the correct answer is probably for those tests to get written, if this path looks acceptable.

synapse/storage/databases/state/store.py
reivilibre: Your question sounds like you think it's not. I'm struggling to understand the docstrings, but it looks like:

* `gatherResults` constructs a `DeferredList` with `fireOnOneErrback = True`
* ```
            @param fireOnOneErrback: (keyword param) a flag indicating that this
            L{DeferredList} will fire when the first L{Deferred} in
            C{deferredList} fires with a failure result without waiting for any
            of the other Deferreds.  When this flag is set, if a Deferred in the
            list errbacks, the DeferredList will errback with a L{FirstError}
            failure wrapping the failure of that Deferred.
    ```
     * i.e. I think this will give you a `FirstError(Exception)` rather than `Exception` if anything happens
* `unwrapFirstError` looks like this:
     ```python
     def unwrapFirstError(failure: Failure) -> Failure:
         # defer.gatherResults and DeferredLists wrap failures.
         failure.trap(defer.FirstError)
         return failure.value.subFailure  # type: ignore[union-attr]  # Issue in Twisted's annotations
     ```
    * essentially `.mapError(lambda err: err.value.subFailure)` to my eye? `self.trap(x)` seems to be more or less `if not isinstance(self, x): raise self` (it wasn't obvious to me)

I don't seem to be the first using it this way: https://github.com/matrix-org/synapse/blob/7468723697e4d292315ce807b5000062a02b37be/synapse/util/caches/descriptors.py#L474-L476 .

Does this sound right now or is there something missing?

synapse/storage/databases/state/store.py
reivilibre: if you `observe` a request that finished, it just gives the result back straight away, so this is fine.

tests/storage/databases/test_state_store.py
reivilibre: I can't see how to get the results off a `Mock` — if you know how that would be appreciated!

synapse/storage/databases/state/store.py
```diff
@@ -106,6 +119,12 @@ def __init__(
             500000,
         )
 
+        # Current ongoing get_state_for_groups in-flight requests
+        # {group ID -> {StateFilter -> ObservableDeferred}}
+        self._state_group_inflight_requests: Dict[
+            int, Dict[StateFilter, ObservableDeferred[StateMap[str]]]
```
richvdh, line 125: Random driveby comment: Consider typing this with `AbstractObservableDeferred` instead:

```suggestion
        self._state_group_inflight_requests: Dict[
            int, Dict[StateFilter, AbstractObservableDeferred[StateMap[str]]]
```

Once you've created the `ObservableDeferred`, the only thing you should be doing with it is `.observe()`ing it, which you can enforce via `AbstractObservableDeferred`.

(somewhat related to my campaign on https://github.com/matrix-org/synapse/issues/11390)

tests/storage/databases/test_state_store.py
clokep: [`call_args` (or `call_args_list`)](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.call_args) should have it.

reivilibre: I've tried my best to read that carefully multiple times and I don't see it make any mention of the return values, just the `*args` and `**kwargs`. Am I just being blind?

clokep: Oh, yes I don't think there's a way to get the return value of the calls. Sorry, misread!

synapse/storage/databases/state/store.py
clokep: I meant the list of `reusable_requests`, I think, but I'm not entirely sure I remember the context anymore.

synapse/storage/databases/state/store.py
clokep: Yeah, that sounds reasonable. Thanks!

clokep:
Discussed yesterday with @reivilibre and he is going to take a look at @richvdh's review before I take another look. Removing my review for now.

tests/storage/databases/test_state_store.py
```diff
@@ -0,0 +1,117 @@
+# Copyright 2021 The Matrix.org Foundation C.I.C.
```
clokep, line 1: 😄 
```suggestion
# Copyright 2022 The Matrix.org Foundation C.I.C.
```

tests/storage/databases/test_state_store.py
```diff
@@ -0,0 +1,117 @@
...
+from synapse.types import MutableStateMap, StateMap
+
+from tests.unittest import HomeserverTestCase
+
+
+class StateGroupInflightCachingTestCase(HomeserverTestCase):
+    def setUp(self) -> None:
+        super().setUp()
+        # Patch out the `_get_state_groups_from_groups`.
+        # This is useful because it lets us pretend we have a slow database.
```
clokep, line 30: Note a huge deal, but I think we usually patch it out on the instance in `prepare` instead?

Something like `patch.object(hs.get_datastore()._get_state_groups_from_groups)` or something like that.

The nice bit of this is that there's no risk of it impacting other tests.

tests/storage/databases/test_state_store.py
```diff
@@ -0,0 +1,117 @@
...
+                else:
+                    for state_key in state_keys:
+                        group_result[(state_type, state_key)] = "abc"
+
+            if state_filter.include_others:
+                group_result[("other.event.type", "state.key")] = "123"
+
+        d.callback(result)
+
+    def test_duplicate_requests_deduplicated(self) -> None:
```
clokep, line 81: This test seems nuanced enough that it could use a docstring saying what the overall steps are that are happening.

synapse/storage/databases/state/store.py
```diff
@@ -269,30 +428,14 @@ async def _get_state_for_groups(
...
-            cache_seq_num_non_members=cache_sequence_nm,
-        )
-
-        # And finally update the result dict, by filtering out any extra
-        # stuff we pulled out of the database.
-        for group, group_state_dict in group_to_state_dict.items():
-            # We just replace any existing entries, as we will have loaded
-            # everything we need from the database anyway.
-            state[group] = state_filter.filter_state(group_state_dict)
+        for group, group_result in zip(incomplete_groups, results_from_requests):
```
clokep, line 437: `incomplete_groups` seems to be a `set`, which the docs say "support partial ordering". I suspect iterating the same `set` multiple times would give the same order, but I don't know if we should trust that?

synapse/storage/databases/state/store.py
```diff
@@ -228,6 +251,142 @@ def _get_state_for_group_using_cache(
...
+                A StateFilter representing what else needs to be requested to fulfill the request
+        """
+
+        inflight_requests = self._state_group_inflight_requests.get(group)
+        if inflight_requests is None:
+            # no requests for this group, need to retrieve it all ourselves
+            return (), state_filter
+
+        state_filter_left_over = state_filter
+        reusable_requests = []
```
clokep, line 273:
```suggestion
        # The list of ongoing requests which will help narrow the current request.
        reusable_requests = []
```

synapse/storage/databases/state/store.py
```diff
@@ -228,6 +251,142 @@ def _get_state_for_group_using_cache(
...
+        for (
+            request_state_filter,
+            request_deferred,
+        ) in inflight_requests.items():
```
clokep, line 277: I suspect this can fit on a single line?

```suggestion
        for (request_state_filter, request_deferred) in inflight_requests.items():
```

synapse/storage/databases/state/store.py
```diff
@@ -228,6 +251,142 @@ def _get_state_for_group_using_cache(
...
+            new_state_filter_left_over = state_filter_left_over.approx_difference(
+                request_state_filter
+            )
+            if new_state_filter_left_over != state_filter_left_over:
+                # reusing this request narrows our StateFilter down a bit.
+                reusable_requests.append(request_deferred)
+            state_filter_left_over = new_state_filter_left_over
```
clokep, line 284: I'm finding some of this a bit confusing as there are three variables that are similar (`state_filter`, `new_state_filter_left_over`, and `state_filter_left_over`). I think one of those is unneeded and we could just rename the input param to `state_filter_left_over`.

I also don't this line and the below can only occur if `new_state_filter_left_over != state_filter_left_over`? But I'm not 100% sure. It might be clearer to check if they're equal and then continue?

synapse/storage/databases/state/store.py
```diff
@@ -228,6 +251,142 @@ def _get_state_for_group_using_cache(
...
+        request_deferred = run_in_background(_the_request)
+        observable_deferred = ObservableDeferred(request_deferred, consumeErrors=True)
+
+        # Insert the ObservableDeferred into the cache
+        group_request_dict = self._state_group_inflight_requests.setdefault(group, {})
+        group_request_dict[db_state_filter] = observable_deferred
+
+        return await make_deferred_yieldable(observable_deferred.observe())
+
+    async def _get_state_for_group_using_inflight_cache(
```
clokep, line 347: I'm struggling a bit with the names of these methods, you have:

1. `_get_state_for_group_using_inflight_cache` which calls out to:
	1. `_get_state_for_group_gather_inflight_requests`, and
	2. `_get_state_for_group_fire_request`

So it seems we have an "inflight" cache with 3 operations: `get`, `gather`, and `fire`. I can't really seem to come up with better names though. It could just help if the docstring for `_get_state_for_group_using_inflight_cache` included a bit more of what is being used:

> 1. Calls `_get_state_for_group_gather_inflight_requests` to fetch any ongoing requests
>    which might overlap with the current request.
> 2. Creates a new request for any state which cannot be captured from ongoing requests.

synapse/storage/databases/state/store.py
reivilibre: Well spotted. Practically there would be very little reason I can see why a set should change iteration order between iterations, but it's not great practice since a mutation can resize the underlying hashtable, causing a rehash which can shuffle things around (at least in most languages. I know Python has guarantees about the order of dict keys, but don't see the same about sets.

> `set`, which the docs say "support partial ordering".

Not that I can see the passage in the documentation that you're referring to, but I suspect this is a red herring and is actually talking about a partial ordering on sets themselves. It'd be called a 'partial ordering' because if you have two sets A and B, you might find that neither `A < B` nor `B < A` holds.

The partial ordering, `<`, seems to mean 'is a subset of' here.
```python
>>> { 4 } < { 4, 8 }
True
>>> { 4, 7 } < { 4, 8 }
False
>>> { 4, 7 } > { 4, 8 }
False
```

Not super relevant but just a bit of trivia I suppose :)

synapse/storage/databases/state/store.py
reivilibre: I suspect the identifiers were once longer, then black split the lines, but apparently doesn't bother to recombine later on after the identifiers are changed to something shorter. Thanks :)

tests/storage/databases/test_state_store.py
reivilibre: glad you mentioned it!

tests/storage/databases/test_state_store.py
reivilibre: Added one; any good?

synapse/storage/databases/state/store.py
reivilibre: I've switched it around following your suggestions

synapse/storage/databases/state/store.py
reivilibre: good suggestion — have added words to that effect, let me know if any more would be useful!

synapse/storage/databases/state/store.py
clokep: I think you're quite right, but I couldn't find any guarantee of ordering of sets. Practically though I couldn't make it go unordered...

erikjohnston: I'd prefer it if we didn't rely on that, especially as the consequences of getting it wrong is going to be horrific. Can we instead do something like:

```python
async def get_from_cache(group, filter):
    state[group] = await self._get_state_for_group_using_inflight_cache(group, filter)

await yieldable_gather_results(
    get_from_cache,
    incomplete_groups,
    state_filter,
)
```

reivilibre: > I'd prefer it if we didn't rely on that, especially as the consequences of getting it wrong is going to be horrific. Can we instead do something like:

This was already changed before your comment, but I think I like your idea more...

erikjohnston: OH woops, clearly I can't read code :)

PR #10871, opened 2021-09-21
Increase parallelism of `trial` test runs to 4 in CI.

reivilibre:


reivilibre:
Yeah, I wasn't sure this would do much, but it was suggested to me and I thought it was interesting enough to give it a try.

This would help if our test runs don't saturate the 2 vCPUs that are given.

What I do notice, though, is that this does speed up the Postgres runs by a few minutes (though I admit it's hard to have too much faith in it from just one sample, but I just looked at quite a few test runs from other PRs and this one was faster than all of them by between 2 and 9 minutes!).

I guess that there's quite a bit of back and forth between Synapse and Postgres which is not CPU-bound (I guess Postgres is actually writing to and reading from disk; our SQLite tests run in-memory if I remember correctly)?

clokep:
Speeding up even those does sound like a win!

PR #10873, opened 2021-09-21
Ensure we mark sent knocks as outliers

### LINKED ISSUE #10872: knock events sent over federation have no state but are not marked as outliers
when we send out a knock event, we persist the sent event in the events table, but do not mark it out as an outlier.

We give it a state_group, but that state_group contains only the knock event itself.

richvdh:
fixes #10872

### DIFF
--- a/changelog.d/10873.bugfix
+++ b/changelog.d/10873.bugfix
@@ -1,0 +1,1 @@
+Fix a bug introduced in Synapse 1.37.0 which caused `knock` events which we sent to remote servers to be incorrectly stored in the local database.
--- a/synapse/handlers/federation.py
+++ b/synapse/handlers/federation.py
@@ -593,6 +593,10 @@ class FederationHandler(BaseHandler):
             target_hosts, room_id, knockee, Membership.KNOCK, content, params=params
         )
 
+        # Mark the knock as an outlier as we don't yet have the state at this point in
+        # the DAG.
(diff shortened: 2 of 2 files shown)


anoadragon453:
Looks like the following Complement tests are failing:

* `TestKnocking/Knocking_on_a_room_with_join_rule_'knock'_should_succeed`
* `TestKnocking/A_user_that_has_already_knocked_is_allowed_to_knock_again_on_the_same_room#01`
* `TestKnocking/A_user_in_the_room_can_reject_a_knock#01`
* `TestKnocking/A_user_can_knock_on_a_room_without_a_reason#01`

Here's a link to their implementation: https://github.com/matrix-org/complement/blob/97183bf2e1572e06ada237977d405a1e9f4ef716/tests/msc2403_test.go#L129

Note that the `#01` at the end of each test name above indicates that this is the second time this test has run. The knocking Complement tests work via first running the tests between two local users, and then one local and one remote user. Thus these failing tests are only occurring in the local and remote user case.

Given the timing of each tests (~5s) before they fail, my initial guess would be that they're all timing out.

richvdh:
"bother". Thanks.

anoadragon453 (approved):
Thanks for catching this. Presumably this bit was taken from the equivalent code in `on_invite_request`, but that method sets the `outlier` and `out_of_band` bits correctly so :shrug:

Issue #10737, opened 2021-09-01
MSC2716 leftover todo tasks


MadLittleMods:
Just trying to keep track of the last remaining little tasks to get [MSC2716](https://github.com/matrix-org/matrix-doc/pull/2716) good to go.

## Todo

 - [x] Verify `chunk_id_from_query` corresponds to an actual insertion event, [`synapse/rest/client/room_batch.py#L302)`](https://github.com/matrix-org/synapse/blob/d1f1b46c2cfe612790aebdd54765953951c94e45/synapse/rest/client/room_batch.py#L302)
    - -> https://github.com/matrix-org/synapse/pull/10776 
 - [x] Ensure insertion events have unique `next_chunk_id` before persisting. Ensure `(room_id, next_chunk_id)` is unique. Avoid cross-talk/conflicts between chunks
    - -> https://github.com/matrix-org/synapse/pull/10877
    - Reject/soft-fail event over federation
    - Disallow persisting on local server
    - Discussed at https://github.com/matrix-org/synapse/pull/10245#discussion_r669641879
 - [x] Split meta events in `/batch_send` response into their own fields, https://github.com/matrix-org/matrix-doc/pull/2716#discussion_r685508994
    - -> https://github.com/matrix-org/synapse/pull/10777
 - [x] Rename `chunk` to `batch` to match `/batch_send` endpoint, https://github.com/matrix-org/matrix-doc/pull/2716#discussion_r684574497
    - -> https://github.com/matrix-org/synapse/pull/10838
 - [x] Rename `?prev_event` to `?prev_event_id`, https://github.com/matrix-org/matrix-doc/pull/2716#discussion_r705872887
    - -> https://github.com/matrix-org/synapse/pull/10839
 - [ ] ~~https://github.com/matrix-org/synapse/issues/10764~~
     - No longer needed because we now use an empty set of `prev_events` instead of fake `prev_events` 
 - [x] https://github.com/matrix-org/synapse/issues/11091
 - [x] https://github.com/matrix-org/synapse/issues/11241
 - [x] Make "marker" events state events, https://github.com/matrix-org/matrix-doc/pull/2716#discussion_r782629097. When the homeserver sees the marker state change or joins the room again, it can go up the marker event `unsigned.replaces_state` chain of the to collect all of the `m.room.marker` events along the way and set as backward extremities on the homeserver.
    - https://github.com/matrix-org/synapse/pull/12718
    - https://github.com/matrix-org/complement/pull/371
[...]

PR #10877, opened 2021-09-22
Ensure `(room_id, next_batch_id)` is unique to avoid cross-talk/conflicts between batches (MSC2716)

MadLittleMods:
Ensure `(room_id, next_batch_id)` is unique across insertion events to avoid cross-talk/conflicts between batches. Currently does not stop duplicate insertion events over federation, see https://github.com/matrix-org/synapse/pull/10877#discussion_r715315789

Part of [MSC2716](https://github.com/matrix-org/matrix-doc/pull/2716)

Part of https://github.com/matrix-org/synapse/issues/10737

Complement changes in https://github.com/matrix-org/complement/pull/207

Briefly discussed at https://github.com/matrix-org/synapse/pull/10245#discussion_r669641879

### Dev notes

`soft_failed` vs `rejected`?

 - `rejected`: when an event doesn't pass event auth rules based on the auth events given and state at the event
    - https://matrix.org/docs/spec/server_server/r0.1.4#rejection
 - `soft_fail`: when an event doesn't pass the event auth rules based on current state of the room
    - " If the event does not pass the auth checks based on the current state of the room (but does pass the auth checks based on the state at that event) it should be "soft failed".", https://matrix.org/docs/spec/server_server/r0.1.4#soft-failure
    - `_check_for_soft_fail`

---

 - `_persist_events_txn`

---

 - `auth_types_for_event`

---

`synapse/storage/databases/main/events.py` -> `_handle_insertion_event`

```python
conflicting_insertion_event_id = self.db_pool.simple_select_one_onecol_txn(
    txn,
    table="insertion_events",
    keyvalues={
        "room_id": event.room_id,
        "next_batch_id": next_batch_id,
    },
    retcol="event_id",
    allow_none=True,
)
if conflicting_insertion_event_id is not None:
    # The current insertion event that we're processing is invalid
    # because an insertion event already exists in the room with the
    # same next_batch_id. We can't allow multiple because the batch
    # pointing will get weird, e.g. we can't determine which insertion
    # event the batch event is pointing to.
    return
```

### Pull Request Checklist

### DIFF
--- a/changelog.d/10877.feature
+++ b/changelog.d/10877.feature
@@ -1,0 +1,1 @@
+Ensure `(room_id, next_batch_id)` is unique across [MSC2716](https://github.com/matrix-org/matrix-doc/pull/2716) insertion events in rooms to avoid cross-talk/conflicts between batches
--- a/changelog.d/10877.feature
+++ b/changelog.d/10877.feature
@@ -1,1 +1,1 @@
-Ensure `(room_id, next_batch_id)` is unique across [MSC2716](https://github.com/matrix-org/matrix-doc/pull/2716) insertion events in rooms to avoid cross-talk/conflicts between batches
+Ensure `(room_id, next_batch_id)` is unique across [MSC2716](https://github.com/matrix-org/matrix-doc/pull/2716) insertion events in rooms to avoid cross-talk/conflicts between batches.
--- a/synapse/handlers/message.py
+++ b/synapse/handlers/message.py
@@ -1390,9 +1390,6 @@ class EventCreationHandler:
                     self.room_prejoin_state_types,
                 )
 
-        room_version = await self.store.get_room_version_id(event.room_id)
-        room_version_obj = KNOWN_ROOM_VERSIONS[room_version]
-
         if event.type == EventTypes.Redaction:
             original_event = await self.store.get_event(
                 event.redacts,
@@ -1401,6 +1398,9 @@ class EventCreationHandler:
                 allow_rejected=False,
                 allow_none=True,
             )
+
+            room_version = await self.store.get_room_version_id(event.room_id)
+            room_version_obj = KNOWN_ROOM_VERSIONS[room_version]
 
             # we can make some additional checks now if we have the original event.
             if original_event:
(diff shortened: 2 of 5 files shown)


synapse/storage/databases/main/events.py
```diff
@@ -1795,6 +1795,24 @@ def _handle_insertion_event(self, txn: LoggingTransaction, event: EventBase):
...
+            retcol="event_id",
+            allow_none=True,
+        )
+        if conflicting_insertion_event_id is not None:
+            # The new  insertion event is invalid because there already exists
+            # and insertion event in the room with the same next_batch_id. We
+            # can't allow multiple because the batch pointing will get weird,
+            # e.g. we can't determine which insertion event the batch event is
+            # pointing to.
+            return
```
MadLittleMods, line 1814: I kinda want to `soft_fail` or mark as `rejected` instead of just ignoring these events.

I can't do any event lookup in `event_auth` to compare to other insertion events.

The closest related spot I can find is `_check_for_soft_fail` 🤔 which still does auth stuff.

synapse/handlers/message.py
```diff
@@ -1461,6 +1462,21 @@ async def persist_and_notify_client_event(
...
+                # The current insertion event that we're processing is invalid
+                # because an insertion event already exists in the room with the
+                # same next_batch_id. We can't allow multiple because the batch
+                # pointing will get weird, e.g. we can't determine which insertion
+                # event the batch event is pointing to.
+                raise SynapseError(
+                    HTTPStatus.BAD_REQUEST,
+                    "Another insertion event already exists with the same next_batch_id",
+                    errcode=Codes.INVALID_PARAM,
+                )
```
MadLittleMods, line 1478: When we receive these events over federation, should we ignore duplicate insertion events? In a previous iteration of this PR, I ignored duplicate conflicting insertion events that came over federation but really wanted to reject/soft_fail instead (felt more right).

Maybe it's better to just let them do what they want and we can just stop this from the `/send` touch point as a warning stop-gap.

synapse/handlers/message.py
```diff
@@ -1461,6 +1462,41 @@ async def persist_and_notify_client_event(
...
+                    # The current insertion event that we're processing is invalid
+                    # because an insertion event already exists in the room with the
+                    # same next_batch_id. We can't allow multiple because the batch
+                    # pointing will get weird, e.g. we can't determine which insertion
+                    # event the batch event is pointing to.
+                    raise SynapseError(
+                        HTTPStatus.BAD_REQUEST,
+                        "Another insertion event already exists with the same next_batch_id",
+                        errcode=Codes.INVALID_PARAM,
+                    )
```
MadLittleMods, line 1498: When we receive events over federation, should we ignore duplicate insertion events? In a previous iteration of this PR, I ignored duplicate conflicting insertion events that came over federation but really wanted to reject/soft_fail instead (felt more right). I can't add a conflict database lookup in `event_auth` because it doesn't have any of that available. `_check_for_soft_fail` seems like the closest spot available for this kind of thing.

Maybe it's better to just let them do what they want and we can just stop this from the `/send` touch point as a warning stop-gap.

synapse/storage/databases/main/room_batch.py
```diff
@@ -18,7 +18,7 @@
 
 
 class RoomBatchStore(SQLBaseStore):
-    async def get_insertion_event_by_batch_id(self, batch_id: str) -> Optional[str]:
+    async def get_insertion_event_by_batch_id(self, room_id: str, batch_id: str) -> Optional[str]:
```
MadLittleMods, line 21: Changed this to match our unique expectation. Only needs to be unique per room.

synapse/storage/databases/main/room_batch.py
```diff
@@ -30,7 +32,7 @@ async def get_insertion_event_by_batch_id(self, batch_id: str) -> Optional[str]:
         """
         return await self.db_pool.simple_select_one_onecol(
             table="insertion_events",
-            keyvalues={"next_batch_id": batch_id},
+            keyvalues={"room_id": room_id, "next_batch_id": batch_id},
```
MadLittleMods, line 35: Changed this to match our unique expectation. Only needs to be unique per room.

erikjohnston (approved):
Do we need to add a unique constraint to the DB tables?

changelog.d/10877.feature
```diff
@@ -0,0 +1 @@
+Ensure `(room_id, next_batch_id)` is unique across [MSC2716](https://github.com/matrix-org/matrix-doc/pull/2716) insertion events in rooms to avoid cross-talk/conflicts between batches.
```
MadLittleMods, line 1: > Do we need to add a unique constraint to the DB tables?

@erikjohnston If possible, but you shied away from it in https://github.com/matrix-org/synapse/pull/10245#discussion_r669641879

> I think we just want to make `event_id` unique? (Note that event IDs are globally unique). 
> 
> I'm not sure whether we can assert that `(room_id, next_chunk_id)` is unique here, I could imagine that one day a remote server gets confused and reuses a chunk ID or something, at which point we probably *don't* want to explode on insertion of the event.
>
> *-- @erikjohnston, https://github.com/matrix-org/synapse/pull/10245#discussion_r669641879*

erikjohnston: Oh lols. I'm not sure I agree with myself there, but haven't really thought about it. Anyway, doesn't need to block this  PR.

MadLittleMods:
Thanks for the review @erikjohnston 🦏

PR #10879, opened 2021-09-22
Include outlier status in `str(event)` for V2/V3 events

richvdh:
I meant to do this before, in #10591, but because I'm stupid I forgot to do it for V2 and V3 events.

I've factored the common code out to `EventBase` to save us having two copies of it.

This means that for `FrozenEvent` we replace `self.get("event_id", None)` with `self.event_id`, which I think is safe. `get()` is an alias for `self._dict.get()`, whereas `event_id()` is an `@property` method which looks up `self._event_id`, which is populated during construction from the same dict. We don't seem to rely on the fallback, because if the `event_id` key is absent from the dict then construction of the `EventBase` object will fail.

Long story short, the only way this could change behaviour is if `event_dict["event_id"]` is changed *after* the `EventBase` object is constructed without updating the `_event_id` field, or vice versa - either of which would be very problematic anyway and the behavior of `str(event)` is the least of our worries.

### DIFF
--- a/changelog.d/10879.misc
+++ b/changelog.d/10879.misc
@@ -1,0 +1,1 @@
+Include outlier status when we log V2 or V3 events.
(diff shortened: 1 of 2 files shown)


reivilibre (approved):
Can't see anything wrong with that! :)

PR #10883, opened 2021-09-22
Factor out a separate `EventContext.for_outlier`

richvdh:
Constructing an `EventContext` for an outlier is actually really simple, and there's no sense in going via an `async` method in the `StateHandler`.

This also means that we can resolve a bunch of FIXMEs.

### DIFF
--- a/changelog.d/10883.misc
+++ b/changelog.d/10883.misc
@@ -1,0 +1,1 @@
+Clean up some of the federation event authentication code for clarity.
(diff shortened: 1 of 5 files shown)


clokep (approved):
Seems reasonable, there seems to be a ton of duplicated code around creating an event, marking it an outlier, giving it a particular context. I'm not sure if we can clean that up at all, (and this PR certainly doesn't make it worse), just noting it.

richvdh:
yes there is, but I think there's a bit of a long game here. I'm really just working on figuring out what half of this stuff even does.

clokep:
> yes there is, but I think there's a bit of a long game here.

Yep! Definitely wasn't try to imply we should fix that here. Thanks for improving this!

PR #10885, opened 2021-09-22
Use direct references for some configuration variables (part 3)

clokep:
Builds on #10798 and #10812.

Each commit was done with `sed` + `black`, then running the linters and fixing some type hints.

After this PR I've checked the variables up to `synapse.config.email`. 😄


anoadragon453 (approved):
:tada:

PR #10887, opened 2021-09-22
Allow `.` and `~` chars in registration tokens

govynnus:
See https://github.com/matrix-org/matrix-doc/pull/3231/commits/79939f955b34113404b5d363a981f790aa7ec0e7

### DIFF
--- a/changelog.d/10887.bugfix
+++ b/changelog.d/10887.bugfix
@@ -1,0 +1,1 @@
+Allow `.` and `~` characters when creating registration tokens as per the [change to MSC3231](https://github.com/matrix-org/matrix-doc/pull/3231/commits/79939f955b34113404b5d363a981f790aa7ec0e7).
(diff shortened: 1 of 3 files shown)


anoadragon453:
Looks like the unit test `test_create_unable_to_generate_token` is now failing. @govynnus could you take a look?

govynnus:
Hmm, I did run the unit tests before pushing, but I'll take a look

govynnus:
Oh it must be because it now doesn't add all possible 1 char tokens to the db, so there's a possibility creating one succeeds.

anoadragon453:
@govynnus Oh, right indeed! Lucky we caught it here then :)

changelog.d/10887.bugfix
```diff
@@ -0,0 +1 @@
+Allow `.` and `~` characters when creating registration tokens as per the [change to MSC3231](https://github.com/matrix-org/matrix-doc/pull/3231/commits/79939f955b34113404b5d363a981f790aa7ec0e7).
```
clokep, line 1: I think we usually just link to the MSC, not the specific commit in it?

clokep (approved):

changelog.d/10887.bugfix
clokep:
```suggestion
Allow the `.` and `~` characters when creating registration tokens as per the change to [MSC3231](https://github.com/matrix-org/matrix-doc/pull/3231).
```

govynnus: Ok, thanks

PR #10889, opened 2021-09-23
Remove unnecessary parentheses around tuples returned from methods

anoadragon453:
I was fairly sure we had cleaned up all of these at some point in the past, but I found a few more while spelunking through old bits of the codebase. They annoyed me enough that I've now rooted them all out for good(?).

There weren't enough occurrences to mess about with `sed`. This is the result of just going through the 20 or so results and manually correcting them. Thus a quick look-over of each one would be best.

### DIFF
--- a/changelog.d/10889.misc
+++ b/changelog.d/10889.misc
@@ -1,0 +1,1 @@
+Clean up some unnecessary parentheses in places around the codebase.
--- a/synapse/config/server.py
+++ b/synapse/config/server.py
@@ -1447,7 +1447,7 @@ def read_gc_thresholds(thresholds):
         return None
     try:
         assert len(thresholds) == 3
-        return (int(thresholds[0]), int(thresholds[1]), int(thresholds[2]))
+        return int(thresholds[0]), int(thresholds[1]), int(thresholds[2])
     except Exception:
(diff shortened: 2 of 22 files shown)


DMRobertson (approved):
Slightly surprised that `black` doesn't have an opinion on this!

richvdh:
I'm also surprised flake8 doesn't complain about it. have we disabled a check or something?

anoadragon453:
PyCharm was the tool to point this out to me. Unfortunately the specific inspection for this in PyCharm doesn't look to be linked to PEP8 anywhere, nor can I see that [we've disabled anything](https://github.com/matrix-org/synapse/blob/26f2bfedbf5493d8a69d1b38147b6236e7606cd3/setup.cfg#L13-L21) related to redundant parentheses in flake8. There also doesn't look to be a relevant [error code](https://pycodestyle.pycqa.org/en/latest/intro.html#error-codes) for redundant parentheses.

Nothing [disabled for black](https://github.com/matrix-org/synapse/blob/26f2bfedbf5493d8a69d1b38147b6236e7606cd3/pyproject.toml#L37-L56) either.

We could code up a check specifically for it, but that sounds a bit overkill.

PR #10891, opened 2021-09-23
Improve typing in user_directory files

DMRobertson:
This makes the user_directory.py in storage pass most of mypy's
checks (including `no-untyped-defs`). Unfortunately that file is in the
tangled web of Store class inheritance so doesn't pass mypy at the moment.

Non-functional change.

The handlers directory has already been mypyed.

### DIFF
--- a/changelog.d/10891.misc
+++ b/changelog.d/10891.misc
@@ -1,1 +1,1 @@
-Improve type hinting in the user_directory code.
+Improve type hinting in the user directory code.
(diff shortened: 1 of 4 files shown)


changelog.d/10891.misc
```diff
@@ -0,0 +1 @@
+Improve type hinting in the user_directory code.
```
reivilibre, line 1:
```suggestion
Improve type hinting in the user directory code.
```
either that or mention an actual package with backticks

reivilibre (approved):
If Mypy is happy, I'm happy.

richvdh (approved):
lgtm

PR #10894, opened 2021-09-23
Add a spamchecker method to allow or deny 3pid invites

babolivier:

Note that a module callback already exists for 3pid invites (https://matrix-org.github.io/synapse/develop/modules/third_party_rules_callbacks.html#check_threepid_can_be_invited) but it doesn't check whether the sender of the invite is allowed to send it.

### DIFF
--- a/docs/modules/spam_checker_callbacks.md
+++ b/docs/modules/spam_checker_callbacks.md
@@ -29,6 +29,41 @@ Called when processing an invitation. The module must return a `bool` indicating
 the inviter can invite the invitee to the given room. Both inviter and invitee are
 represented by their Matrix user ID (e.g. `@alice:example.com`).
 
+### `user_may_send_threepid_invite`
+
+```python
+async def user_may_send_3pid_invite(
+    inviter: str,
+    invitee: Dict[str, str],
+    room_id: str,
+) -> bool
+```
+
+Called when processing an invitation using a third-party identifier (also called a 3PID,
+e.g. an email address or a phone number). The module must return a `bool` indicating
+whether the inviter can invite the invitee to the given room.
+
+The inviter is represented by their Matrix user ID (e.g. `@alice:example.com`), and the
+invitee is represented by a dict describing the third-party identifier to send an
+invitation to, with a `medium` key indicating the identifier's medium (e.g. "email") and
+an `address` key indicating the identifier's address (e.g. `alice@example.com`). See
+[the Matrix specification](https://matrix.org/docs/spec/appendices#pid-types) for more
+information regarding third-party identifiers.
+
+For example, a call to this callback to send an invitation to the email address
+`alice@example.com` would look like this:
+
+```python
+await user_may_send_3pid_invite(
+    "@bob:example.com",  # The inviter's user ID
+    {"medium": "email", "address": "alice@example.com"},  # The 3PID to invite
+    "!some_room:example.com",  # The ID of the room to send the invite into
+)
+```
+
+**Note**: If the third-party identifier is already associated with a matrix user ID,
+[`user_may_invite`](#user_may_invite) will be used instead.
+
 ### `user_may_create_room`
 
 ```python
--- a/synapse/events/spamcheck.py
+++ b/synapse/events/spamcheck.py
@@ -45,6 +45,10 @@ CHECK_EVENT_FOR_SPAM_CALLBACK = Callable[
     Awaitable[Union[bool, str]],
 ]
 USER_MAY_INVITE_CALLBACK = Callable[[str, str, str], Awaitable[bool]]
+USER_MAY_SEND_3PID_INVITE_CALLBACK = Callable[
+    [str, Dict[str, str], str],
+    Awaitable[bool]
+]
 USER_MAY_CREATE_ROOM_CALLBACK = Callable[[str], Awaitable[bool]]
 USER_MAY_CREATE_ROOM_ALIAS_CALLBACK = Callable[[str, RoomAlias], Awaitable[bool]]
 USER_MAY_PUBLISH_ROOM_CALLBACK = Callable[[str, str], Awaitable[bool]]
@@ -163,6 +167,9 @@ class SpamChecker:
     def __init__(self):
         self._check_event_for_spam_callbacks: List[CHECK_EVENT_FOR_SPAM_CALLBACK] = []
         self._user_may_invite_callbacks: List[USER_MAY_INVITE_CALLBACK] = []
+        self._user_may_send_3pid_invite_callbacks: List[
+            USER_MAY_SEND_3PID_INVITE_CALLBACK
+        ] = []
         self._user_may_create_room_callbacks: List[USER_MAY_CREATE_ROOM_CALLBACK] = []
         self._user_may_create_room_alias_callbacks: List[
             USER_MAY_CREATE_ROOM_ALIAS_CALLBACK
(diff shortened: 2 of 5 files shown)


synapse/events/spamcheck.py
```diff
@@ -266,6 +281,32 @@ async def user_may_invite(
...
+        If this method returns false, the threepid invite will be rejected.
+
+        Note that if the threepid is already associated with a Matrix user ID, Synapse
+        will call user_may_invite with said user ID instead.
+
+        Args:
+            inviter_userid: The user ID of the sender of the invitation
+            invitee_threepid: The threepid targeted in the invitation, as a dict including
+                a "medium" key indicating the threepid's medium (e.g. "email") and an
+                "address" key indicating the threepid's address (e.g. "alice@example.com")
```
babolivier, line 298: Wondering if I shouldn't use an `attr.s` class instead here, opinions welcome.

clokep: Do we ever expect anything besides these two fields? Would it make sense to send them as separate parameters to the function?

tests/rest/client/test_rooms.py
```diff
@@ -2315,3 +2315,140 @@ def test_bad_alias(self):
...
+        self.assertEquals(
+            allow_mock.call_args, expected_call_args, allow_mock.call_args
+        )
+
+        # Check that the call to send the invite was made.
+        self.assertEquals(make_invite_mock.call_count, 1)
+
+        # Now change the return value of the callback to deny any invite and test that
+        # we can't send the invite.
+        return_value = False
```
clokep, line 2391: Are we sure this works? Changing a variable like this shouldn't be caught by the scope of `_user_may_send_3pid_invite`?

tests/rest/client/test_rooms.py
```diff
@@ -2315,3 +2315,140 @@ def test_bad_alias(self):
...
+            room_id: str,
+            user: UserID,
+            txn_id: Optional[str],
+            id_access_token: Optional[str] = None,
+        ) -> int:
+            return 0
+
+        mock = Mock(side_effect=_make_and_store_3pid_invite)
+
+        self.hs.get_room_member_handler()._make_and_store_3pid_invite = mock
```
clokep, line 2430: If we don't care about the parameters could we just do:

```suggestion
        self.hs.get_room_member_handler()._make_and_store_3pid_invite = Mock(return_value=0)
```

A similar thing could be done for `_mock_lookup_3pid`. They could probably be inlined then instead.

babolivier: This doesn't work for all the versions of Python we support, since some of them don't support `await`ing on a `Mock` unless it has an `async` side effect

tests/rest/client/test_rooms.py
babolivier: It does work, otherwise the test wouldn't pass.

tests/rest/client/test_rooms.py
babolivier: Though now I don't remember which versions of Python I've tested this with, so I've opened https://github.com/matrix-org/synapse/pull/10928 to make the CI check if that's still the case.

babolivier: As an example, when testing your suggestion locally:

```
2021-09-27 17:13:01+0100 [-] 2021-09-27 17:13:01,823 - synapse.http.server - 93 - ERROR - POST-4 - Failed handle request via 'RoomMembershipRestServlet': <SynapseRequest at 0x7f1bb01eb0a0 method='POST' uri='/_matrix/client/r0/rooms/!vDuFLwPWZJCFsnpqsV:test/invite' clientproto='1.1' site='test'>
	  File "/home/babolivier/Documents/matrix/synapse/synapse/http/server.py", line 258, in _async_render_wrapper
	  File "/home/babolivier/Documents/matrix/synapse/synapse/http/server.py", line 446, in _async_render
	  File "/home/babolivier/Documents/matrix/synapse/synapse/rest/client/room.py", line 790, in on_POST
	    await self.room_member_handler.do_3pid_invite(
	  File "/home/babolivier/Documents/matrix/synapse/synapse/handlers/room_member.py", line 1278, in do_3pid_invite
	    stream_id = await self._make_and_store_3pid_invite(
	TypeError: object int can't be used in 'await' expression
```

synapse/events/spamcheck.py
babolivier: Good point, it would probably be better this way indeed

tests/rest/client/test_rooms.py
clokep: You can use `make_awaitable` from `tests.test_utils` for that:

`Mock(return_value=make_awaitable(0))`.

babolivier: Oooh that is true!

synapse/handlers/room_member.py
```diff
@@ -1260,10 +1260,22 @@ async def do_3pid_invite(
         if invitee:
             # Note that update_membership with an action of "invite" can raise
             # a ShadowBanError, but this was done above already.
+            # We don't check the invite against the spamchecker(s) here (through
+            # user_may_invite) because we'll do it further down the line anyway (in
+            # update_membership_locked).
```
DMRobertson, line 1265: This is a bit confusing because at first glance we seem to do exactly this on 1271. Maybe something like

```suggestion
            # We only run the 3pid spamchecker here; the usual spamchecker 
            # methods will be invoked further down the line (in
            # update_membership_locked).
```

tests/rest/client/test_rooms.py
```diff
@@ -2430,3 +2430,75 @@ def test_bad_alias(self):
...
+                "address": email_to_invite,
+            },
+            access_token=self.tok,
+        )
+        self.assertEquals(channel.code, 200)
+
+        # Check that the callback was called with the right params.
+        expected_call_args = ((self.user_id, "email", email_to_invite, self.room_id),)
+
+        self.assertEquals(mock.call_args, expected_call_args, mock.call_args)
```
DMRobertson, line 2482: Fwiw there's [Mock.assert_called_once_with](https://docs.python.org/3/library/unittest.mock.html?highlight=mock#unittest.mock.Mock.assert_called_once_with) and similarly mock.assert_called_once. Might be slightly less typing or give you a nicer error message? :shrug: 

Six and two threes, I suspect

docs/modules/spam_checker_callbacks.md
```diff
@@ -29,6 +29,41 @@ Called when processing an invitation. The module must return a `bool` indicating
...
+await user_may_send_3pid_invite(
+    "@bob:example.com",  # The inviter's user ID
+    "email",  # The medium of the 3PID to invite
+    "alice@example.com",  # The address of the 3PID to invite
+    "!some_room:example.com",  # The ID of the room to send the invite into
+)
+```
+
+**Note**: If the third-party identifier is already associated with a matrix user ID,
+[`user_may_invite`](#user_may_invite) will be used instead.
```
DMRobertson, line 65: Is this property important enough to have a test check? In other words, if we started calling 3pid spamchecker callbacks in this situation, could that cause pain for people?

(I'm not sure either way---maybe slightly leaning towards "no")

tests/rest/client/test_rooms.py
babolivier: Oh yeah, I completely forgot about those

synapse/handlers/room_member.py
babolivier: > This is a bit confusing because at first glance we seem to do exactly this on 1271

Well yeah but L1271 is on a different code path so the comment make sense (i.e. we managed to turn the 3pid invite into a matrix one but we don't check it now because we already do it further down the line).

DMRobertson: > Well yeah but L1271 is on a different code path

Argh, so it is. Thank you.

docs/modules/spam_checker_callbacks.md
babolivier: Yeah I'm not sure that's necessary.

DMRobertson (approved):

PR #10896, opened 2021-09-23
Factor out common code for persisting fetched auth events

richvdh:
We currently have two codepaths that are used for persisting fetched auth events:
 * `_get_events_and_persist`, which is used to fetch the state events and auth events for a new backwards extremity.
 * `_get_remote_auth_chain_for_event`, which is called when we are missing the auth events for any other regular event.

There is common code between the two paths which can be factored out.

Should be reviewable commit by commit. It's independent of #10901, but the similarity between the two codepaths might be clearer once that has landed.

### DIFF
--- a/synapse/handlers/federation_event.py
+++ b/synapse/handlers/federation_event.py
@@ -1142,7 +1142,7 @@ class FederationEventHandler:
     async def _auth_and_persist_fetched_events(
         self, origin: str, room_id: str, events: Iterable[EventBase]
     ) -> None:
-        """Persist the events fetched by _get_events_and_persist.
+        """Persist the events fetched by _get_events_and_persist or _get_remote_auth_chain_for_event
 
         The events should not depend on one another, e.g. this should be used to persist
         a bunch of outliers, but not a chunk of individual events that depend
@@ -1643,33 +1643,9 @@ class FederationEventHandler:
             x for x in remote_auth_chain if x.event_id not in seen_remotes
         ]
 
-        for auth_event in remote_auth_chain:
-            try:
-                auth_ids = auth_event.auth_event_ids()
-                auth = {
-                    (e.type, e.state_key): e
-                    for e in remote_auth_chain
-                    if e.event_id in auth_ids or e.type == EventTypes.Create
-                }
-
-                logger.debug(
-                    "_check_event_auth %s missing_auth: %s",
-                    event_id,
-                    auth_event.event_id,
-                )
-                missing_auth_event_context = EventContext.for_outlier()
-                missing_auth_event_context = await self._check_event_auth(
-                    destination,
-                    auth_event,
-                    missing_auth_event_context,
-                    claimed_auth_event_map=auth,
-                )
-                await self.persist_events_and_notify(
-                    room_id,
-                    [(auth_event, missing_auth_event_context)],
-                )
-            except AuthError:
-                pass
+        await self._auth_and_persist_fetched_events(
+            destination, room_id, remote_auth_chain
+        )
 
     async def _update_context_for_auth_events(
         self, event: EventBase, context: EventContext, auth_events: StateMap[EventBase]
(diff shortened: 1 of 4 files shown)


synapse/handlers/federation_event.py
```diff
@@ -1137,7 +1136,30 @@ async def get_event(event_id: str) -> None:
...
-        logger.info("Fetched %i events of %i requested", len(event_map), len(event_ids))
+        logger.info("Fetched %i events of %i requested", len(events), len(event_ids))
+        await self._auth_and_persist_fetched_events(destination, room_id, events)
+
+    async def _auth_and_persist_fetched_events(
+        self, origin: str, room_id: str, events: Iterable[EventBase]
+    ) -> None:
+        """Persist the events fetched by _get_events_and_persist.
+
+        The events should not depend on one another, e.g. this should be used to persist
```
erikjohnston, line 1147: I think its worth being explicit that by "depend" we mean due to `prev_events`, and that this class will correctly handle dependencies of `auth_events`.

synapse/handlers/federation_event.py
```diff
@@ -1137,7 +1136,30 @@ async def get_event(event_id: str) -> None:
...
+        a bunch of outliers, but not a chunk of individual events that depend
+        on each other for state calculations.
+
+        We first sort the events so that they do not depend on each other, then
+        persist them.
```
erikjohnston, line 1152: But we just said that the events don't depend on each other :innocent:

synapse/handlers/federation_event.py
```diff
@@ -1168,30 +1190,26 @@ async def get_event(event_id: str) -> None:
                 "Persisting %i of %i remaining events", len(roots), len(event_map)
             )
 
-            await self._auth_and_persist_fetched_events(destination, room_id, roots)
+            await self._auth_and_persist_fetched_events_inner(origin, room_id, roots)
```
erikjohnston, line 1193: Hmm, we actually shouldn't need to persist the events separately, e.g. we don't do this when we are persisting the state/auth chains when joining a room?

synapse/handlers/federation_event.py
richvdh: Note that I'm just moving this text from elsewhere, but ok point taken, I'll try to improve this :)

synapse/handlers/federation_event.py
richvdh: I'm not really following you.

The idea here (and it's an existing idea from https://github.com/matrix-org/synapse/pull/10771, not new in this PR) is that we make sure we auth each event's auth events before the event itself. This is all part of trying to fix https://github.com/matrix-org/synapse/issues/9595. 

And yes, I fear we don't do it on joins, but that's a bug we should also fix.

erikjohnston: For every event we have persisted we need to have persisted its entire auth chain, but I think its fine to persist the entire auth chain in one go? (We might want to sort the event list topologically to make sure the chunking works in the persist_events class). In the persist events class if we fail to persist one event in a batch then we'll fail the entire batch, so there shouldn't be a danger of us persisting an event but not its auth events?

erikjohnston (approved):
There's a question over https://github.com/matrix-org/synapse/pull/10896#discussion_r715497948, but that's orthogonal to this PR

synapse/handlers/federation_event.py
richvdh: We discussed this in #synapse-dev.

Part of the confusion here is that I'm specifically concerned about event *auth*, rather than persisting the events per se. 

Erik's concern is that this could end up with a lot of round-trips between federation reader and event persister for a long auth chain.

So we can look at optimising this so that we do all the auth up-front, and *then* persist everything together. But I think this will be much easier after some more cleanup here, so we've agreed to park it for now.

PR #10898, opened 2021-09-23
Add a spamchecker callback to allow or deny room creation based on invites

babolivier:

This adds a callback that's fairly similar to `user_may_create_room` except it also allows processing based on the invites sent at room creation.


tests/rest/client/test_rooms.py
```diff
@@ -669,6 +669,121 @@ def test_post_room_invitees_ratelimit(self):
...
+            json.dumps({"invite": invited_mxids}).encode("utf8"),
+        )
+        self.assertEqual(200, channel.code)
+
+        # Check that the callback was called with the right arguments.
+        expected_call_args = ((self.user_id, invited_mxids, []),)
+        self.assertEquals(
+            callback_mock.call_args,
+            expected_call_args,
+            callback_mock.call_args,
```
DMRobertson, line 751: IIRC the third argument here is a description of the check---not sure this one would make sense if the test were to regress in the future?

tests/rest/client/test_rooms.py
```diff
@@ -669,6 +669,121 @@ def test_post_room_invitees_ratelimit(self):
...
+
+        # Check that do_3pid_invite was called the right amount of time
+        self.assertEquals(do_3pid_invite_mock.call_count, len(invited_3pids))
+
+        # Check that the callback was called with the right arguments.
+        expected_call_args = ((self.user_id, [], invited_3pids),)
+        self.assertEquals(
+            callback_mock.call_args,
+            expected_call_args,
+            callback_mock.call_args,
```
DMRobertson, line 770: and ditto here

docs/modules/spam_checker_callbacks.md
```diff
@@ -38,6 +38,35 @@ async def user_may_create_room(user: str) -> bool
...
+
+See [the Matrix specification](https://matrix.org/docs/spec/appendices#pid-types) for more
+information regarding third-party identifiers.
+
+If no invite and/or 3PID invite were specified in the room creation request, the
+corresponding list(s) will be empty.
+
+**Note**: This callback is not called when a room is cloned (e.g. during a room upgrade)
+since no invites are sent when cloning a room. To cover this case, modules also need to
+implement `user_may_create_room`.
```
DMRobertson, line 68: This wasn't obvious to me from the diff below. I'm guessing that there's a different code path for cloning a room? (i.e. not via the room handler's `create_room`?)

DMRobertson (approved):

tests/rest/client/test_rooms.py
babolivier: The argument is the message to print if the assertion fails. If it does, the output is truncated, so it's handy to force it to print the full call args for debugging.

docs/modules/spam_checker_callbacks.md
babolivier: Yeah, it's goes through its `clone_existing_room`.

PR #10902, opened 2021-09-23
Update utility code to handle C implementations of frozendict

H-Shay:
Background: some versions of frozendict contain a C implementation of frozendict that was incompatible with our code. This PR adds logic to catch and handle those cases. See also https://github.com/matrix-org/python-canonicaljson/issues/36 for more context.

### DIFF
--- a/changelog.d/10902.misc
+++ b/changelog.d/10902.misc
@@ -1,0 +1,1 @@
+Update utility code to handle C implementations of frozendict.
--- a/synapse/util/__init__.py
+++ b/synapse/util/__init__.py
@@ -50,7 +50,10 @@ def _handle_frozendict(obj: Any) -> Dict[Any, Any]:
     if type(obj) is frozendict:
         # fishing the protected dict out of the object is a bit nasty,
         # but we don't really want the overhead of copying the dict.
-        return obj._dict
+        try:
+            return obj._dict
+        except AttributeError:
+            return dict(obj)
     raise TypeError(
         "Object of type %s is not JSON serializable" % obj.__class__.__name__
     )
--- a/synapse/util/__init__.py
+++ b/synapse/util/__init__.py
@@ -53,6 +53,9 @@ def _handle_frozendict(obj: Any) -> Dict[Any, Any]:
         try:
             return obj._dict
         except AttributeError:
+            # When the C implementation of frozendict is used,
+            # there isn't a `_dict` attribute with a dict
+            # so we resort to making a copy of the frozendict
             return dict(obj)
     raise TypeError(
         "Object of type %s is not JSON serializable" % obj.__class__.__name__


reivilibre:
equivalent: https://github.com/matrix-org/python-canonicaljson/issues/37

abadger:
Is `_handle_frozendict()` still needed?  I see that setup.py says that python3.6 or higher is required for synapse and it looks like frozendict can be dumped without `_handle_frozendict()`:
```
Python 3.6.15 (default, Sep  5 2021, 00:00:00) 
[GCC 10.3.1 20210422 (Red Hat 10.3.1-1)] on linux
>>> import json, frozendict
>>> a = frozendict.frozendict(((1,2),(3,4)))
>>> json.dumps(a)
'{"1": 2, "3": 4}'
>>> 
```

I note that https://github.com/matrix-org/python-canonicaljson/issues/36 is referenced above and python-canonicaljson says it supports python-3.5 and above.  In python-3.5, `_handle_frozendict()` would still be needed:

```
Python 3.5.10 (default, Sep 17 2021, 00:00:00) 
[GCC 10.3.1 20210422 (Red Hat 10.3.1-1)] on linux
>>> import json, frozendict
>>> a = frozendict.frozendict(((1,2),(3,4)))
>>> json.dumps(a)
  File "/usr/lib64/python3.5/json/__init__.py", line 230, in dumps
    return _default_encoder.encode(obj)
  File "/usr/lib64/python3.5/json/encoder.py", line 198, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/usr/lib64/python3.5/json/encoder.py", line 256, in iterencode
    return _iterencode(o, 0)
  File "/usr/lib64/python3.5/json/encoder.py", line 179, in default
    raise TypeError(repr(o) + " is not JSON serializable")
TypeError: <frozendict {1: 2, 3: 4}> is not JSON serializable
```

erikjohnston:
> Is `_handle_frozendict()` still needed? I see that setup.py says that python3.6 or higher is required for synapse and it looks like frozendict can be dumped without `_handle_frozendict()`:

Looks like the pure python version now inherits from `dict`, which means the standard json encoder will correctly serialise it. Looks to still be a problem when using the C extension though:

```python
Python 3.8.10 (default, Jun  2 2021, 10:49:15) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.17.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import frozendict, json                                                                                                                                                                                   

In [2]: frozendict.c_ext                                                                                                                                                                                          
Out[2]: True

In [3]: frozendict.__version__                                                                                                                                                                                    
Out[3]: '2.0.6'

In [4]: json.dumps(frozendict.frozendict({}))                                                                                                                                                                     
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
```

erikjohnston:
(I should also say thanks to @abadger for pointing this out! It's useful info)

DMRobertson:
For completeness, see notes here: https://github.com/matrix-org/python-canonicaljson/issues/36#issuecomment-927816293

squahtx (approved):

PR #10903, opened 2021-09-24
Drop backwards-compatibility support for "outlier"

### LINKED ISSUE #9670: Drop backwards-compatibility support for "outlier" in the database
Since #9411, Synapse no longer makes use of the `outlier` flag in the `internal_metadata` column. Once we are satisfied that nobody will need to roll back to Synapse 1.30 or earlier, we should stop populating that flag.

richvdh:
Before Synapse 1.31 (#9411), we relied on `outlier` being stored in the `internal_metadata` column. We can now assume nobody will roll back their deployment that far and drop the legacy support.

Fixes #9670.

### DIFF
--- a/synapse/storage/schema/__init__.py
+++ b/synapse/storage/schema/__init__.py
@@ -12,9 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# When updating these values, please leave a short summary of the changes below.
-
-SCHEMA_VERSION = 64
+SCHEMA_VERSION = 64  # remember to update the list below when updating
 """Represents the expectations made by the codebase about the database schema
 
 This should be incremented whenever the codebase changes its requirements on the
@@ -35,7 +33,7 @@ Changes in SCHEMA_VERSION = 63:
 """
 
 
-SCHEMA_COMPAT_VERSION = 59
+SCHEMA_COMPAT_VERSION = 60  # 60: "outlier" not in internal_metadata.
 """Limit on how far the synapse codebase can be rolled back without breaking db compat
 
 This value is stored in the database, and checked on startup. If the value in the
--- a/changelog.d/10903.misc
+++ b/changelog.d/10903.misc
@@ -1,0 +1,1 @@
+Drop old functionality which maintained database compatibility with Synapse versions before 1.31.
--- a/synapse/storage/databases/main/events.py
+++ b/synapse/storage/databases/main/events.py
@@ -1276,13 +1276,6 @@ class PersistEventsStore:
                     logger.exception("")
                     raise
 
-                # update the stored internal_metadata to update the "outlier" flag.
-                # TODO: This is unused as of Synapse 1.31. Remove it once we are happy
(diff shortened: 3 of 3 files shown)


synapse/storage/schema/__init__.py
```diff
@@ -35,7 +33,7 @@
 """
 
 
-SCHEMA_COMPAT_VERSION = 59
+SCHEMA_COMPAT_VERSION = 60  # 60: "outlier" not in internal_metadata.
```
richvdh, line 36: strictly speaking, the change to schema version 60 came later than #9411, but near enough.

erikjohnston (approved):

synapse/storage/schema/__init__.py
anoadragon453: Should this note be in a comment on this line, or in a `Changes in SCHEMA_COMPAT_VERSION = 60:` section as above?

richvdh: > Changes in SCHEMA_COMPAT_VERSION = 60

well, that would be a bad name for the section, since it's not really a change in the "schema compat version" - I'm not even sure that makes any sense. But it's a fair question to ask if it should be in a bit of a longer comment rather than a one-liner here.

What I'm trying to avoid is the situation where somebody bumps the number and forgets to update the comment. I figure if they are on the same line, you have to try extra hard to mess it up. But it does only leave me with about 50 characters, which makes for a rather terse comment. 

We could go with a separate comment and *also* a "# remember to update the comment above" comment, but that feels kinda clunky too.

So, open to suggestions/discussion.

anoadragon453: I was mostly just worried about there only being room for a single note here, which would get overwritten in the future when other changes are made.

But are historical changes not a concern here?

reivilibre: Maybe we could match the style of the other comment, but come up with a more accurate heading?

```
Changes leading to SCHEMA_COMPAT_VERSION = 60:
    - "outlier" is no longer stored in event_json.internal_metadata for backwards compatibility
```

I do worry that putting a comment on the line has led you to making it a bit terse that I didn't know what you meant until reading the code.

However, wouldn't mind if you thought that we may well not care about the history and the presence of a comment may not particularly matter — git blame can point you to where things are changed.

richvdh: > But are historical changes not a concern here?

well no, not really. Once we dropped support for synapses with `SCHEMA_VERSION == 60`, we'll no longer really care about the reasons we dropped support for synapses with `SCHEMA_VERSION == 59`.  It's the same reason we don't keep a complete history in https://github.com/matrix-org/synapse/blob/develop/synapse/python_dependencies.py: as long as we don't support Twisted 16, who the hell cares why we don't support Twisted 14?

richvdh: maybe we don't need a comment here at all. After all, there's always git. I just thought it might be helpful for the curious.

PR #10905, opened 2021-09-24
Encode JSON responses on a thread in C, mk2

erikjohnston:
This is the same as #10844, but with fewer commits.

### DIFF
--- a/synapse/http/server.py
+++ b/synapse/http/server.py
@@ -711,6 +711,27 @@ def respond_with_json_bytes(
     return NOT_DONE_YET
 
 
+async def _async_write_json_to_request_in_thread(
+    request: SynapseRequest,
+    json_encoder: Callable[[Any], bytes],
+    json_object: Any,
+):
+    """Encodes the given JSON object on a thread and then writes it to the
+    request.
+
+    This is done so that encoding large JSON objects doesn't block the reactor
+    thread.
+
+    Note: We don't use JsonEncoder.iterencode here as that falls back to the
+    Python implementation (rather than the C backend), which is *much* more
+    expensive.
+    """
+
+    json_str = await defer_to_thread(request.reactor, json_encoder, json_object)
+
+    _write_bytes_to_request(request, json_str)
+
+
 def _write_bytes_to_request(request: Request, bytes_to_write: bytes) -> None:
     """Writes the bytes to the request using an appropriate producer.
 
@@ -830,24 +851,3 @@ def finish_request(request: Request):
         request.finish()
     except RuntimeError as e:
         logger.info("Connection disconnected before response was written: %r", e)
-
-
-async def _async_write_json_to_request_in_thread(
-    request: SynapseRequest,
-    json_encoder: Callable[[Any], bytes],
-    json_object: Any,
-):
-    """Encodes the given JSON object on a thread and then writes it to the
-    request.
-
-    This is done so that encoding large JSON objects doesn't block the reactor
-    thread.
-
-    Note: We don't use JsonEncoder.iterencode here as that falls back to the
-    Python implementation (rather than the C backend), which is *much* more
-    expensive.
-    """
-
-    json_str = await defer_to_thread(request.reactor, json_encoder, json_object)
-
-    _write_bytes_to_request(request, json_str)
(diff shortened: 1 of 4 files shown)


synapse/http/server.py
```diff
@@ -706,15 +707,35 @@ def respond_with_json_bytes(
...
-    # the original `bytes`).
-    bytes_io = BytesIO(json_bytes)
-
-    producer = NoRangeStaticProducer(request, bytes_io)
-    producer.start()
+    _write_json_bytes_to_request(request, json_bytes)
     return NOT_DONE_YET
 
 
+def _write_json_bytes_to_request(request: Request, json_bytes: bytes) -> None:
```
erikjohnston, line 714: From https://github.com/matrix-org/synapse/pull/10844#discussion_r713034847

> is this specific to being json bytes, or could it be used for any byte sequence?

It can be anything that slices to `bytes`, but I don't see any in built typing for that (`ByteString` seems to just be an alias for `Sequence[int]]`...). I can add a protocol, or do `Union[bytes, bytearray]`? I'm not sure its worth its worth it though

synapse/http/server.py
```diff
@@ -706,15 +707,35 @@ def respond_with_json_bytes(
     if send_cors:
         set_cors_headers(request)
 
-    # note that this is zero-copy (the bytesio shares a copy-on-write buffer with
-    # the original `bytes`).
-    bytes_io = BytesIO(json_bytes)
-
-    producer = NoRangeStaticProducer(request, bytes_io)
-    producer.start()
+    _write_json_bytes_to_request(request, json_bytes)
```
erikjohnston, line 710: From https://github.com/matrix-org/synapse/pull/10844#discussion_r713039021

> why is this changing in this PR? It seems unrelated to how we encode JSON responses?

True, I can leave as is, but it just felt cleaner to have both `respond_with_json` and `respond_with_json_bytes` use the same producer. Leaving as is does mean a) returning large responses can lead to timeouts, and b) anything that writes responses via this method in the replication unit tests will fail (due to `Request` converting pull producers to push producers via the default reactor, which is never run. I am sorely tempted to ban all pull producers because I lost like half a day due to that, but I digress).

synapse/http/server.py
```diff
@@ -809,3 +830,24 @@ def finish_request(request: Request):
         request.finish()
     except RuntimeError as e:
         logger.info("Connection disconnected before response was written: %r", e)
```
erikjohnston, line 832: From https://github.com/matrix-org/synapse/pull/10844#discussion_r713032338

> is this really the only possible reason that a RuntimeError can be raised?

Probably not, but this is what it was before.

synapse/http/server.py
richvdh: I'm mostly asking: why don't we just call it `write_bytes_to_request` (and also `s/json_bytes/bytes_to_write/` or something)?

erikjohnston: Ooooooooooooooooooooooooooooooooooooooooooooh.

Yup, happy to rename.

synapse/http/server.py
richvdh: :+1:

synapse/http/server.py
```diff
@@ -809,3 +830,24 @@ def finish_request(request: Request):
...
+
+
+async def _async_write_json_to_request_in_thread(
```
richvdh, line 835: can you move this further up the file (above `_write_json_bytes_to_request`), so that the flow jumps up and down less?

richvdh (approved):
the change looks fine, modulo the nit below.

It makes me sad though. I thought that *not* loading the whole of a HUGE `/sync` response into memory was an epic win when we did that. It makes me wonder if we should find another json encoding implementation that is both efficient *and* iterative. (Or maybe we can even fix the CPython C implementation to work with iterencode, and get it merged into `simplejson`).

erikjohnston:
> It makes me sad though. I thought that _not_ loading the whole of a HUGE `/sync` response into memory was an epic win when we did that. It makes me wonder if we should find another json encoding implementation that is both efficient _and_ iterative. (Or maybe we can even fix the CPython C implementation to work with iterencode, and get it merged into `simplejson`).

Yeah, it is sad and I agree it would be good to find a way that is both fast and iterative. (For future readers: the reason for changing this is that the iterative approach is really *really* slow for large blobs).

PR #10906, opened 2021-09-24
Update postgresql testing script

DMRobertson:
- Use sytest:bionic. Sytest:latest is two years old (do we want
  CI to push out latest at all?) and comes with Python 3.5, which we
  explictly no longer support. The script now runs under PostgreSQL 10
  as a result.
- Pass through argv to `tox` and `trial` so we can run specific test(s) in isolation.

A few changes to make it speedier to re-run the tests:

- Create blank DB in the container, not the script, so we don't have to
  `initdb` each time
- Use a named volume to persist the tox environment, so we don't have to
  fetch and install a bunch of packages from PyPI each time


DMRobertson:
I just tried this with
```shell
scripts-dev/test_postgresql.sh tests.replication.test_sharded_event_persister.EventPersisterShardTestCase
```
Which took 41 seconds on a re-run. Not bad(?)

changelog.d/10906.misc
```diff
@@ -0,0 +1 @@
+Update development testing script `test_postgresql.sh` to use a support Python version and make re-runs quicker.
```
reivilibre, line 1:
```suggestion
Update development testing script `test_postgresql.sh` to use a supported Python version and make re-runs quicker.
```

docker/Dockerfile-pgtests
```diff
@@ -1,12 +1,21 @@
...
+FROM matrixdotorg/sytest:bionic
 
 # The Sytest image doesn't come with python, so install that
 RUN apt-get update && apt-get -qq install -y python3 python3-dev python3-pip
 
 # We need tox to run the tests in run_pg_tests.sh
 RUN python3 -m pip install tox
 
+# Initialise the db
+RUN su -c '/usr/lib/postgresql/10/bin/initdb -D /var/lib/postgresql/data -E "UTF-8" --lc-collate="en_US.UTF-8" --lc-ctype="en_US.UTF-8" --username=postgres' postgres
```
reivilibre, line 12: why American locales? Don't we definitely want `C` as the locale? (Synapse will refuse to set up databases in the wrong locale.)

Perhaps this doesn't matter here (maybe this is just the `postgres` management database?) but it strikes me as odd that we'd specify a specific locale if we don't care about it.

reivilibre: Oh hmm you nicked it from `run_pg_tests.sh`. Odd, but maybe we shouldn't remove it after all if we don't know why it's there

docker/Dockerfile-pgtests
```diff
@@ -1,12 +1,21 @@
 # Use the Sytest image that comes with a lot of the build dependencies
 # pre-installed
-FROM matrixdotorg/sytest:latest
+FROM matrixdotorg/sytest:bionic
```
reivilibre, line 3: I suppose we want this old image so we can give it a fair test on old Python and old Postgres versions?

docker/Dockerfile-pgtests
```diff
@@ -1,12 +1,21 @@
...
+
+# Create a volume for tox's workdir (so we don't have to fetch dependencies from PyPI every run)
+VOLUME /tox
+
 ADD run_pg_tests.sh /pg_tests.sh
-ENTRYPOINT /pg_tests.sh
+
+# Use the "exec form" of ENTRYPOINT (https://docs.docker.com/engine/reference/builder/#entrypoint)
+# so that we can `docker run` this container and pass arguments to pg_tests.sh
+ENTRYPOINT ["/pg_tests.sh"]
```
reivilibre, line 21: Whilst you're here, do you think it'd be sensible to make it keep the same filename when it's copied into the container? It's a bit sad to have two different (context-dependent) names for the same script; I was about to look to see if there was another script called 'pg_tests.sh'.

scripts-dev/test_postgresql.sh
```diff
@@ -0,0 +1,22 @@
...
+
+# Build, and tag
+docker build docker/ -f docker/Dockerfile-pgtests -t synapsepgtests
+
+# Create a place to persist tox's environment so reruns are faster.
+# NB: I'm using a separate docker volume here, because you may have already run
+# "tox -e py36-postgres" on your system. You might not be using the distro used
+# in Dockerfile-pgtests, and so might be using a different CPython build! I don't
+# want there to be any accidental clashes
+docker volume create synapse-pg-test-tox
```
reivilibre, line 19: Is there a benefit to explicitly creating the volume? I can't remember ever needing to do this and thought `docker run` creates volumes automatically.

However I also think that it's convenient to having a nearby path mounted as the volume, rather than having to go hunting in `/var/lib/docker` wherever it is, when you need to read the logs (the logs *are* deposited into `tox`'s environment, right? correct me if wrong).

reivilibre (commented):
Thanks, I'm sure this will be useful to many (setting up a postgres locally can be a little bit faffy)

docker/Dockerfile-pgtests
DMRobertson: Yeah, I just transplanted this into the dockerfile. Even if we don't know why it's there, I think this should probably be `C`. (No-one's used this in ages anyway!)

docker/Dockerfile-pgtests
DMRobertson: That was my thinking---to be conservative. I'm guessing that most newcomers are just going to test against the system python+postgres, and so it's probably easier for them to use the newer ones rather than the older.

(Bionic may be old, but [its EOL is April 2023](https://wiki.ubuntu.com/Releases))

docker/Dockerfile-pgtests
DMRobertson: Very happy to do this!

scripts-dev/test_postgresql.sh
DMRobertson: Yeah, it goes into tox's environment AFAICS.

Maybe we should create a new tox env called say `localtest-postgres`? (I don't want to re-use an existing one, since the container's Python interperter might differ from that on your local system)

There's also the fun and games of making sure that the files persisted to the host aren't written as root (grr).

reivilibre: My thought was something like mounting `$(PWD)/.tox-postgres` or something to `/tox`? Anything sensible is fine, however, as long as people know where to look for the logs (so we should write down where that is in the contributing guide)

reivilibre: > aren't written as root (grr).

hmm, sad. I forgot about that. I don't know how to deal with that.

reivilibre:
One thing worth thinking about: as a user, how can I instruct the testing to run 20 jobs in parallel? I think we might have an env var `TRIAL_FLAGS: "--jobs=20"`, but you probably need to do special tricks to let that seep into the container?

DMRobertson:
> One thing worth thinking about: as a user, how can I instruct the testing to run 20 jobs in parallel? I think we might have an ability to do this with an env var: `TRIAL_FLAGS="--jobs=20"`, but you probably need to do special tricks to let that seep into the container?

Note that run_pg_tests.sh already has `export TRIAL_FLAGS="-j 4"` right before it invokes tox

reivilibre:
> Note that run_pg_tests.sh already has `export TRIAL_FLAGS="-j 4"` right before it invokes tox

ah yes. Slight shame it's not configurable, but don't want to scope creep this if it's not trivial to do

DMRobertson:
Rerunning the script, I get
- a tox environment owned by me at `SYNAPSE_SRC/.tox-pg-container`
  - inside that, a virtualenv and tox's logs
- `_trial_temp` as usual(?), owned by me
- `matrix_synapse.egg-info` as part of the installation process, owned by me
- no files owned by anyone who isn't me, according to `find . -not -uid $(id -u)`

reivilibre (approved):
Thanks!

docs/development/contributing_guide.md
```diff
@@ -170,6 +170,53 @@ To increase the log level for the tests, set `SYNAPSE_TEST_LOG_LEVEL`:
 SYNAPSE_TEST_LOG_LEVEL=DEBUG trial tests
 ```
 
+### Running tests under PostgreSQL
+
+Invoking `trial` as above will use an in-memory SQLite database. This is great for
+quick development and testing. However, we only support using a PostgreSQL database
```
reivilibre, line 176: I'm not aware that this is true, but perhaps:
```suggestion
quick development and testing. However, we recommend using a PostgreSQL database
```

docs/development/contributing_guide.md
```diff
@@ -170,6 +170,53 @@ To increase the log level for the tests, set `SYNAPSE_TEST_LOG_LEVEL`:
...
+export SYNAPSE_POSTGRES_HOST=localhost
+export SYNAPSE_POSTGRES_USER=postgres
+export SYNAPSE_POSTGRES_PASSWORD=mydevenvpassword
+trial
+```
+
+#### Prebuilt container
+
+Since configuring PostgreSQL can be fiddley, we can use of a pre-made
+docker container to host postgres and run our tests for us. To do so, run
```
reivilibre, line 203:
```suggestion
Since configuring PostgreSQL can be fiddly, we can use of a pre-made
Docker container to set up PostgreSQL and run our tests for us. To do so, run
```

docs/development/contributing_guide.md
```diff
@@ -170,6 +170,53 @@ To increase the log level for the tests, set `SYNAPSE_TEST_LOG_LEVEL`:
...
+Any extra arguments to the script will be passed to `tox` and then to `trial`,
+so we can run a specific test in this container with e.g.
+
+```shell
+scripts-dev/test_postgresql.sh tests.replication.test_sharded_event_persister.EventPersisterShardTestCase
+```
+
+The container creates a folder in your Synapse checkout called
+`.tox-pg-container` and uses this as a tox environment. The output of any
+`trial` runs goes into `_trial_temp` in your synapse source directory---the same
```
reivilibre, line 218:
```suggestion
`trial` runs goes into `_trial_temp` in your synapse source directory — the same
```

triple-hyphen was too much for me ;p

reivilibre (approved):
thanks for remembering the docs :)

docs/development/contributing_guide.md
DMRobertson: I had in mind https://matrix-org.github.io/synapse/latest/setup/installation.html?highlight=sqlite#using-postgresql , which says:

> SQLite is only acceptable for testing purposes. SQLite should not be used in a production server. Synapse will perform poorly when using SQLite, especially when participating in large rooms.

That said, still happy to make the change.

docs/development/contributing_guide.md
DMRobertson: Too much LaTeX in my murky past.

docs/development/contributing_guide.md
reivilibre: I didn't know we said something quite so strong about it, but I would probably still make the change anyway as we would still support SQLite in the sense of making sure it doesn't just fall over and explode (even if it does run slowly, though I have to admit an old server of mine was on SQLite in the early days and it was good enough for what I was doing back then).

PR #10910, opened 2021-09-24
Add a spamchecker callback to allow or deny room joins

babolivier:

This adds the `user_may_join_room` spam checker callback that synapse-dinsic has had since https://github.com/matrix-org/synapse/pull/4875.


synapse/events/spamcheck.py
```diff
@@ -259,6 +265,24 @@ async def check_event_for_spam(
 
         return False
 
+    async def user_may_join_room(self, userid: str, room_id: str, is_invited: bool):
```
erikjohnston, line 268: It makes me somewhat sad that we seem to use `userid` in this file whereas eveywhere else in synapse we use `user_id`

synapse/handlers/room_member.py
```diff
@@ -718,6 +725,30 @@ async def update_membership_locked(
...
                     raise AuthError(403, "Guest access not allowed")
 
+            # Figure out whether the user is a server admin to determine whether they
+            # should be able to bypass the spam checker.
+            if (
+                self._server_notices_mxid is not None
+                and requester.user.to_string() == self._server_notices_mxid
+            ):
+                # allow the server notices mxid to join rooms
+                is_requester_admin = True
```
erikjohnston, line 735:
```suggestion
                should_check_join_with_spam_checker = True
```

Or something, as its a bit confusing for `is_requester_admin` to be true when the requester isn't an admin.

synapse/events/spamcheck.py
babolivier: Yeah fair. This was just copy-pasta that I forgot to fix.

```suggestion
    async def user_may_join_room(self, user_id: str, room_id: str, is_invited: bool):
```

docs/modules/spam_checker_callbacks.md
```diff
@@ -19,6 +19,18 @@ either a `bool` to indicate whether the event must be rejected because of spam,
...
+
+```python
+async def user_may_join_room(user: str, room: str, is_invited: bool) -> bool
+```
+
+Called when a user is trying to join a room. The module must return a `bool` to indicate
+whether the user can join the room. The user is represented by their Matrix user ID (e.g.
+`@alice:example.com`) and the room is represented by its Matrix ID (e.g.
+`!room:example.com`). The module is also given a boolean to indicate whether the user
+currently has a pending invite in the room.
```
erikjohnston, line 32: Probably worth explicitly noting that we don't call this function for a) server admins and b) room creations.

erikjohnston (approved):

docs/modules/spam_checker_callbacks.md
babolivier: Ah yeah good call

PR #10912, opened 2021-09-24
e2e: return newly fetched cross-signing keys

TheJJ:
without this patch, only the cached cross-signing keys are returned when
a client requests them, although a federation request is done
in case there's no cache entry.
this patch returns newly fetched entries directly.

This is a stripped-down version of #10668, which needs some discussion and/or root cause analysis.

### DIFF
--- a/changelog.d/10912.bugfix
+++ b/changelog.d/10912.bugfix
@@ -1,0 +1,1 @@
+return E2E cross signing keys fetched over federation when to a client.
--- a/synapse/handlers/e2e_keys.py
+++ b/synapse/handlers/e2e_keys.py
@@ -168,6 +168,22 @@ class E2eKeysHandler:
                 ) = await self.store.get_user_devices_from_cache(query_list)
                 for user_id, devices in remote_results.items():
                     user_devices = results.setdefault(user_id, {})
+
+                    cached_cross_master = user_id in cross_signing_keys["master_keys"]
+                    self_sig_key = cross_signing_keys["self_signing_keys"].get(
+                        user_id, {}
+                    )
+                    cached_cross_selfsigning = bool(self_sig_key)
+
+                    # check if we are missing only one of cross-signing master or
+                    # self-signing key, but the other one is cached.
+                    # as we need both, this will issue a federation request.
+                    # if we don't have any of the keys, either the user doesn't have
+                    # cross-signing set up, or we did not fetch the
+                    # cross-signing keys yet since the device list is not (yet) updated.
+                    if cached_cross_master ^ cached_cross_selfsigning:
+                        user_ids_not_in_cache.add(user_id)
+
                     for device_id, device in devices.items():
                         keys = device.get("keys", None)
                         device_display_name = device.get("device_display_name", None)
@@ -178,21 +194,9 @@ class E2eKeysHandler:
                                 unsigned["device_display_name"] = device_display_name
                             user_devices[device_id] = result
 
-                # check for missing cross-signing keys.
-                for user_id in remote_queries.keys():
-                    cached_cross_master = user_id in cross_signing_keys["master_keys"]
-                    cached_cross_selfsigning = (
-                        user_id in cross_signing_keys["self_signing_keys"]
-                    )
-
-                    # check if we are missing only one of cross-signing master or
-                    # self-signing key, but the other one is cached.
-                    # as we need both, this will issue a federation request.
-                    # if we don't have any of the keys, either the user doesn't have
-                    # cross-signing set up, or the cached device list
-                    # is not (yet) updated.
-                    if cached_cross_master ^ cached_cross_selfsigning:
-                        user_ids_not_in_cache.add(user_id)
+                            # TODO: invalidate device cache if we have cached
+                            # device keys but not cross-signing keys,
+                            # although the user should have cross-signing keys.
 
                 # add those users to the list to fetch over federation.
                 for user_id in user_ids_not_in_cache:
@@ -234,6 +238,9 @@ class E2eKeysHandler:
                     # probably be tracking their device lists. However, we haven't
                     # done an initial sync on the device list so we do it now.
                     try:
+                        # here, we fetch the user's device list
+                        # and cross signing keys,
+                        # and update our cache with them.
                         if self._is_master:
                             user_devices = await self.device_handler.device_list_updater.user_device_resync(
                                 user_id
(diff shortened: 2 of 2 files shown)


synapse/handlers/e2e_keys.py
```diff
@@ -168,32 +168,36 @@ async def query_devices(
...
 
-                # check for missing cross-signing keys.
-                for user_id in remote_queries.keys():
                     cached_cross_master = user_id in cross_signing_keys["master_keys"]
-                    cached_cross_selfsigning = (
-                        user_id in cross_signing_keys["self_signing_keys"]
+                    self_sig_key = cross_signing_keys["self_signing_keys"].get(
+                        user_id, {}
                     )
+                    cached_cross_selfsigning = bool(self_sig_key)
```
clokep, line 176: I'm not sure if I'm missing something, but it looks like this logic is identical to the previous version?

changelog.d/10912.bugfix
```diff
@@ -0,0 +1 @@
+return E2E cross signing keys fetched over federation when to a client.
```
richvdh, line 1: This doesn't really make sense? Please can you expand it into something that will be meaningful to other synapse admins. 

Also, do we know when this bug was introduced? Please include that info.

synapse/handlers/e2e_keys.py
```diff
@@ -234,6 +238,9 @@ async def do_remote_query(destination: str) -> None:
                     # probably be tracking their device lists. However, we haven't
                     # done an initial sync on the device list so we do it now.
                     try:
+                        # here, we fetch the user's device list
+                        # and cross signing keys,
+                        # and update our cache with them.
```
richvdh, line 243: please
use
longer
lines

synapse/handlers/e2e_keys.py
```diff
@@ -168,32 +168,36 @@ async def query_devices(
...
+                    cached_cross_selfsigning = bool(self_sig_key)
 
                     # check if we are missing only one of cross-signing master or
                     # self-signing key, but the other one is cached.
                     # as we need both, this will issue a federation request.
                     # if we don't have any of the keys, either the user doesn't have
-                    # cross-signing set up, or the cached device list
-                    # is not (yet) updated.
+                    # cross-signing set up, or we did not fetch the
+                    # cross-signing keys yet since the device list is not (yet) updated.
```
richvdh, line 183: I'm not really following this. Why would the device list not be updated, and why does it cause us to not fetch the cross-signing keys?

Also, is it relevant to the bug being fixed in this PR, or is it a separate correction you're doing at the same time? (It doesn't matter either way - I'm just trying to understand what this PR is fixing!)

synapse/handlers/e2e_keys.py
```diff
@@ -168,32 +168,36 @@ async def query_devices(
...
-                        device_display_name = device.get("device_display_name", None)
-                        if keys:
-                            result = dict(keys)
-                            unsigned = result.setdefault("unsigned", {})
-                            if device_display_name:
-                                unsigned["device_display_name"] = device_display_name
-                            user_devices[device_id] = result
 
-                # check for missing cross-signing keys.
-                for user_id in remote_queries.keys():
```
richvdh, line 182: so we're now doing the following code just on the results of the query, rather than the devices we asked for. That may well be valid, but can you explain why we need to make this change?

synapse/handlers/e2e_keys.py
```diff
@@ -168,32 +168,36 @@ async def query_devices(
...
                     user_devices = results.setdefault(user_id, {})
-                    for device_id, device in devices.items():
-                        keys = device.get("keys", None)
-                        device_display_name = device.get("device_display_name", None)
-                        if keys:
-                            result = dict(keys)
-                            unsigned = result.setdefault("unsigned", {})
-                            if device_display_name:
-                                unsigned["device_display_name"] = device_display_name
-                            user_devices[device_id] = result
```
richvdh, line 179: It looks like this code is moving further down. Can you explain why we need to do that?

synapse/handlers/e2e_keys.py
```diff
@@ -247,6 +254,15 @@ async def do_remote_query(destination: str) -> None:
                         user_results = results.setdefault(user_id, {})
                         for device in user_devices:
                             user_results[device["device_id"]] = device["keys"]
+
+                        # update the result with the devicelist's cross-signing keys
```
richvdh, line 258: can you expand this comment a bit to say why this is necessary?

synapse/handlers/e2e_keys.py
```diff
@@ -247,6 +254,15 @@ async def do_remote_query(destination: str) -> None:
...
+                        cross_signing_keys["master_keys"][user_id] = user_devices.get(
+                            "master_key"
+                        )
+                        cross_signing_keys["self_signing_keys"][
+                            user_id
+                        ] = user_devices.get("self_signing_key")
```
richvdh, line 264: this will overwrite any existing values in `cross_signing_keys`. Is that correct behaviour?

richvdh (changes requested):
Thanks for making this contribution! Sorry it's been sitting for a while - I think the problem is that nobody on the team understands the changes here.

I'm afraid I'm really struggling to understand what is changing here, and why. I'm not familiar with the existing code, and it's very hard to understand. (I think it was written by someone outside the synapse core team - we certainly wouldn't have a single function over 250 lines long normally).

In other words - we could really do with your help to understand what's changing here, and why. Any comments you can add to the PR to help us understand would be appreciated.

Also: please can you add a regression test to `tests/handlers/test_e2e_keys.py` demonstrating the problem?

synapse/handlers/e2e_keys.py
richvdh: +1. I found the original easier to understand (and has the advantage of not introducing yet another local variable to this massive function).

callahad: The logic is not quite identical: the proposed code would behave differently if `user_id` was present in the self-signing keys dict but set to a falsey value, like an empty dict.

I wonder if this might more clearly express intent:

```suggestion
                    cached_cross_selfsigning = bool(cross_signing_keys["self_signing_keys"].get(user_id)
```

The question then becomes: are there scenarios where this could happen, and would we be better off preventing _those_ instead? (E.g., stripping falsey values from the dicts before testing for membership)

richvdh:
unfortunately this has now bitrotted. @TheJJ if you're still interested in contributing this then that would be wonderful, but in the meantime I'm going to go ahead and close it.

synapse/handlers/e2e_keys.py
callahad: From looking at `get_cross_signing_keys_from_cache()`, it may be possible for a falsey value to end up in `cross_signing_keys["self_signing_keys"]`.

There's a call tree that's basically:

- `get_cross_signing_keys_from_cache()`
    - `store.get_e2e_cross_signing_keys_bulk()` (Docstring: "If a user's cross-signing keys were not found, either their user ID will not be in the dict, __or their user ID will map to None.__")
        - `store._get_bare_e2e_cross_signing_keys_bulk_txn()` (Docstring: "If a user's cross-signing keys were not found, __their user ID will not be in the dict.__")

So worth figuring out what's actually true. Or being defensive here.

(Also worth seeing if this code survived Erik's refactor in the related PR)

callahad: The same code is present in the current version of `e2e_keys.py`, so still worth digging into.

PR #10913, opened 2021-09-24
Fix race conditions when creating media store and config directories

squahtx:
This ought to fix a couple of errors seen on matrix.org:
https://sentry.matrix.org/sentry/synapse-matrixorg/?query=is%3Aunresolved+FileExistsError


clokep (approved):
TIL that parameter exists.

PR #10914, opened 2021-09-24
Consistently exclude users from directory

DMRobertson:
Gather the logic which determines if we should exclude a user from the directory into one place. I think this helps to make the code a lot more followable!

I think of this primarily as a tidy-up. But it does have a perceptible change. With this patch, rebuilding the directory will not include support and deactivated users, but it did before AFAICS.

### DIFF
--- a/synapse/handlers/user_directory.py
+++ b/synapse/handlers/user_directory.py
@@ -132,12 +132,7 @@ class UserDirectoryHandler(StateDeltasHandler):
         # FIXME(#3714): We should probably do this in the same worker as all
         # the other changes.
 
-        # Support users are for diagnostics and should not appear in the user directory.
-        is_support = await self.store.is_support_user(user_id)
-        # When change profile information of deactivated user it should not appear in the user directory.
-        is_deactivated = await self.store.get_user_deactivated_status(user_id)
-
-        if not (is_support or is_deactivated):
+        if not await self.store.is_excluded_from_user_dir(user_id):
             await self.store.update_profile_in_user_dir(
                 user_id, profile.display_name, profile.avatar_url
             )
@@ -229,8 +224,8 @@ class UserDirectoryHandler(StateDeltasHandler):
                     else:
                         logger.debug("Server is still in room: %r", room_id)
 
-                is_support = await self.store.is_support_user(state_key)
-                if not is_support:
+                excluded = await self.store.is_excluded_from_user_dir(state_key)
+                if not excluded:
                     if change is MatchChange.no_change:
                         # Handle any profile changes
                         await self._handle_profile_change(
@@ -357,12 +352,8 @@ class UserDirectoryHandler(StateDeltasHandler):
             # First, if they're our user then we need to update for every user
             if self.is_mine_id(user_id):
 
-                is_appservice = self.store.get_if_app_services_interested_in_user(
-                    user_id
-                )
-
-                # We don't care about appservice users.
-                if not is_appservice:
+                excluded = await self.store.is_excluded_from_user_dir(user_id)
+                if not excluded:
                     for other_user_id in other_users_in_room:
                         if user_id == other_user_id:
                             continue
@@ -374,10 +365,8 @@ class UserDirectoryHandler(StateDeltasHandler):
                 if user_id == other_user_id:
                     continue
 
-                is_appservice = self.store.get_if_app_services_interested_in_user(
-                    other_user_id
-                )
-                if self.is_mine_id(other_user_id) and not is_appservice:
+                excluded = await self.store.is_excluded_from_user_dir(other_user_id)
+                if not excluded:
                     to_insert.add((other_user_id, user_id))
 
             if to_insert:
--- a/synapse/storage/databases/main/user_directory.py
+++ b/synapse/storage/databases/main/user_directory.py
@@ -543,7 +543,7 @@ class UserDirectoryBackgroundUpdateStore(StateDeltasStore):
             desc="get_user_in_directory",
         )
 
-    async def update_user_directory_stream_pos(self, stream_id: int) -> None:
+    async def update_user_directory_stream_pos(self, stream_id: Optional[int]) -> None:
         await self.db_pool.simple_update_one(
             table="user_directory_stream_pos",
             keyvalues={},
--- a/synapse/storage/databases/main/user_directory.py
+++ b/synapse/storage/databases/main/user_directory.py
@@ -238,9 +238,10 @@ class UserDirectoryBackgroundUpdateStore(StateDeltasStore):
 
                 # Update each user in the user directory.
                 for user_id, profile in users_with_profile.items():
-                    await self.update_profile_in_user_dir(
-                        user_id, profile.display_name, profile.avatar_url
-                    )
+                    if not await self.is_excluded_from_user_dir(user_id):
+                        await self.update_profile_in_user_dir(
+                            user_id, profile.display_name, profile.avatar_url
+                        )
 
                 to_insert = set()
 
@@ -349,10 +350,11 @@ class UserDirectoryBackgroundUpdateStore(StateDeltasStore):
         )
 
         for user_id in users_to_work_on:
-            profile = await self.get_profileinfo(get_localpart_from_id(user_id))
-            await self.update_profile_in_user_dir(
-                user_id, profile.display_name, profile.avatar_url
-            )
+            if not await self.is_excluded_from_user_dir(user_id):
+                profile = await self.get_profileinfo(get_localpart_from_id(user_id))
+                await self.update_profile_in_user_dir(
+                    user_id, profile.display_name, profile.avatar_url
+                )
 
             # We've finished processing a user. Delete it from the table.
             await self.db_pool.simple_delete_one(
--- a/synapse/storage/databases/main/user_directory.py
+++ b/synapse/storage/databases/main/user_directory.py
@@ -246,7 +246,7 @@ class UserDirectoryBackgroundUpdateStore(StateDeltasStore):
 
                 if is_public:
                     for user_id in users_with_profile:
-                        if self.get_if_app_services_interested_in_user(user_id):
+                        if await self.is_excluded_from_user_dir(user_id):
                             continue
 
                         to_insert.add(user_id)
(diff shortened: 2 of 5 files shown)


synapse/storage/databases/main/user_directory.py
```diff
@@ -369,6 +371,22 @@ def _get_next_batch(txn: LoggingTransaction) -> Optional[List[str]]:
...
+        if self.hs.is_mine_id(user):
+            # Support users are for diagnostics and should not appear in the user directory.
+            if await self.is_support_user(user):
+                return True
+            # Deactivated users aren't contactable, so should not appear in the user directory.
+            if await self.get_user_deactivated_status(user):
+                return True
+            # App service users aren't usually contactable, so exclude them.
+            if self.get_if_app_services_interested_in_user(user):
+                return True
```
reivilibre, line 387: I'm unsure what to think about this.
I have some appservice-generated users in *my* homeserver's search directory. I can't confirm or deny whether I've done something wrong to wind up with that happening, but thought it was worth mentioning in case this should trigger the question of 'do we really intend to exclude AS users from the directory?'.

Both yes and no make sense here. No, on a lot of networks (IRC probably? Discord when using a guild-bot.), the users aren't contactable.
But yes, on some networks, users are contactable and there's nothing inherently wrong with letting you discover them when trying to start a DM ...

I would guess that the 'No' cases outweigh the 'Yes' cases, so I think what you're doing is sensible.

reivilibre (commented):
Thanks for taking the time to unify these! Seems good to me.

synapse/storage/databases/main/user_directory.py
```diff
@@ -369,6 +371,22 @@ def _get_next_batch(txn: LoggingTransaction) -> Optional[List[str]]:
 
         return len(users_to_work_on)
 
+    async def is_excluded_from_user_dir(self, user: str) -> bool:
+        """Certain classes of local user are omitted from the user directory.
+        Is this user one of them?
+        """
+        if self.hs.is_mine_id(user):
```
richvdh, line 378: please can we do an early return rather than a big nested if statement?

synapse/handlers/user_directory.py
```diff
@@ -132,12 +132,7 @@ async def handle_local_profile_change(
         # FIXME(#3714): We should probably do this in the same worker as all
         # the other changes.
 
-        # Support users are for diagnostics and should not appear in the user directory.
-        is_support = await self.store.is_support_user(user_id)
-        # When change profile information of deactivated user it should not appear in the user directory.
-        is_deactivated = await self.store.get_user_deactivated_status(user_id)
-
-        if not (is_support or is_deactivated):
+        if not await self.store.is_excluded_from_user_dir(user_id):
```
richvdh, line 135: I find the double-negative here a bit mind-bending. Perhaps we should instead have an `should_include_in_user_dir` ?

synapse/handlers/user_directory.py
```diff
@@ -357,12 +352,8 @@ async def _handle_new_user(
             # First, if they're our user then we need to update for every user
             if self.is_mine_id(user_id):
```
richvdh, line 353: think this is redundant?

synapse/storage/databases/main/user_directory.py
```diff
@@ -238,15 +238,16 @@ def _get_next_batch(
 
                 # Update each user in the user directory.
                 for user_id, profile in users_with_profile.items():
-                    await self.update_profile_in_user_dir(
-                        user_id, profile.display_name, profile.avatar_url
-                    )
+                    if not await self.is_excluded_from_user_dir(user_id):
```
richvdh, line 241: it's probably fine, but for the record: what's the logic behind adding this here, and at line 353, where there was no check before?

synapse/storage/databases/main/user_directory.py
richvdh, line 387: wonder if we should do this test first, since it doesn't require a database hit and should be pretty quick.

richvdh: It seems like *most* of the code currently excludes AS users from the directory, but there are enough bugs that they leak through.

At least we're making it consistent here. There might be scope for a future change where we make this configurable per-appservice, but let's leave it for now.

tests/handlers/test_user_directory.py
```diff
@@ -25,12 +24,18 @@
...
 
 from tests import unittest
+from tests.storage.test_user_directory import GetUserDirectoryTables
 from tests.unittest import override_config
 
 
-class UserDirectoryTestCase(unittest.HomeserverTestCase):
-    """
-    Tests the UserDirectoryHandler.
+class UserDirectoryTestCase(GetUserDirectoryTables, unittest.HomeserverTestCase):
```
richvdh, line 31: I'm not a fan of collecting utility functions by inheritance. It makes the dataflows harder to follow.

Make the methods in `GetUserDirectoryTables` async to avoid the dependency on `get_success`, put `self._get_user_directory_tables_helper = GetUserDirectoryTables(self.store)` in `prepare`, and then do `self.get_success(self._get_user_directory_tables_helper.get_users_in_public_rooms(...))`.

tests/handlers/test_user_directory.py
```diff
@@ -119,19 +124,22 @@ def test_handle_user_deactivated_support_user(self):
...
         )
-
-        self.store.remove_from_user_dir = Mock(return_value=defer.succeed(None))
-        self.get_success(self.handler.handle_local_user_deactivated(s_user_id))
-        self.store.remove_from_user_dir.not_called()
+        with patch.object(
+            self.store, "remove_from_user_dir", return_value=defer.succeed(None)
+        ) as mock:
+            self.get_success(self.handler.handle_local_user_deactivated(s_user_id))
+        self.get_success(mock.not_called())
```
richvdh, line 131: Does this do what it's supposed to? [`Mock`](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock) doesn't seem to have a `not_called` method?

tests/handlers/test_user_directory.py
```diff
@@ -119,19 +124,22 @@ def test_handle_user_deactivated_support_user(self):
...
             self.store.register_user(user_id=r_user_id, password_hash=None)
         )
-        self.store.remove_from_user_dir = Mock(return_value=defer.succeed(None))
-        self.get_success(self.handler.handle_local_user_deactivated(r_user_id))
-        self.store.remove_from_user_dir.called_once_with(r_user_id)
+        with patch.object(
+            self.store, "remove_from_user_dir", return_value=defer.succeed(None)
+        ) as mock:
+            self.get_success(self.handler.handle_local_user_deactivated(r_user_id))
+        self.get_success(mock.called_once_with(r_user_id))
```
richvdh, line 142: ... or a `called_once_with` method?

richvdh (changes requested):
I'm struggling to follow what's moving where in the unit tests, and why the changes you're making are safe - but I'm happy to take your word that it's still testing the right sort of thing. (For future reference: it might be preferable to split this into two PRs, where first of all we refactor all the tests, and secondly we clean up the code, might be preferable, as a way of demonstrating that we're not fundamentally changing behaviour).

A few thoughts and suggestions.

tests/handlers/test_user_directory.py
DMRobertson: I don't think so. Has this been broken for a while then!?

tests/handlers/test_user_directory.py
DMRobertson: I guess the original author was missing an `assert_` from the method names.

reivilibre: Hm, this is concerning (and it was very possibly me, though I can't remember). Is there any way we can do this that will properly result in a failure if a typo is made?

synapse/storage/databases/main/user_directory.py
DMRobertson: Consistency with `_handle_deltas` in the handlers code. 

Rebuilding the user directory from scratch (here) should have the same outcome as starting with an empty directory and building it by listening for membership events (`_handle_deltas`). Before my changes, the latter excluded support and deactivated users, so I think the former should too.

tests/handlers/test_user_directory.py
DMRobertson: changing to `assert_not_called` makes the test fail! 

My reading: the test is trying to ensure that `handle_local_user_deactivated` doesn't try to delete a user from the db who would never have been in the directory. But I think `handle_local_user_deactivated` has never had any logic in it, so I think this test has been incorrectly passing forever.

We could make `handle_local_user_deactivated` call `is_excluded_from_user_directory`, but that involves a SELECT from the db anyway to see if they're deactivated or a support user. I'm guessing that's the same order of magnitude of delay as doing a no-op deletion? And so I think the current behaviour is fine.

I guess it's worth having this kind of test here though, to make sure that synapse doesn't fall over when we deactivate a support user?

DMRobertson: > I guess it's worth having this kind of test here though, to make sure that synapse doesn't fall over when we deactivate a support user?

Additional: maybe test_user_directory isn't the right place for such a test.

tests/handlers/test_user_directory.py
DMRobertson: Different typing stubs for `Mock` maybe---removing its `__getattr__`?

richvdh: as of python 3.7, you can call `mock.seal` (https://docs.python.org/3/library/unittest.mock.html#sealing-mocks). But that doesn't really help you now, and I think it just replaces one thing to forget with another thing to forget.

synapse/handlers/user_directory.py
DMRobertson: Glad it's not just me. I was erring towards preserving the original logic below as much as possible, but think it'll be better this way round.

synapse/handlers/user_directory.py
DMRobertson: Because we'll check to see if they're local within `is_excluded_from_user_dir`? I agree...

But. I think putting the localness check into `is_excluded_from_user_dir` makes things more confusing. To a first approximation, it's true that we'll never exclude a remote user from the user directory. With one exception: `users_who_share_private_rooms`. This contains triples `(local_user_id, any_user_id, room_id)` where the two users both belong to the given private room. Only local users live in the first entry of those tuples. (I think this is an optimisation?) To maintain that there's extra logic which needs to ensure we don't put remote users into that column.

I think it's going to be clearer if we have a function `should_include_local_user_in_dir` and make the local-vs-remote logic explicit.

DMRobertson:
Thanks both for your reviews. Closing this in lieu of #10891 and #10960.

PR #10915, opened 2021-09-24
Clean-up type hints in server config

clokep:
This cleans up some of the classes / type hints in the server config file.

The first commit is necessary to continue in my quest to eradicate the magic in `RootConfig`, but the others were easy enough to do while I was there.

### DIFF
--- a/synapse/config/server.py
+++ b/synapse/config/server.py
@@ -236,6 +236,26 @@ class RetentionConfig:
     longest_max_lifetime: Optional[int]
 
 
+@attr.s
+class LimitRemoteRoomsConfig:
+    enabled = attr.ib(
+        validator=attr.validators.instance_of(bool), default=False
+    )
+    complexity = attr.ib(
+        validator=attr.validators.instance_of(
+            (float, int)  # type: ignore[arg-type] # noqa
+        ),
+        default=1.0,
+    )
+    complexity_error = attr.ib(
+        validator=attr.validators.instance_of(str),
+        default=ROOM_COMPLEXITY_TOO_GREAT,
+    )
+    admins_can_join = attr.ib(
+        validator=attr.validators.instance_of(bool), default=False
+    )
+
+
 class ServerConfig(Config):
     section = "server"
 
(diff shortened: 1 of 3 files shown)


synapse/config/server.py
```diff
@@ -184,49 +184,74 @@ def generate_ip_set(
 
 @attr.s(frozen=True)
```
clokep, line 185: This doesn't use `slots` since we splat a dictionary directly into it from the user (e.g. `HttpResourceConfig(**config)`).

synapse/config/server.py
```diff
@@ -184,49 +184,74 @@ def generate_ip_set(
...
+@attr.s(slots=True, frozen=True, auto_attribs=True)
+class RetentionConfig:
+    """Object describing the configuration of the manhole"""
+
+    interval: int
+    shortest_max_lifetime: Optional[int]
+    longest_max_lifetime: Optional[int]
+
+
+@attr.s(frozen=True)
```
clokep, line 239: This doesn't use `slots` since we splat a dictionary directly into it from the user (e.g. `LimitRemoteRoomsConfig(**config)`).

synapse/config/server.py
```diff
@@ -184,49 +184,74 @@ def generate_ip_set(
...
-        type=List[str],
+    names: List[str] = attr.ib(
         factory=list,
         validator=attr.validators.deep_iterable(attr.validators.in_(KNOWN_RESOURCES)),  # type: ignore
     )
-    compress = attr.ib(
-        type=bool,
+    compress: bool = attr.ib(
         default=False,
         validator=attr.validators.optional(attr.validators.instance_of(bool)),  # type: ignore[arg-type]
```
DMRobertson, line 193: Is this a style preference or a functional change? (Perhaps mypy et al only recognise the annotation?)

DMRobertson (commented):
Looks great!

synapse/config/server.py
clokep: This is a style preference, mypy pulls it out either way.

If `auto_attribs=True` is used, it is required.

squahtx (approved):

PR #10916, opened 2021-09-24
Use direct references for configuration variables (part 6).

clokep:
Builds on #10798, #10812, #10885, #10893, and #10897.

~~Depends on #10915.~~

Each commit was done with `sed` + `black`, then running the linters and fixing some type hints. Sorry that this one got a bit big.

Now includes configuration variables except those from `registration`.

### DIFF
--- a/changelog.d/10916.misc
+++ b/changelog.d/10916.misc
@@ -1,0 +1,1 @@
+Use direct references to config flags.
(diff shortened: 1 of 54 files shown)


anoadragon453 (approved):
Other than realising that we should *definitely* split up the `server` config section, this looks good to me!

clokep:
> Other than realising that we should _definitely_ split up the `server` config section, this looks good to me!

I had the same thought while doing this, maybe I'll refactor a bit afterward to add more config sections.

Issue #1496, opened 2016-02-12
Debian package for ARM CPUs (SYN-628)


matrixbot:
As a user I would like to have a ARM-package for Debian, so that I can do a apt-get on a Raspber Pi.

(Imported from https://matrix.org/jira/browse/SYN-628)

(Reported by bob)

nioc:
Hello, any news concerning this?

Extarys:
Is there any progress on this? A docker image to run on a raspberry would allow smaller groups and home users to deploy Matrix/Synapse at home quite easily.

aaronraimist:
Note for anyone coming across this issue: The matrix-synapse package in Debian Backports supports ARM

erikjohnston:
Note (mostly to self): This issue refers to the debs published by matrix.org which don't currently support ARM

ShadowJonathan:
#4599 is possibly related to this

marcdraco:
This is still rumbling on I see. The instructions make no mention that this is not available for ARM, and for the most part they work flawlessly until you spot that little grumble about ARM not being supported. 

For a fairly major package (and a darned important one at that, if we're to take back the Internet from the super-rich who deal in OUR private lives) this is pretty important.

Many people would join "the federation" (live long and prosper) if only they could just buy a little ARM box like a RPi or something similar, dump it behind their router and just use it - I'd help with that sort of project, but I'm 'nixed (pardon the pun) because I can't make it work. Even the Docker image of Synapse doesn't work right now so I'm out of my depth and quite perturbed.

otrapersona:
I second the request for ARM64 binaries.

heini:
Wow, 6 years in the making and still no ARM(64) packages available. Just found out the hard way by launching a `t4g` instance on AWS and trying to install Synapse on it. This is just Python, so what's the problem?

heini:
BTW: Does the package contain native (compiled) binaries, or why is its architecture defined as `amd64` instead of `all`?

jahway603:
Issue https://github.com/matrix-org/synapse/issues/14001 could possibly be resolved if a Debian package for ARM CPUs was made available.

DMRobertson:
https://github.com/matrix-org/synapse/issues/1496#issuecomment-641196368:

> Note (mostly to self): This issue refers to the debs published by matrix.org which don't currently support ARM

[...]

PR #10919, opened 2021-09-25
Update installation instructions

richvdh:
Various updates to the install docs.

It starts with switching a couple of the sections around so might be worth reviewing that first.

### DIFF
--- a/docs/setup/installation.md
+++ b/docs/setup/installation.md
@@ -186,7 +186,7 @@ When following this route please make sure that the [Platform-specific prerequis
 System requirements:
 
 - POSIX-compliant system (tested on Linux & OS X)
-- Python 3.5.2 or later, up to Python 3.9.
+- Python 3.6 or later, up to Python 3.9.
 - At least 1GB of free RAM if you want to join large public rooms like #matrix:matrix.org
 
 To install the Synapse homeserver run:
--- a/docs/setup/installation.md
+++ b/docs/setup/installation.md
@@ -18,18 +18,176 @@ that your email address is probably `user@example.com` rather than
 
 ## Installing Synapse
 
-### Installing from source
+### Prebuilt packages
 
-(Prebuilt packages are available for some platforms - see [Prebuilt packages](#prebuilt-packages).)
+Prebuilt packages are available for a number of platforms. These are recommended
+for most users.
 
-When installing from source please make sure that the [Platform-specific prerequisites](#platform-specific-prerequisites) are already installed.
+#### Docker images and Ansible playbooks
+
+There is an official synapse image available at
+<https://hub.docker.com/r/matrixdotorg/synapse> which can be used with
+the docker-compose file available at
+[contrib/docker](https://github.com/matrix-org/synapse/tree/develop/contrib/docker).
+Further information on this including configuration options is available in the README
+on hub.docker.com.
+
+Alternatively, Andreas Peters (previously Silvio Fricke) has contributed a
+Dockerfile to automate a synapse server in a single Docker image, at
+<https://hub.docker.com/r/avhost/docker-matrix/tags/>
+
+Slavi Pantaleev has created an Ansible playbook,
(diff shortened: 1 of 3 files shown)


docs/setup/installation.md
```diff
@@ -18,19 +18,177 @@ that your email address is probably `user@example.com` rather than
...
+There is an official synapse image available at
+<https://hub.docker.com/r/matrixdotorg/synapse> which can be used with
+the docker-compose file available at
+[contrib/docker](https://github.com/matrix-org/synapse/tree/develop/contrib/docker).
+Further information on this including configuration options is available in the README
+on hub.docker.com.
+
+Alternatively, Andreas Peters (previously Silvio Fricke) has contributed a
+Dockerfile to automate a synapse server in a single Docker image, at
+<https://hub.docker.com/r/avhost/docker-matrix/tags/>
```
DMRobertson, line 37: Hadn't seen this syntax before. I thought it'd make markdown sad because I think you can embed HTML in markdown? But apparently [it's A Thing in GFM](https://github.github.com/gfm/#autolinks).

docs/setup/installation.md
```diff
@@ -18,19 +18,177 @@ that your email address is probably `user@example.com` rather than
...
+
+Slavi Pantaleev has created an Ansible playbook,
+which installs the offical Docker image of Matrix Synapse
+along with many other Matrix-related services (Postgres database, Element, coturn,
```
DMRobertson, line 41: Should that be `Element Web` specifically rather than just `Element`?)

docs/setup/installation.md
```diff
@@ -47,8 +47,10 @@ For more details, see
 
 ##### Matrix.org packages
 
-Matrix.org provides Debian/Ubuntu packages of Synapse via
-<https://packages.matrix.org/debian/>.  To install the latest release:
+Matrix.org provides Debian/Ubuntu packages of Synapse, for the amd64
```
DMRobertson, line 50: Don't we also do this for arm?

DMRobertson (approved):
lgtm! Minor comments, not directly part of your changes:

docs/setup/installation.md
reivilibre: FWIW: I don't think this is GFM-specific. (indeed, CommonMark seems to have it for example).

It's commonly used for e-mail addresses in my experience, probably because e-mail itself formats things like `From: Example <example@example.org>`; I would guess `<..>` carried through as a plaintext convention that wound up in Markdown?

docs/setup/installation.md
reivilibre: The answer seems to be no: https://packages.matrix.org/debian/dists/hirsute/main/.
(Some old distros have i386 too, but that appears to have stopped in July. I don't know if this was intentional or not?)

docs/setup/installation.md
```diff
@@ -18,19 +18,179 @@ that your email address is probably `user@example.com` rather than
...
+doas pkg_add synapse
+```
+
+#### NixOS
+
+Robin Lambertz has packaged Synapse for NixOS at:
+<https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/services/misc/matrix-synapse.nix>
+
+
+### Installing as a python module from PyPI
```
reivilibre, line 182:
```suggestion
### Installing as a Python module from PyPI
```

docs/setup/installation.md
DMRobertson: I was thinking of https://github.com/matrix-org/synapse/blob/develop/.github/workflows/docker.yml#L72 but that's a docker container rather than a debian package

richvdh: cf https://github.com/matrix-org/synapse/issues/1496

> Some distros have i386 too, but that appears to have stopped in July.

It was a long time before that. We stopped doing i386 packages when we dropped the python2 packages, back in January 2019.

PR #10922, opened 2021-09-27
Fix getTurnServer response: integer ttl

lukaslihotzki:
`ttl` must be an integer according to the OpenAPI spec:
https://github.com/matrix-org/matrix-doc/blob/old_master/data/api/client-server/voip.yaml#L70

True division (`/`) returns a float instead (`"ttl": 7200.0`).
Floor division (`//`) returns an integer, so the response is spec compliant.

Like https://github.com/matrix-org/synapse/pull/10175

### DIFF
--- a/changelog.d/10922.bugfix
+++ b/changelog.d/10922.bugfix
@@ -1,0 +1,1 @@
+Fix a minor bug in the response to `/_matrix/client/r0/voip/turnServer`. Contributed by @lukaslihotzki.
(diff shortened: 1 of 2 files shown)


DMRobertson:
Seems reasonable! Do we know if any clients are unintentionally relying on this being a float?

KitsuneRal:
Aside from whatever clients actually use (impact still has to be assessed), whoever tries to tackle it as float does not follow the spec in the first place, see https://matrix.org/docs/spec/client_server/r0.6.1#get-matrix-client-r0-voip-turnserver

clokep (approved):
I think this is fine.

Issue #10863, opened 2021-09-21
Media files in `url_cache/` and `url_cache_thumbnails/` hang around for too long


squahtx:
### Description

Spotted by @richvdh:

Media files in `url_cache/` and `url_cache_thumbnails/` are hanging around for a few days, which is far longer than expected

### Version information

- **Homeserver**: matrix.org

squahtx:
3 and 5 seem to be the expected behaviour:
The code to delete old `url_cache` files has a 2 day threshold for both `url_cache` and `url_cache_thumbnails`: https://github.com/matrix-org/synapse/blob/b93259082c7d8d3fe8376a646e130213d90069dc/synapse/rest/media/v1/preview_url_resource.py#L526-L531

`url_cache` files will be deleted before the 2 day threshold if their expiry time is short enough, but it comes from an oEmbed response:
https://github.com/matrix-org/synapse/blob/b93259082c7d8d3fe8376a646e130213d90069dc/synapse/rest/media/v1/preview_url_resource.py#L439

and when inspecting the `local_media_repository_url_cache` database table on a live instance, about 80% of records expire more than 2 days into the future.

richvdh:
oh, interesting. @clokep is capping the expiry of oembed results in #10814.

clokep:
> oh, interesting. @clokep is capping the expiry of oembed results in #10814.

The current code in #10814 applies it to *all* URL previews, not just oEmbed, but this is kind of a lie since (currently) non-oEmbed ones always get an expiry time of 1 hour.

From the above it looks like we don't expire things when that time passes, but we actually expire things that expired 2 days ago. I think what this actually means is that a URL preview result will last *at most* 3 days (after #10814 is merged).

Hopefully that sounds reasonable?

squahtx:
#10924 ought to fix point 4 in the list above (`url_cache_thumbnails/` retains all empty `yyyy-mm-dd/` directories)

squahtx:
Which leaves 1 and 2:
1. `url_cache/` retains some random empty `yyyy-mm-dd/` directories
2. `url_cache/yyyy-mm-dd/` occasionally contains random files older than 2 days

The best explanation I can come up with for these is that sometimes URL preview requests get stuck downloading the URL and outlive the 2 day cache expiry. Then:
 * when the URL download fails, its cache file is removed, but there is no logic to clean up the `yyyy-mm-dd/` directory
[...]

PR #10924, opened 2021-09-27
Fix empty `url_cache_thumbnails/yyyy-mm-dd/` directories being left behind

squahtx:
Fixes observation 4 in #10863.

When we get around to cleaning up the parents of the thumbnail directory, the thumbnail directory itself has already been deleted.  Instead of continuing with the deletion of the rest of the parent paths, we currently abort, which leaves empty directories lying around:
https://github.com/matrix-org/synapse/blob/f7768f62cbf7579a1a91e694f83d47d275373369/synapse/rest/media/v1/preview_url_resource.py#L565-L570

### DIFF
--- a/synapse/rest/media/v1/preview_url_resource.py
+++ b/synapse/rest/media/v1/preview_url_resource.py
@@ -13,6 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import datetime
+import errno
 import fnmatch
 import itertools
 import logging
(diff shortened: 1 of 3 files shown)


synapse/rest/media/v1/preview_url_resource.py
```diff
@@ -504,20 +504,22 @@ async def _expire_url_cache_data(self) -> None:
...
+            dirs = self.filepaths.url_cache_filepath_dirs_to_delete(media_id)
+            for dir in dirs:
+                try:
                     os.rmdir(dir)
-            except Exception:
-                pass
+                except FileNotFoundError:
+                    pass  # Already deleted, continue with deleting the rest
+                except Exception:
+                    break  # Failed, skip deleting the rest of the parent dirs
```
richvdh, line 522: we should probably log something here?

synapse/rest/media/v1/preview_url_resource.py
```diff
@@ -530,44 +532,50 @@ async def _expire_url_cache_data(self) -> None:
...
```
richvdh, line 556: and here?

synapse/rest/media/v1/preview_url_resource.py
```diff
@@ -530,44 +532,50 @@ async def _expire_url_cache_data(self) -> None:
...
+            # removed by the `rmtree` above.
+            for dir in dirs:
+                try:
                     os.rmdir(dir)
-            except Exception:
-                pass
+                except FileNotFoundError:
+                    pass  # Already deleted, continue with deleting the rest
+                except Exception:
+                    break  # Failed, skip deleting the rest of the parent dirs
```
richvdh, line 578: and here?

richvdh (approved):
lgtm otherwise

synapse/rest/media/v1/preview_url_resource.py
squahtx: Sounds like a good idea. I got tired of the duplicated logic and factored it out into a function.

richvdh (approved):

PR #10925, opened 2021-09-27
Sign the git tag in release script

### LINKED ISSUE #10818: Recent releases are not signed
It seems that [recent releases](https://github.com/matrix-org/synapse/releases) of Synapse are not being signed any longer. I think this is due to the [release script](https://github.com/matrix-org/synapse/blob/14b8c0476f93ea2ed3134e75733e45aa0ab6f5a5/scripts-dev/release.py) not properly signing the tags, but I'm unsure.

(Note that @anoadragon453 mentioned he has been doing the tags manually so his releases do have signatures on them.)

@erikjohnston mentioned this might be due to already having the necessary keys unlocked or not.

erikjohnston:
Fixes #10818

### DIFF
--- a/changelog.d/10925.misc
+++ b/changelog.d/10925.misc
@@ -1,0 +1,1 @@
+Update release script to sign the newly created git tags.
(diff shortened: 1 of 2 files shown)


anoadragon453 (approved):
I'm unable to find the `sign` argument in either the GitPython documentation, or digging into the code itself, but if you've tested this and it works then it's probably fine!

Issue #10920, opened 2021-09-25
"Event not signed by authorising server" when trying to change nickname/avatar in a space-restricted room


pavlukivan:
### Description

I host a homeserver, it has a room with version 9 set to only allow joining for space members, it's upgraded from an old room of version 6. My friend uses the official homeserver, and when he tries to change his per-room avatar it says "Event not signed by authorising server". I am able to reproduce it as well.

### Steps to reproduce

- create a room with space-restricted membership
- join the room with an account on a different homeserver
- try to use /myroomavatar or /myroomnick

### Version information

- **Homeserver**: matrix.org *and* my own (I tested on both)

clokep:
Hm, those checks shouldn't be being applied if the user's previous membership was `join`! I'll take a look.

clokep:
We have a complement test for re-joining the same room (and ensuring it works), but it really should just be sending an updated `m.room.member` event, which is what [Element Web does](https://github.com/matrix-org/matrix-react-sdk/blob/2d1d42b90e8418017348cae1bd17a8a92340fdfb/src/SlashCommands.tsx#L291-L302). Note that they also copy the *entire* previous event and add to it. I think that's probably fine and the server should handle this OK, but not 100% sure.

pavlukivan:
fwiw the steps i listed seem to work no matter the join order, i'm not familiar enough with the synapse codebase to know if that means anything, but just note that it should be fairly easy to reproduce for you as well to check if it's fixed

clokep:
I was able to reproduce, I think what Element Web is doing is fine. The logic is a bit torturous though so haven't solved it yet!

clokep:
Note that #10933 should fix this.

clokep:
Unfortunately the fix just missed Synapse 1.44.0, so will be in 1.45.0. Thanks so much for the detailed report with steps to reproduce! 🎉

PR #10927, opened 2021-09-27
Only do restricted join rules signature checks for room versions 8/9.

### LINKED ISSUE #10923: synapse checks signature on `join_authorised_via_users_server` in `m.room.membership` events for room versions which do not specify it
As of room v8 and later, an extra auth rule is added which says that the event must be signed by the server given in `join_authorised_via_users_server`.  However, synapse also applies that to check to old room versions, and to membership events other than joins. Since spec-compliant implementations will *not* perform the check in those instances, this could lead to spilt-brained rooms.

clokep:
Fixes #10923

The clause was being applied for rooms which don't support this feature. My hope is this will have minimal impact since the additional field shouldn't exist for non- v8/v9 rooms.

### DIFF
--- a/changelog.d/10927.bugfix
+++ b/changelog.d/10927.bugfix
@@ -1,0 +1,1 @@
+Fix a bug introduced in Synapse v1.40.0 where the signature checks for room version 8/9 could be applied to earlier room versions in some situations.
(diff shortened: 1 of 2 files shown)


clokep:
#10920 seems related so I'd like to try to solve that at the same time.

richvdh (approved):
lgtm

Issue #10929, opened 2021-09-28
Room alias consumed when user blocked from publishing a newly-created room


AndrewFerr:
### Description

If Synapse is configured to allow a user to create local room aliases but not publish rooms, then whenever that user attempts to publish a new room with a custom local alias, the room creation will be blocked, as expected.

The problem is that although the room creation is blocked, **the creation of the alias is not**: the alias remains bound to the room ID that would have been assigned to the room, had its creation not been blocked.

This has the effect of the alias being unavailable for other rooms, with most clients not offering a way to free it (since the alias isn't bound to a valid room).

A workaround is to make a direct call to [`DELETE /_matrix/client/r0/directory/room/{roomAlias}`](https://matrix.org/docs/spec/client_server/latest#delete-matrix-client-r0-directory-room-roomalias).

### Steps to reproduce

- In Synapse, apply these config settings:
  - Set `alias_creation_rules` to `allow` for all users
  - Set `room_list_publication_rules` to `deny` for a specific user
- From any client, use the denied user to send [`/createRoom`](https://matrix.org/docs/spec/client_server/latest#post-matrix-client-r0-createroom) with `"visibility": "public"` and `"room_alias_name": <some_custom_alias>`. An easy way to do this is to create a new public room in Element Web, which enforces new public rooms to be published & to have aliases.
- Get a 403 response from Synapse. If using Element Web, get a message of "Server may be unavailable, overloaded, or you hit a bug."
- Try to create the same room again, but with a different user that has permission to publish rooms

<!--
Describe how what happens differs from what you expected.

If you can identify any relevant log snippets from _homeserver.log_, please include
those (please be careful to remove any personal or private data). Please surround them with
``` (three backticks, on a line on their own), so that they are formatted legibly.
-->
Attempts to use the same alias for the new room will fail, since the alias is still in use for the room that failed to be created. Element Web will say "This address is already in use" when typing the alias in the "Room address" section of the "Create a public room" dialog.

### Version information

<!-- IMPORTANT: please answer the following questions, to help us narrow down the problem -->

<!-- Was this issue identified on matrix.org or another homeserver? -->
- **Homeserver**: Testbench homeserver on localhost

<!--
 What version of Synapse is running?

You can find the Synapse version with this command:

$ curl http://localhost:8008/_synapse/admin/v1/server_version

(You may need to replace `localhost:8008` if Synapse is not configured to
listen on that port.)
-->
- **Version**: 1.43.0 (b=develop,d37841787)

[...]

PR #10930, opened 2021-09-28
Don't set new room alias before potential 403 (#10929)

AndrewFerr:
Signed-off-by: Andrew Ferrazzutti <fair@miscworks.net>

### Pull Request Checklist

### DIFF
--- a/synapse/handlers/room.py
+++ b/synapse/handlers/room.py
@@ -752,6 +752,15 @@ class RoomCreationHandler(BaseHandler):
         if not allowed_by_third_party_rules:
             raise SynapseError(403, "Room visibility value not allowed.")
 
+        if is_public:
+            if not self.config.roomdirectory.is_publishing_room_allowed(
+                user_id, room_id, room_alias
+            ):
+                # Lets just return a generic message, as there may be all sorts of
+                # reasons why we said no. TODO: Allow configurable error messages
+                # per alias creation rule?
+                raise SynapseError(403, "Not allowed to publish room")
+
         directory_handler = self.hs.get_directory_handler()
         if room_alias:
             await directory_handler.create_association(
@@ -761,15 +770,6 @@ class RoomCreationHandler(BaseHandler):
                 servers=[self.hs.hostname],
                 check_membership=False,
             )
-
-        if is_public:
-            if not self.config.roomdirectory.is_publishing_room_allowed(
-                user_id, room_id, room_alias
-            ):
-                # Lets just return a generic message, as there may be all sorts of
-                # reasons why we said no. TODO: Allow configurable error messages
-                # per alias creation rule?
-                raise SynapseError(403, "Not allowed to publish room")
 
         preset_config = config.get(
             "preset",
--- a/changelog.d/10929.bugfix
+++ b/changelog.d/10929.bugfix
@@ -1,0 +1,1 @@
+Newly-created public rooms are now only assigned an alias if the room's creation has not been blocked by permission settings. Contributed by @AndrewFerr.


synapse/handlers/room.py
```diff
@@ -752,6 +752,15 @@ async def create_room(
         if not allowed_by_third_party_rules:
             raise SynapseError(403, "Room visibility value not allowed.")
 
+        if is_public:
+            if not self.config.roomdirectory.is_publishing_room_allowed(
+                user_id, room_id, room_alias
+            ):
+                # Lets just return a generic message, as there may be all sorts of
```
DMRobertson, line 759: Not your typo, but while we're here:

```suggestion
                # Let's just return a generic message, as there may be all sorts of
```

DMRobertson (commented):
Thanks for spotting! Just to check I understand, what goes wrong without this change? I'm guessing we'll end up putting the alias in the DB anyway, so that

a) we have an alias to a room that doesn't exist
b) we can't reuse that alias in the future

It'd be great to have a test case that catches this---would you be up for writing one?

DMRobertson:
Ahh sorry, I just found #10929 which describes the problem in detail (thank you again).

> It'd be great to have a test case that catches this---would you be up for writing one?

Just wanted to add that I'd be more than happy to assist with this!

DMRobertson (changes requested):
Oh, minor thing: [the changelog.d entry number needs to match the PR number, not the issue number](https://github.com/matrix-org/synapse/pull/10930/checks?check_run_id=3729646198). With that fixed, CI should run the test suite.

tests/handlers/test_directory.py
```diff
@@ -432,6 +432,48 @@ def test_allowed(self):
         self.assertEquals(200, channel.code, channel.result)
 
 
+class TestCreatePublishedRoomACL(unittest.HomeserverTestCase):
+    user_id = "@test:test"
+    data_template = (
+        '{"room_alias_name": "%%23unofficial_test%%3Atest", "visibility": "%s"}'
+    )
```
DMRobertson, line 439: Does this need to be URL-encoded? I think this is the body of the http request and so should be fine to use `#` and `:` directly.

Having said that, the client-server spec [says](https://spec.matrix.org/unstable/client-server-api/#post_matrixclientr0createroom) that `room_alias_name` is the local part. So I think we just want `unofficial_test` here rather than `#unofficial-test:test`.

```suggestion
    data_template = '{"room_alias_name": "unofficial_test", "visibility": "%s"}'
```

tests/handlers/test_directory.py
```diff
@@ -432,6 +432,48 @@ def test_allowed(self):
...
+            rd_config.is_publishing_room_allowed
+        )
+
+        return hs
+
+    def test_denied(self):
+        channel = self.make_request(
+            "POST", "createRoom", (self.data_template % ("public",)).encode("ascii")
+        )
+        self.assertEquals(403, channel.code, channel.result)
```
DMRobertson, line 464: I don't think this test would fail before your changes. I think we'd need to add logic which checks that we haven't written to the DB. Let me see if I can find an example...

tests/handlers/test_directory.py
```diff
@@ -432,6 +432,48 @@ def test_allowed(self):
...
+
+    def test_allowed(self):
+        channel = self.make_request(
+            "POST", "createRoom", (self.data_template % ("private",)).encode("ascii")
+        )
+        self.assertEquals(200, channel.code, channel.result)
+
+    def test_denied_then_allowed(self):
+        self.test_denied()
+        self.test_allowed()
```
DMRobertson, line 474: Not sure I see why we'd want to repeat these tests in this specific order. What's your thinking here?

tests/handlers/test_directory.py
```diff
@@ -432,6 +432,48 @@ def test_allowed(self):
...
+
+    def test_denied(self):
+        channel = self.make_request(
+            "POST", "createRoom", (self.data_template % ("public",)).encode("ascii")
+        )
+        self.assertEquals(403, channel.code, channel.result)
+
+    def test_allowed(self):
+        channel = self.make_request(
+            "POST", "createRoom", (self.data_template % ("private",)).encode("ascii")
```
DMRobertson, line 468: I guess for private visibility Synapse will just ignore the `room_alias_name` in the request?

DMRobertson (changes requested):

tests/handlers/test_directory.py
DMRobertson: Here's an example where we select columns from a table with a where clause, and assert that we selected no rows:

https://github.com/matrix-org/synapse/blob/dc75fb7f0552e6a9903a7c173672c96610219ec0/tests/storage/databases/main/test_room.py#L57-L65

We'd have to work out what table is being written to and so on.

Alternatively, we could follow your description on #10929 and try to create a room with the same alias, but as a user with permission. That user should succeed in doing so.

> Try to create the same room again, but with a different user that has permission to publish rooms
> Attempts to use the same alias for the new room will fail, since the alias is still in use for the room that failed to be created

That would involve 

- `self.register_user` to make a new user
- `self.login` to generate an access token for that user
- passing `access_token=...` to `self.make_request` calls to act as that user.

Here's an example:

https://github.com/matrix-org/synapse/blob/e584534403b55ad3f250f92592e30b15b01f0201/tests/rest/admin/test_room.py#L1035-L1045



Sorry, that's quite an info dump. How does that feel---does it make sense/feel doable?

tests/handlers/test_directory.py
AndrewFerr: Yes, you're right, this is supposed to be just the localpart. That explains why the resolved alias was eventually getting an extra leading hash...!

tests/handlers/test_directory.py
AndrewFerr: If multiple room creation attempts are made in the same test function, the database persists the effects of the attempts, namely reserved aliases. To make it clearer, I tweaked & added some tests.

The `test_denied_then_retry` tests should succeed because their final creation attempt should be permitted. Without my code changes, they fail because of the requested alias having been consumed from the first (failed) attempt.

tests/handlers/test_directory.py
AndrewFerr: I added a test to create a published room as a denied, then allowed, user.

I managed to do it just by setting a custom `user_id` in calls to `self.helper.create_room_as`, while the test context is managed by an admin user. It seems to work fine, but please let me know if what I did is problematic.

tests/handlers/test_directory.py
```diff
@@ -432,6 +432,60 @@ def test_allowed(self):
...
+        rd_config.read_config(config)
+
+        self.hs.config.roomdirectory.is_publishing_room_allowed = (
+            rd_config.is_publishing_room_allowed
+        )
+
+        return hs
+
+    def test_denied(self):
+        # NOTE Setting is_public=True isn't enough
```
DMRobertson, line 461: That's odd---especially because `is_public` is `True` by default. I think the docstring is lying! I've raised #10951 for this.

tests/handlers/test_directory.py
```diff
@@ -432,6 +432,60 @@ def test_allowed(self):
...
+        self.data["visibility"] = "public"
+        self.helper.create_room_as(
+            self.denied_user_id, extra_content=self.data, expect_code=403
+        )
```
DMRobertson, line 465: This is going to mutate `self.data`, which might affect tests called afterwards. That means that tests can be sensitive to the order they're ran in, or might not run the same in isolation!

For this specific case, I think the other calls to `create_room_as` explicitly set `is_public=False` and so each test will behave the same no matter which order they're ran in. But I think we should try to avoid mutating `self.data` anyway, so the reader doesn't have to reason through this.

How about:

```suggestion
        data = self.data.copy()
        data["visibility"] = "public"
        self.helper.create_room_as(
            self.denied_user_id, extra_content=data, expect_code=403
        )
```

tests/handlers/test_directory.py
```diff
@@ -432,6 +432,60 @@ def test_allowed(self):
...
+            self.user_id, extra_content=self.data, is_public=False, expect_code=200
+        )
+
+    def test_denied_then_retry_without_publish(self):
+        self.test_denied()
+        self.test_allowed_without_publish()
+
+    def test_denied_then_retry_as_allowed(self):
+        self.test_denied()
+        self.test_allowed_as_allowed()
```
DMRobertson, line 486: Ahh I see---this will detect registering the alias in the DB before rejection.

I think that's worth a comment though. Maybe:

```suggestion
    def test_denied_then_retry_as_allowed(self):
        # Try to create a room with alias without permission.
        self.test_denied()
        # Repeat as someone with permission.
        # This used to fail because we registered the alias in the first request before it was denied.
        self.test_allowed_as_allowed()
```

tests/handlers/test_directory.py
```diff
@@ -432,6 +432,60 @@ def test_allowed(self):
         self.assertEquals(200, channel.code, channel.result)
 
 
+class TestCreatePublishedRoomACL(unittest.HomeserverTestCase):
+    user_id = "@admin:test"
```
DMRobertson, line 436: There's code which [treats admin users differently](https://github.com/matrix-org/synapse/blob/b10257e87972d158f4b6a0c7d1fe7239014ea10a/synapse/handlers/room.py#L644), but this isn't properly registering the user as an admin. I think change this to a different name, say `@test2:test`.

Also suggest calling this `accepted_user_id` instead of just `user_id`.

tests/handlers/test_directory.py
```diff
@@ -432,6 +432,60 @@ def test_allowed(self):
...
+        )
+
+    def test_allowed_as_allowed(self):
+        self.helper.create_room_as(
+            self.user_id, extra_content=self.data, is_public=False, expect_code=200
+        )
+
+    def test_denied_then_retry_without_publish(self):
+        self.test_denied()
+        self.test_allowed_without_publish()
```
DMRobertson, line 482: Comment here would be similarly helpful, something like this?

```suggestion
    def test_denied_then_retry_without_publish(self):
        # Try to create a public room with alias without permission. This should be rejected.
        self.test_denied()
        # Retry as the same user, but without publishing to the directory. This should be accepted.
        self.test_allowed_without_publish()
```

DMRobertson (changes requested):
Many thanks for this---I think it's almost there.

It's a bit of a nitpick, but I think the test method names could be a little more descriptive. Perhaps something like this?

- test_denied_without_publication_permission
- test_allowed_when_creating_private_room
- test_allowed_with_publication_permission
- test_can_create_as_private_room_after_rejection
- test_can_create_with_permission_after_rejection

tests/handlers/test_directory.py
richvdh: > This is going to mutate self.data, which might affect tests called afterwards. 

I'm not sure this is correct: I think there will be a new instance of `TestCreatePublishedRoomACL` for each test.

richvdh: ohhh, but each instance will end up pointing at the same `dict`, I guess, because it's a class-level property rather than an instance-level one. Maybe it'd be better to intialise `self.data` in `prepare`?

tests/handlers/test_directory.py
```diff
@@ -458,20 +458,32 @@ def prepare(self, reactor, clock, hs):
         return hs
 
     def test_denied(self):
```
clokep, line 460: Are these helper functions or separate tests (or both?) It looks like they're helpers, but by naming them `test_...` they'll get run as individual tests also.

A docstring might clarify this?

tests/handlers/test_directory.py
DMRobertson: > I'm not sure this is correct: I think there will be a new instance of TestCreatePublishedRoomACL for each test.

Looks like it! See https://gist.github.com/DMRobertson/40fe186843a327658cbee31bb381e1b2

richvdh:
@AndrewFerr unfortunately force-pushing makes it impossible to see what has changed since the previous review. Please could you reverse the force-push?

richvdh (dismissed):
cannot review after force-push

AndrewFerr:
> @AndrewFerr unfortunately force-pushing makes it impossible to see what has changed since the previous review. Please could you reverse the force-push?

Sorry about that. The force-push made 2 changes:
1) included #10963
2) rebased to the tip of `develop` at the time

Since this change must now eventually include #10963, may I wait for that to be merged before making any more changes here?

AndrewFerr:
What I pushed just now is what I believe was the state of this PR before the force-push.

Since #10964 is merged, I have it included in another local branch on the tip of `develop`, which I can push if that would be cleaner. Or should that go in a new PR altogether?

AndrewFerr:
And this latest push includes an extra change that I had made on top of the `create_room`/`is_public` change, which I forgot to mention in the list of changes I said my force-push had included.

richvdh:
I'm not entirely following you, but if you merge from `matrix-org:develop` into your local `develop`, that should update this PR to include any recent changes including #10963.

AndrewFerr:
Alright. Is it fine if I rebase instead of merge?

richvdh:
no, because that makes it hard for us to see what has changed during the rebase. Please merge.

AndrewFerr:
Done.

This merges what I had prior to the rebase that messed everything up, adds a change on top of #10963 (d97145203), and makes a fix to linter errors.

If this is still unusable, I'll make whatever changes are needed to fix it.

richvdh:
Thanks! It looks better now, though it looks like you've included some changes to `saml.py` which are causing a CI failure.

DMRobertson (approved):
LGTM! Thank you for seeing this through! I guess @richvdh might want to approve before merging?

PR #10931, opened 2021-09-28
Fix debian package builds.

erikjohnston:
This was due to dh-virtualenv builds being broken due to Sphinx removing deprecated APIs.

### DIFF
--- a/docker/Dockerfile-dhvirtualenv
+++ b/docker/Dockerfile-dhvirtualenv
@@ -47,8 +47,9 @@ RUN apt-get update -qq -o Acquire::Languages=none \
     && cd /dh-virtualenv \
     && env DEBIAN_FRONTEND=noninteractive mk-build-deps -ri -t "apt-get -y --no-install-recommends"
 
-# build it
-RUN cd /dh-virtualenv && dpkg-buildpackage -us -uc -b
+# Build it. Note that building the docs doesn't work due to differences in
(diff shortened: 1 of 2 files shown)


richvdh:
> Shpinx

schhhhpinx

docker/Dockerfile-dhvirtualenv
```diff
@@ -47,8 +47,9 @@ RUN apt-get update -qq -o Acquire::Languages=none \
     && cd /dh-virtualenv \
     && env DEBIAN_FRONTEND=noninteractive mk-build-deps -ri -t "apt-get -y --no-install-recommends"
 
-# build it
-RUN cd /dh-virtualenv && dpkg-buildpackage -us -uc -b
+# Build it. Note that building the docs doesn't work due to differences in
```
richvdh, line 50:
```suggestion
# Build it. Note that building the docs don't work due to differences in
```

richvdh (approved):
fine, but where is the updated sphinx coming from on bionic? https://packages.ubuntu.com/bionic/python3-sphinx says it hasn't changed since 2018.

erikjohnston:
> fine, but where is the updated sphinx coming from on bionic? https://packages.ubuntu.com/bionic/python3-sphinx says it hasn't changed since 2018.

I can't now seem to repro it failing on bionic, but on sid the package was updated on Saturday: python3-sphinx from 4.1 to 4.2

PR #10933, opened 2021-09-28
Strip "join_authorised_via_users_server" from join events which do not need it.

### LINKED ISSUE #10920: "Event not signed by authorising server" when trying to change nickname/avatar in a space-restricted room

clokep:
This fixes auth errors when changing a display name.

Fixes #10920. See complement tests at matrix-org/complement#208.

Note that the first commit is refactoring (using a constant) I can remove it if we do not think it is helpful...but I think it could be a nice jumping off point to find all the places which use that field.


synapse/handlers/room_member.py
```diff
@@ -908,6 +908,11 @@ async def _should_perform_remote_join(
                 Membership.JOIN,
                 Membership.INVITE,
             ):
+                # The event content should *not* include the authorising user as
+                # it won't be properly signed. Strip it out since it might come
+                # back from a client updating a display name / avatar.
+                content.pop(EventContentFields.AUTHORISING_USER, None)
```
clokep, line 914: I debated about universally stripping this field, but that seemed like not a great solution.

richvdh: This feels like an odd side-effect for this method. Did you consider doing it in `update_membership_locked` or somesuch?

clokep: I did consider doing it unconditionally near where we mess with the profile information. We don't have the room version that early / know if the user is in the room yet, so it would be a bit of a larger change.

Note that this method does already modify the `content` in some cases (by adding the `join_authorised_via_users_server` if it is needed), so it seemed somewhat reasonable that it also removes it in some cases?

I'd be OK changing that though if you think it is clearer!

richvdh: > Note that this method does already modify the `content` in some cases (by adding the `join_authorised_via_users_server` if it is needed), so it seemed somewhat reasonable that it also removes it in some cases?

Yeah, I did see that. It feels like adding it where it's needed is more expected than removing it where it's not, but ymmv.

I think stripping it universally would be fine.

I don't have a hugely strong opinion, but do think it would be clearer if you moved it out. Up to you though.

clokep: > I think stripping it universally would be fine.

That's fine with me.

richvdh (approved):
lgtm

clokep:
I cherry-picked this on 1.44 as 32072dcdac0072049832cda6204cd75be2d4e38f

PR #10934, opened 2021-09-28
Use `RoomVersion` objects

richvdh:
Various refactors to use `RoomVersion` objects instead of room version identifieers.

Should be reviewable commit-by-commit

### DIFF
--- a/synapse/events/builder.py
+++ b/synapse/events/builder.py
@@ -18,10 +18,8 @@ import attr
 from nacl.signing import SigningKey
 
 from synapse.api.constants import MAX_DEPTH
-from synapse.api.errors import UnsupportedRoomVersionError
 from synapse.api.room_versions import (
     KNOWN_EVENT_FORMAT_VERSIONS,
-    KNOWN_ROOM_VERSIONS,
     EventFormatVersions,
     RoomVersion,
 )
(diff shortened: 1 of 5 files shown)


synapse/handlers/federation.py
```diff
@@ -897,9 +897,9 @@ async def on_make_leave_request(
             )
             raise SynapseError(403, "User not from origin", Codes.FORBIDDEN)
 
-        room_version = await self.store.get_room_version_id(room_id)
-        builder = self.event_builder_factory.new(
-            room_version,
+        room_version_obj = await self.store.get_room_version(room_id)
```
clokep, line 900: I think we likely could have kept the variable names as `room_version`?

synapse/events/builder.py
```diff
@@ -197,24 +195,6 @@ def __init__(self, hs: "HomeServer"):
...
-        Returns:
-            EventBuilder
-        """
-        v = KNOWN_ROOM_VERSIONS.get(room_version)
-        if not v:
-            # this can happen if support is withdrawn for a room version
-            raise UnsupportedRoomVersionError()
-        return self.for_room_version(v, key_values)
-
     def for_room_version(
```
clokep, line 198: I wonder if we should rename this to `new(...)` now?

clokep (approved):
LGTM! The comments are both at your discretion!

synapse/handlers/federation.py
richvdh: yeah we could; however renaming them made it easier for me to make sure I was updating all the places they were referenced, and I think it does no harm to have renamed them.

synapse/events/builder.py
richvdh: I kinda like having the verbose name. It makes grepping more effective, for a start.

PR #10938, opened 2021-09-29
Fix MSC2716 `/batch_send` endpoint not being able to create outlier events

MadLittleMods:
Fix MSC2716 `/batch_send` endpoint not being able to create outlier events.

The `EventContext.for_outlier` refactor happened in https://github.com/matrix-org/synapse/pull/10883 and this spot was left out.

Part of https://github.com/matrix-org/matrix-doc/pull/2716

### Pull Request Checklist

### DIFF
--- a/changelog.d/10938.bugfix
+++ b/changelog.d/10938.bugfix
@@ -1,1 +1,1 @@
-Fix [MSC2716](https://github.com/matrix-org/matrix-doc/pull/2716) `/batch_send` endpoint not being able to create outlier events.
+Fix bug introduced in Synapse 1.44 which caused the experimental [MSC2716](https://github.com/matrix-org/matrix-doc/pull/2716) `/batch_send` endpoint to return a 500 error.
--- a/changelog.d/10938.bugfix
+++ b/changelog.d/10938.bugfix
@@ -1,0 +1,1 @@
+Fix [MSC2716](https://github.com/matrix-org/matrix-doc/pull/2716) `/batch_send` endpoint not being able to create outlier events.
--- a/synapse/handlers/message.py
+++ b/synapse/handlers/message.py
@@ -958,7 +958,7 @@ class EventCreationHandler:
         # Pass on the outlier property from the builder to the event
         # after it is created
         if builder.internal_metadata.outlier:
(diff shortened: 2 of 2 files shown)


synapse/handlers/message.py
```diff
@@ -953,18 +953,13 @@ async def create_new_client_event(
             depth=depth,
         )
 
-        old_state = None
+        context = await self.state.compute_event_context(event)
```
richvdh, line 956: computing event context isn't free (eg, it creates state groups in the database). We shouldn't do it if we're about to throw it away.

synapse/handlers/message.py
```diff
@@ -953,18 +953,13 @@ async def create_new_client_event(
...
 
         # Pass on the outlier property from the builder to the event
         # after it is created
         if builder.internal_metadata.outlier:
             event.internal_metadata.outlier = builder.internal_metadata.outlier
```
richvdh, line 961:
```suggestion
            event.internal_metadata.outlier = True
```

changelog.d/10938.bugfix
```diff
@@ -0,0 +1 @@
+Fix [MSC2716](https://github.com/matrix-org/matrix-doc/pull/2716) `/batch_send` endpoint not being able to create outlier events.
```
richvdh, line 1: we should say which version the bug was introduced in:

```suggestion
Fix bug introduced in Synapse 1.44 which caused the experimental [MSC2716](https://github.com/matrix-org/matrix-doc/pull/2716) `/batch_send` endpoint to return a 500 error.
```

(can we be more specific? "create outlier events" is too opaque for an admin - what causes it to generate outliers?)

richvdh (changes requested):
Sorry for missing this one, @MadLittleMods. I didn't realise we had code in the MessageHandler which generated outliers :/

MadLittleMods:
Thanks for the review pass @richvdh! It's hard to see when the MSC2716 tests don't run in CI yet 🙂

richvdh (approved):
thanks!

reivilibre:
Merged #10938 into release-v1.44.

MadLittleMods:
Thanks for the review @richvdh and for the merge @reivilibre 🦜

PR #10939, opened 2021-09-29
Fix logged errors from unit tests.

clokep:
This fixes some errors that appear in the logs during tests (mostly of the "cannot await a Mock" sort of error). This generally happens in background processes or get caught, but we can do better.

Generally if the unit tests pass this should be OK.

There's definitely more than what's fixed here, but this makes it better.

### DIFF
--- a/tests/appservice/test_scheduler.py
+++ b/tests/appservice/test_scheduler.py
@@ -24,7 +24,7 @@ from synapse.appservice.scheduler import (
 from synapse.logging.context import make_deferred_yieldable
 
 from tests import unittest
-from tests.test_utils import make_awaitable, simple_async_mock
+from tests.test_utils import simple_async_mock
 
 from ..utils import MockClock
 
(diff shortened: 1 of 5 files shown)


tests/appservice/test_scheduler.py
```diff
@@ -49,11 +49,10 @@ def test_single_service_up_txn_sent(self):
...
         # mock methods
-        self.store.get_appservice_state = Mock(
-            return_value=defer.succeed(ApplicationServiceState.UP)
-        )
-        txn.send = Mock(return_value=make_awaitable(True))
-        self.store.create_appservice_txn = Mock(return_value=defer.succeed(txn))
+        self.store.get_appservice_state = simple_async_mock(ApplicationServiceState.UP)
+        txn.send = simple_async_mock(True)
+        txn.complete = simple_async_mock(True)
+        self.store.create_appservice_txn = simple_async_mock(txn)
```
richvdh, line 55: is there any change in behaviour here, or is this just a cleaner way of doing it?

richvdh (approved):

tests/appservice/test_scheduler.py
clokep: There's one bugfix in here, which is also handling `txn.complete`, which previously ended up as just a `Mock` which can't be awaited. The rest is just refactoring to be a bit cleaner. (Sorry for not separating those!)

While I was in here I also converted `defer.succeed` since the mocked functions don't return `Deferreds` anymore. Note that we (annoyingly) have two ways of making async mocks at the moment (`simple_async_mock` and `make_awaitable`), essentially you can switch `defer.succeed` -> `make_awaitable` OR you can switch `Mock(return_value=defer.succeed(...))` -> `simple_async_mock(...)`. Sad times.

richvdh: got it. Thanks!

PR #10940, opened 2021-09-29
Split `event_auth.check` into two parts

richvdh:
Broadly, the existing `event_auth.check` function has two parts:
 * a validation section: checks that the event isn't too big, that it has the rught signatures, etc. This bit is independent of the rest of the state in the room, and so need only be done once for each event.
 * an auth section: ensures that the event is allowed, given the rest of the state in the room. This gets done multiple times, against various sets of room state, because it forms part of the state res algorithm.

Currently, this is implemented with `do_sig_check` and `do_size_check` parameters, but I think that makes everything hard to follow. Instead, we split the function in two and call each part separately where it is needed.

Reviewable commit-by-commit.


synapse/event_auth.py
```diff
@@ -64,28 +64,12 @@ def check(
...
-    # example.
-    for auth_event in auth_events.values():
-        if auth_event.room_id != room_id:
-            raise AuthError(
-                403,
-                "During auth for event %s in room %s, found event %s in the state "
-                "which is in room %s"
-                % (event.event_id, room_id, auth_event.event_id, auth_event.room_id),
-            )
-
```
richvdh: this (and the assertion above) are moved further down, into `check_auth_rules_for_event`. This could mean that we end up with a *different* `AuthError` for some bad events, but I don't think that matters: the important thing is that bad events are rejected and good events are accepted.

synapse/state/v1.py
```diff
@@ -329,12 +329,10 @@ def _resolve_auth_events(
         auth_events[(prev_event.type, prev_event.state_key)] = prev_event
         try:
             # The signatures have already been checked at this point
-            event_auth.check(
+            event_auth.check_auth_rules_for_event(
```
richvdh, line 332: this means we skip the check that the event has a `room_id` attribute, but there's no way we could have got this far with an event without one. Similarly in a bunch of places below.

tests/test_event_auth.py
```diff
@@ -37,21 +37,19 @@ def test_random_users_cannot_send_state_before_first_pl(self):
         }
 
         # creator should be able to send state
-        event_auth.check(
+        event_auth.check_auth_rules_for_event(
```
richvdh, line 40: since all these tests are just trying to check that the auth rules are checked, we may as well skip straight to that bit.

synapse/handlers/event_auth.py
```diff
@@ -49,19 +49,16 @@ def __init__(self, hs: "HomeServer"):
         self._store = hs.get_datastore()
         self._server_name = hs.hostname
 
-    async def check_from_context(
+    async def check_auth_rules_from_context(
         self,
-        room_version: str,
+        room_version_obj: RoomVersion,
         event: EventBase,
         context: EventContext,
```
richvdh, line 53: the eagle-eyed reviewer will notice that `context` is unused, so may wonder why I'm not getting rid of it, and why the function is still called `..._from_context`.

The answer is that I'm planning to add some extra tests based on the context, in the near future.

synapse/handlers/federation.py
```diff
@@ -746,7 +746,6 @@ async def on_make_join_request(
 
         # The remote hasn't signed it yet, obviously. We'll do the full checks
         # when we get the event back in `on_send_join_request`
-        validate_event_for_room_version(room_version, event, do_sig_check=False)
```
richvdh: this means we're not checking that the event has a `room_id` attribute, or the sizes of the event fields.

The former seems fine, given we've just built the event. The latter maybe less so, but it'll be checked multiple other times before it gets into the event DAG, so I'm going to go with it anyway.

clokep: > this means we're not checking that the event has a `room_id` attribute

The `room_id` check seems to have been moved into `check_auth_rules_for_event` so I think that will still get called?

clokep: I agree that skipping checking the size of the fields isn't a huge deal here since we're not persisting anything into the room until the `send_*` API call comes back.

clokep (approved):
Looks like a nice clean-up and definitely makes it clearer when we do and don't check signatures.

synapse/handlers/federation.py
richvdh: > The `room_id` check seems to have been moved into `check_auth_rules_for_event` so I think that will still get called?

I don't think so? I'm referring to this check: https://github.com/matrix-org/synapse/pull/10940/files#diff-2f0812c5ff333e8ee94dd38210fc77f98ccd0d8a995f9864d8c839e311b3f1a9R72-R73

clokep: Right, we do still pull out the `room_id` though so I'd expect it to error at least: https://github.com/matrix-org/synapse/pull/10940/files#diff-2f0812c5ff333e8ee94dd38210fc77f98ccd0d8a995f9864d8c839e311b3f1a9R149

Anyway, I think it is fine.

PR #10943, opened 2021-09-29
Annotate `log_function` decorator

reivilibre:
This one was concealing quite a few type errors because it didn't pass through the function signature.

### DIFF
--- a/synapse/federation/federation_client.py
+++ b/synapse/federation/federation_client.py
@@ -250,6 +250,8 @@ class FederationClient(FederationBase):
 
         logger.debug("backfill transaction_data=%r", transaction_data)
 
+        assert transaction_data is not None
+
         room_version = await self.store.get_room_version(room_id)
 
         pdus = [
--- a/synapse/federation/federation_server.py
+++ b/synapse/federation/federation_server.py
@@ -634,7 +634,7 @@ class FederationServer(FederationBase):
 
     async def on_make_knock_request(
         self, origin: str, room_id: str, user_id: str, supported_versions: List[str]
-    ) -> Dict[str, Union[EventBase, str]]:
+    ) -> Dict[str, Union[JsonDict, str]]:
         """We've received a /make_knock/ request, so we create a partial knock
         event for the room and hand that back, along with the room version, to the knocking
         homeserver. We do *not* persist or process this event until the other server has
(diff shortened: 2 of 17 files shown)


synapse/federation/federation_client.py
```diff
@@ -250,6 +250,8 @@ async def backfill(
```
squahtx, line 250: Neither here nor there but it bothers me that we use `__bool__` on `extremities` above, which is a generic `Iterable`. It happens in the other `backfill` methods too.

synapse/handlers/profile.py
```diff
@@ -455,7 +455,11 @@ async def _update_remote_profile_cache(self) -> None:
                 continue
 
             new_name = profile.get("displayname")
+            if new_name is not None and not isinstance(new_name, str):
+                new_name = None
             new_avatar = profile.get("avatar_url")
+            if new_avatar is not None and not isinstance(new_avatar, str):
+                new_avatar = None
```
clokep, line 462: I think all we need is the `isinstance` check here?

```suggestion
            if not isinstance(new_name, str):
                new_name = None
            new_avatar = profile.get("avatar_url")
            if not isinstance(new_avatar, str):
                new_avatar = None
```

synapse/federation/federation_server.py
```diff
@@ -632,7 +634,7 @@ async def on_send_leave_request(
 
     async def on_make_knock_request(
         self, origin: str, room_id: str, user_id: str, supported_versions: List[str]
-    ) -> Dict[str, Union[EventBase, str]]:
+    ) -> Dict[str, Union[JsonDict, str]]:
```
clokep, line 637: Maybe this is simpler?

```suggestion
    ) -> JsonDict:
```

synapse/federation/federation_client.py
```diff
@@ -250,6 +250,8 @@ async def backfill(
 
         logger.debug("backfill transaction_data=%r", transaction_data)
 
+        assert transaction_data is not None
```
clokep, line 253: The `backfill` method of the transport client suggests maybe we should never be returning `None` instead of this. 🤷 It looks like it isn't possible for it to return `None` though since we check `extremities` above. If nothing else, please add a comment about why this assertion is safe.

clokep (commented):
Seems good overall! 🎉

PR #10944, opened 2021-09-29
Fix issue causing sending presence to ASes to fail (due to incomplete type annotations)

### LINKED ISSUE #10941: Exceptions from presence handling in 1.44.0rc1
```
2021-09-29 16:09:28,131 - synapse.metrics.background_process_metrics - 215 - ERROR - notify_interested_services_ephemeral-0 - Background process 'notify_interested_services_ephemeral' threw an exception
  File "synapse/metrics/background_process_metrics.py", line 213, in run
    return await maybe_awaitable(func(*args, **kwargs))
  File "synapse/handlers/appservice.py", line 247, in _notify_interested_services_ephemeral
    events = await self._handle_presence(service, users)
  File "synapse/handlers/appservice.py", line 293, in _handle_presence
    presence_events, _ = await presence_source.get_new_events(
  File "synapse/logging/utils.py", line 69, in wrapped
    return f(*args, **kwargs)
TypeError: get_new_events() got an unexpected keyword argument 'service'
```

looks like something changed here in 4054dfa409fa17b45ab8f265813994956ed97bae

reivilibre:
Fixes #10941.

Cherry-picked out of https://github.com/matrix-org/synapse/pull/10943 (was commit fa5383214aab852c79192aa75bbe6b0343a579ce). Mypy won't actually be checking anything much on *this* PR because `@log_function` erases the signature of the functions it wraps, but in #10943 I have properly annotated it.

Maybe we should have tests for this ... (should I take the chance to do that now?).

### DIFF
--- a/synapse/handlers/account_data.py
+++ b/synapse/handlers/account_data.py
@@ -174,7 +174,6 @@ class AccountDataEventSource(EventSource[int, JsonDict]):
     async def get_new_events(
         self,
         user: UserID,
-        *,
         from_key: int,
         limit: Optional[int],
         room_ids: Collection[str],
--- a/synapse/handlers/account_data.py
+++ b/synapse/handlers/account_data.py
@@ -174,6 +174,7 @@ class AccountDataEventSource(EventSource[int, JsonDict]):
     async def get_new_events(
         self,
         user: UserID,
+        *,
         from_key: int,
         limit: Optional[int],
         room_ids: Collection[str],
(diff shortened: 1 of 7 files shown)


synapse/handlers/account_data.py
```diff
@@ -174,6 +174,7 @@ def get_current_key(self, direction: str = "f") -> int:
     async def get_new_events(
         self,
         user: UserID,
+        *,
```
clokep, line 177: Why make these keyword-only args?

reivilibre: To start with, I was trying to figure out why the signatures of the methods of the subclass were not compatible with the signature of the parent's method, then left them in because it seemed to match up with how these functions are called...

It looks like the `*` can be removed and no further complaints, so I'm happy to take them out.

clokep (approved):

PR #4886, opened 2019-03-19
fix test_auto_create_auto_join_where_no_consent

neilisfragile:

PR #10945, opened 2021-09-29
Clean-up registration tests

clokep:
This has a couple non-functional changes and then attempts to have the `test_auto_create_auto_join_where_no_consent` test what the comments say.

* Mocks out some federation logic.
* Uses `override_config` in places.
* Update the config in `test_auto_create_auto_join_where_no_consent`  to have auto-join rooms, which then breaks the test. (The test wasn't setting `auto_join_rooms` properly so it wasn't attempting to join rooms ever.)

Open questions:

* There's a race in the consent logic that if another user registers before the first user registers & completes consent than the auto-join rooms will never be created. I wonder if it would be more robust to attempt to join the room if it exists, otherwise create it for every user?

See #4886 which tried to fix this previously.

### DIFF
--- a/tests/handlers/test_register.py
+++ b/tests/handlers/test_register.py
@@ -521,9 +521,13 @@ class RegistrationTestCase(unittest.HomeserverTestCase):
 
     @override_config(
         {
-            "user_consent": {"block_events_error": "Error"},
+            "user_consent": {
+                "block_events_error": "Error",
+                "require_at_registration": True,
+            },
             "form_secret": "53cr3t",
             "public_baseurl": "http://test",
+            "auto_join_rooms": ["#room:test"],
         },
     )
     def test_auto_create_auto_join_where_no_consent(self):
@@ -536,18 +540,20 @@ class RegistrationTestCase(unittest.HomeserverTestCase):
         #    * they have not given that consent
         #    * The server is configured to auto-join to a room
         # (and autocreate if necessary)
-        room_alias_str = "#room:test"
-        self.hs.config.auto_join_rooms = [room_alias_str]
 
         # When:-
-        #   * the user is registered and post consent actions are called
+        #   * the user is registered
         user_id = self.get_success(self.handler.register_user(localpart="jeff"))
-        self.get_success(self.handler.post_consent_actions(user_id))
 
         # Then:-
         #   * Ensure that they have not been joined to the room
         rooms = self.get_success(self.store.get_rooms_for_user(user_id))
         self.assertEqual(len(rooms), 0)
+
+        # The user provides consent; ensure they are now in the rooms.
+        self.get_success(self.handler.post_consent_actions(user_id))
+        rooms = self.get_success(self.store.get_rooms_for_user(user_id))
+        self.assertEqual(len(rooms), 1)
 
     def test_register_support_user(self):
         user_id = self.get_success(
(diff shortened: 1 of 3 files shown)


tests/handlers/test_register.py
```diff
@@ -521,25 +538,20 @@ def test_auto_create_auto_join_where_no_consent(self):
...
-        event_creation_handler._block_events_without_consent_error = "Error"
-        event_creation_handler._consent_uri_builder = Mock()
-        room_alias_str = "#room:test"
-        self.hs.config.auto_join_rooms = [room_alias_str]
-
         # When:-
-        #   * the user is registered and post consent actions are called
+        #   * the user is registered
         user_id = self.get_success(self.handler.register_user(localpart="jeff"))
-        self.get_success(self.handler.post_consent_actions(user_id))
```
clokep, line 536: I *think* this change makes sense! If they have consented, then why would they not be joined to the room? It is a bit hard to tell from the logic what is supposed to happen in this case though.

tests/handlers/test_register.py
```diff
@@ -510,6 +519,17 @@ def test_auto_create_auto_join_room_preset_invalid_permissions(self):
         self.assertEqual(rooms, set())
         self.assertEqual(invited_rooms, [])
 
+    @override_config(
+        {
+            "user_consent": {
+                "block_events_error": "Error",
+                "require_at_registration": True,
```
clokep, line 526: This somewhat changes what we're testing by requiring it at registration, which is not the default. 😢

tests/handlers/test_register.py
```diff
@@ -127,7 +132,15 @@ def make_homeserver(self, reactor, clock):
         hs_config["max_mau_value"] = 50
         hs_config["limit_usage_by_mau"] = True
 
-        hs = self.setup_test_homeserver(config=hs_config)
+        # Don't attempt to reach out over federation.
+        self.mock_federation_client = Mock()
+        self.mock_federation_client.make_query.side_effect = CodeMessageException(
+            500, ""
+        )
```
richvdh, line 141: this will *return* an exception object, rather than raise it? Is that correct?

tests/handlers/test_register.py
richvdh: :man_shrugging:

Any idea how this was passing before?

tests/handlers/test_register.py
clokep: Setting `side_effect` to an exception class or instance raises it.

From [the docs](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.side_effect):

> This can either be a function to be called when the mock is called, an iterable or an exception (class or instance) to be raised.
>
> ...
>
> An example of a mock that raises an exception (to test exception handling of an API):
>
> ```pycon
> >>> mock = Mock()
> >>> mock.side_effect = Exception('Boom!')
> >>> mock()
> Traceback (most recent call last):
>   ...
> Exception: Boom!
> ```

The `side_effect` property has a lot of confusing special behavior...

richvdh: ugh.

tests/handlers/test_register.py
richvdh: lovely.

richvdh (approved):

PR #10947, opened 2021-09-29
Add functionality to remove deactivated users from the monthly_active_users table

### LINKED ISSUE #10861: Deactivated users still count towards the MAU count
Which feels like something that at the very least should be configurable, if not counted at all. An example is you can fill a host with 10 users, set a limit of 10 MAU, and if you deactivate your users you will still be affected by the limit.

H-Shay:
Fixes #10861. Prior to this change, deactivated users were not being removed from the monthly_active_users table, thus continuing to count towards the mau_limit.


synapse/storage/databases/main/monthly_active_users.py
```diff
@@ -354,3 +354,17 @@ async def populate_monthly_active_users(self, user_id):
                         await self.upsert_monthly_active_user(user_id)
             elif now - last_seen_timestamp > LAST_SEEN_GRANULARITY:
                 await self.upsert_monthly_active_user(user_id)
+
+    async def remove_deactivated_user(self, user_id: str) -> None:
```
DMRobertson, line 358: Sorry, should have said this yesterday: I think this name should make it clear we're only deleting from the `monthly_active_users` table and not e.g. from the `users` table.

synapse/storage/databases/main/monthly_active_users.py
```diff
@@ -354,3 +354,17 @@ async def populate_monthly_active_users(self, user_id):
...
+        table and resets the "get_monthly_active_count" cache.
+
+        Args:
+            user_id(str): the user_id to remove
+        """
+
+        await self.db_pool.simple_delete_one(
+            table="monthly_active_users", keyvalues={"user_id": user_id}
+        )
+        await self.invalidate_cache_and_stream("get_monthly_active_count", ())
```
DMRobertson, line 370: Ahh, didn't spot that there was a version without the leading underscore (that doesn't take a `txn`). Nice!

synapse/storage/databases/main/monthly_active_users.py
```diff
@@ -354,3 +354,17 @@ async def populate_monthly_active_users(self, user_id):
...
+        Removes a deactivated user from the monthly active user
+        table and resets the "get_monthly_active_count" cache.
+
+        Args:
+            user_id(str): the user_id to remove
+        """
+
+        await self.db_pool.simple_delete_one(
+            table="monthly_active_users", keyvalues={"user_id": user_id}
+        )
```
richvdh, line 369: suggestion: we could avoid the need to call `user_last_seen_monthly_active` in `deactivate_account` by replacing this with `simple_delete`. (you should probably then only do the `invalidate_cache_and_stream` if it returns non-zero.)

synapse/storage/databases/main/monthly_active_users.py
richvdh, line 370: I think we also need to invalidate `get_monthly_active_count_by_service` and `user_last_seen_monthly_active` ?

tests/test_mau.py
```diff
@@ -229,6 +235,32 @@ def test_tracked_but_not_limited(self):
...
+        self.do_sync_for_user(token2)
+
+        # Check that mau count is what we expect
+        count = self.get_success(self.store.get_monthly_active_count())
+        self.assertEqual(count, 2)
+
+        # Deactivate user1
+        url = "/_synapse/admin/v1/deactivate/%s" % user1
+        channel = self.make_request("POST", url, access_token=admin_token)
+        print(channel.json_body)
```
richvdh, line 257: this looks like some debugging that has got left behind.

richvdh (commented):
looks great, but a couple of suggestions.

synapse/storage/databases/main/monthly_active_users.py
```diff
@@ -355,16 +355,26 @@ async def populate_monthly_active_users(self, user_id):
...
+        rows_deleted = await self.db_pool.simple_delete(
+            table="monthly_active_users",
+            keyvalues={"user_id": user_id},
+            desc="simple_delete",
         )
-        await self.invalidate_cache_and_stream("get_monthly_active_count", ())
+
+        if rows_deleted != 0:
+            await self.invalidate_cache_and_stream(
+                "user_last_seen_monthly_active", (user_id)
```
richvdh, line 375: ah, you've fallen into one of python's traps here. `(foo)` *looks* like a tuple, but is actually the same as `foo`. You need a trailing comma to force a tuple:

```suggestion
                "user_last_seen_monthly_active", (user_id,)
```

richvdh (approved):
lgtm other than the below

H-Shay:
Merged.

H-Shay:
Thanks for the review and suggestions, @richvdh!

Issue #10764, opened 2021-09-04
Backfill should gracefully handle the case where no one has the complete state for a given event


MadLittleMods:
*This issue is spawning from https://github.com/matrix-org/synapse/pull/10566#discussion_r700713381 and summarizes the problem/conversation there. Thanks @erikjohnston and @richvdh for the insights on this!*

---

We initially used a `fake_prev_event_id` for MSC2716 to make a chain of events float outside of the normal DAG but have since [opted to use empty `prev_events=[]`](https://github.com/matrix-org/synapse/pull/11114). This issue still valid for the general case where no servers have the state for a given event though.

The refactor in https://github.com/matrix-org/synapse/pull/10645 changed the way we handle fetching state for backfills. Before we'd just get the state at the last event, whereas now we try and get the state for each `prev_event` (including the fake event), which obviously fails. This means that federated homeservers no longer accept the insertion event because it can't fetch the `fake_prev_event_id` state (`Error attempting to resolve state at missing prev_events`, `ERROR 403: We can't get valid state history.`). 

The goal of the code that's rejecting the fake event is protecting against an attack where a remote server could arbitrarily replace the servers view of the current state of the room. **But the mitigation we use is a sledgehammer and we should replace it with something more intelligent that both mitigates the attack and gracefully handles missing events.**

## Potential solutions 

To be determined (TBD).

## Reproduction case

 1. Create an event which references a made up `prev_event_id`
 1. Try to backfill it from federated homeserver

```sh
2021-09-01 20:19:21,560 - synapse.handlers.federation_event - 746 - WARNING - GET-5-$hbHXgulNKlmg9mkGQiBH5SSgq2NL4GO70sIIfHCa-YY - Error attempting to resolve state at missing prev_events
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/synapse/handlers/federation_event.py", line 711, in _resolve_state_at_missing_prevs
    remote_state = await self._get_state_after_missing_prev_event(
  File "/usr/local/lib/python3.8/site-packages/synapse/handlers/federation_event.py", line 777, in _get_state_after_missing_prev_event
    ) = await self.federation_client.get_room_state_ids(
  File "/usr/local/lib/python3.8/site-packages/synapse/federation/federation_client.py", line 370, in get_room_state_ids
    result = await self.transport_layer.get_room_state_ids(
[...]

PR #10948, opened 2021-09-29
Draft: Add proper state and state_groups to historical events so they return state from `/context` (avatar/displayname) (MSC2716)

MadLittleMods:
Add proper state and state_groups to historical events so they return state from `/context` and we see the appropriate avatar/displayname in Element.

See https://gitlab.com/gitterHQ/webapp/-/merge_requests/2229#note_683532091

Also seems to fix https://github.com/matrix-org/synapse/issues/10764

### Pull Request Checklist

### DIFF
--- a/synapse/handlers/message.py
+++ b/synapse/handlers/message.py
@@ -959,7 +959,15 @@ class EventCreationHandler:
             event.internal_metadata.outlier = True
             context = EventContext.for_outlier()
         else:
-            context = await self.state.compute_event_context(event)
+            old_state = None
+            # Define the state for historical messages while we know to get all of
+            # state_groups setup properly when we `compute_event_context`.
+            if builder.internal_metadata.is_historical() and auth_event_ids:
+                old_state = await self.store.get_events_as_list(auth_event_ids)
+
+            context = await self.state.compute_event_context(event, old_state=old_state)
+
+        logger.info("create_new_client_event type=%s, event_id=%s context=%s", event.type, event.event_id, context)
 
         if requester:
             context.app_service = requester.app_service
--- a/synapse/rest/client/room_batch.py
+++ b/synapse/rest/client/room_batch.py
@@ -176,6 +176,8 @@ class RoomBatchSendEventRestServlet(RestServlet):
     async def on_POST(
         self, request: SynapseRequest, room_id: str
     ) -> Tuple[int, JsonDict]:
+        logger.info("room batch send =====================================================")
+        logger.info("=====================================================================")
         requester = await self.auth.get_user_by_req(request, allow_guest=False)
 
         if not requester.app_service:


synapse/handlers/room.py
```diff
@@ -1167,12 +1167,13 @@ async def filter_evts(events: List[EventBase]) -> List[EventBase]:
         # XXX: why do we return the state as of the last event rather than the
         # first? Shouldn't we be consistent with /sync?
         # https://github.com/matrix-org/matrix-doc/issues/687
+        event_id_to_get_state_from = last_event_id
 
         state = await self.state_store.get_state_for_events(
-            [last_event_id], state_filter=state_filter
```
MadLittleMods, line 1172: For `/context`, it only gives you the state from the last event in `events_after` instead of the `event_id` itself. This means that often times, you end up missing the historical floating state for the historical message and instead get some of the real resolved state which doesn't have any of the member events we want.

I was messing around with changing this to return the state for the `event_id` specified but using the last event is part of the spec.

I could change the which events are fetched in `events_before`/`events_after` when specifying a historical event so it only returns itself 🤷. Then it would be the last event of itself.

synapse/handlers/message.py
```diff
@@ -969,7 +972,13 @@ async def create_new_client_event(
...
             context = EventContext.for_outlier()
         else:
-            context = await self.state.compute_event_context(event)
+            old_state = None
+            # Define the state for historical messages while we know to get all of
+            # state_groups setup properly when we `compute_event_context`.
+            if builder.internal_metadata.is_historical() and full_state_ids_at_event:
+                old_state = await self.store.get_events_as_list(full_state_ids_at_event)
+
+            context = await self.state.compute_event_context(event, old_state=old_state)
```
MadLittleMods, line 981: When you pull state from a `state_group`, it grabs the full resolved state from the all of the `state_group` deltas.

So we want to define the state for the historical event as all of the state at that point in time, including member events not matching the `sender` of the event. This way when `/messages` fetches the state for the first event in the batch, it most likely grabs from some historical event with all of the member events necessary for that batch.

This does have the flaw where if you just insert a single event somewhere, it probably won't resolve the state correctly from `/messages` or `/context` since it will grab a non historical event above or below.

Might be good to look at `/mesages` and `/context` to additional add the `state` for any historical messages in that batch.

synapse/handlers/message.py
```diff
@@ -939,6 +916,32 @@ async def create_new_client_event(
...
+            # Create a StateMap[str]
+            auth_event_state_map = {
+                (e.type, e.state_key): e.event_id for e in auth_events
+            }
+            # Actually strip down and use the necessary auth events
+            auth_event_ids = self._event_auth_handler.compute_auth_events(
+                event=temp_event,
+                current_state_ids=auth_event_state_map,
+                for_verification=False,
+            )
```
MadLittleMods, line 943: Moved this auth state stripping into `create_new_client_event` so we can copy `full_state_ids_at_event` before we strip and use it for the `old_state` for `state_group` calculation

synapse/handlers/message.py
MadLittleMods: Sharing `state_groups` and resolving state in v2, https://github.com/matrix-org/synapse/pull/10975

MadLittleMods:
Closing in favor of https://github.com/matrix-org/synapse/pull/10975

PR #10956, opened 2021-09-30
Ensure that we reject events which use rejected events for auth

richvdh:
When we consider whether to accept events, we should not accept those which depend on rejected events for their auth events.

This (together with earlier changes such as https://github.com/matrix-org/synapse/pull/10771 and https://github.com/matrix-org/synapse/pull/10896) forms a partial fix to https://github.com/matrix-org/synapse/issues/9595. There still remain code paths where we do not check the `auth_events` at all.

Complement doesn't work with versions of synapse before 1.8.0 (it relies on send_join v2), but this bug seems to have existed since at least then.

### DIFF
--- a/changelog.d/10956.bugfix
+++ b/changelog.d/10956.bugfix
@@ -1,0 +1,1 @@
+Fix a long-standing bug which meant that events received over federation were sometimes incorrectly accepted into the room state.
--- a/synapse/event_auth.py
+++ b/synapse/event_auth.py
@@ -155,6 +155,12 @@ def check_auth_rules_for_event(
                 "which is in room %s"
                 % (event.event_id, room_id, auth_event.event_id, auth_event.room_id),
             )
+        if auth_event.rejected_reason:
+            raise AuthError(
+                403,
+                "During auth for event %s: found rejected event %s in the state"
+                % (event.event_id, auth_event.event_id),
+            )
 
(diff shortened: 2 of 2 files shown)


DMRobertson (approved):
Seems reasonable and complement is happy: matrix-org/complement#210

Sanity check: once an event is rejected, is there anything that could happen to make us later accept it,

richvdh:
> Sanity check: once an event is rejected, is there anything that could happen to make us later accept it,

no. Rejecting an event is an indication that the event is fundamentally broken and can never be accepted.

erikjohnston:
I'm not sure this is quite right. We now have two situations where we reject events depending on whether we were checking the auth based on 1) the event's auth events or 2) the state at the event. I think this check should only apply to rejections due to the former and not the latter, as:

1. In a handwavey way: the check against auth events is meant to be a check if the event could have been valid at some point, without having to calculate the state of the full room graph. This PR is sort of conflating "is this event valid based on its auth chain" and if events should be rejected based on state. 
2. For events that used to be outliers we may do the second check long after the first (e.g when we backfill), at which point we'll go from accepting events that use that event as an auth event to rejecting events that reference that event.

I see two approaches to handling this:
1. Drop events that don't pass auth based on their auth events, i.e. basically treat them in the same way as events which have missing auth events
2. Add a new `rejected_reason` string to differentiate between the two cases.

PR #10958, opened 2021-09-30
Add type hints to filtering classes.

clokep:
I was trying to figure out how these worked so I added type hints.


squahtx (approved):
lgtm

PR #10959, opened 2021-09-30
Use direct references for configuration variables (part 7).

clokep:
Builds on #10798, #10812, #10885, #10893, #10897, and #10916.

Each commit was done with `sed` + `black`, then running the linters and fixing some type hints. Sorry that this one got a bit big.

~~Builds on #10945.~~

There's a few remaining instances after this that I had missed in previous PRs (or got added after this was all done?) I'll fix those in the final PR.

### DIFF
--- a/changelog.d/10959.misc
+++ b/changelog.d/10959.misc
@@ -1,0 +1,1 @@
+Use direct references to config flags.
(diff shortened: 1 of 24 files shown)


squahtx (approved):
lgtm!

PR #10960, opened 2021-09-30
Consistently exclude from user_directory, round 2

DMRobertson:
Spun off from #10914, alongside #10891. Please review commit-by-commit.

Most of the effort here went into writing test cases. There's some duplication between testing the incremental updates and testing the complete rebuild.

In the rebuild functions I filter `users_with_profile` to exclude users early (rather than at the point of potentially putting them in the DB). That's because there was a code path that missed this check even before my changes. 

I think we could probably do a similar thing in the incremental update functions, but I didn't do that here to keep the changeset minimal.

### DIFF
--- a/synapse/storage/databases/main/user_directory.py
+++ b/synapse/storage/databases/main/user_directory.py
@@ -40,12 +40,10 @@ from synapse.util.caches.descriptors import cached
 
 logger = logging.getLogger(__name__)
 
-
 TEMP_TABLE = "_temp_populate_user_directory"
 
 
 class UserDirectoryBackgroundUpdateStore(StateDeltasStore):
-
     # How many records do we calculate before sending it to
     # add_users_who_share_private_rooms?
     SHARE_PRIVATE_WORKING_SET = 500
@@ -235,13 +233,16 @@ class UserDirectoryBackgroundUpdateStore(StateDeltasStore):
                 )
 
                 users_with_profile = await self.get_users_in_room_with_profiles(room_id)
+                # Throw away users excluded from the directory.
+                users_with_profile = {
+                    user_id: profile
+                    for user_id, profile in users_with_profile.items()
+                    if not self.hs.is_mine_id(user_id)
+                    or await self.should_include_local_user_in_dir(user_id)
+                }
 
                 # Update each user in the user directory.
                 for user_id, profile in users_with_profile.items():
-                    if self.hs.is_mine_id(
-                        user_id
-                    ) and not await self.should_include_local_user_in_dir(user_id):
-                        continue
                     await self.update_profile_in_user_dir(
                         user_id, profile.display_name, profile.avatar_url
                     )
@@ -250,11 +251,6 @@ class UserDirectoryBackgroundUpdateStore(StateDeltasStore):
 
                 if is_public:
                     for user_id in users_with_profile:
-                        if self.hs.is_mine_id(
-                            user_id
-                        ) and not await self.should_include_local_user_in_dir(user_id):
-                            continue
-
                         to_insert.add(user_id)
 
                     if to_insert:
@@ -262,10 +258,10 @@ class UserDirectoryBackgroundUpdateStore(StateDeltasStore):
                         to_insert.clear()
                 else:
                     for user_id in users_with_profile:
+                        # We want the set of pairs (L, M) where L and M are
+                        # in `users_with_profile` and L is local.
+                        # Do so by looking for the local user L first.
                         if not self.hs.is_mine_id(user_id):
-                            continue
-
-                        if not await self.should_include_local_user_in_dir(user_id):
                             continue
 
                         for other_user_id in users_with_profile:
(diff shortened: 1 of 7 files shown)


richvdh:
> (Does deactivating a user remove them from any rooms they're in?)

Yes, though it's asynchronous. (see `_start_user_parting`). You might be able to stub it out for the tests? Or generate a rejoin event for them?

tests/unittest.py
```diff
@@ -596,6 +597,25 @@ def register_user(
...
         return user_id
 
+    def register_appservice_user(
+        self,
+        username: str,
+        appservice_token: str,
+    ) -> str:
+        """Register an appservice user. Requires the client-facing registration API be registered.
+
+        Returns the MXID of the new user."""
```
richvdh, line 607: docs on what the params are would be nice.

tests/unittest.py
```diff
@@ -596,6 +597,25 @@ def register_user(
...
+        urlencoded_token = quote(appservice_token)
```
richvdh, line 608: probably preferable to use the `access_token` param on `make_request`

tests/handlers/test_user_directory.py
```diff
@@ -47,13 +48,29 @@ class UserDirectoryTestCase(unittest.HomeserverTestCase):
...
+        self.appservice = ApplicationService(
+            token="i_am_an_app_service",
+            hostname="test",
+            id="1234",
+            namespaces={"users": [{"regex": r"@as_user.*", "exclusive": True}]},
+            sender="@as:test",
+        )
+
+        mock_load_appservices = Mock(return_value=[self.appservice])
+        with patch(
```
richvdh, line 68: It wouldn't be *crazy* to allow app service configs to be specified inline in the homeserver config, rather than forcing them to be an external file, as currently. See, for example, the `signing_key`/`signing_key_path` settings: the key can be specified inline rather than in a separate file.

That obviously makes testing easier.

synapse/storage/databases/main/user_directory.py
```diff
@@ -259,7 +265,7 @@ def _get_next_batch(
                         if not self.hs.is_mine_id(user_id):
                             continue
 
-                        if self.get_if_app_services_interested_in_user(user_id):
+                        if not await self.should_include_local_user_in_dir(user_id):
                             continue
 
                         for other_user_id in users_with_profile:
```
DMRobertson, line 271: There's a hard-to-spot oversight right below here on -269.

We don't check to see if `other_user_id` ought to be excluded from the directory. We'll gleefully add them to `users_who_share_private_rooms`. This isn't a user-visible problem---if all of the logic elsewhere is correct, then `other_user_id` won't be in the `user_directory` table and so won't show up in searches. But it's a frustrating inconsistency.

Instead I'm going to filter out the excluded users early on, right before we assign a value to `users_with_profile`.

synapse/storage/databases/main/user_directory.py
```diff
@@ -40,12 +40,10 @@
 
 logger = logging.getLogger(__name__)
 
-
```
DMRobertson, line 43: I'm not sure how these crept in. Possibly I accidentally formatted in pycharm (ctrl-alt-l) and `black` was happy with the change?

DMRobertson:
@richvdh I think this is in a complete place now. Hopefully should still be reviewable commit-by-commit. Some, but not all of the tests introduced here fail without the corresponding code changes so I'm not sure this one naturally breaks up further. Still, if it's hard to follow I'm happy to try and pull out pieces.

(In the ideal world I'd interactively rebase to squash c090ab6 with 010314d, but I don't want to screw up anything on the reviewers' end by doing so)

tests/storage/test_user_directory.py
```diff
@@ -237,34 +238,77 @@ def test_initial(self) -> None:
         users = self.get_success(self.user_dir_helper.get_users_in_user_directory())
         self.assertEqual(users, {u1, u2, u3})
 
+    # The next three tests (test_population_excludes_*) all setup
```
richvdh, line 241:
```suggestion
    # The next three tests (test_population_excludes_*) all set up
```

tests/handlers/test_user_directory.py
```diff
@@ -62,6 +81,137 @@ def prepare(self, reactor: MemoryReactor, clock: Clock, hs: HomeServer) -> None:
...
+        self.assertEqual(users, {alice, bob})
+        self.assertEqual(
+            set(in_public), {(alice, public), (bob, public), (alice, public2)}
+        )
+        self.assertEqual(
+            self.user_dir_helper._compress_shared(in_private),
+            {(alice, bob, private), (bob, alice, private)},
+        )
+
+    # The next three tests (test_population_excludes_*) all setup
```
richvdh, line 125:
```suggestion
    # The next three tests (test_population_excludes_*) all set up
```

richvdh (approved):
lgtm!

PR #10962, opened 2021-10-01
Fix logic flaw preventing tracking of MSC2716 events in existing room versions

MadLittleMods:
Fix logic flaw preventing tracking of MSC2716 events in existing room versions

We correctly allowed using the MSC2716 batch endpoint for
the room creator in existing room versions but accidentally didn't track
the events because of a logic flaw.

This prevented you from connecting subsequent batches together because it would
throw the unknown batch ID error.

Complement tests: https://github.com/matrix-org/complement/pull/209

### Pull Request Checklist

### DIFF
--- a/changelog.d/10962.bugfix
+++ b/changelog.d/10962.bugfix
@@ -1,0 +1,1 @@
+Fix [MSC2716](https://github.com/matrix-org/matrix-doc/pull/2716) `/batch_send` endpoint rejecting subsequent batches with unknown batch ID error in existing room versions from the room creator.
--- a/synapse/handlers/federation_event.py
+++ b/synapse/handlers/federation_event.py
@@ -1010,9 +1010,8 @@ class FederationEventHandler:
         room_version = await self._store.get_room_version(marker_event.room_id)
         create_event = await self._store.get_create_event_for_room(marker_event.room_id)
         room_creator = create_event.content.get(EventContentFields.ROOM_CREATOR)
-        if (
-            not room_version.msc2716_historical
-            or not self._config.experimental.msc2716_enabled
+        if not room_version.msc2716_historical and (
+            not self._config.experimental.msc2716_enabled
             or marker_event.sender != room_creator
         ):
             return
--- a/synapse/storage/databases/main/events.py
+++ b/synapse/storage/databases/main/events.py
@@ -1763,9 +1763,8 @@ class PersistEventsStore:
             retcol="creator",
             allow_none=True,
         )
-        if (
-            not room_version.msc2716_historical
-            or not self.hs.config.experimental.msc2716_enabled
+        if not room_version.msc2716_historical and (
+            not self.hs.config.experimental.msc2716_enabled
             or event.sender != room_creator
         ):
             return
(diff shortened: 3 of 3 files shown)


changelog.d/10962.bugfix
```diff
@@ -0,0 +1 @@
+Fix [MSC2716](https://github.com/matrix-org/matrix-doc/pull/2716) `/batch_send` endpoint rejecting subsequent batches with unknown batch ID error in existing room versions from the room creator.
```
MadLittleMods, line 1: Better way to explain?

In existing room versions, the room creator can use the `/batch_send` endpoint. But previously because of the logic flaw, we didn't add the insertion event batch ID's to the database table so when you tried to reference that batch in another chunk, we couldn't find the insertion event and it would throw a `400` `No insertion event corresponds to the given ?batch_id` error.

synapse/handlers/federation_event.py
```diff
@@ -1010,9 +1010,8 @@ async def _handle_marker_event(self, origin: str, marker_event: EventBase) -> No
         room_version = await self._store.get_room_version(marker_event.room_id)
         create_event = await self._store.get_create_event_for_room(marker_event.room_id)
         room_creator = create_event.content.get(EventContentFields.ROOM_CREATOR)
-        if (
-            not room_version.msc2716_historical
-            or not self._config.experimental.msc2716_enabled
+        if not room_version.msc2716_historical and (
+            not self._config.experimental.msc2716_enabled
             or marker_event.sender != room_creator
         ):
```
MadLittleMods, line 1016: We only want to process MSC2716 events when:

 - The room version supports MSC2716
 - Any room version with the `msc2716_enabled` experimental feature enabled and the event is from the room creator

clokep: I'm a bit confused (and haven't been following this work too closely -- so sorry if this is a dumb question), but what's the relationship between the room version supporting MSC2716 vs. having MSC2716 enabled in the config? I would think the latter implied the former, but it sounds like they're completely different?

I don't see how a room version can have `msc2716_enabled` since it is a configuration flag?

MadLittleMods: @clokep Since we're considering MSC2716 pretty backwards compatible, we're allowing MSC2716 things to happen in current room versions but the homeserver still needs the `experimental.msc2716_enabled` config enabled and be the room creator to avoid power level intricacies.

In room versions that support MSC2716, we allow everything regardless.

clokep: Should the first comment read:

> We only want to process MSC2716 events when:
> 
> * The room version supports MSC2716
> * Any room ~~version~~ with the `msc2716_enabled` experimental feature enabled and the event is from the room creator

MadLittleMods: @clokep Either seems valid but I could see how that would be more clear 👍. I was trying to convey that it works in all of the other current room versions.

clokep: Alright. Thanks.

clokep (approved):
I think the changes here make sense, I do wonder if we should consolidate the logic (and not do database queries that might not be used). Neither of those are necessary though.

MadLittleMods:
Thanks for the review and understanding the problem @clokep 🐪

PR #10965, opened 2021-10-01
Annotate `client` on `TransportLayerClient`

reivilibre:


synapse/federation/transport/client.py
```diff
@@ -47,7 +47,7 @@ class TransportLayerClient:
 
     def __init__(self, hs):
         self.server_name = hs.hostname
-        self.client = hs.get_federation_http_client()
+        self.client: MatrixFederationHttpClient = hs.get_federation_http_client()
```
richvdh, line 54: it'd be better to annotate `hs` to be a `HomeServer`, then this would magically work.

synapse/federation/transport/client.py
```diff
@@ -68,12 +68,14 @@ async def get_room_state_ids(
...
 
         path = _create_v1_path("/state_ids/%s", room_id)
-        return await self.client.get_json(
+        resp = await self.client.get_json(
             destination,
             path=path,
             args={"event_id": event_id},
             try_trailing_slash_on_400=True,
         )
+        assert isinstance(resp, dict)
```
richvdh, line 81: This doesn't look correct. `get_json` returns whatever the remote endpoint returns; we have no way to know it is a dict. Checking that it *is* a dict shouldn't be done with assert.

The caller (`federation_client.get_room_state_ids`) has some checks that the shape of the response is correct. It should also be checking that it's a dict.

The same applies to the other assertions in this commit.

synapse/federation/transport/client.py
```diff
@@ -104,7 +108,7 @@ async def get_event(
 
     @log_function
     async def backfill(
-        self, destination: str, room_id: str, event_tuples: Iterable[str], limit: int
+        self, destination: str, room_id: str, event_tuples: List[str], limit: int
```
richvdh, line 111: It'd be nice to have a type definition for "things that can be json-serialised into a list", which would include Tuples as well as Lists.

not a blocker though.

synapse/federation/transport/client.py
```diff
@@ -333,6 +333,12 @@ async def send_leave_v1(
             # sync.
             ignore_backoff=True,
         )
+        assert isinstance(resp, list)
```
richvdh, line 336: again, don't use assert for this, please.

richvdh (changes requested):
sorry, I don't think `assert` is the right tool for this job!

Issue #9346, opened 2021-02-08
We don't delete to_device messages aimed at a local device once that device is deleted


anoadragon453:
When a user's device is deleted - either manually, when they log out or when their account is deactivated - `to_device` messages aimed at that device are not deleted, and end up sitting in the database forever.

https://github.com/matrix-org/synapse/blob/a78016dadfb1680f5f77daae9948086b37cbeef8/synapse/handlers/device.py#L343-L417

There's no point in keeping them around, and should get rid of them when a device goes pop.

Related: https://github.com/matrix-org/synapse/issues/3599

anoadragon453:
After deleting these manually I found that ~48% of the `device_inbox` table was taken up by messages to devices that no longer existed.

Issue #3599, opened 2018-07-24
device_inbox never gets emptied


ara4n:
it's up to 85M rows or so on matrix.org, which seems insane. similarly *_stream never seem to get purged, or device_federation_inbox/outbox

richvdh:
The main problem is that it's hard to tell when we can delete to-device messages. They don't have an expiry time on them.

ukcb:
I made a small **test script** for private use with postgresql. Let's see if it works. :-)

```
#!/usr/bin/perl

use strict;
use warnings;

use DBI;
use DateTime;

my $db_name = 'synapse';
my $db_host = '127.0.0.1';
my $db_port = '5432';
my $db_user = 'synapse';
my $db_pass = 'synapse';

my $time_to_keep_in_days = 7;

my $table_device_inbox = 'device_inbox';
my $table_time_indexer = 'device_inbox_timestamp_indexer';


my $datetime        = DateTime->today;
my $timestamp_today = $datetime->epoch;
$datetime->subtract( days => $time_to_keep_in_days );
my $timestamp_obsolete = $datetime->epoch;


my $dbh = DBI->connect( "dbi:Pg:dbname=$db_name;host=$db_host;port=$db_port",
    $db_user, $db_pass, { AutoCommit => 1, RaiseError => 1, PrintError => 1 } )
  or die $DBI::errstr;

$dbh->do("CREATE TABLE IF NOT EXISTS $table_time_indexer
            ( stream_id BIGINT,
              timestamp BIGINT,
              PRIMARY KEY (stream_id)
            )") or die $dbh->errstr;

my $sth = $dbh->prepare("SELECT stream_id FROM $table_device_inbox")
  or warn $dbh->errstr;
$sth->execute() or warn $dbh->errstr;

my $pool = {};
while ( my $ref = $sth->fetchrow_hashref() ) {
    $pool->{ $ref->{'stream_id'} } = 1;
}

$sth = $dbh->prepare("SELECT stream_id, timestamp FROM $table_time_indexer")
  or warn $dbh->errstr;
$sth->execute() or warn $dbh->errstr;

my $index = {};
while ( my $res = $sth->fetchrow_hashref() ) {
    $index->{ $res->{'stream_id'} } = $res->{'timestamp'};
}


foreach my $key ( keys $pool->%* ) {
    if ( $index->%{$key} ) {
        if ( $index->%{$key} < $timestamp_obsolete ) {
            delete_key_from_tables($key);
        }
    }
[...]

PR #10098, opened 2021-05-30
fix #3599 delete messages from device_inbox table when deleting device

JohannesKleine:
### Pull Request Checklist

* [x] Pull request is based on the develop branch
* [ ] Pull request includes a [changelog file](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md#changelog). The entry should:
  - Be a short description of your change which makes sense to users. "Fixed a bug that prevented receiving messages from other servers." instead of "Moved X method from `EventStore` to `EventWorkerStore`.".
  - Use markdown where necessary, mostly for `code blocks`.
  - End with either a period (.) or an exclamation mark (!).
  - Start with a capital letter.
* [x] Pull request includes a [sign off](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md#sign-off)
* [ ] Code style is correct (run the [linters](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md#code-style))

Issue #9348, opened 2021-02-08
Synapse accept to_device messages to hidden devices


anoadragon453:
As part of cross-signing, devices for the master, self-signing and user-signing cross-signing keys are created and stored in the `devices` table.

These devices should not be getting `to_device` messages sent to them, as they will never be read by any client. It turns out that it is possible to have `to_device` messages aimed at them in the `device_inbox` table through the following:

* Clients can directly send `to_device` messages to one of your cross-signing public keys. These happen to be the device IDs of these hidden devices. Synapse doesn't condition on the device being hidden, and so will accept them.
* Clients often send `to_device` messages to all devices of a user, by specifying `*` as the target. Synapse will then create a `to_device` message for all devices of a user, including hidden ones:

https://github.com/matrix-org/synapse/blob/c7b823525e9598ee4de76099d9ca5ea1ce3977f4/synapse/storage/databases/main/deviceinbox.py#L486-L498

These messages were never getting read, and thus piling up in the database forever. In my testing they didn't seem to be using *that* much space however. Upon removing them for all users on my homeserver and vacuuming the table, the size of `device_inbox` only went from 3420MB to 3410MB.

Initially presumed as a client-side issue, before we discovered Synapse was at fault: https://github.com/vector-im/element-web/issues/15638

jryans:
Related to https://github.com/matrix-org/synapse/issues/3656

anoadragon453:
I'd like to see a solution which:

* Stops adding messages for hidden devices when the user specifies `*`
* Silently ignore messages directed at hidden devices (just as we do today for non-existent ones)
* Removes existing messages that are directed at hidden devices to cut down on database size. I ended up using:

```sql
delete from device_inbox where device_id in (select device_id from devices where hidden = true);
```

DMRobertson:
Should this be closed? AFAICS #11199 will delete the stuff we don't want on Synapse start up, but it won't address the first two bullets in @anoadragon453's https://github.com/matrix-org/synapse/issues/9348#issuecomment-775237673.

anoadragon453:
@DMRobertson #10097 should address the first two points, and has been merged.

I'm not sure if it's worth keeping this open for #11199, unless we're specifically tracking this issue. The third bullet point was mostly mentioned as a nice-to-have.

DMRobertson:
Ahh, sorry. I confused #10097 and #11199. I hadn't spotted the former; I thought the latter was just merged. Agreed entirely.

PR #10969, opened 2021-10-01
Delete messages from `device_inbox` table when deleting device

dklimpel:
Fixes: #3599
Replaces: #10098

Related:

- #9346
- #9348
- #10097

### Pull Request Checklist


tests/handlers/test_device.py
```diff
@@ -160,6 +160,36 @@ def test_delete_device(self):
...
+        # check that the device_inbox was deleted
+        self.get_failure(
+            self.store.db_pool.simple_select_one(
+                table="device_inbox",
+                keyvalues={"user_id": user1, "device_id": "abc"},
+                retcols=("user_id", "device_id"),
+                desc="get_device_id_from_device_inbox",
+            ),
+            synapse.api.errors.StoreError,
+        )
```
DMRobertson, line 191: You might prefer to pass `allow_none=True` to `simple_select_one`. That will return `None` rather than raise an exception if there are no rows matching that query.

tests/storage/databases/main/test_devices.py
```diff
@@ -0,0 +1,95 @@
...
+        # Make sure the background task deleted old device_inbox
+        res = self.get_success(
+            self.store.db_pool.simple_select_onecol(
+                table="device_inbox",
+                keyvalues={},
+                retcol="device_id",
+                desc="get_device_id_from_device_inbox",
+            )
+        )
+        self.assertEqual(1, len(res))
```
DMRobertson, line 95: I think I'd add an extra assertion which checks that `res[0]` refers to `current_device`, not `old_device`.

synapse/storage/databases/main/devices.py
```diff
@@ -1045,6 +1050,63 @@ def _txn(txn):
 
         return rows
 
+    async def _remove_deleted_devices_from_device_inbox(
+        self, progress: dict, batch_size: int
+    ) -> int:
+        """A background update that deletes all device_inboxes for deleted devices.
+
+        This should only need to be run once (when users upgrade to v1.45.0)
```
DMRobertson, line 1058: There are some delta sql scripts mentioned below. Is the intent we run those as migrations on synapse startup to trigger this bg update? (Rather than admins manually triggering it)

DMRobertson:
> Fixes #3599

I think this actually fixes #9346 (?)

synapse/storage/databases/main/devices.py
dklimpel: Yes, this is for migration. The intent ist to run in bg job instead waiting for startup because it need a lot of time to run:
```sql
DELETE FROM device_inbox WHERE device_id NOT IN (SELECT device_id FROM devices);
```

IMO it is an improvment to #10098
Similar to this discussion: https://github.com/matrix-org/synapse/pull/10097#discussion_r643171952 and this PR: #10581

changelog.d/10969.bugfix
```diff
@@ -0,0 +1 @@
+Fix a bug where messages in `device_inbox` table where not deleted, when deleting device(s).
```
DMRobertson, line 1: I'm guessing this has been around basically forever, so I suggest we describe the bug `long-standing`. (Unless we happen to know that it was introduced in a particular release?)

I've tweaked the wording a little too, to emphasise the pain for the server administrator (lots of DB rows!). I've also added a sentence to attribute yourself and Johannes. What do you think?

```suggestion
Fix a long-standing bug where messages in the `device_inbox` table for deleted devices would persist indefinitely. Contributed by @dklimpel and @JohannesKleine.
```

DMRobertson (commented):
Looks good to me. I'm not too familiar with this kind of db wrangling in synapse though, so I'd appreciate another pair of eyes from @matrix-org/synapse-core.

changelog.d/10969.bugfix
dklimpel: I think this is a bug since the beginning.

dklimpel:
> Looks good to me. I'm not too familiar with this kind of db wrangling in synapse though, so I'd appreciate another pair of eyes from @matrix-org/synapse-core.

This is my first bg task. It was not very easy for me.

DMRobertson:
> This is my first bg task. It was not very easy for me.

It's much appreciated! I forgot to say earlier: thank you for picking up where the previous PR left off!

Is there anything we could have done to make it easier for you? E.g. documentation that was lacking?

dklimpel:
At the moment there is no doc for big tasks. But it is an very special topic.

There are missing unit tests for device_inbox. That is the reason why I am poke in database table in unit tests. This is not my favorite style.
Most time is write unit tests. But it is necessary.

synapse/storage/databases/main/devices.py
```diff
@@ -1045,6 +1050,62 @@ def _txn(txn):
...
+                self.db_pool.simple_delete_txn(
+                    txn,
+                    "device_inbox",
+                    {"user_id": row[0], "device_id": row[1], "stream_id": row[2]},
+                )
+
+            if rows:
+                self.db_pool.updates._background_update_progress_txn(
+                    txn, "remove_deleted_devices_from_device_inbox", row
+                )
```
clokep, line 1093: We don't seem to use the progress we've made at all here, it gets passed back into `_remove_deleted_devices_from_device_inbox` but we seem to always run the same query. We might want to do something like `WHERE device_id > last_seen_device_id`.

clokep (commented):
Looks like a good start! Thanks for tackling something new to you, background updates are definitely a bit under documented at the moment.

synapse/storage/databases/main/devices.py
```diff
@@ -1133,6 +1194,12 @@ async def delete_device(self, user_id: str, device_id: str) -> None:
             desc="delete_device",
         )
 
+        await self.db_pool.simple_delete(
+            table="device_inbox",
+            keyvalues={"user_id": user_id, "device_id": device_id},
+            desc="delete_device_inbox",
+        )
```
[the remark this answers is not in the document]
dklimpel: What is the solution?
A transaction: `BEGIN; DELETE ; COMMIT;`
Or is there an other examle in the code?

tests/storage/databases/main/test_devices.py
```diff
@@ -0,0 +1,96 @@
...
+        # ... and tell the DataStore that it hasn't finished all updates yet
+        self.store.db_pool.updates._all_done = False
+
+        # Now let's actually drive the updates to completion
+        while not self.get_success(
+            self.store.db_pool.updates.has_completed_background_updates()
+        ):
+            self.get_success(
+                self.store.db_pool.updates.do_next_background_update(100), by=0.1
+            )
```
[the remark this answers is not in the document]
dklimpel: Do you mean changes in `tests/unittest.py` in
- #10982 ?

synapse/storage/databases/main/devices.py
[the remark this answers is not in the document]
dklimpel: Probably is it also possible with a left join.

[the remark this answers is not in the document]
clokep: I mean the various places in the code where we use `db_pool.runInteraction(...)` with a function that then would use `simple_delete_txn` inside of it.

synapse/storage/databases/main/devices.py
```diff
@@ -1121,18 +1189,27 @@ async def store_device(
...
+            )
+
+            self.db_pool.simple_delete_txn(
+                txn,
+                table="device_inbox",
+                keyvalues={"user_id": user_id, "device_id": device_id},
+            )
+
+        await self.db_pool.runInteraction("delete_device", _delete_device_txn)
         self.device_id_exists_cache.invalidate((user_id, device_id))
```
clokep, line 1213: I wonder if this entire method should just be:

```python
await self.delete_devices(user_id, [device_id])
```

Instead of duplicating the logic of how to delete these two tables.

tests/storage/databases/main/test_devices.py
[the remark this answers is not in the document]
clokep: Yes, that's the one! That's now merged too so if we merge develop in here we should be able to use that helper!

clokep:
> Is it important which class contains the bg task `remove_deleted_devices_from_device_inbox`? `DeviceBackgroundUpdateStore` or `DeviceInboxBackgroundUpdateStore`
> 
> And which one is the correct one?

I don't think it is important, as it is cleaning up `device_inbox` it might make more sense to go into `DeviceInboxBackgroundUpdateStore` though?

PR #10970, opened 2021-10-01
Fix inconsistent behavior of `get_last_client_by_ip`

squahtx:
Make `get_last_client_by_ip` return the same dictionary structure
regardless of whether the data has been persisted to the database.

This change will allow slightly cleaner type hints to be applied later
on.

### DIFF
--- a/synapse/storage/databases/main/client_ips.py
+++ b/synapse/storage/databases/main/client_ips.py
@@ -538,15 +538,20 @@ class ClientIpStore(ClientIpWorkerStore):
         """
         ret = await super().get_last_client_ip_by_device(user_id, device_id)
 
-        # Update what is retrieved from the database with data which is pending insertion.
+        # Update what is retrieved from the database with data which is pending
+        # insertion, as if it has already been stored in the database.
         for key in self._batch_row_update:
-            uid, access_token, ip = key
+            uid, _access_token, ip = key
             if uid == user_id:
                 user_agent, did, last_seen = self._batch_row_update[key]
+
+                if did is None:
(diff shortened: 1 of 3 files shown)


synapse/storage/databases/main/client_ips.py
```diff
@@ -538,15 +538,20 @@ async def get_last_client_ip_by_device(
...
+
+                if did is None:
+                    # These updates don't make it to the `devices` table
+                    continue
+
                 if not device_id or did == device_id:
-                    ret[(user_id, device_id)] = {
+                    ret[(user_id, did)] = {
                         "user_id": user_id,
-                        "access_token": access_token,
```
squahtx, line 549: `access_token` was intentionally removed in 51d28272e20d799b2e35a8a14b3c1d9d5f555d10

synapse/storage/databases/main/client_ips.py
```diff
@@ -538,15 +538,20 @@ async def get_last_client_ip_by_device(
...
             if uid == user_id:
                 user_agent, did, last_seen = self._batch_row_update[key]
+
+                if did is None:
+                    # These updates don't make it to the `devices` table
+                    continue
+
                 if not device_id or did == device_id:
-                    ret[(user_id, device_id)] = {
+                    ret[(user_id, did)] = {
```
squahtx, line 553: `device_id` can be `None`, which means info for all devices should be reported (as noted in the docstring).  
The fetch-from-database code path files entries under their actual device id rather than `None`, so the previous code here was inconsistent.

PR #10968, opened 2021-10-01
Fix error in `get_user_ip_and_agents` when fetching from the database

squahtx:
Fixes #10957, introduced in v1.44.0rc1

PR #10972, opened 2021-10-01
Add type hints to `synapse.storage.databases.main.client_ips`

squahtx:
Built on top of #10968 and #10970, otherwise the types don't check out.  Those commits can be ignored.


synapse/handlers/device.py
```diff
@@ -595,7 +606,7 @@ async def rehydrate_device(
 
 
 def _update_device_from_client_ips(
-    device: JsonDict, client_ips: Dict[Tuple[str, str], JsonDict]
+    device: JsonDict, client_ips: Mapping[Tuple[str, str], Mapping[str, Any]]
```
squahtx, line 609: `client_ips` comes from the return of `get_last_client_ip_by_device`. I downgraded the `Dict`s to `Mapping`s to make mypy happy.

changelog.d/10968.bugfix
```diff
@@ -0,0 +1 @@
+Fix `/admin/whois/{user_id}` endpoint, which was broken in v1.44.0rc1.
```
DMRobertson, line 1: Are these changelogs meant to be in here?

(Guessing they're already merged to develop?)

mypy.ini
```diff
@@ -99,6 +100,9 @@ disallow_untyped_defs = True
 [mypy-synapse.rest.*]
 disallow_untyped_defs = True
 
+[mypy-synapse.storage.databases.main.client_ips]
+disallow_untyped_defs = True
```
DMRobertson, line 104: <3

synapse/handlers/device.py
DMRobertson: Why was mypy unhappy here, sorry---something to do with a `Dict` being mutable?

synapse/module_api/__init__.py
```diff
@@ -773,9 +773,9 @@ async def get_user_ip_and_agents(
             # Sanitize some of the data. We don't want to return tokens.
             return [
                 UserIpAndAgent(
-                    ip=str(data["ip"]),
-                    user_agent=str(data["user_agent"]),
-                    last_seen=int(data["last_seen"]),
+                    ip=data["ip"],
+                    user_agent=data["user_agent"],
+                    last_seen=data["last_seen"],
```
DMRobertson, line 778: Note to self: last_seen was `int`, is now a `str`?

synapse/storage/databases/main/client_ips.py
```diff
@@ -29,8 +41,32 @@
 LAST_SEEN_GRANULARITY = 120 * 1000
 
 
+LastClientIpByDeviceEntry = TypedDict(
```
DMRobertson, line 44: Ooooh `TypedDict`! ooi, was there a reason you preferred the function-call form over the dataclass-like form? (i.e. a class that inherits from TypedDict)

synapse/storage/databases/main/client_ips.py
```diff
@@ -189,7 +229,9 @@ def remove(txn):
                 ),
                 args,
             )
-            res = txn.fetchall()
+            res = cast(
+                List[Tuple[str, str, str, Optional[str], str, int, int]], txn.fetchall()
+            )
```
DMRobertson, line 234: Part of me wishes we were using bits of sqlalchemy so we could [get this kind of inference for free](https://docs.sqlalchemy.org/en/14/orm/extensions/mypy.html). I don't know how much you have to buy in to using the ORM for that to pay off though.

synapse/storage/databases/main/client_ips.py
```diff
@@ -538,15 +599,20 @@ async def get_last_client_ip_by_device(
...
             if uid == user_id:
                 user_agent, did, last_seen = self._batch_row_update[key]
+
+                if did is None:
+                    # These updates don't make it to the `devices` table
+                    continue
+
                 if not device_id or did == device_id:
-                    ret[(user_id, device_id)] = {
+                    ret[(user_id, did)] = {
```
DMRobertson, line 614: Is this a bugfix? (If `device_id` is falsey, should we be using it as part of the key?)

synapse/storage/databases/main/client_ips.py
```diff
@@ -584,15 +650,15 @@ def get_recent(txn):
...
 
         rows = await self.db_pool.runInteraction(
             desc="get_user_ip_and_agents", func=get_recent
         )
 
         results.update(
-            ((row["access_token"], row["ip"]), (row["user_agent"], row["last_seen"]))
-            for row in rows
+            ((access_token, ip), (user_agent, last_seen))
+            for access_token, ip, user_agent, last_seen in rows
```
DMRobertson, line 661: How does this work? On the left, `rows` looks like a dictionary. Ohhh... is `rows` some kind of proxy object that you can treat as both a mapping and a sequence?

synapse/module_api/__init__.py
DMRobertson: Could you explain why this is safe? (Maybe I just need to chase the logic through next week?)

changelog.d/10968.bugfix
squahtx: They're from PRs #10968 and #10972. One's in the release branch and the other's not merged yet. Without those PRs the types don't check out.

synapse/handlers/device.py
squahtx: It's to do with the inner `JsonDict` not being the same as `LastClientIpByDeviceEntry`:
`Argument 2 ... has incompatible type "Dict[Tuple[str, str], LastClientIpByDeviceEntry]"; expected "Dict[Tuple[str, str], Dict[str, Any]]"  [arg-type]`

synapse/module_api/__init__.py
squahtx: I thought `data["last_seen"]` is still an `int`?

`raw_data` used to be a `Dict[str, Union[str, int]]`, but is now a `TypedDict` so these ought to already have the correct type.

synapse/storage/databases/main/client_ips.py
squahtx: I didn't know the dataclass-like form existed. Which style do we prefer?

DMRobertson: Quick grep suggests dataclass-like, e.g.

https://github.com/matrix-org/synapse/blob/94b620a5edd6b5bc55c8aad6e00a11cc6bf210fa/synapse/handlers/register.py#L81-L85

squahtx: Let's go with that then. Do you have any suggestions for better names for these `TypedDict`s?

synapse/storage/databases/main/client_ips.py
squahtx: That looks pretty handy. I wonder how hard it is to make a mypy module that would do that for us.

DMRobertson: I would imagine fairly hard? I'm imagining something which parses sql queries to work out what's being selected, and cross-reference against a schema to work out what type will be fed back into Python.

synapse/storage/databases/main/client_ips.py
squahtx: It is a bugfix, from #10970  

Basically the data store buffers updates in `_batch_row_update` and writes them to the database periodically. The call to `super().get_last_client_ip_by_device(...)` above fetches from database while this code adds data from the in-memory buffer to the results.

The db implementation of `get_last_client_ip_by_device` always uses the correct `device_id` in the key whereas the original code here didn't.

synapse/storage/databases/main/client_ips.py
squahtx: This bugfix is from #10968 - the previous code was broken because `rows` is a tuple and not a dict.

changelog.d/10968.bugfix
DMRobertson: Ahhh, I should have seen your description and only looked at the last commit.

synapse/storage/databases/main/client_ips.py
DMRobertson: I think they're okay as they are. I'm guessing that `Entry` means that they correspond directly to a database table (or some subset of its columns?)

DMRobertson (approved):

synapse/storage/databases/main/client_ips.py
squahtx: I've converted them to class-forms and tried to give them better names. Hopefully they make sense.

At the time `Entry` really didn't mean anything, though I guess they do correspond to columns in database tables. That's an implementation detail internal to the data store, so perhaps it's not the best name.

DMRobertson (approved):
lgtm!

squahtx:
Note to self: do not merge until #10970 has gone through!

PR #10975, opened 2021-10-02
Draft:  Resolve and share `state_groups` for all historical events in batch (MSC2716)

MadLittleMods:
Resolve and share `state_groups` for all historical events in batch

### Pull Request Checklist


synapse/handlers/message.py
```diff
@@ -969,7 +978,16 @@ async def create_new_client_event(
...
-            context = await self.state.compute_event_context(event)
+            old_state = None
+            # Define the state for historical messages while we know to get all of
+            # state_groups setup properly when we `compute_event_context`.
+            if builder.internal_metadata.is_historical() and full_state_ids_at_event:
+                old_state = await self.store.get_events_as_list(full_state_ids_at_event)
+
+            context = await self.state.compute_event_context(
+                event,
+                # old_state=old_state
```
MadLittleMods, line 989: We can revert if we don't need `old_state` here (it's commented out atm) and all of the `auth_event_ids` stuff. All of this stuff since it's part of `v1` in https://github.com/matrix-org/synapse/pull/10948

synapse/rest/client/room_batch.py
```diff
@@ -254,8 +259,8 @@ async def on_POST(
                     room_id=room_id,
                     action=membership,
                     content=event_dict["content"],
-                    outlier=True,
-                    prev_event_ids=[fake_prev_event_id],
+                    # outlier=True,
```
MadLittleMods, line 262: Removed `outlier` so we get `state_groups`

synapse/rest/client/room_batch.py
```diff
@@ -216,6 +220,10 @@ async def on_POST(
         prev_state_ids = list(prev_state_map.values())
         auth_event_ids = prev_state_ids
 
+        # Make the state events float off on their own
+
+        prev_event_id_for_state_chain = generate_fake_prev_event_id()
```
MadLittleMods, line 225: We still have the state events float off on their own, but now their in a floating chain instead of individual.

synapse/rest/client/room_batch.py
```diff
@@ -325,14 +327,12 @@ async def on_POST(
         # an insertion event), in which case we just create a new insertion event
         # that can then get pointed to by a "marker" event later.
         else:
-            prev_event_ids = prev_event_ids_from_query
```
MadLittleMods, line 328: I think this was actually a bug. We only want the base insertion event to be tied to `prev_event_ids_from_query`.

Whereas previously, this attached the base and insertion event part of the batch.

synapse/rest/client/room_batch.py
```diff
@@ -383,11 +383,20 @@ async def on_POST(
         # Prepend the insertion event to the start of the batch (oldest-in-time)
         events_to_create = [insertion_event] + events_to_create
 
+        # Also connect the historical event chain to floating state chain,
+        # which causes the HS to ask for the state at the start of
+        # the batch later.
+        prev_event_ids = [prev_event_id_for_state_chain]
```
MadLittleMods, line 389: Connect our historical batch chain to the float state chain so we still get the floating benefit but also can resolve in the state and share `state_groups` across the batch.

synapse/rest/client/room_batch.py
```diff
@@ -383,11 +383,20 @@ async def on_POST(
...
+
         event_ids = []
         events_to_persist = []
         for ev in events_to_create:
             assert_params_in_dict(ev, ["type", "origin_server_ts", "content", "sender"])
 
+            assert self.hs.is_mine_id(ev["sender"]), "User must be our own: %s" % (
+                event.sender,
+            )
```
MadLittleMods, line 398: Moved this up because it makes sense to assert this earlier rather than later.

synapse/storage/databases/main/events.py
```diff
@@ -2071,13 +2069,18 @@ def _store_event_state_mappings_txn(
 
             state_groups[event.event_id] = context.state_group
 
-        self.db_pool.simple_insert_many_txn(
+        self.db_pool.simple_upsert_many_txn(
```
MadLittleMods, line 2072: Changed this to an upsert because to avoid colliding with a previous entry we put in via `store_state_group_id_for_event_id` earlier before we persisted the events.

synapse/handlers/message.py
```diff
@@ -640,6 +640,12 @@ async def create_event(
             auth_event_ids=auth_event_ids,
             depth=depth,
         )
+        logger.info(
+            "create_new_client_event %s event=%s state_group=%s",
+            event.type,
+            event.event_id,
+            context._state_group,
+        )
```
MadLittleMods, line 648: Maybe can use the same `old_state` trick from https://github.com/matrix-org/synapse/pull/10948 but only do it for `insertion` events which are the root ancestor of each batch.

(also make sure to make the state events as outliers again)

synapse/rest/client/room_batch.py
MadLittleMods: Before (expected) | Before (actual, [because of bug](https://github.com/matrix-org/synapse/pull/10975#discussion_r720625888))
--- | ---
[Mermaid live editor](https://mermaid-js.github.io/mermaid-live-editor/edit/#eyJjb2RlIjoiZmxvd2NoYXJ0IEJUXG4gICAgc3ViZ3JhcGggbGl2ZVxuICAgICAgICBCIC0tLS0tLS0tLS0tLS0tLS0tLS0tPiBBXG4gICAgZW5kXG4gICAgXG4gICAgc3ViZ3JhcGggY2h1bmswXG4gICAgICAgIGNodW5rMC1jaHVua1tbXCJjaHVua1wiXV0gLS0-IGNodW5rMC0yKChcIjJcIikpIC0tPiBjaHVuazAtMSgoMSkpIC0tPiBjaHVuazAtMCgoMCkpIC0tPiBjaHVuazAtaW5zZXJ0aW9uWy9pbnNlcnRpb25cXF1cbiAgICBlbmRcblxuICAgIHN1YmdyYXBoIGNodW5rMVxuICAgICAgICBjaHVuazEtY2h1bmtbW1wiY2h1bmtcIl1dIC0tPiBjaHVuazEtMigoXCIyXCIpKSAtLT4gY2h1bmsxLTEoKDEpKSAtLT4gY2h1bmsxLTAoKDApKSAtLT4gY2h1bmsxLWluc2VydGlvblsvaW5zZXJ0aW9uXFxdXG4gICAgZW5kXG4gICAgXG4gICAgc3ViZ3JhcGggY2h1bmsyXG4gICAgICAgIGNodW5rMi1jaHVua1tbXCJjaHVua1wiXV0gLS0-IGNodW5rMi0yKChcIjJcIikpIC0tPiBjaHVuazItMSgoMSkpIC0tPiBjaHVuazItMCgoMCkpIC0tPiBjaHVuazItaW5zZXJ0aW9uWy9pbnNlcnRpb25cXF1cbiAgICBlbmRcblxuXG4gICAgY2h1bmswLWluc2VydGlvbiAtLT4gZmFrZVByZXZFdmVudDB7e2Zha2VfcHJldl9ldmVudH19XG4gICAgY2h1bmsxLWluc2VydGlvbiAtLT4gZmFrZVByZXZFdmVudDF7e2Zha2VfcHJldl9ldmVudH19XG4gICAgY2h1bmsyLWluc2VydGlvbiAtLT4gZmFrZVByZXZFdmVudDJ7e2Zha2VfcHJldl9ldmVudH19XG5cbiAgICBcbiAgICBjaHVuazAtaW5zZXJ0aW9uQmFzZVsvaW5zZXJ0aW9uXFxdIC0tLS0tLS0tLS0tLS0tLT4gQVxuICAgIGNodW5rMC1jaHVuayAtLi0-IGNodW5rMC1pbnNlcnRpb25CYXNlWy9pbnNlcnRpb25cXF1cbiAgICBjaHVuazEtY2h1bmsgLS4tPiBjaHVuazAtaW5zZXJ0aW9uXG4gICAgY2h1bmsyLWNodW5rIC0uLT4gY2h1bmsxLWluc2VydGlvblxuIiwibWVybWFpZCI6IntcbiAgXCJ0aGVtZVwiOiBcImRlZmF1bHRcIlxufSIsInVwZGF0ZUVkaXRvciI6ZmFsc2UsImF1dG9TeW5jIjp0cnVlLCJ1cGRhdGVEaWFncmFtIjpmYWxzZX0)  | [Mermaid live editor](https://mermaid-js.github.io/mermaid-live-editor/edit/#eyJjb2RlIjoiZmxvd2NoYXJ0IEJUXG4gICAgc3ViZ3JhcGggbGl2ZVxuICAgICAgICBCIC0tLS0tLS0tLS0tLS0tLS0tLS0tPiBBXG4gICAgZW5kXG4gICAgXG4gICAgc3ViZ3JhcGggY2h1bmswXG4gICAgICAgIGNodW5rMC1jaHVua1tbXCJjaHVua1wiXV0gLS0-IGNodW5rMC0yKChcIjJcIikpIC0tPiBjaHVuazAtMSgoMSkpIC0tPiBjaHVuazAtMCgoMCkpIC0tPiBjaHVuazAtaW5zZXJ0aW9uWy9pbnNlcnRpb25cXF1cbiAgICBlbmRcblxuICAgIHN1YmdyYXBoIGNodW5rMVxuICAgICAgICBjaHVuazEtY2h1bmtbW1wiY2h1bmtcIl1dIC0tPiBjaHVuazEtMigoXCIyXCIpKSAtLT4gY2h1bmsxLTEoKDEpKSAtLT4gY2h1bmsxLTAoKDApKSAtLT4gY2h1bmsxLWluc2VydGlvblsvaW5zZXJ0aW9uXFxdXG4gICAgZW5kXG4gICAgXG4gICAgc3ViZ3JhcGggY2h1bmsyXG4gICAgICAgIGNodW5rMi1jaHVua1tbXCJjaHVua1wiXV0gLS0-IGNodW5rMi0yKChcIjJcIikpIC0tPiBjaHVuazItMSgoMSkpIC0tPiBjaHVuazItMCgoMCkpIC0tPiBjaHVuazItaW5zZXJ0aW9uWy9pbnNlcnRpb25cXF1cbiAgICBlbmRcblxuXG4gICAgY2h1bmswLWluc2VydGlvbiAtLT4gZmFrZVByZXZFdmVudDB7e2Zha2VfcHJldl9ldmVudH19XG4gICAgY2h1bmsxLWluc2VydGlvbiAtLT4gZmFrZVByZXZFdmVudDF7e2Zha2VfcHJldl9ldmVudH19XG4gICAgY2h1bmsyLWluc2VydGlvbiAtLT4gZmFrZVByZXZFdmVudDJ7e2Zha2VfcHJldl9ldmVudH19XG5cbiAgICBcbiAgICBjaHVuazAtaW5zZXJ0aW9uQmFzZVsvaW5zZXJ0aW9uXFxdIC0tLS0tLS0tLS0tLS0tLT4gQVxuICAgIGNodW5rMC1pbnNlcnRpb24gLS0-IEFcblxuICAgIGNodW5rMC1jaHVuayAtLi0-IGNodW5rMC1pbnNlcnRpb25CYXNlWy9pbnNlcnRpb25cXF1cbiAgICBjaHVuazEtY2h1bmsgLS4tPiBjaHVuazAtaW5zZXJ0aW9uXG4gICAgY2h1bmsyLWNodW5rIC0uLT4gY2h1bmsxLWluc2VydGlvblxuIiwibWVybWFpZCI6IntcbiAgXCJ0aGVtZVwiOiBcImRlZmF1bHRcIlxufSIsInVwZGF0ZUVkaXRvciI6ZmFsc2UsImF1dG9TeW5jIjp0cnVlLCJ1cGRhdGVEaWFncmFtIjpmYWxzZX0)

synapse/handlers/message.py
```diff
@@ -968,6 +971,13 @@ async def create_new_client_event(
         if builder.internal_metadata.outlier:
             event.internal_metadata.outlier = True
             context = EventContext.for_outlier()
+        elif (
+            event.type == EventTypes.MSC2716_INSERTION
+            and full_state_ids_at_event
+            and builder.internal_metadata.is_historical()
+        ):
+            old_state = await self.store.get_events_as_list(full_state_ids_at_event)
+            context = await self.state.compute_event_context(event, old_state=old_state)
```
MadLittleMods, line 980: Add explicit state to the insertion event so the rest of the batch can inherit the same state and `state_group`

synapse/handlers/federation_event.py
```diff
@@ -1010,9 +1010,8 @@ async def _handle_marker_event(self, origin: str, marker_event: EventBase) -> No
         room_version = await self._store.get_room_version(marker_event.room_id)
         create_event = await self._store.get_create_event_for_room(marker_event.room_id)
         room_creator = create_event.content.get(EventContentFields.ROOM_CREATOR)
-        if (
-            not room_version.msc2716_historical
-            or not self._config.experimental.msc2716_enabled
+        if not room_version.msc2716_historical and (
+            not self._config.experimental.msc2716_enabled
```
MadLittleMods, line 1014: This fix is part of https://github.com/matrix-org/synapse/pull/10962 and just here so I can still test this in Gitter.

synapse/storage/databases/main/room_batch.py
```diff
@@ -36,3 +36,16 @@ async def get_insertion_event_by_batch_id(
             retcol="event_id",
             allow_none=True,
         )
+
+    async def store_state_group_id_for_event_id(
+        self, event_id: str, state_group_id: int
+    ) -> Optional[str]:
+        {
```
DMRobertson, line 43: Not sure what's going on here. The braces to enclose the `await` is legit syntax (it'll build a set, like `{"hello"}`). As written the function is going to return `None`.

DMRobertson: See #11310

synapse/handlers/message.py
```diff
@@ -965,6 +969,13 @@ async def create_new_client_event(
         if builder.internal_metadata.outlier:
             event.internal_metadata.outlier = True
             context = EventContext.for_outlier()
+        elif (
+            event.type == EventTypes.MSC2716_INSERTION
+            and full_state_ids_at_event
+            and builder.internal_metadata.is_historical()
+        ):
+            old_state = await self.store.get_events_as_list(full_state_ids_at_event)
```
richvdh, line 977: @MadLittleMods, @erikjohnston: I'm trying to figure out how this works, so that I can fit it in with my work on faster joins and partial room state.

so... how does this work? `full_state_ids_at_event` is set to `auth_event_ids`, which is absolutely *not* the full state at the event. Rather, as the docstring says, it is "The event ids to use as the auth_events for the new event." - ie, the subset of the state that is needed to auth the event.

MadLittleMods: The relevant call stack for this is:

```
RoomBatchSendEventRestServlet.on_POST
handle_batch_of_events
persist_historical_events
create_event
create_new_client_event
```

In `RoomBatchSendEventRestServlet.on_POST` we [fetch the full state for the event](https://github.com/matrix-org/synapse/blob/c56bfb08bc071368db23f3b1c593724eb4f205f0/synapse/rest/client/room_batch.py#L126-L132) we're inserting next to (specified by `/batch_send?prev_event_id=xxx`) as a base to start from. Then as we persist the `state_events_at_start`, we also add that to the `auth_event_ids` list.

That full `auth_event_ids` list is then passed down that call stack and gets stripped down to only include the necessary bits in the code here. And we use `full_state_ids_at_event` to "Add explicit state to the insertion event so the rest of the batch can inherit the same state and `state_group`"

---

Here is the context where we originally started stripping down the state: https://github.com/matrix-org/synapse/pull/9247#discussion_r651803946

richvdh: right, thanks for the explanation, but this seems extremely confusing. At the very least, if `create_new_client_event` and co are expecting `auth_event_ids` to be the full state rather than just the auth events, then the docstrings should be updated to say so - right now they are flat-out wrong.

But generally, this feels like an abuse of `auth_event_ids` - we shouldn't be using it to mean "normally the auth events, except under special circumstances, when it is the full state". I'm not sure I have any immediate suggestions, but adding a new parameter would be better than the current situation.

@erikjohnston I think you've been following the MSC2716 work a bit more than me - any thoughts here?

erikjohnston: Ah, I think I got confused when reading this as at some point the state at the insertion event was limited to the used auth events, which is no longer the case. 

I agree that we should therefore be passing the full state through as a separate parameter. Though at some point I'm wondering if this code path needs refactoring to make it easier to use in the `/send_batch` case, as its starting to get a bit unwieldy with all the different cases.

MadLittleMods: @richvdh Do you want me to tackle this new full state parameter? Don't want to collide and cause conflicts with your existing refactor.

richvdh: You're unlikely to conflict with me here, so please do go ahead!

PR #10980, opened 2021-10-04
More annotations for `synapse.logging`, part 1

DMRobertson:
Goal is to make `synapse.logging` pass `no-untyped-defs`. This is a first step.

Some work also done in #10943 in this area.

### DIFF
--- a/synapse/logging/context.py
+++ b/synapse/logging/context.py
@@ -22,20 +22,36 @@ them.
 
 See doc/log_contexts.rst for details on how this works.
 """
+import functools
 import inspect
 import logging
 import threading
 import typing
 import warnings
-from typing import TYPE_CHECKING, Optional, Tuple, TypeVar, Union
+from types import TracebackType
+from typing import (
+    TYPE_CHECKING,
+    Any,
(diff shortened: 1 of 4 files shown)


synapse/logging/context.py
```diff
@@ -711,16 +735,19 @@ def nested_logging_context(suffix: str) -> LoggingContext:
     )
 
 
-def preserve_fn(f):
+def preserve_fn(f: F) -> F:
```
squahtx, line 738: We might want add a comment noting that the return type of `F` changes into a `Deferred[T]` if the original return type was not a `Deferred`.

I don't think we came to a consensus in #synapse-devs on whether it was okay to use slightly misleading type hints in this kind of situation.

DMRobertson: Ahh, I didn't spot that this also happens in `run_in_background`. Hmm, maybe this needs more thought?

clokep:
@DMRobertson I think *most* of this PR was done in #11556. Should we close this for now?

DMRobertson:
Agreed!

PR #10981, opened 2021-10-04
Fix potential leak of per-room profiles when the user dir is rebuilt.

DMRobertson:
The meat is in the first commit. The next two are small cleanups that I'd like to make, but don't have a functional impact and aren't necessary.

### DIFF
--- a/synapse/storage/databases/main/user_directory.py
+++ b/synapse/storage/databases/main/user_directory.py
@@ -248,18 +248,15 @@ class UserDirectoryBackgroundUpdateStore(StateDeltasStore):
                     )
 
                 # Now update the room sharing tables to include this room.
-                to_insert = set()
                 is_public = await self.is_room_world_readable_or_publicly_joinable(
                     room_id
                 )
-                if is_public:
-                    for user_id in users_with_profile:
-                        to_insert.add(user_id)
-
-                    if to_insert:
-                        await self.add_users_in_public_rooms(room_id, to_insert)
-                        to_insert.clear()
+                if is_public and users_with_profile:
+                    await self.add_users_in_public_rooms(
+                        room_id, users_with_profile.keys()
+                    )
                 else:
+                    to_insert = set()
                     for user_id in users_with_profile:
                         # We want the set of pairs (L, M) where L and M are
                         # in `users_with_profile` and L is local.
--- a/synapse/storage/databases/main/user_directory.py
+++ b/synapse/storage/databases/main/user_directory.py
@@ -228,10 +228,6 @@ class UserDirectoryBackgroundUpdateStore(StateDeltasStore):
             is_in_room = await self.is_host_joined(room_id, self.server_name)
 
             if is_in_room:
-                is_public = await self.is_room_world_readable_or_publicly_joinable(
-                    room_id
-                )
-
                 # TODO this leaks per-room profiles!
                 users_with_profile = await self.get_users_in_room_with_profiles(room_id)
                 # Throw away users excluded from the directory.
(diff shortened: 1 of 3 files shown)


synapse/storage/databases/main/user_directory.py
```diff
@@ -228,10 +228,7 @@ def _get_next_batch(
             is_in_room = await self.is_host_joined(room_id, self.server_name)
 
             if is_in_room:
-                is_public = await self.is_room_world_readable_or_publicly_joinable(
-                    room_id
-                )
-
+                # TODO this leaks per-room profiles!
```
richvdh, line 231: this seems to be in conflict with the newsfile :thinking: 

I think I don't really understand what the TODO means. How does it leak them?

Perhaps it's ok if you're going to be updating this soon, but otherwise a few more words in this comment might be helpful.

synapse/storage/databases/main/user_directory.py
```diff
@@ -241,8 +242,11 @@ def _get_next_batch(
                     or await self.should_include_local_user_in_dir(user_id)
                 }
 
-                # Update each user in the user directory.
+                # Upsert a user_directory record for each remote user we see.
+                # (Local users are processed in `_populate_user_directory_users`.)
                 for user_id, profile in users_with_profile.items():
+                    if self.hs.is_mine_id(user_id):
+                        continue
```
richvdh, line 245: it might be worth adding a comment here to explain why it's important to skip local users here.

tests/storage/test_user_directory.py
```diff
@@ -76,6 +78,19 @@ async def get_users_in_user_directory(self) -> Set[str]:
         )
         return {row["user_id"] for row in result}
 
+    async def get_profiles_in_user_directory(self) -> Dict[str, ProfileInfo]:
```
richvdh, line 81: this class is starting to get to the point where docstrings for the methods wouldn't go amiss. Could you add a quick docstring for this method?

synapse/storage/databases/main/user_directory.py
```diff
@@ -248,18 +248,15 @@ def _get_next_batch(
...
                     room_id
                 )
-                if is_public:
-                    for user_id in users_with_profile:
-                        to_insert.add(user_id)
-
-                    if to_insert:
-                        await self.add_users_in_public_rooms(room_id, to_insert)
-                        to_insert.clear()
+                if is_public and users_with_profile:
```
richvdh, line 254: although I think it ends up being the same functionally (because it's a no-op if `users_with_profile` is empty), the conditions under which we run the `else` branch (vis: either `is_public` is false, or `is_public` is true but `users_with_profile` is empty) make it confusing to read. Suggest:

```suggestion
                if is_public:
                  if users_with_profile:
```
(with suitable indentation fixes)

richvdh (changes requested):

synapse/storage/databases/main/user_directory.py
DMRobertson: I'll expand the comment. For completeness here:

`get_users_in_room_with_profiles` reads from the `room_memberships` table. That contains a `display_name` and `avatar_url` specific to this room (it doesn't read from the `profiles` table). With my change, we no longer upsert `user_directory` and `user_directory_search` rows for local users in this function, but we *do* continue to do so for remote users.

I meant the TODO as a marker to track where the data we don't want to leak comes from.

synapse/storage/databases/main/user_directory.py
DMRobertson: Good spot, thank you---I hadn't thought of this at all!

richvdh (approved):
lgtm!

DMRobertson:
Many thanks for reviewing!

PR #10982, opened 2021-10-04
Don't drop user dir deltas when server leaves room

DMRobertson:
I spotted this when I was trying to do another refactor/tidy up of this code. Argh! Not sure if we ever hit this (I don't know the details of how `_handle_deltas` is hit---it presumably is sensitive to replication delays?).

Anyway, this squashes a bug and makes be feel better.

### DIFF
--- a/tests/storage/test_roommember.py
+++ b/tests/storage/test_roommember.py
@@ -169,12 +169,7 @@ class CurrentStateMembershipUpdateTestCase(unittest.HomeserverTestCase):
 
     def test_can_rerun_update(self):
         # First make sure we have completed all updates.
-        while not self.get_success(
-            self.store.db_pool.updates.has_completed_background_updates()
-        ):
-            self.get_success(
-                self.store.db_pool.updates.do_next_background_update(100), by=0.1
-            )
+        self.wait_for_background_updates()
 
         # Now let's create a room, which will insert a membership
         user = UserID("alice", "test")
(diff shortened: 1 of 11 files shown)


tests/unittest.py
```diff
@@ -317,6 +317,15 @@ def wait_on_thread(self, deferred, timeout=10):
             self.reactor.advance(0.01)
             time.sleep(0.01)
 
+    def wait_for_background_updates(self) -> None:
+        """Block until we've written completed all background updates."""
```
clokep, line 321: I think there's a word missing or something?

```suggestion
        """Block until all background updates have completed."""
```

clokep (approved):
Seems reasonable to me!

tests/unittest.py
DMRobertson: Ahh, thank you. I had in mind something like "block until we've written to the DB that all background updates have completed". But I think the former is an unnecessary detail.

clokep: Yeah, I don't think the caller really cares about the details, just that they've all finished! (Also, thanks for abstracting this out!)

tests/storage/test_roommember.py
```diff
@@ -197,9 +192,4 @@ def test_can_rerun_update(self):
         self.store.db_pool.updates._all_done = False
```
DMRobertson, line 192: I did have one concern here. Some of these explicitly reset the `_all_done` flag, and others don't. I haven't made the situation worse---apart from an extra stack frame the behaviour should be identical. But I wondered if it was was more consistent to reset the flag in the helper function.

tests/unittest.py
```diff
@@ -317,6 +317,15 @@ def wait_on_thread(self, deferred, timeout=10):
             self.reactor.advance(0.01)
             time.sleep(0.01)
 
+    def wait_for_background_updates(self) -> None:
+        """Block until all background updates have completed."""
```
richvdh, line 321:
```suggestion
        """Block until all background database updates have completed."""
```

tests/storage/test_roommember.py
richvdh: :man_shrugging:

richvdh (approved):
lgtm

PR #10983, opened 2021-10-04
Log stack traces when a missing opentracing span is detected

squahtx:
Make it easier to track down where opentracing spans are going missing
by including stack traces in the logs.

### DIFF
--- a/synapse/logging/opentracing.py
+++ b/synapse/logging/opentracing.py
@@ -339,6 +339,7 @@ def ensure_active_span(message, ret=None):
                     "There was no active span when trying to %s."
                     " Did you forget to start one or did a context slip?",
                     message,
+                    stack_info=True,
                 )
 
                 return ret
--- a/changelog.d/10983.misc
+++ b/changelog.d/10983.misc
@@ -1,0 +1,1 @@
+Log stack traces when a missing opentracing span is detected.


synapse/logging/opentracing.py
```diff
@@ -339,6 +339,7 @@ def ensure_active_span_inner_2(*args, **kwargs):
                     "There was no active span when trying to %s."
                     " Did you forget to start one or did a context slip?",
                     message,
+                    stack_info=True,
```
DMRobertson, line 342: Nice---didn't know about this! (`traceback.print_stack` is my go-to)

DMRobertson (approved):
LGTM!

DMRobertson:
Does this want to be merged into the release branch? My understand is that this causes sentry noise rather than actual pain, so I'd assume just merge into develop as normal?

synapse/logging/opentracing.py
squahtx: sentry has a bug where each line of some stack traces gets turned into a separate issue. I'm hoping this makes it get filed correctly.

squahtx:
I think develop since it's not a priority

synapse/logging/opentracing.py
richvdh: that's a synapse/twisted bug rather than a sentry bug. Basically uncaught exceptions end up getting printed to `sys.stderr`, but synapse intercepts things written to `sys.stderr` and sends them back to the logger, at ERROR level, line by line; there they are of course intercepted by the sentry handler.

The correct solution would be to not have uncaught exceptions printed to `sys.stderr`, and just let the interpreter-level uncaught exception hook handle them (where sentry will correctly report them). Unfortunately that's happening within the bowels of Twisted, so the next best solution is to make sure there aren't any uncaught exceptions coming from our code.

squahtx: interesting, thanks for clearing things up!

PR #10984, opened 2021-10-04
Fix logging context warnings when losing replication connection

squahtx:
Instead of triggering `__exit__` manually on the replication handler's
logging context, use it as a context manager so that there is an
`__enter__` call to balance the `__exit__`.

### DIFF
--- a/changelog.d/10984.misc
+++ b/changelog.d/10984.misc
@@ -1,0 +1,1 @@
+Fix spurious "Expected logging context replication_command_handler-############### was lost" warnings when losing the replication connection.
--- a/synapse/replication/tcp/protocol.py
+++ b/synapse/replication/tcp/protocol.py
@@ -434,8 +434,9 @@ class BaseReplicationStreamProtocol(LineOnlyReceiver):
         if self.transport:
             self.transport.unregisterProducer()
 
-        # mark the logging context as finished
-        self._logging_context.__exit__(None, None, None)
+        # mark the logging context as finished by triggering `__exit__()`
+        with self._logging_context:
+            pass
 
     def __str__(self):
         addr = None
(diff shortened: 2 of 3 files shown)


changelog.d/10984.misc
```diff
@@ -0,0 +1 @@
+Fix spurious "Expected logging context replication_command_handler-############### was lost" warnings when losing the replication connection.
```
babolivier, line 1:
```suggestion
Fix spurious warnings about losing the logging context on the `ReplicationCommandHandler` when losing the replication connection.
```

babolivier (approved):

clokep:
I wonder if it would make more sense to ensure that `__enter__` is called when it is created? (Are there any issues with it being called later?)

squahtx:
> I wonder if it would make more sense to ensure that `__enter__` is called when it is created? (Are there any issues with it being called later?)

I'm not quite sure I understand the suggestion. Do you mean we that we could a) `__enter__` the `LoggingContext` in the `RedisSubscriber` constructor, or that b) all `LoggingContexts` should be `__enter__`ed when constructed?

a) creates problems because the logging context would span random bits of code which we don't want to include in CPU stats. Right now the logging context only measures CPU time when processing a replication command.

b) turns out to be true but I can't see how to enforce it. Some digging reveals that `LoggingContext`s capture the previous logging context on construction, and restore it on `__exit__`. Which means that it's not necessarily correct to stash away a `LoggingContext` and use it later, since the restored logging context may not be the one that was active on `__enter__`. Luckily we only do this kind of stashing in 4 places that I can see: the two instances in this PR, in `SynapseRequest.render` and in the `Measure` context manager.

I'll have to update the PR in light of b).

PR #10985, opened 2021-10-04
Require direct references to configuration variables.

clokep:
Builds on #10798, #10812, #10885, #10893, #10897, #10916, and #10959.

This was done by starting with the last commit, then running unit tests repeatedly until they passed. 😄 Commits are reviewable individually. The first commit is changes which were missed in the above PRs (and a couple which pointed to the wrong configuration class!) There's then a few commits of refactoring. Then we fix-up the configuration classes themselves to contact each other via the `RootConfig` object.

Finally, we can remove the `__getattr__` magic that allows not specifying a configuration class directly.

I'm fairly confident I found all references, but not 100% sure.


clokep:
Any alternate to the last commit could also be to add some more deprecation code in that would warn about properties being accessed?

squahtx:
code changes LGTM

I can't comment on whether it's likely to break third party modules. I don't see why we shouldn't just make the change if the config isn't part of the official module API. Adding in deprecation warnings sounds like extra hassle.

synapse/app/homeserver.py
```diff
@@ -234,7 +234,7 @@ def _configure_named_resource(self, name, compress=False):
             )
 
         if name in ["media", "federation", "client"]:
-            if self.config.media.enable_media_repo:
+            if self.config.server.enable_media_repo:
```
babolivier, line 237: I might be missing context here, but why was this line (wrongly) trying to get the setting from the `media` config? Ditto for the `account_validity` vs `email` config.

Also, and not something that needs to be fixed now, I find it a bit worrying if we have been looking for this setting in the wrong config class for some time and CI didn't pick it up.

clokep: > I might be missing context here, but why was this line (wrongly) trying to get the setting from the `media` config? Ditto for the `account_validity` vs `email` config.

In one of the previous PRs I updated it to look at the wrong configuration.

> Also, and not something that needs to be fixed now, I find it a bit worrying if we have been looking for this setting in the wrong config class for some time and CI didn't pick it up.

It still worked fine before this PR, which is why CI didn't pick it up. It is due to this code:

https://github.com/matrix-org/synapse/blob/94b620a5edd6b5bc55c8aad6e00a11cc6bf210fa/synapse/config/_base.py#L121-L134

(Which I need to double check is removed by this PR actually...)

clokep:
> I'd be tempted to say that modules that don't follow the rules should be expected to break at some point, though off the back of that we should probably update our documentation to make things like this super explicit (I can take that action as I have more work around module documentation lined up in the next couple of weeks).

I think the only way to do this would be to do `module_api._hs.config`, which is pretty obviously reaching into things that aren't meant to be exposed IMO.

synapse/app/homeserver.py
clokep: Yep, I removed that! 🎉

PR #10986, opened 2021-10-04
Hoist `cache_joined_hosts_for_event` to caller

richvdh:
`_check_event_auth` is only called in two places, and only one of those sets `send_on_behalf_of`. Warming the cache isn't really part of auth anyway, so moving it out makes a lot more sense.

### DIFF
--- a/synapse/handlers/federation_event.py
+++ b/synapse/handlers/federation_event.py
@@ -356,6 +356,11 @@ class FederationEventHandler:
                 )
 
         # all looks good, we can persist the event.
+
+        # First, precalculate the joined hosts so that the federation sender doesn't
+        # need to.
+        await self._event_creation_handler.cache_joined_hosts_for_event(event, context)
+
         await self._run_push_actions_and_persist_event(event, context)
         return event, context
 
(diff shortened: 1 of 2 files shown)


synapse/handlers/federation_event.py
```diff
@@ -1291,17 +1296,10 @@ async def _check_event_auth(
         except AuthError as e:
             logger.warning("Failed auth resolution for %r because %s", event, e)
             context.rejected = RejectedReason.AUTH_ERROR
+            return context
```
richvdh, line 1299: nowhere else in this method sets `context.rejected` (and it can't be set before entry), so this is equivalent to the removed `if not context.rejected` line.

synapse/handlers/federation_event.py
```diff
@@ -356,6 +356,11 @@ async def on_send_membership_event(
                 )
 
         # all looks good, we can persist the event.
+
+        # First, precalculate the joined hosts so that the federation sender doesn't
+        # need to.
+        await self._event_creation_handler.cache_joined_hosts_for_event(event, context)
```
clokep, line 362: Previously we would do this even if `context.rejected` was truth-y. I suspect that wasn't helpful since the event is rejected, but wanted to point this out as a change in behavior.

Also, I find the comment a bit odd "fIrst...", but there's no follow-up (no "then" / "next" / "second"). 😄

richvdh: > Previously we would do this even if context.rejected was truth-y. I suspect that wasn't helpful since the event is rejected, but wanted to point this out as a change in behavior.

yeah, it's deliberate, but I think the only difference is that we won't be populating a cache entry that will never be used.

> Also, I find the comment a bit odd "fIrst...", but there's no follow-up (no "then" / "next" / "second")

mmhmm. it's in reference to the line above: "we can persist the event ... But first, precalculate the joined hosts".

clokep: Fair! Not a huge deal, just felt like I was left with a cliffhanger!

PR #10988, opened 2021-10-04
`_check_event_auth`: move event validation earlier

richvdh:
There's little point in doing a fancy state reconciliation dance if the event
itself is invalid.

Likewise, there's no point checking it again in `_check_for_soft_fail`.

### DIFF
--- a/synapse/handlers/federation_event.py
+++ b/synapse/handlers/federation_event.py
@@ -1250,8 +1250,17 @@ class FederationEventHandler:
         # This method should only be used for non-outliers
         assert not event.internal_metadata.outlier
 
+        # first of all, check that the event itself is valid.
         room_version = await self._store.get_room_version_id(event.room_id)
         room_version_obj = KNOWN_ROOM_VERSIONS[room_version]
+
+        try:
+            validate_event_for_room_version(room_version_obj, event)
+        except AuthError as e:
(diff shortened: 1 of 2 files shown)


synapse/handlers/federation_event.py
```diff
@@ -1250,9 +1250,18 @@ async def _check_event_auth(
...
         room_version = await self._store.get_room_version_id(event.room_id)
         room_version_obj = KNOWN_ROOM_VERSIONS[room_version]
 
+        try:
+            validate_event_for_room_version(room_version_obj, event)
+        except AuthError as e:
+            logger.warning("While validating received event %r: %s", event, e)
+            # TODO: use a different rejected reason here?
+            context.rejected = RejectedReason.AUTH_ERROR
+            return context
```
clokep, line 1263: Looks like moving this up means that we won't call `cache_joined_hosts_for_event` in some situations, but I see #10986 moves it out of this method anyway.

clokep (approved):
I think this is fine, I don't think any of the methods called have side-effects that affect the `event`.

synapse/handlers/federation_event.py
richvdh: yeah. We shouldn't be calling `cache_joined_hosts_for_event` on rejected events anyway, since we won't be sending them out.

PR #10992, opened 2021-10-05
Run CI with Python 3.10 and Postgres 14

DMRobertson:
Python 3.10 was [released yesterday](https://www.python.org/downloads/release/python-3100/), and the `setup-python` action supports it [as of three hours ago](https://github.com/actions/python-versions/releases/tag/3.10.0-117470).

Postgres 14 was [released at the end of September](https://www.postgresql.org/docs/14/release-14.html) and the dockerhub container [is at least two hours old](https://hub.docker.com/layers/postgres/library/postgres/14.0/images/sha256-da98eb3d0ac42daede96c9cdf77ac57ef8e3c50614f48fcafe55e691b7d9a590?context=explore).

Bump the CI jobs that run on the latest versions of python and postgres to use these. (Should we use postgres:latest rather than pinning a specific number?)

Also include an explicit matrix job for trial under Python 3.10.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -76,11 +76,11 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.6", "3.7", "3.8", "3.9"]
+        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10"]
         database: ["sqlite"]
         include:
           # Newest Python without optional deps
-          - python-version: "3.9"
+          - python-version: "3.10"
             toxenv: "py-noextras,combine"
 
           # Oldest Python with PostgreSQL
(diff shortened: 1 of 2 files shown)


squahtx (approved):
LGTM if CI passes!

richvdh:
are we going to claim support for python 3.10 in https://matrix-org.github.io/synapse/develop/setup/installation.html#installing-as-a-python-module-from-pypi and https://github.com/matrix-org/synapse/blob/develop/setup.py#L144?

DMRobertson:
Maybe slightly paranoid, but my gut says we should try running a homeserver proper (and not just tests) under 3.10 before doing that.

richvdh:
oh right, looking at this more closely, it looks like we're not running sytest or complement against a python3.10-synapse. Yes, we should do that first.

DMRobertson:
> oh right, looking at this more closely, it looks like we're not running sytest or complement against a python3.10-synapse. Yes, we should do that first.

IIUC we use the Python + Postgres versions that come OOTB in various Debian and Ubuntu releases. So I think we'd need to wait for a sufficiently new release of those? (Though maybe debian testing will contain [Python 3.10](https://packages.debian.org/bookworm/python3.10) and [Postgres 14](https://packages.debian.org/bookworm/database/postgresql-14) soon? Not sure of the mechanics)

danger89:
So ready to upgrade? https://github.com/matrix-org/synapse/issues/13234

PR #10995, opened 2021-10-05
Catch AttributeErrors when calling registerProducer

babolivier:
Looks like this was missed in https://github.com/matrix-org/synapse/pull/10932 despite said PR mentioning it.

### DIFF
--- a/synapse/http/server.py
+++ b/synapse/http/server.py
@@ -563,7 +563,7 @@ class _ByteProducer:
 
         try:
             self._request.registerProducer(self, True)
-        except RuntimeError as e:
+        except (AttributeError, RuntimeError) as e:
             logger.info("Connection disconnected before response was written: %r", e)
 
             # We drop our references to data we'll not use.
--- a/changelog.d/10995.bugfix
+++ b/changelog.d/10995.bugfix
@@ -1,0 +1,1 @@
+Correct a bugfix introduced in Synapse v1.44.0 that wouldn't catch every error of the connection breaks before a response could be written to it.


DMRobertson:
Twisted [says](https://twistedmatrix.com/documents/current/api/twisted.internet.interfaces.IConsumer.html):

> Raises RuntimeError: If a producer is already registered.

I think that means the error as written is misleading, if it catches a RuntimeError?

babolivier:
`_request.registerProducer` calls `_request.channel.registerProducer`, which might return a `RuntimeError` in some cases. However, it doesn't account for the fact that `_request.channel` might be `None`, which is what Twisted sets it to when it loses the connection. This means that calling `registerProducer` on it will raise an `AttributeError`.

DMRobertson:
> `_request.registerProducer` calls `_request.channel.registerProducer`, which might return a `RuntimeError` in some cases. However, it doesn't account for the fact that `_request.channel` might be `None`, which is what Twisted sets it to when it loses the connection. This means that calling `registerProducer` on it will raise an `AttributeError`.

Sorry, I phrased that poorly. I understand that an `AttributeError` can be raised and we should explicitly catch it here. But suppose instead that `_request.channel.registerProducer` raises a RuntimeError because we've called it twice. With this change we'll still catch the RuntimeError here and log an error in exactly the same way as if there were an AttributeError. I'm not sure if that's what we want.

babolivier:
> With this change we'll still catch the RuntimeError here and log an error in exactly the same way as if there were an AttributeError

No, since we also log the error's message.

DMRobertson:
> > With this change we'll still catch the RuntimeError here and log an error in exactly the same way as if there were an AttributeError
> 
> No, since we also log the error's message.

Oh, I see. But in that case (devil's advocate), why not catch every `Exception`?

babolivier:
> > > With this change we'll still catch the RuntimeError here and log an error in exactly the same way as if there were an AttributeError
> > 
> > 
> > No, since we also log the error's message.
> 
> Oh, I see. But in that case (devil's advocate), why not catch every `Exception`?

Because it's usually bad practice and I prefer avoiding that when I can. If we fail for a reason we don't expect we shouldn't hide it in a warning we might miss (and we won't feed back into sentry).

synapse/http/server.py
```diff
@@ -563,7 +563,11 @@ def __init__(
 
         try:
             self._request.registerProducer(self, True)
-        except RuntimeError as e:
+        except (AttributeError, RuntimeError) as e:
+            # Twisted will raise a RuntimeError in some cases.
```
clokep, line 567: Do we know what cases?

clokep (approved):
I think this seems OK. I wondered if it would be clearer to do an explicit check for `self._request.channel` first, but I think that would make the code flow more confusing.

PR #10996, opened 2021-10-05
Fix opentracing and Prometheus metrics for replication requests

squahtx:
This commit fixes two bugs to do with decorators not instrumenting
`ReplicationEndpoint`'s `send_request` correctly. There are two
decorators on `send_request`: Prometheus' `Gauge.track_inprogress()`
and Synapse's `opentracing.trace`.

`Gauge.track_inprogress()` does not have any support for async
functions when used as a decorator. Since async functions behave like
regular functions that return coroutines, only the creation of the
coroutine was covered by the metric and none of the actual body of
`send_request`.

`Gauge.track_inprogress()` returns a regular, non-async function
wrapping `send_request`, which is the source of the next bug.
The `opentracing.trace` decorator would normally handle async functions
correctly, but since the wrapped `send_request` is a non-async function,
the decorator ends up suffering from the same issue as
`Gauge.track_inprogress()`: the opentracing span only measures the
creation of the coroutine and none of the actual function body.

Using `Gauge.track_inprogress()` as a context manager instead of a
decorator resolves both bugs.

### DIFF
--- a/changelog.d/10996.misc
+++ b/changelog.d/10996.misc
@@ -1,0 +1,1 @@
+Fix a bug introduced in Synapse 1.21.0 that causes opentracing and Prometheus metrics for replication requests to be measured incorrectly.
--- a/synapse/replication/http/_base.py
+++ b/synapse/replication/http/_base.py
@@ -182,85 +182,87 @@ class ReplicationEndpoint(metaclass=abc.ABCMeta):
             )
 
         @trace(opname="outgoing_replication_request")
-        @outgoing_gauge.track_inprogress()
         async def send_request(*, instance_name="master", **kwargs):
-            if instance_name == local_instance_name:
-                raise Exception("Trying to send HTTP request to self")
-            if instance_name == "master":
-                host = master_host
-                port = master_port
-            elif instance_name in instance_map:
-                host = instance_map[instance_name].host
-                port = instance_map[instance_name].port
-            else:
-                raise Exception(
-                    "Instance %r not in 'instance_map' config" % (instance_name,)
+            with outgoing_gauge.track_inprogress():
+                if instance_name == local_instance_name:
+                    raise Exception("Trying to send HTTP request to self")
+                if instance_name == "master":
+                    host = master_host
+                    port = master_port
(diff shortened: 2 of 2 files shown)


squahtx:
The only changes are the removal of a decorator and the addition of a context manager.

Clicking the cog and turning on "Hide whitespace changes" may help with the diff

DMRobertson:
Nice spot! Is there a way to stop ourselves from making this mistake in the future? Maybe we could change `opentracing.trace` to warn if it ends up decorating a synchronous function?

squahtx:
> Maybe we could change `opentracing.trace` to warn if it ends up decorating a synchronous function?

Good idea!

synapse/logging/opentracing.py
```diff
@@ -807,6 +807,13 @@ def err_back(result):
                         result.addCallbacks(call_back, err_back)
 
                     else:
+                        if inspect.iscoroutine(result):
+                            logger.error(
+                                "@trace did not wrap %s correctly! "
+                                "The function is not async but returns a coroutine.",
+                                func.__qualname__,
+                            )
```
squahtx, line 815: Should this be an error or warning?

I chose error because I'd like it to show up in sentry. On the other hand it doesn't impact normal operation of the homserver.

clokep: Should this raise an error since it is a programming error that shouldn't occur?

clokep: Maybe an error would be more appropriate then. (Note that we do seem to handle coroutines properly above, so this would only happen if we have a function which returns an awaitable, but is not a coroutine. We do have some of those though...)

squahtx: Alright, I'll change it to an assert then.

clokep: Sorry, I wasn't clear. I meant "maybe logging an error is more appropriate". As in -- don't change anything.

I don't feel strongly either way.

Also, shouldn't we be checking via `inspect.isawaitable`, not `iscoroutine` here?

Issue #10083, opened 2021-05-27
Unexpectedly found that events don't have chain IDs


daenney:
### Description

```
synapse.storage.databases.main.event_federation - 369 - INFO - persist_events-17250 - Unexpectedly found that events don't have chain IDs in room !OGEhHVWSdvArJzumhm:matrix.org: {'$e3j8FpK-lcibZO57q50gANeLSK0dptpF6fCUmBF3rKU', '$uIBJZEYyEtATjznpOiBVPlazSR_PhVKurtsLpTCEmDA', '$Om8XNzoCfBmCG36Zx24Y9WFHRy2RPVQDDBbN8CvSq74', '$7_ZXqJstZZrXU1S-2jgKaEOpgrvWEXdU_NxQ6mxC4_8'} 
```

And logs like:

```
PUT-638717-!OGEhHVWSdvArJzumhm:matrix.org-$OOjkuWpVS-iTT7joyJUx3UDTpKyPDPbdg2YvyfltkWY - auth_event id $GJ4kBt8ivplflUSXy6NB_tYANh98Pi_5hZ2skUeq0v4 for event $uIBJZEYyEtATjznpOiBVPlazSR_PhVKurtsLpTCEmDA is missing
```

### Steps to reproduce

- Look in the logs :disappointed: 

### Version information

- **Homeserver**: Other, g33k.se

If not matrix.org:

- **Version**: 1.34.0

- **Install method**: Debian packages from Element

- **Platform**: Ubuntu 20.04

daenney:
Other folks noticed this too like @Half-Shot @MatMaul and a similar issue has been reported for EMS hosts too by @benbz. @richvdh mentioned it's likely related to #9595

richvdh:
This also causes `send_join` requests to take tens of minutes of CPU time.

richvdh:
@daenney could you do me a favour and run `select * from rejections where event_id='$GJ4kBt8ivplflUSXy6NB_tYANh98Pi_5hZ2skUeq0v4';`

That should help confirm if it's due to #9595.

daenney:
`0 rows retrieved in 63 ms` is what that query got me.

richvdh:
hrm, so it's a slightly different cause to #9595 :/

bdelwood:
I am seeing the same issue that's killing Matrix HQ for me. 

Here are some of the relevant log lines
```
[...]

PR #11001, opened 2021-10-06
Check auth on received events' auth_events

### LINKED ISSUE #6643: Event auth rules are not correctly followed
As per https://github.com/matrix-org/synapse/issues/6605#issuecomment-569832078:

Different code paths use different values of `auth_events` when performing event auth: some use a locally-calculated set of events, whereas others use the auth events provided by the event itself.

The [spec](https://matrix.org/docs/spec/server_server/r0.1.3#checks-performed-on-receipt-of-a-pdu) says:

> The receiving server must ensure that the event:
> ...
> * Passes authorization rules based on the event's auth events, otherwise it is rejected.
> * Passes authorization rules based on the state at the event, otherwise it is rejected.

This is not happening correctly, which allows malformed events into the state of the room.

richvdh:
Currently, when we receive an event whose auth_events differ from those we expect, we state-resolve between the two state sets, and check that the event passes auth based on the resolved state.

This means that it's possible for us to accept events which don't pass auth at their declared auth_events (or where the auth events themselves were rejected), leading to problems down the line like https://github.com/matrix-org/synapse/issues/10083.

This change means we will:
 * ignore any events where we cannot find the auth events
 * reject any events whose auth events were rejected
 * reject any events which do not pass auth at their declared auth_events.

Together with a whole raft of previous work, this is a partial fix to #9595.

Fixes #6643.

Based on #11009.

### DIFF
--- a/changelog.d/11001.bugfix
+++ b/changelog.d/11001.bugfix
@@ -1,0 +1,1 @@
+ Fix a long-standing bug which meant that events received over federation were sometimes incorrectly accepted into the room state.


synapse/handlers/federation_event.py
```diff
@@ -1466,6 +1490,9 @@ async def _update_auth_events_and_context_for_auth(
         # if we have missing events, we need to fetch those events from somewhere.
         #
         # we start by checking if they are in the store, and then try calling /event_auth/.
+        #
+        # TODO: this code is now redundant, since it should be impossible for us to
+        #   get here without already having the auth events.
```
richvdh, line 1495: I'm going to clean this up in another PR.

DMRobertson:
Sytest failing `Inbound federation correctly handles soft failed events as extremities`. Not sure if that's expected?

richvdh:
it's not expected - I'm fixing up the sytest.

PR #11002, opened 2021-10-06
Don't alter directory entries for local users when setting a per-room nickname

DMRobertson:
This goes some way toward fixing #5677. However:

- We're still leaking per-room profiles for remote users
- There are other leaks possible for local users. I've got one PR coming up for that.
- I'd still like to tidy up this code in other ways


DMRobertson:
> > There are other leaks possible for local users. I've got one PR coming up for that.
> 
> Is one of those that `_handle_new_user` also seems to update the profile?

Ding ding ding!

PR #11003, opened 2021-10-06
Fix overwriting profile when making room public

DMRobertson:
Depends on #11002. Suggest reviewing commit-by-commit (I tried to be disciplined).


synapse/handlers/user_directory.py
```diff
@@ -229,25 +229,22 @@ async def _handle_deltas(self, deltas: List[Dict[str, Any]]) -> None:
...
-                            continue
-
-                        profile = ProfileInfo(
-                            avatar_url=event.content.get("avatar_url"),
-                            display_name=event.content.get("displayname"),
-                        )
-
-                        await self._handle_new_user(room_id, state_key, profile)
+                        if not self.is_mine_id(state_key):
+                            await self._upsert_directory_entry_for_remote_user(
```
clokep, line 244: I haven't really looked at it, but I wonder if this could also re-use `_handle_profile_change`?

PR #11005, opened 2021-10-06
Remove the deprecated BaseHandler.

clokep:
Each commit removes the deprecated `BaseHandler` class which some handlers still used as a super-class. Most of them simply needed to access a couple of properties (`store`, `server_name`, etc.)

There was a few places where the handlers depended on the `ratelimit` method -- I moved this out to a separate class which is accessible via `HomeServer.get_request_ratelimiter()`.

Should be reviewable commit-by-commit, this should be non-functional overall.


clokep:
> There're some remaining mentions of `BaseHandler` in `docs/architecture.md`. It's explicitly dated 2014 so it may not be worth updating as part of this PR.

Good shout, but I'm not sure that's worth removing from there. I think we should probably update that document in general.

PR #11006, opened 2021-10-06
Relax `ignore-missing-imports` for modules that have stubs now

DMRobertson:
I searched through typeshed to see if there were any of these that we could remove. I found three modules.

This didn't find any interesting errors. Feels like it was worth doing though.

I noticed our mypy and mypy-zope versions are pinned to a specific minor release. Is it worth bumping these?

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -226,9 +226,6 @@ ignore_missing_imports = True
 [mypy-josepy.*]
 ignore_missing_imports = True
 
-[mypy-jsonschema]
-ignore_missing_imports = True
-
 [mypy-jwt.*]
 ignore_missing_imports = True
 
--- a/mypy.ini
+++ b/mypy.ini
@@ -192,6 +192,13 @@ disallow_untyped_defs = True
 [mypy-tests.storage.test_user_directory]
 disallow_untyped_defs = True
 
+;; Dependencies without annotations
+;; Before ignoring a module, check to see if type stubs are available.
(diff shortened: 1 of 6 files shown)


clokep:
> I noticed our mypy and mypy-zope versions are pinned to a specific minor release. Is it worth bumping these?

Yes!

mypy.ini
```diff
@@ -192,98 +192,97 @@ disallow_untyped_defs = True
 [mypy-tests.storage.test_user_directory]
 disallow_untyped_defs = True
 
-[mypy-pymacaroons.*]
-ignore_missing_imports = True
+;; Dependencies without annotations
+;; Before ignoring a module, check to see if type stubs are available.
+;; The `typeshed` project maintains stubs here:
+;;     https://github.com/python/typeshed/tree/master/stubs
+;; and for each pacakge `foo` there's a corresponding `types-foo` package on PyPI,
```
clokep, line 199:
```suggestion
;; and for each package `foo` there's a corresponding `types-foo` package on PyPI,
```

DMRobertson:
Let's see what CI makes of this. Mypy was fine running locally but... I have faffed with dependencies and maybe screwed something up.

DMRobertson:
> > I noticed our mypy and mypy-zope versions are pinned to a specific minor release. Is it worth bumping these?
> 
> Yes!

Done in the most recent batch of changes.

PR #11022, opened 2021-10-07
disallow-untyped-defs for synapse.replication.http

DMRobertson:


DMRobertson:
I think this is bitrotting and didn't get very far anyway.

PR #11029, opened 2021-10-08
disallow-untyped-defs for the module_api

DMRobertson:
Resurrected this from about a month ago. Merge wasn't too bad.

### DIFF
--- a/synapse/module_api/__init__.py
+++ b/synapse/module_api/__init__.py
@@ -150,12 +150,6 @@ class UserIpAndAgent:
     last_seen: int
 
 
-# This is sad: I'd like to express that the arguments to this Callable are
-# themselves callable. "Callback protocols" let you do this, but I couldn't find
-# a way to do so that mypy agreed was type safe.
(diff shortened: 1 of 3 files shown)


synapse/module_api/__init__.py
```diff
@@ -146,6 +150,12 @@ class UserIpAndAgent:
     last_seen: int
 
 
+# This is sad: I'd like to express that the arguments to this Callable are
+# themselves callable. "Callback protocols" let you do this, but I couldn't find
+# a way to do so that mypy agreed was type safe.
+RegisterCallbacks = Callable[..., None]
```
squahtx, line 156: Is this used anywhere?

squahtx (approved):
lgtm, one question

synapse/module_api/__init__.py
DMRobertson: ... no. Can't remember where I wanted to use it. Maybe there were changes elsewhere that made this redundant.

PR #11030, opened 2021-10-08
Document fail-open behaviour of `check_event_allowed` module API callback

reivilibre:
Signed-off-by: Olivier Wilkinson (reivilibre) <oliverw@matrix.org>

With my module developer hat on, this behaviour didn't seem expected.

It's probably not my say whether we should change this (I guess it's been thought through already) but it should at least be documented, I think.


reivilibre:
So it seems the current behaviour **is** wrong — issue to be opened imminently.

PR #11036, opened 2021-10-08
Reset global cache state before cache tests

### LINKED ISSUE #11007: `test_global_instantiated_before_config_load` is flaky
https://github.com/matrix-org/synapse/runs/3816535717 and https://github.com/matrix-org/synapse/runs/3818555537

This was recently rewritten in #10985.

clokep:
Fixes #11007 hopefully.

From looking at the logs of a bad test run I think what is happening is:

* A test runs that configures caches in someway.
* Our test runs which expects no logging configuration to be done yet.
* The test fails because the resize callback kicks in when we don't expect it to.

~~I'm not really sure how this ever worked, but I think the proper fix is to patch out more global state.~~

This *used* to work since state was reset *before* tests ran, instead of *after* they ran. I made an incorrect assumption that other tests (outside of this file) wouldn't touch this global state.

This reverts #11019 and structures the code a bit more like it was before #10985.

### DIFF
--- a/tests/config/test_cache.py
+++ b/tests/config/test_cache.py
@@ -12,25 +12,23 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from unittest.mock import patch
-
 from synapse.config.cache import CacheConfig, add_resizable_cache
 from synapse.util.caches.lrucache import LruCache
 
 from tests.unittest import TestCase
 
 
-# Patch the global _CACHES so that each test runs against its own state.
-@patch("synapse.config.cache._CACHES", new_callable=dict)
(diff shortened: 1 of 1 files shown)


clokep:
I think this will fix it, but I'm not sure why this is flakey, as opposed to failing every time? Maybe it has to do with the order the tests run in (or which tests get split into separate processes when doing `--jobs=2`).

In the CI run I traced, the `resize_all_caches_func` was configured from the `tests.app.test_phone_stats_home.PhoneHomeTestCase.test_r30_user_must_be_retained_for_at_least_a_month` test, so that leaked state.

PR #11045, opened 2021-10-11
Fix race in `MultiWriterIdGenerator`

### LINKED ISSUE #9424: Event got sent but not synced back down everywhere
The event in question is https://app.element.io/#/room/!RbKRuLTvHIhArQcvjr:matrix.org/$1613561441289661rpDpF:matrix.org?via=matrix.org&via=vector.modular.im&via=t2l.io

The client that sent the event (element desktop) and my dekstop hydrogen running alongside did not get the event from sync.

On element desktop the event was black (e.g. sent) but stuck to the bottom of the room, and on hydrogen I check the storage and it showed that there was no limited sync for that room around those messages, and there is no trace of that event anywhere.

When opening my phone hydrogen an hour or so later, the event was synced down to it in the correct order.

At least T3chguy and Ryan did receive the event in the correct order.

erikjohnston:
The race allowed the current position to advance too far when stream IDs
are still being persisted.

This happened when it received a new stream ID from a remote write
between a new stream ID being allocated and it being added to the set of
unpersisted stream IDs.

Fixes #9424.

### DIFF
--- a/synapse/storage/util/id_generators.py
+++ b/synapse/storage/util/id_generators.py
@@ -542,17 +542,17 @@ class MultiWriterIdGenerator:
                 if self._unfinished_ids and self._in_flight_fetches:
                     # `_in_flight_fetches` stores the maximum safe stream ID, so
                     # we add one to make it equivalent to the minimum unsafe ID.
-                    min_unfinshed = min(
+                    min_unfinished = min(
                         self._unfinished_ids[0], self._in_flight_fetches[0] + 1
                     )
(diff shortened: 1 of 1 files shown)


erikjohnston:
> Looks like a subtle and fiddley problem. How hard would it be to write a test case that mimics or reproduces the problem?

I've been looking into this, the problem is that it looks like we run the DB "threads" actually in our main thread, so adding locks to simulate delays causes the test to deadlock

PR #11070, opened 2021-10-13
Move some config out of the server section.

clokep:
Moves:

* `experimental_msc1849_support_enabled` from the server to the experimental config.
* Creates a new config object for retention.

The overall goal was just to slim down the `server` config a bit.


clokep:
> Do we want to update docs/upgrade.md to tell server admins to restructure their configs?

There's nothing to restructure, this is an internal change. It does have the unfortunate side-effect of moving some bits of the sample config around, but ordering there doesn't really matter.

PR #11075, opened 2021-10-13
Correctly exclude users when making a room public or private

DMRobertson:
Suppose you have a user in a room who is excluded from the user directory. If they change the room from private to public, they are added to the directory, making them no longer excluded.

I think this is a rare and unlikely thing, but it's another step towards consistency.

I sanity checked [the code in version 1.40](https://github.com/matrix-org/synapse/blob/release-v1.40/synapse/handlers/user_directory.py) before I started all these changes. As far as I can see this bug was there and has probably been around for a while. We have tried to exclude support users since at least three years ago (https://github.com/matrix-org/synapse/commit/d2f7c4e6b1efbdd3275d02a19220a10cf00a8f66) and so I think this is just another oversight.

### DIFF
--- a/tests/handlers/test_user_directory.py
+++ b/tests/handlers/test_user_directory.py
@@ -250,7 +250,7 @@ class UserDirectoryTestCase(unittest.HomeserverTestCase):
         self.helper.join(room, bob, tok=bob_token)
 
         # The user directory should reflect this.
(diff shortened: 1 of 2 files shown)


clokep:
Looks good! Thanks for the back and forth on what this is doing!

PR #11077, opened 2021-10-13
Attempt different character encodings when previewing a URL.

clokep:
There's a variety of issues in Sentry (see https://sentry.matrix.org/sentry/synapse-matrixorg/?query=is%3Aunresolved+unicodedecodeerror) about being unable to decode a page as UTF-8. At least some of these pages are `windows-1252` instead of `utf-8`, but need to be guessed as that encoding.

This adapts [the algorithm of BeautifulSoup's dammit module](https://bazaar.launchpad.net/~leonardr/beautifulsoup/bs4/view/head:/bs4/dammit.py#L2565) for URL previews, in particular this means we fully decode the bytes to a string *first* than ask lxml to parse it. We attempt different encodings until we find one which works.

I wonder if we should just be using BeautifulSoup instead of doing this ourselves.

An example URL which should now work (and previously break) is https://www.bibliotecapleyades.net/tierra_hueca/esp_tierra_hueca_2d.htm

### DIFF
--- a/synapse/rest/media/v1/preview_url_resource.py
+++ b/synapse/rest/media/v1/preview_url_resource.py
@@ -295,7 +295,7 @@ class PreviewUrlResource(DirectServeJsonResource):
             with open(media_info.filename, "rb") as file:
                 body = file.read()
 
-            tree = decode_body(body, media_info.media_type)
+            tree = decode_body(body, media_info.uri, media_info.media_type)
             if tree is not None:
                 # Check if this HTML document points to oEmbed information and
                 # defer to that.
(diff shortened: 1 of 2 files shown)


squahtx:
> Yep, this is potentially an issue, but I don't think decoding the bytes should take _too_ long? (I'm assuming you're worried about CPU load?)

Yep, that was my concern. From some quick tests, 10 MiB takes ~5ms when decoding as utf-8 and ~30ms when decoding as windows-1252.

I definitely think it's alright for now, considering BeautifulSoup does the same.

PR #11078, opened 2021-10-13
Fix broken export-data admin command and add a test for it to CI

### LINKED ISSUE #10908: Export data admin command is broken.
Running `python -m synapse.app.admin_cmd -c <config_file> export-data <user_id>` results in:

```
  File "/usr/local/lib/python3.8/runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/local/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/usr/local/lib/python3.8/site-packages/synapse/app/admin_cmd.py", line 236, in <module>
    start(sys.argv[1:])
  File "/usr/local/lib/python3.8/site-packages/synapse/app/admin_cmd.py", line 190, in start
    and not config.worker_log_file
  File "/usr/local/lib/python3.8/site-packages/synapse/config/_base.py", line 340, in __getattr__
    return self._get_unclassed_config(None, item)
  File "/usr/local/lib/python3.8/site-packages/synapse/config/_base.py", line 363, in _get_unclassed_config
    raise AttributeError(item, "not found in %s" % (list(self._configs.keys()),))
AttributeError: ('worker_log_file', "not found in ['modules', 'server', 'tls', 'federation', 'caches', 'database', 'logging', 'ratelimiting', 'media', 'oembed', 'captcha', 'voip', 'registration', 'account_validity', 'metrics', 'api', 'appservice', 'key', 'saml2', 'oidc', 'cas', 'sso', 'jwt', 'auth', 'email', 'authproviders', 'push', 'spamchecker', 'room', 'groups', 'userdirectory', 'consent', 'stats', 'servernotices', 'roomdirectory', 'thirdpartyrules', 'tracing', 'worker', 'redis', 'experimental']")
```

The effect of this is that its currently not possible to export a users data, e.g due to a GDPR request.

c.f. #10900

H-Shay:
As the title states. Fixes #10908. I added a test script to CI that ensures that the command keeps working. 

One note on how I solved this issue: one of the places it was broken was in `filter_events_for_client`, where we attempted to call `get_retention_policy_for_room` on the worker storage, but the storage didn't have access to that function (it's a cached function, and I gathered that we try not to add cached functions to worker storage). However, I found I could avoid that call if I set the `filter_send_to_client` param  to `False` rather than the default, which is `True`. The thinking behind this was that it was a less invasive way of solving the problem and that since the command was an admin command, it was probably okay to send the user data that we had in the database that was being requested. 

For background, the purpose of `filter_send_to_client` is to make sure that data in the room that had passed the date of the room retention policy was not show to clients in the time between when the retention policy expired and it is deleted.

### DIFF
--- a/.ci/scripts/test_export_data_command.sh
+++ b/.ci/scripts/test_export_data_command.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-# Test for the export-data admin command
+# Test for the export-data admin command against sqlite and postgres
 
 set -xe
 cd `dirname $0`/../..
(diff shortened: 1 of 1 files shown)


clokep:
> One note on how I solved this issue: one of the places it was broken was in `filter_events_for_client`, where we attempted to call `get_retention_policy_for_room` on the worker storage, but the storage didn't have access to that function (it's a cached function, and I gathered that we try not to add cached functions to worker storage).

This is already on the worker store though, so I think we could add `RoomWorkerStore` to the `AdminCmdSlavedStore` class used by the command?

PR #11088, opened 2021-10-14
Add a thread relation type per MSC3440.

### LINKED ISSUE #11085: Support MSC3440 thread relations
[MSC3440](https://github.com/matrix-org/matrix-doc/pull/3440) suggests implementing a new relation (`m.thread`, which uses `io.element.thread` as an unstable version). In order to implement this we'll need to:

* [ ] Support storing the new relation into the database (since we seem to filter on `rel_type` before storing them).
* [ ] Include a summary of the thread when bundling relations.

clokep:
Adds a thread relation type and the associated aggregation. It also includes a minor fix to not include empty keys when using `/send_relation`.

Fixes #11085

### DIFF
--- a/synapse/storage/databases/main/relations.py
+++ b/synapse/storage/databases/main/relations.py
@@ -293,7 +293,7 @@ class RelationsWorkerStore(SQLBaseStore):
                 WHERE
                     relates_to_id = ?
                     AND relation_type = ?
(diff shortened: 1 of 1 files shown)


clokep:
I think we're going to want to make one final change here to deny threads to threads...I might do that in a follow-up though.

Issue #11064, opened 2021-10-12
Advertise support for r0.6.1


clokep:
Synapse as of #9224 supports r0.6.1 of the Matrix spec, we should advertise support for that.

turt2live:
The spec is incorrect: the server should be advertising both `r0.6.0` and `r0.6.1`

PR #11097, opened 2021-10-15
Support Client-Server API r0.6.1

aaronraimist:
Fixes https://github.com/matrix-org/synapse/issues/11064

I can't tell if any unstable features should be removed. It looks like `org.matrix.msc2432` is in the r0.6.1 changelog but Element only checks for the `org.matrix.msc2432` flag and not for r0.6.1 so maybe it isn't in r0.6.1.

### Pull Request Checklist

### DIFF
--- a/changelog.d/11097.feature
+++ b/changelog.d/11097.feature
@@ -1,0 +1,1 @@
+Advertise support for Client-Server API r0.6.1.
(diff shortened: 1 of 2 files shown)


erikjohnston (approved):
Thanks!

PR #11098, opened 2021-10-15
Add remaining type hints to `synapse.events`.

clokep:
Builds on #11066 to finish up the type hints in `synapse.events`.


synapse/events/__init__.py
```diff
@@ -111,22 +189,22 @@ def __init__(self, internal_metadata_dict: JsonDict):
...
-    out_of_band_membership: bool = DictProperty("out_of_band_membership")
-    send_on_behalf_of: str = DictProperty("send_on_behalf_of")
-    recheck_redaction: bool = DictProperty("recheck_redaction")
-    soft_failed: bool = DictProperty("soft_failed")
-    proactively_send: bool = DictProperty("proactively_send")
-    redacted: bool = DictProperty("redacted")
-    txn_id: str = DictProperty("txn_id")
-    token_id: int = DictProperty("token_id")
-    historical: bool = DictProperty("historical")
+    out_of_band_membership: DictProperty[bool] = DictProperty("out_of_band_membership")
```
erikjohnston, line 192: This makes me very sad, but looks like mypy is indeed not intelligent enough here.

erikjohnston (approved):

synapse/events/__init__.py
clokep: What aspect makes you sad? That we can't just pretend it is a `bool`?

clokep: (We can, but you need to add a lot of ignore comments.)

erikjohnston: Yeah, exactly, the outside world shouldn't need to know its anything other than a bool

clokep: Would you prefer that we do the type hint and add some ignores then? It isn't too awful, but is a bit annoying. See c980b856f1ebc4a4c33468c5e6c9a98582bb7c0d

erikjohnston: Oh, sorry I should have been clear: Let's just leave as is, I'd rather we didn't use `# type: ignore` where possible

clokep: OK! Descriptors are very hard to type as far as I can tell. 😢

PR #11103, opened 2021-10-18
Don't remove local users from dir when the leave their last room

DMRobertson:
Preserve the invariant "local users are always in the `user_directory` table" so that the "search_all_users" config option works.


DMRobertson:
The sytest failure is a known flake. (https://github.com/matrix-org/sytest/issues/1120). I'm going to rerun CI manually and set to auto merge.

PR #11124, opened 2021-10-19
Drop synapse_storage_transaction_time_bucket

### LINKED ISSUE #11081: `synapse_storage_transaction_time_bucket` prometheus metric has too high cardinality due to desc label
### Description

The `desc` label on `synapse_storage_transaction_time_bucket` results in a very high cardinality metric. For a single instance, there's 248 variants of `desc`, multiplied by 15 buckets. This results in over 3k series for a single host.

Though this might be acceptable if you're only ingesting metrics for a single instance, for Prometheus instances that might be scraping multiple Synapses this quickly becomes a problem.

On one of our internal clusters this is resulting in over a million time series, which is about 5x the amount of time series of the next most problematic timeseries: [`synapse_http_server_response_time_seconds_bucket`](https://github.com/matrix-org/synapse/issues/11082). Though this is not causing storage issues, it causes unnecessarily high CPU load and memory usage ballooning on the ingesters (so we have to run with much bigger instances) and querying these series become problematic.

### Steps to reproduce

- Run lots of Synapse instances
- Scrape them all with a single Prometheus instance
- Prometheus gets sad

### Version information

- **Homeserver**: any
- **Version**: any since this metric got introduced
- **Install method**: unrelated
- **Platform**: unrelated

daenney:
This particular metric has a much too high cardinality due to the fact
that the desc label can have (at present) 248 values. This results in
over 3k series per Synapse. If you have a Prometheus instance that
monitors multiple Synpase instances it results in a huge amount of
additional series to ingest.

The metric in question is also not used in the Synapse dashboard and
the core team has indicated they're happy to drop this metric entirely.

Fixes #11081

### Pull Request Checklist

### DIFF
--- a/changelog.d/11124.removal
+++ b/changelog.d/11124.removal
@@ -1,0 +1,1 @@
+Remove the `synapse_storage_transaction_time_bucket` metric due to the high cardinality of the metric putting undue strain on Prometheus deployments. This metric is not used in Synapse's included Grafana dashboards.
--- a/synapse/storage/database.py
+++ b/synapse/storage/database.py
@@ -64,7 +64,6 @@ perf_logger = logging.getLogger("synapse.storage.TIME")
 sql_scheduling_timer = Histogram("synapse_storage_schedule_time", "sec")
 
 sql_query_timer = Histogram("synapse_storage_query_time", "sec", ["verb"])
-sql_txn_timer = Histogram("synapse_storage_transaction_time", "sec", ["desc"])
 
 
 # Unique indexes which have been added in background updates. Maps from table name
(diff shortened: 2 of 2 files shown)


squahtx (dismissed):
LGTM! One minor suggestion about the wording of the changelog.

reivilibre:
Any thoughts on what's happening with this PR? (is it still wanted? is it going to be a configuration flag? ...)

daenney:
Yeah, we still very much need this. The timeseries aren't all that useful and they take up a decent amount of space. We're simply dropping them at ingestion time on the EMS side, but it would be nicer to not have to.

Haven't had the time yet to check if I can change it to a summary type without consequence.

daenney:
Though I suppose we can close this one, since we'll take a different approach to solving it.

PR #11128, opened 2021-10-19
Add domain specific matching for haproxy config

bb4242:
I ran into trouble getting federation working properly with the haproxy config specified here.  After some debugging, I discovered that many federation HTTP requests, including the ones sent by the [federation tester](https://federationtester.matrix.org/), include the port number in the HTTP Host header field.  For example, instead of `Host: matrix.example.com`, these federation requests look like `Host: matrix.example.com:443`.  At least on haproxy 2.3, the extra port information causes the `acl matrix-host hdr(host) -i matrix.example.com` match to fail, since this is looking for an exact string match by default according to the [haproxy docs](http://cbonte.github.io/haproxy-dconv/2.3/configuration.html#7.1).  This failure, in turn, causes haproxy to return error codes and causes federation to fail.  Using `hdr_dom(host)`, which ignores the port information, fixes the issue in my setup.

### DIFF
--- a/changelog.d/11128.doc
+++ b/changelog.d/11128.doc
@@ -1,1 +1,1 @@
-Improve example HAProxy config in the docs to properly handle host headers with port information. This is required for federation to work correctly.
+Improve example HAProxy config in the docs to properly handle host headers with port information. This is required for federation over port 443 to work correctly.
--- a/changelog.d/11128.doc
+++ b/changelog.d/11128.doc
@@ -1,0 +1,1 @@
+Improve example HAProxy config in the docs to properly handle host headers with port information. This is required for federation to work correctly.
--- a/docs/reverse_proxy.md
+++ b/docs/reverse_proxy.md
@@ -188,7 +188,7 @@ frontend https
   http-request set-header X-Forwarded-For %[src]
 
   # Matrix client traffic
-  acl matrix-host hdr_dom(host) -i matrix.example.com
+  acl matrix-host hdr(host) -i matrix.example.com matrix.example.com:443
   acl matrix-path path_beg /_matrix
   acl matrix-path path_beg /_synapse/client
 
--- a/docs/reverse_proxy.md
+++ b/docs/reverse_proxy.md
@@ -188,7 +188,7 @@ frontend https
   http-request set-header X-Forwarded-For %[src]
 
   # Matrix client traffic
-  acl matrix-host hdr(host) -i matrix.example.com
+  acl matrix-host hdr_dom(host) -i matrix.example.com
   acl matrix-path path_beg /_matrix
   acl matrix-path path_beg /_synapse/client
 


reivilibre:
I'm not familiar with HAProxy at all, so I think I'll have to leave this for someone else who knows what they're on about to review.

However, here are some of my findings:

* the `Host` header **can** (according to HTTP spec) include port numbers and must do so if the port number isn't the default for the protocol (80 for HTTP, 443 over HTTPS).
* https://stackoverflow.com/a/48758710 makes me slightly worried about directly saying OK to `hdr_dom` (even though it looks reasonable at first glance) because it's giving me the impression that `hdr_dom(matrix.org)` would also match `anything.matrix.org`, though again, I have no experience here so can't say for sure.
  * the HAProxy manual says 'check that a **dot-delimited portion of the contents** exactly match one of the provided string patterns. This may be used with binary or string samples.' (emphasis mine)
* It seems like adding another line with the `:443` might do the trick

e.g. would this do?
```haproxy
acl matrix-host hdr(host) -i matrix.example.com
acl matrix-host hdr(host) -i matrix.example.com:443
```

or the HAProxy manual is kind of giving me the feeling that you can do this:
```haproxy
acl matrix-host hdr(host) -i matrix.example.com matrix.example.com:443
```

Again, I've not had the chance to touch haproxy myself yet so sorry for the speculation.

bb4242:
Thanks for the great feedback, @reivilibre!  Haha I'm not an expert in HAProxy either (I just started using it a few days ago after switching from relayd), so hopefully someone with more expertise could chime in here as well.  That said, I tested all of the ideas you mentioned on my local setup, and those experiments confirm what you said.  In particular:

* Yep, `hdr_dom(host) matrix.org` does indeed match subdomains like `anything.matrix.org`, so I agree we probably don't want to use that.
* `acl matrix-host hdr(host) -i matrix.example.com matrix.example.com:443` works and seems nicely specific.  So maybe that's the best bet.

I was kind of hoping there would be a way to simply ignore port information in the host headers (that what I, incorrectly, thought `hdr_dom` was for) as that seems like what we really want here, but based on my cursory reading of the haproxy docs, I don't see an obvious way to do exactly that without resorting to regexes. :grimacing:  I suppose another option might be to use `hdr_beg(host) matrix.example.com`, but that might have some odd edge cases as well.  Maybe someone who's better versed in HAProxy best practices would have some ideas of the best way to go here.

reivilibre:
> `acl matrix-host hdr(host) -i matrix.example.com matrix.example.com:443` works and seems nicely specific.  So maybe that's the best bet.

I think it's looking likely.

> I was kind of hoping there would be a way to simply ignore port information in the host headers … I suppose another option might be to use `hdr_beg(host) matrix.example.com`

hdr_beg(host) matrix.example.com

erikjohnston:
Am I right in thinking this is waiting on changes? @bb4242 please shout if that isn't right :slightly_smiling_face:

reivilibre:
I think this is actually just on us to merge... :)

changelog.d/11128.doc
```diff
@@ -0,0 +1 @@
+Improve example HAProxy config in the docs to properly handle host headers with port information. This is required for federation to work correctly.
```
reivilibre, line 1:
```suggestion
Improve example HAProxy config in the docs to properly handle host headers with port information. This is required for federation over port 443 to work correctly.
```

reivilibre (approved):

Issue #10836, opened 2021-09-16
Prevent large ephemeral event traffic to appservices upon startup


anoadragon453:
@Half-Shot mentioned that upon start up, Synapse will send all missed ephemeral events to application services that have opted into receiving them (see [MSC2409](https://github.com/matrix-org/matrix-doc/pull/2409)). There are some performance issues present with the feature's implementation in a couple of different circumstances.

Some background: a "stream token" is an incrementing integer. Every occurrence (a new typing event, a new read receipt, etc.) gets one of these IDs, and they're (usually rigidly) ordered in this way. Thus we can simply keep track of the last successfully sent stream token per app service to know what the AS needs to still receive.

1. When the bridge is first added to the homeserver: it is assumed to have a last-successful stream token of 0:
https://github.com/matrix-org/synapse/blob/bdfde6dca11a9468372b3c9b327ad3327cbdbe4a/synapse/storage/databases/main/appservice.py#L387-L412

When it comes time to send out the ephemeral events the appservice has missed, Synapse will send *all* read receipts and presence updates that it knows about and that are relevant to the namespaces the appservice has registered:

https://github.com/matrix-org/synapse/blob/a8a27b2b8bac2995c3edd20518680366eb543ac9/synapse/handlers/appservice.py#L266-L273

This historical data isn't very useful to the appservice as it has just been registered.

2. When a bridge has been offline for a while and comes back: it will receive all missed, relevant EDUs since it went offline. In particular, historical EDUs are typically less useful than the current state of things.

I'd be tempted to implement a time bound to the EDUs we send to appservices, such that very old EDUs are no longer considered.

Fixing both of these would help reduce the large amount of EDUs that can be sent to and overload both the appservice and Synapse itself when it tries to pull all of that information out of the database.

Finally, note that the above problems only seem to apply to the implementation of read receipts and presence events. Historical typing events are flat out not sent to appservices:

https://github.com/matrix-org/synapse/blob/a8a27b2b8bac2995c3edd20518680366eb543ac9/synapse/handlers/appservice.py#L235

anoadragon453:
After some investigation by Scott Weber and myself, it appears that only the most recent 100 read receipts will be sent to the appservice:

https://github.com/matrix-org/synapse/blob/a3a7514570f21dcad6f7ef4c1ee3ed1e30115825/synapse/handlers/receipts.py#L236-L266

[...]

PR #11137, opened 2021-10-20
Fix providing a `RoomStreamToken` instance to `_notify_app_services_ephemeral`

anoadragon453:
We typically figure out which EDUs to send to application services based the last stream token that appservice successfully received. This is true for read receipts and user presence, however for typing we always just send the last typing event. This process is started whenever a new read receipt, presence update or typing event is processed by the server. In that case, `notify_interested_services_ephemeral` will be called with the new event and its stream token (so we can check whether we need to send any other events to the appservice may have missed).

With all that in mind, I believe I've found a bug in the implementation. If a `RoomStreamToken` object was provided as the `new_token` argument to `_notify_app_services_ephemeral`, `new_token` would be set to `None`, before then being passed to `notify_interested_services_ephemeral` and then `_notify_interested_services_ephemeral`. The commit that introduced this change (559974fb4bc1872abced395de22ffbd2292b2f8b) was intended to prevent mypy from failing. `(_)notify_app_services_ephemeral` was also configured to allow `new_token` to be an `Optional[int]` as a result.

However, if `None` is provided, `_notify_app_services_ephemeral` will end up passing `None` to `set_type_stream_id_for_appservice`, which will clear each appservice's stream token for the given `stream_key`. ~~Doing so means that we send all presence or read receipt data from a None stream token up until the latest token(!). This seems like a bug to do every time we have a `RoomStreamToken` rather than an `int`.~~ Actually, all stream keys that are passed to `_notify_interested_services_ephemeral` (`typing_key`, `receipt_key`, `presence_key`) are passing `int` values for `new_token`, so the buggy code would have never fired anyways.

Regardless, I propose converting the `RoomStreamToken` to an `int` using `RoomStreamToken.stream` (~~I _think_ this is the right way to convert it. [`RoomStreamToken`](https://github.com/matrix-org/synapse/blob/aa2c027792d04c36b17866710e95a41d31f5d99c/synapse/types.py#L415-L464) can end up being a vector clock when you have multiple event persister workers; in this case `stream` will be the minimum stream token amongst all workers... which I think is fine... maybe?~~ This appears to be correct, we do it [elsewhere as well](https://github.com/matrix-org/synapse/blob/9abbd08e33fc7710d952e870e26368d17915bbba/synapse/handlers/appservice.py#L67-L70).)

~~This may relate to the large bursts of EDU traffic Synapse is sending to AS's ala https://github.com/matrix-org/synapse/issues/10836.~~ See above. The code should not have actually resulted in broken behaviour.

### DIFF
--- a/synapse/handlers/appservice.py
+++ b/synapse/handlers/appservice.py
@@ -212,12 +212,18 @@ class ApplicationServicesHandler:
         if stream_key not in ("typing_key", "receipt_key", "presence_key"):
             return
 
-        # Convert new_token from a RoomStreamToken to an int if necessary.
-        # We just use the minimum stream ordering and ignore the vector clock
-        # component. This is safe to do as long as we *always* ignore the vector
-        # clock components.
-        if isinstance(new_token, RoomStreamToken):
-            new_token = new_token.stream
+        # Assert that new_token is an integer (and not a RoomStreamToken).
+        # All of the supported streams that this function handles uses an
+        # integer to track progress (rather than a RoomStreamToken - a
+        # vector clock implementation - as they don't support multiple
+        # stream writers).
+        #
+        # As a result, we simply assert that new_token is an integer.
+        # If we do end up needing to pass a RoomStreamToken down here
+        # in the future, using RoomStreamToken.stream (the minimum stream
+        # position) to convert to an ascending integer value should work.
+        # Additional context: https://github.com/matrix-org/synapse/pull/11137
+        assert isinstance(new_token, int)
 
         services = [
             service
--- a/synapse/handlers/appservice.py
+++ b/synapse/handlers/appservice.py
@@ -213,10 +213,10 @@ class ApplicationServicesHandler:
             return
 
         # Assert that new_token is an integer (and not a RoomStreamToken).
-        # All of the supported streams that this function handles uses an
+        # All of the supported streams that this function handles use an
         # integer to track progress (rather than a RoomStreamToken - a
-        # vector clock implementation - as they don't support multiple
-        # stream writers).
+        # vector clock implementation) as they don't support multiple
+        # stream writers.
         #
         # As a result, we simply assert that new_token is an integer.
         # If we do end up needing to pass a RoomStreamToken down here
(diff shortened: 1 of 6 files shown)


synapse/handlers/appservice.py
```diff
@@ -212,6 +212,13 @@ def notify_interested_services_ephemeral(
         if stream_key not in ("typing_key", "receipt_key", "presence_key"):
             return
 
+        # Convert new_token from a RoomStreamToken to an int if necessary.
+        # We just use the minimum stream ordering and ignore the vector clock
+        # component. This is safe to do as long as we *always* ignore the vector
+        # clock components.
+        if isinstance(new_token, RoomStreamToken):
+            new_token = new_token.stream
```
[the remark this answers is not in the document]
anoadragon453: Good point -- I've changed this to an assert and left some context.

synapse/handlers/appservice.py
```diff
@@ -212,12 +212,18 @@ def notify_interested_services_ephemeral(
...
             return
 
-        # Convert new_token from a RoomStreamToken to an int if necessary.
-        # We just use the minimum stream ordering and ignore the vector clock
-        # component. This is safe to do as long as we *always* ignore the vector
-        # clock components.
-        if isinstance(new_token, RoomStreamToken):
-            new_token = new_token.stream
+        # Assert that new_token is an integer (and not a RoomStreamToken).
+        # All of the supported streams that this function handles uses an
```
richvdh, line 216:
```suggestion
        # All of the supported streams that this function handles use an
```

synapse/handlers/appservice.py
```diff
@@ -212,12 +212,18 @@ def notify_interested_services_ephemeral(
...
+        # integer to track progress (rather than a RoomStreamToken - a
+        # vector clock implementation - as they don't support multiple
+        # stream writers).
```
richvdh, line 219:
```suggestion
        # vector clock implementation) as they don't support multiple
        # stream writers.
```

richvdh (approved):
lgtm

PR #11140, opened 2021-10-20
Deduplicate presence updates before sending them to application services

anoadragon453:
Mostly addresses problem 2 of https://github.com/matrix-org/synapse/issues/10836#issuecomment-921960097.

We calculate presence updates for application services based on the
users that application service is interested in. For each of these
users, we determine which presence updates they are set to receive,
compile that into a list, and then send every update from the list
to the application service.

However, because a single presence update can cause a notification
for many different users, we're likely to end up with lots of
duplicated presence updates being collected here. Currently, all of
these are sent to the application service.

By using a Set, this deduplication happens automatically.

### DIFF
--- a/synapse/handlers/appservice.py
+++ b/synapse/handlers/appservice.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import logging
-from typing import TYPE_CHECKING, Collection, Dict, Iterable, List, Optional, Union
+from typing import TYPE_CHECKING, Collection, Dict, Iterable, List, Optional, Set, Union
 
 from prometheus_client import Counter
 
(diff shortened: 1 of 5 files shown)


synapse/handlers/appservice.py
```diff
@@ -427,7 +427,7 @@ async def _handle_presence(
                 from_key=from_key,
             )
             time_now = self.clock.time_msec()
-            events.extend(
+            events.update(
```
clokep, line 430: Should this really be a set or do we want to de-duplicate based on user ID?

synapse/handlers/typing.py
```diff
@@ -491,7 +491,7 @@ async def get_new_events_as(
                 ):
                     continue
 
-                events.append(self._make_event_for(room_id))
+                events.add(self._make_event_for(room_id))
```
clokep, line 494: This feels like it being a set might not de-duplicate a lot since the object embeds a list of user IDs?

clokep (commented):
This looks plausible, but I have some questions since I'm not super familiar with this code.

DMRobertson:
Looks like we forgot about this one. Is it worth resurrecting @anoadragon453?

anoadragon453:
Let's close it for now since it's ancient. One day I may return to finish it up.

PR #11147, opened 2021-10-21
Add a ModuleApi method to update a user's membership in a room

babolivier:
Should be fine to run on workers since `hs.get_room_member_handler().update_membership` ends up hitting `synapse.handlers.message.EventCreationHandler._persist_event`, which defers to the right worker if necessary.

### DIFF
--- a/synapse/module_api/__init__.py
+++ b/synapse/module_api/__init__.py
@@ -692,8 +692,8 @@ class ModuleApi:
                     # federation but the remote server responded with a non-standard
                     # status code.
                     logger.error(
-                        "Got non-404 error status when fetching profile for %s"
-                        % target_user_id.to_string(),
+                        "Got non-404 error status when fetching profile for %s",
+                        target_user_id.to_string(),
                     )
 
             # Set the profile where it needs to be set.


buffless-matt:
Out of curiosity, why was [this constraint](https://github.com/matrix-org/synapse/blob/583f22780f44157c50bc2dc5c242e88cc18c7886/synapse/module_api/__init__.py#L961-L964) added?

Looking at [this REST API end-point handler](https://github.com/matrix-org/synapse/blob/583f22780f44157c50bc2dc5c242e88cc18c7886/synapse/rest/client/room.py#L841-L908), I can't see a similar constraint around usage of [update_membership](https://github.com/matrix-org/synapse/blob/583f22780f44157c50bc2dc5c242e88cc18c7886/synapse/handlers/room_member.py#L488-L587).

I ask because I have a module API use-case that involves auto-joins to federated/remote (public) rooms.

buffless-matt:
Also just noticed that I don't think the [shadow ban part](https://github.com/matrix-org/synapse/blob/583f22780f44157c50bc2dc5c242e88cc18c7886/synapse/module_api/__init__.py#L955) of the design contract is implemented. `requester` is defined [here](https://github.com/matrix-org/synapse/blob/583f22780f44157c50bc2dc5c242e88cc18c7886/synapse/module_api/__init__.py#L966), but `create_requester` has a [default value](https://github.com/matrix-org/synapse/blob/583f22780f44157c50bc2dc5c242e88cc18c7886/synapse/types.py#L170) of `False` for `shadow_banned`.

babolivier:
> Out of curiosity, why was [this constraint](https://github.com/matrix-org/synapse/blob/583f22780f44157c50bc2dc5c242e88cc18c7886/synapse/module_api/__init__.py#L961-L964) added?

Because a homeserver cannot send events as a user that doesn't belong on it (e.g. `bob.com` cannot send events as `@alice:alice.com`). We do it early to prevent wasting any resource on events the protocol wouldn't allow us to send.

> Also just noticed that I don't think the [shadow ban part](https://github.com/matrix-org/synapse/blob/583f22780f44157c50bc2dc5c242e88cc18c7886/synapse/module_api/__init__.py#L955) of the design contract is implemented.

That's a good point, though imo it's not clear that it should be done here. Shadow-banning is effectively no-op'ing operations performed by the user (e.g. the user actively tries to join a room), whereas in the context of a module it's the server automatically (or from admin intervention) doing an operation on behalf of the user. I think there are probably use cases where we might still want to be able to manipulate a user (in this case their room memberships) even if they're shadow-banned. Following the same logic, the [room membership admin API](https://matrix-org.github.io/synapse/latest/admin_api/room_membership.html) is also calling `create_requester` while keeping the `shadow_banned` value to `False` (see https://github.com/matrix-org/synapse/blob/e3163e2e11cf8bffa4cb3e58ac0b86a83eca314c/synapse/rest/admin/rooms.py#L456). Other opinions welcome though.

buffless-matt:
> Because a homeserver cannot send events as a user that doesn't belong on it...

Right, of course. Sorry, I got mixed up with something related.

> We don't do this check on client endpoints because clients don't provide the sender's user ID. It's resolved by looking up who the access token used to perform the request belongs to.

Makes sense.

> That's a good point, though imo it's not clear that it should be done here. ...

Ah right, yeah that's a good point. Thanks, just wanted to check it wasn't missed in this PR.

--------------------

So the thing I got mixed up with, is that my use-case requires joining a local user to a remote/federated room, but it doesn't appear as though the implementation from this PR allows for that (which is what I meant to raise, not about why the `sender` must be local).

My use-case  requires that [remote_room_hosts](https://github.com/matrix-org/synapse/blob/e17e5c97e0d2be7a82c782e8ef9774e682968c7b/synapse/handlers/room_member.py#L495) be non-empty (i.e. without it, I'm see [this exception](https://github.com/matrix-org/synapse/blob/e17e5c97e0d2be7a82c782e8ef9774e682968c7b/synapse/handlers/room_member.py#L1682-L1686)), but this module API method [doesn't utilise it](https://github.com/matrix-org/synapse/blob/e17e5c97e0d2be7a82c782e8ef9774e682968c7b/synapse/module_api/__init__.py#L1002-L1008).

Would it be OK if I made a PR to modify the `update_room_membership` module API method, so that the `remote_room_hosts` parameter can be optionally passed through to `update_membership`? If not, could you please propose a suitable alternative design that allows for a local user to be joined to a remote/federated room (and then I'd be happy to make the PR)?

babolivier:
> Would it be OK if I made a PR to modify the update_room_membership module API method, so that the remote_room_hosts parameter can be optionally passed through to update_membership?

Yes I think this would make sense and is probably the best way to solve your issue.

buffless-matt:
> Yes I think this would make sense and is probably the best way to solve your issue.

@babolivier I've created a PR at #13441.

PR #11157, opened 2021-10-22
Require body for read receipts with user-agent exceptions

### LINKED ISSUE #11156: Limit which clients are allowed to send read receipts without a body
The Matrix Spec requires that requests to `POST /rooms/{roomId}/receipt/{receiptType}/{eventId}` have a body, even if it's just `{}`. Unfortunately, older Element Android clients omitted this and sent empty bodies instead.

We worked around this in #10531, but we'd like to eventually remove the workaround per #10534.

To prevent the problem from getting worse, we should limit the scope of the workaround to _only_ apply to known bad clients.

Specifically, instead of unconditionally setting `allow_empty_body=True` here:

https://github.com/matrix-org/synapse/blob/ba00e20234eadae66f105f8bda64e39beed9a92d/synapse/rest/client/receipts.py#L55

We should only set it to `True` if the result of `get_request_user_agent(request)` contains `Android` and satisfies one of the following:

- Contains `Riot`  (e.g., `Old Riot.im`, `Riot`, `RiotX`, `Element (Riot.im)`)
- Starts with `Element/1.[012].*`
- Starts with `SchildiChat/1.[012].*`

(This will unnecessarily allow _all_ Element 1.2.x releases to send empty bodies, instead of just versions < 1.2.1, but doing so really simplifies the glob, and we know that 1.2.1 and later are well behaved, so no harm done.)

rogersheu:
### Pull Request Checklist

Signed-off-by: Roger Sheu <rogersheu92@gmail.com>

Fixes #11156.


rogersheu:
> > I think the fix would either be to fix the unit tests by adding in this extra information or to add in new unit tests addressing each of these possible clients.
> 
> I'd suggest:
> 
> * change `test_read_receipt_with_empty_body` to take an extra `user_agent: str` and `expected_status_code: int` argument
> * decorate this test with `@parameterized.expand` to cover the different cases. Something like:
>   ```python
>   @parameterized.expand([
>       ("Android Element 1.0", 400),
>       ("Android Element 1.4", 200),
>       ...
>   ])
>   def test_read_receipt_with_empty_body(...):
>   ```

>     
>   (The user agents are just examples here; I'm not sure of the specifics!)
> * Pass `custom_headers` to `make_request` to set the user agent header to the test argument.
> * Change the final `assertEqual` call to compare to `expected_status_code`.
> 
Cool, thanks for the guidance. What are the most important platforms/versions to test? I know some instances from some of our previous discussions, but I'm not fully apprised of all the version options out there.

Examples:
* `Element/1.2.2 (Linux; Android 6)` (pass)
* `SchildiChat/1.3.6 (Android 11)` (pass)
* `Element (Riot.im) (Android 9)` (pass)
* `Element/1.2.1` (fail)
* `Element dbg/1.1.8-dev (Android)` (fail)
* 
> > I also ran into an issue (specifically with fcntl) running tests locally since I'm running Windows and not Linux.
> 
> That's unfortunate. Maybe WSL can help with this?

I'll look into it.

synapse/rest/client/receipts.py
```diff
@@ -52,7 +54,14 @@ async def on_POST(
         if receipt_type != "m.read":
             raise SynapseError(400, "Receipt type must be 'm.read'")
 
-        body = parse_json_object_from_request(request, allow_empty_body=True)
+        user_agent = get_request_user_agent(request)
+        pattern = re.compile(r"(?:Element|SchildiChat)/1\.[012]\.")
```
reivilibre, line 58: We want to compile this only once (e.g. make it a constant outside of the function) or alternatively use `re.match` without pre-compiling so that it uses the re compiled pattern cache.

tests/rest/client/test_sync.py
```diff
@@ -417,7 +419,23 @@ def test_hidden_read_receipts(self):
...
-    def test_read_receipt_with_empty_body(self):
+    @parameterized.expand(
+        [
+            (
+                "agent1",
+                "Element/1.2.2 (Linux; U; Android 9; MatrixAndroidSDK_X 0.0.1)",
+                200,
+            ),
+            ("agent2", "Element/1.2.2 (Linux; Android 6)", 200),
+            ("agent3", "SchildiChat/1.3.6 (Android 11)", 200),
```
reivilibre, line 430: I was under the impression that 1.3+ should fail
```suggestion
            ("agent3", "SchildiChat/1.3.6 (Android 11)", 400),
```

(maybe add a pass case for SchildiChat too such as
```
            ("agent7", "SchildiChat/1.1.6 (Android 11)", 200),
```
)

tests/rest/client/test_sync.py
```diff
@@ -426,8 +444,9 @@ def test_read_receipt_with_empty_body(self):
             "POST",
             "/rooms/%s/receipt/m.read/%s" % (self.room_id, res["event_id"]),
             access_token=self.tok2,
+            custom_headers=[("User Agent", user_agent)],
```
reivilibre, line 447: the header has a hyphen-minus in the name
```suggestion
            custom_headers=[("User-Agent", user_agent)],
```

reivilibre (changes requested):
Some corrections. I also noticed your tests are failing (though this might help). Don't forget to click the circle arrows button to re-request a review when you need :)

tests/rest/client/test_sync.py
rogersheu: Ah, I changed that last minute and forgot about the 1.```[012]```. restriction, so this should indeed fail.

tests/rest/client/test_sync.py
rogersheu: Is that standard header formatting or is it specific to this application? I tried sifting through the uses of custom_headers and make_request in the code but could only find a couple of instances.

tests/rest/client/test_sync.py
```diff
@@ -417,7 +419,23 @@ def test_hidden_read_receipts(self):
...
-    def test_read_receipt_with_empty_body(self):
+    @parameterized.expand(
+        [
+            (
+                "agent1",
+                "Element/1.2.2 (Linux; U; Android 9; MatrixAndroidSDK_X 0.0.1)",
+                200,
+            ),
+            ("agent2", "Element/1.2.2 (Linux; Android 6)", 200),
+            ("agent3", "SchildiChat/1.3.6 (Android 11)", 400),  # Will allow empty body again starting at version 1.3+
```
reivilibre, line 430:
```suggestion
            ("agent3", "SchildiChat/1.3.6 (Android 11)", 400),  # Will deny empty body again starting at version 1.3+
```
I think?

tests/rest/client/test_sync.py
reivilibre: It's part of the HTTP standard. I'm not sure if it's strictly case sensitive, but at the very least the hyphen-minus is required and spaces probably aren't even allowed :)

tests/rest/client/test_sync.py
```diff
@@ -417,7 +419,23 @@ def test_hidden_read_receipts(self):
...
+            ("agent4", "Element (Riot.im) (Android 9)", 200),
+            ("agent5", "Element/1.2.1", 400),  # Does not contain "Android"
+            ("agent6", "Element dbg/1.1.8-dev (Android)", 400),  # Different format, missing "/" after Element
```
reivilibre, line 433: I think we're missing a test that old SchildiChat versions will be exempt from spec compliance.
```suggestion
            ("agent6", "Element dbg/1.1.8-dev (Android)", 400),  # Different format, missing "/" after Element
            ("agent7", "SchildiChat/1.2.1 (Android)", 200), 
```

reivilibre:
*The linter wants to have a word with you.*

:smile:

Running `./script-dev/lint.sh` with your virtual environment activated will likely sort it out. Sorry that you only get the status checks after someone approves it; it's a bit painful but it's a GitHub restriction for first-time contributors :(.

tests/rest/client/test_sync.py
rogersheu: Added.

tests/rest/client/test_sync.py
rogersheu: Correct, thanks for catching that!

tests/rest/client/test_sync.py
rogersheu: I see. I'll have to read about that then. Thanks!

rogersheu:
I thought I'd ran the linter, but it turns out I was in the wrong directory when I ran it. Should be better now!

reivilibre (commented):
Thanks for your help!

rogersheu:
Thanks for walking me through these changes, everyone!

PR #11161, opened 2021-10-22
Do not allow MSC3440 threads to fork threads

clokep:
[MSC3440](https://github.com/matrix-org/matrix-doc/pull/3440) requires that a thread cannot be created against an event which already has a relation.

This has quite a few changes in it:

* Improves validation for relations in the C-S API.
* Denies events which attempt to start a thread on a message which already has a relation.


synapse/handlers/message.py
```diff
@@ -1016,9 +1048,12 @@ async def create_new_client_event(
             if already_exists:
                 raise SynapseError(400, "Can't send same reaction twice")
 
-        logger.debug("Created event %s", event.event_id)
-
-        return event, context
+        # If this relation is a thread, then ensure thread head is not part of
+        # a thread already.
```
clokep, line 1052: The MSC says that the thread head should not have any relation. (I.e. it must be an event that is not related to any other event.)

synapse/handlers/message.py
```diff
@@ -1001,13 +1001,45 @@ async def create_new_client_event(
...
+        relation_type = relation.get("rel_type")
+        if not relation_type:
+            return
+
+        # Ensure the parent is real.
+        relates_to = relation["event_id"]
+        parent_event = await self.store.get_event(relates_to)
+        if not parent_event:
+            # TODO Attempt to fetch the event over federation.
+            raise SynapseError(400, "Can't send relation to unknown event")
```
clokep, line 1033: What do we do in this case? It is (theoretically) possible the user's homeserver doesn't have this backfilled *or* that the user doesn't have permission to see this event. Maybe if we can't find the event we should just skip some of the next validation and assume that clients will handle it OK?

clokep: There was a bit of a discussion around this in #synapse-dev:matrix.org (see https://matrix.to/#/!XaqDhxuTIlvldquJaV:matrix.org/$N7UHzQ6lrerKq5LtqaPyjZmAWyP6GPCn3bMwU2BcERA?via=matrix.org&via=vector.modular.im&via=envs.net), the overall conclusion was that servers and clients must be robust to not being able to have an event for some reason so it is *probably* reasonable to skip validation if we don't know about the parent event for some reason. We might also want to see if there are other events using that as a parent relation and deduce that it is *likely* OK.

synapse/storage/databases/main/relations.py
```diff
@@ -132,6 +132,44 @@ def _get_recent_references_for_event_txn(
...
+        Returns:
+            True if the event is the target of any relations.
+        """
+
+        result = await self.db_pool.simple_select_one_onecol(
+            table="event_relations",
+            keyvalues={"relates_to_id": parent_id},
+            retcol="event_id",
+            allow_none=True,
+            desc="event_has_relations",
```
reivilibre, line 169: duplicate description

synapse/storage/databases/main/relations.py
```diff
@@ -132,6 +132,44 @@ def _get_recent_references_for_event_txn(
...
+
+    async def event_related_to(self, parent_id: str) -> bool:
+        """Check if a given event has other event relating to it.
+
+        Args:
+            parent_id: The event to check.
+
+        Returns:
+            True if the event is the target of any relations.
+        """
```
reivilibre, line 162: I'm finding this name hard to read; from the method name I'd expect this to be a binary operator (is _ related to _? Well, literally called a relation (mathematically)...).
It doesn't seem like that, though.

Maybe this is some personal misunderstanding about what 'relations' are in Matrix; though in any case this makes it hard to approach for noobs like me.
Reckon you could extend the comment to describe what the shape of the event ID referred to by `parent_id` is meant to be; or in other words: what are 'you' (well, this function) looking for?

The same for `event_has_relations`, if you wouldn't mind, please.
It may be that a different name would fit them better, but I shouldn't be telling you that until I know what they actually mean :).

reivilibre (changes requested):

synapse/handlers/message.py
```diff
@@ -1016,9 +1055,12 @@ async def create_new_client_event(
...
 
-        logger.debug("Created event %s", event.event_id)
-
-        return event, context
+        # Don't attempt to start a thread if the parent event is a relation.
+        elif relation_type == RelationTypes.THREAD:
+            if await self.store.event_has_relations(relates_to):
+                raise SynapseError(
+                    400, "Cannot start threads from an event with a relation"
+                )
```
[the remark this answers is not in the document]
clokep: > Does this mean that nested threading is not possible?

Yes, it means you can not nest threads (and cannot start threads off of, e.g. a reaction). Is there a way I can make this clearer? Maybe the method should be `event_is_relation` instead of `event_has_relations`. It is checking if the event you're trying to create a relation to also has a relation.

synapse/storage/databases/main/relations.py
clokep: I find the terminology very ambiguous, which makes it hard to write clear names. I'll try to make them a bit more straightforward though.

If you have a two events: `A` and `B` and `B` includes a relation to `A`, then:

* `event_related_to(A)` would return `True` (since `B` references `A`)
* `event_related_to(B)` would return `False` (since no one references `B`)
* `event_has_relations(A)` would return `False` (since `A` doesn't reference anything)
* `event_has_relations(B)` would return `True` (since `B` references `A`)

Hopefully those examples clear it up for your own understanding, but I'll try to reword and expand the docstrings.

clokep: I pushed 558175b3702e17bbb85673560ae5215a253dc97b to attempt to clarify this, let me know if it makes sense!

synapse/handlers/message.py
```diff
@@ -1001,13 +1001,52 @@ async def create_new_client_event(
...
+            # And in the same room.
+            if parent_event.room_id != event.room_id:
+                raise SynapseError(400, "Relations must be in the same room")
+
+        else:
+            # There must be some reason that the client knows the event exists,
+            # see if there are existing relations. If so, assume everything is fine.
+            if not await self.store.event_related_to(relates_to):
+                # Otherwise, the client can't know about the parent event!
+                raise SynapseError(400, "Can't send relation to unknown event")
```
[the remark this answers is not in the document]
clokep: > Is it because you might see a thread where the base message is not visible to you/your homeserver, but the subsequent messages are?

This is the situation that I was thinking of, yes. (The simple case is if you're invited to a room that doesn't share history and you can't see the thread head, but you can see subsequent messages in it).

I think it is OK to divulge in this case that the event ID is valid. Theoretically you would know already since you've seen other events referencing that event.

It could be used to iterate through whether events are in a room, but that seems unreasonable in terms of the space to scan + rate limits, I think.

clokep:
@reivilibre I left a couple of comments and pushed some clarifications, let me know what you think!

synapse/storage/databases/main/relations.py
```diff
@@ -132,41 +132,66 @@ def _get_recent_references_for_event_txn(
...
+        and event_id in the content:
+
+        {
+            "content": {
+                "m.relates_to": {
+                    "rel_type": "m.replace",
+                    "event_id": "$other_event_id"
+                }
+            }
+        }
```
reivilibre, line 148: The small piece of mathematician in me thinks the terminology is off (probably not your fault so much as the spec's or even English in general — maybe relation has more meanings that confuse this), but the example paints a thousand words, so thanks :)

synapse/storage/databases/main/relations.py
```diff
@@ -132,41 +132,66 @@ def _get_recent_references_for_event_txn(
...
+        content:
+
+        {
+            "content": {
+                "m.relates_to": {
+                    "rel_type": "m.replace",
+                    "event_id": "$parent_id"
+                }
+            }
+        }
```
reivilibre, line 180: Thanks!

synapse/storage/databases/main/relations.py
```diff
@@ -132,41 +132,66 @@ def _get_recent_references_for_event_txn(
...
             retcol="event_id",
             allow_none=True,
-            desc="event_has_relations",
+            desc="event_includes_relation",
         )
         return result is not None
 
-    async def event_related_to(self, parent_id: str) -> bool:
-        """Check if a given event has other event relating to it.
+    async def event_has_relation(self, parent_id: str) -> bool:
```
reivilibre, line 166: definitely clunkier, but do you think `event_is_target_of_relation` might paint this better?

I think I'm beginning to see a bit more clearly, that *relations* here are really just arrows between events.
It might be sensible to call the pointer a 'source' and the pointee the 'target' (perhaps could call the other function `event_is_source_of_relation`?).
My suggestion makes sense to me here and now in this context, but it's hard to say if this would get confusing in another context or if other people were trying to make sense of it :/.

The example below is really crucial though; as for the name, I'm happy to trust what you think is best.

(An alternative to source/target would be parent/child, if that's any better — I think it makes lots of sense for reactions, threads, replies and edits so maybe that's the winner?)

synapse/handlers/message.py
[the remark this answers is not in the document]
reivilibre: I feel like `event_has_parent` or similar would clarify this particular case a lot (though it probably wants the word `relation` in there ... `event_has_relation_parent`? hmm)
Mind you, relations are meant to be a bit abstract and don't necessarily mean parent/child, but the current uses for them do kind of seem parent/child-like (and the parent has to have been born first, since the child refers to it by ID...).
Naming is hard.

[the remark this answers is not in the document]
reivilibre: (though I will note that it's sad that nested threading will not be an option here, I'll have to trust that this has been through MSCs and people are happy that there is a workaround for nested threading in the future)

synapse/handlers/message.py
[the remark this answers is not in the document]
reivilibre: Yeah, that seems OK. I don't think we're particularly stringent about this anyway. I can't properly imagine a case where this would hurt us, but that doesn't mean it's not possible (though we probably already divulge this elsewhere)

reivilibre (approved):
Naming is still not entirely to my taste but I'll trust you there — not sure any particular set of words will ever be perfect. All else looks good to me — thanks!

synapse/storage/databases/main/relations.py
clokep: I was thinking a bit more about this last night and we seem to be missing ones for the target event of the relation vs. the event which includes the relation? Definitely open to ideas if you have any!

reivilibre: if I understand you properly:

> to be missing ones [words?] for the target event of the relation vs. the event which includes the relation

parent and child are possibly my favourite for now, and the intuition that the parent existed first matches up with reality.

If this isn't what you were asking about the please rephrase :)

clokep: Yes, I meant `words` not `ones`. I've edited my comment!

clokep:
> Naming is still not entirely to my taste but I'll trust you there — not sure any particular set of words will ever be perfect. All else looks good to me — thanks!

Thanks! I'll sleep on it again and see if I can come up with anything. If not I'll merge, as-is.

synapse/storage/databases/main/relations.py
clokep: It seems that MSC2974 refers to them as "target event" and "relating event".

synapse/handlers/message.py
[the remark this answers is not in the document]
clokep: > (though I will note that it's sad that nested threading will not be an option here, I'll have to trust that this has been through MSCs and people are happy that there is a workaround for nested threading in the future)

There's various reasons given in the MSC, this conversation belongs on there. 😄

PR #5970, opened 2019-09-03
Docker packaging should not su-exec or chmod if already running as UID/GID

michaelkaye:
It's possible to configure docker to set the UID/GID before a container starts.

Some docker based environments, such as openshift, [force containers to run with a specific UID](https://docs.openshift.com/enterprise/3.1/creating_images/guidelines.html#use-uid) and GID and cannot start as 0:0 and drop to a different user to continue execution.

This PR checks whether the environment of the docker container is that of the uid/gid running start.py are the same as the UID/GID specified in the environment.

If this is true, then it does not attempt any chmod / su-exec based operations, as:

- su-exec requires root to function
- chmod will work only if the target uid:gid matches the existing permissions.

so these will not generally succeed if the container is not already running as root.

Existing deployments using root:root to run the container will be unaffected, as their UID/GID will not match.
Existing deployments using another UID/GID to run the container currently do not function as they're unable to use su-exec to change.

If UID != getuid but getuid != root, then running the container will fail with a chmod / su-exec error that it's unable to assume the group, 

I think this could also be expressed as a check for getuid/getgid not being root:root and if that's true, set `ownership = None`, but that would then not attempt to perform the changes, and you would be able to supply a UID/GID but not use them. But happy to have thoughts on whether that would be better.

### Pull Request Checklist

* [X] Pull request is based on the develop branch
* [x] Pull request includes a [changelog file](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.rst#changelog)
* [ ] Pull request includes a [sign off](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.rst#sign-off)

PR #11209, opened 2021-10-29
Docker: avoid changing userid unnecessarily

richvdh:
Two related changes

 * When the image is run with a `generate` param to generate a new config file, don't try to change user ID for the generate step. I think this was introduced accidentally as part of #5970.

 * When we are run as non-root (eg, with a `--user` param to docker, or via a different runtime like OpenShift which does so by default), don't try to change user unless a UID is explicitly given.

   It was possible to work around this before by setting UID to match whatever userid we were being run as, but that was somewhat fiddly.

Reviewable commit-by-commit.

### DIFF
--- a/changelog.d/11209.docker
+++ b/changelog.d/11209.docker
@@ -1,0 +1,1 @@
+Avoid changing userid when started as a non-root user, and no explicit `UID` is set.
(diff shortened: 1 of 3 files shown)


squahtx (approved):
lgtm!

bziemons:
> It was possible to work around this before by setting UID to match whatever userid we were being run as, but that was somewhat fiddly.

Sadly this workaround was broken in a non-obvious way by this change :(

fphammerle:
> Sadly, this workaround was broken in a non-obvious way by this change :(

@bziemons, the workaround is no longer required. Just set `--user 1234:2345` (or whatever) without setting the `UID` and `GID` environment variable.

bziemons:
@fphammerle thanks for your help. The problem is not the current solution, but rather the previous solution not working anymore. I was pointing out the backwards incompatibility and I wish this change would have been done with backwards compatibility in mind.

PR #11210, opened 2021-10-29
Default value for `public_baseurl`

richvdh:
Having a working `public_baseurl` is required for many things in the codebase, and for common deployments, we can derive a working default from `server_name`. So, we might as well do that, to help admins out as much as we can.

One slight wrinkle is that setting `public_baseurl` currently changes some behavior: in particular, it affects whether we put the public_baseurl in the response to `/login`, which tells clients to redirect to that url. So, if we suddenly start returning a *wrong* url there instead of none at all, it could break existing, working deployments.

Also adds some validation for `public_baseurl` while I'm in the area.

### DIFF
--- a/synapse/config/server.py
+++ b/synapse/config/server.py
@@ -16,6 +16,7 @@ import itertools
 import logging
 import os.path
 import re
+import urllib.parse
 from textwrap import indent
 from typing import Any, Dict, Iterable, List, Optional, Set, Tuple, Union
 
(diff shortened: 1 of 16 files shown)


richvdh:
> I think that will make it OK, more of just noting that we ought to be careful with this / ask for extra testing or something?

I guess so. It's worth noting that this didn't affect any of our own deployments the first time round, because we set `public_baseurl` explicitly. I'm also conscious that the type of people likely to test RCs are also the type of people likely to set `public_baseurl`. So, suggestions welcome.

clokep:
> So, suggestions welcome.

The only thing I can think of is specifically asking the people who reported issues previously?

synapse/config/server.py
```diff
@@ -288,6 +289,20 @@ def read_config(self, config, **kwargs):
                 public_baseurl += "/"
         self.public_baseurl = public_baseurl
 
+        # check that public_baseurl is valid
+        try:
+            splits = urllib.parse.urlsplit(self.public_baseurl)
+        except Exception as e:
+            raise ConfigError(f"Unable to parse URL: {e}", ("public_baseurl",))
```
clokep, line 296: I wonder if this has minor potential to cause previously working configs to break (if somehow the generated `public_baseurl` from `server_name` creates an invalid URL). Looks like we already validate that it above though so should be OK.

clokep (approved):
Looks good! Definitely cleaner to always have this set!

synapse/config/server.py
richvdh: I think it should be ok.

I feel like if `server_name` did generate a bad `pubic_baseurl`, that must be a pretty broken setup anyway.

clokep: That's fair! 👍

Issue #8308, opened 2020-09-13
Synapse should be able to serve a `.well-known/matrix/server` file


maquis196:
**I'd like to be able to serve .well-known files directly from the container:**

So usually this isn't a problem for most, if you have a classic reverse proxy setup, its super easy to set up one way or another, if youre like me trying to set up traefik, it becomes more complicated. Its possible to setup a seperate nginx instance to catch such requests but personally I'd like to be able to provide either a file to be mounted or variable to be passed into the container that would serve these files. 

One workaround for this is to mount a file directly into the static directory and have a redirect on that file.

erikjohnston:
It seems we already respond to `/.welll-known/matrix/client` requests in Synpase (if you set the `public_baseurl` config). We should document this.

Do you have a use case for also hosting a `/.well-known/matrix/server` entry?

maquis196:
My use case is my current setup, having traefik as an edge router is a fairly simple setup involving labels (if a regular docker setup), my only issue is that traefik doesnt seem to provide a way of providing "virtual locations" like nginx does, so if i want to host the well-known/matrix/server, I need to either spin up a separate nginx host to provide one file (which is the solution ive chosen in the interim) so all requests to https://homeserver/.well-known/matrix/server get routed to nginx which isnt great.

If synapse inside docker could provide this itself via static assets that i could easily say mount directly or provided via synapse with say and environmental value passed to docker, it would make my life easier. Its potentially a pretty niche use case so wanted to see what everyone thought or if there was a way of doing this easily that i haven't considered?

maquis196:
although also, are we able to use SRV records still? not sure ive seen it working for me even though the dns record looks correct

erikjohnston:
OK, sounds like you do also want to host `/.well-known/matrix/server`. That shouldn't be hard to add alongside the current host `/.well-known/matrix/server`.

Though having said that, the `/.well-known/matrix/server` is really only used to route federation traffic elsewhere. If you're already receiving federation traffic into the container then you don't need to route the traffic anywhere else?

> although also, are we able to use SRV records still? not sure ive seen it working for me even though the dns record looks correct

SRV records will work. If you haven't seen https://federationtester.matrix.org/ then that can help you debug any issues.

maquis196:
[...]

PR #11211, opened 2021-10-29
Support for serving server well-known files

richvdh:
Having Synapse serve a server well-known is a way to have federation traffic redirected from port 8448 to port 443, which can be handy.

Fixes https://github.com/matrix-org/synapse/issues/8308.

Also updates the delegation documentation.


docs/delegate.md
```diff
@@ -38,16 +51,29 @@ should return:
 Note, specifying a port is optional. If no port is specified, then it defaults
 to 8448.
 
-With .well-known delegation, federating servers will check for a valid TLS
-certificate for the delegated hostname (in our example: `synapse.example.com`).
+### Serving a `.well-known/matrix/server` file with Synapse
```
reivilibre, line 54: Up to you, but I might have been tempted to put this section first since it's the 'easy' case.

docs/delegate.md
```diff
@@ -68,27 +94,9 @@ wouldn't need any delegation set up.
...
-In that case, the only reason Synapse needs the certificate is to populate a legacy
-`tls_fingerprints` field in the federation API. This is ignored by Synapse 0.99.0
-and later, and the only time pre-0.99 Synapses will check it is when attempting to
-fetch the server keys - and generally this is delegated via `matrix.org`, which
-is running a modern version of Synapse.
-
-### Do I need the same certificate for the client and federation port?
-
-No. There is nothing stopping you from using different certificates,
-particularly if you are using a reverse proxy.
```
reivilibre, line 94: (Ah, this all reminds me of Synapse from a long time ago. Nostalgic.)

docs/sample_config.yaml
```diff
@@ -93,6 +93,23 @@ pid_file: DATADIR/homeserver.pid
 #
 #public_baseurl: https://example.com/
 
+# Uncomment the following to tell other servers to redirect traffic to port 443.
```
reivilibre, line 96:
```suggestion
# Uncomment the following to tell other servers to communicate on port 443.
```

'redirect' reads a little bit wrong here to me; it also brings to mind HTTP redirects and I can see it giving people some imagery that 8448 will serve redirects to 443 or something.

reivilibre (approved):
Very glad this has been done, it felt like one of those 'obvious' things we should have. Thank you!

docs/delegate.md
richvdh: I decided to put it second, because it's only useful in one very specific case. I thought it was clearer to put the more general solution first.

reivilibre: That's reasonable! I think the way you introduced it will give enough hint that there are more options too, so no problem there

docs/delegate.md
richvdh: It's still there in https://anoadragon453.github.io/synapse/latest/upgrading/MSC1711_certificates_FAQ.html if you want the nostalgia hit :)

docs/sample_config.yaml
richvdh: done, thanks.

Issue #10653, opened 2021-08-18
Encrypted appservices: how to?


turt2live:
The end goal of this adventure is to have an appservice which can participate in encrypted rooms. While this doesn't necessarily provide end-to-end encryption, it does mean that encrypted messages can be decrypted by application services. The driving usecase is largely private context (DM/ping me internally), but is effectively a high-traffic single-user highly reliable bot using the appservice API (sync won't keep up). Other usecases like bridges are worth considering, even if not driving this work, due to the fact that someone will try it despite warnings.

There's 3-4 major pieces in order to get this support working:
* Sending `to-device` messages to appservices for their interested users. This has security context to consider.
* Sending device list changes to appservices.
* Sending One Time Key (OTK) counts to appservices for each of their interested users.
* Sending fallback key usage to appservices for each of their interested users.

Other considerations, like setting up key backups, are already largely solved due to the masquerading support on the endpoints. Account data changes might need to make it down to appservices, but this can be considered future work for the purposes of this conversation.

# Sending device messages (`to-device`)

This is effectively sending ephemeral events to appservices, which Synapse already supports for read receipts, typing, and presence. Currently the streams appear to operate off a `from` and `to` token system, grabbing a batch of events between those tokens with some filter criteria, however for appservices the filter criteria is a bit more nuanced. Specifically, the appservice can easily get to millions of users under its namespace which means millions of inboxes to check. We could add an array of user IDs to the DB fetch function for inboxes to check, however this can, again, be millions of entries long: this would be bad for the DB server.

Appservices already have to care about implicit versus explicit user reservations and will likely have to do additional filtering based upon that, so the proposal I have is largely that the appservice handler grab *all* device messages for all inboxes between the stream tokens, then filter that down to interested users in code. The function would build two piles: implicit interest and explicit interest. Explicit interest would cause the function to delete those messages from the inbox as they'll shortly be "delivered". Implicit interest means those messages won't be deleted, but will still be sent to the appservice. This is where the security context comes into play: this provides a way for appservices to theoretically intercept device messages without the user knowing. However, the server admin will have had to approve the appservice implicitly by installing it, so it might be fine. This might need more discussion.

An issue with deleting the device messages is reliability: where the code appears to build the transaction and where that transaction is sent to the appservice is disjointed. This isn't important for things like presence, typing, or read receipts (to a degree), however for device messages if the server were to be restarted while an undelivered transaction is in the queue then the messages will be lost, leading to UISIs (in this use case). 

[...]

Issue #11150, opened 2021-10-21
There is no retry logic for sending EDUs to application services


anoadragon453:
When we send ephemeral events to appservices, we keep track of a stream token per appservice per EDU type. We do this for read receipts and presence updates (but not for typing, those are a little ephemeral to care about).

When a new read receipt comes through the server, we consider whether that should be sent to any connected application services. If so, we send those read receipts off. We then record the stream token of the read receipt for this appservice, even if we ended up determining that the appservice wasn't interested in the event (so that we wouldn't have to check that again later on).

This system works (although has [scalability concerns](https://github.com/matrix-org/synapse/issues/10836)), however there is no retry logic. We record the updated stream token even if sending to the application service fails:

https://github.com/matrix-org/synapse/blob/e584534403b55ad3f250f92592e30b15b01f0201/synapse/handlers/appservice.py#L239-L245

`submit_ephemeral_events_for_as` kicks off a background task which logs and then ignores exceptions:

https://github.com/matrix-org/synapse/blob/4b965c862dc66c0da5d3240add70e9b5f0aa720b/synapse/appservice/scheduler.py#L156-L159

The stream token is stored and updated for this appservice immediately after the background task kicks off. It makes sense to do this so that we do not end up with duplicated work while processing subsequent calls, even if sending to the appservice is slow, however we still need a way of deprecated/holding off updating the stored stream id until a 200 is returned from the appservice. Only then should we mark the appservice as successfully having processed up to that stream token.

This doesn't matter so much with read receipts or presence updates, but will become much more important when we start passing things like device lists over this channel. A blip in the network will lead to decryption errors down the line.

Note that this behaviour for the current set of supported EDUs is intentional:

https://github.com/matrix-org/synapse/blob/e584534403b55ad3f250f92592e30b15b01f0201/synapse/storage/databases/main/appservice.py#L192-L200

The stream tokens are stored in the `application_services_state` table, which has the schema:

```
synapse=# \d application_services_state
                   Table "public.application_services_state"
         Column         |         Type         | Collation | Nullable | Default 
------------------------+----------------------+-----------+----------+---------
 as_id                  | text                 |           | not null | 
 state                  | character varying(5) |           |          | 
 last_txn               | bigint               |           |          | 
 read_receipt_stream_id | bigint               |           |          | 
 presence_stream_id     | bigint               |           |          | 

Indexes:
    "application_services_state_pkey" PRIMARY KEY, btree (as_id)
```

[...]

PR #5156, opened 2019-05-09
Prevent multiple device list updates from breaking a batch send

anoadragon453:
Fixes https://github.com/matrix-org/synapse/issues/5153 and towards fixing #5095.

This PR:

* Ensures we take device list updates from the DB in order of ascending `stream_id`
* Raises the limit of updates from 20 to 100 (presumably this must be kept under 100 EDU transaction limit)
* Removes one more update from the DB and checks to see if its `stream_id` matches that of the update before it. If so, an update is running over the bounds of the batch, and we try to send everything but updates with that device ID.
* If the above results in us not sending any updates, then we simply ignore this update as it's probably not correct behavior anyways.

cc @richvdh

PR #11215, opened 2021-10-29
Send to-device messages to application services

anoadragon453:
Addresses the ToDevice side of #10653. Partially implements [MSC2409](https://github.com/matrix-org/matrix-doc/pull/2409).

This PR adds the ability for Application Services to receive to-device messages when they are sent by local users, or received from remote homeservers. The control flow once this PR lands will work as follows:

<details>

<summary>Control flow</summary>

```
incoming to-device message (from local or remote, destined for a local user)
🠗
message is dropped into the `device_inbox` table, and assigned a to_device stream ID.
Note that one to-device *message* intended for multiple devices will result in a single stream ID being assigned to multiple rows in the `device_inbox` table.
🠗
`Notifier.on_new_event` fires with the stream ID of the new message, as well as the users who should receive it
...
🠗
`ApplicationServicesHandler._notify_interested_services_ephemeral`: for each registered application service, we pull out the rows from the `device_inbox` table that are 1. have a stream ID in the range `last recorded sent token for this appservice < X <= given stream ID` 2. have a recipient that matches the list of user IDs given from `Notifier.on_new_event`.
🠗
`ApplicationServiceSchedular.submit_ephemeral_events_for_as` receives those events along with the service to send them to. It schedules them to be sent, placing them in the `de.sorunome.msc2409.ephemeral` dict of the AS txn.
🠗
The stream ID is recorded as the latest successful stream ID for this application service.
```

</details>

Discussion points:

- [x] Do we want to use a feature flag here?
    * While receiving to-device messages requires opting in to receiving EDUs by including [a flag in the AS registration file](https://github.com/matrix-org/matrix-doc/pull/2409/files#diff-3f2c9c82b30beec57491d80df02d93c08e1a69c6e4e57899681267a0d51baa63R19-R20), some homeserver admins may be using this feature for existing EDUs without wanting to try out experimental to-device message sending. Thus, adding a feature flag for this under the [MSC2409](https://github.com/matrix-org/matrix-doc/pull/2409) namespace seems sensible.
- [x] How do we want to tackle deleting messages?
    * We want to delete to-device messages that have been confirmed delivered to the AS that's exclusively registered for the users that are set to receive them. Let's delete those messages after we receive confirmation of delivery to the AS.
    * **This will be done in a subsequent PR.** Tracked at https://github.com/matrix-org/synapse/issues/11423.
- [x] Capturing to-device messages being sent to a remote user.
    * **This will be done in a subsequent PR.** (It's a separate stream entirely.)
- [x] When an application service is added for the first time to the homeserver, do we want it to receive historical to-device messages for the users in its namespace?
    * TravisR says no, and I agree. The AS should still receive missed messages if it was down, but that's part of the later reliability step.
    * **This will be done in a subsequent PR.** Tracked at https://github.com/matrix-org/synapse/issues/10836.
- [x] Tests. (It would be nice to add some to Complement as well, however Complement doesn't yet support application services (https://github.com/matrix-org/complement/issues/24), and I don't wish to block this PR on adding that support)
     - [x] Sending to-device messages to {exclusive, non-exclusive} appservice user namespaces work.
     - [x] Sending messages to users that aren't in an AS's user namespace shouldn't be received by that AS. 
     - [x] Does this handle large bursts of messages?

Reviewable commit-by-commit. Note that a large chunk of the diff is tests. ~~Based on https://github.com/matrix-org/synapse/pull/11247.~~


synapse/storage/databases/main/deviceinbox.py
```diff
@@ -116,6 +120,97 @@ def process_replication_rows(self, stream_name, instance_name, token, rows):
...
+        # Bail out if none of these users have any messages
+        for user_id in user_ids:
+            if self._device_inbox_stream_cache.has_entity_changed(
+                user_id, from_stream_id
+            ):
+                break
+        else:
+            return {}, to_stream_id
+
+        def get_new_messages_txn(txn):
```
reivilibre, line 162: type anno please

synapse/storage/databases/main/deviceinbox.py
```diff
@@ -116,6 +120,97 @@ def process_replication_rows(self, stream_name, instance_name, token, rows):
...
+                    (recipient_user_id, recipient_device_id), []
+                ).append(message_dict)
+                total_messages_processed += 1
+
+            # If we don't end up hitting the limit, we still want to return the equivalent
+            # value of `to_stream_id` to the calling function. This is needed as we'll
+            # be processing up to `to_stream_id`, without necessarily fetching a message
+            # with a stream id of `to_stream_id`.
+            if total_messages_processed < limit:
+                stream_pos = to_stream_id
```
reivilibre, line 206: I'm confused about this part.

reivilibre (commented):
Some thoughts

synapse/storage/databases/main/deviceinbox.py
```diff
@@ -116,6 +120,97 @@ def process_replication_rows(self, stream_name, instance_name, token, rows):
...
+        limit: int = 100,
+    ) -> Tuple[Dict[Tuple[str, str], List[JsonDict]], int]:
+        """
+        Retrieve to-device messages for a given set of user IDs.
+
+        Only to-device messages with stream ids between the given boundaries
+        (from < X <= to) are returned.
+
+        Note that multiple messages can have the same stream id. Stream ids are
+        unique to *messages*, but there might be multiple recipients of a message, and
```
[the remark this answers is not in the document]
anoadragon453: So what this is trying to say is that you can have a single message A, which you send to multiple devices. You'll then end up with multiple rows in `device_inbox`, as each row specifies unique (device_id, message) tuples. The rows that are the result of a single action will have the same stream ID, however. Thus you can end up with multiple rows relating to the same "message", with the same stream ID.

Does that make sense? The above definition of message really means "message content". I suppose you could argue that sending the same content to different people means that each instance of that is a separate "message". They will still all have matching stream ID's though.

As an example from my local test DB:

```
synapse=> select * from device_inbox order by stream_id desc limit 5;
      user_id      |                  device_id                  | stream_id |     message_json   | instance_name 
-------------------+---------------------------------------------+-----------+--------------------+---------------
 @admin:localhost  | UPRDRTCDLI                                  |        37 | {"content":{"..."} | master
 @admin:localhost  | LUGNTXIHIQ                                  |        37 | {"content":{"..."} | master
 @admin:localhost  | XKCKVAPGKH                                  |        37 | {"content":{"..."} | master
...
```

Which presumably was the result of a [sendToDevice call](https://matrix.org/docs/spec/client_server/r0.6.1#put-matrix-client-r0-sendtodevice-eventtype-txnid) like:

```jsonc
PUT /_matrix/client/r0/sendToDevice/m.new_device/123 HTTP/1.1
Content-Type: application/json

{
  "messages": {
    "@admin:localhost": {
      "*": {
        // message content
      }
    }
  }
}
```

synapse/storage/databases/main/deviceinbox.py
```diff
@@ -116,6 +120,97 @@ def process_replication_rows(self, stream_name, instance_name, token, rows):
...
+            many_clause_sql, many_clause_args = make_in_list_sql_clause(
+                self.database_engine, "user_id", user_ids
+            )
+
+            sql += (
+                " WHERE %s"
+                " AND ? < stream_id AND stream_id <= ?"
+                " ORDER BY stream_id ASC"
+                " LIMIT ?"
+            ) % many_clause_sql
```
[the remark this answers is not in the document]
anoadragon453: I agree that it is easier to read, and GitHub looks to be automatically syntax highlighting the SQL as well.

So I'm on board!

Edit: done in d9f32492aba7dca3171e66d72417951ef5fd29ee.

synapse/storage/databases/main/deviceinbox.py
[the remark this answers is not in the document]
reivilibre: I think I see what you mean.
I wonder if you think phrasing it like:

> A stream ID can be shared by multiple copies of the same message with different recipient devices

would help?

anoadragon453:
I've rebased this PR on the latest develop. Tests to come soon.

synapse/storage/databases/main/deviceinbox.py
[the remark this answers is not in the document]
anoadragon453: That is much clearer yes, thank you! I've added your suggested wording in 0453d5e53e.

reivilibre:
(self-requesting review because I need to read and understand, but I'm not in a position to properly review the code)

richvdh:
>  The control flow currently works as follows:

before or after this PR?

synapse/handlers/appservice.py
```diff
@@ -440,6 +464,76 @@ async def _handle_presence(
...
+                # Remove 'message_id' from the to-device message, as it's an internal ID
+                message_json.pop("message_id", None)
+
+                message_payload.append(
+                    {
+                        "to_user_id": user_id,
+                        "to_device_id": device_id,
+                        **message_json,
+                    }
+                )
```
richvdh, line 533: this is more of a matter for the MSC, but you can't just smash together to-device messages and EDUs into the same namespace. cf https://github.com/matrix-org/matrix-doc/pull/2409#discussion_r752211109

synapse/config/experimental.py
```diff
@@ -46,3 +46,11 @@ def read_config(self, config: JsonDict, **kwargs):
 
         # MSC3266 (room summary api)
         self.msc3266_enabled: bool = experimental.get("msc3266_enabled", False)
+
+        # MSC2409 (this setting only relates to optionally sending to-device messages).
```
richvdh, line 50: it is ... frustrating... that it doesn't apply to other things covered by MSC2409. still, not relevant here.

synapse/handlers/appservice.py
```diff
@@ -199,8 +202,9 @@ def notify_interested_services_ephemeral(
         Args:
             stream_key: The stream the event came from.
 
-                `stream_key` can be "typing_key", "receipt_key" or "presence_key". Any other
-                value for `stream_key` will cause this function to return early.
+                `stream_key` can be "typing_key", "receipt_key", "presence_key" or
+                "to_device_key". Any other value for `stream_key` will cause this function
+                to return early.
```
richvdh, line 207: > Any other value for `stream_key` will cause this function to return early.

will it? it looks like you've removed the relevant return statement. (which seems a surprising choice, but eh)

synapse/handlers/appservice.py
```diff
@@ -55,6 +55,9 @@ def __init__(self, hs: "HomeServer"):
         self.clock = hs.get_clock()
         self.notify_appservices = hs.config.appservice.notify_appservices
         self.event_sources = hs.get_event_sources()
+        self.msc2409_to_device_messages_enabled = (
```
richvdh, line 58: private fields get underscores. Even in classes that do not currently follow the convention.

```suggestion
        self._msc2409_to_device_messages_enabled = (
```

synapse/handlers/appservice.py
```diff
@@ -307,6 +314,23 @@ async def _notify_interested_services_ephemeral(
                             service, "presence", new_token
                         )
 
+                    elif (
+                        stream_key == "to_device_key"
+                        and self.msc2409_to_device_messages_enabled
```
richvdh, line 319: should this be an `assert` rather than an `and` ? We should already have checked `msc2409_to_device_messages_enabled`?

synapse/handlers/appservice.py
```diff
@@ -307,6 +314,23 @@ async def _notify_interested_services_ephemeral(
...
+                    ):
+                        # Retrieve an iterable of to-device message events, as well as the
```
richvdh, line 321: an iterable is insufficient, given we need to check if it is non-empty.

also, `submit_ephemeral_events_for_as` claims to want a list rather than an iterable (though looking at it, I think the annotation is overly specific)

synapse/handlers/appservice.py
```diff
@@ -307,6 +314,23 @@ async def _notify_interested_services_ephemeral(
...
+                        # maximum stream token of the messages we were able to retrieve.
+                        events = await self._handle_to_device(service, new_token, users)
+                        if events:
```
richvdh, line 324: feels like this `if` condition should be inside `submit_ephemeral_events_for_as` rather than having to repeat it at every single call site.

synapse/handlers/appservice.py
```diff
@@ -307,6 +314,23 @@ async def _notify_interested_services_ephemeral(
...
+                            self.scheduler.submit_ephemeral_events_for_as(
+                                service, events
+                            )
+
+                        # Persist the latest handled stream token for this appservice
+                        await self.store.set_type_stream_id_for_appservice(
```
richvdh, line 330: :thinking: what a funny name this method has :/

synapse/handlers/appservice.py
```diff
@@ -440,6 +464,76 @@ async def _handle_presence(
 
         return events
 
+    async def _handle_to_device(
```
richvdh, line 467: maybe `_get_to_device_messages` or something?

synapse/handlers/appservice.py
```diff
@@ -307,6 +314,23 @@ async def _notify_interested_services_ephemeral(
...
+                        events = await self._handle_to_device(service, new_token, users)
+                        if events:
+                            self.scheduler.submit_ephemeral_events_for_as(
+                                service, events
+                            )
+
+                        # Persist the latest handled stream token for this appservice
+                        await self.store.set_type_stream_id_for_appservice(
+                            service, "to_device", new_token
+                        )
```
richvdh, line 332: I haven't chased through the code: once the events are sent to  `submit_ephemeral_events_for_as`, will they reliably get delivered?

synapse/handlers/appservice.py
```diff
@@ -440,6 +464,76 @@ async def _handle_presence(
...
+        """
+        # Get the stream token that this application service has processed up until
+        from_key = await self.store.get_type_stream_id_for_appservice(
+            service, "to_device"
+        )
+
+        # Filter out users that this appservice is not interested in
+        users_appservice_is_interested_in: List[str] = []
+        for user in users:
+            if isinstance(user, UserID):
```
richvdh, line 495: facepalm. Can't we go and fix the calling code so we don't have to fuck about with this?

(rhetorical question, feel free to park it for now)

synapse/storage/databases/main/deviceinbox.py
```diff
@@ -136,6 +137,79 @@ def process_replication_rows(self, stream_name, instance_name, token, rows):
     def get_to_device_stream_token(self):
         return self._device_inbox_id_gen.get_current_token()
 
+    async def get_new_messages(
```
richvdh, line 140: this needs a name that reflects the fact it is specific to to-device messages.

synapse/storage/databases/main/deviceinbox.py
```diff
@@ -136,6 +137,79 @@ def process_replication_rows(self, stream_name, instance_name, token, rows):
...
+            return {}
+
+        def get_new_messages_txn(txn: LoggingTransaction):
+            # Build a query to select messages from any of the given users that are between
+            # the given stream id bounds
+
+            # Scope to only the given users. We need to use this method as doing so is
+            # different across database engines.
+            many_clause_sql, many_clause_args = make_in_list_sql_clause(
+                self.database_engine, "user_id", user_ids
```
richvdh, line 180: we could limit the `user_ids` here to those where `has_entity_changed` is true.

synapse/storage/databases/main/deviceinbox.py
```diff
@@ -136,6 +137,79 @@ def process_replication_rows(self, stream_name, instance_name, token, rows):
...
+            )
+
+            sql = f"""
+                SELECT user_id, device_id, message_json FROM device_inbox
+                WHERE {many_clause_sql}
+                AND ? < stream_id AND stream_id <= ?
+                ORDER BY stream_id ASC
```
richvdh, line 187: have you checked what the optimiser does with this? I can't remember what indexes we have, but i suspect if you have more than one or two users, it's going to start table-scanning.

synapse/storage/databases/main/deviceinbox.py
```diff
@@ -136,6 +137,79 @@ def process_replication_rows(self, stream_name, instance_name, token, rows):
...
+                recipient_user_id_device_id_to_messages.setdefault(
+                    (recipient_user_id, recipient_device_id), []
+                ).append(message_dict)
+
+            return recipient_user_id_device_id_to_messages
+
+        return await self.db_pool.runInteraction(
+            "get_new_messages", get_new_messages_txn
+        )
+
```
richvdh, line 212: I'm rather uncomfortable that we're duplicating  `get_new_messages_for_device`. Can't we figure out the device ids and use `get_new_messages_for_device` ?

richvdh (changes requested):

anoadragon453:
> before or after this PR?

ah, afterwards.

synapse/handlers/appservice.py
```diff
@@ -233,6 +233,13 @@ def notify_interested_services_ephemeral(
         # Additional context: https://github.com/matrix-org/synapse/pull/11137
         assert isinstance(new_token, int)
 
+        # Ignore to-device messages if the feature flag is not enabled
+        if (
+            stream_key == "to_device_key"
```
richvdh, line 238: didn't we (you) invent constants for this?

synapse/notifier.py
```diff
@@ -444,15 +444,23 @@ def on_new_event(
...
-            except Exception:
-                logger.exception("Error notifying application services of event")
+            # Notify appservices of updates in ephemeral event streams.
+            # Only the following streams are currently supported.
+            if stream_key in (
+                "typing_key",
+                "receipt_key",
+                "presence_key",
+                "to_device_key",
+            ):
```
richvdh, line 454: I didn't really understand the logic for moving this out of `appservice_handler`. Isn't the choice of what streams to care about a matter for the appservice handler, not the notifier?

synapse/storage/schema/main/delta/65/06_msc2409_add_device_id_appservice_stream_type.sql
```diff
@@ -0,0 +1,18 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+-- Add a column to track what to_device stream id that this application
+-- service has been caught up to.
+ALTER TABLE application_services_state ADD COLUMN to_device_stream_id BIGINT;
```
richvdh, line 18: could this be NOT NULL?

tests/handlers/test_appservice.py
```diff
@@ -321,3 +325,241 @@ def _mkservice_alias(self, is_interested_in_alias):
...
+        synapse.rest.admin.register_servlets_for_client_rest_resource,
+        login.register_servlets,
+        room.register_servlets,
+        sendtodevice.register_servlets,
+        receipts.register_servlets,
+    ]
+
+    def prepare(self, reactor, clock, hs):
+        # Mock the application service scheduler so that we can track any
+        # outgoing transactions
```
richvdh, line 341: given we're mocking this bit out, it might be better to call the class `AppserviceHandlerEphemeralEventsTestCase`. Also worth adding a docstring explaining what exactly is, or is not, in scope.

I hate having to read all the way through a testclass to understand what it's trying to test.

tests/handlers/test_appservice.py
```diff
@@ -321,3 +325,241 @@ def _mkservice_alias(self, is_interested_in_alias):
...
+
+            # Add to the count of messages for this application service
+            service_id_to_message_count.setdefault(service.id, 0)
+            service_id_to_message_count[service.id] += len(events)
+
+        # Assert that each interested service received the full count of messages
+        for count in service_id_to_message_count.values():
+            self.assertEqual(count, number_of_messages)
+
+    def _register_interested_and_uninterested_application_services(
```
richvdh, line 505: why do we need to register all the appservices at once? why not just have a single `_register_application_service` which adds the appservice to a `self._services` field in the test class (which will be returned by the mock)?

tests/handlers/test_appservice.py
```diff
@@ -321,3 +325,241 @@ def _mkservice_alias(self, is_interested_in_alias):
...
+
+        hs.get_application_service_handler().scheduler = self.mock_scheduler
+
+        self.device1 = "device1"
+        self.user1 = self.register_user("user1", "password")
+        self.token1 = self.login("user1", "password", self.device1)
+
+        self.device2 = "device2"
+        self.user2 = self.register_user("user2", "password")
+        self.token2 = self.login("user2", "password", self.device2)
```
richvdh, line 353: worth adding some comments about how these users/devices are used later in the tests. Sounds like `user1` is a generic non-AS user, whereas `user2` is used as an exclusive AS user?

tests/handlers/test_appservice.py
```diff
@@ -321,3 +325,241 @@ def _mkservice_alias(self, is_interested_in_alias):
...
+        self.assertEqual(chan.code, 200, chan.result)
+
+        # Check if our application service - that is interested in user2 - received
+        # the to-device message as part of an AS transaction.
+        # Only the user1 -> user2 to-device message should have been forwarded to the AS.
+        #
+        # The uninterested application service should not have been notified at all.
+        self.assertEqual(
+            1, self.mock_scheduler.submit_ephemeral_events_for_as.call_count
+        )
```
richvdh, line 396: you could `assert_called_once` :man_shrugging:

tests/handlers/test_appservice.py
```diff
@@ -321,3 +325,241 @@ def _mkservice_alias(self, is_interested_in_alias):
...
+                        }
+                    ],
+                },
+            )
+            interested_services.append(interested_service)
+
+        for _ in range(uninterested_count):
+            uninterested_services.append(self._make_application_service())
+
+        # Register this application service, along with another, uninterested one
```
richvdh, line 544: this comment looks a bit outdated

richvdh (changes requested):

tests/handlers/test_appservice.py
anoadragon453: They don't need to be registered all at once. This method is for convenience, but perhaps things are abstracted a bit too far.

The design you propose does sound simpler. I'll give that a shot.

synapse/storage/databases/main/deviceinbox.py
anoadragon453: Standard indexes for this table are:

```
synapse=> \d device_inbox
               Table "public.device_inbox"
    Column     |  Type  | Collation | Nullable | Default 
---------------+--------+-----------+----------+---------
 user_id       | text   |           | not null | 
 device_id     | text   |           | not null | 
 stream_id     | bigint |           | not null | 
 message_json  | text   |           | not null | 
 instance_name | text   |           |          | 
Indexes:
    "device_inbox_stream_id_user_id" btree (stream_id, user_id)
    "device_inbox_user_stream_id" btree (user_id, device_id, stream_id)

synapse=> 
```

It seems that adding device IDs to this query speed it up significantly, especially when scanning a large range of stream IDs (e.g if the AS has been offline for a while):

<details>

<summary>query with no device IDs specified</summary>

```
synapse=# EXPLAIN ANALYZE -- no device IDs specified
SELECT user_id, device_id, message_json FROM device_inbox
WHERE user_id in ('@andrewm:amorgan.xyz', '@testguest:amorgan.xyz', '@telegram_68411305:amorgan.xyz', '@telegram_11305:amorgan.xyz', '1', '2', '3', '4', '5', '6', '7', '8', '9', '10', '11', '12', '13', '14', '15', '16', '17', '18', '19', '20')                                           
AND 1000000 < stream_id AND stream_id <= 1518046
ORDER BY stream_id ASC;
                                                                                            QUERY PLAN                                                                                             
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Index Scan using device_inbox_stream_id on device_inbox  (cost=0.43..598059.96 rows=3197064 width=453) (actual time=0.581..10466.706 rows=3257282 loops=1)
   Index Cond: ((1000000 < stream_id) AND (stream_id <= 1518046))
   Filter: (user_id = ANY ('{@andrewm:amorgan.xyz,@testguest:amorgan.xyz,@telegram_68411305:amorgan.xyz,@telegram_11305:amorgan.xyz,1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19,20}'::text[]))
   Rows Removed by Filter: 46549
 Planning time: 0.630 ms
 Execution time: 10654.986 ms
(6 rows)
```

</details>

The largest slowdown (on amorgan.xyz) looks to be down to the size of the stream ID range - where a range size this large (518046) is indeed quite slow. Throwing in some device IDs though:

<details>

<summary>specifying some device IDs</summary>

```
synapse=# EXPLAIN ANALYZE                                                      
SELECT user_id, device_id, message_json FROM device_inbox
WHERE user_id in ('@andrewm:amorgan.xyz', '@testguest:amorgan.xyz', '@telegram_68411305:amorgan.xyz', '@telegram_11305:amorgan.xyz', '1', '2', '3', '4', '5', '6', '7', '8', '9', '10', '11', '12', '13', '14', '15', '16', '17', '18', '19', '20')
AND device_id in ('XFHILLDHIT', 'CJYNODEIGN', 'TKYYIPMJAN', 'WQWGGLVYVX', 'ODIEEXSYLA')
AND 1000000 < stream_id AND stream_id <= 1518046
ORDER BY stream_id ASC;

                                                                                                                                                                                QUERY PLAN                                                                                                                                                                                
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Sort  (cost=27251.02..27275.14 rows=9649 width=453) (actual time=1595.313..1602.499 rows=31182 loops=1)
   Sort Key: stream_id
   Sort Method: external merge  Disk: 17824kB
   ->  Index Scan using device_inbox_user_stream_id on device_inbox  (cost=0.56..24631.44 rows=9649 width=453) (actual time=1.719..1563.038 rows=31182 loops=1)
         Index Cond: ((user_id = ANY ('{@andrewm:amorgan.xyz,@testguest:amorgan.xyz,@telegram_68411305:amorgan.xyz,@telegram_11305:amorgan.xyz,1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19,20}'::text[])) AND (device_id = ANY ('{XFHILLDHIT,CJYNODEIGN,TKYYIPMJAN,WQWGGLVYVX,ODIEEXSYLA,VVDOQSPBZK}'::text[])) AND (1000000 < stream_id) AND (stream_id <= 1518046))
 Planning time: 0.194 ms
 Execution time: 1607.424 ms
(7 rows)
```

</details>

and we take advantage of our `device_inbox_user_stream_id` index.

In practice, the range of stream IDs could grow quite large in practice if the AS has been offline for a while. I'm not sure if the set of users/devices here would realistically get that large (unless a user had a *ton* of devices).

synapse/handlers/appservice.py
anoadragon453: I think I asked about where to put them a few weeks ago. It's been on my mind to convert these to constants once all this in-flight stuff had settled.

synapse/storage/databases/main/deviceinbox.py
anoadragon453: Blocked on https://github.com/matrix-org/synapse/pull/11215/files#r752249314.

synapse/handlers/appservice.py
richvdh: I've got to say, it's sufficiently broken that I'm not terribly happy landing it in synapse mainline as it currently is.

synapse/handlers/appservice.py
richvdh: sounds better!

synapse/handlers/appservice.py
richvdh: it's probably not worth blocking right now. Suggest you add some FIXME-style comments here to note the outstanding work.

synapse/storage/schema/main/delta/65/06_msc2409_add_device_id_appservice_stream_type.sql
richvdh: > The rest of the columns in application_services_state are currently nullable:

yes, but synapse's schema is generally messy. We shouldn't set too much store by what happened in the past.

> For instance, we plan to not send historical to-device messages to an application service when it's first connected. 

hrm, this sounds like it might be a useful use of NULL here then. Has this code landed yet? I don't remember seeing it.

But if you're assigning a special meaning to NULL, you need to document it in the schema definition.

> On the other hand, what benefits would NOT NULL bring?

Mostly the same benefits of not using a TEXT field: less storage overhead, less having to worry about what to do when the data doesn't match your expected format.  Let the database do the validation for us.

synapse/handlers/appservice.py
richvdh: given the other stuff you're changing, I think this is trivial. Just take what you've got and stick it somewhere other than `ephemeral`.

turt2live: ftr the MSC has moved these to a `to_device` array at the top level of the transaction.

anoadragon453: Ah, OK. I was worried about blocking this PR on changes to the MSC. But given that's now gone through and everyone will be using the MSC as the source of truth, it makes sense to me to update the PR.

synapse/storage/schema/main/delta/65/06_msc2409_add_device_id_appservice_stream_type.sql
anoadragon453: > hrm, this sounds like it might be a useful use of NULL here then. Has this code landed yet? I don't remember seeing it.

It hasn't landed, no. From the PR description:

> - [ ] When an application service is added for the first time to the homeserver, do we want it to receive historical to-device messages for the users in its namespace?
>    * TravisR says no, and I agree. The AS should still receive missed messages if it was down, but that's part of the later reliability step.
>    * **This will be done in a subsequent PR.**

Adding the meaning as a comment to the schema definition is a good idea though. I hadn't thought of that. I can do that here while we're adding the migration.

tests/handlers/test_appservice.py
anoadragon453: Done in https://github.com/matrix-org/synapse/pull/11215/commits/bd9d963af2c2c33774b35652087a37ac7c0114a9.

synapse/handlers/appservice.py
anoadragon453: Good idea. I've added comments for the following outstanding items from this PR in c691ef0992da5ad3b4998a698a39b97acfb81aea:

* adding constants for the various stream keys
* fixing a user being either a `str` or a `UserID` when being passed to `notify_interested_services_ephemeral`

There's some existing issues for the outstanding work as well:

* EDU -> AS reliability (https://github.com/matrix-org/synapse/issues/11150)
* deleting to-device messages intended for exclusive namespace users after they've been successfully delivered to the AS (https://github.com/matrix-org/synapse/issues/11423)
* adding logic for not spamming AS's with historical to-device events when they're first connected (https://github.com/matrix-org/synapse/issues/10836)

synapse/handlers/appservice.py
anoadragon453: As noted in https://github.com/matrix-org/synapse/pull/11215/files#r752237473, a todo comment has been added to clean this up later.

synapse/handlers/appservice.py
anoadragon453: As per https://github.com/matrix-org/synapse/pull/11215/files#r752237473, I've added a todo comment for this task.

synapse/storage/databases/main/deviceinbox.py
anoadragon453: Blocked on https://github.com/matrix-org/synapse/pull/11215/files#r752249314.

synapse/storage/schema/main/delta/65/06_msc2409_add_device_id_appservice_stream_type.sql
anoadragon453: Added the comment in 7cf6ad9197.

synapse/config/experimental.py
richvdh: no, I think that will be more confusing. I'm just grumbling.

synapse/storage/databases/main/deviceinbox.py
richvdh: > But this becomes more complicated when dealing with multiple (user_id, device_id) combinations, as the AS API would like to do.

I'm not sure I entirely follow this. As long as you `ORDER BY stream_id`, you will know that you have complete results for all but the last returned `stream_id` - so you just need to drop the last, potentially incomplete, `stream_id` from the results before returning them to the caller.

synapse/storage/schema/main/delta/65/06_msc2409_add_device_id_appservice_stream_type.sql
```diff
@@ -0,0 +1,23 @@
...
+ * limitations under the License.
+ */
+
+-- Add a column to track what to_device stream id that this application
+-- service has been caught up to.
+
+-- We explicitly don't set this field as "NOT NULL", as having NULL as a possible
+-- state is useful for determining if we've ever sent traffic for a stream type
+-- to an appservice. See https://github.com/matrix-org/synapse/issues/10836 for
+-- one way this can be used.
```
richvdh, line 22: thanks for adding this, but I think it is a bit verbose!

```suggestion
-- NULL indicates that this appservice has never received any to_device messages. This
-- can be used to avoid sending a huge dump of messages at startup.
```

synapse/handlers/appservice.py
anoadragon453: As of [this set of commits](https://github.com/matrix-org/synapse/pull/11215#commits-pushed-6d68b8a), we now place to-device messages in the right place. That being under `de.sorunome.msc2409.to_device` at the top of the AS transaction, instead of `de.sorunome.msc2409.ephemeral`.

synapse/storage/schema/main/delta/65/06_msc2409_add_device_id_appservice_stream_type.sql
anoadragon453: This feels like it was more of a note-to-self for future work than a comment really meant for anyone else. I'll keep it in mind for myself :)

synapse/storage/databases/main/deviceinbox.py
anoadragon453: What will the caller than call the method with next time though? If they try again with the last returned `stream_id`, at best they'll get some duplicated messages, and at worst if the number of messages belonging to a single stream ID > the limit, then we'll get in a never-ending loop. `stream_id - 1` has the same problem, and `stream_id + 1` will cause messages to be missed.

An alternative approach:

The caller provides `from` and `to` tokens, as well as an offset value. The storage function returns the latest processed `stream_id` as well as an offset, which the caller then hands right back on the next call, and so on.

This way, if we end up hitting our limit in the middle of a `stream_id`s group of messages, we can identify how much of that group was returned on the last call. We then just offset through that same group on the next call.

We'd need to ensure a `stream_id`s group of messages are always consistently ordered (`ORDER BY stream_id, device_id`). We'd also have to store this offset in the database, though that's just another column.

I think that should work, though there's likely a simpler way to do it...

anoadragon453:
@richvdh requested review as I'm blocked on a resolution to https://github.com/matrix-org/synapse/pull/11215#discussion_r782436115.

synapse/storage/databases/main/deviceinbox.py
richvdh: Following on from your example; suppose we call `get_new_messages_for_devices` with `current_stream_id=0, limit=2`. We add one to the limit, and make the DB query, which returns the following:

```
      user_id      |                  device_id                  | stream_id |     message_json   | instance_name 
-------------------+---------------------------------------------+-----------+--------------------+---------------
 @alice:localhost  | HAKDUDVZXZ                                  |         1 | {"content":{"A"} | master
 @alice:localhost  | HAKDUDVZXZ                                  |         2 | {"content":{"B"} | master
 @bob:localhost    | TYCGLKNPVJ                                  |         2 | {"content":{"B"} | master
```

We can now tell that, at the limit (2), we would have incomplete results for `stream_id==2`. So, `get_new_messages_for_devices` *drops*all the messages with `stream_id==2`, and returns the remainder, with `stream_pos=1`.

Next time round, the caller passes the returned `stream_pos` (ie, 1) back as `current_stream_id` (noting that `current_stream_id` is *ex*clusive) and gets all the messages with stream_id=2.

(You might remember this algorithm from https://github.com/matrix-org/synapse/pull/5156/files#diff-d9219a3564a55443579532e051d6abc32f47e8f4ba44c64459cd4f9d378e22bdR90-R121).

You're right that it fails if there are more messages with a given `stream_id` for the given devices than there are at the limit. But assuming we're talking about a limit of say 100, that would mean that the AS would have to be managing at least 100 devices... how likely a scenario is this?

anoadragon453: > You're right that it fails if there are more messages with a given stream_id for the given devices than there are at the limit. But assuming we're talking about a limit of say 100, that would mean that the AS would have to be managing at least 100 devices... how likely a scenario is this?

Pretty likely I'm afraid - current end-to-end encryption bridge architectures require at least one device per bridged user.

richvdh: bleh. In which case, maybe we're better off with your earlier idea of 

> we could just have public methods which only accept:
> 
>     * multiple `(user_id, device_id)` pairs, no limit
>     * a single `(user_id, device_id)` pair, limit
> 
> and both call to a unified, private method.

having to store an offset into a given stream_id feels pretty icky to me.

anoadragon453: https://github.com/matrix-org/synapse/pull/11215/commits/822e92a62f35f52a1cefe42d5af6bed75da93c62 refactors the two previous methods into two public methods (`get_messages_for_device` and `get_messages_for_user_devices`) which both call the private method `_get_device_messages`.

A `limit` option is only allowed for `get_messages_for_device`, which takes a single user ID / device ID pair.

synapse/storage/databases/main/deviceinbox.py
anoadragon453: `_get_device_messages` (implemented in https://github.com/matrix-org/synapse/commit/822e92a62f35f52a1cefe42d5af6bed75da93c62) queries using the `device_id` of the `device_inbox_user_stream_id`, resulting in a much faster query.

synapse/storage/databases/main/deviceinbox.py
anoadragon453: Done in https://github.com/matrix-org/synapse/commit/822e92a62f35f52a1cefe42d5af6bed75da93c62.

anoadragon453:
@richvdh besides trial old-deps failing, which has stumped me, this is now ready for a re-review!

The bulk of the new changes are in https://github.com/matrix-org/synapse/commit/822e92a62f35f52a1cefe42d5af6bed75da93c62, which refactor the two, somewhat duplicated storage methods into private method - exposing two public interfaces to it. One allows for a `limit`, one does not.

I then ran into some typing issues with `SyncConfig.device_id` being an `Optional[str]`, rather than a `str`. This prompted the creation of https://github.com/matrix-org/synapse/issues/11829, and was worked around in https://github.com/matrix-org/synapse/pull/11215/commits/026cb8a5ca5f43d1b42e31ec500d97f4947a3535.

synapse/storage/databases/main/deviceinbox.py
```diff
@@ -136,63 +137,241 @@ def process_replication_rows(self, stream_name, instance_name, token, rows):
...
+            if len(user_ids) > 1 or len(device_ids) > 1:
+                raise AssertionError(
+                    "Programming error: _get_new_device_messages was passed 'limit' "
+                    "with >1 user id/device id pair"
+                )
+
+        user_ids_to_query: Set[str] = set()
+        device_ids_to_query: Set[str] = set()
+
+        if device_ids is not None:
```
richvdh, line 273: seems like a non-empty `device_ids` list should only be permitted when there is exactly one `user_id` (since device_ids are only unique to a given user, and you might otherwise end up querying a given device_id for the wrong user)

synapse/storage/databases/main/deviceinbox.py
```diff
@@ -136,63 +137,241 @@ def process_replication_rows(self, stream_name, instance_name, token, rows):
...
+                user_ids_to_query.add(user_id)
+
+        def get_device_messages_txn(txn: LoggingTransaction):
+            # Build a query to select messages from any of the given devices that
+            # are between the given stream id bounds.
+
+            # If a list of device IDs was not provided, retrieve all devices IDs
+            # for the given users. We explicitly do not query hidden devices, as
+            # hidden devices should not receive to-device messages.
+            if not device_ids:
```
richvdh, line 293: can we just omit the device_id part of the query in this case? (ie, set `device_id_many_clause_sql="1"` and `device_id_many_clause_args=()` )

synapse/storage/databases/main/deviceinbox.py
```diff
@@ -136,63 +137,241 @@ def process_replication_rows(self, stream_name, instance_name, token, rows):
...
+                recipient_user_id = row[1]
+                recipient_device_id = row[2]
+                message_dict = db_to_json(row[3])
+
+                # Store the device details
+                recipient_device_to_messages.setdefault(
+                    (recipient_user_id, recipient_device_id), []
+                ).append(message_dict)
+
+                if limit and message_count == limit:
```
richvdh, line 358: might be easier to do `if txn.rowcount() == limit`

synapse/storage/schema/main/delta/65/06_msc2409_add_device_id_appservice_stream_type.sql
```diff
@@ -0,0 +1,21 @@
+/* Copyright 2021 The Matrix.org Foundation C.I.C
```
richvdh, line 1: achtung! we're up to schema version 68.

richvdh (commented):
looks generally sensible. A few bits and pieces.

synapse/storage/databases/main/deviceinbox.py
anoadragon453: > Though one thing I could do is take this conditional out of the loop, since we're specifying LIMIT in the SQL above...

Have done so.

anoadragon453:
@richvdh this is now ready for a re-review!

The bulk of the new changes are in https://github.com/matrix-org/synapse/commit/822e92a62f35f52a1cefe42d5af6bed75da93c62, which refactor the two, somewhat duplicated storage methods into private method - exposing two public interfaces to it. One allows for a `limit`, one does not.

I then ran into some typing issues with `SyncConfig.device_id` being an `Optional[str]`, rather than a `str`. This prompted the creation of https://github.com/matrix-org/synapse/issues/11829, and was worked around in https://github.com/matrix-org/synapse/pull/11215/commits/026cb8a5ca5f43d1b42e31ec500d97f4947a3535.

synapse/storage/databases/main/deviceinbox.py
anoadragon453: https://github.com/matrix-org/synapse/pull/11215/commits/089e04168f64be088b89cb696dab76eb8b9f68ad tightens up the interface.

synapse/storage/databases/main/deviceinbox.py
```diff
@@ -252,28 +261,32 @@ async def _get_device_messages(
...
-                    "Programming error: _get_new_device_messages was passed 'limit' "
-                    "but not device_ids. This could lead to querying multiple user ID "
-                    "/ device ID pairs, which is not compatible with 'limit'"
+                    "Programming error: 'device_id' cannot be supplied to "
+                    "_get_device_messages when >1 user_id has been provided"
                 )
 
-            if len(user_ids) > 1 or len(device_ids) > 1:
+            # A limit can only be applied when querying for a single user ID / device ID tuple.
+            if limit:
```
richvdh, line 276: nb that a limit of zero is different from a limit of `None`

```suggestion
            if limit is not None:
```

synapse/storage/databases/main/deviceinbox.py
```diff
@@ -252,28 +261,32 @@ async def _get_device_messages(
...
                 raise AssertionError(
-                    "Programming error: _get_new_device_messages was passed 'limit' "
-                    "with >1 user id/device id pair"
+                    "Programming error: _get_device_messages was passed 'limit' "
+                    "with >1 user_id"
                 )
```
richvdh, line 280: we should maybe expand this to complain whenever we are not querying for a specific user/device id.

```suggestion
       # A limit can only be applied when querying for a single user ID / device ID tuple.
       if device_id is None and limit is not None:
            raise AssertionError(
                "Programming error: _get_device_messages was passed 'limit' "
                "without a specific user_id/device_id"
            )
```

synapse/storage/databases/main/deviceinbox.py
```diff
@@ -290,7 +303,7 @@ def get_device_messages_txn(txn: LoggingTransaction):
             # If a list of device IDs was not provided, retrieve all devices IDs
             # for the given users. We explicitly do not query hidden devices, as
             # hidden devices should not receive to-device messages.
```
richvdh, line 305:
```suggestion
            # hidden devices should not receive to-device messages.
            # Note that this is more efficient than just dropping `device_id` from the query,
            # since device_inbox has an index on `(user_id, device_id, stream_id)`
```

richvdh (approved):
couple of nits, but lgtm

synapse/storage/databases/main/deviceinbox.py
anoadragon453: Done in c334eef8ba.

PR #11216, opened 2021-10-29
UserDirectoryStore.search_user_dir: optimize when search term is empty

sumnerevans:
This PR optimizes `UserDirectoryStore.search_user_dir` when the search term that is provided is empty.

### Pull Request Checklist

### DIFF
--- a/synapse/storage/databases/main/user_directory.py
+++ b/synapse/storage/databases/main/user_directory.py
@@ -811,94 +811,132 @@ class UserDirectoryStore(UserDirectoryBackgroundUpdateStore):
         if isinstance(self.database_engine, PostgresEngine):
             full_query, exact_query, prefix_query = _parse_query_postgres(search_term)
 
-            # If enabled, this config option will rank local users higher than those on
-            # remote instances.
-            if self._prefer_local_users_in_search:
-                # This statement checks whether a given user's user ID contains a server name
-                # that matches the local server
-                statement = "* (CASE WHEN user_id LIKE ? THEN 2.0 ELSE 1.0 END)"
-                additional_ordering_statements.append(statement)
+            if not search_term:
+                logger.debug(
+                    "search_user_dir() No search term provided to user_directory, doing plain list"
+                )
+                sql = """
+                    SELECT d.user_id AS user_id, display_name, avatar_url
+                    FROM user_directory_search as t
+                    INNER JOIN user_directory AS d USING (user_id)
+                    WHERE
+                        %s
+                    LIMIT ?
+                """ % (
+                    where_clause,
+                )
+            else:
+                logger.debug(
+                    "search_user_dir() Search term was provided '%s'" % (search_term)
+                )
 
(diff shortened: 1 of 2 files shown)


synapse/storage/databases/main/user_directory.py
```diff
@@ -811,94 +811,132 @@ async def search_user_dir(
...
-                            vector,
-                            to_tsquery('simple', ?),
-                            8
+            if not search_term:
+                logger.debug(
+                    "search_user_dir() No search term provided to user_directory, doing plain list"
+                )
+                sql = """
+                    SELECT d.user_id AS user_id, display_name, avatar_url
+                    FROM user_directory_search as t
```
reivilibre, line 820: It doesn't look to me like this table is used anymore in this query; it might be sensible to select directly from `user_directory`.

I wonder if this should be ordered? Of course, there's no ranking information since there's no query, but alphabetical ordering by display name and then user ID seems like it would be sensible (depending on what indices are available)?

There's also a setting, `self._prefer_local_users_in_search`, which this optimised query seems to ignore — optionally influencing the ordering according to a case statement as below might be necessary.

synapse/storage/databases/main/user_directory.py
```diff
@@ -811,94 +811,132 @@ async def search_user_dir(
...
-                # that matches the local server
-                #
-                # Note that we need to include a comma at the end for valid SQL
-                statement = "user_id LIKE ? DESC,"
-                additional_ordering_statements.append(statement)
+            if not search_term:
+                logger.debug(
+                    "search_user_dir() No search term provided to user_directory, doing plain list"
+                )
+                sql = """
```
reivilibre, line 895: This query seems to be the same as the one for Postgres.

This function is already a bit messy, but duplicating the same query twice might not be helping things.

I wonder if the structure should be (in rough code):

```python
if not search_term:
    ...
elif isinstance(.., PostgresEngine):
    ...
elif isinstance(.., SqliteEngine):
    ...
```

That might also help to reduce the amount of nesting happening in this function.

reivilibre (commented):
A few thoughts as I was passing by — going to leave this up for review because not sure.

I'm guessing that you must have your reasons for contributing this, but any idea of how much performance this is gaining? It's interesting to hear numbers, if you have any.

synapse/storage/databases/main/user_directory.py
richvdh: big +1 to this - it's pretty hard to review as it stands.

I'd actually recommend splitting it into two functions - 200 line functions are no fun.

richvdh:
I'm taking this off the review queue for now, because @reivilibre's initial comments are very sensible, and refactoring as he suggests will make it much easier to review.

sumnerevans:
Hello, sorry it's taken me so long to respond. I've finally had the chance to investigate why this was in our fork on our side, and it turns out we are actually not using this code after all, so I'm going to revert this in our fork, and close this PR. Sorry for taking your time!

PR #11220, opened 2021-10-30
Fix using MSC2716 batch sending with event persistence workers

### LINKED ISSUE #11219: MSC2716 batch sending breaks when using event persistence workers
### Steps to reproduce

- Have a Synapse instance with workers
- Use MSC2716 `/batch_send`ing

Expected result: successful batch send

Actual result: internal server error
```

### Version information
- **Version**: 1.45.1, 1.46.0rc1
- **Platform**: Docker

### PR DESCRIPTION
**Pull request by tulir**
Fixes #11219 

Not entirely sure if this is the correct solution, but it seems to work.

### Pull Request Checklist

### CODE REVIEW

**Comment by tulir** [comment]
Actually looks like this might not be a complete solution. Some batch sends are throwing a different error

```
2021-10-30 15:50:34,470 - synapse.http.server - 97 - ERROR - POST-35 - Failed handle request via 'ReplicationSendEventRestServlet': <SynapseRequest at 0x7f0a4c4567a0 method='POST' uri='/_synapse/replication/send_event/%24KR5qvRUM0BhZFkLNYV1I6gbCc6M7ts8V1kClr1GaRSg/MfgWrObBsn' clientproto='HTTP/1.1' site=8034>
  File "synapse/http/server.py", line 262, in _async_render_wrapper
  File "synapse/http/server.py", line 451, in _async_render
  File "synapse/replication/http/_base.py", line 307, in _check_auth_and_handle
    return await self.response_cache.wrap(
  File "synapse/util/caches/response_cache.py", line 208, in wrap
    return await make_deferred_yieldable(result)
  File "/twisted/internet/defer.py", line 1661, in _inlineCallbacks
  File "synapse/replication/http/send_event.py", line 129, in _handle_request
    event = await self.event_creation_handler.persist_and_notify_client_event(
  File "synapse/handlers/message.py", line 1390, in persist_and_notify_client_event
    ] = await self.store.get_stripped_room_state_from_event_context(
  File "synapse/storage/databases/main/events_worker.py", line 709, in get_stripped_room_state_from_event_context
    assert current_state_ids is not None
AssertionError
2021-10-30 15:50:34,472 - synapse.access.http.8034 - 400 - INFO - POST-35 - 127.0.0.1 - 8034 - {@_tulir_whatsapp_bot:server} Processed request: 0.002sec/0.001sec (0.000sec, 0.000sec) (0.000sec/0.000sec/0) 55B 500 "POST /_synapse/replication/send_event/%24KR5qvRUM0BhZFkLNYV1I6gbCc6M7ts8V1kClr1GaRSg/MfgWrObBsn HTTP/1.1" "Synapse/1.46.0rc1 (b=master,d887ab481)" [0 dbevts]
```

**Comment by tulir** [comment]
I think I got it fixed by not filling `invite_room_state`. I don't think it's necessary to fill that for batch send member events, and the `get_stripped_room_state_from_event_context` assumes the event is not an outlier so it doesn't work

File: synapse/handlers/message.py [hunk 1]
```diff
@@ -1382,13 +1382,14 @@ async def persist_and_notify_client_event(
...
         if event.type == EventTypes.Member:
             if event.content["membership"] == Membership.INVITE:
-                event.unsigned[
-                    "invite_room_state"
-                ] = await self.store.get_stripped_room_state_from_event_context(
-                    context,
-                    self.room_prejoin_state_types,
-                    membership_user_id=event.sender,
-                )
+                if not event.internal_metadata.outlier:
```
MadLittleMods, line 1385: This seems like it would be a problem with any outlier that includes `invite_room_state`. With MSC2716, we create a lot of outliers though which makes sense why it's obvious to encounter.

erikjohnston: I'm a bit confused why this is only a problem when using workers? The same code paths should be hit either way?

MadLittleMods: @erikjohnston Agreed, it seems that way to me as well ("This seems like it would be an existing problem with any `outlier` that includes `invite_room_state`"). I haven't reproduced this myself.

What's a good way to test worker mode if it is that specific?

@tulir Can you share some reproduction steps to lead us here?

synapse/app/generic_worker.py
```diff
@@ -237,6 +238,7 @@ class GenericWorkerSlavedStore(
     SlavedEventStore,
     SlavedKeyStore,
     RoomWorkerStore,
+    RoomBatchStore,
```
MadLittleMods, line 241: @tulir Might be good to split out this missing import fix for worker mode in a separate PR so we can merge it while we figure out the other problem.

synapse/handlers/message.py
tulir: Just batch sending any invite event in `state_events_at_start` seems to reproduce it. Sending the same invite event without workers doesn't reproduce it.

It looks like on the master process, `current_state_ids` will be an empty dict, but on the worker it will be `None`.

MadLittleMods: It looks like the [Synapse Complement test runner](https://github.com/matrix-org/synapse/blob/b09d90cac9179f84024d4cb3ab4574480c1fd1df/scripts-dev/complement.sh#L41-L54) does have the option to have the `WORKERS` environment variable set and go into worker mode. But I unable to get it working,

```
WORKERS=1 COMPLEMENT_ALWAYS_PRINT_SERVER_LOGS=1 COMPLEMENT_DIR=../complement ./scripts-dev/complement.sh TestImportHistoricalMessages
```

Results in my Docker crashing in different ways:

```
Fatal Error

Docker daemon failed to start

2021-11-10T00:39:12Z dockerd time="2021-11-10T00:39:11.027705833Z" level=info msg="[graphdriver] using prior storage driver: overlay2"
```

```
2021-11-10T03:24:22Z dockerd time="2021-11-10T03:24:22.696959125Z" level=info msg="Container 2faad2c3366d177fba0d1df67959518ac66114e72ddca539434f6c53cdadc084 failed to exit within 10 seconds of signal 15 - using the force
```

---

> It looks like on the master process, `current_state_ids` will be an empty dict, but on the worker it will be `None`.

Could we additionally fix up the difference in behavior?

tulir: I didn't figure out where the difference is coming from yet since the `current_state_ids` is loaded somehow weirdly

tulir: It looks like `_current_state_ids` is initialized to an [empty dict for outlier events](https://github.com/matrix-org/synapse/blob/v1.47.0rc2/synapse/events/snapshot.py#L134-L140), but that field isn't preserved when the `EventContext` is serialized/deserialized

tulir: Made a change to include outlier metadata in serialized event contexts and reverted the other one, seems to work still

synapse/events/snapshot.py
```diff
@@ -198,6 +201,11 @@ def deserialize(storage: "Storage", input: JsonDict) -> "EventContext":
             rejected=input["rejected"],
         )
```
MadLittleMods, line 202: > It looks like `_current_state_ids` is initialized to an [empty dict for outlier events](https://github.com/matrix-org/synapse/blob/v1.47.0rc2/synapse/events/snapshot.py#L134-L140), but that field is sent as a state group ID instead of the raw dict when the `EventContext` is serialized/deserialized
>
> *-- https://github.com/matrix-org/synapse/pull/11220/#discussion_r747057523*

As an alternative, would it be worth changing [`_AsyncEventContextImpl._fill_out_state`](https://github.com/matrix-org/synapse/blob/c99da2d0794de7240e791eba1a2f2d703fdecb60/synapse/events/snapshot.py#L320-L341) to default `current_state_ids={}` and `prev_state_ids={}` if there is no `state_group`?

Is there ever a case where `current_state_ids`/`prev_state_ids` should be `None` on purpose?

synapse/events/snapshot.py
```diff
@@ -198,6 +201,11 @@ def deserialize(storage: "Storage", input: JsonDict) -> "EventContext":
...
 
+        if input["outlier"]:
+            context._prev_state_ids = {}
+            context._current_state_ids = {}
+            context._outlier = True
```
MadLittleMods, line 207: Going down this route, it seems like we could short-circuit and return early for an outlier and just create `EventContext.for_outlier()`:

```py
if input["outlier"]:
  return EventContext.for_outlier()
```

tulir: Other metadata can still be added for outliers (like the appservice metadata below this)

MadLittleMods: I see 👍

synapse/events/snapshot.py
```diff
@@ -112,6 +112,7 @@ class EventContext:
 
     _current_state_ids = attr.ib(default=None, type=Optional[StateMap[str]])
     _prev_state_ids = attr.ib(default=None, type=Optional[StateMap[str]])
+    _outlier = attr.ib(default=False, type=bool)
```
richvdh, line 115: please add documentation for any new fields to the docstring, though I'm a bit surprised this is needed.

synapse/events/snapshot.py
```diff
@@ -198,6 +201,11 @@ def deserialize(storage: "Storage", input: JsonDict) -> "EventContext":
             rejected=input["rejected"],
         )
 
+        if input["outlier"]:
+            context._prev_state_ids = {}
+            context._current_state_ids = {}
```
richvdh, line 206: why are these necessary? shouldn't they be `{}` for an outlier anyway?

richvdh (changes requested):
Thanks for the contribution @tulir - sorry this seems to have fallen down the cracks while everyone waits for everyone else.

I'm afraid I'm struggling to review this, partly due to not being particular familiar with the way the MSC2716 implementation works, but also there seems to have been a bit of churn in the PR and it's unclear which parts of the PR thread are still relevant.

How/why does messing about with the `outlier` flag help fix #11219?

(or are there two bugs being fixed here? In which case I suggest doing as @MadLittleMods says and have separate PRs which explain more clearly what is going on).

synapse/events/snapshot.py
tulir: The problem is that `serialize` doesn't preserve those private fields, so they become `None` when transmitted to an event persistence worker. Being `None` means something will try to re-populate the values, which will fail because fetching those values for outlier events is not possible.

richvdh: Right, I see.

Let's do what @MadLittleMods suggests and have`_AsyncEventContextImpl._fill_out_state `to default current_state_ids={} and prev_state_ids={} if `state_group is None`

richvdh: closing this thread in favour of https://github.com/matrix-org/synapse/pull/11220/files#r747812503

synapse/events/snapshot.py
tulir: Now it does that and it seems to work still

Issue #10797, opened 2021-09-10
Prevent multiple instances of shutting down the same room simultaneously


anoadragon453:
We had a process attempt to shut down the same room using two, separate calls to the Admin API:

* `DELETE /_synapse/admin/v1/rooms/{roomId}`
* `POST /_synapse/admin/v1/{roomId}/delete`

Both of these method handlers eventually call [RoomShutdownHandler.shutdown_room](https://github.com/matrix-org/synapse/blob/0288e6033b08149867e742283bb0139f10b66b01/synapse/handlers/room.py#L1241), and there doesn't look to be any code that deduplicates requests anywhere.

This lead to a state where two long-running shutdown processes for the same room occurred and choked the main processes' CPU, resulting in the need to restart the main process to recover its performance.

dklimpel:
It is related to #10294

dklimpel:
Perhaps this API should return an ID like the purge room API. https://github.com/matrix-org/synapse/blob/develop/docs/admin_api/purge_history_api.md#purge-status-query

PR #6394, opened 2019-11-20
Add dev script to generate full SQL schema files

anoadragon453:
Fixes #6476 

So we ran into a [bug](https://github.com/matrix-org/synapse/issues/6311) that was a result of `INSERT` statements getting lost from converting SQL delta files to full schema. This was due to that process just copying over the schema of fully updated SQLite and Postgres tables, instead of the actual data.

Additionally, there wasn't any automated way (or much documentation) on how to generate full schemas, and I think if we had a canonical, automated way we might do it more often.

The script does this by generating a sqlite database with all background updates applied to it, then making use of the now-working `synapse_port_db` script to convert it to a postgres database. These two databases are then dumped to `full.{sqlite,postgres}.sql` files.

Also converted the full schema README from .txt to .md.

Issue #10792, opened 2021-09-10
Host dead after caches seem to break possibly due to multiple room delete commands in short period of time


jaywink:
### Description

From the logs it looks like a host admin deleted a room multiple times in a row within a few minutes. Initially this caused 403's for message sending for a bot user, which makes sense, but this soon turned to 500's until restarted, after which a 403 was again returned. Also some time after the room delete activity, a sync started to throw long stacktraces with 500 error. This is unconfirmed to have been fixed after restart, but it seems likely given the stack mentions caches.

From logs:

* Multiple events of the following events within a few minutes:
  * `synapse.handlers.room - 1339 - INFO - DELETE-6639- Shutting down room '!roomid:domain.tld'`
  * `synapse.handlers.room - 1348 - INFO - DELETE-6639- Kicking '@botuser:domain.tld' from '!roomid:domain.tld'...`
* `@botuser:domain.tld` starts to see 403
* Purge events seems to run multiple times for the rooms within a short period, with various events like this
  * `synapse.storage.databases.main.purge_events - 412 - INFO - DELETE-6639- [purge] removing !roomid:domain.tld from event_push_summary`
* Room send 403 turns into 500:

```
> 2021-09-09 18:02:40,129 - synapse.http.server - 93 - ERROR - PUT-6693- Failed handle request via 'RoomSendEventRestServlet': <XForwardedForRequest at 0x7f978dff48b0 method='PUT' uri='/_matrix/client/r0/rooms/!roomid:domain.tld/send/m.room.message/m1627686538.10576?access_token=<redacted>' clientproto='HTTP/1.1' site='8008'>
> Sep 9, 2021 @ 18:02:40.132synapse
Traceback (most recent call last):
Sep 9, 2021 @ 18:02:40.132synapse
  File "/usr/local/lib/python3.8/site-packages/twisted/internet/defer.py", line 1661, in _inlineCallbacks
Sep 9, 2021 @ 18:02:40.132synapse
    result = current_context.run(gen.send, result)
Sep 9, 2021 @ 18:02:40.132synapse
  File "/usr/local/lib/python3.8/site-packages/synapse/rest/client/room.py", line 242, in on_POST
Sep 9, 2021 @ 18:02:40.132synapse
    ) = await self.event_creation_handler.create_and_send_nonmember_event(
Sep 9, 2021 @ 18:02:40.132synapse
[...]

Issue #10294, opened 2021-07-01
Room deletion (shutdown) fail in a constant loop due to non-serializable access caused by PostgreSQL isolation levels


PeterCxy:
### Description

When using the [room deletion api](https://github.com/matrix-org/synapse/blob/develop/docs/admin_api/rooms.md#delete-room-api) to remove a large room (such as Matrix HQ) from the server, the purging process, if it needs more than a few seconds to finish, can sometimes enter a constant fail-retry loop due to `unable to serialize access` (because the tables are concurrently accessed and modified by other transactions constantly on a running server).

### Steps to reproduce

- On a moderately busy server (e.g. being in multiple moderately-sized federated rooms), try to purge a large federated room using the delete room api
- Observe that the process gets stuck with `unable to serialize access` being reported in the logs. You can also observe the behavior using PgHero, in which one exact long-running query will appear again and again on a regular interval, indicating Synapse has been retrying it again and again.

<!--
Describe how what happens differs from what you expected.

If you can identify any relevant log snippets from _homeserver.log_, please include
those (please be careful to remove any personal or private data). Please surround them with
``` (three backticks, on a line on their own), so that they are formatted legibly.
-->

### Version information

<!-- IMPORTANT: please answer the following questions, to help us narrow down the problem -->
<!--
 What version of Synapse is running?

You can find the Synapse version with this command:

$ curl http://localhost:8008/_synapse/admin/v1/server_version

(You may need to replace `localhost:8008` if Synapse is not configured to
listen on that port.)
-->
- **Version**: 1.37.1

- **Install method**: pip
<!-- examples: package manager/git clone/pip  -->

- **Platform**: Debian 10 "buster", in a LXC container
<!--
Tell us about the environment in which your homeserver is operating
distro, hardware, if it's running in a vm/container, etc.
-->

### Notes

The error will go away if the isolation level is changed to the lowest `READ COMMITTED` for the room-purging transaction, though I am not sure if this is correct or not, but I assume it should be fine given that we are just deleting everything related to a room.

```
diff --git a/synapse/storage/databases/main/purge_events.py b/synapse/storage/databases/main/purge_events.py
index 7fb7780d0..2619a6602 100644
--- a/synapse/storage/databases/main/purge_events.py
+++ b/synapse/storage/databases/main/purge_events.py
@@ -313,6 +313,7 @@ class PurgeEventsStore(StateGroupWorkerStore, CacheInvalidationWorkerStore):
         )

[...]

PR #11223, opened 2021-10-31
Convert delete room admin API to async endpoint

dklimpel:
Fixes: #10797, #6934

Related to:
- #7651
- #10294
- #10792

I have created two new endpoints:
- `DELETE /_synapse/admin/v2/rooms/<room_id>`
- `GET /_synapse/admin/v2/rooms/<room_id>/delete_status`

changes
- moved functionality of `RoomShutdownHandler` into `PaginationHandler`
- merged shutdown and purge room into one function/process
- did not touched old functions, it should be removed after the new api is stable for a few month
- added seperate unit tests
- added `result` information to `PurgeStatus`

Questions
- ~~Is there a chance to define the new endpoint not as final to improve it in near feature if there are issues?~~
- ~~I do not know why the [purge history endpoint](https://matrix-org.github.io/synapse/latest/admin_api/purge_history_api.html) creates a new `<purge_id>`. I have used the `room_id` instead. IMO this is easier for admins, because they do not have to memory a new id to get the status of process.~~
- ~~test `test_delete_same_room_twice` is flaky, do you have a hint to improve it? Otherwise I have to remove it.~~

### Pull Request Checklist


richvdh:
> Is there a chance to define the new endpoint not as final to improve it in near feature if there are issues?

I think the best way to do this is by flagging it as experimental and "subject to change" in the documentation.

richvdh:
> I do not know why the purge history endpoint creates a new <purge_id>. I have used the room_id instead. IMO this is easier for admins, because they do not have to memory a new id to get the status of process.

I think one reason for this is because there can be multiple requests for the same room. For example, the first request might not go back far enough, or might have failed. By giving each request its own id, you can make sure that you're getting the results for the right request.

I hear your point that it's harder for external tools to use, though. Happy for it to just use room id for now.

dklimpel:
Ok. That is a good point.
Maybe it makes sense to create an id for the uniqueness and to query the last status of the room with the room_id?

dklimpel:
Probably I have to adjust this PR after merging
- #11228

tests/rest/admin/test_room.py
```diff
@@ -687,6 +687,8 @@ def test_delete_same_room_twice(self):
         self.assertEqual(200, status_channel.code, msg=status_channel.json_body)
         self.assertEqual("complete", status_channel.json_body["status"])
 
+    test_delete_same_room_twice.skip = "Disabled by default because it's flaky"
```
richvdh, line 683: the reason this is flaky is that our database caching layer doesn't necessarily complete its callers in order (it's backed by a `set`) - so when we have multiple requests that each do a `get_user_by_access_token` query, there is no guarantee that one completes before the other.

I think there are two things you could do here:
 * arrange for the purge operation to take some time, then do `first_channel.await_result()` *before* making the second request. This would at least check that the second request cannot start once the first is in progress.
 * Or, allow *either* `first_channel` *or* `second_channel` (but not both) to complete with a 200.

richvdh: alternatively, wait for https://github.com/matrix-org/synapse/pull/11229 to land, which I think will fix this.

synapse/handlers/pagination.py
```diff
@@ -472,3 +495,312 @@ async def get_messages(
...
+                raise SynapseError(
+                    400, "User must be our own: %s" % (new_room_user_id,)
+                )
+
+        # we log the purge_id here so that it can be tied back to the
+        # request id in the log lines.
+        logger.info("[shutdown_and_purge] starting shutdown room_id %s", room_id)
+
+        self._purges_by_id[room_id] = PurgeStatus()
+        run_in_background(
```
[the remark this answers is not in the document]
dklimpel: Done

synapse/handlers/pagination.py
```diff
@@ -472,3 +495,312 @@ async def get_messages(
...
+            )
+            self._purges_by_id[room_id].status = PurgeStatus.STATUS_FAILED
+        finally:
+            self._purges_in_progress_by_room.discard(room_id)
+
+            # remove the purge from the list 24 hours after it completes
+            def clear_purge() -> None:
+                del self._purges_by_id[room_id]
+
+            self.hs.get_reactor().callLater(24 * 3600, clear_purge)
```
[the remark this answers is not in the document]
dklimpel: I changed it to reuse `_shutdown_and_purge_room `.

tests/rest/admin/test_room.py
```diff
@@ -418,17 +422,615 @@ def _assert_peek(self, room_id, expect_code):
...
         url = "events?timeout=0&room_id=" + room_id
         channel = self.make_request(
             "GET", url.encode("ascii"), access_token=self.admin_user_tok
         )
+        self.assertEqual(expect_code, channel.code, msg=channel.json_body)
+
+    def _test_result(
+        self,
+        channel: FakeChannel,
+        purge_id,
```
dklimpel, line 1000:
```suggestion
        purge_id: str,
```

synapse/handlers/pagination.py
```diff
@@ -472,3 +495,312 @@ async def get_messages(
...
+                migrated from the old room to the new.
+            new_room_id: A string representing the room ID of the new room.
+        """
+
+        if not new_room_name:
+            new_room_name = self.DEFAULT_ROOM_NAME
+        if not message:
+            message = self.DEFAULT_MESSAGE
+
+        self._purges_in_progress_by_room.add(room_id)
```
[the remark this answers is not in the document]
dklimpel: It is the first line, now.

docs/admin_api/rooms.md
```diff
@@ -470,16 +513,60 @@ The following JSON body parameters are available:
...
-* `failed_to_kick_users` - An array of users (`user_id`) that that were not kicked.
-* `local_aliases` - An array of strings representing the local aliases that were migrated from
-                    the old room to the new.
-* `new_room_id` - A string representing the room ID of the new room.
-
+* `status` - The status will be one of:
+  - `remove members` - The process is removing users from the `room_id`.
+  - `active` - The process is purging the room from databse.
+  - `complete` - The process has completed successfully.
+  - `failed` - The process is aborted, an error has occurred.
```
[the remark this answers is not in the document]
dklimpel: I am not sure what do you mean?
Should there some more information? Or do not return
```jsonc
"result": {
    "kicked_users": [],
    "failed_to_kick_users": [],
    "local_aliases": [],
    "new_room_id": null
}
```

synapse/handlers/pagination.py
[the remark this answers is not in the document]
dklimpel: I did an update and it uses the `purge_id`.
Additional I created a map from `room_id` to `purge_id`.
This makes it possible to query the purges of a room. I hope that is a better solution.

dklimpel:
Do you have any better idea for the name of the returned objet then `delete_status`?
```json
{
    "delete_status": [
        {
            "purge_id": "purgeid1",
            "status": "failed",
            "result": {
                "kicked_users": [],
                "failed_to_kick_users": [],
                "local_aliases": [],
                "new_room_id": null
            }
        }
    ]
}
```

docs/admin_api/rooms.md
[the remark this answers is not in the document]
dklimpel: Is this related to?
- #6934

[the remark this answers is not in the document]
richvdh: Yes, I meant "we should return more information", and it seems to be the same as #6934. We can leave it for now if you like.

docs/admin_api/rooms.md
```diff
@@ -475,16 +524,82 @@ The following JSON body parameters are available:
...
+
+**Note**: This API is new, experimental and "subject to change".
+
+It is possible to query the status of the background task for deleting rooms.
+The status can be queried up to 24 hours after completion of the task,
+or until Synapse is restarted (whichever happens first).
+
+With this API you can get the status of all tasks the last 24h for this `room_id`.
+If you want only get the status of one specific task, you can also use
+the [Purge status query API](purge_history_api.md#Purge-status-query).
```
richvdh, line 537: It feels confusing to tell users to use the *purge* status query for the results of a *delete* operation. I think if this is a thing we need, we should add a `GET /_synapse/admin/v2/rooms/<room_id>/delete_status/<purge_id>` for it.

docs/admin_api/rooms.md
```diff
@@ -475,16 +524,82 @@ The following JSON body parameters are available:
...
+
+A response body like the following is returned:
+
+```json
+{
+    "delete_status": [
+        {
+            "purge_id": "purgeid1",
+            "status": "failed",
+            "result": {
```
richvdh, line 553: given this is just the result of the shutdown operation, maybe:

```suggestion
            "shutdown_result": {
```

docs/admin_api/rooms.md
```diff
@@ -475,16 +524,82 @@ The following JSON body parameters are available:
...
-* `failed_to_kick_users` - An array of users (`user_id`) that that were not kicked.
-* `local_aliases` - An array of strings representing the local aliases that were migrated from
-                    the old room to the new.
-* `new_room_id` - A string representing the room ID of the new room.
-
+- `delete_status` - An array of objects, each containing information about one task.
+  Task objects contain the following fields:
+  - `status` - The status will be one of:
+    - `remove members` - The process is removing users from the `room_id`.
+    - `active` - The process is purging the room from database.
```
richvdh, line 592: `active` is confusing here. How about:

```suggestion
    - `shutting_down` - The process is removing users from the room.
    - `purging` - The process is purging the room and event data from database.
```

synapse/handlers/pagination.py
```diff
@@ -472,3 +502,212 @@ async def get_messages(
...
+                    new_room_user_id=new_room_user_id,
+                    new_room_name=new_room_name,
+                    message=message,
+                    block=block,
+                )
+                self._purges_by_id[purge_id].status = PurgeStatus.STATUS_ACTIVE
+
+                if purge:
+                    logger.info("starting purge room_id %s", room_id)
+                    self._purges_by_id[purge_id].status = PurgeStatus.STATUS_ACTIVE
```
richvdh, line 592: no-op, it's done 4 lines above

synapse/handlers/pagination.py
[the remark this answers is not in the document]
richvdh: > Should I add this logic to the "old" code?

sorry, which logic? which old code?

synapse/handlers/pagination.py
```diff
@@ -472,3 +502,212 @@ async def get_messages(
...
+                raise SynapseError(
+                    400, "User must be our own: %s" % (new_room_user_id,)
+                )
+
+        purge_id = random_string(16)
+
+        # we log the purge_id here so that it can be tied back to the
+        # request id in the log lines.
+        logger.info(
+            "[_shutdown_and_purge_room] starting shutdown room_id %s with purge_id %s",
```
richvdh, line 693: Not sure what the `[_shutdown_and_purge_room]` tag is for here? It doesn't seem to be used anywhere else?
```suggestion
            "starting shutdown room_id %s with purge_id %s",
```

synapse/handlers/pagination.py
```diff
@@ -472,3 +502,212 @@ async def get_messages(
...
+            room_id,
+            purge_id,
+        )
+
+        self._purges_by_id[purge_id] = PurgeStatus()
+        self._purges_by_room.setdefault(room_id, []).append(purge_id)
+        run_as_background_process(
+            "_shutdown_and_purge_room",
```
richvdh, line 701:
```suggestion
            "shutdown_and_purge_room",
```

richvdh (commented):
looks generally good to me. A few comments though.

synapse/handlers/pagination.py
```diff
@@ -49,19 +49,31 @@ class PurgeStatus:
...
+    result: Dict = {}
+
     def asdict(self) -> JsonDict:
         return {"status": PurgeStatus.STATUS_TEXT[self.status]}
 
+    def asdict_with_result(self) -> JsonDict:
+        return {
+            "status": PurgeStatus.STATUS_TEXT[self.status],
+            "result": self.result,
+        }
```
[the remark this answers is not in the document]
dklimpel: Yes, sure.
But then `/purge_histrory_status` gets always this information:
```json
{
  "kicked_users": [],
  "failed_to_kick_users": [],
  "local_aliases": [],
  "new_room_id": null
}
```
Purging does not kick someone. This informations are always empty, confusing and does not help the user.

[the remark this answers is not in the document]
richvdh: does it though? I thought it would get an empty dict.

Anyway, there's nothing stopping a caller using `asdict` on a delete request. It shouldn't just strip the data out.

docs/admin_api/rooms.md
dklimpel: If I change `active` to `purging` I change the `/purge_history_status` API.
If this is okay. I do this. Otherwise I have to create a new status.

synapse/handlers/pagination.py
dklimpel: The idea was to make it easier to read the log file. Entries from `_shutdown_and_purge_room` have the same log context. But this lines does have another context to get the request id.

synapse/handlers/pagination.py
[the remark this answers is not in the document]
dklimpel: https://github.com/matrix-org/synapse/blob/f36434590c1baafad4621afc9d2b583e9f89b6bb/synapse/handlers/pagination.py#L242-L271

There is
https://github.com/matrix-org/synapse/blob/f36434590c1baafad4621afc9d2b583e9f89b6bb/synapse/handlers/pagination.py#L268
instead
` run_as_background_process`

And the "normal" purges will not be added to the list of tasks, like this:
`self._purges_by_room.setdefault(room_id, []).append(purge_id)`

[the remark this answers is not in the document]
richvdh: oh right. Yes, those would probably be good to update.

docs/admin_api/rooms.md
dklimpel: What is with this: `GET /_synapse/admin/v2/rooms/delete_status/<purge_id>`?
Than you can request the status with `purge_id` and do not need the `room_id`.

docs/admin_api/rooms.md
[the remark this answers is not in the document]
dklimpel: I have added an error message of it fails.

synapse/handlers/pagination.py
```diff
@@ -472,3 +502,212 @@ async def get_messages(
...
+        Otherwise local users only leave the room without any information.
+        After that, the room will be removed from the database.
+
+        The new room will be created with the user specified by the
+        `new_room_user_id` parameter as room administrator and will contain a
+        message explaining what happened. Users invited to the new room will
+        have power level `-10` by default, and thus be unable to speak.
+
+        The local server will only have the power to move local user and room
+        aliases to the new room. Users on other servers will be unaffected.
```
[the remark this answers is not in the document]
dklimpel: Done

synapse/handlers/room.py
```diff
@@ -1276,6 +1276,12 @@ def get_current_key_for_room(self, room_id: str) -> Awaitable[str]:
 
 
 class RoomShutdownHandler:
+    """This handles synchronous room shutdowns and is part of the delete room v1 API.
+    It will become deprecated in the future.
+
+    The handler for asynchronous shudowns is part of the `PaginationHandler`.
+    If this handler is removed, `shutdown_room` must to be migrated to `PaginationHandler`
```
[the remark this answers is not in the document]
dklimpel: I have moved it.

docs/admin_api/rooms.md
dklimpel: Sytest fails.
https://github.com/matrix-org/sytest/blob/57c856c0c03f41ef9a6d06c89ca898e594505a4b/tests/48admin.pl#L28

Should I create an patch for sytest?
`   }, while => sub { $_[0]->get eq 'active' || $_[0]->get eq 'purging' });`

dklimpel: And we certainly need an upgrade notice.

richvdh: Sorry, I think this needs to be specific to delete requests.

I think this, and the problem with `asdict`, are really symptoms of re-using `PurgeStatus` for a different type of request.

docs/admin_api/rooms.md
richvdh: > What is with this: `GET /_synapse/admin/v2/rooms/delete_status/<purge_id>`?

sounds fine.

docs/admin_api/rooms.md
dklimpel: Ok. I will create a new class `DeleteStatus` to solve the problem.

richvdh: Thanks. Sorry this is turning into a bit of an epic!

docs/admin_api/rooms.md
```diff
@@ -400,10 +400,10 @@ as room administrator and will contain a message explaining what happened. Users
 to the new room will have power level `-10` by default, and thus be unable to speak.
 
 If `block` is `true`, users will be prevented from joining the old room.
-This option can also be used to pre-emptively block a room, even if it's unknown
-to this homeserver. In this case, the room will be blocked, and no further action
-will be taken. If `block` is `false`, attempting to delete an unknown room is
-invalid and will be rejected as a bad request.
+This option can in [Version 1](#version-1-old-version) also be used to pre-emptively
```
richvdh, line 403: wait, what? why are we removing this ability? It's important to be able to pre-emptively block rooms.

dklimpel: That ist not removed. The ability in Version 1 is created in #11228 and there is also an issue #11283.
IMO a simple block of a room with a bg task is not a very good usage. And it is implemented with raise and catching exceptions. I think v1 of this API should be removed after publishing the block API.

docs/admin_api/purge_history_api.md
```diff
@@ -69,6 +69,7 @@ This API returns a JSON body like the following:
 ```
 
 The status will be one of `active`, `complete`, or `failed`.
+If `status` is `failed` there will be a string `error` with the error message.
```
richvdh, line 72:
```suggestion

If `status` is `failed` there will be a string `error` with the error message.
```

docs/admin_api/rooms.md
```diff
@@ -480,17 +530,123 @@ The following JSON body parameters are available:
...
+
+### Query by `room_id`
 
-* `kicked_users` - An array of users (`user_id`) that were kicked.
-* `failed_to_kick_users` - An array of users (`user_id`) that that were not kicked.
-* `local_aliases` - An array of strings representing the local aliases that were migrated from
-                    the old room to the new.
-* `new_room_id` - A string representing the room ID of the new room, or `null` if
-                  no such room was created.
+With this API you can get the status of all tasks the last 24h for this `room_id`.
```
richvdh, line 543:
```suggestion
With this API you can get the status of all active deletion tasks, and all those completed in the last 24h,
for the given `room_id`.
```

richvdh:
fixes: #6394

synapse/handlers/pagination.py
```diff
@@ -78,11 +132,17 @@ def __init__(self, hs: "HomeServer"):
         self.state_store = self.storage.state
         self.clock = hs.get_clock()
         self._server_name = hs.hostname
+        self._room_shutdown_handler = hs.get_room_shutdown_handler()
 
         self.pagination_lock = ReadWriteLock()
         self._purges_in_progress_by_room: Set[str] = set()
```
richvdh, line 138:
```suggestion
        # IDs of rooms in which there currently an active purge *or delete* operation.
        self._purges_in_progress_by_room: Set[str] = set()
```

synapse/handlers/pagination.py
```diff
@@ -70,6 +121,9 @@ class PaginationHandler:
     paginating during a purge.
     """
 
+    # remove the purge from the list 24 hours after it completes
+    CLEAR_PURGE_TIME = 3600 * 24
```
richvdh, line 125: we tend to use milliseconds rather than seconds.

```suggestion
    # when to remove a completed deletion/purge from the results map
    CLEAR_PURGE_AFTER_MS = 1000 * 3600 * 24   # 24 hours
```

synapse/handlers/pagination.py
```diff
@@ -295,14 +360,17 @@ async def _purge_history(
...
         finally:
             self._purges_in_progress_by_room.discard(room_id)
 
             # remove the purge from the list 24 hours after it completes
             def clear_purge() -> None:
                 del self._purges_by_id[purge_id]
 
-            self.hs.get_reactor().callLater(24 * 3600, clear_purge)
+            self.hs.get_reactor().callLater(
+                PaginationHandler.CLEAR_PURGE_TIME, clear_purge
```
richvdh, line 372:
```suggestion
                PaginationHandler.CLEAR_PURGE_AFTER_MS/1000, clear_purge
```

synapse/handlers/room.py
```diff
@@ -1279,14 +1279,23 @@ def get_current_key_for_room(self, room_id: str) -> Awaitable[str]:
...
 
 class ShutdownRoomResponse(TypedDict):
+    """
+    kicked_users: An array of users (`user_id`) that were kicked.
+    failed_to_kick_users:
+        An array of users (`user_id`) that that were not kicked.
+    local_aliases:
+        An array of strings representing the local aliases that were
+        migrated from the old room to the new.
+    new_room_id: A string representing the room ID of the new room.
```
richvdh, line 1289:
```suggestion
    Attributes:
        kicked_users: An array of users (`user_id`) that were kicked.
        failed_to_kick_users:
            An array of users (`user_id`) that that were not kicked.
        local_aliases:
            An array of strings representing the local aliases that were
            migrated from the old room to the new.
        new_room_id: A string representing the room ID of the new room.
```

synapse/handlers/room.py
```diff
@@ -1312,6 +1320,10 @@ async def shutdown_room(
         block: bool = False,
     ) -> ShutdownRoomResponse:
         """
+        This handles synchronous room shutdowns and is part of the delete room v1 API.
+        The handler for asynchronous shudowns is part of the `PaginationHandler`.
+        If this handler is removed, `shutdown_room` must to be migrated to `PaginationHandler`
+
```
richvdh, line 1326: this is not correct. It is used by both v1 and v2 APIs.

```suggestion
```

synapse/handlers/pagination.py
```diff
@@ -472,3 +557,193 @@ async def get_messages(
...
+                If set to `true`, the room will be purged from database
+                also if it fails to remove some users from room.
+
+        Saves a `RoomShutdownHandler.ShutdownRoomResponse` in `DeleteStatus`:
+        """
+
+        self._purges_in_progress_by_room.add(room_id)
+        try:
+            with await self.pagination_lock.write(room_id):
+
```
richvdh, line 613:
```suggestion
```

docs/admin_api/rooms.md
richvdh: hrm. Ok, let's leave it for now.

synapse/handlers/pagination.py
```diff
@@ -472,3 +557,193 @@ async def get_messages(
...
+            # remove the delete from the list 24 hours after it completes
+            def clear_delete() -> None:
+                del self._delete_by_id[delete_id]
+                self._delete_by_room[room_id].remove(delete_id)
+                if not self._delete_by_room[room_id]:
+                    del self._delete_by_room[room_id]
+
+            self.hs.get_reactor().callLater(
+                PaginationHandler.CLEAR_PURGE_TIME, clear_delete
```
richvdh, line 663:
```suggestion
                PaginationHandler.CLEAR_PURGE_AFTER_MS/1000, clear_delete
```

richvdh (approved):
:tada: thanks @dklimpel , let's get this thing shipped!

dklimpel:
> 🎉 thanks @dklimpel , let's get this thing shipped!

Thank you for your patience. The PR is unusually large and complex.

Issue #5823, opened 2019-08-05
well_known m.homeserver path has errant trailing "/" which breaks discovery in clients such as Pattle


khimaros:
### Description

Matrix clients such as [Pattle](https://git.pattle.im/pattle/android) honor the output of `/.well-known/matrix/client` literally.

A block in `synapse/config/server` [modifies the value of public_baseurl](https://github.com/matrix-org/synapse/blob/c0a1301ccd91cdfbb079f675b3c3dc305f876be7/synapse/config/server.py#L192) by appending a `/` if one is not already present. The value of `public_baseurl` is returned unmodified by `synapse/rest/well_known` which causes the errant trailing `/` to reach the user. By way of example:

```
{"m.identity_server": {"base_url": "https://myhome.server"}, "m.homeserver": {"base_url": "https://myhome.server/"}}
```

### Steps to reproduce

- configure a homeserver
- set `public_baseurl` to `https://myhome.server` or equiv.
- use Pattle to log into that homeserver (configure server in Advanced)

Pattle simply appends `/_matrix` to the URL returned in `m.homeserver` which results in requests to the Matrix server of the form:

```
2019-08-05 17:32:34,571 - synapse.http.server - 79 - INFO - GET-77- <XForwardedForRequest at 0x7f70fdbce240 method='GET' uri='//_matrix/client/r0/register/available?username=myuser' clientproto='HTTP/1.0' site=8008> SynapseError: 400 - Unrecognized request
```

synapse refuses to handle these requests due to the double `//`

Removing all three lines in [this block](https://github.com/matrix-org/synapse/blob/c0a1301ccd91cdfbb079f675b3c3dc305f876be7/synapse/config/server.py#L192) appears to have no ill effect on Riot Android or Riot Web and fixes the discovery issue for Pattle.

### Version information

All versions, including HEAD.

turt2live:
It's not specified to include or not include the trailing slash. The reason for riot-web (and probably the other riots) not exploding is https://github.com/matrix-org/matrix-js-sdk/blob/35f1cdf89c26a0a6567e8983a03a178d0b198b49/src/autodiscovery.js#L462-L464

I'd personally recommend that Pattle do the same, as Synapse is not the only producer of well-known files.

khimaros:
@wilkomanger who is the author of Pattle.

khimaros:
Making the clients more robust seems reasonable to me.

[...]

PR #11224, opened 2021-10-31
comment out code that removes trailing slash as per 5823.

src-r-r:
Signed-off-by: Jordan Hewitt <633901+src-r-r@users.noreply.github.com>

### Pull Request Checklist


callahad:
We actually title newsfragments after the PR, not the issue the fix, so a `git mv changelog.d/{5823,11224}.bugfix` should get the CI back underway :)

src-r-r:
Great, thank you!

dklimpel:
I am not sure if modify the configuration is the right direction.
IMO it should be some logic in `WellKnownBuilder` to be sure that there is no trailing slash.
https://github.com/matrix-org/synapse/blob/e320f5dba32f5b5818b6d5a0059ae388430b9a72/synapse/rest/well_known.py#L31-L47

src-r-r:
> I am not sure if modify the configuration is the right direction. IMO it should be some logic in `WellKnownBuilder` to be sure that there is no trailing slash.
> 
> https://github.com/matrix-org/synapse/blob/e320f5dba32f5b5818b6d5a0059ae388430b9a72/synapse/rest/well_known.py#L31-L47

Gotcha. I'll modify it then. I'm not doubting you; I'm just curious--what's the reasoning?

dklimpel:
The reason is that `public_baseurl` is used many times in code. With your change the tests fail. You have to verify every position in code where `public_baseurl` is used if this change has an impact.
Perhaps there are places where the trailing slash is necessary.

synapse/config/server.py
```diff
@@ -264,9 +264,10 @@ def read_config(self, config, **kwargs):
         self.use_frozen_dicts = config.get("use_frozen_dicts", False)
 
         self.public_baseurl = config.get("public_baseurl")
-        if self.public_baseurl is not None:
-            if self.public_baseurl[-1] != "/":
-                self.public_baseurl += "/"
+        # Commented out due to https://github.com/matrix-org/synapse/issues/8308
+        # if self.public_baseurl is not None:
+        #     if self.public_baseurl[-1] != "/":
+        #         self.public_baseurl += "/"
```
richvdh, line 270: Could you explain how this is related to https://github.com/matrix-org/synapse/issues/8308?

richvdh:
related to #5823

richvdh:
> I am not sure if modify the configuration is the right direction.
> IMO it should be some logic in WellKnownBuilder to be sure that there is no trailing slash.

I'm not necessarily sure I agree with this: it seems a bit silly to explicitly add a trailing slash in `read_config`, and then strip it out again in `WellKnownBuilder`.

But for sure, if we change `read_config` in this way, we'll have to update the many places in the codebase that expect `public_baseurl` to have a trailing slash, which sounds like a lot of work.

ShadowJonathan:
I think the commit history of this PR is a bit borked, i'd recommend looking at how to rebase your commits until the current main branch.

src-r-r:
> I think the commit history of this PR is a bit borked, i'd recommend looking at how to rebase your commits until the current main branch.

I did do a rebase with develop. Should I do a rebase with master?

ShadowJonathan:
No, hold on, something else weird is happening here then

DMRobertson:
This passed most checks, but the sytest integration tests aren't happy. The failures aren't particularly clear, regrettably

From a quick look, I don't think the `needs_all` call here succeeds.

https://github.com/matrix-org/sytest/blob/3125282b7ac6de20bc8b2bcf07ca663f33250131/tests/10apidoc/13ui-auth.pl#L140-L150

I can't see what the failure is though. Maybe we need more `log_if_fail`? Perhaps @richvdh could advise?

DMRobertson (changes requested):
Thanks for bearing with us on this one.

There's quite a few things in these changes that I don't follow, and in particular some config handling code that seems to have been deleted. There might be some confusion between the server and client well-known endpoints too?

Additionally, I think it might be clearer in places to use `str.rstrip` to remove trailing slashes, rather than . (I sympathise with your comment on #5823 about urljoin --- I'd be fine with that too, but painful to implement it across the whole source tree.)

Taking a step back, the plan as I as I understand it is:

- when ingesting config, remove any trailing slashes from the end of the URL.
- keep the trimmed publicbase_url around in memory
- when serving the .well-known/matrix/server file, use the trimmed version
- fix up any remaining places which implicitly were expecting the baseurl config to end in a forward slash
   -  by adding an explicit '/' to the path we build

callahad:
I think we're going around in circles a bit because #5823 superficially _appeared_ simple, but in actually attempting to resolve it, @src-r-r has unearthed a bunch of conflicting / contradictory / undefined behavior.

I suspect it would be a worthwhile effort for someone on the Synapse team to write a new issue to replace #5823, which more completely enumerates the problem, solution, and consequences thereof.

Regardless, it seems clear to me that clients should (must?) be robust against base_url's both with _and_ without trailing slashes, doubly so since matrix.org is inconsistent with Synapse's default behavior.

callahad:
@src-r-r Per discussion on the Matrix Spec https://github.com/matrix-org/matrix-doc/issues/3465#issuecomment-980449599 we've ultimately decided that mandating _either_ that trailing slashes are stripped _or_ that they're added is fraught with peril, and that the only practically viable path forward is to explicitly instruct clients to handle both cases. This has now become part of the Matrix spec, rendering this pull request moot.

I'm terribly sorry that the outcome of your hard work is not __✅ Merged!__, but in many ways this is a much more valuable result: your wrangling with this issue catalyzed improvements to the Matrix Spec in a way that will make the ecosystem itself much more robust. We wouldn't have gotten there without your involvement in this pull request. Thank you.

src-r-r:
Thank you @callahad .

PR #11226, opened 2021-11-01
Fix a bug in unit test `test_block_room_and_not_purge`

dklimpel:
### Pull Request Checklist


changelog.d/11226.bugfix
```diff
@@ -0,0 +1 @@
+Fix a bug in unit test `test_block_room_and_not_purge`.
```
DMRobertson, line 1: I think this should be `11226.misc` because it's not a user-visible bug (rather, a mistake and hole in our test coverage).

Otherwise I'm happy with this. I'll open a separate PR to make the api return 400 if you ask for both block and purge to be False.

DMRobertson (changes requested):
Thanks for this!

DMRobertson (approved):

Issue #2392, opened 2017-07-28
Blacklist room ID from federating


Mikaela:
I think that Synapse should somehow have the ability to block specific room IDs, as somehow specified by homeserver administrators, from federating (preventing users from joining it or adding aliases or even preventing the room from having aliases).

Use case would be content illegal in the country where the homeserver is ran and potential legal trouble for homeserver administrator for hosting the room or content in it.

*I think this issue has been discussed previously at Matrix, but I wasn't able to find an issue either from this repository or matrix-docs and when asking I was told to feel free to file it here.*

turt2live:
Would the shutdown room API be suitable for this? https://github.com/matrix-org/synapse/pull/2291

allquixotic:
+1, Would like to see this added.

richvdh:
It's not really what server ACLs do.

/admin/shutdown_room is probably the closest bet currently, though it's a bit of a sledgehammer against a nut.

MurzNN:
For implement blacklisting of room at now we can use https://github.com/matrix-org/synapse/blob/develop/docs/admin_api/rooms.md#delete-room-api for remove current room, and https://github.com/matrix-org/mjolnir tool for prevent new invites from this room to all local users.

Issue #5575, opened 2019-06-27
Room Shutdown API has no validation/error handling


f0x52:
### Description
after calling `_matrix/client/r0/admin/shutdown_room` on a room, I get invited to the replacement room, but I still continue to receive new events in the old room.

### Steps to reproduce

- join a room
- run the `_matrix/client/r0/admin/shutdown_room` endpoint on it
- continue to receive new events
<!--
Describe how what happens differs from what you expected.

If you can identify any relevant log snippets from _homeserver.log_, please include
those (please be careful to remove any personal or private data). Please surround them with
``` (three backticks, on a line on their own), so that they are formatted legibly.
-->
I don't see any errors in the log, and from my understanding this endpoint should totally remove this room from my server, for the future as well.

### Version information

<!-- IMPORTANT: please answer the following questions, to help us narrow down the problem -->

<!-- Was this issue identified on matrix.org or another homeserver? -->
- **Homeserver**: pixie.town

If not matrix.org:

<!--
What version of Synapse is running?
You can find the Synapse version by inspecting the server headers (replace matrix.org with
your own homeserver domain):
$ curl -v https://matrix.org/_matrix/client/versions 2>&1 | grep "Server:"
-->
- **Version**: Synapse 1.0.0

- **Install method**: pip in virtualenv
<!-- examples: package manager/git clone/pip  -->

- **Platform**: Debian Unstable on bare metal
<!--
Tell us about the environment in which your homeserver is operating
distro, hardware, if it's running in a vm/container, etc.
-->

**Comment by richvdh** [comment]
shutdown_room is undocumented, but afaict it only removes local users from the room. It doesn't remove the room in its entirety from the server.

That said, you should be removed from the room, and the server should reject new events for the room (because all local users have left). Can you share logs that show the shutdown_room running and the new messages arriving?

**Comment by f0x52** [comment]
As this issue is appearing again, with a new version of the same alt-right room sending spam invites and harassment, here are my steps:
- receive invite
- join room
- call shutdown_room endpoint with curl:
```bash
curl 'https://pixie.town/_matrix/client/r0/admin/shutdown_room/\!HkIVpVBRHoIryLnGUn:chat.thisisjoes.site' \
-H "Authorization: Bearer $ADMIN_TOKEN" -XPOST \
-d'{"new_room_user_id": "@f0x:pixie.town", "message": "testing shutdown_room", "room_name": "Replacement Room after supposed shutdown"}'
```
[...]

PR #11228, opened 2021-11-01
Allow admins to proactively block rooms

DMRobertson:
Wasn't thrilled with how this one turned out. The implementation and API shape feels icky.

The first three commits are doc fixups. The rest should be okay to review commit-by-commit.

### DIFF
--- a/changelog.d/11228.feature
+++ b/changelog.d/11228.feature
@@ -1,1 +1,1 @@
-Allow the admin [Delete Room API](https://matrix-org.github.io/synapse/latest/admin_api/rooms.html#delete-room-api) to block a room before this server joins it.
+Allow the admin [Delete Room API](https://matrix-org.github.io/synapse/latest/admin_api/rooms.html#delete-room-api) to block a room without the need to join it.
--- a/changelog.d/11228.feature
+++ b/changelog.d/11228.feature
@@ -1,0 +1,1 @@
+Allow the admin [Delete Room API](https://matrix-org.github.io/synapse/latest/admin_api/rooms.html#delete-room-api) to block a room before this server joins it.
--- a/docs/admin_api/rooms.md
+++ b/docs/admin_api/rooms.md
@@ -391,7 +391,10 @@ The new room will be created with the user specified by the `new_room_user_id` p
 as room administrator and will contain a message explaining what happened. Users invited
 to the new room will have power level `-10` by default, and thus be unable to speak.
 
-If `block` is `True` it prevents new joins to the old room.
+If `block` is `true`, users will be prevented from joining the old room.
+This option can also be used to pre-emptively block a room, even if it's unknown
+to this homeserver. (If `block` is `false`, attempting to delete an unknown room
+is invalid and will be rejected as a bad request.)
 
 This API will remove all trace of the old room from your database after removing
 all local users. If `purge` is `true` (the default), all traces of the old room will
@@ -459,8 +462,9 @@ The following JSON body parameters are available:
               `new_room_user_id` in the new room. Ideally this will clearly convey why the
                original room was shut down. Defaults to `Sharing illegal content on this server
                is not permitted and rooms in violation will be blocked.`
-* `block` - Optional. If set to `true`, this room will be added to a blocking list, preventing
-            future attempts to join the room. Defaults to `false`.
+* `block` - Optional. If set to `true`, this room will be added to a blocking list,
+            preventing future attempts to join the room. Rooms can be blocked
+            even if they're not yet known to the homeserver. Defaults to `false`.
 * `purge` - Optional. If set to `true`, it will remove all traces of the room from your database.
             Defaults to `true`.
 * `force_purge` - Optional, and ignored unless `purge` is `true`. If set to `true`, it
--- a/docs/admin_api/rooms.md
+++ b/docs/admin_api/rooms.md
@@ -478,7 +478,8 @@ The following fields are returned in the JSON response body:
 * `failed_to_kick_users` - An array of users (`user_id`) that that were not kicked.
 * `local_aliases` - An array of strings representing the local aliases that were migrated from
                     the old room to the new.
-* `new_room_id` - A string representing the room ID of the new room.
+* `new_room_id` - A string representing the room ID of the new room, or `null` if
+                  no such room was created.
 
 
 ## Undoing room deletions
--- a/docs/admin_api/rooms.md
+++ b/docs/admin_api/rooms.md
@@ -393,8 +393,9 @@ to the new room will have power level `-10` by default, and thus be unable to sp
 
 If `block` is `true`, users will be prevented from joining the old room.
 This option can also be used to pre-emptively block a room, even if it's unknown
-to this homeserver. (If `block` is `false`, attempting to delete an unknown room
-is invalid and will be rejected as a bad request.)
+to this homeserver. In this case, the room will be blocked, and no further action
+will be taken. If `block` is `false`, attempting to delete an unknown room is
+invalid and will be rejected as a bad request.
 
 This API will remove all trace of the old room from your database after removing
 all local users. If `purge` is `true` (the default), all traces of the old room will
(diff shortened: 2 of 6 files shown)


tests/rest/admin/test_room.py
```diff
@@ -281,6 +284,31 @@ def test_block_room_and_not_purge(self):
...
+        should be ignored.
+        """
+        room_id = "!unknown:test"
+
+        # The room isn't already in the blocked rooms table
+        self._is_blocked(room_id, expect=False)
+
+        # Request the room be blocked.
+        channel = self.make_request(
+            self.method,
```
dklimpel, line 300:
```suggestion
            "DELETE",
```

DMRobertson: Argh, thank you. I think this is an unfortunate merge conflict with your recent changes.

synapse/rest/admin/rooms.py
```diff
@@ -241,7 +241,9 @@ async def _delete_room(
         if purge:
             await pagination_handler.purge_room(room_id, force=force_purge)
 
-        return 200, ret
+        # Cast safety: I'm casting away the knowledge that this is a TypedDict.
```
anoadragon453, line 253:
```suggestion
        # Cast safety: Cast away the knowledge that this is a TypedDict.
```

synapse/handlers/room.py
```diff
@@ -1365,14 +1365,26 @@ async def shutdown_room(
...
-        if not await self.store.get_room(room_id):
-            raise NotFoundError("Unknown room id %s" % (room_id,))
-
-        # This will work even if the room is already blocked, but that is
-        # desirable in case the first attempt at blocking the room failed below.
+        # Action the block first (even if the room doesn't exist yet)
         if block:
+            # This will work even if the room is already blocked, but that is
+            # desirable in case the first attempt at blocking the room failed below.
             await self.store.block_room(room_id, requester_user_id)
```
anoadragon453, line 1372: I think it may be worth adding to the docstring of `block_room` that it's intentional that the function can block rooms that aren't yet known to the homeserver. Lest someone try to add a 'helpful check' in the future.

synapse/handlers/room.py
```diff
@@ -1365,14 +1365,26 @@ async def shutdown_room(
...
+                return {
+                    "kicked_users": [],
+                    "failed_to_kick_users": [],
+                    "local_aliases": [],
+                    "new_room_id": None,
+                }
+            else:
+                # But if you don't want to preventatively block another room,
+                # this function can't do anything useful.
+                raise NotFoundError("Unknown room id %s" % (room_id,))
```
anoadragon453, line 1386: It may be worth specifically called out that it was _not the block_ that failed here? Otherwise server admins might assume they cannot block unknown rooms as well?

```suggestion
                raise NotFoundError("Cannot shut down room: unknown room id %s" % (room_id,))
```

synapse/handlers/room.py
```diff
@@ -1365,14 +1365,26 @@ async def shutdown_room(
         if not RoomID.is_valid(room_id):
             raise SynapseError(400, "%s is not a legal room ID" % (room_id,))
 
-        if not await self.store.get_room(room_id):
-            raise NotFoundError("Unknown room id %s" % (room_id,))
-
-        # This will work even if the room is already blocked, but that is
-        # desirable in case the first attempt at blocking the room failed below.
+        # Action the block first (even if the room doesn't exist yet)
```
anoadragon453, line 1368: I almost feel as though we want to pull the room blocking bit out of this function and put the logic into `_delete_room`. Then only call `shutdown_room` and `purge_room` if the room is known. That may make the control flow slightly simpler?

changelog.d/11228.feature
```diff
@@ -0,0 +1 @@
+Allow the admin [Delete Room API](https://matrix-org.github.io/synapse/latest/admin_api/rooms.html#delete-room-api) to block a room before this server joins it.
```
anoadragon453, line 1:
```suggestion
Allow the admin [Delete Room API](https://matrix-org.github.io/synapse/latest/admin_api/rooms.html#delete-room-api) to block a room without the need to join it.
```

docs/admin_api/rooms.md
```diff
@@ -478,7 +478,8 @@ The following fields are returned in the JSON response body:
 * `failed_to_kick_users` - An array of users (`user_id`) that that were not kicked.
 * `local_aliases` - An array of strings representing the local aliases that were migrated from
                     the old room to the new.
```
anoadragon453, line 480: Could we note somewhere in this doc that a room can be blocked without the need to join it? Such as:

> * `block` - Optional. If set to `true`, this room will be added to a blocking list, preventing
future attempts to join the room. The server does not need to be in the room in order
to block it. Defaults to `false`.

I realise it's already intuitive that the server would not need to be in the room to block it, but the previous restrictions may have become learned already... and it's best to be explicit?

synapse/rest/admin/rooms.py
DMRobertson: Thanks. I'll also add a link to remind the reader why the cast is necessary in the first place.

docs/admin_api/rooms.md
DMRobertson: This bit matches the function docstring. I'll keep those two roughly in sync.

synapse/handlers/room.py
DMRobertson: I was a little reluctant here.

I didn't like the idea that the block could succeed only for the shutdown/purge to fail; I'd rather the whole thing succeed or fail atomically. But that's the situation today and wouldn't be introduced by your change. It would maybe mean checking twice that a room exists, which is a bit sad.

I'm kinda of the view that this is best left alone for now---I feel like blocking might merit being its own separate API call?

dklimpel: I am not sure if a simple block needs a new API. But I suspect it.
I am try to bring the delete API to a background task. #11223
I do not know how to handle a simple block in V2 API. A block request bring to background is not a good solution. If I am a user, I would except to get a response immediately.
An alternative for V2 API is to response for a simple block other than a shutdown or purge. But I do not like an API with two different return types.

dklimpel: It is related to #5575.
Maybe should the API return if the blocked room is known or unknown for the Homeserver.

dklimpel:
It also fixes #2392 ?

synapse/handlers/room.py
anoadragon453: > I'm kinda of the view that this is best left alone for now---I feel like blocking might merit being its own separate API call. How does that sit with you?

I think shutting down a room should still block it, but separating a single block out into its own API call does sound like it'd be cleaner and better long-term. Leaving this alone for now with the hope of refactoring it all later sounds fine. Can you write up a ticket for the idea?

@dklimpel placing the shutdown room handling into a background process does sound like something that should happen regardless. I'm not sure how it relates to moving blocking of a room into a separate API however. Room blocking should be a relatively quick process.

docs/admin_api/rooms.md
```diff
@@ -391,13 +391,16 @@ The new room will be created with the user specified by the `new_room_user_id` p
 as room administrator and will contain a message explaining what happened. Users invited
 to the new room will have power level `-10` by default, and thus be unable to speak.
 
-If `block` is `True` it prevents new joins to the old room.
+If `block` is `true`, users will be prevented from joining the old room.
+This option can also be used to pre-emptively block a room, even if it's unknown
+to this homeserver. (If `block` is `false`, attempting to delete an unknown room
+is invalid and will be rejected as a bad request.)
```
anoadragon453, line 397:
```suggestion
If `block` is `true`, users will be prevented from joining the old room.
This option can also be used to pre-emptively block a room, even if it's unknown
to this homeserver. In this case, the room will be blocked, and no further action
will be taken. If `block` is `false`, attempting to delete an unknown room is
invalid and will be rejected as a bad request.
```

Final bit of clarification after reading through it.

anoadragon453:
> It also fixes #2392?

I think so. Have linked it to this PR, which will close it when this merges.

anoadragon453 (approved):
lgtm otherwise. I left https://github.com/matrix-org/synapse/pull/11228/files#r741843724 open for discussion, even though I don't expect a change off the back of it.

docs/admin_api/rooms.md
DMRobertson: Thanks. I'll update the corresponding docstring too.

PR #8572, opened 2020-10-16
Push some deferred wrangling down into DeferredCache

richvdh:
Rather than `DeferredCache.get()` sometimes returning an `ObservableDeferred` that you have to `.observe`, and sometimes returning an immediate result that you have to wrap with a `Deferred`, we can make it behave consistently. This removes complexity from the call site, improves type safety, and generally leasds to a cleaner API.

~~Based on #8568~~

PR #190, opened 2015-06-18
Fix notifier leak

erikjohnston:

PR #11229, opened 2021-11-01
ObservableDeferred: run observers in order

richvdh:
It came as a bit of a surprise to me that `ObservableDeferred` *doesn't* run its registered observers in order. This has ramifications like meaning that, if we have two things waiting for a database lookup, we don't run them in a first come, first served order.

I'm not really sure that the existing behaviour breaks anything, but I think a bit more determinism is probably a good thing, as is doing what people expect.

### DIFF
--- a/synapse/util/async_helpers.py
+++ b/synapse/util/async_helpers.py
@@ -76,7 +76,7 @@ class ObservableDeferred(Generic[_T]):
     def __init__(self, deferred: "defer.Deferred[_T]", consumeErrors: bool = False):
         object.__setattr__(self, "_deferred", deferred)
         object.__setattr__(self, "_result", None)
-        object.__setattr__(self, "_observers", list())
+        object.__setattr__(self, "_observers", [])
 
         def callback(r):
             object.__setattr__(self, "_result", (True, r))
@@ -84,7 +84,7 @@ class ObservableDeferred(Generic[_T]):
             # once we have set _result, no more entries will be added to _observers,
             # so it's safe to replace it with the empty tuple.
             observers = self._observers
-            object.__setattr__(self, "_observers", tuple())
+            object.__setattr__(self, "_observers", ())
 
             for observer in observers:
                 try:
@@ -104,7 +104,7 @@ class ObservableDeferred(Generic[_T]):
             # once we have set _result, no more entries will be added to _observers,
             # so it's safe to replace it with the empty tuple.
             observers = self._observers
-            object.__setattr__(self, "_observers", tuple())
+            object.__setattr__(self, "_observers", ())
 
             for observer in observers:
                 # This is a little bit of magic to correctly propagate stack
--- a/synapse/util/async_helpers.py
+++ b/synapse/util/async_helpers.py
@@ -17,6 +17,7 @@ import collections
 import inspect
 import itertools
 import logging
+from collections import Collection
 from contextlib import contextmanager
 from typing import (
     Any,
@@ -26,7 +27,6 @@ from typing import (
     Generic,
     Hashable,
     Iterable,
-    List,
     Optional,
     Set,
     TypeVar,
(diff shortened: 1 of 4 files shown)


tests/util/caches/test_deferred_cache.py
```diff
@@ -47,9 +47,7 @@ def check1(r):
             self.assertTrue(set_d.called)
             return r
 
-        # TODO: Actually ObservableDeferred *doesn't* run its tests in order on py3.8.
-        #   maybe we should fix that?
-        # get_d.addCallback(check1)
```
richvdh, line 52: the comment isn't terribly clear, but I think this is what we're fixing.

richvdh: cf: https://github.com/matrix-org/synapse/pull/8572#discussion_r508427387

synapse/util/async_helpers.py
```diff
@@ -127,20 +136,13 @@ def observe(self) -> "defer.Deferred[_T]":
         """
         if not self._result:
             d: "defer.Deferred[_T]" = defer.Deferred()
-
-            def remove(r):
-                self._observers.discard(d)
-                return r
-
-            d.addBoth(remove)
```
richvdh, line 135: this code removed the observer from `self._observers` when it was run. I think it was redundant before, because the observer was `pop()`ed from `self._observers` anyway, but it's doubly-redundant now, since the whole of `self._observers` is thrown away.

(it was added in https://github.com/matrix-org/synapse/pull/190 - no real clues there as to why.)

synapse/util/async_helpers.py
```diff
@@ -127,20 +136,13 @@ def observe(self) -> "defer.Deferred[_T]":
...
-
-            self._observers.add(d)
+            self._observers.append(d)
             return d
         else:
             success, res = self._result
             return defer.succeed(res) if success else defer.fail(res)
 
-    def observers(self) -> "List[defer.Deferred[_T]]":
+    def observers(self) -> "Collection[defer.Deferred[_T]]":
```
DMRobertson, line 145: A bit surprised that we're leaking internal state here!

Could only see [one use](https://github.com/matrix-org/synapse/blob/c7a5e49664ab0bd18a57336e282fa6c3b9a17ca0/synapse/notifier.py#L162-L163) though. Probably fine as it is.

DMRobertson (approved):
I wasn't quite sure what `ObservableDeferred` gives us over plain `Deferred`. I guess the key is this bit from the docstring?

>  These observer deferreds do not affect the callback chain of the original deferred.

That aside...

SGTM! Given that the observer order was unspecified before, hopefully replacing this with a very specific order won't cause any breakages. ([XKCD 1172](https://xkcd.com/1172/) notwithstanding)

synapse/util/async_helpers.py
```diff
@@ -76,12 +76,17 @@ class ObservableDeferred(Generic[_T]):
     def __init__(self, deferred: "defer.Deferred[_T]", consumeErrors: bool = False):
         object.__setattr__(self, "_deferred", deferred)
         object.__setattr__(self, "_result", None)
-        object.__setattr__(self, "_observers", set())
+        object.__setattr__(self, "_observers", list())
```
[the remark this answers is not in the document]
richvdh: yes, for some reason someone decided that making ObservableDeferred a transparent wrapper for a Deferred was a good idea. I hate it.

richvdh:
>> I wasn't quite sure what ObservableDeferred gives us over plain Deferred. I guess the key is this bit from the docstring?

>    These observer deferreds do not affect the callback chain of the original deferred.

yes, exactly. Normally when you add a series of callbacks to a deferred, each callback is passed the result of the *previous* callback. `ObservableDeferred` ensures that each observer sees the same thing.

synapse/util/async_helpers.py
richvdh: yup. Thought about changing it to a method that just returns the number of observers. Ran out of enthusiasm.

PR #11230, opened 2021-11-01
Change display names/avatar URLs to None if they contain null bytes before storing in DB

### LINKED ISSUE #10886: Internal error due to NUL byte in Conduit federation
This internal error just happened to sbd. when joining #conduit:fachschaften.org (the main Conduit room) with Synapse:

```
8ZhxlGdY - Error attempting to resolve state at missing prev_events
Sep 22 10:50:55 debian matrix-synapse[3275]: Traceback (most recent call last):
Sep 22 10:50:55 debian matrix-synapse[3275]:   File "/home/synapse/synapse/env/lib/python3.9/site-packages/synapse/handlers/federation_event.py", line 710, in _resolve_state_at_missing_prevs
Sep 22 10:50:55 debian matrix-synapse[3275]:     remote_state = await self._get_state_after_missing_prev_event(
Sep 22 10:50:55 debian matrix-synapse[3275]:   File "/home/synapse/synapse/env/lib/python3.9/site-packages/synapse/handlers/federation_event.py", line 817, in _get_state_after_missing_prev_event
Sep 22 10:50:55 debian matrix-synapse[3275]:     await self._get_events_and_persist(
Sep 22 10:50:55 debian matrix-synapse[3275]:   File "/home/synapse/synapse/env/lib/python3.9/site-packages/synapse/handlers/federation_event.py", line 1171, in _get_events_and_persist
Sep 22 10:50:55 debian matrix-synapse[3275]:     await self._auth_and_persist_fetched_events(destination, room_id, roots)
Sep 22 10:50:55 debian matrix-synapse[3275]:   File "/home/synapse/synapse/env/lib/python3.9/site-packages/synapse/handlers/federation_event.py", line 1240, in _auth_and_persist_fetched_events
Sep 22 10:50:55 debian matrix-synapse[3275]:     await self.persist_events_and_notify(
Sep 22 10:50:55 debian matrix-synapse[3275]:   File "/home/synapse/synapse/env/lib/python3.9/site-packages/synapse/handlers/federation_event.py", line 1767, in persist_events_and_notify
Sep 22 10:50:55 debian matrix-synapse[3275]:     events, max_stream_token = await self._storage.persistence.persist_events(
Sep 22 10:50:55 debian matrix-synapse[3275]:   File "/home/synapse/synapse/env/lib/python3.9/site-packages/synapse/storage/persist_events.py", line 322, in persist_events
Sep 22 10:50:55 debian matrix-synapse[3275]:     ret_vals = await yieldable_gather_results(enqueue, partitioned.items())
Sep 22 10:50:55 debian matrix-synapse[3275]:   File "/home/synapse/synapse/env/lib/python3.9/site-packages/synapse/storage/persist_events.py", line 239, in handle_queue_loop
Sep 22 10:50:55 debian matrix-synapse[3275]:     ret = await self._per_item_callback(
Sep 22 10:50:55 debian matrix-synapse[3275]:   File "/home/synapse/synapse/env/lib/python3.9/site-packages/synapse/storage/persist_events.py", line 577, in _persist_event_batch
Sep 22 10:50:55 debian matrix-synapse[3275]:     await self.persist_events_store._persist_events_and_state_updates(
Sep 22 10:50:55 debian matrix-synapse[3275]:   File "/home/synapse/synapse/env/lib/python3.9/site-packages/synapse/storage/databases/main/events.py", line 174, in _persist_events_and_state_updates
Sep 22 10:50:55 debian matrix-synapse[3275]:     await self.db_pool.runInteraction(
Sep 22 10:50:55 debian matrix-synapse[3275]:   File "/home/synapse/synapse/env/lib/python3.9/site-packages/synapse/storage/database.py", line 686, in runInteraction
Sep 22 10:50:55 debian matrix-synapse[3275]:     result = await self.runWithConnection(
Sep 22 10:50:55 debian matrix-synapse[3275]:   File "/home/synapse/synapse/env/lib/python3.9/site-packages/synapse/storage/database.py", line 791, in runWithConnection
Sep 22 10:50:55 debian matrix-synapse[3275]:     return await make_deferred_yieldable(
Sep 22 10:50:55 debian matrix-synapse[3275]:   File "/home/synapse/synapse/env/lib/python3.9/site-packages/twisted/python/threadpool.py", line 238, in inContext
Sep 22 10:50:55 debian matrix-synapse[3275]:     result = inContext.theWork()  # type: ignore[attr-defined]
Sep 22 10:50:55 debian matrix-synapse[3275]:   File "/home/synapse/synapse/env/lib/python3.9/site-packages/twisted/python/threadpool.py", line 254, in <lambda>
Sep 22 10:50:55 debian matrix-synapse[3275]:     inContext.theWork = lambda: context.call(  # type: ignore[attr-defined]
Sep 22 10:50:55 debian matrix-synapse[3275]:   File "/home/synapse/synapse/env/lib/python3.9/site-packages/twisted/python/context.py", line 118, in callWithContext
Sep 22 10:50:55 debian matrix-synapse[3275]:     return self.currentContext().callWithContext(ctx, func, *args, **kw)
Sep 22 10:50:55 debian matrix-synapse[3275]:   File "/home/synapse/synapse/env/lib/python3.9/site-packages/twisted/python/context.py", line 83, in callWithContext
Sep 22 10:50:55 debian matrix-synapse[3275]:     return func(*args, **kw)
Sep 22 10:50:55 debian matrix-synapse[3275]:   File "/home/synapse/synapse/env/lib/python3.9/site-packages/twisted/enterprise/adbapi.py", line 293, in _runWithConnection
Sep 22 10:50:55 debian matrix-synapse[3275]:     compat.reraise(excValue, excTraceback)
Sep 22 10:50:55 debian matrix-synapse[3275]:   File "/home/synapse/synapse/env/lib/python3.9/site-packages/twisted/python/deprecate.py", line 298, in deprecatedFunction
Sep 22 10:50:55 debian matrix-synapse[3275]:     return function(*args, **kwargs)
Sep 22 10:50:55 debian matrix-synapse[3275]:   File "/home/synapse/synapse/env/lib/python3.9/site-packages/twisted/python/compat.py", line 404, in reraise
Sep 22 10:50:55 debian matrix-synapse[3275]:     raise exception.with_traceback(traceback)
Sep 22 10:50:55 debian matrix-synapse[3275]:   File "/home/synapse/synapse/env/lib/python3.9/site-packages/twisted/enterprise/adbapi.py", line 284, in _runWithConnection
Sep 22 10:50:55 debian matrix-synapse[3275]:     result = func(conn, *args, **kw)
Sep 22 10:50:55 debian matrix-synapse[3275]:   File "/home/synapse/synapse/env/lib/python3.9/site-packages/synapse/storage/database.py", line 786, in inner_func
Sep 22 10:50:55 debian matrix-synapse[3275]:     return func(db_conn, *args, **kwargs)
Sep 22 10:50:55 debian matrix-synapse[3275]:   File "/home/synapse/synapse/env/lib/python3.9/site-packages/synapse/storage/database.py", line 554, in new_transaction
Sep 22 10:50:55 debian matrix-synapse[3275]:     r = func(cursor, *args, **kwargs)
Sep 22 10:50:55 debian matrix-synapse[3275]:   File "/home/synapse/synapse/env/lib/python3.9/site-packages/synapse/logging/utils.py", line 69, in wrapped
Sep 22 10:50:55 debian matrix-synapse[3275]:     return f(*args, **kwargs)
Sep 22 10:50:55 debian matrix-synapse[3275]:   File "/home/synapse/synapse/env/lib/python3.9/site-packages/synapse/storage/databases/main/events.py", line 396, in _persist_events_txn
Sep 22 10:50:55 debian matrix-synapse[3275]:     self._update_metadata_tables_txn(
Sep 22 10:50:55 debian matrix-synapse[3275]:   File "/home/synapse/synapse/env/lib/python3.9/site-packages/synapse/storage/databases/main/events.py", line 1528, in _update_metadata_tables_txn
Sep 22 10:50:55 debian matrix-synapse[3275]:     self._store_room_members_txn(
Sep 22 10:50:55 debian matrix-synapse[3275]:   File "/home/synapse/synapse/env/lib/python3.9/site-packages/synapse/storage/databases/main/events.py", line 1666, in _store_room_members_txn
Sep 22 10:50:55 debian matrix-synapse[3275]:     self.db_pool.simple_insert_many_txn(
Sep 22 10:50:55 debian matrix-synapse[3275]:   File "/home/synapse/synapse/env/lib/python3.9/site-packages/synapse/storage/database.py", line 931, in simple_insert_many_txn
Sep 22 10:50:55 debian matrix-synapse[3275]:     txn.execute_values(sql, vals, fetch=False)
Sep 22 10:50:55 debian matrix-synapse[3275]:   File "/home/synapse/synapse/env/lib/python3.9/site-packages/synapse/storage/database.py", line 293, in execute_values
Sep 22 10:50:55 debian matrix-synapse[3275]:     return self._do_execute(
Sep 22 10:50:55 debian matrix-synapse[3275]:   File "/home/synapse/synapse/env/lib/python3.9/site-packages/synapse/storage/database.py", line 331, in _do_execute
Sep 22 10:50:55 debian matrix-synapse[3275]:     return func(sql, *args)
Sep 22 10:50:55 debian matrix-synapse[3275]:   File "/home/synapse/synapse/env/lib/python3.9/site-packages/synapse/storage/database.py", line 294, in <lambda>
Sep 22 10:50:55 debian matrix-synapse[3275]:     lambda *x: execute_values(self.txn, *x, fetch=fetch), sql, *args
Sep 22 10:50:55 debian matrix-synapse[3275]:   File "/home/synapse/synapse/env/lib/python3.9/site-packages/psycopg2/extras.py", line 1267, in execute_values
Sep 22 10:50:55 debian matrix-synapse[3275]:     parts.append(cur.mogrify(template, args))
Sep 22 10:50:55 debian matrix-synapse[3275]: ValueError: A string literal cannot contain NUL (0x00) characters.
```

_Originally posted by @jplatte in https://github.com/matrix-org/synapse/issues/9341#issuecomment-924790608_

H-Shay:
Hopefully fixes #10886, the change seems quite simple but I could be missing something here!


DMRobertson (commented):
Seems reasonable. Have we had a chance to test this? Is there some kind of test we could write to cover this?

synapse/storage/databases/main/events.py
```diff
@@ -1641,7 +1641,7 @@ def _store_room_members_txn(self, txn, events, backfilled):
         """Store a room member in the database."""
 
         def str_or_none(val: Any) -> Optional[str]:
```
clokep, line 1643: I wonder if we should update the name of this now that it does more than ensuring it is a string? 🤷 It's a pretty simple function so not a big deal.

changelog.d/11230.bugfix
```diff
@@ -0,0 +1,2 @@
+Fix a bug wherein display names or avatar URLs containing null bytes cause an internal server error
+when stored in the DB.
```
squahtx, line 2:
```suggestion
Fix a long-standing bug wherein display names or avatar URLs containing null bytes cause an internal server error
when stored in the DB.
```
I think this bug's been around forever?

H-Shay:
> Seems reasonable. Have we had a chance to test this? Is there some kind of test we could write to cover this?

@DMRobertson: I added a test that I don't think we should permanently add, it's more of a proof-of-concept that the change works. The test is pretty ugly-I couldn't figure out how to get access to the function I was testing so I copied it over. Nevertheless, the test as it is breaks (on postgres only) in the same way as the error message of the issue that is being addressed without the code change, and passes when the change is added. Let me know if you think this is sufficient!

changelog.d/11230.bugfix
H-Shay: Indeed, I have updated it.

synapse/storage/databases/main/events.py
H-Shay: I updated the name, let me know if you think it's a little clearer.

DMRobertson:
> @DMRobertson: I added a test that I don't think we should permanently add, it's more of a proof-of-concept that the change works. The test is pretty ugly-I couldn't figure out how to get access to the function I was testing so I copied it over. 

I had in mind a test that would try to reproduce the issue from the "outside". My (rough!) understanding of the issue is that

- We try to join a room where someone's name or avatar has a NULL codepoint
- That name/avatar was set as part of an event at some point in the room's history
- Synapse parses the JSON description of that event into a object `e` in memory
- Synapse queues `e` for persistence to the DB
- When `e` comes to be persisted, psycopg2/postgres complains about the NULL codepoint.

The last two bullets are what we're keen to test here. From a bit of reading, it looks like we can call `EventsPersistenceStorage.persist_event` with the `event` and `context` you have in your test already.  We should then be able to remove your copy of `_store_room_members_txn` from the test body. 

I think the test wants to do something like 

```python
self.get_success(
    self.hs.get_storage().persistence.persist_event(...)
)
```
If I understand right, that should fail before your change and succeed after your change. The test should also confirm that the user's event and avatar are correctly stored to the DB, using e.g. `RoomMemberWorkerStore.get_users_in_room_with_profiles`.

----
Does that make sense?

Being honest, I think the above is probably overkill for a small change like this. (I just like to have a test that proves we've made the right change.) If the above is tricky and just eating up time, we can drop it.

reivilibre:
An alternative approach with the testing would be to do it quite a long way from the outside using the test helpers that do the REST requests — I know it's not very unit test-like, but it's useful information to know that the overall request doesn't fail if a user joins a room, sets their name & avatar URL to have null bytes in, and so on. (Thinking about it, that could be a useful addition to Complement since other HS implementations might also get bitten by null bytes.)

tests/storage/test_roommember.py
```diff
@@ -161,6 +161,60 @@ def test_get_joined_users_from_context(self):
...
+                prev_event_ids=[bob_event.event_id],
+                type="m.test.1",
+                content={"displayname": "bad\u0000null", "membership": Membership.JOIN},
+            )
+        )
+
+        # pass event to _store_room_members_txn and verify that it doesn't blow up
+        self.get_success(
+            self.store.db_pool.runInteraction(
+                "store room members", _store_room_members_txn, [event], False
```
reivilibre, line 214:
```suggestion
                "store room members", self.hs.get_storage().persistence.persist_events_store._store_room_members_txn, [event], False
```
That one was hard to dig up, but doing this lets you avoid copying the function (so there won't be any chance we forget to copy it into the test if we change/break it in the future)

tests/storage/test_roommember.py
```diff
@@ -161,6 +161,60 @@ def test_get_joined_users_from_context(self):
...
+        event, context = self.get_success(
+            event_injection.create_event(
+                self.hs,
+                room_id=room,
+                sender=self.u_alice,
+                prev_event_ids=[bob_event.event_id],
+                type="m.test.1",
+                content={"displayname": "bad\u0000null", "membership": Membership.JOIN},
+            )
+        )
```
reivilibre, line 209: I usually prefer to avoid injecting events if you can avoid it (reserve that for when you need invalid events to be persisted for some reason, or to pretend you received one over federation).

Since you have alice's access token and everything, you can use something like

```
        new_profile = {"displayname": "ali\u0000ce"}
        self.helper.change_membership(
            room, self.u_alice, self.u_alice, "join", extra_data=new_profile, tok=self.t_alice
        )
```

tests/storage/test_roommember.py
reivilibre: I don't know if you really need to test this function explicitly/individually if you properly demonstrate that `self.helper.change_membership(` (as discussed in another thread) works (assuming it didn't work before!).

tests/storage/test_roommember.py
H-Shay: The call to `change_membership` does indeed break without the change I made to `_store_room_members_txn` and passes with the change, so hopefully this is sufficient? Thanks for pointing out the use cases for injecting events, it's helpful to be told these sorts of things :)

tests/storage/test_roommember.py
H-Shay: Thanks for showing how to access `_store_room_members_txn`-even though it seems like it's not necessary to call now-I really scratched my head over how to get access to it so I appreciate being shown the answer!

tests/storage/test_roommember.py
reivilibre: Maybe it would be worth also querying the state of the database and checking that they contain Nones instead of something else.
(It just feels weird to have an assert-less test in some strange way, but it's also worth noting that the behaviour is replacing the name with None rather than e.g. leaving it as it was before, if that makes sense?)

H-Shay: Totally makes sense, I shall give that a go.

tests/storage/test_roommember.py
H-Shay: Oh interesting-thanks for pointing that out.

tests/storage/test_roommember.py
```diff
@@ -177,6 +177,16 @@ def test__null_byte_in_display_name_properly_handled(self):
...
 
+        res = self.get_success(
+            self.store.db_pool.simple_select_onecol(
+                "room_memberships", {"user_id": "@alice:test"}, "display_name"
+            )
+        )
+        # verify that the display name was alice before change in membership
+        self.assertEqual(res[0], "alice")
+        # verify that it is now None
+        self.assertEqual(res[1], None)
```
reivilibre, line 188: Bah, slight problem here is that this is *probably* fine for tests, but it's relying on the database returning the rows in insertion order.

SQLite secretly maintains a 'rowid' (row ID) per row and stores rows in a B* tree keyed by those, so I can see it returning them in insertion order.
I assume Postgres does something similar — basically, returning them in order of insertion is 'easiest' for this query (it'd be more work to jumble them up) so I reckon this test will work!
(If the database was using an index to retrieve the rows, then it would probably return them in order of that index)

However, nothing guarantees that these databases will do that. I think a notable time that databases will violate this insertion order is if some rows get deleted and the database shuffles things around (e.g. Postgres' VACUUM; SQLite can recycles rowids afaik and SQLite will generate rowids somewhat randomly if you reach the limit).

I was curious about SQLite's VACUUM and in my case it kept the order when vacuuming, seemingly just rewriting rowids, oh well (I was hoping for some rows to be dramatically reordered! :P).
```sql
sqlite> create table tab1 (val TEXT);
sqlite> insert into tab1 values ("a"), ("b"), ("c");
sqlite> select rowid, val from tab1;
1|a
2|b
3|c
sqlite> delete from tab1 where val = "b";
sqlite> insert into tab1 values ("d"), ("e"), ("f");
sqlite> select rowid, val from tab1;
1|a
3|c
4|d
5|e
6|f
sqlite> vacuum;
sqlite> select rowid, val from tab1;
1|a
2|c
3|d
4|e
5|f
```

Anyway, excuse the tangent, but this is the kind of thing that has the tendency to bite you in the bum :(.
I think what I'd do here is:
* select from `room_memberships` before changing Alice's name
  * check your preconditions *and make a note of the event_id*
* update the name
* select from `room_memberships` again, **filtering out the row with the event_id you had before** and check your postcondition

Check the expected number of rows at each step to make sure that if we get more rows for some reason, they don't cause confusion in the future when the test breaks and some poor soul has to figure out why :p.

I'd probably use a list comprehension to do the filtering in Python (mostly since doing it in SQL here is a pain and it's not as though it's performance sensitive since it's a test), something like that:
```python
rows = [
	row for row in rows if row["event_id"] != first_event_id
]
```

reivilibre (changes requested):
The test looks good other than relying on accidental ordering, though :)

Thanks!

reivilibre (approved):
Thank you!

PR #11231, opened 2021-11-02
Add index to `local_group_updates.stream_id`

erikjohnston:
This should speed up startup times and generally increase performance of
groups.

### DIFF
--- a/synapse/storage/schema/main/delta/65/03_local_group_updates.sql
+++ b/synapse/storage/schema/main/delta/65/03_local_group_updates.sql
@@ -1,0 +1,18 @@
+/* Copyright 2021 The Matrix.org Foundation C.I.C
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+-- Check old events for thread relations.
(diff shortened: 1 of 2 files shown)


anoadragon453 (changes requested):
Some small things I noticed.

I did see some other queries against `local_group_updates` which searched using `user_id`. However, since it's not a unique column, any attempts to add further indexes didn't look to result in any less costly queries.

synapse/storage/schema/main/delta/65/03_local_group_updates.sql
```diff
@@ -0,0 +1,18 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+-- Check index on `local_group_updates.stream_id`.
+INSERT INTO background_updates (ordering, update_name, progress_json) VALUES
+  (6503, 'local_group_updates_index', '{}');
```
anoadragon453, line 18: Looks like someone else sniped us and took `6503`: https://github.com/matrix-org/synapse/blob/237f7eb87ad097ecf40f945630ac49cef1aa1154/synapse/storage/schema/main/delta/65/03remove_hidden_devices_from_device_inbox.sql#L21-L22

Bump to `6504`? :innocent:

anoadragon453 (approved):
Lgtm otherwise, thanks!

synapse/storage/schema/main/delta/65/03_local_group_updates.sql
erikjohnston: It doesn't really matter, but sure

anoadragon453: Oh! I suppose it doesn't matter as this query doesn't rely on being after `remove_hidden_devices_from_device_inbox` for any reason, correct?

erikjohnston: Yup, exactly!

PR #11233, opened 2021-11-02
Add twine and towncrier as dev dependencies

erikjohnston:
We don't pin them as we execute them as commands, rather than use them
as libs.


anoadragon453 (approved):
One of those "why didn't we do this sooner" fixes. Thanks!

PR #11236, opened 2021-11-02
Support filtering by relations per MSC3440

### LINKED ISSUE #11086: Implement MSC3440 relation filtering
[MSC3440](https://github.com/matrix-org/matrix-doc/pull/3440) suggests implementing filtering on relations, specifically filtering for parent events which have:

* `relation_types`: Another event referring to them of particular relation type(s).
* `relation_senders`: Another event referring to them from particular sender(s).

In order to support this we'll need to:

* [ ] Update the filtering code to be async.
* [ ] Update the filtering code to store/process the new filter fields.
* [ ] Update the filtering code to fetch the event relations and filter on them.
* [ ] Update the message pagination code to filter based on the above information.

These will have an unstable values of `io.element.relation_types` and `io.element.relation_senders` as filter keys.

clokep:
Implements filtering via relations per[MSC3440](https://github.com/matrix-org/matrix-doc/pull/3440), specifically filtering for parent events which have:

* `relation_types`: Another event referring to them of particular relation type(s).
* `relation_senders`: Another event referring to them from particular sender(s).

This PR uses the unstable prefixes and required a bit of refactoring (I can split this work out to a separate PR if preferred):

* Updates the `filter_*` methods to be async
* Allow the `FIlter` to access the datastore (via `HomeServer`)

After that we can update the `Filter` implementation to take the additional filterings parameters and apply those in:

* The pagination infrastructure from the database.
* The manual checks against the event objects.

Fixes #11086

### DIFF
--- a/synapse/api/filtering.py
+++ b/synapse/api/filtering.py
@@ -86,6 +86,9 @@ ROOM_EVENT_FILTER_SCHEMA = {
         # cf https://github.com/matrix-org/matrix-doc/pull/2326
         "org.matrix.labels": {"type": "array", "items": {"type": "string"}},
         "org.matrix.not_labels": {"type": "array", "items": {"type": "string"}},
+        # MSC3440, filtering by event relations.
+        "io.element.relation_senders": {"type": "array", "items": {"type": "string"}},
+        "io.element.relation_types": {"type": "array", "items": {"type": "string"}},
     },
 }
 
@@ -291,6 +294,7 @@ class FilterCollection:
 class Filter:
     def __init__(self, hs: "HomeServer", filter_json: JsonDict):
         self._hs = hs
+        self._store = hs.get_datastore()
         self.filter_json = filter_json
 
         self.limit = filter_json.get("limit", 10)
@@ -312,6 +316,11 @@ class Filter:
 
         self.labels = filter_json.get("org.matrix.labels", None)
         self.not_labels = filter_json.get("org.matrix.not_labels", [])
+
+        self.relation_senders = self.filter_json.get(
+            "io.element.relation_senders", None
+        )
+        self.relation_types = self.filter_json.get("io.element.relation_types", None)
 
     def filters_all_types(self) -> bool:
         return "*" in self.not_types
@@ -379,6 +388,14 @@ class Filter:
                 contains_url = isinstance(content.get("url"), str)
                 if contains_url_filter != contains_url:
                     return False
+
+            # TODO This is wildly inefficient.
+            if isinstance(event, EventBase) and (
+                self.relation_senders or self.relation_types
+            ):
+                return await self._store.has_event_relations(
+                    event.event_id, self.relation_senders, self.relation_types
+                )
 
             return True
 
--- a/synapse/api/filtering.py
+++ b/synapse/api/filtering.py
@@ -331,7 +331,7 @@ class Filter:
     def filters_all_rooms(self) -> bool:
         return "*" in self.not_rooms
 
-    async def check(self, event: FilterEvent) -> bool:
+    async def _check(self, event: FilterEvent) -> bool:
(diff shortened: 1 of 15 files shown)


synapse/api/filtering.py
```diff
@@ -420,8 +454,30 @@ def filter_rooms(self, room_ids: Iterable[str]) -> Set[str]:
 
         return room_ids
 
-    def filter(self, events: Iterable[FilterEvent]) -> List[FilterEvent]:
-        return list(filter(self.check, events))
+    async def _check_event_relations(
+        self, events: Iterable[FilterEvent]
+    ) -> List[FilterEvent]:
```
clokep, line 459: Separating this out to do a single query for all events might be a bit premature optimization, but doing a separate query per event seemed quite expensive.

synapse/api/filtering.py
```diff
@@ -444,6 +500,3 @@ def _matches_wildcard(actual_value: Optional[str], filter_value: str) -> bool:
         return actual_value.startswith(type_prefix)
     else:
         return actual_value == filter_value
-
-
-DEFAULT_FILTER_COLLECTION = FilterCollection({})
```
clokep, line 449: This needed to be moved somewhere that it can have access to a `HomeServer`.

synapse/storage/databases/main/stream.py
```diff
@@ -1133,32 +1156,51 @@ def _paginate_room_events_txn(
...
-                select_keywords += "DISTINCT"
+                # Multiple labels could cause the same event to appear multiple times.
+                needs_distinct = True
+
+        # If there is a filter on relation_senders and relation_types join to the
+        # relations table.
+        if event_filter and (
+            event_filter.relation_senders or event_filter.relation_types
+        ):
+            # Filtering by relations could cause thes ame event to appear multiple
```
squahtx, line 1181: typo
```suggestion
            # Filtering by relations could cause the same event to appear multiple
```

squahtx:
I wonder if it'd be cleaner to pass the HomeServer as an argument to `Filter.check` instead of threading it through the constructor...
(I haven't checked)

squahtx:
Can we treat empty lists as a filter allowing nothing through instead of no filter?

I'm imagining a case where someone writes the obvious code to query for something like "all events with a reaction from new users who joined within the last hour", where the list of users comes from another query.
It'd be unexpected for the filter to allow all events through if the query for newly-joined users happened to return 0 users.

squahtx (approved):
LGTM apart from the comments above

clokep:
> I wonder if it'd be cleaner to pass the HomeServer as an argument to `Filter.check` instead of threading it through the constructor... (I haven't checked)

I'm not sure that would be any clearer, it feels nicer to have it when we instantiate it (it should be fewer call sites), but I can try it if you would like!

> Can we treat empty lists as a filter allowing nothing through instead of no filter?
> 
> I'm imagining a case where someone writes the obvious code to query for something like "all events with a reaction from new users who joined within the last hour", where the list of users comes from another query. It'd be unexpected for the filter to allow all events through if the query for newly-joined users happened to return 0 users.

Hmmm, the [current filtering](https://matrix.org/docs/spec/client_server/latest#post-matrix-client-r0-user-userid-filter) has things like:

>  A list of event types to include. If this list is absent then all event types are included.

Which is similar language to what I put into [MSC3440](https://github.com/matrix-org/matrix-doc/pull/3440). Is there a particular spot you're looking at to apply this? I think I implemented the filtering in the same way as the previous filters? Are you just asking for some more explicit `is None` checks?

clokep:
@squahtx and I discussed this a bit on chat; he was hoping to see `is (not) None` checks added, but our conclusion was that it probably makes more sense to be consistent with the pre-existing code.

The conclusion is that we're going to leave this for now. We can re-visit in the future if necessary.

PR #11239, opened 2021-11-03
Remove a debug statement from tests.

clokep:
There was a leftover `print()` statement in one of the tests which I noticed. 😄


DMRobertson (approved):
No other matches for 'print\(' in `tests`.

There were a few in `synapse`, but this isn't the time to start an audit of when we should be printing versus logging!

PR #11114, opened 2021-10-18
Fix historical messages backfilling in random order on remote homeservers (MSC2716)

MadLittleMods:
*Started as an investigation of https://github.com/matrix-org/synapse/issues/11091 but then realized the backfilled messages are really weird on remote homeservers.*

---

Fix historical messages backfilling in random order on remote homeservers ([MSC2716](https://github.com/matrix-org/matrix-doc/pull/2716)).

Fix https://github.com/matrix-org/synapse/issues/11091
Fix https://github.com/matrix-org/synapse/issues/10764

 1. Made the `/backfill` response return messages in `(depth, stream_ordering)` order (previously only sorted by `depth`)
    - Technically, it shouldn't really matter how `/backfill` returns things but I'm just trying to make the `stream_ordering` a little more consistent from the origin to the remote homeservers in order to get the order of messages from `/messages` consistent ([sorted by `(topological_ordering, stream_ordering)`](https://github.com/matrix-org/synapse/blob/develop/docs/development/room-dag-concepts.md#depth-and-stream-ordering)).
    - Even now that we return backfilled messages in order, it still doesn't guarantee the same `stream_ordering` (and more importantly the [`/messages` order](https://github.com/matrix-org/synapse/blob/develop/docs/development/room-dag-concepts.md#depth-and-stream-ordering)) on the other server. For example, if a room has a bunch of history imported and someone visits a permalink to a historical message back in time, their homeserver will skip over the historical messages in between and insert the permalink as the next message in the `stream_order` and totally throw off the sort.
       - This will be even more the case when we add the [MSC3030 jump to date API endpoint](https://github.com/matrix-org/matrix-doc/pull/3030) so the static archives can navigate and jump to a certain date.
       - We're solving this in the future by switching to [online topological ordering](https://github.com/matrix-org/gomatrixserverlib/issues/187) and [chunking](https://github.com/matrix-org/synapse/issues/3785) which by its nature will apply retroactively to fix any inconsistencies introduced by people permalinking
 1. As we're navigating `prev_events` to return in `/backfill`, we order by `depth` first (newest -> oldest) and now also tie-break based on the `stream_ordering` (newest -> oldest). This is technically important because MSC2716 inserts a bunch of historical messages at the same `depth` so it's best to be prescriptive about which ones we should process first. In reality, I think the code already looped over the historical messages as expected because the database is already in order.
[...]

PR #11242, opened 2021-11-03
Split out federated PDU retrieval into a non-cached version

MadLittleMods:
Split out federated PDU retrieval into a non-cached version

Context: https://github.com/matrix-org/synapse/pull/11114/files#r741643968

This code is unused but thought it seemed like a small refactor with a little bit of value to split some of the logic out. And the caching/rate-limiting is handled in the `get_pdu` consumer.

### Pull Request Checklist

* [x] [Code style](https://matrix-org.github.io/synapse/latest/code_style.html) is correct (run the [linters](https://matrix-org.github.io/synapse/latest/development/contributing_guide.html#run-the-linters))

### DIFF
--- a/synapse/federation/federation_client.py
+++ b/synapse/federation/federation_client.py
@@ -286,8 +286,8 @@ class FederationClient(FederationBase):
         timeout: Optional[int] = None,
     ) -> Optional[EventBase]:
         """Requests the PDU with given origin and ID from the remote home
-        server.
-        Does not have any caching or rate limiting!
+        server. Does not have any caching or rate limiting!
+
         Args:
             destination: Which homeserver to query
             event_id: event to fetch
@@ -297,8 +297,10 @@ class FederationClient(FederationBase):
                 of the current block of PDUs. Defaults to `False`
             timeout: How long to try (in ms) each destination for before
                 moving to the next destination. None indicates no timeout.
+
         Returns:
             The requested PDU, or None if we were unable to find it.
+
         Raises:
             SynapseError, NotRetryingDestination, FederationDeniedError
         """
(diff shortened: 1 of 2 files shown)


synapse/federation/federation_client.py
```diff
@@ -277,6 +277,60 @@ async def backfill(
...
+            SynapseError, NotRetryingDestination, FederationDeniedError
+        """
+
+        signed_pdu = None
+
+        transaction_data = await self.transport_layer.get_event(
+            destination, event_id, timeout=timeout
+        )
+
+        logger.info(
```
erikjohnston, line 314:
```suggestion
        logger.debug(
```

We will log the fact that we've requested the event elsewhere, so I don't think there is much need in upping the log level to `INFO`?

synapse/federation/federation_client.py
```diff
@@ -277,6 +277,60 @@ async def backfill(
...
+            for p in transaction_data["pdus"]
+        ]
+
+        if pdu_list and pdu_list[0]:
+            pdu = pdu_list[0]
+
+            # Check signatures are correct.
+            signed_pdu = await self._check_sigs_and_hash(room_version, pdu)
+
+        return signed_pdu
```
erikjohnston, line 332:
```suggestion
            retuurn await self._check_sigs_and_hash(room_version, pdu)

        return None
```

And then we can remove the `signed_pdu = None` line at the top?
```suggestion
            signed_pdu = await self._check_sigs_and_hash(room_version, pdu)

        return signed_pdu
```

synapse/federation/federation_client.py
MadLittleMods: Good catch, no need for this and just adjusted for debugging initially before

erikjohnston (approved):

MadLittleMods:
Thanks for the review @erikjohnston 🦛

Issue #8094, opened 2020-08-14
lots of AssertionError: 'Attempting to create an event with no prev_events'


richvdh:
These suggest something is wrong with the room in question. We should filter out attempts to send to such rooms earlier on.

richvdh:
https://sentry.matrix.org/sentry/synapse-matrixorg/issues/119955

richvdh:
This assertionerror is what happens where previously we would have returned 'no create event in auth events'; as such this is closely related to #7642.

These rooms have no entries in `event_forward_extremities` or `current_state_events`. 

They seem to divide into two sets:

 * rooms with an appservice user (typically an RSS bot) where the last regular user left some time ago. I suspect they have been cleared out by the the `delete_old_current_state_events` database schema update, though am a bit confused as to why this has happened given the appservice user is still a member.
 * rooms where we have received one or more `invite` events over federation, but apparently never (successfully?) joined the room.

richvdh:
> * rooms with an appservice user (typically an RSS bot) where the last regular user left some time ago. I suspect they have been cleared out by the the `delete_old_current_state_events` database schema update, though am a bit confused as to why this has happened given the appservice user is still a member.

It might be worth noting that, at the time we ran this update on matrix.org, it looked at `current_state_events.membership` to see which rooms were still active (that was later changed in #7854). But I don't know why that wouldn't be populated either, at least on matrix.org (other servers might have had a horrible mess, due to #6923).

clokep:
New sentry issue: https://sentry.matrix.org/sentry/synapse-matrixorg/issues/189680/

richvdh:
Still very much a thing: https://sentry.matrix.org/sentry/synapse-matrixorg/issues/200257

anoadragon453:
This was noticed again when @jaywink was trying to invite `@neb_alertmanager:matrix.org` to a room on `2021-12-01 14:08:56` with ID containing `TZrnFfySWI`.

Relevant stack trace:
<details>

<summary>Stack trace</summary>

```
2021-12-01 14:08:56,850 - synapse.http.server - 101 - ERROR - PUT-2311876 - Failed handle request via 'RoomSendEventRestServlet': <XForwardedForRequest at 0x7fe7a53f3bd8 method='PUT' uri='/_matrix/client/r0/rooms/%21redacted:federator.dev/send/m.room.message/go1638367736821254338?access_token=<redacted>' clientproto='HTTP/1.1' site='8138'>
[...]

PR #11243, opened 2021-11-03
Allow events to be created with no `prev_events` (MSC2716)

MadLittleMods:
Allow events to be created with no `prev_events`. The event still needs to have `auth_events` defined to be valid.

Split out from https://github.com/matrix-org/synapse/pull/11114

### Dev notes

```
SYNAPSE_POSTGRES=1 SYNAPSE_TEST_LOG_LEVEL=INFO python -m twisted.trial tests.handlers.test_message.EventCreationTestCase
```

### Pull Request Checklist


synapse/handlers/message.py
```diff
@@ -949,14 +949,24 @@ async def create_new_client_event(
...
+        room_version_obj = await self.store.get_room_version(builder.room_id)
+        if room_version_obj.msc2716_empty_prev_events:
+            # We allow events with no `prev_events` but it better have some `auth_events`
+            assert (
+                builder.type == EventTypes.Create
+                or len(prev_event_ids) > 0
+                # Allow an event to have empty list of prev_event_ids
+                # only if it has auth_event_ids.
+                or (auth_event_ids and len(auth_event_ids) > 0)
+            ), "Attempting to create an event with no prev_events or auth_event_ids"
```
anoadragon453, line 964: To be slightly more specific:

```suggestion
            ), "Attempting to create a non-m.room.create event with no prev_events or auth_event_ids"
```

synapse/handlers/message.py
```diff
@@ -949,14 +949,24 @@ async def create_new_client_event(
...
+        else:
+            # we now ought to have some prev_events (unless it's a create event).
+            assert (
+                builder.type == EventTypes.Create or len(prev_event_ids) > 0
+            ), "Attempting to create an event with no prev_events"
```
anoadragon453, line 969:
```suggestion
            ), "Attempting to create a non-m.room.create event with no prev_events"
```

changelog.d/11243.feature
```diff
@@ -0,0 +1 @@
+Allow experimental room version `org.matrix.msc2716v4` to allow events to be created without `prev_events` (only `auth_events`).
```
anoadragon453, line 1:
```suggestion
Add experimental room version `org.matrix.msc2716v4` to allow events to be created without `prev_events` (only `auth_events`).
```

synapse/handlers/room_member.py
```diff
@@ -644,7 +644,7 @@ async def update_membership_locked(
             if block_invite:
                 raise SynapseError(403, "Invites have been disabled on this server")
 
-        if prev_event_ids:
+        if prev_event_ids is not None:
```
anoadragon453, line 647:
```suggestion
        # An empty prev_events list is allowed by room version "org.matrix.msc2716v4"
        if prev_event_ids is not None:
```

anoadragon453 (changes requested):
It would be nice to have some tests for this, unless it's expected to be covered by https://github.com/matrix-org/complement/pull/214?

tests/handlers/test_message.py
```diff
@@ -156,6 +158,79 @@ def test_duplicated_txn_id_one_call(self):
...
+                    "sender": self.requester.user.to_string(),
+                    "content": {"msgtype": "m.text", "body": random_string(5)},
+                },
+                prev_event_ids=[],
+                # The event should be rejected when there are no auth_events
+                auth_event_ids=[],
+            ),
+            AssertionError,
+        )
+
```
MadLittleMods, line 233: > It would be nice to have some tests for this, unless it's expected to be covered by https://github.com/matrix-org/complement/pull/214?

All of the MSC2716 Complement tests cover this use case since we use it so much there. I've just added some integration tests for this.

I didn't add tests for every case to be completely exhaustive but did make sure this extra functionality works and that we don't allow events with empty `auth_events`

 - Creating a `m.room.create` event
    - :x: In MSC2716 room version
    - :x: In existing stable room version
 - Creating event with empty `prev_event`
    - ✅ In MSC2716 room version
    - :x: In existing stable room version
 - Creating event with empty `prev_event` and empty `auth_event`  (make sure rejected)
    - ✅ In MSC2716 room version
    - :x: In existing stable room version
 - Creating event with `prev_events`
    - :x: In MSC2716 room version
    - :x: In existing stable room version
 - Creating event with `prev_events` but empty `auth_event` (make sure rejected)
    - :x: In MSC2716 room version
    - :x: In existing stable room version

anoadragon453 (approved):
Thanks for the tests! This lgtm, I'd like to just ask another member of the team to double-check the changes make sense (and don't have any security implications).

tests/handlers/test_message.py
```diff
@@ -156,6 +158,79 @@ def test_duplicated_txn_id_one_call(self):
...
+                    "room_id": room_id,
+                    "sender": self.requester.user.to_string(),
+                    "content": {"msgtype": "m.text", "body": random_string(5)},
+                },
+                # Empty prev_events is the key thing we're testing here
+                prev_event_ids=[],
+                auth_event_ids=[memberEvent.event_id],
+            )
+        )
+        self.assertIsNotNone(event)
```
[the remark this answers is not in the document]
MadLittleMods: We're testing the `create_event` function itself in this test, it seems better to not rely on typing in the implementation which could change on purpose or accident. The whole point of the test is to make sure an event is created and exists.

synapse/handlers/message.py
```diff
@@ -949,14 +950,24 @@ async def create_new_client_event(
...
-        ), "Attempting to create an event with no prev_events"
+        room_version_obj = await self.store.get_room_version(builder.room_id)
+        if room_version_obj.msc2716_empty_prev_events:
+            # We allow events with no `prev_events` but it better have some `auth_events`
+            assert (
+                builder.type == EventTypes.Create
+                or len(prev_event_ids) > 0
+                # Allow an event to have empty list of prev_event_ids
+                # only if it has auth_event_ids.
+                or (auth_event_ids and len(auth_event_ids) > 0)
```
erikjohnston, line 964: I'd be tempted to add an assertion that this is an insertion event, to ensure that we're doing the right thing elsewhere. Specifically, we do have a bug somewhere that means that we sometimes try and send events without any prev events and it would be good to keep catching those.

erikjohnston:
```suggestion
                or auth_event_ids
```

if `len(auth_event_ids) == 0` then `auth_event_ids` will be falsey, so the extra check is redundant.

erikjohnston (commented):
Otherwise lgtm

tests/handlers/test_message.py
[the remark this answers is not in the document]
anoadragon453: That's a fair point.

anoadragon453:
I've noticed that interestingly, we seem to have attempts to create events without `prev_events` all the time, which will now start going through: https://sentry.matrix.org/sentry/synapse-matrixorg/issues/200257/?query=is%3Aunresolved (link for those with access).

richvdh:
see https://github.com/matrix-org/synapse/issues/8094 for some background on that assertion.

I'm not enthusiastic about any change which makes it more likely for us to start accidentally creating regular events without `prev_events`, so +1 to the suggestion that we only allow this for insertion events.

synapse/handlers/message.py
MadLittleMods: > I'm not enthusiastic about any change which makes it more likely for us to start accidentally creating regular events without prev_events, so +1 to the suggestion that we only allow this for insertion events.

If we're only allowing for certain type of events, we need to allow no `prev_events` for all the state events too. The state event chain (the alice events at the top) is actually the one with no `prev_events`. We connect the insertion event to the floating state chain.

The only time, the insertion event itself is floating is when no `state_events_at_start` are defined.

synapse/handlers/message.py
```diff
@@ -949,14 +950,24 @@ async def create_new_client_event(
...
         # create event in auth events")
-        assert (
-            builder.type == EventTypes.Create or len(prev_event_ids) > 0
-        ), "Attempting to create an event with no prev_events"
+        room_version_obj = await self.store.get_room_version(builder.room_id)
+        if room_version_obj.msc2716_empty_prev_events:
+            # We allow events with no `prev_events` but it better have some `auth_events`
+            assert (
+                builder.type == EventTypes.Create
+                or len(prev_event_ids) > 0
```
MadLittleMods, line 961: > if `len(auth_event_ids) == 0` then `auth_event_ids` will be falsey, so the extra check is redundant.
>
> *-- https://github.com/matrix-org/synapse/pull/11243#discussion_r745408186*

Based on the other comment, can we apply the same to `prev_event_ids`?

synapse/handlers/message.py
erikjohnston: > Is it really a problem to have events without `prev_events` though? They seem like they would be benign and a much better alternative to a homeserver lying out their teeth about a fake `prev_event`.

The problem is that for the normal case not having any `prev_event` means that we won't be able to calculate the state at the event.

---

One thought I had: can we allow events with no prev events if and only if they're either a) an insertion event or b) an outlier?

Separately: I'm not sure its correct that the insertion event points to anything else, as we need want federation to ask for the state at the insertion event, rather than any of the state events that it references?

synapse/handlers/message.py
```diff
@@ -949,14 +950,24 @@ async def create_new_client_event(
...
-        #
-        # do a quick sanity check here, rather than waiting until we've created the
+        # Do a quick sanity check here, rather than waiting until we've created the
         # event and then try to auth it (which fails with a somewhat confusing "No
         # create event in auth events")
-        assert (
-            builder.type == EventTypes.Create or len(prev_event_ids) > 0
-        ), "Attempting to create an event with no prev_events"
+        room_version_obj = await self.store.get_room_version(builder.room_id)
+        if room_version_obj.msc2716_empty_prev_events:
```
MadLittleMods, line 957: Does this actually require a new room version? (discussed at https://github.com/matrix-org/synapse/pull/11114#discussion_r733475645)

My thinking is that this code is only for creating events, not accepting events. So technically any other homeserver nowadays can create events with no `prev_events` now and it would work.

synapse/handlers/message.py
MadLittleMods: > The problem is that for the normal case not having any `prev_event` means that we won't be able to calculate the state at the event.

Perhaps there are two separate concerns here:

 1. Just making sure app code is doing the right thing for the majority of cases. We expect `prev_events` in all of the non-historical cases.
 1. Having events with no `prev_events` being a problem in a room


For 1, we can add a `allow_no_prev_events: [True|False]` parameter to `create_new_client_event` to only allow it when specified.

For 2, see https://github.com/matrix-org/synapse/pull/11243#discussion_r750816377. I think events with no `prev_events` work just fine in a room. If having no `prev_events` is a problem, then we probably shouldn't do it for the historical events either. Things seem to work without needing to worry about anything.

> One thought I had: can we allow events with no prev events if and only if they're either a) an insertion event or b) an outlier?

I think that would work but I'm leaning away from making any special cases.

synapse/handlers/message.py
```diff
@@ -918,6 +918,7 @@ async def create_new_client_event(
         full_state_ids_at_event = None
         if auth_event_ids is not None:
             # If auth events are provided, prev events must be also.
+            # prev_event_ids could be an empty array though.
             assert prev_event_ids is not None
 
             # Copy the full auth state before it stripped down
```
MadLittleMods, line 924: > Separately: I'm not sure its correct that the insertion event points to anything else, as we need want federation to ask for the state at the insertion event, rather than any of the state events that it references?
>
> *-- [@erikjohnston](https://github.com/matrix-org/synapse/pull/11243#discussion_r750312188)*

The insertion event is connected to the state chain so the whole historical batch can share the same `state_group`. It also looks nice in the DAG semantically to be able to see the state that authed the batch.

For the federation cases, it seems to work 🤔🤷. `auth_event_ids` probably

synapse/handlers/message.py
erikjohnston: Yup!

synapse/handlers/message.py
erikjohnston: I am not that surprised that we accept outliers with empty prev-events, but we'll need special handling for the insertion event to be accepted, i.e. we'll need some sort of check like "this is an insertion event so we need to go and fetch the state rather than trying to calculate it". Though that can be part of the history import MSC I guess

synapse/handlers/message.py
erikjohnston: Hmm, that surprises me a bit. I was very much imagining that the federation code would ask for the state at the insertion event. I guess it probably works because we fetch the state at the start of the chain of state events? And we return the full state, so we then can accept all the state events we add? I don't think that is necessary to make the whole batch have the same state group.

Also, I wonder if that'll make the client UI show the state events when backpaginating, making it look like the state was changed at those points, when actually they were changed much further back?

synapse/handlers/message.py
MadLittleMods: > we'll need special handling for the insertion event to be accepted, i.e. we'll need some sort of check like "this is an insertion event so we need to go and fetch the state rather than trying to calculate it"

Why is this the case? It seems to work in my Complement tests without any of this 🤔

synapse/handlers/message.py
MadLittleMods: @erikjohnston What are your thoughts here?

synapse/handlers/message.py
MadLittleMods: > I don't think that is necessary to make the whole batch have the same state group.

I can play around with it but that seems like something for another PR. It's pretty finicky to get right.


> Also, I wonder if that'll make the client UI show the state events when backpaginating, making it look like the state was changed at those points, when actually they were changed much further back?

Doesn't seem to be a problem for the local origin homeserver with Element probably because they are marked as `outliers`.

I want to test this on a federated homeserver though.

I'm trying to setup a local federated homeserver to see if it's a problem for remote federated servers. Currently working out the hostname/`server_name` problem mapping to a port locally (tips welcome) and basing the config off of the [demo script](https://github.com/matrix-org/synapse/blob/develop/demo/start.sh). I'm attempting to use [`/etc/hosts` and a nginx reverse proxy to `127.0.0.1:8448` (non-tls)](https://www.baeldung.com/linux/mapping-hostnames-ports) . Synapse is configured to be `tls: false` on `8008 and `8448`. This appears to work to get `my.matrix.host` and `other.matrix.host` but doesn't make it happy enough to be able to successfully fetch the room list.

Might just end up creating a Complement test around making sure the state isn't visible between batches ⏩ but would be nice to double-check with real Element about how it looks.

synapse/handlers/message.py
erikjohnston: Spooky! :ghost:  Is it doing something silly like dropping the event with no extremities and then doing a `/state` on the event that references it?

synapse/handlers/message.py
erikjohnston: (I swear this thread wasn't here yesterday, sorry :confused: )

> For 1, we can add a allow_no_prev_events: [True|False] parameter to create_new_client_event to only allow it when specified.

I think that's fine. I understand wanting to avoid special casing, but my assumption was that we'd need to add special handling for events with no prev events anyway, so it'd make sense to reuse that.

synapse/handlers/message.py
erikjohnston: > I'm trying to setup a federated homeserver locally to see if it's a problem for remote federated servers.

FYI you can start some preconfigured HSes that federate with each other by running `./demo/start.sh` (and `./demo/stop.sh`)

synapse/handlers/message.py
MadLittleMods: Removed the new room version and added `allow_no_prev_events: bool = False` to only allow this case whenever we explicitly set it ✅

synapse/handlers/message.py
MadLittleMods: > Also, I wonder if that'll make the client UI show the state events when backpaginating, making it look like the state was changed at those points, when actually they were changed much further back?

> Might just end up creating a Complement test around making sure the state isn't visible between batches ⏩ but would be nice to double-check with real Element about how it looks.

I updated the [Complement test which tests multiple batches over federation to check for historical state](https://github.com/matrix-org/complement/pull/214) and didn't see any problem as well.

MadLittleMods: > I don't think that is necessary to make the whole batch have the same state group.

I created https://github.com/matrix-org/synapse/pull/11487 to add a test around sharing the `state_groups`. Things seem to work just fine if I disconnect the floating state chain from the `insertion` event so we can move forward with that in https://github.com/matrix-org/synapse/pull/11114 after this merges. Both the state chain and the `insertion` event chain can float with no `prev_events`. No issues with `state_groups` or accepting over federation.

synapse/handlers/message.py
MadLittleMods: I'm not sure yet. If you really want me to dive into it, I can ⛳

synapse/handlers/message.py
```diff
@@ -949,14 +953,23 @@ async def create_new_client_event(
...
         # event and then try to auth it (which fails with a somewhat confusing "No
         # create event in auth events")
-        assert (
-            builder.type == EventTypes.Create or len(prev_event_ids) > 0
-        ), "Attempting to create an event with no prev_events"
+        if allow_no_prev_events:
+            # We allow events with no `prev_events` but it better have some `auth_events`
+            assert (
+                builder.type == EventTypes.Create
+                or prev_event_ids
```
erikjohnston, line 963:
```suggestion
```

Personally, since we *do* want them to have auth events

changelog.d/11243.feature
```diff
@@ -0,0 +1 @@
+Allow events to be created without `prev_events` (only `auth_events`).
```
erikjohnston, line 1:
```suggestion
Allow specific, experimental events to be created without `prev_events`. Used by MSCXXXXX
```

Or something? And its probably a misc? Or maybe we actually just want to have a generic `Improve support for MSCXX` changelog so all these PRs go into the same changelog entry. End users/server admins won't care about the guts of implementations like this (which is the audience of these changelogs)

erikjohnston (approved):
This makes sense to me, sorry for long RTT :heart:

MadLittleMods:
Thanks for the review @erikjohnston @anoadragon453 @richvdh 🦎

Happy to see this one get across the line! 🥌

PR #11246, opened 2021-11-04
Additional test for `cachedList`

richvdh:
I was trying to understand how `cachedList` works, and ended up writing this extra test. I figure we may as well keep it.

### DIFF
--- a/tests/util/caches/test_descriptors.py
+++ b/tests/util/caches/test_descriptors.py
@@ -716,7 +716,7 @@ class CachedListDescriptorTestCase(unittest.TestCase):
                 pass
 
             @descriptors.cachedList("fn", "args1")
-            def list_fn(self, args1) -> Deferred[dict]:
+            def list_fn(self, args1) -> "Deferred[dict]":
                 return self.mock(args1)
 
         obj = Cls()
--- a/tests/util/caches/test_descriptors.py
+++ b/tests/util/caches/test_descriptors.py
@@ -17,6 +17,7 @@ from typing import Set
 from unittest import mock
 
 from twisted.internet import defer, reactor
+from twisted.internet.defer import Deferred
 
 from synapse.api.errors import SynapseError
 from synapse.logging.context import (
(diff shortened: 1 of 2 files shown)


tests/util/caches/test_descriptors.py
```diff
@@ -703,6 +704,48 @@ async def list_fn(self, args1, arg2):
...
+    def test_concurrent_lookups(self):
+        """All concurrent lookups should get the same result"""
+
+        class Cls:
+            def __init__(self):
+                self.mock = mock.Mock()
+
+            @descriptors.cached()
+            def fn(self, arg1):
+                pass
```
DMRobertson, line 716: I don't think this is used---or is the `"fn"` on 718 referring to this somehow?

tests/util/caches/test_descriptors.py
```diff
@@ -703,6 +704,48 @@ async def list_fn(self, args1, arg2):
...
+        obj.mock.return_value = deferred_result
+
+        # start off several concurrent lookups of the same key
+        d1 = obj.list_fn([10])
+        d2 = obj.list_fn([10])
+        d3 = obj.list_fn([10])
+
+        # the mock should have been called exactly once
+        obj.mock.assert_called_once_with((10,))
+        obj.mock.reset_mock()
```
DMRobertson, line 733: Is there a reason for the reset? Just tidying up?

DMRobertson (approved):
The test seems reasonable, but `cachedList` seems like spooky magic.

richvdh:
> but cachedList seems like spooky magic.

it's not actually that magic. I can talk you through it if you're interested!

tests/util/caches/test_descriptors.py
richvdh: > is the `"fn"` on 718 referring to this somehow?

yes, it is. an `@cachedList` function is alwaysan extension of an `@cached` function - it doesn't actually have a cache of its own.

So if you don't really care about having a non-list version, you still have to have a stub, just to hold the cache.

There are a few instances of this in the codebase, though there they throw `NotImplementedError`.

tests/util/caches/test_descriptors.py
```diff
@@ -703,6 +704,48 @@ async def list_fn(self, args1, arg2):
...
+        class Cls:
+            def __init__(self):
+                self.mock = mock.Mock()
+
+            @descriptors.cached()
+            def fn(self, arg1):
+                pass
+
+            @descriptors.cachedList("fn", "args1")
+            def list_fn(self, args1) -> Deferred[dict]:
```
DMRobertson, line 719: Oh, CI is failing because the old version of twisted doesn't have generic `Deferred`. I think you want `"Deferred[dict]"` here.

```
2021-11-04T12:47:21.2343360Z ===============================================================================
2021-11-04T12:47:21.2344758Z [ERROR]
2021-11-04T12:47:21.2345162Z Traceback (most recent call last):
2021-11-04T12:47:21.2345884Z   File "/github/workspace/tests/util/caches/test_descriptors.py", line 710, in test_concurrent_lookups
2021-11-04T12:47:21.2346531Z     class Cls:
2021-11-04T12:47:21.2348968Z   File "/github/workspace/tests/util/caches/test_descriptors.py", line 719, in Cls
2021-11-04T12:47:21.2350189Z     def list_fn(self, args1) -> Deferred[dict]:
2021-11-04T12:47:21.2351885Z builtins.TypeError: 'type' object is not subscriptable
2021-11-04T12:47:21.2352322Z 
2021-11-04T12:47:21.2353364Z tests.util.caches.test_descriptors.CachedListDescriptorTestCase.test_concurrent_lookups
2021-11-04T12:47:21.2355276Z -------------------------------------------------------------------------------
```

DMRobertson (changes requested):

tests/util/caches/test_descriptors.py
richvdh: mostly that.

tests/util/caches/test_descriptors.py
richvdh: yes I did. Thanks.

DMRobertson (approved):
lgtm!

PR #11247, opened 2021-11-04
Clarifications and small fixes to to-device related code

anoadragon453:
A lot of little clarifications and some fixes split out from #11215 to help reduce muddying of the new feature.

Most of the comment changes are just helping document what's already there. There are also a couple code changes included:

* Tightening up a type hint that didn't need to be optional.
* Preventing shadowing the `type` built-in by renaming a variable.

Recommended to review commit-by-commit.

### DIFF
--- a/changelog.d/11247.misc
+++ b/changelog.d/11247.misc
@@ -1,1 +1,1 @@
-Clean up some areas of the codebase relating to to-device messages and sending ephemeral events to application services.
+Clean up code relating to to-device messages and sending ephemeral events to application services.
--- a/changelog.d/11247.misc
+++ b/changelog.d/11247.misc
@@ -1,0 +1,1 @@
+Clean up some areas of the codebase relating to to-device messages and sending ephemeral events to application services.
--- a/synapse/handlers/appservice.py
+++ b/synapse/handlers/appservice.py
@@ -262,7 +262,7 @@ class ApplicationServicesHandler:
         new_token: int,
         users: Collection[Union[str, UserID]],
     ) -> None:
-        logger.debug("Checking interested services for %s" % stream_key)
+        logger.debug("Checking interested services for %s", stream_key)
         with Measure(self.clock, "notify_interested_services_ephemeral"):
             for service in services:
                 if stream_key == "typing_key":
--- a/synapse/handlers/appservice.py
+++ b/synapse/handlers/appservice.py
@@ -203,7 +203,7 @@ class ApplicationServicesHandler:
                 value for `stream_key` will cause this function to return early.
 
                 Ephemeral events will only be pushed to appservices that have opted into
-                them.
+                them via the ephemeral AS registration file option.
 
                 Appservices will only receive ephemeral events that fall within their
                 registered user and room namespaces.
@@ -214,6 +214,7 @@ class ApplicationServicesHandler:
         if not self.notify_appservices:
             return
 
+        # Ignore any unsupported streams
         if stream_key not in ("typing_key", "receipt_key", "presence_key"):
             return
 
@@ -230,12 +231,19 @@ class ApplicationServicesHandler:
         # Additional context: https://github.com/matrix-org/synapse/pull/11137
         assert isinstance(new_token, int)
 
+        # Check whether there are any appservices which have registered to receive
+        # ephemeral events.
+        #
+        # Note that whether these events are actually relevant to these appservices
+        # is decided later on.
         services = [
             service
             for service in self.store.get_app_services()
             if service.supports_ephemeral
         ]
         if not services:
+            # Bail out early if none of the target appservices have explicitly registered
+            # to receive these ephemeral events.
             return
 
         # We only start a new background process if necessary rather than
@@ -252,7 +260,7 @@ class ApplicationServicesHandler:
         new_token: int,
         users: Collection[Union[str, UserID]],
     ) -> None:
-        logger.debug("Checking interested services for %s" % (stream_key))
+        logger.debug("Checking interested services for %s" % stream_key)
         with Measure(self.clock, "notify_interested_services_ephemeral"):
             for service in services:
                 if stream_key == "typing_key":
@@ -345,6 +353,9 @@ class ApplicationServicesHandler:
 
         Args:
             service: The application service to check for which events it should receive.
+            new_token: A receipts event stream token. Purely used to double-check that the
+                from_token we pull from the database isn't greater than or equal to this
+                token. Prevents accidentally duplicating work.
 
         Returns:
             A list of JSON dictionaries containing data derived from the read receipts that
@@ -382,6 +393,9 @@ class ApplicationServicesHandler:
         Args:
             service: The application service that ephemeral events are being sent to.
             users: The users that should receive the presence update.
+            new_token: A presence update stream token. Purely used to double-check that the
+                from_token we pull from the database isn't greater than or equal to this
+                token. Prevents accidentally duplicating work.
 
         Returns:
             A list of json dictionaries containing data derived from the presence events
--- a/synapse/handlers/appservice.py
+++ b/synapse/handlers/appservice.py
@@ -203,7 +203,9 @@ class ApplicationServicesHandler:
                 value for `stream_key` will cause this function to return early.
 
                 Ephemeral events will only be pushed to appservices that have opted into
-                them via the ephemeral AS registration file option.
+                receiving them by setting `push_ephemeral` to true in their registration
+                file. Note that while MSC2409 is experimental, this option is called
+                `de.sorunome.msc2409.push_ephemeral`.
 
                 Appservices will only receive ephemeral events that fall within their
                 registered user and room namespaces.
--- a/synapse/handlers/appservice.py
+++ b/synapse/handlers/appservice.py
@@ -188,7 +188,7 @@ class ApplicationServicesHandler:
         self,
         stream_key: str,
         new_token: Union[int, RoomStreamToken],
-        users: Optional[Collection[Union[str, UserID]]] = None,
+        users: Collection[Union[str, UserID]],
     ) -> None:
         """
         This is called by the notifier in the background when an ephemeral event is handled
@@ -249,7 +249,7 @@ class ApplicationServicesHandler:
         # We only start a new background process if necessary rather than
         # optimistically (to cut down on overhead).
         self._notify_interested_services_ephemeral(
-            services, stream_key, new_token, users or []
+            services, stream_key, new_token, users
         )
 
     @wrap_as_background_process("notify_interested_services_ephemeral")
--- a/synapse/handlers/devicemessage.py
+++ b/synapse/handlers/devicemessage.py
@@ -144,14 +144,14 @@ class DeviceMessageHandler:
 
         # Add messages to the database.
         # Retrieve the stream id of the last-processed to-device message.
-        max_stream_id = await self.store.add_messages_from_remote_to_device_inbox(
+        last_stream_id = await self.store.add_messages_from_remote_to_device_inbox(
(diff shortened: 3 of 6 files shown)


synapse/handlers/devicemessage.py
```diff
@@ -135,12 +142,16 @@ async def on_direct_to_device_edu(self, origin: str, content: JsonDict) -> None:
                 message_type, sender_user_id, by_device
             )
 
-        stream_id = await self.store.add_messages_from_remote_to_device_inbox(
+        # Add messages to the database.
+        # Retrieve the stream id of the last-processed to-device message.
+        max_stream_id = await self.store.add_messages_from_remote_to_device_inbox(
```
DMRobertson, line 147: What does the `max` indicate here? Largest stream id that has been processed?

synapse/handlers/appservice.py
```diff
@@ -203,7 +203,7 @@ def notify_interested_services_ephemeral(
                 value for `stream_key` will cause this function to return early.
 
                 Ephemeral events will only be pushed to appservices that have opted into
-                them.
+                them via the ephemeral AS registration file option.
```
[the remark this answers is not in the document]
anoadragon453: "(ephemeral AS registration file) option" is right.

Application services register themselves with a homeserver through a [registration file](https://matrix.org/docs/spec/application_service/r0.1.2#registration). That file is filled with options, for which sending ephemeral events is an opt-in one (https://github.com/Sorunome/matrix-doc/blob/soru%2Bhs/appservice-edus/proposals/2409-appservice-edus.md#changes-to-the-registration-file).

I did feel this was a little convoluted when writing it, so thanks for pointing it out. What do you think about:

> Ephemeral events will only be pushed to appservices that have opted into receiving them by setting `push_ephemeral` to true in their registration file. Note that while MSC2409 is experimental, this option is called `de.sorunome.msc2409.push_ephemeral`.

? Bit wordy but comments are free!

synapse/handlers/devicemessage.py
anoadragon453: Yes, it's just the stream id of the last to-device message that was added to the database. Messages are added to the database in with incrementing stream ids.

```
# Retrieve the stream id of the to-device message that was stored last (the largest stream id).
```

instead perhaps?

synapse/handlers/appservice.py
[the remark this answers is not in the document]
DMRobertson: Bit wordy, but not that big and it's much clearer to me. :+1:

synapse/handlers/devicemessage.py
DMRobertson: I guess my concern here was that "processing" could be a separate step to persisting them in the database. I also wasn't sure if we might be adding messages with stream ids say `10, 11, 12`, but someone else has already put a message with a stream id of `30` into the db.

anoadragon453: The stream id that is assigned to each to-device message is implemented using an ever-incrementing database sequence:

https://github.com/matrix-org/synapse/blob/4535532526581834ab798996ffe73f6d19c25123/synapse/storage/databases/main/deviceinbox.py#L468-L475

So having multiple rows with the same stream id should never occur.

to-device messages are stored here. Later on they are retrieved from the database and sent off to remote homeservers/down user sync streams. Each remote has a last-known-received stream id, and the same is encoded in client sync tokens.

...does that quell your concern?

synapse/handlers/appservice.py
```diff
@@ -252,7 +262,7 @@ async def _notify_interested_services_ephemeral(
         new_token: int,
         users: Collection[Union[str, UserID]],
     ) -> None:
-        logger.debug("Checking interested services for %s" % (stream_key))
+        logger.debug("Checking interested services for %s" % stream_key)
```
clokep, line 265: I think you wanted this?

```suggestion
        logger.debug("Checking interested services for %s", stream_key)
```

synapse/handlers/devicemessage.py
clokep: `last_stream_id` might be a better name in this case?

synapse/storage/databases/main/deviceinbox.py
```diff
@@ -153,12 +156,21 @@ def get_new_messages_for_device_txn(txn):
             txn.execute(
                 sql, (user_id, device_id, last_stream_id, current_stream_id, limit)
             )
+
             messages = []
+            stream_pos = current_stream_id
```
clokep, line 161: Why is this added? In case `txn` is empty?

synapse/storage/databases/main/deviceinbox.py
```diff
@@ -153,12 +156,21 @@ def get_new_messages_for_device_txn(txn):
...
+
             for row in txn:
                 stream_pos = row[0]
                 messages.append(db_to_json(row[1]))
+
+            # If we don't end up hitting the limit, we still want to return the equivalent
+            # value of `current_stream_id` to the calling function. This is needed as we'll
+            # be processing up to `current_stream_id`, without necessarily fetching a message
+            # with a stream token of `current_stream_id`.
```
clokep, line 170: I'm not really following this, wouldn't this cause us to "skip" rows or something?

synapse/storage/databases/main/deviceinbox.py
```diff
@@ -261,12 +273,19 @@ def get_new_messages_for_remote_destination_txn(txn):
             )
             txn.execute(sql, (destination, last_stream_id, current_stream_id, limit))
             messages = []
```
clokep, line 275: Do we need to instantiate `stream_pos` like we do above?

It's a shame that this code looks identical minus the table name.

clokep (approved):

synapse/handlers/appservice.py
anoadragon453: Oh! Yes, that's a better shout.

synapse/storage/databases/main/deviceinbox.py
anoadragon453: It's mostly just to stop my editor complaining that `stream_pos` is never set, though we should never get into a state where `stream_pos` is `None`. We'll either pull something from the database, or we won't have and then `len(messages) < limit`...

Well, unless `limit` is 0, but that'd be unlikely to be set.

So yes, just a peace of mind safeguard really. I can add a comment above it explaining what we're doing with `stream_pos` overall.

synapse/storage/databases/main/deviceinbox.py
anoadragon453: Oh! I didn't notice that until now. Refactoring it is definitely an option.

synapse/storage/databases/main/deviceinbox.py
anoadragon453: We shouldn't, though I see that the explanation isn't clear.

For example, let's say `user_id` is `@alice:localhost`, `last_stream_id` is 1, and `current_stream_id` is 5, limit is 3, and this is what the DB looks like:

```
      user_id      |                  device_id                  | stream_id |     message_json   | instance_name 
-------------------+---------------------------------------------+-----------+--------------------+---------------
 @alice:localhost  | HAKDUDVZXZ                                  |         1 | {"content":{"..."} | master
 @alice:localhost  | HAKDUDVZXZ                                  |         2 | {"content":{"..."} | master
 @alice:localhost  | HAKDUDVZXZ                                  |         3 | {"content":{"..."} | master
 @alice:localhost  | HAKDUDVZXZ                                  |         4 | {"content":{"..."} | master
 @bob:localhost    | TYCGLKNPVJ                                  |         5 | {"content":{"..."} | master
(5 rows)
```

We're going to pull out to-device messages with stream ids 2, 3, 4. `len(messages) == limit`, so `stream_pos` is 4. We return `4` alongside the messages, and `4` is stored.

Later on, a new to-device message comes in, so the DB now looks like:

```
      user_id      |                  device_id                  | stream_id |     message_json   | instance_name 
-------------------+---------------------------------------------+-----------+--------------------+---------------
 @alice:localhost  | HAKDUDVZXZ                                  |         1 | {"content":{"..."} | master
 @alice:localhost  | HAKDUDVZXZ                                  |         2 | {"content":{"..."} | master
 @alice:localhost  | HAKDUDVZXZ                                  |         3 | {"content":{"..."} | master
 @alice:localhost  | HAKDUDVZXZ                                  |         4 | {"content":{"..."} | master
 @bob:localhost    | TYCGLKNPVJ                                  |         5 | {"content":{"..."} | master
 @bob:localhost    | TYCGLKNPVJ                                  |         6 | {"content":{"..."} | master
(6 rows)
```

And we call this function with `@alice:localhost`, `last_stream_id` as 4, and `current_stream_id` as 6. Limit is still 3.

We end up not pulling out any messages, as none of them are addressed to `@alice:localhost`. However, we still need to return `6` for `stream_pos`, so that the calling function knows where to check from on the next round. If we returned `4`, then we'd duplicate work checking stream ids 5, 6 again.

---

Thinking more about it now... there is an edge case where we end up returning exactly `<limit>` rows. Say you had the below DB table: 

```
      user_id      |                  device_id                  | stream_id |     message_json   | instance_name 
-------------------+---------------------------------------------+-----------+--------------------+---------------
 @alice:localhost  | HAKDUDVZXZ                                  |         1 | {"content":{"..."} | master
 @alice:localhost  | HAKDUDVZXZ                                  |         2 | {"content":{"..."} | master
 @alice:localhost  | HAKDUDVZXZ                                  |         3 | {"content":{"..."} | master
 @alice:localhost  | HAKDUDVZXZ                                  |         4 | {"content":{"..."} | master
 @bob:localhost    | TYCGLKNPVJ                                  |         5 | {"content":{"..."} | master
(5 rows)
```

And you called `get_new_messages_for_device` with user_id `@alice:localhost`, `last_stream_id` as 1, `current_stream_id` as 5, with limit 3.

The function would check over rows 2 - 5 in the DB, and return rows 2, 3, 4. `len(messages) == limit`, so `stream_pos` is that of the last row we pulled out: 4. However, the function did check row 5 already. Next time round the function will be called with `last_stream_id` as 4. So we'll end up doing some duplicate work checking row 5 again.

I think this only causes a slight drop in perf though, rather than going the other way and causing us to miss rows...

synapse/storage/databases/main/deviceinbox.py
clokep: Yep, refactoring is more of a long term thing, just noticed it while reviewing. (I didn't closely inspect if they *are* identical by the way. They might just be very similar.)

changelog.d/11247.misc
```diff
@@ -0,0 +1 @@
+Clean up some areas of the codebase relating to to-device messages and sending ephemeral events to application services.
```
clokep, line 1:
```suggestion
Clean up code relating to to-device messages and sending ephemeral events to application services.
```

synapse/storage/databases/main/deviceinbox.py
clokep: After chatting about this it seems that the idea is that if we don't hit the limit that means there's no data left (up to the `current_stream_id`) that relates to the requested filtering (by user/device). Thus it is safe to consider up to `current_stream_id` processed (instead considering only up to the last result's stream position).

I think the following simpler comment describes this?

```suggestion
            # If the limit was not reached we know that there's no more data for this
            # user/device pair up to current_stream_id.
```

anoadragon453: I'd say that's accurate!

anoadragon453: Fixed in a25c8dd4c55f79a20b5046c32329a4e27af705aa.

clokep (approved):

PR #11255, opened 2021-11-04
Fix rolling back when using workers

### LINKED ISSUE #11252: Rolling back is broken on workers when there is a schema version change
```
**********************************************************************************
 Error during initialisation:
    Expected database schema version 64 but got 65: run the main synapse process to upgrade the database schema before starting worker processes.
 There may be more information in the logs.
**********************************************************************************
```

erikjohnston:
Fixes #11252

### DIFF
--- a/synapse/storage/prepare_database.py
+++ b/synapse/storage/prepare_database.py
@@ -142,6 +142,9 @@ def prepare_database(
                     databases=databases,
                 )
             elif version_info.current_version < SCHEMA_VERSION:
+                # If the DB is on an older version than we expect the we refuse
+                # to start the worker (as the main process needs to run first to
+                # update the schema).
                 raise UpgradeDatabaseException(
                     OUTDATED_SCHEMA_ON_WORKER_ERROR
                     % (SCHEMA_VERSION, version_info.current_version)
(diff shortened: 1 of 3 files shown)


synapse/storage/prepare_database.py
```diff
@@ -133,22 +133,20 @@ def prepare_database(
...
-            ):
+            if config.worker.worker_app is None:
+                _upgrade_existing_database(
+                    cur,
+                    version_info,
+                    database_engine,
+                    config,
+                    databases=databases,
+                )
+            elif version_info.current_version < SCHEMA_VERSION:
```
clokep, line 144: The change from `!=` to `<` is the real fix here, I believe?

Is it worth adding a comment saying "If the database hasn't been updated yet, refuse to start" or something?

clokep (approved):
Seems reasonable. I had to work out why we didn't use `SCHEMA_COMPAT_VERSION` here, but this is because we don't want the workers to expect a *newer* version than main (since that would mean that the migration hasn't run yet).

clokep:
Looks like tests are failing here.

synapse/storage/prepare_database.py
richvdh: that's pretty much what the comment at line 134 is saying, though I kindof agree that it might be clearer to move it down now.

PR #11262, opened 2021-11-05
Handle federation inbound instances being killed more gracefully

erikjohnston:
If a worker handling inbound federation dies and gets restarted then the locks held when it got killed will still be considered valid when it starts up again (as the timeouts won't have expired yet). So when we check for staged events in the DB at start up we won't process any of those where the lock for the room was previously held, and then we won't check again until we get another event for that room or we restart, causing those events to get stuck even though the locks will time out.

We fix this in two ways:

1. We ignore locks that match our instance name that we don't actually hold; and
2. Add a periodic check for old staged events.

### DIFF
--- a/synapse/storage/databases/main/lock.py
+++ b/synapse/storage/databases/main/lock.py
@@ -14,6 +14,7 @@
 import logging
 from types import TracebackType
 from typing import TYPE_CHECKING, Dict, Optional, Tuple, Type
+from weakref import WeakValueDictionary
 
 from twisted.internet.interfaces import IReactorCore
 
@@ -61,7 +62,7 @@ class LockStore(SQLBaseStore):
 
         # A map from `(lock_name, lock_key)` to the token of any locks that we
         # think we currently hold.
-        self._live_tokens: Dict[Tuple[str, str], str] = {}
+        self._live_tokens: Dict[Tuple[str, str], Lock] = WeakValueDictionary()
 
         # When we shut down we want to remove the locks. Technically this can
         # lead to a race, as we may drop the lock while we are still processing.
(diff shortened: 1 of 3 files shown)


synapse/storage/databases/main/lock.py
```diff
@@ -112,6 +120,7 @@ def _try_acquire_lock_txn(txn: LoggingTransaction) -> bool:
                             last_renewed_ts = EXCLUDED.last_renewed_ts
                         WHERE
                             worker_locks.last_renewed_ts < ?
+                            OR worker_locks.instance_name = EXCLUDED.instance_name
```
DMRobertson, line 123: Does the `DO UPDATE` above only apply to the row(s?) that's caused a conflict? If not I'm not sure I understand what's happening here.

DMRobertson (approved):
I think this relies on each worker having their own unique instance name. But each instance name corresponds(?) to a different worker configuration filename, so probably they're unique?

I guess there's no way to enforce this, because no synapse process knows about any of the other synapses.

Is there any way to test this ahead of deployment?

synapse/storage/databases/main/lock.py
erikjohnston: Yup, so it's basically equivalent to doing the `INSERT` if there is no conflict, or applying the `UPDATE SET ... WHERE ...` if there is a conflict (with the special `EXCLUDED.*` table matching the `VALUES` in the `INSERT`).  

I.e. in this specific case we're either inserting a new row, or updating the existing row so long as either a) the lock has timed out OR b) the instance name matches.

erikjohnston:
> I think this relies on each worker having their own unique instance name. But each instance name corresponds(?) to a different worker configuration filename, so probably they're unique?

They're really meant to be unique, but it's hard to check.

> I guess there's no way to enforce this, because no synapse process knows about any of the other synapses.

Exactly.

> Is there any way to test this ahead of deployment?

Not beyond the unit tests, I don't think. 

It's worth noting that the locks are *best effort* (since they time out), so nothing _should_ break if multiple instances process the same event, it's just inefficient.

PR #11263, opened 2021-11-05
Add some background update admin APIs

### LINKED ISSUE #11259: Add admin API for pausing background updates
If a background update is causing issues it is useful to be able to temporarily pause running of the updates via an admin API.

Open question: should this only pause the BG updates until next restart? We want to avoid the situation where the BG updates get paused and then forgotten about, as that would be a massive footgun

erikjohnston:
Fixes #11259


dklimpel:
IMO the location of new docs should be "docs/usage/administration/admin_api"
https://github.com/matrix-org/synapse/tree/develop/docs/admin_api#readme

And the summary needs a new entry
https://github.com/matrix-org/synapse/blob/develop/docs/SUMMARY.md

docs/usage/administration/admin_api/background_updates.md
```diff
@@ -0,0 +1,84 @@
...
+}
+```
+
+`enabled` sets whether the background updates are enabled or disabled.
+
+The API returns the `enabled` param.
+
+```json
+{
+    "enabled": 0
```
squahtx, line 80:
```suggestion
    "enabled": false
```

synapse/storage/background_updates.py
```diff
@@ -100,29 +100,58 @@ def __init__(self, hs: "HomeServer", database: "DatabasePool"):
...
+
+        update_name = self._current_background_update
+        if not update_name:
+            return None
+
+        perf = self._background_update_performance.get(update_name)
+        if not perf:
+            perf = BackgroundUpdatePerformance(update_name)
+
+        return self._background_update_performance.get(update_name)
```
squahtx, line 121:
```suggestion
        return perf
```

synapse/storage/database.py
```diff
@@ -446,6 +446,10 @@ def __init__(
                 self._check_safe_to_upsert,
             )
 
+    def name(self) -> str:
+        "Return the name of this database"
+        return self._database_config.name
```
squahtx, line 451: This could be a `@property`

erikjohnston: Yeah, though I'm not a huge fan of using properties for the sake of it :shrug:

Issue #11241, opened 2021-11-03
Historical state events from `state_events_at_start` in the MSC2716 `/batch_send` endpoint get pushed to an application service `/transactions`


MadLittleMods:
[As discovered by @tulir](https://matrix.to/#/!SBYNQlpqkwJzFIdzxI:nevarro.space/$EyGLPHcogIusgKqMjLkWKVC6MOJpFtm-PFfOhtgrqG8?via=nevarro.space&via=beeper.com&via=matrix.org), when using the MSC2716 `/batch_send` endpoint, events specified in `state_events_at_start` will come down ~~incremental `/sync`~~ and the application services `/transactions` endpoints. Since this is historical state, it shouldn't be coming down the live pipe.

We already have a Complement [tests to ensure the historical messages don't come down incremental sync](https://github.com/matrix-org/complement/blob/11adc36c58c29c0603ded8a6276563276d2d5649/tests/msc2716_test.go#L258-L303), but we should also do the same for the state events.

### Relevant code

 - Where we create state events from the `state_events_at_start`: [`synapse/handlers/room_batch.py#L213-L249`](https://github.com/matrix-org/synapse/blob/a271e233e9f846193c22b6d74f33ae7d7f2c1167/synapse/handlers/room_batch.py#L213-L249)
 - `builder.internal_metadata.historical = True`: [`synapse/handlers/message.py#L608`](https://github.com/matrix-org/synapse/blob/a271e233e9f846193c22b6d74f33ae7d7f2c1167/synapse/handlers/message.py#L608)
 - If it's a historical event, then we mark it as `backfilled = True`: [`synapse/handlers/message.py#L1526-L1530`](https://github.com/matrix-org/synapse/blob/a271e233e9f846193c22b6d74f33ae7d7f2c1167/synapse/handlers/message.py#L1526-L1530)

MadLittleMods:
From my testing, the historical state does not come down `/sync` thanks to it being an `outlier` already. But we should make sure it's also `historical` in the internal metadata.

Need to dig into the application service being pushed historical state events on `/transaction` by the homeserver

PR #11265, opened 2021-11-05
Prevent historical state from being pushed to an application service via `/transactions` (MSC2716)

MadLittleMods:
Prevent historical state from being pushed to an application service via `/transactions`. We now mark historical state from the MSC2716 `/batch_send` endpoint as `historical` which makes it `backfilled` and have a negative `stream_ordering` so it doesn't get queried by `/transactions`.

Fix https://github.com/matrix-org/synapse/issues/11241

Complement tests: https://github.com/matrix-org/complement/pull/221

### Dev notes

```
update_membership
update_membership_locked
_local_membership_update
create_event

create_event
```

`/sync` looks for `stream_ordering` but excludes all `outliers`.

https://github.com/matrix-org/synapse/blob/b09d90cac9179f84024d4cb3ab4574480c1fd1df/synapse/storage/databases/main/stream.py#L519-L527

Whereas `/transactions` only cares about `stream_ordering`. Perhaps we should exclude `outliers` in `/transactions`?

https://github.com/matrix-org/synapse/blob/b09d90cac9179f84024d4cb3ab4574480c1fd1df/synapse/storage/databases/main/appservice.py#L358-L368

 
## `/sync` stack trace

 - [`SyncRestServlet.on_GET`](https://github.com/matrix-org/synapse/blob/b09d90cac9179f84024d4cb3ab4574480c1fd1df/synapse/rest/client/sync.py#L113)
 - [`wait_for_sync_for_user`](https://github.com/matrix-org/synapse/blob/b09d90cac9179f84024d4cb3ab4574480c1fd1df/synapse/handlers/sync.py#L299)
 - [`_wait_for_sync_for_user`](https://github.com/matrix-org/synapse/blob/b09d90cac9179f84024d4cb3ab4574480c1fd1df/synapse/handlers/sync.py#L329)
 - [`current_sync_for_user`](https://github.com/matrix-org/synapse/blob/b09d90cac9179f84024d4cb3ab4574480c1fd1df/synapse/handlers/sync.py#L399)
 - [`generate_sync_result`](https://github.com/matrix-org/synapse/blob/b09d90cac9179f84024d4cb3ab4574480c1fd1df/synapse/handlers/sync.py#L1039)
 - [`_generate_sync_entry_for_rooms`](https://github.com/matrix-org/synapse/blob/b09d90cac9179f84024d4cb3ab4574480c1fd1df/synapse/handlers/sync.py#L1455)
 - [`_get_rooms_changed`](https://github.com/matrix-org/synapse/blob/b09d90cac9179f84024d4cb3ab4574480c1fd1df/synapse/handlers/sync.py#L1614)
 - [`get_room_events_stream_for_rooms`](https://github.com/matrix-org/synapse/blob/b09d90cac9179f84024d4cb3ab4574480c1fd1df/synapse/storage/databases/main/stream.py#L409)
 - [`get_room_events_stream_for_room`](https://github.com/matrix-org/synapse/blob/b09d90cac9179f84024d4cb3ab4574480c1fd1df/synapse/storage/databases/main/stream.py#L478-L527)

## `/transactions` stack trace

 - [`notify_interested_services`](https://github.com/matrix-org/synapse/blob/b09d90cac9179f84024d4cb3ab4574480c1fd1df/synapse/handlers/appservice.py#L66-L87)
 - [`_notify_interested_services`](https://github.com/matrix-org/synapse/blob/b09d90cac9179f84024d4cb3ab4574480c1fd1df/synapse/handlers/appservice.py#L90-L135)
    - [`get_new_events_for_appservice`](https://github.com/matrix-org/synapse/blob/b09d90cac9179f84024d4cb3ab4574480c1fd1df/synapse/storage/databases/main/appservice.py#L353-L368)
    - [`submit_event_for_as`](https://github.com/matrix-org/synapse/blob/b09d90cac9179f84024d4cb3ab4574480c1fd1df/synapse/handlers/appservice.py#L133-L135)

[`submit_event_for_as`](https://github.com/matrix-org/synapse/blob/b09d90cac9179f84024d4cb3ab4574480c1fd1df/synapse/appservice/scheduler.py#L94-L95) ->

 - [`enqueue_event`](https://github.com/matrix-org/synapse/blob/b09d90cac9179f84024d4cb3ab4574480c1fd1df/synapse/appservice/scheduler.py#L129-L131)
 - [`_send_request`](https://github.com/matrix-org/synapse/blob/b09d90cac9179f84024d4cb3ab4574480c1fd1df/synapse/appservice/scheduler.py#L137-L157)
 - `appservice.scheduler.send`
 - [`create_appservice_txn`](https://github.com/matrix-org/synapse/blob/b09d90cac9179f84024d4cb3ab4574480c1fd1df/synapse/appservice/scheduler.py#L196-L201)
 - [`AppServiceTransaction.send`](https://github.com/matrix-org/synapse/blob/b09d90cac9179f84024d4cb3ab4574480c1fd1df/synapse/appservice/__init__.py#L338)
 - [`push_bulk`](https://github.com/matrix-org/synapse/blob/b09d90cac9179f84024d4cb3ab4574480c1fd1df/synapse/appservice/api.py#L202)
 - [`/transactions`](https://github.com/matrix-org/synapse/blob/b09d90cac9179f84024d4cb3ab4574480c1fd1df/synapse/appservice/api.py#L220)
 
 

### Pull Request Checklist

### DIFF
--- a/changelog.d/11265.bugfix
+++ b/changelog.d/11265.bugfix
@@ -1,0 +1,1 @@
+Prevent [MSC2716](https://github.com/matrix-org/matrix-doc/pull/2716) historical state events from being pushed to an application service via `/transactions`.
--- a/synapse/appservice/api.py
+++ b/synapse/appservice/api.py
@@ -231,17 +231,22 @@ class ApplicationServiceApi(SimpleHttpClient):
                 json_body=body,
                 args={"access_token": service.hs_token},
             )
-            logger.debug(
-                "push_bulk to %s succeeded! events=%s",
-                uri,
-                [event.get("event_id") for event in events],
-            )
+            if logger.isEnabledFor(logging.DEBUG):
+                logger.debug(
+                    "push_bulk to %s succeeded! events=%s",
+                    uri,
+                    [event.get("event_id") for event in events],
+                )
             sent_transactions_counter.labels(service.id).inc()
             sent_events_counter.labels(service.id).inc(len(events))
             return True
         except CodeMessageException as e:
             logger.warning(
-                "push_bulk to %s received code=%s msg=%s", uri, e.code, e.msg
+                "push_bulk to %s received code=%s msg=%s",
+                uri,
+                e.code,
+                e.msg,
+                exc_info=logger.isEnabledFor(logging.DEBUG),
             )
         except Exception as ex:
             logger.warning(
@@ -250,6 +255,7 @@ class ApplicationServiceApi(SimpleHttpClient):
                 type(ex).__name__,
                 ex,
                 ex.args,
+                exc_info=logger.isEnabledFor(logging.DEBUG),
             )
         failed_transactions_counter.labels(service.id).inc()
         return False
--- a/synapse/appservice/api.py
+++ b/synapse/appservice/api.py
@@ -231,13 +231,26 @@ class ApplicationServiceApi(SimpleHttpClient):
                 json_body=body,
                 args={"access_token": service.hs_token},
             )
+            logger.debug(
+                "push_bulk to %s succeeded! events=%s",
+                uri,
+                [event.get("event_id") for event in events],
+            )
             sent_transactions_counter.labels(service.id).inc()
             sent_events_counter.labels(service.id).inc(len(events))
             return True
         except CodeMessageException as e:
-            logger.warning("push_bulk to %s received %s", uri, e.code)
+            logger.warning(
+                "push_bulk to %s received code=%s msg=%s", uri, e.code, e.msg
+            )
         except Exception as ex:
-            logger.warning("push_bulk to %s threw exception %s", uri, ex)
+            logger.warning(
+                "push_bulk to %s threw exception(%s) %s args=%s",
+                uri,
+                type(ex).__name__,
+                ex,
+                ex.args,
+            )
         failed_transactions_counter.labels(service.id).inc()
         return False
 
--- a/synapse/handlers/room_batch.py
+++ b/synapse/handlers/room_batch.py
@@ -221,6 +221,7 @@ class RoomBatchHandler:
                     action=membership,
                     content=event_dict["content"],
                     outlier=True,
+                    historical=True,
                     prev_event_ids=[prev_event_id_for_state_chain],
                     # Make sure to use a copy of this list because we modify it
                     # later in the loop here. Otherwise it will be the same
@@ -240,6 +241,7 @@ class RoomBatchHandler:
                     ),
                     event_dict,
                     outlier=True,
+                    historical=True,
                     prev_event_ids=[prev_event_id_for_state_chain],
                     # Make sure to use a copy of this list because we modify it
                     # later in the loop here. Otherwise it will be the same
--- a/synapse/handlers/room_member.py
+++ b/synapse/handlers/room_member.py
@@ -268,6 +268,7 @@ class RoomMemberHandler(metaclass=abc.ABCMeta):
         content: Optional[dict] = None,
         require_consent: bool = True,
         outlier: bool = False,
+        historical: bool = False,
     ) -> Tuple[str, int]:
         """
         Internal membership update function to get an existing event or create
@@ -337,6 +338,7 @@ class RoomMemberHandler(metaclass=abc.ABCMeta):
             auth_event_ids=auth_event_ids,
             require_consent=require_consent,
             outlier=outlier,
+            historical=historical,
         )
 
         prev_state_ids = await context.get_prev_state_ids()
@@ -433,6 +435,7 @@ class RoomMemberHandler(metaclass=abc.ABCMeta):
         new_room: bool = False,
         require_consent: bool = True,
         outlier: bool = False,
+        historical: bool = False,
         prev_event_ids: Optional[List[str]] = None,
         auth_event_ids: Optional[List[str]] = None,
     ) -> Tuple[str, int]:
@@ -487,6 +490,7 @@ class RoomMemberHandler(metaclass=abc.ABCMeta):
                 new_room=new_room,
                 require_consent=require_consent,
                 outlier=outlier,
+                historical=historical,
                 prev_event_ids=prev_event_ids,
                 auth_event_ids=auth_event_ids,
             )
@@ -507,6 +511,7 @@ class RoomMemberHandler(metaclass=abc.ABCMeta):
         new_room: bool = False,
         require_consent: bool = True,
         outlier: bool = False,
+        historical: bool = False,
         prev_event_ids: Optional[List[str]] = None,
         auth_event_ids: Optional[List[str]] = None,
     ) -> Tuple[str, int]:
@@ -657,6 +662,7 @@ class RoomMemberHandler(metaclass=abc.ABCMeta):
                 content=content,
                 require_consent=require_consent,
                 outlier=outlier,
+                historical=historical,
             )
 
         latest_event_ids = await self.store.get_prev_events_for_room(room_id)
--- a/synapse/handlers/room_member.py
+++ b/synapse/handlers/room_member.py
@@ -294,6 +294,9 @@ class RoomMemberHandler(metaclass=abc.ABCMeta):
             outlier: Indicates whether the event is an `outlier`, i.e. if
                 it's from an arbitrary point and floating in the DAG as
                 opposed to being inline with the current DAG.
+            historical: Indicates whether the message is being inserted
+                back in time around some existing events. This is used to skip
+                a few checks and mark the event as backfilled.
 
         Returns:
             Tuple of event ID and stream ordering position
@@ -457,6 +460,9 @@ class RoomMemberHandler(metaclass=abc.ABCMeta):
             outlier: Indicates whether the event is an `outlier`, i.e. if
                 it's from an arbitrary point and floating in the DAG as
                 opposed to being inline with the current DAG.
+            historical: Indicates whether the message is being inserted
+                back in time around some existing events. This is used to skip
+                a few checks and mark the event as backfilled.
             prev_event_ids: The event IDs to use as the prev events
             auth_event_ids:
                 The event ids to use as the auth_events for the new event.
@@ -535,6 +541,9 @@ class RoomMemberHandler(metaclass=abc.ABCMeta):
             outlier: Indicates whether the event is an `outlier`, i.e. if
                 it's from an arbitrary point and floating in the DAG as
                 opposed to being inline with the current DAG.
+            historical: Indicates whether the message is being inserted
+                back in time around some existing events. This is used to skip
+                a few checks and mark the event as backfilled.
             prev_event_ids: The event IDs to use as the prev events
             auth_event_ids:
                 The event ids to use as the auth_events for the new event.


synapse/appservice/api.py
```diff
@@ -231,13 +231,26 @@ async def push_bulk(
...
+            )
         except Exception as ex:
-            logger.warning("push_bulk to %s threw exception %s", uri, ex)
+            logger.warning(
+                "push_bulk to %s threw exception(%s) %s args=%s",
+                uri,
+                type(ex).__name__,
+                ex,
+                ex.args,
+            )
```
MadLittleMods, line 253: Without these extra params, it's pretty much impossible to debug errors like:

```
push_bulk to http://host.docker.internal:9111/transactions/1 threw exception 400 Bad Request

push_bulk to http://localhost:9000/transactions/1 threw exception An error occurred while connecting: 99: Cannot assign requested address.

push_bulk to http://0.0.0.0:9111/transactions/1 threw exception Connection was refused by other side: 111: Connection refused.

push_bulk to http://host.docker.internal:9111/transactions/1 received 504
```

When there is more info available like:
```
push_bulk to http://host.docker.internal:9111/transactions/1 threw PartialDownloadError code=b'400' msg=b'Bad Request' response=b'Client sent an HTTP request to an HTTPS server.\n'

push_bulk to http://localhost:9111/transactions/1 threw exception(ConnectError) An error occurred while connecting: 99: Cannot assign requested address. args=('Cannot assign requested address',)

push_bulk to https://host.docker.internal:9111/transactions/1 threw exception(RequestTransmissionFailed) [<twisted.python.failure.Failure OpenSSL.SSL.Error: [('SSL routines', 'tls_process_server_certificate', 'certificate verify failed')]>]
```

synapse/handlers/room_batch.py
```diff
@@ -221,6 +221,7 @@ async def persist_state_events_at_start(
                     action=membership,
                     content=event_dict["content"],
                     outlier=True,
+                    historical=True,
```
MadLittleMods, line 224: All of these changes are just piping `historical` down to `create_event`

```
update_membership
update_membership_locked
_local_membership_update
create_event
```

synapse/handlers/room_batch.py
```diff
@@ -240,6 +241,7 @@ async def persist_state_events_at_start(
                     ),
                     event_dict,
                     outlier=True,
+                    historical=True,
```
MadLittleMods, line 244: All of these changes are just piping `historical` down to `create_event`

```
create_event
```

synapse/appservice/api.py
erikjohnston: An alternate approach here is to do:

```python
logger.warning(
    "push_bulk to %s threw exception %s", 
    uri, 
    ex,
    exc_info=logger.isEnabledFor(logging.DEBUG),
)
```

which should include exception info if the logging level is set to debug. Your method also works

synapse/appservice/api.py
```diff
@@ -231,13 +231,26 @@ async def push_bulk(
                 json_body=body,
                 args={"access_token": service.hs_token},
             )
+            logger.debug(
+                "push_bulk to %s succeeded! events=%s",
+                uri,
+                [event.get("event_id") for event in events],
+            )
```
erikjohnston, line 238:
```suggestion
            if logger.isEnabledFor(logging.DEBUG):
                logger.debug(
                    "push_bulk to %s succeeded! events=%s",
                    uri,
                    [event.get("event_id") for event in events],
                )
```

Otherwise we'll always generate the list, even if the logging isn't set to debug

changelog.d/11265.bugfix
```diff
@@ -0,0 +1 @@
+Prevent [MSC2716](https://github.com/matrix-org/matrix-doc/pull/2716) historical state events from being pushed to an application service via `/transactions`.
```
erikjohnston, line 1: I'm a bit confused why these would come down `/transactions` and not down `/sync`?

MadLittleMods: I was curious a little bit about this as well and saw that `/transactions` only cares about `stream_ordering` so it made sense why it still came down and didn't look into it further.

Diving into this more, here are the details:

`/sync` looks for `stream_ordering` but excludes all `outliers`.

https://github.com/matrix-org/synapse/blob/b09d90cac9179f84024d4cb3ab4574480c1fd1df/synapse/storage/databases/main/stream.py#L519-L527

Whereas `/transactions` only cares about `stream_ordering`. Perhaps we should exclude `outliers` in `/transactions`?

https://github.com/matrix-org/synapse/blob/b09d90cac9179f84024d4cb3ab4574480c1fd1df/synapse/storage/databases/main/appservice.py#L358-L368



 
## `/sync` stack trace

 - [`SyncRestServlet.on_GET`](https://github.com/matrix-org/synapse/blob/b09d90cac9179f84024d4cb3ab4574480c1fd1df/synapse/rest/client/sync.py#L113)
 - [`wait_for_sync_for_user`](https://github.com/matrix-org/synapse/blob/b09d90cac9179f84024d4cb3ab4574480c1fd1df/synapse/handlers/sync.py#L299)
 - [`_wait_for_sync_for_user`](https://github.com/matrix-org/synapse/blob/b09d90cac9179f84024d4cb3ab4574480c1fd1df/synapse/handlers/sync.py#L329)
 - [`current_sync_for_user`](https://github.com/matrix-org/synapse/blob/b09d90cac9179f84024d4cb3ab4574480c1fd1df/synapse/handlers/sync.py#L399)
 - [`generate_sync_result`](https://github.com/matrix-org/synapse/blob/b09d90cac9179f84024d4cb3ab4574480c1fd1df/synapse/handlers/sync.py#L1039)
 - [`_generate_sync_entry_for_rooms`](https://github.com/matrix-org/synapse/blob/b09d90cac9179f84024d4cb3ab4574480c1fd1df/synapse/handlers/sync.py#L1455)
 - [`_get_rooms_changed`](https://github.com/matrix-org/synapse/blob/b09d90cac9179f84024d4cb3ab4574480c1fd1df/synapse/handlers/sync.py#L1614)
 - [`get_room_events_stream_for_rooms`](https://github.com/matrix-org/synapse/blob/b09d90cac9179f84024d4cb3ab4574480c1fd1df/synapse/storage/databases/main/stream.py#L409)
 - [`get_room_events_stream_for_room`](https://github.com/matrix-org/synapse/blob/b09d90cac9179f84024d4cb3ab4574480c1fd1df/synapse/storage/databases/main/stream.py#L478-L527)


## `/transactions` stack trace

 - [`notify_interested_services`](https://github.com/matrix-org/synapse/blob/b09d90cac9179f84024d4cb3ab4574480c1fd1df/synapse/handlers/appservice.py#L66-L87)
 - [`_notify_interested_services`](https://github.com/matrix-org/synapse/blob/b09d90cac9179f84024d4cb3ab4574480c1fd1df/synapse/handlers/appservice.py#L90-L135)
    - [`get_new_events_for_appservice`](https://github.com/matrix-org/synapse/blob/b09d90cac9179f84024d4cb3ab4574480c1fd1df/synapse/storage/databases/main/appservice.py#L353-L368)
    - [`submit_event_for_as`](https://github.com/matrix-org/synapse/blob/b09d90cac9179f84024d4cb3ab4574480c1fd1df/synapse/handlers/appservice.py#L133-L135)

[`submit_event_for_as`](https://github.com/matrix-org/synapse/blob/b09d90cac9179f84024d4cb3ab4574480c1fd1df/synapse/appservice/scheduler.py#L94-L95) ->

 - [`enqueue_event`](https://github.com/matrix-org/synapse/blob/b09d90cac9179f84024d4cb3ab4574480c1fd1df/synapse/appservice/scheduler.py#L129-L131)
 - [`_send_request`](https://github.com/matrix-org/synapse/blob/b09d90cac9179f84024d4cb3ab4574480c1fd1df/synapse/appservice/scheduler.py#L137-L157)
 - `appservice.scheduler.send`
 - [`create_appservice_txn`](https://github.com/matrix-org/synapse/blob/b09d90cac9179f84024d4cb3ab4574480c1fd1df/synapse/appservice/scheduler.py#L196-L201)
 - [`AppServiceTransaction.send`](https://github.com/matrix-org/synapse/blob/b09d90cac9179f84024d4cb3ab4574480c1fd1df/synapse/appservice/__init__.py#L338)
 - [`push_bulk`](https://github.com/matrix-org/synapse/blob/b09d90cac9179f84024d4cb3ab4574480c1fd1df/synapse/appservice/api.py#L202)
 - [`/transactions`](https://github.com/matrix-org/synapse/blob/b09d90cac9179f84024d4cb3ab4574480c1fd1df/synapse/appservice/api.py#L220)

synapse/appservice/api.py
MadLittleMods: Maybe both 👀. The warnings are kinda useless without the little extra info on type of error and arguments. But also useful to have the stack when debugging.

examples with `exc_info=True`
```python
push_bulk to http://host.docker.internal:65302/transactions/1 received code=500 msg=Internal Server Error
  File "/usr/local/lib/python3.8/site-packages/synapse/appservice/api.py", line 229, in push_bulk
    await self.put_json(
  File "/usr/local/lib/python3.8/site-packages/synapse/http/client.py", line 639, in put_json
    raise HttpResponseException(
synapse.api.errors.HttpResponseException: 500: Internal Server Error
```

```python
push_bulk to http://host.docker.internal:49443/transactions/1 threw exception(ConnectionRefusedError) Connection was refused by other side: 111: Connection refused. args=('Connection refused',)
  File "/usr/local/lib/python3.8/site-packages/synapse/appservice/api.py", line 229, in push_bulk
    await self.put_json(
  File "/usr/local/lib/python3.8/site-packages/synapse/http/client.py", line 630, in put_json
    response = await self.request(
  File "/usr/local/lib/python3.8/site-packages/synapse/http/client.py", line 440, in request
    response = await make_deferred_yieldable(request_deferred)
twisted.internet.error.ConnectionRefusedError: Connection was refused by other side: 111: Connection refused.
```

changelog.d/11265.bugfix
richvdh: Apologies for jumping in without context, but

> Perhaps we should exclude outliers in /transactions?

the outlier flag seems like a poor way to decide whether we should push this data. (Yes, outliers shouldn't be sent over `/transactions`, but there are probably many other events which shouldn't be sent).

`FederationEventHandler._process_received_pdu` has a `backfilled` parameter (see https://github.com/matrix-org/synapse/blob/b09d90cac9179f84024d4cb3ab4574480c1fd1df/synapse/handlers/federation_event.py#L945), whose purpose is slightly unclear, but I *think* one of its jobs is this sort of thing. Maybe we should use similar logic to that, somehow?

synapse/handlers/room_batch.py
richvdh: whatever else happens, please please please make sure that any parameters are added to the docstrings *at each level*, with the intended purpose clearly described. 

Relatedly: as a general rule, I would say it is preferable for method parameters to describe a change in behaviour ("`inhibit_sync_to_clent`") rather than describe some bit of state that the function is free to interpret as it wishes: it's much clearer for a later reader to understand how things tie together, particularly when things then get modified later on. Obviously this isn't a hard-and-fast rule, but it's worth considering.

[Currently we have a bit of a mess in some places (`FederationEventHander.backfilled` being a prime example) where you have to dig through 150 methods to find out what a flag actually does, and then somehow reverse-engineer what it is *supposed* to do. Don't do that.]

MadLittleMods: > please please please make sure that any parameters are added to the docstrings at each level, with the intended purpose clearly described.

Will do 🙇 

---

> Relatedly: as a general rule, I would say it is preferable for method parameters to describe a change in behaviour ("inhibit_sync_to_clent") rather than describe some bit of state that the function is free to interpret as it wishes

Seems reasonable. I've created a separate issue to track this https://github.com/matrix-org/synapse/issues/11300

changelog.d/11265.bugfix
MadLittleMods: Whatever we decide on to make `/sync` and `/transactions` behave similarly, we can tackle in another PR. This PR is still good to ship on its own, i.e. It's a good idea to mark historical state as `historical`

---

> `FederationEventHandler._process_received_pdu` has a `backfilled` parameter [...]
> Maybe we should use similar logic to that, somehow?

For `/transactions`, we can't tell whether the event was `backfilled`. The only indication is that the `stream_ordering` would be negative which is what the `/transactions` code already takes into account. And is why the fix in this PR to mark the historical state events `historical` -> `backfilled` works to not show up in `/transactions`.

I just suggested adding `outlier` to `/transactions` query as well so it would match the query in `/sync`

synapse/appservice/api.py
erikjohnston: Fair. Ideally converting an exception to a string should give us useful information, but clearly those exceptions (which we don't control) don't follow that rule :/

erikjohnston:
This approach seems reasonable to me as a stop gap before we try and use the same rules for sending events down `/sync` and `/transactions`

MadLittleMods:
@erikjohnston Approval please 🙂 

Do you want @richvdh to approve as well before merge?

richvdh:
given erik has the context here, I'm happy to defer to him, if he has the bandwidth to look at it. I was just making driveby comments before!

erikjohnston (approved):

MadLittleMods:
Thanks for the review @erikjohnston and @richvdh 🐖

changelog.d/11265.bugfix
MadLittleMods: Created https://github.com/matrix-org/synapse/issues/11394 to track this ⏩

synapse/appservice/api.py
```diff
@@ -231,13 +231,32 @@ async def push_bulk(
                 json_body=body,
                 args={"access_token": service.hs_token},
             )
+            if logger.isEnabledFor(logging.DEBUG):
```
MadLittleMods, line 234: I came back to this and was confused why I did this since it's guarding debug against debug.

AFAICT, this is to avoid potentially expensive string construction when we loop through all of the `events` here. We do this in a few other places in the codebase too.

PR #11268, opened 2021-11-07
Wrong DTLS port in TURN server "Troubleshooting" section.

jmcparland:
Port 5349, not 5479.

### DIFF
--- a/docs/turn-howto.md
+++ b/docs/turn-howto.md
@@ -204,7 +204,7 @@ Here are a few things to try:
    anyone who has successfully set this up.
 
  * Check that you have opened your firewall to allow TCP and UDP traffic to the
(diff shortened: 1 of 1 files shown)


erikjohnston:
Thanks for catching that, it does indeed look like the port is wrong. 

Can you add a changelog entry as per the [contributing guide](https://matrix-org.github.io/synapse/latest/development/contributing_guide.html) please? I.e. something like `changelog.d/11268.doc`: 

> Fix the incorrect DTLS port in TURN server "Troubleshooting" section.

erikjohnston:
Oh! I didn't notice that you had targeted the `master` branch, which is a) fine for docs and b) means you don't need to do a changelog. :+1:

erikjohnston (approved):
Thanks!

Issue #11271, opened 2021-11-08
Get synapse/ to pass mypy


callahad:
Our mypy configuration currently includes an explicit [enumeration](https://github.com/matrix-org/synapse/blob/09cb441a043947ee367820b56d189c02f5fd35a6/mypy.ini#L19-L70) of files to check. That list covers 409 of the 491 Python files in the `synapse/` directory.

We should:

- Get as many of those to pass as possible
- Invert the selection criteria so that all files in `synapse/` are checked by default, while failing files are explicitly excluded.

As of 09cb441a043947ee367820b56d189c02f5fd35a6, the omitted files are:

- `_scripts/*` (3 files)
- `app/*` (16 files)
- `python_dependencies.py`
- `storage/databases/__init__.py`
- `storage/databases/main/*` (39 of 49 files omitted)
- `storage/schema/*` (22 files)

We should probably start with switching from inclusion to exclusion, and burn down from there. Note that the next release of mypy will have [significantly better](https://github.com/python/mypy/pull/11329) syntax for multiple exclusions, but it's not out yet, so we'll have to get by with gnarlier regex-based syntax.

DMRobertson:
Things are much better now. The database modules remain a sticking point (related: #11733). It is very hard to untangle the MRO sphagetti into a form that mypy approves of and some extra thought is required.

The schema files are probably the next best thing to tackle. For bonus points, it'd be great if we could define a `Protocol` which specifies the interface expected of a schema upgrade file and have the mypy plugin enforce that.

clokep:
> The schema files are probably the next best thing to tackle. For bonus points, it'd be great if we could define a `Protocol` which specifies the interface expected of a schema upgrade file and have the mypy plugin enforce that.

I had some issues even getting mypy to see those files properly since they're not in modules (there are not `__init__` files). So I'm unsure there's much we can do there?

PR #11272, opened 2021-11-08
Require all Python to pass mypy unless specifically excluded

callahad:
This is a maximal response to #11271: instead of specifying files to _include_ in mypy checking, all files are considered unless explicitly _excluded._

We may want to just take the first commit instead, and focus only on the `synapse/` directory. Or some middle ground.

For example: instead of getting this whole repo to pass Mypy in a single run, we may want to:

- Explicitly exempt everything in `contrib/`, `tests/`, etc.
- Check standalone scripts, schema migrations, etc. separately and individually, instead of as part of a single omnibus mypy run for the `synapse` module (doing so separately is [encouraged](https://github.com/python/mypy/issues/4008#issuecomment-332032784) by mypy, but the `scripts_are_modules` setting kind of works around it).

### Pull Request Checklist

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -17,24 +17,7 @@ no_implicit_optional = True
 files =
   scripts-dev/sign_json,
   synapse/,
-  tests/replication,
-  tests/test_event_auth.py,
-  tests/test_utils,
-  tests/handlers/test_password_providers.py,
-  tests/handlers/test_room.py,
-  tests/handlers/test_room_summary.py,
-  tests/handlers/test_send_email.py,
-  tests/handlers/test_sync.py,
-  tests/handlers/test_user_directory.py,
-  tests/rest/client/test_login.py,
-  tests/rest/client/test_auth.py,
-  tests/rest/client/test_relations.py,
-  tests/rest/media/v1/test_filepath.py,
-  tests/rest/media/v1/test_oembed.py,
-  tests/storage/test_state.py,
-  tests/storage/test_user_directory.py,
-  tests/util/test_itertools.py,
-  tests/util/test_stream_change_cache.py
+  tests/
 
 # Note: Better exclusion syntax in mypy > 0.910
 # See https://github.com/python/mypy/pull/11329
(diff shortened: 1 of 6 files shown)


DMRobertson (approved):
LGTM. We might want to ensure that the Python outside of the `synapse` package is linted too? But that's a separate change.

callahad:
Would appreciate a second review before merging, since this does pretty significantly change our static type checking strategy across the repo.

reivilibre (approved):
I also think this is sensible and preferable — good to know that you can get the exclude functionality with a monster regex.

mypy.ini
```diff
@@ -9,83 +9,189 @@ mypy_path = stubs
...
-  tests/rest/client/test_auth.py,
-  tests/rest/client/test_relations.py,
-  tests/rest/media/v1/test_filepath.py,
-  tests/rest/media/v1/test_oembed.py,
-  tests/storage/test_state.py,
-  tests/storage/test_user_directory.py,
-  tests/util/test_itertools.py,
-  tests/util/test_stream_change_cache.py
+  # Discover all .py[i] files in the repo
+  .,
```
richvdh, line 19: won't this iterate into virtualenvs etc?

mypy.ini
```diff
@@ -9,83 +9,189 @@ mypy_path = stubs
...
+# See https://github.com/python/mypy/pull/11329
+exclude = (?x)
+  ^(contrib/cmdclient/console.py
+   |contrib/cmdclient/http.py
+   |contrib/experiments/cursesio.py
+   |contrib/experiments/test_messaging.py
+   |contrib/graph/graph.py
+   |contrib/graph/graph2.py
+   |contrib/graph/graph3.py
+   |contrib/jitsimeetbridge/jitsimeetbridge.py
```
richvdh, line 43: I guess I'm not *entirely* convinced we need to enforce mypy for `contrib` stuff, but :man_shrugging:

mypy.ini
```diff
@@ -9,83 +9,189 @@ mypy_path = stubs
...
+  scripts/register_new_matrix_user,
+ #scripts/synapse_port_db,
+  scripts/synapse_review_recent_signups,
+ #scripts/update_synapse_database,
+  scripts-dev/build_debian_packages,
+  scripts-dev/sign_json
+
+# Note: Better exclusion syntax in mypy > 0.910
+# See https://github.com/python/mypy/pull/11329
+exclude = (?x)
```
richvdh, line 35: since this is enabling comments... maybe we could add a comment to say what it is doing?

```suggestion
exclude = (?x)   # enable verbose REs, see https://docs.python.org/3/library/re.html#re.X
```

mypy.ini
callahad: Mypy won't recurse into hidden directories, so as long as your venv is in `.venv/` you're golden.

Are there any other common paths we should specifically exclude?

mypy.ini
callahad: Yeah, me either. Heck, I'm not even sure we should be shipping `contrib/` in this repo at all 😉 

I'll switch this to a blanket exclusion of `|contrib/.*`

mypy.ini
DMRobertson: The README (and maybe docs) recommends putting the venv in an `env` directory with `python3 -m venv ./env`. I did this on day one.

callahad: Hm. Y'all should definitely switch to `.venv`, but I'll add an exclusion for `env/` for now ;)

DMRobertson: Happy to; I was just getting at that we ought to update the README! (and perhaps for sydent/sygnal too?)

richvdh: I'm not enthusiastic about adding this sort of constraint to people's dev environments. I'm not aware of any rule that says "thou shalt have no virtualenvs in your working tree save that they be called `.venv`", and I don't necessarily expect anyone else to be aware of such a rule.

Please can we just list the directories we care about?

richvdh: To add to this: my main concern is if every different project starts having its own, specific rules about where the venv should go. I really don't want to have to start remembering different places for different projects.

So, if this is a thing we want to do, we have to get into a big debate about what The One True Place For Your Virtualenv should be, which isn't a fight I particularly want to have.

richvdh: also also: I have several *different* venvs in my working copy, with different selections of libraries, and python interpreters, so in any case just excluding `env` doesn't help a great deal.

mypy.ini
```diff
@@ -9,83 +9,202 @@ mypy_path = stubs
...
-  tests/rest/client/test_auth.py,
-  tests/rest/client/test_relations.py,
-  tests/rest/media/v1/test_filepath.py,
-  tests/rest/media/v1/test_oembed.py,
-  tests/storage/test_state.py,
-  tests/storage/test_user_directory.py,
-  tests/util/test_itertools.py,
-  tests/util/test_stream_change_cache.py
+  # Discover all .py[i] files in the repo
+  .,
```
richvdh, line 19: do we need to include `.ci/scripts` here?

callahad: Ah, yep!

mypy.ini
callahad: /me gestures wildly at "I'll add an exclusion for env/ for now ;)" and 515ca6bb6c085cd94d1ccf11bd07faf81ea95fde from an hour ago

richvdh: I'm yet to be convinced, I'm afraid.

richvdh: basically anything that's not checked in - my whole point is that this shouldn't break when people have things in their working copy that you don't expect.

richvdh: I guess maybe it would be ok to ignore things that are in the user's various gitignore configs, but that feels a bit magical.

callahad: In [Apache voting terms](http://www.apache.org/foundation/voting.html#expressing-votes-1-0-1-and-fractions), how strongly are your feelings on this issue?

Assume the alternative would be to only cover `synapse/` and `tests/` with mypy, allowing other files in the repo to go without type checking.

richvdh: > In [Apache voting terms](http://www.apache.org/foundation/voting.html#expressing-votes-1-0-1-and-fractions), how strongly are your feelings on this issue?

-0.9ish.

I like this way of resolving the dispute.

> Assume the alternative would be to only cover synapse/ and tests/ with mypy, allowing other files in the repo to go without type checking.

Well, we could surely add other directories - `scripts-dev`, `scripts`, etc. As noted elsewhere, we already have to have multiple entries in the list.

mypy.ini
```diff
@@ -9,83 +9,202 @@ mypy_path = stubs
...
+
+  # Also check scripts that don't end in .py
```
richvdh, line 21: `synctl` is missing from this list.

PR #11278, opened 2021-11-08
Fix typo in `RelationAggregationPaginationServlet` error response

anoadragon453:
So that it now reads: Relation type must be 'm.annotation'.


clokep (approved):
Seems reasonable.

anoadragon453:
Oops, not sure how that got left in draft state. Thanks for the review!

Issue #8159, opened 2020-08-25
Separate default configuration from configuration documentation


erdnaxeli:
This is not _really_ a bug but a really annoying behavior.

The debian package's configuration file change every time. Every time you update your synapse package, you have to go through the process of looking at the diff between the old and the new configuration file to check for relevant modifications. And often there are not.

For example when updating from 1.18.0+buster1 to 1.19.0+buster1 the changes where:
* a blank line removed
* the federation_ip_range_blacklist parameter and all its documentation position has changed
* some other not impacting changes

A very common and very frustrating change is the cosmetic one: removing (or adding) a space after a "#", fixing a typo, changing the indentation, …

This bother me every time I update my synapse. I also manage another one with an ansible playbook, and it makes me feel uncomfortable, because I have no way to known if I need to change my config file. Most of the time I do not change it, let it diverge from the one in the package, and every some months I go through the painful process of doing a full diff (btw finding the default conf is not obvious, the easiest way is to just to the update manually, going through the diff process with apt, and copy back the change to the config file in the ansible role).

My suggestion is that **config files are not for documentation**.  That's a good idea when your config file is ten lines long and never changes, but that's not the case here. The nginx configuration documentation is not in the config file, neither the synapse one should. Please find somewhere else to put the doc, and leave our config file alone :p

If there is a relevant change, like an attribute changing name, or a new attribute, please indicate it in a changelog file (`apt-listchanges` will show it at installation). For everything else, like a documentation change or a typo fix, publish the new documentation somewhere but please (please!) don't do a config file diff.

Thanks you :)

richvdh:
I would love for this to be fixed. The default config file should not be a 3000-line epic, but rather a *minimal* configuration that is just enough to get going. 

(as an aside, a workaround for the particular pain felt by users of the Debian package is to put your local changes into separate config files in the `conf.d` directory. But it's definitely a workaround.)

reivilibre:
#11281 suggests Debian users to use `conf.d`, which again is just a workaround but thought it was worth linking to this issue.

anoadragon453:
Thank you for starting on this with #12368 @H-Shay! You mention in that PR that it's the first step in solving this issue. Could you perhaps list the rest of the steps here?

H-Shay:
[...]

PR #11281, opened 2021-11-09
Suggest using `/etc/matrix-synapse/conf.d/` for configuration with Debian packages

### LINKED ISSUE #11238: apt update/upgrade Configuration file '/etc/matrix-synapse/homeserver.yaml'  ==> Modified (by you or by a script) since installation.
**Description:**
would it be possible to support a local configuration, eg: 
homeserver.yaml.local

the values defined in .local would overwrite anything specificed in the main homeserver.yaml

This way I can leave homeserver.yaml untouched and will not be prompted what to do when updating the server.
(or if there is another way to accomplish the same result, please let me know, thank you.)

when I perform an apt update of my synapse install.
I am always prompted:

Setting up matrix-synapse-py3 (1.46.0+focal1) ...
```
Configuration file '/etc/matrix-synapse/homeserver.yaml'
 ==> Modified (by you or by a script) since installation.
 ==> Package distributor has shipped an updated version.
   What would you like to do about it ?  Your options are:
    Y or I  : install the package maintainer's version
    N or O  : keep your currently-installed version
      D     : show the differences between the versions
      Z     : start a shell to examine the situation
 The default action is to keep your current version.
*** homeserver.yaml (Y/I/N/O/D/Z) [default=N] ?
```

reivilibre:
I don't know if this is actually best, so hopefully someone with Debian opinions can confirm/deny if this is a good idea or not.

Fixes #11238.

### DIFF
--- a/changelog.d/11281.doc
+++ b/changelog.d/11281.doc
@@ -1,0 +1,1 @@
+Suggest users of the Debian packages add configuration to `/etc/matrix-synapse/conf.d/` to prevent, upon upgrade, being asked to choose between their configuration and the maintainer's.
--- a/docs/setup/installation.md
+++ b/docs/setup/installation.md
@@ -76,6 +76,12 @@ The fingerprint of the repository signing key (as shown by `gpg
 /usr/share/keyrings/matrix-org-archive-keyring.gpg`) is
 `AAF9AE843A7584B5A3E4CD2BCF45A512DE2DA058`.
 
+When installing with Debian packages, you might prefer to place files in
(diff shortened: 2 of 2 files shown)


richvdh (approved):
seems plausible to me.

A lot of the problem here is that we're conflating "comprehensive config documentation" and "default configuration file", which means that `/etc/matrix-synapse/homeserver.yaml` changes every single release. For that, see https://github.com/matrix-org/synapse/issues/8159.

PR #11284, opened 2021-11-09
Bundle relations of relations into the `/relations` result.

### LINKED ISSUE #11087: Include bundled events when requesting relations
In order for [MSC3440](https://github.com/matrix-org/matrix-doc/pull/3440) to be successful it proposes being explicit about the behavior in [MSC2675](https://github.com/matrix-org/matrix-doc/pull/2675) that the `/relations` API should include the bundled relations of each returned event.

The behavior here was changed a few times:

* #5595 asked to include the original event (in the `original_event` field), which was done in #5626. Both of these fields included bundle aggregations (which is a bit odd since you're requesting the relations).
* #5654 stopped bundling aggregations in both the original event *and* the relations events.

The proposal here is to include them on the relations event. Note that #5654 includes the following:

> We also do this when pulling the relation events as well, because edits of edits are not something we'd like to support here.

The code in question lies at:

https://github.com/matrix-org/synapse/blob/1f9d0b8a7a6a777d59fe3217724f3e2ddb94a9b2/synapse/rest/client/relations.py#L232-L237

Which asserts that we don't allow relations to relations, but I don't think this is accurate. If this is a problem for edits we might need to solve this a different way.

clokep:
For [MSC3440](https://github.com/matrix-org/matrix-doc/pull/3440), thread relations might have other relations to them (e.g. annotations and edits). It is useful to include those in the `/relations` API response so that an entire thread can be rendered at once.

Fixes #11087, which has some background about this. Note that we've flipped this flag back and forth a few times so it is probably worth looking at the history. This also modifies the serialization to *not* include bundled relations for edits / annotations which I think will assuage the concerns for why it was previously not included.

Overall this makes two changes:

* Events returned from the `/relations` API return bundled events.
* Edits and annotations no longer return bundled events (anywhere).

## TODO

* [x] Write tests
* [ ] Does this need to be behind a config flag?


clokep:
> Fixes #11087, which has some background about this. Note that we've flipped this flag back and forth a few times so it is probably worth looking at the history.

I'm looking for a double check that this all seems reasonable before going forward and writing tests, etc. Does it seem OK to send this additional data here or is it correct that we don't include bundled relations? I'm hesitant to dump too much work into this upfront since we've flipped this flag back and forth a few times.

erikjohnston:
I *think* this is probably fine. I'm wondering if the comment is more referring to the fact it would be pointless to bother checking for aggregations at all?

I'd still be tempted to add explicit checks to ensure that we don't apply edits to edits etc, as I don't *think* those are checks we are able to do when we receive events over federation.

clokep:
> I'd still be tempted to add explicit checks to ensure that we don't apply edits to edits etc, as I don't _think_ those are checks we are able to do when we receive events over federation.

Would it make sense to only include bundled relations for the thread relation type, maybe? I don't know how hard that would be to do though.

clokep:
Going to throw this onto @erikjohnston's review queue to get an answer to the above.

erikjohnston:
> > I'd still be tempted to add explicit checks to ensure that we don't apply edits to edits etc, as I don't _think_ those are checks we are able to do when we receive events over federation.
> 
> Would it make sense to only include bundled relations for the thread relation type, maybe? I don't know how hard that would be to do though.

Yeah. Or flip it around so that we don't bundle aggregation for reactions/edits in `serialize_events`?

synapse/rest/client/relations.py
```diff
@@ -230,12 +230,9 @@ async def on_GET(
...
             event, now, bundle_aggregations=False
         )
-        # Similarly, we don't allow relations to be applied to relations, so we
-        # return the original relations without any aggregations on top of them
-        # here.
-        serialized_events = await self._event_serializer.serialize_events(
-            events, now, bundle_aggregations=False
-        )
+        # For any relations applying to the original event they need their
+        # aggregations applied to them.
```
richvdh, line 234: I'm a bit confused about what this comment is trying to tell me?

synapse/events/utils.py
```diff
@@ -398,7 +398,7 @@ async def serialize_event(
         """Serializes a single event.
 
         Args:
-            event
+            event: The event being serialized.
             time_now: The current time in milliseconds
             bundle_aggregations: Whether to bundle in related events
```
richvdh, line 403: I... don't know what this means.

(also: are bundled relations the same as bundled aggregations?)

richvdh (commented):
I think I need to go and learn what bundled relations are before I can I review this :/

synapse/events/utils.py
```diff
@@ -419,68 +418,91 @@ async def serialize_event(
...
+    ) -> None:
+        """Potentially injects bundled relations into the unsigend portion of the serialized event.
+
+        Args:
+            event: The event being serialized.
+            time_now: The current time in milliseconds
+            serialized_event: The serialized event which may be modified.
+
+        """
+        # Do not bundle aggregations for an event which represents an edit or annotation.
```
richvdh, line 436: persuant to the thread at https://github.com/matrix-org/matrix-doc/pull/2675#discussion_r753343727: we've got "bundle relations" and "bundle aggregations" within 10 lines of each other. Let's stick to one term.

synapse/events/utils.py
```diff
@@ -419,68 +418,91 @@ async def serialize_event(
...
-                if latest_thread_event:
-                    r = serialized_event["unsigned"].setdefault("m.relations", {})
-                    r[RelationTypes.THREAD] = {
-                        # Don't bundle aggregations as this could recurse forever.
-                        "latest_event": await self.serialize_event(
-                            latest_thread_event, time_now, bundle_aggregations=False
-                        ),
-                        "count": thread_count,
-                    }
+            await self._injected_bundled_relations(event, time_now, serialized_event)
```
richvdh, line 421: it's unfortunate that we're combining a refactor and functional changes in the same PR here. AIUI the only non-refactoring change is:
 * when looking at an original event which is itself an edit or annotation of another, do not bundle any related events with it.

Could you confirm?

synapse/events/utils.py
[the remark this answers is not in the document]
richvdh: right, this is replacing the "bundle_aggregations=False" in the original, right?

Seems generally plausible, but still: expand the comment please.

Also: this needs to be made explicit in MSC2675 imho. It may already be lost somewhere in the sea of comments on there, in which case, a link would be appreciated.

richvdh (commented):
plausible? generally a bit worried about the number of unresolved threads on MSC2675.

synapse/events/utils.py
clokep: > (also: are bundled relations the same as bundled aggregations?)

I believe so, the MSC is really confusing at what it calls bundled relations & aggregations.

synapse/events/utils.py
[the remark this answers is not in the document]
clokep: > Also: this needs to be made explicit in MSC2675 imho. It may already be lost somewhere in the sea of comments on there, in which case, a link would be appreciated.

It isn't really explicit, hence the back and forth in previous PRs about it.

synapse/events/utils.py
clokep: I can split this into two PRs if you would like!

This makes two changes:

* Events returned from the `/relations` API return bundled events.
* Edits and annotations no longer return bundled events (anywhere when returned by Synapse).

clokep: I split out the refactoring into a separate PR: #11408.

synapse/rest/client/relations.py
clokep: I've attempted to clarify it!

synapse/events/utils.py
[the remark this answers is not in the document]
clokep: This is partially due to a comment on #5654:

> We also do this when pulling the relation events as well, because edits of edits are not something we'd like to support here.

@erikjohnston also [asserted above](https://github.com/matrix-org/synapse/pull/11284#issuecomment-966499736) that we shouldn't be returning edits of edits.

clokep:
@richvdh I've merged in develop after separating out the refactoring. I can rebase this if it is easier, but the entire diff now should be reasonable to view at once!

I've also updated the description a bit to be more explicit about the changes here, but hopefully with the smaller diff it is straightforward.

changelog.d/11284.feature
```diff
@@ -0,0 +1 @@
+Include bundled relations in response to the `/relations` API, per [MSC3440](https://github.com/matrix-org/matrix-doc/pull/3440).
```
richvdh, line 1: to be clear, we're talking about relations of relations of the original event, right? 

Suppose I request the relations for event E, where:
 * E has relations A, B and C
 * each of A, B and C has its own relations X, Y and Z respectively
 * once this PR lands, we will bundle X, Y and Z in with A, B and C, whereas we would not before.

In other words, we're considering meta-relations...

Is that a correct understanding? If so, I think both this newsfile and the PR summary could be clearer.

richvdh (commented):
generally, the change seems sound - thanks! Some nitpicking about wording though.

changelog.d/11284.feature
clokep: > Is that a correct understanding? If so, I think both this newsfile and the PR summary could be clearer.

Yes, this is correct. I've been having trouble with some of the wording around relations. Do you have any suggestions of how to make it clearer?

richvdh: For the newsfile:

> When returning relation events from the `/relations` API, bundle any relations of those relations into the result, per MSC3440.

It's not great, but it's reasonably accurate.

For the PR summary:

> Bundle relations of relations into the `/relations` result.

clokep: Note that this was also updated in MSC2675. It now reads:

> The following client-server APIs should bundle aggregations with events they return:
>
> * ...
> * `GET /relations`, as proposed in this MSC.

```suggestion
When returning relation events from the `/relations` API, bundle any relations of those relations into the result, per updates to [MSC2675](https://github.com/matrix-org/matrix-doc/pull/2675).
```

richvdh (approved):
lgtm!

PR #11282, opened 2021-11-09
Require mypy for synapse/ & tests/ unless excluded

callahad:
This is a minimal response to #11271: all files in `synapse/` and `tests/` must pass mypy unless explicitly excluded.

Contrast with #11272

Signed-off-by: Dan Callahan <danc@element.io>

### Pull Request Checklist

* [x] Pull request is based on the develop branch
* [x] Pull request includes a [changelog file](https://matrix-org.github.io/synapse/latest/development/contributing_guide.html#changelog). The entry should:
  - Be a short description of your change which makes sense to users. "Fixed a bug that prevented receiving messages from other servers." instead of "Moved X method from `EventStore` to `EventWorkerStore`.".
  - Use markdown where necessary, mostly for `code blocks`.
  - End with either a period (.) or an exclamation mark (!).
  - Start with a capital letter.
* [x] Pull request includes a [sign off](https://matrix-org.github.io/synapse/latest/development/contributing_guide.html#sign-off)
* [x] [Code style](https://matrix-org.github.io/synapse/latest/code_style.html) is correct
  (run the [linters](https://matrix-org.github.io/synapse/latest/development/contributing_guide.html#run-the-linters))

PR #11285, opened 2021-11-09
Ignore missing imports for parameterized.

clokep:
This seems to be from a conflict between #11282 and #11228.

Example failing run: https://github.com/matrix-org/synapse/runs/4154822727?check_suite_focus=true

### DIFF
--- a/changelog.d/11285.misc
+++ b/changelog.d/11285.misc
@@ -1,0 +1,1 @@
+Require all files in synapse/ and tests/ to pass mypy unless specifically excluded.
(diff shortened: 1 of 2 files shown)


DMRobertson (approved):
LGTM assuming CI passes.

Sadly there aren't any type stubs in typeshed for parameterized.

I wonder why this didn't show up in #11282?

clokep:
> I wonder why this didn't show up in #11282

It was based on #11263, so before #11228 merged.

(Visible via `refs/pull/11282/head`.)

PR #11286, opened 2021-11-09
Fix typos for `username_available` admin API

binarymaster:
- Fix typo in the word `available`
- HTTP method should be `GET`, not `POST`

### DIFF
--- a/changelog.d/11286.doc
+++ b/changelog.d/11286.doc
@@ -1,0 +1,1 @@
+Fix typo in the word `available` and fix HTTP method (should be `GET`) for the `username_available` admin API. Contributed by Stanislav Motylkov.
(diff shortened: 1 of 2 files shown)


clokep (approved):
Looks great, thank you so much!

changelog.d/11286.doc
```diff
@@ -0,0 +1 @@
+Fix typo in the word `available` and fix HTTP method (should be `GET`) for the `username_available` admin API.
```
clokep, line 1: We like to credit our contributors, you can include your nick or name if you'd like!

```suggestion
Fix typo in the word `available` and fix HTTP method (should be `GET`) for the `username_available` admin API. Contributed by Stanislav Motylkov. 
```

binarymaster: Well, it technically should land in the git history, but ok, added it :slightly_smiling_face:

clokep: Yep, but this is what we use to generate the changelogs, which are much more public!

Issue #11165, opened 2021-10-22
Improve type hints for data store usage


squahtx:
Everywhere in Synapse, `HomeServer.get_datastore()` is annotated as the `DataStore` god class, which is incorrect for workers and allowed #11163 to slip into matrix.org, causing an outage. It would be a good idea for data store consumers (servlets, handlers, the module API, etc) to declare which aspects of the data store they need, and have CI check that we don't pass them a data store that's missing the required interfaces.

There are three data store classes to consider:
 * `DataStore`, which is the data store class used on the main process
 * `GenericWorkerSlavedStore`, which is the data store class used on worker processes
 * `AdminCmdSlavedStore`, which is the data store class used when running admin commands(?)

`DataStore` and `GenericWorkerSlavedStore` overlap but aren't subtypes of each other. `AdminCmdSlavedStore` is a subset of `GenericWorkerSlavedStore` functionality-wise, but not through inheritance.

It's been suggested by @reivilibre that we define two or three types for use in type hints:
 * `WorkerStore`
 * `MainStore` (a subtype of `WorkerStore`?)
 * `EitherStore = Union[WorkerStore, MainStore]`, if it turns out that `WorkerStore` contains functionality not in `MainStore`

These don't have to be concrete classes and could be `Protocol`s if needed.

We _could_ have more granular store requirements, but it wouldn't catch any extra bugs.

The code is currently structured like this in a lot of places:
```py
class GroupsLocalWorkerHandler:
    def __init__(self, hs: "HomeServer"):
        self.store = hs.get_datastore()
```

#### Proposal 1: `HomeServer[WorkerStore]`
We could add a covariant type parameter to `HomeServer` to have data store consumers declare which type of data store they need:
```py
class GroupsLocalWorkerHandler:
    def __init__(self, hs: "HomeServer[WorkerStore]"):
        self.store = hs.get_datastore()
```
`HomeServer[WorkerStore]` would mean "a homeserver with a data store that supports at least the capabilities of a `WorkerStore`".

We'd have to do this refactor across the entire codebase at once, otherwise the type hints for data stores would be implicitly degraded to `Any` everywhere.

#### Proposal 2: `get_worker_datastore()`
@clokep suggests that we add new `get_*_datastore()` methods to `HomeServer` that raise errors when called on the wrong process:
```py
class GroupsLocalWorkerHandler:
    def __init__(self, hs: "HomeServer"):
        self.store = hs.get_worker_datastore()
```

[...]

PR #11287, opened 2021-11-09
Add missing type hints to `synapse.app`.

clokep:
Finishes off type hints for `synapse.app`.

This should be reviewable commit-by-commit. Please see my comments for explanations of why changes were made.


synapse/app/_base.py
```diff
@@ -60,30 +72,37 @@
...
     """
     _sighup_callbacks.append((func, args, kwargs))
 
 
-def start_worker_reactor(appname, config, run_command=reactor.run):
+def start_worker_reactor(
+    appname: str,
+    config: HomeServerConfig,
+    # mypy can't figure out what the default reactor is.
+    run_command: Callable[[], None] = reactor.run,  # type: ignore[attr-defined]
```
clokep, line 95: There's similar comments in a bunch of places, this is due to how Twisted injects the default reactor into `twisted.internet.reactor`, mypy thinks this is a modular, not an object and gets all upset.

I don't think there's much else to do here (except maybe a mypy plugin?)

synapse/app/_base.py
```diff
@@ -273,12 +301,20 @@ def listen_tcp(bind_addresses, port, factory, reactor=reactor, backlog=50):
         except error.CannotListenError as e:
             check_bind_error(e, address, bind_addresses)
 
-    return r
+    # IReactorTCP returns an object implementing IListeningPort from listenTCP,
+    # but we know it will be a Port instance.
+    return r  # type: ignore[return-value]
```
clokep, line 306: Maybe I should have used a `cast` here? This is vaguely unsafe in that the implementation, not the interface, defines this. Realistically this is deep in Twisted and I can't imagine this changing.

synapse/app/admin_cmd.py
```diff
@@ -65,16 +68,11 @@ class AdminCmdSlavedStore(
 
 
 class AdminCmdServer(HomeServer):
-    DATASTORE_CLASS = AdminCmdSlavedStore
+    DATASTORE_CLASS = AdminCmdSlavedStore  # type: ignore
```
clokep, line 71: I'm essentially punting handling this attribute (for the 3 sub-classes of `HomeServer`) to #11165.

synapse/app/generic_worker.py
```diff
@@ -187,9 +189,8 @@ async def on_POST(self, request: Request, device_id: Optional[str]):
                 # If the header exists, add to the comma-separated list of the first
                 # instance of the header. Otherwise, generate a new header.
                 if x_forwarded_for:
-                    x_forwarded_for = [
-                        x_forwarded_for[0] + b", " + previous_host
-                    ] + x_forwarded_for[1:]
+                    x_forwarded_for = [x_forwarded_for[0] + b", " + previous_host]
+                    x_forwarded_for.extend(x_forwarded_for[1:])
```
clokep, line 193: `requestHeaders.getRawHeaders` is defined as returning a `Sequence` so mypy complained about this. This should be identical, in reality it *does* return a `List`.

synapse/app/generic_worker.py
```diff
@@ -253,24 +254,27 @@ class GenericWorkerSlavedStore(
...
         assert listener_config.http_options is not None
 
         site_tag = listener_config.http_options.tag
         if site_tag is None:
-            site_tag = port
+            site_tag = str(port)
 
         # We always include a health resource.
-        resources: Dict[str, IResource] = {"/health": HealthResource()}
+        resources: Dict[str, Resource] = {"/health": HealthResource()}
```
clokep, line 277: There's many places where I changed `IResource` to `Resouce` this is necessary because `create_resource_tree` calls `resource.listNames()` which is *not* part of the interface, it is an implementation detail of `Resource`.

synapse/util/manhole.py
```diff
@@ -105,7 +105,8 @@ def manhole(settings: ManholeConfig, globals: Dict[str, Any]) -> Factory:
     factory.privateKeys[b"ssh-rsa"] = priv_key  # type: ignore[assignment]
     factory.publicKeys[b"ssh-rsa"] = pub_key  # type: ignore[assignment]
 
-    return factory
+    # ConchFactory is a Factory, not a ServerFactory, but they are identical.
+    return factory  # type: ignore[return-value]
```
clokep, line 109: Maybe this should be a `cast`?

synapse/app/_base.py
```diff
@@ -294,10 +330,13 @@ def listen_ssl(
         except error.CannotListenError as e:
             check_bind_error(e, address, bind_addresses)
 
-    return r
+    # IReactorSSL incorrectly declares that an int is returned from listenSSL,
+    # it actually returns an object implementing IListeningPort, but we know it
+    # will be a Port instance.
```
clokep, line 335: See https://twistedmatrix.com/trac/ticket/10274

clokep: And twisted/twisted#1672.

synapse/app/generic_worker.py
DMRobertson: I wouldn't have spotted an error on the LHS! I was curious, and reproducing it gave me 

> `Unsupported operand types for + ("List[bytes]" and "Sequence[bytes]")

which I didn't realise was a thing. But it is:

```python
>>> [1] + (2,3)
TypeError: can only concatenate list (not "tuple") to list
```

Good to know!

synapse/app/_base.py
DMRobertson: I've come across this kind of thing in Sydent. [In one place](https://github.com/matrix-org/sydent/blob/7bdaac2774840e1f9697d30515f367c5c8c16866/sydent/sydent.py#L105) I ended up opting to introduce `SydentReactor`. I don't know if this kind of approach (with `SynapseReactor`?) would help.

If nothing obvious springs to mind, let's leave it as it is (it's more than fine!)

synapse/app/_base.py
```diff
@@ -151,7 +170,7 @@ def run():
         run()
 
 
-def quit_with_error(error_string: str) -> NoReturn:
+def quit_with_error(error_string: str) -> None:
```
DMRobertson, line 174: I think `NoReturn` was correct, because we call `sys.exit()`.

synapse/app/_base.py
```diff
@@ -161,7 +180,7 @@ def quit_with_error(error_string: str) -> NoReturn:
     sys.exit(1)
 
 
-def handle_startup_exception(e: Exception) -> NoReturn:
+def handle_startup_exception(e: Exception) -> None:
```
DMRobertson, line 184: I think this therefore ought to be `NoReturn` too. I don't know if mypy can/should enforce that.

synapse/app/_base.py
DMRobertson: Given that there are not other lines that follow, I think a cast and a type-ignore are broadly equivalent.

synapse/replication/tcp/resource.py
```diff
@@ -38,7 +38,7 @@
 logger = logging.getLogger(__name__)
 
 
-class ReplicationStreamProtocolFactory(Factory):
+class ReplicationStreamProtocolFactory(ServerFactory):
```
DMRobertson, line 41: What's the motivation here?

DMRobertson: Ahh, [this](https://github.com/twisted/twisted/blob/5c24e99e671c4082a1ddc8dbeb869402294bd0dc/src/twisted/internet/protocol.py#L483) explains it. Nice spot!

synapse/util/manhole.py
DMRobertson: As above, mypy won't do any analysis based on the type of `factory` because it's the last line of this function. I think this is fine.

synapse/app/__init__.py
```diff
@@ -27,7 +28,9 @@
     sys.exit(1)
 
 
-def check_bind_error(e, address, bind_addresses):
+def check_bind_error(
+    e: Exception, address: str, bind_addresses: Container[str]
```
callahad, line 32: This will only ever be a `twisted.internet.error.CannotListenError` -- should we further specialize the type here, or leave well enough alone?

synapse/app/_base.py
callahad: Can we somehow cast `reactor` into something that satisfies the `twisted.internet.interfaces.IReactorCore` interface?

synapse/app/generic_worker.py
clokep: I would have thought the operator would just care for the RHS to be an iterable, but I guess not. 🤷

callahad:
(Oh, weird, had a race condition with DMR's review... thought GitHub had somehow glitched and mistaken my sporadic comments as a review)

synapse/app/_base.py
clokep: Yes, I think `SydentReactor` was based on what we did for `SynapseReactor`. It won't help here though since we're not passing `reactor` in as a parameter, we're importing it at the top of the file. I haven't found a way to tell mypy that an import is of a different type than it hinks.

synapse/app/_base.py
clokep: Doesn't `NoReturn` imply an exception is raised? Although maybe that's how `sys.exist()` works...

synapse/app/homeserver.py
```diff
@@ -413,30 +413,6 @@ def format_config_error(e: ConfigError) -> Iterator[str]:
 
 
 def run(hs: HomeServer):
-    PROFILE_SYNAPSE = False
-    if PROFILE_SYNAPSE:
```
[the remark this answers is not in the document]
clokep: The code doesn't work:

> type object 'ThreadPool' has no attribute '_worker'

It also doesn't have anyway to enable it without modifying code, so I figure this seems less than useful. I believe that cProfile has too much overhead for actually running synapse, but 🤷 I can attempt to get it to work if you'd like.

synapse/app/_base.py
clokep: > Can we somehow cast `reactor` into something that satisfies the `twisted.internet.interfaces.IReactorCore` interface?

I don't think you can cast on import, at least I haven't found anything that allows this.

synapse/app/_base.py
DMRobertson: The docs just say "it won't return", which is a bit ambiguous. I think it's meant to enable detection of unreachable code?

My understanding is that if `f: Callable[..., NoReturn]` then in the following sample

- `x` will never be assigned a value
- `foo()` will never be called

```python
x = f()
foo()
```

FWIW typeshed's [stub](https://github.com/python/typeshed/blob/ab25042efa1c325d242147c51d3cf28478c5a36d/stdlib/sys.pyi#L171) for sys.exit uses NoReturn; see also https://github.com/python/mypy/issues/2280.

synapse/app/__init__.py
clokep: I think leaving it is fine since it just gets re-raised? I don't feel strongly either way though!

callahad: Narrowing it might allow us to get away with not passing `address` and instead accessing `interface` on the exception itself,  but that's a refactor for later :) Happy to call this good enough.

synapse/app/_base.py
callahad: This would do it:

```diff
diff --git a/synapse/app/_base.py b/synapse/app/_base.py
index f2c1028b5d..0641e76bf4 100644
--- a/synapse/app/_base.py
+++ b/synapse/app/_base.py
@@ -22,13 +22,14 @@ import socket
 import sys
 import traceback
 import warnings
-from typing import TYPE_CHECKING, Awaitable, Callable, Iterable
+from typing import TYPE_CHECKING, Awaitable, Callable, Iterable, cast
 
 from cryptography.utils import CryptographyDeprecationWarning
 from typing_extensions import NoReturn
 
 import twisted
-from twisted.internet import defer, error, reactor
+from twisted.internet import defer, error, reactor as _reactor
+from twisted.internet.interfaces import IReactorCore
 from twisted.logger import LoggingFile, LogLevel
 from twisted.protocols.tls import TLSMemoryBIOFactory
 from twisted.python.threadpool import ThreadPool
@@ -57,6 +58,8 @@ from synapse.util.versionstring import get_version_string
 if TYPE_CHECKING:
     from synapse.server import HomeServer
 
+reactor = cast(IReactorCore, _reactor)
+
 logger = logging.getLogger(__name__)
 
 # list of tuples of function, args list, kwargs dict
```

callahad: isort might yell, haven't actually linted, but that's the gist.

DMRobertson: > I haven't found a way to tell mypy that an import is of a different type than it thinks.

Possible ugly workaround?

```python
from twisted.internet import reactor as _reactor
reactor = cast(IReactorCore, _reactor)
```

clokep: Yeah, that's quite ugly. Not sure what I think of it. 😢

synapse/app/__init__.py
clokep: Yep, trying to avoid refactoring too much during these! 🎉

synapse/app/_base.py
```diff
@@ -70,6 +72,11 @@
 if TYPE_CHECKING:
     from synapse.server import HomeServer
 
+# Twisted injects the global reactor to make it easier to import, this confuses
+# mypy which thinks it is a module. Tell it that it a more proper type.
+reactor = cast(ISynapseReactor, _reactor)
```
DMRobertson, line 77: Unfortunate, but hopefully worth it to get the extra checking!

(If we end up making use of this elsewhere, perhaps we can do this import-cast dance in `synapse.types` or somewhere similar?)

DMRobertson (approved):
LGTM! Dunno if @callahad has anything else on his mind?

synapse/app/_base.py
clokep: I had that thought briefly, but decided it wasn't worth tying things more tightly together at the moment!

clokep:
> LGTM! Dunno if @callahad has anything else on his mind?

I think I hit all of the thoughts Dan had, I'm going to merge this but can make more changes if necessary!

DMRobertson:
The diff on `mypy.ini` is great to see! :fireworks:

Issue #10721, opened 2021-08-31
Media-Upload returns 500 for images with width or height of 1px


Barry3D:
### Description

We've recently noticed that uploads to `_matrix/media/r0/upload` always cause an error when one of the images dimensions is only at one pixel.

### Steps to reproduce

- create a file that has one dimension of one pixel
- upload it to synapse (`_matrix/media/r0/upload`)

```
2021-08-31 08:49:51,833 - synapse.http.server - 97 - ERROR - POST-66 - Failed handle request via 'UploadResource': <XForwardedForRequest at 0x7f9a6b2284a8 method='POST' uri='/_matrix/media/r0/upload?filename=bla.png&access_token=<redacted>' clientproto='HTTP/1.0' site='8008'>
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/dist-packages/synapse/http/server.py", line 258, in _async_render_wrapper
    callback_return = await self._async_render(request)
  File "/usr/local/lib/python3.7/dist-packages/synapse/http/server.py", line 286, in _async_render
    callback_return = await raw_callback_return
  File "/usr/local/lib/python3.7/dist-packages/synapse/rest/media/v1/upload_resource.py", line 94, in _async_render_POST
    media_type, upload_name, content, content_length, requester.user
  File "/usr/local/lib/python3.7/dist-packages/synapse/rest/media/v1/media_repository.py", line 182, in create_content
    await self._generate_thumbnails(None, media_id, media_id, media_type)
  File "/usr/local/lib/python3.7/dist-packages/synapse/rest/media/v1/media_repository.py", line 733, in _generate_thumbnails
    self.hs.get_reactor(), thumbnailer.scale, t_width, t_height, t_type
  File "/usr/local/lib/python3.7/dist-packages/twisted/python/threadpool.py", line 238, in inContext
    result = inContext.theWork()  # type: ignore[attr-defined]
  File "/usr/local/lib/python3.7/dist-packages/twisted/python/threadpool.py", line 255, in <lambda>
    ctx, func, *args, **kw
  File "/usr/local/lib/python3.7/dist-packages/twisted/python/context.py", line 118, in callWithContext
    return self.currentContext().callWithContext(ctx, func, *args, **kw)
  File "/usr/local/lib/python3.7/dist-packages/twisted/python/context.py", line 83, in callWithContext
    return func(*args, **kw)
  File "/usr/local/lib/python3.7/dist-packages/synapse/logging/context.py", line 902, in g
[...]

PR #11288, opened 2021-11-09
Fix error in thumbnail generation

Neeeflix:
Changed the aspect() function to return at least size 1 in width and height.

### Pull Request Checklist

### DIFF
--- a/changelog.d/11288.bugfix
+++ b/changelog.d/11288.bugfix
@@ -1,0 +1,1 @@
+Changed the aspect() function to return at least size 1 in width and height.
--- a/changelog.d/11288.bugfix
+++ b/changelog.d/11288.bugfix
@@ -1,1 +1,1 @@
-Changed the aspect() function to return at least size 1 in width and height.
+Fix a long-standing bug where uploading extremely thin images (e.g. 1000x1) would fail. Contributed by @Neeeflix.
--- a/changelog.d/11285.misc
+++ b/changelog.d/11285.misc
@@ -1,0 +1,1 @@
+Require all files in synapse/ and tests/ to pass mypy unless specifically excluded.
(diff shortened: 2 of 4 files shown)


DMRobertson (commented):
Many thanks for this! Could you describe an example of how that this change would fix a bug? (I think it applies to source images with an extremely high or extremely low aspect ratio?)

Neeeflix:
> Many thanks for this! Could you describe an example of how that this change would fix a bug? (I think it applies to source images with an extremely high or extremely low aspect ratio?)

This fix refers to issue #10721. In that case an image with ratio 1x1000 px caused an error in the log when generating the thumbnail.

changelog.d/11288.bugfix
```diff
@@ -0,0 +1 @@
+Changed the aspect() function to return at least size 1 in width and height.
```
DMRobertson, line 1: Thanks for clarifying! We should edit the changelog to describe the user-visible bugfix. Here's my suggestion.

I've also included your username to give you credit for the fix. Please change if if you'd like to be credited differently, or remove it if you'd prefer not to be credited.

```suggestion
Fix a long-standing bug where uploading extremely thin images (e.g. 1000x1) would fail. Contributed by @Neeeflix.
```

DMRobertson (changes requested):

DMRobertson:
Many thanks.

Note for completeness that 142c2a7a6beb97ce380594bdc844861446193f14 includes a sign-off.

DMRobertson:
Also for completeness: your first three commits are by `Studying <study@Sisyphos.lan>`. Given that it came from your fork of the repository I think this might just be an artifact of how `git` is configured on your machine; I don't think that address is meant to be attributed as an author. If I've got that wrong, let me know and we can sort it out.

DMRobertson (approved):

PR #11297, opened 2021-11-10
Add type hints to synapse._scripts

clokep:
These were pretty straightforward, a couple of oddities though.


anoadragon453 (approved):
Straightforward and looks correct! Thanks!

Issue #11294, opened 2021-11-10
Include in SSO documentation that SSO redirect confirmation can be bypassed using the `sso.client_whitelist` option


anoadragon453:
It's [not currently obvious](https://github.com/matrix-org/synapse/issues/11290) that the SSO redirection page (introduced as a security measure) can be bypassed by use of the `sso.client_whitelist` option. This option allows specifying a whitelist of client URIs, for which the redirection page won't appear if the client you are being redirected to during SSO login.

We should include some words about it on the yet-to-be-written Single Sign-On documentation page.

richvdh:
> We should include some words about it on the yet-to-be-written Single Sign-On documentation page.

@anoadragon453 would it be worth starting such a page (with just some very basic content like "Synapse supports single sign-on through the SAML, Open ID Connect or CAS protocols") to make this task less daunting for contributors?

anoadragon453:
@richvdh Excellent suggestion. In fact I think we should do that for all of our current "draft pages". It's much easier for contributors to hit a :pencil2: button on the docs site to add content rather than figuring out where to put files and how to update SUMMARY.md.

I've opened a PR to get the ball rolling for Single Sign-On related pages here: https://github.com/matrix-org/synapse/pull/11298.

PR #11298, opened 2021-11-10
Add documentation page stubs for Single Sign-On, SAML and CAS pages

anoadragon453:
As suggested in https://github.com/matrix-org/synapse/issues/11294, let's use page stubs for content we want to see expanded instead of draft pages. It's much easier for contributors to get started on without having to worry about how to update SUMMARY.md, etc.

This PR adds stubs for the Single Sign-On, CAS and SAML pages.

Reviewable commit-by-commit if you like.


richvdh (approved):
looks grand, thanks!

PR #11303, opened 2021-11-11
Move sql file for `remove_deleted_devices_from_device_inbox` into v65

dklimpel:
Between creation and merging PR #10969 was an schema update in #10975 (v1.46.0). That is the reason why the bg task remove_deleted_devices_from_device_inbox is not running in v1.47.0.rc2. https://github.com/matrix-org/synapse/pull/10969/files#diff-d664ca027d1e3730574905917ca3ab47ab476edca1434fe0ae08b8bce06d1fdb has to move from schema 64 to 65.

### Pull Request Checklist

### DIFF
--- a/changelog.d/11303.feature
+++ b/changelog.d/11303.feature
@@ -1,0 +1,1 @@
+Fix a long-standing bug where messages in the `device_inbox` table for deleted devices would persist indefinitely. Contributed by @dklimpel and @JohannesKleine.
--- a/changelog.d/11303.feature
+++ b/changelog.d/11303.feature
@@ -1,1 +1,0 @@
-Fix an issue which prevented the 'remove deleted devices from device_inbox column' background process from running when updating from a recent Synapse version. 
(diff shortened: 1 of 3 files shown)


anoadragon453 (changes requested):
lgtm otherwise! Thanks for catching this.

anoadragon453 (approved):

erikjohnston:
Do we need to worry about this being re-inserted when the old background update is still in the table?

dklimpel:
I think: No. At second run there is no result. The job removes old entries in the database.

erikjohnston:
I was more thinking of the case of us trying to `INSERT` into the `background_updates` an update that is already there, sine I think there is a unique constraint?

dklimpel:
This PR changed the ordering number. In worst case there a running two jobs parallel and creates more load at DB.

dklimpel:
Should the PR also come in version 1.47.0?

anoadragon453:
@dklimpel Indeed. We'll make a v1.47.0rc3 with it included.

PR #11304, opened 2021-11-11
Move 'remove deleted devices' migration to latest schema version

anoadragon453:
This migration file was included as part of https://github.com/matrix-org/synapse/pull/10969 and at the time of the PR's creation, the latest schema version was 64. When the PR was merged, the latest version was 65, however the migration was not moved. This meant that Synapse's on schema version 65 would not run the migration after updating.

This PR moves the migration file to the latest schema, 65. It is OK if this SQL runs twice, as it simply kicks off a background table cleanup process.

Thanks to @dklimpel for spotting it.


erikjohnston:
Is this not a duplicate of https://github.com/matrix-org/synapse/pull/11303

anoadragon453:
> Is this not a duplicate of #11303

Oh bother. Yes, yes it is.

PR #11307, opened 2021-11-11
Add type hints to some storage classes

clokep:
This adds type hints to some of the simpler storage classes.


synapse/storage/util/id_generators.py
```diff
@@ -87,7 +89,25 @@ def _load_current_id(
     return (max if step > 0 else min)(current_id, step)
 
 
-class StreamIdGenerator:
+class AbstractStreamIdGenerator(metaclass=abc.ABCMeta):
```
clokep, line 92: Not sure how I feel about this, but it seems useful to have a shared super-class for `StreamIdGenerator` and `MultiWriterIdGenerator`?

synapse/storage/databases/main/deviceinbox.py
```diff
@@ -101,6 +119,8 @@ def __init__(self, database: DatabasePool, db_conn, hs: "HomeServer"):
 
     def process_replication_rows(self, stream_name, instance_name, token, rows):
         if stream_name == ToDeviceStream.NAME:
+            # If replication is happening than postgres must be being used.
+            assert isinstance(self._device_inbox_id_gen, MultiWriterIdGenerator)
```
clokep, line 123: This is rather unfortunate since this code can only be called if replication is in use, but not sure if there's a "better" way to do it.

synapse/storage/util/id_generators.py
squahtx: It's definitely more convenient for type hinting than writing `Union[StreamIdGenerator, MultiWriterIdGenerator]` everywhere

synapse/storage/databases/main/deviceinbox.py
squahtx: I can't think of a better way either!

squahtx (approved):
LGTM!

synapse/storage/util/id_generators.py
clokep: > It's definitely more convenient for type hinting than writing `Union[StreamIdGenerator, MultiWriterIdGenerator]` everywhere

Oh I should see where we do that an update it I guess...

clokep: I couldn't find any instances of this, although there's some of `Union[StreamIdGenerator, SlavedIdTracker]`.

PR #11310, opened 2021-11-11
Correct type hint for room_batch.py

DMRobertson:
#10975 introduce a function of the form

```
async def foo(*args, **kwargs) -> Optional[str]:
    {
        await expression
    }
```

This is legit syntax (it's building a set literal) but unusual (see `python -c "from __future__ import braces"`). The function as written will always return None. We don't inspect the return value of this function at its only call site, so it's safe to remove the braces and declare a return type of `None`.

### DIFF
--- a/changelog.d/11310.misc
+++ b/changelog.d/11310.misc
@@ -1,1 +1,1 @@
-Correct a type annotation in `room_batches.py`.
+Add type hints to storage classes.
--- a/changelog.d/11310.misc
+++ b/changelog.d/11310.misc
@@ -1,0 +1,1 @@
+Correct a type annotation in `room_batches.py`.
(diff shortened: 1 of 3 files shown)


clokep (approved):
LGTM if CI passes.

changelog.d/11310.misc
```diff
@@ -0,0 +1 @@
+Correct a type annotation in `room_batches.py`.
```
clokep, line 1: Maybe use the same as #11307?

```suggestion
Add type hints to storage classes.
```

DMRobertson: Ahh, good idea. Thank you!

synapse/storage/databases/main/room_batch.py
```diff
@@ -39,13 +39,11 @@ async def get_insertion_event_id_by_batch_id(
...
-            )
-        }
+    ) -> None:
+        await self.db_pool.simple_upsert(
+            table="event_to_state_groups",
+            keyvalues={"event_id": event_id},
+            values={"state_group": state_group_id, "event_id": event_id},
+            # Unique constraint on event_id so we don't have to lock
+            lock=False,
+        )
```
MadLittleMods, line 49: Looks reasonable 👍 

Thanks for fixing it up @DMRobertson 🙂

PR #11312, opened 2021-11-11
Get db signatures file to pass mypy

DMRobertson:
Introduce an intermediate variable so we don't replace `Dict[str, bytes]` with `Dict[str, str]`.

### DIFF
--- a/synapse/events/builder.py
+++ b/synapse/events/builder.py
@@ -128,14 +128,12 @@ class EventBuilder:
             )
 
         format_version = self.room_version.event_format
+        # The types of auth/prev events changes between event versions.
+        prev_events: Union[List[str], List[Tuple[str, Dict[str, str]]]]
+        auth_events: Union[List[str], List[Tuple[str, Dict[str, str]]]]
         if format_version == EventFormatVersions.V1:
-            # The types of auth/prev events changes between event versions.
-            auth_events: Union[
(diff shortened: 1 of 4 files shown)


synapse/events/builder.py
```diff
@@ -128,14 +128,12 @@ async def build(
...
         if format_version == EventFormatVersions.V1:
-            # The types of auth/prev events changes between event versions.
-            auth_events: Union[
-                List[str], List[Tuple[str, Dict[str, str]]]
-            ] = await self._store.add_event_hashes(auth_event_ids)
-            prev_events: Union[
-                List[str], List[Tuple[str, Dict[str, str]]]
-            ] = await self._store.add_event_hashes(prev_event_ids)
+            auth_events = await self._store.add_event_hashes(auth_event_ids)
+            prev_events = await self._store.add_event_hashes(prev_event_ids)
```
clokep, line 136: Is this changing anything or just moving the types up?

DMRobertson: Just moving the types up. I was confused because `add_event_hashes` returns the `List[Tuple[...]]`, which I thought made the union unnecessary. But in the `else` branch, both event lists are `List[str]`. I hoped that moving it up would make this clearer.

clokep (approved):

DMRobertson:
Cheers!

PR #11313, opened 2021-11-11
Annotations for user_erasure_store

DMRobertson:
I'm not sure why this was excluded---it seemed to be passing for me. But
it's easy enough to fixup.

### DIFF
--- a/synapse/storage/databases/main/user_erasure_store.py
+++ b/synapse/storage/databases/main/user_erasure_store.py
@@ -14,11 +14,12 @@
 
 from typing import Dict, Iterable
 
-from synapse.storage._base import SQLBaseStore
+from synapse.storage.database import LoggingTransaction
+from synapse.storage.databases.main import CacheInvalidationWorkerStore
 from synapse.util.caches.descriptors import cached, cachedList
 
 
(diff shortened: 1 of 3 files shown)


synapse/storage/databases/main/user_erasure_store.py
```diff
@@ -14,11 +14,12 @@
...
 from typing import Dict, Iterable
 
-from synapse.storage._base import SQLBaseStore
+from synapse.storage.database import LoggingTransaction
+from synapse.storage.databases.main import CacheInvalidationWorkerStore
 from synapse.util.caches.descriptors import cached, cachedList
 
 
-class UserErasureWorkerStore(SQLBaseStore):
+class UserErasureWorkerStore(CacheInvalidationWorkerStore):
```
clokep, line 22: I'm not sure about this change, is this due to it calling methods from that?

DMRobertson: Yes, `_invalidate_cache_and_stream` is the method in question. Sorry, should have written that down.

clokep (approved):
I think this should be OK.

DMRobertson:
Merge commit to resolve mypy.ini.

PR #11314, opened 2021-11-11
Attempt to annotate events_forward_extremities

DMRobertson:
It was particularly painful to decipher why the DataStore mro was inconsistent.

### DIFF
--- a/synapse/storage/databases/main/events_forward_extremities.py
+++ b/synapse/storage/databases/main/events_forward_extremities.py
@@ -13,15 +13,20 @@
 # limitations under the License.
 
 import logging
-from typing import Dict, List
+from typing import Any, Dict, List
 
 from synapse.api.errors import SynapseError
-from synapse.storage._base import SQLBaseStore
+from synapse.storage.database import LoggingTransaction
+from synapse.storage.databases.main import CacheInvalidationWorkerStore
+from synapse.storage.databases.main.event_federation import EventFederationWorkerStore
 
 logger = logging.getLogger(__name__)
 
 
-class EventForwardExtremitiesStore(SQLBaseStore):
+class EventForwardExtremitiesStore(
+    EventFederationWorkerStore,
+    CacheInvalidationWorkerStore,
+):
     async def delete_forward_extremities_for_room(self, room_id: str) -> int:
         """Delete any extra forward extremities for a room.
 
@@ -31,7 +36,7 @@ class EventForwardExtremitiesStore(SQLBaseStore):
         Returns count deleted.
         """
 
-        def delete_forward_extremities_for_room_txn(txn):
+        def delete_forward_extremities_for_room_txn(txn: LoggingTransaction) -> int:
             # First we need to get the event_id to not delete
             sql = """
                 SELECT event_id FROM event_forward_extremities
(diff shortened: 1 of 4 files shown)


synapse/storage/databases/main/events_forward_extremities.py
```diff
@@ -13,15 +13,20 @@
...
+from synapse.storage.databases.main.event_federation import EventFederationWorkerStore
 
 logger = logging.getLogger(__name__)
 
 
-class EventForwardExtremitiesStore(SQLBaseStore):
+class EventForwardExtremitiesStore(
+    EventFederationWorkerStore,
+    CacheInvalidationWorkerStore,
+):
```
clokep, line 29: Any concerns this could introduce bugs? The only way I could see that happening is if we're shadowing methods, but I *really* hope we're not doing that unless it is a worker vs. non-worker version.

clokep (approved):

synapse/storage/databases/main/events_forward_extremities.py
DMRobertson: I was scared of doing this, but I'm scared of the storage classes as a whole.

This is a very reasonable question that I don't know how to answer easily.

EventFederationWorkerStore has the following methods. Those without notes aren't defined anywhere else.
  - `get_auth_chain`
  - `get_auth_chain_ids`
  - `_get_auth_chain_ids_using_cover_index_txn`
  - `_get_auth_chain_ids_txn`
  - `get_auth_chain_difference`
     - is also defined on `StateResolutionStore` and is defined on `TestStateResolutionStore`. I think these are wrappers rather than storage mixins. (?)
   - `_get_auth_chain_difference_using_cover_index_txn`
   - `_get_auth_chain_difference_txn`
   - `get_oldest_event_ids_with_depth_in_room`
   - `get_insertion_event_backwards_extremities_in_room`
   - `get_max_depth_of`
   - `get_min_depth_of`
   - `get_prev_events_for_room`
   - `_get_prev_events_for_room_txn`
   - `_get_prev_events_for_room_txn`
   - `get_latest_event_ids_in_room`
   - `get_min_depth`
   - `_get_min_depth_interaction`
   - `get_forward_extremities_for_room_at_stream_ordering`
   - `_get_forward_extremeties_for_room`
   - `get_backfill_events`
   - `_get_backfill_events`
   - `get_missing_events`
     - defined on federation client and transport, but not any other store
   -  `_get_missing_events`
   - `get_successor_events`
   - `_delete_old_forward_extrem_cache`
   - `insert_insertion_extremity`
   - `insert_received_event_to_staging`
   - `remove_received_event_from_staging`
   - `get_next_staged_event_id_for_room`
   - `get_next_staged_event_for_room`
   - `prune_staged_events_in_room`
   - `get_all_rooms_with_staged_incoming_events`
   - `_get_stats_for_federation_staging`

So the only shadowing is by a wrapper rather than a mixin. I _think_ this is safe? But I'm mainly relying on CI not to explode to be honest.

CacheInvalidationWorkerStore to follow.

DMRobertson: `CacheInvalidationWorkerStore` has the following methods. Those without notes aren't defined elsewhere.

- `get_all_updated_caches`
- `process_replication_rows` defined on
  - `CacheInvalidationWorkerStore`
  - `DeviceInboxWorkerStore`
  - `EventsWorkerStore`
  - `PrescenceStore`
  - `ReceiptsWorkerStore`
  - each of these ends with a call to the same method on the `super()`class, so I think this will traverse the whole MRO. I don't know how we can enforce that though.
- `_process_event_stream_row`
- `_invalidate_caches_for_event`
- `invalidate_cache_and_stream`
- `_invalidate_cache_and_stream`
- `_invalidate_all_cache_and_stream`
- `_invalidate_state_caches_and_stream`
- `_send_invalidation_to_replication`
- `get_cache_stream_token_for_writer`

Given that every `process_replication_rows` implementation calls the superclass will be safe provided that `EventForwardExtremitiesStore`'s MRO is consistent with that of `DataStore` and any other top-level Stores. If the latter was not the case, we'd get a big TypeError at import time when trying to define the DataStore type.

clokep: Thanks for the thorough review! 👍 I agree with your assessment!

DMRobertson: I suspect we might need to do this again if I keep pulling in superclasses like this. Maybe there's some way to automate this process?

DMRobertson:
The merge defers to the mypy config added in https://github.com/matrix-org/synapse/commit/95547e53007be2840c78f1c45b0c5a2ea8dcb410.

PR #11316, opened 2021-11-11
Annotations for state_deltas.py

DMRobertson:
I was sad that I couldn't do better for
`_curr_state_delta_stream_cache`. At least it's explicitly called out in
a comment with #TODO.

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -51,7 +51,6 @@ exclude = (?x)
    |synapse/storage/databases/main/search.py
    |synapse/storage/databases/main/signatures.py
    |synapse/storage/databases/main/state.py
-   |synapse/storage/databases/main/state_deltas.py
    |synapse/storage/databases/main/stats.py
    |synapse/storage/databases/main/transactions.py
    |synapse/storage/databases/main/user_directory.py
@@ -181,6 +180,9 @@ disallow_untyped_defs = True
 disallow_untyped_defs = True
 
 [mypy-synapse.storage.databases.main.client_ips]
+disallow_untyped_defs = True
+
+[mypy-synapse.storage.databases.main.state_deltas]
 disallow_untyped_defs = True
 
 [mypy-synapse.storage.util.*]
--- a/synapse/storage/databases/main/state_deltas.py
+++ b/synapse/storage/databases/main/state_deltas.py
@@ -16,11 +16,17 @@ import logging
 from typing import Any, Dict, List, Tuple
 
 from synapse.storage._base import SQLBaseStore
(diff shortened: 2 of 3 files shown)


mypy.ini
```diff
@@ -183,6 +182,9 @@ disallow_untyped_defs = True
 [mypy-synapse.storage.databases.main.room_batch]
 disallow_untyped_defs = True
 
+[mypy-synapse.storage.databases.main.state_deltas]
+disallow_untyped_defs = True
```
clokep, line 186: See #11322 instead. 😄

synapse/storage/databases/main/state_deltas.py
```diff
@@ -16,11 +16,17 @@
...
+from synapse.storage.database import LoggingTransaction
+from synapse.util.caches.stream_change_cache import StreamChangeCache
 
 logger = logging.getLogger(__name__)
 
 
 class StateDeltasStore(SQLBaseStore):
+    # This class must be mixed in with a child class which provides the following
+    # attribute. TODO: can we get static analysis to enforce this?
+    _curr_state_delta_stream_cache: StreamChangeCache
```
reivilibre, line 28: Can we make this an abstract class perhaps, with this being an abstract.. field... if that exists ...

It doesn't seem like it does, but I would be tempted to try an abstract property:

```python
    @property
    @abstractmethod
    def _curr_state_delta_stream_cache(self) -> StreamChangeCache:
        ...
```

Not super clean, but curious if that'd do the trick (and if it's allowed to just overwrite it with a field later)

reivilibre (approved):
Hacky abstract thing is up to you, everything else LGTM (though the mypy.ini change is soon going to be obsolete one way or the other ...)

synapse/storage/databases/main/state_deltas.py
DMRobertson: I got

```
synapse/app/generic_worker.py:222: error: Definition of "_curr_state_delta_stream_cache" in base class "StateDeltasStore" is incompatible with definition in base class "SlavedEventStore"  [misc]
```

after

```
class StateDeltasStore(SQLBaseStore, abc.ABC):
    @property
    @abc.abstractmethod
    def _curr_state_delta_stream_cache(self) -> StreamChangeCache:
        """This class must be mixed in with a child class which provides the
        following attribute."""
        ...
```

DMRobertson: You might be interested in the discussion here: https://github.com/python/mypy/issues/4426

PR #11322, opened 2021-11-12
Generalize the disallowed_untyped_defs in mypy.ini

clokep:
Instead of including specific files we include full modules.

The main benefit to this is that as we removed ignored files from mypy we require all the type hints in those files.

### DIFF
--- a/changelog.d/11322.misc
+++ b/changelog.d/11322.misc
@@ -1,0 +1,1 @@
+Add type hints to storage classes.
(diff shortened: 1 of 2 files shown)


reivilibre:
I suspect this conflicts with #11302

reivilibre (approved):
Happy if CI is happy, though do note that it conflicts with #11302 (that said, no reason we can't just merge this and switch over to #11302 if it suits)

clokep:
> I suspect this conflicts with #11302

And #11315, I just hadn't gotten around to commenting on that yet...

Issue #11319, opened 2021-11-12
DELETE /directory/rooms/ALIAS doesn't handle the case of a nonexistant alias


DMRobertson:
[Spec](https://matrix.org/docs/spec/client_server/r0.6.1#delete-matrix-client-r0-directory-room-roomalias) says we should return 404.

AFAICS we only return 200 (or maybe 500 if there's an internal server error).

https://github.com/matrix-org/synapse/blob/2af6d31b78109a989e27128ac655990c35b29d62/synapse/rest/client/directory.py#L94-L124

Discovered because mypy spotted the return type hint here was wrong (we may return `None`):

https://github.com/matrix-org/synapse/blob/40a1fddd1b7843c25248fca409debab6d8393c59/synapse/storage/databases/main/directory.py#L136-L159

DMRobertson:
Well, having written a test for this, it's working perfectly fine. But I have no idea why; I would like to understand,

DMRobertson:
Oh. It's here.

https://github.com/matrix-org/synapse/blob/75ca0a6168f92dab3255839cf85fb0df3a0076c3/synapse/handlers/directory.py#L188-L193

Well, this was a load of hot air. Closing.

PR #11327, opened 2021-11-12
Test room alias deletion

DMRobertson:
I thought there was a bug in #11319  by my reading of the source code. I learned that my reading was incomplete after I wrote a test case to try and demonstrate the bug. We may as well keep that test case around. While I'm at it, I did some typing too.

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -282,6 +282,9 @@ disallow_untyped_defs = True
 [mypy-tests.storage.test_user_directory]
 disallow_untyped_defs = True
 
+[mypy-tests.rest.client.test_directory]
+disallow_untyped_defs = True
+
 ;; Dependencies without annotations
 ;; Before ignoring a module, check to see if type stubs are available.
 ;; The `typeshed` project maintains stubs here:
--- a/synapse/handlers/directory.py
+++ b/synapse/handlers/directory.py
@@ -204,6 +204,10 @@ class DirectoryHandler:
             )
 
         room_id = await self._delete_association(room_alias)
+        # The call to `_user_can_delete_alias` above ensures the alias exists.
+        # `_delete_association` returns `None` only if the alias doesn't exist.
+        # We leave the assert here for mypy's benefit.
+        assert room_id is not None
 
         try:
             await self._update_canonical_alias(requester, user_id, room_id, room_alias)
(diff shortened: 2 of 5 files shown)


synapse/handlers/directory.py
```diff
@@ -204,6 +204,10 @@ async def delete_association(
             )
 
         room_id = await self._delete_association(room_alias)
+        # The call to `_user_can_delete_alias` above ensures the alias exists.
+        # `_delete_association` returns `None` only if the alias doesn't exist.
+        # We leave the assert here for mypy's benefit.
+        assert room_id is not None
```
clokep, line 210: Isn't this actually hiding a race condition?

synapse/storage/databases/main/directory.py
```diff
@@ -126,7 +126,7 @@ def alias_txn(txn):
 
 
 class DirectoryStore(DirectoryWorkerStore):
-    async def delete_room_alias(self, room_alias: RoomAlias) -> str:
+    async def delete_room_alias(self, room_alias: RoomAlias) -> Optional[str]:
```
clokep, line 129: Does `_delete_room_alias_txn` need to match this?

synapse/handlers/directory.py
DMRobertson: Thanks, you're right. Maybe something deletes the alias after we get `can_delete: True` but before `_delete_association` finishes.

DMRobertson:
Hmm. Failing the test run on olddeps. Not sure why.

mypy.ini
```diff
@@ -282,6 +282,9 @@ disallow_untyped_defs = True
 [mypy-tests.storage.test_user_directory]
 disallow_untyped_defs = True
 
+[mypy-tests.rest.client.test_directory]
+disallow_untyped_defs = True
```
clokep, line 286: I think this is unnecessary with #11322?

clokep: OH NO! That had a bug in it....

tests/rest/client/test_directory.py
```diff
@@ -32,15 +36,19 @@ class DirectoryTestCase(unittest.HomeserverTestCase):
...
-    def make_homeserver(self, reactor, clock):
+    def make_homeserver(self, reactor: MemoryReactor, clock: Clock) -> HomeServer:
         config = self.default_config()
         config["require_membership_for_aliases"] = True
 
         self.hs = self.setup_test_homeserver(config=config)
 
         return self.hs
 
-    def prepare(self, reactor, clock, homeserver):
```
clokep, line 43: FWIW It might be easy to do a PR that modifies this (and `make_homeserver`) with `sed` + `black`. Not sure if it would make things pass or not though.

mypy.ini
DMRobertson: missing a `.*` I think? Ouch!

clokep (approved):

Issue #10946, opened 2021-09-29
Stabilize support for MSC2946: spaces summary


clokep:
Once [MSC2946](https://github.com/matrix-org/matrix-doc/pull/2946) lands we should (at some point) switch to a stable endpoint and make some updates. There's few bits to this that we need to do:

## Phase 1

* [x] Support to the stable endpoint.
* [x] ~~Require `limit` to be > 0.~~ (This has always been correct.)
* [x] Ensure the endpoint is ratelimited.

## Phase 2

* [x] Remove the fallback logic of `get_room_hierarchy` to request the `/spaces` endpoint over federation.
* [x] Remove the `/spaces` client and federation endpoints (and corresponding code in the handlers).
* [x] Update the API responses to remove `allowed_spaces` (this was replaced with `allowed_room_ids`).
* [x] Clean-up code around max-children, which is not used by the `/hierarchy` endpoint.
* [x] Remove `room_id` from the serialized `children_state`, it is redundant and removed from the API response.

## Phase 3

* [x] Remove unstable `/hierarchy` client and federation endpoints.
* [x] Remove the typo `join_rules` field from the response.

clokep:
I think this work needs to be split into a few pieces.

Phase 1:

* Support the stable endpoint.
* Remove some of the legacy code (the `/spaces` endpoint? and surrounding bits?)
* Enforce the `limit` / ratelimiting.

Phase 2:

* Remove the unstable `/hierarchy` endpoint.

clokep:
> Remove some of the legacy code (the `/spaces` endpoint? and surrounding bits?)

It looks like these endpoints still get some usage on matrix.org, so might be best to just leave them for now and remove all the unstable code at once.

clokep:
Client bugs for using the stable endpoint:

* vector-im/element-ios#5144
* vector-im/element-android#4462
* vector-im/element-web#19738

clokep:
For phase 2 it is useful to know which version of clients first started using (unstable or stable) `/hierarchy` instead of `/spaces`.

* Element Android: v1.2.1 (vector-im/element-android#3926)
* Element iOS: v1.6.9 (https://github.com/vector-im/element-ios/blob/v1.6.9/Podfile#L16 via matrix-ios-sdk v0.20.11 -- matrix-org/matrix-ios-sdk#1259)
* matrix-js-sdk v12.4.0 (matrix-org/matrix-js-sdk#1818):
[...]

Issue #5118, opened 2019-05-01
we should optimise rest endpoint dispatch


richvdh:
every incoming GET request gets compared against a list of 227 regexps, many of which are redundant since they duplicate each endpoint onto the various places the client API lives. This is surely an unnecessary overhead.

erikjohnston:
Something like https://github.com/nitely/kua would probably work

richvdh:
possibly, though I'm not sure how it would tie into twisted.

I was mostly just imagining that it would be a good start to say: for the ClientRestResource which is mounted at `/_matrix/client/r0`, only register regexes which will actually match that prefix, rather than all of the `/_matrix/client/api/v1` etc regexes.

clokep:
> every incoming GET request gets compared against a list of 227 regexps

I was a bit curious how this compares to current day:

| Method | Endpoints |
|----|----|
| `GET` | 232 |
| `PUT` | 129 |
| `POST` | 169 |
| `DELETE` | 45 |

I was pretty easily able to reduce this quite a bit by messing with the `client_patterns` to create a single regular expression instead of multiple patterns. I'm not sure if that's a worthwhile optimization or if we need an entire new strategy. Anyway the results were:

| Method | Endpoints |
|----|----|
| `GET` | 80 |
| `PUT` | 39 |
| `POST` | 66 |
| `DELETE` | 15 |

(These were for the `/_matrix/client` endpoints using the demo server. so exact endpoints could vary a bit based on configuration.)

PR #11329, opened 2021-11-12
Support the stable endpoint from MSC2946

clokep:
[MSC2946](https://github.com/matrix-org/matrix-doc/pull/2946) passed FCP and we should now support the stable endpoint to allow clients to use it.

This work is tracked in #10946, but this PR only does the first half of it: adding the stable endpoint and minor bugfixes based on changes to the MSC.


synapse/rest/client/room.py
```diff
@@ -1166,7 +1166,7 @@ async def on_GET(
             )
 
         limit = parse_integer(request, "limit")
-        if limit is not None and limit <= 0:
```
DMRobertson, line 1165: I understand "positive" to mean "> 0".

If passing `limit=0` constitutes a valid request, I'd suggest the error message should say "non-negative" (i.e. >=0) instead of "positive".

The terminology is a little awkward, but standard AFAIK.

DMRobertson (approved):
Seems reasonable! Note for completeness that the corresponding MSC was merged back a fortnight ago.

synapse/rest/client/room.py
DMRobertson: Agreed. MSC says

> * **`limit`**: Optional: a client-defined limit to the maximum number of rooms to return per page. Must an integer greater than zero.

So it was correct before.

clokep: I'll back out that commit, must have confused myself at some point!

synapse/handlers/room_summary.py
```diff
@@ -276,6 +277,8 @@ async def get_room_hierarchy(
         Returns:
             The JSON hierarchy dictionary.
         """
+        await self._request_ratelimiter.ratelimit(requester)
```
clokep, line 280: I'm not sure if this is the correct ratelimiter to be using or not? It seems most of our other requests use a custom rate limiter instead of the request ratelimiter?

DMRobertson (approved):
LGTM!

clokep:
I should check if we need to update any tests here.

synapse/handlers/room_summary.py
DMRobertson: Not sure here---might be worth another pair of eyes?

clokep:
I'm converting this to a draft -- when setting the complement tests to use the new endpoint they were failing. I want to investigate before merging this.

synapse/app/homeserver.py
```diff
@@ -194,6 +194,7 @@ def _configure_named_resource(
                 {
                     "/_matrix/client/api/v1": client_resource,
                     "/_matrix/client/r0": client_resource,
+                    "/_matrix/client/v1": client_resource,
```
clokep, line 197: For the generic worker it seems we just put `client_resource` under `/_matrix/client` once. I'm not sure why we don't do that here too?

synapse/federation/federation_client.py
```diff
@@ -1395,11 +1395,28 @@ async def get_room_hierarchy(
...
+
+                logger.debug(
+                    "Couldn't fetch room hierarchy with the v1 API, falling back to the unstable API"
+                )
+
+                res = await self.transport_layer.get_room_hierarchy_unstable(
+                    destination=destination,
+                    room_id=room_id,
+                    suggested_only=suggested_only,
+                )
```
clokep, line 1419: This is the same pattern we do with e.g. `send_join`.

synapse/federation/transport/client.py
```diff
@@ -1192,10 +1192,24 @@ async def get_space_summary(
         )
 
     async def get_room_hierarchy(
-        self,
-        destination: str,
-        room_id: str,
-        suggested_only: bool,
+        self, destination: str, room_id: str, suggested_only: bool
+    ) -> JsonDict:
```
clokep, line 1196: We could choose whether to v1 vs. unstable based on a parameter but the other bits in this file seem to have separate methods for v1 vs. v2, etc. so I followed that pattern. I think we can clean that up separately if it makes sense?

clokep:
See updated complement tests: matrix-org/complement#225.

I've also prepped a branch that finishes off #10946 and double checked that tests still pass.

.github/workflows/tests.yml
```diff
@@ -374,7 +374,7 @@ jobs:
         working-directory: complement/dockerfiles
 
       # Run Complement
-      - run: go test -v -tags synapse_blacklist,msc2403,msc2946,msc3083 ./tests/...
```
DMRobertson, line 377: There's duplication here with `complement.sh`. (The latter of which now has a merge conflict, I think).

Is there some way to make both of these places read the excluded tags from a single file?

Possibly out-of-scope for this change.

docs/workers.md
```diff
@@ -210,7 +210,7 @@ expressions:
     ^/_matrix/federation/v1/get_groups_publicised$
     ^/_matrix/key/v2/query
     ^/_matrix/federation/unstable/org.matrix.msc2946/spaces/
-    ^/_matrix/federation/unstable/org.matrix.msc2946/hierarchy/
+    ^/_matrix/federation/(v1|unstable/org.matrix.msc2946)/hierarchy/
```
DMRobertson, line 213: Does this want to also say "v3" now that Matrix 1.1 is specced?

docs/workers.md
```diff
@@ -223,7 +223,7 @@ expressions:
     ^/_matrix/client/(api/v1|r0|v3|unstable)/rooms/.*/members$
     ^/_matrix/client/(api/v1|r0|v3|unstable)/rooms/.*/state$
     ^/_matrix/client/unstable/org.matrix.msc2946/rooms/.*/spaces$
-    ^/_matrix/client/unstable/org.matrix.msc2946/rooms/.*/hierarchy$
+    ^/_matrix/client/(v1|unstable/org.matrix.msc2946)/rooms/.*/hierarchy$
```
DMRobertson, line 226: ditto

synapse/rest/client/room.py
```diff
@@ -1138,11 +1138,15 @@ async def on_POST(
...
 class RoomHierarchyRestServlet(RestServlet):
-    PATTERNS = (
+    PATTERNS = [
         re.compile(
             "^/_matrix/client/unstable/org.matrix.msc2946"
             "/rooms/(?P<room_id>[^/]*)/hierarchy$"
         ),
+    ] + list(
+        client_patterns(
+            "/rooms/(?P<room_id>[^/]*)/hierarchy$", releases=("v1",), unstable=False
```
DMRobertson, line 1148: do we want `v3` here?

DMRobertson (commented):
Also seems reasonable after the rebase! I would like input from @matrix-org/synapse-core re https://github.com/matrix-org/synapse/pull/11329/files#r749355348 though.

DMRobertson (changes requested):
At the very least, that merge conflict needs resolving though!

.github/workflows/tests.yml
clokep: It is really annoying, see matrix-org/complement#185. I don't have good ideas how to solve it, but I think it is out of scope of this PR.

docs/workers.md
clokep: No, new endpoints star at `v1` (note that this matches what is in the MSC also).

clokep: Also federation endpoints have always been versioned per-endpoint so have always started at `v1` (they never had `r0` paths).

docs/workers.md
clokep: No, new endpoints start at `v1` (note that this matches what is in the MSC also).

synapse/rest/client/room.py
clokep: No, see above.

.github/workflows/tests.yml
DMRobertson: Glad to see we're both annoyed by this! Agreed, let's leave it.

synapse/app/homeserver.py
richvdh: I think it's because of `WEB_CLIENT_PREFIX`, down at line 274.

richvdh: (which is not to say there aren't way better ways of doing this and I would love it if this stuff were less of a mess. see also https://github.com/matrix-org/synapse/issues/5118)

synapse/handlers/room_summary.py
richvdh: It's not really appropriate, because `get_request_ratelimiter` is configured via the `rc_message` config settings, which are about the number of messages that can be sent.

Let's not introduce another set of config settings just for this, though. Either hardcode it for now or introduce some sort of "generic request" ratelimit setting.

synapse/rest/client/room.py
```diff
@@ -1138,11 +1138,15 @@ async def on_POST(
...
+        )
     )
```
richvdh, line 1150: ugh, I hate client_patterns. this is shorter and clearer to write out longhand.

clokep: I'd prefer to tackle that separately if we wish to change it instead of tacking it onto an unrelated PR.

richvdh: it's not an unrelated PR - you're adding a new use of `client_patterns` when it would be shorter and clearer just to write it out.

But I'll not block this PR on it if you disagree.

clokep: Do we have it documented that we should not be using `client_patterns` for new code? This is the first I've heard of this.

richvdh: > Do we have it documented that we should not be using client_patterns for new code?

That's not *quite* what I said - there may be times it's useful. What I *am* saying is that this:

```python
    PATTERNS = (
        re.compile(
            "^/_matrix/client/unstable/org.matrix.msc2946"
            "/rooms/(?P<room_id>[^/]*)/hierarchy$"
        ),
        re.compile(
            "^/_matrix/client/v1/rooms/(?P<room_id>[^/]*)/hierarchy$"
        ),
    )
```

is clearer and shorter than:

```python
    PATTERNS = [
        re.compile(
            "^/_matrix/client/unstable/org.matrix.msc2946"
            "/rooms/(?P<room_id>[^/]*)/hierarchy$"
        ),
    ] + list(
        client_patterns(
            "/rooms/(?P<room_id>[^/]*)/hierarchy$", releases=("v1",), unstable=False
        )
    )
```

... ie, I don't really feel like `client_patterns` is giving any value here. Just because it exists doesn't mean we should shoehorn it in everywhere we can.

Still, I'll not object if someone wants to go and rid the codebase of it :)

richvdh: > you can easily use client_patterns, which is the standard way we describe these URLs.

well, sometimes.

>  I would find it confusing to switch to client_patterns when removing the unstable API.

I would certainly not recommend doing that.

synapse/handlers/room_summary.py
clokep: I went with the hard-coded settings for now (that are fairly high).

clokep:
@DMRobertson Can you take another look over this?

synapse/rest/client/room.py
```diff
@@ -1140,14 +1140,10 @@ async def on_POST(
 class RoomHierarchyRestServlet(RestServlet):
     PATTERNS = [
         re.compile(
-            "^/_matrix/client/unstable/org.matrix.msc2946"
+            "^/_matrix/client/(v1|unstable/org.matrix.msc2946)"
```
DMRobertson, line 1143: Paranoid: do we refer to capture groups by their index anywhere? The `P<room_id>` makes it look like we always use named capture groups so probably not.

As a sanity check, I tried grepping for `\|unstable`, but didn't find any matches outside of `configure_workers_and_start`. I guess this is the CLIENT_PATTERNS stuff discussed here and elsewhere. (I don't have an opinion on that and am happy to go with the consensus.)

synapse/handlers/room_summary.py
```diff
@@ -93,7 +94,9 @@ def __init__(self, hs: "HomeServer"):
         self._event_serializer = hs.get_event_client_serializer()
         self._server_name = hs.hostname
         self._federation_client = hs.get_federation_client()
-        self._request_ratelimiter = hs.get_request_ratelimiter()
+        self._ratelimiter = Ratelimiter(
+            store=self._store, clock=hs.get_clock(), rate_hz=5, burst_count=10
```
DMRobertson, line 98: Where did these numbers come from, out of interest?

clokep: I made them up, they should allow for 5 queries per second without enforcing any ratelimiting until 10 queries. This might be a bit high, but I'm never quite sure what to put here and usually err on the side of allowing more. 🤷

synapse/rest/client/room.py
clokep: Good question! I wasn't entirely sure, but we end up only passing on [`groupdict()`](https://docs.python.org/3/library/re.html#re.Match.groupdict) from the resulting `Match` object, which does only pass *named* subgroups.

See https://github.com/matrix-org/synapse/blob/b64b6d12d47c012a9996a2585a969d3b962d25fa/synapse/http/server.py#L419-L425

clokep: (I can change it to the suggestion of two different URL paths if you'd prefer, which seems to be what we do in other places. I think there's a benefit to having it match the docs too though!)

DMRobertson: > (I can change it to the suggestion of two different URL paths if you'd prefer, which seems to be what we do in other places. I think there's a benefit to having it match the docs too though!)

I think it's fine as it is, and matching docs is a nice-to-have!

synapse/handlers/room_summary.py
DMRobertson: Fair enough. We can always tighten the limits!

(Though there's a part of me that wonders if rate limiting ought to be handled by the load balancer to keep Synapse itself simpler.)

clokep: Some rate limits can be tweaked by the Synapse config, FWIW. They can also depend on particular data, e.g. you could limit the message rate per user, per room, per message type (or something like that).

synapse/handlers/room_summary.py
DMRobertson: >  It's not really appropriate, because get_request_ratelimiter is configured via the rc_message config settings, which are about the number of messages that can be sent.

Maybe a rename would help express this? (Messages as in to_device messages or events?) Not one for here though.

DMRobertson (approved):
LGTM!

docs/workers.md
```diff
@@ -210,7 +210,7 @@ expressions:
     ^/_matrix/federation/v1/get_groups_publicised$
     ^/_matrix/key/v2/query
     ^/_matrix/federation/unstable/org.matrix.msc2946/spaces/
```
daenney, line 212: Shouldn't this one get `v1` too in that case?

clokep: No, this endpoint was not stabilized and will be removed in the future.

daenney: Cool, thanks!

clokep: No problem! It was still being hit fairly actively on matrix.org last time I checked unfortunately. It was essentially replaced with the `/hierarchy` version.

#10946 tracks this work for cross-referencing.

PR #11330, opened 2021-11-12
Misc typing fixes for tests, part 2 of N

DMRobertson:
I haven't included the mypy config that prompted these errors because there are too many. But I promise this fixed errors!

### DIFF
--- a/tests/rest/client/utils.py
+++ b/tests/rest/client/utils.py
@@ -19,17 +19,7 @@ import json
 import re
 import time
 import urllib.parse
-from typing import (
-    Any,
-    AnyStr,
-    Dict,
-    Iterable,
-    Mapping,
-    MutableMapping,
-    Optional,
-    Tuple,
-    Union,
-)
+from typing import Any, AnyStr, Dict, Iterable, Mapping, MutableMapping, Optional, Tuple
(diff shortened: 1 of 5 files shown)


DMRobertson:
Sorry, I've included some changes from #11327 here. Will rebase.

tests/rest/client/utils.py
```diff
@@ -631,7 +658,13 @@ def initiate_sso_login(
 
         # hit the redirect url again with the right Host header, which should now issue
         # a cookie and redirect to the SSO provider.
-        location = channel.headers.getRawHeaders("Location")[0]
+        def get_location() -> str:
+            location_values = channel.headers.getRawHeaders("Location")
```
clokep, line 662: Does this properly pull in `channel` again when it gets assigned on line 669? Might be clearer to just take it as an input argument.

DMRobertson: It does, but I had to confirm:

```python
>>> x = "a"
>>> def foo(): print(x)
... 
>>> foo()
a
>>> x = "b"
>>> foo()
b
```

That said, I prefer your clearer approach!

clokep (approved):
Sorry for the delay here!

PR #11331, opened 2021-11-12
Fix case in `wait_for_background_updates` where `self.store` does not exist

anoadragon453:
`self.store` is not guaranteed to exist in every test case. Replace with `self.hs.get_datastore()` as `self.hs` is at least available in every instance of `unittest.HomeserverTestCase`.

Silly little change split out from #11215.


DMRobertson:
Oh _that_ was the bug you teased me with?

(maybe we should just make `self.store` part of the base test case?)

tests/unittest.py
```diff
@@ -334,7 +334,7 @@ def wait_on_thread(self, deferred, timeout=10):
     def wait_for_background_updates(self) -> None:
         """Block until all background database updates have completed."""
         while not self.get_success(
-            self.store.db_pool.updates.has_completed_background_updates()
+            self.hs.get_datastore().db_pool.updates.has_completed_background_updates()
         ):
             self.get_success(
                 self.store.db_pool.updates.do_next_background_update(100), by=0.1
```
clokep, line 340: I think you need to replace it here too, maybe just pull `store` out to a separate variable for this function?

clokep (commented):
I ran into this the other day in #11391, would be great to get this merged!

DMRobertson (approved):
Seems reasonable, assuming CI approves.

Any idea why mypy didn't spot this---I guess this code just isn't tested?

clokep:
> Any idea why mypy didn't spot this---I guess this code just isn't tested?

I think just most of the test code isn't covered by mypy, yeah.

PR #11332, opened 2021-11-12
Rollback #11322 due to wrong syntax in mypy.ini

clokep:
It should have been `[mypy-tests.*]` but that obviously won't work.

### DIFF
--- a/changelog.d/11332.misc
+++ b/changelog.d/11332.misc
@@ -1,0 +1,1 @@
+Add type hints to storage classes.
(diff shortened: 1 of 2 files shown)


DMRobertson (approved):
I almost wish they forced you to either write `foo.__init__` or `foo.*` to clarify your intent.

Can I confirm that this bit of config is as intended?

```
[mypy-synapse.storage]
disallow_untyped_defs = True
```

I.e. do we want all storage modules that pass `mypy` to satisfy `disallow_untyped_defs`, or just `storage.__init__`?

clokep:
> I.e. do we want all storage modules that pass `mypy` to satisfy `disallow_untyped_defs`, or just `storage.__init__`?

We want them all. :(

DMRobertson (approved):
Unfortunate, but we definitely want the checking in!

PR #11340, opened 2021-11-15
Properly register all callback hooks for legacy password authentication providers

reivilibre:
Introduced in https://github.com/matrix-org/synapse/commit/cdd308845ba22fef22a39ed5bf904b438e48b491#diff-54e31652219b9be5b5121f808bd34ccf9fa7d56f45d04ab307e501445c11401aR1951.
Found whilst un-erasing type annotations for the module API.

### DIFF
--- a/changelog.d/11340.bugfix
+++ b/changelog.d/11340.bugfix
@@ -1,1 +1,1 @@
-Fix a bug, introduced in Synapse 1.46.0, which caused some callbacks in legacy password authentication provider modules to not be registered. Modules using the generic module API were not affected.
+Fix a bug, introduced in Synapse 1.46.0, which caused the `check_3pid_auth` and `on_logged_out` callbacks in legacy password authentication provider modules to not be registered. Modules using the generic module API were not affected.
--- a/changelog.d/11340.bugfix
+++ b/changelog.d/11340.bugfix
@@ -1,0 +1,1 @@
+Fix a bug, introduced in Synapse 1.46.0, which caused some callbacks in legacy password authentication provider modules to not be registered. Modules using the generic module API were not affected.
--- a/synapse/handlers/auth.py
+++ b/synapse/handlers/auth.py
@@ -1950,7 +1950,7 @@ def load_single_legacy_password_auth_provider(
         # need to use a tuple here for ("password",) not a list since lists aren't hashable
         auth_checkers[(LoginType.PASSWORD, ("password",))] = check_password
 
-    api.register_password_auth_provider_callbacks(hooks, auth_checkers=auth_checkers)
+    api.register_password_auth_provider_callbacks(auth_checkers=auth_checkers, **hooks)
 
 
 CHECK_3PID_AUTH_CALLBACK = Callable[
--- a/synapse/handlers/auth.py
+++ b/synapse/handlers/auth.py
@@ -1828,13 +1828,6 @@ def load_single_legacy_password_auth_provider(
         logger.error("Error while initializing %r: %s", module, e)
         raise
 
-    # The known hooks. If a module implements a method who's name appears in this set
-    # we'll want to register it
-    password_auth_provider_methods = {
-        "check_3pid_auth",
-        "on_logged_out",
-    }
-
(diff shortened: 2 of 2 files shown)


DMRobertson:
Could we have spotted this with some kind of test?

reivilibre:
@DMRobertson type annotations in #11341 would have spotted it (indeed, that's how it was spotted).
Already been in touch with Brendan about this (since he helped me spot that the code was wrong, oops) and he plans to make an issue about the test coverage of some module API stuff later on today.

changelog.d/11340.bugfix
```diff
@@ -0,0 +1 @@
+Fix a bug, introduced in Synapse 1.46.0, which caused some callbacks in legacy password authentication provider modules to not be registered. Modules using the generic module API were not affected.
```
clokep, line 1: Is there any more information about *which* callbacks? How would a sysadmin know if they were affected?

synapse/handlers/auth.py
```diff
@@ -1950,7 +1950,7 @@ def run(*args: Tuple, **kwargs: Dict) -> Awaitable:
         # need to use a tuple here for ("password",) not a list since lists aren't hashable
         auth_checkers[(LoginType.PASSWORD, ("password",))] = check_password
 
-    api.register_password_auth_provider_callbacks(hooks, auth_checkers=auth_checkers)
+    api.register_password_auth_provider_callbacks(auth_checkers=auth_checkers, **hooks)
```
clokep, line 1953: I'm actually finding this code really hard to reason about. It seems like the abstract `hooks` variable saves like 2 lines of code? Can we just be explicit?

babolivier: We probably can. FWIW I suggested we do it this way for consistency with how we load legacy spam checkers and third-party rule modules, but maybe it's just only relevant there.

babolivier:
> Already been in touch with Brendan about this (since he helped me spot that the code was wrong, oops) and he plans to make an issue about the test coverage of some module API stuff later on today.

Have opened https://github.com/matrix-org/synapse/issues/11351

synapse/handlers/auth.py
reivilibre: Stripped this out now and inlined it. I think it's better this way, though maybe it was justifiable with the sheer number of hooks for the spam checker modules.

changelog.d/11340.bugfix
reivilibre: Added

synapse/handlers/auth.py
babolivier: Yeah that was my point, the third-party rules and spam checker modules have enough callbacks that it probably makes it worth it, but that's not really the case here.

clokep: Definitely fair, but when we come back to this code later we probably won't be looking at these all at once. Thanks for updating it!

synapse/handlers/auth.py
```diff
@@ -1950,7 +1946,11 @@ def run(*args: Tuple, **kwargs: Dict) -> Awaitable:
         # need to use a tuple here for ("password",) not a list since lists aren't hashable
         auth_checkers[(LoginType.PASSWORD, ("password",))] = check_password
 
-    api.register_password_auth_provider_callbacks(hooks, auth_checkers=auth_checkers)
+    api.register_password_auth_provider_callbacks(
+        auth_checkers=auth_checkers,
+        check_3pid_auth=check_3pid_auth_hook,
+        on_logged_out=on_logged_out_hook,
```
clokep, line 1952: Let's keep them the same order as the definition:

```suggestion
        check_3pid_auth=check_3pid_auth_hook,
        on_logged_out=on_logged_out_hook,
        auth_checkers=auth_checkers,
```

clokep (approved):
Seems reasonable!

synapse/handlers/auth.py
babolivier: Yeah I definitely agree that it's better without a dict here, I was just trying to provide additional context.

PR #11341, opened 2021-11-15
Add/Unerase annotations to Module API

reivilibre:
This would have caught #11340 (which will fix the type error here, so this is dependent on that).

The root issue is that `@property` requires the function to be annotated if it is going to expose a type annotation.
But writing a concrete type for these functions is not possible with the tools currently given to us (at least, apart from Mypy-specific typing extensions).

I thought expanding the functions to call through made more sense.


synapse/module_api/__init__.py
```diff
@@ -114,7 +150,7 @@ class ModuleApi:
     can register new users etc if necessary.
     """
 
-    def __init__(self, hs: "HomeServer", auth_handler):
+    def __init__(self, hs: "HomeServer", auth_handler) -> None:
```
squahtx, line 153: `auth_handler: AuthHandler`?

synapse/module_api/__init__.py
```diff
@@ -309,7 +419,7 @@ async def is_user_admin(self, user_id: str) -> bool:
         """
         return await self._store.is_server_admin(UserID.from_string(user_id))
 
-    def get_qualified_user_id(self, username):
+    def get_qualified_user_id(self, username: str) -> str:
```
squahtx, line 422: Do we want to update the docstring to strip out the argument types?

synapse/module_api/__init__.py
```diff
@@ -357,7 +467,7 @@ async def get_threepids_for_user(self, user_id: str) -> List[Dict[str, str]]:
         """
         return await self._store.user_get_threepids(user_id)
 
-    def check_user_exists(self, user_id):
+    def check_user_exists(self, user_id: str):
```
squahtx, line 470: Do we want to update the docstring to strip out the argument types?

squahtx (approved):
lgtm, left a few comments

Issue #11344, opened 2021-11-15
Workers no longer halt startup while the database schema needs an update


anoadragon453:
When redeploying matrix.org today on 2a93005a7b8b099f7f3b98637075dda55be63022, we started with a test worker. The commit included a new schema migration ([this one](https://github.com/matrix-org/synapse/commit/b596a1eb80be4b24142d131238dd5402db6bb855)), which should have preventing the test worker from starting up.

Instead, the test worker failed to notice the new schema file, which is a bit worrying:

<details>

<summary>test worker logs</summary>

```
2021-11-15 11:57:25,378 - root - 345 - WARNING - main - ***** STARTING SERVER *****
2021-11-15 11:57:25,378 - root - 346 - WARNING - main - Server /home/synapse/src/synapse/app/generic_worker.py version 1.47.0rc2 (b=matrix-org-hotfixes,2a93005a7)
2021-11-15 11:57:25,378 - root - 347 - INFO - main - Server hostname: matrix.org
2021-11-15 11:57:25,379 - root - 348 - INFO - main - Instance name: test_worker-1
2021-11-15 11:57:25,379 - synapse.server - 309 - INFO - main - Setting up.
2021-11-15 11:57:25,485 - synapse.storage.databases - 65 - INFO - main - [database config 'master']: Checking database server
2021-11-15 11:57:25,488 - synapse.storage.databases - 71 - INFO - main - [database config 'master']: Preparing for databases ['main', 'state']
2021-11-15 11:57:25,488 - synapse.storage.prepare_database - 116 - INFO - main - ['main', 'state']: Checking existing schema version
2021-11-15 11:57:25,502 - synapse.storage.prepare_database - 124 - INFO - main - ['main', 'state']: Existing schema is 65 (+4 deltas)
2021-11-15 11:57:25,503 - synapse.storage.databases - 84 - INFO - main - [database config 'master']: Starting 'main' database
[...]
2021-11-15 11:57:25,921 - synapse.storage.databases.main.event_push_actions - 540 - INFO - main - Searching for stream ordering 1 month ago
2021-11-15 11:57:25,940 - synapse.storage.databases.main.event_push_actions - 545 - INFO - main - Found stream ordering 1 month ago: it's 2393068710
2021-11-15 11:57:25,940 - synapse.storage.databases.main.event_push_actions - 547 - INFO - main - Searching for stream ordering 1 day ago
[...]

PR #11346, opened 2021-11-15
Run _upgrade_existing_database on workers if at current schema_version

anoadragon453:
Fixes https://github.com/matrix-org/synapse/issues/11344.

Workers need to run `_upgrade_existing_database` to figure out if any outstanding delta files exist (it's the thing that actually traverses the filesystem for new delta files to apply). But, workers only need to run this if we're currently on the latest schema version.

If the database reports that we're not on the latest schema version, then we can simply bail out without needing to run `_upgrade_existing_database`.

#11255 made the correct fix for when we roll back the database schema version, in that `_upgrade_existing_database` should not be run if `version_info.schema_version` > SCHEMA_VERSION. But we still need to run this method on workers if `version_info.schema_version` == SCHEMA_VERSION to get them to halt startup, and let the sysadmin know that they need to run the master first.

Merging to the `release-v1.47` branch as this is intended to be included in v1.47.0rc3.

### DIFF
--- a/synapse/storage/prepare_database.py
+++ b/synapse/storage/prepare_database.py
@@ -131,17 +131,13 @@ def prepare_database(
                     "config==None in prepare_database, but database is not empty"
                 )
 
-            # if it's a worker app, refuse to upgrade the database, to avoid multiple
-            # workers doing it at once.
-            if config.worker.worker_app is None:
-                _upgrade_existing_database(
-                    cur,
-                    version_info,
-                    database_engine,
-                    config,
-                    databases=databases,
-                )
-            elif version_info.current_version < SCHEMA_VERSION:
+            # If the schema version needs to be updated, and we are on a worker, we immediately
+            # know to bail out as workers cannot update the database schema. Only one worker
+            # must update the database at the time, therefore we delegate this task to the master.
+            if (
+                config.worker.worker_app is not None
+                and version_info.current_version < SCHEMA_VERSION
+            ):
                 # If the DB is on an older version than we expect then we refuse
                 # to start the worker (as the main process needs to run first to
                 # update the schema).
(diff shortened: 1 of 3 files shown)


erikjohnston:
Could you add a test case for this please? I guess in tests/storage/test_rollback_worker.py

synapse/storage/prepare_database.py
```diff
@@ -150,6 +146,24 @@ def prepare_database(
                     % (SCHEMA_VERSION, version_info.current_version)
                 )
 
+            # Otherwise if we are on the current schema version, we need to check if there
+            # are any unapplied delta files to process. This should be run on all processes,
+            # master or worker. The master will apply the deltas, workers will check if any
+            # outstanding deltas exist and bail out if they do.
+            if version_info.current_version == SCHEMA_VERSION:
```
richvdh, line 153: we need to call `_upgrade_existing_database` to do the upgrade on the master.

I can't help but wonder if this wouldn't be simpler if we moved the logic into `_upgrade_existing_database`, since that's already got conditional code for workers.

anoadragon453: Good point on both. I've moved the logic (simplifying it quite a bit) into `_upgrade_existing_database` in 
5aa7589, and made sure the master process actually runs its migrations :)

synapse/storage/prepare_database.py
```diff
@@ -358,6 +350,18 @@ def _upgrade_existing_database(
 
     is_worker = config and config.worker.worker_app is not None
 
+    # If the schema version needs to be updated, and we are on a worker, we immediately
+    # know to bail out as workers cannot update the database schema. Only one worker
```
richvdh, line 354:
```suggestion
    # know to bail out as workers cannot update the database schema. Only one process
```

richvdh (approved):
lgtm

PR #11347, opened 2021-11-15
Add ability to un-shadow-ban via the admin API.

### LINKED ISSUE #11343: Admin API feature to un-shadowban
**Description:**

https://matrix-org.github.io/synapse/latest/admin_api/user_admin_api.html#shadow-banning-users appears to only be able to apply the shadow ban, not remove it. And it doesn't appear to be a toggle?

So, the Admin API should get a way to un-shadowban a user.

clokep:
Pretty straightforward, adds `DELETE` method to the shadow ban admin API.

Fixes #11343


changelog.d/11347.misc
```diff
@@ -0,0 +1 @@
+Add admin API to un-shadow-ban a user.
```
babolivier, line 1: I think this should be a `feature`

babolivier (approved):
lgtm otherwise

Issue #8246, opened 2020-09-03
Add better method to repopulate room stats than poking the DB (or remove docs).


erikjohnston:
As per #8238 we have docs that say to repopulate the room stats table you can insert `populate_stats_process_rooms` into the background job table. This is a bit icky as a) its not obvious that background jobs are a "public API" and b) if we add further background jobs that depend on `populate_stats_process_rooms` those will not get re-run (which might lead to bugs).

This feels like something that should be an admin API, or some other form of explicit API, to avoid the above issues.

(Note: we *do* have tests that manually inserting `populate_stats_process_rooms` does the Right Thing, but since its not obvious that its a "public API" the tests may just be modified, which is what allowed the regression in #8238)

Issue #8128, opened 2020-08-19
ability to run the dummy-events-for-extremities job manually


richvdh:
sometimes it would be useful to be able to manually (via admin api or manhole) run the dummy-events-for-extremities job for a given set of rooms, or for a lower than normal extremity count threshold.

dklimpel:
Do you mean this job?
This function has no parameter for rooms at the moment.

https://github.com/matrix-org/synapse/blob/4bd54b263ef7e2ac29acdc85e0c6392684c44281/synapse/handlers/message.py#L1637-L1640

https://github.com/matrix-org/synapse/blob/4bd54b263ef7e2ac29acdc85e0c6392684c44281/synapse/handlers/message.py#L471-L475

PR #11352, opened 2021-11-16
Add admin API to run background jobs

dklimpel:
Fixes: #8246
Related to: #8128

Add API to run predefined background jobs to prevent admins to poke in DB.
`POST /_synapse/admin/v1/background_updates/start_job`

### Pull Request Checklist

### DIFF
--- a/docs/usage/administration/admin_api/background_updates.md
+++ b/docs/usage/administration/admin_api/background_updates.md
@@ -102,14 +102,6 @@ with the following body:
 }
 ```
 
-The API returns the `job_name` param.
-
-```json
-{
-    "job_name": "populate_stats_process_rooms"
-}
-```
-
 The following JSON body parameters are available:
 
 - `job_name` - A string which job to run. Valid values are:
--- a/docs/usage/administration/admin_api/background_updates.md
+++ b/docs/usage/administration/admin_api/background_updates.md
@@ -85,7 +85,7 @@ There is also a `GET` version which returns the `enabled` state.
 
 ## Run
 
-This API runs specific background updates. The job starts immediately after calling the API.
+This API schedules a specific background update to run. The job starts immediately after calling the API.
 
 
 The API is:
--- a/docs/usage/administration/admin_api/background_updates.md
+++ b/docs/usage/administration/admin_api/background_updates.md
@@ -42,7 +42,6 @@ For each update:
 `average_items_per_ms` how many items are processed per millisecond based on an exponential average.
 
 
-
 ## Enabled
 
 This API allow pausing background updates.
@@ -82,3 +81,37 @@ The API returns the `enabled` param.
 ```
 
 There is also a `GET` version which returns the `enabled` state.
+
+
+## Run
+
+This API runs specific background updates. The job starts immediately after calling the API.
+
+
+The API is:
+
+```
+POST /_synapse/admin/v1/background_updates/start_job
+```
+
+with the following body:
+
+```json
+{
+    "job_name": "populate_stats_process_rooms"
+}
+```
+
+The API returns the `job_name` param.
+
+```json
+{
+    "job_name": "populate_stats_process_rooms"
+}
+```
+
+The following JSON body parameters are available:
+
+- `job_name` - A string which job to run. Valid values are:
+  - `populate_stats_process_rooms` - Recalculate the stats for all rooms.
+  - `regenerate_directory` - Recalculate the [user directory](../../../user_directory.md) if it is stale or out of sync.
--- a/docs/user_directory.md
+++ b/docs/user_directory.md
@@ -7,7 +7,8 @@ who are present in a publicly viewable room present on the server.
 
 The directory info is stored in various tables, which can (typically after
 DB corruption) get stale or out of sync.  If this happens, for now the
-solution to fix it is to execute the SQL [here](https://github.com/matrix-org/synapse/blob/master/synapse/storage/schema/main/delta/53/user_dir_populate.sql)
+solution to fix it is to use the [admin API](usage/administration/admin_api/background_updates.md#run) or
+execute the SQL [here](https://github.com/matrix-org/synapse/blob/master/synapse/storage/schema/main/delta/53/user_dir_populate.sql)
 and then restart synapse. This should then start a background task to
 flush the current tables and regenerate the directory.
 
--- a/docs/user_directory.md
+++ b/docs/user_directory.md
@@ -6,10 +6,9 @@ on this particular server - i.e. ones which your account shares a room with, or
 who are present in a publicly viewable room present on the server.
 
 The directory info is stored in various tables, which can (typically after
-DB corruption) get stale or out of sync.  If this happens, for now the
-solution to fix it is to use the [admin API](usage/administration/admin_api/background_updates.md#run) or
-execute the SQL [here](https://github.com/matrix-org/synapse/blob/master/synapse/storage/schema/main/delta/53/user_dir_populate.sql)
-and then restart synapse. This should then start a background task to
+DB corruption) get stale or out of sync. If this happens, for now the
+solution to fix it is to use the [admin API](usage/administration/admin_api/background_updates.md#run)
+and execute the job `regenerate_directory`. This should then start a background task to
 flush the current tables and regenerate the directory.
 
 Data model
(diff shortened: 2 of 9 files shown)


docs/usage/administration/admin_api/background_updates.md
```diff
@@ -82,3 +81,37 @@ The API returns the `enabled` param.
 ```
 
 There is also a `GET` version which returns the `enabled` state.
+
+
+## Run
+
+This API runs specific background updates. The job starts immediately after calling the API.
```
clokep, line 88:
```suggestion
This API schedules a specific background update to run. The job starts immediately after calling the API.
```

docs/usage/administration/admin_api/background_updates.md
```diff
@@ -82,3 +81,37 @@ The API returns the `enabled` param.
...
+
+with the following body:
+
+```json
+{
+    "job_name": "populate_stats_process_rooms"
+}
+```
+
+The API returns the `job_name` param.
```
clokep, line 105: Why return the `job_name`?

docs/user_directory.md
```diff
@@ -7,7 +7,8 @@ who are present in a publicly viewable room present on the server.
 
 The directory info is stored in various tables, which can (typically after
 DB corruption) get stale or out of sync.  If this happens, for now the
-solution to fix it is to execute the SQL [here](https://github.com/matrix-org/synapse/blob/master/synapse/storage/schema/main/delta/53/user_dir_populate.sql)
+solution to fix it is to use the [admin API](usage/administration/admin_api/background_updates.md#run) or
+execute the SQL [here](https://github.com/matrix-org/synapse/blob/master/synapse/storage/schema/main/delta/53/user_dir_populate.sql)
```
clokep, line 11: Probably best if we only suggest a single way to do this, I think?

synapse/rest/admin/background_updates.py
```diff
@@ -29,37 +34,38 @@
...
     """Allows temporarily disabling background updates"""
 
-    PATTERNS = admin_patterns("/background_updates/enabled")
+    PATTERNS = admin_patterns("/background_updates/enabled$")
 
     def __init__(self, hs: "HomeServer"):
-        self.group_server = hs.get_groups_server_handler()
-        self.is_mine_id = hs.is_mine_id
-        self.auth = hs.get_auth()
+        self._is_mine_id = hs.is_mine_id
```
clokep, line 40: None of these updates seem to `_is_mind_id`, I think we can remove it everywhere (by the way, thank you for cleaning up some of this code while you were in here!)

synapse/rest/admin/background_updates.py
```diff
@@ -105,3 +110,68 @@ async def on_GET(self, request: SynapseRequest) -> Tuple[int, JsonDict]:
...
+        self._store = hs.get_datastore()
+
+    async def on_POST(self, request: SynapseRequest) -> Tuple[int, JsonDict]:
+        requester = await self._auth.get_user_by_req(request)
+        await assert_user_is_admin(self._auth, requester.user)
+
+        body = parse_json_object_from_request(request)
+        assert_params_in_dict(body, ["job_name"])
+
+        job_name = body.get("job_name")
```
clokep, line 133: We know the `job_name` key must exist, so best to just get it.

```suggestion
        job_name = body["job_name"]
```

synapse/rest/admin/background_updates.py
```diff
@@ -105,3 +110,68 @@ async def on_GET(self, request: SynapseRequest) -> Tuple[int, JsonDict]:
...
+
+        if job_name == "populate_stats_process_rooms":
+            await self._store.db_pool.simple_insert(
```
clokep, line 136: I think these inserts will fail if the job is currently inserted, do we care about that or should we just let it error?

synapse/storage/background_updates.py
```diff
@@ -122,6 +122,8 @@ def get_current_update(self) -> Optional[BackgroundUpdatePerformance]:
 
     def start_doing_background_updates(self) -> None:
         if self.enabled:
+            # if we start a new background update, not all updates ar done
```
clokep, line 125:
```suggestion
            # if we start a new background update, not all updates are done.
```

Does this fix a bug or just an oddity?

clokep (commented):
Looks good overall!

docs/usage/administration/admin_api/background_updates.md
dklimpel: There is no special reason. What is better? Any ideas?Return nothing? Only status `200`?

clokep: I'm not sure, I think just a `200` would be fine? I see that the enable API returns something similar, but I don't see the point in returning an input parameter?

synapse/storage/background_updates.py
dklimpel: IMO this is a bug.
I am wrote the unit tests first. I did not get the right status with `has_completed_background_updates`. After finish all jobs after synapse start, `_all_done` is always `True`. There is no code to set it `False` if there is a new job.

synapse/rest/admin/background_updates.py
dklimpel: Ok. I did not realized this. What is a good solution? Error `500` or status `200` with a message?

synapse/storage/background_updates.py
clokep: My question was poorly worded -- I agree it is a bug! Do you think it has any impact in normal operations or properly not?

dklimpel: IMO it has no other impact. `_all_done` is only analysed in `has_completed_background_updates` and `has_completed_background_update` to skip db query. It should always return the right status. Most calls are in unit tests.

synapse/rest/admin/background_updates.py
dklimpel: Done.

docs/usage/administration/admin_api/background_updates.md
dklimpel: Done

synapse/storage/background_updates.py
clokep: Thanks for the analysis, that was my gut feeling too!

synapse/rest/admin/background_updates.py
```diff
@@ -104,4 +105,78 @@ async def on_GET(self, request: SynapseRequest) -> Tuple[int, JsonDict]:
...
+    async def on_POST(self, request: SynapseRequest) -> Tuple[int, JsonDict]:
+        requester = await self._auth.get_user_by_req(request)
+        await assert_user_is_admin(self._auth, requester.user)
+
+        body = parse_json_object_from_request(request)
+        assert_params_in_dict(body, ["job_name"])
+
+        job_name = body["job_name"]
+
+        if job_name == "populate_stats_process_rooms":
```
clokep, line 130: Looks good! Can we rework this block a bit to reduce duplication. I think we can do:

```python
if job_name == "...":
    jobs = [...]
elif job_name == "...":
    jobs = [...]
else:
    raise SynError(...)

try:
    simple_insert_many(jobs...)
except IntegrityError:
    raise SynError(...)
```

We should be able to inline the error message too then.

clokep: (And we'll end up using `simple_insert_many` even for the `populate_stats_process_rooms` but that's fine!)

synapse/rest/admin/background_updates.py
```diff
@@ -104,4 +105,72 @@ async def on_GET(self, request: SynapseRequest) -> Tuple[int, JsonDict]:
...
+        try:
+            await self._store.db_pool.simple_insert_many(
+                table="background_updates",
+                values=jobs,
+                desc=f"admin_api_run_{job_name}",
+            )
+        except self._store.db_pool.engine.module.IntegrityError:
+            raise SynapseError(
+                HTTPStatus.BAD_REQUEST,
+                "Job %s is already in queue of background updates." % job_name,
```
clokep, line 171:
```suggestion
                "Job %s is already in queue of background updates." % (job_name,),
```

clokep (approved):
Looks good (assuming CI passes!)

clokep:
Thanks so much! 🎉

PR #11353, opened 2021-11-16
Rename `remove_deleted_devices_from_device_inbox` to ensure it is always run

anoadragon453:
This PR creates a new background update called `remove_deleted_devices_from_device_inbox_v2` (note that v2), and makes the original background update a no-op.

This is to allow homeservers that have already started running the old background update to finish it immediately, before then running a new background update, which does the same thing.

As this bg update is just deleting old to-device messages, it's fine to stop in the middle, and then restart, as we would be doing here.

Only a subset of homeservers would have actually started running the bg update. For some background, https://github.com/matrix-org/synapse/pull/11303 moved the migration file that triggered this background update from schema version 64 to 65. While it was in 64, we released v1.47.0rc1, rc2. Any homeserver that updated from a version of Synapse with schema version 64 or lower (<=v1.45.0) would have run the old background update, and potentially have it still running currently. This PR fixes that edge case.


anoadragon453:
The test failures are flakes.

reivilibre:
I'm confused why this is necessary; is it because the delta got moved and so a HS might run the update twice?

If so, another way we've handled this in the past is to issue a `DELETE FROM` in the delta to clear out duplicate jobs before creating the new job.

anoadragon453:
> I'm confused why this is necessary; is it because the delta got moved and so a HS might run the update twice?

Yeah, simply that.

> If so, another way we've handled this in the past is to issue a `DELETE FROM` in the delta to clear out duplicate jobs before creating the new job.

great suggestion! Looks like we have [indeed learned of better ways](https://github.com/matrix-org/synapse/blob/25f43faa70f7cc58493b636c2702ae63395779dc/synapse/storage/schema/main/delta/58/16populate_stats_process_rooms_fix.sql#L20) to do this since.

I'll modify the PR in that direction.

changelog.d/11353.misc
```diff
@@ -0,0 +1 @@
+Fix an issue which prevented the 'remove deleted devices from device_inbox column' background process from running when updating from a recent Synapse version.
```
reivilibre, line 1:
```suggestion
Fix an issue which prevented the 'remove deleted devices from `device_inbox` column' background process from running when updating from a recent Synapse version.
```

reivilibre (approved):
Yup, this is good. Thanks!

PR #11354, opened 2021-11-16
Get the cache storage passing mypy

DMRobertson:
This was a PITA wrangling the inheritance sphagetti.

Review commit-by-commit.

### DIFF
--- a/synapse/replication/slave/storage/events.py
+++ b/synapse/replication/slave/storage/events.py
@@ -48,10 +48,10 @@ logger = logging.getLogger(__name__)
 
 class SlavedEventStore(
     RoomMemberWorkerStore,
-    StreamWorkerStore,
     StateGroupWorkerStore,
     UserErasureWorkerStore,
     BaseSlavedStore,
+    StreamWorkerStore,
 ):
     def __init__(self, database: DatabasePool, db_conn, hs: "HomeServer"):
         super().__init__(database, db_conn, hs)
--- a/synapse/replication/slave/storage/events.py
+++ b/synapse/replication/slave/storage/events.py
@@ -47,7 +47,6 @@ logger = logging.getLogger(__name__)
 
 
 class SlavedEventStore(
-    RoomMemberWorkerStore,
     StateGroupWorkerStore,
     UserErasureWorkerStore,
     BaseSlavedStore,
--- a/synapse/replication/slave/storage/events.py
+++ b/synapse/replication/slave/storage/events.py
@@ -48,7 +48,6 @@ logger = logging.getLogger(__name__)
 
 class SlavedEventStore(
     RoomMemberWorkerStore,
-    EventPushActionsWorkerStore,
     StreamWorkerStore,
     StateGroupWorkerStore,
     UserErasureWorkerStore,
--- a/synapse/replication/slave/storage/events.py
+++ b/synapse/replication/slave/storage/events.py
@@ -47,13 +47,10 @@ logger = logging.getLogger(__name__)
 
 
 class SlavedEventStore(
-    EventFederationWorkerStore,
     RoomMemberWorkerStore,
     EventPushActionsWorkerStore,
     StreamWorkerStore,
     StateGroupWorkerStore,
-    EventsWorkerStore,
-    SignatureWorkerStore,
     UserErasureWorkerStore,
     RelationsWorkerStore,
     BaseSlavedStore,
--- a/synapse/replication/slave/storage/events.py
+++ b/synapse/replication/slave/storage/events.py
@@ -52,7 +52,6 @@ class SlavedEventStore(
     StreamWorkerStore,
     StateGroupWorkerStore,
     UserErasureWorkerStore,
-    RelationsWorkerStore,
     BaseSlavedStore,
 ):
     def __init__(self, database: DatabasePool, db_conn, hs: "HomeServer"):
--- a/synapse/replication/slave/storage/events.py
+++ b/synapse/replication/slave/storage/events.py
@@ -16,14 +16,6 @@ import logging
 from typing import TYPE_CHECKING
 
 from synapse.storage.database import DatabasePool
-from synapse.storage.databases.main.event_federation import EventFederationWorkerStore
-from synapse.storage.databases.main.event_push_actions import (
-    EventPushActionsWorkerStore,
-)
-from synapse.storage.databases.main.events_worker import EventsWorkerStore
-from synapse.storage.databases.main.relations import RelationsWorkerStore
-from synapse.storage.databases.main.roommember import RoomMemberWorkerStore
-from synapse.storage.databases.main.signatures import SignatureWorkerStore
 from synapse.storage.databases.main.state import StateGroupWorkerStore
 from synapse.storage.databases.main.stream import StreamWorkerStore
 from synapse.storage.databases.main.user_erasure_store import UserErasureWorkerStore
--- a/synapse/replication/slave/storage/events.py
+++ b/synapse/replication/slave/storage/events.py
@@ -17,7 +17,6 @@ from typing import TYPE_CHECKING
 
 from synapse.storage.database import DatabasePool
 from synapse.storage.databases.main.state import StateGroupWorkerStore
-from synapse.storage.databases.main.stream import StreamWorkerStore
 from synapse.storage.databases.main.user_erasure_store import UserErasureWorkerStore
 from synapse.util.caches.stream_change_cache import StreamChangeCache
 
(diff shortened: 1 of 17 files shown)


DMRobertson:
Need to fix portdb.

DMRobertson:
That sytest failure on buster multi-postgres is a suspected flake, see https://github.com/matrix-org/sytest/issues/1167

DMRobertson:
Think this is reviewable now. Suggest going commit-by-commit up to 
5343d6f , then the last three are probably best viewed together.

DMRobertson:
I've rebased to try and trim down the commits. There are still quite a few small commits trying to make one surgical change to the storage class inheritance, with comments to try and justify them. I still think it's worth going through commit-by-commit.

I'm still quite scared by all this. I'm not sure if we should go ahead with all of this change (input welcome). I think it's useful to see what dependencies we have, even if they might not all make sense. Maybe this can inform us on how to do things better (https://github.com/matrix-org/synapse/issues/11165) if nothing else.

Debugging this to get a consistent MRO between all the datastore classes was the motivation for #11357.

synapse/storage/databases/main/cache.py
```diff
@@ -85,7 +99,13 @@ def get_all_updated_caches_txn(txn):
...
             """
             txn.execute(sql, (last_id, instance_name, limit))
-            updates = [(row[0], row[1:]) for row in txn]
+            updates: List[Tuple[int, _CacheData]] = []
+            row: Tuple[int, str, Optional[List[str]], Optional[int]]
+            # Type saftey: iterating over `txn` yields `Tuple`, i.e.
+            # `Tuple[Any, ...]` of arbitrary length. Mypy detects assigning a
+            # variadic tuple to a fixed length tuple and flags it up as an error.
+            for row in txn:  # type: ignore[assignment]
+                updates.append((row[0], row[1:]))
```
clokep, line 128: I think this could probably have stayed a list comprehension?

synapse/replication/slave/storage/events.py
```diff
@@ -48,10 +48,10 @@
 
 class SlavedEventStore(
     RoomMemberWorkerStore,
-    StreamWorkerStore,
     StateGroupWorkerStore,
     UserErasureWorkerStore,
     BaseSlavedStore,
+    StreamWorkerStore,
```
clokep, line 45: Is `StreamWorkerStore` inherited via `BaseSlavedStore` anyway? (Why did we move it this time instead of remove it?)

synapse/storage/databases/main/cache.py
```diff
@@ -59,6 +60,7 @@ class CacheInvalidationWorkerStore(
     RelationsWorkerStore,
     EventPushActionsWorkerStore,
     StreamWorkerStore,
+    StatsStore,
```
clokep, line 64: Instead of this we could use `_clock`, which is provided by `SQLBaseStore`. That might be a separate PR though?

DMRobertson: I like the sound of that. I really didn't like making `CacheInvalidationWorkerStore` depend on stats!

DMRobertson: I still need something to provide `self._curr_state_delta_stream_cache` though, so I'll have to inherit from `StateDeltasStore` for now.

clokep: I agree! While we're at it, maybe make `stats` just use the one from `SQLBaseStore` too?

DMRobertson: In the future I think we could refactor the (only) usage of _curr_state_delta_stream_cache to call a method on StateDeltasStore.

synapse/storage/databases/main/cache.py
DMRobertson: Unfortunately I think the annotation to `row` on 123 doesn't apply to an iteration variable inside a comprehension or generator expression.

Replacing 122--128 with the following

```python
            updates: List[Tuple[int, _CacheData]] = []
            row: Tuple[int, str, Optional[List[str]], Optional[int]]
            updates = [(row[0], row[1:]) for row in txn]
```

yields:

```
mypy
synapse/storage/databases/main/cache.py:127: error: List comprehension has incompatible type List[Tuple[Any, Tuple[Any, ...]]]; expected List[Tuple[int, Tuple[str, Optional[List[str]], Optional[int]]]]  [misc]
```

clokep: Ah, that's unfortunate. I think it gets scoped to just the comprehension.

synapse/storage/databases/main/cache.py
clokep: Sure, I meant specifically the `clock` just to be clear!

DMRobertson: Ahh, I think our comments raced each other! I was thinking of a way to make `CacheInvalidationWorkerStore` less icky.

But agreed, I'm only planning to adjust `StatsStore.clock`.

synapse/storage/databases/main/cache.py
DMRobertson: Yeah. I think we'd need to `cast()` inside the comprehension. I felt like the above was the least-bad option.

synapse/replication/slave/storage/events.py
DMRobertson: Let me double check this. I'm all `Store`d out!

DMRobertson: You're spot on! If you'll forgive me for showing off a graph:

image

As for move versus delete, I suspect I moved it in one commit to solve an MRO conflict, and then didn't notice I could remove it later on.

clokep: Maybe! Do you think it is worth changing then?

DMRobertson:
@clokep  thanks for your comments! Could you take another look once CI has run?

synapse/replication/slave/storage/events.py
DMRobertson: Yep, it's removed in 4686190.

clokep:
@DMRobertson Looks like portdb is failing.

DMRobertson:
And `trial` too:

```
   File "/home/runner/work/synapse/synapse/synapse/visibility.py", line 149, in allowed
    oldest_allowed_ts = storage.main.clock.time_msec() - max_lifetime
builtins.AttributeError: 'DataStore' object has no attribute 'clock'
```

DMRobertson:
> And `trial` too:
> 
> ```
>    File "/home/runner/work/synapse/synapse/synapse/visibility.py", line 149, in allowed
>     oldest_allowed_ts = storage.main.clock.time_msec() - max_lifetime
> builtins.AttributeError: 'DataStore' object has no attribute 'clock'
> ```

Why didn't mypy spot this?

```python
                    reveal_type(storage.main)
```

```
synapse/visibility.py:149: note: Revealed type is "Any"
```
Gonads.

I think this has now become a Tomorrow Problem.

clokep:
I believe this comes back to `stores.main` being `DataStoreT`?

DMRobertson:
> I believe this comes back to `stores.main` being `DataStoreT`?

Agreed, but I don't think mypy can see that because `Storage` isn't generic over `DataStoreT` itself. (?)

DMRobertson:
@clokep I had a quick go at getting mypy to spot the problem that made the tests fail. I think it'll be a bit involved though; I'd like to handle that in a separate change and get this wrapped up.

To proceed, I can see three options:
1. revert the last three changes, so that CacheInvalidationWorkerStore inherits from StatsStore again. (i.e. reset back to dfe4779)
2. keep the last three changes, and change the `.clock` attribute reference that's causing tests to fail to `._clock`.
3. keep the last three changes. Introduce a `clock` read-only `@property` on SQLBaseStore that exposes the `_clock`.

I think my preferences would be 3 > 1 > 2; what do you think? Perhaps there's another way to proceed that I'm not seeing.

clokep:
@DMRobertson I think my preference would be a twist on idea 2 -- the twist being splitting it out to a separate PR, which I hope will help isolate the change.

clokep:
Removing my review since I've given thoughts already -- @DMRobertson if you're still waiting for info for me please clarify what I can do to help move this along!

DMRobertson:
> @DMRobertson I think my preference would be a twist on idea 2 -- the twist being splitting it out to a separate PR, which I hope will help isolate the change.

Thanks. I had a quick go at this (I wanted to make `Storage` generic over `DataStoreT` but found it to be quick tricky. I suggest we go with plan 2 and that I leave a comment summarising the situation.

DMRobertson:
More failures, some in portdb and some in sytest.

Feeling a bit swamped by this. I'm wondering if it might be better to close this and try to pull out some easier changes.

DMRobertson:
This has rotted. The last time we talked about the DB monstrosity I think we were in favour of a more drastic, composition-over-inheritance approach; see #11733.

PR #11355, opened 2021-11-16
Avoid sharing room hierarchy responses between users

squahtx:
Different users may be allowed to see different rooms within a space,
so sharing responses between users is inadvisable.

### DIFF
--- a/tests/handlers/test_room_summary.py
+++ b/tests/handlers/test_room_summary.py
@@ -14,6 +14,8 @@
 from typing import Any, Iterable, List, Optional, Tuple
 from unittest import mock
 
+from twisted.internet.defer import ensureDeferred
+
 from synapse.api.constants import (
     EventContentFields,
     EventTypes,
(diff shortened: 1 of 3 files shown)


clokep (changes requested):
Good catch, can you add a test for this?

squahtx:
I've added some tests now

tests/handlers/test_room_summary.py
```diff
@@ -316,6 +318,59 @@ def test_visibility(self):
...
+        result1 = self.get_success(deferred1)
+        result2 = self.get_success(deferred2)
+
+        # Both `get_room_hierarchy` calls should return the same result.
+        expected = [(self.space, [self.room]), (self.room, ())]
+        self._assert_hierarchy(result1, expected)
+        self._assert_hierarchy(result2, expected)
+        self.assertIs(result1, result2)
+
+        # A subsequent `get_room_hierarchy` call should reuse the result.
```
clokep, line 341: How are we testing that it re-uses it? We assert that `result1` != `result3` so I'm not sure what we're checking here?

tests/handlers/test_room_summary.py
```diff
@@ -316,6 +318,59 @@ def test_visibility(self):
             AuthError,
         )
 
+    def test_room_hierarchy_cache(self) -> None:
+        """In-flight room hierarchy requests are deduplicated."""
+        # Run two `get_room_hierarchy` calls up until they block.
```
clokep, line 323: "up until they block" is why we manually call `ensureDeferred` instead of letting `get_success` do that?

tests/handlers/test_room_summary.py
squahtx: Oops, that comment is missing a `not`.

tests/handlers/test_room_summary.py
squahtx: Yep, otherwise they'd run sequentially and the cache wouldn't do anything.

clokep: That's what I thought, just double checking!

clokep (approved):
Thanks so much for finding and fixing this!

PR #11356, opened 2021-11-16
Annotate string constants in `synapse.api.constants` with `Final`

squahtx:
This change makes mypy complain if the constants are ever reassigned,
and, more usefully, makes mypy type them as `Literal`s instead of `str`s,
allowing code of the following form to pass mypy:
```py
def do_something(membership: Literal["join", "leave"], ...): ...

do_something(Membership.JOIN, ...)
```
-----------------------------------------------------
I tried @DMRobertson 's suggestion of `StrEnums`, but we `sys.intern` bits of event JSON in various places, which raises when it encounters enums. And having dict keys sometimes be strings and sometimes be enum values defeats the purpose of interning.

### DIFF
--- a/synapse/api/constants.py
+++ b/synapse/api/constants.py
@@ -17,6 +17,8 @@
 
 """Contains constants from the specification."""
 
+from typing_extensions import Final
+
 # the max size of a (canonical-json-encoded) event
 MAX_PDU_SIZE = 65536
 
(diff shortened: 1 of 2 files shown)


DMRobertson:
> I tried @DMRobertson 's suggestion of StrEnums, but we sys.intern bits of event JSON in various places, which raises when it encounters enums. And having dict keys sometimes be strings and sometimes be enum values defeats the purpose of interning.

That's a real shame (from the POV of type checking): we can't write `def foo(membership: Membership). Ah well.

synapse/api/constants.py
```diff
@@ -39,217 +41,217 @@ class Membership:
 
     """Represents the membership states of a user in a room."""
 
-    INVITE = "invite"
-    JOIN = "join"
-    KNOCK = "knock"
-    LEAVE = "leave"
-    BAN = "ban"
-    LIST = (INVITE, JOIN, KNOCK, LEAVE, BAN)
+    INVITE: Final = "invite"
```
reivilibre, line 44: I was wondering if this should say `Final[str]`. Answer seems to be a big fat NO.
If you make it `Final[str]`, then it types as `str` rather than `Literal['invite']`.
So I think this is fine.

synapse/api/constants.py
```diff
@@ -39,217 +41,217 @@ class Membership:
...
+    JOIN: Final = "join"
+    KNOCK: Final = "knock"
+    LEAVE: Final = "leave"
+    BAN: Final = "ban"
+    LIST: Final = (INVITE, JOIN, KNOCK, LEAVE, BAN)
```
reivilibre, line 49: > `Revealed type is "Tuple[Literal['invite']?, Literal['join']?, Literal['knock']?, Literal['leave']?, Literal['ban']?]"`

Actually, the earlier lines print things like `Revealed type is "Literal['invite']?"` too.

Do we know why it says a question mark after each Literal? ... It makes me worry that something secret is going on.

reivilibre (approved):
Slight concern with the question mark, but overall it seems sensible

synapse/api/constants.py
DMRobertson: Good question. https://mypy.readthedocs.io/en/stable/literal_types.html#declaring-literal-variables says

> If you do not provide an explicit type in the Final, the type of c becomes context-sensitive: mypy will basically try “substituting” the original assigned value whenever it’s used before performing type checking. This is why the revealed type of c is Literal[19]?: the question mark at the end reflects this context-sensitive nature.

reivilibre: Thanks for the link David. It took me a few reads to see what it's going on about, but interesting — question is: do we want the behaviour given by `Final` or the one given by `Literal[…]`?

squahtx: > do we want the behaviour given by `Final` or the one given by `Literal[…]`?

I'm having trouble coming up with an example where it makes a difference. Do you have one in mind?

squahtx: It's really hard to come up with an example where it makes a difference. Here's one:
```py
T = TypeVar("T", bound=str)
def validate_or_throw(value: str, allowed_values: List[T]) -> T:
    ...

ACTIONS_IMPLICIT: Final = ["ban", "unban"]
ACTIONS_EXPLICIT: List[Literal["ban", "unban"]] = ["ban", "unban"]

def do_action(action: Literal["ban", "unban"]):
    pass

def on_GET1(json: Dict):
    action = validate_or_throw(json["action"], ACTIONS_IMPLICIT)
    do_action(action)  # Argument 1 to "do_action" has incompatible type "str"

def on_GET2(json: Dict):
    action = validate_or_throw(json["action"], ACTIONS_EXPLICIT)
    do_action(action)  # ok
```

I'm tempted to stick with `Final` until the context-sensitive nature becomes a problem.

DMRobertson: > I'm tempted to stick with Final until the context-sensitive nature becomes a problem.

I'm fine with this---no worse than the current situation as far as I can see

reivilibre: aye yeah, let's just stick with it

PR #11357, opened 2021-11-16
Commit hacky script to visualise store inheritance

DMRobertson:
Use e.g. with `scripts-dev/storage_inheritance.py DataStore --show`.

It's dirty, but it's useful enough to help disentangle the sphagetti.

### DIFF
--- a/scripts-dev/storage_inheritance.py
+++ b/scripts-dev/storage_inheritance.py
@@ -1,0 +1,180 @@
+#! /usr/bin/env python3
+import argparse
+import os
+import re
+import subprocess
+import sys
+import tempfile
+from typing import Iterable, Optional, Set
+
+import networkx
+
+
+def scrape_storage_classes() -> str:
+    """Grep the for classes ending with "Store" and extract their list of parents.
+
+    Returns the stdout from `rg` as a single string."""
+
+    # TODO: this is a big hack which assumes that each Store class has a unique name.
+    #   That assumption is wrong: there are two DirectoryStores, one in
+    #   synapse/replication/slave/storage/directory.py and the other in
+    #   synapse/storage/databases/main/directory.py
+    #   Would be nice to have a way to account for this.
+
+    return subprocess.check_output(
+        [
+            "rg",
+            "-o",
+            "--no-line-number",
+            "--no-filename",
+            "--multiline",
+            r"class .*Store\((.|\n)*?\):$",
+            "synapse",
+            "tests",
+        ],
+        cwd="/home/dmr/workspace/synapse/",
+    ).decode()
+
+
+oneline_class_pattern = re.compile(r"^class (.*)\((.*)\):$")
+opening_class_pattern = re.compile(r"^class (.*)\($")
+
+
+def load_graph(lines: Iterable[str]) -> networkx.DiGraph:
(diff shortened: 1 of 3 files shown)


DMRobertson:
`scripts-dev/storage_inheritance.py DataStore --show` running at 9e361c8:

image

`scripts-dev/storage_inheritance.py BaseSlavedStore --show` at cd7257bf27b0344266ae507085066c9c6c604b6d:

image

`scripts-dev/storage_inheritance.py --show` at 9e361c8:

image

scripts-dev/storage_inheritance.py
```diff
@@ -0,0 +1,180 @@
+#! /usr/bin/env python3
+import argparse
+import os
+import re
+import subprocess
+import sys
+import tempfile
+from typing import Iterable, Optional, Set
+
+import networkx
```
babolivier, line 10: We should add this to the dev dependencies.

scripts-dev/storage_inheritance.py
```diff
@@ -0,0 +1,180 @@
...
+            "rg",
+            "-o",
+            "--no-line-number",
+            "--no-filename",
+            "--multiline",
+            r"class .*Store\((.|\n)*?\):$",
+            "synapse",
+            "tests",
+        ],
+        cwd="/home/dmr/workspace/synapse/",
```
babolivier, line 35: I assume we probably want to change this so it doesn't point to your home directory.

babolivier:
Also I would suggest writing the file on disk rather than opening it. For example I know Gnome has some weird magic around default apps, and when I tried running the script it opened it with Gimp for some reason despite the fact that I've changed the default app for PNGs to something else some time ago.

DMRobertson:
> Also I would suggest writing the file on disk rather than opening it. 

There's a command line option for this: `--output DEST`.

> For example I know Gnome has some weird magic around default apps, and when I tried running the script it opened it with Gimp for some reason despite the fact that I've changed the default app for PNGs to something else some time ago.

Suggest checking your config with:

```
$ xdg-mime query default image/png
org.gnome.eog.desktop
```

babolivier:
> There's a command line option for this: `--output DEST`.

Hah, I see. Maybe worth adding some doc along with the script so we remember how to use it, then? Otherwise I'm afraid it'll become like the config validator/parser util in Synapse and we'll completely forget it's a thing until someone complains we don't have such a tool or it's broken.

DMRobertson:
> Hah, I see. Maybe worth adding some doc along with the script so we remember how to use it, then?

I use argparse so there's a `--help` out of the box for this

anoadragon453:
What's the expected lifetime of this script? Do we expect to remove it once DataStore is less :spaghetti:?

clokep:
I'm going to remove this from the review queue since it seems to be awaiting info from @DMRobertson.

DMRobertson:
> What's the expected lifetime of this script? Do we expect to remove it once DataStore is less spaghetti?

I hadn't really considered it, but that sounds sensible.

(I can imagine a version that imports a class and inspects `__mro__` attributes to build this graph, but hopefully we aren't relying on inheritance to this extent anywhere else?)

PR #11359, opened 2021-11-16
Fix incorrect return value in tests.

clokep:
This seems to be from a conflict between #11282 and #11346.


clokep:
You can see a failing build: https://github.com/matrix-org/synapse/runs/4227021769?check_suite_focus=true

clokep:
Thanks!

Issue #11173, opened 2021-10-25
Extend ModuleApi with APIs to find out how many messages a user has sent and when the account was created


Yoric:
Both APIs would be very useful to write spam checkers.

rhetenor:
Hey,

The account creation date can already be queried via `ModuleApi.get_userinfo_by_id`. 
For the message count I've just created a pull request that should fix this issue.

babolivier:
As @rhetenor points out, `ModuleApi.get_userinfo_by_id` already tells modules about account creation dates. It only does it for local accounts, but that's because a homeserver has no way of knowing the creation date of an account that was created on another homeserver.

About message counts, we just had a discussion about this in #synapse-dev (https://matrix.to/#/!XaqDhxuTIlvldquJaV:matrix.org/$IQwnIgcWDM3gQ83WquHu_4e0VwNC9kHRGvMTr5PGelQ?via=matrix.org&via=vector.modular.im&via=envs.net) and concluded that the best way to do this would be on the module side, by leveraging the new `on_new_event` callback (https://matrix-org.github.io/synapse/develop/modules/third_party_rules_callbacks.html#on_new_event) that will be added by Synapse 1.47.0 (expected later today), and is called every time Synapse persists a new event. Doing it on Synapse looks like it would likely be quite expensive and non-trivial, for a result that wouldn't be complete for remote users anyway (since a homeserver has no way to know about messages it doesn't see). Plus doing it on the module's side would give the module more freedom to be peckish about what events they want to count, and how it wants to store its counts.

PR #11364, opened 2021-11-16
Add count message function to ModuleAPI

rhetenor:
Referring to issue #11173 this pull requests implements a new endpoint of the ModuleAPI to count the total messages sent by a given user.

### Pull Request Checklist

* [X] [Code style](https://matrix-org.github.io/synapse/latest/code_style.html) is correct


synapse/module_api/__init__.py
```diff
@@ -425,6 +425,15 @@ def register_user(
...
         )
 
+    def get_user_total_message_count(self, user_id):
+        """Counts the total number of messages, sent by a user.
+        Args:
+            user_id (str): full canonical @user:id
+        Returns:
+            defer.Deferred[int] : The number of messages.
+        """
+        return defer.ensureDeferred(self._store.count_total_messages(user_id))
```
babolivier, line 435: Could you use Python's async/await semantics here rather than Deferreds?
It's true some other methods use Deferreds, but the reason for that is mostly that we haven't converted them to async/await yet.

synapse/module_api/__init__.py
```diff
@@ -425,6 +425,15 @@ def register_user(
             )
         )
 
+    def get_user_total_message_count(self, user_id):
+        """Counts the total number of messages, sent by a user.
+        Args:
+            user_id (str): full canonical @user:id
+        Returns:
+            defer.Deferred[int] : The number of messages.
```
babolivier, line 433: On the docstring can you:

* add an empty line between each section
* add the version of Synapse the method will be shipped with - in this case it will likely be 1.48.0

Feel free to refer to other docstrings in the file for examples.

synapse/storage/databases/main/metrics.py
```diff
@@ -521,3 +521,30 @@ def _generate_user_daily_visits(txn):
...
+        """
+
+        def _count_messages(txn):
+
+            sql = """
+                SELECT COALESCE(COUNT(*), 0) FROM events
+                WHERE type = 'm.room.message'
+                    OR type = 'm.room.encrypted'
+                    AND sender LIKE ?
+            """
```
babolivier, line 542: I think it would be wiser to make the list of types to retrieve an argument of the function, that way modules would have more flexibility on what event is considered in the count (and we could potentially save up on performance by counting less events).

Also, we probably want `sender = ?` instead of `LIKE`, I can't see any reason why `user_id` would be a SQL regex.

changelog.d/11364.feature
```diff
@@ -0,0 +1 @@
+Add ModuleAPI endpoint to count total messages sent by a user. Contributed by @rhetenor, @Neeeflix
```
babolivier, line 1:
```suggestion
Add a module API method to count total messages sent by a user. Contributed by @rhetenor, @Neeeflix
```

babolivier (changes requested):
Thanks for contributing to Synapse! It looks mostly fine to me, apart from a few places that need a bit of change.

babolivier:
Actually, after discussing this in #synapse-dev, we don't think this is the way we want to solve this issue. The `events` table is a big one, and doesn't have the right index for the required SQL query. We could add this index in a background update, but that could also be a heavy operation that would require a fair bit of thinking to make sure we get it, and `get_user_total_message_count`, right.

An easier and nicer solution is probably to have the module use the new `on_new_event` callback that will be introduced in Synapse 1.47.0 (which is called every time a new event comes in) to count new events itself (and be able to be more picky about what events it cares about). Sure, the module will be missing some history on the user's activity, but that's also the case when querying the `users` table directly (unless the user is local to the homeserver, how can the homeserver know that it has all of their messages?).

I apologise this isn't a very satisfying conclusion to your attempt at a first contribution, and I hope this will not put you off from contributing to Synapse!

Issue #11363, opened 2021-11-16
./scripts-dev/complement.sh and GHA complement job diverged


Bubu:
It seems like https://github.com/matrix-org/synapse/commit/684d19a11c3b93c9dd5fb90f43d38aa7e8c6005f added the `msc2716` tag to the complement.sh script but it seems that was never added to the GHA workflow here: https://github.com/matrix-org/synapse/blob/develop/.github/workflows/tests.yml#L377

Consequently that (those?) tests are now failing on develop.

clokep:
@MadLittleMods Do you know if those tests should be passing or not?

MadLittleMods:
I think they would but it looks like the Complement tests are using the `org.matrix.msc2716v4` which hasn't merged to Synapse yet (potentially merged in https://github.com/matrix-org/synapse/pull/11243 or https://github.com/matrix-org/synapse/pull/9445 or not merged at all https://github.com/matrix-org/synapse/pull/11243#discussion_r750816377). Feel free to remove, it's always a juggle trying to add it back when developing a new MSC though.

clokep:
> I don't think the tests are running in CI (GHA) because the command is separate from `complement.sh` though?

Right, but for people developing against Synapse it is frequent to manually run complement commands with this script, so it is best to keep it in a state where it is expected to pass.

DMRobertson:
Is this fixed by https://github.com/matrix-org/synapse/pull/11368?

clokep:
> Is this fixed by #11368

No, I'll do a PR to fix it.

PR #11368, opened 2021-11-17
Fix running complement.sh script.

### LINKED ISSUE #11362: scripts-dev/complement.sh broke after 13f084e
### Description

It seems like  13f084e broke the complement.sh script included here, it doesn't find any tests to run anymore:

```
Successfully built 41beaabb3d44
Successfully tagged complement-synapse:latest
no Go files in /home/bubu/projects/oss/synapse/complement-master
```

When manually reverting the changes to that file it works again (the actual tests fail somewhere down the line, but that another matter).

### Steps to reproduce

- run `./scripts-dev/complement.sh` from an up-to-date synapse checkout

clokep:
By reverting changes from #11166 in this script. Specifically commit 13f084eb58c379b2fc073680818d5931e94626f4.

Fixes #11362

### DIFF
--- a/changelog.d/11368.misc
+++ b/changelog.d/11368.misc
@@ -1,0 +1,1 @@
+Fix running `scripts-dev/complement.sh`, which was broken in v1.47.0rc1.
--- a/scripts-dev/complement.sh
+++ b/scripts-dev/complement.sh
@@ -24,7 +24,7 @@
 set -e
 
 # Change to the repository root
-cd "$(dirname "$0")/.."
+cd "$(dirname $0)/.."
 
 # Check for a user-specified Complement checkout
 if [[ -z "$COMPLEMENT_DIR" ]]; then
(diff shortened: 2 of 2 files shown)


clokep:
I opted to fully revert the changes to this script since there's a variety of ways to run this script and that made me more confident.

I debated about removing `msc2716` to also handle #11363 here, but opted not to. Will wait to see if @MadLittleMods has thoughts first.

DMRobertson (commented):
Guessing this was #11166?

clokep:
> Guessing this was #11166?

Yes, see the description.

DMRobertson (approved):
I'd like to understand what went wrong here, but let's get it back to a known-good state for now.

clokep:
> I'd like to understand what went wrong here, but let's get it back to a known-good state for now.

That was my thought process, yes!

PR #11369, opened 2021-11-17
Rename `get_access_token_for_user_id` method to `create_access_token_for_user_id`

reivilibre:
This one was a little bit confusing and I'm not a fan of conflating 'getting' with 'creating', especially around authentication code.


babolivier (approved):
Hah, I got confused by exactly this just yesterday, so thanks!

PR #11370, opened 2021-11-17
Rename `get_refresh_token_for_user_id` to `create_refresh_token_for_user_id`

reivilibre:
Follow up to #11369 because of course I missed one


babolivier (approved):
A bit worrying that no test picked that up.

reivilibre:
> A bit worrying that no test picked that up.

Sorry to mislead — it's a different function with the same naming scheme. :D

PR #8450, opened 2020-10-02
Speed up unit tests when using PostgreSQL

erikjohnston:

PR #11376, opened 2021-11-17
Track ongoing event fetches correctly (again)

### LINKED ISSUE #11167: Synapse does not recover correctly after a database server outage
Earlier today our database server restarted. The database recovered itself; Synapse did not. In particular we saw "in flight requests" stacking up:

image

... and the reverse-proxy returned 429s for many requests.

squahtx:
The previous fix for the ongoing event fetches counter
(8eec25a1d9d656905db18a2c62a5552e63db2667) was both insufficient and
incorrect.

When the database is unreachable, `_do_fetch` never gets run and so
`_event_fetch_ongoing` is never decremented.

The previous fix also moved the `_event_fetch_ongoing` decrement outside
of the `_event_fetch_lock` which allowed race conditions to corrupt the
counter.

Fixes #11167

### DIFF
--- a/synapse/storage/databases/main/events_worker.py
+++ b/synapse/storage/databases/main/events_worker.py
@@ -804,9 +804,10 @@ class EventsWorkerStore(SQLBaseStore):
                     # We exited cleanly but noticed more work.
                     self._maybe_start_fetch_thread()
 
-                if exc is not None and event_fetches_to_fail:
+                if event_fetches_to_fail:
                     # We were the last remaining fetcher and failed.
                     # Fail any outstanding fetches since no one else will handle them.
+                    assert exc is not None
                     with PreserveLoggingContext():
                         for _, deferred in event_fetches_to_fail:
                             deferred.errback(exc)
--- a/synapse/storage/databases/main/events_worker.py
+++ b/synapse/storage/databases/main/events_worker.py
@@ -741,23 +741,40 @@ class EventsWorkerStore(SQLBaseStore):
         try:
             await self.db_pool.runWithConnection(self._do_fetch_txn)
         except BaseException as e:
+            failed_event_list = []
             with self._event_fetch_lock:
-                self._event_fetch_ongoing -= 1
-                event_fetch_ongoing_gauge.set(self._event_fetch_ongoing)
-
-                if self._event_fetch_ongoing == 0 and self._event_fetch_list:
+                if self._event_fetch_ongoing == 1 and self._event_fetch_list:
                     # We are the last remaining fetcher and we have just failed.
-                    # Fail any outstanding event fetches, since no one else will process
-                    # them.
+                    # Fail any outstanding fetches, since they won't get processed
+                    # otherwise.
                     failed_event_list = self._event_fetch_list
                     self._event_fetch_list = []
-                else:
-                    failed_event_list = []
+
+                self._event_fetch_ongoing -= 1
+                event_fetch_ongoing_gauge.set(self._event_fetch_ongoing)
 
             for _, deferred in failed_event_list:
                 if not deferred.called:
                     with PreserveLoggingContext():
                         deferred.errback(e)
+        else:
+            should_restart = False
+            with self._event_fetch_lock:
+                if self._event_fetch_ongoing == 1 and self._event_fetch_list:
+                    # An event fetch has been queued, but we're the last remaining
+                    # fetcher and have already decided to terminate.
+                    # Start a new fetcher.
+                    should_restart = True
+
+                    # `_event_fetch_ongoing` ought to be decremented for ourself and
+                    # incremented for the new fetcher. These cancel out, so we leave it
+                    # alone.
+                else:
+                    self._event_fetch_ongoing -= 1
+                    event_fetch_ongoing_gauge.set(self._event_fetch_ongoing)
+
+            if should_restart:
+                run_as_background_process("fetch_events", self._do_fetch)
 
     def _do_fetch_txn(self, conn: Connection) -> None:
         """Takes a database connection and waits for requests for events from
@@ -776,8 +793,6 @@ class EventsWorkerStore(SQLBaseStore):
                         or single_threaded
                         or i > EVENT_QUEUE_ITERATIONS
                     ):
-                        self._event_fetch_ongoing -= 1
-                        event_fetch_ongoing_gauge.set(self._event_fetch_ongoing)
                         return
                     else:
                         self._event_fetch_lock.wait(EVENT_QUEUE_TIMEOUT_S)
--- a/synapse/storage/databases/main/events_worker.py
+++ b/synapse/storage/databases/main/events_worker.py
@@ -736,49 +736,56 @@ class EventsWorkerStore(SQLBaseStore):
             for e in state_to_include.values()
         ]
 
-    async def _do_fetch(self) -> None:
-        """Services requests for events from the `_event_fetch_list` queue."""
-        try:
-            await self.db_pool.runWithConnection(self._do_fetch_txn)
-        except Exception as e:
-            failed_event_list = []
-            with self._event_fetch_lock:
-                if self._event_fetch_ongoing == 1 and self._event_fetch_list:
-                    # We are the last remaining fetcher and we have just failed.
-                    # Fail any outstanding fetches, since they won't get processed
-                    # otherwise.
-                    failed_event_list = self._event_fetch_list
-                    self._event_fetch_list = []
-
-                self._event_fetch_ongoing -= 1
+    def _start_fetch_thread(self) -> None:
+        """Starts an event fetcher."""
+        with self._event_fetch_lock:
+            if (
+                self._event_fetch_list
+                and self._event_fetch_ongoing < EVENT_QUEUE_THREADS
+            ):
+                self._event_fetch_ongoing += 1
                 event_fetch_ongoing_gauge.set(self._event_fetch_ongoing)
+                should_start = True
+            else:
+                should_start = False
 
-            for _, deferred in failed_event_list:
-                if not deferred.called:
-                    with PreserveLoggingContext():
-                        deferred.errback(e)
-
-            raise
-        else:
-            should_restart = False
-            with self._event_fetch_lock:
-                if self._event_fetch_ongoing == 1 and self._event_fetch_list:
-                    # An event fetch has been queued, but we're the last remaining
-                    # fetcher and have already decided to terminate.
-                    # Start a new fetcher.
-                    should_restart = True
-
-                    # `_event_fetch_ongoing` ought to be decremented for ourself and
-                    # incremented for the new fetcher. These cancel out, so we leave it
-                    # alone.
-                else:
+        async def _do_fetch() -> None:
+            """Services requests for events from the `_event_fetch_list` queue."""
+            exc = None
+            try:
+                await self.db_pool.runWithConnection(self._do_fetch)
+            except Exception as e:
+                exc = e
+                raise
+            finally:
+                should_restart = False
+                failed_event_list = []
+                with self._event_fetch_lock:
                     self._event_fetch_ongoing -= 1
                     event_fetch_ongoing_gauge.set(self._event_fetch_ongoing)
 
-            if should_restart:
-                run_as_background_process("fetch_events", self._do_fetch)
+                    if self._event_fetch_ongoing == 0 and self._event_fetch_list:
+                        # We are the last remaining fetcher and we are about to
+                        # go away. Deal with any outstanding fetches.
+                        if exc is None:
+                            should_restart = True
+                        else:
+                            failed_event_list = self._event_fetch_list
+                            self._event_fetch_list = []
 
-    def _do_fetch_txn(self, conn: Connection) -> None:
+                if should_restart:
+                    self._start_fetch_thread()
+
+                if exc is not None:
+                    for _, deferred in failed_event_list:
+                        if not deferred.called:
+                            with PreserveLoggingContext():
+                                deferred.errback(exc)
+
+        if should_start:
+            run_as_background_process("fetch_events", _do_fetch)
+
+    def _do_fetch(self, conn: Connection) -> None:
         """Takes a database connection and waits for requests for events from
         the _event_fetch_list queue.
         """
@@ -1021,18 +1028,9 @@ class EventsWorkerStore(SQLBaseStore):
         events_d = defer.Deferred()
         with self._event_fetch_lock:
             self._event_fetch_list.append((events, events_d))
-
             self._event_fetch_lock.notify()
 
-            if self._event_fetch_ongoing < EVENT_QUEUE_THREADS:
-                self._event_fetch_ongoing += 1
-                event_fetch_ongoing_gauge.set(self._event_fetch_ongoing)
-                should_start = True
-            else:
-                should_start = False
-
-        if should_start:
-            run_as_background_process("fetch_events", self._do_fetch)
+        self._start_fetch_thread()
 
         logger.debug("Loading %d events: %s", len(events), events)
         with PreserveLoggingContext():
--- a/synapse/storage/databases/main/events_worker.py
+++ b/synapse/storage/databases/main/events_worker.py
@@ -750,37 +750,65 @@ class EventsWorkerStore(SQLBaseStore):
                 should_start = False
 
         async def _fetch_thread() -> None:
-            """Services requests for events from the `_event_fetch_list` queue."""
+            """Services requests for events from `_event_fetch_list`."""
             exc = None
             try:
                 await self.db_pool.runWithConnection(self._fetch_loop)
-            except Exception as e:
+            except BaseException as e:
                 exc = e
                 raise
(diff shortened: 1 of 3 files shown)


synapse/storage/databases/main/events_worker.py
```diff
@@ -602,7 +602,7 @@ async def _get_events_from_cache_or_db(
             # already due to `_get_events_from_db`).
             fetching_deferred: ObservableDeferred[
                 Dict[str, _EventCacheEntry]
-            ] = ObservableDeferred(defer.Deferred())
+            ] = ObservableDeferred(defer.Deferred(), consumeErrors=True)
```
squahtx, line 605: Without `consumeErrors`, twisted thinks that exceptions don't get handled.

synapse/storage/databases/main/events_worker.py
```diff
@@ -736,35 +736,56 @@ async def get_stripped_room_state_from_event_context(
...
+                    # them.
+                    failed_event_list = self._event_fetch_list
+                    self._event_fetch_list = []
+                else:
+                    failed_event_list = []
+
+            for _, deferred in failed_event_list:
+                if not deferred.called:
+                    with PreserveLoggingContext():
+                        deferred.errback(e)
```
squahtx, line 760: This is still wrong, in a different way. If `runWithConnection` raises *after* `_do_fetch_txn` successfully exits, we will decrement `_event_fetch_ongoing` twice.

On the other hand, if we move the decrement out of `_event_fetch_ongoing`, then there is a race:
  1. All fetchers have decided to terminate
  2. An event fetch is queued up, but doesn't start a new fetcher
  3. All fetchers terminate, and that event fetch fails.

synapse/storage/databases/main/events_worker.py
```diff
@@ -736,35 +736,71 @@ async def get_stripped_room_state_from_event_context(
...
+                    failed_event_list = self._event_fetch_list
+                    self._event_fetch_list = []
+
+                self._event_fetch_ongoing -= 1
+                event_fetch_ongoing_gauge.set(self._event_fetch_ongoing)
+
+            for _, deferred in failed_event_list:
+                if not deferred.called:
+                    with PreserveLoggingContext():
+                        deferred.errback(e)
```
squahtx, line 759: You forgot to reraise the exception.

synapse/storage/databases/main/events_worker.py
```diff
@@ -736,35 +736,73 @@ async def get_stripped_room_state_from_event_context(
...
+                    # fetcher and have already decided to terminate.
+                    # Start a new fetcher.
+                    should_restart = True
+
+                    # `_event_fetch_ongoing` ought to be decremented for ourself and
+                    # incremented for the new fetcher. These cancel out, so we leave it
+                    # alone.
+                else:
+                    self._event_fetch_ongoing -= 1
+                    event_fetch_ongoing_gauge.set(self._event_fetch_ongoing)
```
erikjohnston, line 776: Can this happen in a single place please? Probably in a finally block? I do wonder if maybe we should move this into the same function that increments the `self._event_fetch_ongoing` in the first place?

squahtx: I had a go at refactoring the code. Let me know if it looks alright!

synapse/storage/databases/main/events_worker.py
richvdh: one has to wonder if there is ever a usecase for ObservableDeferred to *not* `consumeErrors`.

This inspired me to write https://github.com/matrix-org/synapse/issues/11390.

synapse/storage/databases/main/events_worker.py
```diff
@@ -736,35 +736,80 @@ async def get_stripped_room_state_from_event_context(
...
+                        else:
+                            failed_event_list = self._event_fetch_list
+                            self._event_fetch_list = []
+
+                if should_restart:
+                    self._start_fetch_thread()
+
+                if exc is not None:
+                    for _, deferred in failed_event_list:
+                        if not deferred.called:
```
richvdh, line 781: A comment as to why the deferred might already have been called would be helpful.

synapse/storage/databases/main/events_worker.py
```diff
@@ -736,35 +736,80 @@ async def get_stripped_room_state_from_event_context(
...
+                            with PreserveLoggingContext():
```
richvdh, line 782: I'd be tempted to hoist this outside the loop. (I'm not sure it's needed at all, but it's certainly not needed for each iteration.)

synapse/storage/databases/main/events_worker.py
```diff
@@ -736,35 +736,80 @@ async def get_stripped_room_state_from_event_context(
...
+                    if (
+                        not self.USE_DEDICATED_DB_THREADS_FOR_EVENT_FETCHING
+                        or single_threaded
+                        or i > EVENT_QUEUE_ITERATIONS
+                    ):
+                        return
+                    else:
+                        self._event_fetch_lock.wait(EVENT_QUEUE_TIMEOUT_S)
+                        i += 1
+                        continue
```
richvdh, line 809: while we're refactoring:

```suggestion
                    self._event_fetch_lock.wait(EVENT_QUEUE_TIMEOUT_S)
                    i += 1
                    continue
```

synapse/storage/databases/main/events_worker.py
```diff
@@ -736,35 +736,80 @@ async def get_stripped_room_state_from_event_context(
...
+                exc = e
+                raise
+            finally:
+                should_restart = False
+                failed_event_list = []
+                with self._event_fetch_lock:
+                    self._event_fetch_ongoing -= 1
+                    event_fetch_ongoing_gauge.set(self._event_fetch_ongoing)
+
+                    if self._event_fetch_ongoing == 0 and self._event_fetch_list:
```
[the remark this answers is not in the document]
richvdh: similarly, if there *is* an exception, it looks like we leave the remaining entries in the queue for other threads to pick up - but they may not do so for up to `EVENT_QUEUE_TIMEOUT_S`. That's not an implausible thing to do (backing off after an exception sounds good), but let's comment that it's happening.

synapse/storage/databases/main/events_worker.py
```diff
@@ -736,35 +736,80 @@ async def get_stripped_room_state_from_event_context(
...
+                        if exc is None:
+                            should_restart = True
+                        else:
+                            failed_event_list = self._event_fetch_list
+                            self._event_fetch_list = []
+
+                if should_restart:
+                    self._start_fetch_thread()
+
+                if exc is not None:
```
richvdh, line 779: comment please. What caused us to get to this point, and why do we need to `errback` `failed_event_list` ?

tests/storage/databases/main/test_events_worker.py
```diff
@@ -144,3 +157,108 @@ def test_dedupe(self):
...
+                    "event_json",
+                    {"event_id": event_id},
+                    {
+                        "room_id": self.room_id,
+                        "json": json.dumps({"type": "test", "room_id": self.room_id}),
+                        "internal_metadata": "{}",
+                        "format_version": EventFormatVersions.V3,
+                    },
+                )
+            )
```
richvdh, line 211: for this sort of thing, I'd suggest calling the storage functions (`hs.get_datastore().store_room()`,  `hs.get_storage().persistence.persist_events()`), to decouple the tests a bit from the raw database tables.

tests/storage/databases/main/test_events_worker.py
```diff
@@ -144,3 +157,108 @@ def test_dedupe(self):
...
+
+    @contextmanager
+    def _outage(self) -> Generator[None, None, None]:
+        """Simulate a database outage."""
```
richvdh, line 215:
```suggestion
        """Simulate a database outage.
        
        Returns a context manager. While the context is active, any attempts to connect to the database will fail.
        """
```

tests/storage/databases/main/test_events_worker.py
```diff
@@ -144,3 +157,108 @@ def test_dedupe(self):
...
+        connection_pool = self.store.db_pool._db_pool
+        connection_pool.close()
+        connection_pool.start()
```
richvdh, line 218: what does this do?

richvdh (changes requested):
generally, the logic here seems sound.

However, this sort of stuff is fiddly and non-obvious; it is the sort of thing where detailed comments explaining what is going on and how it is meant to work will pay dividends next time we come to look at it, so I strongly encourage adding a load!

synapse/storage/databases/main/events_worker.py
```diff
@@ -736,35 +736,80 @@ async def get_stripped_room_state_from_event_context(
...
+        i = 0
+        while True:
+            with self._event_fetch_lock:
+                event_list = self._event_fetch_list
+                self._event_fetch_list = []
+
+                if not event_list:
+                    single_threaded = self.database_engine.single_threaded
+                    if (
+                        not self.USE_DEDICATED_DB_THREADS_FOR_EVENT_FETCHING
```
richvdh, line 801: I think `USE_DEDICATED_DB_THREADS_FOR_EVENT_FETCHING` might have outlived its purpose. Certainly it no longer seems to match the comment. I'd be interested to know if something has changed here, and whether we should update the comment or get rid of it entirely.

I'm a bit concerned its presence invalidates your carefully constructed test case.

synapse/storage/databases/main/events_worker.py
```diff
@@ -736,35 +736,80 @@ async def get_stripped_room_state_from_event_context(
...
-                            i += 1
-                            continue
-                    i = 0
+        i = 0
+        while True:
+            with self._event_fetch_lock:
+                event_list = self._event_fetch_list
+                self._event_fetch_list = []
+
+                if not event_list:
```
richvdh, line 798:
```suggestion
                if not event_list:
                    # there are no requests waiting. If we haven't yet reached the maximum iteration limit,
                    # wait for some more requests to turn up. Otherwise, bail out.
```

synapse/storage/databases/main/events_worker.py
```diff
@@ -736,35 +736,80 @@ async def get_stripped_room_state_from_event_context(
...
+                self._event_fetch_list
+                and self._event_fetch_ongoing < EVENT_QUEUE_THREADS
+            ):
+                self._event_fetch_ongoing += 1
+                event_fetch_ongoing_gauge.set(self._event_fetch_ongoing)
+                should_start = True
+            else:
+                should_start = False
+
+        async def _do_fetch() -> None:
```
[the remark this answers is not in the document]
squahtx: It was to bring the decrement closer to the increment

synapse/storage/databases/main/events_worker.py
squahtx: I'm not actually sure. It's copied from the exception path inside the transaction, except it's not clear to me why it's necessary there either. I shall remove it.

tests/storage/databases/main/test_events_worker.py
squahtx: That gets a bit messy with the sqlite in-memory database. All the data gets lost when the connection pool gets closed during the outage and I can't see an obvious way to ensure that the room ID and event IDs are the same when recreating the data.

synapse/storage/databases/main/events_worker.py
```diff
@@ -736,35 +736,80 @@ async def get_stripped_room_state_from_event_context(
             for e in state_to_include.values()
         ]
 
+    def _start_fetch_thread(self) -> None:
+        """Starts an event fetcher."""
```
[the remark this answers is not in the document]
squahtx: Renamed!

synapse/storage/databases/main/events_worker.py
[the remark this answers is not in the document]
squahtx: Renamed them to `_fetch_thread` and `_fetch_loop`

synapse/storage/databases/main/events_worker.py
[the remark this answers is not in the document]
squahtx: > if there is no exception, is it correct that we only restart if `_event_fetch_ongoing == 0` ?
Yes and no
No, because if there are other threads remaining, `_event_fetch_list` only gets processed once one of those threads completes its current work, however long that takes.
Yes, because otherwise we could restart spuriously if `_enqueue_events` has already started a worker for the new work. This can happen because `_maybe_start_fetch_thread` doesn't claim the work when it starts the thread.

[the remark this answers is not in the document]
squahtx: I've rejigged it to always restart on a clean exit when there is more work remaining and added a bunch of comments.

synapse/storage/databases/main/events_worker.py
squahtx: added a comment

tests/storage/databases/main/test_events_worker.py
squahtx: `close` closes all connection pool connections and also shuts down the db threadpool, which is unfortunate.
`start` restores the threadpool.

I've added a comment

synapse/storage/databases/main/events_worker.py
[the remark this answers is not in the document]
squahtx: comments added

synapse/storage/databases/main/events_worker.py
squahtx: I don't think much has changed [since it was introduced](https://github.com/matrix-org/synapse/pull/8450). It looks like a flag to prevent the idling behaviour during tests, since database transactions get run on the main thread.

squahtx: Neither of the two new tests test the threading behaviour deeply enough for it to make a difference

synapse/storage/databases/main/events_worker.py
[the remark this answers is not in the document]
richvdh: personally, I'm of the opinion that not having to grok inner functions outweighs the fact that the increment and decrement are spatially separated - particularly if we add (more) comments to signpost the logic. However, I don't feel strongly and clearly @erikjohnston differs.

synapse/storage/databases/main/events_worker.py
```diff
@@ -736,35 +736,111 @@ async def get_stripped_room_state_from_event_context(
...
+                        else:
+                            # The exception is a `SystemExit`, `KeyboardInterrupt` or
+                            # `GeneratorExit`. Don't try to do anything clever here.
+                            pass
+
+                if should_restart:
+                    # We exited cleanly but noticed more work.
+                    self._maybe_start_fetch_thread()
+
+                if exc is not None and event_fetches_to_fail:
```
richvdh, line 806: the alignment is hard to see, but is this outside the `finally` block? It's unreachable if so - the exception will be raised at line 759

tests/storage/databases/main/test_events_worker.py
richvdh: I see. Might be worth adding a comment at the top of `_populate_events` to help out future readers.

synapse/storage/databases/main/events_worker.py
richvdh: fair enough. I'm still giving it side-eye, but let's leave it alone for now.

richvdh (commented):
a few minor things, but looks generally good. Particular thanks for the detailed comments here!

synapse/storage/databases/main/events_worker.py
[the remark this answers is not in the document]
erikjohnston: (If bringing the increment and decrement together is turning out to be painful I don't feel particularly strongly about it)

synapse/storage/databases/main/events_worker.py
squahtx: It's inside the `finally` block still

synapse/storage/databases/main/events_worker.py
```diff
@@ -736,35 +736,111 @@ async def get_stripped_room_state_from_event_context(
...
+                            # We decided to exit, but then some more work was added
+                            # before `_event_fetch_ongoing` was decremented.
+                            # If a new event fetch thread was not started, we should
+                            # restart ourselves since the remaining event fetch threads
+                            # may take a while to get around to the new work.
+                            # Unfortunately it is not possible to tell whether a new
+                            # event fetch thread was started, so we restart
+                            # unconditionally. If we are unlucky, we will end up with
+                            # extra idle threads holding database connections for up to
+                            # `EVENT_QUEUE_ITERATIONS * EVENT_QUEUE_TIMEOUT_S` seconds.
```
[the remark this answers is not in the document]
squahtx: `EVENT_QUEUE_THREADS` is indeed respected. I was trying to convey that more than one fetch thread may run down this path at the same time.

synapse/storage/databases/main/events_worker.py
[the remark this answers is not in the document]
squahtx: I'll un-inner function `_fetch_thread` since it makes `_maybe_start_fetch_thread` easier to read.

richvdh (approved):
looks great, let's ship it!

PR #11333, opened 2021-11-13
Remove legacy code related to deprecated `trust_identity_server_for_password_resets` config flag

H-Shay:
Fixes #10545. As the title says, removes code associated with handling the deprecated config option, and Synapse refuses to start if this flag is set in the config.

PR #11377, opened 2021-11-17
Add missing type hints to config base classes

clokep:
This adds some missing type hints to `synapse.config._base`, there's a pyi version of this but I'm not really sure why? It lies about what's on `RootConfig` (by defining what's on `HomeServerConfig`). I didn't attempt to change any of that, just made it more correct.

This shook loose some errors which I then fixed.


synapse/config/emailconfig.py
```diff
@@ -147,11 +147,13 @@ def read_config(self, config, **kwargs):
             and config.get("trust_identity_server_for_password_resets", False) is True
         ):
             # Use the first entry in self.trusted_third_party_id_servers instead
-            if self.trusted_third_party_id_servers:
+            if self.root.registration.trusted_third_party_id_servers:
```
clokep, line 150: This was a real bug, but seems like it is getting removed in #11333 anyway.

synapse/config/tls.py
```diff
@@ -295,7 +295,7 @@ def read_tls_certificate(self) -> crypto.X509:
         cert_path = self.tls_certificate_file
         logger.info("Loading TLS certificate from %s", cert_path)
         cert_pem = self.read_file(cert_path, "tls_certificate_path")
-        cert = crypto.load_certificate(crypto.FILETYPE_PEM, cert_pem)
+        cert = crypto.load_certificate(crypto.FILETYPE_PEM, cert_pem.encode())
```
clokep, line 298: I have a feeling that `load_certificate` takes both strings and bytes, but is declared to only take bytes? Not 100% sure what's going on here.

synapse/module_api/__init__.py
```diff
@@ -903,7 +903,7 @@ def read_templates(
             A list containing the loaded templates, with the orders matching the one of
             the filenames parameter.
         """
-        return self._hs.config.read_templates(
+        return self._hs.config.server.read_templates(
```
clokep, line 906: This was a real since #10985 was merged for v1.45.0rc1 -- `read_templates` would not have worked for modules since then.

synapse/config/tls.py
reivilibre: As long as you're meeting its contract and `cert_pem` is a `str` here then this seems perfectly reasonable.

I'm the worst offender of this but we ought to stop second-guessing type signatures (yes yes, I know they're often genuinely wrong too :P) — support for strings might be a backwards compatibility thing, for example, so this seems like the Right Thing To Do™ here.

synapse/module_api/__init__.py
reivilibre: at least you're doing some good typing work with non-academic benefits! :)

reivilibre (approved):

synapse/config/tls.py
clokep: My guess is that it is a legacy code-path for Python 2. I'll look a bit more into it.

synapse/config/emailconfig.py
clokep: Yes, I can wait until that merges first. 👍

synapse/config/tls.py
clokep: It seems they both work, weirdly another place in this file we manually open the cert for binary reading. I wonder if that would make mores sense to do here.

reivilibre: > It seems they both work, weirdly another place in this file we manually open the cert for binary reading. I wonder if that would make mores sense to do here.

well now that you mention it... yeah that could make sense rather than decoding and reencoding :)

clokep:
Note that this is currently failing due to #11395.

synapse/config/tls.py
clokep: Hmm, OK. I might revert this for now then and handle this in a future PR.

Issue #11373, opened 2021-11-17
servers cannot re-join rooms after rotating signing keys and clearing database


richvdh:
if a server is re-provisioned (so that the signing key is replaced, and the event store is cleared), then it cannot re-join any rooms it was previously participating in, because it cannot validate the auth chain:

```
2021-11-17 15:45:59,186 - synapse.federation.federation_base - 63 - WARNING - POST-150 - Signature check failed for $16371629950pDraD:synapse-performance-test.lab.element.dev: 403: event id $16371629950pDraD:synapse-performance-test.lab.element.dev: unable to verify signature for sender synapse-performance-test.lab.element.dev: 401: Invalid signature for server synapse-performance-test.lab.element.dev with key ed25519:a_qdcq: Missing signature for synapse-performance-test.lab.element.dev, ed25519:a_qdcq
2021-11-17 15:45:59,187 - synapse.federation.federation_base - 63 - WARNING - POST-150 - Signature check failed for $16371620780tvsmk:synapse-performance-test.lab.element.dev: 403: event id $16371620780tvsmk:synapse-performance-test.lab.element.dev: unable to verify signature for sender synapse-performance-test.lab.element.dev: 401: Invalid signature for server synapse-performance-test.lab.element.dev with key ed25519:a_qdcq: Missing signature for synapse-performance-test.lab.element.dev, ed25519:a_qdcq
2021-11-17 15:45:59,333 - synapse.http.matrixfederationclient - 684 - WARNING - POST-150 - {GET-O-74} [synapse-performance-test.lab.element.dev] Request failed: GET matrix://synapse-performance-test.lab.element.dev/_matrix/federation/v1/event/%2416371629950pDraD%3Asynapse-performance-test.lab.element.dev: HttpResponseException('404: Not Found')
2021-11-17 15:45:59,336 - synapse.http.matrixfederationclient - 684 - WARNING - POST-150 - {GET-O-75} [synapse-performance-test.lab.element.dev] Request failed: GET matrix://synapse-performance-test.lab.element.dev/_matrix/federation/v1/event/%2416371620780tvsmk%3Asynapse-performance-test.lab.element.dev: HttpResponseException('404: Not Found')
[...]

PR #11379, opened 2021-11-17
Fix verification of objects signed with old local keys

richvdh:
Fixes a bug introduced in #11129: objects signed by the local server, but with keys other than the current one, could not be successfully verified.

We need to check the key id in the signature, and track down the right key.

Fixes: #11373


reivilibre (approved):
Seems sensible to me!

PR #11380, opened 2021-11-17
Publish a `develop` docker image

richvdh:
I'd find it helpful to have a docker image corresponding to current develop, without having to build my own.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -5,7 +5,7 @@ name: Build docker images
 on:
   push:
     tags: ["v*"]
-    branches: [ master, main ]
+    branches: [ master, main, develop ]
   workflow_dispatch:
 
 permissions:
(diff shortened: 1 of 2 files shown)


DMRobertson (approved):
I'm slightly anxious that the ARM build takes a while, and that this means we'll be running it a lot more frequently in CI and limit the CI throughput. (I think we only have a certain number of runners available to the org as a whole?) 

But that's just speculation. It's probably fine?

babolivier:
>I'm slightly anxious that the ARM build takes a while, and that this means we'll be running it a lot more frequently in CI and limit the CI throughput. (I think we only have a certain number of runners available to the org as a whole?)
>
> But that's just speculation. It's probably fine?

Worst case scenario, if we see it causing too much trouble, we can always make it only build non-arm later on. I'd say for now let's go with arm builds, since it's useful for people building on new apple machines.

PR #11382, opened 2021-11-18
Keep fallback key marked as used if it's re-uploaded

uhoreg:
This prevents the following situation:

- client uploads a fallback key.  Network dies, and client doesn't get acknowledgement that the upload succeeded, even though it did
- someone does a `keys/claim` and the fallback key gets marked as used
- client comes back later and tries to re-upload the key
- the key gets replaced and marked as unused
- the client doesn't get notified that the key was used, because the server thinks it is unused

This PR changes the behaviour so that it compares the uploaded key with the existing key, and if it's the same, it doesn't touch it.  Notably, this means that the key will remain marked as used, so that the client will get notified in the `/sync` that it needs to upload a new key.

### Pull Request Checklist

### DIFF
--- a/changelog.d/11382.misc
+++ b/changelog.d/11382.misc
@@ -1,0 +1,1 @@
+Keep fallback key marked as used if it's re-uploaded.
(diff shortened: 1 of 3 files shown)


DMRobertson:
Great spot. Are there security ramifications for this problem? At first glance, it sounds sufficiently important that we should get a complement test written so that other homeservers can benefit from the test.

It'd be great to get CI running on this---I think we just need a newsfile in order to do so.

uhoreg:
sytest failure seems to be a flaky test.  This should be ready for review now.

reivilibre (approved):
Thanks!

PR #11386, opened 2021-11-18
Use auto_attribs for RefreshTokenLookupResult

reivilibre:


DMRobertson (approved):
Ship it!

(I'd like to complain that `dataclasses` provides options for `frozen=True` and `slots=True` though.)

DMRobertson:
> Ship it!
> 
> (I'd like to complain that `dataclasses` provides options for `frozen=True` and `slots=True` though.)

... assuming CI passes

reivilibre:
> .. dataclasses ..

Is there a benefit to using dataclasses over attrs?

DMRobertson:
> > .. dataclasses ..
> 
> Is there a benefit to using dataclasses over attrs?

They're in the stdlib, which saves a dependency if you're not using any of the fuller attrs machinery.

I reckon you can make them `Generic` while still having `slots=True`, which would work around that perennial attrs problem we keep encountering.

(Not proposing we change anything, it's just a perennial grumble.)

DMRobertson:
> edit: seems like Py 3.6 support is a winner for attrs over dataclasses

Oh true. There's a [backport](https://pypi.org/project/dataclasses/) but I'm not sure how fully featured it is.

reivilibre:
> which saves a dependency if you're not using any of the fuller attrs machinery.

I think we do (or will ;-) if I ever get my validattrs stuff in) use fuller machinery, so not sure this is a terrible plight

PR #11388, opened 2021-11-18
Rename unstable `access_token_lifetime` configuration option to `refreshable_access_token_lifetime` to make it clear it only concerns refreshable access tokens.

reivilibre:
This should be OK because the feature is unstable and currently does not match the MSC or the (now stabilised spec).
There are also no client implementations other than a stale PR on Hydrogen (afaict) and nobody appears to have used the feature on matrix.org even though it was enabled for some time.

I'm also not doing this needlessly; I'm planning to add 2 other kinds of lifetime:

* non-refreshable access token lifetime
* refresh token lifetime

### DIFF
--- a/synapse/handlers/register.py
+++ b/synapse/handlers/register.py
@@ -116,7 +116,9 @@ class RegistrationHandler:
             self.pusher_pool = hs.get_pusherpool()
 
         self.session_lifetime = hs.config.registration.session_lifetime
-        self.access_token_lifetime = hs.config.registration.access_token_lifetime
+        self.access_token_lifetime = (
+            hs.config.registration.refreshable_access_token_lifetime
+        )
 
         init_counters_for_auth_provider("")
 
--- a/synapse/handlers/register.py
+++ b/synapse/handlers/register.py
@@ -116,7 +116,7 @@ class RegistrationHandler:
             self.pusher_pool = hs.get_pusherpool()
 
         self.session_lifetime = hs.config.registration.session_lifetime
-        self.access_token_lifetime = (
+        self.refreshable_access_token_lifetime = (
             hs.config.registration.refreshable_access_token_lifetime
         )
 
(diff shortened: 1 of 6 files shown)


clokep:
A couple of thoughts (these aren't requests, just trying to start a discussion):

* Should we error / warn / something if the old name is set?
* Do we want to nest these new settings instead of having them in the root config?

reivilibre:
> Should we error / warn / something if the old name is set?

We could, but my main concern was that doing this for every experimental config option ever gets a bit crufty. Not opposed to it though.

> Do we want to nest these new settings instead of having them in the root config?

Could do, don't really know what the nested name would be though (plus `session_lifetime` is already in `registration` unnested and it's a related option, so it'd feel weird to nest 3 out of 4 of them? Equally, trying to move that one seems like a faffle.)

Open to opinions

richvdh:
> Should we error / warn / something if the old name is set?

If this were a thing that we thought had been widely used, I'd be in favour of doing so. Given it's not documented, and we haven't heard of anyone trying it out, I'm inclined not to bother.

> Do we want to nest these new settings instead of having them in the root config?

I don't think there's a great deal to be gained by nesting them. Grouping them next to each other in the sample config should be sufficient.

Speaking of which, now that the MSC has landed, should we document the config settings? (Or maybe that's a different PR?)

reivilibre:
> should we document the config settings? (Or maybe that's a different PR?)

Yeah, I'm going to do all the polishings when I update this to match the spec'd version (some minor changes, plus removing the unstable prefix). Until then, it's still an unstable feature in Synapse :-)

synapse/handlers/register.py
```diff
@@ -116,7 +116,9 @@ def __init__(self, hs: "HomeServer"):
             self.pusher_pool = hs.get_pusherpool()
 
         self.session_lifetime = hs.config.registration.session_lifetime
-        self.access_token_lifetime = hs.config.registration.access_token_lifetime
+        self.access_token_lifetime = (
```
clokep, line 119: Should we rename the property on `RegistrationHandler` to match?

(Same goes for `RefreshTokenServlet` below.)

clokep (approved):
Seems reasonable.

Issue #11375, opened 2021-11-17
Improve performance of event_thread_relation background update for large databases


bradtgmurray:
### Description
Been debugging some performance issues on our RDS database and came across this ugly chart.
![image](https://user-images.githubusercontent.com/33967/142258065-19af4392-7f75-4f0a-a23b-d9e21205260b.png)

pg_stat_statements say it's this query here:
`SELECT event_id, json FROM event_json LEFT JOIN event_relations USING (event_id) WHERE event_id > '$iC2wL1TGv8qPaZXR8-KMSNuKxxgLLkhru3HeUeetEPI' AND event_relations.event_id IS NULL ORDER BY event_id LIMIT 100`

Which is the top query on our DB by a factor of 3, as it needs to do a surprisingly slow and large index scan of the index on event_relations to see if there are any that exist for the events we'd like to process.

```
Limit  (cost=1.12..90.44 rows=100 width=1064) (actual time=2295.323..2295.979 rows=100 loops=1)
  ->  Merge Anti Join  (cost=1.12..20178406.57 rows=22592023 width=1064) (actual time=2295.322..2295.968 rows=100 loops=1)
        Merge Cond: (event_json.event_id = event_relations.event_id)
        ->  Index Scan using event_json_event_id_key on event_json  (cost=0.57..19154088.46 rows=24493063 width=1064) (actual time=0.022..0.630 rows=108 loops=1)
              Index Cond: (event_id > '$iC2wL1TGv8qPaZXR8-KMSNuKxxgLLkhru3HeUeetEPI'::text)
        ->  Index Only Scan using event_relations_id on event_relations  (cost=0.56..926766.22 rows=6923535 width=44) (actual time=0.003..1908.093 rows=5003206 loops=1)
              Heap Fetches: 778015
Planning Time: 0.153 ms
Execution Time: 2296.006 ms
```

2 seconds to scan through the whole event_relations_id index looking for entries that don't exist.

Note our database is quite large. Our events_json table is about 145gb alone, and we're only on the $i events in this background update and we already have 7 million rows in our event_relations table.

It feels like the batch size calculation here is actually hurting us: https://github.com/matrix-org/synapse/blob/develop/synapse/storage/background_updates.py#L274

The query takes so long that even though we only want it to take 100ms, it takes 2s+, so we always only ever do the minimum amount (100 entries). However, the nature of the query means that we're scanning the same index over and over again to find pretty small sets of entries.

[...]

PR #11391, opened 2021-11-18
Store arbitrary relations from events

### LINKED ISSUE #11387: Rework relations storage and background update to handle any relation type
The `event_thread_relation` background update is fairly heavy and only looks for the unstable `io.element.thread` relation. We will need to essentially re-run this when [MSC3440](https://github.com/matrix-org/matrix-doc/pull/3440) is stabilized to search for `m.thread` relations instead.

It seems a better way to do this would be to store unknown relation types into the database (and potentially filter them on fetch) so that we do not need to crawl all events in the future when more relation types are added. If a future relation has additional data which must be processed, similar to the `key` field of `m.annotation`, then this would allow easily re-processing only the events with that relation.

Note that this was previously discussed at https://github.com/matrix-org/synapse/pull/11088#discussion_r729040616 but rejected. This is being reconsidered in light of #11375 and other problems with running this background update.

The overall plan for how to accomplish this would be:

1. Replace the current background update with a no-op, this allows any servers which have not finished it to give up.
2. Write a new background update that does something similar, but pulls in any relation type.
3. Rework the code in `_handle_event_relations` to save any relation of new events (instead of only known relations).
4. (Optionally) Modify places where we fetch relations to limit to only "known" relations.

This also has the benefit of allowing for easier experimentation with unknown relations. Synapse would (somewhat obviously) not include those in bundled relations since the format of those is custom per relation type.

clokep:
1. Persist any relation received on a message, even if the type is unknown.
2. Take the background update from #11181 (`event_thread_relation` and make it a no-op.
3. Add a new background update, similar to the one from #11181, but storing relations of any type.

The hope here is that if we add a new relation type (e.g. `m.foo`) we could search the `event_relations` table for `m.foo` events and do further processing on those events, instead of having to crawl all events again.

Note that this will have some of the same issues mentioned in #11375 as the method used in the background update is pretty much identical.

Fixes #11387


clokep:
I'd be curious if @richvdh or @erikjohnston have thoughts on how we can improve this query to make it either faster or less expensive!

synapse/storage/databases/main/events.py
```diff
@@ -1,6 +1,6 @@
 # Copyright 2014-2016 OpenMarket Ltd
 # Copyright 2018-2019 New Vector Ltd
```
richvdh, line 2: I believe we're supposed to combine all these (OM and NVL transferred the copyright to m.org)

```suggestion
```

synapse/storage/databases/main/events.py
```diff
@@ -1,6 +1,6 @@
...
-# Copyright 2019 The Matrix.org Foundation C.I.C.
+# Copyright 2019-2021 The Matrix.org Foundation C.I.C.
```
richvdh, line 3:
```suggestion
# Copyright 2014-2021 The Matrix.org Foundation C.I.C.
```

richvdh (commented):
generally seems sound to me, modulo a few comments.

synapse/storage/databases/main/events.py
clokep: I think the latest I've seen is that we're not supposed to and we should just be adding to it?

richvdh: apologies, have reread the wall of text on the wiki about it and sounds like you're right.

clokep: No problem! Seems guidance on this has flipped back and forth a few times!

synapse/storage/databases/main/events_bg_updates.py
```diff
@@ -1099,11 +1102,16 @@ def purged_chain_cover_txn(txn) -> int:
...
-        """Background update handler which will store thread relations for existing events."""
+    async def _event_arbitrary_relations(
+        self, progress: JsonDict, batch_size: int
+    ) -> int:
+        """Background update handler which will store previously unknown relations for existing events."""
         last_event_id = progress.get("last_event_id", "")
 
-        def _event_thread_relation_txn(txn: LoggingTransaction) -> int:
+        def _event_arbitrary_relations_txn(txn: LoggingTransaction) -> int:
+            # Iterate over events which do not appear in the event_relations
```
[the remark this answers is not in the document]
clokep: For future reference, an "anti-join" is joining and then filtering on the null column.

tests/rest/client/test_relations.py
```diff
@@ -811,3 +860,65 @@ def _create_user(self, localpart: str) -> Tuple[str, str]:
         access_token = self.login(localpart, "abc123")
 
         return user_id, access_token
+
+    def test_background_update(self):
```
clokep, line 864: Putting this in this file seems dirty, but it seems best to keep relations tests together? Maybe it would be better to start a storage test file and we can add to it in the future when fleshing out some unit tests?

synapse/storage/databases/main/events_bg_updates.py
```diff
@@ -1127,48 +1137,71 @@ def _event_thread_relation_txn(txn: LoggingTransaction) -> int:
...
-                values=[
-                    {
-                        "event_id": event_id,
-                        "relates_to_Id": parent_id,
-                        "relation_type": RelationTypes.THREAD,
-                    }
-                    for event_id, parent_id in missing_thread_relations
-                ],
-            )
+                # Iterate the parent IDs and invalidate caches.
```
clokep, line 1176: I don't know if we usually clear caches as part of background updates, but in this case I think it could be a bit broken without doing it.

clokep:
@richvdh Thanks for the review! I think I've handled your comments -- I've left a couple of notes of places I was unsure about.

synapse/storage/databases/main/events_bg_updates.py
```diff
@@ -1109,21 +1112,20 @@ async def _event_arbitrary_relations(
...
+                WHERE event_id > ?
                 ORDER BY event_id LIMIT ?
                 """,
                 (last_event_id, batch_size),
             )
 
             results = list(txn)
-            missing_relations = []
+            # (event_id, parent_id, rel_type) for each relation
+            missing_relations: List[Tuple[str, str, str]] = []
```
richvdh, line 1128: we don't know if they are actually missing? maybe just `relations`?

synapse/storage/databases/main/events_bg_updates.py
```diff
@@ -1140,32 +1142,49 @@ def _event_arbitrary_relations_txn(txn: LoggingTransaction) -> int:
                 if not relates_to or not isinstance(relates_to, dict):
                     continue
 
-                # The only expected relation type would be from threads, but
-                # there could be other unknown ones.
+                # If the relation type or parent event ID is not a string a
+                # string, skip it.
```
richvdh, line 1146:
```suggestion
                # If the relation type or parent event ID is not a string, skip it.
```

synapse/storage/databases/main/events_bg_updates.py
```diff
@@ -1140,32 +1142,49 @@ def _event_arbitrary_relations_txn(txn: LoggingTransaction) -> int:
...
+                #
+                # Do not consider relation types that have existed for a long time,
+                # only include the new thread relation and any unknown relations.
```
richvdh, line 1149:
```suggestion
                # Do not consider relation types that have existed for a long time,
                # since they will already be listed in the `event_relations` table.
```

synapse/storage/databases/main/events_bg_updates.py
```diff
@@ -1140,32 +1142,49 @@ def _event_arbitrary_relations_txn(txn: LoggingTransaction) -> int:
...
-                values=[
-                    {
-                        "event_id": event_id,
-                        "relates_to_Id": parent_id,
-                        "relation_type": rel_type,
-                    }
-                    for event_id, parent_id, rel_type in missing_relations
-                ],
-            )
+            # Insert the missing data, note that we upsert here in-case the event
```
richvdh, line 1164:
```suggestion
            # Insert the missing data, note that we upsert here in case the event
```

synapse/storage/databases/main/events_bg_updates.py
richvdh: seems sensible

tests/rest/client/test_relations.py
richvdh: I think it's best here.

richvdh (approved):
lgtm other than a couple of nits

synapse/storage/databases/main/events_bg_updates.py
clokep: Fair. Maybe `relations_to_insert` or something. 🤷

PR #7260, opened 2020-04-10
Check publishing permissions

PeerD:
Check on room creation if the user is allowed to publish the room to the room directory. See #7204 

`Signed-off-by: Peer Diestelhorst <signoff@spy.de>`

### Pull Request Checklist

* [x] Pull request is based on the develop branch
* [x] Pull request includes a [changelog file](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md#changelog). The entry should:
  - Be a short description of your change which makes sense to users. "Fixed a bug that prevented receiving messages from other servers." instead of "Moved X method from `EventStore` to `EventWorkerStore`.".
  - Use markdown where necessary, mostly for `code blocks`.
  - End with either a period (.) or an exclamation mark (!).
  - Start with a capital letter.
* [x] Pull request includes a [sign off](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md#sign-off)
* [x] Code style is correct (run the [linters](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md#code-style))

PR #11392, opened 2021-11-18
Fix checking whether a room can be published on creation.

clokep:
#7260 attempted to fix this, but the check was not correct, I'm actually surprised this hasn't caused an errors to appear anywhere, they would be of the form:

`TypeError: 'RoomAlias' object is not iterable`

### DIFF
--- a/synapse/config/room_directory.py
+++ b/synapse/config/room_directory.py
@@ -1,4 +1,5 @@
 # Copyright 2018 New Vector Ltd
+# Copyright 2021 Matrix.org Foundation C.I.C.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -12,6 +13,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from typing import List
+
+from synapse.types import JsonDict
 from synapse.util import glob_to_regex
 
 from ._base import Config, ConfigError
@@ -20,7 +24,7 @@ from ._base import Config, ConfigError
 class RoomDirectoryConfig(Config):
     section = "roomdirectory"
 
-    def read_config(self, config, **kwargs):
+    def read_config(self, config, **kwargs) -> None:
         self.enable_room_list_search = config.get("enable_room_list_search", True)
 
         alias_creation_rules = config.get("alias_creation_rules")
@@ -47,7 +51,7 @@ class RoomDirectoryConfig(Config):
                 _RoomDirectoryRule("room_list_publication_rules", {"action": "allow"})
             ]
 
-    def generate_config_section(self, config_dir_path, server_name, **kwargs):
+    def generate_config_section(self, config_dir_path, server_name, **kwargs) -> str:
         return """
         # Uncomment to disable searching the public room list. When disabled
         # blocks searching local and remote room lists for local and remote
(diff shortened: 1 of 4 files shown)


clokep:
I spun this out of a type hints PR since it was a real bug fix...

reivilibre:
Does it smell of missing a test if we didn't spot this? :S

synapse/config/room_directory.py
```diff
@@ -20,7 +24,7 @@
 class RoomDirectoryConfig(Config):
     section = "roomdirectory"
 
-    def read_config(self, config, **kwargs):
```
reivilibre, line 23: `config`? `kwargs` maybe?

synapse/config/room_directory.py
```diff
@@ -47,7 +51,7 @@ def read_config(self, config, **kwargs):
                 _RoomDirectoryRule("room_list_publication_rules", {"action": "allow"})
             ]
 
-    def generate_config_section(self, config_dir_path, server_name, **kwargs):
+    def generate_config_section(self, config_dir_path, server_name, **kwargs) -> str:
```
reivilibre, line 54:
```suggestion
    def generate_config_section(self, config_dir_path: str, server_name: str, **kwargs) -> str:
```
not sure about the `kwargs` but is that right?

reivilibre (approved):
Seems good though, thank you :)

synapse/config/room_directory.py
clokep: I have a separate branch which does all of the `read_config` and `generate_config_section` methods at once. Hope that's OK!

clokep:
> Does it smell of missing a test if we didn't spot this? :S

Yes, there was a test for this but the test had a bug in it (it wasn't publishing the room alias when it was supposed to).

I've pushed some additional tests and cleaned up the old ones a bit.

synapse/config/room_directory.py
reivilibre: aye that's reet :)

tests/handlers/test_directory.py
```diff
@@ -394,30 +393,24 @@ class TestCreateAliasACL(unittest.HomeserverTestCase):
...
     servlets = [directory.register_servlets, room.register_servlets]
 
-    def prepare(self, reactor, clock, hs):
-        # We cheekily override the config to add custom alias creation rules
-        config = {}
+    def default_config(self):
+        config = super().default_config()
+
+        # We cheekily override the config to add custom alias creation and
+        # room publication rules.
```
reivilibre, line 400: but you removed the room publication rules? Guess you changed this comment before invalidating it? :)

tests/handlers/test_directory.py
```diff
@@ -505,10 +522,24 @@ def test_allowed_with_publication_permission(self):
...
-            is_public=False,
+            is_public=True,
             expect_code=200,
         )
 
+    def test_denied_publication_with_invalid_alias(self):
+        """
+        Try to create a room, register an alias for it, and publish it,
+        as a user WITH permission to publish rooms.
+        (This is used as both a standalone test & as a helper function.)
```
reivilibre, line 533: Is it?

reivilibre (approved):
Looks good to me other than these two comments :)

tests/handlers/test_directory.py
clokep: Ah, yes. Good catch!

tests/handlers/test_directory.py
clokep: Arg, I copied and pasted this, removed it, deleted it all, then redid it. :(

PR #4384, opened 2019-01-14
Support ACME for certificate provisioning

hawkowl:

PR #11393, opened 2021-11-18
Remove dead code from acme support.

clokep:
I believe this should have been part of #10194. The only commits reference it are from #4384 and #10194. There's no current callers.

### DIFF
--- a/changelog.d/11393.misc
+++ b/changelog.d/11393.misc
@@ -1,0 +1,1 @@
+Remove dead code from supporting ACME.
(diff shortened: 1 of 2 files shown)


babolivier (approved):
Ah, I think I just assumed we were using this for non-acme certs as well, thanks for taking care of it!

clokep:
> Ah, I think I just assumed we were using this for non-acme certs as well, thanks for taking care of it!

No problem! I was trying to add type hints and PyCharm wasn't find any users...a quick `git log -Sis_disk_cert_valid` showed only those two commits!

babolivier:
(how have I lived so long without knowing about `git log -S`?)

Issue #11300, opened 2021-11-10
Refactor function parameters to describe the behavior instead of the state which is interpreted in non-obvious ways


MadLittleMods:
> as a general rule, I would say it is preferable for method parameters to describe a change in behaviour ("`inhibit_sync_to_clent`") rather than describe some bit of state that the function is free to interpret as it wishes: it's much clearer for a later reader to understand how things tie together, particularly when things then get modified later on. Obviously this isn't a hard-and-fast rule, but it's worth considering.
> 
> [Currently we have a bit of a mess in some places (`FederationEventHander.backfilled` being a prime example) where you have to dig through 150 methods to find out what a flag actually does, and then somehow reverse-engineer what it is *supposed* to do. Don't do that.]
> 
> *-- @richvdh, https://github.com/matrix-org/synapse/pull/11265#discussion_r746528719*

### Examples:

The interesting spots will be any of the `bool` function parameters, search for `: bool.*=?.*,`:

 `outlier` and `backfilled` can be refactored out to options like:
 
 - `inhibit_sync_to_client`
 - `inhibit_push_notifications`
 - `use_negative_stream_ordering`
 - etc
 
 
`is_peeking` could be refactored into

 - `inhibit_peekers_viewing_shared_rooms`
 - `include_membership`
 - etc

To refactor, it's probably easiest to dive down to the innermost function to find the actual functionality each function parameter triggers and pipe the new options to the top.

In many cases, we rely on `event.internal_metadata` with `outlier` (or `is_outlier`), `historical` (or `is_historical`), `soft_failed` (or `is_soft_failed`), etc. There are many side-effects from setting something as an `outlier` for example. We could write methods like `event.internal_metadata.should_proactively_send()` with these descriptive behaviors based on those properties, new ex. `event.internal_metadata.should_inhibit_sync_to_client()`. 

### Definition of done

 - [ ] Replace `backfilled` function parameters we pipe everywhere with specific behavior options
    - Related PR's:
       - https://github.com/matrix-org/synapse/pull/11417
 - [ ] Replace `outlier` with specific behavior options
 - [ ] Replace `historical` with specific behavior options (there will be some overlap with `outlier` and `backfilled`
 - [ ] Replace `is_peeking` with specific behavior options

---

Keywords:

 - Refactor logic to rely on descriptive behavior function parameters instead of interpreting a given field

richvdh:
[...]

PR #11396, opened 2021-11-19
Refactor `backfilled` into specific behavior function arguments pt.1

MadLittleMods:
Refactor `backfilled` into specific behavior function arguments pt.1. This PR does not replace all `backfilled` instances. Just `persist_events_and_notify` and all downstream calls

Using keyword-only arguments (`*`) to reduce the mistakes from `backfilled` being left as a positional argument somewhere and being interpreted wrong by our new arguments.

Part of https://github.com/matrix-org/synapse/issues/11300

### Dev notes

 - `persist_events_and_notify` (added `inhibit_push_notifications`)
    - `persist_events` and `persist_event`
       - `_event_persist_queue`

`_event_persist_queue` -> 

 - `_persist_event_batch` (added `should_calculate_state_and_forward_extrems`)
    - `_persist_events_and_state_updates` (added `use_negative_stream_ordering`)
       - `_persist_events_txn`
          - `_update_room_depths_txn` (added `update_room_forward_stream_ordering`)
             - `_store_room_members_txn` (added `inhibit_local_membership_updates`)

### Pull Request Checklist

### DIFF
--- a/synapse/handlers/federation_event.py
+++ b/synapse/handlers/federation_event.py
@@ -1809,7 +1809,24 @@ class FederationEventHandler:
 
         try:
             await self.persist_events_and_notify(
-                event.room_id, [(event, context)], backfilled=backfilled
+                event.room_id,
+                [(event, context)],
+                # We should not send notifications about backfilled events.
+                inhibit_push_notifications=backfilled,
+                # We don't need to calculate the state for backfilled events and
+                # we there is no need to update the forward extrems because we
+                # already know this event happened in the past if it was
+                # backfilled.
+                should_calculate_state_and_forward_extrems=not backfilled,
+                # Backfilled events get a negative stream ordering so they don't
+                # come down incremental `/sync`
+                use_negative_stream_ordering=backfilled,
+                # Backfilled events do not affect the current local state
+                inhibit_local_membership_updates=backfilled,
+                # Backfilled events have negative stream ordering and happened
+                # in the past so we know that we don't need to update the
+                # stream_ordering tip for the room.
+                update_room_forward_stream_ordering=not backfilled,
             )
         except Exception:
             run_in_background(
@@ -1823,6 +1840,10 @@ class FederationEventHandler:
         event_and_contexts: Sequence[Tuple[EventBase, EventContext]],
         *,
         inhibit_push_notifications: bool = False,
+        should_calculate_state_and_forward_extrems: bool = True,
+        use_negative_stream_ordering: bool = False,
+        inhibit_local_membership_updates: bool = False,
+        update_room_forward_stream_ordering: bool = True,
     ) -> int:
         """Persists events and tells the notifier/pushers about them, if
         necessary.
@@ -1835,6 +1856,19 @@ class FederationEventHandler:
             inhibit_push_notifications: Whether to stop the notifiers/pushers
                 from knowing about the event. Usually this is done for any backfilled
                 event.
+            should_calculate_state_and_forward_extrems: Determines whether we
+                need to calculate the state and new forward extremities for the
+                room. This should be set to false for backfilled events.
+            use_negative_stream_ordering: Whether to start stream_ordering on
+                the negative side and decrement. Usually this is done for any
+                backfilled event.
+            inhibit_local_membership_updates: Stop the local_current_membership
+                from being updated by these events. Usually this is done for
+                backfilled events.
+            update_room_forward_stream_ordering: Whether to update the
+                stream_ordering position to mark the latest event as the front
+                of the room. This should only be set as false for backfilled
+                events.
 
         Returns:
             The stream ID after which all events have been persisted.
@@ -1861,7 +1895,11 @@ class FederationEventHandler:
             # Note that this returns the events that were persisted, which may not be
             # the same as were passed in if some were deduplicated due to transaction IDs.
             events, max_stream_token = await self._storage.persistence.persist_events(
-                event_and_contexts, backfilled=backfilled
+                event_and_contexts,
+                should_calculate_state_and_forward_extrems=should_calculate_state_and_forward_extrems,
+                use_negative_stream_ordering=use_negative_stream_ordering,
+                inhibit_local_membership_updates=inhibit_local_membership_updates,
+                update_room_forward_stream_ordering=update_room_forward_stream_ordering,
             )
 
             if self._ephemeral_messages_enabled:
--- a/synapse/handlers/federation_event.py
+++ b/synapse/handlers/federation_event.py
@@ -1814,7 +1814,7 @@ class FederationEventHandler:
                 # We should not send notifications about backfilled events.
                 inhibit_push_notifications=backfilled,
                 # We don't need to calculate the state for backfilled events and
-                # we there is no need to update the forward extrems because we
+                # there is no need to update the forward extrems because we
                 # already know this event happened in the past if it was
                 # backfilled.
                 should_calculate_state_and_forward_extrems=not backfilled,
@@ -1823,7 +1823,7 @@ class FederationEventHandler:
                 use_negative_stream_ordering=backfilled,
                 # Backfilled events do not affect the current local state
                 inhibit_local_membership_updates=backfilled,
-                # Backfilled events have negative stream ordering and happened
+                # Backfilled events have negative stream_ordering and happened
                 # in the past so we know that we don't need to update the
                 # stream_ordering tip for the room.
                 update_room_forward_stream_ordering=not backfilled,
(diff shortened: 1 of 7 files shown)


synapse/replication/http/federation.py
```diff
@@ -69,14 +69,37 @@ def __init__(self, hs: "HomeServer"):
...
-    async def _serialize_payload(store, room_id, event_and_contexts, backfilled):
+    async def _serialize_payload(
+        store,
+        room_id,
+        event_and_contexts,
+        inhibit_push_notifications,
+        should_calculate_state_and_forward_extrems,
+        use_negative_stream_ordering,
+        inhibit_local_membership_updates,
+        update_room_forward_stream_ordering,
```
MadLittleMods, line 80: Is there a way we can define the arguments here like the following? I can't get it to work even without the keyword-only marker.

```py
async def _serialize_payload(
    store,
    room_id,
    event_and_contexts,
    *,
    inhibit_push_notifications: bool = False,
    should_calculate_state_and_forward_extrems: bool = True,
    use_negative_stream_ordering: bool = False,
    inhibit_local_membership_updates: bool = False,
    update_room_forward_stream_ordering: bool = True,
):
```
    
When I do this, I get the following error:

```
synapse/replication/http/federation.py:72: error: Signature of "_serialize_payload" incompatible with supertype "ReplicationEndpoint"  [override]
```

And I can't figure out how to adjust the abstract method to be more flexible, https://github.com/matrix-org/synapse/blob/7ffddd819c2e0e7edff141e987372205894084b6/synapse/replication/http/_base.py#L131-L132

synapse/storage/persist_events.py
```diff
@@ -35,6 +35,7 @@
 )
 
 import attr
+from mypy_extensions import DefaultNamedArg
```
MadLittleMods, line 38: Is there a better way to tackle this?

The build is failing because I used it `builtins.ModuleNotFoundError: No module named 'mypy_extensions'`, https://github.com/matrix-org/synapse/runs/4261552290?check_suite_focus=true#step:6:739

synapse/handlers/federation_event.py
```diff
@@ -1809,7 +1809,24 @@ async def _run_push_actions_and_persist_event(
...
+                should_calculate_state_and_forward_extrems=not backfilled,
+                # Backfilled events get a negative stream ordering so they don't
+                # come down incremental `/sync`
+                use_negative_stream_ordering=backfilled,
+                # Backfilled events do not affect the current local state
+                inhibit_local_membership_updates=backfilled,
+                # Backfilled events have negative stream_ordering and happened
+                # in the past so we know that we don't need to update the
+                # stream_ordering tip for the room.
+                update_room_forward_stream_ordering=not backfilled,
```
MadLittleMods, line 1829: In a future PR, I'll extend the behavior arguments further upstream in place of `backfilled`.

This PR does not replace all `backfilled` instances. Just `persist_events_and_notify` and all downstream calls

synapse/handlers/federation_event.py
```diff
@@ -1859,15 +1904,19 @@ async def persist_events_and_notify(
...
+                update_room_forward_stream_ordering=update_room_forward_stream_ordering,
             )
 
             if self._ephemeral_messages_enabled:
                 for event in events:
                     # If there's an expiry timestamp on the event, schedule its expiry.
                     self._message_handler.maybe_schedule_expiry(event)
 
-            if not backfilled:  # Never notify for backfilled events
+            if not inhibit_push_notifications:  # Never notify for backfilled events
```
MadLittleMods, line 1919: Behavioral argument replacement: `inhibit_push_notifications`

synapse/storage/persist_events.py
```diff
@@ -448,7 +590,7 @@ async def _persist_event_batch(
             # device lists as stale.
             potentially_left_users: Set[str] = set()
 
-            if not backfilled:
+            if should_calculate_state_and_forward_extrems:
```
MadLittleMods, line 593: Behavioral argument replacement: `should_calculate_state_and_forward_extrems`

synapse/storage/databases/main/events.py
```diff
@@ -159,7 +175,7 @@ async def _persist_events_and_state_updates(
         #
         # Note: Multiple instances of this function cannot be in flight at
         # the same time for the same room.
-        if backfilled:
+        if use_negative_stream_ordering:
```
MadLittleMods, line 178: Behavior argument replacement: `use_negative_stream_ordering`

synapse/storage/databases/main/events.py
```diff
@@ -1682,7 +1740,7 @@ def non_null_str_or_none(val: Any) -> Optional[str]:
             # band membership", like a remote invite or a rejection of a remote invite.
             if (
                 self.is_mine_id(event.state_key)
-                and not backfilled
+                and not inhibit_local_membership_updates
```
MadLittleMods, line 1743: Behavioral argument replacement: `inhibit_local_membership_updates`

synapse/storage/databases/main/events.py
```diff
@@ -176,13 +192,15 @@ async def _persist_events_and_state_updates(
...
+                inhibit_local_membership_updates=inhibit_local_membership_updates,
+                update_room_forward_stream_ordering=update_room_forward_stream_ordering,
                 state_delta_for_room=state_delta_for_room,
                 new_forward_extremeties=new_forward_extremeties,
             )
             persist_event_counter.inc(len(events_and_contexts))
 
-            if not backfilled:
+            # TODO: test that this actuall works
+            if stream < 0:
```
MadLittleMods, line 203: TODO: Test that this actually works

synapse/storage/databases/main/events.py
```diff
@@ -1200,21 +1232,27 @@ def _update_room_depths_txn(
...
+                events because backfilled events have negative stream_ordering
+                and happened in the past so we know that we don't need to
+                update the stream_ordering tip for the room.
         """
         depth_updates: Dict[str, int] = {}
         for event, context in events_and_contexts:
             # Remove the any existing cache entries for the event_ids
             txn.call_after(self.store._invalidate_get_event_cache, event.event_id)
-            if not backfilled:
+            if update_room_forward_stream_ordering:
```
MadLittleMods, line 1255: Behavioral argument replacment: `update_room_forward_stream_ordering`

synapse/handlers/federation_event.py
```diff
@@ -1821,7 +1838,12 @@ async def persist_events_and_notify(
         self,
         room_id: str,
         event_and_contexts: Sequence[Tuple[EventBase, EventContext]],
-        backfilled: bool = False,
+        *,
+        inhibit_push_notifications: bool = False,
+        should_calculate_state_and_forward_extrems: bool = True,
+        use_negative_stream_ordering: bool = False,
+        inhibit_local_membership_updates: bool = False,
+        update_room_forward_stream_ordering: bool = True,
```
MadLittleMods, line 1846: @richvdh How does this look? Is this on the right track?

synapse/replication/http/federation.py
richvdh: I'm not entirely sure why, but it's the `:bool` type annotations that are upsetting it rather than the `*` or the default values. If I remove the `:bool`s, it's fine.

synapse/handlers/federation_event.py
richvdh, line 1829: shouldn't `update_room_forward_stream_ordering` be implied by `use_negative_stream_ordering` ?

synapse/handlers/federation_event.py
```diff
@@ -1809,7 +1809,24 @@ async def _run_push_actions_and_persist_event(
...
+                # We don't need to calculate the state for backfilled events and
+                # there is no need to update the forward extrems because we
+                # already know this event happened in the past if it was
+                # backfilled.
+                should_calculate_state_and_forward_extrems=not backfilled,
+                # Backfilled events get a negative stream ordering so they don't
+                # come down incremental `/sync`
+                use_negative_stream_ordering=backfilled,
+                # Backfilled events do not affect the current local state
+                inhibit_local_membership_updates=backfilled,
```
richvdh, line 1825: would it be sensible to have this implied by `should_calculate_state_and_forward_extrems` ?

synapse/handlers/federation_event.py
```diff
@@ -1809,7 +1809,24 @@ async def _run_push_actions_and_persist_event(
...
-                event.room_id, [(event, context)], backfilled=backfilled
+                event.room_id,
+                [(event, context)],
+                # We should not send notifications about backfilled events.
+                inhibit_push_notifications=backfilled,
+                # We don't need to calculate the state for backfilled events and
+                # there is no need to update the forward extrems because we
+                # already know this event happened in the past if it was
+                # backfilled.
+                should_calculate_state_and_forward_extrems=not backfilled,
```
richvdh, line 1820: the `should_` prefix seems a bit redundant here.

synapse/handlers/federation_event.py
```diff
@@ -1850,7 +1895,7 @@ async def persist_events_and_notify(
                     store=self._store,
                     room_id=room_id,
                     event_and_contexts=batch,
-                    backfilled=backfilled,
+                    inhibit_push_notifications=inhibit_push_notifications,
```
richvdh, line 1898: doesn't this need to pass through *all* the keyword args?

synapse/storage/databases/main/events.py
richvdh: > Behavior argument replacement: `use_negative_stream_ordering`

sorry, what does this mean?

richvdh (changes requested):
Thanks for putting this up, @MadLittleMods!

I haven't read all the way through the diff, because it's a bit hard to follow. Could I ask you to break it down a bit - let's start somewhere nearer the bottom of the stack: say `_persist_events_and_state_updates`. I think that will make it much easier to convince ourselves that the changes are correct.

Generally it seems like the right *sort* of thing, but I fear we might be taking it to extremes.  If there are ways of grouping some of these flags together into coherent bits of functionality, I think that would be better. For example - does it ever make sense to set `use_negative_stream_ordering=True` and `update_room_forward_stream_ordering=True` together (or both `False`)? If not, let's get rid of `update_room_forward_stream_ordering` - otherwise it's just a nightmare to use the function.

synapse/storage/databases/main/events.py
MadLittleMods: It's just the root place that I originally replaced and defines the behavior

synapse/handlers/federation_event.py
MadLittleMods: I understand the leap that updating the state in the room also encompasses member state events.

My first impression was that it wasn't very obvious but it makes sense ⏩

synapse/handlers/federation_event.py
MadLittleMods: 👍 I think so. Should match `_serialize_payload`.

It's a loose connection and I was having trouble typing it correctly, https://github.com/matrix-org/synapse/pull/11396#discussion_r752929555, to have the linter yell at me for this.

MadLittleMods:
Thanks for the review @richvdh, great points! I've made a smaller PoC in https://github.com/matrix-org/synapse/pull/11417

PR #11398, opened 2021-11-19
Implement MSC3383: include destination in X-Matrix auth header

jcgruenhage:
This is the implementation for MSC3383, as requested over at https://github.com/matrix-org/matrix-doc/pull/3383.
We've tested this at Famedly already, deployed it to a testserver, federation still works as expected, we've tested it with complement and sytest, and we're planning to roll it out with the 1.47.1 release of our synapse fork next week.

### Pull Request Checklist

* [x] Pull request includes a [sign off](https://matrix-org.github.io/synapse/latest/development/contributing_guide.html#sign-off) (see https://github.com/matrix-org/synapse/pull/11398/commits/9b7433055de93b89046c921a01d001c701095848)


synapse/federation/transport/server/_base.py
```diff
@@ -82,10 +82,15 @@ async def authenticate_request(self, request, content):
...
             if auth.startswith(b"X-Matrix"):
-                (origin, key, sig) = _parse_auth_header(auth)
+                (origin, key, sig, destination) = _parse_auth_header(auth)
                 json_request["origin"] = origin
                 json_request["signatures"].setdefault(origin, {})[key] = sig
 
+                # if the origin_server sent a destination along it needs to match our own server_name
+                if destination is not None and destination != self.server_name:
+                    raise AuthenticationError(
+                        400, "Destination mismatch in auth header", Codes.UNAUTHORIZED
```
DMRobertson, line 92: Could we use `HTTPStatus.BAD_REQUEST` here instead of 400? Or should this be 401, `HTTPStatus.UNAUTHORIZED`?

(Is unauthorized the right outcome here? As I understand it, we could confirm that this is from the originating server---it's just that this request isn't meant for us?)

synapse/federation/transport/server/_base.py
```diff
@@ -140,7 +145,7 @@ def _parse_auth_header(header_bytes):
         header_bytes (bytes): header value
 
     Returns:
-        Tuple[str, str, str]: origin, key id, signature.
+        Tuple[str, str, str, Optional[str]]: origin, key id, signature, destination.
```
DMRobertson, line 148: While we're here, it'd be great if these types could be added as annotations to the function itself (and removed from the docstring)

synapse/http/matrixfederationclient.py
```diff
@@ -715,6 +715,9 @@ def build_auth_headers(
         Returns:
             A list of headers to be added as "Authorization:" headers
         """
+        if destination is None and destination_is is None:
+            raise ValueError("destination and destination_is cannot both be None!")
```
DMRobertson, line 719: Did this check make sense before this change too?

DMRobertson (commented):
A few thoughts! Probably wants more input from @matrix-org/synapse-core.

synapse/federation/transport/server/_base.py
jcgruenhage: Yep, we received a request that isn't meant for us. This means that the authorization was completed correctly, so 400 feels better than 401.

I'll use `HTTPStatus.BAD_REQUEST` though ^^

synapse/http/matrixfederationclient.py
jcgruenhage: Yep

synapse/federation/transport/server/_base.py
jcgruenhage: I'd say that this might be out of scope of this PR, but if you want I can include that. wdyt?

synapse/http/matrixfederationclient.py
jcgruenhage: I've added that because https://github.com/matrix-org/synapse/blob/5f0c507b8922463d1ae98e9f92a56c2c01c32c4e/synapse/http/matrixfederationclient.py#L748 would result in a key error otherwise, which is a relatively weird error message.

changelog.d/11398.feature
```diff
@@ -0,0 +1 @@
+Implement MSC3383 for including the destination in S2S authentication headers.
```
clokep, line 1:
```suggestion
Implement [MSC3383](https://github.com/matrix-org/matrix-doc/pull/3383) for including the destination in server-to-server authentication headers.
```

You can include "Contributed by your name." (or Famedly) if you would like, by the way.

clokep (commented):
I think this looks great overall, but we would need to see the MSC move forward before merging.

We might also be able to include it if it were behind an experimental configuration flag, but I'm unsure in this case.

ShadowJonathan:
> but we would need to see the MSC move forward before merging.

That'd be a bit of a catch-22, as a MSC needs a working implementation before it can be continued towards SCT review, I only see approvals, so unless you'd like to review it as well (and see it move forward yourself), then i think this has to go forward first.

Unless you'd like to see this implemented in an alternative server implementation as well...?

clokep:
> as a MSC needs a working implementation before it usually is seriously considered for SCT review

A pull request is considered a working implementation.

clokep:
@jcgruenhage Looks like this got caught in the linter though, Could you re-run `./scripts-dev/lint.sh`?

jcgruenhage:
Yeah, sorry, I'll run the linting stuff later, didn't have time to clean that up yesterday.

synapse/federation/transport/server/_base.py
DMRobertson: I don't mind leaving it be for now; at some point we'll have to do a pass over this file to improve type coverage anyway.

changelog.d/11398.feature
jcgruenhage: Is it good like this?

jcgruenhage:
Linting stuff is all fixed by now, and I also changed the parsing of the header a bit, because the linter complained on that as well, even though that line hadn't really been touched in 3 years afaict

changelog.d/11398.feature
clokep: Looks great! Thanks!

clokep:
The MSC has passed FCP, so putting this back into the review queue.

@jcgruenhage might be good to merge `develop` into this since it has been a while since CI ran?

clokep:
> Merging develop as opposed to rebasing this? Not that much of a fan of merging long diverged branches, but can do if that's your process.

We prefer merges -- there's some info in the docs about why: https://matrix-org.github.io/synapse/latest/development/git.html

Note that we squash merge so it will end up being a single commit once merged to `develop`.

jcgruenhage:
Okay, sounds good.

synapse/federation/transport/server/_base.py
```diff
@@ -77,15 +79,24 @@ async def authenticate_request(self, request, content):
...
             if auth.startswith(b"X-Matrix"):
-                (origin, key, sig) = _parse_auth_header(auth)
+                (origin, key, sig, destination) = _parse_auth_header(auth)
                 json_request["origin"] = origin
                 json_request["signatures"].setdefault(origin, {})[key] = sig
 
+                # if the origin_server sent a destination along it needs to match our own server_name
+                if destination is not None and destination != self.server_name:
+                    raise AuthenticationError(
+                        HTTPStatus.BAD_REQUEST,
```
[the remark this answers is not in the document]
jcgruenhage: Applied. For context, this code was written before that section of the MSC was added.

DMRobertson (changes requested):
I think this needs the linter script to get it through CI; otherwise, I'm happy.

DMRobertson (approved):
Many thanks!

PR #11400, opened 2021-11-19
Trim DataStore inheritance

DMRobertson:
This might help down the line.

See individual commits for justification.

### DIFF
--- a/changelog.d/11400.misc
+++ b/changelog.d/11400.misc
@@ -1,0 +1,1 @@
+Trim redundant DataStore inheritance.
--- a/synapse/app/admin_cmd.py
+++ b/synapse/app/admin_cmd.py
@@ -28,13 +28,11 @@ from synapse.config.homeserver import HomeServerConfig
 from synapse.config.logger import setup_logging
 from synapse.events import EventBase
 from synapse.handlers.admin import ExfiltrationWriter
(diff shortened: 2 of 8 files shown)


erikjohnston (commented):
I'm hesitant to do this as we're now no longer declaring which classes we are actually expecting to be in the `DataStore` and `GenericWorkerSlavedStore`, which I think is as a bit confusing for mixin classes. I guess the benefit of doing this is that it makes it less likely devs run into MRO errors when shuffling things around, but that should be fairly rare?

DMRobertson:
> I'm hesitant to do this as we're now no longer declaring which classes we are actually expecting to be in the `DataStore` and `GenericWorkerSlavedStore`, which I think is as a bit confusing for mixin classes. I guess the benefit of doing this is that it makes it less likely devs run into MRO errors when shuffling things around, but that should be fairly rare?

I concede that my motivation is to reduce MRO pain. More generally, the present state of affairs is already confusing and this was an attempt to lessen that.

I'm not convinced the storage classes really are mixins, given the implicit dependencies between so many of them.

I'd be sad to drop this ([-0.2 Apaches, say](https://www.apache.org/foundation/voting.html#expressing-votes-1-0-1-and-fractions)) but won't stop us from doing so.

DMRobertson:
This has rotted. #11733 tries to characterise the general problem.

PR #11409, opened 2021-11-22
Improved push typing

Bubu:
I had started this quite a bit ago but never got around to submitting but I thought it might still be useful to have this.

I'm unhappy with a few bits, so maybe someone has a better idea on how to model this, I'll add specific comments inline.

### Pull Request Checklist

### DIFF
--- a/synapse/push/httppusher.py
+++ b/synapse/push/httppusher.py
@@ -27,7 +27,7 @@ from synapse.logging import opentracing
 from synapse.metrics.background_process_metrics import run_as_background_process
 from synapse.push import Pusher, PusherConfig, PusherConfigException
 
-from ..storage.databases.main.event_push_actions import HttpPushAction
+from synapse.storage.databases.main.event_push_actions import HttpPushAction
 from . import push_rule_evaluator, push_tools
 
 if TYPE_CHECKING:
--- a/synapse/push/httppusher.py
+++ b/synapse/push/httppusher.py
@@ -28,6 +28,7 @@ from synapse.metrics.background_process_metrics import run_as_background_process
 from synapse.push import Pusher, PusherConfig, PusherConfigException
 
 from . import push_rule_evaluator, push_tools
+from ..storage.databases.main.event_push_actions import HttpPushAction
 
 if TYPE_CHECKING:
     from synapse.server import HomeServer
@@ -273,7 +274,7 @@ class HttpPusher(Pusher):
                     )
                     break
 
-    async def _process_one(self, push_action: dict) -> bool:
+    async def _process_one(self, push_action: HttpPushAction) -> bool:
         if "notify" not in push_action["actions"]:
             return True
 
--- a/synapse/push/httppusher.py
+++ b/synapse/push/httppusher.py
@@ -26,8 +26,8 @@ from synapse.events import EventBase
 from synapse.logging import opentracing
 from synapse.metrics.background_process_metrics import run_as_background_process
 from synapse.push import Pusher, PusherConfig, PusherConfigException
-
 from synapse.storage.databases.main.event_push_actions import HttpPushAction
+
 from . import push_rule_evaluator, push_tools
 
 if TYPE_CHECKING:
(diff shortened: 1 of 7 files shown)


synapse/push/push_types.py
```diff
@@ -0,0 +1,67 @@
...
+
+class MessageVars(TypedDict, total=False):
+    event_type: str
+    is_historical: bool
+    id: str
+    ts: int
+    sender_name: str
+    sender_avatar_url: Optional[str]
+    sender_hash: int
+    msgtype: Optional[Any]
```
Bubu, line 52: `msgtype` and `format` needing to be `Optional[Any]` doesn't feel particularly great. They come from json fields, so maybe how to type-hint those is already a solved problem in other parts of the codebase?

synapse/push/push_types.py
```diff
@@ -0,0 +1,67 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from typing import Any, List, Optional
+
+from typing_extensions import TypedDict
+
+
+class EmailReason(TypedDict, total=False):
```
Bubu, line 19: The `total=False` isn't great in my opinion, but it's currently necessary as many code path iteratively construct these dicts. Maybe reworking that could be done as a next step?

synapse/push/push_types.py
erikjohnston: I think the usual way of doing this is to do an `isinstance(msgtype, str)` check when pulling it out of the JSON dict (and discarding the value if its not a str)

synapse/push/push_types.py
erikjohnston: Can we keep them as normal dicts while we're building them, and then cast? Or just use local variables and then build it at the end? e.g.:

```python
let room_id = ...
...
let room_name = ...
....
reason: EmailReason  = { ... }
```

Not sure if that works

synapse/push/push_types.py
```diff
@@ -0,0 +1,67 @@
...
+    room_id: str
+    room_name: Optional[str]
+    now: int
+    received_at: int
+    delay_before_mail_ms: int
+    last_sent_ts: int
+    throttle_ms: int
+
+
+class NotifVars(TypedDict):
```
erikjohnston, line 29: While you are here could you had some doc comments for each class please?

erikjohnston (commented):
Thanks! I've left a few comments, but its nice to see some more typing being added. 

Ideally we'd replace the `TypedDict` instances with real `attr.s` classes, but I can imagine that is a bigger change.

synapse/push/httppusher.py
```diff
@@ -27,6 +27,7 @@
 from synapse.metrics.background_process_metrics import run_as_background_process
 from synapse.push import Pusher, PusherConfig, PusherConfigException
 
+from ..storage.databases.main.event_push_actions import HttpPushAction
```
squahtx, line 30:
```suggestion
from synapse.storage.databases.main.event_push_actions import HttpPushAction
```
absolute imports are preferred in synapse

synapse/push/push_types.py
Bubu: Done

synapse/push/push_types.py
ShadowJonathan: That would require changing how these functions iteratively build the struct, while i think that's a good idea in general, i don't know how possible that is, certainly as i remember some of them building these up across multiple functions, passing intermediate results.

Bubu: Yup, I also believe that's the case. I'll take another look and see if I can resolve at least some easy cases here.

Bubu: I looked again and with the current code structure doing anything like this doesn't really make sense as for the reasons @ShadowJonathan mentioned.

The thing we could do is creating None value fields instead of structures with missing fields (so adding a bunch of Optionals instead of `total=False`). I currently have no opinion if this is better or worse.

Anything else would require more restructuring of the code, which I think is better done seperately?

reivilibre: Thanks for looking into it. I think restructuring this may be the right answer eventually but you're right that it's not for here and now.

reivilibre (approved):
Thank you!

PR #11410, opened 2021-11-22
fix push notifications for invites from remote servers

### LINKED ISSUE #8626: We don't send push notifications for invites from remote servers (for rooms the local server isn't in).
The invite comes in via federation `/invite` and gets persisted as an outlier. This flow bypasses the `handle_push_actions_for_event` code.

```python
    def test_remote_invite_sends_email(self):
        fed_handler = self.hs.get_federation_handler()

        fake_invite = {
            "auth_events": ["$dfdjslfdkf", "$sdfdjflsjoiy"],
            "prev_events": ["$sdfdjflsjoiy"],
            "type": "m.room.member",
            "state_key": self.user_id,
            "content": {"membership": "invite"},
            "sender": "@test:remote.example.com",
            "room_id": "!soom_room:remote.example.com",
            "depth": 3872,
        }

        testkey = signedjson.key.generate_signing_key("ver1")

        invite_event = create_local_event_from_event_dict(
            self.clock,
            hostname="remote.example.com",
            signing_key=testkey,
            room_version=RoomVersions.V6,
            event_dict=fake_invite,
        )

        self.get_success(
            fed_handler.on_invite_request(
                "remote.example.com", invite_event, RoomVersions.V6,
            )
        )

        # We should get emailed about the invite
        self._check_for_mail()
```

Is a unit test failing, if added to `tests/push/test_email.py`

Bubu:
So while this solution works and actually solves the referenced bug, I'm not sure if this is doing things correctly. In fact I'm almost certain it's not doing things correctly, but I don't have enough insight into why the push code is structured as is to say why exactly circumventing that structure is bad. This bypasses any database persistence for this specific case of pushes and just directly sends those out.

I'm happy to restructure this with advice from someone with more background knowledge here.

Based on #11409
Closes #8626

### Pull Request Checklist

### DIFF
--- a/changelog.d/11410.bugfix
+++ b/changelog.d/11410.bugfix
@@ -1,0 +1,1 @@
+Fix not sending out push notifications for invites from remote servers.
--- a/synapse/handlers/federation.py
+++ b/synapse/handlers/federation.py
@@ -818,6 +818,8 @@ class FederationHandler:
             event.room_id, [(event, context)]
         )
 
+        await self._federation_event_handler.notify_remote_invite(event, context)
+
         return event
 
     async def do_remotely_reject_invite(
--- a/synapse/handlers/federation_event.py
+++ b/synapse/handlers/federation_event.py
@@ -58,11 +58,14 @@ from synapse.federation.federation_client import InvalidResponseError
 from synapse.logging.context import nested_logging_context, run_in_background
 from synapse.logging.utils import log_function
 from synapse.metrics.background_process_metrics import run_as_background_process
+from synapse.push.bulk_push_rule_evaluator import BulkPushRuleEvaluator
+from synapse.push.httppusher import HttpPusher
 from synapse.replication.http.devices import ReplicationUserDevicesResyncRestServlet
 from synapse.replication.http.federation import (
     ReplicationFederationSendEventsRestServlet,
 )
 from synapse.state import StateResolutionStore
+from synapse.storage.databases.main.event_push_actions import BasePushAction
 from synapse.storage.databases.main.events_worker import EventRedactBehaviour
 from synapse.types import (
     PersistedEventPosition,
@@ -96,6 +99,7 @@ class FederationEventHandler:
     """
 
     def __init__(self, hs: "HomeServer"):
+        self._hs = hs
         self._store = hs.get_datastore()
         self._storage = hs.get_storage()
         self._state_store = self._storage.state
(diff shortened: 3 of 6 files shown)


reivilibre (changes requested):
Would you be happy to merge in (or rebase, at your preference, since nothing has been reviewed yet) develop? I gave it a quick try but there were cases where the better push typing PR seems to have diverged, so I don't trust myself to do it properly.

Bubu:
@reivilibre Rebased on develop! Thanks for looking at it :).

reivilibre (commented):
Thanks for doing the rebase!

I've been staring at this for a while — I suspect it's not 'right' architecturally but I really wanted to say something helpful.

The main red flag I see with it is that push notifications are sent by a specific worker ([the `pusher` worker type](https://matrix-org.github.io/synapse/latest/workers.html#synapseapppusher)) but in this PR, a worker handling federation will be sending the push.

I'll have to be honest with you that I'm not very familiar with the structure of the push code here and what I do know is from a few years ago..

My suspicion is that maybe we should be putting actions into the staging area, *then* persisting the event, with a *catch* clause that will back out the staging area if the persist fails. (That said, I just described the same thing that `_run_push_actions_and_persist_event` does.)

Maybe the answer is that `_run_push_actions_and_persist_event` needs to be adapted to be happy with outliers (at least for invite events) and then the invite code should use that instead of plain `persist_events_and_notify`?

I just worry that I don't know *why* outliers aren't allowed in `_run_push_actions_and_persist_event` and that's probably pretty critical for doing this. It might be that we need to carry some special cases down for these outlier invites.

I'll put this back on the queue in case anyone else has any more intelligent thoughts.

squahtx:
I tried to review and ended up coming to roughly the same conclusion as @reivilibre above. The change doesn't sit right architecturally, but I'm not familiar enough with this part of the code to suggest what should be done instead.

Additionally, won't this send notifications twice for invites into rooms that the homeserver is already participating in?
#8626 seems to suggest that notifications are already sent for such rooms.

richvdh (changes requested):
As others have said, this crosses various architecture boundaries.

What happens if we simply relax the `assert not event.internal_metadata.outlier` in `_run_push_actions_and_persist_event` to allow events with `event.internal_metadata.out_of_band_membership`, and call that instead of `persist_events_and_notify`?

anoadragon453:
This PR has been sitting around for a while. I'm going to close it for now to tidy the review queue a bit, but please re-open it (and rebase on latest `develop`) if you'd like to keep working on it!

PR #11413, opened 2021-11-23
Return the stable `event` field from `/send_join` per MSC3083.

### LINKED ISSUE #11345: /send_join response includes unstable prefix for event field
The response generated for `/send_join` (and parsing of it on the federation client side) still uses an unstable prefix from [MSC3083](https://github.com/matrix-org/matrix-doc/pull/3083): `org.matrix.msc3083.v2.event`.

This should be updated to be just `event`, the stable field name. We probably need to send / handle both for a period of time unfortunately.

See https://github.com/matrix-org/synapse/blob/98c8fc6ce82d9d6b1bd21bf70df6a0e1ce91c1dc/synapse/federation/federation_server.py#L617

clokep:
[MSC3083](https://github.com/matrix-org/matrix-doc/pull/3083) added an additional field to the `/send_join` response: it now returns the final serialized event. Synapse was still returning the unstable field name (`org.matrix.msc3083.v2.event`). This updates Synapse to return both the stable and unstable field names and search for both when parsing responses.

In a few versions we can stop returning the unstable field. I have another branch which does this and passes all tests.

This was an oversight in #10449 (I think, or perhaps just when MSC3083 was merged).

Fixes #11345


synapse/federation/transport/client.py
```diff
@@ -1317,15 +1317,22 @@ def __init__(self, room_version: RoomVersion, v1_api: bool):
...
+        self._coro_unstable_event = ijson.kvitems_coro(
             _event_parser(self._response.event_dict),
             prefix + "org.matrix.msc3083.v2.event",
             use_float=True,
         )
+        self._coro_event = ijson.kvitems_coro(
+            _event_parser(self._response.event_dict),
+            prefix + "event",
+            use_float=True,
+        )
```
clokep, line 1330: This means that the contents of both the `org.matrix.msc3083.v2.event` and `event` fields will stomp over each other. In a well behaving server they should be identical so this should be "OK". It might be possible to do something weird with this though.

It would be possible to keep these as two separate dictionaries and then do one of the following:

* Use `event` if it is non-empty, otherwise fallback to `org.matrix.msc3083.v2.event`
* Compare the two and ensure they are identical.

I'm a little concerned that this could be used to split-brain a room between different Synapse versions, maybe? But I don't really see how so I think I'm just being paranoid.

babolivier (approved):

synapse/federation/transport/client.py
babolivier: > This means that the contents of both the `org.matrix.msc3083.v2.event` and `event` fields will stomp over each other.

Could you turn this into a code comment please? It was quite difficult for me to understand what's going on before I read this comment.

PR #11414, opened 2021-11-23
Support OIDC backchannel logouts

### LINKED ISSUE #11326: Support back-channel logouts from OIDC providers
**Description:**

Right now, if a user logs out from his identity provider, they does not get logged out from their session in Synapse.
OIDC has two mechanism to achieve that: 

 - front-channel logouts, as defined [here](https://openid.net/specs/openid-connect-frontchannel-1_0.html), which works by having the IdP display to the user an invisible iframe of the relying party ; hence the name "front-channel": the communication between the IdP and Synapse would happen in the user's browser
 - back-channel logouts, as defined [here](https://openid.net/specs/openid-connect-backchannel-1_0.html), which works by sending a request from the IdP to the relying party ; hence the name "back-channel": the communication between the IdP and Synapse would happen from server to server

Since we can do server-to-server requests between the IdP and Synapse, we should probably implement back-channel logouts.

When a logout happens, Synapse would get a JWT sent to an endpoint (e.g. `/_synapse/client/oidc/backchannel-logout`) signed by the IdP (we already have the keys discovered because we verify ID tokens signatures) with either the user ID (`sub`) or the session ID (`sid`).
Ideally, we would logout using the session ID, else we would not be able to distinguish what session to log out. It also involves saving that session ID when the login happens.

Things to do to make that happen:
 - the IdP indicates back-channel logout support via the `backchannel_logout_supported` metadata
 - Synapse indicates support for back-channel logout by filling the appropriate fields in the IdP (Keycloak has a *Backchannel Logout URL* field)
 - when logging in, we need to get the session ID from the ID token (should be the `sid` claim). We should probably attach the session ID to the device to avoid having to carry that whenever we refresh the token
 - when a request goes to `/_synapse/client/oidc/backchannel-logout`, Synapse should:
   - decode the JWT
   - get the issuer (`iss` claim) to find out what provider sent the request
   - validate the JWT signature with the keys we have for that provider
   - check a few claims in the JWT:
     - `aud` (audience) must match the `client_id` used by Synapse
     - `iat` (issued at) must not be in the future, nor too far in the past
     - `events` must be `{ "http://schemas.openid.net/event/backchannel-logout": {} }`
     - `jti` (JWT token ID) should be saved to prevent token replay. We might skip that for performance reasons
   - get the session ID from the `sid` claim
   - find the device associated to this session ID in DB, and logout that device

See PSE-127

sandhose:
Fixes #11326 

Based on #11482.

Tested (and works) with Keycloak.

### How to test this?

 - Spin up a Keycloak instance, create a realm
 - Configure Synapse to login via Keycloak: https://matrix-org.github.io/synapse/latest/openid.html#keycloak
 - In Keycloak, on the client used by Synapse, add the following backchannel logout URL: `{public_url}/_synapse/client/oidc/backchannel_logout`
 - Ensure `Backchannel Logout Session Required` is enabled on that client
 - Login via OIDC in any Matrix client
 - Log the user out of Keycloak by going to `{keycloak}/auth/realms/{realms}/account/#/` (which is the self-service console) and click "Sign Out"
 - The user should be logged out of Synapse

There is a Docker Compose stack here to help with testing: https://github.com/sandhose/synapse-keycloak-devstack

Additionally, there are two edge cases which can be tested (and which are covered by tests):

 - when the logout happens before the client sends the login token (e.g. happens during the "Continuing will grant app.element.io access to your account." screen)
 - when the logout happens during the username mapping session, so when the `localpart_template` is not set and the user chooses a username themselves

### Pull Request Checklist


clokep:
Removing this from the review queue until #11482 gets worked out.

synapse/config/oidc.py
```diff
@@ -562,6 +569,9 @@ class OidcProviderConfig:
     # "openid" scope is used.
     jwks_uri: Optional[str]
 
+    # Whether Synapse should react to backchannel logouts
+    backchannel_logout_enabled: bool
```
clokep, line 573: It makes me slightly sad to add another configuration option here, but I don't think there's much we can do here to avoid it?

The metadata seems to include whether this feature is supported or not. Can we auto-enable it in those cases?

synapse/handlers/oidc.py
```diff
@@ -246,6 +253,78 @@ async def handle_oidc_callback(self, request: SynapseRequest) -> None:
 
         await oidc_provider.handle_oidc_callback(request, session_data, code)
 
+    async def handle_backchannel_logout(self, request: SynapseRequest) -> None:
+        """Handle an incoming request to /_synapse/client/oidc/backchannel_logout
+
+        This extracts the logout_token from the request and try to figure out
+        from whom it is coming from. This works by matching the iss claim with
+        the issuer and the aud claim with the client_id.
```
clokep, line 261: I think this is a little more readable?

```suggestion
        This extracts the logout_token from the request and tries to figure out
        which account it is associated with. This works by matching the iss claim
        with the issuer and the aud claim with the client_id.
```

synapse/handlers/oidc.py
```diff
@@ -246,6 +253,78 @@ async def handle_oidc_callback(self, request: SynapseRequest) -> None:
...
+            raise SynapseError(400, "Missing logout_token in request")
+
+        # A JWT looks like this:
+        #    header.payload.signature
+        # where all parts are encoded with urlsafe base64.
+        # The aud and iss claims we care about are in the payload part, which
+        # is a JSON object.
+        try:
+            # This raises if there are too many or not enough segments in the token
+            header, payload, signature = logout_token.rsplit(".", 4)
```
clokep, line 281: Why the `maxsplit=4`? I only see 3 parts listed above, so I think I would expect a `maxsplit=2`?

Although it seems like we really expect the exact number of results to be 3, so I think we can just leave out the `maxsplit`?

synapse/handlers/oidc.py
```diff
@@ -246,6 +253,78 @@ async def handle_oidc_callback(self, request: SynapseRequest) -> None:
...
+        except ValueError:
+            raise SynapseError(400, "Invalid logout_token in request")
+
+        try:
+            payload_bytes = unpaddedbase64.decode_base64(payload)
+            claims = json.loads(payload_bytes)
```
clokep, line 287: Please use `json_decoder` here.

synapse/handlers/oidc.py
```diff
@@ -246,6 +253,78 @@ async def handle_oidc_callback(self, request: SynapseRequest) -> None:
...
+            iss = claims["iss"]
+            aud = claims["aud"]
+            # The aud claim can be either a string or a list of string. Here we
+            # normalize it as a list of strings.
+            if isinstance(aud, str):
+                aud = [aud]
+
+            # Check that we have the right types for the aud and the iss claims
+            assert isinstance(iss, str)
+            assert isinstance(aud, list)
```
clokep, line 302: We should do if-checks and raise here (assertions can be disabled in Python).

synapse/handlers/oidc.py
```diff
@@ -246,6 +253,78 @@ async def handle_oidc_callback(self, request: SynapseRequest) -> None:
...
+            if provider.issuer == issuer and any(
+                a == provider.client_id for a in audience
+            ):
+                oidc_provider = provider
+                break
+
+        if oidc_provider is None:
+            raise SynapseError(400, "Could not find the OP that issued this event")
+
+        # We now figured out what provider should handle the logout request
```
clokep, line 325:
```suggestion
        # Ask the provider to handle the logout request.
```

synapse/handlers/oidc.py
```diff
@@ -246,6 +253,78 @@ async def handle_oidc_callback(self, request: SynapseRequest) -> None:
 
         await oidc_provider.handle_oidc_callback(request, session_data, code)
 
+    async def handle_backchannel_logout(self, request: SynapseRequest) -> None:
+        """Handle an incoming request to /_synapse/client/oidc/backchannel_logout
+
+        This extracts the logout_token from the request and try to figure out
```
clokep, line 259: Maybe it is simple enough, but I wonder if we should have a separate method which parses the logout token -> `iss` & `aud` claims. Maybe overkill though if we don't need to d that anywhere else.

synapse/handlers/oidc.py
```diff
@@ -246,6 +253,78 @@ async def handle_oidc_callback(self, request: SynapseRequest) -> None:
...
+        """
+        logout_token = parse_string(request, "logout_token")
+        if logout_token is None:
+            raise SynapseError(400, "Missing logout_token in request")
+
+        # A JWT looks like this:
+        #    header.payload.signature
+        # where all parts are encoded with urlsafe base64.
+        # The aud and iss claims we care about are in the payload part, which
+        # is a JSON object.
```
clokep, line 278: The libraries we're using don't allow parsing without verifying signatures?

synapse/handlers/oidc.py
```diff
@@ -398,6 +478,22 @@ def _validate_metadata(self, m: OpenIDProviderMetadata) -> None:
...
 
+        if self._config.backchannel_logout_enabled:
+            if not m.get("backchannel_logout_supported", False):
+                logger.warning(
+                    "OIDC Back-Channel Logout is enabled for issuer %r"
+                    "but it does not advertise support for it",
+                    self.issuer,
+                )
+
+            if not m.get("backchannel_logout_session_supported", False):
```
clokep, line 489: I think this needs to be an `elif` since the warning below says that back-channel logout is enabled.

synapse/handlers/oidc.py
```diff
@@ -413,6 +509,14 @@ def _uses_userinfo(self) -> bool:
             or self._user_profile_method == "userinfo_endpoint"
         )
 
+    @property
+    def issuer(self) -> str:
```
clokep, line 513: Docstrings for both these properties please!

synapse/handlers/oidc.py
```diff
@@ -1035,6 +1139,161 @@ def _remote_id_from_userinfo(self, userinfo: UserInfo) -> str:
...
+        # issuer should be the IdP issuer
+        claim_options = {
+            "iss": {"values": [self.issuer]},
+            "aud": {"values": [self.client_id]},
+        }
+
+        logger.debug("Attempting to decode JWT logout_token %r", logout_token)
+
+        # Try to decode the keys in cache first, then retry by forcing the keys
+        # to be reloaded
```
clokep, line 1193: I think we do this pattern elsewhere, should it be a helper to try, then try-with-force?

synapse/handlers/oidc.py
```diff
@@ -1035,6 +1139,161 @@ def _remote_id_from_userinfo(self, userinfo: UserInfo) -> str:
...
+        #       or both.
+        #    5. Verify that the Logout Token contains an events Claim whose
+        #       value is JSON object containing the member name
+        #       http://schemas.openid.net/event/backchannel-logout.
+        #    6. Verify that the Logout Token does not contain a nonce Claim.
+        # This is all verified by the LogoutToken claims class, so at this
+        # point the `sid` claim exists and is a string.
+        sid: str = claims.get("sid")
+
+        # Find in the store if we have a device associated with that SID
```
clokep, line 1226: Maybe this is clearer?

```suggestion
        # Fetch any device(s) in the store associated with the session ID.
```

synapse/handlers/oidc.py
```diff
@@ -1035,6 +1139,161 @@ def _remote_id_from_userinfo(self, userinfo: UserInfo) -> str:
...
+        devices = await self._store.get_devices_by_auth_provider_session_id(
+            auth_provider_id=self.idp_id,
+            auth_provider_session_id=sid,
+        )
+        for device in devices:
+            user_id = device["user_id"]
```
clokep, line 1232: This feels a little fishy -- should we be asserting that the user information matches, or do we not get any of that back and only receive the session ID?

synapse/handlers/oidc.py
```diff
@@ -1035,6 +1139,161 @@ def _remote_id_from_userinfo(self, userinfo: UserInfo) -> str:
...
+            device_id = device["device_id"]
+            logger.info(
+                "Logging out %r (device %r) via OIDC backchannel logout (sid %r).",
+                user_id,
+                device_id,
+                sid,
+            )
+            await self._device_handler.delete_device(user_id, device_id)
```
clokep, line 1240: Probably better to use the `delete_devices` method (it looks like most of the code is the same...feels like something we should clean-up).

synapse/handlers/oidc.py
```diff
@@ -1035,6 +1139,161 @@ def _remote_id_from_userinfo(self, userinfo: UserInfo) -> str:
...
+
+        request.setResponseCode(200)
+        request.setHeader(b"Cache-Control", b"no-cache, no-store")
```
clokep, line 1243: The other place we set this we also set `must-revalidate`: https://github.com/matrix-org/synapse/blob/b64b6d12d47c012a9996a2585a969d3b962d25fa/synapse/http/server.py#L683

We also don't set `Pragma` there. Should this be a helper (like `set_clickjacking_protection_headers`)?

synapse/handlers/oidc.py
```diff
@@ -1035,6 +1139,161 @@ def _remote_id_from_userinfo(self, userinfo: UserInfo) -> str:
...
+    """
+    Holds and verify claims of a logout token, as per
+    https://openid.net/specs/openid-connect-backchannel-1_0.html#LogoutToken
+    """
+
+    REGISTERED_CLAIMS = ["iss", "sub", "aud", "iat", "jti", "events", "sid"]
+
+    def validate(self, now: Optional[int] = None, leeway: int = 0) -> None:
+        """Validate everything in claims payload."""
+        super(LogoutToken, self).validate(now, leeway)
```
clokep, line 1258:
```suggestion
        super().validate(now, leeway)
```

synapse/handlers/oidc.py
```diff
@@ -1035,6 +1139,161 @@ def _remote_id_from_userinfo(self, userinfo: UserInfo) -> str:
...
+        self.validate_sid()
+        self.validate_events()
+        self.validate_nonce()
+
+    def validate_exp(self, now: Optional[int] = None, leeway: int = 0) -> None:
```
clokep, line 1263: Do these get called automatically? (Why do we call them manually above?)

synapse/handlers/oidc.py
```diff
@@ -1035,6 +1139,161 @@ def _remote_id_from_userinfo(self, userinfo: UserInfo) -> str:
...
+        if not isinstance(sid, str):
+            raise InvalidClaimError("sid")
+
+    def validate_nonce(self) -> None:
+        """Ensure the nonce claim is absent"""
+        if "nonce" in self:
+            raise InvalidClaimError("nonce")
+
+    def validate_events(self) -> None:
+        """Ensure the events is present and with the right value"""
```
clokep, line 1286:
```suggestion
        """Ensure the events claim is present and with the right value"""
```

clokep (commented):
Looks pretty sensible, I left a bunch of comments but I think most are minor!

synapse/handlers/oidc.py
sandhose: So, the logout token has two interesting claims: the `sid` that identifies the unique session and the `sub` that identifies the user.
A valid logout token *must* have at least one of them but can also have both. Keycloak sends both for example.

We could check the `sub` claim *if present*, since we're already storing it via the `user_external_ids` table.
One of the reason I did not do that right away, is that while the default is to use `sub` claim for the `external_id`, it can actually be mapped to something else in the `user_mapping_provider`, with the `subject_claim` config with the default mapper.
So either we assume that in an OIDC context no one with a sane mind would use something else than the `sub` claim as `external_id`, or we have to run the mapper first here to be sure to have the right `external_id`, and this is very tricky because [it requires a proper "UserInfo"](https://github.com/matrix-org/synapse/blob/2f053f3f82ca174cc1c858c75afffae51af8ce0d/synapse/handlers/oidc.py#L1262-L1273), which we don't have here.

Keep in mind, the only reason there is a way to use something else than the `sub` claim is for non-OIDC OAuth 2.0 providers, so in [the docs](https://matrix-org.github.io/synapse/latest/openid.html) that would be GitHub and Gitea.

So, either:
 - we leave it as is
 - we check the `sub` claim if it is there using the `user_external_ids` table, assuming people are not mapping something else than the `sub` claim, and properly document that caveat

synapse/handlers/oidc.py
sandhose: There is a theoretical code path here where `get_devices_by_auth_provider_session_id` returns devices from two different users. That would indicate a faulty/weird OIDC Provider or a weird user mapping provider, but still possible. `delete_devices` assumes it's the same `user_id` for each device

synapse/handlers/oidc.py
sandhose: I was kinda just blindly [following the spec](https://openid.net/specs/openid-connect-backchannel-1_0.html#BCResponse). What should I do?

synapse/handlers/oidc.py
sandhose: Yep, by [`JWTClaims#validate`](https://github.com/lepture/authlib/blob/0788d705adc8232434ef497174711f2efce9879f/authlib/jose/rfc7519/claims.py#L88-L106). The manual calls are there because those are additional claims not validated by `JWTClaims`.
I'm overriding the method here to skip the regular `exp` claim validation. But now, reading the spec again, I'm not sure why past me did that, since the spec explicitly says:

>  OPs are encouraged to use short expiration times in Logout Tokens, preferably at most two minutes in the future, to prevent captured Logout Tokens from being replayable. 

So those tokens can have an `exp`, even if it's not in the [list of claims in a logout token](https://openid.net/specs/openid-connect-backchannel-1_0.html#LogoutToken)

synapse/handlers/oidc.py
sandhose: It might make sense to refractor that if/when we implement RISC events (#11514) at some point since we'd have to do the same kind of matching, but I'd leave that refactor to when we do that if you don't mind

sandhose (commented):
Thanks for the review @clokep! I replied to most comments and will fix some things now

I'd like your thoughts on the caveat I highlighted in the PR message:

> One caveat (not sure how to easily fix this): in the following scenario, the user is still logged in even if the OIDC session ended:
>
>  -  start the OIDC login from an "untrusted" client
>  -  when on the "consent" screen served by Synapse (Continuing will grant http://XXXX/ access to your account.), do not click continue yet
>  -  instead, log out from Keycloak
>  -  click on "continue" on the consent screen: the user is successfully logged in, even if the session ended in Keycloak

synapse/handlers/oidc.py
sandhose: Removed the `exp`/`nbf` validation override

synapse/handlers/oidc.py
sandhose: Done

synapse/handlers/oidc.py
clokep: That's fair, it ended up being simpler than I initially feared!

synapse/handlers/oidc.py
clokep: > we check the `sub` claim if it is there using the `user_external_ids` table, assuming people are not mapping something else than the `sub` claim, and properly document that caveat

I suspect this isn't really going to work -- I'm positive that there are people not using the `sub` claim.

synapse/handlers/oidc.py
clokep: Can we add a comment about this?

synapse/python_dependencies.py
```diff
@@ -101,7 +101,7 @@
     "saml2": [
         "pysaml2>=4.5.0",
     ],
-    "oidc": ["authlib>=0.14.0"],
+    "oidc": ["authlib>=0.14.0", "pyjwt>=1.6.4"],
```
clokep, line 104: We annoyingly list `pyjwt` twice now, which seems like it has potential for things to go haywire if we don't keep the versions in sync. I'm unsure what to do about that? (Does `authlib` depend on `pyjwt` by the way, or are they completely separate?)

sandhose: They are completely separate unfortunately :(

synapse/handlers/oidc.py
sandhose: Ha, I did not properly read your suggestion. It's not "which account it is associated with" but rather "from which IdP it is coming from"

synapse/python_dependencies.py
reivilibre: I'm not sure this is a problem — do we have reason to believe that pip won't choose the most restrictive requirement given that it should already be able to cope if a transitive requirement also diverges on what version it needs?

Alternatively if this really does concern us, it could be replaced with a `PYJWT = "pyjwt>=1.6.4"` constant and used in both places, since it's just code after all :).

clokep: It should be fine unless they conflict, I agree! 👍

I was thinking that we probably would want to keep them the same, but reconsidering; I'm unsure if that's true. If our support for JWT logins requires a newer version than we might want to bump one and not the other. This might be confusing for packagers of course. 😄

clokep:
> One caveat (not sure how to easily fix this): in the following scenario, the user is still logged in even if the OIDC session ended:
> 
> * start the OIDC login from an "untrusted" client
> * when on the "consent" screen served by Synapse (Continuing will grant http://XXXX/ access to your account.), do not click continue yet
> * instead, log out from Keycloak
> * click on "continue" on the consent screen: the user is successfully logged in, even if the session ended in Keycloak

This seems like it is probably OK? Shouldn't keycloak be rejecting this if the session ended?

synapse/handlers/oidc.py
clokep: Hmmm, I don't have a strong opinion. 🤷 Seems fine as is, I guess.

synapse/handlers/oidc.py
clokep: Isn't this matching to a particular user though, not just the IdP?

synapse/config/oidc.py
clokep: > Right now the behaviour is to delete the device, which does a "hard" logout. From a client perspective, it means it won't keep E2EE data, so you can't get back the deleted device. Maybe that's not the right behaviour, maybe that should be configurable.
> 
> Would it make sense to have instead a config parameter `backchannel_logout_behaviour` with the following values: `logout` (the default), `soft_logout`, `nothing`, and maybe implement `soft_logout`/other behaviours later?

I don't think doing a hard logout makes sense, actually. It likely makes more sense to do a soft-logout? I'm not really sure what a backchannel-logout is supposed to represent, but logging out of the IdP causing Synapse to destroy a session, causing a user to lose decryption keys sounds not great.

reivilibre:
What's the thinking on the state of this PR? Is it still wanted; does it need more review; is there something to be done to it?

sandhose (commented):
I'm getting back to the PR. I rebased on top of develop and applied a few fixes.

Still on my todo list is:

 - add tests
 - add documentation
 - figure out what to do with the potential login/logout "race condition"
 - merge some of the JWT handling code of the ID token and the logout token

sandhose:
It should be ready to review! Now that #13844 is merged, I was able to revoke login tokens in-flight (there is a test for that), and now that #13910 is merged, I was able to write proper unit tests for everything.

It just lacks documentation, which I can add later today, or in a future PR. At least it is ready to review code-wise

synapse/rest/synapse/client/oidc/backchannel_logout_resource.py
```diff
@@ -0,0 +1,35 @@
+# Copyright 2021 The Matrix.org Foundation C.I.C.
```
clokep, line 1:
```suggestion
# Copyright 2022 The Matrix.org Foundation C.I.C.
```

synapse/config/oidc.py
```diff
@@ -368,6 +374,12 @@ class OidcProviderConfig:
...
+
+    # Whetever Synapse should ignore the `sub` claim in backchannel logouts or not
```
clokep, line 380:
```suggestion
    # Whether Synapse should ignore the `sub` claim in backchannel logouts or not.
```

docs/usage/configuration/config_documentation.md
```diff
@@ -3014,6 +3014,12 @@ Options for each entry include:
      which is set to the claims returned by the UserInfo Endpoint and/or
      in the ID Token.
 
+* `backchannel_logout_enabled`: set to true to process OIDC Back-Channel Logout notifications. 
+  Those notifications are expected to be received on `/_synapse/client/oidc/backchannel_logout`.
+
+* `backchannel_logout_ignore_sub`: by default, the OIDC Back-Channel Logout feature checks that the
+  `sub` claim matches the subject claim received during login. This check can be disabled by setting
+  this to `false`.
```
clokep, line 3022: Why would you disable this?

docs/usage/configuration/config_documentation.md
```diff
@@ -3014,6 +3014,12 @@ Options for each entry include:
      which is set to the claims returned by the UserInfo Endpoint and/or
      in the ID Token.
 
+* `backchannel_logout_enabled`: set to true to process OIDC Back-Channel Logout notifications. 
+  Those notifications are expected to be received on `/_synapse/client/oidc/backchannel_logout`.
```
clokep, line 3018: What does this default to?

synapse/handlers/sso.py
```diff
@@ -790,6 +790,32 @@ def get_mapping_session(self, session_id: str) -> UsernameMappingSession:
...
+
+        for session_id, session in self._username_mapping_sessions.items():
+            if (
+                session.auth_provider_id == auth_provider_id
+                and session.auth_provider_session_id == auth_provider_session_id
+            ):
+                to_delete.append(session_id)
+
+        for session_id in to_delete:
+            logger.info("Expiring mapping session %s", session_id)
```
clokep, line 816: Maybe this is clearer?

```suggestion
            logger.info("Revoking mapping session %s", session_id)
```

synapse/handlers/sso.py
```diff
@@ -790,6 +790,32 @@ def get_mapping_session(self, session_id: str) -> UsernameMappingSession:
...
+
+        This is useful to revoke in-flight mapping sessions when a logout notification
+        arrives from the IdP.
+
+        Args:
+            auth_provider_id: A unique identifier for this SSO provider, e.g.
+                "oidc" or "saml".
+            auth_provider_session_id: The session ID got during login from the SSO IdP.
+        """
+        to_delete = []
```
clokep, line 806: Do we actually expect more than 1 session? That seems surprising to me.

synapse/storage/databases/main/registration.py
```diff
@@ -1920,6 +1920,27 @@ async def consume_login_token(self, token: str) -> LoginTokenLookupResult:
...
+            auth_provider_session_id: The session ID advertised by the SSO Identity
+                Provider
+        """
+        await self.db_pool.simple_update(
+            table="login_tokens",
+            keyvalues={
+                "auth_provider_id": auth_provider_id,
+                "auth_provider_session_id": auth_provider_session_id,
+            },
+            updatevalues={"used_ts": self._clock.time_msec()},
```
clokep, line 1940: Setting it as used invalidates it?

(Naively I was expecting a delete here.)

docs/usage/configuration/config_documentation.md
sandhose: The default is `false`, but I assumed that the wording ("set to true") was enough to suggest that the default is `false`

docs/usage/configuration/config_documentation.md
sandhose: There are (rare) cases where you want to use something else as `subject_claim`, like the entity UUID in your upstream LDAP or something like that. In those cases, the `external_id` we save for the user does not match the `sub` claim.

There is a warning on startup to check that

synapse/handlers/oidc.py
sandhose: @clokep I was referring to this for the `backchannel_logout_ignore_sub`. This is where we show a warning if the `sub` doesn't match, and ignore that check if `backchannel_logout_ignore_sub` is set to true

synapse/handlers/sso.py
sandhose: This is technically possible, although quite unlikely.
Basically, if you start two logins, and do not complete the username mapping sessions, the IDP might return the same `sid` for both cases

synapse/rest/synapse/client/oidc/backchannel_logout_resource.py
sandhose: Oh. Well, I started working on this patch in 2021 :D

synapse/storage/databases/main/registration.py
sandhose: Yeah, it does. Initially in #13844 I was deleting them, but I moved to a `used_ts` field to be able to track token reuse and use after expiration in a prometheus counter. Maybe that could be a separate column though, like `revoked_at`, because it means that logins cancelled in-flight will be considered as "token reuse" in that counter 🤷 
We can always change that later

tests/test_utils/oidc.py
```diff
@@ -183,7 +205,8 @@ def start_authorization(
         code = random_string(10)
         sid = None
         if with_sid:
-            sid = random_string(10)
+            sid = str(self.sid_counter)
+            self.sid_counter += 1
```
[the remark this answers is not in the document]
sandhose: I think it was to have collisions in SID with two distinct providers in tests.
You start two sessions in two different fake providers, they will get the same `sid` (0) but should not be confused when getting a logout token from one provider or the other

synapse/storage/databases/main/registration.py
clokep: Thanks for the explanation!

synapse/handlers/oidc.py
```diff
@@ -247,6 +266,79 @@ async def handle_oidc_callback(self, request: SynapseRequest) -> None:
...
+            raise SynapseError(400, "Missing logout_token in request")
+
+        # A JWT looks like this:
+        #    header.payload.signature
+        # where all parts are encoded with urlsafe base64.
+        # The aud and iss claims we care about are in the payload part, which
+        # is a JSON object.
+        try:
+            # This raises if there are too many or not enough segments in the token
+            _, payload, _ = logout_token.rsplit(".", 4)
```
clokep, line 294: I assume you want at most 3 parts?

```suggestion
            _, payload, _ = logout_token.rsplit(".", 2)
```

synapse/handlers/oidc.py
```diff
@@ -247,6 +266,79 @@ async def handle_oidc_callback(self, request: SynapseRequest) -> None:
...
+        except (TypeError, KeyError):
+            raise SynapseError(400, "Invalid issuer/audience in logout_token")
+
+        # Now that we know the audience and the issuer, we can figure out from
+        # what provider it is coming from
+        oidc_provider: Optional[OidcProvider] = None
+        for provider in self._providers.values():
+            if provider.issuer == issuer and any(
+                a == provider.client_id for a in audience
+            ):
```
clokep, line 332: I believe this is equivalent:

```suggestion
            if provider.issuer == issuer and provider.client_id in audience:
```

synapse/handlers/oidc.py
```diff
@@ -400,6 +493,41 @@ def _validate_metadata(self, m: OpenIDProviderMetadata) -> None:
...
+                    "OIDC Back-Channel Logout is enabled for issuer %r"
+                    "but it does not advertise support for it",
+                    self.issuer,
+                )
+
+            elif not m.get("backchannel_logout_session_supported", False):
+                logger.warning(
+                    "OIDC Back-Channel Logout is enabled and supported "
+                    "by issuer %r but it might not send session ID with "
+                    "logout tokens, which is required for the logouts to work",
```
clokep, line 508:
```suggestion
                    "OIDC Back-Channel Logout is enabled and supported "
                    "by issuer %r but it might not send a session ID with "
                    "logout tokens, which is required for the logouts to work",
```

synapse/handlers/oidc.py
```diff
@@ -400,6 +493,41 @@ def _validate_metadata(self, m: OpenIDProviderMetadata) -> None:
...
+            if not self._config.backchannel_logout_ignore_sub:
+                # If OIDC backchannel logouts are enabled, the provider mapping provider
+                # should use the `sub` claim. We verify that by mapping a dumb user and
+                # see if we get back the sub claim
+                user = UserInfo({"sub": "thisisasubject"})
+                try:
+                    subject = self._user_mapping_provider.get_remote_user_id(user)
+                    if subject != user["sub"]:
+                        raise Exception()
+                except Exception:
```
clokep, line 521: Does this really need to catch a bare exception here?

I think we only want this warning if the sub was not properly returned, not if `get_remote_user_id` throws an error?

synapse/handlers/oidc.py
```diff
@@ -400,6 +493,41 @@ def _validate_metadata(self, m: OpenIDProviderMetadata) -> None:
...
+                    logger.warning(
+                        f"OIDC Back-Channel Logout is enabled for issuer {self.issuer!r} "
+                        f"but it looks like the configured `user_mapping_provider` "
+                        f"does not use the `sub` claim as subject. If it is the case, "
+                        f"and you want Synapse to ignore the `sub` claim in OIDC "
+                        f"Back-Channel Logouts, set `backchannel_logout_ignore_sub` "
+                        f"to `true` in the issuer config."
```
clokep, line 528: I don't know if this is clearer, but I think this works:

```suggestion
                        f"OIDC Back-Channel Logout is enabled for issuer {self.issuer!r} "
                        "but it looks like the configured `user_mapping_provider` "
                        "does not use the `sub` claim as subject. If it is the case, "
                        "and you want Synapse to ignore the `sub` claim in OIDC "
                        "Back-Channel Logouts, set `backchannel_logout_ignore_sub` "
                        "to `true` in the issuer config."
```

synapse/handlers/oidc.py
```diff
@@ -1043,6 +1213,180 @@ def _remote_id_from_userinfo(self, userinfo: UserInfo) -> str:
...
+        # manually to the OP, so not specifying the backchannel-logout URI is
+        # as effective than disabling it here. It might make more sense if we
+        # support dynamic registration in Synapse at some point.
+        if not self._config.backchannel_logout_enabled:
+            logger.warning(
+                f"Received an OIDC Back-Channel Logout request from issuer {self.issuer!r} but it is disabled in config"
+            )
+            raise SynapseError(
+                400, "OpenID Connect Back-Channel Logout is disabled for this provider"
+            )
```
clokep, line 1241: This will get spit back out as a JSON error to the IdP -- is that an appropriate response?

synapse/handlers/oidc.py
```diff
@@ -1043,6 +1213,180 @@ def _remote_id_from_userinfo(self, userinfo: UserInfo) -> str:
...
+        #   Section 5.3 of [JWT].
+        alg_values = metadata.get("id_token_signing_alg_values_supported", ["RS256"])
+
+        # As per sec. 2.6:
+        #    3. Validate the iss, aud, and iat Claims in the same way they are
+        #       validated in ID Tokens.
+        # Which means the audience should contain Synapse's client_id and the
+        # issuer should be the IdP issuer
+        claims_options = {
+            "iss": {"values": [self.issuer]},
```
clokep, line 1259: I'm not quite following why we use `self.issuer` here and we use `metadata["issuer"]` when handling an ID token?

synapse/handlers/oidc.py
```diff
@@ -1043,6 +1213,180 @@ def _remote_id_from_userinfo(self, userinfo: UserInfo) -> str:
...
+        #   A Logout Token MUST be signed and MAY also be encrypted. The same
+        #   keys are used to sign and encrypt Logout Tokens as are used for ID
+        #   Tokens. If the Logout Token is encrypted, it SHOULD replicate the
+        #   iss (issuer) claim in the JWT Header Parameters, as specified in
+        #   Section 5.3 of [JWT].
+        alg_values = metadata.get("id_token_signing_alg_values_supported", ["RS256"])
+
+        # As per sec. 2.6:
+        #    3. Validate the iss, aud, and iat Claims in the same way they are
+        #       validated in ID Tokens.
```
clokep, line 1255: Do we not need / use the `iat` claim?

synapse/handlers/oidc.py
```diff
@@ -1043,6 +1213,180 @@ def _remote_id_from_userinfo(self, userinfo: UserInfo) -> str:
...
+        self._sso_handler.revoke_mapping_sessions_for_provider_session_id(
+            auth_provider_id=self.idp_id,
+            auth_provider_session_id=sid,
+        )
+
+        # Invalidate any in-flight login tokens
+        await self._store.invalidate_login_tokens_by_session_id(
+            auth_provider_id=self.idp_id,
+            auth_provider_session_id=sid,
+        )
```
clokep, line 1305: Would we *ever* do these as two separate operations?

I'm wondering if the `SsoHandler` method should call the store method so you always invalidate both mapping sessions & current login tokens together?

(Might need to rename `revoke_mapping_sessions_for_provider_session_id`...)

clokep: Actually, should the rest of this method all be grouped together into the `SsoHandler` method??

synapse/handlers/oidc.py
```diff
@@ -1043,6 +1213,180 @@ def _remote_id_from_userinfo(self, userinfo: UserInfo) -> str:
...
+                )
+                continue
+
+            logger.info(
+                "Logging out %r (device %r) via OIDC backchannel logout (sid %r).",
+                user_id,
+                device_id,
+                sid,
+            )
+            await self._device_handler.delete_devices(user_id, [device_id])
```
clokep, line 1341: Not a huge deal, but we should probably only call this once for performance reasons.

synapse/handlers/oidc.py
```diff
@@ -1043,6 +1213,180 @@ def _remote_id_from_userinfo(self, userinfo: UserInfo) -> str:
...
+
+        request.setResponseCode(200)
+        request.setHeader(b"Cache-Control", b"no-cache, no-store")
+        request.setHeader(b"Pragma", b"no-cache")
+        finish_request(request)
+
+
+class LogoutToken(JWTClaims):
```
clokep, line 1349: A shame that `authlib` doesn't provide this? 🤷  Or maybe this is just how you're supposed to use `authlib`.

synapse/handlers/oidc.py
```diff
@@ -1112,6 +1456,7 @@ def _get_secret(self) -> bytes:
         logger.info(
             "Generating new JWT for %s: %s %s", self._oauth_issuer, header, payload
         )
+        jwt = JsonWebToken(header["alg"])
```
clokep, line 1459: What's this change for? Is it ensuring `JsonWebToken` doesn't explode?

synapse/handlers/oidc.py
```diff
@@ -247,6 +266,79 @@ async def handle_oidc_callback(self, request: SynapseRequest) -> None:
...
+    async def handle_backchannel_logout(self, request: SynapseRequest) -> None:
+        """Handle an incoming request to /_synapse/client/oidc/backchannel_logout
+
+        This extracts the logout_token from the request and tries to figure out
+        which OpenID Provider it is comming from. This works by matching the iss claim
+        with the issuer and the aud claim with the client_id.
+
+        Since at this point we don't know who signed the JWT, we can't just
+        decode it using authlib since it will always verifies the signature. We
+        have to decode it manually without validating the signature.
```
clokep, line 278: Is it worth mentioning that `OidcProvider.handler_backchannel_logout` *does* then verify the token though?

docs/usage/configuration/config_documentation.md
```diff
@@ -3014,6 +3014,13 @@ Options for each entry include:
      which is set to the claims returned by the UserInfo Endpoint and/or
      in the ID Token.
 
+* `backchannel_logout_enabled`: set to `true` to process OIDC Back-Channel Logout notifications. 
+  Those notifications are expected to be received on `/_synapse/client/oidc/backchannel_logout`.
+  Defaults to `false`.
+
+* `backchannel_logout_ignore_sub`: by default, the OIDC Back-Channel Logout feature checks that the
+  `sub` claim matches the subject claim received during login. This check can be disabled by setting
+  this to `true`. Defaults to `false`.
```
clokep, line 3023: Should this *never* be set if you're using `sub` as your subject claim?

```suggestion
* `backchannel_logout_ignore_sub`: by default, the OIDC Back-Channel Logout feature checks that the
  `sub` claim matches the subject claim received during login. This check can be disabled by setting
  this to `true`. Defaults to `false`.

  You might want to disable this if the `subject_claim` returned by the mapping provider is not `sub`.
```

----

Alternately, there isn't a way to configure the subject claim for backchannel logouts, correct?

tests/rest/client/test_auth.py
```diff
@@ -1165,3 +1166,371 @@ def _txn(txn: LoggingTransaction) -> int:
...
+        config["public_baseurl"] = "http://synapse.test"
+
+        return config
+
+    def create_resource_dict(self) -> Dict[str, Resource]:
+        resource_dict = super().create_resource_dict()
+        resource_dict.update(build_synapse_client_resource_tree(self.hs))
+        return resource_dict
+
+    def is_access_token_valid(self, access_token: str) -> bool:
```
clokep, line 1193: Should we have a super-class to incorporate shared logic?

tests/rest/client/test_auth.py
```diff
@@ -1165,3 +1166,371 @@ def _txn(txn: LoggingTransaction) -> int:
...
+        first_access_token: str = login_resp["access_token"]
+        self.assertTrue(self.is_access_token_valid(first_access_token))
+
+        login_resp, second_grant = self.helper.login_via_oidc(
+            fake_oidc_server, user, with_sid=True
+        )
+        second_access_token: str = login_resp["access_token"]
+        self.assertTrue(self.is_access_token_valid(second_access_token))
+
+        self.assertNotEqual(first_grant.sid, second_grant.sid)
```
clokep, line 1251: Should we assert the access tokens are different too?

tests/rest/client/test_auth.py
```diff
@@ -1165,3 +1166,371 @@ def _txn(txn: LoggingTransaction) -> int:
...
+                    "client_secret": "test-client-secret",
+                    "scopes": ["openid"],
+                    "user_mapping_provider": {
+                        "config": {"localpart_template": "{{ user.sub }}"}
+                    },
+                    "backchannel_logout_enabled": False,
+                }
+            ]
+        }
+    )
```
clokep, line 1405: Most of these tests seem to have the same config, can we put it in `default_config`?

clokep:
@sandhose Overall this looks reasonable. I'm sure some of the questions I asked were answered previously during our last round of reviews, but I'm afraid those have been lost to time. 😢 Sorry about that!

docs/usage/configuration/config_documentation.md
sandhose: No, there isn't. The `sub` check is more of a sanity check, since `sid` are also supposed to be unique within a provider.

Also, you should be using something else than `sub` as subject claim only if you have really good reasons to do it (or if you're talking to an OAuth but non-OIDC provider like GitHub)

synapse/handlers/oidc.py
sandhose: we were importing `authlib.jose.jwt` before and it's not the case anymore. I think I did this because it was not available in all versions of authlib?

synapse/handlers/oidc.py
sandhose: TBF, OIDC Backchannel Logout is still a draft, so even if recent authlib decide to add them, it won't be there in the version we're supporting.

And yes, this is the way to define a custom JWT with the common checks

synapse/handlers/oidc.py
sandhose: There technically is a code path where a single session logs out different users at the same time (like a single account on your OP branches out to two different Matrix account depending on some claim? so you can get the same session ID?), so that's why I did it this way. And since we're probably not logging out that many sessions with a single logout token, this is probably OK in terms of performance.

synapse/handlers/oidc.py
sandhose: I guess it could be worth doing it that way, especially if we want for example to support SAML SLO at some point

synapse/handlers/oidc.py
sandhose: This check is handled done by `_verify_jwt` + whatever is in the `claim_cls`/`LogoutToken`. It should check that `iat <= now()`

synapse/handlers/oidc.py
sandhose: > If the logout succeeded, the RP MUST respond with HTTP 200 OK. However, note that some Web frameworks will substitute an HTTP 204 No Content response for an HTTP 200 OK when the HTTP body is empty. Therefore, OPs should be prepared to also process an HTTP 204 No Content response as a successful response.
>
> If the logout request was invalid or the logout failed, the RP **MUST respond with HTTP 400 Bad Request**. The response **MAY include an HTTP body consisting of a JSON object with error and error_description** parameters conveying the nature of the error that occured, which can assist with debugging. These error response parameters are used as specified in Section 5.2 of OAuth 2.0 [RFC6749]. Like in OAuth 2.0, the parameters are included in the entity-body of the HTTP response using the application/json media type. Also like in OAuth 2.0, the error parameter is REQUIRED when a response body is present and the error_description parameter is OPTIONAL. An error value of invalid_request MAY be used to indicate that there was a problem with the syntax of the logout request. Note that the information conveyed in the response body is intended to help debug deployments; it is not intended that implementations use different error values to trigger different runtime behaviors

So the 400 code is a MUST, the JSON with error + error_description is a SHOULD, soooo this is probably fine?

synapse/handlers/oidc.py
sandhose: I think we also want to catch if `get_remote_user_id` in the mapper with just the `sub` failed, since it probably means that the mapper is using something else than the `sub` claim

synapse/handlers/oidc.py
sandhose: That should actually be a 3. I'm matching one more, so that if there is 4 segment, that split will fail to destruct, raising a `ValueError`

synapse/handlers/oidc.py
sandhose: Probably, will do on Monday

synapse/handlers/oidc.py
sandhose: Done in 6deef8d41

synapse/handlers/oidc.py
sandhose: Good point. Usually `self.issuer == metadata["issuer"]`, but I changed that in e0748f60b

synapse/handlers/oidc.py
clokep: That sounds OK. Can we add a todo saying something about filling in error / error_description please.

synapse/handlers/oidc.py
clokep: Seems to be consistent now. 👍

clokep (approved):
Looks good if tests pass! 🎉

PR #11416, opened 2021-11-23
Don't start Synapse master process if `worker_app` is set

### LINKED ISSUE #11366: Stop Synapse master process from starting if `worker_app` is set
It's easy to mess up and try and start Synapse via `synapse.app.homeserver` with config that contains `worker_app` (e.g. the Docker image by default uses `synapse.app.homeserver`)

H-Shay:
Fixes #11366. Basically checks to ensure that if you are starting Synapse via `synapse.app.homeserver` that you don't have `worker_app` set in the config and refuses to start if so.

### DIFF
--- a/synapse/app/homeserver.py
+++ b/synapse/app/homeserver.py
@@ -357,7 +357,7 @@ def setup(config_options: List[str]) -> SynapseHomeServer:
         # generating config files and shouldn't try to continue.
         sys.exit(0)
 
-    if config.worker:
+    if config.worker.worker_app:
         raise ConfigError(
             "You have specified worker app/s in the config but are attempting to start a non-worker "
             "instance. Please use `python -m synapse.app.generic_worker` instead."
--- a/synapse/app/homeserver.py
+++ b/synapse/app/homeserver.py
@@ -359,7 +359,7 @@ def setup(config_options: List[str]) -> SynapseHomeServer:
 
     if config.worker.worker_app:
         raise ConfigError(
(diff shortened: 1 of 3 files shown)


synapse/app/homeserver.py
```diff
@@ -357,6 +357,13 @@ def setup(config_options: List[str]) -> SynapseHomeServer:
         # generating config files and shouldn't try to continue.
         sys.exit(0)
 
+    if config.worker:
```
erikjohnston, line 360: I would imagine this is always truthy? Since I think `config.worker` is a `WorkerConfig`? 

```suggestion
    if config.worker.worker_app:
```

H-Shay: Yes I discovered this 5 minutes before the end of the day and beginning of vacation last week :)

reivilibre (approved):
LGTM otherwise.

synapse/app/homeserver.py
```diff
@@ -357,6 +357,13 @@ def setup(config_options: List[str]) -> SynapseHomeServer:
         # generating config files and shouldn't try to continue.
         sys.exit(0)
 
+    if config.worker.worker_app:
+        raise ConfigError(
+            "You have specified worker app/s in the config but are attempting to start a non-worker "
+            "instance. Please use `python -m synapse.app.generic_worker` instead."
```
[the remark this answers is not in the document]
H-Shay: I think this makes sense-the user should either remove the `worker_app` option and start a main process or start the worker process using the provided command. Either way it forces them to think about what they are doing.

PR #11417, opened 2021-11-24
Refactor `backfilled` into specific behavior function arguments (`_persist_events_and_state_updates`)

MadLittleMods:
Refactor backfilled into specific behavior function arguments (`_persist_events_and_state_updates`)

Using keyword-only arguments (`*`) to reduce the mistakes from `backfilled` being left as a positional argument somewhere and being interpreted wrong by our new arguments.

Part of #11300

Split-off from https://github.com/matrix-org/synapse/pull/11396

### Dev notes

Call stack:

 - `_persist_events_and_state_updates` (added `use_negative_stream_ordering`)
    - `_persist_events_txn`
       - `_update_room_depths_txn`
          - `_store_room_members_txn` (added `inhibit_local_membership_updates`)

### Pull Request Checklist

### DIFF
--- a/synapse/storage/databases/main/events.py
+++ b/synapse/storage/databases/main/events.py
@@ -1228,10 +1228,8 @@ class PersistEventsStore:
             # backfilled events because backfilled events have negative
             # stream_ordering and happened in the past so we know that we don't
             # need to update the stream_ordering tip/front for the room.
-            if (
-                event.internal_metadata.stream_ordering
-                and event.internal_metadata.stream_ordering >= 0
-            ):
+            assert event.internal_metadata.stream_ordering is not None
+            if event.internal_metadata.stream_ordering >= 0:
                 txn.call_after(
                     self.store._events_stream_cache.entity_has_changed,
                     event.room_id,
--- a/synapse/storage/databases/main/events.py
+++ b/synapse/storage/databases/main/events.py
@@ -1228,7 +1228,10 @@ class PersistEventsStore:
             # backfilled events because backfilled events have negative
             # stream_ordering and happened in the past so we know that we don't
             # need to update the stream_ordering tip/front for the room.
-            if event.internal_metadata.stream_ordering >= 0:
+            if (
+                event.internal_metadata.stream_ordering
+                and event.internal_metadata.stream_ordering >= 0
+            ):
                 txn.call_after(
                     self.store._events_stream_cache.entity_has_changed,
                     event.room_id,
--- a/synapse/storage/databases/main/events.py
+++ b/synapse/storage/databases/main/events.py
@@ -124,7 +124,9 @@ class PersistEventsStore:
         current_state_for_room: Dict[str, StateMap[str]],
         state_delta_for_room: Dict[str, DeltaState],
         new_forward_extremeties: Dict[str, List[str]],
-        backfilled: bool = False,
+        *,
+        use_negative_stream_ordering: bool = False,
+        inhibit_local_membership_updates: bool = False,
     ) -> None:
         """Persist a set of events alongside updates to the current state and
         forward extremities tables.
@@ -137,7 +139,14 @@ class PersistEventsStore:
                 room state
             new_forward_extremities: Map from room_id to list of event IDs
                 that are the new forward extremities of the room.
-            backfilled
+            use_negative_stream_ordering: Whether to start stream_ordering on
+                the negative side and decrement. This should be set as True
+                for backfilled events because backfilled events get a negative
+                stream ordering so they don't come down incremental `/sync`.
+            inhibit_local_membership_updates: Stop the local_current_membership
+                from being updated by these events. This should be set to True
+                for backfilled events because backfilled events in the past do
+                not affect the current local state.
 
         Returns:
             Resolves when the events have been persisted
@@ -159,7 +168,7 @@ class PersistEventsStore:
         #
         # Note: Multiple instances of this function cannot be in flight at
         # the same time for the same room.
-        if backfilled:
+        if use_negative_stream_ordering:
             stream_ordering_manager = self._backfill_id_gen.get_next_mult(
                 len(events_and_contexts)
             )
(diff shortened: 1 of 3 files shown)


synapse/storage/persist_events.py
```diff
@@ -583,7 +583,8 @@ async def _persist_event_batch(
                 current_state_for_room=current_state_for_room,
                 state_delta_for_room=state_delta_for_room,
                 new_forward_extremeties=new_forward_extremeties,
-                backfilled=backfilled,
+                use_negative_stream_ordering=backfilled,
+                inhibit_local_membership_updates=backfilled,
```
MadLittleMods, line 587: > If there are ways of grouping some of these flags together into coherent bits of functionality, I think that would be better.
>
> https://github.com/matrix-org/synapse/pull/11396#pullrequestreview-811548838

Around that idea, there is a nicety around grouping up the behavior under the `backfilled` label.

Somewhere at the multiple top-levels when we refactor this further, we're going to have to make sure the `backfilled` logic is consistent.


Wdyt about having a static lookup map:

```py

class BackfilledAttributes:
    use_negative_stream_ordering = False
    inhibit_local_membership_updates = False
    other_more_condition = True
    extraneous_more_condition = False
    different_more_condition = True
```

And it can be used like this. I don't know. I'm not convinced of this usage and haven't figured a good clean way for the `else ...` spot (whether normal events have their own` NormalAttributes` or just be to negate the backfill one) but it seems close. Something in this vein.

```py
await self.persist_events_store._persist_events_and_state_updates(
    chunk,
    current_state_for_room=current_state_for_room,
    state_delta_for_room=state_delta_for_room,
    use_negative_stream_ordering=BackfilledAttributes.use_negative_stream_ordering if backfilled else ...,
    inhibit_local_membership_updates=BackfilledAttributes.inhibit_local_membership_updates if backfilled else ...,
    other_more_condition=BackfilledAttributes.other_more_condition if backfilled else ...,
    extraneous_more_condition=BackfilledAttributes.extraneous_more_condition if backfilled else ...,
    different_more_condition=BackfilledAttributes.different_more_condition if backfilled else ...,
)
```

synapse/storage/databases/main/events.py
```diff
@@ -1427,7 +1457,12 @@ def _store_rejected_events_txn(self, txn, events_and_contexts):
         return [ec for ec in events_and_contexts if ec[0] not in to_remove]
 
     def _update_metadata_tables_txn(
-        self, txn, events_and_contexts, all_events_and_contexts, backfilled
+        self,
+        txn,
+        *,
+        events_and_contexts,
+        all_events_and_contexts,
```
richvdh, line 1464: given that these don't have default values, does it necessarily make sense for them to be kw-only args? (Not necessarily asking for a change, just musing)

synapse/storage/databases/main/events.py
```diff
@@ -1199,22 +1220,31 @@ def _filter_events_and_contexts_for_duplicates(
...
         for event, context in events_and_contexts:
             # Remove the any existing cache entries for the event_ids
             txn.call_after(self.store._invalidate_get_event_cache, event.event_id)
-            if not backfilled:
+            # This will update the `stream_ordering` position to mark the latest
+            # event as the front of the room. This should not be done for
+            # backfilled events because backfilled events have negative
+            # stream_ordering and happened in the past so we know that we don't
+            # need to update the stream_ordering tip for the room.
+            if not use_negative_stream_ordering:
```
richvdh, line 1247: can we just infer this from `event.internal_metadata.stream_ordering >= 0` ?

synapse/storage/persist_events.py
richvdh: There's something to be said for identifying different "classes" of event (in this case, 'backfilled' and, uh, 'normal') with a list of exactly how those classes differ in behaviour.  In that case you can just pass the 'class' identifier around rather than a sea of booleans.

We do something pretty similar with room versions (see https://github.com/matrix-org/synapse/blob/v1.47.0/synapse/api/room_versions.py#L50 etc) and I don't hate it.

That said, I'm yet to be convinced that such a solution isn't over-engineered in this case - if we've only really got a few degrees of freedom here, then booleans for them might be easier. Can I reserve judgement on this?

MadLittleMods: Great comparison to `RoomVersion` 👌

> Can I reserve judgement on this?

Totally fine to push this to when we have more a need. Just wanted to throw out the inkling of an idea that the original `backfilled` had some benefit and how we could retain some of that in this new refactor ⏩

synapse/storage/databases/main/events.py
MadLittleMods: Keyword-only matches the sole usage already and seems decent to not mix-up the arguments because there isn't a straightforward order from the function name.

synapse/storage/databases/main/events.py
MadLittleMods: This seems reasonable 👍 Cool simplification we're seeing across the arguments here!

synapse/storage/databases/main/events.py
```diff
@@ -1200,21 +1211,27 @@ def _update_room_depths_txn(
...
             # Remove the any existing cache entries for the event_ids
             txn.call_after(self.store._invalidate_get_event_cache, event.event_id)
-            if not backfilled:
+            # Then update the `stream_ordering` position to mark the latest
+            # event as the front of the room. This should not be done for
+            # backfilled events because backfilled events have negative
+            # stream_ordering and happened in the past so we know that we don't
+            # need to update the stream_ordering tip/front for the room.
+            if (
+                event.internal_metadata.stream_ordering
```
richvdh, line 1232: this condition should be unnecessary - all events should have some flavour of stream_ordering by this point.

Turn it into an `assert` if you're having trouble with the typing.

richvdh (approved):
lgtm otherwise

MadLittleMods:
Thanks for the review @richvdh 🐁

Issue #11401, opened 2021-11-19
synapse 1.47 big jump in load due to `remove_hidden_devices_from_device_inbox`


skepticalwaves:
__Edit:__ Tracking expected mitigations in:

* [x] https://github.com/matrix-org/synapse/pull/11422
* [ ] https://github.com/matrix-org/synapse/pull/11421

See discussion below for context

---

### Description

After upgrading to synapse-1.47, via https://github.com/spantaleev/matrix-docker-ansible-deploy, my server experienced a huge increase in both CPU and IO load.

Upon examining the synapse prometheus/grafana stats, I found the following entires:

A large increase in master_0_background_updates in DB transactions
![Screenshot_20211119_151116](https://user-images.githubusercontent.com/66973897/142687731-e43555c6-0662-497c-9bed-34d6576a5a16.png)

And in particular, master-0_remove_hidden_device_from_inbox
![image](https://user-images.githubusercontent.com/66973897/142687011-4d43eb38-91b7-408a-b5de-c46acd5fce7b.png)

It may be related, that the HTTP pusher distribution also changed oddly:
![image](https://user-images.githubusercontent.com/66973897/142687329-1b7fddfe-44fa-4c47-801c-6d20f41e690b.png)

### Steps to reproduce

- upgraded from 1.46 to 1.47

### Version information

- **Version**: 1.47

- **Install method**: https://github.com/spantaleev/matrix-docker-ansible-deploy

- **Platform**: Ubuntu 20.04.3

dklimpel:
There is a background job to clean hidden devices from `device_inbox`. 
Depending on the size of the table, this may take a while.

skepticalwaves:
Would that be:
> Fix a long-standing bug where messages in the device_inbox table for deleted devices would persist indefinitely. Contributed by @dklimpel and @JohannesKleine. (#10969, #11212)?

dklimpel:
This job for deleted devices runs after.

The job for hidden devices is 
> Delete to_device messages for hidden devices that will never be read, reducing database size. (#11199)

skepticalwaves:
The last time such a change was published which was expected to produce unexpected load, it was documented,
https://matrix-org.github.io/synapse/develop/upgrade#re-indexing-of-events-table-on-postgres-databases

skepticalwaves:
2 days later, still have elevated load from this specific process

richvdh:
we've observed a similar problem on one of the hosts on EMS.

[...]

PR #11420, opened 2021-11-24
Improve performance of `remove_hidden_devices_from_device_inbox`

babolivier:
See https://github.com/matrix-org/synapse/issues/11401 for context.

By walking the `devices` table instead of the `device_inbox` one.

Here's the query details on abolivier.bzh's database:

```
synapse=# explain analyze SELECT user_id, device_id, stream_id
FROM devices
INNER JOIN device_inbox USING (user_id, device_id)
WHERE
    hidden = TRUE
    AND user_id >= '@foo:bar'
ORDER BY user_id
LIMIT 100;
                                                                                 QUERY PLAN                                                                                 
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=24.36..13861.73 rows=100 width=65) (actual time=106.046..106.047 rows=0 loops=1)
   ->  Merge Join  (cost=24.36..100898.78 rows=729 width=65) (actual time=106.044..106.046 rows=0 loops=1)
         Merge Cond: ((devices.user_id = device_inbox.user_id) AND (devices.device_id = device_inbox.device_id))
         ->  Index Scan using device_uniqueness on devices  (cost=0.28..180.59 rows=560 width=57) (actual time=0.075..0.911 rows=467 loops=1)
               Index Cond: (user_id >= '@foo:bar'::text)
               Filter: hidden
               Rows Removed by Filter: 147
         ->  Index Only Scan using device_inbox_user_stream_id on device_inbox  (cost=0.54..100217.89 rows=109879 width=42) (actual time=0.141..92.477 rows=101229 loops=1)
               Heap Fetches: 6968
 Planning time: 3.127 ms
 Execution time: 106.130 ms
(11 rows)
```

So not ideal but looks like an improvement over its previous incarnation.

### DIFF
--- a/synapse/storage/databases/main/deviceinbox.py
+++ b/synapse/storage/databases/main/deviceinbox.py
@@ -748,7 +748,7 @@ class DeviceInboxBackgroundUpdateStore(SQLBaseStore):
             last_user_id = progress.get("user_id", "")
 
             sql = """
-                SELECT user_id, device_id, stream_id
+                SELECT device_id, user_id, stream_id
                 FROM devices
                 INNER JOIN device_inbox USING (device_id, user_id)
                 WHERE
--- a/synapse/storage/databases/main/deviceinbox.py
+++ b/synapse/storage/databases/main/deviceinbox.py
@@ -737,31 +737,28 @@ class DeviceInboxBackgroundUpdateStore(SQLBaseStore):
         def _remove_hidden_devices_from_device_inbox_txn(
             txn: LoggingTransaction,
         ) -> int:
-            """stream_id is not unique
-            we need to use an inclusive `stream_id >= ?` clause,
-            since we might not have deleted all hidden device messages for the stream_id
-            returned from the previous query
+            """Retrieve the list of hidden devices that appear in the device_inbox table,
+            ordered by user_id so the query uses the index on the devices table.
 
             Then delete only rows matching the `(user_id, device_id, stream_id)` tuple,
             to avoid problems of deleting a large number of rows all at once
             due to a single device having lots of device messages.
             """
 
-            last_stream_id = progress.get("stream_id", 0)
+            last_user_id = progress.get("user_id", "")
 
             sql = """
-                SELECT device_id, user_id, stream_id
-                FROM device_inbox
+                SELECT user_id, device_id, stream_id
+                FROM devices
+                INNER JOIN device_inbox USING (device_id, user_id)
                 WHERE
-                    stream_id >= ?
-                    AND (device_id, user_id) IN (
-                        SELECT device_id, user_id FROM devices WHERE hidden = ?
-                    )
-                ORDER BY stream_id
+                    user_id >= ?
+                    AND hidden = ?
+                ORDER BY user_id
                 LIMIT ?
             """
 
-            txn.execute(sql, (last_stream_id, True, batch_size))
+            txn.execute(sql, (last_user_id, True, batch_size))
             rows = txn.fetchall()
 
             num_deleted = 0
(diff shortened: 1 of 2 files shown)


synapse/storage/databases/main/deviceinbox.py
```diff
@@ -737,31 +737,28 @@ async def _remove_hidden_devices_from_device_inbox(
...
+            """Retrieve the list of hidden devices that appear in the device_inbox table,
+            ordered by user_id so the query uses the index on the devices table.
 
             Then delete only rows matching the `(user_id, device_id, stream_id)` tuple,
             to avoid problems of deleting a large number of rows all at once
             due to a single device having lots of device messages.
             """
 
-            last_stream_id = progress.get("stream_id", 0)
+            last_user_id = progress.get("user_id", "")
```
richvdh, line 748: the problem here is that we might pick up a half-finished copy of the migration from the old implementation, and because we're dong things in a different order now, this won't work. We'll need to start from scratch again (which might be most easily achieved by using a differently-named field to store the progress).

synapse/storage/databases/main/deviceinbox.py
```diff
@@ -737,31 +737,28 @@ async def _remove_hidden_devices_from_device_inbox(
...
                 WHERE
-                    stream_id >= ?
-                    AND (device_id, user_id) IN (
-                        SELECT device_id, user_id FROM devices WHERE hidden = ?
-                    )
-                ORDER BY stream_id
+                    user_id >= ?
+                    AND hidden = ?
+                ORDER BY user_id
                 LIMIT ?
```
richvdh, line 758: this limit will mean that we may not find all the `device_inbox` rows for a given user - we may only end up deleting half of them before deciding to move onto the next user.

I think a better strategy is to remove the join against `device_inbox`, and just look for hidden devices, without worrying about whether they have `device_inbox` rows. Then, do a `DELETE FROM device_inbox` for each such device.

richvdh (changes requested):

synapse/storage/databases/main/deviceinbox.py
babolivier: I was going to do this initially, but the original docstring mentions that it might be a bit heavy if the device has tons of pending messages in its inbox. For example, on abolivier.bzh's database, I've got devices with over 60k rows in `device_inbox` (which aren't hidden devices but that's probably because I've already run the previous incarnation of this update), which sounds like a lot to delete in one go.

On top of that, I don't see how we might not delete entries for all devices for a given user, given the condition in the query is `user_id >= ?`, so if we don't do every device from a user we should just do the rest in the next run?

richvdh: > On top of that, I don't see how we might not delete entries for all devices for a given user, given the condition in the query is user_id >= ?, so if we don't do every device from a user we should just do the rest in the next run?

oh sorry, you're right.


However, from your analysis in the PR comment:

```
         ->  Index Only Scan using device_inbox_user_stream_id on device_inbox  (cost=0.54..100217.89 rows=109879 width=42) (actual time=0.141..92.477 rows=101229 loops=1)
               Heap Fetches: 6968
```

this is no good. It's a scan of the *entirety* of the `device_inbox_user_stream_id` index: there is no `filter` here. It's acceptable for your local db with 100k rows, but won't be for a larger db.

Honestly, looking at this again, I think we're better off rewriting it again (sorry!) to do the same as #11421 (ie, walk through the `device_inbox` table for a sequence of stream_ids. Hell, why not *combine it* with #11421?

other ideas...
 * we can probably get away with deleting all the messages for a device at once, even if its in the 10s of thousands, though it gets really nasty given we'll be considering 100 devices on the first pass.
 * maybe a two-stage loop: first find a device, then work through a range of `stream_ids` for that device on each run of the bg update. That gets annoyingly fiddly in terms of tracking state, though.

babolivier: Yeah I think you're right here, let's combine it with #11421

PR #11421, opened 2021-11-24
Improve performance of `remove_{hidden,deleted}_devices_from_device_inbox`

babolivier:
See https://github.com/matrix-org/synapse/issues/11401 for context

We do this by using the batch size to cap the stream IDs rather than using it as a limit.


synapse/storage/databases/main/deviceinbox.py
```diff
@@ -653,7 +653,7 @@ async def _remove_deleted_devices_from_device_inbox(
 
         def _remove_deleted_devices_from_device_inbox_txn(
             txn: LoggingTransaction,
-        ) -> int:
+        ) -> Tuple[int, bool]:
             """stream_id is not unique
             we need to use an inclusive `stream_id >= ?` clause,
             since we might not have deleted all dead device messages for the stream_id
```
richvdh, line 659: this is now somewhat misleading. We will delete all messages for a given stream id at once.

synapse/storage/databases/main/deviceinbox.py
```diff
@@ -664,21 +664,28 @@ def _remove_deleted_devices_from_device_inbox_txn(
             due to a single device having lots of device messages.
             """
 
-            last_stream_id = progress.get("stream_id", 0)
+            txn.execute("SELECT max(stream_id) FROM device_inbox")
```
richvdh, line 667: it might be nice to do this once when the update starts, and stash the result in the progress dict. That means we won't re-scan messages which arrive after the upgrade begins.

synapse/storage/databases/main/deviceinbox.py
```diff
@@ -664,21 +664,28 @@ def _remove_deleted_devices_from_device_inbox_txn(
...
+            row = txn.fetchone()
+            if row is None or row[0] is None:
```
richvdh, line 669: any idea what could cause `row[0]` to be `None`?

synapse/storage/databases/main/deviceinbox.py
```diff
@@ -664,21 +664,28 @@ def _remove_deleted_devices_from_device_inbox_txn(
...
+                max_stream_id = 0
```
richvdh, line 670: I'd be tempted just to bail out here. No results means there is nothing to do.
```suggestion
                return 0, True
```

synapse/storage/databases/main/deviceinbox.py
```diff
@@ -664,21 +664,28 @@ def _remove_deleted_devices_from_device_inbox_txn(
...
             sql = """
                 SELECT device_id, user_id, stream_id
                 FROM device_inbox
                 WHERE
-                    stream_id >= ?
+                    stream_id >= ? AND stream_id < ?
                     AND (device_id, user_id) NOT IN (
                         SELECT device_id, user_id FROM devices
                     )
                 ORDER BY stream_id
```
richvdh, line 685: I think the ORDER BY is redundant?

richvdh: In fact, why not just do a DELETE?

```
                DELETE FROM device_inbox
                WHERE
                    stream_id >= ? AND stream_id < ?
                    AND (device_id, user_id) NOT IN (
                        SELECT device_id, user_id FROM devices
                    )
```

synapse/storage/databases/main/deviceinbox.py
```diff
@@ -704,15 +711,15 @@ def _remove_deleted_devices_from_device_inbox_txn(
                     },
                 )
 
-            return num_deleted
+            return num_deleted, stop >= max_stream_id
```
richvdh, line 714: `num_deleted` is the wrong number here. We're interested in the number of `stream_id`s we considered (ie, `batch_size`).

changelog.d/11421.bugfix
```diff
@@ -0,0 +1 @@
+Improve performance of various background updates related to device inboxes.
```
richvdh, line 1: cf https://github.com/matrix-org/synapse/pull/11422#discussion_r756221851

richvdh (changes requested):

synapse/storage/databases/main/deviceinbox.py
richvdh: oh wait, `row[0] is None` if the table is empty. In which case, my question becomes: any idea what could cause `row` to be `None`?

changelog.d/11421.bugfix
babolivier:
```suggestion
Improve performance of various background database schema updates.
```

synapse/storage/databases/main/deviceinbox.py
babolivier: I did it to stop mypy from complaining, but I'm just going to replace it with an ignore and a comment explaining the ignore.

synapse/storage/databases/main/deviceinbox.py
babolivier: > I think the ORDER BY is redundant?

Ah so it is

> In fact, why not just do a DELETE?

We need to get data back to update the progress, but in a DELETE we can only do that using `RETURNING`, which SQLite < 3.35.0 (2021-03-12) doesn't support, and I don't think we want to break the update for them. So I'll split this with a condition on `supports_returning`, and do it in a single `DELETE` if the engine supports it (otherwise it'll do it in two steps like it does now).

synapse/storage/databases/main/deviceinbox.py
babolivier: Oh? As an admin if I read in my logs that an update that deletes stuff from my db "Processed X items" I understand that it deleted X items (not that it deleted items that have a stream ID in a range of size X).

synapse/storage/databases/main/deviceinbox.py
```diff
@@ -653,41 +653,67 @@ async def _remove_deleted_devices_from_device_inbox(
...
+                # If the database engine supports the RETURNING clause, use it and do
+                # everything in one go.
+                sql = """
+                    DELETE FROM device_inbox
+                    WHERE
+                        stream_id >= ? AND stream_id < ?
+                        AND (device_id, user_id) NOT IN (
+                            SELECT device_id, user_id FROM devices
+                        )
+                    RETURNING device_id, user_id, stream_id
```
richvdh, line 688: we don't need any of this data. all we need is the value of `stop` (so we know where to pick up next time), and we have that without RETURNING.

synapse/storage/databases/main/deviceinbox.py
richvdh: the *main* purpose of this value is to feed back into the rate-control algorithm. (The logging is a side-effect).

For the rate-control algorithm to work correctly, we need to be consistent about what sort of thing we are counting. At https://github.com/matrix-org/synapse/pull/11421/files#diff-f0d3815e503803dc732990d65d3d49bcdd08bb60d13c43bff0375cdc06b3538fR676, we're using it to count stream_ids; we therefore need to tell it how many stream_ids we processed, not how many items that resulted in us deleting.

Putting this another way: the RC algorithm will starts us off with a batch size of 100. We consider 100 stream ids, and suppose we end up deleting 1000 inbox entries, all of which takes us 1 second. The RC algorithm is aiming for runtimes of 100ms, so really we should only have done a tenth of that work.
 * if we tell it we deleted 1000 rows in that one second, then next time it will tell us to process (1000/10) items = 100 items again. No improvement.
 * if we tell it we processed 100 stream ids in that one second, then next time it will tell us to process (100/10) items = 10 items - which will be about right, given these hypothetical numbers.

---

If your next question is "what's the point of returning the item count when it should always be the same as the input batch size", well, that's a good question. Sometimes it's hard to break down work at exactly the batch size, so you might end up processing a few extra rows, and returning the number we actually processed is better for the RC algorithm.

richvdh (changes requested):

synapse/storage/databases/main/deviceinbox.py
babolivier: > At https://github.com/matrix-org/synapse/pull/11421/files#diff-f0d3815e503803dc732990d65d3d49bcdd08bb60d13c43bff0375cdc06b3538fR676, we're using it to count stream_ids; we therefore need to tell it how many stream_ids we processed, not how many items that resulted in us deleting.

Right, in this case that's a very confusing way of doing it - without your explanation I would find it very obvious that the intent is to count the rows that are deleted, which happens to be the same number as `batch_size`.

But it makes sense now, I'll update the code.

synapse/storage/databases/main/deviceinbox.py
babolivier: Right, so we don't care about adding more data to the progress?

I was trying to stick to what this comment says:

https://github.com/matrix-org/synapse/blob/9279e5dca090895a38a995574b473fbffb4815c7/synapse/storage/databases/main/deviceinbox.py#L791-L794

richvdh: I don't really know what that comment is meant to mean! I think it might have been a misunderstanding from the original author.

The only reason to store data in the progress table is so that we can read it next time and pick up where we left off. We're not using `user_id` or `device_id`, so there's no point writing them imho.

synapse/storage/databases/main/deviceinbox.py
richvdh: well, ideas about how to make this clearer are of course welcome...

synapse/storage/databases/main/deviceinbox.py
babolivier: (assuming the answer to this question is no, I'm changing this)

synapse/storage/databases/main/deviceinbox.py
```diff
@@ -733,74 +680,94 @@ async def _remove_hidden_devices_from_device_inbox(
...
-                FROM device_inbox
-                WHERE
-                    stream_id >= ?
-                    AND (device_id, user_id) IN (
-                        SELECT device_id, user_id FROM devices WHERE hidden = ?
-                    )
-                ORDER BY stream_id
-                LIMIT ?
-            """
+    def _remove_devices_from_device_inbox_txn(
```
babolivier, line 699: I've tried to be a bit clever here and have both updates call the same function since they're so similar, lmk if it's more desirable to have each run its own nested function

synapse/storage/databases/main/deviceinbox.py
```diff
@@ -733,74 +680,94 @@ async def _remove_hidden_devices_from_device_inbox(
...
+            args += (True,)  # type: ignore[assignment]
+
+        sql = (
+            """
+            DELETE FROM device_inbox
+            WHERE
+                stream_id >= ? AND stream_id < ?
+                AND (device_id, user_id) NOT IN (
+                    %s
                 )
```
richvdh, line 751: think this might be backwards for the hidden case? it ends up as

```
(device_id, user_id) NOT IN (
                    SELECT device_id, user_id FROM devices WHERE hidden=true
)
```

that `NOT` is incorrect?

synapse/storage/databases/main/deviceinbox.py
```diff
@@ -733,74 +680,94 @@ async def _remove_hidden_devices_from_device_inbox(
...
+
+        nested_select = "SELECT device_id, user_id FROM devices"
+        args = (start, stop)
+        if update_name == self.REMOVE_HIDDEN_DEVICES:
+            # If we want to remove hidden devices, select only rows from the devices table
+            # that have `hidden = TRUE`.
+            nested_select += " WHERE hidden = ?"
+            # We need to ignore mypy's error here, otherwise we can't use different
+            # arguments depending on whether we want to filter for hidden devices.
+            args += (True,)  # type: ignore[assignment]
```
richvdh, line 742: think this works fine:

```suggestion
            nested_select += " WHERE hidden"
```

richvdh (changes requested):
le sigh. I tried this on postgres - it's trying to table-scan `devices` for me, which is prohibitively expensive.

try this for size:

```sql
            DELETE FROM device_inbox di
            WHERE
                stream_id >= ? AND stream_id < ?
                AND NOT EXISTS (SELECT * FROM devices d WHERE d.device_id=di.device_id and d.user_id=di.user_id AND NOT hidden)
```

You'd think that `NOT EXISTS` would be the same as `(key) NOT IN`, but clearly not. Sorry, there's no great magic here - I just tried various ways of writing the query until I found one where postgres decided to use the index on `devices`.

Note that it does both updates in one pass. The idea is we want to filter out non-hidden devices from the inbox.

synapse/storage/databases/main/deviceinbox.py
richvdh: per my comment - I think we can actually combine the two updates into one query relatively easily.

synapse/storage/databases/main/deviceinbox.py
babolivier: Oh TIL

synapse/storage/databases/main/deviceinbox.py
babolivier: Ah yes it is

synapse/storage/databases/main/deviceinbox.py
```diff
@@ -639,9 +645,9 @@ def reindex_txn(conn):
     async def _remove_deleted_devices_from_device_inbox(
         self, progress: JsonDict, batch_size: int
     ) -> int:
-        """A background update that deletes all device_inboxes for deleted devices.
+        """No-op.
```
richvdh, line 648: better to use `register_noop_background_update` for this. Stick the "Used to be a background update that deletes all device_inboxes for deleted devices" comment next to the call to `register_noop_background_update`

synapse/storage/schema/main/delta/65/08_device_inbox_background_updates.sql
```diff
@@ -0,0 +1,18 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+-- Background update to clear the inboxes of hidden and deleted devices.
+INSERT INTO background_updates (ordering, update_name, progress_json) VALUES
+  (6508, 'remove_devices_from_device_inbox', '{}');
```
richvdh, line 18: let's give this a slightly better name. `remove_dead_devices_from_device_inbox` or something.

synapse/storage/databases/main/deviceinbox.py
```diff
@@ -733,74 +674,84 @@ async def _remove_hidden_devices_from_device_inbox(
...
+            if "max_stream_id" in progress:
+                max_stream_id = progress["max_stream_id"]
+            else:
+                txn.execute("SELECT max(stream_id) FROM device_inbox")
+                # There's a type mismatch here between how we want to type the row and
+                # what fetchone says it returns, but we silence it because we know that
+                # res can't be None.
+                res: Tuple[Optional[int]] = txn.fetchone()  # type: ignore[assignment]
+                if res[0] is None:
+                    return 0, True
```
richvdh, line 710:
```suggestion
                    # this can only happen if the `device_inbox` table is empty, in which
                    # case we have no work to do.
                    return 0, True
```

synapse/storage/databases/main/deviceinbox.py
```diff
@@ -733,74 +674,84 @@ async def _remove_hidden_devices_from_device_inbox(
...
+                else:
+                    max_stream_id = res[0]
 
-            last_stream_id = progress.get("stream_id", 0)
+            start = progress.get("stream_id", 0)
+            stop = start + batch_size
 
             sql = """
```
richvdh, line 717:
```suggestion
            # delete rows in `device_inbox` which do *not* correspond to a known,
            # unhidden device.
            sql = """
```

synapse/storage/databases/main/deviceinbox.py
```diff
@@ -733,74 +674,84 @@ async def _remove_hidden_devices_from_device_inbox(
...
-        # The task is finished when no more lines are deleted.
-        if not number_deleted:
+        if finished:
             await self.db_pool.updates._end_background_update(
-                self.REMOVE_HIDDEN_DEVICES
+                self.REMOVE_DEVICES_FROM_INBOX,
             )
 
-        return number_deleted
+        return num_deleted
```
richvdh, line 754: we're still returning the wrong thing here:
```suggestion
        return batch_size
```

richvdh (commented):
nearly there!

I think we could probably have gotten away with just turning `remove_deleted_devices_from_device_inbox` into the new update, but this is fine too.

synapse/storage/databases/main/deviceinbox.py
babolivier: Ah, I forgot `register_noop_background_update` was a thing

synapse/storage/databases/main/deviceinbox.py
```diff
@@ -743,15 +710,15 @@ def _remove_devices_from_device_inbox_txn(
 
         num_deleted, finished = await self.db_pool.runInteraction(
```
richvdh, line 711: num_deleted is now unused here - just have `_remove_dead_devices_from_device_inbox_txn` return the `finished` flag

richvdh (approved):
lgtm otherwise

synapse/storage/databases/main/deviceinbox.py
babolivier: Ah yeah indeed, good point

PR #11422, opened 2021-11-24
Lower minumum batch size to 1 for background updates

babolivier:
See https://github.com/matrix-org/synapse/issues/11401 for context


changelog.d/11422.bugfix
```diff
@@ -0,0 +1 @@
+Improve performance of various background updates related to device inboxes.
```
richvdh, line 1:
```suggestion
Improve performance of various background database schema updates.
```

(it's not *just* the device inbox stuff we're targetting here)

richvdh (approved):
lgtm

changelog.d/11422.bugfix
babolivier: Yeah but the idea is to merge this changelog with the ones from the other PRs that sprouted out of https://github.com/matrix-org/synapse/issues/11401

richvdh: update those too :)

babolivier: (ah I've just now seen your comment on #11421)

PR #11427, opened 2021-11-25
Document the usage of refresh tokens.

reivilibre:


docs/usage/configuration/user_authentication/refresh_tokens.md
```diff
@@ -0,0 +1,124 @@
...
+  Unless you want to log inactive sessions out, it is often fine to use a long
+  value here or even leave it unset (infinite).
+  Beware that making it too short will inconvenience clients that do not connect
+  very often, including mobile clients and clients of infrequent users.
+
+
+### Using refresh token expiry to log out inactive sessions
+
+If you'd like to force sessions to be logged out upon inactivity, you can enable
+refreshable access token expiry and refresh token expiry.
```
DMRobertson, line 95: I think it's worth defining activity here. "Active" here means "their client has refreshed their token", as opposed to something like a a daily-active-users metric.

docs/usage/configuration/user_authentication/refresh_tokens.md
```diff
@@ -0,0 +1,124 @@
...
+
+Assuming that clients keep their access token valid whilst they are active, and
+that you set `refresh_token_lifetime` to some value *R* and
+`refreshable_access_token_lifetime` to some value *A* (which is not bigger than *R*),
```
DMRobertson, line 99: I parse this as `A <= R`. Judging by point 1 below, I think you don't want to allow `A = R`?

docs/usage/configuration/user_authentication/refresh_tokens.md
```diff
@@ -0,0 +1,124 @@
...
+  Beware that making it too short will inconvenience clients that do not connect
+  very often, including mobile clients and clients of infrequent users.
+
+
+### Using refresh token expiry to log out inactive sessions
+
+If you'd like to force sessions to be logged out upon inactivity, you can enable
+refreshable access token expiry and refresh token expiry.
+
+Assuming that clients keep their access token valid whilst they are active, and
```
DMRobertson, line 97: > keep their access token valid

What does this mean?

docs/usage/configuration/user_authentication/refresh_tokens.md
```diff
@@ -0,0 +1,124 @@
...
+`refreshable_access_token_lifetime` to some value *A* (which is not bigger than *R*),
+then:
+ 1. sessions may become logged out if they are inactive for more than *R - A*.
+    (Explanation: if a client goes offline just before it was about to refresh its
+    access token, then it must come online before the refresh token expires.)
+ 2. sessions will always become logged out if they are inactive for more than *R*.
+    (Explanation: if a client refreshes its access token just before going offline,
+    then returns just before the refresh token expires, it will be able to continue
+    its session.)
+
```
DMRobertson, line 108: I don't follow this. What is the trigger for a session to be logged out? Your refresh token expires without being used?

DMRobertson: The key thing I didn't understand here is that the start of the inactivity period can be before your access and refresh tokens were generated. I think the example below would be improved by this.

(Is it always the case that the active access and active refresh tokens are generated at the same instant?)

docs/usage/configuration/user_authentication/refresh_tokens.md
```diff
@@ -0,0 +1,124 @@
...
+Clients are able to refresh more frequently than strictly necessary in order to
+avoid the case described by point (1) above.
+
+As an example, if refresh tokens are configured to expire after 20 minutes and
+access tokens are configured to expire after 5 minutes, then sessions will definitely
+end if inactive for 20 minutes, but may end if inactive for as few as 15 minutes.
```
DMRobertson, line 114: Something like

Suppose I am granted an access token and refresh token at 12:00. These last for 5 and 20 minutes respectively.
   - I can keep using my access token until 12:05.
   - If I go inactive at 12:00 immediately, I must use my refresh token before it expires at 12:20. This gives me a 20 minute window.
   - If I go inactive at 12:02, I must use my refresh token before it expires at 12:20. This gives me an 18 minute window.
   - If I go inactive at 12:05, I must use my refresh token before it expires at 12:20. This gives me a 15 minute window.

Feels like an information overload. I wonder if it'd be simpler to just say something like 

> Each time they login or refresh a token, clients must refresh again within time`R` to avoid being logged out.

Hmm. I'm not really convinced by my effort there.

docs/usage/configuration/user_authentication/refresh_tokens.md
```diff
@@ -0,0 +1,124 @@
...
+Traditionally, these access tokens were eternally valid (at least until the user
+explicitly chose to log out).
+
+In some cases, it may be desirable for these access tokens to expire so that the
+potential damage caused by leaking an access token is reduced.
+On the other hand, forcing a user to re-authenticate (log in again) often might
+be too much of an inconvenience.
+
+**Refresh tokens** are a mechanism to avoid some of this inconvenience whilst
+still getting most of the benefits of short access token lifetimes.
```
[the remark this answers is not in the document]
reivilibre: It's not quite OAuth/OIDC (afaik) since my experience with those shows that they use a slightly different format, but the principle is the same. Perhaps I can indeed link to that, saying the principle is the same.

docs/usage/configuration/user_authentication/refresh_tokens.md
```diff
@@ -0,0 +1,124 @@
...
+(Synapse) with the refresh token.
+
+The homeserver will then generate a new access token and refresh token for the user
+and return them. The old refresh token is invalidated and can not be used again*.
+
+Finally, refresh tokens also make it possible for sessions to be logged out if they
+are inactive for too long; see the configuration guide below.
+
+
+*To prevent issues if clients lose connection half-way through refreshing a token,
```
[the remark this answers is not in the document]
reivilibre: I had a look and came to the conclusion that it seemed to be an extension rather than core markdown, sadly.

docs/usage/configuration/user_authentication/refresh_tokens.md
```diff
@@ -0,0 +1,124 @@
...
+issued to users on login. The access token will expire after a predetermined amount
+of time, but otherwise works in the same way as before. When the access token is
+close to expiring (or has expired), the user's client should present the homeserver
+(Synapse) with the refresh token.
+
+The homeserver will then generate a new access token and refresh token for the user
+and return them. The old refresh token is invalidated and can not be used again*.
+
+Finally, refresh tokens also make it possible for sessions to be logged out if they
+are inactive for too long; see the configuration guide below.
```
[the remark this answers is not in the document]
reivilibre: I think I can appreciate why it's not very clear to you — the overall session expiry was something that I didn't come to learn about until after so I didn't think about how that gets confusing. Will see if I can think of a way to clear this up :)

docs/usage/configuration/user_authentication/refresh_tokens.md
```diff
@@ -0,0 +1,124 @@
...
+## Caveats
+
+There are some caveats:
+
+* If a third party gets both your access token and refresh token, they will be able to
+  continue to enjoy access to your session.
+  * This is still an improvement because you (the user) will notice when *your*
+    session expires and you're not able to use your refresh token.
+    That would be a giveaway that someone else has compromised your session.
+    You would be able to log in again and terminate that session.
```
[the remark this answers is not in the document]
reivilibre: Yes, usual creds. No, can't change the password (to do so, you need to have the current password; I imagine this also protects against you leaving a library computer with your account on and someone changing your password, for example).

docs/usage/configuration/user_authentication/refresh_tokens.md
```diff
@@ -0,0 +1,124 @@
...
+    Previously (with long-length access tokens), a third party that has your access
+    token could go undetected for a very long time.
+* Clients need to implement support for refresh tokens in order for them to be a
+  useful mechanism.
+  * It is up to homeserver administrators if they want to issue long-lived access
+    tokens to clients not implementing refresh tokens.
```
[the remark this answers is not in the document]
reivilibre: Long-lived and indefinite are not very distinguishable… all access tokens will expire at the heat death of the universe or perhaps before ;-). I agree that this isn't particularly clear though.

Your choices are therefore:

* long-lived/infinite lifetimes
* short-lived lifetimes
* (don't offer at all — currently you sorta get this behaviour by offering a short lifetime and I would argue that offering short-lived tokens to these clients is just as secure as doing so to refresh-capable clients)

docs/usage/configuration/user_authentication/refresh_tokens.md
```diff
@@ -0,0 +1,124 @@
...
+  In other words, the client must log in again after this time period.
+  In most cases, this can be unset (infinite) or set to a long time (years or months).
+* `refreshable_access_token_lifetime`: lifetime of access tokens that are created
+  by clients supporting refresh tokens.
+  This should be short; a good value might be 5 minutes (`5m`).
+* `nonrefreshable_access_token_lifetime`: lifetime of access tokens that are created
+  by clients which don't support refresh tokens.
+  Make this short if you want to force use of refresh tokens.
+  Make this long if you don't want to inconvenience users of clients which don't
+  support refresh tokens.
```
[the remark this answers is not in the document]
reivilibre: That's a good point. Perhaps this ought to have some special behaviour for `0` to reject logins from incapable clients.
That said, you're no worse off offering a 5 minute nonrefreshable access token if you usually offer 5 minute refreshable access token — though the user experience is a concern here: rejecting the login altogether at least protects the user from thinking they're fine and getting logged out abruptly 5 minutes later.

docs/usage/configuration/user_authentication/refresh_tokens.md
reivilibre: There's nothing too wrong with `A = R`; clients would just have to be diligent about refreshing...

The only confusing part is that the section below explaining the range of inactivities that will kick a client would then have a lower bound of zero.
In reality this depends on client behaviour and it's unlikely that clients would wait until the absolute last second to refresh their tokens, but I was trying to be helpful to admins who are thinking something like 'I want sessions inactive for 15 minutes to be logged out' and trying to paint a picture that it depends on when the clients happen to have refreshed their token — are you really OK with that meaning that an inactivity of (e.g. 5 minutes) could also trigger that behaviour?

docs/usage/configuration/user_authentication/refresh_tokens.md
reivilibre: Short answer: it's misleading but I mean that the clients keep refreshing their access tokens when needed so they always have a valid one.

I'm making a bit of an assumption that an active client wants to be syncing all the time.
To sync, it needs a valid access token.
I'm then saying that an 'active' client that has a valid refresh token but an invalid access token will refresh immediately, because otherwise it can't sync (and therefore it contradicts the fact that it's supposedly an 'active' client and all active clients sync all the time).

docs/usage/configuration/user_authentication/refresh_tokens.md
reivilibre: > the start of the inactivity period can be before your access and refresh tokens were generated. I think the example below would be improved by this.

I'm confused by this and on first hand it sounds wrong but I think it's more that I've misunderstood.

> (Is it always the case that the active access and active refresh tokens are generated at the same instant?)

Yup, maybe I should actually explain that 'refreshing an access token' means that you exchange your (old) refresh token for a new access token AND a new refresh token at the same time. It's kind of implicit knowledge to me now but I suppose I didn't actually tell you that before :P

docs/usage/configuration/user_authentication/refresh_tokens.md
reivilibre: I'm attempting (maybe not very well) to paint a picture of how homeserver admins can configure their servers to have which user-visible behaviours.
What clients should do should hopefully be obvious from the spec.

Basically, if a homeserver admin (or one of our customers...) has a problem statement like 'I want sessions to get logged out automatically if inactive for 15 minutes', this document should help them decide what values they need and hopefully explain why you get a bit of a *range* of times (with the widest range assuming that clients are lazy and only refresh at the last possible moment)...

docs/usage/configuration/user_authentication/refresh_tokens.md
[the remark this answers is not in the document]
DMRobertson: s/based on/inspired by?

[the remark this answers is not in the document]
DMRobertson: was more just thinking of a link to prior art, to provide context and make it clear that this isn't a synapse-specific concept

docs/usage/configuration/user_authentication/refresh_tokens.md
```diff
@@ -0,0 +1,135 @@
...
+* `nonrefreshable_access_token_lifetime`: lifetime of access tokens that are created
+  by clients which don't support refresh tokens.
+  Make this short if you want to effectively force use of refresh tokens.
+  Make this long if you don't want to inconvenience users of clients which don't
+  support refresh tokens.
+* `refresh_token_lifetime`: lifetime of refresh tokens.
+  Unless you want to log inactive sessions out, it is often fine to use a long
+  value here or even leave it unset (infinite).
+  Beware that making it too short will inconvenience clients that do not connect
+  very often, including mobile clients and clients of infrequent users.
```
DMRobertson, line 93: Two uses of inconvenience here. Can we explain how they're inconvenienced, i.e. that they must re-authenticate with the login credentials?

docs/usage/configuration/user_authentication/refresh_tokens.md
```diff
@@ -0,0 +1,135 @@
...
+
+**Note:** All these options only apply to newly-created tokens (i.e. tokens
+obtained from logging in or from refreshing), so won't apply retroactively.
```
DMRobertson, line 96:
```suggestion
**Note:** All four options above only apply when tokens are created (by logging in or refreshing). Changes to these settings do not apply retroactively.
```

docs/usage/configuration/user_authentication/refresh_tokens.md
```diff
@@ -0,0 +1,135 @@
...
+
+Note: this will only affect sessions using refresh tokens. You may wish to
+set a short `nonrefreshable_access_token_lifetime` to prevent this being bypassed
+by clients that do not support refresh tokens.
+
+
+#### Choosing values that guarantee permitting some inactivity
+
+It may be desirable to permit some short periods of inactivity, for example to
+accommodate brief outages in client connectivity.
```
DMRobertson, line 120: I check Element on my phone maybe once or twice a day. Are we expecting that the phone app will perioidically refresh the token in the background even when the app is not open? I'm not sure that will be something we have full control over.

docs/usage/configuration/user_authentication/refresh_tokens.md
```diff
@@ -0,0 +1,135 @@
...
+      token support. In that case, the non-refreshable access token lifetime
+      may be set to a short duration so that a similar level of security is provided.
+
+
+## Configuration Guide
+
+The following configuration options, in the `registration` section, are related:
+
+* `session_lifetime`: maximum length of a session, even if it's refreshed.
+  In other words, the client must log in again after this time period.
```
DMRobertson, line 79: Can we have a sentence like this for `refresh_token_lifetime` too?
e.g. 
> In other words, the client must refresh within this time period to maintain its session.

DMRobertson (commented):
A few smaller things:

docs/usage/configuration/user_authentication/refresh_tokens.md
reivilibre: I believe that would be reasonably classed as inactivity (and hence don't believe we need to worry about the app refreshing in the background).

I don't think people like you or me would be making use of refresh token expiry, or at least not without a pretty long expiry time on the refresh token.

DMRobertson (approved):
LGTM. Thanks for bearing with me.

reivilibre:
> LGTM. Thanks for bearing with me.

Au contraire, thanks for the feedback — it's very difficult to read docs from the position of someone who knows next to nothing after just spending the effort to figure it out and write it up... :)

PR #11428, opened 2021-11-25
Add type annotations to some of the configuration surrounding refresh tokens.

reivilibre:


synapse/handlers/register.py
```diff
@@ -812,7 +812,10 @@ class and RegisterDeviceReplicationServlet.
             assert access_token_expiry is None
             access_token = self.macaroon_gen.generate_guest_access_token(user_id)
         else:
-            if should_issue_refresh_token:
+            if (
+                should_issue_refresh_token
+                and self.refreshable_access_token_lifetime is not None
+            ):
```
DMRobertson, line 818: Bugfix?

Even if so, it's in between unreleased versions so probably doesn't need calling out in the changelog.

DMRobertson: (I wonder if `should_issue_refresh_token and self.refreshable_access_token_lifetime is None` should raise an error?)

reivilibre: It's just because Mypy can't realise that `should_issue_refresh_token` is only true if there is a lifetime configured (it's set several levels up).

Maybe `assert self.refreshable_access_token_lifetime is not None` makes more sense, though I'm not entirely sure.

DMRobertson: Putting it in the `if` condition makes me think that the condition must be considered at runtime. My preference would be an assertion with a comment like 

> Mypy can't realise that should_issue_refresh_token is only true if there is a lifetime configured (it's set several levels up).

DMRobertson (approved):
I'm happy if CI is happy!

PR #11430, opened 2021-11-25
Update MSC2918 refresh token support to confirm with the latest revision: accept the `refresh_tokens` parameter in the request body rather than in the URL parameters.

reivilibre:
The latest revision of [MSC2918](https://github.com/matrix-org/matrix-doc/blob/main/proposals/2918-refreshtokens.md#msc2918-refresh-tokens) does it this way; previously a URL parameter was used.

### DIFF
--- a/synapse/rest/client/login.py
+++ b/synapse/rest/client/login.py
@@ -28,7 +28,6 @@ from synapse.http.server import HttpServer, finish_request
 from synapse.http.servlet import (
     RestServlet,
     assert_params_in_dict,
-    parse_boolean,
     parse_bytes_from_args,
     parse_json_object_from_request,
     parse_string,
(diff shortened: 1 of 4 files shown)


synapse/rest/client/login.py
```diff
@@ -155,11 +154,15 @@ async def on_POST(self, request: SynapseRequest) -> Tuple[int, LoginResponse]:
         login_submission = parse_json_object_from_request(request)
 
         if self._msc2918_enabled:
-            # Check if this login should also issue a refresh token, as per
-            # MSC2918
-            should_issue_refresh_token = parse_boolean(
-                request, name=LoginRestServlet.REFRESH_TOKEN_PARAM, default=False
+            # Check if this registration should also issue a refresh token, as
```
DMRobertson, line 157: Why login -> registration?

DMRobertson (commented):
One typo? Otherwise seems perfectly reasonable

synapse/rest/client/login.py
reivilibre: oops, good spot

DMRobertson (approved):
LGTM. :ship: :it: !

PR #11435, opened 2021-11-26
Stabilise support for MSC2918 refresh tokens as they have now been merged into the Matrix specification.

reivilibre:
Also documents the config options, which is one of the crucial parts of stabilising a feature :).

### DIFF
--- a/synapse/rest/client/login.py
+++ b/synapse/rest/client/login.py
@@ -90,7 +90,7 @@ class LoginRestServlet(RestServlet):
         self.saml2_enabled = hs.config.saml2.saml2_enabled
         self.cas_enabled = hs.config.cas.cas_enabled
         self.oidc_enabled = hs.config.oidc.oidc_enabled
-        self._msc2918_enabled = (
+        self._refresh_tokens_enabled = (
             hs.config.registration.refreshable_access_token_lifetime is not None
         )
 
@@ -163,7 +163,7 @@ class LoginRestServlet(RestServlet):
     async def on_POST(self, request: SynapseRequest) -> Tuple[int, LoginResponse]:
         login_submission = parse_json_object_from_request(request)
 
-        if self._msc2918_enabled:
+        if self._refresh_tokens_enabled:
             # Check if this login should also issue a refresh token, as per MSC2918
             should_issue_refresh_token = login_submission.get(
                 LoginRestServlet.REFRESH_TOKEN_PARAM, False
--- a/synapse/rest/client/login.py
+++ b/synapse/rest/client/login.py
@@ -163,15 +163,16 @@ class LoginRestServlet(RestServlet):
     async def on_POST(self, request: SynapseRequest) -> Tuple[int, LoginResponse]:
         login_submission = parse_json_object_from_request(request)
 
-        if self._refresh_tokens_enabled:
-            # Check if this login should also issue a refresh token, as per MSC2918
-            should_issue_refresh_token = login_submission.get(
-                LoginRestServlet.REFRESH_TOKEN_PARAM, False
-            )
-            if not isinstance(should_issue_refresh_token, bool):
-                raise SynapseError(400, "`refresh_token` should be true or false.")
-        else:
-            should_issue_refresh_token = False
+        # Check to see if the client requested a refresh token.
+        client_requested_refresh_token = login_submission.get(
+            LoginRestServlet.REFRESH_TOKEN_PARAM, False
+        )
+        if not isinstance(client_requested_refresh_token, bool):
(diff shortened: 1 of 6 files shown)


DMRobertson:
Oh sorry, this is still draft. Can take a look when you're ready.

synapse/rest/client/login.py
```diff
@@ -166,12 +166,10 @@ async def on_POST(self, request: SynapseRequest) -> Tuple[int, LoginResponse]:
         if self._msc2918_enabled:
```
DMRobertson, line 166: Is this flag now redundant?

synapse/rest/client/login.py
```diff
@@ -166,12 +166,10 @@ async def on_POST(self, request: SynapseRequest) -> Tuple[int, LoginResponse]:
...
             # Check if this login should also issue a refresh token, as per MSC2918
             should_issue_refresh_token = login_submission.get(
-                "org.matrix.msc2918.refresh_token", False
+                LoginRestServlet.REFRESH_TOKEN_PARAM, False
```
DMRobertson, line 169: Any reason for spelling the class name explicitly here, rather than using `self`?

synapse/rest/client/login.py
DMRobertson: I think the flag is worth keeping (does this homeserver use access tokens?). But I interpreted that flag as as "are we running experimental changes for this MSC"?

Suggest renaming it to `self.offers_refresh_tokens`. But maybe you can think of a better name.

clokep: +1 to renaming it, we tend to use MSC-prefixed enabled flags to mean experimental features!

synapse/rest/client/login.py
```diff
@@ -460,9 +458,7 @@ def _get_auth_flow_dict_for_idp(idp: SsoIdentityProvider) -> JsonDict:
 
 
 class RefreshTokenServlet(RestServlet):
-    PATTERNS = client_patterns(
-        "/org.matrix.msc2918.refresh_token/refresh$", releases=(), unstable=True
-    )
+    PATTERNS = [re.compile("^" + CLIENT_API_PREFIX + "/v1/refresh$")]
```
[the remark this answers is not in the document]
reivilibre: I heard `client_patterns` was evil and thought that meant we should be avoiding them in new code, but if that's not the case then happy to put it back

synapse/rest/client/login.py
reivilibre: (fwiw the rest of this class does that for all the constants)

This is probably a Javaism, where accessing static fields/methods using `this` (`self`) is discouraged as it camouflages the fact that it's static.
I don't know if it's also bad practice in Python code and no linters told me off for it, but it's something I still think I'd avoid for clarity.

Maybe I'm just a fool?

synapse/rest/client/login.py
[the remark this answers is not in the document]
clokep: We seem to be discussing this in #11329, yes. I have not heard this.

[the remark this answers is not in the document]
richvdh: (incidentally, I also think that `CLIENT_API_PREFIX` is pretty pointless. It mostly serves as something I have to go and look up EVERY SINGLE TIME I see it)

synapse/rest/client/login.py
DMRobertson: Don't have an authoritative answer for you other than "it looks weird to me". Not worth worrying about it though.

synapse/rest/client/login.py
```diff
@@ -163,15 +163,16 @@ def on_GET(self, request: SynapseRequest) -> Tuple[int, JsonDict]:
...
+        # Check to see if the client requested a refresh token.
+        client_requested_refresh_token = login_submission.get(
+            LoginRestServlet.REFRESH_TOKEN_PARAM, False
+        )
+        if not isinstance(client_requested_refresh_token, bool):
+            raise SynapseError(400, "`refresh_token` should be true or false.")
+
+        should_issue_refresh_token = (
+            self._refresh_tokens_enabled and client_requested_refresh_token
+        )
```
DMRobertson, line 175: If they request a refresh token but it's not enabled on this server, are clients required to fall back to the old non-refreshable mechanism?

reivilibre: Yep! That'd be the same as connecting to a server that doesn't support refresh tokens.

DMRobertson (approved):
LGTM!

DMRobertson (approved):

richvdh:
for links: https://github.com/matrix-org/matrix-doc/pull/2918

PR #11440, opened 2021-11-26
Fix perspectives requests for multiple keys for the same server

richvdh:
If we tried to request multiple keys for the same server, we would end up dropping some of those requests, which would then cause errors like `Auth events could not be found` when trying to join rooms.

Introduced in #10035.

### DIFF
--- a/synapse/crypto/keyring.py
+++ b/synapse/crypto/keyring.py
@@ -667,27 +667,35 @@ class PerspectivesKeyFetcher(BaseV2KeyFetcher):
             perspective_name,
         )
 
+        request = {}
+        for queue_value in keys_to_fetch:
+            # there may be multiple requests for each server, so we have to merge
+            # them intelligently.
+            request_for_server = {
+                key_id: {
+                    "minimum_valid_until_ts": queue_value.minimum_valid_until_ts,
+                }
+                for key_id in queue_value.key_ids
+            }
+            request.setdefault(queue_value.server_name, {}).update(request_for_server)
+
+        logger.debug("Request to notary server %s: %s", perspective_name, request)
+
         try:
             query_response = await self.client.post_json(
                 destination=perspective_name,
                 path="/_matrix/key/v2/query",
-                data={
-                    "server_keys": {
-                        queue_value.server_name: {
-                            key_id: {
-                                "minimum_valid_until_ts": queue_value.minimum_valid_until_ts,
-                            }
-                            for key_id in queue_value.key_ids
-                        }
-                        for queue_value in keys_to_fetch
-                    }
-                },
+                data={"server_keys": request},
             )
         except (NotRetryingDestination, RequestSendFailed) as e:
             # these both have str() representations which we can't really improve upon
             raise KeyLookupError(str(e))
         except HttpResponseException as e:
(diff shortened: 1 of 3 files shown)


synapse/crypto/keyring.py
```diff
@@ -667,28 +667,36 @@ async def get_server_verify_key_v2_indirect(
...
+
+        logger.debug("Request to notary server %s: %s", perspective_name, request)
+
         try:
             query_response = await self.client.post_json(
                 destination=perspective_name,
                 path="/_matrix/key/v2/query",
-                data={
-                    "server_keys": {
-                        queue_value.server_name: {
```
DMRobertson, line 676: There's the problem: we'd overwrite the value associated given to a given server's name if we saw that name a second time.

synapse/crypto/keyring.py
```diff
@@ -667,28 +667,36 @@ async def get_server_verify_key_v2_indirect(
...
+        for queue_value in keys_to_fetch:
+            # there may be multiple requests for each server, so we have to merge
+            # them intelligently.
+            request_for_server = {
+                key_id: {
+                    "minimum_valid_until_ts": queue_value.minimum_valid_until_ts,
+                }
+                for key_id in queue_value.key_ids
+            }
+            request.setdefault(queue_value.server_name, {}).update(request_for_server)
```
DMRobertson, line 680: I haven't seen this idiom used much (instead of e.g. `defaultdict(dict)`),  but it a) makes sense and b) seems to be used across Synapse.

tests/crypto/test_keyring.py
```diff
@@ -577,6 +578,73 @@ def test_get_keys_from_perspectives(self):
...
+
+            self.assertEqual(
+                list(q[SERVER_NAME].keys()), [testverifykey1_id, testverifykey2_id]
+            )
+            return {"server_keys": [response1, response2]}
+
+        self.http_client.post_json.side_effect = post_json
+
+        # fire off two separate requests; they should get merged together into a
+        # single HTTP hit.
```
DMRobertson, line 624: > they should get merged together into a single HTTP hit.

I'm not sure how the test enforces this. Would a  `self.http_client.post_json.assert_called_once()` help?

richvdh: I guess it would!

Currently the test enforces it by checking that both key ids are present at line 617.

DMRobertson: Ahh yes. I think this test wouldn't spot us making an identical (correctly merged) request twice. But that's a property of the queue handling, not the merging.

DMRobertson (approved):
Thanks for explaining. LGTM!

tests/crypto/test_keyring.py
```diff
@@ -645,6 +645,9 @@ async def post_json(destination, path, data, **kwargs):
         self.assertEqual(k.verify_key.alg, "ed25519")
         self.assertEqual(k.verify_key.version, "ver2")
 
+        # finally, ensured that only one request was sent
```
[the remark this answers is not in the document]
richvdh: doh

DMRobertson (approved):

PR #11439, opened 2021-11-26
Bump ijson dependency to 3.1 (#11438)

danimo:
* [x] Pull request is based on the develop branch
* [x] Pull request includes a [changelog file](https://matrix-org.github.io/synapse/latest/development/contributing_guide.html#changelog). The entry should:
  - Be a short description of your change which makes sense to users. "Fixed a bug that prevented receiving messages from other servers." instead of "Moved X method from `EventStore` to `EventWorkerStore`.".
  - Use markdown where necessary, mostly for `code blocks`.
  - End with either a period (.) or an exclamation mark (!).
  - Start with a capital letter.
* [x] Pull request includes a [sign off](https://matrix-org.github.io/synapse/latest/development/contributing_guide.html#sign-off)
* [x] [Code style](https://matrix-org.github.io/synapse/latest/code_style.html) is correct
  (run the [linters](https://matrix-org.github.io/synapse/latest/development/contributing_guide.html#run-the-linters))

Since e81fa9264873369653171157514ff68226491fff, Synapse depends on
the use_float flag which has been introduced in ijson 3.1 and
is not available in 3.0. This is known to cause runtime errors
with send_join.

Issue #11438, opened 2021-11-26
Dependency issue: Minimum ijson version to low preventing to join rooms


passw0rd123:
### Description
Usage of `use_float` in **ijson** introduced in e81fa9264873369653171157514ff68226491fff needs a minimum version of **3.1**

Set version `"ijson>=3.0"` in https://github.com/matrix-org/synapse/blob/develop/synapse/python_dependencies.py is to low. `use_float` is introduced in ijson 3.1

Joining rooms is failing with the following error.

>**Failed to join room**  
>Failed to send_join via any server

```
 synapse.federation.federation_client: [POST-352731] Failed to send_join via matrix.org
 Traceback (most recent call last):
   File "/usr/lib/python3.6/site-packages/synapse/federation/federation_client.py", line 589, in _try_destination_list
     return await callback(destination)
   File "/usr/lib/python3.6/site-packages/synapse/federation/federation_client.py", line 768, in send_request
     response = await self._do_send_join(room_version, destination, pdu)
   File "/usr/lib/python3.6/site-packages/synapse/federation/federation_client.py", line 911, in _do_send_join
     content=pdu.get_pdu_json(time_now),
   File "/usr/lib/python3.6/site-packages/synapse/federation/transport/client.py", line 314, in send_join_v2
     parser=SendJoinParser(room_version, v1_api=False),
   File "/usr/lib/python3.6/site-packages/synapse/federation/transport/client.py", line 1313, in __init__
     use_float=True,
   File "/usr/lib64/python3.6/site-packages/ijson/common.py", line 251, in items_coro
     *_items_pipeline(backend, prefix, map_type, config)
   File "/usr/lib64/python3.6/site-packages/ijson/utils.py", line 29, in chain
     f = coro_func(f, *coro_args, **coro_kwargs)
   File "/usr/lib64/python3.6/site-packages/ijson/utils.py", line 13, in wrapper
     g = func(*args, **kwargs)
   File "/usr/lib64/python3.6/site-packages/ijson/backends/yajl2_c.py", line 17, in basic_parse_basecoro
     return _yajl2.basic_parse_basecoro(target.send, **kwargs)
 TypeError: 'use_float' is an invalid keyword argument for this function
```

### Steps to reproduce
- join existing room or follow invite
- Element shows th following error message

>**Failed to join room**  
>Failed to send_join via any server

<!--
Describe how what happens differs from what you expected.

If you can identify any relevant log snippets from _homeserver.log_, please include
[...]

PR #11217, opened 2021-10-29
Add `use_float=true` to ijson calls in Synapse

H-Shay:
Fixes  #11135, which appears to have been introduced in #9958. Prior to this change, when `ijson` parsed floats it returned a `Decimal`, which is incompatible with `canonicaljson` and caused an error wherein users were unable to join some rooms that returned a `send_join` response containing floats. I tested this solution by spinning up a homeserver,  reproducing the failed `send_join` to `#ghidra:matrix.org` (a room known to be prone to this error), making the below edits in the code used for the homeserver, restarting, and verifying that I could successfully join `ghidra:matrix.org`.

PR #11441, opened 2021-11-26
Add a test case for the SendJoinParser

DMRobertson:
This would have caught the bug #11438 introduced in #11217 and fixed in #11439. I confirmed this by `pip install ijson==3.0`, running the test and seeing a failure.

### DIFF
--- a/changelog.d/11441.misc
+++ b/changelog.d/11441.misc
@@ -1,1 +1,1 @@
-Add a test case to sanity check that we parse `/send_join` responses correctly.
+Fix a bug introduced in 1.47.0 where `send_join` could fail due to an outdated `ijson` version.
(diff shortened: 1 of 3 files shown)


danimo:
Thanks for the test!

changelog.d/11441.misc
```diff
@@ -0,0 +1 @@
+Add a test case to sanity check that we parse `/send_join` responses correctly.
```
babolivier, line 1: Can we group it with the changelog from https://github.com/matrix-org/synapse/pull/11439?

babolivier (approved):
Thanks for taking care of it! Just a nitpick with the changelog but happy for it to be merged once resolved.

tests/federation/transport/test_client.py
```diff
@@ -0,0 +1,50 @@
+import json
```
clokep, line 1: @DMRobertson Can you add a license header to this file please?

clokep: See #11460.

PR #11445, opened 2021-11-29
Support configuring the lifetime of non-refreshable access tokens separately to refreshable access tokens.

reivilibre:
Precursor to #11435 (which stabilises refresh tokens and documents these configuration flags).

`nonrefreshable_access_token_lifetime` is introduced as a configuration option so that homeserver admins can prevent bypassing auto-logout through the use
of clients that don't advertise support for refresh tokens.

### DIFF
--- a/synapse/config/registration.py
+++ b/synapse/config/registration.py
@@ -144,6 +144,17 @@ class RegistrationConfig(Config):
             )
         self.nonrefreshable_access_token_lifetime = nonrefreshable_access_token_lifetime
 
+        if (
+            self.session_lifetime is not None
+            and self.nonrefreshable_access_token_lifetime is not None
+        ):
+            if self.session_lifetime < self.nonrefreshable_access_token_lifetime:
+                raise ConfigError(
+                    "Both `session_lifetime` and `nonrefreshable_access_token_lifetime` "
+                    "configuration options have been set, but `nonrefreshable_access_token_lifetime` "
+                    " exceeds `session_lifetime`!"
+                )
+
         refresh_token_lifetime = config.get("refresh_token_lifetime")
         if refresh_token_lifetime is not None:
             refresh_token_lifetime = self.parse_duration(refresh_token_lifetime)
--- a/synapse/config/registration.py
+++ b/synapse/config/registration.py
@@ -128,6 +128,17 @@ class RegistrationConfig(Config):
             int
         ] = refreshable_access_token_lifetime
 
+        if (
+            self.session_lifetime is not None
+            and "refreshable_access_token_lifetime" in config
+        ):
+            if self.session_lifetime < self.refreshable_access_token_lifetime:
+                raise ConfigError(
+                    "Both `session_lifetime` and `refreshable_access_token_lifetime` "
+                    "configuration options have been set, but `refreshable_access_token_lifetime` "
+                    " exceeds `session_lifetime`!"
+                )
+
         # The `nonrefreshable_access_token_lifetime` applies for tokens that can NOT be
         # refreshed using a refresh token.
         # If it is None, then these tokens last for the entire length of the session,
@@ -159,6 +170,17 @@ class RegistrationConfig(Config):
         if refresh_token_lifetime is not None:
             refresh_token_lifetime = self.parse_duration(refresh_token_lifetime)
         self.refresh_token_lifetime: Optional[int] = refresh_token_lifetime
+
+        if (
+            self.session_lifetime is not None
+            and self.refresh_token_lifetime is not None
+        ):
+            if self.session_lifetime < self.refresh_token_lifetime:
+                raise ConfigError(
+                    "Both `session_lifetime` and `refresh_token_lifetime` "
+                    "configuration options have been set, but `refresh_token_lifetime` "
+                    " exceeds `session_lifetime`!"
+                )
 
         # The fallback template used for authenticating using a registration token
         self.registration_token_template = self.read_template("registration_token.html")
--- a/synapse/config/registration.py
+++ b/synapse/config/registration.py
@@ -125,6 +125,22 @@ class RegistrationConfig(Config):
             )
         self.refreshable_access_token_lifetime = refreshable_access_token_lifetime
 
+        # The `nonrefreshable_access_token_lifetime` applies for tokens that can NOT be
+        # refreshed using a refresh token.
+        # If it is None, then these tokens last for the entire length of the session,
+        # which is infinite by default.
+        # The intention behind this configuration option is to help with requiring
+        # all clients to use refresh tokens, if the homeserver administrator requires.
+        nonrefreshable_access_token_lifetime = config.get(
+            "nonrefreshable_access_token_lifetime",
+            None,
+        )
+        if nonrefreshable_access_token_lifetime is not None:
+            nonrefreshable_access_token_lifetime = self.parse_duration(
+                nonrefreshable_access_token_lifetime
+            )
+        self.nonrefreshable_access_token_lifetime = nonrefreshable_access_token_lifetime
+
         refresh_token_lifetime = config.get("refresh_token_lifetime")
         if refresh_token_lifetime is not None:
             refresh_token_lifetime = self.parse_duration(refresh_token_lifetime)
--- a/synapse/handlers/register.py
+++ b/synapse/handlers/register.py
@@ -1,4 +1,5 @@
 # Copyright 2014 - 2016 OpenMarket Ltd
+# Copyright 2021 The Matrix.org Foundation C.I.C.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/synapse/handlers/register.py
+++ b/synapse/handlers/register.py
@@ -116,6 +116,9 @@ class RegistrationHandler:
             self.pusher_pool = hs.get_pusherpool()
 
         self.session_lifetime = hs.config.registration.session_lifetime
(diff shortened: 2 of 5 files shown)


synapse/config/registration.py
```diff
@@ -128,6 +128,22 @@ def read_config(self, config, **kwargs):
             int
         ] = refreshable_access_token_lifetime
 
+        # The `nonrefreshable_access_token_lifetime` applies for tokens that can NOT be
+        # refreshed using a refresh token.
+        # If it is None, then these tokens last for the entire length of the session,
```
DMRobertson, line 133: Admins won't be able to input a literal None in the yaml; they'd have to use `null` or omit the key.

synapse/config/registration.py
```diff
@@ -128,6 +128,22 @@ def read_config(self, config, **kwargs):
...
+        # which is infinite by default.
+        # The intention behind this configuration option is to help with requiring
+        # all clients to use refresh tokens, if the homeserver administrator requires.
```
DMRobertson, line 136: You mentioned this in the docs PR but I can't remember the reasoning.

For the use case you describe, it sounds like a server admin would want a boolean flag which, if set:
- will make Synapse invalidate all non-refreshable access tokens currently in its DB on startup
- will make Synapse refuse to grant any new non-refreshable tokens

Or perhaps the point is that I can retroactively enforce a lifetime to tokens I've already granted, so that after $CUTOFF everyone will have to use refreshable tokens?

synapse/handlers/register.py
```diff
@@ -794,13 +798,25 @@ async def register_device_inner(
...
+        if self.nonrefreshable_access_token_lifetime is not None:
+            if access_token_expiry is not None:
+                # Don't allow the non-refreshable access token to outlive the
+                # session.
+                access_token_expiry = min(
+                    now_ms + self.nonrefreshable_access_token_lifetime,
+                    access_token_expiry,
+                )
+            else:
+                access_token_expiry = now_ms + self.nonrefreshable_access_token_lifetime
```
DMRobertson, line 819: Wait, session_lifetime applies to sessions using refreshable _and_ non-refreshable access tokens?

DMRobertson: If that's true, then the above seems reasonable

tests/rest/client/test_auth.py
```diff
@@ -619,7 +632,12 @@ def test_token_refresh(self):
             refresh_response.json_body["refresh_token"],
         )
 
-    @override_config({"refreshable_access_token_lifetime": "1m"})
+    @override_config(
+        {
+            "refreshable_access_token_lifetime": "1m",
+            "nonrefreshable_access_token_lifetime": "10m",
```
DMRobertson, line 638: Not sure I understand why this is needed. Isn't this test only concerned with refreshable access tokens?

tests/rest/client/test_auth.py
```diff
@@ -671,6 +689,101 @@ def test_refreshable_access_token_expiration(self):
             HTTPStatus.UNAUTHORIZED,
         )
 
+    @override_config(
+        {
+            "refreshable_access_token_lifetime": "1m",
+            "nonrefreshable_access_token_lifetime": "10m",
+        }
+    )
+    def test_differnt_expiry_for_refreshable_and_nonrefreshable_access_tokens(self):
```
DMRobertson, line 698: typo: `differnt` -> `different`

tests/rest/client/test_auth.py
```diff
@@ -671,6 +689,101 @@ def test_refreshable_access_token_expiration(self):
...
+        # Now neither token is valid.
+        self.assertFalse(self.is_access_token_valid(refreshable_access_token))
+        self.assertFalse(self.is_access_token_valid(nonrefreshable_access_token))
+
+    @override_config(
+        {"session_lifetime": "10m", "nonrefreshable_access_token_lifetime": "20m"}
+    )
+    def test_nonrefreshable_access_tokens_do_not_outlive_session(self):
+        """
+        Tests that non-refreshable access tokens do not outlive the session.
```
DMRobertson, line 760: This particular configuration doesn't seem useful! Would it make more sense to enforce that all three kinds of token lifetimes are <= session_lifetime when ingesting config?

synapse/config/registration.py
reivilibre: Omitting the key seems in order and at least is consistent with `session_lifetime` which came before me.

synapse/config/registration.py
reivilibre: Not caring about previously-issued tokens is a simplication that all the current code takes advantage of (even `session_lifetime` and friends do this).
If that worries us we could do something about it in the future but probably not relevant right now.

This option is only relevant to newly-minted non-refreshable access tokens.

Setting a short lifetime on non-refreshable access tokens is to satisfy the use case that ALL client sessions that go inactive for 15 minutes (or whatever) must be logged out.
Before this change, clients not supporting refresh tokens will be given an access token for the lifetime of `session_lifetime` (typically a long time).
After this change, that can be narrowed down further by this additional setting. For example, it could be set to `15 minutes` to give a similar level of security as refreshable tokens.
Of course, UX will suffer — so you'd want to be sure that all your users are using clients that support refresh tokens, but the goal is to close the gap that allows bypassing the auto-logout mechanism (or in other words: prevents issuing long-lived access tokens).

synapse/handlers/register.py
reivilibre: Yep, it's an upper bound on the session lifetime. It seemed reasonable to do it this way, especially considering `session_lifetime` existed before refresh tokens did — and there are considerations around using it for reauthenticating with SSO, which the proposed PR preserves.

tests/rest/client/test_auth.py
reivilibre: Ah oups, I went back and forth a bit on whether to roll it in or make it a separate test, then left that there. Reverted.

tests/rest/client/test_auth.py
reivilibre: > This particular configuration doesn't seem useful!

You're not wrong.

Perhaps it would make sense to check this in the config, at least in the case that `nonrefreshable_access_token_lifetime` has actually been explicitly set. (it defaults to infinite and it seems reasonable to let it be capped by the session lifetime otherwise.)

synapse/config/registration.py
DMRobertson: Oh sorry, I thought this comment would end up in the example yaml, where `None` wouldn't be meaningful. Ignore this.

tests/rest/client/test_auth.py
reivilibre: (I think I handled this: let me know if suits)

synapse/config/registration.py
DMRobertson: > Not caring about previously-issued tokens is a simplication that all the current code takes advantage of [...]

Okay, but ignoring existing tokens seems at odds with the use case of "requiring all clients to use refresh tokens".  That sounds like a pretty cut-and-dry security requirement which I imagine e.g. an admin operating a security-conscious company homeserver would want to enforce universally, even to existing clients.

Perhaps you're aiming to also facilitate an admin who who wants to encourage, but not require refresh tokens.

In a nutshell, I'm not sure what you mean by `requiring` in the quoted source code.

[more to follow]

DMRobertson: > Setting a short lifetime on non-refreshable access tokens is to satisfy the use case that ALL client sessions that go inactive for 15 minutes (or whatever) must be logged out.

Okay, so this is a best-effort means to continue service to old clients while still enforcing the activity requirement.

If we're going to ignore existing non-refreshable access tokens though, we can't say that we've enforced this requirement for all clients. (Perhaps you want to make further changes to that end in an upcoming PR?)

reivilibre: I guess you're making a fair point. Perhaps the answer here is that the admin needs to revoke all the existing tokens first (or, of course, set up the server using this facility from day one).
It doesn't really seem any different from the security requirement that `session_lifetime` imposes though and that is also something that is determined when an access token is issued (and not updated retrospectively).

Perhaps this is something that should be addressed through documentation? The current `session_lifetime` config option makes note of this fact, perhaps the new options need a similar note in the sample config?

DMRobertson: Reflecting on this a bit, I don't think I had any objections to the proposed source code changes. I can also see that everything you've said above corresponds faithfully to the MSC.

I think I'd be happy with
- notes in the docs and config which explains that the settings don't apply retroactively
- an explanation in the docs of how to nuke existing access tokens if admins really want to kill off existing non-refreshable tokens
- a guide/FAQ in the docs of how to enforce the things we're talking about here
  - requiring people to use refresh tokens
  - requiring people to refresh once every $THRESHOLD

DMRobertson (changes requested):
See https://github.com/matrix-org/synapse/pull/11445#discussion_r759497373 - I think we should check all three token lifetimes against the session lifetime.

DMRobertson:
> See [#11445 (comment)](https://github.com/matrix-org/synapse/pull/11445#discussion_r759497373) - I think we should check all three token lifetimes against the session lifetime.

(But am happy to be convinced otherwise)

tests/rest/client/test_auth.py
reivilibre: Sensible — included now

synapse/config/registration.py
reivilibre: Taking these suggestions to the other PRs (the config notices are already in #11435 now(

DMRobertson (approved):
LGTM!

PR #11446, opened 2021-11-29
Fix media repository failing when media store path contains symlinks

squahtx:
### Pull Request Checklist


synapse/rest/media/v1/filepath.py
```diff
@@ -43,47 +43,79 @@ def _wrapped(self: "MediaFilePaths", *args: Any, **kwargs: Any) -> str:
...
-
-        for path in paths_to_check:
-            # path may be an absolute or relative path, depending on the method being
-            # wrapped. When "appending" an absolute path, `os.path.join` discards the
-            # previous path, which is desired here.
-            normalized_path = os.path.normpath(os.path.join(self.real_base_path, path))
-            if (
-                os.path.commonpath([normalized_path, self.real_base_path])
-                != self.real_base_path
-            ):
```
squahtx, line 81: The bug was that `path` was an absolute, un-canonicalized path, so `normalized_path` was un-canonicalized while `self.real_base_path` was.

synapse/rest/media/v1/filepath.py
```diff
@@ -43,47 +43,79 @@ def _wrapped(self: "MediaFilePaths", *args: Any, **kwargs: Any) -> str:
...
+                # directory, since the media store directory may itself be a relative
+                # path.
+                if relative:
+                    normalized_path = os.path.normpath(
+                        os.path.join(self.base_path, path)
+                    )
+                else:
+                    normalized_path = os.path.normpath(path)
+
+                normalized_base_path = os.path.normpath(self.base_path)
```
clokep, line 89: Not a big deal, but should this be done outside this function, once? (Or maybe that precludes someone from modifying a symlink while synapse is running, but that sounds like a bad idea anyway.)

synapse/rest/media/v1/filepath.py
```diff
@@ -43,47 +43,79 @@ def _wrapped(self: "MediaFilePaths", *args: Any, **kwargs: Any) -> str:
...
+                # Construct the path that will ultimately be used.
+                # We cannot guess whether `path` is relative to the media store
+                # directory, since the media store directory may itself be a relative
+                # path.
+                if relative:
+                    normalized_path = os.path.normpath(
+                        os.path.join(self.base_path, path)
+                    )
+                else:
+                    normalized_path = os.path.normpath(path)
```
clokep, line 87: I wonder if it would be clearer to normalize only in one spot?

```suggestion
                if relative:
                    path = os.path.join(self.base_path, path)
                normalized_path = os.path.normpath(path)
```

synapse/rest/media/v1/filepath.py
```diff
@@ -43,47 +43,79 @@ def _wrapped(self: "MediaFilePaths", *args: Any, **kwargs: Any) -> str:
...
+
+                normalized_base_path = os.path.normpath(self.base_path)
+
+                # Check that the path lies within the media store directory.
+                # `os.path.commonpath` does not take `../`s into account and
+                # considers `a/b/c` and `a/b/c/../d` to have a common path of
+                # `a/b/c`, so we have to normalize the paths first.
```
clokep, line 94: This comment makes it sound like we an't use `commonpath`, but then we *do* use it below... do we need to do something about these issues?

tests/rest/media/v1/test_filepath.py
```diff
@@ -486,3 +487,41 @@ def _test_path_validation(
                         f"{value!r} unexpectedly passed validation: "
                         f"{method} returned {path_or_list!r}"
                     )
+
+    def test_symlink(self):
```
clokep, line 491: Should we add tests for some of the edge cases using `..` and `.`?

clokep (commented):
Looks plausible, but I'm not really sure. I left a few comments that will hopefully clarify things for me!

synapse/rest/media/v1/filepath.py
squahtx: That's a lot better, thank you!

synapse/rest/media/v1/filepath.py
squahtx: Yes, it's possible to normalize only once. `normpath` is completely independent from the filesystem as far as I know. It's only `realpath` that resolves symlinks.

synapse/rest/media/v1/filepath.py
squahtx: It sounds like the comment could do with rewording. It was intended to explain why we use `normpath` before `commonpath`. 

As for the `normpath` issue below, I'm really struggling to find a way to resolve it.

As an example, if you have a media store at `/var/lib/synapse`, symlinked to `/mnt/media`, then `/var/lib/synapse/../synapse/some_file` refers to `/mnt/synapse/some_file` and not `/mnt/media/some_file`.
`normpath` will elide the `..` and make the path look like it's legitimate.

`realpath` would resolve the path correctly, but then server admins who have symlinks _within_ the media store directory would have problems, eg. if `remote_content` was symlinked elsewhere.

Since this check is a last resort and not _the_ check for path traversal, I opted to be permissive and let the issue slide.

(todo tomorrow: revisit comments)

tests/rest/media/v1/test_filepath.py
squahtx: There's another set of checks (not touched in this PR) that inspects individual path components and rejects `..` and `.`, which makes it hard to test the unhappy paths in the decorator.

At the time I wasn't sure how to test the decorator in isolation but I think I have an idea.

synapse/rest/media/v1/filepath.py
clokep: I wonder if it is reasonable to not allow relative paths or not allow symlinks somewhere in here?

tests/rest/media/v1/test_filepath.py
clokep: Ah, I must have missed that! I'll double check.

synapse/rest/media/v1/filepath.py
squahtx: I agree, it'd be nice to not allow symlinks within the media store. On the other hand, I can definitely imagine a server admin wanting to offload one of the big media subdirectories elsewhere and expecting a symlink to work.

As for relative paths (paths containing `..` or `.`), in theory we _shouldn't_ encounter any if the main checks work. Then again this decorator is wholly unnecessary if the main checks work.

tests/rest/media/v1/test_filepath.py
squahtx: I've added a couple of negative test cases

synapse/rest/media/v1/filepath.py
squahtx: I've tried to reword the comments. Hopefully they are less confusing

clokep (approved):
Pretty sure this looks good! 👍

PR #11449, opened 2021-11-29
Expose worker & homeserver as entrypoints in `setup.py`

Ma27:
This exposes `worker` & `synapse` as executables using the [`entry_points`](https://setuptools.pypa.io/en/latest/userguide/entry_point.html)-feature of `setup.py`.

This change is used by the `matrix-synapse`-package in `nixpkgs` to automatically install `worker` & `homeserver` into `$out/bin` to make them available in the output-path of the package. This change only affects `generic_worker.py` since all other workers seem to call `generic_worker.py` and everything else happens based on the worker's configuration.

Not sure if that's useful to you, but I thought I'd file these patches anyways :) 

### Pull Request Checklist

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -154,7 +154,8 @@ setup(
     python_requires="~=3.6",
     entry_points={
         'console_scripts': [
-            'homeserver = synapse.app.homeserver:main'
+            'homeserver = synapse.app.homeserver:main',
+            'worker = synapse.app.generic_worker:main'
         ]
     },
     classifiers=[
--- a/setup.py
+++ b/setup.py
@@ -155,7 +155,7 @@ setup(
     entry_points={
         "console_scripts": [
             "synapse_homeserver = synapse.app.homeserver:main",
-            "synapse_worker = synapse.app.generic_worker:main"
+            "synapse_worker = synapse.app.generic_worker:main",
         ]
     },
     classifiers=[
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -152,6 +152,12 @@ def exec_file(path_segments):
     long_description=long_description,
     long_description_content_type="text/x-rst",
     python_requires="~=3.6",
+    entry_points={
+        'console_scripts': [
+            'homeserver = synapse.app.homeserver:main',
+            'worker = synapse.app.generic_worker:main'
```
reivilibre, line 158: I think we ought to be careful about name clashes.
Typing `homeserver` and `worker` on Ubuntu, I already get suggestions for packages to install.

```suggestion
            'synapse_homeserver = synapse.app.homeserver:main',
            'synapse_worker = synapse.app.generic_worker:main'
```

Otherwise we risk doing a Postgres where `createuser` is a command which actually creates a Postgres user (even though it doesn't sound like it) :).

reivilibre (changes requested):
Thanks!
Please would you be able to add a newsfile?
For example, `changelog.d/11449.feature`
'Expose `synapse_homeserver` and `synapse_worker` commands as entry points to run Synapse's main process and worker processes, respectively. Contributed by @Ma27.'
(Contributed by ... is up to you — optional, can also use your name if you would prefer, for example.)

Ma27:
@reivilibre thanks, done! :)

PR #11450, opened 2021-11-29
Introduce `update_user_directory_on`

### LINKED ISSUE #10339: Shared rooms API doesn't work if the worker handling the request doesn't update the user directory
#7785 added `^/_matrix/client/unstable/uk.half-shot.msc2666/user/shared_rooms/.*$`.

We check the `update_user_directory` config flag when handling this API, to basically sanity check that the tables we're about to query are roughly up-to-date (`users_who_share_private_rooms`, `users_in_public_rooms`). However, if the request is handled on a *different* synapse worker which does not have that config flag set, then the request is wrongly refused.

I'm not sure if there is an easy way to check if another Synapse process is configured to update the user directory, or whether we should just remove the check. Another option would just be to ensure this endpoint is handled on the same worker.

ShadowJonathan:
Fixes #10339

This changes the `update_user_directory` config variable to a `update_user_directory_on` variable, which more explicitly points or denotes which process is supposed to be working user directory updates. This defaults to the main process.

### Pull Request Checklist


synapse/config/server.py
```diff
@@ -321,9 +321,11 @@ def read_config(self, config, **kwargs):
                 self.presence_router_config,
             ) = load_module(presence_router_config, ("presence", "presence_router"))
 
-        # Whether to update the user directory or not. This should be set to
-        # false only if we are updating the user directory in a worker
-        self.update_user_directory = config.get("update_user_directory", True)
```
ShadowJonathan, line 326: Should we keep this config option for now (for backwards compatibility), but let it emit a warning if defined?

What should we do if it has a `False` value? Default to "synapse.app.user_dir"?

ShadowJonathan:
Blocked on https://github.com/matrix-org/sytest/issues/1173 to let sytest issues proceed correctly.

synapse/config/server.py
ShadowJonathan: I think with a `False` value, we should interpret it as "undefined", unless the local worker is the user_dir worker, then it'll suddenly "match", this should be enough to be backwards-compatible, but we should emit a compatibility warning when this happens.

clokep: We would probably want to have some sort of backwards compatibility here, yes. We usually do that when updating the configuration.

ShadowJonathan: Alright, but do you think the backwards compatibility method described in my second comment looks alright?

ShadowJonathan:
https://github.com/matrix-org/sytest/pull/1174 unblocks this, decided its easier to just introduce the config change, than process the other issue.

ShadowJonathan:
I added backwards compatibility to this, this is unblocked now.

turt2live:
I have not looked at the patch, but it'd be quite nice to keep the disabling side effect of the flag. In my use case, t2bot.io doesn't update the user directory at all and there is no worker for it, so the original intent of the setting was to save CPU/DB cycles. 

I will admit that it's a use case with sample size 1 though.

ShadowJonathan:
*Technically* it would still work, as no string can match `False`, so consider that a feature :)

turt2live:
can it get documented as a feature please? 😇 

Relying on magic behaviour isn't great.

synapse/app/admin_cmd.py
```diff
@@ -211,7 +211,7 @@ def start(config_options: List[str]) -> None:
         config.logging.no_redirect_stdio = True
 
     # Explicitly disable background processes
-    config.server.update_user_directory = False
+    config.server.worker_to_update_user_directory = ""
```
squahtx, line 214:
```suggestion
    config.server.worker_to_update_user_directory = False
```
Should this be `False`?

synapse/config/server.py
```diff
@@ -35,6 +35,14 @@
 
 logger = logging.Logger(__name__)
 
+# an object to pass to the "default" parameter, with which we can
+# `is` against to be sure we have an undefined config option.
+UNDEFINED = object()
```
squahtx, line 40: Is `UNDEFINED` purely internal to this file?
ie. Will outside code never observe `ServerConfig.worker_to_update_user_directory` with this value?

synapse/config/server.py
```diff
@@ -321,9 +329,30 @@ def read_config(self, config, **kwargs):
                 self.presence_router_config,
             ) = load_module(presence_router_config, ("presence", "presence_router"))
 
-        # Whether to update the user directory or not. This should be set to
-        # false only if we are updating the user directory in a worker
-        self.update_user_directory = config.get("update_user_directory", True)
+        # Which worker is responsible for updating the user directory,
+        # None means the main process handles this.
+        self.worker_to_update_user_directory = config.get(
```
squahtx, line 334: Could we add a type annotation to document the expected set of values?
something like `Union[None, False, str, ANY_USER_DIRECTORY_WORKER]`

synapse/app/generic_worker.py
```diff
@@ -458,21 +458,22 @@ def start(config_options: List[str]) -> None:
         config.appservice.notify_appservices = False
 
     if config.worker.worker_app == "synapse.app.user_dir":
-        if config.server.update_user_directory:
+        from synapse.config.server import ANY_USER_DIRECTORY_WORKER
```
squahtx, line 461: I think there's a preference for having imports at the top of the file where possible.

synapse/rest/client/shared_rooms.py
```diff
@@ -42,19 +42,11 @@ def __init__(self, hs: "HomeServer"):
...
         self, request: SynapseRequest, user_id: str
     ) -> Tuple[int, JsonDict]:
 
-        if not self.user_directory_active:
-            raise SynapseError(
-                code=400,
-                msg="The user directory is disabled on this server. Cannot determine shared rooms.",
-                errcode=Codes.FORBIDDEN,
-            )
-
```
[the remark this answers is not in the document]
ShadowJonathan: Yes, see https://github.com/matrix-org/synapse/issues/10339#issuecomment-965523795 and https://github.com/matrix-org/synapse/issues/10339#issuecomment-971509019.

This makes the admin explicitly select a worker to handle this, and de-magicifies the "oh, the user directory handles this if its false", with this value, any worker (if it is specified and running) can handle the update, the sample config is updated to reflect this behaviour, and warn the admin that it's their responsibility to see that the worker handling it is online.

The default value is `None`, meaning the main process handles this.

synapse/config/server.py
ShadowJonathan: The last one isn't a type, so that wouldn't pass mypy

synapse/app/generic_worker.py
ShadowJonathan: This is to avoid cyclic dependencies, but i think it could be okay enough, let me check.

synapse/config/server.py
ShadowJonathan: No, though i'll add an `assert` to make that sure

synapse/rest/client/shared_rooms.py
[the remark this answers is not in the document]
squahtx: I see, thanks for explaining!

synapse/config/server.py
ShadowJonathan: Yup, i'll add a comment though

image

squahtx: That's very true. Elsewhere in the code we appear to create a single entry enum: https://github.com/matrix-org/synapse/blob/develop/synapse/util/caches/cached_call.py#L72

Though if we're going to make an enum we might as well roll the "main process" and "nobody" values into it.

squahtx: A comment is a good alternative!

squahtx (approved):
lgtm!

squahtx:
@ShadowJonathan The user directory sytests without workers seem to be failing after I merged the sytest PR. Could you investigate?

docs/sample_config.yaml
```diff
@@ -142,6 +142,17 @@ presence:
...
+# User Directory Updates
+#
+# If specified, a worker with that name will be the only one able to update
+# the user directory. Important for querying shared rooms. If the worker is
+# specified, but not running, the user directory may become outdated.
+#
+# Defaults to null, meaning the main process will handle this.
+#
+# Set to false to make sure no process will handle the user_directory.
+#worker_to_update_user_directory: null
```
richvdh, line 154: please have a look at https://github.com/matrix-org/synapse/blob/develop/docs/code_style.md#configuration-file-format and ensure this matches.

docs/sample_config.yaml
```diff
@@ -142,6 +142,17 @@ presence:
...
 
+# User Directory Updates
+#
+# If specified, a worker with that name will be the only one able to update
+# the user directory. Important for querying shared rooms. If the worker is
+# specified, but not running, the user directory may become outdated.
+#
+# Defaults to null, meaning the main process will handle this.
+#
+# Set to false to make sure no process will handle the user_directory.
```
richvdh, line 153: so it can be a string, or a boolean, or `null`? ew. can we find a way to limit it to just one type or `null`?

docs/sample_config.yaml
richvdh: It should probably be in the `workers` section?

richvdh: and maybe it should be called `update_user_directory_on` for consistency with `run_background_tasks_on` ?

docs/sample_config.yaml
richvdh, line 153: why would you want *no* process to handle the user directory? why would that be useful?

docs/sample_config.yaml
richvdh: (should it default to the same value as `run_background_tasks_on`?)

richvdh (commented):
sorry, just some drive-by comments from looking at the sample config. Afraid I haven't looked at anything else!

richvdh:
> Afraid I haven't looked at anything else!

oh, but others have, so yay.

richvdh (changes requested):
config setting needs updates, as above

docs/sample_config.yaml
ShadowJonathan: We could, the extra type was suggested by @turt2live to basically "disable" the config, though maybe we can change that to be `""`, an empty string.

ShadowJonathan: See [this comment](https://github.com/matrix-org/synapse/pull/11450#issuecomment-987109870)

docs/sample_config.yaml
ShadowJonathan: I've addressed all three first issues, but the last issue is still unresolved, i'm leaving this thread open for discussion on that.

docs/sample_config.yaml
```diff
@@ -2657,6 +2657,12 @@ opentracing:
 #
 #run_background_tasks_on: worker1
 
+# The worker that is used to run user directory update tasks
+# (e.g. users in public rooms, shared rooms between users.).
+#
+# If not provided, or null, this defaults to the main process.
+#update_user_directory_on: null
```
richvdh, line 2664: this still doesn't match the conventions specified at https://github.com/matrix-org/synapse/blob/develop/docs/code_style.md#configuration-file-format. Please go and read it.

richvdh (changes requested):

docs/sample_config.yaml
richvdh: I guess it's only fair to give you some clues.

> A line consisting of only `#`.
> ...
> For other options, the example should give some non-default value which is likely to be useful to the reader.

synapse/config/workers.py
```diff
@@ -26,6 +27,16 @@
 )
 from .server import ListenerConfig, parse_listener_def
 
+logger = logging.Logger(__name__)
+
+# an object to pass to the "default" parameter, with which we can
+# `is` against to be sure we have an undefined config option.
+UNDEFINED = object()
```
richvdh, line 34: convention is to call things like this `SENTINEL` rather than `UNDEFINED`, though tbh I think it would be better to avoid the need for it with an explicit test like:

```
if 'update_user_directory' in config:
```

synapse/config/workers.py
```diff
@@ -26,6 +27,16 @@
...
+
+# an object to pass to self.worker_to_update_user_directory if
+# update_user_directory was defined, is used to compare with 'is'.
+ANY_USER_DIRECTORY_WORKER = object()
```
richvdh, line 38: can we not avoid the need for this ugliness by deciding whether we should update the user directory in this module whether we should update the user dir (in the same way that we do with `run_background_tasks`, `start_pushers`, `send_federation`, etc)?

It feels like we've invented a completely different wheel when we already have a perfectly functional one.

richvdh (changes requested):

synapse/config/workers.py
ShadowJonathan: Ah, wasn't aware of that trick, will apply it. :+1:

synapse/config/workers.py
ShadowJonathan: You mean something along the lines of checking for `worker_app` here if `update_user_directory` is defined, and then setting `update_user_directory_on = worker_name` depending on that?

docs/sample_config.yaml
ShadowJonathan: Ah gotcha, will update. :+1:

synapse/config/workers.py
richvdh: no, I mean getting rid of `config.worker.update_user_directory_on` altogether, and just having a boolean `config.worker.update_user_directory`, in much the same way as we have a `config.worker.send_federation`.

ShadowJonathan: That discussion is out of scope for this PR, please comment on https://github.com/matrix-org/synapse/issues/10339 with that, as this PR is applying the suggestions made there.

richvdh: I can't see anything about the implementation in #10339? It seems to just be talking about the format of the config file (which I'm not suggesting you change).

ShadowJonathan: https://github.com/matrix-org/synapse/issues/10339 talks about the disposition that we can't be sure the user directory will be updated at any time, I suggested to make the config value accept None to indicate explicitly that the user directory will be updated on the main process, and any string value for any worker name.

@anoadragon453 approved of this at the time, this PR is executing what is said in that issue, if I were to move ahead on this PR, then I need to know the consensus, so please discuss on that issue if you think this is not a good idea.

reivilibre: @ShadowJonathan I think Rich is not referring to making any config format change, but instead that the config object only stores true/false whether **this** worker will run them.

I refer to `run_background_tasks` (one of the fields Rich mentioned):
```python
        background_tasks_instance = config.get("run_background_tasks_on") or "master"
        self.run_background_tasks = (
            self.worker_name is None and background_tasks_instance == "master"
        ) or self.worker_name == background_tasks_instance
```

Put in other words:
A Synapse process doesn't care exactly *who (which process)* is updating the directory. It only cares whether *it's* the one doing it or not.

ShadowJonathan: That snippet has the same behaviour as this current config change, it doesn't test for true/false, are you sure you got the right snippet?

ShadowJonathan: With some private clarification from rich (after i saw we were miscommunicating on this issue), i'll alter the internal variable to `update_user_directory: bool`, which is derived from the yaml config value `update_user_directory_on: Option[str]`, with backwards compat from the yaml config option `update_user_directory`

synapse/app/generic_worker.py
```diff
@@ -458,21 +458,14 @@ def start(config_options: List[str]) -> None:
...
-        if config.server.update_user_directory:
+        if not config.worker.update_user_directory:
             sys.stderr.write(
-                "\nThe update_user_directory must be disabled in the main synapse process"
-                "\nbefore they can be run in a separate worker."
-                "\nPlease add ``update_user_directory: false`` to the main config"
-                "\n"
+                "\nThe user directory worker is not allowed to perform user directory updates per the config."
+                "\nPlease add or edit "
+                f'``update_user_directory_on: "{config.worker.worker_name}"`` in the config\n'
```
richvdh, line 465: errr this seems a bit of an odd constraint. why isn't the user directory worker allowed to do the user directory updates?

(the previous code was checking that user directory updates were disabled on the *main* process before they were enabled implicitly on a worker by setting `worker_app: synapse.app.user_dir`)

richvdh: also note: if you're updating the text of the warning, use a `"""` at the top rather than this clunky multiple-lines-of-`"` thing.

synapse/config/workers.py
```diff
@@ -296,6 +306,43 @@ def read_config(self, config, **kwargs):
...
+
+                if update_user_directory:
+                    # Main process should update
+                    uud_result = is_master_process
+                else:
+                    # user directory worker should update
+                    uud_result = self.worker_app == "synapse.app.user_dir"
+            else:
+                # Both values are undefined, assume None for update_user_directory_on
+                uud_result = is_master_process
```
richvdh, line 342: this seems a very verbose, and not very clear, way of writing this logic.

```suggestion
        if "update_user_directory" in config:
            # Backwards compatibility case
            logger.warning(USER_UPDATE_DIRECTORY_DEPRECATION_WARNING)

            if config["update_user_directory"]:
                # Main process should update
                update_user_directory_on = "master"
            else:
                # user directory worker should update
                update_user_directory_on = "synapse.app.user_dir"
        else:
            update_user_directory_on = config.get("update_user_directory_on") or "master"
        
        self.update_user_directory = (
            is_master_process and update_user_directory_on == "master"
        ) or self.worker_name == update_user_directory_on
```

tests/utils.py
```diff
@@ -169,7 +169,7 @@ def default_config(name, parse=False):
         "default_room_version": DEFAULT_ROOM_VERSION,
         # disable user directory updates, because they get done in the
         # background, which upsets the test runner.
-        "update_user_directory": False,
+        "update_user_directory_on": "",
```
richvdh, line 172: if the intention is to set this to a non-existent worker, might be better to make that clearer:

```suggestion
        "update_user_directory_on": "non-existent worker",
```

tests/handlers/test_user_directory.py
```diff
@@ -56,7 +56,7 @@ class UserDirectoryTestCase(unittest.HomeserverTestCase):
 
     def make_homeserver(self, reactor: MemoryReactor, clock: Clock) -> HomeServer:
         config = self.default_config()
-        config["update_user_directory"] = True
+        config["update_user_directory_on"] = None
```
richvdh, line 59: suggest a comment here (and below) to clarify what exactly this is doing.

richvdh (changes requested):
this is definitely much clearer and easier to follow now, thanks! A few notes though.

synapse/config/workers.py
```diff
@@ -296,6 +306,43 @@ def read_config(self, config, **kwargs):
             self.worker_name is None and background_tasks_instance == "master"
         ) or self.worker_name == background_tasks_instance
```
richvdh, line 307: this should probably be fixed to use `instance_name` while you're here:

```suggestion
            self.instance_name == background_tasks_instance
        )
```

synapse/config/workers.py
```diff
@@ -296,6 +306,43 @@ def read_config(self, config, **kwargs):
...
 
+        # update_user_directory_on controls which worker is responsible for updating the user directory.
+        # None means that the main process should handle this, so does the absence of the config key.
+        #
+        # Note that due to backwards compatibility, we must also test for update_user_directory, and apply it if
+        # update_user_directory_on is not defined at the same time.
+        self.update_user_directory: bool
```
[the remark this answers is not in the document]
ShadowJonathan: It's to explicitly set out the intended type, and this'll also snag mypy if any of the branches does not set this variable.

synapse/app/generic_worker.py
ShadowJonathan: Good point, should i turn this into a warning then, to log when `update_user_directory_on` isnt specified to the local worker (if it's a user_dir one)?

synapse/config/workers.py
ShadowJonathan: This'd make the above `self.update_user_directory: bool` assertion also redundant.

synapse/app/generic_worker.py
```diff
@@ -126,6 +126,11 @@
 
 logger = logging.getLogger("synapse.app.generic_worker")
 
+DIRECTORY_UPDATE_WARNING = """
+The user directory worker is not allowed to perform user directory updates per the config.
```
richvdh, line 130: right, I see the confusion here now.

"not allowed to" sounds like I've tried to enable something invalid.  Maybe `s/not allowed to/not configured to/`.

But I don't really understand why this is a thing we need to warn about. Why is it a problem if the updates are done on the main process?

Note that the current wording sounds very much like an error, indicating something that definitely won't work.

synapse/app/generic_worker.py
```diff
@@ -459,12 +464,7 @@ def start(config_options: List[str]) -> None:
 
     if config.worker.worker_app == "synapse.app.user_dir":
         if not config.worker.update_user_directory:
-            sys.stderr.write(
-                "\nThe user directory worker is not allowed to perform user directory updates per the config."
-                "\nPlease add or edit "
-                f'``update_user_directory_on: "{config.worker.worker_name}"`` in the config\n'
-            )
-            sys.exit(1)
+            logger.warning(DIRECTORY_UPDATE_WARNING.format(config.worker.worker_name))
```
richvdh, line 467: given that this happens before logging is set up, I think it probably just ends up in stderr anyway. It may be best to stick to `sys.stderr.write` for consistency.

synapse/config/workers.py
```diff
@@ -302,46 +301,36 @@ def read_config(self, config, **kwargs):
         # No effort is made to ensure only a single instance of these tasks is
         # running.
         background_tasks_instance = config.get("run_background_tasks_on") or "master"
-        self.run_background_tasks = (
-            self.worker_name is None and background_tasks_instance == "master"
-        ) or self.worker_name == background_tasks_instance
+        self.run_background_tasks = self.instance_name == background_tasks_instance
```
richvdh, line 304: suggest explicit parens here for clarity
```suggestion
        self.run_background_tasks = (self.instance_name == background_tasks_instance)
```

synapse/config/workers.py
```diff
@@ -302,46 +301,36 @@ def read_config(self, config, **kwargs):
...
-                    uud_result = self.worker_app == "synapse.app.user_dir"
-            else:
-                # Both values are undefined, assume None for update_user_directory_on
-                uud_result = is_master_process
+            update_user_directory_on = (
+                config.get("update_user_directory_on") or "master"
+            )
 
-        self.update_user_directory = uud_result
+        self.update_user_directory: bool = (
```
richvdh, line 331: really quite sure that the `:bool` is redundant here

tests/handlers/test_user_directory.py
```diff
@@ -56,6 +56,7 @@ class UserDirectoryTestCase(unittest.HomeserverTestCase):
 
     def make_homeserver(self, reactor: MemoryReactor, clock: Clock) -> HomeServer:
         config = self.default_config()
+        # Always process user directory updates on this main homeserver process.
```
richvdh, line 59: this isn't terribly clear.
```suggestion
        # Enable user directory updates
```

tests/handlers/test_user_directory.py
```diff
@@ -1014,7 +1014,7 @@ class TestUserDirSearchDisabled(unittest.HomeserverTestCase):
 
     def make_homeserver(self, reactor: MemoryReactor, clock: Clock) -> HomeServer:
         config = self.default_config()
-        config["update_user_directory"] = True
+        config["update_user_directory_on"] = None
```
richvdh, line 1018: this still needs a comment

richvdh (changes requested):

synapse/config/workers.py
ShadowJonathan: Black clears those, iirc

tests/handlers/test_user_directory.py
ShadowJonathan: Yes, but this comment exists to explain *why* the user updates flag is set to true there.

synapse/app/generic_worker.py
ShadowJonathan: > Maybe s/not allowed to/not configured to/.

👍

> Why is it a problem if the updates are done on the main process?

That's not the problem, the problem is is that the configuration does not appoint the local worker to do the updates.

The config can now appoint any worker to do this job by its name, so the job of the user directory worker isn't that unique anymore, but it's uniqueness is still supported by synapse.

> Note that the current wording sounds very much like an error

That's because it is, the local worker has a "type" of being a designated user update worker, while the configuration does not allow it to do just that. This is more or less a warning that "your config, as-is, is contradictory"

The previous version exited the process with `1` when hitting this point, should that be re-introduced?

synapse/config/workers.py
ShadowJonathan: Yeah it does, not much i can do here

richvdh:
Ugh, I remain hopelessly confused about what is going on here, and how it fixes the original issue (https://github.com/matrix-org/synapse/issues/10339). I have commented there.

ShadowJonathan:
I'll be closing this PR in light of the retrospective at https://github.com/matrix-org/synapse/issues/10339#issuecomment-1039175679, where it became clear this PR is the wrong solution to this problem.

PR #11318, opened 2021-11-12
Add support for `/_matrix/client/v3` APIs

aaronraimist:
Step 2 of https://github.com/matrix-org/synapse/issues/11079

I also put `/keys/device_signing/upload` on the stable v3 API while I was here. After a quick look, that was the only new API in Matrix 1.1 I could find that wasn't already using a stable API prefix.

### Pull Request Checklist

* [x] Pull request is based on the develop branch
* [x] Pull request includes a [changelog file](https://matrix-org.github.io/synapse/latest/development/contributing_guide.html#changelog). The entry should:
  - Be a short description of your change which makes sense to users. "Fixed a bug that prevented receiving messages from other servers." instead of "Moved X method from `EventStore` to `EventWorkerStore`.".
  - Use markdown where necessary, mostly for `code blocks`.
  - End with either a period (.) or an exclamation mark (!).
  - Start with a capital letter.
* [x] Pull request includes a [sign off](https://matrix-org.github.io/synapse/latest/development/contributing_guide.html#sign-off)
* [x] [Code style](https://matrix-org.github.io/synapse/latest/code_style.html) is correct
  (run the [linters](https://matrix-org.github.io/synapse/latest/development/contributing_guide.html#run-the-linters))

Issue #11079, opened 2021-10-14
Matrix 1.1 checklist


turt2live:
Matrix 1.1 is coming up and consists of major changes to how the protocol operates from a versioning perspective. From a Synapse perspective, the ordered details are:

1. [ ] Support r0.6.1 as normal (https://github.com/matrix-org/synapse/issues/11064)
2. [ ] Add/duplicate listeners for all `/r0` endpoints to also listen on `/v3`
3. [ ] Add support for the feature list below.
4. [ ] Advertise `v1.1` in `/_matrix/client/versions` (after the spec has been released)

The community has also expressed interest in having a config flag to test what the transition would look like in their clients for when servers start advertising the Matrix version. I think this can be accomplished by adding an unstable flag to `/versions` of `org.matrix.msc2844` from the original MSC - no server implementations (to my knowledge) used the flag for anything during the MSC period, so we can and should use it for something like this. Its presence would denote whether Step 2 above is complete.

----

There's not currently anything that is expected to be added to v1.1 at this point, so the following curated features become stable:

* [ ] Key backup
* [ ] Knocking
* [ ] Social Login (`/login/sso/redirect/:idp`)
* [ ] Cross signing
* [ ] Secret Storage
* [ ] `device_id` in login fallback ([MSC2604](https://github.com/matrix-org/matrix-doc/pull/2604))
* [ ] 404 `M_NOT_FOUND` errors on psuh rule endpoints ([MSC2663](https://github.com/matrix-org/matrix-doc/pull/2663))
* [ ] `reason` and `score` are optional on the `/report` API
* [ ] Guests can get access to the members of a room ([MSC2689](https://github.com/matrix-org/matrix-doc/pull/2689))
* [ ] `device_id` on `/account/whoami`
* [ ] Room version 7 (for knocking)
* [ ] Room version 8 and 9 (time permitting)
* [ ] Add `/_matrix/identity/versions` API (if you use version checking against identity services)

Do note that Spaces are not included in this 1.1 release at present. The MSCs which have landed are still stable enough to use stable identifiers, however the functionality is not presently included in a released version of the spec. This should be fine as there's no API endpoints to worry about for versioning purposes, though servers may need to be aware of the `m.space` room type.

----

[...]

PR #11451, opened 2021-11-29
Register the login redirect endpoint for v3

clokep:
This is a follow-up to #11318 which registers the login redirect endpoint for v3 since it doesn't use `client_patterns`.

There are a variety of other places that `/r0` is mentioned in our code, some of those are places where Synapse generates a URL back to itself (e.g. login fallback, SSO redirect URLs, etc.) We don't *have* to update those today, but we should eventually.

Part of #11079

### DIFF
--- a/synapse/rest/client/room.py
+++ b/synapse/rest/client/room.py
@@ -1138,12 +1138,12 @@ class RoomSpaceSummaryRestServlet(RestServlet):
 
 
 class RoomHierarchyRestServlet(RestServlet):
-    PATTERNS = [
+    PATTERNS = (
         re.compile(
             "^/_matrix/client/(v1|unstable/org.matrix.msc2946)"
             "/rooms/(?P<room_id>[^/]*)/hierarchy$"
         ),
-    ]
+    )
 
     def __init__(self, hs: "HomeServer"):
         super().__init__()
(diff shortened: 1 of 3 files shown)


synapse/rest/client/room.py
```diff
@@ -1138,12 +1138,12 @@ async def on_POST(
 
 
 class RoomHierarchyRestServlet(RestServlet):
-    PATTERNS = [
+    PATTERNS = (
```
clokep, line 1141: The list -> tuple change is is just for consistency with other URLs.

clokep:
I tested this by changing `test_login` to use `v3` URLs and noted that the tests did not all pass until this change was made. I don't know what we should be testing against though.

clokep:
I wonder if we should put this in v1.48.0 since that is when this change was initially made?

DMRobertson:
> We don't have to update those today, but we should eventually.

Mind filing an issue, if we don't have one already?

> I wonder if we should put this in v1.48.0 since that is when this change was initially made?

I think that ship has sailed now.

DMRobertson (approved):
Thank you for catching this!

Are there any other endpoints we might have missed? Is there some way to do a survey?

clokep:
> Are there any other endpoints we might have missed? Is there some way to do a survey?

I did check for other endpoints not using `client_patterns`, but could have missed one still.

clokep:
> > We don't have to update those today, but we should eventually.
> 
> Mind filing an issue, if we don't have one already?

See #11474.

PR #11453, opened 2021-11-29
Eliminate a few `Any`s in `LruCache` type hints

squahtx:
`Any`s make IDE navigation difficult, so let's get rid of them.

### DIFF
--- a/synapse/util/caches/deferred_cache.py
+++ b/synapse/util/caches/deferred_cache.py
@@ -22,6 +22,7 @@ from typing import (
     Iterable,
     MutableMapping,
     Optional,
+    Sized,
     TypeVar,
     Union,
     cast,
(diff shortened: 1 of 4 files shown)


synapse/util/caches/deferred_cache.py
```diff
@@ -104,7 +104,13 @@ def metrics_cb() -> None:
             max_size=max_entries,
             cache_name=name,
             cache_type=cache_type,
-            size_callback=(lambda d: len(d) or 1) if iterable else None,
+            size_callback=(
+                (lambda d: len(d) or 1)  # type: ignore[arg-type]
```
reivilibre, line 108: wonder how this compares on a 'unfortunate but hard to work around':

```suggestion
                (lambda d: len(cast(Sized, d)) or 1)
```

I *think* (but don't feel too strongly) that I prefer the cast here, mostly because it shows exactly what you need to override, whereas the comment makes me have to think about which 'arg' it's talking about.

reivilibre (approved):
LGTM; change at your discretion.

PR #11459, opened 2021-11-30
synctl stop: wait for processes to exit

richvdh:
If you're trying to shut down Synapse, it's rather handy if it *actually* shuts
down before you move on.

### DIFF
--- a/synctl
+++ b/synctl
@@ -109,15 +109,14 @@ def start(pidfile: str, app: str, config_files: Iterable[str], daemonize: bool) 
         return False
 
 
-def stop(pidfile: str, app: str) -> bool:
+def stop(pidfile: str, app: str) -> int:
     """Attempts to kill a synapse worker from the pidfile.
     Args:
         pidfile: path to file containing worker's pid
         app: name of the worker's appservice
 
     Returns:
-        True if the process stopped successfully
-        False if process was already stopped or an error occured
+        process id, or 0 if the process was not running
     """
 
     if os.path.exists(pidfile):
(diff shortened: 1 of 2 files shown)


synctl
```diff
@@ -109,38 +109,36 @@ def start(pidfile: str, app: str, config_files: Iterable[str], daemonize: bool)
...
+def stop(pidfile: str, app: str) -> int:
     """Attempts to kill a synapse worker from the pidfile.
     Args:
         pidfile: path to file containing worker's pid
         app: name of the worker's appservice
 
     Returns:
-        True if the process stopped successfully
-        False if process was already stopped or an error occured
+        process id, or 0 if the process was not running
```
reivilibre, line 119: On *Linux*, pid 0 is reserved for the scheduler ... is that guaranteed more generally on other systems, though?
This is likely to be inconsequential but I would probably go for an `Option[int]` if it were me, if only because I feel it expresses the intent more clearly.

synctl
```diff
@@ -288,32 +286,23 @@ def main():
...
 
     if action == "stop" or action == "restart":
-        has_stopped = True
+        running_pids = []
         for worker in workers:
-            if not stop(worker.pidfile, worker.app):
-                # A worker could not be stopped.
-                has_stopped = False
+            pid = stop(worker.pidfile, worker.app)
+            if pid:
```
reivilibre, line 292: in which case this would naturally be

```suggestion
            if pid is not None:
```

synctl
richvdh: fair enough.

richvdh:
>  apparently FreeBSD allocates PIDs randomly,

Interesting, TIL. Possibly we should wait for the pidfile to disappear instead? But if synapse crashes, it might not clean up its pidfile...

reivilibre:
> > apparently FreeBSD allocates PIDs randomly,
> 
> Interesting, TIL. Possibly we should wait for the pidfile to disappear instead? But if synapse crashes, it might not clean up its pidfile...

I don't have an answer for you, though relying on Synapse cleaning up its pidfile does seem a bit unreliable; it's hard to think of anything better than what's here now.
(I checked: `waitpid()` and friends are only for your children — didn't find any bulletproof approaches for waiting on processes that aren't your children.)

PR #11465, opened 2021-11-30
Additional type hints for config module.

clokep:
Adds some missing type hints to `synapse.config` on my journey to fully type it.

It will be several more PRs before that's done, unfortunately.

### DIFF
--- a/synapse/config/appservice.py
+++ b/synapse/config/appservice.py
@@ -1,4 +1,5 @@
 # Copyright 2015, 2016 OpenMarket Ltd
+# Copyright 2021 The Matrix.org Foundation C.I.C.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -13,14 +14,14 @@
 # limitations under the License.
 
 import logging
-from typing import Dict
+from typing import Dict, List
 from urllib import parse as urlparse
 
 import yaml
 from netaddr import IPSet
 
 from synapse.appservice import ApplicationService
-from synapse.types import UserID
+from synapse.types import JsonDict, UserID
 
 from ._base import Config, ConfigError
 
@@ -30,12 +31,12 @@ logger = logging.getLogger(__name__)
 class AppServiceConfig(Config):
     section = "appservice"
 
-    def read_config(self, config, **kwargs):
+    def read_config(self, config, **kwargs) -> None:
         self.app_service_config_files = config.get("app_service_config_files", [])
         self.notify_appservices = config.get("notify_appservices", True)
         self.track_appservice_user_ips = config.get("track_appservice_user_ips", False)
 
-    def generate_config_section(cls, **kwargs):
+    def generate_config_section(cls, **kwargs) -> str:
         return """\
         # A list of application service config files to use
         #
@@ -50,7 +51,9 @@ class AppServiceConfig(Config):
         """
 
 
-def load_appservices(hostname, config_files):
+def load_appservices(
+    hostname: str, config_files: List[str]
+) -> List[ApplicationService]:
     """Returns a list of Application Services from the config files."""
     if not isinstance(config_files, list):
         logger.warning("Expected %s to be a list of AS config files.", config_files)
@@ -93,7 +96,9 @@ def load_appservices(hostname, config_files):
     return appservices
 
 
-def _load_appservice(hostname, as_info, config_filename):
+def _load_appservice(
+    hostname: str, as_info: JsonDict, config_filename: str
+) -> ApplicationService:
     required_string_fields = ["id", "as_token", "hs_token", "sender_localpart"]
     for field in required_string_fields:
         if not isinstance(as_info.get(field), str):
@@ -115,9 +120,9 @@ def _load_appservice(hostname, as_info, config_filename):
     user_id = user.to_string()
 
     # Rate limiting for users of this AS is on by default (excludes sender)
-    rate_limited = True
-    if isinstance(as_info.get("rate_limited"), bool):
-        rate_limited = as_info.get("rate_limited")
+    rate_limited = as_info.get("rate_limited")
+    if not isinstance(rate_limited, bool):
+        rate_limited = True
 
     # namespace checks
     if not isinstance(as_info.get("namespaces"), dict):
--- a/synapse/config/cache.py
+++ b/synapse/config/cache.py
@@ -1,4 +1,4 @@
-# Copyright 2019 Matrix.org Foundation C.I.C.
+# Copyright 2019-2021 Matrix.org Foundation C.I.C.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -17,6 +17,8 @@ import re
 import threading
 from typing import Callable, Dict, Optional
 
+import attr
+
 from synapse.python_dependencies import DependencyException, check_requirements
 
 from ._base import Config, ConfigError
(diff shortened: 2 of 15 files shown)


synapse/config/appservice.py
```diff
@@ -30,12 +31,12 @@
 class AppServiceConfig(Config):
     section = "appservice"
 
-    def read_config(self, config, **kwargs):
```
DMRobertson, line 33: `config: Dict[str, Any]`?

kwargs seems to be `config_dir_path: str` and `data_dir_path: str` but I don't think these are frequently used. But let me not scope creep...

synapse/config/appservice.py
```diff
@@ -30,12 +31,12 @@
...
+    def read_config(self, config, **kwargs) -> None:
         self.app_service_config_files = config.get("app_service_config_files", [])
         self.notify_appservices = config.get("notify_appservices", True)
         self.track_appservice_user_ips = config.get("track_appservice_user_ips", False)
 
-    def generate_config_section(cls, **kwargs):
+    def generate_config_section(cls, **kwargs) -> str:
```
DMRobertson, line 39: Do these methods ever do anything other than returning a string literal unconditionally?

I guess maybe one of them fills in a hostname or similar.

(I'm just wondering if this might be better off just being a static yaml file somewhere. Again, scope creep though.)

synapse/config/appservice.py
```diff
@@ -50,7 +51,9 @@ def generate_config_section(cls, **kwargs):
         """
 
 
-def load_appservices(hostname, config_files):
+def load_appservices(
+    hostname: str, config_files: List[str]
+) -> List[ApplicationService]:
     """Returns a list of Application Services from the config files."""
     if not isinstance(config_files, list):
         logger.warning("Expected %s to be a list of AS config files.", config_files)
```
DMRobertson, line 59: Bit surprised this is  a warning rather than a refusal to startup!

synapse/config/cas.py
```diff
@@ -28,7 +29,7 @@ class CasConfig(Config):
 
     section = "cas"
 
-    def read_config(self, config, **kwargs):
+    def read_config(self, config, **kwargs) -> None:
```
DMRobertson, line 32: `config: Dict[str, Any]` while we're here.

synapse/config/cas.py
```diff
@@ -51,7 +52,7 @@ def read_config(self, config, **kwargs):
             self.cas_displayname_attribute = None
             self.cas_required_attributes = []
 
-    def generate_config_section(self, config_dir_path, server_name, **kwargs):
+    def generate_config_section(self, config_dir_path, server_name, **kwargs) -> str:
```
DMRobertson, line 55: `str` for both named args I think?

synapse/config/oidc.py
```diff
@@ -36,7 +36,7 @@
 class OIDCConfig(Config):
     section = "oidc"
 
-    def read_config(self, config, **kwargs):
+    def read_config(self, config, **kwargs) -> None:
```
DMRobertson, line 39: `config: Dict[str, Any]` too

synapse/config/saml2.py
```diff
@@ -33,7 +34,7 @@
 )
 
 
-def _dict_merge(merge_dict, into_dict):
+def _dict_merge(merge_dict: dict, into_dict: dict) -> None:
     """Do a deep merge of two dicts
 
     Recursively merges `merge_dict` into `into_dict`:
```
DMRobertson, line 40: Perhaps emphasise that this edits `into_dict` in-place?

synapse/config/saml2.py
```diff
@@ -184,7 +185,7 @@ def read_config(self, config, **kwargs):
 
     def _default_saml_config_dict(
         self, required_attributes: set, optional_attributes: set
```
DMRobertson, line 187: `Set[str]` for both, I think?

synapse/config/sso.py
```diff
@@ -49,7 +49,7 @@ class SSOConfig(Config):
 
     section = "sso"
 
-    def read_config(self, config, **kwargs):
+    def read_config(self, config, **kwargs) -> None:
```
DMRobertson, line 52: `config: Dict[str, Any]`

synapse/config/cache.py
```diff
@@ -162,7 +164,7 @@ def generate_config_section(self, **kwargs):
           #sync_response_cache_duration: 2m
         """
 
-    def read_config(self, config, **kwargs):
+    def read_config(self, config, **kwargs) -> None:
```
DMRobertson, line 167: `config: Dict[str, Any]`

synapse/config/database.py
```diff
@@ -119,7 +120,7 @@ def __init__(self, *args, **kwargs):
 
         self.databases = []
 
-    def read_config(self, config, **kwargs):
+    def read_config(self, config, **kwargs) -> None:
```
DMRobertson, line 123: `config: Dict[str, Any]`

synapse/config/database.py
```diff
@@ -163,12 +164,12 @@ def read_config(self, config, **kwargs):
...
             self.set_databasepath(database_path)
 
-    def generate_config_section(self, data_dir_path, **kwargs):
+    def generate_config_section(self, data_dir_path, **kwargs) -> str:
         return DEFAULT_CONFIG % {
             "database_path": os.path.join(data_dir_path, "homeserver.db")
         }
 
-    def read_arguments(self, args):
+    def read_arguments(self, args: argparse.Namespace) -> None:
```
DMRobertson, line 172: Aside: I wish there was a way for mypy to know which attributes would be available on `args` at typechecktime.

synapse/config/logger.py
```diff
@@ -142,13 +143,13 @@
 class LoggingConfig(Config):
     section = "logging"
 
-    def read_config(self, config, **kwargs):
+    def read_config(self, config, **kwargs) -> None:
```
DMRobertson, line 146: `config: Dict[str, Any]`

synapse/config/server.py
```diff
@@ -1291,7 +1294,9 @@ def is_threepid_reserved(reserved_threepids, threepid):
     return False
 
 
-def read_gc_thresholds(thresholds):
+def read_gc_thresholds(
+    thresholds: Optional[List[Any]],
```
DMRobertson, line 1298: Maybe `List[SupportsInt]` if you're feeling fancy!

DMRobertson: Also, below on +1314, looks like `listener` could be `Dict: [str, Any]`

DMRobertson (commented):
Thanks for this. A few thoughts:

synapse/config/appservice.py
clokep: I have a separate branch built on this which adds type hints to `read_config`.

synapse/config/appservice.py
clokep: Some of them return dynamic strings.

synapse/config/server.py
clokep: I'm not sure. That's what is *expected* by these methods, but not what is necessarily passed in?

synapse/config/database.py
clokep: You mean if you could note what type of arguments exist in the `Namespace`?

synapse/config/appservice.py
DMRobertson: Ahh, my mistake.

synapse/config/database.py
DMRobertson: I was thinking more that I'd like mypy to be able to deduce which args would exist, but an explict annotation would be good too!

DMRobertson:
@clokep I think this needs a linter script run.

synapse/config/server.py
clokep: I'm not really sure what the best practice is here, but this is where we're taking any YAML junk and ensuring it is correct, so I think it would be a lie to say it has to accept integers?

clokep:
> @clokep I think this needs a linter script run.

Doh, fixed!

synapse/config/appservice.py
clokep: No problem! I should have included it in the description. :) There's a lot to these classes, so trying to break it up logically...but maybe only logically in my head. :)

synapse/config/server.py
DMRobertson: Agreed. (The argument I had in mind was that mypy could detect type errors if we described the intended inputs. But it's not going to be able to do that when the yaml comes at runtime.)

DMRobertson (approved):
Looks great. Bring on the next round!

PR #11467, opened 2021-11-30
Add a note about huge pages to our Postgres doc

H-Shay:
As the title states.

### DIFF
--- a/docs/postgres.md
+++ b/docs/postgres.md
@@ -118,6 +118,9 @@ performance:
 Note that the appropriate values for those fields depend on the amount
 of free memory the database host has available.
 
+Additionally, admins of large deployments might want to consider using huge pages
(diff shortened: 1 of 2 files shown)


docs/postgres.md
```diff
@@ -118,6 +118,9 @@ performance:
 Note that the appropriate values for those fields depend on the amount
 of free memory the database host has available.
 
+Additionally, admins of large deployments might want to consider using huge pages
```
DMRobertson, line 121: "huge pages" is a linux kernel feature, so this only applies to deployments running on Linux. Maybe it's worth saying "large deployments on Linux".

(We don't really support running it on anything else AFAIK, so maybe this is a needless distinction to draw. I leave it to your discretion!)

DMRobertson (approved):
Thanks for writing this!

PR #11468, opened 2021-11-30
Clean-up get_version_string

clokep:
This cleans-up `get_version_string` a bit by factoring out common-code, fixing up some types and explicitly closing `/dev/null` (instead of assuming garbage collection will do it).

Note that this code is also shared by sydent so we might want to move it out at some point.

### DIFF
--- a/synapse/util/versionstring.py
+++ b/synapse/util/versionstring.py
@@ -65,9 +65,9 @@ def get_version_string(module: ModuleType) -> str:
             git_commit = _run_git_command("", "rev-parse", "--short", "HEAD")
 
             dirty_string = "-this_is_a_dirty_checkout"
-            is_dirty = _run_git_command("", "describe", "--dirty=" + dirty_string).endswith(
-                dirty_string
-            )
+            is_dirty = _run_git_command(
+                "", "describe", "--dirty=" + dirty_string
+            ).endswith(dirty_string)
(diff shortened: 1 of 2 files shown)


synapse/util/versionstring.py
```diff
@@ -44,71 +46,36 @@ def get_version_string(module: ModuleType) -> str:
...
     try:
-        null = open(os.devnull, "w")
-        cwd = os.path.dirname(os.path.abspath(module.__file__))
+        with open(os.devnull, "w") as null:
+            cwd = os.path.dirname(os.path.abspath(module.__file__))
+
+            def _run_git_command(prefix: str, *params: str) -> str:
+                try:
+                    result = (
+                        subprocess.check_output(["git", *params], stderr=null, cwd=cwd)
```
DMRobertson, line 55: It might be more idiomatic to use [`stderr=subprocess.DEVNULL`](https://docs.python.org/3/library/subprocess.html?highlight=subprocess#subprocess.DEVNULL)? I'd have to convince myself that it doesn't open and leak a file descriptor though. (I'd expect it doesn't, but...)

clokep: That might be better, I'll take a look! Thanks!

DMRobertson (approved):
LGTM!

PR #11475, opened 2021-12-01
Link background update controller docs to summary

babolivier:
Looks like I forgot to do that in #11306

### DIFF
--- a/changelog.d/11475.feature
+++ b/changelog.d/11475.feature
@@ -1,0 +1,1 @@
+Add plugin support for controlling database background updates.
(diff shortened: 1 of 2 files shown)


DMRobertson (approved):
Argh, sorry I missed this.

Might want to merge this directly into master after develop so it applies immediately on the docs website? (IIRC we build it from master, not develop)

babolivier:
> Might want to merge this directly into master after develop so it applies immediately on the docs website? (IIRC we build it from master, not develop)

https://matrix-org.github.io/synapse/develop builds off develop, https://matrix-org.github.io/synapse/latest builds off master. But it doesn't matter in this case since #11306 hasn't shipped yet (it's lined up for 1.49.0).

PR #11477, opened 2021-12-01
Import mock from unittest.

clokep:
These tests fail for me locally with `builtins.ModuleNotFoundError: No module named 'mock'`

I'm not sure why they're working in CI since we shouldn't be install `mock` anymore.

This also adds a license header. Broken from #11306.

### DIFF
--- a/changelog.d/11477.feature
+++ b/changelog.d/11477.feature
@@ -1,0 +1,1 @@
+Add plugin support for controlling database background updates.
--- a/tests/storage/test_background_update.py
+++ b/tests/storage/test_background_update.py
@@ -1,4 +1,18 @@
-from mock import Mock
+# Copyright 2021 The Matrix.org Foundation C.I.C.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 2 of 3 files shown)


erikjohnston:
> I think this is having some issues with async mocks. I don't understand how the original code remotely worked without `mock` installed though?

We added `mock` back in in #11306 as `mock` backports support for `AsyncMock`?

clokep:
> > I think this is having some issues with async mocks. I don't understand how the original code remotely worked without `mock` installed though?
> 
> We added `mock` back in in #11306 as `mock` backports support for `AsyncMock`?

That's really confusing -- we put a bunch of work into removing this dependency. 😢 

Anyway, it would be great to shout that dependencies updated!

erikjohnston:
Ah, sorry I didn't realise we'd done that :slightly_frowning_face:   TBH it was mostly me raging at the amount of faff to work around lack of async utils in py3.6

clokep:
For cross referencing -- see #9772.

clokep:
Oh hmm, I guess we should still grab the license header here. I'll do a separate PR.

PR #11478, opened 2021-12-01
Include bundled aggregations in /sync and related fixes

### LINKED ISSUE #11476: /sync is not bundling aggregations for limited rooms
[MSC2675](https://github.com/matrix-org/matrix-doc/pull/2675) says that `/sync` is one of the APIs which should bundle aggregations:

> `GET /sync`, only for room sections in the response where `limited` field is `true`; this amounts to all rooms in the response if the `since` request parameter was not passed, also known as an initial sync.

Currently it always disables bundled aggregations for this endpoint, see #5220.

We should also double check the other endpoints:

* [x] `GET /rooms/{roomId}/messages` -- [see code](https://github.com/matrix-org/synapse/blob/a265fbd397ae72b2d3ea4c9310591ff1d0f3e05c/synapse/handlers/pagination.py#L545-L558)
* [x] `GET /rooms/{roomId}/context` -- [see code](https://github.com/matrix-org/synapse/blob/a265fbd397ae72b2d3ea4c9310591ff1d0f3e05c/synapse/rest/client/room.py#L709-L723) (note that this doesn't bundle them for state events, unclear if this is "fine" or not)
* [x] `GET /rooms/{roomId}/event/{eventId}` -- [see code](https://github.com/matrix-org/synapse/blob/a265fbd397ae72b2d3ea4c9310591ff1d0f3e05c/synapse/rest/client/room.py#L664-L666)
* [ ] `GET /sync`
* [x] `GET /relations` -- see #11087

clokep:
This has a few fixes related to *when* we include bundled aggregations:

* Renames `bundle_relations` -> `bundle_aggreations` since this is what [MSC2675](https://github.com/matrix-org/matrix-doc/pull/2675) settled on. (Also update some of the surrounded code / comments.)
* Clarifies comments for when *not* to bundle aggregations.
* Bundles aggregations for a limited `/sync` (as specified in MSC2675).
* Does *not* bundle aggregations for `/initial_sync` (as allowed in the MSC2675).
* Consistently do not include bundled aggregations for state events -- this is specified in MSC2675.

Fixes #11476

### DIFF
--- a/changelog.d/11478.bugfix
+++ b/changelog.d/11478.bugfix
@@ -1,1 +1,1 @@
-Include bundled relations during a limited `/sync` request, per [MSC2675](https://github.com/matrix-org/matrix-doc/pull/2675).
+Include bundled relation aggregations during a limited `/sync` request, per [MSC2675](https://github.com/matrix-org/matrix-doc/pull/2675).
--- a/changelog.d/11478.bugfix
+++ b/changelog.d/11478.bugfix
@@ -1,0 +1,1 @@
+Include bundled relations during a limited `/sync` request, per [MSC2675](https://github.com/matrix-org/matrix-doc/pull/2675).
--- a/synapse/events/utils.py
+++ b/synapse/events/utils.py
@@ -306,6 +306,7 @@ def format_event_for_client_v2_without_room_id(d: JsonDict) -> JsonDict:
 def serialize_event(
     e: Union[JsonDict, EventBase],
     time_now_ms: int,
+    *,
     as_client_event: bool = True,
     event_format: Callable[[JsonDict], JsonDict] = format_event_for_client_v1,
     token_id: Optional[str] = None,
@@ -393,6 +394,7 @@ class EventClientSerializer:
         self,
         event: Union[JsonDict, EventBase],
         time_now: int,
+        *,
         bundle_aggregations: bool = True,
         **kwargs: Any,
     ) -> JsonDict:
--- a/synapse/events/utils.py
+++ b/synapse/events/utils.py
@@ -404,7 +404,8 @@ class EventClientSerializer:
             event: The event being serialized.
             time_now: The current time in milliseconds
             bundle_aggregations: Whether to include the bundled aggregations for this
-                event. Only applies to non-state events.
+                event. Only applies to non-state events. (State events never include
+                bundled aggregations.)
             **kwargs: Arguments to pass to `serialize_event`
 
         Returns:
--- a/synapse/events/utils.py
+++ b/synapse/events/utils.py
@@ -393,7 +393,7 @@ class EventClientSerializer:
         self,
         event: Union[JsonDict, EventBase],
         time_now: int,
-        bundle_relations: bool = True,
+        bundle_aggregations: bool = True,
         **kwargs: Any,
     ) -> JsonDict:
         """Serializes a single event.
@@ -401,7 +401,7 @@ class EventClientSerializer:
         Args:
             event: The event being serialized.
             time_now: The current time in milliseconds
-            bundle_relations: Whether to include the bundled relations for this
+            bundle_aggregations: Whether to include the bundled aggregations for this
                 event.
             **kwargs: Arguments to pass to `serialize_event`
 
@@ -416,18 +416,18 @@ class EventClientSerializer:
 
         # If MSC1849 is enabled then we need to look if there are any relations
         # we need to bundle in with the event.
-        # Do not bundle relations if the event has been redacted
+        # Do not bundle aggregations if the event has been redacted
         if not event.internal_metadata.is_redacted() and (
-            self._msc1849_enabled and bundle_relations
+            self._msc1849_enabled and bundle_aggregations
         ):
-            await self._injected_bundled_relations(event, time_now, serialized_event)
+            await self._injected_bundled_aggregations(event, time_now, serialized_event)
 
         return serialized_event
 
-    async def _injected_bundled_relations(
+    async def _injected_bundled_aggregations(
         self, event: EventBase, time_now: int, serialized_event: JsonDict
     ) -> None:
-        """Potentially injects bundled relations into the unsigned portion of the serialized event.
+        """Potentially injects bundled aggregations into the unsigned portion of the serialized event.
 
         Args:
             event: The event being serialized.
@@ -435,7 +435,7 @@ class EventClientSerializer:
             serialized_event: The serialized event which may be modified.
 
         """
-        # Do not bundle relations for an event which represents an edit or an
+        # Do not bundle aggregations for an event which represents an edit or an
         # annotation. It does not make sense for them to have related events.
         relates_to = event.content.get("m.relates_to")
         if isinstance(relates_to, (dict, frozendict)):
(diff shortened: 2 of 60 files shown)


synapse/handlers/initial_sync.py
```diff
@@ -165,7 +165,11 @@ async def handle_room(event: RoomsForUser) -> None:
 
                 invite_event = await self.store.get_event(event.event_id)
                 d["invite"] = await self._event_serializer.serialize_event(
-                    invite_event, time_now, as_client_event
+                    invite_event,
+                    time_now,
+                    # No need to bundle aggregations for state events.
+                    bundle_aggregations=False,
+                    as_client_event=as_client_event,
```
clokep, line 172: This potentially fixes an unrelated bug -- `as_client_event` was being passed in as `bundle_aggregations` instead of being passed through as a `kwarg` so `as_client_event` would always end up being `True` (the default).

synapse/handlers/initial_sync.py
```diff
@@ -226,6 +232,8 @@ async def handle_room(event: RoomsForUser) -> None:
                 d["state"] = await self._event_serializer.serialize_events(
                     current_state.values(),
                     time_now=time_now,
+                    # No need to bundle aggregations for state events.
+                    bundle_aggregations=False,
```
clokep, line 236: This a bunch of places where we do this -- we do a similar check in `serialize_event` (by checking `event.is_state()`, but I think it is worthwhile to keep it here since we *know* that these are state events it is more efficient to just pass the boolean. This might be over-optimizing though and I can remove these if it is clearer!

synapse/events/utils.py
```diff
@@ -416,26 +416,26 @@ async def serialize_event(
 
         # If MSC1849 is enabled then we need to look if there are any relations
         # we need to bundle in with the event.
```
richvdh, line 418:
```suggestion
        # If MSC1849 is enabled then we need to look if there are any relation
        # aggregations we need to bundle in with the event.

File: synapse/rest/client/sync.py [hunk 4]
```diff
@@ -544,8 +544,13 @@ def serialize(events: Iterable[EventBase]) -> Awaitable[List[JsonDict]]:
                     event.room_id,
                 )
 
-        serialized_state = await serialize(state_events)
-        serialized_timeline = await serialize(timeline_events)
+        # No need to bundle aggregations for state events.
+        serialized_state = await serialize(state_events, bundle_aggregations=False)
+        # Only bundle aggregations if the room is limited, as clients could be
+        # missing events.
```
richvdh, line 550: This comment is confusing. Why would bundling aggregations be a problem if clients were missing events?

[The MSC doesn't really explain why we only bundle aggregations if the timeline is limited, either.]

synapse/handlers/initial_sync.py
richvdh: ooof. maybe `bundle_relations` should be a kw-only arg to catch these problems earlier.

synapse/handlers/initial_sync.py
richvdh: I'm ambivalent

changelog.d/11478.bugfix
```diff
@@ -0,0 +1 @@
+Include bundled relations during a limited `/sync` request, per [MSC2675](https://github.com/matrix-org/matrix-doc/pull/2675).
```
richvdh, line 1:
```suggestion
Include bundled relation aggregations during a limited `/sync` request, per [MSC2675](https://github.com/matrix-org/matrix-doc/pull/2675).
```

synapse/handlers/initial_sync.py
richvdh: actually I'm a bit confused. You seem to have added a bunch of these, and then removed them again in a48bd2ad. What's that about?

richvdh (commented):
seems generally fine to me! A few nits and a couple of questions.

synapse/handlers/initial_sync.py
clokep: I think that makes sense. I can make that change.

synapse/handlers/initial_sync.py
clokep: Right, sorry should have explained that!

1. I initially added them to the places where we only serialize a state event.
2. I then added the check for `is_state()`.
3. I then removed passing the flag for places we were serializing a single event, but left it in places where we serialize many events.

As I said, maybe this is an over-optimization.

Do you think I should just leave them all in?

synapse/rest/client/sync.py
clokep: I believe the idea is that if the sync is limited we include the bundled aggregations so that the clients are able to display the event (with all reactions, etc.) without having to fetch the rest of the events.

synapse/handlers/initial_sync.py
clokep: I added this in 2f409daa8cb1834191bd5632ad3feab1584846df.

synapse/rest/client/sync.py
clokep: Maybe the following is clearer:

> Only bundle aggregations if the room is limited, otherwise the client should already have all the information necessary to render events.

Does that make sense?

synapse/handlers/initial_sync.py
richvdh: > Do you think I should just leave them all in?

I kinda do, to be honest. I think that would be clearer, and have little effect on performance.

I don't have super-strong feelings though.

richvdh: > I kinda do, to be honest.

wait no, the opposite. Let me try again.

I think you should let `serialize_event` sort it out, and just set `bundle_aggregations` where you need to override the behaviour. I think that would be clearer, and have little effect on performance.

synapse/rest/client/sync.py
richvdh: > > Only bundle aggregations if the room is limited, otherwise the client should already have all the information necessary to render events.
>
> Does that make sense?

Right yes, that makes more sense. Alternatively:

```suggestion
        # Don't bother to bundle aggregations if the timeline is unlimited, as
        # clients will have all the necessary information.
```

synapse/handlers/initial_sync.py
clokep: > I think you should let `serialize_event` sort it out, and just set `bundle_aggregations` where you need to override the behaviour. I think that would be clearer, and have little effect on performance.

OK. I think that's reasonable. I'll make sure that it is documented to not include them for state events too. 👍  Thanks!

clokep: Hopefully 2e76df3b023604f2b4e2d77eea895663d62ea35a makes this clear.

clokep:
Thank you for the review @richvdh! I think I've handled all your comments! 🎉

synapse/events/utils.py
```diff
@@ -404,7 +404,7 @@ async def serialize_event(
             event: The event being serialized.
             time_now: The current time in milliseconds
             bundle_aggregations: Whether to include the bundled aggregations for this
-                event.
+                event. Only applies to non-state events.
```
richvdh, line 407:
```suggestion
                event. Only applies to non-state events. (State events never include bundled aggregations.)
```

richvdh (approved):
lgtm!

clokep:
Thanks for the reviews! ❤️

PR #11479, opened 2021-12-01
Add a license header and comment.

clokep:
From #11306.


DMRobertson (approved):
Thanks for the fix.

(I'm starting to wonder if we ought to have a CI check that enforces a copyright notice for each file in `synapse` and `tests`?)

((IANAL, but given the repository has a licence, it seems a bit redundant to put it on each file. :shrug:))

PR #11482, opened 2021-12-01
Save the OIDC session ID (sid) with the device on login

sandhose:
This is part of #11326

### Pull Request Checklist

### DIFF
--- a/synapse/handlers/auth.py
+++ b/synapse/handlers/auth.py
@@ -183,10 +183,11 @@ class LoginTokenAttributes:
 
     user_id = attr.ib(type=str)
 
-    # the SSO Identity Provider that the user authenticated with, to get this token
     auth_provider_id = attr.ib(type=str)
+    """The SSO Identity Provider that the user authenticated with, to get this token."""
 
     oidc_sid = attr.ib(type=Optional[str])
+    """The OIDC session ID (sid) obtained from the ID token."""
 
 
 class AuthHandler:
--- a/synapse/handlers/auth.py
+++ b/synapse/handlers/auth.py
@@ -39,6 +39,7 @@ import attr
 import bcrypt
 import pymacaroons
 import unpaddedbase64
+from pymacaroons.exceptions import MacaroonVerificationFailedException
 
 from twisted.web.server import Request
 
--- a/synapse/handlers/auth.py
+++ b/synapse/handlers/auth.py
@@ -1670,7 +1670,7 @@ class AuthHandler:
                 during successful login. Must be JSON serializable.
             new_user: True if we should use wording appropriate to a user who has just
                 registered.
-            auth_provider_session_id: The session ID got during login from the SSO IdP.
+            auth_provider_session_id: The session ID from the SSO IdP received during login.
         """
         # If the account has been deactivated, do not proceed with the login
         # flow.
--- a/synapse/handlers/auth.py
+++ b/synapse/handlers/auth.py
@@ -186,8 +186,8 @@ class LoginTokenAttributes:
     auth_provider_id = attr.ib(type=str)
     """The SSO Identity Provider that the user authenticated with, to get this token."""
 
-    oidc_sid = attr.ib(type=Optional[str])
-    """The OIDC session ID (sid) obtained from the ID token."""
+    auth_provider_session_id = attr.ib(type=Optional[str])
+    """The session ID advertised by the SSO Identity Provider."""
 
 
 class AuthHandler:
@@ -1654,7 +1654,7 @@ class AuthHandler:
         client_redirect_url: str,
         extra_attributes: Optional[JsonDict] = None,
         new_user: bool = False,
-        oidc_sid: Optional[str] = None,
+        auth_provider_session_id: Optional[str] = None,
     ) -> None:
         """Having figured out a mxid for this user, complete the HTTP request
 
@@ -1670,7 +1670,7 @@ class AuthHandler:
                 during successful login. Must be JSON serializable.
             new_user: True if we should use wording appropriate to a user who has just
                 registered.
-            oidc_sid: The session ID (sid) got from a OIDC login.
+            auth_provider_session_id: The session ID got during login from the SSO IdP.
         """
         # If the account has been deactivated, do not proceed with the login
         # flow.
@@ -1691,7 +1691,7 @@ class AuthHandler:
             extra_attributes,
             new_user=new_user,
             user_profile_data=profile,
-            oidc_sid=oidc_sid,
+            auth_provider_session_id=auth_provider_session_id,
         )
 
     def _complete_sso_login(
@@ -1703,7 +1703,7 @@ class AuthHandler:
         extra_attributes: Optional[JsonDict] = None,
         new_user: bool = False,
         user_profile_data: Optional[ProfileInfo] = None,
-        oidc_sid: Optional[str] = None,
+        auth_provider_session_id: Optional[str] = None,
     ) -> None:
         """
         The synchronous portion of complete_sso_login.
@@ -1725,7 +1725,9 @@ class AuthHandler:
 
         # Create a login token
         login_token = self.macaroon_gen.generate_short_term_login_token(
-            registered_user_id, auth_provider_id=auth_provider_id, oidc_sid=oidc_sid
+            registered_user_id,
+            auth_provider_id=auth_provider_id,
+            auth_provider_session_id=auth_provider_session_id,
         )
 
         # Append the login token to the original redirect URL (i.e. with its query
@@ -1830,7 +1832,7 @@ class MacaroonGenerator:
         self,
         user_id: str,
         auth_provider_id: str,
-        oidc_sid: Optional[str] = None,
+        auth_provider_session_id: Optional[str] = None,
         duration_in_ms: int = (2 * 60 * 1000),
     ) -> str:
         macaroon = self._generate_base_macaroon(user_id)
@@ -1839,8 +1841,10 @@ class MacaroonGenerator:
         expiry = now + duration_in_ms
         macaroon.add_first_party_caveat("time < %d" % (expiry,))
         macaroon.add_first_party_caveat("auth_provider_id = %s" % (auth_provider_id,))
-        if oidc_sid is not None:
-            macaroon.add_first_party_caveat("oidc_sid = %s" % (oidc_sid,))
+        if auth_provider_session_id is not None:
+            macaroon.add_first_party_caveat(
+                "auth_provider_session_id = %s" % (auth_provider_session_id,)
+            )
         return macaroon.serialize()
 
     def verify_short_term_login_token(self, token: str) -> LoginTokenAttributes:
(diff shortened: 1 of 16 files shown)


synapse/handlers/auth.py
```diff
@@ -1851,15 +1860,24 @@ def verify_short_term_login_token(self, token: str) -> LoginTokenAttributes:
         user_id = get_value_from_macaroon(macaroon, "user_id")
         auth_provider_id = get_value_from_macaroon(macaroon, "auth_provider_id")
 
+        oidc_sid: Optional[str] = None
+        try:
+            oidc_sid = get_value_from_macaroon(macaroon, "oidc_sid")
+        except MacaroonVerificationFailedException:
```
clokep, line 1866: This seems to be undefined -- missing import probably from splitting the PR?

synapse/handlers/auth.py
```diff
@@ -1851,15 +1860,24 @@ def verify_short_term_login_token(self, token: str) -> LoginTokenAttributes:
...
+            pass
+
         v = pymacaroons.Verifier()
         v.satisfy_exact("gen = 1")
         v.satisfy_exact("type = login")
         v.satisfy_general(lambda c: c.startswith("user_id = "))
         v.satisfy_general(lambda c: c.startswith("auth_provider_id = "))
+        v.satisfy_general(lambda c: c.startswith("oidc_sid = "))
```
clokep, line 1874: If I'm remembering how macaroons work -- this will be fine if there's no `oidc_sid` also?

synapse/handlers/auth.py
sandhose: Yep, definitely. Not around my computer anymore, so feel free to push a fix on my branch, else I'll do that tomorrow

synapse/storage/databases/main/devices.py
```diff
@@ -1079,6 +1102,8 @@ async def store_device(
             device_id: id of device
             initial_device_display_name: initial displayname of the device.
                 Ignored if device exists.
+            auth_provider_id: The SSO IdP the user used, if any.
+            oidc_sid: The session ID (sid) got from a OIDC login.
```
clokep, line 1106: What happens if these differ? (Can the same device end up logging it with a different OIDC session?)

clokep (commented):
Looks somewhat reasonable, seems to mostly just be threading the session ID through many methods. There's a few things I'm unsure about though, please see comments!

synapse/storage/databases/main/devices.py
sandhose: Good question. I wonder in what scenario this could happen though?
It would be more accurate if the oidc_sid & auth_provider_id were linked to the access token, but that means they have to be carried on when refreshing the access token, so I linked them to the device instead.

synapse/handlers/oidc.py
```diff
@@ -684,9 +681,6 @@ async def _parse_id_token(self, token: Token, nonce: str) -> UserInfo:
             # If we got an `access_token`, there should be an `at_hash` claim
             # in the `id_token` that we can check against.
             claims_params["access_token"] = token["access_token"]
-            claims_cls = CodeIDToken
-        else:
-            claims_cls = ImplicitIDToken
```
sandhose, line 689: Looking again at what `CodeIDToken` and `ImplicitIDToken` verify in `authlib`, it really did not make sense to have that logic here.

synapse/handlers/auth.py
sandhose: Fixed

synapse/handlers/auth.py
```diff
@@ -1665,6 +1668,7 @@ async def complete_sso_login(
                 during successful login. Must be JSON serializable.
             new_user: True if we should use wording appropriate to a user who has just
                 registered.
+            oidc_sid: The session ID (sid) got from a OIDC login.
```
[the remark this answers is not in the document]
sandhose: Erf, not sure about that. Sure, there is a similar thing in CAS & SAML2, but I'd rather rename that parameter & column if we end up doing the same thing with those if that's OK

synapse/handlers/auth.py
```diff
@@ -185,6 +185,8 @@ class LoginTokenAttributes:
     # the SSO Identity Provider that the user authenticated with, to get this token
     auth_provider_id = attr.ib(type=str)
 
+    oidc_sid = attr.ib(type=Optional[str])
```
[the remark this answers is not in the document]
sandhose: Done

synapse/handlers/auth.py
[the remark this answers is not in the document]
clokep: If I'm recalling correctly, renaming database columns isn't a trivial thing so I'd like use to get it right up front! I'll have a think about this though...

synapse/storage/databases/main/devices.py
clokep: Maybe if soft-logouts are enabled and the user has to login again on the same device? Would that trigger this?

sandhose: Do we have evidence that clients do that?
I see how it could happen with API calls, but not really in the wild with clients.
The other way around is more plausible (multiple devices/clients connected with the same OIDC session) and is supported by this

synapse/handlers/auth.py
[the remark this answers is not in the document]
clokep: Confirmed at https://matrix.to/#/!XaqDhxuTIlvldquJaV:matrix.org/$8kjA_AdB32SFwZ_dstHYLedVCN6YDZhrHIzb5zRxyWI?via=matrix.org&via=vector.modular.im&via=envs.net that renaming columns on sqlite doesn't work great, at least not for the versions we support (you have to rewrite the table).

sandhose:
@clokep Ready for another review!

I ended up having the device <-> auth session association in a separate table to avoid potentials issues when reusing a device. I also renamed `oidc_sid` to `auth_provider_session_id` (to go with `auth_provider_id`)

synapse/storage/databases/main/devices.py
```diff
@@ -139,6 +139,27 @@ async def get_devices_by_user(self, user_id: str) -> Dict[str, Dict[str, str]]:
 
         return {d["device_id"]: d for d in devices}
 
+    async def get_devices_by_auth_provider_session_id(
+        self, auth_provider_id: str, auth_provider_session_id: str
+    ) -> List[Dict[str, Any]]:
+        """Retrieve the list of devices associated with an OIDC session ID (sid).
```
clokep, line 145:
```suggestion
        """Retrieve the list of devices associated with a SSO IdP session ID.
```

(This method seems unused here, but I think it's use is fairly obvious if we're saving this into the database...)

synapse/storage/databases/main/devices.py
```diff
@@ -1115,6 +1143,18 @@ async def store_device(
                 if hidden:
                     raise StoreError(400, "The device ID is in use", Codes.FORBIDDEN)
 
+            if auth_provider_id and auth_provider_session_id:
```
clokep, line 1146: I don't know if it matters if this should be in a transaction with the `simple_upsert` above or not?

synapse/storage/schema/main/delta/65/11_devices_auth_provider_session.sql
```diff
@@ -0,0 +1,27 @@
...
+  user_id TEXT NOT NULL,
+  device_id TEXT NOT NULL,
+  auth_provider_id TEXT NOT NULL,
+  auth_provider_session_id TEXT NOT NULL
+);
+
+CREATE INDEX device_auth_providers_devices
+  ON device_auth_providers (user_id, device_id);
+CREATE INDEX device_auth_providers_sessions
+  ON device_auth_providers (auth_provider_id, auth_provider_session_id);
```
clokep, line 27: We're considing that each user + device can only appear once and that each auth provider can only use a session ID once?

Do sessions ever need to time out or anything? It looks like this table will grow indefinitely at the moment.

synapse/handlers/oidc.py
sandhose: The only things it changes is the verification done on claims in the ID token. `CodeIDToken` is for the flow we're using, `ImplicitIDToken` is for the implicit flow (where you get the access token/ID token directly in the callback params).

synapse/storage/databases/main/devices.py
sandhose: Yep, and definitely used in the OIDC logout PR afterwards

synapse/storage/databases/main/devices.py
sandhose: there are already multiple things done here not in a transaction, that's why I did not bother

synapse/storage/schema/main/delta/65/11_devices_auth_provider_session.sql
sandhose: Nope, you can have multiple logins with the same session ID, and because of it being technically possible to have different session for the same device, I moved this to this separate table

synapse/handlers/oidc.py
```diff
@@ -117,7 +117,8 @@ async def load_metadata(self) -> None:
         for idp_id, p in self._providers.items():
             try:
                 await p.load_metadata()
-                await p.load_jwks()
+                if not p._uses_userinfo:
+                    await p.load_jwks()
```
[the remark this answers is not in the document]
sandhose: Problem is that we are raising if we call load_jwks and there is no jwks_uri in the OP metadata (which can happen)
Before, we were just returning an empty key set if the userinfo endpoint was used, which was fine because we never tried to verify ID tokens in that case.
Now, we verify ID tokens if they are present, even if we are using the userinfo endpoint, so we can't just return an empty set, but we still want to raise if we try to verify an ID token while there is no jwks_uri 
This here is just to crash if Synapse can't figure out the OP metadata & keyset on startup. We don't want it to crash if the OP does not have keys (like GitHub) because they are using the userinfo endpoint instead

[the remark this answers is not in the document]
clokep: Thanks!

synapse/storage/databases/main/devices.py
clokep: I figured that was the rationale. 👍

synapse/storage/schema/main/delta/65/11_devices_auth_provider_session.sql
clokep: Sorry -- I misread these as unique indexes. 👍

synapse/handlers/sso.py
```diff
@@ -415,6 +416,8 @@ async def complete_sso_login_request(
             extra_login_attributes: An optional dictionary of extra
                 attributes to be provided to the client in the login response.
 
+            auth_provider_session_id: An optional session ID from the OIDC login
```
clokep, line 419:
```suggestion
            auth_provider_session_id: An optional session ID from the IdP.
```

clokep:
Should we augment `test_oidc` to have a situation where we *do* save the session ID?

sandhose:
> Should we augment `test_oidc` to have a situation where we _do_ save the session ID?

Done!

clokep (approved):

PR #11483, opened 2021-12-01
Add most of the missing type hints to `synapse.federation`.

clokep:
This adds most of the missing type hints to `synapse.federation`, but skips a few methods which are dicey. I think those can be done in a follow-up since this PR became decent sized and it is better to get the partial type hints in then to have none at all.

### DIFF
--- a/synapse/federation/transport/server/_base.py
+++ b/synapse/federation/transport/server/_base.py
@@ -239,9 +239,7 @@ class BaseFederationServlet:
         self.ratelimiter = ratelimiter
         self.server_name = server_name
 
-    def _wrap(
-        self, func: Callable[..., Awaitable[Tuple[int, JsonDict]]]
-    ) -> ServletCallback:
+    def _wrap(self, func: Callable[..., Awaitable[Tuple[int, Any]]]) -> ServletCallback:
         authenticator = self.authenticator
         ratelimiter = self.ratelimiter
 
--- a/synapse/federation/transport/server/_base.py
+++ b/synapse/federation/transport/server/_base.py
@@ -15,11 +15,11 @@
 import functools
 import logging
 import re
-from typing import Optional, Tuple
+from typing import Any, Awaitable, Callable, Optional, Tuple, cast
 
 from synapse.api.errors import Codes, FederationDeniedError, SynapseError
 from synapse.api.urls import FEDERATION_V1_PREFIX
-from synapse.http.server import HttpServer
+from synapse.http.server import HttpServer, ServletCallback
 from synapse.http.servlet import parse_json_object_from_request
 from synapse.http.site import SynapseRequest
 from synapse.logging import opentracing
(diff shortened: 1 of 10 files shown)


synapse/federation/transport/server/_base.py
```diff
@@ -233,31 +239,37 @@ def __init__(
         self.ratelimiter = ratelimiter
         self.server_name = server_name
 
-    def _wrap(self, func):
+    def _wrap(
+        self, func: Callable[..., Awaitable[Tuple[int, JsonDict]]]
```
squahtx, line 243: As much as I dislike `Any`, I think the return type of `func` is `Awaitable[Tuple[int, Any]]` since there are a bunch of handlers in `synapse/federation/transport/server/federation.py` that return something other than a `JsonDict`.

synapse/federation/transport/server/_base.py
```diff
@@ -301,7 +313,7 @@ async def new_func(request, *args, **kwargs):
                                 "client disconnected before we started processing "
                                 "request"
                             )
-                            return -1, None
+                            return None
```
squahtx, line 316: Will this change have any unwanted side effects?

I'm failing to find where the return of this function is consumed...

clokep: Oh yeah this goes through many many layers, I should have deep linked. Sorry about that!

This essentially ends up being called at:

https://github.com/matrix-org/synapse/blob/b64b6d12d47c012a9996a2585a969d3b962d25fa/synapse/http/server.py#L262-L266

So by returning `None` we're short-circuiting calling `self._send_response`. For the case of the federation resources, they're registered as part of the [`TransportLayerServer`](https://github.com/matrix-org/synapse/blob/a6f1a3abecf8e8fd3e1bff439a06b853df18f194/synapse/federation/transport/server/__init__.py#L46) class, which is a sub-class of [`JsonResource`](https://github.com/matrix-org/synapse/blob/b64b6d12d47c012a9996a2585a969d3b962d25fa/synapse/http/server.py#L355), which is a sub-class of `DirectServeJsonResource` which essentially calls out to `response_with_json`:

https://github.com/matrix-org/synapse/blob/b64b6d12d47c012a9996a2585a969d3b962d25fa/synapse/http/server.py#L325-L339

If we investigate `response_with_json` we'll see that this also inspects `request._disconnected` and bails:

https://github.com/matrix-org/synapse/blob/b64b6d12d47c012a9996a2585a969d3b962d25fa/synapse/http/server.py#L667-L674

Pretty much, the callbacks already handle returning `None` to say "I'm done", but for some reason the federation code does something different and relies on us separate check to give up. I'm making this a bit more consistent with the C-S code. I think.

Let me know if that isn't clear!

synapse/federation/transport/server/_base.py
clokep: It really needs to be "something that can be serialized to JSON" I think... which we don't have a good type for. 😢 I can make it `Any` for now and we can improve if we'd like.

(I think the only things returned are `JsonDict` or `Tuple[int, JsonDict]` but there could be other weirder things.)

squahtx: Sounds good, I'd avoid trying to define Jsonable/JsonObject for now, to avoid scope creep.

clokep: Oh, this seems to better match the return type of `_wrap` anyway. 👍

synapse/federation/transport/server/_base.py
squahtx: Thanks for clearing that up!

squahtx (approved):
LGTM!

PR #11486, opened 2021-12-01
`scripts-dev/sign_json`: support for signing events

richvdh:


scripts-dev/federation_client.py
```diff
@@ -15,6 +15,25 @@
...
+    user_id=...
+
+   # make_join
+   federation_client.py "/_matrix/federation/v1/make_join/$room_id/$user_id?ver=5" > make_join.json
+
+   # sign
+   jq -M .event make_join.json | sign_json --sign-event-room-version=$(jq -r .room_version make_join.json) -o signed-join.json
+
+   # send_join
+   federation_client.py -X PUT "/_matrix/federation/v2/send_join/$room_id/x" --body $(<signed-join.json) > send_join.json
```
clokep, line 34: Looks like this switched from 4-space to 3-space indent randomly?

```suggestion
    # make_join
    federation_client.py "/_matrix/federation/v1/make_join/$room_id/$user_id?ver=5" > make_join.json

    # sign
    jq -M .event make_join.json | sign_json --sign-event-room-version=$(jq -r .room_version make_join.json) -o signed-join.json

    # send_join
    federation_client.py -X PUT "/_matrix/federation/v2/send_join/$room_id/x" --body $(<signed-join.json) > send_join.json
```

clokep: Is the `/x` at the end of the `send_join` call correct?

clokep (approved):
Seems reasonable, wasn't sure about one thing though.

scripts-dev/federation_client.py
richvdh: > Is the /x at the end of the send_join call correct?

Well, it's supposed to be the event id of the join event. But Synapse ignores it (https://github.com/matrix-org/synapse/blob/develop/synapse/federation/transport/server/federation.py#L396-L397), and calculating the correct event id is fiddly.

So, it's deliberate, at least :)

clokep: Ah, right! :) Thanks!

PR #11487, opened 2021-12-02
Test to ensure we share the same `state_group` across the whole historical batch (MSC2716)

MadLittleMods:
Test to ensure we share the same `state_group` across the whole historical batch.

We did some work on making sure the `state_groups` were shared in https://github.com/matrix-org/synapse/pull/10975 and I don't want to regress.

It also helps me test the details of this statement: https://github.com/matrix-org/synapse/pull/11243/files#discussion_r750819703

### Dev notes

```
SYNAPSE_POSTGRES=1 SYNAPSE_TEST_LOG_LEVEL=INFO python -m twisted.trial tests.rest.client.test_room_batch
```

### Pull Request Checklist

### DIFF
--- a/tests/rest/client/test_room_batch.py
+++ b/tests/rest/client/test_room_batch.py
@@ -88,8 +88,6 @@ class RoomBatchTestCase(unittest.HomeserverTestCase):
     def prepare(self, reactor: MemoryReactor, clock: Clock, hs: HomeServer) -> None:
         self.clock = clock
         self.storage = hs.get_storage()
-        # self.user_id = self.register_user("tester", "foobar")
-        # self.access_token = self.login("tester", "foobar")
 
         self.virtual_user_id = self.register_appservice_user(
             "as_user_potato", self.appservice.token
--- a/tests/rest/client/test_room_batch.py
+++ b/tests/rest/client/test_room_batch.py
@@ -1,0 +1,189 @@
+import logging
+
+from synapse.rest import admin
+from synapse.rest.client import login, register, room, room_batch
+from synapse.api.constants import (
+    EventContentFields,
+    EventTypes,
+)
+from synapse.appservice import ApplicationService
+from synapse.server import HomeServer
+from synapse.types import JsonDict
+from synapse.util import Clock
+
+from twisted.test.proto_helpers import MemoryReactor
+
+from typing import (
+    List,
+    Tuple,
+)
+from tests import unittest
+from unittest.mock import Mock, patch
+
+logger = logging.getLogger(__name__)
+
+
+def _create_join_state_events_for_batch_send_request(
+    virtual_user_ids: List[str],
+    insert_time: int,
+) -> List[JsonDict]:
+    return [
+        {
+            "type": EventTypes.Member,
+            "sender": virtual_user_id,
+            "origin_server_ts": insert_time,
+            "content": {
+                "membership": "join",
+                "displayname": "display-name-for-%s" % (virtual_user_id,),
+            },
+            "state_key": virtual_user_id,
+        }
+        for virtual_user_id in virtual_user_ids
+    ]
+
+
+def _create_message_events_for_batch_send_request(
+    virtual_user_id: str, insert_time: int, count: int
+) -> List[JsonDict]:
+    return [
+        {
+            "type": EventTypes.Message,
+            "sender": virtual_user_id,
+            "origin_server_ts": insert_time,
+            "content": {
+                "msgtype": "m.text",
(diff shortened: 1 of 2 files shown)


tests/rest/client/test_room_batch.py
```diff
@@ -0,0 +1,182 @@
...
+            mock_load_appservices,
+        ):
+            hs = self.setup_test_homeserver(config=config)
+        return hs
+
+    def prepare(self, reactor: MemoryReactor, clock: Clock, hs: HomeServer) -> None:
+        self.clock = clock
+        self.storage = hs.get_storage()
+        # self.user_id = self.register_user("tester", "foobar")
+        # self.access_token = self.login("tester", "foobar")
```
DMRobertson, line 92: Do we want to keep these lines around?

tests/rest/client/test_room_batch.py
```diff
@@ -0,0 +1,182 @@
...
+
+        self.virtual_user_id = self.register_appservice_user(
+            "as_user_potato", self.appservice.token
```
DMRobertson, line 95: Hah, is this copied from the user directory stuff I did?

tests/rest/client/test_room_batch.py
```diff
@@ -0,0 +1,182 @@
...
+
+        return room_id, event_id_a, event_id_b, event_id_c
+
+    @unittest.override_config({"experimental_features": {"msc2716_enabled": True}})
+    def test_same_state_groups_for_whole_historical_batch(self):
+        """Make sure that when using the `/batch_send` endpoint to import a
+        bunch of historical messages, it re-uses the same `state_group` across
+        the whole batch. This is an easy optimization to make sure we're getting
+        right because the state for the whole batch is contained in
+        `state_events_at_start` and can be shared across everything.
```
DMRobertson, line 144: I think this only makes sense of events in historical batches cannot alter the state of a room. Is that the case?

tests/rest/client/test_room_batch.py
```diff
@@ -0,0 +1,182 @@
...
+
+        # Fetch the state_groups
+        state_group_map = self.get_success(
+            self.storage.state.get_state_groups_ids(room_id, historical_event_ids)
+        )
+
+        # We expect all of the historical events to be using the same state_group
+        # so there should only be a single state_group here!
+        self.assertEqual(
+            len(state_group_map.keys()),
```
DMRobertson, line 178: Assuming `state_group_map` is a dict here, you could just use `len(state_group_map)` if you prefer.

DMRobertson (commented):
Seems reasonable! A few minor ~~things~~ questions:

tests/rest/client/test_room_batch.py
MadLittleMods: Correct, the `state_events_at_start` do not affect the current room state at all.

tests/rest/client/test_room_batch.py
MadLittleMods: I think so 😁

tests/rest/client/test_room_batch.py
MadLittleMods: I prefer the keys one just because it's more explicit and can tell easier what's going on to my newer Python eyes. Feel free to poke the other way again if you rather have it that way.

Just did a quick search and I see it used the way I have it in other places in the code. Probably a mix of the two throughout.

tests/rest/client/test_room_batch.py
DMRobertson: How would we handle importing a conversation that includes e.g. a display name change? Normally that'd be a "m.room.membership" state change event. I guess there'd need to be two separate batches: one before the name change, and one after?

DMRobertson: (Sorry, appreciate this is a wider question about the MSC rather than this PR as such. Just trying to get the context)

tests/rest/client/test_room_batch.py
DMRobertson: That's fine---just wanted to flag up the option!

(I for one am always surprised that iterating over a dict yields keys, not (key, value) pairs. :shrug: )

tests/rest/client/test_room_batch.py
MadLittleMods: >  I guess there'd need to be two separate batches: one before the name change, and one after?

Your assumption is correct. State is per batch.

DMRobertson (approved):

MadLittleMods:
Thanks for the review @DMRobertson 🐻

PR #11488, opened 2021-12-02
Add type annotations to `tests.storage.test_appservice`.

reivilibre:
I intend to do another pass through to convert it to not use inlineCallbacks if possible.
Need to find out why it doesn't use HomeserverTestCase; perhaps because it's old?

### DIFF
--- a/synapse/appservice/__init__.py
+++ b/synapse/appservice/__init__.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 import logging
 import re
+from enum import Enum
 from typing import TYPE_CHECKING, Iterable, List, Match, Optional
 
 from synapse.api.constants import EventTypes
@@ -27,7 +28,7 @@ if TYPE_CHECKING:
 logger = logging.getLogger(__name__)
 
 
-class ApplicationServiceState:
+class ApplicationServiceState(Enum):
     DOWN = "down"
     UP = "up"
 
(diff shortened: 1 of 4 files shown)


synapse/appservice/__init__.py
```diff
@@ -27,7 +28,7 @@
 logger = logging.getLogger(__name__)
 
 
-class ApplicationServiceState:
+class ApplicationServiceState(Enum):
```
squahtx, line 31: Silly question, but this change isn't visible to the module API right?

reivilibre: not a silly question! As far as I can tell, no it's not (and I can't think of any interfaces where it'd make sense). :)

synapse/storage/databases/main/appservice.py
```diff
@@ -173,7 +173,7 @@ async def get_appservice_state(
             desc="get_appservice_state",
         )
         if result:
-            return result.get("state")
+            return ApplicationServiceState(result.get("state"))
```
erikjohnston, line 176: Just a small note to say that this changed the behaviour of the function. Previously this function returned `None` if `state` was `null` in the DB, but with this change the function errors instead.

This is fixed in https://github.com/matrix-org/synapse/pull/12680#discussion_r874515829

PR #11495, opened 2021-12-02
Clean up `tests.test_visibility` to remove legacy code.

reivilibre:


tests/test_visibility.py
```diff
@@ -292,7 +289,8 @@ def test_large_room(self):
             self.assertEqual(events_to_filter[i].event_id, filtered[i].event_id)
             self.assertEqual(filtered[i].content["extra"], "zzz")
 
-    test_large_room.skip = "Disabled by default because it's slow"
+    # (Mypy is not happy about a function having a field, but we ignore it because it's fine here.)
+    test_large_room.skip = "Disabled by default because it's slow"  # type: ignore[attr-defined]
```
squahtx, line 293: Why don't we use the `@unittest.skip` decorator (in general)?

tests/test_visibility.py
```diff
@@ -185,17 +179,15 @@ def inject_message(self, user_id, content=None):
...
 
-        yield defer.ensureDeferred(
-            self.storage.persistence.persist_event(event, context)
-        )
+        self.get_success(self.storage.persistence.persist_event(event, context))
         return event
 
-    @defer.inlineCallbacks
-    def test_large_room(self):
+    # FIXME This test is broken!
```
squahtx, line 189: Perhaps this should be a second skip reason?

tests/test_visibility.py
```diff
@@ -301,22 +299,26 @@ class _TestStore:
...
         # data for get_events: a map from event_id to event
-        self.events = {}
+        self.events: Dict[str, EventBase] = {}
 
         # data for get_state_ids_for_events mock: a map from event_id to
-        # a map from (type_state_key) -> event_id for the state at that
+        # a map from (type, state_key) -> event_id for the state at that
         # event
-        self.state_ids_for_events = {}
+        self.state_ids_for_events: Dict[str, Dict[Tuple[str, Optional[str]], str]] = {}
```
squahtx, line 309: Is `(Mutable)StateMap[str]` appropriate for the types in this file?

Also does the `Optional[str]` need to be optional?

tests/test_visibility.py
reivilibre: I don't know. Personally speaking I don't know how to skip tests, but also wouldn't write a test that I'd skip by default anyway :).

reivilibre:
Sorry to pull the rug from under your feet, @squahtx but I just removed the `test_large_room` test because it is clearly using APIs that have changed a long time ago, suggesting that nobody has cared enough to run it in so long that there's little point modernising it now.

squahtx (approved):
lgtm! less code makes it easier to review

Issue #11496, opened 2021-12-02
Drop `state_events` and `rejections` tables


richvdh:
Both `state_events` and `rejections` are tables which add a single piece of data to a subset of events (`state_key` and `reason`, respectively).

I believe this would better handled with nullable columns in the `events` table: 
  * `state_events` and `rejections` are *never* read from without a join to `events`, so we can save a join whereever we read from them
  * it's not like adding the data to `events` is going to transform it from a small, cacheable table to something huge (so I don't expect much difference in the case where we *don't* join to `rejections` or `state_events`). In any case, when we look in `events` we normally want the rejection status too.
  * obviously, populating three tables (and maintaining two extra indexes) is extra work to do when storing events.

However, getting from here to there is not an easy task by any means, and will take quite a bit of work to do in a way that doesn't break backwards compatibility. Here are the tasks as far as I can see them:

 * [ ] disambiguate any references to the `state_key` column, in preparation for adding new columns to `events`.
 * [ ] Add `state_key` and `rejection_reason` columns to `events`, and populate them for all new events.
 * [ ] Add a background update to populate `state_key` and `rejection_reason` for all existing events.
 * [ ] Add some code which determines if the background update has completed, and if so, sets a flag which makes reads use the new columns instead of joining to the old tables.
 * [ ] Add some more code which drops the old tables and stops populating them.

Each of those steps will require SCHEMA_VERSION and/or SCHEMA_COMPAT version updates to ensure that we never break compatibility with the previous Synapse release.

richvdh:
It's no longer entirely clear that putting `rejections` into the `events` table is a good idea... it is possible for events to change rejection status (thanks to faster-joins) and in particular, different servers may have different perspectives on whether an event has been rejected or not. The implication is that, if we were to move to a model where a single event store could back multiple server deployments, the rejection status would have to go and live somewhere else.

PR #11497, opened 2021-12-02
Disambiguate queries on `state_key`

richvdh:
This is part of my long-held secret scheme to kill off the `state_events` and `rejections` tables: see https://github.com/matrix-org/synapse/issues/11496.

The first step is just some column disambiguation.


synapse/storage/databases/main/events.py
```diff
@@ -575,9 +575,9 @@ def _add_chain_cover_index(
         # fetch their auth event info.
         while missing_auth_chains:
             sql = """
-                SELECT event_id, events.type, state_key, chain_id, sequence_number
+                SELECT event_id, events.type, se.state_key, chain_id, sequence_number
                 FROM events
-                INNER JOIN state_events USING (event_id)
+                INNER JOIN state_events se USING (event_id)
```
reivilibre, line 580:
```suggestion
                INNER JOIN state_events AS se USING (event_id)
```

synapse/storage/schema/__init__.py
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-SCHEMA_VERSION = 65  # remember to update the list below when updating
+SCHEMA_VERSION = 66  # remember to update the list below when updating
```
richvdh, line 15: this is needed, because adding the new columns will break compatibility with synapse versions before this PR, so we'll need to bump SCHEMA_COMPAT_VERSION.

synapse/storage/databases/main/events_worker.py
```diff
@@ -1408,10 +1408,10 @@ def get_all_new_forward_event_rows(
         ) -> List[Tuple[int, str, str, str, str, str, str, str, str]]:
             sql = (
                 "SELECT e.stream_ordering, e.event_id, e.room_id, e.type,"
-                " state_key, redacts, relates_to_id, membership, rejections.reason IS NOT NULL"
+                " se.state_key, redacts, relates_to_id, membership, rejections.reason IS NOT NULL"
                 " FROM events AS e"
                 " LEFT JOIN redactions USING (event_id)"
-                " LEFT JOIN state_events USING (event_id)"
+                " LEFT JOIN state_events se USING (event_id)"
```
reivilibre, line 1414:
```suggestion
                " LEFT JOIN state_events AS se USING (event_id)"
```
(personal preference, though it is also used above)

synapse/storage/databases/main/events_worker.py
```diff
@@ -1449,11 +1449,11 @@ def get_ex_outlier_stream_rows_txn(
         ) -> List[Tuple[int, str, str, str, str, str, str, str, str]]:
             sql = (
                 "SELECT event_stream_ordering, e.event_id, e.room_id, e.type,"
-                " state_key, redacts, relates_to_id, membership, rejections.reason IS NOT NULL"
+                " se.state_key, redacts, relates_to_id, membership, rejections.reason IS NOT NULL"
                 " FROM events AS e"
                 " INNER JOIN ex_outlier_stream AS out USING (event_id)"
                 " LEFT JOIN redactions USING (event_id)"
-                " LEFT JOIN state_events USING (event_id)"
+                " LEFT JOIN state_events se USING (event_id)"
```
reivilibre, line 1456:
```suggestion
                " LEFT JOIN state_events AS se USING (event_id)"
```

synapse/storage/databases/main/events_worker.py
```diff
@@ -1507,10 +1507,10 @@ def get_all_new_backfill_event_rows(
         ) -> Tuple[List[Tuple[int, Tuple[str, str, str, str, str, str]]], int, bool]:
             sql = (
                 "SELECT -e.stream_ordering, e.event_id, e.room_id, e.type,"
-                " state_key, redacts, relates_to_id"
+                " se.state_key, redacts, relates_to_id"
                 " FROM events AS e"
                 " LEFT JOIN redactions USING (event_id)"
-                " LEFT JOIN state_events USING (event_id)"
+                " LEFT JOIN state_events se USING (event_id)"
```
reivilibre, line 1513:
```suggestion
                " LEFT JOIN state_events AS se USING (event_id)"
```

synapse/storage/databases/main/events_worker.py
```diff
@@ -1537,11 +1537,11 @@ def get_all_new_backfill_event_rows(
 
             sql = (
                 "SELECT -event_stream_ordering, e.event_id, e.room_id, e.type,"
-                " state_key, redacts, relates_to_id"
+                " se.state_key, redacts, relates_to_id"
                 " FROM events AS e"
                 " INNER JOIN ex_outlier_stream AS out USING (event_id)"
                 " LEFT JOIN redactions USING (event_id)"
-                " LEFT JOIN state_events USING (event_id)"
+                " LEFT JOIN state_events se USING (event_id)"
```
reivilibre, line 1544:
```suggestion
                " LEFT JOIN state_events AS se USING (event_id)"
```

reivilibre (approved):
Seems sensible to me, though I haven't checked that this is exhaustive. I suppose the tests will do that when you create an ambiguity though?

synapse/storage/schema/__init__.py
clokep, line 15: I'm having trouble convincing myself this needs to be bumped? It looks like this is essentially a no-op and should be backwards and forwards compatible?

Or are we just being conservative and bumping it anyway?

reivilibre: I believe the issue is that Rich wants to add a `state_key` column to `events`, at which point a rollback becomes impossible because prior versions of Synapse will have these ambiguous queries left in.

clokep:
I think this is missing one:

https://github.com/matrix-org/synapse/blob/2b82ec425fccb0ef626242779f7ccd4d77a0685c/synapse/storage/databases/main/event_federation.py#L1551-L1559

synapse/storage/schema/__init__.py
richvdh: @clokep did you see my comments at https://github.com/matrix-org/synapse/pull/11497#discussion_r761353811?

clokep: Ahhhh, yeah I wasn't thinking of rolling forward and then backwards again. 👍

clokep: > @clokep did you see my comments at https://github.com/matrix-org/synapse/pull/11497#discussion_r761353811?

I have now. :)

richvdh:
> I think this is missing one:

great spot! I've just been through and double-checked again, and I can't find any more.

clokep:
Thanks for double checking! I ran `pygrep -r "JOIN state_events" synapse` and matched with the changes you had made. (That picked up a couple in `synapse/storage/databases/main/events_bg_updates.py` too, but those already looked to be OK!)

PR #11503, opened 2021-12-03
Refactor `tests.util.setup_test_homeserver` and `tests.server.setup_test_homeserver`.

reivilibre:

Only `tests.server.setup_test_homeserver` is called by anything at the moment
(all users of the one from `tests.util` have been *removed* in 3 prior PRs)

### DIFF
--- a/changelog.d/11503.misc
+++ b/changelog.d/11503.misc
@@ -1,0 +1,1 @@
+Refactor `tests.util.setup_test_homeserver` and `tests.server.setup_test_homeserver`.
--- a/tests/server.py
+++ b/tests/server.py
@@ -467,7 +467,7 @@ class ThreadPool:
         return d
 
 
-def make_test_homeserver_synchronous(server: HomeServer) -> None:
+def _make_test_homeserver_synchronous(server: HomeServer) -> None:
     """
     Make the given test homeserver's database interactions synchronous.
     """
@@ -851,6 +851,6 @@ def setup_test_homeserver(
     hs.get_auth_handler().validate_hash = validate_hash
 
     # Make the threadpool and database transactions synchronous for testing.
-    make_test_homeserver_synchronous(hs)
+    _make_test_homeserver_synchronous(hs)
 
     return hs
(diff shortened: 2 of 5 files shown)


clokep (commented):
I'm finding this hard to follow, I think the following operations are being done:

* Rename `tests.util.setup_test_homeserver` to `make_test_homeserver_synchronous`
* Move `tests.server.setup_test_homeserver` to `tests.util`

I'm not sure what the goal of this is though? Would it make more sense to just do the rename? Why do we want to do the move -- it seems that it just causes a ton of imports from `tests.util`.

reivilibre:
> I'm finding this hard to follow, I think the following operations are being done:
> 
>     * Rename `tests.util.setup_test_homeserver` to `make_test_homeserver_synchronous`

Yes, though it changes the calling style a little bit.

>     * Move `tests.server.setup_test_homeserver` to `tests.util`

That's right. That way, the old interface is preserved.
I could equally have just inlined `tests.server.setup_test_homeserver` into `tests.util.setup_test_homeserver`, but it was already longer than the screen and I thought it was nice to preserve the separated single responsibility of what is now `make_test_homeserver_synchronous`.

> I'm not sure what the goal of this is though?

Make things easier to follow. Having two `setup_test_homeserver` functions, one of which was unused except by the other, is pretty confusing — especially considering one does slightly more than the other.

> Would it make more sense to just do the rename?
> Why do we want to do the move -- it seems that it just causes a ton of imports from `tests.util`.

Having this function spread over two files feels odd and makes it harder to follow, I think.

Do you think it would maybe be better to move everything over to `tests.server`? It would mean updating quite a few imports but I guess it's less generic than `util`. There are too many `util` modules (3 or more?) in the tests so maybe that would be better..

reivilibre:
I've added the underscore and moved across `LEAVE_DB` (everything else is used in many other places, so I didn't want to churn too much).
Is that better or do you still think it'd be better to put this in `tests.util`? My main concern, I guess, is that `tests.util` is a bit of a catch-all name (and we have 3 of the things in the tests package :/).

clokep (approved):
I think this is overall an improvement, I'm sure we could clean it up more, but let's move on!

I think it might actually make sense to revert the last commit so that all the things read from environment variables are in a single spot? I'll leave that to your decision though!

PR #11507, opened 2021-12-03
Support MSC3266 room summaries over federation

deepbluev7:

### Pull Request Checklist


synapse/handlers/room_summary.py
```diff
@@ -1096,9 +1096,48 @@ async def get_room_summary(
...
-            #      as MSC needs more maturing on that side.
-            raise SynapseError(400, "Federation is not currently supported.")
+            # Reuse the hierarchy query over federation
+            if remote_room_hosts is None:
+                raise SynapseError(400, "Missing via to query remote room")
+
+            fed_rooms = await self._summarize_remote_room(
+                _RoomQueueEntry(room_id, remote_room_hosts),
+                suggested_only=True,
+                max_children=1,
```
squahtx, line 1106: Should this be `max_children=0`?

synapse/handlers/room_summary.py
```diff
@@ -1096,9 +1096,48 @@ async def get_room_summary(
...
                 room_summary["membership"] = membership or "leave"
         else:
-            # TODO federation API, descoped from initial unstable implementation
-            #      as MSC needs more maturing on that side.
-            raise SynapseError(400, "Federation is not currently supported.")
+            # Reuse the hierarchy query over federation
+            if remote_room_hosts is None:
+                raise SynapseError(400, "Missing via to query remote room")
+
+            fed_rooms = await self._summarize_remote_room(
```
squahtx, line 1103: `_summarize_remote_room` seems to use `/spaces/<room_id>` which I can't find in the final version of msc2946.
Should we be using this endpoint?
For reference, `_summarize_remote_room_hierarchy` uses `/hierarchy/<room_id>`

synapse/handlers/room_summary.py
```diff
@@ -1096,9 +1096,48 @@ async def get_room_summary(
...
+            )
+
+            for room_entry in fed_rooms:
+                room = room_entry.room
+                fed_room_id = room_entry.room_id
+
+                if fed_room_id == room_id:
+                    # If no user is specified, test against an invalid userid
+                    if requester is None:
+                        requester = ""
```
squahtx, line 1117: This is hacky; perhaps we could update `_is_remote_room_accessible` to handle `requester=None` to mean an unauthenticated local user?

For local rooms, we already do this with `_is_local_room_accessible` indirectly.

synapse/handlers/room_summary.py
```diff
@@ -1096,9 +1096,48 @@ async def get_room_summary(
...
+
+                    # The results over federation might include rooms that the we,
+                    # as the requesting server, are allowed to see, but the requesting
+                    # user is not permitted see.
```
squahtx, line 1121:
```suggestion
                    # user is not permitted to see.
```

synapse/handlers/room_summary.py
```diff
@@ -1096,9 +1096,48 @@ async def get_room_summary(
...
+                    if await self._is_remote_room_accessible(
+                        requester, fed_room_id, room
+                    ):
+                        # Before returning to the client, remove the allowed_room_ids
+                        # and allowed_spaces keys.
+                        room.pop("allowed_room_ids", None)
+                        room.pop("allowed_spaces", None)
+
+                        # A remote room can't be in the joined state
+                        room["membership"] = "leave"
```
squahtx, line 1133: Will this be inaccurate if the user has been invited to a remote room?

squahtx:
Thank you for getting the last bits of msc3266 implemented!
The PR's on the right track, though I have a few questions.

synapse/handlers/room_summary.py
deepbluev7: Dammit, you noticed me being lazy. Yes, I will fix that! :D

synapse/handlers/room_summary.py
deepbluev7: I think for invites we already store a shallow copy of the room locally, so we shouldn't end up in this case, but I am not too familiar with how that works, so actually no idea! Same about knocks, I suppose...

deepbluev7:
Thank you for the quick review, I'll see if I can get those details fixed over the week!

deepbluev7:
Since it has been so long, I rebased my changes now and addressed most feedback. It also is adapted to the latest changes on the MSC now. Not sure about the still open review comment though.

synapse/handlers/room_summary.py
```diff
@@ -713,8 +715,12 @@ async def _build_room_entry(self, room_id: str, for_federation: bool) -> JsonDic
             ),
             "guest_can_join": stats["guest_access"] == "can_join",
             "room_type": create_event.content.get(EventContentFields.ROOM_TYPE),
+            "im.nheko.summary.version": stats["version"],
         }
 
+        if stats["encryption"]:
+            entry["im.nheko.summary.encryption"] = stats["encryption"]
```
deepbluev7, line 722: I'm not sure, if I need the if to leave fields out and if it is a better idea to leave the encryption field out completely for unencrypted rooms and if I need to do the same for the version or not, since old rooms have no version. Otoh, that wasn't done for the name and topic, so...

synapse/handlers/room_summary.py
```diff
@@ -812,15 +816,50 @@ async def get_room_summary(
...
+                    # user is not permitted to see.
+                    #
+                    # Filter the returned results to only what is accessible to the user.
+                    if await self._is_remote_room_accessible(
+                        requester, fed_room_id, room
+                    ):
+                        # Before returning to the client, remove the allowed_room_ids
+                        # and allowed_spaces keys.
+                        room.pop("allowed_room_ids", None)
+                        room.pop("allowed_spaces", None)
```
squahtx, line 848: Can we remove this line? I don't think `allowed_spaces` is part of the spec anymore.
```suggestion
                        # Before returning to the client, remove the allowed_room_ids
                        # key.
                        room.pop("allowed_room_ids", None)
```

synapse/handlers/room_summary.py
```diff
@@ -812,15 +816,50 @@ async def get_room_summary(
...
+                        return room
+
+            raise NotFoundError("Room not found or is not accessible")
 
         return room_summary
 
 
 @attr.s(frozen=True, slots=True, auto_attribs=True)
 class _RoomQueueEntry:
+
```
squahtx, line 862: nit: stray newline in the diff
```suggestion
```

synapse/handlers/room_summary.py
```diff
@@ -645,7 +645,7 @@ async def _is_remote_room_accessible(
         due to an invite, etc.
 
         Args:
-            requester: The user requesting the summary.
+            requester: The user requesting the summary. If not passed only world readability is checked.
```
squahtx, line 648: nit: line length
```suggestion
            requester: The user requesting the summary. If not passed only world
                readability is checked.
```

squahtx (commented):
Thanks for making the changes. I've forgotten a lot of the context about this PR.

We could do with a test for the new code in `tests/handlers/test_room_summary.py`.  
There are a few examples of mocking federation responses in there.

squahtx:
@matrix-org/synapse-core Can we get input on https://github.com/matrix-org/synapse/pull/11507#discussion_r762965709?

tl;dr. what does `is_host_joined` return for a remote room that a local user has received an invite to?

synapse/handlers/room_summary.py
richvdh: > tl;dr. what does is_host_joined return for a remote room that a local user has received an invite to?

I've not read the context of this PR, I'm afraid, but `is_host_joined` looks at `current_state_events`, which will be empty for a room we are not a current member of - whether we have received an invite or not.

deepbluev7: Is there any function, that can return if for a room we are invited, in the knock state or joined?

deepbluev7:
I added a very basic test based on the room hierarchy tests. It is not very extensive, but most stuff is tested for hierarchies anyway.

tests/handlers/test_room_summary.py
```diff
@@ -1092,3 +1092,29 @@ def test_visibility(self):
         )
         result = self.get_success(self.handler.get_room_summary(user2, self.room))
         self.assertEqual(result.get("room_id"), self.room)
+
+    def test_fed(self):
+        """
+        Return data over federation and ensure that it is handled properly.
+        """
+        fed_hostname = self.hs.hostname + "2"
+        fedroom = "#subroom:" + fed_hostname
```
squahtx, line 1101: nit: For consistency with our naming convention, this ought to be `fed_room`

synapse/handlers/room_summary.py
squahtx: `_is_local_room_accessible` has a check for invites to rooms that we don't have state for. Maybe it can be copied?

squahtx:
Thanks for adding the test!

tests/handlers/test_room_summary.py
deepbluev7: Right, I renamed it from subroom, which for some reason didn't use an underscore :3

synapse/handlers/room_summary.py
deepbluev7: I guess I can just test that. However that wouldn't solve knocks, I think

deepbluev7: Okay, I think I found a solution (with anoa's help): `get_local_current_membership_for_user_in_room` seems to also work for remote rooms unless it is otherwise modified by purging the room for example. We tested it with knocks and invites.

synapse/handlers/room_summary.py
```diff
@@ -713,6 +716,8 @@ async def _build_room_entry(self, room_id: str, for_federation: bool) -> JsonDic
             ),
             "guest_can_join": stats["guest_access"] == "can_join",
             "room_type": create_event.content.get(EventContentFields.ROOM_TYPE),
+            "im.nheko.summary.version": stats["version"],
```
clokep, line 719: In some situations this could potentially fix the hack on L656-657 about assuming `knock` is valid. Would probably not be available often enough though.

We should guard these with the unstable config flag, probably?

synapse/handlers/room_summary.py
```diff
@@ -812,9 +817,51 @@ async def get_room_summary(
...
+            )
+
+            if room_entry:
+                room = room_entry.room
+                fed_room_id = room_entry.room_id
+
+                if fed_room_id == room_id:
+                    # The results over federation might include rooms that the we,
+                    # as the requesting server, are allowed to see, but the requesting
+                    # user is not permitted to see.
```
clokep, line 840:
```suggestion
                    # The results over federation might include rooms that we,
                    # as the requesting server, are allowed to see, but the requesting
                    # user is not permitted to see.
```

synapse/handlers/room_summary.py
```diff
@@ -812,9 +817,51 @@ async def get_room_summary(
...
+            ) = await self._summarize_remote_room_hierarchy(
+                _RoomQueueEntry(room_id, remote_room_hosts),
+                suggested_only=True,
+            )
+
+            if room_entry:
+                room = room_entry.room
+                fed_room_id = room_entry.room_id
+
+                if fed_room_id == room_id:
```
clokep, line 837: When would this not be true? (Or is just insurance?)

synapse/handlers/room_summary.py
```diff
@@ -812,9 +817,51 @@ async def get_room_summary(
...
+                    # as the requesting server, are allowed to see, but the requesting
+                    # user is not permitted to see.
+                    #
+                    # Filter the returned results to only what is accessible to the user.
+                    if await self._is_remote_room_accessible(
+                        requester, fed_room_id, room
+                    ):
+                        # Before returning to the client, remove the allowed_room_ids
+                        # key.
+                        room.pop("allowed_room_ids", None)
```
clokep, line 848: I think you need to make a copy of `room` before modifying it to avoid impacting any cached information from `FederationClient._get_room_hierarchy_cache` which is used by `FederationClient.get_room_hierarchy`

This is what `_RoomEntry.as_json()` does.

synapse/handlers/room_summary.py
deepbluev7: I guess it can't happen here anymore. This is a leftover from me iterating over the list of rooms, which could potentially have child rooms. I guess in practice this can never be false, but it also doesn't hurt anything and makes bugs more obvious imo. I can remove it however.

synapse/handlers/room_summary.py
deepbluev7: Sure, I always forget that python is by ref and not by value. Thanks for spotting!

synapse/handlers/room_summary.py
clokep: If we wanted to assert it I think we would want to do it in the shared code and raise an error on that case, instead of silently disregarding the bad data.

synapse/handlers/room_summary.py
deepbluev7: I think we should wait for this to be stable first and available for a while and then we can use that to fix the hack, since we need the remote server to use it. After a few months it is probably safe to just assume a room is version 1, if no version is returned.

I think an unstable flag would probably make more sense for the whole feature instead of just some namespaced fields. I was hoping the MSC just FCPs and passes in the next few weeks, since it is quite small and then we can just stabilize it (so I don't need to figure out how to add unstable flags :D). But I'll defer that to your judgement. If you want a flag, I can add it. I am just lazy .-.

synapse/handlers/room_summary.py
deepbluev7: Added an assert to the federation client now :3

synapse/handlers/room_summary.py
clokep: > I think an unstable flag would probably make more sense for the whole feature instead of just some namespaced fields. I was hoping the MSC just FCPs and passes in the next few weeks, since it is quite small and then we can just stabilize it (so I don't need to figure out how to add unstable flags :D). But I'll defer that to your judgement. If you want a flag, I can add it. I am just lazy .-.

There already is one for the whole feature -- I'm suggesting to re-use it and guard these fields with it.

deepbluev7: Okay, I think I figured it out. I cache the flag in the Handler, as it doesn't cache the hs yet and I think just caching the flag makes more sense? No idea :D

synapse/federation/federation_client.py
```diff
@@ -1426,6 +1426,8 @@ async def send_request(
             room = res.get("room")
             if not isinstance(room, dict):
                 raise InvalidResponseError("'room' must be a dict")
+            if not room.get("room_id") == room_id:
```
squahtx, line 1429:
```suggestion
            if room.get("room_id") != room_id:
```

squahtx:
Otherwise looks good from my end!

squahtx (approved):
Looks good from my end!
Awaiting re-review from @clokep before merging.

synapse/handlers/room_summary.py
```diff
@@ -812,9 +820,48 @@ async def get_room_summary(
...
+            (
+                room_entry,
+                children_room_entries,
+                inaccessible_children,
+            ) = await self._summarize_remote_room_hierarchy(
+                _RoomQueueEntry(room_id, remote_room_hosts),
+                suggested_only=True,
+            )
+
+            if room_entry:
```
clokep, line 836: I'm finding the level of indentation here rather difficult to follow. It could be nicer to check `if not room_entry` and raise immediately, rather than at the bottom of the method, but won't block on that.

synapse/handlers/room_summary.py
```diff
@@ -812,9 +820,48 @@ async def get_room_summary(
...
+                room = dict(room_entry.room)
+                fed_room_id = room_entry.room_id
```
clokep, line 838: I think the `fed_room_id` variable isn't useful now that we assert `fed_room_id` is equal to `room_id` in the federation code?

synapse/handlers/room_summary.py
deepbluev7: I've inverted the logic now and combined the 2 if cases, which gets rid of 2 levels of indentation. If I don't combine them I would have to duplicate the error messages, because I want the error to be the same for inaccessible or not found.

clokep (approved):
Looks great!

squahtx:
Is this something we want to merge immediately or are we waiting for the MSC process?

clokep:
> Is this something we want to merge immediately or are we waiting for the MSC process?

I think it can be merged since it is already partially merged (and behind a config flag).

Issue #7720, opened 2020-06-18
Synapse can't connect to an IPv6-only mail server via hostname


sandhose:
### Description

Synapse only does IPv4 address resolution for the `email.smtp_host` when sending email.
This might be (arguably) an issue on Twisted side though.

Internally, our SMTP server is accessible only with IPv6 (because dual-stack services is not a thing in Kubernetes right now) ; while we could probably work around, I think it's still a bug worth fixing, because IPv6-only environments/clouds are getting more and more traction.

### Steps to reproduce

- have a SMTP host with only an IPv6 (`AAAA`) entry
- configure it as the `email.smtp_host` on Synapse
- try to send an email

Here is the error we got from Sentry:
```
DNSLookupError: DNS lookup failed: postfix.mail.svc.cluster.local.
  File "synapse/handlers/identity.py", line 350, in send_threepid_validation
    await send_email_func(email_address, token, client_secret, session_id)
  File "synapse/push/mailer.py", line 207, in send_add_threepid_mail
    template_vars,
  File "synapse/push/mailer.py", line 323, in send_email
    requireTransportSecurity=self.hs.config.require_transport_security,
```

### Additional informations

I figured that Synapse was only doing `A` resolution by capturing on the container net interface. I wondered why this was the case for SMTP but not in other parts like the federation HTTP client.

It seems that it is because Synapse uses [`twisted.mail.smtp.sendmail`](https://github.com/twisted/twisted/blob/4f49872fe7623eb755d01441a0e5299f6f2363ff/src/twisted/mail/smtp.py#L2093-L2095) to send emails without doing explicit DNS resolution first, and `sendmail` directly does a `reactor.connectTcp` which chooses the address type (IPv4 or IPv6) [based on the host](https://github.com/twisted/twisted/blob/4f49872fe7623eb755d01441a0e5299f6f2363ff/src/twisted/internet/tcp.py#L1538). Because the host supplied to `sendmail` is a hostname and not an IP, Twisted considers it is an IPv4, not an IPv6.

I'm not sure if that should be fixed on Synapse or Twisted side.

### Version information

- **Homeserver**: not matrix.org
- **Version**: 1.15.1
- **Install method**: Kubernetes/docker image
- **Platform**: Containers on bare-metal

cc @ludovicm67, my colleague who found this issue

clokep:
> I'm not sure if that should be fixed on Synapse or Twisted side.

[...]

Issue #7695, opened 2020-06-14
v1.15 upgrade broken ipv6 replication config (ipv6 literals)


grinapo:
### Description
v1.13.x workers config broken after v1.15.0 upgrade. Server dies in mysterious ways.

```
2020-06-14 23:49:11,767 - synapse.http.client - 283 - INFO - replication-POSITION-27- Sending request GET http://::1:None/_synapse/replication/get_repl_stream_updates/federation/XSBWfTbBUR?from_token=0&upto_token=2
2020-06-14 23:49:11,767 - synapse.http.client - 330 - INFO - replication-POSITION-27- Error sending request to  GET http://::1:None/_synapse/replication/get_repl_stream_updates/federation/XSBWfTbBUR?from_token=0&upto_token=2: URLParseError expected integer for port, not ':1:None'
2020-06-14 23:49:11,768 - synapse.metrics.background_process_metrics - 215 - ERROR - replication-POSITION-27- Background process 'replication-POSITION' threw an exception
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/hyperlink/_url.py", line 993, in from_text
    port = int(port)
ValueError: invalid literal for int() with base 10: ':1:None'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/synapse/metrics/background_process_metrics.py", line 213, in run
    return (yield result)
hyperlink._url.URLParseError: expected integer for port, not ':1:None'
```

So far it has been working as:
```
worker_app: synapse.app.synchrotron
worker_replication_host: '::1'
worker_replication_port: 9092
```

Apart from it may have require `worker_replication_http_port` now (not sure though) the `…_host` is completely broken. I can try various valid combos without success, like

```
worker_replication_host: '[::1]'
```

but then it stupidly tries to interpret it as a hostname. I did not find a working solution, so I had to downgrade all connections to ipv4.

### Steps to reproduce
- use ipv6 literals in replication
- nothing works 

### Version information
- **Homeserver**: grin.hu
- **Version**: 1.15.0
- **Install method**: Debian (non NV)
- **Platform**: debian/buster

clokep:
Thanks for the bug report! I was able to reproduce this in sytest by running in worker mode with a command like:

```bash
[...]

Issue #6644, opened 2020-01-06
IPv6 seems unsupported for metrics-listener


cherti:
### Description

metrics-listener makes synapse crash when assigned an IPv6-bind_address

### Steps to reproduce

Add
```
listeners:
  - type: metrics
    port: 19000
    bind_addresses:
      - '127.0.0.1'
```
to the `/etc/matrix-synapse/homeserver.yaml` and everything works, change it to either
```
listeners:
  - type: metrics
    port: 19000
    bind_addresses:
      - '::1'
```
or
```
listeners:
  - type: metrics
    port: 19000
    bind_addresses:
      - '127.0.0.1'
      - '::1'
```
and synapse will not start anymore, instead yielding the following traceback:

```
twisted: [] Traceback (most recent call last):
twisted: []   File "/usr/lib/python3/dist-packages/synapse/app/_base.py", line 271, in start
twisted: []     hs.start_listening(listeners)
twisted: []   File "/usr/lib/python3/dist-packages/synapse/app/homeserver.py", line 293, in start_listening
twisted: []     _base.listen_metrics(listener["bind_addresses"], listener["port"])
twisted: []   File "/usr/lib/python3/dist-packages/synapse/app/_base.py", line 160, in listen_metrics
twisted: []     start_http_server(port, addr=host, registry=RegistryProxy)
twisted: []   File "/usr/lib/python3/dist-packages/synapse/metrics/_exposition.py", line 242, in start_http_server
twisted: []     httpd = _ThreadingSimpleServer((addr, port), CustomMetricsHandler)
twisted: []   File "/usr/lib/python3.7/socketserver.py", line 452, in __init__
twisted: []     self.server_bind()
twisted: []   File "/usr/lib/python3.7/http/server.py", line 137, in server_bind
twisted: []     socketserver.TCPServer.server_bind(self)
twisted: []   File "/usr/lib/python3.7/socketserver.py", line 466, in server_bind
twisted: []     self.socket.bind(self.server_address)
twisted: [] socket.gaierror: [Errno -9] Address family for hostname not supported
twisted: [] Unhandled error in Deferred:
twisted: [] 
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/synapse/app/_base.py", line 271, in start
    hs.start_listening(listeners)
  File "/usr/lib/python3/dist-packages/synapse/app/homeserver.py", line 293, in start_listening
    _base.listen_metrics(listener["bind_addresses"], listener["port"])
[...]

PR #11508, opened 2021-12-05
Tests for IPv6 bugs / redis replication over IPv6

evilham:
This will help close some long-standing issues:
("X" means we are adding a test for it)
[ ] AS            https://github.com/matrix-org/synapse/issues/4092
[ ] metrics EP    https://github.com/matrix-org/synapse/issues/6644
[X] replication   https://github.com/matrix-org/synapse/issues/7695
[X] smtp          https://github.com/matrix-org/synapse/issues/7720
[X] redis         https://github.com/matrix-org/synapse/issues/10694

In order to run these IPv6-related tests, the test suite must run with:
env SYNAPSE_POSTGRES=YES DEBUG_IPv6=YES trial tests

Sponsored by: ungleich.ch
Signed-off-by: @telmich Nico Schottelius <foss@ungleich.ch>

### DIFF
--- a/changelog.d/16016.doc
+++ b/changelog.d/16016.doc
@@ -1,0 +1,2 @@
+Clarify comment on the keys/upload over replication enpoint.
+
--- a/synapse/replication/http/devices.py
+++ b/synapse/replication/http/devices.py
@@ -107,8 +107,7 @@ class ReplicationUploadKeysForUserRestServlet(ReplicationEndpoint):
     Calls to e2e_keys_handler.upload_keys_for_user(user_id, device_id, keys) on
     the main process to accomplish this.
 
-    Defined in https://spec.matrix.org/v1.4/client-server-api/#post_matrixclientv3keysupload
-    Request format(borrowed and expanded from KeyUploadServlet):
+    Request format for this endpoint (borrowed and expanded from KeyUploadServlet):
 
         POST /_synapse/replication/upload_keys_for_user
 
@@ -117,6 +116,7 @@ class ReplicationUploadKeysForUserRestServlet(ReplicationEndpoint):
         "device_id": "<device_id>",
         "keys": {
             ....this part can be found in KeyUploadServlet in rest/client/keys.py....
+            or as defined in https://spec.matrix.org/v1.4/client-server-api/#post_matrixclientv3keysupload
         }
     }
 


richvdh (commented):
I appreciate the effort, but I don't really see the value of adding tests which are disabled by default. They'll just bitrot until they get removed again (see https://github.com/matrix-org/synapse/pull/11495/files#diff-cda5424a49ddae4fe5181900d1b607a9daf5c9afa539ed1fdb01622de069e499L198 for a recent example).

evilham:
Hey @richvdh, thanks for taking a look.

Maybe I should have elaborated a tad more:
the motivation with @telmich to start with these tests is that they provide a very simple path to check that these known bugs are solved without breaking Synapse's pipeline, and a simpler way to fix the issue with replication workers (the main motivation for this), since that is indeed more involved than it looks like at first.
Note that the tests are not really new, they re-use the code from existing tests, just with different fake DNS entries and changes in `tests/server.py` mean that this whole class of issues will be caught in the future, so that is also a win.

Indeed, I'd recommend that @xfk abandons and re-opens https://github.com/matrix-org/synapse/pull/10717#discussion_r699626644 with just the fix for the SMTP bug and enabling the corresponding test by default, that way we both check that the bug is fixed and that regressions do not happen.
If they do not in a reasonable time, since the original fix was somewhat simple and already signed-off, I'd see that this happens.

Also based on your comment on https://github.com/matrix-org/synapse/pull/10717#discussion_r699626644 I started re-working the replication bits, which is the most complex of these bugs, and will need to ask a couple questions to finish that.

richvdh:
Yes, having the tests is very useful, and your efforts in contributing them are very much appreciated. It's just that from a practical point of view I'd much rather we combine the fix and the tests into the same PR - so for example let's have a PR which fixes #7720 and lands the new test.

Perhaps we can mark this PR as a draft for now so that people can use it as inspiration for writing those PRs?

evilham:
OK, let's redirect this PR into fixing #10694 (and #7695 as a particular case), if in the meantime @xfk wants to open a PR with the fix for #7720 and the corresponding (passing and always enabled) test from this PR, that'd be great.

Based on previous discussion and particularly this comment https://github.com/matrix-org/synapse/pull/10717#discussion_r699626644, I've been looking into refactoring the communication with Redis using a `ClientService`, which proves more challenging than I first expected.

Here are some things I've noticed, please correct if any of these is wrong since understanding this right is quite important:

- Redis is only ever used through `synapse.server.HomeServer.get_outbound_redis_connection` and `RedisDirectTcpReplicationClientFactory`
- The `outbound_redis_connection` is used in the `synapse.replication.tcp.external_cache.ExternalCache` which is using the `RedisProtocol` directly.
- This `ExternalCache` is used in `synapse.federation.sender.FederationSender` in a subscription, read-only fashion, but it is used in a write-fashion by `synapse.handlers/message.EventCreationHandler`. This is a tad unexpected and confusing.
- It is also used by the `synapse.replication.tcp.handler.ReplicationCommandHandler`, passing it to `synapse.replication.tcp.redis.RedisDirectTcpReplicationClientFactory`.
- `RedisDirectTcpReplicationClientFactory` creates its own connection to Redis and it is used in tandem with the `outbound_redis_connection`, it uses the `outbound_redis_connection` in a write-fashion to send commands and the associated connection is used in a subscription, read-only fashion.

Now, the doubts:
- `synapse.replication.tcp.redis.lazyConnection` (used for the `outbound_redis_connection`) claims to return a `RedisProtocol`, but it realy returns a `txredisapi.ConnectionHandler` (see: https://github.com/matrix-org/synapse/blob/2f053f3f82ca174cc1c858c75afffae51af8ce0d/synapse/replication/tcp/redis.py#L383). Which.... yes and no, it wraps the behaviour of a `RedisProtocol`, but it is doing some magic regarding connection pools. I don't think this should happen.
- There is a periodic [ping](https://github.com/matrix-org/synapse/blob/2f053f3f82ca174cc1c858c75afffae51af8ce0d/synapse/replication/tcp/redis.py#L230) that claims to keep the connection alive, I wonder if this would be needed if we switch to `ClientService`, how exactly did `txredisapi` "stop"? From my understanding of Twisted, the instance of the `RedisProtocol` should be treated as something that is bound to change, so if the connection is lost, you should get another one, with the proper factories and all that, it should be ready to use. Maybe the fact that `txredisapi` is exposing weird objects like the `ConnectionHandler` lead to treating it as something semi-persistent and to these issues?

My overall plan to make this happen is:
- Stay away from `txredisapi.ConnectionHandler` and `txredisapi.RedisFactory`, they are trying to do way too much and are inflexible.
- Create our own `SimpleRedisFactory` (which might be upstreamed to `txredisapi` in the future, but in the meantime should live in synapse as well due to packaging) which does none of the reconnection and pool magic.
- Use this `SimpleRedisFactory` with a `RedisClientService` to have free connections with exponential backoff and reset of the timers.
- Never pass the `RedisProtocol` from `synapse.server.HomeServer` (or ever), but instead the `RedisClientService` which should have ways of letting users of the service know when the connection is lost / recovered.
- `RedisDirectTcpReplicationClientFactory` should use two such `RedisClientService` objects, the one passed by the `HomeServer` and its own connection service.

Does this sound OK? am I missing something?

richvdh:
> - Redis is only ever used through `synapse.server.HomeServer.get_outbound_redis_connection` and `RedisDirectTcpReplicationClientFactory`
> - The `outbound_redis_connection` is used in the `synapse.replication.tcp.external_cache.ExternalCache` which is using the `RedisProtocol` directly.
> - This `ExternalCache` is used in `synapse.federation.sender.FederationSender` in a subscription, read-only fashion, but it is used in a write-fashion by `synapse.handlers/message.EventCreationHandler`. This is a tad unexpected and confusing.
> - It [`outbound_redis_connection`] is also used by the `synapse.replication.tcp.handler.ReplicationCommandHandler`, passing it to `synapse.replication.tcp.redis.RedisDirectTcpReplicationClientFactory`.
> - `RedisDirectTcpReplicationClientFactory` creates its own connection to Redis and it is used in tandem with the `outbound_redis_connection`, it uses the `outbound_redis_connection` in a write-fashion to send commands and the associated connection is used in a subscription, read-only fashion.

This is all correct.

> This [use of `ExternalCache`] is a tad unexpected and confusing.

I don't think you need to worry too much about how `ExternalCache` works: for our purposes it's just a class which needs access to a redis connection where it can send commands and receive the replies. But I'm not sure which part is unexpected: the fact that `EventCreationHandler` writes to the cache? What would be the use of a cache which is never populated?


> Now, the doubts:
> - `synapse.replication.tcp.redis.lazyConnection` (used for the `outbound_redis_connection`) claims to return a `RedisProtocol`, but it realy returns a `txredisapi.ConnectionHandler`. Which.... yes and no, it wraps the behaviour of a `RedisProtocol`, but it is doing some magic regarding connection pools. I don't think this should happen.

Maybe. It would certainly be nice if there were a type definition for the interface exposed by `RedisProtocol`, so that `lazyConnection` can say "here is a thing which behaves the same as a `RedisProtocol`, but isn't one". But `txredisapi` doesn't define such a class.

> - There is a periodic [ping](https://github.com/matrix-org/synapse/blob/2f053f3f82ca174cc1c858c75afffae51af8ce0d/synapse/replication/tcp/redis.py#L230) that claims to keep the connection alive,

Did you link to the right line here? That's nothing to do with pings, but rather about whether `txredisapi.RedisFactory` reconnects after a connection failure. But I think it is specific to `RedisFactory` and not required for `ClientService`.

> My overall plan to make this happen is:
> - Stay away from `txredisapi.ConnectionHandler` and `txredisapi.RedisFactory`, they are trying to do way too much and are inflexible.

+1

> - Create our own `SimpleRedisFactory` (which might be upstreamed to `txredisapi` in the future, but in the meantime should live in synapse as well due to packaging) which does none of the reconnection and pool magic.

You may find that a factory created with [`Factory.forProtocol`](https://twistedmatrix.com/documents/current/api/twisted.internet.protocol.Factory.html#forProtocol) is sufficent here.

> - Use this `SimpleRedisFactory` with a `RedisClientService` to have free connections with exponential backoff and reset of the timers.

Are you intending to make `RedisClientService` a derived class of `ClientService`? I would avoid doing so - either wrap the `ClientService` *inside* another class, or just pass the `ClientService` around.

> - Never pass the `RedisProtocol` from `synapse.server.HomeServer` (or ever), but instead the `RedisClientService` which should have ways of letting users of the service know when the connection is lost / recovered.

I'm not sure if you actually need to know that? Could you just call `ClientService.whenConnected` to get the current connection?

> - `RedisDirectTcpReplicationClientFactory` should use two such `RedisClientService` objects, the one passed by the `HomeServer` and its own connection service.
>
> Does this sound OK? am I missing something?

I think it sounds about right. To be honest, I'm struggling to figure out how it all fits together when looking at it in the abstract.

richvdh:
@evilham: are you still interested in working on this?

evilham:
Yep, I want to get this fixed and it's on my TODO, but things have gotten in the way >,<. Will try to re-prioritise, thanks for the ping!

clokep:
This has a bunch of conflicts that need to be solved; I wonder if a more reasonable way to handle this would be to:

* Close this PR;
* Take the tests for each item you've listed above (e.g. SMTP) and open a new PR with just those tests;
* Fix support for whatever that item tests in the same PR;
* Review it!

Hopefully this is pretty doable and folks can crib off this.

PR #11511, opened 2021-12-06
Correctly ignore invites from ignored users

### LINKED ISSUE #11506: Invites from ignored users may not be ignored
Midway down the digestive system of Synapse's sync code, we fetch all membership changes for a user across the sync period.

After 1709, `event` will be the last element in the list of events, or it'll be unset if there were no such events. They're fetched in ascending stream ordering order, so `event` will be the membership change with highest stream ordering.

https://github.com/matrix-org/synapse/blob/d26808dd854006bd26a2366c675428ce0737238c/synapse/handlers/sync.py#L1703-L1709

There are no other writes to `event`, but there is a read:

https://github.com/matrix-org/synapse/blob/d26808dd854006bd26a2366c675428ce0737238c/synapse/handlers/sync.py#L1794-L1800

I smell a bug. 

- Suppose my `membership_change_events` consists of three events in this order:
   - I'm invited to room A by someone I've ignored
   - I'm invited to room B by someone I've not ignord.
- `event` will be the invite event in room B.
- mem_change_events_by_room_id will be the dictionary `{A: [invite_A], B: [invite_B]}`.
- We start looping over the keys of this dictionary.
- When I process room A, we test `event.sender not in ignored users`
- **`event` refers `invite_B`, whose sender is not ignored. We then push invite_A down to the user.

Well this is just embarrassing. Should be an easy fix though.

DMRobertson:
Fixes #11506. Corresponding test case at matrix-org/complement#267.

### DIFF
--- a/changelog.d/11511.bugfix
+++ b/changelog.d/11511.bugfix
@@ -1,0 +1,1 @@
+Fix a long-standing bug where invites from ignored users would not be omitted from incremental syncs.
--- a/changelog.d/11511.bugfix
+++ b/changelog.d/11511.bugfix
@@ -1,1 +1,1 @@
-Fix a long-standing bug where invites from ignored users would not be omitted from incremental syncs.
+Fix a long-standing bug where invites from ignored users were included in incremental syncs.
--- a/synapse/handlers/sync.py
+++ b/synapse/handlers/sync.py
@@ -1771,6 +1771,7 @@ class SyncHandler:
 
             if not non_joins:
                 continue
+            last_non_join = non_joins[-1]
 
             # Check if we have left the room. This can either be because we were
             # joined before *or* that we since joined and then left.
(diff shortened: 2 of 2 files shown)


squahtx (approved):
LGTM!

I found the double negative in the changelog slightly hard to read.
`s/not be omitted/included/`?

DMRobertson:
Thanks, I'll reword that. Is it best to wait for the complement test to be approved before merging too?

squahtx:
> Thanks, I'll reword that. Is it best to wait for the complement test to be approved before merging too?

Probably! I guess we'd want the complement test to be included in the CI for this change.

DMRobertson:
> Probably! I guess we'd want the complement test to be included in CI for this change.

It has done, because the branch names match. see https://github.com/matrix-org/synapse/runs/4431172326?check_suite_focus=true#step:7:1048

DMRobertson:
That sytest is a known flake. I'll rerun CI.

PR #11515, opened 2021-12-06
Refactors in `_generate_sync_entry_for_rooms`

DMRobertson:
Small things that helped me when I was reading all of this stuff.

- 0af0643c7 is probably the most useful.
- f9dff2d60 and 963de4e5c pull out isolated chunks of code into their own functions.
- 0e2c6f6fc: I'm not sure if this makes things clearer or just introduces more noise. Input welcome.

### DIFF
--- a/synapse/handlers/sync.py
+++ b/synapse/handlers/sync.py
@@ -1612,6 +1612,12 @@ class SyncHandler:
         )
 
     async def _get_ignored_users(self, user_id: str) -> FrozenSet[str]:
+        """Retrieve the users ignored by the given user from their global account_data.
+
+        Returns an empty set if
+        - there is no global account_data entry for ignored_users
+        - there is such an entry, but it's not a JSON object.
+        """
         ignored_account_data = (
             await self.store.get_global_account_data_by_type_for_user(
                 AccountDataTypes.IGNORED_USER_LIST, user_id=user_id
--- a/synapse/handlers/sync.py
+++ b/synapse/handlers/sync.py
@@ -1512,8 +1512,8 @@ class SyncHandler:
             account_data_by_room: Dictionary of per room account data
 
         Returns:
-            Returns a 4-tuple describing rooms we've joined or left, and users who've
-            joined or left rooms we're in. This gets used later in
+            Returns a 4-tuple describing rooms the user has joined or left, and users who've
+            joined or left rooms any rooms the user is in. This gets used later in
             `_generate_sync_entry_for_device_list`.
 
             Its entries are:
--- a/synapse/handlers/sync.py
+++ b/synapse/handlers/sync.py
@@ -184,20 +184,6 @@ class GroupsSyncResult:
 
 
 @attr.s(slots=True, frozen=True, auto_attribs=True)
-class MembershipChangesSummary:
-    """
-    A summary of rooms we've joined or left, plus users who we've seen join or leave.
-
-    Used to create the DeviceLists struct below.
-    """
-
-    newly_joined_rooms: Collection[str]
-    newly_joined_or_invited_or_knocked_users: Collection[str]
-    newly_left_rooms: Collection[str]
-    newly_left_users: Collection[str]
-
-
-@attr.s(slots=True, frozen=True, auto_attribs=True)
 class DeviceLists:
     """
     Attributes:
@@ -1126,9 +1112,11 @@ class SyncHandler:
 
         logger.debug("Fetching room data")
 
-        membership_changes_summary = await self._generate_sync_entry_for_rooms(
+        res = await self._generate_sync_entry_for_rooms(
             sync_result_builder, account_data_by_room
         )
+        newly_joined_rooms, newly_joined_or_invited_or_knocked_users, _, _ = res
+        _, _, newly_left_rooms, newly_left_users = res
 
         block_all_presence_data = (
             since_token is None and sync_config.filter_collection.blocks_all_presence()
@@ -1137,8 +1125,8 @@ class SyncHandler:
             logger.debug("Fetching presence data")
             await self._generate_sync_entry_for_presence(
                 sync_result_builder,
-                membership_changes_summary.newly_joined_rooms,
-                membership_changes_summary.newly_joined_or_invited_or_knocked_users,
+                newly_joined_rooms,
+                newly_joined_or_invited_or_knocked_users,
             )
 
         logger.debug("Fetching to-device data")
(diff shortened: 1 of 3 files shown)


DMRobertson:
Sytest failure wasn't a known flake, but only happened on one instance.

Here's the logs:

```
21061-2021-12-06 18:04:19,595 - synapse.http.server - 101 - ERROR - POST-2523 - Failed handle request via 'RoomCreateRestServlet': <SynapseRequest at 0x7f8d18dc8f98 method='POST' uri='/_matrix/client/r0/createRoom?access_token=<redacted>' clientproto='HTTP/1.1' site='8824'>
21062-Traceback (most recent call last):
21063-  File "/venv/lib/python3.7/site-packages/synapse/http/server.py", line 262, in _async_render_wrapper
21064-    callback_return = await self._async_render(request)
21065-  File "/venv/lib/python3.7/site-packages/synapse/http/server.py", line 451, in _async_render
21066-    callback_return = await raw_callback_return
21067-  File "/venv/lib/python3.7/site-packages/synapse/rest/client/room.py", line 90, in on_POST
21068-    requester, self.get_room_config(request)
21069-  File "/venv/lib/python3.7/site-packages/synapse/handlers/room.py", line 919, in create_room
21070-    last_stream_id,
21071-  File "/venv/lib/python3.7/site-packages/synapse/replication/tcp/client.py", line 302, in wait_for_stream_position
21072-    await make_deferred_yieldable(deferred)
21073-  File "/venv/lib/python3.7/site-packages/twisted/internet/defer.py", line 859, in _runCallbacks
21074-    current.result, *args, **kwargs
21075-  File "/venv/lib/python3.7/site-packages/synapse/util/async_helpers.py", line 532, in convert_cancelled
21076-    raise defer.TimeoutError("Timed out after %gs" % (timeout,))
21077-twisted.internet.defer.TimeoutError: Timed out after 30s
21078:2021-12-06 18:04:19,599 - synapse.access.http.8824 - 421 - I
```

synapse/handlers/sync.py
```diff
@@ -1630,6 +1618,22 @@ async def handle_room_entries(room_entry: "RoomSyncResultBuilder") -> None:
             newly_left_users,
         )
 
+    async def _get_ignored_users(self, user_id: str) -> FrozenSet[str]:
```
clokep, line 1614: Docstring please!

synapse/handlers/sync.py
```diff
@@ -1630,6 +1618,22 @@ async def handle_room_entries(room_entry: "RoomSyncResultBuilder") -> None:
...
+        ignored_account_data = (
+            await self.store.get_global_account_data_by_type_for_user(
+                AccountDataTypes.IGNORED_USER_LIST, user_id=user_id
+            )
+        )
```
clokep, line 1619: I wonder if we can use the `ignored_users` table instead of this...? But that's a real change, not just a refactor.

synapse/handlers/sync.py
```diff
@@ -1506,16 +1512,22 @@ async def _generate_sync_entry_for_rooms(
             account_data_by_room: Dictionary of per room account data
 
         Returns:
-            Returns a 4-tuple whose entries are:
+            Returns a 4-tuple describing rooms we've joined or left, and users who've
+            joined or left rooms we're in. This gets used later in
+            `_generate_sync_entry_for_device_list`.
```
clokep, line 1517:
```suggestion
            Returns a 4-tuple describing rooms the user has joined or left, and users who've
            joined or left rooms any rooms the user is in. This gets used later in
            `_generate_sync_entry_for_device_list`.
```

synapse/handlers/sync.py
```diff
@@ -2353,6 +2333,34 @@ class SyncResultBuilder:
     to_device: List[JsonDict] = attr.Factory(list)
 
 
+def _calculate_user_changes(
+    sync_result_builder: "SyncResultBuilder",
+) -> Tuple[Set[str], Set[str]]:
```
clokep, line 2344: I wonder if this should be start of `SyncResultBuilder`?

synapse/handlers/sync.py
```diff
@@ -183,6 +183,20 @@ def __bool__(self) -> bool:
         return bool(self.join or self.invite or self.leave)
 
 
+@attr.s(slots=True, frozen=True, auto_attribs=True)
+class MembershipChangesSummary:
```
clokep, line 187: I'm not sure if using a class here instead of the 4-tuple is much clearer.

clokep (commented):
Seems reasonable overall. 👍

synapse/handlers/sync.py
DMRobertson: Didn't know about that, thanks. Though it's frustrating that it's not clear where the source of truth is. I'll add a #TODO.

clokep: They should be in-sync, this table was added as an optimization. It might not take into account everything you need though!

synapse/handlers/sync.py
DMRobertson: I'm glad you agree---I wasn't convinced either. I'll revert https://github.com/matrix-org/synapse/commit/0e2c6f6fcd13237514896d64af8b039e2cf4d9f4.

synapse/handlers/sync.py
DMRobertson: Not sure what you mean here, sorry. Perhaps that this ought to be placed above the definition of `SyncResultBuilder`?

Oh, actually I think you mean the quotes are unnecessary.

(My personal preference would be to move SyncResultBuilder and all other POD structs/classes all the way up to the top of the file, so the quotes aren't needed throughout.)

synapse/handlers/sync.py
DMRobertson: Coming in dba7c941a.

synapse/handlers/sync.py
DMRobertson: Note upcoming in ecccfb405.

synapse/handlers/sync.py
DMRobertson: ... reverted in c38478998.

synapse/handlers/sync.py
DMRobertson: Trimmed quotes in dd39d8276.

clokep: What I meant to say was:

> I wonder if this should be *part* of `SyncResultBuilder`?

As in a method on that class! Clearly I failed to say anything reasonable.

DMRobertson: Ahh, I see. I considered this, but opted to make `_calculate_user_changes` a free function, to try and suggest that it's just reading data and not mutating the `SyncResultBuilder` instance.

I think this is just an old C++ habit dying hard though. Happy to go with whatever you think is clearest!

clokep: I think it is fine either way, `SyncResultBuilder` seems to just be a way to pass around mutable state so probably best to keep logic out of it.

clokep (approved):

PR #11518, opened 2021-12-06
Correctly register shutdown handler for presence workers

### LINKED ISSUE #11517: Error on shutdown when prescence is enabled
Spotted when I was reading through sytest logs.

```
21082-2021-12-06 18:04:19,715 - synapse.metrics.background_process_metrics - 246 - ERROR - generic_presence.on_shutdown-0 - Background process 'generic_presence.on_shutdown' threw an exception
21083-Traceback (most recent call last):
21084-  File "/venv/lib/python3.7/site-packages/synapse/metrics/background_process_metrics.py", line 242, in run
21085-    return await func(*args, **kwargs)
21086-TypeError: object NoneType can't be used in 'await' expression
```

The problem is that we register a call to `run_as_background_process` before shutdown here. We pass it the callable `self._on_shutdown`. The latter is a synchronous function, but `run_as_background_process` expects an async function.

https://github.com/matrix-org/synapse/blob/75ca0a6168f92dab3255839cf85fb0df3a0076c3/synapse/handlers/presence.py#L416-L428

The symptom was introduced in #10847 when we removed `maybe_awaitable`. But the underlying cause is that `_on_shutdown` didn't return a Deferred back when we expected it to. (Plus `addSystemEventTrigger` obscures the call from mypy. Probably extra confusion when this was rewritten from twisted defers to asyncs.)

DMRobertson:
Fixes #11517. See issue for the diagnosis.


clokep (approved):
Seems reasonable.

Anyway we could have caught this? (mypy? Something else?)

DMRobertson:
> Anyway we could have caught this? (mypy? Something else?)

I don't think there's an easy way. The problem is that this mypy doesn't know that we're going to call the third argument with the fourth and fifths as parameters.

```
        hs.get_reactor().addSystemEventTrigger(
            "before",
            "shutdown",
            run_as_background_process,
            "generic_presence.on_shutdown",
            self._on_shutdown,
        )
```

[ParamSpec](https://docs.python.org/3/library/typing.html?highlight=paramspec#typing.ParamSpec) might(?) help in the future but mypy hasn't got full support for that [yet](https://github.com/python/mypy/issues/8645). And anyway, that'd only be useful if twisted's annotations were updated, and I'm not sure that the twisted authors have a great deal of time to review changes ATM.

We could instead pass a callable that takes no arguments:

```
        hs.get_reactor().addSystemEventTrigger(
            "before",
            "shutdown",
            lambda: run_as_background_process(
                "generic_presence.on_shutdown", self._on_shutdown,
            ),
        )
```

Mypy rejects this.

```
$ mypy synapse/handlers/presence.py 
synapse/handlers/presence.py:419: error: Argument 2 to "run_as_background_process" has incompatible type "Callable[[], None]"; expected "Callable[..., Awaitable[Optional[Any]]]"  [arg-type]
Found 1 error in 1 file (checked 1 source file)
```

PyCharm makes it a warning too!

PR #11522, opened 2021-12-07
Fix the test breakage introduced by #11435 as a result of concurrent PRs

reivilibre:
Typical story, one of the PRs changed something that needed tests changing,

another PR had a test added because 'this needs more tests!'

and oops.

Sorry :)


squahtx (approved):
lgtm!

PR #11523, opened 2021-12-07
Fix 'delete room' admin api to work on incomplete rooms

### LINKED ISSUE #6547: Purge Room API reports "Unknown room" for room that exists
I have been in some kind of half in half out state with `!OGEhHVWSdvArJzumhm:matrix.org`, see https://github.com/matrix-org/synapse/issues/6520. Today I clicked clear cache in Riot and sync was failing with the following error:
Screen Shot 2019-12-13 at 8 02 10 PM

Attempting to use the purge room API on `!OGEhHVWSdvArJzumhm:matrix.org` returned:
```
{
    "errcode": "M_NOT_FOUND",
    "error": "Unknown room !OGEhHVWSdvArJzumhm:matrix.org"
}
```

The only way I could figure out how to get my sync to work was to go into the Synapse manhole and run `hs.get_datastore().purge_room('!OGEhHVWSdvArJzumhm:matrix.org')` which let the sync complete.

richvdh:
If, for some reason, we don't have the create event, we should still be able to purge a room.

Fixes #6547.

### DIFF
--- a/synapse/handlers/room.py
+++ b/synapse/handlers/room.py
@@ -1535,20 +1535,13 @@ class RoomShutdownHandler:
             await self.store.block_room(room_id, requester_user_id)
 
         if not await self.store.get_room(room_id):
-            if block:
-                # We allow you to block an unknown room.
-                return {
-                    "kicked_users": [],
-                    "failed_to_kick_users": [],
-                    "local_aliases": [],
-                    "new_room_id": None,
-                }
-            else:
-                # But if you don't want to preventatively block another room,
-                # this function can't do anything useful.
-                raise NotFoundError(
-                    "Cannot shut down room: unknown room id %s" % (room_id,)
-                )
+            # if we don't know about the room, there is nothing left to do.
+            return {
+                "kicked_users": [],
+                "failed_to_kick_users": [],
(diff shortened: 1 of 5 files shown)


synapse/handlers/room.py
```diff
@@ -1535,20 +1535,13 @@ async def shutdown_room(
...
-                    "failed_to_kick_users": [],
-                    "local_aliases": [],
-                    "new_room_id": None,
-                }
-            else:
-                # But if you don't want to preventatively block another room,
-                # this function can't do anything useful.
-                raise NotFoundError(
-                    "Cannot shut down room: unknown room id %s" % (room_id,)
-                )
```
babolivier, line 1551: I'm a bit on the fence about this. I'd think it more helpful to only return a 200 response if something has actually happened. But it's not a hill I'm willing to die on.

babolivier (approved):

synapse/handlers/room.py
richvdh: I don't disagree, but at this point, we don't know if something has actually happened, because we haven't yet tried to delete things. Keeping track of "did anything actually happen" is fiddly, at best.

PR #11524, opened 2021-12-07
Fix `ModuleApi.looping_background_call` for non-async functions

squahtx:
After #10847, `looping_background_call` would print an error in the logs
every time a non-async function was called. Since the error would be
caught and ignored immediately, there were no other side effects.


synapse/module_api/__init__.py
```diff
@@ -1014,9 +1015,7 @@ def looping_background_call(
                 run_as_background_process,
                 msec,
                 desc,
-                f,
-                *args,
-                **kwargs,
+                lambda: maybe_awaitable(f(*args, **kwargs)),
```
DMRobertson, line 1018: A shame mypy couldn't have spotted this.

I wonder how feasible a mypy plugin would be here?

DMRobertson (approved):
I'm happy if CI is happy.

synapse/module_api/__init__.py
squahtx: `ParamSpec` would have caught this, if only mypy had support for it.
I'm not a big fan of plugins since they tie you to a specific type checker.

PR #11527, opened 2021-12-07
Revert "Move `glob_to_regex` and `re_word_boundary` to `matrix-python-common` (#11505)

squahtx:
This reverts commit a77c36989785c0d5565ab9a1169f4f88e512ce8a.

Our downstream package maintainers need more notice before adding a new package.
To be revisited in the new year.

### DIFF
--- a/changelog.d/11527.misc
+++ b/changelog.d/11527.misc
@@ -1,0 +1,1 @@
+Temporarily revert usage of `matrix-python-common`.
--- a/changelog.d/11505.misc
+++ b/changelog.d/11505.misc
@@ -1,0 +1,1 @@
+Move `glob_to_regex` and `re_word_boundary` to `matrix-python-common`.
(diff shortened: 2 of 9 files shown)


changelog.d/11527.misc
```diff
@@ -0,0 +1 @@
+Temporarily revert usage of `matrix-python-common`.
```
clokep, line 1: Can we nuke both changelogs after this merges so we don't add anything to the changes doc?

DMRobertson (approved):
Seems reasonable to me, let's sanity check that CI succeeds though?

clokep (approved):
Seems reasonable, I assume this was a `git revert`.

changelog.d/11527.misc
squahtx: Yep, I'll be cutting them out of the changes doc after.

PR #11530, opened 2021-12-07
Strip unauthorized fields from `unsigned` object in events received over federation

### LINKED ISSUE #11080: Synapse is overly accepting of content in the `unsigned` object in events received over federation
Synapse makes use of various properties within the `unsigned` object of events - either internally, or by passing them on to clients. One example is `replaces_state`, which is used to store the event id of the previous event with the same `type` and `state_key`, and is later used to populate the `prev_content` property for events served to clients.

The problem is that homeservers are free to populate `unsigned`, without it affecting the event hashes or signatures; a malicious or buggy homeserver could therefore populate the content with incorrect data.

Taking the example of `replaces_state`, Synapse overwrites this property when receiving an event, but only if there was previously an event with the same `type` and `state_key` in the room state; it is otherwise passed through unchanged. So, a malicious homeserver could confuse remote servers' clients by sending incorrect values of `replaces_state` over federation.

---

The specification is not clear on how unspecified properties within `unsigned` should be handled, but I think they should be stripped off by the receiving homeserver. This will ensure that if, in future, the C-S API spec is extended to specify new properties be added to `unsigned`, there will be no confusion about whether they were added by the local or remote homeserver.

As far as I am aware, the only properties that *should* be allowed in `unsigned` over federation are:
 * `invite_room_state`
 * `knock_room_state`
 * `age` - though see also https://github.com/matrix-org/synapse/issues/8429.

[Aside: in an ideal world, we might have different properties for "things added by the remote homeserver - treat with caution!" vs "things added by the local homeserver - can be trusted". However, that ship has probably sailed for now.]

H-Shay:
Fixes #11080. Currently Synapse does not check if the fields in `unsignedData` correspond to what is outlined in the spec, this PR adds functionality to do so and strips any fields which are unauthorized.


tests/test_federation.py
```diff
@@ -276,3 +276,108 @@ def test_cross_signing_keys_retry(self):
...
+                },
+            }
+        )
+        self.get_success(federation_event_handler.on_receive_pdu("test.serv", event3))
+
+        event = self.get_success(self.store.get_event("$event3:test.serv"))
+        event_dict = event.get_dict()
+        self.assertTrue("age" in event_dict["unsigned"])
+        self.assertTrue("invite_room_state" not in event_dict["unsigned"])
+        self.assertTrue("more warez" not in event_dict["unsigned"])
```
callahad, line 383: Can use `self.assertIn` and `self.assertNotIn` here for slightly nicer error messages if it fails

tests/test_federation.py
```diff
@@ -276,3 +276,111 @@ def test_cross_signing_keys_retry(self):
             "ed25519:" + remote_self_signing_key in self_signing_key["keys"].keys(),
         )
         self.assertTrue(remote_self_signing_key in self_signing_key["keys"].values())
+
+
+class StripUnsignedFromEventsTestCase(MessageAcceptTests):
+    def setUp(self):
+        super().setUp()
```
DMRobertson, line 283: This looks redundant---the superclass method will get called if this override is deleted.

tests/test_federation.py
```diff
@@ -276,3 +276,111 @@ def test_cross_signing_keys_retry(self):
...
+
+    def strip_unauthorized_unsigned_values(self):
```
DMRobertson, line 285: Can you check that these tests get run by inspecting the `trial` output? IIRC only methods whose name starts with `test` are actually considered as tests to be run.

synapse/handlers/federation_event.py
```diff
@@ -1939,3 +1949,33 @@ def _sanity_check_event(self, ev: EventBase) -> None:
...
+
+        filtered_unsigned = {}
+
+        # Unsigned should never be None by the time we are here but mypy doesn't know that
+        assert unsigned is not None
+        for k, v in unsigned.items():
+            if k in whitelist:
+                filtered_unsigned[k] = v
+
+        event_dict["unsigned"] = filtered_unsigned
```
DMRobertson, line 1975: This mutates the `event_dict` in place. Assuming that's intentional, I think we should note this in the docstring.

(Note that EventBase.get_dict() returns a copy of the event's internal dict, so... we're not mutating the original event. I guess I was just a bit surprised that operate in two modes simulatneously: we mutate the `event_dict` parameter but build a new EventBase.)

synapse/handlers/federation_event.py
```diff
@@ -1939,3 +1949,33 @@ def _sanity_check_event(self, ev: EventBase) -> None:
...
+        unsigned = event_dict.get("unsigned")
+
+        if pdu.type == EventTypes.Member:
+            whitelist = ["knock_room_state", "invite_room_state", "age"]
+        else:
+            whitelist = ["age"]
+
+        filtered_unsigned = {}
+
+        # Unsigned should never be None by the time we are here but mypy doesn't know that
```
DMRobertson, line 1969: Is that because of the checks preceding each call of `strip_unsigned_values`? It might be reasonable to just use `unsigned = event_dict["unsigned"]` and avoid the assert in this case.

tests/test_federation.py
```diff
@@ -276,3 +276,111 @@ def test_cross_signing_keys_retry(self):
...
+                "prev_events": [(most_recent, {})],
+                "unsigned": {"malicious garbage": "hackz", "more warez": "more hackz"},
+            }
+        )
+        self.get_success(federation_event_handler.on_receive_pdu("test.serv", event1))
+
+        event = self.get_success(self.store.get_event("$event1:test.serv"))
+        event_dict = event.get_dict()
+        # Make sure unauthorized fields are stripped from unsigned
+        self.assertNotIn("hackz", event_dict["unsigned"])
```
DMRobertson, line 312: I think this assertion will pass if you reverted your code changes. The same applies to line +348 below.

It's going to assert `hacks not in event_dict["unsigned"]`. But `x in my_dict` tests that a _key_ `x` is contained in a dictionary:

```
>>> 1 in {"a": 1}
False
>>> "a" in {"a": 1}
True
```

tests/test_federation.py
```diff
@@ -276,3 +276,111 @@ def test_cross_signing_keys_retry(self):
...
+        self.get_success(
+            federation_event_handler.on_send_membership_event("test.serv", event2)
+        )
+
+        event = self.get_success(self.store.get_event("$event2:test.serv"))
+        event_dict = event.get_dict()
+        self.assertIn("age", event_dict["unsigned"])
+        self.assertNotIn("hackz", event_dict["unsigned"])
+        # Invite_room_state is allowed in events of type m.room.member
+        self.assertIn("invite_room_state", event_dict["unsigned"])
```
DMRobertson, line 350: Could also check that the values for `age` and `invite_room_state` are unaltered by your change.

tests/test_federation.py
```diff
@@ -276,3 +276,111 @@ def test_cross_signing_keys_retry(self):
...
+                    "invite_room_state": [],
+                },
+            }
+        )
+        self.get_success(federation_event_handler.on_receive_pdu("test.serv", event3))
+
+        event = self.get_success(self.store.get_event("$event3:test.serv"))
+        event_dict = event.get_dict()
+        self.assertIn("age", event_dict["unsigned"])
+        # Invite_room_state field is only associated with event type m.room.member
```
DMRobertson, line 384: perhaps change `associated with event type ...` to `permitted in "unsigned" for event type ...`?

Not sure how much clearer that is, feel free to use your discretion!

DMRobertson (dismissed):
A few thoughts. Happy to talk through them as and when.

synapse/handlers/federation_event.py
H-Shay: Right I can see how that would be a little ...unclean... What might be a more graceful way to handle this? I think it makes sense to build a new EventBase rather than mutating the original event, would it make sense to make a copy of the `event_dict` parameter (even though it is already a copy, as you mentioned) and use that to build the EventBase rather than the mutated parameter?

tests/test_federation.py
H-Shay: They do run but only if you run them individually, which I can now see how that would be a problem. I didn't realized I needed to add `test` to the beginning of the method, thank you!

tests/test_federation.py
H-Shay: Got it thanks!

DMRobertson:
@H-Shay is this ready for another look?

DMRobertson (commented):
Looks reasonable to me, but I'd appreciate a second review from someone who's familiar with the event format. Maybe @richvdh?

changelog.d/11530.misc
```diff
@@ -0,0 +1 @@
+Add functionality to strip the unsignedData object of unauthorized fields in events arriving over federation.
```
richvdh, line 1: I'd say this is a bugfix!

> Fix a long-standing issue which could cause Synapse to incorrectly accept data in the `unsigned` field of events received over federation.

Or words to that effect.

richvdh (commented):
I think this is ok as far as it goes, but it misses events that we *pull* from the remote server, which are currently subject to the same bugs.

Could we just do the filtering at the point where we first turn the `dict` we receive in the JSON into an `EventBase` (ie, `event_from_pdu_json`)? I think that would be easier, as well as making sure we catch all the places it's a problem.

H-Shay:
> I think this is ok as far as it goes, but it misses events that we _pull_ from the remote server, which are currently subject to the same bugs.
> 
> Could we just do the filtering at the point where we first turn the `dict` we receive in the JSON into an `EventBase` (ie, `event_from_pdu_json`)? I think that would be easier, as well as making sure we catch all the places it's a problem.

Sure thing, that's a much better idea. Thanks for the suggestion!

synapse/federation/federation_base.py
```diff
@@ -238,6 +238,10 @@ def event_from_pdu_json(
     # origin, etc etc)
     assert_params_in_dict(pdu_json, ("type", "depth"))
 
+    # Strip any unauthorized values from "unsigned" if they exist
+    if "unsigned" in pdu_json and pdu_json["unsigned"] != {}:
+        pdu_json = _strip_unsigned_values(pdu_json)
```
richvdh, line 243: it's worth noting that, because `_strip_unsigned_values` modifies the actual dict that `pdu_json` refers to, this is the same as:

```suggestion
        _strip_unsigned_values(pdu_json)
```

Indeed, I think it's confusing to have `_strip_unsigned_values` return the dict at all, since it gives the incorrect impression that it returns a *different* dict, rather than modifying the original.

richvdh (commented):
thank you! a couple of minor comments...

synapse/federation/federation_base.py
H-Shay: That makes sense, thank you for pointing that out!

tests/test_federation.py
```diff
@@ -276,3 +278,91 @@ def test_cross_signing_keys_retry(self):
             "ed25519:" + remote_self_signing_key in self_signing_key["keys"].keys(),
         )
         self.assertTrue(remote_self_signing_key in self_signing_key["keys"].values())
+
+
+class StripUnsignedFromEventsTestCase(MessageAcceptTests):
```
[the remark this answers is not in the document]
H-Shay: Thanks for pointing this out, it's helpful to know.

synapse/federation/federation_base.py
```diff
@@ -245,3 +249,27 @@ def event_from_pdu_json(pdu_json: JsonDict, room_version: RoomVersion) -> EventB
...
+
+    if pdu_dict["type"] == "m.room.member":
+        whitelist = ["knock_room_state", "invite_room_state", "age"]
+    else:
+        whitelist = ["age"]
+
+    filtered_unsigned = {k: v for k, v in unsigned.items() if k in whitelist}
+    pdu_dict["unsigned"] = filtered_unsigned
+
+    return pdu_dict
```
richvdh, line 275: let's remove this: It's no longer used, and I think that returning a value makes it look like we might be returning something different from the input.

```suggestion
```

likewise, s/`return pdu_dict`/`return`/ above, and you'll need to change the annotation to `-> None`.

richvdh (approved):
lgtm otherwise!

synapse/federation/federation_base.py
H-Shay: Right makes sense, also thank you for the quick and enlightening review!

PR #11531, opened 2021-12-07
Add receipt types constant

clokep:
Trying to get a bit familiar with the receipts code, figured I would:

* Add a constant for `m.read`.
* Finish up some missing type hints.

### DIFF
--- a/synapse/storage/databases/main/receipts.py
+++ b/synapse/storage/databases/main/receipts.py
@@ -18,6 +18,7 @@ from typing import TYPE_CHECKING, Any, Dict, Iterable, List, Optional, Tuple
 
 from twisted.internet import defer
 
+from synapse.api.constants import ReceiptTypes
 from synapse.replication.slave.storage._slaved_id_tracker import SlavedIdTracker
 from synapse.replication.tcp.streams import ReceiptsStream
 from synapse.storage._base import SQLBaseStore, db_to_json, make_in_list_sql_clause
@@ -88,7 +89,7 @@ class ReceiptsWorkerStore(SQLBaseStore):
 
     @cached()
     async def get_users_with_read_receipts_in_room(self, room_id):
-        receipts = await self.get_receipts_for_room(room_id, "m.read")
+        receipts = await self.get_receipts_for_room(room_id, ReceiptTypes.READ)
         return {r["user_id"] for r in receipts}
 
     @cached(num_args=2)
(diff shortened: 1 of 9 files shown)


clokep:
There's a few spots where we add/remove fields from JSON objects with the `m.read` flag -- it is unclear to me if those should use the same constant though. Any thoughts?

richvdh:
> There's a few spots where we add/remove fields from JSON objects with the m.read flag -- it is unclear to me if those should use the same constant though. Any thoughts?

I'd have thought we should - they're referring to the same concept? Do you have examples?

synapse/storage/databases/main/receipts.py
```diff
@@ -78,7 +89,7 @@ def __init__(self, database: DatabasePool, db_conn, hs: "HomeServer"):
             "ReceiptsRoomChangeCache", self.get_max_receipt_stream_id()
         )
 
-    def get_max_receipt_stream_id(self):
+    def get_max_receipt_stream_id(self) -> int:
         """Get the current max stream ID for receipts stream
 
         Returns:
```
richvdh, line 95: might be nice to remove this

richvdh (approved):
lgtm!

clokep:
> > There's a few spots where we add/remove fields from JSON objects with the m.read flag -- it is unclear to me if those should use the same constant though. Any thoughts?
> 
> I'd have thought we should - they're referring to the same concept? Do you have examples?

The REST layer pulls some bits out, for example:

https://github.com/matrix-org/synapse/blob/086adf55f9b3624853977eb470fb6b61faec32ce/synapse/rest/client/read_marker.py#L51

I think we also generate JSON payloads with the field for sync (and maybe push).

It seems there's also the `ReadReceiptEventFields` which I could possibly add something too...

richvdh:
> The REST layer pulls some bits out, for example:
> 
> https://github.com/matrix-org/synapse/blob/086adf55f9b3624853977eb470fb6b61faec32ce/synapse/rest/client/read_marker.py#L51

That certainly looks like it should be using the same constant to me.

synapse/storage/databases/main/receipts.py
clokep: Good catch! I found one other too!

clokep:
> > The REST layer pulls some bits out, for example:
> > https://github.com/matrix-org/synapse/blob/086adf55f9b3624853977eb470fb6b61faec32ce/synapse/rest/client/read_marker.py#L51
> 
> That certainly looks like it should be using the same constant to me.

I've updated a few more places. 👍

PR #11538, opened 2021-12-08
Add experimental support for MSC3202: allowing application services to masquerade as specific devices.

reivilibre:
- [x] add tests

somewhat commit-by-commit friendly?


synapse/storage/databases/main/devices.py
```diff
@@ -120,6 +122,26 @@ async def get_device(self, user_id: str, device_id: str) -> Dict[str, Any]:
             desc="get_device",
         )
 
+    async def get_device_opt(
```
anoadragon453, line 125: I think it'd be nicer if we just switched `get_device` to return an `Optional`, and then change up the callers to look for `None` instead of a `StoreError`.

That would also fix up callers that seem to currently 500 if a device isn't found(!)

tests/api/test_auth.py
```diff
@@ -210,6 +210,67 @@ def test_get_user_by_req_appservice_valid_token_bad_user_id(self):
...
+        app_service = Mock(
+            token="foobar", url="a_url", sender=self.test_user, ip_range_whitelist=None
+        )
+        app_service.is_interested_in_user = Mock(return_value=True)
+        self.store.get_app_service_by_token = Mock(return_value=app_service)
+        # This just needs to return a truth-y value.
+        self.store.get_user_by_id = simple_async_mock({"is_guest": False})
+        self.store.get_user_by_access_token = simple_async_mock(None)
+        # This also needs to just return a truth-y value
+        self.store.get_device_opt = simple_async_mock(None)
```
anoadragon453, line 264: This one needs to be falsey.

tests/api/test_auth.py
```diff
@@ -210,6 +210,67 @@ def test_get_user_by_req_appservice_valid_token_bad_user_id(self):
...
+
+        request = Mock(args={})
```
anoadragon453, line 266: Hmm, I see the argument for consistency, but does `args={}` actually do anything? Likewise above.

anoadragon453 (changes requested):
Some minor things. Looks good to me otherwise! Was a pleasure to review commit-by-commit.

Handy link to [the MSC](https://github.com/matrix-org/matrix-doc/pull/3202).

synapse/storage/databases/main/devices.py
reivilibre: Span out to #11565 .

tests/api/test_auth.py
reivilibre: `args` is where the access token, desired user ID and desired device ID go. Did I misunderstand? Is there a nicer way to do this?

anoadragon453: Ohhh, hah. I thought `args` was some kind of option that Mock accepted, rather than just laying out the spec for `SynapseRequest`. That makes much more sense.

anoadragon453 (approved):
lgtm after merging in develop!

synapse/api/auth.py
```diff
@@ -155,24 +155,34 @@ async def get_user_by_req(
...
-                        device_id="dummy-device",  # stubbed
+                        device_id="dummy-device"
+                        if device_id is None
+                        else device_id,  # stubbed
                     )
 
-                requester = create_requester(user_id, app_service=app_service)
+                requester = create_requester(
+                    user_id, app_service=app_service, device_id=device_id
+                )
```
turt2live, line 177: for some reason this isn't enough to cause `/account/whoami` to return the device ID back.

turt2live: nevermind, I was given the wrong experimental flag.

turt2live:
Just wanted to drop by and say this works :)

PR #7021, opened 2020-03-02
Fix inconsistent handling of upper and lower cases of email addresses.

dklimpel:
Fix #7016

### Pull Request Checklist

* [x] Pull request is based on the develop branch
* [x] Pull request includes a [changelog file](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md#changelog). The entry should:
  - Be a short description of your change which makes sense to users. "Fixed a bug that prevented receiving messages from other servers." instead of "Moved X method from `EventStore` to `EventWorkerStore`.".
  - Use markdown where necessary, mostly for `code blocks`.
  - End with either a period (.) or an exclamation mark (!).
  - Start with a capital letter.
* [x] Pull request includes a [sign off](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md#sign-off)
* [x] Code style is correct (run the [linters](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md#code-style))

Signed-off-by: Dirk Klimpel dirk@klimpel.org

PR #11547, opened 2021-12-09
Ensure emails are canonicalized before fetching associated user.

### LINKED ISSUE #11458: Emails are not canonicalised on /_matrix/client/*/pushers/set
### Description

Emails sent to `/pushers/set` are not canonicalised, while they probably should.

### Steps to reproduce

- choose an email with an uppercase letter in it, e.g. `FooBar@example.com`
- set that email to a user via the admin API. The email will end up normalized in the DB
- try to set an email pusher with the same email. It should work but doesn't and returns with an `Email not found` error

### Version information

- **Homeserver**: other
- **Version**: 1.47.0
- **Install method**: Docker/Kubernetes
- **Platform**: N/A

---

Here, the `pushkey` should definitely be canonicalised for emails:

https://github.com/matrix-org/synapse/blob/8c7a531e277f98ac6b7981b9738649f3a70feb94/synapse/push/pusherpool.py#L115-L118

using `canonicalise_email`:

https://github.com/matrix-org/synapse/blob/8c7a531e277f98ac6b7981b9738649f3a70feb94/synapse/util/threepids.py#L66-L69

clokep:
Updates two callers of `get_user_id_by_threepid` to canonicalize the email argument. I audited other callers and ensured that they were fine (by calling `canonicalise_email` or `validate_email`).

Fixes #11458

### DIFF
--- a/changelog.d/11547.bugfix
+++ b/changelog.d/11547.bugfix
@@ -1,0 +1,1 @@
+Fix a bug introduced in Synapse 1.17.0 where a pusher created for an email with capital letters would fail to be created.
--- a/synapse/push/pusherpool.py
+++ b/synapse/push/pusherpool.py
@@ -27,6 +27,7 @@ from synapse.push.pusher import PusherFactory
 from synapse.replication.http.push import ReplicationRemovePusherRestServlet
 from synapse.types import JsonDict, RoomStreamToken
 from synapse.util.async_helpers import concurrently_execute
+from synapse.util.threepids import canonicalise_email
 
 if TYPE_CHECKING:
     from synapse.server import HomeServer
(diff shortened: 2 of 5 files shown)


changelog.d/11547.bugfix
```diff
@@ -0,0 +1 @@
+Fix a bug introduced in Synapse 1.17.0 where a pusher created for an email with capital letters would fail to be created.
```
clokep, line 1: #7021 added the canonical email stuff (I think) which was in Synapse 1.17.0, I'm not 100% sure what the behavior was before that though.

squahtx (approved):
lgtm!

PR #11549, opened 2021-12-09
Add type hints to `synapse/storage/databases/main/e2e_room_keys.py`

squahtx:


synapse/handlers/e2e_room_keys.py
```diff
@@ -58,7 +61,9 @@ async def get_room_keys(
         version: str,
         room_id: Optional[str] = None,
         session_id: Optional[str] = None,
-    ) -> List[JsonDict]:
+    ) -> Dict[
+        Literal["rooms"], Dict[str, Dict[Literal["sessions"], Dict[str, RoomKey]]]
+    ]:
```
squahtx, line 66: The previous annotation was incorrect. The caller and tests expect a JsonDict to be returned:

https://github.com/matrix-org/synapse/blob/0cc3bf97b4399234cf20f52ae4c09d03661225ff/synapse/rest/client/room_keys.py#L199-L223
https://github.com/matrix-org/synapse/blob/0cc3bf97b4399234cf20f52ae4c09d03661225ff/tests/handlers/test_e2e_room_keys.py#L404-L409

synapse/storage/databases/main/e2e_room_keys.py
```diff
@@ -92,31 +114,39 @@ async def add_e2e_room_keys(self, user_id, version, room_keys):
...
+            session_id: Optional. the session whose room_key we're querying, if any.
                 If specified, we also require the room_id to be specified.
                 If not specified, we return all the keys in this version of
                 the backup (or for the specified room)
 
         Returns:
-            A list of dicts giving the session_data and message metadata for
-            these room keys.
+            A dict giving the session_data and message metadata for these room keys.
+            `{"rooms": {room_id: {"sessions": {session_id: room_key}}}}`
```
squahtx, line 141: The previous annotation was incorrect. See the PR comment above

synapse/storage/databases/main/e2e_room_keys.py
```diff
@@ -271,23 +313,25 @@ async def delete_e2e_room_keys(
...
         txn.execute(
             "SELECT MAX(version) FROM e2e_room_keys_versions "
             "WHERE user_id=? AND deleted=0",
             (user_id,),
         )
-        row = txn.fetchone()
-        if not row:
+        row = cast(Tuple[Optional[int]], txn.fetchone())
+        if row[0] is None:
             raise StoreError(404, "No current backup version")
```
squahtx, line 324: The previous code was buggy. `SELECT MAX(...)` will return a single row with `NULL` if it's done over 0 rows.

synapse/handlers/e2e_room_keys.py
DMRobertson: Any idea why mypy didn't spot this? I guess the tests aren't checked, but I'd hope room_keys was

synapse/storage/databases/main/e2e_room_keys.py
DMRobertson: oh, because MAX(version) will aggregate to NULL if there are no rows permitted by the where clause? Might that be worth a comment?

synapse/storage/databases/main/e2e_room_keys.py
```diff
@@ -328,26 +374,26 @@ def _get_e2e_room_keys_version_info_txn(txn):
...
                 "SELECT MAX(version) FROM e2e_room_keys_versions WHERE user_id=?",
                 (user_id,),
             )
-            current_version = txn.fetchone()[0]
+            current_version = cast(Tuple[Optional[int]], txn.fetchone())[0]
             if current_version is None:
-                current_version = "0"
+                current_version = 0
 
             new_version = str(int(current_version) + 1)
```
DMRobertson, line 398: Do we still need the call to `int` here?

Edit: oh, there's some weird strings/int stuff going on here. I think I expanded on this in another comment.

synapse/types.py
```diff
@@ -59,9 +59,11 @@
 StateMap = Mapping[StateKey, T]
 MutableStateMap = MutableMapping[StateKey, T]
 
-# the type of a JSON-serialisable dict. This could be made stronger, but it will
-# do for now.
+# JSON types. These could be made stronger, but will do for now.
+# A JSON-serialisable dict.
 JsonDict = Dict[str, Any]
+# A JSON-serialisable object.
+JsonObject = object
```
DMRobertson, line 66: This surprised me. I guessed that "JsonObject" and "JsonDict" would be one and the same, because javascript/json calls its mapping types objects (c.f. json.org):

> An object is an unordered set of name/value pairs. An object begins with {left brace and ends with }right brace. Each name is followed by :colon and the name/value pairs are separated by ,comma.

Perhaps `JsonSerializable`?

synapse/handlers/e2e_room_keys.py
```diff
@@ -58,7 +61,9 @@ async def get_room_keys(
         version: str,
```
DMRobertson, line 61: [Alarm bells](https://youtu.be/HcFd5j1cios?t=24).

```
synapse=> \d e2e_room_keys_versions
         Table "public.e2e_room_keys_versions"
  Column   |   Type   | Collation | Nullable | Default 
-----------+----------+-----------+----------+---------
 user_id   | text     |           | not null | 
 version   | bigint   |           | not null | 
 algorithm | text     |           | not null | 
 auth_data | text     |           | not null | 
 deleted   | smallint |           | not null | 0
 etag      | bigint   |           |          | 
Indexes:
    "e2e_room_keys_versions_idx" UNIQUE, btree (user_id, version)
```

version is a bigint in the db, so I'm surprised to see a string here. What's going on?

DMRobertson: Oh. It looks like version comes from a query parameter.

https://github.com/matrix-org/synapse/blob/d1f1b46c2cfe612790aebdd54765953951c94e45/synapse/rest/client/room_keys.py#L197

Can we change this to `parse_integer` at the rest layer, then pass an integer to and from the handler and the DB layer?

DMRobertson (commented):
Many thanks for this!

Not sure about JsonObject (see below), and if you're up for it I'd love to see the str/int version confusion sorted out at the rest layer.

synapse/types.py
squahtx: Ah, bad naming on my part. I was thinking `object`, as in the class that's at the base of the type hierarchy.

synapse/handlers/e2e_room_keys.py
squahtx: It looks like `e2e_room_keys_handler.get_room_keys` has a `@trace` decorator, which isn't annotated

DMRobertson: Interesting! I think @H-Shay is taking a look at annotating that file, maybe this will show up the problem.

synapse/types.py
clokep: +1 that I found this confusing a JSON object == a dict in Python!

synapse/storage/databases/main/e2e_room_keys.py
squahtx: I noticed that this line is strictly incorrect: `new_version` ought to be an `int` here because we insert it into a bigint column in the database.

synapse/handlers/e2e_room_keys.py
squahtx: Normally I'd agree: stringly-typed programming is a code smell.

However the matrix spec says that `version` is an opaque string identifying a backup: https://spec.matrix.org/v1.1/client-server-api/#post_matrixclientv3room_keysversion
`version`s are only issued when a backup is uploaded, by the homeserver. As such, the correct response to a non-numeric version is a 404 (no such backup) rather than the 400 `parse_integer` would raise.

Because the spec says `version`s are opaque identifiers, I'd consider versions-being-integers to be an implementation detail. Only the code responsible for issuing them, `EndToEndRoomKeyStore`, has to know about it. Keeping that detail contained within the data store does have a benefit: if we were to decide that we wanted to use, say, sha256 hashes instead of incrementing integers, we'd only need to update the data store and not the handler and rest layers.

synapse/types.py
squahtx: Renamed it!

synapse/storage/databases/main/e2e_room_keys.py
squahtx: Good spot, the cast to `int` is redundant.

synapse/handlers/e2e_room_keys.py
clokep: It might make sense to include a comment somewhere in the store method saying why we convert it to an int there?

squahtx: Having said that, the handling for non-numeric versions is not pretty code.

squahtx:
I've went through and added the missing error handling for non-numeric version strings.
The code should behave the same as if a numeric, non-existant version was passed.

synapse/handlers/e2e_room_keys.py
squahtx: > It might make sense to include a comment somewhere in the store method saying why we convert it to an int there?

Good call. I think it ought to be a class-level docstring describing the design choice, since it applies to every method.

synapse/storage/databases/main/e2e_room_keys.py
```diff
@@ -301,8 +325,14 @@ async def delete_e2e_room_keys(
                 If not specified, we delete all the keys in this version of
                 the backup (or for the specified room)
         """
+        try:
+            version_int = int(version)
+        except ValueError:
+            # Our versions are all ints so if we can't convert it to an integer,
+            # it doesn't exist.
+            return
```
DMRobertson, line 342: Not sure about this---if someone provides e.g. an empty string for the version and tries to delete stuff with that version, they might think they've deleted something when in fact they've deleted nothing.

But... maybe the point is that this should be handled at the handler/rest levels?

synapse/storage/databases/main/e2e_room_keys.py
```diff
@@ -57,12 +57,18 @@ async def update_e2e_room_key(
         Raises:
             StoreError
         """
+        try:
+            version_int = int(version)
+        except ValueError:
+            # Our versions are all ints so if we can't convert it to an integer,
+            # it doesn't exist.
+            raise StoreError(404, "No row found")
```
DMRobertson, line 74: Maybe "invalid version" for the exception message?

DMRobertson: Ditto on line 99 below

synapse/storage/databases/main/e2e_room_keys.py
```diff
@@ -38,6 +38,15 @@ class RoomKey(TypedDict):
 
 
 class EndToEndRoomKeyStore(SQLBaseStore):
+    """The store for end to end room key backups.
+
+    See https://spec.matrix.org/v1.1/client-server-api/#server-side-key-backups
+
+    As per the spec, backups are identified by an opaque version string. Internally,
+    version identifiers are assigned using incrementing integers. Non-numeric version
+    strings are treated as if they do not exist, since we would have never issued them.
```
DMRobertson, line 47: I think I said this on the changes in the previous commit, but worried this might let some client requests fail silently while otherwise indicating success.

synapse/handlers/e2e_room_keys.py
DMRobertson: That's a surprise---thanks for digging it up!

synapse/storage/databases/main/e2e_room_keys.py
squahtx: The intent of this code was to match the behaviour when a non-existant numeric version is given. `simple_delete` below will always succeed, even if it deletes 0 rows. Only the `_one()` variants will raise a 404 error (which ought to correspond to all the places where I do raise a 404).

Granted, the existing behaviour may be undesirable, but I'd rather not be changing the (intended) API contract as part of this PR.

DMRobertson: That's very fair, and you're right to call me out for scope creep!

synapse/storage/databases/main/e2e_room_keys.py
DMRobertson: Thanks, appreciate that!

DMRobertson (approved):

PR #11553, opened 2021-12-10
Add details for how to set up TURN behind NAT

AndrewFerr:
Signed-off-by: Andrew Ferrazzutti <fair@miscworks.net>

### Pull Request Checklist

    * (Not sure if this is required, since this is only a documentation change. I did run `mdbook build` and `mdbook serve` with no issues, however.)

### DIFF
--- a/docs/turn-howto.md
+++ b/docs/turn-howto.md
@@ -15,8 +15,8 @@ The following sections describe how to install [coturn](<https://github.com/cotu
 
 For TURN relaying with `coturn` to work, it must be hosted on a server/endpoint with a public IP.
 
-Hosting TURN behind a NAT (even with appropriate port forwarding) is known to cause issues
-and to often not work.
+Hosting TURN behind a NAT requires port forwaring and for the NAT gateway to have a public IP.
+However, even with appropriate configuration, NAT is known to cause issues and to often not work.
 
 ## `coturn` setup
 
@@ -103,6 +103,21 @@ This will install and start a systemd service called `coturn`.
     denied-peer-ip=192.168.0.0-192.168.255.255
     denied-peer-ip=172.16.0.0-172.31.255.255
 
+    # recommended additional local peers to block, to mitigate external access to internal services.
+    # https://www.rtcsec.com/article/slack-webrtc-turn-compromise-and-bug-bounty/#how-to-fix-an-open-turn-relay-to-address-this-vulnerability
+    no-multicast-peers
+    denied-peer-ip=0.0.0.0-0.255.255.255
+    denied-peer-ip=100.64.0.0-100.127.255.255
+    denied-peer-ip=127.0.0.0-127.255.255.255
+    denied-peer-ip=169.254.0.0-169.254.255.255
+    denied-peer-ip=192.0.0.0-192.0.0.255
+    denied-peer-ip=192.0.2.0-192.0.2.255
+    denied-peer-ip=192.88.99.0-192.88.99.255
+    denied-peer-ip=198.18.0.0-198.19.255.255
+    denied-peer-ip=198.51.100.0-198.51.100.255
+    denied-peer-ip=203.0.113.0-203.0.113.255
+    denied-peer-ip=240.0.0.0-255.255.255.255
+
     # special case the turn server itself so that client->TURN->TURN->client flows work
     allowed-peer-ip=10.0.0.1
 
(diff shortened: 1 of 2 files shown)


AndrewFerr:
Since I managed to get my coturn working behind an OpenVPN subnet, I figured I'd share the steps I used (especially since the documentation asked for it :slightly_smiling_face:)

I assume there isn't anything fundamentally unique about OpenVPN's NATting, and that the settings I used should work for any kind of NAT.

This PR also includes updates some other TURN config information/warnings:
* extra local peers to block, based on https://www.rtcsec.com/article/slack-webrtc-turn-compromise-and-bug-bounty/
* a notice about Let's Encrypt certificates with TLS/DTLS, based on https://github.com/vector-im/element-android/issues/1533

These can be split into separate commits if preferred.

docs/turn-howto.md
```diff
@@ -129,26 +144,84 @@ This will install and start a systemd service called `coturn`.
     We recommend that you only try to set up TLS/DTLS once you have set up a
     basic installation and got it working.
 
+    NB: If your TLS certificate was provided by Let's Encrypt, TLS/DTLS will
+    not work with any Matrix client that uses Chromium's WebRTC library (such
+    as Element Android). See [this case](https://github.com/vector-im/element-android/issues/1533) for more details.
```
richvdh, line 149: I'm struggling to believe this is true.  Matrix.org's TLS cert is provided by LE, and it seems to work? Doesn't Google Chrome use the same WebRTC library?

The issue just says "socket buffer operation error", which doesn't really tell me a lot. I have no idea where that error appears or what its effect is.

docs/turn-howto.md
```diff
@@ -129,26 +144,84 @@ This will install and start a systemd service called `coturn`.
...
+
+    ```
+    net.ipv4.ip_forward=1
+    net.ipv6.conf.all.forwarding=1
+    ```
 
-    If you get this working, let us know!
+    (The `all` in any of the the IPv6 paths/settings mentioned above may be
+    substituted for the name of any specific network interface that must
+    forward traffic.)
```
richvdh, line 224: This is pretty long now. It might be a good idea to move it out to a separate section "configuring a firewall to forward traffic" or something?

docs/turn-howto.md
```diff
@@ -129,26 +144,84 @@ This will install and start a systemd service called `coturn`.
...
+    # same, but for ipv6
+    ip6tables -t nat -A PREROUTING -i $EXTERNAL_INTERFACE_NAME -p tcp -m multiport --dports 3478,3479,5349,5350 -j DNAT --to-destination $INTERNAL_TURNSERVER_IPv6_ADDRESS
+    ip6tables -t nat -A PREROUTING -i $EXTERNAL_INTERFACE_NAME -p udp -m multiport --dports 3478,3479,5349,5350,49152:65535 -j DNAT --to-destination $INTERNAL_TURNSERVER_IPv6_ADDRESS
+    ip6tables -t nat -A PREROUTING -i $EXTERNAL_INTERFACE_NAME -p tcp -m multiport --dports 5222,5269 -j DNAT --to-destination $INTERNAL_TURNSERVER_IPv6_ADDRESS
+    ```
+
+    For this to work, the NAT server must be configured to forward IPv4/IPv6
+    traffic. To check if forwarding is enabled, run
+    `cat /proc/sys/net/ipv4/ip_forward` and `cat /proc/sys/net/ipv6/conf/all/forwarding`
+    and confirm they both print `1`. If not, they may be set at runtime like so:
```
richvdh, line 207: surely, a NAT server which *doesn't* forward traffic isn't a whole lot of use as a NAT server?

richvdh (commented):
thanks for this! a few comments.

docs/turn-howto.md
reivilibre: Have to agree, LP.net uses LE and it seems to work without complaint.

AndrewFerr: Yes, there can indeed be connectivity issues when using one of the mobile Element clients. The same issue affects the Jitsi Meet app on Android.

I'll add links to the Element iOS & Jitsi Meet issues:
https://github.com/vector-im/element-ios/issues/2712
https://github.com/jitsi/jitsi-meet/issues/6383#issuecomment-632659991

From my testing, calls fail to connect only if at least one client is Element Android, call traffic is required to go through the TURN server, and only `turns` connections are available. (If unencrypted TURN URIs are available, they continue to work as fallbacks.)

To test forcing traffic through the TURN server (as opposed to using direct p2p connections, which AIUI is what STUN tries to set up--I'm still foggy on the full details of this), I made calls to/from a work laptop on a corporate VPN. To ensure encryption was used, I either disabled all `turn:` entries from my homeserver's `turn_uris`, or I left them enabled, but blocked traffic to/from ports 3478-3479 for my laptop & (Android) mobile, with either device serving as the other end of a call with the work laptop.

The results of this were:
* Calls between two web browsers (personal & work laptops) were successful, whether or not `turn:` URIs were disabled, or ports 3478-3479 were blocked. Running `tcpdump` on my personal laptop showed traffic flowing through port 5349.
* Calls between my phone & the work laptop failed unless I enabled `turn:` URIs or unblocked the ports.

Critically, the results were identical even when making calls from a matrix.org account, and thus while using its TURN server & its certificate.

(If it makes a difference, turn.matrix.org's `turns:` URI uses TCP on port 443, not 5349--but if anything, that should help connections be more reliable, as no firewall should block TLS traffic on port 443.)

Results were similar when testing my self-hosted Jitsi: calls from the Android app were unable to get p2p connections with `turns`, and instead fell back to using the videobridge, while calls between two web browsers could always use p2p.

(FWIW, Jitsi Meet is what informed me of the possibility of non-TURN p2p connections: calls between my personal laptop & my phone on mobile data get tagged as "(p2p)", whereas calls involving the work laptop behind its VPN were labelled as "(p2p)(turn)", unless the videobridge was used as a fallback. Any setup that gets marked as "(p2p)" in Jitsi Meet always results in a successful call in Element.)

docs/turn-howto.md
AndrewFerr: Good idea, I'll get to it.

docs/turn-howto.md
AndrewFerr: True, but this is also provided as a light guide for how to set up a NAT server that a homserver admin may need to configure themselves (such as an OpenVPN server). But if this is out-of-scope for a TURN setup article, I'm fine with removing this.

richvdh: If I were setting up a NAT server, I certainly wouldn't make forwarding *incoming* rtc traffic the place I would start. So yeah, I think this is out-of-scope.

docs/turn-howto.md
richvdh: ookay, thanks for the detailed information, but could you put it all on the relevant issue? (https://github.com/vector-im/element-android/issues/1533, I guess.) It's valuable research, which is going to get lost in this PR thread, and it needs to be reported to the people who maintain the software with the bugs!

(aside: I note that https://github.com/vector-im/element-ios/issues/2712#issuecomment-869274355 says it's fine if you correctly publish the intermediate certificates - is that not correct)?

We can consider adding a warning to `turn-howto.md`, but it's an orthogonal problem to NAT - please could you remove it from this PR for now?

AndrewFerr: Removed, and commented in https://github.com/vector-im/element-android/issues/1533#issuecomment-996409199

docs/turn-howto.md
AndrewFerr: On second thought, I removed this entirely, as setting up your own NAT server isn't relevant for knowing how to get TURN working under a pre-existing NAT.

docs/turn-howto.md
AndrewFerr: Agreed. Removed.

docs/turn-howto.md
```diff
@@ -134,21 +150,35 @@ This will install and start a systemd service called `coturn`.
...
 
-    If you want to try it anyway, you will at least need to tell coturn its
-    external IP address:
+    You may optionally limit the TURN server to listen only on the local
+    address that is mapped by NAT to the external address:
 
     ```
-    external-ip=192.88.99.1
+    listening-ip=INTERNAL_TURNSERVER_IPv4_ADDRESS
+    external-ip=EXTERNAL_NAT_IPv4_ADDRESS/INTERNAL_TURNSERVER_IPv4_ADDRESS
```
richvdh, line 166: this `external/internal` notation is only required if there is more than one external IP address, right?

docs/turn-howto.md
```diff
@@ -234,6 +261,17 @@ Here are a few things to try:
    Try removing any AAAA records for your TURN server, so that it is only
    reachable over IPv4.
 
+ * If your TURN server is behind a NAT:
```
richvdh, line 264:
```suggestion
 * If your TURN server is behind NAT:
```

docs/turn-howto.md
```diff
@@ -234,6 +261,17 @@ Here are a few things to try:
...
+
+    * double-check that your NAT gateway is correctly forwarding all TURN
+      ports to the NAT-internal address of your TURN server. If advertising
```
richvdh, line 267: should we expand on "all TURN ports"? in my experience, people remember to forward 3478 and 5349, but forget about the UDP relay ports (49152-65535).

docs/turn-howto.md
AndrewFerr: That should be the case. I'll test what happens when using multiple `listening-ip`s with only one `external-ip`. If all public->private address pairings are made automatically, I'll remove line 166.

AndrewFerr: After some testing, it looks like the config is much more flexible than I thought. The `external/internal` notation isn't needed with multiple active listening IPs, or even multiple relay IPs--as long as one paring works, just the `external` notation is enough.

I'll trim down this section accordingly.

docs/turn-howto.md
AndrewFerr: Listed the ports in the latest commit.

docs/turn-howto.md
AndrewFerr: Done.

richvdh (approved):
thank you for your perseverance here!

PR #11556, opened 2021-12-10
Squah/type hint synapse logging context

squahtx:
### Pull Request Checklist


stubs/txredisapi.pyi
```diff
@@ -17,20 +17,21 @@
 from typing import Any, List, Optional, Type, Union
 
 from twisted.internet import protocol
+from twisted.internet.defer import Deferred
 
 class RedisProtocol(protocol.Protocol):
     def publish(self, channel: str, message: bytes): ...
-    async def ping(self) -> None: ...
-    async def set(
+    def ping(self) -> "Deferred[None]": ...
```
squahtx, line 24: After looking at the txredisapi source, I'm fairly sure these aren't async methods and return `Deferred`s instead.

https://github.com/IlyaSkriblovsky/txredisapi/blob/c89a49829d1c48baa10d03c41d39c261c02ad6e1/txredisapi.py#L655-L659

https://github.com/IlyaSkriblovsky/txredisapi/blob/c89a49829d1c48baa10d03c41d39c261c02ad6e1/txredisapi.py#L577

synapse/http/federation/matrix_federation_agent.py
```diff
@@ -309,12 +310,14 @@ def __init__(
...
-    def connect(self, protocol_factory: IProtocolFactory) -> defer.Deferred:
+    def connect(
+        self, protocol_factory: IProtocolFactory
+    ) -> "defer.Deferred[IProtocol]":
         """Implements IStreamClientEndpoint interface"""
 
         return run_in_background(self._do_connect, protocol_factory)
 
-    async def _do_connect(self, protocol_factory: IProtocolFactory) -> None:
+    async def _do_connect(self, protocol_factory: IProtocolFactory) -> IProtocol:
```
squahtx, line 320: The return value comes from `make_deferred_yieldable`, which is why mypy didn't complain before.

synapse/logging/context.py
```diff
@@ -711,16 +723,61 @@ def nested_logging_context(suffix: str) -> LoggingContext:
...
+    f: Callable[..., Awaitable[R]],
+) -> Callable[..., "defer.Deferred[R]"]:
+    # The `type: ignore[misc]` above suppresses
+    # "Overloaded function signatures 1 and 2 overlap with incompatible return types"
+    ...
+
+
+@overload
+def preserve_fn(f: Callable[..., R]) -> Callable[..., "defer.Deferred[R]"]:
+    ...
```
squahtx, line 740: Another place where `ParamSpec` would help.

synapse/util/async_helpers.py
```diff
@@ -234,6 +236,59 @@ def yieldable_gather_results(
...
+    ],
+    consumeErrors: bool = ...,
+) -> "defer.Deferred[Tuple[T1, T2, T3]]":
+    ...
+
+
+def gather_results(  # type: ignore[misc]
+    deferredList: Tuple["defer.Deferred[T1]", ...],
+    consumeErrors: bool = False,
+) -> "defer.Deferred[Tuple[T1, ...]]":
```
DMRobertson, line 280: I'm getting C++ [parameter pack](https://en.cppreference.com/w/cpp/language/parameter_pack) vibes here!

I guess the overloads do the trick.

synapse/federation/federation_server.py
```diff
@@ -360,13 +359,13 @@ async def _handle_incoming_transaction(
         # want to block things like to device messages from reaching clients
         # behind the potentially expensive handling of PDUs.
         pdu_results, _ = await make_deferred_yieldable(
-            defer.gatherResults(
-                [
+            gather_results(
+                (
```
DMRobertson, line 363: If I understand rightly, the difference here is that we're returning a Tuple (which we can annotate heterogeneously) as opposed to a list? But otherwise `pdu_results, _` is identical

synapse/logging/context.py
```diff
@@ -778,7 +778,10 @@ def run_in_background(f, *args, **kwargs) -> defer.Deferred:
     return res
 
 
-def make_deferred_yieldable(deferred):
+T = TypeVar("T")
+
+
+def make_deferred_yieldable(deferred: "defer.Deferred[T]") -> "defer.Deferred[T]":
     """Given a deferred (or coroutine), make it follow the Synapse logcontext
     rules:
```
DMRobertson, line 845: Judging by the doctstring I'd've thought the annotation ought to be `Union["defer.Deferred[T]", Awaitable[T]]`.

This might have cropped up in https://github.com/matrix-org/synapse/pull/10980 --- I remember something like this being painful.

synapse/logging/context.py
```diff
@@ -720,7 +733,30 @@ def g(*args, **kwargs):
     return g
 
 
-def run_in_background(f, *args, **kwargs) -> defer.Deferred:
+@overload
+def run_in_background(  # type: ignore[misc]
+    f: Callable[..., Awaitable[R]], *args: Any, **kwargs: Any
+) -> "defer.Deferred[R]":
+    # The `type: ignore[misc]` above suppresses
+    # "Overloaded function signatures 1 and 2 overlap with incompatible return types"
```
DMRobertson, line 761: Does mypy still use this overload with the type-ignore?

synapse/logging/context.py
```diff
@@ -720,7 +733,30 @@ def g(*args, **kwargs):
...
+    f: Callable[..., R], *args: Any, **kwargs: Any
+) -> "defer.Deferred[R]":
+    ...
+
+
+def run_in_background(
+    f: Union[
+        Callable[..., R],
+        Callable[..., Awaitable[R]],
+    ],
```
DMRobertson, line 776: +764 seems to suggest that `f` could return a `defer.Deferred[R]` too.

(Is Deferred[T] considered compatible with Awaitable[T]?)

synapse/logging/context.py
```diff
@@ -751,6 +787,9 @@ def run_in_background(f, *args, **kwargs) -> defer.Deferred:
     # At this point we should have a Deferred, if not then f was a synchronous
     # function, wrap it in a Deferred for consistency.
     if not isinstance(res, defer.Deferred):
+        # All `Awaitable`s in Synapse are either coroutines or `Deferred`s
+        assert not isinstance(res, Awaitable)
+
```
DMRobertson, line 812: Not sure about this, but also not sure if I can put my finger on why this gives me a bad feeling.

synapse/logging/context.py
squahtx: It is: https://github.com/twisted/twisted/blob/trunk/src/twisted/internet/defer.py#L347

The `T` in `Deferred[T]` is contravariant for some reason. If it weren't contravariant, we wouldn't need the `type: ignore`s for the overloads, for reasons I haven't figured out yet.

synapse/federation/federation_server.py
squahtx: Yes, the values of `pdu_results, _` should be the same before and after this PR.

synapse/logging/context.py
squahtx: I don't think mypy ever uses the base version of a method when there are `@overload`s. I'm hoping 3-tuples are enough for everybody

synapse/logging/context.py
squahtx: Oops, I'd entirely forgotten that there was a previous PR that tried to type hint these. I didn't mean to step on your toes!

Yes, the docstring contradicts the type hint. I wasn't sure what to do here. I can't find anywhere where we feed a coroutine to this function and I would remove the coroutine support, but it feels risky?

I could change it to accept an `Awaitable`, then I'd need another assert in the implementation, since we can't really handle non-coroutine, non-deferred awaitables.

synapse/util/async_helpers.py
squahtx: If anyone asks, I'm going to claim that the overloads were inspired by [Rust](https://doc.rust-lang.org/std/primitive.tuple.html#impl-Debug-4) rather than [C#](https://docs.microsoft.com/en-us/dotnet/api/system.valuetuple-4) or [TypeScript](https://github.com/microsoft/TypeScript/blob/4e39023396f03907f7c9fecb9ad72d5f42256eff/lib/lib.es2015.promise.d.ts#L41).

synapse/logging/context.py
squahtx: I considered accepting a union, but decided it wasn't worth the hassle of replacing `Awaitable`s throughout synapse.

I'm not sure how we'd handle other awaitables here. Create a dummy async function which `await`s them?

synapse/logging/context.py
DMRobertson: Oh! Well that makes life much easier!

> The `T` in `Deferred[T]` is contravariant for some reason.

Maybe because the `T` gets passed to a callback, and `Callable[T, ...]` is contravariant in `T`?

synapse/logging/context.py
DMRobertson: Not to worry---I had let that one rot a little and I don't think I had as robust an understanding of what's going on then as you do now!

Maybe we can remove `(or coroutine)` from the docstring, if we're convinced that this always is given a `Deferred`.

synapse/logging/context.py
DMRobertson: I guess I just wanted to confirm that this overload is getting used, and that mypy isn't just falling back to `Any` or something due to the type-ignore.

Can we do a reveal_type on something returned by this overload specifically, to be convinced that mypy is using this overload definition?

synapse/logging/context.py
DMRobertson: Ahh. I don't think I'd fully appreciated that `Awaitable` and `Coroutine` were different concepts.

Would it help to have a Synapse-only type alias for `Union[Coroutine-that-returns-T, Deferred[T]]`?

squahtx:
I noticed I was missing a few type hints, sorry @DMRobertson !

synapse/logging/context.py
```diff
@@ -399,17 +415,6 @@ def __exit__(self, type, value, traceback) -> None:
...
-    def copy_to(self, record) -> None:
-        """Copy logging fields from this context to a log record or
-        another LoggingContext
-        """
-
-        # we track the current request
-        record.request = self.request
-
-        # we also track the current scope:
-        record.scope = self.scope
```
squahtx, line 411: The `copy_to` methods don't seem to be used anymore

synapse/logging/context.py
squahtx: Whoops, my comment above does not make any sense. I thought this was about `gather_results`.

mypy does make use of the overload:
https://mypy-play.net/?mypy=latest&python=3.10&gist=1cbb7ab4a946e9b043218e7b1300694b

DMRobertson: Brilliant, thank you!

synapse/logging/context.py
```diff
@@ -711,16 +721,61 @@ def nested_logging_context(suffix: str) -> LoggingContext:
...
+    # "Overloaded function signatures 1 and 2 overlap with incompatible return types"
+    ...
+
+
+@overload
+def preserve_fn(f: Callable[..., R]) -> Callable[..., "defer.Deferred[R]"]:
+    ...
+
+
+def preserve_fn(
```
DMRobertson, line 741: As an aside, I don't really know what the name `preserve_fn` means here. Not one for this PR though.

squahtx: Neither do I really

synapse/logging/context.py
squahtx: That makes sense!

synapse/logging/context.py
squahtx: After some discussion in #synapse-dev, the best thing to do seems to be to leave the assert as is.

synapse/logging/context.py
squahtx: I've elected to remove the coroutine support, since mypy thinks it is unused, and I can't find any usage with coroutines manually.

DMRobertson (approved):
LGTM!

I had a quick look through matches for `make_deferred_yieldable\(` in `synapse` and it all seemed to be using deferreds.

squahtx:
> I had a quick look through matches for `make_deferred_yieldable\(` in `synapse` and it all seemed to be using deferreds.

Thank you for double checking!

synapse/logging/context.py
```diff
@@ -751,6 +805,10 @@ def run_in_background(f, *args, **kwargs) -> defer.Deferred:
     # At this point we should have a Deferred, if not then f was a synchronous
     # function, wrap it in a Deferred for consistency.
     if not isinstance(res, defer.Deferred):
+        # `res` is not a `Deferred` and not a `Coroutine`.
+        # There are no other types of `Awaitable`s we expect to encounter in Synapse.
+        assert not isinstance(res, Awaitable)
+
```
squahtx, line 811: This turns out to be false. We have `DoneAwaitable`s from `maybe_awaitable` and completed `Future`s from `make_awaitable`.

I've added support for these in #12505.

PR #11558, opened 2021-12-10
Stop populating `state_events.prev_state`

richvdh:
this field is never read, so we may as well stop populating it.


squahtx (approved):
lgtm, I can't spot any usage of the field

PR #11560, opened 2021-12-10
skip some dict munging in event persistence

richvdh:
This creates something I've been after for a while, which is a database helper method like `simple_insert_many`, but which takes raw row values, rather than {key=>value} dicts.

This saves us a bunch of dict munging and safety-checking, and makes it easier to use generators rather than creating intermediate lists and dicts.

### DIFF
--- a/synapse/storage/database.py
+++ b/synapse/storage/database.py
@@ -896,6 +896,9 @@ class DatabasePool:
     ) -> None:
         """Executes an INSERT query on the named table.
 
+        The input is given as a list of dicts, with one dict per row.
+        Generally simple_insert_many_values should be preferred for new code.
+
         Args:
             table: string giving the table name
             values: dict of new column names and values for them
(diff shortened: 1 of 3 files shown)


synapse/storage/database.py
```diff
@@ -933,23 +939,66 @@ def simple_insert_many_txn(
             if k != keys[0]:
                 raise RuntimeError("All items must have the same keys")
 
+        return DatabasePool.simple_insert_many_values_txn(txn, table, keys[0], vals)
+
+    async def simple_insert_many_values(
```
richvdh, line 944: incidentally, ideas for better names for this thing are welcome.

clokep (approved):
Seems reasonable. This has always bugged me a big. I don't have an idea for a better name, but would be nice if we could get rid of the old one at some point and steal that name. 😄

PR #11561, opened 2021-12-10
Allow tracking puppeted users for MAU

jaywink:
Currently when puppeting another user, the user doing the puppeting is tracked for client ip's and MAU (if configured).

When tracking MAU is important, it becomes necessary to be possible to also track the client IP's and MAU of puppeted users. As an example a client that manages user creation and creation of tokens via the Synapse admin API, passing those tokens for the client to use.

This PR adds optional configuration to enable tracking of puppeted users into monthly active users. The default behaviour stays the same.


changelog.d/11561.feature
```diff
@@ -0,0 +1 @@
+Add config flag `mau_track_puppeted_users` to allow tracking puppeted users in terms of monthly active users.
```
squahtx, line 1:
```suggestion
Add `mau_track_puppeted_users` config flag to include puppeted users in the monthly active users count.
```

docs/sample_config.yaml
```diff
@@ -439,6 +439,11 @@ manhole_settings:
 #  - medium: 'email'
 #    address: 'reserved_user@example.com'
 
+# If enabled, puppeted users can also be tracked. By default when
+# puppeting another user, the user who has created the access token
+# for puppeting is tracked. If this is enabled, both requests are tracked.
+#mau_track_puppeted_users: false
```
squahtx, line 445:
```suggestion
# If enabled, puppeted users will be counted in monthly active users.
# By default, when puppeting another user, only the puppeteer is counted.
# If this is enabled, both users are counted.
#mau_track_puppeted_users: false
```

synapse/api/auth.py
```diff
@@ -208,6 +209,18 @@ async def get_user_by_req(
...
+                    user_info.user_id != user_info.token_owner
+                    and self._mau_track_puppeted_users
+                ):
+                    await self.store.insert_client_ip(
+                        user_id=user_info.user_id,
+                        access_token=access_token,
+                        ip=ip_addr,
+                        user_agent=user_agent,
+                        device_id=device_id,
+                    )
```
squahtx, line 223: `insert_client_ip` also affects the list of client ips and devices, which aren't directly related to MAU. I think we should either call `populate_monthly_active_users` directly, or rename the config option to be clearer about the side effects.

edit: `populate_monthly_active_users` isn't available on workers unfortunately

squahtx (changes requested):
I think we should document the non-MAU side effects of `insert_client_ip` a bit better

plus some minor wording suggestions which might not be applicable any more

synapse/api/auth.py
jaywink: Yeah I was struggling a bit on how to submit this due to exactly this reason that feels a bit wrong. As a change it feels like the least invasive change to how things currently work and behave though, so I went with this suggestion.

There are other settings that affect MAU that are outside the MAU config settings, like `track_appservice_user_ips`. We could do the same here. Possibly `ServerConfig` item (on the top level) as `track_puppeted_user_ips` (logical because it also contains MAU config items). Would that sound better or do you have other suggestions where the config would logically live?

squahtx: I wasn't aware we had `track_appservice_user_ips`, thanks! I'd be in favour of us following the same pattern and going with `track_puppeted_user_ips`.

jaywink: Have adjusted! Moved the config flag to api config, which seems more relevant, given this is used in api auth. The server config feels a bit bloated already with different things and the appservice option was in the appservice config, so these couldn't be bundled.

Should years in copyright be updated btw when changing files?

synapse/config/api.py
```diff
@@ -59,6 +60,11 @@ def generate_config_section(cls, **kwargs) -> str:
            #
            #additional_event_types:
            #  - org.example.custom.event.type
+
+        # If enabled, puppeted user IP's can also be tracked. By default when
+        # puppeting another user, the user who has created the access token
+        # for puppeting is tracked. If this is enabled, both requests are tracked.
```
squahtx, line 66: `track_appservice_user_ips` has a note about mau tracking ("Implicitly enables MAU tracking for application service users."). Shall we add a similar note here?

synapse/api/auth.py
squahtx: > Should years in copyright be updated btw when changing files?

I think so, but I'm not going to block the PR over it!

synapse/config/api.py
jaywink: I thought I added that /o\ I was going to at least, I swear :sweat_smile: 

Added now!

jaywink:
I'm putting this PR as draft, noticed an issue testing with an actual host this patch is needed for, related to consent. Will update in a bit.

jaywink:
@squahtx Actually the errors we were seeing deployment ended up not being related to these changes directly. The issue was this particular customer was using a custom client and not expecting users to handle consent. Activating user IP tracking caused consent flow to fire (due to [replication user_ip handler](https://github.com/matrix-org/synapse/blob/bf72d10dbf506f5ea486d67094b6003947d38fb7/synapse/server_notices/server_notices_sender.py#L51) getting triggered by synchrotron worker) and that caused issues. Ends up server notices and consent was erronously configured for this client and disabling that makes this patch work as it should.

Of course this patch does mean if `track_puppeted_user_ips` is set to `true` and server notices/consent is enabled, those requests will start getting hit by the consent flow. But that feels like the correct thing to do anyway?

Not sure if something to look at in general, but it feels like request IP's are being logged for worker (or just synchrotron?) requests both in `auth/api` but also in the replication handler `on_USER_IP` handler on master (https://github.com/matrix-org/synapse/blob/af54167516c7211937efa5b800853f3088ef5178/synapse/replication/tcp/handler.py#L401). I assume that is intended.

Anyway, running this for the particular customer for a full day now makes me feel this should be good to go.

squahtx:
> request IP's are being logged for worker (or just synchrotron?) requests both in `auth/api` but also in the replication handler `on_USER_IP` handler on master

I would have thought that the IP logging in `auth/api` would bounce over to `on_USER_IP` on master. On workers, `insert_client_ip` is provided by `SlavedClientIpStore` which sends it over replication.

squahtx (approved):
LGTM!

docs/sample_config.yaml
```diff
@@ -1502,6 +1502,12 @@ room_prejoin_state:
    #additional_event_types:
    #  - org.example.custom.event.type
 
+# If enabled, puppeted user IP's can also be tracked. By default when
+# puppeting another user, the user who has created the access token
+# for puppeting is tracked. If this is enabled, both requests are tracked.
+# Implicitly enables MAU tracking for puppeted users.
+#track_puppeted_user_ips: false
```
richvdh, line 1509: * this is missing a `#` line
* I'm struggling to follow the description. What does the first sentence mean? What does "If this is enabled" mean (just uncommenting the example setting, ie setting `track_puppeted_user_ips: false`?)
* convention is that the example should be the *opposite* to the default, which doesn't seem to be the case here.

More details at https://matrix-org.github.io/synapse/develop/code_style.html#configuration-file-format

PR #11564, opened 2021-12-13
checks for generators in database functions

richvdh:
A couple of safety-checks to hopefully stop people doing what I just did, and create a storage function which only works the first time it is called (and not when it is re-run due to a database concurrency error or similar). It's not water-tight, of course, but it should catch the most obvious failure modes.

Ideally, I'd have this throw a ValueError or somesuch, but I'm a bit worried it'll break existing code that I haven't yet found, so for now at least, let's just log an error.

Also fixes up a particular case that I discovered during testing. In theory this is bug which could have caused us to send fewer presence updates than we should have done. In practice, it's such an edge-case I can't really imagine anyone will ever have hit it.

### DIFF
--- a/synapse/storage/database.py
+++ b/synapse/storage/database.py
@@ -548,14 +548,21 @@ class DatabasePool:
         # will fail if we have to repeat the transaction.
         # For now, we just log an error, and hope that it works on the first attempt.
         # TODO: raise an exception.
-        for arg in args:
+        for i, arg in enumerate(args):
             if inspect.isgenerator(arg):
-                logger.error("Programming error: generator passed to new_transaction")
+                logger.error(
+                    "Programming error: generator passed to new_transaction as "
+                    "argument %i to function %s",
+                    i,
+                    func,
+                )
         for name, val in kwargs.items():
             if inspect.isgenerator(val):
                 logger.error(
-                    "Programming error: generator passed to new_transaction as argument %s",
+                    "Programming error: generator passed to new_transaction as "
+                    "argument %s to function %s",
                     name,
+                    func,
                 )
         # also check variables referenced in func's closure
         if inspect.isfunction(func):
@@ -564,7 +571,8 @@ class DatabasePool:
                 for i, cell in enumerate(f.__closure__):
                     if inspect.isgenerator(cell.cell_contents):
                         logger.error(
-                            "Programming error: function %s references generator %s via its closure",
+                            "Programming error: function %s references generator %s "
+                            "via its closure",
                             f,
                             f.__code__.co_freevars[i],
                         )
--- a/synapse/storage/database.py
+++ b/synapse/storage/database.py
@@ -528,6 +528,12 @@ class DatabasePool:
         the function will correctly handle being aborted and retried half way
         through its execution.
 
+        Similarly, the arguments to `func` (`args`, `kwargs`) should not be generators,
+        since they could be evaluated multiple times (which would produce an empty
+        result on the second or subsequent evaluation). Likewise, the closure of `func`
+        must not reference any generators.  This method attempts to detect such usage
+        and will log an error.
+
         Args:
             conn
             desc
--- a/synapse/storage/database.py
+++ b/synapse/storage/database.py
@@ -552,14 +552,16 @@ class DatabasePool:
                     name,
                 )
         # also check variables referenced in func's closure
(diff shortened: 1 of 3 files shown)


synapse/storage/database.py
```diff
@@ -536,6 +538,31 @@ def new_transaction(
...
+                logger.error("Programming error: generator passed to new_transaction")
+        for name, val in kwargs.items():
+            if inspect.isgenerator(val):
+                logger.error(
+                    "Programming error: generator passed to new_transaction as argument %s",
+                    name,
+                )
+        # also check variables referenced in func's closure
+        if inspect.isfunction(func):
+            f = cast(types.FunctionType, func)
```
richvdh, line 556: mypy doesn't understand that `FunctionType` is a particular type of Callable, hence this business.

clokep:
It seems like there's a bunch of other spots where we accept `Iterable` in `synapse/storage/database.py` -- are those all wrong and should be updated to `Collection`?

synapse/storage/database.py
```diff
@@ -1226,9 +1253,9 @@ async def simple_upsert_many(
         self,
         table: str,
         key_names: Collection[str],
-        key_values: Collection[Iterable[Any]],
+        key_values: Collection[Collection[Any]],
         value_names: Collection[str],
-        value_values: Iterable[Iterable[Any]],
+        value_values: Collection[Collection[Any]],
```
clokep, line 1258: Does `simple_upsert_many_txn` need to match these?

synapse/storage/database.py
```diff
@@ -536,6 +538,31 @@ def new_transaction(
             **kwargs
         """
 
+        # Robustness check: ensure that none of the arguments are generators, since that
+        # will fail if we have to repeat the transaction.
+        # For now, we just log an error, and hope that it works on the first attempt.
```
clokep, line 543: Should we update the docstring of this method to specifically call out not using generators?

synapse/storage/database.py
```diff
@@ -536,6 +538,31 @@ def new_transaction(
...
+        # also check variables referenced in func's closure
+        if inspect.isfunction(func):
+            f = cast(types.FunctionType, func)
+            if f.__closure__:
+                for i, cell in enumerate(f.__closure__):
+                    if inspect.isgenerator(cell.cell_contents):
+                        logger.error(
+                            "Programming error: function %s references generator %s via its closure",
+                            f,
+                            f.__code__.co_freevars[i],
```
clokep, line 563: Should this be `co_cellvars` instead of `co_freevars`? https://docs.python.org/3/reference/datamodel.html#index-55 is a bit hard to follow...

synapse/storage/database.py
richvdh: I don't think so, because `simple_upsert_many_txn` only goes through the iterables once.

richvdh:
> are those all wrong and should be updated to Collection?

probably. But I didn't want this PR to turn into an epic of hunting them down and chasing the fixes up through the codebase. Sorry!

synapse/storage/database.py
richvdh: empirically not.

I've not been able to find a good reference for this, either. In short, suppose you have:

```python
def foo():
    x=1
    def bar():
         print(x)
```

then, for `foo`, `x` is a `cellvar`, but for `bar`, it is a `freevar`.

```
>>> def foo():
...     x=1
...     def bar():
...          print(x)
...     print(f"bar: {bar.__code__.co_cellvars}, {bar.__code__.co_freevars}")
... 
>>> print(f"foo: {foo.__code__.co_cellvars}, {foo.__code__.co_freevars}")
foo: ('x',), ()
>>> foo()
bar: (), ('x',)
```

clokep: Thanks for the example, that's enlightening! Re-reading the docs again it seems like this can only refer to variables from a closure, so I guess the ordering and length must be equal.

clokep (approved):
Looks good, can we file an issue to follow-up on the other methods to get them changed from `Iterable` -> `Collection`?

richvdh:
done with #11569.

richvdh:
wait, I've just realised this doesn't actually log enough to tell you where the problem is :facepalm:

clokep:
Oh, doesn't seem to log which function is called. That's not useful, sorry I missed that! 😢

PR #11565, opened 2021-12-13
Make `get_device` return None if the device doesn't exist rather than raising an exception.

reivilibre:


changelog.d/11565.misc
```diff
@@ -0,0 +1 @@
+Make `get_device` return None if the device doesn't exist rather than raising an exception.
```
squahtx, line 1:
```suggestion
Make `get_device` return `None` if the device doesn't exist rather than raising an exception.
```

synapse/storage/databases/main/devices.py
```diff
@@ -101,23 +101,24 @@ def count_devices_by_users_txn(txn, user_ids):
...
         hidden.
 
         Args:
             user_id: The ID of the user which owns the device
             device_id: The ID of the device to retrieve
         Returns:
-            A dict containing the device information
-        Raises:
-            StoreError: if the device is not found
+            A dict containing the device information, or None if the device does not exist.
```
squahtx, line 114:
```suggestion
            A dict containing the device information, or `None` if the device does not
            exist.
```

squahtx (approved):
LGTM!

reivilibre:
Thank you for the swift review!

PR #11566, opened 2021-12-13
Move HTML parsing to a separate file for URL previews.

clokep:
This is an internal refactor to:

* Split out the HTML parsing code from the resource code by moving it to a separate file. (Brings a ~1k LOC file to ~600 LOC.)
* Fix a circular import in the oEmbed code (which uses the HTML parsing code).
* Renames some of the HTML parsing methods to:
  * Make it clear which methods are "internal" to the module.
  * Clarify what the methods do.

### DIFF
--- a/changelog.d/11566.misc
+++ b/changelog.d/11566.misc
@@ -1,0 +1,1 @@
+Split the HTML parsing code from the URL preview resouce code.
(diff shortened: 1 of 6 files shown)


changelog.d/11566.misc
```diff
@@ -0,0 +1 @@
+Split the HTML parsing code from the URL preview resouce code.
```
squahtx, line 1:
```suggestion
Split the HTML parsing code from the URL preview resource code.
```

squahtx (approved):
Looks good!

clokep:
Thanks for the prompt review!

PR #11570, opened 2021-12-13
Remove redundant `COALESCE()`s around `COUNT()`s in database queries

squahtx:
`COUNT()` never returns `NULL`. A `COUNT(*)` over 0 rows is 0 and a
`COUNT(NULL)` is also 0.

The [Postgres documentation](https://www.postgresql.org/docs/current/functions-aggregate.html) also has some words to this effect:
> It should be noted that except for count, these functions return a null value when no rows are selected. In particular, sum of no rows returns null, not zero as one might expect, and array_agg returns null rather than an empty array when there are no input rows. The coalesce function can be used to substitute zero or an empty array for null when necessary.


clokep:
Is this also true of sqlite?

squahtx:
> Is this also true of sqlite?

The sqlite docs aren't as clear: https://www.sqlite.org/lang_aggfunc.html

In testing, sqlite's count has the same behavior as postgres:
```sql
$ sqlite
SQLite version 2.8.17
Enter ".help" for instructions
sqlite> CREATE TABLE test(a INT);
sqlite> SELECT COUNT(*) FROM test;
0

sqlite> INSERT INTO test VALUES (NULL);
sqlite> SELECT COUNT(*) FROM test;
1 -- COUNT(*) counts rows
sqlite> SELECT COUNT(a) FROM test;
0
```

DMRobertson (approved):

PR #11571, opened 2021-12-13
Add missing type hints to synapse.http.

clokep:
This starts adding some missing type hints to `synapse.http`. There's still quite a bit to do but wanted to continue making baby steps!

I wasn't sure if I should add these to `mypy.ini` or not, but listing individual files seems painful.

### DIFF
--- a/synapse/http/server.py
+++ b/synapse/http/server.py
@@ -276,7 +276,7 @@ class _AsyncResource(resource.Resource, metaclass=abc.ABCMeta):
             f = failure.Failure()
             self._send_error_response(f, request)
 
-    async def _async_render(self, request: SynapseRequest) -> Optional[Tuple[int, Any]]:  # type: ignore[return]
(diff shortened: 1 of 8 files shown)


synapse/http/server.py
```diff
@@ -271,7 +276,7 @@ async def _async_render_wrapper(self, request: SynapseRequest):
             f = failure.Failure()
             self._send_error_response(f, request)
 
-    async def _async_render(self, request: Request):
+    async def _async_render(self, request: SynapseRequest) -> Optional[Tuple[int, Any]]:  # type: ignore[return]
```
squahtx, line 279: What complaint is being suppressed here?

clokep: Seems like it isn't needed. Not sure why I added it!

squahtx (approved):
lgtm!

DMRobertson:
I think a lot of this is duplicated within Sydent and Sygnal---would be keen to see stuff migrated to matrix-python-common if you see a good opportunity to do so!

PR #11572, opened 2021-12-13
Fix nginx config

koyuawsmbrtn:
### Pull Request Checklist

`Signed-off-by: koyu <me@koyu.space>`

### DIFF
--- a/docs/reverse_proxy.md
+++ b/docs/reverse_proxy.md
@@ -63,7 +63,7 @@ server {
 
     server_name matrix.example.com;
 
-    location ~* ^(\/_matrix|\/_synapse\/client) {
+    location ~* ^(\/_matrix|\/_synapse|\/client) {
         # note: do not add a path (even a single /) after the port in `proxy_pass`,
         # otherwise nginx will canonicalise the URI and cause signature verification
         # errors.


dklimpel:
IMO this line is correct. The used URLs are `/_matrix` and `/_synapse/client`, e.g. used in OIDC (https://matrix-org.github.io/synapse/latest/openid.html#sample-configs)

What is wrong with this line?

koyuawsmbrtn:
I used that line on my Raspberry Pi and it didn't work, `_synapse` just 404ed

DMRobertson:
Agreed with @dklimpel, this doesn't look right. Can you find the log line for the 404 so we can see exactly which URL was not found?

koyuawsmbrtn:
This doesn't 404 anymore since I changed it, but this is the URL Element was trying to reach: https://im.koyu.space/_synapse/admin/v1/users/%40koyu%3Aim.koyu.space/admin

DMRobertson:
> This doesn't 404 anymore since I changed it, but this is the URL Element was trying to reach: https://im.koyu.space/_synapse/admin/v1/users/%40koyu%3Aim.koyu.space/admin

Ahh, I see---the admin API. The very [bottom of that documentation page](https://github.com/matrix-org/synapse/pull/11572) mentions this:

> Endpoints for administering your Synapse instance are placed under /_synapse/admin. These require authentication through an access token of an admin user. However as access to these endpoints grants the caller a lot of power, we do not recommend exposing them to the public internet without good reason.

For that reason, we don't want to encourage people to expose the admin API by default, and so I don't think we'd merge this change. We recommend that you only accept admin API requests from a machine whose access is strictly controlled, rather than the entire public internet.

Note: `\/_matrix|\/_synapse|\/client` will route requests to `$PUBLIC_BASEURL/client` to synapse---but Synapse has no such endpoint. If you still want to make this config change, I'd suggest either `\/_matrix|\/_synapse` or `\/_matrix|\/_synapse/client|\/_synapse/admin`.

richvdh:
> this is the URL Element was trying to reach: https://im.koyu.space/_synapse/admin/v1/users/%40koyu%3Aim.koyu.space/admin

This doesn't sound right to me. Element should not be trying to use the admin api - if it is, that is a bug in element.

richvdh:
@koyuawsmbrtn as a general point for any future contributions - it's very helpful if your PR can explain up front what it is you are trying to achieve, to save people having to ask.  Thank you for getting involved though!

koyuawsmbrtn:
> > this is the URL Element was trying to reach: [im.koyu.space/_synapse/admin/v1/users/%40koyu%3Aim.koyu.space/admin](https://im.koyu.space/_synapse/admin/v1/users/%40koyu%3Aim.koyu.space/admin)
> 
> This doesn't sound right to me. Element should not be trying to use the admin api - if it is, that is a bug in element.

Even if I am admin on the server?

callahad:
> Even if I am admin on the server?

Yep - the Admin API is only designed for use by external / standalone tools since it's a Synapse specific thing. Element itself _shouldn't_ ever try to access Synapse Admin APIs; it should be able to do everything it needs to through standardized Matrix APIs.

dklimpel:
The URLs `/_synapse/admin/` are only for administration. A normal users and element client do not need access to this API.

callahad:
It looks like Element may indeed be trying to access those APIs in [a few places](https://github.com/matrix-org/matrix-js-sdk/blob/develop/src/client.ts#L5977-L5992). We'll have to file a bug about that; thank you for bringing it to our attention!

Issue #11399, opened 2021-11-19
Include whether the requesting users has responded to an MSC3440 thread


clokep:
Per an [update to MSC3440](https://github.com/matrix-org/matrix-doc/pull/3440/commits/2b76a6eed919e2bb9f3a6836bcaee9a5d7510f3a). (Note that we're going to be consistent and not use a camel-case version as suggested there.)

clokep:
This seems like it might not be super feasible since we don't have the requesting user when serializing events. We could probably pass through the user to those places, but I'm not sure if that is desirable (in terms of caching and such). What do we think of this?

Relevant code is near:

* https://github.com/matrix-org/synapse/blob/ba00e20234eadae66f105f8bda64e39beed9a92d/synapse/events/utils.py#L467-L480, which calls
* https://github.com/matrix-org/synapse/blob/ba00e20234eadae66f105f8bda64e39beed9a92d/synapse/storage/databases/main/relations.py#L273-L327

clokep:
This is not necessary, the filtering adding to MSC3440 is meant to be used for this purpose.

clokep:
We might need to do this for thread previews in the timeline.

PR #11574, opened 2021-12-14
Convert EventStreamResult to attrs.

clokep:
I was attempting to convince myself that `serialize_events` never took a `JsonDict` and only ever received `EventBase` instances...well turns out it does take them via the event streams.

I fixed up some of the types around this to use `attrs` instead of a `namedtuple` to avoid `Any` being passed around.

Vaguely related to #11399 since that spurred the original work.


squahtx (approved):
lgtm

PR #11576, opened 2021-12-14
Remove the 'password_hash' from the Users Admin API endpoint response dictionary

anoadragon453:
A `password_hash` field is returned by both `GET` and `PUT` methods of `/_synapse/admin/v2/users/<user_id>`. There is little use for this to be returned, and returning it is an unnecessary security risk.

Note that I don't believe this field was included on purpose. It is simple a consequence of using `get_user_by_id` directly for this Admin API endpoint (which we probably shouldn't do, as it means that changing an internal storage method changes an external API shape!).

~~As this is a backwards-incompatible change, the endpoint version has been bumped.~~ This ended up being considered a bug, slight security risk, and not intentionally included in the first place. Additionally, we're not aware of any usecases for it. Thus we have simply dropped it from the existing v2 API.

Reviewable commit-by-commit.

### DIFF
--- a/synapse/handlers/admin.py
+++ b/synapse/handlers/admin.py
@@ -55,21 +55,43 @@ class AdminHandler:
 
     async def get_user(self, user: UserID) -> Optional[JsonDict]:
         """Function to get user details"""
-        ret = await self.store.get_user_by_id(user.to_string())
-        if ret:
-            profile = await self.store.get_profileinfo(user.localpart)
-            threepids = await self.store.user_get_threepids(user.to_string())
-            external_ids = [
-                ({"auth_provider": auth_provider, "external_id": external_id})
-                for auth_provider, external_id in await self.store.get_external_ids_by_user(
-                    user.to_string()
-                )
-            ]
-            ret["displayname"] = profile.display_name
-            ret["avatar_url"] = profile.avatar_url
-            ret["threepids"] = threepids
-            ret["external_ids"] = external_ids
-        return ret
+        user_info_dict = await self.store.get_user_by_id(user.to_string())
+        if user_info_dict is None:
+            return None
+
+        # Restrict returned information to a known set of fields. This prevents additional
+        # fields added to get_user_by_id from modifying Synapse's external API surface.
+        user_info_to_return = [
+            "admin",
+            "deactivated",
+            "shadow_banned",
+            "creation_ts",
+            "appservice_id",
+            "consent_server_notice_sent",
+            "consent_version",
+            "user_type",
+        ]
+
+        # Restrict returned keys to a known set.
+        user_info_dict = {
+            key: value for key, value in user_info_dict if key in user_info_to_return
+        }
+
+        # Add additional user metadata
+        profile = await self.store.get_profileinfo(user.localpart)
+        threepids = await self.store.user_get_threepids(user.to_string())
+        external_ids = [
+            ({"auth_provider": auth_provider, "external_id": external_id})
+            for auth_provider, external_id in await self.store.get_external_ids_by_user(
+                user.to_string()
+            )
+        ]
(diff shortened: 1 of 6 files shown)


richvdh:
I'd be tempted to just change the `v2` endpoint. Anything relying on `password_hash` existing sounds pretty broken to me.

@dklimpel any thoughts here?

dklimpel:
I have no idea why an administrator should need the `password_hash` of a user. IMO it would be sufficient to remove the `password_hash` from v2. This would be a security benefit now and not later. If security is the reason for removing than should  backwards-incompatible not be a blocker. A patch which does not solve the problem is not a patch.

A useful information for an administrator maybe is to know if a password is set, especialy if Synapse is used with an IdP for OIDC.

anoadragon453:
@richvdh I'm happy to just change v2 as well, given that I don't know of anyone using it in the wild. Do you think we need a deprecation/warning for this at all?

It's less a pressing security matter, more that it's just information that is slightly sensitive and does not need to be exposed. Thus, it should not be.

richvdh:
> @richvdh I'm happy to just change v2 as well, given that I don't know of anyone using it in the wild. Do you think we need a deprecation/warning for this at all?

personally I wouldn't bother with a deprecation period.

synapse/handlers/admin.py
```diff
@@ -55,21 +55,47 @@ async def get_whois(self, user: UserID) -> JsonDict:
...
+        profile = await self.store.get_profileinfo(user.localpart)
+        threepids = await self.store.user_get_threepids(user.to_string())
+        external_ids = [
+            ({"auth_provider": auth_provider, "external_id": external_id})
+            for auth_provider, external_id in await self.store.get_external_ids_by_user(
+                user.to_string()
+            )
+        ]
+        user_info_dict["displayname"] = profile.display_name
+        user_info_dict["avatar_url"] = profile.avatar_url
```
DMRobertson, line 94: I know that these two can be `None` if not set. Should we point out they might be `null` in the response? Maybe overkill.

synapse/rest/admin/users.py
```diff
@@ -399,10 +398,10 @@ async def on_PUT(
                     target_user, requester, body["avatar_url"], True
                 )
 
-            user = await self.admin_handler.get_user(target_user)
-            assert user is not None
+            user_info_dict = await self.admin_handler.get_user(target_user)
+            assert user_info_dict is not None
 
-            return 201, user
+            return 201, user_info_dict
```
DMRobertson, line 404: HTTPStatus.CREATED while we're here? (201 wasn't one I'd memorised)

synapse/handlers/admin.py
anoadragon453: Oh, yes that would be good to mention. In fact, it may be nice to mention the return type of each field explicitly in a table in the documentation.

I'll add some (`// can be null if not set`) comments to this PR, and then make a separate PR for describing each field (much as we have under the [PUT bit of the Users API](https://matrix-org.github.io/synapse/latest/admin_api/user_admin_api.html#create-or-modify-account)).

synapse/handlers/admin.py
```diff
@@ -55,21 +55,47 @@ async def get_whois(self, user: UserID) -> JsonDict:
...
-            ret["threepids"] = threepids
-            ret["external_ids"] = external_ids
-        return ret
+        user_info_dict = await self.store.get_user_by_id(user.to_string())
+        if user_info_dict is None:
+            return None
+
+        # Restrict returned information to a known set of fields. This prevents additional
+        # fields added to get_user_by_id from modifying Synapse's external API surface.
+        user_info_to_return = [
```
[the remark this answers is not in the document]
DMRobertson: Good spot!

PR #11578, opened 2021-12-14
update m.direct list for InviteAutoAccepter module

chagai95:
### Pull Request Checklist

  Signed-off-by: Chagai Friedlander me@chagai.website

  Sorry for not running the linters, will try tomorrow.

### DIFF
--- a/synapse/handlers/room_member.py
+++ b/synapse/handlers/room_member.py
@@ -505,37 +505,47 @@ class RoomMemberHandler(metaclass=abc.ABCMeta):
 
     async def update_m_direct(
         self,
-        target: UserID,
+        sender: str,
+        target: str,
         room_id: str,
     ) -> None:
         """Update a user's m_direct list.
 
         Params:
+            sender: The user who sent the invite, used as the key to update the list.
             target: The user whose user's m_direct list is being updated.
             room_id: The room ID which is being added to the list.
         """
+
         logger.debug(
-            "InviteAutoAccepter: update_m_direct is triggered %s", room_id
+            "InviteAutoAccepter: update_m_direct is triggered with room id: %s, sender user id: %s, target user id: %s",
+            room_id,
+            sender,
+            target,
         )
 
         # Retrieve user account data
-        user_account_data, _ = await self.store.get_account_data_for_user(target.to_string())
+        user_account_data, _ = await self.store.get_account_data_for_user(target)
 
         # Retrieve m direct list
         direct_rooms = user_account_data.get(AccountDataTypes.DIRECT, {})
 
         # Check which key this room is under
         if isinstance(direct_rooms, dict):
-            for key, room_id_list in direct_rooms.items():
-                if room_id not in room_id_list:
+            m_list_changed = False
+            if sender in direct_rooms:
+                if room_id not in direct_rooms[sender]:
                     # Add new room_id to this key
-                    direct_rooms[key].append(room_id)
-
-                    # Save back to user's m.direct account data
-                    await self.account_data_handler.add_account_data_for_user(
-                        target.to_string(), AccountDataTypes.DIRECT, direct_rooms
-                    )
-                    break
+                    direct_rooms[sender].append(room_id)
+                    m_list_changed = True
+            else:
+                direct_rooms[sender] = [room_id]
+                m_list_changed = True
+            if m_list_changed:
+                # Save back to user's m.direct account data
+                await self.account_data_handler.add_account_data_for_user(
+                    target, AccountDataTypes.DIRECT, direct_rooms
+                )
 
     async def update_membership_locked(
         self,
--- a/synapse/handlers/room_member.py
+++ b/synapse/handlers/room_member.py
@@ -503,6 +503,40 @@ class RoomMemberHandler(metaclass=abc.ABCMeta):
 
         return result
 
+    async def update_m_direct(
+        self,
+        target: UserID,
+        room_id: str,
+    ) -> None:
+        """Update a user's m_direct list.
+
+        Params:
+            target: The user whose user's m_direct list is being updated.
+            room_id: The room ID which is being added to the list.
+        """
+        logger.debug(
+            "InviteAutoAccepter: update_m_direct is triggered %s", room_id
+        )
+
+        # Retrieve user account data
+        user_account_data, _ = await self.store.get_account_data_for_user(target.to_string())
+
+        # Retrieve m direct list
+        direct_rooms = user_account_data.get(AccountDataTypes.DIRECT, {})
+
+        # Check which key this room is under
+        if isinstance(direct_rooms, dict):
+            for key, room_id_list in direct_rooms.items():
+                if room_id not in room_id_list:
+                    # Add new room_id to this key
+                    direct_rooms[key].append(room_id)
+
+                    # Save back to user's m.direct account data
+                    await self.account_data_handler.add_account_data_for_user(
+                        target.to_string(), AccountDataTypes.DIRECT, direct_rooms
+                    )
+                    break
+
     async def update_membership_locked(
         self,
         requester: Requester,
--- a/synapse/module_api/__init__.py
+++ b/synapse/module_api/__init__.py
@@ -880,16 +880,6 @@ class ModuleApi:
             content=content,
         )
 
-        # Update the m direct list
-        is_direct = content.get("is_direct", None)
-        logger.debug("InviteAutoAccepter: is_direct is %s", is_direct)
-        if is_direct:
-            await self._hs.get_room_member_handler().update_m_direct(
-                sender=sender,
-                target=target,
-                room_id=room_id,
-            )
-
         # Try to retrieve the resulting event.
         event = await self._hs.get_datastore().get_event(event_id)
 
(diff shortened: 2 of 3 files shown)


synapse/handlers/room_member.py
```diff
@@ -503,6 +503,50 @@ async def update_membership(
 
         return result
 
+    async def update_m_direct(
```
chagai95, line 506: I'm not sure about async for the function I added, does it need this?

synapse/module_api/__init__.py
```diff
@@ -880,6 +880,16 @@ async def update_room_membership(
             content=content,
         )
 
+        # Update the m direct list
+        is_direct = content.get("is_direct", None)
+        logger.debug("InviteAutoAccepter: is_direct is %s", is_direct)
+        if is_direct:
+            await self._hs.get_room_member_handler().update_m_direct(
```
chagai95, line 887: I'm not sure about await for the function I added, does it need to wait?

synapse/module_api/__init__.py
```diff
@@ -880,6 +880,16 @@ async def update_room_membership(
             content=content,
         )
 
+        # Update the m direct list
+        is_direct = content.get("is_direct", None)
+        logger.debug("InviteAutoAccepter: is_direct is %s", is_direct)
+        if is_direct:
```
chagai95, line 886: I notices a log where the sender and target are the same, is this somehow because there is also an event for the user who created the room, should I exclude such cases?
  `if target not is sender and is_direct`

synapse/module_api/__init__.py
```diff
@@ -880,6 +880,16 @@ async def update_room_membership(
             content=content,
         )
 
+        # Update the m direct list
+        is_direct = content.get("is_direct", None)
```
reivilibre, line 884: Is `is_direct` being in the content of an event something you've seen elsewhere? Looking at join events for my DMs I can't see an example of this.
Oh however I do see it in the *invite* event's content.
The usual way this seems to work (according to the spec) is that the client adds it to `m.direct` after accepting the invite (it could prompt the user first or ignore it, but Element doesn't do either of those).

After asking around, I think it would be better to expose a separate function for setting whether the room is in `m.direct` for a given user, and have the auto-accept module call into that (since it knows if an invite has `is_direct` set).

reivilibre (changes requested):
Thanks for contributing! I hope the comments above are helpful?

synapse/module_api/__init__.py
chagai95: Ah yes sorry I probably forgot to add the content part being passed to this pull request.

chagai95: I'm not aware of the per room one and what it's used for (or could be used for) so I'm guessing the global one is what we care about.

chagai95: Can I ask why it's better to separate this into two separate functions? 🙈

synapse/module_api/__init__.py
chagai95: Ah yes sorry I probably forgot to add the content part being passed to this pull request. so I guess I do need that check.

synapse/module_api/__init__.py
reivilibre: I think there are a few reasons why a separate function arguably makes sense.

Clients have to make a separate request for setting a room as direct or not, so it makes some sense to reflect that design in the module API. Exposing account data is a bit more general than just the `m.direct` setting and may well have other uses in the future (for example, tags other than `m.direct`, lists of ignored users, and so on).
The two operations are also somewhat distinct; joining a room modifies room state, but setting it as direct modifies the *user's* account state. The current design doesn't leave a way for the status to be changed after the room is joined (and it also has no meaning if the target membership is anything other than 'join').

benparsons:
@reivilibre thanks for picking this up - can you tell me what the next steps are? Is there an option to merge and iterate on the requirement?

reivilibre:
> @reivilibre thanks for picking this up - can you tell me what the next steps are? Is there an option to merge and iterate on the requirement?

I don't think the code will work as intended as it currently stands, since it assumes that `join` events will contain `is_direct` (this is not the case — `invite` events contain that).
It's also difficult to merge and change later since this becomes public API the moment it goes into a release, which then causes backward compatibility problems if we try to change it.

The next steps seem to be:

* take most of the actual logic from this PR to a new PR in the auto accept module
* write a PR against Synapse that exposes two methods to get and update globally-scoped account data for a given user. It's likely fine to just pass through something similar to the existing functions on the data store; the main thing that's needed is that modules can access it and that it's documented as public API rather than having modules reach into the depths of Synapse's code (which we can and most likely will break when shuffling code around).

I'll be away until after Christmas but if there are any questions I'll be happy to address them when I'm here.

chagai95:
It actually works exactly as intended but I completely understand your concern, from how I understood the spec it could also be implemented this way but it's not how it should be done.

I did add the content to the membership=join member event and used (abused) that to check if the m.direct list should be updated or not, isn't that more like a feature than a bug? 😉

reivilibre:
> I did add the content to the membership=join member event and used (abused) that to check if the m.direct list should be updated or not, isn't that more like a feature than a bug? wink

My reading of the spec is that `is_direct` only has a meaning for `invite`s — as such, putting it in a join event for this one use case is a bit of a hack (and makes things very confusing, especially because it doesn't behave in the same way as a client sending that event would).

The existing client workflow is to send a plain join, then read-modify-write the account data. It seems only reasonable to expect a module to do the same.

*If* you were going to place a flag in the event content, then realistically it should be a vendor-specific-prefixed name like `org.matrix.synapse.is_direct` since `is_direct` has no current meaning in the spec in that context (but might do in the future!). However I'm mostly saying this to illustrate the issue I have with it: it doesn't feel right at all.

> It seems to me like adding a new API call from synapse for account data might be a little bit of an overkill for our use case (but nevertheless pretty cool to have as a feature in general)

Being a bit extreme, you could say this to some degree about many API calls that Synapse exposes, however I think it's sensible to prefer module logic to exist in modules and for the interfaces to be generic.
Nowhere (except for room upgrades :/) in Synapse currently cares about what rooms are `m.direct`.
On the other hand, passing through the existing functions for changing account data seems fairly easy to do and gives the same functionality as the client endpoints do.

N.B. I wouldn't object to leaving out the room-scoped account data functions for another time; no need to go and add everything right away — it's fine to keep it simple.

> I was also under the impression that's not something that should be easy for system admins to tamper with because it's a personal user setting?

This is not a system admin, this is a module. You could make the same point about sending messages or indeed automatically accepting invites on behalf of a user: that's a personal user choice. So I think if it's fine to accept an invite on behalf of a user, then it's fine to also amend the account data to set the direct flag on it.

> What I'm trying to say is that the correct solution for this it seems to be is to implement canonical DMs and that'll probably break everything anyway?

Maybe, but I don't think Canonical DMs are part of the spec and *may never be* (they've been floating around for years I think..). I think we'll need to cross that bridge if/when we come to it.
I'm not up to date on Canonical DMs but it seems to me that the client API would break in the same way.


Hopefully that addresses your concerns?

chagai95:
Yes, very detailed explanation, thanks for that! will hopefully get around soon to adding another function for that, otherwise if anyone else wants to take over and do what needs to be done, that'll be great.

chagai95:
I exposed another method to update the m direct list and created a PR on the auto_accept_invite module as well using that exposed function. I also realized I understood yet one more  thing wrong (sender/target) and did that hopefully the correct way this time.

synapse/module_api/__init__.py
```diff
@@ -889,6 +889,40 @@ async def update_room_membership(
 
         return event
 
+    async def update_m_direct(
```
reivilibre, line 892: Sorry, I believe I have been misunderstood here.

What I was trying to say is that `update_m_direct` should live in the `synapse-auto-accept-invite` module.

Instead, I would suggest adding `add_account_data_for_user` (which is just a thin wrapper around `add_account_data_for_user`) and `get_account_data_for_user` (a thin wrapper again) to the Module API.

reivilibre (changes requested):

synapse/module_api/__init__.py
chagai95: Yep but as a first step I added just updating the list. I guess the logic for that is also better in the module, could we keep that also for a next step?

babolivier: The Module API is supposed to be a stable API. So I don't think we should add things to it if we're already planning to get rid of them soon, as it would mean we would need to maintain an unnecessary layer of backwards compatibility when we do the switch. So since we already have an idea of what the long-term solution should look like (and it's not a complex one) I think we should just do it.
As a nitpick, I think `add_account_data_for_user` should instead be named `set_account_data_for_user` - it makes it clearer that the method can be used for both creating new account data and updating existing one (sadly this makes it a bit inconsistent with the method from the account data handler, but I think it's best to make things clearer to module developers).

reivilibre: > As a nitpick, I think add_account_data_for_user should instead be named set_account_data_for_user

No objections here with that; it does reflect that it will overwrite account data so sounds right.

synapse/handlers/room_member.py
```diff
@@ -503,6 +503,50 @@ async def update_membership(
...
+        self,
+        sender: str,
+        target: str,
+        room_id: str,
+    ) -> None:
+        """Update a user's m_direct list.
```
anoadragon453, line 512: Note that this PR should ensure that there is a way of allowing the Module API to ignore updates from itself via the proposed `on_account_data_change` callback (https://github.com/matrix-org/synapse/pull/12327). Perhaps by adding an argument to that method to inform modules that it was a module vs. a user that updated the account data.

dklimpel:
IMO this is replaced by #12391

reivilibre:
Superseded by #12391 indeed — thanks though!

PR #8059, opened 2020-08-11
Re-implement unread counts (again)

babolivier:
The third and (hopefully) final installment of our summer series: how to implement unread counts in Synapse. As a season finale, this one contains a flashback on previous episodes, bringing back some of the storage code from https://github.com/matrix-org/synapse/pull/7673 and some of the sync logic from https://github.com/matrix-org/synapse/pull/7736.

The MSC I'm implementing here is still MSC2654, so as far as testing is concerned I just copied the tests from the latter PR.

This PR should be reviewable commit by commit.

Fixes #8007

PR #11580, opened 2021-12-14
Require Collections for simple_* methods.

### LINKED ISSUE #11569: require `Collection`s rather than `Iterable`s in storage functions
There are various methods in `DatabasePool` (such as `simple_insert_many_values`) which claim to take arbitrary `Iterable`s as arguments when in fact they cannot be generators (since they may be evaluated more than once: see https://github.com/matrix-org/synapse/pull/11564).

We should change their definitions to `Collection`.

clokep:
This fixes #11569 and continues the work from #11564 to use `Collection` instead of `Iterable` for the inputs to the various `simple_*` methods in the database code. I'll leave some notes inline.


synapse/storage/database.py
```diff
@@ -1680,8 +1680,8 @@ async def simple_select_many_batch(
         self,
         table: str,
         column: str,
-        iterable: Iterable[Any],
-        retcols: Iterable[str],
+        iterable: Collection[Any],
```
clokep, line 1683: `iterable` should be a `Collection` since we check if it is truth-y, generators are always truth-y so by requiring a `Collection` the short circuit works better.

synapse/storage/databases/main/pusher.py
```diff
@@ -196,27 +196,6 @@ async def get_if_user_has_pusher(self, user_id: str):
         # This only exists for the cachedList decorator
         raise NotImplementedError()
 
-    @cachedList(
-        cached_method_name="get_if_user_has_pusher",
-        list_name="user_ids",
-        num_args=1,
-    )
-    async def get_if_users_have_pushers(
```
clokep, line 204: This was unused since #8059 as far as I can tell.

synapse/visibility.py
```diff
@@ -96,7 +96,7 @@ async def filter_events_for_client(
         if isinstance(ignored_users_dict, dict):
             ignore_list = frozenset(ignored_users_dict.keys())
 
-    erased_senders = await storage.main.are_users_erased(e.sender for e in events)
+    erased_senders = await storage.main.are_users_erased([e.sender for e in events])
```
clokep, line 99: This was potential for a real bug.

synapse/storage/database.py
richvdh: this makes the comment at line 1710 a bit misleading.

Maybe we should just get rid of the short-circuit (it's not like the rest of the method is exactly heavy-weight for an empty iterable).

(Also, we could use `batch_iter` instead of rolling our own, less-efficient, version; `batch_iter` *does* support iterables.)

synapse/visibility.py
richvdh: I don't think so? `are_users_erased` uses `simple_select_many_batch`, which does work correctly with generators, modulo not short-circuiting for an empty generator.

richvdh:
(thank you for picking this up, @clokep !)

synapse/visibility.py
clokep: Fair, it came out requiring that to be a collection.

synapse/storage/database.py
clokep: I've made this change to use `batch_iter`.

synapse/storage/databases/main/user_erasure_store.py
```diff
@@ -40,7 +40,7 @@ async def is_user_erased(self, user_id: str) -> bool:
         return bool(result)
 
     @cachedList(cached_method_name="is_user_erased", list_name="user_ids")
-    async def are_users_erased(self, user_ids: Iterable[str]) -> Dict[str, bool]:
+    async def are_users_erased(self, user_ids: Collection[str]) -> Dict[str, bool]:
```
richvdh, line 43: this can be an Iterable now

synapse/storage/databases/state/store.py
```diff
@@ -612,7 +612,7 @@ def _purge_unreferenced_state_groups(
         )
 
     async def get_previous_state_groups(
-        self, state_groups: Iterable[int]
+        self, state_groups: Collection[int]
```
richvdh, line 615: so can this

richvdh (approved):
lgtm otherwise

synapse/storage/databases/main/user_erasure_store.py
clokep: Arg... sorry I should have reverted that commit. 😢

synapse/storage/databases/main/pusher.py
clokep: The signature of this is now "fine" that it could stay and mypy won't complain, but if it is unused...I think we should remove it?

PR #11583, opened 2021-12-15
Disable aggregation bundling on `/sync` responses

richvdh:
A partial revert of #11478. This turns out to have had a significant CPU impact on initial-sync handling. For now, let's disable it, until we find a more efficient way of achieving this.

(PR targeting `release-v1.49` while we decide if we want to make a 1.49.1 for this)

### DIFF
--- a/changelog.d/11583.bugfix
+++ b/changelog.d/11583.bugfix
@@ -1,0 +1,1 @@
+Fix a performance regression in `/sync` handling, introduced in 1.49.0.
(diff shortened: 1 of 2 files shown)


clokep:
The changes look fine, but are sad. 😢 Tests are failing since we test that `/sync` returns the relation in `tests/rest/client/test_relations.py`.

richvdh:
This doesn't actually seem to have made much difference to the CPU usage figures, so ... I don't know what's going on.

clokep (approved):
I'm going to approve this since it does what's on the tin. I'm going to leave it up to @richvdh to merge this if it seems necessary.

richvdh:
This was crucial to mitigate #11611 on `matrix.org`.

PR #11587, opened 2021-12-15
Cache empty responses from `/user/devices`

### LINKED ISSUE #11586: We don't cache the fact that a user has no devices
Spotted by @squahtx.

Some users, e.g. appservice users, won't have any devices whatsoever. If we query their homeserver to retrieve that user's devices, it should respond with an empty list of `devices`. In the processing of that response, by the time we reach this line we have a `devices` list from a successful response whose size is at most 999. 

We transform `devices` into a dictionary `{d["device_id"]: d for d in devices}`. Assuming there are no repeated device_ids, this is a reversible transformation.

https://github.com/matrix-org/synapse/blob/e5cdb9e2339e321e8a77a898d362d7fbc476303b/synapse/handlers/device.py#L951-L958

Here, `cached_devices` is:
 
https://github.com/matrix-org/synapse/blob/5305a5e88144828419249fd9e4c5198d92276a44/synapse/storage/databases/main/devices.py#L555-L565

Suppose we have never cached devices for this user, and that the `devices` list is empty. Then `cached_devices` will be an empty dictionary and so too will the transformed version of devices. Thus we'll set `ignore_devices = True` and not cache the fact that this user has no devices.

Then, whenever someone wants this user's device again, we'll repeat the federation request again.

DMRobertson:
Fixes #11586. Or it will fix, anyway.

### DIFF
--- a/synapse/handlers/device.py
+++ b/synapse/handlers/device.py
@@ -948,14 +948,14 @@ class DeviceListUpdater:
             devices = []
             ignore_devices = True
         else:
-            stream_id = await self.store.get_device_list_last_stream_id_for_remote(
+            prev_stream_id = await self.store.get_device_list_last_stream_id_for_remote(
                 user_id
             )
             cached_devices = await self.store.get_cached_devices_for_user(user_id)
 
             # To ensure that a user with no devices is cached, we skip the resync only
             # if we have a stream_id from previously writing a cache entry.
-            if stream_id is not None and cached_devices == {
+            if prev_stream_id is not None and cached_devices == {
                 d["device_id"]: d for d in devices
             }:
                 logging.info(
--- a/synapse/handlers/device.py
+++ b/synapse/handlers/device.py
@@ -948,7 +948,9 @@ class DeviceListUpdater:
             devices = []
             ignore_devices = True
         else:
-            stream_id = await self.store.get_device_list_last_stream_id_for_remote(user_id)
+            stream_id = await self.store.get_device_list_last_stream_id_for_remote(
+                user_id
+            )
             cached_devices = await self.store.get_cached_devices_for_user(user_id)
 
             # To ensure that a user with no devices is cached, we skip the resync only
--- a/synapse/handlers/device.py
+++ b/synapse/handlers/device.py
@@ -948,14 +948,12 @@ class DeviceListUpdater:
             devices = []
             ignore_devices = True
         else:
-            sid = await self.store.get_device_list_last_stream_id_for_remote(user_id)
+            stream_id = await self.store.get_device_list_last_stream_id_for_remote(user_id)
             cached_devices = await self.store.get_cached_devices_for_user(user_id)
 
-            # If this is the first time we've queried this user, then `sid is None` and
-            # `cached_devices` will be empty. If the remote user has no devices (i.e.
-            # `devices` is empty), we should cache this fact. For this reason, we skip
-            # only if `sid is not None`.
-            if sid is not None and cached_devices == {
+            # To ensure that a user with no devices is cached, we skip the resync only
+            # if we have a stream_id from previously writing a cache entry.
+            if stream_id is not None and cached_devices == {
                 d["device_id"]: d for d in devices
             }:
                 logging.info(
--- a/synapse/handlers/device.py
+++ b/synapse/handlers/device.py
@@ -948,8 +948,16 @@ class DeviceListUpdater:
             devices = []
             ignore_devices = True
         else:
+            sid = await self.store.get_device_list_last_stream_id_for_remote(user_id)
             cached_devices = await self.store.get_cached_devices_for_user(user_id)
-            if cached_devices == {d["device_id"]: d for d in devices}:
+
+            # If this is the first time we've queried this user, then `sid is None` and
+            # `cached_devices` will be empty. If the remote user has no devices (i.e.
+            # `devices` is empty), we should cache this fact. For this reason, we skip
+            # only if `sid is not None`.
+            if sid is not None and cached_devices == {
+                d["device_id"]: d for d in devices
+            }:
                 logging.info(
                     "Skipping device list resync for %s, as our cache matches already",
                     user_id,
--- a/synapse/handlers/federation_event.py
+++ b/synapse/handlers/federation_event.py
@@ -1008,8 +1008,6 @@ class FederationEventHandler:
             sender_key = event.content.get("sender_key")
 
             cached_devices = await self._store.get_cached_devices_for_user(event.sender)
-            if cached_devices is None:
-                cached_devices = {}
 
             resync = False  # Whether we should resync device lists.
 
--- a/synapse/handlers/federation_event.py
+++ b/synapse/handlers/federation_event.py
@@ -1008,6 +1008,8 @@ class FederationEventHandler:
             sender_key = event.content.get("sender_key")
 
             cached_devices = await self._store.get_cached_devices_for_user(event.sender)
+            if cached_devices is None:
+                cached_devices = {}
 
             resync = False  # Whether we should resync device lists.
 
--- a/synapse/storage/databases/main/devices.py
+++ b/synapse/storage/databases/main/devices.py
@@ -525,7 +525,7 @@ class DeviceWorkerStore(SQLBaseStore):
         """Get the devices (and keys if any) for remote users from the cache.
 
         Args:
-            query_list: List of (user_id, device_ids) pairs. If device_ids is
+            query_list: List of (user_id, device_ids), if device_ids is
                 falsey then return all device ids for that user.
 
         Returns:
@@ -555,10 +555,7 @@ class DeviceWorkerStore(SQLBaseStore):
                 device = await self._get_cached_user_device(user_id, device_id)
                 results.setdefault(user_id, {})[device_id] = device
             else:
-                user_devices = await self.get_cached_devices_for_user(user_id)
-                if user_devices is None:
-                    user_devices = {}
-                results[user_id] = user_devices
+                results[user_id] = await self.get_cached_devices_for_user(user_id)
 
         set_tag("in_cache", results)
         set_tag("not_in_cache", user_ids_not_in_cache)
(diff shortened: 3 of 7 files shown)


DMRobertson:
cc @squahtx

synapse/storage/databases/main/devices.py
```diff
@@ -573,16 +576,53 @@ async def _get_cached_user_device(self, user_id: str, device_id: str) -> JsonDic
...
+            FROM device_lists_remote_extremeties
+                LEFT JOIN device_lists_remote_cache USING(user_id)
+            WHERE user_id = ?
+        """
+        txn.execute(query, (user_id,))
+        devices: List[Tuple[str, Optional[str], Optional[str]]] = txn.fetchall()
+        if not devices:
+            return None
+        elif devices[0][1] is None:
+            return {}
```
clokep, line 621: Should we check the length here?

synapse/storage/databases/main/devices.py
```diff
@@ -573,16 +576,53 @@ async def _get_cached_user_device(self, user_id: str, device_id: str) -> JsonDic
...
+
+    def _get_cached_devices_for_user_txn(
+        self, txn: LoggingTransaction, user_id: str
+    ) -> Optional[Dict[str, JsonDict]]:
+        # Four cases:
+        # 1. No stream id, no cached devices. Query yields no rows. Return None.
+        # 2. No stream id, >= 1 cached devices. Invalid state. Query will yield no rows.
+        #    return None.
+        # 3. Stream id, no cached devices. Return empty dict. Query returns one row
+        #    (non-NULL stream_id, NULL, NULL). Return empty dict.
```
clokep, line 607: This says "Return empty dict." twice and I'm unsure if this is a 3a and 3b case, or if iti is a typo?

synapse/storage/databases/main/devices.py
```diff
@@ -573,16 +576,53 @@ async def _get_cached_user_device(self, user_id: str, device_id: str) -> JsonDic
...
+        # 4. Stream id, >= 1 cached devices. Query return 1 or more row
+        #    (non-NULL stream_id, non-NULL device_id, non-NULL content). Return dict.
+        query = """
+            SELECT stream_id, device_id, content
+            FROM device_lists_remote_extremeties
```
clokep, line 612: I have no idea if this is the proper usage of `device_lists_remote_extremeties` unfortunately.

tests/handlers/test_e2e_keys.py
```diff
@@ -831,3 +837,62 @@ def test_query_devices_remote_sync(self):
                 }
             },
         )
+
+    @parameterized.expand(
+        [([],), ([{"device_id": "device_1"}, {"device_id": "device_2"}],)]
```
clokep, line 842: It might be worth describing what the different inputs are here and why they matter.

synapse/handlers/federation_event.py
```diff
@@ -1008,6 +1008,8 @@ async def _process_received_pdu(
             sender_key = event.content.get("sender_key")
 
             cached_devices = await self._store.get_cached_devices_for_user(event.sender)
+            if cached_devices is None:
+                cached_devices = {}
```
clokep, line 1012: (Based on the comment above in `devicemessage.py`...) is this expected?

clokep (commented):
I think this generally looks OK, but I'm not confident the usage of the stream IDs makes sense.

synapse/storage/databases/main/devices.py
DMRobertson: My thinking is written down here: https://github.com/matrix-org/synapse/issues/11586#issuecomment-995042917. I'm also nervous though!

synapse/storage/databases/main/devices.py
DMRobertson: If `devices` is truthy (618) it has at least one entry. But having said that, FWIW I think this could be clearer if I made two separate queries: one to see if there's an entry in the extremities table, then another to read from the cache table. 

I wanted to write a single query here---in the spirit of the postgres book----as an exercise and as a bit of practice to see how it came out. I'm not really convinced by the result TBH. Maybe two queries is better. It's not like this is going to be an application bottleneck anyway.

synapse/storage/databases/main/devices.py
DMRobertson: Typo.

I think needing a long form comment to explain the query isn't a good sign for clarity and maintainability.

synapse/handlers/federation_event.py
DMRobertson: I don't think so, but I'm not all that sure to be honest. I wanted to at least err on the side of maintaining the existing behaviour. Similarly on line 558 below.

tests/handlers/test_e2e_keys.py
DMRobertson: Good idea, thanks!

DMRobertson:
> I think this generally looks OK, but I'm not confident the usage of the stream IDs makes sense.

FWIW I'm not interested in the specific stream ID as such. I'm using it in the query to detect that we've queried a homeserver for this user's devices before.

I agree that it's not great though. I have a bit of an icky feeling about the whole change the more I look at it. Perhaps sleeping on it will help.

DMRobertson:
> Perhaps sleeping on it will help.

To be explicit, perhaps sleeping on it will help me to later rework this and make it clearer.

synapse/storage/databases/main/devices.py
clokep: Right there will be at least one query, but what is it possible to get multiple rows and have one of them (the first one) but `<something>, NULL, NULL`? Or is that impossible from the query?

DMRobertson:
> Perhaps sleeping on it will help.

Having hibernated over the Christmas break, I think I should
- do away with the all-in-one query
- use `get_device_list_last_stream_id_for_remote` instead to fetch the stream_id, which is cached in-application
- only do so when we are considering updating the cache
  - so a generic read from the cache won't return `None`.

DMRobertson:
The commit history is a bit noisy, but I think this is a lot better now. I'll gamble that CI passes and stick it up for review.

synapse/storage/databases/main/devices.py
DMRobertson: The long query is gone now, so there's no need for this.

synapse/storage/databases/main/devices.py
DMRobertson: The changes to the query here have been reverted.

synapse/handlers/device.py
```diff
@@ -948,8 +948,16 @@ async def user_device_resync(
             devices = []
             ignore_devices = True
         else:
+            sid = await self.store.get_device_list_last_stream_id_for_remote(user_id)
```
clokep, line 951: I originally assumed `sid` = session ID, let's be more specific.

```suggestion
            stream_id = await self.store.get_device_list_last_stream_id_for_remote(user_id)
```

synapse/handlers/federation_event.py
DMRobertson: This is gone too after the revert. `get_cached_devices_for_user` will return an empty dict
- if we've never requested this user's devices,
- or if we have, but their HS returned an empty device list.

Instead, we distinguish between these two cases at the point where we check to see if the cache needs updating.

tests/handlers/test_e2e_keys.py
```diff
@@ -831,3 +837,89 @@ def test_query_devices_remote_sync(self):
...
+                e2e_handler.query_devices(
+                    request_body,
+                    timeout=10,
+                    from_user_id=local_user_id,
+                    from_device_id="some_device_id",
+                )
+            )
+
+            # We should not have made a second federation request.
+            mocked_federation_request.assert_called_once()
```
clokep, line 920: I would be clearer to me if we called `reset()` on the mock, then `assert_not_called()` I think? Functionally equivalent though, so whichever you prefer.

synapse/handlers/device.py
```diff
@@ -948,8 +948,16 @@ async def user_device_resync(
...
             cached_devices = await self.store.get_cached_devices_for_user(user_id)
-            if cached_devices == {d["device_id"]: d for d in devices}:
+
+            # If this is the first time we've queried this user, then `sid is None` and
+            # `cached_devices` will be empty. If the remote user has no devices (i.e.
+            # `devices` is empty), we should cache this fact. For this reason, we skip
+            # only if `sid is not None`.
```
clokep, line 957: I find this a bit wordy, I'm not sure this following explains it though:

> Ensure that a user with no devices is cached by checking if there's a stream ID.

tests/handlers/test_e2e_keys.py
```diff
@@ -831,3 +836,94 @@ def test_query_devices_remote_sync(self):
...
+        e2e_handler = self.hs.get_e2e_keys_handler()
+
+        # Pretend we're sharing a room with the user we're querying. If not,
+        # `_query_devices_for_destination` will return early.
+        mock_get_rooms = mock.patch.object(
+            self.store,
+            "get_rooms_for_user",
+            new_callable=mock.MagicMock,
+            return_value=make_awaitable(["some_room_id"]),
+        )
```
clokep, line 888: Why the change from using `patch` directly? I'm surprised this needs `new_callable` here, I've hardly ever seen that used. Why do we need to provide it?

DMRobertson: > Why the change from using patch directly?

I noticed that the file was already importing `mock` in a few places before I came along. I thought this was slightly more consistent.

> I'm surprised this needs new_callable here, I've hardly ever seen that used. Why do we need to provide it?

The previous version without `new_callable` worked fine on 3.7 but failed on 3.10 (and I'm guessing would have failed on 3.8--3.9). I thought I understood why, but since the tests are all failing I clearly didn't. Argh. Let me take another look.

clokep: From [the docs](https://docs.python.org/3.7/library/unittest.mock.html#unittest.mock.patch):

> *new_callable* allows you to specify a different class, or callable object, that will be called to create the *new* object. By default `MagicMock` is used.

So this sounds to me like we're just passing in the default.

DMRobertson: That's only the case in 3.7. In 3.8 and up:

> _new_callable_ allows you to specify a different class, or callable object, that will be called to create the _new_ object. By default `AsyncMock` is used for async functions and `MagicMock` for the rest.

DMRobertson: Yes, exactly: those are the failures in the checks for 0999ed8.

There's a new failure in 13e0aec caused by me overwriting `stream_id` after the rename! Fixed in 6fbf316, hopefully.

clokep: Yeah, I think we'll be able to clean-up a lot of code once we're only on 3.8! Thanks for the explanation!

DMRobertson: (I'd started writing out a wall of text to cover the mechanics in detail, but... fortunately you've seen this awaitable-awaitable before!)

clokep (approved):
I think this looks good.

PR #11588, opened 2021-12-15
Use mock from standard library

V02460:
~~As Python 3.6 is EOL, the backported version of the mock package is no longer needed and can be replaced by the one found in the standard library. Workarounds for emulating `AsyncMock` are no longer required either.~~

Synapse currently uses the backported non-standard library mock in one single test. This PR rewrites that test to no longer require the non-standard mock and removes the dependency from Synapse.

### DIFF
--- a/tests/storage/test_background_update.py
+++ b/tests/storage/test_background_update.py
@@ -12,8 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Use backported mock for AsyncMock support on Python 3.6.
-from mock import Mock
+from unittest.mock import Mock
 
 from twisted.internet.defer import Deferred, ensureDeferred
 
--- a/tests/storage/test_background_update.py
+++ b/tests/storage/test_background_update.py
@@ -116,11 +116,10 @@ class BackgroundUpdateControllerTestCase(unittest.HomeserverTestCase):
 
         # Mock out the AsyncContextManager
         class MockCM:
-            pass
+            __aenter__ = simple_async_mock(return_value=None)
+            __aexit__ = simple_async_mock(return_value=None)
 
         self._update_ctx_manager = MockCM
-        self._update_ctx_manager.__aenter__ = simple_async_mock(return_value=None)
-        self._update_ctx_manager.__aexit__ = simple_async_mock(return_value=None)
 
         # Mock out the `update_handler` callback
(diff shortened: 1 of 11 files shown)


clokep:
CI will fail on this until we remove Python 3.6, unfortunately.

I suspect we can also replace `make_awaitable`, but I'm not 100% sure on that one!

DMRobertson:
Related: #11588

clokep:
This looks good overall -- I'm removing this from the review queue for now since it is blocked on our CI dropping support for Python 3.6.

callahad:
(Note: that we _are_ open to PRs that drop 3.6 from our CI pipeline now that 1.49.0 is out)

H-Shay:
There is a PR open now for dropping 3.6 from CI: #11595

V02460:
Some corrections: I just noticed that `AsyncMock` is not available in Python 3.7 either, so removing `simple_async_mock` has to wait for another 1½ years >.<

Looking at the code now, it seems you went through this discussion before, but there is currently a single test in Synapse which requires a backported `Mock` on Python 3.7 (tests.storage.test_background_update.BackgroundUpdateControllerTestCase.test_controller). If that instance is changed, mock could still be removed – unless you want to keep your options.

V02460:
CI completes successfully now, even with Python 3.6!

tests/storage/test_background_update.py
```diff
@@ -116,14 +115,15 @@ def prepare(self, reactor, clock, homeserver):
...
-        self._update_ctx_manager.__aenter__ = Mock(
-            return_value=make_awaitable(None),
-        )
-        self._update_ctx_manager.__aexit__ = Mock(return_value=make_awaitable(None))
+        class MockCM:
+            pass
+
+        self._update_ctx_manager = MockCM
+        self._update_ctx_manager.__aenter__ = simple_async_mock(return_value=None)
+        self._update_ctx_manager.__aexit__ = simple_async_mock(return_value=None)
```
clokep, line 123: Would it work to define these on the class directly? That seems a bit cleaner!

V02460: Yes!

clokep (approved):
That test seems to be rather confusing, but this doesn't make it anymore so. Thanks for taking care of this! 👍

PR #11592, opened 2021-12-16
Do not bundle aggregations for APIs which shouldn't include them.

clokep:
This tightens up the API endpoints we return bundled aggregations from. We were incorrectly bundling them for `/search`, `/events`, and `/notifications`. This changes the default of the `serialize_event` method to *not bundle aggregations* since there's only a small list of endpoints that [MSC2675](https://github.com/matrix-org/matrix-doc/pull/2675) declares as having bundled aggregations:

> The following client-server APIs should bundle aggregations with events they return:
>
>- `GET /rooms/{roomId}/messages`
>- `GET /rooms/{roomId}/context/{eventId}`
>- `GET /rooms/{roomId}/event/{eventId}`
>- `GET /sync`, only for room sections in the response where `limited` field is `true`; this amounts to all rooms in the response if the `since` request parameter was not passed, also known as an initial sync.
>- `GET /relations`, as proposed in this MSC.
>
> Deprecated APIs like `/initialSync` and `/events/{eventId}` are *not* required to bundle aggregations.

This should have been part of #11478 but I missed it!

### DIFF
--- a/synapse/rest/admin/rooms.py
+++ b/synapse/rest/admin/rooms.py
@@ -745,13 +745,19 @@ class RoomEventContextServlet(RestServlet):
 
         time_now = self.clock.time_msec()
         results["events_before"] = await self._event_serializer.serialize_events(
-            results["events_before"], time_now
+            results["events_before"],
+            time_now,
+            bundle_aggregations=True,
         )
         results["event"] = await self._event_serializer.serialize_event(
-            results["event"], time_now
+            results["event"],
+            time_now,
+            bundle_aggregations=True,
         )
         results["events_after"] = await self._event_serializer.serialize_events(
-            results["events_after"], time_now
+            results["events_after"],
+            time_now,
+            bundle_aggregations=True,
         )
         results["state"] = await self._event_serializer.serialize_events(
             results["state"], time_now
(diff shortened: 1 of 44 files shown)


clokep:
Honestly -- I wonder if we should disable bundling by default and only enable it on the APIs that are declared to use it? It seems somewhat like a footgun to enable it by default.

clokep:
> Honestly -- I wonder if we should disable bundling by default and only enable it on the APIs that are declared to use it? It seems somewhat like a footgun to enable it by default.

Judging by @richvdh's 👍 I 'm going to do this.

synapse/rest/admin/rooms.py
```diff
@@ -745,13 +745,19 @@ async def on_GET(
 
         time_now = self.clock.time_msec()
         results["events_before"] = await self._event_serializer.serialize_events(
-            results["events_before"], time_now
+            results["events_before"],
+            time_now,
+            bundle_aggregations=True,
```
clokep, line 750: As these are admin APIs we can do whatever we want, but the intention is for them to match the Client-Server APIs, so I gave them the same behavior!

richvdh (approved):
seems sensible to me!

PR #11594, opened 2021-12-16
Add type hints to event_push_actions.

clokep:
I was taking a look at notification counts and got a bit side tracked adding type hints...I figured I should finish them up.

This is broken into 3 commits, the first adds type hints and the subsequent ones switch from `TypedDict` to `attrs` since the return values don't need to be dicts for any reason.

### DIFF
--- a/synapse/storage/databases/main/event_push_actions.py
+++ b/synapse/storage/databases/main/event_push_actions.py
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import logging
-from typing import TYPE_CHECKING, Dict, List, Optional, Tuple, Union
+from typing import TYPE_CHECKING, Any, Dict, List, Optional, Tuple, Union, cast
 
 import attr
 from typing_extensions import TypedDict
@@ -34,8 +34,11 @@ if TYPE_CHECKING:
 logger = logging.getLogger(__name__)
 
 
-DEFAULT_NOTIF_ACTION = ["notify", {"set_tweak": "highlight", "value": False}]
-DEFAULT_HIGHLIGHT_ACTION = [
+DEFAULT_NOTIF_ACTION: List[Union[dict, str]] = [
+    "notify",
+    {"set_tweak": "highlight", "value": False},
+]
+DEFAULT_HIGHLIGHT_ACTION: List[Union[dict, str]] = [
     "notify",
     {"set_tweak": "sound", "value": "default"},
     {"set_tweak": "highlight"},
@@ -56,7 +59,13 @@ class EmailPushAction(HttpPushAction):
     received_ts: Optional[int]
 
 
-def _serialize_action(actions, is_highlight):
+class NotifCounts(TypedDict):
+    notify_count: int
+    unread_count: int
+    highlight_count: int
+
+
+def _serialize_action(actions: List[Union[dict, str]], is_highlight: bool) -> str:
     """Custom serializer for actions. This allows us to "compress" common actions.
 
     We use the fact that most users have the same actions for notifs (and for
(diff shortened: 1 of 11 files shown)


synapse/storage/databases/main/event_push_actions.py
```diff
@@ -149,15 +154,15 @@ async def get_unread_event_push_actions_by_room_for_user(
...
+        txn: LoggingTransaction,
+        room_id: str,
+        user_id: str,
+        last_read_event_id: Optional[str],
+    ) -> NotifCounts:
         stream_ordering = None
 
         if last_read_event_id is not None:
-            stream_ordering = self.get_stream_id_for_event_txn(
+            stream_ordering = self.get_stream_id_for_event_txn(  # type: ignore[attr-defined]
```
clokep, line 165: I'm sure we could do something better here, but not sure it is worth blocking this PR for?

synapse/storage/databases/main/event_push_actions.py
```diff
@@ -539,12 +553,11 @@ async def remove_push_actions_from_staging(self, event_id: str) -> None:
         """
 
         try:
-            res = await self.db_pool.simple_delete(
```
clokep, line 542: This returns an `int`, but declares it returns `None`. The return value is never used.

synapse/storage/databases/main/event_push_actions.py
```diff
@@ -393,7 +405,7 @@ def get_after_receipt(txn):
             )
             args = [user_id, user_id, min_stream_ordering, max_stream_ordering, limit]
             txn.execute(sql, args)
-            return txn.fetchall()
+            return txn.fetchall()  # type: ignore[return-value]
```
clokep, line 408: I could `cast` this also, but that just duplicates code? 🤷

synapse/storage/databases/main/event_push_actions.py
```diff
@@ -972,32 +1010,42 @@ def f(txn):
...
-        try:
-            if action.get("set_tweak", None) == "highlight":
-                return action.get("value", True)
-        except AttributeError:
-            pass
+        if not isinstance(action, dict):
+            continue
+
+        if action.get("set_tweak", None) == "highlight":
+            return action.get("value", True)
```
clokep, line 1037: I'm fairly certain this is the same behavior but I suppose there could have been something else being passed in here that wasn't a `dict`, but did implement `get(...)`?

synapse/storage/databases/main/event_push_actions.py
```diff
@@ -57,6 +57,13 @@ class EmailPushAction(HttpPushAction):
     received_ts: Optional[int]
 
 
+@attr.s(slots=True, frozen=True, auto_attribs=True)
+class UserPushAction(EmailPushAction):
+    topological_ordering: int
+    highlight: bool
+    profile_tag: str
```
richvdh, line 64: I'm struggling to understand what this actually represents, and why it's different to the other types of PushAction. If you've managed to grok it, maybe you could add some docstring?

richvdh (approved):
seems reasonable to me!

synapse/storage/databases/main/event_push_actions.py
clokep: It seems to be required by the spec, for what it is worth: https://spec.matrix.org/latest/client-server-api/#get_matrixclientv3notifications

It is defined there as:

> The profile tag of the rule that matched this event.

I can add something along those lines if it is useful, but I'm not sure it really is. It might make more sense to say that these applies to the `/notifications` endpoint somewhere? But it applies to a store method so it gets a bit convoluted?

richvdh: In case it was unclear: by "this" I meant the entire `UserPushAction` class.

If `UserPushAction` corresponds to something in the spec, it would probably be worth saying that.

clokep: Ah, fair! I thought you meant `profile_tag` in particular. :)

clokep: I'll add some docstrings to the three classes then. 👍

Issue #11552, opened 2021-12-10
Python 3.6 / PostgreSQL 9.6 Deprecation Checklist / Chores


callahad:
- [x] Bump the version of Black we use for code formatting
- [x] Bump the `tool.black.target-version` in pyproject.toml
- [x] Run Black on all the things
- [x] Update our Trove classifiers to drop support for 3.6
- [x] Update CI configuration to stop testing on 3.6 (including updating the `olddeps` test run)
- [x] Update .deb build script to stop building for Ubuntu Bionic (18.04) ~~and Ubuntu Hirsute (21.04)~~
- [x] Get rid of the SyTest image that uses Bionic (and figure out what to replace it with)
- [x] Grep source and docs for anything referencing Python 3.6, Bionic, or Hirsute and excise it
- [x] Run pyupgrade
- [ ] ~~Consider taking advantage of `from __future__ import annotations` from [PEP 563](https://docs.python.org/3.10/whatsnew/3.7.html#pep-563-postponed-evaluation-of-annotations) to speed up application start?~~
- [x] Get rid of anything referencing PostgreSQL 9.6; our new minimum version is 10. (Check SyTest too)

richvdh:
>  Update .deb build script to stop building for Ubuntu Bionic (18.04) and Ubuntu Hirsute (21.04)

In theory we could support these versions of ubuntu via deadsnakes (https://launchpad.net/~deadsnakes/+archive/ubuntu/ppa). I'm conscious that 21.04 is still the current LTS version of ubuntu, so is probably quite popular. (It's what I use on my dev box, which is going to make things interesting...)

callahad:
> In theory we could support these versions of ubuntu via deadsnakes

Individuals are absolutely welcome to continue running Synapse on older Ubuntu releases, including via the deadsnakes PPA, but the Synapse team itself will not build third party Debian packages for those releases.

> I'm conscious that 21.04 is still the current LTS version of ubuntu

It's not :) 20.04 (Focal) is the current LTS, and we will still support it.

21.04 (Hirsute) is a normal release which hits end-of-life on January 1st:

![Ubuntu release lifecycle graph](https://user-images.githubusercontent.com/24193/145558641-ed7efdd5-1dfa-4215-99f8-c945184d6b39.png)

richvdh:
> >     I'm conscious that 21.04 is still the current LTS version of ubuntu
>
> It's not :) 20.04 (Focal) is the current LTS, and we will still support it.

[...]

PR #11595, opened 2021-12-16
Drop EOL python 3.6 from CI

H-Shay:
As the title states. Drops python 3.6 and postgres 9.6 from the CI and removes the python 3.6 environment from tox.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,5 @@
 [tox]
-envlist = packaging, py36, py37, py38, py39, check_codestyle, check_isort
+envlist = packaging, py37, py38, py39, check_codestyle, check_isort
 
 # we require tox>=2.3.2 for the fix to https://github.com/tox-dev/tox/issues/208
 minversion = 2.3.2
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,5 @@
 [tox]
-envlist = packaging, py37, py38, py39, check_codestyle, check_isort
+envlist = packaging, py37, py38, py39, py310, check_codestyle, check_isort
 
 # we require tox>=2.3.2 for the fix to https://github.com/tox-dev/tox/issues/208
 minversion = 2.3.2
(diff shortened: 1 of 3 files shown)


clokep:
Related to #11552.

tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = packaging, py36, py37, py38, py39, check_codestyle, check_isort
+envlist = packaging, py37, py38, py39, check_codestyle, check_isort
```
callahad, line 2: Huh, looks like we somehow missed adding py310 here... I wouldn't balk at sneaking that into this PR

DMRobertson: Though if I'm honest, I don't understand tox well enough to know what the ramifications of this are. Tests [always seem to use `toxenv: "py"`](https://github.com/matrix-org/synapse/blob/a4521ce0a8d252e77ca8bd261ecf40ba67511a31/.github/workflows/tests.yml#L81-L97), whatever that means

callahad: Using `py` is an implicit match to the currently running `python` version. Since our CI environment will itself supply the correct Python version for a test run, we don't need to enumerate it separately for Tox.

Where the envlist matters is running tox locally, where you want _Tox_ to handle juggling the requested Python versions.

(E.g., just running `tox` is equivalent to running `tox -e py37,py38,py39,check_codestyle,check_isort`... and which will only succeed if you have Python 3.7, 3.8, and 3.9 available on your local system.)

H-Shay: Happy to sneak in py310, will update.

richvdh (changes requested):
IIRC the `buster` builds of sytest use 3.6 as well?

callahad:
@richvdh Shouldn't the `buster` builds be using 3.7? https://packages.debian.org/buster/python3

richvdh:
argh. yes. sorry. So confused.

richvdh (approved):
lgtm. Sorry for the run-around!

H-Shay:
No worries! Thank you!

PR #11596, opened 2021-12-16
Update to the current version of Black and run it on Synapse codebase

H-Shay:
As the title states. Part of the cleanup from dropping Python 3.6: #11552.

### DIFF
--- a/changelog.d/11596.misc
+++ b/changelog.d/11596.misc
@@ -1,0 +1,1 @@
+Update black version and run it on all the files.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,7 +35,7 @@
         showcontent = true
 
 [tool.black]
-target-version = ['py37']
+target-version = ['py37', 'py38', 'py39', 'py310']
 exclude = '''
 
 (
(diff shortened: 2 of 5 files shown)


clokep:
It is unclear to me if this is meant to wait for #11595 or not? I think it is fine since CI passes, but theoretically it might be possible that this would cause invalid syntax in 3.6?

callahad:
Either PR can be merged independently; looks like Black is only changing whitespace in this upgrade.

clokep:
> Either PR can be merged independently; looks like Black is only changing whitespace in this upgrade.

Yep, I looked at the diff, just wasn't clear to me if it is was "safe" to merge based on targeting the wrong version now.

callahad:
Looks like the only [new syntax features in 3.7](https://docs.python.org/3.10/whatsnew/3.7.html) are PEP563's `from __future__ import annotations` and marking `async` and `await` as reserved keywords.

callahad:
From reading https://github.com/psf/black/issues/751 it looks like we "should" be enumerating all Pythons versions we want Black to support for our code. So changing:

```diff
- target-version = ['py37']
+ target-version = ['py37', 'py38', 'py39', 'py310']
```

Which is daft. But if that's how the tool wants to be used...

richvdh (approved):
lgtm!

PR #11599, opened 2021-12-17
Add documentation for missing worker types.

### LINKED ISSUE #9046: Document that `to_device` streams and encryption endpoints can be on workers
c.f. https://github.com/matrix-org/synapse/pull/9044

### LINKED ISSUE #9477: Worker docs don't mention how to route `/typing`

### LINKED ISSUE #10010: Document that presence can be on a worker
Specifically, the new experimental feature to move presence to a worker doesn't seem to be documented anywhere.

clokep:
Fixes #9046, fixes #9477, fixes #10010.

Most of the endpoints were taken from the matrix.org configuration. I'm less confident about the presence one, but I think it is correct.

I tried to make it a bit clearer that you ideally want to route the endpoints associated with a stream writer to that stream writer, but not 100% sure I got there in a clear manner. 😄


docs/workers.md
```diff
@@ -238,19 +241,40 @@ expressions:
...
+    # Account data requests (for the account_data stream writer)
+    ^/_matrix/client/(api/v1|r0|v3|unstable)/.*/tags
+    ^/_matrix/client/(api/v1|r0|v3|unstable)/.*/account_data
+
+    # Receipts requests (for the receipts stream writer)
+    ^/_matrix/client/(api/v1|r0|v3|unstable)/rooms/.*/receipt
+    ^/_matrix/client/(api/v1|r0|v3|unstable)/rooms/.*/read_markers
+
+    # Presence requests (for the presence stream writer)
+    ^/_matrix/client/(api/v1|r0|v3|unstable)/presence/.*/status$
```
richvdh, line 276: hrm, this feels a bit confusing, because we're talking about stream writers before we introduce them.

(also, there's something to be said for this being a static list of endpoints, that can be C&Ped wholesale).

Would it make more sense for this to move into the "Stream writers" section, with separate lists for each stream writer?

docs/workers.md
```diff
@@ -178,8 +178,11 @@ recommend the use of `systemd` where available: for information on setting up
 
 ### `synapse.app.generic_worker`
 
-This worker can handle API requests matching the following regular
-expressions:
+This worker can handle API requests matching the following regular expressions.
+
+Note that stream writers should have their endpoints routed directly to them for
+efficiency. If this is not done, the requests will be proxied to the proper
+worker.
```
richvdh, line 185: this is maybe a simplification? `POST` and `PUT` requests will be proxied, but `GET` requests will be handled directly?

docs/workers.md
richvdh: hum, or maybe not, because even if we leave these things on the master, then the generic worker will proxy them to the right place?

in which case, maybe we should actually maintain two copies of this part of the list, one here under "generic worker" where we comment on the fact that `POST`s and `PUT`s will be proxied, and another *under* the "stream writers" section?

richvdh (commented):
a few thoughts. Thanks for doing this!

docs/workers.md
clokep: I think I missed that detail when looking at the implementation, thanks!

docs/workers.md
clokep: I struggled with how to include them!

Having them twice is probably clearest, but there's concern about drift between them. I agree I don't like introducing this before the "stream writers" section though.

clokep: The only reasonable thing I could think of was to list them twice, lease let me know what you think!

docs/workers.md
```diff
@@ -347,6 +368,25 @@ stream_writers:
...
+
+The following endpoints should be routed direclty to the workers configured as
+stream writers for the `events` stream:
+
+    ^/_matrix/client/(api/v1|r0|v3|unstable)/rooms/.*/redact
+    ^/_matrix/client/(api/v1|r0|v3|unstable)/rooms/.*/send
+    ^/_matrix/client/(api/v1|r0|v3|unstable)/rooms/.*/state/
+    ^/_matrix/client/(api/v1|r0|v3|unstable)/rooms/.*/(join|invite|leave|ban|unban|kick)$
+    ^/_matrix/client/(api/v1|r0|v3|unstable)/join/
+    ^/_matrix/client/(api/v1|r0|v3|unstable)/profile/
```
richvdh, line 388: I don't think this is right. It's at least not what we do on matrix.org. These requests come in to a generic worker (in the form of the `event_creator`s), which does a bunch of work before passing them onto the things handling the events streams (the `event_persister`s).

docs/workers.md
```diff
@@ -359,6 +399,43 @@ stream_writers:
         - event_persister2
 ```
 
+##### The `typing` stream
+
+The following endpoints should be routed direclty to the workers configured as
```
richvdh, line 404:
```suggestion
The following endpoints should be routed directly to the workers configured as
```

likewise several times below.

docs/workers.md
```diff
@@ -251,6 +257,23 @@ expressions:
...
+    # Account data requests
+    ^/_matrix/client/(api/v1|r0|v3|unstable)/.*/tags
+    ^/_matrix/client/(api/v1|r0|v3|unstable)/.*/account_data
+
+    # Receipts requests
+    ^/_matrix/client/(api/v1|r0|v3|unstable)/rooms/.*/receipt
+    ^/_matrix/client/(api/v1|r0|v3|unstable)/rooms/.*/read_markers
+
+    # Presence requests
+    ^/_matrix/client/(api/v1|r0|v3|unstable)/presence/.*/status$
```
richvdh, line 275: given this is the only thing starting with `/presence`, maybe just:

```suggestion
    # Presence requests
    ^/_matrix/client/(api/v1|r0|v3|unstable)/presence/
```

docs/workers.md
```diff
@@ -347,6 +368,25 @@ stream_writers:
     events: event_persister1
 ```
 
+Each of the streams have associated endpoints which should have `POST` and `PUT`
+requests routed to the workers handling that stream. Otherwise, those requests
+will be proxied to the proper worker.
```
richvdh, line 373: Sadly, I don't think this is correct. Examples:

* if you send a `PUT /typing` request to the wrong worker, it will 500.
* if you send a `PUT /sendToDevice` request or a `POST /receipt` request to the wrong worker... I dunno. I think it will still work? It won't proxy it, anyway.

I think it's roughly correct for `account_data`, `tags` and `presence`

richvdh (changes requested):
:man_shrugging:

docs/workers.md
clokep: Oh, so the `event_persister` itself doesn't handle any endpoints! 👍

docs/workers.md
clokep: That's quite confusing! My initial thought -- is this a bug? Are those supposed to be proxied to the proper workers and we just don't? It would be much clearer to document if the various endpoints behaved the same.

docs/workers.md
clokep: Would it make sense to do the same for the presence stream (below)?

docs/workers.md
richvdh: yeah. No idea what the intention is here.

docs/workers.md
richvdh: yes.

docs/workers.md
clokep: @erikjohnston Any idea what's *supposed* to happen here?

docs/workers.md
```diff
@@ -178,8 +178,10 @@ recommend the use of `systemd` where available: for information on setting up
 
 ### `synapse.app.generic_worker`
 
-This worker can handle API requests matching the following regular
-expressions:
+This worker can handle API requests matching the following regular expressions.
+`GET` requests will be handled directly while `POST` and `PUT` requests will be
```
erikjohnston, line 182: This isn't strictly true. Some `GET` requests are proxied and not all `POST` or `PUT` requests are. I'm not really sure that this is useful information, and not something that is actionable by server admins :shrug:

docs/workers.md
```diff
@@ -251,6 +257,23 @@ expressions:
...
+    ^/_matrix/client/(api/v1|r0|v3|unstable)/.*/tags
+    ^/_matrix/client/(api/v1|r0|v3|unstable)/.*/account_data
+
+    # Receipts requests
+    ^/_matrix/client/(api/v1|r0|v3|unstable)/rooms/.*/receipt
+    ^/_matrix/client/(api/v1|r0|v3|unstable)/rooms/.*/read_markers
+
+    # Presence requests
+    ^/_matrix/client/(api/v1|r0|v3|unstable)/presence/.*/status$
+
```
erikjohnston, line 276: Have these endpoints been added to the SyTest worker config so that we actually have tests that these work?

docs/workers.md
erikjohnston: Ideally I think we want requests to be proxied to the right worker, so that the end game supports load balancers just round robining all requests to all workers (even if that is not the most efficient way of doing it).

docs/workers.md
```diff
@@ -251,6 +257,23 @@ expressions:
     ^/_matrix/client/(api/v1|r0|v3|unstable)/join/
     ^/_matrix/client/(api/v1|r0|v3|unstable)/profile/
 
+    # Typing requests
+    ^/_matrix/client/(api/v1|r0|v3|unstable)/rooms/.*/typing
```
erikjohnston, line 261: This isn't true unless the worker is configured to handle typing? The idea with this list of regexes is that you can just straight up copy and paste this into your load balancers

docs/workers.md
clokep: Almost all of them were, see matrix-org/sytest#1186.

docs/workers.md
clokep: Is this kind of the same conversation as https://github.com/matrix-org/synapse/pull/11599#discussion_r779768039 then?

docs/workers.md
clokep: This makes this almost impossible to document or explain what is happening. 😢

erikjohnston: Can't we say something to the effect of "These endpoints can be routed to any worker. If a worker is set up to handle a stream then additional endpoints may be required to route to said worker, refer to the `streams` section below for further information."? And then for each stream type list what endpoints should be routed to them?

docs/workers.md
```diff
@@ -238,6 +238,12 @@ expressions:
     ^/_matrix/client/(api/v1|r0|v3|unstable)/joined_rooms$
     ^/_matrix/client/(api/v1|r0|v3|unstable)/search$
 
+    # Encryption requests
+    ^/_matrix/client/(api/v1|r0|v3|unstable)/keys/query$
+    ^/_matrix/client/(api/v1|r0|v3|unstable)/keys/changes$
```
clokep, line 243: These two endpoints aren't actually available at `api/v1`, seems silly to route those anywhere...

clokep: It seems like there's quite a few endpoints that this is true for...

richvdh: arguably we shouldn't be documenting any `api/v1` or `unstable` prefixes. Any clients using them are buggy...

docs/workers.md
clokep: I see, this endpoint does get registered but doesn't work unless it is the typing stream.

docs/workers.md
clokep: I've added some text like that, hopefully it is clearer now.

I'm *fairly* certain all the endpoints I documented can be sent to any worker and will be proxied properly. 👍

erikjohnston (approved):
This looks good to me now! Thanks Patrick.

It may be worth waiting for someone else to double check the actual wording makes sense? You know I'm blind to words :/

clokep:
> It may be worth waiting for someone else to double check the actual wording makes sense? You know I'm blind to words :/

Let's see if @richvdh has any opinions! 😄

docs/workers.md
```diff
@@ -225,19 +227,23 @@ expressions:
...
+    ^/_matrix/client/(r0|v3|unstable)/publicised_groups$
+    ^/_matrix/client/(r0|v3|unstable)/publicised_groups/
     ^/_matrix/client/(api/v1|r0|v3|unstable)/rooms/.*/event/
     ^/_matrix/client/(api/v1|r0|v3|unstable)/joined_rooms$
     ^/_matrix/client/(api/v1|r0|v3|unstable)/search$
 
+    # Encryption requests
+    ^/_matrix/client/(r0|v3|unstable)/keys/query$
+    ^/_matrix/client/(r0|v3|unstable)/keys/changes$
+    ^/_matrix/client/(r0|v3|unstable)/keys/claim
```
richvdh, line 244:
```suggestion
    ^/_matrix/client/(r0|v3|unstable)/keys/claim$
```

docs/workers.md
```diff
@@ -225,19 +227,23 @@ expressions:
...
+    ^/_matrix/client/(r0|v3|unstable)/room_keys
```
richvdh, line 245: it might be better to list the endpoints explicitly?

```suggestion
    ^/_matrix/client/(r0|v3|unstable)/room_keys/keys$
    ^/_matrix/client/(r0|v3|unstable)/room_keys/version$
    ^/_matrix/client/(r0|v3|unstable)/room_keys/version/
```

docs/workers.md
```diff
@@ -347,6 +365,12 @@ stream_writers:
     events: event_persister1
 ```
 
+Each of the streams have associated endpoints which should have requests routed
```
richvdh, line 368: `events` doesn't seem to have any:

```suggestion
Some of the streams have associated endpoints which should have requests routed
```

docs/workers.md
```diff
@@ -178,8 +178,10 @@ recommend the use of `systemd` where available: for information on setting up
 
 ### `synapse.app.generic_worker`
 
-This worker can handle API requests matching the following regular
-expressions:
+This worker can handle API requests matching the following regular expressions.
+These endpoints can be routed to any worker. If a worker is set up to handle a
+stream then additional endpoints may be required to route to said worker, refer
+to the [stream writers](#stream-writers) section below for further information.
```
richvdh, line 184: I'm a bit confused about this. Are we saying that `sendToDevice` and things can go to any `generic_worker`, *unless* there is a worker which is defined to handle the `to_device` stream, in which case you have to route `sendToDevice` to that specific worker?

richvdh (commented):
I think the words are generally fine. I'm still finding this a bit confusing though.

docs/workers.md
richvdh: hrm yes.

maybe just:

```suggestion
    ^/_matrix/client/(r0|v3|unstable)/room_keys/
```

docs/workers.md
```diff
@@ -178,8 +178,10 @@ recommend the use of `systemd` where available: for information on setting up
 
 ### `synapse.app.generic_worker`
 
-This worker can handle API requests matching the following regular
-expressions:
+This worker can handle API requests matching the following regular expressions.
+These endpoints can be routed to any worker. If a worker is set up to handle a
+stream then additional endpoints may be required to route to said worker: refer
+to the [stream writers](#stream-writers) section below for further information.
```
richvdh, line 184:
```suggestion
These endpoints can be routed to any worker. If a worker is set up to handle a
stream then, for maximum efficiency, additional endpoints should be routed to that worker: refer
to the [stream writers](#stream-writers) section below for further information.
```

richvdh (approved):
lgtm otherwise

clokep:
Thank you both for the reviews! 🎉 I know we can improve this even more, but happy to see this merged!

PR #11602, opened 2021-12-17
Return JSON errors instead of HTML errors for unknown resources. 

### LINKED ISSUE #11600: Synapse responds with a 404 with HTML content for unknown URL prefixes
See https://github.com/vector-im/element-web/issues/19738#issuecomment-996804209 for some background, but if you request an endpoint for Synapse that doesn't have a prefix registered (see [here](https://github.com/matrix-org/synapse/blob/71f9966f2790c6b24281bb9f109bff28ff05d962/synapse/app/homeserver.py#L185-L196)) then Synapse responds with a 404 including HTML content (the default Twisted 404 handler).

This wasn't an issue when adding new `/unstable` or `/r0` endpoints since the prefix was registered, but the `/hierarchy` endpoint is the first `/v1` endpoint we had. This will also bite us once we need to rev an endpoint to `/v2` (or `/v4`) since neither of those are registered right now.

clokep:
Fixes #11600 by returning JSON errors with CORS for unknown URLs from prefixes.

You can test this before/after this change via:

```bash
$ curl http://localhost:8080/_matrix/client/v2/foo
$ curl -vv -X OPTIONS http://localhost:8080/_matrix/client/v2/foo
$ curl http://localhost:8080/_matrix/client/v2/
$ curl http://localhost:8080/_matrix/client/
```

These now all return something like:

```json
{"errcode":"M_UNRECOGNIZED","error":"Unrecognized request"}
```

instead of:

```html
<html>
  <head><title>404 - No Such Resource</title></head>
  <body>
    <h1>No Such Resource</h1>
    <p>Sorry. No luck finding that resource.</p>
  </body>
</html>
```

### DIFF
--- a/changelog.d/11602.bugfix
+++ b/changelog.d/11602.bugfix
@@ -1,0 +1,1 @@
+Fix a long-standing bug that some unknown endpoints would return HTML 404 errors instead of JSON 400 errors.
--- a/changelog.d/11602.bugfix
+++ b/changelog.d/11602.bugfix
@@ -1,1 +1,1 @@
-Fix a long-standing bug that some unknown endpoints would return HTTP 404 errors instead of JSON `M_UNRECOGNIZED` errors.
+Fix a long-standing bug that some unknown endpoints would return HTML error pages instead of JSON `M_UNRECOGNIZED` errors.
--- a/changelog.d/11602.bugfix
+++ b/changelog.d/11602.bugfix
@@ -1,1 +1,1 @@
-Fix a long-standing bug that some unknown endpoints would return HTML 404 errors instead of JSON 400 errors.
+Fix a long-standing bug that some unknown endpoints would return HTTP 404 errors instead of JSON `M_UNRECOGNIZED` errors.
--- a/synapse/app/homeserver.py
+++ b/synapse/app/homeserver.py
@@ -27,6 +27,7 @@ import synapse
 import synapse.config.logger
 from synapse import events
 from synapse.api.urls import (
+    CLIENT_API_PREFIX,
     FEDERATION_PREFIX,
     LEGACY_MEDIA_PREFIX,
     MEDIA_R0_PREFIX,
@@ -192,13 +193,7 @@ class SynapseHomeServer(HomeServer):
 
             resources.update(
                 {
-                    "/_matrix/client/api/v1": client_resource,
-                    "/_matrix/client/r0": client_resource,
-                    "/_matrix/client/v1": client_resource,
-                    "/_matrix/client/v3": client_resource,
-                    "/_matrix/client/unstable": client_resource,
-                    "/_matrix/client/v2_alpha": client_resource,
-                    "/_matrix/client/versions": client_resource,
+                    CLIENT_API_PREFIX: client_resource,
                     "/.well-known": well_known_resource(self),
                     "/_synapse/admin": AdminRestResource(self),
                     **build_synapse_client_resource_tree(self),
--- a/synapse/http/server.py
+++ b/synapse/http/server.py
@@ -559,6 +559,17 @@ class OptionsResource(resource.Resource):
         return super().getChildWithDefault(path, request)
 
 
+class MissingResource(resource.Resource):
+    """Similar to resource.NoResource, but returns a JSON error with proper CORS headers."""
+
+    def render(self, request: SynapseRequest) -> int:
+        return_json_error(failure.Failure(UnrecognizedRequestError()), request)
+        return NOT_DONE_YET
+
+    def getChild(self, name: str, request: Request) -> resource.Resource:
+        return self
+
+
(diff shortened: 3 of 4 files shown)


reivilibre:
Overall sensible, but wondering: is it correct to return this error code for unknown paths? We seem to already use that error code to mean an array of different things depending on the endpoint and I would be weary about doing this without checking because it then becomes hard for a client/homeserver to tell what the problem is (is the endpoint not recognised? a login method? ...).

I would ideally like to see a spec reference that supports this (even if that means inventing one now, I guess :)).

(To state the unstated but fairly obvious when reading the code: this changes 404s to 400s)

clokep:
> Overall sensible, but wondering: is it correct to return this error code for unknown paths?

We already do this for all unknown paths under known prefixes, e.g. `curl http://localhost:8080/_matrix/client/v1/foo` or `curl http://localhost:8080/_matrix/client/r0/foo`. This makes it consistent.

> I would ideally like to see a spec reference that supports this (even if that means inventing one now, I guess :)).

`M_UNRECOGNIZED` is in [the spec](https://spec.matrix.org/latest/client-server-api/#other-error-codes) under "error codes specific to certain endpoints":

> `M_UNRECOGNIZED` The server did not understand the request.

> We also do things with the result codes from other servers over federation to know if they implement certain endpoints or not; will this break that?

I'll double check what this does to federation!

reivilibre:
> We already do this for all unknown paths under known prefixes

Ah, that was something I didn't realise! In that case I hope this is probably fine as is? :)

changelog.d/11602.bugfix
```diff
@@ -0,0 +1 @@
+Fix a long-standing bug that some unknown endpoints would return HTML 404 errors instead of JSON 400 errors.
```
reivilibre, line 1: I'd probably change to this:
```suggestion
Fix a long-standing bug that some unknown endpoints would return HTTP 404 errors instead of JSON `M_UNRECOGNIZED` errors.
```

clokep:
> > We already do this for all unknown paths under known prefixes
> 
> Ah, that was something I didn't realise! In that case I hope this is probably fine as is? :)

Should have mentioned it better in the description -- the issue went into detail about it. 😄

reivilibre (approved):
LGTM

reivilibre:
> the issue went into detail about it. smile

I did actually read that part but only understand what was gotten at with hindsight ... oh well :)

changelog.d/11602.bugfix
clokep: Errr, yes. Too many abbreviations in my head! Thanks!

clokep: No, this was right the way it was. The point of this PR is to return JSON errors instead of HTML ones.

richvdh:
I do wonder if we should make the expected behaviour here explicit in the spec (for both C-S and S-S). In particular, it would be helpful if HTTP clients knew what they could expect when a server doesn't yet implement a new endpoint. For example, synapse code like this:

```
    def _is_unknown_endpoint(
        self, e: HttpResponseException, synapse_error: Optional[SynapseError] = None
    ) -> bool:
        # There is no good way to detect an "unknown" endpoint.
        #
        # Dendrite returns a 404 (with no body); synapse returns a 400
        # with M_UNRECOGNISED.
        return e.code == 404 or (
            e.code == 400 and synapse_error.errcode == Codes.UNRECOGNIZED
        )
```

clokep:
> I do wonder if we should make the expected behaviour here explicit in the spec (for both C-S and S-S). In particular, it would be helpful if HTTP clients knew what they could expect when a server doesn't yet implement a new endpoint.

I agree! I can propose something... do you see having something in the spec as blocking this?

richvdh:
> do you see having something in the spec as blocking this?

no, not really... though if you put it through the MSC process you may find there are opinions expressed there that mean we end up reworking this PR.

clokep:
>> We also do things with the result codes from other servers over federation to know if they implement certain endpoints or not; will this break that?
>
> I'll double check what this does to federation!

This changes nothing for the federation APIs since those are currently simply registered under the `_matrix/federation` prefix. Thus we don't end up with the same problem where we have "unknown" federation versions.

Although that gives me an idea which might be simpler than this!

clokep:
@reivilibre I've updated this a bit to simplify. The only difference in behavior now is that `/_matrix` will still return the 404 with an HTML page. I'm not sure why anyone would hit that URL directly though so I don't see that as a big deal?

reivilibre (approved):
Looks good and simple

clokep:
> > do you see having something in the spec as blocking this?
> 
> no, not really... though if you put it through the MSC process you may find there are opinions expressed there that mean we end up reworking this PR.

I found an old issue and cross-linked at least: matrix-org/matrix-doc#1492

PR #11603, opened 2021-12-17
Add opentracing types

H-Shay:
This one was a little tricky due to the conditional import. After some consultation it was decided it was prudent to ignore the case where `opentracing` is set to `None`, rather than doing a larger refactor that is difficult to test. Happy to do it a different way but I figure this way would allow the types to be used and still checks that we are using `opentracing` correctly in this file, and doesn't involve a non-negligible refactor.

### DIFF
--- a/changelog.d/11603.misc
+++ b/changelog.d/11603.misc
@@ -1,0 +1,1 @@
+Add opentracing type stubs and fix associated mypy errors.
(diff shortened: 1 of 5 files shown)


richvdh (approved):
this is a great start, which I'm going to go ahead and merge, since I have other things that build on it :)

Issue #11390, opened 2021-11-18
`ObservableDeferred`'s API is a mess


richvdh:
... and I hate it.

In particular:
 * it doesn't `consumeErrors` by default, which in 99% of cases leads to log spam about unhandled failures
 * it does some awful `__getattr__` / `__setattr__` hijinks to "transparently" pass through the methods on the underlying `Deferred`, which makes it hard to type correctly and generally leads to magical behaviour.
 * I find it weird that you can fire an `ObservableDeferred` either by directly calling its `callback`/`errback` methods, or by calling them on the deferred you pass into it.

I would like to replace it with a much simpler implementation. Here is a proposal for how we can do it without rewriting everything at once.

 1. Define an abstract base class which `ObservableDeferred` implements, called `DeferredObservable`, which just exposes the `observe` method:
     ```python
     class DeferredObservable(Generic[T], metaclass=abc.ABCMeta):
         @abc.abstractmethod
         def observe(self) -> "Deferred[T]": ...
     ```
 2. Replace consumers (rather than creators) of `ObservableDeferred` with `DeferredObservable`
 3. Create a new class `SimpleDeferredObservable` which implementents `DeferredObservable`, and just has a `callback`, an `errback`, and an `observe` method. Both methods should return `None`.
 4. Replace uses of `ObservableDeferred` which currently create a brand new `Deferred` with `SimpleDeferredObservable`
 5. Create a new class `DeferredObserver` which wraps a `SimpleDeferredObservable`:
     ```python
     class DeferredObserver(DeferredObservable[T]):
         def __init__(self): 
             self._observable = SimpleDeferredObservable[T]()
        def observe(self) -> Deferred[T]:
             return self._observable.observe()
        def chain_to_deferred(self, deferred: Deferred[T]) -> None:
             deferred.addCallbacks(self._observable.callback, self._observable.errback)
     ```
 6. Replace uses  of `ObservableDeferred` which chain to an existing Deferred with `DeferredObserver`
 7. Remove now-redundant `ObservableDeferred`

PR #11607, opened 2021-12-20
Improve opentracing support for `ResponseCache`

richvdh:
This adds some opentracing annotations to `ResponseCache`, to make it easier to see what's going on; in particular, it adds a link back to the initial trace which is actually doing the work of generating the response.

Since I was refactoring a bit of `ResponseCache` anyway, I also made a start on #11390 by factoring out an AbstractObservableDeferred.

Reviewable commit-by-commit (ish; c302d9389 has some important fixes to 
0d1dcb9)

### DIFF
--- a/synapse/handlers/room.py
+++ b/synapse/handlers/room.py
@@ -172,7 +172,7 @@ class RoomCreationHandler:
         user_id = requester.user.to_string()
 
         # Check if this room is already being upgraded by another person
-        for key in self._upgrade_response_cache.pending_result_cache:
+        for key in self._upgrade_response_cache.keys():
             if key[0] == old_room_id and key[1] != user_id:
                 # Two different people are trying to upgrade the same room.
                 # Send the second an error.
--- a/synapse/util/caches/response_cache.py
+++ b/synapse/util/caches/response_cache.py
@@ -12,7 +12,17 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import logging
-from typing import Any, Awaitable, Callable, Dict, Generic, Iterable, Optional, TypeVar
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Awaitable,
+    Callable,
+    Dict,
+    Generic,
+    Iterable,
+    Optional,
+    TypeVar,
+)
 
 import attr
 
@@ -29,6 +39,9 @@ from synapse.util.async_helpers import AbstractObservableDeferred, ObservableDef
 from synapse.util.caches import register_cache
 
 logger = logging.getLogger(__name__)
+
+if TYPE_CHECKING:
+    import opentracing
 
 # the type of the key in the cache
 KV = TypeVar("KV")
@@ -69,7 +82,7 @@ class ResponseCacheEntry:
     easier to cache Failure results.
     """
 
-    opentracing_span_context: Any
+    opentracing_span_context: "Optional[opentracing.SpanContext]"
     """The opentracing span which generated/is generating the result"""
 
 
@@ -128,7 +141,7 @@ class ResponseCache(Generic[KV]):
         self,
         context: ResponseCacheContext[KV],
         deferred: "defer.Deferred[RV]",
-        opentracing_span_context: Any,
+        opentracing_span_context: "Optional[opentracing.SpanContext]",
     ) -> ResponseCacheEntry:
         """Set the entry for the given key to the given deferred.
 
@@ -221,7 +234,7 @@ class ResponseCache(Generic[KV]):
             if cache_context:
                 kwargs["cache_context"] = context
 
-            span_context = None
+            span_context: Optional[opentracing.SpanContext] = None
 
             async def cb() -> RV:
                 # NB it is important that we do not `await` before setting span_context!
--- a/synapse/util/caches/response_cache.py
+++ b/synapse/util/caches/response_cache.py
@@ -19,6 +19,10 @@ import attr
 from twisted.internet import defer
 
 from synapse.logging.context import make_deferred_yieldable, run_in_background
+from synapse.logging.opentracing import (
+    start_active_span,
+    start_active_span_follows_from,
+)
 from synapse.util import Clock
 from synapse.util.async_helpers import AbstractObservableDeferred, ObservableDeferred
 from synapse.util.caches import register_cache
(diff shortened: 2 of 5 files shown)


synapse/util/caches/response_cache.py
```diff
@@ -63,10 +74,7 @@ class ResponseCache(Generic[KV]):
     """
 
     def __init__(self, clock: Clock, name: str, timeout_ms: float = 0):
-        # This is poorly-named: it includes both complete and incomplete results.
-        # We keep complete results rather than switching to absolute values because
-        # that makes it easier to cache Failure results.
-        self.pending_result_cache: Dict[KV, ObservableDeferred] = {}
+        self.result_cache: Dict[KV, ResponseCacheEntry] = {}
```
richvdh, line 84: since we're changing the type, we may as well take the opportunity to fix the name.

synapse/handlers/room.py
```diff
@@ -172,7 +172,7 @@ async def upgrade_room(
         user_id = requester.user.to_string()
 
         # Check if this room is already being upgraded by another person
-        for key in self._upgrade_response_cache.pending_result_cache:
+        for key in self._upgrade_response_cache.result_cache:
```
richvdh, line 175: this is the only reference to `ObservableDeferred.pending_result_cache` outside of `ObservableDeferred` itself.

synapse/handlers/room.py
```diff
@@ -172,7 +172,7 @@ async def upgrade_room(
         user_id = requester.user.to_string()
 
         # Check if this room is already being upgraded by another person
-        for key in self._upgrade_response_cache.pending_result_cache:
+        for key in self._upgrade_response_cache.keys():
```
richvdh, line 175: this is the only reference to `ResponseCache.pending_result_cache` outside `ResponseCache` itself.

synapse/util/caches/response_cache.py
```diff
@@ -63,10 +74,7 @@ class ResponseCache(Generic[KV]):
     """
 
     def __init__(self, clock: Clock, name: str, timeout_ms: float = 0):
-        # This is poorly-named: it includes both complete and incomplete results.
-        # We keep complete results rather than switching to absolute values because
-        # that makes it easier to cache Failure results.
-        self.pending_result_cache: Dict[KV, ObservableDeferred] = {}
+        self._result_cache: Dict[KV, ResponseCacheEntry] = {}
```
richvdh, line 84: since we're changing its type (so have to change all references), we might as well fix the name while we're here.

synapse/util/caches/response_cache.py
```diff
@@ -89,43 +97,34 @@ def keys(self) -> Iterable[KV]:
         Note that the returned iterator is not safe in the face of concurrent execution:
         behaviour is undefined if `wrap` is called during iteration.
         """
-        return self.pending_result_cache.keys()
+        return self._result_cache.keys()
 
-    def get(self, key: KV) -> Optional[defer.Deferred]:
+    def _get(self, key: KV) -> Optional[ResponseCacheEntry]:
```
richvdh, line 109: `get` was unused outside `wrap`.

synapse/util/caches/response_cache.py
```diff
@@ -75,75 +90,82 @@ def __init__(self, clock: Clock, name: str, timeout_ms: float = 0):
...
+        opentracing_span_context: Any,
+    ) -> ResponseCacheEntry:
         """Set the entry for the given key to the given deferred.
 
         *deferred* should run its callbacks in the sentinel logcontext (ie,
         you should wrap normal synapse deferreds with
         synapse.logging.context.run_in_background).
 
-        Returns a new Deferred (which also doesn't follow the synapse logcontext rules).
-        You will probably want to make_deferred_yieldable the result.
```
clokep, line 116: Was this changed at some point in the past to no longer be true? I don't think I see changes in here related to this?

synapse/util/caches/response_cache.py
```diff
@@ -54,6 +58,20 @@ class ResponseCacheContext(Generic[KV]):
...
+class ResponseCacheEntry:
+    result: AbstractObservableDeferred
+    """The (possibly incomplete) result of the operation.
+
+    Note that we continue to store an ObservableDeferred even after the operation
+    completes (rather than switching to an immediate value), since that makes it
+    easier to cache Failure results.
+    """
+
+    opentracing_span_context: Any
```
clokep, line 71: Is there a better type we can use for opentracing spans or is this awaiting some of the work that @H-Shay is doing?

clokep (commented):
Looks reasonable overall, but tests are failing!

synapse/util/caches/response_cache.py
richvdh: no, it was true until this PR. see https://github.com/matrix-org/synapse/pull/11607/files#diff-4e303a929ca8d0022d41884af5d1dd65997f3c21e596ee09b66d806144770b25R168 which changes the result from a `Deferred` to a `ResponseCacheEntry`.

synapse/util/caches/response_cache.py
```diff
@@ -75,75 +90,82 @@ def __init__(self, clock: Clock, name: str, timeout_ms: float = 0):
...
-        Returns a new Deferred (which also doesn't follow the synapse logcontext rules).
-        You will probably want to make_deferred_yieldable the result.
-
         Args:
             context: Information about the cache miss
             deferred: The deferred which resolves to the result.
+            opentracing_span_context: An opentracing span wrapping the calculation
 
         Returns:
             A new deferred which resolves to the actual result.
```
richvdh, line 144:
```suggestion
        Returns:
            The cache entry object.
```

synapse/util/caches/response_cache.py
clokep: I was not specific enough -- is it useful to somewhere say something about the synapse logcontext rules still or is that no longer needed?

clokep (approved):
I think this makes sense, then opentracing APIs are a bit hard to grasp though.

synapse/util/caches/response_cache.py
richvdh: well, it's no longer relevant in `_set`, at least.  ObservableDeferred has its own set of guidance on the logcontext rules (see https://github.com/matrix-org/synapse/blob/d4033b2be6eae2eda9fd5791f7efd80f396963d2/synapse/util/async_helpers.py#L83-L86) - though that does make me realise I should probably add some notes about it on `AbstractObservableDeferred`.

clokep: That was a bit what I was wondering. I did notice that we call `make_deferred_yieldable` internally. 👍

richvdh: right, done in 6a89434.

PR #11608, opened 2021-12-20
Cap the number of in-flight requests for state from a single group

reivilibre:
Small incremental change above #10870 that prevents too many concurrent requests being in-flight for a single state group, by just requesting the whole state filter from the database.

### DIFF
--- a/synapse/storage/databases/state/store.py
+++ b/synapse/storage/databases/state/store.py
@@ -294,13 +294,15 @@ class StateGroupDataStore(StateBackgroundUpdateStore, SQLBaseStore):
                 # to cover our StateFilter and give us the state we need.
                 break
 
-        if state_filter_left_over != StateFilter.none():
-            if len(inflight_requests) >= MAX_INFLIGHT_REQUESTS_PER_GROUP:
-                # There are too many requests for this group.
-                # To prevent even more from building up, we request the whole
-                # state filter to guarantee that we can be reused by any subsequent
-                # requests for this state group.
-                return (), StateFilter.all()
+        if (
+            state_filter_left_over != StateFilter.none()
+            and len(inflight_requests) >= MAX_INFLIGHT_REQUESTS_PER_GROUP
+        ):
+            # There are too many requests for this group.
+            # To prevent even more from building up, we request the whole
+            # state filter to guarantee that we can be reused by any subsequent
+            # requests for this state group.
+            return (), StateFilter.all()
 
         return reusable_requests, state_filter_left_over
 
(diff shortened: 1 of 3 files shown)


synapse/storage/databases/state/store.py
```diff
@@ -288,6 +294,14 @@ def _get_state_for_group_gather_inflight_requests(
                 # to cover our StateFilter and give us the state we need.
                 break
 
+        if state_filter_left_over != StateFilter.none():
+            if len(inflight_requests) >= MAX_INFLIGHT_REQUESTS_PER_GROUP:
+                # There are too many requests for this group.
+                # To prevent even more from building up, we request the whole
+                # state filter to guarantee that we can be reused by any subsequent
+                # requests for this state group.
+                return (), StateFilter.all()
```
clokep, line 303: Probably clearer without nesting. I also tweaked the comments a bit, but I'm unsure if the changes are correct.

```suggestion
        if state_filter_left_over != StateFilter.none() and len(inflight_requests) >= MAX_INFLIGHT_REQUESTS_PER_GROUP:
            # There are too many in-flight requests for this group.
            # To prevent even more from building up, request the whole
            # state to guarantee that it can be reused by any subsequent
            # requests for this state group.
            return (), StateFilter.all()
```

tests/storage/databases/test_state_store.py
```diff
@@ -131,3 +131,69 @@ def test_duplicate_requests_deduplicated(self) -> None:
...
+
+    def test_inflight_requests_capped(self) -> None:
+        """
+        Tests that the number of in-flight requests is capped to 5.
+
+        - requests several pieces of state separately
+          (5 to hit the limit, 1 to 'shunt out', another that
+        - checks to see that the torrent of requests is shunted out by
+          rewriting one of the filters as the 'all' state filter
+        - requests after that one do not cause any additional queries
```
clokep, line 143: The bullets here seem incorrect?

tests/storage/databases/test_state_store.py
```diff
@@ -131,3 +131,69 @@ def test_duplicate_requests_deduplicated(self) -> None:
...
+        """
+        CAP_COUNT = 5
```
clokep, line 145: Should this import `MAX_INFLIGHT_REQUESTS_PER_GROUP`?

tests/storage/databases/test_state_store.py
```diff
@@ -131,3 +131,69 @@ def test_duplicate_requests_deduplicated(self) -> None:
...
+
+        reqs = []
+
+        # Request 7 different keys (1..=7) of the `some.state` type.
```
clokep, line 149: It is a bit confusing to use `1..=7` in this context.

clokep (commented):
Seems OK overall, just a couple confusing comments!

clokep (approved):

PR #11610, opened 2021-12-20
Order in-flight state group queries in biggest-first order

reivilibre:
The rough idea is that we should favour re-using a small number of broad state filters that are already in-flight, rather than needing to wait for them ALL.

Builds on #10870 and #11608.

### DIFF
--- a/synapse/storage/databases/state/store.py
+++ b/synapse/storage/databases/state/store.py
@@ -85,7 +85,7 @@ def state_filter_rough_priority_comparator(
     have any particular meaning. The representation may also change
 
     The current implementation returns a tuple of the form:
-        * -1 for include_others, -0 otherwise
+        * -1 for include_others, 0 otherwise
         * -(number of entries in state_filter.types)
     """
     return -int(state_filter.include_others), -len(state_filter.types)
--- a/synapse/storage/databases/state/store.py
+++ b/synapse/storage/databases/state/store.py
@@ -25,6 +25,7 @@ from typing import (
 )
 
 import attr
+from sortedcontainers import SortedDict
 
 from twisted.internet import defer
 
@@ -147,7 +148,7 @@ class StateGroupDataStore(StateBackgroundUpdateStore, SQLBaseStore):
         # Current ongoing get_state_for_groups in-flight requests
         # {group ID -> {StateFilter -> ObservableDeferred}}
         self._state_group_inflight_requests: Dict[
-            int, Dict[StateFilter, AbstractObservableDeferred[StateMap[str]]]
+            int, SortedDict[StateFilter, AbstractObservableDeferred[StateMap[str]]]
         ] = {}
 
         def get_max_state_group_txn(txn: Cursor) -> int:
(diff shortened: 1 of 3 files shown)


synapse/storage/databases/state/store.py
```diff
@@ -279,7 +300,12 @@ def _get_state_for_group_gather_inflight_requests(
 
         # The list of ongoing requests which will help narrow the current request.
         reusable_requests = []
-        for (request_state_filter, request_deferred) in inflight_requests.items():
+
+        # Iterate over existing requests in roughly biggest-first order.
+        # reversed(inflight_requests) has an efficient iterator implementation,
+        # but reversed(inflight_requests.items()) does not, sadly.
```
clokep, line 306: Not sure if it helps, but you could negate the values of `state_filter_rough_size_comparator` (or return 0 for `include_other` / 1 otherwise; `-len(state_filter.types)`) to "automatically" reverse this and not need to worry about the perf of `reversed`.

synapse/storage/databases/state/store.py
```diff
@@ -72,6 +73,26 @@ def __len__(self) -> int:
         return len(self.delta_ids) if self.delta_ids else 0
 
 
+# Return type of state_filter_rough_size_comparator. Must be comparable.
+StateFilterRoughSizeComparator = Tuple[int, int]
```
clokep, line 77: I think this is used only once, so doesn't seem very helpful to have it as a separate type?

synapse/storage/databases/state/store.py
reivilibre: Yeah, maybe that will be better. Let's give it a shot.

synapse/storage/databases/state/store.py
```diff
@@ -72,6 +73,24 @@ def __len__(self) -> int:
...
+    """
+    Returns a comparable value that roughly indicates the relative size of this
+    state filter compared to others.
+    'Larger' state filters should sort first when using ascending order, so
+    this is essentially the opposite of 'size'.
+    It should be treated as a rough guide only and should not be interpreted to
+    have any particular meaning. The representation may also change
+
+    The current implementation returns a tuple of the form:
+        * -1 for include_others, -0 otherwise
```
clokep, line 88:
```suggestion
        * -1 for include_others, 0 otherwise
```

tests/storage/databases/test_state_store.py
```diff
@@ -350,3 +355,100 @@ def test_inflight_requests_capped(self) -> None:
...
+            ),
+            state_filter_rough_priority_comparator(
+                state_filters[smallest_state_filter_idx]
+            ),
+            "Test invalid: bigger state filter is not actually bigger.",
+        )
+
+        # Spawn the initial two requests
+        for state_filter in state_filters:
+            ensureDeferred(
```
clokep, line 424: I find the bare `ensureDeferred` here surprising, but I think it is so we "schedule" things to run, but don't run want them to complete yet?

clokep (approved):

tests/storage/databases/test_state_store.py
reivilibre: Yeah, that was my thinking. We don't actually care about the results of these for once. It's not something I'd do outside of a test.

Issue #11611, opened 2021-12-20
large initialsync may never complete


richvdh:
For a sufficiently large account, it's possible that we will never be able to serialise the response before the client times out. This has recently been made a lot worse by #11478 which makes the serialisation process much slower, but it was a potential problem even before that.

The *intention* is that it shouldn't matter too much if it takes a long time to generate a sync response, because once we finally manage it, we will cache the response, which we can then serve up quickly the next time the client makes a request. However, this oversimplifies the situation.

We can basically consider there are four steps in serving a /sync response:

1. Figure out which rooms and events should be in the response. The end result of this step is a `SyncResult` object, which is cached.
2. "Serialise" each event in the response. This is where we turn the events from the internal python objects (`FrozenEvent`) into plain JSON dicts which match the format specified by the C-S API; in particular we add the `unsigned.age` field (which is the main reason this is done after caching); as of #11478 it is also the point at which we add bundled aggregations. The result of this step is a (potentially huge) `JsonDict` (ie, a `dict` containing other simple json-serializable values)
3. JSON-serialise the response dict to a `bytes`.
4. Finally, we start streaming the response to the client.

Since the result of step 1 is cached (at least on `matrix.org`), we can effectively neglect it here (though it is important to remember that it is only cached for 2 minutes... more on that later); so the steps of concern to us are steps 2 and 3, which must complete in less than the client's request timeout.

Some notes on timeouts:
 * Cloudflare times out requests if the server doesn't start sending its response within 100 seconds of the request.
 * However, haproxy doesn't notice this until we start trying to send back the response - so Synapse still sees the request as active until our haproxy times out the request, which happens after 180 seconds.
 * Before we start each of steps 2 and 3, we check that the client connection is still up, and skip that step if not. (Step 4 is aborted as soon as the connection drops.)

**In short**: we need to serialise the `FrozenEvent`s to `dict`s (step 2) and JSON-serialise the `dict`s to `bytes` within 100 seconds. 

That should be plenty though, right? Wrong.

---

The first problem is that #11478 makes step 2 take *224* seconds (assuming all the relevant aggregations are already cached from the database. The first time through, when we have do database activity it takes over 800 seconds):

![image](https://user-images.githubusercontent.com/1389908/146765239-29585531-ccb4-40e0-9a1e-f0355e10043e.png)

https://jaeger.int.matrix.org/trace/1a723cd9877a1cc9

[...]

PR #11612, opened 2021-12-20
Bundle aggregations outside of the serialization method

clokep:
This is some partial work towards #11611, the main benefit of which is that the event serialization process is now synchronous and does not access the database.

This requires the callers which wish to include bundled aggregations to manually fetch them and pass them in (as opposed to just passing a flag).

~~Based on #11592~~

This will be the basis for some future work for #11611:

* Including the bundled aggregations in the response cache for sync -- see #11659.
* Fetching the bundled aggregations of multiple events at once (to reduce database round trips) -- see #11660 for doing this with edits.

### DIFF
--- a/synapse/events/utils.py
+++ b/synapse/events/utils.py
@@ -14,17 +14,7 @@
 # limitations under the License.
 import collections.abc
 import re
-from typing import (
-    TYPE_CHECKING,
-    Any,
-    Callable,
-    Dict,
-    Iterable,
-    List,
-    Mapping,
-    Optional,
-    Union,
-)
+from typing import Any, Callable, Dict, Iterable, List, Mapping, Optional, Union
 
 from frozendict import frozendict
 
@@ -36,9 +26,6 @@ from synapse.util.async_helpers import yieldable_gather_results
 from synapse.util.frozenutils import unfreeze
 
 from . import EventBase
-
-if TYPE_CHECKING:
-    from synapse.server import HomeServer
 
 # Split strings on "." but not "\." This uses a negative lookbehind assertion for '\'
 # (?<!stuff) matches if the current position in the string is not preceded
@@ -385,17 +372,12 @@ class EventClientSerializer:
     clients.
     """
 
-    def __init__(self, hs: "HomeServer"):
-        self.store = hs.get_datastore()
-        self._msc1849_enabled = hs.config.experimental.msc1849_enabled
-        self._msc3440_enabled = hs.config.experimental.msc3440_enabled
-
     async def serialize_event(
         self,
         event: Union[JsonDict, EventBase],
         time_now: int,
         *,
-        bundle_aggregations: bool = False,
+        bundle_aggregations: Optional[Dict[str, JsonDict]] = None,
         **kwargs: Any,
     ) -> JsonDict:
         """Serializes a single event.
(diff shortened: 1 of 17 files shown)


clokep:
Another downside of this is that the cached `EventBase` values become slightly mutable, which might be unexpected!

clokep:
Comparing this to the steps in #11611, I think that this would help be moving the calculation of bundled aggregations from step 2 to step 1, thus ending up with them cached. This would still take over 800 seconds the first time, but should now be a no-op once they're cached (instead of taking 224 seconds). I think we can improve the queries there as well to be a bit more efficient!

synapse/storage/databases/main/relations.py
```diff
@@ -515,6 +545,98 @@ def _get_if_user_has_annotated_event(txn: LoggingTransaction) -> bool:
             "get_if_user_has_annotated_event", _get_if_user_has_annotated_event
         )
 
+    async def _get_bundled_aggregation_for_event(
```
DMRobertson, line 548: Is this brand new logic? Ah, no, it's moved from `_injected_bundled_aggregations` in the next commit.

synapse/events/utils.py
```diff
@@ -418,66 +400,41 @@ async def serialize_event(
...
-        # Do not bundle aggregations if any of the following at true:
-        #
-        # * Support is disabled via the configuration or the caller.
-        # * The event is a state event.
-        # * The event has been redacted.
-        if (
-            self._msc1849_enabled
-            and bundle_aggregations
-            and not event.is_state()
-            and not event.internal_metadata.is_redacted()
```
DMRobertson, line 431: This bit moved to the top of the helper function in the previous commit.

synapse/events/utils.py
```diff
@@ -418,66 +400,41 @@ async def serialize_event(
...
-        if annotations.chunk:
-            aggregations[RelationTypes.ANNOTATION] = annotations.to_dict()
-
-        references = await self.store.get_relations_for_event(
-            event_id, room_id, RelationTypes.REFERENCE, direction="f"
-        )
-        if references.chunk:
-            aggregations[RelationTypes.REFERENCE] = references.to_dict()
+        # Make a copy in-case the object is cached.
+        aggregations = aggregations.copy()
```
DMRobertson, line 431: The changes in this function seem to be:

- the logic which builds `aggregations` is pulled out to the helper in the previous commit
- aggregations are precalculated and passed into this function
- any side effects from the old code are maintained: read the relevant data from `aggregations` and kick off those same side effects

DMRobertson, line 431: Do we need this? Ahh yes: there's a mutation on +469.

DMRobertson (commented):
At a high level this looks like a sensible way to pull out the computation of aggregations from the serialisation logic. I did find this all a little daunting though!

> I think that this would help be moving the calculation of bundled aggregations from step 2 to step 1, thus ending up with them cached.

I don't quite follow this though. Which things are now cached that weren't before?

> Another downside of this is that the cached EventBase values become slightly mutable, which might be unexpected!

Where is this happening, sorry?

clokep:
> At a high level this looks like a sensible way to pull out the computation of aggregations from the serialisation logic. I did find this all a little daunting though!
> 
> > I think that this would help be moving the calculation of bundled aggregations from step 2 to step 1, thus ending up with them cached.
> 
> I don't quite follow this though. Which things are now cached that weren't before?

They aren't *yet*, but #11659 does this (and builds on this).

> > Another downside of this is that the cached EventBase values become slightly mutable, which might be unexpected!
> 
> Where is this happening, sorry?

This was an old comment about the prior approach, should have deleted it, sorry about that! You can ignore this comment.

synapse/storage/databases/main/relations.py
clokep: Yes, it is mostly copied! Sorry about not making that clear!

synapse/events/utils.py
clokep: My hope is that it is identical logic, yep! This still does the serialization of the aggregations, but that's perfectly reasonable I think.

clokep: I wondered if it would make more sense to make the input to this an attrs class so that the types are much clearer, which would mean we would build the dictionary here instead. Would that be clearer? (I think I'd rather do that as a follow-up since it is a decent amount of changes, but could do it here if you'd like!)

DMRobertson: I think that'd be nice, but not crucial. Sounds like it'd be best as a separate change to me.

DMRobertson (approved):
Thanks for explaining, I think this looks good.

OOI how much test coverage do we have of this bundling logic? (I.e. how reassuring is the successful CI run here?)

clokep:
> OOI how much test coverage do we have of this bundling logic? (I.e. how reassuring is the successful CI run here?)

The main test-case for this is https://github.com/matrix-org/synapse/blob/6c68e874b1eafe75db51e46064c0d3af702b4358/tests/rest/client/test_relations.py#L455-L582

This definitely failed quite a few times until I got it right. 😄

PR #11615, opened 2021-12-20
Expose the registered device ID from the `register_appservice_user` test helper.

reivilibre:
This is part of the E2E AS work, spun out onto develop for sanity's sake.

PR #11616, opened 2021-12-20
Allow specifying the application service-specific `user_id` parameter in the `join` test helper.

reivilibre:
Spun out from E2E AS OTK work.


changelog.d/11616.misc
```diff
@@ -0,0 +1 @@
+Allow specifying the application service-specific `user_id` parameter in the `join` test helper.
```
clokep, line 1: Should this have the same changelog as #11615?

tests/rest/client/utils.py
```diff
@@ -147,12 +147,20 @@ def invite(self, room=None, src=None, targ=None, expect_code=200, tok=None):
             expect_code=expect_code,
         )
 
-    def join(self, room=None, user=None, expect_code=200, tok=None):
+    def join(
+        self,
+        room=None,
+        user=None,
+        expect_code=200,
+        tok=None,
```
clokep, line 155: It would be nice to fill in these type hints while we are here! If that sets up crazy mypy work though then feel free to ignore this comment!

tests/rest/client/utils.py
```diff
@@ -217,15 +226,24 @@ def change_membership(
...
         path = "/_matrix/client/r0/rooms/%s/state/m.room.member/%s" % (room, targ)
+        next_arg_char = "?"
+
         if tok:
-            path = path + "?access_token=%s" % tok
+            path += "?access_token=%s" % tok
+            next_arg_char = "&"
+
+        if appservice_user_id:
+            path += f"{next_arg_char}user_id={appservice_user_id}"
```
clokep, line 246: I think you want to use [`urlencode`](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlencode) here instead.

clokep (changes requested):
Seems like a reasonable change to make.

tests/rest/client/utils.py
reivilibre: Perfect!

changelog.d/11616.misc
reivilibre: I don't really know, truthfully

clokep: I think it should, something like "Refactor test helpers for application services." seems fine.

Ideally if the actual changes land in the same release we can reword them there to point to the actual improvement / bugfix.

tests/rest/client/utils.py
```diff
@@ -217,15 +227,25 @@ def change_membership(
...
+        url_params: Dict[str, str] = {}
+
         if tok:
-            path = path + "?access_token=%s" % tok
+            url_params["access_token"] = tok
+
+        if appservice_user_id:
+            url_params["user_id"] = appservice_user_id
+
+        path += "?" + urlencode(url_params)
```
clokep, line 248: It probably doesn't matter much, but if we want to be perfectly consistent with past behavior we could wrap this in an `if url_params`.

clokep (approved):

PR #11617, opened 2021-12-20
Add support for MSC3202: sending one-time key counts and fallback key usage states to Application Services.

reivilibre:
It's difficult to divide this one up any further (apart from minor things ~ 10 lines for config, etc, or leaving things unimplemented as stubs); if you can see a way that would do it effectively then let me know and I'll try to do that.

The general idea here is:
- Application Services will now receive one-time key counts and fallback key usage states for their users' devices
- The idea we settled on in #10653 is as follows:
    - Compute the set of 'interesting users' to the application service for this transaction, hereby defined as:
        - the AS's users that are in rooms which receive an event in this transaction; and
        - the Application Service `sender_localpart` (unsure if this has a more proper name, but it's vaguely the 'main user ID' for the application service and is important enough that it should not miss OTK/FBK changes)
     - Retrieve counts / FBK states for all of those and send them out

As discussed in #10653, doing it this way is easier implementation-wise (since otherwise we would need to somehow track changes) and is chattier than needed.

This should largely be a commit-by-commit review, though there are a few 'boring' commits.

[MSC3202 is available here.](https://github.com/matrix-org/matrix-doc/pull/3202)


synapse/appservice/scheduler.py
```diff
@@ -202,15 +215,92 @@ async def _send_request(self, service: ApplicationService) -> None:
...
                     return
 
+                one_time_key_counts: Optional[TransactionOneTimeKeyCounts] = None
+                unused_fallback_keys: Optional[TransactionUnusedFallbackKeys] = None
+
+                if (
+                    self._msc3202_transaction_extensions_enabled
+                    and service.msc3202_transaction_extensions
+                ):
+                    # Lazily compute the one-time key counts and fallback key
```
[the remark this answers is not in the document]
reivilibre: Yeah, misleading word.

synapse/config/experimental.py
```diff
@@ -59,3 +59,9 @@ def read_config(self, config: JsonDict, **kwargs):
         self.msc2409_to_device_messages_enabled: bool = experimental.get(
             "msc2409_to_device_messages_enabled", False
         )
+
+        # Portion of MSC3202 related to transaction extensions:
+        # sending one-time key counts and fallback key usage to application services.
+        self.msc3202_transaction_extensions: bool = experimental.get(
+            "msc3202_transaction_extensions", False
+        )
```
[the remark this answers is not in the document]
reivilibre: Happy to merge if you think that's best — I wasn't right sure because they're really quite separate features that could easily have been separate MSCs, but not fussed either way.

synapse/storage/databases/main/appservice.py
```diff
@@ -335,12 +359,16 @@ def _get_oldest_unsent_txn(txn):
 
         events = await self.get_events_as_list(event_ids)
 
+        # TODO: should we recalculate one-time key counts and unused fallback
+        #       key counts here?
```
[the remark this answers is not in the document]
reivilibre: Bad comment. Plan for now is to leave this to a later piece of work focussed on 'reliability' (which is somewhat more important for to-device messages than OTKs/UFBKs as you ever only care about the latest state for the latter .

Have changed comment to more accurately reflect the plan.

synapse/appservice/scheduler.py
```diff
@@ -202,15 +215,92 @@ async def _send_request(self, service: ApplicationService) -> None:
...
+                        to_device_messages_to_send,
+                        one_time_key_counts,
+                        unused_fallback_keys,
                     )
                 except Exception:
                     logger.exception("AS request failed")
         finally:
             self.requests_in_flight.discard(service.id)
 
+    async def _determine_interesting_users_for_msc3202_otk_counts_and_fallback_keys(
```
erikjohnston, line 252: Can we move this into `_compute_msc3202_otk_counts_and_fallback_keys`? Rather than having two functions, one of which has a horrifically long function name :)

synapse/storage/databases/main/end_to_end_keys.py
```diff
@@ -439,6 +443,100 @@ def _count_e2e_one_time_keys(txn: LoggingTransaction) -> Dict[str, int]:
...
+
+            result: TransactionOneTimeKeyCounts = {}
+
+            for user_id, device_id, algorithm, count in txn:
+                device_count_by_algo = result.setdefault(user_id, {}).setdefault(
+                    device_id, {}
+                )
+                if algorithm is not None:
+                    # algorithm will be None if this device has no keys.
+                    device_count_by_algo[algorithm] = count
```
erikjohnston, line 481:
```suggestion
                # algorithm will be None if this device has no keys.
                if algorithm is not None:
                    device_count_by_algo = result.setdefault(user_id, {}).setdefault(
                        device_id, {}
                    )
                    device_count_by_algo[algorithm] = count
```

Rather than creating empty dictionaries?

synapse/storage/databases/main/end_to_end_keys.py
```diff
@@ -439,6 +443,100 @@ def _count_e2e_one_time_keys(txn: LoggingTransaction) -> Dict[str, int]:
...
+
+            result: TransactionUnusedFallbackKeys = {}
+
+            for user_id, device_id, algorithm in txn:
+                device_unused_keys = result.setdefault(user_id, {}).setdefault(
+                    device_id, []
+                )
+                if algorithm is not None:
+                    # algorithm will be None if this device has no keys.
+                    device_unused_keys.append(algorithm)
```
erikjohnston, line 532: Ditto to above here

erikjohnston (commented):
Otherwise LGTM

synapse/storage/databases/main/end_to_end_keys.py
reivilibre: I need to page in the MSC again (so good on you for mentioning, this could definitely use a comment), but I think we *want* the empty dicts to signal 'no keys'.

reivilibre: Actually: the MSC says these can be omitted, but I think it's wrong. [I've opened a question to double check](https://github.com/matrix-org/matrix-doc/pull/3202/files#r807866130).

synapse/storage/databases/main/end_to_end_keys.py
reivilibre: I believe this is the same situation as unused fallback keys; if you run out of OTKs for a particular algorithm then you need to hear about it.

erikjohnston (approved):

synapse/appservice/api.py
```diff
@@ -262,6 +268,16 @@ async def push_bulk(
...
             )
 
+        if service.msc3202_transaction_extensions:
+            if one_time_key_counts:
+                body[
+                    "org.matrix.msc3202.device_one_time_key_counts"
+                ] = one_time_key_counts
+            if unused_fallback_keys:
+                body[
+                    "org.matrix.msc3202.device_unused_fallback_keys"
```
Half-Shot, line 278: Having tried to play with this just now, I think this is wrong. The MSC states [`device_unused_fallback_key_types`](https://github.com/matrix-org/matrix-spec-proposals/blob/travis/msc/otk-dl-appservice/proposals/3202-encrypted-appservices.md#proposal)

PR #11618, opened 2021-12-20
Improve opentracing for incoming HTTP requests

richvdh:
The main point of this PR is to ensure that the opentracing span for an incoming HTTP request stays open until we *really* finish dealing with it (including encoding the response and sending back the resultant bytes).

Doing so requires a bit of refactoring, to make `SynapseRequest` responsible for closing the span, as well as removing some redundant stuff from the federation handler.

Also adds a bit of extra opentracing logging for the stages of handling a request.

Suggest reviewing commit-by-commit.

### DIFF
--- a/synapse/federation/transport/server/_base.py
+++ b/synapse/federation/transport/server/_base.py
@@ -22,13 +22,11 @@ from synapse.api.urls import FEDERATION_V1_PREFIX
 from synapse.http.server import HttpServer, ServletCallback
 from synapse.http.servlet import parse_json_object_from_request
 from synapse.http.site import SynapseRequest
-from synapse.logging import opentracing
 from synapse.logging.context import run_in_background
 from synapse.logging.opentracing import (
-    SynapseTags,
+    set_tag,
     span_context_from_request,
-    start_active_span,
-    tags,
+    start_active_span_follows_from,
     whitelisted_homeserver,
 )
 from synapse.server import HomeServer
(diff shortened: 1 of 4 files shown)


synapse/federation/transport/server/_base.py
```diff
@@ -279,29 +277,19 @@ async def new_func(
...
+            # if the origin is authenticated and whitelisted, link to its span context
             context = None
             if origin and whitelisted_homeserver(origin):
                 context = span_context_from_request(request)
 
-            scope = start_active_span(
-                "incoming-federation-request", child_of=context, tags=request_tags
+            scope = start_active_span_follows_from(
+                "incoming-federation-request", contexts=(context,) if context else ()
             )
```
richvdh, line 290: this changes the "incoming-federation-request" span so that its parent is no longer the outgoing request of the remote server, but rather the regular per-incoming-request span that is started in `AsyncResource`. Instead, the outgoing request span is set as a "reference" on the incoming request span.

Given that it's extremely unusual for us to have a single Jaeger instance which contains two federating servers, this seems by far the more useful approach.

synapse/federation/transport/server/_base.py
```diff
@@ -279,30 +277,19 @@ async def new_func(
...
-                    request, "incoming-federation-request", tags=request_tags
-                )
-            else:
-                scope = start_active_span(
-                    "incoming-federation-request", tags=request_tags
-                )
+                context = span_context_from_request(request)
 
-            with scope:
-                opentracing.inject_response_headers(request.responseHeaders)
```
clokep, line 304: For my own reference, this is also called via `trace_servlet` which is used by `_AsyncResource` which (I think) is an eventual super-class of all federation servlets, so we were doing work twice when we didn't need to?

clokep (approved):
LGTM!

synapse/federation/transport/server/_base.py
richvdh: yes, that's right. Sorry I didn't make it clear!

PR #11619, opened 2021-12-20
Various opentracing enhancements

richvdh:
This is a grab-bag of extra opentracing spans around the codebase, created in the process of debugging #11611.

Review commit-by-commit.

### DIFF
--- a/synapse/api/auth.py
+++ b/synapse/api/auth.py
@@ -32,7 +32,7 @@ from synapse.appservice import ApplicationService
 from synapse.events import EventBase
 from synapse.http import get_request_user_agent
 from synapse.http.site import SynapseRequest
-from synapse.logging import opentracing as opentracing
+from synapse.logging.opentracing import active_span, force_tracing, start_active_span
 from synapse.storage.databases.main.registration import TokenLookupResult
 from synapse.types import Requester, StateMap, UserID, create_requester
 from synapse.util.caches.lrucache import LruCache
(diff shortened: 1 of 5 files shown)


synapse/api/auth.py
```diff
@@ -149,6 +149,42 @@ async def get_user_by_req(
...
+                    # request tracing is enabled for this user, so we need to force it
+                    # tracing on for the parent span (which will be the servlet span).
+                    #
+                    # It's too late for the get_user_by_req span to inherit the setting,
+                    # so we also force it on for that.
+                    force_tracing()
+                    force_tracing(parent_span)
+                parent_span.set_tag(
+                    "authenticated_entity", requester.authenticated_entity
+                )
```
clokep, line 169: I had to convince myself that `authenticated_entity` was the same here:

* For appservices `authenticated_entity` isn't set during creation of the requester and we report the tag as `user_id`; this was essentially inlined logic from [`create_requester`](https://github.com/matrix-org/synapse/blob/ecfcd9bbbed62f8ca5d42bd8ba02b696e9079573/synapse/types.py#L183-L184)
* For non-appservices, this was set to `user_info.token_owner`, which is the same thing that is passed as the `authenticated_entity` in `create_requester`.

synapse/http/server.py
```diff
@@ -759,7 +761,20 @@ async def _async_write_json_to_request_in_thread(
...
+        # much that contributed.
+        if opentracing_span:
+            opentracing_span.log_kv({"event": "scheduled"})
+        res = json_encoder(json_object)
+        if opentracing_span:
+            opentracing_span.log_kv({"event": "encoded"})
+        return res
+
+    with start_active_span("encode_json_response"):
+        span = active_span()
```
clokep, line 776: We seem to be using this same logic in a bunch of places, it looks similar to the `trace` decorator, but not the same. It feels really duplicative, but maybe that's just open tracing?

(I guess in this case we need the span itself so we can call `log_kv` on it...)

synapse/rest/client/sync.py
```diff
@@ -239,29 +239,33 @@ async def encode_response(
...
             raise Exception("Unknown event format %s" % (filter.event_format,))
 
-        joined = await self.encode_joined(
-            sync_result.joined,
-            time_now,
-            access_token_id,
-            filter.event_fields,
-            event_formatter,
-        )
+        with start_active_span("sync.encode_joined"):
```
clokep, line 242: Could we use `trace` on each of these functions or do you prefer the context manager approach?

clokep (approved):
Overall this looks useful! Thanks for adding these! I left a few questions, but none of those are blockers to merging.

synapse/api/auth.py
richvdh: sorry, yes, I must have gone through that same thought process, but forgot to write it down here!

synapse/http/server.py
richvdh: well, it's less opentracing, and more our wrappers for opentracing (which make opentracing itself an optional dependency). If you have a real opentracing, you can do:

```
    with start_active_span("encode_json_response") as scope:
        span = scope.span
        span.log_kv(...)
```

... but with our wrappers, `scope` can be `None`, so you need the extra `active_span()` call.

synapse/rest/client/sync.py
richvdh: oh, good idea.

richvdh: done in fd7b6a0

synapse/api/auth.py
clokep: No problem! Definitely better to put all this logic in one place though!

synapse/http/server.py
clokep: Ah, OK! That makes a bit more sense!

PR #11621, opened 2021-12-21
Remove account data (including client config, push rules and ignored users) upon user deactivation.

### LINKED ISSUE #11539: Account data isn't deleted upon deactivation
We currently don't delete account data when a user is deactivated. This means we retain SSSS, lists of ignored users, etc, in the database. We probably want to delete them.

reivilibre:
Not sure how dodgy purging the stream tables is. Will need to ask to make sure.

- [x] tests

Fixes #11539.


dklimpel:
https://github.com/matrix-org/synapse/blob/221595414751f7b8fd0c79772c5ac4ffefeca10a/docs/admin_api/user_admin_api.md#L368

Is this part of `account_data`?

dklimpel:
https://github.com/matrix-org/synapse/issues/11539#issuecomment-988994841 asked also for a background update.

dklimpel:
PR for background task:
- #11655

synapse/storage/databases/main/account_data.py
```diff
@@ -546,6 +546,45 @@ def _add_account_data_for_user(
         for ignored_user_id in previously_ignored_users ^ currently_ignored_users:
             self._invalidate_cache_and_stream(txn, self.ignored_by, (ignored_user_id,))
 
+    async def purge_account_data_for_user(self, user_id: str) -> None:
+        """
+        Removes ALL the account data for a user.
+        Intended to be used upon user deactivation.
+
+        Also purges the user from the ignored_users cache table
+        and the push_rules cache tables.
```
clokep, line 555: Are these part of account data or are they unrelated? (This is confusing to me that they're called out separately.)

tests/handlers/test_deactivate_account.py
```diff
@@ -0,0 +1,78 @@
...
+
+class DeactivateAccountTestCase(HomeserverTestCase):
+    servlets = [
+        login.register_servlets,
+        admin.register_servlets,
+        account.register_servlets,
+    ]
+
+    def prepare(self, reactor: MemoryReactor, clock: Clock, hs: HomeServer) -> None:
+        super(DeactivateAccountTestCase, self).prepare(reactor, clock, hs)
```
clokep, line 33: No need to call `super()` here, this method is designed to be replaced.

tests/handlers/test_deactivate_account.py
```diff
@@ -0,0 +1,78 @@
...
+        self._store = hs.get_datastore()
+
+        self.user = self.register_user("user", "pass")
+        self.token = self.login("user", "pass")
+
+    def test_account_data_deleted_upon_deactivation(self) -> None:
+        """
+        Tests that account data is removed upon deactivation.
+        """
+        # Add some account data
```
clokep, line 43: Should we test ignoring a user and/or push rules?

tests/handlers/test_deactivate_account.py
```diff
@@ -0,0 +1,78 @@
...
+                },
+                access_token=self.token,
+            )
+        )
+        self.assertEqual(req.code, 200, req)
+
+        # Check that the account data does not persist.
+        self.assertIsNone(
+            self.get_success(
+                self._store.get_global_account_data_by_type_for_user(
```
clokep, line 73: Do we need to do anything with room specific account data?

synapse/storage/databases/main/account_data.py
reivilibre: They're caches of the information in the account data — much like we have `state_events` to cache state events even though the canonical information is in `event_json`.
Will clarify since I suppose it's non-obvious.

clokep: I wasn't really sure what "cache table" meant, I suppose. Would it make more sense to just list all the tables it clears? Or call them derived or something?

tests/handlers/test_deactivate_account.py
reivilibre: Terribly good idea... revealed that I wasn't actually doing anything to room account data. Thanks :)

synapse/storage/databases/main/account_data.py
```diff
@@ -546,6 +546,49 @@ def _add_account_data_for_user(
         for ignored_user_id in previously_ignored_users ^ currently_ignored_users:
             self._invalidate_cache_and_stream(txn, self.ignored_by, (ignored_user_id,))
 
+    async def purge_account_data_for_user(self, user_id: str) -> None:
+        """
+        Removes ALL the account data for a user.
+        Intended to be used upon user deactivation.
+
+        Also purges the user from the ignored_users table and the push_rules tables,
+        because those are derived from account data.
```
clokep, line 555:
```suggestion
        Removes the account data for a user.

        This is intended to be used upon user deactivation and also removes any
        derived information from account data (e.g. push rules and ignored users).

		Args:
			user_id: The user ID to remove data for.
```

synapse/storage/databases/main/account_data.py
```diff
@@ -546,6 +546,49 @@ def _add_account_data_for_user(
...
+        """
+
+        def purge_account_data_for_user_txn(txn: LoggingTransaction) -> None:
+            # Purge from the primary account_data table.
```
clokep, line 559:
```suggestion
            # Purge from the primary account_data tables.
```

tests/handlers/test_deactivate_account.py
```diff
@@ -0,0 +1,269 @@
...
+                    },
+                    "erase": True,
+                },
+                access_token=self.token,
+            )
+        )
+        self.assertEqual(req.code, 200, req)
+
+        # Clear the cache (for testing)
+        self._store.get_global_account_data_by_type_for_user.invalidate_all()
```
clokep, line 83: This deactivation seems like it would be good to put into a helper on this test case!

tests/handlers/test_deactivate_account.py
```diff
@@ -0,0 +1,269 @@
...
+                    },
+                    "erase": True,
+                },
+                access_token=self.token,
+            )
+        )
+        self.assertEqual(req.code, 200, req)
+
+        # Clear the cache (for testing)
+        self._store.get_account_data_for_room_and_type.invalidate_all()
```
clokep, line 139: Shouldn't we clear the caches properly too when deactivating the user? That would alleviate the need to do it just for tests.

clokep (changes requested):
Looks pretty good, my main concern is that we seem to not be clearing caches and then working around that in the tests, which feels wrong!

tests/handlers/test_deactivate_account.py
reivilibre: Yeah, was wondering about this, but also was wondering whether it might be expensive to do (presumably it'd have to be put into the stream to other workers).
I need to figure out how to clear caches properly (I want to clear for all of a single user ID at once and ignore the second arg? — that might alleviate the above concerns).

synapse/storage/databases/main/account_data.py
```diff
@@ -546,6 +547,74 @@ def _add_account_data_for_user(
...
+            )
+            self._invalidate_cache_and_stream(
+                txn, self.get_push_rules_for_user, (user_id,)
+            )
+            self._invalidate_cache_and_stream(
+                txn, self.get_push_rules_enabled_for_user, (user_id,)
+            )
+            # This user might be contained in the ignored_by cache for other users,
+            # so we have to invalidate it all.
+            self._invalidate_all_cache_and_stream(txn, self.ignored_by)
```
clokep, line 611: I'm not sure I agree -- the other users still have this user ignored and might e.g. pull stale data with messages from this user. Those cached entries are still valid (and we're not modifying them). The goal is to purge useless cached data from the user being deactivated.

reivilibre: `ignored_by` is a confusing name, I suspect that might be tripping you up.

A is in ignored_by(B) iff A ignores B.

Because we're deleting the fact that *A ignores B*, we have to invalidate `ignored_by(B)` for all users *B* that A ignores. (It's easier in practice to invalidate the whole cache, I think — or to leave it alone, but we've said we're trying not to be too lazy with this).

Note: during the deactivation of A, we only delete the 'A ignores B' facts but not 'B ignores A' facts — if there are still stale messages from A, users ignoring A will continue to do so.

Maybe `ignored_by` should be `get_ignorers_of`, but even that sounds a little bit clunky

clokep (approved):
Looks good!

tests/handlers/test_deactivate_account.py
```diff
@@ -0,0 +1,219 @@
...
+                self._store.get_account_data_for_room_and_type(
+                    self.user, room_id, "m.fully_read"
+                )
+            ),
+        )
+
+    def _is_custom_rule(self, push_rule: Dict[str, Any]) -> bool:
+        """
+        Default rules start with a dot: such as .m.rule and .im.vector.
+        This function returns true iff a rule is custom (not default).
```
deepbluev7, line 138: iff -> if

tests/handlers/test_deactivate_account.py
```diff
@@ -0,0 +1,219 @@
...
+        """
+        return "/." not in push_rule["rule_id"]
```
deepbluev7, line 140: Why didn't you check `push_rule["default"]`? While this check did find an error in my code, I'm a bit confused why it is written that way. :D

tests/handlers/test_deactivate_account.py
clokep: iff means "if and only if", it isn't a typo. (Although I'm not a big fan of including abbreviations like this.)

deepbluev7: Huh, thanks! TIL!

tests/handlers/test_deactivate_account.py
deepbluev7: (Was explained to me now, thanks a lot)

tests/handlers/test_deactivate_account.py
clokep: It comes up somewhat frequently which is why I usually try to spell it out! Sorry I missed that in review!

deepbluev7: No worries. I learned something new, which is always great. Thank you for taking your time to explain it to me, even when this is already closed! <3

PR #11623, opened 2021-12-21
Do not attempt to bundled aggregations for /members and /state.

clokep:
I made a minor mistake in #11592 where it requested to bundle aggregations when fetching the room state for `/members` and `/state` -- the bundled aggregations wouldn't actually get added ([since we don't include them for state events](https://github.com/matrix-org/synapse/blob/dd4778875213d9cb8be7ee71d32751fbd6cdaba2/synapse/events/utils.py#L430)), but it means that we're "opting-in" to something that makes no sense for the APIs.

### DIFF
--- a/changelog.d/11623.bugfix
+++ b/changelog.d/11623.bugfix
@@ -1,0 +1,1 @@
+Fix a long-standing bug where responses included bundled aggregations when they should not, per [MSC2675](https://github.com/matrix-org/matrix-doc/pull/2675).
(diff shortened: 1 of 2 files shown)


clokep:
I double and triple checked all places where we set `bundle_aggregations=True` now and I'm fairly certain they're correct! 🎉

reivilibre (approved):
LGTM

PR #11625, opened 2021-12-21
Pin to frozendict<2.1.2

richvdh:
... to work around breakage on buster (https://github.com/Marco-Sulla/python-frozendict/issues/41)

### DIFF
--- a/changelog.d/11625.misc
+++ b/changelog.d/11625.misc
@@ -1,0 +1,1 @@
+Work around a build problem on Debian Buster.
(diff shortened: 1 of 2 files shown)


clokep (approved):

richvdh:
(turns out that our newsfile linter compares against develop rather than the target of the PR, so it's complaining that I've removed a file, which I haven't)

K900:
The upstream issue is now fixed, and the pin breaks distros that ship a newer frozendict instead (Arch, NixOS, etc.). Maybe it's time to revert?

PR #11631, opened 2021-12-22
Elide the requested event in `/event_auth`

richvdh:
[MSC3585](https://github.com/matrix-org/matrix-doc/pull/3585) says this is a thing we can do.


synapse/handlers/federation.py
```diff
@@ -423,7 +423,7 @@ async def send_invite(self, target_host: str, event: EventBase) -> EventBase:
     async def on_event_auth(self, event_id: str) -> List[EventBase]:
         event = await self.store.get_event(event_id)
         auth = await self.store.get_auth_chain(
-            event.room_id, list(event.auth_event_ids()), include_given=True
+            event.room_id, list(event.auth_event_ids())
```
richvdh, line 426: this was totally misguided. Note that we are passing in `event_id`'s *`auth_events`* here, rather than event_id itself - which is why we want to `include_given`.

Arguably we could still simplify this by just passing in `event_id` rather than doing `get_event` on it first; but I'm going to close this for now.

PR #11632, opened 2021-12-22
`FederationClient.backfill`: stop flagging events as outliers

### LINKED ISSUE #8894: Backfilled events marked as outliers break backfill
A homeserver backfilling from another homeserver was found to be raising an `AssertionError` at:

https://github.com/matrix-org/synapse/blob/30fba6210834a4ecd91badf0c8f3eb278b72e746/synapse/handlers/federation.py#L921-L923

It's currently unclear which event is problematic and why it might be marked as an outlier. The logs of the backfilling server also contained many instances of event signature validation failures. This was due to a remote server changing their signing key without also changing the key ID (or putting the old key in `old_signing_keys`). This may be a red herring however.

The `assert` and a similar assert at:

https://github.com/matrix-org/synapse/blob/30fba6210834a4ecd91badf0c8f3eb278b72e746/synapse/handlers/federation.py#L891-L894

both do not give any information about the failing event.

Thus the first step is to improve these assert statements and figure out why these backfilled events are marked as outliers. After that, we may want to make the backfill code drop these events instead of simple raising, similar to what was done in #8350.

richvdh:
Events returned by `backfill` should not be flagged as outliers.

Fixes:

```
AssertionError: null
  File "synapse/handlers/federation.py", line 313, in try_backfill
    dom, room_id, limit=100, extremities=extremities
  File "synapse/handlers/federation_event.py", line 517, in backfill
    await self._process_pulled_events(dest, events, backfilled=True)
  File "synapse/handlers/federation_event.py", line 642, in _process_pulled_events
    await self._process_pulled_event(origin, ev, backfilled=backfilled)
  File "synapse/handlers/federation_event.py", line 669, in _process_pulled_event
    assert not event.internal_metadata.is_outlier()
```

See https://sentry.matrix.org/sentry/synapse-matrixorg/issues/231992

Fixes #8894.

### DIFF
--- a/synapse/federation/federation_client.py
+++ b/synapse/federation/federation_client.py
@@ -272,7 +272,7 @@ class FederationClient(FederationBase):
 
         # Check signatures and hash of pdus, removing any from the list that fail checks
         pdus[:] = await self._check_sigs_and_hash_and_fetch(
-            dest, pdus, outlier=True, room_version=room_version
+            dest, pdus, room_version=room_version
         )
 
         return pdus
--- a/changelog.d/11632.bugfix
+++ b/changelog.d/11632.bugfix
@@ -1,0 +1,1 @@
+Fix a bug introduced in Synapse 1.19.3 which could sometimes cause `AssertionError`s when backfilling rooms over federation.
--- a/synapse/handlers/federation_event.py
+++ b/synapse/handlers/federation_event.py
@@ -666,7 +666,9 @@ class FederationEventHandler:
         logger.info("Processing pulled event %s", event)
 
         # these should not be outliers.
-        assert not event.internal_metadata.is_outlier()
+        assert (
+            not event.internal_metadata.is_outlier()
+        ), "pulled event unexpectedly flagged as outlier"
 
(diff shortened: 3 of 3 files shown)


synapse/federation/federation_client.py
```diff
@@ -272,7 +272,7 @@ async def backfill(
 
         # Check signatures and hash of pdus, removing any from the list that fail checks
         pdus[:] = await self._check_sigs_and_hash_and_fetch(
-            dest, pdus, outlier=True, room_version=room_version
+            dest, pdus, room_version=room_version
```
richvdh, line 275: note that the events returned by `event_from_pdu_json` a couple of lines above are *not* flagged as outliers. the `outlier` flag here tells `_check_sigs_and_hash_and_fetch` that, if it pulls another copy of the event due to an incorrect signature, it should flag it as an outlier.

(I think the existence of an `outlier` param on `_check_sigs_and_hash_and_fetch` is misguided, but that's a separate PR.)

clokep (approved):
Makes sense that these are consistent.

PR #11633, opened 2021-12-22
Drop Bionic from Debian builds 

H-Shay:
Related to #11552, stops building bionic, updates the `dh-virtualenv` (which we can do now that we have dropped bionic) and updates Trove classifier to indicate that we have dropped support for EOL python 3.6.

### DIFF
--- a/changelog.d/11633.misc
+++ b/changelog.d/11633.misc
@@ -1,0 +1,2 @@
+Stop building bionic, update dh-virtualenv and update Trove classifiers, tasks all associated with dropping
+EOL Python3.6. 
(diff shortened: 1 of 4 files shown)


changelog.d/11633.misc
```diff
@@ -0,0 +1,2 @@
+Stop building bionic, update dh-virtualenv and update Trove classifiers, tasks all associated with dropping
+Python 3.6, which is EOL.
```
richvdh, line 2: I think these need to be on one line. In any case I'd probably just say "Drop support for Python 3.6 and Ubuntu 18.04."

richvdh (approved):
lgtm!

PR #11634, opened 2021-12-22
Refactor the way we set `outlier`

richvdh:
Currently, the federation code is rather inconsistent about where we set `outlier`. Most often it is set when we turn the pdu received over the wire into an `EventBase` (in `event_from_pdu_json`), but this is far from universal, and there are plenty of places where this doesn't happen (for example, for the auth/state event in the response to `/send_join`), and instead the flag is set later.

It doesn't really make much sense to set the flag when we're constructing the event, because at that point we don't really know if it's going to be an outlier or not. (In other words, being an outlier isn't a property of the shape of the event itself, but rather how we persist it.) Accordingly, it seems like it should be the job of `_auth_and_persist_outliers` to set this flag.

This PR is a series of bite-sized commits, each of which should hopefully be easy to understand and verify.

Based on #11632.


synapse/handlers/federation_event.py
```diff
@@ -1312,6 +1310,9 @@ def prep(event: EventBase) -> Optional[Tuple[EventBase, EventContext]]:
                         return None
                     auth.append(ae)
 
+                # we're not bothering about room state, so flag the event as an outlier.
+                event.internal_metadata.outlier = True
```
clokep, line 1314: Should the docstring of `_auth_and_persist_outliers` (or `_auth_and_persist_outliers_inner`) be updated to note that it will mark events as outliers?

clokep (approved):
Seems reasonable.

Issue #3024, opened 2018-03-25
Search for URLs in messages doesn't return any results


WGH-:
### Description

Message text search doesn't work with URL. Even if you search for URL exactly as it was in a message, search will not return any results.

### Steps to reproduce

- Post some link in some room, e.g. https://www.youtube.com/watch?v=oHg5SJYRHA0
- Enter the same link in the search box: https://www.youtube.com/watch?v=oHg5SJYRHA0
- Press enter

The search results will not include the message with the link, which is unexpected.

### Version information

- **Homeserver**: self-hosted
- **Version**:        matrix-synapse==0.26.0
- **Install method**: pip      
- **Platform**: Linux, PostgreSQL 9.6

WGH-:
Plain SQL query returns the message like a charm:
```
synapse=> select event_id, content from events where content like '%https://www.youtube.com/watch?v=oHg5SJYRHA0%';
            event_id            |                                  content                                  
--------------------------------+---------------------------------------------------------------------------
 $15219863901020QNrdM:torlan.ru | {"body":"https://www.youtube.com/watch?v=oHg5SJYRHA0","msgtype":"m.text"}
(1 row)

```

skylord123:
Just ran into this same issue. Tried searching a room for a specific link and it wasn't finding anything but I found it when searching manually.

werner291:
Related to https://github.com/vector-im/riot-web/issues/4486 ?

werner291:
This test case passes:
```
    def test_finds_message_with_url(self):
        """
        Same as test_finds_message, but now with a URL.
        See https://github.com/matrix-org/synapse/issues/3024
        """
        # The other user sends some messages
        self.helper.send(self.room, body="Hi!", tok=self.other_access_token)
        self.helper.send(self.room, body="https://www.youtube.com/watch?v=oHg5SJYRHA0", tok=self.other_access_token)
        self.helper.send(self.room, body="There!", tok=self.other_access_token)

        request, channel = self.make_request(
            "POST",
            "/search?access_token=%s" % (self.access_token,),
            {
                "search_categories": {
                    "room_events": {"search_term": "youtube.com", "order_by": "recent"}
                }
            },
        )

        self.render(request)

        # Check we get the results we expect -- one search result, of the sent
        # messages
        self.assertEqual(channel.code, 200)
        results = channel.json_body["search_categories"]["room_events"]
        self.assertEqual(results["count"], 1)
[...]

PR #11635, opened 2021-12-23
Unified search query syntax using the full-text search capabilities of the underlying DB

novocaine:
Support a unified search query syntax which leverages more of the full-text search of each DB we support.

We now support, with the same syntax across Postgresql and Sqlite:

- quoted "search terms"
- `AND`, `OR`, `-` (negation) set operators ~~and parentheses~~
- Matching words based on their stem, e.g. searches for "dog" matches documents containing "dogs". 

This is achieved by 

- If on pgsql 11+, pass the user inputer to `websearch_to_tsquery`
- If on an earlier version of pgsql or sqlite, manually parse the query and transform it into the underlying DB syntax supporting the same query dialect

Multiple terms separated by a space are implicitly ANDed.

Possible Issues:

1. There is no escaping of full-text syntax that might be supported by the DB; e.g. `NOT`, `NEAR`, `*` in sqlite, `&`, `|` in pgsql. This runs the risk that people might discover this as accidental functionality and depend on something we don't guarantee.
2. On both pgsql and sqllite, english text is assumed for stemming. To support other languages, it's work: either the target language needs to be known at the time of indexing the message (via room metadata, or otherwise), or it'll be necessary to create a separate index for each language supported. Alternatively, we could just disable stemming for now (but still use the other improvements in this PR).

## Before

Prior to this PR, we transformed all full-text queries into a DB full-text query by 

1. splitting the query on all non-word characters  (i.e. not in the python regex class `\w`) 
3. suffixing all the words with "*" to use prefix matching
4. re-joining the words using '&' as the separator.
 
For example, the user input `foo? bar` is transformed into `foo* & bar*`. This happens for both sqlite and postgres.

This had some drawbacks - 

* On sqlite, & isn't actually an operator, its actually just being stripped and then the remaining spaces are interpreted to mean AND. 
* Stripping the query of non-word characters means quote search and NOT aren't available to users, despite both engines supporting these operators. Note that client-side seshat search does support these operators as its based on on [tantivy](https://github.com/quickwit-inc/tantivy)
* Prefix search probably isn't what users expect as that's not how search engines usually work; it probably interferes with the expected stemming behaviour; and it might have significant performance impact
 
[Sqlite docs](https://www.sqlite.org/fts3.html#full_text_index_queries)
[Postgres docs](https://www.postgresql.org/docs/11/textsearch-controls.html)

And for comparison with how Element clients support search, see the [Tantivy docs](https://docs.rs/tantivy/latest/tantivy/query/struct.QueryParser.html)

### Pull Request Checklist

### Sign-off

Signed-off-by: James Salter <iteration@gmail.com>


synapse/storage/databases/main/search.py
```diff
@@ -437,7 +440,7 @@ async def search_msgs(self, room_ids, search_term, keys):
                 "SELECT room_id, count(*) as count FROM event_search"
                 " WHERE value MATCH ?"
             )
-            count_args = [search_term] + count_args
```
novocaine, line 440: I think it was a bug to pass `search_term` here rather than `search_query`

synapse/storage/databases/main/search.py
```diff
@@ -727,20 +738,25 @@ def _to_postgres_options(options_dict):
     return "'%s'" % (",".join("%s=%s" % (k, v) for k, v in options_dict.items()),)
 
 
-def _parse_query(database_engine, search_term):
+def _parse_query_for_sqlite(search_term: str) -> str:
     """Takes a plain unicode string from the user and converts it into a form
-    that can be passed to database.
-    We use this so that we can add prefix matching, which isn't something
-    that is supported by default.
+    that can be passed to sqllite's matchinfo().
```
novocaine, line 743: This is currently a no-op, but we will probably want to add stuff here to handle certain cases e.g. https://github.com/matrix-org/synapse/issues/3024

novocaine:
There's a test failing on sqlite because it relies on some sort of prefixing (searches for `labels` to match `label`); this test is passing on postgres because there we're using the database's stemming. 

sqlite does actually [support stemming](https://www.sqlite.org/fts3.html#tokenizer), so I'm going to try to enable that to fix (this relies on a db migration, so might get tricky)

synapse/storage/schema/main/delta/25/fts.py
```diff
@@ -37,7 +37,7 @@
 
 SQLITE_TABLE = (
     "CREATE VIRTUAL TABLE event_search"
-    " USING fts4 ( event_id, room_id, sender, key, value )"
+    " USING fts4 (tokenize=porter, event_id, room_id, sender, key, value )"
```
novocaine, line 40: I think this is desirable to avoid dropping and recreating this table later, but it might mean we need to update `full_schemas` which seems to be a cached copy of the materialized schema

novocaine: the problem is that if we do not update it, during creation of a new db will involve creating `event_search` and then dropping it and creating it again with `tokenize=porter`

reivilibre: I don't think it makes sense to edit an old delta, as it means that there won't be consistency on what version 25 means regardless of when you installed Synapse (and note that delta 25 won't be run by new homeservers because new homeservers will start with the full snapshot at version 54).
The full schemas are point-in-time snapshots (not manually created), which would be equivalent to running all the deltas leading up to it.
We don't edit them once they've been released 'into the wild'.

Instead, the table should be dropped+recreated or altered in a new delta — yes it's slightly annoying for a new homeserver to have to do this 'create, drop, create' dance, but it takes negligible time. At a later date, a new full schema snapshot can be created so that new homeservers don't have to run all the historical deltas.
As for the background job: The old background job can be deleted and then recreated — this way you don't get two jobs queued up to do the same task.

richvdh: this delta (v25) won't get used by anyone creating a new db today, because we'll start from the full schema in `synapse/storage/schema/main/full_schemas/54`, and only apply the deltas from 55 onwards.

The only time delta 25 is used is if someone has an incredibly old installation (pre-#326) and attempts to upgrade that directly to a modern synapse version. It'd be *nice* if it worked, but there's certainly no need to make it efficient, and realistically we're not going to test upgrades from Synapse versions that old.

synapse/storage/databases/main/search.py
```diff
@@ -770,20 +790,168 @@ def _to_postgres_options(options_dict: JsonDict) -> str:
...
+                phrase_word = words[i]
+            tokens.append(Phrase(phrase))
+        elif word[0] == "-":
+            tokens.append(Not())
+            tokens.append(word[1:])
+        elif word.lower() == "or":
+            tokens.append(Or())
+        elif word.lower() == "and":
+            tokens.append(And())
+        elif word == "(":
```
erikjohnston, line 861: Should this not check just the start of the word? So that it matches `(foo AND bar)`?

synapse/storage/schema/main/delta/65/12_update_sqlite_fts4_tokenizer.py
```diff
@@ -0,0 +1,59 @@
+# Copyright 2021 The Matrix.org Foundation C.I.C.
```
erikjohnston, line 1: We'll need to update the branch and move this into `delta/70` folder

clokep: It will need to go into `delta/71` since the schema version has been bumped.

sliedes:
@richvdh asked in #12609 if this change would fix it (searching for a literal text "C++20"). I assume this would, but let me confirm my understanding and ask some hopefully useful questions on this change, not all related to searching for that literal. I looked through the changes, but I'm completely unfamiliar with the code base of Synapse. So, please forgive me if some of my comments are trivial or less than useful.

So, do I understand correctly that:

- using "C++20" in quotes would match that exact text in a comment (and I assume links, but not any preview text or such? Though I think either way I would consider this to fix #12609), case insensitive
- words inside the literals will not be stemmed

Also, I think if the sqlite "*" allows doing a prefix search, I bet it will 100% be discovered and also quite useful, though unintentional.

synapse/storage/databases/main/search.py
clokep: I think this is getting lucky that the syntax is essentially the same for websearch vs. SQL search here, so we fall-through to the "last" case (which just appends `"(foo"`, then when we convert back to SQL syntax append that token back out so you get `(foo | bar)` instead of `( foo | bar )`.

This will breakdown though if you do something like `("f o o" and bar)` though I *think*?

synapse/storage/databases/main/search.py
```diff
@@ -770,20 +790,168 @@ def _to_postgres_options(options_dict: JsonDict) -> str:
...
+
+class RParen:
+    pass
+
+
+Token = Union[str, Phrase, Not, Or, And, LParen, RParen]
+TokenList = List[Token]
+
+
+def _tokenize_query(query: str) -> TokenList:
```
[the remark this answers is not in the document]
clokep: I tested some of the corner cases, there's still a couple having to do with quotes though.

synapse/storage/databases/main/search.py
```diff
@@ -760,20 +781,193 @@ def _to_postgres_options(options_dict: JsonDict) -> str:
...
+class Phrase:
+    phrase: List[str]
+
+
+class SearchToken(enum.Enum):
+    Not = enum.auto()
+    Or = enum.auto()
+    And = enum.auto()
+    LParen = enum.auto()
+    RParen = enum.auto()
```
clokep, line 794: > `AND`, `OR`, `-` (negation) set operators and parentheses
>
> If on pgsql 11+, pass the user inputer to `websearch_to_tsquery`

It seems that `websearch_to_tsquery` doesn't actually support parentheses in query text (they get removed); it also doesn't seem to recognize `AND` as an operator:

```sql
$ SELECT websearch_to_tsquery('simple', 'fox AND (nope OR brown)');
       websearch_to_tsquery       
----------------------------------
 'fox' & 'and' & 'nope' | 'brown'
```

The underlying `to_tsquery` does support them though.

Note that (ignoring the bogus `AND`), this isn't correct due to operator precedence: `A & (B | C) == A & B | A & C`, where we have `A & B | C`.

synapse/storage/databases/main/search.py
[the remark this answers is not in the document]
clokep: This *doesn't* handle quotes particularly well at the moment, but I think we need to figure out whether we care about parentheses first: https://github.com/matrix-org/synapse/pull/11635#discussion_r999714347

clokep:
@erikjohnston any thoughts on the currently open comment threads?

synapse/storage/databases/main/search.py
erikjohnston: I don't think the complexity in supporting parentheses is worth it? It feels quite niche for someone to need it and for them to know about it, but :shrug:

synapse/storage/databases/main/search.py
```diff
@@ -444,20 +454,25 @@ async def search_msgs(
...
+            search_query, tsquery_func = _parse_query_for_pgsql(
+                search_term, self.database_engine
+            )
             sql = (
-                "SELECT ts_rank_cd(vector, to_tsquery('english', ?)) AS rank,"
+                f"SELECT ts_rank_cd(vector, {tsquery_func}('english', ?)) AS rank,"
                 " room_id, event_id"
                 " FROM event_search"
-                " WHERE vector @@ to_tsquery('english', ?)"
+                f" WHERE vector @@  {tsquery_func}('english', ?)"
```
erikjohnston, line 464: Can we move these to multiline strings while we're here :innocent:

synapse/storage/databases/main/search.py
```diff
@@ -760,20 +781,204 @@ def _to_postgres_options(options_dict: JsonDict) -> str:
...
+
+    Args:
+        search_term: A user supplied search query.
+        engine: The database engine.
+
+    Returns:
+        A tuple of (parsed search_term, tsquery func to use).
+    """
+
+    if engine.supports_websearch_to_tsquery:
```
erikjohnston, line 974: How awful would it be to not do all the fancy handling of and vs or etc for old postgres? Especially given PG10 is EoL in a fortnight...

erikjohnston (commented):
Otherwise I think its good

synapse/storage/databases/main/search.py
clokep: I was planning to do a follow-up PR to update the entire module to multi-line strings. Would that be acceptable?

synapse/storage/databases/main/search.py
clokep: I think we could remove a bunch of code. We could just fallback to `plain_to_tsquery` or something on old postgres, which won't work great but shouldn't break anything.

erikjohnston (approved):

PR #11642, opened 2021-12-23
Fix AssertionErrors after purging events

richvdh:
If you purged a bunch of events from your database, and then restarted synapse
without receiving more events, then you would get a bunch of AssertionErrors on
restart.

This fixes the situation by rewinding the stream processors.

(I've also updated the changelog for #11536, since this was the cause of the mixup I was trying to debug there.)

### DIFF
--- a/scripts-dev/check-newsfragment
+++ b/scripts-dev/check-newsfragment
@@ -42,8 +42,8 @@ echo "--------------------------"
 echo
 
 matched=0
-for f in $(git diff --name-only FETCH_HEAD... -- changelog.d); do
-    # check that any modified newsfiles on this branch end with a full stop.
+for f in $(git diff --diff-filter=d --name-only FETCH_HEAD... -- changelog.d); do
+    # check that any added newsfiles on this branch end with a full stop.
     lastchar=$(tr -d '\n' < "$f" | tail -c 1)
     if [ "$lastchar" != '.' ] && [ "$lastchar" != '!' ]; then
         echo -e "\e[31mERROR: newsfragment $f does not end with a '.' or '!'\e[39m" >&2
--- a/synapse/handlers/stats.py
+++ b/synapse/handlers/stats.py
@@ -80,6 +80,17 @@ class StatsHandler:
         # If self.pos is None then means we haven't fetched it from DB
         if self.pos is None:
             self.pos = await self.store.get_stats_positions()
+            room_max_stream_ordering = self.store.get_room_max_stream_ordering()
+            if self.pos > room_max_stream_ordering:
+                # apparently, we've processed more events than exist in the database!
+                # this can happen if events are removed with history purge or similar.
+                logger.warning(
+                    "Event stream ordering appears to have gone backwards (%i -> %i): "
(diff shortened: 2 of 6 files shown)


scripts-dev/check-newsfragment
```diff
@@ -42,8 +42,8 @@ echo "--------------------------"
 echo
 
 matched=0
-for f in $(git diff --name-only FETCH_HEAD... -- changelog.d); do
-    # check that any modified newsfiles on this branch end with a full stop.
+for f in $(git diff --diff-filter=d --name-only FETCH_HEAD... -- changelog.d); do
```
richvdh, line 45: `--diff-filter=d` tells git to ignore deleted files here. This should probably be a separate PR but at this stage I've got about 15 of the things open :-S

scripts-dev/check-newsfragment
```diff
@@ -42,8 +42,8 @@ echo "--------------------------"
...
+    # check that any added newsfiles on this branch end with a full stop.
```
clokep, line 46: So this is added or modified, but not deleted?

synapse/handlers/stats.py
```diff
@@ -80,6 +80,17 @@ async def _unsafe_process(self) -> None:
...
+            if self.pos > room_max_stream_ordering:
+                # apparently, we've processed more events than exist in the database!
+                # this can happen if events are removed with history purge or similar.
+                logger.warning(
+                    "Event stream ordering appears to have gone backwards (%i -> %i): "
+                    "rewinding stats processor",
+                    self.pos,
+                    room_max_stream_ordering,
+                )
+                self.pos = room_max_stream_ordering
```
clokep, line 93: I'm having trouble convincing myself this is a "safe" operation to do. I thought the max stream ordering was monotonically increasing, even if you purged events but I might be misunderstanding how it is used?

scripts-dev/check-newsfragment
richvdh: yes.

synapse/handlers/stats.py
richvdh: well, that's a fair question.

The fact is that currently, on sqlite, we *do* reuse stream orderings, because we have no way of knowing which stream orderings were used before synapse was restarted. 

For things like the stats and userdirectory, that's not really a problem, because any events that get a stream ordering later than the max at startup must be new events, so I think this must be safe.

One place it *is* a problem is in client /sync - since stream_orderings are encoded in sync tokens, any events that reuse old stream_orderings probably won't get synced to the client. Ironically, we *don't* log any warnings in that case.

So, sure, a better solution here is not to reuse stream orderings at all, but that's not easy. Meanwhile I think we may as well update these bits to stop them complaining.

scripts-dev/check-newsfragment
richvdh: see https://git-scm.com/docs/git-diff#Documentation/git-diff.txt---diff-filterACDMRTUXB82308203 for docs on this btw.

clokep (approved):

PR #11654, opened 2021-12-29
Add type hints to `synapse/storage/databases/main/events_bg_updates.py`

dklimpel:
### Pull Request Checklist

### DIFF
--- a/synapse/storage/databases/main/events_bg_updates.py
+++ b/synapse/storage/databases/main/events_bg_updates.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 import logging
-from typing import TYPE_CHECKING, Dict, List, Optional, Tuple
+from typing import TYPE_CHECKING, Dict, List, Optional, Sequence, Set, Tuple, cast
 
 import attr
 
@@ -240,12 +240,14 @@ class EventsBackgroundUpdatesStore(SQLBaseStore):
 
         ################################################################################
 
-    async def _background_reindex_fields_sender(self, progress, batch_size):
+    async def _background_reindex_fields_sender(
+        self, progress: JsonDict, batch_size: int
+    ) -> int:
         target_min_stream_id = progress["target_min_stream_id_inclusive"]
         max_stream_id = progress["max_stream_id_exclusive"]
         rows_inserted = progress.get("rows_inserted", 0)
 
-        def reindex_txn(txn):
+        def reindex_txn(txn: LoggingTransaction) -> int:
             sql = (
                 "SELECT stream_ordering, event_id, json FROM events"
                 " INNER JOIN event_json USING (event_id)"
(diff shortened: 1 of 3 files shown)


synapse/storage/databases/main/events_bg_updates.py
```diff
@@ -1023,10 +1034,10 @@ def _calculate_chain_cover_txn(
         PersistEventsStore._add_chain_cover_index(
             txn,
             self.db_pool,
-            self.event_chain_id_gen,
+            self.event_chain_id_gen,  # type: ignore[attr-defined]
             event_to_room_id,
             event_to_types,
-            event_to_auth_chain,
+            cast(Dict[str, Sequence[str]], event_to_auth_chain),
```
dklimpel, line 1040: If not cast here:

```sh
synapse/storage/databases/main/events_bg_updates.py:1040: error: Argument 6 to "_add_chain_cover_index" of "PersistEventsStore" has incompatible type "Dict[str, List[str]]"; expected "Dict[str, Sequence[str]]"  [arg-type]
synapse/storage/databases/main/events_bg_updates.py:1040: note: "Dict" is invariant -- see https://mypy.readthedocs.io/en/stable/common_issues.html#variance
synapse/storage/databases/main/events_bg_updates.py:1040: note: Consider using "Mapping" instead, which is covariant in the value type
```

Is there a better solution?

clokep: I'm unsure if there's a better solution, this seems like it might be fine? In this method the sequence is mutable, while in `_add_chain_cover_index` it doesn't ever need it to be mutable so we're telling mypy it can treat the second argument as non-mutable.

clokep:
I think this looks fine, but if anyone has thoughts on the single open thread that would be appreciated!

synapse/storage/databases/main/events_bg_updates.py
reivilibre: This part of Python's type system is always a pain.
There's no nice solution here, as much as I always try to see one.
To get around it without having to cheat the system, you'd have to be able to 'consume' a `Dict[str, List[str]]` and turn it into `Dict[str, Sequence[str]]` such that you couldn't ever re-use the `Dict[str, List[str]]` (since it might contain e.g. tuples).

You could probably do that by cloning it, since the clone and the old one wouldn't be the same thing, but that seems like a waste of time/memory when the dict has already been built.

Realistically, you're probably better off leaving this `cast` in, as much as that makes me sad. With any luck there are some tests that check this part of the code that will ensure the types are right... :-)

reivilibre (approved):
LGTM as is — only two points where I wish things could be better are largely a downfall of Python and the current codebase, not this PR, so happy for it to be as-is.

synapse/storage/databases/main/events_bg_updates.py
```diff
@@ -1181,14 +1192,14 @@ def _event_arbitrary_relations_txn(txn: LoggingTransaction) -> int:
...
                     )
-                    self._invalidate_cache_and_stream(
-                        txn, self.get_aggregation_groups_for_event, cache_tuple
+                    self._invalidate_cache_and_stream(  # type: ignore[attr-defined]
+                        txn, self.get_aggregation_groups_for_event, cache_tuple  # type: ignore[attr-defined]
                     )
-                    self._invalidate_cache_and_stream(
-                        txn, self.get_thread_summary, cache_tuple
+                    self._invalidate_cache_and_stream(  # type: ignore[attr-defined]
+                        txn, self.get_thread_summary, cache_tuple  # type: ignore[attr-defined]
```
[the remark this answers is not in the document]
clokep: I've been doing the same in my code, I figure the stores inheritance stuff is a separate problem to solve.

PR #11655, opened 2021-12-29
Add a background database update to purge account data for deactivated users.

### LINKED ISSUE #11539: Account data isn't deleted upon deactivation

reivilibre:
- [x] tests
- [x] check that the schema delta version number doesn't need to be incremented

Follows up on #11621.

Fixes #11539.


richvdh:
NB this is a background *database update* rather than a "background job" (which I would expect to refer to things that we start with `run_as_background_process`)

synapse/storage/schema/main/delta/68/02_delete_account_data_for_deactivated_accounts.sql
```diff
@@ -0,0 +1,20 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+
+-- We want to retroactively delete account data for users that were already
+-- deactivated.
+INSERT INTO background_updates (ordering, update_name, progress_json) VALUES
+  (6520, 'delete_account_data_for_deactivated_users', '{}');
```
squahtx, line 20:
```suggestion
  (6802, 'delete_account_data_for_deactivated_users', '{}');
```
Should this be 6802, to match the schema version and file name?

tests/handlers/test_deactivate_account.py
```diff
@@ -217,3 +217,111 @@ def test_ignored_users_deleted_upon_deactivation(self) -> None:
...
+                    AccountDataTypes.DIRECT,
+                )
+            ),
+        )
+
+        # Re-run the retroactive deletion update
+        self._rerun_retroactive_account_data_deletion_update()
+
+        # Check that the account data was cleared.
+        self._store.get_global_account_data_by_type_for_user.invalidate_all()
```
squahtx, line 275: Not having the caches be cleared automatically looks like a footgun. I think this has already been discussed in the [previous PR](https://github.com/matrix-org/synapse/pull/11621#pullrequestreview-855636496), so I won't block.

reivilibre: Agh good spot — this line shouldn't be needed anymore, but I forgot about that. The curse of having too much stuff going on, but at least the reviewers are competent :-).

synapse/storage/schema/main/delta/68/02_delete_account_data_for_deactivated_accounts.sql
reivilibre: Good spot, forgot to update that, thanks!
I've also bumped this to `03` since someone took `02`, just for the sake of it.

PR #11657, opened 2021-12-29
Fix a type annotation in `test_account_data.py` and remove it from the Mypy exclusion list.

reivilibre:
Just a tiny one that I spotted when thinking about editing this file (though I later didn't anyway...).

### DIFF
--- a/changelog.d/11657.misc
+++ b/changelog.d/11657.misc
@@ -1,0 +1,1 @@
+Fix a type annotation in `test_account_data.py` and remove it from the Mypy exclusion list.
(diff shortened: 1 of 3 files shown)


changelog.d/11657.misc
```diff
@@ -0,0 +1 @@
+Fix a type annotation in `test_account_data.py` and remove it from the Mypy exclusion list.
```
clokep, line 1: To align it with a bunch of other changes for the next release?

```suggestion
Add missing type hints to storage classes.
```

clokep (approved):

changelog.d/11657.misc
clokep: I guess this is actually in a test, but...close enough?

PR #11593, opened 2021-12-16
Fix get federation status of destination if no error occured

dklimpel:
At the moment you get a `NotFoundError` if you request `GET /_synapse/admin/v1/federation/destinations/<destination>` and the destination has no problems.

The reason is that `get_destination_retry_timings` returns `None` if there is no problem with the destination.

This PR fix it:
- add DB function to query that a destination exists
- update rest API
- add a test

### Pull Request Checklist

* [x] Pull request is based on the develop branch
* [x] Pull request includes a [changelog file](https://matrix-org.github.io/synapse/latest/development/contributing_guide.html#changelog). The entry should:
  - Be a short description of your change which makes sense to users. "Fixed a bug that prevented receiving messages from other servers." instead of "Moved X method from `EventStore` to `EventWorkerStore`.".
  - Use markdown where necessary, mostly for `code blocks`.
  - End with either a period (.) or an exclamation mark (!).
  - Start with a capital letter.
* [x] Pull request includes a [sign off](https://matrix-org.github.io/synapse/latest/development/contributing_guide.html#sign-off)
* [x] [Code style](https://matrix-org.github.io/synapse/latest/code_style.html) is correct
  (run the [linters](https://matrix-org.github.io/synapse/latest/development/contributing_guide.html#run-the-linters))

Signed-off-by: Dirk Klimpel dirk@klimpel.org

PR #11658, opened 2021-12-29
Add admin API to get a list of federated rooms

dklimpel:
Based on #11593

Add admin API
```
GET /_synapse/admin/v1/federation/destinations/<destination>/rooms
```

### Pull Request Checklist


synapse/storage/databases/main/transactions.py
```diff
@@ -561,6 +561,54 @@ def get_destinations_paginate_txn(
...
+            count = cast(Tuple[int], txn.fetchone())[0]
+
+            rooms = self.db_pool.simple_select_list_paginate_txn(
+                txn=txn,
+                table="destination_rooms",
+                orderby="room_id",
+                start=start,
+                limit=limit,
+                retcols=("room_id", "stream_ordering"),
+                order_direction=order,
```
DMRobertson, line 604: I think this is going to do an OFFSET `start` LIMIT `limit`. Naively this will have to scan for `start +  limit` results and discard the `start`. 

That said, we're ordering by `room_id` with a fixed destination, and there's a btree index on `(destination, room_id)`. So maybe this is able to jump to the correct offset without having to scan `start` rows? Unsure.

Either way, it's probably fine as it is. If it's slow in practice, we can take another look.

synapse/rest/admin/federation.py
```diff
@@ -145,3 +145,59 @@ async def on_GET(
...
+    PATTERNS = admin_patterns("/federation/destinations/(?P<destination>[^/]*)/rooms$")
+
+    def __init__(self, hs: "HomeServer"):
+        self._auth = hs.get_auth()
+        self._store = hs.get_datastore()
+
+    async def on_GET(
+        self, request: SynapseRequest, destination: str
+    ) -> Tuple[int, JsonDict]:
+        await assert_requester_is_admin(self._auth, request)
```
DMRobertson, line 172: Aside: do we have to write this line everywhere in the admin API? I wonder if some kind of parent `AdminRestServlet` class could pull out this check for us.

DMRobertson (changes requested):
Looks broadly good. A few suggestions and thoughts in passing:

synapse/rest/admin/federation.py
DMRobertson: FWIW I wasn't worried about the duplication; was more worried that we might forget this check.

But yeah, probably fine as is. We have bigger fires.

changelog.d/11658.feature
```diff
@@ -0,0 +1 @@
+Add admin API to get a list of federated rooms.
```
DMRobertson, line 1:
```suggestion
Add an admin API to get a list of rooms that federate with a given remote homeserver.
```

DMRobertson (approved):
Great! Thank you!

DMRobertson (changes requested):
Oh, that's odd---merge conflicts with develop, I think from #11593? Could you resolve those, please?

DMRobertson (approved):

Issue #11644, opened 2021-12-24
Pull in type hints for pynacl #478


DMRobertson:
In my spare time (i.e. not paid Element time) I've been working to add type annotations to pynacl, see pyca/pynacl#660. This is almost done, and the maintainer is keen to include them in a new minor release, "in the next few days/weeks".

Once there's a release, we should remove pynacl from the list of imports we ignore in Sydent's mypy config. Then we should fixup any complaints. The same also holds true for signedjson and sydent (see https://github.com/matrix-org/sydent/issues/478).

PR #11659, opened 2021-12-29
Include bundled aggregations in the sync response cache.

### LINKED ISSUE #11476: /sync is not bundling aggregations for limited rooms

### LINKED ISSUE #11585: Avoid re-calculating bundled aggregations for cached results
Related to #11583 -- currently when we generate a sync response we cache it for a period of time (as clients sometimes timeout and retry). Unfortunately the bundled aggregations are added during the serialization step, which occurs after caching, and is re-done even when using a cached response.

Ideally we should be calculating the bundled aggregations earlier in a different manner and the serialization methods should be synchronous.

See https://matrix.to/#/!XaqDhxuTIlvldquJaV:matrix.org/$1E_K64Dd5p6iFMS4agak62bn5JZl3Ry-6XiTXYzOmhU?via=matrix.org&via=vector.modular.im&via=envs.net for a brief conversation around this.

clokep:
This has two benefits:

* Consistently return the same data for a cached response.
* Reduces the load of subsequent requests when using cached data.

There's more improvements we could do here (like caching the serialized bytes, but that's annoying because of `age`), but I'd prefer to keep this PR small and simple.

As mentioned in #11612:

> Comparing this to the steps in #11611, I think that this would help be moving the calculation of bundled aggregations from step 2 to step 1, thus ending up with them cached. This would still take over 800 seconds the first time, but should now be a no-op once they're cached (instead of taking 224 seconds).

Part of #11611. Reverts #11583. Fixes #11585. Fixes #11476.

~~Based on #11612~~


clokep:
I've double checked and other spots where we fetch bundled aggregations will not have the same issue of being on the wrong side of a response cache (none of the other instances are cached as far as I can tell).

clokep:
Note that this *does* re-enable bundling of aggregations for `/sync` which was disabled in #11583, we'll want to verify that performance seems OK after doing this.

changelog.d/11659.bugfix
```diff
@@ -0,0 +1 @@
+Include the bundled aggregations in the `/sync` response.
```
DMRobertson, line 1: Is this part of an MSC by any chance? If so maybe we should link it here?

synapse/handlers/sync.py
```diff
@@ -630,10 +633,17 @@ async def _load_filtered_recents(
 
             prev_batch_token = now_token.copy_and_replace("room_key", room_key)
 
+        # Don't bother to bundle aggregations if the timeline is unlimited,
+        # as clients will have all the necessary information.
```
DMRobertson, line 637: Is the intent here to let e.g. let clients show accurate reaction counts, even if some of the reaction events were skipped over in a gappy sync?

synapse/handlers/sync.py
```diff
@@ -630,10 +633,17 @@ async def _load_filtered_recents(
...
+        bundled_aggregations = None
+        if limited:
```
DMRobertson, line 639: On +645 we effectively replace `limited` with `limited or newly_joined_room`. I'm not sure why; but might it make sense to do the same here?

DMRobertson (commented):
Code looks reasonable, but I we might want to test it on `matrix-org-hotfixes` to get a sense of the performance impact before we merge?

changelog.d/11659.bugfix
clokep: It is part of MSC2675, yes.

See https://github.com/uhoreg/matrix-doc/blob/aggregations-helpers/proposals/2675-aggregations-server.md#bundling

synapse/handlers/sync.py
clokep: Yes, pretty much.

synapse/handlers/sync.py
clokep: I don't think it does, but I'll double check.

clokep: The previous code used what gets set as `limited` on `TimelineBatch`, so I think you're right here! 👍

synapse/handlers/sync.py
clokep: (I should note that this comment was copied from `synapse/rest/client/sync.py`, but if we can make it clearer, please let me know!)

clokep:
Failing build is #11644.

changelog.d/11659.bugfix
clokep:
```suggestion
Include the bundled aggregations in the `/sync` response, per [MSC2675](https://github.com/matrix-org/matrix-doc/pull/2675).
```

clokep:
> Code looks reasonable, but I think we might want to test it on `matrix-org-hotfixes` to get a sense of the performance impact before we merge?

Maybe, let's discuss this in a meeting at some point?

DMRobertson:
> > Code looks reasonable, but I think we might want to test it on `matrix-org-hotfixes` to get a sense of the performance impact before we merge?
> 
> Maybe, let's discuss this in a meeting at some point?

No strong opinions here! Perhaps it's just easier to merge as normal and ensure that testing on m.org explicitly includes a step to measure the performance impact?

clokep:
> No strong opinions here! Perhaps it's just easier to merge as normal and ensure that testing on m.org explicitly includes a step to measure the performance impact?

I think we usually just merge it and then check when we do a deploy next. 🤷

synapse/handlers/sync.py
```diff
@@ -636,7 +636,7 @@ async def _load_filtered_recents(
         # Don't bother to bundle aggregations if the timeline is unlimited,
         # as clients will have all the necessary information.
         bundled_aggregations = None
-        if limited:
```
DMRobertson: I wonder if these are equivalent actually, see https://github.com/matrix-org/synapse/pull/11659/commits/acdc04a897a04f20033396652016d563cbf273a3#diff-70c84622edff898486bdc02d7a579db4e52a3e0b26c14020bed0ca1fb8ea2018L519-L524

DMRobertson: TL;DR:
- sync code is confusing
- I think this is fine as it stands

DMRobertson (approved):
LGTM assuming CI happy!

synapse/handlers/sync.py
clokep: Yes, it gets flipped a bit. Would it be clearer to set it again once to `limited or newly_joined_room` and use that twice?

DMRobertson: IMO leave it as is, in favour of another sync code tidy up to clarify what `limited` means and all the various bits of flipping that go on.

clokep:
> > No strong opinions here! Perhaps it's just easier to merge as normal and ensure that testing on m.org explicitly includes a step to measure the performance impact?
> 
> I think we usually just merge it and then check when we do a deploy next. 🤷

Actually I can try using the performance testing thing we have with large accounts, that would be useful to do! 👍

synapse/handlers/sync.py
clokep: Makes sense to me. Thanks!

clokep:
> > > No strong opinions here! Perhaps it's just easier to merge as normal and ensure that testing on m.org explicitly includes a step to measure the performance impact?
> > 
> > 
> > I think we usually just merge it and then check when we do a deploy next. 🤷
> 
> Actually I can try using the performance testing thing we have with large accounts, that would be useful to do! 👍

In order to answer this I wanted to compare the performance of an initial `/sync` across different Synapse versions using a large anonymized account in postgres. I hooked this up to a Synapse running locally after enabling the response cache and did the following for each version:

* Wait for any background updates to finish
* Made a sync request to warm up database caches
* Repeating several times:
  * Wait enough time for the response cache to clear.
  * Make a sync request and record the elapsed time from the logs.
  * Make a second sync request (which would use the response cache) and record the elapsed time from the logs.

For comparison purposes I was interested in testing:
* v1.49.0
* v1.49.1
* `develop` (d41c4654db03dc76812e98222feb62668a1bf94c)
* `develop` (d41c4654db03dc76812e98222feb62668a1bf94c) patched to re-enable bundled aggregations
* #11659
* #11659 + #11660
* #11659 + #11660 + additional changes for the thread summary bundled aggregation

The timings were fairly consistent over 5 runs with the median and mean close to matching. I don't think the raw values are as interesting (as my computer is probably not representative of where Synapse would be running), but the ratio due to the changes should be useful. I chose to normalize the current `develop` since that seemed most appropriate. A ratio of 1.00 means no change, lower is better.

| | Uncached | Cached|
|---|---|---|
| v1.49.0 | 1.89 | 52.89 |
| v1.49.1 | 1.05 | 2.64 |
| `develop` | -- | -- |
| `develop` with agg | 3.21 | 134.97 |
| #11659 | 1.61 | 1.01 |
| #11659 + #11660 | 1.37 | 0.98 |
| #11659 + #11660 + threads | 1.42 | 1.02 |

My takeaways from the above:

* This PR is about 1.6x slower than `develop` for the uncached response, but we can improve that quite a bit with ideas like #11660.
* This PR (and the follow-ons) have no impact on time to return the cached response.
* Something was done between v1.49.1 and `develop` to make the cached response quite a bit faster.

DMRobertson:
Thanks Patrick, this information and process looks like gold dust!

I was a bit surprised by the 50x and 100x numbers, because I wouldn't expect things to be slower with caches. But I think the numbers in the cached column are relative to the performance of `develop` with warm caches, not `develop` with cold caches. (In other numbers, compare numbers columnwise, and not rowwise).

I think this looks good to merge!

DMRobertson (approved):

clokep:
> I was a bit surprised by the 50x and 100x numbers, because I wouldn't expect things to be slower with caches. But I think the numbers in the cached column are relative to the performance of `develop` with warm caches, not `develop` with cold caches. (In other numbers, compare numbers columnwise, and not rowwise).

Sorry this wasn't clear -- the **Uncached** column means "warmed caches, but nothing in the response cache", while the **Cached** column means "warmed caches, and there's a result ready in the response cache". The numbers are normalized column-wise, so you cannot compare across a row.

The 50x and 100x numbers are due to the bug in v1.49.0 that caused it to *not* cache the bundled aggregations (that's what this PR is fixing!). The "`develop` with agg" row is essentially re-introducing that bug, but on `develop` for comparison purposes.

PR #11660, opened 2021-12-29
Fetch edits for multiple events in a single query

clokep:
This fetches edits for multiple events in a single query, which should reduce load when trying to calculate bundled aggregations for many events.

I did some very unscientific testing this seems to result in ~10 - 20% speedup when doing a `/sync` with a large account. I'm not sure that's accurate to real workloads, but it seems to truly help.

My hope would be to do similar queries for the other bundled aggregations, although I'm only confident this is possible for thread summaries.

Part of #11825.

~~Based on #11612~~

~~Should be reviewable commit-by-commit.~~ Sorry, we're way past that...I can rebase if people would like. 😄


synapse/storage/databases/main/relations.py
```diff
@@ -295,60 +332,96 @@ def _get_aggregation_groups_for_event_txn(
...
+            return edits_by_original
+
         # We only allow edits for `m.room.message` events that have the same sender
         # and event type. We can't assert these things during regular event auth so
         # we have to do the checks post hoc.
 
         # Fetches latest edit that has the same type and sender as the
         # original, and is an `m.room.message`.
+        #
+        # TODO Should this ensure it does not return results for state events / redacted events?
```
clokep, line 375: These checks were done by `_get_bundled_aggregation_for_event` for the original event, which this replaces. So I think we should do them? I wasn't sure how to check them though.

synapse/storage/databases/main/relations.py
```diff
@@ -13,26 +13,63 @@
...
+    ):
+        super().__init__(database, db_conn, hs)
+
+        self._msc1849_enabled = hs.config.experimental.msc1849_enabled
+        self._msc3440_enabled = hs.config.experimental.msc3440_enabled
+
+        self.get_applicable_edit: LruCache[str, Optional[EventBase]] = LruCache(
+            cache_name="get_applicable_edit",
+            max_size=hs.config.caches.event_cache_size,  # TODO
+        )
```
clokep, line 71: I'm not sure how we usually decide what to use the `max_size` in our caches? I guess this could be set to whatever the default for the `@cached` decorator is since we're replacing that.

synapse/storage/databases/main/relations.py
```diff
@@ -295,60 +332,96 @@ def _get_aggregation_groups_for_event_txn(
...
         # and event type. We can't assert these things during regular event auth so
         # we have to do the checks post hoc.
 
         # Fetches latest edit that has the same type and sender as the
         # original, and is an `m.room.message`.
+        #
+        # TODO Should this ensure it does not return results for state events / redacted events?
         sql = """
-            SELECT edit.event_id FROM events AS edit
+            SELECT original.event_id, edit.event_id FROM events AS edit
```
clokep, line 377: Ideally we would be only pull out the latest edit event per original event, but I haven't found a reasonable way to do that (maybe a lateral join, but that's not supported on sqlite). Any thoughts on how to improve this would be appreciated!

synapse/storage/databases/main/relations.py
clokep: I'm also not 100% sure that this is truly interchangeable -- does the cache invalidation code work still with this?

clokep:
I'm going to request review on this while leaving it in draft. I suspect there's a bit more to do here, but would like to get some thoughts from the team! I know there's at least a couple of TODO comments to handle.

clokep:
I rebased this on top of #11659 as I don't see any downside to doing that, see https://github.com/matrix-org/synapse/pull/11659#issuecomment-1012119621 for some performance testing of this.

synapse/storage/databases/main/relations.py
```diff
@@ -355,17 +352,17 @@ async def get_applicable_edit(
                 original.event_id = relates_to_id
                 AND edit.type = original.type
                 AND edit.sender = original.sender
+                AND edit.room_id = original.room_id
```
DMRobertson, line 383: Is the point here that the DB already knows which room the original event was in, so we don't need to provide it ourselves?

(Just to make sure I understand more generally: is one event allowed to relate to another event in a different room?)

synapse/storage/databases/main/relations.py
```diff
@@ -63,6 +64,11 @@ def __init__(
         self._msc1849_enabled = hs.config.experimental.msc1849_enabled
         self._msc3440_enabled = hs.config.experimental.msc3440_enabled
 
+        self.get_applicable_edit: LruCache[str, Optional[EventBase]] = LruCache(
```
DMRobertson, line 68: Not sure about the name---`get_applicable_edit` sounds like a method to me. `applicable_edit_cache` maybe?

synapse/storage/databases/main/relations.py
```diff
@@ -325,57 +331,96 @@ def _get_aggregation_groups_for_event_txn(
...
 
-        def _get_applicable_edit_txn(txn: LoggingTransaction) -> Optional[str]:
-            txn.execute(sql, (event_id, RelationTypes.REPLACE))
-            row = txn.fetchone()
-            if row:
-                return row[0]
-            return None
-
-        edit_id = await self.db_pool.runInteraction(
+        def _get_applicable_edit_txn(txn: LoggingTransaction) -> Dict[str, str]:
```
DMRobertson, line 391: `edits` plural here, I think?

synapse/storage/databases/main/relations.py
DMRobertson: Sounds like we do want them.

Some kind of anti-join on the `redactions` and `state_events` tables? Not sure.

synapse/storage/databases/main/relations.py
DMRobertson: Does latest edit mean "edit with largest depth"?

Might we be able to use a trick like https://stackoverflow.com/a/27802817/5252017 ? (Sorry, not an expert here)

synapse/storage/databases/main/relations.py
DMRobertson: Unsure about these :(

synapse/storage/databases/main/relations.py
```diff
@@ -671,12 +664,21 @@ async def get_bundled_aggregations(
...
-        results = {}
+        # event ID -> bundled aggregation in non-serialized form.
+        results: Dict[str, Dict[str, Any]] = {}
+
+        event_ids = [event.event_id for event in events]
+
+        # Fetch any edits.
+        edits = await self._get_applicable_edits(event_ids)
+        for event_id, edit in edits.items():
+            results.setdefault(event_id, {})[RelationTypes.REPLACE] = edit
```
DMRobertson, line 675: I think the `setdefault` dance is used here to handle the case that an `event_id` occurs twice in `edits.items()`. Is that actually possible?

synapse/storage/databases/main/relations.py
```diff
@@ -671,12 +664,21 @@ async def get_bundled_aggregations(
...
+
+        # Fetch other relations per event.
         for event in events:
             event_result = await self._get_bundled_aggregation_for_event(event)
             if event_result is not None:
-                results[event.event_id] = event_result
+                results.setdefault(event.event_id, {}).update(event_result)
```
DMRobertson, line 682: When I was reading this, I thought "why should edits be special here?". I guess the point is that newer edits overrule older ones. But newer e.g. emoji reactions accumulate with older ones---the behavious is different?

DMRobertson (commented):
Some of this was a little over my head and I think I might be clutching at straws in places. Hopefully the feedback is vaguely useful? Happy to talk it through.

DMRobertson:
I've put this back to "draft" because I didn't realise there were some questions/work outstanding at first.

synapse/storage/databases/main/relations.py
clokep: > Is the point here that the DB already knows which room the original event was in, so we don't need to provide it ourselves?

Yeah, that's pretty much what we're taking advantage of. We want to do that so we can (potentially) query for edits to events across multiple rooms.

> (Just to make sure I understand more generally: is one event allowed to relate to another event in a different room?)

It is not allowed.

synapse/storage/databases/main/relations.py
clokep: Ah, I did this to reduce code-churn, but definitely should be cleaned up. Thanks for the prompt!

synapse/storage/databases/main/relations.py
clokep: No, it isn't necessary here, but is below when we call update. This was just me being dumb!

clokep: Note that `dits` is a map so an `event_id` cannot occur twice.

synapse/storage/databases/main/relations.py
clokep: Sorry, I'm not following. Are you asking why this is `update(...)` instead of setting a value? It's because `_get_bundled_aggregation_for_event` returns a dictionary of event ID -> aggregation type -> aggregation, while `_get_applicable_edits` returns event ID -> aggregation. We want to combine the edits into that dictionary and do ti by updating. We could also do the generic bundling first and then inject the edit into it (if that's clearer). That might do weird things with caching though and I would need to investigate!

synapse/storage/databases/main/relations.py
clokep: That was my initial thought too, thanks!

synapse/storage/databases/main/relations.py
clokep: It means the largest `origin_server_ts` (which is why we're ordering by that descending.)

reivilibre: > maybe a lateral join, but that's not supported on sqlite

N.B. SQLite does support lateral joins as long as you don't write the word `LATERAL` — if that's the only thing blocking you, you should be able to work around that by only inserting that word for Postgres.

synapse/storage/databases/main/relations.py
DMRobertson: Sorry, I wasn't doing a good job of expressing myself. I can see that you need to do a setdefault-update dance _here_ because we may have already added an edit for `event.event_id` in the previous loop. That bit's clear enough to me.

I was more wondering about the big picture. Why do we handling one kind of relation (edits on +672--675) separately to other kinds of relations (+677--681)?

clokep: Maybe the changes in #11752 will help with understanding the long term approach?

DMRobertson: I think I see now. I was just trying to work out if edits were somehow an "inherently different" kind of relation to the other kinds. (I.e. is the mental model different). But emphasising that this is a performance fix by reducing the number of DB queries we do makes sense!

synapse/storage/databases/main/relations.py
clokep: We have what we need in Python (we already have an `EventBase`) so easiest to do it there, I think. See 813218006b5f752f298ed0b1afac0aa799655055.

synapse/storage/databases/main/relations.py
clokep: @reivilibre Do you have any reference for that? I've been unable to get it to work (and my searching online has yielded "you can't do this on SQLite").

erikjohnston: On Postgres you can do `SELECT DISTINCT ON (original.event_id) ...` which will choose the first row (as defined by the `ORDER BY` clause). I don't think SQLite has that, and so you would need to use a different query, at which point you may as well fall back to the old behaviour.

Potentially you could use window functions and `first_value` , which I *think* both postgres and sqlite support, but those sorts of queries really are voodoo magic.

clokep: ❤️ @erikjohnston Thank you! That gave me some breadcrumbs to realize we do something very similar elsewhere:

https://github.com/matrix-org/synapse/blob/3e0536cd2afb5a640619bd872fc27b068ec3eb9b/synapse/storage/databases/main/end_to_end_keys.py#L616-L639

I think we can abstract out the changes in 8400c20086e7747336cc1cc174628ae182788bd7, but that is proving a bit tedious / invasive so I'd like to do it separately.

synapse/storage/databases/main/relations.py
```diff
@@ -363,37 +378,67 @@ async def get_applicable_edit(
...
+                INNER JOIN events AS original ON
+                    original.event_id = relates_to_id
+                    AND edit.type = original.type
+                    AND edit.sender = original.sender
+                    AND edit.room_id = original.room_id
+                WHERE
+                    %s
+                    AND relation_type = ?
+                    AND edit.type = 'm.room.message'
+                GROUP BY (original.event_id)
```
clokep, line 415: This query seems to not return the expected value. It seems to be unstable -- in the tests the `origin_server_ts` should never be duplicated, but sometimes this query is returning the max `edit.event_id` instead.

I suspect we should only ever use a single `MAX()` (or `MIN()` to get this behavior from SQLite, trying to see if there's a way to workaround this...

clokep: I ended up switching to the older, less efficient version for SQLite (where all edits are pulled) since the `MAX`/`MIN` trick doesn't seem to work well when you have multiple columns you want to order over.

clokep:
I retargeted @matrix-org/synapse-core with my review since this has undergone quite a bit of changes, but it might make sense for @DMRobertson or @reivilibre to look at it again with some previous context?

synapse/storage/databases/main/relations.py
```diff
@@ -363,37 +378,67 @@ async def get_applicable_edit(
...
+            sql = """
+                SELECT DISTINCT ON (original.event_id) original.event_id, edit.event_id FROM events AS edit
+                INNER JOIN event_relations USING (event_id)
+                INNER JOIN events AS original ON
+                    original.event_id = relates_to_id
+                    AND edit.type = original.type
+                    AND edit.sender = original.sender
+                    AND edit.room_id = original.room_id
+                WHERE
+                    %s
```
DMRobertson, line 394: I think this will get replaced with to `relates_to_id IN (...)`. A bit unfortunate that you have to skip down a few lines and do some guesswork to deduce this, but I guess that's just the way we have it.

synapse/storage/databases/main/relations.py
```diff
@@ -363,37 +378,67 @@ async def get_applicable_edit(
...
-                AND relation_type = ?
-                AND edit.room_id = ?
-                AND edit.type = 'm.room.message'
-            ORDER by edit.origin_server_ts DESC, edit.event_id DESC
-            LIMIT 1
-        """
+        if isinstance(self.database_engine, PostgresEngine):
+            # The `DISTINCT ON` clause will pick the *first* row it encounters,
+            # so ordering by origin server ts + event ID desc will ensure we get
+            # the latest edit.
```
DMRobertson, line 384: I'm not sure what the event ID has to do with it. Aren't they some pseudorandom identifier? Or perhaps the point is that we want a completely deterministic order in case of edits with the same timestamp?

synapse/storage/databases/main/relations.py
clokep: This is pretty typical of how we write these queries.

synapse/storage/databases/main/relations.py
clokep: We need a deterministic way for all servers to come to the same conclusion.

clokep:
> Do we have any tests which make use of the new behaviour (i.e. that would fetch edits for N events in 1 query rather than N?)

Yes, `tests.rest.client.test_relations.RelationsTestCase.test_multi_edit`. I fought with this test being flaky for a long time. 👍

synapse/storage/databases/main/relations.py
```diff
@@ -363,37 +378,67 @@ async def get_applicable_edit(
...
+            txn.execute(sql % (clause,), args)
+            return dict(cast(Iterable[Tuple[str, str]], txn.fetchall()))
+
+        edit_ids = await self.db_pool.runInteraction(
+            "get_applicable_edits", _get_applicable_edits_txn
         )
 
-        if not edit_id:
-            return None
+        edits = await self.get_events(edit_ids.values())  # type: ignore[attr-defined]
```
erikjohnston, line 432: Can we make this class depend on `EventsWorkerStore` instead of ignoring the error?

erikjohnston (approved):

synapse/storage/databases/main/relations.py
clokep: I can attempt to do that! Note that we have similar ignores all over, see #11165.

clokep: This gives other errors (similar to #11165) about inconsistent MROs. I'm going to leave this to be solved in #11165.

erikjohnston: Fair!

PR #11664, opened 2021-12-30
Add admin API to get users' account data

dklimpel:
Add admin API
```
GET /_synapse/admin/v1/users/<user_id>/accountdata
```

### Pull Request Checklist

### DIFF
--- a/docs/admin_api/user_admin_api.md
+++ b/docs/admin_api/user_admin_api.md
@@ -480,6 +480,81 @@ The following fields are returned in the JSON response body:
 - `joined_rooms` - An array of `room_id`.
 - `total` - Number of rooms.
 
+## Account Data
+Gets information about account data for a specific `user_id`.
+
+The API is:
+
+```
+GET /_synapse/admin/v1/users/<user_id>/accountdata
+```
+
+A response body like the following is returned:
+
+```json
+{
+    "account_data": {
+        "global": {
+            "m.secret_storage.key.LmIGHTg5W": {
+                "algorithm": "m.secret_storage.v1.aes-hmac-sha2",
+                "iv": "fwjNZatxg==",
+                "mac": "eWh9kNnLWZUNOgnc="
+            },
+            "im.vector.hide_profile": {
+                "hide_profile": true
+            },
+            "org.matrix.preview_urls": {
+                "disable": false
+            },
+            "im.vector.riot.breadcrumb_rooms": {
+                "rooms": [
+                    "!LxcBDAsDUVAfJDEo:matrix.org",
+                    "!MAhRxqasbItjOqxu:matrix.org"
+                ]
+            },
+            "m.accepted_terms": {
+                "accepted": [
+                    "https://example.org/somewhere/privacy-1.2-en.html",
+                    "https://example.org/somewhere/terms-2.0-en.html"
+                ]
+            },
+            "im.vector.setting.breadcrumbs": {
+                "recent_rooms": [
+                    "!MAhRxqasbItqxuEt:matrix.org",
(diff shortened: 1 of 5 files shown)


synapse/rest/admin/users.py
```diff
@@ -66,12 +66,12 @@ class UsersRestServletV2(RestServlet):
     """
 
     def __init__(self, hs: "HomeServer"):
-        self.store = hs.get_datastore()
-        self.auth = hs.get_auth()
+        self._store = hs.get_datastore()
+        self._auth = hs.get_auth()
```
reivilibre, line 70: This rename might be better served as a separate PR to keep the changes being proposed easy to track.

reivilibre (commented):
Seems generally fine.

Somewhat curious about the motivation (use case) for this data.

Does it make sense for this to be an admin API endpoint? I thought you could already create an access token to control an arbitrary user as an admin — would that not allow you to use the existing client-server APIs?

dklimpel:
> Does it make sense for this to be an admin API endpoint? I thought you could already create an access token to control an arbitrary user as an admin — would that not allow you to use the existing client-server APIs?

IMO with this argument, there would be almost no need for an admin API at all.
If you want to support users, you want to know what the server knows about the user to find out why a client behaves the way it does. I can also use it to check whether all of a user's data has really been deleted (#11539) or if key backup is configured or not.

Creating a token and then using the client APIs is complicated (also manual direct access in the database). If I log in as another user (create a access token), I have to justify why I can fake to be the user and have to prove that I do not know the password.
I don't know any client with which I can simulate the user, because a client allows a login with password and not with access token.

reivilibre:
> IMO with this argument, there would be almost no need for an admin API at all.

Yes, I'm aware of that (see below for my misunderstanding).

> If you want to support users, you want to know what the server knows about the user to find out why a client behaves the way it does. I can also use it to check whether all of a user's data has really been deleted (#11539) or if key backup is configured or not.

Good to know what this is useful for, though — as I was fairly confused as to why you would actually want the information.

> Creating a token and then using the client APIs is complicated (also manual direct access in the database). If I log in as another user (create a access token), I have to justify why I can fake to be the user and have to prove that I do not know the password.

I think this came from a misunderstanding. In Synapse logs, I have previously seen that the user can be shown as `@admin:server|@user:server` in the case where an admin is controlling a user. I thought there was a way to use your admin access token and specify a user ID you wanted to control (appservices work a bit like this, which probably explains why I thought it existed).
In that case you'd just be able to use your admin access token with the client API.

However, I must have imagined things (it's only used in certain cases like creating room admins, etc …. https://github.com/matrix-org/synapse/blob/dd4778875213d9cb8be7ee71d32751fbd6cdaba2/synapse/rest/admin/rooms.py#L602-L605) — indeed it looks like you'd have to mess in the database to mint a token (not what we want).

Will give it a proper review shortly :).

docs/admin_api/user_admin_api.md
```diff
@@ -480,6 +480,81 @@ The following fields are returned in the JSON response body:
...
+
+- `user_id` - fully qualified: for example, `@user:server.com`.
+
+**Response**
+
+The following fields are returned in the JSON response body:
+
+- `account_data` - An array containing the account data for the user
+  - `global` - An array containing the global account data for the user
+  - `rooms` - An array containing the account data per room for the user
```
reivilibre, line 556: I would call these 'maps'.
```suggestion
- `account_data` - A map containing the account data for the user
  - `global` - A map containing the global account data for the user
  - `rooms` - A map containing the account data per room for the user
```

reivilibre (approved):

PR #11666, opened 2021-12-30
Include the topic event in the prejoin state, per MSC3173.

clokep:
[MSC3173](https://github.com/matrix-org/matrix-doc/pull/3173) suggests that the topic of a room should be included in the prejoin state. Synapse does not currently do this, but it is easy enough to include it.

### DIFF
--- a/changelog.d/11666.feature
+++ b/changelog.d/11666.feature
@@ -1,0 +1,1 @@
+Include the room topic in the stripped state included with invites and knocking.
--- a/docs/sample_config.yaml
+++ b/docs/sample_config.yaml
@@ -1488,6 +1488,7 @@ room_prejoin_state:
    # - m.room.encryption
    # - m.room.name
    # - m.room.create
+   # - m.room.topic
    #
    # Uncomment the following to disable these defaults (so that only the event
    # types listed in 'additional_event_types' are shared). Defaults to 'false'.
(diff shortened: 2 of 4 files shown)


DMRobertson:
Seems sensible! Do we have any tests which cover this?

DMRobertson:
> Do we have any tests which cover this?

FWIW, looks like the list of events to retrain in the stripped summary gets used in
- handling a knock request
- calculating `invite_room_state` when persisting an invite or knock event

clokep:
> Seems sensible! Do we have any tests which cover this?

I'm not sure if we have specific tests which look at which state this is used in, but I'll double check.

clokep:
Some of the knocking tests asserting that the prejoin state was sent properly. I added the topic to those tests to ensure it works properly.

DMRobertson (approved):
Thanks, LGTM! I guess the test didn't fail before the most recent commits because all of the stripped state stuff is optional on the homeserver's part?

clokep:
> Thanks, LGTM! I guess the test didn't fail before the most recent commits because all of the stripped state stuff is optional on the homeserver's part?

Modifying the test on `develop` did cause it to fail, but it didn't fail on this branch because we only send the state that exists and the test wasn't setting a topic, so as far as it was concerned everything was working fine.

clokep:
> > Thanks, LGTM! I guess the test didn't fail before the most recent commits because all of the stripped state stuff is optional on the homeserver's part?
> 
> Modifying the test on `develop` did cause it to fail, but it didn't fail on this branch because we only send the state that exists and the test wasn't setting a topic, so as far as it was concerned everything was working fine.

(This is a bit brittle, but I'm not sure of a better way to do this...)

PR #11677, opened 2022-01-03
Fix SimpleHttpClient not sending Accept header in `get_json`

Fr3shTea:
I suppose that in `SimpleHttpClient.get_json` function, `actual_headers` dict should be used in `SimpleHttpClient.get_raw` call instead of passed `headers` dict.

### DIFF
--- a/changelog.d/11677.bugfix
+++ b/changelog.d/11677.bugfix
@@ -1,0 +1,1 @@
+Fix wrong variable reference in `SimpleHttpClient.get_json` that results in abscense of "Accept" header in the request.
--- a/changelog.d/11677.bugfix
+++ b/changelog.d/11677.bugfix
@@ -1,1 +1,1 @@
-Fix wrong variable reference in `SimpleHttpClient.get_json` that results in abscense of "Accept" header in the request.
+Fix wrong variable reference in `SimpleHttpClient.get_json` that results in the absence of the `Accept` header in the request.
--- a/synapse/http/client.py
+++ b/synapse/http/client.py
@@ -588,7 +588,7 @@ class SimpleHttpClient:
         if headers:
             actual_headers.update(headers)  # type: ignore
 
-        body = await self.get_raw(uri, args, headers=headers)
+        body = await self.get_raw(uri, args, headers=actual_headers)
         return json_decoder.decode(body.decode("utf-8"))
 
     async def put_json(


reivilibre (changes requested):
Thanks for spotting this and sending a fix in!
Please would you add a news file — you're very welcome (encouraged, even) to also add something like `Contributed by <your name or @Fr3shTea>.` to the end of it so you will be credited in the change log.

changelog.d/11677.bugfix
```diff
@@ -0,0 +1 @@
+Fix wrong variable reference in `SimpleHttpClient.get_json` that results in abscense of "Accept" header in the request.
```
reivilibre, line 1:
```suggestion
Fix wrong variable reference in `SimpleHttpClient.get_json` that results in the absence of the `Accept` header in the request.
```

reivilibre (approved):
Thank you!

changelog.d/11677.bugfix
```diff
@@ -0,0 +1 @@
+Fix wrong variable reference in `SimpleHttpClient.get_json` that results in the absence of the `Accept` header in the request.
```
richvdh, line 1: This isn't a *great* newsfile, on two counts:

* People using synapse don't normally care *how* a bug gets fixed (and if they do, they can click through to the PR). So "wrong variable reference" and "`SimpleHttpClient.get_json`" aren't useful here and we should just omit the first part of the sentence.
* We talk about "the request", but what request? You're assuming that people know that `SimpleHttpClient` is used to make outbound HTTP requests, but most people reading the changelog *don't* know that, and in any case it's not used for all outbound requests. Anyway, what is the impact of omitting an `Accept` header?

Generally it would be great if we could rephrase this in terms of a change in functionality that people will actually see.

@Fr3shTea maybe if you could propose some better wording we can just commit it to the repo.

Fr3shTea: This is a very small change and I'm not sure that it will have significant effect in terms of change in functionality. Maybe something is relying on that `Accept` header, but I'm not sure about that either. I mostly created this newsfile because reivilibre told me to do that and I can't really think of any better wording for that small change.

Fr3shTea: Maybe newsfile might be unnecessary for such fix?

richvdh: generally if we don't think it's going to have a noticeable effect, we give it a `.misc` newsfile so that it appears under "Internal changes".

I've actually realised this has now made it into the changelog for 1.50: https://github.com/matrix-org/synapse/blob/83acdb23fe318f8a41d2a6b800d994d46754a903/CHANGES.md, so it's too late to change the newsfile anyway.

@reivilibre can I hand this over to you to do *something* with in the 1.50 changelog?

reivilibre: @richvdh Yup I'll take care of it

PR #11680, opened 2022-01-04
update ngnix reverse-proxy example

richvdh:
this should not be a case-insensitive match.

### DIFF
--- a/docs/reverse_proxy.md
+++ b/docs/reverse_proxy.md
@@ -63,7 +63,7 @@ server {
 
     server_name matrix.example.com;
 
-    location ~* ^(\/_matrix|\/_synapse\/client) {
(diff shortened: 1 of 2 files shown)


docs/reverse_proxy.md
```diff
@@ -63,7 +63,7 @@ server {
 
     server_name matrix.example.com;
 
-    location ~* ^(\/_matrix|\/_synapse\/client) {
+    location ~ ^(/_matrix|/_synapse/client) {
```
richvdh, line 66: we may as well remove the backslashes here while we're at it - they are purely obfuscatory

clokep (approved):
From reading the nginx config docs this seems reasonable.

PR #11683, opened 2022-01-04
Drop support for and remove references to EOL Python 3.6

H-Shay:
As the title states, removes references to python 3.6 and upgrades to Python 3.7 where necessary.


H-Shay:
Requires https://github.com/matrix-org/sytest/pull/1181, which is blocked by https://github.com/matrix-org/sytest/issues/1114, but there is a PR open to fix that: https://github.com/matrix-org/sytest/pull/1182

H-Shay:
Looks like `Pillow 4` is not compatible with Python 3.7: https://pillow.readthedocs.io/en/latest/installation.html
Can we bump the version? Looks like we will need to bump it to at least 5.2

tests/crypto/test_event_signing.py
```diff
@@ -36,8 +36,8 @@
 class EventSigningTestCase(unittest.TestCase):
     def setUp(self):
         self.signing_key = nacl.signing.SigningKey(SIGNING_KEY_SEED)
-        self.signing_key.alg = KEY_ALG
-        self.signing_key.version = KEY_VER
+        self.signing_key.alg = KEY_ALG  # type: ignore
+        self.signing_key.version = KEY_VER  # type: ignore
```
clokep, line 40: This was separately fixed in #11714

.ci/scripts/test_old_deps.sh
```diff
@@ -1,12 +1,14 @@
...
+# Prevent tzdata from asking for user input
+export DEBIAN_FRONTEND=noninteractive
+
 set -ex
 
 apt-get update
-apt-get install -y python3 python3-dev python3-pip libxml2-dev libxslt-dev xmlsec1 zlib1g-dev tox
+apt-get install -y python3 python3-dev python3-pip libxml2-dev libxslt-dev xmlsec1 zlib1g-dev tox libjpeg-dev libwebp-dev
 
 export LANG="C.UTF-8"
```
H-Shay, line 14: These were added to support `Pillow`

synapse/python_dependencies.py
```diff
@@ -107,7 +107,7 @@
     # `systemd.journal.JournalHandler`, as is documented in
     # `contrib/systemd/log_config.yaml`.
     "systemd": ["systemd-python>=231"],
-    "url_preview": ["lxml>=3.5.0"],
+    "url_preview": ["lxml>=4.2.0"],
     "sentry": ["sentry-sdk>=0.7.2"],
     "opentracing": ["jaeger-client>=4.0.0", "opentracing>=2.2.0"],
     "jwt": ["pyjwt>=1.6.4"],
```
H-Shay, line 113: These were bumped to be compatible with py 37 and up

tox.ini
```diff
@@ -117,8 +117,7 @@ usedevelop=true
 skip_install = true
 usedevelop = false
 deps =
-    # Old automat version for Twisted
-    Automat == 0.3.0
+    Automat == 0.8.0
     lxml
     {[base]deps}
```
H-Shay, line 123: This was bumped for python 3.8 compatibility

synapse/python_dependencies.py
```diff
@@ -70,7 +70,7 @@
     "pyasn1>=0.1.9",
     "pyasn1-modules>=0.0.7",
     "bcrypt>=3.1.0",
-    "pillow>=4.3.0",
```
DMRobertson, line 73: I think this has come up before, but I was a bit surprised to see this. Broadly speaking, a version of Pillow that successfully ran on 3.6 should also successfully run on 3.7, 3.8, 3.9... etc. (There are deprecations and removals between minor Python releases, but these are uncommon.)  The pain is that Pillow uses a C extension and therefore needs binaries to be built for each interpreter version.

As you pointed out, olddeps was failing here: https://github.com/matrix-org/synapse/runs/4745589267?check_suite_focus=true

For completeness, I think what's happening here is:
- we used to be able to pull a binary wheel from PyPI on bionic. [See e.g.](https://github.com/matrix-org/synapse/runs/4840259529?check_suite_focus=true#step:4:1212)
  > 2022-01-17T11:47:41.2123816Z Collecting pillow==4.3.0
  > 2022-01-17T11:47:41.2318857Z   Downloading Pillow-4.3.0-cp36-cp36m-manylinux1_x86_64.whl (5.8 MB)
- Focal uses Python 3.8.
- Pillow doesn't "support" Python 3.8 [until Pillow 6.2.1](https://pillow.readthedocs.io/en/stable/installation.html#python-support). I think "support" means "provide binary wheels for".
- We still want debian oldstable to use its version, 5.4.1, so don't want to bump the min version to 6.2.1.
- For some reason, `pip/tox` in the olddeps job chooses Pillow 5.4.0 exactly. This doesn't have a Python 3.8 wheel.
- Therefore we need to build the C extension ourselves from scratch. And that's also why we need `libjpeg-devel`, for instance.

synapse/python_dependencies.py
DMRobertson: Have we run this one past the package maintainers?

DMRobertson: For completeness, judging by https://github.com/matrix-org/synapse/runs/4764560526?check_suite_focus=true#step:4:1379, it looks like the C-extension in lxml 3.5 wasn't compatible with CPython 3.8, but the C-extension in lxml 4.2.0 is compatible?

tox.ini
```diff
@@ -117,8 +117,7 @@ usedevelop=true
 skip_install = true
 usedevelop = false
 deps =
-    # Old automat version for Twisted
-    Automat == 0.3.0
+    Automat == 0.8.0
```
DMRobertson, line 120: Also a bit surprised by this one. I would have thought that twisted would specify which version of `Automat` it needs!

DMRobertson (changes requested):
Thanks for seeing this through. I've tried to elaborate on _why_ we needed to bump things for `olddeps`. Hope it's useful, as a historical record if nothing else.

Two small things:
- Have we run the lxml version bump past package maintainers?
- The twisted_trunk workflow speaks of Python 3.6. Could probably change this to "3.7" or "3.x"?

synapse/python_dependencies.py
H-Shay: yes that's my understanding-something about the C-headers changing?

H-Shay:
I've shouted in the packaging room, haven't heard of issues, so hopefully this is good to go.

DMRobertson (approved):
LGTM. Thanks for your diligence on this one!

anoadragon453 (commented):
Heads up that there are a couple checks for Python 3.7+ that can be removed here:

https://github.com/matrix-org/synapse/blob/20c6d85c6e8f721b8688b7f1361c7a7bab2449fd/synapse/app/_base.py#L467-L480

synapse/app/_base.py
```diff
@@ -468,16 +467,12 @@ def run_sighup(*args: Any, **kwargs: Any) -> None:
     # everything currently allocated are things that will be used for the
     # rest of time. Doing so means less work each GC (hopefully).
     #
-    # This only works on Python 3.7
-    if platform.python_implementation() == "CPython" and sys.version_info >= (3, 7):
```
clokep, line 472: I think we need to still check the platform implementation is `CPython` here -- I think pypy doesn't have these APIs.

callahad: pypy has `gc.collect` but not `gc.freeze`

In the interest of duck typing, we should probably check `hasattr(gc, 'freeze')` instead of examining `python_implementation()`

(Or we could say we're standardizing on CPython and rip the Pypy bits out of our codebase; no shame in that)

clokep: Either of those seems ok, just a shame to delete something that's working.

H-Shay: Great, I will re-implement the relevant parts of this check. Since this is already merged, what would be the best way of going about that? Opening a new PR restoring that functionality?

clokep: > Great, I will re-implement the relevant parts of this check. Since this is already merged, what would be the best way of going about that? Opening a new PR restoring that functionality?

Exactly! It probably makes sense to just use the same changelog entry so they'll get merged by towncrier.

PR #11684, opened 2022-01-04
Run `pyupgrade --py37-plus --keep-percent-format` on Synapse

H-Shay:
As the title states. I am sending up (hopefully!) manageable chunks of pyupgrade changes, directory by directory. Much more to come.


DMRobertson:
Do we want `--py37-plus` here? Quoting from pyupgrade's readme:

> --py37-plus will remove Python <= 3.6 only blocks

H-Shay:
Yes apparently I am an idiot and read that wrong. Thank you!

H-Shay:
(Although I just ran it an doesn't seem to be a difference). I will fix title and newsfragment, though.

PR #11685, opened 2022-01-04
Run `pyupgrade --py37-plus --keep-percent-format` on Synapse

H-Shay:
As the title states! Part of the task list for dropping EOL Python 3.6.


dklimpel:
There is a number error in the news file.

changelog.d/11685.misc
```diff
@@ -0,0 +1,2 @@
+Run `pyupgrade --py37-plus --keep-percent-format` on Synapse.
+
```
clokep, line 2: Seems to have an extra blank line, not sure if the news scripts will deal with that OK?

```suggestion
Run `pyupgrade --py37-plus --keep-percent-format` on Synapse.
```

clokep (approved):
LGTM!

PR #11690, opened 2022-01-05
Include `io.element.thread` capability for MSC3440.

### LINKED ISSUE #11689: Add a capabilties flag for MSC3440
Per https://github.com/matrix-org/matrix-doc/commit/fc81bbd836955876c931c95277249981b3e2778c

clokep:
Fixes #11689, per MSC3440.

### DIFF
--- a/synapse/rest/client/capabilities.py
+++ b/synapse/rest/client/capabilities.py
@@ -73,6 +73,9 @@ class CapabilitiesRestServlet(RestServlet):
                 "enabled": self.config.registration.enable_3pid_changes
             }
 
+        if self.config.experimental.msc3440_enabled:
+            response["capabilities"]["io.element.thread"] = {"enabled": True}
+
         return 200, response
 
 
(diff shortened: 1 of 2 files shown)


synapse/rest/client/capabilities.py
```diff
@@ -73,6 +73,9 @@ async def on_GET(self, request: SynapseRequest) -> Tuple[int, JsonDict]:
                 "enabled": self.config.registration.enable_3pid_changes
             }
 
+        if self.config.experimental.msc3440_enabled:
+            response["capabilities"]["io.element.thread"] = {"enabled": True}
```
richvdh, line 77: I think this should maybe be `unstable.*`. will ask the SCT for opinions

clokep: I think in the past we've just namespaced them (i.e. not using `m.*`), but easy enough to change as long as clients know.

richvdh: seems I was being a bit of a crank.

PR #11691, opened 2022-01-05
Use buildkit's cache feature to speed up docker builds

richvdh:
Having spent much of the last week attempting to run complement tests from somewhere with damp string instead of internet... something had to be done.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -366,6 +366,8 @@ jobs:
       # Build initial Synapse image
       - run: docker build -t matrixdotorg/synapse:latest -f docker/Dockerfile .
         working-directory: synapse
+        env:
+          DOCKER_BUILDKIT: 1
 
       # Build a ready-to-run Synapse image based on the initial image above.
       # This new image includes a config file, keys for signing and TLS, and
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -11,7 +11,7 @@
 # There is an optional PYTHON_VERSION build argument which sets the
 # version of python to build against: for example:
 #
-#    DOCKER_BUILDKIT=1 docker build -f docker/Dockerfile --build-arg PYTHON_VERSION=3.6 .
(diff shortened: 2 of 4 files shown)


docker/Dockerfile
```diff
@@ -1,14 +1,17 @@
...
 # synapse repository:
 #
-#    docker build -f docker/Dockerfile .
+#    DOCKER_BUILDKIT=1 docker build -f docker/Dockerfile .
 #
 # There is an optional PYTHON_VERSION build argument which sets the
 # version of python to build against: for example:
 #
-#    docker build -f docker/Dockerfile --build-arg PYTHON_VERSION=3.6 .
+#    DOCKER_BUILDKIT=1 docker build -f docker/Dockerfile --build-arg PYTHON_VERSION=3.6 .
```
DMRobertson, line 14: Maybe `PYTHON_VERSION=3.7` while we're at it?

.github/workflows/tests.yml
```diff
@@ -366,6 +366,8 @@ jobs:
       # Build initial Synapse image
       - run: docker build -t matrixdotorg/synapse:latest -f docker/Dockerfile .
         working-directory: synapse
+        env:
+          DOCKER_BUILDKIT: 1
```
DMRobertson, line 370: Am I right in thinking that this doesn't make the `tests` workflow faster; but it ensures the built image has support for this caching setup?

docker/Dockerfile
richvdh: good point. Or more!

.github/workflows/tests.yml
richvdh: neither, I'm afraid. (The idea is not to change anything about the built image; it is all about the process of building that image).

This is needed because otherwise the `docker build` command will complain about not understanding the `--mount` flags in the Dockerfile. Those flags are unlikely to do much in this workflow, but they are super helpful when working locally, hence wanting them to be there.

DMRobertson (approved):

DMRobertson:
One downside of this is that it's tricky to interactively debug problems from a buildkit build. https://github.com/moby/buildkit/issues/1472 has more to say.

PR #11692, opened 2022-01-05
Use auto_attribs/native type hints for attrs classes.

clokep:
This updates our various `attrs` classes to use `auto_attribs` where reasonable and native type hints everywhere (e.g. `foo: bar = attr.ib(...)` instead of `foo = attr.ib(type=bar, ...)`).

This got a bit big so I can cut it in half if people would like, but it should be reviewable commit-by-commit and pretty straightforward.

There were a few instances of classes without type hints which I filled in while I was doing this, I also change some classes to `frozen`, but I suspect more could be done.


clokep:
(This would be nice to enforce with a lint or something, but i haven't checked if any such thing exists.)

synapse/types.py
```diff
@@ -225,10 +225,10 @@ class DomainSpecificString(metaclass=abc.ABCMeta):
 
     SIGIL: ClassVar[str] = abc.abstractproperty()  # type: ignore
 
-    localpart = attr.ib(type=str)
-    domain = attr.ib(type=str)
+    localpart: str
+    domain: str
 
-    # Because this is a frozen class, it is deeply immutable.
+    # Because this is a frozen class, it is deeplyx immutable.
```
DMRobertson, line 231:
```suggestion
    # Because this is a frozen class, it is deeply immutable.
```

synapse/util/async_helpers.py
```diff
@@ -309,12 +309,12 @@ def gather_results(  # type: ignore[misc]
...
 
-@attr.s(slots=True)
+@attr.s(slots=True, auto_attribs=True)
 class _LinearizerEntry:
     # The number of things executing.
-    count = attr.ib(type=int)
+    count: int
     # Deferreds for the things blocked from executing.
-    deferreds = attr.ib(type=collections.OrderedDict)
+    deferreds: collections.OrderedDict
```
DMRobertson, line 317: I wonder if this could just be `dict` now we're on 3.7+? Probably not worth changing.

synapse/util/caches/dictionary_cache.py
```diff
@@ -33,22 +33,21 @@
...
+        known_absent: Keys that were looked up in the dict and were not there.
         value: The full or partial dict value
     """
 
-    full = attr.ib(type=bool)
-    known_absent = attr.ib(type=Set[Any])  # should be: Set[DKT]
-    value = attr.ib(type=Dict[Any, Any])  # should be: Dict[DKT, DV]
+    full: bool
+    known_absent: Set[Any]  # should be: Set[DKT]
+    value: Dict[Any, Any]  # should be: Dict[DKT, DV]
```
DMRobertson, line 50: https://github.com/python-attrs/attrs/issues/313 again! Argh.

DMRobertson: Now we're on 3.7 I wonder if we can use a stdlib [dataclass](https://docs.python.org/3/library/dataclasses.html) here? We're not using any of the extra attrs machinery here.

synapse/logging/_remote.py
```diff
@@ -54,10 +54,10 @@ class LogProducer:
 
     # This is essentially ITCPTransport, but that is missing certain fields
     # (connected and registerProducer) which are part of the implementation.
-    transport = attr.ib(type=Connection)
-    _format = attr.ib(type=Callable[[logging.LogRecord], str])
-    _buffer = attr.ib(type=deque)
-    _paused = attr.ib(default=False, type=bool, init=False)
+    transport: Connection
+    _format: Callable[[logging.LogRecord], str]
+    _buffer: deque
```
DMRobertson, line 59: Can we make this hint a [`typing.Deque`](https://docs.python.org/3/library/typing.html?highlight=typing%20deque#typing.Deque) so we can specify the type of its elements?

synapse/storage/databases/main/registration.py
```diff
@@ -69,14 +69,14 @@ class TokenLookupResult:
...
-    valid_until_ms = attr.ib(type=Optional[int], default=None)
-    token_owner = attr.ib(type=str)
-    token_used = attr.ib(type=bool, default=False)
+    user_id: str
+    is_guest: bool = False
+    shadow_banned: bool = False
+    token_id: Optional[int] = None
+    device_id: Optional[str] = None
+    valid_until_ms: Optional[int] = None
+    token_owner: str = attr.ib()
```
DMRobertson, line 78:
```suggestion
    token_owner: str
```

DMRobertson (commented):
Thanks for this. I've a few suggestions that are probably out of scope for this one, and a few smaller typos.

I was a bit scared at the start when some of the classes were made `frozen=True`. I think I'll probably trust CI on this one...

synapse/util/caches/dictionary_cache.py
clokep: Maybe, but I'd rather that be left to a separate PR.

synapse/storage/databases/main/registration.py
clokep: This needs to be `attr.ib()` for two reasons:

* The one after it has a default.
* Without this, the decorator below won't work (`@token_owner.default` bit).

clokep:
> I was a bit scared at the start when some of the classes were made `frozen=True`. I think I'll probably trust CI on this one...

The majority of them are either things from config or things from the database, so they really shouldn't be modified (and I did manually check most of them).

DMRobertson (commented):
Thanks for clarifying. The other comments are suggestions for other PRs; I think this is great as it is!

DMRobertson (approved):

synapse/util/async_helpers.py
clokep: I think keeping it as `OrderedDict` is clearer for now!

PR #11695, opened 2022-01-06
Support spaces with > 50 rooms in the /hierarchy endpoint

### LINKED ISSUE #11134: Hierarchy pagination is limited to 50 items per space
`_matrix/client/unstable/org.matrix.msc2946/rooms/{roomId}/hierarchy` returns a next_batch string for spaces with more than 50 children as expected.
Pparsing the pagination token then only returns 11 rooms in my case, even though there should be much more. The Space in question has 135 space children.
Different spaces return different seemingly random numbers after the second call, but never the actual number.

**To Reproduce**

- call `_matrix/client/unstable/org.matrix.msc2946/rooms({roomId}/hierarchy` for a space with more than 50 children, ideally more than 100

- repeat the call and parse the next_batch string

- wrong number of spaces are returned

**Expected behavior**

- return actual number of spaces and if >50 another next_batch string

tested with v1.43.0

clokep:
This updates the support for the `/hierarchy` endpoint to support > 50 rooms in a single space.

This is done by duplicating some logic for the legacy `/spaces` API in callers to `_summarize_local_room`, which is also used by the `/hierarchy` endpoint.

The initial commit asserts the current logic on `develop` and should pass those tests, I then updated it for fixing this bug which is only fixed for `/hierarchy`, not `/spaces`.

Fixes #11134

### DIFF
--- a/synapse/handlers/room_summary.py
+++ b/synapse/handlers/room_summary.py
@@ -153,6 +153,9 @@ class RoomSummaryHandler:
         rooms_result: List[JsonDict] = []
         events_result: List[JsonDict] = []
 
+        if max_rooms_per_space is None or max_rooms_per_space > MAX_ROOMS_PER_SPACE:
+            max_rooms_per_space = MAX_ROOMS_PER_SPACE
+
         while room_queue and len(rooms_result) < MAX_ROOMS:
             queue_entry = room_queue.popleft()
             room_id = queue_entry.room_id
@@ -167,7 +170,7 @@ class RoomSummaryHandler:
             # The client-specified max_rooms_per_space limit doesn't apply to the
             # room_id specified in the request, so we ignore it if this is the
             # first room we are processing.
-            max_children = max_rooms_per_space if processed_rooms else None
+            max_children = max_rooms_per_space if processed_rooms else MAX_ROOMS
 
             if is_in_room:
                 room_entry = await self._summarize_local_room(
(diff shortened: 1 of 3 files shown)


clokep:
The downside of this is that if you have a space with an absurd number of rooms in it then this could cause an extremely large response / large amount of time serializing JSON. The data has already been pulled from the database though, so I don't think it will impact there at all.

clokep:
Although over federation this will be pretty meh -- it will return:

* The requested room with all of the children state.
* The summary of the first 50 rooms.

This would leave the requesting server to have to query the rest of the rooms individually.

synapse/handlers/room_summary.py
```diff
@@ -583,7 +590,9 @@ async def get_federation_hierarchy(
 
         # Iterate through each child and potentially add it, but not its children,
         # to the response.
-        for child_room in root_room_entry.children_state_events:
+        for child_room in itertools.islice(
+            root_room_entry.children_state_events, MAX_ROOMS_PER_SPACE
+        ):
```
clokep, line 595: This changes the `/hierarchy` federation API in a not great way, it would now return:

* The requested room with all of the children state (so references to potentially unlimited).
* The summary of the first 50 rooms.

This would cause the requesting server to query an unknown children rooms individually over federation.

The spec doesn't mention anything about this, but the MSC does:

> Similarly, if a server-set limit on the size of the response is reached, additional rooms are not added to the response and can be queried individually.

squahtx (approved):
LGTM, keeping the caveats in mind.

PR #11699, opened 2022-01-06
Remove a Python 2-ism and improve type hints.

clokep:
Removes a Python 2-ism in our code where indexing into a byte-string on Python 2 returned a bytes object, but on Python 3 returns an int.

### DIFF
--- a/synapse/types.py
+++ b/synapse/types.py
@@ -405,8 +405,7 @@ def map_username_to_mxid_localpart(
     # we also do the =-escaping to mxids starting with an underscore.
     username = re.sub(b"^_", b"=5f", username)
 
-    # we should now only have ascii bytes left, so can decode back to a
-    # unicode.
+    # we should now only have ascii bytes left, so can decode back to a string.
     return username.decode("ascii")
 
 
--- a/synapse/types.py
+++ b/synapse/types.py
@@ -21,6 +21,7 @@ from typing import (
     ClassVar,
     Dict,
     Mapping,
+    Match,
     MutableMapping,
     Optional,
     Tuple,
@@ -380,7 +381,7 @@ def map_username_to_mxid_localpart(
             onto different mxids
 
     Returns:
-        unicode: string suitable for a mxid localpart
+        string suitable for a mxid localpart
     """
     if not isinstance(username, bytes):
         username = username.encode("utf-8")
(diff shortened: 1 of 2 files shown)


synapse/types.py
```diff
@@ -380,29 +381,24 @@ def map_username_to_mxid_localpart(
...
+        def f1(m: Match) -> bytes:
             return b"_" + m.group().lower()
 
         username = UPPER_CASE_PATTERN.sub(f1, username)
     else:
         username = username.lower()
 
-    # then we sort out non-ascii characters
-    def f2(m):
-        g = m.group()[0]
```
clokep, line 400: `m.group()` would be something like `b"x"` (which is equivalent to `"x"` on Python 2).

On Python 2, `g` would be something like `"x"`, so you call `ord(g)` to get `120`.

On Python 3, `g` would be something like `120` directly.

synapse/types.py
```diff
@@ -380,29 +381,24 @@ def map_username_to_mxid_localpart(
...
+        string suitable for a mxid localpart
     """
     if not isinstance(username, bytes):
         username = username.encode("utf-8")
 
     # first we sort out upper-case characters
     if case_sensitive:
 
-        def f1(m):
+        def f1(m: Match) -> bytes:
```
DMRobertson, line 392: Should this be the slightly more specific `Match[bytes]`? And in `f2` below.

DMRobertson (approved):
I think this looks good! I was a bit worried that `NON_MXID_CHARACTER_PATTERN` was a `Pattern[str]`, but line +364/5 has a `.encode("ascii")` which convinced me that the pattern is a `Pattern[bytes]`.

Minor, but I think +409 should say "decode to a string" rather than "to a unicode" in Py3 land? Otherwise LGTM!

DMRobertson:
> Otherwise LGTM!

(assuming CI is happy of course!)

clokep:
> I think this looks good! I was a bit worried that `NON_MXID_CHARACTER_PATTERN` was a `Pattern[str]`, but line +364/5 has a `.encode("ascii")` which convinced me that the pattern is a `Pattern[bytes]`.

Yesssss, it's confusing. 😄 

> Minor, but I think +409 should say "decode to a string" rather than "to a unicode" in Py3 land? Otherwise LGTM!

Done!

PR #11701, opened 2022-01-06
Test case for an edgy race

DMRobertson:
I have tried to capture the race described by @richvdh in [this comment](https://github.com/matrix-org/synapse/pull/11532#discussion_r769104461).

It passes when I ran the test against develop locally, and it fails when I run it against #11532.

### DIFF
--- a/tests/handlers/test_sync.py
+++ b/tests/handlers/test_sync.py
@@ -11,15 +11,14 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
 from typing import Optional
-from unittest.mock import Mock
+from unittest.mock import MagicMock, Mock, patch
 
 from synapse.api.constants import EventTypes, JoinRules
 from synapse.api.errors import Codes, ResourceLimitError
 from synapse.api.filtering import Filtering
 from synapse.api.room_versions import RoomVersions
-from synapse.handlers.sync import SyncConfig
+from synapse.handlers.sync import SyncConfig, SyncResult
 from synapse.rest import admin
 from synapse.rest.client import knock, login, room
 from synapse.server import HomeServer
@@ -27,6 +26,8 @@ from synapse.types import UserID, create_requester
 
 import tests.unittest
 import tests.utils
+from tests.test_utils import make_awaitable
+from tests.test_utils.event_injection import inject_event
 
 
 class SyncTestCase(tests.unittest.HomeserverTestCase):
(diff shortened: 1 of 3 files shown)


tests/handlers/test_sync.py
```diff
@@ -186,6 +187,139 @@ def test_unknown_room_version(self):
         self.assertNotIn(invite_room, [r.room_id for r in result.invited])
         self.assertNotIn(knock_room, [r.room_id for r in result.knocked])
 
+    def test_ban_wins_race_with_join(self):
+        """Check rooms appear under the "leave" section if they lose a race to a ban.
+
+        A complicated edge case. Best to quote RvdH from
+        https://github.com/matrix-org/synapse/pull/11532#discussion_r769104461:
```
clokep, line 194: I'd rather ensure we have the full contents described here, instead of linking off to a GitHub comment. Below seems verbatim, so I'm not sure the link offers much more.

tests/handlers/test_sync.py
```diff
@@ -186,6 +187,139 @@ def test_unknown_room_version(self):
...
+        #
+        # We do this by
+        # 1. Ban eve from the room.
+        # 2. Have eve join the room.
+        #    - mock the prev_events for the join event to precede the ban
+
+        # First, the ban.
+        eve = self.register_user("eve", "password")
+        eve_token = self.login(eve, "password")
+        self.get_success(
```
clokep, line 266: Can this just be a normal `send_state`?

tests/handlers/test_sync.py
```diff
@@ -186,6 +187,139 @@ def test_unknown_room_version(self):
...
+        eve_requester = create_requester(eve)
+        eve_sync_config = generate_sync_config(eve)
+        eve_sync_after_ban: SyncResult = self.get_success(
+            self.sync_handler.wait_for_sync_for_user(eve_requester, eve_sync_config)
+        )
+        since_token = eve_sync_after_ban.next_batch
+        # Sanity check this sync result. We should be banned from the room.
+        self.assertEqual(len(eve_sync_after_ban.joined), 0)
+        self.assertEqual(len(eve_sync_after_ban.archived), 1)
+        self.assertEqual(eve_sync_after_ban.archived[0].room_id, room_id)
```
clokep, line 288: Why would this even be in the user's sync response if they never joined the room? Doesn't this essentially leak bans to a user?

tests/handlers/test_sync.py
```diff
@@ -186,6 +187,139 @@ def test_unknown_room_version(self):
...
+
+        mocked_get_prev_events = patch.object(
```
clokep, line 290: I think the comment about mocking from above should go above this line.

tests/handlers/test_sync.py
```diff
@@ -186,6 +187,139 @@ def test_unknown_room_version(self):
...
+            self.hs.get_datastore(),
+            "get_prev_events_for_room",
+            new_callable=MagicMock,
+            return_value=make_awaitable([moderator_pl_event_id]),
+        )
+        with mocked_get_prev_events:
+            self.helper.join(room_id, eve, tok=eve_token)
+
+        # Eve makes a second, incremental sync. The join event should appear, but not
+        # under the "joined" section.
```
clokep, line 300: We don't seem to be asserting that the join event appears?

tests/handlers/test_sync.py
```diff
@@ -186,6 +187,139 @@ def test_unknown_room_version(self):
...
+                eve_sync_config,
+                since_token=since_token,
+            )
+        )
+        # The incremental sync should have nothing to show.
+        self.assertEqual(len(eve_incremental_sync_after_join.joined), 0)
+        self.assertEqual(len(eve_incremental_sync_after_join.archived), 0)
+
+        # If we did a third initial sync, we should _still_ see that eve is banned.
+        eve_initial_sync_after_join: SyncResult = self.get_success(
```
clokep, line 313: The type can't change, so maybe?

```suggestion
        eve_initial_sync_after_join = self.get_success(
```

tests/handlers/test_sync.py
```diff
@@ -186,6 +187,139 @@ def test_unknown_room_version(self):
...
+            self.sync_handler.wait_for_sync_for_user(
+                eve_requester,
+                eve_sync_config,
+            )
+        )
+        self.assertEqual(len(eve_initial_sync_after_join.joined), 0)
+        self.assertEqual(len(eve_initial_sync_after_join.archived), 1)
```
clokep, line 320: I'm confused at why this seems to flip in and out of archived, something here feels buggy, but maybe it is working as expected?

clokep (commented):
Looks plausible, but I have some questions!

tests/handlers/test_sync.py
DMRobertson: This one is coming from the moderator, who we're pretending is on a remote homeserver. But on reflection, I don't think it matters who sends the ban or where their homeserver is. (The scenario Rich describes involving federation is just one way that it could happen).

Maybe there's no need for the `moderator` user at all---maybe the `owner` can do the banning. I'll have a look at this tomorrow!

clokep: Yeah, the federation example might just be a likely way for this to happen, but perhaps something similar could happen with e.g. workers.

tests/handlers/test_sync.py
DMRobertson: I didn't expect this to show up either.

My recollection of "archived" from when I looked at the sync machinery before Christmas is that nobody really knows what it's meant to do our how it currently works.

> Doesn't this essentially leak bans to a user?

Maybe---I'd want to try this out in a simpler case without any fakery that injects invents though.

tests/handlers/test_sync.py
DMRobertson: I think this comment might be wrong (it certainly wasn't the behaviour I saw). I think I might need to double check the scenario with @richvdh here.

tests/handlers/test_sync.py
DMRobertson: This is a different variable: `initial` instead of `incremental`.

(In the ideal world, we'd annotate `get_success` so this isn't necessary; but I refrained from doing that here.)

tests/handlers/test_sync.py
DMRobertson: There are three syncs:

1. `eve_sync_after_ban`, an  initial sync,
2. `eve_incremental_sync_after_join`, which follows (1), and
3. `eve_initial_sync_after_join`.

(Maybe the naming is too verbose here?)

(1 + 2) and (3) are meant to cover the same range of events in the room. I think (1+2) and (3) are consistent.

> maybe it is working as expected?

maybe; again, I'm not sure we know what "expected" actually is.

tests/handlers/test_sync.py
clokep: Oh, my eyes deceived me. Carry on!

tests/handlers/test_sync.py
DMRobertson: > > I think (1+2) and (3) are consistent.
> 
> Consistent in that 2 shows no change since the state of the room has not changed according to eve?

Yes, exactly this.

tests/handlers/test_sync.py
clokep: Seems to be working as expected.

tests/rest/client/utils.py
```diff
@@ -196,6 +196,16 @@ def leave(self, room=None, user=None, expect_code=200, tok=None):
             expect_code=expect_code,
         )
 
+    def ban(self, room=None, src=None, targ=None, expect_code=200, tok=None):
```
clokep, line 199: Type hints since we're adding this please! (And a docstring.)

DMRobertson: Ah, I wanted to hold off on this and change all of these helpers in one go. They all lack type hints, all incorrectly allow a bunch of required arguments to default to None, and a docstring. We could do better here by using `functools.partialmethod`.

DMRobertson: > We could do better here by using `functools.partialmethod`.

Ahh, that's not strictly true: for "join" and "leave" the src and target are one and the same.

DMRobertson: And PyCharm doesn't have good support for partial/partialmethod. See e.g. https://youtrack.jetbrains.com/issue/PY-18448 or https://youtrack.jetbrains.com/issue/PY-37275

clokep: My thought was mostly that any new code should get type hints, even if we do intend to replace it in the future or something. Thanks for adding them!

richvdh (approved):
lgtm!

PR #11702, opened 2022-01-06
Optionally use an on-disk sqlite db in tests

DMRobertson:
When debugging a test it is sometimes useful to inspect the state of the
DB. This is not easy when the db is in-memory: one cannot attach the
sqlite CLI to another process's DB. (I suppose you could ferret out the
sqlite connection object from within the debugger and sending text
directly but... that is painful and doesn't have tab completion.)

With this change, if SYNAPSE_TEST_SQLITE_DB_LOCATION is set, we try to
write the DB to disk at that location (deleting any existing db file as
necessary).

Quick and dirty.

### DIFF
--- a/tests/server.py
+++ b/tests/server.py
@@ -14,6 +14,7 @@
 import hashlib
 import json
 import logging
+import os
 import time
 import uuid
 import warnings
@@ -71,6 +72,7 @@ from tests.utils import (
     POSTGRES_HOST,
     POSTGRES_PASSWORD,
     POSTGRES_USER,
+    SQLITE_TEST_DB_LOCATION,
     USE_POSTGRES_FOR_TESTS,
     MockClock,
     default_config,
@@ -739,9 +741,16 @@ def setup_test_homeserver(
             },
         }
     else:
+        if SQLITE_TEST_DB_LOCATION != ":memory:":
+            # nuke the DB on disk
+            try:
+                os.remove(SQLITE_TEST_DB_LOCATION)
+            except FileNotFoundError:
+                pass
+
         database_config = {
             "name": "sqlite3",
-            "args": {"database": ":memory:", "cp_min": 1, "cp_max": 1},
+            "args": {"database": SQLITE_TEST_DB_LOCATION, "cp_min": 1, "cp_max": 1},
         }
 
     if "db_txn_limit" in kwargs:
--- a/tests/server.py
+++ b/tests/server.py
@@ -745,9 +745,7 @@ def setup_test_homeserver(
         if SQLITE_PERSIST_DB:
             # The current working directory is in _trial_temp, so this gets created within that directory.
             test_db_location = os.path.abspath("test.db")
-            logger.debug(
-                "Will persist db to %s", test_db_location
-            )
+            logger.debug("Will persist db to %s", test_db_location)
             # Ensure each test gets a clean database.
             try:
                 os.remove(test_db_location)
--- a/tests/server.py
+++ b/tests/server.py
@@ -747,7 +747,7 @@ def setup_test_homeserver(
             logger.debug(
                 "Will persist db to %s", test_db_location
             )
-            # nuke the DB on disk if it already exists
+            # Ensure each test gets a clean database.
             try:
                 os.remove(test_db_location)
             except FileNotFoundError:
--- a/tests/server.py
+++ b/tests/server.py
@@ -743,6 +743,7 @@ def setup_test_homeserver(
         }
     else:
         if SQLITE_PERSIST_DB:
+            # The current working directory is in _trial_temp, so this gets created within that directory.
             test_db_location = os.path.abspath("test.db")
             logger.debug(
                 "Will persist db to %s", test_db_location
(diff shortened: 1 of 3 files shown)


tests/utils.py
```diff
@@ -42,6 +42,10 @@
 POSTGRES_PASSWORD = os.environ.get("SYNAPSE_POSTGRES_PASSWORD", None)
 POSTGRES_BASE_DB = "_synapse_unit_tests_base_%s" % (os.getpid(),)
 
+# When debugging a specific test, it's occasionally useful to write the
+# DB to /tmp and query it with the sqlite CLI.
+SQLITE_TEST_DB_LOCATION = os.environ.get("SYNAPSE_TEST_SQLITE_DB_LOCATION", ":memory:")
```
clokep, line 47: I debated if this should default to `None` so we could encapsulate the magic `":memory:"` string in a single spot, what do you think?

tests/server.py
```diff
@@ -739,9 +741,16 @@ def setup_test_homeserver(
             },
         }
     else:
+        if SQLITE_TEST_DB_LOCATION != ":memory:":
+            # nuke the DB on disk
+            try:
+                os.remove(SQLITE_TEST_DB_LOCATION)
+            except FileNotFoundError:
+                pass
```
clokep, line 749: This feels a little icky, it seems for psql we generate a unique db name for each test. I'm not sure that's really doable here though.

Would it make more sense to not provide a location but just provide that, yes, you want to use an on-disk database and we put it in a known location inside of `_trial_temp`?

DMRobertson: > This feels a little icky,

agreed---icky, but hopefully useful? I'm mostly intending for this to be used when debugging one specific test interactively to try.

> we put it in a known location inside of _trial_temp

That could work---though I think if you have multiple instances of `trial` running then you end up with `_trial_temp`, `_trial_temp-1`, ... . (This occasionally comes up when I'm debugging one test in PyCharm while running another---the two modes are independent.)

clokep: I meant "wherever trial thinks the root of my current test running is", not `_trial_temp` specifically, sorry for not being clear!

DMRobertson: roger that. I was worried at first that this would be hard to find, but it appears to just be the CWD when we run the tests.

tests/server.py
```diff
@@ -739,9 +742,24 @@ def setup_test_homeserver(
             },
         }
     else:
+        if SQLITE_PERSIST_DB:
+            test_db_location = os.path.abspath("test.db")
+            logger.debug(
+                "Will persist db to %s",
+            )
+            # nuke the DB on disk if it already exists
```
clokep, line 750: Will it ever exist? Doesn't the temp directory get fully destroyed before a run?

DMRobertson: My understanding: the directory gets destroyed before `trial` starts to run its tests, but not between test runs`; I expect we want each test to run on its own DB from scratch.

clokep: Ah-ha! Right. 👍

tests/server.py
```diff
@@ -739,9 +742,24 @@ def setup_test_homeserver(
             },
         }
     else:
+        if SQLITE_PERSIST_DB:
+            test_db_location = os.path.abspath("test.db")
+            logger.debug(
+                "Will persist db to %s",
+            )
```
clokep, line 749:
```suggestion
            logger.debug(
                "Will persist db to %s", test_db_location
            )
```

tests/server.py
clokep:
```suggestion
            # Ensure each test gets a clean database.
```

tests/server.py
```diff
@@ -739,9 +742,24 @@ def setup_test_homeserver(
             },
         }
     else:
+        if SQLITE_PERSIST_DB:
+            test_db_location = os.path.abspath("test.db")
```
clokep, line 746: I think this needs something saying it gets into `_trial_temp`?

```suggestion
            # The current working directory is in _trial_temp, so this gets created within that directory.
            test_db_location = os.path.abspath("test.db")
```

clokep (approved):
LGTM if CI passes!

tests/utils.py
```diff
@@ -42,6 +42,10 @@
 POSTGRES_PASSWORD = os.environ.get("SYNAPSE_POSTGRES_PASSWORD", None)
 POSTGRES_BASE_DB = "_synapse_unit_tests_base_%s" % (os.getpid(),)
 
+# When debugging a specific test, it's occasionally useful to write the
+# DB to disk and query it with the sqlite CLI.
+SQLITE_PERSIST_DB = os.environ.get("SYNAPSE_TEST_PERSIST_SQLITE_DB") is not None
```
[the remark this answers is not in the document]
DMRobertson: Why `str(...)` --- are the values bytes by default?

Come to think of it, maybe I could do `"VARNAME" in os.environ`, since I'm only interested in if it's set or not. I'll probably leave it be for now though.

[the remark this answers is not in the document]
clokep: > Why `str(...)` --- are the values bytes by default?

It makes `None` -> `"None"` 😄 

This is only really useful if you want to ensure that `true`-ish string values are used.

PR #11707, opened 2022-01-07
Prettier complement logs

kegsay:
Modifies Github Actions to make Complement logs prettier.

E.g https://github.com/matrix-org/synapse/runs/4740778609?check_suite_focus=true#step:7:18


DMRobertson (approved):
Looks great! Guessing this works because `gotestfmt` is now baked into the complement image?

kegsay:
Yep, spot on, I've updated the image to include `gotestfmt` (so there's no hit to run times).

PR #11714, opened 2022-01-10
Deal with mypy errors w/ type-hinted pynacl 1.5.0

### LINKED ISSUE #11644: Pull in type hints for pynacl #478

DMRobertson:
Fixes #11644.

I really don't like that we're monkey patching pynacl SignedKey
instances with alg and version objects. But I'm too scared to make the
changes necessary right now.

(Ideally I would replace `signedjson.types.SingingKey` with a runtime class which
wraps or inherits from `nacl.signing.SigningKey`.)

### DIFF
--- a/tests/crypto/test_event_signing.py
+++ b/tests/crypto/test_event_signing.py
@@ -36,6 +36,9 @@ HOSTNAME = "domain"
 
 class EventSigningTestCase(unittest.TestCase):
     def setUp(self):
+        # NB: `signedjson` expects `nacl.signing.SigningKey` instances which have been
+        # monkeypatched to include new `alg` and `version` attributes. This is captured
+        # by the `signedjson.types.SigningKey` protocol.
         self.signing_key: signedjson.types.SigningKey = nacl.signing.SigningKey(
             SIGNING_KEY_SEED
         )
--- a/tests/crypto/test_event_signing.py
+++ b/tests/crypto/test_event_signing.py
@@ -14,6 +14,7 @@
 
 
 import nacl.signing
+import signedjson.types
 from unpaddedbase64 import decode_base64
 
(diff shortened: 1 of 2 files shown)


tests/crypto/test_event_signing.py
```diff
@@ -35,7 +36,9 @@
 
 class EventSigningTestCase(unittest.TestCase):
     def setUp(self):
-        self.signing_key = nacl.signing.SigningKey(SIGNING_KEY_SEED)
+        self.signing_key: signedjson.types.SigningKey = nacl.signing.SigningKey(
```
clokep, line 39: Maybe add a comment above this saying what's happening.

clokep (approved):
Not the most elegant fix, but looks fine if build passes!

tests/crypto/test_event_signing.py
DMRobertson: Good idea, thanks. What do you make of https://github.com/matrix-org/synapse/pull/11714/commits/683cccbb61c41816d144928ab03d4b1b5bfe83a7 ?

clokep: Seems reasonable!

DMRobertson: Cheers. Let's get this merged to make everyone's CI happy.

PR #11715, opened 2022-01-10
Document the `SYNAPSE_TEST_PERSIST_SQLITE_DB` unit test env var

anoadragon453:
Adds some documentation to the contributing guide for the awesome new `SYNAPSE_TEST_PERSIST_SQLITE_DB` environment variable added in #11702.


docs/development/contributing_guide.md
```diff
@@ -169,6 +169,21 @@ To increase the log level for the tests, set `SYNAPSE_TEST_LOG_LEVEL`:
 SYNAPSE_TEST_LOG_LEVEL=DEBUG trial tests
 ```
 
+By default, tests will use an in-memory SQLite database for test data. For additional
+help with debugging, one can use an on-disk SQLite database file instead, in order to
+review database state during and after running tests. This can be adjusted using the
+`SYNAPSE_TEST_PERSIST_SQLITE_DB` environment variable:
+
+```sh
+SYNAPSE_TEST_PERSIST_SQLITE_DB=test.db trial tests
```
DMRobertson, line 178: We don't actually read the value of this variable, just the fact that it's been set. The path`test.db` is hard-coded for now.

```suggestion
SYNAPSE_TEST_PERSIST_SQLITE_DB=1 trial tests
```

docs/development/contributing_guide.md
```diff
@@ -169,6 +169,21 @@ To increase the log level for the tests, set `SYNAPSE_TEST_LOG_LEVEL`:
...
+```
+
+The database file can then be inspected with:
+
+```sh
+sqlite3 test.db
```
DMRobertson, line 184:
```suggestion
sqlite3 `_trial_temp` test.db
```
Probably not worth mentioning `_trial_temp-1`, `_trial_temp-2` etc here.

docs/development/contributing_guide.md
```diff
@@ -169,6 +169,21 @@ To increase the log level for the tests, set `SYNAPSE_TEST_LOG_LEVEL`:
 SYNAPSE_TEST_LOG_LEVEL=DEBUG trial tests
 ```
 
+By default, tests will use an in-memory SQLite database for test data. For additional
+help with debugging, one can use an on-disk SQLite database file instead, in order to
+review database state during and after running tests. This can be adjusted using the
```
DMRobertson, line 174: Can we explicitly mention that this database gets nuked at the start of an individual test? I.e. you only see the DB state for the current test, or the last test to finish.

DMRobertson (changes requested):

docs/development/contributing_guide.md
anoadragon453: I didn't take the suggestion exactly here, as it looks like it may have been a bit of a typo? Great point though!

DMRobertson: Ahh yes, confusing my markdown and shell. Thanks!

DMRobertson (approved):
LGTM! (Have you had a chance to try this out? I'm hoping that it doesn't just Work On My Machine)

anoadragon453:
@DMRobertson Indeed it does! And it's a really welcome change. Thank you for doing it :slightly_smiling_face:

PR #11716, opened 2022-01-10
Fix docstring on `add_account_data_for_user`.

reivilibre:
It looks like a copy and paste mistake: the previous function has the same docstring and *is* intended for room account data.


anoadragon453 (approved):
Good spot!

Issue #9813, opened 2021-04-14
Enable automatic garbage collection


ShadowJonathan:
cc @erikjohnston, regarding the allocation mysteries where python is taking up more memory than it's actually using for objects.

According to @richvdh, the only reason why we're currently using manual GC is because python 2 didn't support [`gc.callbacks`](https://docs.python.org/3/library/gc.html#gc.callbacks), using this to catch metrics instead might slightly improve performance elsewhere, such as in memory pressure and CPU usage, as we're giving cPython the reigns of the garbage collector again.

This might need some performance testing.

erikjohnston:
I think another reason is that we want to ensure that the CPU used doing the GC isn't reported as part of a log context? By triggering it manually we can ensure that it happens in the sentinel log context, though that may be possible with the callbacks too.

Given the reactor ticks around (hopefully) more frequently than we do GCs, I don't think it makes any difference.

clokep:
I re-worded this a bit. The task is to stop using manual GC and to switch to the automatic GC. The way to do this is by using `gc.callbacks`.

erikjohnston:
I'm not sure why we'd want to bother? I don't see what automatic GC will buy us?

clokep:
Hah, well that's why I reworded. Is the task here to switch to automatic GC or to switch to using `gc.callbacks`? 😄 B is a pre-req for A, but doesn't necessarily imply that we should be doing A.

richvdh:
#9902 added a `gc_min_seconds_between` config setting, which requires GC to be manual, so I don't think this is something we want to do.

PR #11723, opened 2022-01-11
Simplify GC prometheus metrics

richvdh:
Rather than hooking into the reactor loop, just add a timed task that runs every 100 ms to do the garbage collection.

Part 1 of a quest to simplify the reactor monkey-patching.

### DIFF
--- a/synapse/metrics/_gc.py
+++ b/synapse/metrics/_gc.py
@@ -1,0 +1,195 @@
+# Copyright 2015-2022 The Matrix.org Foundation C.I.C.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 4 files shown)


synapse/metrics/_gc.py
```diff
@@ -81,6 +83,42 @@ def collect(self) -> Iterable[Metric]:
 if not running_on_pypy:
     REGISTRY.register(GCCounts())
 
+    # disable automatic GC, and replace it with a task that runs every 100ms, so that
+    # we can get some metrics about GC activity.
```
clokep, line 87: If this is just about gathering metrics, could we use the [`gc.callbacks` mechanism](https://docs.python.org/3/library/gc.html#gc.callbacks) instead and not disable automatic gc?

richvdh: per https://github.com/matrix-org/synapse/issues/9813#issuecomment-1009175679: I don't think so.

clokep: Thanks for answering the same question again. 😢

squahtx (approved):
lgtm

PR #11724, opened 2022-01-11
Improve `reactor_tick_time` metric

richvdh:
The existing implementation of the `python_twisted_reactor_tick_time` metric is pretty useless, because it *only* measures the time taken to execute timed calls and callbacks from threads. That neglects everything that happens off the back of I/O, which is obviously quite a lot for us.

To improve this, I've hooked into a different place in the reactor - in particular, where it calls `epoll`. That call is the only place it should wait for something to happen - the rest of the loop *should* be quick.

I've also removed `python_twisted_reactor_pending_calls`, because I don't believe anyone ever looks at it, and it's a nuisance to populate.


synapse/metrics/__init__.py
```diff
@@ -377,11 +377,6 @@ def collect(self) -> Iterable[Metric]:
     "Tick time of the Twisted reactor (sec)",
     buckets=[0.001, 0.002, 0.005, 0.01, 0.025, 0.05, 0.1, 0.2, 0.5, 1, 2, 5],
 )
-pending_calls_metric = Histogram(
-    "python_twisted_reactor_pending_calls",
-    "Pending calls",
-    buckets=[1, 2, 5, 10, 25, 50, 100, 250, 500, 1000],
-)
```
squahtx, line 501: `contrib/prometheus/consoles/synapse.html` still mentions this metric (https://github.com/matrix-org/synapse/blob/release-v1.49/contrib/prometheus/consoles/synapse.html#L100)

Also `docs/metrics-howto.md` could do with an update.

synapse/metrics/_reactor_metrics.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+class EpollWrapper:
+    """a wrapper for an epoll object which records the time between polls"""
+
+    def __init__(self, poller: "select.epoll"):
+        self.last_polled = time.time()
+        self._poller = poller
+
+    def poll(self, *args, **kwargs) -> List[Tuple[int, int]]:  # type: ignore[no-untyped-def]
```
squahtx, line 42: I'm not too keen on the types here, but the way it's done currently does have the benefit of forward and backward compatibility.

Just a comment, not suggesting that it needs to be changed.

synapse/metrics/_reactor_metrics.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+try:
+    # if the reactor has a `_poller` attribute, which is an `epoll` object
+    # (ie, it's an EPollReactor), we wrap the `epoll` with a thing that will
+    # measure the time between ticks
+    from select import epoll
+
+    poller = reactor._poller  # type: ignore[attr-defined]
+except (AttributeError, ImportError):
```
squahtx, line 77: When would the reactor not be epoll-based?

synapse/metrics/__init__.py
richvdh: > `contrib/prometheus/consoles/synapse.html` still mentions this metric (https://github.com/matrix-org/synapse/blob/release-v1.49/contrib/prometheus/consoles/synapse.html#L100)

mmhmm. I don't think that thing's really maintained, tbh, which is why it's in `contrib`. Still, I've removed this.

> Also `docs/metrics-howto.md` could do with an update.

What do you suggest? it's only mentioned under "things that changed in 0.31.0". Or are you suggesting I document the metric's removal there? (I'm not sure that would have much value, but ymmv)

synapse/metrics/_reactor_metrics.py
richvdh: whenever `select.epoll` doesn't exist, which basically means we're not running on Linux: https://docs.python.org/3/library/select.html#select.epoll

clokep: You can see how the reactor is chosen at: https://github.com/twisted/twisted/blob/twisted-20.3.0/src/twisted/internet/default.py#L41-L53 (there's a handful of others you can install manually too...)

Would it make more sense to check if the reactor is. Both `PollReactor` and `EPollReactor` have a `_poller` attribute, but obviously only one is of type `epoll`.

synapse/metrics/__init__.py
squahtx: I was thinking of documenting the removal there. But if that's not a thing we usually do, that's fine.

synapse/metrics/_reactor_metrics.py
squahtx: There's an `isinstance(poller, epoll)` check below which I think does what you're asking?

squahtx (approved):

synapse/metrics/_reactor_metrics.py
clokep: > There's an `isinstance(poller, epoll)` check below which I think does what you're asking?

Errr, yes the code is doing what I thought. I have no idea what the second sentence was trying to ask at the start. 👍 (I've edited it...)

richvdh:
We deployed this to matrix.org back in January:

image

Issue #11719, opened 2022-01-10
`AssertionError: get_device_updates_by_remote returned too many EDUs` causing federation problems


richvdh:
We're seeing quite a lot of these, and they seem to be interrupting federation traffic:

```
2022-01-10 10:03:58,307 - synapse.federation.sender.per_destination_queue - 364 - ERROR - federation_transaction_transmission_loop-9933549 - TX [....] Failed to send transaction
Capture point (most recent call last):
  File "/usr/local/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/local/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/synapse/src/synapse/app/federation_sender.py", line 22, in <module>
    start(sys.argv[1:])
  File "/home/synapse/src/synapse/app/generic_worker.py", line 505, in start
    _base.start_worker_reactor("synapse-generic-worker", config)
  File "/home/synapse/src/synapse/app/_base.py", line 126, in start_worker_reactor
    run_command=run_command,
  File "/home/synapse/src/synapse/app/_base.py", line 178, in start_reactor
    run()
  File "/home/synapse/src/synapse/app/_base.py", line 162, in run
    run_command()
  File "/home/synapse/env-py37/lib/python3.7/site-packages/twisted/internet/base.py", line 1318, in run
    self.mainLoop()
  File "/home/synapse/env-py37/lib/python3.7/site-packages/twisted/internet/base.py", line 1328, in mainLoop
    reactorBaseSelf.runUntilCurrent()
  File "/home/synapse/src/synapse/metrics/__init__.py", line 645, in f
    ret = func(*args, **kwargs)
  File "/home/synapse/env-py37/lib/python3.7/site-packages/twisted/internet/base.py", line 967, in runUntilCurrent
    f(*a, **kw)
  File "/home/synapse/env-py37/lib/python3.7/site-packages/twisted/internet/defer.py", line 662, in callback
    self._startRunCallbacks(result)
  File "/home/synapse/env-py37/lib/python3.7/site-packages/twisted/internet/defer.py", line 764, in _startRunCallbacks
    self._runCallbacks()
  File "/home/synapse/env-py37/lib/python3.7/site-packages/twisted/internet/defer.py", line 859, in _runCallbacks
    current.result, *args, **kwargs
[...]

PR #11729, opened 2022-01-12
Fix a bug introduced in Synapse v1.0.0 whereby device list updates would not be sent to remote homeservers if there were too many to send at once.

### LINKED ISSUE #11727: Device list updates to be sent over federation will be skipped if there are too many at one time
`get_device_updates_by_remote` always returns the next stream position as the latest device stream token, even in cases where it is limited to a maximum number of rows to return.
The effect of this is that any device updates that don't fit within one `limit`-sized batch will be skipped.

Discovered whilst looking into #11719. Due to the nature of my fix to #11719, I'll likely fix them both at the same time.

reivilibre:
Fixes #11727: you can't resume reading the stream if there are more than `limit` entries because you get told to continue from the end of the stream, not the end of where you read up to.

Of note: #7010: stream_ids used to not be unique, but the ones produced in modern times are (since v1.13.0). Thanks Erik for saving me a lot of pain!

It appears to have been introduced in https://github.com/matrix-org/synapse/commit/2d1d7b7e6f2bec3b96b0d23993369ce46aad4f32, which has been there since the good ol' days of v1.0.0 :).

Required change for the fix to #11719, which is a release blocker.

### DIFF
--- a/changelog.d/11729.bugfix
+++ b/changelog.d/11729.bugfix
@@ -1,0 +1,1 @@
+Fix a bug introduced in Synapse v1.0.0 whereby device list updates would not be sent to remote homeservers if there were too many to send at once.
--- a/changelog.d/11729.bugfix
+++ b/changelog.d/11729.bugfix
@@ -1,1 +1,1 @@
-Fix a bug introduced in Synapse v1.0.0 whereby device list updates would not be sent to remote homeservers if there were too many to send at once.
+Fix a bug introduced in Synapse v1.0.0 whereby some device list updates would not be sent to remote homeservers if there were too many to send at once.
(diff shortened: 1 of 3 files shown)


changelog.d/11729.bugfix
```diff
@@ -0,0 +1 @@
+Fix a bug introduced in Synapse v1.0.0 whereby device list updates would not be sent to remote homeservers if there were too many to send at once.
```
babolivier, line 1:
```suggestion
Fix a bug introduced in Synapse v1.0.0 whereby some device list updates would not be sent to remote homeservers if there were too many to send at once.
```

babolivier (approved):

DMRobertson:
Ahh, this is targeted at 1.50, which doesn't have the CI fix.

PR #11730, opened 2022-01-12
Fix a bug introduced in Synapse v1.50.0rc1 whereby outbound federation could fail because too many EDUs were produced for device updates.

### LINKED ISSUE #11719: `AssertionError: get_device_updates_by_remote returned too many EDUs` causing federation problems

reivilibre:
Follows on from: #11729.

Fixes #11719.

Basically, sending both a prefixed and unprefixed copy of cross-signing key updates meant that `get_device_updates_by_remote` was blowing the budget and producing more EDUs than the limit it was given.
Working around this is a little bit ugly but the principle is simple: if there won't be room for a device update, stop early!

The ugliness arises because depending on the row that comes back from the database (and whether it's a duplicate of another or not), there could be 0, 1 or 2 EDUs produced.

This can all be stripped out once we remove the duplication of cross-signing key updates.

### DIFF
--- a/changelog.d/11730.bugfix
+++ b/changelog.d/11730.bugfix
@@ -1,0 +1,1 @@
+Fix a bug introduced in Synapse v1.50.0rc1 whereby outbound federation could fail because too many EDUs were produced for device updates.
--- a/synapse/storage/databases/main/devices.py
+++ b/synapse/storage/databases/main/devices.py
@@ -191,7 +191,7 @@ class DeviceWorkerStore(SQLBaseStore):
     @trace
     async def get_device_updates_by_remote(
         self, destination: str, from_stream_id: int, limit: int
-    ) -> Tuple[int, List[Tuple[str, dict]]]:
+    ) -> Tuple[int, List[Tuple[str, JsonDict]]]:
         """Get a stream of device updates to send to the given remote server.
 
         Args:
--- a/synapse/storage/databases/main/devices.py
+++ b/synapse/storage/databases/main/devices.py
@@ -378,6 +378,9 @@ class DeviceWorkerStore(SQLBaseStore):
         """
         devices = (
             await self.get_e2e_device_keys_and_signatures(
+                # Because these are (user_id, device_id) tuples with all
+                # device_ids not being None, the returned list's length will not
+                # exceed that of query_map.
                 query_map.keys(),
                 include_all_devices=True,
                 include_deleted_devices=True,
--- a/synapse/storage/databases/main/devices.py
+++ b/synapse/storage/databases/main/devices.py
@@ -280,6 +280,15 @@ class DeviceWorkerStore(SQLBaseStore):
         query_map = {}
         cross_signing_keys_by_user = {}
         for user_id, device_id, update_stream_id, update_context in updates:
+            is_master_key_update = (
+                user_id in master_key_by_user
+                and device_id == master_key_by_user[user_id]["device_id"]
+            )
+            is_self_signing_key_update = (
+                user_id in self_signing_key_by_user
+                and device_id == self_signing_key_by_user[user_id]["device_id"]
+            )
+
             if (
                 user_id in master_key_by_user
                 and device_id == master_key_by_user[user_id]["device_id"]
--- a/synapse/storage/databases/main/devices.py
+++ b/synapse/storage/databases/main/devices.py
@@ -222,6 +222,8 @@ class DeviceWorkerStore(SQLBaseStore):
             limit,
         )
 
+        # len(updates) <= limit.
+
         # Return an empty list if there are no updates
         if not updates:
             return now_stream_id, []
@@ -302,11 +304,17 @@ class DeviceWorkerStore(SQLBaseStore):
 
             last_processed_stream_id = update_stream_id
 
+        # len(query_map) + len(cross_signing_keys_by_user) <= len(updates) here,
+        # so len(query_map) + len(cross_signing_keys_by_user) <= limit.
+
         results = await self._get_device_update_edus_by_remote(
             destination, from_stream_id, query_map
         )
 
-        # add the updated cross-signing keys to the results list
+        # len(results) <= len(query_map) here,
+        # so len(results) + len(cross_signing_keys_by_user) <= limit.
+
+        # Add the updated cross-signing keys to the results list
         for user_id, result in cross_signing_keys_by_user.items():
             result["user_id"] = user_id
             results.append(("m.signing_key_update", result))
--- a/synapse/storage/databases/main/devices.py
+++ b/synapse/storage/databases/main/devices.py
@@ -366,12 +366,15 @@ class DeviceWorkerStore(SQLBaseStore):
         Args:
             destination: The host the device updates are intended for
             from_stream_id: The minimum stream_id to filter updates by, exclusive
-            query_map (Dict[(str, str): (int, str|None)]): Dictionary mapping
-                user_id/device_id to update stream_id and the relevant json-encoded
-                opentracing context
+            query_map: Dictionary mapping (user_id, device_id) to
+                (update stream_id, the relevant json-encoded opentracing context)
 
         Returns:
-            List of objects representing an device update EDU
+            List of objects representing a device update EDU.
+
+        Postconditions:
+            The returned list has a length not exceeding that of the query_map:
+                len(result) <= len(query_map)
         """
         devices = (
             await self.get_e2e_device_keys_and_signatures(
--- a/synapse/storage/databases/main/devices.py
+++ b/synapse/storage/databases/main/devices.py
@@ -332,6 +332,11 @@ class DeviceWorkerStore(SQLBaseStore):
                 previous_update_stream_id, _ = query_map.get(key, (0, None))
 
                 if update_stream_id > previous_update_stream_id:
+                    # FIXME If this overwrites an older update, this discards the
+                    #  previous OpenTracing context.
+                    #  It might make it harder to track down issues using OpenTracing.
+                    #  If there's a good reason why it doesn't matter, a comment here
+                    #  about that would not hurt.
                     query_map[key] = (update_stream_id, update_context)
 
             # As this update has been added to the response, advance the stream
--- a/synapse/storage/databases/main/devices.py
+++ b/synapse/storage/databases/main/devices.py
@@ -280,6 +280,17 @@ class DeviceWorkerStore(SQLBaseStore):
         query_map = {}
         cross_signing_keys_by_user = {}
         for user_id, device_id, update_stream_id, update_context in updates:
+            # Calculate the remaining length budget.
+            # Note that, for now, each entry in `cross_signing_keys_by_user`
+            # gives rise to two device updates in the result, so those cost twice
+            # as much (and are the whole reason we need to separatly calculate
+            # the budget; we know len(updates) <= limit otherwise!)
+            # N.B. len() on dicts is cheap since they store their size.
+            remaining_length_budget = limit - (
+                len(query_map) + 2 * len(cross_signing_keys_by_user)
+            )
+            assert remaining_length_budget >= 0
+
             is_master_key_update = (
                 user_id in master_key_by_user
                 and device_id == master_key_by_user[user_id]["device_id"]
--- a/synapse/storage/databases/main/devices.py
+++ b/synapse/storage/databases/main/devices.py
@@ -200,9 +200,10 @@ class DeviceWorkerStore(SQLBaseStore):
             limit: Maximum number of device updates to return
 
         Returns:
-            A mapping from the  current stream id (ie, the stream id of the last
-            update included in the response), and the list of updates, where
-            each update is a pair of EDU type and EDU contents.
+            - The current stream id (i.e. the stream id of the last update included
+              in the response); and
+            - The list of updates, where each update is a pair of EDU type and
+              EDU contents.
         """
         now_stream_id = self.get_device_stream_token()
 
--- a/synapse/storage/databases/main/devices.py
+++ b/synapse/storage/databases/main/devices.py
@@ -351,6 +351,7 @@ class DeviceWorkerStore(SQLBaseStore):
             results.append(("m.signing_key_update", result))
             # also send the unstable version
             # FIXME: remove this when enough servers have upgraded
+            #        and remove the length budgeting above.
             results.append(("org.matrix.signing_key_update", result))
 
         return last_processed_stream_id, results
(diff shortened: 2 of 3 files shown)


synapse/storage/databases/main/devices.py
```diff
@@ -200,9 +200,10 @@ async def get_device_updates_by_remote(
             limit: Maximum number of device updates to return
 
         Returns:
-            A mapping from the  current stream id (ie, the stream id of the last
-            update included in the response), and the list of updates, where
-            each update is a pair of EDU type and EDU contents.
+            - The current stream id (i.e. the stream id of the last update included
+              in the response); and
+            - The list of updates, where each update is a pair of EDU type and
+              EDU contents.
```
DMRobertson, line 206: Maybe the `dict` in the type hint should be a `JSONDict`?

synapse/storage/databases/main/devices.py
```diff
@@ -221,6 +222,8 @@ async def get_device_updates_by_remote(
             limit,
         )
 
+        # len(updates) <= limit.
```
DMRobertson, line 225: What's going on here? Oh, you're tracking an upper bound on the size of `updates`.

Maybe this?

```suggestion
        # We need to ensure `updates` doesn't grow too big. 
        # Currently: `len(updates) <= limit`.
```

synapse/storage/databases/main/devices.py
```diff
@@ -302,11 +304,17 @@ async def get_device_updates_by_remote(
 
             last_processed_stream_id = update_stream_id
 
+        # len(query_map) + len(cross_signing_keys_by_user) <= len(updates) here,
```
DMRobertson, line 338: Agreed. In the worst case scenario, each update is for a distinct user and is added to either query+map or cross_signing_keys_by_user not both.

synapse/storage/databases/main/devices.py
```diff
@@ -302,11 +304,17 @@ async def get_device_updates_by_remote(
...
+        # so len(query_map) + len(cross_signing_keys_by_user) <= limit.
+
         results = await self._get_device_update_edus_by_remote(
             destination, from_stream_id, query_map
         )
 
-        # add the updated cross-signing keys to the results list
+        # len(results) <= len(query_map) here,
```
DMRobertson, line 345: Agreed, but I had to look at your comment on _get_device_update_edus_by_remote for this

synapse/storage/databases/main/devices.py
```diff
@@ -302,11 +304,17 @@ async def get_device_updates_by_remote(
...
+        # so len(results) + len(cross_signing_keys_by_user) <= limit.
```
DMRobertson, line 346: What does "HL-style" mean in the description? Haskell?

synapse/storage/databases/main/devices.py
```diff
@@ -342,7 +342,11 @@ def _get_device_updates_by_remote_txn(
             limit: Maximum number of device updates to return
 
         Returns:
-            List: List of device updates
+            List: List of device update tuples:
+                - user_id
+                - device_id
+                - stream_id
+                - opentracing_context
```
DMRobertson, line 381: Could be an attrs/typing.NamedTuple/dataclass here in the future?

synapse/storage/databases/main/devices.py
```diff
@@ -280,6 +280,17 @@ async def get_device_updates_by_remote(
         query_map = {}
         cross_signing_keys_by_user = {}
         for user_id, device_id, update_stream_id, update_context in updates:
+            # Calculate the remaining length budget.
+            # Note that, for now, each entry in `cross_signing_keys_by_user`
+            # gives rise to two device updates in the result, so those cost twice
+            # as much (and are the whole reason we need to separatly calculate
```
DMRobertson, line 286:
```suggestion
            # as much (and are the whole reason we need to separately calculate
```

changelog.d/11730.bugfix
```diff
@@ -0,0 +1 @@
+Fix a bug introduced in Synapse v1.50.0rc1 whereby outbound federation could fail because too many EDUs were produced for device updates.
```
DMRobertson, line 1: Does this knobble all outbound federation or just some?

tests/storage/test_devices.py
```diff
@@ -155,6 +155,99 @@ def test_get_device_updates_by_remote_can_limit_properly(self):
         # Check the newly-added device_ids are contained within these updates
         self._check_devices_in_updates(device_ids, device_updates)
 
+    def test_get_device_updates_by_remote_cross_signing_key_updates(
+        self,
+    ) -> None:
+        """
+        Tests that `get_device_updates_by_remote` limits the length of the return properly
+        when cross-signing key updates are present.
```
DMRobertson, line 163: Maybe?

```suggestion
        Tests that `get_device_updates_by_remote` limits the length of the return value
        properly when cross-signing key updates are present.
```

tests/storage/test_devices.py
```diff
@@ -155,6 +155,99 @@ def test_get_device_updates_by_remote_can_limit_properly(self):
...
+        )
+
+        # Get device updates meant for this remote
+        next_stream_id, device_updates = self.get_success(
+            self.store.get_device_updates_by_remote("somehost", -1, limit=3)
+        )
+
+        # Even though the limit has not been hit, the next update will be duplicated
+        # but there's no space for it here.
+        self.assertEqual(len(device_updates), 2, device_updates)
```
DMRobertson, line 228: Can you spell out for me which updates we're expecting here. I.e. what would `device_updates` be with a higher limit, and what is it effectively truncated to?

tests/storage/test_devices.py
```diff
@@ -155,6 +155,99 @@ def test_get_device_updates_by_remote_can_limit_properly(self):
...
+        # one, but the cross-signing key update is emitted twice, once with an unprefixed
+        # type and one with an unstable-prefixed type)
+        # (This is a temporary arrangement for backwards compatibility!)
+        self.assertEqual(len(device_updates), 2, device_updates)
+        self.assertEqual(
+            device_updates[0][0], "m.signing_key_update", device_updates[0]
+        )
+        self.assertEqual(
+            device_updates[1][0], "org.matrix.signing_key_update", device_updates[1]
+        )
```
DMRobertson, line 249: Perhaps also check there are no more updates remaining after this?

DMRobertson (commented):
Seems sensible; a few questions.

There's a part of me that wonders if it'd've been easier to pass in `limit//2` instead of `limit` to `get_device_updates_by_remote`. But I think it's better that `get_device_updates_by_remote` obeys the limit you give to it.

reivilibre:
> There's a part of me that wonders if it'd've been easier to pass in limit//2 instead of limit to get_device_updates_by_remote.

Oh trust me I was tempted :D... but then you're potentially making twice as many federation requests for foolish reasons and I didn't like the idea of causing that :)

synapse/storage/databases/main/devices.py
reivilibre: Hoare Logic, but it's totally irrelevant since I didn't actually go to that much effort to prove it :P

DMRobertson: time to rewrite synapse in prolog

synapse/storage/databases/main/devices.py
reivilibre: It appears to be a JSON encoded thing after looking at some other code, but I feel more comfortable leaving it as an opaque clob for the time being

changelog.d/11730.bugfix
reivilibre: Good question. It would potentially only be to some destinations, but not clear if that meant federation was totally scuppered for that destination or only some transactions...

DMRobertson: Makes sense. Let's leave it as it is.

DMRobertson:
Think this needs a lint, I might have accidentally gifted you a trailling space. Sorry.

tests/storage/test_devices.py
reivilibre: Added some words, let me know what you think

synapse/storage/databases/main/devices.py
```diff
@@ -300,6 +300,19 @@ async def get_device_updates_by_remote(
...
+            )
+
+            if (
+                is_cross_signing_key_update
+                and user_id not in cross_signing_keys_by_user
+            ):
+                # This will give rise to 2 device updates.
+                # If we don't have the budget, stop here!
+                if remaining_length_budget < 2:
+                    break
```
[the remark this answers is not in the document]
reivilibre: We only advance the stream position as far as the loop goes (see a little bit down). I added a comment to it to make it more noticeable.

synapse/storage/databases/main/devices.py
reivilibre: Do you think I should add some words?

tests/storage/test_devices.py
```diff
@@ -223,11 +229,16 @@ def test_get_device_updates_by_remote_cross_signing_key_updates(
             self.store.get_device_updates_by_remote("somehost", -1, limit=3)
         )
 
-        # Even though the limit has not been hit, the next update will be duplicated
-        # but there's no space for it here.
+        # Here we expect the device updates for `device_id1` and `device_id2`.
+        # That means we only receive 2 updates this time around.
+        # If we had a higher limit, we would expect to see the pair of
+        # (unstable-prefixed & unprefixed) signing key updates for the device
+        # represented by `fakeMaster` and `fakeSelfSigning`.
```
DMRobertson, line 236: Thanks, this helps a lot. (I thought that `device_1` and `device_2` might be expecting to receive a pair of updates each; don't fully grok the cross-signing machinery.)

tests/storage/test_devices.py
DMRobertson: Much appreciated, thank you!

synapse/storage/databases/main/devices.py
DMRobertson: Nah, I think it's okay!

synapse/storage/databases/main/devices.py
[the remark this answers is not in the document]
DMRobertson: Ahh, I think I see. We tell the caller to update the stream id to `last_processed_stream_id` via return value. This only gets bumped at the end of a the loop body over `updates`.

synapse/storage/databases/main/devices.py
```diff
@@ -277,40 +281,83 @@ async def get_device_updates_by_remote(
...
                 key = (user_id, device_id)
 
+                if key not in query_map and remaining_length_budget < 1:
+                    # We don't have space for a new entry
+                    break
+
                 previous_update_stream_id, _ = query_map.get(key, (0, None))
 
                 if update_stream_id > previous_update_stream_id:
                     query_map[key] = (update_stream_id, update_context)
```
DMRobertson, line 335: What's going on here? Could we have problems if the same key occurs twice in `updates`? (Ahh, looks like that'll never happen because of how it's sourced from the DB?)

reivilibre: I think it's intentional: later ones overwrite new ones (hence this `if`-block)

DMRobertson: I was just worried that this might throw away a previous `update_context`, which sounds bad. But we're not changing that soo... maybe we should just leave it

DMRobertson (approved):
LGTM, thanks for bearing with me.

synapse/storage/databases/main/devices.py
reivilibre: That's 'just' the opentracing span. Maybe I'll put a FIXME note in there about it; not sure what the right answer is at this instant — sounds like it could make it hard to trace things down which is not great, but it's not release blocking considering it was always like this ...

Issue #11736, opened 2022-01-13
Admin API room list sorting is not stable


dsonck92:
### Description

When selecting any kind of sort for the `/_synapse/admin/v1/rooms` endpoint, the sorting is unstable.

### Steps to reproduce

- Call `/_synapse/admin/v1/rooms?order_by=creator` (or any other field that has a lot of the same values)
- Note down the IDs, considering these are unique
- Call `/_synapse/admin/v1/rooms?from=xxx&order_by=creator`
- Carefully check the returned IDs. At some point, already seen IDs are returned again.

This happens, in my case, for `creator`, since I have about 1000 rooms created by an application service. These all carry the same `creator` and thus, at least for PostgreSQL, the sorting of rows from the same creator is arbitrary.

I expect that, when I sort on something, and select a new page, I get new results that haven't been seen before.

### Version information

- **Homeserver**: matrix.sonck.nl

If not matrix.org:

- **Version**: 1.49.2

- **Install method**: Docker Image

- **Platform**: Kubernetes on Linux

PR #6720, opened 2020-01-16
Admin API to list, filter and sort rooms

anoadragon453:
Fixes #6424 

~~TODO: docs~~

PR #11737, opened 2022-01-13
Make pagination of rooms in admin api stable

dsonck92:
Always add `state.room_id` after the configurable ORDER BY. Otherwise,
for any sort, certain pages can contain results from
other pages. (Especially when sorting by creator, since there may
be many rooms by the same creator)

Fixes: #11736 

### Pull Request Checklist

### DIFF
--- a/synapse/storage/databases/main/room.py
+++ b/synapse/storage/databases/main/room.py
@@ -552,7 +552,7 @@ class RoomWorkerStore(CacheInvalidationWorkerStore):
             INNER JOIN room_stats_current curr USING (room_id)
             INNER JOIN rooms USING (room_id)
             %s
-            ORDER BY %s %s
+            ORDER BY %s %s, state.room_id ASC
             LIMIT ?
             OFFSET ?
         """ % (
--- a/synapse/storage/databases/main/room.py
+++ b/synapse/storage/databases/main/room.py
@@ -551,24 +551,24 @@ class RoomWorkerStore(CacheInvalidationWorkerStore):
             FROM room_stats_state state
             INNER JOIN room_stats_current curr USING (room_id)
             INNER JOIN rooms USING (room_id)
-            %s
-            ORDER BY %s %s, state.room_id ASC
+            {where}
+            ORDER BY {order_by} {direction}, state.room_id {direction}
             LIMIT ?
             OFFSET ?
-        """ % (
-            where_statement,
-            order_by_column,
-            "ASC" if order_by_asc else "DESC",
+        """.format(
+            where=where_statement,
+            order_by=order_by_column,
+            direction="ASC" if order_by_asc else "DESC",
         )
 
         # Use a nested SELECT statement as SQL can't count(*) with an OFFSET
         count_sql = """
             SELECT count(*) FROM (
               SELECT room_id FROM room_stats_state state
-              %s
+              {where}
             ) AS get_room_ids
-        """ % (
-            where_statement,
+        """.format(
+            where=where_statement,
         )
 
         def _get_rooms_paginate_txn(
--- a/tests/rest/admin/test_room.py
+++ b/tests/rest/admin/test_room.py
@@ -1363,8 +1363,10 @@ class RoomTestCase(unittest.HomeserverTestCase):
         room_id_3 = self.helper.create_room_as(self.admin_user, tok=self.admin_user_tok)
 
         # Also create a list sorted by IDs for properties that are equal (and thus sorted by room_id)
-        all_ids = [room_id_1, room_id_2, room_id_3]
-        all_ids.sort()
+        sorted_by_room_id_asc = [room_id_1, room_id_2, room_id_3]
+        sorted_by_room_id_asc.sort()
+        sorted_by_room_id_desc = sorted_by_room_id_asc.copy()
+        sorted_by_room_id_desc.reverse()
 
         # Set room names in alphabetical order. room 1 -> A, 2 -> B, 3 -> C
         self.helper.send_state(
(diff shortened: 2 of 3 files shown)


DMRobertson (commented):
Good spot. An oversight in #6720?

A few questions:

* What will happen if the request asks to sort by room name descending? Then we'll end up doing `... ORDER BY state.name DESC, state.name ASC ...` which looks like it might be an invalid query? Postgres seems happy with this though.
* This won't be stable if a room's name changes between pagination queries? (I don't think we have a good way to fix this. Could tie-break on room_id, but that's not stable either, because new room ids get created all the time.)

I think the tests for this endpoint might need tweaking too. I've just seen that this is marked as a draft, so I guess you're probably working on this now? Either way, let us know if you could use help!

dsonck92:
I actually realized later that it was wiser to sort on room_id instead, name is not necessarily unique. I just realized that I might have kept the old commit message (which was also used as auto-fill by github).

Yes, I do agree that sorting on room_id may shift results as rooms come and go, however, the current implementation kinda leaves it up to implementation details how the sort will become. In PostgreSQL (at least), the implicit order is mostly defined by insertion and touching order. However, as I worked through my own Synapse Admin derivative, the current implementation can show a room on 3 different pages. If it's based on room_id, then, sure, maybe a room or two get added, so rooms may be duplicated if they happened before the current page, but it doesn't completely throw out the previous sort.

dsonck92:
And yes, small oversight in that commit. I also checked the user list endpoint, but this one does have a fallback. I'm indeed fixing the tests because, apparently they magically always worked (probably because the table mostly responsible for testing was never touched afterwards, keeping the implicit ordering). But, I can push soon as tests work again!

dsonck92:
Regarding a true stable way, coming from GraphQL where there's the opinionated Relay spec, this could potentially be fixed by cursors. In essence, instead of relying on, say, a page offset, you actually encode some cursor data to use to resume. For instance, if you sort on room_id, you can also compare room_id and select anything > last room_id. This way, if you have more items inserted before the current page, this cursor moves along, and the client would only notice the total amount of results changing. However, for the opposite direction, it's more annoying to do (likely some "earlier, swap implicit sort, limit by page, reorder for client"). Not to mention calculating the offset as feedback for which page you're on, and the fact that you now skipped maybe 1.1 pages instead of exactly 1 or 2.

tests/rest/admin/test_room.py
```diff
@@ -1413,31 +1419,31 @@ def _order_test(
...
 
-        _order_test("join_rules", [room_id_1, room_id_2, room_id_3])
-        _order_test("join_rules", [room_id_1, room_id_2, room_id_3], reverse=True)
+        _order_test("join_rules", all_ids)
+        _order_test("join_rules", all_ids, reverse=True)
 
-        _order_test("guest_access", [room_id_1, room_id_2, room_id_3])
-        _order_test("guest_access", [room_id_1, room_id_2, room_id_3], reverse=True)
+        _order_test("guest_access", all_ids)
+        _order_test("guest_access", all_ids, reverse=True)
```
DMRobertson, line 1442: Are we okay with the order not reversing if we fall back to the tie-breaker and ask for items to be listed in reverse order? It's not what I'd expect, but I don't know if it makes a huge difference when writing an admin frontend.

synapse/storage/databases/main/room.py
```diff
@@ -552,7 +552,7 @@ async def get_rooms_paginate(
             INNER JOIN room_stats_current curr USING (room_id)
             INNER JOIN rooms USING (room_id)
             %s
-            ORDER BY %s %s
+            ORDER BY %s %s, state.room_id ASC
```
DMRobertson, line 555: Per my other comment, should this maybe be `state.room_id %s` where the placeholder is `"ASC" if order_by_asc else "DESC"`?

tests/rest/admin/test_room.py
```diff
@@ -1413,31 +1419,31 @@ def _order_test(
...
 
-        _order_test("federatable", [room_id_1, room_id_2, room_id_3])
-        _order_test("federatable", [room_id_1, room_id_2, room_id_3], reverse=True)
+        _order_test("federatable", all_ids)
+        _order_test("federatable", all_ids, reverse=True)
 
-        _order_test("public", [room_id_1, room_id_2, room_id_3])
+        _order_test("public", all_ids)
         # Different sort order of SQlite and PostreSQL
         # _order_test("public", [room_id_3, room_id_2, room_id_1], reverse=True)
```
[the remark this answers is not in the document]
dsonck92: I think the tests should all work equally for PostgreSQL and SQlite yes.

tests/rest/admin/test_room.py
```diff
@@ -1360,6 +1362,10 @@ def _order_test(
         room_id_2 = self.helper.create_room_as(self.admin_user, tok=self.admin_user_tok)
         room_id_3 = self.helper.create_room_as(self.admin_user, tok=self.admin_user_tok)
 
+        # Also create a list sorted by IDs for properties that are equal (and thus sorted by room_id)
+        all_ids = [room_id_1, room_id_2, room_id_3]
```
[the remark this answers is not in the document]
dsonck92: Yes, I think that's a better name.

tests/rest/admin/test_room.py
```diff
@@ -1419,31 +1421,30 @@ def _order_test(
             "joined_local_members", [room_id_1, room_id_2, room_id_3], reverse=True
         )
 
-        _order_test("version", all_ids)
-        _order_test("version", all_ids, reverse=True)
+        _order_test("version", sorted_by_room_id_asc, reverse=True)
+        _order_test("version", sorted_by_room_id_desc)
```
dsonck92, line 1425: I have no idea why this requires the reverse to be swapped. It was required locally, maybe CI tests differently, but I cannot explain it since all rooms should be the same version, and reversing works the same (and correctly) for the rest of the test cases

DMRobertson: That's odd. Thanks for flagging it up; let's see what CI says and maybe we can investigate in more detail later.

DMRobertson: Ahh, I think I see it now. The initial sort order is controlled by `order_by_asc`, which differs between different sorting orders. See e.g.:

https://github.com/matrix-org/synapse/blob/d2e595f379720ac0fc8225db1cd999fe31e76cfa/synapse/storage/databases/main/room.py#L506-L511

`dir=b` reverses this rather than setting it to False.

I guess the idea is to use the "natural" sort direction by default while still allowing the user to reverse it. Not sure if I agree with that choice, but it looks like the behaviour will be consistent.

DMRobertson: (This goes back to #6720)

DMRobertson (changes requested):
Could you leave a comment above the order test for `"version"` explaining what's going on? Something like `# Note: versions are sorted in descending order when dir=f` perhaps?

The other sort orders that behave in this way are "joined_members", "joined_local_members" and "state_events". Those all have explicit tests that don't rely on the tie-breaker mechanism, but perhaps similar comments would be worthwhile? I'll leave that to your discretion.

Once that's done I think this is good to go!

tests/rest/admin/test_room.py
anoadragon453: > I guess the idea is to use the "natural" sort direction by default while still allowing the user to reverse it. 

That indeed was the intention, and back then there were only two sort orders - size of joined members and alphabetical room names. It seemed sensible for users to want to view rooms sorted by size of joined members descending, whereas users would want to see room names sorted from a->z.

Of course, client software could just set the `dir=b` flag itself when sorting alphabetically, but that's extra work for the client. Default directions are noted in the [docs](https://matrix-org.github.io/synapse/latest/admin_api/rooms.html#list-room-api).

I'm just coming into this conversation without much context... but I fail to see why setting the direction order based on ordering type is confusing.

DMRobertson: > but I fail to see why setting the direction order based on ordering type is confusing.

In a nutshell, we have different interpretations of what `dir=b` ought to mean
- as-implemented: reverse the natural ordering
- our guess: sort the results descending by the given criterion

dsonck92: Aah I see, yes, in my case this will end up confusing:

image

Because the arrow directly translates to `b` or `f` (as usually done through APIs, but usually `asc` `desc` where meaning is well defined), and there is some meaning behind the arrow. So now, blindly translating the direction to `b` or `f` will give the wrong arrow direction. This could be fixed client-side by keeping track of the natural way of sorting, but then the logic becomes more complex and based on how it's written in Synapse. (One could argue it's already tightly bound to synapse as it works with this specific API, but it makes the sorting non-trivial)

dsonck92: Personally, I feel like the initial sort direction is the choice of the client, and the backend just returns exactly what you ask it. That way, the client can be written knowing exactly what it will return without looking at the docs/source code and without having some lookup table to correctly swap arrows based on field. Because at that stage, you're essentially doing double work: backend is opinionated about sorting direction, so will pick a (well defined, but seemingly arbitrary) direction that can be flipped, and client compensates for this opinion by swapping the arrow to match the direction.

DMRobertson (approved):
LGTM. Thanks for seeing this through!

tests/rest/admin/test_room.py
anoadragon453: Thanks for the discussion both - I've ended up coming to the same conclusion. It is unnecessary to maintain this "lookup table" on both ends.

I've created a separate issue to track this at https://github.com/matrix-org/synapse/issues/11759.

PR #11740, opened 2022-01-13
Mention python3-venv and libpq-dev dependencies in contribution guide

andybalaam:


docs/development/contributing_guide.md
```diff
@@ -20,7 +20,9 @@ recommended for development. More information about WSL can be found at
 <https://docs.microsoft.com/en-us/windows/wsl/install>. Running Synapse natively
 on Windows is not officially supported.
 
-The code of Synapse is written in Python 3. To do pretty much anything, you'll need [a recent version of Python 3](https://wiki.python.org/moin/BeginnersGuide/Download).
+The code of Synapse is written in Python 3. To do pretty much anything, you'll need [a recent version of Python 3](https://wiki.python.org/moin/BeginnersGuide/Download), including support for [virtual environments](https://docs.python.org/3/library/venv.html). On Debian-derived Linux distros running `sudo apt install python3-venv` should be enough.
```
callahad, line 23:
```suggestion
The code of Synapse is written in Python 3. To do pretty much anything, you'll need [a recent version of Python 3](https://www.python.org/downloads/). Your Python also needs support for [virtual environments](https://docs.python.org/3/library/venv.html). This is usually built-in, but some Linux distributions like Debian and Ubuntu split it out into its own package. Running `sudo apt install python3-venv` should be enough.
```

callahad (approved):
Tiiiiny little wordsmithing nits. Up to you to accept/reject them.

(Also changing to point to the python.org homepage, rather than the Python wiki, for download instructions. The wiki is way out of date.)

PR #11742, opened 2022-01-13
Replace uses of simple_insert_many with simple_insert_many_values.

clokep:
And the `_txn` form, and then renames `simple_insert_many_values` back to `simple_insert_many`. This finishes the work started in #11560 and should give us some more gains by avoiding creating intermediary dictionaries.

Should probably be reviewed commit-by-commit.

### DIFF
--- a/changelog.d/11742.misc
+++ b/changelog.d/11742.misc
@@ -1,0 +1,1 @@
+Minor efficiency improvements when inserting many values into the database.
--- a/synapse/storage/database.py
+++ b/synapse/storage/database.py
@@ -933,41 +933,6 @@ class DatabasePool:
 
         txn.execute(sql, vals)
 
-    @staticmethod
-    def simple_insert_many_txn(
-        txn: LoggingTransaction, table: str, values: List[Dict[str, Any]]
-    ) -> None:
-        """Executes an INSERT query on the named table.
-
-        The input is given as a list of dicts, with one dict per row.
-        Generally simple_insert_many_values_txn should be preferred for new code.
-
-        Args:
-            txn: The transaction to use.
-            table: string giving the table name
-            values: dict of new column names and values for them
-        """
-        if not values:
-            return
-
-        # This is a *slight* abomination to get a list of tuples of key names
-        # and a list of tuples of value names.
-        #
-        # i.e. [{"a": 1, "b": 2}, {"c": 3, "d": 4}]
-        #         => [("a", "b",), ("c", "d",)] and [(1, 2,), (3, 4,)]
-        #
-        # The sort is to ensure that we don't rely on dictionary iteration
-        # order.
-        keys, vals = zip(
-            *(zip(*(sorted(i.items(), key=lambda kv: kv[0]))) for i in values if i)
-        )
-
-        for k in keys:
-            if k != keys[0]:
-                raise RuntimeError("All items must have the same keys")
-
-        return DatabasePool.simple_insert_many_values_txn(txn, table, keys[0], vals)
-
     async def simple_insert_many_values(
         self,
         table: str,
--- a/synapse/storage/database.py
+++ b/synapse/storage/database.py
@@ -933,21 +933,6 @@ class DatabasePool:
 
         txn.execute(sql, vals)
 
-    async def simple_insert_many(
-        self, table: str, values: List[Dict[str, Any]], desc: str
-    ) -> None:
-        """Executes an INSERT query on the named table.
-
-        The input is given as a list of dicts, with one dict per row.
-        Generally simple_insert_many_values should be preferred for new code.
-
-        Args:
-            table: string giving the table name
-            values: dict of new column names and values for them
-            desc: description of the transaction, for logging and metrics
-        """
-        await self.runInteraction(desc, self.simple_insert_many_txn, table, values)
-
     @staticmethod
     def simple_insert_many_txn(
         txn: LoggingTransaction, table: str, values: List[Dict[str, Any]]
--- a/synapse/storage/database.py
+++ b/synapse/storage/database.py
@@ -933,7 +933,7 @@ class DatabasePool:
 
         txn.execute(sql, vals)
 
-    async def simple_insert_many_values(
+    async def simple_insert_many(
         self,
         table: str,
         keys: Collection[str],
@@ -952,11 +952,11 @@ class DatabasePool:
             desc: description of the transaction, for logging and metrics
         """
         await self.runInteraction(
-            desc, self.simple_insert_many_values_txn, table, keys, values
+            desc, self.simple_insert_many_txn, table, keys, values
         )
 
     @staticmethod
-    def simple_insert_many_values_txn(
+    def simple_insert_many_txn(
         txn: LoggingTransaction,
         table: str,
         keys: Collection[str],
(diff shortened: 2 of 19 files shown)


synapse/storage/databases/main/events_bg_updates.py
```diff
@@ -803,36 +804,27 @@ def get_rejected_events(
...
                 # need to deduplicate as we have a unique constraint.
                 for auth_id in set(event.auth_event_ids()):
-                    auth_events.append(
-                        {
-                            "room_id": event.room_id,
-                            "event_id": event.event_id,
-                            "auth_id": auth_id,
-                        }
-                    )
+                    auth_events.append((event.event_id, event.room_id, auth_id))
```
clokep, line 814: Note that these were reordered to have `event_id`, `room_id` as the first items in the tuple, this matches what we do for `state_events` and seems a bit clearer IMO.

synapse/storage/databases/main/user_directory.py
```diff
@@ -105,8 +105,10 @@ def _make_staging_area(txn: LoggingTransaction) -> None:
                 GROUP BY room_id
             """
             txn.execute(sql)
-            rooms = [{"room_id": x[0], "events": x[1]} for x in txn.fetchall()]
-            self.db_pool.simple_insert_many_txn(txn, TEMP_TABLE + "_rooms", rooms)
+            rooms = list(txn.fetchall())
+            self.db_pool.simple_insert_many_txn(
+                txn, TEMP_TABLE + "_rooms", keys=("room_id", "events"), values=rooms
+            )
             del rooms
```
clokep, line 112: I wondered if I could just set `values=txn.fetchall()`, but I recalled there being something sketchy with iterating a `txn` as you start another query? Figured it was safer to cast it to a list first.

synapse/storage/databases/main/account_data.py
```diff
@@ -536,9 +536,9 @@ def _add_account_data_for_user(
         self.db_pool.simple_insert_many_txn(
             txn,
             table="ignored_users",
+            keys=("ignorer_user_id", "ignored_user_id"),
             values=[
-                {"ignorer_user_id": user_id, "ignored_user_id": u}
-                for u in currently_ignored_users - previously_ignored_users
+                (user_id, u) for u in currently_ignored_users - previously_ignored_users
```
clokep, line 541: There's a handful (maybe 1/3rd?) of callers which could benefit from passing in some constant values for some keys, but I figured that could be left to another round of optimization, if we wanted to complicate the API. (I'm unsure if there's real benefit to it...)

DMRobertson:
Looks like a few tests are failing---something about a non-NULL constraint in registration_tokens.token

tests/rest/admin/test_registration_tokens.py
```diff
@@ -223,20 +223,13 @@ def test_create_unable_to_generate_token(self) -> None:
...
-                    "expiry_time": None,
-                }
-            )
+            tokens.append((c, None, 0, 0, None))
         self.get_success(
             self.store.db_pool.simple_insert_many(
                 "registration_tokens",
-                tokens,
-                "create_all_registration_tokens",
+                keys={"token", "uses_allowed", "pending", "completed", "expiry_time"},
```
DMRobertson, line 230: This is a set---should it be a list or tuple (i.e. ordered?)

DMRobertson: Next thought: why didn't mypy spot this? Because `keys: Collection[str]`. Perhaps `Sequence[str]` is what we want?

DMRobertson: (This is in draft---sorry if I'm treading on your toes here)

clokep: Thanks, yeah this wasn't ready for review yet I was still running tests and noticed this and another issue!

synapse/storage/database.py
```diff
@@ -933,41 +933,6 @@ def simple_insert_txn(
...
-
-        Args:
-            txn: The transaction to use.
-            table: string giving the table name
-            values: dict of new column names and values for them
-        """
-        if not values:
-            return
-
-        # This is a *slight* abomination to get a list of tuples of key names
```
DMRobertson, line 968: For some value of "slight"!

changelog.d/11742.misc
```diff
@@ -0,0 +1 @@
+Minor efficiency improvements when inserting many values into the database.
```
DMRobertson, line 1: So I guess the efficiency savings here are:
- the interpreter doesn't have to have to have `PyObject*` pointers repeatedly pointing to the same set of keys (column names)
  - not sure if it'd repeatedly allocate for copies of the same string literals?
- the interpreter doesn't have to spend time calculating hashes, resizing the hash table, ... and any other fun involved in maintaining a dictionary
- we don't have to spend time doing sorting in the code removed in https://github.com/matrix-org/synapse/pull/11742/commits/36919b1f415cdc895548c253d4583fe8380201ca, and we save a function call too

DMRobertson (approved):
This looks great---nice to have one fewer way to do things.

I think the type hint for `simple_insert_many` is dangerously incorrect though:
- `keys: Collection[str]` allows `keys: Set[str]`, but we rely on the iteration order of `keys` to be consistent with the iteration order of `values`.
  ```
  >>> from typing import *
  >>> isinstance(set(), Collection)
  True
  >>> isinstance(set(), Sequence)
  False
  ```
   Suggest `keys: Sequence[str]`.
- Similarly `values: Collection[Collection[Any]]` looks like it ought to be `values: Collection[Sequence[Any]]`.

Related: #11564 #11580

Do you agree? If so, perhaps this is best fixed in a separate PR.

changelog.d/11742.misc
clokep: I was mostly thinking of the allocation of all the dictionaries, but yes to all of that!

I also copied this changelog from the previous PR.

synapse/storage/databases/main/user_directory.py
```diff
@@ -105,8 +105,10 @@ def _make_staging_area(txn: LoggingTransaction) -> None:
                 GROUP BY room_id
             """
             txn.execute(sql)
-            rooms = [{"room_id": x[0], "events": x[1]} for x in txn.fetchall()]
-            self.db_pool.simple_insert_many_txn(txn, TEMP_TABLE + "_rooms", rooms)
+            rooms = list(txn.fetchall())
+            self.db_pool.simple_insert_many_values_txn(
+                txn, TEMP_TABLE + "_rooms", keys=("room_id", "events"), values=rooms
+            )
```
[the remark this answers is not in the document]
clokep: Yeah this looked a bit sketchy to me in general, but I wasn't going to try to refactor it too much!

clokep:
> I think the type hint for `simple_insert_many` is dangerously incorrect though:

That may be true, and might also apply to other `simple_*` methods, probably worth a quick audit. 👍 

Thanks for the review!

synapse/storage/databases/main/user_directory.py
reivilibre: I thought `list(txn)` and `txn.fetchall()` were equivalent to be honest. Doubt it matters but this looks like it's building the list twice needlessly. See https://www.python.org/dev/peps/pep-0249/#fetchall (Twisted notes that it's compatible with the standard DBAPI).

PR #11710, opened 2022-01-09
Fix missing app variable in mail subject

br4nnigan:
homeserver.yaml documentation claims that you can use the %(app)s variable in password_reset and email_validation subjects, but if you do you end up with an error 500

for seo reasons here's the complete error

```
2022-01-09 19:57:33,737 - synapse.handlers.identity - 415 - ERROR - POST-43 - Error sending threepid validation email to operator@mydomain.net
Traceback (most recent call last):
  File "/opt/venvs/matrix-synapse/lib/python3.7/site-packages/synapse/handlers/identity.py", line 412, in send_threepid_validation
    await send_email_func(email_address, token, client_secret, session_id)
  File "/opt/venvs/matrix-synapse/lib/python3.7/site-packages/synapse/push/mailer.py", line 212, in send_add_threepid_mail
    % {"server_name": self.hs.config.server.server_name},
KeyError: 'app'
2022-01-09 19:57:33,738 - synapse.http.server - 92 - INFO - POST-43 - <XForwardedForRequest at 0x7f3ebc607ef0 method='POST' uri='/_matrix/client/r0/account/3pid/email/requestToken' clientproto='HTTP/1.0' site='8008'> SynapseError: 500 - An error was encountered when sending the email
```

PR #11745, opened 2022-01-13
 Fix missing app variable in mail subject

DMRobertson:
From  #11710, originally by @br4nnigan.

My changes are running the linter and adding a changelog. I meant to do this in #11710 but screwed up my usage of (ma)git royally; apologies to all.


DMRobertson:
Setting this to automerge without review, because I reviewed #11710 and my only changes are linting + changelog.

PR #11752, opened 2022-01-14
Fetch thread summaries for multiple events in a single query

clokep:
This uses the same approach as #11660, but for the thread relation type instead of the edit relation type.

I did similar performance testing on this as in #11660 and #11612 and it seems to be ~15-20% improvement compared to develop when MSC3440 enabled. (The changes in this PR bring performance roughly on par as develop with MSC3440 *disabled*.)

Part of #11825.

### DIFF
--- a/synapse/storage/databases/main/relations.py
+++ b/synapse/storage/databases/main/relations.py
@@ -466,17 +466,36 @@ class RelationsWorkerStore(SQLBaseStore):
         ) -> Tuple[Dict[str, int], Dict[str, str]]:
             # Fetch the count of threaded events and the latest event ID.
             # TODO Should this only allow m.room.message events.
-            sql = """
-                SELECT parent.event_id, child.event_id FROM events AS child
-                INNER JOIN event_relations USING (event_id)
-                INNER JOIN events AS parent ON
-                    parent.event_id = relates_to_id
-                    AND parent.room_id = child.room_id
-                WHERE
-                    %s
-                    AND relation_type = ?
-                ORDER BY child.topological_ordering DESC, child.stream_ordering DESC
-            """
+            if isinstance(self.database_engine, PostgresEngine):
+                # The `DISTINCT ON` clause will pick the *first* row it encounters,
+                # so ordering by topologica ordering + stream ordering desc will
+                # ensure we get the latest event in the thread.
+                sql = """
+                    SELECT DISTINCT ON (parent.event_id) parent.event_id, child.event_id FROM events AS child
+                    INNER JOIN event_relations USING (event_id)
+                    INNER JOIN events AS parent ON
+                        parent.event_id = relates_to_id
+                        AND parent.room_id = child.room_id
+                    WHERE
+                        %s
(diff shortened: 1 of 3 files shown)


synapse/storage/databases/main/relations.py
```diff
@@ -454,106 +455,175 @@ def _get_applicable_edits_txn(txn: LoggingTransaction) -> Dict[str, str]:
...
+            )
+            args.append(RelationTypes.THREAD)
 
-            txn.execute(sql, (event_id, room_id, RelationTypes.THREAD))
-            row = txn.fetchone()
-            if row is None:
-                return 0, None
+            txn.execute(sql % (clause,), args)
+            latest_event_ids = {}
+            for parent_event_id, child_event_id in txn.fetchall():
```
erikjohnston, line 521:
```suggestion
            for parent_event_id, child_event_id in txn:
```

synapse/storage/databases/main/relations.py
```diff
@@ -454,106 +455,175 @@ def _get_applicable_edits_txn(txn: LoggingTransaction) -> Dict[str, str]:
...
+                    SELECT DISTINCT ON (parent.event_id) parent.event_id, child.event_id FROM events AS child
+                    INNER JOIN event_relations USING (event_id)
+                    INNER JOIN events AS parent ON
+                        parent.event_id = relates_to_id
+                        AND parent.room_id = child.room_id
+                    WHERE
+                        %s
+                        AND relation_type = ?
+                    ORDER BY parent.event_id, child.topological_ordering DESC, child.stream_ordering DESC
+                """
```
erikjohnston, line 497: If you wanted to combine the queries for last value and counts, you could use a windows:

```sql
SELECT DISTINCT ON (parent.event_id) parent.event_id, count(*) OVER w, first_value(child.event_id) OVER w FROM events AS child
INNER JOIN event_relations USING (event_id)
INNER JOIN events AS parent ON
    parent.event_id = relates_to_id
    AND parent.room_id = child.room_id
WHERE
    relates_to_id = ?
    AND relation_type = 'io.element.thread'
WINDOW w AS (
    PARTITION BY parent.event_id
    ORDER BY parent.event_id, child.topological_ordering DESC, child.stream_ordering DESC
    RANGE BETWEEN UNBOUNDED PRECEDING AND UNBOUNDED FOLLOWING
);
```

which I think is as efficient. What's going on here is that the `count() OVER w` and `first_value(..) OVER w` is operating on the set of rows defined in the window `w`, which in this case is all rows with the same `parent.event_id`.

But it's probably easier to keep them as separate queries :)

erikjohnston (approved):

synapse/storage/databases/main/relations.py
clokep: We chatted a bit about this and it seems like it would diverge the Postgresql and sqlite code more, which isn't ideal. @erikjohnston said he was mostly curious if it was doable. 👍 

I'll leave this to a future enhancement.

Issue #8631, opened 2020-10-22
Device lists are desynchronizing


turt2live:
### Description

People seem to be seeing/getting outdated device lists from Synapse regardless of server. This is manifesting in the fairly normal ways: wrong names, devices that should have been deleted, etc. New devices seem to be included correctly all the time, but usually with a mix of incorrect devices.

`insert into device_lists_remote_resync (user_id, added_ts) values ('@victim:example.org', (extract(epoch from now())::bigint * 1000));` fixes the problem for the relevant victims.

Appears to be a wider issue than just matrix.org

Known cases:
* Someone deleted their extra devices, but no one saw that.
* Someone cross-signed their new device, but no one believed them.
* Someone renamed their device, but no one saw that.

The rename trick for trying to confuse/convince the server into updating doesn't appear to work anymore either - only the database insert functions.

Started appearing months ago.

clokep:
Is this a duplicate / superset of #8459?

turt2live:
ah ha, I did report it at the time - I'm so smart. I'd close https://github.com/matrix-org/synapse/issues/8459 in favour of this as this is the more general version of "it's all broken", but I'll leave that up to you.

richvdh:
related: #4827?

deepbluev7:
We have a similar issue. We have a bot with 1500 devices (for E2EE performance testing). Its device list only shows up on some servers though. This may be easier to reproduce with such bigger device lists?

Erethon:
Just ran into this myself, a friend on matrix.org had about 480 devices, deleted all of them except 4 and my homeserver reported 318 devices still being around. Running the `insert` query from the OP fixed it for me.

deepbluev7:
In our case it was https://github.com/matrix-org/synapse/blob/b460a088c647a6d3ea0e5a9f4f80d86bb9e303b3/synapse/handlers/device.py#L902 which stopped syncing our 1500 devices list.

rihardsk:
@erikjohnston i'm still facing the issue, yes. On Synapse 1.48.

DMRobertson:
From my reading of the spec, clients either

- watch `device_lists.changed/left` in the `sync` response, or
- periodically query `/keys/changes` after syncs

to discover device list changes or that they can forget a user's devices entirely.

Assuming my understanding is correct, I see three ways this can go wrong on Synapse's end. Either:

1. the `/sync` response incorrectly calculates the list of `changed` or `left` users in `device_lists`; or
[...]

PR #11760, opened 2022-01-17
Debug for device lists updates

DMRobertson:
Debug for #8631. I want to establish if we actually _are_ sending out the
correct EDUs over federation.

I'm having a hard time tracking down what's going wrong in that issue.
In the reported example, I could see server A sending federation traffic
to server B and all was well. Yet B reports out-of-sync device updates
from A.

I couldn't see what was _in_ the events being sent from A to B. So I
have added some entirely untested crude logging to track

- when we have updates to send to a remote HS
- the edus we actually accumulate to send
- when a federation transaction includes a device list update edu
- when such an EDU is received

This is a bit of a sledgehammer. Is this a good idea? I'm anxious that this might create too much logspam.

### DIFF
--- a/synapse/federation/sender/transaction_manager.py
+++ b/synapse/federation/sender/transaction_manager.py
@@ -132,7 +132,7 @@ class TransactionManager:
                 ]
                 if device_list_updates:
                     issue_8631_logger.debug(
-                        "transaction [%s] includes device list updates: %s",
+                        "about to send txn [%s] including device list updates: %s",
                         transaction.transaction_id,
                         device_list_updates,
                     )
--- a/synapse/federation/sender/transaction_manager.py
+++ b/synapse/federation/sender/transaction_manager.py
@@ -35,6 +35,7 @@ if TYPE_CHECKING:
     import synapse.server
 
 logger = logging.getLogger(__name__)
+issue_8631_logger = logging.getLogger("synapse.8631_debug")
 
 last_pdu_ts_metric = Gauge(
     "synapse_federation_last_sent_pdu_time",
@@ -124,6 +125,17 @@ class TransactionManager:
                 len(pdus),
                 len(edus),
             )
+            if issue_8631_logger.isEnabledFor(logging.DEBUG):
+                DEVICE_UPDATE_EDUS = {"m.device_list_update", "m.signing_key_update"}
+                device_list_updates = [
+                    edu.content for edu in edus if edu.edu_type in DEVICE_UPDATE_EDUS
+                ]
+                if device_list_updates:
+                    issue_8631_logger.debug(
+                        "transaction [%s] includes device list updates: %s",
+                        transaction.transaction_id,
+                        device_list_updates,
+                    )
 
             # Actually send the transaction
 
--- a/synapse/federation/transport/server/federation.py
+++ b/synapse/federation/transport/server/federation.py
@@ -105,7 +105,7 @@ class FederationSendServlet(BaseFederationServerServlet):
                 ]
                 if device_list_updates:
                     issue_8631_logger.debug(
-                        "transaction [%s] includes device list updates: %s",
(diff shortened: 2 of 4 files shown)


synapse/federation/sender/transaction_manager.py
```diff
@@ -124,6 +125,17 @@ async def send_new_transaction(
                 len(pdus),
                 len(edus),
             )
+            if issue_8631_logger.isEnabledFor(logging.DEBUG):
+                DEVICE_UPDATE_EDUS = {"m.device_list_update", "m.signing_key_update"}
+                device_list_updates = [
+                    edu.content for edu in edus if edu.edu_type in DEVICE_UPDATE_EDUS
+                ]
+                if device_list_updates:
```
clokep, line 133: Would we want to log if the transaction doesn't include any?

synapse/federation/transport/server/federation.py
```diff
@@ -95,6 +96,20 @@ async def on_PUT(
...
+            if issue_8631_logger.isEnabledFor(logging.DEBUG):
+                DEVICE_UPDATE_EDUS = {"m.device_list_update", "m.signing_key_update"}
+                device_list_updates = [
+                    edu.content
+                    for edu in transaction_data.get("edus", [])
+                    if edu.edu_type in DEVICE_UPDATE_EDUS
+                ]
+                if device_list_updates:
+                    issue_8631_logger.debug(
+                        "transaction [%s] includes device list updates: %s",
```
clokep, line 108: Do you want to make this log line unique from the other? It looks like this is the receiving side?

synapse/storage/databases/main/devices.py
```diff
@@ -222,13 +223,18 @@ async def get_device_updates_by_remote(
             limit,
         )
 
-        # We need to ensure `updates` doesn't grow too big.
-        # Currently: `len(updates) <= limit`.
+        # Note for later that `len(updates) <= limit`.
```
clokep, line 226: Not sure if this comment change makes sense?

clokep (commented):
Looks OK overall. What would be a plan for enabling this? I could see the logs ballooning quite fast!

synapse/storage/databases/main/devices.py
DMRobertson: Ahhh: when I first reviewed this I assumed we were growing `updates` in the rest of the function. We're not though: we gradually build up `results` instead. cc @reivilibre

synapse/federation/sender/transaction_manager.py
DMRobertson: I don't think so---I'm guessing most of these EDUs are typing notifications, which I don't want to spam the logs with.

synapse/federation/transport/server/federation.py
DMRobertson: Ahh, good idea. Thanks!

synapse/federation/sender/transaction_manager.py
clokep: Ah, good point! 👍

synapse/storage/databases/main/devices.py
clokep: It might be an OK change, it just surprised me a bit to be in a PR about adding debug to figure out what is happening... 😄

DMRobertson: That's fair. I can pull that out for now.

clokep (approved):

DMRobertson:
> What would be a plan for enabling this?

Not sure: I don't have anything better to suggest than "turn it on for matrix.org and wait for a bug report about dodgy devices". I was considering making it so that we only log something if there's a device list update for a user in the list of "force-traced" users. But since the S-S communication isn't associated with a requester I think this would require some thought; I suspected it'd be quicker to just shove it in and see how spammy it is.

PR #11761, opened 2022-01-18
Remove `log_function` and its uses

richvdh:
I've never found this terribly useful. I think it was added in the early days
of Synapse, without much thought as to what would actually be useful to log,
and has just been cargo-culted ever since.

Rather, it tends to clutter up debug logs with useless information.


anoadragon453 (commented):
I don't think I've used this function much when debugging either - especially as I so rarely need to head down to DEBUG log level these days.

I'd like to explicitly double-check with @erikjohnston though.

clokep (approved):
LGTM if CI is happy!

richvdh:
(@erikjohnston said he was happy in chat)

Issue #11763, opened 2022-01-18
Synapse 1.50.0 returns 404 Unrecognized request for all client requests if a `webclient` resource is enabled


anoadragon453:
TL;DR if you have the `client` and `webclient` resources enabled on a port in your `listeners` config section, attempts to access `client` resources may fail with a `404`, **breaking client functionality**. Workaround is to either:

* Re-order your `resources` config option such that `webclient` appears before `client`. Example:
    ```yaml
    resources:
      - names: [webclient, client, ...]
      ...
   ```

    though note that if you rely on the `webclient` resource, this workaround will render it inaccessible. Thus it is functionally equivalent to the next workaround:

* Remove `webclient` from your list of resources, or
* [downgrade](https://matrix-org.github.io/synapse/latest/upgrade#rolling-back-to-older-versions) to Synapse 1.49.2.

First noticed in https://github.com/matrix-org/synapse/issues/11762.

This first appears to be related to changes made in https://github.com/matrix-org/synapse/pull/11602

anoadragon453:
A fix for this issue was implemented in https://github.com/matrix-org/synapse/pull/11764 has been released in [Synapse v1.50.1](https://github.com/matrix-org/synapse/releases/tag/v1.50.1).

PR #11764, opened 2022-01-18
Revert #11602 to prevent webclient overriding client resource

anoadragon453:
Fixes https://github.com/matrix-org/synapse/issues/11763

#11602 compiled known client resource prefixes under one, simplified '/_matrix/client' prefix. Unfortunately, '/_matrix/client' is already a prefix used by the 'webclient' resource. As we use a Python dictionary to hold a mapping from prefix to resource, configuring a 'webclient' resource after a 'client' resource would end up with all client requests being redirected to the 'webclient' resource (which doesn't know how to handle things like `/login` or `/sync`).

This PR essentially reverts #11602 in order to provide a working version of Synapse v1.50.x. A proper fix will come later. Additionally, this bug has highlighted how fragile the resource code is currently.

### DIFF
--- a/changelog.d/11764.bugfix
+++ b/changelog.d/11764.bugfix
@@ -1,0 +1,1 @@
+Fixes a bug introduced in Synapse 1.50.0rc1 that could cause Matrix clients to be unable to connect to Synapse instances with the 'webclient' resource enabled.
--- a/synapse/app/homeserver.py
+++ b/synapse/app/homeserver.py
@@ -27,7 +27,6 @@ import synapse
 import synapse.config.logger
 from synapse import events
 from synapse.api.urls import (
-    CLIENT_API_PREFIX,
     FEDERATION_PREFIX,
     LEGACY_MEDIA_PREFIX,
     MEDIA_R0_PREFIX,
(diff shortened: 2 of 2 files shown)


babolivier:
> This PR fixes the bug while maintaining JSON responses as the PR intended

Does it maintain JSON responses _as the PR intended_? It seems to me like the initial intent of the PR was to respond with JSON if the path wasn't matching of the client prefixes we list explicitly.

clokep:
> > This PR fixes the bug while maintaining JSON responses as the PR intended
> 
> Does it maintain JSON responses _as the PR intended_? It seems to me like the initial intent of the PR was to respond with JSON if the path wasn't matching of the client prefixes we list explicitly.

Yeah, I'd be curious if you a request `/_matrix/client/not-a-real-prefix` returns a JSON 404 response or not.

anoadragon453:
You are both correct! HTML is still returned:

```
> curl http://localhost:8008/_matrix/client/not-a-real-prefix
<html>
  <head><title>404 - No Such Resource</title></head>
  <body>
    <h1>No Such Resource</h1>
    <p>Sorry. No luck finding that resource.</p>
  </body>
</html>
```

Seems I didn't read the code hard enough. I'll update the description to reflect that this PR is now essentially fully reverting #11602. I assume we are still fine with that?

clokep:
It somewhat means we cannot add any new prefixes, but we can try to fix that separately, I suppose.

babolivier:
I am. Responding with JSON on unknown prefixes is nice but imo not something we need to fix now now now. So I'd suggest to park it until we get in a more comfortable position wrt web client redirections.

clokep:
Note that this will reopen #11764.

anoadragon453:
I have tested this branch locally and it looks to resolve the issue.

Issue #9078, opened 2021-01-12
Deprecate and then remove web_client_location


aaronraimist:
Having Synapse host a web client is a security risk, as mentioned here: https://github.com/matrix-org/synapse#security-note and I doubt many people are still using it. I think you should deprecate the `web_client_location` option in `homeserver.yaml`.

It would also close at least one issue: https://github.com/matrix-org/synapse/issues/1250

aaronraimist:
For sure the option to point it at a path on disk and have it be served from `/_matrix/client` should be removed. It is possible that people are relying on the redirect to another URL option but I wouldn't think very many people would be using it.

I'm not sure if this would be considered an unspecced API https://github.com/matrix-org/synapse/issues/8334. As far as I know nothing in the spec mentions hosting a client at `/_matrix/client`.

clokep:
Thanks for filing this! I kept meaning to file an issue about removing this... leaving the redirect seems safe, but would be simpler to remove it altogether. 🤷

clokep:
We had a discussion about this today in #synapse-dev with the conclusion:

* we deprecate *non-redirect* `web_client_location`, and the `webclient` listener, with intention to remove support for it in a release or two.
* in the meantime, we can also make `/` redirect directly to the `web_client_location` if it's absolute, rather than bouncing via `/_matrix/client`

clokep:
I believe this is now fixed by #11895 -- the `webclient` listener no longer exists and `web_client_location` must be a HTTP(S) URL which gets turned into a redirect.

PR #11765, opened 2022-01-18
Add a regression test for using both webclient and client resources simultaneously

anoadragon453:
Add a regression test for the bug detailed in https://github.com/matrix-org/synapse/issues/11763.

This test configures a `SynapseSite` with both a `webclient` and `client` resource, then makes a `GET` request against both `/_matrix/client` and `/_matrix/client/v3/login`, check that the appropriate status codes and response header/bodies were returned. Both possible orderings of configuring `webclient` and `client` are attempted (`[webclient, client]` vs `[client, webclient]`).

Manipulating resources in unit tests ended up a lot more complicated than originally desired. I was hoping to just be able to `override_config` with the `listeners` section and be done with it. Unfortunately, resources for homeservers created when inheriting `HomeserverTestCase` are hardcoded: https://github.com/matrix-org/synapse/blob/d6fb96e056f79de220d8d59429d89a61498e9af3/tests/unittest.py#L257-L267

Note that I don't think inheriting from `HomeserverTestCase` is actually necessary for this test - but it may be useful if we ever add more test cases to this class (if we do before we drop `webclient` entirely: #9078).

### DIFF
--- a/tests/http/test_webclient.py
+++ b/tests/http/test_webclient.py
@@ -11,6 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+from http import HTTPStatus
 from typing import Dict
 
 from twisted.web.resource import Resource
@@ -90,7 +91,8 @@ class WebClientTests(HomeserverTestCase):
             method="GET",
             path="/_matrix/client",
         )
-        self.assertEqual(channel.code, 302)
+        # Check that we are being redirected to the webclient location URI.
+        self.assertEqual(channel.code, HTTPStatus.FOUND)
         self.assertEqual(
             channel.headers.getRawHeaders("Location"), ["https://example.org"]
         )
@@ -102,5 +104,5 @@ class WebClientTests(HomeserverTestCase):
             method="GET",
             path="/_matrix/client/v3/login",
         )
-        self.assertEqual(channel.code, 200)
+        self.assertEqual(channel.code, HTTPStatus.OK)
         self.assertIn("flows", channel.json_body)
--- a/tests/http/test_webclient.py
+++ b/tests/http/test_webclient.py
@@ -1,0 +1,106 @@
+# Copyright 2022 The Matrix.org Foundation C.I.C.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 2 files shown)


tests/http/test_webclient.py
```diff
@@ -0,0 +1,106 @@
...
+
+        # Ensure that the *webclient* resource is behaving as expected (we get redirected to
+        # the configured web_client_location)
+        channel = make_request(
+            self.reactor,
+            site=test_site,
+            method="GET",
+            path="/_matrix/client",
+        )
+        self.assertEqual(channel.code, 302)
```
DMRobertson, line 93: Could we have an `HTTPStatus....` here? I can never remember what anything other than `200, 400, 404, 500` means. For consistency, ditto for the 200 below.

DMRobertson (changes requested):
Minor nit.

Sanity check: does this fail if run against 1.50.0?

tests/http/test_webclient.py
babolivier: fwiw most tests use integers for status codes (and so does most of the servlet code when specifying the status to respond with). Though a few bits of the code use `HTTPStatus`, so it looks like we're already inconsistent on this, so :shrug:

anoadragon453: I initially did this as I like to use literals in tests where possible (to ensure what we have in the code makes sense). However if these status codes ever change (unlikely) then presumably we'll want to update the tests anyhow.

anoadragon453:
> Sanity check: does this fail if run against 1.50.0?

It does indeed!

tests/http/test_webclient.py
anoadragon453: I updated the 302 to an `HTTPStatus.FOUND`. The status code name is pretty vague, so I added a comment on top as well.

clokep: > fwiw most tests use integers for status codes (and so does most of the servlet code when specifying the status to respond with). Though a few bits of the code use `HTTPStatus`, so it looks like we're already inconsistent on this, so 🤷

I think we've been slowly updating to use `HTTPStatus`. That seems to be @DMRobertson's plan at least.

DMRobertson: That's right; maybe I ought to just do it in one big bang if we're going to use them at all.

Being fully candid, I don't have a justification for this except for "I think it's nicer". Happy to drop this if people find it noisey or not useful.

DMRobertson (approved):
LGTM!

tests/http/test_webclient.py
anoadragon453: I think it's nicer and more useful than the status code alone. And if one wants to know the status code, they can simply check the definition.

PR #11766, opened 2022-01-18
Allow overriding the complement ref

nico-famedly:
Allows you to define COMPLEMENT_REF to use a different branch or commit
than master, when running complement and not using a local
COMPLEMENT_DIR.

Signed-off-by: Nicolas Werner <n.werner@famedly.com>

### Pull Request Checklist

### DIFF
--- a/scripts-dev/complement.sh
+++ b/scripts-dev/complement.sh
@@ -33,7 +33,7 @@ cd "$(dirname $0)/.."
 # Check for a user-specified Complement checkout
 if [[ -z "$COMPLEMENT_DIR" ]]; then
   COMPLEMENT_REF=${COMPLEMENT_REF:-master}
-  echo "COMPLEMENT_DIR not set. Fetching the latest Complement checkout..."
+  echo "COMPLEMENT_DIR not set. Fetching Complement checkout from ${COMPLEMENT_REF}..."
   wget -Nq https://github.com/matrix-org/complement/archive/${COMPLEMENT_REF}.tar.gz
   tar -xzf ${COMPLEMENT_REF}.tar.gz
   COMPLEMENT_DIR=complement-${COMPLEMENT_REF}
--- a/scripts-dev/complement.sh
+++ b/scripts-dev/complement.sh
@@ -8,7 +8,8 @@
 # By default the script will fetch the latest Complement master branch and
 # run tests with that. This can be overridden to use a custom Complement
 # checkout by setting the COMPLEMENT_DIR environment variable to the
-# filepath of a local Complement checkout.
+# filepath of a local Complement checkout or by setting the COMPLEMENT_REF
+# environment variable to pull a different branch or commit.
 #
 # By default Synapse is run in monolith mode. This can be overridden by
 # setting the WORKERS environment variable.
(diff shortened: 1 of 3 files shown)


nico-famedly:
This makes it easier to override the complement version, when the current master branch of it is broken in our CI and I thought I'd share it, because it might be useful for someone else too. You can achieve the same by manually pulling the complement repo in your CI script and then checking out the right ref, but that is a bit more effort and harder to adapt as needed. Feel free to close it, if you think the current method of overriding the ref is already enough.

scripts-dev/complement.sh
```diff
@@ -31,11 +32,12 @@ cd "$(dirname $0)/.."
...
 if [[ -z "$COMPLEMENT_DIR" ]]; then
+  COMPLEMENT_REF=${COMPLEMENT_REF:-master}
   echo "COMPLEMENT_DIR not set. Fetching the latest Complement checkout..."
-  wget -Nq https://github.com/matrix-org/complement/archive/master.tar.gz
-  tar -xzf master.tar.gz
-  COMPLEMENT_DIR=complement-master
-  echo "Checkout available at 'complement-master'"
+  wget -Nq https://github.com/matrix-org/complement/archive/${COMPLEMENT_REF}.tar.gz
+  tar -xzf ${COMPLEMENT_REF}.tar.gz
+  COMPLEMENT_DIR=complement-${COMPLEMENT_REF}
```
clokep, line 39: We should probably update `.gitignore` to match this pattern?

scripts-dev/complement.sh
```diff
@@ -31,11 +32,12 @@ cd "$(dirname $0)/.."
 
 # Check for a user-specified Complement checkout
 if [[ -z "$COMPLEMENT_DIR" ]]; then
```
clokep, line 34: I'm curious why using `COMPLEMENT_DIR` won't work for you? Do you not want to pull the entire repo for some reason? (It is relatively small IIRC.)

nico-famedly: Well, it is easier to update a single environment variable than having to add "install git, clone repo, checkout right commit, set environment variable" to our CI. It's not like we can't do that, but having to enabled/disable that every time the upstream complement doesn't work with the current synapse release is more effort than just telling it to use the right commit. I also just figured out how to do that correctly, after I hacked in the support for specifying a ref and using a ref is less effort. It might also be useful in the future, if complement ever gets branches for synapse versions like "synapse-1.50" or so (because sometimes the branches are just incompatible).

scripts-dev/complement.sh
nico-famedly: Should be fixed.

scripts-dev/complement.sh
nico-famedly: I can also easily run the complement job in our CI with the complement ref overwritten, which makes it easier to figure out what commit actually works. We had to do that today, because the complement CA changes somehow don't work in our CI and we couldn't figure out why yet.

anoadragon453: > Well, it is easier to update a single environment variable than having to add "install git, clone repo, checkout right commit, set environment variable" to our CI.

Boiled down, this PR is the equivalent of adding the listed steps inside of `complement.sh`. One could also just easily add a "install git, clone repo, checkout right commit, set environment variable" bit to their dockerfile, and control which commit is checked out with another environment variable. And yes, that would work without adding another env var to `complement.sh`.

But this PR proposes upstreaming it as optional functionality into the script, so that every project doesn't need to set that up their own. It's a convenience that could save downstream developers time, so while it's not a necessary option (and is code we'll need to maintain), I think it's justifiable.

nico-famedly: It's also a really small change, because it just replaces the hardcoded master with an overrideable variable.

clokep: > It might also be useful in the future, if complement ever gets branches for synapse versions like "synapse-1.50" or so (because sometimes the branches are just incompatible).

We do branch matching in our CI for this reason, see: https://github.com/matrix-org/synapse/blob/c072c0b82969879cd08d0d6e669ded509f1e833b/.github/workflows/tests.yml#L342-L364

We don't use `complement.sh` directly for Synapse CI, I'm not sure it is really designed for such things. Anyway, I agree that the overall change is relatively minor.

scripts-dev/complement.sh
```diff
@@ -31,11 +32,12 @@ cd "$(dirname $0)/.."
...
+  COMPLEMENT_REF=${COMPLEMENT_REF:-master}
   echo "COMPLEMENT_DIR not set. Fetching the latest Complement checkout..."
```
clokep, line 36: We should probably clarify this log line too, what do you think of:

```suggestion
  echo "COMPLEMENT_DIR not set. Fetching Complement checkout from ${COMPLEMENT_REF}..."
```

scripts-dev/complement.sh
nico-famedly: Hm, I guess we should do that too at some point then. I was mostly happy we got it working at all :D

clokep (approved):

PR #11669, opened 2022-01-01
Fix preview of some gif URLs

blastrock:
Related to #11563

This PR fixes the preview of tenor.com URLs.

Before: 
image

After:
image

This was broken because oEmbed JSON from tenor.com looks like:

```json
{
  "type": "video",
  "version": "1.0",
  "provider_name": "Tenor",
  "provider_url": "https://tenor.com/",
  "cache_age": 86400,
  "html": "<iframe src=\"https://tenor.com/embed/4945328\" width=\"600\" height=\"400\" frameborder=\"0\"></iframe>",
  "width": 600,
  "height": 400
}
```

And we don't know how to handle the iframe. This PR fixes it by aborting the oEmbed preview and falling back to the Open Graph one (which I think was what was done before).

Note that this does not completely fix #11563 because imgur previews are still broken. The oEmbed looks like

```json
{
  "version": "1.0",
  "type": "rich",
  "provider_name": "Imgur",
  "provider_url": "https://imgur.com",
  "width": 540,
  "height": 500,
  "html": "<blockquote class=\"imgur-embed-pub\" lang=\"en\" data-id=\"a/Ojsfo1t\"><a href=\"https://imgur.com/a/Ojsfo1t\">self-quarantine New Years Celebration</a></blockquote><script async src=\"//s.imgur.com/min/embed.js\" charset=\"utf-8\"></script>"
}
```

There are multiple issues here:
1. The type is "rich", so we can't know if there is supposed to be an image or video in that preview.
2. The `html` does not contain the preview per-se, we need to interpret the HTML and JavaScript (that script sources another script!) to render it.
3. The Open Graph response does not contain the preview, only the imgur logo itself.

So this PR will not fix imgur previews.

### Pull Request Checklist

* [x] Pull request is based on the develop branch
* [x] Pull request includes a [changelog file](https://matrix-org.github.io/synapse/latest/development/contributing_guide.html#changelog). The entry should:
  - Be a short description of your change which makes sense to users. "Fixed a bug that prevented receiving messages from other servers." instead of "Moved X method from `EventStore` to `EventWorkerStore`.".
  - Use markdown where necessary, mostly for `code blocks`.
  - End with either a period (.) or an exclamation mark (!).
  - Start with a capital letter.
* [x] Pull request includes a [sign off](https://matrix-org.github.io/synapse/latest/development/contributing_guide.html#sign-off)
* [x] [Code style](https://matrix-org.github.io/synapse/latest/code_style.html) is correct
[...]

PR #11767, opened 2022-01-18
Support rendering previews with data: URLs in them

clokep:
I noticed while reviewing/testing #11669 that if a webpage has a `data:` on it (which we choose for the `og:image` response) then the URL preview breaks. An example page for this is https://www.reddit.com/r/matrixdotorg/comments/s5m47u/cant_connect_to_my_home_server_in_element/

The image which ends up being chosen is the reddit logo, which they embed as a `data:` URL. After this PR we get a working preview:

image

~~The only thing that is a bit weird about the current implementation is that it will allow Synapse to directly render `data:` URLs if requests, e.g. using [the "HTML document" example from MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URIs) and URL encoding it:~~  This is no longer true and only image or oEmbed URLs found in HTML would be parsed if they're `data:` URLs.

```sh
$ curl --header "Authorization: Bearer $TOK" http://localhost:8080/_matrix/media/r0/preview_url\?url\=data%3Atext%2Fhtml%2C%253Ch1%253EHello%252C%2520World%2521%253C%252Fh1%253E
{"og:title":"Hello, World!","og:description":"Hello, World!"}
```

~~I don't think there's any risk to doing this since we're not rendering JavaScript, but it feels a bit icky.~~

The code should probably handle if you embed something weird (e.g. HTML) as the `src` of the image, however. (Note that "other" schemes usually get rejected via [deep in Twisted internals for `Agent`](https://github.com/twisted/twisted/blob/release-21.2.0-10091/src/twisted/web/client.py#L1605-L1613)). Without this change the error is obscure since the `rebase_url` method is broken and you end up trying to preview `data:www.reddit.com/r/matrixdotorg/comments/s5m47u/cant_connect_to_my_home_server_in_element/image/png;base64,<the base 64 data>`.


synapse/rest/media/v1/preview_url_resource.py
```diff
@@ -367,7 +368,128 @@ async def _do_preview(self, url: str, user: UserID, ts: int) -> bytes:
 
         return jsonog.encode("utf8")
 
-    async def _download_url(self, url: str, user: UserID) -> MediaInfo:
+    async def _download_url(
+        self, url: str, output_stream: BinaryIO
+    ) -> Tuple[int, str, int, str, Optional[str], int, Optional[str]]:
```
clokep, line 373: Maybe this should be another type instead of just a massive tuple? 🤷

synapse/rest/media/v1/preview_url_resource.py
```diff
@@ -367,7 +368,128 @@ async def _do_preview(self, url: str, user: UserID, ts: int) -> bytes:
...
+        Returns:
+            A tuple of:
+                Media length, URL downloaded, the HTTP response code,
+                the media type, the downloaded file name, the number of
+                milliseconds the result is valid for, the etag header.
+        """
+
+        try:
+            logger.debug("Trying to get preview for url '%s'", url)
+            length, headers, uri, code = await self.client.get_file(
```
clokep, line 390: We might want to assert schemes here as additional error checking.

synapse/rest/media/v1/preview_url_resource.py
```diff
@@ -377,51 +499,28 @@ async def _download_url(self, url: str, user: UserID) -> MediaInfo:
...
+                (
+                    length,
+                    uri,
+                    code,
+                    media_type,
+                    download_name,
+                    expires,
+                    etag,
+                ) = await self._parse_data_url(url, f)
             else:
```
clokep, line 512: We might want to assert HTTP(S) schemes here as additional error checking.

clokep:
> The only thing that is a bit weird about the current implementation is that it will allow Synapse to directly render `data:` URLs if requests

We could probably avoid this with a flag, now that I think a bit more about it. So directly given URLs couldn't be `data:`. I'm not sure if this offers any real protection though.

synapse/rest/media/v1/preview_html.py
```diff
@@ -323,10 +323,13 @@ def _iterate_over_text(
...
+    # Add a scheme, if one does not exist.
+    if not url_parts[0]:
         url_parts[0] = base_parts[0] or "http"
-    if not url_parts[1]:  # fix up hostname
+    # Fix up the hostname, if this is not a data URL.
+    if url_parts[0] != "data" and not url_parts[1]:
         url_parts[1] = base_parts[1]
+        # If the path does not start with a /, nest it under the base path.
         if not url_parts[2].startswith("/"):
             url_parts[2] = re.sub(r"/[^/]+$", "/", base_parts[2]) + url_parts[2]
```
DMRobertson, line 334: `url_parts` seems to be a named tuple these days. I don't knowif you'd consider `url_parts.schema` an improvement over `url_parts[0]` though.

C.f. [docs](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlparse) and [typeshed](https://github.com/python/typeshed/blob/fe6233a8c3d4ca708a0da2ab281a20fbd148215c/stdlib/urllib/parse.pyi#L56-L62)

synapse/rest/media/v1/preview_html.py
```diff
@@ -323,10 +323,13 @@ def _iterate_over_text(
 def rebase_url(url: str, base: str) -> str:
```
DMRobertson, line 323: I think this function effectively:
- returns `url` unmodified if it is a fully specified, absolute URL
- otherwise resolves `url` as a relative URL against `base` and returns the resolved result

If so, a suggested docstring:

```python
"""
Resolves a potentially relative `URL` against an absolute `base` URL.

For example:
    >>> rebase_url("subpage", "https://example.com/foo/")
    'https://example.com/hello/subpage'
    >>> rebase_url("sibling", "https://example.com/foo")
    'https://example.com/sibling'
    >>> rebase_url("/hello", "https://example.com/foo/")
    'https://example.com/hello'
    >>> rebase_url("https://alice.com/a/", "https://bob.com/b/")
    'https://alice.com/a'
"""
```

synapse/rest/media/v1/preview_url_resource.py
```diff
@@ -367,7 +368,128 @@ async def _do_preview(self, url: str, user: UserID, ts: int) -> bytes:
...
+            raise SynapseError(
+                500,
+                "Failed to parse data URL: %s"
+                % (traceback.format_exception_only(sys.exc_info()[0], e),),
+                Codes.UNKNOWN,
+            )
+
+        length = output_stream.seek(1)
+
+        media_type = url_info.headers.get_content_type()
```
DMRobertson, line 469: Should we validate this against the mime type of the data URL? I don't know if that actually provides any security benefit though.

synapse/rest/media/v1/preview_url_resource.py
```diff
@@ -417,6 +418,38 @@ async def _download_url(
...
```
DMRobertson, line 469: Should we cross reference this with the mime-type in the data url?

(Do we want to have some kind of whitelist of trusted mime types which are okay to preview?)

synapse/rest/media/v1/preview_url_resource.py
```diff
@@ -367,7 +368,128 @@ async def _do_preview(self, url: str, user: UserID, ts: int) -> bytes:
...
+            logger.warning("Error parsing data: URL %s: %r", url, e)
+
+            raise SynapseError(
+                500,
+                "Failed to parse data URL: %s"
+                % (traceback.format_exception_only(sys.exc_info()[0], e),),
+                Codes.UNKNOWN,
+            )
+
+        length = output_stream.seek(1)
```
DMRobertson, line 467: Not sure I follow this. Isn't this seeking the stream to the first byte of the output, and returning the integer `1`? (Judging by [docs](https://docs.python.org/3/library/io.html#io.IOBase.seek)). If so, we might be making the same mistake [here](https://github.com/matrix-org/synapse/blob/d3cf0730f8bf3de58d37060eff0dca3f69b4a0e1/tests/server.py#L316).

Should this be [`output_stream.tell()`](https://docs.python.org/3/library/io.html#io.IOBase.tell)?

DMRobertson (commented):
A few drive-by comments.

Not sure of the security ramifications here; I'll stick into the queue for a second opinion. But FWIW: instead of downloading an arbitrary data blob from someone else, we have an arbitrary binary blob provided to us via a URL---is there any difference? If Synapse itself was processing the data blob generically, then I could forsee problems like the billion laughs attack or some invalid image data designed to trip up an image renderer.

I agree with you on the ickiness. It does seem... odd and confused to allow someone to say "please preview this data that I already have".

synapse/rest/media/v1/preview_html.py
clokep: I attempted to change this code, but we actually modify the pieces, which is why it is a list. 🤷 Not converting `base_parts` to a list and using the named properties might be clearer though?

synapse/rest/media/v1/preview_url_resource.py
clokep: This *is* the mime-type from the data URL, it gets parsed out for us and shoved into a `headers` property for some reason.

[The docs for `urlopen`](https://docs.python.org/3/library/urllib.request.html#urllib.request.urlopen) have some info on how `data` URLs are handled:

> For FTP, file, and data URLs [...], this function returns a [`urllib.response.addinfourl`](https://docs.python.org/3/library/urllib.request.html#urllib.response.addinfourl) object.

The `addinfourl` object has a `headers` property:

> Returns the headers of the response in the form of an [`EmailMessage`](https://docs.python.org/3/library/email.message.html#email.message.EmailMessage.get_content_type) instance.

The `EmailMessage` object has a [`get_content_type()`](https://docs.python.org/3/library/email.message.html#email.message.EmailMessage.get_content_type) property:

> Return the message’s content type, coerced to lower case of the form *maintype/subtype*. If there is no *Content-Type* header in the message return the value returned by [`get_default_type()`](https://docs.python.org/3/library/email.message.html#email.message.EmailMessage.get_default_type). If the *Content-Type* header is invalid, return `text/plain`.

This all doesn't really make sense for `data` URLs, but from trial and error it seems to do the correct thing.

synapse/rest/media/v1/preview_url_resource.py
clokep: Yes, this should definitely be using `tell()`! 🎉

synapse/rest/media/v1/preview_html.py
clokep: Looks like we don't have tests for this function either, I'll add some.

synapse/rest/media/v1/preview_html.py
clokep: I made the change to `base_parts`.

synapse/rest/media/v1/preview_url_resource.py
clokep: I added a clarifying comment in 116015f0315f65908f296a010c9f29a767556a8c -- hopefully that helps?

synapse/rest/media/v1/preview_url_resource.py
clokep: I made the change here since it seems to just be a minor wart in our testing framework.

clokep:
> Not sure of the security ramifications here; I'll stick into the queue for a second opinion. But FWIW: instead of downloading an arbitrary data blob from someone else, we have an arbitrary binary blob provided to us via a URL---is there any difference? If Synapse itself was processing the data blob generically, then I could forsee problems like the billion laughs attack or some invalid image data designed to trip up an image renderer.
> 
> I agree with you on the ickiness. It does seem... odd and confused to allow someone to say "please preview this data that I already have".

Thanks for the review! I agree that it isn't too different from what we're doing. It is worth investigating how `urlopen` would work in some situations though (I'll take a look at that!) In particular this might not handle our max download size properly...but if it is embedded in something we already downloaded then by definition it must be smaller than the max download size. I think.

I've updated the PR to *not* allow directly previewing `data:` URLs because I don't think there's a valid use-case there.

synapse/rest/media/v1/preview_url_resource.py
clokep: I fixed this in b9a1c66814c3de9334c45ba7cd1dd64850ffa58c and added a test for it in d38a07ccbd0c6a2ba70ba3a8f6fde3a15bf0a4b6.

synapse/rest/media/v1/preview_html.py
```diff
@@ -321,14 +321,33 @@ def _iterate_over_text(
 
 
 def rebase_url(url: str, base: str) -> str:
-    base_parts = list(urlparse.urlparse(base))
+    """
+    Resolves a potentially relative `url` against an absolute `base` URL.
```
reivilibre, line 325: Do you know how this differs from [`urljoin`](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urljoin) in the standard lib?
Not necessarily opposed to rolling our own, but it'd be nice to know if we had specific motivation for doing so.

synapse/rest/media/v1/preview_url_resource.py
reivilibre: Yeeeees, that'd be nice, please do :-)

synapse/rest/media/v1/preview_url_resource.py
reivilibre: Yeah I feel like I'd be happier having a few select allowed protocols rather than singling out `data:`; not sure it really makes sense to ask us to preview an `ftp:` URI either for example.

reivilibre (changes requested):
Seems sensible enough — just a few points

synapse/rest/media/v1/preview_url_resource.py
clokep: Yep, note that those *will* fail right now (we won't try them), but it would be better to be explicit, I think!

synapse/rest/media/v1/preview_html.py
clokep: I didn't know of it, but I'll look into it!

synapse/rest/media/v1/preview_url_resource.py
clokep: I think this might actually be tougher to get correct than I initially though since we seem to support previewing scheme-less URLs. I'm not sure it makes sense to refactor that as part of this.

synapse/rest/media/v1/preview_html.py
clokep: I *think* this will work OK, but I'd rather punt this to a separate PR since it is pretty unrelated to this work. Would that be OK?

reivilibre: yeah OK

synapse/rest/media/v1/preview_url_resource.py
reivilibre: Yeah, fair enough.

reivilibre (approved):
Thanks!

Issue #11769, opened 2022-01-19
AttributeError: 'RegistrationHandler' object has no attribute 'pusher_pool'


anoadragon453:
When testing out commit d1e6333f12d8a121c649c6176e0ac5e915345366 on matrix.org today, we ended up with the following stacktrace:

<details>

<summary>stacktrace</summary>

```
2022-01-19 13:25:46,124 - synapse.http.server - 165 - ERROR - GET-847 - Failed handle request <XForwardedForRequest at 0x7f01352cf198 method='GET' uri='/_synapse/client/sso_register' clientproto='HTTP/1.1' site='8087'>
Capture point (most recent call last):
  File "/usr/local/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/local/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/synapse/src/synapse/app/generic_worker.py", line 514, in <module>
    main()
  File "/home/synapse/src/synapse/app/generic_worker.py", line 510, in main
    start(sys.argv[1:])
  File "/home/synapse/src/synapse/app/generic_worker.py", line 505, in start
    _base.start_worker_reactor("synapse-generic-worker", config)
  File "/home/synapse/src/synapse/app/_base.py", line 126, in start_worker_reactor
    run_command=run_command,
  File "/home/synapse/src/synapse/app/_base.py", line 179, in start_reactor
  File "/home/synapse/src/synapse/app/_base.py", line 163, in run
  File "/home/synapse/env-py37/lib/python3.7/site-packages/twisted/internet/base.py", line 1318, in run
    self.mainLoop()
  File "/home/synapse/env-py37/lib/python3.7/site-packages/twisted/internet/base.py", line 1328, in mainLoop
    reactorBaseSelf.runUntilCurrent()
  File "/home/synapse/env-py37/lib/python3.7/site-packages/twisted/internet/base.py", line 967, in runUntilCurrent
    f(*a, **kw)
  File "/home/synapse/env-py37/lib/python3.7/site-packages/twisted/internet/defer.py", line 662, in callback
    self._startRunCallbacks(result)
  File "/home/synapse/env-py37/lib/python3.7/site-packages/twisted/internet/defer.py", line 764, in _startRunCallbacks
    self._runCallbacks()
  File "/home/synapse/env-py37/lib/python3.7/site-packages/twisted/internet/defer.py", line 859, in _runCallbacks
[...]

Issue #10882, opened 2021-09-22
Email notifications are not automatically enabled when authenticating via SSO


manning-ncsa:
### Description

Users that authenticate via oidc_provider do not have email notifications enabled by default, even though the server is configured correctly and they have an email address associated with their account.

The [problem lies here](https://github.com/matrix-org/synapse/blob/b3590614da7e3e17e75530a9d4808df17be9b127/synapse/handlers/sso.py#L949-L951), where `access_token` is set to `None` in `post_registration_actions()`. This [leads to the call](https://github.com/matrix-org/synapse/blob/b3590614da7e3e17e75530a9d4808df17be9b127/synapse/handlers/register.py#L861) to `_register_email_threepid()` with a null `access_token`, which in turn causes [the condition here](https://github.com/matrix-org/synapse/blob/b3590614da7e3e17e75530a9d4808df17be9b127/synapse/handlers/register.py#L913-L917) to evaluate to false, thus preventing the [email notification pusher](https://github.com/matrix-org/synapse/blob/b3590614da7e3e17e75530a9d4808df17be9b127/synapse/handlers/register.py#L927-L937) from being added.

### Steps to reproduce

- Configure Synapse to enable email notifications for new users: `email.notif_for_new_users: true` and `email.enable_notifs: true`.
- Log in a user for the first time using a custom oidc_provider (in our case Keycloak).
- Verify that they automatically have an email address associated with their account.
- Open the user settings and see that email notifications are disabled.

I expect the email notifications to be enabled upon first login.

### Proposed Solution

One solution to this is to include `LoginType.SSO` in the `auth_result` object passed to the `post_registration_actions()` function [here](https://github.com/matrix-org/synapse/blob/b3590614da7e3e17e75530a9d4808df17be9b127/synapse/handlers/sso.py#L944-L947) so that [prior to the `_register_email_threepid()` call](https://github.com/matrix-org/synapse/blob/b3590614da7e3e17e75530a9d4808df17be9b127/synapse/handlers/register.py#L861) the `access_token` could be set to `True` if `auth_result[LoginType.SSO] == True`. (Disclaimer: I have not tried this to verify it is a good solution. The idea came from discussion with @kyrias.)

### Version information

- **Homeserver**: independent self-hosted homeserver

If not matrix.org:

- **Version**: `"server_version":"1.42.0", "python_version":"3.8.12"`

[...]

Issue #10699, opened 2021-08-25
The `synapse_review_recent_signups` script should have a flag to ignore and filter appservice users


maranda:
As the topic says, a server whit bridges would otherwise have a pretty polluted output.

lukasdenk:
I'll work on it :)

lukasdenk:
@maranda: I solved the issue in the above PR. Could you please close the issue? Thanks :)

PR #11675, opened 2022-01-03
Add a flag to the `synapse_review_recent_signups` script to ignore and filter appservice users.

lukasdenk:
Solve issue #10699

Adds a flag to the `synapse_review_recent_signups` script to ignore and filter appservice users.

**Signed-off-by:** Lukas Denk <lukasdenk@web.de>

### Pull Request Checklist

* [x] Pull request is based on the develop branch
* [x] Pull request includes a [changelog file](https://matrix-org.github.io/synapse/latest/development/contributing_guide.html#changelog). The entry should:
  - Be a short description of your change which makes sense to users. "Fixed a bug that prevented receiving messages from other servers." instead of "Moved X method from `EventStore` to `EventWorkerStore`.".
  - Use markdown where necessary, mostly for `code blocks`.
  - End with either a period (.) or an exclamation mark (!).
  - Start with a capital letter.
* [x] Pull request includes a [sign off](https://matrix-org.github.io/synapse/latest/development/contributing_guide.html#sign-off)
* [x] [Code style](https://matrix-org.github.io/synapse/latest/code_style.html) is correct
  (run the [linters](https://matrix-org.github.io/synapse/latest/development/contributing_guide.html#run-the-linters))

PR #11770, opened 2022-01-19
Partially revert #11675; prevent attempting to create pushers on workers

anoadragon453:
Closes https://github.com/matrix-org/synapse/issues/11769. Necessary context is in the issue.

### DIFF
--- a/changelog.d/11770.feature
+++ b/changelog.d/11770.feature
@@ -1,0 +1,1 @@
+Add a flag to the `synapse_review_recent_signups` script to ignore and filter appservice users.
--- a/synapse/handlers/register.py
+++ b/synapse/handlers/register.py
@@ -979,18 +979,16 @@ class RegistrationHandler:
         if (
             self.hs.config.email.email_enable_notifs
             and self.hs.config.email.email_notif_for_new_users
+            and token
         ):
             # Pull the ID of the access token back out of the db
             # It would really make more sense for this to be passed
             # up when the access token is saved, but that's quite an
             # invasive change I'd rather do separately.
-            if token:
-                user_tuple = await self.store.get_user_by_access_token(token)
-                # The token better still exist.
-                assert user_tuple
-                token_id = user_tuple.token_id
-            else:
-                token_id = None
+            user_tuple = await self.store.get_user_by_access_token(token)
+            # The token better still exist.
+            assert user_tuple
+            token_id = user_tuple.token_id
 
             await self.pusher_pool.add_pusher(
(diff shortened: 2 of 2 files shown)


richvdh:
so does this reopen #10699? (if not, why was this change in #11769 in the first place?)

anoadragon453:
> so does this reopen #10699?

It does not. Testing the branch of this PR on my homeserver with:

```
synapse_review_recent_signups -c homeserver.yaml -a
```

I find that appservice users are still properly excluded when `-a` is passed.

> (if not, why was this change in #11675 in the first place?)

I'm not sure. It's not necessary to fix #10699, and was not suggested during review. My theory is that it was an intended fix for a separate bug (`_register_email_threepid` being called on workers) that got mixed in (@lukasdenk?).

anoadragon453:
> It would be nice to at least test this manually though.

Doing so now.

anoadragon453:
That took a while, but on my personal server I was able to confirm that if you attempt to use `/register` with a email/password via a worker process:

* on `develop`, the entire flow will fail.
* on `anoa/pusher_regression`, the user will be registered, but they won't end up getting a pusher assigned to send them email notifications.

So the failure case (not being able to register) was worse than I thought it was (I assumed you just wouldn't get email notifications). The latter being broken has been the case forever, and is more preferable than registration being broken. So yes, this PR seems to take us back to the status quo (though it too needs fixing).

anoadragon453:
Registration tests via a worker would be good to add, yes. Though writing worker-based tests are quite a faff, so I suggest we get those in for the proper "fix email notifs when registering with pushers" PR.

anoadragon453:
> (if not, why was this change in #11675 in the first place?)

It looks like this may have been an attempt to fix https://github.com/matrix-org/synapse/issues/10882. I would encourage contributors to separate out fixes into separate PRs to avoid confusion during review!

DMRobertson (approved):
Thanks for going through the motions and testing it manually!

anoadragon453:
It was a good suggestion. You can never be too careful with regressions!

lukasdenk:
> I'm not sure. It's not necessary to fix #10699, and was not suggested during review. My theory is that it was an intended fix for a separate bug (`_register_email_threepid` being called on workers) that got mixed in (@lukasdenk?).

Sorry, that was a mistake. I mistakenly introduced some changes from another issue I worked on (#10882) in that PR.

anoadragon453:
@lukasdenk It happens! We should have caught it in review, but we're all only human.

PR #11771, opened 2022-01-19
Fix mypy for platforms without epoll support.

clokep:
Follow-up to #11724 to fix mypy on macOS.

### DIFF
--- a/changelog.d/11771.misc
+++ b/changelog.d/11771.misc
@@ -1,0 +1,1 @@
+Improve accuracy of `python_twisted_reactor_tick_time` prometheus metric.
--- a/synapse/metrics/_reactor_metrics.py
+++ b/synapse/metrics/_reactor_metrics.py
@@ -35,7 +35,7 @@ tick_time = Histogram(
 class EpollWrapper:
     """a wrapper for an epoll object which records the time between polls"""
 
-    def __init__(self, poller: "select.epoll"):
+    def __init__(self, poller: "select.epoll"):  # type: ignore[name-defined]
         self.last_polled = time.time()
         self._poller = poller
 
@@ -71,7 +71,7 @@ try:
     # if the reactor has a `_poller` attribute, which is an `epoll` object
     # (ie, it's an EPollReactor), we wrap the `epoll` with a thing that will
     # measure the time between ticks
-    from select import epoll
+    from select import epoll  # type: ignore[attr-defined]
 
     poller = reactor._poller  # type: ignore[attr-defined]
 except (AttributeError, ImportError):


clokep:
Without this change I get:

```
synapse/metrics/_reactor_metrics.py:38: error: Name "select.epoll" is not defined  [name-defined]
synapse/metrics/_reactor_metrics.py:74: error: Module "select" has no attribute "epoll"; maybe "poll"?  [attr-defined]
```

DMRobertson (commented):
I think this means that `reveal_type(select.epoll)` will be `Any` on macos, if we were to run mypy under not-linux?

Slightly unsettled at the idea that the typechecking result is implicitly dependent on your platform, but this is Python I suppose (and maybe we have that problem anyway with the choice of python version?).

Typeshed seems to [do something similar](https://github.com/python/typeshed/blob/fe6233a8c3d4ca708a0da2ab281a20fbd148215c/stdlib/select.pyi#L102-L103).

I'd like to double check that the `'type: ignore` doesn't lose information when we run mypy under linux (in particular, under CI). Let me checkout the PR...

DMRobertson:
> I'd like to double check that the `'type: ignore` doesn't lose information when we run mypy under linux

On my machine:

```
dmr on titan in synapse on  clokep/macos-mypy via 🐍 v3.10.1 (env) 
2022-01-19 16:29:44 ✔  $ mypy
synapse/metrics/_reactor_metrics.py:47: note: Revealed type is "select.epoll"

dmr on titan in synapse on  clokep/macos-mypy via 🐍 v3.10.1 (env) took 3s 
2022-01-19 16:30:33 ✔  $ git diff -- synapse
  diff --git a/synapse/metrics/_reactor_metrics.py b/synapse/metrics/_reactor_metrics.py
  index f38f79831..cf1f69536 100644
  --- a/synapse/metrics/_reactor_metrics.py
  +++ b/synapse/metrics/_reactor_metrics.py
  @@ -44,7 +44,7 @@ def poll(self, *args, **kwargs) -> List[Tuple[int, int]]:  # type: ignore[no-unt
           # how long it takes to run everything in the reactor - ie, how long anything
           # waiting for the next tick will have to wait.
           tick_time.observe(time.time() - self.last_polled)
  -
  +        reveal_type(self._poller)
           ret = self._poller.poll(*args, **kwargs)
   
           self.last_polled = time.time()
```

So I think this LGTM!

DMRobertson (approved):
It might be worth a comment to justify the type-ignore, noting that `select.epoll` is only available on Linux? Maybe that's overkill. I'll leave it to your discretion!

clokep:
> It might be worth a comment to justify the type-ignore, noting that `select.epoll` is only available on Linux? Maybe that's overkill. I'll leave it to your discretion!

I think it is pretty implicit from the attribute / import checks in those methods.

PR #11774, opened 2022-01-19
Add deprecation warnings for `webclient` listener and non-HTTP(S) `web_client_location`.

clokep:
This ends up having a slight behavior change: if `web_client_location` points to an HTTP(S) URL than `/` will now redirect to it, regardless of whether `webclient` is a configured listener.

As part of this I removed the documentation in the sample config for `webclient` listeners.

First part of #9078.

### DIFF
--- a/synapse/app/homeserver.py
+++ b/synapse/app/homeserver.py
@@ -132,8 +132,10 @@ class SynapseHomeServer(HomeServer):
         self._module_web_resources_consumed = True
 
         # try to find something useful to redirect '/' to
-        if WEB_CLIENT_PREFIX in resources:
-            root_resource: Resource = RootOptionsRedirectResource(WEB_CLIENT_PREFIX)
+        if self.config.server.web_client_location_is_redirect:
+            root_resource: Resource = RootOptionsRedirectResource(
+                self.config.server.web_client_location
+            )
         elif STATIC_PREFIX in resources:
             root_resource = RootOptionsRedirectResource(STATIC_PREFIX)
         else:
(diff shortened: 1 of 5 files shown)


synapse/app/homeserver.py
```diff
@@ -132,8 +132,10 @@ def _listener_http(
         self._module_web_resources_consumed = True
 
         # try to find something useful to redirect '/' to
-        if WEB_CLIENT_PREFIX in resources:
-            root_resource: Resource = RootOptionsRedirectResource(WEB_CLIENT_PREFIX)
+        if self.config.server.web_client_location_is_redirect:
```
DMRobertson, line 135: Can you spell out why these two are equivalent? I'm not following (but not sure how the `resources` stuff works.

synapse/config/server.py
```diff
@@ -1349,9 +1349,15 @@ def parse_listener_def(listener: Any) -> ListenerConfig:
     return ListenerConfig(port, bind_addresses, listener_type, tls, http_config)
 
 
+NO_MORE_NONE_HTTP_WEB_CLIENT_LOCATION_WARNING = """
+Synapse no longer supports serving a web client. To remove this warning, ensure
```
DMRobertson, line 1353: I think the "ensure" might be unneeded here?

DMRobertson (changes requested):
A few small things:

synapse/app/homeserver.py
clokep: They are not equivalent, as i mentioned in the description:

> if `web_client_location` points to an HTTP(S) URL than `/` will now redirect to it, regardless of whether `webclient` is a configured listener.

synapse/config/server.py
clokep: Yeahhhh I must have reworded this sentence or something. Thanks!

```suggestion
Synapse no longer supports serving a web client. To remove this warning,
```

DMRobertson (approved):
LGTM!

synapse/app/homeserver.py
clokep: I *think* this is acceptable and what we're going for, but I'm not 100% sure. 👍

PR #11775, opened 2022-01-19
Fix a bug that corrupted the cache of federated space hierarchies

### LINKED ISSUE #11773: Subspaces are randomly empty in space hierarchy (over federation?)
### Description

When browsing a space in Element Web, the children of subspaces will randomly be excluded. Repeatedly viewing the same space will get you different missing children randomly.

I've looked at the network responses in the browser and rooms are definitely missing from the homeserver response, so it's not Element Web's doing.

Something to do with federation maybe? I've not noticed it when the entire space is on the same homeserver as my user.

### Steps to reproduce

1. Create a space with lots of nested subspaces on one homeserver.
2. Using a different homeserver, eg. matrix.org, view the space hierarchy in Element Web repeatedly.
3. Notice runs of subspaces appearing with 0 rooms randomly.

### Version information

- **Homeserver**: matrix.org: 8e8a00829fcab64397705146570bda7082a21075
- **Homeserver**: homeserver with spaces: 1.49.2

squahtx:
`FederationClient.get_room_hierarchy()` caches its return values, so
avoid modifying the returned room summary and clone it instead.

Fixes #11773.

### DIFF
--- a/synapse/handlers/room_summary.py
+++ b/synapse/handlers/room_summary.py
@@ -780,6 +780,7 @@ class RoomSummaryHandler:
         try:
             (
                 room_response,
+                children_state_events,
                 children,
                 inaccessible_children,
             ) = await self._federation_client.get_room_hierarchy(
@@ -803,10 +804,8 @@ class RoomSummaryHandler:
             if "room_id" in c and isinstance(c["room_id"], str)
         }
 
-        room_response = dict(room_response)
-        children_state = room_response.pop("children_state", ())
         return (
-            _RoomEntry(room_id, room_response, children_state),
+            _RoomEntry(room_id, room_response, children_state_events),
             children_by_room_id,
             set(inaccessible_children),
         )
--- a/synapse/handlers/room_summary.py
+++ b/synapse/handlers/room_summary.py
@@ -803,8 +803,10 @@ class RoomSummaryHandler:
             if "room_id" in c and isinstance(c["room_id"], str)
         }
 
+        room_response = dict(room_response)
+        children_state = room_response.pop("children_state", ())
         return (
-            _RoomEntry(room_id, room_response, room_response.pop("children_state", ())),
+            _RoomEntry(room_id, room_response, children_state),
             children_by_room_id,
             set(inaccessible_children),
         )
--- a/changelog.d/11775.bugfix
+++ b/changelog.d/11775.bugfix
@@ -1,0 +1,1 @@
+Fix a long-standing bug where space hierarchy over federation would only work correctly some of the time.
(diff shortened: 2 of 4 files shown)


synapse/handlers/room_summary.py
```diff
@@ -803,8 +803,10 @@ async def _summarize_remote_room_hierarchy(
             if "room_id" in c and isinstance(c["room_id"], str)
         }
 
+        room_response = dict(room_response)
+        children_state = room_response.pop("children_state", ())
         return (
-            _RoomEntry(room_id, room_response, room_response.pop("children_state", ())),
```
squahtx, line 807: room_response is a JsonDict, which is why .pop() is allowed.

This bug feels like something that mypy could have prevented if we had a JsonMapping type. Except we can't just define it as a `Mapping`, since Mappings aren't necessarily JSON-serializable.

DMRobertson: Maybe https://docs.python.org/3.10/library/types.html#types.MappingProxyType, but we'd have to register something to make those json encodable. NewType probably the best way forward, short of some kind of `const/mutable` semantics in the type system.

synapse/handlers/room_summary.py
```diff
@@ -803,8 +803,10 @@ async def _summarize_remote_room_hierarchy(
             if "room_id" in c and isinstance(c["room_id"], str)
         }
 
+        room_response = dict(room_response)
+        children_state = room_response.pop("children_state", ())
```
clokep, line 807: Would it make more sense to make this a separate return value from `get_room_hierarchy` so that we do no modifications in here (and no copies)?

clokep:
Good find, by the way! 👍 Do you think this is worth a regression test?

synapse/handlers/room_summary.py
squahtx: Fewer extra copies sounds good!

squahtx:
> Do you think this is worth a regression test?

Probably a good idea, added one.

callahad:
Throwing a release blocker label on this to ensure it gets into 1.51.0rc1

anoadragon453 (approved):
Looks like existing comments have been addressed, and there are none more from me. Good find! Caching bugs can be notoriously tricky to root out.

squahtx:
Thank you both for reviewing!

I'm going to merge to unblock the rc. @clokep if you spot anything I've missed, let me know and I'll PR a fix.

clokep:
> I'm going to merge to unblock the rc. @clokep if you spot anything I've missed, let me know and I'll PR a fix.

Seems reasonable to me! 🎉

PR #11776, opened 2022-01-19
Comments and typing for `_update_outliers_txn`

richvdh:
A couple of surprises for me here, so thought I'd document them

### DIFF
--- a/synapse/storage/databases/main/events.py
+++ b/synapse/storage/databases/main/events.py
@@ -1254,20 +1254,22 @@ class PersistEventsStore:
         for room_id, depth in depth_updates.items():
             self._update_min_depth_for_room_txn(txn, room_id, depth)
 
-    def _update_outliers_txn(self, txn, events_and_contexts):
+    def _update_outliers_txn(
+        self,
+        txn: LoggingTransaction,
+        events_and_contexts: List[Tuple[EventBase, EventContext]],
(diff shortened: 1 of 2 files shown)


synapse/storage/databases/main/events.py
```diff
@@ -1285,15 +1289,22 @@ def _update_outliers_txn(self, txn, events_and_contexts):
...
-                # was an outlier or not.
+                # If the incoming event is rejected then we don't care if the event
+                # was an outlier or not - what we have is at least as good.
                 continue
 
             outlier_persisted = have_persisted[event.event_id]
             if not event.internal_metadata.is_outlier() and outlier_persisted:
                 # We received a copy of an event that we had already stored as
                 # an outlier in the database. We now have some state at that
                 # so we need to update the state_groups table with that state.
```
clokep, line 1300: (I know you didn't edit this sentence, but...) this feels like it is missing a word "We now have state at that so": that *what*? Maybe:

> We now have some state at that event, so we need to update the state_groups table with that state.

clokep (commented):
I'm not sure of the ramifications of this, as you note, but it seems to be true.

synapse/storage/databases/main/events.py
richvdh: sure, done.

clokep (approved):

PR #11782, opened 2022-01-20
Refuse to start on Python 3.6.

DMRobertson:
Follow up to #11595.


clokep:
I think this overlaps with #11683?

DMRobertson:
Ahh, you're quire right. I was confusing #11683 with #11595!

PR #11784, opened 2022-01-20
Apply a timeout to reading the body when fetching a file.

### LINKED ISSUE #8302: Synapse doesn't limit data/time for http previews
### Description

I have an Icecast stream running and a while back decided to send a link to it to someone over Matrix. A couple of days later I check the admin panel and noticed the following:
image
It seems that, while attempting to make the preview for the url, synapse started listening to the stream and it never stopped, because the stream never ends. Now this is a very specific case which could be solved by me properly disallowing preview fetching on my home network, but imagine if there were other homeservers in that room.

This points to the fact that synapse doesn't properly limit itself when trying to generate http previews, causing it to keep busy forever. In fact, this seems to happen multiple times as at the time of writing there are 3 synapse clients listening to the stream. 

### Steps to reproduce

- Have something with an infinite stream (such as http radio)
- Post a link to this stream in a room
- Any server having previews enabled will now be listening to the stream indefinitely. 

<!--

``` (three backticks, on a line on their own), so that they are formatted legibly.
-->
### Expected Behaviour

It would probably be best if Synapse kills the "thread" when it notices it can't make the preview/it takes too long. Looking at the logs, it seems Synapse does notice that the stream is too large, but it still keeps listening:

```
2020-09-12 10:29:43,512 - synapse.http.client - 339 - INFO - GET-3961617 - Received response to GET https://mkmkm.thebias.nl/: 200
2020-09-12 10:33:34,790 - synapse.http.server - 75 - INFO - GET-3961617 - <XForwardedForRequest at 0x7f5c555f3710 method='GET' uri='/_matrix/media/r0/preview_url?url=https%3A%2F%2Fmkmkm.thebias.nl%2F&ts=1599673680000' clientproto='HTTP/1.0' site=8008> SynapseError: 502 - Requested file is too large > 10485760 bytes
2020-09-12 10:33:34,790 - synapse.http.server - 531 - WARNING - GET-3961617 - Not sending response to request <XForwardedForRequest at 0x7f5c555f3710 method='GET' uri='/_matrix/media/r0/preview_url?url=https%3A%2F%2Fmkmkm.thebias.nl%2F&ts=1599673680000' clientproto='HTTP/1.0' site=8008>, already disconnected.
2020-09-12 10:33:34,791 - synapse.access.http.8008 - 311 - INFO - GET-3961617 - 2001:67c:2564:331:5d2:b8cb:8cf7:a474 - 8008 - {@rafael:thebias.nl} Processed request: 231.856sec/-171.857sec (0.004sec, 0.000sec) (0.000sec/0.002sec/1) 0B 200! "GET /_matrix/media/r0/preview_url?url=https%3A%2F%2Fmkmkm.thebias.nl%2F&ts=1599673680000 HTTP/1.0" "Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:81.0) Gecko/20100101 Firefox/81.0" [0 dbevts]
```

### Version information

- **Homeserver**: Personal Homeserver

<!--

listen on that port.)
-->
- **Version**: 1.19.1

- **Install method**: debian package

- **Platform**: debian 10
<!--
-->

### PR DESCRIPTION
**Pull request by clokep**
Fixes #8302.

Applies a timeout to reading the body for `SimpleHttpClient.get_file()`. This brings the code more inline with `MatrixFederationHttpClient.get_file`, although that uses `addTimeout` instead of `timeout_deferred`.

I tested this by attempting to preview a stream (I used a local radio station), waiting several minutes and noticing the file was still growing. With this patch the download stops after 60 seconds and the client receives an error message:

```json
{"errcode":"M_TOO_LARGE","error":"Requested file took too long to download"}
```

We are rather inconsistent with using timeouts when reading bodies, we seem to do it in some places, but not many other places. It is unclear if this is on purpose or not though. This particular method is only used by URL previews.

### DIFF
--- a/changelog.d/11784.bugfix
+++ b/changelog.d/11784.bugfix
@@ -1,0 +1,1 @@
+Fix a long-standing bug that media streams could cause long-lived connections when generating URL previews.
--- a/synapse/http/client.py
+++ b/synapse/http/client.py
@@ -734,7 +734,7 @@ class SimpleHttpClient:
             d = read_body_with_max_size(response, output_stream, max_size)
 
             # Ensure that the body is not read forever.
-            d = timeout_deferred(d, 60, self.hs.get_reactor())
+            d = timeout_deferred(d, 30, self.hs.get_reactor())
 
             length = await make_deferred_yieldable(d)
         except BodyExceededMaxSize:
--- a/synapse/http/client.py
+++ b/synapse/http/client.py
@@ -731,13 +731,22 @@ class SimpleHttpClient:
         # straight back in again
 
         try:
-            length = await make_deferred_yieldable(
-                read_body_with_max_size(response, output_stream, max_size)
-            )
+            d = read_body_with_max_size(response, output_stream, max_size)
+
+            # Ensure that the body is not read forever.
+            d = timeout_deferred(d, 60, self.hs.get_reactor())
+
+            length = await make_deferred_yieldable(d)
         except BodyExceededMaxSize:
             raise SynapseError(
                 HTTPStatus.BAD_GATEWAY,
                 "Requested file is too large > %r bytes" % (max_size,),
+                Codes.TOO_LARGE,
+            )
+        except defer.TimeoutError:
+            raise SynapseError(
+                HTTPStatus.BAD_GATEWAY,
+                "Requested file took too long to download",
                 Codes.TOO_LARGE,
             )
         except Exception as e:


anarcat:
60 seconds is still a long time. shouldn't there be a byte size limit instead? or better: both?

clokep:
> shouldn't there be a byte size limit instead? or better: both?

There is also a byte-size limit, it is implemented inside of the `read_body_with_max_size`, the actual limit is configurable, but defaults to 10 megabytes, I believe.

reivilibre (approved):
Sounds sensible

dkasak:
FWIW, my instinct is also to make this shorter, probably closer to 10 seconds. HTTP requests usually don't take this long to complete and failure is not catastrophic anyway given that this is just for URL previews.

clokep:
> FWIW, my instinct is also to make this shorter, probably closer to 10 seconds. HTTP requests usually don't take this long to complete and failure is not catastrophic anyway given that this is just for URL previews.

I chose 60 seconds fairly arbitrarily (and it matched some other timeouts in our code). 10 seconds seems rather short if a large image is being previewed, but maybe not (for reference, what bandwidth would it take to hit the 10M limit at a certain time -- 10s: 1000 kbps vs. 60s: 167 kbps).

I'm also a bit worried that some people might have cranked the max preview size up even more than the 10M limit. Would lowering it to 30s sound reasonable?

PR #11786, opened 2022-01-20
Correctly await on_logged_out callbacks

babolivier:
Looks like this was an oversight from https://github.com/matrix-org/synapse/pull/10548

### DIFF
--- a/changelog.d/11786.bugfix
+++ b/changelog.d/11786.bugfix
@@ -1,0 +1,1 @@
+Fix a bug introduced in Synapse 1.46.0 that prevented `on_logged_out` module callbacks from being correctly awaited by Synapse.
--- a/synapse/handlers/auth.py
+++ b/synapse/handlers/auth.py
@@ -2281,7 +2281,7 @@ class PasswordAuthProvider:
         # call all of the on_logged_out callbacks
         for callback in self.on_logged_out_callbacks:
             try:
-                callback(user_id, device_id, access_token)
+                await callback(user_id, device_id, access_token)
             except Exception as e:
                 logger.warning("Failed to run module API callback %s: %s", callback, e)
                 continue
--- a/tests/handlers/test_password_providers.py
+++ b/tests/handlers/test_password_providers.py
@@ -727,10 +727,15 @@ class PasswordAuthProviderTests(unittest.HomeserverTestCase):
         tok = self.login("rin", "password")
 
         on_logged_out = Mock(return_value=make_awaitable(None))
-        self.hs.get_password_auth_provider().on_logged_out_callbacks.append(on_logged_out)
+        self.hs.get_password_auth_provider().on_logged_out_callbacks.append(
+            on_logged_out
+        )
 
         channel = self.make_request(
-            "POST", "/_matrix/client/v3/logout", {}, access_token=tok,
+            "POST",
+            "/_matrix/client/v3/logout",
+            {},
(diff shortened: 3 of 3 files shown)


synapse/handlers/auth.py
```diff
@@ -2281,7 +2281,7 @@ async def on_logged_out(
         # call all of the on_logged_out callbacks
         for callback in self.on_logged_out_callbacks:
             try:
-                callback(user_id, device_id, access_token)
+                await callback(user_id, device_id, access_token)
```
DMRobertson, line 2284: Argh are there any tests of the module API which could/should have caught this?

Related: https://github.com/python/mypy/issues/2499, https://bugs.python.org/issue30491

changelog.d/11786.bugfix
```diff
@@ -0,0 +1 @@
+Fix a bug introduced in Synapse 1.46.0 that prevented `on_logged_out` module callbacks from being correctly awaited by Synapse.
```
DMRobertson, line 1: Did `on_logged_out` exist before that version?

synapse/handlers/auth.py
babolivier: Hmm good point, I'll add a test for this

changelog.d/11786.bugfix
babolivier: Not in the new module system, but in the legacy password auth provider interface yes

tests/handlers/test_password_providers.py
```diff
@@ -719,6 +721,25 @@ def custom_auth_no_local_user_fallback_test_body(self):
...
+        )
+
+        channel = self.make_request(
+            "POST",
+            "/_matrix/client/v3/logout",
+            {},
+            access_token=tok,
+        )
+        self.assertEqual(channel.code, 200)
+        on_logged_out.assert_called_once()
```
DMRobertson, line 741: Hold on, does this actually detect the bug? I.e. does it fail without the fix?

`on_logged_out` is a substitution for `callback`. The problem isn't that `callback` didn't get called, it's that `callback(...)` didn't get awaited.

babolivier: Oh that's true huh

babolivier: So it turns out there's some magic in our test rig that turns `RuntimeWarning: coroutine 'PasswordAuthProviderTests.test_on_logged_out.<locals>.on_logged_out' was never awaited` into an error and fails the test. I've improved the test to check that we actually run the code in the function (and so that we await on it) just in case.

DMRobertson: Interesting. But we still didn't see a failure--- even with that magic. Does that mean that the tests simply never hit this path?

babolivier: Yes, the test suite was missing a test that checked that `on_logged_out` gets called.

DMRobertson: Oh right, because the test suite never added any callbacks to run in the first place. Gotcha, thanks.

DMRobertson (approved):
Thanks for bearing with me!

babolivier:
> Thanks for bearing with me!

No worries, thanks for reviewing this thoroughly, and sorry for the silly mistakes - looks like I haven't totally picked multitasking back up after the holidays :sweat_smile:

PR #11788, opened 2022-01-20
Make the `get_global_account_data_by_type_for_user` cache be a tree-cache whose key is prefixed with the user ID

reivilibre:
A review of this PR should have a bit of scrutiny behind it to make sure nothing has been missed; the type checker won't help us here for swapping a str and another str.

This is useful as part of #11621 to allow invalidating all the rows by user ID.

### DIFF
--- a/synapse/handlers/sync.py
+++ b/synapse/handlers/sync.py
@@ -1619,7 +1619,7 @@ class SyncHandler:
         # TODO: Can we `SELECT ignored_user_id FROM ignored_users WHERE ignorer_user_id=?;` instead?
         ignored_account_data = (
             await self.store.get_global_account_data_by_type_for_user(
-                AccountDataTypes.IGNORED_USER_LIST, user_id=user_id
+                user_id=user_id, data_type=AccountDataTypes.IGNORED_USER_LIST
             )
         )
 
--- a/synapse/storage/databases/main/account_data.py
+++ b/synapse/storage/databases/main/account_data.py
@@ -158,9 +158,9 @@ class AccountDataWorkerStore(CacheInvalidationWorkerStore):
             "get_account_data_for_user", get_account_data_for_user_txn
         )
 
-    @cached(num_args=2, max_entries=5000)
+    @cached(num_args=2, max_entries=5000, tree=True)
     async def get_global_account_data_by_type_for_user(
-        self, data_type: str, user_id: str
+        self, user_id: str, data_type: str
     ) -> Optional[JsonDict]:
         """
         Returns:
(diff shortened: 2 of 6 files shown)


synapse/handlers/sync.py
```diff
@@ -1619,7 +1619,7 @@ async def _get_ignored_users(self, user_id: str) -> FrozenSet[str]:
         # TODO: Can we `SELECT ignored_user_id FROM ignored_users WHERE ignorer_user_id=?;` instead?
         ignored_account_data = (
             await self.store.get_global_account_data_by_type_for_user(
-                AccountDataTypes.IGNORED_USER_LIST, user_id=user_id
+                user_id=user_id, data_type=AccountDataTypes.IGNORED_USER_LIST
```
reivilibre, line 1622: How does this interact with the caching? It uses kwargs so I don't know how our caching handles that

synapse/storage/databases/main/account_data.py
```diff
@@ -476,7 +476,7 @@ async def add_account_data_for_user(
             self._account_data_stream_cache.entity_has_changed(user_id, next_id)
             self.get_account_data_for_user.invalidate((user_id,))
             self.get_global_account_data_by_type_for_user.invalidate(
-                (account_data_type, user_id)
+                (user_id, account_data_type)
             )
```
clokep, line 480: The room case of this seems to prefill instead of invalidate, weirdly. I think this change is fine though, just noting a discrepancy:

https://github.com/matrix-org/synapse/blob/3e0536cd2afb5a640619bd872fc27b068ec3eb9b/synapse/storage/databases/main/account_data.py#L444-L446

synapse/handlers/sync.py
clokep: I think the cached functions just base it off name, so this should work OK. I would be fine with you making them into normal args if you'd prefer though.

clokep (approved):
This seems fine.

synapse/handlers/sync.py
reivilibre: I dug in to the (slightly confusing) code a bit and it looks like you're right :). Good to know.

synapse/storage/databases/main/account_data.py
reivilibre: It is a little bit odd, but it's correct either way — maybe it was an intentional optimisation, maybe it's just inconsistent.

clokep: I think prefilling would be a bit of an optimization, but probably not worth it. 🤷

PR #11789, opened 2022-01-20
Make `get_account_data_for_room_and_type` a tree cache

reivilibre:
This is useful as part of #11621 to allow invalidating all the rows by user ID.

### DIFF
--- a/synapse/storage/databases/main/account_data.py
+++ b/synapse/storage/databases/main/account_data.py
@@ -210,7 +210,7 @@ class AccountDataWorkerStore(CacheInvalidationWorkerStore):
             "get_account_data_for_room", get_account_data_for_room_txn
         )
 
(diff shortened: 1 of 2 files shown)


synapse/storage/databases/main/account_data.py
```diff
@@ -210,7 +210,7 @@ def get_account_data_for_room_txn(
             "get_account_data_for_room", get_account_data_for_room_txn
         )
 
-    @cached(num_args=3, max_entries=5000)
+    @cached(num_args=3, max_entries=5000, tree=True)
     async def get_account_data_for_room_and_type(
```
clokep, line 214: This seems OK, where we invalid this it looks like we do something weird with whether there's a `room_id` or not, but I don't think that would change from this PR.

clokep (approved):

PR #11790, opened 2022-01-20
Add a module callback to set username at registration

babolivier:
This is in the context of mainlining the Tchap fork of Synapse. Currently in Tchap usernames are derived from the user's email address (extracted from the UIA results, more specifically the `m.login.email.identity` step).
This change also exports the `check_username` method from the registration handler as part of the module API, so that a module can check if the username it's trying to generate is correct and doesn't conflict with an existing one, and fallback gracefully if not.
Edit: I've just realised `check_user_exists` is already exported by the module API, but I still think it's worthwhile having `check_username` available so that modules can check if the username they try to register follows the Matrix spec.


babolivier:
Example of a module using this callback: https://github.com/matrix-org/synapse-username-from-threepid/pull/1

docs/modules/password_auth_provider_callbacks.md
```diff
@@ -105,6 +105,64 @@ device ID), and the (now deactivated) access token.
...
+
+```python
+async def get_username_for_registration(
+    uia_results: Dict[str, Any],
+    params: Dict[str, Any],
+) -> Optional[str]
+```
+
+Called when registering a new user, allowing the module to define the username to give the user.
+The username is the local part of a user's full Matrix ID, i.e. it's `alice` in `@alice:example.com`.
```
DMRobertson, line 120: I think we want `e.g.` rather than `i.e.` here.

docs/modules/password_auth_provider_callbacks.md
```diff
@@ -105,6 +105,64 @@ device ID), and the (now deactivated) access token.
...
+        "address": "alice@example.com",
+        "validated_at": 1642701357084,
+    },
+    "m.login.msisdn": {  # User has provided and verified a phone number
+        "medium": "msisdn",
+        "address": "33123456789",
+        "validated_at": 1642701357084,
+    },
+    "org.matrix.msc3231.login.registration_token": "sometoken",  # User has registered through the flow described in MSC3231
+}
```
DMRobertson, line 148: Is a link to https://spec.matrix.org/v1.1/client-server-api/#authentication-types useful here?

I couldn't see `m.login.terms` in that link, and I couldn't see `m.login.sso` in this example. I don't know if that's actually a concern though.

docs/modules/password_auth_provider_callbacks.md
```diff
@@ -105,6 +105,64 @@ device ID), and the (now deactivated) access token.
...
+_First introduced in Synapse v1.52.0_
+
+```python
+async def get_username_for_registration(
+    uia_results: Dict[str, Any],
+    params: Dict[str, Any],
+) -> Optional[str]
+```
+
+Called when registering a new user, allowing the module to define the username to give the user.
```
DMRobertson, line 119: I don't think we actually spell out anywhere that the return value the username the module wants to be used. Maybe something like this?

Called when registering a new user. Modules can choose a username for a user by returning the desired username from this callback.

synapse/handlers/auth.py
```diff
@@ -2285,3 +2300,45 @@ async def on_logged_out(
...
+                res = await callback(uia_results, params)
+
+                if isinstance(res, str):
+                    return res
+                elif res is not None:
+                    # mypy complains that this line is unreachable because it assumes the
+                    # data returned by the module fits the expected type. We just want
+                    # to make sure this is the case.
+                    logger.warning(  # type: ignore[unreachable]
+                        "Ignoring non-string value returned by module in "
```
DMRobertson, line 2333: Should this say which module returned the bad value? I guess that's not easily available.

synapse/module_api/__init__.py
```diff
@@ -1202,6 +1208,22 @@ async def defer_to_thread(
         """
         return await defer_to_thread(self._hs.get_reactor(), f, *args, **kwargs)
 
+    async def check_username(self, username: str) -> None:
```
DMRobertson, line 1211: Perhaps `check_valid_username` here? (I wasn't sure what was being checked until I read the docstring)

DMRobertson (commented):
Seems good! I've left a few comments, but I think they're mostly nitpicks rather than anything critical.

synapse/module_api/__init__.py
DMRobertson: Oh, I misread. I thought this was only checking the grammar, but it also checks to see if the username is in use.

Maybe `check_username_available`?

babolivier: Well both are partially misleading imo. `check_username` checks the grammar _and_ the availability.

DMRobertson: Well put! Let's leave it be.

docs/modules/password_auth_provider_callbacks.md
babolivier: SSO isn't included because it actually happens in a separate code path (and I'm not sure having modules interact with SSO this way is a good idea) - I'll add a precision about this in the doc. `m.login.terms` is https://github.com/matrix-org/matrix-doc/pull/1692 which is (*sigh*) still not merged.

I'll add a link but will specify that this includes _most_ auth types, not all of them.

synapse/handlers/auth.py
babolivier: I've reworded this warning, and made it log the repr for the callback, which looks like this:

```
2022-01-26 11:50:23,425 - synapse.handlers.auth - 2332 - WARNING - POST-17 - Ignoring non-string value returned by get_username_for_registration callback <bound method UsernameFromThreepid.set_username_from_threepid of <username_from_threepid.UsernameFromThreepid object at 0x7fa77ea42020>>: [...]
```

which might be a bit too verbose but better this than not verbose enough.

DMRobertson (approved):
LGTM!

DMRobertson:
Merging in develop to run mypy with recent fixes.

PR #11792, opened 2022-01-20
Add `state_key` and `rejection_reason` to `events`

richvdh:
Part II of my long-held secret scheme to kill off the state_events and rejections tables: see #11496.

Here we add new columns to `events`, and start populating them for new arrivals.

~~Based on #11793~~


synapse/storage/schema/__init__.py
```diff
@@ -60,9 +60,9 @@
```
reivilibre, line 60: I think we can reuse schema version 68 because it hasn't gone out yet (right)? But should we note the addition of these 2 columns in the `SCHEMA_VERSION` change log above this line?

synapse/storage/schema/main/delta/67/01event_columns.sql
```diff
@@ -0,0 +1,26 @@
+/* Copyright 2022 The Matrix.org Foundation C.I.C
```
reivilibre, line 1: This file is in `67/` — should it be in `68/`?

reivilibre (approved):
Seems fine other than these little bookkeeping points. The plan you've proposed seems sensible.

synapse/storage/schema/main/delta/67/01event_columns.sql
richvdh: bah, when I first wrote this, 67 was the latest version :)

synapse/storage/schema/__init__.py
richvdh: > I think we can reuse schema version 68 because it hasn't gone out yet (right)?

yep.

> But should we note the addition of these 2 columns in the SCHEMA_VERSION change log above this line?

good point, we should note something. But, what's interesting here isn't so much the addition of the new columns, as the fact that synapse populates them for new events.

At some point in the future, I'll want to add a background update that populates the new columns for existing events. Once that starts, I'll want to make sure that nobody rolls back to a version of Synapse that doesn't populate them for *new* events. I'll do that by bumping SCHEMA_COMPAT_VERSION (to 68, in this case), which gets stored in the db (in `schema_compat_version.compat_version`), and earlier versions of Synapse will then refuse to start.

If I wasn't populating the columns, it's questionable as to whether I would need to record anything here: the fact that the columns exist doesn't have much bearing on which versions of Synapse you can run with (other than the fact they need disambiguating, which we've already covered below).

PR #11793, opened 2022-01-20
Add `FrozenEvent.get_state_key` and use it in a couple of places

richvdh:
This is more efficient, since we only have to look up `state_key` in the event dict once, rather than three (!) times. It also makes type annotations work rather better.


clokep (approved):
Seems reasonable to me.

Issue #6574, opened 2019-12-19
we should kill off the event_reference_hashes table


richvdh:
as far as I can tell, the values we read from it are never used. They are dutifully passed all the way down the stack into `create_new_client_event`, which throws them away.

richvdh:
so I started on this in #6629. It turns out that `event_reference_hashes` is still used when building V1 events (ie, events for room_versions 1 and 2), so we can't kill it off just yet.

richvdh:
Thinking about this a little more: I still think we should kill off `event_reference_hashes`: we can replace the hash lookups for v1 event population with calculations. The reasons for this are:

* most events are (or soon will be) not v1 events, so calculating and storing the reference hash is wasteful
* of those that are v1 events, most are only used once or twice in references, so again: calculating and storing them is wasteful. (The exception is things like the create event and PL event of busy rooms, but caching the calculation result will mitigate this)
* Calculating event references isn't that hard and if it saves waiting for a database lookup it's probably a win anyway
* Event persistence (as opposed to creation) is a bottleneck since it's hard to parallelise, so it's good to move any work we possible can out of the event persistence path.

PR #11794, opened 2022-01-20
Stop reading from `event_reference_hashes`

richvdh:
Preparation for dropping this table altogether. Part of #6574.


reivilibre (approved):
Looks good to me — can't spot any other reads from this table with `rg`.

synapse/storage/databases/main/signatures.py
```diff
@@ -41,18 +44,27 @@ async def get_event_reference_hashes(
 
         Returns:
              A mapping of event ID to a mapping of algorithm to hash.
+             Returns an empty dict for a given event id if that event is unknown.
```
richvdh, line 47: I should have said: it's not obvious to me that this is sensible behaviour, but it's what it was doing before, so I'm at least not making it worse.

(And somehow the sytests do hit this path.)

PR #11795, opened 2022-01-20
Drop unused table `public_room_list_stream`.

richvdh:
This is a follow-up to #10565.

### DIFF
--- a/synapse/storage/databases/main/purge_events.py
+++ b/synapse/storage/databases/main/purge_events.py
@@ -390,7 +390,6 @@ class PurgeEventsStore(StateGroupWorkerStore, CacheInvalidationWorkerStore):
             "event_search",
             "events",
             "group_rooms",
-            "public_room_list_stream",
             "receipts_graph",
             "receipts_linearized",
             "room_aliases",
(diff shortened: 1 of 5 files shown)


synapse/storage/databases/main/purge_events.py
```diff
@@ -390,7 +390,6 @@ def _purge_room_txn(self, txn, room_id: str) -> List[int]:
             "event_search",
             "events",
             "group_rooms",
-            "public_room_list_stream",
```
richvdh, line 393: the fact this was here means that if you:

 * roll forward to synapse 1.52 (?)
 * then roll back again
 * then do a "delete room" via the admin API

... then it won't work, and you'll get a 500 error. I feel like that's a sufficiently unlikely combo that it's not worth worrying about, but if it is of concern then we could do another round of removing references (together with a `TRUNCATE TABLE`) and do the actual deletion in the future.

reivilibre (approved):
Given that it's only affecting the admin API (and perhaps the experimental message retention — not sure), I'm happy to accept that.

richvdh:
> and perhaps the experimental message retention period feature — not sure

I don't think that's affected, as it only purges individual events, not entire rooms.

reivilibre:
> > and perhaps the experimental message retention period feature — not sure
> 
> I don't think that's affected, as it only purges individual events, not entire rooms.

Even better then! :)

Issue #11567, opened 2021-12-13
stop using `REPEATABLE READ` everywhere


richvdh:
currently, all postgres transactions are run in `REPEATABLE READ` isolation level, which is almost certainly more than we really need. The difficulty is the need to audit all of our transactions to check which actually need it.

richvdh:
per #4473: 

> Maybe we could do this with an option to `runInteraction` which indicates whether we need to change the isolation level (defaulting to "yes"). Over time we can go through our transactions and turn the option off for transactions we are happy don't require this.

clokep:
Maybe related to #4993?

PR #11799, opened 2022-01-21
Db txn set isolation level

Fizzadar:
Part of: https://github.com/matrix-org/synapse/issues/11567

Some things to figure out:

+ How should codepaths calling `runInteraction` specify the TXN level? They exist as constants within `psycopg2` which may not be installed. One option is a mapping of `str` -> level in the postgres engine module.
+ Is `None` the right value to reset the TXN level back to the (synapse specific) default

Signed-off-by: Nick Barrett nick@beeper.com

### Pull Request Checklist

* [ ] Pull request includes a [changelog file](https://matrix-org.github.io/synapse/latest/development/contributing_guide.html#changelog).

### DIFF
--- a/synapse/storage/database.py
+++ b/synapse/storage/database.py
@@ -702,6 +702,7 @@ class DatabasePool:
         func: Callable[..., R],
         *args: Any,
         db_autocommit: bool = False,
+        isolation_level: Optional[int] = None,
         **kwargs: Any,
     ) -> R:
         """Starts a transaction on the database and runs a given function
@@ -724,6 +725,7 @@ class DatabasePool:
                 called multiple times if the transaction is retried, so must
                 correctly handle that case.
 
+            isolation_level: Set the server isolation level for this transaction.
             args: positional args to pass to `func`
             kwargs: named args to pass to `func`
 
@@ -763,6 +765,7 @@ class DatabasePool:
         func: Callable[..., R],
         *args: Any,
         db_autocommit: bool = False,
+        isolation_level: Optional[int] = None,
         **kwargs: Any,
     ) -> R:
         """Wraps the .runWithConnection() method on the underlying db_pool.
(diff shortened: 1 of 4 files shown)


babolivier (changes requested):
Thanks a lot for looking into this!

> Is `None` the right value to reset the TXN level back to the (synapse specific) default

I'd say it's a good enough sentinel value.

Your changes seem to be missing this case, though, causing `attempt_to_set_isolation_level` to never get called when the isolation level is `None`.

Fizzadar:
Thanks for the review @babolivier!

> Your changes seem to be missing this case, though, causing attempt_to_set_isolation_level to never get called when the isolation level is None.

Is there anything to do in this case? `None` would mean no change from the default so no need to change the transaction level at all?

babolivier (approved):
> > Your changes seem to be missing this case, though, causing attempt_to_set_isolation_level to never get called when the isolation level is None.
> 
> Is there anything to do in this case? `None` would mean no change from the default so no need to change the transaction level at all?

Sorry, I misread your code :facepalm: I thought the `finally` clause would set the isolation level to `isolation_level`, not `None`, hence my confusion. So actually it's all good from me. Ideally this would come with a unit test but I'm not sure how we'd do it, so let's not block on that.

Could you merge `develop` in, though, please? There's a fix to the mypy failure in there.

synapse/storage/database.py
```diff
@@ -702,6 +702,7 @@ async def runInteraction(
         func: Callable[..., R],
         *args: Any,
         db_autocommit: bool = False,
+        isolation_level: Optional[int] = None,
```
richvdh, line 705: looks like this parameter isn't being passed into `runWithConnection`.

babolivier: I've opened https://github.com/matrix-org/synapse/pull/11847 to fix this.

PR #11804, opened 2022-01-24
Implementation for MSC3664: Pushrules for relations

deepbluev7:
An inefficient implementation for https://github.com/matrix-org/matrix-doc/pull/3664 based on the work from @bradtgmurray: https://gitlab.com/beeper/synapse/-/commit/44a1728b6b021f97900c89e0c00f7d1a23ce0d43

This probably should also be hidden behind an experimental flag, but I have no idea how to do that :3

I have also tested it on my homeserver and there it seems to work properly.

### Pull Request Checklist


bradtgmurray:
Thanks for doing the upstreaming work here, @deepbluev7. Happy to provide whatever signoff is needed.

deepbluev7:
@bradtgmurray It seems like adding a comment that has the signoff line on this PR should be enough. Thank you!

bradtgmurray:
Signed-off-by: Brad Murray brad@beeper.com

DMRobertson:
[A unit test seems to be failing](https://github.com/matrix-org/synapse/runs/4921981135?check_suite_focus=true). Can be reproduced with `trial tests.rest.client.test_relations.RelationsTestCase.test_deny_invalid_event`.

```
  File "/home/runner/work/synapse/synapse/tests/rest/client/test_relations.py", line 125, in test_deny_invalid_event
    self.assertEquals(200, channel.code, channel.json_body)
  File "/home/runner/work/synapse/synapse/.tox/py-noextras/lib/python3.10/site-packages/twisted/trial/_synctest.py", line 424, in assertEqual
    super().assertEqual(first, second, msg)
  File "/opt/hostedtoolcache/Python/3.10.2/x64/lib/python3.10/unittest/case.py", line 845, in assertEqual
    assertion_func(first, second, msg=msg)
  File "/opt/hostedtoolcache/Python/3.10.2/x64/lib/python3.10/unittest/case.py", line 838, in _baseAssertEqual
    raise self.failureException(msg)
twisted.trial.unittest.FailTest: 200 != 404 : {'errcode': 'M_NOT_FOUND', 'error': 'Could not find event foo'}

tests.rest.client.test_relations.RelationsTestCase.test_deny_invalid_event
```

Any idea what's going on here?

deepbluev7:
Ah, it seems like the issue is that synapse throws an exception, if an event is not in the store, which causes a 404 when evaluating the pushrule

deepbluev7:
That should fix it.

deepbluev7:
Well, https://github.com/matrix-org/synapse/blob/develop/tests/handlers/test_deactivate_account.py was new :D

deepbluev7:
Why would mypy error out now? I didn't touch that? D:

DMRobertson:
> Why would mypy error out now? I didn't touch that? D:

Should be fixed by #11832 and #11834. Can you merge in develop please?

deepbluev7:
Sure, will do

synapse/push/baserules.py
```diff
@@ -238,6 +238,25 @@ def make_base_prepend_rules(
         ],
         "actions": ["dont_notify"],
     },
+    # Enable notifications for replies without fallback
```
erikjohnston, line 241: Instead of adding default rules I wonder if its better to leave them out, and if clients want to test them they can add them itself?

synapse/push/push_rule_evaluator.py
```diff
@@ -145,7 +170,13 @@ def matches(
...
 
-    def _event_match(self, condition: dict, user_id: str) -> bool:
+    def _event_match(
+        self,
+        condition: dict,
+        user_id: str,
+        event: EventBase,
+        event_value_cache: Dict[str, str],
+    ) -> bool:
         pattern = condition.get("pattern", None)
```
erikjohnston, line 180: Since you're changing this can you add a docstring please? Doesn't need to be super detailed, just what the function does, roughly what the arguments are and what it returns

erikjohnston (commented):
This looks plausible enough to me, in terms of validating that the MSC works.

erikjohnston:
> This probably should also be hidden behind an experimental flag, but I have no idea how to do that :3

You add the flag to https://github.com/matrix-org/synapse/blob/9f9d82aa846332189e818f51d49daf2335780014/synapse/config/experimental.py#L47-L48

and then you can access it via `hs.config.experimental.*`

synapse/push/baserules.py
deepbluev7: How about defaulting the default rule to disabled, if the experimental option in the HS config isn't set? Clients can still add their own push rules that way or decide to enable it and the HS admin can enable it globally?

erikjohnston: TBH, I'm not sure what the effect of that would be for clients? I vaguely recall that adding new default push rules is a bit of a PITA, but I can't really remember why. Plus not sure how clients would react to these new push rules (it's probably fine, but I'm not 100% sure)

What about only including them if the experimental option is set?

deepbluev7: So if you don't enable it without the option set, it shouldn't affect notifications, because it isn't enabled. The major issue I could foresee, is that some clients would fail to parse the rule? I'm not sure how strict push rule parsers in clients are, but I would expect them to just ignore unknown conditions in disabled rules.

If you don't include the rule at all, clients can't know, if the condition type is supported. That might be fine and we could expose a flag in /capabilities instead. That option has a bit more implementation overhead, but that might be worth it for the safety? I can look into it this weekend.

richvdh: > I vaguely recall that adding new default push rules is a bit of a PITA, but I can't really remember why. Plus not sure how clients would react to these new push rules (it's probably fine, but I'm not 100% sure)

Related: https://github.com/matrix-org/matrix-doc/pull/3664#discussion_r789520216

synapse/push/baserules.py
```diff
@@ -238,6 +238,25 @@ def make_base_prepend_rules(
...
+    {
+        "rule_id": "global/override/.im.nheko.msc3664.reply",
+        "conditions": [
+            # Only send notification if the reply is to your message
+            {
+                "kind": "im.nheko.msc3664.related_event_match",
+                "key": "sender",
+                "rel_type": "m.in_reply_to",
+                "pattern_type": "user_id",
+                "_id": "_reply",
```
Fizzadar, line 251: I think I just got caught out by this - the `_id` field is [used to cache conditon results in the bulk evaluator](https://github.com/matrix-org/synapse/blob/e24ff8ebe3d4119d377355402245947f7de61c00/synapse/push/bulk_push_rule_evaluator.py#L277-L283) (seems to be the only place it's used) and so is unsafe to use when matching against the current user ID ([the cache is utilised on the room/event, rather than room/event/user, level](https://github.com/matrix-org/synapse/blob/e24ff8ebe3d4119d377355402245947f7de61c00/synapse/push/bulk_push_rule_evaluator.py#L210)).

DMRobertson:
What's the status of this @deepbluev7/@bradgtmurray ? This will have bitrotted with the rust push rules change; are you interested in pursuing this?

AFAICS the outstanding bits were a docstring, and an experimental config flag to guard the new behaviour?

deepbluev7:
It didn't bitrot with the rust changes, it caused conflicts with the threads relation stuff and I didn't have the energy to resolve them (since I basically need to rewrite it, since that change somewhat clobbered the variable names I was using). Also it is still not resolved how `is_falling_back` should be interpreted, which adds a whole bunch of additional complexity, if you want to keep the proper behaviour, see: https://github.com/matrix-org/matrix-spec-proposals/pull/3825

clokep:
> it caused conflicts with the threads relation stuff and I didn't have the energy to resolve them (since I basically need to rewrite it, since that change somewhat clobbered the variable names I was using)

This was mostly backed out FWIW. (If you're talking about MSC3772.)

deepbluev7:
I might have something soon now, seems like the conflicts were much easier to resolve now and it was just a few lines to implement in Rust. It will take a while for me to compile rust analyzer and stuff so that I can actually test, if the code works though. So I might have something by Monday.

deepbluev7:
Sup, updated, added capabilities and experimental config support, ported it to rust. No docstring, since the functions there have no docstrings anyway.

rust/src/push/mod.rs
```diff
@@ -299,6 +301,18 @@ pub struct EventMatchCondition {
...
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct RelatedEventMatchCondition {
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub key: Option<Cow<'static, str>>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub pattern: Option<Cow<'static, str>>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub pattern_type: Option<Cow<'static, str>>,
+    pub rel_type: Cow<'static, str>,
+}
```
erikjohnston, line 314: Do we need to support `include_fallbacks`?

synapse/push/bulk_push_rule_evaluator.py
```diff
@@ -225,6 +226,34 @@ async def action_for_event_by_user(
...
+                relation_type != "m.thread"
+                or not event.content.get("m.relates_to", {}).get(
+                    "is_falling_back", False
+                )
+            ):
+                related_event = await self.store.get_event(
+                    reply_event_id, allow_none=True
+                )
+                if related_event is not None:
+                    related_events["m.in_reply_to"] = _flatten_dict(related_event)
```
erikjohnston, line 255: I think its probably worth pulling this out into a helper function? It's a bit dense to be in the middle of this already fairly long function

erikjohnston (commented):
Looks good, just some Qs and tidy ups

erikjohnston:
@deepbluev7 as an aside I'd love to know how you found hacking on the Rust bit from a dev ex point of view?

rust/src/push/mod.rs
```diff
@@ -414,13 +434,25 @@ impl FilteredPushRules {
...
-                .get(&*r.rule_id)
-                .unwrap_or(&r.default_enabled);
-            (r, enabled)
-        })
+        self.push_rules
+            .iter()
+            .filter(|rule| {
+                // Ignore disabled experimental push rules
+                if !self.msc3664_enabled
+                    && rule.rule_id == "global/override/.im.nheko.msc3664.reply"
```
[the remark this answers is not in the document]
deepbluev7: It was done before for MSC3772. This just makes sure you can toggle them off in the config, which was asked for before.

rust/src/push/mod.rs
deepbluev7: Ah, I forgot to read my own MSC... I guess we should!

synapse/push/bulk_push_rule_evaluator.py
deepbluev7: Sure, sounds good to me!

deepbluev7:
> @deepbluev7 as an aside I'd love to know how you found hacking on the Rust bit from a dev ex point of view?

It's quite nice. Definitely makes some type errors easier to track down than Python. Biggest problems I had were, that poetry didn't tell me about compilation errors and figuring out this line:

```
let haystack = if let Some(haystack) = event.get(&**key) {
```

Also the separate install step is slightly annoying, but I was also just too lazy to setup the other tool that fixes that, so that is partially my own doing.

(Note, I have no prior rust experience, apart from reading what people write about it so far.)

rust/src/push/mod.rs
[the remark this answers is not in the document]
erikjohnston: Aaaah, I hadn't realised that the experimental MSC3772 had been removed.

erikjohnston:
These changes look good, but I think there must have been a problem with merging in develop? I'm failing to see what has happened but the tests are really *not* happy

erikjohnston (commented):
Other than tests going wibbly, LGTM

nico-famedly:
I think it was actually a typing error from switching Dict to dict. I should probably read up on what is the difference :D

DMRobertson:
> I think it was actually a typing error from switching Dict to dict. I should probably read up on what is the difference :D

`dict` with subscripts only works on Python 3.9+.

nico-famedly:
Well, that seems to have fixed it at least!

erikjohnston (approved):

PR #11806, opened 2022-01-24
Fix logic for dropping old events in fed queue

### LINKED ISSUE #11802: synapse gets stuck in a loop with `Invalid prev_events for <event_id>`
If inbound federation traffic backs up in a v1 room, synapse starts spamming the logs with messages like:

```
2022-01-23 00:42:23,593 - synapse.storage.databases.main.event_federation - 1436 - INFO - _process_incoming_pdus_in_room_inner-1197913 - Invalid prev_events for <event_id>
```

this means that rather than dropping excess traffic as it should, it ends up in a tight loop and fails to process incoming traffic from other sources.

anoadragon453:
This PR is a cherry-pick of a patch by @richvdh and fixes #11802.

There was a typo for `prev_events_tuple` in #10390 which lead to the infinite looping.

Fixes infinite loops of:

> logger.info("Invalid prev_events for %s", event_id)

### DIFF
--- a/changelog.d/11806.bugfix
+++ b/changelog.d/11806.bugfix
@@ -1,1 +1,1 @@
-Fix a bug introduced in v1.40.0 that caused Synapse to fail to process incoming federation traffic after handling a large amount of events in a v1 room.
+Fix a bug introduced in Synapse 1.40.0 that caused Synapse to fail to process incoming federation traffic after handling a large amount of events in a v1 room.
--- a/changelog.d/11806.bugfix
+++ b/changelog.d/11806.bugfix
@@ -1,0 +1,1 @@
+Fix a bug introduced in v1.40.0 that caused Synapse to fail to process incoming federation traffic after handling a large amount of events in a v1 room.
--- a/tests/storage/test_event_federation.py
+++ b/tests/storage/test_event_federation.py
@@ -12,10 +12,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from typing import Tuple, Union
+
 import attr
 from parameterized import parameterized
 
-from synapse.api.room_versions import RoomVersions
+from synapse.api.room_versions import RoomVersion, RoomVersions
 from synapse.events import _EventInternalMetadata
 from synapse.util import json_encoder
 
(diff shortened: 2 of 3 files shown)


anoadragon453:
I'm just writing some tests for this, bear with...

tests/storage/test_event_federation.py
```diff
@@ -506,11 +508,24 @@ def insert_event(txn):
         )
         self.assertSetEqual(difference, set())
 
-    def test_prune_inbound_federation_queue(self):
-        "Test that pruning of inbound federation queues work"
+    @parameterized.expand(
```
anoadragon453, line 511: We could additionally have this run for all known RoomVersions...

babolivier:
For the paper trail, the commit that introduces this patch to matrix-org-hotfixes is https://github.com/matrix-org/synapse/commit/2b9f741f3a9dee93e9744774b342c35ce60062c4

tests/storage/test_event_federation.py
babolivier: Yeah I think it might be better to programmatically build this list out of `KNOWN_ROOM_VERSIONS`, that way we're sure we don't accidentally reproduce this bug in a future room version.

tests/storage/test_event_federation.py
```diff
@@ -506,11 +508,24 @@ def insert_event(txn):
...
+        [
+            (RoomVersions.V1,),
+            (RoomVersions.V9,),
+        ]
+    )
+    def test_prune_inbound_federation_queue(self, room_version: RoomVersion):
```
babolivier, line 517: ooi have you tested that this test fails without the patch?

tests/storage/test_event_federation.py
```diff
@@ -506,11 +508,24 @@ def insert_event(txn):
...
+        """Test that pruning of inbound federation queues work"""
 
         room_id = "some_room_id"
 
+        def prev_event_format(prev_event_id: str) -> Union[Tuple[str, dict], str]:
+            """Account for differences in prev_events format across room versions"""
+            if room_version == RoomVersions.V1:
```
babolivier, line 524: If we change the test parameters to include all known room versions, this condition needs to also include v2 rooms according to the spec

babolivier (approved):
lgtm overall

tests/storage/test_event_federation.py
anoadragon453: yep, it fails.

changelog.d/11806.bugfix
```diff
@@ -0,0 +1 @@
+Fix a bug introduced in v1.40.0 that caused Synapse to fail to process incoming federation traffic after handling a large amount of events in a v1 room.
```
babolivier, line 1: Oh actually, for consistency:

```suggestion
Fix a bug introduced in Synapse 1.40.0 that caused Synapse to fail to process incoming federation traffic after handling a large amount of events in a v1 room.
```

PR #11808, opened 2022-01-24
Make get_room_version use cached get_room_version_id.

### LINKED ISSUE #11706: `get_room_version` doesn't use the cache
we have two functions for getting the version of a room from the database, `get_room_version_id` and `get_room_version`. The former gets the raw string, and then caches it to speed up future lookups. The second looks up the raw string, and then consults the list of known room versions to decode it into a `RoomVersion` object. The problem is that it doesn't cache the result, so is generally slower than it should be.

If `get_room_version` needs to exist at all, it should probably just wrap `get_room_version_id`.

lukasdenk:
Fixes #11706.

Make method `get_room_version` use cached `get_room_version_id`.

### Pull Request Checklist

### DIFF
--- a/synapse/storage/databases/main/state.py
+++ b/synapse/storage/databases/main/state.py
@@ -42,6 +42,16 @@ logger = logging.getLogger(__name__)
 MAX_STATE_DELTA_HOPS = 100
 
 
+def _retrieve_and_check_room_version(room_id: str, room_version_id: str) -> RoomVersion:
+    v = KNOWN_ROOM_VERSIONS.get(room_version_id)
+    if not v:
+        raise UnsupportedRoomVersionError(
+            "Room %s uses a room version %s which is no longer supported"
+            % (room_id, room_version_id)
+        )
+    return v
+
+
 # this inherits from EventsWorkerStore because it calls self.get_events
 class StateGroupWorkerStore(EventsWorkerStore, SQLBaseStore):
     """The parts of StateGroupStore that can be called from workers."""
@@ -63,7 +73,7 @@ class StateGroupWorkerStore(EventsWorkerStore, SQLBaseStore):
                 removed from Synapse.
         """
         room_version_id = await self.get_room_version_id(room_id)
-        return self._retrieve_and_check_room_version(room_id, room_version_id)
+        return _retrieve_and_check_room_version(room_id, room_version_id)
 
     def get_room_version_txn(
         self, txn: LoggingTransaction, room_id: str
@@ -79,18 +89,7 @@ class StateGroupWorkerStore(EventsWorkerStore, SQLBaseStore):
                 removed from Synapse.
         """
         room_version_id = self.get_room_version_id_txn(txn, room_id)
-        return self._retrieve_and_check_room_version(room_id, room_version_id)
-
-    def _retrieve_and_check_room_version(
-        self, room_id: str, room_version_id: str
-    ) -> RoomVersion:
-        v = KNOWN_ROOM_VERSIONS.get(room_version_id)
-        if not v:
-            raise UnsupportedRoomVersionError(
-                "Room %s uses a room version %s which is no longer supported"
-                % (room_id, room_version_id)
-            )
-        return v
+        return _retrieve_and_check_room_version(room_id, room_version_id)
 
     @cached(max_entries=10000)
     async def get_room_version_id(self, room_id: str) -> str:
--- a/synapse/storage/databases/main/state.py
+++ b/synapse/storage/databases/main/state.py
@@ -62,11 +62,8 @@ class StateGroupWorkerStore(EventsWorkerStore, SQLBaseStore):
                 Typically this happens if support for the room's version has been
                 removed from Synapse.
         """
-        return await self.db_pool.runInteraction(
-            "get_room_version_txn",
-            self.get_room_version_txn,
-            room_id,
-        )
+        room_version_id = self.get_room_version_id(room_id)
+        return self.__retrieve_and_check_room_version(room_id, room_version_id)
 
     def get_room_version_txn(
         self, txn: LoggingTransaction, room_id: str
@@ -82,14 +79,17 @@ class StateGroupWorkerStore(EventsWorkerStore, SQLBaseStore):
                 removed from Synapse.
         """
         room_version_id = self.get_room_version_id_txn(txn, room_id)
-        v = KNOWN_ROOM_VERSIONS.get(room_version_id)
+        return self.__retrieve_and_check_room_version(room_id, room_version_id)
 
+    def __retrieve_and_check_room_version(
+        self, room_id: str, room_version_id: str
+    ) -> RoomVersion:
+        v = KNOWN_ROOM_VERSIONS.get(room_version_id)
         if not v:
             raise UnsupportedRoomVersionError(
                 "Room %s uses a room version %s which is no longer supported"
                 % (room_id, room_version_id)
             )
-
         return v
 
     @cached(max_entries=10000)
--- a/synapse/storage/databases/main/state.py
+++ b/synapse/storage/databases/main/state.py
@@ -62,7 +62,7 @@ class StateGroupWorkerStore(EventsWorkerStore, SQLBaseStore):
                 Typically this happens if support for the room's version has been
                 removed from Synapse.
         """
-        room_version_id = self.get_room_version_id(room_id)
+        room_version_id = await self.get_room_version_id(room_id)
         return self.__retrieve_and_check_room_version(room_id, room_version_id)
 
     def get_room_version_txn(
--- a/synapse/storage/databases/main/state.py
+++ b/synapse/storage/databases/main/state.py
@@ -63,7 +63,7 @@ class StateGroupWorkerStore(EventsWorkerStore, SQLBaseStore):
                 removed from Synapse.
         """
         room_version_id = await self.get_room_version_id(room_id)
-        return self.__retrieve_and_check_room_version(room_id, room_version_id)
+        return self._retrieve_and_check_room_version(room_id, room_version_id)
 
     def get_room_version_txn(
         self, txn: LoggingTransaction, room_id: str
@@ -79,9 +79,9 @@ class StateGroupWorkerStore(EventsWorkerStore, SQLBaseStore):
                 removed from Synapse.
         """
         room_version_id = self.get_room_version_id_txn(txn, room_id)
-        return self.__retrieve_and_check_room_version(room_id, room_version_id)
+        return self._retrieve_and_check_room_version(room_id, room_version_id)
 
-    def __retrieve_and_check_room_version(
+    def _retrieve_and_check_room_version(
         self, room_id: str, room_version_id: str
     ) -> RoomVersion:
         v = KNOWN_ROOM_VERSIONS.get(room_version_id)
--- a/tests/handlers/test_room_summary.py
+++ b/tests/handlers/test_room_summary.py
@@ -715,7 +715,7 @@ class SpaceSummaryTestCase(unittest.HomeserverTestCase):
 
     def test_unknown_room_version(self):
         """
-        If an room with an unknown room version is encountered it should not cause
+        If a room with an unknown room version is encountered it should not cause
         the entire summary to skip.
         """
         # Poke the database and update the room version to an unknown one.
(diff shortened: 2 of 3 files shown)


synapse/storage/databases/main/state.py
```diff
@@ -62,11 +62,8 @@ async def get_room_version(self, room_id: str) -> RoomVersion:
                 Typically this happens if support for the room's version has been
                 removed from Synapse.
         """
-        return await self.db_pool.runInteraction(
-            "get_room_version_txn",
-            self.get_room_version_txn,
-            room_id,
-        )
+        room_version_id = self.get_room_version_id(room_id)
```
reivilibre, line 65:
```suggestion
        room_version_id = await self.get_room_version_id(room_id)
```
I think this is the reason the tests are failing

reivilibre (changes requested):
Seems sensible — though would like to check the tests agree with me :-).
Thank you!

lukasdenk:
> (By the way: if you write 'Fixes #11706' instead of 'Solves #11706', GitHub will detect that and link the issue to this PR so that the issue will be closed automatically when the PR is merged. I've changed it for you above, but may be useful for your reference)

Thanks :)

synapse/storage/databases/main/state.py
```diff
@@ -82,14 +79,17 @@ def get_room_version_txn(
                 removed from Synapse.
         """
         room_version_id = self.get_room_version_id_txn(txn, room_id)
-        v = KNOWN_ROOM_VERSIONS.get(room_version_id)
+        return self.__retrieve_and_check_room_version(room_id, room_version_id)
 
+    def __retrieve_and_check_room_version(
```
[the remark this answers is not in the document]
lukasdenk: ok

tests/handlers/test_room_summary.py
```diff
@@ -718,6 +717,9 @@ def test_unknown_room_version(self):
         If an room with an unknown room version is encountered it should not cause
         the entire summary to skip.
         """
+        self.hs.get_datastores().main._attempt_to_invalidate_cache(
+            "get_room_version_id", (self.room,)
+        )
```
reivilibre, line 722: Just for clarity, I would put this after you do the UPDATE below (and maybe add a comment, but I think it'd be more obvious if it came after the update).
The current position of it had me wondering if something was leaking from a previous test or something strange.

I would also write this instead:

`self.hs.get_datastores().main.get_room_version_id.invalidate((self.room,))` — I haven't seen the form you've written before, but the documentation says it will do nothing if the cache doesn't exist. It's probably better to get an error if the cache doesn't exist so we can fix the test :).

reivilibre (changes requested):
Also note isort is complaining (possibly about the `import mock` line in the test that you seem to have introduced).

synapse/storage/databases/main/state.py
```diff
@@ -82,14 +79,17 @@ def get_room_version_txn(
                 removed from Synapse.
         """
         room_version_id = self.get_room_version_id_txn(txn, room_id)
-        v = KNOWN_ROOM_VERSIONS.get(room_version_id)
+        return self._retrieve_and_check_room_version(room_id, room_version_id)
 
+    def _retrieve_and_check_room_version(
```
richvdh, line 84: make this a top-level function rather than a member method? (it's fractionally faster to call a top-level function than a method)

lukasdenk: Hmm..., don't you think, it decreases code readability?

tests/handlers/test_room_summary.py
lukasdenk: Done

synapse/storage/databases/main/state.py
reivilibre: I think it would be fine to make this a top-level function here; arguably it might *improve* readability since this method takes a `self` reference that it doesn't use (but from the outside it's not always clear why it does!).
It's also only 60 lines into the code unit, so it wouldn't be that far away anyway

tests/handlers/test_room_summary.py
```diff
@@ -727,6 +727,7 @@ def test_unknown_room_version(self):
                 desc="updated-room-version",
             )
         )
+        self.hs.get_datastores().main.get_room_version_id.invalidate((self.room,))
```
reivilibre, line 730: Why is this needed?

It's bad practice to add invalidations for the tests since it might cover up an invalidation problem in the real code (I ran into this too during one of my recent PRs).
It's not clear to my why it's needed in the first place, but ideally we should fix it so it's not needed, or failing that, add a comment about it?

reivilibre (changes requested):
Sorry for the delay; thanks for bearing with us!

tests/handlers/test_room_summary.py
lukasdenk: The test tests whether a room summary still succeeds, even if the room's version is invalid. To do so, it sets the room version id to `unknown-room-version` in the lines above. However the room version has already been requested before. Therefore, `get_room_version` will not return `unknown-room-version` as the room version for `room_id`. Hence, `get_room_version`'s cache must be invalidated.

lukasdenk: What is weird though, is that removing these lines

https://github.com/matrix-org/synapse/blob/3655585e85bfebd9165151d9d69fd907d65d6db0/tests/handlers/test_room_summary.py#L722-L729

(and not invalidating `get_version_id`) results in a different exception than including the above lines (and not invalidating `get_version_id`). However, in both cases, `get_version_id` returns the same, namely `'6'`. However, I would expect both scenarios would result in the same exception. @reivilibre Do you have any idea why this could be the case? Should I dig more deeply into the reasons? 


The first approach fails here

https://github.com/matrix-org/synapse/blob/3655585e85bfebd9165151d9d69fd907d65d6db0/tests/handlers/test_room_summary.py#L734

with

```
Failure
  File "/home/lukas/PycharmProjects/synapse/env/lib/python3.8/site-packages/twisted/internet/defer.py", line 478, in addCallbacks
    self._runCallbacks()
  File "/home/lukas/PycharmProjects/synapse/env/lib/python3.8/site-packages/twisted/internet/defer.py", line 858, in _runCallbacks
    current.result = callback(  # type: ignore[misc]
  File "/home/lukas/PycharmProjects/synapse/env/lib/python3.8/site-packages/twisted/trial/_asynctest.py", line 152, in deferTestMethod
    d = self._run(self._testMethodName, result)
  File "/home/lukas/PycharmProjects/synapse/env/lib/python3.8/site-packages/twisted/trial/_asynctest.py", line 122, in _run
    d = defer.maybeDeferred(
--- <exception caught here> ---
  File "/home/lukas/PycharmProjects/synapse/env/lib/python3.8/site-packages/twisted/internet/defer.py", line 191, in maybeDeferred
    result = f(*args, **kwargs)
  File "/home/lukas/PycharmProjects/synapse/env/lib/python3.8/site-packages/twisted/internet/utils.py", line 209, in runWithWarningsSuppressed
    raise exc_info[1].with_traceback(exc_info[2])
  File "/home/lukas/PycharmProjects/synapse/env/lib/python3.8/site-packages/twisted/internet/utils.py", line 205, in runWithWarningsSuppressed
    result = f(*a, **kw)
  File "/home/lukas/PycharmProjects/synapse/tests/handlers/test_room_summary.py", line 737, in test_unknown_room_version
    self._assert_rooms(result, expected)
  File "/home/lukas/PycharmProjects/synapse/tests/handlers/test_room_summary.py", line 178, in _assert_rooms
    self.assertCountEqual(
  File "/usr/lib/python3.8/unittest/case.py", line 1272, in assertCountEqual
    self.fail(msg)
  File "/home/lukas/PycharmProjects/synapse/env/lib/python3.8/site-packages/twisted/trial/_synctest.py", line 361, in fail
    raise self.failureException(msg)
twisted.trial.unittest.FailTest: Element counts were not equal:
First has 1, Second has 0:  '!XgELWsagGJJtucVyif:test'
```

while the latter fails here

https://github.com/matrix-org/synapse/blob/3655585e85bfebd9165151d9d69fd907d65d6db0/tests/handlers/test_room_summary.py#L731

with 

```
Failure
  File "/home/lukas/PycharmProjects/synapse/env/lib/python3.8/site-packages/twisted/internet/defer.py", line 478, in addCallbacks
    self._runCallbacks()
  File "/home/lukas/PycharmProjects/synapse/env/lib/python3.8/site-packages/twisted/internet/defer.py", line 858, in _runCallbacks
    current.result = callback(  # type: ignore[misc]
  File "/home/lukas/PycharmProjects/synapse/env/lib/python3.8/site-packages/twisted/trial/_asynctest.py", line 152, in deferTestMethod
    d = self._run(self._testMethodName, result)
  File "/home/lukas/PycharmProjects/synapse/env/lib/python3.8/site-packages/twisted/trial/_asynctest.py", line 122, in _run
    d = defer.maybeDeferred(
--- <exception caught here> ---
  File "/home/lukas/PycharmProjects/synapse/env/lib/python3.8/site-packages/twisted/internet/defer.py", line 191, in maybeDeferred
    result = f(*args, **kwargs)
  File "/home/lukas/PycharmProjects/synapse/env/lib/python3.8/site-packages/twisted/internet/utils.py", line 209, in runWithWarningsSuppressed
    raise exc_info[1].with_traceback(exc_info[2])
  File "/home/lukas/PycharmProjects/synapse/env/lib/python3.8/site-packages/twisted/internet/utils.py", line 205, in runWithWarningsSuppressed
    result = f(*a, **kw)
  File "/home/lukas/PycharmProjects/synapse/tests/handlers/test_room_summary.py", line 734, in test_unknown_room_version
    result = self.get_success(self.handler.get_space_summary(self.user, self.space))
  File "/home/lukas/PycharmProjects/synapse/tests/unittest.py", line 522, in get_success
    return self.successResultOf(d)
  File "/home/lukas/PycharmProjects/synapse/env/lib/python3.8/site-packages/twisted/trial/_synctest.py", line 700, in successResultOf
    self.fail(
  File "/home/lukas/PycharmProjects/synapse/env/lib/python3.8/site-packages/twisted/trial/_synctest.py", line 361, in fail
    raise self.failureException(msg)
twisted.trial.unittest.FailTest: Success result expected on <Deferred at 0x7f6e46cdd460 current result: None>, found failure result instead:
  File "/home/lukas/PycharmProjects/synapse/env/lib/python3.8/site-packages/twisted/internet/defer.py", line 662, in callback
    self._startRunCallbacks(result)
  File "/home/lukas/PycharmProjects/synapse/env/lib/python3.8/site-packages/twisted/internet/defer.py", line 764, in _startRunCallbacks
    self._runCallbacks()
  File "/home/lukas/PycharmProjects/synapse/env/lib/python3.8/site-packages/twisted/internet/defer.py", line 858, in _runCallbacks
    current.result = callback(  # type: ignore[misc]
  File "/home/lukas/PycharmProjects/synapse/env/lib/python3.8/site-packages/twisted/internet/defer.py", line 1751, in gotResult
    current_context.run(_inlineCallbacks, r, gen, status)
--- <exception caught here> ---
  File "/home/lukas/PycharmProjects/synapse/env/lib/python3.8/site-packages/twisted/internet/defer.py", line 1661, in _inlineCallbacks
  File "/home/lukas/PycharmProjects/synapse/synapse/handlers/room_summary.py", line 176, in get_space_summary
    room_entry = await self._summarize_local_room(
  File "/home/lukas/PycharmProjects/synapse/synapse/handlers/room_summary.py", line 651, in _summarize_local_room
    if not await self._is_local_room_accessible(room_id, requester, origin):
  File "/home/lukas/PycharmProjects/synapse/synapse/handlers/room_summary.py", line 860, in _is_local_room_accessible
    join_rules_event = await self._store.get_event(join_rules_event_id)
  File "/home/lukas/PycharmProjects/synapse/synapse/storage/databases/main/events_worker.py", line 385, in get_event
    raise NotFoundError("Could not find event %s" % (event_id,))
synapse.api.errors.NotFoundError: 404: Could not find event $7nE7ON6WDexQYs-xMFG5SMOfIe2wx00BQPwgItkyyIg
```

reivilibre:
```
synapse.api.errors.NotFoundError: 404: Could not find event $7nE7ON6WDexQYs-xMFG5SMOfIe2wx00BQPwgItkyyIg
```

Looking at this one, it seems like it occurs because the event fetching code fetches the room version separately and then encounters a room version it doesn't understand: it then skips fetching that event. (This is a design choice: events in unknown room versions are treated as nonexistent events because we don't know how to decode them)

I think that's fine; in practice the version of a room doesn't change and so there's no way the cache would get out of sync with the database.

```
twisted.trial.unittest.FailTest: Element counts were not equal: First has 1, Second has 0: '!XgELWsagGJJtucVyif:test'
```
This is happening because if you don't set the room version to something unknown, the code will (correctly) generate a room summary entry for it, which the test doesn't expect for unknown room versions.

I think your code is therefore fine as-is!

reivilibre (approved):
Thanks and so sorry for this one falling behind!

PR #11810, opened 2022-01-24
Skip the initial amd64-only Docker build

richvdh:
PyNaCl's recent 1.5.0 release on PyPi includes arm64 wheels, which means our arm64 docker images now build in a sensible amount of time, so we can skip the amd64-only build.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -34,6 +34,8 @@ jobs:
           username: ${{ secrets.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_TOKEN }}
 
+      # TODO: consider using https://github.com/docker/metadata-action instead of this
+      # custom magic
       - name: Calculate docker image tag
         id: set-tag
         run: |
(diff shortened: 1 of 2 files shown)


DMRobertson (approved):
Seems reasonable to me, especially since we wait for both of these as part of the release process.

Is there anything else building from source on arm64? (i.e. is there anything else particularly slow that would benefit from an upstream wheel?)

richvdh:
> (i.e. is there anything else particularly slow that would benefit from an upstream wheel?)

probably! you should be able to see what the slow bits are from the logs.

DMRobertson:
> > (i.e. is there anything else particularly slow that would benefit from an upstream wheel?)
> 
> probably! you should be able to see what the slow bits are from the logs.

Serves me right for asking!

A quick skim shows:
- 5min doing an `apt-get update && apt-get install`
- 12 min doing a `pip install`.
  - setup.py for `thrift` takes 1min; `psycopg2` takes 2min. 5 minutes downloading packages.

PR #11811, opened 2022-01-24
CI: run Complement on the VM, not inside Docker

kegsay:
This requires https://github.com/matrix-org/complement/pull/289 - the branch name matches this PR so CI should test this new configuration.

The run: https://github.com/matrix-org/synapse/runs/4924828255?check_suite_focus=true

We now run Complement on the VM instead of inside a Docker container. This is to allow Complement to bind to any high-numbered port when it starts up its own federation servers. We want to do this to allow for more concurrency when running complement tests.

Previously, Complement only ever bound to `:8448` when running its own federation server. This prevented multiple federation tests running at the same time as they would fight each other on the port. This did however allow Complement to run in Docker, as the host could just port forward `:8448` to allow homeserver containers to communicate to Complement. Now that we are using random ports however, we cannot use Docker to run Complement.

This ends up being a good thing because:

 - Running Complement tests locally is closer to how they run in CI.
 - Allows the `CI` env var to be removed in Complement.
 - Slightly speeds up runs as we don't need to pull down the Complement image prior to running tests. This assumes GHA caches actions sensibly.

### Pull Request Checklist

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -329,6 +329,7 @@ jobs:
       # See https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#adding-a-system-path
       - name: "Set Go Version"
         run: |
+          # Add Go 1.17 to the PATH: see https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-Readme.md#environment-variables-2
           echo "$GOROOT_1_17_X64/bin" >> $GITHUB_PATH
           echo "~/go/bin" >> $GITHUB_PATH
 
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -382,7 +382,7 @@ jobs:
 
       # Run Complement
       - run: |
-          set -o pipefail &&
+          set -o pipefail
           go test -v -json -tags synapse_blacklist,msc2403 ./tests/... 2>&1 | gotestfmt
         shell: bash
         name: Run Complement Tests
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -323,17 +323,22 @@ jobs:
     if: ${{ !failure() && !cancelled() }}
     needs: linting-done
     runs-on: ubuntu-latest
-    container:
-      # https://github.com/matrix-org/complement/blob/master/dockerfiles/ComplementCIBuildkite.Dockerfile
-      image: matrixdotorg/complement:latest
-      env:
-        CI: true
-      ports:
-        - 8448:8448
-      volumes:
-        - /var/run/docker.sock:/var/run/docker.sock
 
     steps:
+      # Env vars are set file a file given by $GITHUB_PATH. We need both Go 1.17 and GOPATH on env to run Complement.
+      # See https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#adding-a-system-path
+      - name: "Set Go Version"
+        run: |
+          echo "$GOROOT_1_17_X64/bin" >> $GITHUB_PATH
+          echo "~/go/bin" >> $GITHUB_PATH
+
+      - name: "Install Complement Dependencies"
+        # We don't need to install Go because it is included on the Ubuntu 20.04 image:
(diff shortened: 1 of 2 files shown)


.github/workflows/tests.yml
```diff
@@ -323,17 +323,22 @@ jobs:
...
-      image: matrixdotorg/complement:latest
-      env:
-        CI: true
-      ports:
-        - 8448:8448
-      volumes:
-        - /var/run/docker.sock:/var/run/docker.sock
 
     steps:
+      # Env vars are set file a file given by $GITHUB_PATH. We need both Go 1.17 and GOPATH on env to run Complement.
```
babolivier, line 328: > Env vars are set file a file given by $GITHUB_PATH

Sorry I'm having troubles parsing this sentence, could you clarify it a bit please?

richvdh: not just any env var - specifically `PATH`, aiui.

.github/workflows/tests.yml
```diff
@@ -323,17 +323,22 @@ jobs:
...
+      # See https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#adding-a-system-path
+      - name: "Set Go Version"
+        run: |
+          echo "$GOROOT_1_17_X64/bin" >> $GITHUB_PATH
```
richvdh, line 332:
```suggestion
          # Add Go 1.17 to the PATH: see https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-Readme.md#environment-variables-2
          echo "$GOROOT_1_17_X64/bin" >> $GITHUB_PATH
```

.github/workflows/tests.yml
```diff
@@ -323,17 +323,22 @@ jobs:
...
+          echo "~/go/bin" >> $GITHUB_PATH
+
+      - name: "Install Complement Dependencies"
+        # We don't need to install Go because it is included on the Ubuntu 20.04 image:
+        # See https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-Readme.md specifically GOROOT_1_17_X64
```
richvdh, line 337: this feels a bit out of place - better to add a comment above as suggested.

.github/workflows/tests.yml
```diff
@@ -323,17 +323,22 @@ jobs:
...
-      volumes:
-        - /var/run/docker.sock:/var/run/docker.sock
 
     steps:
+      # Env vars are set file a file given by $GITHUB_PATH. We need both Go 1.17 and GOPATH on env to run Complement.
+      # See https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#adding-a-system-path
+      - name: "Set Go Version"
+        run: |
+          echo "$GOROOT_1_17_X64/bin" >> $GITHUB_PATH
+          echo "~/go/bin" >> $GITHUB_PATH
```
richvdh, line 333: maybe a silly question: why do we need this?

richvdh (commented):
generally seems sensible to me. A few nits about comments.

.github/workflows/tests.yml
kegsay: Clarified

.github/workflows/tests.yml
kegsay: Clarified

kegsay:
Unsure why `Tests / lint (mypy) (pull_request) Failing after 3m — lint (mypy) ` is randomly failing now with:
```
mypy run-test: commands[0] | mypy
synapse/events/validator.py:249: error: Variable "jsonschema.validators.Draft7Validator" is not valid as a type  [valid-type]
synapse/events/validator.py:249: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
Found 1 error in 1 file (checked 640 source files)
ERROR: InvocationError for command /home/runner/work/synapse/synapse/.tox/mypy/bin/mypy (exited with code 1)
```

DMRobertson:
> Unsure why `Tests / lint (mypy) (pull_request) Failing after 3m — lint (mypy) ` is randomly failing now with:

#11817 should have fixed this, and it should be included with your most recent merge.

richvdh (approved):
lgtm otherwise

PR #11816, opened 2022-01-24
Check that `gc` method is available before using in `synapse/app/_base` 

H-Shay:
A similar check was removed in #11683. The check determined whether the program was being run on Python 3.7 before using a garbage collection method and thus was deemed redundant in this comment: https://github.com/matrix-org/synapse/pull/11683#pullrequestreview-859931689, so I removed it. However, it was pointed out here: https://github.com/matrix-org/synapse/pull/11683#discussion_r790695704 that not all implementations of Python have that API so I have restored a version of the removed check in this new PR.

### DIFF
--- a/synapse/app/_base.py
+++ b/synapse/app/_base.py
@@ -468,7 +468,7 @@ async def start(hs: "HomeServer") -> None:
     # rest of time. Doing so means less work each GC (hopefully).
     #
     # gc.freeze may not be available in all Python implementations, thus we check here
-    if hasattr(gc, 'freeze'):
+    if hasattr(gc, "freeze"):
         gc.collect()
         gc.freeze()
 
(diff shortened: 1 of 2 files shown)


clokep (approved):
LGTM!

Looking at the removed code it was a bit silly that we were doing the same exact check twice in a row...

synapse/app/_base.py
```diff
@@ -467,12 +467,14 @@ def run_sighup(*args: Any, **kwargs: Any) -> None:
     # everything currently allocated are things that will be used for the
     # rest of time. Doing so means less work each GC (hopefully).
     #
-    gc.collect()
-    gc.freeze()
-
-    # Speed up shutdowns by freezing all allocated objects. This moves everything
-    # into the permanent generation and excludes them from the final GC.
-    atexit.register(gc.freeze)
+    # gc.freeze may not be available in all Python implementations, thus we check here
```
callahad, line 470: It may be useful to actually include the string "pypy" in here somewhere, since that's the most common alternative to CPython that we care about... sort of a breadcrumb for us to find our way back with later.

```suggestion
    # PyPy does not (yet?) implement gc.freeze()
```

Issue #11780, opened 2022-01-20
support for python 3.10


richvdh:
Python 3.10.0 final was released 3 months ago, on 2021-10-04. We should add support for it.

richvdh:
I think this involves:
 * [x] adding 3.10 to the list of distributions we run `trial` tests on (already done in #11595)
 * [ ] building new sytest docker images which include python 3.10, and adding it to the list for CI
 * [ ] updating https://matrix-org.github.io/synapse/develop/setup/installation.html#installing-as-a-python-module-from-pypi to note 3.10 is now supported.

Relatedly, following the y-1 rule suggested at https://github.com/matrix-org/synapse/issues/8674#issuecomment-719990241, it's time to update the docker images to python 3.9.

DMRobertson:
> * adding 3.10 to the list of distributions we run `trial` tests on (already done in

See also https://github.com/matrix-org/synapse/pull/10992/

anoadragon453:
> building new sytest docker images which include python 3.10 (I think we could use Debian 'bookworm' as a basis for this) and adding it to the list for CI

It looks like this last remaining item was done in https://github.com/matrix-org/sytest/pull/1184, and indeed we [now have bookworm images on DockerHub](https://hub.docker.com/layers/matrixdotorg/sytest/bookworm/images/sha256-09d9c6752a25aadd668d492283e70f7858e337df8576009f82e995dda0ef3503).

With that final box ticked, can this issue be closed?

richvdh:
yes, if you open a new one to track:

> Relatedly, following the y-1 rule suggested at https://github.com/matrix-org/synapse/issues/8674#issuecomment-719990241, it's time to update the docker images to python 3.9.

:)

PR #11817, opened 2022-01-24
Ignore the jsonschema type.

clokep:
types-jsonschema 4.4.0 broke our mypy, see python/typeshed#7028.

Previously the type was just an alias for `Any`, so we're not missing much by ignoring it.

PR #11820, opened 2022-01-25
Update installation docs to indicate that we support Python 3.10

H-Shay:
As the title states. One of the tasks in #11780.

### DIFF
--- a/changelog.d/11820.misc
+++ b/changelog.d/11820.misc
@@ -1,1 +1,0 @@
-Update pypi installation docs to indicate that we now support Python 3.10.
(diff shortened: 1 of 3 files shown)


changelog.d/11820.misc
```diff
@@ -0,0 +1 @@
+Update pypi installation docs to indicate that we now support Python 3.10.
```
anoadragon453, line 1: This should be a `.doc`.

anoadragon453 (approved):
CI failures are unrelated to changes in this PR. See https://github.com/matrix-org/synapse/pull/11817 for the fix.

Thanks for continuing on with this quest!

PR #11823, opened 2022-01-25
Minor updates, and docs, for schema delta files

richvdh:
1. make the functions in python delta files optional
2. document the things


clokep (approved):
Seems reasonable. I wonder if we can remove these from any past migrations?

richvdh:
> I wonder if we can remove these from any past migrations?

yeah, we could...

PR #11830, opened 2022-01-25
Fix another jsonschema typecheck error

DMRobertson:
Similar to #11817.

In `_create_power_level_validator` we
- retrieve `validator`. This is a class implementing the
  `jsonschema.protocols.Validator` interface. In other words,
  `validator: Type[jsonschema.protocols.Validator]`.
- we then create an second validtor class by modifying the original
  `validator`. We return that class, which is also of type
  `Type[jsonschema.protocols.Validator]`.

So the original annotation was incorrect: it claimed we were returning
an instance of jsonSchema.Draft7Validator, not the class (or a subclass)
itself. (Stricly speaking this is incorrect, because `POWER_LEVELS_SCHEMA`
isn't pinned to a particular version of JSON Schema. But there are other
complications with the type stubs if you try to fix this; I felt like
the change herein was a decent compromise that better expresses intent).

(I suspect/hope the typeshed project would welcome an effort to improve
the jsonschema stubs. Let's see if I get some spare time.)

### DIFF
--- a/changelog.d/11830.misc
+++ b/changelog.d/11830.misc
@@ -1,0 +1,1 @@
+Correct a type annotation in the event validation logic.
--- a/changelog.d/11817.misc
+++ b/changelog.d/11817.misc
@@ -1,1 +1,1 @@
-Compatibility with updated type hints for jsonschema 4.4.0.
+Correct a type annotation in the event validation logic.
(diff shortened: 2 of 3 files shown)


changelog.d/11830.misc
```diff
@@ -0,0 +1 @@
+Correct a type annotation in the event validation logic.
```
clokep, line 1: Can we update `11817.misc` to be this also? 😄

clokep (approved):
This seems reasonable if CI passes!

clokep:
The complement test fail was flaky, so merging anyway.

Issue #11537, opened 2021-12-08
Lock Dependencies in Synapse


callahad:
Hi folks,

Right now most of Synapse's dependencies only declare a minimum version bound, like `Twisted>=18.9.0` ([cite](https://github.com/matrix-org/synapse/blob/release-v1.49/synapse/python_dependencies.py#L50-L90)). This means that every time we build release artifacts or install Synapse from source, we unconditionally pull in the latest versions of our dependencies _at that moment._

This creates unnecessary risk, as our dependencies can change out from under us without warning. For example, installing Synapse 1.49.0 may work today but fail tomorrow if one of our dependencies releases a new version overnight.

This exact scenario bit us with the release of attrs 21.1.0 ([synapse#9936](https://github.com/matrix-org/synapse/issues/9936)) on May 6th. We were forced to release Synapse 1.33.1 less than a day after 1.33.0 as the attrs release broke our ability to install Synapse from source, build working Debian packages, or create functioning Docker images, even though _nothing_ in our repositories had changed.

The absence of locked dependencies also impedes our ability to pursue continuous delivery, maintain LTS releases, or easily backport security fixes as we cannot recreate older release artifacts without _also_ implicitly updating all of the dependencies included therein.

## Definition of Done

* For any given Synapse commit, it is possible to repeatably create identical virtualenvs.

## Further Discussion / Constraints

Resolving this implies that it must be possible to enumerate exact versions of all dependencies included in any given upstream release of Synapse, using only a clone of the Synapse repository. This is important for auditing, as it allows us to easily answer questions like "did we ever ship a release with a vulnerable version of that dependency?"

Moreover, any solution _must_ record hash digests to protect against remote tampering, such as with pip's [hash-checking mode](https://pip.pypa.io/en/stable/cli/pip_install/#hash-checking-mode).

[...]

PR #11834, opened 2022-01-26
Avoid type annotation problems in prom-client

DMRobertson:
Upper bound to < 0.13.0. Not directly pinning it so the package maintainers have flexibility.

#11537 would have helped.

### DIFF
--- a/changelog.d/11834.misc
+++ b/changelog.d/11834.misc
@@ -1,0 +1,1 @@
+Workaround a type annotation problem in `prometheus_client` 0.13.0.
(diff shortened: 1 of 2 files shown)


reivilibre (approved):

reivilibre:
> #11537 would have helped.

Curious how — are we intending to use the pinned deps in CI as well? If so, are we going to have any CI for the 'latest' deps too so we get some kind of gradual warning about upcoming issues rather than a large batch of them appearing at once when we update the deps?

DMRobertson:
> > #11537 would have helped.
> 
> Curious how — are we intending to use the pinned deps in CI as well? 

Certainly for deps that are required at runtime. Unsure what to do about dev-time dependencies like mypy, but I was planning to do the same.

> If so, are we going to have any CI for the 'latest' deps too so we get some kind of gradual warning about upcoming issues rather than a large batch of them appearing at once when we update the deps?

Dependabot might provide this---my understanding is that it will periodically open PRs which bump dependencies to their newest versions. That may be sufficient (though it wouldn't account for the Twisted trunk CI job.

PR #11835, opened 2022-01-26
Prevent duplicate push notifications for room reads

### LINKED ISSUE #11704: Useless pushes with no changes sent when reading rooms with no notifs enabled
### Description

Duplicate pushes are sent whenever you read a room, even if notifications are disabled for this room and there is no change in the notifs count.

### Steps to reproduce

- read a room with no notifs (read receipt is sent)
- a push is sent, identical to the previous push, with no event_id/room_id and an unchanged `count`

This is annoying for big accounts that are also extensively used on a computer device, because the phone basically never sleeps.

lukasdenk:
Fixes #11704.

Make a `POST` to `/rooms/<room_id>/receipt/m.read/<event_id>` only trigger a push notification when the count of unread messages is different to the one in the last successfully sent push.

@reivilibre:
-  As I believe, every message sent to a room triggers a push notification to be sent. So is there actually a scenario in which a `POST` to `/rooms/<room_id>/receipt/m.read/<event_id>` will still trigger a push now?
- I have not written an extra test since `test_push_unread_count_message_count` will now fail in case a `POST` to `/rooms/<room_id>/receipt/m.read/<event_id>` erroneaously causes a push notification to be sent.

### Pull Request Checklist

### DIFF
--- a/tests/push/test_http.py
+++ b/tests/push/test_http.py
@@ -572,7 +572,7 @@ class HTTPPusherTests(HomeserverTestCase):
         #
         # This push should still only contain an unread count of 1 (for 1 unread room)
         self.assertEqual(
-            self.push_attempts[5][2]["notification"]["counts"]["unread"], 1
+            self.push_attempts[4][2]["notification"]["counts"]["unread"], 1
         )
 
     @override_config({"push": {"group_unread_count_by_room": False}})
@@ -588,7 +588,7 @@ class HTTPPusherTests(HomeserverTestCase):
         # We're counting every unread message, so there should now be 4 since the
         # last read receipt
         self.assertEqual(
-            self.push_attempts[5][2]["notification"]["counts"]["unread"], 4
+            self.push_attempts[4][2]["notification"]["counts"]["unread"], 4
         )
 
     def _test_push_unread_count(self):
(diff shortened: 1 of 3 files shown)


lukasdenk:
Made WIP because I still need to change sth small

tests/push/test_http.py
```diff
@@ -673,30 +673,20 @@ def _test_push_unread_count(self):
         )
```
reivilibre, line 673: there's a comment just above 'This will actually trigger a new notification to be sent out so that even if the user does not receive another message, their unread count goes down'...

It doesn't make much sense to me though because it was **already** at zero right before. 

I would say 'I think it may be worth removing the comment', but this test is trying to test that pushes go out when you read something. As it stands, we're not testing that anymore.
Do you think you can adapt the test so that it will actually get the chance to send a push after a read receipt?

The test currently looks like:

* send message → send count of 0 (because we haven't got a read receipt in the room yet)
* add read receipt → send count of 0 (because we sent a read receipt) — **your changes stop this from happening**, so we don't actually end up testing that read receipts do anything
* send message → send count of 1

I wonder if it should look like this:

* send message → send count of 0 (because we haven't got a read receipt in the room yet)
* add read receipt → **make sure** nothing new has been sent (that's still a valuable piece of test information) — do it with an assert on len(push_attempts)
* send message → send count of 1
* add read receipt (to the second message) → **make sure** we send a count of 0 
* send message → send count of 1

reivilibre (changes requested):
Looks good, thank you — just worth checking the tests are still testing what they're meant to test

tests/push/test_http.py
lukasdenk: Ok, I think I will start working on that issue by the beginning of next week.

lukasdenk: @reivilibre: I still refactored the test a bit. If you do not like it, the commit before preserves
 the old coding style.

tests/push/test_http.py
```diff
@@ -642,81 +643,87 @@ def _test_push_unread_count(self):
...
+
+        self._check_push_attempt(expected_push_attempts, 1)
+
+        self._send_read_request(access_token, second_message_event_id, room_id)
+        expected_push_attempts += 1
+
+        self._check_push_attempt(expected_push_attempts, 0)
+
+        # Since we're grouping by room, sending more messages shouldn't increase the
+        # unread count, as they're all being sent in the same room
```
reivilibre, line 671: Although you say that sending more messages shouldn't increase the unread count, you're not actually checking that yet (I think)?
I think you want to use `_check_push_attempt` after each send. At the very least, that will make it very easy to follow and count along!

reivilibre: (I now notice that this is just how the code was before... but if you wouldn't mind enhancing it, that would be appreciated!)

reivilibre (changes requested):
This is good, thanks: almost there!

tests/push/test_http.py
lukasdenk: I updated the comment now.

reivilibre (approved):

reivilibre:
Thanks for this one! I know it will be appreciated!

MatMaul:
Thanks a lot @lukasdenk !

PR #11836, opened 2022-01-26
Create singletons for `StateFilter.{all,none}()`

richvdh:
No point recreating these for each call, since they are frozen


clokep (approved):
/me stares longingly at [`cached_property`](https://docs.python.org/dev/library/functools.html#functools.cached_property).

PR #11837, opened 2022-01-26
Experimental support to include bundled aggregations in search results (MSC3666)

### LINKED ISSUE #11824: Implement MSC3666: Bundled aggregations for /search
[MSC3666](https://github.com/matrix-org/matrix-doc/pull/3666) defines that `/search` should include results with bundled aggregations.

This should be fairly straightforward. 👍

clokep:
Implements [MSC3666](https://github.com/matrix-org/matrix-doc/pull/3666) behind a config flag. I think the main risk with this is that it adds significant load to the `/search` results, but I'm not really sure how to test that.

Fixes #11824


synapse/handlers/search.py
```diff
@@ -238,8 +238,6 @@ async def search(
 
             results = search_result["results"]
 
-            results_map = {r["event"].event_id: r for r in results}
-
```
richvdh, line 242: ugh. If this function wasn't 50000 lines long, we'd be able to follow this sort of thing much more easily.

clokep: Yeah, PyCharm noticed it and declared it unused. 😢 

I can refactor this method a bit first if you'd like. I think removing the `context` calculation would be easy enough.

richvdh (commented):
seems sensible to me

synapse/handlers/search.py
richvdh: it would certainly be nice to refactor it if you have a few tuits. Suggest doing as a followup though.

synapse/handlers/search.py
```diff
@@ -418,12 +420,29 @@ async def search(
...
+                # returned, including contextual events.
+                itertools.chain(
+                    # The events_before and events_after for each context.
+                    itertools.chain.from_iterable(
+                        itertools.chain(context["events_before"], context["events_after"])  # type: ignore[arg-type]
+                        for context in contexts.values()
+                    ),
+                    # The returned events.
+                    allowed_events,
+                ),
```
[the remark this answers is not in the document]
clokep: Hmm, the results are cached so hopefully we would just immediately hit the cache, but depending on that seems not very nice. Good call! 👍

synapse/handlers/search.py
clokep: I took a look at doing this and it quickly spiraled. Will do as a follow-up.

synapse/storage/databases/main/relations.py
```diff
@@ -682,10 +682,20 @@ async def get_bundled_aggregations(
             A map of event ID to the bundled aggregation for the event. Not all
             events may have bundled aggregations in the results.
         """
+        # The already processed event IDs. Tracked separately from the result
+        # since the result omits events which do not have bundled aggregations.
+        seen_events = set()
```
clokep, line 687: I was very much hoping to be able to do `events = set(events)`, but I don't think that would work as events aren't comparable as far as I can tell.

richvdh (approved):
lgtm!

PR #11843, opened 2022-01-27
Remove the obsolete MSC1849 flag.

clokep:
MSC1849 was replaced with MSC2675, which has now been merged. Thus we can remove the experimental config flag (which defaulted to `True` anyway).


reivilibre (approved):
Looks correct!

PR #11844, opened 2022-01-27
Configurable limits on avatars

babolivier:
Only allow files which file size and content types match configured
limits to be set as avatar.

TODO: write tests


babolivier:
Looks like I've messed up a rebase - closing in favour of https://github.com/matrix-org/synapse/pull/11846

Issue #10701, opened 2021-08-26
Allow modules to run only on specific worker(s)


anoadragon453:
There are some usecases for wanting a Synapse module to execute functionality once when it is running on a specific worker. Usecases include:

* A module that auto-accepts invites for certain users. Accepting the invite should only be done by a single instance of the module - not all of them at once.
* A module that wishes to send presence to other modules. This should only be done by one instance of a module (to avoid sending duplicate presence), and should only be done on workers that support sending federation traffic.

To achieve this, it is sometimes suggested to add worker-specific config for a module, in order to tell a module running on one worker to carry out a task, but to not do the same for the module config on other workers.

This method is fragile though - not least because you may wish to share module config between all worker instances, and having to have separate config per-worker is overhead to manage. Additionally, because homeserver config keys are overwritten when specifying multiple homeserver config files (as is commonly done with workers), you can't really configure a worker in one way on the main process, and in a subtly different way on the worker. The worker's config will need to be entirely duplicated.

As a proposed solution, we simply make the worker's name (`None` if the main process) available to modules. That way, we could simply state in the module's config that certain functionality should only run if the name of the worker that the module is running on matches against some configured value(s).

dklimpel:
There is a workaround for disable modules in workers, you can set workers configuration:
```yaml
modules: []
```

This disable the modules on workers. But it only helps if you disable modules on all workers an run the module in main process.

babolivier:
> As a proposed solution, we simply make the worker's name (`None` if the main process) available to modules.

For some reason I was sure this was already something we do, but apparently it isn't.

> That way, we could simply state in the module's config that certain functionality should only run if the name of the worker that the module is running on matches against some configured value(s).

To clarify, the module would be the one responsible for figuring out whether something can run based on the worker name, right? This sentence can be read in a way that implies we add some sort of special-cased config key to module configs to have Synapse automatically run the module on a specific worker.

anoadragon453:
> To clarify, the module would be the one responsible for figuring out whether something can run based on the worker name, right? This sentence can be read in a way that implies we add some sort of special-cased config key to module configs to have Synapse automatically run the module on a specific worker.

What I'm thinking is that you would have a supported field in your worker config that's something along the lines of:

```yaml
config:
    do_flarb_if_worker_name: flarb_runner
```

And then in your module code, have a condition similar to:

```python
 def flarb():
    if self.module_api.worker_name != self.config.do_flarb_if_worker_name:
[...]

PR #11845, opened 2022-01-27
Allow modules to run only on specific worker

dklimpel:
Fixes: #10701

This is not the requested solution of #10701
I am not sure if a module should decide independently where it runs.

This makes it possible to configure the workers at a central location (homerserver.yaml).
I am not sure if a config value is needed to run a module on all servers with the same configuration.

### Pull Request Checklist

### DIFF
--- a/changelog.d/11845.feature
+++ b/changelog.d/11845.feature
@@ -1,0 +1,1 @@
+Allow modules to run only on specific worker.
--- a/docs/modules/index.md
+++ b/docs/modules/index.md
@@ -14,14 +14,19 @@ To use a module on Synapse, add it to the `modules` section of the configuration
 ```yaml
 modules:
   - module: my_super_module.MySuperClass
+    worker_name: None
     config:
       do_thing: true
   - module: my_other_super_module.SomeClass
+    worker_name: my_module_worker
     config: {}
 ```
 
 Each module is defined by a path to a Python class as well as a configuration. This
 information for a given module should be available in the module's own documentation.
+
+`worker_name` defines on wich worker the module is loaded. If `worker_name` is not set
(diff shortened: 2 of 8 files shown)


babolivier:
>  I am not sure if a module should decide independently where it runs.

This is kind of already what we do with `ModuleApi.looping_background_call`, i.e. allowing the module to decide whether to run its background process on all of the processes or just the one in charge of background jobs.

Personally I'm more keen to delegating the logic of figuring out what to run on which worker. The module interface rewrite was in part motivated by the fact that the old module system would force you to write split your module if you wanted part of it to implement interface X and part of it to implement interface Y (which meant the only way to implement both at once was to get really hacky, e.g. sharing state via global variables, etc).
It feels like going backwards to me to force module developers to split their module just because they don't want to run method A and method B on the same worker. I think the modules themselves should have the possibility to decide which worker each functionality runs on, just like they have the possibility to decide which callback categories they want to implement.

anoadragon453:
Additionally, this solution only allows for running a module on a single worker, whereas there's often cases where modules should run across a set of multiple workers.

babolivier:
> > Additionally, this solution as it stands only allows for running a module on a single worker, whereas there's often cases where modules should run across a set of multiple workers.
> 
> IMO you can run with this solution a module on multiple workers: e.g.:
> 
> ```yaml
> modules:
>   - module: my_super_module.MySuperClass
>     worker_name: Worker1
>   - module: my_super_module.MySuperClass
>     worker_name: Worker2
> ```

babolivier:
(sorry, I missclicked)

dklimpel:
I close the PR and wait for another solution. If this kind of solution has a chance, I can reopen it.

Issue #1297, opened 2015-03-13
Changing displayname is catastrophically slow. (SYN-311)


matrixbot:
It takes roughly 70s for me to change my displayname, and 100% cpu on matrix.org.

(Imported from https://matrix.org/jira/browse/SYN-311)

(Reported by @ara4n)

deepbluev7:
I'd say this is still an issue. For example running Nhekos unit tests, if you run them twice on the same docker container, the display name changes (and avatar changes) are atrociously slow. The second run usually means, that you have 20 to 40 rooms already (didn't count them exactly). This results in the following runtimes:

```
[==========] Running 37 tests from 2 test suites.
[----------] Global test environment set-up.
[----------] 35 tests from ClientAPI
[ RUN      ] ClientAPI.Register
[       OK ] ClientAPI.Register (400 ms)
[ RUN      ] ClientAPI.LoginSuccess
[       OK ] ClientAPI.LoginSuccess (565 ms)
[ RUN      ] ClientAPI.LoginWrongPassword
[       OK ] ClientAPI.LoginWrongPassword (242 ms)
[ RUN      ] ClientAPI.LoginWrongUsername
[       OK ] ClientAPI.LoginWrongUsername (13 ms)
[ RUN      ] ClientAPI.LoginFlows
[       OK ] ClientAPI.LoginFlows (10 ms)
[ RUN      ] ClientAPI.SSORedirect
[       OK ] ClientAPI.SSORedirect (1 ms)
[ RUN      ] ClientAPI.EmptyUserAvatar
[       OK ] ClientAPI.EmptyUserAvatar (48359 ms)
[ RUN      ] ClientAPI.RealUserAvatar
[       OK ] ClientAPI.RealUserAvatar (47640 ms)
[ RUN      ] ClientAPI.ChangeDisplayName
[       OK ] ClientAPI.ChangeDisplayName (48443 ms)
[ RUN      ] ClientAPI.EmptyDisplayName
[       OK ] ClientAPI.EmptyDisplayName (49116 ms)
[ RUN      ] ClientAPI.CreateRoom
[       OK ] ClientAPI.CreateRoom (2431 ms)
[ RUN      ] ClientAPI.TagRoom
[       OK ] ClientAPI.TagRoom (3045 ms)
[ RUN      ] ClientAPI.LogoutSuccess
[       OK ] ClientAPI.LogoutSuccess (2509 ms)
[...]

PR #11846, opened 2022-01-27
Configurable limits on avatars

babolivier:
Only allow files which file size and content types match configured
limits to be set as avatar.

Most of the inspiration from the non-test code comes from https://github.com/matrix-org/synapse-dinsic/pull/19


synapse/config/server.py
```diff
@@ -1168,6 +1181,22 @@ def generate_config_section(
         #
         #allow_per_room_profiles: false
 
+        # The largest allowed size for a user avatar. If not defined, no
+        # restriction will be imposed.
```
clokep, line 1185: My initial question was if this was bytes or pixel size -- from the value being `10M` I see it is bytes. Not sure it is worth clarifying.

synapse/config/server.py
```diff
@@ -1168,6 +1181,22 @@ def generate_config_section(
...
+        #
+        # Note that user avatar changes will not work if this is set without
+        # using Synapse's media repository.
```
clokep, line 1188: Should these settings be part of `synapse/config/repository.py` instead?

synapse/handlers/profile.py
```diff
@@ -307,6 +316,76 @@ async def set_avatar_url(
...
+            # of a PUT request on /avatar_url that the file was at least previewed by the
+            # user locally (and therefore downloaded to the remote media cache).
+            logger.warning("Forbidding avatar change to %s: avatar not on server", mxc)
+            return False
+
+        if self.max_avatar_size:
+            # Ensure avatar does not exceed max allowed avatar size
+            if media_info["media_length"] > self.max_avatar_size:
+                logger.warning(
+                    "Forbidding avatar change to %s: avatar must be less than %s bytes",
```
clokep, line 368: Is it more worthwhile to log the size of the requested avatar instead of the max size? (This matches what we do for the mime-type below).

synapse/handlers/profile.py
```diff
@@ -307,6 +316,76 @@ async def set_avatar_url(
...
+                    mxc,
+                    self.max_avatar_size,
+                )
+                return False
+
+        if self.allowed_avatar_mimetypes:
+            # Ensure the avatar's file type is allowed
+            if (
+                self.allowed_avatar_mimetypes
+                and media_info["media_type"] not in self.allowed_avatar_mimetypes
```
clokep, line 378: I don't think so, but can `media_info["media_type"]` be `null`?

tests/handlers/test_profile.py
```diff
@@ -248,3 +249,104 @@ def test_set_my_avatar_if_disabled(self):
...
+        # don't get an error on a non-existing file then it means all of the checks were
+        # successfully skipped.
+        allowed = self.get_success(
+            self.handler._check_avatar_size_and_mime_type("mxc://test/unknown_file")
+        )
+        self.assertTrue(allowed)
+
+    @unittest.override_config(
+        {
+            "max_avatar_size": 50,
```
clokep, line 267: Not a requirement, but if you remove the trailing comma the entire override should collapse to a single line above the test. 👍

tests/handlers/test_profile.py
```diff
@@ -248,3 +249,104 @@ def test_set_my_avatar_if_disabled(self):
...
+
+        allowed = self.get_success(
+            self.handler._check_avatar_size_and_mime_type("mxc://test/small")
+        )
+        self.assertTrue(allowed)
+
+        allowed = self.get_success(
+            self.handler._check_avatar_size_and_mime_type("mxc://test/big")
+        )
+        self.assertFalse(allowed)
```
clokep, line 303: For some reason the combination of the variable being called `allowed` and the call to `get_success` made me really confused at why this wasn't rejected until I realized it was `assertFALSE`, not `assertTrue`.

Maybe calling it `result` instead of `allowed` would be better? (Same goes for the other tests.)

tests/rest/client/test_profile.py
```diff
@@ -150,6 +155,173 @@ def _get_avatar_url(self, name=None):
...
+            access_token=self.owner_tok,
+        )
+        self.assertEqual(channel.code, 403, channel.result)
+        self.assertEqual(
+            channel.json_body["errcode"], Codes.FORBIDDEN, channel.json_body
+        )
+
+        channel = self.make_request(
+            "PUT",
+            "/profile/%s/avatar_url" % (self.owner,),
```
clokep, line 187: Maybe f-strings if you like?

```suggestion
            f"/profile/{self.owner}/avatar_url",
```

synapse/config/server.py
babolivier: Not sure, I see them more as profile-related settings since the action of uploading to the media repo won't be blocked on size and type.

synapse/handlers/profile.py
babolivier: It doesn't look like it. Same for the size/length.

synapse/config/server.py
clokep: I don't feel strongly either way, just figured I'd ask!

babolivier: Fair enough :)

synapse/handlers/profile.py
```diff
@@ -307,6 +317,63 @@ async def set_avatar_url(
 
         await self._update_join_states(requester, target_user)
 
+    @cached()
```
clokep, line 320: Not a hug deal, but do we expect this to be called often enough with the same `mxc` that we should cache it?

clokep (approved):
Looks good, left one comment but not a blocker!

synapse/handlers/profile.py
babolivier: Yes, because this is called both when setting your global profile and when setting a per-room one. Except setting a global profile then updates your per-room profile in every room you're in. So if I'm in hundreds of rooms then a single global avatar change is going to trigger this function to be called hundreds of times.

clokep: Right! Pretty much trying to make #1297 not worse!

babolivier: Yep exactly!

PR #11847, opened 2022-01-27
Pass `isolation_level` to `runWithConnection`

babolivier:
This was missed in https://github.com/matrix-org/synapse/pull/11799


Fizzadar (approved):
Thank you for catching and fixing this @richvdh @babolivier!

clokep (approved):

Issue #5584, opened 2019-07-01
Document configuration option vocabulary/taxonomy and then simplify 


michaelkaye:
A document that contains the vocabulary of the synapse configuration and the expected values of those named fields would assist in keeping code consistent.

A few things have popped up recently, here and in other branches, so I thought i'd start an issue to collect examples.

I think we should first generate a statement on what the best naming scheme is for things, aim to apply that in future, then come back and deprecate/rename existing options to bring them into line.

Related issues:
* https://github.com/matrix-org/synapse/issues/3913 (we should be clear this is used as a prefix or a template)
* https://github.com/matrix-org/synapse/issues/2326 (we should be clear on what is a template and what isn't)
* https://github.com/matrix-org/synapse/issues/3523 (we should be consistent in requiring and handling trailing slashes in URLs in our configuration)

michaelkaye:
For instance, we have secrets with different key names:
* `pepper` - a secret 
* `macaroon_secret_key` - another secret
* `registration_shared_secret` - another secret

Secrets should be obviously a secret, and should match the form `XXX_secret`

We have:
* `pid_file` - a path to the writable pid file
* `signing_key_path` - a path to a readable secret file
* `uploads_path` - a path to a writable directory
* `log_config` - a path to a file, not a configuration itself.

All paths should match the form `XXX_path` if they're for directories
All paths should match the form `XXX_file` if they're to a single file

michaelkaye:
We also have:
* `turn_allow_guests`- a boolean
* `allow_guest_access` - another boolean
* `enable_metrics` - a boolean
* `disable_set_displayname` - another boolean
* `report_stats` - a boolean.
* `use_presence` - another boolean

This should be something like "all boolean-type enabling options should use the format: "enable_XXX" (or if we want both enable and disable as defaultable things, prefix with "enable_XXX" or "disable_XXX") - but what we shouldn't do is mix `enable`, `allow`, `use`, `report` and so on.

richvdh:
Another contender: durations:
* `email.validation_token_lifetime`
* `key_refresh_interval`
* `account_validity.period`
* `account_validity.renew_at`
* `saml2.saml_session_lifetime`
* `stats.bucket_size`
* `stats.retention`
* `turn_user_lifetime`
* `rc_federation.selep_delay`

Most (all?) of the above can be expressed with units eg `6h`, `2d`.

* `mau_trial_days` (in days rather than milliseconds)

ptman:
[...]

PR #11849, opened 2022-01-27
Enable cache time-based expiry by default

### LINKED ISSUE #11787: Enable cache time-based expiry by default
The 'new' time-based expiry for caches in Synapse seems to have helped out with memory usage (empirically, based on what people are telling me).
It seems like it would be worth enabling by default so that people don't have to know about this obscure option to get the benefits.

Related: #11641 (it's not obvious that even though Synapse's memory usage only grows by default, it is capable of trimming back down).

H-Shay:
Changes the default config option for time based cache expiry from None to 30m. Fixes #11787, where you will also find more discussion on this change.


changelog.d/11849.misc
```diff
@@ -0,0 +1 @@
+Enable cache time-based expiry by default.
```
anoadragon453, line 1: It's not clear to me here that my existing Synapse installation won't start expiring caches based on time. Perhaps something like:
```suggestion
Enable cache time-based expiry by default for newly generated homeserver config files.
```

anoadragon453 (changes requested):
Note that this only enables cache expiry by default for newly generated config files, not existing installations. This is good because:

1. We don't want to change settings out from under people.
2. There isn't a way to differentiate between "someone's never set this option" and "someone explicitly disabled this option".

So this looks like the right solution to me.

reivilibre:
> Note that this only enables cache expiry by default for newly generated config files, not existing installations. This is good because…

I actually am not so sure (and it's not the impression I got from the meeting yesterday where we mentioned this).

I think we should enable it by default for existing configs because the original point about it being an obscure option still stands and generally homeservers should benefit from this change.

> There isn't a way to differentiate between "someone's never set this option" and "someone explicitly disabled this option".

Setting it to null would explicitly disable the option; on the other hand, once we ship a default config file with options that aren't the default in the config loaders, we lose the ability to distinguish '30 minutes, explicit' and 'no preference'.
We also might cause strange behaviour later on if we decide to remove this option and start showing deprecation warnings and errors to operators *who never set the option manually anyway*.

I find it strange that we would enable an option by default only for 'new' homeservers — I appreciate the caution but it also conflicts with the intuition that a homeserver admin only needs to have the options they changed present in their config file.

By leaving an option unmarked in the config, I would interpret this to mean 'I, the admin, don't care and want the software to do the default reasonable behaviour' — in this case I would say the time-based expiry is a great boon for memory usage and seems like the most reasonable behaviour when the admin hasn't expressed a preference.

There is a case to be made that we should stop documenting the options inside the config file and instead have a nice config manual or something, because the comment will become out of date for old installations. Still, we could announce that in the upgrade notes in case anyone was caught out by the change in default behaviour.

H-Shay:
Fascinating-it never occurred to me that we would change the behavior of existing Synapse homeservers. My understanding was something along the lines of: "we've used this setting and feel confident it's beneficial, we will make it the default going forwards." My personal feeling is that it is a little strange to change the default behavior out from under people rather than having them explicitly op-in to the *new* default (which I figured we would shout about in the release notes) but that's just a feeling. I am happy to implement whatever the consensus determines is the best approach.  
Also plus one for having a config manual vs comments, I will open an issue about that as I think it would be beneficial.

richvdh:
> My personal feeling is that it is a little strange to change the default behavior out from under people rather than having them explicitly op-in to the new default (which I figured we would shout about in the release notes) but that's just a feeling. 

This is certainly a reasonable argument, but I think the point is that most people don't want to have to think this hard about maintaining their deployments - they want it to "just work" and it's up to us to make it do so as best we can, which includes changing the default behaviour where we have reason to believe that will improve things for the majority of users.

There could be a *minority* who would prefer to retain the existing behaviour, but they can update their config files if they want to.

It's worth noting that we change Synapse's default behaviour all the time when it comes to bugfixes and the like. You could argue that not having time-based expiry enabled is a bug.

anoadragon453:
Thanks for the course-correction folks! There may be a small user-base that read the option and left it commented, assuming that would remain disabled. We would then be changing the behaviour for them.

However, shouting about it in the changelog (or perhaps upgrade notes if we deem necessary) should be enough for even this theoretical edge case.

anoadragon453 (changes requested):
Going forward on this PR, I think we should:

1. Change the default (when the option is not specified) of `expiry_time` to `30m`.
2. Assuming we'd like to continue allowing people to disable cache expiry, modify the sample config to make disabling the option a little simpler. We could either:
    * tell people to write `expiry_time: null` to disable the option, but this isn't really consistent with the rest of the config file.
    * introduce a new `expire_caches` boolean option (defaulting to `true`), which would be my personal preference.

callahad:
> tell people to write `expiry_time: null` to disable the option, but this isn't really consistent with the rest of the config file.

Would `expiry_time: 0` or `expiry_time: -1` be acceptable? Those seem to be common enough conventions in other software, but I don't see much precedent within Synapse. Adding a whole 'nother option seems pretty heavyweight, so I'd bias toward overloading the time.

richvdh:
`expiry_time` is a pretty rubbish name for it, tbh (cf https://github.com/matrix-org/synapse/issues/5584#issuecomment-509284013). It might be a reasonable time to sort that out.

H-Shay:
Thanks everyone for chiming in, I will make these changes. Also thanks for taking time to explain your reasoning, it's very helpful for me to hear.

H-Shay:
Per Rich's comment I renamed the user-facing flag to something (hopefully) a little more descriptive, but I wasn't sure if I should rename the internal var as well.

synapse/config/cache.py
```diff
@@ -217,9 +221,16 @@ def read_config(self, config, **kwargs) -> None:
                     e.message  # noqa: B306, DependencyException.message is a property
                 )
 
-        expiry_time = cache_config.get("expiry_time")
-        if expiry_time:
-            self.expiry_time_msec: Optional[int] = self.parse_duration(expiry_time)
+        expire_caches = cache_config.get("expire_caches", True)
+        cache_entry_ttl = cache_config.get("cache_entry_ttl")
+
```
anoadragon453, line 226: Since we're changing config options, we should include backwards-compatibility support for the previous config option name. Thus if someone set `expiry_time` previously, that same value should continue to work even after this update. At some point :tm: in the future we can remove support for the old option name.

synapse/config/cache.py
```diff
@@ -217,9 +221,16 @@ def read_config(self, config, **kwargs) -> None:
...
+        if expire_caches:
+            if cache_entry_ttl:
+                self.expiry_time_msec: Optional[int] = self.parse_duration(
+                    cache_entry_ttl
+                )
+            else:
+                self.expiry_time_msec = 1800000
```
anoadragon453, line 233: May be worth adding a quick comment to indicate that this equates to 30 minutes.

synapse/config/cache.py
```diff
@@ -217,9 +221,16 @@ def read_config(self, config, **kwargs) -> None:
                     e.message  # noqa: B306, DependencyException.message is a property
                 )
 
-        expiry_time = cache_config.get("expiry_time")
-        if expiry_time:
-            self.expiry_time_msec: Optional[int] = self.parse_duration(expiry_time)
+        expire_caches = cache_config.get("expire_caches", True)
+        cache_entry_ttl = cache_config.get("cache_entry_ttl")
```
anoadragon453, line 225: We could also set the default here, via:

```suggestion
        cache_entry_ttl = cache_config.get("cache_entry_ttl", "30m")
```

which may be slightly more intuitive if one is scanning the code looking for default option values.

synapse/config/cache.py
```diff
@@ -148,11 +148,15 @@ def generate_config_section(self, **kwargs) -> str:
           per_cache_factors:
             #get_users_who_share_room_with_user: 2.0
 
-          # Controls how long an entry can be in a cache without having been
-          # accessed before being evicted. Defaults to None, which means
-          # entries are never evicted based on time.
+          # Controls whether cache entries are evicted after a specified time
+          # period. Defaults to true.
+          # expire_caches: false
```
anoadragon453, line 153:
```suggestion
          # Controls whether cache entries are evicted after a specified time
          # period. Defaults to true.
          #
          # expire_caches: false
```

anoadragon453: This block should also have one of those "Uncomment to disable this feature" sentences.

changelog.d/11849.misc
anoadragon453, line 1: It may be worth mentioning the new configuration options that are available, or otherwise pointing to the relevant section of the sample config for the full details.

This could also be done via our [upgrading docs](https://github.com/matrix-org/synapse/blob/develop/docs/upgrade.md), where we have more room.

anoadragon453 (changes requested):
> I wasn't sure if I should rename the internal var as well.

Doing so is fine, as long as it doesn't affect anyone relying on Synapse (such as methods exposed by the module API. These variables are private however, and people shouldn't be reaching into and relying on our private variables.

For Synapse devs, if an in-flight PR relied on the old variable name then they'd discover a merge conflict after your PR had merged - alerting them, and allowing them to correct it.

A few things below, but overall this is looking great!

docs/upgrade.md
```diff
@@ -86,6 +86,13 @@ process, for example:
     ```
 # Upgrading to v1.(next)
 
+## Time-based cache expiry is now enabled by default
+
+Formerly, entries in the cache were not evicted regardless of whether they were accessed after storing.
+This behavior has now changed. By default entries in the cache are now evicted after 30m of not being accessed. 
+To change the default behavior, go to the `caches` section of the config and change the `expire_caches` and 
+`cache_entry_ttl` flags as necessary. Please note that these flags replace the `expiry_time` flag in the config.  
+
```
reivilibre, line 95: I don't know why there's this extra section above `v1.53.0` — I would probably bring them down into 1.53 since we haven't released an RC for it yet.

changelog.d/11849.misc
reivilibre: I would likely still sneak a keyword or two in there: `expiry_time` will be superseded by `expire_caches` and `cache_entry_ttl`.
Helps perk people's ears up and helps if you ever need to do a Ctrl+F for one of the config option names.

synapse/config/cache.py
```diff
@@ -232,7 +235,18 @@ def read_config(self, config, **kwargs) -> None:
...
+            logger.warning(
+                "You have set two incompatible flags, expiry_time and expire_caches. Please only use the "
+                "expire_caches and cache_entry_ttl flags and delete the expiry_time flag as it is "
+                "deprecated."
+            )
         if expiry_time:
+            logger.warning(
+                "Expiry_time is a deprecated flag, please use the expire_caches and cache_entry_ttl flags "
+                "instead."
+            )
```
reivilibre, line 249: This is very nitty, but I would call these *options* rather than *flags*.
At least in my mind, a flag is a boolean.

docs/upgrade.md
```diff
@@ -86,12 +86,6 @@ process, for example:
...
 # Upgrading to v1.(next)
 
-## Time-based cache expiry is now enabled by default
-
-Formerly, entries in the cache were not evicted regardless of whether they were accessed after storing.
-This behavior has now changed. By default entries in the cache are now evicted after 30m of not being accessed. 
-To change the default behavior, go to the `caches` section of the config and change the `expire_caches` and 
-`cache_entry_ttl` flags as necessary. Please note that these flags replace the `expiry_time` flag in the config.  
 
 ## Stablisation of MSC3231
```
reivilibre, line 90: I know it's not your mess, but as you're here, please could you clean this up by removing the 'v1.(next)' heading and bringing the announcement into the v1.53 section?
I personally think a 'v1.(next)' section would be easily forgotten as it's not in our release notes. Or at least, I'd forget! :)

reivilibre (approved):
Looks good; thanks!

docs/upgrade.md
H-Shay: sure thing

synapse/config/cache.py
H-Shay: This is a useful distinction, thank you!

reivilibre (approved):

H-Shay:
@anoadragon453 any chance I can get you to dismiss your stale review when you get the chance?

anoadragon453:
> @anoadragon453 any chance I can get you to dismiss your stale review when you get the chance?

Of course, and note that it's fine to do so yourself as well if the review is outdated (assuming you have permission?).

changelog.d/11849.misc
```diff
@@ -0,0 +1 @@
+Enable cache time-based expiry by default. The `expiry_time` config flag will be superseded by `expire_caches` and `cache_entry_ttl`.
```
richvdh, line 1: this should probably be a `.feature`?

and "will be" should be "has been"?

docs/upgrade.md
```diff
@@ -111,6 +111,13 @@ to:
 
 Please update any relevant reverse proxy or firewall configurations appropriately.
 
+## Time-based cache expiry is now enabled by default
+
+Formerly, entries in the cache were not evicted regardless of whether they were accessed after storing.
+This behavior has now changed. By default entries in the cache are now evicted after 30m of not being accessed. 
+To change the default behavior, go to the `caches` section of the config and change the `expire_caches` and 
+`cache_entry_ttl` flags as necessary. Please note that these flags replace the `expiry_time` flag in the config.
```
richvdh, line 119: Might be worth mentioning that `expiry_time` will still work (for now), but is deprecated?

richvdh (commented):
sorry, a few belated notes. Feel free to commit fixes straight to develop.

This seems to add an empty https://github.com/matrix-org/synapse/blob/develop/synapse/config/background_updates.py, which looks wrong. Please could it be removed?

H-Shay:
Thanks for catching these mistakes-I have made the requested changes.

clokep:
> Thanks for catching these mistakes-I have made the requested changes.

For cross-referencing: 9c4563c5cd97dbcfe3eb43e05b30ca3c830d6937

PR #11852, opened 2022-01-28
Build Docker images for using worker-mode Synapse in automated test tools.

reivilibre:
This is wanted by mx-tester.


.github/workflows/docker.yml
```diff
@@ -63,3 +63,69 @@ jobs:
...
+
+      - name: Set up Docker Buildx
+        id: buildx
+        uses: docker/setup-buildx-action@v1
+
+      - name: Inspect builder
+        run: docker buildx inspect
+
+      - name: Login to GitHub Container Registry (for worker-testing-only image)
+        if: github.event_name != 'pull_request'
```
richvdh, line 93: it looks like the push is not conditional on this, so why is the login?

(maybe we can make the entire `job` conditional instead of the individial steps?)

.github/workflows/docker.yml
```diff
@@ -63,3 +63,69 @@ jobs:
...
+          platforms: arm64
+
+      - name: Set up Docker Buildx
+        id: buildx
+        uses: docker/setup-buildx-action@v1
+
+      - name: Inspect builder
+        run: docker buildx inspect
+
+      - name: Login to GitHub Container Registry (for worker-testing-only image)
```
richvdh, line 92: the "(for worker-testing-only image)" is a bit weird here. Why only this step?

.github/workflows/docker.yml
```diff
@@ -63,3 +63,69 @@ jobs:
...
+        if: github.event_name != 'pull_request'
+        uses: docker/login-action@v1
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      # This must match EXACTLY the one in the `build` pipeline.
```
richvdh, line 100: could use a yaml alias thingy to avoid the duplication

.github/workflows/docker.yml
```diff
@@ -63,3 +63,69 @@ jobs:
...
+              refs/tags/*)
+                  tag=${GITHUB_REF#refs/tags/}
+                  ;;
+              *)
+                  tag=${GITHUB_SHA}
+                  ;;
+          esac
+          echo "::set-output name=tag::$tag"
+
+      # This image is solely intended to be used for automated test tools,
```
richvdh, line 120: .oO (we should probably use https://github.com/peter-evans/dockerhub-description to upload/sync the image descriptions on ghcr/dockerhub)

.github/workflows/docker.yml
DMRobertson: `anchor` and `merge` are the formal names, I think.

reivilibre: Apparently GHA doesn't support YAML anchors. I was surprised but this is what the net says.

reivilibre:
Turns out this wasn't flexible enough, so mx-tester builds its own image now (I think).

PR #11854, opened 2022-01-28
Add a callback to allow modules to deny 3PID

babolivier:
Part of the Tchap Synapse mainlining.

This allows modules to implement extra logic to figure out whether a given 3PID can be added to the local homeserver. In the Tchap use case, this will allow a Synapse module to interface with the custom endpoint `/internal_info` described [here](https://github.com/matrix-org/sydent/blob/dinsic/docs/info.md).

### DIFF
--- a/docs/modules/password_auth_provider_callbacks.md
+++ b/docs/modules/password_auth_provider_callbacks.md
@@ -171,14 +171,15 @@ generated).
 _First introduced in Synapse v1.52.0_
 
 ```python
-async def is_3pid_allowed(self, medium: str, address: str) -> bool
+async def is_3pid_allowed(self, medium: str, address: str, registration: bool) -> bool
 ```
 
 Called when attempting to bind a third-party identifier (i.e. an email address or a phone
 number). The module is given the medium of the third-party identifier (which is `email` if
-the identifier is an email address, or `msisdn` if the identifier is a phone number). The
-module must return a boolean indicating whether the identifier can be allowed to be bound
-to an account on the local homeserver.
+the identifier is an email address, or `msisdn` if the identifier is a phone number) and
+its address, as well as a boolean indicating whether the attempt to bind is happening as
+part of registering a new user. The module must return a boolean indicating whether the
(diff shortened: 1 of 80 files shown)


docs/modules/password_auth_provider_callbacks.md
```diff
@@ -166,6 +166,24 @@ any of the subsequent implementations of this callback. If every callback return
 the username provided by the user is used, if any (otherwise one is automatically
 generated).
 
+## `is_3pid_allowed`
+
+_First introduced in Synapse v1.52.0_
```
babolivier, line 171: Self-reminder to update this if this PR is merged before the 1.52.0rc1 cut-off.

babolivier:
```suggestion
_First introduced in Synapse v1.53.0_
```

docs/modules/password_auth_provider_callbacks.md
```diff
@@ -166,6 +166,25 @@ any of the subsequent implementations of this callback. If every callback return
...
+
+_First introduced in Synapse v1.53.0_
+
+```python
+async def is_3pid_allowed(self, medium: str, address: str, registration: bool) -> bool
+```
+
+Called when attempting to bind a third-party identifier (i.e. an email address or a phone
+number). The module is given the medium of the third-party identifier (which is `email` if
+the identifier is an email address, or `msisdn` if the identifier is a phone number) and
```
erikjohnston, line 179: We should probably note that new `medium`s could potentially be added in the future, so implementors should handle that? (It is somewhat unlikely, but :shrug: )

erikjohnston (approved):
Seems sensible to me :+1:

docs/modules/password_auth_provider_callbacks.md
babolivier: I'll open an issue since that's not the only place we say `medium` can only have one of two values.

babolivier: Opened as https://github.com/matrix-org/synapse/issues/11935

PR #11856, opened 2022-01-28
Add missing type hints to replication.http.

clokep:
Fully types `synapse.replication.http`.

I have a follow-up that does the rest of `synapse.replication` and forbids untyped defs.


synapse/replication/http/_base.py
```diff
@@ -144,19 +150,20 @@ async def _serialize_payload(**kwargs):
...
-            tuple[int, dict]: HTTP status code and a JSON serialisable dict
-            to be used as response body of request.
+            HTTP status code and a JSON serialisable dict to be used as response
+            body of request.
         """
-        pass
 
     @classmethod
-    def make_client(cls, hs: "HomeServer"):
+    def make_client(cls, hs: "HomeServer") -> Callable:
```
clokep, line 166: This is pretty weak, but the inputs depend on `_serialize_payload`, so I'm not really sure if there's anything better to do?

synapse/replication/http/_base.py
```diff
@@ -182,7 +189,7 @@ def make_client(cls, hs: "HomeServer"):
             )
 
         @trace(opname="outgoing_replication_request")
-        async def send_request(*, instance_name="master", **kwargs):
+        async def send_request(*, instance_name: str = "master", **kwargs: Any) -> Any:
```
clokep, line 192: Again this is very weak, but I think they can return anything? This matches the definition of `request_func` below.

synapse/replication/http/account_data.py
```diff
@@ -48,14 +52,18 @@ def __init__(self, hs: "HomeServer"):
         self.clock = hs.get_clock()
 
     @staticmethod
-    async def _serialize_payload(user_id, account_data_type, content):
+    async def _serialize_payload(  # type: ignore[override]
```
clokep, line 55: Each instance of this have different arguments, defined as `kwargs` in the super-class. Not sure if there's something better to do than ignore the override?

synapse/replication/http/login.py
```diff
@@ -91,8 +96,8 @@ async def _handle_request(self, request, user_id):
             auth_provider_session_id=auth_provider_session_id,
         )
 
-        return 200, res
+        return 200, cast(JsonDict, res)
```
clokep, line 99: `register_device_inner` returns a `TypedDict`, so we just tell mypy to go away.

synapse/replication/http/federation.py
```diff
@@ -248,14 +263,16 @@ def __init__(self, hs: "HomeServer"):
         self.store = hs.get_datastore()
 
     @staticmethod
-    async def _serialize_payload(room_id, args):
```
clokep, line 251: The only caller of this seems to only pass a `room_id`, I'm unsure if this is fixing a bug or not:

https://github.com/matrix-org/synapse/blob/251b5567ecc8fb3d6debaa3f77f6ec2620877d36/synapse/handlers/federation.py#L1379-L1389

synapse/replication/http/_base.py
clokep: I think we could do something better, yes. I don't think it is worth holding up the PR for it though. Would adding a TODO help? Or maybe filing an issue?

synapse/replication/http/_base.py
clokep: Yeah this is the whole "how do you represent a JSON result" issue.

synapse/replication/http/_base.py
DMRobertson: I think mypy will complain (with the right `--option`) about a generic type like `Callable` used without fully specifying its parameters, so I don't think this one needs to be called out in particular.

synapse/replication/http/federation.py
clokep: It seems to be called via `do_invite_join` which is definitely called, it seems to eventually come from either `RoomMemberMaster` or `RoomMemberWorker`, maybe we're not routing anything that calls the worker version to the worker?

clokep:
@DMRobertson I think this is back to you? I'm not sure what to do about the [`_serialize_payload` change discussed above](https://github.com/matrix-org/synapse/pull/11856#discussion_r794798813). The rest I'm fairly comfortable with!

synapse/replication/http/federation.py
DMRobertson: I think there is a clue here:

https://github.com/matrix-org/synapse/blob/251b5567ecc8fb3d6debaa3f77f6ec2620877d36/synapse/handlers/federation.py#L452-L474

I couldn't see any other call site of `_clean_room_for_join`. So yeah, I think this is never called on workers as you say. Given that, let's go ahead with the fix as written. (Strictly speaking it probably ought to be a separate change, but I don't mind sneaking it in here.)

synapse/replication/http/login.py
DMRobertson: FWIW I've been trying to leave a comment justifying every `cast` I make. But maybe TypedDict -> Dict is common enough that it's okay.

DMRobertson (approved):
LGTM---many thanks.

(Still marked as draft but I don't think you have further changes planned here?)

PR #11865, opened 2022-01-31
Print warning if synctl_cache_factor is set in config

### LINKED ISSUE #6354: document synctl_cache_factor
we have configs out there using synctl_cache_factor to set SYNAPSE_CACHE_FACTOR but it's not documented anywhere, which makes it look obsolete or abandoned or deprecated. we should doc it so that when people look to see what it does, they can figure it out

lukasdenk:
**_Eventually_** closes #6354 (According to [this](https://github.com/matrix-org/synapse/issues/6354#issuecomment-552811915) comment)

### Warning:
(!) Only **_eventually_** closes #6354.

### Content:

Deprecate using `synctl` with the config option `synctl_cache_factor`.
Print a warning if a user still uses this option. Instead suggest the user to: 
 - Either set the environment variable 'SYNAPSE_CACHE_FACTOR'
 - or set 'caches.global_factor' in the configs.

### Pull Request Checklist


changelog.d/11865.removal
```diff
@@ -0,0 +1,4 @@
+Deprecate using `synctl` with the config option `synctl_cache_factor`.
+Print a warning if a user still uses this option. Instead suggest the user to: 
+ - Either set the environment variable 'SYNAPSE_CACHE_FACTOR'
+ - or set 'caches.global_factor' in the configs.
```
H-Shay, line 4: I don't think lines 2-4 are needed-generally you don't need to add implementation details to a changelog. Perhaps something like "Deprecate using `synctl` with the config option `synctl_cache_factor` and print a warning if a user still uses this option" might suffice?

synctl
```diff
@@ -228,6 +238,7 @@ def main():
     start_stop_synapse = True
 
     if cache_factor:
+        logger.warning(SYNCTL_CACHE_FACTOR_WARNING)
         os.environ["SYNAPSE_CACHE_FACTOR"] = str(cache_factor)
```
H-Shay, line 242: If we are warning and stating that this option is deprecated, should we still set the env variable on the next line?

changelog.d/11865.removal
lukasdenk: OK

synctl
lukasdenk: I think so. Doesn't deprecating a feature just _prepare_ the user that the feature will be removed in the feature? Otherwise it would be a feature removal, not a deprecation. However, the [above mentioned comment](https://github.com/matrix-org/synapse/issues/6354#issuecomment-552811915) demands a deprecation, not a removal.  
But of course, I can also remove it.

synctl
```diff
@@ -37,6 +38,15 @@ YELLOW = "\x1b[1;33m"
 RED = "\x1b[1;31m"
 NORMAL = "\x1b[m"
 
+SYNCTL_CACHE_FACTOR_WARNING = """\
+Setting 'synctl_cache_factor' in the config is deprecated. Instead, please do
+one of the following:
+ - Either set the environment variable 'SYNAPSE_CACHE_FACTOR'
+ - or set 'caches.global_factor' in the configs.
```
anoadragon453, line 45:
```suggestion
 - or set 'caches.global_factor' in the homeserver config.
```

synctl
```diff
@@ -228,6 +238,7 @@ def main():
     start_stop_synapse = True
 
     if cache_factor:
+        logger.warning(SYNCTL_CACHE_FACTOR_WARNING)
```
anoadragon453, line 241: I don't believe `logger` is necessary here - `write(SYNCTL_CACHE_FACTOR_WARNING)` should do the same thing.

anoadragon453 (changes requested):
Looks good to me otherwise. Thanks for doing this!

synctl
lukasdenk: @anoadragon453 Are you sure? I used the logger because it is also used for this warning:

https://github.com/matrix-org/synapse/blob/0cc3bf97b4399234cf20f52ae4c09d03661225ff/synapse/config/key.py#L74-L78

anoadragon453: `synctl` is a separate script from Synapse. It's used to start up Synapse and run it as a daemon. We do use `logging` extensively in Synapse, but not in `synctl` (currently).

I don't know if there's an advantage to using logging over `write()` commands in `synctl`, but if there is we should change them all out - and in a separate PR.

anoadragon453 (changes requested):
Blocked on a resolution to https://github.com/matrix-org/synapse/pull/11865#discussion_r797475535.

synctl
lukasdenk: :+1:

erikjohnston:
(Requesting from the team again as anoa is off atm)

DMRobertson (approved):
Seems good to me. I'd be tempted to write the error to `stderr` rather than `stdout`, but I don't think it makes much difference here.

PR #11866, opened 2022-01-31
Add query parameter `ts` to allow appservices set the `origin_server_ts` for state events

### LINKED ISSUE #10634: Support AS timestamp massaging for state events
[MSC3316](https://github.com/matrix-org/matrix-doc/pull/3316) recommends that `PUT /_matrix/client/r0/rooms/{roomId}/state/{eventType}/{stateKey}` should support a `ts` key for appservices to override the `origin_server_ts` value of an event. This is already supported in Synapse for normal events, as the feature has been in Synapse for a long time (albeit [excluded from the spec](https://matrix.org/docs/spec/application_service/r0.1.2#timestamp-massaging); MSC3316 aims to revive it).

lukasdenk:
Closes #10634.

Allow appservices to set the `origin_server_ts` of a state event by providing the query parameter `ts` in `PUT /_matrix/client/r0/rooms/{roomId}/state/{eventType}/{stateKey}`.

### Pull Request Checklist


synapse/rest/client/room.py
```diff
@@ -204,6 +204,11 @@ async def on_PUT(
         if state_key is not None:
             event_dict["state_key"] = state_key
 
+        # Twisted will have processed the args by now.
+        assert request.args is not None
+        if b"ts" in request.args and requester.app_service:
+            event_dict["origin_server_ts"] = parse_integer(request, "ts", 0)
```
anoadragon453, line 210: I think `None` would make more sense as a default value for `parse_integer`, as that implies the parameter could not be found. Whereas `0` is (technically) a valid value for `origin_server_ts`.

Note with that, you'd no longer need the `b"ts" in request.args` conditional check:

```python
        if requester.app_service:
            event_dict["origin_server_ts"] = parse_integer(request, "ts")
```

synapse/handlers/room_member.py
```diff
@@ -261,6 +261,7 @@ async def _local_membership_update(
         target: UserID,
         room_id: str,
         membership: str,
+        origin_server_ts: Optional[int],
```
anoadragon453, line 264: It looks like there's a missing `= None` here?
```suggestion
        origin_server_ts: Optional[int] = None,
```

synapse/handlers/room_member.py
```diff
@@ -279,6 +280,7 @@ async def _local_membership_update(
             target:
             room_id:
             membership:
+            origin_server_ts:
```
anoadragon453, line 283: Please add documentation where possible. Likewise below.

synapse/rest/client/room.py
anoadragon453, line 210: What if the provided `origin_server_ts` is a negative number?

synapse/rest/client/room.py
```diff
@@ -204,6 +204,11 @@ async def on_PUT(
...
+
```
anoadragon453, line 211: It would be nice to have some tests for this endpoint. Sadly it doesn't look like there are any unit tests for the existing `ts` parameter on `PUT /_matrix/client/r0/rooms/{roomId}/send/{eventType}/{txnId}`. There are a number of tests for both relevant endpoints in general in https://github.com/matrix-org/synapse/blob/a19d01c3d95f5dbd3a4bb181cb70dacd44135a8b/tests/rest/client/test_rooms.py however.

As this is expected to be a feature used across multiple homeserver implementations, it would be good to include tests in https://github.com/matrix-org/complement - the homeserver-agnostic test suite. But if you're not comfortable writing in Golang, then I'd be happy with some Synapse unit tests as well.

anoadragon453 (changes requested):
Thanks for this @lukasdenk! As noted in https://github.com/matrix-org/synapse/issues/10634#issuecomment-1026041324, we'll need to use an unstable prefix (something like `org.matrix.msc3316.ts`) in place of `ts` before the MSC can be merged. That unstable prefix first should be decided on in https://github.com/matrix-org/matrix-doc/pull/3316, then used here.

I'd also like to see this PR introduce the unstable prefix as an acceptable alternative parameter alongside `ts` for `PUT /_matrix/client/r0/rooms/{roomId}/send/{eventType}/{txnId}`. Otherwise we'll end up in a state where one endpoint supports the stable parameter, while the other endpoint only supports the unstable one - which could complicate client code.

Linting isn't passing currently. Note that you can run the linting tools locally (and even have many issues auto-corrected), as described in [the contributing guide](https://matrix-org.github.io/synapse/latest/development/contributing_guide.html#run-the-linters).

synapse/rest/client/room.py
lukasdenk: In the [MSC3316](https://github.com/matrix-org/matrix-doc/pull/3316) this PR is about, the creator mentions that
> Timestamps should already be considered untrusted over federation, and
application services are trusted server components

(see  [here](https://github.com/matrix-org/matrix-doc/blob/555e779f49728b893e37f04ded325457fff07422/proposals/3316-appservice-timestamp-massaging.md?plain=1#L43-L46))

So maybe we can assume that the value is `<0`? Or we could throw an exception? The original lines from which I copied the code also do not handle negative values specifically.

https://github.com/matrix-org/synapse/blob/f0768663eb240f5d95aaad476d890adb4c55e673/synapse/rest/client/room.py#L268-L271

synapse/rest/client/room.py
lukasdenk: Ok, I'll work on it. However, for the complement tests: I don't know Golang, so I'll see how quickly I can get into it.

synapse/rest/client/room.py
lukasdenk: Done :)

synapse/handlers/room_member.py
lukasdenk: Done

anoadragon453:
Heads up - when this is ready for review again just click "request review" next to my avatar to let me know :)

image

The little :repeat: icon.

lukasdenk:
@anoadragon453 Do you have any idea why `trial-olddeps` fails with `tests.rest.client.test_rooms does not exist`?

tests/rest/client/utils.py
```diff
@@ -142,14 +151,25 @@ def create_room_as(
...
-        self.change_membership(
+    def invite(
+        self,
+        room=None,
+        src=None,
+        targ=None,
+        expect_code=200,
+        tok=None,
+        appservice_user_id=None,
+        msc3316_ts=None,
```
richvdh, line 162: rather than having to extend all of these methods with new parameters specific to your test case, I'd recommend inlining them, and calling `send_request` directly. (Or possibly `HomeserverTestCase.send_request`, which sets the `reactor` and `site` parameters for you.)

So, instead of calling `self.helper.invite(room, appservice_user, normal_user, tok=appservice.token, msc3316_ts=ts, appservice_user_id=appservice_user)`, you would do:

```python
        url_params = {
            "user_id": appservice_user,
            "org.matrix.msc3316.ts": ts,
        }
        channel = self.make_request(
            "PUT",
            path=f"/_matrix/client/r0/rooms/{room}/state/m.room.member/{normal_user}?"+url_encode(url_params),
            content={"membership": "invite"},
            access_token=appservice.token,
        )
        self.assertEqual(channel.code, 200)
        event_id = channel.json_body["event_id"]
```

To my mind, this is much clearer than going via a set of utility functions whose effects are hard to understand.

The same applies to `create_room_as`, `send`, etc.

tests/rest/client/test_rooms.py
```diff
@@ -966,6 +970,81 @@ async def user_may_join_room(
...
+            self.room,
+            tok=self.appservice.token,
+            msc3316_ts=self.ts,
+            appservice_user_id=self.appservice_user,
+        )["event_id"]
+        self._check_event_ts(event_id)
+
+    def _check_event_ts(self, event_id):
+        res = self.get_success(self.hs.get_datastore().get_event(event_id))
+        self.assertEquals(self.ts, res.origin_server_ts)
```
richvdh, line 1045: This could do with a comment:

```suggestion
        # check that the event was successfully persisted to the database with the correct timestamp.
        res = self.get_success(self.hs.get_datastore().get_event(event_id))
        self.assertEquals(self.ts, res.origin_server_ts)
```

tests/rest/client/test_rooms.py
```diff
@@ -966,6 +970,81 @@ async def user_may_join_room(
         self.helper.join(self.room3, self.user2, expect_code=403, tok=self.tok2)
 
 
+class RoomAppserviceTsParamTestCase(RoomBase):
```
richvdh, line 973: I think this should inherit directly from `HomeserverTestCase` rather than `RoomBase`? It seems to override everything in `RoomBase`?

tests/rest/client/test_rooms.py
```diff
@@ -966,6 +970,81 @@ async def user_may_join_room(
...
+    servlets = [
+        room.register_servlets,
+        synapse.rest.admin.register_servlets,
+        register.register_servlets,
+    ]
+
+    def prepare(self, reactor, clock, homeserver):
+        self.ts = 1
```
richvdh, line 981: it's a bit odd to make this a property of the test object, rather than a local variable.

tests/rest/client/test_rooms.py
richvdh, line 1045: given that this is only three lines, I'd just inline it for clarity.

richvdh (changes requested):
Hi @lukasdenk, thanks for looking at this and bearing with us!

It looks good, but I've got a few comments.

synapse/rest/client/room.py
```diff
@@ -204,6 +204,11 @@ async def on_PUT(
         if state_key is not None:
             event_dict["state_key"] = state_key
 
+        # Twisted will have processed the args by now.
+        assert request.args is not None
```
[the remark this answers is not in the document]
anoadragon453: Ah, indeed. `args` is not initially a dict, but `None`: https://github.com/twisted/twisted/blob/a033c848d3adc1359d88a9806db8bbd427313841/src/twisted/web/http.py#L796

So the correct type is `Optional[dict/Any]`, in which case the assert here will appease mypy.

lukasdenk:
Ok, I'll work on it probably next week. Writing exams this week :)

richvdh:
good luck!

synapse/rest/client/room.py
```diff
@@ -204,6 +204,13 @@ async def on_PUT(
...
+        if requester.app_service:
+            event_dict["origin_server_ts"] = parse_integer(
+                request, "org.matrix.msc3316.ts"
+            )
```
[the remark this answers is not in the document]
richvdh: > Another option in this case might just be to wait for [MSC3316](https://github.com/matrix-org/matrix-doc/pull/3316) to complete FCP, which means we can just use the stable parameter name (ts) and not have to worry about supporting the unstable prefixes or using experimental feature flags.

now that MSC3316 is in FCP, I strongly recommend this option.

DMRobertson:
> Do you have any idea how that can be? As far as I remember, a month ago, the test worked locally.

Synapse now depends on a newer version of `matrix-common` than you have installed. Run `pip install -U matrix-common` and try running the tests again.

tests/rest/client/utils.py
```diff
@@ -246,6 +267,12 @@ def change_membership(
                 using an application service access token in `tok`.
             expect_code: The expected HTTP response code
             expect_errcode: The expected Matrix error code
+            msc3316_ts: If the user is an appservice, msc3316_ts will be used as the event's
+                "original_server_ts"
```
MadLittleMods, line 271:
```suggestion
                "origin_server_ts"
```

anoadragon453:
Hi @lukasdenk. This PR looks like it was pretty close to being completed. Do you have any time to get it over the line?

The first step that I would suggest doing is to pull in the latest changes from the `develop` branch with:

```
git pull origin develop
```

and then to resolve the git conflicts.

tulir:
All the unstable prefixes can also be removed now since the MSC is stable and fully released as of Matrix v1.3

lukasdenk:
> Hi @lukasdenk. This PR looks like it was pretty close to being completed. Do you have any time to get it over the line?
> 
> The first step that I would suggest doing is to pull in the latest changes from the `develop` branch with:
> 
> ```
> git pull origin develop
> ```
> 
> and then to resolve the git conflicts.

It is quite close, I know. But I don't think I have the time to finish it. Sorry.

clokep:
@lukasdenk Hope you don't mind, but I merged in `develop` and handled some of the review comments. Hopefully we can get this over the line! Seems like you did most of the work already! 🎉

tests/rest/client/utils.py
clokep: I ended up rolling these all back *except* `create_room_as`, it should be easy enough to call that manually though, if you prefer?

synapse/rest/client/room.py
```diff
@@ -335,8 +343,13 @@ async def on_POST(
...
         assert request.args is not None
-        if b"ts" in request.args and requester.app_service:
-            event_dict["origin_server_ts"] = parse_integer(request, "ts", 0)
+        if requester.app_service:
+            if b"ts" in request.args:
+                event_dict["origin_server_ts"] = parse_integer(request, "ts")
+            elif b"org.matrix.msc3316.ts" in request.args:
+                event_dict["origin_server_ts"] = parse_integer(
+                    request, "org.matrix.msc3316.ts"
+                )
```
tulir, line 352: Looks like this one was missed while removing the unstable prefixes

tests/rest/client/utils.py
```diff
@@ -116,6 +119,10 @@ def create_room_as(
                 Note that if is_public is set, the "visibility" key will be overridden.
                 If room_version is set, the "room_version" key will be overridden.
             custom_headers: HTTP headers to include in the request.
+            appservice_user_id: The `user_id` URL parameter to pass.
+                This allows driving an application service user
+                using an application service access token in `tok`.
+
```
richvdh, line 125: as a sidebar: I kinda hate these `do_foo_as` methods which mess about with the auth mechanism: I find them harder to use and understand than just doing the request. But I'll not insist on a rewrite here.

richvdh (approved):
lgtm!

tests/rest/client/utils.py
clokep: I backed out these changes. 👍

PR #11868, opened 2022-01-31
Allow modules to retrieve server and worker names

babolivier:
Fixes https://github.com/matrix-org/synapse/issues/10701

### DIFF
--- a/synapse/module_api/__init__.py
+++ b/synapse/module_api/__init__.py
@@ -401,6 +401,23 @@ class ModuleApi:
         """
         return self._hs.config.email.email_app_name
 
+    @property
+    def server_name(self) -> str:
+        """The server name for the local homeserver.
+
+        Added in Synapse v1.53.0.
+        """
+        return self._server_name
+
+    @property
+    def worker_name(self) -> Optional[str]:
+        """The name of the worker this specific instance is running as per the
+        "worker_app" configuration setting, or None if it's the main process.
+
+        Added in Synapse v1.53.0.
+        """
+        return self._hs.config.worker.worker_app
+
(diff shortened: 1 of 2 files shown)


synapse/module_api/__init__.py
```diff
@@ -401,6 +401,23 @@ def email_app_name(self) -> str:
         """
         return self._hs.config.email.email_app_name
 
+    @property
+    def server_name(self) -> str:
+        """The server name for the local homeserver.
+
+        Added in Synapse v1.53.0.
```
babolivier, line 408: Self-reminder to change this (and the one below) if it gets through quickly enough for 1.52 (unlikely since the RC cut-off is tomorrow)

synapse/module_api/__init__.py
```diff
@@ -401,6 +401,23 @@ def email_app_name(self) -> str:
...
+        return self._server_name
+
+    @property
+    def worker_name(self) -> Optional[str]:
+        """The name of the worker this specific instance is running as per the
+        "worker_app" configuration setting, or None if it's the main process.
+
+        Added in Synapse v1.53.0.
+        """
+        return self._hs.config.worker.worker_app
```
dklimpel, line 419: I think it's confusing.
https://github.com/matrix-org/synapse/blob/0da2301b21399b85a955d92dc355024a67b8cb40/docs/workers.md#L141-L142

Worker knows `worker_app` and `worker_name`. These are different parameters. And here is it a mix of both and a switch from `worker_name` and `worker_app`. Shouldn't the names stay the same?
Maybe it makes sense to make both parameters available to the module.

babolivier: Oh good point, thanks for catching. I meant to expose `worker_name`. But you're probably right that it's best to expose both.

PR #11869, opened 2022-01-31
Fixes for opentracing scopes

richvdh:
`start_active_span` was inconsistent as to whether it would activate the span immediately, or wait for `scope.__enter__` to happen (it depended on whether the current logcontext already had an associated scope). The inconsistency was rather confusing if you were hoping to set up a couple of separate spans before activating either.

Looking at the other implementations of opentracing `ScopeManager`s, the intention is that it *should* be activated immediately, as the name implies. Indeed, the idea is that you don't have to use the scope as a contextmanager at all - you can just call `.close` on the result. Hence, our cleanup has to happen in `.close` rather than `.__exit__`.

So, the main change here is to ensure that `start_active_span` does activate the span, and that `scope.close()` does close the scope.

We also add some tests, which requires a `tracer` param so that we don't have to rely on the global variable in unit tests.

### DIFF
--- a/synapse/logging/opentracing.py
+++ b/synapse/logging/opentracing.py
@@ -489,7 +489,6 @@ def start_active_span_follows_from(
         contexts: the previous spans to inherit from
         inherit_force_tracing: if set, and any of the previous contexts have had tracing
            forced, the new span will also have tracing forced.
-
         tracer: override the opentracing tracer. By default the global tracer is used.
     """
     if opentracing is None:
--- a/synapse/logging/opentracing.py
+++ b/synapse/logging/opentracing.py
@@ -443,10 +443,14 @@ def start_active_span(
     start_time=None,
     ignore_active_span=False,
     finish_on_close=True,
+    *,
+    tracer=None,
 ):
-    """Starts an active opentracing span. Note, the scope doesn't become active
-    until it has been entered, however, the span starts from the time this
-    message is called.
+    """Starts an active opentracing span.
+
+    Records the start time for the span, and sets it as the "active span" in the
+    scope manager.
+
     Args:
         See opentracing.tracer
     Returns:
@@ -456,7 +460,11 @@ def start_active_span(
     if opentracing is None:
         return noop_context_manager()  # type: ignore[unreachable]
 
-    return opentracing.tracer.start_active_span(
+    if tracer is None:
+        # use the global tracer by default
+        tracer = opentracing.tracer
+
+    return tracer.start_active_span(
         operation_name,
         child_of=child_of,
         references=references,
@@ -468,7 +476,11 @@ def start_active_span(
 
 
 def start_active_span_follows_from(
-    operation_name: str, contexts: Collection, inherit_force_tracing=False
+    operation_name: str,
+    contexts: Collection,
+    *,
+    inherit_force_tracing=False,
+    tracer=None,
 ):
     """Starts an active opentracing span, with additional references to previous spans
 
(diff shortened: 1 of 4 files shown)


synapse/logging/opentracing.py
```diff
@@ -443,10 +443,14 @@ def start_active_span(
     start_time=None,
     ignore_active_span=False,
     finish_on_close=True,
+    *,
```
richvdh, line 446: since this isn't a standard parameter for the underlying `opentracing.tracer.start_active_span`, I wanted to make it a kw-only param so that we can safely add more positional params to pass through in future.

synapse/logging/opentracing.py
```diff
@@ -477,12 +489,18 @@ def start_active_span_follows_from(
         contexts: the previous spans to inherit from
         inherit_force_tracing: if set, and any of the previous contexts have had tracing
            forced, the new span will also have tracing forced.
+
+        tracer: override the opentracing tracer. By default the global tracer is used.
```
babolivier, line 493:
```suggestion
        tracer: override the opentracing tracer. By default the global tracer is used.
```

synapse/logging/scopecontextmanager.py
```diff
@@ -101,8 +118,7 @@ def __init__(self, manager, span, logcontext, enter_logcontext, finish_on_close)
             logcontext (LogContext):
                 the logcontext to which this scope is attached.
             enter_logcontext (Boolean):
-                if True the logcontext will be entered and exited when the scope
-                is entered and exited respectively
+                if True the logcontext will exited when the scope is finished
```
babolivier, line 121:
```suggestion
                if True the logcontext will be exited when the scope is finished
```

tests/logging/test_opentracing.py
```diff
@@ -0,0 +1,185 @@
...
+
+        # the spans should be reported in order of their finishing.
+        self.assertEqual(
+            self._reporter.get_spans(), [scope2.span, scope1.span, root_scope.span]
+        )
+
+    def test_overlapping_spans(self) -> None:
+        """Overlapping spans which are not neatly nested should work
+
+        ... provided they
```
babolivier, line 132: Provided they...?

tests/logging/test_opentracing.py
```diff
@@ -0,0 +1,185 @@
...
+                )
+                self.assertEqual(self._tracer.active_span, scope2.span)
+                self.assertEqual(
+                    scope2.span.context.parent_id, scope1.span.context.span_id
+                )
+
+                with scope1, scope2:
+                    pass
+
+                # the root scope should be restored
```
babolivier, line 117: I'm having troubles understanding this bit - doesn't the PR description say that closing a scope should happen in `.close()` rather than `.__exit__()`? Aren't we closing `scope1` and `scope2` here?

tests/logging/test_opentracing.py
richvdh: :man_shrugging:

tests/logging/test_opentracing.py
richvdh: *both* `__exit__()` *and* `.close()` should close the scope (which happens by virtue of `Scope.__exit__()` calling `self.close()`.

tests/logging/test_opentracing.py
```diff
@@ -0,0 +1,185 @@
...
+        # the spans should be reported in order of their finishing.
+        self.assertEqual(
+            self._reporter.get_spans(), [scope2.span, scope1.span, root_scope.span]
+        )
+
+    def test_overlapping_spans(self) -> None:
+        """Overlapping spans which are not neatly nested should work
+
+        ... provided they
+        """
```
richvdh, line 133:
```suggestion
        """Overlapping spans which are not neatly nested should work"""
```

tests/logging/test_opentracing.py
babolivier: > which happens by virtue of `Scope.__exit__()` calling `self.close()`

Oh, I see, that's the bit of information I was missing, thanks.

babolivier (approved):

PR #11870, opened 2022-01-31
Improve opentracing for federation requests

richvdh:
The idea here is to set the parent span for incoming federation requests to the *outgoing* span on the other end. That means that you can see (most of) the full end-to-end flow when you have a process that includes federation requests.

However, in order not to lose information, we still want a link to the `incoming-federation-request` span from the servlet, so we have to create another span to do exactly that.


erikjohnston:
Sorry, I'm a bit confused how all the span mentioned currently tie together (or not) and how you want them to tie together? I'm just a bit surprised its not simply `outgoing-request -> incoming-federation-request -> servlet` or something like that?

richvdh:
(maybe a better approach here is to rethink our whitelisting - do it by IP address, for example, so that we don't need to wait until we've authenticated the remote server)

erikjohnston (approved):
Thanks for walking me through! I think technically one should be able to change the parent of an active span, but I don't think jaeger exposes a way of doing that sadly

PR #11878, opened 2022-02-01
Only clear the get_users_in_room caches when necessary.

### LINKED ISSUE #11875: Overly aggressive purging of get_users_in_room cache?
We invalidate the `get_users_in_room` cache (and a few others) as part of [`_invalidate_state_caches`](https://github.com/matrix-org/synapse/blob/5305a5e88144828419249fd9e4c5198d92276a44/synapse/storage/_base.py#L59-L75), we [calculate a `members_changed`](https://github.com/matrix-org/synapse/blob/2aa37a4250675f6d9feb57ec0dce65b2a6a3cde6/synapse/storage/databases/main/events.py#L1105-L1120) with the comment:

> Figure out the changes of membership to invalidate the `get_rooms_for_user` cache.

But don't seem to be using it in the opposite direction (`get_users_for_room`).

tl;dr I think we are invalidating this cache in situations we do not need to.

clokep:
We seem to clear the `get_users_in_room` and `get_users_in_room_with_profiles` caches unnecessarily -- when any room state changes, not just when memberships change.

I'm not sure if this is a problem in real-life usage, but I don't see any reason to do this.

Fixes #11875

### DIFF
--- a/changelog.d/11878.misc
+++ b/changelog.d/11878.misc
@@ -1,0 +1,1 @@
+Do not needlessly clear the `get_users_in_room` and `get_users_in_room_with_profiles` caches when any room state changes.
--- a/synapse/storage/_base.py
+++ b/synapse/storage/_base.py
@@ -57,7 +57,7 @@ class SQLBaseStore(metaclass=ABCMeta):
         pass
 
     def _invalidate_state_caches(
-        self, room_id: str, members_changed: Iterable[str]
+        self, room_id: str, members_changed: Collection[str]
     ) -> None:
         """Invalidates caches that are based on the current state, but does
         not stream invalidations down replication.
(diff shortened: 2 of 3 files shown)


clokep:
Mostly wanting to see a CI run of this change.

DMRobertson (approved):
Nice spot! I would guess that most state changes in most rooms _are_ membership events, so don't think we'd expect this to have a huge impact? Nice to have though!

clokep:
> Nice spot! I would guess that most state changes in most rooms _are_ membership events, so don't think we'd expect this to have a huge impact? Nice to have though!

That would be my guess, yes! 👍 

A downside of this is that we might still clear the cache multiple times in a row for any systems being replicated to. I didn't think that was a big enough deal to be worth solving / is somewhat orthogonal to this.

DMRobertson:
> A downside of this is that we might still clear the cache multiple times in a row for any systems being replicated to.

Note sure what you're referring to here---could you expand?

clokep:
> > A downside of this is that we might still clear the cache multiple times in a row for any systems being replicated to.
> 
> Note sure what you're referring to here---could you expand?

`_invalidate_state_caches_and_stream` will invalidate the local cache (via `_invalidate_state_caches`) with the entire membership list, but it then streams the cache change with the membership changes in chunks of 50 (to limit the message size being sent).

The receiving system then calls `_invalidate_state_caches` via `CacheInvalidationWorkerStore.process_replication_rows` with each chunk of members. So if you have 51 members it would come across replication as two different rows, causing the cache to be cleared twice.

PR #11880, opened 2022-02-01
Convert `ApplicationServiceTestCase` to use `simple_async_mock`

anoadragon453:
This ended up being a requirement for testing sending device lists to application services (https://github.com/matrix-org/synapse/pull/11881). I've split it out into a separate PR for ease of review.

### DIFF
--- a/changelog.d/11880.misc
+++ b/changelog.d/11880.misc
@@ -1,0 +1,1 @@
+Convert `ApplicationServiceTestCase` to use `simple_async_mock`.
--- a/tests/appservice/test_appservice.py
+++ b/tests/appservice/test_appservice.py
@@ -164,9 +164,7 @@ class ApplicationServiceTestCase(unittest.TestCase):
         )
         self.service.namespaces[ApplicationService.NS_USERS].append(_regex("@irc_.*"))
         self.event.sender = "@irc_foobar:matrix.org"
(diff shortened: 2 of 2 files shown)


clokep (approved):
Seems fine if CI passes.

clokep:
Looks like this needs black run on it. 👍

DMRobertson:
Meta: isn't `simple_async_mock` redundant now we're only supporting 3.7+? (I.e. doesn't `patch.object` now detect that we're patching an async function and internally create an `AsyncMock` for us?)

clokep:
> Meta: isn't `simple_async_mock` redundant now we're only supporting 3.7+? (I.e. doesn't `patch.object` now detect that we're patching an async function and internally create an `AsyncMock` for us?)

We somehow got ourselves convinced of this, but `AsyncMock` doesn't actually exist until 3.8. 😢

DMRobertson:
> > Meta: isn't `simple_async_mock` redundant now we're only supporting 3.7+? (I.e. doesn't `patch.object` now detect that we're patching an async function and internally create an `AsyncMock` for us?)
> 
> We somehow got ourselves convinced of this, but `AsyncMock` doesn't actually exist until 3.8. cry

Argh, I think I've made this mistake before! Thanks for correcting me.

PR #11881, opened 2022-02-01
Send device list updates to application services (MSC3202) - part 1

anoadragon453:
This PR sends device lists to application services as specified by the current copy of [MSC3202](https://github.com/matrix-org/matrix-doc/pull/3202). This feature is additionally guarded by the existing `msc3202_transaction_extensions` experimental feature flag. 

In order to help break up the work, this PR only implements the "changed" field of device list updates ("left" is always empty). Application Services will be able to work with this, though it could over time lead to UISIs. Implementing the "left" field is the intention for part 2 of this work.

Based on https://github.com/matrix-org/synapse/pull/12193, though it shouldn't block review. Strongly recommended to review commit-by-commit, starting from https://github.com/matrix-org/synapse/pull/11881/commits/51be04b918bbbaa64e8f8aab7c1394e1db6401d8.


synapse/handlers/appservice.py
```diff
@@ -253,15 +257,37 @@ def notify_interested_services_ephemeral(
...
+                    "typing_key",
+                    "receipt_key",
+                    "presence_key",
+                    "to_device_key",
+                )
+                and service.supports_ephemeral
+            )
+            or (
+                stream_key == "device_list_key"
+                and service.msc3202_transaction_extensions
```
anoadragon453, line 289: Note that this is a different variable from `self._msc3202_transaction_extensions_enabled` above!

synapse/appservice/scheduler.py
```diff
@@ -212,7 +227,40 @@ async def _send_request(self, service: ApplicationService) -> None:
...
+                    summary = self.queued_device_list_summaries[service.id].pop(0)
+
+                    # For every user in the incoming "changed" set:
+                    #   * Remove them from the existing "left" set if necessary
+                    #     (as we need to start tracking them again)
+                    #   * Add them to the existing "changed" set if necessary.
+                    for user_id in summary.changed:
+                        if user_id in device_list_summary.left:
+                            device_list_summary.left.remove(user_id)
+                        device_list_summary.changed.add(user_id)
```
clokep, line 246: I think what you want can easily be done with set math?

```suggestion
                    device_list_summary.left -= summary.changed
                    device_list_summary.changed.update(summary.changed)
```

synapse/appservice/scheduler.py
```diff
@@ -212,7 +227,40 @@ async def _send_request(self, service: ApplicationService) -> None:
...
+
+                    # For every user in the incoming "left" set:
+                    #   * Remove them from the existing "changed" set if necessary
+                    #     (we no longer need to track them)
+                    #   * Add them to the existing "left" set if necessary.
+                    for user_id in summary.left:
+                        if user_id in device_list_summary.changed:
+                            device_list_summary.changed.remove(user_id)
+                        device_list_summary.left.add(user_id)
```
clokep, line 255:
```suggestion
                    device_list_summary.changed -= summary.left
                    device_list_summary.left.update(summary.left)
```

synapse/appservice/scheduler.py
```diff
@@ -212,7 +227,40 @@ async def _send_request(self, service: ApplicationService) -> None:
                 ]
                 del all_to_device_messages[:MAX_TO_DEVICE_MESSAGES_PER_TRANSACTION]
 
-                if not events and not ephemeral and not to_device_messages_to_send:
+                # Consolidate any pending device list summaries into a single, up-to-date
+                # summary.
+                # Note: this code assumes that in a single DeviceLists, a user will
+                # never be in both "changed" and "left" sets.
```
clokep, line 233: I think the code below will actually work fine if the sets are not disjoint, with `left` winning...which might not be what you want. 👍

synapse/appservice/scheduler.py
```diff
@@ -212,7 +227,40 @@ async def _send_request(self, service: ApplicationService) -> None:
...
+                    for user_id in summary.left:
+                        if user_id in device_list_summary.changed:
+                            device_list_summary.changed.remove(user_id)
+                        device_list_summary.left.add(user_id)
+
+                if (
+                    not events
+                    and not ephemeral
+                    and not to_device_messages_to_send
+                    # Note that DeviceLists implements __bool__
```
clokep, line 261: I'm not sure this is helpful?

synapse/storage/databases/main/appservice.py
```diff
@@ -430,7 +434,7 @@ def get_new_events_for_appservice_txn(txn):
     async def get_type_stream_id_for_appservice(
         self, service: ApplicationService, type: str
     ) -> int:
-        if type not in ("read_receipt", "presence", "to_device"):
+        if type not in ("read_receipt", "presence", "to_device", "device_list"):
```
clokep, line 437: Should we have a constant somewhere with the known stream types?

synapse/appservice/__init__.py
```diff
@@ -400,6 +401,7 @@ def __init__(
         to_device_messages: List[JsonDict],
         one_time_key_counts: TransactionOneTimeKeyCounts,
         unused_fallback_keys: TransactionUnusedFallbackKeys,
+        device_list_summary: DeviceLists,
```
clokep, line 404: (I chose a semi-random spot for this comment)

I'm surprised how much of the code needs to have the new `device_list_summary` threaded though it -- should we be creating an `AppServiceTransaction` (or some other object) high up in the stack and passing that along instead?

If so, could probably be done as a follow-up...especially if we're going to be adding more data to transactions.

clokep (commented):
Looks pretty reasonable! I left a bunch of comments / questions, but I think none of them are major changes / reworks.

synapse/appservice/scheduler.py
```diff
@@ -212,7 +227,36 @@ async def _send_request(self, service: ApplicationService) -> None:
...
+                while self.queued_device_list_summaries.get(service.id, []):
+                    # Pop a summary off the front of the queue
+                    summary = self.queued_device_list_summaries[service.id].pop(0)
+
+                    # For every user in the incoming "changed" set:
+                    #   * Remove them from the existing "left" set if necessary
+                    #     (as we need to start tracking them again)
+                    #   * Add them to the existing "changed" set if necessary.
+                    device_list_summary.left -= summary.changed
+                    device_list_summary.changed.update(summary.changed)
```
anoadragon453, line 244: Oh, annoying mypy complains about this suggested change:

```
synapse/appservice/scheduler.py:243: error: Property "left" defined in "DeviceLists" is read-only  [misc]
synapse/appservice/scheduler.py:250: error: Property "changed" defined in "DeviceLists" is read-only  [misc]
```

synapse/handlers/appservice.py
```diff
@@ -542,6 +582,98 @@ async def _get_to_device_messages(
...
+            True if the application service is interested in the user's device lists, False
+            otherwise.
+        """
+        # This method checks against both the sender localpart user as well as if the
+        # user is in the appservice's user namespace.
+        if appservice.is_interested_in_user(user_id):
+            return True
+
+        # FIXME: This is quite an expensive check. This method is called per device
+        #  list change.
```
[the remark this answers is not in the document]
anoadragon453: Overlooking the methods again, both `get_rooms_for_user` and `is_interested_in_room`, so perhaps this actually isn't that bad of a loop, assuming that the caches are warm and appropriately sized (`is_interested_in_room` might be a bit small at 1000 items, 500 with the default 0.5 global cache factor).

I'm not sure of a better way to perform this check however -- a lot of checking just needs to happen here. Plus this comment isn't really helpful, as it doesn't lay out how to solve the problem.

I'm going to just replace it with one that explains the purpose of the loop.

synapse/types.py
```diff
@@ -743,6 +744,26 @@ class ReadReceipt:
     data: JsonDict
 
 
+@attr.s(slots=True, frozen=True, auto_attribs=True)
+class DeviceLists:
```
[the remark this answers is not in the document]
anoadragon453: Ah, well the `Lists` in `DeviceLists` is referring to a user's device list, rather than the lists in the Python object. Though I see how that could be confusing?

`DeviceListUpdates` sounds the best to me? Paired with an actual docstring? An instance is an object containing two sets of user IDs which indicate changes to users' device lists.

synapse/types.py
```diff
@@ -743,6 +744,26 @@ class ReadReceipt:
...
+    Attributes:
+        changed: user_ids whose devices may have changed
+        left: user_ids whose devices we no longer track
+    """
+
+    # We need to use a factory here, otherwise `set` is not evaluated at
+    # object instantiation, but instead at class definition instantiation.
+    # The latter happening only once, thus always giving you the same sets
+    # across multiple DeviceLists instances.
+    # Also see: don't define mutable default arguments.
```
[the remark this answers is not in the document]
anoadragon453: We do, though why we do it isn't documented anywhere else, and I did initially struggle without using a factory.

It's not only relevant to this bit of code, though lacking anywhere else good to put it, leaving it may still help someone?

tests/handlers/test_appservice.py
```diff
@@ -627,6 +638,115 @@ def _register_application_service(
...
+    # Test across a variety of configuration values
+    @parameterized.expand(
+        [
+            (True, True, True),
+            (True, False, False),
+            (False, True, False),
+            (False, False, False),
+        ]
+    )
+    @unittest.override_config({"experimental_features": {"": False}})
```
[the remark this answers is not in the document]
anoadragon453: Oops, that's the result of a cut and paste I think. The config override is no longer necessary anyways, as we now set it by modifying `self.as_handler._msc3202_transaction_extensions_enabled`.

synapse/appservice/scheduler.py
```diff
@@ -212,7 +227,40 @@ async def _send_request(self, service: ApplicationService) -> None:
                 ]
                 del all_to_device_messages[:MAX_TO_DEVICE_MESSAGES_PER_TRANSACTION]
 
-                if not events and not ephemeral and not to_device_messages_to_send:
+                # Consolidate any pending device list summaries into a single, up-to-date
+                # summary.
+                # Note: this code assumes that in a single DeviceLists, a user will
+                # never be in both "changed" and "left" sets.
+                device_list_summary = DeviceLists()
+                while self.queued_device_list_summaries.get(service.id, []):
```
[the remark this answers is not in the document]
anoadragon453: The reason to `pop` items out of `self.queued_device_list_summaries` is so that we end up with an empty list for `self.queued_device_list_summaries[service.id]`.

But do you mean that this would be simpler by changing to a `for` loop, and emptying `self.queued_device_list_summaries.get[service.id]` once we're done?

synapse/appservice/scheduler.py
anoadragon453: The spec [doesn't particularly say](https://spec.matrix.org/v1.2/client-server-api/#extensions-to-sync-1) what a client should do if a user ends up in both lists, though I would assume that `left` should win (as if a user updated their list, but they've now left all e2ee rooms with you, then why do you care?).

...So maybe it's OK to pass a device list update with a user with both lists anyways?

anoadragon453:
(develop was merged in in order to fix failing lint CI)

synapse/types.py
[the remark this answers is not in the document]
clokep: I found this to be a fundamental property of using `attr`, but if you find it useful than that's fine with me.

synapse/appservice/scheduler.py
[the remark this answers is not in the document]
clokep: Yeah, I was thinking more of "loop over it, then clear it" instead of mutating it each time. Not a huge deal, just a bit odd I think?

synapse/appservice/scheduler.py
clokep: Yeah, that makes more sense!

clokep:
@anoadragon453 This needs a merge for `develop` it seems!

clokep (approved):
LGTM!

PR #11882, opened 2022-02-01
Printf debugging for MSISDN validation

DMRobertson:
I propose committing this, running it on m.org's master for a bit, then reverting once we're sufficiently confident in the results.


synapse/rest/client/account.py
```diff
@@ -468,6 +468,12 @@ async def on_POST(self, request: SynapseRequest) -> Tuple[int, JsonDict]:
 
         msisdn = phone_number_to_msisdn(country, phone_number)
 
+        # Didn't like the sound of logging `client_secret`, but the spec says it is
+        # "A unique string generated by the client, and used to identify the validation
+        # attempt." I.e. something to facilitate deduplication. I don't think it's a
+        # sensitive secret per se.
```
babolivier, line 474: It's not so much used to deduplicate but rather to make sure the next request(s) come from the client that started the process. I'm not sure what amount of harm can be done if someone gets a hold of someone else's client secret, but my paranoid self would prefer being cautious and not logging it if we don't need it.

synapse/rest/client/account.py
```diff
@@ -518,6 +525,7 @@ async def on_POST(self, request: SynapseRequest) -> Tuple[int, JsonDict]:
         threepid_send_requests.labels(type="msisdn", reason="add_threepid").observe(
             send_attempt
         )
+        logger.info("MSISDN %s is already in use by %s", msisdn, existing_user_id)
```
babolivier, line 528: This doesn't look like it's in the right place?

babolivier (changes requested):
Also CI is failing

synapse/rest/client/account.py
DMRobertson: Thanks, I didn't grok that it was to be sent in another request.

synapse/rest/client/account.py
DMRobertson: no, looks like a copy paste fail. thanks.

babolivier:
I'm assuming the CI failing might fixed by https://github.com/matrix-org/synapse/pull/11906

babolivier (approved):
CI failing isn't related to this PR, let's not block it on that.

PR #11017, opened 2021-10-07
Fix CI to run the unit tests without optional deps

erikjohnston:
This also turns off calculating code coverage, as we didn't use it and it was a lot of noise

PR #11888, opened 2022-02-02
Disable coverage calculation for olddeps build.

richvdh:
We disabled coverage calculation for most of CI in #11017, but the olddeps
build uses a separate script and got forgotten.


DMRobertson (approved):
Sorry for missing this. Can't see any matches for `tox.*combine` other than the one removed here. Is it worth removing the combine testenv from tox.ini?

richvdh:
> Is it worth removing the combine testenv from tox.ini?

<shrug> In theory, it still works and might be useful to someone?

PR #11892, opened 2022-02-02
Add a ratelimiter for 3pid invite

babolivier:
This is part of the Tchap mainlining work. This introduces a ratelimiter specific to sending 3PID invites, which is used instead of the `RequestRatelimiter`. Introduced in synapse-dinsic in https://github.com/matrix-org/synapse-dinsic/pull/11 (though will need to be re-introduced as it got removed by a merge).


clokep (commented):
Looks pretty straightforward!

clokep (approved):
LGTM!

Issue #11891, opened 2022-02-02
Review differences between `matrix-org-hotfixes` and release branch


DMRobertson:
CI currently fails with a mypy error on the hotfixes branch. This made me wonder how different hotfixes are to our release candidates.

We should review these changes. At the very least, I would like to have a reference for the differences between hotfixes and the release branch.

```
$ git diff --stat matrix-org-hotfixes..v1.52.0rc1 
   changelog.d/11583.bugfix                     |  1 -
   synapse/handlers/message.py                  |  2 +-
   synapse/handlers/room_member.py              | 56 ++++++++++++++++++--------------------------------------
   synapse/push/httppusher.py                   |  5 -----
   synapse/storage/database.py                  | 40 +---------------------------------------
   synapse/storage/databases/main/client_ips.py |  2 +-
   synapse/storage/databases/main/search.py     |  2 +-
```

richvdh:
dup https://github.com/matrix-org/synapse/issues/4826 I think...

(the diff is pretty small nowadays, and I would love for us to finish cleaning that up)

DMRobertson:
Ahh yes. Sorry, I should search before I file.

Issue #4826, opened 2019-03-07
Bring matrix-org-hotfixes changes into mainline


erikjohnston:
There aren't many changes, so we may as well move the rest into mainline (behind config flags or whatever)

ShadowJonathan:
I got a diff [here](https://github.com/matrix-org/synapse/compare/2b7dd21655b1ed2db490853d2cdbf6fb38704d81...edac710bc0c4dc1cd226d9ffe73a00b42c2b67d8#files_bucket) for the current (22-4-2021) changes in `-hotfixes` that aren't in develop.

DMRobertson:
State of play as I see it today:

- 2017/05/23: [e5537cf98333b428dbc481ed443daed2f0cfa074](https://github.com/matrix-org/synapse/commit/e5537cf98333b428dbc481ed443daed2f0cfa074) and introduced a `member_limiter` on the
  `RoomMemberHandler`, later tweaked by subsequent commits. This wraps the call to `update_membership_locked` and there looks to be timing-related changes.
- 2017/10/12: [fae708c0e8c35930f1172322b7c0e9f0b1b3f9a4](https://github.com/matrix-org/synapse/commit/fae708c0e8c35930f1172322b7c0e9f0b1b3f9a4) from [matrix-appservice-irc#506](https://github.com/matrix-org/matrix-appservice-irc/issues/506) suppresses the requirement that application services can only lookup joined members if they control a user which is already in this room.
- 2018/02/14: [66dcbf47a36b5ca0e88d4658578d6fb5e6dbd910](https://github.com/matrix-org/synapse/commit/66dcbf47a36b5ca0e88d4658578d6fb5e6dbd910) disables auto search for prefixes in event search.
- 2018/02/14: [8f8ea91eefcc43c5ac24e85b14a86af4da53e6e0](https://github.com/matrix-org/synapse/commit/8f8ea91eefcc43c5ac24e85b14a86af4da53e6e0) bumps client_ip LAST_SEEN_GRANULARITY from 2 to 10 minutes.
- 2018/06/04: [9e38981ae47d03467a954c3c540c51b567f6e50b](https://github.com/matrix-org/synapse/commit/9e38981ae47d03467a954c3c540c51b567f6e50b) and subsequent commits send http pushes directly to a local address, rather than via cloudflare.
[...]

PR #11894, opened 2022-02-02
Require a description of changes on hotfix branch

DMRobertson:
I noted in #11891 (c.f. https://github.com/matrix-org/synapse/issues/4826) that we don't really have a good idea of how the hotfixes branch differs to the releases branch. I propose that we require PRs against the hotfixes branch to include a changelog, and that we enforce this via a github action.

If we commit directly to the branch then this check is avoided. But that's not unique to this specific workflow.

Note that the Hotfixes workflow failed in 18f87a4 but suceeded in de277d8. The checks still fail overall in both cases because linting fails on `matrix-org-hotfixes`.

### DIFF
--- a/.github/workflows/hotfixes-changelog.yml
+++ b/.github/workflows/hotfixes-changelog.yml
@@ -1,0 +1,23 @@
+name: Hotfixes
+on:
+  pull_request:
+    branches:
+      - 'matrix-org-hotfixes'
+jobs:
+  check-changelog:
+    name: Check change is described
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - run: git fetch origin ${{ github.base_ref }}
+
+      - name: Check CHANGES_HOTFIX.md has changed
+        run: |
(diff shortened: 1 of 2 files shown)


richvdh:
I'm unconvinced. Changes to `matrix-org-hotfixes` often skip the PR phase, not least because they often happen in a hurry - which means the check will get skipped. I'd be hesitant to add obstacles to putting our stuff in git - better it goes in git with sketchy information than doesn't go in git at all.

 A file such as you propose will be prone to getting out of date, and doesn't offer a *whole* lot more information than git logs do.

Rather, we should get rid of the differences (#4824), and then if we ever create a difference again, open a GH issue to track it and its removal.

DMRobertson:
> Rather, we should get rid of the differences (#4824), and then if we ever create a difference again, open a GH issue to track it and its removal.

My anxiety is that "we should create an issue" isn't enforced or automated, and so things will fall through the cracks. But I concede the suggestion here is pretty nuclear, and that

> A file such as you propose [...] doesn't offer a whole lot more information than git logs do.

is pretty persuasive.

PR #11895, opened 2022-02-02
Remove support for the webclient listener.

### LINKED ISSUE #1250: The static webclient content served by synapse has no cache control headers (SYN-174)
We should tell clients to cache the HTML & JS served up by Synapse, unless we're actually running in a debug environment.  Unless we think that a LB should be responsible for this?

(Imported from https://matrix.org/jira/browse/SYN-174)

(Reported by @ara4n)

clokep:
Per the plan in #9078 this removes support for the `webclient` listener and requires `web_client_redirect` to be a HTTP(S) URL.

Note that the deprecation landed in Synapse v1.51.0 (#11774) and Synapse v1.52.0rc1 has been cut so `develop` will be come Synapse v1.53.0 eventually.

As a result of this:

* Adding a `webclient` listeners is now a configuration error.
* `web_client_redirect` to a non-HTTP(S) location is now a configuration error.
* `/` will still redirect if `web_client_redirect` is configured.
* `/_matrix/client` will no longer redirect.

Fixes  #1250.


changelog.d/11895.removal
```diff
@@ -0,0 +1 @@
+Drop support for `webclient` listeners, they are now ignored. The `web_client_location` configuration must point to a HTTP(S) URL, if configured.
```
babolivier, line 1: > they are now ignored

It looks like they're now raising errors rather than being ignored?

babolivier (approved):
lgtm otherwise, thanks for taking care of this!

changelog.d/11895.removal
clokep: Good catch, thank you!

PR #11900, opened 2022-02-03
Remove excess condition on `knock->leave` check

turt2live:
This appears to be a bug in that `RoomVersion` is the class object while `room_version` is the local variable. 

Introduced by https://github.com/matrix-org/synapse/pull/6739

### Pull Request Checklist


changelog.d/11900.bugfix
```diff
@@ -0,0 +1 @@
+Fix a bug introduced in Synapse 1.37.0rc1 (when knocking was added) where event authorization could fail on valid knocks to a room.
```
clokep, line 1: To ensure I understand, this is causing a user to not be able to go from knock -> leave?

richvdh: argh. Changes to the auth algorithm mean that different implementations reach different conclusions about room state, which can lead to split-braining.

This is exactly the same debate we're currently having elsewhere about stringy PLs: in theory we can't change this without a new room version.

turt2live: I'm highly unconvinced that this particular issue needs spec involvement. 

The path already would have caused different conclusions to be reached, thus we should just fix it and off we go.

anoadragon453: Note that we do have tests for knocking in Complement. Checking that we can successfully rescind a knock is one of them: https://github.com/matrix-org/complement/blob/154fdc069700bf43694281505da5bf9e2f79bf21/tests/msc2403_test.go#L169-L178

synapse/event_auth.py
```diff
@@ -369,7 +369,7 @@ def _is_membership_change_allowed(
 
     # Require the user to be in the room for membership changes other than join/knock.
     if Membership.JOIN != membership and (
```
richvdh, line 371: why are all these conditions written backwards :exploding_head:

synapse/event_auth.py
```diff
@@ -369,7 +369,7 @@ def _is_membership_change_allowed(
...
-        RoomVersion.msc2403_knocking and Membership.KNOCK != membership
+        room_version.msc2403_knocking and Membership.KNOCK != membership
```
richvdh, line 372: this doesn't seem right. The full condition is now:

`Membership.JOIN != membership and room_version.msc2403_knocking and Membership.KNOCK != membership`.

For room versions that do not support knocking, that condition is false, so the tests inside this block (in particular, you need to be a member of the room to send an invite/kick/ban event) will not take place. I suspect (I haven't checked) that's the reason CI is failing.

changelog.d/11900.bugfix
richvdh: @turt2live I'm still slighty confused about which transitions we're fixing here. Note that `RoomVersion.msc2403_knocking` is truthy, so the current condition is equivalent to:

```
membership != Membership.JOIN and membership != Membership.KNOCK
```

I don't think that actually causes a bug here, but please prove me wrong.

richvdh (changes requested):

synapse/event_auth.py
turt2live: (we need to maintain the room version check because `knock` is not a valid transition in other room versions)

dkasak: @turt2live: Perhaps I phrased it poorly, but my point is the `knock` transition is forbidden further down in the code https://github.com/matrix-org/synapse/blob/cc3170bdce3f7b7021ce25d1bb066daae16dae78/synapse/event_auth.py#L458 regardless of whether the condition on line 372 has the check for the room version or not. It seems correct to let it be handled there.

turt2live: So after looking at this for a long hard while, I'm even more confused than when I started. The code is clearly a typo but indeed doesn't really make sense at all intent-wise. 

The following is a diagram from the spec:


**Note**: There are 3 bugs I noticed while writing this. `invite->knock (via /knock only)`, `ban->invite (via /invite only)`, and `external->leave (via /kick only)` are both valid transitions not covered by the current diagram.

If we further annotate the diagram with illegal transitions, visually de-prioritizing the legal ones, we end up with this:

membership illegal

The red, thick, lines denote the ones we realistically care about. The orange ones are entirely `*->external` which is "we can't delete state, so duh". The solid pink/purple ones are legal transitions from the previous diagram, but with conditions on being able to happen.

So, taking this into account and trying to understand the code: I think it's trying to protect against the thick red lines specifically - the categorically illegal transitions, particularly when the event sender and state key are the same (a self membership change, like a unbanning yourself). Because the user is not legally in the room, they should not be able to perform these transitions.

The code however represents this in a confusing way. If I understand it correctly (I'm not confident that I do), it's trying to say: if the membership equates to not being "in" the room (knocked or joined), then disallow unless the user is trying to leave on their own terms. 

To account for this understanding, I think the condition is *meant* to be: 
```python
if membership != Membership.JOIN or (room_version.msc2403_knocking and membership != Membership.KNOCK):
```

The `if` contained within will do the self-leave check, and will also filter out other illegal transitions like `ban->leave` (without `/unban`) because of the `caller_invited or caller_knocked` and `target_user_id == event.user_id` checks.

So the question is: does this make sense, and is it correct? @dkasak @richvdh 

(if yes, I'll also figure out how to condense this into a useful changelog entry. I still don't think it needs a new room version, but also somewhat unclear if this is the actual problem)

dkasak: > if the membership equates to not being "in" the room (knocked or joined), then disallow unless the user is trying to leave on their own terms.

I think your understanding is correct. To paraphrase, it's trying to say "For memberships other than join or knock do this check: If the user is invited or knocked *and* now trying to leave, allow. Otherwise, deny."

So we want to skip over that entire block if the new membership is something other than join or knock. We want to do this *regardless* of whether knocking is supported.

If knocking *is* supported, we want to skip over that code block, since knocks are one of two memberships for which we don't want to execute that code block.

If knocking *isn't* supported, we still don't want to enter that code block only to report to the user that "X not in room Y." That makes no sense. Instead, if there is no knocking support, we want to say "Unknown membership knock". This will happen by falling through and failing the condition on line 458, and ending up in the else branch.

```python
if membership != Membership.JOIN or (room_version.msc2403_knocking and membership != Membership.KNOCK):
```

I don't think this is correct. Consider what happens with this condition when membership is `join`
and knocking is supported.
```
>>> membership = "join"
>>> knocking_supported = True
>>> membership != "join" or (knocking_supported and membership != "knock")
True
```

clokep: The guard for the room version is there because `knock` doesn't have any meaning if the room version doesn't support it.

The condition does look suspect as is though, the goal was to only have the `membership != KNOCK` check if the room version supports that, otherwise only have the `membership != JOIN`.

It is a bit weird though that we then have a `caller_knocked` check later on though saying that they can leave if they've knocked. So maybe this conditional really is meant to just be if they're *not in the room*, meaning not joined?

changelog.d/11900.bugfix
turt2live: Let's ignore this thread until the others are resolved please.

synapse/event_auth.py
turt2live: > Then if someone tried sending `knock` to a room which doesn't support knocking while not being a member of that room, we would report "USER not in room ROOM". Is this correct?
> 
> I would assume we would want to report "Unknown membership knock" instead.

Indeed, we should be saying "unknown membership" instead. 

With that being said, @richvdh's [comment on another thread](https://github.com/matrix-org/synapse/pull/11900#discussion_r801983820) sparked a little bit more investigation since my last comment. Specifically that `RoomVersion.msc2403_knocking` is actually truthy, so the condition is always `membership != join && true && membership != knock` (boiling down to just `membership != join && membership != knock`). I've been assuming that using the class member would result in a `None`-like type, thus falsy, but turns out that's wrong.

With all the surrounding code and other auth checks we still ultimately validate all the membership transitions correctly, which leaves the remaining question as to why we need to check room version at all. Turns out, @dkasak was right all along and we don't need to. The part I was missing was that it's impossible to have `caller_knocked` ever be true in an un-knockable room because of two other conditions. For largely my own sake, those conditions are:

1. `caller_knocked` itself checks the room version
2. Sending a `membership` of `knock` will trigger the "Unknown membership" auth error because of the room version check in the later if/else ladder. 

The failing `caller_knocked` check would mean an appropriate "Not in room" error for the join-requiring membership transitions.

However, if we want to maintain a room version check here for sanity, it'd look something like this:
```python
    if Membership.JOIN != membership and (
        not room_version.msc2403_knocking or Membership.KNOCK != membership
    ):
```

So coming full circle: do people prefer to see the room version check for safety, or is a simplified `membership not in [join, knock]` more suitable?

turt2live: also for my own sake: limiting the condition to just `membership != join` means that initial knocks can't happen (`external->knock`) because "user not in room".

richvdh (approved):
thanks

Issue #11152, opened 2021-10-21
Room namespace registrations are not considered when sending EDUs to Application Services


anoadragon453:
Application Services can specify user, room ID and room alias namespaces in [their registration file](https://matrix.org/docs/spec/application_service/r0.1.2#registration). The idea is such that any event that would involve either a user or a room specified by those namespaces would be sent to the homeserver.

In the case of Ephemeral Data Units (aka EDUs, like typing events, presence updates, read receipts etc), only the *user* namespace is taken into account when determining whether to forward an event to an application service. However, room ID and alias namespaces are *not* considered:

Read receipts
https://github.com/matrix-org/synapse/blob/eb9ddc8c2e807e691fd1820f88f7c0bf43822661/synapse/handlers/receipts.py#L263-L269

Typing events
https://github.com/matrix-org/synapse/blob/aa2c027792d04c36b17866710e95a41d31f5d99c/synapse/handlers/typing.py#L479-L482

As an example, if I read a message in room X (meaning I sent a read receipt for a message in that room) an appservice should receive that read receipt in 3 cases:

1. A user that lies within the appservice's user namespace is in the room.
2. The room's ID is in the appservice's room ID namespace.
3. The room's alias is in the appservice's room alias namespace.

As it stands, only 1. actually succeeds today. That means if an appservice has a room alias namespace which includes room X - yet none of the users in the room are in that appservice's user namespace - then that appservice will **not** receive the read receipt. It should have done, and [MSC2409](https://github.com/Sorunome/matrix-doc/blob/soru%2Bhs/appservice-edus/proposals/2409-appservice-edus.md#expectations-of-when-an-edu-should-be-pushed-to-an-appservice) states as such.

This is currently only an issue for read receipts and typing events (EDUs which are tied to a room). Presence updates aren't tied to a room, and are based on which users you share a room with. This logic is correct.

PR #11903, opened 2022-02-03
Derive users to notify application services of edus from user streams in `on_new_event`

anoadragon453:
Otherwise we were completely ignoring all users discovered from rooms being passed to `on_new_event`, here:

https://github.com/matrix-org/synapse/blob/c15d37575882b976a99e5e616d738d904ba3e754/synapse/notifier.py#L437-L438

Luckily, the only EDU-handling caller that sets the `rooms` argument of `on_new_event` is the handler of device list updates, which haven't yet been sent to application services before. So this didn't cause any EDUs to get dropped on the floor. But it's something that we'll need to fix for implementing #11881.

Note that this is separate from, and does not fix https://github.com/matrix-org/synapse/issues/11152.

Best reviewed commit-by-commit.


anoadragon453:
It turns out user and room streams are only created on calls to `/sync` or `/events`, which applications services wouldn't and shouldn't have to do in this context.

So using the streams here is not a good solution.

PR #8437, opened 2020-10-01
Send some ephemeral events to appservices

Half-Shot:
This PR is split from #8366.

This change will cause Synapse to send ephemeral events to appservices in accordance with [MSC2409](https://github.com/matrix-org/matrix-doc/pull/2409). This PR only sends typing, presence and read receipt information to appservices, but does not handle device information such as to_device messages or device list changes which will be part of #8366.

PR #11905, opened 2022-02-03
Notify users, rather than rooms, of device list updates

anoadragon453:
When a device list update occurs, [`Notifier.on_new_event`](https://github.com/matrix-org/synapse/blob/610578e27177136b6a4b187783c490d35c688272/synapse/notifier.py#L406) is called with the key `device_list`. This method determines the set of users to notify of this update from both a collection of `users` and a collection of `rooms`. 

Users are consolidated before notifying (waking up sync streams etc.) occurs. `Notifier.rooms_to_users_state` is used to translate from a given room ID to a set of users in the room. Those users are then merged into the list of those to notify.

This all works pretty well, until we [hooked into this method](https://github.com/matrix-org/synapse/pull/8437) to send EDUs to AS's:

https://github.com/matrix-org/synapse/blob/610578e27177136b6a4b187783c490d35c688272/synapse/notifier.py#L461-L472

This method only receives the original collection of users, not the one derived from rooms. Eventually, it should accept a list of room IDs as well (https://github.com/matrix-org/synapse/issues/11152), but not in the case of device list updates.

Technically speaking, device list updates aren't tied to a room. We just use rooms to determine who to send our device list updates to.

Therefore, this PR aims to fix both problems:

* For device list changes, `on_new_event` should receive a collection of users, derived from the rooms the requester is in. Relevant users should then be notified.
* `AppserviceHandler.notify_interested_services_ephemeral` will automatically receive a complete list of users who should be notified, allowing us to send device list updates to AS's.

### DIFF
--- a/changelog.d/11905.misc
+++ b/changelog.d/11905.misc
@@ -1,0 +1,1 @@
+Some changes to eventually support the implementation of sending device list updates to application services.
--- a/changelog.d/11905.misc
+++ b/changelog.d/11905.misc
@@ -1,1 +1,1 @@
-Some changes to eventually support the implementation of sending device list updates to application services.
+Preparation to support sending device list updates to application services.
--- a/synapse/handlers/device.py
+++ b/synapse/handlers/device.py
@@ -495,13 +495,11 @@ class DeviceHandler(DeviceWorkerHandler):
                 "Notifying about update %r/%r, ID: %r", user_id, device_id, position
             )
 
-        room_ids = await self.store.get_rooms_for_user(user_id)
-
         # specify the user ID too since the user should always get their own device list
         # updates, even if they aren't in any rooms.
-        self.notifier.on_new_event(
-            "device_list_key", position, users=[user_id], rooms=room_ids
-        )
+        users_to_notify = users_who_share_room.union(user_id)
+
+        self.notifier.on_new_event("device_list_key", position, users=users_to_notify)
 
         if hosts:
             logger.info(
--- a/synapse/handlers/device.py
+++ b/synapse/handlers/device.py
@@ -497,7 +497,7 @@ class DeviceHandler(DeviceWorkerHandler):
 
         # specify the user ID too since the user should always get their own device list
         # updates, even if they aren't in any rooms.
-        users_to_notify = users_who_share_room.union(user_id)
+        users_to_notify = users_who_share_room.union({user_id})
 
         self.notifier.on_new_event("device_list_key", position, users=users_to_notify)
 


synapse/handlers/device.py
```diff
@@ -495,13 +495,11 @@ async def notify_device_update(
...
             )
 
-        room_ids = await self.store.get_rooms_for_user(user_id)
-
         # specify the user ID too since the user should always get their own device list
         # updates, even if they aren't in any rooms.
-        self.notifier.on_new_event(
-            "device_list_key", position, users=[user_id], rooms=room_ids
-        )
+        users_to_notify = users_who_share_room.union(user_id)
```
clokep, line 500: I'm a bit confused by this, it doesn't seem to just be re-using something but passing in a totally different value?

I think this is wrong though, `.union(user_id)` will cause each character to be iterated and added separately to the result.

changelog.d/11905.misc
```diff
@@ -0,0 +1 @@
+Some changes to eventually support the implementation of sending device list updates to application services.
```
clokep, line 1: This is wordy:

```suggestion
Preparation to support sending device list updates to application services.
```

synapse/handlers/device.py
anoadragon453: > I'm a bit confused by this, it doesn't seem to just be re-using something but passing in a totally different value?

We are creating a new value, but through copying rather than recalculating it.

`on_new_event` eventually needs to work out which users to send the updates to. Doing this by deriving a list of users from a list of room IDs is expensive. We already have an appropriate list of users to notify in `users_who_share_room`, so this PR proposes to just use that.

I've just realised another angle to this change. Device list updates are separate from rooms, and thus shouldn't end up passing a list of room IDs to `on_new_event`, but a list of users. For instance, if an application service explicitly put a room's ID in its room ID namespace (effectively subscribing to updates from the room), it shouldn't then receive device list updates from users in that room. It should only get them if a user in the room falls within its user namespace, as device list updates are between users.

> I think this is wrong though, `.union(user_id)` will cause each character to be iterated and added separately to the result.

Ah, that is a silly mistake. I was surprised CI didn't catch it though. I've written a test to do so in the future: https://github.com/matrix-org/synapse/pull/11909.

clokep: Over chat @anoadragon453 explained to me that this fixes a case of local users not having their sync streams woken up when users in the same room have new devices (that was my interpretation at least).

Do we need to test for this behavior? Are there any currently open issues about this? (I thought we had one, but can't find it on looking again.)

clokep (changes requested):
From our conversation over Matrix @anoadragon453 is going to attempt to add a test for this, if it seems non-trivial we'll re-evaluate.

anoadragon453:
I wrote a test to check that device list updates come down to local users when they share a room in Complement at https://github.com/matrix-org/complement/pull/305.

However part-way through I realised that there wouldn't be a simple way of verifying with Complement that the user stream would be updated properly. In Complement, we often check if something comes down /sync *eventually* - something we do to prevent race conditions.

But in this case we're trying to check that the *current* sync stream is woken up; something that you can really only do with a unit test.

So not wasted effort. But a unit test will be written up tomorrow instead.

synapse/handlers/device.py
anoadragon453: > this fixes a case of local users not having their sync streams woken up when users in the same room have new devices (that was my interpretation at least).

Correct interpretation! (Un)fortunately, my assertion was not correct. `/sync` *does* end up populating the `room_to_user_streams` dict of `Notifier` on each `/sync`. This hidden bit in the call stack does it:

https://github.com/matrix-org/synapse/blob/64ec45fc1b0856dc7daacca7d3ab75d50bd89f84/synapse/notifier.py#L487-L488

Having the appropriate entry in `room_to_user_streams` means that passing a room ID to `Notifier.on_new_event` will correctly translate to, and thus wake up, the associated user stream. And lo, `/sync` streams do get woken up correctly in this case.

Understanding this did take writing a unit test to really prove though - and hooray for more tests! Tests have been put up separately in https://github.com/matrix-org/synapse/pull/11972.

So what remains for this PR? It doesn't fix a bug by passing users to `on_new_event`; `room_ids` works perfectly fine. The only benefit at this point is to save recalculating the list of users from the given `room_ids` before we eventually call `notify_interested_services_ephemeral` at the end of `on_new_events`.

Now that I have a better understanding of how the notifier code works, I've realised that calling `notify_interested_services_ephemeral` in `on_new_event` is a bit of a hack. It's probably better to wrap this method up into a Notifier callback. We'd need to set up user streams on appservice registration though... a topic for another issue.

For now, a solution is to have `notify_interested_services_ephemeral` accept a collection of room IDs and use those to deduce which users to notify. It is extra effort over what this PR proposes, but I'm not convinced that this PR is the Right Way To Do Things:tm: anymore.

So! Closing this, and will reopen with a change that doesn't influence other (nuanced) components so much.

clokep: Thanks for the explanation! It sounds like this code is confusingly entangled with other bits. 😢

anoadragon453: Actually actually, on further reflection, creating a user stream for every AS user would be much more overhead than simply calculating which AS's to send an event to when it happens.

So... the current code isn't really a hack. In fact, it's probably the best way to do it. But the AS-related code in `Notifier` code use a refactor to make different paths better isolated from each other.

anoadragon453:
Closing this as per https://github.com/matrix-org/synapse/pull/11905#discussion_r804743202.

clokep (approved):
I think this is OK.

My follow-up question is if `on_new_event` needs to take room IDs, but it seems to be used in my places still! 👍

anoadragon453:
> My follow-up question is if on_new_event needs to take room IDs, but it seems to be used in my places still!

Yes indeed, even for the `notify_interested_services_ephemeral` use case! For events that are tied to a room, the room ID should be included. This could then be used to notify appservices that [explicitly subscribe to updates from a room](https://spec.matrix.org/v1.2/application-service-api/#registration), which they can do even if they haven't specified a user namespace.

PR #11912, opened 2022-02-04
Add a docstring to `add_device_change_to_streams` and fix some nearby types

anoadragon453:
Yet more refactors split out from #11881.

Public functions lacking docstrings, and some missing/incorrect type hints.

### DIFF
--- a/synapse/storage/databases/main/devices.py
+++ b/synapse/storage/databases/main/devices.py
@@ -1584,7 +1584,7 @@ class DeviceStore(DeviceWorkerStore, DeviceBackgroundUpdateStore):
         txn: LoggingTransaction,
         user_id: str,
         device_ids: Iterable[str],
-        hosts: Iterable[str],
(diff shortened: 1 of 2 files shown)


synapse/storage/databases/main/devices.py
```diff
@@ -1573,11 +1583,11 @@ def _add_device_outbound_poke_to_stream_txn(
         self,
         txn: LoggingTransaction,
         user_id: str,
-        device_ids: Collection[str],
-        hosts: List[str],
+        device_ids: Iterable[str],
+        hosts: Iterable[str],
```
erikjohnston, line 1587: I think this has to be a `Collection` as we iterate over it multiple times?

anoadragon453: Oh, good catch.

erikjohnston (approved):

PR #11913, opened 2022-02-04
Tests: replace mocked Authenticator with the real thing

richvdh:
If we prepopulate the test homeserver with a key for a remote homeserver, we
can make federation requests to it without having to stub out the
authenticator. This has two advantages:

 * means that what we are testing is closer to reality (ie, we now have
   complete tests for the incoming-request-authorisation flow)

 * some tests require that other objects be signed by the remote server (eg,
   the event in `/send_join`), and doing that would require a whole separate
   set of mocking out. It's much simpler just to use real keys.

### DIFF
--- a/tests/unittest.py
+++ b/tests/unittest.py
@@ -17,6 +17,7 @@ import gc
 import hashlib
 import hmac
 import inspect
+import json
 import logging
 import secrets
 import time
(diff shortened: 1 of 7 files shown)


tests/unittest.py
```diff
@@ -755,42 +757,116 @@ def inject_room_member(self, room: str, user: str, membership: Membership) -> No
...
-        federation_server.register_servlets(hs, self, authenticator, ratelimiter)
+
+def _auth_header_for_request(
+    origin: str,
+    destination: str,
+    signing_key: signedjson.key.SigningKey,
+    method: str,
+    path: str,
+    content: Optional[JsonDict],
+) -> str:
```
erikjohnston, line 850: Slightly surprised we don't already have a function that we can reuse to do this, but :shrug:

erikjohnston (approved):

PR #11915, opened 2022-02-04
Rename various ApplicationServices interested methods

anoadragon453:
Another splitting out of refactoring work from #11881. This PR attempts to simplify the mental model of part of the ApplicationService class's interface - that which revolves around checking the AS's interest in various entities (rooms, users, events etc). It also attempts to correct some missing checks that may have been a result of the complexity.

In summary, the main changes are:

* Rename existing `ApplicationService` class methods:
	  * ApplicationServices.is_interested -> is_interested_in_event and add caching
	  * ApplicationService.is_interested_in_user -> is_user_in_namespace
	  * ApplicationService.is_interested_in_alias -> is_room_alias_in_namespace
	  * ApplicationService.is_interested_in_room_id -> is_room_id_in_namespace
* Add `is_interested_in_user` and `is_interested_in_room` methods which include more checks that the previous iterations
* Modified some calls to `ApplicationService.matches_user_in_member_list` - which only checked whether an AS has a member in the room - to instead call `ApplicationService.is_interested_in_room`, which additionally checks if the room ID and alias(es) of the room are in that AS's namespaces. This fixes https://github.com/matrix-org/synapse/issues/11152.

### DIFF
--- a/synapse/appservice/__init__.py
+++ b/synapse/appservice/__init__.py
@@ -263,19 +263,6 @@ class ApplicationService:
             room_id, store, on_invalidate=cache_context.invalidate
         )
 
-    def _matches_room_id(self, event: EventBase) -> bool:
-        if hasattr(event, "room_id"):
-            return self.is_room_id_in_namespace(event.room_id)
-        return False
-
-    async def _matches_aliases(self, event: EventBase, store: "DataStore") -> bool:
-        alias_list = await store.get_aliases_for_room(event.room_id)
-        for alias in alias_list:
-            if self.is_room_alias_in_namespace(alias):
-                return True
-
-        return False
-
     async def is_interested_in_event(
         self, event: EventBase, store: "DataStore"
     ) -> bool:
--- a/synapse/appservice/__init__.py
+++ b/synapse/appservice/__init__.py
@@ -214,6 +214,68 @@ class ApplicationService:
                 return True
         return False
 
+    def is_interested_in_user(
+        self,
+        user_id: str,
+    ) -> bool:
+        """
+        Returns whether the application is interested in a given user ID.
+
+        The appservice is considered to be interested in a user if either: the
+        user ID is in the appservice's user namespace, or if the user is the
+        appservice's configured sender_localpart.
+
+        Args:
+            user_id: The ID of the user to check.
+
+        Returns:
+            True if the application service is interested in the user, False if not.
+        """
+        return (
+            # User is the appservice's sender_localpart user
+            user_id == self.sender
+            # User is in the appservice's user namespace
+            or self.is_user_in_namespace(user_id)
+        )
+
+    @cached(num_args=1, cache_context=True)
+    async def is_interested_in_room(
+        self,
+        room_id: str,
+        store: "DataStore",
+        cache_context: _CacheContext,
+    ) -> bool:
+        """
+        Returns whether the application service is interested in a given room ID.
+
+        The appservice is considered to be interested in the room if either: the ID or one
+        of the aliases of the room is in the appservice's room ID or alias namespace
+        respectively, or if one of the members of the room fall into the appservice's user
+        namespace.
+
+        Args:
+            room_id: The ID of the room to check.
+            store: The homeserver's datastore class.
+
+        Returns:
+            True if the application service is interested in the room, False if not.
+        """
+        # Check if this room ID matches the appservice's room ID namespace
+        if self.is_room_id_in_namespace(room_id):
+            return True
+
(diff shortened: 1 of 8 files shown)


tests/handlers/test_appservice.py
```diff
@@ -325,17 +324,19 @@ def test_notify_interested_services_ephemeral_out_of_order(self):
             interested_service, ephemeral=[]
         )
 
-    def _mkservice(self, is_interested, protocols=None):
+    def _mkservice(self, is_interested: bool, protocols=None):
```
clokep, line 327: Should we update this parameter name too while we're here?

tests/handlers/test_appservice.py
```diff
@@ -325,17 +324,19 @@ def test_notify_interested_services_ephemeral_out_of_order(self):
...
+        service.is_interested_in_user.return_value = make_awaitable(is_interested)
+        service.is_interested_in_room.return_value = make_awaitable(is_interested)
         service.token = "mock_service_token"
         service.url = "mock_service_url"
         service.protocols = protocols
         return service
 
-    def _mkservice_alias(self, is_interested_in_alias):
+    def _mkservice_alias(self, is_room_alias_in_namespace):
         service = Mock()
```
clokep, line 338: Type hint on the param while we're here?

synapse/appservice/__init__.py
```diff
@@ -263,8 +263,9 @@ async def is_interested_in_room(
             room_id, store, on_invalidate=cache_context.invalidate
         )
 
+    @cached(num_args=1, cache_context=True)
```
clokep, line 266: I'm not sure this will work exactly as we're hoping -- `EventBase` equality is checked by it being the exact same object (i.e. if you built two `EventBase` instances with identical properties they aren't equal).

This might be OK, but I'm not 100% sure.

synapse/appservice/__init__.py
```diff
@@ -316,7 +316,7 @@ async def is_interested_in_presence(
 
         # Then find out if the appservice is interested in any of those rooms
         for room_id in room_ids:
-            if await self.matches_user_in_member_list(
+            if await self.is_interested_in_room(
```
clokep, line 319: Sorry -- I'm not following this change, it seems bigger than just "make `matches_user_in_member_list` private"

I think the change is OK from the rest of the commit message, but I'm having trouble following it.

anoadragon453: Ah, yes it is. `is_interested_in_room` calls `_matches_user_in_member_list`, but it will also return True if:

* the room's ID is part of the application service's room ID namespace
* one of the room's aliases are part of the application service's room alias namespace

which is the intended functionality according to [the spec](https://spec.matrix.org/v1.2/application-service-api/#registration):

> Application services register “namespaces” of user IDs, room aliases and room IDs. These namespaces are represented as regular expressions. An application service is said to be “interested” in a given event if one of the IDs in the event match the regular expression provided by the application service, such as the room having an alias or ID in the relevant namespaces. Similarly, the application service is said to be interested in a given event if one of the application service’s namespaced users is the target of the event, or is a joined member of the room where the event occurred.

synapse/appservice/__init__.py
anoadragon453: Ah, good catch. Ideally we'd cache on the event's ID instead of `EventBase`. I don't suppose there's an easy way to do that with `@cached`, though I can just create an `LruCache` much as we do for `_get_event_cache`:

https://github.com/matrix-org/synapse/blob/e2e1d90a5e4030616a3de242cde26c0cfff4a6b5/synapse/storage/databases/main/events_worker.py#L240-L243

tests/handlers/test_appservice.py
anoadragon453: The functionality of this `is_interested` parameter has actually changed from just tweaking whether the created appservice mock will be interested in events, to any event, user or room.

I've fixed this, as well as added docstrings and types in eb779e030.

synapse/appservice/__init__.py
anoadragon453: From the [conversation in #synapse-dev](https://matrix.to/#/!XaqDhxuTIlvldquJaV:matrix.org/$mflf7_3Je3Cm_wzGNMP_u_ShDFxb2qqoVqGFMkjC8NE?via=matrix.org&via=element.io&via=vector.modular.im), I believe we concluded on adding an `event_id` argument to `is_interested_in_event`. Doing so was simple, though updating the tests was slightly tedious.

Done in 453fdacb7!

clokep (approved):

PR #11936, opened 2022-02-08
Don't attempt URL previews for `audio/*` and `video/*` content types

dkasak:
This prevents Synapse from attempting to construct URL previews for URLs resolving to a `Content-Type` with the prefixes `audio/` and `video/`. This makes sense given that Synapse will never generate a URL preview for such a file in the first place, and will help prevent issues with streaming media servers, such as #8302.

### DIFF
--- a/synapse/http/client.py
+++ b/synapse/http/client.py
@@ -726,6 +726,14 @@ class SimpleHttpClient:
                 HTTPStatus.BAD_GATEWAY, "Got error %d" % (response.code,), Codes.UNKNOWN
             )
 
+        if b"Content-Type" in resp_headers:
+            content_type = resp_headers[b"Content-Type"][0]
+            if _is_av_media(content_type):
+                raise SynapseError(
+                    HTTPStatus.BAD_GATEWAY,
+                    ("Unsupported content type for URL previews: %r" % content_type),
+                )
+
         # TODO: if our Content-Type is HTML or something, just read the first
         # N bytes into RAM rather than saving it all to disk only to read it
         # straight back in again
@@ -760,6 +768,12 @@ class SimpleHttpClient:
             response.request.absoluteURI.decode("ascii"),
             response.code,
         )
+
+
+def _is_av_media(content_type: bytes) -> bool:
+    return content_type.lower().startswith(
+        b"video/"
+    ) or content_type.lower().startswith(b"audio/")
 
 
 def _timeout_to_request_timed_out_error(f: Failure):
--- a/tests/rest/media/v1/test_url_preview.py
+++ b/tests/rest/media/v1/test_url_preview.py
@@ -243,6 +243,78 @@ class URLPreviewTests(unittest.HomeserverTestCase):
         self.assertEqual(channel.code, 200)
         self.assertEqual(channel.json_body["og:title"], "\u0434\u043a\u0430")
 
+    def test_video_rejected(self):
+        self.lookups["matrix.org"] = [(IPv4Address, "10.1.2.3")]
+
+        end_content = b"anything"
+
+        channel = self.make_request(
+            "GET",
+            "preview_url?url=http://matrix.org",
+            shorthand=False,
+            await_result=False,
+        )
+        self.pump()
+
+        client = self.reactor.tcpClients[0][2].buildProtocol(None)
+        server = AccumulatingProtocol()
+        server.makeConnection(FakeTransport(client, self.reactor))
+        client.makeConnection(FakeTransport(server, self.reactor))
+        client.dataReceived(
+            (
+                b"HTTP/1.0 200 OK\r\nContent-Length: %d\r\n"
+                b"Content-Type: video/mp4\r\n\r\n"
+            )
+            % (len(end_content))
+            + end_content
+        )
+
+        self.pump()
+        self.assertEqual(channel.code, 502)
+        self.assertEqual(
+            channel.json_body,
+            {
+                "errcode": "M_UNKNOWN",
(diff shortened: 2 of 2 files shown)


changelog.d/11936.bugfix
```diff
@@ -0,0 +1 @@
+Don't attempt URL previews for URLs with audio/* or video/* content types. This prevents Synapse from making useless longer-lived connections to streaming media servers.
```
clokep, line 1:
```suggestion
Don't attempt URL previews for URLs with `audio/*` or `video/*` content types. This prevents Synapse from making useless longer-lived connections to streaming media servers.
```

synapse/http/client.py
```diff
@@ -726,6 +726,14 @@ async def get_file(
                 HTTPStatus.BAD_GATEWAY, "Got error %d" % (response.code,), Codes.UNKNOWN
             )
 
+        if b"Content-Type" in resp_headers:
```
clokep, line 729: Unfortunately `get_file` is meant to be a generic utility, not specific to URL previews, so I don't think putting this check here makes much sense. Maybe passing in a list of allowed (or not allowed) prefixes would work though?

dkasak: Ah right, this is used for media downloads as well. How about passing a predicate function rather than a list? That makes it easier to encode validating by prefix and other possible behaviours.

clokep: Yep, that sounds reasonable to me!

synapse/http/client.py
```diff
@@ -726,6 +733,14 @@ async def get_file(
                 HTTPStatus.BAD_GATEWAY, "Got error %d" % (response.code,), Codes.UNKNOWN
             )
 
+        if is_allowed_content_type and b"Content-Type" in resp_headers:
+            content_type = resp_headers[b"Content-Type"][0]
+            if not is_allowed_content_type(content_type):
+                raise SynapseError(
+                    HTTPStatus.BAD_GATEWAY,
+                    ("Unsupported content type for URL previews: %r" % content_type),
```
clokep, line 741: I don't think it is great to leak the `b` prefix in our error messages:

```suggestion
                    ("Unsupported content type for URL previews: %s" % (content_type.decode(errors='ignore'),)),
```

This kind of leaks that we're using this only for URL previews, but not sure there's a better thing to do besides just letting the passed in callable raise.

clokep (commented):
Looks pretty good! I made a couple of suggestions for readability!

synapse/http/client.py
dkasak: > This kind of leaks that we're using this only for URL previews, but not sure there's a better thing to do besides just letting the passed in callable raise.

Yeah, now that you pointed it out, it sticks out as ugly. I think I'll just say "Requested file's content type not allowed for this operation" and live with the less specific error message. Looking the calling contexts of `get_file`, the operation being performed should be obvious from the surrounding log lines anyway.

clokep: Sounds perfect. Thank you!

synapse/rest/media/v1/preview_url_resource.py
```diff
@@ -403,6 +403,7 @@ async def _download_url(self, url: str, output_stream: BinaryIO) -> DownloadResu
                 output_stream=output_stream,
                 max_size=self.max_spider_size,
                 headers={"Accept-Language": self.url_preview_accept_language},
+                is_allowed_content_type=_is_not_av_media,
```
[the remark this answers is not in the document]
clokep: It isn't just `text/html`, there's a handful of support formats, but it is pretty much text and images. (I think the oEmbed bits go through a different API.)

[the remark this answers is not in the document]
clokep: It looks like this gets used in 3 places:

1. To download the previewed URL, which is expected to have a `Content-Type` of:
	1. `image/`
	2. `text/html` or `application/xhtml`
	3. `application/json` if it was an oEmbed URL
2. Pre-caching images, which is expected to have a `Content-Type` of `image/`

Seems simpler enough to do via a list of allowed properties, might be a bit finicky if we add more, but I don't think there are plans for that.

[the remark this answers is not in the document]
dkasak: Sure, if we can exhaustively list the types we want to preview, an allow list is even better. Pushed a commit that does this.

synapse/rest/media/v1/preview_url_resource.py
```diff
@@ -761,3 +762,16 @@ def _is_html(content_type: str) -> bool:
...
+    """Returns True for content types for which we will perform URL preview and False
+    otherwise."""
+
+    content_type = content_type.lower()
+    return (
+        content_type.startswith(b"text/html")
+        or content_type.startswith(b"application/xhtml")
+        or content_type.startswith(b"image/")
+        or content_type.startswith(b"application/json")
+    )
```
clokep, line 777: Probably best to refer to our helpers that already exist so we can update just a single spot for each of them in the future.

```suggestion
    return (
        _is_media(content_type)
        or _is_html(content_type)
        or _is_json(content_type)
    )
```

clokep (approved):
LGTM! Left one minor comment, but I don't feel too strongly about it. Let me know if there are any other thoughts before we merge!

dkasak:
It doesn't, but the attacker will only be DoS-ing themselves in that case. How does that help them?

Note that there are separate time and size limits too.

anarcat:
> It doesn't, but the attacker will only be DoS-ing themselves in that case. How does that help them?

there are two ends to a potential DoS attack...

PR #11939, opened 2022-02-08
Remove unnecessary ignores due to Twisted upgrade.

### LINKED ISSUE #10899: Remove type-ignore for `IReactorTCP.connectTCP`'s `hostname` argument
The fix proposed in #10895 for #10851 requires a series temporary `# type: ignore`.

We should remove this as soon as we're able: namely, when twisted release a version which includes [this commit](https://github.com/twisted/twisted/commit/ec7445d6361ce04424b3570a81263dc6760f15de).

clokep:
Fixes #10899 now that Twisted has released a new update we don't need to ignore the host parameter for `connectTCP` anymore.

This will require updating local installs to the latest Twisted or mypy will fail.

### DIFF
--- a/changelog.d/11939.misc
+++ b/changelog.d/11939.misc
@@ -1,0 +1,1 @@
+Remove an unnecessary ignoring of type hints due to fixes in upstream packages.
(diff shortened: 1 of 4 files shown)


DMRobertson (approved):
Many thanks for this!

PR #11941, opened 2022-02-08
Support the `dir` parameter for `/relations`.

clokep:
Implements MSC3715 to add a `dir` parameter to `/relations`.

This API is currently unstable and the default behavior is the same, so I think it is safe to just add, but willing to hear alternatives!

### DIFF
--- a/changelog.d/11941.feature
+++ b/changelog.d/11941.feature
@@ -1,0 +1,1 @@
+Support the `dir` parameter on the `/relations` endpoint, per [MSC3715](https://github.com/matrix-org/matrix-doc/pull/3715).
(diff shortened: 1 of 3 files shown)


reivilibre (approved):
Code looks fine.

I'm not entirely convinced that shouldn't be unstable-prefixed; it hasn't made it into the spec after all. (It could be you have good reasons that I don't understand, though)
If the MSC is going to merge soon then perhaps it could be merged upon MSC merge.
Alternatively it would be nice to hear from someone more spec-savvy about whether we really want to ship this unstable feature sans prefix.

clokep:
> I'm not entirely convinced that shouldn't be unstable-prefixed; it hasn't made it into the spec after all. (It could be you have good reasons that I don't understand, though)

I've added a prefix. 👍

Issue #11942, opened 2022-02-08
1.52.0 fails to start with `prometheus_client` version error


olmari:
1.52.0 Debian Bullseye matrix.org repo, HS hacklab.fi

Upgrading workerized synapse to 1.52.0 and synapse won't start, nothing on logs either so very early fail at least without debug logs enabled. I emergency downgraded our HS to 1.51.0 and it started right up again.

While I promised to provide strace too on #synapse:matrix.org, I did promise it later when I can afford to halt our ~100 user homeserver. However there is few others who experienced similar isue, with venv pip installation too, and culplrit is likely needing to have also updated `treq` package alongside new `twist`.

> Our issue seems having originated from an incompatibility between twisted 22.1.0 and treq which was too old but did not get pulled in as a dependency during pip --upgrade.

> Manually upgrading treq and twisted to 22.1.0 has fixed the issue for us.
Related: RedHat [Bug 1953535 - ImportError: cannot import name '_PY3' from 'twisted.python.compat'](https://bugzilla.redhat.com/show_bug.cgi?id=1953535)

So there is very likely rootcause there.

I did make issue now as early as possible for all to know, will fill strace later if not confirmed before.

richvdh:
yeah, treq 20.4.0 and earlier are incompatible with recent versions of twisted (https://github.com/twisted/treq/issues/288). You'll need to upgrade that as well.

richvdh:
I just realised that @olmari is reporting this as a problem with the matrix.org debian packages, which should include up-to-date versions of both Twisted and treq, so this is very much unexpected.

richvdh:
Strongly suspect @olmari's issue is unrelated to treq. We'll need logs to investigate further.

If nothing is written to the log file, it is almost certain something will have been written to stderr. Assuming you're using systemd for process management, you should find something from `journalctl`.

olmari:
Indeed, might not be `treq` related after all, but `prometheus_client`.. We do have metrix enabled, should that be cornerstone.

`journalctl -eu matrix-synapse` gives:

```
Feb 08 16:06:58 morpheus systemd[1]: Starting Synapse master...
Feb 08 16:06:59 morpheus matrix-synapse[3385]: ERROR:root:Needed prometheus_client>=0.4.0,<0.13.0, got prometheus-client==0.13.1
Feb 08 16:06:59 morpheus matrix-synapse[3385]: Missing Requirements: "prometheus_client>=0.4.0,<0.13.0"
[...]

PR #11943, opened 2022-02-08
Recommend upgrading treq alongside twisted

babolivier:
Fixes https://github.com/matrix-org/synapse/issues/11942

Updates the upgrade notes and changelog. I've already updated https://github.com/matrix-org/synapse/releases/tag/v1.52.0.

### DIFF
--- a/changelog.d/11943.doc
+++ b/changelog.d/11943.doc
@@ -1,0 +1,1 @@
+Recommend upgrading treq alongside twisted.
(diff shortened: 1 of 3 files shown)


richvdh:
doesn't fix #11942, since that is (on closer inspection) a problem with our debs, not pip

babolivier:
> doesn't fix https://github.com/matrix-org/synapse/issues/11942, since that is (on closer inspection) a problem with our debs, not pip

Yep, I've updated the PR description, looks like we crossed streams here.

changelog.d/11943.doc
```diff
@@ -0,0 +1 @@
+Recommend upgrading treq alongside twisted.
```
richvdh, line 1: advise *against* a changelog file for a doc change targetting `master`: it won't make any sense to include this for 1.53.0.

richvdh (approved):
lgtm otherwise

PR #11950, opened 2022-02-09
Remove deprecated user_may_create_room_with_invites callback

babolivier:
The deprecation has been issued in https://github.com/matrix-org/synapse/pull/11206, with an advertised EOL of January 2022. I think it's safe to remove it for good now.


changelog.d/11950.removal
```diff
@@ -0,0 +1 @@
+Remove deprecated `user_may_create_room_with_invites` spam checker callback.
```
clokep, line 1: It looks like the upgrade notes mention a replacement for this. Should we update `UPGRADE.md` again with this information (or include it here or mention to see those notes or something else)?

synapse/module_api/__init__.py
```diff
@@ -217,9 +216,6 @@ def register_spam_checker_callbacks(
         user_may_invite: Optional[USER_MAY_INVITE_CALLBACK] = None,
         user_may_send_3pid_invite: Optional[USER_MAY_SEND_3PID_INVITE_CALLBACK] = None,
         user_may_create_room: Optional[USER_MAY_CREATE_ROOM_CALLBACK] = None,
-        user_may_create_room_with_invites: Optional[
```
clokep, line 220: Should we make the parameters to `register_spam_checker_callbacks` keyword only via the `*` flag?

babolivier: We could, I'd suggest opening a new issue for this.

changelog.d/11950.removal
babolivier: Yeah that sounds like a good idea.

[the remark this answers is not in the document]
babolivier: It's worth noting that the initial goal of this callback was to limit what invites users could send right after creating the room. The use cases you mention are valid and can probably be implemented using [`on_create_room`](https://matrix-org.github.io/synapse/develop/modules/third_party_rules_callbacks.html#on_create_room) (which the deprecation notice doesn't mention, I'll make sure that the removal notice does). Since this callback came out of a design failure on my part, was only partly working due to `user_may_invite` (see https://github.com/matrix-org/synapse/pull/11206 for details), and seemed niche enough that there would likely not be anyone needing it, it seemed like the right thing to remove it.

> how to upgrade existing code

The expectation is that this callback was quite short-lived so there should be very little existing code. Moreover, the deprecation notice was issued back in October, which gives ample time to existing projects to come talk to us about how we can resolve things.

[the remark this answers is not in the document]
Half-Shot: Yeah, it sounds like the ultimate solution to this is to provide more details in on_create_room. It's just a shame I came into need of this mechanism only two days ago. I'll see if I can propose something.

babolivier:
I have also replaced the `v1.(next)` section of the upgrade notes with `v1.53.0`

docs/upgrade.md
```diff
@@ -119,17 +130,15 @@ The new `capabilities`
...
 are now active by default.
 
-# Upgrading to v1.53.0
-
-## Dropping support for `webclient` listeners and non-HTTP(S) `web_client_location`
+## Removal of `user_may_create_room_with_invites`
 
-Per the deprecation notice in Synapse v1.51.0, listeners of type  `webclient`
-are no longer supported and configuring them is a now a configuration error.
+As announced with the release of [Synapse 1.47.0](https://matrix-org.github.io/synapse/latest/upgrade#deprecation-of-the-user_may_create_room_with_invites-module-callback),
```
clokep, line 135: It might make sense to use just the fragment here so if you're viewing it on GitHub or the website it works fine.

```suggestion
As announced with the release of [Synapse 1.47.0](#deprecation-of-the-user_may_create_room_with_invites-module-callback),
```

clokep (approved):

synapse/module_api/__init__.py
clokep: I put up a PR at #11975

Issue #11753, opened 2022-01-14
Stabilize support for MSC2675


clokep:
[MSC2675](https://github.com/matrix-org/matrix-doc/pull/2675) is currently in FCP (to complete tomorrow, barring any concerns). We have a few bits of work to do to stabilize support in Synapse for it:

* Add the `v1` prefix for `/relations`.
* Fix-up redaction support:
	* Ensure that redacted relations are not bundled nor returned from `/relations.
	* Support fetching relations of a redacted event. (Currently this returns nothing.)
* Properly handle relations for ignored users in `/relations` and bundled aggregations.

Those were all the differences that I found by reading through it, but there could be more.

clokep:
It is not explicit in the MSC, but it seems we also don't return any bundled aggregations for a redacted event. We likely should since that would match the behavior of being able to call `/relations` on it.

clokep:
Another piece of this that seems broken is that currently you cannot paginate this API using the results of a `/sync` or `/messages` API.

clokep:
> Another piece of this that seems broken is that currently you cannot paginate this API using the results of a `/sync` or `/messages` API.

This is due to the usage of `RelationPaginationToken` instead of `StreamToken` by the returned APIs. I think a way forward would be to:

* Switch the APIs to accept a serialized `StreamToken` or `RelationPaginationToken` (in which case we generate a `StreamToken` with only `room_key` filled in from the parsed `RelationPaginationToken`).
* Update the generated tokens to be serialied `StreamTokens`.

This all seems fairly straightforward (although finicky), but [the similar code in the `PaginationHandler`](https://github.com/matrix-org/synapse/blob/68acb0a29dcb03a0ecbcebdb95e09c5999598f42/synapse/handlers/pagination.py#L448-L491) does some rather complicated bits with topological ordering.

clokep:
This is done, minus removing support for the `unstable` `/relations` endpoint, which I think needs to wait for the next spec release.

clokep:
> This is done, minus removing support for the `unstable` `/relations` endpoint, which I think needs to wait for the next spec release.

I think Matrix v1.3 included this endpoint, so we can probably finish up this work now. Also see #14104.

PR #11952, opened 2022-02-09
Support pagination tokens from sync/messages in the relations API

clokep:
This is part of #11753 -- MSC2675 asserts that a token from `/sync` or `/messages` should be usable in `/relations`.

This is a bit complicated since the `/relations` (and `/aggregations`) APIs use custom pagination tokens, not stream tokens like we use elsewhere.

The approach to fix this is to:

1. Accept either stream tokens or legacy tokens to fetch the next batch.
2. Convert a legacy token to a stream token, if necessary.
3. Update all the code to use stream tokens.
4. Generate stream tokens when paginating.

This should be reviewable commit-by-commit. Some of them are small refactoring bits.

A couple of other thoughts I'd be interested in from a reviewer:

1. Should the shared logic between `RelationPaginationServlet` and `RelationAggregationGroupPaginationServlet` be moved somewhere? (A `RelationHandler` or maybe the `PaginationHandler`?)
2. Should I try to get rid of the weird `PaginationChunk` class that doesn't seem very useful? (Maybe easier to get rid of if the answer to 1 is "Yes"?)


synapse/storage/relations.py
```diff
@@ -75,7 +78,7 @@ def from_string(string: str) -> "RelationPaginationToken":
         except ValueError:
             raise SynapseError(400, "Invalid relation pagination token")
 
-    def to_string(self) -> str:
+    async def to_string(self, store: "DataStore") -> str:
```
reivilibre, line 81: how come? (& for the one below)

tests/rest/client/test_relations.py
```diff
@@ -241,6 +241,60 @@ def test_repeated_paginate_relations(self):
         found_event_ids.reverse()
         self.assertEquals(found_event_ids, expected_event_ids)
 
+    def test_pagination_from_sync_messages(self):
```
reivilibre, line 281: I'm having trouble following this one in general; anything you can do to clarify would be a big win I think, however I will try and make some notes about what I'm struggling with..

tests/rest/client/test_relations.py
```diff
@@ -241,6 +241,60 @@ def test_repeated_paginate_relations(self):
...
+        """Pagination tokens from sync and messages should work."""
```
reivilibre, line 282: 'should work' is a little bit vague

tests/rest/client/test_relations.py
```diff
@@ -241,6 +241,60 @@ def test_repeated_paginate_relations(self):
...
+        self.assertEquals(200, channel.code, channel.json_body)
+        annotation_id = channel.json_body["event_id"]
+        # Send an event after the relation events.
+        self.helper.send(self.room, body="Latest event", tok=self.user_token)
+
+        # Request sync, but limit the timeline to get a pagination token before
+        # the reaction.
+        filter = urllib.parse.quote_plus(
+            '{"room": {"timeline": {"limit": 1}}}'.encode()
+        )
```
reivilibre, line 293: Is this 'syncing from the top' as it were, so a later sync would include the 'A' reaction?

tests/rest/client/test_relations.py
```diff
@@ -241,6 +241,60 @@ def test_repeated_paginate_relations(self):
...
+            "GET", f"/sync?filter={filter}", access_token=self.user_token
+        )
+        self.assertEquals(200, channel.code, channel.json_body)
+        room_timeline = channel.json_body["rooms"]["join"][self.room]["timeline"]
+        sync_prev_batch = room_timeline["prev_batch"]
+        self.assertIsNotNone(sync_prev_batch)
+        # Ensure the event above is not in the batch.
+        self.assertNotIn(
+            annotation_id, [ev["event_id"] for ev in room_timeline["events"]]
+        )
```
reivilibre, line 304: For some reason this was pretty unclear to me.
I suppose `event above` wasn't the clearest and I managed to not notice `annotation_id` being snaked in from the reaction.
I didn't know that reactions were called annotations, so probably a PEBKAC.

tests/rest/client/test_relations.py
```diff
@@ -241,6 +241,60 @@ def test_repeated_paginate_relations(self):
...
+
+        channel = self.make_request(
```
reivilibre, line 306: It would be nice to have a bit of commentary as to what we're hoping to show with this request.

I think it's requesting the previous message before the end of the last sync... does that mean we're expecting to see the same message as we got during sync?

tests/rest/client/test_relations.py
```diff
@@ -241,6 +241,60 @@ def test_repeated_paginate_relations(self):
...
+            "GET",
+            f"/rooms/{self.room}/messages?dir=b&limit=1",
+            access_token=self.user_token,
+        )
+        self.assertEquals(200, channel.code, channel.json_body)
+        messages_end = channel.json_body["end"]
+        self.assertIsNotNone(messages_end)
+        # Ensure the event above is not in the chunk.
```
reivilibre, line 314: 'event above' → 'reaction event'?

tests/rest/client/test_relations.py
```diff
@@ -241,6 +241,60 @@ def test_repeated_paginate_relations(self):
...
+        self.assertNotIn(
+            annotation_id, [ev["event_id"] for ev in channel.json_body["chunk"]]
+        )
+
+        # Request the relations with the token from sync/messages -- this is a
+        # tiny bit silly since how do we know the parent ID? Just assume we had
+        # it from an earlier sync and the above are gappy.
```
reivilibre, line 321: This is a struggle; likely PEBKAC but I suppose it would help me to understand a bit more what's going on here anyway.

the token from sync/messages → from each of /sync and /messages ?

```suggestion
        # Request the relations with the token from each of /sync and /messages -- this is a
        # tiny bit silly since how do we know the parent ID? Just assume we had
        # it from an earlier sync and the above are gappy.
```

Parent ID? Is that the ID of the event which the reaction was sent for?

> and the above are gappy.

I'm struggling to follow that.

reivilibre (changes requested):
Slow to review due to my inexperience in the area, but seems reasonable.

Difficult to answer your questions right now; have messaged you

tests/rest/client/test_relations.py
clokep: Would this be clearer:

> Ensure the relation event is not in the batch returned from sync.

tests/rest/client/test_relations.py
clokep: "Parent ID" is a term used throughout this file, yes. It is the event that the relation is sent against.

I'm trying to say that we're filling in information that the client might be missing. I'll try to expand it.

tests/rest/client/test_relations.py
clokep: I'm not sure what you mean by "from the top", this returns the latest events, as a normal sync does.

tests/rest/client/test_relations.py
clokep: I rewrote a bunch of this in 62cba899b9b63ef6bf53ef7fb6ee54a7294aedd2, hopefully that makes it clearer.

tests/rest/client/test_relations.py
reivilibre: That is confusing to me, since it says the pagination token will be before the reaction; what am I missing?

tests/rest/client/test_relations.py
```diff
@@ -278,16 +278,16 @@ def test_repeated_paginate_relations(self):
         found_event_ids.reverse()
         self.assertEquals(found_event_ids, expected_event_ids)
 
-    def test_pagination_from_sync_messages(self):
-        """Pagination tokens from sync and messages should work."""
+    def test_pagination_from_sync_and_messages(self):
+        """Pagination tokens from /sync and /messages can be used to paginate /relations."""
```
reivilibre, line 282: parfaît

tests/rest/client/test_relations.py
```diff
@@ -278,16 +278,16 @@ def test_repeated_paginate_relations(self):
...
         channel = self._send_relation(RelationTypes.ANNOTATION, "m.reaction", "A")
         self.assertEquals(200, channel.code, channel.json_body)
         annotation_id = channel.json_body["event_id"]
         # Send an event after the relation events.
         self.helper.send(self.room, body="Latest event", tok=self.user_token)
 
-        # Request sync, but limit the timeline to get a pagination token before
-        # the reaction.
+        # Request /sync, limiting it such that only the latest event is returned
+        # (and not the relation).
```
reivilibre, line 290: Aha, that's better. I would (rightly or wrongly!) have thought this would have been 'after' the reaction, not 'before' as it said before. This clears that up.

tests/rest/client/test_relations.py
```diff
@@ -311,14 +313,17 @@ def test_pagination_from_sync_messages(self):
...
+        # Ensure the relation event is not in the chunk returned from /messages.
         self.assertNotIn(
             annotation_id, [ev["event_id"] for ev in channel.json_body["chunk"]]
         )
 
-        # Request the relations with the token from sync/messages -- this is a
-        # tiny bit silly since how do we know the parent ID? Just assume we had
-        # it from an earlier sync and the above are gappy.
+        # Request /relations with the pagination token from the /sync and /messages
+        # requests above.
```
reivilibre, line 322: This sounds like it suggests that the pagination token from `/sync` and `/messages` is identical.
If that's true, I would perhaps like it to be clarified (since the for loop makes the opposite suggestion). If it's not, I would tweak a bit
```suggestion
        # Request /relations with the pagination tokens from both the /sync and /messages
        # requests above, in turn.
```

clokep: I think they actually are (just due to the nature of the data in the unit test), but that's not important to the test.

Issue #11940, opened 2022-02-08
Suggest improvements to Twisted's release process


richvdh:
The Twisted team have recently made a couple of releases in quick succession, both addressing security issues as well as including other changes. This has been difficult for both us and our downstream packagers to work with.

The Twisted release manager would welcome suggestions for improvements for this process in the form of a PR to https://github.com/twisted/twisted/blob/trunk/docs/core/development/policy/release-process.rst. It would be great if we could find the time to do this.

reivilibre:
On the topic of library security vulnerability announcements, I seem to remember that I like OpenSSL's way of doing things where they give some advance notice before making a security release, along with the severity of the security release. Something like that would probably make sense for Twisted since it's a library — and it could be useful to mention a broad category (DoS? Information disclosure? etc) so that library users have a clue about what to expect.

clokep:
> (it might also be helpful if we could document our _own_ release cycle, if for no other reason than to document a way that works for us.)

I've started this at #11954.

adiroiban:
True. A pre-release would be nice.

If people are hanging at IRC or Gitter, we talk about the release and pre-release and the planning... so you somehow get that info :)


Another option for release docs: 
https://github.com/pyca/cryptography/blob/main/docs/doing-a-release.rst

callahad:
We'll take an hour and throw together a PR to Twisted's release process as a strawman, but ultimately the folks who _do_ the releases are the ones who have to accept and implement any changes, so we want to strongly defer to them rather than trying to arguing the specifics of any given change.

As mentioned in comments above, there is lots of great prior art. Django's is decent, too: https://docs.djangoproject.com/en/dev/internals/security/#how-django-discloses-security-issues

adiroiban:
Twisted PR at https://github.com/twisted/twisted/pull/1712 if you want to send your feedback. Thanks!

callahad:
Thank you for doing that, Adi! Will leave a comment over there, closing this issue here.

PR #11954, opened 2022-02-09
Document the release cycle

clokep:
Tries to formalize the conversation from #9914 that we've been following recently.

Also related to #11940.

### DIFF
--- a/docs/development/releases.md
+++ b/docs/development/releases.md
@@ -2,10 +2,10 @@
 
 New releases of Synapse nominally follow a two week release cycle:
 
-* Day 0: Synapse N - 1 is released.
-* Day 7: Synapse N release candidate 1 is released.
-* Days 7 - 13: Synapse N release candidates 2+ are released, if bugs are found.
-* Day 14: Synapse N is released.
+* Day 0: Synapse `N - 1` is released.
+* Day 7: Synapse `N` release candidate 1 is released.
+* Days 7 - 13: Synapse `N` release candidates 2+ are released, if bugs are found.
+* Day 14: Synapse `N` is released.
 
 Note that this schedule might be modified depending on the availability of the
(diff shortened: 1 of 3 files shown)


docs/development/releases.md
```diff
@@ -0,0 +1,33 @@
+# Synapse Release Cycle
+
+New releases of Synapse nominally follow a two week release cycle:
+
+* Day 0: Synapse N - 1 is released.
+* Day 7: Synapse N release candidate 1 is released.
+* Days 7 - 13: Synapse N release candidates 2+ are released, if bugs are found.
+* Day 14: Synapse N is released.
```
reivilibre, line 8: I always find it hard to format these semi-maths in this kind of text in a way where they're readable as maths and not weird punctuation; some emphasis with italics or code blocks would possibly help:

```suggestion
* Day 0: Synapse `N - 1` is released.
* Day 7: Synapse `N` release candidate 1 is released.
* Days 7 - 13: Synapse `N` release candidates 2+ are released, if bugs are found.
* Day 14: Synapse `N` is released.
```
(alternatively an example can do better here, even if it means hardcoding some numbers)

Is it worth mentioning that we aim for Tuesdays?

docs/development/releases.md
```diff
@@ -0,0 +1,33 @@
...
+
+Note that this schedule might be modified depending on the availability of the
+Synapse team, e.g. releases may be skipped to avoid holidays.
+
+## Bugfix releases
+
+If a bug is found after release that is deemed to severe enough (by a combination
```
reivilibre, line 15:
```suggestion
If a bug is found after release that is deemed severe enough (by a combination
```

reivilibre (approved):
Thanks for writing this up!

PR #10588, opened 2021-08-12
Stop building a debian package for Groovy Gorilla

DMRobertson:
Closes #10204

PR #11961, opened 2022-02-10
Drop support for EOL Ubuntu 21.04

### LINKED ISSUE #11525: Ubuntu 21.04 (Hirsute Hippo) reaches End of Life on January 20 2022
It will soon be time to remove it from the supported distros.

https://lists.ubuntu.com/archives/ubuntu-announce/2021-December/000275.html

H-Shay:
As the title states. Fixes #11525.


changelog.d/11961.removal
```diff
@@ -0,0 +1 @@
+Drop support for EOL Ubuntu 21.04.
```
clokep, line 1: I think we should probably be a bit more specific, stealing from #10588:

```suggestion
No longer build `.deb` packages for Ubuntu 21.04 Hirsute Hippo, which has now EOLed.
```

clokep (approved):

Issue #11457, opened 2021-11-30
When the max number of `to_device` messages for a sync is reached, the next sync does not return immediately


BillCarsonFr:
It appears that when you have reached the maximum number of to_device per sync, the next call to sync does not return immediatly with the remaining to_devices.
There must be new incoming data in order for the sync to trigger and return.

Expected:
If there are pending to_devices to deliver the sync should return immediatly instead of waiting for timeout or new data?

babolivier:
For extra context, we currently limit to 100 `to_device` messages in a sync. We return immediately (with the remaining messages) when we receive new data (i.e. data that's been created since the previous sync returned) but if we have leftover `to_device` messages in the database we don't count it as new data and don't return them until either new data comes in or we reach the 30s timeout.

If we have leftover `to_device` messages we couldn't fit in the previous sync, we definitely should return the following sync immediately with them.

bmarty:
We want the Matrix protocol to have good performance, and so the Matrix clients, so I would rise up the severity of this issue, S-Minor is not enough to me.
Some workaround have been implemented on some clients (JS and iOS), but not on Android. I will create an issue there to implement the workaround: set the timeout parameter of sync request to 0 until the number of toDevice event is 0.

richvdh:
It seems counterproductive for clients to have to spend time implementing workarounds for server-side bugs. @callahad: can we fit this in?

callahad:
Bumping severity + priority per discussion above. @DMRobertson may be able to look at this once he finishes up work on another sync issue.

BillCarsonFr:
Slightly off-topic, but currently clients will do a `catchup` sync until the `to_device` array is null. During this period client won't upload new otk (because their might be pre key olm message pending and we don't want to rotate keys to early).

The thing is that client might be doing that endlessly if there are always new to_device events.

I wonder if we don't need a `limited` field for to device..

erikjohnston:
I was only able to reproduce this in unit tests on SQLite. @BillCarsonFr were you experiencing this on servers using postgres?

erikjohnston:
So looks like the bug I fixed in #11966 is a recent regression and hasn't made it to a release yet. 

Can someone confirm if this issue is still present?

richvdh:
> Can someone confirm if this issue is still present?

given nobody is reporting this any more, I'm going to assume it is now resolved. We can reopen if anyone can still reproduce.

PR #11966, opened 2022-02-11
Fix to-device being dropped in limited sync in SQLite.

erikjohnston:
If ther are more than 100 to-device messages pending for a device
`/sync` will only return the first 100, however the next batch token was
incorrectly calculated and so all other pending messages would be
dropped.

This is due to `txn.rowcount` only returning the number of rows that
*changed*, rather than the number *selected* in SQLite.

c.f. #11457

### DIFF
--- a/changelog.d/11966.bugfix
+++ b/changelog.d/11966.bugfix
@@ -1,0 +1,1 @@
+Fix to-device messages being dropped during limited sync when using SQLite.
--- a/synapse/storage/databases/main/deviceinbox.py
+++ b/synapse/storage/databases/main/deviceinbox.py
@@ -362,7 +362,10 @@ class DeviceInboxWorkerStore(SQLBaseStore):
             # intended for each device.
             last_processed_stream_pos = to_stream_id
             recipient_device_to_messages: Dict[Tuple[str, str], List[JsonDict]] = {}
+            rowcount = 0
             for row in txn:
+                rowcount += 1
+
                 last_processed_stream_pos = row[0]
                 recipient_user_id = row[1]
                 recipient_device_id = row[2]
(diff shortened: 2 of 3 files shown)


reivilibre:
For reference:

> As required by the Python DB API Spec, the [rowcount](https://docs.python.org/3/library/sqlite3.html#sqlite3.Cursor.rowcount) attribute “is -1 in case no executeXX() has been performed on the cursor or the rowcount of the last operation is not determinable by the interface”. This includes SELECT statements because we cannot determine the number of rows a query produced until all rows were fetched.

[Python SQLite3 docs](https://docs.python.org/3/library/sqlite3.html#sqlite3.Cursor.rowcount)

Interesting to know that the Python DBAPI spec includes SELECTs in the 'rowcount'. In other languages, I've typically only seen inserts, updates and deletes (e.g. modifications) have a count.

However returning `-1` when it's not available is such a massive trap!!!

reivilibre (approved):
Looks good to me. Have you already checked if there are other cases where we rely on the `rowcount` attribute for selects?

erikjohnston:
> Looks good to me. Have you already checked if there are other cases where we rely on the `rowcount` attribute for selects?

Yeah I had a quick look and everywhere else seemed fine

richvdh:
did you check https://github.com/matrix-org/synapse/pull/11215/files#diff-f0d3815e503803dc732990d65d3d49bcdd08bb60d13c43bff0375cdc06b3538fR376 ?

richvdh:
oh wait. I see. 

I'll get my coat.

richvdh:
though I would say: please could the changelog say when the bug was introduced?

erikjohnston:
> though I would say: please could the changelog say when the bug was introduced?

Woops. /me does that now

erikjohnston:
Introduced by https://github.com/matrix-org/synapse/pull/11215, which I don't actually think is in a release yet??

Issue #11249, opened 2021-11-04
Prototype MSC2775: Lazy loading over federation


callahad:
Large, complex rooms with many servers can take many minutes to join over federation.

One possible option for addressing this is described in [MSC2775: Lazy loading over federation](https://github.com/matrix-org/matrix-doc/pull/2775).

We should prototype it in Synapse to determine its viability.

richvdh:
I am tracking progress on this work at https://hackmd.io/R20t3MTaSO6NvfPQh978QA.

richvdh:
I'm closing this issue in favour of a milestone, https://github.com/matrix-org/synapse/milestone/6.

PR #11967, opened 2022-02-11
Implement MSC3706: partial state in `/send_join` response

richvdh:
Part of my work on #11249: this lands the changes needed on the "server" side of federation to reduce the response size of the `/send_join` response.

There are a few changes. First, we add a query-param allowing federation "clients" to opt in to the new behaviour.

If a client opts in, we then:
 * reduce the number if state events we return
 * only return the auth events needed to auth those state events, and only those not duplicated by the state list.
 * return a list of servers currently in the room.

It's all guarded by an experimental config flag.

Suggest reviewing commit-by-commit as there are a couple of preparatory commits.

### DIFF
--- a/synapse/federation/federation_server.py
+++ b/synapse/federation/federation_server.py
@@ -571,7 +571,7 @@ class FederationServer(FederationBase):
     ) -> JsonDict:
         state_ids = await self.handler.get_state_ids_for_pdu(room_id, event_id)
         auth_chain_ids = await self.store.get_auth_chain_ids(room_id, state_ids)
-        return {"pdu_ids": state_ids, "auth_chain_ids": auth_chain_ids}
+        return {"pdu_ids": state_ids, "auth_chain_ids": list(auth_chain_ids)}
 
     async def _on_context_state_request_compute(
         self, room_id: str, event_id: Optional[str]
--- a/synapse/federation/federation_server.py
+++ b/synapse/federation/federation_server.py
@@ -652,19 +652,27 @@ class FederationServer(FederationBase):
         )
 
         prev_state_ids = await context.get_prev_state_ids()
-        state_ids = list(prev_state_ids.values())
-        auth_chain = await self.store.get_auth_chain(room_id, state_ids)
-        state = await self.store.get_events(state_ids)
+        state_event_ids = prev_state_ids.values()
 
+        auth_chain_event_ids = await self.store.get_auth_chain_ids(
+            room_id, state_event_ids
+        )
+
+        auth_chain_events = await self.store.get_events_as_list(auth_chain_event_ids)
+        state_events = await self.store.get_events_as_list(state_event_ids)
+
+        # we try to do all the async stuff before this point, so that time_now is as
+        # accurate as possible.
         time_now = self._clock.time_msec()
-        event_json = event.get_pdu_json()
-        return {
+        event_json = event.get_pdu_json(time_now)
+        resp = {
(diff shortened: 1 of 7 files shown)


synapse/federation/federation_server.py
```diff
@@ -652,19 +652,27 @@ async def on_send_join_request(
...
-        auth_chain = await self.store.get_auth_chain(room_id, state_ids)
-        state = await self.store.get_events(state_ids)
+        state_event_ids = prev_state_ids.values()
 
+        auth_chain_event_ids = await self.store.get_auth_chain_ids(
+            room_id, state_event_ids
+        )
+
+        auth_chain_events = await self.store.get_events_as_list(auth_chain_event_ids)
+        state_events = await self.store.get_events_as_list(state_event_ids)
```
richvdh, line 684: `get_events_as_list` is equivalent to `get_events`, except that it doesn't build a dict of the results. Since we only ever used the `.values` of the result, the dict-building was entirely redundant. (The only thing it could plausibly have done was dedup the results, but since `state_event_ids` is a set anyway, there couldn't have been any dups in the first place)

synapse/federation/federation_server.py
```diff
@@ -1347,3 +1374,39 @@ async def on_query(self, query_type: str, args: dict) -> JsonDict:
...
+    # be an auth event for the new join, so we may as well return it)
+    current_membership_event_id = prev_state_ids.get(
+        (EventTypes.Member, join_event.state_key)
+    )
+    if current_membership_event_id is not None:
+        state_event_ids.add(current_membership_event_id)
+
+    # TODO: return a few more members:
+    #   - those with invites
+    #   - those that are kicked? / banned
```
reivilibre, line 1410: Interested in more rationale/elaboration here, if you have any more thoughts.
I guess I can see invites and bans because those are useful for validating other users' join events later.
However, due to e.g. annoying spammers, kicks and bans could possibly be a large amount of the state events for some rooms.

reivilibre (approved):
Looks good! I can see what's going on. :)

synapse/federation/federation_server.py
```diff
@@ -645,19 +646,40 @@ async def on_invite_request(
...
     ) -> Dict[str, Any]:
         event, context = await self._on_send_membership_event(
             origin, content, Membership.JOIN, room_id
         )
 
         prev_state_ids = await context.get_prev_state_ids()
-        state_event_ids = prev_state_ids.values()
+
+        state_event_ids: Collection[str]
+        servers_in_room: Optional[Collection[str]]
```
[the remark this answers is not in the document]
richvdh: > I guess also because the joining homeserver might need to send events and for that, it needs to know which servers are in the room 

yeah, this is the main driver right now.

synapse/federation/federation_server.py
richvdh: well, MSC2775 lists a whole bunch of members we might want to return. I'm unconvinced by some of them, so this is here just as a thing to return to.

synapse/federation/transport/server/federation.py
```diff
@@ -422,7 +432,15 @@ async def on_PUT(
     ) -> Tuple[int, JsonDict]:
         # TODO(paul): assert that event_id parsed from path actually
```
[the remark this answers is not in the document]
richvdh: It's @leonerd  :). He did some contracting for us back in the pre-NVL days. Which tells us how long this particular TODO has gone undone :'(

PR #11971, opened 2022-02-11
Prepare for rename of default complement branch

richvdh:
use `HEAD` rather than hardcoding `master`


clokep (approved):
Seems fine if CI is happy.

PR #11972, opened 2022-02-11
Add some tests for propagation of device list changes between local users

anoadragon453:
Born out of trying to test a particular condition in https://github.com/matrix-org/synapse/pull/11905.

There's some duplication between these methods, though not so much that I think refactoring it out is sorely needed. Feel free to refute that claim though!

### DIFF
--- a/tests/rest/client/test_device_lists.py
+++ b/tests/rest/client/test_device_lists.py
@@ -1,0 +1,167 @@
+# Copyright 2022 The Matrix.org Foundation C.I.C.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from unittest.mock import Mock
+
+from synapse.rest import admin, devices, room, sync
+from synapse.rest.client import account, login, register
+
+from tests import unittest
+
+
+class DeviceListsTestCase(unittest.HomeserverTestCase):
+    """Tests regarding device list changes."""
+
+    servlets = [
+        admin.register_servlets_for_client_rest_resource,
+        login.register_servlets,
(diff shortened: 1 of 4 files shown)


tests/rest/client/test_device_lists.py
```diff
@@ -0,0 +1,167 @@
...
+        self.assertIsNotNone(new_room_id)
+
+        # Pretend that Bob joined this room on a previous homeserver startup.
+        #
+        # There is a quirk; if Bob has joined the room since the last homeserver
+        # restart, then an additional mechanism is employed to ensure Bob will start
+        # receiving device list updates for that room.
+        # In order to test the case where Bob has previously been joined to the room,
+        # instead of restarting Synapse, we can just stub out the notifier machinery
+        # that is triggered when a user joins a room.
```
erikjohnston, line 62: I'm fairly dubious about doing this, as it really is wrenching into the guts of the code and changing things (rather than just mocking out an interface). Why do we need this? My reading of the code is that `_user_joined_room` will only have an effect if the user is currently syncing when they joined the room?

erikjohnston (commented):
Otherwise LGTM

tests/rest/client/test_device_lists.py
anoadragon453: Hmm, a user stream entry is created when a user syncs, but I fail to see where it's removed at the end of a sync. I only see streams being removed when they expire: https://github.com/matrix-org/synapse/blob/5de46ac757b8e924a7c29afcfe041d95430b8724/synapse/notifier.py#L701

Nevertheless, putting in a breakpoint [here](https://github.com/matrix-org/synapse/blob/5de46ac757b8e924a7c29afcfe041d95430b8724/synapse/notifier.py#L713) resulted in no trigger when running the full test, so this is indeed not necessary! Removed in edc28b044.

erikjohnston: It happens here: https://github.com/matrix-org/synapse/blob/5de46ac757b8e924a7c29afcfe041d95430b8724/synapse/notifier.py#L155-L164

(I spent a while tracking that down)

erikjohnston (approved):

tests/rest/client/test_device_lists.py
anoadragon453: Right, though the only code I can see that calls this is `remove_expired_streams`.

erikjohnston: Oh, that seems odd that it doesn't get removed when the stream gets notified :shrug:

PR #11974, opened 2022-02-11
Optimise calculating device_list changes in `/sync`. 

erikjohnston:
For users with large accounts it is inefficient to calculate the set of
users they share a room with (and takes a lot of space in the cache).
Instead we can look at users whose devices have changed since the last
sync and check if they share a room with the syncing user.

This should fix the synchrotron with pinned CPU on matrix.org.


synapse/storage/databases/main/roommember.py
```diff
@@ -504,6 +504,63 @@ def _get_rooms_for_user_with_stream_ordering_txn(
             for room_id, instance, stream_id in txn
         )
 
+    @cachedList(
+        cached_method_name="get_rooms_for_user_with_stream_ordering",
+        list_name="user_ids",
+    )
+    async def get_rooms_for_users_with_stream_ordering(
```
reivilibre, line 511: may want a little docstring here, mostly for the return type, which is not very obvious.

It's roughly `user → (room, event_pos)` to my reading. What does the `event_pos` mean here?

… seems to be the position of the event that gives the user their membership in the room?

synapse/storage/databases/main/roommember.py
```diff
@@ -504,6 +504,63 @@ def _get_rooms_for_user_with_stream_ordering_txn(
...
+        self, txn, user_ids: Collection[str]
+    ) -> Dict[str, FrozenSet[GetRoomsForUserWithStreamOrdering]]:
+
+        clause, args = make_in_list_sql_clause(
+            self.database_engine,
+            "c.state_key",
+            user_ids,
+        )
+
+        if self._current_state_events_membership_up_to_date:
```
reivilibre, line 530: Are we stuck with this thing until the rest of time? I'm guessing it's some kind of optimisation we came up with that got populated in the background?

synapse/storage/databases/main/roommember.py
reivilibre: Do we have no need to invalidate this if a user leaves or so on?

synapse/handlers/sync.py
```diff
@@ -1289,23 +1289,55 @@ async def _generate_sync_entry_for_device_list(
...
+                    changed_users
+                )
+
+                for changed_user_id, entries in result.items():
+                    # Check if the changed user shares any rooms with the user,
+                    # or if the changed user is the syncing user (as we always
+                    # want to include device list updates of their own devices).
+                    if (
+                        any(e.room_id in joined_rooms for e in entries)
+                        or user_id == changed_user_id
```
reivilibre, line 1320: I'd probably write it this way so it short-circuits without iterating the set when it's the same user...
```suggestion
                        user_id == changed_user_id
                        or any(e.room_id in joined_rooms for e in entries)
```

synapse/handlers/sync.py
```diff
@@ -1289,23 +1289,55 @@ async def _generate_sync_entry_for_device_list(
...
+            # the last sync then it's likely more efficient to compare the rooms
+            # they're in with the rooms the syncing user is in.
+            #
+            # If we don't have that info cached then we get all the users that
+            # share a room with our user and check if those users have changed.
+            changed_users = self.store.get_cached_device_list_changes(
+                since_token.device_list_key
             )
+            if changed_users is not None:
+                result = await self.store.get_rooms_for_users_with_stream_ordering(
```
reivilibre, line 1310: What's the purpose of the stream ordering? I've stared at this a bit and can't see where it's used...?

reivilibre (changes requested):

synapse/storage/databases/main/roommember.py
```diff
@@ -504,6 +504,63 @@ def _get_rooms_for_user_with_stream_ordering_txn(
...
+                FROM current_state_events AS c
+                INNER JOIN room_memberships AS m USING (room_id, event_id)
+                INNER JOIN events AS e USING (room_id, event_id)
+                WHERE
+                    c.type = 'm.room.member'
+                    AND m.membership = ?
+                    AND {clause}
+            """
+
+        txn.execute(sql, [Membership.JOIN] + list(args))
```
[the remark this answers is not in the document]
erikjohnston: Oh, hmm. We do the clone in a bunch of places but I'm not sure why. For some reason I thought it could return a tuple

synapse/storage/databases/main/roommember.py
erikjohnston: Ah, missed that this didn't have a docstring.

I think you're missing that this is using a `@cachedList`, which means that its simply a batch version of `get_rooms_for_user_with_stream_ordering` and reuses the same cache. Hence why it includes the stream ordering and the like. Also means that invalidation is already taken care of.

synapse/handlers/sync.py
erikjohnston: As per https://github.com/matrix-org/synapse/pull/11974#discussion_r804846713, since we reuse another cache we need to pull out the stream ordering. We could have another function that strips out the stream ordering (which is what `get_user_in_room` does for the non-batch version), but that feels inefficient

synapse/storage/databases/main/roommember.py
reivilibre: Oooh, I see, very cute. That makes sense!

reivilibre (approved):

synapse/storage/databases/main/roommember.py
erikjohnston: At some point I think we can change it from a background update to happen synchronously? Or require that people upgrade from vX -> vY and then vY -> vZ? I think we've done something along these lines before?

PR #11975, opened 2022-02-11
Require that modules register their callbacks using keyword arguments

clokep:
From my suggestion at https://github.com/matrix-org/synapse/pull/11950#discussion_r803013130

This should give us more flexibility to add/remove module callbacks since they must be named as keyword arguments.

This isn't 100% backwards compatible though, unfortunately, if someone *was* not using kwargs.

I poked around at all the modules I could find that we've created and they seem to use keyword arguments.

### DIFF
--- a/changelog.d/11975.bugfix
+++ b/changelog.d/11975.bugfix
@@ -1,0 +1,1 @@
+Require that modules register their callbacks using keyword arguments.
--- a/synapse/module_api/__init__.py
+++ b/synapse/module_api/__init__.py
@@ -211,6 +211,7 @@ class ModuleApi:
 
     def register_spam_checker_callbacks(
         self,
+        *,
         check_event_for_spam: Optional[CHECK_EVENT_FOR_SPAM_CALLBACK] = None,
         user_may_join_room: Optional[USER_MAY_JOIN_ROOM_CALLBACK] = None,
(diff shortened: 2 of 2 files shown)


babolivier:
@clokep Would you be happy with updating the other `register_[...]_callbacks` functions the same way in this PR? I'm a bit uneasy with the restriction only applying to one of them.

babolivier (commented):
I think we should add a note to https://matrix-org.github.io/synapse/latest/modules/writing_a_module.html#registering-a-callback to explicitly call out that arguments must be kwargs (even though it's sort of implied) so module developers don't try to wonder why their modules don't work.

clokep:
> I think we should add a note to https://matrix-org.github.io/synapse/latest/modules/writing_a_module.html#registering-a-callback to explicitly call out that arguments must be kwargs (even though it's sort of implied) so module developers don't try to wonder why their modules don't work.

It seems to already say this (emphasis mine):

> Modules can register callbacks using one of the module API's `register_[...]_callbacks` methods. The *callback functions are passed to these methods as keyword arguments*, with the callback name as the argument name and the function as its value.

I think this is pretty clear, although it doesn't quite say "must". Do you have any suggestions to this text?

babolivier (approved):
Oh looks like I've read that bit too quickly. Yay for consistency with the doc then :stuck_out_tongue:

PR #11980, opened 2022-02-11
Add config settings for background update parameters

H-Shay:
This pull request adds functionality to allows users to set their own values for database background update parameters relating to the minimum background batch size, the default background batch size, the duration of a batch of background updates, whether or not to sleep between batches and if so, how long to sleep for. This should have the effect of allowing users to speed up or slow down background updates.


docs/sample_config.yaml
```diff
@@ -2729,3 +2729,36 @@ redis:
...
+
+# Minimum size a batch of background updates can be. Must be greater than 0. Defaults to 1. Uncomment and
+# set a size to change the default.
+#
+#min_batch_size: 10
+
+# The batch size to use for the first iteration of a new background update. The default is 100.
+# Uncomment and set a size to change the default.
+#
+#default_batch_size: 50
```
erikjohnston, line 2764: I think this would probably be best grouped under a key? e.g.;

```yaml
background_updates:
    min_batch_size: 10
    default_batch_size: 50
```

H-Shay: Great suggestion, I have implemented it.

synapse/config/background_updates.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+class BackgroundUpdateConfig(Config):
+    section = "background_updates"
+
+    def generate_config_section(self, **kwargs) -> str:
+        return """\
+
+
+        ## Background Updates ##
```
erikjohnston, line 25:
```suggestion
        ## Background Updates ##
```

According to the style guide there is meant to only be two blank lines before a new config section https://matrix-org.github.io/synapse/latest/code_style.html#configuration-file-format

tests/config/test_background_update.py
```diff
@@ -0,0 +1,58 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import yaml
+
+from synapse.storage.background_updates import BackgroundUpdater
+
+from tests.unittest import HomeserverTestCase, override_config
+
+
+class RoomDirectoryConfigTestCase(HomeserverTestCase):
```
erikjohnston, line 21:
```suggestion
class BackgroundUpdateConfigTestCase(HomeserverTestCase):
```

erikjohnston (approved):
Awesome, thanks shay :)

PR #11992, opened 2022-02-14
Fix a bug with thread summaries when the latest event is edited

### LINKED ISSUE #11973: Thread summary does not take into account edited events
If the latest event in a thread is edited, the edit is not taken into account when calculating the thread summary.

This would happen if you have:

1. A root event (`A`).
2. A threaded event (`B`).
3. An edit to `B` (`C`).

The summary shows `B` as the latest event instead of `C`.

clokep:
If the latest event in a thread is edited the edit is not reflected properly in the bundled aggregation. This PR fixes that behavior and includes a regression test.

Fixes #11973

### DIFF
--- a/synapse/storage/databases/main/relations.py
+++ b/synapse/storage/databases/main/relations.py
@@ -53,9 +53,13 @@ logger = logging.getLogger(__name__)
 
 @attr.s(slots=True, frozen=True, auto_attribs=True)
 class _ThreadAggregation:
+    # The latest event in the thread.
     latest_event: EventBase
+    # The latest edit to the latest event in the thread.
     latest_edit: Optional[EventBase]
+    # The total number of events in the thread.
     count: int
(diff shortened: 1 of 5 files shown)


synapse/storage/databases/main/relations.py
```diff
@@ -461,7 +462,7 @@ def get_thread_summary(self, event_id: str) -> Optional[Tuple[int, EventBase]]:
     @cachedList(cached_method_name="get_thread_summary", list_name="event_ids")
     async def _get_thread_summaries(
         self, event_ids: Collection[str]
-    ) -> Dict[str, Optional[Tuple[int, EventBase]]]:
+    ) -> Dict[str, Optional[Tuple[int, EventBase, Optional[EventBase]]]]:
```
erikjohnston, line 465: Can update docstring pls?

erikjohnston (approved):

PR #11994, opened 2022-02-15
`_auth_and_persist_outliers`: drop events we have already seen

richvdh:
`_auth_and_persist_outliers` is called in three places:
 * `process_remote_join`
 * `_get_remote_auth_chain_for_event`
 * `_get_events_and_persist`.

The first two of those contain code to filter out any events we have already seen, with a call to `have_seen_events`. (The third, `_get_events_and_persist`, doesn't bother, because in theory it is only called for events we don't already have).

I'm shortly going to add a fourth caller, which will also need to filter out events we have already seen. Rather than add yet a third copy of the deduplication code, let's just push it down into `auth_and_persist_outliers`.

### DIFF
--- a/changelog.d/11994.misc
+++ b/changelog.d/11994.misc
@@ -1,0 +1,1 @@
+Move common deduplication code down into `_auth_and_persist_outliers`.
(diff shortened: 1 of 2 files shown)


squahtx (approved):
lgtm as a refactoring!

mypy is complaining about the type of remote_events

clokep (approved):
Seems reasonable.

Issue #11995, opened 2022-02-15
Mainline matrix-org-hotfixes changes to `RoomMemberHandler.update_membership`


babolivier:
Specifically this bit:

https://github.com/matrix-org/synapse/blob/6705391eec909d3af0d6d1931fd474fc72bb0952/synapse/handlers/room_member.py#L486-L522

It's been causing merge conflicts when merging release branches into `matrix-org-hotfixes` a few times lately due to new releases adding kwargs to `update_membership_locked`, e.g.:

```diff
diff --cc synapse/handlers/room_member.py
index 8ee1a6470,bf1a47efb..000000000
--- a/synapse/handlers/room_member.py
+++ b/synapse/handlers/room_member.py
@@@ -483,43 -500,25 +501,65 @@@ class RoomMemberHandler(metaclass=abc.A
  
          key = (room_id,)
  
++<<<<<<< HEAD
 +        as_id = object()
 +        if requester.app_service:
 +            as_id = requester.app_service.id
 +
 +        then = self.clock.time_msec()
 +
 +        with (await self.member_limiter.queue(as_id)):
 +            diff = self.clock.time_msec() - then
 +
 +            if diff > 80 * 1000:
 +                # haproxy would have timed the request out anyway...
 +                raise SynapseError(504, "took to long to process")
 +
 +            with (await self.member_linearizer.queue(key)):
 +                diff = self.clock.time_msec() - then
 +
 +                if diff > 80 * 1000:
 +                    # haproxy would have timed the request out anyway...
 +                    raise SynapseError(504, "took to long to process")
 +
 +                result = await self.update_membership_locked(
 +                    requester,
 +                    target,
 +                    room_id,
 +                    action,
 +                    txn_id=txn_id,
 +                    remote_room_hosts=remote_room_hosts,
 +                    third_party_signed=third_party_signed,
 +                    ratelimit=ratelimit,
 +                    content=content,
 +                    new_room=new_room,
 +                    require_consent=require_consent,
 +                    outlier=outlier,
 +                    historical=historical,
 +                    prev_event_ids=prev_event_ids,
 +                    auth_event_ids=auth_event_ids,
 +                )
++=======
+         with (await self.member_linearizer.queue(key)):
+             result = await self.update_membership_locked(
+                 requester,
+                 target,
+                 room_id,
+                 action,
+                 txn_id=txn_id,
+                 remote_room_hosts=remote_room_hosts,
+                 third_party_signed=third_party_signed,
[...]

PR #11996, opened 2022-02-15
Limit concurrent AS joins

babolivier:
Initially introduced in `matrix-org-hotfixes` by https://github.com/matrix-org/synapse/commit/e5537cf98333b428dbc481ed443daed2f0cfa074 (and tweaked by later commits).

Fixes https://github.com/matrix-org/synapse/issues/11995

See also https://github.com/matrix-org/synapse/issues/4826

I haven't ported https://github.com/matrix-org/synapse/commit/ca21957b8a4942e829050834297afb290338c120 as part of this PR because it looks like something that should be done separately and would need a switch from hardcoded settings to configuration knobs.

### DIFF
--- a/changelog.d/11996.misc
+++ b/changelog.d/11996.misc
@@ -1,0 +1,1 @@
+Limit concurrent applications service joins.
--- a/changelog.d/11996.misc
+++ b/changelog.d/11996.misc
@@ -1,1 +1,1 @@
-Limit concurrent applications service joins.
+Limit concurrent joins from applications services.
--- a/synapse/handlers/room_member.py
+++ b/synapse/handlers/room_member.py
@@ -82,6 +82,7 @@ class RoomMemberHandler(metaclass=abc.ABCMeta):
         self.event_auth_handler = hs.get_event_auth_handler()
 
         self.member_linearizer: Linearizer = Linearizer(name="member")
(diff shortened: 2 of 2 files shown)


changelog.d/11996.misc
```diff
@@ -0,0 +1 @@
+Limit concurrent applications service joins.
```
clokep, line 1: Maybe this is clearer?

```suggestion
Limit concurrent joins from applications services.
```

synapse/handlers/room_member.py
```diff
@@ -500,25 +501,30 @@ async def update_membership(
...
-                historical=historical,
-                allow_no_prev_events=allow_no_prev_events,
-                prev_event_ids=prev_event_ids,
-                auth_event_ids=auth_event_ids,
-            )
+        as_id = object()
+        if requester.app_service:
+            as_id = requester.app_service.id
+
+        with (await self.member_as_limiter.queue(as_id)):
```
clokep, line 508: I feel like this is confusing enough that it needs a comment above it:

> Linearize first by the application service, then the room ID.

I'm not sure if that's helpful or not though.

clokep (approved):
Seems OK.

clokep:
Test failure is matrix-org/sytest#1044.

PR #2846, opened 2018-02-03
Add a Dockerfile for synapse

kaiyou:
*This is currently not ready for merge, the pull request is meant for tracking the progress on the branch.*

Changes
- [x] Dockerfile building properly
- [x] Basic Python start script
- [x] Template homeserver.yaml and log configuration
- [ ] Generate keys upon startup
- [ ] Handle shared secrets for TURN, registration, Macaroon
- [ ] Provide an example Docker Compose file with a separate TURN server

Issues
- [ ] SQL schema files are not copied properly when building

PR #11997, opened 2022-02-15
Docker: remove `VOLUME` directive

richvdh:
The driver for this is to stop Complement complaining about it, but as far as I can tell it was pointless and needed to go away anyway.

I'm a *bit* unclear about what exactly `VOLUME` does, but I *think* what it means is that, if you don't override it with an explicit `-v` argument, then `docker run` will create a temporary volume, and copy things into it. The temporary volume is then deleted when the container finishes.

That only sounds useful if your image has something to copy into it (otherwise you may as well just use the default root filesystem), and our image notably doesn't copy anything into `/data`.

So... this wasn't doing anything, except annoying Complement?

(It was introduced in https://github.com/matrix-org/synapse/pull/2846, in case that helps.)


richvdh:
cf https://docs.docker.com/engine/reference/builder/#volume

Issue #8428, opened 2020-09-30
Summary of performance impact of running on resource constrained devices such as SBCs


emorrp1:
### Description

I've been running my homeserver on a cubietruck at home now for some time and am often replying to statements like "you need loads of ram to join large rooms" with "it works fine for me". I thought it might be useful to curate a summary of the issues you're likely to run into to help as a scaling-down guide, maybe highlight these for development work or end up as documentation.

#### Presence

This is the main reason people have a poor matrix experience on resource constrained homeservers. Element web will frequently be saying the server is offline while the python process will be pegged at 100% cpu. This feature is used to tell when other users are active (have a client app in the foreground) and therefore more likely to respond, but requires a lot of network activity to maintain even when nobody is talking in a room.

![Screenshot_2020-10-01_19-29-46](https://user-images.githubusercontent.com/71895/94848963-a47a3580-041c-11eb-8b6e-acb772b4259e.png)

While synapse does have some performance issues with presence #3971, the fundamental problem is that this is an easy feature to implement for a centralised service at nearly no overhead, but federation makes it combinatorial #8055. There is also a client-side config option which disables the UI and idle tracking [enable_presence_by_hs_url] to blacklist the largest instances but I didn't notice much difference, so I recommend disabling the feature entirely at the server level as well.

[enable_presence_by_hs_url]: https://github.com/vector-im/element-web/blob/v1.7.8/config.sample.json#L45

#### Joining

Joining a "large", federated room will initially fail with the below message in Element web, but waiting a while (10-60mins) and trying again will succeed without any issue. What counts as "large" is not message history, user count, connections to homeservers or even a simple count of the state events, it is instead how long the state resolution algorithm takes. However, each of those numbers are reasonable proxies, so we can use them as estimates since user count is one of the few things you see before joining.

![Screenshot_2020-10-02_17-15-06](https://user-images.githubusercontent.com/71895/94945781-18771500-04d3-11eb-8419-83c2da73a341.png)

[...]

PR #11998, opened 2022-02-15
Enable complexity checking in complexity checking docs example

anoadragon453:
@youphyun noticed in https://github.com/matrix-org/synapse/issues/8428#issuecomment-1040223558 that the recommended config for complexity checking on single board computers (SBCs) did not actually enable complexity checking.

This PR uncomments the `enabled: true` line in order to actually enable it.

Complexity checking is disabled by default: https://github.com/matrix-org/synapse/blob/119edf51eb3e4f5ed5139dc370f5d7aed46edc1c/synapse/config/server.py#L231-L232

### DIFF
--- a/docs/other/running_synapse_on_single_board_computers.md
+++ b/docs/other/running_synapse_on_single_board_computers.md
@@ -38,8 +38,7 @@ use_presence: false
 # joins a new remote room. If it is above the complexity limit, the server will
 # disallow joining, or will instantly leave.
 limit_remote_rooms:
-  # Uncomment to enable room complexity checking.
-  #enabled: true
+  enabled: true
   complexity: 3.0
 
 # Database configuration


docs/other/running_synapse_on_single_board_computers.md
```diff
@@ -38,8 +38,7 @@ use_presence: false
 # joins a new remote room. If it is above the complexity limit, the server will
 # disallow joining, or will instantly leave.
```
clokep, line 39: The above sounds like you may or may not want to enable it. I think we should reword this to be clearer:

> Disable room "complexity" checking before a user joins a new remote room.
> If it is above the complexity limit, the server will disallow joining, or will
> instantly leave.

clokep: Ahhh, well that's confusing that it is copied from the sample config when bits and pieces of the values are modified. That's probably why it doesn't make sense.

I think it would be clearer, yes. We're telling people to paste this into their config, so it should describe what that change actually does IMO?

anoadragon453: I've made the comments more relevant in c30b24717. Hopefully that's better?

docs/other/running_synapse_on_single_board_computers.md
```diff
@@ -31,28 +31,29 @@ Anything that requires modifying the device list [#7721](https://github.com/matr
...
   name: psycopg2
   args:
     user: matrix-synapse
-    # Generate a long, secure one with a password manager
+    # Generate a long, secure password using a password manager
     password: hunter2
     database: matrix-synapse
     host: localhost
-    cp_min: 5
-    cp_max: 10
```
clokep, line 55: Why this change?

clokep (approved):

docs/other/running_synapse_on_single_board_computers.md
anoadragon453: Those are the default values and don't seem to be relevant to SBCs. Rather, I think they were just copy-pasted over.

clokep: Makes sense. Thanks!

PR #11999, opened 2022-02-15
Fix incorrect `get_rooms_for_user` for remote user

erikjohnston:
When the server leaves a room the `get_rooms_for_user` cache is not
correctly invalidated for the remote users in the room. This means that
subsequent calls to `get_rooms_for_user` for the remote users would
incorrectly include the room (it shouldn't be included because the
server no longer knows anything about the room).

### DIFF
--- a/synapse/storage/databases/main/events.py
+++ b/synapse/storage/databases/main/events.py
@@ -975,6 +975,17 @@ class PersistEventsStore:
             to_delete = delta_state.to_delete
             to_insert = delta_state.to_insert
 
+            # Figure out the changes of membership to invalidate the
+            # `get_rooms_for_user` cache.
+            # We find out which membership events we may have deleted
+            # and which we have added, then we invalidate the caches for all
+            # those users.
+            members_changed = {
+                state_key
+                for ev_type, state_key in itertools.chain(to_delete, to_insert)
+                if ev_type == EventTypes.Member
+            }
+
             if delta_state.no_longer_in_room:
                 # Server is no longer in the room so we delete the room from
                 # current_state_events, being careful we've already updated the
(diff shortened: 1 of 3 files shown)


tests/storage/test_events.py
```diff
@@ -329,3 +329,63 @@ def test_do_not_prune_gap_if_not_dummy(self):
...
+
+    def prepare(self, reactor, clock, homeserver):
+        self.state = self.hs.get_state_handler()
+        self.persistence = self.hs.get_storage().persistence
+        self.store = self.hs.get_datastore()
+
+    def test_remote_user_cache_invalidated(self):
+        """Test that if the server leaves a room the `get_users_in_room` cache
+        is invalidated for remote users.
+        """
```
clokep, line 349: I think this is meant to be:

```suggestion
        """Test that if the server leaves a room the `get_rooms_for_user` cache
        is invalidated for remote users.
        """
```

Although I wonder if we should check the `get_users_for_room` one too while we're here?

synapse/storage/databases/main/events.py
```diff
@@ -993,6 +1004,11 @@ def _update_current_state_txn(
                 """
                 txn.execute(sql, (stream_id, self._instance_name, room_id))
 
+                # We also want to invalidate the membership caches for remote
+                # users that were in the room.
+                users_in_room = self.store.get_users_in_room_txn(txn, room_id)
+                members_changed.update(users_in_room)
```
clokep, line 1010: I might just be me, but I'm finding the remote-ness discussion confusing.

I think what we want to do is:

1. If the server has left the room, `members_changed` should be all users in the room (so `get_rooms_for_user` gets invalidated for everyone).
2. If the server is still in the room, `members_changed` should be calculated from the state changes to the room.

Is that accurate or no?

erikjohnston: Yup, that is correct. The thing is that `to_delete` will already include all local users in the room (as otherwise the server won't have "left" the room). So the impact of this bug is for remote users only, but I'll update the comment to clarify

tests/storage/test_events.py
erikjohnston: > Although I wonder if we should check the get_users_for_room one too while we're here?

I've written a test for it, and looks like it was being correctly invalidated.

clokep (approved):

PR #12000, opened 2022-02-15
Track cache invalidations

erikjohnston:
Currently we only track evictions due to size or time constraints.


clokep (approved):
Pretty sure this is OK, I guess we'll need to update the dashboard at some point too?

erikjohnston:
> Pretty sure this is OK, I guess we'll need to update the dashboard at some point too?

It should just magically work, we already display the reason

PR #10758, opened 2021-09-03
Improve the modules doc

babolivier:
* Split up the documentation in several files rather than one huge one
* Add examples for each callback category
* Other niceties like fixing https://github.com/matrix-org/synapse/issues/10632
* Add titles to callbacks so they're easier to find in the navigation panels and link to

Can be reviewed commit by commit.

PR #12003, opened 2022-02-15
Explain the meaning of spam checker callbacks' return values

DMRobertson:
#10758 documented various module API callbacks which must return a `bool`. It did _not_ specify the meaning of those bools, leaving module authors to either inspect Synapse's source, or else guess the meaning of the return value based on the function's name.

In this change I have tried to explicitly spell out what each return value indicates. Please double-check this carefully: I did this in a hurry, and am not familiar with the module API.

Motivation: Most callbacks return `False` to prevent some kind of action from taking place. Unfortunately `check_username_for_spam` is the opposite (return `True` to exclude from user directory searches; `False` for no effect). This isn't obvious. As a result, for about a week or two we ran a module on matrix.org [which effectively considered all users spammy](https://github.com/matrix-org/message_limit/commit/ac7d84b1143d24460aa342009601213a4808f5fc#diff-107ab1be23f8882205b0f170ecb9893068bebe844b8df5199fa263fa7a8b2b61R34-R35).

### DIFF
--- a/docs/modules/spam_checker_callbacks.md
+++ b/docs/modules/spam_checker_callbacks.md
@@ -50,7 +50,7 @@ context of a room creation.
 
 If multiple modules implement this callback, they will be considered in order. If a
 callback returns `True`, Synapse falls through to the next one. The value of the first
-callback that does not return `True` will be used. If this happens, Synapse will not call 
+callback that does not return `True` will be used. If this happens, Synapse will not call
 any of the subsequent implementations of this callback.
 
 ### `user_may_invite`
@@ -64,7 +64,7 @@ async def user_may_invite(inviter: str, invitee: str, room_id: str) -> bool
 Called when processing an invitation. The module must return a `bool` indicating whether
 the inviter can invite the invitee to the given room. Both inviter and invitee are
 represented by their Matrix user ID (e.g. `@alice:example.com`). Return `False` to prevent
-the invitation; otherwise return `True` to permit it;
+the invitation; otherwise return `True` to permit it.
 
 If multiple modules implement this callback, they will be considered in order. If a
 callback returns `True`, Synapse falls through to the next one. The value of the first
@@ -177,7 +177,7 @@ async def check_username_for_spam(user_profile: Dict[str, str]) -> bool
 
 Called when computing search results in the user directory. The module must return a
 `bool` indicating whether the given user profile can appear in search results. Return
-`True` to indicate that the user is spammy and exclude them user from search results;
+`True` to indicate that the user is spammy and exclude them from search results;
 otherwise return `False`.
 
 The profile is represented as a dictionary with the following keys:
--- a/docs/modules/spam_checker_callbacks.md
+++ b/docs/modules/spam_checker_callbacks.md
@@ -16,10 +16,12 @@ _First introduced in Synapse v1.37.0_
 async def check_event_for_spam(event: "synapse.events.EventBase") -> Union[bool, str]
 ```
 
-Called when receiving an event from a client or via federation. The module can return
-either a `bool` to indicate whether the event must be rejected because of spam, or a `str`
-to indicate the event must be rejected because of spam and to give a rejection reason to
-forward to clients.
+Called when receiving an event from a client or via federation. The callback must return
+either:
+- an error message string, to indicate the event must be rejected because of spam and 
+  give a rejection reason to forward to clients;
+- the boolean `True`, to indicate that the event is spammy, but not provide further details; or
+- the booelan `False`, to indicate that the event is not considered spammy.
 
 If multiple modules implement this callback, they will be considered in order. If a
 callback returns `False`, Synapse falls through to the next one. The value of the first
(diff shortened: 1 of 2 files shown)


docs/modules/spam_checker_callbacks.md
```diff
@@ -16,10 +16,12 @@ _First introduced in Synapse v1.37.0_
...
 ```
 
-Called when receiving an event from a client or via federation. The module can return
-either a `bool` to indicate whether the event must be rejected because of spam, or a `str`
-to indicate the event must be rejected because of spam and to give a rejection reason to
-forward to clients.
+Called when receiving an event from a client or via federation. The callback must return
+either:
+- an error message string, to indicate the event must be rejected because of spam and 
+  give a rejection reason to forward to clients;
```
clokep, line 22: Note that this won't be forwarded to clients over federation, but I don't know if that nuance is worth explaining.

docs/modules/spam_checker_callbacks.md
```diff
@@ -58,7 +63,8 @@ async def user_may_invite(inviter: str, invitee: str, room_id: str) -> bool
 
 Called when processing an invitation. The module must return a `bool` indicating whether
```
clokep, line 64: Interestingly this is called for both client-server API and federation, while `user_may_send_3pid_invite` is not (I suspect it doesn't make sense to, not sure it is worth calling out).

docs/modules/spam_checker_callbacks.md
```diff
@@ -166,7 +176,11 @@ async def check_username_for_spam(user_profile: Dict[str, str]) -> bool
 ```
 
 Called when computing search results in the user directory. The module must return a
-`bool` indicating whether the given user profile can appear in search results. The profile
+`bool` indicating whether the given user profile can appear in search results. Return
+`True` to indicate that the user is spammy and exclude them user from search results;
+otherwise return `False`.
+
+The profile
 is represented as a dictionary with the following keys:
```
clokep, line 184:
```suggestion
The profile is represented as a dictionary with the following keys:
```

docs/modules/spam_checker_callbacks.md
```diff
@@ -225,8 +239,9 @@ async def check_media_file_for_spam(
 ) -> bool
 ```
 
-Called when storing a local or remote file. The module must return a boolean indicating
-whether the given file can be stored in the homeserver's media store.
+Called when storing a local or remote file. The module must return a `bool` indicating
+whether the given file can be stored in the homeserver's media store. Return `False` if 
+to prevent this file from being stored; otherwise return `True`.
```
clokep, line 244: I believe this one is incorrect, see https://github.com/matrix-org/synapse/blob/10a88ba91cb16ccf757984f0a7d41ddf8b4dc07f/synapse/rest/media/v1/media_storage.py#L148-L156

```suggestion
Called when storing a local or remote file. The module must return a `bool` indicating
whether the given file can be stored in the homeserver's media store. Return `True`
to prevent this file from being stored; otherwise return `False`.
```

clokep (commented):
Thanks for improving those docs!

babolivier:
> This isn't obvious. As a result, for about a week or two we ran a module on matrix.org [which effectively considered all users spammy](https://github.com/matrix-org/message_limit/commit/ac7d84b1143d24460aa342009601213a4808f5fc#diff-107ab1be23f8882205b0f170ecb9893068bebe844b8df5199fa263fa7a8b2b61R34-R35).

For clarification, this was caused by the documentation lying (and saying that `True` meant users can appear in the user directory, which isn't the case, which is my bad), not by it being unclear. Though :+1: for clarification, it's worth doing it regardless.

docs/modules/spam_checker_callbacks.md
DMRobertson: Could say `local clients` here maybe?

docs/modules/spam_checker_callbacks.md
DMRobertson: Thanks for spotting this. I propose wording this slightly differently---commit to follow shortly.

DMRobertson:
> > This isn't obvious. As a result, for about a week or two we ran a module on matrix.org [which effectively considered all users spammy](https://github.com/matrix-org/message_limit/commit/ac7d84b1143d24460aa342009601213a4808f5fc#diff-107ab1be23f8882205b0f170ecb9893068bebe844b8df5199fa263fa7a8b2b61R34-R35).
> 
> For clarification, this was caused by the documentation lying (and saying that `True` meant users can appear in the user directory, which isn't the case, which is my bad), not by it being unclear. Though +1 for clarification, it's worth doing it regardless.

I'm not sure I'd agree with that---I genuinely wasn't sure what the documentation was describing until I consulted the source code. 

From chat elsewhere, I think the causes were
- putting together the module under time pressure;
- (reasonably!) assuming that this callback behaved like the others (return `True` for "no effect"); and
- booleans being inherently ambiguous (as opposed to some kind of named enum).

I'm not sure there was time to actually consult the documentation. But as you say, we may as well keep our docs in order.

clokep (approved):
Looks great, thank you!

PR #12005, opened 2022-02-16
`send_join` response: get create event from `state`, not `auth_chain`

richvdh:
https://github.com/matrix-org/matrix-doc/blob/rav/proposal/partial_state_on_join/proposals/3706-partial-state-in-send-join.md#changes-to-the-response proposes changing the `/send_join` response:

> Any events returned within `state` can be omitted from `auth_chain`.

Currently, we rely on `m.room.create` being returned in `auth_chain`, but since the `m.room.create` event must necessarily be part of the state, the above change will break this.

In short, let's look for `m.room.create` in `state` rather than `auth_chain`.

### DIFF
--- a/changelog.d/12005.misc
+++ b/changelog.d/12005.misc
@@ -1,0 +1,1 @@
+Preparation for faster-room-join work: when parsing the `send_join` response, get the `m.room.create` event from `state`, not `auth_chain`.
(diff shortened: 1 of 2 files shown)


babolivier (approved):

richvdh:
related: #11249

PR #12008, opened 2022-02-16
Remove support for the legacy structured logging configuration.

### LINKED ISSUE #12006: Remove legacy structured logging configuration
See [`setup_structured_logging`](https://github.com/matrix-org/synapse/blob/bf72d10dbf506f5ea486d67094b6003947d38fb7/synapse/logging/_structured.py#L136) and the helpers it uses (bits about `drains` and such). There are also some bits in the documentation.

clokep:
Fixes #12006


changelog.d/12008.removal
```diff
@@ -0,0 +1 @@
+Remove support for the legacy structured logging configuration (using `structured: true` in the Synapse configuration).
```
babolivier, line 1:
```suggestion
Remove support for the legacy structured logging configuration (i.e. using `structured: true` in the Synapse configuration).
```

Feels a bit like nitpicking, but I can see people reading this as "you should now use `structured: true`". Also could be worth linking to the upgrade notes.

docs/upgrade.md
```diff
@@ -85,6 +85,15 @@ process, for example:
     dpkg -i matrix-synapse-py3_1.3.0+stretch1_amd64.deb
     ```
 
+# Upgrading to v1.54.0
+
+## Legacy structured logging configuration removal
+
+This release removes support for the `structured: true` logging configuration
+which was deprecated in Synapse v1.23.0. If your logging configuration  contains
```
babolivier, line 93:
```suggestion
which was deprecated in Synapse v1.23.0. If your logging configuration contains
```

synapse/config/logger.py
```diff
@@ -138,6 +137,12 @@
 removed in Synapse 1.3.0. You should instead set up a separate log configuration file.
 """
 
+STRUCTURED_ERROR = """\
+Support for the structured configuration option was removed in Synapse 1.54.0.
+You should instead use the standard logging configuration. See
+https://matrix-org.github.io/synapse/v1.52/structured_logging.html
```
babolivier, line 143:
```suggestion
https://matrix-org.github.io/synapse/v1.54/structured_logging.html
```

So the link doesn't point to a doc saying that legacy configs will be automatically converted.

babolivier (approved):
Few wording nits but lgtm otherwise.

synapse/config/logger.py
clokep: Thanks! I originally removed the info completely so wanted to link to an old version which had it still, but then realized we should leave the conversion notes for a little while longer.

Issue #11031, opened 2021-10-08
Module callbacks should not fail open: exceptions should not equal acceptance


reivilibre:
Currently, if a `check_event_allowed` callback raises an exception, the check is skipped and we fail open (we move on to the next callback if there is one, or we accept the event otherwise).

This behaviour is undesirable for a security feature; intuitively, a failure in a security system should not be ignored.

If we failed closed by default, you could still wrap your module callback in `try`-`catch` and `return True` on failure if you desired the fail open behaviour.

On the other hand, failing open means that there is no way to fail closed without causing issues with federation: when we receive an event over federation and the callback fails, neither 'accept the event' or 'permanently reject the event' seems like a good answer: a failure instead should probably mean 'I can't accept this right now, but in the future I might be able to'.

To summarise:

* true should accept
* false should successfully reject
* an exception should be a failure, but neither a rejection nor an acceptance.
  * it has been suggested that we wrap these in a new `PluginFailedException`

(as an aside, we also mentioned that module authors should not be using `SynapseError` in their modules except as a response to a module's own HTTP resource: modules should prefer to use a generic exception type and allow Synapse to convert it to a 500 if that's the correct thing to do)

context: [the PR sparking this discussion](https://github.com/matrix-org/synapse/pull/11030); [discussion in #synapse-dev](https://matrix.to/#/!XaqDhxuTIlvldquJaV:matrix.org/$necZJEIYJFesBolvcdvN2MfxQfmdUH4P0_fjLahOb6E?via=matrix.org&via=vector.modular.im&via=envs.net)

PR #12009, opened 2022-02-16
Allow modules to set a display name on registration

babolivier:
Part of the Tchap mainlining. Very similar to [`get_username_for_registration`](https://matrix-org.github.io/synapse/latest/modules/password_auth_provider_callbacks.html#get_username_for_registration), except for setting the display name. The reason this is needed is because Tchap's Synapse fork derives the display name from the email address on registration - see [this code](https://github.com/matrix-org/synapse-dinsic/blob/f20caae101a482f19c81fbf36d4952388274d2d8/synapse/rest/client/register.py#L700-L710).

### DIFF
--- a/docs/modules/password_auth_provider_callbacks.md
+++ b/docs/modules/password_auth_provider_callbacks.md
@@ -166,6 +166,34 @@ any of the subsequent implementations of this callback. If every callback return
 the username provided by the user is used, if any (otherwise one is automatically
 generated).
 
+### `get_display_name_for_registration`
+
+_First introduced in Synapse v1.54.0_
+
+```python
+async def get_display_name_for_registration(
+    uia_results: Dict[str, Any],
+    params: Dict[str, Any],
+) -> Optional[str]
+```
+
+Called when registering a new user. The module can return a display name to set for the
+user being registered by returning it as a string, or `None` if it doesn't wish to force a
+display name for this user.
+
+This callback is called once [User-Interactive Authentication](https://spec.matrix.org/latest/client-server-api/#user-interactive-authentication-api)
+has been completed by the user. It is not called when registering a user via SSO. It is
+passed two dictionaries, which include the information that the user has provided during
+the registration process. These dictionaries are identical to the ones passed to
+[`get_username_for_registration`](#get_username_for_registration), so refer to the
+documentation of this callback for more information about them.
+
+If multiple modules implement this callback, they will be considered in order. If a
+callback returns `None`, Synapse falls through to the next one. The value of the first
+callback that does not return `None` will be used. If this happens, Synapse will not call
+any of the subsequent implementations of this callback. If every callback return `None`,
+the username will be used (e.g. `alice` if the user being registered is `@alice:example.com`).
+
 ## `is_3pid_allowed`
 
 _First introduced in Synapse v1.53.0_
@@ -195,7 +223,6 @@ The example module below implements authentication checkers for two different lo
 - `m.login.password` (defined in [the spec](https://matrix.org/docs/spec/client_server/latest#password-based))
     - Expects a `password` field to be sent to `/login`
     - Is checked by the method: `self.check_pass` 
-
 
 ```python
 from typing import Awaitable, Callable, Optional, Tuple
--- a/docs/modules/password_auth_provider_callbacks.md
+++ b/docs/modules/password_auth_provider_callbacks.md
@@ -166,12 +166,12 @@ any of the subsequent implementations of this callback. If every callback return
 the username provided by the user is used, if any (otherwise one is automatically
 generated).
 
-### `get_display_name_for_registration`
+### `get_displayname_for_registration`
 
 _First introduced in Synapse v1.54.0_
 
 ```python
-async def get_display_name_for_registration(
+async def get_displayname_for_registration(
     uia_results: Dict[str, Any],
     params: Dict[str, Any],
 ) -> Optional[str]
--- a/docs/modules/password_auth_provider_callbacks.md
+++ b/docs/modules/password_auth_provider_callbacks.md
@@ -85,7 +85,7 @@ If the authentication is unsuccessful, the module must return `None`.
 If multiple modules implement this callback, they will be considered in order. If a
 callback returns `None`, Synapse falls through to the next one. The value of the first
 callback that does not return `None` will be used. If this happens, Synapse will not call
-any of the subsequent implementations of this callback. If every callback return `None`,
+any of the subsequent implementations of this callback. If every callback returns `None`,
 the authentication is denied.
 
 ### `on_logged_out`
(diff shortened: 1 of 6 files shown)


docs/modules/password_auth_provider_callbacks.md
```diff
@@ -166,6 +166,34 @@ any of the subsequent implementations of this callback. If every callback return
...
+has been completed by the user. It is not called when registering a user via SSO. It is
+passed two dictionaries, which include the information that the user has provided during
+the registration process. These dictionaries are identical to the ones passed to
+[`get_username_for_registration`](#get_username_for_registration), so refer to the
+documentation of this callback for more information about them.
+
+If multiple modules implement this callback, they will be considered in order. If a
+callback returns `None`, Synapse falls through to the next one. The value of the first
+callback that does not return `None` will be used. If this happens, Synapse will not call
+any of the subsequent implementations of this callback. If every callback return `None`,
```
clokep, line 194:
```suggestion
any of the subsequent implementations of this callback. If every callback returns `None`,
```

synapse/module_api/__init__.py
```diff
@@ -328,6 +332,7 @@ def register_password_auth_provider_callbacks(
             is_3pid_allowed=is_3pid_allowed,
             auth_checkers=auth_checkers,
             get_username_for_registration=get_username_for_registration,
+            get_display_name_for_registration=get_display_name_for_registration,
```
clokep, line 335: Do we want to note in the docstring that a new property is being added?

synapse/rest/client/register.py
```diff
@@ -694,11 +694,19 @@ async def on_POST(self, request: SynapseRequest) -> Tuple[int, JsonDict]:
                 session_id
             )
 
+            desired_display_name = await (
```
clokep, line 697: It seems a little weird to call this `desired` since it isn't from the user.

tests/handlers/test_password_providers.py
```diff
@@ -877,20 +913,20 @@ def _test_3pid_allowed(self, username: str, registration: bool):
 
         m.assert_called_once_with("email", "bar@test.com", registration)
 
-    def _setup_get_username_for_registration(self) -> Mock:
+    def _setup_get_name_for_registration(self, callback_name: str) -> Mock:
         """Registers a get_username_for_registration callback that appends "-foo" to the
```
clokep, line 917: Please update the docstring!

tests/handlers/test_password_providers.py
```diff
@@ -817,6 +824,35 @@ def test_3pid_allowed(self):
         self._test_3pid_allowed("rin", False)
         self._test_3pid_allowed("kitay", True)
 
+    def test_display_name(self):
+        """Tests that the get_display_name_for_registration callback can define the
+        display name of a user when registering.
+        """
+        self._setup_get_name_for_registration(
+            callback_name=self.CALLBACK_DISPLAYNAME,
+        )
```
clokep, line 833: Should we be using the return value of the mock (e.g. ensuring it is called or something?)

clokep (commented):
Seems pretty reasonable, a few minor comments.

synapse/module_api/__init__.py
babolivier: I think the mention of the Synapse version in the callback doc should be good enough?

clokep: But the docstring is wrong now, it says it was introduced in a certain version, but that's only partially true?

docs/modules/password_auth_provider_callbacks.md
```diff
@@ -162,10 +162,38 @@ return `None`.
 If multiple modules implement this callback, they will be considered in order. If a
 callback returns `None`, Synapse falls through to the next one. The value of the first
 callback that does not return `None` will be used. If this happens, Synapse will not call
-any of the subsequent implementations of this callback. If every callback return `None`,
+any of the subsequent implementations of this callback. If every callback returns `None`,
 the username provided by the user is used, if any (otherwise one is automatically
 generated).
 
+### `get_display_name_for_registration`
```
clokep, line 169:
```suggestion
### `get_displayname_for_registration`
```

Although I'm not sure why this was changed? `display_name` seems clearer? Does the spec call it `displayname` or something?

tests/handlers/test_password_providers.py
```diff
@@ -817,6 +806,56 @@ def test_3pid_allowed(self):
         self._test_3pid_allowed("rin", False)
         self._test_3pid_allowed("kitay", True)
 
+    def test_displayname(self):
+        """Tests that the get_display_name_for_registration callback can define the
```
clokep, line 810:
```suggestion
        """Tests that the get_displayname_for_registration callback can define the
```

docs/modules/password_auth_provider_callbacks.md
```diff
@@ -162,10 +162,38 @@ return `None`.
...
+
+_First introduced in Synapse v1.54.0_
+
+```python
+async def get_display_name_for_registration(
```
clokep, line 174:
```suggestion
async def get_displayname_for_registration(
```

synapse/handlers/auth.py
```diff
@@ -2350,6 +2365,49 @@ async def get_username_for_registration(
 
         return None
 
+    async def get_displayname_for_registration(
```
clokep, line 2368: Not a huge deal, but it might make sense to make a helper function which does most of this, these could then become lightweight wrappers of something like:

```python
async def get_displayname_for_registration(self, uia_results, params):
    return await self._try_callbacks(self.get_displayname_for_registration_callbacks, uia_results, params)

async def _try_callbacks(self, callbacks, *args, **kwargs):
    ....
```

clokep: (This could definitely be a follow-up too, just noting that it seems we have a lot of boilerplate code here.)

synapse/module_api/__init__.py
babolivier: I wonder if we should just remove the "Introduced in..." mentions for the register callbacks functions, given we already track that info for individual callbacks in the docs it sounds more confusing than useful.

docs/modules/password_auth_provider_callbacks.md
babolivier: Yeah, the spec is consistently writing it `displayname` as far as I can tell.

richvdh: If only that were true: https://github.com/matrix-org/matrix-doc/issues/763

synapse/handlers/auth.py
babolivier: Agreed, I'd definitely like to have some standardised wrapping like this as a follow-up. (see also the aside mentioned in the description of https://github.com/matrix-org/synapse/issues/11031)

docs/modules/password_auth_provider_callbacks.md
babolivier: :sob:

synapse/module_api/__init__.py
babolivier: I've opened https://github.com/matrix-org/synapse/issues/12017 to track this (since it's not exclusive to auth provider callbacks).

PR #12013, opened 2022-02-16
Faster joins: Support for calling `/federation/v1/state`

richvdh:
This is an endpoint that we have server-side support for, but no client-side support. It's going to be useful for resyncing partial-stated rooms, so let's introduce it.

### DIFF
--- a/synapse/federation/federation_client.py
+++ b/synapse/federation/federation_client.py
@@ -464,11 +464,11 @@ class FederationClient(FederationBase):
         )
 
         valid_auth_events = await self._check_sigs_and_hash_and_fetch(
-            destination, auth_events, room_version
+            destination, auth_event_map.values(), room_version
         )
 
         valid_state_events = await self._check_sigs_and_hash_and_fetch(
-            destination, state_events, room_version
+            destination, state_event_map.values(), room_version
         )
 
         return valid_state_events, valid_auth_events
--- a/synapse/federation/federation_client.py
+++ b/synapse/federation/federation_client.py
@@ -17,6 +17,7 @@
 import copy
 import itertools
 import logging
+from functools import partial
 from typing import (
     TYPE_CHECKING,
     Awaitable,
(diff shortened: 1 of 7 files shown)


richvdh:
ugh, this really needs some tests.

reivilibre (changes requested):
This is still a draft so I'm not sure you actually wanted review ... can't argue with tests though :)

synapse/federation/federation_client.py
```diff
@@ -413,6 +414,74 @@ async def get_room_state_ids(
...
+        result = await self.transport_layer.get_room_state(
+            room_version,
+            destination,
+            room_id,
+            event_id,
+        )
+        state_events = result.state
+        auth_events = result.auth_events
+
+        # we may as filter out any duplicates from the response, which avoids some
```
richvdh, line 439:
```suggestion
        # we may as well filter out any duplicates from the response, which avoids some
```

synapse/federation/federation_client.py
```diff
@@ -413,6 +414,74 @@ async def get_room_state_ids(
...
+        # potential failure modes later.
```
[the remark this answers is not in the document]
richvdh: Yeah, I can't really remember :facepalm:

I don't think any duplication would actually be that problematic, it's just a bit inefficient.

```suggestion
        # we may as well filter out any duplicates from the response, to save
        # processing them multiple times. (In particular, events may be present in 
        # `auth_events` as well as `state`, which is redundant).
```

synapse/federation/federation_client.py
```diff
@@ -413,6 +414,74 @@ async def get_room_state_ids(
...
+        #
+        # We don't rely on the sort order of either of them, which makes it easier
```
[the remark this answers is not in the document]
richvdh: it's literally just saying that there's no need to preserve the order of the lists from the remote server, so we can just stick them all in a dict for easier indexing.

(as an example of a reason we *might* rely on ordering: imagine a world where we processed the auth events strictly in the order they came from the remote server. In that case, the auth_events for event X would have to precede X in the list, so if we were filtering duplicates, we'd have to be more careful about preserving ordering.

But, we don't do that, so going into a dict and back is fine.)

```suggestion
        # We don't rely on the sort order of the events, so we can just stick them in a dict.
```

tests/federation/test_federation_client.py
```diff
@@ -0,0 +1,147 @@
...
+            }
+        )
+        member_event_dict = self.add_hashes_and_signatures(
+            {
+                "room_id": test_room_id,
+                "type": "m.room.member",
+                "sender": creator,
+                "state_key": creator,
+                "content": {"membership": "join"},
+                "prev_events": [],
```
reivilibre, line 65: I'm surprised that these non-m.room.create events don't have any prev events (I'd expect these 3 to form a chain, or at least for the latter 2 to point to the room creation event).

tests/federation/test_federation_client.py
```diff
@@ -0,0 +1,147 @@
...
+        )
+
+        # mock up the response, and have the agent return it
+        self._mock_agent.request.return_value = defer.succeed(
+            _mock_response(
+                {
+                    "pdus": [
+                        create_event_dict,
+                        member_event_dict,
+                        pl_event_dict,
```
reivilibre, line 90: I suppose this somewhat goes back to my earlier point of confusion: I would expect the latter 2 events of these to depend on the first in some way, but the code you've written so far claims that the order doesn't matter.

Supposing my expectation was right: that these events should be interdependent in some way, how does it make sense for us to process them in 'any' order where we might come across some before the events they depend on?
I realise we have a concept of 'outliers' but I think I'd benefit from better seeing how this fits together.

reivilibre (changes requested):
Looks sensible; I'd basically say this is good to go; probably requesting changes in my understanding moreso than in your code

tests/federation/test_federation_client.py
richvdh: yeah, they should, but the bit of code under test doesn't care about that - it just deserialises each event, rather than checking the contents thereof. I'll add a comment.

tests/federation/test_federation_client.py
```diff
@@ -0,0 +1,147 @@
...
+        # mock out the Agent used by the federation client, which is easier than
+        # catching the HTTPS connection and do the TLS stuff.
+        self._mock_agent = mock.create_autospec(twisted.web.client.Agent, spec_set=True)
+        homeserver.get_federation_http_client().agent = self._mock_agent
+
+    def test_get_room_state(self):
+        creator = f"@creator:{self.OTHER_SERVER_NAME}"
+        test_room_id = "!room_id"
+
+        # mock up some events to use in the response
```
richvdh, line 45:
```suggestion
        # mock up some events to use in the response.
        # In real life, these would have things in `prev_events` and `auth_events`, but that's a bit
        # annoying to mock up, and the code under test doesn't care, so we don't bother.
```

tests/federation/test_federation_client.py
richvdh: The thing is that `FederationClient.get_room_state` makes it the *caller's* problem to deal with the interdependencies, which is why you aren't seeing any code here which cares about this.

If it helps put it into context, #12040 builds on this PR with a call to `get_room_state` at https://github.com/matrix-org/synapse/pull/12040/files#diff-a91beeaf44154c199a11f5539f2358704a1df966ff74ad8f2681b64430843f8cR968. It just passes the results into `_auth_and_persist_outliers` (https://github.com/matrix-org/synapse/blob/develop/synapse/handlers/federation_event.py#L1225) which handles the ordering problem by making sure that we persist each event's auth_events before the event itself.

tests/federation/test_federation_client.py
richvdh: done in 0ac9ad2.

tests/federation/test_federation_client.py
reivilibre: Aha, thank you :)

reivilibre (approved):
Thanks for the explanation!

PR #12015, opened 2022-02-17
Configure `tox` to use `venv`

richvdh:
As the comment says, virtualenv is a pile of fail.


clokep (approved):
Sad that this isn't the default. 😢

PR #12016, opened 2022-02-17
Fix bug in `StateFilter.return_expanded()` and add some tests.

reivilibre:
If you rollback to just before the fix and run the tests, you'll see the error that gets hit:

```
  File "/home/rei/work/synapse/tests/storage/test_state.py", line 1086, in test_return_expanded
    StateFilter.freeze(
  File "/home/rei/work/synapse/synapse/storage/state.py", line 211, in return_expanded
    types=frozendict({EventTypes.Member: self.types[EventTypes.Member]}),
builtins.KeyError: 'm.room.member'
```

I suspect this hasn't actually ever been hit in production (hence the `misc` label) but I noticed
it was wrong when working on #10870.

### DIFF
--- a/tests/storage/test_state.py
+++ b/tests/storage/test_state.py
@@ -992,3 +992,112 @@ class StateFilterDifferenceTestCase(TestCase):
             StateFilter.none(),
             StateFilter.all(),
         )
+
+
+class StateFilterTestCase(TestCase):
+    def test_return_expanded(self):
+        """
+        Tests the behaviour of the return_expanded() function that expands
+        StateFilters to include more state types (for the sake of cache hit rate).
+        """
+
+        self.assertEqual(StateFilter.all().return_expanded(), StateFilter.all())
+
+        self.assertEqual(StateFilter.none().return_expanded(), StateFilter.none())
+
+        # Concrete-only state filters stay the same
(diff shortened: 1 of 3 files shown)


DMRobertson (changes requested):
I can't see a test case which ensures that member-wildcards are not expanded. Think it might be worth having one (this machinery seems tricky to grok).

Are there are end-user consequences of this bug?

tests/storage/test_state.py
```diff
@@ -1050,14 +1050,14 @@ def test_return_expanded(self):
...
+        # Wildcard member-only state filters stay the same
         self.assertEqual(
             StateFilter.freeze(
-                {EventTypes.Member: {"@wombat:test", "@alicia:test"}},
+                {EventTypes.Member: None},
                 include_others=False,
             ).return_expanded(),
             StateFilter.freeze(
-                {EventTypes.Member: {"@wombat:test", "@alicia:test"}},
+                {EventTypes.Member: None},
```
DMRobertson, line 1060: Thanks---though I think both of these test cases are worth having?

reivilibre: The test case just before was a dupe unless I misread.

PR #12018, opened 2022-02-17
Remove unstable MSC3283 flags

### LINKED ISSUE #11962: Remove unstable flags for MSC3283
#11933 added support for the stable flags, so we should remove the unstable ones in the next release

erikjohnston:
Fixes #11962

Q: Does this warrant a note in the upgrades file?


clokep (approved):
Seems reasonable. Might be worth linking to the MSC in the changelog.

erikjohnston:
> Seems reasonable. Might be worth linking to the MSC in the changelog.

Good point well made

Issue #11474, opened 2021-12-01
Update generated URLs to point to v3


clokep:
Related to #11079, we generate some URLs internally that point to `r0` URLs, some places we do this is for:

* Callback URLs used in SSO
* Attempting to fetch media from remote servers

Maybe other places. We should eventually update these to use `v3` URLs. The ones that generate URLs that go back to ourselves should be easy enough (if people have their reverse proxies configured properly), but the ones to other servers are a bit more complicated. For federation we try `v2` URLs then fallback to `v1` URLs, but that's rather annoying. Not sure if there's something better to do.

PR #12019, opened 2022-02-17
Use v3 endpoints for fallback auth (Matrix 1.1)

turt2live:
Untested, but no reason to believe this wouldn't work. Part of [Matrix 1.1](https://github.com/matrix-org/synapse/issues/11079) work, but not blocking.

### Pull Request Checklist

### DIFF
--- a/changelog.d/12019.misc
+++ b/changelog.d/12019.misc
@@ -1,0 +1,1 @@
+Use Matrix v1.1 endpoints (`/_matrix/client/v3/auth/...`) in fallback auth HTML forms.
--- a/synapse/rest/client/auth.py
+++ b/synapse/rest/client/auth.py
@@ -62,7 +62,7 @@ class AuthRestServlet(RestServlet):
         if stagetype == LoginType.RECAPTCHA:
             html = self.recaptcha_template.render(
                 session=session,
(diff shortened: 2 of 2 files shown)


clokep:
Related to #11474, specifically, I think?

turt2live:
Related, yes. I didn't run across them in my relatively quick checks against the v1.1 feature set, so would recommend they be fixed in a future PR.

babolivier:
I'm wondering if we don't want the v3 and r0 URLs coexist for a bit to give clients time to update their paths and to give users time to update their clients.

reivilibre:
> I'm wondering if we don't want the v3 and r0 URLs coexist for a bit to give clients time to update their paths and to give users time to update their clients.

I'm not sure this is what this is getting at. It appears to be setting the URL for fallback auth to be `v3` rather than `r0`. The interaction is handled in a general-purpose web browser so clients have nothing to update.

(Please correct me if I'm wrong!)

clokep:
Both URLs should work fine still, this just changes where Synapse redirects / point people too.

The only risk is that someone is block `v3` in their reverse proxy for some reason.

erikjohnston (approved):
Sounds like everyone thinks this is fine.

PR #12025, opened 2022-02-18
Update the olddeps CI check to use an old version of markupsafe

babolivier:
Our olddeps check don't check for the oldest possible versions on _transitive_ dependencies. Markupsafe 2.1 was released recently, which breaks Jinja 2.x. So olddeps installs Jinja 2.9 (since it's the oldest version allowed) and markupsafe 2.1 (since it's the latest version of the transitive dep), which breaks everything.

See https://github.com/pallets/jinja/issues/1587

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -124,6 +124,7 @@ usedevelop = false
 deps =
     Automat == 0.8.0
     lxml
+    markupsafe < 2.1
     {[base]deps}
 
 commands =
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -124,6 +124,7 @@ usedevelop = false
 deps =
     Automat == 0.8.0
     lxml
+    markupsafe < 2.1
```
richvdh, line 127: could you add a comment here please. Why is this needed?

richvdh (changes requested):

richvdh (approved):
thx!

PR #12032, opened 2022-02-18
AS transaction isolation tweaks

### LINKED ISSUE #11620: SerializationFailure on appservice outgoing transaction (updating `last_txn`)
### Description

We are seeing AS' go into recoverer mode after a serialization failure when updating `application_services_state.last_txn` (https://github.com/matrix-org/synapse/blob/v1.48.0/synapse/storage/databases/main/appservice.py#L272). We recently fixed a similar issue in https://github.com/matrix-org/synapse/issues/11195 which targets the same database row, so I suspect the issue now is that this `last_txn` update occaisionally clashes with the stream position update we added the linearizer for.

I see that the `last_txn` implementation was added a few years back (https://github.com/matrix-org/synapse/commit/0a60bbf4fac4262da3fee702ca46d2f019597ef1) and hasn't changed since - as far as I can tell the only use of this column is to check the previous value hasn't incremented more than one. This does appear to get triggered from time to time still - I can see 12 log entires in the last 7 days in our deployment.

### Possible Solutions

- move this column to it's own table, not great but would get around the immediate issue
- fix the ordering issue (or work around it) and remove the column/check entirely

I notice the recoverer just pulls the lowest transaction and sends that - could the entire sending mechanism work just like that instead? Simply have the `_TransactionController` write out transactions to the database in the batches and have process (currently `_Recoverer`) that just pulls from these constantly. This would remove any need to track the last transaction ID for a given AS.

### Version information

- v1.48
- worker based deployment with dedicated appservice pusher worker

Fizzadar:
Probably fixes #11620.

Let me know if you'd prefer this in two PRs!

Signed-off-by: Nick Mills-Barrett <nick@beeper.com>

### Pull Request Checklist

### DIFF
--- a/synapse/storage/databases/main/appservice.py
+++ b/synapse/storage/databases/main/appservice.py
@@ -26,6 +26,7 @@ from synapse.events import EventBase
 from synapse.storage._base import SQLBaseStore, db_to_json
 from synapse.storage.database import DatabasePool, LoggingDatabaseConnection
 from synapse.storage.databases.main.events_worker import EventsWorkerStore
+from synapse.storage.engines._base import IsolationLevel
 from synapse.types import JsonDict
 from synapse.util import json_encoder
 
@@ -294,7 +295,9 @@ class ApplicationServiceTransactionWorkerStore(
             )
 
         await self.db_pool.runInteraction(
-            "complete_appservice_txn", _complete_appservice_txn
+            "complete_appservice_txn",
+            _complete_appservice_txn,
+            isolation_level=IsolationLevel.READ_COMMITTED,
         )
 
     async def get_oldest_unsent_txn(
--- a/synapse/storage/databases/main/appservice.py
+++ b/synapse/storage/databases/main/appservice.py
@@ -434,16 +434,12 @@ class ApplicationServiceTransactionWorkerStore(
                 % (stream_type,)
             )
 
-        def set_appservice_stream_type_pos_txn(txn):
-            stream_id_type = "%s_stream_id" % stream_type
-            txn.execute(
-                "UPDATE application_services_state SET %s = ? WHERE as_id=?"
-                % stream_id_type,
-                (pos, service.id),
-            )
+        stream_id_type = "%s_stream_id" % stream_type
 
-        await self.db_pool.runInteraction(
-            "set_appservice_stream_type_pos", set_appservice_stream_type_pos_txn
+        await self.db_pool.simple_update_one(
+            "application_services_state",
+            {"as_id": service.id},
+            {stream_id_type: pos},
         )
 
 
--- a/changelog.d/12032.misc
+++ b/changelog.d/12032.misc
@@ -1,0 +1,1 @@
+Reduce AS state transaction isolation level. Contributed by Nick @ Beeper.


synapse/storage/databases/main/appservice.py
```diff
@@ -434,16 +437,12 @@ async def set_appservice_stream_type_pos(
...
-                (pos, service.id),
-            )
+        stream_id_type = "%s_stream_id" % stream_type
 
-        await self.db_pool.runInteraction(
-            "set_appservice_stream_type_pos", set_appservice_stream_type_pos_txn
+        await self.db_pool.simple_update_one(
+            "application_services_state",
+            {"as_id": service.id},
+            {stream_id_type: pos},
```
squahtx, line 445: This transaction may still encounter a SerializationFailure since it's on REPEATABLE READ.

synapse/storage/databases/main/appservice.py
```diff
@@ -294,7 +295,9 @@ def _complete_appservice_txn(txn):
             )
 
         await self.db_pool.runInteraction(
-            "complete_appservice_txn", _complete_appservice_txn
+            "complete_appservice_txn",
+            _complete_appservice_txn,
+            isolation_level=IsolationLevel.READ_COMMITTED,
```
squahtx, line 300: I didn't realise one of the transactions in question was non-trivial! That complicates things a little since `last_txn` can now change out from under us in between `_get_last_txn` and the upsert. Previously, we'd get a SerializationFailure and retry until we get a consistent view throughout the transaction.

I think this is okay though since `last_txn_id` is only used for a debug check, which will now be less reliable - if two `complete_appservice_txn` transactions run simultaneously with the same `txn_id`, neither of them may trip the debug check. It's best to leave a comment noting that `last_txn_id` can change out from under us before the upsert because we're in READ COMMITTED mode.

synapse/storage/databases/main/appservice.py
erikjohnston: It won't, as we use `autocommit` mode for one off queries like this, which acts more like `READ COMMITTED`

erikjohnston: (Which is non-obvious unless you know about it!)

synapse/storage/databases/main/appservice.py
Fizzadar: Another thought - should `txn_id` be generated using a sequence? The debug check and manual generation of TXN ID seems unncessary to me (this would potentially open up running multiple appservice workers with additional work I think?).

Currently I believe this is safe but only because there is only a single process-per-AS working on transactions (enforced by https://github.com/matrix-org/synapse/blob/develop/synapse/appservice/scheduler.py#L181).

squahtx: For some reason the debug check trips on matrix.org from time to time, perhaps around the time synapse workers are restarting(?). It's not immediately clear to me why it happens.

I can't see why we don't use a sequence for `txn_id`. Something to do with sqlite support maybe?

Fizzadar: Possibly historical reasons? The AS codepaths look old on blame to me! Not super familiar with the rest of synapse but sequences in use/defined in https://github.com/matrix-org/synapse/blob/develop/synapse/storage/util/sequence.py so potentially would be beneficial to switch to those. Need more context to make that call though, and probably beyond the scope of this PR!

Fizzadar:
This has now been fixed in https://github.com/matrix-org/synapse/pull/12209!

PR #12036, opened 2022-02-18
Rename shared_rooms to mutual_rooms

ShadowJonathan:
To be in line with proposal changes from [matrix-org/matrix-doc@`fbbb2d9` (#2666)](https://github.com/matrix-org/matrix-doc/pull/2666/commits/fbbb2d95640457424a018d57df3b121d8a805665)

### Pull Request Checklist

* [x] Pull request is based on the develop branch
* [x] Pull request includes a [changelog file](https://matrix-org.github.io/synapse/latest/development/contributing_guide.html#changelog). The entry should:
  - Be a short description of your change which makes sense to users. "Fixed a bug that prevented receiving messages from other servers." instead of "Moved X method from `EventStore` to `EventWorkerStore`.".
  - Use markdown where necessary, mostly for `code blocks`.
  - End with either a period (.) or an exclamation mark (!).
  - Start with a capital letter.
  - Feel free to credit yourself, by adding a sentence "Contributed by @github_username." or "Contributed by [Your Name]." to the end of the entry.
* [x] Pull request includes a [sign off](https://matrix-org.github.io/synapse/latest/development/contributing_guide.html#sign-off)
* [x] [Code style](https://matrix-org.github.io/synapse/latest/code_style.html) is correct
  (run the [linters](https://matrix-org.github.io/synapse/latest/development/contributing_guide.html#run-the-linters))

`Signed-off-by: Jonathan de Jong <jonathan@automatia.nl>`

PR #12038, opened 2022-02-18
Remove mutual_rooms `update_user_directory` check, and add extra documentation

ShadowJonathan:
Resolves https://github.com/matrix-org/synapse/issues/10339

### Pull Request Checklist

### DIFF
--- a/docs/workers.md
+++ b/docs/workers.md
@@ -528,9 +528,18 @@ the following regular expressions:
 
     ^/_matrix/client/(r0|v3|unstable)/user_directory/search$
 
-When using this worker you must also set `update_user_directory: False` in the
+When using this worker you must also set `update_user_directory: false` in the
 shared configuration file to stop the main synapse running background
 jobs related to updating the user directory.
+
+Above endpoint is not *required* to be routed to this worker. By default,
+`update_user_directory` is set to `true`, which means the main process
+will handle updates. All workers configured with `client` can handle the above
+endpoint as long as either this worker or the main process are configured to
+handle it, and are online.
+
+If `update_user_directory` is set to `false`, and this worker is not running,
+the above endpoint may give outdated results.
 
 ### `synapse.app.frontend_proxy`
 
--- a/synapse/rest/client/shared_rooms.py
+++ b/synapse/rest/client/shared_rooms.py
@@ -48,13 +48,6 @@ class UserSharedRoomsServlet(RestServlet):
         self, request: SynapseRequest, user_id: str
     ) -> Tuple[int, JsonDict]:
 
-        if not self.user_directory_active:
-            raise SynapseError(
-                code=400,
-                msg="The user directory is disabled on this server. Cannot determine shared rooms.",
-                errcode=Codes.FORBIDDEN,
-            )
-
         UserID.from_string(user_id)
 
         requester = await self.auth.get_user_by_req(request)
--- a/synapse/rest/client/shared_rooms.py
+++ b/synapse/rest/client/shared_rooms.py
@@ -43,10 +43,18 @@ class UserSharedRoomsServlet(RestServlet):
         self.auth = hs.get_auth()
         self.store = hs.get_datastores().main
         self.user_directory_active = hs.config.server.update_user_directory
+        self.hs = hs
 
     async def on_GET(
         self, request: SynapseRequest, user_id: str
     ) -> Tuple[int, JsonDict]:
+
+        if not self.hs.config.userdirectory.user_directory_search_enabled:
+            raise SynapseError(
+                code=400,
+                msg="User directory searching is disabled",
+                errcode=Codes.UNKNOWN,
+            )
 
         UserID.from_string(user_id)
 
--- a/changelog.d/12038.misc
+++ b/changelog.d/12038.misc
@@ -1,0 +1,1 @@
+Remove check on `update_user_directory` for shared rooms handler (MSC2666), and update/expand documentation.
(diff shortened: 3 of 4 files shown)


synapse/rest/client/shared_rooms.py
```diff
@@ -48,13 +48,6 @@ async def on_GET(
         self, request: SynapseRequest, user_id: str
     ) -> Tuple[int, JsonDict]:
 
-        if not self.user_directory_active:
```
erikjohnston, line 51: This check probably wants to be replaced with `self.hs.config.userdirectory.user_directory_search_enabled` as we do for `/user_directory/search`:

https://github.com/matrix-org/synapse/blob/9e83521af860cb33a7459dbe74188ce5ef39f446/synapse/rest/client/user_directory.py#L61-L62

docs/workers.md
```diff
@@ -467,10 +467,19 @@ the following regular expressions:
...
 jobs related to updating the user directory.
 
+Above endpoint is not *required* to be routed to this worker. By default,
+`update_user_directory` is set to `true`, which means the main process
+will handle updates. All workers configured with `client` can handle the above
+endpoint as long as either this worker or the main process are configured to
+handle it, and are online.
+
+If `update_user_directory` is set to `false`, and this worker is not running,
+the above endpoint may give outdated results.
```
erikjohnston, line 481: It looks like it will actually return nothing.

docs/workers.md
```diff
@@ -467,10 +467,19 @@ the following regular expressions:
...
-When using this worker you must also set `update_user_directory: False` in the
+When using this worker you must also set `update_user_directory: false` in the
 shared configuration file to stop the main synapse running background
 jobs related to updating the user directory.
 
+Above endpoint is not *required* to be routed to this worker. By default,
+`update_user_directory` is set to `true`, which means the main process
+will handle updates. All workers configured with `client` can handle the above
+endpoint as long as either this worker or the main process are configured to
+handle it, and are online.
```
erikjohnston, line 478: Bleurgh, this is a bit annoying to document. Ideally we'd deprecate `synapse.app.user_dir` and replace it with an option in `generic_worker` (like for background tasks).

synapse/rest/client/shared_rooms.py
ShadowJonathan: No, please review this comment; https://github.com/matrix-org/synapse/issues/10339#issuecomment-1039175679

TLDR; This PR explicitly deals with the removal of this line, the only thing this bit of code needs is for the user directory updater to exist *somewhere* in the cluster, and by default it'll exist on the main process.

The pros and cons of this have already been agreed upon by anoa and rich in that issue.

docs/workers.md
ShadowJonathan: If the user directory worker has been enabled some time in the past, but it's been disabled some time thereafter, then the database might have stale information, so this might return outdated data.

synapse/rest/client/shared_rooms.py
erikjohnston: > TLDR; This PR explicitly deals with the removal of this line, the only thing this bit of code needs is for the user directory updater to exist _somewhere_ in the cluster, and by default it'll exist on the main process.

if `self.hs.config.userdirectory.user_directory_search_enabled` is `False` then the user directory updater doesn't exist anywhere, as its completely disabled. The issue seems to be only discussing the worker case, rather than the user directory being disabled entirely? I don't really understand why we would want to process the request if the directory is disabled, we should either `400` or early return with an empty response like we do for the search endpoint.

docs/workers.md
erikjohnston: Oh, right yup, sorry I misread which config option this was talking about.

docs/workers.md
ShadowJonathan: There's some talk about this in #10339, I don't particularly know how to resolve this now.

synapse/rest/client/shared_rooms.py
erikjohnston: I'm happy with 400ing and changing if the MSC decides something different :+1:

docs/workers.md
anoadragon453: Given the scale of the discussion in #10339, I would document the current state now and get this piece of the puzzle merged. Then we can worry about from the transition from worker type -> generic worker option (https://github.com/matrix-org/synapse/issues/10441).

ShadowJonathan: So should I additionally document the `synapse.app.user_dir` worker type in `workers.md`?

ShadowJonathan: Actually, the worker type is already documented, i think that we shouldnt encode future intentions in the documentation, until we actually are working on it.

synapse/rest/client/shared_rooms.py
```diff
@@ -43,16 +43,17 @@ def __init__(self, hs: "HomeServer"):
...
         self.store = hs.get_datastores().main
         self.user_directory_active = hs.config.server.update_user_directory
+        self.hs = hs
 
     async def on_GET(
         self, request: SynapseRequest, user_id: str
     ) -> Tuple[int, JsonDict]:
 
-        if not self.user_directory_active:
+        if not self.hs.config.userdirectory.user_directory_search_enabled:
```
erikjohnston, line 52: Can we make this a variable rather than saving `self.hs`? It's a bit of a smell to have `self.hs` lying around if you can avoid it.

erikjohnston (commented):
Otherwise LGTM

ShadowJonathan:
I'll rebase this on https://github.com/matrix-org/synapse/pull/12036

erikjohnston (approved):

PR #12039, opened 2022-02-18
remote join processing: get create event from state, not auth_chain

richvdh:
A follow-up to #12005, in which I apparently missed that there are a bunch of other places that assume the create event is in the auth chain.


richvdh:
related: #11249

Issue #9864, opened 2021-04-22
We don't invalidate state cache when deleting state groups during room history purge


erikjohnston:
The error in the logs is:

```
2021-04-21 12:48:14,592 - synapse.handlers.federation - 1239 - ERROR - GET-2201 - Failed to backfill from localhost:8875 because Trying to persist state with unpersisted prev_group: 3369
Traceback (most recent call last):
  File "/venv/lib/python3.7/site-packages/synapse/handlers/federation.py", line 1208, in try_backfill
    dom, room_id, limit=100, extremities=extremities
  File "/venv/lib/python3.7/site-packages/synapse/handlers/federation.py", line 1031, in backfill
    context = await self.state_handler.compute_event_context(event)
  File "/venv/lib/python3.7/site-packages/synapse/state/__init__.py", line 385, in compute_event_context
    current_state_ids=state_ids_after_event,
  File "/venv/lib/python3.7/site-packages/synapse/storage/state.py", line 592, in store_state_group
    event_id, room_id, prev_group, delta_ids, current_state_ids
  File "/venv/lib/python3.7/site-packages/synapse/storage/databases/state/store.py", line 492, in store_state_group
    "store_state_group", _store_state_group_txn
  File "/venv/lib/python3.7/site-packages/synapse/storage/database.py", line 667, in runInteraction
    **kwargs,
  File "/venv/lib/python3.7/site-packages/synapse/storage/database.py", line 743, in runWithConnection
    self._db_pool.runWithConnection(inner_func, *args, **kwargs)
  File "/venv/lib/python3.7/site-packages/twisted/python/threadpool.py", line 238, in inContext
    result = inContext.theWork()  # type: ignore[attr-defined]
  File "/venv/lib/python3.7/site-packages/twisted/python/threadpool.py", line 255, in <lambda>
    ctx, func, *args, **kw
  File "/venv/lib/python3.7/site-packages/twisted/python/context.py", line 118, in callWithContext
    return self.currentContext().callWithContext(ctx, func, *args, **kw)
  File "/venv/lib/python3.7/site-packages/twisted/python/context.py", line 83, in callWithContext
    return func(*args, **kw)
  File "/venv/lib/python3.7/site-packages/twisted/enterprise/adbapi.py", line 293, in _runWithConnection
    compat.reraise(excValue, excTraceback)
  File "/venv/lib/python3.7/site-packages/twisted/python/deprecate.py", line 298, in deprecatedFunction
[...]

PR #12040, opened 2022-02-18
Optimise `_get_state_after_missing_prev_event`: use `/state`

richvdh:
If we're missing most of the events in the room state, then we may as well call the `/state` endpoint, instead of individually requesting each and every event.

Builds on #12013 which added the `get_room_state` method.

Requires the changes to sytest which landed in matrix-org/sytest#1223, and those which have yet to land in https://github.com/matrix-org/sytest/pull/1231.

### DIFF
--- a/synapse/handlers/federation_event.py
+++ b/synapse/handlers/federation_event.py
@@ -897,10 +897,24 @@ class FederationEventHandler:
         logger.debug("We are also missing %i auth events", len(missing_auth_events))
 
         missing_events = missing_desired_events | missing_auth_events
-        logger.debug("Fetching %i events from remote", len(missing_events))
-        await self._get_events_and_persist(
-            destination=destination, room_id=room_id, event_ids=missing_events
-        )
+
+        # Making an individual request for each of 1000s of events has a lot of
+        # overhead. On the other hand, we don't really want to fetch all of the events
+        # if we already have most of them.
+        #
+        # As an arbitrary heuristic, if we are missing more than 10% of the events, then
+        # we fetch the whole state.
+        #
+        # TODO: might it be better to have an API which lets us do an aggregate event
+        #   request
+        if (len(missing_events) * 10) >= len(auth_event_ids) + len(state_event_ids):
+            logger.debug("Requesting complete state from remote")
+            await self._get_state_and_persist(destination, room_id, event_id)
+        else:
+            logger.debug("Fetching %i events from remote", len(missing_events))
+            await self._get_events_and_persist(
+                destination=destination, room_id=room_id, event_ids=missing_events
+            )
 
         # we need to make sure we re-load from the database to get the rejected
         # state correct.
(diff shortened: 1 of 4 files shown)


richvdh:
Something odd is going on here.

The tests are - inconsistently - failing on "Can backfill purged history". This seems to affect largely worker-mode deployments.

1. The first thing that is odd is that the worker-mode deployments perform completely differently to monolith deployments; I think this is because we aren't correctly clearing the caches (https://github.com/matrix-org/synapse/issues/9864) which means it's working on monoliths for the wrong reason. This is not new; it just goes to explain some of the erratic behaviour.

2. The `/messages` request triggers a `backfill`, which returns 10 events. It seems that, at least on workers depending on timings, these 10 events might be the test messages, or might include a membership event. It kinda shouldn't matter - we should end up pulling all of them in anyway - but it does mean that we might end up processing things in a different order.  Again, this is not new.

3. Possibly related to 2: server-1 appears not to receive some of the test events, and returns 404s for later `/state_ids` requests for them. I don't really understand why this is possible - even if it misses them, it should backfill them from server-0. That means that server-0 then fails to process some of the events in the `/backfill` result and ignores them. Once more, I don't think this is new.

4. server-1 ends up persisting some events (in particular, its invite) as outliers. Later, server-0 makes a `/state_ids` request, which returns an empty list (https://github.com/matrix-org/synapse/issues/12074#issuecomment-1050084761). This is fairly clearly wrong, but again, not new.

5. What *is* new is that we calculate we have less than 10% of the events in the state (because  0 * 10 >= 0) so then make a `/state` request, which then returns a 500 (#12074 again). Arguably we could change that `>=` to a `>` to work around it, but it kinda misses the point here.

6. If you're unlucky, server-0 ends up blacklisting server-1 due to the 500s, and then the whole test fails. I haven't figured out what causes the blacklist to happen sometimes and not others.

In short, #12074 is the main problem here, though there is a general catalogue of fail.

richvdh:
> Something odd is going on here.

this is all (hopefully) now fixed.

synapse/handlers/federation_event.py
```diff
@@ -959,6 +973,27 @@ async def _get_state_after_missing_prev_event(
...
+            destination, room_id, event_id=event_id, room_version=room_version
+        )
+        logger.info("/state returned %i events", len(auth_events) + len(state_events))
+
+        await self._auth_and_persist_outliers(
+            room_id, itertools.chain(auth_events, state_events)
+        )
+
+        # we also need the event itself.
+        if not await self._store.have_seen_events(room_id, event_id):
```
DMRobertson, line 992: I think you want

```suggestion
        if not await self._store.have_seen_events(room_id, (event_id,)):
```

Mypy can't spot this because `x: str` implies `x: Iterable[str]`. @squahtx noted this pain point a few weeks ago I think.

(Given that CI passed, are we missing some kind of test here?)

DMRobertson (changes requested):

synapse/handlers/federation_event.py
DMRobertson: Xref: https://github.com/python/typing/issues/256

richvdh: Good spot, thank you.

I've fixed this by implementing `have_seen_event` and using that instead, which seems to have a lower potential for foot-shooting.

I've also added a UT which catches this.

The failure mode here was relatively edge-casey (it requires us to *have* the event, but not its state), and benign (it resulted in a redundant call to `_matrix/federation/event`), which was why this wasn't being picked up in the tests before.

richvdh:
ugh mypy. I'll fix this up another time.

DMRobertson (approved):
The unit test is a little daunting, but LGTM.

PR #12041, opened 2022-02-18
Run `_handle_queued_pdus` as a background process

richvdh:
... to ensure it gets a proper log context, mostly.

(Background: any events that arrive over federation while a join is in progress  get stuck in a queue which is then processed once the join completes. `_handle_queued_pdus` processes those events.)


clokep (approved):
These were logging against the sentinel context, I assume?

clokep:
Looks like you need to import this though. 👍 Guessing it was split from another branch?

richvdh:
> These were logging against the sentinel context, I assume?

No, in the context for the join request (with later warnings about finished log contexts).

PR #12042, opened 2022-02-18
Fix type hints for txredis.

clokep:
The type hints were subtly wrong for txredis. A bunch of places it stated we used `RedisProtocol`, but it was actually a `ConnectionHandler`... this is made more confusing that `ConnectionHandler` essentially choosing connections from the factory's pool of connections. There's also a `ReplicationCommandHandler`, which is a different thing specific to Synapse.

You end up with something like this (with lots of things referring to each other...). Fat lines are sub-classes, thin lines are references. (Note that the `pool` below can have multiple `RedisSubscriber` instances, I put one for clarity and because we set `poolsize=1`).

```mermaid
flowchart TD
RedisDirectTcpReplicationClientFactory ==> SynapseRedisFactory
RedisDirectTcpReplicationClientFactory -->|synapse_outbound_redis_connection| ConnectionHandler
RedisDirectTcpReplicationClientFactory -->|synapse_handler| ReplicationCommandHandler 
SynapseRedisFactory ==> RedisFactory
ConnectionHandler -->|_factory| RedisFactory
RedisFactory --> |handler| ConnectionHandler
RedisFactory --> |pool| RedisSubscriber
RedisSubscriber -->|synapse_outbound_redis_connection| ConnectionHandler
RedisSubscriber -->|synapse_handler| ReplicationCommandHandler
RedisSubscriber ==> SubscriberProtocol
SubscriberProtocol ==> RedisProtocol
```


clokep:
Something that's unclear to me -- we seem to be (sometimes?) sharing a `ConnectionHandler` across factories, I suspect that's wrong and they're meant to be one-to-one.

clokep:
> Something that's unclear to me -- we seem to be (sometimes?) sharing a `ConnectionHandler` across factories, I suspect that's wrong and they're meant to be one-to-one.

@erikjohnston Any thoughts on this? Is `ConnectionHandler` meant to be per-factory or a singleton used everywhere? We seem to be doing weird things with this.

erikjohnston:
What do you mean by sharing?

> > Something that's unclear to me -- we seem to be (sometimes?) sharing a `ConnectionHandler` across factories, I suspect that's wrong and they're meant to be one-to-one.
> 
> @erikjohnston Any thoughts on this? Is `ConnectionHandler` meant to be per-factory or a singleton used everywhere? We seem to be doing weird things with this.

Where are you seeing that? I'm not seeing anywhere in our code that we instansiate `ConnectionHandler`, we just pass around the type?

clokep:
> Where are you seeing that? I'm not seeing anywhere in our code that we instansiate `ConnectionHandler`, we just pass around the type?

It is a bit confusing, but `RedisDirectTcpReplicationClientFactory` ends with two `ConnectionHandlers`, one of which might be unused (?):

https://github.com/matrix-org/synapse/blob/d0e78af35e519ff76bd23e786007f3e7130d90f7/synapse/replication/tcp/redis.py#L315-L331

The call to `super().__Init__(...)` (which goes to txredis) instantiates a handler:

https://github.com/IlyaSkriblovsky/txredisapi/blob/c89a49829d1c48baa10d03c41d39c261c02ad6e1/txredisapi.py#L2289

But we also store one as `synapse_outbound_redis_connection`. This then gets stored onto each `RedisSubscriber`, but `RedisSubscriber` seems to only ever use `synapse_outbound_redis_connection`. Mostly I'm just confused at what happens to the other handler, does it just sit around doing nothing? It seems to get passed into a `Deferred` that we don't seem to care about?

erikjohnston:
Ah right, yes. So at any given time there are *two* connections to Redis. One is dedicated to receiving pub/sub messages (i.e. the connection that does `SUBSCRIBE`, which takes over the connection), and one for sending all other commands. The `synapse_outbound_redis_connection` is the latter, basically, and so gets passed around to anywhere that might need to send a command to Redis.

clokep:
> Ah right, yes. So at any given time there are _two_ connections to Redis. One is dedicated to receiving pub/sub messages (i.e. the connection that does `SUBSCRIBE`, which takes over the connection), and one for sending all other commands. The `synapse_outbound_redis_connection` is the latter, basically, and so gets passed around to anywhere that might need to send a command to Redis.

Right! Well, that's confusing, but regardless I think this PR makes the type hints more accurate?

erikjohnston (approved):

PR #12049, opened 2022-02-21
WIP: manage deps with poetry

DMRobertson:
WIP


DMRobertson:
- complement job is unhappy because of the open registration change.
- twisted trunk needs fixing, need to consult with @squahtx who I think might have already fixed this elsewhere
- debian packages grrr. We could probably land this without those though.
- need to pull in `black` and `jinja2` dep fun
- need to squash the hell out of this and tidy up commit messages

DMRobertson:
Right, this has served its purpose. Time to start afresh.

PR #12052, opened 2022-02-21
Move isort config to `pyproject.toml`

DMRobertson:
To avoid confusion with setuptools, I'd like to move this config into pyproject.toml. (I don't think this is strictly required for #11537, but I think it's slightly neater.)

### DIFF
--- a/changelog.d/12052.misc
+++ b/changelog.d/12052.misc
@@ -1,0 +1,1 @@
+Move `isort` configuration to `pyproject.toml`.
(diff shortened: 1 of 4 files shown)


squahtx (approved):
lgtm, not sure why sytest is failing

DMRobertson:
> lgtm, not sure why sytest is failing

I think it's just waiting for sytest#1192. Rich has the conn there.

PR #12059, opened 2022-02-22
Prune setup.cfg some more

DMRobertson:
Similar to #12052: I want to move stuff away from setup.cfg where possible. However, this time I have an additional motivation. I've only been able to get `tox` and `poetry` to cooperate if `setup.cfg` and `setup.py` are nowhere to be seen. Otherwise, when `tox` invokes `pip`, the latter guesses we're in a setuptools project and tries to run `setup.py develop`, which fails.

Merging this will mean that only `check_manifest` config remains in setup.cfg. That isn't needed for poetry (it doesn't use a manifest file, but config in pyproject.toml). So we will soon be able to remove setup.cfg altogether.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,11 @@
+# TODO: incorporate this into pyproject.toml if flake8 supports it in the future.
+# See https://github.com/PyCQA/flake8/issues/234
+[flake8]
+# see https://pycodestyle.readthedocs.io/en/latest/intro.html#error-codes
+# for error codes. The ones we ignore are:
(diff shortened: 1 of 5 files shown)


squahtx (approved):
LGTM!
Not sure what the trial section is for, but if CI passes I'm happy

DMRobertson:
> LGTM! Not sure what the trial section is for, but if CI passes I'm happy

Think I wrote something in the commit message, but TL;DR I don't think it has any effect. Let's see what CI thinks though, as you say

PR #12062, opened 2022-02-22
Add module callbacks called for reacting to deactivation and profile update

babolivier:
Part of the Tchap mainlining work.

Adds 2 new module callbacks: `on_profile_update` and `on_deactivation`, which are called after a profile update and a deactivation, respectively.

### DIFF
--- a/changelog.d/12062.feature
+++ b/changelog.d/12062.feature
@@ -1,1 +1,1 @@
-Add module callbacks to react to user deactivations and profile updates.
+Add module callbacks to react to user deactivation status changes (i.e. deactivations and reactivations) and profile updates.
--- a/changelog.d/12062.feature
+++ b/changelog.d/12062.feature
@@ -1,0 +1,1 @@
+Add module callbacks to react to user deactivations and profile udpates.
--- a/changelog.d/12062.feature
+++ b/changelog.d/12062.feature
@@ -1,1 +1,1 @@
-Add module callbacks to react to user deactivations and profile udpates.
+Add module callbacks to react to user deactivations and profile updates.
--- a/docs/modules/third_party_rules_callbacks.md
+++ b/docs/modules/third_party_rules_callbacks.md
@@ -150,7 +150,7 @@ If multiple modules implement this callback, Synapse runs them all in order.
 
 ### `on_profile_update`
 
-_First introduced in Synapse v1.5X.0_
+_First introduced in Synapse v1.54.0_
 
 ```python
 async def on_profile_update(
@@ -185,7 +185,7 @@ If multiple modules implement this callback, Synapse runs them all in order.
 
 ### `on_user_deactivation_status_changed`
 
-_First introduced in Synapse v1.5X.0_
+_First introduced in Synapse v1.54.0_
 
 ```python
 async def on_user_deactivation_status_changed(
--- a/docs/modules/third_party_rules_callbacks.md
+++ b/docs/modules/third_party_rules_callbacks.md
@@ -166,8 +166,8 @@ deactivated (in which case their display name is set to an empty string (`""`) a
 avatar URL is set to `None`). The module is passed the Matrix ID of the user whose profile
 has been updated, their new profile, as well as a boolean that is `True` if the update
 was triggered by a server admin (and `False` otherwise). Note that this boolean is also
-`True` if the profile change happens as a result of the user logging through Single
-Sing-On, and if a server admin updates their own profile.
+`True` if the profile change happens as a result of the user logging in through Single
+Sign-On, or if a server admin updates their own profile.
 
 If multiple modules implement this callback, Synapse runs them all in order.
 
--- a/docs/modules/third_party_rules_callbacks.md
+++ b/docs/modules/third_party_rules_callbacks.md
@@ -165,15 +165,17 @@ Called after updating a local user's profile. The update can be triggered either
 user themselves or a server admin. The update can also be triggered by a user being
 deactivated (in which case their display name is set to an empty string (`""`) and the
 avatar URL is set to `None`). The module is passed the Matrix ID of the user whose profile
-has been updated, their new profile, as well as a boolean that is `True` if the update
-was triggered by a server admin (and `False` otherwise), and a boolean that is `True` if
-the update is a result of the user being deactivated. Note that the `by_admin` boolean is
-also `True` if the profile change happens as a result of the user logging in through
-Single Sign-On, or if a server admin updates their own profile.
+has been updated, their new profile, as well as a `by_admin` boolean that is `True` if the
+update was triggered by a server admin (and `False` otherwise), and a `deactivated`
+boolean that is `True` if the update is a result of the user being deactivated.
+
+Note that the `by_admin` boolean is also `True` if the profile change happens as a result
+of the user logging in through Single Sign-On, or if a server admin updates their own
+profile.
 
 Per-room profile changes do not trigger this callback to be called. Synapse administrators
 wishing this callback to be called on every profile change are encouraged to disable
-per-room profile globally using the `allow_per_room_profiles` configuration setting in
+per-room profiles globally using the `allow_per_room_profiles` configuration setting in
 Synapse's configuration file.
 This callback is not called when registering a user, even when setting it through the
 [`get_displayname_for_registration`](https://matrix-org.github.io/synapse/latest/modules/password_auth_provider_callbacks.html#get_displayname_for_registration)
@@ -181,18 +183,24 @@ module callback.
 
 If multiple modules implement this callback, Synapse runs them all in order.
 
-### `on_deactivation`
+### `on_user_deactivation_status_changed`
 
 _First introduced in Synapse v1.5X.0_
 
 ```python
-async def on_deactivation(user_id: str, by_admin: bool) -> None:
+async def on_user_deactivation_status_changed(
+    user_id: str, deactivated: bool, by_admin: bool
+) -> None:
 ```
 
-Called after deactivating a local user. The deactivation can be triggered either by the
-user themselves or a server admin. The module is passed the Matrix ID of the user that's
-been deactivated, as well as a boolean that is `True` if the deactivation was triggered
-by a server admin (and `False` otherwise).
+Called after deactivating a local user, or reactivating them through the admin API. The
+deactivation can be triggered either by the user themselves or a server admin. The module
+is passed the Matrix ID of the user whose status is changed, as well as a `deactivated`
+boolean that is `True` if the user is being deactivated and `False` if they're being
+reactivated, and a `by_admin` boolean that is `True` if the deactivation was triggered by
+a server admin (and `False` otherwise). This latter `by_admin` boolean is always `True`
+if the user is being reactivated, as this operation can only be performed through the
+admin API.
 
 If multiple modules implement this callback, Synapse runs them all in order.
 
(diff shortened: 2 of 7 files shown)


docs/modules/third_party_rules_callbacks.md
```diff
@@ -148,6 +148,44 @@ deny an incoming event, see [`check_event_for_spam`](spam_checker_callbacks.md#c
...
+```
+
+Called after updating a local user's profile. The update can be triggered either by the
+user themselves or a server admin. The update can also be triggered by a user being
+deactivated (in which case their display name is set to an empty string (`""`) and the
+avatar URL is set to `None`). The module is passed the Matrix ID of the user whose profile
+has been updated, their new profile, as well as a boolean that is `True` if the update
+was triggered by a server admin (and `False` otherwise). Note that this boolean is also
+`True` if the profile change happens as a result of the user logging through Single
+Sing-On, and if a server admin updates their own profile.
```
anoadragon453, line 170:
```suggestion
`True` if the profile change happens as a result of the user logging in through Single
Sign-On, or if a server admin updates their own profile.
```

docs/modules/third_party_rules_callbacks.md
```diff
@@ -148,6 +148,44 @@ deny an incoming event, see [`check_event_for_spam`](spam_checker_callbacks.md#c
 
 If multiple modules implement this callback, Synapse runs them all in order.
 
+### `on_profile_update`
```
anoadragon453, line 151: Is it possible to distinguish in this method whether a user was deactivated versus simply setting their profile information to a display name of `""` and an avatar URL of `None`? I could see implementations of this method commonly having a conditional of:

```python
if new_profile.display_name == "" and new_profile.avatar_url is None:
    return
```

in which case users would be able to trigger a profile update without this code catching it.

Can we track whether a profile update is the result of a deactivation and refrain from running this method in that instance?

docs/modules/third_party_rules_callbacks.md
```diff
@@ -148,6 +148,44 @@ deny an incoming event, see [`check_event_for_spam`](spam_checker_callbacks.md#c
...
+
+```python
+async def on_profile_update(
+    user_id: str,
+    new_profile: "synapse.module_api.ProfileInfo",
+    by_admin: bool,
+) -> None:
+```
+
+Called after updating a local user's profile. The update can be triggered either by the
```
anoadragon453, line 163: Could you also note here that per-room profile updates do not trigger this callback?

synapse/events/third_party_rules.py
```diff
@@ -370,3 +380,37 @@ async def _get_state_map_for_room(self, room_id: str) -> StateMap[EventBase]:
...
+                logger.exception(
+                    "Failed to run module API callback %s: %s", callback, e
+                )
+
+    async def on_deactivation(self, user_id: str, by_admin: bool) -> None:
+        """Called after a user has been deactivated.
+
+        Args:
+            user_id: The deactivated user.
+            by_admin: Whether the deactivation is performed by a server admin.
```
anoadragon453, line 408:
```suggestion
            by_admin: Whether the deactivation was performed by a server admin.
```

as the rest of the docstring is in past tense.

docs/modules/third_party_rules_callbacks.md
```diff
@@ -148,6 +148,44 @@ deny an incoming event, see [`check_event_for_spam`](spam_checker_callbacks.md#c
...
+Sing-On, and if a server admin updates their own profile.
+
+If multiple modules implement this callback, Synapse runs them all in order.
+
+### `on_deactivation`
+
+_First introduced in Synapse v1.5X.0_
+
+```python
+async def on_deactivation(user_id: str, by_admin: bool) -> None:
```
anoadragon453, line 179: Note that we officially support re-activating users, which currently modules would not be told about: https://github.com/matrix-org/synapse/blob/18862f20b5495bdc556c54e92fd4b1efdc718ba7/synapse/rest/admin/users.py#L333-L335

It might be worthwhile to modify this callback to add support for information modules about re-activated users?

docs/modules/third_party_rules_callbacks.md
```diff
@@ -148,6 +148,44 @@ deny an incoming event, see [`check_event_for_spam`](spam_checker_callbacks.md#c
...
+    user_id: str,
+    new_profile: "synapse.module_api.ProfileInfo",
+    by_admin: bool,
+) -> None:
+```
+
+Called after updating a local user's profile. The update can be triggered either by the
+user themselves or a server admin. The update can also be triggered by a user being
+deactivated (in which case their display name is set to an empty string (`""`) and the
+avatar URL is set to `None`). The module is passed the Matrix ID of the user whose profile
```
anoadragon453, line 166: Could you note that this method will not be called when a user sets their global profile information for the first time upon registration?

tests/rest/client/test_third_party_rules.py
```diff
@@ -530,3 +532,170 @@ def _update_power_levels(self, event_default: int = 0):
...
+        profile_info = args[1]
+        self.assertEqual(profile_info.display_name, displayname)
+        self.assertEqual(profile_info.avatar_url, avatar_url)
+
+    def test_on_profile_update_admin(self):
+        """Tests that the on_profile_update module callback is correctly called on
+        profile update triggered by a server admin.
+        """
+        displayname = "Foo"
+        avatar_url = "mxc://matrix.org/oWQDvfewxmlRaRCkVbfetyEo"
```
anoadragon453, line 596: lol

tests/rest/client/test_third_party_rules.py
```diff
@@ -530,3 +532,170 @@ def _update_power_levels(self, event_default: int = 0):
...
+        self.assertFalse(args[2])
+
+        # Check that we've got the right profile data.
+        profile_info = args[1]
+        self.assertEqual(profile_info.display_name, displayname)
+        self.assertEqual(profile_info.avatar_url, avatar_url)
+
+    def test_on_profile_update_admin(self):
+        """Tests that the on_profile_update module callback is correctly called on
+        profile update triggered by a server admin.
```
anoadragon453, line 593:
```suggestion
        """Tests that the on_profile_update module callback is correctly called on
        profile updates triggered by a server admin.
```

tests/rest/client/test_third_party_rules.py
```diff
@@ -530,3 +532,170 @@ def _update_power_levels(self, event_default: int = 0):
             },
             tok=self.tok,
         )
+
+    def test_on_profile_update(self):
+        """Tests that the on_profile_update module callback is correctly called on
+        profile update.
```
anoadragon453, line 538:
```suggestion
        """Tests that the on_profile_update module callback is correctly called on
        profile updates.
```

changelog.d/12062.feature
```diff
@@ -0,0 +1 @@
+Add module callbacks to react to user deactivations and profile udpates.
```
anoadragon453, line 1:
```suggestion
Add module callbacks to react to user deactivations and profile updates.
```

anoadragon453 (changes requested):
Looks pretty good on the whole! Some notes and questions below.

docs/modules/third_party_rules_callbacks.md
babolivier: > Can we track whether a profile update is the result of a deactivation and refrain from running this method in that instance?

I don't think we should skip this method in this case. It sounds to me like a way to get modules out of sync with a user's profile (e.g. if they maintain their own db table for profiles, and the user gets reactivated and changes just their display name, the module will still think they've got their old avatar).
We could pass a flag onto the profile handler then to the callback. The initial intention was to have modules react to deactivation separately in `on_deactivation`, but I guess it doesn't hurt to have this information available there as well.

docs/modules/third_party_rules_callbacks.md
babolivier: I wonder if it's not better to add a `on_reactivation` callback in the future (when we end up needing one)? It feels more like the right way to convey this information to modules.

docs/modules/third_party_rules_callbacks.md
babolivier: (done in https://github.com/matrix-org/synapse/pull/12062/commits/698afcf29a50f644423d587bbb04923249ecd9db)

docs/modules/third_party_rules_callbacks.md
anoadragon453: Compiling both into one method would have the advantage of reducing the number of module callbacks we have, while potentially having the downside of a slightly more confusing function signature (`by_admin` would always be `True` on reactivation).

So I suppose it depends on how much value we put on keeping the callback surface small. If not much, then having both `on_deactivation` and `on_reactivation` methods are fine.

Small nit: `on_deactivation` -> `on_user_deactivation`

docs/modules/third_party_rules_callbacks.md
```diff
@@ -148,6 +148,54 @@ deny an incoming event, see [`check_event_for_spam`](spam_checker_callbacks.md#c
...
+
+_First introduced in Synapse v1.5X.0_
+
+```python
+async def on_deactivation(user_id: str, by_admin: bool) -> None:
+```
+
+Called after deactivating a local user. The deactivation can be triggered either by the
+user themselves or a server admin. The module is passed the Matrix ID of the user that's
+been deactivated, as well as a boolean that is `True` if the deactivation was triggered
```
anoadragon453, line 194:
```suggestion
been deactivated, as well as a boolean `by_admin` that is `True` if the deactivation was triggered
```

babolivier:
@anoadragon453 Re-requesting a review despite previous approval since, as discussed in DMs, I've ended up changing the callback for deactivation to also include reactivation.

changelog.d/12062.feature
```diff
@@ -0,0 +1 @@
+Add module callbacks to react to user deactivation status changes (i.e. deactivations and reactivations) and profile updates.
```
anoadragon453, line 1: Would it be worth linking to the documentation for these new methods here?

anoadragon453 (approved):

docs/modules/third_party_rules_callbacks.md
```diff
@@ -148,6 +148,62 @@ deny an incoming event, see [`check_event_for_spam`](spam_checker_callbacks.md#c
 
 If multiple modules implement this callback, Synapse runs them all in order.
 
+### `on_profile_update`
+
+_First introduced in Synapse v1.5X.0_
```
babolivier, line 153:
```suggestion
_First introduced in Synapse v1.54.0_
```

docs/modules/third_party_rules_callbacks.md
```diff
@@ -148,6 +148,62 @@ deny an incoming event, see [`check_event_for_spam`](spam_checker_callbacks.md#c
...
+Synapse's configuration file.
+This callback is not called when registering a user, even when setting it through the
+[`get_displayname_for_registration`](https://matrix-org.github.io/synapse/latest/modules/password_auth_provider_callbacks.html#get_displayname_for_registration)
+module callback.
+
+If multiple modules implement this callback, Synapse runs them all in order.
+
+### `on_user_deactivation_status_changed`
+
+_First introduced in Synapse v1.5X.0_
```
babolivier, line 188:
```suggestion
_First introduced in Synapse v1.54.0_
```

changelog.d/12062.feature
anoadragon453: It might be useful if module developers want to start using the new methods? We don't name them here either.

PR #12067, opened 2022-02-23
Remove more references to `get_datastore`

richvdh:
These were added in #12001, and were snuck in since #12031 was started.

Also a couple of other cleanups while we're in the area.


reivilibre (approved):
CI OK ⇒ LGTM

PR #12073, opened 2022-02-24
Remove the unstable `/spaces` endpoint.

clokep:
This is phase 2 of #10946 -- it removes the unstable `/spaces` endpoint (both Client-Server and Server-Server versions) and various supporting code for it. It also updates bits of the underlying code for backwards compatibility.

Note that the *unstable* `/hierarchy` endpoint (which replaced this) landed in Synapse v1.41.0 and per [some research](https://github.com/matrix-org/synapse/issues/10946#issuecomment-1049408858) seems well supported in clients.


synapse/handlers/room_summary.py
```diff
@@ -958,9 +660,8 @@ async def _is_remote_room_accessible(
         ):
             return True
 
-        # Check if the user is a member of any of the allowed spaces
-        # from the response.
-        allowed_rooms = room.get("allowed_room_ids") or room.get("allowed_spaces")
```
clokep, line 963: I wonder if we should leave this longer since I'm pretty sure a response over federation from an older server will still have this?

synapse/federation/federation_client.py
```diff
@@ -1358,61 +1358,6 @@ async def get_room_complexity(
...
-        Returns:
-            a parsed FederationSpaceSummaryResult
-
-        Raises:
-            SynapseError if we were unable to get a valid summary from any of the
-               remote servers
-        """
-
-        async def send_request(destination: str) -> FederationSpaceSummaryResult:
-            res = await self.transport_layer.get_space_summary(
```
squahtx, line 1396: Can we remove `TransportLayerClient.get_space_summary()` too?

synapse/federation/federation_client.py
```diff
@@ -1706,89 +1599,34 @@ def from_json_dict(cls, d: JsonDict) -> "TimestampToEventResponse":
...
+    content = d.get("content")
+    if not isinstance(content, dict):
+        raise ValueError("Invalid event: 'content' must be a dict")
 
-        return cls(rooms, parsed_events)
+    via = content.get("via")
+    if not isinstance(via, Sequence):
+        raise ValueError("Invalid event: 'via' must be a list")
+    if any(not isinstance(v, str) for v in via):
+        raise ValueError("Invalid event: 'via' must be a list of strings")
```
squahtx, line 1632: Not something that I'd expect to be fixed in this PR but strings do count as `Sequence[str]`s, so we need to be a bit more specific in the first `isinstance` check.

synapse/handlers/room_summary.py
squahtx: Looks like this impacts Synapse 1.35.0 to 1.40.0, inclusive?
Personally I'd prefer to get the removal over and done with. `allowed_spaces` does come from an _unstable_ implementation after all.

clokep: That seems reasonable to me, yes.

synapse/federation/federation_client.py
clokep: Good point, I can do a follow-up for that. 👍  Seems like this has been wrong since the code was introduced in #9653.

squahtx (approved):
LGTM!

synapse/federation/federation_client.py
clokep: Thanks for filing the follow-up: #12123.

Issue #11694, opened 2022-01-06
Viewing room list in a Space / Searching for rooms in a Space is very slow


reivilibre:
Still need to find out what's going on here or even what the related requests are, but can't see an issue about it and this is surely something we ought to resolve if we want people to enjoy using spaces.

### Description

The 'Explore Space' screen in Element is very slow, even if typing in a search term.

### Steps to reproduce

- go to a (decent size?) space: my example is Element's space
    - I suspect this issue may not be visible from the matrix.org homeserver since I expect someone would have said something by now!
- open the 'Explore rooms' view in Element
- type in a *keyword*
- watch spinner for 2 minutes (in my case; probably varies depending upon how far down the room is in the tree?)
- a result appears (spinner is still going, so it's probably not a complete result set. However this was the room I wanted in my situation so I could finish here.)

I would expect searching to be much faster than that; let's say 10 seconds or less to be generous given that I suppose I can understand that it may be making a few round trips to remote homeservers.

### Version information

- **Homeserver**: librepush.net
- **Version**: 1.49.0, but this issue has been here for as long as I can remember the feature existing
- **Install method**: pip from PyPI
- **Platform**: amd64 VPS

clokep:
> I would expect searching to be much faster than that; let's say 10 seconds or less to be generous given that I suppose I can understand that it may be making a few round trips to remote homeservers.

Searching is done client side, so your client is paginating through the entire space finding all results. I Don't know if they're shown as they get matched or at the end though.

See #10523 also for performance with this API.

reivilibre:
I got some time to look into this... Here are some notes

## Client-side view of what's going on

The slow request seems to be this endpoint (with some example parameters):

`GET https://matrix.librepush.net/_matrix/client/unstable/org.matrix.msc2946/rooms/!OJBlkJuUrsKnqtNnTi%3Amatrix.org/hierarchy?suggested_only=false&from=GiSyKxltKfZBfrhSVEXQVjUb&limit=20`

Waiting: 9.43 sec.

Each one only includes a few results, so it makes scrolling through the list quite slow and frustrating. You get a drip of a few new rooms every 10 seconds. It's not great but at least it feels less dead?
**Search is performed client-side as a filter, performing back-to-back requests if needed.** This really exacerbates the problem when performing a search. *This is probably quite poor even if the requests weren't so slow. This is the equivalent of a table scan to search the space, but with high latency and tiny (20) block sizes...!*


## Synapse-side view of what's going on

[...]

PR #12077, opened 2022-02-24
Properly failover for unknown endpoints from Conduit/Dendrite

clokep:
Related to #11694.

There was a bug in the failover logic for finding "unknown endpoints" in that it considered any 404 an unknown endpoint. This isn't always accurate though since some federation endpoints actually return a 404 for missing objects.

Note that I wrote [MSC3743](https://github.com/matrix-org/matrix-doc/pull/3743) to make this behavior less...annoying to deal with. That MSC has some information about where the bodies of the 404s come from (pretty much Conduit and Dendrite).

### DIFF
--- a/changelog.d/12077.bugfix
+++ b/changelog.d/12077.bugfix
@@ -1,0 +1,1 @@
+Fix a long-standing bug where Synapse would make additional failing requests over federation for missing data.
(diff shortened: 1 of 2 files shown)


squahtx (approved):
lgtm

PR #12088, opened 2022-02-25
Use importlib.metadata to read requirements

DMRobertson:
Part of #11537. We will have to use `importlib.metadata` to read our dependencies at runtime once we have moved to a declarative config. I do this now to ease the transition.

Before this change, python_dependencies has four confused purposes:

1. setup.py evaluates this to define matrix-synapse's dependencies and optional
   dependencies.
2. The twisted olddeps CI job edit this file in-place so that we install the oldest
   version of all dependencies. Similarly, the twisted trunk CI job patches this file
   to use twisted trunk from github.
3. This data is made available at runtime for the application to check dependencies
   are present and correctly versioned.
4. Finally, the script can be executed directly to print out all dependencies
   (though this lacks the extras grouping information). This is mentioned in the
   upgrade notes for 0.7.0 and nowhere else.

Purposes (1) and (2) have no right belonging in application code. I am relocating and reimplementing purpose (3) here. I propose nuking purpose (4) as well while we're at it---again, this is not something the application source code should be concerned with.

I've tested this with some dirty mocked test cases, and also:

```
$ pip install 'twisted<18.9' 'twisted>16'
Collecting twisted<18.9
  Downloading Twisted-18.7.0.tar.bz2 (3.1 MB)
     |████████████████████████████████|
 ...
    Running setup.py install for twisted ... done
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
matrix-synapse 1.53.0 requires Twisted>=18.9.0, but you have twisted 18.7.0 which is incompatible.
Successfully installed PyHamcrest-2.0.3 twisted-18.7.0

$ python -m synapse.app
ERROR:root:Need Twisted>=18.9.0, but got Twisted==18.7.0
Missing Requirements: "Twisted"
To install run:
    pip install --upgrade --force "Twisted"
```

### DIFF
--- a/synapse/util/check_dependencies.py
+++ b/synapse/util/check_dependencies.py
@@ -105,7 +105,7 @@ def check_requirements(extra: Optional[str] = None) -> None:
     else:
         raise ValueError(f"Synapse does not provide the feature '{extra}'")
 
-    deps_needed = []
+    deps_unfulfilled = []
     errors = []
 
     for (requirement, must_be_installed) in dependencies:
@@ -113,15 +113,15 @@ def check_requirements(extra: Optional[str] = None) -> None:
             dist: metadata.Distribution = metadata.distribution(requirement.name)
         except metadata.PackageNotFoundError:
             if must_be_installed:
-                deps_needed.append(requirement.name)
+                deps_unfulfilled.append(requirement.name)
                 errors.append(_not_installed(requirement, extra))
         else:
             if not requirement.specifier.contains(dist.version):
-                deps_needed.append(requirement.name)
+                deps_unfulfilled.append(requirement.name)
                 errors.append(_incorrect_version(requirement, dist.version, extra))
 
-    if deps_needed:
+    if deps_unfulfilled:
         for err in errors:
             logging.error(err)
 
-        raise DependencyException(deps_needed)
+        raise DependencyException(deps_unfulfilled)
--- a/synapse/util/check_dependencies.py
+++ b/synapse/util/check_dependencies.py
@@ -1,0 +1,107 @@
+import logging
+from typing import List
+
+from pkg_resources import (
+    DistributionNotFound,
+    Requirement,
+    VersionConflict,
+    get_provider,
+)
+
+from synapse.python_dependencies import CONDITIONAL_REQUIREMENTS, REQUIREMENTS
+
+
+class DependencyException(Exception):
+    @property
+    def message(self):
+        return "\n".join(
+            [
+                "Missing Requirements: %s" % (", ".join(self.dependencies),),
+                "To install run:",
+                "    pip install --upgrade --force %s" % (" ".join(self.dependencies),),
+                "",
+            ]
+        )
+
+    @property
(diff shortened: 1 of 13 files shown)


DMRobertson:
Sytest is not happy. Grr.

DMRobertson:
#12089 seems to have hit the same set of sytest failures(?)

anoadragon453:
Sytest failures are affecting develop as well, yes: https://github.com/matrix-org/synapse/runs/5356507389?check_suite_focus=true

tests/util/test_check_dependencies.py
```diff
@@ -0,0 +1,95 @@
...
+        raise NotImplementedError()
+
+    def read_text(self, filename):
+        raise NotImplementedError()
+
+
+old = DummyDistribution("0.1.2")
+new = DummyDistribution("1.2.3")
+
+# could probably use stdlib TestCase --- no need for twisted here
```
squahtx, line 32: I've got no strong opinions on which `TestCase` class we use.

squahtx (approved):
LGTM!

Just to confirm, this PR only touches purpose (3)?

synapse/util/check_dependencies.py
```diff
@@ -29,79 +27,103 @@ def dependencies(self):
...
+    elif extra in EXTRAS:
+        dependencies = _dependencies_for_extra(extra)
+    else:
+        raise ValueError(f"Synapse does not provide the feature '{extra}'")
 
-    # first check if the markers specify that this requirement needs installing
-    if req.marker is not None and not req.marker.evaluate():
-        # not required for this environment
-        return
+    deps_needed = []
```
squahtx, line 110: nit: `needed` is a tad misleading. `deps_unfulfilled` maybe?

DMRobertson: Can do. This came from the original implementation, but finding the right words for all of this stuff is tricky.

synapse/util/check_dependencies.py
```diff
@@ -0,0 +1,129 @@
...
+    def dependencies(self) -> Iterable[str]:
+        for i in self.args[0]:
+            yield '"' + i + '"'
+
+
+EXTRAS = {
+    v
+    for k, v in metadata.distribution("matrix-synapse").metadata.items()
+    if k == "Provides-Extra"
+}
```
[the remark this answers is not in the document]
DMRobertson: That seems neater---will check this out!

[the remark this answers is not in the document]
DMRobertson: Seems kosher to me: https://importlib-metadata.readthedocs.io/en/latest/api.html#importlib_metadata.PackageMetadata.get_all

squahtx (approved):

DMRobertson:
Needs a linter script run, but I'll put it to auto-merge after that.

DMRobertson:
This change didn't mark `importlib_metadata` as a requirement! We are fortunate that the backport is pulled in by other dependencies.

PR #12089, opened 2022-02-25
Fix 'Unhandled error in Deferred'

richvdh:
Fixes a CRITICAL "Unhandled error in Deferred" log message which happened when a function wrapped with `@cachedList` failed.

There's a couple of other tweaks here - review commit by commit.

### DIFF
--- a/synapse/util/caches/descriptors.py
+++ b/synapse/util/caches/descriptors.py
@@ -346,10 +346,10 @@ class DeferredCacheListDescriptor(_CacheDescriptorBase):
     """Wraps an existing cache to support bulk fetching of keys.
 
     Given an iterable of keys it looks in the cache to find any hits, then passes
-    the tuple of missing keys to the wrapped function.
+    the set of missing keys to the wrapped function.
 
-    Once wrapped, the function returns a Deferred which resolves to the list
-    of results.
+    Once wrapped, the function returns a Deferred which resolves to a Dict mapping from
+    input key to output value.
     """
 
     def __init__(
@@ -452,14 +452,14 @@ class DeferredCacheListDescriptor(_CacheDescriptorBase):
                     # the wrapped function has completed. It returns a dict.
                     # We can now update our own result map, and then resolve the
                     # observable deferreds in the cache.
-                    for e in missing:
+                    for e, d1 in deferreds_map.items():
                         val = res.get(e, None)
                         # make sure we update the results map before running the
                         # deferreds, because as soon as we run the last deferred, the
                         # gatherResults() below will complete and return the result
                         # dict to our caller.
                         results[e] = val
-                        deferreds_map[e].callback(val)
+                        d1.callback(val)
 
                 def errback_all(f: Failure) -> None:
                     # the wrapped function has failed. Propagate the failure into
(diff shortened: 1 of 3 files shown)


synapse/util/caches/descriptors.py
```diff
@@ -455,28 +457,23 @@ def complete_all(res: Dict[Hashable, Any]) -> None:
                         deferreds_map[e].callback(val)
                         results[e] = val
 
-                def errback(f: Failure) -> Failure:
-                    # the wrapped function has failed. Invalidate any cache
-                    # entries we're supposed to be populating, and fail
-                    # their deferreds.
-                    for e in missing:
-                        key = arg_to_cache_key(e)
-                        cache.invalidate(key)
```
richvdh, line 464: this seems to have been redundant (since the cache will invalidate the entry itself when we `errback` the deferred), so I have removed it

synapse/util/caches/descriptors.py
```diff
@@ -444,7 +444,9 @@ def arg_to_cache_key(arg: Hashable) -> Hashable:
                     deferred: "defer.Deferred[Any]" = defer.Deferred()
                     deferreds_map[arg] = deferred
                     key = arg_to_cache_key(arg)
-                    cache.set(key, deferred, callback=invalidate_callback)
```
richvdh, line 447: this was the source of the problem.

`cache.set` creates a new observer, which we were then throwing away. So when the error happened, the failure would be propagated into the observer, but there was nothing to handle it.

One option was to change `cache.set` to avoid creating an observed, but that gets fiddly because it makes the other call to `cache.set` inconvenient/less efficient.

So instead we change things round and add the resulting observer to the list of deferreds we wait for down at line 479.

synapse/util/caches/descriptors.py
```diff
@@ -444,39 +444,40 @@ def arg_to_cache_key(arg: Hashable) -> Hashable:
...
                 def complete_all(res: Dict[Hashable, Any]) -> None:
-                    # the wrapped function has completed. It returns a
-                    # a dict. We can now resolve the observable deferreds in
-                    # the cache and update our own result map.
+                    # the wrapped function has completed. It returns a dict.
+                    # We can now update our own result map, and then resolve the
+                    # observable deferreds in the cache.
                     for e in missing:
                         val = res.get(e, None)
-                        deferreds_map[e].callback(val)
```
richvdh, line 455: per the comment, having to move this to after the update to `results`, since we no longer wait for all the calls to `complete_all` to complete at line 477

synapse/util/caches/descriptors.py
```diff
@@ -346,10 +346,10 @@ class DeferredCacheListDescriptor(_CacheDescriptorBase):
     """Wraps an existing cache to support bulk fetching of keys.
 
     Given an iterable of keys it looks in the cache to find any hits, then passes
-    the tuple of missing keys to the wrapped function.
+    the set of missing keys to the wrapped function.
 
-    Once wrapped, the function returns a Deferred which resolves to the list
-    of results.
+    Once wrapped, the function returns a Deferred which resolves to a Dict mapping from
+    input key to output value.
```
richvdh, line 353: this is a fix to the documentation - it always returned a Dict, rather than a list.

erikjohnston (approved):

PR #12090, opened 2022-02-25
Use the proper serialization format when bundling aggregations

### LINKED ISSUE #12085: Change latest_event format in thread bundled relationship
We need to send down `ClientEventWithoutRoomID` rather than `ClientEvent` to be consistent with the event format that comes down `/sync` as [discussed in the thread MSC](https://github.com/matrix-org/matrix-doc/pull/3440/files#r812880226)

clokep:
Use the serialization parameters passed into the function to serialize the latest thread event in bundled aggregations.

This includes a bit of a refactor, not just the fix, it has a few commits:

1. Don't use `kwargs` since it hides type hints. (This is more verbose, but I think the extra protection is worth it?)
2. The actual fix, which is passing the extra parameters to `serialize_event` when bundling the latest thread event.
3. Fixing an incorrect type hint (it is a token *id*, which is an int).
4. Changes to bundled the configuration together into an object instead of passing each parameter separately. (I'm not 100% sure this change makes sense, but I think overall it is a bit nicer? Idk.)

Fixes #12085


synapse/events/utils.py
```diff
@@ -494,7 +509,10 @@ def _inject_bundled_aggregations(
 
             # Don't bundle aggregations as this could recurse forever.
             serialized_latest_event = self.serialize_event(
-                thread.latest_event, time_now, bundle_aggregations=None
+                thread.latest_event,
+                time_now,
+                config=config,
```
clokep, line 514: I don't know if this is correct or if we should only pass in the format, similar to what we do for redactions: https://github.com/matrix-org/synapse/blob/45f45404de2d0c4d68954eddc2dc905e50dfafe9/synapse/events/utils.py#L349-L352

synapse/appservice/api.py
```diff
@@ -321,16 +321,18 @@ def _serialize(
             serialize_event(
```
clokep, line 321: Part of me thinks we should be consistent and use `EventSerializer` here, but it doesn't really matter unless we might bundled aggregations. I dislike that this is two separate methods though.

clokep:
The sytest failures are unrelated, I believe.

synapse/appservice/api.py
reivilibre: sounds like it might be worthwhile as an extra pass if you have the time and it looks sensible as you're doing it?

reivilibre (approved):

synapse/events/utils.py
clokep: This was discussed a bit in #synapse-dev:matrix.org and we agree that things should definitely be consistent. @erikjohnston thinks that:

> Yeah, my natural instinct is that we should propagate the options for both. In practice I doubt it matters tooo much

I'm going to make that change here.

clokep:
@reivilibre Could you take a quick gander at this?

reivilibre (approved):

PR #12091, opened 2022-02-25
Add restrictions by default to open registration in Synapse

### LINKED ISSUE #12002: Have a restrictive default limit on user registration
something identified as being helpful in countering abuse: most servers have no need to allow large numbers of users to register. The default config could be quite restrictive.

H-Shay:
Enabling registration on a homeserver with no verification (i.e. captcha or email verification) is a known cause for federation spam and abuse. This PR adds restrictions to enabling registration by causing Synapse to refuse to start when registration is enabled with no verification, unless the config flag `enable_registration_without_verification` is also enabled. Fixes #12002.


synapse/app/homeserver.py
```diff
@@ -348,6 +348,22 @@ def setup(config_options: List[str]) -> SynapseHomeServer:
...
         synapse.metrics.MIN_TIME_BETWEEN_GCS = config.server.gc_seconds
 
+    if (
+        config.registration.enable_registration
+        and not config.registration.enable_registration_without_verification
+    ):
+        if (
+            not config.captcha.enable_registration_captcha
+            and not config.registration.registrations_require_3pid
+        ):
```
anoadragon453, line 358: We can probably add `config.registration.registration_requires_token` (token authenticated registration) to this list? 

https://matrix-org.github.io/synapse/latest/usage/administration/admin_api/registration_tokens

H-Shay: Sounds fair to me, I'll add it to the list and update the documentation.

H-Shay:
The sytest failures on this PR will be resolved by this change I made to sytest (currently this is just a branch, once this PR is accepted I will PR the sytest fix): https://github.com/matrix-org/sytest/commit/6497e565d99151f8df13f11a10fbe38cd43373f3

H-Shay:
The complement failure should be resolved by the changes here: https://github.com/matrix-org/complement/compare/shay/allow_registration?expand=1
I'll PR this to complement once this PR is accepted.

clokep:
Not a huge deal -- but I would appreciate if disabled this by default in the `demo/start.sh` script! In that script we modify some of the configuration already, so should be easy enough to add another line with the config flag!

erikjohnston:
What's the deal with the CI failures?

squahtx:
CI failures seem to be:
```
[server]:   File "/venv/lib/python3.9/site-packages/synapse/app/homeserver.py", line 361, in setup
[server]:     raise ConfigError(
[server]: synapse.config._base.ConfigError: You have enabled open registration without any verification. This is a known vector forspam and abuse. If you have enabled registration, please add email, capcha, or token-based verification, oruse the config option `enable_registration_without_verification` in conjunction with `enable_verification.
```

We'll want to update sytest to use the new config option.
Also spaces in the error message have gone missing.

DMRobertson:
> The sytest failures on this PR will be resolved by this change I made to sytest (currently this is just a branch, once this PR is accepted I will PR the sytest fix): [matrix-org/sytest@6497e56](https://github.com/matrix-org/sytest/commit/6497e565d99151f8df13f11a10fbe38cd43373f3)

We've set up CI so that synapse PRs will try to run against sytest branches with the same name, and vice versa. Similarly for complement too. If you rename that sytest branch from `shay/allow_regirstration` to `shay/restrict_registration` I think CI will be happier.

H-Shay:
> We've set up CI so that synapse PRs will try to run against sytest branches with the same name, and vice versa. Similarly for complement too

Thanks DMR, that was super helpful to know! The CI is all sorted now, this should be good to go.

synapse/app/homeserver.py
```diff
@@ -348,6 +348,23 @@ def setup(config_options: List[str]) -> SynapseHomeServer:
...
+        if (
+            not config.captcha.enable_registration_captcha
+            and not config.registration.registrations_require_3pid
+            and not config.registration.registration_requires_token
+        ):
+
+            raise ConfigError(
+                "You have enabled open registration without any verification. This is a known vector for "
+                "spam and abuse. If you have enabled registration, please add email, capcha, or token-based verification, or "
+                "use the config option `enable_registration_without_verification` in conjunction with `enable_verification."
```
anoadragon453, line 364:
```suggestion
                "spam and abuse. If you would like to allow public registrations, please consider adding email, captcha, or token-based verification. "
                "Otherwise this check can be removed by setting the `enable_registration_without_verification` config option to `true`."
```

may be a bit clearer.

anoadragon453 (changes requested):
Getting there, thanks for tackling this!

reivilibre:
> Thanks DMR, that was super helpful to know! The CI is all sorted now, this should be good to go.

I think it would be worth opening the SyTest PR now.

Do you know if the SyTest changes even break CI against an unpatched Synapse? I would imagine they don't, so it may make more sense to land that first so that we never have broken CI on the Synapse develop branch whilst waiting for SyTest changes. (I don't see why it would be a big problem to review concurrently either, so no need to worry about that :))

docs/upgrade.md
```diff
@@ -87,6 +87,10 @@ process, for example:
 
 # Upgrading to v1.55.0
 
+## Open registration without verification is now disabled by default
+Synapse will refuse to start if registration is enabled without email, captcha, or token-based verification unless the new config 
+flag `enable_registration_without_verification` is set to "true".
+
 ## `synctl` script has been moved
```
reivilibre, line 94:
```suggestion
## Open registration without verification is now disabled by default

Synapse will refuse to start if registration is enabled without email, captcha, or token-based verification unless the new config 
flag `enable_registration_without_verification` is set to "true".


## `synctl` script has been moved
```

Bah, of trivial importance, but I think we usually try and format the markdown so there's one line between a heading and the paragraph, then 2 lines between the paragraph and the next heading. (This is just so that it's still 'nice enough' to read the raw Markdown, which is (was?) kind of the point of Markdown)

tests/config/test_registration_config.py
```diff
@@ -76,3 +78,19 @@ def test_session_lifetime_must_not_be_exceeded_by_smaller_lifetimes(self):
         HomeServerConfig().parse_config_dict(
             {"session_lifetime": "31m", "refresh_token_lifetime": "31m", **config_dict}
         )
+
+    def test_refuse_to_start_if_open_registration_and_no_verification(self):
```
reivilibre, line 82: It's only config and straight-forward code so it's tempting not to bother, but this is only testing the negative. You may also be tempted to test that startup is able to proceed if both `enable_registration` and `enable_registration_without_verification` are true.

reivilibre (approved):
Tiny stuff, absolutely fine otherwise. Thank you!

tests/config/test_registration_config.py
H-Shay: Hmm I tested this condition locally and it works. Looking at the test though it feels like it's not necessary to add permanently to the tests. If you feel strongly about it I will, though.

H-Shay:
> I think it would be worth opening the SyTest PR now.

Here are the sytest and complement PR's: 
https://github.com/matrix-org/complement/pull/350
https://github.com/matrix-org/sytest/pull/1217

I can hold off on merging this until these are approved and merged.

PR #12092, opened 2022-02-25
Replace assertEquals with assertEqual.

clokep:
I believe Python 3.10 warns about this by default now? Regardless we should use the non-deprecated alias.

This was done with:

```bash
$ sed -i '' "s/assertEquals/assertEqual/g" tests/**/*.py
$ ./scripts-dev/lint.sh
```

(Repeat the sed line for the various aliases at https://docs.python.org/3/library/unittest.html#deprecated-aliases)

### DIFF
--- a/changelog.d/12092.misc
+++ b/changelog.d/12092.misc
@@ -1,0 +1,1 @@
+User `assertEqual` instead of the deprecated `assertEquals` in test code.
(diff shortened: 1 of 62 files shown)


anoadragon453 (approved):
unittest.TestCase.assertEqual(this_pr, "lgtm!")

SyTest failures are unrelated to these changes.

clokep:
I figure this is safe to merge anyway since it is a unit test only change. Thanks for the review!

PR #11890, opened 2022-02-02
Fix losing incoming EDUs if debug logging enabled

DMRobertson:
Missed in #11760.

Fixes #11889. Homeservers should only be affected if the
`synapse.8631_debug` logger was enabled for DEBUG mode.

I am not sure if this merits a bugfix release: I think the logging can
be disabled in config if anyone is affected? But it is still pretty bad.

PR #12098, opened 2022-02-28
Actually fix bad debug logging rejecting device list & signing key transactions

DMRobertson:
Missed in #11890.


tests/federation/transport/test_server.py
```diff
@@ -38,3 +38,19 @@ def test_open_public_room_list_over_federation(self):
             "/_matrix/federation/v1/publicRooms",
         )
         self.assertEqual(200, channel.code)
+
+    @DEBUG
+    def test_edu_debugging_doesnt_explode(self):
+        """Sanity check fed. RX succeeds with `synapse.debug_8631` logging enabled.
```
babolivier, line 44: I don't know what "RX" means, could you update this comment please?

changelog.d/12098.bugfix
```diff
@@ -0,0 +1 @@
+Fix a bug introduced in Synapse 1.51.0rc1 where incoming federation transactions containing at least one EDU would be dropped if debug logging was enabled for `synapse.8631_debug`. Synapse 1.53.0rc1 included a partial fix this ([#11890](https://github.com/matrix-org/synapse/pull/11890)), but the fix was incorrect: EDUs containing device list or signing key updates would still be dropped.
```
babolivier, line 1: Not sure it's worth mentioning #11890 since it didn't end up fixing it?

babolivier (approved):
lgtm otherwise, modulo CI

tests/federation/transport/test_server.py
DMRobertson: TX: transmit
RX: receive

not sure where the abbreviation comes from. Synapse uses the former in `transaction_manager.py` and `per_destination_queue.py`. Slightly surprised to see it doesn't use the latter.

DMRobertson: Possible etymology: https://en.wikipedia.org/wiki/RS-232

changelog.d/12098.bugfix
DMRobertson: I just want to clarify that I've not accidentally copied and pasted a bugfix line from 1.53's release notes.

babolivier: Fair enough, though I'd tend to think looking at the PR is enough to tell that

tests/federation/transport/test_server.py
babolivier: Thanks! fwiw I just read `TX` in those file as an abbreviated version of `txn` for "transaction", TIL TX and RX mean transmit and receive.

babolivier (approved):
Sytest seems to be failing

DMRobertson:
> Sytest seems to be failing

Yes, though in the same way that develop is failing. #12100 is tackling this.

Issue #3440, opened 2018-06-25
Smarter URL preview


ptman:
### Description

URL previews of login pages or 401/403 aren't very useful. It would be nice URL preview could have some logic to skip those. E.g. links to private google docs or sites needing login that aren't blacklisted.

Maybe even previews with transient error HTTP status codes (!= 200?) could have a way of requesting a refresh. With some sort of rate limit to prevent DoS.

jaywink:
![Selection_999(762)](https://user-images.githubusercontent.com/1174866/144430614-3a02c127-57b4-4544-ae16-06c07c29bb09.png)

It sure would be nice to get rid of these totally useless previews when redirected to sign-in. Trouble is it's a standard 302 in this case for example, and there are totally valid cases for a 302 other than "taking you to login".

Half-Shot:
Should synapse attempt to follow a 3XX (for a limited number of hops) and the resulting status checked? If we're not going to follow a 302, then I'd be pretty happy to say those shouldn't be rendered. 

If there is useful content on a 3XX page, then that feels like a standard isn't being followed.

clokep:
We already do [skip previews of non-2xx pages](https://github.com/matrix-org/synapse/blob/626bd75f4847f36747c162348e309b65cc1646b6/synapse/http/client.py#L697-L701) (I think we always have TBH); Synapse [does follow 3xx redirects](https://github.com/matrix-org/synapse/blob/626bd75f4847f36747c162348e309b65cc1646b6/synapse/media/url_previewer.py#L169).

If you can't access something in GitLab it seems to redirect (via a 302) to the signin page, which returns a 200 response. I'll re-title this issue, but I don't think there's much we can do unless there's a standard way to "identify" a login page?

PR #12099, opened 2022-02-28
Move experimental support for MSC3440 to /versions.

clokep:
Per changes to the MSC: [matrix-org/matrix-doc@`e640f6b` (#3440)](https://github.com/matrix-org/matrix-doc/pull/3440/commits/e640f6b95d828076003626d131c51a846be58e05).

### DIFF
--- a/changelog.d/12099.misc
+++ b/changelog.d/12099.misc
@@ -1,0 +1,1 @@
+Move experimental support for [MSC3440](https://github.com/matrix-org/matrix-doc/pull/3440) to /versions.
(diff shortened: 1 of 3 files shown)


anoadragon453 (approved):
The code change looks fine to me.

Just to double-check, will this break any known client implementations of MSC3440?

clokep:
> Just to double-check, will this break any known client implementations of [MSC3440](https://github.com/matrix-org/matrix-doc/pull/3440)

It will cause client implementations to think the server does not have the feature enabled until they are updated. I believe the known client implementations all have the feature behind an experimental flag though and this should be acceptable. I'll double check however.

clokep:
> > Just to double-check, will this break any known client implementations of [MSC3440](https://github.com/matrix-org/matrix-doc/pull/3440)
> 
> It will cause client implementations to think the server does not have the feature enabled until they are updated. I believe the known client implementations all have the feature behind an experimental flag though and this should be acceptable. I'll double check however.

To come back to this -- I think it should be fine since the support is gated behind a server config flag anyway, so it is not disabled by default in Synapse.

PR #12100, opened 2022-02-28
Fix `PushRuleEvaluator` and `Filter` to work on frozendicts

richvdh:
frozendicts do not (necessarily) inherit from dict, so this needs to handle them correctly.

It seems like the Python version of `frozendict` *does* inherit from `dict`, which is why we have gotten away with this for so long. Versions of `frozendict` since 2.2 make it easier to install the C version (there are now prebuilt `manylinux` wheels).

### DIFF
--- a/changelog.d/12100.bugfix
+++ b/changelog.d/12100.bugfix
@@ -1,0 +1,1 @@
+Fix a long-standing bug which could cause push notifications to malfunction if `use_frozen_dicts` was set in the configuration.
--- a/synapse/api/filtering.py
+++ b/synapse/api/filtering.py
@@ -22,6 +22,7 @@ from typing import (
     Dict,
     Iterable,
     List,
+    Mapping,
     Optional,
     Set,
     TypeVar,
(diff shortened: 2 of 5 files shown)


clokep:
Seems to be a regression (by me) from #11098. 😢

richvdh:
> Seems to be a regression (by me) from #11098.

Interesting, I hadn't spotted that that had changed.

Anyway yes, that looks like part of it, but you still need a few other things to go wrong before this bug actually bites. Hopefully the unit test will help here in future.

changelog.d/12100.bugfix
```diff
@@ -0,0 +1 @@
+Fix a long-standing bug which could cause push notifications to malfunction if `use_frozen_dicts` was set in the configuration.
```
squahtx, line 1:
```suggestion
Fix a bug introduced in 1.47.0 which could cause push notifications to malfunction if `use_frozen_dicts` was set in the configuration.
```

squahtx (approved):
lgtm

clokep:
(And the second one seems to be a regression from #11194?)

richvdh:
> (And the second one seems to be a regression from https://github.com/matrix-org/synapse/pull/11194?)

sorry yes, I said that in the commit comment.

PR #12101, opened 2022-02-28
Limit the size of the aggregation_key

erikjohnston:
There's no reason to let people use long keys.


synapse/handlers/message.py
```diff
@@ -1073,6 +1073,9 @@ async def _validate_event_relation(self, event: EventBase) -> None:
             if already_exists:
                 raise SynapseError(400, "Can't send same reaction twice")
 
+            if len(aggregation_key) > 500:
+                raise SynapseError(400, "Aggregation key is too long")
```
clokep, line 1077: Probably best to do this before `has_user_annotated_event` to avoid unnecessary database operations?

clokep (approved):

PR #12111, opened 2022-03-01
Advertise Python 3.10 support in setup.py

anoadragon453:
This PR updates our classifiers defined in `setup.py` to advertise Synapse's support for Python 3.10.

I'm not entirely sure if this was omitted from https://github.com/matrix-org/synapse/issues/11780 on purpose or simply missed, but presenting it here for review.

### DIFF
--- a/changelog.d/12111.misc
+++ b/changelog.d/12111.misc
@@ -1,0 +1,1 @@
+Advertise support for Python 3.10 in packaging files.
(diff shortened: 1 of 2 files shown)


clokep (approved):
Seems fine, but needs a newsfile.

richvdh:
> I'm not entirely sure if this was omitted from https://github.com/matrix-org/synapse/issues/11780 on purpose or simply missed, but presenting it here for review.

no, I forgot. Now added. Please add anything else you remember so that we have a reference for 3.11.

Issue #3528, opened 2018-07-13
we shouldn't process requests where the requester has timed out/gone away


richvdh:
when we get busy, sometimes we queue up requests, which means that once we get round to processing them, there is no point in doing so. See #3444 for a good example.

richvdh:
a nice way to do this might be via cancellation of Deferreds?

richvdh:
keywords, since I just spent 10 minutes searching for this: cancel, disconnect

squahtx:
Cancelling Deferreds to abort processing looks very promising!

#### How Deferred.cancel() works
When `Deferred.cancel()` is called on a Deferred wrapping a coroutine, the call propagates down into the Deferred the coroutine is waiting on. That Deferred will either propagate the `cancel()` further or fail with a `twisted.internet.defer.CancelledError`, which gets raised inside the coroutine. This means that *any await expression may raise a CancelledError*, including trivial ones like `sleep()`!

The coroutine may then deal with the CancelledError using normal exception handling. The Deferred wrapping the coroutine will not automatically fail with a CancelledError unless it bubbles up through the coroutine.

asyncio has a similar feature, where an `asyncio.CancelledError` will be raised instead. It's interesting to note that twisted CancelledErrors are Exceptions, while asyncio CancelledErrors are BaseExceptions.

#### Complications
Unfortunately aborting processing of requests isn't as simple as calling `Deferred.cancel()` when clients disconnect:
* In many places, we catch and swallow plain `Exception`s without regard for `CancelledError`s. The result is that cancellation is ignored, which isn't too bad, since that's almost what happens today.
* More importantly, we have quite a few constructs (gatherResults, ObservableDeferred, ReadWriteLock, DeferredCache, run_in_background, defer_to_thread and more) which do fancy things with Deferreds and may not have been designed with cancellation in mind.
  Some of them allow cancel() calls to propagate when they shouldn't, don't propagate cancel() calls when they should or don't clean up after CancelledErrors correctly when they come from unexpected places.
  This can result in stuck requests piling up and all sorts of general weirdness. Ideally each construct needs to be checked, tested and fixed which is going to be very time-consuming.

Perhaps we could limit cancellation to a handful of interesting endpoints, like `/sync`? That way we wouldn't have to make all of synapse support cancellation right now. Though that might not be much help since all the interesting endpoints probably make use of many of the constructs that need checking.

richvdh:
[...]

PR #12120, opened 2022-03-01
Add cancellation support to `ReadWriteLock`

squahtx:
Best reviewed commit by commit.

Part of #3528.

`delay_cancellation` comes from #12180.

### Pull Request Checklist


synapse/util/async_helpers.py
```diff
@@ -506,22 +507,24 @@ def __init__(self) -> None:
...
+        @asynccontextmanager
+        async def _ctx_manager() -> AsyncIterator[None]:
             try:
+                # We wait for the latest writer to finish writing. We can safely ignore
+                # any existing readers... as they're readers.
+                # May raise a `CancelledError` if the `Deferred` wrapping us is
+                # cancelled. The `Deferred` we are waiting on must not be cancelled,
+                # since we do not own it.
+                if curr_writer:
+                    await make_deferred_yieldable(stop_cancellation(curr_writer))
```
squahtx, line 527: I'm not sure of the benefit of using async context managers.
It mainly benefits the cleanup for the read path, where the cleanup for not-holding-the-lock and holding-the-lock happen to be the same. On the write path we have to handle both cases differently, by deferring the cleanup until we actually acquire the lock.

synapse/util/async_helpers.py
```diff
@@ -541,24 +543,40 @@ def write(self, key: str) -> AsyncContextManager:
...
+                def release() -> None:
+                    with PreserveLoggingContext():
+                        new_defer.callback(None)
+                    # `self.key_to_current_writer[key]` may be missing if there was another
+                    # writer waiting for us and it completed entirely within the
+                    # `new_defer.callback()` call above.
+                    if self.key_to_current_writer.get(key) == new_defer:
+                        self.key_to_current_writer.pop(key)
+
+                if to_wait_on_defer.called:
```
richvdh, line 573: could we, instead of all this, do:

```python
             to_wait_on_defer = defer.gatherResults(to_wait_on)
             to_wait_on_defer.addBoth(lambda _: release())
             await make_deferred_yieldable(to_wait_on_defer)
             yield
```

... possibly with more `stop_cancellation` ?  Or does that not work?

tests/util/test_rwlock.py
```diff
@@ -100,23 +100,54 @@ async def action():
         self.assertTrue(acquired_d.called)
         release_d.callback(None)
 
+    def _start_reader_or_writer(
+        self,
+        read_or_write: Callable[[str], AsyncContextManager],
+        key: str,
+        name: str,
+    ) -> Tuple["Deferred[None]", "Deferred[None]"]:
+        """Starts a reader or writer which acquires the lock, blocks, then completes."""
```
richvdh, line 109: some docs on the params and return vals would be helpful.

tests/util/test_rwlock.py
```diff
@@ -32,76 +34,120 @@ def _assert_called_before_not_after(self, lst, first_false):
 
     def test_rwlock(self):
         rwlock = ReadWriteLock()
+        key = "key"
+
+        def start_reader_or_writer(
```
richvdh, line 39: it's a bit confusing that this is slightly different to the class-level `_start_reader_or_writer`. Can we combine them, even if it means returning three deferreds rather than two?

richvdh (changes requested):
nice. a few comments though.

synapse/util/async_helpers.py
```diff
@@ -546,11 +546,10 @@ async def write(self, key: str) -> ContextManager:
...
         self.key_to_current_writer[key] = new_defer
 
-        await make_deferred_yieldable(defer.gatherResults(to_wait_on))
-
-        @contextmanager
-        def _ctx_manager() -> Iterator[None]:
+        @asynccontextmanager
+        async def _ctx_manager() -> AsyncIterator[None]:
             try:
+                await make_deferred_yieldable(defer.gatherResults(to_wait_on))
```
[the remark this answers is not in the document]
squahtx: I'm all in favour of removing footguns. I'lk move the setup of `read()` and `write()` into the context manager.

synapse/util/async_helpers.py
```diff
@@ -541,24 +543,40 @@ def write(self, key: str) -> AsyncContextManager:
...
-                with PreserveLoggingContext():
-                    new_defer.callback(None)
-                # `self.key_to_current_writer[key]` may be missing if there was another
-                # writer waiting for us and it completed entirely within the
-                # `new_defer.callback()` call above.
-                if self.key_to_current_writer.get(key) == new_defer:
-                    self.key_to_current_writer.pop(key)
+
+                def release() -> None:
+                    with PreserveLoggingContext():
```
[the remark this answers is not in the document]
squahtx: I'm not wholly convinced the reasoning applies to `defer.callback()` since it's fairly "low level".

We do wrap `.callback(...)` calls in `PreserveLoggingContext` [elsewhere in Synapse](https://github.com/matrix-org/synapse/blob/release-v1.53/synapse/util/batching_queue.py#L164-L166).

More concretely, I believe that if there is a coroutine waiting on `defer`, then inside `defer.callback(...)`:
1. The coroutine's logging context will be restored because the Deferred went through `make_deferred_yieldable`. The logging context from outside the `defer.callback()` call will be forgotten.
2. The coroutine will run for a bit.
3. Then either:
    a) await another `make_deferred_yieldable` Deferred and restore the sentinel logging context, or
    b) terminate, at which point it will have the logging context it had when it was originally started.

synapse/util/async_helpers.py
```diff
@@ -541,24 +543,40 @@ def write(self, key: str) -> AsyncContextManager:
...
+
+                if to_wait_on_defer.called:
+                    release()
+                else:
+                    # We don't have the lock yet, probably because we were cancelled
+                    # while waiting for it. We can't call `release()` yet, since
+                    # `new_defer` must only resolve once all previous readers and
+                    # writers have finished.
+                    # NB: `release()` won't have a logcontext in this path.
+                    to_wait_on_defer.addCallback(lambda _: release())
```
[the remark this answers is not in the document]
squahtx: I'll make the change for the sake of robustness, but all of the Deferreds that represent lock release, and thus also `to_wait_on_defer`, should never error.

synapse/util/async_helpers.py
squahtx: `release()` must only be called after the yield, so I'm guessing you meant:
```py
try:
    await make_deferred_yieldable(stop_cancellation(to_wait_on_defer))
    yield
finally:
    to_wait_on_defer.addBoth(lambda _: release())
```
which I was originally hesitant to do because of re-entrancy in the happy path:
1. When we exit the await, it will be because we are inside a call to `to_wait_on_defer.callback(...)`.
2. During the `yield` / body of the context manager, we may or may not do a blocking await. If we block, then the `to_wait_on_defer.callback()` call gets exited, otherwise we remain within the `to_wait_on_defer.callback()` call.
3. We reach the finally block.
    If we're no longer inside the `callback()` call, `release()` will be called instantly with the current logging context.
    Otherwise if we're still inside the `callback()` call, `release()` won't be called until we next block or terminate and will have the sentinel context.

The sometimes-extra delay before `release()` is tricky to reason about, but I think the proposed code would be alright.

---------------------------------------------------------------------

As an alternative, I'm very tempted by @erikjohnston's suggestion of delaying CancelledErrors:
```py
try:
    await make_deferred_yieldable(delay_cancellation(to_wait_on_defer))
    yield
finally:
    release()  # but inline again
```
Where `delay_cancellation` would hold on to the CancelledError and only raise it once `to_wait_on_defer` resolves.

We'd get better maintainability at the cost of having the crud associated with cancelled requests lingering a little while longer, but maybe that's a worthwhile tradeoff.

tests/util/test_rwlock.py
```diff
@@ -32,58 +34,71 @@ def _assert_called_before_not_after(self, lst, first_false):
...
+            start_reader_or_writer(rwlock.write),  # 6
         ]
-        ds = [defer.ensureDeferred(d) for d in ds]
+        # `Deferred`s that resolve when each reader or writer acquires the lock.
+        acquired_ds = [acquired_d for acquired_d, _release_d in ds]
+        # `Deferred`s that will trigger the release of locks when resolved.
+        release_ds = [release_d for _acquired_d, release_d in ds]
 
-        self._assert_called_before_not_after(ds, 2)
+        self._assert_called_before_not_after(acquired_ds, 2)
```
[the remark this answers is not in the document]
squahtx: I found this test case hard to understand too. I'll try to make it more readable.

tests/util/test_rwlock.py
squahtx: Yep, let's combine them.

tests/util/test_rwlock.py
squahtx: Sounds like a good idea. Added some docs.

synapse/util/async_helpers.py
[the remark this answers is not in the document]
richvdh: hrm, yeah, I think you're right. I was always of the belief that wrapping `.callback` with `PreserveLoggingContext` was cargo-culting, but your reasoning seems good.

synapse/util/async_helpers.py
richvdh: > release() must only be called after the yield, 

oh, duh. Yes, of course.

> The sometimes-extra delay before release() is tricky to reason about, but I think the proposed code would be alright.

ugh, right, yes, let's avoid that.

And yes, `delay_cancellation` looks like it might be worth exploring.

squahtx: I had a go at using a delay in cadfe0add7717341ae90b091cc7178ffe224cffe.
We now depend on #12180 of course.

synapse/util/async_helpers.py
```diff
@@ -538,29 +541,36 @@ def _ctx_manager() -> Iterator[None]:
...
-        await make_deferred_yieldable(defer.gatherResults(to_wait_on))
-
-        @contextmanager
-        def _ctx_manager() -> Iterator[None]:
+            to_wait_on_defer = defer.gatherResults(to_wait_on)
             try:
+                # Wait for all current readers and the latest writer to finish.
+                # May raise a `CancelledError` immediately after the wait if the
+                # `Deferred` wrapping us is cancelled. We must only release the lock
+                # once we have acquired it, hence the delay.
```
richvdh, line 567:
```suggestion
                # once we have acquired it, hence the use of `delay_cancellation`
                # rather than `stop_cancellation`.
```

...  if I understand correctly?

richvdh (approved):
lgtm, subject to updates for consistency with #12180

PR #12121, opened 2022-03-01
Properly invalidate caches when an event with a relation is redacted

### LINKED ISSUE #12104: Redacted events leaked via the latest event in a thread
If the latest event in a thread is redacted, the unredacted content can be leaked via the bundled aggregation (until it is removed from the database).

This only occurs in MSC3440 support is enabled on the server.

clokep:
This fleshes out some of the redaction of relation tests a bit and fixes a bug where the caches were not invalidated properly when an event with a relation is redacted.

Work towards #11753 and fixes #12104

~~Builds on #12113~~


synapse/storage/databases/main/cache.py
```diff
@@ -197,9 +197,7 @@ def _invalidate_caches_for_event(
             self.get_invited_rooms_for_local_user.invalidate((state_key,))
 
         if relates_to:
-            self.get_relations_for_event.invalidate((relates_to,))
-            self.get_aggregation_groups_for_event.invalidate((relates_to,))
-            self.get_applicable_edit.invalidate((relates_to,))
+            self._invalidate_relations_for_event(relates_to)
```
clokep, line 200: We were previously not properly invalidating thread relations.

synapse/storage/databases/main/events.py
```diff
@@ -1943,15 +1943,29 @@ def _handle_batch_event(self, txn: LoggingTransaction, event: EventBase):
...
+        relates_to = self.db_pool.simple_select_one_onecol_txn(
+            txn,
+            table="event_relations",
+            keyvalues={"event_id": redacted_event_id},
+            retcol="relates_to_id",
+            allow_none=True,
+        )
+        # Any relation information for the related events must be cleared.
+        if relates_to is not None:
+            txn.call_after(self.store._invalidate_relations_for_event, relates_to)
```
clokep, line 1967: This will only redact this properly on the process which is persisting events (I believe) -- any ideas on how to stream this properly to other servers?

erikjohnston: There are two options:

1. Call `self.store._invalidate_cache_and_stream` for each cache you want to invalidate (instead of `txn.call_after(cache.invalidate)`. This both invalidates the cache locally and streams the invalidation to other workers. Or
2. Do what we do for current state event invalidation and send a special `RELATION_REDACTED` cache invalidation over replication, which then other workers will see and invalidate all the right caches:

https://github.com/matrix-org/synapse/blob/a8da0469070771b0b63c97a96f3221afea2aa2e7/synapse/storage/databases/main/cache.py#L42-L44

The latter method is really only worth it if we're frequently invalidating lots of caches at once

erikjohnston (commented):
LGTM other than needing to stream invalidations to other processes

synapse/storage/databases/main/events.py
clokep: I suspect the former is the way to go then, I don't see this as being a super frequent?

clokep: Done in 6687bbd8b524cb9f6f7751f2b0509a3d636c6a50, thanks!

synapse/storage/databases/main/events.py
```diff
@@ -1943,15 +1943,43 @@ def _handle_batch_event(self, txn: LoggingTransaction, event: EventBase):
...
         any redacted relations from the database.
 
         Args:
             txn
-            redacted_event_id (str): The event that was redacted.
+            redacted_event_id: The event that was redacted.
         """
 
+        # Fetch current relations.
+        relates_to = self.db_pool.simple_select_one_onecol_txn(
```
erikjohnston, line 1958: I've just spent 5 minutes wondering how this `relates_to` differs for the `relates_to` in the event cache invalidation processing. Maybe worth renaming this to `redacted_relates_to`?

erikjohnston (approved):

synapse/storage/databases/main/events.py
clokep: Right, so this is slightly different because the redaction event doesn't refer to `relates_to`, we want the `relates_to` from the event being redacted.

So you have:

```mermaid
flowchart LR
A[Redaction] -->|redacts| B[Redacted Event]
B -->|m.relates_to| C[Related Event]
```

And we need to clear the caches for "Related Event". 👍

PR #12127, opened 2022-03-02
Make release script write correct no-op changelog

erikjohnston:
As we want to include the previous version in the "No new changes..." string.

### DIFF
--- a/scripts-dev/release.py
+++ b/scripts-dev/release.py
@@ -17,6 +17,8 @@
 """An interactive script for doing a release. See `cli()` below.
 """
 
+import glob
+import os
 import re
 import subprocess
 import sys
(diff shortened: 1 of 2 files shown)


scripts-dev/release.py
```diff
@@ -523,5 +525,29 @@ class VersionSection:
...
+    with open("CHANGES.md", "r+") as f:
+        existing_content = f.read()
+        f.seek(0, 0)
+        f.write(new_changes)
+        f.write("\n")
+        f.write(existing_content)
+
+    # Remove all the news fragments
+    for f in glob.iglob("changelog.d/*.*"):
+        os.remove(f)
```
anoadragon453, line 549: Would this delete https://github.com/matrix-org/synapse/blob/develop/changelog.d/.gitignore?

anoadragon453 (commented):
lgtm besides one small issue.

scripts-dev/release.py
erikjohnston: Nope, glob ignores `.*` files by default

anoadragon453 (approved):
lgtm then!

PR #12128, opened 2022-03-02
Check if instances are lists, not sequences.

### LINKED ISSUE #12123: Don't use `isinstance(..., Sequence)`
In a couple of places, we validate incoming lists of strings with `isinstance(..., Sequence)`. Unfortunately this lets plain strings slip through.
We have to make these checks more specific.

clokep:
Since a string is a sequence, and this is all data from JSON we really only care about lists.

Fixes #12123

### DIFF
--- a/changelog.d/12128.misc
+++ b/changelog.d/12128.misc
@@ -1,0 +1,1 @@
+Fix data validation to compare to lists, not sequences.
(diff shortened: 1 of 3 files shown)


clokep:
I did some quick searching and couldn't find any others.

squahtx (approved):

PR #12130, opened 2022-03-02
Allow retrieving the relations of a redacted event

clokep:
MSC2675 states:

> Requesting `/relations` on a redacted event should still return any existing relation events. This is in line with other APIs like `/context` and `/messages`.

Currently an empty chunk is returned, this modifies this behavior to return the relations.

It additionally returns the bundled aggregations for a redacted event as MSC2675 only declares:

> Aggregations are never bundled into state events.

...and this seems inline with being able to view them from `/relations`, care must be taken to not leak edits of a redacted event as is the behavior for MSC2676.

~~Builds on #12121.~~
~~Builds on #12189.~~

Part of #11753.


synapse/storage/databases/main/relations.py
```diff
@@ -91,10 +91,18 @@ def __init__(
...
+        self.get_aggregation_groups_for_event.invalidate((event_id,))
+        self.get_applicable_edit.invalidate((event_id,))
+        self.get_thread_summary.invalidate((event_id,))
+        self.get_thread_participated.invalidate((event_id,))
+
     @cached(tree=True)
     async def get_relations_for_event(
         self,
-        event_id: str,
+        event: EventBase,
```
clokep, line 105: I wonder if this messes up the caching...

clokep: See https://github.com/matrix-org/synapse/pull/11915#discussion_r818657212 for more info...

clokep: Hmm -- I'm struggling to refactor this a bit since we would *have* to provide the `event_id` (and a matching `event`), but a bunch of the other properties, which we cache on, are optional / have defaults.

Any thoughts on how to refactor this? Or maybe we could have `cached` learn that an `EventBase` should be cached by event ID instead of the actual object?

clokep: > Any thoughts on how to refactor this? Or maybe we could have `cached` learn that an `EventBase` should be cached by event ID instead of the actual object?

I took a look at this and it seems doable, but probably adds code to a hot path which would be used rarely. As far as I can tell there's only 3 situations where we do something similar:

1. `get_relations_for_event` and `is_interested_in_event` (over `EventBase` and its `event_id` property).
2. `_get_joined_hosts` (over `_StateCacheEntry` and its `state_group` property).

This one is a bit ickier though since we have other parameters after it, I think the best thing to do is just default it to `None`, then raise an error if it is `None`.

synapse/storage/databases/main/events.py
```diff
@@ -1980,6 +1978,15 @@ def _handle_redact_relations(
                 txn, self.store.get_thread_participated, (redacted_relates_to,)
             )
 
+        # Caches which might leak edits must be invalidated for the event being
```
clokep, line 1981: This feels duplicative to what happens in `_invalidate_caches_for_event`, but I think that doesn't run on the main process?

synapse/storage/databases/main/relations.py
```diff
@@ -91,7 +91,7 @@ def __init__(
 
         self._msc3440_enabled = hs.config.experimental.msc3440_enabled
 
-    @cached(tree=True)
+    @cached(num_args=9, tree=True)
```
clokep, line 94: Without this the cache does nothing (since `EventBase` instances are never equal).

synapse/storage/databases/main/events.py
erikjohnston: `_invalidate_caches_for_event` happens on the main process? Otherwise the event caches on the main process would be wrong?

synapse/storage/databases/main/relations.py
erikjohnston: I don't think it would be that hard to make `@cached` accept a `cached_param_names`  parameter fwiw. The `num_args` basically just gets converted into a list of args to cache:

https://github.com/matrix-org/synapse/blob/26211fec24d8d0a967de33147e148166359ec8cb/synapse/util/caches/descriptors.py#L91-L106

erikjohnston (commented):
I think this is good, I'm just a bit wary of how unwieldy the type signature now is for `get_relations_for_event` in terms of having a non-optional optional param.

synapse/storage/databases/main/relations.py
clokep: Meaning so that we define something like `def f(event_id, event, blah = None, foo = True)` but cache on `event_id, blah, foo`?

clokep: Although you'd need to make those parameters incompatible and the descriptors are already fairly complicated... hmm.... I think it is probably OK since not passing it is a programming error?

clokep: See #12189 for being able to skip args.

synapse/storage/databases/main/events.py
clokep: I don't see any place that would be called on the main process, it only gets called via `CacheInvalidationWorkerStore.process_replication_rows` (either directly or via `_process_event_stream_row`).

I see some calls directly to `_invalidate_get_event_cache` though? These lines should actually be next to that, most likely...

clokep: I made a change in f15a0310d8f0483ef2789e198ab972763b968c40 which consolidates the logic a bit, which gives apples-to-apples comparison of what to clear when redacting an event.

synapse/storage/databases/main/events.py
```diff
@@ -1619,9 +1619,15 @@ def prefill():
...
-        # invalidate the cache for the redacted event
+    def _store_redaction(self, txn: LoggingTransaction, event: EventBase) -> None:
+        # invalidate the caches for the redacted event
         txn.call_after(self.store._invalidate_get_event_cache, event.redacts)
+        self.store._invalidate_cache_and_stream(
+            txn, self.store.get_relations_for_event, (event.redacts,)
+        )
+        self.store._invalidate_cache_and_stream(
+            txn, self.store.get_applicable_edit, (event.redacts,)
+        )
```
erikjohnston, line 1630: I don't think we need to stream the cache invalidation here, as they happen as part of the event replication? Removing it means we can shove less down the replication streams

clokep: Ah, you're right. The other bit of code I pointed to will get called after replication!

erikjohnston (approved):

Issue #7721, opened 2020-06-18
Performance of device list updates could be improved (aka /login is unusably slow)


babolivier:
I was investigating why logging in with a new device was leading up to poor performances on my account. It seems like both sending the device list update to other homeservers and other homeservers in turn hitting `GET /_matrix/federation/v1/user/devices/{userId}`.

Looking a bit further, it looks like both code paths end up calling [`_get_e2e_device_keys_txn`](https://github.com/matrix-org/synapse/blob/bd6dc17221741d4ceae05ae769a70696ae939336/synapse/storage/data_stores/main/end_to_end_keys.py#L83-L175) in the database code, which seems to be pretty expensive, and is called multiple time in a row, yet isn't cached. This results in quite bad performances:

![image](https://user-images.githubusercontent.com/5547783/85020132-c8fd1b00-b167-11ea-8cad-c71dcf3bb75a.png)

On the bottom right graph the `_get_e2e_device_keys_txn` transaction (in blue) is originating from sending out the `m.device_list_update` EDUs, while the `get_devices_with_keys_by_user` one (in purple) is originating from responding to `GET /_matrix/federation/v1/user/devices/{userId}`. Both transactions involve calling the `_get_e2e_device_keys_txn` function (in fact, the first one does only that).
On the bottom left graph these two transactions are colored respectively in green and purple.

erikjohnston:
I think the next steps here are to add a cache :+1:

hifi:
In #12048 I had an issue where logging in with a new device pretty much kills synapse for a while because of the device notify on login. Just batching up and slowing down sending the notifies would help a lot.

richvdh:
It's worth mentioning that this effectively prevents logging in with Hydrogen (cf https://github.com/vector-im/hydrogen-web/issues/699)

richvdh:
oh, and fluffychat, which seems to do the same thing

erikjohnston:
Oh, hmm. The problem described in the top of this issue (`_get_e2e_device_keys_txn` goes slow) and the slow login due to `/r0/keys/upload` may not have the same underlying cause. E.g. `/

Note that there seems to be two related but distinct problems:
1. `_get_e2e_device_keys_txn` is slow as mentioned in the issue description
2. `notify_device_updates` is slow

@richvdh ooi what were you seeing? Or was it a combination of the two?

richvdh:
[...]

PR #12132, opened 2022-03-02
Spread out sending device lists to remote hosts

erikjohnston:
This does to device list updates as to what #10398 did for presence.

The net effect of this is that we spread out starting new transmission loops when sending a device list update, up to a maximum of 30s. Hosts that already have a transmission loop running will send out the update immediately.

It's not super clear to me if this is acceptable. It means that a new device may not be able to read newly received messages for 30s after login. We may want to try and prioritise some hosts over others.

c.f. #7721

### DIFF
--- a/changelog.d/12132.feature
+++ b/changelog.d/12132.feature
@@ -1,0 +1,1 @@
+Improve performance of logging in for large accounts.
--- a/synapse/federation/send_queue.py
+++ b/synapse/federation/send_queue.py
@@ -244,7 +244,7 @@ class FederationRemoteSendQueue(AbstractFederationSender):
 
         self.notifier.on_new_replication_data()
 
-    def send_device_messages(self, destination: str) -> None:
+    def send_device_messages(self, destination: str, immediate: bool = False) -> None:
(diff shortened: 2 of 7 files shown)


babolivier (commented):
The code lgtm, but I'm feeling a bit uneasy about the change in general. 30s feels like quite a big delay for sending device lists update, and is probably more than enough time for a user to log in and start sending encrypted messages. I assume we're trying to process presence updates and device list updates in the same place to simplify the code, but my feeling is that device list updates are much more time-sensitive than presence updates (as they reflect much more badly on the overall client UX if they're late than the latter).

Maybe a solution could be to run two separate `_DestinationWakeupQueue`s side by side in the federation sender with different values for `_MAX_TIME_IN_QUEUE`? (I think something between 5s and 10s for device list updates would be more acceptable, but ymmv)

babolivier:
> > and is probably more than enough time for a user to log in and start sending encrypted messages
> 
> note that the problem is more about _receiving_ messages. Sending messages should be mostly fine.

Yep, I was thinking about e2ee the wrong way round.

PR #12133, opened 2022-03-02
Allow modules to sign JSON using the homeserver's keys

babolivier:
Part of the Tchap mainlining work. I'm currently working on mainlining the bits that replicate profile information to a central Sydent server, and the current code includes signing replication batches before sending them over. Thus the module implementing this feature needs this capability as well.

### DIFF
--- a/changelog.d/12133.feature
+++ b/changelog.d/12133.feature
@@ -1,0 +1,1 @@
+Allow modules to sign JSON using the homeserver's keys.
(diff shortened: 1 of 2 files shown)


richvdh (approved):
I guess so, though I wonder if it would be more general to expose the signing_key and let the module sign the json itself.

richvdh:
oh also, if you're going to stick with this approach, it might be best to add a test to `ModuleApiTestCase` to ensure it keeps working.

babolivier:
Though actually, now that I've paged the context back in, that's for profile replication which we're not doing anymore, so it's probably better to close this PR.

PR #12135, opened 2022-03-02
Add experimental environment variable to enable asyncio reactor

erikjohnston:
This is mostly just to try to see if it passes CI


richvdh:
related: #8642

clokep:
Looks like the only failure is a known-flake: matrix-org/sytest#1201.

DMRobertson:
> Looks like the only failure is a known-flake: [matrix-org/sytest#1201](https://github.com/matrix-org/sytest/issues/1201).

Though that issue is new today. Maybe there's a recent flakyness regression?

erikjohnston:
I wouldn't be surprised if using a different reactor changed the timings enough to surface some more flakey tests.

callahad:
Make it opt-in and let's roll.

Looks like we'll need Twisted >= 21.2 for this to work for us.

synapse/__init__.py
```diff
@@ -25,6 +25,27 @@
...
+            print("Using asyncio reactor requires Twisted>=21.2.0")
+            sys.exit(1)
+
+        import asyncio
+
+        from twisted.internet import asyncioreactor
+
+        asyncioreactor.install(asyncio.get_event_loop())
+    except ImportError:
+        pass
```
clokep, line 47: Is it worth printing something here? Otherwise it is a silent failure where you *think* it is working, but get no feedback.

(Or really -- should we be logging at all somewhere?)

clokep: Weird, but OK! 👍

synapse/__init__.py
```diff
@@ -25,6 +25,27 @@
     print("Synapse requires Python 3.7 or above.")
     sys.exit(1)
 
+# Allow using the asyncio reactor via env var.
+if bool(os.environ.get("SYNAPSE_ASYNC_IO_REACTOR", False)):
+    try:
+        from incremental import Version
```
clokep, line 31: Should `incremental` be added into `python_dependencies` since we're now using it directly?

clokep (approved):
Seems reasonable!

synapse/__init__.py
erikjohnston: Possibly, though I was imagining this only being a temporary thing while we had the experimental flag, so not much point adding it to the dependencies?

DMRobertson:
We are open to the idea of having this on by default if we can convince ourselves that this is safe/reliable for production use. Anecdotes welcome.

PR #12136, opened 2022-03-02
Remove unused mocks from `test_typing`

DMRobertson:
Noticed in #12108.

It's not clear what these do. `get_user_by_access_token` has the wrong
signature, including the return type. Tests all pass without these. I
think we should nuke them.

### DIFF
--- a/changelog.d/12136.misc
+++ b/changelog.d/12136.misc
@@ -1,0 +1,1 @@
+Remove unused mocks from `test_typing`.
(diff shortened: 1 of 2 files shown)


anoadragon453 (approved):
Some now unnecessary imports are blocking CI, but if it passes then LGTM.

DMRobertson:
> Some now unnecessary imports are blocking CI, but if it passes then LGTM.

One day I will learn to run the linter script.

PR #12138, opened 2022-03-02
Remove backwards compatibility with RelationPaginationToken.

clokep:
This removes the backwards compatibility (added in #11952) with the `RelationPaginationToken` for the `/relations` and `/aggregations/.../` endpoints.

From a user POV, this means that if someone makes a request on Synapse < v.1.52.0, then the administrator upgrades to Synapse >= v1.55.0 the pagination token will be rejected.

I think this is a fairly unlikely situation, but I'm not sure. 🤷 

Related to #11753.

### DIFF
--- a/changelog.d/12138.removal
+++ b/changelog.d/12138.removal
@@ -1,0 +1,1 @@
+Remove backwards compatibilty with pagination tokens from the `/relations` and `/aggregations` endpoints generated from Synapse < v1.52.0.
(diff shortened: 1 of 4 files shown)


anoadragon453 (approved):
Goodbye tech debt! :fire: 

> From a user POV, this means that if someone makes a request on Synapse < v.1.52.0, then the administrator upgrades to Synapse >= v1.55.0 the pagination token will be rejected.

That sounds pretty niche to me as well. It appears that the user will simply receive a 400 in response in that case, rather than an unhandled exception being raised, which is good.

I would be more cautious around backwards compatibility of sync tokens as they're longer lived. But less so for pagination tokens for `/relations`.

PR #12141, opened 2022-03-02
Correctly register deactivation and profile update module callbacks

babolivier:
The callbacks introduced in https://github.com/matrix-org/synapse/pull/12062 aren't wired up correctly so that they can actually be registered by modules. Let's fix that before we release 1.54.0 proper.

### DIFF
--- a/changelog.d/12141.bugfix
+++ b/changelog.d/12141.bugfix
@@ -1,0 +1,1 @@
+Fix a bug introduced in Synapse 1.54.0rc1 preventing the new module callbacks introduced in this release from being registered by modules.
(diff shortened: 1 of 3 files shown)


DMRobertson:
I think if you merge in `develop`, complement's CI should succeed.

babolivier:
> I think if you merge in `develop`, complement's CI should succeed.

This PR is again the release branch, so we probably don't want to do that. I'm fairly certain this PR doesn't impact complement passing, so maybe we can ignore it failing?

anoadragon453:
Apologies for missing this initially!

> I'm fairly certain this PR doesn't impact complement tests (or at least not in a way that would allow the Complement tests to fail but the Synapse tests to pass), so maybe we can ignore it failing?

Indeed. The test failures here stem from https://github.com/matrix-org/complement/pull/328, and is unrelated to this PR. Just to double-check, I've run `COMPLEMENT_DIR="path/to/complement" ./scripts-dev/complement.sh` locally on `babolivier/callbacks` with an up-to-date Complement with https://github.com/matrix-org/complement/pull/328 reverted and all tests pass.

PR #12142, opened 2022-03-02
Move the `snapcraft` configuration to `contrib`.

DMRobertson:
We're happy for people to package this as a snap image if it's useful,
but we don't support or maintain it. I'd like to move the config to
`contrib` to reflect this state of affairs.


clokep (approved):

DMRobertson:
I merged in develop, to pull in #12137 and keep CI happy.

Issue #11274, opened 2021-11-08
Synapse's documentation source files are split into two organisational systems and is confusing


anoadragon453:
As noted in https://github.com/matrix-org/synapse/pull/10086, part of the transition to the new documentation website was figuring out a structure for the existing documentation pages. These ended up being organised into what is now the layout configured by [SUMMARY.md](https://github.com/matrix-org/synapse/blob/a55e1ec9afc45a36f962e1e552128aeab5b1ee3c/docs/SUMMARY.md).

This allowed us to organise the documentation on the website, yet the source files were still left [in a pile](https://github.com/matrix-org/synapse/tree/9799c569bb481622b5882b2008a24e6c4658c431/docs) in the `docs/` directory. The decision was made to not move them into directories that mirror those in SUMMARY.md, as doing so would break existing (non-permanent) links to documentation files on GitHub.

This has currently left us in a state where our documentation is split into two organisational systems, as can be seen with the documentation for the Admin API. Docs exist in

1. https://github.com/matrix-org/synapse/tree/9799c569bb481622b5882b2008a24e6c4658c431/docs/admin_api and
2. https://github.com/matrix-org/synapse/tree/9799c569bb481622b5882b2008a24e6c4658c431/docs/usage/administration/admin_api

Enough time has now passed that links to Synapse's documentation should now be pointing to the documentation website instead of the source files. Thus, it is proposed to now shuffle the documentation files such that they mirror the directory structure of SUMMARY.md/the website. Perma-links (though referencing a specific commit) to the source files will continue to work, but all other links to documentation source files (in comments, not code) throughout the codebase should be updated to point to the website instead.

- [ ] Update existing links to documentation source files throughout the codebase to point at the website instead.
- [ ] Organise files in the `docs/` folder to match that of the website.
- [ ] Add a warning that several documentation source files has been moved to the changelog, to encourage external links to be updated.

Afterwards we'll only have a single system of organisation for the `docs/` folder, and it'll be obvious where to look for a file when coming from the documentation website.

anoadragon453:
I'm not sure if I realised it at the time, but URLs to doc pages are currently based on their filepath, rather than their organisation in SUMMARY.md. So moving files will break links to the documentation.

To help with this, I propose introducing stub pages in the old locations that redirect to the right place. These would be built in our CI rather than actually include the pages in `docs/`, as they're only really a concern for the matrix-org.github.io deployment. A file somewhere in the `docs/` folder would detail what pages mapped to what.

[...]

PR #12143, opened 2022-03-02
Clean-up demo scripts & documentation

clokep:
* Rewrites the demo documentation to be clearer, accurate, and moves it to our documentation tree.
* Improvements to the demo scripts:
	* `clean.sh` now runs `stop.sh` first so you don't get zombies
	* Uses some more modern configuration (and removes some obsolete ones).
	* Consistently use the HTTP ports for server name, etc.
	* Nuke the `demo/etc` directory and separate everything into the `demo/808x` directories.

I find most of this an improvement, but please shout if you disagree!


docs/development/demo.md
```diff
@@ -0,0 +1,40 @@
+# Synapse demo setup
+
+**DO NOT USE THESE DEMO SERVERS IN PRODUCTION**
+
+Requires you to have a [Synapse development environment setup](https://matrix-org.github.io/synapse/develop/development/contributing_guide.html#4-install-the-dependencies).
+
+The demo setup allows running three federation Synapse servers, with server
+names `localhost:8080`, `localhost:8081`, and `localhost:8082`.
+
+You can access them via any Matrix client over HTTP at `localhost:808{1,2,3}` or
```
H-Shay, line 10: Forgive me for my confusion, but I don't quite understand this part and thought I'd ask! The doc says that the script starts three servers with the names `localhost:8080`, `localhost:8081`, and `localhost:8082`. However, on the next line it says to access them at `localhost:808{1,2,3}`. My first thought upon looking that line was that this was saying to access them at `localhost:8081`, `localhost:8082`, or `localhost:8083` which I don't think is correct? I think beginners might take the `{1,2,3}` literally rather than reading it as a set of variables referring to `localhost:8080`, `localhost:8081`, etc. which is what I am assuming you meant?

clokep: > However, on the next line it says to access them at `localhost:808{1,2,3}`. My first thought upon looking that line was that this was saying to access them at `localhost:8081`, `localhost:8082`, or `localhost:8083` which I don't think is correct?

That is what this line is trying to say (and is correct). I can expand it out.

clokep: I updated this (and another instance) in d46c39b5dcbc02844d3980731be62438b8bb57ff.

H-Shay:
Looks great! Thank you for doing this, it's much clearer than what was originally there.

H-Shay (approved):

docs/SUMMARY.md
```diff
@@ -82,6 +82,7 @@
   - [Release Cycle](development/releases.md)
   - [Git Usage](development/git.md)
   - [Testing]()
+    - [Demo scripts](development/demo.md)
```
anoadragon453, line 85: Heads up that this should be under `development/testing/demo.md` according to the [doc docs](https://github.com/matrix-org/synapse/tree/develop/docs#adding-to-the-documentation).

The fact many things in this file *don't* yet follow that convention is tracked at #11274.

PR #12149, opened 2022-03-03
Add test for `ObservableDeferred`'s cancellation behaviour

squahtx:
`ObservableDeferred` already did the right thing, so let's ensure it continues to do the right thing with a test.

### Pull Request Checklist

### DIFF
--- a/changelog.d/12149.misc
+++ b/changelog.d/12149.misc
@@ -1,0 +1,1 @@
+Add test for `ObservableDeferred`'s cancellation behaviour.
--- a/tests/util/test_async_helpers.py
+++ b/tests/util/test_async_helpers.py
@@ -100,6 +100,34 @@ class ObservableDeferredTest(TestCase):
         self.assertEqual(str(results[0].value), "gah!", "observer 1 errback result")
         self.assertEqual(str(results[1].value), "gah!", "observer 2 errback result")
 
+    def test_cancellation(self):
+        """Test that cancelling an observer does not affect other observers."""
+        origin_d: "Deferred[int]" = Deferred()
+        observable = ObservableDeferred(origin_d, consumeErrors=True)
+
+        observer1 = observable.observe()
+        observer2 = observable.observe()
+        observer3 = observable.observe()
+
+        self.assertFalse(observer1.called)
+        self.assertFalse(observer2.called)
+        self.assertFalse(observer3.called)
+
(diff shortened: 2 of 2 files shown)


squahtx:
We should probably remove cancelled `Deferreds` from the `ObservableDeferred._observers` list?

squahtx:
> We should probably remove cancelled `Deferreds` from the `ObservableDeferred._observers` list?

After some investigation I'm not convinced this is worth it (and it's steering into microoptimization territory):

We can define a custom canceller that will remove the `Deferred` from the observers list:
```py
def handle_cancel(d):
    self._observers.remove(d)
    # or self._observers[index_of_d] = None
    # since _observers is a list and not a set.
d = Deferred(handle_cancel)
```

On cancellation, we would reclaim ~1472 B, according to sys.getsizeof:
```
 48 B - d
144 B - d.__dict__
 56 B - d.callbacks, which is []
 88 B - d.result, which is a Failure
144 B - d.result.__dict__
 56 B - d.result.frames, which is []
 56 B - d.result.stack, which is []
112 B - d.result.parents, which is ["twisted.internet.defer.CancelledError", "builtins.Exception", "builtins.BaseException", "builtins.object"]
 86 B - d.result.parents[0]
 67 B - d.result.parents[1]
 71 B - d.result.parents[2]
 64 B - d.result.parents[3]
 88 B - d.result.value, which is a CancelledError
104 B - d.result.value.__dict__, which is {}
 48 B - d._debugInfo, which is a DebugInfo
104 B - d._debugInfo.__dict__
136 B - def handle_cancel()
```
But we need to consider that the canceller itself adds 136 B of extra overhead per observer that isn't cancelled. We break even when the extra overhead = the percentage cancelled * 1472 B.

So we'd need 9.2% (136 B / 1472 B) of observers to be cancelled for us to break even in terms of memory usage, on average.
We can get this breakeven threshold down to 3.4% (48 B / 1384 B) by defining a callable canceller class with `__slots__`.

Only about 0.0011% of requests on matrix.org get cancelled, so the proposed cleanup is a few orders of magnitude away from saving any memory.

richvdh:
Yeah, given the `observers` list gets cleared once the deferred completes anyway, I really don't think it's worth clearing it.

richvdh (approved):
looks great.

squahtx:
done as part of #3528

Issue #12148, opened 2022-03-03
Stabilize support for MSC3440: thread relation


clokep:
[MSC3440](https://github.com/matrix-org/matrix-doc/pull/3440) is currently in a final comment period. We'll need to make a few tweaks to stabilize support for it:

* [x] Use the proper serialization configuration: #12085
* [x] Support stable identifiers:
	* `io.element.relation_senders` -> `related_by_senders`
	* `io.element.relation_types` -> `related_by_rel_types`
	* `io.element.thread` -> `m.thread`
* [ ] Remove experimental configuration flag and unstable identifiers.

We'll probably want to overlap support of stable & unstable identifiers for a couple of releases.

See also: #11753.

clokep:
See also a cache invalidation bug with thread summaries: #12104

Nira34:
https://matrix.org/

clokep:
Support for the unstable identifiers will be dropped in v1.58.0. Note that this will cause old clients to fallback to their local-only implementation of unstable threads, but this should be fine since people will eventually update to stable threads / clients.

PR #12151, opened 2022-03-03
Support stable identifiers for MSC3440: Threading

clokep:
This is blocked on [MSC3440](https://github.com/matrix-org/matrix-doc/pull/3440) completing the final comment period.

It supports both the unstable and stable identifiers (with the intention to remove the unstable identifiers after two releases).

Part of #12148


synapse/storage/databases/main/relations.py
```diff
@@ -514,14 +504,15 @@ def _get_thread_summaries_txn(
                         AND parent.room_id = child.room_id
                     WHERE
                         %s
-                        AND relation_type = ?
+                        AND (relation_type = ? OR relation_type = ?)
```
clokep, line 507: This (and the similar changes) mean that having both `m.thread` and `io.element.thread` in a thread would work...for now (until support for `io.element.thread` is dropped).

I prefer doing it this way though since it does not double the queries to get stable and unstable threads. Thoughts on whether this is acceptable?

clokep:
Setting a review flag on this since it is fully ready for review and only blocked on waiting for MSC3440 to complete FCP!

richvdh (commented):
I'm a bit hesitant to support the unstable identifiers (filter settings, relationship types) without a config setting guarding it, irrespective of the "TODO Remove this in Synapse >= v1.57.0" comment. I've got a suspicion its removal will be forgotten, and meanwhile a bunch of things will continue blithely using the unstable identifiers.

Not ruling it out though, just an initial reaction.

clokep:
> I'm a bit hesitant to support the unstable identifiers (filter settings, relationship types) without a config setting guarding it, irrespective of the "TODO Remove this in Synapse >= v1.57.0" comment. I've got a suspicion its removal will be forgotten, and meanwhile a bunch of things will continue blithely using the unstable identifiers.
> 
> Not ruling it out though, just an initial reaction.

That seems reasonable to me. I've made this change. 👍

synapse/storage/databases/main/relations.py
clokep: The above is still true, but only if you have the experimental config flag enabled.

synapse/events/utils.py
```diff
@@ -515,11 +515,13 @@ def _inject_bundled_aggregations(
...
                 )
 
-            serialized_aggregations[RelationTypes.THREAD] = {
+            thread_summary = {
                 "latest_event": serialized_latest_event,
                 "count": thread.count,
                 "current_user_participated": thread.current_user_participated,
             }
+            serialized_aggregations[RelationTypes.THREAD] = thread_summary
+            serialized_aggregations[RelationTypes.UNSTABLE_THREAD] = thread_summary
```
richvdh, line 524: could this be guarded by the feature flag?

synapse/rest/client/versions.py
```diff
@@ -101,6 +101,7 @@ def on_GET(self, request: Request) -> Tuple[int, JsonDict]:
                     "org.matrix.msc3030": self.config.experimental.msc3030_enabled,
                     # Adds support for thread relations, per MSC3440.
                     "org.matrix.msc3440": self.config.experimental.msc3440_enabled,
+                    "org.matrix.msc3440.stable": True,
```
richvdh, line 104:
```suggestion
                    "org.matrix.msc3440.stable": True,  # TODO: remove when "v1.3" is added above
```

synapse/storage/databases/main/relations.py
```diff
@@ -514,16 +514,22 @@ def _get_thread_summaries_txn(
...
             clause, args = make_in_list_sql_clause(
                 txn.database_engine, "relates_to_id", event_ids
             )
+
             args.append(RelationTypes.THREAD)
+            if self._msc3440_enabled:
+                relations_clause = "(relation_type = ? OR relation_type = ?)"
+                args.append(RelationTypes.UNSTABLE_THREAD)
+            else:
+                relations_clause = "relation_type = ?"
```
richvdh, line 530: I'd find this easier to follow:

```suggestion
            if self._msc3440_enabled:
                relations_clause = "(relation_type = ? OR relation_type = ?)"
                args.extend((RelationTypes.THREAD, RelationTypes.UNSTABLE_THREAD))
            else:
                relations_clause = "relation_type = ?"
                args.append(RelationTypes.THREAD)
```

likewise below

richvdh (commented):
looks generally good to me!

synapse/events/utils.py
clokep: Good catch! 👍 Guess my commits weren't quite as atomic as I thought.

richvdh (approved):

PR #12154, opened 2022-03-03
Avoid generating state groups for local out-of-band leaves

richvdh:
If we locally generate a rejection for an invite received over federation, it
is stored as an outlier (because we probably don't have the state for the
room). However, currently we still generate a state group for it (even though
the state in that state group will be nonsense).

By setting the `outlier` param on `create_event`, we avoid the nonsensical
state.

### DIFF
--- a/changelog.d/12154.misc
+++ b/changelog.d/12154.misc
@@ -1,0 +1,1 @@
+Avoid generating state groups for local out-of-band leaves.
--- a/synapse/handlers/room_member.py
+++ b/synapse/handlers/room_member.py
@@ -1736,8 +1736,8 @@ class RoomMemberMasterHandler(RoomMemberHandler):
             txn_id=txn_id,
             prev_event_ids=prev_event_ids,
             auth_event_ids=auth_event_ids,
+            outlier=True,
         )
-        event.internal_metadata.outlier = True
         event.internal_metadata.out_of_band_membership = True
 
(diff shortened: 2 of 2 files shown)


synapse/handlers/room_member.py
```diff
@@ -1736,8 +1736,8 @@ async def _generate_local_out_of_band_leave(
             txn_id=txn_id,
             prev_event_ids=prev_event_ids,
             auth_event_ids=auth_event_ids,
+            outlier=True,
```
clokep, line 1739: For my own reference, by passing this here:

1. The event still gets it set on `internal_metadata` via `create_new_client_event`
2. Avoids calling `compute_event_context`, which (I think) is what creates the state group.

clokep (approved):

changelog.d/12154.misc
```diff
@@ -0,0 +1 @@
+Avoid generating state groups for local out-of-band leaves.
```
clokep, line 1: Is there anything to "clean-up" in the database from this?

richvdh: well in theory we might be able to go and delete the incorrectly created state groups. In practice, I'm not sure it's worthwhile.

synapse/handlers/room_member.py
richvdh: correct!

Issue #12074, opened 2022-02-24
`/_matrix/federation/v1/state` returns a 500 error if called for an outlier


richvdh:
```
2022-02-24 09:59:19,270 - synapse.http.server - 100 - ERROR - GET-281 - Failed handle request via 'FederationStateV1Servlet': <SynapseRequest at 0x7f683468c0a0 method='GET' uri='/_matrix/federation/v1/state/%21ItGBZyOeGywOWwJbSK%3Alocalhost%3A8840?event_id=%242DMzm1OALGG81SL30L-ArnxcWKfwrlvkv-bqehxWNdk' clientproto='HTTP/1.1' site='8849'>
Traceback (most recent call last):
  File "/venv/lib/python3.9/site-packages/twisted/internet/defer.py", line 1660, in _inlineCallbacks
    result = current_context.run(gen.send, result)
StopIteration: 0

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/venv/lib/python3.9/site-packages/twisted/internet/defer.py", line 1660, in _inlineCallbacks
    result = current_context.run(gen.send, result)
StopIteration: {'ed25519:a_VhKa': FetchKeyResult(verify_key=<nacl.signing.VerifyKey object at 0x7f6834b83550>, valid_until_ts=1645782543370)}

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/venv/lib/python3.9/site-packages/synapse/http/server.py", line 269, in _async_render_wrapper
    callback_return = await self._async_render(request)
  File "/venv/lib/python3.9/site-packages/synapse/http/server.py", line 471, in _async_render
    callback_return = await raw_callback_return
  File "/venv/lib/python3.9/site-packages/synapse/federation/transport/server/_base.py", line 323, in new_func
    response = await func(
  File "/venv/lib/python3.9/site-packages/synapse/federation/transport/server/federation.py", line 160, in on_GET
    return await self.handler.on_room_state_request(
  File "/venv/lib/python3.9/site-packages/synapse/federation/federation_server.py", line 535, in on_room_state_request
    await self._state_resp_cache.wrap(
  File "/venv/lib/python3.9/site-packages/synapse/util/caches/response_cache.py", line 250, in wrap
    return await make_deferred_yieldable(entry.result.observe())
  File "/venv/lib/python3.9/site-packages/twisted/internet/defer.py", line 1660, in _inlineCallbacks
    result = current_context.run(gen.send, result)
[...]

PR #12155, opened 2022-03-03
Update client-visibility filtering for outlier events

richvdh:
Avoid trying to get the state for outliers, which isn't a sensible thing to do.

### DIFF
--- a/synapse/visibility.py
+++ b/synapse/visibility.py
@@ -81,8 +81,9 @@ async def filter_events_for_client(
 
     types = ((EventTypes.RoomHistoryVisibility, ""), (EventTypes.Member, user_id))
 
+    # we exclude outliers at this point, and then handle them separately later
     event_id_to_state = await storage.state.get_state_for_events(
-        frozenset(e.event_id for e in events),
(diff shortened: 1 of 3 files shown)


synapse/visibility.py
```diff
@@ -81,8 +81,9 @@ async def filter_events_for_client(
 
     types = ((EventTypes.RoomHistoryVisibility, ""), (EventTypes.Member, user_id))
 
+    # we exclude outliers at this point, and then handle them separately later
     event_id_to_state = await storage.state.get_state_for_events(
-        frozenset(e.event_id for e in events),
+        frozenset(e.event_id for e in events if not e.internal_metadata.outlier),
```
richvdh, line 86: it's worth noting that currently, `get_state_for_events` returns an empty list when called on an outlier (which is why this currently works).

However, I'm planning to change it to raise an exception (because I'm pretty sure that doing so is normally a programming error and produces mess like https://github.com/matrix-org/synapse/issues/12074), which is why this PR is needed.

erikjohnston (approved):

PR #4864, opened 2019-03-15
Migrate the user directory initial population to a background task

hawkowl:

PR #12157, opened 2022-03-03
Fix a bug in background updates wherein background updates are never run using the default batch size

H-Shay:
AFICT from the code, the approach to background updates should be to run the first batch with the default batch size, and then tune from there (relevant code starts here): https://github.com/matrix-org/synapse/blob/8533c8b03d8916e3805c7d0e0020226017680147/synapse/storage/background_updates.py#L408

However, in #4864 a guard clause was introduced here in this function https://github.com/matrix-org/synapse/blob/8533c8b03d8916e3805c7d0e0020226017680147/synapse/storage/background_updates.py#L100 such that this function would never return `None`, and thus the code above would never hit the conditional branch requiring `None`, which is the branch that sets the batch size to the default size. Thus the first iteration of the background update was always run with the minimum batch size. 

This was not noticed because at the time and for a very long time after this change, because the default batch size and the minimum batch size were the same size (100). The minimum batch size was changed recently to 1 by #11422, and as a part of that PR the tests were refactored so again this bug was missed. 

In this PR I have attempted to fix the bug and reset the tests to check that the correct batch size is being run.


erikjohnston (approved):
:tada:

PR #12163, opened 2022-03-04
Reduce to-device queries for /sync.

erikjohnston:
Right now we always hit the DB, even if the stream change cache reports that there are no new to-device messages for the users.


changelog.d/12163.misc
```diff
@@ -0,0 +1 @@
+Reduce number of DB queries made during processing of `/sync`.
```
anoadragon453, line 1:
```suggestion
Reduce number of DB queries made during processing of `/sync`.
```

anoadragon453 (approved):
Good catch!

PR #12165, opened 2022-03-04
Remove references to "msc2403"

anoadragon453:
Complement uses golang build flags to determine whether certain tests should be run. After https://github.com/matrix-org/complement/pull/331 is merged, setting this flag will no longer be required.

This PR also cleans up usage of the `msc2403_enabled` experimental config option, which was removed in https://github.com/matrix-org/synapse/pull/10167.

This is a small PR, but for best effect review commit-by-commit.


squahtx (approved):

anoadragon453:
https://github.com/matrix-org/complement/pull/331 has now been merged, so this can be too.

PR #12166, opened 2022-03-04
Relax version guard for packaging

DMRobertson:
It’s just occurred to me that #12088 pulled in the “packaging” package (~=21.3). I pulled in the newest version I had at the time, and I didn’t check this with the packagers, sorry. (Also sorry to bring up the P-word again.)

I only use it for packaging.requirements.Requirements. Which was added in packaging 16.1: https://github.com/pypa/packaging/releases/tag/16.1

https://pkgs.org/download/python3-packaging suggests that the oldest version we care about is 17.1 in Ubuntu Bionic. So I think with this bound we're hunky dory.

### DIFF
--- a/changelog.d/12166.misc
+++ b/changelog.d/12166.misc
@@ -1,0 +1,1 @@
+Relax the version guard for "packaging" added in #12088.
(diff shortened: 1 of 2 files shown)


richvdh (approved):
This is failing CI because it's running against sytest develop, which expects `synapse/synctl` to be a python entrypoint, rather than a script that gets copied in (cf https://github.com/matrix-org/sytest/pull/1202/files).

The change lgtm, so I'm going to go ahead and merge it.

PR #12168, opened 2022-03-05
Implement changes to MSC2285

SimonBrandner:
### Pull Request Checklist


synapse/handlers/receipts.py
```diff
@@ -178,35 +179,27 @@ def filter_out_hidden(events: List[JsonDict], user_id: str) -> List[JsonDict]:
...
+                    for rr_user_id, user_rr in m_read_private.items():
+                        if rr_user_id == user_id:
+                            new_users[rr_user_id] = user_rr.copy()
+
+                    # Set new users unless empty
+                    if len(new_users.keys()) > 0:
+                        new_event["content"][event_id] = {
+                            ReceiptTypes.READ_PRIVATE: new_users
+                        }
+                    continue
```
clokep, line 200: I think you can avoid the loop and heavily simplify this:

```suggestion
                    user_rr = m_read_private.get(user_id)
                    if user_rr:
                        new_event["content"][event_id] = {
                            ReceiptTypes.READ_PRIVATE: {user_id: user_rr.copy()}
                        }
```

Although I think the current code (and my code above) will have issues if the current user provides both a private and non-private read receipt for the same event. This is a silly thing to do, but I suspect we should send both back?

synapse/push/push_tools.py
```diff
@@ -24,7 +23,7 @@ async def get_badge_count(store: DataStore, user_id: str, group_by_room: bool) -
     invites = await store.get_invited_rooms_for_local_user(user_id)
     joins = await store.get_rooms_for_user(user_id)
 
-    my_receipts_by_room = await store.get_receipts_for_user(user_id, ReceiptTypes.READ)
+    my_receipts_by_room = await store.get_receipts_for_user(user_id)
```
clokep, line 26: I'm not sure if the changes to the various store methods to remove the receipt makes sense? I think it makes an assumption that all receipts are some sort of read-receipt, which might not be true?

It might make more sense to have `get_receipts_for_user` and friends to take a list of receipt types or to return a map of receipt type -> latest.

clokep (commented):
Thanks for the pull request! I've left a bunch of comments, although haven't fully looked over the tests yet. Some of them are a bit general, so wanted to leave them earlier and have a conversation before going over the entire PR finely!

Let me know what I've misunderstood. 😁

SimonBrandner:
> Thanks for the pull request! I've left a bunch of comments, although haven't fully looked over the tests yet. Some of them are a bit general, so wanted to leave them earlier and have a conversation before going over the entire PR finely!

Thank you for the review and the suggestions! I've fixed some of the things but I still have to look into a few

clokep:
Changes so far look great! 🎉  Let me know if you have questions on any of my comments (or shout/rerequest when it is ready for review again!)

synapse/rest/client/receipts.py
```diff
@@ -46,42 +46,47 @@ def __init__(self, hs: "HomeServer"):
...
-        hidden = body.get(ReadReceiptEventFields.MSC2285_HIDDEN, False)
-
-        if not isinstance(hidden, bool):
-            raise SynapseError(
-                400,
-                "Param %s must be a boolean, if given"
-                % ReadReceiptEventFields.MSC2285_HIDDEN,
-                Codes.BAD_JSON,
-            )
+        parse_json_object_from_request(request, allow_empty_body)
```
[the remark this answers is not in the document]
SimonBrandner: From what I gather the simplest option, might be just writing a comment unless I am missing something, I guess some of the logic from `parse_json_object_from_request()` could be copied over here but I don't like that much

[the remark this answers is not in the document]
clokep: A comment explaining it would be fine. 👍

synapse/push/push_tools.py
SimonBrandner: At the moment all receipts are some sort of a read receipt, so the assumption is true, at least right now, so I am a little hesitant to make changes (especially since I am not that confident with the SQL stuff). That being said, I can certainly give it a shot, if you think it's worth it

synapse/handlers/receipts.py
```diff
@@ -178,35 +183,23 @@ def filter_out_hidden(events: List[JsonDict], user_id: str) -> List[JsonDict]:
...
-                        # we just ignore the read receipt
-                        continue
+                # Filter the private read receipts to only the requesting user
+                m_read_private = event_content.get(ReceiptTypes.READ_PRIVATE, None)
+                if m_read_private:
+                    user_rr = m_read_private.get(user_id, None)
+                    if user_rr:
+                        new_event["content"][event_id] = {
+                            ReceiptTypes.READ_PRIVATE: {user_id: user_rr.copy()}
+                        }
```
clokep, line 199: I think this will mean that if a user has a private read receipt at event `A`, then we'll not return any other user's at that event.

We likely want:

```suggestion
                        new_event["content"].setdefault(event_id, {})[ReceiptTypes.READ_PRIVATE] = {user_id: user_rr.copy()}
```

This would be a great thing to add a unit test for!

synapse/push/push_tools.py
clokep: Yeah... let me think a bit more about it, but that does seem silly to abstract it for no reason.

synapse/handlers/receipts.py
SimonBrandner: You're right, this could cause an issue, I am trying to figure out the optimal solution

clokep: Feels to me like we would definitely want a test with multiple receipt types on the same event too, since this is the first time we're doing it.

SimonBrandner:
AFAICT, the tests fails are unrelated?

tests/rest/client/test_sync.py
```diff
@@ -639,13 +635,73 @@ def test_unread_counts(self) -> None:
...
+            "/rooms/%s/receipt/org.matrix.msc2285.read.private/%s"
+            % (self.room_id, res2["event_id"]),
+            {},
+            access_token=self.tok,
+        )
+        self.assertEqual(channel.code, 200, channel.json_body)
+        self._check_unread_count(0)
+
+    @override_config({"experimental_features": {"msc2285_enabled": True}})
+    def test_doesnt_reverse_read_receipts(self) -> None:
```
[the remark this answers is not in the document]
SimonBrandner: I've renamed the test and updated the comments - is it clearer now?

synapse/handlers/receipts.py
```diff
@@ -168,6 +166,13 @@ def __init__(self, hs: "HomeServer"):
 
     @staticmethod
     def filter_out_hidden(events: List[JsonDict], user_id: str) -> List[JsonDict]:
```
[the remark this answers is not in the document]
SimonBrandner: You're right, it is definitely worth looking into though I'd suggest leaving it for another PR

synapse/storage/databases/main/receipts.py
```diff
@@ -541,14 +534,20 @@ def insert_linearized_receipt_txn(
...
-                if int(so) >= stream_ordering:
+            txn.execute(sql, (room_id, user_id))
+
+            for so, eid, rt in txn:
+                if int(so) >= stream_ordering and (
+                    receipt_type == rt
+                    or (
+                        rt == ReceiptTypes.READ
+                        and receipt_type == ReceiptTypes.READ_PRIVATE
+                    )
```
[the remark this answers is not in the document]
SimonBrandner: This was meant to prevent the read receipts from going "up" in the timeline to an older event. This happens when we have a read-receipt for event B and then receive a read receipt for event A (which is older than B) of a different type. We then just get the latest read-receipt that the server has received which is the read-receipt for event A.

This doesn't always work but should be fixable. Anyway, I am actually not sure what is the most optimal solution here. 

Should we ignore new read receipts for events for which we already have a read receipt of a similar type (the types that influence notif counts) or should we somehow try to handle this when getting the read receipts from the DB instead?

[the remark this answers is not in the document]
SimonBrandner: I've pushed some more changes with the fix and added tests though I am still not exactly sure about the best solution here

synapse/handlers/receipts.py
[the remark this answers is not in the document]
clokep: Yes, definitely wasn't asking for that change here! Sorry for being unclear!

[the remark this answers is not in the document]
SimonBrandner: No worries!

SimonBrandner:
CI failure due to https://github.com/matrix-org/synapse/issues/12324, I assume

tests/rest/client/test_sync.py
```diff
@@ -639,13 +653,74 @@ def test_unread_counts(self) -> None:
...
+        self.assertEqual(channel.code, 200, channel.json_body)
+        self._check_unread_count(0)
+
+    @override_config({"experimental_features": {"msc2285_enabled": True}})
+    def test_read_receipts_only_go_down(self) -> None:
+        # Join the new user
+        self.helper.join(room=self.room_id, user=self.user2, tok=self.tok2)
+
+        # We tests this for both receipt types that influence notification counts
+        for receipt_type in [ReceiptTypes.READ, ReceiptTypes.READ_PRIVATE]:
```
clokep, line 690: A better option for this might be to use `parameterized`, looks like this is used on the `test_read_receipt_with_empty_body` test.

synapse/push/push_tools.py
clokep: I'm going to consolidate this conversation into https://github.com/matrix-org/synapse/pull/12168/files#r829202648

synapse/storage/databases/main/receipts.py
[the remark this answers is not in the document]
clokep: My thought was that we should simplify this logic here and always insert the latest read receipt / hidden read receipt into the room (ensuring separately they don't go "up" in a room), and then when we fetch the (read) receipt we can do some logic on them to see which is "later". We could do this in a couple of ways:

* By updating `get_receipts_for_user` to be `get_latest_receipt_for_user` and modifying it to search across multiple receipt types (it will require some custom SQL, but I think it will be fairly straightforward!)
* Renaming `get_receipts_for_user` to `get_read_receipt_for_user` and embedding the read vs. hidden read receipt logic into it.

In either way the logic should be pretty similar.

clokep (commented):
A couple of minor comments and then the conversation about the SQL bits. 😄

tests/rest/client/test_sync.py
```diff
@@ -412,17 +412,28 @@ def test_hidden_read_receipts(self) -> None:
...
+        channel = self.make_request(
+            "POST",
+            f"/rooms/{self.room_id}/receipt/m.read/{res['event_id']}",
+            {},
+            access_token=self.tok2,
+        )
+        self.assertEqual(channel.code, 200)
+
+        # Test that we didn't override the private read receipt
         self.assertEqual(self._get_read_receipt(), None)
```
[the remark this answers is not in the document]
SimonBrandner: No, it wouldn't, if we allowed this, it could lead to read-receipts going up in the timeline and other potentially weird things.

tests/rest/client/test_sync.py
SimonBrandner: Ah, thanks! I was thinking there was a better way to do this! :sweat_smile:

synapse/push/push_tools.py
SimonBrandner: :+1: I am going to mark this one as resolved

tests/rest/client/test_sync.py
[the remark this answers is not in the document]
clokep: I'm not sure I follow -- shouldn't we choose whichever is latest between read receipts and hidden read receipts?

SimonBrandner:
@clokep, I am not very confident in the SQL (and other) changes in the PR - **please review carefully**

synapse/rest/client/read_marker.py
```diff
@@ -48,34 +49,51 @@ async def on_POST(
...
+
+        read_marker_event_id = body.get(ReceiptTypes.FULLY_READ, None)
         if read_marker_event_id:
             await self.read_marker_handler.received_client_read_marker(
                 room_id,
                 user_id=requester.user.to_string(),
                 event_id=read_marker_event_id,
             )
 
+        for key in body.keys():
```
clokep, line 79: This is somewhat optional, but I wonder if it would be clearer in here to:

1. Calculate `valid_receipt_types = {READ}`, if `msc2285_enabled` then add `READ_PRIVATE`.
2. Iterate over the validate receipt types fetching from the body and calling `received_client_read_marker`.
3. Doing the assertion that `set(body.keys()) <= (valid_receipt_types | {FULLY_READ})`

I think we also want the assertion to happen before we persist anything!

synapse/storage/databases/main/receipts.py
```diff
@@ -133,38 +133,63 @@ async def get_receipts_for_room(
             desc="get_receipts_for_room",
         )
 
-    @cached(num_args=3)
```
clokep, line 136: I think we need to keep `get_last_receipt_event_id_for_user` and `get_receipts_for_user` as cached methods, receipts get fetched quite often.

clokep: I'm suspecting you got rid of these because you want to be able to invalidate a cache using on partial arguments -- the way to do this is to setup the cache as a tree:

```python
@cached(tree=True)
async def get_last_receipt_event_id_for_user(self, user_id, room_id, receipt_type):
    ...
```

Then when you invalidate it you can give partial arguments:

```python
get_last_receipt_event_id_for_user.invalidate((user_id, room_id,))
```

SimonBrandner: That doesn't seem to work as `List` isn't hashable :/

synapse/rest/client/read_marker.py
SimonBrandner: Thank you for the suggestion - it makes the code much clearer. I ended up not doing number 2 since we sometimes call `received_client_read_marker()` and sometimes `received_client_receipt()`, so it seems iterating would look a little weird; it might make more sense if there is more receipt types added

synapse/storage/databases/main/receipts.py
clokep: Right, most things aren't. I guess I see two options:

1. Use something hashable (a `tuple` is easier, but a `frozenset` is likely more correct).
2. Modify the methods to be read receipts.

I suspect going the second route is easier, but I would be OK with either.

synapse/rest/client/read_marker.py
```diff
@@ -48,27 +49,38 @@ async def on_POST(
...
         body = parse_json_object_from_request(request)
-        read_event_id = body.get(ReceiptTypes.READ, None)
-        hidden = body.get(ReadReceiptEventFields.MSC2285_HIDDEN, False)
 
-        if not isinstance(hidden, bool):
+        valid_receipt_types = {ReceiptTypes.READ, ReceiptTypes.FULLY_READ}
+        if self.config.experimental.msc2285_enabled:
+            valid_receipt_types.add(ReceiptTypes.READ_PRIVATE)
+
+        if set(body.keys()) > (valid_receipt_types):
```
clokep, line 57: I think the parents are unnecessary:

```suggestion
        if set(body.keys()) > valid_receipt_types:
```

synapse/storage/databases/main/receipts.py
SimonBrandner: It seems that doesn't work? When `invalidate()` is called, we would need to invalidate the list of RR types but also all other lists of RR types which include some of the same elements? That does not seem to happen - am I missing something?

synapse/storage/databases/main/receipts.py
```diff
@@ -133,59 +133,95 @@ async def get_receipts_for_room(
...
             row[0]: {
                 "event_id": row[1],
-                "topological_ordering": row[2],
-                "stream_ordering": row[3],
+                "receipt_type": row[2],
+                "topological_ordering": row[3],
+                "stream_ordering": row[4],
             }
             for row in rows
         }
```
clokep, line 227: I don't think this is quite right, it will only return a single receipt per room, even though we're fetching (potentially) multiple receipts per room.

clokep:
I think this is close to ready, we should check that we aren't leaking the private read receipt to other users at all (I'm fairly certain we have tests for this, but want to double check them with the refactoring that I made).

synapse/storage/databases/main/receipts.py
```diff
@@ -144,60 +144,126 @@ async def get_receipts_for_room(
...
         Args:
             user_id: The user to fetch receipts for.
             room_id: The room ID to fetch the receipt for.
-            receipt_type: The receipt type to fetch.
+            receipt_type: The receipt types to fetch. Earlier receipt types
+                are given priority if multiple receipts point to the same event.
 
         Returns:
-            The event ID of the latest receipt, if one exists; otherwise `None`.
+            The latest receipt and stream ordering, if one exists.
```
SimonBrandner, line 160: Stream ordering?

SimonBrandner, line 160: Stream ordering

SimonBrandner (commented):
@clokep, your changes look good to me!

PR #12175, opened 2022-03-07
Do not return allowed_room_ids from /hierarchy response.

### LINKED ISSUE #12169: Extra fields in space hierarchy listing response
### Description

Synapse includes at least two fields in the `GET …/hierarchy` space listing response that it probably shouldn't:

* `join_rules` which seems to be duplicating `join_rule`
* `allowed_room_ids`, for restricted rooms

I originally reported this in the spec since I assumed `allowed_room_ids` was actually meant to be received by clients: https://github.com/matrix-org/matrix-spec/issues/984#issuecomment-1059750228

### Steps to reproduce

- Open Element dev Console (Ctrl+Shift+I), Network tab
- List rooms of a space with restricted rooms
- Observe a `GET …/hierarchy` request and its response

### Version information

- **Homeserver**: flipdot.org

If not matrix.org:

- **Version**: 1.53.0
- **Install method**: 🤷🏼
- **Platform**: 🤷🏼

clokep:
Note that this was done properly for the `/spaces` API, but not the `/hierarchy` API.

Now with a test! 🎉 

Fixes #12169.

### DIFF
--- a/changelog.d/12175.bugfix
+++ b/changelog.d/12175.bugfix
@@ -1,0 +1,1 @@
+Fix a bug where non-standard information was returned from the `/hierarchy` API. Introduced in Synapse v1.41.0.
--- a/synapse/handlers/room_summary.py
+++ b/synapse/handlers/room_summary.py
@@ -295,7 +295,7 @@ class RoomSummaryHandler:
             # inaccessible to the requesting user.
             if room_entry:
                 # Add the room (including the stripped m.space.child events).
-                rooms_result.append(room_entry.as_json())
+                rooms_result.append(room_entry.as_json(for_client=True))
 
                 # If this room is not at the max-depth, check if there are any
                 # children to process.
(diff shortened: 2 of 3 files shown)


squahtx (approved):
Looks good!

Was `join_rules` in #12169 was an unstable field?

Also I'm guessing we aren't very concerned if remote homeservers return extra unspecced fields?

clokep:
> Was `join_rules` in #12169

I'm not quite following, sorry. Are you asking about `join_rules` vs. `join_rule`? This was a typo in the original MSC and I think that we want to keep for now for backwards compatibility.

> Also I'm guessing we aren't very concerned if remote homeservers return extra unspecced fields?

Hmm, that's a good point! I'm unsure if we usually rebuild the responses or not. Will take a look at some other spots in the code.

squahtx:
\> `join_rules` vs. `join_rule`
Sorry if that wasn't clear. I was trying to ask why this PR doesn't need to remove the `join_rules` key mentioned in #12169.

clokep:
> > Also I'm guessing we aren't very concerned if remote homeservers return extra unspecced fields?
> 
> Hmm, that's a good point! I'm unsure if we usually rebuild the responses or not. Will take a look at some other spots in the code.

It looks like in other spots we just pass through the responses over federation. I'm not sure how good of an idea that is, but at least we're consistent? 😄 

(It does have the benefit of making it slightly easier to add fields in the future.)

clokep:
> > `join_rules` vs. `join_rule`
> 
> Sorry if that wasn't clear. I was trying to ask why this PR doesn't need to remove the `join_rules` key mentioned in #12169.

We want to keep that for backwards compatibility for now, I think. I've added a note to #10946 about it though!

PR #12177, opened 2022-03-08
Fix a bug introduced in 1.54.0rc1 which meant that Synapse would refuse to start if pre-release versions of dependencies were installed.

### LINKED ISSUE #12176: RC versions of dependencies don't satisfy the run-time dependency checker; `Need Twisted>=18.9.0, but got Twisted==21.7.0rc3` (1.54.0rc1 suspected regression)
When deploying `1.54.0rc1` to matrix.org and some personal homeservers that had an RC of Twisted installed, the dependency checker complained:

`Need Twisted>=18.9.0, but got Twisted==21.7.0rc3`

For some reason it appears that being an RC makes the version insufficient, even though the version is higher. Using the non-RC version works fine.

Possibly fall-out from https://github.com/matrix-org/synapse/pull/12088?

I wonder if the same logic as e.g. `pip` is being used, in that it would never select an RC version as being satisfactory unless it was a hard match?

reivilibre:
Lovely `fix_bug=True` flag, if only it was always that easy :).

Fixes #12176.

### DIFF
--- a/changelog.d/12177.bugfix
+++ b/changelog.d/12177.bugfix
@@ -1,0 +1,1 @@
+Fix a bug introduced in 1.54.0rc1 which meant that Synapse would refuse to start if pre-release versions of dependencies were installed.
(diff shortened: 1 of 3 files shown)


richvdh (approved):
lgtm!

richvdh:
CI will be failing because sytest is relying on stuff not in the release branch.

PR #12106, opened 2022-02-28
Add `stop_cancellation` utility function

squahtx:
This'll be generally useful for #3528.

In particular, it gets used in https://github.com/matrix-org/synapse/commit/ef5273847c5b0a8c71d2071b8299434dd457b026 (to be PRed)

### Pull Request Checklist

* [x] Pull request is based on the develop branch
* [x] Pull request includes a [changelog file](https://matrix-org.github.io/synapse/latest/development/contributing_guide.html#changelog). The entry should:
  - Be a short description of your change which makes sense to users. "Fixed a bug that prevented receiving messages from other servers." instead of "Moved X method from `EventStore` to `EventWorkerStore`.".
  - Use markdown where necessary, mostly for `code blocks`.
  - End with either a period (.) or an exclamation mark (!).
  - Start with a capital letter.
  - Feel free to credit yourself, by adding a sentence "Contributed by @github_username." or "Contributed by [Your Name]." to the end of the entry.
* [x] Pull request includes a [sign off](https://matrix-org.github.io/synapse/latest/development/contributing_guide.html#sign-off)
* [x] [Code style](https://matrix-org.github.io/synapse/latest/code_style.html) is correct
  (run the [linters](https://matrix-org.github.io/synapse/latest/development/contributing_guide.html#run-the-linters))

PR #12180, opened 2022-03-08
Add `delay_cancellation` utility function

squahtx:
`delay_cancellation` behaves like `stop_cancellation` (#12106), except it
delays `CancelledError`s until the original `Deferred` resolves.
This is handy for unifying cleanup paths and ensuring that uncancelled
coroutines don't use finished logcontexts.

Comes from a suggestion by @erikjohnston.

As part of these changes I've ended up updating some of the `stop_cancellation` code and tests.

Generally useful for #3528.

### Pull Request Checklist

### DIFF
--- a/synapse/util/async_helpers.py
+++ b/synapse/util/async_helpers.py
@@ -686,11 +686,12 @@ def stop_cancellation(deferred: "defer.Deferred[T]") -> "defer.Deferred[T]":
             Synapse logcontext rules.
 
     Returns:
-        A new `Deferred`, which will contain the result of the original `Deferred`,
-        but will not propagate cancellation through to the original. When cancelled,
-        the new `Deferred` will fail with a `CancelledError` and will not follow the
-        Synapse logcontext rules. `make_deferred_yieldable` should be used to wrap
-        the new `Deferred`.
+        A new `Deferred`, which will contain the result of the original `Deferred`.
+        The new `Deferred` will not propagate cancellation through to the original.
+        When cancelled, the new `Deferred` will fail with a `CancelledError`.
+
+        The new `Deferred` will not follow the Synapse logcontext rules and should be
+        wrapped with `make_deferred_yieldable`.
     """
     new_deferred: defer.Deferred[T] = defer.Deferred()
     deferred.chainDeferred(new_deferred)
--- a/synapse/util/async_helpers.py
+++ b/synapse/util/async_helpers.py
@@ -698,17 +698,15 @@ def stop_cancellation(deferred: "defer.Deferred[T]") -> "defer.Deferred[T]":
     return new_deferred
 
 
-def delay_cancellation(deferred: "defer.Deferred[T]", all: bool) -> "defer.Deferred[T]":
+def delay_cancellation(deferred: "defer.Deferred[T]") -> "defer.Deferred[T]":
     """Delay cancellation of a `Deferred` until it resolves.
 
     Has the same effect as `stop_cancellation`, but the returned `Deferred` will not
     resolve with a `CancelledError` until the original `Deferred` resolves.
 
     Args:
-        deferred: The `Deferred` to protect against cancellation. Must not follow the
-            Synapse logcontext rules if `all` is `False`.
-        all: `True` to delay multiple cancellations. `False` to delay only the first
-            cancellation.
+        deferred: The `Deferred` to protect against cancellation. May optionally follow
+            the Synapse logcontext rules.
 
     Returns:
         A new `Deferred`, which will contain the result of the original `Deferred`.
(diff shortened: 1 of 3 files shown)


tests/util/test_async_helpers.py
```diff
@@ -313,13 +318,22 @@ async def caller():
...
 
-class StopCancellationTests(TestCase):
-    """Tests for the `stop_cancellation` function."""
+@parameterized_class(
+    ("wrap_deferred",),
+    [
+        (lambda _self, deferred: stop_cancellation(deferred),),
+        (lambda _self, deferred: delay_cancellation(deferred, all=True),),
+    ],
+)
```
squahtx, line 327: This is rather ugly. Alternatives are welcome.

I previously tried an abstract base class, but trial tried to instantiate it.

synapse/util/async_helpers.py
```diff
@@ -686,12 +686,69 @@ def stop_cancellation(deferred: "defer.Deferred[T]") -> "defer.Deferred[T]":
...
+    """Delay cancellation of a `Deferred` until it resolves.
+
+    Has the same effect as `stop_cancellation`, but the returned `Deferred` will not
+    resolve with a `CancelledError` until the original `Deferred` resolves.
+
+    Args:
+        deferred: The `Deferred` to protect against cancellation. Must not follow the
+            Synapse logcontext rules if `all` is `False`.
+        all: `True` to delay multiple cancellations. `False` to delay only the first
+            cancellation.
```
richvdh, line 711: I'm led to wonder whether `all=False` is ever going to be useful? It seems to add complexity here so if we can avoid the need for it that would be good.

tests/util/test_async_helpers.py
richvdh: >  but trial tried to instantiate it.

did you try giving it a different name? AIUI trial picks the things to instantiate based on their name. (IIRC it's things ending in `Test`, `Tests` or `TestCase`.)

richvdh: (otherwise: rather than using a lambda, just have a boolean, and make `wrap_deferred` a real method that inspects the boolean before delegating to `stop_cancellation`/`delay_cancellation`. It's marginally less ugly.)

synapse/util/async_helpers.py
```diff
@@ -686,12 +686,69 @@ def stop_cancellation(deferred: "defer.Deferred[T]") -> "defer.Deferred[T]":
...
+        delay_deferred: "defer.Deferred[T]" = defer.Deferred()
+        deferred.chainDeferred(delay_deferred)
+
+        if all:
+            # Intercept cancellations recursively. Each cancellation will cause another
+            # `Deferred` to be inserted into the chain.
+            delay_deferred.addErrback(cancel_errback)
+
+        # Override the result with the `CancelledError`.
+        delay_deferred.addBoth(lambda _: failure)
```
richvdh, line 747: would it be easier to give `new_deferred` a custom canceller, which swallows the cancellation but sets a flag. Then we add a callback which substitutes the result with a `CancelledError` if the canceller was called?

richvdh (commented):
looks generally good, but a few thoughts

tests/util/test_async_helpers.py
squahtx: > try giving it a different name

Renaming the class doesn't work. trial seems to look for subclasses of `TestCase` rather than by matching on names.

I'm not a fan of using booleans as enums, so I'll switch it to a string. Which is still not an enum but at least makes the meaning clear.

synapse/util/async_helpers.py
squahtx: Good idea, I was starting to wonder the same. Let's remove the option for now and always delay all cancellations.

synapse/util/async_helpers.py
squahtx: A custom canceller was actually my first idea, but I found that [twisted expects the canceller to `.callback()` or `.errback()` the `Deferred`, otherwise it will errback with a `CancelledError` immediately](https://github.com/twisted/twisted/blob/18a7cb39e28b301e5ff76e447a852e04b7d7cb20/src/twisted/internet/defer.py#L731-L741).

The custom canceller would certainly be a lot cleaner if it worked!

squahtx: Actually I have an idea involving `Deferred.pause()`...

synapse/util/async_helpers.py
```diff
@@ -686,12 +686,45 @@ def stop_cancellation(deferred: "defer.Deferred[T]") -> "defer.Deferred[T]":
...
+        resolves before failing with a `CancelledError`.
+
+        The new `Deferred` will follow the Synapse logcontext rules if `deferred`
+        follows the Synapse logcontext rules. Otherwise the new `Deferred` should be
+        wrapped with `make_deferred_yieldable`.
     """
-    new_deferred: defer.Deferred[T] = defer.Deferred()
+
+    def handle_cancel(new_deferred: "defer.Deferred[T]") -> None:
+        new_deferred.pause()
```
richvdh, line 723: a couple of words of explanation might be useful here.

```suggestion
        # when the new deferred is cancelled, we `pause` it to stop the cancellation propagating.
        # we then unpause it once the wrapped deferred completes, to propagate the exception.
        new_deferred.pause()
```

richvdh (approved):
lgtm!

PR #12181, opened 2022-03-08
Convert boolean event content to strings for push rule evaluation

Fizzadar:
Quick change so boolean event fields can be evaluated in push rules (as strings). Alternative could be a different matcher type (`event_match_boolean`) but that felt like overkill.

Signed of by Nick Mills-Barrett <nick@beeper.com>

### Pull Request Checklist

### DIFF
--- a/synapse/push/push_rule_evaluator.py
+++ b/synapse/push/push_rule_evaluator.py
@@ -232,6 +232,8 @@ def _flatten_dict(
     if result is None:
         result = {}
     for key, value in d.items():
+        if isinstance(value, bool):
+            value = str(value)
         if isinstance(value, str):
             result[".".join(prefix + [key])] = value.lower()
         elif isinstance(value, Mapping):
--- a/changelog.d/12181.misc
+++ b/changelog.d/12181.misc
@@ -1,0 +1,1 @@
+Convert event content booleans to strings when evaluating push rules. Contributed by Nick @ Beeper.


richvdh:
This does seem sensible, but I'm undecided how to handle the spec side of this.

As of https://github.com/matrix-org/matrix-doc/pull/3690 (ie, current unstable spec), the spec contains the words:

>  If the property specified by key ... does not have a string value, then the condition will not match.

Obviously, this PR is in conflict with that.

So we've got a couple of ways to proceed:

 * Decide that this was the only sensible way to proceed, that the spec should always have allowed matching on boolean fields, that it was effectively a synapse bug that it didn't, and YOLO another PR to the spec.
 * Decide that this really is a change to existing behaviour, and that we need an MSC to change the spec in this way.

Strictly speaking I think it should be an MSC, but I'm not going to insist on a bunch of process if folks think that's ridiculous.

However, it does bring into question: what should we do about other types of fields? (In particular, numbers and `null`s)

richvdh (changes requested):
just marking this as needing changes for now, while we consider how to proceed

clokep:
> Strictly speaking I think it should be an MSC, but I'm not going to insist on a bunch of process if folks think that's ridiculous.
> 
> However, it does bring into question: what should we do about other types of fields? (In particular, numbers and `null`s)

The inconsistency between non-string types is what really makes me think the behavior should be specced. It really isn't obvious to me that you would want a boolean (or number) value to be matched as a string also.

I'd rather see this go through the MSC process, I think, but would definitely defer to people with more history of how push rules are supposed to work!

deepbluev7:
While I generally do want to be able to match booleans (and other types), I'm not sure if stringifying them is the right approach.

synapse/push/push_rule_evaluator.py
```diff
@@ -232,6 +232,8 @@ def _flatten_dict(
     if result is None:
         result = {}
     for key, value in d.items():
+        if isinstance(value, bool):
+            value = str(value)
```
reivilibre, line 236: I don't like this implementation because it will give `"True"` and `"False"` as strings. Other languages represent them differently. The encoding of booleans as strings should really be specified unambiguously so push rules can be written to work on any homeserver.

Fizzadar: Agreed, the below string check lowercases them but it's not explicit in any way.

reivilibre: ah, well, it feels less bad then (Python's capitalisation is particularly unusual, I guess), but I suppose we still ought to make it explicit in the spec how things should be stringified if this is what we're doing.

Fizzadar:
Given the discussion - definitely feels like this should be an MSC! I've not got much time at the moment but can probably write something up soonish.

> While I generally do want to be able to match booleans (and other types), I'm not sure if stringifying them is the right approach.

I'd also lean this way, string matching on booleans feels like a hack. Same for numbers/null values.

clokep:
> Given the discussion - definitely feels like this should be an MSC! I've not got much time at the moment but can probably write something up soonish.

For cross-linking this is [MSC3758](https://github.com/matrix-org/matrix-spec-proposals/pull/3758).

Fizzadar:
Closing this out, the MSC3758 takes a much better approach (match the value + type).

PR #12187, opened 2022-03-09
Clean-up some unused variables

clokep:
These were noticed while looking at some of the caching descriptors (+ PyCharm flagging them as unused).


squahtx (approved):
lgtm

`current_state_ids` doesn't seem to have ever been used

PR #12189, opened 2022-03-09
Allow for ignoring some arguments when caching.

clokep:
Work for #12130 -- this makes a few changes to our caching mechanisms:

* `@cached` can now take an `uncached_args` which is an iterable of names to not use in the cache key.
* Requires `@cached`, @cachedList` and `@lru_cache` to use keyword arguments for clarity.
* Asserts that keyword-only arguments in cached functions are not accepted. (I tested this briefly and I don't believe this works properly.)

### DIFF
--- a/synapse/util/caches/descriptors.py
+++ b/synapse/util/caches/descriptors.py
@@ -129,7 +129,7 @@ class _CacheDescriptorBase:
 
         self.add_cache_context = cache_context
 
-        self.cache_key_builder = get_cache_key_builder(
+        self.cache_key_builder = _get_cache_key_builder(
             self.arg_names, self.arg_defaults
         )
 
@@ -613,7 +613,7 @@ def cachedList(
     return cast(Callable[[F], _CachedFunction[F]], func)
 
 
-def get_cache_key_builder(
+def _get_cache_key_builder(
     param_names: Sequence[str], param_defaults: Mapping[str, Any]
 ) -> Callable[[Sequence[Any], Mapping[str, Any]], CacheKey]:
     """Construct a function which will build cache keys suitable for a cached function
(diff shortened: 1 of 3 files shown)


synapse/util/caches/descriptors.py
```diff
@@ -69,13 +70,19 @@ def __init__(
...
     ):
         self.orig = orig
 
         arg_spec = inspect.getfullargspec(orig)
         all_args = arg_spec.args
 
+        if arg_spec.kwonlyargs:
+            raise ValueError(
+                "_CacheDescriptorBase does not support keyword-only arguments."
+            )
```
erikjohnston, line 84: Can you add a note as to why? I think its just that its buggy and so needs to be fixed before it can be used?

erikjohnston (approved):

clokep:
This is subtly broken actually, it causing it to sometimes use the wrong argument. 😢 Trying to reproduce in a simpler fashion...

clokep:
@erikjohnston I think this has changed enough (and is fiddly enough) that it makes sense to get another review. I found this while integrating it into #12130 -- when the uncached argument (`event` in my case) was passed as an arg (not a kwarg) then it still got used and would skip one of the expected args.

synapse/util/caches/descriptors.py
```diff
@@ -667,16 +677,21 @@ def _get_cache_key_gen(
     This is essentially the same operation as `inspect.getcallargs`, but optimised so
     that we don't need to inspect the target function for each call.
     """
+    if param_names == ():
+        pass
```
erikjohnston, line 681: I assume this is spurious and can be deleted?

erikjohnston (approved):

PR #12192, opened 2022-03-09
Rename get_tcp_replication.

clokep:
It is used for what we call "TCP replication" (as opposed to HTTP replication), but it can go over direct TCP or Redis which makes it rather confusing.

Regardless, all the other methods are named `get_class_name`, while this returns something completely different (`ReplicationCommandHandler`, not a `TcpReplication`).

### DIFF
--- a/synapse/app/generic_worker.py
+++ b/synapse/app/generic_worker.py
@@ -417,7 +417,7 @@ class GenericWorkerServer(HomeServer):
             else:
                 logger.warning("Unsupported listener type: %s", listener.type)
 
-        self.get_tcp_replication().start_replication(self)
+        self.get_replication_command_handlere().start_replication(self)
 
 
 def start(config_options: List[str]) -> None:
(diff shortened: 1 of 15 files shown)


synapse/app/generic_worker.py
```diff
@@ -417,7 +417,7 @@ def start_listening(self) -> None:
             else:
                 logger.warning("Unsupported listener type: %s", listener.type)
 
-        self.get_tcp_replication().start_replication(self)
+        self.get_replication_command_handlere().start_replication(self)
```
erikjohnston, line 420: It's *so* nice to discover someone else making a stupid typo everywhere, usually its just me :D  (Also, I enjoy the idea of naming "handler" "handle here" :p)

```suggestion
        self.get_replication_command_handler().start_replication(self)
```

clokep: Hah, that's what I get for trying to do this quickly!

erikjohnston (approved):

PR #12193, opened 2022-03-09
Start application service stream token tracking from 1

anoadragon453:
Warning: this explanation is a bit complex. Let me know if by reading to the end of it it'd be easier to explain over voice.

---

An off-by-one error. Pulled out of #11881, though I felt it deserved a bit more attention than being buried in there.

The `application_services_state` table is used, among other things, to track the last stream token (5, 6, 7...) of a given stream type (presence, typing, device lists, etc.) that a given application service has previously been informed of. The `ApplicationServiceHandler` then uses this to determine the set (from the stored stream token up to the latest known stream token) of new presence updates, typing notifications, etc. to send to the application service.

For example, say this is the stream of presence updates that have occurred:

```
presence update stream tokens:
	2 3 4 5 6 7 8
```

the `application_services_state` helps track which events have already been sent out to a given appservice:

```
presence update stream tokens:
	2 3 4 5 6 7 8
        |
       AS1
```

with this, the `ApplicationServiceHandler` knows that it needs to send out the presence updates with stream tokens 4-8 the next time.

I noticed an edge case during #11881, in that the default value for a stream token across Synapse is 1: `StreamtokenGenerator` sequences start at `step` (defined as 1 for everything except backfill):

https://github.com/matrix-org/synapse/blob/032688854babeea832cbb4f762fc70fe31e73cc6/synapse/storage/util/id_generators.py#L86-L89

`MultiWritertokenGenerator` sequences also start at 1 (though the first entity will actually start at 2):

https://github.com/matrix-org/synapse/blob/10a88ba91cb16ccf757984f0a7d41ddf8b4dc07f/synapse/storage/util/id_generators.py#L344-L349

As an aside, when we create an event, we usually ask for the *next* available ID, which is initially 2. So the minimum stream token for any entity is 2.

This is inconsistent with `get_type_stream_id_for_appservice`, which assumes stream tokens start from 0 by default:

https://github.com/matrix-org/synapse/blob/2cc5ea933dbe65445e3711bb3f05022b007029ea/synapse/storage/databases/main/appservice.py#L448-L449

Why would that be a problem? Well, it turns out some bits of the codebase assume stream token positions start at 1, and checking against 0 can cause bugs. I ran into one here:

https://github.com/matrix-org/synapse/blob/53f9d018ac369c0b89145ec19a71eaaaacb8ea17/synapse/storage/databases/main/devices.py#L702-L708

where we call `get_all_entities_changed` with a minimum stream token to figure out which user tokens have had their device lists changed. This method has a conditional here:

https://github.com/matrix-org/synapse/blob/f8d0f72b27e158738f3c75a38399b967f7478011/synapse/util/caches/stream_change_cache.py#L147-L156

that we kept unintentionally hitting, as `self._earliest_known_stream_pos` is 1 by default (it is derived from the current stream token when the homeserver starts). Thus this function was returning `None`, even though there was a device list change (at stream token 2) that had occurred, and that our appservice should know about.

---

The impact of this is that the appservice would never receive word of that first device list change, as `from_key` would then be incremented.

For any positive impacts on existing EDU -> AS code, I can confirm that presence's logic has a similar bug: `from_key` would be `0` here:

https://github.com/matrix-org/synapse/blob/e24ff8ebe3d4119d377355402245947f7de61c00/synapse/handlers/presence.py#L1630-L1660

which looks like it would lead to sending all known presence states to the AS (#10836?).

Typing didn't have this bug as we don't record stream tokens for that, and receipts is not affected either as it just applies the range to the database without a stream cache potentially bailing out in the middle.


erikjohnston:
Well this is fun, thanks for the explanation! The fix certainly makes sense.

Though I wonder if we actually *want* to send all the current presence updates to the application service on first start? Otherwise it won't ever see people that are e.g. always online.

It's worth noting that the code has specifically chosen to send all presence updates to the streamer if it falls too far behind. If that is much of a problem then we should also fix that.

anoadragon453:
@erikjohnston Thanks for taking a look!

By "all known presence states" above, I meant that the AS would be receiving presence updates for *every user*, not just *users the AS is interested in receiving the presence of*.

Looking through again though, I think I misread this bit of code as including *every user*, which it isn't.

https://github.com/matrix-org/synapse/blob/e24ff8ebe3d4119d377355402245947f7de61c00/synapse/handlers/presence.py#L1657-L1660

So actually, this change is removing presence updates from the initial burst for users that have never set a presence state. Instead of receiving an "offline" state for them, they'll just be omitted. This _is_ a change in behaviour, but probably isn't a problem as the AS would just assume those users would be offline anyways?

For clarity, we're now going through the truthy body of this conditional, whereas before `from_key` would be zero, and we'd drop down to the falsey:

https://github.com/matrix-org/synapse/blob/e24ff8ebe3d4119d377355402245947f7de61c00/synapse/handlers/presence.py#L1630-L1660

...I'm going to change that to `if from_key is not None` here as well to avoid it catching 0 at some other point.

synapse/storage/databases/main/appservice.py
```diff
@@ -446,7 +446,8 @@ def get_type_stream_id_for_appservice_txn(txn):
             )
             last_stream_id = txn.fetchone()
             if last_stream_id is None or last_stream_id[0] is None:  # no row exists
-                return 0
+                # stream tokens always start from 1
```
erikjohnston, line 449:
```suggestion
                # Stream tokens always start from 1, to avoid foot guns around `0` being falsey.
```

changelog.d/12193.bugfix
```diff
@@ -0,0 +1 @@
+Fix a long-standing bug where excess presence events may be sent to an application service after it is initially configured.
```
erikjohnston, line 1: This is probably no longer a bug?

erikjohnston (approved):

PR #12195, opened 2022-03-09
Allow modules to store already existing 3PID associations

babolivier:
Part of the Tchap mainlining. On Tchap, users register with their work email address, which is automatically associated on Sydent using Sydent's internal API. These associations then need to be stored in Synapse's database. The idea is to move that into a module that implements [`on_user_registration`](https://matrix-org.github.io/synapse/latest/modules/account_validity_callbacks.html#on_user_registration), binds the 3pid, and then inserts the association into the database.

Ideally I would have copied over `add_user_bound_threepid` into the module, but it requires doing an upsert which isn't trivial from a module.

### DIFF
--- a/synapse/module_api/__init__.py
+++ b/synapse/module_api/__init__.py
@@ -659,7 +659,8 @@ class ModuleApi:
     def record_user_external_id(
         self, auth_provider_id: str, remote_user_id: str, registered_user_id: str
     ) -> defer.Deferred:
-        """Record a mapping from an external user id to a mxid
+        """Record a mapping between an external user id from a single sign-on provider
+        and a mxid.
 
         Added in Synapse v1.9.0.
 
@@ -1281,7 +1282,7 @@ class ModuleApi:
 
         The association must already exist on the remote identity server.
 
-        Added in Synapse v1.55.0.
+        Added in Synapse v1.56.0.
 
         Args:
             user_id: The user ID that's been associated with the 3PID.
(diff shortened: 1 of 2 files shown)


synapse/module_api/__init__.py
```diff
@@ -1274,6 +1274,24 @@ async def check_username(self, username: str) -> None:
         """
         await self._registration_handler.check_username(username)
 
+    async def store_remote_3pid_association(
```
anoadragon453, line 1277: From the eyes of a module developer, this method may look confusingly similar to `record_user_external_id`, in that both "store a mapping between a user ID and a third-party identifier".

It may help to update the docstring of `record_user_external_id` to clarify exactly what system (authentication through third-party Single Sign-on providers) that it's interacting with.

synapse/module_api/__init__.py
```diff
@@ -1274,6 +1274,24 @@ async def check_username(self, username: str) -> None:
...
+        The association must already exist on the remote identity server.
+
+        Added in Synapse v1.55.0.
+
+        Args:
+            user_id: The user ID that's been associated with the 3PID.
+            medium: The medium of the 3PID (current supported values are "msisdn" and
+                "email").
+            address: The address of the 3PID.
+            id_server: The identity server the 3PID association has been registered on.
```
anoadragon453, line 1291: It's probably worth mentioning that this should be in the form `domain.com` and should not include a protocol scheme nor path.

synapse/module_api/__init__.py
```diff
@@ -1274,6 +1274,24 @@ async def check_username(self, username: str) -> None:
...
+        """
+        await self._store.add_user_bound_threepid(user_id, medium, address, id_server)
```
anoadragon453, line 1293: Do we want to return whether an upsert occurred (`add_user_bound_threepid` returns a bool).

babolivier: > `add_user_bound_threepid` returns a bool

It doesn't.

anoadragon453: Oops, I see.

Do we want it to?

babolivier: Don't think so. If we realise we do then it's probably something we can figure out later.

anoadragon453: I suppose since we don't return anything from `store_remote_3pid_association` at the moment, making it return something in the future isn't really a backwards-incompatible change. So we could decide we'd want to later without much issue.

Issue #11584, opened 2021-12-15
Remove groups/communities support from Synapse


clokep:
The groups/communities feature in Synapse is unspecced (defined in [MSC1513](https://github.com/matrix-org/matrix-doc/issues/1513), kind of, but is abandoned) -- spaces ([MSC1772](https://github.com/matrix-org/matrix-doc/pull/1772) and friends).

There are still some useful features driven by Groups (namely flair) so it is unclear when this will happen. Note that Element Web has the capability to convert a Group -> Space.

A rough plan of how we might want to do this is:

* [x] Add an experimental config flag for groups that guards the unstable endpoints for groups.
* [x] Add a deprecation notice to our release and upgrade notes. (Should probably be done at the same time as 1.)
* [x] Default the above flag to `false` to disable those endpoints at the end of the deprecation period. The configuration flag should emit a warning if it is enabled.
* [x] (After some time...) Remove the groups code.
	* Endpoints
	* Handlers
	* Update the worker documentation
	* `RelatedGroups` constant
* [ ] (After some time...) Bump the schema and remove the database tables for groups.

Related to #8334.

clokep:
I'm not really sure if there's much benefit to step 3 -- we might want to just jump over it and do step 4 at the end of the deprecation period.

clokep:
Removal is waiting for Synapse 1.61.

clokep:
Almost all of the groups related code I could find has now been removed. There's a tiny bit left which will be handled in #12940.

PR #12200, opened 2022-03-09
Deprecate the groups/communities endpoints

clokep:
The first part of #11584, this:

1. Adds an experimental configuration flag for groups (but enables it by default).
2. Places the groups/communities endpoints behind the configuration flag.
3. Documents that groups/communities are deprecated.

### DIFF
--- a/changelog.d/12200.removal
+++ b/changelog.d/12200.removal
@@ -1,0 +1,1 @@
+The groups/communities feature in Synapse has been deprecated.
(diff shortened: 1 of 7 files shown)


H-Shay (commented):
This appears sane to me.

H-Shay (approved):

clokep:
Thanks for the review!

PR #12202, opened 2022-03-10
Retention test: avoid relying on state at purged events

richvdh:
This test was relying on poking events which weren't in the database into `filter_events_for_client`, which is conflicting with my changes elsewhere.

The intention of the test is to see what happens when `filter_events_for_client` is called on events which should have expired, but are still in the db. So let's *actually* test that by leaving them in the db!


clokep (approved):
Seems reasonable to me!

PR #12206, opened 2022-03-10
Remove unnecessary passes.

clokep:
I was attempting to use `autoflake` to find unused variables, but it just removed a bunch of unneeded `pass` calls. I see no reason not to do this though. 🤷


erikjohnston (approved):
sure

PR #12209, opened 2022-03-11
Use a sequence to generate AS transaction IDs, drop `last_txn` AS state

### LINKED ISSUE #11620: SerializationFailure on appservice outgoing transaction (updating `last_txn`)

Fizzadar:
Drafting this pending any concerns and will then look into tests. Aims here:

+ fix #11620 without altering transaction isolation levels
+ precursor to parallising the `appservice` workload across multiple worker instances

Switching to a sequence means there's no need to track `last_txn` on the AS state table to generate new TXN IDs. This also means that there is no longer contention between the AS scheduler and AS handler on updates to the `application_services_state` table, which will prevent serialization errors during the complete AS txn transaction. This should properly fix #11620 without the need for dropping the TXN isolation levels.

NOTE: we have been running the TXN isolation level change in https://github.com/matrix-org/synapse/pull/12032 successfully for a couple of weeks now. Since making the change we have not seen a single instance of the debug line removed in this PR. My conclusion is that the tracking of last TXN ID and the associated debug statement actually cause the problem due to the contention on the AS state table writes. (I believe this all pre-dates the use of sequences).

Another thing of note is this means all AS's share the same sequence; I don't think there's any actual issues with this other than the TXN ID "jumping" between as specific AS' transactions - the spec doesn't specify these need to be sequential. (if needed one could generate a sequence-per-AS, but that feels wrong!).

If this is all good, I'll update the tests and work on the required DB migration so this can be rolled out.

Signed off by Nick Mills-Barrett <nick@beeper.com>

### Pull Request Checklist


clokep:
@Fizzadar requested an initial review on this:

> basically a quick confirm I've not missed something obvious that would prevent switching to a sequence generator before I look into tests and the DB migration required as I suspect that'll take longer than the changes so far!

synapse/storage/databases/main/appservice.py
```diff
@@ -72,6 +74,21 @@ def __init__(
         )
         self.exclusive_user_regex = _make_exclusive_regex(self.services_cache)
 
+        def get_max_as_txn_id(txn: Cursor) -> int:
+            txn.execute(
+                "SELECT COALESCE(max(txn_id), 0) FROM application_services_txns"
+            )
```
richvdh, line 80: problem here is that it will mean a table-scan of `application_services_txns` on startup - which could be very slow.

That said, this is only used on SQLite, and tbh anyone running a busy AS with sqlite probably needs to move over to postgres anyway.

So perhaps we can just log a warning ("falling back to slow query, you need to port to postgres") if we ever hit this path.

synapse/storage/databases/main/appservice.py
```diff
@@ -285,40 +288,10 @@ async def complete_appservice_txn(
...
-                    service.id,
-                )
-
-            # Set current txn_id for AS to 'txn_id'
-            self.db_pool.simple_upsert_txn(
-                txn,
-                "application_services_state",
-                {"as_id": service.id},
-                {"last_txn": txn_id},
-            )
```
richvdh, line 311: as a general rule, we'd generally prefer to make this sort of change over two releases of Synapse. In the first release, we'd stop reading from the `last_txn` column; we would then follow up in a second release by removing the code that *writes* to `last_txn`.  If we combine the changes into one release, it means that anybody who has a problem with the release and has to roll back will have problems (in this case, I think all the AS txns since upgrade will be replayed).

So for bonus points, it would be great to split this into two changes.

richvdh (commented):
sorry for the delay on this, @Fizzadar.

The approach here looks very sensible. I think the only reason we're using a field in `application_services_state` is that this code is so old it predates the use of sequences anywhere in Synapse.

synapse/storage/databases/main/appservice.py
Fizzadar: https://github.com/matrix-org/synapse/pull/12209/commits/cceb2b9caa4bfd0137e46b3bf4328d151084a6cb

synapse/storage/databases/main/appservice.py
Fizzadar: V.sensible! https://github.com/matrix-org/synapse/pull/12209/commits/722f6ce13e86402c47ec58b3b24964f93851828b

synapse/storage/schema/main/delta/68/06as_txn_seq.py
```diff
@@ -0,0 +1,48 @@
...
+        start_val = max(last_txn_max, txn_max) + 1
+
+        cur.execute(
+            "CREATE SEQUENCE application_services_txn_id_seq START WITH %s",
+            (start_val,),
+        )
+
+
+def run_upgrade(*args, **kwargs):
+    pass
```
richvdh, line 48: this is no longer required.

```suggestion
```

synapse/storage/schema/main/delta/68/06as_txn_seq.py
```diff
@@ -0,0 +1,48 @@
+# Copyright 2022 Beeper
```
richvdh, line 1: it would be good to bump `synapse.storage.schema.SCHEMA_VERSION` (and move this to `69/`). Then in future, when we remove the upsert on `application_services_state`, we can bump `SCHEMA_COMPAT_VERSION` to ensure that nobody tries rolling back to a version of synapse that relies on it.

richvdh (commented):
lgtm otherwise (though I'm afraid a couple of conflicts have slipped in).

synapse/storage/schema/main/delta/68/06as_txn_seq.py
Fizzadar: https://github.com/matrix-org/synapse/pull/12209/commits/d62adeee3e503b0094b62f21cf20e141687138fb

synapse/storage/schema/main/delta/68/06as_txn_seq.py
Fizzadar: Covered in https://github.com/matrix-org/synapse/pull/12209/commits/d62adeee3e503b0094b62f21cf20e141687138fb

Fizzadar:
Updated per comments, merged in develop and added a note to the changelog about the upgrade!

changelog.d/12209.misc
```diff
@@ -1 +1 @@
-Switch to using a sequence to generate AS transaction IDs. Contributed by Nick Beeper.
+Switch to using a sequence to generate AS transaction IDs. Contributed by Nick Beeper. If running synapse with a dedicated appservice worker, this MUST be shutdown before upgrading the main process and database.
```
richvdh, line 1:
```suggestion
Switch to using a sequence to generate AS transaction IDs. Contributed by Nick Beeper. If running synapse with a dedicated appservice worker, this MUST be stopped before upgrading the main process and database.
```

richvdh (approved):
thanks very much for this, @Fizzadar!

I've added an entry to the upgrade notes to warn about shitting down the AS worker.

Issue #12205, opened 2022-03-10
Tweak Sign In With Apple/SSO copy to make required data explicit


nadonomy:
![d84f2561-607c-45f2-9d43-35c1784e7cca_005](https://user-images.githubusercontent.com/4626865/157723026-0e224f19-2e07-445d-b628-4fec50cd11a7.png)

Above is a screenshot of what users see today after tapping 'Sign In With Apple' (SIWA) in iOS. According to [Apple's guidelines](https://developer.apple.com/design/human-interface-guidelines/sign-in-with-apple/overview/data-management/), we need to make explicit the data is required vs an optional request.

- [ ] When using SIWA, update the headline copy
         from 'Your account is nearly ready'
         to 'Choose your account name'
- [ ] When using SIWA, update the byline copy
         from 'Check your details before creating an account on server'
         to 'This is required to create your account, and you can't change this later.'
- [ ] When using SIWA, prefill the username with whatever Apple has passed us

nadonomy:
I'd also welcome us changing the copy generically if that's the preferred path to avoid special casing this for one SSO provider.

babolivier:
> I'd also welcome us changing the copy generically if that's the preferred path to avoid special casing this for one SSO provider.

This is a generic page (not special-cased to Apple, the logo is just templated in), so if we change the copy there the change will propagate across all of the SSO providers

PR #12210, opened 2022-03-11
Update the SSO username picker template to comply with SIWA guidelines

babolivier:
Fixes https://github.com/matrix-org/synapse/issues/12205

cc @reivilibre

### DIFF
--- a/synapse/res/templates/sso_auth_account_details.html
+++ b/synapse/res/templates/sso_auth_account_details.html
@@ -138,7 +138,7 @@
         <div class="username_input" id="username_input">
           <label for="field-username">Username</label>
           <div class="prefix">@</div>
-          <input type="text" name="username" id="field-username" value="{{ user_attributes.remote_user_id }}" autofocus>
+          <input type="text" name="username" id="field-username" value="{{ user_attributes.localpart }}" autofocus>
           <div class="postfix">:{{ server_name }}</div>
         </div>
         <output for="username_input" id="field-username-output"></output>
--- a/synapse/res/templates/sso_auth_account_details.html
+++ b/synapse/res/templates/sso_auth_account_details.html
@@ -130,15 +130,15 @@
   </head>
   <body>
     <header>
-      <h1>Your account is nearly ready</h1>
-      <p>Check your details before creating an account on {{ server_name }}</p>
+      <h1>Choose your user name</h1>
+      <p>This is required to create your account on {{ server_name }}, and you can't change this later.</p>
     </header>
     <main>
       <form method="post" class="form__input" id="form">
         <div class="username_input" id="username_input">
           <label for="field-username">Username</label>
           <div class="prefix">@</div>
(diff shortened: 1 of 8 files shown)


synapse/rest/synapse/client/pick_username.py
```diff
@@ -98,6 +98,9 @@ async def _async_render_GET(self, request: Request) -> None:
             "user_attributes": {
                 "display_name": session.display_name,
                 "emails": session.emails,
+                "remote_user_id": map_username_to_mxid_localpart(
```
richvdh, line 101: I think these params are documented somewhere (possibly the sample config?), so the doc will need an update

babolivier: Oh good point.

synapse/res/templates/sso_auth_account_details.html
```diff
@@ -130,15 +130,15 @@
...
+      <h1>Choose your user name</h1>
+      <p>This is required to create your account on {{ server_name }}, and you can't change this later.</p>
     </header>
     <main>
       <form method="post" class="form__input" id="form">
         <div class="username_input" id="username_input">
           <label for="field-username">Username</label>
           <div class="prefix">@</div>
-          <input type="text" name="username" id="field-username" autofocus>
+          <input type="text" name="username" id="field-username" value="{{ user_attributes.remote_user_id }}" autofocus>
```
richvdh, line 141: I'm a bit concerned that this might not be a sensible thing to do for non-Apple providers (often the remote_user_id is an entirely opaque string), so I think it would be better to make this optional.

Also: if your remote_user_id matches an MXID that is already registered (presumably via direct registration or a different SSO provider), we'll prepopulate the form with an invalid option, which is not great UX.

One way to do this: The `oidc_config` already has a field `localpart_template`, which defines a way to build a localpart from the remote details (and handles clashing MXIDs). However, the act of setting it means that the user is not given the opportunity to change the generated MXID. So, we could add another field to the config `confirm_generated_localpart: true|false` or somesuch, which means we'll show this page even if `localpart_template` is set.

richvdh (changes requested):

synapse/res/templates/sso_auth_account_details.html
babolivier: > often the remote_user_id is an entirely opaque string

Yes indeed, I've just realised I've misunderstood what it's meant to be.

Your suggestion looks sound, I'll go with that, thanks!

richvdh (approved):
lgtm otherwise

babolivier:
I've had a chat out-of-bound with @nadonomy and it sounds like the changes are fine, so let's merge it and get it deployed it soon.

nadonomy (commented):
@babolivier @callahad sorry PEBKAC on my behalf for not submitting this before. Would be great to get this right given the pedantry at review time.

PR #12212, opened 2022-03-11
Implement a Jinja2 filter to extract localparts from email addresses

babolivier:
When discussing config changes for https://github.com/matrix-org/synapse/issues/12205, we concluded that we want to use email address localparts. We don't have a filter for that currently, so this PR adds one:

```python
>>> from synapse.util.templates import _localpart_from_email_filter
>>> from jinja2.environment import Environment
>>> env = Environment()
>>> env.filters.update({"localpart_from_email": _localpart_from_email_filter})
>>> t = env.from_string("{{ email|localpart_from_email }}")
>>> t.render(email="foo@bar.com")
'foo'
>>> t.render(email="foo@bar.com@example.com")
'foo@bar.com'
>>> 
```

### DIFF
--- a/docs/templates.md
+++ b/docs/templates.md
@@ -36,6 +36,13 @@ Turns a `mxc://` URL for media content into an HTTP(S) one using the homeserver'
 
 Example: `message.sender_avatar_url|mxc_to_http(32,32)`
 
+```python
+localpart_from_email(address: str) -> str
+```
+
+Returns the local part of an email address (e.g. `alice` in `alice@example.com`).
+
+Example: `user.email_address|localpart_from_email`
 
 ## Email templates
 
--- a/changelog.d/12212.feature
+++ b/changelog.d/12212.feature
@@ -1,0 +1,1 @@
+Add a new Jinja2 template filter to extract the local part of an email address.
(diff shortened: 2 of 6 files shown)


reivilibre:
How does this work for e-mail addresses whose localparts aren't valid MXIDs? (e.g. capitalisation, strange symbols, and so on)

E-mail address localparts can be pretty strange (I think they can even be quoted!) but it's probably good enough to strip/replace invalid characters

babolivier:
> How does this work for e-mail addresses whose localparts aren't valid MXIDs? (e.g. capitalisation, strange symbols, and so on)

We already sanitise whatever we're getting from the IdP before we use it as an MXID localpart:

https://github.com/matrix-org/synapse/blob/1d4c5c230a4b02f913fb944c06e30a12c4d1b488/synapse/handlers/oidc.py#L1393

> E-mail address localparts can be pretty strange (I think they can even be quoted!)

They can! I've seen Thunderbird do that when I was doing a bit of testing earlier.

reivilibre (approved):
Looks good to me, thanks :)

docs/templates.md
```diff
@@ -36,6 +36,13 @@ Turns a `mxc://` URL for media content into an HTTP(S) one using the homeserver'
 
 Example: `message.sender_avatar_url|mxc_to_http(32,32)`
 
+```python
+localpart_from_email(address: str) -> str
+```
+
+Returns the local part of an email address (e.g. `alice` in `alice@example.com`).
+
+Example: `user.email_address|localpart_from_email`
```
richvdh, line 45: if this is only available in the SSO templates, we should probably say so.

That said, is there any reason for it only to be available in the SSO templates?

richvdh:
oh bother, too late

PR #12213, opened 2022-03-11
Prevent a sync request from removing a user's busy presence status

dbkr:
In trying to use the MSC3026 busy presence status, the user's status
would be set back to 'online' next time they synced. The change makes
it so that sycning does not affect a user's presence status if it
is currently set to 'busy': it must be removed through the presence
API.

The MSC defers to implementations on the behaviour of busy presence,
so I think this remains compatible with the MSC.

I've replaced the explicit presence-set call from the sync code and moved
this logic into the presence handler, which feels like a better place for it. I've
added some tests to assert the behaviour remains what it's supposed to be.

I haven't added this code to the worker version of user_syncing as there
was no code there to set presence status before, but having removed the
set_state from sync, I'm not certain whether it needs to be there or not,
so I'd appreciate a check from someone who knows the worker setup better.

### Pull Request Checklist

* [ ] Pull request includes a [sign off](https://matrix-org.github.io/synapse/latest/development/contributing_guide.html#sign-off)


deepbluev7:
What happens if a user explicitly sets the presence to online in the sync request in one client and the other client had set the presence to busy, but didn't sync for a while? Shouldn't clients instead set their presence explicitly in /sync every time and then the max calculated from that? Because one of my clients will always be busy, when I have 2 open at the same time.

synapse/handlers/presence.py
```diff
@@ -151,7 +151,7 @@ def __init__(self, hs: "HomeServer"):
 
     @abc.abstractmethod
     async def user_syncing(
-        self, user_id: str, affect_presence: bool
+        self, user_id: str, affect_presence: bool, presence_state: str
```
squahtx, line 154: The docstring below could use an update now that there's a new argument.
The same for the rest of the argument changes in this PR.

synapse/replication/http/presence.py
```diff
@@ -114,7 +114,6 @@ async def _handle_request(  # type: ignore[override]
         await self._presence_handler.set_state(
             UserID.from_string(user_id),
             content["state"],
-            content["ignore_status_msg"],
```
squahtx, line 117: Can `ignore_status_msg` be removed from `_serialize_payload` above as well?

synapse/handlers/presence.py
```diff
@@ -985,7 +986,8 @@ async def user_syncing(
                 await self._update_states(
                     [
                         prev_state.copy_and_replace(
-                            last_user_sync_ts=self.clock.time_msec()
+                            last_active_ts=self.clock.time_msec(),
+                            last_user_sync_ts=self.clock.time_msec(),
                         )
                     ]
                 )
```
squahtx, line 993: > I haven't added this code to the worker version of user_syncing as there
> was no code there to set presence status before, but having removed the
> set_state from sync, I'm not certain whether it needs to be there or not,
> so I'd appreciate a check from someone who knows the worker setup better.

We're going to have to update the worker version of `user_syncing` too, otherwise single process synapse and synapse with workers will behave differently.

In worker mode, `self.presence_handler` is a `PresenceHandler` on the worker configured as the presence writer and `WorkerPresenceHandler` on other workers.

eg. on a matrix.org-like setup, `/sync`s would arrive at a worker with a `WorkerPresenceHandler`. `WorkerPresenceHandler.user_syncing` would call `mark_as_coming_online`, which calls `send_user_sync`, which calls `ReplicationCommandHandler.send_user_sync`. The data then arrives on the presence writer at `ReplicationCommandHandler.on_USER_SYNC`, which calls `PresenceHandler.update_external_syncs_row`. `update_external_syncs_row` looks a lot like `user_syncing` and we should probably update this whole code path too.

synapse/handlers/presence.py
dbkr: Done (where the was docstring for any of the args)

synapse/replication/http/presence.py
dbkr: Yeah, I would assume so

synapse/handlers/presence.py
squahtx: Thanks for keeping the docstrings updated!

synapse/replication/http/presence.py
squahtx: Can we remove the parameter too? I believe it was only used in the call to `_set_state_client()` in `WorkerPresenceHandler.set_state()`, which has already been updated.

squahtx (commented):
Thanks for updating the PR!

The worker mode changes look like a step in the right direction. There's a slight discrepancy in  whether `last_active_ts` gets updated between worker and non-worker mode. I'm not sure whether it's important to us. I guess it depends on how clients want to use it?

**Before this PR**
| prev_state                        | set_presence  | new presence set | last_active_ts updated | last_user_sync_ts updated |
|-----------------------------------|---------------|------------------|------------------------|---------------------------|
| {offline,unavailable,online,busy} | offline       | ❌                | ❌                      | ❌                         |
| {offline,unavailable,online,busy} | unavailable   | ✔️                | ❌                      | ✔️                         |
| {offline,unavailable,online,busy} | {online,busy} | ✔️                | ✔️                      | ✔️                         |

**After this PR, single process**
| prev_state                   | set_presence              | new presence set | last_active_ts updated | last_user_sync_ts updated |
|------------------------------|---------------------------|------------------|------------------------|---------------------------|
| {offline,unavailable,online} | offline                   | ❌                | ❌                      | ❌                         |
| {offline,unavailable,online} | {unavailable,online,busy} | ✔️                | ✔️                      | ✔️                         |
| busy                         | offline                   | ❌                | ❌                      | ❌                         |
| busy                         | {unavailable,online,busy} | ❌                | ✔️                      | ✔️                         |

**After this PR, with workers**
| prev_state                   | set_presence       | new presence set | last_active_ts updated | last_user_sync_ts updated |
|------------------------------|--------------------|------------------|------------------------|---------------------------|
| {offline,unavailable,online} | offline            | ❌                | ❌                      | ❌                         |
| {offline,unavailable,online} | unavailable        | ✔️                | ❌ (!)                  | ✔️                         |
| {offline,unavailable,online} | {online,busy}      | ✔️                | ✔️                      | ✔️                         |
| busy                         | offline            | ❌                | ❌                      | ❌                         |
| busy                         | unavailable        | ❌                | ❌ (!)                  | ✔️                         |
| busy                         | {online,busy}      | ❌                | ❌ (!)                  | ✔️                         |

synapse/replication/http/presence.py
dbkr: ah, good spot - done

synapse/handlers/presence.py
```diff
@@ -471,6 +476,16 @@ async def user_syncing(
         if not affect_presence or not self._presence_enabled:
             return _NullContextManager()
 
+        if affect_presence:
+            prev_state = await self.current_state_for_user(user_id)
+            if prev_state != PresenceState.BUSY:
+                # XXX: Why does this pass force_notify = True? This is copied
+                # from when the sync rest handler set the presence itself, but
+                # I don't really understand why this would be necessary.
```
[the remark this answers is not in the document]
dbkr: ah, good spot! You're right, and yes, I think we do need it since we need to not affect / clear the status message. I've added the param back.

dbkr:
Thanks for going through working out what all the various behaviours were. It's a good point that the last_active_ago would have been affected which was probably not the correct thing. I've updated this for the non-worker case. I was hoping to move this into a common method between the worker & non-worker cases to use the same logic, but the worker doesn't have `_update_states` so it's not as simple as that. I'll look at this again tomorrow - hopefully it can be done without adding a new replication endpoint.

dbkr:
Having looked at this some more, I've realised that the previous behaviour always updated last_active_ts because set_state() updates last_active_ts, so it looks like all the logic in user_syncing to update or not update last_active_ts isn't really doing anything, as far as I can see. Testing this by writing a sytest seems to confirm this: performing a normal sync request (previously online, set_presence=online) caused last_active_ago to be updated. I'm guessing this is a bug, at this point I'm inclined to leave that bug for a different PR. I can leave it as it is and leave the behaviour 'correct' (I think) in single-process mode and keeping the old behaviour in worker mode, or use set_state everywhere and keep the existing behaviour in both modes - any preference?

squahtx:
Hm, tricky. I'd prefer that we keep the single process and worker mode behaviour the same, so that it's easier to reason about, even if it's subtly wrong.

ie. make `PresenceHandler.user_syncing` and `PresenceHandler.update_external_syncs_row` do the same thing, and call `set_state` (somewhere) in both configurations.

dbkr:
Yeah, agreed there's strong argument for keeping it consistent and fixing the bug everywhere separately. Done, and filed https://github.com/matrix-org/synapse/issues/12424 to track the incorrect behaviour.

synapse/handlers/presence.py
```diff
@@ -967,27 +988,55 @@ async def user_syncing(
...
                     [
                         prev_state.copy_and_replace(
-                            last_user_sync_ts=self.clock.time_msec()
+                            # We do want to update state here but currently do so with separate
+                            # set_state below.
+                            # state=presence_state,
+                            last_user_sync_ts=self.clock.time_msec(),
                         )
                     ]
                 )
```
squahtx, line 1028: I'm finding this very confusing (sorry!). Could we do the following refactoring:
1. Move the `if prev_state.state != PresenceState.BUSY:` block before all this, but before `await self.current_state_for_user`. That way it matches the order of things in worker mode. It's not so much the order that matters, but the value of `prev_state.state` when the `if prev_state.state == PresenceState.OFFLINE` block gets run.
2. Roll back this block of changes to bring `user_syncing` and `update_external_syncs_row` back into line.
3. Replace `prev_state.state` with the presence after the `set_state` call, because in worker mode, we'd be reading the presence after the call to `set_state`.

ie.
```py
            prev_state = await self.current_state_for_user(user_id)

            if prev_state.state != PresenceState.BUSY:
                ...
                await self.set_state(
                    UserID.from_string(user_id), {"presence": presence_state}, True
                )
                prev_state = prev_state.copy_and_replace(state=presence_state)

            if prev_state.state == PresenceState.OFFLINE:
                # If they're currently offline then bring them online, otherwise
                # just update the last sync times.
                await self._update_states(
                    [
                        prev_state.copy_and_replace(
                            state=PresenceState.ONLINE,
                            last_active_ts=self.clock.time_msec(),
                            last_user_sync_ts=self.clock.time_msec(),
                        )
                    ]
                )
            else:
                await self._update_states(
                    [
                        prev_state.copy_and_replace(
                            last_user_sync_ts=self.clock.time_msec()
                        )
                    ]
                )
```

which is analogous to the following code path in worker mode:
```py
        prev_state = await self.current_state_for_user(user_id)
        if prev_state != PresenceState.BUSY:
            await self.set_state(
                UserID.from_string(user_id), {"presence": presence_state}, True
            )

        # followed by a call to `update_external_syncs_row`
```

dbkr: Yeah, I think I see how that makes more sense - done

dbkr: ...although this was then re-checking the value of `prev_state.state` after modifying it, so I've made it make a copy in a new variable before making modifications

squahtx: > ...although this was then re-checking the value of prev_state.state after modifying it

That part was intentional: in worker mode we call `set_state` on the worker, then bounce to the writer, which reads a `prev_state` that comes from the `set_state` call.

I hope all the back and forth of review isn't proving too bad. I think we're almost there for what it's worth!

dbkr: Hmm, I see what you mean. Unfortunately doing that caused a bunch of sytest failures: https://github.com/matrix-org/synapse/runs/5973187907?check_suite_focus=true - I thought the CI sytest ran with the synapse in worker mode, but I guess not if it's executing this code path? Does this suggest that the worker mode code would fail this sytest?

squahtx: That's unfortunate. CI has multiple sytest runs, including both single process and worker configurations. The worker mode configuration seems to be passing. I'll try to investigate what's going on tomorrow.

squahtx: Turns out `prev_state = prev_state.copy_and_replace(state=presence_state)` is insufficient, since it doesn't pick up the update to `last_active_ts` from the `set_state` call.

Instead, `prev_state = await self.current_state_for_user(user_id)` after `set_state` _ought_ to make sytest pass.
This looks awful, but ought to hit the in-memory cache, since we've just written to it a few lines before.

dbkr: Ah, makes sense - done

squahtx: That's done the trick!

synapse/handlers/presence.py
```diff
@@ -967,23 +988,38 @@ async def user_syncing(
...
+                # XXX: We set_state separately here and just update the last_active_ts above
+                # This keeps the logic as similar as possible between the worker and single
+                # process modes. Using set_state will actually cause last_active_ts to be
+                # updated always, which is not what the spec calls for, but synapse has done
+                # this for... forever, I think.
+                await self.set_state(
+                    UserID.from_string(user_id), {"presence": presence_state}, True
+                )
+                prev_state = await self.current_state_for_user(user_id)
+
```
squahtx, line 1004: We could do with a comment here along the lines of
```py
            # To keep the single process behaviour consistent with worker mode, run the
            # same logic as `update_external_syncs_row`, even thought it looks weird.
```

synapse/handlers/presence.py
```diff
@@ -967,23 +988,38 @@ async def user_syncing(
...
             if prev_state.state == PresenceState.OFFLINE:
-                # If they're currently offline then bring them online, otherwise
-                # just update the last sync times.
                 await self._update_states(
                     [
                         prev_state.copy_and_replace(
-                            state=PresenceState.ONLINE,
+                            state=presence_state,
```
squahtx, line 1009: I think we can only get here if `presence_state` is `OFFLINE`, which is never, given the current implementations of `SyncRestServlet.on_GET` and `EventStreamHandler.get_stream`.
I'm inclined to revert this line, not that it really matters:
```suggestion
                            state=PresenceState.ONLINE,
```

synapse/handlers/presence.py
```diff
@@ -967,23 +988,38 @@ async def user_syncing(
...
                             last_active_ts=self.clock.time_msec(),
                             last_user_sync_ts=self.clock.time_msec(),
                         )
                     ]
                 )
+            # otherwise, set the new presence state & update the last sync time,
+            # but don't update last_active_ts as this isn't an indication that
+            # they've been active.
```
squahtx, line 1017: Do we want to update this comment to note that `last_active_ts` may have been updated by `set_state()` above?

synapse/handlers/presence.py
```diff
@@ -967,23 +988,38 @@ async def user_syncing(
...
             else:
                 await self._update_states(
                     [
                         prev_state.copy_and_replace(
-                            last_user_sync_ts=self.clock.time_msec()
+                            last_user_sync_ts=self.clock.time_msec(),
```
squahtx, line 1022: small tidy up of the diff:
```suggestion
                            last_user_sync_ts=self.clock.time_msec()
```

squahtx (commented):
Did a final round of review of the overall diff. Just some picky tidy-ups left, otherwise I'm happy for us to merge

synapse/handlers/presence.py
```diff
@@ -967,9 +988,24 @@ async def user_syncing(
...
+                # updated always, which is not what the spec calls for, but synapse has done
+                # this for... forever, I think.
+                await self.set_state(
+                    UserID.from_string(user_id), {"presence": presence_state}, True
+                )
+                # To keep the single process behaviour consistent with worker mode, run the
+                # same logic as `update_external_syncs_row`, even though it looks weird
+                # (this should come from the in-memory cache)
+                prev_state = await self.current_state_for_user(user_id)
+
```
squahtx, line 1007: Sorry if I wasn't very clear just now. The weirdness (in my mind at least) is calling `set_state` and then replacing an `OFFLINE` state that we've just set with `ONLINE` (not that it can happen). It's crying out for a refactor which we can't do without mucking with the inter-worker API.

```suggestion
                # Retrieve the new state for the logic below. This should come from the
                # in-memory cache.
                prev_state = await self.current_state_for_user(user_id)

            # To keep the single process behaviour consistent with worker mode, run the
            # same logic as `update_external_syncs_row`, even though it looks weird.
```

dbkr: Ah, yes - I see what you mean.

PR #12216, opened 2022-03-14
Add some missing type hints to cache datastore.

clokep:
This does not fully type the cache datastore, but includes some type hints which were missing near some code I had touched recently!

### DIFF
--- a/synapse/storage/databases/main/cache.py
+++ b/synapse/storage/databases/main/cache.py
@@ -23,6 +23,7 @@ from synapse.replication.tcp.streams.events import (
     EventsStream,
     EventsStreamCurrentStateRow,
     EventsStreamEventRow,
+    EventsStreamRow,
 )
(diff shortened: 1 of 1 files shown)


synapse/storage/databases/main/cache.py
```diff
@@ -107,7 +111,9 @@ def get_all_updated_caches_txn(txn):
             "get_all_updated_caches", get_all_updated_caches_txn
         )
 
-    def process_replication_rows(self, stream_name, instance_name, token, rows):
+    def process_replication_rows(
+        self, stream_name: str, instance_name: str, token: int, rows: Iterable[Any]
+    ) -> None:
```
clokep, line 116: Note that the `Iterable[Any]` seems to match what we do for the other instances of `process_replication_rows`, it isn't ideal but I think we want them to all match?

reivilibre (approved):

PR #12219, opened 2022-03-14
Clean-up logic for rebasing URLs

clokep:
@reivilibre asked me to use `urljoin` instead of our own rebasing logic in https://github.com/matrix-org/synapse/pull/11767#discussion_r789506708, from that I further cleaned up some of the logic around rebasing URLs.

This should be reviewable commit-by-commit, but the gist is:

1. Use [`urljoin`](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urljoin) instead of re-implementing the logic.
2. Realization that we do not need to call `rebase_url` in multiple places, so remove a call to it.
3. Inline the rebase URL logic.

I think overall this is a nice clean-up, but let me know if you have questions!

### DIFF
--- a/synapse/rest/media/v1/preview_html.py
+++ b/synapse/rest/media/v1/preview_html.py
@@ -16,7 +16,7 @@ import itertools
 import logging
 import re
 from typing import TYPE_CHECKING, Dict, Generator, Iterable, Optional, Set, Union
-from urllib.parse import urlparse, urlunparse
+from urllib.parse import urlparse, urljoin
 
 if TYPE_CHECKING:
     from lxml import etree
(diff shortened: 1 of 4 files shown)


synapse/rest/media/v1/preview_html.py
```diff
@@ -144,9 +143,7 @@ def decode_body(
     return etree.fromstring(body, parser)
 
 
-def parse_html_to_open_graph(
-    tree: "etree.Element", media_uri: str
-) -> Dict[str, Optional[str]]:
+def parse_html_to_open_graph(tree: "etree.Element") -> Dict[str, Optional[str]]:
     """
     Parse the HTML document into an Open Graph response.
```
H-Shay, line 149: Just wondering if you want to remove the `media_url` parameter from the docstring below since it is being removed from the function signature?

clokep: Good catch! Thanks! 👍

PR #12196, opened 2022-03-09
Document stream writer `typing`, `to_device`, `account_data`, `receipts`, and `presence` can only be used on a single worker.

cremesk:

PR #12220, opened 2022-03-14
docs/worker: add missing typing request location

cremesk:
@clokep we both missed and forgot something in the PR #12196. ;)

### DIFF
--- a/changelog.d/12220.doc
+++ b/changelog.d/12220.doc
@@ -1,0 +1,1 @@
+add missing typing requests location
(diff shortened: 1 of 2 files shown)


clokep:
> @clokep oh okay. I was so unaware of that. now I'm a little irritated myself.

#11599 attempted to make bits of this clearer, but it is definitely confusing! 😢 I think being able to allow the generic worker to properly forward is the desired end-goal. For now I think I'm going to close this PR unless you can figure out a way to make this clearer?

PR #12225, opened 2022-03-15
Use the ignored_users table when testing event visibility.

### LINKED ISSUE #11964: Consider using the "ignored_users" database table when filtering client events
When ignoring a user, clients modify a `m.ignored_user_list` key in their account data ([spec](https://spec.matrix.org/v1.2/client-server-api/#ignoring-users)). Synapse will detect if this is happening, and modify an equivalent database table named `ignored_users`. The idea is that the database table is faster to read from than digging the value out of account data.

The table is used when calculating bulk push rules:

https://github.com/matrix-org/synapse/blob/10a88ba91cb16ccf757984f0a7d41ddf8b4dc07f/synapse/push/bulk_push_rule_evaluator.py#L212-L216
https://github.com/matrix-org/synapse/blob/af795173bec447da44b65b739d05d0083b02229d/synapse/storage/databases/main/account_data.py#L367-L385

Yet it is not used when ignored users are considered while filtering events for clients (i.e. for `/sync`):

https://github.com/matrix-org/synapse/blob/64ec45fc1b0856dc7daacca7d3ab75d50bd89f84/synapse/handlers/sync.py#L1613-L1634
https://github.com/matrix-org/synapse/blob/4c2096599c9780290703e14df63963e77d058dda/synapse/visibility.py#L89-L91

I'm not sure how much of an optimisation this would really be (as `get_global_account_data_by_type_for_user` is cached on key (`user_id`, `account_data_type`)). In addition, we'd need an index of the `ignorer_user` column, which doesn't currently exist.

On the flip side - it is a nice idea to consolidate where we pull this information from.

clokep:
Instead of parsing the account data in multiple places, this uses the `ignored_users` table to fetch all users that are ignored by a user (previously we were only using it to calculate all users which ignore a specific user).

There was concern in #11964 that this would need another index, but it does not the uniqueness that's on the table can be used:

```sql
=> EXPLAIN SELECT ignored_user_id FROM ignored_users WHERE ignorer_user_id = '@clokep:matrix.org';
                                             QUERY PLAN                                             
----------------------------------------------------------------------------------------------------
 Index Only Scan using ignored_users_uniqueness on ignored_users  (cost=0.41..4.45 rows=2 width=25)
   Index Cond: (ignorer_user_id = '@clokep:matrix.org'::text)
(2 rows)
```

Fixes #11964 

Reviewable commit-by-commit.

### DIFF
--- a/changelog.d/12225.misc
+++ b/changelog.d/12225.misc
@@ -1,0 +1,1 @@
+Use the `ignored_users` table in additional places instead of re-parsing the account data.
--- a/synapse/handlers/sync.py
+++ b/synapse/handlers/sync.py
@@ -28,7 +28,7 @@ from typing import (
 import attr
 from prometheus_client import Counter
 
-from synapse.api.constants import AccountDataTypes, EventTypes, Membership, ReceiptTypes
+from synapse.api.constants import EventTypes, Membership, ReceiptTypes
 from synapse.api.filtering import FilterCollection
 from synapse.api.presence import UserPresenceState
 from synapse.api.room_versions import KNOWN_ROOM_VERSIONS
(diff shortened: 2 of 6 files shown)


DMRobertson (approved):
LGTM!

PR #12227, opened 2022-03-15
Refactor relations code to avoid duplication

clokep:
This is some pre-work towards #11753, it injects a handler layer between the rest and datastore layers, as much of our other code has.

A partial goal of this is to make the relations code a bit closer to the pagination code, so this mirrors more of the `RoomMessageListRestServlet` -> `PaginationHandler.get_messages()` -> `StreamWorkerStore.paginate_room_events()` flow.

### DIFF
--- a/changelog.d/12227.misc
+++ b/changelog.d/12227.misc
@@ -1,0 +1,1 @@
+Refactor the relations endpoints to add a `RelationsHandler`.
--- a/synapse/handlers/pagination.py
+++ b/synapse/handlers/pagination.py
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import logging
-from typing import TYPE_CHECKING, Any, Collection, Dict, List, Optional, Set
+from typing import TYPE_CHECKING, Collection, Dict, List, Optional, Set
 
 import attr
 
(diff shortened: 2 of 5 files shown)


clokep:
There's probably a bit more we could refactor here, but I think this is enough for what I need for #11753.

clokep:
> There's probably a bit more we could refactor here, but I think this is enough for what I need for #11753.

I do kind of feel like the `get_bundled_aggregations` (and maybe `_get_bundled_aggregation_for_event`) could be in here though? WDYT?

reivilibre (approved):

reivilibre:
> I do kind of feel like the get_bundled_aggregations (and maybe _get_bundled_aggregation_for_event) could be in here though? WDYT?

I would think so, probably. If it looks hairier after you've done it then you'll know it was a bad move, but sounds about right to me :)

clokep:
> > I do kind of feel like the get_bundled_aggregations (and maybe _get_bundled_aggregation_for_event) could be in here though? WDYT?
> 
> I would think so, probably. If it looks hairier after you've done it then you'll know it was a bad move, but sounds about right to me :)

Good shout! I think I'll do that in a follow-up since the changes in here are blocking other work. 👍

PR #12228, opened 2022-03-15
Only fetch participated for events with threads

clokep:
This makes the code match the comment: we should only attempt to fetch whether the current user has participated in a thread *if there's actually a thread at all on that event*.

As part of this I implemented a way for unit tests to assert the number of database transactions occurring, which I checked is decreased by this PR. 🎉 

~~Depends on #12232.~~

Reviewable commit-by-commit (much of the code-change is refactoring some test code).

### DIFF
--- a/tests/rest/client/test_relations.py
+++ b/tests/rest/client/test_relations.py
@@ -15,7 +15,7 @@
 
 import itertools
 import urllib.parse
-from typing import Any, Dict, List, Optional, Tuple
+from typing import Any, Callable, Dict, List, Optional, Tuple
 from unittest.mock import patch
 
 from twisted.test.proto_helpers import MemoryReactor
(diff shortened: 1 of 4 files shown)


tests/rest/client/test_relations.py
```diff
@@ -1291,6 +1067,264 @@ def test_aggregation_pagination_within_group(self) -> None:
...
+            "GET",
+            f"/rooms/{self.room}/context/{self.parent_id}",
+            access_token=self.user_token,
+        )
+        self.assertEqual(200, channel.code, channel.json_body)
+        assert_bundle(channel.json_body["event"])
+
+        # Request sync.
+        # bundled aggregations).
+        filter = urllib.parse.quote_plus(b'{"room": {"timeline": {"limit": 4}}}')
```
squahtx, line 1135: Is the middle line a typo?

clokep: I think it is from copying and pasting? I'll remove it!

tests/rest/client/test_relations.py
```diff
@@ -1291,6 +1067,264 @@ def test_aggregation_pagination_within_group(self) -> None:
...
+        channel = self.make_request(
+            "GET",
+            f"/rooms/{self.room}/context/{self.parent_id}",
+            access_token=self.user_token,
+        )
+        self.assertEqual(200, channel.code, channel.json_body)
+        assert_bundle(channel.json_body["event"])
+
+        # Request sync.
+        # bundled aggregations).
```
clokep, line 1134:
```suggestion
```

PR #12234, opened 2022-03-16
Fix a bug if unsigned is not a returned event field.

### LINKED ISSUE #12233: KeyError: 'unsigned' in _inject_bundled_aggregations
https://sentry.matrix.org/sentry/synapse-matrixorg/issues/243572/

This seems to happen if a filter is provided which specifies specific fields to return, investigating more now though...

clokep:
Fixes #12233

There was a user-supplied filter to only return a few fields (not including `unsigned`) which causes the bundled aggregations code to break.

According to the spec for `event_fields` (emphasis mine):

>  List of event fields to include. If this list is absent then all fields are included. The entries may include ‘.’ characters to indicate sub-fields. So [‘content.body’] will include the ‘body’ field of the ‘content’ object. A literal ‘.’ character in a field name may be escaped using a ‘\’. **A server may include more fields than were requested.**

I did debate if we should try to do some magic to not inject bundled aggregations if `unsigned` is not a returned field, but this is a bit finicky since someone could filter out just `unsigned.age`, etc. Additionally, edits actually modify the content, so we want to always apply those (unless `content` is not returned). Anyway, you see how this gets finicky.


tests/rest/client/test_relations.py
```diff
@@ -981,6 +981,28 @@ def test_background_update(self) -> None:
             [annotation_event_id_good, thread_event_id],
         )
 
+    def test_bundled_aggregations_with_filter(self):
+        """If unsigned is not a returned field, the bundled aggregations still get added."""
```
reivilibre, line 985: I would remark in this test description that this is a design choice for ease and not something that the spec requires (indeed, maybe one day it will make sense to save the bytes and leave it off).

reivilibre (approved):
Thanks.

Ought this go to the release branch though?

clokep:
FTR we decided to put this into 1.55.0 since we haven't cut an RC for it yet.

PR #12235, opened 2022-03-16
Do not consider events by ignored users for bundled aggregations

clokep:
This is the last bit needed for #11753; we need to ensure that if a user ignores another user, their related events are properly ignored in bundled aggregations.

See #12285 for filtering events from ignored users in `/relations`.

Note that for the cached methods we per-calculate the list of ignored users and pass it into each other method. The reasoning for this is that users who have the same set of ignored users will then share the cache (as opposed to if we passed in the requester) -- my thought being that most users will not have anyone ignored *or* that it is highly likely for a small set of users to be ignored. (I.e. essentially that there's a small set of people who get ignored for being spammy.)


synapse/handlers/relations.py
```diff
@@ -223,13 +238,21 @@ async def get_bundled_aggregations(
...
+            event_result = await self._get_bundled_aggregation_for_event(
+                event, ignored_users
+            )
             if event_result:
                 results[event.event_id] = event_result
 
         # Fetch any edits (but not for redacted events).
+        #
+        # Note that there is no use in limiting edits by ignored users since the
+        # parent event should be ignored in the first place if the user is ignored.
```
clokep, line 255: Does this sound like sound logic? I couldn't come up with a situation where we would need to check edits,

synapse/handlers/relations.py
```diff
@@ -113,6 +116,9 @@ async def get_relations(
         if event is None:
             raise SynapseError(404, "Unknown parent event.")
 
+        # Note that ignored users are not passed into get_relations_for_event
+        # below. Ignored users are handled in filter_events_for_client (and by
+        # noy passing them in here we should get a better cache hit rate).
```
clokep, line 121: Does this sound reasonable or do we think we should do the initial filtering by ignoring users too? (I think we would essentially do it twice then -- once in `get_relations_for_event`, and then again in `filter_events_for_client`.

erikjohnston:
> Note that for the cached methods we per-calculate the list of ignored users and pass it into each other method. The reasoning for this is that users who have the same set of ignored users will then share the cache (as opposed to if we passed in the requester) -- my thought being that most users will not have anyone ignored _or_ that it is highly likely for a small set of users to be ignored. (I.e. essentially that there's a small set of people who get ignored for being spammy.)

My hunch is that we'd pretty much only hit the cache for users with zero ignored users, as I think its going to be relatively rare for different users to have the *exact* same set of ignored users (even if they are mostly the same). It also means that we cache the set of ignored users in the cache key, which feels weird and very bleurgh for people with large numbers of ignored users (which would be rare tbf)

For `get_relations_for_event` can we instead lift the filtering of the ignored users out of the DB query? (You might want to do it in a loop to fetch more results to replace any filtered out).

I'm not really sure what to do about calculating the counts in the other queries. Some thoughts I have are:

1. Filter ignored users to those that are also in the room, to hopefully increase cache hits
2. Don't bother caching the query if `ignored_users` is non-empty
3. Have a separate cached query that looks up the counts by user, e.g. for `get_aggregation_groups_for_event` could have a query to look up `(event_id, sender) -> (type, aggregation_key)` so that you could call that for each ignored user, and use the results to fiddle with the counts. :shrug: 

Thoughts?

clokep:
> For `get_relations_for_event` can we instead lift the filtering of the ignored users out of the DB query? (You might want to do it in a loop to fetch more results to replace any filtered out).

I originally thought about having it also use the `filter_events_for_client`, but that would require resolving the event IDs to events, which is not cheap. I suppose we could also request the sender and do the filtering in Python though. This might need a bit more of refactoring...as right now the results of `get_relations_for_event` are essentially splatted out into responses (and bundled aggregations).

> * Filter ignored users to those that are also in the room, to hopefully increase cache hits

That could maybe work?

> * Don't bother caching the query if `ignored_users` is non-empty

This sounds like the "easy" thing to do, but I'd be curious what percentage of users ignore at least someone... otherwise we might just be making the cache useless.

> * Have a separate cached query that looks up the counts by user, e.g. for `get_aggregation_groups_for_event` could have a query to look up `(event_id, sender) -> (type, aggregation_key)` so that you could call that for each ignored user, and use the results to fiddle with the counts. 🤷

Something like that might work, I'll investigate. I think it is a bit more fiddly, but probably the best plan?

Thanks for the thoughts! I'll poke a bit more at this.

erikjohnston:
> This sounds like the "easy" thing to do, but I'd be curious what percentage of users ignore at least someone... otherwise we might just be making the cache useless.

Yeah, no idea, though if you also filter the ignored users to those in the room it might not be too bad? :shrug:

erikjohnston (commented):
(Bah, forgot to hit the submit review button)

synapse/storage/databases/main/relations.py
```diff
@@ -287,14 +296,71 @@ async def get_aggregation_groups_for_event(
...
-            txn.execute(sql, where_args)
+            txn.execute(sql, args)
 
             return [{"type": row[0], "key": row[1], "count": row[2]} for row in txn]
 
         return await self.db_pool.runInteraction(
             "get_aggregation_groups_for_event", _get_aggregation_groups_for_event_txn
         )
 
+    async def get_aggregation_groups_for_users(
```
clokep, line 307: Not sure if this needs caching?

synapse/storage/databases/main/relations.py
```diff
@@ -521,6 +587,67 @@ def _get_thread_summaries_txn(
 
         return summaries
 
+    async def get_threaded_messages_per_user(
```
clokep, line 590: Not sure if this needs caching?

clokep:
@erikjohnston Could you take another look at this? I attempted to take the idea of having the "normal" query, which is cached, and then substracting out info from ignored users, as you proposed. Let me know if you want to discuss!

synapse/storage/database.py
```diff
@@ -2225,9 +2230,18 @@ def make_in_list_sql_clause(
...
+            sql = f"{column} != ANY(?)"
+        return sql, [list(iterable)]
     else:
-        return "%s IN (%s)" % (column, ",".join("?" for _ in iterable)), list(iterable)
+        values = ",".join("?" for _ in iterable)
+        if include:
+            sql = f"{column} IN ({values})"
+        else:
+            sql = f"{column} NOT IN ({values})"
+        return sql, list(iterable)
```
erikjohnston, line 2244: I think these changes are now no longer used?

synapse/storage/databases/main/relations.py
```diff
@@ -46,6 +49,12 @@
 logger = logging.getLogger(__name__)
 
 
+@attr.s(slots=True, frozen=True, auto_attribs=True)
+class _RelatedEvent:
+    event_id: str
+    sender: str
```
erikjohnston, line 55: Docstring please :)

synapse/handlers/relations.py
```diff
@@ -216,6 +314,96 @@ async def _get_bundled_aggregation_for_event(
...
+                        continue
+
+                    # The *last* event returned is the one that is cared about.
+                    #
+                    # This event shuold exist.
+                    event = await self._event_handler.get_event(
+                        user, room_id, potential_events[-1].event_id
+                    )
+                    if event is None:
+                        continue
```
erikjohnston, line 393: Do we want to warn here if we don't expect that this should ever get hit?

synapse/storage/databases/main/relations.py
erikjohnston: I don't think these need caching, given that the common case will be a no-op anyway. We can always revisit it if we we see it taking time

synapse/storage/databases/main/relations.py
```diff
@@ -287,14 +296,71 @@ async def get_aggregation_groups_for_event(
...
+        args.extend(users_args)
+
+        sql = f"""
+            SELECT type, aggregation_key, COUNT(DISTINCT sender)
+            FROM event_relations
+            INNER JOIN events USING (event_id)
+            WHERE relates_to_id = ? AND room_id = ? AND relation_type = ? AND {users_sql}
+            GROUP BY relation_type, type, aggregation_key
+            ORDER BY COUNT(*) DESC
+            LIMIT ?
```
erikjohnston, line 350: Have you checked the query plans for these? I'm struggling to reason about if these are "safe" or not?

erikjohnston (commented):
This makes sense to me, modulo checking that the queries are good

synapse/storage/databases/main/relations.py
clokep: Good call! Will do.

clokep:
@erikjohnston I made the changes requested and we discussed the query plans, I think this is probably good to go. I've just merged in `develop` again to double check CI!

erikjohnston (approved):
:shipit:

PR #12237, opened 2022-03-16
Move get_bundled_aggregations to relations handler.

clokep:
Follow-up to #12227, which moves some more code to the handler. `get_bundled_aggregations` is pretty high-level and doesn't make sense to be in the datastore IMO. @reivilibre seemed to agree in the review of #12227.

This is a lift-and-shift of `get_bundled_aggregations`, `_get_bundled_aggregations_for_event`; and then a rename of some private methods.

### DIFF
--- a/changelog.d/12237.misc
+++ b/changelog.d/12237.misc
@@ -1,0 +1,1 @@
+Refactor the relations endpoints to add a `RelationsHandler`.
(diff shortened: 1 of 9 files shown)


squahtx (approved):
lgtm

CI seems stuck. I'd give it a kick except I can't see how without committing to your branch.

clokep:
Merging in develop seems to have kicked it. Probably while GH was having issues?

PR #12243, opened 2022-03-17
Sync more worker regexes in the documentation.

### LINKED ISSUE #12229: Worker docs has 2 different endpoint regex for same endpoint
At https://github.com/matrix-org/synapse/blob/develop/docs/workers.md (master is same in this regard at the moment) there exist 2 different regex endpoints for some endpoints. Mostly those that is also mentioned in stream writers has more regex matches than the otherwise same endpoint given earlier in generic worker list..

For example under generic worker listing:
```
# Device requests
^/_matrix/client/(r0|v3|unstable)/sendToDevice/
```
and then later in stream writer `to_device` listing:
```
^/_matrix/client/(api/v1|r0|v3|unstable)/sendToDevice/
```

clokep:
Fixes #12229.

These were missed in [`44853ea` (#11599)](https://github.com/matrix-org/synapse/pull/11599/commits/44853ea070646a3bdac4cf567fd43997afea6932).

### DIFF
--- a/changelog.d/12243.doc
+++ b/changelog.d/12243.doc
@@ -1,1 +1,1 @@
-Remove incorrect `api/v1` prefixes in the worker documentation for some endpoints.
+Remove incorrect prefixes in the worker documentation for some endpoints.
--- a/changelog.d/12243.doc
+++ b/changelog.d/12243.doc
@@ -1,0 +1,1 @@
+Remove incorrect `api/v1` prefixes in the worker documentation for some endpoints.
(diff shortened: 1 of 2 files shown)


olmari (approved):
LGTM in documentation side of things, I assume they are now proper current endpoints too :)

clokep:
I had another branch with some related changes, figured I'd dump them here.

squahtx (approved):
lgtm

`docker/configure_workers_and_start.py` has copies of some of these regexes that are now out of sync. Do we want to update them at some point?

PR #12246, opened 2022-03-17
Correct `check_username_for_spam` annotations and docs

### LINKED ISSUE #12245: `check_username_for_spam`'s documentation incorrectly claims that `avatar_url` and `display_name` are never `None`
### Description

The Synapse documentation claims that the `check_username_for_spam` spam checker callback receives a `userprofile` that is a `Dict[str, str]`. In practice, there can be `None`s in the `userprofile` dict, eg.
```py
{
    avatar_url: None, 
    display_name: None, 
    user_id: '@[redacted]:matrix.org'
}
```

See https://sentry.matrix.org/sentry/synapse-matrixorg/issues/243697/ for examples of `userprofile`s with missing avatars and display names.

`docs/modules/spam_checker_callbacks.md` and `CHECK_USERNAME_FOR_SPAM_CALLBACK` could do with an update.

DMRobertson:
Fixes #12245.

I'm not completely sure how I feel about using a `TypedDict` here, but

- using a namedtuple or attrs struct would mean the diff had to update a bunch of tests
- we don't do very much with the user profile data; deserialising to a struct only to reseralise to a dict seemed like overkill
- more compellingly, the spam checker callback already expects a dictionary, so let's keep giving them a dictionary

Not convinced the documentation change is super easy to read. All thoughts are welcome.

### DIFF
--- a/docs/modules/spam_checker_callbacks.md
+++ b/docs/modules/spam_checker_callbacks.md
@@ -172,7 +172,7 @@ any of the subsequent implementations of this callback.
 _First introduced in Synapse v1.37.0_
 
 ```python
-async def check_username_for_spam(user_profile: synapse.types.UserProfile) -> bool
+async def check_username_for_spam(user_profile: synapse.module_api.UserProfile) -> bool
 ```
 
 Called when computing search results in the user directory. The module must return a
--- a/docs/modules/spam_checker_callbacks.md
+++ b/docs/modules/spam_checker_callbacks.md
@@ -172,7 +172,7 @@ any of the subsequent implementations of this callback.
 _First introduced in Synapse v1.37.0_
 
 ```python
-async def check_username_for_spam(user_profile: Dict[str, str]) -> bool
+async def check_username_for_spam(user_profile: synapse.types.UserProfile) -> bool
 ```
 
 Called when computing search results in the user directory. The module must return a
@@ -180,11 +180,13 @@ Called when computing search results in the user directory. The module must retu
 searches. Return `True` to indicate that the user is spammy and exclude them from 
 search results; otherwise return `False`.
 
-The profile is represented as a dictionary with the following keys:
+The profile is represented as a dictionary with the following keys.
 
-* `user_id`: The Matrix ID for this user.
-* `display_name`: The user's display name.
-* `avatar_url`: The `mxc://` URL to the user's avatar.
+* `user_id: str`. The Matrix ID for this user.
+* `display_name: Optional[str]`. The user's display name, or `None` if this user
+  has not set a display name.
+* `avatar_url: Optional[str]`. The `mxc://` URL to the user's avatar, or `None`
+  if this user has not set an avatar.
 
 The module is given a copy of the original dictionary, so modifying it from within the
 module cannot modify a user's profile when included in user directory search results.
--- a/docs/modules/spam_checker_callbacks.md
+++ b/docs/modules/spam_checker_callbacks.md
@@ -180,7 +180,7 @@ Called when computing search results in the user directory. The module must retu
 searches. Return `True` to indicate that the user is spammy and exclude them from 
 search results; otherwise return `False`.
 
-The profile is represented as a dictionary with the following keys.
+The profile is represented as a dictionary with the following keys:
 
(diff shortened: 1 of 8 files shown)


synapse/types.py
```diff
@@ -63,6 +64,10 @@
 # JSON types. These could be made stronger, but will do for now.
 # A JSON-serialisable dict.
 JsonDict = Dict[str, Any]
+# A JSON-serialisable mapping; roughly speaking an immutable JSONDict.
+# Useful when you have a TypedDict which isn't going to be mutated and you don't want
+# to cast to JsonDict everywhere.
+JsonMapping = Mapping[str, Any]
```
squahtx, line 70: It's worth noting that `Mapping`s aren't necessarily JSON-serialisable, unless they're really `dict`s or `frozendict`s. Essentially when we use `JsonMapping`, we're pinky promising that it's really a `(frozen)dict`.

The same kind of thing applies to `JsonSerializable` below I guess.

squahtx (approved):
Looks good!
Left some questions

docs/modules/spam_checker_callbacks.md
```diff
@@ -172,19 +172,21 @@ any of the subsequent implementations of this callback.
 _First introduced in Synapse v1.37.0_
 
 ```python
-async def check_username_for_spam(user_profile: Dict[str, str]) -> bool
+async def check_username_for_spam(user_profile: synapse.types.UserProfile) -> bool
```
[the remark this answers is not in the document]
DMRobertson: No idea. Perhaps it makes more sense to rexport the type in `synapse.module_api`.

synapse/types.py
DMRobertson: I wasn't 100% sold on this either. I wanted to see if there was a way to make more use of `TypedDicts` without having to cast those back to `JsonDicts` everywhere. Maybe I could just make the return type of this function `SearchResult`? That feels a bit off though---it's exposing a type from storage.databases.main in rest.client.

https://github.com/matrix-org/synapse/blob/7fe7b42e4b3f7c9b0172bc79dd84deafc94a23b7/synapse/rest/client/user_directory.py#L41

DMRobertson:
Looks like I need to import TypedDict from typing_extensions on older Pythons.

docs/modules/spam_checker_callbacks.md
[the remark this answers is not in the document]
squahtx: That sounds like a good idea!

synapse/types.py
squahtx: I do think there is a case for having a `JsonMapping` type, to signal that functions aren't allowed to modify the contents of a dict. Especially around caches.

I'm not _opposed_ to making the return type some sort of `TypedDict`. I think we really ought to have a file somewhere full of `TypedDict` definitions that come from the spec, since that's the source of truth for the return type here (and not the database layer).

In any case I'm happy with the `JsonMapping` as is.

DMRobertson: > I do think there is a case for having a `JsonMapping` type, to signal that functions aren't allowed to modify the contents of a dict. Especially around caches.

I broadly sympathise, but I don't think that'd be fool proof: e.g. given `d: JsonMapping = {"a": [1,2,3]}` I think `d["a"].append(4)` would still type check?

> I think we really ought to have a file somewhere full of TypedDict definitions that come from the spec, since that's the source of truth for the return type here (and not the database layer).

I like the sound of this; e.g. `synapse.types.spec.UserDirectorySearchResult`.  I guess the dream would be to automatically generate such a file from the openapi definition in the spec.

squahtx: > > I do think there is a case for having a `JsonMapping` type, to signal that functions aren't allowed to modify the contents of a dict. Especially around caches.
> 
> I broadly sympathise, but I don't think that'd be fool proof: e.g. given `d: JsonMapping = {"a": [1,2,3]}` I think `d["a"].append(4)` would still type check?

True! Unless we manage to define `JsonMapping = Mapping[str, ImmutableJsonSerializable]` recursively, which mypy won't let us do. (grumble grumble this would be possible in typescript)

squahtx (approved):

PR #12251, opened 2022-03-18
Move `update_client_ip` background job from the main process to the background worker.

### LINKED ISSUE #10425: Move `update_client_ips` to the background worker
https://github.com/matrix-org/synapse/blob/c0df6bae066fe818bb80d41af65503be7a07275d/synapse/storage/databases/main/client_ips.py#L474-L486

I have a feeling that we wanted to keep this on the main process as it needs to only be run by one, single process. However, this is no longer a concern now that we can dedicate all background work to a single worker through the `run_background_tasks_on: <worker_name>` config option.

This operation has been reported as potentially CPU intensive on some deployments, so moving it off the main process would be a win.

(Also random thought: can this be converted to something that operates entirely in the database (postgres)?)

reivilibre:
This is intended to reduce load on the main process.

Fixes #10425.

The strategy is as follows:

- Get rid of `MonthlyActiveUsersStore` and inline it into `MonthlyActiveUsersWorkerStore`, but enforce a single writer (the background process)
- Get rid of `ClientIpStore` and inline it into `ClientIpWorkerStore`, also enforcing a single writer.
- Get rid of `SlavedClientIpStore` and inline it into `ClientIpWorkerStore` since it's almost entirely duplicated code anyway.

This should be a commit-by-commit reviewable PR.

Requires SyTest changes: https://github.com/matrix-org/sytest/pull/1222


synapse/replication/tcp/handler.py
```diff
@@ -235,6 +235,8 @@ def __init__(self, hs: "HomeServer"):
         if self._is_master:
             self._server_notices_sender = hs.get_server_notices_sender()
 
+        self._is_background_worker = hs.config.worker.run_background_tasks
```
reivilibre, line 238: Needs updating to only include Redis

synapse/replication/tcp/handler.py
```diff
@@ -401,12 +403,27 @@ def on_USER_IP(
...
     async def _handle_user_ip(self, cmd: UserIpCommand) -> None:
+        """
+        Handles a User IP, branching depending on whether we are the main process
+        and/or the background worker.
+        """
+        if self._is_master:
+            await self._handle_user_ip_as_master(cmd)
+
+        if self._is_background_worker:
+            await self._handle_user_ip_as_background_worker(cmd)
```
erikjohnston, line 426: I think it would be clearer to inline `_handle_user_ip` into `on_USER_IP`? Otherwise its a bit confusing

erikjohnston (approved):
Otherwise LGTM!

synapse/replication/tcp/handler.py
reivilibre: Only concern with that is that it will mean that uninvolved workers end up creating and scheduling coroutines as background processes as well, whereas they didn't before.
I don't know how much that worries us, but since the code was already structured this way for some reason, I didn't want to remove it without understanding.

erikjohnston: I'm confused, I'm not talking about a functional change here? `_handle_user_ip` is only called by `on_USER_IP`?

What's confusing me really is that the `self._is_master or self._should_insert_client_ips` a) looks odd by itself, and b) is entirely redundant given `_handle_user_ip` then does the exact same checks. You may as well just always call `self._handle_user_ip` in `on_USER_IP`, at which point having it as a separate function seems a bit silly?

Or am I missing something here?

reivilibre: `on_USER_IP` returns either `None` or an awaitable. If it's an awaitable, it goes through extra stuff to schedule it as a background process; if it's just `None` then nothing happens.

Notably, the function can't (easily/nicely) be inlined and still preserve this behaviour of returning `None` if there's nothing to do, as I'd have to make `on_USER_IP` an `async def` in order to run potentially two coroutines back-to-back.

It may not be a big deal, but I wasn't sure and since replication overhead is pretty sensitive, I was a bit uncomfortable taking a leap of faith.

erikjohnston: OH. Right, with you now. How obscure.

reivilibre: Yes, and it took me forever where to find the code that calls `on_%s` because the name is dynamically constructed and I forgot that we like `%s`-formatting and was searching for `"on_"` :')

erikjohnston: Can you add a comment in `on_USER_IP` about the whole awaitable thing please?

reivilibre: Done — I also inlined the two specific `handle_user_ip` functions, since looking at them now, they're so short it wasn't buying any clarity.

PR #12252, opened 2022-03-18
Optimise `_update_client_ips_batch_txn` to batch together database operations.

reivilibre:
This easy optimisation was noticed in #10425 by anoa, and it seemed like the right time to implement these
(along with #12251). Not expecting super magical powers, but it might be a nice little improvement.

Depends on: #12251

### DIFF
--- a/synapse/storage/database.py
+++ b/synapse/storage/database.py
@@ -1856,11 +1856,6 @@ class DatabasePool:
                 f"{len(key_values)} key rows and {len(value_values)} value rows: should be the same number."
             )
 
-        # List of value names, then key names
-        allnames: List[str] = []
-        allnames.extend(value_names)
-        allnames.extend(key_names)
-
         # List of tuples of (value values, then key values)
         # (This matches the order of `allnames` and the order of the query)
         args = [tuple(x) + tuple(y) for x, y in zip(value_values, key_values)]
--- a/synapse/storage/database.py
+++ b/synapse/storage/database.py
@@ -63,7 +63,7 @@ if TYPE_CHECKING:
     from synapse.server import HomeServer
 
 # python 3 does not have a maximum int value
-MAX_TXN_ID = 2 ** 63 - 1
+MAX_TXN_ID = 2**63 - 1
 
 logger = logging.getLogger(__name__)
 
--- a/synapse/storage/database.py
+++ b/synapse/storage/database.py
@@ -1280,7 +1280,8 @@ class DatabasePool:
             value_names: The value column names
             value_values: A list of each row's value column values.
                 Ignored if value_names is empty.
-            lock: True to lock the table when doing the upsert.
+            lock: True to lock the table when doing the upsert. Unused if the database engine
+            	supports native upserts.
         """
 
         # We can autocommit if we are going to use native upserts
@@ -1320,7 +1321,8 @@ class DatabasePool:
             value_names: The value column names
             value_values: A list of each row's value column values.
                 Ignored if value_names is empty.
-            lock: True to lock the table when doing the upsert.
+            lock: True to lock the table when doing the upsert. Unused if the database engine
+            	supports native upserts.
         """
         if self.engine.can_native_upsert and table not in self._unsafe_to_upsert_tables:
             return self.simple_upsert_many_txn_native_upsert(
--- a/synapse/storage/database.py
+++ b/synapse/storage/database.py
@@ -1360,11 +1360,17 @@ class DatabasePool:
         if not value_names:
             value_values = [() for x in range(len(key_values))]
 
+        if lock:
+            # Lock the table just once, to prevent it being done once per row.
+            # Note that, according to Postgres' documentation, once obtained,
+            # the lock is held for the remainder of the current transaction.
+            self.engine.lock_table(txn, "user_ips")
+
         for keyv, valv in zip(key_values, value_values):
             _keys = {x: y for x, y in zip(key_names, keyv)}
             _vals = {x: y for x, y in zip(value_names, valv)}
 
-            self.simple_upsert_txn_emulated(txn, table, _keys, _vals, lock=lock)
+            self.simple_upsert_txn_emulated(txn, table, _keys, _vals, lock=False)
 
     def simple_upsert_many_txn_native_upsert(
         self,
--- a/synapse/storage/database.py
+++ b/synapse/storage/database.py
@@ -63,7 +63,7 @@ if TYPE_CHECKING:
     from synapse.server import HomeServer
 
 # python 3 does not have a maximum int value
-MAX_TXN_ID = 2**63 - 1
+MAX_TXN_ID = 2 ** 63 - 1
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


clokep (commented):
Looks pretty good!

synapse/storage/database.py
```diff
@@ -1792,6 +1799,86 @@ def simple_update_txn(
...
+            value_values: A list of each row's value column values.
+        """
+
+        if len(value_values) != len(key_values):
+            raise ValueError(
+                f"{len(key_values)} key rows and {len(value_values)} value rows: should be the same number."
+            )
+
+        # List of tuples of (value values, then key values)
+        # (This matches the order of `allnames` and the order of the query)
```
clokep, line 1860: This still refers to `allnames`.

```suggestion
        # (This matches the order needed for the query)
```

synapse/storage/database.py
```diff
@@ -1355,7 +1362,7 @@ def simple_upsert_many_txn_emulated(
             _keys = {x: y for x, y in zip(key_names, keyv)}
             _vals = {x: y for x, y in zip(value_names, valv)}
 
-            self.simple_upsert_txn_emulated(txn, table, _keys, _vals)
+            self.simple_upsert_txn_emulated(txn, table, _keys, _vals, lock=lock)
```
clokep, line 1365: I'm guessing this works fine since we lock the table (as part of the transaction) in the first iterate of the loop, but then it stays locked until the end of the transaction?

1. Is this accurate?
2. If so, is there overhead to locking it for every value? (Should we only lock the table on the first iteration of the loop?)

synapse/storage/database.py
```diff
@@ -1268,6 +1268,7 @@ async def simple_upsert_many(
         value_names: Collection[str],
         value_values: Collection[Collection[Any]],
         desc: str,
+        lock: bool = True,
```
clokep, line 1271: Note that I *think* this is fine to default to `True` since `simple_upsert_txn_emulated` (which this eventually calls) defaults to `True`. So pretty much this was the default behavior.

synapse/storage/database.py
```diff
@@ -1279,21 +1280,23 @@ async def simple_upsert_many(
             value_names: The value column names
             value_values: A list of each row's value column values.
                 Ignored if value_names is empty.
+            lock: True to lock the table when doing the upsert.
```
clokep, line 1283: This should likely say it only applies for emulated?

```suggestion
            lock: True to lock the table when doing the upsert. Unused if the database engine
            	supports native upserts.
```

synapse/storage/database.py
```diff
@@ -1316,14 +1320,15 @@ def simple_upsert_many_txn(
...
```
clokep, line 1323:
```suggestion
            lock: True to lock the table when doing the upsert. Unused if the database engine
            	supports native upserts.
```

clokep (commented):
Pretty sure this is fine, but a couple questions on the locking!

synapse/storage/database.py
```diff
@@ -1360,11 +1360,17 @@ def simple_upsert_many_txn_emulated(
         if not value_names:
             value_values = [() for x in range(len(key_values))]
 
+        if lock:
+            # Lock the table just once, to prevent it being done once per row.
+            # Note that, according to Postgres' documentation, once obtained,
+            # the lock is held for the remainder of the current transaction.
+            self.engine.lock_table(txn, "user_ips")
```
clokep, line 1367: I don't think this code is used on postgres though since that doesn't need the emulated support here?

synapse/storage/database.py
clokep: @reivilibre judging from your commits my comment above was accurate, I think?

synapse/storage/database.py
reivilibre: Well, sigh, you're absolutely right. Locking on SQLite is a no-op (apparently! I haven't fully reasoned through how that's sufficient, but the code is a no-op...?!).

* Postgres: always supports native upsert, so locking doesn't apply.
* SQLite: may not support native upsert, locking is a no-op.

Why do upserts have locking at all then? Maybe I made the mistake of assuming that there must have been a reason for it because of the code that was here before...


In one of the cases, an emulated upsert can be a `SELECT` followed by an `INSERT`. In the other case, it's an `UPDATE` followed by an `INSERT`...

The latter case (UPDATE, INSERT) doesn't sound like it would need a lock, because both of those operations need a PENDING lock. There's no way another writer can get in the middle of those two.

(Assuming traditional SQLite3 locking) The first case (SELECT, INSERT) may be problematic because `SELECT` only needs a SHARED lock.  But it needs to upgrade to a PENDING lock when it begins the `INSERT`. If there's another PENDING or stronger lock, it will fail and will need to retry the transaction.

I can't tell whether this works properly in WAL mode or not; the WAL docs aren't particularly clear about how transactions interact like the old-style locking page is. That said, the page on transactions has this to say:

> A read transaction is used for reading only. A write transaction allows both reading and writing. A read transaction is started by a SELECT statement, and a write transaction is started by statements like CREATE, DELETE, DROP, INSERT, or UPDATE (collectively "write statements"). If a write statement occurs while a read transaction is active, then the read transaction is upgraded to a write transaction if possible. If some other database connection has already modified the database or is already in the process of modifying the database, then upgrading to a write transaction is not possible and the write statement will fail with [SQLITE_BUSY](https://www.sqlite.org/rescode.html#busy).

Assuming that's still true for WAL mode, then I think it's safe on SQLite.

---

Ah! However, there's one case where Postgres DOES need the emulated support, so I need to take back some of the things I just wrote :/.
That's when the unique index on the table is not ready yet and the table is 'unsafe to upsert'. (This is only a notable effect with Postgres, since SQLite's indices block when adding; Postgres' are added concurrently sometimes). In that case, locking does serve a useful role because Postgres' MVCC means that duplicates could be introduced (and then lead to the failure of creating the index altogether).
Aside: `user_directory_search` is an eternal exception — it's *always* 'unsafe to upsert' on SQLite....


That turned into something more complex than I thought :/

clokep (approved):
From our conversations, I think this is fine.

synapse/storage/database.py
reivilibre: > > A read transaction is used for reading only. A write transaction allows both reading and writing. A read transaction is started by a SELECT statement, and a write transaction is started by statements like CREATE, DELETE, DROP, INSERT, or UPDATE (collectively "write statements"). If a write statement occurs while a read transaction is active, then the read transaction is upgraded to a write transaction if possible. If some other database connection has already modified the database or is already in the process of modifying the database, then upgrading to a write transaction is not possible and the write statement will fail with [SQLITE_BUSY](https://www.sqlite.org/rescode.html#busy).
> 
> Assuming that's still true for WAL mode, then I think it's safe on SQLite.

I tried this and it seems safe on both WAL and non-WAL mode.

The advantage of WAL mode in this case is that a transaction merely running 'select' and hanging around doesn't prevent you from committing. But that transaction will have to restart (at least if it's in conflict; not sure about generally) — it gets told the database is locked each time it tries to commit or update.

I learnt a bit about SQLite's WAL mode, so that's good at least

PR #12262, opened 2022-03-21
Refuse to start if DB has an unsafe locale

### LINKED ISSUE #12055: Abort startup if unsafe DB locale
### Description

From https://github.com/matrix-org/synapse/issues/10718 :

> > Perhaps instead of a warning that should be made to abort startup unless a new setting `corrup_my_db_silently_on_libc_update: true` is set.
> 
> We're fine going with this approach to at least help people avoid silent corruption... details tbd.

We should probably also add the commands needed to drop the broken data from a DB to https://matrix-org.github.io/synapse/latest/postgres.html#fixing-incorrect-collate-or-ctype and also add:

> We should also probably link to information on how Glibc collation changes can break PostgreSQL. E.g., https://wiki.postgresql.org/wiki/Locale_data_changes

### Steps to reproduce

- have DB use wrong locale
- update libc with locale change
- notice corrupt DB

<!--

``` (three backticks, on a line on their own), so that they are formatted legibly.
-->

### Version information

<!-- Was this issue identified on matrix.org or another homeserver? 
- **Homeserver**:

If not matrix.org:
-->
<!--

listen on that port.)
-->
- **Version**: seen e.g. on 1.41.0, but bug exists before and after

- **Install method**: any

- **Platform**: Linux glibc
<!--
-->

### PR DESCRIPTION
**Pull request by H-Shay**
Currently Synapse will refuse to start new deployments if they have an unsafe DB locale. This PR extends that behavior to existing deployments, unless the user sets the new config flag `allow_unsafe_locale` added in this PR. Fixes #12055.

### DIFF
--- a/changelog.d/12262.misc
+++ b/changelog.d/12262.misc
@@ -1,0 +1,1 @@
+Refuse to start if DB has non-c locale, unless config flag `allow_unsafe_db_locale` is set to true.
--- a/changelog.d/12262.misc
+++ b/changelog.d/12262.misc
@@ -1,1 +1,1 @@
-Refuse to start if DB has non-c locale, unless config flag `allow_unsafe_db_locale` is set to true.
+Refuse to start if DB has non-`C` locale, unless config flag `allow_unsafe_db_locale` is set to true.
--- a/synapse/config/database.py
+++ b/synapse/config/database.py
@@ -37,8 +37,8 @@ DEFAULT_CONFIG = """\
 # 'txn_limit' gives the maximum number of transactions to run per connection
 # before reconnecting. Defaults to 0, which means no limit.
 #
-# 'allow_unsafe_locale' is an option specific to postgres. Under the default behavior, Synapse will refuse to
-# start if the postgres db is set to a non-c locale. You can override this behavior (which is *not* recommended)
+# 'allow_unsafe_locale' is an option specific to Postgres. Under the default behavior, Synapse will refuse to
+# start if the postgres db is set to a non-C locale. You can override this behavior (which is *not* recommended)
 # by setting 'allow_unsafe_locale' to true. Note that doing so may corrupt your database. You can find more information
 # here: https://matrix-org.github.io/synapse/latest/postgres.html#fixing-incorrect-collate-or-ctype and here:
 # https://wiki.postgresql.org/wiki/Locale_data_changes
--- a/synapse/config/database.py
+++ b/synapse/config/database.py
@@ -37,6 +37,12 @@ DEFAULT_CONFIG = """\
 # 'txn_limit' gives the maximum number of transactions to run per connection
 # before reconnecting. Defaults to 0, which means no limit.
 #
+# 'allow_unsafe_locale' is an option specific to postgres. Under the default behavior, Synapse will refuse to
+# start if the postgres db is set to a non-c locale. You can override this behavior (which is *not* recommended)
+# by setting 'allow_unsafe_locale' to true. Note that doing so may corrupt your database. You can find more information
+# here: https://matrix-org.github.io/synapse/latest/postgres.html#fixing-incorrect-collate-or-ctype and here:
+# https://wiki.postgresql.org/wiki/Locale_data_changes
+#
 # 'args' gives options which are passed through to the database engine,
 # except for options starting 'cp_', which are used to configure the Twisted
 # connection pool. For a reference to valid arguments, see:
--- a/synapse/storage/engines/postgres.py
+++ b/synapse/storage/engines/postgres.py
@@ -47,10 +47,18 @@ class PostgresEngine(BaseDatabaseEngine):
         self.default_isolation_level = (
             self.module.extensions.ISOLATION_LEVEL_REPEATABLE_READ
         )
+        self.config = database_config
 
     @property
     def single_threaded(self) -> bool:
         return False
+
+    def get_db_locale(self, txn):
+        txn.execute(
+            "SELECT datcollate, datctype FROM pg_database WHERE datname = current_database()"
+        )
+        collation, ctype = txn.fetchone()
+        return collation, ctype
 
     def check_database(self, db_conn, allow_outdated_version: bool = False):
         # Get the version of PostgreSQL that we're using. As per the psycopg2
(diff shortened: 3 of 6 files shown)

### CODE REVIEW

File: changelog.d/12262.misc [hunk 1]
```diff
@@ -0,0 +1 @@
+Refuse to start if DB has non-c locale, unless config flag `allow_unsafe_db_locale` is set to true.
```
reivilibre, line 1:
```suggestion
Refuse to start if DB has non-`C` locale, unless config flag `allow_unsafe_db_locale` is set to true.
```

C is a proper noun... probably. And I suppose it's a constant.

synapse/config/database.py
```diff
@@ -37,6 +37,12 @@
 # 'txn_limit' gives the maximum number of transactions to run per connection
 # before reconnecting. Defaults to 0, which means no limit.
 #
+# 'allow_unsafe_locale' is an option specific to postgres. Under the default behavior, Synapse will refuse to
+# start if the postgres db is set to a non-c locale. You can override this behavior (which is *not* recommended)
```
reivilibre, line 41:
```suggestion
# 'allow_unsafe_locale' is an option specific to Postgres. Under the default behavior, Synapse will refuse to
# start if the postgres db is set to a non-C locale. You can override this behavior (which is *not* recommended)
```

synapse/storage/engines/postgres.py
```diff
@@ -72,33 +81,31 @@ def check_database(self, db_conn, allow_outdated_version: bool = False):
...
-            )
-            collation, ctype = txn.fetchone()
+            collation, ctype = self.get_db_locale(txn)
             if collation != "C":
-                logger.warning(
-                    "Database has incorrect collation of %r. Should be 'C'\n"
-                    "See docs/postgres.md for more information.",
-                    collation,
-                )
+                if not allow_unsafe_locale:
```
reivilibre, line 86: Would be tempted to flatten this nested if to `if collation != "C" and not allow_unsafe_locale:`.
Actually, scratch that, I'd be tempted to still log a warning when the config flag is enabled, just thinking that it's good to have the log file reveal when a 'tainted configuration' is being used.

tests/storage/test_unsafe_locale.py
```diff
@@ -0,0 +1,38 @@
...
+from unittest.mock import patch
+
+from synapse.storage.database import make_conn
+from synapse.storage.engines._base import IncorrectDatabaseSetup
+from synapse.storage.engines.postgres import PostgresEngine
+
+from tests.unittest import HomeserverTestCase
+
+
+class UnsafeLocaleTest(HomeserverTestCase):
```
reivilibre, line 23: You will probably need to add the import for this, but looking at other tests with `rg` (since I vaguely remembered seeing it), this is how we usually skip Postgres-specific tests.
That way, they're neither marked as a failure nor a success on SQLite, which I guess is pedantically more correct.
Would imagine it avoids the overhead of setting up a test homeserver + database just to not do anything :-).

```suggestion
class UnsafeLocaleTest(HomeserverTestCase):
    if not USE_POSTGRES_FOR_TESTS:
        skip = "Requires Postgres"

```

tests/storage/test_unsafe_locale.py
```diff
@@ -0,0 +1,38 @@
...
+    @patch("synapse.storage.engines.postgres.PostgresEngine.get_db_locale")
+    def test_unsafe_locale(self, mock_db_locale):
+        mock_db_locale.return_value = ("B", "B")
```
reivilibre, line 26: ROFL at `B`...... Not sure I've ever seen *that* locale before :D. (no problem with leaving it; it's 'not C', just not a 'not C' I was expecting)

tests/storage/test_unsafe_locale.py
```diff
@@ -0,0 +1,38 @@
...
+        database = self.hs.get_datastores().databases[0]
+
+        # Only run this on postgres databases
+        if isinstance(database.engine, PostgresEngine):
```
reivilibre, line 30: If we skip the test on Postgres we don't need to worry about obscuring the test code with this guard (and get to avoid boomeranging it).
In other situations, I've been told enough times (since I always forget in my own code), this could have been an early return to prevent the boomeranging also.

tests/storage/test_unsafe_locale.py
```diff
@@ -0,0 +1,38 @@
...
+from synapse.storage.database import make_conn
+from synapse.storage.engines._base import IncorrectDatabaseSetup
+from synapse.storage.engines.postgres import PostgresEngine
+
+from tests.unittest import HomeserverTestCase
+
+
+class UnsafeLocaleTest(HomeserverTestCase):
+    @patch("synapse.storage.engines.postgres.PostgresEngine.get_db_locale")
+    def test_unsafe_locale(self, mock_db_locale):
```
reivilibre, line 25:
```suggestion
    def test_unsafe_locale(self, mock_db_locale) -> None:
```
Just so we can claim this file is type-checked :innocent: :-) (occasionally tests can be good places that exercise the type annotations a bit, providing valid type errors for mistakes in the actual code. Not sure it'd buy much here, but why not)

Don't know what the type of the Mock is, would be tempted to put it in but also suppose it stands very little chance of mattering.

tests/storage/test_unsafe_locale.py
```diff
@@ -0,0 +1,38 @@
...
+        # Only run this on postgres databases
+        if isinstance(database.engine, PostgresEngine):
+            db_conn = make_conn(
+                database._database_config, database.engine, "test_unsafe"
+            )
+            with self.assertRaises(IncorrectDatabaseSetup):
+                database.engine.check_database(db_conn)
+            with self.assertRaises(IncorrectDatabaseSetup):
+                database.engine.check_new_database(db_conn)
+            db_conn.close()
```
reivilibre, line 38: I would be tempted to add another test: one that runs `get_db_locale` *without* mocking it, checking that:

- the query works
- `"C", "C"` is being used in the tests

reivilibre (changes requested):
Looking good!

tests/storage/test_unsafe_locale.py
H-Shay: I figured it didn't matter and was a little cheeky joke but if you want me to change it to POSIX or whatever I am happy to do so.

reivilibre (approved):
Perfect! Thanks :)

PR #12266, opened 2022-03-22
Optionally include account validity in MSC3720 account status responses

babolivier:
Add a configuration option to include account validity expiration info in the responses to MSC3720 account status requests. This is because DINUM wants to use this info in Tchap.

I'm reluctant to add this new configuration flag to the sample configuration. It's not compliant with the MSC, account validity isn't part of the Matrix spec (and imo shouldn't be), and I don't want clients to start relying on them (apart from closed, controlled deployments like Tchap). Happy to discuss this further if the reviewer thinks otherwise.

### DIFF
--- a/synapse/config/server.py
+++ b/synapse/config/server.py
@@ -676,7 +676,7 @@ class ServerConfig(Config):
         ):
             raise ConfigError("'custom_template_directory' must be a string")
 
-        self.use_account_validity_in_account_status = (
+        self.use_account_validity_in_account_status: bool = (
             config.get("use_account_validity_in_account_status") or False
         )
 
--- a/synapse/config/server.py
+++ b/synapse/config/server.py
@@ -676,6 +676,10 @@ class ServerConfig(Config):
         ):
             raise ConfigError("'custom_template_directory' must be a string")
 
+        self.use_account_validity_in_account_status = (
+            config.get("use_account_validity_in_account_status") or False
+        )
+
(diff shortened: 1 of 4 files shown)


synapse/config/server.py
```diff
@@ -676,6 +676,10 @@ def read_config(self, config, **kwargs):
         ):
             raise ConfigError("'custom_template_directory' must be a string")
 
+        self.use_account_validity_in_account_status: bool = (
+            config.get("use_account_validity_in_account_status") or False
+        )
+
```
reivilibre, line 682: Considering the MSC is unstable, should this be in the experimental/unstable part of the config?

reivilibre (approved):
I think this is  the right move. At least the property is prefixed so that it's very obvious it's non-standard (this is something that bit us in the past with unspecced things; at least the way you've made it here makes it clear to client developers).

I only wonder if the config setting should be in the 'experimental' config section, like other MSC flags are.

synapse/config/server.py
babolivier: Experimental feels more to me like the place where we'd add toggles to enable entire MSC implems. Plus I'm not sure this qualifies as experimental in the same way other things in Synapse do - it's partly meant as a stopgap until Tchap stops relying on account validity (which should eventually be replaced by some LDAP-style central user directory). The long term plan is more to have it disappear when we don't need it anymore, rather than make it stable and expose it to the world.

PR #7519, opened 2020-05-15
Kill off some old python 2 code

richvdh:
This stuff has been annoying me for a while.

Each commit should stand alone.

(There's plenty more that can be removed; this was just what I found in 10 minutes.)

PR #3247, opened 2018-05-19
Misc Python3 fixes

NotAFile:
* use stand-in value if maxint is not available (this is the case with the py3
   arbitrary size ints)
 * fix py3 intern and remove unnecessary py3 encode (the encode("ascii") is
   afaict used to save space, however that doesn't bring any benefits on py3
 * py3-ize url preview (misc stuff I bunched together because it was blocking
   the unit tests)

PR #12267, opened 2022-03-22
Add type hints to the main state store.

clokep:
Just a little bit more. 👍


synapse/util/caches/__init__.py
```diff
@@ -195,8 +195,10 @@ def register_cache(
     )
 }
 
+T = TypeVar("T", Optional[str], str)
```
clokep, line 198: This is attempting to make it so that if we know we have a `str` before calling `intern_string` we don't lose that information (and have it become `Optional[str]` on the output). I'm not 100% sure this is correct though.

So pretty much `intern_string(Optional[str]) -> Optional[str]` while `intern_string(str) -> str`. Could maybe also be done with an overload?

DMRobertson: > Could maybe also be done with an overload?

For sure. I don't mind this way though: [typing.AnyStr](https://docs.python.org/3/library/typing.html?highlight=anystr#typing.AnyStr) is a nice example of this in the stdlib.

With that said: is there an example where this function is passed `None`? I guess `intern_dict` could have a key which maps to `None`?

synapse/util/caches/__init__.py
```diff
@@ -195,8 +195,10 @@ def register_cache(
...
 
-def intern_string(string: Optional[str]) -> Optional[str]:
+
+def intern_string(string: T) -> T:
     """Takes a (potentially) unicode string and interns it if it's ascii"""
```
DMRobertson, line 202: Is the comment about the string being ascii relevant now? Ditto for the try/except.

I can't see any mention of either at https://docs.python.org/3/library/sys.html?highlight=sys%20intern#sys.intern . I'd guess this was some PY2 compat stuff we can just ditch?

The history seems to be
-  https://github.com/matrix-org/synapse/pull/2155
- https://github.com/matrix-org/synapse/pull/3247
- https://github.com/matrix-org/synapse/pull/7519

DMRobertson (commented):
For `state.py`, I think Dirk might have beaten you to it in #12255.

synapse/util/caches/__init__.py
DMRobertson: (I was gonna say "a name more specific than `T`" might be nice. But I couldn't think of a good one. `FixedOptionalString`?

clokep: > With that said: is there an example where this function is passed `None`? I guess `intern_dict` could have a key which maps to `None`?

I think the only place it could be is the `intern_dict` call for the state key? I'm not 100% sure though...

DMRobertson: Nor I! Let's leave it be for now.

DMRobertson (approved):
Sorry for the delay here @clokep. LGTM!

PR #12269, opened 2022-03-22
Use psycopg2 type stubs

DMRobertson:
This didn't find anything interesting---one problem that I've suppressed
for now. But it feels good to feed more types into mypy's belly where
possible.

### DIFF
--- a/synapse/storage/database.py
+++ b/synapse/storage/database.py
@@ -305,7 +305,15 @@ class LoggingTransaction:
         from psycopg2.extras import execute_values  # type: ignore
 
         return self._do_execute(
-            lambda *x: execute_values(self.txn, *x, fetch=fetch), sql, *args
+            # Type ignore: mypy is unhappy because if `x` is a 5-tuple, then there will
+            # be two values for `fetch`: one given positionally, and another given
+            # as a keyword argument. We might be able to fix this by
+            # - propagating the signature of psycopg2.extras.execute_values to this
+            #   function, or
+            # - changing `*args: Any` to `values: T` for some appropriate T.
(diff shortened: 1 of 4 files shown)


reivilibre (approved):
Thanks!

synapse/storage/database.py
```diff
@@ -305,7 +305,15 @@ def execute_values(self, sql: str, *args: Any, fetch: bool = True) -> List[Tuple
...
-            lambda *x: execute_values(self.txn, *x, fetch=fetch), sql, *args
+            # Type ignore: mypy is unhappy because if `x` is a 5-tuple, then there will
+            # be two values for `fetch`: one given positionally, and another given
+            # as a keyword argument. We might be able to fix this by
+            # - propagating the signature of psycopg2.extras.execute_values to this
+            #   function, or
+            # - changing `*args: Any` to `values: T` for some appropriate T.
+            lambda *x: execute_values(self.txn, *x, fetch=fetch),  # type: ignore[misc]
+            sql,
+            *args,
```
DMRobertson, line 316: Cross-referencing:
- stub: https://github.com/python/typeshed/blob/dd1265c0e639c8231e8c9a66ddb116c30c8e07c2/stubs/psycopg2/psycopg2/extras.pyi#L181
- source: https://github.com/psycopg/psycopg2/blob/c34bf2f2f90fbbe65e8c4c2650734aa151ddd6ed/lib/extras.py#L1190

DMRobertson: I _think_ (as Patrick suggested in #synapse-dev) that we could change `*args` and `*x` to be `args` and `x`. But the rest of the source makes a lot of use of `args: Any` and I wasn't completely convinced that the change was safe or made things clearer.

PR #12271, opened 2022-03-23
Flesh out documentation for running SyTest against Synapse, including use of Postgres and worker mode.

reivilibre:
I thought it would be sensible to document these whilst I remember.

### DIFF
--- a/changelog.d/12271.doc
+++ b/changelog.d/12271.doc
@@ -1,1 +1,1 @@
-Flesh out documentation for running SyTest against Synapse, including use of Postgres and worker mode.
+Clarify documentation for running SyTest against Synapse, including use of Postgres and worker mode.
--- a/changelog.d/12271.doc
+++ b/changelog.d/12271.doc
@@ -1,0 +1,1 @@
+Flesh out documentation for running SyTest against Synapse, including use of Postgres and worker mode.
--- a/docs/development/contributing_guide.md
+++ b/docs/development/contributing_guide.md
@@ -261,7 +261,7 @@ This configuration should generally cover your needs.
 - To run with Postgres, supply the `-e POSTGRES=1 -e MULTI_POSTGRES=1` environment flags.
 - To run with Synapse in worker mode, supply the `-e WORKERS=1 -e REDIS=1` environment flags (in addition to the Postgres flags).
 
-For more details about other configurations, see [documentation in the SyTest repo](https://github.com/matrix-org/sytest/blob/develop/docker/README.md).
+For more details about other configurations, see the [Docker-specific documentation in the SyTest repo](https://github.com/matrix-org/sytest/blob/develop/docker/README.md).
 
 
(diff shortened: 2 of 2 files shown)


clokep:
I'm unsure if we do cross-link somewhere, but I usually reference https://github.com/matrix-org/sytest/tree/develop/docker#target-specific-details

docs/development/contributing_guide.md
```diff
@@ -254,8 +254,14 @@ configuration:
 ```sh
 $ docker run --rm -it -v /path/where/you/have/cloned/the/repository\:/src:ro -v /path/to/where/you/want/logs\:/logs matrixdotorg/sytest-synapse:buster
 ```
+(Note that the paths must be full paths! You could also write `$(realpath relative/path)` if needed.)
 
-This configuration should generally cover  your needs. For more details about other configurations, see [documentation in the SyTest repo](https://github.com/matrix-org/sytest/blob/develop/docker/README.md).
+This configuration should generally cover your needs.
+
+- To run with Postgres, supply the `-e POSTGRES=1 -e MULTI_POSTGRES=1` environment flags.
```
clokep, line 261: `MULTI_POSTGRES` seems undocumented on the sytest side, but seems to run separate databases for main and state. I feel like that's a somewhat weird thing to normally test against and is probably not necessarily here?

reivilibre: It seems more correct than running them in unified mode, since that means that queries can run against both (which of course won't work in multi-Postgres mode).
However this was just what I was told to run.

clokep: > It seems more correct than running them in unified mode, since that means that queries can run against both (which of course won't work in multi-Postgres mode).

I'm not sure why it seems more correct? Can you elaborate a bit about what you mean!

reivilibre: Sure, I can elaborate a bit on that last clause about what I mean!

If you have two separate Postgres databases, then fewer potential queries will work in that context than if you only had one. Put another way, a test result from a test with one unified database offers no implication that the code will work with split databases.
As a general idea, I think it's good to test things with the most restrictive configuration possible — in that sense, it'd be somewhat weird to test against the unified database (though of course there are reasons for that too, and in any case it's good to cover all the corners eventually, but maybe that explains why as a first approximation it seems less correct to test in a situation where the databases aren't isolated).

clokep: Can we document this on the sytest side so that when someone tries to figure out what they do by following the link it says something there?

clokep (approved):

changelog.d/12271.doc
```diff
@@ -0,0 +1 @@
+Flesh out documentation for running SyTest against Synapse, including use of Postgres and worker mode.
```
clokep, line 1: Maybe?

```suggestion
Clarify documentation for running SyTest against Synapse, including use of Postgres and worker mode.
```

PR #12272, opened 2022-03-23
Add cache for `get_membership_from_event_ids`

erikjohnston:
This should speed up push rule calculations for rooms with large numbers of local users when the main push rule cache fails.

### DIFF
--- a/synapse/storage/databases/main/events.py
+++ b/synapse/storage/databases/main/events.py
@@ -1745,7 +1745,7 @@ class PersistEventsStore:
                 (event.state_key,),
             )
 
-            # The `_get_joined_profile_from_event_id` is immutable, except for the
+            # The `_get_membership_from_event_id` is immutable, except for the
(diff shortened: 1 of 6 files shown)


synapse/storage/databases/main/events.py
```diff
@@ -1745,6 +1745,13 @@ def non_null_str_or_none(val: Any) -> Optional[str]:
                 (event.state_key,),
             )
 
+            # The `_get_joined_profile_from_event_id` is immutable, except for the
```
reivilibre, line 1748: The name in this comment doesn't match the cache you're invalidating; copy paste issue?

reivilibre (approved):
LGTM otherwise :)

synapse/storage/databases/main/events.py
erikjohnston: OMG you have no idea how often I got confused between these two.

PR #12273, opened 2022-03-23
Include bundled aggregations for the latest event in a thread

### LINKED ISSUE #12270: Thread bundled relationships can return a split-brained latest_event
### Description

The latest_event returned in an `m.thread` bundled relationship can be split-brained if it was edited, the content got updated but the `m.replace` relation is not signalled in any way, which means clients now have to manually fetch the relations on the latest_event to check whether it was edited or not.

`/event/` call on the thread root `await mxMatrixClientPeg.matrixClient.fetchRoomEvent("!cCpRBTFuXwPmOCUFRL:matrix.org", "$2fQ-tX_N6eXHllDPHN3gYD_2ghfbZkGfjvjQdUqxtFQ")`
```json
{
  "content": {
    "body": "root",
    "msgtype": "m.text",
    "org.matrix.msc1767.text": "root"
  },
  "origin_server_ts": 1647960269875,
  "room_id": "!cCpRBTFuXwPmOCUFRL:matrix.org",
  "sender": "@webdevguru_test31:matrix.org",
  "type": "m.room.message",
  "unsigned": {
    "age": 66278821,
    "m.relations": {
      "m.thread": {
        "latest_event": {
          "content": {
            "body": "testfoobar",
            "format": "org.matrix.custom.html",
            "formatted_body": "testfoobar",
            "msgtype": "m.text",
            "org.matrix.msc1767.message": [
              {
                "body": "testfoobar",
                "mimetype": "text/plain"
              },
              {
                "body": "testfoobar",
                "mimetype": "text/html"
              }
            ],
            "m.relates_to": {
              "event_id": "$2fQ-tX_N6eXHllDPHN3gYD_2ghfbZkGfjvjQdUqxtFQ",
              "is_falling_back": true,
              "m.in_reply_to": {
                "event_id": "$6WWJ7Hga3ZtlTehlThl-v2_doVAFs5EaqgCoffLM9gc"
              },
              "rel_type": "m.thread"
            }
          },
          "origin_server_ts": 1648024628463,
          "room_id": "!cCpRBTFuXwPmOCUFRL:matrix.org",
          "sender": "@webdevguru_test31:matrix.org",
          "type": "m.room.message",
          "unsigned": {
            "age": 1920233
          },
          "event_id": "$e8ENHc50vigxQU3NgeZ6I5A68dL1IGqouonJb500jpk",
          "user_id": "@webdevguru_test31:matrix.org",
          "age": 1920233
        },
        "count": 7,
        "current_user_participated": true
      }
    }
  },
  "event_id": "$2fQ-tX_N6eXHllDPHN3gYD_2ghfbZkGfjvjQdUqxtFQ",
  "user_id": "@webdevguru_test31:matrix.org",
  "age": 66278821
}
```

`/event/` call on the latest event in the thread `await mxMatrixClientPeg.matrixClient.fetchRoomEvent("!cCpRBTFuXwPmOCUFRL:matrix.org", "$e8ENHc50vigxQU3NgeZ6I5A68dL1IGqouonJb500jpk")`
```json
{
  "content": {
    "body": "testfoobar",
    "format": "org.matrix.custom.html",
    "formatted_body": "testfoobar",
    "msgtype": "m.text",
    "org.matrix.msc1767.message": [
      {
        "body": "testfoobar",
        "mimetype": "text/plain"
      },
      {
        "body": "testfoobar",
        "mimetype": "text/html"
      }
    ],
    "m.relates_to": {
      "event_id": "$2fQ-tX_N6eXHllDPHN3gYD_2ghfbZkGfjvjQdUqxtFQ",
      "is_falling_back": true,
      "m.in_reply_to": {
        "event_id": "$6WWJ7Hga3ZtlTehlThl-v2_doVAFs5EaqgCoffLM9gc"
      },
      "rel_type": "m.thread"
    }
  },
  "origin_server_ts": 1648024628463,
  "room_id": "!cCpRBTFuXwPmOCUFRL:matrix.org",
  "sender": "@webdevguru_test31:matrix.org",
  "type": "m.room.message",
  "unsigned": {
    "age": 1940276,
    "m.relations": {
      "m.replace": {
        "event_id": "$8fzcKK9FohMXMGhvGM13dYoLM1S6Ekgw8Oz2ojW2dl8",
        "origin_server_ts": 1648025699050,
        "sender": "@webdevguru_test31:matrix.org"
      }
    }
  },
  "event_id": "$e8ENHc50vigxQU3NgeZ6I5A68dL1IGqouonJb500jpk",
  "user_id": "@webdevguru_test31:matrix.org",
  "age": 1940276
}
```

Observe the event ID and content are the same, but the latter includes the `m.replace` relation which should be sanely returned in the first request also.

### Steps to reproduce

- Create a thread with root $root
- Reply to thread $root with event $event (content A)
- Edit event $event with event $edit (content B)
- Ask the server for $root via /event/ API and receive latest_event $event with content B and no details about m.replace or $edit

I would have expected latest_event to either be entirely unedited or edited with the relation.

### Version information

matrix.org

clokep:
Fixes #12270; we were previously handling edits to the latest event in a thread (see #11973), but not notifying the client that this *was* an edit via the bundled aggregations.

It is wanted that the bundled aggregations for the latest event in a thread include everything so it is easy to render the thread summary.

I think this is OK, but does go a little bit down a slippery slope -- I wonder if it would have been better to not include the full event (at the risk of causing many more event look-ups).


changelog.d/12273.bugfix
```diff
@@ -0,0 +1 @@
+Fix a bug introduced in Synapse v1.48.0 the latest event provided in bundled aggregations where it did not include bundled aggregations.
```
richvdh, line 1: This makes no sense!

synapse/events/utils.py
```diff
@@ -32,13 +32,13 @@
 from synapse.api.constants import EventContentFields, EventTypes, RelationTypes
 from synapse.api.errors import Codes, SynapseError
 from synapse.api.room_versions import RoomVersion
+from synapse.handlers.relations import BundledAggregations
```
richvdh, line 35: is this deliberate?

I'm thinking that the import was probably conditional for a reason; and that in any case we don't really want to add dependencies on `synapse.handlers` from `synapse.events`.

synapse/events/utils.py
```diff
@@ -472,7 +471,7 @@ def _inject_bundled_aggregations(
         event: EventBase,
         time_now: int,
         config: SerializeEventConfig,
-        aggregations: "BundledAggregations",
+        bundled_aggregations: Dict[str, "BundledAggregations"],
```
richvdh, line 474: docstring needs an update

synapse/events/utils.py
```diff
@@ -487,15 +486,22 @@ def _inject_bundled_aggregations(
         """
         serialized_aggregations = {}
 
-        if aggregations.annotations:
-            serialized_aggregations[RelationTypes.ANNOTATION] = aggregations.annotations
+        # The aggregations must exist for this event.
```
richvdh, line 489:
```suggestion
        # We have already checked that aggregations exist for this event.
```

synapse/events/utils.py
```diff
@@ -487,15 +486,22 @@ def _inject_bundled_aggregations(
         """
         serialized_aggregations = {}
```
richvdh, line 487: this is a bit out of place now, move down to after `event_aggregations = bundled_aggregations[event.event_id]` ?

synapse/events/utils.py
```diff
@@ -506,18 +512,16 @@ def _inject_bundled_aggregations(
             }
 
         # If this event is the start of a thread, include a summary of the replies.
-        if aggregations.thread:
-            thread = aggregations.thread
+        if event_aggregations.thread:
+            thread = event_aggregations.thread
 
             # Don't bundle aggregations as this could recurse forever.
```
richvdh, line 518: this comment seems to be misplaced now. And do we need to worry about it recursing forever?

synapse/handlers/relations.py
```diff
@@ -153,7 +151,7 @@ async def get_relations(
 
     async def _get_bundled_aggregation_for_event(
         self, event: EventBase, user_id: str
-    ) -> Optional[BundledAggregations]:
+    ) -> Tuple[Optional[JsonDict], Optional[JsonDict]]:
```
richvdh, line 154: docstring needs an update

richvdh: does the method need renaming? it seems like it no longer returns the "bundled aggregation" [sic]?

synapse/handlers/relations.py
```diff
@@ -210,8 +203,13 @@ async def get_bundled_aggregations(
             user_id: The user requesting the bundled aggregations.
 
         Returns:
-            A map of event ID to the bundled aggregation for the event. Not all
-            events may have bundled aggregations in the results.
+            A map of event ID to the bundled aggregation for the event.
```
richvdh, line 206: I'm a bit confused about the terminology here. The return type is `BundledAggregation`**`s`**, but this text suggests there can only be one?

synapse/storage/databases/main/relations.py
```diff
@@ -444,7 +444,7 @@ def get_thread_summary(self, event_id: str) -> Optional[Tuple[int, EventBase]]:
     @cachedList(cached_method_name="get_thread_summary", list_name="event_ids")
     async def get_thread_summaries(
         self, event_ids: Collection[str]
-    ) -> Dict[str, Optional[Tuple[int, EventBase, Optional[EventBase]]]]:
+    ) -> Dict[str, Optional[Tuple[int, EventBase]]]:
         """Get the number of threaded replies, the latest reply (if any), and the latest edit for that reply for the given event.
```
richvdh, line 448: this is no longer correct

synapse/handlers/relations.py
```diff
@@ -223,42 +221,61 @@ async def get_bundled_aggregations(
...
+            [event_id for event_id, summary in summaries.items() if summary],
+            user_id,
         )
         for event_id, summary in summaries.items():
             if summary:
-                thread_count, latest_thread_event, edit = summary
+                thread_count, latest_thread_event = summary
+
+                # If the latest event in a thread is not already being fetched,
+                # add it to the events. (Note that we don't mo
```
richvdh, line 243: this seems to have run out of wind

richvdh (changes requested):
I've got to say I'm struggling to follow all this, though it seems plausible enough. I have a few questions and a few requests for changes.

It seems like other HS implementations could easily make the same mistake, and that we should make the behaviour explicit in the spec and/or add a complement test to enforce it.

clokep:
There was some chatter about this near https://matrix.to/#/!XaqDhxuTIlvldquJaV:matrix.org/$poGNZCThWxibXwbicuCh1J3IRDaIc5Brz1npmIA2u2c?via=matrix.org&via=element.io&via=vector.modular.im

synapse/events/utils.py
clokep: I think an older version required this, but definitely not needed now. 👍

synapse/handlers/relations.py
clokep: I think it was just a typo, it should be "bundled aggregations". There is only a single "bundle" per type of relation, but since there are multiple types of relations there are multiple "aggregations". 🤷 The terminology is all a bit confusing, yes.

synapse/events/utils.py
clokep: I think we do still in some odd circumstances. I need to ponder this a little more.

synapse/events/utils.py
```diff
@@ -506,19 +515,24 @@ def _inject_bundled_aggregations(
...
-            thread = aggregations.thread
+        # Include any threaded replies to this event.
+        #
+        # Note that is it not valid to start a thread on an event which already
+        # has a relation. Doing so would cause an infinite loop.
+        event_relation = event.content.get("m.relates_to")
+        include_thread = (
+            not isinstance(event_relation, dict)
+            or event_relation.get("rel_type") is None
+        )
```
clokep, line 526: Maybe it would make more sense to pass this in as a flag, like is done for `apply_edits`?

clokep:
@richvdh Sorry for the extreme delaying in getting back to this. I've made some updates, resolved some conflicts. If you've lost all context and want to throw this back on the team pile that's quite fine too!

clokep:
It seems I have a failing test, sorry about that.

tests/rest/client/test_relations.py
```diff
@@ -1070,7 +1033,143 @@ def assert_annotations(bundled_aggregations: JsonDict) -> None:
...
+        # Fetch the thread root, to get the bundled aggregation for the thread.
+        relations = self._get_bundled_aggregations()
+
+        # The latest event should have bundled aggregations.
+        self.assertIn(RelationTypes.THREAD, relations)
+        thread_summary = relations[RelationTypes.THREAD]
+        self.assertIn("latest_event", thread_summary)
+
+        # The latest event should not have any bundled aggregations (since it
+        # only has a thread attached.
```
richvdh, line 1109:
```suggestion
        # The latest event should not have any bundled aggregations (since the
        # only relation to it is another thread, which is invalid).
```

likewise below.

tests/rest/client/test_relations.py
```diff
@@ -1070,7 +1033,143 @@ def assert_annotations(bundled_aggregations: JsonDict) -> None:
...
+    def test_thread_loop(self) -> None:
+        """Ensure that bogus events do not cause the bundled aggregations code to iterate forever."""
+        last_event = self.parent_id
+
+        # Disable the validation to pretend this came over federation.
+        with patch(
+            "synapse.handlers.message.EventCreationHandler._validate_event_relation",
+            new=lambda self, event: make_awaitable(None),
+        ):
+            for _ in range(2):
```
richvdh, line 1094: can't help but feel that the loop is too clever for its own good and is making it harder to follow the flow here. I'd just unroll it (and note that only one of the `_send_relation` calls needs the patch).

tests/rest/client/test_relations.py
```diff
@@ -1070,7 +1033,143 @@ def assert_annotations(bundled_aggregations: JsonDict) -> None:
...
+                )
+                last_event = channel.json_body["event_id"]
+
+        # Fetch the thread root, to get the bundled aggregation for the thread.
+        relations = self._get_bundled_aggregations()
+
+        # The latest event should have bundled aggregations.
+        self.assertIn(RelationTypes.THREAD, relations)
+        thread_summary = relations[RelationTypes.THREAD]
+        self.assertIn("latest_event", thread_summary)
```
richvdh, line 1106: we should probably check that the `latest_event` is the event we think it is (ie, the first one sent by the loop above)?

tests/rest/client/test_relations.py
```diff
@@ -1070,7 +1033,143 @@ def assert_annotations(bundled_aggregations: JsonDict) -> None:
...
+        # We expect that the edit message appears in the thread summary in the
+        # unsigned relations section.
+        relations_dict = channel.json_body["unsigned"].get("m.relations")
+        self.assertIn(RelationTypes.THREAD, relations_dict)
+
+        thread_summary = relations_dict[RelationTypes.THREAD]
+        self.assertIn("latest_event", thread_summary)
+        latest_event_in_thread = thread_summary["latest_event"]
+        self.assertEqual(latest_event_in_thread["content"]["body"], "I've been edited!")
+        # The event should also have edit appear under the bundled aggregations.
```
richvdh, line 1168: I'm not sure what this comment is trying to say, but it doesn't seem to make sense right now.

richvdh (changes requested):
I'm afraid I *had* completely paged this out, so my apologies if I'm retreading old ground here.

This is broadly looking good and making sense. A few thoughts and questions.

Main question though: is this behaviour defined in an MSC somewhere, or at least mentioned as a comment on one? It feels quite subtle and should definitely make it into the spec.

Relatedly: I feel like we need a Complement test/sytest for it.

synapse/handlers/relations.py
```diff
@@ -433,13 +434,38 @@ async def get_bundled_aggregations(
...
+            latest_thread_event = thread.latest_event
+            if latest_thread_event and latest_thread_event.event_id not in events_by_id:
+                events_by_id[latest_thread_event.event_id] = latest_thread_event
+
         # Fetch other relations per event.
         for event in events_by_id.values():
-            event_result = await self._get_bundled_aggregation_for_event(
+            (
+                annotations,
+                references,
```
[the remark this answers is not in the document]
clokep: It unfortunately will not, I think it only has that behavior for parameters, not for destructured results?

tests/rest/client/test_relations.py
```diff
@@ -1070,7 +1033,143 @@ def assert_annotations(bundled_aggregations: JsonDict) -> None:
                 bundled_aggregations.get("latest_event"),
             )
 
-        self._test_bundled_aggregations(RelationTypes.THREAD, assert_annotations, 9)
+        self._test_bundled_aggregations(RelationTypes.THREAD, assert_thread, 10)
+
+    @unittest.override_config({"experimental_features": {"msc3666_enabled": True}})
```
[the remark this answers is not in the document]
clokep: That was missed from #12393, it seems.

tests/rest/client/test_relations.py
```diff
@@ -1070,7 +1033,143 @@ def assert_annotations(bundled_aggregations: JsonDict) -> None:
...
+        # Disable the validation to pretend this came over federation.
+        with patch(
+            "synapse.handlers.message.EventCreationHandler._validate_event_relation",
+            new=lambda self, event: make_awaitable(None),
+        ):
+            for _ in range(2):
+                channel = self._send_relation(
+                    RelationTypes.THREAD, "m.room.test", parent_id=last_event
+                )
+                last_event = channel.json_body["event_id"]
```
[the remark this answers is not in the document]
clokep: I've updated the docstring to say it is a thread to a thread, which is not allowed.

tests/rest/client/test_relations.py
clokep: 823a6b11c05dae9d2fde70045cb7e4e63b6766d4

tests/rest/client/test_relations.py
clokep: 823a6b11c05dae9d2fde70045cb7e4e63b6766d4

tests/rest/client/test_relations.py
clokep: 823a6b11c05dae9d2fde70045cb7e4e63b6766d4

tests/rest/client/test_relations.py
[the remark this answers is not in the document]
clokep: I did a separate PR #12511, thanks for pointing this out!

synapse/events/utils.py
```diff
@@ -506,19 +515,24 @@ def _inject_bundled_aggregations(
                 "sender": edit.sender,
             }
 
-        # If this event is the start of a thread, include a summary of the replies.
-        if aggregations.thread:
-            thread = aggregations.thread
+        # Include any threaded replies to this event.
+        #
+        # Note that is it not valid to start a thread on an event which already
+        # has a relation. Doing so would cause an infinite loop.
```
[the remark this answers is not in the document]
clokep: Just long, good call. 👍

synapse/handlers/relations.py
```diff
@@ -256,10 +254,11 @@ async def get_annotations_for_event(
 
         return filtered_results
 
-    async def _get_bundled_aggregation_for_event(
+    async def _get_bundled_annotations_and_references_for_event(
```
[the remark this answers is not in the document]
clokep: Yeah, I actually wonder if we even need it at all and can just inline the calls out to the inner methods...

clokep:
This somehow grew to a point where there's quite a bit of refactoring in it, I've split that out to #12519 to focus this PR more on behavior changes.

synapse/events/utils.py
```diff
@@ -506,19 +515,24 @@ def _inject_bundled_aggregations(
...
+        event_relation = event.content.get("m.relates_to")
+        include_thread = (
+            not isinstance(event_relation, (dict, frozendict))
```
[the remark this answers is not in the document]
clokep: We seem to be about 50-50 on which to use:

```
$ grep -rE "\(dict, frozendict\)" synapse 
synapse/util/frozenutils.py:    if isinstance(o, (dict, frozendict)):
synapse/storage/databases/main/state.py:        if not isinstance(predecessor, (dict, frozendict)):
synapse/events/utils.py:    elif isinstance(value, (dict, frozendict)):
synapse/handlers/relations.py:            if isinstance(relates_to, (dict, frozendict)):
$ grep -r "abc.Mapping" synapse
synapse/crypto/event_signing.py:    if not isinstance(hashes, collections.abc.Mapping):
synapse/events/validator.py:        "object", lambda checker, thing: isinstance(thing, collections.abc.Mapping)
synapse/events/utils.py:    if not isinstance(old_power_levels, collections.abc.Mapping):
synapse/events/utils.py:        if isinstance(v, collections.abc.Mapping):
```

I can switch to `collections.abc.Mapping` if we think that's clearer.

synapse/handlers/relations.py
```diff
@@ -301,18 +297,19 @@ async def _get_bundled_aggregation_for_event(
...
                 "chunk": [{"event_id": event.event_id} for event in references]
             }
 
             if next_token:
-                aggregations.references["next_batch"] = await next_token.to_string(
+                serialized_references["next_batch"] = await next_token.to_string(
                     self._main_store
                 )
 
         # Store the bundled aggregations in the event metadata for later use.
```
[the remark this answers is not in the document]
clokep: No...but it got removed at some point. 🤷

tests/rest/client/test_relations.py
clokep: c48b22056e428bee9458ad473a1fb5acb8bf73ab hopefully clarifies this.

synapse/handlers/relations.py
```diff
@@ -369,21 +371,49 @@ async def get_bundled_aggregations(
...
         # Fetch any ignored users of the requesting user.
         ignored_users = await self._main_store.ignored_users(user_id)
 
+        # Threads are special as the latest event of a thread might cause additional
+        # events to be fetched. Thus, we check those first!
+
+        # Fetch thread summaries (but only for the directly requested events).
+        threads = await self.get_threads_for_events(
+            # It is not valid to start a thread on an event which already has a relation.
+            [eid for eid in events_by_id.keys() if eid not in relations_by_id],
```
clokep, line 395: I moved the filtering to here (instead of `_inject_bundled_aggregations`) as this better matches what we already did for annotations and edits. (See below, near line 411.)

This should avoid requesting data from the database which we will just throw away later on.

clokep:
@richvdh I think this is ready for another look, I answered the questions from your last review above: https://github.com/matrix-org/synapse/pull/12273#issuecomment-1103860011

I've resolved all threads which I think are closed.

synapse/events/utils.py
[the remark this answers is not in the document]
richvdh: hmph. Quite a few of those `(dict, frozendict)`s are quite recent. eg [11df4ec](https://github.com/matrix-org/synapse/commit/11df4ec6c2178bdd25652644a94c213c3064d693#diff-50843e15ebdedb5e273adb2561e7b74f754262bed804c4c63f5c7e2407954996L161) goes the other way,

I have a mild preference for using Mapping, but it's very mild.

synapse/handlers/relations.py
```diff
@@ -453,19 +405,43 @@ async def get_bundled_aggregations(
...
-            ) = await self._get_bundled_annotations_and_references_for_event(
-                event, ignored_users
+            # Edits and annotations may not have related annotations or references.
+            if relations_by_id.get(event.event_id) in (
+                RelationTypes.ANNOTATION,
+                RelationTypes.REPLACE,
+            ):
+                continue
+
+            annotations = await self.get_annotations_for_event(
```
richvdh, line 420:
```suggestion
            # look for annotations (reactions) which we should bundle in with this  event
            annotations = await self.get_annotations_for_event(
```

synapse/handlers/relations.py
```diff
@@ -453,19 +405,43 @@ async def get_bundled_aggregations(
...
+                event.event_id, event.room_id, ignored_users=ignored_users
+            )
+            if annotations:
+                results.setdefault(
+                    event.event_id, BundledAggregations()
+                ).annotations = {"chunk": annotations}
+
+            references, next_token = await self.get_relations_for_event(
```
richvdh, line 428:
```suggestion
            # look for references (replies) which we should bundle in with this event.
            references, next_token = await self.get_relations_for_event(
```

synapse/handlers/relations.py
```diff
@@ -453,19 +405,43 @@ async def get_bundled_aggregations(
             latest_thread_event = thread.latest_event
             if latest_thread_event and latest_thread_event.event_id not in events_by_id:
                 events_by_id[latest_thread_event.event_id] = latest_thread_event
+                # The latest event in the thread must have a thread relation.
```
richvdh, line 408: I'm not following this. "must have" according to whom? Evidently it doesn't have a thread relation, otherwise we wouldn't have to do this?

synapse/events/utils.py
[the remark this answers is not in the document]
clokep: > I have a mild preference for using Mapping, but it's very mild.

I have a strong preference for consistency! I made the changes separately: #12564. Will update any new occurrences in this PR though.

synapse/handlers/relations.py
```diff
@@ -438,11 +390,11 @@ async def get_bundled_aggregations(
         # events to be fetched. Thus, we check those first!
 
         # Fetch thread summaries (but only for the directly requested events).
-        #
-        # Note that you can't have threads off of other related events, but it is
-        # possible for a malicious homeserver to inject them anyway.
         threads = await self.get_threads_for_events(
-            events_by_id.keys(), user_id, ignored_users
+            # It is not valid to start a thread on an event which already has a relation.
```
richvdh, line 394: Currently this somehow implies to me an event which is the *target* of a relation.

```suggestion
            # It is not valid to start a thread on an event which itself relates to another event.
```

clokep:
> > Main question though: is this behaviour defined in an MSC somewhere, or at least mentioned as a comment on one? It feels quite subtle and should definitely make it into the spec.
> 
> No, not really. The only description of this field is a single line in the MSC:

matrix-org/matrix-spec-proposals#3785 was made to clarify this.

synapse/handlers/relations.py
```diff
@@ -369,21 +371,51 @@ async def get_bundled_aggregations(
             event.event_id: event for event in events if not event.is_state()
         }
 
+        # A map of event ID to the relation in that event, if there is one.
+        relations_by_id = {}
```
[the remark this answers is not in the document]
clokep: It doesn't really matter if it is a `str` or not since we only check it against string constants, but I can assert that if you would prefer.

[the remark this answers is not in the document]
richvdh: don't mind, it was only to make the type annotation nicer really.

synapse/handlers/relations.py
```diff
@@ -453,19 +405,43 @@ async def get_bundled_aggregations(
...
+                relations_by_id[latest_thread_event.event_id] = RelationTypes.THREAD
 
         # Fetch other relations per event.
         for event in events_by_id.values():
-            (
-                annotations,
-                references,
-            ) = await self._get_bundled_annotations_and_references_for_event(
-                event, ignored_users
+            # Edits and annotations may not have related annotations or references.
```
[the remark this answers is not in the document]
clokep: This comment was originally added in #11284 (and then has moved around quite a bit...); it originally comes from a comment in the REST layer:

https://github.com/matrix-org/synapse/blob/6a5dd485bd82b269e7e169c0385290d081eae801/synapse/rest/client/relations.py#L233-L235

As part of #11284 I seem to have made it so that this behavior applies only to annotations (reactions) and references (replies).

I think the following might be clearer:

> It doesn't make sense for an annotation (ie, reaction) or reference (ie, reply) to have any other events related to them.

The original comment seems confusing that it talks about edits though, I suspect that was a typo. There was a thread on #11284 which talks about this: https://github.com/matrix-org/synapse/pull/11284#discussion_r754362364

Note that [MSC2677](https://github.com/uhoreg/matrix-doc/blob/aggregations-reactions/proposals/2677-reactions.md#edge-cases) says:

> Can you react to a reaction?
> * Yes, at the protocol level. But you shouldn't expect clients to do anything useful with it.

...but it doesn't say anything about whether those should be bundled. Maybe this check is just bogus here and we should remove it?

synapse/handlers/relations.py
clokep: The latest event is an event which has a `rel_type` = `m.thread` to an event. (That's what we just searched for above.) The latest events are *new events* found, not part of the original input set, so we're expanding the information we calculated above.

synapse/handlers/relations.py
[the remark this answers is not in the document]
clokep: Actually, I can probably change the `relation_type is None` check to an `isinstance(relation_type, str)` to make it a `Dict[str, str]` -- would that be clearer?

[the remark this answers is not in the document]
clokep: I made this change in 6bcb2dce95f4434b64378615a3cb4118436a211c, which I think makes things clearer. 👍

synapse/handlers/relations.py
richvdh: yup, much clearer, thanks.

richvdh (approved):
lgtm. Sorry this one has taken so long!

clokep:
> lgtm. Sorry this one has taken so long!

No problem! Thank you for the thorough review!

Issue #12274, opened 2022-03-23
org.matrix.login.jwt login succeeded even if user is deactivated


laurent-treeb:
Version: 1.46 to 1.54

 1. Server configured with `jwt_config.enabled=true`
 1. Login a test user  with `type=org.matrix.login.jwt` and token
 1. With admin user, deactivate the test user `/_synapse/admin/v1/deactivate/{userid}`
 1. Relogin the test user  with `type=org.matrix.login.jwt` and token 
    - -> here the login is success
    - => should be refused with `403` `M_USER_DEACTIVATED`

laurent-dol:
https://github.com/matrix-org/synapse/pull/12276

laurent-dol:
The pull request doesn't manage the reactivation of the user.
When a user is reactivate via 
/_synapse/admin/v2/users/{userid} and "deactivated": false
a password value is required.

But in case of org.matrix.login.jwt, there is no password, so maybe the password should be optional on reactivation, but i don't evaluate the impact on others features.

anoadragon453:
Thank you for filing this and for providing a PR that aims to fix it! I agree that a JWT user should be unable to log in when their account is deactivated.

> When a user is reactivate via /_synapse/admin/v2/users/{userid} and "deactivated": false a password value is required.

The inability to reactivate users without a local password is tracked at https://github.com/matrix-org/synapse/issues/10397.

bradjones1:
Is this not a security issue?

PR #7240, opened 2020-04-07
Do not allow a deactivated user to login via SSO.

clokep:
This checks the deactivated status of a user during login for SSO.

If a user is deactivated, they are shown an HTML page explaining what happened.

As part of this it makes the `complete_sso_login` on the auth handler async, unfortunately to keep API compatibility in `ModuleApi` this PR adds a separate `complete_sso_login_async` and deprecates `complete_sso_login`. The sync version will not do the deactivation check.

PR #12276, opened 2022-03-23
#12274 org.matrix.login.jwt login succeeded even if user is deactivated

laurent-treeb:
### Pull Request Checklist

* [ ] Pull request includes a [sign off](https://matrix-org.github.io/synapse/latest/development/contributing_guide.html#sign-off)

### DIFF
--- a/synapse/rest/client/login.py
+++ b/synapse/rest/client/login.py
@@ -18,7 +18,7 @@ from typing import TYPE_CHECKING, Any, Awaitable, Callable, Dict, List, Optional
 
 from typing_extensions import TypedDict
 
-from synapse.api.errors import Codes, LoginError, SynapseError
+from synapse.api.errors import Codes, LoginError, SynapseError, UserDeactivatedError
 from synapse.api.ratelimiting import Ratelimiter
 from synapse.api.urls import CLIENT_API_PREFIX
 from synapse.appservice import ApplicationService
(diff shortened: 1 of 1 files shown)


dklimpel:
Is it useful to create a test to avoid it in feature?

synapse/rest/client/login.py
```diff
@@ -319,6 +319,12 @@ async def _complete_login(
                 canonical_uid = await self.registration_handler.register_user(
                     localpart=UserID.from_string(user_id).localpart
                 )
+            else: # check if user deactivated
+                deactivated = await self.hs.get_datastore().get_user_deactivated_status(canonical_uid)
+                if deactivated:
+                    raise UserDeactivatedError("This account has been deactivated")
```
clokep, line 325: This doesn't look quite right since it is in the `create_non_existent_users` block.

I'd be curious where we check for deactivation for password users / SSO users and why that isn't being applied properly. Can we consolidate that logic into a single spot?

anoadragon453: Annoyingly the place where we currently check for user deactivation is while authenticating against the local password database: https://github.com/matrix-org/synapse/blob/e24ff8ebe3d4119d377355402245947f7de61c00/synapse/handlers/auth.py#L1379-L1383.

Thankfully we can check if a user is deactivated via the `deactivated` column on the `users` table (rather than using an empty password hash as a heuristic).

So I'd recommend pulling the user deactivated check out of `_check_local_password` and instead moving it into `_validate_userid_login`, such that it can be applied to all user logins; not just those against the local password database.

clokep: Thanks for digging into that @anoadragon453! That sounds sensible to me... I think we did something like this for SSO too at some point.

Looks like we did something specific for SSO too (see #7240).

anoadragon453: Ah, it makes sense that you'd want to show an error page for SSO flows, rather than raising an `UserDeactivatedError`.

Still, the local password and JWT flows could be consolidated.

clokep:
Closing due to lack of response. If you're still interested in working on this please take a look at the review comments above and let us know.

Issue #12124, opened 2022-03-01
HAproxy advised configuration sends IPv4-mapped address on X-Forwarded-For


villepeh:
### Description

I decided to try out HAproxy as Synapse reverse proxy. It looks like the [advised configuration](https://github.com/matrix-org/synapse/blob/develop/docs/reverse_proxy.md#haproxy) sends an incorrect IP with X-Forwarded-For, at least if your server supports both IPv4 and IPv6. It seems to "append" v6 address to your v4 address. I'm assuming this will cause problems with black-/whitelisting and such.

![image](https://user-images.githubusercontent.com/100730729/156256129-245bcf36-e577-471a-8b48-dd9a037297c3.png)

Also visible on HAproxy logs

![image](https://user-images.githubusercontent.com/100730729/156256673-629265fe-9ea2-4561-8420-e11522b8d336.png)

### Steps to reproduce

- Install Synapse
- Install HAproxy 2.5.4
- Follow the reverse proxy instructions [here](https://github.com/matrix-org/synapse/blob/develop/docs/reverse_proxy.md#haproxy)
- Start up Synapse and connect.

Expected behavior: Client and server being reported with a correct IP address, not this "v4 + v6 hybrid".

### Fix

I just followed the instructions here instead: [LINK](https://discourse.haproxy.org/t/make-haproxy-stop-prepending-to-ipv4-addresses-in-x-forwarded-for/3368)

Instead of:
`frontend https`
`  bind :::443 v4v6 ssl crt /etc/ssl/haproxy/ strict-sni alpn h2,http/1.1`
`  [...]`

...use this:
`frontend https`
`bind *:443,:::443 v6only ssl crt /etc/ssl/haproxy/ strict-sni alpn h2,http/1.1`
`[...]`

(and do the same for the federation port if needed)

Reverse proxy now forwards correct IPv4 and IPv6 address to Synapse. 

Perhaps the guide should be updated?

### Version information

- **Homeserver**: Synapse

- **Version**: 1.53.0

- **Install method**: pip (python 3.9)

- **Platform**: Slackware Linux.

(My first post here. Hope I got this right :)

reivilibre:
(I looked at http://cbonte.github.io/haproxy-dconv/configuration-1.6.html#5.1-v6only to see what the options mean.)

[...]

PR #12279, opened 2022-03-23
HAProxy guide update

villepeh:
This PR includes HAProxy reverse proxy guide update. HAProxy will still listen to IPv4 and IPv6 ports but doesn't send IPv4-mapped address to homserver anymore. Also moved forwarding headers to the **backend** part as recommended by several guides.

PR also includes a simple delegation guide.

Signed-off-by: Villepeh


villepeh:
#12124

villepeh:
Changelog filename mess up fixed.

H-Shay:
Looks like the CI has gone wonky-maybe rerun the GHA workflows?

villepeh:
> Looks like the CI has gone wonky-maybe rerun the GHA workflows?

@clokep fixed something that allowed me to merge with the develop branch. Hopefully the PR is fine now.

docs/reverse_proxy.md
```diff
@@ -195,17 +193,23 @@ frontend https
...
-  http-request set-header X-Forwarded-For %[src]
-
+  # Uncomment the line in the end to disable TLS <= v1.2
+  bind *:8448,[::]:8448 ssl crt /path/to/certificate.pem alpn h2,http/1.1 # ssl-min-ver TLSv1.2
+  
   default_backend matrix
 
 backend matrix
-  server matrix 127.0.0.1:8008
+  option forwardfor
```
richvdh, line 202: unfortunately this is not secure, due to https://github.com/matrix-org/synapse/issues/9471 (for which fixes are welcome!) - `option forwardfor` makes haproxy *append* an entry to any existing `X-Forwarded-For` header, whereas synapse uses the *first* entry in the header.

docs/reverse_proxy.md
```diff
@@ -195,17 +193,23 @@ frontend https
...
+  http-request set-header X-Forwarded-Proto https
+  server matrix localhost:8008 # 127.0.0.1:8008 or [::1]:8008 should also work
+```
+[Delegation](delegate.md) example:
+```
+frontend https
+  # Add this anywhere in your frontend config
+  http-request return status 200 content-type application/json lf-string '{"m.server": "matrix.example.com:443"}' if { path -i -m str /.well-known/matrix/server }
+  http-request return status 200 content-type application/json lf-string '{"m.server": {"base_url": "matrix.example.com:443"}}' if { path -i -m str /.well-known/matrix/client }
```
richvdh, line 211: this is not a valid `/.well-known/matrix/client` file.

Also: don't you need to set CORS headers for a client .well-known go work?

docs/reverse_proxy.md
```diff
@@ -195,17 +193,23 @@ frontend https
...
+  # Uncomment the line in the end to disable TLS <= v1.2
+  bind *:8448,[::]:8448 ssl crt /path/to/certificate.pem alpn h2,http/1.1 # ssl-min-ver TLSv1.2
+  
   default_backend matrix
 
 backend matrix
-  server matrix 127.0.0.1:8008
+  option forwardfor
+  http-request set-header X-Forwarded-Proto https
+  server matrix localhost:8008 # 127.0.0.1:8008 or [::1]:8008 should also work
```
richvdh, line 204: this isn't very clear. Why are we giving multiple options here? Which is the user supposed to do? (What was wrong with the original?)

docs/reverse_proxy.md
```diff
@@ -182,11 +182,9 @@ matrix.example.com {
 
 ```
 frontend https
-  bind :::443 v4v6 ssl crt /etc/ssl/haproxy/ strict-sni alpn h2,http/1.1
-  http-request set-header X-Forwarded-Proto https if { ssl_fc }
-  http-request set-header X-Forwarded-Proto http if !{ ssl_fc }
-  http-request set-header X-Forwarded-For %[src]
-
+  # Uncomment the line in the end to disable TLS <= v1.2
```
richvdh, line 185: again, I don't think this is the place to be giving the user lots of different options. Let's make a single recommendation. If advanced users want to do things like restrict TLS versions they can research how to do it themselves.

docs/reverse_proxy.md
```diff
@@ -182,11 +182,9 @@ matrix.example.com {
 
 ```
 frontend https
-  bind :::443 v4v6 ssl crt /etc/ssl/haproxy/ strict-sni alpn h2,http/1.1
```
richvdh, line 185: the use of a directory here was deliberate - could you explain why changing to a single file is prefererable?

richvdh (changes requested):
thanks for this! I've got a few questions and comments though.

villepeh:
I think I got a bit too carried away with "what works for me" so I limited the changes to #12124 for now.

> unfortunately this is not secure, due to https://github.com/matrix-org/synapse/issues/9471 (for which fixes are welcome!) - option forwardfor makes haproxy append an entry to any existing X-Forwarded-For header, whereas synapse uses the first entry in the header.

Thanks for letting me know. This can be solved by adding "http-request del-header x-forwarded-for" before `option forwardfor`. But we might as well keep the original setting since it's just one line

> this is not a valid /.well-known/matrix/client file.
> Also: don't you need to set CORS headers for a client .well-known to work?

Thanks, I fixed mine :) I need to study more about CORS headers before I advising other server admins. I reverted this one.

> this isn't very clear. Why are we giving multiple options here? Which is the user supposed to do? (What was wrong with the original?)

I prefer local IPv6 traffic, because... uum... the future? 🙈  anyway, localhost is in half of the tutorials and 127.0.0.1 is in the other half so I kept that one here so it'll redirect traffic to whichever is listening.

> again, I don't think this is the place to be giving the user lots of different options. Let's make a single recommendation. If advanced users want to do things like restrict TLS versions they can research how to do it themselves.

Got carried away again. Sorry about hat

> the use of a directory here was deliberate - could you explain why changing to a single file is prefererable?

My distro doesn't have that folder so I thought it was just an example. I reverted it back. I created a _haproxy.key_ file that has cert.pem and fullchain.pem cat'd together. I reverted this one.

villepeh:
Localhost can actually cause issues in some scenarios: [resolver defaults to ipv6 even if local ipv6 not available in network stack](https://github.com/haproxy/haproxy/issues/508). Going back to 127.0.0.1.

changelog.d/12279.doc
```diff
@@ -0,0 +1 @@
+HAProxy reverse proxy guide update to stop sending IPv4-mapped address to homeserver.
```
richvdh, line 1:
```suggestion
HAProxy reverse proxy guide update to stop sending IPv4-mapped address to homeserver. Contributed by @villepeh.
```

richvdh (approved):
nice and simple now! :heart:

richvdh:
> Signed-off-by: Villepeh

Just to make you aware, we normally need a legally identifiable name here (see https://matrix-org.github.io/synapse/latest/development/contributing_guide.html#sign-off). Since this is a small change to the docs, I don't think it's too important for this PR, but if you make more contributions in future we'll need a legal name.

PR #12283, opened 2022-03-24
Re-enable concurrency on Complement

kegsay:
The issues which meant concurrency caused flakey tests has been resolved for over a month now without issue ( https://github.com/matrix-org/complement/pull/318 ) so it's safe to re-enable this for Synapse.

For context, this reduces the total time from ~9m 52s to ~8m 11s. Most of the savings are during the actual running of the tests, which has gone from 7m 20s to 5m 47s.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -377,7 +377,7 @@ jobs:
       # Run Complement
       - run: |
           set -o pipefail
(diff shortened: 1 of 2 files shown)


clokep (approved):
CI seems happy, matches what's in the Complement CI.

PR #12284, opened 2022-03-24
Add a configuration option for rewriting base URLs when interacting with ISs

babolivier:
This work is happening in the context of the Tchap mainlining.

Identity servers are services that need to be accessible by both clients (lookups, associations, etc) and homeservers (storing 3PID invites). In the Tchap infra, for security reasons, any IS URL that doesn't need to be accessible to clients are firewalled off and only accessible via the internal network. Thus the URL (both address/hostname and protocol) to use isn't the same depending on whether the communication is happening from a client on the internet or a homeserver on the local network.

This change introduces a `rewrite_identity_server_base_urls` configuration setting allowing admins to provide a substitution maps for base URLs when to use by Synapse when interacting with identity services.

### DIFF
--- a/changelog.d/12284.feature
+++ b/changelog.d/12284.feature
@@ -1,0 +1,1 @@
+Add a configuration option for rewriting base URLs when interacting with identity servers from Synapse.
--- a/docs/sample_config.yaml
+++ b/docs/sample_config.yaml
@@ -539,6 +539,23 @@ templates:
   #
   #custom_template_directory: /path/to/custom/templates/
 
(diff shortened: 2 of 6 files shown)


DMRobertson:
This feels like a pretty bespoke option and I'm not sure that Synapse is the right place for this. Could this be better done by some other means of configuring TChap's internal systems? E.g. DNS CNAME records/iptables rules?

babolivier:
Yep, that could make sense. I'll have a chat with @MatMaul to ask what he thinks about this.

richvdh:
taking this out the review queue for now then.

PR #12288, opened 2022-03-24
Mention the new behaviour on unsafe database locale in the upgrade notes

babolivier:
Follow-up to https://github.com/matrix-org/synapse/pull/12262

If for some reason a homeserver's database isn't on the right locale, it's very scary if Synapse suddenly refuses to start after an update and tells the admin they need to recreate their database, without a decent heads up. This PR adds such heads up to the upgrade notes.

### DIFF
--- a/docs/upgrade.md
+++ b/docs/upgrade.md
@@ -104,7 +104,7 @@ experimental_features:
 Synapse now refuses to start when using PostgreSQL with incorrect values of `COLLATE` and
 `CTYPE` unless the config flag `allow_unsafe_locale`, found in the database section of
 the configuration file, is set to `true`. See the [PostgreSQL documentation](https://matrix-org.github.io/synapse/latest/postgres.html)
-for more information.
+for more information and instructions on how to fix a database with incorrect values.
 
 # Upgrading to v1.55.0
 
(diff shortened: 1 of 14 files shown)


clokep:
Can we point people to a link or anything about how to fix this / check this beforehand?

babolivier:
I don't think we have documentation on how to check this, but I've linked to the postgres docs which has steps to fix this longer term.

docs/upgrade.md
```diff
@@ -99,6 +99,13 @@ experimental_features:
   groups_enabled: false
 ```
 
+## Change in behaviour for PostgreSQL databases with unsafe locale
+
+Synapse now refuses to start when using PostgreSQL with incorrect values of `COLLATE` and
```
squahtx, line 104: With my pretend synapse admin hat on, it's not immediately clear what a correct value of `COLLATE` and `CTYPE` is. It's not mentioned in https://matrix-org.github.io/synapse/latest/postgres.html#fixing-incorrect-collate-or-ctype either. I'm guessing it's "C"?

H-Shay: Yes, this change applies to any non-`C` locale

H-Shay, line 104:
```suggestion
Synapse now refuses to start when using PostgreSQL with non-`C` values for `COLLATE` and
```
Maybe this is clearer? Thanks for doing this btw, I didn't think to do so!

babolivier: No worries :)

squahtx (approved):
lgtm!

PR #12289, opened 2022-03-24
Replace instances of deprecated `Jinja2.Markup` with `markupsafe.Markup`

anoadragon453:
The Jinja2 recently [removed](https://jinja.palletsprojects.com/en/3.1.x/changes/#version-3-1-0) the long-deprecated `Jinja2.Markup` property, instead advising callers to import `Markup` from the [markupsafe](https://pypi.org/project/MarkupSafe/) library directly.

This PR:

* Adds a direct dependency on MarkupSafe as we now import from it. Pinned to the currently major version of `2`.
* Replaces uses of `Jinja2.Markup` with `markupsafe.Markup`

### DIFF
--- a/synapse/python_dependencies.py
+++ b/synapse/python_dependencies.py
@@ -75,6 +75,7 @@ REQUIREMENTS = [
     "attrs>=19.2.0,!=21.1.0",
     "netaddr>=0.7.18",
     "Jinja2>=2.9",
+    "MarkupSafe<=3.0.0",
     "bleach>=1.4.3",
     # We use `ParamSpec`, which was added in `typing-extensions` 3.10.0.0.
     "typing-extensions>=3.10.0",
--- a/synapse/python_dependencies.py
+++ b/synapse/python_dependencies.py
@@ -75,7 +75,7 @@ REQUIREMENTS = [
     "attrs>=19.2.0,!=21.1.0",
     "netaddr>=0.7.18",
     "Jinja2>=2.9",
-    "MarkupSafe<=3.0.0",
+    "MarkupSafe>=2.0",
     "bleach>=1.4.3",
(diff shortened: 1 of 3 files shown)


synapse/python_dependencies.py
```diff
@@ -75,6 +75,7 @@
     "attrs>=19.2.0,!=21.1.0",
     "netaddr>=0.7.18",
     "Jinja2>=2.9",
+    "MarkupSafe<=3.0.0",
```
clokep, line 78: I'm curious where the 3.0.0 version came from (and why less than, instead of greater than some package number).

clokep: Where did the 2.1.1 come from though? Is that what Jinja2 was pinning at?

anoadragon453: That's just the latest version of `MarkupSafe`. Jinja2 uses `MarkupSafe>=2.0`: https://github.com/pallets/jinja/blob/e542e100027aaa131df73ad06b71ba9a4ee348c8/setup.py#L6

clokep:
Also, do we want to target release or develop?

anoadragon453:
> Also, do we want to target release or develop?

Ah! Good catch. This is intended to be 1.55.1, so it should be against `release-v1.55`.

anoadragon453:
(force pushed to rebase the existing commits on top of the `release-v1.55` branch.

synapse/python_dependencies.py
richvdh: for reference, if your goal is to pin to `2.x`, you probably want `~=2.0`. See https://peps.python.org/pep-0440/#compatible-release.

clokep: The goal from my POV was to not change what we were already pulling in, but making it an explicit dependency.

PR #12291, opened 2022-03-24
Remove unused test code.

clokep:
This code is not imported anywhere, pretty sure it can go. 🔥


DMRobertson (approved):
AFAICS the last time this was used was [about a year ago](https://github.com/matrix-org/synapse/pull/9396/files#diff-4755362f1ed91daea3a616c198f24f0f44a54f8910e3a40f24ef207c26572de5L44). Kill it with fire.

PR #12293, opened 2022-03-24
Remove the unused and unstable `/aggregations` endpoint.

clokep:
As background: an old version of MSC2675 specified an `/aggregations` endpoint in addition to the `/relations` endpoint. This endpoint allowed paginating through the bundled aggregations of `m.annotation` relations.

This PR makes the following modifications:

* It removes the unstable, unspecced (and not even part of a current MSC!) endpoint for `/aggregations`.
* It then cleans up the code to fetch the bundled aggregations for `m.annotation` to remove the pagination tokens which were only useful for the `/aggregations` endpoint. (Note that the bundled aggregation form of `m.annotation` is defined in MSC2677, but it is rather divorced from current reality.)

The `/aggregations` endpoint is unused, no clients implement it and matrix.org has no hits in the logs. Additionally, no known clients use the bundled aggregations for `m.annotation`.

Kind of part of #11753 / kind of blocking #12235 (due to needing to take into account this code).


synapse/handlers/relations.py
```diff
@@ -190,13 +190,11 @@ async def _get_bundled_aggregation_for_event(
...
-        annotations = await self._main_store.get_aggregation_groups_for_event(
+        annotations, limited = await self._main_store.get_aggregation_groups_for_event(
             event_id, room_id
         )
-        if annotations.chunk:
-            aggregations.annotations = await annotations.to_dict(
-                cast("DataStore", self)
-            )
+        if annotations:
+            aggregations.annotations = {"chunk": annotations, "limited": limited}
```
clokep, line 197: Previously this would have given pagination tokens instead of `limited`. I chose `limited` instead since I would think that a client wants *some* way to know if it got all the results or not. If it hasn't, then it should go check `/relations`. 🤷 (`limited` also now matches what the MSC says...)

clokep:
Note that this PR is based on a conversation I had with @turt2live about what to do with this code. We came to the agreement that it is in a really weird spot as unspecced/unstable, but also not part of any current MSC and should just go. If a future MSC wants to add it back, we can pull it back out of git history fairly easily.

synapse/storage/databases/main/relations.py
```diff
@@ -270,79 +261,38 @@ async def get_aggregation_groups_for_event(
...
-            SELECT type, aggregation_key, COUNT(DISTINCT sender), MAX(stream_ordering)
+            SELECT type, aggregation_key, COUNT(DISTINCT sender)
             FROM event_relations
             INNER JOIN events USING (event_id)
-            WHERE {where_clause}
+            WHERE relates_to_id = ? AND room_id = ? AND relation_type = ?
             GROUP BY relation_type, type, aggregation_key
-            {having_clause}
-            ORDER BY COUNT(*) {order}, MAX(stream_ordering) {order}
+            ORDER BY COUNT(*) DESC, MAX(stream_ordering) DESC
```
clokep, line 283: I'm not 100% sure that these order by queries still make sense or not? We could possible remove `MAX(stream_ordering)`?

synapse/handlers/relations.py
richvdh: I'm not really following this. Are we no longer allowing pagination? Typically the absence of `next_batch`implies that we've reached the end of the results (see https://spec.matrix.org/v1.2/appendices/#pagination)

Which MSC says we should return `limited` ?

Isn't changing the format of the response to `/relations` orthogonal to removing `/aggregations` ?

synapse/storage/databases/main/relations.py
richvdh: yes please

synapse/storage/databases/main/relations.py
```diff
@@ -270,79 +261,38 @@ async def get_aggregation_groups_for_event(
...
-            event_type: Only fetch events with this event type, if given.
             limit: Only fetch the `limit` groups.
-            direction: Whether to fetch the highest count first (`"b"`) or
-                the lowest count first (`"f"`).
-            from_token: Fetch rows from the given token, or from the start if None.
-            to_token: Fetch rows up to the given token, or up to the end if None.
 
         Returns:
             List of groups of annotations that match. Each row is a dict with
             `type`, `key` and `count` fields.
```
richvdh, line 268: this is no longer accurate

richvdh (commented):
:tada: for removing code. I have questions though.

synapse/handlers/relations.py
clokep: > I'm not really following this. Are we no longer allowing pagination? Typically the absence of `next_batch`implies that we've reached the end of the results (see https://spec.matrix.org/v1.2/appendices/#pagination)

Well `/aggregations` isn't really part of any MSC right now. It was removed from the (merged) MSC2675. MSC2677 defines the bundled aggregation format for `m.annotations`, but is mostly lies. The changes in here make it more closely follow MSC2677.

> Which MSC says we should return `limited` ?

MSC2677, see https://github.com/uhoreg/matrix-doc/blob/aggregations-reactions/proposals/2677-reactions.md#bundled-relations

It still wouldn't match perfectly though since we would not include the `count` key.

> Isn't changing the format of the response to `/relations` orthogonal to removing `/aggregations` ?

This doesn't change the response of `/relations`, it changes the bundled aggregations format. The idea is that you can still use `/relations` and aggregate locally even without `/aggregations` though. So the information isn't lost, just a bit harder to get to.

Let me know if that makes sense, or we can discuss offline if that's ab it easier!

clokep:
There was some chatter about this near https://matrix.to/#/!XaqDhxuTIlvldquJaV:matrix.org/$poGNZCThWxibXwbicuCh1J3IRDaIc5Brz1npmIA2u2c?via=matrix.org&via=element.io&via=vector.modular.im

synapse/handlers/relations.py
clokep: Based on our conversation in #synapse-dev:matrix.org I've removed this flag.

clokep:
@richvdh I've updated this PR description based on our conversation and removed the `limited` field from the response. I think it is now up to MSC2677 to define a reasonable format for this when someone is ready to push that through standardization.

clokep:
I seem to have been wrong that there's no MSC for this any longer -- it seems this got moved into [MSC3571](https://github.com/matrix-org/matrix-spec-proposals/pull/3571).

I still think the right thing is to remove this until it is needed and properly defined, however.

PR #12295, opened 2022-03-24
Move MSC2654 support behind an experimental configuration flag.

clokep:
Note sure if this is something we really want to do, but it seems a bit weird to me that we're returning an unstable field by default which is [something we've said we don't want to do](https://matrix-org.github.io/synapse/latest/development/experimental_features.html). 🤷 

Anyway, this moves the support for MSC2654 behind an experimental configuration flag.


squahtx (approved):
lgtm!

Off the top of your head, are there any clients using this field at the moment?

clokep:
From my quick looking across GitHub: https://github.com/search?p=4&q=org.matrix.msc2654.unread_count&type=Code I don't see other clients really using it. There's a few WIP SDKs which use it, but most of them seem to properly handle it not existing.

Issue #12294, opened 2022-03-24
`ImportError` after upgrade to Synapse 1.55.{1,2}


richvdh:
I upgraded Synapse to 1.55.1, and got this error:

```
Mar 24 18:04:24 river synapse[16993]: Traceback (most recent call last):
Mar 24 18:04:24 river synapse[16993]:   File "/usr/lib/python3.7/runpy.py", line 193, in _run_module_as_main
Mar 24 18:04:24 river synapse[16993]:     "__main__", mod_spec)
Mar 24 18:04:24 river synapse[16993]:   File "/usr/lib/python3.7/runpy.py", line 85, in _run_code
Mar 24 18:04:24 river synapse[16993]:     exec(code, run_globals)
Mar 24 18:04:24 river synapse[16993]:   File "/home/synapse/src/synapse/app/homeserver.py", line 28, in <module>
Mar 24 18:04:24 river synapse[16993]:     import synapse.config.logger
Mar 24 18:04:24 river synapse[16993]:   File "/home/synapse/src/synapse/config/__init__.py", line 15, in <module>
Mar 24 18:04:24 river synapse[16993]:     from ._base import ConfigError, find_config_files
Mar 24 18:04:24 river synapse[16993]:   File "/home/synapse/src/synapse/config/_base.py", line 37, in <module>
Mar 24 18:04:24 river synapse[16993]:     import jinja2
Mar 24 18:04:24 river synapse[16993]:   File "/home/synapse/env/lib/python3.7/site-packages/jinja2/__init__.py", line 12, in <module>
Mar 24 18:04:24 river synapse[16993]:     from .environment import Environment
Mar 24 18:04:24 river synapse[16993]:   File "/home/synapse/env/lib/python3.7/site-packages/jinja2/environment.py", line 25, in <module>
Mar 24 18:04:24 river synapse[16993]:     from .defaults import BLOCK_END_STRING
Mar 24 18:04:24 river synapse[16993]:   File "/home/synapse/env/lib/python3.7/site-packages/jinja2/defaults.py", line 3, in <module>
Mar 24 18:04:24 river synapse[16993]:     from .filters import FILTERS as DEFAULT_FILTERS  # noqa: F401
Mar 24 18:04:24 river synapse[16993]:   File "/home/synapse/env/lib/python3.7/site-packages/jinja2/filters.py", line 13, in <module>
Mar 24 18:04:24 river synapse[16993]:     from markupsafe import soft_unicode
[...]

PR #12297, opened 2022-03-24
Pin Jinja to <3.1.0

anoadragon453:
Pins [Jinja2](https://pypi.org/project/jinja2/) to <3.1.0, as 3.1.0 removed the deprecated `jinja2.Markup` class which we still rely on.

An alternative to https://github.com/matrix-org/synapse/pull/12289, which ended in https://github.com/matrix-org/synapse/issues/12294.


richvdh (approved):
I think so.

PR #12302, opened 2022-03-25
Add a callback to react to 3PID associations

babolivier:
Happening in the context of the Tchap mainlining, to enable https://github.com/matrix-org/synapse-bind-sydent

A first draft of that module uses `on_user_registration`, however at the time it's called newly registered users don't have their 3PIDs stored in the database yet. Using this callback will slightly change the module's behaviour (in that it won't only be triggered on registrations) but that should be fine.

### DIFF
--- a/docs/modules/third_party_rules_callbacks.md
+++ b/docs/modules/third_party_rules_callbacks.md
@@ -247,6 +247,24 @@ admin API.
 
 If multiple modules implement this callback, Synapse runs them all in order.
 
+### `on_threepid_bind`
+
+_First introduced in Synapse v1.56.0_
+
+```python
+async def on_threepid_bind(user_id: str, medium: str, address: str) -> None:
+```
+
+Called after creating an association between a local user and a third-party identifier
+(email address, phone number). The module is given the Matrix ID of the user the
+association is for, as well as the medium (`email` or `msisdn`) and address of the
+third-party identifier.
+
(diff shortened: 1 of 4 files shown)


docs/modules/third_party_rules_callbacks.md
```diff
@@ -247,6 +247,24 @@ admin API.
...
 
+### `on_threepid_bind`
+
+_First introduced in Synapse v1.56.0_
+
+```python
+async def on_threepid_bind(user_id: str, medium: str, address: str) -> None:
+```
+
+Called after creating an association between a local user and a third-party identifier
```
H-Shay, line 258: Maybe just clarify that the association here is on the homeserver?

babolivier: It does say below:

> Note that this callback is _not_ called after a successful association on an _identity
> server_.

Which I think is enough?

H-Shay: I saw that and my only thought was that the below statement makes one draw an inference rather than explicitly stating that the association in question is on the homeserver but it's a very subtle point!

babolivier: I see what you mean. Considering the only alternative to the association happening on the IS is that it's happening on the HS, I think that should be good enough. We can always revisit it later if we find that it confuses people though.

PR #12310, opened 2022-03-28
Add a configuration to exclude rooms from sync response

babolivier:
Part of the Tchap mainlining.

This is in order to implement the new plan for user discovery. See https://github.com/matrix-org/synapse-dinsic/issues/125 for the high-level description and https://docs.google.com/document/d/1wQw1OMgqn0blbv4Jt7ApZhR8mtN39ZLkNHKuNY-Xd_M/edit for the detailed action plan (which this work is a part of).

### DIFF
--- a/synapse/config/server.py
+++ b/synapse/config/server.py
@@ -1240,7 +1240,7 @@ class ServerConfig(Config):
           #custom_template_directory: /path/to/custom/templates/
 
         # List of rooms to exclude from sync responses.
-        # No room is excluded by default.
+        # By default, no room is excluded.
         #
         #exclude_rooms_from_sync:
         #    - !foo:example.com
--- a/synapse/config/server.py
+++ b/synapse/config/server.py
@@ -680,6 +680,10 @@ class ServerConfig(Config):
             config.get("use_account_validity_in_account_status") or False
         )
 
+        self.rooms_to_exclude_from_sync: List[str] = (
+            config.get("exclude_rooms_from_sync") or []
+        )
+
     def has_tls_listener(self) -> bool:
         return any(listener.tls for listener in self.listeners)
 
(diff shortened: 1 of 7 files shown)


synapse/config/server.py
```diff
@@ -1234,6 +1238,12 @@ def generate_config_section(
           # information about using custom templates.
           #
           #custom_template_directory: /path/to/custom/templates/
+
+        # List of rooms to exclude from sync responses.
```
H-Shay, line 1242: Can some info be added here about the use case for excluding rooms and the end result (i.e. that end-users won't see these rooms in their clients, etc?)

synapse/storage/databases/main/roommember.py
```diff
@@ -384,14 +388,19 @@ async def get_rooms_for_local_user_where_membership_is(
             self._get_rooms_for_local_user_where_membership_is_txn,
             user_id,
             membership_list,
+            excluded_rooms,
         )
 
         # Now we filter out forgotten rooms
         forgotten_rooms = await self.get_forgotten_rooms_for_user(user_id)
         return [room for room in rooms if room.room_id not in forgotten_rooms]
```
H-Shay, line 396: Out of curiosity, is there any reason that the excluded rooms couldn't be filtered out here, like the forgotten rooms are, rather than changing the query to the database in `_get_rooms_for_local_user_where_membership_is_txn`?

babolivier: My rationale here is that we already know from the start which rooms we want to exclude, so ignoring them from the result would minimise the work and keep the memory usage to a minimum. I'm not sure why we're not already doing that for forgotten rooms, I'll ask around.

babolivier: According to #synapse-dev, it's likely a combination of `NOT IN` not being very well supported at the time and the forgotten rooms list being user-defined and not capped thus it's possible it becomes large enough to cause issues if used directly in the query. I think we're good on both points with the excluded rooms, since it's an admin-defined static list that should remain fairly small, but I'll still change it just to simplify the query and the code (and it shouldn't make a huge difference).

babolivier:
CI failure is due to https://github.com/matrix-org/synapse/issues/12324

H-Shay (approved):
LGTM

anoadragon453:
Passing note: these days this would probably be better implemented with room types, which clients could hide those that they didn't understand.

PR #12311, opened 2022-03-28
Improve type annotations for `execute_values`.

reivilibre:
Should be a commit-by-commit review.

This was something that I agreed to look into whilst I was in the area.
Turns out the `*args` was not necessary; there's only one arg here (an iterable of value rows).

I also added a comment to `_do_execute`, where we could improve the type annotations easily a little bit
once `ParamSpec` is more usable.

### DIFF
--- a/synapse/storage/database.py
+++ b/synapse/storage/database.py
@@ -324,16 +324,6 @@ class LoggingTransaction:
         "Strip newlines out of SQL so that the loggers in the DB are on one line"
         return " ".join(line.strip() for line in sql.splitlines() if line.strip())
 
-    # TODO(ParamSpec):
-    #     Once Mypy supports Concatenate, this could be written as
-    #     def _do_execute(
-    #         self,
-    #         func: Callable[Concatenate[str, P], R],
-    #         sql: str,
-    #         *args: P.args,
-    #         **kwargs: P.kwargs,
-    #     ) -> R:
-    #     so long as we also pass kwargs.
     def _do_execute(self, func: Callable[..., R], sql: str, *args: Any) -> R:
         sql = self._make_sql_one_line(sql)
 
--- a/synapse/storage/database.py
+++ b/synapse/storage/database.py
@@ -290,7 +290,9 @@ class LoggingTransaction:
         if isinstance(self.database_engine, PostgresEngine):
             from psycopg2.extras import execute_batch
 
-            self._do_execute(lambda *x: execute_batch(self.txn, *x), sql, args)
+            self._do_execute(
+                lambda the_sql: execute_batch(self.txn, the_sql, args), sql
+            )
         else:
             self.executemany(sql, args)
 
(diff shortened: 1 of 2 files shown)


synapse/storage/database.py
```diff
@@ -326,6 +324,16 @@ def _make_sql_one_line(self, sql: str) -> str:
...
+    # TODO(ParamSpec):
+    #     Once Mypy supports Concatenate, this could be written as
+    #     def _do_execute(
+    #         self,
+    #         func: Callable[Concatenate[str, P], R],
+    #         sql: str,
+    #         *args: P.args,
+    #         **kwargs: P.kwargs,
+    #     ) -> R:
+    #     so long as we also pass kwargs.
```
clokep, line 336: Seems better as an issue that we file?

synapse/storage/database.py
```diff
@@ -305,15 +307,11 @@ def execute_values(self, sql: str, *args: Any, fetch: bool = True) -> List[Tuple
...
         return self._do_execute(
-            # Type ignore: mypy is unhappy because if `x` is a 5-tuple, then there will
-            # be two values for `fetch`: one given positionally, and another given
-            # as a keyword argument. We might be able to fix this by
-            # - propagating the signature of psycopg2.extras.execute_values to this
-            #   function, or
-            # - changing `*args: Any` to `values: T` for some appropriate T.
-            lambda *x: execute_values(self.txn, *x, fetch=fetch),  # type: ignore[misc]
+            lambda the_sql, the_values: execute_values(
+                self.txn, the_sql, the_values, fetch=fetch
```
clokep, line 311: (This code was pre-existing, but...)

`the_sql` doesn't change for each iterate, correct? So really no reason to pass it in for each one?

reivilibre: Sorry, what do you mean?

`the_sql` is given by `_do_execute`.

`the_values` are just a pass-through of `values` and could have been part of the lambda's closure, but in analogue to `execute_batch` (and indeed the way this code was written before), it's passed through. If we want to change this I suppose I have no objections, though I wonder if the idea was that `_do_execute` could have transformed the args if necessary...?

synapse/storage/database.py
reivilibre: yeah, if you think so, could do that instead.

synapse/storage/database.py
clokep: Errr, yes...that didn't make sense.

My confusion is -- why are we passing in `fetch` directly into the lambda, but we pass in `the_sql` and `the_values` as arguments to the lambda?

clokep (approved):
Seems fine, there's a pending question but won't block merging this on it!

synapse/storage/database.py
clokep: Ah, I did not realize that `_do_execute` modified the passed in `sql`!

reivilibre: I've changed this around a bit, is that better to you?

clokep: I do find that clearer, yes!

PR #12313, opened 2022-03-28
Un-revert Jinja2 fix

babolivier:
* Reverts https://github.com/matrix-org/synapse/pull/12296
* Pins `Jinja2>=3.0`

Fixes https://github.com/matrix-org/synapse/issues/12294

### DIFF
--- a/changelog.d/12313.misc
+++ b/changelog.d/12313.misc
@@ -1,0 +1,1 @@
+Correctly import `Markup` from `MarkupSafe` instead of `Jinja2`.
--- a/changelog.d/12313.misc
+++ b/changelog.d/12313.misc
@@ -1,1 +1,1 @@
-Import `Markup` from `MarkupSafe` as importing it from `Jinja2` has been deprecated for some time and recently removed.
+Fix compatibility with the recently-released Jinja 3.1.
--- a/changelog.d/12313.misc
+++ b/changelog.d/12313.misc
@@ -1,1 +1,1 @@
-Correctly import `Markup` from `MarkupSafe` instead of `Jinja2`.
+Import `Markup` from `MarkupSafe` as importing it from `Jinja2` has been deprecated for some time and recently removed.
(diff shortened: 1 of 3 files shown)


changelog.d/12313.misc
```diff
@@ -0,0 +1 @@
+Correctly import `Markup` from `MarkupSafe` instead of `Jinja2`.
```
richvdh, line 1: could you expand on this? Why are we doing this?

synapse/python_dependencies.py
```diff
@@ -74,8 +74,7 @@
     # Note: 21.1.0 broke `/sync`, see #9936
     "attrs>=19.2.0,!=21.1.0",
     "netaddr>=0.7.18",
-    # Jinja2 3.1.0 removes the deprecated jinja2.Markup class, which we rely on.
-    "Jinja2<3.1.0",
+    "Jinja2~=3.0",
```
richvdh, line 77: why do we need 3.x? a comment would be good.

changelog.d/12313.misc
```diff
@@ -0,0 +1 @@
+Import `Markup` from `MarkupSafe` as importing it from `Jinja2` has been deprecated for some time and recently removed.
```
richvdh, line 1: How about simply:

```suggestion
Fix compatibility with the recently-released Jinja 3.1?
```

synapse/python_dependencies.py
```diff
@@ -74,8 +74,8 @@
     # Note: 21.1.0 broke `/sync`, see #9936
     "attrs>=19.2.0,!=21.1.0",
     "netaddr>=0.7.18",
-    # Jinja2 3.1.0 removes the deprecated jinja2.Markup class, which we rely on.
-    "Jinja2<3.1.0",
+    # Jinja 2.x is incompatible with recent versions of MarkupSafe, on which it depends.
+    "Jinja2~=3.0",
```
richvdh, line 78: The `olddeps` logic ([here](https://github.com/matrix-org/synapse/blob/develop/tox.ini#L125)) doesn't handle `~=`, so you'll either need to fix that, or change this to a simple `>=`.

babolivier:
```suggestion
    "Jinja2>=3.0",
```

richvdh (approved):
thanks!

PR #12314, opened 2022-03-28
Exclude outliers in `on_backfill_request`

richvdh:
When we are processing a `/backfill` request from a remote server, exclude any
outliers from consideration early on. We can't return outliers anyway (since we
don't know the state at the outlier), and filtering them out earlier means that
we won't attempt to calculate the state for them.

part of my work on fixing #12074.

### DIFF
--- a/synapse/storage/databases/main/event_federation.py
+++ b/synapse/storage/databases/main/event_federation.py
@@ -1073,9 +1073,15 @@ class EventFederationWorkerStore(SignatureWorkerStore, EventsWorkerStore, SQLBas
             /* Get the depth and stream_ordering of the prev_event_id from the events table */
             INNER JOIN events
             ON prev_event_id = events.event_id
+
+            /* exclude outliers from the results (we don't have the state, so cannot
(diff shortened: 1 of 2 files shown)


synapse/storage/databases/main/event_federation.py
```diff
@@ -1073,9 +1073,15 @@ def _get_connected_prev_event_backfill_results_txn(
...
+
+            /* exclude outliers from the results (we don't have the state, so cannot
+             * verify if the requesting server can see them).
+             */
+            WHERE NOT events.outlier
+            
             /* Look for an edge which matches the given event_id */
-            WHERE event_edges.event_id = ?
-            AND event_edges.is_state = ?
+            AND event_edges.event_id = ? AND NOT event_edges.is_state
```
richvdh, line 1083: this is a quick simplification while I'm passing.

H-Shay (approved):
Makes sense to me!

richvdh:
thanks!

Issue #12201, opened 2022-03-10
Sometimes we have entries in `event_to_state_groups` for outliers, and I don't know why


richvdh:
Generally we don't expect to have the state at an outlier event, so having entries in `event_to_state_groups` is surprising to me. I haven't investigated why this happens, or if it causes any real problems.

MadLittleMods:
This seems very plausible (by the name of the function): We have various functions which run into `_auth_and_persist_outliers` and makes it way down to `_store_event_state_mappings_txn` which upserts into `event_to_state_groups`.

```
_auth_and_persist_outliers
_auth_and_persist_outliers_inner
persist_events_and_notify
persist_events
enqueue
_event_persist_queue -> _persist_event_batch
_persist_events_and_state_updates
_persist_events_txn
_store_event_state_mappings_txn (upserts `event_to_state_groups`)
```

richvdh:
but `_store_event_state_mappings_txn` shouldn't do anything for outliers. It contains the following:

```
            if event.internal_metadata.is_outlier():
                # ...
                continue
```

MadLittleMods:
@richvdh Good call 🤦‍♂️

What is the date of the most recent entry in `event_to_state_groups` that is an `outlier`? (share your database queries if you look it up) I'm wondering if it's still an ongoing problem of accidentally adding outliers or possibly a previous bug that is now fixed.

richvdh:
I've not been able to find any examples on matrix.org, but I don't know if that means they don't exist, or just that a couple of hours of table-scanning failed find them.

I did find a number of examples on sw1v.org, with this query:

```
select to_timestamp(received_ts/1000), * from events e join event_to_state_groups esg using (event_id) where e.outlier order by stream_ordering desc limit 10;
```

The most recent examples there are from 2021-12-02, so it may have been fixed since.

Some other things I noticed about these entries:
 * they all have positive `stream_ordering`s
 * they *also* have entries in `ex_outlier_stream`, with *negative* stream orderings.

That means they were originally persisted with positive stream-orderings, and we later attempted to de-outlier them with a negative stream-ordering. There's nothing particularly wrong with that; what is odd is that they are still marked as outliers after going through the de-outliering process, given that we always set `outlier=False` whenever we add an entry to `ex_outlier_stream`: https://github.com/matrix-org/synapse/blob/release-v1.53/synapse/storage/databases/main/events.py#L1320-L1332.

[...]

PR #12316, opened 2022-03-28
Room batch: fix up handling of unknown prev_event_ids

richvdh:
Don't attempt to get the state at an unknown `prev_event_id`.

part of my work on fixing #12074.

### DIFF
--- a/synapse/rest/client/room_batch.py
+++ b/synapse/rest/client/room_batch.py
@@ -123,6 +123,19 @@ class RoomBatchSendEventRestServlet(RestServlet):
                     errcode=Codes.INVALID_PARAM,
                 )
 
+        # Make sure that the prev_event_ids exist and aren't outliers - ie, they are
+        # regular parts of the room DAG where we know the state.
+        non_outlier_prev_events = await self.store.have_events_in_timeline(
+            prev_event_ids_from_query
+        )
+        for prev_event_id in prev_event_ids_from_query:
+            if prev_event_id not in non_outlier_prev_events:
+                raise SynapseError(
+                    HTTPStatus.BAD_REQUEST,
+                    "prev_event %s does not exist, or is an outlier" % (prev_event_id,),
+                    errcode=Codes.INVALID_PARAM,
+                )
+
         # For the event we are inserting next to (`prev_event_ids_from_query`),
         # find the most recent state events that allowed that message to be
         # sent. We will use that as a base to auth our historical messages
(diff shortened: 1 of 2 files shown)


squahtx (approved):
lgtm

more of an msc2716 question: What is `prev_event_id` supposed to be when a `batch_id` is provided? If it's the insertion event associated with `batch_id`, then surely that's redundant?

synapse/rest/client/room_batch.py
```diff
@@ -123,6 +123,19 @@ async def on_POST(
                     errcode=Codes.INVALID_PARAM,
```
MadLittleMods, line 123: > more of an [msc2716](https://github.com/matrix-org/matrix-doc/pull/2716) question: What is `prev_event_id` supposed to be when a `batch_id` is provided? If it's the insertion event associated with `batch_id`, then surely that's redundant?
>
> *-- https://github.com/matrix-org/synapse/pull/12316#pullrequestreview-924611628*

It's used as the place we're inserting next to. Which means, inheriting the `depth` and state at that point in time.

Batches of history aren't connected together in the DAG by `prev_event_id`. There is only a loose connection from the `batch` event of the next batch pointing back at the `insertion` event in the previous batch. But it's also possible to derive the same information from the previous batch but it just hasn't been that way. It can change.

synapse/rest/client/room_batch.py
```diff
@@ -123,6 +123,19 @@ async def on_POST(
...
+        non_outlier_prev_events = await self.store.have_events_in_timeline(
+            prev_event_ids_from_query
+        )
+        for prev_event_id in prev_event_ids_from_query:
+            if prev_event_id not in non_outlier_prev_events:
+                raise SynapseError(
+                    HTTPStatus.BAD_REQUEST,
+                    "prev_event %s does not exist, or is an outlier" % (prev_event_id,),
+                    errcode=Codes.INVALID_PARAM,
+                )
```
MadLittleMods, line 137: We should add a test for this case, [`tests/rest/client/test_room_batch.py`](https://github.com/matrix-org/synapse/blob/1f32b90b0ff0f4f4f499f95e8150efe438388270/tests/rest/client/test_room_batch.py)

MadLittleMods, line 137: This is surprising that it's necessary. Is `get_most_recent_full_state_ids_from_event_id_list` -> `state_store.get_state_ids_for_event(...)` return state for an `outlier`? It seems to look at `state_groups` which we don't expect to exist for an `outlier` but we know it can happen somehow: https://github.com/matrix-org/synapse/issues/12201#issuecomment-1081745471

I'd expect `get_most_recent_full_state_ids_from_event_id_list` to find no state for an `outlier` and the `if not state_event_ids:` check would have fired before

richvdh: > I'd expect get_most_recent_full_state_ids_from_event_id_list to find no state for an outlier and the if not state_event_ids: check would have fired before

currently, that's the case. I'm seeking to change it in https://github.com/matrix-org/synapse/pull/12191.

Issue #12305, opened 2022-03-26
Add developer docs to explain pagination tokens


MadLittleMods:
Add developer docs to explain pagination tokens.

The comment docs explain general things around the pagination tokens well but when I was confronted with `s2633508_17_338_6732159_1082514_541479_274711_265584_1`, it wasn't obvious to me how to decipher it. I knew the `stream_ordering` (`s2633508`) part but it was really fuzzy what the other numbers were and the comment docs don't explain that part. I only really figured it out while drafting this issue and looking at the code more.

Relevant code:

 - [`RoomStreamToken`](https://github.com/matrix-org/synapse/blob/3c41d87b67d3a62edfc660b4fe8f2545f5dbee4f/synapse/types.py#L418-L597)
 - [`StreamToken`](https://github.com/matrix-org/synapse/blob/3c41d87b67d3a62edfc660b4fe8f2545f5dbee4f/synapse/types.py#L600-L678)
 
 
Relevant endpoints:

 - `/sync`
    - `?since`
    - `next_batch`
    - `prev_batch`
 - `/messages`
    - `?from`
    - `?to`
    - `start`
    - `end`
 - others

## Live tokens (`stream_ordering`)

https://github.com/matrix-org/synapse/blob/3c41d87b67d3a62edfc660b4fe8f2545f5dbee4f/synapse/types.py#L436-L437

ex.

 - `s2633508_17_338_6732159_1082514_541479_274711_265584_1`
    1. `room_key`: `s2633508` -> `2633508` `stream_ordering`
    1. `presence_key`: `17`
    1. `typing_key`: `338`
    1. `receipt_key`: `6732159`
    1. `account_data_key`: `1082514`
    1. `push_rules_key`: `541479`
    1. `to_device_key`: `274711`
    1. `device_list_key`: `265584`
    1. `groups_key`: `1`
 - `s1_33_0_1_1_1_1_7_1`
 - `s843_0_0_0_0_0_0_0_0`

Each number key are concatenated together in this order:

https://github.com/matrix-org/synapse/blob/3c41d87b67d3a62edfc660b4fe8f2545f5dbee4f/synapse/types.py#L636-L649

And represent the position of the various fields in the `/sync` response:

```
{
  "next_batch": "s12_4_0_1_1_1_1_4_1",
  "presence": {
    "events": [
      {
        "type": "m.presence",
        "sender": "@the-bridge-user:hs1",
        "content": {
          "presence": "offline",
          "last_active_ago": 103
[...]

PR #12317, opened 2022-03-29
Explain how to decipher live and historic pagination tokens

MadLittleMods:
Explain how to decipher live and historic pagination tokens

 - `s2633508_17_338_6732159_1082514_541479_274711_265584_1`
 - `t426-2633508_17_338_6732159_1082514_541479_274711_265584_1`

Fix https://github.com/matrix-org/synapse/issues/12305

### Pull Request Checklist

### DIFF
--- a/synapse/types.py
+++ b/synapse/types.py
@@ -433,6 +433,11 @@ class RoomStreamToken:
     (`topological_ordering` in the event graph) and tie-broken by
     `stream_ordering` (when the event arrived at the homeserver).
 
+    If you're looking for more info about what a token with all of the
+    underscores means, ex.
+    `s2633508_17_338_6732159_1082514_541479_274711_265584_1`, see the docstring
+    for `StreamToken` below.
+
     ---
 
     Live tokens start with an "s" followed by the `stream_ordering` of the event
--- a/synapse/types.py
+++ b/synapse/types.py
@@ -453,8 +453,8 @@ class RoomStreamToken:
     Historic tokens start with a "t" followed by the `depth`
     (`topological_ordering` in the event graph) of the event that comes before
     the position of the token, followed by "-", followed by the
-    `stream_ordering` of the event it comes after along with rest of the same
-    keys from the live tokens. An example token is:
+    `stream_ordering` of the event that comes before the position of the token.
+    An example token is:
 
         t426-2633508
 
--- a/synapse/types.py
+++ b/synapse/types.py
@@ -624,7 +624,7 @@ class RoomStreamToken:
 @attr.s(slots=True, frozen=True, auto_attribs=True)
 class StreamToken:
     """A collection of keys joined together by underscores in the following
-    order and represent the position in their respective streams.
+    order and which represent the position in their respective streams.
 
     ex. `s2633508_17_338_6732159_1082514_541479_274711_265584_1`
         1. `room_key`: `s2633508` which is a `RoomStreamToken`
--- a/synapse/types.py
+++ b/synapse/types.py
@@ -426,85 +426,32 @@ class RoomStreamToken:
     Tokens can either be a point in the live event stream or a cursor going
     through historic events.
 
-    When traversing the live event stream, events are ordered by when they
-    arrived at the homeserver.
+    When traversing the live event stream, events are ordered by
+    `stream_ordering` (when they arrived at the homeserver).
 
-    When traversing historic events, events are first ordered by their
-    "depth" ("topological_ordering" in the event graph) and tie-broken by
-    "stream_ordering" (when the event arrived at the homeserver).
-
-    ---
-
-    Live tokens start with an "s" followed by the "stream_ordering" id of the
-    event that comes after. The rest of the keys are joined together by
-    underscores in the following order and represent the position of various
-    data.
-
-    ex. `s2633508_17_338_6732159_1082514_541479_274711_265584_1`
-        1. `room_key`: `s2633508` -> `2633508` `stream_ordering`
-        2. `presence_key`: `17`
-        3. `typing_key`: `338`
-        4. `receipt_key`: `6732159`
-        5. `account_data_key`: `1082514`
-        6. `push_rules_key`: `541479`
-        7. `to_device_key`: `274711`
-        8. `device_list_key`: `265584`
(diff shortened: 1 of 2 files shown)


synapse/types.py
```diff
@@ -421,22 +421,92 @@ class RoomStreamToken:
```
MadLittleMods, line 421: What's the preference on referring to these tokens? "pagination tokens" or "position tokens"?

squahtx: I've only ever heard them called "pagination tokens".

synapse/types.py
```diff
@@ -421,22 +421,92 @@ class RoomStreamToken:
...
+    "depth" ("topological_ordering" in the event graph) and tie-broken by
+    "stream_ordering" (when the event arrived at the homeserver).
+
+    ---
 
     Live tokens start with an "s" followed by the "stream_ordering" id of the
-    event it comes after. Historic tokens start with a "t" followed by the
-    "topological_ordering" id of the event it comes after, followed by "-",
-    followed by the "stream_ordering" id of the event it comes after.
+    event that comes after. The rest of the keys are joined together by
```
squahtx, line 439: Does the token represent a position before or after the event with the given stream_ordering? I'm confused as to which is right.

synapse/types.py
```diff
@@ -421,22 +421,92 @@ class RoomStreamToken:
...
+    ex. `s2633508_17_338_6732159_1082514_541479_274711_265584_1`
+        1. `room_key`: `s2633508` -> `2633508` `stream_ordering`
+        2. `presence_key`: `17`
+        3. `typing_key`: `338`
+        4. `receipt_key`: `6732159`
+        5. `account_data_key`: `1082514`
+        6. `push_rules_key`: `541479`
+        7. `to_device_key`: `274711`
+        8. `device_list_key`: `265584`
+        9. `groups_key`: `1`
```
squahtx, line 452: Aren't these values from `StreamToken`s, rather than `RoomStreamToken`? Would it be more appropriate to update that docstring instead?

MadLittleMods: I was just adding to the docstring that seems more like the source of truth. It seems like all of this info should be on one or the other for easier reading and getting the full picture vs splitting it up between.

I don't understand the distinction between `RoomStreamToken` and `StreamToken`? They're intertwined and I can't spot the difference with the `cls` stuff.

synapse/types.py
MadLittleMods: I don't know for sure but based on the previous comments, I would say the token represents the position before the given "id". But the "id" itself is the one that comes after the position. Can be phrased either way. Better way to explain?

Here are the previous comments around this.

> Live tokens start with an "s" followed by the "stream_ordering" id of the event it comes after.
    
> Historic tokens start with a "t" followed by the "topological_ordering" id of the event it comes after

squahtx: After a bit of a call with @erikjohnston:
`stream_ordering`s uniquely identify persisted events. When a `RoomStreamToken` contains a `stream_ordering`, the token means "the position just after the event identified by `stream_ordering`". I'm not sure if it works differently for backfill.

That is, I think the wording should be "the event it comes after" instead of "the event that comes after".
And the same for line +491.

synapse/types.py
squahtx: Using the example above, a `StreamToken` would look like "s2633508_17_338_6732159_1082514_541479_274711_265584_1", with the "s2633508" part being a `RoomStreamToken`.
A `RoomStreamToken` would look like "s2633508", "t426-2633508" or "m56\~2.58\~3.59".

`cls` in a `RoomStreamToken` method looks confusing but it really means "RoomStreamToken". And `cls` in a `StreamToken` method means "StreamToken".

------

I'm not all that keen on having the documentation for `StreamToken` or `RoomStreamToken` live in a different docstring. At the very least, it would be nice if one or the other docstring had some wording along the lines of "See the docstring for `(Room)StreamToken` for a description of/details on <...>"

synapse/types.py
MadLittleMods: The nuance here is making me read this 1000 times 😵

---

The first explanation makes sense 👍

> `stream_ordering`s uniquely identify persisted events. When a `RoomStreamToken` contains a `stream_ordering`, the token means "the position just after the event identified by `stream_ordering`".


How does this sound?

```diff
+ Live tokens start with an "s" followed by the "stream_ordering" id of the
+ event that comes before position of the token.
```

synapse/types.py
MadLittleMods: Thank you for the explanation and breakdown @clokep and @squahtx 🙇 

Split the info out to their appropriate classes 👍


---

The naming of these two classes pretty ambiguous. It's all understandable just not immediately clear to me:

 - `RoomStreamToken` is used as the `room_key` in a `StreamToken`. The naming doesn't lend itself to be obvious that a `room_key` is a `RoomStreamToken`
 - `RoomStreamToken` and `StreamToken` includes the word "stream" which immediately jumps my brain to `stream_ordering` and `s999` type tokens. Instead of generic streams of data
 


Spit-balling, these seem more obvious although this is a different conversation to have ⏩

  - `RoomStreamToken` -> `RoomPositionKey` (or `RoomStreamPosition`, `RoomPosition`)
  - `StreamToken` -> `PaginationToken` (or `StreamPaginationToken`)

synapse/types.py
```diff
@@ -421,22 +421,44 @@ class RoomStreamToken:
...
+
+        s2633508
+
+    ---
+
+    Historic tokens start with a "t" followed by the `depth`
+    (`topological_ordering` in the event graph) of the event that comes before
+    the position of the token, followed by "-", followed by the
+    `stream_ordering` of the event it comes after along with rest of the same
+    keys from the live tokens. An example token is:
```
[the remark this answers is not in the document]
MadLittleMods: Good cactch, forgot to update this last place 👍

> I suspect the `depth` and `stream_ordering` usually come from the same event?

Looking at the code, it does seem to pull `depth` and `stream_ordering` from a given event when making a historic token. But I could see it being different. As there can gap in `stream_ordering` between in events in the room and all of them are still valid to represent that position.

squahtx (approved):
lgtm

Issue #12247, opened 2022-03-18
Expose user_directory to the module system


Samonitari:
**Description:**

It would be nice if modules could extend the user_directory search. If I am right, there is no callback interface, and you cannot register the needed web resource in a module, by the current restrictions
Reasons/example scenario:
- the [ldap auth backend](https://github.com/matrix-org/matrix-synapse-ldap3) handles the job of relying on a central, non-synapse managed user database. Other backends also apply
- if an auth backend is used, you either:
  - can't find the existing (on the backend) users, or
  - with config `user_directory.seach_all_users: true` you find **all** (maybe? not tested)
- there should be something in-between the previous two options, like find anybody in the same LDAP organization unit or group. In other words, the discoverability could (or rather should, in some deployments) be sourced from the same backend the auth one uses.
- ma1sd kinda did this, but that is abandonware (and the whole identity server thing needs love), and it made it work by intercepting the `/_matrix/client/r0/user_directory/` HTTP requests, which is not elegant at all. The module system would be the optimal place to do this.

Callbacks would be better than selectively enabling the mentioned resource, I presume.

hanadi92:
Another ugly temporary way to achieve that: extend the `UserDirectoryHandler` with your custom logic and overwrite the `api.hs.get_user_directory_handler` with a ModuleApi class.

PR #12318, opened 2022-03-29
add Module API callback for user directory search

Samonitari:
Signed-off-by: Krisztian Szegi <k.git.hub@meszaros-szegi.cloud>

### Pull Request Checklist

### DIFF
--- a/synapse/handlers/user_directory.py
+++ b/synapse/handlers/user_directory.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 import logging
-from typing import TYPE_CHECKING, Any, Dict, List, Optional
+from typing import TYPE_CHECKING, Any, Awaitable, Callable, Dict, List, Optional
 
 import synapse.metrics
 from synapse.api.constants import EventTypes, HistoryVisibility, JoinRules, Membership
@@ -27,6 +27,9 @@ if TYPE_CHECKING:
     from synapse.server import HomeServer
 
 logger = logging.getLogger(__name__)
+
+# Types for callbacks to be registered via the module api
+ON_SEARH_USERS_CALLBACK = Callable[[str, str, int], Awaitable[Optional[SearchResult]]]
 
 
 class UserDirectoryHandler(StateDeltasHandler):
@@ -76,6 +79,14 @@ class UserDirectoryHandler(StateDeltasHandler):
             # we start populating the user directory
             self.clock.call_later(0, self.notify_new_event)
 
+        self._on_search_users_callbacks: List[ON_SEARH_USERS_CALLBACK] = []
+
+    def register_user_directory_callbacks(
+        self, on_search_users: Optional[ON_SEARH_USERS_CALLBACK] = None
+    ) -> None:
+        if on_search_users is not None:
+            self._on_search_users_callbacks.append(on_search_users)
+
     async def search_users(
         self, user_id: str, search_term: str, limit: int
     ) -> SearchResult:
@@ -96,6 +107,23 @@ class UserDirectoryHandler(StateDeltasHandler):
                 }
         """
         results = await self.store.search_user_dir(user_id, search_term, limit)
+
+        merged_results = results["results"]
+        existing_user_ids = [user["user_id"] for user in merged_results]
+        for callback in self._on_search_users_callbacks:
+            external_users = await callback(user_id, search_term, limit)
+            if external_users is not None:
+                # for user in external_users["results"]:
+                #     if user["user_id"] not in existing_user_ids:
+                #         merged_results.append(user)
+                merged_results += [
+                    user
+                    for user in external_users["results"]
+                    if user["user_id"] not in existing_user_ids
+                ]
+
+        results["results"] = merged_results
+        results["limited"] = len(merged_results) > limit
 
         # Remove any spammy users from the results.
         non_spammy_users = []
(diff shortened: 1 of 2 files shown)


synapse/handlers/user_directory.py
```diff
@@ -97,6 +108,23 @@ async def search_users(
         """
         results = await self.store.search_user_dir(user_id, search_term, limit)
 
+        merged_results = results["results"]
+        existing_user_ids = [user["user_id"] for user in merged_results]
+        for callback in self._on_search_users_callbacks:
+            external_users = await callback(user_id, search_term, limit)
+            if external_users is not None:
+                # for user in external_users["results"]:
```
Samonitari, line 116: Which is preferred: loop style, or the list comprehension below?

synapse/handlers/user_directory.py
```diff
@@ -97,6 +108,23 @@ async def search_users(
         """
         results = await self.store.search_user_dir(user_id, search_term, limit)
 
+        merged_results = results["results"]
+        existing_user_ids = [user["user_id"] for user in merged_results]
```
Samonitari, line 112: merging the external user directory with synapse's works so that: if user_id match -> user match.
I don't know if it is realistic that the display name or avatar is different in the "backend", but this seems like the safe(r) thing to do

synapse/handlers/user_directory.py
```diff
@@ -97,6 +108,23 @@ async def search_users(
...
+        for callback in self._on_search_users_callbacks:
+            external_users = await callback(user_id, search_term, limit)
+            if external_users is not None:
+                # for user in external_users["results"]:
+                #     if user["user_id"] not in existing_user_ids:
+                #         merged_results.append(user)
+                merged_results += [
```
Samonitari, line 119: Which is preferred: loop style, or the list comprehension below?

Samonitari:
I'd like this to be merged, so #12247 could be closed.
I could then add the callback to [matrix-synapse-ldap3](https://github.com/matrix-org/matrix-synapse-ldap3) auth backend, making the latter a 100% replacement for the dead ma1sd, at least in a LDAP based setup.

dklimpel:
Is there no docs needed for `synapse/docs/modules/`?

Samonitari:
Fair enough - I'll add the docs, and remove the commented out loop!

synapse/handlers/user_directory.py
Samonitari: Okay, list comprehension it is, removing commented out loop

babolivier (changes requested):
Thanks for your contribution! I've left a few points I'd like to see addressed before we can merge this. I'd also like to see some tests for this new callback - see [this snippet](https://github.com/matrix-org/synapse/blob/15382b1afad65366df13c3b9040b6fdfb1eccfca/tests/rest/client/test_third_party_rules.py#L566-L898) for a few examples of tests that involve module callbacks. I'd also like to see some documentation about this, but from the previous comments on this PR it looks like you're already on it.

synapse/handlers/user_directory.py
```diff
@@ -28,6 +28,9 @@
 
 logger = logging.getLogger(__name__)
 
+# Types for callbacks to be registered via the module api
+ON_SEARH_USERS_CALLBACK = Callable[[str, str, int], Awaitable[Optional[SearchResult]]]
```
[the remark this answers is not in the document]
Samonitari: Fair enough, I didn't consider the sequential meaning `on_[...]` would bring
`get_external_results` feels a bit too general, and external potentially isn't the most correct either, as these users could be already mirrored to database.
`SEARCH_USERS_CALLBACK` simply?

synapse/handlers/user_directory.py
```diff
@@ -97,6 +108,20 @@ async def search_users(
...
+        merged_results = results["results"]
+        existing_user_ids = [user["user_id"] for user in merged_results]
+        for callback in self._on_search_users_callbacks:
+            external_users = await callback(user_id, search_term, limit)
+            if external_users is not None:
+                merged_results += [
+                    user
+                    for user in external_users["results"]
+                    if user["user_id"] not in existing_user_ids
+                ]
```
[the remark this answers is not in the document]
Samonitari: Fair enough, correcting

synapse/handlers/user_directory.py
```diff
@@ -97,6 +108,20 @@ async def search_users(
...
+
+        results["results"] = merged_results
+        results["limited"] = len(merged_results) > limit
```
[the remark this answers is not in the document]
Samonitari: I did only really understood the logic behind `limited` *after* I committed.
SQL query does not report if the `limit` was needed or not - so the +1 is a workaround so we can now if there **are** more users. And that is a bit "meh":
Maybe the workaround *could* also strip the the last found result **if** query returned `limit + 1`, as: technically we maybe *are* falsely reporting limited when in fact, there is *exactly* `limit+1` users.

OTOH, the correct merging would be maybe:
- if `limited` is false before merge, not merge more "external" users if limit is reached, and set limited
- if `limited` is true before merge, don't merge at all

janpawellek:
Cool, thank you for the PR 🥳 

We also need this feature, thus your PR looks very promising. Keep up your good work!
Looking forward to get it into Synapse 😃

clokep:
@Samonitari or @janpawellek -- do you wish to continue this PR? It looks like it needs some updates and the feedback addressed.

Samonitari:
Yes, but to be a bit more direct - and to not make any bloody estimation lies - sometime this year I'll do this.
I'll keep you updated, but the next two weeks I am consolidating two hosting servers into one - I am halfway there!

Issue #12281, opened 2022-03-24
Missing sync events during historical batch imports


Fizzadar:
Split out from https://github.com/matrix-org/synapse/issues/9424#issuecomment-1061992367 onwards. I have inlined the relevant context from that issue below.

We have re-created the following issue; I've included the event list and trimmed sync responses in this gist: https://gist.github.com/Fizzadar/316dabfe9be30ebaa48e288de2380de1. Some notes:

+ The first sync containing the invite seems to correctly pull everything
+ The second sync includes member information but misses out the prior state
+ The invite also appears to arrive after the room creation event
+ Prior syncs do not reference the room at all (can provide JSON dumps if useful)

Notably, this is only occuring during initial backfill using MSC2716 batch sends. When this is disabled on the bridge we have not yet replicated the problem seen here.

The DAG:

```
synapse=> SELECT stream_ordering, type, state_key, event_id, (ej.json::json)->'prev_events' AS prev_events FROM events
LEFT JOIN event_json AS ej USING (event_id)
WHERE events.room_id ='!SjekxtmsBigrUMcMkX:beeper-dev.com' ORDER BY stream_ordering ASC;
 stream_ordering |             type             |                          state_key                           |                   event_id                   |                   prev_events
-----------------+------------------------------+--------------------------------------------------------------+----------------------------------------------+--------------------------------------------------
         -113203 | org.matrix.msc2716.insertion |                                                    | $G5GLwX6lc8MXKqwauoM1p9lkLD6S5GfwcosFlPra0ws | []
         -113202 | m.room.encrypted             |                                                    | $I3PsF44AMu5I_XOZQ2Hm1cOhcdLJCbA8NcAMS69hLEE | ["$G5GLwX6lc8MXKqwauoM1p9lkLD6S5GfwcosFlPra0ws"]
         -113201 | m.room.encrypted             |                                                    | $hIO0ZI-ZYr8gWGkWtvBiPfCA0jKdjiZP7PCnEfuutTY | ["$I3PsF44AMu5I_XOZQ2Hm1cOhcdLJCbA8NcAMS69hLEE"]
         -113200 | m.room.encrypted             |                                                    | $-jWSmA7NJ1dbl5b86K8FAKN2ydt-qoOw8ftCdSEN4A4 | ["$hIO0ZI-ZYr8gWGkWtvBiPfCA0jKdjiZP7PCnEfuutTY"]
[...]

Issue #3305, opened 2018-05-29
incremental /sync may give incomplete results when rejoining a room


richvdh:
I rather hope this will get fixed when we make depth not be a thing, but:

* incremental /sync looks for rooms you have joined since your previous sync
* it does this by looking at your membership state in each room at your 'since' token (https://github.com/matrix-org/synapse/blob/933bf2dd357842b0e7c3fc7a1111d89ab52f5329/synapse/handlers/sync.py#L1087 ish), and looking for those that are not 'join'
* however, `get_state_at` uses `get_recent_events_for_room` which returns the event with the greatest depth, so if somebody has been playing silly buggers with the depth, you may end up looking at a point in the graph where you were a member, even though at the point of the 'since' token you were not.

PR #12319, opened 2022-03-29
Fix missing sync events during historical batch imports

Fizzadar:
Discovered after much in-depth investigation in #12281. Adding an explicit `order_by` argument to change the ordering of results returned from room pagination queries and updating use in the sync & message handlers.

Closes: #12281 
Closes: #3305

Signed off by: Nick Mills-Barrett <nick@beeper.com>

### Pull Request Checklist

* [x] Pull request includes a [changelog file](https://matrix-org.github.io/synapse/latest/development/contributing_guide.html#changelog)


synapse/storage/databases/main/stream.py
```diff
@@ -1099,6 +1115,7 @@ def _paginate_room_events_txn(
         from_token: RoomStreamToken,
         to_token: Optional[RoomStreamToken] = None,
         direction: str = "b",
+        order_by: str = "topological",
```
MadLittleMods, line 1118: Instead of adding a new `order_by` argument, I think we can just check whether `from_token.topological is not None` to determine whether to sort by `topological` or `stream_ordering`

synapse/handlers/sync.py
```diff
@@ -683,12 +683,11 @@ async def get_state_at(
...
-        # get_recent_events_for_room operates by topo ordering. This therefore
-        # does not reliably give you the state at the given stream position.
-        # (https://github.com/matrix-org/synapse/issues/3305)
         last_events, _ = await self.store.get_recent_events_for_room(
-            room_id, end_token=stream_position.room_key, limit=1
+            room_id,
+            end_token=stream_position.room_key,
+            limit=1,
+            order_by="stream",
         )
```
MadLittleMods, line 691: We should add a test for https://github.com/matrix-org/synapse/issues/12281 so it doesn't regress.

[Other MSC2716 `/batch_send` test](https://github.com/matrix-org/synapse/blob/287a9c1e20e1e271d69be581d6bb8ee5a8de9d88/tests/rest/client/test_room_batch.py) for reference. We can also use the barebones reproduction steps explained in https://github.com/matrix-org/synapse/issues/12281#issuecomment-1081199403 for the test.

I think it would be possible write a [Complement](https://github.com/matrix-org/complement/pull/352) test instead but since the problem is so closely tied to `stream_ordering`, it would probably be best to lookup and manipulate from a trial unit test.

MadLittleMods, line 691: We should add a test for https://github.com/matrix-org/synapse/issues/3305 so it doesn't regress.

I don't have clear reproduction steps for this one and maybe we can just consider the test we write for https://github.com/matrix-org/synapse/issues/12281 as a subset of this problem and only need one test 🤷

synapse/storage/databases/main/stream.py
Fizzadar: My concern with this approach is the magic nature of it, a reader new to the code won’t know if a token is topological or not and thus the ordering becomes much harder to reason about.
Unless I’m missing something on the token definition that is! Happy to switch to using the token detection if that would be preferred?

synapse/handlers/sync.py
Fizzadar: Pulled together [a test case here](https://github.com/matrix-org/synapse/pull/12319/commits/e440db94d638b778a5a6b7ac2dc909cb37093445) - not sure this is the right location for it as it could sit in sync or batch.

synapse/storage/databases/main/stream.py
MadLittleMods: I think we just want it to work correctly with whatever token is provided.

Is there a case where we want to sort by `depth`(`topological_ordering`) when we give it a stream token? Or vice versa?

Fizzadar: Makes sense - I did spot other cases where the token defines this kind of behavior elsewhere as well since my initial change; added https://github.com/matrix-org/synapse/pull/12319/commits/ef07fcda333cd22825661eee5f8615b258ac3505 to address.

Fizzadar: Hm. This appears to have broken some complement tests now!

MadLittleMods: > Hm. This appears to have broken some complement tests now!

Looking into this more. I think this may be a bug in [`get_messages(...)`](https://github.com/matrix-org/synapse/blob/bebf994ee804ef63ce16801c6694713fcd685320/synapse/handlers/pagination.py#L419). 

The following code will always return a `stream_ordering` token but when we're working with `/messages`, it should be a topological token (I think). With the previous code, I feel like it just masked this problem because we always did the `topological_order` sort with `_paginate_room_events_txn` before regardless of what the token was.

https://github.com/matrix-org/synapse/blob/bebf994ee804ef63ce16801c6694713fcd685320/synapse/handlers/pagination.py#L444

Relevant call stack:

```
on_GET (/messages)
get_messages
paginate_room_events
_paginate_room_events_txn
```


## Debugging that failing Complement test

Relevant Complement test: https://github.com/matrix-org/complement/blob/1f527b05ea90afb5ab4d164980f8522576d2b219/tests/msc2716_test.go#L84-L187

Running the test:
```
COMPLEMENT_ALWAYS_PRINT_SERVER_LOGS=1 COMPLEMENT_DIR=../complement ./scripts-dev/complement.sh TestImportHistoricalMessages/parallel/Historical_events_resolve_in_the_correct_order
```


See these steps to access database for homeserver after Complement test runs: https://github.com/matrix-org/complement/pull/165/files

```sh
$ docker exec -it complement_fed_1_fed.hs_with_application_service.hs1_1 /bin/bash
$ apt-get update && apt-get install -y sqlite3
$ sqlite3
> .open /conf/homeserver.db
# To get the room ID, just look at any of the events
> SELECT * FROM events;

# Pretty print the output better
> .mode column
> .headers on
```

This is expected `topological_order` sort and how it behaves when the test passes (you can see how this [matches up to the test here](https://github.com/matrix-org/complement/blob/1f527b05ea90afb5ab4d164980f8522576d2b219/tests/msc2716_test.go#L155-L166)):

```
SELECT depth, stream_ordering, type, state_key, event_id FROM events
WHERE events.room_id ='!FEvmZPRLBIEafdrJWQ:hs1'
ORDER BY depth DESC, stream_ordering DESC;
depth  stream_ordering  type                          state_key             event_id
-----  ---------------  ----------------------------  --------------------  --------------------------------------------
# eventIDsBefore
11     12               m.room.message                                      $Wt7wI7MnraTqyDIu83-zWTa6SKOhw2mj3vU9bumrpwg
10     11               m.room.message                                      $NqSxI26ke8xr9B2eCH05gTSDzpzryw1Jde1EuElONwA
# baseInsertionEventID0
10     -3               org.matrix.msc2716.insertion                        $vGkND0_QMEhU12MJkxc7QwNPv5OFI8bfmzg1TPSgrC8
# historical batch 0
10     -4               org.matrix.msc2716.batch                            $E407MKifUI5x_KQMS1JvJ3WAbRPaDRAR814b0IJ0DIg
10     -5               m.room.message                                      $TS2vmOnEL8bQvbNOzKEYLgvsBIaNW_h8RFckjQsnej4
10     -6               m.room.message                                      $wlBalhDWofWEwxz1ZQoEOMilN6veASLHmIL2CReDoLw
10     -7               m.room.message                                      $840NKtUjST7gNG5zteq17-jfGYHkPH2n_hATD3OOH4I
10     -8               org.matrix.msc2716.insertion                        $ww-VPKdpa_vOE5l2GLpKZNBtgbvn2oEKi1c8KlIvLsU
# historical batch 1
10     -10              org.matrix.msc2716.batch                            $v_sDaykItrVQW-HEd3dt7voIoazLpuw9EtN4syyjCjQ
10     -11              m.room.message                                      $G8KlqUhMsknRszhw3dEJznLWqzWf7bKMRLvtVaOd5mU
10     -12              m.room.message                                      $9IoiumsFGnrXIbcnDG_LjIJjqsQU_z35jTx7fxB13iQ
10     -13              m.room.message                                      $3fxfwV6oz9DZ20pwlhWJgExd9n7hsP7k9w3GM1LyQgw
10     -14              org.matrix.msc2716.insertion                        $lK21br0ZyavuewyBlPAr57WZHItKHvG37ffIBeBDgZ4
# eventIDsAfter
9      10               m.room.message                                      $Z-D4EQEzgrwhr-R1_-JjmAUHIGUjusQp5yh8Dbe4Q6Q
8      9                m.room.message                                      $tjnuGZaismCDTPUoHJ87agAaH9CKyTIIfCYsjKRBiSs
# start of room
7      8                m.room.member                 @alice:hs1            $t1JOLDZLR48r91UmWn0kPmRgwt97usvz1p1Zcx7SJqs
6      7                m.room.name                                         $UhxaGndH86gNZTya1Nuv09eBGtMGc2KBvWYhpeSys1k
5      6                m.room.history_visibility                           $l0f9onBohOxAbGDQ8gudGHHpEgBWAZwSbSkdGdjDo7A
4      5                m.room.join_rules                                   $UZwW_RMaYUyyk8x3WbNowoRNhd-lW85QWD6qBdSEly8
3      4                m.room.power_levels                                 $Yooqdyfxw9m-xWzMnkUIcDMlJlDJQjnDDd2CAdF-kzg
2      3                m.room.member                 @the-bridge-user:hs1  $QynAfgtUHeBUjN1YXvQ6Ye5E1hyE9q59bO2TH89HtQg
1      2                m.room.create                                       $gkDWQFo9jfu_w9pbzSypK_dF9aqnw3aCecnG5wQiCGQ
1      -2               m.room.member                 @maria:hs1            $dWDbmnkE_RpnCmD1Q6ZJHCVMDQvWeJ0kSzCc0GBqHSA
1      -9               m.room.member                 @maria:hs1            $hhIfpjLltBmvMjwTALIGSAfIeY-d3n2vwRXG2s-b-QQ
```

This is the `stream_ordering` sort it tries to use which is not the order we expect in the test.
```
SELECT depth, stream_ordering, type, state_key, event_id FROM events
WHERE events.room_id ='!FEvmZPRLBIEafdrJWQ:hs1'
ORDER BY stream_ordering DESC, depth DESC;
depth  stream_ordering  type                          state_key             event_id
-----  ---------------  ----------------------------  --------------------  --------------------------------------------
11     12               m.room.message                                      $Wt7wI7MnraTqyDIu83-zWTa6SKOhw2mj3vU9bumrpwg
10     11               m.room.message                                      $NqSxI26ke8xr9B2eCH05gTSDzpzryw1Jde1EuElONwA
9      10               m.room.message                                      $Z-D4EQEzgrwhr-R1_-JjmAUHIGUjusQp5yh8Dbe4Q6Q
8      9                m.room.message                                      $tjnuGZaismCDTPUoHJ87agAaH9CKyTIIfCYsjKRBiSs
7      8                m.room.member                 @alice:hs1            $t1JOLDZLR48r91UmWn0kPmRgwt97usvz1p1Zcx7SJqs
6      7                m.room.name                                         $UhxaGndH86gNZTya1Nuv09eBGtMGc2KBvWYhpeSys1k
5      6                m.room.history_visibility                           $l0f9onBohOxAbGDQ8gudGHHpEgBWAZwSbSkdGdjDo7A
4      5                m.room.join_rules                                   $UZwW_RMaYUyyk8x3WbNowoRNhd-lW85QWD6qBdSEly8
3      4                m.room.power_levels                                 $Yooqdyfxw9m-xWzMnkUIcDMlJlDJQjnDDd2CAdF-kzg
2      3                m.room.member                 @the-bridge-user:hs1  $QynAfgtUHeBUjN1YXvQ6Ye5E1hyE9q59bO2TH89HtQg
1      2                m.room.create                                       $gkDWQFo9jfu_w9pbzSypK_dF9aqnw3aCecnG5wQiCGQ
1      -2               m.room.member                 @maria:hs1            $dWDbmnkE_RpnCmD1Q6ZJHCVMDQvWeJ0kSzCc0GBqHSA
10     -3               org.matrix.msc2716.insertion                        $vGkND0_QMEhU12MJkxc7QwNPv5OFI8bfmzg1TPSgrC8
10     -4               org.matrix.msc2716.batch                            $E407MKifUI5x_KQMS1JvJ3WAbRPaDRAR814b0IJ0DIg
10     -5               m.room.message                                      $TS2vmOnEL8bQvbNOzKEYLgvsBIaNW_h8RFckjQsnej4
10     -6               m.room.message                                      $wlBalhDWofWEwxz1ZQoEOMilN6veASLHmIL2CReDoLw
10     -7               m.room.message                                      $840NKtUjST7gNG5zteq17-jfGYHkPH2n_hATD3OOH4I
10     -8               org.matrix.msc2716.insertion                        $ww-VPKdpa_vOE5l2GLpKZNBtgbvn2oEKi1c8KlIvLsU
1      -9               m.room.member                 @maria:hs1            $hhIfpjLltBmvMjwTALIGSAfIeY-d3n2vwRXG2s-b-QQ
10     -10              org.matrix.msc2716.batch                            $v_sDaykItrVQW-HEd3dt7voIoazLpuw9EtN4syyjCjQ
10     -11              m.room.message                                      $G8KlqUhMsknRszhw3dEJznLWqzWf7bKMRLvtVaOd5mU
10     -12              m.room.message                                      $9IoiumsFGnrXIbcnDG_LjIJjqsQU_z35jTx7fxB13iQ
10     -13              m.room.message                                      $3fxfwV6oz9DZ20pwlhWJgExd9n7hsP7k9w3GM1LyQgw
10     -14              org.matrix.msc2716.insertion                        $lK21br0ZyavuewyBlPAr57WZHItKHvG37ffIBeBDgZ4
```

changelog.d/12319.bugfix
```diff
@@ -0,0 +1 @@
+Fix bug with incremental sync missing events when rejoining/backfilling. Contributed by Nick @ Beeper.
```
MadLittleMods, line 1: As a note, I'm not a Synapse reviewer/maintainer. After we get the tests passing again ✅, we should assign this back to `matrix-org/synapse-core`

Fizzadar:
@MadLittleMods I'm a bit confused on next steps here - this seems to have unearthed another (complex looking) bug I don't have time (currently) to work on a fix for.

My suggestion would be to roll back to the explicit argument fix which resolves this issue and create a new issue to tackle the `get_messages` bug (which can also remove the new argument).

MadLittleMods:
> My suggestion would be to roll back to the explicit argument fix which resolves this issue and create a new issue to tackle the `get_messages` bug (which can also remove the new argument).

@Fizzadar Go for it 👍 We can get some feedback from the Synapse team whether that is acceptable

synapse/storage/databases/main/stream.py
MadLittleMods: @Fizzadar I've created https://github.com/matrix-org/synapse/pull/12370 to fix the `/messages` bug causing the Complement failures you were seeing.

changelog.d/12319.bugfix
MadLittleMods, line 1: For the reviewer: The CI failures described in https://github.com/matrix-org/synapse/pull/12319#discussion_r839267470 are fixed by https://github.com/matrix-org/synapse/pull/12370

The first thing to decide is whether we like the fix from this PR. If we like the fix there, then we also need to consider the fix in https://github.com/matrix-org/synapse/pull/12370 to fix the `/messages` bug that was uncovered in this PR.

MadLittleMods:
> @MadLittleMods I'm a bit confused on next steps here - this seems to have unearthed another (complex looking) bug I don't have time (currently) to work on a fix for.

As an update, I've fixed up the other [complex bug](https://github.com/matrix-org/synapse/pull/12319#discussion_r840276412) via https://github.com/matrix-org/synapse/pull/12370

So we can probably leave this PR as-is. I've assigned it back to the Synapse core for review ⏩

tests/rest/client/test_room_batch.py
```diff
@@ -178,3 +179,112 @@ def test_same_state_groups_for_whole_historical_batch(self) -> None:
...
+        state_event_types = [
+            event["type"]
+            for event in sync_response.json_body["rooms"]["join"][room_id]["state"][
+                "events"
+            ]
+        ]
+
+        assert (
+            "m.room.create" in state_event_types
+        ), "Missing room full state in sync response"
```
MadLittleMods, line 290: Tested and fails on `develop` but passes with this PR ✅ (as expected)

tests/rest/client/test_room_batch.py
```diff
@@ -178,3 +179,112 @@ def test_same_state_groups_for_whole_historical_batch(self) -> None:
...
+        response = self.helper.send_event(
+            room_id=room_id,
+            type=EventTypes.Message,
+            content={
+                "msgtype": "m.text",
+                "body": "C",
+            },
+            tok=self.appservice.token,
+        )
+        event_to_hang_id = response["event_id"]
```
MadLittleMods, line 228: Instead of creating a new `event_to_hang_id`. We can use one of the events that `_create_test_room` already creates with the room

```py
room_id, event_id_a, _, _ = self._create_test_room()
```

MadLittleMods: ehhh, nvm. It's important that we `/batch_send` off of the latest event in the room. 

But we should add a comment about why this is important ^

erikjohnston:
Thanks @Fizzadar and @MadLittleMods for tracking this down! 

My main concern is how this changes clients back paginating using a `prev_batch` token from a `/sync` `timeline` section. Those tokens are currently stream tokens, but `/messages` will return them in topological order with a topological next token. This change means that the first `/messages` will return events in stream order but will still return a topological next token, meaning that the next `/messages` will return events in a topological order.

I don't think that is a change we really want to make. My understanding is that that behaviour works correctly currently? Or does that also have issues with backflled/historical events?

changelog.d/12319.bugfix
MadLittleMods: @Fizzadar https://github.com/matrix-org/synapse/pull/12370 is now merged. Merge `develop`  into this PR to see the tests passing again ✅

synapse/handlers/message.py
```diff
@@ -175,10 +175,6 @@ async def get_state_events(
         state_filter = state_filter or StateFilter.all()
```
MadLittleMods, line 175: > My main concern is how this changes clients back paginating using a `prev_batch` token from a `/sync` `timeline` section. Those tokens are currently stream tokens, but `/messages` will return them in topological order with a topological next token. This change means that the first `/messages` will return events in stream order but will still return a topological next token, meaning that the next `/messages` will return events in a topological order.
>
> *-- @erikjohnston, https://github.com/matrix-org/synapse/pull/12319#issuecomment-1090135172*

`/messages` could be updated to always return in `topological_ordering` regardless of what type of pagination token is given. That's what it was doing before this PR anyway.

> I don't think that is a change we really want to make. My understanding is that that behaviour works correctly currently? Or does that also have issues with backfilled/historical events?

I think we have two assumptions that we should probably enforce:

 - `/messages` should always sort by `topological_ordering` regardless of pagination token given
 - `/sync` should always sort by `stream_ordering` regardless of pagination token given

This also aligns with what we have documented:

>  - `/sync` returns things in the order they arrive at the server (`stream_ordering`).
>  - `/messages` (and `/backfill` in the federation API) return them in the order determined by the event graph `(topological_ordering, stream_ordering)`.
>
> *-- [Room DAG concepts -> Depth and stream ordering](https://github.com/matrix-org/synapse/blob/f931c0602a1dcfd40696c53779d80dc68c9c9e17/docs/development/room-dag-concepts.md#depth-and-stream-ordering)*

The current behavior of `/messages` and `/sync` returns `events` as expected but not `state`.

There is a bug in how incremental `/sync` returns `state` which this PR aims to fix. The best way to understand the problem is probably reading https://github.com/matrix-org/synapse/issues/12281#issuecomment-1081199403 - but basically when calculating what `state` to return, `/sync` is ordering events by `topological_ordering` even though it should be `stream_ordering`.

---

I [initially assumed](https://github.com/matrix-org/synapse/pull/12319#discussion_r837946739) we would want flexibility in how these endpoints sorted events according to the type of pagination token given but it's sounding like we actually want to enforce a given sort according to the endpoint. In which case, we can revert to @Fizzadar initial approach of plumbing a `order_by` argument which we can set.

changelog.d/12319.bugfix
Fizzadar: Pushed, and complement passed OK! Not had a chance to look at the sytest failure yet!

tests/rest/client/test_room_batch.py
Fizzadar: https://github.com/matrix-org/synapse/pull/12319/commits/b2c6c20f4ab8e68b56b39b869af29c56e81d0ff7

synapse/handlers/message.py
erikjohnston: (Sorry for not getting to this yesterday, it sat at about the second thing on my todo list all day :/)

Thanks for the clarification. There's the added bonus that `/sync` will return events in pagination order for the first batch of events for the room, i.e. in an initial sync or when a user joins the room. (The rationale being that it's equivalent to getting no events then immediately doing a back pagination). This is relevant as we use `get_recent_events_for_room` for that case as well.



> I [initially assumed](https://github.com/matrix-org/synapse/pull/12319#discussion_r837946739) we would want flexibility in how these endpoints sorted events according to the type of pagination token given but it's sounding like we actually want to enforce a given sort according to the endpoint. In which case, we can revert to @Fizzadar initial approach of plumbing a `order_by` argument which we can set.

I wonder if we're just making this harder for ourselves by re-using the same query to do both `/messages` and when looking up the state for `/sync`? I think the query we want to run is simply:

```sql
SELECT event_id FROM events
WHERE room_id = ? AND stream_ordering <= ?
ORDER BY stream_ordering DESC
LIMIT 1
```

So it might be simpler to just have a separate `get_last_event_in_room_before` function that we call when getting the state instead? Rather than extending the already slightly horrific `_paginate_room_events_txn`  function?

---

Entirely separately, and I'm not suggesting we necessarily do this now, but it occurs to me that we have a `get_forward_extremities_for_room_at_stream_ordering` store function that maps from `room_id` to the list of forward extremities in the room at the time. This can then be used to calculate the "current" state of the room at the time more accurately than the current method. The downside of this approach is that  `get_forward_extremities_for_room_at_stream_ordering` will fail if the stream ordering is too old.


Another approach that has also literally just now occurred to me is that we could use the `current_state_delta_stream` to work out the changes in state between two stream orderings in the room, as the `stream_id` column correspond to the stream orderings of the changes (I think?).

Anyway, just wanted to record my thoughts here.

erikjohnston: @Fizzadar sorry for going around the houses really slowly on this :disappointed:

synapse/handlers/sync.py
```diff
@@ -661,16 +661,12 @@ async def get_state_at(
             stream_position: point at which to get state
             state_filter: The state filter used to fetch state from the database.
         """
-        # FIXME this claims to get the state at a stream position, but
-        # get_recent_events_for_room operates by topo ordering. This therefore
-        # does not reliably give you the state at the given stream position.
-        # (https://github.com/matrix-org/synapse/issues/3305)
-        last_events, _ = await self.store.get_recent_events_for_room(
-            room_id, end_token=stream_position.room_key, limit=1
+        last_event = await self.store.get_last_event_in_room_before_stream_ordering(
```
erikjohnston, line 664:
```suggestion
        # FIXME: This gets the state at the latest event before the stream ordering,
        # which might not be the same as the "current state" of the room at the time
        # of the stream token if there were multiple forward extremities at the time.
        last_event = await self.store.get_last_event_in_room_before_stream_ordering(
```

This PR makes things a lot better, but its still not technically 100% correct. I think it makes sense to then leave in a `FIXME` comment.

erikjohnston (approved):

tests/rest/client/test_room_batch.py
```diff
@@ -178,3 +179,114 @@ def test_same_state_groups_for_whole_historical_batch(self) -> None:
...
+                content={"msgtype": "m.text", "body": "C"},
+                tok=self.appservice.token,
+            )
+
+        # Join the room as the normal user
+        self.helper.join(room_id, user_id, tok=user_tok)
+
+        # Create event to hang backfill batch from, and send batch - historical batches
+        # must be hung from the lmost recent event after the member join event to see
+        # this failure case.
```
MadLittleMods, line 220: I think we should explain more of the "why".

```suggestion
        # Create an event to hang the historical batch from - In order to see
        # the failure case originally reported in #12281, the historical batch
        # must be hung from the most recent event in the room so the base
        # insertion event ends up with the highest `topogological_ordering`
        # (`depth`) in the room but will have a negative `stream_ordering`
        # because it's a `historical` event. Previously, when assembling the
        # `state` for the `/sync` response, the bugged logic would sort by
        # `topological_ordering` descending and pick up the base insertion
        # event because it has a negative `stream_ordering` below the given
        # pagination token. Now we properly sort by `stream_ordering`
        # descending which puts `historical` events with a negative
        # `stream_ordering` way at the bottom and aren't selected as expected.
```

tests/rest/client/test_room_batch.py
MadLittleMods: Review continued here https://github.com/matrix-org/synapse/pull/12319/files#r848686533

erikjohnston:
Thank you again for bearing with us @Fizzadar :heart:

PR #12320, opened 2022-03-29
Bump `black` and `click` versions

DMRobertson:
To include https://github.com/psf/black/pull/2966 which is necessary
after https://github.com/pallets/click/issues/2198.

This should hopefully fix CI. Shoutout to #11537.

### DIFF
--- a/changelog.d/12320.misc
+++ b/changelog.d/12320.misc
@@ -1,0 +1,1 @@
+Bump the version of `black` for compatibility with the latest `click` release.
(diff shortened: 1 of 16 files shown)


DMRobertson:
Note that we pin both `black` and `click` in setup.py. It's just that `tox` decides to install the latest version of black and click. /shrug.

Given that the linter has passed I'm gonna set this to automerge. If there's a problem with bumping `click` then we can revert.

PR #12321, opened 2022-03-29
Track device list updates per room.

erikjohnston:
This is a first step in dealing with #7721.

The idea is basically that rather than calculating the full set of users a device list update needs to be sent to up front, we instead simply record the rooms the user was in at the time of the change. This will allow a few things:

1. we can defer calculating the set of remote servers that need to be poked about the change; and
2. during `/sync` and `/keys/changes` we can avoid also avoid calculating users who share rooms with other users, and instead just look at the rooms that have changed.

However, care needs to be taken to correctly handle server downgrades. As such this PR writes to both `device_lists_changes_in_room` and the `device_lists_outbound_pokes` table synchronously. In a future release we can then bump the database schema compat version to `69` and then we can assume that the new `device_lists_changes_in_room` exists and is handled.

There is a temporary option to disable writing to `device_lists_outbound_pokes` synchronously, allowing us to test the new code path does work (and by implication upgrading to a future release and downgrading to this one will work correctly).

Note: Ideally we'd do the calculation of room to servers on a worker (e.g. the background worker), but currently only master can write to the `device_list_outbound_pokes` table.

Reviewable commit-by-commit.


synapse/storage/databases/main/devices.py
```diff
@@ -1636,3 +1657,37 @@ def _add_device_outbound_poke_to_stream_txn(
...
+                "device_id",
+                "opentracing_context",
+            ),
+            values=[
+                (
+                    next(next_stream_id),
+                    room_id,
+                    user_id,
+                    device_id,
+                    json_encoder.encode(context),
```
richvdh, line 1727: can we do this once, rather than once for every room and every device?

synapse/storage/databases/main/devices.py
```diff
@@ -1636,3 +1657,37 @@ def _add_device_outbound_poke_to_stream_txn(
...
+    def _add_device_outbound_room_poke_txn(
+        self,
+        txn: LoggingTransaction,
+        user_id: str,
+        device_ids: Iterable[str],
+        room_ids: Collection[str],
+        stream_ids: List[str],
+        context: Dict[str, str],
+    ) -> None:
+        next_stream_id = iter(stream_ids)
```
richvdh: shouldn't this be called `stream_id_iterator` or something?

synapse/storage/databases/main/devices.py
```diff
@@ -1529,50 +1529,53 @@ async def add_device_change_to_streams(
         if not device_ids:
             return None
 
-        async with self._device_list_id_gen.get_next_mult(
-            len(device_ids)
-        ) as stream_ids:
-            await self.db_pool.runInteraction(
-                "add_device_change_to_stream",
-                self._add_device_change_to_stream_txn,
+        num_stream_ids = max(
```
richvdh: it might be good to add a comment clarifying why this is a `max` rather than a total (vis, that we use the same set of stream ids for each of the three tables)

synapse/storage/databases/main/devices.py
```diff
@@ -1529,50 +1529,53 @@ async def add_device_change_to_streams(
...
+                txn,
                 user_id,
                 device_ids,
                 stream_ids,
             )
 
-        if not room_ids:
-            return stream_ids[-1]
+            if not room_ids:
+                return
```
richvdh, line 1571: I'd be quite tempted to omit this (I believe `simple_insert_many_txn` handles an empty list correctly?)

failing that, some justification as to why we can bail out here might be good. "If the user isn't in any rooms, then clearly they also cannot share any rooms with remote users, so we can bail out now."

synapse/storage/schema/main/delta/69/01device_list_oubound_by_room.sql
```diff
@@ -18,7 +18,11 @@ CREATE TABLE device_lists_changes_in_room (
...
     user_id TEXT NOT NULL,
     device_id TEXT NOT NULL,
+    -- We track if we've calculated the hosts in the room and updated
+    -- `device_lists_outbound_pokes`.
+    converted_to_destinations BOOLEAN NOT NULL,
     opentracing_context TEXT
 );
 
 CREATE UNIQUE INDEX device_lists_changes_in_stream_id ON device_lists_changes_in_room(stream_id);
+CREATE INDEX device_lists_changes_in_stream_id_converted ON device_lists_changes_in_room(stream_id) WHERE NOT converted_to_destinations;
```
richvdh, line 28:
```suggestion
CREATE INDEX device_lists_changes_in_stream_id_unconverted ON device_lists_changes_in_room(stream_id) WHERE NOT converted_to_destinations;
```

synapse/storage/schema/main/delta/69/01device_list_oubound_by_room.sql
```diff
@@ -14,15 +14,15 @@
...
     device_id TEXT NOT NULL,
+    room_id TEXT NOT NULL,
+    stream_id BIGINT NOT NULL,  -- This matches `device_lists_stream.stream_id`
     -- We track if we've calculated the hosts in the room and updated
     -- `device_lists_outbound_pokes`.
     converted_to_destinations BOOLEAN NOT NULL,
     opentracing_context TEXT
 );
 
-CREATE UNIQUE INDEX device_lists_changes_in_stream_id ON device_lists_changes_in_room(stream_id);
```
richvdh: can we have *some* sort of unique index? it seems like `(stream_id, room_id)` should be unique?

synapse/storage/schema/main/delta/69/01device_list_oubound_by_room.sql
```diff
@@ -14,15 +14,15 @@
  */
 
 CREATE TABLE device_lists_changes_in_room (
-    stream_id BIGINT NOT NULL,
-    room_id TEXT NOT NULL,
     user_id TEXT NOT NULL,
     device_id TEXT NOT NULL,
+    room_id TEXT NOT NULL,
+    stream_id BIGINT NOT NULL,  -- This matches `device_lists_stream.stream_id`
```
richvdh, line 20: > -- This matches `device_lists_stream.stream_id`

can we enforce this with a constraint?

synapse/storage/databases/main/devices.py
```diff
@@ -1670,29 +1669,27 @@ def _add_device_outbound_room_poke_txn(
...
                     user_id,
                     device_id,
+                    room_id,
+                    stream_id,
                     True,  # As we're updating `device_lists_outbound_pokes` at the same time.
                     json_encoder.encode(context),
                 )
                 for room_id in room_ids
-                for device_id in device_ids
+                for device_id, stream_id in zip(device_ids, stream_ids)
```
richvdh, line 1730: can we stick some comments in here, and in the `delta` file, about how the stream_ids work? It seems like we get a different stream id for each device id, but there might be multiple rooms for each stream id?

synapse/storage/schema/main/delta/69/01device_list_oubound_by_room.sql
```diff
@@ -18,7 +18,11 @@ CREATE TABLE device_lists_changes_in_room (
     room_id TEXT NOT NULL,
     user_id TEXT NOT NULL,
     device_id TEXT NOT NULL,
+    -- We track if we've calculated the hosts in the room and updated
+    -- `device_lists_outbound_pokes`.
```
richvdh, line 22:
```suggestion
    -- We have a background process which goes through this table and converts
    -- entries into rows in `device_lists_outbound_pokes`. Once we have processed
    -- a row, we mark it as such by setting `converted_to_destinations=TRUE`.
```

synapse/storage/databases/main/devices.py
```diff
@@ -1693,3 +1693,82 @@ def _add_device_outbound_room_poke_txn(
...
+            self.db_pool.simple_update_txn(
+                txn,
+                table="device_lists_changes_in_room",
+                keyvalues={
+                    "user_id": user_id,
+                    "device_id": device_id,
+                    "stream_id": stream_id,
+                    "room_id": room_id,
+                },
+                updatevalues={"converted_to_destinations": True},
```
richvdh, line 1795: this might be a silly question, but why do we do it this way, instead of keeping track of the point we've processed up to?

synapse/storage/schema/main/delta/69/01device_list_oubound_by_room.sql
```diff
@@ -18,7 +18,11 @@ CREATE TABLE device_lists_changes_in_room (
...
+    converted_to_destinations BOOLEAN NOT NULL,
```
richvdh, line 23: remember that boolean columns need messing about in `synapse_port_db`

synapse/handlers/device.py
```diff
@@ -484,15 +490,18 @@ async def notify_device_update(
 
         room_ids = await self.store.get_rooms_for_user(user_id)
 
-        hosts: Set[str] = set()
-        if self.hs.is_mine_id(user_id):
-            for room_id in room_ids:
-                joined_users = await self.store.get_users_in_room(room_id)
-                hosts.update(get_domain_from_id(u) for u in joined_users)
+        hosts: Optional[Set[str]] = None
+        if self.use_new_device_lists_changes_in_room:
```
richvdh, line 494: is this the wrong way round? we should *skip* calculating hosts if `use_new_device_lists_changes_in_room`?

richvdh (changes requested):

synapse/storage/schema/main/delta/69/01device_list_oubound_by_room.sql
erikjohnston: Annoyingly we delete from `device_lists_stream` the next time the user updates their device, and I *think* we'll want to keep the rows in this table when that happens.

synapse/handlers/device.py
erikjohnston: Bah, yes. And this is the only place that uses it directly ─ everything else infers it from `hosts` being `None` ─ so nothing broke. 

I'm not sure if there is an easy way to test that this is the right way round?

synapse/storage/databases/main/devices.py
erikjohnston: Err, we could. I don't think I had much reasoning beyond not having to worry about extra tables or handling the duplicate stream IDs. 

I've also been thinking about how you'd distribute this across multiple workers, but I'm not sure this helps that or not :shrug:

synapse/storage/databases/main/devices.py
erikjohnston: The `max` got replaced, but I've added a comment the equivalent code

erikjohnston:
(The last three commits are bug fixes, rather than responding to review comments)

synapse/storage/databases/main/devices.py
```diff
@@ -1555,6 +1555,8 @@ async def add_device_change_to_streams(
 
         num_stream_ids = len(device_ids)
         if hosts:
+            # The `device_lists_outbound_pokes` wants a different stream ID for
+            # each row, which is a row per host per device update.
```
richvdh:
```suggestion
            # `device_lists_outbound_pokes` wants a different stream ID for
            # each row, which is a row per host per device update.
            #
            # We use the same set of stream_ids in both `device_lists_stream`
            #  and `device_lists_outbound_pokes`, so we need a total of
            # max(len(device_ids), len(hosts) * len(device_ids)).
```

synapse/storage/databases/main/devices.py
richvdh: mmmhmm. Generally I prefer to keep our tables append-only where possible (it saves a world of caching problems later on), but fair enough.

synapse/handlers/device.py
```diff
@@ -469,19 +488,26 @@ async def notify_device_update(
             # No changes to notify about, so this is a no-op.
             return
 
-        users_who_share_room = await self.store.get_users_who_share_room_with_user(
```
richvdh, line 472: FederationSenderDevicesTestCases still stubs this out. Is that still necessary?

synapse/handlers/device.py
richvdh: you might be able to block the `_handle_new_device_update_async` job in the tests (eg by setting `_handle_new_device_update_is_processing`), and check what actually gets written to the db?

Not sure it's worth it though.

synapse/storage/databases/main/devices.py
richvdh: oh well, apparently this is no longer correct.

synapse/storage/databases/main/devices.py
```diff
@@ -1546,34 +1554,62 @@ async def add_device_change_to_streams(
...
+                txn,
                 user_id,
                 device_ids,
                 hosts,
-                stream_ids,
+                stream_ids_for_outbound_pokes,
                 context,
             )
 
+        # The `device_lists_stream` wants a stream ID per device update.
```
richvdh, line 1594:
```suggestion
        # `device_lists_stream` wants a stream ID per device update.
```

synapse/storage/databases/main/devices.py
```diff
@@ -1546,34 +1554,62 @@ async def add_device_change_to_streams(
...
+        num_stream_ids = len(device_ids)
+
+        if hosts:
+            # The `device_lists_outbound_pokes` wants a different stream ID for
```
richvdh, line 1598:
```suggestion
            # `device_lists_outbound_pokes` wants a different stream ID for
```

richvdh (changes requested):

synapse/storage/schema/main/delta/69/01device_list_oubound_by_room.sql
```diff
@@ -0,0 +1,38 @@
...
+ */
+
+CREATE TABLE device_lists_changes_in_room (
+    user_id TEXT NOT NULL,
+    device_id TEXT NOT NULL,
+    room_id TEXT NOT NULL,
+
+    -- This initialy matches `device_lists_stream.stream_id`. Note that we
+    -- delete older values from `device_lists_stream`, so we can't use a foreign
+    -- constraint here.
```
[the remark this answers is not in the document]
erikjohnston: That commit is about `device_list_outbound_pokes`? Rather than `device_lists_stream` or `device_lists_changes_in_room`?

synapse/handlers/device.py
erikjohnston: I guess this code is going to be ripped out in the next release anyway

synapse/storage/databases/main/devices.py
erikjohnston: Ah yeah, it was surfacing a bug with the `get_all_device_list_changes_for_remotes` query. I thought this was how it worked previously, but that wasn't true so I've changed it to the old behaviour.

synapse/storage/schema/main/delta/69/01device_list_oubound_by_room.sql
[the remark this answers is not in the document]
richvdh: oh. confused. sorry.

richvdh (approved):
lgtm!

synapse/storage/databases/main/devices.py
```diff
@@ -1645,16 +1682,146 @@ def _add_device_outbound_poke_to_stream_txn(
...
+                    encoded_context,
+                )
+                for room_id in room_ids
+                for device_id, stream_id in zip(device_ids, stream_ids)
+            ],
+        )
+
+    async def get_uncoverted_outbound_room_pokes(
+        self, limit: int = 10
+    ) -> List[Tuple[str, str, str, int, Optional[Dict[str, str]]]]:
```
DMRobertson, line 1752: I think the return type here is incorrect: the optional dict should be Optional[str]. (Presumably we json-decode the string elsewhere when needed.)

Xref https://github.com/matrix-org/synapse/pull/12485#discussion_r854315184

Issue #12110, opened 2022-03-01
Add tests for or prevent non-member events in `state_events_at_start` in MSC2716 batch sending


tulir:
Synapse 1.53 broke this branch of batch sending: https://github.com/matrix-org/synapse/blob/v1.53.0/synapse/handlers/room_batch.py#L232-L253

#12083 seems to somehow fix it, but the feature should get some tests or be removed entirely.

MadLittleMods:
Closing as https://github.com/matrix-org/synapse/pull/12329 was merged to fix this use case in Synapse and Complement tests added in https://github.com/matrix-org/complement/pull/354

PR #12329, opened 2022-03-30
Allow non-member state sent in room batch to resolve for historic events (MSC2716)

MadLittleMods:
Allow non-member state sent in room batch to resolve for historic events. `/batch_send` `state_events_at_start`

Previously, non-member state didn't resolve because [`filter_events_for_client`](https://github.com/matrix-org/synapse/blob/e0bb2681340752f2716f1f386139ca37c53f26cd/synapse/visibility.py#L44) [filters out all `outlier` state](https://github.com/matrix-org/synapse/blob/e0bb2681340752f2716f1f386139ca37c53f26cd/synapse/visibility.py#L79-L81) except for out-of-band membership.

Part of https://github.com/matrix-org/synapse/issues/12110

Complement test: https://github.com/matrix-org/complement/pull/354

Part of [MSC2716](https://github.com/matrix-org/matrix-spec-proposals/pull/2716)

### Pull Request Checklist

### DIFF
--- a/changelog.d/12329.bugfix
+++ b/changelog.d/12329.bugfix
@@ -1,0 +1,1 @@
+Fix non-member state events not resolving for historical events when used in [MSC2716](https://github.com/matrix-org/matrix-spec-proposals/pull/2716) `/batch_send` `state_events_at_start`.
--- a/synapse/handlers/room_batch.py
+++ b/synapse/handlers/room_batch.py
@@ -156,8 +156,8 @@ class RoomBatchHandler:
     ) -> List[str]:
         """Takes all `state_events_at_start` event dictionaries and creates/persists
         them in a floating state event chain which don't resolve into the current room
-        state. They are floating because they reference no prev_events and are marked
-        as outliers which disconnects them from the normal DAG.
+        state. They are floating because they reference no prev_events which disconnects
+        them from the normal DAG.
 
         Args:
             state_events_at_start:
@@ -218,23 +218,18 @@ class RoomBatchHandler:
                     # The rest should hang off each other in a chain.
                     allow_no_prev_events=index == 0,
                     prev_event_ids=prev_event_ids_for_state_chain,
-                    # Since each state event is marked as an outlier, the
-                    # `EventContext.for_outlier()` won't have any `state_ids`
-                    # set and therefore can't derive any state even though the
-                    # prev_events are set. Also since the first event in the
-                    # state chain is floating with no `prev_events`, it can't
-                    # derive state from anywhere automatically. So we need to
-                    # set some state explicitly.
+                    # The first event in the state chain is floating with no
+                    # `prev_events` which means it can't derive state from
+                    # anywhere automatically. So we need to set some state
+                    # explicitly.
                     #
                     # Make sure to use a copy of this list because we modify it
                     # later in the loop here. Otherwise it will be the same
-                    # reference and also update in the event when we append later.
+                    # reference and also update in the event when we append
+                    # later.
                     state_event_ids=state_event_ids.copy(),
                 )
             else:
-                # TODO: Add some complement tests that adds state that is not member joins
-                # and will use this code path. Maybe we only want to support join state events
-                # and can get rid of this `else`?
                 (
                     event,
                     _,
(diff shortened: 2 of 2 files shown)


synapse/handlers/room_batch.py
```diff
@@ -213,9 +213,6 @@ async def persist_state_events_at_start(
                     room_id=room_id,
                     action=membership,
                     content=event_dict["content"],
-                    # Mark as an outlier to disconnect it from the normal DAG
-                    # and not show up between batches of history.
-                    outlier=True,
```
MadLittleMods, line 218: Removing these as `outliers` because [`filter_events_for_client`](https://github.com/matrix-org/synapse/blob/e0bb2681340752f2716f1f386139ca37c53f26cd/synapse/visibility.py#L44) [filters out all `outlier` state](https://github.com/matrix-org/synapse/blob/e0bb2681340752f2716f1f386139ca37c53f26cd/synapse/visibility.py#L79-L81) except for out-of-band membership.

This state chain is still disconnected from the normal DAG because the first event has empty `prev_events` so they state just floats in a chain of its own.

synapse/handlers/room_batch.py
```diff
@@ -213,9 +213,6 @@ async def persist_state_events_at_start(
                     room_id=room_id,
                     action=membership,
                     content=event_dict["content"],
-                    # Mark as an outlier to disconnect it from the normal DAG
```
richvdh, line 216: there's some text in the docstring for this function:

> ... are marked as outliers which disconnects them from the normal DAG.

which sounds like it is no longer correct?

synapse/handlers/room_batch.py
richvdh: I'm struggling to understand how the diagram relates to the change here. Which of the events on the diagram correspond to this code path?

richvdh (commented):
generally not setting these things as outliers seem sensible, but I'm not quite following the justification.

synapse/handlers/room_batch.py
MadLittleMods: > I'm struggling to understand how the diagram relates to the change here. Which of the events on the diagram correspond to this code path?

Just giving it as a reference. It shows that the state chain floats by no `prev_events` and that the state chain is disconnected from the historical message chain.

synapse/handlers/room_batch.py
MadLittleMods: Good call, updated the spots that reference `outlier`

changelog.d/12329.bugfix
```diff
@@ -0,0 +1 @@
+Fix non-member state events not resolving for historical events when used in [MSC2716](https://github.com/matrix-org/matrix-spec-proposals/pull/2716) `/batch_send` `state_events_at_start`.
```
MadLittleMods, line 1: > generally not setting these things as outliers seem sensible, but I'm not quite following the justification.

Shout with any more clarification questions 👍

synapse/handlers/room_batch.py
richvdh: ok, so this code is handling the events in the yellow boxes labelled `batch0`, `batch1`, `batch2` ?

MadLittleMods: `persist_state_events_at_start` is handling the state events **above the yellow boxes** (`m.room.member` events in the diagram). The historical state chain.

The yellow boxes are the historical messages chains

MadLittleMods:
> Fundamentally, we have the state at these events so they shouldn't be marked as outliers, so this seems correct.

This is a great point too!  Wasn't thinking in that context.

Thanks for the review @richvdh 🐂

synapse/handlers/room_batch.py
richvdh: > persist_state_events_at_start is handling the state events above the yellow boxes (m.room.member events in the diagram). The historical state chain.

well in that case, this seems wrong. those events seem to be random state events where we haven't wired them up to the event dag? why would we have the state at those events?

Either I'm severely understanding your diagram, or something is odd here.

MadLittleMods: > well in that case, this seems wrong. those events seem to be random state events where we haven't wired them up to the event dag?

Correct, they float in their own chain.

> why would we have the state at those events?

But we technically do have the state for them because we're manually providing it via `state_event_ids`

PR #12330, opened 2022-03-30
Disable proactive sends for remote joins

richvdh:
Do not attempt to send remote joins out over federation. Normally, it will do
nothing; occasionally, it will do the wrong thing.

### DIFF
--- a/synapse/handlers/federation_event.py
+++ b/synapse/handlers/federation_event.py
@@ -469,6 +469,12 @@ class FederationEventHandler:
             if context.rejected:
                 raise SynapseError(400, "Join event was rejected")
 
+            # the remote server is responsible for sending our join event to the rest
+            # of the federation. Indeed, attempting to do so will result in problems
+            # when we try to look up the state before the join (to get the server list)
+            # and discover that we do not have it.
+            event.internal_metadata.proactively_send = False
+
             return await self.persist_events_and_notify(room_id, [(event, context)])
 
     async def backfill(
--- a/changelog.d/12330.misc
+++ b/changelog.d/12330.misc
@@ -1,0 +1,1 @@
+Avoid trying to calculate the state at outlier events.


synapse/handlers/federation_event.py
```diff
@@ -469,6 +469,12 @@ async def process_remote_join(
             if context.rejected:
                 raise SynapseError(400, "Join event was rejected")
 
+            # the remote server is responsible for sending our join event to the rest
+            # of the federation. Indeed, attempting to do so will result in problems
+            # when we try to look up the state before the join (to get the server list)
+            # and discover that we do not have it.
+            event.internal_metadata.proactively_send = False
+
```
DMRobertson, line 477: > Indeed, attempting to do so will result in problems

are these problems local to us only, or are there consequences for the rest of the federation? (i.e. does this scenario need to be called out in the spec?)

DMRobertson (commented):
I wonder if `proactively_send` should be opt-in rather than opt-out? (I'm guessing from the title that this is one example of a class of similar bugs.)

DMRobertson (approved):

synapse/handlers/federation_event.py
richvdh: the main "problem" I'm trying to avoid is that I want to make getting the state at an outlier raise an exception (#12191), which is definitely a "local to us" problem.

The [spec](https://spec.matrix.org/unstable/server-server-api/#joining-rooms) has this to say:

> The **resident** server must also send the event to other servers participating in the room.

(my emphasis)

An example of a less local-to-us problem is that we might actually succeed in looking up the state, and then send another copy of the join event to the whole federation. Still, in the worst case that's just duplicate data, and isn't really any different from us deciding to, say, send out an event generated by another server.

In short - I think the spec says all it needs to?

richvdh:
> I wonder if `proactively_send` should be opt-in rather than opt-out?

not convinced. There really aren't many occasions where the `sender` of an event is local, yet we don't want to send it out to any servers in the room.

PR #12332, opened 2022-03-30
Handle outliers in `/federation/v1/event`

richvdh:
The intention here is to avoid doing state lookups for outliers in `/_matrix/federation/v1/event`. Unfortunately that's expanded into something of a rewrite of `filter_events_for_server`, which ended up trying to do that operation in a couple of places.

First of all, note that there is a long-standing fudge here, where we return outliers via `/_matrix/federation/v1/event` to any server that is *currently* in the room, even though we don't know the `history_visibility` in force at the time of the event, nor if the server was in the room at that time (which are normally prerequisites for allowing this). This *has* to be the case, because the federation protocol relies on being able to pull `auth_events` for a given event in this manner (and we have no way to show that the server should *not* be able to see those events).

It currently works somewhat via the back door, because `get_state_ids_for_events` returns an empty set for outliers; that equates to a missing `m.room.history_visibility` event, which equates to `shared`, ie "visible to any server which is now in the room".

`filter_events_for_server` does *two* state lookups: first of all it just looks up the history_visibility events. If those are *all* unrestricted, there is a fast path for it to bail out. Otherwise, it does a *second* state lookup, including (currently) both history_visibility events *and* membership events.

Rather than apply the same special-casing for outliers twice, I've refactored the code so that, rather than having a separate fast path, I've changed the "slow" path so that it only looks up membership events for events where it has already decided that there is a restricted history visibility. That therefore side-steps the outlier issue for that second state lookup.

### DIFF
--- a/synapse/visibility.py
+++ b/synapse/visibility.py
@@ -381,8 +381,7 @@ async def _event_to_history_vis(
     visibility_ids = {
         vis_event_id
         for vis_event_id in (
-            state_ids.get(_HISTORY_VIS_KEY)
-            for state_ids in event_to_state_ids.values()
+            state_ids.get(_HISTORY_VIS_KEY) for state_ids in event_to_state_ids.values()
         )
         if vis_event_id
     }
--- a/synapse/visibility.py
+++ b/synapse/visibility.py
@@ -1,4 +1,5 @@
 # Copyright 2014 - 2016 OpenMarket Ltd
+# Copyright (C) The Matrix.org Foundation C.I.C. 2022
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -12,7 +13,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import logging
-from typing import Dict, FrozenSet, List, Optional
+from typing import Collection, Dict, FrozenSet, List, Optional, Tuple
+
+from typing_extensions import Final
 
 from synapse.api.constants import EventTypes, HistoryVisibility, Membership
 from synapse.events import EventBase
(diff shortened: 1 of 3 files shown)


synapse/visibility.py
```diff
@@ -293,67 +296,28 @@ def is_sender_erased(event: EventBase, erased_senders: Dict[str, bool]) -> bool:
...
+            if memtype == Membership.JOIN:
+                return True
+            elif memtype == Membership.INVITE:
+                if visibility == HistoryVisibility.INVITED:
+                    return True
+
+        # server has no users in the room: redact
+        return False
 
     if not check_history_visibility_only:
```
DMRobertson, line 322: Struggling to follow this. I think github's choice of diff algorithm hasn't made it easy to understand how the bits got moved around.

Before: this ran after the computation of `event_to_state_ids`, `visibility_ids` and `all_open`.

After: this runs _before_ the computation of `event_to_memberships`, `event_to_history_vis`. `all_open` is no more.

Is there any significance to the reordering? I can't see any, but would appreciate confirmation

synapse/visibility.py
```diff
@@ -362,34 +326,104 @@ def check_event_is_visible(event: EventBase, state: StateMap[EventBase]) -> bool
...
+    )
+
+    visibility_ids = {
+        vis_event_id
+        for vis_event_id in (
+            state_ids.get((EventTypes.RoomHistoryVisibility, ""))
+            for state_ids in event_to_state_ids.values()
+        )
+        if vis_event_id
+    }
```
DMRobertson, line 388: Found the nested comprehensions tricky to read, but I don't think I can see a neater way to write it.

synapse/visibility.py
```diff
@@ -405,36 +439,26 @@ def check_event_is_visible(event: EventBase, state: StateMap[EventBase]) -> bool
         for key, event_id in key_to_eid.items()
     }
 
-    def include(typ, state_key):
-        if typ != EventTypes.Member:
-            return True
-
+    def include(state_key):
```
DMRobertson, line 442:
```suggestion
    def include(state_key: str) -> bool:
```

synapse/visibility.py
```diff
@@ -405,36 +439,26 @@ def check_event_is_visible(event: EventBase, state: StateMap[EventBase]) -> bool
...
             return False
         return state_key[idx + 1 :] == server_name
 
     event_map = await storage.main.get_events(
-        [e_id for e_id, key in event_id_to_state_key.items() if include(key[0], key[1])]
+        [
+            e_id
+            for e_id, (_, state_key) in event_id_to_state_key.items()
+            if include(state_key)
+        ]
```
DMRobertson, line 454: Sanity check: for this to be safe, event_id_to_state_key should only contain membership events. That's true by construction on +426.

synapse/visibility.py
```diff
@@ -362,34 +326,104 @@ def check_event_is_visible(event: EventBase, state: StateMap[EventBase]) -> bool
...
+    Returns a map from event id to history_visibility setting
+    """
+
+    # outliers get special treatment here. We don't have the state at that point in the
+    # room (and attempting to look it up will raise an exception), so all we can really
+    # do is assume that the requesting server is allowed to see the event. That's
+    # equivalent to there not being a history_visibility event, so we just exclude
+    # any outliers from the query.
+    event_to_state_ids = await storage.state.get_state_ids_for_events(
+        frozenset(e.event_id for e in events if not e.internal_metadata.is_outlier()),
```
DMRobertson, line 377: We don't ignore outliers below on +425. Are the events passed to `_event_to_membership` are never outliers?

DMRobertson (commented):
This was tricky to follow. I think the rewrite looks sane, and excluding the outliers seems to be a one line change (+377?).

WIth that said:

synapse/visibility.py
```diff
@@ -293,67 +296,28 @@ def is_sender_erased(event: EventBase, erased_senders: Dict[str, bool]) -> bool:
...
-            e.content.get("history_visibility")
-            in (None, HistoryVisibility.SHARED, HistoryVisibility.WORLD_READABLE)
-            for e in event_map.values()
-        )
+        # We now loop through all membership events looking for
+        # membership states for the requesting server to determine
+        # if the server is either in the room or has been invited
+        # into the room.
+        for ev in memberships.values():
+            assert get_domain_from_id(ev.state_key) == server_name
```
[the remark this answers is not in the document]
richvdh: > Oh, it's because `_event_to_memberships` is going to spew out membership events for that server only. So this is just a defensive assertion?

indeed. We should never see membership events from other servers (and that's not new, so I don't really know why we had the `continue` before).

synapse/visibility.py
richvdh: No, there's no particular significance to the reordering, other than that it seemed to make sense to me to put `_event_to_history_vis` and `_event_to_memberships` next to each other. The functionality should be much the same.

synapse/visibility.py
richvdh: yes, that's the main intention of this PR.

outliers will always get a result of `HistoryVisibility.SHARED` from `_event_to_history_vis`, which means they will be filtered out by the list comprehension in the args to `_event_to_membership` (line +341).

synapse/visibility.py
```diff
@@ -362,34 +326,104 @@ def check_event_is_visible(event: EventBase, state: StateMap[EventBase]) -> bool
...
+    # any outliers from the query.
+    event_to_state_ids = await storage.state.get_state_ids_for_events(
+        frozenset(e.event_id for e in events if not e.internal_metadata.is_outlier()),
+        state_filter=StateFilter.from_types(types=(_HISTORY_VIS_KEY,)),
+    )
+
+    visibility_ids = {
+        vis_event_id
+        for vis_event_id in (
+            state_ids.get((EventTypes.RoomHistoryVisibility, ""))
```
richvdh, line 384:
```suggestion
            state_ids.get(_HISTORY_VIS_KEY)
```

synapse/visibility.py
richvdh: Yeah. We could build the set incrementally:

```python
visibility_ids = set()
for state_ids in event_to_state_ids.values():
    vis_event_id = state_ids.get(_HISTORY_VIS_KEY)
    if vis_event_id:
        visibility_ids.add(vis_event_id)
```

... but I'm assuming (without checking) that the comprehension is faster.

richvdh:
@DMRobertson: Thanks for ploughing through this. Certainly it's not easy to follow from the diff. I've tried to answer your questions, though it looks like you've already grokked it pretty well.

synapse/visibility.py
DMRobertson: Does this also want to be applied to line 80 and 163? (Would need the constant to be moved higher in the file I think?)

Issue #12328, opened 2022-03-29
Using URL component `port` in `url_preview_url_blacklist` leads to `TypeError`


lxp:
### Description

If `url_preview_url_blacklist` uses the `port` URL component an `TypeError` is thrown.

### Steps to reproduce

- Use an `url_preview_url_blacklist` with the URL component `port` in your homeserver.yaml.

```yml
url_preview_url_blacklist:
  # blacklist any URL with an explicit port
  - port: '*'
```

- When an URL preview for an URL explicit port set is requested, the following exception is logged:

```
2022-03-29 18:49:55,029 - synapse.http.server - 100 - ERROR - GET-685 - Failed handle request via 'PreviewUrlResource': <XForwardedForRequest at 0x7fa7fb6a1d00 method='GET' uri='/_matrix/media/r0/preview_url?url=https%3A%2F%2Fgoogle.com%3A1234%2F&ts=1648572540000' clientproto='HTTP/1.0' site='8008'>
Traceback (most recent call last):
File "/usr/local/lib/python3.9/site-packages/synapse/http/server.py", line 268, in _async_render_wrapper
callback_return = await self._async_render(request)
File "/usr/local/lib/python3.9/site-packages/synapse/http/server.py", line 296, in _async_render
callback_return = await raw_callback_return
File "/usr/local/lib/python3.9/site-packages/synapse/rest/media/v1/preview_url_resource.py", line 212, in _async_render_GET
if not fnmatch.fnmatch(getattr(url_tuple, attrib), pattern):
File "/usr/local/lib/python3.9/fnmatch.py", line 40, in fnmatch
name = os.path.normcase(name)
File "/usr/local/lib/python3.9/posixpath.py", line 54, in normcase
return os.fspath(s)
TypeError: expected str, bytes or os.PathLike object, not int
```

- The same also applies to a regex configuration leading to a slightly different exception:

```yml
url_preview_url_blacklist:
  # blacklist any URL with an explicit port
  - port: '^.*$'
```

```
2022-03-29 18:52:00,582 - synapse.http.server - 100 - ERROR - GET-9 - Failed handle request via 'PreviewUrlResource': <XForwardedForRequest at 0x7f33e13238b0 method='GET' uri='/_matrix/media/r0/preview_url?url=https%3A%2F%2Fgoogle.com%3A1234%2F&ts=1648572720000' clientproto='HTTP/1.0' site='8008'>
Traceback (most recent call last):
[...]

PR #688, opened 2016-04-03
URL previewing support

ara4n:
* add design sketch doc for a URL preview API
 * add a new `SpiderHttpClient` derived from `SimpleHttpClient`, which follows redirects and handles gzip CTE correctly
 * add `get_file` support to `SimpleHttpClient`, knowingly duplicated for now from matrixfederationclient.
 * add a `preview_url_resource` to implement the new media/r0/preview_url API.  This:
   * spiders the a given URL, extracting or synthesising OpenGraph metadata for it via `lxml`, returning the metadata as a JSON blob
   * returns a cache (either from on disk or in memory) for the metadata of the URL as of the requested point in time, if available.
   * deduplicates requests to spider a URL such that only one req is in flight at any point.
 * adds support for thumbnailing SVGs (by just passing back the original image for now)
 * adds the `local_media_repository_url_cache` table to the DB for the on-disk URL cache
 * adds `get_url_cache` and `store_url_cache` to `media_repository.py` to wrap the new table

N.B. that following redirects will not work correctly until https://twistedmatrix.com/trac/ticket/8265 is merged.  Unsure if it's worth maintaining our own Twisted fork until that happens.

Given I'm hardly a python/twisted expert, review would be particularly appreciated on:
 * Whether I should be doing anything with log contexts - I've seen some worrying warnings about log contexts not being preserved
 * Whether I should be doing anything smarter with the types or flavours of exceptions I raise. Particularly, some exceptions from the async_get method seem to lose their stack traces entirely - i haven't spotted the pattern yet.

This is part of a set of PRs spanning vector-web, matrix-react-sdk, matrix-js-sdk and synapse.
See also vector-im/vector-web#1343 and matrix-org/matrix-react-sdk#260 and https://github.com/matrix-org/matrix-js-sdk/pull/122

PR #12333, opened 2022-03-30
Ensure the type of URL attributes is always str when matching against preview blacklist

babolivier:
Fixes https://github.com/matrix-org/synapse/issues/12328

Introduced in https://github.com/matrix-org/synapse/pull/688


babolivier:
Removing the request for review as this needs a test.

DMRobertson (approved):
It's sad that mypy couldn't detect this. It knows what `url_tuple` is; but the getattr call on 191 means that `value: Any` and so there's no hope of meaningful inference from that point onward.

PR #12334, opened 2022-03-30
Nuke the tox `packaging` job

DMRobertson:
It only checks the manifest. Once #11537 lands there won't be a manifest
any more.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -27,7 +27,6 @@ jobs:
           - "check_codestyle"
           - "check_isort"
           - "mypy"
-          - "packaging"
 
     steps:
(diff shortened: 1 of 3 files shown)


reivilibre (approved):
I suppose the only question I have is: Is there an equivalent for Poetry? (Is it worth it?)

DMRobertson:
> Is there an equivalent for Poetry?

Not that I'm aware of. We could write our own tool for this (compare pyproject.toml include/exclude directives against source control), but I don't think there's a great need for it. (And I think such a tool would be better placed in poetry itself).

In #12337 I'll produce diffs between sdists wheels which should hopefully give us enough confidence.

PR #12335, opened 2022-03-30
Ignore .envrc for direnv users

DMRobertson:
`direnv` + either 
- `layout python` or 
- `layout poetry` with config as described in https://github.com/direnv/direnv/wiki/Python#poetry 

works well for me. Its configuration file is `.envrc`. I happen to know @reivilibre uses it too and finds is a pleasurable.

I haven't committed an `.envrc` file because I don't want to
introduce clutter for people who aren't using direnv, and because I
suppose in principle someone could choose to use a different layout.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,6 +30,9 @@ __pycache__/
 /media_store/
 /uploads
 
+# For direnv users
+/.envrc
+
 # IDEs
 /.idea/
 /.ropeproject/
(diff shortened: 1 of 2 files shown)


reivilibre (approved):
Thanks!

Just out of interest, you can add these entries to `.git/info/exclude` if working on a repository that doesn't have it ignored, but I think it's nicer to have it built in.

PR #12336, opened 2022-03-30
Remove `dockerfile-pgtests`

DMRobertson:
AFAICS these aren't used any more. They're going to break when I do
poetry work (#11537), because tox won't (easily) recreate the locked
poetry environment. I think it's less effort to throw it away than to
fix it.


anoadragon453 (approved):
I chuckled at the branch name.

Issue #12107, opened 2022-02-28
Move scripts to Synapse package and expose as `entry_point`s; add suffices to scripts-dev


DMRobertson:
`black` and `flake8` will ignore such files by default. flake8 has a `filename` config option, but I haven't been able to use this to have flake8 run over the scripts it would otherwise script.

`mypy` does not have this problem: one can specify an explicit list of files to example or ignore. `isort` doesn't have such configuration, but does look for shebang lines to detect python scripts without a .py suffix.

Given a magic wand, I would like to:
- rename all scripts to end in .py
- change the documentation to include the new .py suffix
- change the debian package build so that the new scripts are available by the old, suffixless filename

Bullet three here is the hard one.

This would allow me to run `black .` and `flake8 .` both locally and in CI without thinking.

DMRobertson:
> * change the debian package build so that the new scripts are available by the old, suffixless filename
> 
> 
> Bullet three here is the hard one.

Aha! This is done in `matrix-synapse-py3.links`.

richvdh:
I am not enthusiastic about naming commands according to the language they happen to be written in. It feels a bit like Hungarian notation. Linux prefers `#!` lines over file extensions for a reason.

However, Setuptools allows an entry point to be defined in a file named `*.py` and still get installed into the virtualenv with its regular name - see https://setuptools.pypa.io/en/latest/userguide/entry_point.html.

It looks like poetry's [scripts](https://python-poetry.org/docs/pyproject/#scripts) do much the same thing.

DMRobertson:
An alternative would be to keep passing around the list of explicit targets to lint, both in CI and in lint.sh. That's fine and is probably easier---but it would mean maintaining duplicate lists, which is the kind of thing I'm trying to move us away from with poetry.

On entry_points and scripts: I'm glad you brought that up. When I looked at `register_new_matrix_user` I thought "this could just be an entry point".

https://github.com/matrix-org/synapse/blob/4b965c862dc66c0da5d3240add70e9b5f0aa720b/scripts/register_new_matrix_user#L16-L19

I think this script was moved to within Synapse (#4085) in order to be tested here:

https://github.com/matrix-org/synapse/blob/4b965c862dc66c0da5d3240add70e9b5f0aa720b/tests/scripts/test_new_matrix_user.py#L17

[...]

PR #12337, opened 2022-03-30
[WIP] Use Poetry to manage Synapse's dependencies

### LINKED ISSUE #11537: Lock Dependencies in Synapse

DMRobertson:
Reviewable commit-by-commit. Hopefully. Ignoring the lockfile, the diffstats for each individual commit should hopefully be managable.

I'd appreciate eyes on the output from CI. I'd like to ensure that it's still reasonably easy to follow, and clear that we're using the poetry locked environment rather than pip-installing the newest version of everything.

This set of changes:

- uses poetry rather than setuptools to declare Synapse's project metadata and dependencies;
- updates the docker image and debian package to use dependency versions managed in the poetry lockfile;
- runs CI in the locked poetry environment; and
- updates docs to strongly recommend using `poetry` for development purposes.

Included at the start are a couple of DEBUG commits which will ensure that the CI run for 
https://github.com/matrix-org/synapse/pull/12337/commits/5f8db3ed0e8b062cb88fb0ce06f08240e749cfcf will build _all_ debian packages and run the twisted CI job. Before this is merged, they'll need to be undone.

Closes #11537.

TODO:
- [ ] bump version number to appropriate version before merging,
- [x] compare the wheels and sdists generated on develop and on this branch.
- [x] CI that pip installs the latest released version of _everything_, to anticipate problems from people installing from wheels. (We'll no longer get notifications of this via our CI). This is probably best done in a separate PR.


debian/changelog
```diff
@@ -1,3 +1,10 @@
+matrix-synapse-py3 (1.55.2+nmu1) UNRELEASED; urgency=medium
+
+  * Non-maintainer upload.
+  * Use poetry to manage the bundled virtualenv included with this package.
```
DMRobertson, line 4: I have no idea if this is appropriate.

docker/Dockerfile-dhvirtualenv
```diff
@@ -85,6 +85,15 @@ RUN apt-get update -qq -o Acquire::Languages=none \
         libpq-dev \
         xmlsec1
 
+RUN pip3 install -U \
+    pip \
+    setuptools \
+    wheel
```
DMRobertson, line 91: Shay says: debian buster needed a newer version of pip to pull in wheels rather than build  from source. Not sure if an upgrade to setuptools is required here.

pyproject.toml
```diff
@@ -65,4 +50,235 @@ known_twisted = ["twisted", "OpenSSL"]
 multi_line_output = 3
 include_trailing_comma = true
 combine_as_imports = true
+skip_gitignore = true
 
+[tool.poetry]
+name = "matrix-synapse"
+version = "1.54.0rc1"
```
DMRobertson, line 57: This is now out-of-date.

.github/workflows/tests.yml
```diff
@@ -129,18 +119,18 @@ jobs:
           || true
 
   trial-olddeps:
+    # Note: sqlite only; no postgres
     if: ${{ !cancelled() && !failure() }} # Allow previous steps to be skipped, but not fail
-    needs: linting-done
+#    needs: linting-done
```
DMRobertson, line 124: Debug that needs to be undone.

pyproject.toml
DMRobertson: I guess we should aim this at the 1.57 release? Would be nice to mark this as a prerelease somehow if possible. Can't remember the right notation for this though, will have to check https://peps.python.org/pep-0440/

.ci/scripts/test_old_deps.sh
```diff
@@ -9,12 +12,57 @@ set -ex
...
+# delete all lines referring to psycopg2 --- so no testing of postgres support
+# Omit systemd: we're not logging to journal here.
+
+sed -i-backup \
+   -e "s/[~>]=/==/g" \
+   -e "/psycopg2/d" \
+   -e 's/pyOpenSSL = "==16.0.0"/pyOpenSSL = "==17.0.0"/' \
+   -e '/psycopg2/d' \
+   -e '/systemd/d' \
+   pyproject.toml
```
squahtx, line 41: Is the double `/psycopg2/d` a typo?

.github/workflows/tests.yml
```diff
@@ -20,20 +20,11 @@ jobs:
...
-
-    steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-python@v2
-      - run: pip install tox
-      - run: tox -e ${{ matrix.toxenv }}
+    # This does a vanilla `poetry install` - no extras. I'm slightly anxious
+    # that we might skip some typechecks on code that uses extras. However,
+    # I think the right way to fix this is to turn up mypy's strictness:
+    # disallow unknown imports and be accept fewer uses of `Any`.
```
squahtx, line 26:
```suggestion
    # disallow unknown imports and accept fewer uses of `Any`.
```

README.rst
```diff
@@ -293,39 +293,41 @@ directory of your choice::
     git clone https://github.com/matrix-org/synapse.git
     cd synapse
 
-Synapse has a number of external dependencies, that are easiest
-to install using pip and a virtualenv::
+Synapse has a number of external dependencies. We maintain a fixed development
+environment using [poetry](). First, install poetry. We recommend
```
squahtx, line 297: is this missing a url?

squahtx (commented):
skimmed the changes - would appreciate more eyes on this!

.github/workflows/twisted_trunk.yml
```diff
@@ -7,31 +7,35 @@ on:
   workflow_dispatch:
   pull_request:
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
```
babolivier, line 11: Can you add a comment explaining why we need this please?

README.rst
babolivier, line 297:
```suggestion
environment using [poetry](https://python-poetry.org/). First, install poetry. We recommend
```

README.rst
```diff
@@ -293,39 +293,41 @@ directory of your choice::
...
-    python -m synapse.app.homeserver \
+    poetry run python -m synapse.app.homeserver \
       --server-name my.domain.name \
       --config-path homeserver.yaml \
       --generate-config \
       --report-stats=[yes|no]
 
     # Start the app
-    python -m synapse.app.homeserver --config-path homeserver.yaml
+    poetry run python -m synapse.app.homeserver --config-path homeserver.yaml
```
babolivier, line 330: Shouldn't we use `poetry run synapse_homeserver [...]` here?

docs/code_style.md
```diff
@@ -6,13 +6,8 @@ The Synapse codebase uses a number of code formatting tools in order to
...
 
-The necessary tools are detailed below.
-
-First install them with:
-
-```sh
-pip install -e ".[lint,mypy]"
-```
+The necessary tools are detailed below. They should be installed by poetry as part
+of Synapse's dev dependencies. See [the contributing guide](https://matrix-org.github.io/synapse/develop/development/contributing_guide.html#4-install-the-dependencies) for instructions.
```
babolivier, line 10:
```suggestion
of Synapse's dev dependencies. See [the contributing guide](https://matrix-org.github.io/synapse/latest/development/contributing_guide.html#4-install-the-dependencies) for instructions.
```

docs/development/contributing_guide.md
```diff
@@ -48,16 +48,16 @@ can find many good git tutorials on the web.
 
 # 4. Install the dependencies
 
-Once you have installed Python 3 and added the source, please open a terminal and
-setup a *virtualenv*, as follows:
+Synapse uses the [poetry](https://python-poetry.org/) project to lock its dependencies
```
babolivier, line 51:
```suggestion
Synapse uses the [poetry](https://python-poetry.org/) project to manage its dependencies
```

This sounds more natural to me, especially since it's supposed to also apply to "development environment".

docs/upgrade.md
```diff
@@ -19,32 +19,35 @@ this document.
...
-        If Synapse was installed using git then upgrade to the latest
-        version by running:
-
+    3.  If you're using `poetry` to manage a Synapse installation, run:
         ```bash
-        git pull
-        pip install --upgrade .
+        poetry install
         ```
+        Include any relevant extras with `-E`, e.g. `poetry install -E "postgres oidc"`.
```
babolivier, line 48: I think we should use `--extras` here instead for consistency.

docs/upgrade.md
```diff
@@ -19,32 +19,35 @@ this document.
...
+
+    2.  If you're running Synapse in a virtualenv, make sure to activate it before
+        upgrading. For example, if Synapse is installed in a virtualenv in `~/synapse/env` then
         run:
 
         ```bash
         source ~/synapse/env/bin/activate
+        pip install --upgrade .
         ```
+        Include any relevant extras with `-e`, e.g. `pip install --upgrade .[postgres,oidc]`.
```
babolivier, line 42:
```suggestion
        Include any relevant extras between square brackets, e.g. `pip install --upgrade .[postgres,oidc]`.
```

babolivier (changes requested):
Thanks a lot for working on this! Really looking forward to seeing this PR land.

docker/Dockerfile-dhvirtualenv
babolivier: > Not sure if an upgrade to setuptools is required here.

Why would it?

.ci/scripts/test_old_deps.sh
DMRobertson: yes, thanks.

README.rst
DMRobertson: absolutely!

.ci/scripts/test_old_deps.sh
DMRobertson:
```suggestion
```

.ci/scripts/test_old_deps.sh
```diff
@@ -9,12 +12,57 @@ set -ex
...
+# but make the pyopenssl 17.0, which can work against an
+# OpenSSL 1.1 compiled cryptography (as older ones don't compile on Travis).
+# delete all lines referring to psycopg2 --- so no testing of postgres support
+# Omit systemd: we're not logging to journal here.
+
+sed -i-backup \
+   -e "s/[~>]=/==/g" \
+   -e "/psycopg2/d" \
+   -e 's/pyOpenSSL = "==16.0.0"/pyOpenSSL = "==17.0.0"/' \
+   -e '/psycopg2/d' \
```
DMRobertson, line 39:
```suggestion
```

synapse/__init__.py
```diff
@@ -68,7 +70,7 @@
 except ImportError:
     pass
 
-__version__ = "1.55.2"
+__version__ = get_distribution_version_string("matrix-synapse")
```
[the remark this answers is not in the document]
DMRobertson: Agreed, but I think it'll be easier to fixup the release script when it comes to a release with this change landed.

docs/code_style.md
DMRobertson: I think developers are typically going to be developing based on `develop` rather than `master`.

.github/workflows/twisted_trunk.yml
```diff
@@ -56,11 +60,18 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Patch dependencies
-        run: .ci/patch_for_twisted_trunk.sh
+        run: >-
+          ln -s -T /venv /src/.venv &&
+          .ci/patch_for_twisted_trunk.sh &&
+          unlink /src/.venv
```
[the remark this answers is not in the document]
squahtx: This one's on me for not adding a comment.

`/src` is a volume mount of the repo (`- ${{ github.workspace }}:/src` above).
`.ci/patch_for_twisted_trunk.sh` invokes poetry, which has been configured to look for a virtualenv at `./.venv`. But the sytest-synapse container wants control over the location of the virtualenv and has it at `/venv`. So we cheekily symlink it in to make poetry happy, then clean up after ourselves.

```suggestion
          # `.ci/patch_for_twisted_trunk.sh` invokes poetry, which looks for a
          # virtualenv in `/src/.venv`. Set up the virtualenv for poetry, run
          # the script, then tidy up after ourselves. `/src` is a mounted
          # volume, so it's impolite to be making changes to it.
          ln -s -T /venv /src/.venv &&
          .ci/patch_for_twisted_trunk.sh &&
          unlink /src/.venv
```

.ci/patch_for_twisted_trunk.sh
```diff
@@ -1,8 +1,22 @@
...
+# We could make full use of the poetry machinery (rather than just `pip install ...`)
+# so that we can catch dependency resolver problems that could arise from twisted
+# bumping its dependencies...
+# sed -ibackup -e 's!^Twisted = .*!Twisted = { git = "https://github.com/twisted/twisted.git", rev = "trunk" }!' pyproject.toml
+# poetry lock --no-update
+# poetry install --no-interaction --extras "all test"
+
+# ...except we run into https://github.com/python-poetry/poetry/issues/5311, where
+# poetry insists on installing an old version of treq, which isn't actually compatible
+# with recent twisted releases. So let's just install twisted trunk using pip.
```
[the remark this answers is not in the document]
DMRobertson: No; that comment arrived a little after we settled on using pip. (And as Sean writes above, it would only catch problems which arise from Twisted bumping its dependencies, which seems unlikely at this point in Twisted's lifecycle.)

.github/workflows/twisted_trunk.yml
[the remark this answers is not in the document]
babolivier: Makes sense, thanks for explaining!

.ci/scripts/test_old_deps.sh
```diff
@@ -9,12 +12,56 @@ set -ex
...
+
+# patch the project definitions in-place
+# replace all lower bounds with exact bounds
+# but make the pyopenssl 17.0, which can work against an
+# OpenSSL 1.1 compiled cryptography (as older ones don't compile on Travis).
+# delete all lines referring to psycopg2 --- so no testing of postgres support
+# Omit systemd: we're not logging to journal here.
+
+sed -i-backup \
+   -e "s/[~>]=/==/g" \
```
DMRobertson, line 36: This character class should include a caret to account for caret requirements: https://python-poetry.org/docs/dependency-specification#caret-requirements

docs/code_style.md
babolivier: Fair enough.

.github/workflows/twisted_trunk.yml
DMRobertson: This is a fairly standard snippet which means that this workflow will only run once for a given PR. See e.g.

https://github.com/matrix-org/sytest/blob/2780655ce1aa81d06e32c34eba758054eeb69a56/.github/workflows/pipeline.yml#L7-L12

Also used in synapse tests.yml.

That said, we don't need this concurrency blob here once we're back to using a cronjob. I'll remove it later.

docker/Dockerfile
```diff
@@ -46,7 +46,7 @@ RUN \
     && rm -rf /var/lib/apt/lists/*
 
 # Copy just what we need to pip install
-COPY MANIFEST.in README.rst setup.py /synapse/
+COPY MANIFEST.in README.rst /synapse/
```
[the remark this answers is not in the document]
DMRobertson: It should, yes. It gets removed later though in [`87ecb33` (#12337)](https://github.com/matrix-org/synapse/pull/12337/commits/87ecb333e3d2f6a687b812cfbeda45ccb7d768e7). I don't want to cause reviewer pain by rebasing now, so propose leaving it as it is.

pyproject.toml
```diff
@@ -66,3 +66,197 @@ multi_line_output = 3
...
+# ijson 3.1.4 fixes a bug with "." in property names
+ijson = ">=3.1.4"
+matrix-common = "~=1.1.0"
+# We need packaging.requirements.Requirement, added in 16.1.
+packaging = ">=16.1"
+
+
+# Optional Dependencies
+# ---------------------
+matrix-synapse-ldap3 = { version = ">=0.1", optional = true }
```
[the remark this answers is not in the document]
DMRobertson: I'd be happy to drop it, but wouldn't want to do so in this PR.

`poetry install` will let you keep packages around in its venv even if they're no longer declared as dependencies. (`poetry install --remove-untracked` will clean up such orphans, but that's not the default behaviour.)

docker/Dockerfile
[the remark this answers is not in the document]
babolivier: So it does (I guess that's the issue with commit-by-commit reviews). Thanks for pointing this out!

synapse/__init__.py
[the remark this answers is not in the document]
DMRobertson: #12342 to track this.

pyproject.toml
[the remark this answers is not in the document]
babolivier: Fair enough.

.gitignore
```diff
@@ -15,6 +15,10 @@ _trial_temp*/
 .DS_Store
 __pycache__/
 
+# We do want the poetry lockfile. TODO: is there a good reason for ignoring
+# '*.lock' above? If not, let's nuke it.
```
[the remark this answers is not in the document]
DMRobertson:
```suggestion
# We do want the poetry lockfile. Ignoring '*.lock' above seems to ignore tox and trial lock files.
```

pyproject.toml
```diff
@@ -66,3 +66,197 @@ multi_line_output = 3
...
+opentracing = { version = ">=2.2.0", optional = true }
+jaeger-client = { version = ">=4.0.0", optional = true }
+pyjwt = { version = ">=1.6.4", optional = true }
+txredisapi = { version = ">=1.4.7", optional = true }
+hiredis = { version = "*", optional = true }
+Pympler = { version = "*", optional = true }
+parameterized = { version = ">=0.7.4", optional = true }
+
+## Developer dependencies should not get included in "all".
+##
```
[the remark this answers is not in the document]
DMRobertson: I can't fully remember what the intent was here. I think it's better served by the comments explaining which extras get omitted from all.

```suggestion
## Developer dependencies should not get included in "all".
##
```

README.rst
DMRobertson: Can do.

```suggestion
    poetry run synapse_homeserver \
      --server-name my.domain.name \
      --config-path homeserver.yaml \
      --generate-config \
      --report-stats=[yes|no]

    # Start the app
    poetry run synapse_homeserver --config-path homeserver.yaml
```

README.rst
```diff
@@ -293,39 +293,41 @@ directory of your choice::
...
+(to stop, you can use `poetry run ./demo/stop.sh`)
 
 See the [demo documentation](https://matrix-org.github.io/synapse/develop/development/demo.html)
 for more information.
 
 If you just want to start a single instance of the app and run it directly::
 
     # Create the homeserver.yaml config once
-    python -m synapse.app.homeserver \
+    poetry run python -m synapse.app.homeserver \
```
DMRobertson, line 323:
```suggestion
    poetry run synapse_homeserver \
```

README.rst
```diff
@@ -293,39 +293,41 @@ directory of your choice::
...
       --server-name my.domain.name \
       --config-path homeserver.yaml \
       --generate-config \
       --report-stats=[yes|no]
 
     # Start the app
-    python -m synapse.app.homeserver --config-path homeserver.yaml
+    poetry run python -m synapse.app.homeserver --config-path homeserver.yaml
```
DMRobertson, line 330:
```suggestion
    poetry run synapse_homeserver --config-path homeserver.yaml
```

README.rst
DMRobertson: Yes, https://github.com/matrix-org/synapse/pull/12337/commits/5dc21d95b7a4b0f3f1143aa012994cc3a04d7c26 should do this.

docs/code_style.md
```diff
@@ -35,7 +35,7 @@ pip install -e ".[lint,mypy]"
     Check all application and test code with:
 
     ```sh
-    flake8 synapse tests
+    flake8 .
```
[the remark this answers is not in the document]
DMRobertson: That's right. Motivation here: https://github.com/matrix-org/synapse/issues/12107

In a nutshell: I think we should be configuring linters to select sensible files, rather than propagating lists of directories around everywhere.

This isn't specific to poetry, but I've snuck it in here because I'm making use of this action: https://github.com/matrix-org/backend-meta/blob/main/.github/workflows/python-poetry-ci.yml#L19-L25 which doesn't accept a list of targets for for the reason described in the previous sentence.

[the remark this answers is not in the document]
babolivier: Makes sense, thanks for explaining!

.ci/scripts/test_old_deps.sh
DMRobertson: I left a TODO for this in e96fd31ef.

debian/build_virtualenv
```diff
@@ -43,7 +43,8 @@ dh_virtualenv \
     --preinstall="wheel" \
     --extra-pip-arg="--no-cache-dir" \
     --extra-pip-arg="--compile" \
-    --extras="all,systemd,test"
+    --extras="all,systemd,test" \
+    --requirements="requirements.txt"
```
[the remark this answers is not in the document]
DMRobertson: This is 6fd43c4b9.

DMRobertson:
@babolivier I think I've gone over most of your comments. Could you take another look?

(Happy to solicit input from anyone given the enormousness of this enterprise)

.ci/scripts/test_old_deps.sh
```diff
@@ -9,12 +12,66 @@ set -ex
...
+
+del data['tool']['poetry']['dev-dependencies']
+
+with open('pyproject.toml', 'w') as f:
+    toml.dump(data, f)
+"
+python3 -c "$REMOVE_DEV_DEPENDENCIES"
+
+echo "::group::Patched pyproject.toml"
+diff -u pyproject.toml-backup pyproject.toml
```
DMRobertson, line 71: `toml.dump` unfortunately doesn't preserve the formatting so the diff isn't so useful. Also, `diff` returns a nonzero status code causing the job to fail. I'll change this to just cat out the patched file.

```suggestion
cat pyproject.toml
```

docker/Dockerfile-dhvirtualenv
H-Shay: I checked on my local branch and looks like you can rip it out-also wheel.

DMRobertson: Thanks. I wonder why we needed to upgrade pip after the poetry changes, but not before?

DMRobertson: See 18f91a7ac.

DMRobertson:
In https://gist.github.com/DMRobertson/f23cd46b1f1b60ff5369ed890271cce0 I've computed the difference between wheels and sdists before and after this change. AFAICS all changes are explained by

- changes to Synapse in this branch, or
- differences between how `bdist_wheel` and `poetry-core` produce wheels.

There's still a lot of moving parts here though, so I'd appreciate extra eyes.

debian/build_virtualenv
```diff
@@ -32,6 +32,8 @@ esac
 
 # Use --builtin-venv to use the better `venv` module from CPython 3.4+ rather
 # than the 2/3 compatible `virtualenv`.
+# Use the requirements exported from poetry so that we reproduce its locked
+# environment exactly. We generate requirements.txt in build_debian.sh.
```
richvdh, line 36: arg. This isn't really a thing we're meant to do.

The idea of debian packaging is that you can take a vanilla Debian VM, install the build dependencies, and expect `dpkg-buildpackage` to just work. Obviously, this adds dependencies on the environment.

A temporary workaround in the interests of landing this thing might be to *commit* requirements.txt (possibly inside the `debian` directory), but that's a thing we'll have to maintain manually. 

The correct solution is to run the `poetry export` here.

docker/Dockerfile
```diff
@@ -16,18 +16,31 @@
 
 ARG PYTHON_VERSION=3.9
 
+FROM docker.io/python:${PYTHON_VERSION}-slim as base
```
richvdh, line 19: I'm a bit confused why we have an extra `FROM` statement now?

docker/Dockerfile
```diff
@@ -45,33 +58,36 @@ RUN \
...
+
+# Copy just what we need to run `poetry install`
+COPY pyproject.toml poetry.lock README.rst /synapse/
 
+# Install to the Python installation which hosts `pip`. In this case, it's the system
+# Python.
+ENV POETRY_VIRTUALENVS_IN_PROJECT=true \
+    POETRY_VIRTUALENVS_CREATE=true \
+    POETRY_HOME=/opt/poetry
 # To speed up rebuilds, install all of the dependencies before we copy over
```
richvdh, line 71:
```suggestion

# To speed up rebuilds, install all of the dependencies before we copy over
```

docker/Dockerfile
```diff
@@ -45,33 +58,36 @@ RUN \
...
-COPY synapse/__init__.py /synapse/synapse/__init__.py
-COPY synapse/python_dependencies.py /synapse/synapse/python_dependencies.py
+WORKDIR /synapse
+
+# Copy just what we need to run `poetry install`
+COPY pyproject.toml poetry.lock README.rst /synapse/
 
+# Install to the Python installation which hosts `pip`. In this case, it's the system
+# Python.
+ENV POETRY_VIRTUALENVS_IN_PROJECT=true \
```
richvdh, line 68: I think using ENV here will mean we end up with these env vars exported in the final image? Are you sure that's what we want?

docker/Dockerfile
```diff
@@ -45,33 +58,36 @@ RUN \
...
-# This is aiming at installing the `install_requires` and `extras_require` from `setup.py`
-RUN --mount=type=cache,target=/root/.cache/pip \
-  pip install --prefix="/install" --no-warn-script-location \
-    /synapse[all]
+RUN --mount=type=cache,target=/opt/poetry/artifacts \
+    --mount=type=cache,target=/opt/poetry/.cache/pypoetry/cache \
+  /root/.local/bin/poetry install --no-dev --no-root --no-interaction --no-ansi --extras all
 
-# Copy over the rest of the project
+# Copy over the synapse source code.
```
richvdh, line 78:
```suggestion
# Copy over the rest of the synapse source code.
```

docker/Dockerfile
```diff
@@ -93,7 +109,7 @@ RUN \
     openssl \
     && rm -rf /var/lib/apt/lists/*
 
-COPY --from=builder /install /usr/local
+COPY --from=builder /synapse/ /synapse
```
richvdh, line 112: /me wonders how many people rely on things being installed in `/usr/local`.

Certainly this is going to break anyone who builds docker images that install extension modules.

docker/Dockerfile-dhvirtualenv
```diff
@@ -85,6 +85,16 @@ RUN apt-get update -qq -o Acquire::Languages=none \
         libpq-dev \
         xmlsec1
 
+# Debian 10 (Buster)'s python3-pip package includes pip 18.1. This version was unable to
+# pull wheels. (I'd guess we need support for manylinux2010, manylinux2014 or PEP 600,
+# introduced in pip 19.0, 19.3 and 20.3 respectively.)
+# In any case, we upgrade pip rather than compile CPython extensions from source.
+RUN pip3 install -U pip
```
richvdh, line 92: doesn't the `--upgrade-pip` in the `dh_virtualenv` invocation take care of this.

docker/Dockerfile-dhvirtualenv
```diff
@@ -85,6 +85,16 @@ RUN apt-get update -qq -o Acquire::Languages=none \
...
+
+# We normally use poetry 1.1.12. However we need to use a prerelease of 1.2 here,
+# which fixes a bug in poetry export that stops us from exporting for Python 3.8 builds.
+RUN pip3 install poetry==1.2.0b1
```
richvdh, line 96: again, we can't do this here. If our build depends on these versions of pip and poetry, we need to do the installation *inside* the build, not i the docker image.

docker/start.py
```diff
@@ -9,6 +9,8 @@
 
 import jinja2
 
+VIRTUALENV_INTERPRETER = "/synapse/.venv/bin/python"
```
richvdh, line 12: can we just use [`sys.executable`](https://docs.python.org/3/library/sys.html#sys.executable) here rather than hardcoding it?

```suggestion
VIRTUALENV_INTERPRETER = sys.executable
```

(or better yet, inline it below)

docs/code_style.md
```diff
@@ -46,11 +41,9 @@ pip install -e ".[lint,mypy]"
     Auto-fix imports with:
 
     ```sh
-    isort -rc synapse tests
+    isort .
     ```
 
-    `-rc` means to recursively search the given directories.
-
```
richvdh, line 53: is `-rc` no longer needed?

pyproject.toml
```diff
@@ -36,24 +36,9 @@
...
-    | _trial_temp.*
-    | build
-    | dist
-    | debian
-  )/
-)
-'''
+# black ignores everything in .gitignore by default, see
+# https://black.readthedocs.io/en/stable/usage_and_configuration/file_collection_and_discovery.html#gitignore
+# Use `extend-exclude` if you want to exclude something in addition to this.
```
richvdh, line 41: given the size of this pr, I'd be looking for everything we can do to move stuff out of it. These changes to lint (and in `code_style.md` and `lint.sh`) look like candidates.

richvdh (changes requested):
this looks broadly great - most of the comments here are fairly small.

Concerns about the debian build though. Sorry, I'm going to be difficult about that. Have suggested a workaround for now, to get this landed.

---

Question: suppose we land the changes to `pyproject.toml` and `poetry.lock` without *any other changes* for now. Notwithstanding the fact that means we now have two places we have to maintain the dependency list, does that break anything? Or does it leave us in a position where you can use *either* `pip` *or* `poetry` to do the installation?

I'm thinking it opens up a path to splitting up this PR into much smaller lumps, since we can then follow up much smaller PRs like:
 * switch the `olddeps` build over to poetry
 * switch `.ci/scripts/test_synapse_port_db.sh` over to poetry
 * switch the docker image over to poetry

babolivier (commented):
Forgot to send these comments

docker/start.py
DMRobertson: Done in #12385.

debian/build_virtualenv
callahad: > and expect dpkg-buildpackage to just work. 

So it does, kind of.

The impurity here is that our builds expect a `requirements.txt` to exist, which we're providing out of band (via `poetry export` in the builder container).

If that file is missing, like in the case of a raw `dpkg-buildpackage` invocation in a pristine Debian environment, then dh-virtualenv [silently skips it](https://github.com/spotify/dh-virtualenv/blob/6d5a12aae672fc7204ba4093e882cef871d58b14/dh_virtualenv/deployment.py#L201-L202) and then [eventually calls](https://github.com/spotify/dh-virtualenv/blob/6d5a12aae672fc7204ba4093e882cef871d58b14/dh_virtualenv/deployment.py#L270-L272) `pip install .` which has the effect of grabbing the latest compatible versions of all deps.

This obviously circumvents our attempts at dependency locking, but it does have the interesting property that "if a requirements.txt file is present, our `debian/rules` script will use it to constrain exactly which versions of dependencies are selected for the virtualenv."

@richvdh In ascending order of implementation difficulty, would you rather we:

1. Codify the above as intended behavior, updating comments / logging to be more explicit about what's happening.
2. Modify `debian/rules` (well, `debian/build_virtualenv`, really) to install Poetry from the Internet (probably into an ephemeral venv? Or using [pipx](https://pypa.github.io/pipx/)?), run `poetry export`, then nuke our Poetry installation
3. Modify all workflows / CI / etc., to ensure that we always have an up-to-date `poetry export > requirements.txt` checked into the repo whenever `poetry.lock` gets updated.
4. Manually package Poetry for all of our various Debian versions, much like we do dh-virtualenv (we can't use downstream poetry packages since the `export` command only works correctly for us in the 1.2 prereleases)

...or something else entirely?

richvdh: > If that file is missing, like in the case of a raw dpkg-buildpackage invocation in a pristine Debian environment, then dh-virtualenv [silently skips it](https://github.com/spotify/dh-virtualenv/blob/6d5a12aae672fc7204ba4093e882cef871d58b14/dh_virtualenv/deployment.py#L201-L202) and then [eventually calls](https://github.com/spotify/dh-virtualenv/blob/6d5a12aae672fc7204ba4093e882cef871d58b14/dh_virtualenv/deployment.py#L270-L272) pip install . which has the effect of grabbing the latest compatible versions of all deps.

hrm, that is interesting; I hadn't realised that; it does change my feelings here a bit.

I was expecting us to do option 2 (update `debian/build_virtualenv` to install and run Poetry, somehow), and that still feels like the cleanest option, but on reflection I could settle for option 1.

callahad: To be honest, I was surprised at the silent success, but the siren song of "it's not a bug, it's a feature!"

(Though I do feel like it's a bug in dh-virtualenv which may eventually get fixed, biting us in the arse. For example, the dh-poetry project Shay found *never* worked as described, but appeared to due to that behavior)

Let's run with option 2... which will be janky in its own way, but dh-virtualenv itself is a bit of a rickety ship.

callahad: Resolved in f0d72acebfd373b6428e8841898e90851a941192

docker/Dockerfile-dhvirtualenv
callahad: Resolved in f0d72acebfd373b6428e8841898e90851a941192

docker/Dockerfile-dhvirtualenv
callahad: Resolved in f0d72acebfd373b6428e8841898e90851a941192

callahad:
Note: CI is currently passing across the board except for Complement, which will be resolved by:

- #12404

docker/Dockerfile-dhvirtualenv
callahad: Resolved in f0d72acebfd373b6428e8841898e90851a941192

pyproject.toml
callahad: Merged in develop, resolving this. Now that develop has a `pyproject.toml` file with this metadata, we shouldn't need to worry about this further in this PS.

.github/workflows/twisted_trunk.yml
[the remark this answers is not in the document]
callahad: Fixed in 232a270d45124259e54c031b6b23baa246fab9de

.github/workflows/tests.yml
```diff
@@ -128,18 +119,18 @@ jobs:
...
     steps:
       - uses: actions/checkout@v2
       - name: Test with old deps
         uses: docker://ubuntu:focal # For old python and sqlite
+        # Note: focal seems to be using 3.8, but the oldest is 3.7?
         with:
           workdir: /github/workspace
           entrypoint: .ci/scripts/test_old_deps.sh
-        env:
-          TRIAL_FLAGS: "--jobs=2"
```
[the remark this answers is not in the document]
DMRobertson: I now pass `-j 2` to `trial` directly in the olddeps script; see https://github.com/matrix-org/synapse/pull/12337/files#diff-2aee3456887f885b373d6c8b3912454677082f5c6e11715a3548c1a58a1688cfR77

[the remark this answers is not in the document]
DMRobertson: Though I may as well preserve the spelling `--jobs=2`.

.ci/scripts/test_old_deps.sh
```diff
@@ -9,12 +12,66 @@ set -ex
...
-exec tox -e py3-old
+# TODO: in the future, we could use an implementation of
+#   https://github.com/python-poetry/poetry/issues/3527
+#   https://github.com/pypa/pip/issues/8085
+# to select the lowest possible versions, rather than resorting to this sed script.
+
+# Patch the project definitions in-place:
+# - Replace all lower and tilde bounds with exact bounds
+# - Make the pyopenssl 17.0, which can work against an
+#   OpenSSL 1.1 compiled cryptography (as older ones don't compile on Travis).
```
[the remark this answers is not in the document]
DMRobertson: I'm going to pull out the olddeps changes into #12407 and address this there.

.github/workflows/tests.yml
```diff
@@ -128,18 +119,18 @@ jobs:
...
   trial-olddeps:
+    # Note: sqlite only; no postgres
     if: ${{ !cancelled() && !failure() }} # Allow previous steps to be skipped, but not fail
     needs: linting-done
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
       - name: Test with old deps
         uses: docker://ubuntu:focal # For old python and sqlite
+        # Note: focal seems to be using 3.8, but the oldest is 3.7?
```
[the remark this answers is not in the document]
DMRobertson: I'm going to pull out the olddeps changes into #12407 and address this there.

.ci/scripts/test_old_deps.sh
```diff
@@ -9,12 +12,66 @@ set -ex
...
+# - Delete all lines referring to psycopg2 --- so no testing of postgres support.
+# - Omit systemd: we're not logging to journal here.
+
+# TODO: also replace caret bounds, see https://python-poetry.org/docs/dependency-specification/#version-constraints
+# We don't use these yet, but IIRC they are the default bound used when you `poetry add`.
```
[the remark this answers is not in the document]
DMRobertson: I'm going to pull out the olddeps changes into #12407 and address this there.

.ci/scripts/test_old_deps.sh
```diff
@@ -9,12 +12,66 @@ set -ex
...
+# The sed expression 's/\^/==/g' ought to do the trick. But it would also change
+# `python = "^3.7"` to `python = "==3.7", which would mean we fail because olddeps
+# runs on 3.8 (#12343).
+
+sed -i-backup \
```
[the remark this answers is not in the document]
DMRobertson: I'm going to pull out the olddeps changes into #12407 and address this there.

.ci/scripts/test_old_deps.sh
```diff
@@ -9,12 +12,66 @@ set -ex
...
+   -e '/systemd/d' \
+   pyproject.toml
+
+# Use poetry to do the installation. This ensures that the versions are all mutually
+# compatible (as far the package metadata declares, anyway); pip's package resolver
+# is more lax.
+#
+# Rather than `poetry install --no-dev`, we drop all dev dependencies from the
+# toml file. This means we don't have to ensure compatibility between old deps and
+# dev tools.
```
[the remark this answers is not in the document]
DMRobertson: I'm going to pull out the olddeps changes into #12407 and address this there.

.github/workflows/tests.yml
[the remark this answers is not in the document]
DMRobertson: I'm going to pull out the olddeps changes into #12407 and address this there.

docker/Dockerfile
DMRobertson: #12385 has changed this to avoid the breakage.

docker/Dockerfile
DMRobertson: Done in https://github.com/matrix-org/synapse/pull/12385.

docker/Dockerfile
DMRobertson: Dealt with in https://github.com/matrix-org/synapse/pull/12385.

docker/Dockerfile
DMRobertson: Dealt with in https://github.com/matrix-org/synapse/pull/12385.

docker/Dockerfile
DMRobertson: Dealt with in https://github.com/matrix-org/synapse/pull/12385.

.ci/scripts/test_old_deps.sh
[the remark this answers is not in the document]
DMRobertson: Answer: not any more. I tried to print a diff between the source tree and patched `pyproject.toml`. But the toml fiddling via Python destroys all formatting so the diff was pretty useless.

.ci/patch_for_twisted_trunk.sh
```diff
@@ -1,8 +1,22 @@
 #!/bin/sh
 
-# replaces the dependency on Twisted in `python_dependencies` with trunk.
+# replaces the dependency on Twisted with Twisted's trunk.
```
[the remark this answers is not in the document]
callahad: Resolved in 9225764bc8fae4dd1e5dc8483975ec1d14e7d2a7

.github/workflows/twisted_trunk.yml
```diff
@@ -55,11 +59,18 @@ jobs:
...
```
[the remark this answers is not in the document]
callahad: Resolved in 232a270d45124259e54c031b6b23baa246fab9de

Note that GitHub Actions runs multi-line `run:` stanzas with `bash -eo pipefail` by default, so you can use a simple `run: |` and dispense with the `&&` combinations.

docker/start.py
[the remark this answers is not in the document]
callahad: This has been resolved by merging in `develop`

.github/workflows/twisted_trunk.yml
callahad: We won't *need* it once merged, but there's no harm in leaving it in -- it will simply prevent the Twisted trunk workflow from running multiple times concurrently for the same branch. Let's pull it out after merging, since it's handy for testing before then.

.github/workflows/tests.yml
```diff
@@ -155,23 +146,24 @@ jobs:
...
     needs: linting-done
     runs-on: ubuntu-latest
     strategy:
       matrix:
         python-version: ["pypy-3.7"]
+        extras: ["all"]
 
     steps:
       - uses: actions/checkout@v2
+      # TODO: why is this list different from  above?
```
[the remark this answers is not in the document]
callahad: Updating comment.

.github/workflows/tests.yml
```diff
@@ -20,19 +20,11 @@ jobs:
...
-
-    steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-python@v2
-      - run: pip install tox
-      - run: tox -e ${{ matrix.toxenv }}
+    # This does a vanilla `poetry install` - no extras. I'm slightly anxious
+    # that we might skip some typechecks on code that uses extras. However,
+    # I think the right way to fix this is to turn up mypy's strictness:
+    # disallow unknown imports and accept fewer uses of `Any`.
```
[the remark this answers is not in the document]
callahad: Addressing in a pending commit.

.ci/scripts/test_export_data_command.sh
```diff
@@ -2,29 +2,23 @@
...
-pip install psycopg2
-
-# Install Synapse itself. This won't update any libraries.
-pip install -e .
-
 echo "--- Generate the signing key"
 
 # Generate the server's signing key.
-python -m synapse.app.homeserver --generate-keys -c .ci/sqlite-config.yaml
+poetry run synapse_homeserver --generate-keys -c .ci/sqlite-config.yaml
```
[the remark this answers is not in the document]
callahad: Let's leave well enough alone; this works and can be refactored after merging if we decide invoking the script itself with poetry is more ergonomic.

pyproject.toml
callahad: Changes which can be applied independently are being extract into their own PRs; marking as resolved.

docs/code_style.md
callahad: Correct. Since isort 5.0.0 (July 2020) the `--recursive` option is removed and directories are sorted recursively by default.

DMRobertson:
I've ripped out what I need to from this. Many thanks all for you eyes and words.

PR #12340, opened 2022-03-31
docs: Don't render the table of contents on the print page

anoadragon453:
If you try to print our docs currently: https://matrix-org.github.io/synapse/v1.55/print.html

...you will presented with about 6 pages of page headers before you finally get to the actual content. This is due to our custom table-of-contents renderer rendering all page titles upon page load. This is very useful for most pages, but not the print page (which shows all contents in the entire book).

This PR exits the table-of-contents loading code early if it notices that it's running on the print page.

### DIFF
--- a/changelog.d/12340.doc
+++ b/changelog.d/12340.doc
@@ -1,0 +1,1 @@
+Fix rendering of the documentation site when using the 'print' feature.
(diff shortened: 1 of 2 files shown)


anoadragon453:
@squahtx Yes! One can build the documentation locally using [these steps](https://github.com/matrix-org/synapse/tree/develop/docs#building-the-documentation). Then navigate to http://localhost:3000/print.html to see the difference between the branches.

Before:

image

After:

image

In the second version we actually end up with a blank page as the first page, with the content starting on the second page. But it's a step in the right direction versus 6 pages of nonsense, a blank page, and then the content.

PR #12344, opened 2022-03-31
Disable groups/communities by default.

clokep:
This is the next bit of #11584 -- disabling groups by default, but leaving the configuration flag there to enable it.

**Blocked until develop is on Synapse v1.58.0.**

Requires matrix-org/sytest#1233.

### DIFF
--- a/tests/rest/admin/test_admin.py
+++ b/tests/rest/admin/test_admin.py
@@ -63,6 +63,7 @@ class DeleteGroupTestCase(unittest.HomeserverTestCase):
         self.other_user = self.register_user("user", "pass")
         self.other_user_token = self.login("user", "pass")
 
+    @unittest.override_config({"experimental_features": {"groups_enabled": True}})
     def test_delete_group(self) -> None:
         # Create a new group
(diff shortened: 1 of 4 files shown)


clokep:
Apparently we said we wouldn't do this until 1.58.0, I thought we were doing it in 1.57.0 😢

tests/rest/admin/test_admin.py
```diff
@@ -63,6 +63,7 @@ def prepare(self, reactor: MemoryReactor, clock: Clock, hs: HomeServer) -> None:
         self.other_user = self.register_user("user", "pass")
         self.other_user_token = self.login("user", "pass")
 
+    @unittest.override_config({"experimental_features": {"groups_enabled": True}})
```
reivilibre, line 66: Do we really only have one test that uses groups?

Edit: ah, I see. The flag changes whether or not the servlets get registered; the other tests register the servlets themselves.

reivilibre (approved):
Looking forward to seeing these gone :-)

PR #12346, opened 2022-04-01
Remove redundant `get_success` calls in test code

richvdh:
There are a bunch of places we call `get_success` on an immediate value, which is unnecessary. Let's rip them out, and remove the redundant functionality in `get_success` and friends.


squahtx (approved):
lgtm if the tests pass!

PR #12347, opened 2022-04-01
Add type hints for `tests/unittest.py`.

richvdh:
In particular, add type hints for `get_success` and friends, which are then helpful in a bunch of places.

Needs a few fixes, so review commit-by-commit.

### DIFF
--- a/tests/storage/databases/main/test_lock.py
+++ b/tests/storage/databases/main/test_lock.py
@@ -28,7 +28,7 @@ class LockTestCase(unittest.HomeserverTestCase):
         """
         # First to acquire this lock, so it should complete
         lock = self.get_success(self.store.try_acquire_lock("name", "key"))
-        self.assertIsNotNone(lock)
+        assert lock is not None
 
         # Enter the context manager
         self.get_success(lock.__aenter__())
@@ -45,7 +45,7 @@ class LockTestCase(unittest.HomeserverTestCase):
 
         # We can now acquire the lock again.
         lock3 = self.get_success(self.store.try_acquire_lock("name", "key"))
-        self.assertIsNotNone(lock3)
+        assert lock3 is not None
         self.get_success(lock3.__aenter__())
         self.get_success(lock3.__aexit__(None, None, None))
 
@@ -53,7 +53,7 @@ class LockTestCase(unittest.HomeserverTestCase):
         """Test that we don't time out locks while they're still active"""
 
         lock = self.get_success(self.store.try_acquire_lock("name", "key"))
-        self.assertIsNotNone(lock)
+        assert lock is not None
 
         self.get_success(lock.__aenter__())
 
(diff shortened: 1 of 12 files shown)


tests/unittest.py
```diff
@@ -511,30 +534,36 @@ async def run_bg_updates():
...
         self.reactor.pump([by] * 100)
 
-    def get_success(self, d, by=0.0):
-        deferred: Deferred[TV] = ensureDeferred(d)
+    def get_success(
+        self,
+        d: Awaitable[TV],
+        by: float = 0.0,
+    ) -> TV:
+        deferred: Deferred[TV] = ensureDeferred(d)  # type: ignore[arg-type]
```
clokep, line 548: What does mypy want the type of `ensureDeferred` to be?

tests/storage/databases/main/test_lock.py
```diff
@@ -28,7 +28,7 @@ def test_simple_lock(self):
         """
         # First to acquire this lock, so it should complete
         lock = self.get_success(self.store.try_acquire_lock("name", "key"))
-        self.assertIsNotNone(lock)
+        assert lock is not None
```
clokep, line 31: I find it quite annoying that mypy doesn't handle `assertIs(Not)None` the same way...

python/mypy#5088 for cross-refs.

tests/unittest.py
```diff
@@ -642,11 +671,11 @@ def register_appservice_user(
 
     def login(
         self,
-        username,
-        password,
-        device_id=None,
+        username: str,
+        password: str,
+        device_id: Optional[str] = None,
         custom_headers: Optional[Iterable[Tuple[AnyStr, AnyStr]]] = None,
```
clokep, line 677: Should this be the `CustomHeaderType`?

clokep (commented):
Seems good, just a couple of questions!

tests/unittest.py
clokep: Seems it was a `Coroutine` or `Deferred`: https://github.com/twisted/twisted/blob/5c24e99e671c4082a1ddc8dbeb869402294bd0dc/src/twisted/internet/defer.py#L1094-L1096

tests/unittest.py
richvdh: yup!

tests/unittest.py
richvdh: yeah. We end up passing in things here that implement `Coroutine["Deferred[_T]", object, "_T"]`, (so they work in practice) but aren't typed that way (they just implement `Awaitable[_T]`). To keep the tests themselves clean, it's easier just to `ignore[arg-type]`.

tests/unittest.py
richvdh, line 677:
```suggestion
        custom_headers: Optional[Iterable[CustomHeaderType]] = None,
```

tests/unittest.py
clokep: That makes sense, thanks!

PR #12348, opened 2022-04-01
Move single-use methods out of `TestCase`

richvdh:
These methods are only used by a single testcase, so they shouldn't be
cluttering up the base `TestCase` class.


clokep:
(Assuming CI is good!)

Issue #11498, opened 2021-12-02
Synapse selects `"preset": "public_chat"` by default, while it should be "private_chat"


ShadowJonathan:
### Description

I noticed this while looking at the spec;

While creating rooms [`/createRooms`](https://spec.matrix.org/v1.1/client-server-api/#post_matrixclientv3createroom) recommends, for a default `"preset"`, to use `"private_chat"`, recommending it should look at `"visibility"`, which defaults to `"private"`.

However, synapse creates public_chat rooms when given no options.

I know that the spec is only saying "should" here, but since Dendrite and Conduit also display this behaviour, it could cause confusion that a "known good" server such as Synapse is going against spec here, creating bugs in software where previously there were none.

### Version information

- **Version**: v1.48.0

- **Install method**: Docker

- **Platform**: Linux

reivilibre:
Even our test helpers say it should be private by default (of course the tests in practice rely on them being public):
```python
            is_public: If True, the `visibility` parameter will be set to
                "public". If False, it will be set to "private". If left
                unspecified, the server will set it to an appropriate default
                (which should be "private" as per the CS spec).
```

PR #12350, opened 2022-04-01
Default to `private` room visibility rather than `public` when a client does not specify one, according to spec.

reivilibre:
Our tests pretty heavily depend on the broken behaviour, so I changed our test helper to default to creating public rooms rather than 'default' rooms, since I think that matches the casual style of most tests anyway.
(Tests are usually not too fussed about privacy rules and just want some users in a room!)

Should be a commit-by-commit review. Just a quick fix to #11498.


squahtx (approved):
lgtm

Are we missing a test for spec compliance somewhere?

PR #12353, opened 2022-04-01
Convert `Linearizer` tests from `inlineCallbacks` to async

### LINKED ISSUE #12114: `Linearizer` can get stuck due to cancellation
Once a `Linearizer` task acquires the lock, it does a 0-duration sleep to reset the call stack:
https://github.com/matrix-org/synapse/blob/605d161d7d585847fd1bb98d14d5281daeac8e86/synapse/util/async_helpers.py#L466

In the rare event that the task is cancelled while sleeping, the lock is not released and the `Linearizer` becomes stuck.

Can be reproduced by modifying the cancellation test slightly. Cancelling the 2nd task after 1st task releases the lock (instead of before) will make the test get stuck indefinitely:
```py
        with cm1:
            pass

        d2.cancel()
```
instead of 
https://github.com/matrix-org/synapse/blob/release-v1.53/tests/util/test_linearizer.py#L159-L162

squahtx:
To fix #12114, it might be useful to convert the `Linearizer` from raw `Deferred`s to async. But first, we need to convert the tests.

Probably easier to review commit by commit. There's a commit in the middle that adds comments, so that the correspondence between `inlineCallbacks` and async code is clearer.

### Pull Request Checklist

### DIFF
--- a/tests/util/test_linearizer.py
+++ b/tests/util/test_linearizer.py
@@ -122,7 +122,7 @@ class LinearizerTestCase(unittest.TestCase):
     def test_lots_of_queued_things(self) -> None:
         """Tests lots of fast things queued up behind a slow thing.
 
-        The stack should *not* explode when the fast thing completes.
+        The stack should *not* explode when the slow thing completes.
         """
         linearizer = Linearizer()
         key = ""
--- a/tests/util/test_linearizer.py
+++ b/tests/util/test_linearizer.py
@@ -37,6 +37,7 @@ class LinearizerTestCase(unittest.TestCase):
         d2 = linearizer.queue(key)
         self.assertFalse(d2.called)
 
+        # Once the first task is done, the second task can continue.
         with cm1:
             self.assertFalse(d2.called)
 
@@ -56,13 +57,14 @@ class LinearizerTestCase(unittest.TestCase):
         d1 = linearizer.queue(key)
         cm1 = yield d1
 
-        # Since d1 gets called immediately, "is_queued" should return false.
+        # Since the first task acquires the lock immediately, "is_queued" should return
+        # false.
         self.assertFalse(linearizer.is_queued(key))
 
         d2 = linearizer.queue(key)
         self.assertFalse(d2.called)
 
-        # Now d2 is queued up behind successful completion of cm1
+        # Now the second task is queued up behind the first.
         self.assertTrue(linearizer.is_queued(key))
 
         with cm1:
@@ -71,7 +73,7 @@ class LinearizerTestCase(unittest.TestCase):
             # cm1 still not done, so d2 still queued.
             self.assertTrue(linearizer.is_queued(key))
 
-        # And now d2 is called and nothing is in the queue again
+        # And now the second task acquires the lock and nothing is in the queue again.
         self.assertFalse(linearizer.is_queued(key))
 
         with (yield d2):
@@ -118,12 +120,14 @@ class LinearizerTestCase(unittest.TestCase):
         d3 = limiter.queue(key)
         cm3 = yield d3
 
+        # These next two tasks have to wait.
         d4 = limiter.queue(key)
         self.assertFalse(d4.called)
 
         d5 = limiter.queue(key)
         self.assertFalse(d5.called)
 
+        # Once the first task completes, the fourth task can continue.
         with cm1:
             self.assertFalse(d4.called)
             self.assertFalse(d5.called)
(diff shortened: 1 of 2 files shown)


tests/util/test_linearizer.py
```diff
@@ -13,160 +13,205 @@
...
+            # the reactor to advance.
+            self._pump()
+
+        return d, acquired_d, unblock
+
+    def _pump(self) -> None:
+        """Pump the reactor to advance `Linearizer`s."""
+        assert isinstance(reactor, ReactorBase)
+        while reactor.getDelayedCalls():
+            reactor.runUntilCurrent()
```
squahtx, line 71: I think `Linearizer`s don't use the same reactor as the Homeserver in our tests. We've never noticed because there's no contention in tests.

tests/util/test_linearizer.py
```diff
@@ -13,160 +13,205 @@
...
+             * A `Deferred` that resolves once the task acquires the lock.
+             * A function that unblocks the task. Must be called by the caller
+               to allow the task to release the lock and complete.
+        """
+        acquired_d: "Deferred[None]" = Deferred()
+        unblock_d: "Deferred[None]" = Deferred()
+
+        async def task() -> None:
+            with await linearizer.queue(key):
+                acquired_d.callback(None)
```
squahtx, line 54: Unlike the previous code, we enter the context manager immediately once `linearizer.queue()` resolves.

So
```py
cm1 = yield d1
with cm1:
    self.assert(...)
```
turns into
```py
# context manager entered already
self.assertTrue(acquired_d1.called)

self.assert(...)

unblock()  # exit the context manager
```

And the asserts that were once inside the context manager often become duplicates and are removed.

tests/util/test_linearizer.py
```diff
@@ -75,8 +80,10 @@ def test_linearizer_is_queued(self):
         self.assertFalse(linearizer.is_queued(key))
 
     def test_lots_of_queued_things(self):
-        # we have one slow thing, and lots of fast things queued up behind it.
-        # it should *not* explode the stack.
+        """Tests lots of fast things queued up behind a slow thing.
+
+        The stack should *not* explode when the fast thing completes.
```
DMRobertson, line 125: Should "the fast thing" be "the slow thing" here? Otherwise I'm not sure which of the fast things this refers to.

tests/util/test_linearizer.py
```diff
@@ -119,15 +119,15 @@ def test_linearizer_is_queued(self):
...
+    def test_lots_of_queued_things(self) -> None:
         """Tests lots of fast things queued up behind a slow thing.
 
         The stack should *not* explode when the fast thing completes.
         """
         linearizer = Linearizer()
         key = ""
 
-        async def func(i, wait_for=None) -> None:
+        async def func(i: int, wait_for: Optional["Deferred[None]"] = None) -> None:
```
DMRobertson, line 130: is (was?) `wait_for` unused?

DMRobertson (commented):
Fairly daunting, but the commit-by-commit breakdown was helpful. The transformations in the 7842add1052f81193b154547ecd55ec5a78de09f seem fairly mechanical---I hope I haven't missed too much out by skimming over those.

tests/util/test_linearizer.py
squahtx: Yep, that's a typo.

tests/util/test_linearizer.py
squahtx: it's used in the very first call (the "slow thing")

squahtx:
> The transformations in the https://github.com/matrix-org/synapse/commit/7842add1052f81193b154547ecd55ec5a78de09f seem fairly mechanical---I hope I haven't missed too much out by skimming over those.

Aww, that was kind of the main change. If you've convinced yourself the transformations look sensible then that's alright!

DMRobertson:
> > The transformations in the [7842add](https://github.com/matrix-org/synapse/commit/7842add1052f81193b154547ecd55ec5a78de09f) seem fairly mechanical---I hope I haven't missed too much out by skimming over those.
> 
> Aww, that was kind of the main change. If you've convinced yourself the transformations look sensible then that's alright!

I'll take another look!

tests/util/test_linearizer.py
DMRobertson: On -92 I can see `func(0, sleep=True)`, but I can't see an analogous call to func after the changes.

Maybe the point is that the slow thing is now:
- acquiring the linearizer lock on +139
- "waiting" to release it on +144?

If so, I still think `wait_for` is unused?

DMRobertson (changes requested):
I'm happy apart from this comment: https://github.com/matrix-org/synapse/pull/12353#discussion_r841638159

tests/util/test_linearizer.py
DMRobertson:
```suggestion
        async def func(i: int) -> None:
```

squahtx: It's gone now!

DMRobertson (approved):
LGTM!

PR #12356, opened 2022-04-01
Make `scripts-dev` pass `mypy --disallow-untyped-defs`

DMRobertson:
I couldn't find a way to actually configure `--disallow-untyped-defs` in the `scripts-dev` section. I think this might be easier in pyproject.toml syntax rather than mypy.ini syntax; see also #11302 and #11315.

The release script was a bit fiddlier than I expected to get it typechecking, so that's broken down into smaller commits.

Follows on from #12355.


scripts-dev/release.py
```diff
@@ -296,7 +301,9 @@ def tag(gh_token: Optional[str]):
 
     click.echo_via_pager(changes)
     if click.confirm("Edit text?", default=False):
-        changes = click.edit(changes, require_save=False)
+        edited_changes = click.edit(changes, require_save=False)
+        assert edited_changes is not None
+        changes = edited_changes
```
squahtx, line 306: The `click` docs suggest that `edited_changes` would be `None` when the user closes the editor without saving: https://click.palletsprojects.com/en/7.x/utils/#launching-editors

Do we want to abort via an assert or use the unedited changes in that case?

scripts-dev/release.py
```diff
@@ -163,6 +165,8 @@ def prepare():
 
     # Switch to the release branch.
     parsed_new_version = version.parse(new_version)
+    # parse() shouldn't be returning a LegacyVersion. Whatever that is.
+    assert isinstance(parsed_new_version, version.Version)
```
squahtx, line 169: The docs suggest that we'll only see a `LegacyVersion` if we feed in non-PEP 440 compliant strings: https://packaging.pypa.io/en/latest/version.html#packaging.version.parse

Which ought to never happen with our version numbering, so the assert is fine!

squahtx (commented):
Looks good, one question

scripts-dev/release.py
DMRobertson: I found the docs hard to read. It seems like there might be 

Testing with `print(click.edit("Hello", require_save=...))` yielded:

|                      | require_save: False | requre_save: True |
|----------------------|---------------------|-------------------|
| **quit without save**    | original text       | None              |
| **save without changes** | original text       | original text     |
| **save with  changes**   | changed text        | changed text      |

So I think this assert is for mypy's benefit only. Perhaps it would be clearer to cast this to `str` with a comment?

squahtx: Curious. In that case I'm happy with the assert!

squahtx (approved):

DMRobertson:
@squahtx there were some merge conflicts to resolve after my poetry-related changes on the 1.58 release branch. Could you take another look please?

squahtx (approved):
I'm still happy.
What's the removal of the `redbaron` dependency about?

DMRobertson:
> I'm still happy. What's the removal of the `redbaron` dependency about?

That comes from [this commit ](https://github.com/matrix-org/synapse/commit/a54d9b0508c17a29f2f02c87054cd8990d476054) on the 1.58 release branch. We used to use redbaron to parse `__init__.py` and update the version number in place. That's now handled by asking poetry to update pyproject.toml for us (from [this commit](https://github.com/matrix-org/synapse/commit/730fcda5463144f09ff1b34198c678021469ee28) on the release branch).

PR #12360, opened 2022-04-03
Implement MSC3759: Leave event metadata for deactivated users

Half-Shot:
This PR implements [MSC3759](https://github.com/matrix-org/matrix-spec-proposals/pull/3759), which includes an extra key on the leave membership events sent when deactivating a user.

N.B. I couldn't find any tests that tested the existing behavior, so some pointers on how to write those tests would be good :)

### DIFF
--- a/synapse/config/experimental.py
+++ b/synapse/config/experimental.py
@@ -81,3 +81,6 @@ class ExperimentalConfig(Config):
 
         # MSC2654: Unread counts
         self.msc2654_enabled: bool = experimental.get("msc2654_enabled", False)
+
+        # MSC3759: Leave event metadata for deactivated users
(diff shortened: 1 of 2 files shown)


richvdh (changes requested):
I'm not *entirely* convinced by the change from the spec POV, but this certainly seems to do what the MSC says, so that's fine.

In terms of writing a test: there's a `DeactivateAccountTestCase` in `tests.rest.admin.test_user`. It looks like you'll need to add a new test-case that starts by joining the user to a room (by, for example, calling `self.helper.create_room_as`).

Half-Shot:
I'll await spec feedback before plowing on for the moment, but thanks for the hints.

Issue #10304, opened 2021-07-02
Wrong type for /publicRooms limit


timokoesters:
According to https://spec.matrix.org/unstable/server-server-api/#get_matrixfederationv1publicrooms it should be an integer, but here it is stringified:

https://github.com/matrix-org/synapse/blob/a7a37437bc0364d6cde93f3ec264e06ed6324068/synapse/federation/transport/client.py#L425

reivilibre:
Note that there are two issues here:

1. that Synapse sends strings against the spec
2. that Synapse accepts strings (i.e. doesn't enforce the spec).

Fixing the first should be easy; fixing the second probably wants some lenience to allow other Synapse servers to upgrade.

PR #12364, opened 2022-04-04
Fix a spec compliance issue where requests to the `/publicRooms` federation API would specify `limit` as a string.

reivilibre:
Fixes the sending half of #10304.

Synapse is able to **accept** either strings or integers (and likely floats and bools ... sigh), but should really only emit spec-approved integers.

Enforcing this requirement on the receiving side is out of scope for this PR, but we might consider it someday, after giving enough time for Synapse versions to spread throughout the network.

### DIFF
--- a/changelog.d/12364.bugfix
+++ b/changelog.d/12364.bugfix
@@ -1,0 +1,1 @@
+Fix a spec compliance issue where requests to the `/publicRooms` federation API would specify `limit` as a string.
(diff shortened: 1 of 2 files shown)


DMRobertson (approved):
Should sytest/complement have caught this? I.e. is there a gap somewhere in the test coverage?

reivilibre:
> Should sytest/complement have caught this? I.e. is there a gap somewhere in the test coverage?

Indeed; I have a list of 2 shortcomings that I intend to add to Complement when I learn how — this is one of them.

PR #12367, opened 2022-04-04
Prefill the device_list_stream_cache

erikjohnston:
This makes restarting the synchrotrons less of a pain


reivilibre (approved):

erikjohnston:
The issue was that `/sync` has an expensive code path if a user comes along with a device list stream token that is older than those in the associated stream change cache. When restarting the cache started off empty, meaning that *all* `/sync` requests immediately hit the bad code path just after restart. By prefilling the cache we reduce the number of `/sync` requests that hit that failure mode.

PR #12368, opened 2022-04-04
Add a manual documenting config file options

H-Shay:
This is the first step to fixing #8159. Once this documentation is approved, I can move forward with pulling out all the redundant documentation from the configuration file itself. 

A few things: 
-For the most part the description of each config setting is pulled directly from the comments in the original document, I tried not change it too much, most of the changes I made were to add context or to clarify the documentation in a different context. If you see anything that is unclear please flag it and I can change it. 
 -The document is huge. It probably makes sense to have people only review a section at a time (there is an index of sections at the beginning of the file). If you review a section, please note in the comment section below which section you looked at, so others don't duplicate your work.
-I've been staring at this document for far too long and I am sure I have missed spelling/grammar errors or urls which should be converted to links. Please keep an eye out for them!


docs/usage/configuration/config_documentation.md
```diff
@@ -0,0 +1,3464 @@
...
+
+[Push](#push)
+
+[Rooms](#rooms)
+
+[Opentracing](#opentracing)
+
+[Workers](#workers)
+
+[Background Updates](#background-updates)
```
babolivier, line 115: I don't think an index is necessary since the docs website already provides automatically.

docs/usage/configuration/config_documentation.md
```diff
@@ -0,0 +1,3464 @@
+## Welcome
+
+This is intended as a guide to the synapse configuration. The behavior of a Synapse instance can be modified 
+through the many configuration settings documented here-each config option is explained,
```
richvdh, line 4:
```suggestion
through the many configuration settings documented here — each config option is explained, 
```

docs/usage/configuration/config_documentation.md
```diff
@@ -0,0 +1,3464 @@
+## Welcome
```
babolivier, line 1: This could probably be changed into `Configuring Synapse` or something like that.

babolivier (commented):
That's some impressive work, thanks Shay!

Some drive-by comments having a very quick look at it:

* I think it'd be very useful to be able to link to specific configuration settings, so maybe consider making their names titles
* the page needs to be added to `SUMMARY.md` so it can appear on the configuration website

(plus the two I've added on the diff)

docs/usage/configuration/config_documentation.md
```diff
@@ -0,0 +1,3464 @@
...
+
+m = minute
+
+h = hour
+
+d = day
+
+w = week
+
+y = year
```
richvdh, line 26: A bullet list might make this clearer:

```suggestion
* `s` = second
* `m` = minute
* `h` = hour
* `d` = day
* `w` = week
* `y` = year
```

richvdh (approved):
this is fantastic, thank you so much!

anoadragon453 (commented):
I would love the ability to be able to link to specific config options, rather than only section headers. To do this, we'd need to create a header for each config option, something like:

```
#### Config option: `log_config`
```

While this works, it unfortunately will result in adding every config option to the table of contents sidebar, which makes it less useful.

Creating a method of marking headers in our documentation as "do not add this to the table of contents" would be worthwhile in a few cases, and something I can add to [the table of contents code](https://github.com/matrix-org/synapse/blob/fd9856e4a98fb3fa9c139317b0a3b79f22aff1c7/docs/website_files/table-of-contents.js#L74-L76). We could then later apply it to this document to allow granular linking.

docs/usage/configuration/config_documentation.md
```diff
@@ -0,0 +1,3464 @@
+## Welcome
+
+This is intended as a guide to the synapse configuration. The behavior of a Synapse instance can be modified 
+through the many configuration settings documented here-each config option is explained, 
+including what the default is, how to change the default and what sort of behavior the setting governs.
```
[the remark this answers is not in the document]
H-Shay: fancy!

docs/usage/configuration/config_documentation.md
```diff
@@ -0,0 +1,3464 @@
...
+
+Example configuration:
+```
+modules:
+  - module: my_super_module.MySuperClass
+    config:
+      do_thing: true
+  - module: my_other_super_module.SomeClass
+    config: {}
+```
```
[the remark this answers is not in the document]
H-Shay: Thank you for the script snippet!

H-Shay:
> I would love the ability to be able to link to specific config options, rather than only section headers. To do this, we'd need to create a header for each config option, something like:
> 
> ```
> #### Config option: `log_config`
> ```
> 
> While this works, it unfortunately will result in adding every config option to the table of contents sidebar, which makes it less useful.
> 
> Creating a method of marking headers in our documentation as "do not add this to the table of contents" would be worthwhile in a few cases, and something I can add to [the table of contents code](https://github.com/matrix-org/synapse/blob/fd9856e4a98fb3fa9c139317b0a3b79f22aff1c7/docs/website_files/table-of-contents.js#L74-L76). We could then later apply it to this document to allow granular linking.

Would something like adding a custom html tag to each option be an acceptable solution? I.e.:
```
<h7>Config option: `server_name`</h7>
```
The tags aren't visible when it's rendered, but linking still works (I built the book locally to verify that), and we could add a little line in [here](https://github.com/matrix-org/synapse/blob/fd9856e4a98fb3fa9c139317b0a3b79f22aff1c7/docs/website_files/table-of-contents.js#L84) that says something like "skip me if I'm an H7"?

PR #12370, opened 2022-04-05
Generate historic pagination token for `/messages` when no `?from` token provided

MadLittleMods:
Generate historic pagination (`tXXX-XXX_...`) token for `/messages` when no `?from` token provided. Previously, it generated a live pagination token (`sXXX_...`) but always sorted by `topological_ordering` anyway when querying for messages.

For more documentation around historic vs live pagination tokens, see https://github.com/matrix-org/synapse/pull/12317

Part of https://github.com/matrix-org/synapse/issues/12281

Context: https://github.com/matrix-org/synapse/pull/12319#discussion_r840276412

To be clear, this PR in itself doesn't fix any bug. But when combined with the changes from https://github.com/matrix-org/synapse/pull/12319, it does fix a bug with `/messages` was previously masked. See the "Context" section below.

**The first thing to decide is whether we like the fix from https://github.com/matrix-org/synapse/pull/12319. If we like the fix there, then we also need to consider the fix here to fix the bug that is uncovered.**

But this PR does update `/messages` to do the right thing imo so it would be a good change regardless. We should be using historic tokens as the default for `/messages`.

---

Notice that `start` is now a historic token.

`/messages?dir=b`
| Before | After |
| --- | --- |
| <pre>{<br>    "chunk": [...],<br>    "start": "s962_0_0_0_0_0_0_0_0",<br>    "end": "t181-536_0_0_0_0_0_0_0_0"<br>}</pre> | <pre>{<br>    "chunk": [...],<br>    "start": "t190-962_0_0_0_0_0_0_0_0",<br>    "end": "t181-536_0_0_0_0_0_0_0_0"<br>}</pre> |

`/messages?dir=f`
| Before | After |
| --- | --- |
| <pre>{<br>    "chunk": [],<br>    "start": "s962_0_0_0_0_0_0_0_0",<br>    "end": "s962_0_0_0_0_0_0_0_0"<br>}</pre> | <pre>{<br>    "chunk": [],<br>    "start": "t190-962_0_0_0_0_0_0_0_0",<br>    "end": "t190-962_0_0_0_0_0_0_0_0"<br>}</pre> |

### Context

Context: https://github.com/matrix-org/synapse/pull/12319#discussion_r840276412

In https://github.com/matrix-org/synapse/pull/12319, [`get_state_at(since_token)`](https://github.com/matrix-org/synapse/blob/287a9c1e20e1e271d69be581d6bb8ee5a8de9d88/synapse/handlers/sync.py#L673-L679) -> [`get_recent_events_for_room`](https://github.com/matrix-org/synapse/blob/287a9c1e20e1e271d69be581d6bb8ee5a8de9d88/synapse/storage/databases/main/stream.py#L654-L657) -> [`_paginate_room_events_txn`](https://github.com/matrix-org/synapse/blob/287a9c1e20e1e271d69be581d6bb8ee5a8de9d88/synapse/storage/databases/main/stream.py#L1095) is being updated to sort according to the type of token that it's given. So if a historic token is provided, sort by `topological_ordering` first, or if it's a live token, sort by `stream_ordering` first. https://github.com/matrix-org/synapse/pull/12319 aims to fix problems with our incremental `/sync` not including the `state` it should (a [known FIXME comment](https://github.com/matrix-org/synapse/blob/f0b03186d96305fd44d74a89bf4230beec0c5c31/synapse/handlers/sync.py#L664-L667)).

But that PR also unearthed a bug that's been hiding in `/messages`. The problem was previously masked because even though we [generated a](https://github.com/matrix-org/synapse/blob/bebf994ee804ef63ce16801c6694713fcd685320/synapse/handlers/pagination.py#L444) [live token](https://github.com/matrix-org/synapse/blob/bebf994ee804ef63ce16801c6694713fcd685320/synapse/storage/databases/main/stream.py#L404-L429) (`stream_ordering`) before, it still [always sorted by `topological_ordering`](https://github.com/matrix-org/synapse/blob/287a9c1e20e1e271d69be581d6bb8ee5a8de9d88/synapse/storage/databases/main/stream.py#L1231). We expect to get messages by `topological_ordering` when calling `/messages` by default. Although it should work if it's provided with a live token in the `?from` or `?to` query parameters.

>  - `/messages` [...] return them in the order determined by the event graph `(topological_ordering, stream_ordering)`.
 >
 > *-- https://github.com/matrix-org/synapse/blob/f608e6c8cf48d8d5327c5ac04746f4f7f3992e31/docs/development/room-dag-concepts.md*

When you use the changes from https://github.com/matrix-org/synapse/pull/12319 with `/messages` without a `?from` token, it can give messages in the wrong order in some situations because it now sorts by `stream_ordering` because the token generated when you don't provide one is a live token (`stream_ordering`). Say the room has less than 100 messages which all fit within the `/messages?dir=b&limit=100` response; all of the federation backfilled messages with negative `stream_ordering` will be at the end of the response even though they should be intertwined according to their `topological_ordering`(`depth`). The same problem occurs with MSC2716 batches because they also have negative `stream_ordering`.

So this PR, aims to fix the `/messages` sorting problem by generating a historic pagination token to have as the starting bounds (known as [`from_token` in the code](https://github.com/matrix-org/synapse/blob/bebf994ee804ef63ce16801c6694713fcd685320/synapse/handlers/pagination.py#L444)).

### Dev notes

Add `"tabulate>=0.8.9",` to [`synapse/python_dependencies.py#L91`](https://github.com/matrix-org/synapse/blob/bebf994ee804ef63ce16801c6694713fcd685320/synapse/python_dependencies.py#L91)

Maybe can format tabular data without a dependency: https://stackoverflow.com/a/9536084/796832

```py
logger.info(
    "asdf_get_debug_events_in_room_ordered_by_depth %s",
    await self.store.asdf_get_debug_events_in_room_ordered_by_depth(
        room_id
    ),
)
logger.info(
    "asdf_get_debug_events_in_room_ordered_by_stream_ordering %s",
    await self.store.asdf_get_debug_events_in_room_ordered_by_stream_ordering(
        room_id
    ),
)


from tabulate import tabulate
            
async def asdf_get_debug_events_in_room_ordered_by_depth(self, room_id: str) -> Any:
    """Gets the topological token in a room after or at the given stream
    ordering.
    Args:
        room_id
    """
    sql = (
        "SELECT depth, stream_ordering, type, state_key, event_id FROM events"
        " WHERE events.room_id = ?"
        " ORDER BY depth DESC, stream_ordering DESC;"
    )
    rows = await self.db_pool.execute(
        "asdf_get_debug_events_in_room_ordered_by_depth", None, sql, room_id
    )

    headers = ["depth", "stream_ordering", "type", "state_key", "event_id"]
    return tabulate(rows, headers=headers)

async def asdf_get_debug_events_in_room_ordered_by_stream_ordering(self, room_id: str) -> Any:
    """Gets the topological token in a room after or at the given stream
    ordering.
    Args:
        room_id
    """
    sql = (
        "SELECT depth, stream_ordering, type, state_key, event_id FROM events"
        " WHERE events.room_id = ?"
        " ORDER BY stream_ordering DESC, depth DESC;"
    )
    rows = await self.db_pool.execute(
        "asdf_get_debug_events_in_room_ordered_by_depth", None, sql, room_id
    )

    headers = ["depth", "stream_ordering", "type", "state_key", "event_id"]
    return tabulate(rows, headers=headers)
```

### Pull Request Checklist

### DIFF
--- a/changelog.d/12370.bugfix
+++ b/changelog.d/12370.bugfix
@@ -1,0 +1,1 @@
+Fix `/messages` returning backfilled and [MSC2716](https://github.com/matrix-org/synapse/pull/12319) historic messages our of order.
--- a/synapse/handlers/room.py
+++ b/synapse/handlers/room.py
@@ -1445,7 +1445,7 @@ class RoomEventSource(EventSource[RoomStreamToken, EventBase]):
         return self.store.get_room_max_token()
 
     def get_current_key_for_room(self, room_id: str) -> Awaitable[RoomStreamToken]:
-        return self.store.get_room_events_max_id(room_id)
+        return self.store.get_current_room_stream_token_for_room_id(room_id)
 
 
 class ShutdownRoomResponse(TypedDict):
--- a/synapse/handlers/room.py
+++ b/synapse/handlers/room.py
@@ -1444,7 +1444,7 @@ class RoomEventSource(EventSource[RoomStreamToken, EventBase]):
     def get_current_key(self) -> RoomStreamToken:
         return self.store.get_room_max_token()
 
-    def get_current_key_for_room(self, room_id: str) -> Awaitable[str]:
+    def get_current_key_for_room(self, room_id: str) -> Awaitable[RoomStreamToken]:
         return self.store.get_room_events_max_id(room_id)
 
 
--- a/scripts-dev/complement.sh
+++ b/scripts-dev/complement.sh
@@ -71,4 +71,4 @@ fi
 
 # Run the tests!
 echo "Images built; running complement"
-go test -v -tags synapse_blacklist,msc2716,msc3030 -count=1 $EXTRA_COMPLEMENT_ARGS ./tests/
+go test -v -tags synapse_blacklist,msc2716,msc3030 -count=1 $EXTRA_COMPLEMENT_ARGS ./tests/...
--- a/scripts-dev/complement.sh
+++ b/scripts-dev/complement.sh
@@ -71,4 +71,4 @@ fi
 
 # Run the tests!
 echo "Images built; running complement"
-go test -v -tags synapse_blacklist,msc2716,msc3030 -count=1 $EXTRA_COMPLEMENT_ARGS ./tests/...
+go test -v -tags synapse_blacklist,msc2716,msc3030 -count=1 $EXTRA_COMPLEMENT_ARGS ./tests/
--- a/synapse/handlers/pagination.py
+++ b/synapse/handlers/pagination.py
@@ -446,49 +446,9 @@ class PaginationHandler:
                     room_id
                 )
             )
+            # We expect `/messages` to use historic pagination tokens by default but
+            # `/messages` should still works with live tokens when manually provided.
             assert from_token.room_key.topological
-            # from_live_token = (
-            #     self.hs.get_event_sources().get_current_token_for_pagination()
-            # )
-            # # Convert the live token (sXXX) into a historic token (tXXX-XXX)
-            # # which is more suitable for /messages.
-            # current_stream_ordering = from_live_token.room_key.stream
-            # current_topographical_ordering = (
-            #     await self.store.get_current_topological_token(
-            #         room_id, current_stream_ordering
-            #     )
-            # )
-            # from_token = from_live_token.copy_and_replace(
-            #     "room_key",
-            #     RoomStreamToken(
-            #         current_topographical_ordering, current_stream_ordering
-            #     ),
-            # )
-            # logger.info(
-            #     "get_messages(room_id=%s)\n\tfrom_token=%s\n\tcurrent_stream_ordering=%s\n\tcurrent_topographical_ordering=%s\n\tfrom_live_token=%s",
-            #     room_id,
-            #     await from_token.to_string(self.store),
-            #     current_stream_ordering,
-            #     current_topographical_ordering,
-            #     await from_live_token.to_string(self.store),
-            # )
-            logger.info(
-                "get_messages(room_id=%s)\n\tfrom_token=%s",
-                room_id,
-                await from_token.to_string(self.store),
-            )
-            logger.info(
-                "asdf_get_debug_events_in_room_ordered_by_depth %s",
-                await self.store.asdf_get_debug_events_in_room_ordered_by_depth(
-                    room_id
-                ),
-            )
-            logger.info(
-                "asdf_get_debug_events_in_room_ordered_by_stream_ordering %s",
-                await self.store.asdf_get_debug_events_in_room_ordered_by_stream_ordering(
-                    room_id
-                ),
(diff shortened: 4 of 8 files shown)


synapse/handlers/room.py
```diff
@@ -1444,7 +1444,7 @@ async def get_new_events(
     def get_current_key(self) -> RoomStreamToken:
         return self.store.get_room_max_token()
 
-    def get_current_key_for_room(self, room_id: str) -> Awaitable[str]:
+    def get_current_key_for_room(self, room_id: str) -> Awaitable[RoomStreamToken]:
         return self.store.get_room_events_max_id(room_id)
```
MadLittleMods, line 1448: Previously, `get_current_key_for_room` was unused along with `get_room_events_max_id`.

It seemed almost exactly what I was looking for so I repurposed both functions with proper `RoomStreamToken` types instead of returning the string version that was interpolated together.

changelog.d/12370.bugfix
```diff
@@ -0,0 +1 @@
+Fix `/messages` returning backfilled and [MSC2716](https://github.com/matrix-org/synapse/pull/12319) historic messages our of order.
```
MadLittleMods, line 1: We can probably use the same changelog as https://github.com/matrix-org/synapse/pull/12319 as this PR doesn't actually fix anything without additionally merging that PR.

In any case, this PR does update `/messages` to do the right thing imo so it would be a good change regardless.

synapse/handlers/room.py
```diff
@@ -1444,8 +1444,8 @@ async def get_new_events(
     def get_current_key(self) -> RoomStreamToken:
         return self.store.get_room_max_token()
 
-    def get_current_key_for_room(self, room_id: str) -> Awaitable[str]:
-        return self.store.get_room_events_max_id(room_id)
+    def get_current_key_for_room(self, room_id: str) -> Awaitable[RoomStreamToken]:
+        return self.store.get_current_room_stream_token_for_room_id(room_id)
```
MadLittleMods, line 1448: Previously, `get_current_key_for_room` was unused along with `get_room_events_max_id`.

It seemed almost exactly what I was looking for so I repurposed both functions with proper `RoomStreamToken` types instead of returning the string version that was interpolated together.

MadLittleMods:
> I think this PR makes sense independently of https://github.com/matrix-org/synapse/pull/12319.

Agreed 👍 But I think we need to update the changelog https://github.com/matrix-org/synapse/pull/12370#discussion_r842253671

Will send another PR for that ⏩

babolivier:
If someone (like me) is wondering why we're allowing `/messages` requests without a `from` even though the spec says it's required, it's because we're following [MSC3567](https://github.com/matrix-org/matrix-spec-proposals/pull/3567) which just hasn't made it into a release of the spec yet.

PR #12372, opened 2022-04-05
Prefill more stream change caches.

erikjohnston:
We also make it so that `get_cache_dict` handles large gaps in stream IDs better (for `device_lists_stream` table).

### DIFF
--- a/synapse/storage/database.py
+++ b/synapse/storage/database.py
@@ -2030,24 +2030,29 @@ class DatabasePool:
         max_value: int,
         limit: int = 100000,
     ) -> Tuple[Dict[Any, int], int]:
-        # Fetch a mapping of room_id -> max stream position for "recent" rooms.
-        # It doesn't really matter how many we get, the StreamChangeCache will
-        # do the right thing to ensure it respects the max size of cache.
-        sql = (
-            "SELECT %(entity)s, MAX(%(stream)s) FROM %(table)s"
-            " WHERE %(stream)s > ? - %(limit)s"
-            " GROUP BY %(entity)s"
-        ) % {
-            "table": table,
-            "entity": entity_column,
-            "stream": stream_column,
-            "limit": limit,
-        }
+        """Gets roughly the last N changes in the given stream table as a
+        map from entity to the stream ID of the most recent change.
+
+        Also returns the minimum stream ID.
+        """
+
+        # This may return many rows for the same entity, but the `limit` is only
+        # a suggestion so we don't care that much.
+        sql = f"""
+            SELECT {entity_column}, {stream_column}
+            FROM {table}
+            ORDER BY {stream_column} DESC
+            LIMIT ?
+        """
 
         txn = db_conn.cursor(txn_name="get_cache_dict")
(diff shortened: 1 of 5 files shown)


synapse/storage/database.py
```diff
@@ -2030,24 +2030,29 @@ def get_cache_dict(
...
+        """
+
+        # This may return many rows for the same entity, but the `limit` is only
+        # a suggestion so we don't care that much.
+        sql = f"""
+            SELECT {entity_column}, {stream_column}
+            FROM {table}
+            ORDER BY {stream_column} DESC
+            LIMIT ?
+        """
```
babolivier, line 2046: This change means that if an entity has many updates to its name (and others don't) we will end up with a smaller cache than before, right? Are we fine with that (eg because the limit is high enough that we don't really have to care)?

changelog.d/12372.feature
```diff
@@ -0,0 +1 @@
+Reduce overhead of restarting synchrotrons.
```
babolivier, line 1: Doesn't it actually create more overhead, since we need to pull thousands of rows out of the database at startup?

erikjohnston: Yeah, so the reason this works is that we do the work once at start up, rather than many times for lots of different requests. The key thing here is that the stream change caches do not get updated on reads, but on the writes (or when we prefill them on startup).

Also c.f. https://github.com/matrix-org/synapse/pull/12367#issuecomment-1088574192 why prefilling is good in general

erikjohnston:
OH BLAST. I've just realised that this doesn't work for streams where we can have multiple rows per stream ID. GAH.

synapse/storage/database.py
erikjohnston: Before we were looking for rows in the range `[MAX - LIMIT, MAX]`, which will return less than `LIMIT` ¹, so I think this will always return more rows than before.

¹ ... except if we have rows with the same stream ID, in which case this breaks.

erikjohnston:
> OH BLAST. I've just realised that this doesn't work for streams where we can have multiple rows per stream ID. GAH.

I've fixed this by simply adding one to the returned minimum stream ID

changelog.d/12372.feature
babolivier: Yes, I get why prefilling is a good thing, I'm just curious about the changelog claiming it reduces the restart the overhead of restarting synchrotrons but it actually increases it.

babolivier (approved):

changelog.d/12372.feature
erikjohnston: Ah, I think I see what you mean: yes it does cause the initial restart to take longer, but then the first 10 minutes of up time uses much less DB and CPU. So I mean more along the lines of "reduce resource usage for the first few minutes after restart". Not sure if there is a better way of wording the changelog?

babolivier: >Not sure if there is a better way of wording the changelog?

Don't think there is no, at least not in the way that would be user-friendly.

PR #12376, opened 2022-04-05
Allow specifying the Postgres database's port when running unit tests with Postgres.

reivilibre:
Due to weirdness, my local dev machine has its Postgres UNIX socket specified as number 5433 rather than 5432,
so I need to be able to specify the port to run the tests locally.

This adds an environment variable to the tests that allows you to specify the Postgres port.

I also update the documentation and describe how to use this environment variable, as well as the fact that most of these
are optional if you're using a well-configured Postgres server with a UNIX socket, as is common on Debians/Ubuntus.

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -35,6 +35,11 @@ LEAVE_DB = os.environ.get("SYNAPSE_LEAVE_DB", False)
 POSTGRES_USER = os.environ.get("SYNAPSE_POSTGRES_USER", None)
 POSTGRES_HOST = os.environ.get("SYNAPSE_POSTGRES_HOST", None)
 POSTGRES_PASSWORD = os.environ.get("SYNAPSE_POSTGRES_PASSWORD", None)
+POSTGRES_PORT = (
+    int(os.environ["SYNAPSE_POSTGRES_PORT"])
+    if "SYNAPSE_POSTGRES_PORT" in os.environ
+    else None
+)
 POSTGRES_BASE_DB = "_synapse_unit_tests_base_%s" % (os.getpid(),)
 
 # When debugging a specific test, it's occasionally useful to write the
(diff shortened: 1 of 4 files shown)


tests/utils.py
```diff
@@ -35,6 +35,11 @@
 POSTGRES_USER = os.environ.get("SYNAPSE_POSTGRES_USER", None)
 POSTGRES_HOST = os.environ.get("SYNAPSE_POSTGRES_HOST", None)
 POSTGRES_PASSWORD = os.environ.get("SYNAPSE_POSTGRES_PASSWORD", None)
+POSTGRES_PORT = (
+    int(os.environ["SYNAPSE_POSTGRES_PORT"])
+    if "SYNAPSE_POSTGRES_PORT" in os.environ
+    else None
+)
```
DMRobertson, line 42: tfw no `Option::map` :(

tests/utils.py
```diff
@@ -73,6 +79,7 @@ def setupdb():
             database=POSTGRES_BASE_DB,
             user=POSTGRES_USER,
             host=POSTGRES_HOST,
+            port=POSTGRES_PORT,
```
DMRobertson, line 82: I'm assuming `port=None` is the default behaviour. CI seems to be making good progress anyway...

DMRobertson (approved):
I think there is a loose convention that test-only env vars are prefixed with `SYNAPSE_TEST_`, e.g. `SYNAPSE_TEST_PERSIST_SQLITE_DB`, `SYNAPSE_TEST_LOG_LEVEL`, `SYNAPSE_TEST_PATCH_LOG_CONTEXTS`.

Oh, but the existing test PG env vars ignore this. Okay, fair enough.

tests/utils.py
reivilibre: Yes, very much so!!

tests/utils.py
reivilibre: Yeah, indeed, psycopg2 is nicely designed that way

PR #12380, opened 2022-04-05
Add opentracing spans to calls to external cache

erikjohnston:
I don't think redis is causing any issues, but we should be adding spans whenever we talk to an external service.

### DIFF
--- a/synapse/replication/tcp/external_cache.py
+++ b/synapse/replication/tcp/external_cache.py
@@ -96,7 +96,7 @@ class ExternalCache:
 
         with opentracing.start_active_span(
             "ExternalCache.set",
-            tags={opentracing.SynapseTags: cache_name},
+            tags={opentracing.SynapseTags.CACHE_NAME: cache_name},
         ):
             with response_timer.labels("set").time():
                 return await make_deferred_yieldable(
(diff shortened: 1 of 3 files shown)


erikjohnston:
(CI is broken on develop)

synapse/replication/tcp/external_cache.py
```diff
@@ -94,7 +94,10 @@ async def set(self, cache_name: str, key: str, value: Any, expiry_ms: int) -> No
 
         logger.debug("Caching %s %s: %r", cache_name, key, encoded_value)
 
-        with opentracing.start_active_span("ExternalCache.set"):
+        with opentracing.start_active_span(
+            "ExternalCache.set",
+            tags={opentracing.SynapseTags: cache_name},
```
clokep, line 99: I assume this should be?

```suggestion
            tags={opentracing.SynapseTags.CACHE_NAME: cache_name},
```

synapse/replication/tcp/external_cache.py
```diff
@@ -110,7 +113,10 @@ async def get(self, cache_name: str, key: str) -> Optional[Any]:
         if self._redis_connection is None:
             return None
 
-        with opentracing.start_active_span("ExternalCache.get"):
+        with opentracing.start_active_span(
+            "ExternalCache.get",
+            tags={opentracing.SynapseTags: cache_name},
```
clokep, line 118:
```suggestion
            tags={opentracing.SynapseTags.CACHE_NAME: cache_name},
```

synapse/replication/tcp/external_cache.py
erikjohnston: :facepalm:

clokep (approved):

PR #12381, opened 2022-04-05
Poetry: initial pyproject.toml and poetry.lock

DMRobertson:
As a first step towards poetry, commit its configuration and lockfile. Crucially, we leave the setuptools configuration in place and declare that our `[build-system]` is setuptools. This means that poetry isn't used unless one explicitly goes out of their way to run `poetry install`.

This is not a sustainable state to be in, because now the project metadata is duplicated in two places.

CI seems to be happy and this will allow us to break up the review.

I am assuming that:
- the twisted trunk CI job will continue to run successfully via setuptools
- the full set of debian builds succeed (only the debian sid build runs on PRs)

Xref #11537 .

### DIFF
--- a/changelog.d/12381.misc
+++ b/changelog.d/12381.misc
@@ -1,0 +1,1 @@
+Use `poetry` to manage Synapse's dependencies.
--- a/changelog.d/12381.misc
+++ b/changelog.d/12381.misc
@@ -1,1 +1,1 @@
-Use `poetry` to manage Synapse's dependencies.
+Lay groundwork for using `poetry` to manage Synapse's dependencies.
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,10 @@ _trial_temp*/
 .DS_Store
 __pycache__/
 
+# We do want the poetry lockfile. TODO: is there a good reason for ignoring
+# '*.lock' above? If not, let's nuke it.
+!poetry.lock
+
 # stuff that is likely to exist when you run a server locally
 /*.db
 /*.log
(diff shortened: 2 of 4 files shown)


changelog.d/12381.misc
```diff
@@ -0,0 +1 @@
+Use `poetry` to manage Synapse's dependencies.
```
richvdh, line 1: Let's start with something slightly less grand:

```suggestion
Lay groundwork for using `poetry` to manage Synapse's dependencies.
```

Hopefully we can get the rest of this landed in time for a release, in which case we can update the changelog entries.

richvdh (approved):
yes!

DMRobertson:
For completeness:

- sytest [is running under poetry](https://github.com/matrix-org/synapse/runs/5837441497?check_suite_focus=true#step:5:67), thanks to Sean's work in https://github.com/matrix-org/sytest/pull/1220
- tox is ignoring the lockfile (e.g. CI installs frozendict 2.3.1. but 2.3.0 is the locked version)
- naked pip installs, e.g. those done by portdb, seem to be ignoring the lockfile too. ([compare frozendict versions](https://github.com/matrix-org/synapse/runs/5837441045?check_suite_focus=true#step:6:190))
- no signs of poetry in the `complement` job
- I have kicked off a [twisted trunk run](https://github.com/matrix-org/synapse/actions/runs/2097566201) manually, which seems to be making good progress
- debian build [suceeded](https://github.com/matrix-org/synapse/runs/5837407235?check_suite_focus=true) and [used the new version of frozendict](https://github.com/matrix-org/synapse/runs/5837407235?check_suite_focus=true#step:6:239)

DMRobertson:
@richvdh I'd forgotten there were other commits which change pyproject.toml on #12337. I've now cherry-picked the relevant changes over. Could you take another look?

PR #12384, opened 2022-04-05
Add missing dependency on importlib_metadata

DMRobertson:
Missed in #12088. We got away with it because we were indirectly
requiring it via other packages.

Without this change, we encounter pain in `poetry export`; and besides,
we should be explicit about our direct dependencies.

### DIFF
--- a/synapse/python_dependencies.py
+++ b/synapse/python_dependencies.py
@@ -89,6 +89,9 @@ REQUIREMENTS = [
     "matrix-common~=1.1.0",
     # We need packaging.requirements.Requirement, added in 16.1.
     "packaging>=16.1",
+    # At the time of writing, we only use functions from the version `importlib.metadata`
+    # which shipped in Python 3.8. This corresponds to version 1.4 of the backport.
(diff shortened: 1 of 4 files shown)


clokep (approved):
In light of #12381, does `poetry.lock` need to be updated?

DMRobertson:
> In light of #12381, does `poetry.lock` need to be updated?

Thanks---yes, it does. https://github.com/matrix-org/setup-python-poetry/pull/2 would have caught this... but that's not in place yet.

synapse/python_dependencies.py
```diff
@@ -91,7 +91,7 @@
     "packaging>=16.1",
     # At the time of writing, we only use functions from the version `importlib.metadata`
     # which shipped in Python 3.8. This corresponds to version 1.4 of the backport.
-    "importlib_metadata>=1.4; python_version < 3.8",
+    "importlib_metadata>=1.4",
```
DMRobertson, line 94: I think I was supposed to wrap `3.8` in single quotes here. Oh well.

PR #12389, opened 2022-04-06
Update type annotations for compatiblity with prometheus_client 0.14

### LINKED ISSUE #12387: CI run against Twisted trunk is failing
See https://github.com/matrix-org/synapse/actions/runs/2101100787

richvdh:
Principally, `prometheus_client.REGISTRY.register` now requires its argument to extend `prometheus_client.Collector`.

Additionally, `Gauge.set` is now annotated so that passing `Optional[int]` causes an error.

Fixes #12387.

### DIFF
--- a/changelog.d/12389.misc
+++ b/changelog.d/12389.misc
@@ -1,0 +1,1 @@
+Update type annotations for compatiblity with prometheus_client 0.14.
--- a/synapse/metrics/__init__.py
+++ b/synapse/metrics/__init__.py
@@ -18,7 +18,6 @@ import logging
 import os
 import platform
 import threading
-from abc import ABC, abstractmethod
 from typing import (
     Callable,
     Dict,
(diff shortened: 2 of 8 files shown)


squahtx (approved):

reivilibre (changes requested):
Mypy seems fine but looks like you've introduced a circular import :/. Seems fine otherwise.

EDIT: this was meant to be an approval (once the bug is fixed)

squahtx:
LGTM though CI is unhappy due to circular imports

DMRobertson:
Perhaps move `Collector` into `synapse.metrics.types`, and import that into `synapse.metrics.*`?

richvdh:
> Perhaps move Collector into synapse.metrics.types, and import that into synapse.metrics.*?

indeed. thanks!

DMRobertson:
(Sorry for wading in on this; I'm just keen to get CI going on my other branches)

PR #12391, opened 2022-04-06
Add Module API for reading global account data.

reivilibre:
This is part of my effort towards solving https://github.com/matrix-org/synapse-auto-accept-invite/issues/5.

Putting global account data is harder than it looks in a workerised setup, so I've descoped that from this PR for now, sorry!


tests/module_api/test_account_data_manager.py
```diff
@@ -110,3 +110,49 @@ def test_get_global_no_mutability(self) -> None:
...
+        self.assertEqual(
+            self.get_success(
+                self._store.get_global_account_data_by_type_for_user(
+                    self.user_id, "test.data"
+                )
+            ),
+            {"wombat": True},
+        )
+
+    def test_put_global_validation(self) -> None:
```
clokep, line 135: Should we also test bad data types for these parameters?

synapse/module_api/__init__.py
```diff
@@ -1386,3 +1396,65 @@ async def remove_room_from_public_room_list(self, room_id: str) -> None:
...
+        Added in Synapse v1.57.0.
+        """
+        self._validate_user_id(user_id)
+
+        # Ensure the user exists, so we don't just write to users that aren't there.
+        if await self._store.get_userinfo_by_id(user_id) is None:
+            raise ValueError(f"User {user_id} does not exist on this server.")
+
+        # Use the account data handler to write the account data.
+        # Crucially, this makes a replication request if necessary.
```
clokep, line 1459: I feel like noting that this works over replication doesn't really matter? We should just strive that all our code does this?

synapse/module_api/__init__.py
```diff
@@ -1386,3 +1396,65 @@ async def remove_room_from_public_room_list(self, room_id: str) -> None:
             room_id: The ID of the room.
         """
         await self._store.set_room_is_public(room_id, False)
+
+
+class AccountDataManager:
```
clokep, line 1401: You're making this a separate object just for organization purposes, I assume?

synapse/module_api/__init__.py
```diff
@@ -431,6 +433,14 @@ def public_room_list_manager(self) -> "PublicRoomListManager":
         """
         return self._public_room_list_manager
 
+    @property
+    def account_data_manager(self) -> "AccountDataManager":
```
clokep, line 437: Any reason we don't just set the object as `account_data_manager` to start with? Just to show it really is public?

synapse/module_api/__init__.py
reivilibre: Yeah, indeed — a couple of other things do this but otherwise the module API is one large blob and it feels quite disorganised. It may make it hard to split up into multiple files later on.

synapse/module_api/__init__.py
reivilibre: I don't really know, but this is a convention through the rest of the module API. Perhaps it's to make it read-only? Happy to go with the simple approach if you think it's best; that's what I would have done had there not already been this pattern here.

synapse/module_api/__init__.py
```diff
@@ -1386,3 +1396,65 @@ async def remove_room_from_public_room_list(self, room_id: str) -> None:
...
+        Added in Synapse v1.57.0.
+        """
+        self._validate_user_id(user_id)
+
+        data = await self._store.get_global_account_data_by_type_for_user(
+            user_id, data_type
+        )
+        # We clone to prevent the module accidentally mutating the dict that
+        # lives in the cache, as that could introduce nasty bugs.
+        return copy.deepcopy(data)
```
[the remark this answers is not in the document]
reivilibre: I was half-and-half on this one. I obviously don't want modules to mutate internal data.
But if you're going to give the module a clone anyway, what does freezing it buy you? They already can't damage anything, since it's a clone.

Maybe it makes more sense anyway. I guess I'll do it for now, it'll be easier to revert later than the other way around if we're not keen.

synapse/module_api/__init__.py
reivilibre: well, sure, I guess. It was just a bit of narrative so you can see what's actually going on without having to go too deep. I suppose it doesn't matter.

tests/module_api/test_account_data_manager.py
reivilibre: Good idea

synapse/module_api/__init__.py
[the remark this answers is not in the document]
clokep: Freezing it makes it immutable, so will raise errors instead of allowing mutation but doing nothing. That seems clearer to me?

synapse/module_api/__init__.py
clokep: Makes sense, just wanted to make sure there wasn't other magic going on!

synapse/module_api/__init__.py
clokep: Following what else is done is fine too! 👍

clokep (approved):
Seems reasonable.

PR #12395, opened 2022-04-06
Remove the unstable event field for `/send_join` per MSC3083.

### LINKED ISSUE #11345: /send_join response includes unstable prefix for event field

clokep:
Fixes #11345 which goes into more detail about this, but I think we can drop support for the unstable identifier in `send_join` responses from MSC3083, which was erroneously used for a bit after stabilization.

The stable identifier was added in Synapse v1.49.0; the unstable identifiers were added in Synapse v1.41.0; so 1.41.0 - 1.48.0 are potentially of concern? Of those I only see any significant set of servers on 1.42.0.

### DIFF
--- a/changelog.d/12395.misc
+++ b/changelog.d/12395.misc
@@ -1,0 +1,1 @@
+Remove an unstable identifier from [MSC3083](https://github.com/matrix-org/matrix-doc/pull/3083).
--- a/synapse/federation/federation_server.py
+++ b/synapse/federation/federation_server.py
@@ -687,8 +687,6 @@ class FederationServer(FederationBase):
         time_now = self._clock.time_msec()
         event_json = event.get_pdu_json(time_now)
         resp = {
-            # TODO Remove the unstable prefix when servers have updated.
-            "org.matrix.msc3083.v2.event": event_json,
             "event": event_json,
             "state": [p.get_pdu_json(time_now) for p in state_events],
             "auth_chain": [p.get_pdu_json(time_now) for p in auth_chain_events],
--- a/synapse/federation/transport/client.py
+++ b/synapse/federation/transport/client.py
@@ -1379,16 +1379,6 @@ class SendJoinParser(ByteParser[SendJoinResponse]):
                 prefix + "auth_chain.item",
                 use_float=True,
             ),
-            # TODO Remove the unstable prefix when servers have updated.
-            #
-            # By re-using the same event dictionary this will cause the parsing of
-            # org.matrix.msc3083.v2.event and event to stomp over each other.
-            # Generally this should be fine.
-            ijson.kvitems_coro(
-                _event_parser(self._response.event_dict),
-                prefix + "org.matrix.msc3083.v2.event",
-                use_float=True,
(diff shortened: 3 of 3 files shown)


erikjohnston:
So am I understanding correctly that servers on v1.42.0 will not be able to join restricted rooms, since they won't be able to get the correctly signed event out of the `/send_join` response?

clokep:
> So am I understanding correctly that servers on v1.42.0 will not be able to join restricted rooms, since they won't be able to get the correctly signed event out of the `/send_join` response?

Correct, any servers on Synapse v1.41.0 - v1.48.0 joining via a server that is on Synapse > v1.57.0 (or whenever this lands) would not complete properly since they would be unable to retrieve the proper signed event.

We can leave it for now, I just don't know when it has been "enough" time.

erikjohnston:
Yeah, fair. I'm not sure either, so maybe let's discuss this as a team?

clokep:
> Confirm what would occur if an old version tried to join via the new version (i.e. if the expected field was missing).

I tested this by setting up a public space and a restricted room (limited to members of the space) as `@alice:foo`. I then tried operations from `@bob:bar`. I modified Synapse to not return either the stable or unstable `event` field from MSC3083.

1. `@bob:bar` attempted to join `#restricted:foo` and received the proper message about not being a member of any of the required spaces.
2. `@bob:bar` joined `#space:foo` with no issue.
3. `@bob:bar` attempted to join `#restricted:foo` and received an error message ("Join event was rejected"). `@alice:foo` saw `@bob:bar` in the room at this point.
4. `@bob:bar` attempted to join `#restricted:foo` again and the join was successful for them.

The rationale for why this occurs is that the event is properly added by `foo` into the room on the first join, but since i this not returned `bar` uses the event they originally sent, see:

https://github.com/matrix-org/synapse/blob/v1.42.0/synapse/federation/federation_client.py#L759-L779

Unfortunately the event will fail auth rules and be rejected later on.

During the secondary join, `foo` inserts the join event that matches what is provided by `bar` (i.e. it no longer has the `join_authorised_user_id` in it) and the auth rules pass. Note that this event *does* have an auth event depending on the initial join of the user.

Being able to `/send_join` when you're already in the room seems odd, but I think is a bit unrelated to the current question.

@erikjohnston Does this assuage your concerns? I think we are fine since a second join would work, although the initial error message is confusing.

erikjohnston:
That doesn't sound like the worst failure mode then, in which case I'm happy for this to merge now.

erikjohnston (approved):

PR #12399, opened 2022-04-06
Await un-partial-stating after a partial-state join

richvdh:
When we join a room via the faster-joins mechanism, we end up with "partial state" at some points on the event DAG. Many parts of the codebase need to wait for the full state to load. So, we implement a mechanism to keep track of which events have partial state, and wait for them to be fully-populated.

There are a few commits here, which it might be easier to review individually.

This should pass against the tests in https://github.com/matrix-org/complement/pull/359, though they haven't yet landed.

Part of #11249.


synapse/handlers/federation_event.py
```diff
@@ -515,6 +515,7 @@ async def update_state_for_partial_state_event(
                 )
                 return
```
squahtx, line 516: Is there logic somewhere to re-attempt to de-partial-state successor events?

synapse/storage/state.py
```diff
@@ -579,7 +584,7 @@ async def get_state_groups_ids(
         if not event_ids:
             return {}
 
-        event_to_groups = await self.stores.main._get_state_group_for_events(event_ids)
+        event_to_groups = await self._get_state_group_for_events(event_ids)
```
squahtx, line 587: `synapse/storage/persist_events.py` has two calls to `DataStore._get_state_group_for_events`. Do we need to update those too?

synapse/storage/util/partial_state_events_tracker.py
```diff
@@ -0,0 +1,116 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class PartialStateEventsTracker:
+    """Keeps track of which events have partial state, after a partial-state join"""
+
+    def __init__(self, store: EventsWorkerStore):
+        self._store = store
+        self._observers: Dict[str, Set[Deferred]] = defaultdict(set)
```
squahtx, line 34:
```suggestion
        self._observers: Dict[str, Set[Deferred[None]]] = defaultdict(set)
```

We could also do with a comment explaining that the keys are event ids.

synapse/storage/util/partial_state_events_tracker.py
```diff
@@ -0,0 +1,116 @@
...
+
+        try:
+            # some of them may have been un-lazy-joined between us checking the db and
+            # registering the observer, in which case we'd wait forever for the
+            # notification. Call back the observers now.
+            for event_id, partial in (
+                await self._store.get_partial_state_events(observers.keys())
+            ).items():
+                if not partial:
+                    observers[event_id].callback(None)
```
squahtx, line 96: If `notify_un_partial_stated` gets called during the 2nd `get_partial_state_events` query, we can `.callback` the same `Deferred` twice, which will make twisted raise an `AlreadyCalledError`.

synapse/storage/util/partial_state_events_tracker.py
```diff
@@ -0,0 +1,116 @@
...
+
+from synapse.logging.context import PreserveLoggingContext, make_deferred_yieldable
+from synapse.storage.databases.main.events_worker import EventsWorkerStore
+from synapse.util import unwrapFirstError
+
+logger = logging.getLogger(__name__)
+
+
+class PartialStateEventsTracker:
+    """Keeps track of which events have partial state, after a partial-state join"""
```
squahtx, line 30: This only works for single process synapses for now, is that right?

synapse/handlers/federation_event.py
richvdh: yeah, [_sync_partial_state_room](https://github.com/matrix-org/synapse/blob/develop/synapse/handlers/federation.py#L1388) (which calls this method) basically works in a loop, finding remaining partial-stated events.

synapse/storage/util/partial_state_events_tracker.py
richvdh: yeees, though that's more because we don't call `notify_un_partial_stated` if another worker un-partial-states an event than anything particular to do with this class. Or are you seeing something I'm not?

synapse/storage/util/partial_state_events_tracker.py
```diff
@@ -0,0 +1,116 @@
...
+
+    def __init__(self, store: EventsWorkerStore):
+        self._store = store
+        self._observers: Dict[str, Set[Deferred[None]]] = defaultdict(set)
```
richvdh, line 34:
```suggestion
        # a map from event id to a set of Deferreds which are waiting for that event to be
        # un-partial-stated.
        self._observers: Dict[str, Set[Deferred[None]]] = defaultdict(set)
```

reivilibre (commented):
squahtx seems to have beaten me to this review, but I did already write...

synapse/storage/state.py
```diff
@@ -785,6 +790,23 @@ def _get_state_for_groups(
...
+    ) -> Mapping[str, int]:
+        """Returns mapping event_id -> state_group
+
+        Args:
+            event_ids: events to get state groups for
+            await_full_state: if true, will block if we do not yet have complete
+               state at this event.
+        """
+        if await_full_state:
+            await self._partial_state_events_tracker.await_full_state(event_ids)
```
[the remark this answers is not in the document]
richvdh: That's part of the plan. Another part of the plan is to get cancellation working (https://github.com/matrix-org/synapse/issues/3528) so that we don't stack up requests. A final part of the plan is probably to change the api to do something different, but we're a way off that yet.

synapse/storage/util/partial_state_events_tracker.py
squahtx: Just checking if I understood what's left to do. There wasn't anything else driving that comment.

squahtx (approved):
lgtm!

PR #12402, opened 2022-04-06
Add missing type hints to config classes.

clokep:
Does what it says on the tin, I originally wanted to more strictly type `generate_config_section` and `read_config`...but they get passed a bunch of arguments which  most don't use, so leaving them as `**kwargs: Any` seemed "good enough". 🤷 

Note that this creates *a lot* of `Any` things in our config (as we generally don't assert the types of anything we read from there). I think generally improving the parsing of our configuration is a follow-on task though.

### DIFF
--- a/synapse/config/key.py
+++ b/synapse/config/key.py
@@ -100,7 +100,7 @@ class KeyConfig(Config):
     section = "key"
 
     def read_config(
-        self, config: JsonDict, config_dir_path: Optional[str], **kwargs: Any
+        self, config: JsonDict, config_dir_path: str, **kwargs: Any
     ) -> None:
         # the signing key can be specified inline or in a separate file
         if "signing_key" in config:
--- a/synapse/config/key.py
+++ b/synapse/config/key.py
@@ -99,11 +99,14 @@ class TrustedKeyServer:
 class KeyConfig(Config):
     section = "key"
 
-    def read_config(self, config, config_dir_path, **kwargs):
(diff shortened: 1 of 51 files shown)


synapse/config/registration.py
```diff
@@ -196,7 +196,9 @@ def read_config(self, config, **kwargs):
 
         self.inhibit_user_in_use_error = config.get("inhibit_user_in_use_error", False)
 
-    def generate_config_section(self, generate_secrets=False, **kwargs):
+    def generate_config_section(
+        self, generate_secrets: Optional[bool] = False, **kwargs: Any
```
babolivier, line 200: I think this should be a `bool`? I don't see any place in the code passing `None` for `generate_secrets`.

clokep: I think I just based this on what `generate_config` has for type hints... let me double check.

synapse/config/key.py
```diff
@@ -99,11 +99,14 @@ class TrustedKeyServer:
 class KeyConfig(Config):
     section = "key"
 
-    def read_config(self, config, config_dir_path, **kwargs):
+    def read_config(
+        self, config: JsonDict, config_dir_path: Optional[str], **kwargs: Any
```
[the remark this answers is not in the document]
clokep: Good call! Seems like we can make it required (and `data_dir_path`).

babolivier (approved):

PR #12404, opened 2022-04-06
Move complement setup stuff into the Synapse repo

richvdh:
Fixes https://github.com/matrix-org/complement/issues/330 (or it will, once we remove the old files).

It's not *quite* a lift-and-shift: I've also taken the opportunity to get rid of the custom CA that we used to use to sign the TLS certs, which has been superceded by the CA exposed by Complement.

### DIFF
--- a/scripts-dev/complement.sh
+++ b/scripts-dev/complement.sh
@@ -50,25 +50,18 @@ if [[ -n "$WORKERS" ]]; then
 
   export COMPLEMENT_BASE_IMAGE=complement-synapse-workers
   COMPLEMENT_DOCKERFILE=SynapseWorkers.Dockerfile
+
   # And provide some more configuration to complement.
-  export COMPLEMENT_CA=true
   export COMPLEMENT_SPAWN_HS_TIMEOUT_SECS=25
 else
   export COMPLEMENT_BASE_IMAGE=complement-synapse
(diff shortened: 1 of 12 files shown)


DMRobertson (approved):

scripts-dev/complement.sh
```diff
@@ -50,25 +50,18 @@ if [[ -n "$WORKERS" ]]; then
...
 
 # Build the Complement image from the Synapse image we just built.
-docker build -t $COMPLEMENT_BASE_IMAGE -f "$COMPLEMENT_DIR/dockerfiles/$COMPLEMENT_DOCKERFILE" "$COMPLEMENT_DIR/dockerfiles"
-
-cd "$COMPLEMENT_DIR"
-
-EXTRA_COMPLEMENT_ARGS=""
-if [[ -n "$1" ]]; then
-  # A test name regex has been set, supply it to Complement
-  EXTRA_COMPLEMENT_ARGS+="-run $1 "
```
MadLittleMods, line 69: This behavior change was a bit confusing to figure out. Updated the docs in https://github.com/matrix-org/synapse/pull/12664

It looks like this was necessary to be able to do `synapse/scripts-dev/complement.sh -json 2>&1` in `.github/workflows/tests.yml` above

PR #12406, opened 2022-04-07
Add a module API to allow modules to add push rules for local users

babolivier:
This is happening in the context of the Tchap mainlining. See https://github.com/matrix-org/matrix-dinsic/issues/856 for more context.

### DIFF
--- a/synapse/module_api/__init__.py
+++ b/synapse/module_api/__init__.py
@@ -97,6 +97,14 @@ from synapse.logging.context import (
 )
 from synapse.metrics.background_process_metrics import run_as_background_process
 from synapse.rest.client.login import LoginResponse
+from synapse.rest.client.push_rule import (
+    InvalidRuleException,
+    RuleSpec,
+    _check_actions,
+    _namespaced_rule_id,
+    _namespaced_rule_id_from_spec,
+    _priority_class_from_spec,
+)
 from synapse.storage import DataStore
 from synapse.storage.background_updates import (
     DEFAULT_BATCH_SIZE_CALLBACK,
@@ -192,6 +200,7 @@ class ModuleApi:
         self._clock: Clock = hs.get_clock()
         self._registration_handler = hs.get_registration_handler()
         self._send_email_handler = hs.get_send_email_handler()
+        self._notifier = hs.get_notifier()
         self.custom_template_dir = hs.config.server.custom_template_directory
 
         try:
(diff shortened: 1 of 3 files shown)


synapse/module_api/__init__.py
```diff
@@ -1340,6 +1349,91 @@ async def store_remote_3pid_association(
         """
         await self._store.add_user_bound_threepid(user_id, medium, address, id_server)
 
+    async def add_push_rule_for_user(
```
clokep, line 1352: I think much of this method is copied from `PushRuleRestServlet` -- would it make sense to create a `PushRuleHandler` to encapsulate that logic and call it from both here and `PushRuleRestServlet`? (Parts of that class could probably be moved into the handler too...)

babolivier: Yes probably.

babolivier:
Turns out this was the wrong approach, so I changed this PR to instead implement an API for modules to change the actions for a given rule. I've also moved a few things to a push rules handler, so the module API doesn't import directly from client resources.

synapse/module_api/errors.py
```diff
@@ -16,14 +16,18 @@
...
 from synapse.config._base import ConfigError
+from synapse.handlers.push_rules import InvalidRuleException
 
 __all__ = [
     "InvalidClientCredentialsError",
     "RedirectException",
     "SynapseError",
     "ConfigError",
+    "InvalidRuleException",
+    "NotFoundError",
```
richvdh, line 32: ugh, I'm not that enthusiastic about exposing NotFoundError. It's a bit of a poster-child for our antipattern of raising HTTP errors from the depths of database code.

I see we already have a `RuleNotFoundException`. Can we use that?

synapse/handlers/push_rules.py
```diff
@@ -0,0 +1,129 @@
...
+    async def set_rule_attr(
+        self, user_id: str, spec: RuleSpec, val: Union[bool, JsonDict]
+    ) -> None:
+        """Set an attribute (enabled or actions) on an existing push rule.
+
+        Args:
+            user_id: the user for which to modify the push rule.
+            spec: the spec of the push rule to modify.
+            val: the value to change the attribute to.
+        """
```
richvdh, line 48: it looks like you're relying on this raising particular exceptions in particular situations, so that's part of the calling contract for this method. Please could you details of the exceptions to the docstring?

synapse/handlers/push_rules.py
```diff
@@ -0,0 +1,129 @@
...
+
+@attr.s(slots=True, frozen=True, auto_attribs=True)
+class RuleSpec:
+    scope: str
+    template: str
+    rule_id: str
+    attr: Optional[str]
+
+
+class PushRulesHandler:
```
richvdh, line 34: it'd be nice to give this a brief description in a docstring, even if it's just a oneliner.

synapse/handlers/push_rules.py
```diff
@@ -0,0 +1,129 @@
...
+            if is_default_rule:
+                if namespaced_rule_id not in BASE_RULE_IDS:
+                    raise SynapseError(404, "Unknown rule %r" % (namespaced_rule_id,))
+            await self._store.set_push_rule_actions(
+                user_id, namespaced_rule_id, actions, is_default_rule
+            )
+        else:
+            raise UnrecognizedRequestError()
+
+    def notify_user(self, user_id: str) -> None:
```
richvdh, line 90: docstring please

synapse/rest/client/push_rule.py
```diff
@@ -81,8 +78,12 @@ async def on_PUT(self, request: SynapseRequest, path: str) -> Tuple[int, JsonDic
...
 
         if spec.attr:
-            await self.set_rule_attr(user_id, spec, content)
-            self.notify_user(user_id)
+            try:
+                await self._push_rules_handler.set_rule_attr(user_id, spec, content)
+            except InvalidRuleException as e:
+                raise SynapseError(400, "Invalid actions: %s" % e)
+
+            self._push_rules_handler.notify_user(user_id)
```
richvdh, line 86: shouldn't notify_user happen automatically as a side-effect of `set_rule_attr`?

synapse/handlers/push_rules.py
```diff
@@ -0,0 +1,129 @@
...
+            spec: the spec of the push rule to modify.
+            val: the value to change the attribute to.
+        """
+        if spec.attr not in ("enabled", "actions"):
+            # for the sake of potential future expansion, shouldn't report
+            # 404 in the case of an unknown request so check it corresponds to
+            # a known attribute first.
+            raise UnrecognizedRequestError()
+
+        namespaced_rule_id = namespaced_rule_id_from_spec(spec)
```
richvdh, line 55: I'm sorely tempted to inline this rather than lug `namespaced_rule_id_from_spec` around

```suggestion
        namespaced_rule_id = f"global/{spec.template}/{spec.rule_id}"
```

synapse/server.py
```diff
@@ -850,3 +851,7 @@ def get_request_ratelimiter(self) -> RequestRatelimiter:
             self.config.ratelimiting.rc_message,
             self.config.ratelimiting.rc_admin_redaction,
         )
+
+    @cache_in_self
+    def get_push_rules_handler(self) -> PushRulesHandler:
+        return PushRulesHandler(self)
```
richvdh, line 857: put this next to the other Handlers, maybe?

synapse/module_api/__init__.py
```diff
@@ -1340,6 +1342,75 @@ async def store_remote_3pid_association(
...
+        Args:
+            actions: the actions to check.
+
+        Returns:
+            True if the actions are valid, False otherwise.
+        """
+        try:
+            check_actions(actions)
+        except InvalidRuleException:
+            return False
```
richvdh, line 1365: given we're exposing `InvalidRuleException` anyway, why not let the exception propagate? it will give the caller more info than a simple boolean.

richvdh (changes requested):
seems sensible, but I have a few thoughts.

dklimpel:
Is there a plan to add something like this?
- "rewrite Sygnal URL"
   https://github.com/matrix-org/synapse-dinsic/pull/69

richvdh (commented):
looks broadly good. still a few nits.

synapse/handlers/push_rules.py
```diff
@@ -37,12 +37,13 @@ class PushRulesHandler:
...
         self._notifier = hs.get_notifier()
-        self._store = hs.get_datastores().main
+        self._main_store = hs.get_datastores().main
 
     async def set_rule_attr(
         self, user_id: str, spec: RuleSpec, val: Union[bool, JsonDict]
     ) -> None:
-        """Set an attribute (enabled or actions) on an existing push rule.
+        """Set an attribute (enabled or actions) on an existing push rule, and notify
+        listeners (e.g. sync handler) of the change.
```
richvdh, line 46:
```suggestion
        """Set an attribute (enabled or actions) on an existing push rule.
        
        Notifies listeners (e.g. sync handler) of the change.
```

richvdh (approved):
lgtm!

PR #12407, opened 2022-04-07
Poetry: select olddeps using `poetry`

DMRobertson:
There are a few moving parts here.

- Once we ditch setup.py we'll need to patch the dependencies in pyproject.toml to maintain this CI job.
- If we were using poetry-core as our PEP 517 build system, it would suffice to `pip install .[all,test]`. But we're gradually transitioning to that; setuptools remains the build system for now.
- Therefore use the "full" poetry to _lock_ an olddeps environment and run using `poetry run`.

### DIFF
--- a/.ci/scripts/test_old_deps.sh
+++ b/.ci/scripts/test_old_deps.sh
@@ -1,6 +1,9 @@
 #!/usr/bin/env bash
-# this script is run by GitHub Actions in a plain `focal` container; it installs the
-# minimal requirements for tox and hands over to the py3-old tox environment.
+# this script is run by GitHub Actions in a plain `focal` container; it
+# - installs the minimal system requirements, and poetry;
+# - patches the project definition file to refer to old versions only;
+# - creates a venv with these old versions using poetry; and finally
+# - invokes `trial` to run the tests with old deps.
 
 # Prevent tzdata from asking for user input
 export DEBIAN_FRONTEND=noninteractive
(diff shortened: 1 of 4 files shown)


tox.ini
```diff
@@ -105,29 +105,7 @@ commands =
...
 usedevelop=true
 
-# A test suite for the oldest supported versions of Python libraries, to catch
-# any uses of APIs not available in them.
-[testenv:py3-old]
-skip_install = true
-usedevelop = false
-deps =
-    Automat == 0.8.0
-    lxml
```
clokep, line 115: I don't see these being moved into the new bits, but maybe that's OK?

DMRobertson: It should be okay.

- Automat isn't a direct dependency (it's not in python_dependencies,py nor pyproject.toml), so I don't think we have any business pinning it to a particular version anyway;
- lxml will be installed [it's part of `all`](https://github.com/matrix-org/synapse/blob/dd5cc37aa465df8e33fd872de2528dc53740b2a6/pyproject.toml#L217-L227), and in the config we install `all` extras
- markupsafe is also not a direct dependency.
- the base dependencies are pip, setuptools, and stuff related to `coverage`. We've stopped using coverage (and I suspect it'd need some config fiddling to have it work in poetry anyway).

.ci/scripts/test_old_deps.sh
```diff
@@ -9,12 +12,74 @@ set -ex
...
+# The sed expression 's/\^/==/g' ought to do the trick. But it would also change
+# `python = "^3.7"` to `python = "==3.7"`, which would mean we fail because olddeps
+# runs on 3.8 (#12343).
+
+# TODO: I'd prefer to use something like this
+#   https://github.com/python-poetry/poetry/issues/3527
+#   https://github.com/pypa/pip/issues/8085
+# rather than this sed script.
+
+sed -i-backup \
```
richvdh, line 43: is there any point in keeping a backup here?

```suggestion
sed -i \
```

richvdh (approved):
lgtm!

.ci/scripts/test_old_deps.sh
DMRobertson: Argh, I could have sworn I already removed this somewhere else.

(A previous version of this script diffed the pre- and post patch toml. Unfortunately `toml.dump`'s output style is different to our input, so the resulting diff isn't very meaningful.)

richvdh (approved):
lgtm otherwise

.ci/scripts/test_old_deps.sh
```diff
@@ -20,42 +20,38 @@ export LANG="C.UTF-8"
...
+#   https://github.com/python-poetry/poetry/issues/3527
+#   https://github.com/pypa/pip/issues/8085
+# to select the lowest possible versions, rather than resorting to this sed script.
+
 # Patch the project definitions in-place:
 # - Replace all lower and tilde bounds with exact bounds
-# - Make the pyopenssl 17.0, which is the oldest version that works with
-#   a `cryptography` compiled against OpenSSL 1.1.
+# - Make the pyopenssl 17.0, which can work against an
+#   OpenSSL 1.1 compiled cryptography (as older ones don't compile on Travis).
```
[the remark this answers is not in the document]
DMRobertson: It is. Thanks.

PR #12408, opened 2022-04-07
Do not add groups to sync results if disabled.

clokep:
This was missed in #12200 -- if we disable groups/communities we should stop returning them as part of sync results.

### DIFF
--- a/synapse/rest/client/sync.py
+++ b/synapse/rest/client/sync.py
@@ -301,14 +301,13 @@ class SyncRestServlet(RestServlet):
         if archived:
             response["rooms"][Membership.LEAVE] = archived
 
-        # By the time we get here groups is no longer optional.
-        assert sync_result.groups is not None
-        if sync_result.groups.join:
-            response["groups"][Membership.JOIN] = sync_result.groups.join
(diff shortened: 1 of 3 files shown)


synapse/rest/client/sync.py
```diff
@@ -301,14 +301,13 @@ async def encode_response(
         if archived:
             response["rooms"][Membership.LEAVE] = archived
 
-        # By the time we get here groups is no longer optional.
-        assert sync_result.groups is not None
```
DMRobertson, line 305: I double checked this:

`_generate_sync_entry_for_groups` was computing this and making it non-None. But we're not calling that function any more so it'll continue to be None.

(I'm not quite sure why it `.groups` was allowed to be None in the first place? But it's going away soon anyway.)

DMRobertson (approved):
LGTM! :fire:

synapse/rest/client/sync.py
clokep: Thanks for double check, that was what I found as well! 👍 (I also ran some sync tests after disabling groups and nothing 💥 ...)

PR #12409, opened 2022-04-07
Fix rejecting invites over federation

erikjohnston:
Currently causes future incremental syncs to fail.

Broke by #12191


clokep (approved):
I think?

PR #12410, opened 2022-04-07
Fix fetching public rooms over federation

erikjohnston:
Broke by #12364

### DIFF
--- a/synapse/federation/transport/client.py
+++ b/synapse/federation/transport/client.py
@@ -509,7 +509,7 @@ class TransportLayerClient:
             if third_party_instance_id:
                 args["third_party_instance_id"] = (third_party_instance_id,)
             if limit:
-                args["limit"] = [limit]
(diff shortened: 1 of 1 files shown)


synapse/federation/transport/client.py
```diff
@@ -509,7 +509,7 @@ async def get_public_rooms(
             if third_party_instance_id:
                 args["third_party_instance_id"] = (third_party_instance_id,)
             if limit:
-                args["limit"] = [limit]
+                args["limit"] = [str(limit)]
```
DMRobertson, line 512: This wasn't caught because mypy thinks `args: Dict[str, Any]`.

DMRobertson: And we even write as much on line 506.

DMRobertson: Can we add `args: QueryArgs` here?

PR #8806, opened 2020-11-24
Add type hints to matrix federation client / agent.

clokep:
This adds type hints for more of the federation client HTTP paths, in particular the matrix federation client, matrix federation agent, and the well known resolver.

This should be reviewable commit-by-commit.

PR #8372, opened 2020-09-22
Add type annotations to SimpleHttpClient

richvdh:

PR #12415, opened 2022-04-07
Unify HTTP query parameter type hints

### LINKED ISSUE #12411: Improve type hints relating to query parameters
The regression fixed by https://github.com/matrix-org/synapse/pull/12410 was not spotted because we annotated `args` as a `Dict[str, Any]` but then added `args["limit"]` as an `int`. This is fine because `int` is a subtype of `Any`.

https://github.com/matrix-org/synapse/blob/1a90c1e3af71fe57c5a4bce52c60e6cd7509b0c2/synapse/federation/transport/client.py#L506-L519

We pass args to `get_json`, which expects `args: Optional[QueryArgs]`, which is:

https://github.com/matrix-org/synapse/blob/bc9dff1d9597251a15a15475cb8e8194b2d14910/synapse/http/matrixfederationclient.py#L102

This isn't found by mypy because `args: Dict[str, Any]` means that we do no typechecking of the values in args.

All functions which accept `args: QueryArgs` forward args as the `query` attribute of `MatrixFederationRequest`. That attribute is typed more weakly as `Dict[Any, Any]`.

https://github.com/matrix-org/synapse/blob/bc9dff1d9597251a15a15475cb8e8194b2d14910/synapse/http/matrixfederationclient.py#L147-L149

The `query` attribute is seems to only ever be passed to `encode_query_args`. That expects a type similar to `QueryArgs`, and converts its argument list in a form that can be passed to `urlencode`.

https://github.com/matrix-org/synapse/blob/a121507cfec0ffce45a89f5a1019034eda5b0c70/synapse/http/client.py#L914-L935

We also have a `QueryParams` type elsewhere, which again is only forwarded to or passed to `urlencode`.

https://github.com/matrix-org/synapse/blob/a121507cfec0ffce45a89f5a1019034eda5b0c70/synapse/http/client.py#L100-L102

I would like to see:
- A stronger type annotation for `MatrixFederationRequest.params`
- Some unification of `QueryArgs` and `QueryParams`
- Some kind of audit of how we pass query parameters to `MatrixFederationRequest`
  - at the very least something that would have caught the regression fixed by #12410.

DMRobertson:
Fixes #12411. Please see that issue for my motivation; otherwise, it should be accessible commit-by-commit.

These are mostly just type-hinting changes: but please note commits 904d57fd7e6259f1455940edc7eb3bce17c2e458 and 00f22d7c258531d68846e96273ddb3706b5f9b61, which include runtime changes. Those changes are intended to be no-ops, but please do check my reasoning.

The [documentation](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlencode) and [implementation](https://github.com/python/cpython/blob/3.10/Lib/urllib/parse.py#L911) of urlencode will likely be of interest!

### DIFF
--- a/synapse/federation/transport/client.py
+++ b/synapse/federation/transport/client.py
@@ -504,7 +504,7 @@ class TransportLayerClient:
         else:
             path = _create_v1_path("/publicRooms")
 
-            args: Dict[str, Any] = {
+            args: Dict[str, Union[str, Iterable[str]]] = {
                 "include_all_networks": "true" if include_all_networks else "false"
             }
             if third_party_instance_id:
--- a/synapse/federation/transport/client.py
+++ b/synapse/federation/transport/client.py
@@ -44,6 +44,7 @@ from synapse.api.urls import (
 from synapse.events import EventBase, make_event_from_dict
 from synapse.federation.units import Transaction
 from synapse.http.matrixfederationclient import ByteParser
+from synapse.http.types import QueryParams
 from synapse.types import JsonDict
 
 logger = logging.getLogger(__name__)
@@ -255,7 +256,7 @@ class TransportLayerClient:
         self,
         destination: str,
         query_type: str,
-        args: dict,
+        args: QueryParams,
         retry_on_dns_fail: bool,
         ignore_backoff: bool = False,
         prefix: str = FEDERATION_V1_PREFIX,
--- a/synapse/http/matrixfederationclient.py
+++ b/synapse/http/matrixfederationclient.py
@@ -162,10 +162,7 @@ class MatrixFederationRequest:
 
         destination_bytes = self.destination.encode("ascii")
         path_bytes = self.path.encode("ascii")
-        if self.query:
-            query_bytes = encode_query_args(self.query)
-        else:
-            query_bytes = b""
+        query_bytes = encode_query_args(self.query)
 
         # The object is frozen so we can pre-compute this.
         uri = urllib.parse.urlunparse(
@@ -482,10 +479,7 @@ class MatrixFederationHttpClient:
         method_bytes = request.method.encode("ascii")
         destination_bytes = request.destination.encode("ascii")
         path_bytes = request.path.encode("ascii")
-        if request.query:
-            query_bytes = encode_query_args(request.query)
-        else:
-            query_bytes = b""
+        query_bytes = encode_query_args(request.query)
 
         scope = start_active_span(
             "outgoing-federation-request",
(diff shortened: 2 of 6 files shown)


synapse/federation/transport/client.py
```diff
@@ -503,7 +504,7 @@ async def get_public_rooms(
         else:
             path = _create_v1_path("/publicRooms")
 
-            args: Dict[str, Any] = {
+            args: Dict[str, Union[str, Iterable[str]]] = {
```
clokep, line 507: Does it make sense to use `MutableMapping` for `QueryParams` (or have a `MutableQueryParams`)?

synapse/http/matrixfederationclient.py
```diff
@@ -98,10 +99,6 @@
 
 _next_id = 1
 
-
-QueryArgs = Dict[str, Union[str, List[str]]]
```
clokep, line 102: We pretty much just had two implementations of the same thing?

synapse/http/types.py
```diff
@@ -0,0 +1,7 @@
+from typing import Iterable, Mapping, Union
```
clokep, line 1: License header, please.

synapse/http/matrixfederationclient.py
DMRobertson: Correct: one from #8806 and #8372. I think yourself and Rich basically ended up at the same solution.

synapse/federation/transport/client.py
DMRobertson: I think I prefer it keeping the (immutable) Mapping. The non-mutable version expresses that functions taking a `QueryParams` won't mutate those parameters. It's inconvenient here if you conditionally build up a parameter dictionary, but that's rare:  we almost always build a parameter dictionary `args` as a dictionary literal and pass it immediately to an HTTP function (without further mutating `args`).

synapse/http/matrixfederationclient.py
```diff
@@ -165,10 +162,7 @@ def __attrs_post_init__(self) -> None:
 
         destination_bytes = self.destination.encode("ascii")
         path_bytes = self.path.encode("ascii")
-        if self.query:
-            query_bytes = encode_query_args(self.query)
-        else:
-            query_bytes = b""
```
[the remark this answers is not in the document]
DMRobertson: I hadn't thought of this, but fortunately it will work:

```
$ python
Python 3.10.4 (main, Mar 25 2022, 00:00:00) [GCC 11.2.1 20220127 (Red Hat 11.2.1-9)] on linux
>>> import urllib.parse
>>> urllib.parse.urlencode({}, True)
""
```

synapse/federation/transport/client.py
clokep: That seems fine! 👍

clokep (approved):
Nice clean-up! 🌮

DMRobertson:
I'm going to manually merge. I think the sytest failure is a flake, see https://github.com/matrix-org/synapse/issues/12419

PR #12418, opened 2022-04-08
CI: Fix the export-data job to run under poetry

DMRobertson:
Pulled out from #12337.

### DIFF
--- a/.ci/scripts/test_export_data_command.sh
+++ b/.ci/scripts/test_export_data_command.sh
@@ -2,29 +2,23 @@
 
 # Test for the export-data admin command against sqlite and postgres
 
+# Expects Synapse to have been already installed with `poetry install -e .[postgres]
+
 set -xe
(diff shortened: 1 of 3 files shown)


.ci/scripts/test_export_data_command.sh
```diff
@@ -2,29 +2,23 @@
 
 # Test for the export-data admin command against sqlite and postgres
 
+# Expects Synapse to have been already installed with `poetry install -e .[postgres]
```
clokep, line 5: Should this be?

```suggestion
# Expects Synapse to have been already installed with `poetry install --extras postgres`.
# Expects `poetry` to be available on the `PATH`.
```

From #12412

DMRobertson: It definitely should be, thank you. I suppose my brain was frazzled into confusing `poetry` and `pip` at some point.

clokep (approved):

PR #12420, opened 2022-04-08
Stop maintaining a list of lint targets

DMRobertson:
instead, configure the tools to find sensible targets.

Pulled out from #12337; part (but not all) of
e53e99edba7ccda8319ebd4fe832637cbe9261cc. Related: #12107.


DMRobertson:
CI isn't going to actually use this. But it makes the poetry diff smaller.

reivilibre (approved):

PR #12421, opened 2022-04-08
Make `synapse._scripts` pass typechecks

DMRobertson:
A first pass. I plan to run a second pass to make the scripts pass `--disallow-untyped-defs`.

### DIFF
--- a/synapse/_scripts/synapse_port_db.py
+++ b/synapse/_scripts/synapse_port_db.py
@@ -21,12 +21,13 @@ import logging
 import sys
 import time
 import traceback
-from typing import Dict, Iterable, Optional, Set
+from types import TracebackType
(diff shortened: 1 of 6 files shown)


synapse/_scripts/synapse_port_db.py
```diff
@@ -236,9 +243,12 @@ def get_instance_name(self):
         return "master"
 
 
-class Porter(object):
-    def __init__(self, **kwargs):
-        self.__dict__.update(kwargs)
+class Porter:
```
clokep, line 246: Ah looks like these never got some of the Python 3 upgrades, we should probably run `pyupgrade` on them...

clokep (approved):
Seems OK to me.

PR #12423, opened 2022-04-08
Add some type hints to datastore

dklimpel:
### Pull Request Checklist

### DIFF
--- a/synapse/storage/databases/main/appservice.py
+++ b/synapse/storage/databases/main/appservice.py
@@ -14,7 +14,7 @@
 # limitations under the License.
 import logging
 import re
-from typing import TYPE_CHECKING, List, Optional, Pattern, Tuple
+from typing import TYPE_CHECKING, Any, Collection, Dict, List, Optional, Pattern, Tuple
 
 from synapse.appservice import (
     ApplicationService,
@@ -26,7 +26,11 @@ from synapse.appservice import (
 from synapse.config.appservice import load_appservices
 from synapse.events import EventBase
 from synapse.storage._base import db_to_json
-from synapse.storage.database import DatabasePool, LoggingDatabaseConnection
+from synapse.storage.database import (
+    DatabasePool,
+    LoggingDatabaseConnection,
+    LoggingTransaction,
+)
 from synapse.storage.databases.main.events_worker import EventsWorkerStore
 from synapse.storage.databases.main.roommember import RoomMemberWorkerStore
 from synapse.storage.types import Cursor
@@ -92,7 +96,7 @@ class ApplicationServiceWorkerStore(RoomMemberWorkerStore):
 
         super().__init__(database, db_conn, hs)
 
-    def get_app_services(self):
+    def get_app_services(self) -> List[ApplicationService]:
         return self.services_cache
 
     def get_if_app_services_interested_in_user(self, user_id: str) -> bool:
@@ -256,7 +260,7 @@ class ApplicationServiceTransactionWorkerStore(
             A new transaction.
         """
 
-        def _create_appservice_txn(txn):
+        def _create_appservice_txn(txn: LoggingTransaction) -> AppServiceTransaction:
             new_txn_id = self._as_txn_seq_gen.get_next_id_txn(txn)
 
             # Insert new txn into txn table
@@ -291,7 +295,7 @@ class ApplicationServiceTransactionWorkerStore(
             service: The application service which was sent this transaction.
         """
 
-        def _complete_appservice_txn(txn):
+        def _complete_appservice_txn(txn: LoggingTransaction) -> None:
             # Set current txn_id for AS to 'txn_id'
             self.db_pool.simple_upsert_txn(
                 txn,
@@ -322,7 +326,9 @@ class ApplicationServiceTransactionWorkerStore(
             An AppServiceTransaction or None.
         """
 
-        def _get_oldest_unsent_txn(txn):
+        def _get_oldest_unsent_txn(
+            txn: LoggingTransaction,
+        ) -> Optional[Dict[str, Any]]:
             # Monotonically increasing txn ids, so just select the smallest
             # one in the txns table (we delete them when they are sent)
             txn.execute(
@@ -364,7 +370,7 @@ class ApplicationServiceTransactionWorkerStore(
         )
 
     async def set_appservice_last_pos(self, pos: int) -> None:
-        def set_appservice_last_pos_txn(txn):
+        def set_appservice_last_pos_txn(txn: LoggingTransaction) -> None:
             txn.execute(
                 "UPDATE appservice_stream_position SET stream_ordering = ?", (pos,)
             )
@@ -378,7 +384,9 @@ class ApplicationServiceTransactionWorkerStore(
     ) -> Tuple[int, List[EventBase]]:
         """Get all new events for an appservice"""
 
-        def get_new_events_for_appservice_txn(txn):
+        def get_new_events_for_appservice_txn(
+            txn: LoggingTransaction,
+        ) -> Tuple[int, Collection[str]]:
             sql = (
                 "SELECT e.stream_ordering, e.event_id"
                 " FROM events AS e"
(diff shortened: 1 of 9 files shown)


synapse/storage/databases/main/appservice.py
```diff
@@ -378,7 +384,9 @@ async def get_new_events_for_appservice(
     ) -> Tuple[int, List[EventBase]]:
         """Get all new events for an appservice"""
 
-        def get_new_events_for_appservice_txn(txn):
+        def get_new_events_for_appservice_txn(
+            txn: LoggingTransaction,
+        ) -> Tuple[int, Collection[str]]:
```
DMRobertson, line 389: Any particular reason for returning a `Collection` rather than a `List`?

synapse/storage/databases/main/registration.py
```diff
@@ -121,7 +121,7 @@ def __init__(
         database: DatabasePool,
         db_conn: LoggingDatabaseConnection,
         hs: "HomeServer",
-    ):
+    ) -> None:
```
DMRobertson, line 124: For interest: provided at least one other argument is typed, Mypy will allow you to omit the return type from `__init__`.

(I don't have strong feelings either way on this.)

synapse/storage/databases/main/registration.py
```diff
@@ -394,7 +394,9 @@ async def get_users_expiring_soon(self) -> List[Dict[str, Any]]:
             A list of dictionaries, each with a user ID and expiration time (in milliseconds).
         """
 
-        def select_users_txn(txn, now_ms, renew_at):
+        def select_users_txn(
+            txn: LoggingTransaction, now_ms: int, renew_at: int
+        ) -> List[Dict[str, Any]]:
```
DMRobertson, line 399: I'm always reluctant to use `Any`: it means the typechecker can't do anything with this data.

There's only one call site for this function. We could consider refactoring it to return `List[Tuple[str, int]]`.

(To be explicit: there's nothing wrong with this change; I just like empowering mypy as much as possible.)

synapse/storage/databases/main/stream.py
```diff
@@ -742,7 +753,7 @@ def _f(txn):
                 " LIMIT 1"
             )
             txn.execute(sql, (room_id, stream_ordering))
-            return txn.fetchone()
+            return cast(Tuple[int, int, str], txn.fetchone())
```
DMRobertson, line 756: Is this correct? Couldn't the query select 0 rows, so that `txn.fetchone() is None`?

synapse/storage/databases/main/stream.py
```diff
@@ -770,7 +781,7 @@ def get_stream_id_for_event_txn(
         self,
         txn: LoggingTransaction,
         event_id: str,
-        allow_none=False,
+        allow_none: Literal[False] = False,
```
DMRobertson, line 784: This doesn't look right; get_stream_id_for_event_txn passes `allow_none=True`.  Could you explain your thinking?

synapse/storage/databases/main/stream.py
```diff
@@ -985,7 +996,9 @@ async def get_all_new_events_stream(
             the `current_id`).
         """
 
-        def get_all_new_events_stream_txn(txn):
+        def get_all_new_events_stream_txn(
+            txn: LoggingTransaction,
+        ) -> Tuple[int, Collection[str]]:
```
DMRobertson, line 1001: Similarly to earlier: why `Collection` here rather than `List`?

DMRobertson (changes requested):
Many thanks Dirk, this looks great. I have a few questions:

synapse/storage/databases/main/appservice.py
dklimpel: No. I have changed it to `List`.

synapse/storage/databases/main/stream.py
dklimpel: No reason. I have changed it to `List`.

synapse/storage/databases/main/stream.py
dklimpel: Oh. This was a mistake.

synapse/storage/databases/main/stream.py
dklimpel: If is set `allow_none: bool = False,` I am get:
```
synapse/storage/databases/main/stream.py:786: error: No overload variant of "simple_select_one_onecol_txn" of "DatabasePool" matches argument types "LoggingTransaction", "str", "Dict[str, str]", "str", "bool"  [call-overload]
synapse/storage/databases/main/stream.py:786: note: Possible overload variants:
synapse/storage/databases/main/stream.py:786: note:     def simple_select_one_onecol_txn(cls, txn: LoggingTransaction, table: str, keyvalues: Dict[str, Any], retcol: str, allow_none: Literal[False] = ...) -> Any
synapse/storage/databases/main/stream.py:786: note:     def simple_select_one_onecol_txn(cls, txn: LoggingTransaction, table: str, keyvalues: Dict[str, Any], retcol: str, allow_none: Literal[True] = ...) -> Optional[Any]
```

I cannot see the mistake:
This should match with:
https://github.com/matrix-org/synapse/blob/98ec375b26ff167f91d40cf5e531d5640fca3635/synapse/storage/database.py#L1564-L1571

synapse/storage/databases/main/registration.py
dklimpel: I have refactore it to use `List[Tuple[str, int]]`

synapse/storage/databases/main/appservice.py
DMRobertson: Cheers! FWIW, my rule of thumb here is that `Collection` and the [other abstract types](https://docs.python.org/3/library/typing.html?highlight=typing#abstract-base-classes) are good for annotating function parameters. but I think return types are often best typed more literally.

(That is just an Opinion though...)

synapse/storage/databases/main/stream.py
DMRobertson: My understanding:

1. `StreamWorkerStore.get_stream_id_for_event_txn` is declared to return an `int`.
2. `StreamWorkerStore.get_stream_id` returns whatever `DatabasePool.simple_select_one_onecol_txn` returns. The former passes a value of `allow_none` through to the latter.
3. `DatabasePool.simple_select_one_onecol_txn` returns `Optional[Any]`. If it's called with `allow_none=False`, then it returns `Any`; the `None` case is handled by raising `StoreError`. (See https://github.com/matrix-org/synapse/blob/e630722f11821b4d7610bceae256674114f9803b/synapse/storage/database.py#L1539-L1561; this is where the overloads mentioned in the error message come from).

Mypy has spotted that if we call `get_stream_id_for_event_txn` with `allow_none=True`, then `simple_select_one_onecol_txn` might return `None`. And that would contradict the return type annotation of `get_stream_id_for_event_txn`.

We only seem to call `get_stream_id_for_event_txn` in two places, both in this function:

https://github.com/matrix-org/synapse/blob/3e0536cd2afb5a640619bd872fc27b068ec3eb9b/synapse/storage/databases/main/event_push_actions.py#L184-L216

In particular line 194 of the quoted source looks like it's expecting `stream_ordering` to be possibly None.

DMRobertson: Given the above, I think the right thing to do is

1. change `get_stream_id_for_event_txn` to return `Optional[int]` instead of `int`.
2. Try changing `EventPushActionsWorkerStore` to inherit from `StreamWorkerStore` and removing the type-ignores where we call `get_stream_id_for_event_txn`. Doing so might fail due with DataStore MRO problems (related: #11165 and more generally #11733). If so, that's fine and we can leave things as they are.

Sorry Dirk, that's maybe a bit of information overload. Does that make sense? (If not, is there anything I can clarify?)

dklimpel: Does it make sense to open an extra issue for this? And do not touch this type hint here?

DMRobertson: See #12437.

DMRobertson (changes requested):
As per [your comment](https://github.com/matrix-org/synapse/pull/12423#discussion_r847376077), let's leave the annotation of `allow_none` alone for now; I've raised #12437. Once that's done I think this looks good to merge!

synapse/storage/databases/main/stream.py
dklimpel: Thank you for opening the new issue.

DMRobertson:
Thanks again, Dirk! The complement test failure should be rectified by #12439.

PR #12425, opened 2022-04-08
Use poetry lockfile in twisted trunk CI job

DMRobertson:
Pulled out from #12337.

Co-authored-by: Sean Quah <seanq@element.io>
Co-authored-by: Dan Callahan <danc@element.io>

### DIFF
--- a/.github/workflows/twisted_trunk.yml
+++ b/.github/workflows/twisted_trunk.yml
@@ -21,7 +21,6 @@ jobs:
           python-version: "3.x"
           extras: "all"
       - run: |
-          poetry remove twisted
           poetry add --extras tls git+https://github.com/twisted/twisted.git#trunk
           poetry install --no-interaction --extras "all test"
       - run: poetry run mypy
@@ -37,7 +36,6 @@ jobs:
           python-version: "3.x"
           extras: "all test"
       - run: |
-          poetry remove twisted
           poetry add --extras tls git+https://github.com/twisted/twisted.git#trunk
           poetry install --no-interaction --extras "all test"
       - run: poetry run trial --jobs 2 tests
@@ -65,16 +63,14 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Patch dependencies
-        # Note: The script below runs poetry against a virtualenv in /src/.venv/
-        #       ...but the sytest-synapse container expects it to be in /venv/
-        #       So we temporarily symlink it, run poetry, then remove the link.
-        #       Tidying up is important because `/src` is a mounted volume.
+        # Note: The poetry commands want to create a virtualenv in /src/.venv/,
+        #       but the sytest-synapse container expects it to be in /venv/.
+        #       We symlink it before running poetry so that poetry actually
+        #       ends up installing to `/venv`.
         run: |
           ln -s -T /venv /src/.venv
-          poetry remove twisted
           poetry add --extras tls git+https://github.com/twisted/twisted.git#trunk
           poetry install --no-interaction --extras "all test"
-          rm /src/.venv
         working-directory: /src
       - name: Run SyTest
         run: /bootstrap.sh synapse
--- a/.github/workflows/twisted_trunk.yml
+++ b/.github/workflows/twisted_trunk.yml
@@ -6,16 +6,25 @@ on:
 
   workflow_dispatch:
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   mypy:
     runs-on: ubuntu-latest
 
     steps:
       - uses: actions/checkout@v2
-      - uses: actions/setup-python@v2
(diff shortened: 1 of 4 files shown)


.github/workflows/twisted_trunk.yml
```diff
@@ -55,11 +65,24 @@ jobs:
...
+          poetry install --no-interaction --extras "all test"
+          rm /src/.venv
         working-directory: /src
       - name: Run SyTest
         run: /bootstrap.sh synapse
         working-directory: /src
+        env:
+          # Use offline mode to avoid reinstalling the pinned version of
+          # twisted.
+          OFFLINE: 1
```
DMRobertson, line 85: I think he twisted trunk CI jobs from today were using the locked environment without this patch?

DMRobertson:
I'm going to trigger the twisted trunk CI job manually against this branch. Hold on to your hats: https://github.com/matrix-org/synapse/actions/runs/2116790909

DMRobertson:
@squahtx and @callahad have the context on this one; I've just mainly been commit wrangling.

DMRobertson:
Kicking off another run: https://github.com/matrix-org/synapse/actions/runs/2116943435

DMRobertson:
> Kicking off another run: [matrix-org/synapse/actions/runs/2116943435](https://github.com/matrix-org/synapse/actions/runs/2116943435)

This was successful!

.github/workflows/twisted_trunk.yml
```diff
@@ -55,11 +65,24 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Patch dependencies
-        run: .ci/patch_for_twisted_trunk.sh
+        # Note: The script below runs poetry against a virtualenv in /src/.venv/
+        #       ...but the sytest-synapse container expects it to be in /venv/
+        #       So we temporarily symlink it, run poetry, then remove the link.
```
richvdh, line 70: slightly amazed that works, but ok!

.github/workflows/twisted_trunk.yml
```diff
@@ -55,11 +65,24 @@ jobs:
...
+        #       Tidying up is important because `/src` is a mounted volume.
```
richvdh, line 71: I'm not really following why that matters? What cares about things being in `/src` ?

richvdh (commented):
lgtm, but I have a question about the `rm`

.github/workflows/twisted_trunk.yml
DMRobertson: I must defer to a higher power here: @squahtx

squahtx: To be fair it probably doesn't matter, since `/src` is a volume mount of `${{ github.workspace }}` and I believe nothing bad happens if you scribble all over it.

(Later on, `scripts/synapse_sytest.sh` in sytest will make a copy of `/src`, `rm -rf .venv` and re-do the symlink. Nothing will break there if a stray virtualenv is left in.)

richvdh:
fixes: #12458

.github/workflows/twisted_trunk.yml
```diff
@@ -55,11 +65,24 @@ jobs:
...
+        run: |
+          ln -s -T /venv /src/.venv
+          poetry remove twisted
+          poetry add --extras tls git+https://github.com/twisted/twisted.git#trunk
+          poetry install --no-interaction --extras "all test"
+          rm /src/.venv
```
richvdh, line 77:
```suggestion
        # Note: The poetry commands want to create a virtualenv in /src/.venv/,
        #       but the sytest-synapse container expects it to be in /venv/,
        #       so we symlink it before running poetry so that poetry actually
        #       ends up installing to `/venv`.
        run: |
          ln -s -T /venv /src/.venv
          poetry remove twisted
          poetry add --extras tls git+https://github.com/twisted/twisted.git#trunk
          poetry install --no-interaction --extras "all test"
```

.github/workflows/twisted_trunk.yml
richvdh: ok, then I suggest we remove the `rm` and update the comment. Have made a suggestion below.

callahad:
Suspect we can remove the `poetry remove twisted` lines now that #12444 has merged... trying.

callahad:
Manual GHA run at https://github.com/matrix-org/synapse/actions/runs/2160337360

callahad:
Nope, we can't. Reverting.

callahad:
*Just* running `poetry add --extras tls git+https://github.com/twisted/twisted.git#trunk` (without first calling `remove`) will work on Poetry 1.2, but not in 1.1.

callahad:
Manual CI run again trunk works: https://github.com/matrix-org/synapse/actions/runs/2160417080

callahad:
Merging on the basis of @richvdh's prior "lgtm" and question about the `rm` invocation which was removed following discussion. The only other changes since then have been to comments.

PR #12434, opened 2022-04-11
Run lints under poetry in CI; remove lint tox jobs

DMRobertson:
Pulled out from #12337.

- We don't recommend tox in `code_style.md`
- After #12107 it's much easier for black, isort and flake8 to find the
  scripts we want them to lint; we don't need to maintain `lint_targets`
  in the tox config. See also #12420.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -20,19 +20,13 @@ jobs:
       - run: scripts-dev/config-lint.sh
 
   lint:
-    runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        toxenv:
-          - "check_codestyle"
-          - "check_isort"
-          - "mypy"
-
-    steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-python@v2
-      - run: pip install tox
-      - run: tox -e ${{ matrix.toxenv }}
+    # This does a vanilla `poetry install` - no extras. I'm slightly anxious
+    # that we might skip some typechecks on code that uses extras. However,
+    # I think the right way to fix this is to mark any extras needed for
+    # typechecking as development dependencies. To detect this, we ought to
+    # turn up mypy's strictness: disallow unknown imports and be accept fewer
+    # uses of `Any`.
+    uses: "matrix-org/backend-meta/.github/workflows/python-poetry-ci.yml@v1"
 
   lint-crlf:
     runs-on: ubuntu-latest
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,5 @@
 [tox]
-envlist = py37, py38, py39, py310, check_codestyle, check_isort
+envlist = py37, py38, py39, py310
 
 # we require tox>=2.3.2 for the fix to https://github.com/tox-dev/tox/issues/208
 minversion = 2.3.2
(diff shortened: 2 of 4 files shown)


DMRobertson:
I think the sytest failure is a flake: https://github.com/matrix-org/sytest/issues/1136

.github/workflows/tests.yml
```diff
@@ -20,19 +20,13 @@ jobs:
...
-      - uses: actions/setup-python@v2
-      - run: pip install tox
-      - run: tox -e ${{ matrix.toxenv }}
+    # This does a vanilla `poetry install` - no extras. I'm slightly anxious
+    # that we might skip some typechecks on code that uses extras. However,
+    # I think the right way to fix this is to mark any extras needed for
+    # typechecking as development dependencies. To detect this, we ought to
+    # turn up mypy's strictness: disallow unknown imports and be accept fewer
+    # uses of `Any`.
+    uses: "matrix-org/backend-meta/.github/workflows/python-poetry-ci.yml@v1"
```
DMRobertson, line 29: Hyperlinked for your pleasure: https://github.com/matrix-org/backend-meta/blob/v1/.github/workflows/python-poetry-ci.yml

.github/workflows/tests.yml
```diff
@@ -20,19 +20,13 @@ jobs:
...
-    steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-python@v2
-      - run: pip install tox
-      - run: tox -e ${{ matrix.toxenv }}
+    # This does a vanilla `poetry install` - no extras. I'm slightly anxious
+    # that we might skip some typechecks on code that uses extras. However,
+    # I think the right way to fix this is to mark any extras needed for
+    # typechecking as development dependencies. To detect this, we ought to
+    # turn up mypy's strictness: disallow unknown imports and be accept fewer
```
H-Shay, line 27: Just curious if this has been done or is waiting on something?

tox.ini
```diff
@@ -138,18 +124,3 @@ setenv =
 commands =
     python -m synmark {posargs:}
 
-[testenv:check_codestyle]
-extras = lint
-commands =
-    python -m black --check --diff {[base]lint_targets}
```
H-Shay, line 144: Just curious why there is no longer a `--diff` flag in the call to black in the new setup: https://github.com/matrix-org/backend-meta/blob/2001ab8d60a922b0d1f48bd5f3b38765e9f74113/.github/workflows/python-poetry-ci.yml#L22

DMRobertson: No particular reason: that's an oversight. (I'm not sure it's hugely useful to have the --diff, but I'll fix that up)

DMRobertson: https://github.com/matrix-org/backend-meta/blob/9588c2bbc156aa62ac09d6663335b282da625a30/.github/workflows/python-poetry-ci.yml#L22 now includes --diff. The v1 tag has been moved forward to point to this. (Which is normally bad practice, but it's what's [recommended for actions](https://docs.github.com/en/enterprise-server@3.4/actions/creating-actions/about-custom-actions#using-release-management-for-actions).)

.github/workflows/tests.yml
DMRobertson: Not done, not waiting on anything in particular.

(We could easily pass extras on to the workflow---but I'm reluctant to, because I want the dev dependencies automatically pulled in by `poetry install` to be _exactly_ what you need to do all development tasks, including typechecking.)

H-Shay (approved):
This LGTM, I think there was some question as to whether to add the lint targets back in or not, but I don't think that should preclude approval?

PR #12435, opened 2022-04-11
Enable certificate checking during complement tests

### LINKED ISSUE #11803: complement test image disables HTTP cert checking
The image we build to run under complement disables HTTP cert checking for federation connections:

https://github.com/matrix-org/complement/blob/master/dockerfiles/synapse/homeserver.yaml#L44-L46

@kegsay suggests we don't do anything about this until https://github.com/matrix-org/complement/issues/65 is fixed

richvdh:
Fixes #11803.


clokep (approved):
Seems good if CI is happy!

Issue #15, opened 2014-11-18
Homeserver installation issue


winsontan520:
I face some issue here. I got this when I execute "python setup.py develop --user"
below my console output:

```
~/Git/synapse$ python setup.py develop --user
running develop
running egg_info
writing requirements to SynapseHomeServer.egg-info/requires.txt
writing SynapseHomeServer.egg-info/PKG-INFO
writing top-level names to SynapseHomeServer.egg-info/top_level.txt
writing dependency_links to SynapseHomeServer.egg-info/dependency_links.txt
writing entry points to SynapseHomeServer.egg-info/entry_points.txt
reading manifest file 'SynapseHomeServer.egg-info/SOURCES.txt'
reading manifest template 'MANIFEST.in'
warning: no files found matching '*.sql' under directory 'synapse/persistence/schema'
writing manifest file 'SynapseHomeServer.egg-info/SOURCES.txt'
running build_ext
Creating /Users/admin/Library/Python/2.7/lib/python/site-packages/SynapseHomeServer.egg-link (link to .)
SynapseHomeServer 0.0.1 is already the active version in easy-install.pth
Installing synapse-homeserver script to /Users/admin/Library/Python/2.7/bin

Installed /Users/admin/Git/synapse
Processing dependencies for SynapseHomeServer==0.0.1
Searching for cryptography>=0.2.1
Reading https://pypi.python.org/simple/cryptography/
Best match: cryptography 0.6.1
Downloading https://pypi.python.org/packages/source/c/cryptography/cryptography-0.6.1.tar.gz#md5=1c48fd78742d85c9cffdcceb6809e45b
Processing cryptography-0.6.1.tar.gz
Writing /var/folders/0t/xt8fspyd76995w8lq__67dlr0000gn/T/easy_install-Mg_TNT/cryptography-0.6.1/setup.cfg
Running cryptography-0.6.1/setup.py -q bdist_egg --dist-dir /var/folders/0t/xt8fspyd76995w8lq__67dlr0000gn/T/easy_install-Mg_TNT/cryptography-0.6.1/egg-dist-tmp-M8QmES
Traceback (most recent call last):
  File "setup.py", line 58, in <module>
    """
  File "/opt/local/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/distutils/core.py", line 152, in setup
    dist.run_commands()
[...]

PR #12439, opened 2022-04-11
Workaround pip bug installing latest treq and not-latest twisted from hashes

DMRobertson:
The requirements file generated by `poetry export` isn't correctly processed by `pip install -r requirements.txt`. It contains twisted and treq, both pinned to 22.2.0.

When `pip` installs treq, it notices that `Twisted[tls]` is required. It then tries to acquire the latest twisted release, only to fail (because this hash isn't listed in the requirements file).From e.g. https://github.com/matrix-org/synapse/runs/5977154990?check_suite_focus=true

> ```
> #15 9.204 Collecting Twisted[tls]>=18.7.0
> #15 9.205 ERROR: In --require-hashes mode, all requirements must have their versions pinned with ==. These do not:
> #15 9.205     Twisted[tls]>=18.7.0 from https://files.pythonhosted.org/packages/db/99/38622ff95bb740bcc991f548eb46295bba62fcb6e907db1987c4d92edd09/Twisted-22.4.0-py3-none-any.whl#sha256=f9f7a91f94932477a9fc3b169d57f54f96c6e74a23d78d9ce54039a7f48928a2 (from treq==22.2.0->-r /synapse/requirements.txt (line 724))
> #15 ERROR: executor failed running [/bin/sh -c pip install --prefix="/install" --no-warn-script-location -r /synapse/requirements.txt]: exit code: 1
> ```

The underlying pip issue is https://github.com/pypa/pip/issues/9644. A comment notes that one can avoid this behaviour with by `pip install`ing with the `--no-deps` flag. Let us do so.

(At first glance, the problem looks like https://github.com/python-poetry/poetry/issues/5311, but that was a bug in `poetry install`; this is `poetry export`, whose behaviour is fine AFAICS).

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -63,6 +63,11 @@ COPY pyproject.toml poetry.lock README.rst /synapse/
 
 RUN /root/.local/bin/poetry export --extras all -o /synapse/requirements.txt
 
+# Workaround https://github.com/pypa/pip/issues/9644. Without this, pip tries to install
+# Twisted[tls] using the latest twisted and fails; it doesn't realisie that there is
+# already a version of twisted given in the requirments.
+RUN sed -i /synapse/requirements.txt -e 's/^twisted==/twisted[tls]==/'
+
 ###
 ### Stage 1: builder
 ###
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -63,11 +63,6 @@ COPY pyproject.toml poetry.lock README.rst /synapse/
 
 RUN /root/.local/bin/poetry export --extras all -o /synapse/requirements.txt
 
-# Workaround https://github.com/pypa/pip/issues/9644. Without this, pip tries to install
-# Twisted[tls] using the latest twisted and fails; it doesn't realisie that there is
-# already a version of twisted given in the requirments.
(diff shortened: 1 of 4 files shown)


callahad:
If this works, then we can also remove the `poetry remove twisted` lines from the Twisted trunk CI script

DMRobertson:
This didn't work: the change didn't alter the exported `requirements.txt`. (I was confused by local docker layer caching).

I think it's time to inspect the exported requirements.

docker/Dockerfile
```diff
@@ -63,6 +63,11 @@ COPY pyproject.toml poetry.lock README.rst /synapse/
 
 RUN /root/.local/bin/poetry export --extras all -o /synapse/requirements.txt
 
+# Workaround https://github.com/pypa/pip/issues/9644. Without this, pip tries to install
+# Twisted[tls] using the latest twisted and fails; it doesn't realisie that there is
```
clokep, line 67:
```suggestion
# Twisted[tls] using the latest twisted and fails; it doesn't realise that there is
```

DMRobertson:
The commit history and comments etc are a little confusing in retrospect. Fortunately, the diff consists of just two lines.

erikjohnston (approved):

PR #12441, opened 2022-04-11
Bump twisted version to the latest in lockfile

H-Shay:
May be necessary in lieu of https://github.com/matrix-org/synapse/pull/12439?

### DIFF
--- a/changelog.d/12441.misc
+++ b/changelog.d/12441.misc
@@ -1,1 +1,1 @@
-Bump twisted version in poetry.lock to work around pip bug #9644.
+Bump twisted version in `poetry.lock` to work around [pip bug #9644](https://github.com/pypa/pip/issues/9644).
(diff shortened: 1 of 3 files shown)


callahad:
Twisted 22.2 to 22.4 changelog looks pretty minimal / low risk: https://github.com/twisted/twisted/pull/1714/files

changelog.d/12441.misc
```diff
@@ -0,0 +1 @@
+Bump twisted version in poetry.lock to work around pip bug #9644.
```
squahtx, line 1:
```suggestion
Bump twisted version in `poetry.lock` to work around [pip bug #9644](https://github.com/pypa/pip/issues/9644).
```

squahtx (approved):
No objections to bumping twisted here, regardless of whether #12439 lands.

PR #12448, opened 2022-04-12
Update UPGRADE notes to include more details about v1.57 upgrade failure mode

### LINKED ISSUE #12447: Update UPGRADE notes to include more details about v1.57 upgrade failure mode
If you fail to stop the AS worker when upgrading Synapse then the AS worker may refuse to start, spitting out an error message that says to run some SQL.

Is there is a concern that this is still confusing? Do we need to add more words somewhere about what to do?

richvdh:
Fixes #12447.


squahtx (approved):
LGTM
you've worded it far better than I would have managed

PR #12449, opened 2022-04-12
Use `poetry` to manage the virtualenv in debian packages

DMRobertson:
Pulled out from #12337.

I have very little context on this one: that's with @H-Shay and @callahad.

Co-authored-by: Dan Callahan <danc@element.io>
Co-authored-by: Shay <hillerys@element.io>


debian/build_virtualenv
```diff
@@ -30,8 +30,22 @@ case $(dpkg-architecture -q DEB_HOST_ARCH) in
         ;;
 esac
 
-# Use --builtin-venv to use the better `venv` module from CPython 3.4+ rather
-# than the 2/3 compatible `virtualenv`.
+# Manually install Poetry and export a pip-compatible `requirements.txt`
+# We need a Poetry pre-release as the export command is buggy in < 1.2
+if [ -e requirements.txt ]; then
```
richvdh, line 35: the problem with this is that it means we can't run `dpkg-buildpackage` twice (the second one fails). We kinda need to decide if `requirements.txt` is a thing that we generate (and should hence delete in [`dh_clean`](https://manpages.debian.org/buster/debhelper/dh_clean.1.en.html)) or is a thing that we expect to exist at the outset.

Possibly just giving it a different name would help.

DMRobertson: We should generate it.

I can add an entry to `debian/clean`, but running locally I'm not convinced that
- the generated file gets removed from `/synapse/build`
- the clean entry is necessary fora rebuild to suceed

debian/changelog
```diff
@@ -1,3 +1,10 @@
+matrix-synapse-py3 (1.58.0+nmu1) UNRELEASED; urgency=medium
```
DMRobertson, line 1: Please check this version number is appropriate.

richvdh: don't sweat the specific version number here; the release process fixes this up to match whatever we are releasing.

callahad:
We could also just whack an `rm` in there after the `dh-virtualenv` invocation, instead of dealing with `dh_clean`

callahad:
My reasoning behind the `if [ -e requirements.txt ]` check is that, at some point in the future, we may want to keep a `requirements.txt` in the main repo. If we do that, we should probably *not* silently clobber or ignore it.

Switching to `exported_requirements.txt` in this script implicitly says "we're fine ignoring it, should it ever exist." Are we sure that's the behavior we want? (Fully deferring to @richvdh on this question)

debian/build_virtualenv
richvdh: > We should generate it.
> 
> I can add an entry to `debian/clean`, but running locally I'm not convinced that
> 
>     * the generated file gets removed from `/synapse/build`

It seems to work for me.

Working on dad410484c2f746a6ff056f630c7085222f84914:

```
$ DEB_BUILD_OPTIONS=nocheck dpkg-buildpackage -b -uc
...
$ ls exported_requirements.txt
exported_requirements.txt
$ dpkg-buildpackage -T clean
dpkg-buildpackage: info: source package matrix-synapse-py3
dpkg-buildpackage: info: source version 1.58.0+nmu1
dpkg-buildpackage: info: source distribution UNRELEASED
dpkg-buildpackage: info: source changed by Synapse Packaging Team <packages@matrix.org>
 fakeroot debian/rules clean
dh clean --with python-virtualenv
   dh_clean
$ ls exported_requirements.txt
ls: cannot access 'exported_requirements.txt': No such file or directory
```

>     * the clean entry is necessary for a rebuild to suceed


Here's what I see, working on d3bf3177e6 (the last version with the check in place):
```
$ DEB_BUILD_OPTIONS=nocheck dpkg-buildpackage -b -uc
...
$ ls requirements.txt 
requirements.txt
$ DEB_BUILD_OPTIONS=nocheck dpkg-buildpackage -b -uc
...
make[1]: Entering directory '/home/rav/work/synapse/poetry'
./debian/build_virtualenv
requirements.txt already exists; aborting
make[1]: *** [debian/rules:52: override_dh_virtualenv] Error 1
make[1]: Leaving directory '/home/rav/work/synapse/poetry'
make: *** [debian/rules:61: binary] Error 2
dpkg-buildpackage: error: fakeroot debian/rules binary subprocess returned exit status 2
```

DMRobertson: That's encouraging. (I was simply running `build_debian.sh` inside a debian container.)

DMRobertson:
`clean` seems to be working. I have no opinions on the test for `requirements.txt` (other than I'd prefer it not to be commited to the source tree at all). Is there anything else that needs to be done here?

debian/build_virtualenv
richvdh: @callahad wrote:

> My reasoning behind the `if [ -e requirements.txt ]` check is that, at some point in the future, we may want to keep a `requirements.txt` in the main repo. If we do that, we should probably _not_ silently clobber or ignore it.

I must admit I'm struggling to forsee a world in which we would want to commit `requirements.txt`. 

But I feel like if we do make that change, we'll manage to remember to excise the `poetry export` commands at the same time.

richvdh (approved):
lgtm.

PR #12450, opened 2022-04-12
Fix typos in release script docs

erikjohnston:


scripts-dev/release.py
```diff
@@ -69,11 +69,12 @@ def cli():
         # ... wait for assets to build ...
 
         ./scripts-dev/release.py publish
+
```
erikjohnston, line 72: This newline is needed, annoyingly, as otherwise it renders the two commands on a single line

scripts-dev/release.py
```diff
@@ -69,11 +69,12 @@ def cli():
...
         ./scripts-dev/release.py upload
 
         # Optional: generate some nice links for the announcement
 
-        ./scripts-dev/release.py upload
+        ./scripts-dev/release.py announce
```
DMRobertson, line 77: oops. thanks.

scripts-dev/release.py
erikjohnston: Oh fun, TIL.

PR #12451, opened 2022-04-12
Update documentation to reflect that `run_background_tasks_on` is no longer experimental.

reivilibre:
We've been running them on m.org for a long time; asking around, it seems like it's just been forgotten about :).


clokep (approved):
There's some similar language for the stream writers, I'm unsure if those can be removed or not though.

dklimpel:
> There's some similar language for the stream writers, I'm unsure if those can be removed or not though.

I had the same thought.

https://github.com/matrix-org/synapse/blob/703f142cf7656bd4f468068157a6bed3596be377/docs/workers.md?plain=1#L344-L346

clokep (approved):

PR #12454, opened 2022-04-12
Limit `device_id` size to 512B

H-Shay:
As the title states. This PR limits the `device_id` field to a reasonable 512B.


changelog.d/12454.misc
```diff
@@ -0,0 +1 @@
+Limit max size of device_id to less than 8KB.
```
DMRobertson, line 1: Should this say 512B?

synapse/rest/client/login.py
```diff
@@ -342,6 +343,16 @@ async def _complete_login(
...
+
+        # Check that device_id is not greater than a reasonable 512B
+        device_id_size = sys.getsizeof(device_id)
+        if device_id_size > 512:
+            raise LoginError(
+                400,
+                "Device_id cannot be greater than 512B.",
+                errcode=Codes.INVALID_PARAM,
+            )
+
```
DMRobertson, line 355: I have two minor nitpicks which you should might want to ignore for the purposes of this review, but which might be of interest.

Nitpicking: this is asserting that (C)Python's internal representation of the string is no larger than 512 bytes. Postgres likey has a different representation with a different size, but the two are almost certainly close enough so that there we'll still be within the 8KB limit.

Another nitpick: future versions of CPython might choose to use a different string representation where the same string is smaller or larger in-memory. On my system, Python 3.10 uses 49 bytes for the empty string, whereas Python 2 uses 37 bytes. 

```shell
$ python2 --version; python2 -c "import sys; print(sys.getsizeof(''))"
Python 2.7.18
37

$ python3 --version; python3 -c "import sys; print(sys.getsizeof(''))"
Python 3.10.4
49
```

If this changes under our feet then we might have device IDs that are rejected after a homeserver runner upgrades their Python version.

synapse/rest/client/login.py
```diff
@@ -342,6 +343,16 @@ async def _complete_login(
             user_id = canonical_uid
 
         device_id = login_submission.get("device_id")
+
+        # Check that device_id is not greater than a reasonable 512B
+        device_id_size = sys.getsizeof(device_id)
+        if device_id_size > 512:
+            raise LoginError(
+                400,
+                "Device_id cannot be greater than 512B.",
```
DMRobertson, line 352: I think the error message could be a point of confusion. On my system, a 464-character ASCII string (which would be encoded as 464 bytes in UTF-8) just breaches the proposed representation size limit:

```python
>>> sys.getsizeof("1"*464)
513
```

So the error message could be a point of confusion here. I'd suggest either:
- be vaguer: say something like "device_id is too long" without committing to specifics
- use a different metric: e.g. reject strings whose `len()`is bigger than 512. That should correspond to strings of at most 512 unicode codepoints (whose raw data would be encoded in UTF-8 in at most 2048 bytes).

Given my slightly paranoid nitpicks in the other comment, I think I'd opt for the latter; but I defer to your and the team's judgement!

tests/rest/client/test_login.py
```diff
@@ -384,6 +385,38 @@ def test_session_can_hard_logout_all_sessions_after_being_soft_logged_out(
         channel = self.make_request(b"POST", "/logout/all", access_token=access_token)
         self.assertEqual(channel.result["code"], b"200", channel.result)
 
+    def test_login_with_overly_large_device_id_fails(self) -> None:
+        self.register_user("mickey", "cheese")
+
+        # create a device_id larger than 512B
+        device_id = 1000**1000
```
DMRobertson, line 392: Note that `device_id` is an integer here rather than a string. I expect this will get encoded as a JSON number, which is contrary to the spec!

To make large strings, one can use the `*` operator: e.g.

```python
>>> "hello"*50
'hellohellohellohellohellohellohellohellohellohellohellohellohellohellohellohellohellohellohellohellohellohellohellohellohellohellohellohellohellohellohellohellohellohellohellohellohellohellohellohellohellohellohellohellohellohellohellohellohellohello'
```

synapse/rest/client/login.py
clokep: I think we usually limit by string length (number of characters)?

changelog.d/12454.misc
H-Shay: D'oh

synapse/rest/client/login.py
H-Shay: Very useful nitpicks! I have changed it to limit by size, I appreciate the explanation.

tests/rest/client/test_login.py
```diff
@@ -384,6 +384,37 @@ def test_session_can_hard_logout_all_sessions_after_being_soft_logged_out(
...
+            "type": "m.login.password",
+            "user": "mickey",
+            "password": "cheese",
+            "device_id": device_id,
+        }
+
+        # make a login request with the bad device_id
+        channel = self.make_request(
+            "POST",
+            "/_matrix/client/r0/login",
```
anoadragon453, line 403:
```suggestion
            "/_matrix/client/v3/login",
```

just to use the latest endpoint version.

synapse/rest/client/login.py
```diff
@@ -342,6 +342,15 @@ async def _complete_login(
             user_id = canonical_uid
 
         device_id = login_submission.get("device_id")
+
+        # If device_id is present, check that device_id is not longer than a reasonable 512 characters
+        if device_id and len(device_id) > 512:
+            raise LoginError(
+                400,
+                "Device_id cannot be longer than 512 characters.",
```
anoadragon453, line 350:
```suggestion
                "device_id cannot be longer than 512 characters.",
```

Our query parameters are case-sensitive, for better or worse :upside_down_face: 

Alternatively, if you want to keep the capital first letter:

```suggestion
                "Parameter 'device_id' cannot be longer than 512 characters.",
```

tests/rest/client/test_login.py
```diff
@@ -384,6 +384,37 @@ def test_session_can_hard_logout_all_sessions_after_being_soft_logged_out(
...
+            custom_headers=None,
+        )
+
+        # test that the login fails with the correct error code/message
+        self.assertEqual(channel.code, 400)
+        self.assertEqual(
+            channel.json_body,
+            {
+                "errcode": "M_INVALID_PARAM",
+                "error": "Device_id cannot be longer than 512 characters.",
```
anoadragon453, line 414: I tend to be against checking localised error text in tests, as:

1. If we update the text, we'll need to do so in multiple places
2. Clients should not be performing logic based on the text anyways, as it may change.

As such, I think we should only check the `errcode` is correct here.

anoadragon453 (approved):
Just some minor nits really, otherwise this lgtm!

This does have the potential to break someone's workflow, though I'd surprised to know why someone would *need* >512 character length device IDs...

tests/rest/client/test_login.py
H-Shay: Right, thanks for explaining that. Makes sense!

PR #12455, opened 2022-04-12
Reintroduce the lint targets in the linter script

DMRobertson:
I still think (+0.4) there is value in making CI lint everything by
default, with sensible exclusions configured as appropriate. But that
doesn't have to cause pain for local development that just happens to
have Python files lying around in their synapse source tree.


H-Shay (approved):
LGTM, thank you for the cleanup in the docs.

richvdh (approved):
lgtm

PR #12464, opened 2022-04-13
Dockerfile-workers: reduce the amount we install

richvdh:
This is an attempt to reduce the rebuild time. In short, we reduce the amount of stuff that the dockerfile installs.


docker/configure_workers_and_start.py
```diff
@@ -512,9 +512,10 @@ def generate_worker_files(environ, config_path: str, data_dir: str):
     )
 
     # Supervisord config
+    os.makedirs('/etc/supervisor', exist_ok=True)
     convert(
         "/conf/supervisord.conf.j2",
-        "/etc/supervisor/conf.d/supervisord.conf",
+        "/etc/supervisor/supervisord.conf",
```
richvdh, line 518: Previously, we would `apt-get install supervisor`, which would install a config file in `/etc/supervisor/supervisord.conf`, which would include `/etc/supervisor/conf.d/*.conf`. We now install the generated config file directly to `/etc/supervisor/supervisord.conf` instead.

docker/configure_workers_and_start.py
```diff
@@ -532,14 +533,6 @@ def generate_worker_files(environ, config_path: str, data_dir: str):
         os.mkdir(log_dir)
 
 
-def start_supervisord():
-    """Starts up supervisord which then starts and monitors all other necessary processes
-
-    Raises: CalledProcessError if calling start.py return a non-zero exit code.
-    """
-    subprocess.run(["/usr/bin/supervisord"], stdin=subprocess.PIPE)
```
richvdh, line 540: inlining this and replacing with `os.exec` is just a driveby while I was passing. Avoiding the unnecessary subprocess makes signal handling more robust.

docker/configure_workers_and_start.py
reivilibre: This will make it trickier to add a `/etc/supervisor/conf.d/postgres.conf` file from a `COPY` step in the Dockerfile; I'm not sure it's an improvement. (I could be convinced.)

I suppose the include could be added back in to our config when it becomes useful.

reivilibre (approved):

docker/configure_workers_and_start.py
richvdh: yeah, if we've got a usecase for the include, we can just add it back into `supervisord.conf.j2`.

richvdh: in fact, lemme do it now.

richvdh: done in [dadfa72](https://github.com/matrix-org/synapse/pull/12464/commits/dadfa72dd04767a3da33ba6de797523191d868df)

PR #12465, opened 2022-04-13
Only send out device list updates for our own users

erikjohnston:
Broke in #12365


clokep (approved):
I think so? Should we add a test to ensure this?

erikjohnston:
> I think so? Should we add a test to ensure this?

Woops, forgot that part. Have a test.

clokep (approved):

PR #12466, opened 2022-04-13
Dockerfile-workers: give the master its own log config

richvdh:
When we run a worker-mode synapse under docker, everything gets logged to stdout. Currently, output from the workers is tacked with a worker name, for example:

```
2022-04-13 15:27:56,810 - worker:frontend_proxy1 - synapse.util.caches.lrucache - 154 - INFO - LruCache._expire_old_entries-0 - Dropped 0 items from caches
```

- note `worker:frontend_proxy1`. No such tag is applied to log lines from the master, which makes somewhat confusing reading.

To fix this, we generate a dedicated log config file for the master in the same way that we do for the workers, and use that.

### DIFF
--- a/docker/configure_workers_and_start.py
+++ b/docker/configure_workers_and_start.py
@@ -483,7 +483,7 @@ def generate_worker_files(environ, config_path: str, data_dir: str):
     # Finally, we'll write out the config files.
 
     # log config for the master process
-    master_log_config = generate_worker_log_config(environ, "master")
+    master_log_config = generate_worker_log_config(environ, "master", data_dir)
     shared_config["log_config"] = master_log_config
 
     # Shared homeserver config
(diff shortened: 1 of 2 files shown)


docker/configure_workers_and_start.py
```diff
@@ -496,6 +482,10 @@ def generate_worker_files(environ, config_path: str, data_dir: str):
 
     # Finally, we'll write out the config files.
 
+    # log config for the master process
+    master_log_config = generate_worker_log_config(environ, "master")
```
clokep, line 486: Does this need the `data_dir` parameter?

richvdh: evidently. Not quite sure how I managed this.

clokep (approved):
Seems fine. 👍

PR #12467, opened 2022-04-13
complement-synapse-workers: factor out separate entry point script

richvdh:
... with a bit more verbosity.


docker/complement/SynapseWorkers.Dockerfile
```diff
@@ -13,8 +13,8 @@ RUN curl -OL "https://github.com/caddyserver/caddy/releases/download/v2.3.0/cadd
   tar xzf caddy_2.3.0_linux_amd64.tar.gz && rm caddy_2.3.0_linux_amd64.tar.gz && mv caddy /root
 
 # Install postgresql
-RUN apt-get update
-RUN apt-get install -y postgresql
+RUN apt-get update && \
+  DEBIAN_FRONTEND=noninteractive apt-get install --no-install-recommends -y postgresql-13
```
richvdh, line 17: bit of a driveby to avoid installing the recommends packages, and to install postgres 13 specifically since we depend on that in the next line.

PR #12468, opened 2022-04-13
Update `delay_cancellation` to accept any awaitable

squahtx:
This will mainly be useful when dealing with module callbacks, which are
all typed as returning `Awaitable`s instead of coroutines or
`Deferred`s.

### Pull Request Checklist

### DIFF
--- a/synapse/util/async_helpers.py
+++ b/synapse/util/async_helpers.py
@@ -14,6 +14,7 @@
 # limitations under the License.
 
 import abc
+import asyncio
 import collections
 import inspect
 import itertools
@@ -743,11 +744,14 @@ def delay_cancellation(awaitable: Awaitable[T]) -> Awaitable[T]:
     # First, convert the awaitable into a `Deferred`.
     if isinstance(awaitable, defer.Deferred):
         deferred = awaitable
-    elif isinstance(awaitable, Coroutine):
+    elif asyncio.iscoroutine(awaitable):
+        # Ideally we'd use `Deferred.fromCoroutine()` here, to save on redundant
+        # type-checking, but we'd need Twisted >= 21.2.
         deferred = defer.ensureDeferred(awaitable)
     else:
         # We have no idea what to do with this awaitable.
-        # Let the caller `await` it normally.
+        # We assume it's already resolved, such as `DoneAwaitable`s or `Future`s from
+        # `make_awaitable`, and let the caller `await` it normally.
         return awaitable
 
     def handle_cancel(new_deferred: "defer.Deferred[T]") -> None:
--- a/synapse/util/async_helpers.py
+++ b/synapse/util/async_helpers.py
@@ -25,6 +25,7 @@ from typing import (
     Awaitable,
     Callable,
     Collection,
+    Coroutine,
     Dict,
     Generic,
     Hashable,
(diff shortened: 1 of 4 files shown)


synapse/util/async_helpers.py
```diff
@@ -701,27 +702,54 @@ def stop_cancellation(deferred: "defer.Deferred[T]") -> "defer.Deferred[T]":
...
+        The new `Deferred` will follow the Synapse logcontext rules if `awaitable`
         follows the Synapse logcontext rules. Otherwise the new `Deferred` should be
         wrapped with `make_deferred_yieldable`.
     """
 
+    # First, convert the awaitable into a `Deferred`.
+    if isinstance(awaitable, defer.Deferred):
+        deferred = awaitable
+    elif isinstance(awaitable, Coroutine):
+        deferred = defer.ensureDeferred(awaitable)
```
richvdh, line 747: hrm, ideally we'd use `deferred.fromCoroutine(awaitable)` here, to save some redundant type checking, but that's only in Twisted 21.2. Might be worth adding a comment.

synapse/util/async_helpers.py
```diff
@@ -701,27 +702,54 @@ def stop_cancellation(deferred: "defer.Deferred[T]") -> "defer.Deferred[T]":
...
+
+        The new `Deferred` will follow the Synapse logcontext rules if `awaitable`
         follows the Synapse logcontext rules. Otherwise the new `Deferred` should be
         wrapped with `make_deferred_yieldable`.
     """
 
+    # First, convert the awaitable into a `Deferred`.
+    if isinstance(awaitable, defer.Deferred):
+        deferred = awaitable
+    elif isinstance(awaitable, Coroutine):
```
richvdh, line 746: I slightly wonder if this is reliable.

Possibly we want [`asyncio.iscoroutine`](https://docs.python.org/3/library/asyncio-task.html#asyncio.iscoroutine) which allows a wider range of types, and has some caching.

synapse/util/async_helpers.py
```diff
@@ -701,27 +702,54 @@ def stop_cancellation(deferred: "defer.Deferred[T]") -> "defer.Deferred[T]":
...
+        deferred = defer.ensureDeferred(awaitable)
+    else:
+        # We have no idea what to do with this awaitable.
+        # Let the caller `await` it normally.
+        return awaitable
```
richvdh, line 751: what sort of awaitables do we have that are neither coroutine objects nor Deferreds, outside of things returned by `make_awaitable` in unit tests? Would it be better to raise in this case? If so, perhaps we just call `defer.ensureDeferred(awaitable)` without bothering to do the type checks.

richvdh (commented):
looks good, but a few questions

synapse/util/async_helpers.py
squahtx: I'll add a comment.

synapse/util/async_helpers.py
squahtx: That's a good idea. `Deferred.fromCoroutine` accepts anything that `asyncio.iscoroutine` identifies as a coroutine, so it looks safe to do.

synapse/util/async_helpers.py
squahtx: Apart from `Futures` returned by `make_awaitable`, we have [`DoneAwaitables` from `maybe_awaitable`](https://github.com/matrix-org/synapse/blob/release-v1.56/synapse/util/async_helpers.py#L666). More generally, I think any awaitable that's already "complete" can be awaited in synapse.

I'm not keen on raising, since I do think we should support arbitrary completed awaitables, ie. if `await x` works, `await delay_cancellation(x)` should also work.

I'll add a comment mentioning the types of awaitable we'd expect here.

richvdh (approved):
lgtm!

Issue #12437, opened 2022-04-11
Incorrect type annotation in `get_stream_id_for_event_txn`


DMRobertson:
Originally [from this comment thread](https://github.com/matrix-org/synapse/pull/12423#discussion_r846349017).

`StreamWorkerStore.get_stream_id_for_event_txn` is declared to return an `int`. If you call it with `allow_none=True`, it returns `None`, which contravenes the type hints. I think this behaviour is intentional and the type hint is incorrect, judging by how it's used. We only call `get_stream_id_for_event_txn` in two places, both in this function:

https://github.com/matrix-org/synapse/blob/3e0536cd2afb5a640619bd872fc27b068ec3eb9b/synapse/storage/databases/main/event_push_actions.py#L184-L216

In particular line 194 of the quoted source looks like it's expecting `stream_ordering` to be possibly None.

Given the above, I think the right thing to do is

1. change `get_stream_id_for_event_txn` to return `Optional[int]` instead of `int`.
2. Try changing `EventPushActionsWorkerStore` to inherit from `StreamWorkerStore` and removing the type-ignores where we call `get_stream_id_for_event_txn`. Doing so might fail due with DataStore MRO problems (related: #11165 and more generally #11733). If so, that's fine and we can leave things as they are.

gautamgiri-dev:
Hello @DMRobertson I have made some changes as suggested by you followed by some changes based on linting errors thrown by mypy. Could you please review it.. and please assign this issue to me

clokep:
This was fixed in #12716.

PR #12470, opened 2022-04-13
fix type annotation in `get_stream_id_for_event_txn`

gautamgiri-dev:
Fixes: #12437 
`StreamWorkerStore.get_stream_id_for_event_txn` now has return type `Optional[int]` which is consistent with the `NoneType` when `allow_none=True`.

Signed-off-by: Gautam Kumar Giri <personal.gautamgiri@gmail.com>

### Pull Request Checklist

### DIFF
--- a/synapse/storage/database.py
+++ b/synapse/storage/database.py
@@ -1560,6 +1560,21 @@ class DatabasePool:
     ) -> Optional[Any]:
         ...
 
+    @overload
+    @classmethod
+    def simple_select_one_onecol_txn(
+        cls,
+        txn: LoggingTransaction,
+        table: str,
+        keyvalues: Dict[str, Any],
+        retcol: str,
+        allow_none: bool = False,
+    ) -> Optional[int]:
+        ret = cls.simple_select_one_onecol_txn(cls, txn, table=table, keyvalues=keyvalues, retcol=retcol, allow_none=allow_none)
+        if ret:
+            return int(ret)
+        return ret
+
     @classmethod
     def simple_select_one_onecol_txn(
         cls,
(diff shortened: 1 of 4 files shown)


synapse/storage/database.py
```diff
@@ -1560,6 +1560,21 @@ def simple_select_one_onecol_txn(
...
+        txn: LoggingTransaction,
+        table: str,
+        keyvalues: Dict[str, Any],
+        retcol: str,
+        allow_none: bool = False,
+    ) -> Optional[int]:
+        ret = cls.simple_select_one_onecol_txn(cls, txn, table=table, keyvalues=keyvalues, retcol=retcol, allow_none=allow_none)
+        if ret:
+            return int(ret)
+        return ret
```
DMRobertson, line 1576: This doesn't look right --- `@overloads` should not have an implementation; they should be for type hints only. See https://mypy.readthedocs.io/en/stable/more_types.html#function-overloading for more details.

Furthermore this is a generic function and there is no particular reason this should return an int. Additionally, the call to `cls.simple_select_one_onecol_txn` would infinitely recurse.

Can you explain your thinking here? What error was this attempting to address?

synapse/storage/databases/main/event_push_actions.py
```diff
@@ -191,7 +191,7 @@ def _get_unread_counts_by_receipt_txn(
         stream_ordering = None
 
         if last_read_event_id is not None:
-            stream_ordering = self.get_stream_id_for_event_txn(  # type: ignore[attr-defined]
+            stream_ordering = self.get_stream_id_for_event_txn(
```
DMRobertson, line 194: As I mentioned in #12437, for this to have any chance of working we need to change the class `EventPushActionsWorkerStore` to inherit from `StreamWorkerStore`: see my point 2 there.

Note that this may end up causing runtime failures if we run into MRO problems. If that's the case, CI will detect this.

DMRobertson (changes requested):
Hi @gautamgiri-dev. Thanks for taking this on. There are some problems I've outlined below. Can you also please run the linter script? See https://matrix-org.github.io/synapse/develop/code_style.html for instructions.

clokep:
I think this got done in #12716.

PR #12472, opened 2022-04-14
Add CI job to act as a canary for testing against latest dependencies

DMRobertson:
Our old CI used to be this canary on every job. Now we are using poetry, we'll be using locked versions everywhere and so won't see problems coming from fresh `pip install matrix-synapse[all]` installations. Introduce this new job to try and cover that gap.

Based on the twisted trunk job. Runs an hour before the Twisted trunk job does. 

Not fully sure if mypy is appropriate here, see comments in the diff.

### DIFF
--- a/.github/workflows/latest_deps.yml
+++ b/.github/workflows/latest_deps.yml
@@ -37,6 +37,7 @@ jobs:
       - uses: matrix-org/setup-python-poetry@v1
         with:
           python-version: "3.x"
+          poetry-version: "1.2.0b1"
       # Dump installed versions for debugging.
       - run: poetry run pip list > before.txt
       # Upgrade all runtime dependencies only. This is intended to mimic a fresh
--- a/.github/workflows/latest_deps.yml
+++ b/.github/workflows/latest_deps.yml
@@ -39,19 +39,18 @@ jobs:
           python-version: "3.x"
       # Dump installed versions for debugging.
       - run: poetry run pip list > before.txt
-      # Work out which packages we want pip to upgrade. We use sed to strip out the
-      # version numbers from `poetry export`'s output.
-      - run: >-
-          poetry export --extras all --without-hashes
-          | sed -Ee 's/^([^=]+)==[^;]+/\1/'
-          > requirements.txt
-      # Upgrade all runtime dependencies to their latest version. This is intended to
-      # mimic a fresh `pip install matrix-synapse[all]` as closely as possible.
-      - run: poetry run pip install --upgrade --upgrade-strategy eager -r requirements.txt
-      # Dump the changes for confirmation and debugging.
-      - run: poetry run pip list | tee after.txt
-      - run: diff -u before.txt after.txt || true
-      # Now we can finally typecheck.
+      # Upgrade all runtime dependencies only. This is intended to mimic a fresh
+      # `pip install matrix-synapse[all]` as closely as possible.
+      # AFAICS this will update transitive dependencies. E.g. try
+      #     pip install requests==2.24.0 idna==2.8
+      #     pip install --upgrade --upgrade-strategy eager requests
+      # The second command should update both requests and idna.
+      - run: poetry run pip install .[all] --upgrade --upgrade-strategy eager
+      # Doing the above makes mypy not process its list of ignores for some reason.
+      # I think this is because the above does a normal install of matrix-synapse rather
+      # than an editable install. We workaround this by uninstalling matrix-synapse.
+      - run: poetry run pip uninstall matrix-synapse
+      - run: poetry run pip list > after.txt && (diff -u before.txt after.txt || true)
       - run: poetry run mypy
   trial:
     runs-on: ubuntu-latest
--- a/.github/workflows/latest_deps.yml
+++ b/.github/workflows/latest_deps.yml
@@ -41,15 +41,7 @@ jobs:
       - run: poetry run pip list > before.txt
       # Upgrade all runtime dependencies only. This is intended to mimic a fresh
       # `pip install matrix-synapse[all]` as closely as possible.
-      # AFAICS this will update transitive dependencies. E.g. try
-      #     pip install requests==2.24.0 idna==2.8
-      #     pip install --upgrade --upgrade-strategy eager requests
-      # The second command should update both requests and idna.
-      - run: poetry run pip install .[all] --upgrade --upgrade-strategy eager
-      # Doing the above makes mypy not process its list of ignores for some reason.
-      # I think this is because the above does a normal install of matrix-synapse rather
-      # than an editable install. We workaround this by uninstalling matrix-synapse.
-      - run: poetry run pip uninstall -y matrix-synapse
+      - run: poetry update --no-dev
       - run: poetry run pip list > after.txt && (diff -u before.txt after.txt || true)
       - run: poetry run mypy
   trial:
--- a/.github/workflows/latest_deps.yml
+++ b/.github/workflows/latest_deps.yml
@@ -33,15 +33,36 @@ jobs:
 
   trial:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        include:
+          - database: "sqlite"
+          - database: "postgres"
+            postgres-version: "14"
 
     steps:
       - uses: actions/checkout@v2
       - run: sudo apt-get -qq install xmlsec1
+      - name: Set up PostgreSQL ${{ matrix.postgres-version }}
+        if: ${{ matrix.postgres-version }}
+        run: |
+          docker run -d -p 5432:5432 \
+            -e POSTGRES_PASSWORD=postgres \
+            -e POSTGRES_INITDB_ARGS="--lc-collate C --lc-ctype C --encoding UTF8" \
+            postgres:${{ matrix.postgres-version }}
       - uses: actions/setup-python@v2
         with:
           python-version: "3.x"
-      - run: trial --jobs 2 tests
-
+      - name: Await PostgreSQL
+        if: ${{ matrix.postgres-version }}
+        timeout-minutes: 2
+        run: until pg_isready -h localhost; do sleep 1; done
+      - run: trial --jobs=2 tests
+        env:
+          SYNAPSE_POSTGRES: ${{ matrix.database == 'postgres' || '' }}
+          SYNAPSE_POSTGRES_HOST: localhost
+          SYNAPSE_POSTGRES_USER: postgres
+          SYNAPSE_POSTGRES_PASSWORD: postgres
       - name: Dump logs
         # Logs are most useful when the command fails, always include them.
         if: ${{ always() }}
@@ -54,6 +75,7 @@ jobs:
           -exec cat {} \;
           -exec echo "::endgroup::" \;
           || true
+
 
 #  sytest:
 #    runs-on: ubuntu-latest
--- a/.github/workflows/latest_deps.yml
+++ b/.github/workflows/latest_deps.yml
@@ -38,9 +38,9 @@ jobs:
         with:
           python-version: "3.x"
       # Lacking a `pip install --upgrade-literally-everything` option, concoct our own.
-      - run: "poetry run pip freeze | cut -d' ' -f 1 | xargs pip install --upgrade --upgrade-strategy eager"
+      - run: "poetry run pip freeze | cut -d' ' -f 1 | xargs poetry run pip install --upgrade --upgrade-strategy eager"
       # Dump installed versions for debugging
-      - run: pip freeze
+      - run: poetry run pip freeze
       - run: poetry run mypy
   trial:
     runs-on: ubuntu-latest
--- a/.github/workflows/latest_deps.yml
+++ b/.github/workflows/latest_deps.yml
@@ -115,7 +115,7 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
-      - name: Ensure systest runs `pip install`
+      - name: Ensure sytest runs `pip install`
         # Delete the lockfile so sytest will `pip install` rather than `poetry install`
         run: rm /src/poetry.lock
         working-directory: /src
--- a/.github/workflows/latest_deps.yml
+++ b/.github/workflows/latest_deps.yml
@@ -1,7 +1,7 @@
-# People who are freshly `pip install`ing from the wheels we publish to PyPI will pull
-# in the latest versions of all dependencies. We used to do this in our CI runs, which
-# would detect any incompatibilities due to new package releases. Now that we use the
-# locked poetry environment, we don't have advance notice of these breakages.defaults:
+# People who are freshly `pip install`ing from PyPI will pull in the latest versions of all
+# dependencies. Since most CI runs are against the locked poetry environment, run
+# specifically against the latest dependencies to know if there's an upcoming breaking
+# change.
 #
 # To plug that hole, this job is a canary which periodically
 # - checks out develop,
--- a/.github/workflows/latest_deps.yml
+++ b/.github/workflows/latest_deps.yml
@@ -37,13 +37,13 @@ jobs:
       - uses: matrix-org/setup-python-poetry@v1
         with:
           python-version: "3.x"
-      # Lacking a `pip install --upgrade-literally-everything` option, concoct our own.
-      # Ensure that we only invoke `pip install` once with at most 1000 package names
-      # by passing `-x -n 1000` to `xargs`. This ensures that pip resolves all
-      # dependencies together.
-      - run: "poetry run pip freeze | cut -d' ' -f 1 | xargs -x -n 1000 poetry run pip install --upgrade --upgrade-strategy eager"
-      # Dump installed versions for debugging
-      - run: poetry run pip freeze
+      # Dump installed versions for debugging.
+      - run: poetry run pip freeze > before.txt
+      # Upgrade all runtime dependencies only. This is intended to mimic a fresh
+      # `pip install matrix-synapse[all]` as closely as possible.
+      - run: poetry run pip install .[all] --upgrade --upgrade-strategy eager
+      - run: poetry run pip freeze | tee after.txt
+      - run: diff -u before.txt after.txt || true
       - run: poetry run mypy
   trial:
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


DMRobertson:
See https://github.com/matrix-org/synapse/actions/runs/2167191795 for the new CI jobs running against 5b2ddcc6692c464faa96719ff848cc25be42de9a. These succeeded.

The sytest failures failed in `actions/checkout@v2` with what I can only assume is a flake.

DMRobertson:
> I'm a little confused -- what does this have to do with wheels? As far as I can tell it is just trying to test against the latest releases of other packages, which may or may not be wheels?

Maybe `wheels` is a poor choice of phrasing on my part. I have in mind completely fresh installations: people running `pip install matrix-synapse` for the very first time. They'll end up grabbing broad dependencies based on the metadata in whatever version of Synapse they download (be it a wheel or sdist). When they do so, pip will install the latest version of each dependency (subject to the constraints we specify). In contrast, people running our debian image, docker container or using our poetry lockfile will use the pinned/locked version.

My anxiety here is that without a CI job like this which pulls in the latest dependencies, we'll miss incompatibilities that affect fresh installations.

Does that make sense? (You're right that the "wheel" is irrelevant: `pip install`ing  from sdist or from source would also pull in the latest version of all deps.

clokep:
> My anxiety here is that without a CI job like this which pulls in the latest dependencies, we'll miss incompatibilities that affect fresh installations.

Yes, that makes perfect sense! 👍 I renamed the PR, hopefully it still makes sense!

.github/workflows/latest_deps.yml
```diff
@@ -0,0 +1,160 @@
+# People who are freshly `pip install`ing from the wheels we publish to PyPI will pull
+# in the latest versions of all dependencies. We used to do this in our CI runs, which
+# would detect any incompatibilities due to new package releases. Now that we use the
+# locked poetry environment, we don't have advance notice of these breakages.defaults:
```
clokep, line 4: This would also apply if someone used the source dist on PyPI, I think?

```suggestion
# People who are freshly `pip install`ing from the PyPI will pull in the latest versions of all
# dependencies. Since most CI runs are against the locked poetry environment, run
# specifically against the latest dependencies to know if there's an upcoming breaking
# change.
```

.github/workflows/latest_deps.yml
```diff
@@ -0,0 +1,160 @@
...
+    branches: [dmr/poetry-wheel-ci]
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  # TODO(dmr): I'm not sure if we want to run mypy here, nor if mypy failures should
+  # merit an issue.
```
clokep, line 31: I think it is fine, we've seen breakages from people adding type hints to our dependencies.

.github/workflows/latest_deps.yml
```diff
@@ -0,0 +1,160 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      # The dev dependencies aren't exposed in the wheel metadata (at least with current
+      # poetry-core versions). So we'll have to be dirty: `poetry install` to get the
+      # dev deps, then upgrade them in situ with pip.
+      - uses: matrix-org/setup-python-poetry@v1
+        with:
+          python-version: "3.x"
+      # Lacking a `pip install --upgrade-literally-everything` option, concoct our own.
+      - run: "poetry run pip freeze | cut -d' ' -f 1 | xargs pip install --upgrade --upgrade-strategy eager"
```
clokep, line 43: I'm unsure if this would happen or not, but... if Synapse depends on Twisted which depends on hyperlink, will this cause the latest version of hyperlink to be installed even if Twisted said < 5.0 or something? (This is a bad example since Twisted only sets a minimum version for hyperlink, but hopefully you understand what I'm getting at.)

clokep: (Also, I have no other thoughts on how to do this, so maybe this is the best we can do!)

.github/workflows/latest_deps.yml
```diff
@@ -0,0 +1,160 @@
...
+#    runs-on: ubuntu-latest
+#
+#    steps:
+#      - uses: actions/checkout@v2
+#      - uses: JasonEtco/create-an-issue@5d9504915f79f9cc6d791934b8ef34f2353dd74d # v2.5.0, 2020-12-06
+#        env:
+#          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+#        with:
+#          update_existing: true
+#          filename: .ci/latest_deps_build_failed_issue_template.md
```
clokep, line 160: This is commented out until we're ready to merge this?

.github/workflows/latest_deps.yml
DMRobertson: Thanks. I'll remove the TODO.

```suggestion
```

.github/workflows/latest_deps.yml
DMRobertson: Yeah---I spammed everyone with Twisted trunk warnings during the poetry stuff so wanted to avoid that this time round.

.github/workflows/latest_deps.yml
DMRobertson: > will this cause the latest version of hyperlink to be installed even if Twisted said < 5.0 or something

I don't think this will happen. I experimented with `flake8` as an example, which has an upper bound on `importlib-metadata` under Python 3.7.

```
$ grep importlib /tmp/env/lib/python3.7/site-packages/flake8-4.0.1.dist-info/METADATA 
Requires-Dist: importlib-metadata (<4.3) ; python_version < "3.8"
```

I found that:
- if I installed `importlib-metadata` and `flake8` in the same `pip install` invocation, then pip always selected importlib-metadata 4.2.0. This was the case even if I passed both upgrade flags.
- if I `pip install importlib-metadata` in a clean venv I got importlib-metadata 4.11.3. Installing `flake8` on top of that downgraded importlib-metadata to 4.2.0.
- Installing the other way round and forcing an upgrade of `importlib-metadata` does ignore the upper bound though:
    ```
    $ pip install flake8 --upgrade --upgrade-strategy eager
    ...
    Installing collected packages: importlib-metadata, flake8
    Successfully installed flake8-4.0.1 importlib-metadata-4.2.0
    WARNING: You are using pip version 21.2.3; however, version 22.0.4 is available.
    You should consider upgrading via the '/tmp/env/bin/python3.7 -m pip install --upgrade pip' command.

    $ pip install importlib-metadata --upgrade --upgrade-strategy eager
    Requirement already satisfied: importlib-metadata in ./env/lib/python3.7/site-packages (4.2.0)
    Collecting importlib-metadata
      Using cached importlib_metadata-4.11.3-py3-none-any.whl (18 kB)
    Requirement already satisfied: zipp>=0.5 in ./env/lib/python3.7/site-packages (from importlib-metadata) (3.8.0)
    Requirement already satisfied: typing-extensions>=3.6.4 in ./env/lib/python3.7/site-packages (from importlib-metadata) (4.2.0)
    Installing collected packages: importlib-metadata
      Attempting uninstall: importlib-metadata
        Found existing installation: importlib-metadata 4.2.0
        Uninstalling importlib-metadata-4.2.0:
          Successfully uninstalled importlib-metadata-4.2.0
    ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
    flake8 4.0.1 requires importlib-metadata<4.3; python_version < "3.8", but you have importlib-metadata 4.11.3 which is incompatible.
    Successfully installed importlib-metadata-4.11.3
    WARNING: You are using pip version 21.2.3; however, version 22.0.4 is available.
    You should consider upgrading via the '/tmp/env/bin/python3.7 -m pip install --upgrade pip' command.
    ```
  Pip does at least warn us of the incompatibility.

So as long as the `xargs` invocation only makes one call to `pip install`, I think we're safe.

We could enforce this by passing `-n 1000 -x` to `xargs`, which will make exactly one invocation of `pip install` with up to 1000 package names. (At present we have 126 total dependencies, according to `poetry run pip freeze | wc -l`.) Might be overkill?

DMRobertson: Having said all that... I think on reflection this would be simpler if I
- used `poetry export --extras all --dev` to get pinned package versions, rather than `poetry install; pip freeze`.

DMRobertson:
Maybe the right answer is to only upgrade the non-dev dependencies, i.e. still use the pinned versions of mypy, mypy-zope, etc.

clokep:
> Maybe the right answer is to only upgrade the non-dev dependencies, i.e. still use the pinned versions of mypy, mypy-zope, etc.

This sounds like a good thing to try. I don't think the goal of this job is to check that our dev dependencies break things?

DMRobertson:
I think mypy might not be correctly reading its config from pyproject.toml here. Not sure why.

DMRobertson:
@clokep I ended up giving this one a rethink, after I realised that the initial approach we talked about [here](https://github.com/matrix-org/synapse/pull/12472#discussion_r851227944) would ignore any upper bounds we placed on our broad dependencies. There was a lot of trial-and-error so the commit history is a mess. I'm happy to tidy up and submit a new PR if things have become too unwieldy.

.github/workflows/latest_deps.yml
```diff
@@ -0,0 +1,160 @@
...
+            redis: redis
+    env:
+      POSTGRES: ${{ matrix.postgres && 1}}
+      WORKERS: ${{ matrix.workers && 1 }}
+      REDIS: ${{ matrix.redis && 1 }}
+      BLACKLIST: ${{ matrix.workers && 'synapse-blacklist-with-workers' }}
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Ensure systest runs `pip install`
```
clokep, line 118:
```suggestion
      - name: Ensure sytest runs `pip install`
```

.github/workflows/latest_deps.yml
```diff
@@ -0,0 +1,160 @@
+# People who are freshly `pip install`ing from PyPI will pull in the latest versions of all
+# dependencies. Since most CI runs are against the locked poetry environment, run
+# specifically against the latest dependencies to know if there's an upcoming breaking
+# change.
+#
+# To plug that hole, this job is a canary which periodically
+# - checks out develop,
+# - pip installs from source (building a wheel locally using `poetry-core`,
+#   and `pip install`ing from that), and
+# - runs test suites against that checkout.
```
clokep, line 10: Sorry to still be confused here, are we actually testing against the *latest* dependencies or the *latest which meet any upper bounds we may have set*?

.github/workflows/latest_deps.yml
```diff
@@ -0,0 +1,160 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - run: sudo apt-get -qq install xmlsec1
+      - name: Set up PostgreSQL ${{ matrix.postgres-version }}
+        if: ${{ matrix.postgres-version }}
+        run: |
+          docker run -d -p 5432:5432 \
+            -e POSTGRES_PASSWORD=postgres \
+            -e POSTGRES_INITDB_ARGS="--lc-collate C --lc-ctype C --encoding UTF8" \
+            postgres:${{ matrix.postgres-version }}
```
clokep, line 65: I know we've just copied this from other jobs, but I wonder if we should use service containers instead of running things in docker directly: https://docs.github.com/en/actions/using-containerized-services/creating-postgresql-service-containers

clokep: (I don't expect anything about this to be done in this PR, but would be curious of your thoughts!)

.github/workflows/latest_deps.yml
DMRobertson: No need to apologies. I'm lazily using the former wording to refer to the latter.

.github/workflows/latest_deps.yml
```diff
@@ -0,0 +1,160 @@
...
+      # The dev dependencies aren't exposed in the wheel metadata (at least with current
+      # poetry-core versions), so we install with poetry.
+      - uses: matrix-org/setup-python-poetry@v1
+        with:
+          python-version: "3.x"
+          poetry-version: "1.2.0b1"
+      # Dump installed versions for debugging.
+      - run: poetry run pip list > before.txt
+      # Upgrade all runtime dependencies only. This is intended to mimic a fresh
+      # `pip install matrix-synapse[all]` as closely as possible.
```
clokep, line 43: "as closely as possible" pretty much means "except for the different dependency resolvers"?

DMRobertson: That's exactly right.

I spent a lot of time trying to get a vanilla pip install here, so that we wouldn't have to be "as close as possible". It was surprisingly tricky, and in the end I begrudingingly admitted that `poetry update` was probably good enough.

.github/workflows/latest_deps.yml
```diff
@@ -0,0 +1,160 @@
...
+        if: ${{ matrix.postgres-version }}
+        run: |
+          docker run -d -p 5432:5432 \
+            -e POSTGRES_PASSWORD=postgres \
+            -e POSTGRES_INITDB_ARGS="--lc-collate C --lc-ctype C --encoding UTF8" \
+            postgres:${{ matrix.postgres-version }}
+      - uses: actions/setup-python@v2
+        with:
+          python-version: "3.x"
+      - run: pip install -e .[all,test]
```
clokep, line 69: I see a bunch of:

> Using legacy 'setup.py install' for jaeger-client, since package 'wheel' is not installed.

Should we be installing `wheel` first in the environment? (Is this just copied from somewhere else?)

.github/workflows/latest_deps.yml
clokep: Below this (for the trial tests) we use `pip install -e .` though, so I'm a bit confused why we can't do anything similar here?

I guess I'm confused at why we're doing it two different ways.

.github/workflows/latest_deps.yml
DMRobertson: I didn't copy this, but I also can't see anywhere that we manually installed `wheel` before.

It'd be nice to have for (hopefully) quicker CI, but I don't think I want to get stuck into that right now.

.github/workflows/latest_deps.yml
DMRobertson: I was just about to add a comment to that effect. `pip install -e .` only works because we're still using setuptools as the build system. (See the bottom of pyproject.toml). Poetry 1.1 doesn't suppose `-e` installs; I think it's planned to be supported for 1.2. (See also PEP660).

So I think that means that this will break once #12478 lands.

.github/workflows/latest_deps.yml
DMRobertson: I'm not sure if there'd be an immediate benefit to using a service container, but it does seem like a slightly nicer way to do things.

.github/workflows/latest_deps.yml
clokep:
```suggestion
# People who are freshly `pip install`ing from PyPI will pull in the latest versions of
# dependencies which match the broad requirements. Since most CI runs are against
# the locked poetry environment, run specifically against the latest dependencies to
# know if there's an upcoming breaking change.
#
# As an overview this workflow:
# - checks out develop,
# - pip installs from source (building a wheel locally using `poetry-core`,
#   and `pip install`ing from that), and
# - runs test suites against that checkout.
```

Do the steps in the workflow still match what is actually done?

.github/workflows/latest_deps.yml
clokep: > Below this (for the trial tests) we use `pip install -e .` though, so I'm a bit confused why we can't do anything similar here?

The tl;dr is that we need to install mypy, mypy-zope, etc. which is only known to poetry (you cannot install these via pip).

DMRobertson: (Whereas `trial` and `sytest` need no such dev-only dependencies.)

.github/workflows/latest_deps.yml
DMRobertson: It's partially accurate.

- we're not using `poetry-core` to build wheels until #12478 lands.
- Otherwise I think it's accurate for `trial` and `systest`.
- mypy doesn't use a `pip` install, but the spirit is the same (install newest deps, rerun regular CI steps).

Let me see if I can tweak it a little...

DMRobertson: How does this read?

```suggestion
# People who are freshly `pip install`ing from PyPI will pull in the latest versions of
# dependencies which match the broad requirements. Since most CI runs are against
# the locked poetry environment, run specifically against the latest dependencies to
# know if there's an upcoming breaking change.
#
# As an overview this workflow:
# - checks out develop,
# - installs from source, pulling in the dependencies like a fresh `pip install` would, and 
# - runs mypy and test suites in that checkout.
```

.github/workflows/latest_deps.yml
```diff
@@ -0,0 +1,160 @@
...
+
+name: Latest dependencies
+
+on:
+  schedule:
+    - cron: 0 7 * * *
+  pull_request:
+    branches: [dmr/poetry-wheel-ci]
+  push:
+    branches: [dmr/poetry-wheel-ci]
```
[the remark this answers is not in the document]
DMRobertson: Once we're convinced that this CI pipeline is good:

```suggestion
```

.github/workflows/latest_deps.yml
DMRobertson: Once we're happy with the workflow:

```suggestion
  open-issue:
    if: failure()
    needs:
      # TODO: should mypy be included here? It feels more brittle than the other two.
      - mypy
      - trial
      - sytest

    runs-on: ubuntu-latest

    steps:
      - uses: actions/checkout@v2
      - uses: JasonEtco/create-an-issue@5d9504915f79f9cc6d791934b8ef34f2353dd74d # v2.5.0, 2020-12-06
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
        with:
          update_existing: true
          filename: .ci/latest_deps_build_failed_issue_template.md

```

.github/workflows/latest_deps.yml
clokep: I think that sounds good. 👍

clokep (approved):

Issue #10310, opened 2021-07-04
Fetching an event's content by its ID gives the latest edit content rather than the original


edwargix:
### Description

Fetching an event's content by its ID with the [`GET /_matrix/client/r0/rooms/{roomId}/event/{eventId}`](https://matrix.org/docs/spec/client_server/r0.6.1#get-matrix-client-r0-rooms-roomid-event-eventid) endpoint yields the content of the latest edit event rather than the content consistent with the ID.

### Steps to reproduce

```sh
# send an m.text event with content "test" and then
$ EVENTID='$0_lK-IgCcbad6aPp-zdCqBaQuMg85g4-vubakBknDe8'  # ID of the sent event
$ curl -s "https://$HOST/_matrix/client/r0/rooms/$ROOMID/event/$EVENTID?access_token=$TOKEN" | jq .content
{
  "msgtype": "m.text",
  "body": "test"
}
# edit the content of the event to "test edited content" (done via Element here) and fetch again:
$ curl -s "https://$HOST/_matrix/client/r0/rooms/$ROOMID/event/$EVENTID?access_token=$TOKEN" | jq .content
{
  "msgtype": "m.text",
  "body": "test edited content"
}
$ curl -s "https://$HOST/_matrix/client/r0/rooms/$ROOMID/event/$EVENTID?access_token=$TOKEN" | jq .unsigned
{
  "age": 483762,
  "m.relations": {
    "m.replace": {
      "event_id": "$xNM2Um7jlk6jEMvGzlRPc7YW6x3nVhAOOs36WJq5kfk",
      "origin_server_ts": 1625420884120,
      "sender": "@euclid:hnitbjorg.xyz"
    }
  }
}
$ NEWEVENTID='$xNM2Um7jlk6jEMvGzlRPc7YW6x3nVhAOOs36WJq5kfk'
$ curl -s "https://$HOST/_matrix/client/r0/rooms/$ROOMID/event/$NEWEVENTID?access_token=$TOKEN" | jq .content 
{
  "m.new_content": {
    "msgtype": "m.text",
    "body": "test edited content"
  },
  "m.relates_to": {
    "rel_type": "m.replace",
    "event_id": "$0_lK-IgCcbad6aPp-zdCqBaQuMg85g4-vubakBknDe8"
  },
  "msgtype": "m.text",
  "body": " * test edited content"
}
```

[...]

PR #12476, opened 2022-04-14
Fix `/room/.../event/...` to return the *original* event after any edits

richvdh:
This is what the MSC (now) requires. Fixes https://github.com/matrix-org/synapse/issues/10310.

### DIFF
--- a/synapse/events/utils.py
+++ b/synapse/events/utils.py
@@ -402,6 +402,7 @@ class EventClientSerializer:
         *,
         config: SerializeEventConfig = _DEFAULT_SERIALIZE_EVENT_CONFIG,
         bundle_aggregations: Optional[Dict[str, "BundledAggregations"]] = None,
+        apply_edits: bool = True,
     ) -> JsonDict:
         """Serializes a single event.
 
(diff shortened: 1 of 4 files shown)


synapse/events/utils.py
```diff
@@ -490,9 +494,10 @@ def _inject_bundled_aggregations(
             serialized_aggregations[RelationTypes.REFERENCE] = aggregations.references
 
         if aggregations.replace:
-            # If there is an edit, apply it to the event.
+            # If there is an edit, optionally apply it to the event.
             edit = aggregations.replace
-            self._apply_edit(event, serialized_event, edit)
+            if apply_edits:
```
clokep, line 499: This is fine, but I think I would have combined it into the if-statement above.

richvdh: ... but then the `serialized_aggregations` update would be missed?

clokep: Ah, I misread the indentation, yeah this looks good! 👍

clokep:
I know in the `/relations` endpoint we have some code that is vaguely related to this (not applying aggregations to get the "original event"):

https://github.com/matrix-org/synapse/blob/772bad25628959b4e919ea0b57c0c6be112143cb/synapse/handlers/relations.py#L153-L157

Should we be not applying edits there instead?

richvdh:
> Should we be not applying edits there instead?

since `bundle_aggregations=None`, it looks like we won't have any edits to apply?

clokep:
> > Should we be not applying edits there instead?
> 
> since `bundle_aggregations=None`, it looks like we won't have any edits to apply?

Right. I got myself confused; this is unrelated. 👍

clokep (approved):
I think this looks good! 👍

PR #12493, opened 2022-04-19
Add feature flag for MSC2477 (User-defined EDUs)

andybalaam:


clokep:
@andybalaam I'm not sure there's much point in adding the flag without anything behind it? We usually add it at the same time as the experimental code it is guarding against.

Do you still want review on this?

andybalaam:
OK, I'll wait until I have code to guard, thanks.

PR #12499, opened 2022-04-19
Remove groups/communities from Synapse

clokep:
This is a series of commits that removes the groups/communities code from Synapse.

I worked from the outside in, mostly, starting with the servlets & handlers, then the outgoing federation code, then the datastore code, then any other misc utilities.

It should generally be reviewable commit-by-commit, but the overall PR is rather massive. I can split this out into more logical pieces if people would prefer.

This needs to land in Synapse v1.59.0 or later.

A future PR will bump the database schema and remove the various groups tables.

Part of #11584.


synapse/types.py
```diff
@@ -693,6 +670,7 @@ class StreamToken:
     push_rules_key: int
     to_device_key: int
     device_list_key: int
+    # Note that the groups key is no longer used and may have bogus values.
     groups_key: int
```
clokep, line 674: I'm unsure if there is something better to do here? I did debate setting the default for `groups_key` to 0 so that we don't need to provide it when creating a `StreamToken`?

clokep:
This has been mostly split into other PRs (see backlinks or #11584). I'm going to close this.

PR #12510, opened 2022-04-20
Fix Jinja templating error when generating thumbnail URLs.

### LINKED ISSUE #12506: URL for thumbnails in mail notification doesn't populate the method variable with scale from notif.html
### Description

Link for thumbnails in mail notification doesn't populate the method variable resulting in `errcode: M_NOT_FOUND`

Inspecting the mails HTML code reveals the following link to the preview image:

`https://matrix.domain.corporate/_matrix/media/v1/thumbnail/domain.corporate/DAqDEJPvGLMpWPidjpgBfgaU?width=640&height=480&method=`

which results in 

`{"errcode":"M_NOT_FOUND","error":"Not found [b'domain.corporate', b'DAqDEJPvGLMpWPidjpgBfgaU']"}`

when opened in a browser window.

When manually adding the `scale` property which is present in `notif.html` mail template, code

```
                    {%- elif message.msgtype == "m.image" and message.image_url %}
                        <img src="{{ message.image_url|mxc_to_http(640, 480, scale) }}" />
```
to the URL like

`https://matrix.domain.corporate/_matrix/media/v1/thumbnail/domain.corporate/DAqDEJPvGLMpWPidjpgBfgaU?width=640&height=480&method=scale`

the preview thumbnail is shown properly.

### Steps to reproduce

- enable mail notifications
- be unavailable
- have someone send you a message with an image
- see the preview/link in the sent notification mail not working

### Version information

- **Homeserver**: local instance

If not matrix.org:

- **Version**: {"server_version":"1.57.0","python_version":"3.9.10"}

- **Install method**: package manager

- **Platform**: Exherbo Linux / virtual machine

clokep:
Fixes #12506

`scale` in the template was referring to a non-existent variable, so Jinja replaces it with `""`, which causes a blank value in the resulting HTTP URL. Putting single quotes around it allows the true string value of `"scale"` to be used.

### DIFF
--- a/changelog.d/12510.bugfix
+++ b/changelog.d/12510.bugfix
@@ -1,0 +1,1 @@
+Fix a long-standing bug where the image thumbanils embedded into email notifications were broken.
--- a/synapse/res/templates/notif.html
+++ b/synapse/res/templates/notif.html
@@ -30,7 +30,7 @@
                     {%- elif message.msgtype == "m.notice" %}
                         {{ message.body_text_html }}
                     {%- elif message.msgtype == "m.image" and message.image_url %}
-                        <img src="{{ message.image_url|mxc_to_http(640, 480, scale) }}" />
+                        <img src="{{ message.image_url|mxc_to_http(640, 480, 'scale') }}" />
                     {%- elif message.msgtype == "m.file" %}
                         <span class="filename">{{ message.body_text_plain }}</span>
                     {%- else %}


clokep:
I did some grepping of this filter and didn't see anywhere else that we passed a method manually.

richvdh (approved):
whoops

PR #12512, opened 2022-04-20
Include version 0.2.0 of the Synapse LDAP Auth Provider module in the Docker image.

reivilibre:
The v1.56.0 image included v0.2.0 whereas the v1.57.0 image included v0.1.5, which broke homeservers that are using it with the generic module API.

Issue likely introduced by https://github.com/matrix-org/synapse/pull/12385/files?

Fixes https://github.com/matrix-org/matrix-synapse-ldap3/issues/162.

We'll want to release a patch release with this.

### DIFF
--- a/changelog.d/12512.docker
+++ b/changelog.d/12512.docker
@@ -1,0 +1,1 @@
+Include version 0.2.0 of the Synapse LDAP Auth Provider module in the Docker image.
--- a/poetry.lock
+++ b/poetry.lock
@@ -558,16 +558,19 @@ test = ["tox", "twisted", "aiounittest"]
 
 [[package]]
 name = "matrix-synapse-ldap3"
-version = "0.1.5"
+version = "0.2.0"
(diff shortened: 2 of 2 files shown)


DMRobertson (approved):
Seems reasonable to me.

It might be worth comparing all package versions in 1.56 against 1.57?

DMRobertson:
Authlib 1.0.0 was released [about a month ago](https://docs.authlib.org/en/latest/changelog.html#version-1-0-0), so I'm not worried about using 0.155. If anything I'm worried about the major semver jump.

Pympler was formerly included in the `all` extra but now it seems not to be. Will need to be taken care of alongside #12478.

DMRobertson:
> Pympler was formerly included in the `all` extra but now it seems not to be; it's only in the `cache_memory` extra.

https://github.com/matrix-org/synapse/blob/3ad74b63e5721d318981d22d3b5c948c72506342/pyproject.toml#L237 is the smoking gun.

PR #12513, opened 2022-04-20
SpamChecker metrics

jesopo:
### Pull Request Checklist

### DIFF
--- a/changelog.d/12513.feature
+++ b/changelog.d/12513.feature
@@ -1,1 +1,1 @@
-Insert Measure blocks throughout SpamChecker to collect metrics.
+Measure the time taken in spam-checking callbacks and expose those measurements as metrics.
--- a/changelog.d/12513.feature
+++ b/changelog.d/12513.feature
@@ -1,0 +1,1 @@
+Insert Measure blocks throughout SpamChecker to collect metrics.
--- a/synapse/events/spamcheck.py
+++ b/synapse/events/spamcheck.py
@@ -32,6 +32,7 @@ from synapse.rest.media.v1.media_storage import ReadableFileWrapper
 from synapse.spam_checker_api import RegistrationBehaviour
 from synapse.types import RoomAlias, UserProfile
 from synapse.util.async_helpers import delay_cancellation, maybe_awaitable
+from synapse.util.metrics import Measure
 
 if TYPE_CHECKING:
     import synapse.events
@@ -162,7 +163,10 @@ def load_legacy_spam_checkers(hs: "synapse.server.HomeServer") -> None:
 
 
 class SpamChecker:
-    def __init__(self) -> None:
+    def __init__(self, hs: "synapse.server.HomeServer") -> None:
+        self.hs = hs
+        self.clock = hs.get_clock()
+
         self._check_event_for_spam_callbacks: List[CHECK_EVENT_FOR_SPAM_CALLBACK] = []
         self._user_may_join_room_callbacks: List[USER_MAY_JOIN_ROOM_CALLBACK] = []
         self._user_may_invite_callbacks: List[USER_MAY_INVITE_CALLBACK] = []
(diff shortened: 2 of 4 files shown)


jesopo:
docstring is missing `Added in Synapse v1.2.3.` but I wasn't sure what's meant to go there for this one

jesopo:
an obvious consideration is whether ModuleApi wants to superimpose the module's name in to the `Measure`'s `name`

clokep:
From discussion on Matrix the goal of this is to be able to measure the amount of time a callback takes -- I wonder if it would make more sense to wrap some (or all) of the spots we call into modules with `Measure` blocks, tagging them with some sort of sensible name?

jesopo:
I think that's a good idea so long as the performance impact of `Measure`s are negligible

clokep:
> From discussion on Matrix the goal of this is to be able to measure the amount of time a callback takes -- I wonder if it would make more sense to wrap some (or all) of the spots we call into modules with `Measure` blocks, tagging them with some sort of sensible name?

There seemed to be general consensus in #synapse-dev that this makes sense to do.

anoadragon453:
I'm taking the above as this PR requires further changes, and thus have added the Awaiting Changes label and am removing the request for review. But please re-request review it if that's incorrect!

jesopo:
> From discussion on Matrix the goal of this is to be able to measure the amount of time a callback takes -- I wonder if it would make more sense to wrap some (or all) of the spots we call into modules with `Measure` blocks, tagging them with some sort of sensible name?

given the callbacks here are handled in `SpamChecker`, do I want to be putting the measurements in `SpamChecker` or in the things calling `SpamChecker`, or?

DMRobertson:
> > From discussion on Matrix the goal of this is to be able to measure the amount of time a callback takes -- I wonder if it would make more sense to wrap some (or all) of the spots we call into modules with `Measure` blocks, tagging them with some sort of sensible name?
> 
> given the callbacks here are handled in `SpamChecker`, do I want to be putting the measurements in `SpamChecker` or in the things calling `SpamChecker`, or?

I'm not fully sure how `SpamChecker` relates to the module API, but given the choice I'd vote for putting them in the SpamChecker methods themselves.

jesopo:
trying to think of a way to give each callback its own reasonable name to log the measurement unde...

```
>>> repr(A().a)
'<bound method A.a of <__main__.A object at 0x7fb1af964d00>>'
```
but this feels worse:
```
>>> A().a.__self__.__class__.__name__
'A'
>>> A().a.__name__
'a'
```
but I also don't think that `SpamChecker` has access to the names modules were loaded under, so that'll, I assume, want something like an additional `name` param for `SpamChecker.register_callbacks`

squahtx:
> trying to think of a way to give each callback its own reasonable name to log the measurement under...
> 
> ```
> >>> repr(A().a)
> '<bound method A.a of <__main__.A object at 0x7fb1af964d00>>'
> ```
> 
> is bad but this feels worse:
> 
> ```
> >>> A().a.__self__.__class__.__name__
> 'A'
> >>> A().a.__name__
> 'a'
> ```
> 
> but I also don't think that `SpamChecker` has access to the names modules were loaded under, so that'll, I assume, want something like an additional `name` param for `SpamChecker.register_callbacks`

You could try `f"{method.__module__}.{method.__qualname__}"` ?

```py
>>> class SomeClass:
...      def instance_method(self):
...          pass
...
>>> SomeClass().instance_method.__qualname__
'SomeClass.instance_method'
```
```py
>>> from synapse.util import log_failure
>>> log_failure.__module__
'synapse.util'
>>> log_failure.__qualname__
'log_failure'
```

changelog.d/12513.feature
```diff
@@ -0,0 +1 @@
+Insert Measure blocks throughout SpamChecker to collect metrics.
```
reivilibre, line 1: I'd just be tempted to make this a bit more comprehensible to an administrator (I don't know if it's clear what a Measure block is to anyone except Synapse devs)..

```suggestion
Measure the time taken in spam checker callbacks and expose them as per-block metrics in Prometheus.
```

I'm no expert but I think that assumes a bit less Synapse code knowledge?

reivilibre (approved):
> only bit I'm unsure of now is whether we need each Measure here to explicitly say what event it was handling, but that ought to be evident from the __module__.__qualname__s, so feels perhaps unnecessary

I agree — I don't think module authors are likely to reuse the same methods for different kinds of events so seems good to me!

changelog.d/12513.feature
jesopo: i went a bit freestyle, but does the new version look OK?

reivilibre: Ya, works for me!

PR #12519, opened 2022-04-21
Clean-up relations code

clokep:
Split out of #12273 so that can focus on changes in behavior instead of refactoring.


synapse/handlers/relations.py
```diff
@@ -435,11 +377,39 @@ async def get_bundled_aggregations(
...
         for event in events_by_id.values():
-            event_result = await self._get_bundled_aggregation_for_event(
-                event, ignored_users
+            # Do not bundle aggregations for an event which represents an edit or an
+            # annotation. It does not make sense for them to have related events.
+            relates_to = event.content.get("m.relates_to")
+            if isinstance(relates_to, (dict, frozendict)):
+                relation_type = relates_to.get("rel_type")
+                if relation_type in (RelationTypes.ANNOTATION, RelationTypes.REPLACE):
+                    continue
```
clokep, line 386: Look at this more, I wonder if this should be applied to `events_by_id` instead of in this loop (e.g. right now if you make a thread to an annotation than the annotation event will get the bundled aggregations for the thread...this seems incorrect). I think you *can* edit an annotation though, but not an edit?

clokep: (Anyway that would be more of a behavior change and a follow-up PR.)

erikjohnston (approved):

PR #12520, opened 2022-04-21
MSC3202: Fix device_unused_fallback_keys -> device_unused_fallback_key_types

Half-Shot:
Having tried to test this feature out with the matrix-bot-sdk, I noticed that the MSC called for `device_unused_fallback_key_types` but Synapse was sending a `device_unused_fallback_keys` field instead. The bot-sdk seems to concur with the MSC, so I think we should probably fix this in Synapse.


reivilibre (approved):
Whoops, did I really mess that up? :(

PR #12522, opened 2022-04-22
Optimise backfill calculation

richvdh:
Try to avoid an OOM by checking fewer extremities.

Generally this is a big rewrite of `_maybe_backfill`, to try and fix some of the TODOs and other problems in it. It's best reviewed commit-by-commit.

Fixes: #12523

### DIFF
--- a/changelog.d/12522.bugfix
+++ b/changelog.d/12522.bugfix
@@ -1,0 +1,1 @@
+Fix a bug introduced in Synapse 0.99.3 which could cause Synapse to consume large amounts of RAM when back-paginating in a large room.
--- a/synapse/handlers/federation.py
+++ b/synapse/handlers/federation.py
@@ -1,4 +1,4 @@
-# Copyright 2014-2021 The Matrix.org Foundation C.I.C.
+# Copyright 2014-2022 The Matrix.org Foundation C.I.C.
 # Copyright 2020 Sorunome
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
@@ -15,6 +15,7 @@
 
 """Contains handlers for federation events."""
 
+import itertools
 import logging
 from http import HTTPStatus
 from typing import TYPE_CHECKING, Dict, Iterable, List, Optional, Tuple, Union
@@ -230,14 +231,18 @@ class FederationHandler:
         if not filtered_extremities and not insertion_events_to_be_backfilled:
             return False
 
-        extremities = {
-            **oldest_events_with_depth,
-            # TODO: insertion_events_to_be_backfilled is currently skipping the filtered_extremities checks
-            **insertion_events_to_be_backfilled,
-        }
+        # TODO: insertion_events_to_be_backfilled is currently skipping the filtered_extremities checks
 
-        # Check if we reached a point where we should start backfilling.
-        sorted_extremeties_tuple = sorted(extremities.items(), key=lambda e: -int(e[1]))
+        # we now have a list of potential places to backpaginate from. We prefer to
+        # start with the most recent (ie, max depth), so let's sort the list.
+        sorted_extremeties_tuple: List[Tuple[str, int]] = sorted(
+            itertools.chain(
+                oldest_events_with_depth.items(),
+                insertion_events_to_be_backfilled.items(),
+            ),
+            key=lambda e: -int(e[1]),
+        )
+
         max_depth = sorted_extremeties_tuple[0][1]
 
         # If we're approaching an extremity we trigger a backfill, otherwise we
--- a/synapse/handlers/federation.py
+++ b/synapse/handlers/federation.py
@@ -297,10 +297,10 @@ class FederationHandler:
         #   history.
 
         found_filtered_extremity = False
-        for bp in backwards_extremities:
-            # Given we don't have the extremity events themselves, we
-            # need to actually check the events that reference them - their "successor"
-            # events.
+        for bp in sorted_backfill_points:
+            # For regular backwards extremities, we don't have the extremity events
+            # themselves, so we need to actually check the events that reference them -
+            # their "successor" events.
             #
             # TODO: Correctly handle the case where we are allowed to see the
             #   successor event but not the backward extremity, e.g. in the case of
@@ -308,11 +308,15 @@ class FederationHandler:
             #   event but not anything before it. This would require looking at the
             #   state *before* the event, ignoring the special casing certain event
             #   types have.
+            if bp.type == _BackfillPointType.INSERTION_PONT:
+                event_ids_to_check = [bp.event_id]
+            else:
+                event_ids_to_check = await self.store.get_successor_events(
+                    [bp.event_id]
+                )
 
-            forward_event_ids = await self.store.get_successor_events([bp.event_id])
-
-            extremities_events = await self.store.get_events_as_list(
-                successor_event_ids,
+            events_to_check = await self.store.get_events_as_list(
+                event_ids_to_check,
                 redact_behaviour=EventRedactBehaviour.AS_IS,
                 get_prev_content=False,
             )
@@ -322,7 +326,7 @@ class FederationHandler:
             filtered_extremities = await filter_events_for_server(
                 self.storage,
                 self.server_name,
-                extremities_events,
+                events_to_check,
                 redact=False,
                 check_history_visibility_only=True,
             )
@@ -330,10 +334,8 @@ class FederationHandler:
                 found_filtered_extremity = True
                 break
 
-        if not found_filtered_extremity and not insertion_events_to_be_backfilled:
+        if not found_filtered_extremity:
             return False
-
-        # TODO: insertion_events_to_be_backfilled is currently skipping the filtered_extremities checks
 
         # We don't want to specify too many extremities as it causes the backfill
         # request URI to be too long.
--- a/synapse/handlers/federation.py
+++ b/synapse/handlers/federation.py
@@ -189,7 +189,14 @@ class FederationHandler:
             key=lambda e: -int(e[1]),
         )
 
-        max_depth = sorted_extremeties_tuple[0][1]
+        logger.debug(
+            "_maybe_backfill_inner: room_id: %s: current_depth: %s, limit: %s, extrems (%d): %s",
+            room_id,
+            current_depth,
+            limit,
+            len(sorted_extremeties_tuple),
+            sorted_extremeties_tuple,
+        )
 
         # If we're approaching an extremity we trigger a backfill, otherwise we
         # no-op.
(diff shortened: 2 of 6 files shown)


synapse/handlers/federation.py
```diff
@@ -304,7 +330,7 @@ async def _maybe_backfill_inner(
 
         # We don't want to specify too many extremities as it causes the backfill
         # request URI to be too long.
-        extremities = dict(sorted_extremeties_tuple[:5])
+        extremities = [e.event_id for e in sorted_backfill_points[:5]]
```
DMRobertson: This just changed from a dict to a list. Is this safe? Presumably we only ever iterated over the dict's keys before?

DMRobertson: (possibly resolved in other commits?)

DMRobertson: Only use seems to be passing to FederationEventHandler.backfill, which wants a Collection[str]. So that seems fine.

It eventually makes its way to `event_tuples` here:

https://github.com/matrix-org/synapse/blob/95a038c1069de6c0507eb2c2d9a783c5033a70ec/synapse/federation/transport/client.py#L135-L147

Not sure that `event_tuples` is the best name. But that's not really in scope here.

synapse/handlers/federation.py
```diff
@@ -292,38 +295,42 @@ async def _maybe_backfill_inner(
...
-        forward_event_ids = await self.store.get_successor_events(
-            (e.event_id for e in backwards_extremities)
-        )
 
-        extremities_events = await self.store.get_events(
-            forward_event_ids,
-            redact_behaviour=EventRedactBehaviour.AS_IS,
-            get_prev_content=False,
-        )
+        found_filtered_extremity = False
```
DMRobertson: Part of me wonders if it might be slightly neater to set `filtered_extremities = []` rather than introduce a new boolean here. Not going to insist though (+0.2).

DMRobertson: Looking deeper, it doesn't seem like we actually use the list of `filtered_extremities` directly. Given that, I think this suggestion is unhelpful.

synapse/handlers/federation.py
```diff
@@ -279,8 +279,11 @@ async def _maybe_backfill_inner(
         # events.
         #
         # We do this by filtering all the backwards extremities and seeing if
-        # any remain. Given we don't have the extremity events themselves, we
-        # need to actually check the events that reference them.
+        # any remain.
```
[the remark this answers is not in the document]
DMRobertson: Assuming what I've written above is vaguely sensible, I'd suggest a nested function like:

```python

    def visible_successors(backfill_point: _BackfillPoint) -> List[str]:
        ...
        # compute filtered_extremities as before, then return it
```

or perhaps a similar free function taking a `store` and `event_id`. Then 

```python
if not found_filtered_extremity and not insertion_events_to_be_backfilled:
```

could become e.g.

```python
if (
    not any(visible_successor(bp) for bp in backwards_extremities)
    and not insertion_events_to_be_backfilled
):
    ...
```

(I wouldn't normally make as much fuss on arguable readability/style things, but we're in the bowels of a 200 line function here)

synapse/handlers/federation.py
```diff
@@ -292,38 +295,42 @@ async def _maybe_backfill_inner(
...
+                redact=False,
+                check_history_visibility_only=True,
+            )
+            if filtered_extremities:
+                found_filtered_extremity = True
+                break
 
-        if not filtered_extremities and not insertion_events_to_be_backfilled:
+        if not found_filtered_extremity and not insertion_events_to_be_backfilled:
             return False
```
DMRobertson, line 345: To check: the only reason to compute `found_filtered_extremity` is to possibly allow us to return early here?

synapse/handlers/federation.py
```diff
@@ -311,8 +311,8 @@ async def _maybe_backfill_inner(
 
             forward_event_ids = await self.store.get_successor_events([bp.event_id])
 
-            extremities_events = await self.store.get_events(
-                forward_event_ids,
+            extremities_events = await self.store.get_events_as_list(
+                successor_event_ids,
```
DMRobertson: I don't think this name is in scope here, as of this commit (https://github.com/matrix-org/synapse/commit/cb880815a9cebafa37c3ef1aca0a17bdae6c03c7)

synapse/handlers/federation.py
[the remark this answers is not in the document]
DMRobertson: 1f0a63e6b57ca94775408c3b3ad81ff0df3dbceb changes the situation somewhat. Let's take a look at the whole when I get to the last commit...

synapse/handlers/federation.py
richvdh: right, sorry, yes, this was subtle, and I should have split it out or made it clearer.

As you note, `extremities` it gets passed into `backfill`, which does nothing but iterate over it. And iterating a `dict` iterates that dict's keys. So, modulo deduplication not happening (which should be ok, because we shouldn't have duplicates here anyway), this is functionally identical.

synapse/handlers/federation.py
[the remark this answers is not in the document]
richvdh: In case it helps, this is talking about "filtering" in the sense of "a call to `filter_events_for_server`".

(and yes, that's really not clear when first reading.)

synapse/handlers/federation.py
richvdh: Yes, though it's worth noting that https://github.com/matrix-org/synapse/pull/12522/commits/1f0a63e6b57ca94775408c3b3ad81ff0df3dbceb changes this.

synapse/handlers/federation.py
richvdh: ugh, sorry. It should be `forward_event_ids`. This gets fixed up later in the patch set.

synapse/handlers/federation.py
```diff
@@ -265,31 +254,99 @@ async def _maybe_backfill_inner(
...
+        # enough good ones, and then stop.
+        #
+        # *Note*: the spec wants us to keep backfilling until we reach the start
+        # of the room in case we are allowed to see some of the history. However
+        # in practice that causes more issues than its worth, as a) its
+        # relatively rare for there to be any visible history and b) even when
+        # there is its often sufficiently long ago that clients would stop
+        # attempting to paginate before backfill reached the visible history.
+
+        extremities_to_request: Set[str] = set()
```
DMRobertson, line 298: Is it important that this is a set rather than list? I.e. do we expect this to do useful deduplication?

DMRobertson: We only ever add `bp.event_id`, and I think these are already duplicate-free by the construction of `sorted_backfill_points`?

richvdh: yeah, good point. We should probably switch to a List.

synapse/handlers/federation.py
```diff
@@ -265,31 +254,99 @@ async def _maybe_backfill_inner(
...
+        # However, we only want to paginate from a particular extremity if we can
+        # actually see the events we'll get, as otherwise we'll just spend a lot of
+        # resources to get redacted events.
+        #
+        # We do this by filtering all the backwards extremities and seeing if
+        # any remain.
+        #
+        # Doing this filtering can be expensive (we load the full state for the room
+        # at each of the sucessor events), so we check them one at a time until we find
+        # enough good ones, and then stop.
```
DMRobertson, line 289: Is the bracketed clause "we load... the sucessor events" still accurate, now that we include `INSERTION_POINT`s in this logic too?

synapse/handlers/federation.py
```diff
@@ -265,31 +254,99 @@ async def _maybe_backfill_inner(
...
-        extremities = dict(sorted_extremeties_tuple[:5])
+        # We still need to narrow down the list of extremities we pass to the remote
+        # server. We limit to 5 of them, to avoid the request URI becoming too long.
+        #
+        # However, we only want to paginate from a particular extremity if we can
+        # actually see the events we'll get, as otherwise we'll just spend a lot of
+        # resources to get redacted events.
+        #
+        # We do this by filtering all the backwards extremities and seeing if
+        # any remain.
```
DMRobertson, line 285: I still think filtering is an odd verb (even if it makes sense given that we call `filter_events_for_server`). Could this be rewritten? An attempt:

```
        # For performance's sake, we only want to paginate from a particular extremity
        # if we can actually see the events we'll get. Otherwise, we'd just spend a lot
        # of resources to get redacted events. We check each extremity in turn and
        # ignore those which don't merit backfilling.
        # 
        # Additionally, we limit ourselves to backfilling from at most 5 extremities,
        # for two reasons:
        # 
        # - The check which determines if we can see an extremity's events can be
        #   expensive (we load the full state for the room at each of the sucessor 
        #   events).
        # - We want to avoid the server-server API request URI becoming too long.
```

DMRobertson: (The point in https://github.com/matrix-org/synapse/pull/12522/files#r857783795 notwithstanding)

DMRobertson (approved):
I'm a little unsure about the a couple of comments (see discussion of "filtering" and "successor events", but I think the logical changes are sound. I'm not too fussed about the Set versus List comment.

I'll leave those minor things to your discretion; otherwise LGTM.

synapse/handlers/federation.py
DMRobertson: sgtm

synapse/handlers/federation.py
richvdh: I totally agree it's oddly phrased. It's just something I copied from the original implementation without stopping to think if it could be improved.  I'll update it.

richvdh: done in [8487963](https://github.com/matrix-org/synapse/pull/12522/commits/84879633516be303f5b795cb5aac3468da953d88)

changelog.d/12522.bugfix
```diff
@@ -0,0 +1 @@
+Fix a bug introduced in Synapse 0.99.3 which could cause Synapse to consume large amounts of RAM when back-paginating in a large room.
```
DMRobertson, line 1: Suppose we're in the worst case, where we have many backfill points to consider and none of them/their successor events are visible to us. Wouldn't we end up doing the same amount of work as before this change? (I.e. isn't this a bugfix for the best/average case rather than the worst?)

DMRobertson (approved):

changelog.d/12522.bugfix
richvdh: > Wouldn't we end up doing the same amount of work as before this change?

I think we won't try to load all state maps into memory *at the same time* (which we used to do in the depths of `filter_events_for_server`, in `_event_to_memberships`). So yeah, we might end up churning a bunch of CPU which we didn't before, but at least we won't OOM.

synapse/handlers/federation.py
```diff
@@ -92,6 +96,24 @@ def get_domains_from_state(state: StateMap[EventBase]) -> List[Tuple[str, int]]:
     return sorted(joined_domains.items(), key=lambda d: d[1])
 
 
+class _BackfillPointType(Enum):
+    # a regular backwards extremity (ie, an event which we don't yet have, but which
+    # is referred to by other events in the DAG)
+    BACKWARDS_EXTREMITY = enum.auto()
```
MadLittleMods, line 102: I think  "backward extremity" (not plural) is the correct spelling using the database as the source of truth (and the docs https://github.com/matrix-org/synapse/blob/develop/docs/development/room-dag-concepts.md#backward-extremity but I derived those from the code).

PR #12527, opened 2022-04-22
Add an index to config manual 

H-Shay:
Adds an index for the config manual so users can jump to headers. Note that this may end up being temporary as the the lack of an index could also be fixed in the website code, but for now this is the quickest and easiest way to fix this and the manual is not really usable without an index.

### DIFF
--- a/docs/usage/configuration/config_documentation.md
+++ b/docs/usage/configuration/config_documentation.md
@@ -64,7 +64,49 @@ apply if you want your config file to be read properly. A few helpful things to 
   In addition, each setting has an example of its usage, with the proper indentation
   shown. 
 
-  
+## Index
+[Modules](#modules)
+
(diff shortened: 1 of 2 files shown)


DMRobertson (approved):
Broadly LGTM. One nit, which I leave to your discretion.

Related: https://github.com/rust-lang/mdBook/issues/1523

docs/usage/configuration/config_documentation.md
```diff
@@ -64,7 +64,49 @@ apply if you want your config file to be read properly. A few helpful things to
   In addition, each setting has an example of its usage, with the proper indentation
   shown. 
 
-  
+## Index
```
DMRobertson, line 67: Nit: I think this might be a table of contents rather than an index?

PR #12528, opened 2022-04-22
disallow-untyped-defs in `docker` and `stubs` directories

### LINKED ISSUE #12190: Run mypy over docker directory
There's a couple of scripts in there which it would be nice to typecheck.

- change mypy.ini's `files` list to include `docker`
- run `mypy` and see what errors crop up
- fix them until none remain

There's the `contrib` directory too, but we don't promise to support that diligently.

DMRobertson:
Recommend commit-by-commit review.

I wasn't able to configure mypy to run with disallow-untyped-defs on these directories, so it'll be unenforced for now. (See also #11302.)

Resolves #12190.


docker/configure_workers_and_start.py
```diff
@@ -547,7 +549,7 @@ def generate_worker_log_config(
     return log_config_filepath
 
 
-def main(args, environ):
+def main(args: List[str], environ: MutableMapping[str, str]) -> None:
```
squahtx, line 552: I can't figure out why `environ` needs to be mutable?

stubs/txredisapi.pyi
```diff
@@ -34,11 +36,11 @@ class RedisProtocol(protocol.Protocol):
...
 class SubscriberProtocol(RedisProtocol):
-    def __init__(self, *args, **kwargs): ...
+    def __init__(self, *args: object, **kwargs: object): ...
     password: Optional[str]
-    def subscribe(self, channels: Union[str, List[str]]): ...
-    def connectionMade(self): ...
-    def connectionLost(self, reason): ...
+    def subscribe(self, channels: Union[str, List[str]]) -> "Deferred[None]": ...
+    def connectionMade(self) -> None: ...
+    def connectionLost(self, reason: Failure = ...) -> None: ...
```
squahtx, line 43: Where does the default value for `reason` come from?

docker/configure_workers_and_start.py
DMRobertson: We mutate the `environ`, see e.g.g +559; I was trying to be a light touch here and avoid logic/behaviour changes.

(I was surprised to learn that the `os.environ` mapping is mutable!)

squahtx: Ah, I missed that!

docker/start.py
```diff
@@ -83,7 +89,7 @@ def generate_config_from_template(config_dir, config_path, environ, ownership):
 
     # Convert SYNAPSE_NO_TLS to boolean if exists
     if "SYNAPSE_NO_TLS" in environ:
-        tlsanswerstring = str.lower(environ["SYNAPSE_NO_TLS"])
+        tlsanswerstring = str.lower(environ["SYNAPSE_NO_TLS"])  # type
```
squahtx, line 92: What does "# type" do?

stubs/txredisapi.pyi
DMRobertson: By default it's a failure which holds a `ConnectionDone` instance. If you feel like spelunking:

- https://twistedmatrix.com/documents/current/api/twisted.internet.protocol.Protocol.html#connectionLost
- https://github.com/twisted/twisted/blob/5ed041274f70285299bcb3dcf11da70604416f63/src/twisted/internet/protocol.py#L571
- https://github.com/twisted/twisted/blob/5ed041274f70285299bcb3dcf11da70604416f63/src/twisted/internet/protocol.py#L524
- https://github.com/twisted/twisted/blob/5ed041274f70285299bcb3dcf11da70604416f63/src/twisted/internet/error.py#L221

DMRobertson: Actually, that is for the abstract `Protocol`. Let me double check that txredisapi doesn't override that somehow...

docker/start.py
DMRobertson: Nothing. I think this was a `#type: ignore` I started adding but thought better of it. Will remove, thanks!

stubs/txredisapi.pyi
DMRobertson: Oh. It does: https://github.com/IlyaSkriblovsky/txredisapi/blob/c89a49829d1c48baa10d03c41d39c261c02ad6e1/txredisapi.py#L335. Moveover it seems like lots of the implementations of `Protocol` in twisted itself don't have a default for `why`.

In that case, let me remove the default argument.

docker/start.py
DMRobertson: https://github.com/matrix-org/synapse/pull/12528/commits/283d9614018a154f0570a43d2b2dde4dfac52b4c

stubs/txredisapi.pyi
DMRobertson: https://github.com/matrix-org/synapse/pull/12528/commits/b4198640f0fad24af70a43c982ecfe7ed6dad460

squahtx (approved):
lgtm!

PR #12529, opened 2022-04-22
Handle cancellation in `EventsWorkerStore._get_events_from_cache_or_db`

squahtx:
Multiple calls to `EventsWorkerStore._get_events_from_cache_or_db` can
reuse the same database fetch, which is initiated by the first call.
Ensure that cancelling the first call doesn't cancel the other calls
sharing the same database fetch.

Signed-off-by: Sean Quah <seanq@element.io>

### Pull Request Checklist

### DIFF
--- a/synapse/storage/databases/main/events_worker.py
+++ b/synapse/storage/databases/main/events_worker.py
@@ -75,7 +75,7 @@ from synapse.storage.util.id_generators import (
 from synapse.storage.util.sequence import build_sequence_generator
 from synapse.types import JsonDict, get_domain_from_id
 from synapse.util import unwrapFirstError
-from synapse.util.async_helpers import ObservableDeferred
+from synapse.util.async_helpers import ObservableDeferred, delay_cancellation
 from synapse.util.caches.descriptors import cached, cachedList
 from synapse.util.caches.lrucache import LruCache
 from synapse.util.iterutils import batch_iter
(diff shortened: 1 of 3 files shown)


synapse/storage/databases/main/events_worker.py
```diff
@@ -640,42 +640,57 @@ async def _get_events_from_cache_or_db(
...
+                        missing_events_ids,
+                    )
+                except Exception as e:
+                    with PreserveLoggingContext():
+                        fetching_deferred.errback(e)
+                    raise e
+                finally:
+                    # Ensure that we mark these events as no longer being fetched.
+                    for event_id in missing_events_ids:
+                        self._current_event_fetches.pop(event_id, None)
```
DMRobertson, line 680: Why `None` here? Would we ever expect some other async bit of execution to remove this event from `_current_event_fetches`?

synapse/storage/databases/main/events_worker.py
```diff
@@ -640,42 +640,57 @@ async def _get_events_from_cache_or_db(
...
+                # going to pull from the DB. We use a single deferred that resolves
+                # to all the events we pulled from the DB (this will result in this
+                # function returning more events than requested, but that can happen
+                # already due to `_get_events_from_db`).
+                fetching_deferred: ObservableDeferred[
+                    Dict[str, EventCacheEntry]
+                ] = ObservableDeferred(defer.Deferred(), consumeErrors=True)
                 for event_id in missing_events_ids:
-                    self._current_event_fetches.pop(event_id, None)
+                    self._current_event_fetches[event_id] = fetching_deferred
```
DMRobertson, line 662: What happens if `event_id` already exists in `self._current_event_fetches`? Perhaps this is never the case because of how `missing_events_ids` is constructed, line 640 in particular?

tests/storage/databases/main/test_events_worker.py
```diff
@@ -281,3 +282,119 @@ def test_recovery(self) -> None:
...
+
+            # Unblock the database fetch.
+            unblock.callback(None)
+            # A `CancelledError` should be raised out of the first `get_event` call.
+            exc = self.get_failure(get_event1, CancelledError).value
+            self.assertIsInstance(exc, CancelledError)
+            # The second `get_event` call should complete successfully.
+            self.get_success(get_event2)
+
+    def test_second_get_event_cancelled(self):
```
DMRobertson, line 386: What happens/should happen if both `get_event`s are cancelled? Should the database query be automatically cancelled somehow?

DMRobertson (commented):
Seems reasonable to me. What's the benefit of this change---prep work for having cancelled requests propagate through to cancelling the underlying queries?

synapse/storage/databases/main/events_worker.py
squahtx: I don't think that can happen because of how `missing_event_ids` is constructed: `already_fetching_ids` is `intersection(missing_event_ids, _current_event_fetches)`.
If it did manage to happen, it'd mean we're doing extra database fetches that could be avoided.

Note that all this is just the previous code, indented into a function.

synapse/storage/databases/main/events_worker.py
squahtx: I have no idea. It's from the previous version of the code, introduced in #10119.

`event_id` would only be missing if we managed to replace an existing fetch above, which we shouldn't do.

tests/storage/databases/main/test_events_worker.py
squahtx: That's a good question. Currently the database query will complete with nothing waiting for it. This is mainly for simplicity's sake. We've been following the same sort of pattern in other places where there are operations that are shared or risky to cancel.

Letting the cancellation go deeper is a lot more work. There is a pool of up to 3 event-fetcher threads which service queues of event ids. Each thread uses the same database connection for its entire lifetime. We'd need:
* a way to determine whether there's anything waiting for `get_missing_events_from_db`'s fetch
* logic to dequeue or skip cancelled items in the queue
* and optionally a way to cancel an in-flight db query without wrecking the event fetcher thread
tl;dr. improving the granularity of cancellations is a lot of extra work that I'm not yet convinced is worth it.

squahtx:
> What's the benefit of this change---prep work for having cancelled requests propagate through to cancelling the underlying queries?

It's to prevent the cancellation of a request from leaking over to a different unrelated request that happens to be retrieving the same events. ie. the first test case fails without this change. There's no plan to do any cancellation at the db query level right now.

synapse/storage/databases/main/events_worker.py
DMRobertson: > event_id would only be missing if we managed to replace an existing fetch above, which we shouldn't do.

That was my suspicion---thanks for confirming!

DMRobertson:
> It's to prevent the cancellation of a request from leaking over to a different unrelated request that happens to be retrieving the same events. ie. the first test case fails without this change. There's no plan to do any cancellation at the db query level right now.

Thanks, that's very clear. 

>  ie. the first test case fails without this change

Does this mean this change is a `.bugfix` rather than a `.misc` change?

squahtx:
> Does this mean this change is a .bugfix rather than a .misc change?

I think it's more of a feature than a bugfix. We're adding cancellation "support" to a place where there was previously none.
It depends on whether the intent of the previous code was to support cancellation, I guess?

squahtx:
Thanks for the review!

PR #12541, opened 2022-04-25
Mark Dockerfile as requiring BuildKit

jaywink:
The Synapse Dockerfile uses Buildkit and some GitLab CI environments at least seem to be confused about the extra arguments to some Dockerfile commands, if the version is not defined.

Docs: https://docs.docker.com/develop/develop-images/build_enhancements/#overriding-default-frontends


changelog.d/12541.misc
```diff
@@ -0,0 +1 @@
+Define Dockerfile version to fix issues with building images in some GitLab CI environments due to Buildkite.
```
DMRobertson, line 1: We've got a changelog section for the docker image. Can you rename this file to `12541.docker`?

DMRobertson:
```suggestion
Explicitly opt-in to using [BuildKit-specific features](https://github.com/moby/buildkit/blob/master/frontend/dockerfile/docs/syntax.md) in the Dockerfile. This fixes issues with building images in some GitLab CI environments.
```

DMRobertson (changes requested):
One changelog nitpick; otherwise, thank you for this!

Xref: Presumably the pain you experienced on GitLab was introduced by #11691.

